PR #2, opened 2019-04-10
Initial implementation of OCP install test

clacroix12:
Note that this includes some temporary changes to run.py to allow execution of the test case. Further changes to run.py will remove ceph specific functionality that is no longer needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ ven*
 config.yaml
 .vscode
 *.bak
+openshift-install
+data/pull-secret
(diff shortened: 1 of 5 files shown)


vasukulkarni:
Looking great!! , thanks!,  I see some commented out code, Do you want to remove the commented code?

clacroix12:
@vasukulkarni if it's in run.py there were some things i will end up removing but others that will remain with some changes. I didn't want to remove anything just yet as I will handle that in the next MR

PR #8, opened 2019-04-22
download binary based on os type and bail out sooner for unsupported …

vasukulkarni:
download binary based on os type and bail out sooner for unsupported windows platform.

### DIFF
--- a/run.py
+++ b/run.py
@@ -11,6 +11,7 @@ import textwrap
 import urllib3
 from docopt import docopt
 from getpass import getuser
+from ocsci.enums import ReturnCode
 from utility.utils import (
     timestamp, create_run_dir, create_report_portal_session, email_results,
     close_and_remove_filehandlers
@@ -119,8 +120,8 @@ def run(args):
 
     # Check if we are running on windows and bail out sooner
     if platform.system() == "Windows":
-        log.info("OS not supported by Openshift Installer")
-        return 1
+        log.error("Windows OS is not supported by Openshift Installer")
+        return ReturnCode.UNSUPPORTED_WINDOWS_RUN
 
     if suite_file:
         suites_path = os.path.abspath(suite_file)
@@ -210,5 +211,5 @@ def run(args):
 if __name__ == '__main__':
     args = docopt(doc)
     rc = run(args)
-    log.info("Final rc of test run %d" % rc)
-    sys.exit(rc)
+    log.info("Final rc of test run %d", rc.value)
+    sys.exit(rc.value)
--- a/run.py
+++ b/run.py
@@ -5,6 +5,7 @@ monkey.patch_all()
 import yaml
 import sys
 import os
+import platform
 import logging
 import textwrap
 import urllib3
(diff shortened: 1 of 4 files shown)


vasukulkarni:
Tested on Mac/Darwin

run.py
```diff
@@ -109,6 +110,11 @@ def run(args):
     if cleanup_name:
         pass  # TODO: cleanup cluster and skip test execution
 
+    # Check if we are running on windows and bail out sooner
+    if platform.system() == "Windows":
+        log.info("OS not supported by Openshift Installer")
+        return 1
```
RazTamir: @vasukulkarni , I'm thinking about managed return codes like exceptions so we will be able to handle them outside of the test itself.
I still don't have a full vision of how to use it but think it will be handy in the future.
In this case,
ERRWINDOWS = <any non zero number>

return ERRWINDOWS

WDYT?

vasukulkarni: @RazTamir yep that makes sense to me, we will have to build the list as we go, we can start with something small for now.

RazTamir: And of course I forgot Petr comment:

```python
ERRWINDOWS = 15 # Any non zero return code

return ERRWINDOWS
```

clacroix12: @RazTamir are you looking for something like an Enum here? I am experimenting with something similar for test status codes.

Ex.
```
class TestStatus(Enum):
    PASSED = 0
    FAILED = 1
    NOT_EXECUTED = 2
    SKIPPED = 3
```

tests/test_ocs_basic_install.py
```diff
@@ -59,7 +60,13 @@ def run(**kwargs):
     else:
         log.info("Downloading openshift installer")
         ver = config.get('installer-version')
-        url = f"https://github.com/openshift/installer/releases/download/{ver}/openshift-install-linux-amd64"
+        if platform.system() == "Darwin":
+            os_type = "darwin"
+        elif platform.system() == "Linux":
+            os_type = "linux"
+        else:
```
clacroix12: Should we raise an Exception here instead of setting to None? The download will fail anyways.

run.py
vasukulkarni: @clacroix12 I know brought this up today, I am thinking Enum as well, ex: TestExitErrorCodes, different from your class but can be in the same file.

tests/test_ocs_basic_install.py
vasukulkarni: I thought of that, then there are 3 options Darwin, Linux and Windows. Windows will exit sooner, I can still add an exception here we can check when it reaches here.

run.py
RazTamir: Yes @clacroix12

run.py
```diff
@@ -109,6 +110,11 @@ def run(args):
     if cleanup_name:
         pass  # TODO: cleanup cluster and skip test execution
 
+    # Check if we are running on windows and bail out sooner
+    if platform.system() == "Windows":
+        log.info("OS not supported by Openshift Installer")
```
petr-balogh: This should be error log level. And please put OS name in the message.

tests/test_ocs_basic_install.py
clacroix12: As long as we are just raising an exception, the runner should handle the rest for us. It will except it, log, fail the test, and move on if not configured to abort on failure.

PR #14, opened 2019-04-24
TestCase class for runner to utilize per test in suite

clacroix12:

### DIFF
--- a/ocsci/framework.py
+++ b/ocsci/framework.py
@@ -14,9 +14,9 @@ log = logging.getLogger(__name__)
 
 class TestCase(object):
     """
-        Generic test case object to contain configuration details from suite.
-        Handles test execution as well as pre and post execution steps
-        for Report Portal and Polarion.
+    Generic test case object to contain configuration details from suite.
+    Handles test execution as well as pre and post execution steps
+    for Report Portal and Polarion.
 
     Attributes
         name (str): name of test case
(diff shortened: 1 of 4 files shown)


ocsci/framework.py
```diff
@@ -0,0 +1,144 @@
...
+log = logging.getLogger(__name__)
+
+
+class TestCase(object):
+    """
+        Generic test case object to contain configuration details from suite.
+        Handles test execution as well as pre and post execution steps
+        for Report Portal and Polarion.
+
+    Attributes
```
petr-balogh: Args:

ocsci/framework.py
```diff
@@ -0,0 +1,144 @@
...
+from utility.polarion import post_to_polarion
+from utility.utils import create_unique_test_name, configure_logger, timestamp
+from .enums import TestStatus
+
+log = logging.getLogger(__name__)
+
+
+class TestCase(object):
+    """
+        Generic test case object to contain configuration details from suite.
```
petr-balogh: One indent left for this text block

ocsci/framework.py
clacroix12: These are not arguments though, args are outlined in the pydoc for the `__init__`. These are object attributes. Not all are passed to the `__init__` upon instantiation.

ocsci/framework.py
clacroix12: fixed

PR #17, opened 2019-04-30
Fix new url for download installer

petr-balogh:
Cause they changed the path of installer, changed also here.

### DIFF
--- a/tests/test_ocs_basic_install.py
+++ b/tests/test_ocs_basic_install.py
@@ -56,20 +56,24 @@ def run(**kwargs):
 
     # Download installer
     installer_filename = "openshift-install"
+    tarball = f"{installer_filename}.tar.gz"
     if os.path.isfile(installer_filename):
         log.info("Installer exists, skipping download")
     else:
         log.info("Downloading openshift installer")
-        ver = config.get('installer-version')
+        ver = config.get('installer-version', '4.1.0-rc.0')
         if platform.system() == "Darwin":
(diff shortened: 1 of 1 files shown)


tests/test_ocs_basic_install.py
```diff
@@ -56,20 +56,24 @@ def run(**kwargs):
 
     # Download installer
     installer_filename = "openshift-install"
+    tarball = f"{installer_filename}.tar.gz"
     if os.path.isfile(installer_filename):
         log.info("Installer exists, skipping download")
     else:
         log.info("Downloading openshift installer")
-        ver = config.get('installer-version')
+        ver = config.get('installer-version', '4.1.0-rc.0')
```
RazTamir: Can we pass this parameter through the yaml file this test is executed from? or at least maintain this parameter in a config file so we won't need to change here and in different files in the future?

clacroix12: That's where this value is pulled from. `config` is the config data in the suite file. The value `4.1.0-rc.0` is just the default value in case it's not set in the suite.

RazTamir: Yep,
I was talking about this hard-coded value. 
Because it is a default in case the version doesn't exist, let's maintain it in a proper way

vasukulkarni: I will add a trello card to set some values as default that can come in either via yaml or python file.

PR #18, opened 2019-04-30
jenkins_rc value and other fixes

clacroix12:
Fixed:
- jenkins_rc value at the end of the run
- suite config value being passed to TestCases
- test abort_on_fail scenario condition

### DIFF
--- a/run.py
+++ b/run.py
@@ -11,7 +11,7 @@ import textwrap
 import urllib3
 from docopt import docopt
 from getpass import getuser
-from ocsci.enums import ReturnCode
+from ocsci.enums import ReturnCode, TestStatus
 from utility.utils import (
     timestamp, create_run_dir, create_report_portal_session, email_results,
     close_and_remove_filehandlers, configure_logger,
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -172,13 +172,13 @@ def run(args):
...
     tcs = []
-    jenkins_rc = 0
+    jenkins_rc = TestStatus.FAILED
     test_data = dict()
     if cluster_name:
         test_data['cluster-name'] = cluster_name
 
     for test in tests:
-        config = test.get('config', {})
+        config = test.get('test').get('config', {})
```
petr-balogh: if the test won't be defined here test.get('test') and it will return None as default value for get you will fail with next one .get as None type has no method called get.

vasukulkarni: Its not possible to reach test which is None, since tests is a list of test from yaml.

clacroix12: @petr-balogh is there a better way for me to access that sub-key without hitting an exception if the initial one doesn't exist? It's really not a big deal here because like vasu said the only way for us to hit that is for you to be using a malformed suite file. However, I would like to know if there's a better approach to this general issue.

vasukulkarni: I can only think of nested if checks, but really dont see much value trying to check for malformed item which can be a dict or something else too. The review or test runs should throw them out.

petr-balogh: I see, thanks for explenation.  @clacroix12 not sure about elegant way.. I usually used .get('key', {}).get('key2', default) .

PR #19, opened 2019-05-01
Add defaults.py file

RazTamir:
Added INSTALLER_VERSION
 and few more code guidelines fixes

### DIFF
--- a/ocs/defaults.py
+++ b/ocs/defaults.py
@@ -1,0 +1,4 @@
+"""
+Defaults module
+"""
+INSTALLER_VERSION = '4.1.0-rc.0'
--- a/tests/test_ocs_basic_install.py
+++ b/tests/test_ocs_basic_install.py
@@ -5,6 +5,7 @@ import random
 import subprocess
 import sys
 
+import ocs.defaults as default
 import requests
 import yaml
 from ocs.exceptions import UnsupportedOSType
@@ -62,7 +63,7 @@ def run(**kwargs):
         log.info("Installer exists, skipping download")
     else:
         log.info("Downloading openshift installer")
-        ver = config.get('installer-version', '4.1.0-rc.0')
+        ver = config.get('installer-version', default.INSTALLER_VERSION)
         if platform.system() == "Darwin":
             os_type = "mac"
         elif platform.system() == "Linux":
@@ -78,7 +79,10 @@ def run(**kwargs):
 
     # Deploy cluster
     log.info("Deploying cluster")
-    run_cmd(f"./openshift-install create cluster --dir {cluster_path} --log-level debug")
+    run_cmd(
+        f"./openshift-install create cluster --dir {cluster_path} "
+        f"--log-level debug"
+    )
 
     # Test cluster access
     log.info("Testing access to cluster")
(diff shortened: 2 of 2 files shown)


tests/test_ocs_basic_install.py
```diff
@@ -114,9 +124,13 @@ def run_cmd(cmd, **kwargs):
```
ebenahar: Docsring is missing "Raises"

tests/test_ocs_basic_install.py
```diff
@@ -114,9 +124,13 @@ def run_cmd(cmd, **kwargs):
 
     """
     log.info(f"Executing command: {cmd}")
-    r = subprocess.run(cmd.split(), stdout=sys.stdout, stderr=sys.stderr, **kwargs)
+    r = subprocess.run(
```
vasukulkarni: It fits in single line right and is following the guidelines?

RazTamir: it was 84 chars

tests/test_ocs_basic_install.py
RazTamir: Will fix in #21

vasukulkarni: I see this resolved in #21

vasukulkarni:
@RazTamir please rebase this on new master, since merge of #21 is causing conflict

RazTamir:
> @RazTamir please rebase this on new master, since merge of #21 is causing conflict

Done

ocs/defaults.py
```diff
@@ -0,0 +1,4 @@
+"""
+Defaults module
+"""
+INSTALLER_VERSION = '4.1.0-rc.0'
```
petr-balogh: @RazTamir I am just thinking about use yaml format for the defaults, Here we can load yaml format to DEFAULTS variable and from other modules we can do.

```python
from defaults import DEFAULTS

ver = config.get('installer-version', DEFAULTS['INSTALLER_VERSION'])
```

Advantage for it is that we can easily use it when rendering templates.
So we can do some other helper function which can do something like this:

```python
def get_updated_defaults(data={}):
"""
Returns updated defaults with provided data

Args:
    data (dict): Dictionary with new data (default: {})

Returns:
    dict: Default values with updated data
"""
defaults_copy = DEFAULTS.copy()  # deep copy maybe better here 
return defaults_copy.update(data)
```

so once rendering some template we can take all the data from defaults easily instead of naming them explicitly and just update with new data we need? WDYT about this approach?

RazTamir: @petr-balogh sounds good to me.
I'll abandon this change.

I won't be able to implement this new approach soon so if someone can take this task it will be great

petr-balogh: Maybe no need to abandon, we can start with what you have and just in the comment you can point to this discussion as TODO, so this can be done in new PR.

RazTamir: Ok, created this for future conversion to yaml

petr-balogh: @RazTamir ok, thanks, but let's also other ppl to give their 2 cents here. Maybe it's not good idea for all. @clacroix12 @vasukulkarni @dahorak @vavuthu @shylesh @ebenahar .

vasukulkarni: Taking this up in a next PR

PR #20, opened 2019-05-01
Fix Dict comprehension in openshift_ops.py

RazTamir:


oc/openshift_ops.py
```diff
@@ -1,13 +1,13 @@
...
 from kubernetes import config
 from openshift.dynamic import DynamicClient, exceptions
 
 log = logging.getLogger(__name__)
 
 
 class OCP(object):
-    """Class which contains various utility functions for interacting
+    """
+    Class which contains various utility functions for interacting
```
shylesh: Doc string start line is bit confusing. According to guidelines, google style has started doc string at the line where quotes start.
Could you please clarify on which one should we stick on to?

clacroix12: I believe most of our docstrings already have the quotes on lines by themselves, with the text in between them (see ex.). I personally think this way is cleaner and will take less work getting the entire repo to conform to our guidelines. We will want to add a section into our coding guidelines for this regardless of what we choose.

ex:
```
def my_example(ex_arg):
    """
    Example docstring.

    Args:
        ex_arg: an example arg

    Raises:
        ExampleException: you get it

    Returns:
        something
    """
    pass
```

RazTamir: Added here #22

shylesh: LGTM

petr-balogh:
And fixed docstring convention from #22

petr-balogh:
> And fixed docstring convention from #22

Sorry, was in wrong field when I did squash and merge.
This had to be in squashed commit message.

PR #21, opened 2019-05-01
Move CommandFailedException to exceptions.py

RazTamir:


ocs/exceptions.py
```diff
@@ -4,3 +4,7 @@ class CommandFailed(Exception):
 
 class UnsupportedOSType(Exception):
     pass
+
+
+class CommandFailedException(Exception):
```
vasukulkarni: we have CommandFailed Exception in the same file, probably this was overlooked before during the merge, we can use the original one and delete the extra in install.py

RazTamir: right,

Fixing now.
So now I just need to:
git commit (no --amend)
and push new commit?

clacroix12: @RazTamir yeah you can just push a new commit and upon merging to master if you choose to squash your commits into one github will let you do that

vasukulkarni:
@clacroix12 needs 2 approvals, please squash and merge

PR #22, opened 2019-05-01
Adding docstring to coding guidelines

RazTamir:


docs/coding_guidelines.md
```diff
@@ -74,6 +74,32 @@ print("This is old style of formatting and should be avoided:")
...
+* The description should be as follow:
+```python
+def my_example(arg1, arg2):
+    """
+    This function is for showing how a proper docstring should
+    look like
+
+    Args:
+        arg1 (<type of argument>): Description of what this argument is and if
+        the
```
clacroix12: Should "the" be indented in 4 spaces with the rest following it?

RazTamir: Thanks!

shylesh:
LGTM

docs/coding_guidelines.md
```diff
@@ -74,6 +74,31 @@ print("This is old style of formatting and should be avoided:")
...
+```python
+def my_example(arg1, arg2):
+    """
+    This function is for showing how a proper docstring should
+    look like
+
+    Args:
+        arg1 (<type of argument>): Description of what this argument is and if
+            the line is too long we will indent according to this (4 spaces)
+        arg2 (int): This argument is an integer
```
petr-balogh: Do you think that we should also document default values?

I am used to do:
```python
def my_example(arg1, arg2=10):
    """
    This function is for showing how a proper docstring should
    look like

    Args:
        arg1 (<type of argument>): Description of what this argument is and if
            the line is too long we will indent according to this (4 spaces)
        arg2 (int): This argument is an integer (default: 10)
```

RazTamir: Done

PR #23, opened 2019-05-02
Deployment of OCS with templates

clacroix12:

PR #24, opened 2019-05-03
Wip rook infra changes

shylesh:
Introducing 3 commits for Rook infrastructure class and related changes along with a simple test.

### DIFF
--- a/oc/openshift_ops.py
+++ b/oc/openshift_ops.py
@@ -22,6 +22,68 @@ class OCP(object):
         self.v1_projects = dyn_client.resources.get(
             api_version='project.openshift.io/v1', kind='Project'
         )
+        self.v1_pods = dyn_client.resources.get(
+            api_version='v1', kind='Pod'
+        )
+
+    def get_pods(self, **kw):
+        """
+        Get pods in specific namespace or across oc cluster.
+
+        Args:
+            **kw: ex: namespace=rook-ceph, label_selector='x==y'
+
+        Returns:
+            list: of pods names, if no namespace provided then this function
+                returns all pods across openshift cluster.
+        """
+        resource = self.v1_pods
+
+        try:
+            pod_data = resource.get(**kw)
+            log.info(pod_data)
+        except exceptions.NotFoundError:
+            log.error("Failed to get pods: resource not found.")
+            raise
+        except Exception:
+            log.error("Unexpected error.")
+            raise
+
+        return [item.metadata.name for item in pod_data.items]
+
+    def get_labels(self, pod_name, pod_namespace):
+        """
+        Get labels from specific pod
+
+        Args:
+            pod_name (str): Name of pod in oc cluster
+            pod_namespace (str): pod namespace in which the pod lives
+
+        Raises:
+            NotFoundError: If resource not found
+
+        Returns:
+            dict: All the openshift labels on a given pod
+        """
+
+        resource = self.v1_pods.status
+
+        try:
+            pod_meta = resource.get(
+                name=pod_name,
+                namespace=pod_namespace,
+            )
+        except exceptions.NotFoundError:
+            log.error("Failed to get pods: resource not found.")
+            raise
+        except Exception:
+            log.error("Unexpected error")
+            raise
+
+        data = pod_meta['metadata']['labels']
+        pod_labels = {k: v for k, v in data.items()}
+
+        return pod_labels
 
     def get_projects(self):
         """
--- a/ocs/pod.py
+++ b/ocs/pod.py
@@ -17,30 +17,29 @@ class Pod(object):
     """Handles per pod related context
 
         Attributes:
-            name (str):      name of the pod in oc cluster
-            namespace(str):  openshift namespace where this pod lives
-            labels (list):   list of oc labels associated with pod
+            _name (str):      name of the pod in oc cluster
+            _namespace(str):  openshift namespace where this pod lives
+            labels (dict):   A dict of oc labels associated with pod
             roles (list):    This could be oc roles like Master, etcd OR
                              ceph roles like mon, osd etc
 
 
     """
 
-    def __init__(self, name=None, namespace=None, labels=None, roles=None):
+    def __init__(self, name=None, namespace=None, labels=None, roles=[]):
         """Context detail per pod
 
             Args:
                 name (string):      name of the pod in oc cluster
                 namespace (string): namespace in which pod lives
-                labels (list):      list of oc labels associated with pod
+                labels (dict):      dictionary of oc labels associated with pod
                 roles (list):       This could be oc roles like Master, etcd OR
-                                   ceph roles like mon, osd etc
-
+                    ceph roles like mon, osd etc.
         """
         self._name = name
         self._namespace = namespace
-        self.labels = labels
-        self.roles = roles
+        self._labels = labels
+        self._roles = roles
         # TODO: get backend config !!
 
     @property
@@ -50,6 +49,22 @@ class Pod(object):
     @property
     def namespace(self):
         return self._namespace
+
+    @property
+    def roles(self):
+        return self._roles
+
+    @property
+    def labels(self):
+        return self._labels
+
+    def set_role(self, role):
+        """
+        Set a role for this pod
+        Args:
+            role (str): New role to be assigned for this pod
+        """
+        self._roles.append(role)
 
     def exec_command(self, **kw):
         """ Handles execution of a command on a pod
--- a/ocs/rook.py
+++ b/ocs/rook.py
@@ -1,0 +1,140 @@
+"""
+A module for all rook functionalities and abstractions.
+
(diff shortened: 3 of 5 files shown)


ocs/rook.py
```diff
@@ -0,0 +1,142 @@
...
+class RookCluster(object):
+    """
+    Handles all cluster related operations from ceph perspective
+
+    This class has depiction of ceph cluster. Contains references to
+    pod objects which represents ceph cluster entities.
+
+    Attributes:
+        _ocs_pods (list) : A list of  ceph cluster related pods
+
```
petr-balogh: no need for extra empty lines between attributes I think, Also if you are using dot at the end, use it on each of line or omit on each line.

ocs/rook.py
```diff
@@ -0,0 +1,142 @@
...
+        _cluster_name (str): Name of ceph cluster.
+
+        _namespace (str): openshift Namespace where this cluster lives.
+
+        _api_client (APIClient): api-client used for interacting with openshift.
+
+        _url_prefix (str): url prefix for REST api interaction
+
+    Methods:
```
petr-balogh: Where are those methods implemented? I don't think that we should have even document methods as they are documented in their own docstrings.

ocs/rook.py
```diff
@@ -0,0 +1,142 @@
...
+       add_mon(**kw) : add a new mon
+       remove_mon(id): Remove a mon with id id
+       add_osd(**kw):  add a new osd to cluster
+    """
+
+    def __init__(self, **config):
+        # cluster_name is name of cluster in rook of type CephCluster
+        self._cluster_name = config.get('cluster_name', 'rook-ceph')
+        self._namespace = config.get('namespace', 'rook-ceph')
```
petr-balogh: For cluster namespace we will be using probably openshift-storage as we have it as default in template as you can see:  https://github.com/red-hat-storage/ocs-ci/pull/23/files#diff-788816bc8d86d39b43365dfba5670b65R16 , So would be good to put it to defaults and use it from defaults

ocs/rook.py
```diff
@@ -0,0 +1,142 @@
...
+    def cluster_name(self):
+        return self._cluster_name
+
+    @property
+    def namespace(self):
+        return self._namespace
+
+    @property
+    def pods(self):
+        # TODO: Decide whether to return list or yield
```
petr-balogh: Not sure what advantage we will get here when using generator if we already have the list defined.

vasukulkarni: There is advantage when its a generator, each function call can process one required 'role' and yield that value instead of complete list, if someone wants full list from generator they can always call list()

oc/openshift_ops.py
```diff
@@ -22,6 +22,70 @@ def __init__(self):
...
+        """
+        Get pods in specific namespace or across oc cluster.
+
+        Args:
+            **kw: ex: namespace=rook-ceph, label_selector='x==y'
+
+        Returns:
+            list: of pods names
+
+            if no namespace provided then this function returns all pods
```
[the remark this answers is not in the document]
shylesh: Done

oc/openshift_ops.py
```diff
@@ -22,6 +22,70 @@ def __init__(self):
...
+            raise
+
+        return [item.metadata.name for item in pod_data.items]
+
+    def get_labels(self, pod_name, pod_namespace):
+        """
+        Get labels from specific pod
+
+        Args:
+            pod_name (str)
```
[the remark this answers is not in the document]
shylesh: Done

ocs/pod.py
```diff
@@ -30,17 +30,18 @@ def __init__(self, name=None, namespace=None, labels=None, roles=None):
         """Context detail per pod
 
             Args:
-                name (string):      name of the pod in oc cluster
-                namespace (string): namespace in which pod lives
-                labels (list):      list of oc labels associated with pod
-                roles (list):       This could be oc roles like Master, etcd OR
-                                   ceph roles like mon, osd etc
-
+                _name (string):      name of the pod in oc cluster
```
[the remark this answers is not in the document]
shylesh: Done

oc/openshift_ops.py
```diff
@@ -22,6 +22,70 @@ def __init__(self):
...
+            pod_namespace (str)
+
+        Raises:
+            NotFoundError: If resource not found
+
+        Returns:
+            labels (dict): All the labels (key and value)
```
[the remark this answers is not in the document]
shylesh: Done

ocs/pod.py
```diff
@@ -30,17 +30,18 @@ def __init__(self, name=None, namespace=None, labels=None, roles=None):
...
+                _labels (dict):      list of oc labels associated with pod
+                _roles (list):       This could be oc roles like Master, etcd OR
+                    ceph roles like mon, osd etc.
         """
         self._name = name
         self._namespace = namespace
-        self.labels = labels
-        self.roles = roles
+        self._labels = labels
+        if not roles:
```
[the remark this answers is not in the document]
shylesh: Originally it was like that but after referring to few "dangerous default value" threads I changed this way but I didn't realize that this function will be called only once so this should work. I changed it .

ocs/pod.py
```diff
@@ -51,6 +52,18 @@ def name(self):
     def namespace(self):
         return self._namespace
 
+    @property
+    def roles(self):
+        return self._roles
+
+    @roles.setter
+    def roles(self, role):
```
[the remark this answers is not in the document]
shylesh: Done, created a separate function set_role() for this purpose.

ocs/rook.py
shylesh: Done

ocs/rook.py
shylesh: Sorry, this was written for my reference but forgot to remove it. Done

ocs/rook.py
shylesh: This one I couldn't figure out where to put it and how to retrieve . Which template file will have this ? Could you please point me to an example.

ocs/rook.py
shylesh: As of now yield looks overkill but considering atleast 20+ pods in ceph cluster (going fwd cluster may grow) and mostly users will be iterating over pods for one or the other operation I made it as generator. If no objection then I will leave as is.

ocs/pod.py
```diff
@@ -17,30 +17,29 @@ class Pod(object):
...
 
-    def __init__(self, name=None, namespace=None, labels=None, roles=None):
+    def __init__(self, name=None, namespace=None, labels=None, roles=[]):
         """Context detail per pod
 
             Args:
                 name (string):      name of the pod in oc cluster
                 namespace (string): namespace in which pod lives
-                labels (list):      list of oc labels associated with pod
+                labels (dict):      list of oc labels associated with pod
```
RazTamir: Dictionary of oc labels associated with pod

ocs/rook.py
```diff
@@ -0,0 +1,133 @@
...
+            # Instantiate pod object for this pod
+            podobj = Pod(pod, self._namespace, pod_labels)
+            self._ocs_pods.append(podobj)
+
+    def _get_pod_labels(self, pod_name, namespace):
+        """
+        Get labels(openshift labels) on a pod
+
+        Args:
+            pod_name (str)
```
RazTamir: add description to both

ocs/rook.py
```diff
@@ -0,0 +1,133 @@
...
+            namespace (str)
+
+        Returns:
+            labels (dict): from pod metadata
```
RazTamir: Dict: Labels from pod metadata

shylesh:
I Apologize for force push because of rebase.

ocs/rook.py
shylesh: Done

ocs/rook.py
shylesh: Done

ocs/pod.py
shylesh: Done

ocs/rook.py
shylesh: @petr-balogh ,
After further discussion with Vasu, I went ahead with defaults.py approach. Since rook object has to be instantiated somewhere after deployment we thought we will do reading from config there instead of doing it inside RookCluster class.

ocs/rook.py
shylesh: I assume we are ok with this.

vasukulkarni:
FYI: Shylesh has to squash them appropriately before this can be merged.

ocs/rook.py
vasukulkarni: We will have to keep the defaults in defaults.py, I will add the trello on how we can just use one place for defaults and move few things from template to here, redenering the template with defaults is something we have to look into.

ocs/pod.py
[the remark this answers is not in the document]
vasukulkarni: I see this is resolved too.

PR #26, opened 2019-05-07
Add CODEOWNERS file for auto assign reviewers

petr-balogh:


petr-balogh:
> LGTM

Please go to review and approve, so you have to see the green check button next to your name in Reviewers section, we need to have at least of two of those.  I see:  RazTamir reviewed on behalf of red-hat-storage/top-level-reviewers 4 minutes ago so not sure how it will be good cause I saw that group disappeared from reviewers replaced by you.

PR #27, opened 2019-05-07
Adding functionality to run on existing cluster

petr-balogh:


docs/usage.md
```diff
@@ -5,26 +5,42 @@
...
 ## Examples
 
-Run OCS install suite:
+* Run OCS install suite:
 
 ```bash
-python run.py --cred ~/aws.yml --suite suites/ocs_basic_install.yml --log-level info
-```
\ No newline at end of file
+python3 run.py --cred ~/aws.yml --suite suites/ocs_basic_install.yml \
```
clacroix12: Same here (and code block below) and the one below, we should really encourage using a virtualenv instead of the system installed python3.

petr-balogh: I am using venv, but wasn't sure about how it's handel python2 or python3 . Thanks for pointing it out.. Now I really see that:

```bash
(.venv) [jenkins@qe-ocs4-slave-prod-3 ocs-ci]$ which python
~/ocs-ci/.venv/bin/python
(.venv) [jenkins@qe-ocs4-slave-prod-3 ocs-ci]$ python --version
Python 3.6.6
```

tests/test_ocs_basic_install.py
```diff
@@ -91,9 +96,10 @@ def run(**kwargs):
     )
 
     # Test cluster access
-    log.info("Testing access to cluster")
-    os.environ['KUBECONFIG'] = f"{cluster_path}/auth/kubeconfig"
-    run_cmd("oc cluster-info")
+    if not OCP.set_kubeconfig(
```
vasukulkarni: Can we check this at the begining of the test and we will be able to skip quite a bit of code if we are using existing cluster based on cli?

docs/usage.md
```diff
@@ -12,19 +12,35 @@ python run.py --help
...
 
 ## Useful Arguments
 
 Some non-required arguments that we end up using a lot:
 
 * `--log-level <level>` - set the log level that is output to stdout.
+* `--cluster-name <name>` - name of cluster.
+* `--cluster-path <path>` - path where to create the directory which will
+    contains all the installation/authentication information about cluster.
+    `Use this parameter when running on already deployed cluster!` You can
```
vavuthu: for running on already deployed cluster, are we not using --reuse option

tests/test_ocs_basic_install.py
vasukulkarni: Absolutely there is no need for that,  so there are 2 ways to handle this,  (a) either skip this in runner and set the kubecontext (b) do it at the start of the test and the email results will report this test as pass,  We can take the approach (b) for simplicity.

docs/usage.md
vasukulkarni: @vavuthu reuse option is gone,  we can bring that when we are going to test with internal openstack and the purpose was not to create VMs and other resources again, I think the subtle difference here is  'use existing ocs cluster'  vs 'existing vm's and resources'

clacroix12:
@petr-balogh When we say "cluster" do we mean OCP or OCS? Is the intention here to install OCS on an existing OCP cluster or to execute tests against an existing OCS installation?

petr-balogh:
Intention here is run whathever else test on already intalled ocp and ocs cluster as we do it at once now. And not run installation again once installed. @clacroix12

clacroix12:
If that is the case, shouldn't we be skipping the install somehow in these changes? I'm not seeing how we do that. We should probably be reading in the cluster_path, testing connectivity, and returning TestStatus.SKIPPED. @petr-balogh

petr-balogh:
I wasn't sure how we will have test suites defined as I saw only one test suite. Will each test suite contains also installation? If so we should do that, if suite will cointain separate tests than you can run like it's now. I wanted to discuss this on tomorrow´s mtg. Cause I thought that we willl run the installation in separate run and then we will continue with run in different stage of job which will run the tests in diff run.

clacroix12:
The idea is that if a suite calls for an OCS installation it will include that as one of it's tests in the suite file. It won't be called automatically or anything like that. Additional tests will be included in the suite to be executed after the installation if the suite requires it. Suites should be self-contained and not require any other suites to be run in order to be executed.

vasukulkarni:
> If that is the case, shouldn't we be skipping the install somehow in these changes? I'm not seeing how we do that. We should probably be reading in the cluster_path, testing connectivity, and returning TestStatus.SKIPPED. @petr-balogh

We can just check this at the start of the test and use the above TestStatus.SKIPPED

tests/test_ocs_basic_install.py
```diff
@@ -34,19 +36,27 @@ def run(**kwargs):
...
     cluster_dir_parent = "/tmp"
-    base_name = test_data.get('cluster-name', 'ocs-ci-cluster')
+    cluster_name = test_data.get('cluster-name')
+    cluster_path = test_data.get('cluster-path')
     cid = random.randint(10000, 99999)
-    cluster_name = f'{base_name}-{cid}'
-
-    cluster_path = os.path.join(cluster_dir_parent, cluster_name)
+    if not (cluster_name and cluster_path):
+        cluster_name = f"{default.CLUSTER_NAME}-{cid}"
```
mbukatov: Why is necessary to reset `cluster_name` when `cluster_path` is not defined? I'm just curious.

petr-balogh: good catch, I wanted yesterday to short this one but didn't realize this default name there. Thanks, will fix in next commit

petr-balogh: @mbukatov I hope I found the solution for this in next commit. Can you please take a look if it makes sense now? Thanks

vasukulkarni:
@clacroix12 anything from your side? If not we can merge I think

tests/test_ocs_basic_install.py
```diff
@@ -34,19 +36,28 @@ def run(**kwargs):
     log.info("Generating install-config")
     # TODO: determine better place to create cluster directories - (log dir?)
     cluster_dir_parent = "/tmp"
-    base_name = test_data.get('cluster-name', 'ocs-ci-cluster')
+    cluster_name = test_data.get('cluster-name')
```
clacroix12: Should we be requiring the user to pass both the cluster-name as well as the directory? If we have the directory we can infer the name based on that. I think this whole block would be cleaner if we started with a check to see if the `cluster_path` was provided. If so we set the name in the test data as well, test access to the cluster, and return SKIPPED. Otherwise, if no `cluster_path` is set, we can just do what we were doing before with the addition of setting the default value to `default.CLUSTER_NAME`.

petr-balogh: @clacroix12 the purpose of this was mainly cause of automation runs on Jenkins slaves. I would like to have some directory specified like /home/jenkins/cluster-dir  - as you can see I don't have any name of the cluster in the name of directory.  The jenkins node can during it's deployment create a cluster directory in the share mount point, so /home/jenkins/cluster-dir will be on all the slaves all the time the same (You always know where to find it), and we can create //nfs_share:/clusters_data/unique_cluster_name_id/ which will be mounted in /home/jenkins/cluster-dir . With solution with just cluster_path and taking the name from this directory.

petr-balogh: @vasukulkarni @RazTamir @dahorak  can you please take a look to this conversation and agree? So we can merge and move on or just decide to do it diff way and I can change the logic. But as I said as we will have one jenkins node per cluster, and till cluster is up we will have jenkins slave up, I prefer to have the data about the cluster in for example mentioned folder /home/jenkins/cluster-dir

RazTamir: From my POV, this file 'test_ocs_basic_install.py' should be used only for deployment purposes and not part of the testing phase, hence, the user shouldn't have an already existing cluster/path/directory.
When running tests locally, we can pass the path to the AWS directory but it seems like not a convenient way

petr-balogh: Maybe I got the whole thing you meant @clacroix12 Is this what you've meant: https://github.com/red-hat-storage/ocs-ci/pull/27/commits/8b84809100a0211ff5e08a5f919a3036098b396b move the check at the start?

dahorak: From my POV, I agree with the flexibility provided by the suggested changes... but my experience with this framework is still very limited, so maybe I'm missing some crucial context.

clacroix12: @petr-balogh yes that's essentially what I meant. However since you have the `cluster_path` and `OCP.set_kubeconfig` paired up in that `if` we will only return skipped if the cluster is ready. In the even that the cluster_path is set and in a bad state, we will continue with a deployment. Is that intended?

clacroix12: @RazTamir I'm not sure I agree. The installation of OCP is something we will be testing as well, is it not? I don't see why this shouldn't be treated as a test.

petr-balogh: @clacroix12 actually it's just checking if OCP is up and running and you are able to connect with credentials you have in cluster_path.  I am not checking any other state of the OCS cluster and assume that user know that cluster is up and running. (We can improve in next PR to check also OCS cluster and then decide what to do if it's not ready) But as I said, I want to be able to have cluster_path specified for jenkins slave in for example /home/jenkins/ocs_cluster_dir and even in first run when cluster doesn't exist, it will be created from scratch. It's up to user to have cluster in good shape to run on when decided to run on existing cluster.

petr-balogh: @clacroix12 as there are another PRs kind of depend on this change or can reuse some parts from it, I would go and merge this one if you and also @vasukulkarni approve. You can merge and align those kind of depending PRs.  I can do another requests if there are some in new PR, WDYT?

vasukulkarni:
I have already approved this one, but it got dismissed die to push. Once you resolve the conversation with Coady, Feel free to merge this.

petr-balogh:
> I have already approved this one, but it got dismissed due to push. Once you resolve the conversation with Coady, Feel free to merge this.

@vasukulkarni yep, I know, we have it set in setting of this repo that it should be approved again after push of other change cause you know, every change can introduce another bug which can other ppl see.  So just please approve again so I would like to follow the rules we have set to have two approves before merge. Thanks

tests/test_ocs_basic_install.py
clacroix12: Right, I guess what I mean is if they specify a cluster-path for a cluster that isn't running, we will go forward with the installation and most likely hit other issues because we're using the same directory as another install. I think we can merge this but we should really revisit this logic here and make sure we aren't going to hit any weird cases like that.

PR #28, opened 2019-05-09
Ignore venv for flake8 in tox

petr-balogh:


clacroix12:
Was tox using the venv you created to run ocs-ci? It has always created it's own for each env in the envlist for me. @petr-balogh

petr-balogh:
> Was tox using the venv you created to run ocs-ci? It has always created it's own for each env in the envlist for me. @petr-balogh

I created .venv in ocs-ci root and when I ran locally on my mac it looked like tox is stuck and after few minutes it started showing it going over all files from venv so I decided to put it rather here and whit this option it works for me.

clacroix12:
@petr-balogh makes sense to me! I was just curious as to what prompted this change.

docs/getting_started.md
```diff
@@ -10,9 +10,15 @@
...
 It is recommended that you use a python virtual environment to install the necessary dependencies
 
-1. Setup a python 3.7 virtual environment. This is actually quite easy to do now.
-    * `python3.7 -m venv <path/to/venv>`
-    * `source <path/to/venv>/bin/activate`
+1. Setup a python 3.7 virtual environment. This is actually quite easy to do
+    now.
+
+> Use hidden `.venv` or normal `venv` folder for virtual env as we are
+> ignoring this in flake8 configurtation in tox.
```
mbukatov: Sorry for syntax nitpicking, but it would help if this block of text were aligned with the previous text, removing the blockquote marks. That way, the following list would not be considered to be a code block.

petr-balogh: Wanted to emphasize this by blockquote but I don't mind to let it as part of the previous text

PR #29, opened 2019-05-09
Implement TestStatus properly for our usage

petr-balogh:


ocsci/enums.py
```diff
@@ -2,10 +2,13 @@
...
 
 class TestStatus(Enum):
+    # As we have function to transform RC to TestStatus Enum we should add new
+    # statuses with negative number to avoid miss tranforming from Unix RC to
+    # some of our statuses!
     PASSED = 0
     FAILED = 1
-    NOT_EXECUTED = 2
-    SKIPPED = 3
+    NOT_EXECUTED = -1
```
shylesh: I still see a problem here. As you said user can directly return ret value of shell script then he is also free to return ret value of some program which is executed through exec.
In this case ret values could be -1 (ex: read sys call etc), so do you thin you want to allow user to return straight away whetever the ret value they got directly from execution ? (either through shell or through exec).

mbukatov: For return code of a shell script or any other process executed from the test, clash is not possible, because if you try to return a negative number in any unix process, it will be converted to single byte unsigned integer.

This clash could happen only if one uses a return code from some python function which uses negative numbers to report errors.

shylesh: > For return code of a shell script or any other process executed from the test, clash is not possible, because if you try to return a negative number in any unix process, it will be converted to single byte unsigned integer.
> 
> This clash could happen only if one uses a return code from some python function which uses negative numbers to report errors.

Yes, you are right. Even ret values from exec is translated into unsigned int. So this should not be a problem.

ocsci/framework.py
```diff
@@ -101,6 +102,22 @@ def _setup(self):
...
             )
 
+    @staticmethod
+    def rc_to_status(rc):
+        """
+        Transform int unix return code to TestStatus
+
+        Args:
+            rc (int): Return code
+        Returns:
```
ebenahar: Empty line is missing before 'Returns'

petr-balogh: Done

petr-balogh:
@clacroix12 @vasukulkarni I just fixed the comment from Martin, can you please approve again and merge? Thx

PR #30, opened 2019-05-09
Set abort-on-fail to true for ocs install

clacroix12:


vasukulkarni:
Ahh, Sorry I check in framework.py, so the check already exists. Thanks

PR #31, opened 2019-05-09
Added destroy cluster functionality

clacroix12:


ocsci/framework.py
```diff
@@ -141,4 +144,9 @@ def _teardown(self):
             # TODO: update post_to_polarion to look for correct keys
         self.end = time.time()
         self.duration = self.end - self.start
+
+        if self.destroy_cluster:
```
petr-balogh: Do I understand that it will be called after deployment finish in its teardown and if any other test is in test suite it cannot succeed cause we will destroy cluster before continue to next test case?

utility/utils.py
```diff
@@ -662,3 +665,55 @@ def download_file(url, filename):
         r = requests.get(url)
         f.write(r.content)
     assert r.ok
+
+
+def destroy_cluster(cluster_path):
```
petr-balogh: what about to have it as another test case? Then we can skip it depends on some parameter?

ocsci/framework.py
clacroix12: Yes you are correct if those subsequent tests do not create a new cluster, so there is some responsibility on those creating the suite to not opt into destroying their cluster before they perform tests on it. That's why this defaults to False.

utility/utils.py
clacroix12: What do we benefit from it being it's own test case? I can see that complicating how we call it when we use the CLI usage pattern, as well as adding more than just a single line to a suite file in order to destroy the cluster after a certain test.

run.py
```diff
@@ -53,6 +53,8 @@
   --log-level <LEVEL>               Set logging level
   --cluster-name <name>             Name that will be used for cluster creation
   --no-email                        Do not send results email
+  --destroy <PATH>                  Destroy cluster using given path to the
```
petr-balogh: Just thinking about why to specify path here when we have parameter cluster-path in this PR https://github.com/red-hat-storage/ocs-ci/pull/27/files#diff-4be2078a125fe364f7198675d51bc688R26 . cannot we use this parameter and --destroy will be just true or false depends if defined?

utility/utils.py
```diff
@@ -662,3 +665,55 @@ def download_file(url, filename):
...
+        metadata_file = os.path.join(cluster_path, "metadata.json")
+        with open(metadata_file) as f:
+            metadata = json.loads(f.read())
+        cluster_name = metadata.get("clusterName")
+        region_name = metadata.get("aws").get("region")
+
+        # Execute destroy cluster using OpenShift installer
+        log.info(f"Destroying cluster defined in {cluster_path}")
+        run_cmd(destroy_cmd)
+
```
vavuthu: can we have check whether openshift installer is present or not. If not present we need to download it again. The reason why i am asking is if the user clones the repo and use the destroy functionality, it will fail.

Also we can we make download installer as separate lib in utility with version as parameter like
def download_installer(installer_version):

utility/utils.py
```diff
@@ -662,3 +665,55 @@ def download_file(url, filename):
...
+        # Find and delete volumes
+        aws = AWS(region_name)
+        volume_pattern = f"{cluster_name}*"
+        log.debug(f"Finding volumes with pattern: {volume_pattern}")
+        volumes = aws.get_volumes_by_name_pattern(volume_pattern)
```
vavuthu: It might work with cluster deployed with ocs-ci framework. if the user added the EBS with different name, then this logic won't help.

instead we can get instances of all workers, then delete the volumes associated with instance

ocsci/framework.py
petr-balogh: Yep, but I was thinking that better approach would be to allow user to do this destroy cluster at the end of the execution. If I understand correctly now there are two options:

1) set destroy-cluster in test suite yaml file for particular test case (user will specify to last one I guess), if they will add one test to test suite they will need to move this param to new last test case.
2) do the destroy cluster at the start of the execution with --destroy parameter

But what if user would like to run whole test suite and decide not to destroy at the end. If destroy is defined in test suite yaml file, there is currently no way to skip destroy right?

So if we would like to support this, we probably will have to have two params. --destroy and --no-destroy or --no-destroy and destroy by default and the end of execution and --destroy-at-start to call the destroy at the start as you do.

utility/utils.py
petr-balogh: ok, I see now when I looked at PR with fresh brain after night.

utility/utils.py
clacroix12: I don't think we should be trying to have ocs-ci support every use case, especially those where people manually intervened with how the cluster is created. I think ocs-ci should be able to clean up after the resources that it creates. If we start to worry about every single thing that a user can do outside of ocs-ci the scope of this operation becomes quite large.

utility/utils.py
clacroix12: Downloading the installer should definitely be it's own utility, I agree. I can break that out and then call it in the event that the installer doesn't exist. This will be useful since we clean up the installer after destroying the cluster as well.

run.py
clacroix12: `--cluster-path` doesn't exist on master yet which is why I didn't use it :) We can change this to have `--destroy` set a boolean and pair it up with `--cluster-path` if that is better.

ocsci/framework.py
clacroix12: I think your approach of having the destroy be it's own test would be better. This makes it less clunky to add additional tests into an established suite. Also providing a `--no-destroy` option that the test could read in and return TestStatus.SKIPPED in the event of would be nice. This would allow users to leave their clusters up without having to manually edit the suite file before running, as well as let us trigger this through jenkins more easily.

utility/utils.py
clacroix12: Thinking about this again, it really doesn't change how we call it from the CLI side. The test case however will basically just end up calling the function in the same way.

utility/utils.py
vasukulkarni: Yeah we cannot handle things done outside ocs-ci, there is no clear direction what is needed and what is not.

vasukulkarni:
I think `--no-destroy` option should be a different PR, since the logic there is to override the value set in the test suite.

clacroix12:
@vasukulkarni it's actually pretty minimal. Basically if that option is set we add a value for it in the test_data, which the test reads in and skips if it should. I have it changed locally but need to test still

ocsci/framework.py
clacroix12: One thing to add to this however is that if the destroy is it's own test, and we hit issues during our installation, we have the test set to 'abort-on-fail.' This will stop us from getting to the destroy test as we stop test execution. Just something to think about, and a tradeoff if we decide to go with the 'destroy in its own test' approach.

clacroix12:
I expect some merge conflicts after #27 is merged to master but this is what I have at this point after addressing the concerns brought up with the original commit.

ocsci/framework.py
RazTamir: From my side, I don't think that destroy-cluster should be in the test cases level at all as this is an infrastructure operation (logic wise).
In case someone wants to destroy it's cluster it should be a different phase in Jenkins and not part of the test suite

utility/utils.py
RazTamir: I agree with Coady here, in case some tests are creating AWS/Azure/etc. volumes outside of the deplyoment, the teardown phase of the relevant test should take care of cleaning the additional objects

petr-balogh:
> In overall it looks good but we need to review the logic again as I don't see a scenario where a user executes destroy cluster right after executing a test case (analysis has to happen and in case the test pass, the cluster can remain right?)

@RazTamir , let's rebase this PR and merge it to have at least some functionality to delete the cluster from ocs-ci as soon as possible. We can think of how to move the destroy outside of the TC, but as we have already deploy defined as TC I would merge this one and move redesign in different PR.  I see it's part of one topic in the agenda in next automation mtg.

clacroix12:
merge conflict has been addressed, this is ready for final review and merge.

run.py
```diff
@@ -32,7 +32,8 @@
         [--log-level <LEVEL>]
         [--cluster-name <NAME>]
         [--no-email]
-  run.py --cleanup=NAME [--osp-cred <FILE>]
+        [--no-destroy]
```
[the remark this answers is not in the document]
vasukulkarni: I think the [--no-destroy] is useful only in the first docopt option, the second cli only takes [-destroy] option + optional log.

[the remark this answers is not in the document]
clacroix12: Vasu is right, the second usage pattern is strictly for destroying a specified cluster. I didn't think it made sense to include the `--no-destroy` option for that usage.

clacroix12:
@petr-balogh i rebased on latest master and force pushed. Please verify that things look okay now.

PR #33, opened 2019-05-09
Configurable api client

shylesh:
This PR has 2 commits. 
1. Which handles the api-client configurations at run time
2. A simple test for api-client infrastructure

### DIFF
--- a/oc/openshift_ops.py
+++ b/oc/openshift_ops.py
@@ -6,6 +6,7 @@ from openshift.dynamic import DynamicClient, exceptions
 
 from ocs.exceptions import CommandFailed
 from utility.utils import run_cmd
+import ocs.defaults as default
 
 log = logging.getLogger(__name__)
 
@@ -27,14 +28,14 @@ class OCP(object):
         self.v1_projects = dyn_client.resources.get(
             api_version='project.openshift.io/v1', kind='Project'
         )
-        self.v1_pods = dyn_client.resources.get(
-            api_version='v1', kind='Pod'
+        self.pods = dyn_client.resources.get(
+            api_version=default.API_VERSION, kind='Pod'
         )
-        self.v1_deployments = dyn_client.resources.get(
-            api_version='v1', kind='Deployment'
+        self.deployments = dyn_client.resources.get(
+            api_version=default.API_VERSION, kind='Deployment'
         )
-        self.v1_services = dyn_client.resources.get(
-            api_version='v1', kind='Service'
+        self.services = dyn_client.resources.get(
+            api_version=default.API_VERSION, kind='Service'
         )
 
     @staticmethod
@@ -74,7 +75,7 @@ class OCP(object):
             list: of pods names, if no namespace provided then this function
                 returns all pods across openshift cluster.
         """
-        resource = self.v1_pods
+        resource = self.pods
 
         try:
             pod_data = resource.get(**kw)
@@ -103,7 +104,7 @@ class OCP(object):
             dict: All the openshift labels on a given pod
         """
 
-        resource = self.v1_pods.status
+        resource = self.pods.status
 
         try:
             pod_meta = resource.get(
--- a/oc/openshift_ops.py
+++ b/oc/openshift_ops.py
@@ -30,6 +30,38 @@ class OCP(object):
         self.v1_pods = dyn_client.resources.get(
             api_version='v1', kind='Pod'
         )
+        self.v1_deployments = dyn_client.resources.get(
+            api_version='v1', kind='Deployment'
+        )
+        self.v1_services = dyn_client.resources.get(
+            api_version='v1', kind='Service'
+        )
+
+    @staticmethod
+    def call_api(method, **kw):
+        """
+        This function makes generic REST calls
+
+        Args:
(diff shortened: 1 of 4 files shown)


ocs/api_client.py
```diff
@@ -0,0 +1,255 @@
...
+    This implements all functionalities like create, patch, delete using
+    oc commands.
+    """
+
+    @property
+    def name(self):
+        return "OCCLIClient"
+
+
+class KubClient(APIClientBase):
```
vasukulkarni: why not KubeClient?

ocs/api_client.py
```diff
@@ -0,0 +1,255 @@
...
+    """
+    All activities using upstream kubernetes python client
+    """
+
+    @property
+    def name(self):
+        return "KubClient"
```
vasukulkarni: s/KubClient/KubeClient

vasukulkarni:
Other than minor nits, LGTM,  Thanks for abstracting out functions into Kube vs Openshift which will make it easy to run on both environments.

ocs/api_client.py
shylesh: Done

ocs/api_client.py
shylesh: Done

ocs/api_client.py
```diff
@@ -0,0 +1,255 @@
...
+
+            Note: Returns could be tricky because it varies from client to
+            client. If its oc-cli then extra work needs to be done in the
+            specific implementation.
+        """
+
+        raise NotImplementedError("get_labels method is not implemented")
+
+
+# All openshift REST client specific imports here
```
petr-balogh: I would move all imports at top of module and order them in pep8 way.

ocs/api_client.py
```diff
@@ -0,0 +1,255 @@
...
+
+        resource = self.rest_client.v1_services
+        kw.update({'resource': resource})
+        return self.api_create(**kw)
+
+    def api_get(self, **kw):
+        return self.rest_client.call_api("GET", **kw)
+
+    def api_post(self, **kw):
+        pass
```
petr-balogh: super(). api_post(**kw) - this way we will raise NotImplementedError as you have defined in abstract method.  The same below.

ocs/api_client.py
```diff
@@ -0,0 +1,255 @@
...
+class OCRESTClient(APIClientBase):
+    """ All activities using openshift REST client"""
+
+    def __init__(self):
+        """TODO: get REST client instance from ctx which is shared globally"""
+        self.rest_client = openshift_ops.OCP()
+
+    @property
+    def name(self):
+        return "OCRESTClient"
```
petr-balogh: what about 
```python
return self.__class__.__name__
```

petr-balogh:
And please once WIP is over, remove it from the name. Thanks

oc/openshift_ops.py
```diff
@@ -25,6 +25,38 @@ def __init__(self):
         self.v1_pods = dyn_client.resources.get(
             api_version='v1', kind='Pod'
         )
+        self.v1_deployments = dyn_client.resources.get(
+            api_version='v1', kind='Deployment'
+        )
+        self.v1_services = dyn_client.resources.get(
```
RazTamir: Can we remove the v1 from the variable name?
The api version is v1 but the name shouldn't mention that

+ can we maintain the 'v1' in the defaults so as soon as the api version with change, we will be able to change only 1 parameter?

shylesh:
> Does `test_api_client.py` require a running cluster to work properly or would this be better suited as a unit test? Right now it's included in the same module as our product tests and I don't think it belongs there as it's testing part of our framework.

This test assumes that openshift cluster is up and running and creates an example service although it has nothing specific to rook cluster. However this is exercising the api-client interface which we intend to use for our general activities which involves oc create, patch, post etc. 

I am not sure whether we can keep it here or move it to some other place, do you feel it should be kept seperate ?

clacroix12:
> I am not sure whether we can keep it here or move it to some other place, do you feel it should be kept seperate ?

I'm more just trying to understand the nature of this test, and when we plan to execute it. It doesn't look like it's been added to any of our suites and seems like it would be better as a unit test if anything.

ocs/api_client.py
shylesh: Done

ocs/api_client.py
shylesh: Done

ocs/api_client.py
shylesh: Actually I was not sure whether config will have same parameter value as class name hence kept that string. However with this change user has to use same value as class name.

Done

oc/openshift_ops.py
shylesh: Done.

shylesh:
Looks like I have merge conflicts. Once if you guys agree on review related changes I will rebase and resolve the conflict.

petr-balogh:
@shylesh go ahead with rebase and solve the conflict.

vasukulkarni:
@shylesh please rebase

clacroix12:
LGTM

PR #35, opened 2019-05-10
filesystem creation for ceph

shreekarSS:
Added template for filesystem, updated rook resource creation for filesystem.yaml

### DIFF
--- a/templates/ocs-deployment/filesystem.yaml
+++ b/templates/ocs-deployment/filesystem.yaml
@@ -1,0 +1,54 @@
+#################################################################################################################
+# Create a filesystem with settings with replication enabled for a production environment.
+# A minimum of 3 OSDs on different nodes are required in this example.
+#  kubectl create -f filesystem.yaml
+#################################################################################################################
+
+apiVersion: ceph.rook.io/v1
+kind: CephFilesystem
+metadata:
+  name: myfs
+  namespace: {{ cluster_namespace | default('openshift-storage') }}
+spec:
+  # The metadata pool spec. Must use replication.
+  metadataPool:
+    replicated:
+      size: 3
+  # The list of data pool specs. Can use replication or erasure coding.
+  dataPools:
+    - failureDomain: host
+      replicated:
+        size: 3
+  # The metadata service (mds) configuration
+  metadataServer:
+    # The number of active MDS instances
+    activeCount: 1
+    # Whether each active MDS instance will have an active standby with a warm metadata cache for faster failover.
+    # If false, standbys will be available, but will not have a warm cache.
+    activeStandby: true
+    # The affinity rules to apply to the mds deployment
+    placement:
+    #  nodeAffinity:
+    #    requiredDuringSchedulingIgnoredDuringExecution:
+    #      nodeSelectorTerms:
+    #      - matchExpressions:
+    #        - key: role
+    #          operator: In
+    #          values:
+    #          - mds-node
+    #  tolerations:
+    #  - key: mds-node
+    #    operator: Exists
+    #  podAffinity:
+    #  podAntiAffinity:
+    # A key/value list of annotations
+    annotations:
+    #  key: value
+    resources:
+    # The requests and limits set here, allow the filesystem MDS Pod(s) to use half of one CPU core and 1 gigabyte of memory
+    #  limits:
+    #    cpu: "500m"
+    #    memory: "1024Mi"
+    #  requests:
+    #    cpu: "500m"
+    #    memory: "1024Mi"
--- a/tests/test_ocs_basic_install.py
+++ b/tests/test_ocs_basic_install.py
@@ -160,7 +160,7 @@ def run(**kwargs):
     create_oc_resource(
         "prometheus-rules.yaml", rook_data, cluster_path, _templating
     )
-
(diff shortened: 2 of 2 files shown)


tests/test_ocs_basic_install.py
```diff
@@ -160,7 +160,7 @@ def run(**kwargs):
     create_oc_resource(
         "prometheus-rules.yaml", rook_data, cluster_path, _templating
     )
-
+    create_oc_resource('filesystem.yaml', rook_data, cluster_path, _templating)
```
vasukulkarni: Lets move this in Rook Class as `def create_filesystem()`,  I think once I see the admin guide, we can revisit how to handle deployment with required services.

vasukulkarni:
@shreekarSS are you working on this one along with a test?

shylesh:
Approximately it should look like , rookcluster.create_fs(param1, param2 ...). 
create_fs function should consider creating the fs resource yaml by putting params into template and invoke creation through create call to openshift. Also this function should do verification of fs creation in the cluster.

roughly:

class RookCluster():
    def create_fs(param1, param2...):
          service_data = "
          api_ver: <v>
          expand yaml by replacing params
          "
          api_client.create(body=body, namespace=self.namespace)

RazTamir:
Why rook cluster and not OCS?

shylesh:
> Why rook cluster and not OCS?

Because a ceph filesystem OR RBD image closely tied with a ceph cluster.   It would be inherently meaningful to access something like cluster.fs OR cluster.RBD etc instead of treating them as independent units.  In case if you create a OCS class I feel ceph cluster will be a composition of OCS class because OCS contains a ceph cluster along with other non-ceph entities  and a ceph cluster contains filesystem OR RBD.

My 2 cents.

templates/ocs-deployment/filesystem.yaml
```diff
@@ -0,0 +1,54 @@
+#################################################################################################################
+# Create a filesystem with settings with replication enabled for a production environment.
+# A minimum of 3 OSDs on different nodes are required in this example.
+#  kubectl create -f filesystem.yaml
+#################################################################################################################
```
mbukatov: If you have used [this file](https://github.com/rook/rook/blob/master/cluster/examples/kubernetes/ceph/filesystem.yaml) as a base here, please make it clear in the comment.

Also if that is the case, I would even suggest to add the file in it's original form in one commit, and then changing it for the purposes of ocs qe work.

vasukulkarni:
> Why rook cluster and not OCS?

the current class is named Rook :)

templates/ocs-deployment/filesystem.yaml
```diff
@@ -0,0 +1,54 @@
...
+      size: 3
+  # The list of data pool specs. Can use replication or erasure coding.
+  dataPools:
+    - failureDomain: host
+      replicated:
+        size: 3
+  # The metadata service (mds) configuration
+  metadataServer:
+    # The number of active MDS instances
+    activeCount: 1
```
ramkiperiy: Please make this as param, where default 1 MDS count, or the user can specify the count.

templates/ocs-deployment/filesystem.yaml
```diff
@@ -0,0 +1,54 @@
...
+spec:
+  # The metadata pool spec. Must use replication.
+  metadataPool:
+    replicated:
+      size: 3
+  # The list of data pool specs. Can use replication or erasure coding.
+  dataPools:
+    - failureDomain: host
+      replicated:
+        size: 3
```
ramkiperiy: change this as a param "replicasize"

shreekarSS: yes Ram, I have made func create_fs() with all  configurable params needed

templates/ocs-deployment/filesystem.yaml
```diff
@@ -0,0 +1,54 @@
+#################################################################################################################
+# Create a filesystem with settings with replication enabled for a production environment.
+# A minimum of 3 OSDs on different nodes are required in this example.
+#  kubectl create -f filesystem.yaml
+#################################################################################################################
+
+apiVersion: ceph.rook.io/v1
+kind: CephFilesystem
+metadata:
+  name: myfs
```
petr-balogh: Shouldn't be the name also parametrized?

templates/ocs-deployment/filesystem.yaml
```diff
@@ -0,0 +1,54 @@
...
+  namespace: {{ cluster_namespace | default('openshift-storage') }}
+spec:
+  # The metadata pool spec. Must use replication.
+  metadataPool:
+    replicated:
+      size: 3
```
petr-balogh: Will there be the case when we want to have different replica size? If so, make it parametrized with default value 3.

templates/ocs-deployment/filesystem.yaml
petr-balogh: +1

templates/ocs-deployment/filesystem.yaml
petr-balogh: +1 on parameterized here as well

RazTamir:
@shreekarSS , is this patch still needed?

PR #42, opened 2019-05-14
adds a basic pytest structure for running ocs tests

andrewschoen:
This adds a ``conftest.py`` file that will inspect the test path at collection time and give appropriate markers if the test file lives in a directory with the name of a ceph daemon type. For example, all tests that live in a file at ``tests/mon/test_mons.py`` would get a ``mons`` marker. This is really helpful when wanting to run daemon specific tests.

This also includes a simple tox.ini file that can be used to run pytest against the ``ocs-ci/tests`` folder.


tests/tox.ini
```diff
@@ -0,0 +1,11 @@
+[tox]
+envlist = py27,py35,py36,py37
```
vasukulkarni: @andrewschoen we have moved out of py27 and anyway its just couple of months for EOL for 27, Can we remove py27?

andrewschoen: Sure, I can do that. I ran this with ``py37`` locally anyway.

tests/tox.ini
```diff
@@ -0,0 +1,11 @@
...
+envlist = py35,py36,py37
+skip_missing_interpreters = true
+skipsdist = true
+
+[testenv]
+deps=
+  -rrequirements.txt
+  pytest
+commands=
+    pytest -v {posargs:.}
```
clacroix12: I checked [the docs](https://tox.readthedocs.io/en/latest/example/general.html) but wasn't able to find anything on it, would you elaborate what the `.` is doing in `{posargs:.}`?

tests/tox.ini
```diff
@@ -0,0 +1,11 @@
+[tox]
+envlist = py35,py36,py37
```
clacroix12: I don't think we can use py35 here since we utilize f-strings which were introduced in 3.6.

tests/tox.ini
andrewschoen: It's just running pytest in the current directory, totally not necessary I suppose. I find ``{posargs}`` useful though, if by some reason we ever need to point at a different test directory interactively.

Would you prefer if I remove that default value there? If it improves readability I'm ok with that.

tests/tox.ini
andrewschoen: I can remove it. Honestly the python version here isn't all that terribly important. If we only want ``ocs-ci`` to support ``py36`` or ``py37`` we can pick one and then name the tox environment something more descriptive than the python version. That's probably be a good idea anyway.

tests/tox.ini
clacroix12: Ah I see now, that makes sense. For some reason I was thinking it was something for tox but you're just passing `.` through to pytest. That makes sense as the default functionality to me.

tests/tox.ini
clacroix12: Yeah that sounds like a better approach. If we have multiple envs in `envlist` won't we end up executing tests once per env? I don't think that's what we want to happen for our ocs tests.

andrewschoen: I've renamed the env ``ocs-tests`` and defaulted to ``py37``. I'm open to suggestions on better naming for sure, but I think that works. You can specify which env you want to run with the ``-e`` flag so I don't think that'd be a problem.

tests/tox.ini
```diff
@@ -0,0 +1,11 @@
+[tox]
+envlist = {ocs-tests}
+basepython = py37
+skipsdist = true
+
+[testenv]
+deps=
+  -rrequirements.txt
```
vavuthu: is that typo ( -r requirements.txt )

tests/tox.ini
```diff
@@ -0,0 +1,11 @@
...
+  pytest
```
petr-balogh: I see pytest is defined just here in tox requirements. But if we are going support pytest in OCS-CI, we will be implementing some plugins and so on I think that it should be requirements defined for whole ocs-ci project in setup.py.

tests/tox.ini
petr-balogh: it's not typo @vavuthu , if you will do pip install -rrequirements.txt it work like a char as just -r is the first argument and then the path to requirement file following. It's more readable when there is space I agree, but doesn't matter.

petr-balogh: Do we really need to separate and have other tox.ini file in tests directory? Cannot we just have different env defined in tox.ini file in root directory of the project?

tests/requirements.txt
```diff
@@ -0,0 +1 @@
+-e ../
```
petr-balogh: What about ppl who would like to run pytest. Currently now the flow is like create venv for ocs-ci like 
```bash
python3 -m venv .venv
source .venv/bin/activate
pip install -r requirements.txt
python run.py ......
```

If I would like to run pytest now, I don't have defined the pytest requirement anywhere just for tox.
So if we will stick with this project, I would have one place for all the requirements in setup.py as everything will be binded so closely with whole framework I think.  Or are you tending to have separate requirement file for tests requirement, so once ppl will need to have some package to be installed for their test that it should be defined in this file and not in the setup.py?  If so I still think that at least pytest and it's plugins should be in the setup.py.

Also as the user if I would like to install this requirements to the same venv as I have for ocs-ci I cannot do something like this from root directory of the project 
```bash
pip install -r tests/requirements.txt
Directory '../' is not installable. File 'setup.py' not found.
```
I have to `cd tests` and run `pip install -r requirements.txt`.

So please document a bit how this should be used for the users.

tests/tox.ini
```diff
@@ -0,0 +1,11 @@
+[tox]
```
mbukatov: What is the purpose of this file here?

tests/tox.ini
mbukatov: This is also useful if one needs to pass any other pytest option to pytest run executed via tox, eg. like `tox -e py36 -- --pdb --runxfail tests/test_foo.py`.

That said, I don't see any value in using tox here to run ocsqe test cases.

tests/requirements.txt
mbukatov: +1 I don't see any value in using tox here to run ocsqe test cases.

tests/tox.ini
andrewschoen: Good point, I'll make that change.

andrewschoen:
I do like to tox to run these tests, but I'm not sold on it if the rest of the team disagrees. Even if we keep the ``tests/tox.ini`` file those who chose to can run pytest directly. I've used it on past projects to great success, especially the generative envlist feature. It can be used to create a dynamic testing matrix very easily. https://tox.readthedocs.io/en/latest/config.html#generating-environments-conditional-settings

And yes, this definitely needs some documentation. Fully agree there.

tests/requirements.txt
andrewschoen: I've added ``pytest`` to setup.py and added a comment below about tox usage.

tests/tox.ini
andrewschoen: We could have a different env defined in the tox.ini file in the root. With as simple as it is now, that might be a good option. I've found in the past though that multiple, less complicated tox.ini files are better than trying to manage one very complicated tox.ini that does everything.

RazTamir:
@clacroix12 @petr-balogh @vasukulkarni is this PR still relevant?

clacroix12:
@RazTamir i don't think so. We are handling marks differently with petr's PR and it seems like we have decided to execute pytest directly instead of through tox.

tests/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+    daemon in it, tag the test with the appropriate
+    marker. This allows for us to filter and test
+    by specific daemon type.
+
+    For example, all tests that live in a file at
+    tests/mon/test_mons.py would be tagged with
+    a 'mons' marker.
+    """
+    for item in items:
+        test_path = item.location[0]
```
petr-balogh: Hi @andrewschoen, what item.location[0] contains? First directory in which the test file belongs to?  We discussed to have structure starting with teams folders in QE. And we merged PR #61 which implements pytest for us and the basic config we had in old runner to pytest. So if this structure is needed from engineering can this been discussed on our regular meeting? Thanks

clacroix12:
@petr-balogh @RazTamir can we close this PR given the recent changes we have made with regards to pytest?

andrewschoen:
> @petr-balogh @RazTamir can we close this PR given the recent changes we have made with regards to pytest?

I'll go ahead and close this. I can reopen it if we decide there's things here we might still need.

PR #43, opened 2019-05-15
Implementing issue#48 and adding 2 new test cases

RazTamir:
This is 2 new tests - please review

### DIFF
--- a/ocs/defaults.py
+++ b/ocs/defaults.py
@@ -8,3 +8,6 @@ AWS_REGION = 'us-east-2'
 ROOK_CLUSTER_NAMESPACE = 'openshift-storage'
 KUBECONFIG_LOCATION = 'auth/kubeconfig'  # relative from cluster_dir
 CLUSTER_NAME = f"ocs-ci-cluster-{getuser()}"
+VOLUME_DELETED = 'persistentvolume "{volume_name}" deleted'
+CEPHFS_DELETED = '"{cephfs_name}" deleted'
+NOT_FOUND = 'NotFound'
--- a/ocs/defaults.py
+++ b/ocs/defaults.py
@@ -8,6 +8,3 @@ AWS_REGION = 'us-east-2'
 ROOK_CLUSTER_NAMESPACE = 'openshift-storage'
 KUBECONFIG_LOCATION = 'auth/kubeconfig'  # relative from cluster_dir
 CLUSTER_NAME = f"ocs-ci-cluster-{getuser()}"
-VOLUME_DELETED = 'persistentvolume "{volume_name}" deleted'
-CEPHFS_DELETED = '"{cephfs_name}" deleted'
-NOT_FOUND = 'NotFound'
--- a/ocs/ocp.py
+++ b/ocs/ocp.py
@@ -1,0 +1,156 @@
+"""
+General OCP object
+"""
+import os
+import logging
+import yaml
+
+from munch import munchify
+from utility.utils import TimeoutSampler
+from utility.utils import run_cmd
+
+log = logging.getLogger(__name__)
+
+
+class OCP(object):
+    """
+    A basic OCP object to run basic 'oc' commands
+    """
+
+    def __init__(self, api_version='v1', kind='Service', namespace=None):
+        """
+        Initializer function
+
+        Args:
+            api_version (str): TBD
+            kind (str): TBD
+            namespace (str): The name of the namespace to use
+        """
+        self._api_version = api_version
+        self._kind = kind
+        self._namespace = namespace
+
+    @property
+    def api_version(self):
+        return self._api_version
+
+    @property
+    def kind(self):
+        return self._kind
+
+    @property
+    def namespace(self):
+        return self._namespace
+
+    def exec_oc_cmd(self, command):
+        """
+        Executing 'oc' command
+
+        Args:
+            command (str): The command to execute (e.g. create -f file.yaml)
+                without the initial 'oc' at the beginning
+
+        Returns:
+            Munch Obj: this object represents a returned yaml file
+        """
+        oc_cmd = "oc "
+        kubeconfig = os.getenv('KUBECONFIG')
+        if self.namespace:
+            oc_cmd += f"-n {self.namespace} "
+
+        if kubeconfig:
+            oc_cmd += f"--kubeconfig {kubeconfig} "
+
+        oc_cmd += command
+        out = run_cmd(cmd=oc_cmd)
+        return munchify(yaml.safe_load(out))
+
+    def get(self, resource_name='', out_yaml_format=True):
+        """
+        Get command - 'oc get <resource>'
+
+        Args:
+            resource_name (str): The resource name to fetch
+
+        Example:
+            get('my-pv1')
+
+        Returns:
+            Munch Obj: this object represents a returned yaml file
+        """
+        command = f"get {self.kind} {resource_name}"
+        if out_yaml_format:
+            command += " -o yaml"
+        return self.exec_oc_cmd(command)
+
+    def create(self, yaml_file, out_yaml_format=True):
+        """
+        Creates a new resource
+
+        Args:
+            yaml_file (str): Path to a yaml file to use in 'oc create -f
+                file.yaml
+            out_yaml_format (bool): Determines if the output should be
+                formatted to a yaml like string
+
+        Returns:
+            Munch Obj: this object represents a returned yaml file
+        """
+        command = f"create -f {yaml_file}"
+        if out_yaml_format:
+            command += " -o yaml"
+
+        return self.exec_oc_cmd(command)
+
+    def delete(self, yaml_file, wait=True):
+        """
+        Deletes a resource
+
+        Args:
+            yaml_file (str): Path to a yaml file to use in 'oc delete -f
+                file.yaml
+            wait (bool): Determines if the delete command should wait to
+                completion
+
+        Returns:
+            Munch Obj: this object represents a returned yaml file
+        """
+        command = f"delete -f {yaml_file}"
+        if wait:
+            command += " --wait=True"
+        return self.exec_oc_cmd(command)
+
+    def apply(self, yaml_file):
+        """
+        Applies configuration changes to a resource
+
+        Args:
+            yaml_file (str): Path to a yaml file to use in 'oc apply -f
+                file.yaml
+
+        Returns:
+            Munch Obj: this object represents a returned yaml file
+        """
+        command = f"apply -f {yaml_file}"
+        return self.exec_oc_cmd(command)
+
+    def wait_for_resource_status(
+        self, resource_name, condition, timeout=30, sleep=3
+    ):
+        """
+        Wait for a resource to reach to a desired condition
+
+        Args:
+            resource_name (str): The name of the resource to wait
+                for (e.g.my-pv1)
+            condition (str): The desired state the resource should be at
+                This is referring to: status.phase presented in the resource
+                yaml file
+                (e.g. status.phase == Available)
+        """
+        for sample in TimeoutSampler(
+            timeout, sleep, self.get, resource_name
+        ):
+            if sample.status.phase == condition:
+                return True
+        return False
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,8 @@ setup(
         'kubernetes<9.0.0,>=8.0.0',  # to satisfy openshift package requirement
         'openshift',
         'boto3',
+        'munch',
+        'ipdb'
     ],
     zip_safe=True,
     include_package_data=True,
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,6 @@ setup(
         'openshift',
         'boto3',
         'munch',
-        'ipdb'
     ],
     zip_safe=True,
     include_package_data=True,
--- a/templates/ocs-deployment/PersistentVolume.yaml
+++ b/templates/ocs-deployment/PersistentVolume.yaml
@@ -1,0 +1,16 @@
+---
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: {{ pv_name | default('ocsci-pv') }}
+  namespace: {{ cluster_namespace | default('openshift-storage') }}
+spec:
+  capacity:
+    storage: {{ pv_size | default('5Gi') }}
+  accessModes:
+    - {{ access_mode | default('ReadWriteOnce') }}
+  persistentVolumeReclaimPolicy: Recycle
+  storageClassName: {{ storageclass_namespace | default('slow') }}
+  nfs:
+    path: /tmp
+    server: 172.17.0.2
--- a/tests/test_add_mds_to_cluster.py
+++ b/tests/test_add_mds_to_cluster.py
@@ -10,7 +10,7 @@ from ocs import ocp
 from munch import munchify
 from ocs import exceptions
 from ocsci.enums import TestStatus
-from utility import utils
+from utility import utils, templating
 
 log = logging.getLogger(__name__)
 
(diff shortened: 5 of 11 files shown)


tests/test_add_mds_to_cluster.py
```diff
@@ -0,0 +1,144 @@
...
+
+def run(**kwargs):
+    """
+    A simple function to exercise a resource creation through api-client
+    """
+    fs_name = 'my-cephfs'
+    active_count = '1'
+    new_active_count = '2'
+    ceph_fs_resource = dyn_client.resources.get(
+        api_version='v1', kind='CephFilesystem'
```
shylesh: @RazTamir 
You are using api_version='v1' , Don't we have to use ceph.rook.io/v1 ? OR it doesn't matter at all ?

I am still not familiar with api namespaces and rook's interaction in kubernets, just a hunch?Can you confirm it once.

RazTamir: It doesn't seem like a problem to use only 'v1'.
I don't know what is the difference

utility/utils.py
```diff
@@ -11,6 +11,7 @@
 import traceback
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
+from munch import munchify
```
fbalak: `munch` is missing in requirements or setup files.

RazTamir: Will take care of everything in next patch

tests/test_add_mds_to_cluster.py
```diff
@@ -0,0 +1,144 @@
+"""
+A test for creating a CephFS
+"""
+import logging
+from pdb import set_trace
```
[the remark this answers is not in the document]
RazTamir: Your'e right.
I should have commented that this is for demo reasons

ocs/ocp.py
```diff
@@ -0,0 +1,106 @@
...
+        oc_cmd = "oc "
+        kubeconfig = os.getenv('KUBECONFIG')
+        if self.namespace:
+            oc_cmd += f"-n {self.namespace} "
+
+        if kubeconfig:
+            oc_cmd += f"--kubeconfig {kubeconfig} "
+
+        oc_cmd += command
+        out = run_cmd(cmd=oc_cmd)
```
ebenahar: in case of a failure, I believe that the output won't digestible munchify(yaml.safe_load(out))

ocs/ocp.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def apply(self, yaml_file):
+        """
+        TBD
+        """
+        command = f"apply -f {yaml_file}"
+        return self.exec_oc_cmd(command)
+
+    def wait_for_resource_status(
+        self, resource_kind, resource_name, condition, timeout=30, sleep=3
```
ebenahar: save as constants

ebenahar:
Please also change the commit message

ocs/ocp.py
RazTamir: in a failure, an exception will raise from run_cmd so it won't reach munchify

ebenahar: Right :)

tests/test_add_mds_to_cluster.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def run(**kwargs):
+    """
+    A simple function to exercise a resource creation through api-client
+    """
+    fs_data = {}
+    fs_name = 'my-cephfs1'
+    fs_data['fs_name'] = fs_name
+    fs_data['active_count'] = '1'
```
ebenahar: maybe we can create a default value for that

tests/test_add_mds_to_cluster.py
```diff
@@ -0,0 +1,86 @@
...
+    new_active_count = '3'
```
ebenahar: here too

tests/test_pv_creation.py
```diff
@@ -0,0 +1,77 @@
...
+
+PV_YAML = os.path.join("templates/ocs-deployment", "PersistentVolume.yaml")
+TEMP_YAML_FILE = 'test.yaml'
+
+RESOURCE = ocp.OCP(
+    kind='PersistentVolume', namespace=defaults.ROOK_CLUSTER_NAMESPACE
+)
+
+
+def create_pv(**kwargs):
```
ebenahar: these basic functions (like create/delete pv) deserve to be moved to a common place

ebenahar:
@RazTamir this PR is pretty big, some added functionalities, like run oc cmd, are very crucial for many of us to start adding cases. Thinking it would be better to separate to smaller patches

utility/utils.py
```diff
@@ -754,3 +756,129 @@ def download_openshift_installer(version=defaults.INSTALLER_VERSION):
         run_cmd(f"tar xzvf {tarball}")
 
     return installer_filename
+
+
+def run_oc_cmd(cmd):
```
ebenahar: redundant?

tests/test_pv_creation.py
RazTamir: of course but I think that for now, until we won't move to pytest, I prefer to keep things simpler for the test itself.
Could you please open a new issue for that and add me as an assignee?

Thanks

utility/utils.py
RazTamir: nope,
run_cmd is for running any command
run_oc_cmd is for running oc and taking care of munchify etc.

RazTamir:
@ebenahar about the size of the patch,
If this PR will be merged soon it will be no issue (:
I have to keep it together so I'll be able to test the new functionalities

utility/utils.py
ebenahar: @RazTamir I meant if it's a duplicate with exec_oc_cmd() from ocp.py

RazTamir: nice catch!
removed from utils.py

tests/test_pv_creation.py
ebenahar: @RazTamir https://github.com/red-hat-storage/ocs-ci/issues/50

ocs/ocp.py
```diff
@@ -0,0 +1,156 @@
...
+            oc_cmd += f"-n {self.namespace} "
+
+        if kubeconfig:
+            oc_cmd += f"--kubeconfig {kubeconfig} "
+
+        oc_cmd += command
+        out = run_cmd(cmd=oc_cmd)
+        return munchify(yaml.safe_load(out))
+
+    def get(self, resource_name='', out_yaml_format=True):
```
petr-balogh: what about to user to pass additional extra parameters to this oc get commonad (not only in `oc get` but others as well). 

I suggest like additional parameter `parameters` which can be list of the optional parameters as we probably don't want to enable all functionality here. But to give the user to option use more that is defined here? (Of course on user responsibility, so if they will add parameter --output json they will break the behavior here)

so if user will add parameters=['--request-timeout', 60] for example they can change the behavior of the command regarding the [documentation](https://www.mankier.com/1/oc-get#--request-timeout)

ocs/defaults.py
```diff
@@ -8,3 +8,6 @@
 ROOK_CLUSTER_NAMESPACE = 'openshift-storage'
 KUBECONFIG_LOCATION = 'auth/kubeconfig'  # relative from cluster_dir
 CLUSTER_NAME = f"ocs-ci-cluster-{getuser()}"
+VOLUME_DELETED = 'persistentvolume "{volume_name}" deleted'
```
petr-balogh: I feel like this deserve better place than defaults, same below.

utility/utils.py
```diff
@@ -754,3 +755,124 @@ def download_openshift_installer(version=defaults.INSTALLER_VERSION):
         run_cmd(f"tar xzvf {tarball}")
 
     return installer_filename
+
+
+def generate_yaml_from_template(file_, **kwargs):
```
petr-balogh: I would keep all templating utils in this [templating module](https://github.com/red-hat-storage/ocs-ci/blob/master/utility/templating.py) or class itself if possible as static method for example.

For example we can define custom filters there which are part of env defined there like to_nice_yaml and so on.

utility/utils.py
```diff
@@ -754,3 +755,124 @@ def download_openshift_installer(version=defaults.INSTALLER_VERSION):
...
+    """
+    Delete file_name
+
+    Args:
+        file_name (str): Path to the file you want to delete
+    """
+    os.remove(file_name)
+
+
+class TimeoutExpiredError(Exception):
```
petr-balogh: Consider move to [exceptions](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs/exceptions.py)

ocs/ocp.py
RazTamir: I agree but because this is an initial PR I will leave the advanced implementation to a different PR.
We can always extend it

ocs/defaults.py
RazTamir: I agree,
Any proposed file name?

utility/utils.py
RazTamir: wasn't aware of this file.
Will move it there in case this is not a dup

ocs/ocp.py
```diff
@@ -0,0 +1,156 @@
...
+        Executing 'oc' command
+
+        Args:
+            command (str): The command to execute (e.g. create -f file.yaml)
+                without the initial 'oc' at the beginning
+
+        Returns:
+            Munch Obj: this object represents a returned yaml file
+        """
+        oc_cmd = "oc "
```
dahorak: Just FYI: this should be later modified to use `get_openshift_client()` function from PR https://github.com/red-hat-storage/ocs-ci/pull/49 (but I'm not sure, which PR will be merged earlier).

utility/utils.py
RazTamir: Done

ocs/ocp.py
RazTamir: +1.
If you can open an issue for this it will be great so we won't forget to fix it

ocs/defaults.py
clacroix12: It seems like something that can just be handled within the script? You're really just building out a string with the volume name, no need to have a central place for something like this.

vasukulkarni:
LGTM since this is a *first cut* for OCP class and sample tests. Thanks for posting the run results.

shylesh:
@RazTamir ,
I have a concern here.
Will the users consuming the ocp methods the way your are using in test programs ?
If Yes, then it will be tightly coupled system and might lead extensibility problems.

For ex: consider a simple test scenario where you want to
1. create a filesystem
2. start io
3. bringdown mds container while I/O is in progress 

Here step1 straight forward "oc create", step 2 little bit convoluted(How to get client pod info ?) but doable.
when it comes to step3 how do you know which is the mds for this filesystem ? you need a utility function here to figure out , atleast I dont see it in resource information. Basically for other scenarios like this where we need some ceph cluster data in conjunction, we will end up having lot of utility functions.

So I had few class designs where we will be creating "has a" heirarchy.
For ex: a cephfs is a kubernetes volume type (from kubernetes perspective), at the same time an fs should be associated with a ceph cluster, hence
```python
class CephFSVolume(KubernetesVolume):
    self.cluster = RookCluster() 
   
    def create()
    def delete()
    def update()

class RookCluster()
    self.mds = []
    self.mons = []
    api_client =  ctx['api_client']   ----> which would be pointing to OCP object
   def create_fs():
         yamldata= get_from_template()
         api_client.create(yamldata)

   def get_pod(role):


To restart mds:
cfs = CephFSVolume()
pod = cfs.cluster.get_pod(mds)
pod.restart()

Thoughts...

**Comment by RazTamir** [comment]
Hi @shylesh ,
The OCP class is the base class that all others should (this is my intention) inherit from and not from Object.
We will definitely need to expand the functionality we have but not in OCP obj as this is a class for mimicking running 'oc' commands and that's it.
If you want to create this design I think this is a good approach but your rook cluster will need to have the capabilities of running 'oc' commands so it should inherit from OCP class and not creating an instance inside:
```python
api_client =  ctx['api_client']
```
I'm not sure what capabilities KubernetesVolume provides so not sure why to inherit from it
My initial thought was to create an object for every kind we will use:
```python
class CephFilesystem(OCP)

class PersistentVolume(OCP)

class PersistentVolumeClaim(PersistentVolume)
```
and so on --> pure OOP will ensure our design is better maintained

Thoughts?

RazTamir:
BTW,
In your example you still not solving the issue to identify which pods is running the filesystem.
```python
pod = cfs.cluster.get_pod(mds)
```
How do you know which mds is the one who holds the FS?
So in any case, we will need to create all those functions

setup.py
```diff
@@ -24,6 +24,8 @@
         'kubernetes<9.0.0,>=8.0.0',  # to satisfy openshift package requirement
         'openshift',
         'boto3',
+        'munch',
+        'ipdb'
```
petr-balogh: I don't think we need to have ipdb as part of the installation. If user want to debug it needs to be installed by the user in the venv.

ocs/defaults.py
petr-balogh: agree with Coady that this can be just part of the module where are you using this

templates/ocs-deployment/PersistentVolume.yaml
```diff
@@ -0,0 +1,16 @@
+---
```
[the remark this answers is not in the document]
RazTamir: That's true but we still want to test also PV creation only right?

setup.py
RazTamir: Done

petr-balogh:
@fbalak can you please re-approve the change? So then I can merge. Thanks

tests/test_add_mds_to_cluster.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    A simple function to exercise a resource creation through api-client
+    """
+    fs_data = {}
+    fs_name = 'my-cephfs1'
+    fs_data['fs_name'] = fs_name
+    # fs_data['active_count'] = '1'
+    new_active_count = '3'
+    assert create_ceph_fs(**fs_data)
+    assert verify_fs_exist(fs_name)
```
nehaberry: @RazTamir would it be better to have a check for number of MDS pods (oc get pods) as well ? Also, Since we are going to increase the active count in next step, only pod count will change and hence re-verifying MDS pod counts again may be a good validation. Let me know if its already handled and I misinterpreted.

RazTamir: Great input Neha.
I'll take care of it in a new PR as this already merged

shylesh:
Reply Inline.

> Hi @shylesh ,
> The OCP class is the base class that all others should (this is my intention) inherit from and not from Object.

[Shylesh] This is a global context object, I mean after deployment we will decide what api client we use and instantiate there. Put this client object in a ctx which will be passed all along tests and its upto tests to decide what they want to do.

> We will definitely need to expand the functionality we have but not in OCP obj as this is a class for mimicking running 'oc' commands and that's it.

[Shylesh] Exactly that is where I am coming, as per your test script, you were directly accessing OCP functions and I had concern over tests directly accessing OCP class and dealing with functions like createfs, createpv, which is supposed to be abstracted and exposed through class objects.

> If you want to create this design I think this is a good approach but your rook cluster will need to have the capabilities of running 'oc' commands so it should inherit from OCP class and not creating an instance inside:
> 
> ```python
> api_client =  ctx['api_client']
> ```
> 
[Shylesh] As I said rook will use the api_client instance from global ctx.

> I'm not sure what capabilities KubernetesVolume provides so not sure why to inherit from it
> My initial thought was to create an object for every kind we will use:
> 

[Shylesh] KuberneteVolume is the broad category into which all cephfsvolume, rbdvolume, pvcs fall into which I found from kubernetes doc. This may define some common functionalities like create, delete etc of-course which will be overridden in most of the cases.

> ```python
> class CephFilesystem(OCP)

[Shylesh] I feel inheritance is an anti-pattern here in this case, CephFilesystem "IS NOT A" OCP instead it just leverages facilities of OCP . I feel this is a candidate for "composition" rather than "inheritance". Any cephfs operation should go through ceph cluster because ceph cluster object will have more context on what are the methods and attributes available for cluster related ops.

[Shylesh] Other inheritance like RookCluster(OCS) also doesn't seem to be right candidate, because a OCS has rook as one of the components but rook itself is not ocs. Please find another design doc I have shared with infra team.
> 
> class PersistentVolume(OCP)
> 
> class PersistentVolumeClaim(PersistentVolume)
> ```
> 
> and so on --> pure OOP will ensure our design is better maintained
> 
> Thoughts?

As you said this is not a final design but I just wanted to dump my thoughts here.

Thanks,
Shylesh

RazTamir:
+1 on all.
Thanks @shylesh

PR #49, opened 2019-05-16
WIP: Download openshift client

dahorak:
* handle download of Openshift client (`oc`)
* download both Openshift client and installer into `./bin` directory

### DIFF
--- a/utility/utils.py
+++ b/utility/utils.py
@@ -18,7 +18,9 @@ from jinja2 import Environment, FileSystemLoader, select_autoescape
 from reportportal_client import ReportPortalServiceAsync
 
 from ocs import defaults
-from ocs.exceptions import CommandFailed, UnsupportedOSType
+from ocs.exceptions import (
+    CommandFailed, UnsupportedOSType, TimeoutExpiredError,
+)
 from ocsci.enums import TestStatus
 from .aws import AWS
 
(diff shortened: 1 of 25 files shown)


utility/utils.py
```diff
@@ -750,7 +749,63 @@ def download_openshift_installer(version=defaults.INSTALLER_VERSION):
             f"https://mirror.openshift.com/pub/openshift-v4/clients/ocp/"
             f"{version}/openshift-install-{os_type}-{version}.tar.gz"
         )
+        # Prepare BIN_DIR
+        try:
+            os.mkdir(defaults.BIN_DIR)
+        except FileExistsError:
+            pass
```
petr-balogh: I would at least write debug message to log about that bin directory exist instead of pass WDYT?

utility/utils.py
```diff
@@ -750,7 +749,63 @@ def download_openshift_installer(version=defaults.INSTALLER_VERSION):
...
+    else:
+        log.info("Downloading openshift client")
+        if platform.system() == "Darwin":
+            os_type = "mac"
+        elif platform.system() == "Linux":
+            os_type = "linux"
+        else:
+            raise UnsupportedOSType
+        url = (
+            f"https://mirror.openshift.com/pub/openshift-v4/clients/ocp/"
```
petr-balogh: I see taht path is almost same as here: https://github.com/red-hat-storage/ocs-ci/pull/49/files#diff-ef93c603b643a8362ba61f8d9797bac2R749, so I would move this URL to some different place with format varaibles and would format it on this place and for installer as well with provided variables.

utility/utils.py
```diff
@@ -750,7 +749,63 @@ def download_openshift_installer(version=defaults.INSTALLER_VERSION):
...
+        str: path of the client binary
+
+    """
+    client_binary_path = os.path.join(defaults.BIN_DIR, 'oc')
+    if os.path.isfile(client_binary_path):
+        log.info("Client exists, skipping download")
+        # TODO: check client version
+    else:
+        log.info("Downloading openshift client")
+        if platform.system() == "Darwin":
```
petr-balogh: Identical code about getting system platform from https://github.com/red-hat-storage/ocs-ci/pull/49/files#diff-ef93c603b643a8362ba61f8d9797bac2R742 , please do function and reuse on both places in this case.

utility/utils.py
```diff
@@ -750,7 +749,63 @@ def download_openshift_installer(version=defaults.INSTALLER_VERSION):
...
+            os_type = "mac"
+        elif platform.system() == "Linux":
+            os_type = "linux"
+        else:
+            raise UnsupportedOSType
+        url = (
+            f"https://mirror.openshift.com/pub/openshift-v4/clients/ocp/"
+            f"{version}/openshift-client-{os_type}-{version}.tar.gz"
+        )
+        # Prepare BIN_DIR
```
petr-balogh: I already saw this code above, you know what to do ;)

utility/utils.py
```diff
@@ -750,7 +749,63 @@ def download_openshift_installer(version=defaults.INSTALLER_VERSION):
...
+    Args:
+        version (str): version of the client to download
+
+    Returns:
+        str: path of the client binary
+
+    """
+    client_binary_path = os.path.join(defaults.BIN_DIR, 'oc')
+    if os.path.isfile(client_binary_path):
+        log.info("Client exists, skipping download")
```
RazTamir: I would only print in case the client is not exist

utility/utils.py
```diff
@@ -713,31 +712,31 @@ def destroy_cluster(cluster_path):
...
+        str: path of the installer binary
 
     """
     installer_filename = "openshift-install"
-    tarball = f"{installer_filename}.tar.gz"
-    if os.path.isfile(installer_filename):
+    installer_binary_path = os.path.join(defaults.BIN_DIR, installer_filename)
+    if os.path.isfile(installer_binary_path):
         log.info("Installer exists, skipping download")
+        # TODO: check installer version
```
vasukulkarni: Yeah this seems tricky, probably better to overwrite to be *sure* we got the latest version.

mbukatov:
@dahorak Would it make sense to download both clients during (pre)setup and then update `PATH` so that the `bin` directory is there with priority? This way, one could just use `run_cmd("oc cluster-info")` instead of obtaining oc object and `run_cmd(f"{oc} cluster-info")`. It would also limit the merge conflicts and additional changes required.

That said, your approach is explicit and could be extended to a case when you have 2 different versions of oc in one test.

utility/utils.py
dahorak: Addressed in 0056ce6f3e59d0365c7ef238bd4b29cd3b1e36bc.

utility/utils.py
dahorak: Updated in 0056ce6f3e59d0365c7ef238bd4b29cd3b1e36bc.

utility/utils.py
dahorak: Updated in 0056ce6f3e59d0365c7ef238bd4b29cd3b1e36bc.

utility/utils.py
dahorak: Updated in 0056ce6f3e59d0365c7ef238bd4b29cd3b1e36bc.

utility/utils.py
dahorak: I've changed to print this message only at *debug* level, is it ok?

utility/utils.py
dahorak: Not sure, what would be the best approach here, would need it to another PR.

dahorak:
@mbukatov Addressed in 095192659b2a8b746295c7cbd0748208cd7de06b.

utility/utils.py
```diff
@@ -715,47 +714,136 @@ def destroy_cluster(cluster_path):
...
+        os.chdir(defaults.BIN_DIR)
+        url = get_openshift_mirror_url('openshift-client', version)
+        tarball = "openshift-client.tar.gz"
+        download_file(url, tarball)
+        run_cmd(f"tar xzvf {tarball}")
+        os.remove(tarball)
+        # return to the previous working directory
+        os.chdir(previous_dir)
+
+    add_path_to_env_path(defaults.BIN_DIR)
```
mbukatov: This should not be hidden there, but called at the same level where `get_openshift_client()` is called.

Also: we need to make sure `PATH` is configured properly for tests executed on already provisioned cluster (when ` tests/test_ocs_basic_install.py` is not executed). That said, since we are moving to pytest, we could just note this to be a known issue and solve it with pytest fixture properly after this pull request is merged.

dahorak: My intention with this approach was, that anybody who needs `oc` command will call the `get_openshift_client()` function, which will ensure that the client is available and the use either the `oc` command rely on the PATH configured by the function or will use the full path to `oc` command returned by the function.

dahorak:
Since PR https://github.com/red-hat-storage/ocs-ci/pull/61 is merged, I'll update and clean this PR and probably force push it.

dahorak:
Based on @mbukatov's suggestion, I'll close this PR and create new one with the changes based on the update from PR  https://github.com/red-hat-storage/ocs-ci/pull/61.

PR #52, opened 2019-05-16
Check for pull-secret file in ocsci-data as well and raise exception …

vasukulkarni:
Check for pull-secret file in ocsci-data as well and raise exception if we cannot
find in required dir's.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -46,8 +46,9 @@ cp ocs-ci.yaml.template ~/.ocs-ci.yaml
 In order to deploy a cluster to AWS with the Openshift Installer,
 you will need to download the pull secret for your account.
 Download this file from [openshift.com](https://cloud.openshift.com/clusters/install)
-and place in the `data` directory at the root level of the project.
-If there is no `data` directory, create one.
+and place in the `data` directory at the root level of the project
+If there is no `data` directory, create one or you can place the file in
+`$HOME/ocsci-data/`.
 The name of the file should be `pull-secret`.
 
 ## Tests
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -47,8 +47,9 @@ In order to deploy a cluster to AWS with the Openshift Installer,
 you will need to download the pull secret for your account.
 Download this file from [openshift.com](https://cloud.openshift.com/clusters/install)
 and place in the `data` directory at the root level of the project
-If there is no `data` directory, create one or you can place the file in
-`$HOME/ocsci-data/`.
+(If there is no `data` directory, create one) or you can place the file in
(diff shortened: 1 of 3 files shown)


docs/getting_started.md
```diff
@@ -46,8 +46,9 @@ cp ocs-ci.yaml.template ~/.ocs-ci.yaml
 In order to deploy a cluster to AWS with the Openshift Installer,
 you will need to download the pull secret for your account.
 Download this file from [openshift.com](https://cloud.openshift.com/clusters/install)
-and place in the `data` directory at the root level of the project.
-If there is no `data` directory, create one.
+and place in the `data` directory at the root level of the project
+If there is no `data` directory, create one or you can place the file in
+`$HOME/ocsci-data/`.
```
mbukatov: Maybe we could state that you can place the data directory either in root dir of the project or in home directory, with indication what is a safe default choice.

clacroix12:
Looks like there's a merge conflict in `exceptions.py` but other than that it LGTM.

docs/getting_started.md
```diff
@@ -46,8 +46,10 @@ cp ocs-ci.yaml.template ~/.ocs-ci.yaml
 In order to deploy a cluster to AWS with the Openshift Installer,
 you will need to download the pull secret for your account.
 Download this file from [openshift.com](https://cloud.openshift.com/clusters/install)
-and place in the `data` directory at the root level of the project.
-If there is no `data` directory, create one.
+and place in the `data` directory at the root level of the project
+(If there is no `data` directory, create one) or you can place the file in
+`$HOME/ocsci-data/`. Placing the data in `$HOME/ocsci-data/` is safer and you dont
```
petr-balogh: please try to also keep the line number under 79 characters, I see here in github that the word `dont` is on new line. BTW don't . 

The same line length below.

vasukulkarni: we have let docs go upto 120 lines since its rendered in browser to keep it consistent with previous changes or else my line would appear truncated compared to existing doc.

tests/test_ocs_basic_install.py
```diff
@@ -44,8 +45,16 @@ def run(**kwargs):
     if not cluster_path:
         cluster_path = os.path.join(cluster_dir_parent, cluster_name)
     run_cmd(f"mkdir -p {cluster_path}")
-    pull_secret_path = os.path.join(templating.TOP_DIR, "data", "pull-secret")
-    with open(pull_secret_path, "r") as f:
+    pull_secret_file = os.path.join(templating.TOP_DIR, "data", "pull-secret")
```
clacroix12: Would it be worth creating a function that retrieves the pull secret now that there is more than one filepath to check? I also think there were changes to how we read the pull-secret in #133  that we need to consider before making this change.

PR #61, opened 2019-05-17
Basic pytest conf for custom plugins

petr-balogh:


petr-balogh:
Hey, didn't find any PR about some POC for running tests, so I just finished with one example for configuration, still WIP, just some basic which we can discuss if we will chose this approach or different one.

You can try to run:
`py.test --ocs-conf conf/ocs_basic_install.yml ./tests/`
or `py.test --ocs-conf -m run_this conf/ocs_basic_install.yml ./tests/`

ocsci/enums.py
```diff
@@ -1,8 +1,8 @@
 from enum import Enum
 
 
-class TestStatus(Enum):
-    # As we have function to transform RC to TestStatus Enum we should add new
+class StatusOfTest(Enum):
```
clacroix12: Do we even need this enum anymore if we are utilizing pytest? I think tests should be asserting anything they need as passing criteria and we should let pytest handle the test status on it's own. Thoughts?

petr-balogh: We don't need them ;) I just didn't want to deal with cleanup of everything yet till we have pytest ready. There was issue with Test in the name of class and pytest tried to execute it. So I used easiest way for me and just renamed all occurrences for now to prevent pytest to consider it as test ;)

pytest_customization/plugins/ocscilib.py
```diff
@@ -0,0 +1,37 @@
+"""
+This plugin allows you to setup all basic configuratio for pytest we need
```
RazTamir: configuratio*n*

petr-balogh: done

docs/getting_started.md
```diff
@@ -28,18 +28,18 @@ It is recommended that you use a python virtual environment to install the neces
...
-The template can be found at the top level of the repository, `ocs-ci.yaml.template`.
+The default file is in conf/ocsci/default_config.yaml.
+TODO: Once moved to pytest update properly this file: `ocs-ci.yaml.template`.
+
 The required keys are in the template. Values are placeholders and should be replaced by legitimate values.
 Values for report portal or polarion are only required if you plan on posting to that particular service.
 
-Move a copy of the template to your user directory and edit it from there with the proper values.
+Move a copy of the template to your conf directory and edit it from there with
+the proper values and pass it with --ocs-conf parameter to pytest.
```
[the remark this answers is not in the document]
petr-balogh: done

docs/usage.md
```diff
@@ -55,3 +77,12 @@ python run.py --cluster-name=my-testing-cluster \
...
     --no-email
 ```
+
+### For pytest
+
+```bash
+ py.test -m deployment  --ocsci-conf conf/ocsci/custom_config.yaml \
+    --cluster-conf conf/ocs_basic_install.yml \
+    --cluster-name kerberos_ID-ocs-deployment \
+    --cluster-path /home/my_user/my-ocs-dir tests/
```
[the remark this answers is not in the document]
petr-balogh: Actually I didn't run this command on existing cluster. This was first deployment, and I just wanted to make sure I have my cluster dir under cluster-path, also wanted to have some specific name, so I think that if I want to achieve this I need to pass both params. Anyway, I didn't change logic how we had it. Just moving them to diff place for pytest. I think that we can improve or change the logic in some new PR if needed.

utility/utils.py
```diff
@@ -844,3 +845,24 @@ def wait_for_func_status(self, result):
...
+
+
+def update_dict_recursively(d, u):
+    """
+    Update dict recursively to not delete nested dict under second and more
+    nested level.
+
+    Args:
+        d (dict): Dict to update
+        u (dict): Other dict used for update d dict
```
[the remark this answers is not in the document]
petr-balogh: done

conf/ocsci/default_config.yaml
```diff
@@ -0,0 +1,55 @@
...
+  cli_params: {}  # this will be filled with CLI parameters data
+
+# In this section we are storing all deployment related configuration
+DEPLOYMENT:
+  installer_version: "4.1.0-rc.3"
+
+# Section for reporting configuration
+REPORTING:
+  email:
+    address: "ocs-ci@redhat.com"
```
fbalak: Is this address ok?

Isn't there a chance that it can generate a lot of spam from all users that try to run tests in their local environment and won't change this address? Also is this address ok here in upstream?

petr-balogh: Currently in pytest we don't have this emailing enabled at all.
Once we implement it, it would be with some --email parameter enabled (By default I wouldn't send any email as don't see the sense of it for local run when you see the results in your console)

Just thinking in advance and wanted to put something under this reporting section as example ;)

mbukatov: If this is just an example, I would rather make it explicit by using `example.com` domain.

mbukatov:
Just a note about an edge case: this will work fine, but it won't be possible to use configuration in parametrization code.

petr-balogh:
Verified deployment:

py.test -m deployment --cluster-conf conf/ocs_basic_install.yml --cluster-name pbalogh-pr61 --cluster-path /home/jenkins/cluster-dir-pr-61 --log-level info tests/
OUTPUT OMITTED 
INFO     tests.test_ocs_basic_install:test_ocs_basic_install.py:200 HEALTH_OK, install successful.
PASSED  
===============================================================================
1 passed, 1 deselec
ted in 2130.18 seconds ===============================================================================

RazTamir:
@petr-balogh is this still WIP or I can start a review like it is ready for merge?

clacroix12:
> Just a note about an edge case: this will work fine, but it won't be possible to use configuration in parametrization code.

@mbukatov can you elaborate on this a bit? Not sure if we plan on heavily using parametrize but we should discuss this if something we're doing here entirely blocks this feature of pytest for us.

pytest_customization/plugins/ocscilib.py
```diff
@@ -0,0 +1,141 @@
+"""
```
mbukatov: Would it make sense to make plugins directory placed directly in root of the repo?

mbukatov:
> @mbukatov can you elaborate on this a bit? Not sure if we plan on heavily using parametrize but we should discuss this if something we're doing here entirely blocks this feature of pytest for us.

I rechecked and the proposed solution seems to work. I overlooked the fact that you are actually just adding your own command line parameters and process the config on your own (I was assuming that you are adding pytest config options ...).

The use case I have in mind is this: we have some values in the config file, eg:

```
DEMO:
 - 1
 - 2
```

And we would like to run test case parametrized by these values:

```
import pytest

from ocsci import config as ocsci_config

@pytest.mark.parametrize("item", ocsci_config.DEMO)
def test_demo_parametrized_config(item):
    assert item is not None
```

And as I already noted, this works fine:

```
(venv) [ocsqe@localhost ocs-ci]$ python -m pytest -p pytest_customization.plugins.ocscilib  --ocsci-conf conf/ocsci/default_config.yaml tests/test_demo.py -v 
================================================================ test session starts ================================================================
platform linux -- Python 3.7.3, pytest-4.5.0, py-1.8.0, pluggy-0.11.0 -- /home/ocsqe/projects/ocs-ci/venv/bin/python
cachedir: .pytest_cache
rootdir: /home/ocsqe/projects/ocs-ci, inifile: pytest.ini
collected 2 items                                                                                                                                   

tests/test_demo.py::test_demo_parametrized_config[1] PASSED                                                                                   [ 50%]
tests/test_demo.py::test_demo_parametrized_config[2] PASSED                                                                                   [100%]

============================================================= 2 passed in 0.02 seconds ==============================================================
```

tox.ini
```diff
@@ -4,8 +4,6 @@ envlist = py37,flake8
 [testenv]
 deps =
     -rrequirements.txt
-    pytest
-commands = {envpython} -m pytest
```
mbukatov: Dropping it like this will disable (aka break) unit test run. We should just reconfigure it so that tox still runs unit tests, by adding parameters like `--ignore=tests` or `-c` (if we need custom pytest.ini for unittests):

```
commands = {envpython} -m pytest  --ignore=tests -c unittests/pytest.ini unittests
```

conf/ocsci/default_config.yaml
```diff
@@ -0,0 +1,55 @@
+---
```
mbukatov: Btw why this can't be in `conf/` directory directly?

pytest_customization/plugins/ocscilib.py
mbukatov: I mean, which other pytest customization could we have here other than plugins?

petr-balogh:
> @petr-balogh is this still WIP or I can start a review like it is ready for merge?

yes it's ready to review

conf/ocsci/default_config.yaml
petr-balogh: I wanted to separate ocsci configuration here, there are going to be a lot of configuration for different cluster configurations we are going to have. So we will need to somehow structure them. It will be mess when we will mix those configurations in one folder I think.

tox.ini
petr-balogh: I agree, but as we don't have any unittests related to our changed ocsci, I've just disabled this till we will have some and then we can re-enable, but yeah. I can do what you have suggested.

conf/ocsci/default_config.yaml
mbukatov: Ah, ok.

pytest_customization/plugins/ocscilib.py
petr-balogh: I was thinking about defining for example markers module here as well and define all markers like

```python
import pytest

tier1 = pytest.mark.tier1
tier2 = pytest.mark.tier3
```

then from test modules we can import

```python
from pytest_customization.markers import tier1


@tier1
def test_tier1():
    pass
```

Cannot thing now about any other, cause hooks are like plugins but we can separate them to hooks package maybe.

mbukatov: I would rather not do that. This is additional level of abstraction, which would increase complexity a bit without any benefit. When I look at `@tier1` function annotation, I have no idea what it actually means. But looking at `@pytest.mark.tier1` gives a clear indication that this is a pytest marker.

That said, maintaining list of valid markers is not a bad idea. We can do that via [Registering markers](https://docs.pytest.org/en/latest/example/markers.html#registering-markers), which can be combined with `--strict-markers` to prevent using unregistered markers (it could be enabled by default in pytest.ini as well).

petr-balogh: I think that  in latest pytest there is warning message automatically printed that you are using non registered markers, not sure about if it's just warning or it's by default strict. I am already registering those markers in pytest.ini file as you can see @mbukatov .

And about abstraction, we used it in RHV a lot and found it quite handy, cause it allowed us to define something like this:
```python
BEFORE_UPGRADE = 1
UPGRADE = 2
BEFORE_UPGRADE_HOSTS = 3
UPGRADE_HOSTS = 4
AFTER_UPGRADE_HOSTS = 5
UPGRADE_CL = 6
AFTER_UPGRADE_CL = 7
UPGRADE_DC = 8
AFTER_UPGRADE = 9

# Tests ordering
FIRST_TEST = AFTER_UPGRADE + 1
TEST = FIRST_TEST + 1
LAST_TEST = TEST + 1

# Polarion decorator
polarion = pytest.mark.polarion

# Bugzilla decorator
# https://github.com/rhevm-qe-automation/pytest_marker_bugzilla
bz = pytest.mark.bugzilla

# Jira decorator
# https://github.com/vkondula/pytest_jira/
jira = pytest.mark.jira

network = pytest.mark.network
sla = pytest.mark.sla
storage = pytest.mark.storage
coresystem = pytest.mark.coresystem
virt = pytest.mark.virt
integration = pytest.mark.integration

# used for tests that are not adjusted to GE or tests that we don't want to run
do_not_run = pytest.mark.do_not_run(value=17)
endless = pytest.mark.endless(value='endless')
upgrade = pytest.mark.upgrade(value='upgrade')
tier1 = pytest.mark.tier1(value=1)
tier2 = pytest.mark.tier2(value=2)
tier3 = pytest.mark.tier3(value=3)
tier4 = pytest.mark.tier4(value=4)
timeout = pytest.mark.timeout
storages = pytest.mark.storages

tier_markers = [tier1, tier2, tier3, tier4, upgrade, do_not_run]
team_markers = [network, sla, storage, coresystem, virt, upgrade]

# order markers for ordering tests
order_before_upgrade = pytest.mark.run(order=BEFORE_UPGRADE)
order_upgrade = pytest.mark.run(order=UPGRADE)
order_before_upgrade_hosts = pytest.mark.run(order=BEFORE_UPGRADE_HOSTS)
order_upgrade_hosts = pytest.mark.run(order=UPGRADE_HOSTS)
order_after_upgrade_hosts = pytest.mark.run(order=AFTER_UPGRADE_HOSTS)
order_upgrade_cluster = pytest.mark.run(order=UPGRADE_CL)
order_after_upgrade_cluster = pytest.mark.run(order=AFTER_UPGRADE_CL)
```

We had some plugins which for example added default timeout for different tiers.
Or thanks to value=X we were able to easily know which number of test is directly from test.
Another usage was about ordering tests for pytest order plugin as you can see our order markers for upgrade tests.

So instead of ppl decorated the tests all the time like @pytest.mark.run(order=UPGRADE_CL) they just used @order_upgrade_cluster .

So this will is advantage we used for the markers defined like this and I think that ppl will know about those markers if we will document them properly.

@RazTamir  WDYT here?

RazTamir: I'm used to this way as this is how we worked in RHV so +1.
If there were any issues with that in the past, I don't know but if you know @petr-balogh so maybe we can enhance it here as part of the implementation.

In overall, using @tier1 is very easy usage imo and people will get used to it (@mbukatov comment above).

This whole set provides a great flexibility

pytest.ini
```diff
@@ -0,0 +1,9 @@
+[pytest]
+markers =
+    run_this: testing marker for run this test, useful for development
+    deployment: marker for deployment
```
fbalak: I think that all deployment should be kept in fixtures. It shouldn't be done by tests.

pytest_customization/plugins/ocscilib.py
mbukatov: I see your point with odering markers, but otherwise I'm not convinced that it's a good idea.

I'm curious what is this about: `pytest.mark.tier1(value=1)`. What is this `value=1` used for, when the marker is called `tier1` already? I know that one can read this value from fixture or test code, but what is the point here exactly?

vasukulkarni: > I think that in latest pytest there is warning message automatically printed that you are using non registered markers, not sure about if it's just warning or it's by default strict. I am already registering those markers in pytest.ini file as you can see @mbukatov .
> 
> And about abstraction, we used it in RHV a lot and found it quite handy, cause it allowed us to define something like this:
> 
> ```python
> BEFORE_UPGRADE = 1
> UPGRADE = 2
> BEFORE_UPGRADE_HOSTS = 3
> UPGRADE_HOSTS = 4
> AFTER_UPGRADE_HOSTS = 5
> UPGRADE_CL = 6
> AFTER_UPGRADE_CL = 7
> UPGRADE_DC = 8
> AFTER_UPGRADE = 9
> 
> # Tests ordering
> FIRST_TEST = AFTER_UPGRADE + 1
> TEST = FIRST_TEST + 1
> LAST_TEST = TEST + 1
> 
> # Polarion decorator
> polarion = pytest.mark.polarion
> 
> # Bugzilla decorator
> # https://github.com/rhevm-qe-automation/pytest_marker_bugzilla
> bz = pytest.mark.bugzilla
> 
> # Jira decorator
> # https://github.com/vkondula/pytest_jira/
> jira = pytest.mark.jira
> 
> network = pytest.mark.network
> sla = pytest.mark.sla
> storage = pytest.mark.storage
> coresystem = pytest.mark.coresystem
> virt = pytest.mark.virt
> integration = pytest.mark.integration
> 
> # used for tests that are not adjusted to GE or tests that we don't want to run
> do_not_run = pytest.mark.do_not_run(value=17)
> endless = pytest.mark.endless(value='endless')
> upgrade = pytest.mark.upgrade(value='upgrade')
> tier1 = pytest.mark.tier1(value=1)
> tier2 = pytest.mark.tier2(value=2)
> tier3 = pytest.mark.tier3(value=3)
> tier4 = pytest.mark.tier4(value=4)
> timeout = pytest.mark.timeout
> storages = pytest.mark.storages
> 
> tier_markers = [tier1, tier2, tier3, tier4, upgrade, do_not_run]
> team_markers = [network, sla, storage, coresystem, virt, upgrade]
> 
> # order markers for ordering tests
> order_before_upgrade = pytest.mark.run(order=BEFORE_UPGRADE)
> order_upgrade = pytest.mark.run(order=UPGRADE)
> order_before_upgrade_hosts = pytest.mark.run(order=BEFORE_UPGRADE_HOSTS)
> order_upgrade_hosts = pytest.mark.run(order=UPGRADE_HOSTS)
> order_after_upgrade_hosts = pytest.mark.run(order=AFTER_UPGRADE_HOSTS)
> order_upgrade_cluster = pytest.mark.run(order=UPGRADE_CL)
> order_after_upgrade_cluster = pytest.mark.run(order=AFTER_UPGRADE_CL)
> ```
> 
> We had some plugins which for example added default timeout for different tiers.
> Or thanks to value=X we were able to easily know which number of test is directly from test.
> Another usage was about ordering tests for pytest order plugin as you can see our order markers for upgrade tests.
> 
> So instead of ppl decorated the tests all the time like @pytest.mark.run(order=UPGRADE_CL) they just used @order_upgrade_cluster .
> 
> So this will is advantage we used for the markers defined like this and I think that ppl will know about those markers if we will document them properly.
> 
> @RazTamir WDYT here?

Looking at the above ordering makes me *nervous*, Just to set the ordering right, is this the only way to do it? We are not only going to confuse the new comers here but unnecessarily create a complexity and for this reason alone its a  -1 from me.  I cannot believe how our previously simplicity of just ordering the tests in yaml solved this problem so neatly.

mbukatov: > Looking at the above ordering makes me _nervous_, Just to set the ordering right, is this the only way to do it? We are not only going to confuse the new comers here but unnecessarily create a complexity and for this reason alone its a -1 from me. I cannot believe how our previously simplicity of just ordering the tests in yaml solved this problem so neatly.

Ordering test cases using markers is not a bad idea. That said, in our case with OCS, most of the ordering could be resolved via fixture tracking automatically and so explicit ordering via markers may not be that important anyway.

petr-balogh: @vasukulkarni The thing is that the tests in pytest way shouldn't be ordered at all and shouldn't depend on some previous one and should be triggered independently.  By default they will run in the order how they are defined in module class I think.

I am not telling that we should start using ordering now, I am just telling what is the advantage of have defined those markers on one place and just import them and mart test with @tier1.

In upgrade tests we needed to really have test ordered to do some operations before datacenter updated, before cluster updated and so on.. So we used this way how to order tests by order-plugin and just mark the tests in which order they should run in upgrade testing.

@mbukatov value=1 fro tier 1 was used in some of our pytest plugin we had and needed to consider the number of tier. So was easier to take it from value from pre defined marker which everyone used than to pare the number of tier from the name of marker. Also we had for example upgrade tear, and as you can see it doesn't contain any number in the marker.

At some point we will get to situation when we will need to inject some values to markers and I think that have them defined on one place will be easy to use them.

RazTamir: @vasukulkarni I'll try to explain from a tester pov:

We can create a file with classes like:

```pyhon
@manage
class Manage()
    pass

@E2E
class e2e()
    pass
```

And in each test we will have something like:
```python
from testm_classes import Manage

@tier1  # This should be very easy to understand based on the test content or Polarin
class TestCaseOCS12345(Manage):  # this test will inherit the @manage decorator
```

The ordering can be done in the teams_classes.py like the team decorators.

So to recap, the only thing that a tester should take care of is adding a decorator on its class

pytest.ini
petr-balogh: @fbalak I am just moving stuff from what we had done by previous runner to be able to run it by pytest to let us move to pytest runner as soon as possible. We still didn't decided how to handle deployment or cleanup here in pytest. This will be part of different PR if we decide to do it different way.

dahorak: I think, that since we will have to test also the installation itself, it is IMHO test.

pytest_customization/plugins/ocscilib.py
mbukatov: > @mbukatov value=1 fro tier 1 was used in some of our pytest plugin we had and needed to consider the number of tier. So was easier to take it from value from pre defined marker which everyone used than to pare the number of tier from the name of marker. Also we had for example upgrade tear, and as you can see it doesn't contain any number in the marker.
>
> At some point we will get to situation when we will need to inject some values to markers and I think that have them defined on one place will be easy to use them.

If you know beforehand that you need to process value of fixtures in plugins, it's easier to just use markers like `@pytest.marker.tier(value=1)`.

pytest.ini
fbalak: Ok

pytest_customization/plugins/ocscilib.py
petr-balogh: @mbukatov yep, it's possible but we faced the issue with inheritance and reading the real value of marker kwargs value as this wasn't easily possible that time. But this issue was already fixed in pytetst I think.

Just found some related issues with markers we faced in past:
https://github.com/pytest-dev/pytest/issues/2515
https://github.com/pytest-dev/pytest/issues/3605

petr-balogh: @mbukatov another thing which I've just realized and which was the main point is that you cannot call pytest -m tier(1)   so that's why pytest -m tier1 and team for example

mbukatov: @petr-balogh good point

tox.ini
petr-balogh: done

pytest_customization/plugins/ocscilib.py
petr-balogh: I added some structure in next commit. So please review once more @red-hat-storage/top-level-reviewers

tests/ecosystem/deployment/test_ocs_basic_install.py
```diff
@@ -0,0 +1,214 @@
+import json
```
petr-balogh: FYI this is basically old test case just changed few stuff which moved to pytest. So pleas don't look on code improvements here which were already implemented like this in [origin test](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/test_ocs_basic_install.py)

ebenahar:
@petr-balogh I haven't seen bugzilla configuration for the bz decorator. Shouldn't it be part of the scope here?

mbukatov:
I don't know. I not a big fan of having the markers defined in 4 different places.

petr-balogh:
> @petr-balogh I haven't seen bugzilla configuration for the bz decorator. Shouldn't it be part of the scope here?

I made basic functionality here, the rest of things needs to be done in separated PRs I think.
Also can be made by someone else than me.

petr-balogh:
> I don't know. I not a big fan of having the markers defined in 4 different places.

You are talking about added: ocsci/__init__.py ?

This if for easier imports for users.
As I don't expect that markers will be added so frequently I would rather define it here like this again to have easy import just like:  `from ocsci import tir1, manage`
and then just mark the test with 
```python
@tier1
@manage
```

but if most of you don't like this approach we can omit this one.

petr-balogh:
I submitted another commit with some documentation, let's discuss this on our mtg today in 1 hour.

mbukatov:
> Overall LGTM. Still not sure I'm convinced we get a lot of benefit from importing the markers in `ocsci/__init__.py` for the added complexity but it's not a blocker for me.

+1 I guess we would need to revamp the end result anyway, to polish edges a bit.

petr-balogh:
@clacroix12 @RazTamir ok, just verified that old runner is still working.
I've ran:
`python run.py --cluster-name pbalogh-pr61 --cluster-path /home/jenkins/cluster-dir-pr-61  -s suites/basic_testing.yml --no-destroy --no-email`
And test cases passed on existing cluster.
Destroyed cluster with:
`python run.py  --cluster-path /home/jenkins/cluster-dir-pr-61 --destroy`
Also cluster destroyed without any issue.

Tried to run test like:
```python
import logging
import pytest

from ocs import ocp
from ocs import defaults
from ocsci.config import ENV_DATA
from ocsci import run_this, EcosystemTest, tier1


OCP = ocp.OCP(
    kind='pods', namespace=defaults.ROOK_CLUSTER_NAMESPACE
)

@run_this
def test_not_run_me_fail_pass():
    logger.info("Hey from test which should pass")
    logger.info(
        "You can easily access data from ENV_DATA like cluster_name: %s",
        ENV_DATA['cluster_name']
    )
    assert 1 == 1, "This will not reach this message"


@tier1
class TestExampleClass(EcosystemTest):
    def test_example_method(self):
        logger.info("Hello from test method inside test class")
        pods = OCP.get()['items']
        number_of_pods = len(pods)
        logger.info(
            f"Found {number_of_pods} pods in namespace: "
            f"{defaults.ROOK_CLUSTER_NAMESPACE}"
        )
        for pod in pods:
            logger.info(f"Pod : {pod.metadata.name}")
        assert number_of_pods, "No pod exists!"
```

Ran with `Py.test -m tier1 --cluster-conf conf/ocs_basic_install.yml --cluster-name pbalogh-pr61 --cluster-path /home/jenkins/cluster-dir-pr-61  --capture=no tests/` - status: passed,

So I think we can merge

RazTamir:
@petr-balogh I have no idea what I'm missing but I can't execute as pytest doesn't recognize the custom parameters:

(.venv) [~/git/ocs-ci]$ pytest -v -m "deployment" --ocsci-conf conf/ocsci/custom_config.yaml --cluster-conf conf/ocs_basic_install.yml --cluster-name ratamir_ocs_cluster --cluster-path /home/ratamir/ocs_automation
[DEPRECATION WARNING]: ansible.constants.BECOME_METHODS is deprecated, please use ansible.plugins.loader.become_loader. This list is statically defined and may not include all become 
methods. This feature will be removed in version 2.10. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --ocsci-conf --cluster-conf conf/ocs_basic_install.yml --cluster-name ratamir_ocs_cluster --cluster-path /home/ratamir/ocs_automation
  inifile: /home/ratamir/git/ocs-ci/pytest.ini
  rootdir: /home/ratamir/git/ocs-ci

PR #70, opened 2019-05-20
Add wait() method

RazTamir:
Something similar to:
 https://github.com/openshift/openshift-restclient-python#watch

### DIFF
--- a/ocs/ocp.py
+++ b/ocs/ocp.py
@@ -65,7 +65,7 @@ class OCP(object):
         out = run_cmd(cmd=oc_cmd)
         return munchify(yaml.safe_load(out))
 
-    def get(self, resource_name='', out_yaml_format=True):
+    def get(self, resource_name='', out_yaml_format=True, selector=None):
         """
         Get command - 'oc get <resource>'
 
@@ -79,6 +79,8 @@ class OCP(object):
             Munch Obj: this object represents a returned yaml file
         """
         command = f"get {self.kind} {resource_name}"
+        if selector is not None:
+            command += f"--selector={selector}"
         if out_yaml_format:
             command += " -o yaml"
         return self.exec_oc_cmd(command)
@@ -117,7 +119,7 @@ class OCP(object):
         """
         command = f"delete -f {yaml_file}"
         if wait:
-            command += " --wait=True"
+            command += " --wait=true"
         return self.exec_oc_cmd(command)
 
     def apply(self, yaml_file):
@@ -134,23 +136,42 @@ class OCP(object):
         command = f"apply -f {yaml_file}"
         return self.exec_oc_cmd(command)
 
-    def wait_for_resource_status(
-        self, resource_name, condition, timeout=30, sleep=3
+    def watch(
+        self, condition, resource_name='', selector=None, resource_count=0,
+        timeout=60, sleep=3
     ):
         """
         Wait for a resource to reach to a desired condition
 
         Args:
-            resource_name (str): The name of the resource to wait
-                for (e.g.my-pv1)
             condition (str): The desired state the resource should be at
                 This is referring to: status.phase presented in the resource
                 yaml file
                 (e.g. status.phase == Available)
+            resource_name (str): The name of the resource to wait
+                for (e.g.my-pv1)
+            selector (str): The resource selector to search with.
+                Example: 'app=rook-ceph-mds'
+            resource_count (int): How many resources expected to be
+
+        Returns:
+            bool: True in case all resources reached desired condition,
+                False otherwise
+
         """
         for sample in TimeoutSampler(
-            timeout, sleep, self.get, resource_name
+            timeout, sleep, self.get, resource_name, True, selector
         ):
-            if sample.status.phase == condition:
-                return True
+            in_condition = []
+            pods = sample['items']
+            for pod in pods:
+                if pod.status.phase == condition:
+                    in_condition.append(pod)
+                if resource_count:
+                    if len(in_condition) == resource_count and (
+                        len(pods) == len(in_condition)
+                    ):
+                        return True
+                elif len(pods) == len(in_condition):
+                    return True
         return False
--- a/ocs/ocp.py
+++ b/ocs/ocp.py
@@ -136,7 +136,7 @@ class OCP(object):
         command = f"apply -f {yaml_file}"
         return self.exec_oc_cmd(command)
 
-    def watch(
+    def wait(
         self, condition, resource_name='', selector=None, resource_count=0,
         timeout=60, sleep=3
     ):
@@ -162,16 +162,22 @@ class OCP(object):
         for sample in TimeoutSampler(
             timeout, sleep, self.get, resource_name, True, selector
         ):
-            in_condition = []
-            pods = sample['items']
-            for pod in pods:
-                if pod.status.phase == condition:
-                    in_condition.append(pod)
-                if resource_count:
-                    if len(in_condition) == resource_count and (
-                        len(pods) == len(in_condition)
-                    ):
+            # More than 1 resources returned
+            if sample.kind == 'List':
+                in_condition = []
+                sample = sample['items']
+                for item in sample:
+                    if item.status.phase == condition:
+                        in_condition.append(item)
+                    if resource_count:
+                        if len(in_condition) == resource_count and (
+                            len(sample) == len(in_condition)
+                        ):
+                            return True
+                    elif len(sample) == len(in_condition):
                         return True
-                elif len(pods) == len(in_condition):
+            # Only 1 resource returned
+            else:
+                if sample.status.phase == condition:
                     return True
         return False
--- a/ocs/ocp.py
+++ b/ocs/ocp.py
@@ -104,20 +104,32 @@ class OCP(object):
 
         return self.exec_oc_cmd(command)
 
-    def delete(self, yaml_file, wait=True):
+    def delete(self, yaml_file=None, resource_name='', wait=True):
         """
         Deletes a resource
 
         Args:
             yaml_file (str): Path to a yaml file to use in 'oc delete -f
                 file.yaml
+            resource_name (str): Name of the resource you want to delete
             wait (bool): Determines if the delete command should wait to
                 completion
 
         Returns:
             Munch Obj: this object represents a returned yaml file
         """
-        command = f"delete -f {yaml_file}"
+        if yaml_file is None and not resource_name:
+            log.error(
+                "At least one of resource_name or yaml_file have to "
+                "be provided"
+            )
+            return False
+
+        command = f"delete "
+        if resource_name:
+            command += f"{self.kind} {resource_name}"
+        else:
+            command += f"-f {yaml_file}"
         if wait:
             command += " --wait=true"
         return self.exec_oc_cmd(command)
--- a/ocs/ocp.py
+++ b/ocs/ocp.py
@@ -154,7 +154,7 @@ class OCP(object):
 
     def wait_for_resource(
         self, condition, resource_name='', selector=None, resource_count=0,
-        timeout=60, sleep=3
+        to_delete=False, timeout=60, sleep=3
     ):
         """
         Wait for a resource to reach to a desired condition
@@ -169,6 +169,8 @@ class OCP(object):
             selector (str): The resource selector to search with.
                 Example: 'app=rook-ceph-mds'
             resource_count (int): How many resources expected to be
+            to_delete (bool): Determines if wait_for_resource should wait for
+                a resource to be deleted
             timeout (int): Time in seconds to wait
             sleep (int): Sampling time in seconds
 
@@ -198,29 +200,7 @@ class OCP(object):
                             return True
                     elif len(sample) == len(in_condition):
                         return True
-
-        return False
-
-    def wait_for_delete(
-        self, resource_name='', selector=None, timeout=60, sleep=3
-    ):
-        """
-        Wait for a resource to reach to a desired condition
-
-        Args:
-            resource_name (str): The name of the resource to wait
-                for (e.g.my-pv1)
-            selector (str): The resource selector to search with.
-                Example: 'app=rook-ceph-mds'
-
-        Returns:
-            bool: True in case all resources reached desired condition,
-                False otherwise
-
-        """
-        for sample in TimeoutSampler(
-            timeout, sleep, self.get, resource_name, True, selector
-        ):
-            if not sample:
+            if to_delete and not sample:
                 return True
+
         return False
--- a/ocs/ocp.py
+++ b/ocs/ocp.py
@@ -152,7 +152,7 @@ class OCP(object):
         command = f"apply -f {yaml_file}"
         return self.exec_oc_cmd(command)
 
-    def wait(
+    def wait_for_resource(
         self, condition, resource_name='', selector=None, resource_count=0,
         timeout=60, sleep=3
     ):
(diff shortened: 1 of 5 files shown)


ocs/ocp.py
```diff
@@ -134,23 +136,42 @@ def apply(self, yaml_file):
         command = f"apply -f {yaml_file}"
         return self.exec_oc_cmd(command)
 
-    def wait_for_resource_status(
-        self, resource_name, condition, timeout=30, sleep=3
+    def watch(
```
shylesh: Isn't this functionality same as "oc wait" ?

RazTamir: That's true!

My intention was to split into watch() and wait() that will use watch - fixing

vasukulkarni:
Needs to be squashed and merged

ocs/ocp.py
```diff
@@ -134,23 +148,48 @@ def apply(self, yaml_file):
         command = f"apply -f {yaml_file}"
         return self.exec_oc_cmd(command)
 
-    def wait_for_resource_status(
-        self, resource_name, condition, timeout=30, sleep=3
+    def wait(
```
shylesh: @RazTamir ,

My point here was, if we already have a "oc wait" command which does same job , why are you using indirect method of achieving same result using "get" method ?
From openshift doc this is what I understood, 

```
wait
Wait for a specific condition on one or more resources.

Example: Wait for the python-1-mz2rf Pod to be deleted
$ oc wait --for=delete pod/python-1-mz2rf
```
If this is what exactly you are doing why can't you just wrap "oc wait " command instead of using "oc get" cli. Also "oc wait --timeout=yyy" is  available.

Am I missing something here ?

RazTamir: Hi @shylesh ,
I think I taked about the differences from oc wait cmd in the automation session yesterday (watch it (: )
From what I saw, oc wait can be used for delete and pod/ready which is very limited to our needs.
In that implementation, you are open to any condition and any resource to sample.
Also, you can specify how many resources should be in that condition

ocs/ocp.py
```diff
@@ -102,22 +104,34 @@ def create(self, yaml_file, out_yaml_format=True):
...
                 file.yaml
+            resource_name (str): Name of the resource you want to delete
             wait (bool): Determines if the delete command should wait to
                 completion
 
         Returns:
             Munch Obj: this object represents a returned yaml file
         """
-        command = f"delete -f {yaml_file}"
+        if yaml_file is None and not resource_name:
```
petr-balogh:
```python
if not (yaml_file or resource_name):
```
this will do the same I think

ocs/ocp.py
```diff
@@ -102,22 +104,34 @@ def create(self, yaml_file, out_yaml_format=True):
...
+            log.error(
+                "At least one of resource_name or yaml_file have to "
+                "be provided"
+            )
+            return False
```
petr-balogh: returning different type than is defined in Returns section where you mentioned Munch Obj.
Shouldn't we rather raise the exception here as this is unexpected from user to not pass any of those params?

ocs/ocp.py
petr-balogh: shouldn't we rename the function name to wait_for_resource than case to not confuse with wait form cli?

ocs/ocp.py
```diff
@@ -134,23 +148,48 @@ def apply(self, yaml_file):
         command = f"apply -f {yaml_file}"
         return self.exec_oc_cmd(command)
 
-    def wait_for_resource_status(
-        self, resource_name, condition, timeout=30, sleep=3
+    def wait(
+        self, condition, resource_name='', selector=None, resource_count=0,
+        timeout=60, sleep=3
```
petr-balogh: Missing documentation for timeout and sleep

ocs/ocp.py
```diff
@@ -134,23 +148,48 @@ def apply(self, yaml_file):
...
 
         Args:
-            resource_name (str): The name of the resource to wait
-                for (e.g.my-pv1)
             condition (str): The desired state the resource should be at
                 This is referring to: status.phase presented in the resource
                 yaml file
                 (e.g. status.phase == Available)
+            resource_name (str): The name of the resource to wait
+                for (e.g.my-pv1)
```
petr-balogh: mention default values in docstring (default: def_value), for all below.

ocs/ocp.py
```diff
@@ -134,23 +148,48 @@ def apply(self, yaml_file):
...
+                        in_condition.append(item)
+                    if resource_count:
+                        if len(in_condition) == resource_count and (
+                            len(sample) == len(in_condition)
+                        ):
+                            return True
+                    elif len(sample) == len(in_condition):
+                        return True
+            # Only 1 resource returned
+            else:
```
petr-balogh: what if you had resource_count set to 2? I think this case you should return True in this else branch and continue to sampling

ocs/ocp.py
```diff
@@ -102,22 +104,34 @@ def create(self, yaml_file, out_yaml_format=True):
 
         return self.exec_oc_cmd(command)
 
-    def delete(self, yaml_file, wait=True):
+    def delete(self, yaml_file=None, resource_name='', wait=True):
```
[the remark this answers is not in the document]
RazTamir: yes,
All checked

ocs/ocp.py
RazTamir: yep, thanks:
```python
>>> if yaml is None and not res:
...     print "123"
...     
... 
123
>>> if not (yaml and res):
...     print "123"
...     
... 
123

```

ocs/ocp.py
RazTamir: fixed to raise CommandError

ocs/ocp.py
RazTamir: fixed

ocs/ocp.py
RazTamir: do we really need that?
a python func with parameter and '=some_value' is enough to understand imo
WDYT?

ocs/ocp.py
RazTamir: Not sure I understand

ocs/ocp.py
petr-balogh: I see in you example is `and` , but I think that you would like to do the operation if none of both yaml and res is not defined right? 
So for this example I see:
nor image

Maybe I am missing something, can you elaborate please?

ocs/ocp.py
petr-balogh: actually don't mind, I just was using this as we mentioned it in our [coding guidelines](https://github.com/red-hat-storage/ocs-ci/blob/master/docs/coding_guidelines.md)

ocs/ocp.py
petr-balogh: You have parameter resource_count right? I understand it that you can wait til 2 resources are created.
So you wait for 2 resources for example. But on line [178](https://github.com/red-hat-storage/ocs-ci/pull/70/files#diff-9d7ae45bb7efabea0529c53dbca720f9R178) you are checking if kind is List.

I am not sure what is the behavior here, but if it won't be list and only one resource returned as you mentioned on line 191 you are going directly to else block.
checking it's status and if it's == condition returning `True`.

So you actually ending your wait with True status, but you have only 1 resource created but expected to wait for 2 resources.

Do I missing something or misunderstand this function?

ocs/ocp.py
RazTamir: yeah this example is not valid.
In the code is 'or'

ocs/ocp.py
RazTamir: No you are right.
Fixing now and nice catch!

RazTamir: So it seems like that List is always returned from this command:
2019-05-22 16:25:05,073 - utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /tmp/ratamir-test-ocs4-31174/auth/kubeconfig get Pod --selector=app=rook-ceph-mdsa -o yaml
DEBUG:utility.utils:CMD output: apiVersion: v1
items: []
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""

RazTamir: Ok found the thing:
1) when providing resource_name you are expected to get 1 resource 
2) when using selector --> can have multiple so List is returned

RazTamir:
Verified:
TEST NAME              TEST DESCRIPTION                                     DURATION                        STATUS
testing OCS            PV creation testing                                  4.293256521224976               PASSED
testing OCS            Modifying a CephFilesystem                           54.55412030220032               PASSED

ocs/ocp.py
shylesh: > Hi @shylesh ,
> I think I taked about the differences from oc wait cmd in the automation session yesterday (watch it (: )
> From what I saw, oc wait can be used for delete and pod/ready which is very limited to our needs.
> In that implementation, you are open to any condition and any resource to sample.
> Also, you can specify how many resources should be in that condition

Hi @RazTamir ,

I don't think "oc wait" is specific to pod and delete ops. we can exercise it for any resource and conditions will be specific to resources. For ex: for pods condition could be Creating, Running etc whereas for deployments condition=Running doesn't hold and we have to check list of applicable conditions.

I feel finding number of resources matching your selector deserves its own function rather than combining it with wait. In general wait semantics should never be worried about count of specific condition match and only either fail based on timeout OR return True if  condition met by resources shortlisted through selectors. It should be upto the consumer to decide whether they want count OR not based on context of test and may issue further get call.

shylesh: @RazTamir 

I did a dirty cli experiment and it works may not be perfect one.
```bash
 $ ~/oc-tools/oc wait --timeout=10s  -l rook_cluster=rook-ceph deployment --for=condition=available -n  rook-ceph -o=jsonpath='{.kind}{"\n"}' | wc -l
7
```
This returns number of deployments which satisfies given condition.

RazTamir: +1 on your comments.

Few things in my mind:
1) At some scenarios, the pods will spin up after a few seconds and not right away which means the return value from this cli command will be 0 because when you executed it, the number of pods is still 0
2) I will split the function into wait() and wait for_resouce_count() and hope this will serve your feedback
3) the fact that a cli command provides a functionality doesn't mean we have to stick with it.

shylesh: > +1 on your comments.
> 
> Few things in my mind:
> 
> 1. At some scenarios, the pods will spin up after a few seconds and not right away which means the return value from this cli command will be 0 because when you executed it, the number of pods is still 0

[shylesh] Not necessarily, because wait will not return until condition is met OR timeout. If pod has condition=Running that means pod is up and running so I guess there could be no window
of missing the pod, though I haven't tried.  

> 2. I will split the function into wait() and wait for_resouce_count() and hope this will serve your feedback
[shylesh] that would work.

> 3. the fact that a cli command provides a functionality doesn't mean we have to stick with it.
[Shylesh] I agree. I was just curious to see how that command works hence I tried.

RazTamir: @shylesh , I'm just going through all data I can find regarding oc wait and I can't find too much information about waiting on different conditions other than available/ready.

For now, I will change the name of the function to wait_for_resource and will leave the implementation of a pure wait to later phase

suites/basic_testing.yml
```diff
@@ -1,14 +1,14 @@
 tests:
+#- test:
```
petr-balogh: was it commented out by mistake?

tests/test_add_mds_to_cluster.py
```diff
@@ -35,8 +35,8 @@ def create_ceph_fs(**kwargs):
         yaml.dump(file_y, yaml_file, default_flow_style=False)
     log.info(f"Creating a new Ceph FileSystem")
     assert CEPHFS.create(yaml_file=TEMP_YAML_FILE)
-
-    assert POD.wait(
+    from ipdb import set_trace;set_trace()
```
petr-balogh: Also probably doesn't belong here

ocs/ocp.py
shylesh: > @shylesh , I'm just going through all data I can find regarding oc wait and I can't find too much information about waiting on different conditions other than available/ready.
> 
> For now, I will change the name of the function to wait_for_resource and will leave the implementation of a pure wait to later phase

Yes, please go ahead as this is one of the useful feature required for writing tests. We can revisit this later.

suites/basic_testing.yml
RazTamir: yes, fixed

tests/test_add_mds_to_cluster.py
RazTamir: True, fixed

tests/test_add_mds_to_cluster.py
```diff
@@ -43,35 +50,37 @@ def modify_fs(new_active_count):
...
     log.info(f"Deleting the file system")
-    stat = OCP.delete(TEMP_YAML_FILE)
+    stat = CEPHFS.delete(resource_name=fs_name)
     if CEPHFS_DELETED.format(cephfs_name=fs_name) in stat:
         return True
     return False
 
 
-def verify_fs_exist(fs_name):
+def verify_fs_exist(pod_count):
```
Shrivaibavi: The filesystem exists even after we delete the all MDS pods. Can we check for the name of the pod in the filesystem instead of pod_count. The pod name gets created along with the filesystem name, hence we could exactly know the pod has got created in the filesystem.
Please correct me if my understanding was wrong.

RazTamir: I'm not deleting pods here but the cephfilesystem.
I'm checking for pod count because when deleting the cephfilesystem, I'd expect that all pods will be deleted as well

ocs/ocp.py
```diff
@@ -134,23 +152,75 @@ def apply(self, yaml_file):
...
+                        if len(in_condition) == resource_count and (
+                            len(sample) == len(in_condition)
+                        ):
+                            return True
+                    elif len(sample) == len(in_condition):
+                        return True
+
+        return False
+
+    def wait_for_delete(
```
ebenahar: the name indicates we wait for the resource to be deleted while the implementation doesn't

suites/basic_testing.yml
```diff
@@ -1,14 +1,14 @@
 tests:
 - test:
     name: testing OCS
-    desc: Minimal testing of OCS
+    desc: PV creation testing
```
ebenahar: I think it would be better to leave this empty

suites/basic_testing.yml
```diff
@@ -1,14 +1,14 @@
...
     module: test_pv_creation.py
     config:
       destroy-cluster: False
     abort-on-fail: True
 - test:
     name: testing OCS
-    desc: Minimal testing of OCS
+    desc: Modifying a CephFilesystem
```
ebenahar: same

ocs/ocp.py
RazTamir: Fixed,
I added the functionality of waiting for deletion to wait_for_resource so you can specify 'to_delete' = True and the func will wait until no results will exist from the timoutSampler (sample is empty)

ocs/ocp.py
```diff
@@ -134,23 +152,55 @@ def apply(self, yaml_file):
...
+
         """
         for sample in TimeoutSampler(
-            timeout, sleep, self.get, resource_name
+            timeout, sleep, self.get, resource_name, True, selector
         ):
-            if sample.status.phase == condition:
+            # Only 1 resource expected to be returned
+            if resource_name:
+                if sample.status.phase == condition:
```
ebenahar: @RazTamir in case sample is of kind List, does it have status.phase?
Approving the PR but just want to make sure that this was checked.

RazTamir: No, but in case resource name is provided, you can be sure that only one resource will return

Issue #76, opened 2019-05-21
change the ceph failure domain to "host"


ramkiperiy:
Ceph failure domain should be "host" in storage-manifest.yaml.

With OSD level failure domain, failure of one node with multiple OSDs will lead to IO failure and data unavailability. 

Replica size should be 3.

ramkiperiy:
Closing this issue

PR #77, opened 2019-05-21
Updating storage-manifest.yaml with failure domain and replica_size

ramkiperiy:
Changed the failure domain from OSD to host and updated the replica_size to 3 in jinja template


ramkiperiy:
fix for #76

templates/ocs-deployment/storage-manifest.yaml
```diff
@@ -5,9 +5,9 @@ metadata:
   name: rbd
   namespace: {{ cluster_namespace | default('openshift-storage') }}
 spec:
-  failureDomain: osd
+  failureDomain: host
```
vasukulkarni: This is interesting in AWS :)

PR #89, opened 2019-05-24
Create a private namespace.

wusui:
In order to create pvc's outside of openshift-storage, I need a namespace for them.


utility/make_private_namespace.py
```diff
@@ -0,0 +1,45 @@
+""" Create a namespace """
+import os
+import subprocess
+import shlex
+from utility import templating
+from ocs.utils import create_oc_resource
+
+
+def make_private_namespace(name, region='us-east-2'):
```
RazTamir: create_...

utility/make_private_namespace.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Create a namespace.
+
+    Args:
+        name (str): namespace to be created
```
RazTamir: please use capital letters in the beginning of description

utility/make_private_namespace.py
```diff
@@ -0,0 +1,45 @@
...
+        region (str): aws region.  Defaults to us-east-2
+
+    Returns:
```
RazTamir: I don't understand this test,
It will never fail

utility/make_private_namespace.py
```diff
@@ -0,0 +1,45 @@
...
+    output, error = process.communicate()
+    if error:
+        print(f"Error executing: {oc_cmd}")
+        return
+    prev_ns = [x for x in output.decode().split('\n') if x.startswith(f"{name} ")]
+    if prev_ns:
+        print(f"{name} already exists")
+        return
+    env_data = dict()
+    env_data['platform'] = 'AWS'
```
RazTamir: there is no 'platform' in the private.yaml

utility/make_private_namespace.py
```diff
@@ -0,0 +1,45 @@
...
+    env_data['cluster_namespace'] = name
+    env_data['region'] = region
+    local_dir = f"/tmp/{name}"
+    if not os.path.exists(local_dir):
+        os.makedirs(local_dir)
+    templ_parm = templating.Templating()
+    create_oc_resource('private.yaml', local_dir, templ_parm, env_data)
+
+
+if __name__ == "__main__":
```
RazTamir: please take a look at the recording of the live demos.
We are not executing simple python scripts

utility/make_private_namespace.py
wusui: Done.  I should probably change the name of the file to create_private_namespace but I am not sure if this would confuse the review process.

utility/make_private_namespace.py
wusui: Done

utility/make_private_namespace.py
wusui: I have changed this routine and the comment to return False on failures.

utility/make_private_namespace.py
wusui: Done.  Accidentally left this in.

utility/make_private_namespace.py
wusui: Done.

utility/make_private_namespace.py
```diff
@@ -0,0 +1,41 @@
...
+
+def create_private_namespace(name, region='us-east-2'):
+    """
+    Create a namespace.
+
+    Args:
+        name (str): Namespace to be created
+        region (str): AWS region.  Defaults to us-east-2
+
+    Returns:
```
clacroix12: Instead of returning True/False can we raise an Exception in the event that we are unable to create the namespace? We have a number of custom Exceptions in `ocs/exceptions.py` where you can add a new one if necessary.

templates/ocs-deployment/private.yaml
```diff
@@ -0,0 +1,9 @@
+###################################################################################################################
+# Create a private namespace 
+###################################################################################################################
+---
+# Namespace used to test individual scripts. 
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ cluster_namespace }}
```
clacroix12: If this file is not used as a part of the actual OCS deployment it doesn't belong in `templates/ocs-deployment`. We should create another directory under `templates/` to house this new file.

utility/make_private_namespace.py
wusui: Using  OCP.exec_oc_cmd simplifies this.  It raises the exceptions that I think are appropriate.

utility/make_private_namespace.py
```diff
@@ -0,0 +1,41 @@
...
+        False if unable to create the new namespace.  If True, a directory in
+        /tmp with the namespace's name should contain the private.yaml file
+        that created this namespace.
+
+    """
+    oc_cmd = "oc get namespace"
+    process = subprocess.Popen(shlex.split(oc_cmd), stdout=subprocess.PIPE)
+    output, error = process.communicate()
```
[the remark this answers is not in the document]
wusui: Fixed in update. OCP.exec_oc_cmd is now being used.

templates/ocs-deployment/private.yaml
wusui: I added a new directory (test-deployment) and changed the create_oc_resource call.

wusui:
Now that I have looked at this a bit, create_oc_resource raises and exception when the namespace already
exists.  If that is what we want to happen here, then I don't think that the "get namespace" exec is necessary here.

utility/namespace.py
```diff
@@ -25,5 +25,5 @@ def create_private_namespace(name, region='us-east-2'):
     if not os.path.exists(local_dir):
         os.makedirs(local_dir)
     templ_parm = templating.Templating()
-    create_oc_resource('private.yaml', local_dir, templ_parm,
+    create_oc_resource('namespace.yaml', local_dir, templ_parm,
```
RazTamir: fix parenthesis according to the coding guidelines

utility/namespace.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Raises AlreadyExists exception if the namespace already exists.
+
+    Args:
+        name (str): Namespace to be created
+        region (str): AWS region.  Defaults to us-east-2
+    """
+    env_data = dict()
+    env_data['cluster_namespace'] = name
+    env_data['region'] = region
```
mbukatov: What does a region do here? I'm wondering how is a k8s namespace related to it?

mbukatov:
Just a note, don't read this as objection against this pr: I'm considering whether having just the template yaml and a fixture for a k8s namespace would be better.

utility/namespace.py
wusui: Actually, there is no 'region' information in the generated yaml that is used by the oc command.  I think that I can safely remove it.

utility/namespace.py
wusui: Done

RazTamir:
Hi @wusui , you can achieve this functionality with ocp.py/OCP.create().
kind should be 'namespace' and just pass the name of it.
In case you want to create a new namespace and also move to it, you can use new_project() in ocp.py

utility/namespace.py
```diff
@@ -0,0 +1,29 @@
...
+    namespace.
+
+    Raises AlreadyExists exception if the namespace already exists.
+
+    Args:
+        name (str): Namespace to be created
+    """
+    env_data = dict()
+    env_data['cluster_namespace'] = name
+    local_dir = f"/tmp/{name}"
```
shylesh: I didn't understand the purpose of storing yaml locally, if we need any spec info related to the namespace we can extract it from "oc get -o yaml" command. May be I am missing something.

vasukulkarni:
@wusui do we still want this?

wusui:
I don't think we  need this anymore

Issue #12, opened 2019-04-23
RFE: handle oc binary doesn't exist


vasukulkarni:
probably check in the runner and bail out sooner...

```
2019-04-23 10:51:07,465 - test_ocs_basic_install - INFO - Executing command: oc cluster-info
2019-04-23 10:51:07,480 - __main__ - ERROR - Traceback (most recent call last):
  File "run.py", line 198, in run
    rc = test_mod.run(**test_kwargs)
  File "/Users/vasukulkarni/ocs-ci/tests/test_ocs_basic_install.py", line 80, in run
    run_cmd("oc cluster-info")
  File "/Users/vasukulkarni/ocs-ci/tests/test_ocs_basic_install.py", line 111, in run_cmd
    r = subprocess.run(cmd.split(), stdout=sys.stdout, stderr=sys.stderr, **kwargs)
  File "/Users/vasukulkarni/ocs-ci/venv2/lib/python3.7/site-packages/gevent/subprocess.py", line 1672, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/Users/vasukulkarni/ocs-ci/venv2/lib/python3.7/site-packages/gevent/subprocess.py", line 627, in __init__
    restore_signals, start_new_session)
  File "/Users/vasukulkarni/ocs-ci/venv2/lib/python3.7/site-packages/gevent/subprocess.py", line 1505, in _execute_child
    raise child_exception
FileNotFoundError: [Errno 2] No such file or directory: 'oc'
```

clacroix12:
We can do this pretty easily with `shutil.which("oc")`. If we don't find anything, we can log and raise an exception.

shylesh:
This could be problematic. In my case i downloaded oc-tools (which had oc binary) to a non-standard binary path which could lead to shutil.which("oc") fail. If we enforce restriction saying oc binary should be present in /usr/bin then this might work.

clacroix12:
@shylesh shutil.which() will return it if it is on the path. if it's not on the path, the `oc` commands we execute will end up failing anyways.

shylesh:
If this discussion is all about just fail if it doesn't exist then Yes it makes sense. What I meant is oc could be in custom path which may not be noticed by which() unless its in paths scanned by shutil.which(). Either we can document the path where we expect oc binary should be (either /usr/bin OR any path which could be added later in python path.  Also if oc-tools has different versions how to tell end users about this ? I feel its better we handle it inside out script.

[...]

PR #92, opened 2019-05-24
Download openshift client

dahorak:
Replacement of https://github.com/red-hat-storage/ocs-ci/pull/49 with changes made on top of recently merged PR https://github.com/red-hat-storage/ocs-ci/pull/61. 

* handle download of Openshift client (`oc`)
* download both Openshift client and installer into `./bin` directory
* point `PATH` env. variable to the `./bin` directory
* fixes: https://github.com/red-hat-storage/ocs-ci/issues/12, https://github.com/red-hat-storage/ocs-ci/issues/99


utility/utils.py
```diff
@@ -716,47 +715,136 @@ def destroy_cluster(cluster_path):
...
-    Download the openshift installer
+    Download the OpenShift installer binary, if not already present.
+    Update env. PATH and get path of the openshift installer binary.
 
     Args:
         version (str): version of the installer to download
 
     Returns:
-        str: name of the installer binary after being unpacked
+        str: path to the installer binary
```
RazTamir: Path with capital P

utility/utils.py
```diff
@@ -716,47 +715,136 @@ def destroy_cluster(cluster_path):
...
+    return installer_binary_path
+
+
+def get_openshift_client(version=defaults.CLIENT_VERSION):
+    """
+    Download the OpenShift client binary, if not already present.
+    Update env. PATH and get path of the oc binary.
+
+    Args:
+        version (str): version of the client to download
```
RazTamir: Capital

utility/utils.py
```diff
@@ -716,47 +715,136 @@ def destroy_cluster(cluster_path):
...
+
+    Returns:
+        str: path of the client binary
```
RazTamir: Capital

utility/utils.py
```diff
@@ -716,47 +715,136 @@ def destroy_cluster(cluster_path):
...
+    return client_binary_path
+
 
-    return installer_filename
+def get_openshift_mirror_url(file_name, version):
+    """
+    Format url to OpenShift mirror (for client and installer download).
+
+    Args:
+        file_name (str): name of file
```
RazTamir: Capital - fix all

utility/utils.py
dahorak: Fixed by 22ba458.

utility/utils.py
dahorak: Fixed by 22ba458.

utility/utils.py
dahorak: Fixed by 22ba458.

utility/utils.py
dahorak: Fixed by 22ba458.

utility/utils.py
```diff
@@ -716,47 +715,136 @@ def destroy_cluster(cluster_path):
...
+        delete_file(installer)
         return StatusOfTest.PASSED
 
     except Exception:
         log.error(traceback.format_exc())
         return StatusOfTest.FAILED
 
 
-def download_openshift_installer(version=defaults.INSTALLER_VERSION):
+def get_openshift_installer(version=defaults.INSTALLER_VERSION):
```
dahorak: @petr-balogh how can I rewrite this: `version=defaults.INSTALLER_VERSION` (and the other *defaults* variables like `CLIENT_VERSION` and `BIN_DIR`), to use the values from `default_config.yaml` file, instead of the old approach with `defaults` module?

petr-balogh: @dahorak you can define it in mentioned `default_config.yaml` in default section. But I think that this will be RUN specific, so we can directly maybe access this from RUN section and than it can be overwritten by passing custom config by --ocsci-conf custom-config.yaml by overwrite it in section it will be defined.

ocs/defaults.py
```diff
@@ -12,6 +13,7 @@
 CEPH_IMAGE = "quay.io/rhceph-dev/rhceph-4.0-rhel-8"
 ROOK_IMAGE = "quay.io/rhceph-dev/rook"
 DEPLOYMENT_PLATFORM = 'AWS'
+BIN_DIR = './bin'
```
petr-balogh: Ok, let's move this under the RUN section to default_config.yaml,

ocs/defaults.py
```diff
@@ -4,6 +4,7 @@
 from getpass import getuser
 
 INSTALLER_VERSION = '4.1.0-rc.3'
+CLIENT_VERSION = INSTALLER_VERSION
```
petr-balogh: Define this in default_config.yaml in RUN section, I think it's part of RUN and won't be used anywhere else I think. Anyway it can be taken from RUN section from ocsci/config.py module from whatever place

vasukulkarni: Can you link what file are you talking about here?

vasukulkarni:
I think this one is ready for merge, probably we can take the defaults changes later on after discussion.

utility/utils.py
dahorak: @petr-balogh I've probably miss something, or there is some unexpected behaviour...

If the `DEPLOYMENT['installer_version']` construction used in `test_ocs_basic_install.py` file (in `test_deployment` function), it works as expected.
But if I try to import it and use similar construction in this file (`utility/utils.py`), it fails with `KeyError: 'bin_dir'`.

I've tried to log content of `DEPLOYMENT` on both places, and in the `test_....py` file it contains expected keys and values, but in `utils.py` file it is empty dictionary:
```
09:02:26 - MainThread - tests.ecosystem.deployment.test_ocs_basic_install - INFO - {'bin_dir': './bin/', 'client_version': '4.1.0-rc.3', 'installer_version': '4.1.0-rc.3'}
09:02:26 - MainThread - utility.utils - INFO - {}
```

Is there any required order of imported modules, or how the configuration should be initialized to be available also in other files, not only in the `test_...py` file?

dahorak: Ok, I've hopefully find solution for this. Instead of:
```
from ocsci.config import DEPLOYMENT
DEPLOYMENT['installer_version']
```
I have to use something like:
```
from ocsci import config
config.DEPLOYMENT['installer_version']
```

dahorak:
I agree with @vasukulkarni, that it will be better to do the movement of the configuration to the `default_config.yaml` in separated PR, because right now, it is not possible to do the full movement, mainly because of cluster destroy functionality is still based on the old runner.

.gitignore
```diff
@@ -13,4 +13,5 @@ config.yaml
 *.bak
 openshift-install*
 data/pull-secret
-dist
+bin/
+dist
```
RazTamir: New empty line?

dahorak: yeah... it's github "feature" :-/ ...  when resolving conflict on the web, it will not add the new line as normal editor did... will fix it.

utility/utils.py
petr-balogh: I spent time on this issue yesterday and didn't come up yet with something with I am satisfied. So I am looking into this issue today as well and hope I will finally find some good solution for it.

ocs/defaults.py
petr-balogh: @vasukulkarni was talking about this file: https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/default_config.yaml . But I am trying to solve some problem which @dahorak  mentioned in his another comment which is not easy fix cause of import modules before we are actually loading config in some cases.

PR #94, opened 2019-05-24
Add Ceph cmd execution capability

ebenahar:


ocs/ocp.py
```diff
@@ -154,3 +154,20 @@ def wait_for_resource_status(
...
+        """
+        Execute a Ceph command on the Ceph tools pod
+
+        Args:
+            ceph_cmd (str): The Ceph command to execute on the Ceph tools pod
+
+        Returns:
+            str: Ceph command output in a Json format
+        """
+        pods_list = shlex.split(self.get(out_yaml_format=False))
```
RazTamir: What will happen if I will call this func with OCP object that is initialized for a resource that is not a "pod"?
Also, I think this function shouldn't be under OCP object but this can stay for now until we'll have better design and hierarchy

ocs/ocp.py
```diff
@@ -154,3 +154,20 @@ def wait_for_resource_status(
             if sample.status.phase == condition:
                 return True
         return False
+
+    # this method location is temporary, will be changed to ocs class
+    def exec_ceph_cmd(self, ceph_cmd):
```
RazTamir: Maybe you can make it more generic and implement 'exec_ccmd_on_pod'?
I'm not sure what are the differences between tools pod and mon pod for example in terms of executing commands but in case it is not that different maybe worth to handle it

ocs/ocp.py
ebenahar: @RazTamir you're right. My initial intention was to puth this function in a location that will instantiate an ocp object of kind pod.
For now, I'll do it in this same file and I'll move it to a more appropriate location later.

ocs/ocp.py
ebenahar: @RazTamir see new commit

ocs/ocp.py
```diff
@@ -154,3 +155,49 @@ def wait_for_resource_status(
...
+
+    def exec_cmd_on_pod(self, pod_name, command):
+        """
+
+        Args:
+            pod_name (str): The pod on which the command should be executed
+            command (str): The command to execute on the given pod
+
+        Returns:
+            Munch Obj: this object represents a returned yaml file
```
RazTamir: *T*his

ocs/ocp.py
```diff
@@ -66,11 +66,12 @@ def exec_oc_cmd(self, command):
         out = run_cmd(cmd=oc_cmd)
         return munchify(yaml.safe_load(out))
 
-    def get(self, resource_name='', out_yaml_format=True):
+    def get(self, selector='', resource_name='', out_yaml_format=True):
```
RazTamir: should be implemented as part of https://github.com/red-hat-storage/ocs-ci/pull/70/files#diff-9d7ae45bb7efabea0529c53dbca720f9R70

RazTamir:
@ebenahar can you please rebase?
Let's merge this patch quickly

ebenahar:
Verified. Executed a simple test that prints the output of 'ceph df' command:

```python
import logging
from ocs import ocp

class Test1:
    def test_1(self):
        out = ocp.exec_ceph_cmd("ceph df")
        logger.info(f"{out}")
```


tests/manage/test_1.py::Test1::test_1 
-------------------------------------------------------------------------------------------------- live log call --------------------------------------------------------------------------------------------------
14:14:12 - MainThread - utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ebenahar/ocs-ci-ebenahar-63293/auth/kubeconfig get pods --selector=app=rook-ceph-tools -o yaml
14:14:13 - MainThread - utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ebenahar/ocs-ci-ebenahar-63293/auth/kubeconfig rsh rook-ceph-tools-75867b94fc-nqj9q ceph df --format json-pretty
14:14:16 - MainThread - tests.manage.test_1 - INFO - {'stats': {'total_bytes': 318901321728, 'total_avail_bytes': 315630010368, 'total_used_bytes': 50085888, 'total_used_raw_bytes': 3271311360, 'total_used_raw_ratio': 0.010258067399263382}, 'stats_by_class': {'ssd': {'total_bytes': 318901321728, 'total_avail_bytes': 315630010368, 'total_used_bytes': 50085888, 'total_used_raw_bytes': 3271311360, 'total_used_raw_ratio': 0.010258067399263382}}, 'pools': [{'name': 'rbd', 'id': 1, 'stats': {'stored': 0, 'objects': 0, 'kb_used': 0, 'bytes_used': 0, 'percent_used': 0, 'max_avail': 99894976512}}, {'name': 'ocsci-cephblockpool', 'id': 10, 'stats': {'stored': 13035556, 'objects': 17, 'kb_used': 38496, 'bytes_used': 39419904, 'percent_used': 0.00013152052997611463, 'max_avail': 99894976512}}]}
PASSED

ocs/ocp.py
```diff
@@ -204,3 +205,50 @@ def wait_for_resource(
...
+    """
+    Get the Ceph tools pod
+
+    Returns:
+        str: The Ceph tools pod name
+    """
+    ocp_pod_obj = OCP(kind='pods', namespace=defaults.ROOK_CLUSTER_NAMESPACE)
+    ct_pod = ocp_pod_obj.get(
+        selector='app=rook-ceph-tools'
+    ).toDict()['items'][0]['metadata']['name']
```
petr-balogh: Look as Coady accessing this directly via jsonpath [here](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/ecosystem/deployment/test_ocs_basic_install.py), we can then reuse this function in installation as well.

ebenahar: Thanks @petr-balogh. 
You refer to:
```python
    tools_pod = run_cmd(
        f"oc -n {namespace} get pod -l 'app=rook-ceph-tools' "
        f"-o jsonpath='{{.items[0].metadata.name}}'"
    )
```

Is there any advantage of using it over what I did?

ocs/ocp.py
```diff
@@ -204,3 +205,50 @@ def wait_for_resource(
...
+    Args:
+        ceph_cmd (str): The Ceph command to execute on the Ceph tools pod
+
+    Returns:
+        dict: Ceph command output
+    """
+    ocp_pod_obj = OCP(kind='pods', namespace=defaults.ROOK_CLUSTER_NAMESPACE)
+    ct_pod = get_ceph_tools_pod()
+    ceph_cmd += " --format json-pretty"
+    return ocp_pod_obj.exec_cmd_on_pod(ct_pod, ceph_cmd).toDict()
```
mbukatov: Not all ceph command line tools has `--format json-pretty` option. Eg. `rados` knows it, but eg. `rbd` doesn't. But I don't think we need to address it here, one would just run `exec_cmd_on_pod(get_ceph_tools_pod(), cmd)` when one needs something like that.

ocs/ocp.py
petr-balogh: just that the you will safe toDict call and you will get directly the name from oc command. And if no such pod found that you will get raised exception from run_cmd.  But I don't mind, anyway, it's merged already :) :+1:

PR #101, opened 2019-05-27
Improving OCP.create() to support creation of resource name

RazTamir:


ocs/ocp.py
```diff
@@ -87,20 +87,31 @@ def get(self, resource_name='', out_yaml_format=True, selector=None):
...
+            raise CommandFailed(
+                "At least one of resource_name or yaml_file have to "
+                "be provided"
+            )
+        command = "create "
+        if yaml_file:
+            command += f"-f {yaml_file}"
+        elif resource_name:
+            # e.g "oc namespace my-project"
+            command = f"{self.kind} {resource_name}"
```
ebenahar: not sure if 'project' is considered as a kind here

RazTamir: It is not.
That's why added example line before:
# e.g "oc namespace my-project"
namespace and project are same

ebenahar: @RazTamir +1

ocs/ocp.py
```diff
@@ -152,6 +163,21 @@ def apply(self, yaml_file):
         command = f"apply -f {yaml_file}"
```
keesturam: Did you miss to delete this line?

RazTamir: no, why?
This is the command for apply

keesturam: if my understanding is correct, this nullifies whatever we set for  "command" in the previous lines.

keesturam: the misunderstanding was due to the fact that there were hidden lines of code :-) Thanks @RazTamir for pointing out that! LGTM.

Issue #102, opened 2019-05-27
When --cluster-name parameter is defined it is not applied


petr-balogh:

petr-balogh:
fixed in #103

PR #103, opened 2019-05-27
Fix issue with naming cluster

petr-balogh:
Fixing issue #102


pytest_customization/ocscilib.py
```diff
@@ -94,10 +94,11 @@ def process_cluster_cli_params(config):
         f"{ocsci_config.ENV_DATA['cluster_name']}-{getuser()}"
     )
     cluster_name = get_cli_param(config, 'cluster_name')
-    base_cluster_name = default_cluster_name
+    if cluster_name:
+        default_cluster_name = cluster_name
```
mbukatov: Wouldn't be  setting `cluster_name = default_cluster_name` and then using `cluster_name` more readable?

RazTamir:
@petr-balogh I think this is ready for merge

PR #115, opened 2019-05-29
run_cmd() should return even decoded stderr and return code of the command

PrasadDesala:
Currently run_cmd is only returning the decoded stdout of command.

We use this function for running most of the oc commands. It will
be good if return even the decoded stderr and return code
of the command as these will help to do negative scenarios
validations

With this PR, run_cmd will return below,
1) decoded stdout of the command
2) decoded stderr of the command
3) return code of the command

### DIFF
--- a/utility/utils.py
+++ b/utility/utils.py
@@ -635,6 +635,8 @@ def run_cmd(cmd, **kwargs):
 
     Returns:
         (str) Decoded stdout of command
+        (str) Decoded stderr of command
+        (str) return code of command
 
     """
     log.info(f"Executing command: {cmd}")
@@ -655,7 +657,7 @@ def run_cmd(cmd, **kwargs):
             f"Error during execution of command: {cmd}."
             f"\nError is {r.stderr.decode()}"
         )
-    return r.stdout.decode()
+    return r.stdout.decode(), r.stderr.decode(), r.returncode
 
 
 def download_file(url, filename):
--- a/utility/utils.py
+++ b/utility/utils.py
@@ -634,7 +634,7 @@ def run_cmd(cmd, **kwargs):
         CommandFailed: In case the command execution fails
 
     Returns:
-        (str, str, str) Decoded stdout, decoded stderr, and return code of command
+        tuple (str, str, int): Decoded stdout, decoded stderr, and return code of command
 
     """
     log.info(f"Executing command: {cmd}")
--- a/utility/utils.py
+++ b/utility/utils.py
@@ -634,7 +634,8 @@ def run_cmd(cmd, **kwargs):
         CommandFailed: In case the command execution fails
 
     Returns:
-        tuple (str, str, int): Decoded stdout, decoded stderr, and return code of command
+        tuple (str, str, int): Decoded stdout, decoded stderr, and 
+                               return code of command
 
     """
     log.info(f"Executing command: {cmd}")
--- a/utility/utils.py
+++ b/utility/utils.py
@@ -634,9 +634,7 @@ def run_cmd(cmd, **kwargs):
         CommandFailed: In case the command execution fails
 
     Returns:
-        (str) Decoded stdout of command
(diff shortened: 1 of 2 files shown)


utility/utils.py
```diff
@@ -635,6 +635,8 @@ def run_cmd(cmd, **kwargs):
 
     Returns:
         (str) Decoded stdout of command
+        (str) Decoded stderr of command
```
vasukulkarni: You want to convert this to list here.

clacroix12: Would it not be a tuple instead?

Ex.
```
(str, str, str) Decoded stdout, decoded stderr, and return code of command
```

vasukulkarni: yeah either list or tuple, the return signature should change, tuple is more valid here.

PrasadDesala: Done. Changed the return type to tuple.

utility/utils.py
```diff
@@ -634,7 +634,7 @@ def run_cmd(cmd, **kwargs):
         CommandFailed: In case the command execution fails
 
     Returns:
-        (str) Decoded stdout of command
+        (str, str, str) Decoded stdout, decoded stderr, and return code of command
```
ebenahar: tuple (str, str, str):

RazTamir:
@PrasadDesala I have a question about the existing implementation.
Currently, the command is returning the output, and in case of failure, rasing CommandError with the stderr as part of the exception so you will be able to parse it like:

```python
try:
    exec_oc_cmd()
except CommandError as ex:
   if "Failed to create X" in ex
return True
```

This example is for negative case - this is not sufficient?

PrasadDesala:
> @PrasadDesala I have a question about the existing implementation.
> Currently, the command is returning the output, and in case of failure, rasing CommandError with the stderr as part of the exception so you will be able to parse it like:
> 
> ```python
> try:
>     exec_oc_cmd()
> except CommandError as ex:
>    if "Failed to create X" in ex
> return True
> ```
> 
> This example is for negative case - this is not sufficient?

Yes, this way is doable and even infact I initially validated the same way in my local script. 
Two reasons for returning the values:
1) I would want to check whether the oc command failed or not, for which I want to make use of the return value from the command.
eg:
        if ret != 0:
            assert "PVC creation with invalid name succeeded : NOT expected"
2) I assumed that most of the negative scenarios would require such validations and thought it would help if the function returns all 3 (ret, out, err) so that can be used where ever necessary. 
By having it this way, in the test script we can simply validate without exceptions , 
if "text" in err:
    print ("expected")

These are just my thoughts, please correct me if I am wrong.

RazTamir:
You are not wrong at all.
This is just a different approach (better imo) but just wanted to understand the reason behind it

utility/utils.py
PrasadDesala: Updated the PR fixing this comment.

utility/utils.py
```diff
@@ -655,7 +656,7 @@ def run_cmd(cmd, **kwargs):
             f"Error during execution of command: {cmd}."
             f"\nError is {r.stderr.decode()}"
         )
-    return r.stdout.decode()
+    return (r.stdout.decode(), r.stderr.decode(), r.returncode)
```
RazTamir: I'm not sure I understand this change.
In case of error (i.e. retcode != 0) an exception will be raised and nothing will return
In case of positive flow, you can be sure retcode will be 0 and reterror will be empty

Please elaborate

PrasadDesala: > I'm not sure I understand this change.
> In case of error (i.e. retcode != 0) an exception will be raised and nothing will return
> In case of positive flow, you can be sure retcode will be 0 and reterror will be empty
> 
> Please elaborate

Raz, you are right. My bad, I missed it in my code read that the function is returning the exception when there is an error.

This function seems to be handling well and requires no changes. Apologies from my side.
I'll close this PR.

PR #116, opened 2019-05-29
Function for unique naming

ebenahar:


ebenahar:
Verified. Executed this simple test that prints the created name:

```python
import logging
from tests import helpers

class Test1:
    def test_create_name(self):
        name = helpers.create_unique_object_name('test', 'project')
        log.info(f"{name}")
```

tests/manage/test_1.py::Test1::test_create_name 
-------------------------------------------------------------------------------------------------- live log call --------------------------------------------------------------------------------------------------
12:10:25 - MainThread - tests.manage.test_1 - INFO - project_test_2912102521
PASSED

tests/helpers.py
```diff
@@ -0,0 +1,24 @@
...
+Helper functions file for OCS QE
+"""
+import datetime
+
+
+def create_unique_object_name(object_description, object_type):
+    """
+    Creates a unique object name by using the object_description
+    and object_type, as well as the current date/time string.
+    This can be used for any objects such as VMs, disks, clusters etc.
```
petr-balogh: VMs, disks, cluster etc -> Change to something OCS related ;)

ebenahar: Done :)

tests/helpers.py
```diff
@@ -0,0 +1,23 @@
+"""
+Helper functions file for OCS QE
+"""
+import datetime
+
+
+def create_unique_object_name(object_description, object_type):
```
RazTamir: I will change it to resource_description and resource_type

PrasadDesala: +1
Also, I think we need to change the function name the same way as the args name. Something to..
"create_unique_resource_name"

PrasadDesala:
Also just noticed that, this file is placed in tests/helpers.py which may not be a perfect place given that it is not a test case.
**Please ignore this comment if you have plans to move it at a later stage.

ebenahar:
@PrasadDesala I think this is the correct location since this function will be used for resources created as part of the tests

keesturam:
Looks good to me.

mbukatov:
> @PrasadDesala I think this is the correct location since this function will be used for resources created as part of the tests

I disagree. We should maintain some separation between test code and utility modules. Moreover if someone decides to create subdirectories in the tests directory, there is no reliable way to import the module.

ebenahar:
@PrasadDesala @mbukatov Ok, I get your point.
How about utility/utils.py?

PR #127, opened 2019-05-30
Update installer version to 4.1.0-rc.5

vavuthu:


vavuthu:
Tested with new installer version

$ oc get clusterversion
NAME      VERSION      AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.1.0-rc.5   True        False         22m     Cluster version is 4.1.0-rc.5

mbukatov:
Why this particular version? I see that as of now, the latest 4-stable rc version is [4.1.0-rc.7](https://openshift-release.svc.ci.openshift.org/releasestream/4-stable/release/4.1.0-rc.7). Btw how do we check that this matches the container images we use?

vavuthu:
> Why this particular version? I see that as of now, the latest 4-stable rc version is [4.1.0-rc.7](https://openshift-release.svc.ci.openshift.org/releasestream/4-stable/release/4.1.0-rc.7). 

I saw that rc7 is there but I am not confident that whether we can use rc7 or not bcoz its not poined to the [latest link](https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/ ). 

Btw how do we check that this matches the container images we use?
I am not sure which container images you are referring to. Currently we are using upstream bits of rook and ceph.

mbukatov:
> > Why this particular version? I see that as of now, the latest 4-stable rc version is [4.1.0-rc.7](https://openshift-release.svc.ci.openshift.org/releasestream/4-stable/release/4.1.0-rc.7).
> 
> I saw that rc7 is there but I am not confident that whether we can use rc7 or not bcoz its not poined to the [latest link](https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/).

Good point. Thanks for the explanation. 

> > Btw how do we check that this matches the container images we use?
> I am not sure which container images you are referring to. Currently we are using upstream bits of rook and ceph.

Sorry for the confusion, I mean version of openshift itself, which one can check via `oc adm release info --commits`.  And a better question would be: does it matter if installer and oc client version is bit different from openshift version?

PR #129, opened 2019-05-30
Testcase to remove MDS from the cluster (OCS-361)

Shrivaibavi:


Shrivaibavi:
Looks like my branch is messed up. hence closing PR

PR #133, opened 2019-05-30
Fix pull-secret handling

zmc:
openshift-install expects the pullSecret value to be a JSON string. Mine is
stored as a multiline object, and so the existing code to read it failed. I
really think a better approach is to use the existing parsers for both JSON and
YAML here, as they are extremely well-tested and supported.


zmc:
Before this commit, deployment would fail; after it, it succeeds 🎉

vasukulkarni:
thanks @zmc

PR #137, opened 2019-05-31
WIP: TestCase OCS-288 creating pvc with random sc

prsurve:

### DIFF
--- a/tests/manage/ocs_288.py
+++ b/tests/manage/ocs_288.py
@@ -1,7 +1,6 @@
 """
 A test for creating pvc with random sc
 """
-import base64
 import logging
 import os
 import random
@@ -12,33 +11,21 @@ import yaml
 
 from ocs import defaults
 from ocs import ocp
-from ocsci import ManageTest, tier1
+from ocsci.config import ENV_DATA
+from ocsci.testlib import tier1, ManageTest
+from resources import pod
+from resources.ocs import OCS
+from tests import helpers
 from utility import templating
 
 log = logging.getLogger(__name__)
-CEPH_FILESYSTEM = os.path.join(
-    "templates/ocs-deployment", "cephfilesystem.yaml"
-)
-SC_RBD_YAML = os.path.join(
-    "templates/CSI/rbd", "storageclass.yaml"
-)
-SECRET_RBD = os.path.join(
-    "templates/CSI/rbd", "secret.yaml"
-)
-SC_CEPHFS_YAML = os.path.join(
-    "templates/CSI/cephfs", "storageclass.yaml"
-)
+TEMPLATE_DIR = "templates"
+
 RBD_POOL_YAML = os.path.join(
-    "templates/CSI/rbd", "CephBlockPool.yaml"
+    "templates/ocs-deployment", "cephblockpool.yaml"
 )
-PVC_RBD_YAML = os.path.join(
-    "templates/CSI/rbd", "pvc.yaml"
-)
-SECRET_CEPHFS = os.path.join(
-    "templates/CSI/cephfs", "secret.yaml"
-)
+
 TEMP_YAML_FILE = 'temp.yaml'
-TEMP_YAML_FILE_FS = 'temp_fs.yaml'
 
 SC = ocp.OCP(
     kind='StorageClass', namespace=defaults.ROOK_CLUSTER_NAMESPACE
@@ -58,11 +45,11 @@ PVC = ocp.OCP(
 CFS = ocp.OCP(
     kind='CephFilesystem', namespace=defaults.ROOK_CLUSTER_NAMESPACE
 )
+tools_pod = pod.get_ceph_tools_pod()
 
 
 @pytest.fixture(scope='class')
-def test_fixture(request):
-
+def ocs288_fixture(request):
     def finalizer():
         teardown()
 
@@ -74,10 +61,15 @@ def setup():
     """
     Setting up the environment for the test
     """
-    assert create_secret(file_name=SECRET_RBD)
-    assert create_secret(file_name=SECRET_CEPHFS)
+    log.info("Creating RBD Secret")
+    assert create_secret_rbd()
+    log.info("Creating CEPHFS Secret")
+    assert create_secret_cephfs()
+    log.info("Creating CEPH FileSystem")
     assert create_cephfilesystem()
+    log.info("Creating RBD Storageclass")
     assert create_multiple_rbd_storageclasses(count=5)
+    log.info("Creating CEPHFS Storageclass")
     assert create_storageclass_cephfs()
 
 
@@ -85,11 +77,21 @@ def teardown():
     """
     Tearing down the environment
     """
+    global RBD_SECRET_OBJ, CEPHFS_SECRET_OBJ, CEPHFS_OBJ, CEPHFS_SC_OBJ
+    log.info("Deleting PVC")
     assert delete_pvc()
+    log.info("Deleting CEPH BLOCK POOL")
     assert delete_cephblockpool()
+    log.info("Deleting RBD Secret")
+    RBD_SECRET_OBJ.delete()
+    log.info("Deleting CEPHFS Secret")
+    CEPHFS_SECRET_OBJ.delete()
+    log.info("Deleting CEPH FILESYSTEM")
+    CEPHFS_OBJ.delete()
+    log.info("Deleting Cephfs Storageclass")
+    CEPHFS_SC_OBJ.delete()
+    log.info("Deleting RBD Storageclass")
     assert delete_storageclass()
-    assert delete_cephfilesystem(file_name=TEMP_YAML_FILE_FS)
-    assert delete_secret()
 
 
 def delete_pvc():
@@ -126,30 +128,6 @@ def delete_cephblockpool():
     return True
 
 
-def delete_cephfilesystem(file_name):
-    """
-    Function to delete CephFileSystem
-    :param file_name:
-    :return:
-    """
-    log.info("Deleting CephFileSystem")
-    assert CFS.delete(yaml_file=file_name)
-    return True
-
-
-def delete_secret():
-    """
-    Function to delete Secret
-    :param secret_list:
-    :return:
-    """
-    secret_list = get_secret()
-    for item in secret_list:
-        log.info(f"Deleting secret {item}")
-        assert SECRET.delete(resource_name=item)
-    return True
-
-
 def create_rbd_pool(pool_name):
     """
     Create Blockpool with specified name
@@ -157,8 +135,12 @@ def create_rbd_pool(pool_name):
     :param pool_name:
     :return:
     """
+
     pool_data = {}
-    pool_data['rbd_pool'] = pool_name
+
+    pool_data['cephblockpool_name'] = pool_name
+    pool_data['rook_api_version'] = defaults.ROOK_API_VERSION
+
     file_y = templating.generate_yaml_from_jinja2_template_with_data(
         RBD_POOL_YAML, **pool_data
     )
@@ -179,15 +161,15 @@ def validate_pool_creation(pool_name):
     """
     ceph_validate = False
     k8s_validate = False
-    cmd = "ceph osd lspools"
-    out = ocp.exec_ceph_cmd(cmd)
+    cmd = "ceph osd pool ls"
+    out = tools_pod.exec_ceph_cmd(ceph_cmd=cmd)
     for item in out:
-        if item['poolname'] == pool_name:
+        if item == pool_name:
             ceph_validate = True
     pool_obj = POOL.get()
     sample = pool_obj['items']
     for item in sample:
-        if item.metadata.name == pool_name:
+        if item.get('metadata')['name'] == pool_name:
             k8s_validate = True
     if ceph_validate and k8s_validate:
         return True
@@ -216,16 +198,19 @@ def create_cephfilesystem():
     :param kwargs:
     :return:
     """
-    file_y = templating.generate_yaml_from_jinja2_template_with_data(
-        CEPH_FILESYSTEM
+    fs_data = defaults.CEPHFILESYSTEM_DICT.copy()
+    fs_data['metadata']['name'] = helpers.create_unique_resource_name(
+        'test', 'cephfs'
     )
-    with open(TEMP_YAML_FILE_FS, 'w') as yaml_file:
-        yaml.dump(file_y, yaml_file, default_flow_style=False)
-    log.info(f"Creating a new CephFileSystem")
-    assert CFS.create(yaml_file=TEMP_YAML_FILE_FS)
-    fs_name = file_y.get('metadata')['name']
-    POD.wait_for_resource(condition="Running", selector='app=rook-ceph-mds')
-    assert validate_cephfilesystem(fs_name=fs_name)
+    log.info(fs_data)
+    fs_data['metadata']['namespace'] = ENV_DATA['cluster_namespace']
+    global CEPHFS_OBJ
+    CEPHFS_OBJ = OCS(**fs_data)
+    CEPHFS_OBJ.create()
+    assert POD.wait_for_resource(
+        condition='Running', selector='app=rook-ceph-mds'
+    )
+    assert validate_cephfilesystem(fs_name=fs_data['metadata']['name'])
     return True
 
 
@@ -238,9 +223,11 @@ def validate_cephfilesystem(fs_name):
     ceph_validate = False
     k8s_validate = False
     cmd = "ceph fs ls"
-    out = ocp.exec_ceph_cmd(cmd)
+    log.info(fs_name)
+    out = tools_pod.exec_ceph_cmd(ceph_cmd=cmd)
     if out:
         out = out[0]['name']
+        log.info(out)
         if out == fs_name:
             log.info("FileSystem got created from Ceph Side")
             ceph_validate = True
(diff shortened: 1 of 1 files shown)


tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+    """
+    assert delete_pvc(pvc_list)
+    assert delete_storageclass_and_cephblockpool(storageclass_list)
+    assert delete_cephfilesystem(file_name=TEMP_YAML_FILE_FS)
+    assert delete_secret(secret_list)
+
+
+def delete_pvc(pvc_list):
+    """
+    Function to delete pvc
```
RazTamir: fix docstring according to coding guidelines:
https://github.com/red-hat-storage/ocs-ci/blob/master/docs/coding_guidelines.md

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+        file_name, **secret_data
+    )
+    with open(TEMP_YAML_FILE, 'w') as yaml_file:
+        yaml.dump(file_y, yaml_file, default_flow_style=False)
+    assert SECRET.create(yaml_file=TEMP_YAML_FILE)
+    secret_list.append(file_y.get('metadata')['name'])
+    return True
+
+
+def create_pvc(**kwargs):
```
RazTamir: you are only passing 1 argument here - count.
So no need for **kwargs.

It should be 
```python
def create_pvc(count=1)
```

And with that, you also taking care of line 350

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+        pvc_count = int(kwargs.get('count'))
+    else:
+        pvc_count = 1
+    for i in range(pvc_count):
+        sc_name = random.choice(storageclass_list)
+        pvc_data = {}
+        pvc_data['rbd_storageclass_name'] = sc_name
+        pvc_data['pvc_name'] = "rbd-pvc-" + ''.join(
+            [random.choice(string.ascii_lowercase)
+             for _ in range(3)]) + str(sc_name)
```
RazTamir: please add explanation of what you are doing here.
I think we already have a function to do that

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+
+        file_y = templating.generate_yaml_from_jinja2_template_with_data(
+            PVC_RBD_YAML, **pvc_data
+        )
+        with open(TEMP_YAML_FILE, 'w') as yaml_file:
+            yaml.dump(file_y, yaml_file, default_flow_style=False)
+        assert PVC.create(yaml_file=TEMP_YAML_FILE)
+        pvc_name = pvc_data.get('pvc_name')
+        pvc_name = pvc_data.get('pvc_name')
```
RazTamir: why double call?

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+
+def setup(self):
+    """
+    Setting up the environment for the test
+    """
+    assert create_secret(file_name=SECRET_RBD)
+    assert create_secret(file_name=SECRET_CEPHFS)
+    assert create_cephfilesystem()
+    assert create_storageclass_rbd(get_pool_get_sc_name())
+    assert create_storageclass_rbd(get_pool_get_sc_name())
```
RazTamir: why all those redundant calls?

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+SECRET = ocp.OCP(
+    kind='Secret', namespace="default"
+)
+PVC = ocp.OCP(
+    kind='PersistentVolumeClaim', namespace=defaults.ROOK_CLUSTER_NAMESPACE
+)
+CFS = ocp.OCP(
+    kind='CephFilesystem', namespace=defaults.ROOK_CLUSTER_NAMESPACE
+)
+storageclass_list = []
```
RazTamir: should be class attributes

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+    self = request.node.cls
+
+    def finalizer():
+        teardown(self)
+
+    request.addfinalizer(finalizer)
+    setup(self)
+
+
+def setup(self):
```
RazTamir: you are not using self at all, please remove

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+    assert create_cephfilesystem()
+    assert create_storageclass_rbd(get_pool_get_sc_name())
+    assert create_storageclass_rbd(get_pool_get_sc_name())
+    assert create_storageclass_rbd(get_pool_get_sc_name())
+    assert create_storageclass_rbd(get_pool_get_sc_name())
+    assert create_storageclass_rbd(get_pool_get_sc_name())
+    assert create_storageclass_cephfs()
+
+
+def teardown(self):
```
RazTamir: not using self - please remove

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+    """
+    Tearing down the environment
+    """
+    assert delete_pvc(pvc_list)
```
RazTamir: pvc_list should be a class attribure and you can approach it with self.pvc_list

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+
+def get_storageclass():
+    """
+    Funtion for getting all storageclass
+    :return: storageclass list
+    """
+    sc_obj = SC.get()
+    sample = sc_obj['items']
+    ignore_sc = 'gp2'
+    for item in sample:
```
RazTamir: Use list comprehension - https://www.pythonforbeginners.com/basics/list-comprehensions-in-python :
```python
storageclass_list = [
    item.metadata.name for item in sample if (
        item.metadata.name not in ignore_sc
    )
]
```

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+        if item.metadata.name not in ignore_sc:
+            storageclass_list.append(item.metadata.name)
+    if len(storageclass_list) >= 1:
```
RazTamir: if len(storageclass_list):

len of empty is 0 and if 0: --> False

tests/manage/ocs_288.py
prsurve: Here we need to create multiple storageclass

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+    k8s side
+
+    :param pool_name:
+    :return:
+    """
+    ceph_validate = False
+    k8s_validate = False
+    cmd = "ceph osd lspools"
+    out = ocp.exec_ceph_cmd(cmd)
+    for item in out:
```
ebenahar: no need for going through all the items in the dict, just do check the 'poolname' key:

```python
if item['poolname'] == pool_name:
```

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+    Create a new CSI StorageClass
+    """
+    pool_name = kwargs.get('rbd_pool')
+    sc_name = kwargs.get('rbd_storageclass_name')
+    assert create_rbd_pool(pool_name=pool_name)
+    assert validate_pool_creation(pool_name)
+    if validate_pool_creation(pool_name):
+        log.info("Resource validation Passed")
+    else:
+        return 1
```
RazTamir: In line 280 you are returning a bool and here int.
What is the right returning type?
Please add a description to this function with an explanation of the return type

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+    )
+    with open(TEMP_YAML_FILE, 'w') as yaml_file:
+        yaml.dump(file_y, yaml_file, default_flow_style=False)
+    log.info(f"Creating a new RBD StorageClass")
+    assert SC.create(yaml_file=TEMP_YAML_FILE)
+    assert validate_storageclass(sc_name)
+    return True
+
+
+def get_pool_get_sc_name():
```
RazTamir: generate_pool_and_sc_names()

tests/manage/ocs_288.py
RazTamir: ok,
So please create a function called create_multiple_rbd_storageclasses() and inside it, in a loop, call to generate_pool_and_sc_names() (see my comment on the get_pool_get_sc_name() function)

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+        return False
+
+
+def validate_storageclass(sc_name):
+    """
+    Validate if storageClass is been created or not
+    """
+    sc_obj = SC.get()
+    sample = sc_obj['items']
+    for item in sample:
```
RazTamir: why not use get(resource_name=sc_name) and that's it?

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+    out = ocp.exec_ceph_cmd(cmd)
+    if out:
+        out = out[0]['name']
+        if out == fs_name:
+            log.info("FileSystem got created from Ceph Side")
+            ceph_validate = True
+        else:
+            log.error("FileSystem was not present at Ceph Side")
+            return False
+    result = CFS.get()
```
RazTamir: same - use get(resource_name=fs_name)

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+    sample = result['items']
+    for item in sample:
+        if item.metadata.name == fs_name:
+            log.info(f"Filesystem got created from kubernetes Side")
+            k8s_validate = True
+        else:
+            log.error("Filesystem was not create at Kubernetes Side")
+            return False
+    if ceph_validate and k8s_validate:
+        return True
```
RazTamir: return True if (ceph_validate and k8s_validate) else False

tests/manage/ocs_288.py
prsurve: Here i am creating pvc with prefix rbd-pvc and appending 3 random letter and at suffix appending storageclass name
Example:-
1:- rbd-pvc-eaccsi-cephfs (here eac is random text and csi-cephfs is storageclass)
2:- rbd-pvc-fwqocsci-csi-pool-ejtn (here fwq is random text and ocsci-csi-pool-ejtn is storageclass name)

tests/manage/ocs_288.py
prsurve: The output of cmd is in list format inside list their is dictionary so here we need to iterate to get the poolname

tests/manage/ocs_288.py
prsurve: keeping self now because I moved variables inside a class so to access them using self

tests/manage/ocs_288.py
prsurve: keeping self now because I moved variables inside a class so to access them using self

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+import pytest
+import yaml
+
+from ocs import defaults
+from ocs import ocp
+from ocsci import ManageTest, tier1
+from utility import templating
+
+log = logging.getLogger(__name__)
+CEPH_FILESYSTEM = os.path.join("templates/ocs-deployment",
```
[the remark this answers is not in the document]
prsurve: done

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,407 @@
...
+    :param pvc_list:
+    :return:
+    """
+    for item in pvc_list:
+        log.info(f"Deleting pvc {item}")
+        assert PVC.delete(resource_name=item)
+    return True
+
+
+def delete_storageclass_and_cephblockpool(storageclass_list):
```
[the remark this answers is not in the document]
prsurve: done

tests/manage/ocs_288.py
prsurve: done

tests/manage/ocs_288.py
prsurve: removed in latest commit

tests/manage/ocs_288.py
prsurve: replaced with single function create_multiple_rbd_storageclasses just need to pass count it will create specified no of rbd storgeclass

tests/manage/ocs_288.py
prsurve: done

tests/manage/ocs_288.py
prsurve: done

tests/manage/ocs_288.py
prsurve: done

tests/manage/ocs_288.py
prsurve: done

tests/manage/ocs_288.py
prsurve: done

tests/manage/ocs_288.py
prsurve: done

tests/manage/ocs_288.py
prsurve: done

tests/manage/ocs_288.py
prsurve: Done

tests/manage/ocs_288.py
prsurve: done

tests/manage/ocs_288.py
```diff
@@ -195,30 +204,28 @@ def validate_storageclass(sc_name):
...
-    for item in sample:
-        if item.metadata.name in sc_name:
-            log.info(f"StorageClass got created")
-            return True
-    return False
+    sc_obj = SC.get(resource_name=sc_name)
+    if sc_obj['metadata']['name']:
+        log.info(f"StorageClass got created")
+        return True
+    else:
```
RazTamir: no need for else:
just return False

tests/manage/ocs_288.py
```diff
@@ -296,23 +300,22 @@ def get_pool_get_sc_name():
...
+    storageclass = [
+        item.metadata.name for item in sample if (
+            item.metadata.name not in ignore_sc
+        )
+    ]
+    return storageclass
 
 
-def create_secret(file_name):
+def create_secret(self, file_name):
```
RazTamir: this is not good.
As this function is not a class method, it is very bad to use self here.

tests/manage/ocs_288.py
```diff
@@ -334,36 +337,30 @@ def create_secret(file_name):
...
-
         file_y = templating.generate_yaml_from_jinja2_template_with_data(
             PVC_RBD_YAML, **pvc_data
         )
         with open(TEMP_YAML_FILE, 'w') as yaml_file:
             yaml.dump(file_y, yaml_file, default_flow_style=False)
         assert PVC.create(yaml_file=TEMP_YAML_FILE)
         pvc_name = pvc_data.get('pvc_name')
-        pvc_name = pvc_data.get('pvc_name')
         pvc_list.append(pvc_name)
```
RazTamir: Ok few things about this function.
1) It should create pvcs - if you also what to wait for resources, please add new parameter called 'wait' type bool and in case it is True, wait_for_resource()
2) if this func should create pvcs - let it return the pvc list and don't take care of maintaining 'pvc_list' which is a class attribute.

tests/manage/ocs_288.py
ebenahar: @prsurve thanks, please ignore my comment

tests/manage/ocs_288.py
prsurve: 1-:Here in this function i am creating pvc as well as checking that pvc is in Bound
2:- I will remove pvc_list and create another function to get pvc_list for teardown step

tests/manage/ocs_288.py
prsurve: Removed self in the latest commit

tests/manage/ocs_288.py
prsurve: done

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,436 @@
...
+PVC = ocp.OCP(
+    kind='PersistentVolumeClaim', namespace=defaults.ROOK_CLUSTER_NAMESPACE
+)
+CFS = ocp.OCP(
+    kind='CephFilesystem', namespace=defaults.ROOK_CLUSTER_NAMESPACE
+)
+
+
+@pytest.fixture(scope='class')
+def test_fixture(request):
```
fbalak: Please change name of the fixture to the resource it provides or setup it does and update docstring so it is clear.

fbalak: And if this fixture could be shared among more testcases you can put it into `conftest.py` as shown in https://docs.pytest.org/en/latest/example/simple.html

tests/manage/ocs_288.py
```diff
@@ -0,0 +1,436 @@
...
+def create_rbd_pool(pool_name):
+    """
+    Create Blockpool with specified name
+
+    :param pool_name:
+    :return:
+    """
+    pool_data = {}
+    pool_data['rbd_pool'] = pool_name
+    file_y = templating.generate_yaml_from_jinja2_template_with_data(
```
vasukulkarni: Templating method is gone,  Can you check the new OCS() class and use that instead. If you need help ping me in QE chat.

prsurve: Done

tests/manage/ocs_288.py
prsurve: Done

vasukulkarni:
Remove the WIP branch and post the test results as well please,  Also final change is for POOL dict instead of template. Thanks

PR #139, opened 2019-05-31
Fix to get_ceph_tools_pod()

ebenahar:


ocs/ocp.py
```diff
@@ -257,11 +257,9 @@ def get_ceph_tools_pod():
         str: The Ceph tools pod name
     """
     ocp_pod_obj = OCP(kind='pods', namespace=defaults.ROOK_CLUSTER_NAMESPACE)
-    ct_pod = ocp_pod_obj.get(
-        selector='app=rook-ceph-tools'
-    ).toDict()['items'][0]['metadata']['name']
-    assert ct_pod, f"No Ceph tools pod found"
-    return ct_pod
+    ct_pods = ocp_pod_obj.get(selector='app=rook-ceph-tools')
+    assert ct_pods, f"No Ceph tools pods found"
```
vasukulkarni: Instead can we raise an exception, so that user can retry?

ocs/ocp.py
```diff
@@ -257,11 +257,9 @@ def get_ceph_tools_pod():
         str: The Ceph tools pod name
```
RazTamir: As soon as Shylesh implementation of the OCS design will be ready I think we want to return a Pod object here

ocs/ocp.py
ebenahar: @vasukulkarni not sure I follow. It does raise an exception, an assertion error

vasukulkarni: @ebenahar assert indicates a strong failure, we can just raise exception "CephToolsPodNotReady" and it will be clear at try/except block.

clacroix12: If we are concerned with wanting to retry can't we use `oc wait --for condition=ready` instead of `get`? This way we can configure the appropriate timeout and still just simply assert the success of that command?

vasukulkarni:
Can we close this one, Since I see its handled in recent PR?

RazTamir:
@ebenahar - is this still relevant?

ebenahar:
@RazTamir no, this is fixed already

PR #141, opened 2019-06-01
Installer is truncating the cluster name, use upto 8 chars from username

vasukulkarni:
Installer is truncating the cluster name so lets append our clusternames with upto 8

chars from getuser.

### DIFF
--- a/ocsci/pytest_customization/ocscilib.py
+++ b/ocsci/pytest_customization/ocscilib.py
@@ -93,7 +93,7 @@ def process_cluster_cli_params(config):
     # TODO: determine better place for parent dir
     cluster_dir_parent = "/tmp"
     default_cluster_name = (
-        f"{ocsci_config.ENV_DATA['cluster_name']}-{getuser()}"
+        f"{ocsci_config.ENV_DATA['cluster_name']}-{getuser()[:8]}"
     )
     cluster_name = get_cli_param(config, 'cluster_name')
     if cluster_name:


ocsci/pytest_customization/ocscilib.py
```diff
@@ -93,7 +93,7 @@ def process_cluster_cli_params(config):
     # TODO: determine better place for parent dir
     cluster_dir_parent = "/tmp"
     default_cluster_name = (
-        f"{ocsci_config.ENV_DATA['cluster_name']}-{getuser()}"
+        f"{ocsci_config.ENV_DATA['cluster_name']}-{getuser()[:8]}"
```
clacroix12: Are we limiting the length of `cluster_name` as well somewhere? Otherwise we might still hit issues where the name we give the installer is too long, depending on what you pass to `--cluster-name`, right?

Also is there a reason we construct the `default_cluster_name` here as well as have a value for it in [defaults.py](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs/defaults.py#L30)

vasukulkarni: This PR is for the default auto-generated name,  the default_cluster_name is just 'ocs-ci' as defined in our defaults, the behavior hasn't changed from before.

clacroix12: Right, so what I'm seeing here is that we get `ocsci_config.ENV_DATA['cluster_name']` which is `f"ocs-ci-cluster-{getuser()}"` and then append `-{getuser()[:8]}`. Wouldn't this result in having the username in the cluster_name twice?

In defaults.py `ENV_DATA['cluster_name'] = CLUSTER_NAME`, and `CLUSTER_NAME = f"ocs-ci-cluster-{getuser()}"`

Just trying to make sure we are constructing the name correctly in the default case. This issue isn't due to these changes, but unless I'm missing something it seems like we weren't constructing this correctly before.

vasukulkarni: Petr's PR changed this a bit,  ocsci_config.ENV_DATA['cluster_name'] is just "ocs-ci" ->https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/default_config.yaml#L52

vasukulkarni:
I have tested this and removing "needs-testing" label.

ocsci/pytest_customization/ocscilib.py
clacroix12: I see now, I was looking at `ocs/defaults.py` instead of `conf/ocsci/default_config.yaml`

Issue #66, opened 2019-05-20
ocs-ci/docs/getting_started.md: No mention of cloning the ocs-ci repo before Installing->Step#2


nehaberry:
Doc link - https://github.com/red-hat-storage/ocs-ci/blob/master/docs/getting_started.md

The document doesn't mention **to clone the OCS-CI repo** before asking to install " pip install -r requirement.txt". It would be helpful to have this information for first time users. 

Doc link
--------------
Getting Started ->Installing -> Step 2 : It mentions running "pip install -r requirements.txt" .

Current content:
-----------------------
1. Setup a python 3.7 virtual environment. This is actually quite easy to do now. 
--snip--

2. **Install requirements with pip install -r requirements.txt**

Change suggested:
--------------------------------
Please add a comment to **clone the repo** before trying to execute Step#2.

1. Clone the ocs-ci repository and run "pip install -r requirements.txt"  ( or any other equivalent statement for user's clear understanding)

Thanks,
Neha

PR #144, opened 2019-06-01
Fixing issue #66

petr-balogh:
Mention clone of repository before run pip install cmd.


petr-balogh:
Once merged, please close also the issue #66

PR #146, opened 2019-06-03
Added minor fixes for exec_ceph_cmd function

keesturam:
- Added additional argument format for 'exec_ceph_cmd' function
 - Added ability to handle non munchable objects returned from ceph commands

### DIFF
--- a/ocs/ocp.py
+++ b/ocs/ocp.py
@@ -5,7 +5,7 @@ import os
 import logging
 import yaml
 from ocsci.config import ENV_DATA
-from munch import munchify
+from munch import munchify, Munch
 
 from ocs.exceptions import CommandFailed
 from utility.utils import TimeoutSampler
@@ -53,7 +53,7 @@ class OCP(object):
                 without the initial 'oc' at the beginning
 
         Returns:
-            Munch Obj: this object represents a returned yaml file
+            Munch Obj if the object represents a returned yaml file, else returns a list
         """
         oc_cmd = "oc "
         kubeconfig = os.getenv('KUBECONFIG')
@@ -65,7 +65,10 @@ class OCP(object):
 
         oc_cmd += command
         out = run_cmd(cmd=oc_cmd)
-        return munchify(yaml.safe_load(out))
+        out = yaml.safe_load(out)
+        if isinstance(out, list):
+            return out
+        return munchify(out)
 
     def get(self, resource_name='', out_yaml_format=True, selector=None):
         """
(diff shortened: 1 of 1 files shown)


ocs/ocp.py
```diff
@@ -53,7 +53,7 @@ def exec_oc_cmd(self, command):
                 without the initial 'oc' at the beginning
 
         Returns:
-            Munch Obj: this object represents a returned yaml file
+            Munch Obj if the object represents a returned yaml file, else returns a list
```
RazTamir: When it will return a list?

keesturam: The output from Ceph commands is either in the form of dictionary or list. I've seen this in a couple of occasions.

The following is one such case.
```
sh-4.2# rbd ls rbd --format json 
["pvc-29e8c2cf-85d4-11e9-a278-0a334edd4bdc","pvc-771654f7-85d3-11e9-9454-02a058fe7dec"]
```
converting such an output to dict fails.
```
>   return [item.toDict() for item in out if item]
E   AttributeError: 'str' object has no attribute 'toDict'
```

ocs/ocp.py
```diff
@@ -264,22 +267,24 @@ def get_ceph_tools_pod():
...
+    ceph_cmd += f" --format {format_arg}"
     out = ocp_pod_obj.exec_cmd_on_pod(ct_pod, ceph_cmd)
 
     # For some commands, like "ceph fs ls", the returned output is a list
-    if isinstance(out, list):
-        return [item.toDict() for item in out if item]
-    return out.toDict()
+    if isinstance(out, Munch):
+        return out.toDict()
+    return yaml.safe_load(out)
```
jilju:
```
return yaml.safe_load(out)
```
Is yaml.safe_load() still required? Here, 'out' will be a list

keesturam: out is a string here.

ocs/ocp.py
clacroix12: Would it be better to always return a list than to have two different return types? Even if the list has one item, I think the consistency would be better than having any code that consumes this need to consider both cases.

RazTamir: I probably missing something.
This method is executing oc command so how 
```python
sh-4.2# rbd ls rbd --format json
```
is related to it?

vasukulkarni:
@keesturam can you address the comments here and also due to new changes, it needs rebased or done as new PR

vasukulkarni:
@keesturam please rebase and test.

vasukulkarni:
@keesturam ping?

keesturam:
@vasukulkarni sure. I should have this done in a couple of days.

RazTamir:
@keesturam still relevant?
This PR is really old

clacroix12:
Closing due to inactivity. Feel free to re-open / create a new PR to address these changes if necessary.

PR #147, opened 2019-06-03
OCS-383 and OCS-384 - Test to validate reclaim policy - 

keesturam:
Automates the following tests

OCS-383 - FT - OCP_Validate Retain policy is honored
OCS-384 - FT - OCP_Validate Delete policy is honored

### DIFF
--- a/ocs/defaults.py
+++ b/ocs/defaults.py
@@ -32,6 +32,7 @@ API_VERSION = "v1"
 CEPH_IMAGE = "ceph/ceph:v14.2.0-20190410"
 ROOK_IMAGE = "rook/ceph:master"
 DEPLOYMENT_PLATFORM = 'AWS'
+TEST_NAMESPACE = 'ocs-qe'
 
 # This section is suppose to be available just from ocsci/config.py module from
 # ENV_DATA dictionary. Once we drop support of old runner we will delete this
--- a/templates/RBD_PersistentVolumeClaim.yaml
+++ b/templates/RBD_PersistentVolumeClaim.yaml
@@ -1,0 +1,13 @@
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: {{pvc_name | default('test-rbd-device')}}
+  namespace: {{project_name | default('default')}}
+spec:
+  accessModes:
+    - {{access_mode | default("ReadWriteOnce")}}
+  resources:
+    requests:
+      storage: {{ pvc_size | default("10") }}Gi
+  storageClassName: {{sc_name | default("rook-ceph-block")}}
--- a/tests/test_reclaim_policy.py
+++ b/tests/test_reclaim_policy.py
@@ -1,0 +1,271 @@
+"""
+Automates the following tests
+
+OCS-383 - FT - OCP_Validate Retain policy is honored
+OCS-384 - FT - OCP_Validate Delete policy is honored
+
+"""
+
+import logging
+import yaml
+import os
+import pytest
+
+from ocs import ocp
+from utility import utils, templating
+from ocsci.config import ENV_DATA
+from ocsci.testlib import tier1, ManageTest
+import ocs.defaults as defaults
+
+log = logging.getLogger(__name__)
+PVC_YAML = os.path.join(
+    "templates", "RBD_PersistentVolumeClaim.yaml"
+)
+POD_YAML = os.path.join(
+    "templates", "RBD_NginxPod.yaml"
+)
+SC_YAML = os.path.join(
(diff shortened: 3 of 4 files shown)


tests/test_reclaim_policy.py
```diff
@@ -0,0 +1,271 @@
...
+    ct_pod = ocp.get_ceph_tools_pod()
+    rbd_cmd = f"rbd rm {pool_name}/{vol}"
+    return ocp_pod_obj.exec_cmd_on_pod(ct_pod, rbd_cmd)
+
+
+@tier1
+@pytest.mark.usefixtures(
+    test_fixture.__name__,
+)
+class TestReclaimPolicy(ManageTest):
```
ebenahar: @keesturam Nice usage of 2 cases within the same class!

tests/test_reclaim_policy.py
```diff
@@ -0,0 +1,271 @@
...
+
+def rbd_rm_volume(vol, pool_name='rbd'):
+    """
+
+    Args:
+        vol(str): rbd image name, which is also PV name
+        pool_name(str): ceph pool used. Defaults to rbd
+
+    Returns:
+        None if the deletion is successful or
```
ebenahar: In that case, it would be better to raise an exception (assertionError for example) in case the returned value from exe_cmd_on_pod is a command failure one.
And in case of success, the function should be void

tests/test_reclaim_policy.py
```diff
@@ -0,0 +1,271 @@
...
+
+def create_pvc(**kwargs):
+    """
+    Creates a Persistent Volume Claim
+    Args:
+        **kwargs:
+
+    Returns:
+        (bool): True if creation is successful, False otherwise
+
```
ebenahar: delete empty line.
Apply to all relevant places

tests/test_reclaim_policy.py
```diff
@@ -0,0 +1,271 @@
...
+    """
+
+    log.info(f'deleting storage class: {sc_name}')
+    return SC.delete(resource_name=sc_name)
+
+
+def create_pvc(**kwargs):
+    """
+    Creates a Persistent Volume Claim
+    Args:
```
ebenahar: add one empty line above Args. 
Apply to all relevant places

ocs/defaults.py
```diff
@@ -32,6 +32,7 @@
 CEPH_IMAGE = "ceph/ceph:v14.2.0-20190410"
 ROOK_IMAGE = "rook/ceph:master"
 DEPLOYMENT_PLATFORM = 'AWS'
+TEST_NAMESPACE = 'ocs-qe'
```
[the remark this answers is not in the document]
mbukatov: I would say it's ok to keep it here. We don't have a guideline about namespace usage in test code (eg. I would like to see namespace fixture with default function scope), but it's clear that tests should create k8s objects in custom test namespace.

templates/RBD_PersistentVolumeClaim.yaml
```diff
@@ -0,0 +1,13 @@
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: {{pvc_name | default('test-rbd-device')}}
```
vasukulkarni: templating part should not be used any more, please use new method.

templates/RBD_PersistentVolumeClaim.yaml
```diff
@@ -0,0 +1,13 @@
...
+  namespace: {{project_name | default('default')}}
+spec:
+  accessModes:
+    - {{access_mode | default("ReadWriteOnce")}}
```
vasukulkarni: same

ocs/defaults.py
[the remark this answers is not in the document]
clacroix12: @mbukatov that makes sense. If multiple tests are utilizing this and potentially overriding the value which is propagated to other tests then I can see it being here. I just want to avoid having simple variables get put here when they really belong inside the test script themselves.

tests/test_reclaim_policy.py
```diff
@@ -0,0 +1,271 @@
...
+
+    Returns:
+        (bool): True if creation is successful, False otherwise
+
+    """
+
+    file_sc = templating.generate_yaml_from_jinja2_template_with_data(
+        SC_YAML, **kwargs
+    )
+    with open(TEMP_SC_YAML_FILE, 'w') as yaml_file:
```
vasukulkarni: please use new method which uses dict instead of modifying yaml

keesturam:
closing this PR.
A new PR with the test cases using OCS class will be sent along with the changes suggested in this PR's comments.

PR #97, opened 2019-05-27
[WIP]: Adding Test case for OCS-269 - AllocateSizeMoreThanClusterSize

keesturam:
This PR introduces test case OCS-269 and template files for storageclass, application pod, persistent volume claim.

PR #98, opened 2019-05-27
OCS-331: Create CephFS Storage Class with invalid values

fbalak:
- Add `conftest.py` with fixture `invalid_cephfs_storageclass`.
- Use `tmpdir` fixture for yaml files created from templates.
- Add `invalid_cephfs_storageclass` test case.

PR #148, opened 2019-06-03
Adding test for OCS-269

keesturam:
Test for OCS-269. Fixes all comments for #97

### DIFF
--- a/ocs/defaults.py
+++ b/ocs/defaults.py
@@ -32,6 +32,8 @@ API_VERSION = "v1"
 CEPH_IMAGE = "ceph/ceph:v14.2.0-20190410"
 ROOK_IMAGE = "rook/ceph:master"
 DEPLOYMENT_PLATFORM = 'AWS'
+TEST_NAMESPACE = 'ocs-qe'
+GB = 1024 ** 3
 
 # This section is suppose to be available just from ocsci/config.py module from
 # ENV_DATA dictionary. Once we drop support of old runner we will delete this
@@ -48,3 +50,4 @@ STORAGE_API_VERSION = 'storage.k8s.io/v1'
 ROOK_API_VERSION = 'ceph.rook.io/v1'
 OCP_API_VERSION = 'project.openshift.io/v1'
 OPENSHIFT_REST_CLIENT_API_VERSION = 'v1'
+
--- a/ocs/defaults.py
+++ b/ocs/defaults.py
@@ -50,4 +50,3 @@ STORAGE_API_VERSION = 'storage.k8s.io/v1'
 ROOK_API_VERSION = 'ceph.rook.io/v1'
 OCP_API_VERSION = 'project.openshift.io/v1'
 OPENSHIFT_REST_CLIENT_API_VERSION = 'v1'
-
--- a/tests/test_AllocateSizeMoreThanClusterSize.py
+++ b/tests/test_AllocateSizeMoreThanClusterSize.py
@@ -1,0 +1,332 @@
+"""
+Automates the following test
+
+OCS-269 - FT-OCP-Create-PV-AllocateSizeMoreThanClusterSize
+Verify a PVC creation by allocating more storage
+than what is available in Ceph
+"""
+
+import logging
+import yaml
+import os
+import pytest
+import ocs.defaults as defaults
+
+from ocs import ocp
+from utility import utils, templating
+from utility.utils import run_cmd
+from ocsci.testlib import tier1, ManageTest
+
+log = logging.getLogger(__name__)
+
+PVC_YAML = os.path.join(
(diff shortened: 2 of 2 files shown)


tests/test_AllocateSizeMoreThanClusterSize.py
```diff
@@ -0,0 +1,332 @@
...
+NAMESPACE = ocp.OCP(
+    kind='Project', namespace=defaults.TEST_NAMESPACE
+)
+OCP = ocp.OCP(
+    kind='Pod', namespace=defaults.TEST_NAMESPACE
+)
+
+
+@pytest.fixture(scope='class')
+def test_fixture(request):
```
fbalak: This fixture should have name representing the resource it provides.

tests/test_AllocateSizeMoreThanClusterSize.py
```diff
@@ -0,0 +1,332 @@
...
+    """
+    Over allocate a PVC above cluster's overall capacity
+    """
+    self = request.node.cls
+
+    def finalizer():
+        teardown(self)
+    request.addfinalizer(finalizer)
```
fbalak: I think that `request.addfinalizer(teardown)` is enough and there doesn't need to be another layer (`finalizer` method).

tests/test_AllocateSizeMoreThanClusterSize.py
```diff
@@ -0,0 +1,332 @@
...
+)
+POD_YAML = os.path.join(
+    "templates", "RBD_NginxPod.yaml"
+)
+SC_YAML = os.path.join(
+    "templates", "RBD_StorageClass.yaml"
+)
+
+
+TEMP_PVC_YAML_FILE = '/tmp/pvc_test.yaml'
```
fbalak: I suggest to use `tmpdir` builtin fixture creating temporary files. Example in https://github.com/red-hat-storage/ocs-ci/pull/98/files#diff-484462fced51d1a06b1d93b4a44dd535R31

tests/test_AllocateSizeMoreThanClusterSize.py
```diff
@@ -0,0 +1,332 @@
...
+    Returns:
+        (bool): True if creation is successful, False otherwise
+
+    """
+
+    sc_name = (kwargs['sc_name'])
+    sc_get = SC.get()
+    storage_classes = [item['metadata']['name'] for item in sc_get['items']]
+    log.info(f'checking if {sc_name} exists already')
+    if sc_name in storage_classes:
```
ebenahar: why is it needed?

tests/test_AllocateSizeMoreThanClusterSize.py
mbukatov: @fbalak note that `teardown` is called with `self` argument, and so this layer needs to be there as long as passing reference to `self` into `finalizer` function is intended. You could make it shorter using lambda, but that's about it :) I don't like this kind of code organization either, but I don't think it's a good idea to block this PR because of this, esp. when lot of other people got similar code merged. I would rather draw a line at some point when we get an agreement on this, change this at once and then start to enforce it during reviews.

ocs/defaults.py
```diff
@@ -32,6 +32,8 @@
 CEPH_IMAGE = "ceph/ceph:v14.2.0-20190410"
 ROOK_IMAGE = "rook/ceph:master"
 DEPLOYMENT_PLATFORM = 'AWS'
+TEST_NAMESPACE = 'ocs-qe'
```
clacroix12: Does `TEST_NAMESPACE` really belong in `defaults.py`? This seems like it should just be a local variable in whatever script needs it.

ocs/defaults.py
```diff
@@ -32,6 +32,8 @@
...
+GB = 1024 ** 3
```
clacroix12: This should be in `constants.py` instead of `defaults.py`

ocs/defaults.py
clacroix12: Nevermind this, reasoning was addressed in #147

tests/test_AllocateSizeMoreThanClusterSize.py
fbalak: Ah, ok.

tests/test_AllocateSizeMoreThanClusterSize.py
fbalak: This is no longer relevant as temporary yaml files are handled by `ocs-ci` classes.

ocs/defaults.py
fbalak: What does this default variable represent?

keesturam:
closing this PR.
A new PR with the test cases using OCS class will be sent along with the changes suggested in this PR's comments.

PR #132, opened 2019-05-30
Helper functions

ebenahar:

PR #154, opened 2019-06-05
Add resources

ebenahar:


resources/ceph_block_pool.py
```diff
@@ -0,0 +1,51 @@
...
+            self.cbp_data.get('apiVersion'), self.cbp_data.get('kind'),
+            self.cbp_data.get('metadata').get('namespace')
+        )
+        self._name = self.cbp_data.get('metadata').get('name')
+
+    @property
+    def name(self):
+        return self._name
+
+    def reload(self):
```
fbalak: Would it make sense to move methods `reload()`, `get()` and `delete()` (and probably `create()`) to `BaseOCSClass` so it can be accessed by all children of the class?

resources/ceph_block_pool.py
```diff
@@ -0,0 +1,51 @@
...
+    Provides basic methods to operate on CephBlockPool kind.
+    """
+    def __init__(self, **kwargs):
+        """
+        Initializer method
+
+        kwargs:
+            Copy of ocs/defaults.py::CEPHBLOCKPOOL_DICT dictionary
+        """
+        self.cbp_data = defaults.CEPHBLOCKPOOL_DICT
```
RazTamir: copy()

dahorak:
Sorry if I miss something, but why is the template file `templates/ocs-deployment/cephblockpool_new.yam` empty?

resources/ceph_block_pool.py
RazTamir: not all of them but probably delete and get can be moved.
Create - every class will implement differently the creation of the resource based on several things like the ceph interface, etc
reload - probably can move as well but as there is no actual usage of it, yey, we can't tell

resources/ceph_block_pool.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @property
+    def name(self):
+        return self._name
+
+    def reload(self):
+        self.cbp_data = load_yaml_to_dict(self.temp_yaml.name)
+
+    def get(self):
+        return self.ocp.get()
```
RazTamir: this will return all cephblockpools in the cluster.
maybe you want to send return self.ocp.get(resource_name=self.name)

resources/ceph_block_pool.py
ebenahar: @fbalak Yes, I agree, will do that

resources/ceph_block_pool.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def create(self):
+        log.info(f"Creating a Ceph block pool {self.name}")
+        with open(self.temp_yaml.name, 'w') as yaml_file:
```
RazTamir: maybe create a function for this:
```python
def dump_dict_to_temp_yaml(data, temp_yaml):
    with open(temp_yaml, 'w') as yaml_file:
    return yaml.dump(data, yaml_file)
```

you can add it in templating file

resources/ceph_block_pool.py
ebenahar: @RazTamir Actually, it would be better if we do the copy outside, like done here - https://github.com/red-hat-storage/ocs-ci/pull/132/files

I will do this here too

resources/ceph_block_pool.py
ebenahar: done

resources/ceph_block_pool.py
RazTamir: I'm not sure which part I should look for but in the next line, you are changing the original dict

resources/ceph_block_pool.py
ebenahar: @dahorak @RazTamir for now everything is generic so moved all the base class, see next commit

resources/ceph_block_pool.py
ebenahar: done

resources/ceph_block_pool.py
ebenahar: @RazTamir I think that in the case of storageclass, we can do the distinction between block and file outside the object creation method. For example, in the helpers function of storageclass creation, we'll send a copy of defaults.STORAGECLASSCEFFS_DICT which will contain the relevant keys for storageclss of ceph fs interface type

ebenahar:
@dahorak regarding cepfs yaml, I would like to wait for it to be added as part of a separate PR

resources/ceph_block_pool.py
RazTamir: So why do we need the kind classes if we are going to use functionality from helpers?

ebenahar: @RazTamir for future implementation for methods which will be relevant for specific kinds. Not sure if there will be any though

RazTamir: I'm not sure this is a good direction as we wanted to simplify the usage of kinds, with that change we just wrapped the basic OCP methods to enable also yaml dump/load - that's it.

resources/ceph_file_system.py
```diff
@@ -14,49 +12,16 @@
...
-
-    def delete(self):
-        self.ocp.delete(resource_name=self.name)
-
-    def apply(self, **data):
-        with open(self.temp_yaml.name, 'w') as yaml_file:
-            yaml.dump(data, yaml_file)
-        assert self.ocp.apply(yaml_file=self.temp_yaml.name)
+        self.data = kwargs
+        super(CephFileSystem, self).__init__(**kwargs)
```
vasukulkarni: Dont we still need the Kind, Metadata and namespace to be static here for each kind?

vasukulkarni:
Looks like the Base OCS class does all create/apply/delete now which looks simplified and probably somethings can be moved there, but wondering what can be specific for each kind :)

RazTamir:
Great observation Vasu.
After @filipbalak comment https://github.com/red-hat-storage/ocs-ci/pull/154#discussion_r290720586 we realized we are going to need only those simple oc methods.

So not really need to have all these classes (:

ebenahar:
@dahorak @vasukulkarni @RazTamir Thanks guys, closing this PR and we'll continue in https://github.com/red-hat-storage/ocs-ci/pull/132

vasukulkarni:
@RazTamir @ebenahar I think we can still keep the FileSystem, StorageClass etc with different "kind" and  "namespace" and later we find something specific to those Kinds, the methods can be implemented there (other than create, delete, apply etc)

ebenahar:
@vasukulkarni wouldn't it be redundant to keep these classes which don't do anything just for a possibility that in the future we might need them? 
I think that if we'll need to implement special things for certain kinds, we can do the change quite easily. This is better than maintaining an unsued code IMO

RazTamir:
+1 on Elad's commet.
I just finished review all the 13 PRs for new test cases and I can't see anything that is specific to a kind. 

Let's create the kind classes when needed. 
My 2 cents

vasukulkarni:
> @vasukulkarni wouldn't it be redundant to keep these classes which don't do anything just for a possibility that in the future we might need them?
> I think that if we'll need to implement special things for certain kinds, we can do the change quite easily. This is better than maintaining an unsued code IMO

works for me, if the 13 PRs dont need any specific operation at this stage then why worry now!, but we shoudn't create lot of helper functions to deal with kinds and at that stage they are candidate for specific kind methods.

PR #159, opened 2019-06-06
Implementing logging per test case.

vavuthu:
Implementing logging per test case.

 * creates the directory structure for logging
 * creates symlink for easy access to logs

Example for symlink:
```
[vavuthu@localhost tmp]$ ls -lrt | tail -2
drwxrwxr-x.   3 vavuthu vavuthu     60 Jun  6 15:52 logs_1559816529
lrwxrwxrwx.   1 vavuthu vavuthu     20 Jun  6 15:52 logs -> /tmp/logs_1559816529
[vavuthu@localhost tmp]$ 
```

Example for directory structure:
lets say  tests/manage/test_manage.py conatins class `TestManage` having 2 test cases. 
o/p: 
[vavuthu@localhost tmp]$ ls logs/tests/manage/test_manage/TestManage/
test_manage_test_case1.log  test_manage_test_case2.log
[vavuthu@localhost tmp]$ 

OVERALL logging structure:
======================
```
[vavuthu@localhost tmp]$ tree logs
logs
└── tests
    ├── ecosystem
    │   ├── deployment
    │   │   └── test_ocs_basic_install
    │   │       └── TestDeployment
    │   │           └── test_deployment.log
    │   └── destroy
    │       └── test_destroy_cluster
    │           └── TestDestroy
    │               └── test_destroy_cluster.log
    ├── manage
    │   └── test_manage
    │       └── TestManage
    │           ├── test_manage_test_case1.log
    │           └── test_manage_test_case2.log
    └── test_pv_creation
        └── TestPvCreation
            └── test_pv_creation.log

13 directories, 5 files
[vavuthu@localhost tmp]$ 
```

### DIFF
--- a/ocsci/pytest_customization/ocsci_logging.py
+++ b/ocsci/pytest_customization/ocsci_logging.py
@@ -10,9 +10,9 @@ FORMATTER = logging.Formatter(
     " - %(message)s"
 )
 ep_time = int(time.time())
-log_dir = "/logs_" + str(ep_time)
-log_dir_path = ocsci_config.RUN['log_dir'] + log_dir
-sym_link = ocsci_config.RUN['log_dir'] + "/logs"
+log_dir = f"logs_{ep_time}"
+log_dir_path = os.path.join(ocsci_config.RUN['log_dir'], log_dir)
+sym_link = os.path.join(ocsci_config.RUN['log_dir'], "logs")
 
 
 def create_directory_path(path):
@@ -43,7 +43,7 @@ def pytest_runtest_setup(item):
     log_structure = item.nodeid.split("::")[0][:-3]
     test_class_name = item.nodeid.split("::")[1]
     log_path = os.path.join(log_dir_path, log_structure, test_class_name)
-    test_log = item.name + ".log"
+    test_log = f"{item.name}.log"
     log_file = os.path.join(log_path, test_log)
 
     if not os.path.exists(log_path):
--- a/ocsci/pytest_customization/ocsci_logging.py
+++ b/ocsci/pytest_customization/ocsci_logging.py
@@ -1,0 +1,65 @@
+import logging
+import os
+import time
+import pytest
+
+from ocsci import config as ocsci_config
+
+FORMATTER = logging.Formatter(
+    "%(asctime)s - %(levelname)s - %(name)s.%(funcName)s.%(lineno)d"
+    " - %(message)s"
+)
+ep_time = int(time.time())
+log_dir = "/logs_" + str(ep_time)
+log_dir_path = ocsci_config.RUN['log_dir'] + log_dir
+sym_link = ocsci_config.RUN['log_dir'] + "/logs"
+
+
+def create_directory_path(path):
+    """
+    Creates directory if path doesn't exists
+    """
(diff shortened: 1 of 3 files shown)


ocsci/pytest_customization/ocsci_logging.py
```diff
@@ -0,0 +1,65 @@
...
+import pytest
+
+from ocsci import config as ocsci_config
+
+FORMATTER = logging.Formatter(
+    "%(asctime)s - %(levelname)s - %(name)s.%(funcName)s.%(lineno)d"
+    " - %(message)s"
+)
+ep_time = int(time.time())
+log_dir = "/logs_" + str(ep_time)
```
clacroix12: Use fstring instead of concat and casting. `log_dir = f"/logs_{ep_time}"`

ocsci/pytest_customization/ocsci_logging.py
```diff
@@ -0,0 +1,65 @@
...
+log_dir_path = ocsci_config.RUN['log_dir'] + log_dir
```
clacroix12: Use `os.path.join(ocsci_config.RUN['log_dir'], log_dir)` instead of concat here and on the next line.

ocsci/pytest_customization/ocsci_logging.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def pytest_runtest_setup(item):
+    """
+    Adding unique log handler for each test
+    """
+    log_structure = item.nodeid.split("::")[0][:-3]
+    test_class_name = item.nodeid.split("::")[1]
+    log_path = os.path.join(log_dir_path, log_structure, test_class_name)
+    test_log = item.name + ".log"
```
clacroix12: Use fstring instead of concat. `test_log = f"{item.name}.log"`

ocsci/pytest_customization/ocsci_logging.py
vasukulkarni: It is better to log in the upper log_dir then in the tree structure, makes it easier to browse when things fail.

ocsci/pytest_customization/ocsci_logging.py
vavuthu: Done

ocsci/pytest_customization/ocsci_logging.py
vavuthu: > Use `os.path.join(ocsci_config.RUN['log_dir'], log_dir)` instead of concat here and on the next line.

Done

vavuthu: > It is better to log in the upper log_dir then in the tree structure, makes it easier to browse when things fail.

Even initially I thought the same but when our test cases grows, and if 2 test cases has same test case id, then logging in the single directory create problems of overriding the logs. Even though if we append number at the end of file, its difficult for users to differentiate between 2 identical test cases. 

Traversing through directory structure of logs is easy because user knows where the test case resides.

Please share your thoughts

vasukulkarni: Sure, Lets this one goes!

PR #165, opened 2019-06-06
Enabling unit testing, plus a config/defaults test

zmc:
* Make it possible to run unit tests with just `py.test`
* Don't break the functionality of `run-ci`
* Colocate the pytest plugin tests with the plugins themselves
* Add a basic test that compares an actual default config with what the defaults module specifies


tox.ini
```diff
@@ -6,7 +6,7 @@ deps =
     -rrequirements.txt
 setenv =
  PYTHONPATH=.
-commands = {envpython} {envbindir}/run-ci --ignore=tests -c unittests/pytest.ini {posargs:unittests}
+commands = py.test --ignore=tests -c unittests/pytest.ini {posargs:mita oc ocs resources unittests utility ocsci}
```
vasukulkarni: This was intentionally run-ci (due to a comment in old pr) to test the entry-point.

zmc: That makes sense - however, if we move the logic from the entry point into one or more of the packages themselves, we can simply unit test that logic.

vasukulkarni: Mind dropping this one until we have something more in unittests? :)

vasukulkarni:
Overall LGTM

tox.ini
zmc: Sure - I just repushed.

I'm ready to write more unit tests, but there's actually an issue that needs addressing first. Because pytest runs all tests in the same process, modifications to `ocsci.config` will persist between tests. So we need a way to reset the configuration. The current "module as a container for configuration" approach makes that really tough.

I have a decent implementation ready that could fix this using a singleton, but I don't want to tack it onto this PR. We should discuss that soon!

vasukulkarni:
@clacroix12 @petr-balogh FYI

ocsci/pytest_customization/tests/test_pytest.py
```diff
@@ -18,6 +18,9 @@ def test_help_message(testdir):
     Check that ``py.test --help`` output lists custom options of
     ocscilib pytest plugin.
     """
+    testdir.makeconftest(textwrap.dedent("""
+        pytest_plugins = ['ocsci.pytest_customization.ocscilib']
+    """))
```
mbukatov: My initial feedback here is that this should not be necessary, as the pytest is executed within the same process, so this should not be necessary. Has it any effect?

setup.py
```diff
@@ -33,10 +33,6 @@
         'console_scripts': [
             'run-ci=run_ocsci:main',
         ],
-        'pytest11': [
-            'ocscilib = ocsci.pytest_customization.ocscilib',
-            'marks = ocsci.pytest_customization.marks',
-        ]
```
mbukatov: How did you register these modules for normal test run (not unittests)?

mbukatov: I'm ok with dropping this from setup, but we need to add it into global conftest (which then would need to be shielded from unittests) or in other appropriate place.

ocsci/pytest_customization/tests/test_pytest.py
zmc: Executing unit tests inside of `run-ci` has side-effects that are not always desirable for unit tests; I think unit tests should be run using `py.test` directly - and with this PR, that does work - though @vasukulkarni asked that I not change `tox.ini` just yet.

setup.py
zmc: See https://github.com/red-hat-storage/ocs-ci/pull/165/files#diff-52ce28eee4cc4cda242e445bb4ac28d1R113

zmc:
Leaving unit tests in a separate directory won't save us from having to update them when we move the code under test. I'd be happy to rebase and update docs - I hadn't seen those before your comment.

setup.py
vasukulkarni: This should be resolved I guess

vasukulkarni:
@mbukatov Is there anything blocking in current structure for this PR to be merged?

zmc:
Just want to note that I have a follow-up PR in progress that does a little more restructuring and will include more unit tests :)

andrewschoen:
> I would rather not move unittest into python modules until we stabilize the structure of the repository. This is because we could need to tweak unittests every time we move something.

IMHO, I think it's reasonable to have the testing code for a module exist in the same module. If you're using tox to initiate the tests then the invocation shouldn't change regardless of the test location anyway.

vasukulkarni:
@zmc needs rebase, sorry dont know what pr changed caused this

zmc:
@vasukulkarni it was https://github.com/red-hat-storage/ocs-ci/pull/159.

zmc:
@vasukulkarni Rebased.

mbukatov:
> > I would rather not move unittest into python modules until we stabilize the structure of the repository. This is because we could need to tweak unittests every time we move something.
> 
> IMHO, I think it's reasonable to have the testing code for a module exist in the same module. If you're using tox to initiate the tests then the invocation shouldn't change regardless of the test location anyway.

I'm not against having unit tests next to the python modules themselves. I was raising few issues I saw with the original pull request.

ocsci/pytest_customization/tests/test_pytest.py
mbukatov: This doesn't answer my question.

ocs/defaults.py
```diff
@@ -71,6 +70,25 @@
     'rook_image': ROOK_IMAGE,
 }
 
+DEPLOYMENT = {
```
petr-balogh: @zmc  Didn't get why those section were moved here? They are not used from the defaults in any places from the code but from ocsci/config.py.

My intention was to remove a lot of variable from here and use them from ocsci/config.py directly which are loaded from the default_config.yaml as I mentioned in the comment in this file. But as we still had some support for old runner I didn't do this cleanup yet and wanted to do it in the PR once I am back from PTO.

This is comment you can see in this file.
```python
# Those variables below are duplicate at the moment from default_config.yaml
# and once we drop support for old runner we will remove those variables from
# here and will have them defined only on one place.

# Be aware that variables defined above and below are not used anywhere in the
# config files and their sections when we rendering config!
```

Issue #161, opened 2019-06-06
Create a constants file


ebenahar:
Move all constants from defaults.py to there

davidwilson3:
I'd like to take this if it's still up for grabs

clacroix12:
Addressed in #166

PR #164, opened 2019-06-06
Fix template directory structure and test references

vasukulkarni:
Signed-off-by: Vasu Kulkarni <vasu@redhat.com>

PR #166, opened 2019-06-06
Constants module

clacroix12:
Created constants module. Moved constant vars from defaults and kinds modules to new module. #161

### DIFF
--- a/docs/writing_tests.md
+++ b/docs/writing_tests.md
@@ -18,11 +18,31 @@ can import directly from `ocsci.testlib` module with this statement:
 `from ocsci.testlib import tier1` for example.
 
 
-## Base test calsses for teams
+## Base test classes for teams
 
 Those are located in [testlib.py](../ocsci/testlib.py) which you can also
 import from `ocsci.testlib` module with statement:
 `from ocsci.testlib import manage` which is base test class for manage team.
+
+
+## Constants and Defaults
+
+Many of our tests utilize defaults and constants. These are both defined in
+`ocs/constants.py` and `ocs/defaults.py` respectively. Constants and defaults
+are fairly similar but functionally different which is why we have chosen
+to separate them into their own modules.
+
+If your test requires one of these you can easily import it.
+If you intend to implement a new one (generally if more than one test will
+utilize it), please consider whether or not that value might change between
+different test executions. If it's something like a filepath (unchanging),
+it's probably a constant. If tests may overwrite the value, it's most likely a
+default.
+
+Note these modules are not intended to be a dumping ground for any variable
+your test might need. These are designed to be homes for widely used variables
+that need to be consistent across a test execution. You can learn more from
+viewing the existing constants and defaults in their respective modules.
 
 
 ## Other notes
--- a/ocs/constants.py
+++ b/ocs/constants.py
@@ -1,0 +1,32 @@
+"""
+Constants module.
+
+This module contains any values that are widely used across the framework,
+utilities, or tests that will predominantly remain unchanged.
+
+In the event values here have to be changed it should be under careful review
+and with consideration of the entire project.
+
+"""
+import os
+
+# Directories
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+TOP_DIR = os.path.dirname(THIS_DIR)
+TEMPLATES_DIR = os.path.join(TOP_DIR, "templates")
+
+# Statuses
+STATUS_PENDING = 'Pending'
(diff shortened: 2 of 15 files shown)


vasukulkarni:
I know there is an issue, but do we really need Constants, defaults can serve that purpose :-) , Users have to know too many files to import, we could keep things simple.

clacroix12:
@vasukulkarni We thought it was important to differentiate between defaults and constants. Defaults are intended to be overridden if your test case or deployment calls for it, constants aren't. We also already had a `kinds.py` that this PR gets rid of in favour of using `constants.py` so it really isn't a change in the number of locations we have these sorts of variables.

vasukulkarni:
I am +1 for dropping kinds.py and I understand the subtle difference between two but we have favored  ease of use previously in libraries and other things so just for those couple barewords users have to know 2 different files to import and technically dir changes per system breaking the meaning of constant :)

ocs/defaults.py
```diff
@@ -14,22 +14,15 @@
 import os
 from getpass import getuser
 
+from ocs.constants import TEMPLATES_DIR
 from utility.templating import load_yaml_to_dict
 
-THIS_DIR = os.path.dirname(os.path.abspath(__file__))
-TOP_DIR = os.path.dirname(THIS_DIR)
-TEMPLATE_DIR = os.path.join(TOP_DIR, "templates/ocs-deployment/")
+DEPLOYMENT_TEMPLATE_DIR = os.path.join(TEMPLATES_DIR, "ocs-deployment")
```
vavuthu: Could you add below 2 so that most of test cases benifits:

TEMPLATE_DIR_RBD = os.path.join(TOP_DIR, "templates/CSI/rbd/")
TEMPLATE_DIR_CEPHFS = os.path.join(TOP_DIR, "templates/CSI/cephfs/")

clacroix12: Yeah I see that vasu has added a number of new constants in for those directories, if we decide to move forward with the constants module I will add those all in.

clacroix12:
@vasukulkarni

clacroix12:
> technically template_dir changes per system breaking the meaning of constant

@vasukulkarni I'm not sure what you mean by this. Sure the actual filepath will change based on where you clone there repo but it's a constant in the sense it always points you to the templates directory. 

I also don't think that having people import from constants is a huge ask. I would argue that having people figure out that our constants are actually in our defaults module is more confusing. I would prefer to have well defined modules and files as opposed to combining things that are functionally different to cut down on the number of imports people have to use.

ocs/constants.py
```diff
@@ -0,0 +1,32 @@
...
+
+In the event values here have to be changed it should be under careful review
+and with consideration of the entire project.
+
+"""
+import os
+
+# Directories
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+TOP_DIR = os.path.dirname(THIS_DIR)
```
zmc: I'd argue that THIS_DIR and TOP_DIR shouldn't be stored anywhere

zmc:
`TEMPLATES_DIR` may be confusing as a constant since it can differ across process lifetimes, but what makes it a constant is that *during* a process' lifetime, it won't change. Consider some valid JavaScript: that is ubiquitous `const now = new Date();`

ocs/constants.py
clacroix12: `THIS_DIR` I understand, but that was originally just used to make constructing `TOP_DIR` a little more readable. Why wouldn't we want to keep `TOP_DIR` though?

zmc: Oh, this is my mistake - blame it on a lack of coffee 🤦‍♂ 
Comment only stands for THIS_DIR

clacroix12: Totally agree, will remove `THIS_DIR`

vasukulkarni:
Based on other review comments, lets merge this after #164

docs/writing_tests.md
```diff
@@ -18,13 +18,33 @@ can import directly from `ocsci.testlib` module with this statement:
...
 
-## Base test calsses for teams
+## Base test classes for teams
 
 Those are located in [testlib.py](../ocsci/testlib.py) which you can also
 import from `ocsci.testlib` module with statement:
 `from ocsci.testlib import manage` which is base test class for manage team.
 
 
+## Constants and Defaults
```
RazTamir: @clacroix12 it will be great if you can send some update to ocs-qe about this change

clacroix12: @RazTamir will do

PR #172, opened 2019-06-07
Add  cluster_new.yaml, which is cluster.yaml without jinja templates

ebondare:
For my test case I need a file similar to cluster.yaml but without jinja templates, so I added cluster_new.yaml, as we have cephfilesystem.yaml and cephfilesystem_new.yaml.

### DIFF
--- a/templates/ocs-deployment/cluster_new.yaml
+++ b/templates/ocs-deployment/cluster_new.yaml
@@ -1,0 +1,130 @@
+#################################################################################################################
+# Define the settings for the rook-ceph cluster with common settings for a production cluster.
+# All nodes with available raw devices will be used for the Ceph cluster. At least three nodes are required
+# in this example. See the documentation for more details on storage settings available.
+
+# For example, to create the cluster:
+#   kubectl create -f common.yaml
+#   kubectl create -f operator.yaml
+#   kubectl create -f cluster.yaml
+#################################################################################################################
+
+apiVersion: ceph.rook.io/v1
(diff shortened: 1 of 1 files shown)


templates/ocs-deployment/cluster_new.yaml
```diff
@@ -0,0 +1,130 @@
+#################################################################################################################
```
ebenahar: @vasukulkarni @ebondare shouldn't we move this under CSI dir and call it cluster.py?

vasukulkarni: @ebondare how are you using this one? Are you modifying config and reloading without jinja? If so I think the best place is CSI/ folder

vasukulkarni: @ebondare ?

vasukulkarni: @petr-balogh we will have to use this in deployment only then its going to be useful?

ebondare: > @ebondare how are you using this one? Are you modifying config and reloading without jinja? If so I think the best place is CSI/ folder

As a matter of fact, the testcase I needed this file for was taken from me, so I don't need it anymore. I'll just close this pull request, right?

petr-balogh: Yep, please close, if we will start using files without templates in deployment we will update our file or update. Thanks

Issue #88, opened 2019-05-23
Come up with approach to execute deployment, test execution, cleanup in one run.


petr-balogh:
We can use [pytest ordering](https://pytest-ordering.readthedocs.io/en/develop/) plugin to make sure that deployment test is running as the first one test, and the destroy cluster will be run at the end of the execution.  We can specify execution for example like: `pytest -m "deployment or (tier1 and manage) or destroy"` to run deployment, then tier1 tests of manage team and destroy the cluster.

andrewschoen:
I think the deployment phase would be nicely done as a pytest fixture with a scope of session. This way the cluster can be created at the start of the test run and tore down when the tests complete. I've used this pattern in the past to create databases that are used during the test run and discarded afterwards. 

He's a link to some pytest docs about this: https://docs.pytest.org/en/latest/fixture.html#scope-sharing-a-fixture-instance-across-tests-in-a-class-module-or-session

This is is an example of how I've done this in the past for a shared test database: https://github.com/ceph/shaman/blob/master/shaman/tests/conftest.py#L101

I'd be happy to start the work here if we think it's an approach worth exploring.

vasukulkarni:
Even I think session based scope fits better,  but we also want this to be optional when using existing cluster.  probably its better to discuss in meeting how to fix the ordering part.

andrewschoen:
> Even I think session based scope fits better, but we also want this to be optional when using existing cluster. probably its better to discuss in meeting how to fix the ordering part.

I'd be happy to discuss this in a meeting. I think we'd handle using a existing cluster exactly how it's currently done. We can test to see if cluster information was given and if we can connect to it or not. https://github.com/red-hat-storage/ocs-ci/blob/master/tests/ecosystem/deployment/test_ocs_basic_install.py#L27-L33

PR #173, opened 2019-06-07
adds a pytest session fixture to automatically deploy a test cluster

andrewschoen:
This moves all the code from the deployment test into a pytest session fixture that is auto run each time tests are invoked. If given a path to an existing cluster the deployment will be skip. The cluster will also automatically destroyed after tests complete unless ``--skip-teardown`` is provided.

If we like this approach I can change docs and remove the existing deployment and destroy tests.

Fixes: #88


andrewschoen:
I tested this locally by running the following command:
```
run-ci -m "tier1 and manage" --ocsci-conf conf/config.yaml --cluster-conf conf/ocs_basic_install.yml --cluster-name andrewschoen-ocs-deployment --cluster-path ./clusters/test --skip-teardown
```

Then I ran it once again to make sure it'd reuse the cluster and then again without ``--skip-teardown`` to make sure the cluster would be destroyed.

ocsci/pytest_customization/ocscilib.py
```diff
@@ -42,6 +42,13 @@ def pytest_addoption(parser):
         dest='cluster_name',
         help="Name of cluster",
     )
+    parser.addoption(
+        '--skip-teardown',
+        dest='skip_teardown',
+        action="store_true",
+        default=False,
```
vasukulkarni: I would prefer True as default, Cluster setup is time consuming.

andrewschoen: I see that point, but I think we should change the flag to be ``--teardown`` and default it to ``False`` in that case though. Thoughts?

vasukulkarni: Agree on the flag switch, thanks!

clacroix12:
@andrewschoen are we able to only execute a deployment with this being a fixture or would we essentially need to create an empty test in order to spin up a cluster without running other tests?

andrewschoen:
> @andrewschoen are we able to only execute a deployment with this being a fixture or would we essentially need to create an empty test in order to spin up a cluster without running other tests?

I just added a ``test_cluster_is_running`` test that has the ``deployment`` mark. With that I'm able to add a ``-m "deployment"`` and deploy or teardown a cluster without running tests.

dahorak:
I understand the intention to move the deployment into fixture and from the pytest/terminology POV it makes sense, but...

1. I think, that cluster deployment and teardown should be considered as fully-fledged tests, with all the related reporting etc.
1. It should be easily possible to *just* deploy the cluster. This is possible via the `test_cluster_is_running`... but it is not immediately clear...
1. It should be easily and clearly possible to destroy existing cluster. Same as previous point, this is possible via the `test_cluster_is_running` test, but it is even more ambiguous, than the deployment (running `test_cluster_is_running` with mark `deployment` to destroy cluster... looks really strange).

I would propose to keep separation of the two tests for deployment and destroy as it is now - so it will be possible to run those tests directly, properly report results from them, etc. ...and additionally create similar fixture as proposed in this PR, which will allow us to run the particular tests without the necessity of dealing with the deployment and teardown (this will probably require to move some parts of the tests code into common functions, so it will be reusable both in the tests and in the fixture, without duplicating the code).

andrewschoen:
> I understand the intention to move the deployment into fixture and from the pytest/terminology POV it makes sense, but...
> 
> 1. I think, that cluster deployment and teardown should be considered as fully-fledged tests, with all the related reporting etc.

Nothing has changed from the existing tests here, the only thing that's different is how and when they're executed. If failures occur during deployment or teardown they'll be treated the same as a failing test.

> 2. It should be easily possible to _just_ deploy the cluster. This is possible via the `test_cluster_is_running`... but it is not immediately clear...

The ``test_cluster_is_running`` is not deploying the cluster, it is testing if the cluster is available after deployment. For the session fixture to run at least one test needs to be selected. I'd recommend keeping deployment separate from validating that the deployment produced a viable cluster. We could have a suite that's selected when we want to deploy that does a variety of checks on a cluster to ensure the install was correct.

> 3. It should be easily and clearly possible to destroy existing cluster. Same as previous point, this is possible via the `test_cluster_is_running` test, but it is even more ambiguous, than the deployment (running `test_cluster_is_running` with mark `deployment` to destroy cluster... looks really strange).
> 

The way to destroy a cluster is to use the ``--teardown`` flag. If the deployment mark is confusing we could reuse ``destroy`` and have a different test it collects, so it would skip things like cluster verification which doesn't make sense when destroying a cluster. I still consider the teardown of a cluster a "deployment" sort of action though.

> I would propose to keep separation of the two tests for deployment and destroy as it is now - so it will be possible to run those tests directly, properly report results from them, etc. 

I'm unsure what reporting we're missing by deployment being a session fixture, could you please explain?

> ...and additionally create similar fixture as proposed in this PR, which will allow us to run the particular tests without the necessity of dealing with the deployment and teardown 

This is already possible. If ``--cluster-path`` is used and the cluster can be accessed then deployment is skipped and testing against the existing cluster will proceed. Teardown will **only** happen if the ``--teardown`` flag is also provided though.

dahorak:
> > I understand the intention to move the deployment into fixture and from the pytest/terminology POV it makes sense, but...
> > 
> > 1. I think, that cluster deployment and teardown should be considered as fully-fledged tests, with all the related reporting etc.
> 
> Nothing has changed from the existing tests here, the only thing that's different is how and when they're executed. If failures occur during deployment or teardown they'll be treated the same as a failing test.

If I'm not mistaken, when failure occur in the fixture (deployment or teardown), it will be reported as failure (or maybe error) of the particular test for which the fixture was used, not as failure of "deployment test" (not sure, how exactly it works for fixtures with scope wider than only to one test though).

> 
> > 1. It should be easily possible to _just_ deploy the cluster. This is possible via the `test_cluster_is_running`... but it is not immediately clear...
> 
> The `test_cluster_is_running` is not deploying the cluster, it is testing if the cluster is available after deployment. For the session fixture to run at least one test needs to be selected. I'd recommend keeping deployment separate from validating that the deployment produced a viable cluster. We could have a suite that's selected when we want to deploy that does a variety of checks on a cluster to ensure the install was correct.

I understand that, `test_cluster_is_running` test is not deploying the cluster... but if you want to deploy the cluster, you need to run this (or similar) test... Not directly some "deploy cluster" test.

> 
> > 1. It should be easily and clearly possible to destroy existing cluster. Same as previous point, this is possible via the `test_cluster_is_running` test, but it is even more ambiguous, than the deployment (running `test_cluster_is_running` with mark `deployment` to destroy cluster... looks really strange).
> 
> The way to destroy a cluster is to use the `--teardown` flag. If the deployment mark is confusing we could reuse `destroy` and have a different test it collects, so it would skip things like cluster verification which doesn't make sense when destroying a cluster. I still consider the teardown of a cluster a "deployment" sort of action though.

Yep, running cluster verification before destroying cluster doesn't make much sense, but even if we will use another test, which will not run the verification before the teardown, what would happen, if it will be executed against non existing cluster? If I understand it well, it will firstly create it and then destroy, which seems quite weird.

> 
> > I would propose to keep separation of the two tests for deployment and destroy as it is now - so it will be possible to run those tests directly, properly report results from them, etc.
> 
> I'm unsure what reporting we're missing by deployment being a session fixture, could you please explain?

Reporting installation (and prospectively teardown) as separated test case. As I've mentioned above, if I understand it correctly, failure in fixture is reported to the particular test for which the fixture was used, not as separated test case.

> 
> > ...and additionally create similar fixture as proposed in this PR, which will allow us to run the particular tests without the necessity of dealing with the deployment and teardown
> 
> This is already possible. If `--cluster-path` is used and the cluster can be accessed then deployment is skipped and testing against the existing cluster will proceed. Teardown will **only** happen if the `--teardown` flag is also provided though.

Yes, from the functional point of view, it is the same what is covered in this PR, but if we would like to have the main functionality in regular tests, then the code should be probably slightly restructured (e.g. to avoid code duplication).

andrewschoen:
> If I'm not mistaken, when failure occur in the fixture (deployment or teardown), it will be reported as failure (or maybe error) of the particular test for which the fixture was used, not as failure of "deployment test" (not sure, how exactly it works for fixtures with scope wider than only to one test though).

The failure will be reported in a ``live log setup`` section, or was when I tested locally. A session scoped fixture is only ran once for the entire test run so if the deployment fails then no actual tests will run so I'm not sure how a deployment failure could be confused here. A fixture with a scope of method would fail for each individual test that uses it though.

> I understand that, `test_cluster_is_running` test is not deploying the cluster... but if you want to deploy the cluster, you need to run this (or similar) test... Not directly some "deploy cluster" test.

The idea is that there is no "deploy cluster" test anymore. Deploying a test cluster is just part of running any set of tests and is done automatically if a testing cluster is not already deployed and provided.

> Yep, running cluster verification before destroying cluster doesn't make much sense, but even if we will use another test, which will not run the verification before the teardown, what would happen, if it will be executed against non existing cluster? If I understand it well, it will firstly create it and then destroy, which seems quite weird.

Why would you want to teardown a non existing cluster? If you run the ``run-ci`` command with the ``--teardown`` flag and you do not provide a cluster one will be spun up for testing, used for testing and then destroyed.

> Reporting installation (and prospectively teardown) as separated test case. As I've mentioned above, if I understand it correctly, failure in fixture is reported to the particular test for which the fixture was used, not as separated test case.

Because this is a ``session`` scoped fixture it's going to operate a little differently here. You will see the error in ``setup`` section of the logs, not tied to each individual test like a ``method`` scoped fixture would.

> Yes, from the functional point of view, it is the same what is covered in this PR, but if we would like to have the main functionality in regular tests, then the code should be probably slightly restructured (e.g. to avoid code duplication).

I guess I'm not fully understanding why we'd want this in regular tests, could you please explain the benefit you see of this deployment and teardown code executing in a test method as opposed to a session-scoped fixture? It seems the same thing is being tested as a session-scoped fixture to me.

I I force a failure I see something like this from the session fixture.

```
================================================================================ test session starts ================================================================================
platform darwin -- Python 3.7.3, pytest-4.6.2, py-1.8.0, pluggy-0.12.0 -- /Users/andrewschoen/dev/ocs-ci/venv/bin/python
cachedir: .pytest_cache
rootdir: /Users/andrewschoen/dev/ocs-ci, inifile: pytest.ini
plugins: ocs-ci-0.1, reportportal-1.0.5
collected 10 items / 9 deselected / 1 selected

tests/ecosystem/deployment/test_ocs_basic_install.py::test_cluster_is_running
---------------------------------------------------------------------------------- live log setup -----------------------------------------------------------------------------------
13:43:41 - MainThread - tests.conftest - INFO - Running OCS basic installation
ERROR                                                                                                                                                                         [100%]

====================================================================================== ERRORS =======================================================================================
_____________________________________________________________________ ERROR at setup of test_cluster_is_running _____________________________________________________________________

request = <SubRequest 'cluster' for <Function test_cluster_is_running>>

    @pytest.fixture(scope="session", autouse=True)
    def cluster(request):
        log.info("Running OCS basic installation")
        cluster_path = ENV_DATA['cluster_path']
>       assert 0, "Failing the deployment simulation..."
E       AssertionError: Failing the deployment simulation...
E       assert 0

tests/conftest.py:32: AssertionError
-------------------------------------------------------------------------------- Captured log setup ---------------------------------------------------------------------------------
13:43:41 - MainThread - tests.conftest - INFO - Running OCS basic installation
```

tests/conftest.py
```diff
@@ -0,0 +1,255 @@
...
+    if is_cluster_running(cluster_path):
+        pytest.skip(
+            "The installation is skipped cause the cluster is running"
+        )
+
+    # Generate install-config from template
+    log.info("Generating install-config")
+    run_cmd(f"mkdir -p {cluster_path}")
+    pull_secret_path = os.path.join(
+        templating.TOP_DIR,
```
clacroix12: Once #166 is merged this will need to be imported from `constants` and not `templating`.

tests/conftest.py
```diff
@@ -0,0 +1,255 @@
...
+
+from oc.openshift_ops import OCP
+from ocs.exceptions import CommandFailed, CephHealthException
+from ocs.utils import create_oc_resource, apply_oc_resource
+from ocsci import config
+from ocsci.config import RUN, ENV_DATA, DEPLOYMENT
+from utility import templating
+from utility.aws import AWS
+from utility.retry import retry
+from utility.utils import destroy_cluster, run_cmd, get_openshift_installer, get_openshift_client, is_cluster_running
```
clacroix12: I know the rule isn't set in our flake8 check yet but let's keep this line < 79 chars.

tests/ecosystem/deployment/test_ocs_basic_install.py
```diff
@@ -13,13 +13,18 @@
 from utility import templating
 from utility.aws import AWS
 from utility.retry import retry
-from utility.utils import run_cmd, get_openshift_installer, get_openshift_client
+from utility.utils import run_cmd, get_openshift_installer, get_openshift_client, is_cluster_running
```
clacroix12: Same here regarding line length. < 79 chars.

dahorak:
> The failure will be reported in a `live log setup` section, or was when I tested locally. A session scoped fixture is only ran once for the entire test run so if the deployment fails then no actual tests will run so I'm not sure how a deployment failure could be confused here. A fixture with a scope of method would fail for each individual test that uses it though.

The main question is, if we consider cluster deployment as regular test or not? If not, then we can close this discussion as it doesn't make much sense. But if yes, then, we need to be able to report it as separated test case (e.g. in JUnit xml, reportportal, etc.)

My understanding is, that we (OCS4 QE team) are responsible for testing the whole deployment of the OCS4, which also include the installation (and destroy) part. Based on that, I think that installation should be test case similar to any other with same reporting capabilities etc..

> The idea is that there is no "deploy cluster" test anymore. Deploying a test cluster is just part of running any set of tests and is done automatically if a testing cluster is not already deployed and provided.

That is the initial and main question here... if "deploy cluster" should or should not be considered as test case.

> Why would you want to teardown a non existing cluster? If you run the `run-ci` command with the `--teardown` flag and you do not provide a cluster one will be spun up for testing, used for testing and then destroyed.

This might happen accidentally or because of some failure in the check if the cluster is available or not...

> Because this is a `session` scoped fixture it's going to operate a little differently here. You will see the error in `setup` section of the logs, not tied to each individual test like a `method` scoped fixture would.

How it will be reported in JUnit xml, into reportportal, etc.?

> I guess I'm not fully understanding why we'd want this in regular tests, could you please explain the benefit you see of this deployment and teardown code executing in a test method as opposed to a session-scoped fixture? It seems the same thing is being tested as a session-scoped fixture to me.

My point of view described above.

tests/conftest.py
andrewschoen: Thanks for the heads up! I'll get that changed.

tests/conftest.py
andrewschoen: Will do

andrewschoen:
> > The failure will be reported in a `live log setup` section, or was when I tested locally. A session scoped fixture is only ran once for the entire test run so if the deployment fails then no actual tests will run so I'm not sure how a deployment failure could be confused here. A fixture with a scope of method would fail for each individual test that uses it though.
> 
> The main question is, if we consider cluster deployment as regular test or not? If not, then we can close this discussion as it doesn't make much sense. But if yes, then, we need to be able to report it as separated test case (e.g. in JUnit xml, reportportal, etc.)

That's not a question I have the answer to, but I think that even if we do need to report a deployment failure by this session fixture as a separate test case we certainly could. We could also just change the ``test_cluster_is_running`` back to ``test_deployment`` or something so if there is a failure it can report with a different name.

> My understanding is, that we (OCS4 QE team) are responsible for testing the whole deployment of the OCS4, which also include the installation (and destroy) part. Based on that, I think that installation should be test case similar to any other with same reporting capabilities etc..

I understand that. I think there is a difference here between testing the ocs-ci deployment and if the bits used in the deployment resulted in a viable cluster though. I'd advocate against creating test cases/bugs out of failed ocs-ci deployments, but instead created a smoke test of sorts that verifies common install requirements. I.e. ports open, processes running, etc.

I just want to be careful we're not creating product bugs out of ocs-ci deployment issues. If that makes any sense. But maybe we do want to do that?

> > Why would you want to teardown a non existing cluster? If you run the `run-ci` command with the `--teardown` flag and you do not provide a cluster one will be spun up for testing, used for testing and then destroyed.
> 
> This might happen accidentally or because of some failure in the check if the cluster is available or not...

The teardown finalizer should run even if the deployment step fails. That's a nice thing about using a finalizer as opposed to the ``yield`` statement in fixtures.

> > Because this is a `session` scoped fixture it's going to operate a little differently here. You will see the error in `setup` section of the logs, not tied to each individual test like a `method` scoped fixture would.
> 
> How it will be reported in JUnit xml, into reportportal, etc.?

I definitely acknowledge the point of needing to report test failures, I'm just not sure a session fixture based deployment keeps us from doing that.

vasukulkarni:
I think we can try to discuss this in *thursday's* call and close on this one, the discussion seems lengthy here.

dahorak:
> > This might happen accidentally or because of some failure in the check if the cluster is available or not...
> 
> The teardown finalizer should run even if the deployment step fails. That's a nice thing about using a finalizer as opposed to the `yield` statement in fixtures.

My point was, that if I already have cluster, but the check if the specified cluster exists during the setup phase fails for some reason, it will silently create new cluster... (because it will think, that the specified cluster doesn't exist).

Thanks for the hint about difference between `yield` and finalizer.

tests/conftest.py
```diff
@@ -0,0 +1,255 @@
...
+@pytest.fixture(scope="session", autouse=True)
+def cluster(request):
+    log.info("Running OCS basic installation")
+    cluster_path = ENV_DATA['cluster_path']
+    # Add a finalizer to teardown the cluster after test execution is finished
+    if config.teardown:
+        request.addfinalizer(cluster_teardown)
+        log.info("Will teardown cluster because --teardown was provided")
+    # Test cluster access and if exist just skip the deployment.
+    if is_cluster_running(cluster_path):
```
RazTamir: Let's execute it only when user requests it to by specifying in the command line (i.e. --deploy_cluster)

vasukulkarni:
@dahorak please resolve your comments and add your approval too. Thanks

andrewschoen:
> I'm not sure, where is the problem, since I've provided the `--cluster-path` argument?

I added a commit last night that added a ``required=True`` argument for ``--cluster-path``, I've removed that for now. Unsure why pytest was not respecting that.

I think I've got the rest of the feedback in place here.

tests/conftest.py
```diff
@@ -46,13 +58,254 @@ def invalid_cephfs_storageclass(request):
...
+        log.info("The installation is skipped because the cluster is running")
+        return
+    elif not config.deploy:
+        pytest.fail(
+            "The given cluster can not be connected to: {}".format(cluster_path)
```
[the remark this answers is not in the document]
andrewschoen: Fixed in b6cce108d72495d3e6e4f94168773527d4b0db5a

mbukatov:
Hmm, `unittests/test_config.py` fails, https://travis-ci.org/red-hat-storage/ocs-ci/jobs/547225127:

```
AttributeError: module 'ocs.defaults' has no attribute 'deploy'
```

In other words, the test basically complains that `deploy` and `teardown` are not present in `ocs/defaults.py`. @petr-balogh @vasukulkarni What is the expected resolution here?

vasukulkarni:
> Hmm, `unittests/test_config.py` fails, https://travis-ci.org/red-hat-storage/ocs-ci/jobs/547225127:
> 
> ```
> AttributeError: module 'ocs.defaults' has no attribute 'deploy'
> ```
> 
> In other words, the test basically complains that `deploy` and `teardown` are not present in `ocs/defaults.py`. @petr-balogh @vasukulkarni What is the expected resolution here?

I am removing this in #218 but blocked by deployment test, Lets merge this one which is low risk #251

andrewschoen:
> In other words, the test basically complains that `deploy` and `teardown` are not present in `ocs/defaults.py`. @petr-balogh @vasukulkarni What is the expected resolution here?

I'm happy to update these tests here if they need to be, but it looks like they're going to be removed anyway?

tests/conftest.py
```diff
@@ -47,13 +59,257 @@ def invalid_cephfs_storageclass(request):
...
+        return
+    elif config.teardown and not config.deploy:
+        log.info("Attempting teardown of non-accessible cluster: %s", cluster_path)
+        return
+    elif not config.deploy and not config.teardown:
+        msg = "The given cluster can not be connected to: {}. ".format(cluster_path)
+        msg += "Provide a valid --cluster-path or use --deploy to deploy a new cluster"
+        pytest.fail(msg)
+    else:
+        log.info("A testing cluster will be deployed and cluster information stored at: %s", cluster_path)
```
mbukatov: Wouldn't this deploy new cluster in case of communication problems between `oc` and the already existing cluster, when `--deploy` is specified?

tests/conftest.py
```diff
@@ -47,13 +59,257 @@ def invalid_cephfs_storageclass(request):
...
+    if is_cluster_running(cluster_path):
+        log.info("The installation is skipped because the cluster is running")
+        return
+    elif config.teardown and not config.deploy:
+        log.info("Attempting teardown of non-accessible cluster: %s", cluster_path)
+        return
+    elif not config.deploy and not config.teardown:
+        msg = "The given cluster can not be connected to: {}. ".format(cluster_path)
+        msg += "Provide a valid --cluster-path or use --deploy to deploy a new cluster"
+        pytest.fail(msg)
```
mbukatov: It may be nice to fail when both teardown and deploy are specified. It looks like this will just run teardown if cluster is running, and deploy and teardown if cluster exists, but is not accessible.

mbukatov:
> > In other words, the test basically complains that `deploy` and `teardown` are not present in `ocs/defaults.py`. @petr-balogh @vasukulkarni What is the expected resolution here?
> 
> I'm happy to update these tests here if they need to be, but it looks like they're going to be removed anyway?

Yep. If I read explanation from Vasu right, we can just ignore this unit tests here.

dahorak:
Also it seems to be slightly inconsistent, that deployment test was removed, but destroy test not... But since we will have to add/update both destroy and deploy tests, I'm OK to leave it in this state in this PR.

tests/ecosystem/deployment/test_ocs_basic_install.py
```diff
@@ -1,250 +1,12 @@
...
-from ocs.parallel import parallel
+from ocsci.config import ENV_DATA
+from ocsci.testlib import deployment
+from utility.utils import is_cluster_running
 
 log = logging.getLogger(__name__)
 
 
 @deployment
-class TestDeployment(EcosystemTest):
```
petr-balogh: How we will deal with the testing of the different deployments?  We will need to test deployment on different platforms and put results to the polarion. So I was thinking about has different test cases for different platform with different polarion_id, and than have some hook to deselect tests which won't be for that specific platform. Can you give me some ideas how we will deal with this? Thanks

ocsci/pytest_customization/ocscilib.py
```diff
@@ -101,5 +115,11 @@ def process_cluster_cli_params(config):
         cluster_name = f"{cluster_name}-{cid}"
     if not cluster_path:
         cluster_path = os.path.join(cluster_dir_parent, cluster_name)
+    # adds teardown to osci_config.RUN['cli_params']
+    teardown = get_cli_param(config, "teardown", default=False)
```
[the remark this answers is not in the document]
mbukatov: Good point.

tests/ecosystem/deployment/test_ocs_basic_install.py
mbukatov: @petr-balogh I agree. We had some discussion about that last week and we stressed out that we will need a separate set of setup/deployment test cases. @dahorak is going to propose a change of this simple approach when this is needed, so that the 1st version of this setup/teardown fixture could be merged soon.

Basically one would run particular setup, and then run particular tests. The setup fixture will then do nothing when it checks that the cluster is already there.

tests/conftest.py
andrewschoen: I was able to run teardown on a cluster that did not pass the accessibility check. You'll need to be able to specify ``--teardown`` and ``--deploy`` if you want to spin up a testing cluster and immediately destroy it after testing. I think this is something we should allow.

tests/conftest.py
andrewschoen: Yes, if you have ``--deploy`` specified it will deploy a cluster if the one given can not be connected to.

docs/usage.md
```diff
@@ -43,15 +43,20 @@ to the pytest.
...
+    If the information given can not be used to access the cluster then
+    test execution will fail. If you wish to deploy a new cluster, give
+    a path to a new directory and also use the `--deploy` argument.
 * `--ocsci-conf` - with this configuration you can overwrite the default
     OCS-CI parameters defined in `conf/ocsci/default_config.yaml`
 * `--cluster-conf` - with this configuration you can overwrite the default
     parameters for cluster and deployment. See the example of such file
     [here](../conf/ocs_basic_install.yml).
+* `--deploy` - if this is given and a cluster can not be accessed from the
+    provided `--cluster-path` then a new test cluster will be deployed.
```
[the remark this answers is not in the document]
andrewschoen: Yes, a new cluster will only be deployed if the given cluster path can not be connected to and ``--deploy`` is given.

ocsci/pytest_customization/ocscilib.py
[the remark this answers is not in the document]
andrewschoen: This was just a style preference on my part. I find that ``config.teardown`` is much easier to use in code than ``config.RUN['cli_params']['teardown']``. Is this something you'd like me to change?

tests/conftest.py
```diff
@@ -1,14 +1,26 @@
...
 
 from ocs import constants
 from resources.ocs import OCS
 from utility.environment_check import environment_checker  # noqa: F401
+from oc.openshift_ops import OCP
+from ocs import constants
+from ocs.exceptions import CommandFailed, CephHealthException
+from ocs.utils import create_oc_resource, apply_oc_resource
+from ocsci import config
+from ocsci.config import RUN, ENV_DATA, DEPLOYMENT
```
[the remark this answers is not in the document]
andrewschoen: Yup, I'll need to update this PR to account for the work in #210 for sure.

tests/conftest.py
```diff
@@ -47,13 +59,257 @@ def invalid_cephfs_storageclass(request):
...
+    destroy_cluster(ENV_DATA['cluster_path'])
+    log.info("Destroying the test cluster complete")
+
+
+@pytest.fixture(scope="session", autouse=True)
+def cluster(request):
+    log.info("Running OCS basic installation")
+    cluster_path = ENV_DATA['cluster_path']
+    # Add a finalizer to teardown the cluster after test execution is finished
+    if config.teardown:
```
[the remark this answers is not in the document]
andrewschoen: You can, see my comment above. Would you like me to remove these and access RUN directly?

ocsci/pytest_customization/ocscilib.py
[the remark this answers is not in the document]
petr-balogh: If you would like to use it on more places in the funftion you can put it to some variableat the top of function but if we have some sections in the config, let's follow and use those please.

tests/conftest.py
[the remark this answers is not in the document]
petr-balogh: Yep cause it will cause the issues after #210 merged you cannot directly import some section from config.

tests/ecosystem/deployment/test_ocs_basic_install.py
andrewschoen: @petr-balogh off the top of my head I think the session fixture added here could become a proxy method that calls out to many other types of deployments. This could be chosen by the user by a new cli option perhaps? We could also use the pytest test collection hook and automatically select tests that would validate the selected deployment method. I believe those tests are what @dahorak was going to propose after this PR to validate the AWS deployment.

docs/usage.md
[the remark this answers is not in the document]
zmc: Thought we decided not to do at deploy at all if the cluster path already existed, to prevent leaking resources?

andrewschoen:
Thanks for the feedback everyone. I'm working on those, rebasing and will address the failing test as well.

andrewschoen:
This should be good to go now.

docs/usage.md
```diff
@@ -114,10 +113,11 @@ you can run following command:
 run-ci -m deployment --ocsci-conf conf/ocsci/custom_config.yaml \
```
vasukulkarni: probably we can remove -m deployment from usage (also in destroy part)

vasukulkarni:
overall LGTM

docs/usage.md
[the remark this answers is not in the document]
vasukulkarni: I think we can take this up as follow up PR and not block on this

tests/conftest.py
```diff
@@ -1,17 +1,322 @@
...
+        log.info("The installation is skipped because the cluster is running")
+        return
+    elif teardown and not deploy:
+        log.info("Attempting teardown of non-accessible cluster: %s", cluster_path)
+        return
+    elif not deploy and not teardown:
+        msg = "The given cluster can not be connected to: {}. ".format(cluster_path)
+        msg += "Provide a valid --cluster-path or use --deploy to deploy a new cluster"
+        pytest.fail(msg)
+    else:
```
clacroix12: From @zmc 's comment above:
> Thought we decided not to do at deploy at all if the cluster path already existed, to prevent leaking resources?

I thought this as well and I think that check needs to happen here, correct? Theoretically we could provide a `--cluster-path` that has information about an existing cluster, however that cluster is inaccessible for some reason. We would then proceed to spin up a new cluster and place the information about that cluster on top of the files that already exist for the provided cluster. I think we need an additional check here.

```suggestion
    elif cluster_dir_is_empty:
```
Something along this line really where we check to make sure that there aren't resources for another cluster in the provided path. If there are we should fail and log an error to the user. If not we can proceed with the cluster deployment and use the provided `--cluster-path`

vasukulkarni: spin up of cluster will only happen if --deploy is also provided, so this is same behaviour as current (if user provides -m deployment , a new deployment is expected)

clacroix12: Resolving as we've decided to take this up in a separate issue/PR

RazTamir:
@andrewschoen @petr-balogh I'm not sure if this was asked in this long conversation but what will happen when we'll have more deployments (VMware, IPI, UPI, RHHI.next, etc.)?
This bind the AWS installation to the framework which is bad (apologies if I missed something)

petr-balogh:
@RazTamir we had discussion about this Raz. Currently we don't have any other deployment and on last mtg I mentioned that we will have to refactor and decompose the deployment from the state how it's written now. 

For sure we will need to provide details about the deployment/platform etc to the config. So once we will run the deployment it will know which platform and details where to deploy and will use proper deployment depend on the platform.

I think that we will need to decompose the deployment to multiple stages.
1) In case of UPI - prepare resources (), in case of IPI - skip this stage
2) Install OCP
3) Depend on the platform we maybe needs to do some tweaks after installation like we currently do for AWS (Creating disks and attaching them to work nodes)
3) Installation of OCS

So if we will have this code in some class/classes in some module I would then just in the fixture (how it's now) initiate the proper class with parameters for specific platform and it should does the trick.

PR #174, opened 2019-06-07
Add helper functions for pvc

RazTamir:


RazTamir:
All is verified except the resize cause I hit bug https://bugzilla.redhat.com/show_bug.cgi?id=1713579

tests/test_add_mds_to_cluster.py
```diff
@@ -49,8 +49,6 @@ def setup(self):
     CEPH_OBJ = OCS(**self.fs_data)
```
shylesh: @RazTamir 
Why all are using OCS() class instead of StorageClass() ?
I still see StorageClass(OCS) inheriting and my master branch is uptodate .

Any other guidelines for the usage ?

RazTamir: things are changing every day :)
so we realized most of the functionality is around get(), create(), apply(), etc. so why do we need to maintain all those classes?
for these helper functions, it does make sense to have them inside a PVC class.
So, in general, I'd say it depend on the needs.
For example, for storage class (I'm still not familiar with all actions) I'm not sure we need more functionality other than create() and delete() so for now not in use so people are using OCS directly.

Apologies for all the changes but I hope we reached some good point where we are on the right direction

vasukulkarni: In last PR https://github.com/red-hat-storage/ocs-ci/pull/132 we agreed to defer multiple classes to keep things simple because there wasn't functionality specific to class that needed to be implemented, If the *future* has reached today then there should be one more discussion :)

vasukulkarni: I think the confusion is we still have to StorageKind and PVC class, probably should get rid of those in the interim.

shylesh: @RazTamir @vasukulkarni 
Yes true, I got confused because of those files and I mislead Shreekara as well.

shylesh: > things are changing every day :)
> so we realized most of the functionality is around get(), create(), apply(), etc. so why do we need to maintain all those classes?
As you could already see these are just basic functions required but in a larger context , for ex: In PVC resize you would use apply as one of the step but apply() itself doesn't have significance in carrying semantics of the actions on particular kind.
This is the pattern we were discussing about where there is need of classes for each kind and grouping the methods.

Anyways as we started with simplified approach, going forward we can expand on need basis.

> for these helper functions, it does make sense to have them inside a PVC class.
> So, in general, I'd say it depend on the needs.
> For example, for storage class (I'm still not familiar with all actions) I'm not sure we need more functionality other than create() and delete() so for now not in use so people are using OCS directly.
> 
> Apologies for all the changes but I hope we reached some good point where we are on the right direction

resources/pvc.py
```diff
@@ -0,0 +1,67 @@
...
+    def get_backed_pv(self):
+        """
+        Returns the backed PV name of pvc_name in namespace
+
+        Returns:
+            str: PV name
+        """
+        return self.data.get('spec').get('volumeName')
+
+    def get_size(self):
```
shylesh: same here as well, making this as @property so that access wil be pvc.size

vasukulkarni:
What is the workflow for tests now?  we should have create_pvc() in OCS that would return this object

```python
OCS = OCS()
pvc  = OCS.create_pvc(my_modified_dict)
pvc.resize()
#also nice to have
all_pvcs = OCS.get_pvcs()

```

resources/pvc.py
```diff
@@ -0,0 +1,67 @@
...
+    def get_status(self):
+        """
+        Returns the PVC status
+
+        Returns:
+            str: PVC status
+        """
+        return self.data.get('status').get('phase')
+
+    def get_backed_pv(self):
```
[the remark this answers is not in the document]
RazTamir: good idea!
Done

resources/pvc.py
RazTamir: done

RazTamir:
So we have 2 options:
for new PVC:
```python
pvc_obj = OCS(**PVC_DICT)
pvc_obj.create()
pvc_obj.resize()
#also nice to have
all_pvcs = OCS.get_pvcs()
```
For existing PVC
```python
pvc_oc_client = OCP('v1', "pvc', 'openshift-storage')
 # Next line will return a dict like in PVC_DICT but for existing PVC
pvc_obj = pvc_oc_client.get(resource_name='mypv') 
pvc_obj.resize(new_size=10)
```

resources/pvc.py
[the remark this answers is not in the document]
RazTamir: For size, it looks good thing to have but for the pv, I'm not sure.
Thought why yes?

vasukulkarni:
> So we have 2 options:
> for new PVC:
> 
> ```python
> pvc_obj = OCS(**PVC_DICT)
> pvc_obj.create()
> pvc_obj.resize()
> #also nice to have
> all_pvcs = OCS.get_pvcs()
> ```
> 
> For existing PVC
> 
> ```python
> pvc_oc_client = OCP('v1', "pvc', 'openshift-storage')
>  # Next line will return a dict like in PVC_DICT but for existing PVC
> pvc_obj = pvc_oc_client.get(resource_name='mypv') 
> pvc_obj.resize(new_size=10)
> ```

It looks mixed up and confusing to me, I can also do
```python
pvc_obj = PVC(**PVC_DICT)
```
which will be same as using OCS, so instead we can just use one object that will keep track of its 3 different objects which users can modify? and we are going to have many pv and pvc objects so its better to keep track in the OCS object for verifications.

RazTamir:
yeah sorry about that :(
This was the intention PVC(**PVC_DICT)

vasukulkarni:
> yeah sorry about that :(
> This was the intention PVC(**PVC_DICT)

Ok, probably we can take up that as different PR then, Thanks!

resources/pvc.py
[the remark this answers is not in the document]
vasukulkarni: Could be just "VolumeName" to be in sync with yaml.

vasukulkarni:
Overall LGTM as separate object, other enhancements can be different PR.

resources/pvc.py
vasukulkarni: new commit not yet pushed, I guess.

RazTamir:
> > yeah sorry about that :(
> > This was the intention PVC(**PVC_DICT)
> 
> Ok, probably we can take up that as different PR then, Thanks!

@vasukulkarni , take up what in different PR?
Did I miss comments?

vasukulkarni:
> > > yeah sorry about that :(
> > > This was the intention PVC(**PVC_DICT)
> > 
> > 
> > Ok, probably we can take up that as different PR then, Thanks!
> 
> @vasukulkarni , take up what in different PR?
> Did I miss comments?

We can give access to these objects from OCS() so that users later on can get those and modify without having to do bookkeeping in individual tests.

resources/pvc.py
RazTamir: @shylesh can I take care of those comments in next PR?
I totally messed up my branch with a scale test case and don't want to block this PR

resources/ocs.py
```diff
@@ -22,6 +22,15 @@ def __init__(self, **kwargs):
 
         Args:
             kwargs (dict):
+                1) For existing resource, use OCP.get() to get the resource's
```
ebenahar: Use OCS.reload() instead?

resources/pvc.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __init__(self, **kwargs):
+        """
+        Initializer function
+        kwargs:
+            See parent class for kwargs information
+        """
+        super(PVC, self).__init__(**kwargs)
+        self.temp_yaml = tempfile.NamedTemporaryFile(
+            mode='w+', prefix='PVC_', delete=False
```
ebenahar: Not sure why this is necessary here. In OCS() it does the same thing:
```python
        self.temp_yaml = tempfile.NamedTemporaryFile(
            mode='w+', prefix=self._kind, delete=False
        )
```

resources/pvc.py
```diff
@@ -0,0 +1,67 @@
...
+        )
+
+    def get_status(self):
```
ebenahar: isn't it relevant for any other kind?

resources/pvc.py
RazTamir: True true

resources/pvc.py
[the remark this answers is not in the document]
RazTamir: done

resources/pvc.py
RazTamir: I'm not sure if all kinds have status.phase as status

resources/ocs.py
RazTamir: Nice!
I hope this will work

RazTamir:
Fixes issue #178

zmc:
This BR broke flake8:
`./resources/pvc.py:4:1: F401 'tempfile' imported but unused`

Was CI not run?

vasukulkarni:
huh, it wont allow to merge for failed tests, checking..

RazTamir:
How is that possible that all checks passed?
In any case, sorry about that.
I'll check why my pre-commit git hook failed to catch it as well

zmc:
It didn't pass, but it wasn't reported to github for some reason. I'd investigate but I lack the proper permissions :)

PR #180, opened 2019-06-10
Force openshift binaries download and log versions

dahorak:
* add possibility to force OpenShift installer and client download (avoid issues with version mishmash)
* log downloaded version of installer and client


dahorak:
The intention is, to force download of openshift installer and client during deployment.

prsurve:
Hi @dahorak can you update installer_version as well so after pr gets merged everyone can use latest openshift installer

tests/ecosystem/deployment/test_ocs_basic_install.py
```diff
@@ -63,10 +63,11 @@ def test_deployment(self):
 
         # Download installer
         installer = get_openshift_installer(
-            DEPLOYMENT['installer_version']
+            DEPLOYMENT['installer_version'],
+            force_download=True,
         )
         # Download client
-        get_openshift_client()
+        get_openshift_client(force_download=True)
```
petr-balogh: Same here, but as I pointed in some our previous conversation I think that client is RUN specific I think and should be done each run in case the user will change the version in config.

tests/ecosystem/deployment/test_ocs_basic_install.py
```diff
@@ -63,10 +63,11 @@ def test_deployment(self):
 
         # Download installer
         installer = get_openshift_installer(
-            DEPLOYMENT['installer_version']
+            DEPLOYMENT['installer_version'],
+            force_download=True,
```
petr-balogh: I think that we should allow to user to avoid to install it all the time by passing some variable from our custom config, so ppl can create their custom config where they will specify to not download each time and take those data from config like:
force_download=DEPLOYMENT['force_download_installer']

I remember that Harish told me that at home he has bad network connection and he want to use the file which was downloaded before so that it's on ppls responsibility to delete the file if the want new one, WDYT?

dahorak:
@prsurve the version was always updated in separated PR (e.g.: https://github.com/red-hat-storage/ocs-ci/pull/127, https://github.com/red-hat-storage/ocs-ci/pull/37), so I would keep this habit and propose to update the version in separated PR.

tests/ecosystem/deployment/test_ocs_basic_install.py
dahorak: In current state of the tests, the client is downloaded only here in the deployment test (no other test is using the `get_openshift_client()` function), so I think it is ok, to force the download in the same way as for the installer... It might change once we will come with some more flexible approach to get the client for each test  (e.g. fixture)...

tests/ecosystem/deployment/test_ocs_basic_install.py
dahorak: @petr-balogh yeah, this make sense... will update it that way...

dahorak: Fixed in bc1c461.

tests/ecosystem/deployment/test_ocs_basic_install.py
petr-balogh: Yep, for sure the download client each time should be implemented in different PR. Just pointed to use that section here as you did. Thanks

utility/utils.py
```diff
@@ -728,6 +729,7 @@ def destroy_cluster(cluster_path):
 def get_openshift_installer(
     version=DEPLOYMENT['installer_version'],
     bin_dir=RUN['bin_dir'],
+    force_download=False,
```
vasukulkarni: I think you can get rid of default 'False' here, since its always set by default.

utility/utils.py
```diff
@@ -768,6 +776,7 @@ def get_openshift_installer(
 def get_openshift_client(
     version=RUN['client_version'],
     bin_dir=RUN['bin_dir'],
+    force_download=False,
```
vasukulkarni: same

conf/ocsci/default_config.yaml
```diff
@@ -37,12 +37,14 @@ RUN:
   kubeconfig_location: 'auth/kubeconfig' # relative from cluster_dir
   cli_params: {}  # this will be filled with CLI parameters data
   client_version: '{{ DEPLOYMENT["installer_version"] }}'
+  force_download_client: True
   bin_dir: './bin'
 
 # In this section we are storing all deployment related configuration but not
 # the environment related data as those are defined in ENV_DATA section.
 DEPLOYMENT:
   installer_version: "4.1.0-rc.5"
```
RazTamir: I saw the previous comment about the version but any chance to change this to latest and that's it?
By that we are sure to get the latest downstream which is what we want right?

utility/utils.py
dahorak: @vasukulkarni, I'm not sure if I understand what you meant... what is the problem to have default `False` here?

conf/ocsci/default_config.yaml
dahorak: @RazTamir it is not directly possible, because the files in `latest` [directory](https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/) still have particular version in the name, so there will have to be some additional code, which will parse the index page and chose the proper version...
So it is definitely worth to consider it, but I would propose to create new PR for that functionality.
Also I'm not sure, how the `latest` link works, because now (2019-06-11 14:45 CEST), it points to version 4.1.0 from June 5, but there is already newer version `4.1.1` from today...

utility/utils.py
vasukulkarni: @dahorak what I meant was the default value of True is set in config.yaml and seeing False in function initialization confused me a bit, I think we can safely remove this initialization.

vasukulkarni:
Also needs rebase for failing CI tests

utility/utils.py
dahorak: The reason, why it is set to `False` here and to `True` in the default configuration is, that this function (and mainly the second for getting client) might be used multiple times during any test run, to get the path of the required binary, so it doesn't make sense to download everything again multiple times, if it already exists. The variable in `default_config.yaml` is meant just for deployment - during which on the other hand make sense to force the download, to be sure it is used the proper version. With this in mind, it probably make sense, to move the `force_download_client` variable also to the `DEPLOYMENT` section...
Also I can't remove the initialization, since there are already two arguments with initialization before and I don't want to reorder the parameters of those functions.

dahorak: > With this in mind, it probably make sense, to move the `force_download_client` variable also to the `DEPLOYMENT` section...

Done in fc02eb0.

dahorak:
@vasukulkarni The reason of failing Travis seems to be, that I've defined the `force_download_...` variables only in `conf/ocsci/default_config.yaml`, but not in `ocs/defaults.py`. What is the reason to define it in both places? My understanding was, that the defaults should be in the `default_config.yaml` and the `defaults.py` should be nearly empty... and automatically loaded from the yaml file...

vasukulkarni:
@dahorak Please see this PR https://github.com/red-hat-storage/ocs-ci/pull/218 ,  Right now it can fail if the variable exists in both the places and the change happened only in one place, it cannot fail if its only in defaults.py or in the yaml, Can you check both the files and update to same value.

vasukulkarni:
@dahorak can you rebase and test, thanks!

dahorak:
> @dahorak can you rebase and test, thanks!

Rebased and tested.

PR #182, opened 2019-06-10
Change of ceph image version to v14

petr-balogh:
We agreed with Raz that we want to use latest version of upsteram all
the time so when we define v14 we don't need to maintain this version
when new version released and update it here once we decide to go to
v15.


vasukulkarni:
@petr-balogh how does this work with downstream images?

petr-balogh:
> @petr-balogh how does this work with downstream images?

we will have a custom config with the downstream images which we will pass via --cluster-conf param, so than we can easily overwrite the images by downstream images quite easily. But as @RazTamir explained in the email this evening by default we will be testing upstream images, just for jenkins runs we will be using the DS and of course if ppl want to test DS build they can pass custom config as well.

For those DS images I've just submitted another PR here #187

PR #183, opened 2019-06-10
Change installer and client to version 4.1.0

petr-balogh:
Currently the lastest version is pointing to 4.1.0 so we won't need
need to change this version each time when there is new RC like:
4.1.0-rc.9, but need to change just once there will be new version like
4.1.1 or 4.2.0.

### DIFF
--- a/conf/ocsci/default_config.yaml
+++ b/conf/ocsci/default_config.yaml
@@ -42,7 +42,7 @@ RUN:
 # In this section we are storing all deployment related configuration but not
 # the environment related data as those are defined in ENV_DATA section.
 DEPLOYMENT:
-  installer_version: "4.1.0-rc.5"
+  installer_version: "4.1.0"
 
 # Section for reporting configuration
 REPORTING:


nehaberry:
@petr-balogh even though we are not using the ocs/[defaults.py](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs/defaults.py), do we need to make the version change in ocs/defaults.py as well ? Just to keep everything in sync ?

Please suggest.

**Snip of Current content** 
# Be aware that variables defined above and below are not used anywhere in the
# config files and their sections when we rendering config!

INSTALLER_VERSION = '4.1.0-rc.5'

vasukulkarni:
@nehaberry the defaults.py is still in use, you really dont have to update in 2 places , note in the yaml is DEFAULTS not DEPLOYMENT

nehaberry:
@vasukulkarni completely agree. What I meant with defaults.py not being used is  that some of the variables mentioned in the files wont be used (L39 - 44 docstring).



```

# Those variables below are duplicate at the moment from default_config.yaml
# and once we drop support for old runner we will remove those variables from
# here and will have them defined only on one place.

# Be aware that variables defined above and below are not used anywhere in the
# config files and their sections when we rendering config!
```

Hence, there wont be a sync as this file has a hardcoded variable as "INSTALLER_VERSION = '4.1.0-rc.5'" , even though it is ignored during deployment :) 

But completely agree with the explanation and its ok to be left as it is

nehaberry:
Tried install with "  installer_version: "4.1.0" in my local branch and it succeeded.

$ oc get clusterversion
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.1.0     True        False         9m34s   Cluster version is 4.1.0

$ oc version
Client Version: version.Info{Major:"4", Minor:"1+", GitVersion:"v4.1.0-201905091432+4910781-dirty", GitCommit:"4910781", GitTreeState:"dirty", BuildDate:"2019-05-09T19:19:42Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.4+838b4fa", GitCommit:"838b4fa", GitTreeState:"clean", BuildDate:"2019-05-19T23:51:04Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}

PR #184, opened 2019-06-10
Fixing typo

prsurve:


mbukatov:
It seems to me that this is not a typo. The example is trying to show how to run the deployment  when one uses a custom configuration (to make the example closer to the real use cases). If the defaults are all ok for your case, you don't need to specify a config file explicitly like that.

RazTamir:
As Martin mentioned this is not a type

PR #187, opened 2019-06-10
Create cluster config with DS images

petr-balogh:


conf/ocsci/downstream_config.yaml
```diff
@@ -0,0 +1,4 @@
+---
+ENV_DATA:
+  ceph_image: 'quay.io/rhceph-dev/rhceph-4.0-rhel-8:latest'
+  rook_image: 'quay.io/rhceph-dev/rook:4.2-38.jenkins'
```
RazTamir: what about CSI and OCP?

petr-balogh: @RazTamir I sent email yesterday with the question to Boris and Ken as reply to email with the subject: Versioned rook builds in quay .  So till I will have some information from them I cannot specify other images here and once we will get some info it can be done in different PR. 

About OCP, not sure what exactly do you mean, we are using some specific version of installer which was changed in this PR: https://github.com/red-hat-storage/ocs-ci/pull/183/files

RazTamir: +1

mbukatov: Yesterday, I deployed cluster using:

```
rook_image: "quay.io/rhceph-dev/rook"
ceph_image: "quay.io/rhceph-dev/rhceph-4.0-rhel-8"
```
and all OCS images were coming from quay:

```
$ oc  get pods --namespace openshift-storage -o jsonpath="{..image}" | tr -s ' ' '\n' | sort | uniq
quay.io/cephcsi/cephfsplugin:v1.0.0
quay.io/cephcsi/rbdplugin:v1.0.0
quay.io/k8scsi/csi-attacher:v1.0.1
quay.io/k8scsi/csi-node-driver-registrar:v1.0.2
quay.io/k8scsi/csi-provisioner:v1.0.1
quay.io/k8scsi/csi-snapshotter:v1.0.1
quay.io/rhceph-dev/rook
quay.io/rhceph-dev/rook:latest
```

I don't know how these particular versions are selected, but I expect this will be handled by OLM in the end anyway.

mbukatov: @petr-balogh I would rather use `latest` tag instead of hardcoding particular `4.2-38.jenkins` tag, even though that I understand that in CI, you may want to set the tag to a particular version based on some build notification. I'm asking Boris if my understanding is correct.

petr-balogh: @mbukatov I saw few your emails about the versions of images, but didn't have time to go properly over them, so will try once I will have time or when I am back from Israel.  As you said, the versions will be handled by OLM which is not implemented yet AFAIK. I have some questions to Boris in mentioned email but I expected that we will get new build from engineering and we will get the information that new build will contains information about version of images we have to use for now till OLM will be ready but not sure here and we have to get some agreement on this from engineering:
```yaml
  ceph_image: 'quay.io/rhceph-dev/rhceph-4.0-rhel-8:4.2-38.jenkins'
￼   rook_image: 'quay.io/rhceph-dev/rook:4.2-38.jenkins'
```

Also we will have to finish with all the testing before we will move to the new one build, so I prefer to stick with some specific version in production jobs till we finish with all the testing and will be ready to consume the new build.

Anyway we will have to get some UMB message from engineering to move to new release which we will start testing.

mbukatov: I agree that for the CI, you need to specify exact versions anyway (there are many reasons for that, as we both noted above). But for a safe default one will use by hand, using `latests` tag seems more reasonable and I would expect it to be present this way in the repo. Boris just verified that the tag will always point to the latest jenkins build.

ktdreyer:
I'm commenting here to remind myself that when we rename the `rook` repository to `rook-ceph`, we will need to update this downstream YAML file.

conf/ocsci/downstream_config.yaml
mbukatov: @petr-balogh could you change `4.2-38.jenkins` tag to `latest`?

petr-balogh: @mbukatov thanks for reminder.  Done

mbukatov:
I tested this today via:

```
$ python run_ocsci.py --cluster-path ~/data/cluster-2019-06-20 --ocsci-conf=conf/ocsci/downstream_config.yaml --cluster-name mbukatov-ocsqe -m deployment tests
```

And  the cluster get deployed. That said, it reported one failure (which rotate out of scrollback buffer), the only issue I seen was:

```
12:14:58 - MainThread - utility.utils - INFO - Executing command: oc -n openshift-storage exec rook-ceph-tools-6cc97c4c4-jrdbf ceph health
12:15:00 - MainThread - utility.utils - DEBUG - CMD output: HEALTH_WARN crush map has legacy tunables (require firefly, min is hammer)
12:15:00 - MainThread - utility.retry - WARNING - Ceph cluster health is not OK. Health: HEALTH_WARN crush map has legacy tunables (require firefly, min is hammer)
```

petr-balogh:
I think that we are OK with this config for tasting latest DS images. I've just deployed the cluster but verification of health status of ceph failing with error mentioned in bug: https://bugzilla.redhat.com/show_bug.cgi?id=1722425

Issue #193, opened 2019-06-11
Resource StorageClass is broken


fbalak:
Class `StorageClass` from `resources.storage_class` have following issues:

- Directory in paths to yaml template is swapped. CephFS uses RBD dir and RBD uses CephFS dir.
- Path to yaml points to not existing file (`{rbd,cephfs}_storageclass.yaml`).
- Class gets `namespace` from template but storage classes don't use namespaces. This will be harder to fix because `OCS` class that is parent for this class expects `namespace` as argument.

RazTamir:
see my comment https://github.com/red-hat-storage/ocs-ci/pull/194#pullrequestreview-248047997

fbalak:
Fixed by https://github.com/red-hat-storage/ocs-ci/pull/213.

PR #194, opened 2019-06-11
Fix storageclass path

fbalak:
- Fixes https://github.com/red-hat-storage/ocs-ci/issues/193
- Storage class doesn't use `namespace`. Added default namespace so that the class can inherit from `OCS` class.

### DIFF
--- a/resources/storage_class.py
+++ b/resources/storage_class.py
@@ -6,7 +6,7 @@ import logging
 import yaml
 import tempfile
 
-from ocs import constants, defaults
+from ocs import constants
 from resources.ocs import OCS
 
 log = logging.getLogger(__name__)
--- a/resources/storage_class.py
+++ b/resources/storage_class.py
@@ -30,13 +30,13 @@ class StorageClass(OCS):
         self.interface = interface
         if self.interface == 'cephfs':
             self.yaml_path = os.path.join(
-                constants.TEMPLATE_CSI_RBD_DIR, "rbd_storageclass.yaml"
+                constants.TEMPLATE_CSI_FS_DIR, "storageclass.yaml"
             )
             # TODO: Implement
             pass
         if self.interface == 'rbd':
             self.yaml_path = os.path.join(
-                constants.TEMPLATE_CSI_FS_DIR, "cephfs_storageclass.yaml"
+                constants.TEMPLATE_CSI_RBD_DIR, "storageclass.yaml"
             )
             # TODO: Implement
             pass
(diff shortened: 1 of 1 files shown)


resources/storage_class.py
```diff
@@ -30,22 +30,26 @@ def __init__(self, interface, **kwargs):
...
+                constants.TEMPLATE_CSI_RBD_DIR, "storageclass.yaml"
             )
             # TODO: Implement
             pass
 
         self.sc_data = yaml.safe_load(open(self.yaml_path, 'r'))
         self.sc_data.update(kwargs)
+
+        # TODO(HACK): Fix properly inheritance to OCS, storage classes don't use namespace
+        self.sc_data['metadata']['namespace'] = defaults.ROOK_CLUSTER_NAMESPACE
```
mbukatov: [StorageClass](https://kubernetes.io/docs/concepts/storage/storage-classes/) objects are not in any [namespace](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/), in the same way as other low level cluster wide resources such as nodes or PVs. If the current design doesn't allow you to specify no namespace for a storage class, the design is invalid.

fbalak: Ok, created another issue https://github.com/red-hat-storage/ocs-ci/issues/195

fbalak:
@RazTamir What is correct way to create storage class in current implementation (if this file will be removed)?

resources/storage_class.py
nehaberry: > [StorageClass](https://kubernetes.io/docs/concepts/storage/storage-classes/) objects are not in any [namespace](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/), in the same way as other low level cluster wide resources such as nodes or PVs. If the current design doesn't allow you to specify no namespace for a storage class, the design is invalid.

+1.. Storageclass should not need any namespace

RazTamir:
@fbalak @mbukatov @nehaberry please review this - https://github.com/red-hat-storage/ocs-ci/pull/196
It should take care for any kind without namespace

fbalak:
`namespace` argument is not passed to `StorageClass` parent now. It should work as expected after https://github.com/red-hat-storage/ocs-ci/pull/196 is merged.

@RazTamir I think that this file and `secret.py` files should be removed in different PR. I am also not sure how to correctly create resources from these files when they are deleted. `StorageClass` class gets `interface` as its required parameter so it seems that splitting sc templates into `CephFS` and `RBD` is handled by this class.

RazTamir:
@fbalak you can use OCS in ocs.py directly like any other kind

fbalak:
Closing as https://github.com/red-hat-storage/ocs-ci/pull/213 is merged.

PR #196, opened 2019-06-11
Fix storageclass doesn't have namespace

RazTamir:


resources/ocs.py
```diff
@@ -35,7 +35,9 @@ def __init__(self, **kwargs):
         self.data = kwargs
         self._api_version = self.data.get('api_version')
         self._kind = self.data.get('kind')
-        self._namespace = self.data.get('metadata').get('namespace')
+        self._namespace = None
+        if 'metadata' in self.data:
```
fbalak: This check is not complete. This way it tries to get `namespace` even when it is not provided but some other metadata are present.
Something like:
```suggestion
        if 'metadata' in self.data and 'namespace' in self.data['metadata']:
```

fbalak: Or it is probably ok when `get` is used.

mbukatov: @fbalak Assuming no namespace is represented via `self._namespace = None`, your suggestion is not necessary.

RazTamir: yeah that was my intention - that in case of meta is there but namespace not --> None and we are all set (:

resources/ocs.py
```diff
@@ -35,7 +35,9 @@ def __init__(self, **kwargs):
...
+            self._namespace = self.data.get('metadata').get('namespace')
         self._name = self.data.get('metadata').get('name')
```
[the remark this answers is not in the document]
RazTamir: good catch!
Fixing now

PR #202, opened 2019-06-11
Fix test for defaults

petr-balogh:


andrewschoen:
> Not sure if I like the idea of having values for our unit tests mixed in with our legitimate defaults.

I agree. Couldn't we store a testing config in ``unittests/default_config.yml`` and then point ``run_ocsci.init_ocsci_conf`` at that?

petr-balogh:
@zmc I commented about the issue today here: https://github.com/red-hat-storage/ocs-ci/commit/f5912f7c3a60db949c2c3d012a9773cc3f6fb6cd  actually the test how was written didn't test the logic for the defaults.

The test how was written actually didn't allow you to put any different value to the default_config.yaml than is defined in defaults.py. I wanted to remove those values from defaults.py but see that you added all the config sections to defaults in this PR https://github.com/red-hat-storage/ocs-ci/pull/165/files#diff-5d4c2b9710e56826b51cfbf768c7e646 . Those sections are supposed to be loaded to the ocsci/config.py module and in case how you wrote the test you will all the time need to duplicate the values for the config and for the defaults and if you will pass the custom config it will even overwrite the default_config.yaml so that case it will also fail in case you pass some custom config to that test.

zmc:
https://github.com/red-hat-storage/ocs-ci/pull/206

Issue #207, opened 2019-06-11
Project directory restructure


clacroix12:
There have been several concerns about the structure of our project at the moment. A lot of this stems from the origins and evolution of the project. This issue will focus on organizing the directories into a structure that makes sense for us now. 

Proposed changes:

- Contents of `oc` and `mita` move to `ocs` module.
- `Resources` module moves to `ocs` module.
- Removal of `suites` directory.
- Removal of `run.py` (old test runner).
- Rename `ocsci` module to something like `framework` (suggestions welcome) to avoid confusion with `ocs` module. 

This leaves us with something that looks like this:
```
ocs-ci/
    conf/
        examples/
    docs/
    ocs/
        resources/
    framework/
        pytest_customization
    templates/
        CSI/
        ocs-deployment/
        pv_pvc/
        reporting/
    tests/
        e2e/
        ecosystem/
        manage/
    unittests/
    utility/
```

Thoughts, concerns, and suggestions welcome.

clacroix12:
Another idea using a more standard python project approach. Essentially moving all source under `ocs_ci`, keeping only unit tests (for now) and docs at the top level.
```
ocs-ci/
    docs/
    ocs_ci/
        conf/
        ocs/
        framework/
        templates/
        tests/
        utility/
    unittests/ (this eventually gets removed in favor of unit tests localized to their respective modules)
```

RazTamir:
+1 for the first with few questions:
- why docs not under framework? or a similar place?
- conf, the same question
 besides that looks good

clacroix12:
@RazTamir docs are for more than just the framework as they include things like how to write tests, how to contribute to the project as a whole, and coding guidelines that apply everywhere. I don't see them as something that ONLY applies to the framework so I wouldn't put it under that directory. 

`conf` could definitely move somewhere else but I'm not sure framework is the right place either. Once we get rid of the defaults yaml there the only thing that is left is the optional cluster configuration files, but these are really only used for the deployment.

clacroix12:
Something else to add to this is that we have `tests/helpers.py`. As such I think having a `utility` module at the top level is probably unnecessary. These will most likely end up being framework specific functions and as such can be consolidated with the `framework` module as well. 

Note `utility` needs some serious attention as there is a lot of code to be cleaned up here. Lots of remnants of ceph-ci and the old runner. This will be taken up in another issue but I think it's worth mentioning.

zmc:
[...]

PR #210, opened 2019-06-11
More unit test groundwork

zmc:


ocsci/__init__.py
```diff
@@ -6,3 +6,32 @@
 https://docs.pytest.org/en/latest/reference.html
 under section PYTEST_DONT_REWRITE
 """
+from dataclasses import dataclass, field, fields
+
+
+@dataclass
```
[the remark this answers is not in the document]
zmc: Very good point - I can update references to 3.6.

ocsci/tests/test_config.py
```diff
@@ -1,15 +1,19 @@
 # -*- coding: utf-8 -*-
-import ocs
+from pytest import fixture
+
+import ocs.defaults
 import ocsci
-import run_ocsci
+import ocsci.main
```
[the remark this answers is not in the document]
zmc: As of now, yes, because `ocsci/__init__.py` doesn't import `main`. It appears to me that it could, though, if we wanted it to.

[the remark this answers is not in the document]
clacroix12: Ah I see. I think this is okay then. It ensures you only get `main` if you explicitly import it. I can see other things needing to import `ocsci` but not necessarily get `main` along with it. Let's leave as is.

petr-balogh:
@zmc just looked at glance cause I unfortunately don't have enough time cause of CI gathering I am attending this week. See comment below and consider this before the merge.
@dahorak be aware of this that if it will get merged that we need to have python 3.7 on our Jenkins slaves cause of dataclasses are 3.7 feature and AFAIK on our centos slaves we have python 3.6 so it will break our jenkins automation.

tox.ini
```diff
@@ -4,9 +4,19 @@ envlist = py37,flake8
 [testenv]
 deps =
     -rrequirements.txt
+    pytest-cov
 setenv =
  PYTHONPATH=.
```
mbukatov: Do we still need to keep this `PYTHONPATH` hack there? It shouldn't be there in the first place and I see you changed the way the unittests are wired a bit.

zmc: Yeah, it looks like if we drop this, `constants.TEMPLATE_DEPLOYMENT_DIR` ends up having the value of `.tox/py37/lib/python3.7/site-packages/templates/ocs-deployment` instead of `templates/ocs-deployment`, which doesn't work.

I believe this may signal a larger issue with packaging; I've also noticed that all of our top-level modules end up in `$venv/lib/python3.7/site-packages`, which seems wrong...

ocsci/__init__.py
```diff
@@ -6,3 +6,32 @@
 https://docs.pytest.org/en/latest/reference.html
 under section PYTEST_DONT_REWRITE
 """
+from dataclasses import dataclass, field, fields
```
vasukulkarni: Can you add some header docs please on how this is getting used from high level

tox.ini
```diff
@@ -4,9 +4,19 @@ envlist = py37,flake8
...
-commands = {envpython} {envbindir}/run-ci --ignore=tests -c unittests/pytest.ini {posargs}
+commands = py.test \
```
vasukulkarni: can we add some unit test to check run-ci invocation is not broken?

mbukatov: No. If the unittests are working without that, there is no reason to run unittests via `run-ci`.

vasukulkarni: @mbukatov my point is also to test run-ci breakage in CI tests or else we will not catch that. I am fine running unit tests via py.test but we also want to test run-ci is not broken and that was the main reason we added it.

mbukatov: @vasukulkarni But that is not a unit test. We added the run-ci there as a hack to make the unit tests work with the new config code. If that is no longer necessary (I haven't rechecked myself), let's remove the hack.

zmc: @vasukulkarni can you help me understand what specifically you want covered by unit tests here?

vasukulkarni: @zmc the change here remove run-ci to run unittests and uses py.test, I am saying if you can add a unit test that can just invoke run-ci and even check usage that should be fine.

zmc: @vasukulkarni like this? https://github.com/red-hat-storage/ocs-ci/pull/210/commits/afdae23cc00434c8d69e749eafab894baa3cb489

vasukulkarni: @zmc yep thanks!

zmc:
unit tests got broken again, due to mismatched defaults: https://github.com/red-hat-storage/ocs-ci/pull/182

vasukulkarni:
@zmc sorry can't blame #182 it was all green there. So maybe we really want to update the test to ingore defaults change in 2 places?

zmc:
@vasukulkarni it was green because its branch didn't contain the test that checks for this.

ocsci/__init__.py
zmc: @vasukulkarni added!

ocsci/__init__.py
```diff
@@ -6,3 +6,34 @@
...
 """
+# Use the new python 3.7 dataclass decorator, which provides an object similar
+# to a namedtuple, but allows type enforcement and defining methods.
+from dataclasses import dataclass, field, fields
+
+
+@dataclass
+class Config:
+    DEPLOYMENT: dict = field(default_factory=dict)
+    ENV_DATA: dict = field(default_factory=dict)
```
andrewschoen: This is a total nitpick, but is there a strong reason to keep this all upper case still?

vasukulkarni:
@zmc lets wait for some of the old PR's to be merged as they are almost ready, else it will cause them to not just rebase but fix conflicts too.

ocsci/__init__.py
zmc: I'd be in favor of moving to lower case, but I'm not sure about the rest of the group

clacroix12: I think they should be lower case. We should probably 'reserve' upper case variables for things that come from `constants`, `defaults`, and _maybe_  global vars.

zmc:
@petr-balogh @andrewschoen @clacroix12 It took some tinkering to get Travis to cooperate, but now we have:
* A compatible backport of `dataclasses` for 3.6
* Unit tests and `flake8` running under both Python versions in Travis

zmc:
Rebased again

ocsci/__init__.py
andrewschoen: I agree. Should we do this in a different PR perhaps?

clacroix12: @andrewschoen @zmc yeah let's make that change in another PR, I'd rather not hold this work back for it since it's really just matching what those variables looked like before.

zmc:
@clacroix12 @andrewschoen I had to repush yet again to update references to the old config module - some had snuck in in the last couple days. Are there blockers to getting this merged?

clacroix12:
@zmc The only thing I can think of is if there are any test case PRs that we are really pushing to merge ASAP. 

@RazTamir / @vasukulkarni I see we have six Test Case PRs open but I'm unsure which (if any) are the high priority ones we were talking about getting in before the changes in this PR as well as #207. If any of those should hold up this PR let's identify which ones (and mark with high priority) otherwise I'd like to get this PR merged.

RazTamir:
Sure @clacroix12 ,
I'll review pending testcase PRs but this is a huge PR to merge - please ensure this is verified through:
- deployment
- destroy cluster
- running all manage tests

zmc:
Test case PRs will require minimal changes, or those changes can happen here in another rebase. I'm running a deployment now, and can run the manage cases once that's complete

tox.ini
```diff
@@ -1,12 +1,24 @@
 [tox]
-envlist = py37,flake8
+# Use a new pip so we can use PEP 508 URL requirements
+requires = pip >= 19.1.1
```
mbukatov: Are you sure about this? [Release notes of pip 19.1.1](https://pip.pypa.io/en/stable/news/#id1) doesn't mention anything about [PEP 508 URL requirements feature](https://www.python.org/dev/peps/pep-0508/#examples). Also I don't see you to use this feature in your pull request, looking at `git blame` it was added by Coady.

Pip 19.1.1 is very recent (2019-05-06), so that it isn't even packaged in Fedora 30.  When I tried it with `python3-pip-19.0.3-1.fc30.noarch`, it worked just fine. So I would rather not have the minimal version high for no reason. 

I suggest to either drop minimal version or find which pip version actually introduced the feature you are referencing here.

run_ocsci.py
```diff
@@ -1,123 +1,15 @@
...
-        '-p', 'ocsci.pytest_customization.ocscilib',
-        '-p', 'ocsci.pytest_customization.marks',
-        '-p', 'ocsci.pytest_customization.ocsci_logging',
-    ])
+    HERE = os.path.abspath(os.path.dirname(__file__))
     from utility.utils import add_path_to_env_path
     add_path_to_env_path(os.path.join(HERE, 'bin'))
-    exit(pytest.main(arguments))
+    arguments = sys.argv[1:]
+    exit(ocsci.main.main(arguments))
```
mbukatov: I would double check issues similar to https://github.com/red-hat-storage/ocs-ci/issues/233 (test that one can install both with clean state and then with cli tools already placed in bin directory).

tox.ini
clacroix12: @mbukatov when I attempted to use the url requirement pip told me I needed to update. I wasn't able to find the specific release version that included that PEP so I just went with the latest. I just tested with a fresh venv and it seems like installing `19.0` works but `18.0` doesn't. I think as long as we require `pip>=19` we should be good to go on this one.

mbukatov: Ah, ok. I checked the release notes and see that the feature was added in 18.1 (2018-10-05):

> Allow PEP 508 URL requirements to be used as dependencies.

So let's use that.

zmc: We need to specify a version [`>= 18.1`](https://pip.pypa.io/en/stable/news/#id90) for the PEP 508 feature. You're correct that I didn't add the usage of it, but tests in TravisCI with python3.6 will fail without a version that's new enough.

Re: Fedora - these dependencies will be installed by pip, and not the package manager.
Re: the version being recent - is there a reason to not use the latest version? I guess I can back it down to 18.1 if you really think it matters

run_ocsci.py
zmc: I did test that

zmc:
@RazTamir I ran the manage tests with this branch and with master. There are identical breakages, which look to be caused by the use of `global`s 😦

vasukulkarni:
@clacroix12 Lets give time till Tuesday, There are 2 more out of 4 old Testcases that can be merged, Also as Raz mentioned, this has to go through Deployment and couple of Tier1 tests

zmc:
Won't there just be more test cases to merge by then?

RazTamir:
Thanks Zack,

I just executed and noticed that as well. 
Maybe not relevant to this discussion but I want to create a function that will be executed after each test (fixture maybe) that will ensure there are no leftovers after the test. 
It will help us to ensure new tests are not causing any problems for existing tests by leaving leftovers

vasukulkarni:
> Won't there just be more test cases to merge by then?

@zmc not going to hold newer testcases that came after this one

tox.ini
mbukatov: > Re: Fedora - these dependencies will be installed by pip, and not the package manager.
> Re: the version being recent - is there a reason to not use the latest version? I guess I can back it down to 18.1 if you really think it matters

When running tests with tox, you are correct that the actual test environment is installed via pip into temporary virtualenv, which is isolated from whatever is packaged in the distribution, and for this reason it doesn't really matter which version we pick. But this doesn't apply for the python, pip, virtualenv and tox itself, which needs to be already installed before running the tests.

So I believe that the best approach here is to make sure that the tests will work with tox as packaged in Fedora (or any other recent distro), so that when one has tox installed from the distribution package, the tests will just work.

If we require the latest pip or tox, one needs to install new pip over the one from the distribution, which could create additional issues.

I noted this in the [README file for the unit tests](https://github.com/red-hat-storage/ocs-ci/blob/master/docs/unit_tests.md).

mbukatov:
Ah, we may need to add `.coverage` into `.gitignore` file.

tox.ini
zmc: Testing with the latest Fedora is fine, but we should also be testing with RHEL8 IMO. The reason we need to update pip, though is that Travis builds failing without doing that. It also doesn't need to install over the system-provided copy - it does that in a virtualenv, like everything else tox does.

zmc:
This has already been used to test deployments, among other things. Waiting on hearing that we want to merge this, as rebasing it every day has felt like a waste

vasukulkarni:
@zmc if you can post couple of tier1 test logs, then we can merge today.

zmc:
@vasukulkarni ok - is there a way to do this in Jenkins yet?

vasukulkarni:
@zmc Its been manual invocation, waht folks are doing is running the test in local laptop and pasting the logs in chat. I dont think we can run real tests here within jenkins.

vasukulkarni:
@zmc also can we remove this commit now: 33c74f241c3379b1ab0bb164c5c30e3a780ddd7f since it works with 3.6

petr-balogh:
> @vasukulkarni ok - is there a way to do this in Jenkins yet?

@zmc we have already some job here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/ but still it's in progress and we need to do some additional stuff like add logs to NFS share and some other stuff as well. The job is done by @dahorak, so he can answer about the details.

zmc:
@vasukulkarni what do you mean it works in 3.6?

vasukulkarni:
> > @vasukulkarni ok - is there a way to do this in Jenkins yet?
> 
> @zmc we have already some job here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/ but still it's in progress and we need to do some additional stuff like add logs to NFS share and some other stuff as well. The job is done by @dahorak, so he can answer about the details.

@petr-balogh Its more about running the tests with wip branch than with master.

zmc:
I'd still like to have a standard way of running this, even for non-master builds. We can address that later, though.

petr-balogh:
@zmc please document somewhere the process of using the values from config to avoid the issue we hit today with accessing values in the module level and give some background about this behavior please. (At least in next PR) I will approve this one if we want to merge it ASAP.

This I was struggling before and wanted to give this ability to access those values directly from the code on global level but this was maybe wrong assumption and not good approach.  We should have discussion about this approach on our next call same as discussion about defaults.py.

Thanks

tox.ini
mbukatov: @zmc see https://github.com/red-hat-storage/ocs-ci/pull/254

PR #216, opened 2019-06-12
Fixing test case name

Edm571:
Why are there no differences?


mbukatov:
@Edm571 you just renamed a file, so the github diff is empty as expected.

PR #220, opened 2019-06-13
ocs 368

wusui:
OCS 368 -- Setup only.  Test to come later.  I would like this checked out first.


wusui:
I will move this to tests/manage and rename the branch to wip-...

clacroix12:
@wusui just FYI You don't need to rename the branch, just add "WIP" to the title of the PR and it will be sufficient.

PR #224, opened 2019-06-13
Methods to switch between projects

### LINKED ISSUE #223: Add method to switch to another project
It will be helpful having a method to switch to another project.

jilju:
*Switch to another project
*Switch to default project
Fixes #223

### DIFF
--- a/ocs/ocp.py
+++ b/ocs/ocp.py
@@ -8,6 +8,7 @@ import yaml
 from ocs.exceptions import CommandFailed
 from utility.utils import TimeoutSampler
 from utility.utils import run_cmd
+from ocs import defaults
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


ocs/ocp.py
```diff
@@ -251,3 +252,34 @@ def wait_for_resource(
...
+    Returns:
+        bool: True on success, False otherwise
+    """
+    log.info(f'Switching to project {project_name}')
+    cmd = f'oc project {project_name}'
+    success_msgs = [
+        f'Now using project "{project_name}"',
+        f'Already on project "{project_name}"'
+    ]
+    if any(msg in run_cmd(cmd) for msg in success_msgs):
```
vavuthu: run_cmd will run twice if we are passing the project_name which we are already present

jilju: Done

ocs/ocp.py
```diff
@@ -251,3 +252,35 @@ def wait_for_resource(
...
+
+
+def switch_to_default_project():
+    """
+    Switch to default project
+
+    Returns:
+        bool: True on success, False otherwise
+    """
+    return switch_to_project(defaults.ROOK_CLUSTER_NAMESPACE)
```
nehaberry: @jilju by default project, user may assume the OCP namespace called "default".

Hence, either change the function name to "switch_to_default_rook_cluster_project" or some other name.

jilju: Done

mbukatov:
I don't think that this functionality is a good idea for automated tests. I would rather use the namespace in an explicit way then switching it like this.

zmc:
How might this be used?

jilju:
> I don't think that this functionality is a good idea for automated tests. I would rather use the namespace in an explicit way then switching it like this.

This method is just to switch between projects. For example, to delete a namespace which is created for a test case , you may need to switch back to a different project before running 'oc delete project project_name'

jilju:
> How might this be used?

@zmc 
Example:
This method can be used (just for the third command and its verification) while automating the below process. I know this is not a good example, but a simple way to describe the usage of this method.
```
$ oc new-project testproject
Now using project "testproject"
$ oc project
Using project "testproject"
$ oc project openshift-storage
Now using project "openshift-storage"
$ oc delete project testproject
project.project.openshift.io "testproject" deleted
```

Issue #225, opened 2019-06-13
Option to run command asynchronously


jilju:
It will be helpful to have an option like this -
Run a command locally and ignore. Later, check the status of the command and fetch the result.

RazTamir:
Hi Jilju,
Can you provide a background for this request?

jilju:
> Hi Jilju,
> Can you provide a background for this request?

@RazTamir 
This is to address change request https://github.com/red-hat-storage/ocs-ci/pull/208#discussion_r292865732

jilju:
PR #227 is merged. Hence closing this issue.

PR #227, opened 2019-06-13
Method to execute command in an asynchronous way

jilju:
Adding a method which can be used to run a command and check the status and result of the command at a later point of time.
Issue #225

### DIFF
--- a/utility/utils.py
+++ b/utility/utils.py
@@ -966,3 +966,33 @@ def get_random_str(size=13):
     """
     chars = string.ascii_lowercase + string.digits
     return ''.join(random.choice(chars) for _ in range(size))
+
+
+def run_async(command):
+    """
+    Run command locally and return without waiting for completion
+
+    Args:
+        command (str): The command to run.
+
+    Returns:
+        An open descriptor to be used by the calling function.
+    """
+    log.info(f"Executing command: {command}")
+    popen_obj = subprocess.Popen(
+        command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, shell=True
+    )
+
+    def async_communicate():
+        """
+        Wait for command to complete and fetch the result
+
+        Returns:
+            retcode, stdout, stderr of the command
+        """
+        stdout, stderr = popen_obj.communicate()
+        retcode = popen_obj.returncode
+        return retcode, stdout, stderr
+
+    popen_obj.async_communicate = async_communicate
+    return popen_obj
--- a/utility/utils.py
+++ b/utility/utils.py
@@ -977,10 +977,16 @@ def run_async(command):
 
     Returns:
         An open descriptor to be used by the calling function.
+
+    Example:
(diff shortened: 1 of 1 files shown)


utility/utils.py
```diff
@@ -966,3 +966,33 @@ def get_random_str(size=13):
...
+    def async_communicate():
+        """
+        Wait for command to complete and fetch the result
+
+        Returns:
+            retcode, stdout, stderr of the command
+        """
+        stdout, stderr = popen_obj.communicate()
+        retcode = popen_obj.returncode
+        return retcode, stdout, stderr
```
vavuthu: stdout is of type class 'bytes'. Do you want to return str instead of bytes.

To return str, you can add encoding='utf-8' to Popen()

vavuthu:
@jilju Could you add example on usage in doc string

utility/utils.py
jilju: Done

jilju:
> @jilju Could you add example on usage in doc string

@vavuthu Added

utility/utils.py
```diff
@@ -966,3 +966,39 @@ def get_random_str(size=13):
...
+    Returns:
+        An open descriptor to be used by the calling function.
+
+    Example:
+        command = 'oc delete pvc pvc1'
+        proc = run_async(command)
+        ret, out, err = proc.async_communicate()
+    """
+    log.info(f"Executing command: {command}")
+    popen_obj = subprocess.Popen(
```
vasukulkarni: you have to return this object for it to be meaningful for tests to later communicate with it.

utility/utils.py
```diff
@@ -966,3 +966,39 @@ def get_random_str(size=13):
...
+        command, stdin=subprocess.PIPE, stdout=subprocess.PIPE, shell=True,
+        encoding='utf-8'
+    )
+
+    def async_communicate():
```
vasukulkarni: dont think it is necessary to have this function as it is one liner at test code.

zmc:
What's the use-case for this, as opposed to simply invoking Popen directly?

utility/utils.py
jilju: popen_obj is the return value of this method.

utility/utils.py
jilju: Yes, open_obj can be used inside the test case to get the stdout and stderr using popen_obj.communicate() and then get return code using popen_obj.return code.
The method async_communicate just club those and make it simpler to use in testcase.
ret, out, err = popen_obj.async_communicate()

jilju:
> What's the use-case for this, as opposed to simply invoking Popen directly?

This can be used to run a command and check the status and result of the command at a later point of time. For example, start executing a long running command and continue with next steps of the test case. Verify the result of the command later.

utility/utils.py
vasukulkarni: missed the other function is a closure, thanks for clarifying.

PR #228, opened 2019-06-13
Fixing script name and adding retry functionality

prsurve:
- Changed script name 
- Changed tier from tier1 to tier2 as per polarion test case
- Added retry functionality to validate_cephfilesystem function
- Fixed create_cephfilesystem to avoid the creation of mds if already exists

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -7,9 +7,15 @@ import datetime
 import logging
 
 from ocs import constants, defaults, ocp
+<<<<<<< HEAD
 from ocsci import config
+=======
+from ocs.exceptions import TimeoutExpiredError
+from ocsci.config import ENV_DATA
+>>>>>>> 1:- Added retry capability for validate_cephfilesystem
 from resources import pod
 from resources.ocs import OCS
+from utility.retry import retry
 
 logger = logging.getLogger(__name__)
 
@@ -269,6 +275,7 @@ def get_cephfs_data_pool_name():
     return out[0]['data_pools'][0]
 
 
+@retry(TimeoutExpiredError, tries=5, delay=3, backoff=1)
 def validate_cephfilesystem(fs_name):
     """
      Verify CephFileSystem exists at ceph and k8s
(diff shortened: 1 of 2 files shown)


tests/helpers.py
```diff
@@ -373,11 +376,19 @@ def delete_cephblockpool():
 
 def create_cephfilesystem():
     """
-    Function for deploying CephFileSystem (MDS)
+    Function for deploying CephFileSystem (MDS) if CephFilesystem exists
+    deploying will be skipped
```
nehaberry: Can we write it like this?

Function for deploying CephFileSystem (MDS) . If CephFilesystem exists,
    deployment will be skipped
(OR)

Function for deploying CephFileSystem (MDS) . 
If CephFilesystem exists, MDS deployment will be skipped

tests/helpers.py
```diff
@@ -253,6 +255,7 @@ def get_cephfs_data_pool_name():
     return out[0]['data_pools'][0]
 
 
+@retry(TimeoutExpiredError, tries=5, delay=3, backoff=1)
 def validate_cephfilesystem(fs_name):
     """
      Verify CephFileSystem exists at ceph and k8s
```
nehaberry: It would be better to replace k8s with "OCP" .

Even line 291 to 295

    if result['metadata']['name']:
        logger.info(f"Filesystem got created from OCP Side")
        k8s_validate = True
    else:
        logger.error("Filesystem was not create at OCP Side")

prsurve: Done

tests/helpers.py
prsurve: Done

tests/helpers.py
```diff
@@ -373,11 +376,19 @@ def delete_cephblockpool():
...
 def create_cephfilesystem():
     """
-    Function for deploying CephFileSystem (MDS)
+    Function for deploying CephFileSystem (MDS). If CephFilesystem exists
+    deployment will be skipped
 
     Returns:
         bool: True if CephFileSystem creates successful
     """
+    POD = pod.get_all_pods(
```
ebenahar: What if CephFS exists from a previous test case as a leftover? I think we want to know about it and not to skip the creation. Maybe let the test the call this function decide wether the CephFS should exist or not.

prsurve: Hi @ebenahar Is it ok if we pass an arg to create_cephfilesystem like force_create so if it's true we can delete old mds and create a new one

ebenahar: @prsurve sounds good

prsurve: will work on this today

prsurve: done

tests/helpers.py
```diff
@@ -387,14 +391,23 @@ def delete_cephblockpool():
...
+        force_create (bool): If true it will delete old CFS and recreate new one, If False will skip creation of CFS
 
     Returns:
         bool: True if CephFileSystem creates successful
     """
     fs_data = copy.deepcopy(defaults.CEPHFILESYSTEM_DICT)
+    if force_create:
+        logger.info("Deleting CephFileSystem if exists")
+        delete_all_cephfilesystem()
+    else:
```
ebenahar: this still doesn't answer the situation in which you have a leftover FS and you want to know about it. I think it would be better to assert in case a FS already exists

PR #226, opened 2019-06-13
Adding MDS pod creation to ocs_basic_install

nehaberry:
This PR addresses #112 and #181 and #253 

1. Added MDS pods as part of rook-ceph deployment using the new kind classes (fixes #112)
2.  Changed the operator-openshift-with-csi.yaml (fixes #253)
3. Due to chnages in deployment, moved check for agent pods after running cluster.yaml
4. Added checks for Ceph pods (fixes #181)

Note: I already had a PR (#151) which was mistakenly closed by me. Hence creating a new PR 
Old PR - https://github.com/red-hat-storage/ocs-ci/pull/151

Please review

PR #229, opened 2019-06-14
Update: tests/manage/test_ocs_336_346.py

shreekarSS:
- Updated test func test_ocs_336() with removed fixtures which were creating and deleting cephfs (depends on PR https://github.com/red-hat-storage/ocs-ci/pull/226)
 - Added cleanup function cleanup_resource(), which checks resource if exists, else deletes
 - Added reload() for pvc object, which updates pvc status
 - Added pool creation for rbd --> test_ocs_346()

### DIFF
--- a/tests/manage/test_ocs_336_346.py
+++ b/tests/manage/test_ocs_336_346.py
@@ -1,5 +1,4 @@
 import logging
-import pytest
 
 from ocs import ocp, defaults, constants
 from ocsci.config import ENV_DATA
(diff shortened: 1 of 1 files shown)


shreekarSS:
http://pastebin.test.redhat.com/771823 --> test log

tests/manage/test_ocs_336_346.py
```diff
@@ -1,57 +1,17 @@
 import logging
```
RazTamir: By removing the setup and teardown you changed the test itself to something is not written in Polarion.
In addition, we need to separate between setup, test and teardown - this change cause all steps to be part of the test which is not good

shreekarSS: Updated with new PR https://github.com/red-hat-storage/ocs-ci/pull/239

Issue #200, opened 2019-06-11
Organize fixtures in a conventionalized way


ebenahar:
We should agree on a standard of where to put fixtures implementation. 
Currently, we have fixtures written in different places - [inside the test file](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/test_add_mds_to_cluster.py#L21), [in conftest.py](https://github.com/red-hat-storage/ocs-ci/pull/98/files#diff-484462fced51d1a06b1d93b4a44dd535R1) and [in fixtures.py ](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/fixtures.py).

mbukatov:
I would say that having some [fixtures](https://docs.pytest.org/en/latest/fixture.html) in test files and others in conftest.py files doesn't constitute a conflict or breach on standard pytest conventions.

It makes sense to keep fixtures which automates specific setup for small number of test cases (as present in a particular python file) in the test file, while moving the fixture into conftest makes it usable for wider number of tests, and so this makes sense for more general setup/teardown.

I'm against keeping fixtures in fixtures.py files, as it conflicts with standard pytest conventions, and creates additional layer on top of pytest for no good reason. Also both people new to pytest and experienced pytest users will find it weird (both will get into question why we are using fixtures in a different way compared to all documentation related to pytest found on the internet).

ebenahar:
Thanks @mbukatov, sounds good to me. Replaced fixtures.py with conftest.py - https://github.com/red-hat-storage/ocs-ci/pull/203

mbukatov:
Summary based on:

 * https://github.com/red-hat-storage/ocs-ci/pull/237#discussion_r294360820
 * https://github.com/red-hat-storage/ocs-ci/pull/237#issuecomment-502802865

## Description

Instead of referencing fixtures via direct reference like this:

```
from somewhere import setup_something

@pytest.mark.usefixtures(  # noqa: F405
    setup_something.__name__,
    ...
```

let's use [canonical pytest way of using fixtures](https://docs.pytest.org/en/latest/fixture.html), placing fixture into conftest.py file or in python file with the test (as noted above, this depends on expected reusability of the fixture):

```
@pytest.mark.usefixtures("setup_something")
class SomeTest():
    ...
```

or (for test functions or methods):

```
def test_something(setup_something):
    ...
```

mbukatov:
I don't think this was addressed (eg. there still is `tests/fixtures.py` file).

PR #237, opened 2019-06-15
Environment status checks

RazTamir:
- Compare env status before and after test execution for leftovers
 - New installer version 4.1.2


RazTamir:
A sample output:

            raise exceptions.LeftoversDetected(
>               f"\nThere are leftovers in the environment after test case:"
                f"\nResources added: {leftovers['Leftovers added']}"
                f"\nResources removed: {leftovers['Leftovers removed']}"
            )
E           ocs.exceptions.LeftoversDetected: 
E           There are leftovers in the environment after test case:
E           Resources added: [{'sc': {'new_value': '123456789'}}]
E           Resources removed: []

tests/test_environment_check.py:127: LeftoversDetected

utility/environment_check.py
```diff
@@ -0,0 +1,128 @@
...
+    leftovers = {'Leftovers added': [], 'Leftovers removed': []}
+    for kind in diffs_list:
+        if ADDED_RESOURCE in kind[''.join(kind.keys())]:
+            leftovers['Leftovers added'].append({
+                ''.join(kind.keys()):
+                kind[''.join(kind.keys())][ADDED_RESOURCE][
+                    ''.join(kind[''.join(kind.keys())][ADDED_RESOURCE])
+                ]
+            })
+            leftover_detected = True
```
ebenahar: if this one is True, there is no reason to check the removed resources. I suggest checking if there are removed resources only in case there are no added resources

ebenahar: however, you still want to check if there is anything to append to leftovers['Leftovers removed'] even if the previous leftover detection result was true.
So maybe having to leftover_detected args, one for added and one for removed

RazTamir: I'm not sure I follow:
regarding the first comment:
I can create a secret and not delete it (leftover added) and delete a pod from the env by mistake (leftover removed).
Regarding the second comment:
I will print everything in any case (for kind in diffs_list)

ebenahar: @RazTamir See my edit to the first comment

ebenahar: @RazTamir my mistake :) in case there are added resources and no removed ones, leftover_detected will remain True

RazTamir: > if this one is True, there is no reason to check the removed resources. I suggest checking if there are removed resources only in case there are no added resources.
A scenario where I created a resource and forgot to delete it will result in added_resource.
If I also deleted some other resource so I want to print it as well.
I think I'm missing something
> Actually, if there are added leftovers and no removed leftovers, leftover_detected will be False.
Why?
I'm not re-setting levftovers_detected to False ever again - after the first catch of leftover, it will print it
Same, I think I miss something

RazTamir:
More sample output:
I created (manually) PVC and deleted it during the test execution:


E           ocs.exceptions.LeftoversDetected: 
E           There are leftovers in the environment after test case:
E           Resources added: []
E           Resources removed: [{'pvcs': {'apiVersion': 'v1', 'kind': 'PersistentVolumeClaim', 'metadata': {'annotations': {'pv.kubernetes.io/bind-completed': 'yes', 'pv.kubernetes.io/bound-by-controller': 'yes', 'volume.beta.kubernetes.io/storage-provisioner': 'ceph.rook.io/block'}, 'creationTimestamp': '2019-06-15T17:00:03Z', 'finalizers': ['kubernetes.io/pvc-protection'], 'name': 'raz', 'namespace': 'openshift-storage', 'resourceVersion': '360973', 'selfLink': '/api/v1/namespaces/openshift-storage/persistentvolumeclaims/raz', 'uid': '04e28370-8f8f-11e9-9a05-061efedd80e4'}, 'spec': {'accessModes': ['ReadWriteOnce'], 'dataSource': None, 'resources': {'requests': {'storage': '1Gi'}}, 'storageClassName': 'rook-ceph-block', 'volumeMode': 'Filesystem', 'volumeName': 'pvc-04e28370-8f8f-11e9-9a05-061efedd80e4'}, 'status': {'accessModes': ['ReadWriteOnce'], 'capacity': {'storage': '1Gi'}, 'phase': 'Bound'}}}]

RazTamir:
And the opposite:
Started the test execution with clean env and as part of the test, created new PVC:

E           ocs.exceptions.LeftoversDetected: 
E           There are leftovers in the environment after test case:
E           Resources added: [{'pvs': {'apiVersion': 'v1', 'kind': 'PersistentVolume', 'metadata': {'annotations': {'pv.kubernetes.io/provisioned-by': 'ceph.rook.io/block'}, 'creationTimestamp': '2019-06-15T18:02:59Z', 'finalizers': ['kubernetes.io/pv-protection'], 'name': 'pvc-cf6c46b9-8f97-11e9-b80d-061efedd80e4', 'resourceVersion': '381786', 'selfLink': '/api/v1/persistentvolumes/pvc-cf6c46b9-8f97-11e9-b80d-061efedd80e4', 'uid': 'cf9a04e8-8f97-11e9-a7cf-0a4235d7f512'}, 'spec': {'accessModes': ['ReadWriteOnce'], 'capacity': {'storage': '1Gi'}, 'claimRef': {'apiVersion': 'v1', 'kind': 'PersistentVolumeClaim', 'name': 'raz', 'namespace': 'openshift-storage', 'resourceVersion': '381779', 'uid': 'cf6c46b9-8f97-11e9-b80d-061efedd80e4'}, 'flexVolume': {'driver': 'ceph.rook.io/openshift-storage', 'fsType': 'xfs', 'options': {'clusterNamespace': 'openshift-storage', 'dataBlockPool': '', 'image': 'pvc-cf6c46b9-8f97-11e9-b80d-061efedd80e4', 'pool': 'rbd', 'storageClass': 'rook-ceph-block'}}, 'persistentVolumeReclaimPolicy': 'Retain', 'storageClassName': 'rook-ceph-block', 'volumeMode': 'Filesystem'}, 'status': {'phase': 'Bound'}}}, {'pvcs': {'apiVersion': 'v1', 'kind': 'PersistentVolumeClaim', 'metadata': {'annotations': {'pv.kubernetes.io/bind-completed': 'yes', 'pv.kubernetes.io/bound-by-controller': 'yes', 'volume.beta.kubernetes.io/storage-provisioner': 'ceph.rook.io/block'}, 'creationTimestamp': '2019-06-15T18:02:59Z', 'finalizers': ['kubernetes.io/pvc-protection'], 'name': 'raz', 'namespace': 'openshift-storage', 'resourceVersion': '381789', 'selfLink': '/api/v1/namespaces/openshift-storage/persistentvolumeclaims/raz', 'uid': 'cf6c46b9-8f97-11e9-b80d-061efedd80e4'}, 'spec': {'accessModes': ['ReadWriteOnce'], 'dataSource': None, 'resources': {'requests': {'storage': '1Gi'}}, 'storageClassName': 'rook-ceph-block', 'volumeMode': 'Filesystem', 'volumeName': 'pvc-cf6c46b9-8f97-11e9-b80d-061efedd80e4'}, 'status': {'accessModes': ['ReadWriteOnce'], 'capacity': {'storage': '1Gi'}, 'phase': 'Bound'}}}]
E           Resources removed: []

utility/environment_check.py
```diff
@@ -0,0 +1,127 @@
...
+REMOVED_RESOURCE = 'iterable_item_removed'
+
+
+@pytest.fixture(scope='class')
+def environment_checker(request):
+    request.addfinalizer(get_status_after_execution)
+    get_status_before_execution()
+
+
+def get_status_before_execution():
```
ebenahar: Shouldn't we fetch this data out of the CephCluster yaml?
This will ensure we're comparing against the desired state, and will apply also for deploymet - we'll be able to execute this health check after the deployment is done

RazTamir: As discussed f2f, there are a lot of implications by fetching that info from cephcluster yaml.
As we need a quick solution that provides us with what we need, this PR answers that

utility/environment_check.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+POD = ocp.OCP(kind=constants.POD)
+SC = ocp.OCP(kind=constants.STORAGECLASS)
+CEPHFILESYSTEM = ocp.OCP(kind=constants.CEPHFILESYSTEM)
+CEPHBLOCKPOOL = ocp.OCP(kind=constants.CEPHBLOCKPOOL)
+PV = ocp.OCP(kind=constants.PV)
+PVC = ocp.OCP(kind=constants.PVC)
+SECRET = ocp.OCP(kind=constants.SECRET)
+NS = ocp.OCP(kind=constants.NAMESPACE)
```
shylesh: How about adding deployment, service kinds as well .

utility/environment_check.py
```diff
@@ -0,0 +1,127 @@
+"""
```
shylesh: Instead of utility function how about handling it in OCS.create() and OCS.delete() itself. Lets say we add a decorator @register_resource for create() and delete() functions. Every create will register (resource_kind, name, namespace) in a global list and every delete() will remove that element from global list. 

1. If we try to delete the resource which we haven't created, that element will not be in list so raise exception
2. If we created element but forgot to delete, during finalize just scan that list and raise exception.

utility/environment_check.py
```diff
@@ -0,0 +1,127 @@
...
+        all_namespaces=True
+    )['items']
+    defaults.ENV_STATUS_PRE['namespace'] = NS.get(all_namespaces=True)['items']
+
+
+def get_status_after_execution():
+    """
+
+    """
+    defaults.ENV_STATUS_POST['pod'] = POD.get(all_namespaces=True)['items']
```
shylesh: Do you want to capture complete details about a resource ? How about simply printing (name, namespace, kind)

ocs/ocp.py
```diff
@@ -88,6 +92,8 @@ def get(self, resource_name='', out_yaml_format=True, selector=None):
             dict: Dictionary represents a returned yaml file
         """
         command = f"get {self.kind} {resource_name}"
+        if all_namespaces and not self.namespace:
+            command += "-A"
```
shylesh: Why all namespaces ? If openshift creates OR deletes its house keeping resources even that would be counted right ? And if user wants to do garbage collection(assuming if user gets leftovers exception and wants to clean up the garbage) I feel it could lead to disaster. Am I missing something ?

ocs/exceptions.py
```diff
@@ -14,12 +14,11 @@ class UnexpectedBehaviour(Exception):
...
 
 
-# TODO: Delete once moved to pytest
-class UnknownStatusOfTestException(Exception):
+class ClassCreationException(Exception):
     pass
 
 
-class ClassCreationException(Exception):
+class LeftoversDetected(Exception):
```
shylesh: Somehow I am not convinced about this name. I feel following would be more readable.
1. ResourceLeftoversException()
2. GarbageLeftoversException().

Also Leftovers doesn't indicate the resources which deleted by test but not created, because there is nothing leftover there.

shylesh:
I have a small confusion on idea behind this PR.

Are we worried about aws resource consumption after whole test run and hence urge users to cleanup ?
If this is the case then we can make use of labels and annotations, where for any resource we create attach a label or annotations, Something like TO_DELETE=true. At the end of the run just scan all the resources with above annotation or label and delete it.

Please correct me if this totally doesn't make any sense.

bin/README.md
```diff
@@ -1,10 +0,0 @@
-# bin directory
-
-This directory is placed into `PATH` environment variable so that tools located
-here are available to all tests (this is done in
-[`run_ocsci.py`](../run_ocsci.py) wrapper).
-
-The directory is intentionally empty (there should be no other files committed
-there). Command line tools like `oc` and `openshift-install` are installed
-there during deployment (see
-[test_ocs_basic_install.py](../tests/ecosystem/deployment/test_ocs_basic_install.py)).
```
mbukatov: Why is this file deleted? I consider having a README file in every important directory useful to keep it's purpose visible, which helps us with maintenance and usage in the long term.

ocs/exceptions.py
RazTamir: I will change to ResourceLeftoversException().
But what about your "Also" part?

ocs/ocp.py
RazTamir: - > Why all namespaces ? If openshift creates OR deletes its house keeping resources even that would be counted right ?
You are right about the first part - I would not expect the OCP will add new entities just like that without any request
 And if user wants to do garbage collection(assuming if user gets leftovers exception and wants to clean up the garbage) I feel it could lead to disaster. Am I missing something ?
So the user should fix the teardown (:

utility/environment_check.py
RazTamir: To make it work on every create() delete(), you will need to maintain a map of all resources creation and deletion so you will know what is created as part of the resource.
For example, when I create a cephfilesystem, also 2 mds pods are being created so how can I count it?
I think the approach you are suggesting is much more complicated and doesn't add too much value.

utility/environment_check.py
RazTamir: sure, I will add it

utility/environment_check.py
RazTamir: yes,

for this initial PR, I'm just checking if new/old resources left/removed after the test finishes but in the future (for now, it is nice to have) we will add more functionality to check also changes in existing resources.
For example, if I'm modifying some existing resource, I want to ensure it is returning to its original values.

bin/README.md
RazTamir: I have no idea,
I will restore it in next PR.
Sorry

RazTamir:
> we can make use of labels and annotations, where for any resource we create attach a label or annotations, Something like TO_DELETE=true. At the end of the run just scan all the resources with above annotation or label and delete it.
> 
> Please correct me if this totally doesn't make any sense.

This is an interesting idea.
The only thing that I can think of that might cause issues is extra resources that are being created as part of others (see the previous comment about cephfilesystem)

bin/README.md
mbukatov: I added a response here, but can't see it anymore, so sorry if I'm replying twice. I would suggest to just return the file in a separate commit in this pull request, so that when this PR is merged via github's Squash and Merge, the change won't be visible as if nothing happened to the readme file in the final commit.

utility/environment_check.py
```diff
@@ -0,0 +1,127 @@
...
+        all_namespaces=True
+    )['items']
+    defaults.ENV_STATUS_POST['pv'] = PV.get(all_namespaces=True)['items']
+    defaults.ENV_STATUS_POST['pvc'] = PVC.get(all_namespaces=True)['items']
+    defaults.ENV_STATUS_POST['secret'] = SECRET.get(
+        all_namespaces=True
+    )['items']
+    defaults.ENV_STATUS_POST['namespace'] = NS.get(
+        all_namespaces=True
+    )['items']
```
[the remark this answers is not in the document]
RazTamir: can you please elaborate on what is the issue?
Not sure I understand this "get all pods for given list of kinds"

bin/README.md
RazTamir: I have this file in my local repo and it doesn't seem to be visible here.
In case it will not be part of the merge, I will add it right after the merge

utility/environment_check.py
mbukatov: > To make it work on every create() delete(), you will need to maintain a map of all resources creation and deletion so you will know what is created as part of the resource.
> For example, when I create a cephfilesystem, also 2 mds pods are being created so how can I count it?
> I think the approach you are suggesting is much more complicated and doesn't add too much value.

I agree. I would rather keep this simple at first. We can revisit the labeling idea (as noted in different comment) later, if we found it necessary, but right now I don't have enough experience to guess if that will be useful or not.

utility/environment_check.py
mbukatov: > if this one is True, there is no reason to check the removed resources. I suggest checking if there are removed resources only in case there are no added resources.
> Actually, if there are added leftovers and no removed leftovers, leftover_detected will be False.

I read it that `leftover_detected` variable covers both addition and removal of something. This could be noted eg. at line:

```
leftover_detected = False
leftovers = {'Leftovers added': [], 'Leftovers removed': []}
```

 And with that in mind, it make sense both cases each time, as these are independent and we care about both.

utility/environment_check.py
```diff
@@ -0,0 +1,127 @@
...
+    )
+    namespace_diff = DeepDiff(
+        defaults.ENV_STATUS_PRE['namespace'],
+        defaults.ENV_STATUS_POST['namespace']
+    )
+    diffs_list = [
+        {'pods': pod_diff}, {'sc': sc_diff}, {'cephfs': cephfs_diff},
+        {'cephbp': cephbp_diff}, {'pvs': pv_diff}, {'pvcs': pvc_diff},
+        {'secret': secret_diff}, {'ns': namespace_diff},
+    ]
```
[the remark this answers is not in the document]
RazTamir: Thanks!

I just realized it remained in a debugging form - fixed

RazTamir:
Sample output after new commit:

E           ocs.exceptions.ResourceLeftoversException: 
E           There are leftovers in the environment after test case:
E           Resources added: [{'pvs': {'apiVersion': 'v1', 'kind': 'PersistentVolume', 'metadata': {'annotations': {'pv.kubernetes.io/provisioned-by': 'ceph.rook.io/block'}, 'creationTimestamp': '2019-06-17T12:32:27Z', 'finalizers': ['kubernetes.io/pv-protection'], 'name': 'pvc-f6e62345-90fb-11e9-a630-02ec909f52e0', 'resourceVersion': '1175465', 'selfLink': '/api/v1/persistentvolumes/pvc-f6e62345-90fb-11e9-a630-02ec909f52e0', 'uid': 'f73f17c4-90fb-11e9-a630-02ec909f52e0'}, 'spec': {'accessModes': ['ReadWriteOnce'], 'capacity': {'storage': '1Gi'}, 'claimRef': {'apiVersion': 'v1', 'kind': 'PersistentVolumeClaim', 'name': 'raz', 'namespace': 'openshift-storage', 'resourceVersion': '1174477', 'uid': 'f6e62345-90fb-11e9-a630-02ec909f52e0'}, 'flexVolume': {'driver': 'ceph.rook.io/openshift-storage', 'fsType': 'xfs', 'options': {'clusterNamespace': 'openshift-storage', 'dataBlockPool': '', 'image': 'pvc-f6e62345-90fb-11e9-a630-02ec909f52e0', 'pool': 'rbd', 'storageClass': 'rook-ceph-block'}}, 'persistentVolumeReclaimPolicy': 'Retain', 'storageClassName': 'rook-ceph-block', 'volumeMode': 'Filesystem'}, 'status': {'phase': 'Released'}}}, {'pvcs': {'apiVersion': 'v1', 'kind': 'PersistentVolumeClaim', 'metadata': {'annotations': {'pv.kubernetes.io/bind-completed': 'yes', 'pv.kubernetes.io/bound-by-controller': 'yes', 'volume.beta.kubernetes.io/storage-provisioner': 'ceph.rook.io/block'}, 'creationTimestamp': '2019-06-17T12:39:07Z', 'finalizers': ['kubernetes.io/pvc-protection'], 'name': 'rrr', 'namespace': 'openshift-storage', 'resourceVersion': '1176567', 'selfLink': '/api/v1/namespaces/openshift-storage/persistentvolumeclaims/rrr', 'uid': 'e61bad28-90fc-11e9-a630-02ec909f52e0'}, 'spec': {'accessModes': ['ReadWriteOnce'], 'dataSource': None, 'resources': {'requests': {'storage': '1Gi'}}, 'storageClassName': 'rook-ceph-block', 'volumeMode': 'Filesystem', 'volumeName': 'pvc-e61bad28-90fc-11e9-a630-02ec909f52e0'}, 'status': {'accessModes': ['ReadWriteOnce'], 'capacity': {'storage': '1Gi'}, 'phase': 'Bound'}}}]
E           Resources removed: []

ocs/defaults.py
```diff
@@ -66,6 +66,9 @@
     'bin_dir': './bin',
 }
 
+ENV_STATUS_PRE = {}
+ENV_STATUS_POST = {}
```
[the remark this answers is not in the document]
RazTamir: do you suggest to move it to the exvironment_checker.py file?

bin/README.md
mbukatov: It is not going to be part of the merge in it's current state, as the file is removed in commit f92cbaeab758283b2c983d717f991cac7f80cdfa.

If you have the file in your local repository, you don't see it in `git status` output because `bin` directory is in gitignore. I was thinking if there is anything wrong with our code, but didn't noted anything which could have caused this.

Anyway, you can fix it quickly via:

```
$ rm bin/README.md # just in case
$ git checkout c9fc4fd30d8eba7cab3d5787de0eeb6a679a534b -- bin/README.md
$ git commit -m 'reverting README.md removal'
```

Then, when this is merged via "Squash and Merge", the resulting commit won't touch this file as if it were never deleted.

ocs/defaults.py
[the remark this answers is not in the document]
mbukatov: Yes, this would be better as it would prevent the state to be read or changed anywhere in the code. And it would make the code in env. checker self contained.

utility/environment_check.py
shylesh: > To make it work on every create() delete(), you will need to maintain a map of all resources creation and deletion so you will know what is created as part of the resource.
> For example, when I create a cephfilesystem, also 2 mds pods are being created so how can I count it?
> I think the approach you are suggesting is much more complicated and doesn't add too much value.

In this specific case it could still work, if you are expecting the test case to create and delete filesystem, we will be tracking the resource like this [(myfs, openshift-storage,CephFilesystem)]. During cleanup just delete myfs so any resource created as part of this fs will be wiped out.

If you want to create fs here and delete it elsewhere then this method won't work.

utility/environment_check.py
shylesh: sure

ocs/ocp.py
shylesh: Yep, that's why I said house keeping, which may not require any external request although I am not sure about any specific case.

utility/environment_check.py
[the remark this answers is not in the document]
mbukatov: Method OCP.get() is using inner state of OCP object to get kind and namespace. If there were a function which takes both as an argument instead, we could do something like:

```
kind_list = ["pod", "sc", "cephfs", ... ]
env_status = {}
for kind in kind_list:
    env_status[kind] = ocp_get(kind=kind, all_namespacess=True)
```

without initializing 8 different objects for each kind. It would also make it possible to move the code above into function which could be just called with env_status_pre in setup, and env_status_post in teardown.

ocs/defaults.py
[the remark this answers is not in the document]
RazTamir: done

utility/environment_check.py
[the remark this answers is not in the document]
RazTamir: creating such a function ocp_get() is pretty easy but I really want this checks to be part of the infrastructure as soon as possible cause there are many existing test cases that not cleaning the env after finish - this will enforce it.

As soon as this will be merged, I will implement such function and change the calls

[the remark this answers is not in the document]
RazTamir: creating such a function ocp_get() is pretty easy but I really want this checks to be part of the infrastructure as soon as possible cause there are many existing test cases that not cleaning the env after finish - this will enforce it.

As soon as this will be merged, I will implement such function and change the calls

RazTamir:
Hi @red-hat-storage/top-level-reviewers ,
I know there are a lot of things to improve in this mechanism and I'll be happy to do so but this PR is pretty urgent as we have many test case PRs coming in that do should not pass the quality level for test case - leaving leftovers in the environment and more bad, deleting existing ones.
So in case there is nothing really important to fix, I'll be happy to merge this PR and all improvements will be done in next PR (feel free to open issues about it and assign it to me)

ocsci/testlib.py
```diff
@@ -1,6 +1,10 @@
 from ocsci.pytest_customization.marks import *  # noqa: F403
+from utility import environment_check as ec
 
 
+@pytest.mark.usefixtures(  # noqa: F405
+    ec.environment_checker.__name__,
```
andrewschoen: You should only have to use the string 'environment_checker' here. Loading this fixture into ``tests/conftest.py`` makes is automatically accessible.

Also, why wouldn't we want to define this as an ``autouse`` fixture and just have it clean up after each test class has finished executing?

utility/environment_check.py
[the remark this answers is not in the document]
mbukatov: Agreed, as I noted in the [1st comment here](https://github.com/red-hat-storage/ocs-ci/pull/237/files#r294251297), we don't need to address it in this PR.

ocsci/testlib.py
mbukatov: I agree, the fixture usage is not optimal and what you are suggesting it a canonical pytest way. But since this is not fully followed in code which is already merged, I would not block this PR on this right now. It would be better to agree on change how to use fixtures, change it in all places and then start enforcing it during reviews.

ocs/defaults.py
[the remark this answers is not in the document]
mbukatov: I don't see any change.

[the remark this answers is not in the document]
RazTamir: Pushed now

RazTamir:
So all, please assign all issues to me

utility/environment_check.py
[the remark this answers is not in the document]
vasukulkarni: @mbukatov  It can be just a util function, dont know what it has to do with OCP class design.

vasukulkarni:
@RazTamir I am assuming some testcases are not cleaning up the objects they created which ideally should be testcases cleanup, but as such having this helper is important but I would like it to be configurable too, I will raise an issue for you and will not block this PR so that we can have clean runs.

shylesh:
> > we can make use of labels and annotations, where for any resource we create attach a label or annotations, Something like TO_DELETE=true. At the end of the run just scan all the resources with above annotation or label and delete it.
> > Please correct me if this totally doesn't make any sense.
> 
> This is an interesting idea.
> The only thing that I can think of that might cause issues is extra resources that are being created as part of others (see the previous comment about cephfilesystem)

According to my understanding, consider CephFS creation, it creates few pods and other resources.
If you delete cephfs lets say using its name ocs.delete('myfs') all the resources like pods etc created through this fs resource will be destroyed.
So effectively we need to keep only cephfs reference rather than mds pods created. I don't have complete working knowledge of openshift resources and how it will be used through out tests but I felt this is feasible.

RazTamir:
The "resources will be destroyed" part is something we as QE need to ensure and can't rely on.
Everything else is correct and I think this can be a great approach for teardowns - you can maintain a list of resources you created and just delete as part of your teardown. The env check will ensure everything is ok. 
Please open an issue and assign it to me. 

Thanks

ocsci/testlib.py
mbukatov: This is covered in https://github.com/red-hat-storage/ocs-ci/issues/200

utility/environment_check.py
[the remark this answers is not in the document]
mbukatov: https://github.com/red-hat-storage/ocs-ci/issues/248

utility/environment_check.py
```diff
@@ -0,0 +1,133 @@
...
+Util for environment check befora and after test to compare and find stale
+leftovers
+"""
+import logging
+import pytest
+
+from ocs import ocp, constants, exceptions
+from deepdiff import DeepDiff
+
+log = logging.getLogger(__name__)
```
petr-balogh: @RazTamir I see that log logger is not used anywhere, didn't want you print some info to the log?

RazTamir:
Only in the raise exception.
In case all good, I want to leave less logs a possible

PR #238, opened 2019-06-17
Update: tests/helpers.py

shreekarSS:
- Added a variable 'component' in functions create_pvc(),create_storage_class() and create_secret() which passed to create_unique_resource_name() which creates resources.
 - Resources can be identified easily with this component name in the resource name

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -110,14 +110,17 @@ def create_secret(interface_type):
         secret_data = defaults.CSI_RBD_SECRET.copy()
         del secret_data['data']['kubernetes']
         secret_data['data']['admin'] = get_admin_key()
+        component = 'rbd'
     elif interface_type == constants.CEPHFILESYSTEM:
         secret_data = defaults.CSI_CEPHFS_SECRET.copy()
         del secret_data['data']['userID']
         del secret_data['data']['userKey']
         secret_data['data']['adminID'] = constants.ADMIN_BASE64
         secret_data['data']['adminKey'] = get_admin_key()
+        component = 'cephfs'
+
     secret_data['metadata']['name'] = create_unique_resource_name(
-        'test', 'secret'
+        f'test-{component}', 'secret'
     )
     secret_data['metadata']['namespace'] = defaults.ROOK_CLUSTER_NAMESPACE
 
@@ -136,8 +139,7 @@ def create_ceph_block_pool(pool_name=None):
     """
     cbp_data = defaults.CEPHBLOCKPOOL_DICT.copy()
     cbp_data['metadata']['name'] = pool_name if pool_name else create_unique_resource_name(
-        'test', 'cbp'
-    )
+        'test', 'cbp')
     cbp_data['metadata']['namespace'] = defaults.ROOK_CLUSTER_NAMESPACE
     cbp_obj = create_resource(**cbp_data, wait=False)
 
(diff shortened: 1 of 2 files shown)


shreekarSS:
Examples:
secret-test-cephfs-1713121946
 pvc-test-cephfs-1713123327
pvc-test-rbd-1713122955

tests/helpers.py
```diff
@@ -110,14 +110,17 @@ def create_secret(interface_type):
         secret_data = defaults.CSI_RBD_SECRET.copy()
         del secret_data['data']['kubernetes']
         secret_data['data']['admin'] = get_admin_key()
+        component = 'rbd'
```
RazTamir: Add 'rbd' in constant file

tests/helpers.py
```diff
@@ -110,14 +110,17 @@ def create_secret(interface_type):
...
     elif interface_type == constants.CEPHFILESYSTEM:
         secret_data = defaults.CSI_CEPHFS_SECRET.copy()
         del secret_data['data']['userID']
         del secret_data['data']['userKey']
         secret_data['data']['adminID'] = constants.ADMIN_BASE64
         secret_data['data']['adminKey'] = get_admin_key()
+        component = 'cephfs'
```
RazTamir: same

tests/helpers.py
```diff
@@ -136,8 +139,7 @@ def create_ceph_block_pool(pool_name=None):
     """
     cbp_data = defaults.CEPHBLOCKPOOL_DICT.copy()
     cbp_data['metadata']['name'] = pool_name if pool_name else create_unique_resource_name(
-        'test', 'cbp'
-    )
+        'test', 'cbp')
```
RazTamir: fix parenthesis

tests/helpers.py
shreekarSS: Added in constants.py

tests/helpers.py
shreekarSS: Added in constants.py

tests/helpers.py
shreekarSS: Fixed!

tests/helpers.py
```diff
@@ -110,14 +110,17 @@ def create_secret(interface_type):
...
         secret_data = defaults.CSI_CEPHFS_SECRET.copy()
         del secret_data['data']['userID']
         del secret_data['data']['userKey']
         secret_data['data']['adminID'] = constants.ADMIN_BASE64
         secret_data['data']['adminKey'] = get_admin_key()
+        component = constants.CEPHFS_COMPONENT
+
     secret_data['metadata']['name'] = create_unique_resource_name(
-        'test', 'secret'
+        f'test-{component}', 'secret'
```
ebenahar: I think 'component' should replace the second arg, in this case - 'secret'

shreekarSS: secret would be the type of resource.
resource name would look like tthis --> secret-test-cephfs-1713121946

tests/helpers.py
```diff
@@ -207,9 +214,16 @@ def create_pvc(sc_name, pvc_name=None):
...
+        component = constants.RBD_COMPONENT
+    elif 'cephfs' in sc_name:
+        component = constants.CEPHFS_COMPONENT
+    else:
+        component = None
+
     pvc_data = defaults.CSI_PVC_DICT.copy()
     pvc_data['metadata']['name'] = pvc_name if pvc_name else create_unique_resource_name(
-        'test', 'pvc'
+        f'test-{component}', 'pvc'
```
jilju:
```
        f'test-{component}', 'pvc'
```
It is possible that 'component' can be NoneType here. It will give 'test-None'.

jilju: This is with reference to line 222, in case sc_name does not contain sub string 'rbd' or 'cephfs'

prsurve: jilju while creating storage with helpers.py it will have rbd or CephFS. But if someone creates storageclass explicitly I.e without helpers.py then componet will be none or later on future if we have support for rgw and till then this code is not been fixed for rgw for that componet will be none

PR #243, opened 2019-06-17
Polarion support

clacroix12:
Added the ability to mark tests with their respective `polarion_id`s in order to support posting our test results to Polarion. When we export our test results to a junit xml file (`--junit-xml report.xml`) any test cases that are marked with polarion_ids will have a custom property set that reflects the value. 

Ex.
```
<properties><property name="polarion-testcase-id" value="OCS-foo"/></properties>
```

Note that these changes do not add the ability to post the results to Polarion, only to support the tool we will use to take our xml test results and post the results.

### DIFF
--- a/conf/ocsci/default_config.yaml
+++ b/conf/ocsci/default_config.yaml
@@ -47,7 +47,7 @@ REPORTING:
   email:
     address: "ocs-ci@redhat.com"
   polarion:
-    project_id: "OCS"
+    project_id: "OpenShiftContainerStorage"
 
 # This is the default information about environment. Will be overwritten with
 # --cluster-conf file.yaml data you will pass to the pytest.
--- a/conf/ocsci/default_config.yaml
+++ b/conf/ocsci/default_config.yaml
@@ -46,6 +46,8 @@ DEPLOYMENT:
 REPORTING:
   email:
     address: "ocs-ci@redhat.com"
+  polarion:
+    project_id: "OCS"
 
 # This is the default information about environment. Will be overwritten with
 # --cluster-conf file.yaml data you will pass to the pytest.
--- a/ocs/defaults.py
+++ b/ocs/defaults.py
@@ -55,6 +55,9 @@ REPORTING = {
     'email': {
         'address': 'ocs-ci@redhat.com',
     },
+    'polarion': {
+        'project_id': 'OCS',
+    }
 }
 
 RUN = {
--- a/ocs/defaults.py
+++ b/ocs/defaults.py
@@ -56,7 +56,7 @@ REPORTING = {
         'address': 'ocs-ci@redhat.com',
     },
     'polarion': {
-        'project_id': 'OCS',
+        'project_id': 'OpenShiftContainerStorage',
     }
 }
 
--- a/ocsci/pytest_customization/ocscilib.py
+++ b/ocsci/pytest_customization/ocscilib.py
@@ -103,3 +103,13 @@ def process_cluster_cli_params(config):
         cluster_path = os.path.join(cluster_dir_parent, cluster_name)
     ocsci_config.ENV_DATA['cluster_name'] = cluster_name
     ocsci_config.ENV_DATA['cluster_path'] = cluster_path
+
+
+def pytest_collection_modifyitems(session, config, items):
+    """
+    Add Polarion ID property to test cases that are marked with one.
+    """
+    for item in items:
+        for marker in item.iter_markers(name="polarion_id"):
+            polarion_id = marker.args[0]
+            item.user_properties.append(("polarion-testcase-id", polarion_id))
--- a/ocsci/pytest_customization/ocscilib.py
+++ b/ocsci/pytest_customization/ocscilib.py
@@ -6,8 +6,8 @@ The basic configuration is done in run_ocsci.py module casue we need to load
 all the config before pytest run. This run_ocsci.py is just a wrapper for
 pytest which proccess config and passes all params to pytest.
 """
+import logging
 import os
-
 import random
 
 import ocs
@@ -16,6 +16,8 @@ from ocsci import config as ocsci_config
 __all__ = [
     "pytest_addoption",
 ]
+
+log = logging.getLogger(__name__)
 
 
 def pytest_addoption(parser):
@@ -110,6 +112,16 @@ def pytest_collection_modifyitems(session, config, items):
     Add Polarion ID property to test cases that are marked with one.
     """
     for item in items:
-        for marker in item.iter_markers(name="polarion_id"):
-            polarion_id = marker.args[0]
-            item.user_properties.append(("polarion-testcase-id", polarion_id))
+        try:
+            marker = item.get_closest_marker(name="polarion_id")
+            if marker:
+                polarion_id = marker.args[0]
+                item.user_properties.append(
+                    ("polarion-testcase-id", polarion_id)
+                )
+        except IndexError:
+            log.warning(
+                f"polarion_id marker found with no value for "
+                f"{item.name} in {item.fspath}",
+                exc_info=True
+            )
--- a/ocsci/pytest_customization/marks.py
+++ b/ocsci/pytest_customization/marks.py
@@ -34,6 +34,7 @@ scale = pytest.mark.scale
 deployment = pytest.mark.deployment
 destroy = pytest.mark.destroy
 upgrade = pytest.mark.upgrade
+polarion_id = pytest.mark.polarion_id
 
 # mark the test class with marker below to ignore leftover check
 ignore_leftovers = pytest.mark.ignore_leftovers
--- a/pytest.ini
+++ b/pytest.ini
@@ -19,6 +19,7 @@ markers =
     scale: scale related tests
     deployment: deplyment related tests
     upgrade: upgrade related tests
+    polarion_id: ID of test case used for reporting to Polarion
 
 
 log_format = %(asctime)s - %(threadName)s - %(name)s - %(levelname)s - %(message)s
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,7 @@ import pytest
 import yaml
 
 from ocs import constants
+from ocsci.config import REPORTING
 from resources.ocs import OCS
 from utility.environment_check import environment_checker  # noqa: F401
 
@@ -57,3 +58,12 @@ def invalid_cephfs_storageclass(request):
         f"{request.param['storageclass_name']}"
     )
     storageclass.delete()
+
+
+@pytest.fixture(scope="session", autouse=True)
+def polarion_testsuite_properties(record_testsuite_property):
+    """
+    Configures polarion testsuite properties for junit xml
+    """
+    polarion_project_id = REPORTING['polarion']['project_id']
+    record_testsuite_property('polarion-projct-id', polarion_project_id)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,7 +4,7 @@ import pytest
 import yaml
 
 from ocs import constants
-from ocsci.config import REPORTING
+from ocsci import config
 from resources.ocs import OCS
 from utility.environment_check import environment_checker  # noqa: F401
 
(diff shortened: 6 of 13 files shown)


ocsci/pytest_customization/ocscilib.py
```diff
@@ -103,3 +103,13 @@ def process_cluster_cli_params(config):
...
     ocsci_config.ENV_DATA['cluster_path'] = cluster_path
+
+
+def pytest_collection_modifyitems(session, config, items):
+    """
+    Add Polarion ID property to test cases that are marked with one.
+    """
+    for item in items:
+        for marker in item.iter_markers(name="polarion_id"):
+            polarion_id = marker.args[0]
```
vasukulkarni: Please add a skip when we dont have an ID

ocsci/pytest_customization/ocscilib.py
```diff
@@ -103,3 +103,13 @@ def process_cluster_cli_params(config):
...
     ocsci_config.ENV_DATA['cluster_name'] = cluster_name
     ocsci_config.ENV_DATA['cluster_path'] = cluster_path
+
+
+def pytest_collection_modifyitems(session, config, items):
+    """
+    Add Polarion ID property to test cases that are marked with one.
+    """
+    for item in items:
+        for marker in item.iter_markers(name="polarion_id"):
```
petr-balogh: I would rather use https://docs.pytest.org/en/latest/reference.html#_pytest.nodes.Node.get_closest_marker WDYT?

ocsci/pytest_customization/ocscilib.py
```diff
@@ -103,3 +103,13 @@ def process_cluster_cli_params(config):
...
+            polarion_id = marker.args[0]
+            item.user_properties.append(("polarion-testcase-id", polarion_id))
```
petr-balogh: We will need to add one more global required parameter under testsuites to junit file. See the documentation:
https://mojo.redhat.com/docs/DOC-1073077#jive_content_id_polarionprojectid_required

ocsci/pytest_customization/ocscilib.py
mbukatov: Do you mean a situation when there is a "polarion_id" marker, but used without argument, like this?

```
@pytest.mark.polarion_id
def test_foo_bar(setup_something):
    ...
```

In such case, the pytest immediately fails with `IndexError: tuple index out of range` exception.
And it is true that we should handle this error better in some way. But I would rather not skip or ignore it without some warning.

ocsci/pytest_customization/ocscilib.py
mbukatov: Yes, but I would rather not add it for every test case individually.

ocsci/pytest_customization/ocscilib.py
vasukulkarni: Yeah when the ID is missing we will hit the error you pasted above,  if we can handle this in unittest that will be great but if not we can handle here.

vasukulkarni:
> This looks good.
> 
> Note that `polarion-project-id` property is mandatory for `testsuites` element (as noted in docs of polarion ci importer). But the junit xml report starts with `testsuite` element:
> 
> ```
> (venv) [ocsqe@localhost ocs-ci]$ xmllint --format report.xml
> <?xml version="1.0" encoding="utf-8"?>
> <testsuite errors="0" failures="1" name="pytest" skipped="0" tests="2" time="0.125">
>   <testcase classname="tests.test_foo" file="tests/test_foo.py" line="2" name="test_foo_bar" time="0.002">
>     <properties>
>       <property name="polarion-testcase-id" value="FOO-123"/>
>     </properties>
>   </testcase>
>   <testcase classname="tests.test_foo" file="tests/test_foo.py" line="6" name="test_foo_baz" time="0.001">
>     <properties>
>       <property name="polarion-testcase-id" value="FOO-120"/>
>     </properties>
>     <failure message="assert 1 == 0   -1   +0">@pytest.mark.polarion_id("FOO-120")
>     def test_foo_baz():
> &gt;       assert 1 == 0
> E       assert 1 == 0
> E         -1
> E         +0
> 
> tests/test_foo.py:9: AssertionError</failure>
>   </testcase>
> </testsuite>
> ```
> 
> So the question is if importer will recognize `polarion-project-id` property in `testsuite` element (which we could provide in pytest), or whether we need to wrap the report into `testsuites` element and add the project id there (which could be done in the CI job pushing the report into the importer).
> 
> That said, I'm ok with letting this in, as it's not clear at which point we should add `polarion-project-id` property.

project-id and test-run-id can be set at jenkins level? since they are just one time per testsuite?

ocsci/pytest_customization/ocscilib.py
clacroix12: > And it is true that we should handle this error better in some way. But I would rather not skip or ignore it without some warning.

I agree with this, we can raise a warning in that scenario so it doesn't go unnoticed.

ocsci/pytest_customization/ocscilib.py
clacroix12: Reiterating what we discussed earlier: I think `get_closest_marker` is a good idea. This could save us in the event that both a test method and the test class it resides in are marked with polarion_ids that conflict. With `get_closest_marker` we would only use the value from the test level marker and not the class. This is still a situation we don't want to have present in our tests, but we also don't want to post results to the wrong tests in the event it happens. +1 for `get_closest_marker`.

ocsci/pytest_customization/ocscilib.py
clacroix12: Yeah like @mbukatov said in his top level comment we might need to wrap the output from our tests in a `<testsuites>` element as it doesn't exist in the report by default. The property you linked to needs to be a property of `<testsuites>` not the `<testcase>` so I wouldn't add it in this loop.

clacroix12:
> project-id and test-run-id can be set at jenkins level? since they are just one time per testsuite?

@vasukulkarni That's one possibility as we will already need to wrap the junit.xml in a `<testsuites>` element anyways. We might as well handle the `<testsuites>` level properties in the same operation. The other possibility is we format it and add the properties ourselves as a session level finalizer if we have passed the CLI args to export the junit.xml results. Either of which I think can be addressed outside of this PR once we have made some progress with running the actual importer.

ocsci/pytest_customization/ocscilib.py
petr-balogh: Ok, I've just dig in to how it's done in rhev and you can see one of generated xunit file which works for polarion importer in RHV automation here: https://rhv-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/4.3/job/rhv-4.3-ge-runner-tier1/81/artifact/xunit_output.xml 

I see that it can be in testusite under properties as you can see here
```xml
<testsuite errors="0" failures="3" name="pytest" skips="42" tests="706" time="68101.387">
    <properties>
        <property name="polarion-custom-isautomated" value="True"/>
        <property name="polarion-testrun-status-id" value="inprogress"/><property name="polarion-custom-plannedin" value="RHV_4_3"/>
        <property name="polarion-user-id" value="ci-user"/>
        <property name="polarion-project-id" value="RHEVM3"/><property name="polarion-response-myproduct" value="rhvm"/>
        <property name="polarion-testrun-id" value="RHV_4_3_tier1_x86_64"/><property name="polarion-custom-arch" value="x8664"/>
    </properties>
    <testcase classname="rhevmtests.compute.sla.cpu_qos.cpu_qos_test.TestQoSAndCpuProfileCRUD" file="rhevmtests/compute/sla/cpu_qos/cpu_qos_test.py" line="60" name="test_a_add_qos" time="11.999502182"><properties><property name="polarion-testcase-id" value="RHEVM-14931"/></properties>
        <system-out>
...output omitted...
```

petr-balogh:
> > project-id and test-run-id can be set at jenkins level? since they are just one time per testsuite?
> 
> @vasukulkarni That's one possibility as we will already need to wrap the junit.xml in a `<testsuites>` element anyways. We might as well handle the `<testsuites>` level properties in the same operation. The other possibility is we format it and add the properties ourselves as a session level finalizer if we have passed the CLI args to export the junit.xml results. Either of which I think can be addressed outside of this PR once we have made some progress with running the actual importer.

I prefer to prepare whole xunit file in ocs-ci as we can then just use it even in some manual upload to polarion without any further changes, and in the future we will need maybe add more properties.

clacroix12:
I wanted to bring attention to a small issue I'm seeing with these updates. In adding the log statement in `ocscilib.py` it seems like some of the pytest output gets captured and placed in the `<system-out>` of one of my test functions in the junit report xml. I'm not exactly sure why and it appears to only happen for the first test case that is executed.  I have confirmed that simply commenting out the log statement is enough to get this output to not show up in the junit xml.

```xml
    <testcase classname="tests.test_pytest_html.TestPolarionExample"
              file="tests/test_pytest_html.py" line="30"
              name="test_without_marker" time="0.003">
        <system-out>#x1B[1m
            collected 3 items #x1B[0m

            tests/test_pytest_html.py::TestPolarionExample::test_without_marker
        </system-out>
    </testcase>
```

This isn't a blocker and as such didn't want to hold up this PR for it, but I wanted to call it out and see if anyone would have any ideas as to why this is happening.

ocsci/pytest_customization/ocscilib.py
clacroix12: I added a session fixture to handle adding these properties. For now it only adds the `polarion-project-id` but we can update it to add any others we might need.

ocsci/pytest_customization/ocscilib.py
clacroix12: Done

ocsci/pytest_customization/ocscilib.py
clacroix12: Done

ocsci/pytest_customization/ocscilib.py
dahorak: What if one automated test covers more test cases in Polarion? (not sure, if it is "allowed" or not, but I think, it is possible to have such case...)

On the other side it doesn't make sense to add Polarion ID on class level, because it doesn't make sense to have more automated tests reporting to one test case in Polarion...

mbukatov: @dahorak I'm not sure if the importer will be able to handle this anyway.

ocsci/pytest_customization/ocscilib.py
mbukatov: Good.

mbukatov:
> This isn't a blocker and as such didn't want to hold up this PR for it, but I wanted to call it out and see if anyone would have any ideas as to why this is happening.

This looks like a peculiarity of the pytest xml report. Logs and output from collecting phase doesn't belong to any test and so the xml report seems to just add it into the 1st one instead of dropping it from the xml report.

I don't think this is a big deal, as in normal operation in CI or any other valid test run, we are not expected to have any warnings reported in pytest hooks. The warning here is for author of the test code to notice something needs to be addressed before merging.

conf/ocsci/default_config.yaml
```diff
@@ -46,6 +46,8 @@ DEPLOYMENT:
 REPORTING:
   email:
     address: "ocs-ci@redhat.com"
+  polarion:
+    project_id: "OCS"
```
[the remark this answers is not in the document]
petr-balogh: +1 for catch

ocs/defaults.py
```diff
@@ -55,6 +55,9 @@
     'email': {
         'address': 'ocs-ci@redhat.com',
     },
+    'polarion': {
+        'project_id': 'OCS',
```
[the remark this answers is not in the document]
clacroix12: Done

conf/ocsci/default_config.yaml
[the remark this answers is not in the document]
clacroix12: Done

RazTamir:
With this PR is now merged, could you please provide information, what is needed for test cases to include in order to be in sync with Polarion?

PR #245, opened 2019-06-18
Fixed test_pvc_invalid_inputs test failure

PrasadDesala:
*. Fixed cluster namespace failure
*. As this is a negative test case, changed the tier level
   from tier1 to tier3


PrasadDesala:
> LGTM but the conflict needs to be resolved.

Done

PR #249, opened 2019-06-18
Resource name enhancement

shreekarSS:
ocs/constants.py
  - Added component names for rbd and cephfs

tests/helpers.py
  - Added a variable 'component' in functions create_pvc(),create_storage_class() and create_secret() which passed to create_unique_resource_name() which creates resources.
  - Resources can be identified easily with this component name in the resource name

### DIFF
--- a/resources/pvc.py
+++ b/resources/pvc.py
@@ -4,7 +4,7 @@ General PVC object
 import logging
 
 from ocs import constants
-from ocs.defaults import ENV_DATA
+from ocs.defaults import ENV_DATA, CSI_PVC_DICT
 from ocs.ocp import OCP
 from resources.ocs import OCS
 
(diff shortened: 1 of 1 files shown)


tests/helpers.py
```diff
@@ -110,14 +112,17 @@ def create_secret(interface_type):
...
         secret_data = defaults.CSI_CEPHFS_SECRET.copy()
         del secret_data['data']['userID']
         del secret_data['data']['userKey']
         secret_data['data']['adminID'] = constants.ADMIN_BASE64
         secret_data['data']['adminKey'] = get_admin_key()
+        component = constants.CEPHFS_COMPONENT
+
     secret_data['metadata']['name'] = create_unique_resource_name(
-        'test', 'secret'
+        f'test-{component}', 'secret'
```
ebenahar: do we need the second arg 'secret'? I think we can remove this arg from create_unique_resource_name()

shreekarSS: The second arg 'secret' is    'resource_type' in create_unique_resource_name() which returns  return f"{resource_type}-{resource_description[:23]}-{current_date_time[:10]}"

If we remove second arg then, we can't identify which resource type are we creating

ebenahar: yes, I meant that in create_uniques_resource_name, to return:
```python
return f"{resource_description[:23]}-{current_date_time[:10]}"
```

ebenahar: Please disregard my above comment

PR #251, opened 2019-06-18
remove tests_config which was added with wrong assumption

vasukulkarni:
remove tests_config which was added with wrong assumption to compare defaults set in two places.


petr-balogh:
> This test doesn't compare defaults from two places - it calls the main function which initializes the configuration, then compares the result to the values found in the defaults module. It doesn't specifically care about the yaml file.

But the defaults module didn't meant to hold all the config sections defined in the config and also shouldn't contain the same values defined in [DEFAULTS] section in default_config.yaml , the [DEFAULTS] section from custom config or default one were suppose to extend values in defaults.py module but we will need to get rid of this functionality cause it's confusing the ppl a lot. I will explain about my intention once more on our next call.

zmc:
Sounds like there's still disagreement on where the single set of defaults is going to live. I've been working on the assumption that it would be the python module, but perhaps I'm wrong?

vasukulkarni:
> Sounds like there's still disagreement on where the single set of defaults is going to live. I've been working on the assumption that it would be the python module, but perhaps I'm wrong?

We agreed that it will be in defaults.py where more than 99% of those values exists now,  I am closing this in favor of #218 where its just going to be one file.

PR #258, opened 2019-06-19
Update kubernetes version to latest

shreekarSS:
- Updated kubernetes version to latest

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
         'paramiko==2.4.2',
         'pyyaml>=4.2b1',
         'jinja2==2.10.1',
-        'kubernetes<9.0.0,>=8.0.0',  # to satisfy openshift package requirement
+        'kubernetes==9.0.0',
         'openshift',
         'boto3',
         'munch',


shreekarSS:
pip show kubernetes
\Name: kubernetes
Version: 9.0.0
Summary: Kubernetes python client
Home-page: https://github.com/kubernetes-client/python
Author: Kubernetes
Author-email: 
License: Apache License Version 2.0
Location: /home/sshreeka/wip/ocs/update_kuber/ocs-ci/lib/python3.6/site-packages
Requires: certifi, websocket-client, six, requests-oauthlib, urllib3, pyyaml, requests, setuptools, python-dateutil, google-auth

setup.py
```diff
@@ -22,7 +22,7 @@
         'paramiko==2.4.2',
         'pyyaml>=4.2b1',
         'jinja2==2.10.1',
-        'kubernetes<9.0.0,>=8.0.0',  # to satisfy openshift package requirement
+        'kubernetes<=10.0.0',
```
clacroix12: I see in the comment you tested with 9.0.0. Is 10.0.0 tested as well? Also do we want `<=` or are we trying to stay up to date with kubernetes latest version from here on? Just trying to get a grasp of what we're achieving with this change.

shreekarSS: As per Vasu's comment, we are sticking to v10. I have tested with 10.0.0.  Previously with lower than 9.0.0 used to get installed which we were seeing some deprecated warnings yaml load things. All these are fixed in latest kubernetes.

clacroix12: With that in mind is `<=10.0.0` what we want? It seems like we want `>=10.0.0` or `==10.0.0`. As it is now we just ensure you aren't going past v10 which I don't think is what we want. If someone already has v8 installed for example, this change won't ensure their version is updated.

mbukatov:
Could someone elaborate what is this about? And back the claims with proper references?

mbukatov:
Ah, I see that we have [`openshift`](https://pypi.org/project/openshift/) PyPI package [specified in our `setup.py`](https://github.com/red-hat-storage/ocs-ci/blob/afb5d6d82991ebd236fe0a727cc03ecbd93d8842/setup.py#L26), and new version 0.9.0 was just released yesterday (2019-06-19), which seems to trigger this problem, correct?

Would it make more sense to pin kubernetes and/or openshift to a particular version, which matches what we have in our OCP/OCS builds? How do we verify that the versions we use here are correct wrt OCP/OCS builds?

setup.py
mbukatov: If we have a complex version constraints here, why to specify a range of versions instead of a single version?

jhutar: I had 8.0.1 and got:
```
ERROR: openshift 0.9.0 has requirement kubernetes~=9.0.0, but you'll have kubernetes 8.0.1 which is incompatible.
```
when `pip install -r requirements.txt`. Could we make this `kubernetes<=10.0.0,>=9.0.0` or so? Or maybe I'm missing something?

vasukulkarni: Last time the PR was raised latest was 9.0, now its 10.0,  @shreekarSS can you use '==' and pin this version and test, If its working for 10.0, we can stick to it.

shreekarSS: @vasukulkarni I  tried with kubernetes==10.0.0, which says 
  Could not find a version that satisfies the requirement kubernetes==10.0.0 (from ocs-ci==0.1->-r requirements.txt (line 1)) (from versions: 1.0.0a2, 1.0.0a3, 1.0.0a4, 1.0.0a5, 1.0.0b1, 1.0.0b2, 1.0.0b3, 1.0.0, 1.0.1, 1.0.2, 2.0.0a1, 2.0.0b1, 2.0.0, 3.0.0a1, 3.0.0b1, 3.0.0, 4.0.0a1, 4.0.0b1, 4.0.0, 5.0.0b1, 5.0.0, 6.0.0b1, 6.0.0, 6.1.0, 7.0.0a1, 7.0.0b1, 7.0.0, 7.0.1, 8.0.0a1, 8.0.0b1, 8.0.0, 8.0.1, 9.0.0a1, 9.0.0b1, 9.0.0, 10.0.0a1)
No matching distribution found for kubernetes==10.0.0 (from ocs-ci==0.1->-r requirements.txt (line 1))
 
What if we  do kubernetes==9.0.0? it will install the latest(9.0.0 as per https://pypi.org/project/kubernetes/)

vasukulkarni: yeah lets stick to 9.0.0,  v10 is in prerelease https://github.com/kubernetes-client/python/releases

shreekarSS: Updated with 9.0.0 @vasukulkarni

PR #259, opened 2019-06-19
Move config initialization to Config class

zmc:
This is marked as a WIP because I still want to update docs.

### DIFF
--- a/ocsci/__init__.py
+++ b/ocsci/__init__.py
@@ -8,7 +8,17 @@ under section PYTEST_DONT_REWRITE
 """
 # Use the new python 3.7 dataclass decorator, which provides an object similar
 # to a namedtuple, but allows type enforcement and defining methods.
+import collections
+import os
+import yaml
 from dataclasses import dataclass, field, fields
+
+from ocs import constants
+
+DEFAULT_CONFIG_PATH = os.path.join(
+    constants.TOP_DIR,
+    "conf/ocsci/default_config.yaml"
+)
 
 
 @dataclass
@@ -22,9 +32,33 @@ class Config:
         self.reset()
 
     def reset(self):
+        """
+        Clear all configuration data and load defaults
+        """
         for f in fields(self):
             setattr(self, f.name, f.default_factory())
-        self.RUN['cli_params'] = dict()
+        self.update(self.get_defaults())
+
+    def get_defaults(self):
+        """
+        Return a fresh copy of the default configuration
+        """
+        with open(DEFAULT_CONFIG_PATH) as file_stream:
+            return yaml.safe_load(file_stream)
+
+    def update(self, user_dict: dict):
+        """
+        Override configuration items with items in user_dict, without wiping
+        out non-overridden items
+        """
+        field_names = [f.name for f in fields(self)]
+        for k, v in user_dict.items():
+            if k not in field_names:
+                raise ValueError(
+                    f"{k} is not a valid config section. "
+                    f"Valid sections: {field_names}")
+            section = getattr(self, k)
+            merge_dict(section, v)
 
     def to_dict(self):
         # We don't use dataclasses.asdict() here, because that function appears
@@ -34,6 +68,44 @@ class Config:
         return {
             name: getattr(self, name) for name in field_names
         }
+
+
+def merge_dict(orig: dict, new: dict) -> dict:
+    """
+    Update a dict recursively, with values from 'new' being merged into 'orig'.
+
+    Args:
+        orig (dict): The object that will receive the update
+        new  (dict): The object which is the source of the update
+
+    Example:
+    orig = {
+        'dict': {'one': 1, 'two': 2},
+        'list': [1, 2],
+        'string': 's',
+    }
+    new = {
+        'dict': {'one': 'one', 'three': 3},
+        'list': [0],
+        'string': 'x',
+    }
+    merge_dict(orig, new) ->
+    {
+        'dict': {'one': 'one', 'two': 2, 'three': 3}
+        'list': [0],
+        'string', 'x',
+    }
+    """
+    for k, v in new.items():
+        if isinstance(orig, collections.Mapping):
+            if isinstance(v, collections.Mapping):
+                r = merge_dict(orig.get(k, dict()), v)
+                orig[k] = r
+            else:
+                orig[k] = v
+        else:
+            orig = {k: v}
+    return orig
 
 
 config = Config()
--- a/ocsci/main.py
+++ b/ocsci/main.py
@@ -1,18 +1,10 @@
-import collections
 import os
 
-from jinja2 import Template
 import pytest
 import yaml
 
 import ocsci
 from utility import utils
-
-
-HERE = os.path.abspath(os.path.dirname(__file__))
-OCSCI_DEFAULT_CONFIG = os.path.join(
-    HERE, "../conf/ocsci/default_config.yaml"
-)
 
 
 def get_param(param, arguments, default=None):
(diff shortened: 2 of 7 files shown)


ocsci/main.py
```diff
@@ -45,50 +43,19 @@ def init_ocsci_conf(arguments=[], default_config=OCSCI_DEFAULT_CONFIG):
     """
     custom_config = get_param('--ocsci-conf', arguments)
     cluster_config = get_param('--cluster-conf', arguments)
-    config_data = ocsci.config.to_dict()
     with open(default_config) as file_stream:
         default_config_data = yaml.safe_load(file_stream)
-    update_dict_recursively(config_data, default_config_data)
+        rendered_defaults = render_yaml_with_j2_context(default_config_data)
```
petr-balogh: Before the rendering of jinja2 was done at the end once all configs were merged, so you could use j2 variables in all config files. But now you are rendering via jinja just default_config_data and the rest of the config files loosing ability to use jinja2 there. So if we don't want to add additional logic, it should be removed from documentation and from examples of config where I mentioned those functionality.

ocsci/__init__.py
```diff
@@ -21,10 +24,26 @@ class Config:
...
+        if with_defaults:
+            self._init_defaults()
+
+    def _init_defaults(self):
+        field_names = [f.name for f in fields(self)]
+        defaults_dict = {k: getattr(defaults, k) for k in field_names}
+        self.update(defaults_dict)
+
+    def update(self, user_dict):
+        field_names = [f.name for f in fields(self)]
```
petr-balogh: add docstring please

ocsci/__init__.py
```diff
@@ -21,10 +24,26 @@ class Config:
     def __post_init__(self):
         self.reset()
 
-    def reset(self):
+    def reset(self, with_defaults=True):
```
petr-balogh: add docstring please

ocsci/__init__.py
```diff
@@ -36,4 +55,38 @@ def to_dict(self):
         }
 
 
+def merge_dict(orig: dict, new: dict) -> dict:
+    """
+    Update a dict recursively.
+
```
petr-balogh: missing Args: in the docstring, I know you are using annotations for describing types, but we agreed in our conventions that we will use google sphinx style for docstrings. Thanks

ocsci/tests/test_config.py
```diff
@@ -12,8 +12,61 @@ def reset_config(self):
...
+
+    def test_custom_conf(self):
+        user_dict = dict(
+            REPORTING=dict(email='unit@test.com'),
+            RUN=dict(log_dir='/dev/null'),
+        )
+        ocsci.config.update(user_dict)
+        assert ocsci.config.REPORTING['email'] == 'unit@test.com'
+        assert ocsci.config.RUN['log_dir'] == '/dev/null'
+        assert ocsci.config.RUN['bin_dir'] == ocs.defaults.RUN['bin_dir']
```
petr-balogh: I want to avoid of duplications of sections like RUN, ENV_DATA, REPORTING in both places defaults.py module and the default_config.yaml file. I still don't see the reason why to keep updated on both places, so I would rather get rid of  default_config.yaml file and have all those default dictionaries (Sections) in defaults.py and then those should be used in the init of the config. We have then --ocsci-config and --cluster-config parameters to allow us overwrite those sections as well WDYT?

zmc: I thought the plan was the have defaults in only one location?

ocsci/main.py
zmc: To me, it seems unnecessarily complex to allow jinja2 in configuration. If a lot of people disagree, then we should discuss it.

It's definitely on my to-do list to submit docs updates with this PR

ocsci/tests/test_config.py
petr-balogh: Yes, that what I told. All the defaults with all the sections like RUN and so on can live in defaults.py, you can use them in init of the config from defaults.py instead of from default_config.yaml as we do now. Then we can get rid of this file default_config.yaml and keep updating all the default values only in defaults.py moudule.

zmc: ok, good - I see @vasukulkarni has a WIP in #218 for that.

ocsci/main.py
petr-balogh: Yes I agree that we can simplify this and avoid using jinja templates in the rest of config files, just pointing that it needs to be documented. Cause currently it's documented in examples like you can use those jinja templates.

ocsci/tests/test_config.py
petr-balogh: #218 is just moving this DEFAULTS section away from the default_config.yaml, but as we have now all the sections defined also in defaults.py I would get rid of this default_config.yaml and as I said, we can use sections from defaults.py for init the config.  Sounds ok @vasukulkarni @clacroix12 @zmc ?

clacroix12: I don't see the point in keeping `defaults.py` either. We have our configuration object instantiated in `ocsci.__init__.py`. Why keep our default configuration values separate from where we instantiate the object?

zmc: @petr-balogh oh, I thought he was going to drop the whole file in that PR.

@clacroix12 The defaults used by `Config` are coming from `defaults.py`

clacroix12: @zmc once the things relevant to configuration are moved out I really just see a bunch of dictionaries that are loaded from yaml. I'm not really sure why these are in our `defaults` module in the first place. Maybe we can rename the module once the config stuff is out, or maybe there's a better home for these dictionaries. All I know is that the `defaults` module doesn't really have a clear purpose to me right now.

vasukulkarni: we are going to keep defaults.py as it started with a purpose and it still provides a simple way to add things for qe folks, Also folks are writing tests using current model so we dont want to keep changing it behind the scenes.

zmc:
I believe I've updated all the docs sections that need it

ocsci/tests/test_config.py
```diff
@@ -11,10 +11,68 @@ def reset_config(self):
...
-        assert env_data['rook_image'] == 'rook/ceph:master'
-        assert reporting_data['email']['address'] == 'ocs-ci@redhat.com'
-        assert run_data['bin_dir'] == './bin'
+        config_sections = ocsci.config.to_dict().keys()
+        for section_name in config_sections:
+            section = getattr(ocsci.config, section_name)
+            assert section == ocsci.config.get_defaults()[section_name]
+
+    def test_defaults_specific(self):
+        # FIXME: Is this test still useful?
```
vasukulkarni: I think its still useful to check specific to verify init worked (in case aboe keys() return emtpy)

zmc: Sure, good point - I'll just remove the FIXME there then.

clacroix12: Is there a way to verify init worked without checking specific default values? I can just see updating this test in the event that we change one of these defaults being tedious.

ocsci/__init__.py
```diff
@@ -8,8 +8,16 @@
...
 # Use the new python 3.7 dataclass decorator, which provides an object similar
 # to a namedtuple, but allows type enforcement and defining methods.
+import collections
+import os
+import yaml
 from dataclasses import dataclass, field, fields
 
+DEFAULT_CONFIG_PATH = os.path.join(
+    os.path.abspath(os.path.dirname(__file__)),
+    "../conf/ocsci/default_config.yaml"
```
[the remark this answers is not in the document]
zmc: Done!

ocsci/tests/test_main.py
```diff
@@ -15,31 +26,25 @@ def test_help(self):
         assert '--cluster-path' in result
         assert '--cluster-name' in result
 
+    @mock.patch('ocsci.main.pytest.main')
+    @mock.patch.object(config, 'update')
+    def test_no_args(self, config_update, pytest_main, testdir):
```
petr-balogh: @zmc I am not familiar with mock, so I have tried to understand this and hope I understand this test right.

This mocking the ocsci.main.pytest.main, so once you call the main.main([]) without any arguments, you are checking if `update` on mocked `update` function from `config` module was called, but as you changed logic in the main module on [line 33](https://github.com/red-hat-storage/ocs-ci/pull/259/files#diff-3e198ff054d2633195cb6203e4b9cbaaR37) and just  returning in this case you won't do any call on update function and thats the assert on line 33 is about to test if no call of update function has been done. Right?

Maybe some description of this test would be good to understand of this test.

ocsci/tests/test_main.py
```diff
@@ -15,31 +26,25 @@ def test_help(self):
...
+            '.yaml',
+            ocsci_conf='RUN: null',
+            cluster_conf='DEPLOYMENT: null',
+        ).dirname
+        main.main([
+            '--ocsci-conf', os.path.join(tempdir, 'ocsci_conf.yaml'),
+            '--cluster-conf', os.path.join(tempdir, 'cluster_conf.yaml'),
+        ])
+        assert config_update.call_args_list == [
+            mock.call(dict(RUN=None)),
```
petr-balogh: This checking just if mocked update function was called twice with mentioned dicts as I understand this test correctly? Also would be good to add some docstring to describe a test.

ocsci/tests/test_main.py
zmc: That's correct! Here, we're just making sure that `update()` isn't called unnecessarily.

ocsci/tests/test_main.py
zmc: Yes, we're making sure that `update()` was called with the correct values. It's a good idea to add docs for this, but if it's ok, I'd like to avoid invalidating the existing reviews - next time I touch this file I can make sure to add some explanation it that helps.

Issue #233, opened 2019-06-14
Downloading openshift client is broken


dahorak:
Some of the recent changes in `master` branch break the way how configuration files are loaded and processed, which leads to issue with downloading openshift client.
Fresh deployment (without already downloaded client) fails.
(removed unrelated parts)

```
$ run-ci tests/ecosystem/deployment -m deployment --cluster-name dahorak-cluster01 --cluster-path dahorak-cluster01
================================================ test session starts ================================================
platform linux -- Python 3.7.3, pytest-4.5.0, py-1.8.0, pluggy-0.12.0
cachedir: .pytest_cache
rootdir: ocs-ci, inifile: pytest.ini
plugins: reportportal-1.0.5
collected 1 item                                                                                                    

tests/ecosystem/deployment/test_ocs_basic_install.py::TestDeployment::test_deployment 

  ...

16:03:04 - MainThread - utility.utils - INFO - Downloading openshift client
FAILED                                                                                                        [100%]

===================================================== FAILURES ======================================================
__________________________________________ TestDeployment.test_deployment ___________________________________________

self = <tests.ecosystem.deployment.test_ocs_basic_install.TestDeployment object at 0x7fb538fa4ef0>

    def test_deployment(self):
        log.info("Running OCS basic installation")
        cluster_path = ENV_DATA['cluster_path']
        # Test cluster access and if exist just skip the deployment.
        if RUN['cli_params'].get('cluster_path') and OCP.set_kubeconfig(
            os.path.join(cluster_path, RUN.get('kubeconfig_location'))
        ):
            pytest.skip(
                "The installation is skipped cause the cluster is running"
            )
    
        # Generate install-config from template
        log.info("Generating install-config")
        run_cmd(f"mkdir -p {cluster_path}")
        pull_secret_path = os.path.join(
            TOP_DIR,
            "data",
            "pull-secret"
        )
    
        # TODO: check for supported platform and raise the exception if not
        # supported. Currently we support just AWS.
    
        _templating = templating.Templating()
        install_config_str = _templating.render_template(
            "install-config.yaml.j2", ENV_DATA
        )
        # Parse the rendered YAML so that we can manipulate the object directly
        install_config_obj = yaml.safe_load(install_config_str)
        with open(pull_secret_path, "r") as f:
            # Parse, then unparse, the JSON file.
            # We do this for two reasons: to ensure it is well-formatted, and
            # also to ensure it ends up as a single line.
[...]

PR #261, opened 2019-06-20
Configure PATH from configuration

dahorak:
* configure `PATH` based on configuration `RUN['bin_dir']`

### DIFF
--- a/ocsci/main.py
+++ b/ocsci/main.py
@@ -7,6 +7,7 @@ import yaml
 
 import ocs.defaults
 import ocsci
+from utility import utils
 
 
 HERE = os.path.abspath(os.path.dirname(__file__))
@@ -113,7 +114,5 @@ def main(arguments):
         '-p', 'ocsci.pytest_customization.marks',
         '-p', 'ocsci.pytest_customization.ocsci_logging',
     ])
-    from ocsci import config
-    from utility.utils import add_path_to_env_path
-    add_path_to_env_path(os.path.expanduser(config.RUN['bin_dir']))
+    utils.add_path_to_env_path(os.path.expanduser(ocsci.config.RUN['bin_dir']))
     return pytest.main(arguments)
--- a/ocsci/main.py
+++ b/ocsci/main.py
@@ -113,4 +113,7 @@ def main(arguments):
         '-p', 'ocsci.pytest_customization.marks',
         '-p', 'ocsci.pytest_customization.ocsci_logging',
     ])
+    from ocsci import config
+    from utility.utils import add_path_to_env_path
+    add_path_to_env_path(os.path.expanduser(config.RUN['bin_dir']))
     return pytest.main(arguments)
--- a/run_ocsci.py
+++ b/run_ocsci.py
@@ -1,13 +1,9 @@
-import os
 import sys
 
 import ocsci.main
 
 
 def main():
-    HERE = os.path.abspath(os.path.dirname(__file__))
-    from utility.utils import add_path_to_env_path
-    add_path_to_env_path(os.path.join(HERE, 'bin'))
     arguments = sys.argv[1:]
     exit(ocsci.main.main(arguments))
 


setup.py
```diff
@@ -33,7 +33,7 @@
     ],
     entry_points={
         'console_scripts': [
-            'run-ci=run_ocsci:main',
+            'run-ci=ocsci.main:main',
```
vasukulkarni: so run-ci should be broken in current master? @zmc the test you added is supposed to catch that right?

vasukulkarni: Ahh ignore, you moved it out.

ocsci/main.py
```diff
@@ -106,11 +107,15 @@ def render_yaml_with_j2_context(yaml_data):
...
     arguments.extend([
         '-p', 'ocsci.pytest_customization.ocscilib',
         '-p', 'ocsci.pytest_customization.marks',
         '-p', 'ocsci.pytest_customization.ocsci_logging',
     ])
-    return pytest.main(arguments)
+    from ocsci import config
+    from utility.utils import add_path_to_env_path
+    add_path_to_env_path(os.path.expanduser(config.RUN['bin_dir']))
+    exit(pytest.main(arguments))
```
vasukulkarni: return ?

dahorak: After discussion with Martin, I've decided not to remove the `run_ocsci.py` wrapper in this PR.

dahorak:
@mbukatov I had to fix small tox issue (remove `os` module from `run_ocsci.py`), sorry for the force push, you was too quick in the review.

ocsci/main.py
```diff
@@ -113,4 +113,7 @@ def main(arguments):
         '-p', 'ocsci.pytest_customization.marks',
         '-p', 'ocsci.pytest_customization.ocsci_logging',
     ])
+    from ocsci import config
```
zmc: You can simply access `ocsci.config`.

ocsci/main.py
```diff
@@ -113,4 +113,7 @@ def main(arguments):
...
+    from utility.utils import add_path_to_env_path
```
zmc: Is it possible to do this import at the module level?

dahorak: Importing `config` module (or any module which import `config` module) before config initialization (`init_ocsci_conf()` - few lines before) breaks the configuration in the way, that some values wasn't available (https://github.com/red-hat-storage/ocs-ci/issues/233).

ocsci/main.py
dahorak: good catch, let me try it

ocsci/main.py
zmc: Importing the config module isn't what causes the problem - it's code that runs at import time using values from it before it's initialized. Since this code executes *after* that, it will be fine.

mbukatov: Hmm, maybe the problem you are referencing to was partly caused by using mutable objects as defaults, fixed in 8ed5847d8024fd828821000cc8fad6a9dcbfeac9

zmc: That certainly seems plausible :)

ocsci/main.py
dahorak: Fixed in f5ccedd.

ocsci/main.py
dahorak: Fixed in f5ccedd.

ocsci/main.py
```diff
@@ -113,4 +114,5 @@ def main(arguments):
         '-p', 'ocsci.pytest_customization.marks',
         '-p', 'ocsci.pytest_customization.ocsci_logging',
     ])
+    utils.add_path_to_env_path(os.path.expanduser(ocsci.config.RUN['bin_dir']))
```
petr-balogh: If we will need to do more post init action in the future we can create some function for this purpose which will contain those and call it from here.  But I am ok for this PR, just FYI if you will add some more next time.

petr-balogh:
@zmc hope you are OK with the change, I am merging it how Daniel updated, in case you are missing something please open the issue and assign @dahorak, thanks!

PR #274, opened 2019-06-21
Resource name enhancement

shreekarSS:
ocs/constants.py

Added component names for rbd and cephfs
tests/helpers.py

Added a variable 'component' in functions create_pvc(),create_storage_class() and create_secret() which passed to create_unique_resource_name() which creates resources.
Resources can be identified easily with this component name in the resource name
Signed-off-by: Shreekar sshreeka@redhat.com

### DIFF
--- a/ocs/constants.py
+++ b/ocs/constants.py
@@ -43,3 +43,7 @@ IGNORE_SC = "gp2"
 # encoded value of 'admin'
 ADMIN_BASE64 = 'YWRtaW4='
 GB = 1024 ** 3
+
+# component
+RBD_COMPONENT = 'rbd'
+CEPHFS_COMPONENT = 'cephfs'
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -5,6 +5,8 @@ import base64
 import copy
 import datetime
 import logging
+import random
+import string
 
 from ocs import constants, defaults, ocp
 from ocsci import config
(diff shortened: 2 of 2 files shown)


shreekarSS:
test log: http://pastebin.test.redhat.com/773909

ocs/constants.py
```diff
@@ -43,3 +43,7 @@
 # encoded value of 'admin'
 ADMIN_BASE64 = 'YWRtaW4='
 GB = 1024 ** 3
+
+# component
+RBD_COMPONENT = 'rbd'
```
RazTamir: INTERFACE?
I asked ceph ppl how they used to call the ceph backend type and been told 'interface'

tests/helpers.py
```diff
@@ -215,8 +224,16 @@ def create_pvc(sc_name, pvc_name=None):
         OCS: An OCS instance for the PVC
     """
     pvc_data = copy.deepcopy(defaults.CSI_PVC_DICT)
+    if 'rbd' in sc_name:
```
RazTamir: that's not a good way to determine if that's a rbd or cephfs.
Simply pass the interface_type to the function

nehaberry:
+1 to this PR.

It would be beneficial to have the prefix "rbd" or "cephfs" in all resource names, be it storageclass, pool names, pvc names, pod names, etc.

tests/helpers.py
nehaberry: @shreekarSS  @ebenahar I have one doubt. In helpers.create_storage_class, we are using interface_type and interface_name. 
```
create_storage_class(
    interface_type, interface_name, secret_name, sc_name=None
):
...
 sc_data['parameters']['pool'] = interface_name
```

Since, both CephFS and RBD storageclaasses have similar **parameter** name called **pool** (i.e  sc_data['parameters']['pool'] ) , would it be more convenient to call **"interface_name**" as "**pool_name**" .

Please suggest

shreekarSS: I think 'pool_name' would be  more convenient @nehaberry

shreekarSS: > that's not a good way to determine if that's a rbd or cephfs.
> Simply pass the interface_type to the function

I've modified create_pvc() to have interface_type , changing the files impacted

vasukulkarni:
needs rebase due to directory structure change

clacroix12:
@shreekarSS any update on this PR? Needs to be rebased on latest master and have comments addressed.

PR #275, opened 2019-06-21
Templates update plus scripts

petr-balogh:
This patch will allow us to keep track on template files and easily update them and see
what changed and easily submit patch for new changes.

There are 2 new scripts:

* `create-patch.sh` - allow you to create new patch file
* `apply-patch.sh` - allow you to apply created patch file on new downloaded template file

### DIFF
--- a/scripts/apply-patch.sh
+++ b/scripts/apply-patch.sh
@@ -4,7 +4,7 @@
 # particular file.
 
 # If you don't provide any parameter:
-# This script look for all .patch files in ocs-ci repo and trying to download
+# This script looks for all .patch files in ocs-ci repo and tries to download
 # original file written in the first line of patch file. (Count that patch was
 # created by create-patch.sh script with labeling and contains URL)
 
--- a/scripts/apply-patch.sh
+++ b/scripts/apply-patch.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+
+# If you pass patch file as first parameter this script apply patch to just
+# particular file.
+
+# If you don't provide any parameter:
+# This script look for all .patch files in ocs-ci repo and trying to download
+# original file written in the first line of patch file. (Count that patch was
+# created by create-patch.sh script with labeling and contains URL)
+
+set -e
+
+script_dir=$(dirname "$0")
+echo $script_dir
+download_and_patch_file () {
+    patch_file=$1
+    file_name=${patch_file%".patch"}
+    rm -f $file_name
+    read -r download_link<$patch_file
+    download_link=$(echo  $download_link | cut -d " " -f 2)
+    echo "Downloading new version of file: $file_name from source: $download_link"
+    wget $download_link -O $file_name
+    patch $file_name < $patch_file
+    echo "File: $file_name was sucessfully patched"
+}
+
+PATCH_FILE=$1
+
+if [ -n "$PATCH_FILE" ]; then
+    download_and_patch_file "$PATCH_FILE"
+else
+    # If you will move the script to some different location you will need to
+    # change following line!
+    pushd $script_dir/../
+    find ./ -name '*.patch' | while read patch_file
+    do
+        echo "Found patch file: $patch_file"
+        download_and_patch_file "$patch_file"
+    done
+    popd
+fi
--- a/templates/README.md
+++ b/templates/README.md
@@ -17,7 +17,7 @@ and applying patches.
 
 Once we will have all patch files we can just run the script
 [scripts/apply-patch.sh](../scripts/apply-patch.sh) and see what was changed
-via `git diff` command  and we can send PR with latest changes to our
+via `git diff` command and you can send PR with latest changes to our
 templates. Advantage is also that if there will be some conflict, the script
 `apply-patch.sh` will fail and we will need to solve the conflict.
 
--- a/templates/README.md
+++ b/templates/README.md
@@ -1,0 +1,36 @@
+# Templates
+
+In this folder you can find our templates used in the tests and deployment.
+
+If you are creating some new template from some source like
+[rook examples](https://github.com/rook/rook/tree/master/cluster/examples/kubernetes/ceph)
+please create the patch file for this template in the same path with the same
+name + add `.patch` suffix. You can see one example of patch file
+[here](ocs-deployment/cluster.yaml.patch) which was created by
+[scripts/create-patch.sh](../scripts/create-patch.sh) script.
+
+It will provide us the way to keep track of changes of our template
+files till we will find some better solution.
+
+See the the [documentation](../scripts/README.md) for new scripts for creating
+and applying patches.
+
+Once we will have all patch files we can just run the script
+[scripts/apply-patch.sh](../scripts/apply-patch.sh) and see what was changed
+via `git diff` command  and we can send PR with latest changes to our
+templates. Advantage is also that if there will be some conflict, the script
+`apply-patch.sh` will fail and we will need to solve the conflict.
+
+In case of new changes provided, you need to submit the patch for new changes
+and you need to create new patch file to update the last one to be able to
+watch new changes!
+
+The intention is to have some job which will run maybe once a day and will send
+the notification in case of new changes, so then we cannot manually watch all
+the changes.
+
+If no change is done in the template and no need to  it's worth to at least
+the mention the source URL of the template or link to documentation in the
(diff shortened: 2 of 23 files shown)


mbukatov:
I would rather get rid of the jinja2 templates, which would allow us to keep the original yaml files with minimal changes.

petr-balogh:
> I would rather get rid of the jinja2 templates, which would allow us to keep the original yaml files with minimal changes.

@mbukatov  it's not only about jinja2 templates, I mean template whole manifest file which we are copy here, and I think that ppl are doing some minimal changes like changing namespace to openshift-storage.

This way we can just run the script apply-patch.sh which will fetch all new changes and you will directly see what changed and you can just create new patch.

Can you propose other way how to track those changes in state how we have it now as part of our repository?  In the ideal world I would like to get those manifests/templates from engineering so then we don't have to maintain them but there are a lot of changes coming in rook repository for files we are using and we need to keep track on those currently and update them.

mbukatov:
> it's not only about jinja2 templates, I mean template whole manifest file which we are copy here, and I think that ppl are doing some minimal changes like changing namespace to openshift-storage.

@petr-balogh I understand your point with patches, which are a way to allow rebasing of the yaml based jinja2 templates without actually rebasing the git repository. But that is necessary under assumption that there will be lot of changes in these files.

I imagine that without converting the yaml files into jinja2 templates, there would be very few changes in these files (ideally do no changes in most cases), which would mean that the heavy change tracking (on a level of a rebase helper tool) won't be necessary.

> Can you propose other way how to track those changes in state how we have it now as part of our repository?

When adding the file, I will commit the file without any changes from the original source, linking the source url, and commit id in the text of a commit. We may also consider adding the url in the 1st line as a comment.

Then I can add a simple change (like a change of rook-ceph namespace to openshift-storage) as another commit. Such changes should be very rare.

When I need to update the file based on the new version, I will repeat the process.

The commit graph for a particular file would look like (latest commit on the top):

```
* toolbox_pod: change of default namespace
* refreshing toolbox_pod.yaml file 
* toolbox_pod: change of default namespace
* adding toolbox_pod.yaml file 
```

When I would like to inspect changes and origin of the file, I will just use standard git tools, like `git log` or `git blame`.

That said, I agree that there is a risk that we will fail to keep the changes minimal as possible, and fail to follow this simple process. And that the more automated solution could be useful.

The concerns above makes me think that the best way to address them is to keep the files in original state without any changes, only referencing the source. We could add a check which will verify this property during tox run.

Another possibility is to refer to the files without copying it into our repo. But I don't know how many sources we have.

If there are few changes in a particular file which we need to do in most or all cases, this doesn't matter as the test code won't use the yaml file directly anyway. I image unitity function or fixture which will create dictionary for the test based on the yaml file. See also: https://github.com/red-hat-storage/ocs-ci/issues/276

templates/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+It will provide us the way to keep track of changes of our template
+files till we will find some better solution.
+
+See the the [documentation](../scripts/README.md) for new scripts for creating
+and applying patches.
+
+Once we will have all patch files we can just run the script
+[scripts/apply-patch.sh](../scripts/apply-patch.sh) and see what was changed
+via `git diff` command  and we can send PR with latest changes to our
```
vasukulkarni: s/and we/and you

vasukulkarni:
> > it's not only about jinja2 templates, I mean template whole manifest file which we are copy here, and I think that ppl are doing some minimal changes like changing namespace to openshift-storage.
> 
> @petr-balogh I understand your point with patches, which are a way to allow rebasing of the yaml based jinja2 templates without actually rebasing the git repository. But that is necessary under assumption that there will be lot of changes in these files.
> 
> I imagine that without converting the yaml files into jinja2 templates, there would be very few changes in these files (ideally do no changes in most cases), which would mean that the heavy change tracking (on a level of a rebase helper tool) won't be necessary.
> 
> > Can you propose other way how to track those changes in state how we have it now as part of our repository?
> 
> When adding the file, I will commit the file without any changes from the original source, linking the source url, and commit id in the text of a commit. We may also consider adding the url in the 1st line as a comment.
Can you elaborate on how to do this with current structure?

> 
> Then I can add a simple change (like a change of rook-ceph namespace to openshift-storage) as another commit. Such changes should be very rare.
> 
> When I need to update the file based on the new version, I will repeat the process.
> 
> The commit graph for a particular file would look like (latest commit on the top):
> 
> ```
> * toolbox_pod: change of default namespace
> * refreshing toolbox_pod.yaml file 
> * toolbox_pod: change of default namespace
> * adding toolbox_pod.yaml file 
> ```
> 
> When I would like to inspect changes and origin of the file, I will just use standard git tools, like `git log` or `git blame`.
> 
> That said, I agree that there is a risk that we will fail to keep the changes minimal as possible, and fail to follow this simple process. And that the more automated solution could be useful.
> 
> The concerns above makes me think that the best way to address them is to keep the files in original state without any changes, only referencing the source. We could add a check which will verify this property during tox run.
> 
> Another possibility is to refer to the files without copying it into our repo. But I don't know how many sources we have.
> 
> If there are few changes in a particular file which we need to do in most or all cases, this doesn't matter as the test code won't use the yaml file directly anyway. I image ~unitity function or~ fixture which will create dictionary for the test based on the yaml file (via utility function). See also: #276

clacroix12:
> Another possibility is to refer to the files without copying it into our repo. But I don't know how many sources we have.

This was my thought as well. We provide a way for people to manipulate the data they get from these yaml files through python, why do we need to alter them from their original state in yaml? Shouldn't we should just be pulling from the source so we do not need to maintain a copy of them in our repository, and any changes that need to happen past it's default state can happen once the file is read into a dictionary?

scripts/apply-patch.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+
+# If you pass patch file as first parameter this script apply patch to just
+# particular file.
+
+# If you don't provide any parameter:
+# This script look for all .patch files in ocs-ci repo and trying to download
```
ebenahar: Replace:
'look' -> 'looks'
'and trying' -> 'and tries'

templates/README.md
petr-balogh: done

scripts/apply-patch.sh
petr-balogh: done

petr-balogh:
> > Then I can add a simple change (like a change of rook-ceph namespace to openshift-storage) as another commit. Such changes should be very rare.
> > When I need to update the file based on the new version, I will repeat the process.
> > The commit graph for a particular file would look like (latest commit on the top):
> > ```
> > * toolbox_pod: change of default namespace
> > * refreshing toolbox_pod.yaml file 
> > * toolbox_pod: change of default namespace
> > * adding toolbox_pod.yaml file 
> > ```
> > 
> > 
> > When I would like to inspect changes and origin of the file, I will just use standard git tools, like `git log` or `git blame`.
> > That said, I agree that there is a risk that we will fail to keep the changes minimal as possible, and fail to follow this simple process. And that the more automated solution could be useful.

What I am proposing here is working just with current situation. I am not going as part of this PR to change the logic of current state and will just provide the way how to simplify updates of those manifests till we will move to some more convenient solution. 

The problem I see here is that if you download new file and and you try to apply all the commits which we will have in git tree on top of this file that you will end up with conflicts and you will need to solve them and then maybe you will have another conflict in other commit as well and this is not comfortable. Way and automation will fail on this I think. 

> > The concerns above makes me think that the best way to address them is to keep the files in original state without any changes, only referencing the source. We could add a check which will verify this property during tox run.
> > Another possibility is to refer to the files without copying it into our repo. But I don't know how many sources we have.
> > If there are few changes in a particular file which we need to do in most or all cases, this doesn't matter as the test code won't use the yaml file directly anyway. I image ~unitity function or~ fixture which will create dictionary for the test based on the yaml file (via utility function). See also: #276
I think that

petr-balogh:
Mainly aligning for latest changes: https://github.com/rook/rook/pull/3324/files#diff-c95a038077e874ba304e11f95e7c02c8

We will need to agree on some approach how to take those files (But not as part of this PR). For deployment we need to change a lot of values like values also in lists which contain rook-ceph namespace on a lot of places not just key: value pair but it's part of string values like here: https://github.com/red-hat-storage/ocs-ci/pull/275/files#diff-908e6b99b72d8cbd4c2acdd08a6c1625

PR #277, opened 2019-06-21
Utility cleanup

clacroix12:
Another pass at cleaning up unused code, this time mostly focusing on the `utilities` module. There is still some more work to be done here. @vasukulkarni I will most likely need your help on identifying what we would like to keep here that was brought over from ceph. I'm not sure what we will want to get rid of vs. what we will want to adapt to our needs here. Same goes for `ocs/utils.py` and `ocs/ceph.py` but I will raise this in a new issue. 

I tried isolating most of the removals in their own commits so we can back off if I removed something I shouldn't have just yet. Hopefully this helps parsing the removals a bit easier as well.


petr-balogh:
@clacroix12 thanks, feel free to merge once you online

PR #278, opened 2019-06-21
ensure that the given --cluster-path if empty if --deploy is also provided

andrewschoen:
This is to guard against accidentally writing over an existing clusters information if for whatever reason it could not be accessed temporarily.

Also, I'm adding an ``ocsci.exceptions`` module and ``utility.system`` module. The ``ocsci.exceptions`` module we can use to store custom exceptions which can provide more meaningful exception messages back to the user. I was thinking the ``utility.system`` could contain system-level utility methods, for example file system modifications. Eventually we can port stuff from ``utility.utils`` that make sense to do so. 

This is a followup from #173

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,7 +12,7 @@ from oc.openshift_ops import OCP
 from ocs import constants, ocp, defaults
 from ocs.exceptions import CommandFailed, CephHealthException
 from ocs.utils import create_oc_resource, apply_oc_resource
-from utility import templating
+from utility import templating, system
 from utility.aws import AWS
 from utility.retry import retry
 from utility.utils import destroy_cluster, run_cmd, get_openshift_installer, get_openshift_client, is_cluster_running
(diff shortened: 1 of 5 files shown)


andrewschoen:
I'll address the failing tests, but wanted to get this posted early for review.

utility/system.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+
+def is_path_empty(path):
```
clacroix12: I know it's pretty self explanatory but we should still have some pydoc for this function.

zmc:
I appreciate the addition of new modules here

tests/conftest.py
```diff
@@ -58,6 +58,10 @@ def cluster(request):
         msg = "The given cluster can not be connected to: {}. ".format(cluster_path)
         msg += "Provide a valid --cluster-path or use --deploy to deploy a new cluster"
         pytest.fail(msg)
+    elif not system.is_path_empty(cluster_path) and deploy:
+        msg = "The given cluster path is not empty: {}. ".format(cluster_path)
```
petr-balogh: you can just format the string like

```python
msg = (
    f"The given cluster path is not empty: {cluster_path}. "
    f"Provide an empty --cluster-path and --deploy to deploy a new cluster"
)
```
This is the preferred way from our contribution [docs](https://github.com/red-hat-storage/ocs-ci/blob/master/docs/coding_guidelines.md).

Don't you want to mention why the directory has to be empty? Is it cause you would like to prevent to override the data which was generated via the installer and prevent of their lost?

andrewschoen: @petr-balogh sure thing, I'll format it that way next time. Would you like me to modify this message here? I'm thinking perhaps the docs can explain more about this?

petr-balogh: I don't want to block this PR, if it's ready for merge and verified please merge this. And docs can be improved in some next PR if you feel it's worth to mention it.

PR #286, opened 2019-06-24
Make it possible to run tests/test_api_client.py::test_create_simple_service >once

jhutar:


tests/test_api_client.py
```diff
@@ -29,15 +30,15 @@ def test_create_simple_service():
...
+      name: myservice-%s
     spec:
       selector:
         app: MyApp
       ports:
         - protocol: TCP
           port: 8089
           targetPort: 9369
-    """
+    """ % int(time.time() * 1000)
```
clacroix12: We shouldn't be using this method of string formatting, please use f-strings instead. Check out the [coding guidelines](https://github.com/red-hat-storage/ocs-ci/blob/master/docs/coding_guidelines.md) for some examples.

shylesh:
> Ah, I think that this test case won't probably have any link to polarion either so maybe we can delete it if it's not a valid test. @shylesh is this some valid TC defined somewhere? Or this is like unittest for testing if client is working? I am not sure if we should run this kind of testcases in our executions.
> 
> We will need to discuss this topic on our next automation call on Thursday. Adding it to agenda.
> 
> So first problem I see that this TC doesn't teardown resources it created. So rather than add some timestamp, please change the logic here and teardown the resource properly in the fixture.
> 
> @jhutar if you will run some tests, you can run with parameter -m tier1 to trigger just tier1 tests or any other tear if you want.
> 
> Thanks

@petr-balogh ,
No, this doesn't have any polarion test. API-client is unmaintained code now, do we have to keep this or remove it ?

petr-balogh:
> > Ah, I think that this test case won't probably have any link to polarion either so maybe we can delete it if it's not a valid test. @shylesh is this some valid TC defined somewhere? Or this is like unittest for testing if client is working? I am not sure if we should run this kind of testcases in our executions.
> > We will need to discuss this topic on our next automation call on Thursday. Adding it to agenda.
> > So first problem I see that this TC doesn't teardown resources it created. So rather than add some timestamp, please change the logic here and teardown the resource properly in the fixture.
> > @jhutar if you will run some tests, you can run with parameter -m tier1 to trigger just tier1 tests or any other tear if you want.
> > Thanks
> 
> @petr-balogh ,
> No, this doesn't have any polarion test. API-client is unmaintained code now, do we have to keep this or remove it ?

Let' discuss on our next call, already added to agenda. For now submit the PR to fix this test to properly teardown. This PR can be closed then I think

vasukulkarni:
@petr-balogh @shylesh I have raised issue 293 to address this, we dont have to remove it since its still helpful to test out libraries.

tests/test_api_client.py
jhutar: Hello. Thanks and please check now!

clacroix12:
Closing this out as it's been a month. If this is a change we would still want feel free to reopen, rebase/address merge conflicts, and run tests to verify.

PR #287, opened 2019-06-25
In some cases command with stderr is perfectly OK

jhutar:
I have spent some time investigating error in `tests/test_pv_creation.py` output:

```
08:55:27 - MainThread - tests.test_pv_creation - INFO - Deleting the Persistent Volume my-pv1
08:55:27 - MainThread - utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig ../dev-scripts/ocp/auth/kubeconfig delete -f test.yaml
08:55:27 - MainThread - utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig ../dev-scripts/ocp/auth/kubeconfig get PersistentVolume my-pv1 -o yaml
08:55:27 - MainThread - utility.utils - ERROR - CMD error:: Error from server (NotFound): persistentvolumes "my-pv1" not found
08:55:27 - MainThread - tests.test_pv_creation - INFO - PV my-pv1 doesn't exist
```

where test is asserting that PV was deleted, but `run_cmd` emits error. I think that is confusing and that it is up to test to assert and log error only if it is error.

### DIFF
--- a/utility/utils.py
+++ b/utility/utils.py
@@ -650,7 +650,10 @@ def run_cmd(cmd, **kwargs):
     )
     log.debug(f"CMD output: {r.stdout.decode()}")
     if r.stderr:
-        log.error(f"CMD error:: {r.stderr.decode()}")
+        if r.returncode == 0:
+            log.debug(f"CMD error: {r.stderr.decode()}")
+        else:
+            log.error(f"CMD error: {r.stderr.decode()}")
     if r.returncode:
         raise CommandFailed(
             f"Error during execution of command: {cmd}."


utility/utils.py
```diff
@@ -649,8 +649,7 @@ def run_cmd(cmd, **kwargs):
         **kwargs
     )
     log.debug(f"CMD output: {r.stdout.decode()}")
-    if r.stderr:
-        log.error(f"CMD error:: {r.stderr.decode()}")
+    log.debug(f"CMD error: {r.stderr.decode()}")
```
petr-balogh: I think that 'if' statement can remain as no need to write to log if no stderr exist. WDYT?

clacroix12:
I'm not sure I agree with this change. If we encounter an error when executing a command we should log it by default. I would rather see us pass a particular keyword to `run_cmd` that signals an error is expected and to log stderr at debug level.

Something like this:
```python
    err_msg = f"CMD error:: {r.stderr.decode()}"
    if r.stderr and kwargs.get('expect_error'):
        log.debug(err_msg)
    elif r.stderr:
        log.error(err_msg)
```

This way you could pass `expect_error=True` to `run_cmd` if you didn't want to log stderr at error level, yet remaining calls to `run_cmd` would remain unaffected. It doesn't need to actually be `expect_error` but I think utilizing kwargs in this case is better than changing behavior across the board, especially when this function is already so heavily used.

vasukulkarni:
We should raise CommandFailed exception in library and catch in the test and handle it, or we could also use the kwargs as Coady mentioned.  We have been using assertion error inside libraries, which I think is bad practise because everything falls under one category.

clacroix12:
> We should raise CommandFailed exception in library and catch in the test and handle it, or we could also use the kwargs as Coady mentioned. We have been using assertion error inside libraries, which I think is bad practise because everything falls under one category.

I don't think this functionality is changing, just the logging of the errors that eventually raise the Exception. Totally agree we should continue raising the exception and if the test is expecting it, that's what try/except blocks are made for.

utility/utils.py
vasukulkarni: The only change I see is removing the *if* statement and I agree there is no need for any change here.

clacroix12: @vasukulkarni `log.error` -> `log.debug`

RazTamir: @clacroix12 @petr-balogh @vasukulkarni I think there is no need for this change for several reasons:
1) We are the only QE that will analyze those tests and this is well known that in case of an error, we should use try: except:
2) there is nothing harm in printing the error to the console - the error is real and if we are expecting for error it doesn't mean we should log it in debug level

mbukatov: @petr-balogh 
> I think that 'if' statement can remain as no need to write to log if no stderr exist. WDYT?

Explicit loogging of empty stderr in debug level is more reasonable and easier to read for people who doesn't know the internals of the utils here.

mbukatov: @RazTamir I **strongly disagree** with you here. The code which we are talking about is in general ` run_cmd()` function, and for this reason, we don't know if stderr output can be interpreted int as an error. This could be done in a specific context only.

The argument that since we are the only ones who reads these logs doesn't mean it's good that the choices in logging doesn't make sense. Neither it's an evidence that current design is helpful.

On the other hand, the current design is not useful in the other use case neither, when I'm writing a code which should fail if stderr is present, because it's current design doesn't allow me to do that.

This all means that currently, we will see ERROR log line reported for stderr in both cases:

 * when this is expected or doesn't indicate any error for the test
 * when this is a problem, actual error

It should be clear that this clearly is not a good design.

mbukatov: This mean that I can't use this function for writing a tests anyway.

jhutar:
Hello. I do not insist on this at all.

Mine thinking was pytest log levels should be used to express test situations (test failed, passed, or debug when some info is needed only when trying to understand what is going below the surface). Fact that some command printed on stderr might be perfectly OK from test point of view.

Also I agree that `run_cmd` besides `stdout` should also return `stderr` and `exit_code` (and maybe even something like `duration`...?) as the logic on what is good and what is bad should be in the test. Of course some `run_cmd_no_stderr` and `run_cmd_rc_0` wrappers (wit better names obviously :-)) would be handy.

Anyway, I created this PR just because I was confused and because of the **ERROR** in the log, I have missed the test actually passed, so please feel free just to reject it.

utility/utils.py
petr-balogh: I think that we can have another function which will return just tuple of (rc, stdout, stderr) and not raise any exception and will allow TC to process those data with specific context and react on RC, OUT and ERR as they need. @mbukatov is this what you mean?

RazTamir: Hi @mbukatov ,
I'll be happy to have a BJ meeting on that PR as I don't fully understand the issues you are facing

mbukatov: I quickly discussed this with @RazTamir, and we reached to following suggestion for this pull request to be merged:

* when the retcode is nonzero, report the stderr using ERROR level
* when the retcode is zero, log the stderr using DEBUG level

This will address the problem that we sometimes report ERROR log lines when it's not appropriate from the test case point of view. And on the other hand, it will keep reporting ERROR log lines when there is most likely some problem.

I understand that the current design is not ideal, and one solution could be to have another function as @petr-balogh proposes in https://github.com/red-hat-storage/ocs-ci/pull/287#discussion_r298087904. That said, I would like to find a minimal way for this pull request to be merged now, so that we improve the code a bit right immediately.

mbukatov:
@jhutar 

> Mine thinking was pytest log levels should be used to express test situations (test failed, passed, or debug when some info is needed only when trying to understand what is going below the surface). Fact that some command printed on stderr might be perfectly OK from test point of view.

I agree with you.

> Also I agree that `run_cmd` besides `stdout` should also return `stderr` and `exit_code` (and maybe even something like `duration`...?) as the logic on what is good and what is bad should be in the test. Of course some `run_cmd_no_stderr` and `run_cmd_rc_0` wrappers (wit better names obviously :-)) would be handy. And that would be bigger change than plain `log.error` -> `log.debug` change.

Right. But I think that improvements of the design are out of scope of your simple fix, and could be done later after some additional thinking (and other changes such as https://github.com/red-hat-storage/ocs-ci/pull/301which are going on right now).

> Anyway, I created this PR just because I was confused and because of the **ERROR** in the log, I have missed the test actually passed, so please feel free just to reject it.

I discussed this with @RazTamir (details in https://github.com/red-hat-storage/ocs-ci/pull/287#discussion_r298236513), and our suggestion is:

* when the retcode is nonzero, report the stderr using ERROR level
* when the retcode is zero, log the stderr using DEBUG level

The idea is to address the concerns above, but keep the fix as simple as possible, so that we can merge it soon. Does that make sense to you? I don't want to deter you from proposing simple fixes like that, and neither I want to block it requesting full redesign as noted above.

utility/utils.py
RazTamir: > I quickly discussed this with @RazTamir, and we reached to following suggestion for this pull request to be merged:
> 
> * when the retcode is nonzero, report the stderr using ERROR level
> * when the retcode is zero, log the stderr using DEBUG level
WARNING level not DEBUG
so something like:
```python
if r.stderr and not r.returncode:
    log.warning(f"CMD error: {r.stderr.decode()}")
elif r.returncode :
    raise CommandFailed()
```
> 
> This will address the problem that we sometimes report ERROR log lines when it's not appropriate from the test case point of view. And on the other hand, it will keep reporting ERROR log lines when there is most likely some problem.
> 
> I understand that the current design is not ideal, and one solution could be to have another function as @petr-balogh proposes in [#287 (comment)](https://github.com/red-hat-storage/ocs-ci/pull/287#discussion_r298087904). That said, I would like to find a minimal way for this pull request to be merged now, so that we improve the code a bit right immediately.

jhutar: Well, actually, when we are talking about this and you decided to go this way: if RC is 0 and there is an stderr, that is suspicious and might deserve ERROR log lvl, while if RC is >0, having stderr is pretty normal/expected.

jhutar:
Hello @mbukatov . I mostly wanted to report mine confusion, so whatever works for you, works for me. Thank you! I agree.

jhutar:
Pushed new version of path as per your proposal. Thank you in advance.

RazTamir:
Hi @jhutar ,
I added a comment earlier but not sure why it is not here so, I suggested a small fix based on the discussion:
```python
if r.stderr and not r.returncode:
    log.warning(f"CMD warning:: {r.stderr.decode()}")
elif r.returncode:
    raise CommandFailed(...)
```

jhutar:
Hello @RazTamir . I think that would cause this:
```
returncode | stderr | stderr printed?        | return
-----------+--------+------------------------+----------------
0          | ""     |                        | return stdout
0          | "xxx"  | printed in warning lvl | return stdout
1          | ""     |                        | CommandFailed()
1          | "xxx"  | not printed            | CommandFailed()
```
(see last row when stderr would be hidden)

RazTamir:
stderr will be printed in the exception:
```python
raise CommandFailed(
    f"Error during execution of command: {cmd}."
)
```

Am I missing something?

jhutar:
Oh, got it. Depends on what is in:

```
elif r.returncode:
    raise CommandFailed(...)
```

I think. Anyway, looks like there is quite precise idea on how the patch should look like, but I can not credit myself for it now, so please feel free to commit it. Thank you!

Issue #125, opened 2019-05-30
openshift-storage namespace is getting overriden even if we have user-defined namespace


ramkiperiy:
Updated custom namespace in 2 locations as below.

1. conf/ocsci/default_config.yaml ==> ROOK_CLUSTER_NAMESPACE: "rook-ceph"
2. ocs/defaults.py ==> ROOK_CLUSTER_NAMESPACE = 'rook-ceph'

After editing in above 2 places still my oc commands are generated with "openshift-storage" namespace.

This param should be overridden from one location.

Workaround: This is not a good suggestion but works for executing tests.
defaults.ROOK_CLUSTER_NAMESPACE = 'rook-ceph'  ==> in the test_case.py

`self = <tests.test_ocs_260.TestCaseOCS260 object at 0x7f7216e1f048>

    def test_run_ocs_260(self):
        """
        Test case OCS-260
    
        """
        pvc_count = 4
        pv_list = []
    
>       create_ceph_block_pool()

tests/test_ocs_260.py:400: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/test_ocs_260.py:103: in create_ceph_block_pool
    assert CBP.create(yaml_file=TEMP_YAML)
../../ocs-ci/ocs/ocp.py:119: in create
    return self.exec_oc_cmd(command)
../../ocs-ci/ocs/ocp.py:67: in exec_oc_cmd
    out = run_cmd(cmd=oc_cmd)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cmd = ['oc', '-n', 'openshift-storage', '--kubeconfig', '/home/rperiyas/Documents/tools/OCS4x/auto_frame/cluster/auth/kubeconfig', 'create', ...]
kwargs = {}
r = CompletedProcess(args=['oc', '-n', 'openshift-storage', '--kubeconfig', '/home/rperiyas/Documents/tools/OCS4x/auto_fra...oes not match the namespace "openshift-storage". You must pass \'--namespace=rook-ceph\' to perform this operation.\n')

    def run_cmd(cmd, **kwargs):
        """
        Run an arbitrary command locally
    
        Args:
            cmd (str): command to run
[...]

Issue #294, opened 2019-06-25
Change of project before the execution will break a lot of tests.


petr-balogh:
Found the issue that if you change project to default or some other before execution like `oc project default` than some tests will fail as you can see:

```python
test_cephfilesystem_creation
tests.test_add_mds_to_cluster.TestCephFilesystemCreation
Time Taken: 56.03s
Failed: test teardown failure
def finalizer():
>       teardown(self)

tests/test_add_mds_to_cluster.py:30: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/test_add_mds_to_cluster.py:50: in teardown
    self.fs_data = CEPHFS.get(defaults.CEPHFILESYSTEM_NAME)
ocs/ocp.py:102: in get
    return self.exec_oc_cmd(command)
ocs/ocp.py:66: in exec_oc_cmd
    out = run_cmd(cmd=oc_cmd)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cmd = ['oc', '--kubeconfig', '/Users/pbalogh/tmp/cluseters/pbalogh-us3-master/auth/kubeconfig', 'get', 'CephFileSystem', 'ocsci-cephfs', ...]
kwargs = {}
r = CompletedProcess(args=['oc', '--kubeconfig', '/Users/pbalogh/tmp/cluseters/pbalogh-us3-master/auth/kubeconfig', 'get',...urncode=1, stdout=b'', stderr=b'Error from server (NotFound): cephfilesystems.ceph.rook.io "ocsci-cephfs" not found\n')

    def run_cmd(cmd, **kwargs):
        """
        Run an arbitrary command locally
    
        Args:
            cmd (str): command to run
    
        Raises:
            CommandFailed: In case the command execution fails
    
        Returns:
            (str) Decoded stdout of command
    
        """
        log.info(f"Executing command: {cmd}")
        if isinstance(cmd, str):
            cmd = shlex.split(cmd)
        r = subprocess.run(
            cmd,
[...]

PR #295, opened 2019-06-25
Move default_config.yaml to python defaults module

petr-balogh:


petr-balogh:
@zmc @vasukulkarni @clacroix12  this was discussed on last automation call that we will get rid of default_config.yaml file.

Also I am moving all usages of ROOK_CLUSTER_NAMESPACE and using cluster_namespace from config.ENV_DATA which should corresponding all the time with one from config.

I think that it should also fix issue #125 .

Anyway I think that in defaults.py we still had a lot of constants so I moved them there.

Struggling with verification cause broken leftovers check which failing tests and also some other odds errors which I see even in master. Also hit the issue https://github.com/red-hat-storage/ocs-ci/issues/294 while verification which also caused a lot of problems. So still need to proper verify.

ocs/defaults.py
```diff
@@ -7,19 +7,47 @@
...
-CLIENT_VERSION = INSTALLER_VERSION
-ROOK_CLUSTER_NAMESPACE = 'openshift-storage'
-KUBECONFIG_LOCATION = 'auth/kubeconfig'  # relative from cluster_dir
-API_VERSION = "v1"
-CEPHFILESYSTEM_NAME = 'ocsci-cephfs'
+
+#############################################################################
+# Default Configuration - access those vars from config and not from here!  #
+#############################################################################
+ENV_DATA = {
```
vasukulkarni: Instead of constants, As per some discussion last week, I think few said they belong to *defaults* and many of the current entries in *defaults* needs a new file name.

conf/README.md
```diff
@@ -81,12 +81,18 @@ from ocs import defaults
 
 # From you code you can access those data like
 
-Taking data from ENV_DATA will always use right cluster_namespace passed via
-`--ocsci-conf` config file or default one defined in `default_config.yaml`.
-function_which_taking_namespace(namespace=config.ENV_DATA['cluster_namespace'])
+# Taking data from ENV_DATA will always use right cluster_namespace passed via
+#`--ocsci-conf` or `--cluster-conf` config files.
+
+# It's not possible to access config data from module level as config is not
```
vasukulkarni: I would reword since its conveying different meaning, maybe just "config data should not be accessed at module level and instead should be used within function definitions or fixtures"

zmc:
Seems we're still not quite in the same page. Several of us met last week and [discussed this](https://github.com/red-hat-storage/ocs-ci/pull/218#pullrequestreview-252938640)

Let's cover this in tomorrow's meeting!

petr-balogh:
> Seems we're still not quite in the same page. Several of us met last week and [discussed this](https://github.com/red-hat-storage/ocs-ci/pull/218#pullrequestreview-252938640)
> 
> Let's cover this in tomorrow's meeting!

things why I wanted to get rid of yaml file are:
- I thought that we discussed this last week probably on our different mtg that you had later on with Vasu
- that ppl are confused and trying to put this default config to --ocsci-conf as a default config if there won't be such file ppl won't import it
- we don't need to have extra layer of loading yaml file if it will be in python module
- we can use some python expressions if needed
- values can be accessed directly if needed from some reason

If ppl want to keep the defautl_config.yaml I can change the PR to keep it and will have just the rest of changes which take care about using namespace from config instead of defaults.py and move the rest of constant values to constants and deleting some unused vars.

ocs/defaults.py
petr-balogh: @vasukulkarni not sure I follow here. I can move this config related default to diff module, or we can decide to have it still in yaml file, see my comment below. And discuss probably tomorrow.

conf/README.md
petr-balogh: done in next commit

clacroix12:
@petr-balogh is this something we want to move forward with or should we close?

petr-balogh:
> @petr-balogh is this something we want to move forward with or should we close?

Lowered priority here as I dealt with more priority tasks.

I will close this one and will open new one PR later on.

PR #299, opened 2019-06-26
Fix issue with error while testing version of oc

petr-balogh:
Sometime there is an issue while installing new cluster:
./bin/oc version
Client Version: version.Info{Major:"4", Minor:"1+",
GitVersion:"v4.1.2-201906121056+3569a06-dirty", GitCommit:"3569a06",
GitTreeState:"dirty", BuildDate:"2019-06-12T15:47:26Z", GoVersion:"go1.11.5",
 Compiler:"gc", Platform:"linux/amd64"}
Unable to connect to the server: dial tcp:
lookup api.ksandha-85879.qe.rh-ocs.com on 10.75.5.25:53: no such host

It somehow remember the old cluster which is not available yet.

### DIFF
--- a/utility/utils.py
+++ b/utility/utils.py
@@ -529,8 +529,13 @@ def get_openshift_client(
         delete_file(tarball)
         # return to the previous working directory
         os.chdir(previous_dir)
-
-    client_version = run_cmd(f"{client_binary_path} version")
+    try:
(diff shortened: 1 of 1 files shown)


petr-balogh:
We need to find other way how to deal with this kubeconfig. I found that we can print config with

`oc config view` and then we should probably do oc config delete-cluster cluster-name

PR #311, opened 2019-06-28
Add autouse fixture for polarion project ID

petr-balogh:
- Auto use fixture added
- Fixed project-id property name
- Added custom polarion property for description to add jenkins build URL


petr-balogh:
Tested in our jobs and worked well

PR #313, opened 2019-06-28
Add option to return raw_output from oc run_cmd

vasukulkarni:

### DIFF
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -43,13 +43,17 @@ class OCP(object):
     def namespace(self):
         return self._namespace
 
-    def exec_oc_cmd(self, command):
+    def exec_oc_cmd(self, command, raw_output=False):
         """
         Executing 'oc' command
 
         Args:
             command (str): The command to execute (e.g. create -f file.yaml)
                 without the initial 'oc' at the beginning
+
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/ocp.py
```diff
@@ -43,14 +43,18 @@ def kind(self):
     def namespace(self):
         return self._namespace
 
-    def exec_oc_cmd(self, command):
+    def exec_oc_cmd(self, command, raw_output=False):
```
RazTamir: 1) I'd do the opposite - out_yaml_format=True for consistency - see below functions
and if you agree with that,
```python
 if not out_yaml_format:
   ....
```

vasukulkarni: used same name

ocs_ci/ocs/ocp.py
```diff
@@ -43,14 +43,17 @@ def kind(self):
...
         return self._namespace
 
-    def exec_oc_cmd(self, command):
+    def exec_oc_cmd(self, command, out_yaml_format=True):
         """
         Executing 'oc' command
 
         Args:
             command (str): The command to execute (e.g. create -f file.yaml)
                 without the initial 'oc' at the beginning
```
fbalak:
```suggestion
```

PR #326, opened 2019-07-02
Adding test case: OCS-521 and OCS-522

PrasadDesala:


PrasadDesala:
========================================================================== short test summary info ==========================================================================
PASSED test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py::TestVerifyAllFieldsInScYamlWithOcDescribe::test_verify_all_fields_in_sc_yaml_with_oc_describe_RBD
PASSED test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py::TestVerifyAllFieldsInScYamlWithOcDescribe::test_verify_all_fields_in_sc_yaml_with_oc_describe_CEPHFS
======================================================================== 2 passed in 128.75 seconds =========================================================================

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
```diff
@@ -0,0 +1,131 @@
...
+    setup()
+
+
+def setup():
+    """
+    Setting up the environment
+    *. Creating RBD and CephFS Secret
+    *. Creating Ceph block pool
+    """
+    global RBD_SECRET_OBJ
```
fbalak: Are `RBD_SECRET_OBJ` and `CEPHFS_SECRET_OBJ` used somewhere in the test code? If not then it would be better to work with them just in fixture and don't make them global variables (e.g. use them in one fixture function with `yield`).

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
```diff
@@ -0,0 +1,131 @@
...
+    log.info("RBD and CephFS Secret's deleted successfully")
+    log.info(f"Deleting Ceph block pool {POOL}")
+    assert POOL.delete()
+    log.info(f"Ceph block pool:{POOL} deleted successfully")
+
+
+@tier1
+@pytest.mark.usefixtures(test_fixture.__name__)
+@pytest.mark.usefixtures(create_ceph_block_pool.__name__)
+@pytest.mark.polarion_id("OCS-521")
```
dahorak: How this should be reported to Polarion, when two tests have the same Polarion ID? I think, that `polarion_id` marker should be used only on test level.

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
```diff
@@ -0,0 +1,131 @@
...
+    *. Creating RBD and CephFS Secret
+    *. Creating Ceph block pool
+    """
+    global RBD_SECRET_OBJ
+    log.info("Creating RBD Secret")
+    RBD_SECRET_OBJ = helpers.create_secret(constants.CEPHBLOCKPOOL)
+
+    log.info("Creating CEPHFS Secret")
+    global CEPHFS_SECRET_OBJ
+    CEPHFS_SECRET_OBJ = helpers.create_secret(constants.CEPHFILESYSTEM)
```
fbalak: Why not use `create_rbd_secret` fixture from `contest.py` and create there also fixture `create_cephfs_secret`?

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
```diff
@@ -0,0 +1,131 @@
...
+        Test function for RBD
+        """
+        log.info("Creating a RBD Storage Class")
+        self.sc_data = templating.load_yaml_to_dict(
+            constants.CSI_RBD_STORAGECLASS_YAML
+        )
+        self.sc_data['metadata']['name'] = helpers.create_unique_resource_name(
+            'test', 'csi-rbd'
+        )
+        global RBD_SC_OBJ
```
fbalak: I suggest to delete this resource at the end of test function so there won't have to be global variable.

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
```diff
@@ -0,0 +1,131 @@
...
+
+@tier1
+@pytest.mark.usefixtures(test_fixture.__name__)
+@pytest.mark.usefixtures(create_ceph_block_pool.__name__)
+@pytest.mark.polarion_id("OCS-521")
+class TestVerifyAllFieldsInScYamlWithOcDescribe(ManageTest):
+    """
+    Test class for storageclass detail check by oc describe
+    """
+    def test_verify_all_fields_in_sc_yaml_with_oc_describe_RBD(self):
```
fbalak: I suggest to create just one test function and parametrize it instead of creating two test function with the same code.

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
```diff
@@ -0,0 +1,131 @@
...
+    log.info(f"Ceph block pool:{POOL} deleted successfully")
+
+
+@tier1
+@pytest.mark.usefixtures(test_fixture.__name__)
+@pytest.mark.usefixtures(create_ceph_block_pool.__name__)
+@pytest.mark.polarion_id("OCS-521")
+class TestVerifyAllFieldsInScYamlWithOcDescribe(ManageTest):
+    """
+    Test class for storageclass detail check by oc describe
```
fbalak: Please explain the testcase in more detail in docstring.

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
PrasadDesala: In Polarion I wrote it as a single generic test case which covers both RBD and CephFS as this scenario needs to be tested on both RBD and CephFS.
May be given that it is a tier-1 case, we can fail the test case in polarion if atleast one fails.
Thoughts?
Let me know if you have any other suggestions.

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
```diff
@@ -0,0 +1,131 @@
...
+    log.info(f"Deleting CEPHFS Secret: {CEPHFS_SECRET_OBJ}")
+    assert CEPHFS_SECRET_OBJ.delete()
+    log.info("RBD and CephFS Secret's deleted successfully")
+    log.info(f"Deleting Ceph block pool {POOL}")
+    assert POOL.delete()
+    log.info(f"Ceph block pool:{POOL} deleted successfully")
+
+
+@tier1
+@pytest.mark.usefixtures(test_fixture.__name__)
```
ebenahar: These 2 fixtures should be written under the same decorator:

```python
@pytest.mark.usefixtures(
    test_fixture.__name__,
    create_ceph_block_pool.__name__
)
```

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
PrasadDesala: I will create two separate cases in polarion for each interface.

PrasadDesala: Done, Created two separate test cases in polarion

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
PrasadDesala: Done.
================================================================================== PASSES ===================================================================================
========================================================================== short test summary info ==========================================================================
PASSED test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py::TestVerifyAllFieldsInScYamlWithOcDescribe::test_verify_all_fields_in_sc_yaml_with_oc_describe[RBD]
PASSED test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py::TestVerifyAllFieldsInScYamlWithOcDescribe::test_verify_all_fields_in_sc_yaml_with_oc_describe[CEPHFS]
======================================================================== 2 passed in 240.22 seconds =========================================================================

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
PrasadDesala: Done

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
PrasadDesala: Done

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
PrasadDesala: Done

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
PrasadDesala: Done

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
PrasadDesala: Done

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
```diff
@@ -0,0 +1,91 @@
...
+        elif interface == "CEPHFS":
+            log.info("Creating a CephFS Storage Class")
+            self.sc_data = templating.load_yaml_to_dict(
+                constants.CSI_CEPHFS_STORAGECLASS_YAML
+            )
+            self.sc_data['metadata']['name'] = (
+                helpers.create_unique_resource_name(
+                    'test', 'csi-cephfs'
+                )
+            )
```
dahorak: I think, that at least part of this code might be combined together. For
example the part with `helpers.create_unique_resource_name(...)` might be
easily replaced by general piece of code like this:

```python
        self.sc_data['metadata']['name'] = (
            helpers.create_unique_resource_name(
                'test', f'csi-{interface.lower()}'
            )
        )
```
And probably even whole `if...elif...` block of code might be replaced by
something like this:
```python
        log.info(f"Creating a {interface} Storage Class")
        self.sc_data = templating.load_yaml_to_dict(
            getattr(constants, f"CSI_{interface}_STORAGECLASS_YAML")
        )
        self.sc_data['metadata']['name'] = (
            helpers.create_unique_resource_name(
                'test', f'csi-{interface.lower()}'
            )
        )
```

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
```diff
@@ -0,0 +1,91 @@
...
+        log.info(describe_out)
+
+        # Confirm that sc yaml details matches oc describe sc output
+        value = {
+            k: describe_out[k] for k in set(describe_out) - set(self.sc_data)
+        }
+        if len(value) == 1 and value['volumeBindingMode'] == 'Immediate':
+            log.info("OC describe sc output matches storage class yaml")
+        else:
+            assert ("OC describe sc output didn't match storage class yaml")
```
dahorak: First parameter for assert should be condition, second (optional) might be error message - in your case, the assert will not do anything as the error message will be taken as *condition*  and always converted to `True`.

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
PrasadDesala: Really good comment. Thanks Fixed it.

tests/manage/test_verify_all_fields_in_sc_yaml_with_oc_describe_sc.py
PrasadDesala: Done

PR #327, opened 2019-07-02
Collect OCS logs when a test case failed.

vavuthu:
This PR will enable us  to automatically collect OCS logs when a test case failed. It helps in filing bugs.

### DIFF
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -21,6 +21,7 @@ from ocs_ci.utility.utils import (
     get_rook_version,
     get_csi_versions
 )
+from ocs_ci.ocs.utils import collect_ocs_logs
 
 __all__ = [
     "pytest_addoption",
@@ -183,3 +184,13 @@ def pytest_collection_modifyitems(session, config, items):
                 f"{item.name} in {item.fspath}",
                 exc_info=True
             )
+
+
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    outcome = yield
+    rep = outcome.get_result()
+    # we only look at actual failing test calls, not setup/teardown
+    if rep.when == "call" and rep.failed:
+        test_case_name = item.name
+        collect_ocs_logs(test_case_name)
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -73,6 +73,13 @@ def pytest_addoption(parser):
         dest='email',
         help="Email ID to send results",
     )
+    parser.addoption(
+        '--collect-logs',
+        dest='collect-logs',
+        action="store_true",
+        default=False,
+        help="Collect OCS logs when test case failed",
+    )
 
 
 def pytest_configure(config):
@@ -162,6 +169,7 @@ def process_cluster_cli_params(config):
     ocsci_config.RUN['cli_params']['deploy'] = get_cli_param(config, "deploy", default=False)
     ocsci_config.ENV_DATA['cluster_name'] = cluster_name
     ocsci_config.ENV_DATA['cluster_path'] = cluster_path
+    get_cli_param(config, 'collect-logs')
     if get_cli_param(config, 'email') and not get_cli_param(config, '--html'):
         pytest.exit("--html option must be provided to send email reports")
 
@@ -191,6 +199,10 @@ def pytest_runtest_makereport(item, call):
     outcome = yield
     rep = outcome.get_result()
     # we only look at actual failing test calls, not setup/teardown
-    if rep.when == "call" and rep.failed:
+    if (
+        rep.when == "call"
+        and rep.failed
+        and ocsci_config.RUN.get('cli_params').get('collect-logs')
+    ):
         test_case_name = item.name
         collect_ocs_logs(test_case_name)
--- a/ocs_ci/ocs/utils.py
+++ b/ocs_ci/ocs/utils.py
@@ -19,6 +19,9 @@ from ocs_ci.ocs.ceph import RolesContainer, CommandFailed, Ceph, CephNode
 from ocs_ci.ocs.clients import WinNode
 from ocs_ci.ocs.openstack import CephVMNode
 from ocs_ci.ocs.parallel import parallel
+from ocs_ci.utility.utils import create_directory_path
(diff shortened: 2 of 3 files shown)


ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -192,3 +192,13 @@ def pytest_collection_modifyitems(session, config, items):
                 f"{item.name} in {item.fspath}",
                 exc_info=True
             )
+
+
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    outcome = yield
```
mbukatov: What are you trying to do here?

vasukulkarni:
good addition!, can you make this configurable please so that we can skip for local runs.

mykaul:
Don't we prefer to use must-gather?

ocs_ci/utility/utils.py
```diff
@@ -922,3 +922,11 @@ def get_csi_versions():
             version = container['image'].split("/")[-1].split(":")[1]
             csi_versions[name] = version
     return csi_versions
+
+
+def create_directory_path(path):
+    """
+    Creates directory if path doesn't exists
+    """
+    if not os.path.exists(path):
```
clacroix12: Should we log something in the event the directory already exists? Right now it will just move on silently.

ocs_ci/ocs/utils.py
```diff
@@ -624,3 +627,25 @@ def apply_oc_resource(
...
+
+    Args:
+        dir_name (str): directory name to store OCS logs. Logs will be stored
+        in dir_name suffix with _ocs_logs.
+
+    Returns:
+        None
+
+    """
+    log_dir_path = os.path.join(ocsci_config.RUN['log_dir'], f"failed_testcase_ocs_logs_{ocsci_config.RUN['run_id']}")
```
clacroix12:
```suggestion
    log_dir_path = os.path.join(
        ocsci_config.RUN['log_dir'],
        f"failed_testcase_ocs_logs_{ocsci_config.RUN['run_id']}"
    )
```

ocs_ci/ocs/utils.py
```diff
@@ -624,3 +627,25 @@ def apply_oc_resource(
...
+
+def collect_ocs_logs(dir_name):
+    """
+    Collects OCS logs
+
+    Args:
+        dir_name (str): directory name to store OCS logs. Logs will be stored
+        in dir_name suffix with _ocs_logs.
+
+    Returns:
```
clacroix12: You can omit the `Returns` section if it doesn't return anything.

vavuthu: Done

ocs_ci/ocs/utils.py
vavuthu: Done

ocs_ci/utility/utils.py
vavuthu: Done

vavuthu:
> Don't we prefer to use must-gather?

used must-gather in commit  a305f52b84c6ab8a6fe29a3733b715899387d4c4

ocs_ci/ocs/utils.py
```diff
@@ -624,3 +627,25 @@ def apply_oc_resource(
...
+    log_dir_path = os.path.join(
+        ocsci_config.RUN['log_dir'],
+        f"failed_testcase_ocs_logs_{ocsci_config.RUN['run_id']}"
+    )
+    create_directory_path(log_dir_path)
+    dir_name = f"{dir_name}_ocs_logs"
+    dump_dir = os.path.join(log_dir_path, dir_name)
+    cmd = f"adm must-gather --image=ashishranjan738/ocs-must-gather --dest-dir={dump_dir}"
+    log.info(f"OCS logs are placed in location {dump_dir}")
+    occli = OCP(namespace=defaults.ROOK_CLUSTER_NAMESPACE)
```
petr-balogh: please use namespace from config ENV_DATA section

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -183,3 +192,17 @@ def pytest_collection_modifyitems(session, config, items):
                 f"{item.name} in {item.fspath}",
                 exc_info=True
             )
+
+
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_runtest_makereport(item, call):
```
petr-balogh: @vavuthu  how long does it take to get all data from must gather?

Vijay or @nehaberry do you know if we will get some advantage to run it after each failed test case? 

Isn't enough to run only at the end of the execution to get all the logs after the the execution and run it only once?  I guess that if if we will have 2 failed test cases that we will spend time to collecting logs twice with same results as we do only once at the end of execution or do I miss something?

**I see advantage if**: 

- must gather collect logs just from last execution of must gather of if we can specify somehow timestamp
- if we will miss some data from some deleted pods from the test which are not able to collect after some time . (This I am not sure how it works)

@nehaberry or @vavuthu do you know if must gather collect also logs for deleted pods?

Vijay, as far as I understand this hook is running after the test finish and it runs after finalizers of fixtures right?  Just thinking if must gather doesn't collect the logs from deleted pods that it probably will miss some logs from resources we will delete in finalizer. But not sure how it works.

Thanks for clarifications

ocs_ci/ocs/utils.py
vavuthu: > please use namespace from config ENV_DATA section

ahh,, Its strange I have removed namespace parameters in current commit , its reflecting the same in my branch but not in PR

ocs_ci/framework/pytest_customization/ocscilib.py
vavuthu: @petr-balogh ran must gather couple of times and  I can see it took ~31 sec
$ time oc adm must-gather --image=ashishranjan738/ocs-must-gather:latest

real	0m31.134s
user	0m0.723s
sys	0m0.259s

It will collect logs before tear down executes ( after test case fails but before finalizers executes ) so that we will have logs exactly in the state where test case failed which will be easy for debugging.

I am not sure whether it will collect deleted pods or not ( need to check )

petr-balogh: Great, if it will run after finalizers (tear down) that it's good addition and ~31 sec is not to much.  We will see if it will takes more after some longer running execution.

But that case I agree that it's better that all the logs will be more readable if they will end with the failure of the test case and won't have additional logs with teardown.

Thanks

PR #352, opened 2019-07-03
tox: Fixes for running flake8

zmc:


zmc:
Note: the issue I'm fixing with the second commit never affected TravisCI, since it places its virtualenv outside of the project directory.

tox.ini
```diff
@@ -17,9 +17,10 @@ commands = py.test \
 deps =
     flake8
     flake8-mutable
-commands = flake8
+commands = flake8 ocs_ci
```
clacroix12: We will want to include `tests` here as well since they are outside of `ocs_ci`

tox.ini
```diff
@@ -17,9 +17,10 @@ commands = py.test \
 deps =
     flake8
     flake8-mutable
-commands = flake8
+commands = flake8 ocs_ci tests
 
 [flake8]
+basepython = python3
```
MeridianExplorer: So I'm really not sure about this; Just bringing it up for discussion;
I think it might be better to specifically state python3.6/python3.7, considering our code depends on Python>=3.6.
According to [the docs](https://tox.readthedocs.io/en/latest/config.html#conf-basepython), python3 simply calls.... python3, which I assume might be below version 3.6, and might cause possible issues with flake8.
If python3 is safe and shouldn't hinder the process in any way, feel free to disregard my comment :)

zmc: Well, we need this to work on 3.6 *or* 3.7, so I'm not sure if we can express that properly here

vasukulkarni: so what are the next steps here? drop the PR?

zmc: I would vote for merging it since it addresses real issues. I just have no reviews...

vasukulkarni: what about @MeridianExplorer comment? it will only pickup one version of python for CI tests?

zmc: If you check the Travis build attached to this PR, you'll see it runs as normal

vasukulkarni: thanks, yep dont see any issues.

vasukulkarni:
@clacroix12 @mbukatov @petr-balogh FYI, lets close on this one.

PR #353, opened 2019-07-03
Drop debug-level logging for openshift-install

zmc:
I expect that some people will rather make this configurable than to simply drop `--log-level debug` entirely. I'll be willing to do that work *if* we want it. I also wouldn't be surprised if nobody would actually miss that output, though, so I'm starting with this.

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -413,8 +413,7 @@ def destroy_cluster(cluster_path):
 
     destroy_cmd = (
         f"{installer} destroy cluster "
-        f"--dir {cluster_path} "
-        f"--log-level debug"
+        f"--dir {cluster_path}"
     )
 
     try:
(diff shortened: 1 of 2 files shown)


vasukulkarni:
@zmc this should be configurable, for jenkins as well as folks in deployment team for them to debug, can you add the option please. thanks

clacroix12:
@zmc @vasukulkarni should we just hook into the log level set for pytest to keep things consistent?

vasukulkarni:
> @zmc @vasukulkarni should we just hook into the log level set for pytest to keep things consistent?

That makes sense to me, but there is a current problem with 'debug' that is causing huge logs in couple of tests due to which installer's debug will have to suffer :(

clacroix12:
@zmc @vasukulkarni I opened #395 which makes it so the openshift-installer will use pytest's `log_cli_level`.

vasukulkarni:
I think we can drop this due to #395

clacroix12:
Closing due to #395

Issue #285, opened 2019-06-24
Use the provided setup and teardown class methods pytest supports


andrewschoen:
In the following test there is a ``setup`` and ``teardown`` method that should be a part of the class and not used as class fixtures. https://github.com/red-hat-storage/ocs-ci/blob/master/tests/manage/ocs_288.py

The ``setup`` and ``teardown`` should be a part of ``class TestCaseOCS288``. In this way we can avoid the use of the GLOBAL keyword in those methods. Things like ``RBD_SECRET_OBJ`` and ``CEPHFS_SECRET_OBJ`` should not need to be global data, they can just exist as properties of the class.

For reference: https://docs.pytest.org/en/latest/xunit_setup.html

RazTamir:
In this link, the usage of setup and teardown are from the unittest world and replaced with fixutres in pytest:
"Note
While these setup/teardown methods are simple and familiar to those coming from a unittest or nose background, you may also consider using pytest’s more powerful fixture mechanism which leverages the concept of dependency injection, allowing for a more modular and more scalable approach for managing test state, especially for larger projects and for functional testing. You can mix both fixture mechanisms in the same file but test methods of unittest.TestCase subclasses cannot receive fixture arguments."

andrewschoen:
@RazTamir yeah, fixtures can work here too. My issue is really with the use of ``GLOBAL`` keyword here. Those really feel like state that should exist on the class and not global state.

alfredodeza:
@RazTamir the reference to pytest's documentation explains the usage of teardown (via the finalizer) in a fixture, however the test used as an example is using a fixture to teardown the test - this is an anti-pattern because the fixture isn't providing anything to the test itself except to teardown. I think this is why it has to depend on using `global`.

Using asserts to verify correctness like in the teardown section of the fixture is also not a common workflow in Python (unlike in C and C++). Raising exceptions is usually preferred.

RazTamir:
The only way I can see this work is to use setup fixture that will initialize all those attributes and the class will use it.

alfredodeza:
I don't think there is a need at all to have a fixture to setup/teardown the test methods. That is what setup and teardown are for. If more than one class is going to use the same setup/teardown helpers then it would be OK to use a base class (even though I dislike inheritance).

Perhaps if you could expand more as to why this approach is being preferred?

clacroix12:
[...]

PR #374, opened 2019-07-08
WIP: switch managage/test_add_mds_to_cluster.py to setup/teardown class methods

andrewschoen:
An example of what #285 is trying to address.

This allows us to avoid global variable usage while still leveraging pytest's setup and teardown capabilities.

### DIFF
--- a/tests/manage/test_add_mds_to_cluster.py
+++ b/tests/manage/test_add_mds_to_cluster.py
@@ -3,8 +3,6 @@ A test for creating a CephFS
 """
 import logging
 
-import pytest
-
 from ocs_ci.ocs import constants, defaults, ocp
 from ocs_ci.framework import config
 from ocs_ci.framework.testlib import tier1, ManageTest
@@ -14,51 +12,12 @@ log = logging.getLogger(__name__)
 
 CEPHFS_DELETED = '"{cephfs_name}" deleted'
 
-POD = ocp.OCP(kind=constants.POD, namespace=config.ENV_DATA['cluster_namespace'])
-CEPHFS = ocp.OCP(kind=constants.CEPHFILESYSTEM)
-CEPH_OBJ = None
-
-
-@pytest.fixture(scope='class')
-def test_fixture(request):
-    """
-    Create disks
-    """
-    self = request.node.cls
-
-    def finalizer():
-        teardown(self)
-    request.addfinalizer(finalizer)
-    setup(self)
-
-
-def setup(self):
-    """
-    Setting up the environment for the test
-    """
-    global CEPHFS, CEPH_OBJ
-    CEPHFS = ocp.OCP(kind=constants.CEPHFILESYSTEM)
-    self.fs_data = CEPHFS.get(defaults.CEPHFILESYSTEM_NAME)
-    self.fs_name = self.fs_data['metadata']['name']
-    CEPH_OBJ = OCS(**self.fs_data)
-
-
-def teardown(self):
-    """
-    Tearing down the environment
-    """
-    self.fs_data = CEPHFS.get(defaults.CEPHFILESYSTEM_NAME)
-    self.fs_data['spec']['metadataServer']['activeCount'] = (
-        self.original_active_count
-    )
-    CEPH_OBJ.apply(**self.fs_data)
-    assert verify_fs_exist(self.original_active_count * 2)
-
 
 def verify_fs_exist(pod_count):
     """
     Verifying if a ceph FS exist
     """
+    POD = ocp.OCP(kind=constants.POD, namespace=config.ENV_DATA['cluster_namespace'])
     assert POD.wait_for_resource(
         condition='Running', selector='app=rook-ceph-mds',
         resource_count=pod_count
@@ -70,9 +29,6 @@ def verify_fs_exist(pod_count):
 
 
 @tier1
-@pytest.mark.usefixtures(
-    test_fixture.__name__,
-)
 class TestCephFilesystemCreation(ManageTest):
     """
     Testing creation of Ceph FileSystem
@@ -80,6 +36,20 @@ class TestCephFilesystemCreation(ManageTest):
     new_active_count = 2
     original_active_count = 1
     fs_name = None
+
+    def setup_class(self):
+        self.CEPHFS = ocp.OCP(kind=constants.CEPHFILESYSTEM)
+        self.fs_data = self.CEPHFS.get(defaults.CEPHFILESYSTEM_NAME)
+        self.fs_name = self.fs_data['metadata']['name']
+        self.CEPH_OBJ = OCS(**self.fs_data)
+
+    def teardown_class(self):
+        self.fs_data = self.CEPHFS.get(defaults.CEPHFILESYSTEM_NAME)
+        self.fs_data['spec']['metadataServer']['activeCount'] = (
+            self.original_active_count
+        )
+        self.CEPH_OBJ.apply(**self.fs_data)
+        assert verify_fs_exist(self.original_active_count * 2)
 
     def test_cephfilesystem_creation(self):
         """
@@ -89,5 +59,5 @@ class TestCephFilesystemCreation(ManageTest):
         self.fs_data['spec']['metadataServer']['activeCount'] = (
             self.new_active_count
         )
-        CEPH_OBJ.apply(**self.fs_data)
+        self.CEPH_OBJ.apply(**self.fs_data)
         assert verify_fs_exist(self.new_active_count * 2)
--- a/tests/manage/test_add_mds_to_cluster.py
+++ b/tests/manage/test_add_mds_to_cluster.py
@@ -2,6 +2,7 @@
 A test for creating a CephFS
 """
 import logging
+import pytest
 
 from ocs_ci.ocs import constants, defaults, ocp
 from ocs_ci.framework import config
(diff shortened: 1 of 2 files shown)


RazTamir:
Hi Andrew,
This setup and teardown are unittest methods and we are not going with that direction please. 

We are about to share with the team the right way to use fixtures.

andrewschoen:
> Hi Andrew,
> This setup and teardown are unittest methods and we are not going with that direction please.
> 
> We are about to share with the team the right way to use fixtures.

They are the old unittest style, but they are supported by pytest. My issue is with usage of global variables, if we can avoid that in a fixture then that's great. I'm curious, can you explain why you prefer the original approach over this method though?

Feel free to close the PR if this isn't a direction you'd like to go, but I felt there was maybe some confusion on what I meant on #285 so I wanted to make this to provide some clarity.

ebenahar:
Hi @andrewschoen, please have a look in my reply to the related issue - https://github.com/red-hat-storage/ocs-ci/issues/285#issuecomment-508416215

As Raz mentioned, along with explaining the right way of using fixtures, we'll ask everyone to avoid using global vars.

vasukulkarni:
Setting aside xUnit style setup/teardown, I think self. makes total sense to me to get rid of globals.

RazTamir:
Let me know what do you think

```python

@fixture
def initialize_params(self):
    self.x = OCP()
    self.y = PVC()

@usefixture(
    initialize_params.__name__,
    do_1.__name__,
    do_2.__name__,
)
class test_my_test():
    """
    Some test class
    """
    def test_123(self):
        self.x.create()
        pvc_size = self.y.size



```

andrewschoen:
If we wanted to use fixtures, I'd probably take this approach.

```
@fixture(scope="class")
def ocp(self):
    return OCP()

@fixture(scope="class")
def pvc():
    return PVC()

@usefixture(
    "do_1",
    "do_2",
)
class test_my_test():
    """
    Some test class
    """
    def test_123(self, ocp, pvc):
        ocp.create()
        pvc_size = pvc.size
```

Because we have the scope set to class pytest will return the same ``OCP`` and ``PVC`` instance for all test methods within the same class.

andrewschoen:
I suppose if the ``OCP.create()`` only needs to happen once that could happen in the fixture itself.

```
@fixture(scope="class")
def ocp(self):
    ocp = OCP()
    ocp.create()
    return ocp
```

vasukulkarni:
> Let me know what do you think
> 
> ```python
> @fixture
> def initialize_params(self):
>     self.x = OCP()
>     self.y = PVC()
> 
> @usefixture(
>     initialize_params.__name__,
>     do_1.__name__,
>     do_2.__name__,
> )
> class test_my_test():
>     """
>     Some test class
>     """
>     def test_123(self):
>         self.x.create()
>         pvc_size = self.y.size
> ```

I think the additional function to initialize params and pass it as fixture can be avoided, else, we will have to decorate other fixtures to use params fixtures, If we can keep it simple with just self. in the fixture function or setup_class method that will be straighforward for all users.

clacroix12:
> I think the additional function to initialize params and pass it as fixture can be avoided

100%. If we have functions that need access to the same variables we can simply move them inside the test class and reference them with `self.whatever`. There's no need to make this more complicated than it needs to be. Classes provide us with an extremely easy way to share data across methods, let's use it. 

As far as the design of the fixtures goes, I agree with @andrewschoen 's approach. The example fixtures have singular purposes and provide a resource for the tests to utilize easily. This actually makes them good candidates to move to `conftest.py` so other test cases can utilize them instead of being heavily coupled with the tests in this file.

RazTamir:
I agree on to initialize the OCP() in the relevant fixtures but, what will happen in this scenario:
```python
def fixture1(request):
    self.pod = OCP(kind=pod)
    do_more_things_pod()

def fixture2(request):
    self.pod = OCP(kind=pod)
    do_other_things_with_pod()

@usefixtures(fixture1, fxture2)
class test_something():
    pass
```

This will override my first instance and might cause unwanted behavior

clacroix12:
@RazTamir it depends on what your intention is with `self.pod` in this example. Is `self.pod` meant to be shared across fixtures? If so you wouldn't want to instantiate it twice. Is it supposed to be a local `pod` variable that isn't shared? Then you wouldn't define it with `self.` and instead just use a local variable that won't be overwritten by other fixtures.

andrewschoen:
> I agree on to initialize the OCP() in the relevant fixtures but, what will happen in this scenario:
> 
> ```python
> def fixture1(request):
>     self.pod = OCP(kind=pod)
>     do_more_things_pod()
> 
> def fixture2(request):
>     self.pod = OCP(kind=pod)
>     do_other_things_with_pod()
> 
> @usefixtures(fixture1, fxture2)
> class test_something():
>     pass
> ```
> 
> This will override my first instance and might cause unwanted behavior

This is one of the reasons we shouldn't use self and instead just return an OCP object. If we do that then pytest can keep these isolated between test methods or classes. Then we can have test cases that perform the actions and assert the results.

zmc:
@RazTamir Won't using `self` inside a fixture just raise `NameError`? Also, the class definition there doesn't look right.

To me, there are times where using fixtures make sense, and times where `setup()`/`setup_class()` and friends make sense. This case looks like one where class-scoped fixtures might be better - but it's mainly a question of style, not technical correctness.

Of course, `global`s should be purged either way.

RazTamir:
The main (not the only one) issue with "old" setup and teardown is that in case the setup fails, the teardown is not even executed

andrewschoen:
I've updated this to use fixtures. I'm having trouble deploying a cluster so I haven't tested this quite yet, but wanted to post the changes regardless.

tests/manage/test_add_mds_to_cluster.py
```diff
@@ -69,25 +29,45 @@ def verify_fs_exist(pod_count):
     return False
 
 
+@pytest.fixture(scope="class")
+def cephfs():
+    cephfs = ocp.OCP(kind=constants.CEPHFILESYSTEM)
+    return cephfs
+
+
+@pytest.fixture(scope="class")
```
andrewschoen: Because we only have one test method in this class I'm not entirely sure what the correct scope should be. Setting it to ``class`` would mean that each test method within the class using the fixture would get the same copy of ``fs_data``. If we set it to ``function`` then each test method would get a fresh copy. The original teardown was resetting a value on ``fs_data`` to the original value. If we set the scope to ``function`` that wouldn't be necessary.

It seems to me that if we were gonna move these functions to conftest.py we'd want the scope set to ``function`` so each test method could be isolated?

clacroix12: @andrewschoen I agree with function scope, especially if they move to `conftest.py`. Unless it's necessary to share state between tests, function scope makes sense. Creating new objects is cheap and having state shared automatically can cause unintended issues if it isn't accounted for.

tests/manage/test_add_mds_to_cluster.py
```diff
@@ -2,92 +2,58 @@
...
-    """
-    self.fs_data = CEPHFS.get(defaults.CEPHFILESYSTEM_NAME)
-    self.fs_data['spec']['metadataServer']['activeCount'] = (
-        self.original_active_count
-    )
-    CEPH_OBJ.apply(**self.fs_data)
-    assert verify_fs_exist(self.original_active_count * 2)
+@pytest.fixture(scope="class")
+def ceph_obj(request, fs_data):
+    ceph_obj = OCS(**fs_data)
```
RazTamir: And if I want 1 more ceph object? a different one

tests/manage/test_add_mds_to_cluster.py
```diff
@@ -2,92 +2,58 @@
...
 class TestCephFilesystemCreation(ManageTest):
     """
     Testing creation of Ceph FileSystem
     """
     new_active_count = 2
-    original_active_count = 1
-    fs_name = None
 
-    def test_cephfilesystem_creation(self):
+    def test_cephfilesystem_creation(self, ceph_obj, fs_data):
```
RazTamir: Maybe I'm missing something but this implementation limiting us

tests/manage/test_add_mds_to_cluster.py
clacroix12: If we wanted a fixture to be able to provide any number of resources we could do what @andrewschoen suggested in yesterdays meeting and have the fixture return a function that you can call to create the object instead of returning the object directly.

tests/manage/test_add_mds_to_cluster.py
clacroix12: What limitations is this putting on us?

tests/manage/test_add_mds_to_cluster.py
andrewschoen: Yeah, fixtures can be factory methods as well. Here's some good documentation on that. https://docs.pytest.org/en/latest/fixture.html#factories-as-fixtures

tests/manage/test_add_mds_to_cluster.py
vasukulkarni: resolving this after our current discussion today

tests/manage/test_add_mds_to_cluster.py
vasukulkarni: for some objects class makes sense and for some function, I think this is all test dependent, at the conftest.py we should default to class level and individual tests can create their own function level scopes

vasukulkarni:
@andrewschoen I am using this as template to work on one more test script, mind testing this one, so we can merge this?

tests/manage/test_add_mds_to_cluster.py
clacroix12: @vasukulkarni I think it would be better for us to align with what pytest itself defaults to which is function scope. If a fixture is explicitly designed to happen once per test class then we can set class scope, otherwise we can assume that a fixture is creating a resource for one test only, and tests are receiving resources that were not possibly altered by other tests.

vasukulkarni: @clacroix12 this is going to be costly if we keep doing create/delete resources which other tests can utilize, The more I think we should better let the tests decide this and not framework.

clacroix12: I agree that tests should choose, however scope is determined via the fixture decorator so I think we need to provide both class and function scope fixtures to our tests so that they can choose which is appropriate for their use-cases. 

I just didn't want us configuring every one of our fixtures to class scope when that can cause unintended issues where tests conflict with each other. As our collection of tests grows we should really try to keep test cases in as isolated of an environment as possible. I think it's better to create/delete resources for tests a bit more aggressively than we might need to in order to avoid issues like a newly implemented test altering the class scoped resource which causes failures in other test cases.

RazTamir:
@andrewschoen can we close this PR?

vasukulkarni:
Closing the PR due to inactivity and also it served as an example and few things have changed now. Please reopen if needed. Thanks

Issue #86, opened 2019-05-23
Integrate Bugzilla plugin


RazTamir:
We need to integrate Bugzilla to our framework so test cases will be skipped in case there's an open bug.

The implementation in RHV already exists so we can use it

mbukatov:
Fixed by https://github.com/red-hat-storage/ocs-ci/pull/375

PR #375, opened 2019-07-08
Add bugzilla integration

petr-balogh:
Fixing the issue #86

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -52,6 +52,18 @@ to the pytest.
    necessary styles and scripts
 * `--email` - to send the email reports of the test run which was generated
    by --html option. MUST specify --html to send email reports.
+* `--bugzilla` - allows you to skip tests with unresolved bug (test case
+  needs to be properly decorated as you can see in
+  [documentation](./writing_tests.md)). You will also need to
+  create configuration file for `pytest_marker_bugzilla` plugin. See the
+  [plugin documentation](https://github.com/eanxgeek/pytest_marker_bugzilla)
+  or example of `/etc/bugzilla.cfg` file:
+  ```bash
+  [DEFAULT]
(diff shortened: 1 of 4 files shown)


petr-balogh:
Verified in [jenkins job](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/110/testReport/tests.manage.test_pv_creation/TestPvCreation/test_pv_creation/) where I added testing commit to decorate one test case, after verification I deleted this commit.

docs/usage.md
```diff
@@ -52,6 +52,18 @@ to the pytest.
    necessary styles and scripts
 * `--email` - to send the email reports of the test run which was generated
    by --html option. MUST specify --html to send email reports.
+* `--bugzilla` - allows you to skip tests with unresolved bug (test case
+  needs to be properly decorated as you can see in
```
vasukulkarni: If we have the bugzilla mark but dont have the /etc/bz conf file, will it create a problem or skip the test?

petr-balogh: Just tested this scenario that I moved the configuration file to different location and saw that the test was executed as regularly without any skip and without any error that bugzilla config file is missing.  Anyway you can decide if you pass the --bugzilla parameter to the execution. So if you don't have config file, it doesn't sense to pass --bugzilla cause it has no effect.

You can see also from the code of plugin [here](https://github.com/eanxgeek/pytest_marker_bugzilla/blob/master/pytest_marker_bugzilla.py#L315)
```python
if config.getvalue("bugzilla") and config.getvalue('bugzilla_url'):
```
Which will continue to initialize plugin only if bugzilla parameter provided and bugzilla_url is available.

Issue #335, opened 2019-07-02
--teardown on partially destroyed cluster fails


vasukulkarni:
we need to ignore those errors during teardown

```
NTERNALERROR>     raise ex[1].with_traceback(ex[2])
INTERNALERROR>   File "/home/ebenahar/git/ocs-ci/ocs-venv/lib64/python3.7/site-packages/pluggy/callers.py", line 187, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>   File "/home/ebenahar/git/ocs-ci/ocs_ci/framework/pytest_customization/ocscilib.py", line 103, in pytest_configure
INTERNALERROR>     ceph_version = get_ceph_version()
INTERNALERROR>   File "/home/ebenahar/git/ocs-ci/ocs_ci/utility/utils.py", line 876, in get_ceph_version
INTERNALERROR>     ct_pod = pod.get_ceph_tools_pod()
INTERNALERROR>   File "/home/ebenahar/git/ocs-ci/ocs_ci/ocs/resources/pod.py", line 173, in get_ceph_tools_pod
INTERNALERROR>     )['items'][0]
INTERNALERROR> IndexError: list index out of range
I don't have the ceph pods:
[ebenahar@localhost ocs-ci]$ oc -n openshift-storage --kubeconfig /home/ebenahar/ebenahar-ocs/auth/kubeconfig get pod
NAME                                  READY   STATUS    RESTARTS   AGE
rook-ceph-operator-5c6fd4b7db-2twzd   1/1     Running   0          51m
rook-discover-c8sv5                   1/1     Running   0          42m
rook-discover-cs64z                   1/1     Running   0          42m
rook-discover-g6rsc                   1/1     Running   0          42m
```

fbalak:
I believe that this is not fixed yet. I currently have incorrectly deployed cluster:
```
$ oc get pods -n openshift-storage
NAME                                  READY   STATUS                  RESTARTS   AGE
csi-cephfsplugin-2zq6c                2/2     Running                 0          49m
csi-cephfsplugin-provisioner-0        3/3     Running                 0          49m
csi-cephfsplugin-psx6z                2/2     Running                 0          49m
csi-cephfsplugin-px8rr                2/2     Running                 0          49m
csi-rbdplugin-dkrkn                   2/2     Running                 0          49m
csi-rbdplugin-ln59c                   2/2     Running                 0          49m
[...]

PR #377, opened 2019-07-09
Destroy cluster in bad shape

petr-balogh:
We cannot collect data like version info from the cluster when it's not
in good shape. So when destroy cluster, we just skip the version
collecting with change in this PR.

Should also fix the issue #335

### DIFF
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -86,34 +86,42 @@ def pytest_configure(config):
         process_cluster_cli_params(config)
         # Add OCS related versions to the html report and remove extraneous metadata
         markers_arg = config.getoption('-m')
-        if not ("deployment" in markers_arg and ocsci_config.RUN['cli_params']['deploy']):
-            print("Collecting Cluster versions")
-            # remove extraneous metadata
-            del config._metadata['Python']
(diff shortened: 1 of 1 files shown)


nehaberry:
@petr-balogh is there a change in executing the --teardown command ? do we need to pass any option to not check for "cluster version ?

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -86,34 +86,42 @@ def pytest_configure(config):
...
+        print("Collecting Cluster versions")
+        # remove extraneous metadata
+        del config._metadata['Python']
+        del config._metadata['Packages']
+        del config._metadata['Plugins']
+        del config._metadata['Platform']
+
+        try:
+            # add cluster version
+            clusterversion = get_cluster_version()
```
vasukulkarni: I think we need timeout in addition to try/except

petr-balogh: @vasukulkarni  I didn't touch original code, it's just moved one indent left cause I've changed condition logic on the top and returning to skip the rest of the code.

vasukulkarni: got it, thanks, make sense to skip the complete check

Issue #323, opened 2019-07-01
limit the size of console logs we are printing out in some tier1 tests


vasukulkarni:
some tier tests are printing way too much data to console, makes it difficult to debug what acutally failed, fix those.  As a side effect our html report is bloated too.. 

- [ ] tier1: identify/ fix tests
- [x]  html report size: check html report size

vasukulkarni:
So the issue mainly seems to be from the single debug line(all namespace output i guess)
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L376

vasukulkarni:
Can be solved from this PR too https://github.com/red-hat-storage/ocs-ci/pull/306

vasukulkarni:
#306 is merged, testing with latest master!

vasukulkarni:
@clacroix12 . @andrewschoen  can you check this one, I still see debug logs printed with default settings.

vasukulkarni:
This is fixed!

PR #383, opened 2019-07-10
Suppress captured logs at end of session

clacroix12:
Automatically add --no-print-logs option to suppress dumping logs at the end of a session and to reduce html report file size.

This change solves two issues:
- Captured logs being dumped at the end of a session makes viewing the overall results in the console practically impossible.
- Enormous file size of html reports caused giant load times and reduces the value of these _summary_ reports

Before:
- tier1-report.html - 431M
- 55 sec. load time in browser

After:
- tier1-report.html - 906K
- < 1 sec. load time in browser

I don't think this 100% fixes the issue described in #323 but it is a significant start.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -27,5 +27,5 @@ log_format = %(asctime)s - %(threadName)s - %(name)s - %(levelname)s - %(message
 log_cli=true
 log_cli_level = INFO
 log_level = DEBUG
-addopts = --ignore=ocs_ci
+addopts = --ignore=ocs_ci --no-print-logs
 testpaths = tests


vasukulkarni:
So If I go to individual logs in /tmp/logs and load the log file, I assume that size is smaller and can be loaded in browser? can you confirm? Thanks

clacroix12:
@vasukulkarni this isn't changing the size of the log files no, that is handled by our `log_level` in `pytest.ini` as well as _how_ our tests are logging. This change is suppressing the log dump at the end of a session and reducing the html report size. Actual log files are unaffected.

vasukulkarni:
OK i forgot to try --log-level info!

Issue #386, opened 2019-07-10
Pod is not being deleted as part of test_run_io teardown which causes the test teardown to stuck during PVC deletion


ebenahar:
tests.manage.test_pvc_deletion_during_io.TestDeletePVCWhileRunningIO#test_run_io

tests/manage/test_pvc_deletion_during_io.py::TestDeletePVCWhileRunningIO::test_run_io 
------------------------------------------------------------------------------------------------- live log setup --------------------------------------------------------------------------------------------------
19:27:36 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ebenahar/ebenahar-ocs-dir/auth/kubeconfig get Pod --selector=app=rook-ceph-tools -o yaml
19:27:37 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ebenahar/ebenahar-ocs-dir/auth/kubeconfig rsh rook-ceph-tools-76c7d559b6-6zmpg ceph auth get-key client.admin --format json-pretty
19:27:40 - MainThread - ocs_ci.ocs.resources.ocs - INFO - Adding Secret with name secret-test-1019274019
19:27:40 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ebenahar/ebenahar-ocs-dir/auth/kubeconfig create -f /tmp/Secretrjhpxjd6 -o yaml
19:27:41 - MainThread - ocs_ci.ocs.resources.ocs - INFO - Adding CephBlockPool with name cbp-test-1019274107
19:27:41 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ebenahar/ebenahar-ocs-dir/auth/kubeconfig create -f /tmp/CephBlockPoolg220d3hh -o yaml
19:27:41 - MainThread - tests.helpers - INFO - Verifying that block pool cbp-test-1019274107 exists
19:27:41 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ebenahar/ebenahar-ocs-dir/auth/kubeconfig get Pod --selector=app=rook-ceph-tools -o yaml
[...]

PR #339, opened 2019-07-02
Cleanup pod and pvc resources after test, also introduced a helper to…

shylesh:
… find pv for pvc
Addresses issue https://github.com/red-hat-storage/ocs-ci/issues/338

Signed-off-by: Shylesh Kumar <shmohan@redhat.com>

PR #387, opened 2019-07-10
Fix for https://github.com/red-hat-storage/ocs-ci/issues/386

RazTamir:


tests/manage/test_pvc_deletion_during_io.py
```diff
@@ -52,6 +52,13 @@ def test_run_io_and_delete_pvc(self):
             if "NotFound" in str(ex):
                 pass
 
+
+@pytest.mark.usefixtures(delete_pod.__name__)
+class TestRunningIO(TestDeletePVCWhileRunningIO):
```
ebenahar: I don't think we should inherit from the test class.
Let's create a base class which the basic fixtures and inherit from it for both TCs

RazTamir: I agree.
Will fix it

tests/fixtures.py
```diff
@@ -144,3 +144,16 @@ def create_rbd_pod(request):
         interface_type=constants.CEPHBLOCKPOOL,
         pvc_name=class_instance.pvc_obj.name
     )
+
+
+@pytest.fixture()
```
petr-balogh: I think this is duplicate work of those two PRs:
https://github.com/red-hat-storage/ocs-ci/pull/396/files#diff-8a66b2566ee3ea0285647f603db9174bR159
https://github.com/red-hat-storage/ocs-ci/pull/339/files#diff-10993bfc055d543860f899ca9733c55aR20

ebenahar:
Not needed anymore, the test will be removed with https://github.com/red-hat-storage/ocs-ci/pull/435

PR #390, opened 2019-07-10
Wait for fio results to complete

vasukulkarni:
Wait for fio results to complete as this is the most common use case.

### DIFF
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -200,8 +200,8 @@ class Pod(OCS):
             jobs (int): Number of jobs to execute FIO
             runtime (int): Number of seconds IO should run for
             fio_filename(str): Name of fio file created on app pod's mount point
-            wait (bool): weather to wait for results or run in async mode
-                         default is wait for fio run to complete
+            wait (bool): Whether to wait for results or run in async mode
+                (default: True - waits for fio run to complete)
         """
         name = 'test_workload'
         spec = self.pod_data.get('spec')
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -177,7 +177,7 @@ class Pod(OCS):
 
     def run_io(
         self, storage_type, size, io_direction='rw', rw_ratio=75,
-        jobs=1, runtime=60, fio_filename=None
+        jobs=1, runtime=60, fio_filename=None, wait=True
     ):
         """
         Execute FIO on a pod
@@ -200,6 +200,8 @@ class Pod(OCS):
             jobs (int): Number of jobs to execute FIO
             runtime (int): Number of seconds IO should run for
             fio_filename(str): Name of fio file created on app pod's mount point
+            wait (bool): weather to wait for results or run in async mode
+                         default is wait for fio run to complete
         """
         name = 'test_workload'
         spec = self.pod_data.get('spec')
@@ -230,6 +232,9 @@ class Pod(OCS):
             io_params['filename'] = fio_filename
 
         self.fio_thread = wl.run(**io_params)
+        if wait:
+            logger.info("Waiting for fio job to complete")
+            self.get_fio_results()
 
 
 # Helper functions for Pods
--- a/tests/manage/test_pvc_disruptive.py
+++ b/tests/manage/test_pvc_disruptive.py
@@ -3,6 +3,7 @@ import logging
 from ocs_ci.ocs import constants
 from tests import helpers, disruption_helpers
 from ocs_ci.framework.testlib import ManageTest, tier4
+from time import sleep
 from tests.fixtures import (
     create_rbd_storageclass, create_ceph_block_pool, create_cephfs_storageclass,
     create_rbd_secret, create_cephfs_secret
@@ -46,7 +47,11 @@ class BaseDisruption(ManageTest):
         assert self.pod_obj.ocp.wait_for_resource(
             condition=constants.STATUS_RUNNING, resource_name=self.pod_obj.name, timeout=120
         )
-        self.pod_obj.run_io(storage_type=self.storage_type, size='1G')
+        self.pod_obj.run_io(
+            storage_type=self.storage_type, size='1G', wait=False
+        )
+        # wait for atleast 10 seconds for IO to start
+        sleep(10)
         if operation_to_disrupt == 'run_io':
             DISRUPTION_OPS.delete_resource()
 
--- a/tests/manage/test_pvc_deletion_during_io.py
+++ b/tests/manage/test_pvc_deletion_during_io.py
@@ -57,7 +57,7 @@ class TestDeletePVCWhileRunningIO(ManageTest):
         """
         Test IO
         """
-        self.pod_obj.run_io('fs', '1G')
+        self.pod_obj.run_io('fs', '1G', wait=False)
         logging.info("Waiting for results")
         fio_result = self.pod_obj.get_fio_results()
         logging.info("IOPs after FIO:")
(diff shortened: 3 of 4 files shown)


ocs_ci/ocs/resources/pod.py
```diff
@@ -200,6 +200,8 @@ def run_io(
             jobs (int): Number of jobs to execute FIO
             runtime (int): Number of seconds IO should run for
             fio_filename(str): Name of fio file created on app pod's mount point
+            wait (bool): weather to wait for results or run in async mode
+                         default is wait for fio run to complete
```
RazTamir: fix indentation

ocs_ci/ocs/resources/pod.py
```diff
@@ -200,6 +200,8 @@ def run_io(
             jobs (int): Number of jobs to execute FIO
             runtime (int): Number of seconds IO should run for
             fio_filename(str): Name of fio file created on app pod's mount point
+            wait (bool): weather to wait for results or run in async mode
```
ebenahar: s/weather/Weather

tests/manage/test_pvc_disruptive.py
```diff
@@ -46,7 +47,11 @@ def disruptive_base(self, operation_to_disrupt, resource_to_delete):
         assert self.pod_obj.ocp.wait_for_resource(
             condition=constants.STATUS_RUNNING, resource_name=self.pod_obj.name, timeout=120
         )
-        self.pod_obj.run_io(storage_type=self.storage_type, size='1G')
+        self.pod_obj.run_io(
+            storage_type=self.storage_type, size='1G', wait=False
+        )
+        # wait for atleast 10 seconds for IO to start
```
ebenahar: maybe we can verify if the IO started from within run_io()?
For example, in ocs_ci.ocs.resources.pod.run_io_in_bg, we check for file existence after starting the IO of creating and writing to files - ocs_ci.ocs.resources.pod.check_file_existence

vasukulkarni: @ebenahar the run_io_in_bg is different, and run_io uses fio, I think we are safe here with 10 seconds of wait and testing.

vasukulkarni:
Tested with a modified test, I see there is another issue with build where we can't delete pvc backed by rbd

```
--------------------------------------------------------------------------------------------- live log call ----------------------------------------------------------------------------------------------
12:54:21 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /Users/vasukulkarni/cluster-path/auth/kubeconfig rsh pod-test-rbd-1112540265 which apt-get
12:54:23 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /Users/vasukulkarni/cluster-path/auth/kubeconfig rsh pod-test-rbd-1112540265 apt-get update
12:54:28 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /Users/vasukulkarni/cluster-path/auth/kubeconfig rsh pod-test-rbd-1112540265 apt-get -y install fio
12:54:35 - MainThread - ocs_ci.utility.utils - WARNING - Command warning:: debconf: delaying package configuration, since apt-utils is not installed
12:54:35 - ThreadPoolExecutor-1_0 - ocs_ci.utility.workloads.fio - INFO - Running cmd: fio --name=fio-rand-readwrite --filename=/var/lib/www/html/fio-rand-readwrite --readwrite=randrw --bs=4K --direct=1 --numjobs=1 --time_based=1 --runtime=60 --size=1G --iodepth=4 --invalidate=1 --fsync_on_close=1 --rwmixread=75 --ioengine=libaio --output-format=json
12:54:35 - MainThread - ocs_ci.ocs.workload - INFO - Done submitting..
12:54:35 - ThreadPoolExecutor-1_0 - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /Users/vasukulkarni/cluster-path/auth/kubeconfig rsh pod-test-rbd-1112540265 fio --name=fio-rand-readwrite --filename=/var/lib/www/html/fio-rand-readwrite --readwrite=randrw --bs=4K --direct=1 --numjobs=1 --time_based=1 --runtime=60 --size=1G --iodepth=4 --invalidate=1 --fsync_on_close=1 --rwmixread=75 --ioengine=libaio --output-format=json
12:54:35 - MainThread - ocs_ci.ocs.resources.pod - INFO - Waiting for fio job to complete
12:56:02 - MainThread - root - INFO - Waiting for results
12:56:02 - MainThread - root - INFO - IOPs after FIO:
12:56:02 - MainThread - root - INFO - Read: 1217.439128
12:56:02 - MainThread - root - INFO - Write: 406.429679
PASSED                                                                                                                                                                                             [100%]
------------------------------------------------------------------------------------------- live log teardown --------------------------------------------------------------------------------------------
12:56:02 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /Users/vasukulkarni/cluster-path/auth/kubeconfig delete PersistentVolumeClaim pvc-test-1112535699
```

tests/manage/test_pvc_disruptive.py
RazTamir: Why not using self.fio_thread.running()?

vasukulkarni: the thread will run in nano second, what I am waiting for here is for sometime so that I/O's are in flight. More sleep doesn't hurt here.

RazTamir: I don't want to overcomplicate things but:
1) what will happen if yum install fio hangs?
2) why not to create a simple function that actually checks that the df is changing and the disk size is being consumed.

We have to be accurate and that's why sleep is bad usage - everywhere

clacroix12: @RazTamir 

> what will happen if yum install fio hangs?

I assume `yum` itself would time out and we would end up raising a `CommandFailed` exception. I couldn't find where this install actually happens though so that assumption could be wrong. 

> why not to create a simple function that actually checks that the df is changing and the disk size is being consumed.

I think this is a good idea, we should be verifying actual disk activity.

> We have to be accurate and that's why sleep is bad usage - everywhere

Perhaps we could be more efficient and have the function that verifies disk activity return a new custom exception that our `retry` decorator expects and utilize that instead of just sleeping.

vasukulkarni: This is not *my* test :) (the test itself is quite disruptive here compared to other disruptive tests), but as a part of signature change, I noticed that some sleep is better than nothing to give I/O some chance - we are using stable tools to test I/O so I am not super concerned about the tool failure and we have routines to check for stats).

we have method to get the FIO stats but because of the nature of *this* test, I cannot use it here, I will try to add something here to be sure, but sleep as such is not a bad thing when used appropriately, its been used across various projects(including rook ;-).

keesturam: > I don't want to overcomplicate things but:
> 
> 1. what will happen if yum install fio hangs?

one shouldn't run yum install within a container. That's not the way containers are designed. I see that we do that as part of our test today. I'm going to raise an issue to use a custom container image with fio pre-installed. 

> 2. why not to create a simple function that actually checks that the df is changing and the disk size is being consumed.
> 
> We have to be accurate and that's why sleep is bad usage - everywhere

clacroix12: @RazTamir  Is this something we need to hold this PR up for or are we okay with moving forward with these changes and potentially addressing the concern around using sleep in a subsequent PR?

RazTamir: We can handle it later but we'll need to track it we new issue

vasukulkarni: I will file a ticket as follow on PR and merge this one after rebase and new testing.

clacroix12:
@vasukulkarni this will need a rebase

RazTamir:
@vasukulkarni do we still need this?

vasukulkarni:
I will try to address it this week and see what all new changes went in.

vasukulkarni:
Too old, will revisit again

PR #395, opened 2019-07-11
Use pytest log_cli_level for openshift-installer log level

clacroix12:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -74,22 +74,23 @@ def polarion_testsuite_properties(record_testsuite_property):
         )
 
 
-def cluster_teardown():
+def cluster_teardown(log_level="DEBUG"):
     log.info("Destroying the test cluster")
-    destroy_cluster(config.ENV_DATA['cluster_path'])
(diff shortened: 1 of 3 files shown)


clacroix12:
I have tested deployment as well as cluster destroy and both processes passed without issue. I can post logs if necessary.

tests/conftest.py
```diff
@@ -383,3 +384,15 @@ def environment_checker(request):
...
+@pytest.fixture(scope="session")
+def log_cli_level(pytestconfig):
+    """
+    Retrieves the log_cli_level set in pytest.ini
+
+    Returns:
+        str: log_cli_level set in pytest.ini or DEBUG if not set
+
+    """
+    return pytestconfig.getini('log_cli_level') or 'DEBUG'
```
vasukulkarni: can you test if it respects the cli option that pytest provides?

vasukulkarni:
Overall LGTM

tests/conftest.py
clacroix12: @vasukulkarni yes I did, and it does

PR #399, opened 2019-07-12
Fix .gitlab-ci

zmc:
It was failing because of the lack of the py36 interpreter

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,11 +4,14 @@ before_script:
    - pip install tox
 
 stages:
-   - build
+   - lint and unit test
 
-tox:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -4,11 +4,14 @@ before_script:
    - pip install tox
 
 stages:
-   - build
-
-tox:
-   stage: build
-   script:
-      - tox
-   tags:
```
clacroix12: Did we not need this docker tag?

zmc: Not that I can see! https://gitlab.com/zcerza/ocs-ci/pipelines/70773478

PR #405, opened 2019-07-15
Fix import of delete_all_pvcs

petr-balogh:
Change approach for PVCs fixtures and not using globals for them.

### DIFF
--- a/tests/manage/test_create_pvc_random_storage_class.py
+++ b/tests/manage/test_create_pvc_random_storage_class.py
@@ -49,8 +49,6 @@ def teardown():
     Tearing down the environment
     """
     global RBD_SECRET_OBJ, CEPHFS_SECRET_OBJ
-    log.info("Deleting PVC")
-    assert pvc.delete_all_pvcs()
     log.info("Deleting CEPH BLOCK POOL")
     assert helpers.delete_cephblockpool()
     log.info("Deleting RBD Secret")
@@ -85,17 +83,19 @@ def create_multiple_rbd_storageclasses(count=1):
     return True
 
 
-def create_pvc(storageclass_list, count=1):
+def create_pvc(storageclass_list, pvcs, count=1):
     """
     Function for creating pvc and multiple pvc
 
     Args:
         storageclass_list (list): This will contain storageclass list
+        pvcs (list): reference to list object where to append created pvcs
         count (int): count specify no of pvc want's to create
     """
     for i in range(count):
         sc_name = random.choice(storageclass_list)
         pvc_obj = helpers.create_pvc(sc_name)
+        pvcs.append(pvc_obj)
         log.info(f"{pvc_obj.name} got Created and got Bounded")
     return True
 
@@ -113,8 +113,20 @@ def create_storageclass_cephfs():
     return True
 
 
+@pytest.fixture(scope='class')
+def delete_pvcs(request):
+    cls_ref = request.node.cls
+
+    def fin():
+        log.info("Deleting PVCs")
+        if cls_ref.created_pvcs:
+            assert pvc.delete_pvcs(cls_ref.created_pvcs)
+    request.addfinalizer(fin)
+
+
 @tier1
 @pytest.mark.usefixtures(
+    delete_pvcs.__name__,
     ocs288_fixture.__name__,
 )
 @pytest.mark.polarion_id("OCS-288")
@@ -122,11 +134,12 @@ class TestCreatePVCRandomStorageClass(ManageTest):
     """
     Creating PVC with random SC
     """
+    created_pvcs = []
 
     def test_create_pvc_with_random_sc(self):
         storageclass_list = helpers.get_all_storageclass_name()
         if len(storageclass_list):
-            assert create_pvc(storageclass_list, count=20)
+            assert create_pvc(storageclass_list, self.created_pvcs, count=20)
         else:
             log.error("No Storageclass Found")
             return False
--- a/tests/manage/test_pvc_concurrent_deletion_creation.py
+++ b/tests/manage/test_pvc_concurrent_deletion_creation.py
@@ -8,7 +8,7 @@ from ocs_ci.ocs import constants, ocp, exceptions
 from ocs_ci.utility.utils import run_async
 from ocs_ci.framework.testlib import tier1, ManageTest
 from ocs_ci.ocs.resources.pod import get_ceph_tools_pod
-from ocs_ci.ocs.resources.pvc import create_multiple_pvc, delete_all_pvcs
+from ocs_ci.ocs.resources.pvc import create_multiple_pvc, delete_pvcs
 from tests.fixtures import (
     create_rbd_storageclass, create_ceph_block_pool, create_rbd_secret
 )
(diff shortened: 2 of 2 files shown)


petr-balogh:
I tried to run verification but see that those test cases failed on the same error as few days ago `Jul 12, 2019 10:32 PM`

**TC: test_create_pvc_with_random_sc**
[Here is link for execution with this PR](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/127/testReport/tests.manage.test_create_pvc_random_storage_class/TestCreatePVCRandomStorageClass/test_create_pvc_with_random_sc/)
[previous execution](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/121/testReport/tests.manage.test_create_pvc_random_storage_class/TestCreatePVCRandomStorageClass/test_create_pvc_with_random_sc/)

**test_multiple_pvc_concurrent_creation_deletion**
[With this PR](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/127/testReport/tests.manage.test_pvc_concurrent_deletion_creation/TestMultiplePvcConcurrentDeletionCreation/test_multiple_pvc_concurrent_creation_deletion/)
[previous execution](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/121/testReport/tests.manage.test_pvc_concurrent_deletion_creation/TestMultiplePvcConcurrentDeletionCreation/test_multiple_pvc_concurrent_creation_deletion/)

tests/manage/test_create_pvc_random_storage_class.py
```diff
@@ -85,17 +83,19 @@ def create_multiple_rbd_storageclasses(count=1):
     return True
 
 
-def create_pvc(storageclass_list, count=1):
+def create_pvc(storageclass_list, pvcs, count=1):
```
andrewschoen: I think what I'd probably do here is make a fixture out of ``create_pvc`` that is essentially this same code and returns a list of pvc objects. This "factory function" fixture can then handle the teardown for each pvc that it creates.

This way the tests that need to create pvcs can just request the fixture, call it  with a count to create as many pvcs as they need and can depend on pytest cleaning things up with finalizers.

Usage might look something like:

```
def test_foo(create_pvcs):
    pvcs = create_pvcs(count=20)
    # perform tests on the pvcs
```

Here's some docs on fixtures as factory functions that's a good reference: https://docs.pytest.org/en/latest/fixture.html#factories-as-fixtures

vasukulkarni: It makes sense to use functions and return them, we could easily reuse those functions for multiple cases, I am bit surprised how many tests are creating pv, pvcs, storageclass etc, when they could be just couple of reusable functions.

mbukatov: I agree with @andrewschoen

petr-balogh:
> I don't like pushing data into an object via a class reference, a fixture should pass a value via yield or return to the test function/method in clear, explicit way.
> 
> In [Coding guidelines for OCS-CI project](https://github.com/red-hat-storage/ocs-ci/blob/master/docs/coding_guidelines.md) we state:
> 
> > Code Readability takes priority, Lets avoid crazy hacks and tricks and stick to straight forward code blocks that are easy to understand for all.
> 
> Both global variable usage and pushing data into an object via a class reference is a hack
Isn't that what I am trying to avoid in this PR to using both you mentioned? As I didn't know about proposed solution by Andrew I just used the solution I knew and didn't wanted to change all the rest of the code in this initial PR without hearing yours opinion if you have different one. That's why we are discussing right now about this and we are getting nice input.
, which makes understanding and maintenance problematic. Hack which is equally problematic can't be considered a solution here.
> 
> We should also consider [proposal](https://github.com/red-hat-storage/ocs-ci/pull/405/files#r303483549) from @andrewschoen

Solution proposed by @andrewschoen looks good to me for this use case:
we need to have shared function providing some resources which should be created as a part of the test as the creation of pvc is here in this case part of test. But teardown needs to be done outside of the test in the fixture which can be done by proposed fixture factories functions pretty well.

But we have other use cases where we need to do creation of PVC as a setup of the test case so it shouldn't be called from the test method so this use case we cannot probably use this factory as the function needs to be called inside the the test method. 

In the PR where I just moved `project` and `pvcs` to fixture where PVC depends on project fixture as done [here](https://github.com/red-hat-storage/ocs-ci/pull/405/files#diff-3447e8cca8af24e7d2779be35775526d).  I didn't move the rest so far before see your opinions here.
But for example to access some class member attribute like pvc_base_name which was done in original PR is not a hack in my opinion and in some cases using another one fixture can maybe be additional complexity instead of just read one string attribute from class reference. (Of course I can imagine it as another fixture which will pvcs fixture will depend on)

tests/manage/test_create_pvc_random_storage_class.py
RazTamir: There's one problem with this function factory and is the fact that the creation of objects should happen in the setup and not part of the test.
In thi case, pvcs should be created in a setup phase and not be considered as part of the test

RazTamir:
Few comments,
1) usage of yield is not good for us as in case the setup will fail, the teardown will not be called as stated in the pytest fixtures documentation. 
2) creation of objects should be done in setup and not in the test itself 
3) deletions should be done in the teardown unless the test is about deleting objects. 
4) we need to also think about the simplest way that will be intuitive as much as possible for everyone
5) usage of request.node.cls is well documented and considered as valid way to access other scoped attributes. 

I saw cnv QE way to use request.node.cls specific attributes (not self.attr) so my question is, why do you think request.node.cls is a bad approach? (it works pretty good so far)

tests/manage/test_create_pvc_random_storage_class.py
clacroix12: @RazTamir Would something like this work for you?

```python
@pytest.fixture()
def multiple_pvcs(create_pvcs):
    return create_pvcs(count=20)

def test_foo(multiple_pvcs):
    for pvc in multiple_pvcs:
        pvc.do_whatever()
```

This keeps the creation of pvcs out of the test logic while benefitting from the customization of the factory style fixture. It just takes having another fixture which the test uses actually utilize the factory fixture to create them.

RazTamir: @clacroix12 , with that usage, how can I pass to multiple_pvcs() more parameters that will depend on the test needs? for example the namespace of the PVCs, size, etc.

mbukatov: Fixture `multiple_pvcs` from the example above would be specific for the particular test(s), so that if the test(s) needs more parameters, you will specify it in the fixture. This is in contrast with `create_pvcs`, which is very general.

andrewschoen:
> Few comments,
> 
> 1. usage of yield is not good for us as in case the setup will fail, the teardown will not be called as stated in the pytest fixtures documentation.

Agreed. The proposal I gave would use finalizers, so teardown would always happen.

> 2. creation of objects should be done in setup and not in the test itself

Declaring a fixture using a decorator on the class or requesting it in a function that needs it does the same thing, as long as we're talking about ``class`` level scope. If the fixture is ``function`` scope then each function will have it's own "setup". I don't see how my proposal disagrees with this.

> 3. deletions should be done in the teardown unless the test is about deleting objects.

Agreed. This happens in the finalizers created by the fixture.

> 4. we need to also think about the simplest way that will be intuitive as much as possible for everyone

I would contend having a library of well-documented fixtures that are well contained and reusable would be the easiest way. This way the QE team does not have to know about ``self`` or scoping or anything else. They just request the fixtures they need that do the actions they need for the specific test.

> 5. usage of request.node.cls is well documented and considered as valid way to access other scoped attributes.

It is valid, but maybe not for every use case. One drawback that you've brought up is needing to have some sort of standard on how we name class level properties to avoid collisions. We open ourselves up to the same kind of issues we saw with globals. This also locks us to ``class`` level scope for everything, which might not be what we want. The usage of fixtures in the way I've proposed is also well documented.

> I saw cnv QE way to use request.node.cls specific attributes (not self.attr) so my question is, why do you think request.node.cls is a bad approach? (it works pretty good so far)

As stated above it locks us to ``class`` scope and I believe that pytest already has better methods for handling what you're trying to do here, which is fixtures at a class level scope that are requested by test methods (when needed) and handle their own teardown.

RazTamir:
@petr-balogh do we need this?

petr-balogh:
No, this one was alrady fixed by another PR, this one was proposing to do it in new fixture style, but this is already covered by Filip's PR

Issue #415, opened 2019-07-17
Logs are placed to /tmp/ directory instead of to RUN['log_dir']


dahorak:
All logs from pytest are placed to `/tmp/...` directory, no matter what is configured in `RUN['log_dir']`. Only symlink is created from the particular `/tmp/...` directory to directory specified in `RUN['log_dir']`. This is an issue in Jenkins jobs, because logs are not properly placed to the expected place on NFS share and thus are not available.

dahorak:
The solution will be to use `--logger-logsdir` parameter passed automatically from the `run-ci` wrapper with proper argument.

PR #416, opened 2019-07-17
pass --logger-logsdir parameter to pytest

dahorak:
- place pytest logs to `RUN[`log_dir`]` to `pytest-logs-{run_id}` subdirectory
- generate `RUN['run_id']` directly in `run-ci` wrapper, so we can use the
  value for the `--logger-logsdir` parameter
- fixes https://github.com/red-hat-storage/ocs-ci/issues/415

### DIFF
--- a/ocs_ci/framework/main.py
+++ b/ocs_ci/framework/main.py
@@ -46,15 +46,14 @@ def init_ocsci_conf(arguments=None):
         with open(os.path.expanduser(cluster_config)) as file_stream:
             cluster_config_data = yaml.safe_load(file_stream)
             framework.config.update(cluster_config_data)
+    framework.config.RUN['run_id'] = int(time.time())
 
 
 def main(arguments):
     init_ocsci_conf(arguments)
-    run_id = int(time.time())
-    framework.config.RUN['run_id'] = run_id
     pytest_logs_dir = os.path.join(os.path.expanduser(
         framework.config.RUN['log_dir']
-    ), f'pytest-logs-{run_id}')
+    ), f"pytest-logs-{framework.config.RUN['run_id']}")
     utils.create_directory_path(pytest_logs_dir)
     arguments.extend([
         '-p', 'ocs_ci.framework.pytest_customization.ocscilib',
--- a/ocs_ci/framework/main.py
+++ b/ocs_ci/framework/main.py
@@ -1,4 +1,5 @@
 import os
+import time
 
 import pytest
 import yaml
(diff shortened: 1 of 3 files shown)


dahorak:
With this fix, the logs directory from Jenkins job (installation, tests and destroy) looks like this:

logs_directory

tests/conftest.py
```diff
@@ -48,8 +48,7 @@ def pytest_logger_config(logger_config):
     logger_config.set_formatter_class(OCSLogFormatter)
 
 
-ep_time = int(time.time())
-config.RUN['run_id'] = ep_time
+ep_time = config.RUN['run_id']
```
petr-balogh: This is not used anywhere else so we can probably just use this in log_dir on line 52:
`logs_{config.RUN['run_id']}`

ocs_ci/framework/main.py
```diff
@@ -49,10 +50,17 @@ def init_ocsci_conf(arguments=None):
 
 def main(arguments):
     init_ocsci_conf(arguments)
+    run_id = int(time.time())
```
petr-balogh: Maybe move line 53 and 54 to init_ocsci_conf as it's part of initial config and probably doesn't have to be in main function directly.

For the rest of the functionality we can maybe thin of move it to some init_ocsci function which will be called after init_ocsci_conf and move there lines 55-58 and maybe also 65-67.

WDYT?

tests/conftest.py
dahorak: Fixed by 6e0d644.

ocs_ci/framework/main.py
dahorak: First part is fixed by 6e0d644.

About the additional function `init_ocsci`, I'm not sure... If we will want to keep the `arguments.extend([...` part in the `main` function, we still need to define the `pytest_logs_dir`, which is basically the same how it is now... and if we will also move the `arguments.extend([...` part to the new function, there is not much what will left in the `main` function ... so not sure if it will bring more clarity or not...

petr-balogh: > First part is fixed by [6e0d644](https://github.com/red-hat-storage/ocs-ci/commit/6e0d644f1ce2d093d81cb592cce6b750b4f8afc3).
> 
> About the additional function `init_ocsci`, I'm not sure... If we will want to keep the `arguments.extend([...` part in the `main` function, we still need to define the `pytest_logs_dir`, which is basically the same how it is now... and if we will also move the `arguments.extend([...` part to the new function, there is not much what will left in the `main` function ... so not sure if it will bring more clarity or not...

I don't want to move arguments as they needs to be passed at the end, just lines I mentioned and if you will store this expanded path in the config, from main you can directly access from config and you don't need to expand users on multiple places if you will overwrite the config value.

In the feature we will need to add some other operation and I don't want to mess the main function here.

petr-balogh:
I will approve how it's now but if others thing as well that will be better to keep the main function clean and move other logic outside then please do the change otherwise merge when get second approve ;)

ocs_ci/framework/main.py
```diff
@@ -45,14 +46,20 @@ def init_ocsci_conf(arguments=None):
...
             cluster_config_data = yaml.safe_load(file_stream)
             framework.config.update(cluster_config_data)
+    framework.config.RUN['run_id'] = int(time.time())
 
 
 def main(arguments):
     init_ocsci_conf(arguments)
+    pytest_logs_dir = os.path.join(os.path.expanduser(
+        framework.config.RUN['log_dir']
+    ), f"pytest-logs-{framework.config.RUN['run_id']}")
```
vasukulkarni: s/pytest-logs/ocs-ci/

dahorak: @vasukulkarni the default location used by pytest is `/tmp/pytest-of-<username>/pytest-X/`, that was the reason, why I put the `pytest` prefix there.
If you think, it should rather contain `oc-ci` prefix, I would propose to use `ocs-ci-logs`, WDYT?

petr-balogh:
Let's merge this one as we are missing logs without this in our jenkins executions. In case something else needs to be addressed, please open issue to Daniel.  If we will in the future would tend to add more functionality to main function I would go with my suggestion about move the other code to different function and keep main function clean as much as possible.

PR #421, opened 2019-07-18
Wait for CBP creation

ebenahar:


tests/helpers.py
```diff
@@ -340,11 +342,14 @@ def verify_block_pool_exists(pool_name):
     """
     logger.info(f"Verifying that block pool {pool_name} exists")
     ct_pod = pod.get_ceph_tools_pod()
-    pools = ct_pod.exec_ceph_cmd('ceph osd lspools')
-    logger.info(f'POOLS are {pools}')
-    for pool in pools:
-        if pool_name in pool.get('poolname'):
-            return True
+
```
RazTamir: Add raises

ebenahar: I would like to return bool here. So I'll catch TimeoutExpiredError and return False to be caught outside with assert in case the pool is not found

PR #422, opened 2019-07-18
Add delete flag to OCS class

petr-balogh:
This will be useful for finalizers/teardowns as you can skip this object
which was already deleted from test method for example.


ocs_ci/ocs/resources/ocs.py
```diff
@@ -15,6 +15,9 @@ class OCS(object):
     """
     Base OCSClass
     """
+    # This _delete flag is set to True if the delete method was called on
+    # object of this class and was successfull.
+    _deleted = False
```
vasukulkarni: s/_deleted/deleted or is_deleted, since we are planning to use it outside class.

petr-balogh: Done by property

PR #427, opened 2019-07-19
Change yaml.load to yaml.safe_load

### LINKED ISSUE #378: yaml.load() without Loader=... is deprecated
yaml.load() without Loader=... is deprecated
I think we should change yaml.load() to yaml.safe_load() or yaml.full_load()

ebondare:
Fix #378

### DIFF
--- a/ocs_ci/ocs/monitoring.py
+++ b/ocs_ci/ocs/monitoring.py
@@ -23,7 +23,7 @@ def create_configmap_cluster_monitoring_pod(sc_name):
     config_map = templating.load_yaml_to_dict(
         constants.CONFIGURE_PVC_ON_MONITORING_POD
     )
-    config = yaml.load(config_map['data']['config.yaml'])
+    config = yaml.safe_load(config_map['data']['config.yaml'])
     config['prometheusK8s']['volumeClaimTemplate']['spec']['storageClassName'] = sc_name
     config['alertmanagerMain']['volumeClaimTemplate']['spec']['storageClassName'] = sc_name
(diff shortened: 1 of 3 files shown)


ebondare:
Link to jenkins build: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/175/

clacroix12:
There are two other instances of calling `yaml.load` that I can see in the repo:

https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/prometheus.py#L71
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/templating.py#L20

I see that we specify the loader that is used for them though. Is that acceptable and the reason why they aren't being changed to `safe_load` as well?

ebondare:
> There are two other instances of calling `yaml.load` that I can see in the repo:
> 
> https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/prometheus.py#L71
> https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/templating.py#L20
> 
> I see that we specify the loader that is used for them though. Is that acceptable and the reason why they aren't being changed to `safe_load` as well?

Yes, I think it is acceptable.
https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation
Only the use of PyYAML's yaml.load function without specifying the Loader=... parameter has been deprecated, so I didn't change the lines where the loader was specified.

PR #418, opened 2019-07-17
Update fixtures from fixtures.py

fbalak:
I suggest this change to improve usability and readability of fixtures. I believe that we should try not to presume in fixtures that they are used in class so fixtures shouldn't contain any parent class reference. I tried to update fixtures from `tests/fixtures.py` to be used as described in https://docs.pytest.org/en/latest/fixture.html.

- Move fixtures from `tests/fixtures.py` to `tests/conftest.py`.
- Remove direct imports of fixtures. Now the fixtures can be called in tests or fixtures as parameter.
- Add links to parent objects in hierarchy into fixtures (e.g. from `rbd_pod` can be `rbd_secret` called like `rbd_pod_factory().pvc.storageclass.secret`).
- Turn fixtures into factory functions which create resource and handles teardown.
- Update tests that used those fixtures.
- Fixes https://github.com/red-hat-storage/ocs-ci/issues/417
- Fixes https://github.com/red-hat-storage/ocs-ci/issues/263
- Fixes https://github.com/red-hat-storage/ocs-ci/issues/200

Signed-off-by: fbalak <fbalak@redhat.com>

PR #428, opened 2019-07-19
Adding examples and documentation for fixtures

petr-balogh:


docs/fixture_usage.md
```diff
@@ -0,0 +1,119 @@
...
+The finalizer should be added as soon as possible to make sure the finalizer
+will be called at the end. (Otherwise we will hit the same issue as we have
+with yield)
+
+## Use cases
+
+TODO: Mention function factories and use cases and when to use which approach.
+Something like:
+
+* Setup resource,  test (using resources) , teardown - regular usage of fixture
```
clacroix12:
```suggestion
* Setup resource, test (using resources), teardown - regular usage of fixture
```

docs/fixture_usage.md
```diff
@@ -0,0 +1,119 @@
...
+  shouldn’t fail on those deleted resources
+
+## Avoid
+
+Here is the list of things we should avoid in our fixtures:
+
+* Using references to `request.node.cls` for set, altering or reading class
+  attributes.
+* Using globals for sharing data.
+* Using finalizer after yield.
```
clacroix12: We won't be using `yield` at all if we are enforcing the use of `request.addfinalizer()`.

docs/fixture_usage.md
```diff
@@ -0,0 +1,119 @@
...
+* Using `@pytest.mark.usefixtures`.
+* Using more than one assert in the teardown when another teardown action is
+  between them. (This will lead that the rest of resources won't be cleaned
+  once assertion exception raised).
+
+## Examples of fixtures and tests
+
+Here are few simple examples how to use fixtures.
```
clacroix12:
```suggestion
Here are few simple examples of how to use fixtures.
```

docs/python-examples/fixtures/conftest.py
```diff
@@ -0,0 +1,109 @@
...
+
+    request.addfinalizer(fin)
+    return pvc
+
+
+@pytest.fixture(scope='class')
+def pvcs_factory(request, storage_class):
+    """
+    Fixture factory for creating pvcs.
+    This fixture is returning function with with which you can create objects
```
clacroix12:
```suggestion
    This fixture returns a function with with which you can create objects
```

docs/python-examples/fixtures/conftest.py
```diff
@@ -0,0 +1,109 @@
...
+    def fin():
+        if not pvc.is_deleted:
+            pvc.delete()
+
+    request.addfinalizer(fin)
+    return pvc
+
+
+@pytest.fixture(scope='class')
+def pvcs_factory(request, storage_class):
```
clacroix12: Would a naming convention of `create_{resource}` be better for factories instead of `{resource}_factory`? When a test actually uses the function I think it makes more sense for it to be `pvc = create_pvc(storage_class)` than `pvc = pvc_factory(storage_class)`

docs/python-examples/fixtures/conftest.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    Fixture factory for creating pvcs.
+    This fixture is returning function with with which you can create objects
+    of PVC as a part of test and cares about teardown of created PVCs.
+
+    Usage of this fixter is for usecases when:
```
clacroix12:
```suggestion
    Usage of this fixture is for use cases when:
```

docs/python-examples/fixtures/conftest.py
```diff
@@ -0,0 +1,109 @@
...
+        teardown of created objects.
+
+        Args:
+            count (int): desired count of pvcs to be created
+            storage_class (StorageClass): reference to storage class object
+            pvcs (list): reference to pvcs list where to append all created
+                pvcs by helper function ocs.create_pvcs
+        """
+        ret_value = ocs.create_pvcs(count, storage_class)
+        pvcs += ret_value
```
clacroix12:
```suggestion
        pvcs.append(ret_value)
```

docs/python-examples/fixtures/ocs.py
```diff
@@ -0,0 +1,71 @@
...
+        self.is_deleted = False
+
+    def delete(self):
+        logger.info(f"Deleting pvc: {self.name}")
+        self.is_deleted = True
+
+
+class StorageClass:
+    """
+    Example dummy class of PVC
```
clacroix12:
```suggestion
    Example dummy class of StorageClass
```

docs/python-examples/fixtures/ocs.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    def __init__(self, name):
+        """
+        Constructor for PVC
```
clacroix12:
```suggestion
        Constructor for StorageClass
```

docs/python-examples/fixtures/ocs.py
```diff
@@ -0,0 +1,71 @@
...
+
+        Args:
+            name (str): name of pvc
```
clacroix12:
```suggestion
            name (str): name of storage class
```

docs/python-examples/fixtures/test_fixtures.py
```diff
@@ -0,0 +1,109 @@
...
+    logger.info(f"Doing something with pvc: {pvc}")
+
+
+@pytest.fixture(scope='class')
+def three_pvcs(pvcs_factory, storage_class):
+    """
+    This is fixture related just to one specific test in this module so no
+    need to have it defined in conftest
+    """
+    return pvcs_factory(3, storage_class)
```
clacroix12: `pvcs_factory` only returns one pvc at a time, you'll need to call it several times in this example or update the fixture definition to support passing a count. I know these are only examples but I can see someone getting confused by this. Same goes for the example test cases below that use the factory.

docs/python-examples/fixtures/test_fixtures.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class TestCreatingPVCsFromTest:
+    # This needs to be discussed! How to solve the issue of sharing resources
```
clacroix12: Not sure I understand the question, what is the issue with sharing resources between tests when the fixture is class scope?

docs/python-examples/fixtures/test_fixtures.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        pvcs_created = pvcs_factory(pvcs_number, storage_class)
+        len(pvcs_created) == pvcs_number
+        logger.info([p.name for p in pvcs_created])
+
+    def test_share_pvcs(self, pvcs_factory, storage_class):
+        my_shared_pvcs = pvcs_factory(2, storage_class)
+        logger.info(f"Shared pvcs usage: {[p.name for p in my_shared_pvcs]}")
+
+        # Is this acceptable to do below for share PVCs to test 3rd test?
```
clacroix12: I think if multiple tests need to 'share' the output from a class scope factory fixture, we should create a wrapper fixture that uses the factory to create the shared resources and returns the list of them. That way the tests that need the list of resources can just use the wrapper fixture to get the list. For example these tests would use the`three_pvcs` fixture to get the list of 3 pvcs and not the `pvc_factory` fixture directly.

docs/python-examples/fixtures/test_fixtures.py
```diff
@@ -0,0 +1,109 @@
...
+
+        # If you need to do something with pvc outside class you need to just
+        # send whatever as reference like in example bellow:
+        pvc_to_alter = self.shared_pvcs[1]
+        alter_pvc(pvc_to_alter)
+
+
+class TestPVCsCreatedInSetup:
+    """
+    This is example of test which needs 3 PVCs and do some operation on top of
```
clacroix12:
```suggestion
    This is an example of a test which needs 3 PVCs and does some operation on top of
```

docs/fixture_usage.md
```diff
@@ -0,0 +1,119 @@
+# Guidelines for fixtures usage in OCS-CI
+
+## Basic information
+
+What is pytest Fixture? You can find definition and examples documentation in
```
clacroix12:
```suggestion
What is a pytest Fixture? You can find definition and examples documentation in
```

docs/python-examples/fixtures/test_fixtures.py
petr-balogh: Pvcs_factory returns the function which is called here and parametrized to give you three pvcs.

docs/fixture_usage.md
petr-balogh: You can in some cases use fixture with yield if you do not need to do any cleanup/teardown on maybe.  That's why I wrote it like that.

docs/python-examples/fixtures/conftest.py
petr-balogh: Sure

docs/python-examples/fixtures/conftest.py
petr-balogh: Ret value is a list of pvcs, cannot do append otherwise pvcs will look like [[pvc1, pVc2]] and after second call [[pvc1, pvc2],[pvc3, pvc4, pvc5]] for example.

docs/python-examples/fixtures/test_fixtures.py
petr-balogh: What about case those resources needs to be created as a part of test?

docs/fixture_usage.md
clacroix12: If there isn't any cleanup/teardown to perform we would use `return` and not `yield`. I think any scenario in which we would need yield would mean we need to add a finalizer instead.

docs/python-examples/fixtures/conftest.py
clacroix12: Ah, good call

docs/python-examples/fixtures/test_fixtures.py
clacroix12: I see that now, I was looking at the wrong definition, scratch this!

docs/python-examples/fixtures/test_fixtures.py
clacroix12: In this scenario you have one test that needs to create the resources, and other tests that need to utilize those created in the other test. If that's the case then sure using a class attribute to pass those created resources from one test to another is fine. As long as the fixture isn't altering that class attribute directly and it's done at the test level like you have here.

docs/fixture_usage.md
petr-balogh: Done

docs/fixture_usage.md
```diff
@@ -0,0 +1,119 @@
...
+[pytest web](https://docs.pytest.org/en/latest/fixture.html). Please read this
+as the first.
```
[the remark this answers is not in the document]
petr-balogh: Done

docs/fixture_usage.md
```diff
@@ -0,0 +1,119 @@
...
+
+We should keep our fixtures organized and have all the dependencies well
+defined in fixtures.
+
+All the fixtures should be documented - mandatory is document return value.
```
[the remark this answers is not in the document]
petr-balogh: Done

docs/fixture_usage.md
petr-balogh: Done

docs/fixture_usage.md
petr-balogh: done

docs/python-examples/fixtures/conftest.py
petr-balogh: done

docs/python-examples/fixtures/conftest.py
petr-balogh: done

docs/python-examples/fixtures/conftest.py
petr-balogh: done

docs/python-examples/fixtures/ocs.py
petr-balogh: done

docs/python-examples/fixtures/ocs.py
petr-balogh: done

docs/python-examples/fixtures/ocs.py
petr-balogh: done

docs/python-examples/fixtures/test_fixtures.py
petr-balogh: This comment is here to discuss the approach we are discussing in next comment on line 47.

docs/python-examples/fixtures/test_fixtures.py
petr-balogh: done

docs/python-examples/fixtures/ocs.py
```diff
@@ -0,0 +1,71 @@
+import random
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_random_name(obj_type):
+    """
+    Random
```
ebenahar: Add Returns

docs/python-examples/fixtures/ocs.py
```diff
@@ -0,0 +1,71 @@
...
+        list: PVCs objects
+    """
+    _pvcs = []
+    for number in range(count):
+        pvc_name = get_random_name('pvc')
+        _pvcs.append(PVC(pvc_name, storage_class))
+    return _pvcs
+
+
+class PVC:
```
ebenahar: Why do we need these classes if we can use helpers.py create resource functions?

petr-balogh: It's just and dummy class which represent the resource class for the examples of fixtures. Those are just short dummy examples and I didn't add helpers.py module for simplicity here and have helper function which creates resources directly in this ocs.py example module.

docs/python-examples/fixtures/ocs.py
petr-balogh: done

docs/python-examples/fixtures/conftest.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    sc_name = ocs.get_random_name('storage_class')
+    sc = ocs.StorageClass(sc_name)
+    logger.info(f"Creating storage class: {sc.name}")
+
+    def fin():
+        if not sc.is_deleted:
+            sc.delete()
+
+    request.addfinalizer(fin)
```
RazTamir: This should be the first line being executed in the fixture

docs/python-examples/fixtures/conftest.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    pvc_name = ocs.get_random_name('pvc')
+    pvc = ocs.PVC(pvc_name, storage_class)
+    logger.info(f"Creating pvc: {pvc.name}")
+
+    def fin():
+        if not pvc.is_deleted:
+            pvc.delete()
+
+    request.addfinalizer(fin)
```
RazTamir: same - please fix all

docs/python-examples/fixtures/test_fixtures.py
```diff
@@ -0,0 +1,112 @@
...
+class TestCreatingPVCsFromTest:
+    # If the test suite needs to share some resources it's possible to do it
+    # via class member like below, when you fill resources from test case.
+    # We just need to document that test suit needs run like whole cause if
+    # some test case depends on resources created by other test case in same 
+    # test suite (class) it's not possible to run just last test case.
+    shared_pvcs = []
+
+    @pytest.mark.parametrize("pvcs_number", (2, 4))
+    def test_create_pvcs(self, pvcs_number, create_pvcs, storage_class):
```
RazTamir: this create_pvcs is the fixture that uses the function factory fixture?
This is going to be a nightmare

docs/fixture_usage.md
andrewschoen: I agree with @clacroix12 here, I don't really see a strong reason to use ``yield`` at all.

docs/python-examples/fixtures/test_fixtures.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        You can access all needed resources via fixtures as parameters in
+        method definition like in example above (create_pvcs or
+        storage_class).
+        """
+        pvcs_created = create_pvcs(pvcs_number, storage_class)
+        len(pvcs_created) == pvcs_number
```
andrewschoen: did you mean to ``assert`` here?

docs/fixture_usage.md
petr-balogh: ok, removing in next commit

docs/python-examples/fixtures/test_fixtures.py
petr-balogh: It's not that bad I think. create_pvcs is fixture factory which return you the wrapper function defined in fixture factory which allow you to collect the created object for the finalizer once the function is called. Take a look here:  https://github.com/red-hat-storage/ocs-ci/pull/418/files @fbalak is already implementing it in his PR

docs/python-examples/fixtures/conftest.py
petr-balogh: done

docs/python-examples/fixtures/conftest.py
petr-balogh: done

docs/python-examples/fixtures/test_fixtures.py
andrewschoen: If a fixture is a ``class`` scope you still won't need that class attribute though. If the first test uses the ``class`` level fixture and modifies it, then the next class will get that modified version. But to @clacroix12 point, using a class attribute is still a feasible solution if done inside a test explicitly. That's a lot less confusing than fixtures that magically change class attributes.

docs/python-examples/fixtures/conftest.py
```diff
@@ -0,0 +1,111 @@
...
+    sc_name = ocs.get_random_name('storage_class')
+    sc = ocs.StorageClass(sc_name)
+    logger.info(f"Creating storage class: {sc.name}")
+
+    def fin():
+        if not sc.is_deleted:
+            sc.delete()
+
+    request.addfinalizer(fin)
+    return sc_name
```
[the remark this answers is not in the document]
petr-balogh: it's just example as I don't need the object in example I just returned the name.

docs/python-examples/fixtures/conftest.py
petr-balogh: I thought as well but tested it and it really works. Just download the examples and run ;)

docs/python-examples/fixtures/test_fixtures.py
petr-balogh: Yep, will fix in next commit once I will send.

docs/python-examples/fixtures/test_fixtures.py
clacroix12: > this create_pvcs is the fixture that uses the function factory fixture?
> This is going to be a nightmare

It's really no different from a fixture calling a bunch of utility methods to create resources except we get the additional benefit of the factory fixture finalizers cleaning everything up after we are done with them.

docs/python-examples/fixtures/conftest.py
```diff
@@ -0,0 +1,111 @@
...
+    def fin():
+        if not pvc.is_deleted:
+            pvc.delete()
+
+    request.addfinalizer(fin)
+    return pvc
+
+
+@pytest.fixture(scope='class')
+def create_pvcs(request, storage_class):
```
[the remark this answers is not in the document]
petr-balogh: @andrewschoen the reason here was to really make sure the order for finalizer (forgot to mention this in comment), if I didn't specified it here you can see that storage class was deleted in different order that PVC in example here where you have some vimdiff output from two executions, on the left side it's with defined storage_class here in factory and on right side without.    https://paste.fedoraproject.org/paste/Gc8WLgbYBTZkkkcybZlK4Q .  

If we will use just storage class from test how we will make sure that the required resource will be deleted in right order? as we need to create in order . SC, PVC and delete in order PVC, SC for example and there will be more complex matrix I think for some resources. 

We need to make sure that our fixture factories will be able to be parametrized and we need to keep order for setup and delete the resources in right way.

docs/python-examples/fixtures/test_fixtures.py
petr-balogh: @andrewschoen problem here is that the fixture factory returns the function, so when you want to create some resources by this function in one of the test and in the other test you simply cannot get the same resources as if you will use create_pvcs_factory then you will get the same instance of function for the class but not the resources created in previous call from different test in that same class.

docs/python-examples/fixtures/conftest.py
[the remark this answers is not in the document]
clacroix12: > If we will use just storage class from test how we will make sure that the required resource will be deleted in right order? as we need to create in order . SC, PVC and delete in order PVC, SC for example and there will be more complex matrix I think for some resources.

I don't think we should design our fixtures to cover both of these cases. There's really no reason to make things more complicated than they need to be. Either the pvc fixture can depend on the storage_class fixture as a dependency and handle cleanup itself, or it can require the test to depend on the storage_class fixture and pass that to the function the pvc fixture returned. In either case the resources are cleaned up through their respective fixture finalizers.

docs/python-examples/fixtures/conftest.py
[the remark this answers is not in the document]
petr-balogh: I updated it in next commit to return object to not be confusing

docs/python-examples/fixtures/conftest.py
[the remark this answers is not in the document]
petr-balogh: @andrewschoen @clacroix12  As we need to keep the storage class fixture parameter in this factory to make proper order of cleaning I updated tests to use it like this.  Also made the factory as a function returning just 1 PVC, so if user needs more of them he should create them in some loop from the test or fixture like I have three_pvcs fixture.  Please take a look and tell me if it looks better now.

docs/python-examples/fixtures/conftest.py
```diff
@@ -0,0 +1,102 @@
...
+    pvc_name = ocs.get_random_name('pvc')
+    logger.info(f"Creating pvc: {pvc_name}")
+    pvc = ocs.PVC(pvc_name, storage_class, some_parameter="Default value")
+    return pvc
+
+
+@pytest.fixture(scope='class')
+def pvc_factory(request, storage_class):
+    """
+    Fixture factory for creating pvcs.
```
vasukulkarni: Not a big fan of this due to (1), It adds complexity for many qe folks on how to use it 
(2) can be easily handled at test level without need for this.

clacroix12: It _can_ be handled at the test level but the point is that by utilizing these fixtures the tests no longer have to care about cleaning up resources OR creating that resources dependencies as well, and we don't need to ensure that tests are doing so properly. By using the factory fixture they can dictate when they create resources and let the finalizer clean up correctly based on scope.

I'm not seeing the added complexity, usage is actually quite simple:

```
def my_test(pvc_factory):
    pvc = pvc_factory()
    # now the test can do whatever it wants with this pvc object and not care about cleaning up
```

vasukulkarni: Works for me :)

docs/python-examples/fixtures/test_fixtures.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    Dummy helper function to alter pvc
+
+    Args:
+        pvc (PVC): PVC object reference
+    """
+    logger.info(f"Doing something with pvc: {pvc}")
+
+
+@pytest.fixture(scope='class')
```
vasukulkarni: I will add some commit on top of this to cleanup some examples like this which are confusing

docs/python-examples/fixtures/test_fixtures.py
```diff
@@ -0,0 +1,134 @@
+import logging
+
+import pytest
+
+logger = logging.getLogger(__name__)
+
+
+def alter_pvc(pvc):
+    """
+    Dummy helper function to alter pvc
```
vasukulkarni: dont think this test is necessary due to class methods for altering within test.

docs/python-examples/fixtures/ocs.py
```diff
@@ -0,0 +1,73 @@
...
+        self.is_deleted = False
+
+    def delete(self):
+        logger.info(f"Deleting pvc: {self.name}")
+        self.is_deleted = True
+
+
+class StorageClass:
+    """
+    Example dummy class of StorageClass
```
vasukulkarni: This is framework class example, I think its really not needed in this PR.

docs/python-examples/fixtures/ocs.py
```diff
@@ -0,0 +1,73 @@
...
+    Returns:
+        list: PVCs objects
+    """
+    pvc_name = get_random_name('pvc')
+    return PVC(pvc_name, storage_class, some_parameter)
+
+
+class PVC:
+    """
+    Example dummy class of PVC
```
vasukulkarni: This is a framework class example, not really needed in this PR.

docs/python-examples/fixtures/conftest.py
```diff
@@ -0,0 +1,102 @@
...
+    This fixture returns function with which you can create objects of PVC as
+    a part of test and cares about teardown of created PVCs.
```
fbalak:
```suggestion
    a part of test and takes care of teardown for created PVCs.
```

docs/python-examples/fixtures/conftest.py
```diff
@@ -0,0 +1,102 @@
...
+
+    Usage of this fixture is for use cases when:
+    * you need to create PVC resources as part of test and the teardown is
+      done automatically by its teardown
+    * you need to create PVC resources as part of test and you can also delete
```
fbalak:
```suggestion
    * you need to create PVC resources as part of test and you can also delete them
```

docs/python-examples/fixtures/test_fixtures.py
fbalak: This one looks ok. It can be useful to see how to created basic setup fixture providing more pregenerated resources. On the other hand, developer can look into some test in `tests` directory and look for some examples there.

docs/fixture_usage.md
```diff
@@ -0,0 +1,148 @@
...
+Fixture is compose of its setup part, the code which is done for prepare
+resource and teardown part which is after yield of added via
+`request.addfinalizer` method. It's strictly forced to use `addfinalizer`
+method cause it can happen that what is after yield won't be proceed when
+exception will be thrown in the middle of the setup.
+
+The finalizer should be added as soon as possible to make sure the finalizer
+will be called at the end. (Otherwise we will hit the same issue as we have
+with yield)
+
```
fbalak: I would add paragraph about [usage of `contest.py`](https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions) file and about ability to create it in each test directory and that fixtures contained in it can be used in all tests from all child directories. So there can be different conftest for manage tests and for ecosystem. In root directory can be conftest that contains fixtures for all tests.

docs/python-examples/fixtures/ocs.py
petr-balogh: As all of those are just examples I don't see any issue to have those dummy classes to represent some object for those examples as I am using it in finalizers with delete method and is_deleted flag which is in my opinion more understandable/readable with this example.

docs/python-examples/fixtures/test_fixtures.py
petr-balogh: same

docs/python-examples/fixtures/test_fixtures.py
petr-balogh: This specific case was discussed on automation mtg AFAIR and that's why I added it here, so don't think its confusing, or can you elaborate what is confusing here? Please before sending additional commit to this PR, please raise the comment about what is confusing or why and how it should be changed to have some discussion before putting effort to changing stuff. I see that this is already approved by @andrewschoen and @clacroix12 . Anyway I see some suggestions from @fbalak which I will fix tomorrow.

vasukulkarni: we have pvc_factory that is an example for creating "N" objects, so we have to  Also, I touched about this topic in our last meeting, we wanted to document in simple terms for QE folks on how they can follow new fixture methods and not get into details of how classes are designed(it was never discussed afaik)  and how delete is handled for them, 

Also,  is_delete flag is unnecessary, since its a class level feature(outside the scope of finalizer) and whether you check in finalizer or not it doesn't matter.

I have pushed my commit on top, we can discuss on monday and this I also want to document in gdoc. This has *wider* implications for all team members, So I want to ensure everybody is getting the core concept of what we are saying here.

vasukulkarni: Here is my diff: https://github.com/red-hat-storage/ocs-ci/pull/428/commits/c731450a2872cee90a11843d574a2775d872b05b

docs/python-examples/fixtures/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+@pytest.fixture(scope='class')
+def storage_class(request):
+    """
+    Storage class fixture
+
+    Returns:
+        StorageClass: object of storage class
+    """
+    def fin():
+            sc.delete()
```
alfredodeza: is this using a `TAB` or 8 spaces? looks odd here and on other `fin()` calls

petr-balogh: There was one line before sc.delete with if condition which was deleted by @vasukulkarni here: https://github.com/red-hat-storage/ocs-ci/commit/c731450a2872cee90a11843d574a2775d872b05b#diff-ae168d5153b97475ccfbc0d0a3d6ad76L20

PR #437, opened 2019-07-23
Knip 821

wusui:
Make sure ceph cluster works on minimal test and extract baseline fio values for comparison with later tests.


wusui:
I'll sync before I push this again.

Issue #436, opened 2019-07-22
Deployment logs are not being generated


clacroix12:
I am attempting to execute a fresh deployment with only the required arguments being passed to `run-ci`. Live logs are not appearing in the console, as well as log files are not being generated in the log directory for the test run.

This is the entire output I am seeing from the fresh deployment. I suspect the internal error is due to the fact that there are actually no file handlers. 
```
venv) ➜  ocs-ci git:(master) ✗ run-ci -m deployment --cluster-path ~/clusters/deploy-test-1 --deploy
The kubeconfig file /Users/clacroix/clusters/deploy-test-1/auth/kubeconfig doesn't exist!
=========================================================================================================== test session starts ============================================================================================================
platform darwin -- Python 3.7.3, pytest-5.0.1, py-1.8.0, pluggy-0.12.0
rootdir: /tmp/pytest-logs-1563813071
plugins: marker-bugzilla-0.9.1.dev2, reportportal-1.0.5, html-1.21.1, metadata-1.8.0, logger-0.5.0
collected 74 items / 73 deselected / 1 selected

../../../../tmp/pytest-logs-1563813071/tests/ecosystem/deployment/test_ocs_basic_install.py
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/Users/clacroix/workspace/ocs-ci/venv/lib/python3.7/site-packages/_pytest/main.py", line 213, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>   File "/Users/clacroix/workspace/ocs-ci/venv/lib/python3.7/site-packages/_pytest/main.py", line 257, in _main
INTERNALERROR>     config.hook.pytest_runtestloop(session=session)
INTERNALERROR>   File "/Users/clacroix/workspace/ocs-ci/venv/lib/python3.7/site-packages/pluggy/hooks.py", line 289, in __call__
INTERNALERROR>     return self._hookexec(self, self.get_hookimpls(), kwargs)
INTERNALERROR>   File "/Users/clacroix/workspace/ocs-ci/venv/lib/python3.7/site-packages/pluggy/manager.py", line 87, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook, methods, kwargs)
INTERNALERROR>   File "/Users/clacroix/workspace/ocs-ci/venv/lib/python3.7/site-packages/pluggy/manager.py", line 81, in <lambda>
[...]

PR #444, opened 2019-07-23
Deployment logging fix

clacroix12:
This PR addresses the issues documented in #436. I still wasn't able to determine with 100% certainty what the cause of the issues was, although I highly suspect it was some combination of setting the log_dir through `--logging-logsdir` as well as defining `pytest_logger_logdirlink` in our `conftest.py` which enables symlinking our logs. 

- Removed sym linking of pytest logs
- Moved log directory path construction to utility method
- Removed explicit call to create the log directory as pytest handles this for us

Fixes: #436 

### DIFF
--- a/ocs_ci/framework/main.py
+++ b/ocs_ci/framework/main.py
@@ -52,6 +52,7 @@ def init_ocsci_conf(arguments=None):
 def main(arguments):
     init_ocsci_conf(arguments)
     pytest_logs_dir = utils.ocsci_log_path()
+    utils.create_directory_path(framework.config.RUN['log_dir'])
     arguments.extend([
         '-p', 'ocs_ci.framework.pytest_customization.ocscilib',
         '-p', 'ocs_ci.framework.pytest_customization.marks',
--- a/ocs_ci/framework/main.py
+++ b/ocs_ci/framework/main.py
@@ -51,10 +51,7 @@ def init_ocsci_conf(arguments=None):
 
 def main(arguments):
     init_ocsci_conf(arguments)
-    pytest_logs_dir = os.path.join(os.path.expanduser(
-        framework.config.RUN['log_dir']
-    ), f"pytest-logs-{framework.config.RUN['run_id']}")
-    utils.create_directory_path(pytest_logs_dir)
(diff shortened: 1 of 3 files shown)


clacroix12:
I'd still like to test the following things before merging this but so far everything seems functional.
- [ ] XML reports
- [ ] HTML reports
- [ ] Email summary
- [ ] --collect-logs

RazTamir:
Hi Coady,

Sorry if I already asked, but can we remove the functionality that was added in the PR that caused this regression?

ocs_ci/framework/main.py
```diff
@@ -51,10 +51,7 @@ def init_ocsci_conf(arguments=None):
 
 def main(arguments):
     init_ocsci_conf(arguments)
-    pytest_logs_dir = os.path.join(os.path.expanduser(
-        framework.config.RUN['log_dir']
-    ), f"pytest-logs-{framework.config.RUN['run_id']}")
-    utils.create_directory_path(pytest_logs_dir)
```
zmc: We still need to create `log_dir`

zmc:
@RazTamir Which PR are you referring to?

clacroix12:
@zmc he's referring to #416 

@RazTamir The issue is that those changes are necessary for our jenkins jobs to run. I think implementing a change to fix it for both local and jenkins runs is better than removing that commit entirely before getting the fix in.

zmc:
Ah. I imagine the issue was caused by using that flag while continuing to use the symlinks.

clacroix12:
Confirmed log files appear for deployment, tests, and teardown.

https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/203

dahorak:
> Hi Coady,
> 
> Sorry if I already asked, but can we remove the functionality that was added in the PR that caused this regression?

As mentioned @clacroix12 above, without the parameter `--logger-logsdir` (used in #416), pytest saves all logs to `/tmp/...` directory, but in Jenkins job we need to be able to define custom directory to save the logs and all other files to NFS Share, so it will be accessible when the job finishes.

PR #450, opened 2019-07-24
Fix Issue #449 - KeyError: 'nodeName' during pod creation

sidhant-agrawal:
Handle the condition when 'nodeName' parameter is not present in pod YAML template.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -128,10 +128,12 @@ def create_pod(
     pod_data['metadata']['namespace'] = namespace
     if pvc_name:
         pod_data['spec']['volumes'][0]['persistentVolumeClaim']['claimName'] = pvc_name
+
     if node_name:
         pod_data['spec']['nodeName'] = node_name
     else:
-        del pod_data['spec']['nodeName']
+        if 'nodeName' in pod_data.get('spec'):
+            del pod_data['spec']['nodeName']
 
     pod_obj = pod.Pod(**pod_data)
     pod_name = pod_data.get('metadata').get('name')


tests/helpers.py
```diff
@@ -128,10 +128,14 @@ def create_pod(
     pod_data['metadata']['namespace'] = namespace
     if pvc_name:
         pod_data['spec']['volumes'][0]['persistentVolumeClaim']['claimName'] = pvc_name
+
     if node_name:
         pod_data['spec']['nodeName'] = node_name
     else:
-        del pod_data['spec']['nodeName']
```
ebenahar: you can replate try and except here with:

```python
if 'nodeName' in data['spec']:
    del pod_data['spec']['nodeName']
```

sidhant-agrawal: Done

RazTamir: Not sure I understand this change.
pod_data['spec']['nodeName'] is part of the yaml file by default and in case you want to bind the pod to run on specific node, you assign the node name to this attribute in the yaml file.
How this change fixes the issue of not providing node_name?

CC: @sidhant-agrawal @ebenahar

sidhant-agrawal: @RazTamir 
pod_data['spec']['nodeName'] is not a part of standard pod.yaml which current mostly of the tests are using from ocs_ci/templates/CSI/rbd/pod.yaml and ocs_ci/templates/CSI/cephfs/pod.yaml

RazTamir: ok thanks

RazTamir:
Please verify and let's merge this

sidhant-agrawal:
Ran the test tests/manage/test_create_storage_class_pvc.py which contains pod creation for both rbd and cephfs based pvc.

`====================================================== 2 passed in 450.85 seconds =======================================================`

PR #457, opened 2019-07-25
Fix assert for 'No Ceph tools pod found', was getting IndexError instead

jhutar:


ocs_ci/ocs/resources/pod.py
```diff
@@ -262,11 +262,11 @@ def get_ceph_tools_pod():
...
         kind=constants.POD, namespace=config.ENV_DATA['cluster_namespace']
     )
-    ct_pod = ocp_pod_obj.get(
+    ct_pod_items = ocp_pod_obj.get(
         selector='app=rook-ceph-tools'
-    )['items'][0]
-    assert ct_pod, f"No Ceph tools pod found"
-    ceph_pod = Pod(**ct_pod)
+    )['items']
+    assert len(ct_pod_items) > 0, f"No Ceph tools pod found"
```
ebenahar: You can also do:
```python 
assert ct_pod_items, "No Ceph tools pod found"
```

clacroix12:
This looks like a good change. I would take a look at @ebenahar's comment and go that route as it's a bit cleaner. Also you need to sign off on your commits before this will be merge-able. Take a look here: https://github.com/red-hat-storage/ocs-ci/pull/457/checks?check_run_id=177113936

ocs_ci/ocs/resources/pod.py
jhutar: Done, thank you!

jhutar:
@clacroix12 Should be OK now. Thank you!

PR #465, opened 2019-07-25
Allow PV deletion validation in PVC.delete()

ebenahar:

### DIFF
--- a/ocs_ci/ocs/resources/pvc.py
+++ b/ocs_ci/ocs/resources/pvc.py
@@ -87,6 +87,7 @@ class PVC(OCS):
 
         Returns:
             str: The result of PVC deletion
+
         """
         backed_pv = None
         if validate_pv_deleted:
--- a/ocs_ci/ocs/resources/pvc.py
+++ b/ocs_ci/ocs/resources/pvc.py
@@ -74,6 +74,28 @@ class PVC(OCS):
         pv_obj.reload()
         return pv_obj
 
+    def delete(self, wait=True, force=False, validate_pv_deleted=False):
+        """
+        Delete the PVC
+
+        Args:
+            wait (bool): True for wait for PVC deletion, False otherwise
+            force (bool): True for force deletion with --grace-period=0,
+                False otherwise
+            validate_pv_deleted (bool): True for respective PV deletion
+                validation, False otherwise
+
+        Returns:
+            str: The result of PVC deletion
+        """
+        backed_pv = None
+        if validate_pv_deleted:
+            backed_pv = self.backed_pv_obj
+        result = super().delete(wait=wait, force=force)
+        if validate_pv_deleted:
+            backed_pv.ocp.wait_for_delete(resource_name=self.backed_pv)
+        return result
+
     def resize_pvc(self, new_size, verify=False):
         """
         Returns the PVC size pvc_name in namespace
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -228,9 +228,10 @@ def create_pvcs(request):
         """
         Delete multiple PVCs
         """
+        validate_pv_deleted = getattr(class_instance, 'validate_pv_deleted', False)
         if hasattr(class_instance, 'pvc_objs'):
             for pvc_obj in class_instance.pvc_objs:
-                pvc_obj.delete()
+                pvc_obj.delete(validate_pv_deleted=validate_pv_deleted)
             for pvc_obj in class_instance.pvc_objs:
                 pvc_obj.ocp.wait_for_delete(pvc_obj.name)
 
--- a/tests/manage/pv_services/test_run_io_multiple_pods.py
+++ b/tests/manage/pv_services/test_run_io_multiple_pods.py
@@ -35,6 +35,7 @@ class BaseRunIOMultiplePods(ManageTest):
     """
     Run IO on multiple pods in parallel
     """
+    validate_pv_deleted = True
     num_of_pvcs = 10
     pvc_size_int = 5
     interface = None


ocs_ci/ocs/resources/pvc.py
```diff
@@ -74,6 +74,12 @@ def backed_pv_obj(self):
         pv_obj.reload()
         return pv_obj
 
+    def delete(self, wait=True, force=False, validate_pv_deleted=False):
+        result = super().delete(wait=wait, force=force)
+        if validate_pv_deleted:
+            self.backed_pv_obj.ocp.wait_for_delete(resource_name=self.backed_pv)
```
vavuthu: This validation always fails since  backed_pv_obj method will reloads the pvc instance which is already deleted in line 78.

ocs_ci/ocs/resources/pvc.py
```diff
@@ -74,6 +74,12 @@ def backed_pv_obj(self):
         pv_obj.reload()
         return pv_obj
 
+    def delete(self, wait=True, force=False, validate_pv_deleted=False):
```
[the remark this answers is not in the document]
ebenahar: done

ocs_ci/ocs/resources/pvc.py
ebenahar: Thanks, changed accordingly without verification due to PVC not getting bound bug.

ocs_ci/ocs/resources/pvc.py
```diff
@@ -74,6 +74,28 @@ def backed_pv_obj(self):
...
+
+        Returns:
+            dict: Dictionary represents a returned yaml file
+        """
+        backed_pv = None
+        if validate_pv_deleted:
+            backed_pv = self.backed_pv_obj
+        result = super().delete(wait=wait, force=force)
+        if validate_pv_deleted:
+            backed_pv.ocp.wait_for_delete(resource_name=self.backed_pv)
```
RazTamir: How this will behave when the reclaim policy is to retain?
I think we need to figure this out

ebenahar: @RazTamir yes, exactly, will have to find a way to sync between the two.
But even if we find a way to do pv deletion validation only in case the reclaim policy is delete, in some places, we'll won't to skip the validation, even if the policy is to delete. For instance in disruptive scenarios, in which we don't want to wait for PVC and PV deletion to complete

RazTamir: Ok,

Let's just ensure this will not be stuck anything because the reclaim policy is to retain and it will fail on timeout

ocs_ci/ocs/resources/pvc.py
```diff
@@ -74,6 +74,28 @@ def backed_pv_obj(self):
...
+
+        Args:
+            wait (bool): True for wait for PVC deletion, False otherwise
+            force (bool): True for force deletion with --grace-period=0,
+                False otherwise
+            validate_pv_deleted (bool): True for respective PV deletion
+                validation, False otherwise
+
+        Returns:
+            dict: Dictionary represents a returned yaml file
```
vavuthu: NIT: Leave empty line

vavuthu: Could you check the return type. looks like we are getting str

ebenahar: > NIT: Leave empty line

Where?


> Could you check the return type. looks like we are getting str

yaml.safe_load in ocs_ci.ocs.ocp.OCP#exec_oc_cmd returns a dict

> /home/ebenahar/git/ocs-ci/ocs_ci/ocs/ocp.py(80)exec_oc_cmd()
     79             from ipdb import set_trace; set_trace()
---> 80             return yaml.safe_load(out)
     81         return out

ipdb> type(yaml.safe_load(out))                                                                                                                                                                                    
<class 'dict'>

ocs_ci/ocs/resources/pvc.py
ebenahar: the default is False, we'll have to be careful when changing to True

ebenahar:
Verification:
http://pastebin.test.redhat.com/783710

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -47,5 +47,5 @@ ENV_DATA:
   # Do not change to specific version like v14.2.1-20190430 if not needed
   # cause we don't need to update it each time new 14.x version is released
   # but only once when move to new version like v15.
-  ceph_image: 'ceph/daemon-base:latest-nautilus-devel'
+  ceph_image: 'ceph/ceph:v14.2.2-20190722'
```
vasukulkarni: The commit message and changes dont match, Please add a new commit for this change and also change PR subject to reflect the same (better yet to have this commit as a different PR - not sure why it has to be part of PVC delete)

ebenahar: This change was already merged here - https://github.com/red-hat-storage/ocs-ci/pull/466
I'll rebase

ocs_ci/ocs/resources/pvc.py
vavuthu: > > NIT: Leave empty line
> 
> Where?
> 
        Returns:
            dict: Dictionary represents a returned yaml file

        """

> > Could you check the return type. looks like we are getting str
> 
> yaml.safe_load in ocs_ci.ocs.ocp.OCP#exec_oc_cmd returns a dict
> 
> > /home/ebenahar/git/ocs-ci/ocs_ci/ocs/ocp.py(80)exec_oc_cmd()
> > 79             from ipdb import set_trace; set_trace()
> > ---> 80             return yaml.safe_load(out)
> > 81         return out
> 
> ipdb> type(yaml.safe_load(out))
> <class 'dict'>

```
>>> pvc_obj = helpers.create_pvc(sc_name=sc_name, size='1Gi')
>>> pvc_obj.name
'pvc-test-2912182753'
>>> pvc_obj.status
'Bound'
>>> 
>>> 
>>> result = pvc_obj.delete()
>>> type(result)
<class 'str'>
>>> result
'persistentvolumeclaim "pvc-test-2912182753" deleted'
>>> 
```

ebenahar: Fixed

tests/fixtures.py
```diff
@@ -228,9 +228,10 @@ def finalizer():
         """
         Delete multiple PVCs
         """
+        validate_pv_deleted = getattr(class_instance, 'validate_pv_deleted', False)
         if hasattr(class_instance, 'pvc_objs'):
             for pvc_obj in class_instance.pvc_objs:
-                pvc_obj.delete()
+                pvc_obj.delete(validate_pv_deleted=validate_pv_deleted)
```
RazTamir: I don't think this is the behavior we want.
We will delete every pv and will wait for it to be deleted instead of deleing all and than wait for all (much faster)

ebenahar:
Closing this PR. PV deletion validation will be done outside PVC.delete() and is being implemented in https://github.com/red-hat-storage/ocs-ci/pull/476

PR #471, opened 2019-07-26
Add E2e baseline I/O test.

wusui:
Signed-off-by: Warren Usui <wusui@redhat.com>

### DIFF
--- a/tests/e2e/test_minimal_io_and_set_baseline.py
+++ b/tests/e2e/test_minimal_io_and_set_baseline.py
@@ -1,0 +1,62 @@
+"""
+OCS-685
+
+Demonstrate that the ceph configuration can handle a minimal level of I/O
+and establish a baseline for comparision of later I/O tests.
+"""
+import pytest
+import logging
+import json
+
+from ocs_ci.framework.testlib import ManageTest, tier1
+from tests.fixtures import (
+    create_rbd_storageclass, create_rbd_pod, create_pvc, create_ceph_block_pool,
+    create_rbd_secret, create_project
+)
+
+logger = logging.getLogger(__name__)
+
+
+def do_io_actions(io_type, io_info):
+    """
+    Display fio statistics and stash fio results somewhere.
+
+    Args:
+        io_type (str):  either "Read" or "Write"
+        io_info (dict): results of fio run.
+    """
+    logging.info(
+        f'{io_type}: {io_info.get("iops")} IOPS '
+    )
+    logging.info(
+        f'{io_type}: {io_info.get("clat")} Completion Latency '
+    )
+    with open(f'/tmp/{io_type}_fio_data_baseline.json', 'w') as out_json:
+        json.dump(io_info, out_json)
+
+
+@pytest.mark.usefixtures(
+    create_rbd_secret.__name__,
+    create_ceph_block_pool.__name__,
+    create_rbd_storageclass.__name__,
+    create_project.__name__,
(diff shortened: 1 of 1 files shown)


tests/e2e/test_minimal_io_and_set_baseline.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+    logging.info(
+        f'{io_type}: {io_info.get("clat")} Completion Latency '
+    )
+    with open(f'/tmp/{io_type}_fio_data_baseline.json', 'w') as out_json:
+        json.dump(io_info, out_json)
+
+
+@pytest.mark.usefixtures(
+    create_rbd_secret.__name__,
```
vasukulkarni: This method of fixture is no more valid.

tests/e2e/test_minimal_io_and_set_baseline.py
```diff
@@ -0,0 +1,60 @@
...
+logger = logging.getLogger(__name__)
+
+
+def do_io_actions(io_type, io_info):
+    """
+    Display fio statistics and stash fio results somewhere.
+
+    Args:
+        io_type (str):  either "Read" or "Write"
+        io_info (dict): results of fio run.
```
vavuthu: NIT: Leave empty line
eg:

```
    Args:
        io_type (str):  either "Read" or "Write"
        io_info (dict): results of fio run.

    """
```

vavuthu: io_info (dict) ---> this is the not the complete results of fio run. In the test case, from the test case, we are passing subset of fio results ( like read , write ). Please change the description accordingly.

vavuthu:
@wusui Do we have any threshold values for Read and Write ? In which scenarios this test case will fail.

tests/e2e/test_minimal_io_and_set_baseline.py
wusui: How should this be done now that this method is invalid?

tests/e2e/test_minimal_io_and_set_baseline.py
wusui: Okay.

wusui: Okay.

wusui:
Vavuthu -- I have sent email to Rick Sussman to get minimum values to check here.  I will add code that will check against these values.

tests/e2e/test_minimal_io_and_set_baseline.py
```diff
@@ -0,0 +1,60 @@
...
+    create_rbd_storageclass.__name__,
+    create_project.__name__,
+    create_pvc.__name__,
+    create_rbd_pod.__name__,
+)
+@pytest.mark.polarion_id("OCS-685")
+class TestMinimalLimit(ManageTest):
+
+    @tier1
+    def test_run_io(self):
```
keesturam: This test covers only IO run on rbd volume. please change the function name accordingly. 
We'll also have to cover cephfs and raw block io.

tests/e2e/test_minimal_io_and_set_baseline.py
wusui: Has the new method of doing this been implemented yet?

RazTamir:
@wusui see my previous comment - do we still need this?

vasukulkarni:
Closing this one, Please reopen if needed.

PR #476, opened 2019-07-29
Fix PVC deletion and PV deletion

RazTamir:

### DIFF
--- a/tests/e2e/test_running_cluster_monitoring_with_persistent_storage.py
+++ b/tests/e2e/test_running_cluster_monitoring_with_persistent_storage.py
@@ -4,7 +4,6 @@ import pytest
 
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.framework.testlib import tier1, E2ETest
-from ocs_ci.ocs.resources.pvc import delete_pvcs
 from ocs_ci.ocs.monitoring import (
     create_configmap_cluster_monitoring_pod,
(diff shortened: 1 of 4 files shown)


tests/e2e/test_running_cluster_monitoring_with_persistent_storage.py
```diff
@@ -26,20 +26,18 @@ def test_fixture(request):
     """
     Setup and teardown
     """
-    self = request.node.cls
-
     def finalizer():
-        teardown(self)
+        teardown()
```
clacroix12: Since `teardown` doesn't take any arguments we can add it as a finalizer directly.
```
request.addfinalizer(teardown)
```

tests/helpers.py
```diff
@@ -609,7 +609,28 @@ def get_all_pvs():
...
+        pvc_objs (list): List of the pvc objects to be deleted
+        wait (bool): True if need to wait for PVCs deletion
+
+    Returns:
+        bool: True if deletion is successful
+    """
+    status = pvc.delete_pvcs(pvc_objs)
+    if wait:
+        for pvc_obj in pvc_objs:
+            if pvc_obj.backed_pv_obj().get('spec').get(
```
vavuthu: This validation always fail since backed_pv_obj method will reloads the pvc instance which is already deleted in line 623

RazTamir:
Closing this PR for now

Issue #409, opened 2019-07-16
add property to xunit with: "polarion-testrun-id" 


petr-balogh:
We need to update also this property to xunit result:
<property name="polarion-testrun-id" value="OCS_4_2_tier1"/>

We need to add version + tier number for now which we can read dynamically from pytest.

We can compose the run id from all passed markers.

Example how to retrieve markers is [here](https://stackoverflow.com/questions/52585701/how-to-retrieve-the-markers-for-py-test-in-conftest-py)

PR #477, opened 2019-07-29
update reporting (add Test Run ID to reports)

dahorak:
* add Test Run ID to html and email reports
  (for example:  OCS_US_4-2_tier1)
* add folowing properties to xml jUnit report:
  * polarion-testrun-id
  * polarion-custom-plannedin
  * polarion-testrun-status-id

fixes: https://github.com/red-hat-storage/ocs-ci/issues/409

### DIFF
--- a/conf/ocsci/downstream_config.yaml
+++ b/conf/ocsci/downstream_config.yaml
@@ -12,5 +12,5 @@ ENV_DATA:
   # rook_csi_attacher_image:
   cluster_namespace: 'rook-ceph'
 REPORTING:
-  # Upstream: 'US' or Downstream: 'DS'
+  # Upstream: 'US' or Downstream: 'DS' - used only for reporting (Test Run ID)
   us_ds: 'DS'
--- a/conf/ocsci/downstream_config.yaml
+++ b/conf/ocsci/downstream_config.yaml
@@ -11,3 +11,6 @@ ENV_DATA:
   # rook_csi_snapshotter_image:
   # rook_csi_attacher_image:
   cluster_namespace: 'rook-ceph'
+REPORTING:
+  # Upstream: 'US' or Downstream: 'DS'
+  us_ds: 'DS'
--- a/conf/ocsci/downstream_config.yaml
+++ b/conf/ocsci/downstream_config.yaml
@@ -12,5 +12,5 @@ ENV_DATA:
   # rook_csi_attacher_image:
   cluster_namespace: 'rook-ceph'
 REPORTING:
-  # Upstream: 'US' or Downstream: 'DS' - used only for reporting (Test Run ID)
+  # Upstream: 'US' or Downstream: 'DS', used only for reporting (Test Run Name)
   us_ds: 'DS'
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -36,6 +36,9 @@ REPORTING:
     address: "ocs-ci@redhat.com"
   polarion:
     project_id: "OpenShiftContainerStorage"
+  ocs_version: '4.2'
+  # Upstream: 'US' or Downstream: 'DS'
+  us_ds: 'US'
 
 # This is the default information about environment. Will be overwritten with
 # --cluster-conf file.yaml data you will pass to the pytest.
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -36,8 +36,7 @@ REPORTING:
     address: "ocs-ci@redhat.com"
   polarion:
     project_id: "OpenShiftContainerStorage"
-  ocs_version: '4.2'
-  # Upstream: 'US' or Downstream: 'DS'
+  # Upstream: 'US' or Downstream: 'DS' - used only for reporting (Test Run ID)
   us_ds: 'US'
 
 # This is the default information about environment. Will be overwritten with
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -36,7 +36,7 @@ REPORTING:
     address: "ocs-ci@redhat.com"
   polarion:
     project_id: "OpenShiftContainerStorage"
-  # Upstream: 'US' or Downstream: 'DS' - used only for reporting (Test Run ID)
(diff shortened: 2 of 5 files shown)


ocs_ci/framework/conf/default_config.yaml
```diff
@@ -36,6 +36,9 @@ REPORTING:
     address: "ocs-ci@redhat.com"
   polarion:
     project_id: "OpenShiftContainerStorage"
+  ocs_version: '4.2'
```
mbukatov: I would rather not specify such values in config files. This could backfire because:
- it has no actual meaning (OCS version is specified by various image versions), when this value is updated, it has no effect 
- it's creates information duplicity, which is prone to various human errors

Moreover when we have ocs version operator, it will have full control over this and the only canonical way to get version would be to ask it in the same way, as we can ask cluster version operator now.

conf/ocsci/downstream_config.yaml
```diff
@@ -11,3 +11,6 @@ ENV_DATA:
   # rook_csi_snapshotter_image:
   # rook_csi_attacher_image:
   cluster_namespace: 'rook-ceph'
+REPORTING:
+  # Upstream: 'US' or Downstream: 'DS'
+  us_ds: 'DS'
```
mbukatov: I'm not sure this is a good idea. Maybe you can add into the comment that this is just for test run id format used in jenkins, and it should not be touched or used in any other way.

ocs_ci/framework/conf/default_config.yaml
mbukatov: Maybe you can rename it to "test run prefix" or something like that, and drop it from the default config file.

conf/ocsci/downstream_config.yaml
dahorak: It is in `REPORTING` section, which might be slightly indicating, that it is only for reporting... but I'll add it to the comment too.

ocs_ci/framework/conf/default_config.yaml
dahorak: Fixed by 0b2ca9ff.

ocs_ci/utility/utils.py
```diff
@@ -954,6 +954,34 @@ def ocsci_log_path():
     )
 
 
+def get_testrun_id():
```
clacroix12: Should we rename this to `get_polarion_testrun_id` as to not confuse the return value from this with `config['RUN']['run_id']`?

dahorak: I was thinking about the `polarion` prefix, but it is used also for email and html reports... what do you thing about `get_testrun_name` instead?

clacroix12: I think `get_testrun_name` makes sense

dahorak: Fixed by 065f186.

ocs_ci/utility/utils.py
```diff
@@ -954,6 +954,34 @@ def ocsci_log_path():
...
+    """
+    Prepare testrun ID for Polarion (and other reports).
+
+    Return config.REPORTING["polarion"]["testrun_id"], if configured.
+    Otherwise prepare testrun ID based on Upstream/Downstream information,
+    OCS version and used markers.
+
+    Returns:
+        str: String containing testrun ID
+    """
```
vavuthu: Leave empty line between Returns and """
eg: 
```
Returns:
        str: String containing testrun ID

"""
```

dahorak:
Verified for `tier1` and `tier3` tests:
* Email report subject: *ocs-ci results for OCS_US_4-2_tier3 (RUN ID: 1564556323)*
* Line in Environment table: *Test Run ID: OCS_US_4-2_tier3*[1]
* jUnit XML Report:
```
<?xml version="1.0" encoding="utf-8"?>
<testsuite errors="0" failures="0" name="jenkins-us-tier3" skipped="0" tests="5" time="360.964">
  <properties>
    <property name="polarion-project-id" value="OpenShiftContainerStorage"/>
    <property name="polarion-custom-description" value="https://...jenkins.../job/qe-deploy-ocs-cluster/258/"/>
    <property name="polarion-testrun-id" value="OCS_US_4-2_tier3"/>
    <property name="polarion-testrun-status-id" value="inprogress"/>
  </properties>
...
```

[1] *Test Run ID* is renamed to *Test Run Name* in commit bc1b3fc.

ocs_ci/utility/utils.py
```diff
@@ -954,6 +954,34 @@ def ocsci_log_path():
...
+    markers = config.RUN['cli_params'].get('-m', '').replace(" ", "-")
+    us_ds = config.REPORTING.get("us_ds")
+    if config.REPORTING["polarion"].get("testrun_id"):
```
vavuthu: is testrun_id is same as run_id ?If not where do we updating testrun_id to REPORTING section

ocs_ci/utility/utils.py
```diff
@@ -954,6 +954,34 @@ def ocsci_log_path():
...
+        testrun_id = f"OCS_{us_ds}_{config.REPORTING.get('testrun_id_part', '')}_{markers}"
+    else:
+        testrun_id = f"OCS_{us_ds}_{config.REPORTING.get('testrun_id_part', '')}"
+    # replace invalid character(s) by '-'
+    testrun_id = testrun_id.translate(
+        str.maketrans(
+            {key: '-' for key in ''' \\/.:*"<>|~!@#$?%^&'*(){}+`,=\t'''}
+        )
+    )
+    return testrun_id
```
vavuthu: is testrun_id is unique for every run

ocs_ci/utility/utils.py
dahorak: Fixed in b123b44.

ocs_ci/utility/utils.py
dahorak: I've renamed the `testrun_id` to `testrun_name` in b123b44, to be more consistent with the `get_testrun_name`.

`testrun_name` (previously `testrun_id`) is optional, if not provided, the *Test Run Name* is generated from other sources (see the following lines).

ocs_ci/utility/utils.py
dahorak: Now, `testrun_id` (now renamed as `testrun_name`) should be the same for the same configuration.

PR #484, opened 2019-07-29
patch gp2 ebs based storage class 

vasukulkarni:
and use cephfs/rbd as default storageclass.

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -278,6 +278,14 @@ class Deployment(object):
         assert ceph_health_check(
             namespace=config.ENV_DATA['cluster_namespace']
         )
+        # patch gp2 (EBS) storage class as 'non-default'
+        logger.info("Patch gp2 storageclass as non-default")
+        patch = " '{\"metadata\": {\"annotations\":{\"storageclass.kubernetes.io/is-default-class\":\"false\"}}}' "
+        run_cmd(
+            f"oc patch storageclass gp2 "
+            f"-p {patch} "
+            f"--request-timeout=120s"
+        )
 
     def destroy_cluster(self):
         """


tests/conftest.py
```diff
@@ -325,6 +325,13 @@ def cluster_teardown_finalizer():
     # Verify health of ceph cluster
     # TODO: move destroy cluster logic to new CLI usage pattern?
     log.info("Done creating rook resources, waiting for HEALTH_OK")
+    # patch gp2 (EBS) storage class as 'non-default'
+    patch = " '{\"metadata\": {\"annotations\":{\"storageclass.kubernetes.io/is-default-class\":\"false\"}}}' "
+    run_cmd(
+        f"oc patch storageclass gp2 "
+        f"-p {patch} "
+        f"--timeout=120s"
```
vavuthu: looks like --timeout option is not available with oc patch. instead we can use --request-timeout

```
(venv1) [vavuthu@localhost ocs-ci]$ oc patch storageclass gp2 -p  '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}' --timeout=120s
Error: unknown flag: --timeout

venv1) [vavuthu@localhost ocs-ci]$ oc patch storageclass gp2 -p  '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}' --request-timeout="120s"
storageclass.storage.k8s.io/gp2 patched
(venv1) [vavuthu@localhost ocs-ci]$ 

```

vasukulkarni: Thanks fixing it!

vasukulkarni:
```
09:14:52 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc patch storageclass gp2 -p  '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'  --request-timeout=120s

$ oc get storageclass
NAME                        PROVISIONER             AGE
gp2                         kubernetes.io/aws-ebs   17h
rook-ceph-block (default)   ceph.rook.io/block      17h
```

tests/conftest.py
```diff
@@ -325,6 +325,13 @@ def cluster_teardown_finalizer():
     # Verify health of ceph cluster
     # TODO: move destroy cluster logic to new CLI usage pattern?
     log.info("Done creating rook resources, waiting for HEALTH_OK")
```
clacroix12: I would just move this log statement after your patch code so logging the workflow matches what logging is saying.

vasukulkarni: I can add a new logging statement before patch, the HEALTH_OK logging looks accurate to me since we have reached before this point int time?

clacroix12: Basically what I want to avoid is if your block of code were to raise an error for any reason, the log would say we were waiting for HEALTH_OK which isn't technically true since we never got to the health check. You can add a log statement in before your patch, that's totally fine. I just wanted to move the one on line 327 after your patch so we don't log that we are waiting for HEALTH_OK before we actually are.

vasukulkarni: Sorry I missed the *waiting* part in the log(was thinking HEALTH_OK has already reached), makes sense to me! , will fix it.

vavuthu:
LGTM.

Only thing is I feel ceph_health_check should be the last step in deployment but its not a blocker 
for this to merge. WDYT

vasukulkarni:
> LGTM.
> 
> Only thing is I feel ceph_health_check should be the last step in deployment but its not a blocker
> for this to merge. WDYT

@vavuthu I thought about this one,  HEALTH_OK indicates that storage is perfect and we can go ahead and make gp2 non-default :)

RazTamir:
Hi @vasukulkarni @vavuthu @clacroix12 , this hack is not good.
We are fixing something in our local ocs-ci repository that might not reflect in the actual product hence we are testing something different than what the customer will get.
I saw bug https://bugzilla.redhat.com/show_bug.cgi?id=1734443 but wonder if we have a revert for this as soon as the bug will be fixed

vasukulkarni:
@RazTamir its not a hack, As filed in bz, we don't want gp2 - customers should be using ceph provisioned storage instead of storage backed by ebs. when we use GUI to create new projects the default storageclass is used(which is gp2). I am also going to file an issue here to track bz changes here.

PR #487, opened 2019-07-30
update docs on how to address new review comments on old PR's

vasukulkarni:

### DIFF
--- a/docs/code_review.md
+++ b/docs/code_review.md
@@ -27,8 +27,8 @@
 * All comments should be addressed or responded and once comment is considered
     as closed we should Resolve conversation to close the comment.
 * If there are new review comments for a PR that is older than 4 days, the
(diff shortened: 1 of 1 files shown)


docs/code_review.md
```diff
@@ -26,6 +26,9 @@
     approved  from github by at least 2 project maintainers before merge.
 * All comments should be addressed or responded and once comment is considered
     as closed we should Resolve conversation to close the comment.
+* If there are new review comments for a PR that is older than 4 days, the
```
clacroix12: Is age the correct metric to track for this or do we care more about how long it has been since the last update?

vasukulkarni: I think the age is correct, once we get couple review comments in the first few days and has been fixed, the original age is what matters and any new comments that appear from new folks should be just addressed as follow on PRs.

PR #494, opened 2019-07-31
WIP: Update deployment to use a 4.2 nightly build

clacroix12:
- Updated the OCP client/installer versions to a 4.2 nightly build
- Added .json extension to the terraform.tfvars file
- Changed location of where we pull deployment artifacts from

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -19,14 +19,14 @@ RUN:
   run_id: null  # this will be redefined in the execution
   kubeconfig_location: 'auth/kubeconfig' # relative from cluster_dir
   cli_params: {}  # this will be filled with CLI parameters data
-  client_version: '4.1.4'
+  client_version: '4.2.0-0.nightly-2019-07-30-073644'
   bin_dir: './bin'
   google_api_secret: '~/.ocs-ci/google_api_secret.json'
 
 # In this section we are storing all deployment related configuration but not
 # the environment related data as those are defined in ENV_DATA section.
 DEPLOYMENT:
-  installer_version: "4.1.4"
+  installer_version: "4.2.0-0.nightly-2019-07-30-073644"
   force_download_installer: True
   force_download_client: True
 
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -56,6 +56,10 @@ and place in the `data` directory at the root level of the project.
 If there is no `data` directory, create one.
 The name of the file should be `pull-secret`.
 
+In addition you will need to add a registry auth to your pull-secret to
+support deploying CI / Nightly builds. Please follow the instructions
+[here](https://mojo.redhat.com/docs/DOC-1204026) to do so.
+
 ## Tests
 
 ### AWS and CentralCI Authentication files
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -67,7 +67,8 @@ class AWSBase(Deployment):
         Args:
             size (int): Size of volume in GB (default: 100)
         """
-        with open(os.path.join(self.cluster_path, "terraform.tfvars")) as f:
+        tfvars_file = "terraform.tfvars.json"
+        with open(os.path.join(self.cluster_path, tfvars_file)) as f:
             tfvars = json.load(f)
 
         cluster_id = tfvars['cluster_id']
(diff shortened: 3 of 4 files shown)


clacroix12:
There are very few code changes to enable this, however the biggest issue with this is the requirement for anyone executing ocs-ci (including jenkins) to update their pull-secret with an additional registry auth. I have outlined the process to do so [here](https://mojo.redhat.com/docs/DOC-1204026).

@RazTamir and I discussed giving the team some time to go through this process before merging these changes so I am setting this as `WIP` to block merging for the time being. We figured a week would be a sufficient amount of time so please do read through the document and update your pull-secrets. Feel free to ping me with any issues.

clacroix12:
> Can we at least put the link to the mojo page somewhere in the documentation in this project?

Yes this is a good idea, I can update our section on the pull secret to include this doc.

> But as mojo is internal it will be accessible just for ppl from RH. I am not on VPN so I didn't check the content of the mojo yet, just asking if it contains some confidential info which cannot be shared here?

The requirement is to be a part of the OpenShift organization in order to retrieve the registry auth for the pull-secret. Part of the process includes linking your redhat rover profile with your github ID so I _assumed_ it was redhat internal only.

dahorak:
Because of the current outage of our Jenkins, I wasn't able to test it there, but locally the installation passed. Also I've update pull-secret Secret File in Jenkins, so it *should* be possible to try it there, once the outage will be resolved.

Another task will be automatic rotation of the registry token.

dahorak:
Verified also in by following Jenkins job build on tier1 tests: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/279/

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -19,14 +19,14 @@ RUN:
   run_id: null  # this will be redefined in the execution
   kubeconfig_location: 'auth/kubeconfig' # relative from cluster_dir
   cli_params: {}  # this will be filled with CLI parameters data
-  client_version: '4.1.4'
+  client_version: '4.2.0-0.nightly-2019-07-30-073644'
```
mbukatov: Any reason to have this hardcoded like this?

Also what is the reason for the default to be a nightly build and not CI build? What would we suggest the team to prefer?

clacroix12: > Any reason to have this hardcoded like this?

No reason other than this is the current way we are handling OCP client and installer versions. Note this is the default config that can be overwritten by passing a different value through with `--ocsci-conf` so they aren't "hardcoded" as much as they are default values.

> Also what is the reason for the default to be a nightly build and not CI build? What would we suggest the team to prefer?

This was the latest version I locally tested with, so I stuck with it. The point is to ensure we are testing 4.2. I have no _personal_ preference between CI and Nightly.

mbukatov: Ok, so this looks as a safe default and we expect people to check which version to use?

clacroix12: > we expect people to check which version to use?

I'm not sure what you mean by this. If users need to test using a specific version of OCP client or installer then yeah they should be checking what version we are using by default and override if necessary.

vasukulkarni: I see what Martin is saying,  Is there a -latest-nightly tag that one can use?

prsurve: @vasukulkarni we can get latest 4.2 nightly build from https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/4.2.0-0.nightly/latest

prsurve: from above link we can get download link of latest binary and we can download and deploy cluster i just made a small script to get downloadlink and download oc-client and oc-installer 
https://gist.github.com/prsurve/dda470d8ad42e17af8f1a026ce9d63a7

clacroix12: If we want to default to using the latest nightly then let's create an issue for that and I can either update this PR to handle it or do it in a subsequent PR. Defaulting to the latest CI or nightly release wasn't a requirement so much as simply moving to 4.2 was so I didn't go that route.

RazTamir: IMO there's no reason for consuming latest.
We are not doing QA for OCP and should not waste time on testing the failures landed there. 
Let's stick with a stable version and once in a while, move on to a new one

petr-balogh:
Hi @RazTamir , I see you merged this with WIP in the name of PR which should be removed from the name of PR before merge otherwise I think you had to encounter and did the force merge and some warning which basically block ppl to not to merge the PR when it contains WIP in the name.

RazTamir:
I wasn't blocked :)
I just clicked 'rebase and merge'

PR #497, opened 2019-08-01
Test cases for ec2 instances restart (force True/False)

ebenahar:


ebenahar:
=========================================================================================== 1 passed in 331.61 seconds ============================================================================================

Full logs:
[logs_ungraceful_shutdown.zip](https://github.com/red-hat-storage/ocs-ci/files/3458898/logs_ungraceful_shutdown.zip)

ocs_ci/ocs/ocp.py
```diff
@@ -373,3 +373,41 @@ def switch_to_default_rook_cluster_project():
...
+        list: Cluster node dictionaries
+
+    """
+    nodes_obj = OCP(kind='node')
+    return nodes_obj.get()['items']
+
+
+def wait_for_nodes_ready(num_of_nodes):
+    """
+    Wait until all nodes are in Ready status
```
vasukulkarni: boto has waiters - It is efficient to use that call instead of sampler

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Waiter.InstanceRunning

ocs_ci/utility/aws.py
```diff
@@ -262,6 +257,26 @@ def detach_and_delete_volume(self, volume, timeout=120):
...
+            instance_id (str): ID of the instance to stop
+            status (int): The status code to wait for
+            timeout (int): The number in seconds to wait for the instance
+                reach the status
+
+        Returns:
+            bool: True in case the instance reached the status, False otherwise
+
+        """
+        for sample in utils.TimeoutSampler(
```
vasukulkarni: Lets use boto waiters instead of function sampler - https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Waiter.InstanceRunning

ocs_ci/utility/aws.py
```diff
@@ -301,11 +313,76 @@ def start_ec2_instance(self, instance_id, wait=False):
...
+    instance_ids, instance_names = get_instances_ids_and_names(instances)
+
+    for instance_id, instance_name in zip(instance_ids, instance_names):
+        if aws.get_instances_status_by_id(instance_id) == constants.INSTANCE_STOPPED:
+            logger.info(f"Starting instance {instance_name}")
+            aws.start_ec2_instance(instance_id)
+
+    for instance_id, instance_name in zip(instance_ids, instance_names):
+        logger.info(f"Waiting for instance {instance_name} to reach status running")
+        aws.wait_for_instance_status(
```
vasukulkarni: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Instance.wait_until_running

ocs_ci/ocs/ocp.py
ebenahar: @vasukulkarni but here I'm checking the nodes from OCP side with 'oc get node -o yaml' to make sure all nodes reached Ready

vasukulkarni: ok thanks, I see it,  OCP also has wait for function which could be used but either way looks fine.

ocs_ci/utility/aws.py
ebenahar: @vasukulkarni Nice!
Also found these 2:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Instance.wait_until_running
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Instance.wait_until_stopped

ocs_ci/utility/aws.py
```diff
@@ -262,6 +257,26 @@ def detach_and_delete_volume(self, volume, timeout=120):
             delete_response
         )
 
+    def wait_for_instance_status(self, instance_id, status, timeout=TIMEOUT):
+        """
+
```
clacroix12: Missing description

ocs_ci/utility/aws.py
vasukulkarni: 👍

ocs_ci/utility/aws.py
ebenahar: method deleted

tests/manage/cluster/cluster_shutdown/test_shutdown.py
```diff
@@ -0,0 +1,105 @@
...
+
+    request.addfinalizer(finalizer)
+
+    instances = ocp.get_all_nodes()
+    return instances
+
+
+@pytest.fixture()
+def vms():
+    # TODO
```
vasukulkarni: you can bring in the TODO's in next PR other than that it all LGTM.

tests/manage/cluster/cluster_shutdown/test_shutdown.py
```diff
@@ -0,0 +1,105 @@
...
+        assert ocp.wait_for_nodes_ready(len(instances)), (
+            "Not all nodes reached status Ready"
+        )
+
+        ceph_cluster = CephCluster()
+        assert ceph_health_check(
+            namespace=config.ENV_DATA['cluster_namespace']
+        )
+        ceph_cluster.cluster_health_check(timeout=60)
+
```
nehaberry: @ebenahar as part of validate cluster, will it be good to have a small check for the folllowing:

1. able to create 1 PVC + POD for both RBD and Cephfs ( confiming that cephcluster is usable even after full cluster shutdown)

Please share your opinion.

tests/manage/cluster/cluster_shutdown/test_shutdown.py
```diff
@@ -0,0 +1,105 @@
...
+    @pytest.mark.skipif(
+        condition=config.ENV_DATA['platform'] != 'AWS',
+        reason="Tests are not running on AWS deployed cluster"
+    )
+    def test_ungraceful_shutdown_aws(self, instances):
+        """
+        Test ungraceful cluster shutdown - AWS
+        """
+
+        aws.stop_instances(instances)
```
hnallurv: This test case needs 'Abrupt'(=ungraceful) shutdown. Are we sure by 'stopping' the instance that we are shutting down a node 'abruptly'? Can we please confirm this?...

Isn't 'stop instance' trying to bring down all the services running in that instance in an ordered way - which is 'graceful' shutdown?

tests/manage/cluster/cluster_shutdown/test_shutdown.py
hnallurv: Usually the Tier 4 tests need some background operations to keep running throughout the test. As part of that, the above mentioned operation should be covered. Not sure cluster_health_check will sample the results of such background operations. Or if the background operations itself confirm the success of pvc, pod and IOs success, then no need to explicitly check that.

tests/manage/cluster/cluster_shutdown/test_shutdown.py
hnallurv: I think we need to use '--force' operation to simulate the abrupt failure. Please check: https://docs.aws.amazon.com/cli/latest/reference/ec2/stop-instances.html

tests/manage/cluster/cluster_shutdown/test_shutdown.py
ebenahar: @nehaberry yes, I'll add these checks after the cluster health check.
@hnallurv I think for that, we can execute this test with --io_in_bg that will run IO in the background. Although, I'm pretty sure we'll see ugly exceptions from the IO thread due to the shutdown. Will try that later

tests/manage/cluster/cluster_shutdown/test_shutdown.py
ebenahar: @hnallurv We already do this:
https://github.com/red-hat-storage/ocs-ci/blob/658b851ed503e274bb5ddee020979dd5931cae63/ocs_ci/utility/aws.py#L278

tests/manage/cluster/cluster_shutdown/test_shutdown.py
ebenahar: done

ebenahar:
The test passed. However, leftover check is failing. Seems like upon cluster restart, new pods with new names are spawned. @RazTamir need a way to ignore that
[logs_shutdown.zip](https://github.com/red-hat-storage/ocs-ci/files/3461968/logs_shutdown.zip)


------------------------------------------------------------------------ generated html file: file:///home/ebenahar/git/ocs-ci/report.html ------------------------------------------------------------------------
======================================================================================= 1 passed, 1 error in 814.96 seconds =======================================================================================

ebenahar:
Added ignore_leftovers decorator

ocs_ci/utility/aws.py
```diff
@@ -68,6 +62,18 @@ def ec2_resource(self):
             )
         return self._ec2_resource
 
+    def get_ec2_instance(self, instance_id):
+        """
```
clacroix12: This method needs a description as well

ebenahar:
=========================================================================================== 1 passed in 711.84 seconds ============================================================================================
[shutdown_logs.zip](https://github.com/red-hat-storage/ocs-ci/files/3462329/shutdown_logs.zip)

ocs_ci/utility/aws.py
ebenahar: done

ocs_ci/utility/aws.py
```diff
@@ -301,11 +304,73 @@ def start_ec2_instance(self, instance_id, wait=False):
...
+        tuple: lists of node IDs and names
+
+    """
+    instance_names = [node.get('metadata').get('name') for node in instances]
+    instance_ids = [
+        'i-' + node.get('spec').get(
+            'providerID'
+        ).partition('i-')[-1] for node in instances
+    ]
+    return instance_ids, instance_names
```
clacroix12: Would it be better to just return a dict here like `{instance_id: instance_name}` instead of having to zip it up after the fact or is there a reason to keep them in separate lists?

ocs_ci/utility/aws.py
```diff
@@ -68,6 +62,19 @@ def ec2_resource(self):
...
 
+    def get_ec2_instance(self, instance_id):
+        """
+        Get instance of ec2 Instance
+
+        Args:
+            instance_id (str): The ID of the instance to get
+
+        Returns:
+            boto3.Instance instance of ec2 instance resource
```
clacroix12:
```suggestion
            boto3.Instance: instance of ec2 instance resource
```

tests/helpers.py
```diff
@@ -149,6 +149,20 @@ def create_pod(
     return pod_obj
 
 
+def create_project():
```
vasukulkarni: We have this in OCP class, looks like you added recently

ebenahar: I think it's more convenient to have this function in helpers

ocs_ci/utility/aws.py
ebenahar: done

ocs_ci/utility/aws.py
ebenahar: right, thanks. converted to dict

ebenahar:
=========================================================================================== 1 passed in 696.49 seconds ============================================================================================
[shutdown_logs.zip](https://github.com/red-hat-storage/ocs-ci/files/3463686/shutdown_logs.zip)

ebenahar:
=========================================================================================== 2 passed in 1398.37 seconds ===========================================================================================
[shutdown_logs.zip](https://github.com/red-hat-storage/ocs-ci/files/3464212/shutdown_logs.zip)

ocs_ci/utility/aws.py
```diff
@@ -262,50 +269,75 @@ def detach_and_delete_volume(self, volume, timeout=120):
...
+            force (bool): True for force instance stop, False otherwise
 
-        Returns:
-            bool: True in case operation succeeded, False otherwise
         """
-        res = self.ec2_client.stop_instances(
-            InstanceIds=[instance_id], Force=True
-        )
+        instance_ids = [key for key in instances.keys()]
+        instance_names = [val for val in instances.values()]
```
MeridianExplorer: Would replace 283 and 284 with a more elegant solution - 
```suggestion
        instance_ids, instance_names = zip(*instances.items())
```

ocs_ci/utility/aws.py
```diff
@@ -262,50 +269,75 @@ def detach_and_delete_volume(self, volume, timeout=120):
...
+        logger.info(f"Stopping instances {instance_names} with Force={force}")
+        assert self.ec2_client.stop_instances(
```
MeridianExplorer: stop_instances() always returns a list containing information about the instance(s) status(es), and thus will always evaluate to True.
I think it'll be best to parse the returning dictionary, and print+analyze the results accordingly.
More here:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.stop_instances

ocs_ci/utility/aws.py
```diff
@@ -262,50 +269,75 @@ def detach_and_delete_volume(self, volume, timeout=120):
...
-        state = res.get('StartingInstances')[0].get('CurrentState').get('Code')
-        return state == PENDING
+        instances = {
+            key: val for key, val in instances.items() if (
+                self.get_instances_status_by_id(key) == constants.INSTANCE_STOPPED
+            )
+        }
+        if instances:
+            instance_ids = [key for key in instances.keys()]
+            instance_names = [val for val in instances.values()]
```
MeridianExplorer: Can be zipped like the ones in stop_ec2_instances

ocs_ci/utility/aws.py
```diff
@@ -262,50 +269,75 @@ def detach_and_delete_volume(self, volume, timeout=120):
...
+        instances (list): Nodes dictionaries, returned by 'oc get node -o yaml'
+
+    Returns:
+        dict: The ID keys and the name values of the instances
+
+    """
+    return {
+        'i-' + instance.get('spec').get('providerID').partition('i-')[
+            -1
+        ]: instance.get('metadata').get('name') for instance in instances
```
MeridianExplorer: The wrapping here makes this highly unreadable

tests/manage/cluster/nodes/test_nodes_restart.py
```diff
@@ -0,0 +1,200 @@
...
+    Returns:
+        tuple: empty lists of resources
+
+    """
+    projects = list()
+    secrets = list()
+    pools = list()
+    storageclasses = list()
+    pvcs = list()
+    pods = list()
```
MeridianExplorer:
```suggestion
    projects, secrets, pools, storageclasses, pvcs, pods = [None] for i in range(6)
```

tests/manage/cluster/nodes/test_nodes_restart.py
```diff
@@ -0,0 +1,200 @@
...
+                pod.delete()
+                pod.ocp.wait_for_delete(pod.name)
+        if pvcs:
+            for pvc in pvcs:
+                pvc.delete()
+                pvc.ocp.wait_for_delete(pvc.name)
+        if storageclasses:
+            for sc in storageclasses:
+                sc.delete()
+                sc.ocp.wait_for_delete(sc.name)
```
MeridianExplorer: for resource in zip(pods, pvcs, storageclasses, secrets):
  resources.delete()
  resource.ocp.wait_for_delete(resource.name)

tests/manage/cluster/nodes/test_nodes_restart.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+        ceph_cluster.cluster_health_check(timeout=60)
+
+        # Create resources and run IO for both FS and RBD
+        projects = resources[0]
+        secrets = resources[1]
+        pools = resources[2]
+        storageclasses = resources[3]
+        pvcs = resources[4]
+        pods = resources[5]
```
MeridianExplorer:
```suggestion
        projects, secrets, pools, storageclasses, pvcs, pods = resources[:6]
```

ocs_ci/utility/aws.py
ebenahar: done

tests/manage/cluster/nodes/test_nodes_restart.py
ebenahar: done

tests/manage/cluster/nodes/test_nodes_restart.py
ebenahar: we can't do what you've suggested (even though it's very cool) because pools[1] is an existing FS pool that we just get its name and don't want to delete

tests/manage/cluster/nodes/test_nodes_restart.py
ebenahar: done

tests/manage/cluster/nodes/test_nodes_restart.py
MeridianExplorer: I'm not sure I understood, but please do note that the pools aren't included in this - I only minified pods, pvcs, storageclasses and secrets :)

I noticed pools and projects have different deletion procedures and I suggest to leave them as they are (and unify all the ones who use the same methodology)

ebenahar: now I see that :)

Regarding zip usage - it iterates over the tuple and doesn't get into each item in each of the tuple lists
Instead, I'll iterate over each of the tuple lists

ocs_ci/utility/aws.py
```diff
@@ -304,18 +309,32 @@ def start_ec2_instances(self, instances, wait=False):
                 False otherwise
 
         """
+        stopping_instances = {
```
RazTamir: This code block shouldn't be here.
This function should start ec2 instances and that's it - no need to check if the instances is in stopping state.
This sampling should happen, but not here

ebenahar:
=========================================================================================== 2 passed in 1255.28 seconds ===========================================================================================
[shutdown_logs.zip](https://github.com/red-hat-storage/ocs-ci/files/3465183/shutdown_logs.zip)

ocs_ci/ocs/ocp.py
```diff
@@ -376,3 +376,42 @@ def switch_to_default_rook_cluster_project():
...
+    Returns:
+        bool: True if all nodes reached status Ready
+
+    """
+    try:
+        for sample in TimeoutSampler(timeout=120, sleep=3, func=get_all_nodes):
+            if len(sample) == num_of_nodes:
+                ready_nodes = 0
+                for node in sample:
+                    for con in node.get('status').get('conditions'):
```
MeridianExplorer: I feel 'con' might be confusing - what do you think about renaming it to condition?

ebenahar: done

ocs_ci/utility/aws.py
```diff
@@ -262,50 +268,79 @@ def detach_and_delete_volume(self, volume, timeout=120):
...
-        state = res.get('StoppingInstances')[0].get('CurrentState').get('Code')
-        return state == STOPPING
-
-    def start_ec2_instance(self, instance_id, wait=False):
+            for instance_id, instance_name in instances.items():
+                logger.info(
+                    f"Waiting for instance {instance_name} to reach status stopped"
+                )
+                instance = self.get_ec2_instance(instance_id)
+                instance.wait_until_stopped()
```
MeridianExplorer: Just to make sure - if we'll receive a timeout, the test *should* fail and go to teardown, right? There's no need for try-except

tests/manage/cluster/nodes/test_nodes_restart.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    nodes = ocp.get_all_nodes()
+    ec2_instances = aws.get_instances_ids_and_names(nodes)
+
+    def finalizer():
+        """
+        Make sure all instances are running
+        """
+        stopping_instances = {
+            key: val for key, val in ec2_instances.items() if (
```
MeridianExplorer: If we're not using the values, I think it's better to iterate only over the dictionary's keys
```suggestion
            instanceid for instanceid in ec2_instances if (
```

tests/manage/cluster/nodes/test_nodes_restart.py
```diff
@@ -0,0 +1,206 @@
...
+                aws_obj.get_instances_status_by_id(key) == constants.INSTANCE_STOPPED
+            )
+        }
+        if stopped_instances:
+            aws_obj.start_ec2_instances(instances=stopped_instances, wait=True)
+
+    request.addfinalizer(finalizer)
+
+    return ec2_instances
```
MeridianExplorer: Do we need a check to see whether all the instances started?

tests/manage/cluster/nodes/test_nodes_restart.py
```diff
@@ -0,0 +1,206 @@
...
+@pytest.fixture()
+def resources(request):
+    """
+    Delete the resources created during the test
+
+    Returns:
+        tuple: empty lists of resources
+
+    """
+    projects, secrets, pools, storageclasses, pvcs, pods = ([] for i in range(6))
```
MeridianExplorer: I suspect the uninitialized list might cause issues - maybe it'll be safer to initialize it using
```suggestion
    projects, secrets, pools, storageclasses, pvcs, pods = ([None] for i in range(6))
```

ocs_ci/ocs/ocp.py
```diff
@@ -376,3 +376,42 @@ def switch_to_default_rook_cluster_project():
...
+                        if 'True' in con.get('status'):
+                            if con.get('type') == 'Ready':
+                                ready_nodes += 1
+                    if ready_nodes == num_of_nodes:
+                        return True
+    except TimeoutExpiredError:
+        log.error("Not all nodes reached status Ready")
```
nehaberry: @ebenahar will it be possible to display exactly which all nodes didn't reach Ready state ? Just a question

tests/manage/cluster/nodes/test_nodes_restart.py
dahorak: @MeridianExplorer @ebenahar Do you think, this is more readable then the original version?
If you want to squash it to one line, I would propose this:
```
    projects = secrets = pools = storageclasses = pvcs = pods = list()
```

ocs_ci/utility/aws.py
ebenahar: in case of more timeout (10 minutes), we'll go directly to teardown

tests/manage/cluster/nodes/test_nodes_restart.py
ebenahar: we are doing it inside  aws_obj.start_ec2_instances()

ocs_ci/ocs/ocp.py
ebenahar: Thanks, I'll add this as part of https://github.com/red-hat-storage/ocs-ci/issues/504

tests/manage/cluster/nodes/test_nodes_restart.py
ebenahar: @dahorak it's more organized. I don't have a preference though

PR #505, opened 2019-08-05
Add force download for client and installer

petr-balogh:


ocs_ci/deployment/aws.py
```diff
@@ -89,7 +89,8 @@ def __init__(self):
         self.name = self.__class__.__name__
         super(AWSIPI, self).__init__()
         self.installer = utils.get_openshift_installer(
-            config.DEPLOYMENT['installer_version']
+            config.DEPLOYMENT['installer_version'],
+            force_download=config.DEPLOYMENT['force_download_installer'],
```
fbalak: Just maybe it would be good to add something like:
```
force_download = DEPLOYMENT.get('force_download_installer') in ("true", "True", True)
```

petr-balogh: The yaml file provides you the ability to use boolean values and as we are loading the yaml, ppl are responsible to use proper boolean value like described for example here:  https://yaml.org/type/bool.html .

As we have defined default values as True for both client and installer I didn't use get as we should have default value always defined.

Issue #509, opened 2019-08-06
Force download of client and installer should be done during deployment or teardown only


sidhant-agrawal:
Client and Installer are being force download even when other scripts are run.

```
$ run-ci -m "tier2" --cluster-name sagrawal-ocs --cluster-path /home/sagrawal/ocs-runs/ tests/manage/pv_services/test_delete_create_pvc_same_name.py --no-print-logs
Collecting Cluster versions
========================================================== test session starts ==========================================================
platform linux -- Python 3.7.4, pytest-5.0.1, py-1.8.0, pluggy-0.12.0
rootdir: /home/sagrawal/Projects/github/ocs/ocs-ci-fork, inifile: pytest.ini
plugins: html-1.21.1, reportportal-1.0.5, metadata-1.8.0, marker-bugzilla-0.9.1.dev2, logger-0.5.1
collected 2 items                                                                                                                       

tests/manage/pv_services/test_delete_create_pvc_same_name.py::TestDeleteCreatePVCSameName::test_pvc_delete_create_same_name[CephBlockPool] 
------------------------------------------------------------ live log setup -------------------------------------------------------------
11:28:48 - MainThread - tests.conftest - INFO - All logs located at /tmp/ocs-ci-logs-1565071111
11:28:48 - MainThread - ocs_ci.deployment.factory - INFO - Deployment key = aws_ipi
11:28:48 - MainThread - ocs_ci.deployment.factory - INFO - Current deployment platform: AWS,deployment type: ipi
11:28:48 - MainThread - ocs_ci.utility.utils - INFO - Downloading openshift installer (4.2.0-0.nightly-2019-07-30-073644).
11:30:20 - MainThread - ocs_ci.utility.utils - INFO - Executing command: tar xzvf openshift-install.tar.gz openshift-install
11:30:22 - MainThread - ocs_ci.utility.utils - INFO - Executing command: ./bin/openshift-install version
11:30:22 - MainThread - ocs_ci.utility.utils - INFO - OpenShift Installer version: ./bin/openshift-install v4.2.0-201907300219-dirty
                                                      built from commit 9323138c52b773e31b7c98d8a934f9ddaa6ccaee
                                                      release image registry.svc.ci.openshift.org/ocp/release@sha256:eb93f85c7caec2d7603f384f1a0ae6c004fc3dcdff90c957b074c4ff1b790e2b
[...]

PR #513, opened 2019-08-07
set default value of force_download_client to False

vasukulkarni:
set default value of force_download_client to False to revert back to old
behavior i,e not download oc client during tier runs with existing cluster.

Fixes: https://github.com/red-hat-storage/ocs-ci/issues/509


clacroix12:
I thought we wanted to do this for test execution only and still default to True if we are deploying or destroying a cluster?

RazTamir:
Yes we do.
This should depend on the '-m deployment' parameter passed to run-ci command

petr-balogh:
This can be closed as I posted #514 to do force download only when deployment.

PR #517, opened 2019-08-07
update installer and client version

dahorak:
to current latest nightly builds

It is probably affected by this known issue: https://bugzilla.redhat.com/show_bug.cgi?id=1736800 (affecting clusters running more than 24 hours).

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -19,14 +19,14 @@ RUN:
   run_id: null  # this will be redefined in the execution
   kubeconfig_location: 'auth/kubeconfig' # relative from cluster_dir
   cli_params: {}  # this will be filled with CLI parameters data
-  client_version: '4.2.0-0.nightly-2019-07-30-073644'
+  client_version: '4.2.0-0.nightly-2019-08-06-195545'
   bin_dir: './bin'
   google_api_secret: '~/.ocs-ci/google_api_secret.json'
 
 # In this section we are storing all deployment related configuration but not
 # the environment related data as those are defined in ENV_DATA section.
 DEPLOYMENT:
-  installer_version: "4.2.0-0.nightly-2019-07-30-073644"
+  installer_version: "4.2.0-0.nightly-2019-08-06-195545"
   force_download_installer: True
   force_download_client: True
 


dahorak:
Installation and destroy passed, waiting for tier1 test results.

dahorak:
As mentioned in the description, this version will be probably affected by BZ 1736800, which means that any cluster will be broken after 24 hours, but I think it is better than the current situation, where we are not able to even install it.

dahorak:
@nehaberry @RazTamir are you OK with this update even with the known issue https://bugzilla.redhat.com/show_bug.cgi?id=1736800?

RazTamir:
If we know about this issue why we want to change the versions at all?

dahorak:
@RazTamir without this change we are not able to install the cluster at all (= CI is broken), with this change we will know, that any installed cluster will be broken in 24 hours.. Which is ok for CI... And for manual testing, if people are able to install with the older version, they can change it manually... or rebuild it once a day...

dahorak:
tier1 test results are ok

vasukulkarni:
I am sorry, but I dont understand why we have to use dated versions, this seems so wrong, if OCS = latest OCP + OCS, then our pipeline is a broken testing.

RazTamir:
OCS = OCS latest + OCP stable

vasukulkarni:
> @RazTamir without this change we are not able to install the cluster at all (= CI is broken), with this change we will know, that any installed cluster will be broken in 24 hours.. Which is ok for CI... And for manual testing, if people are able to install with the older version, they can change it manually... or with the new version rebuild it once a day... (I'm not sure, what is the real issue here, because it seems that for some people the installation with the older version works and for some other not...)

Since the cluster is going to be broken in 24 hours, I *propose* that instead of this change, you take --installer-version and --client-version as params and make it more flexible.

clacroix12:
> Since the cluster is going to be broken in 24 hours, I _propose_ that instead of this change, you take --installer-version and --client-version as params and make it more flexible.

We don't need additional parameters to enable this, we can already pass a custom config with `--ocsci-conf` which sets the client and installer versions.

vasukulkarni:
> > Since the cluster is going to be broken in 24 hours, I _propose_ that instead of this change, you take --installer-version and --client-version as params and make it more flexible.
> 
> We don't need additional parameters to enable this, we can already pass a custom config with `--ocsci-conf` which sets the client and installer versions.

Do you see the context I am trying to say here? I know you have the conf file, but for CI system(*nightly* runs) they have to edit the file which is not trivial compared to modifying the CLI in jenkins.

petr-balogh:
Yep, anyway to not block CI I merged this one and we will discuss in mail thread and ACK on what we really want, the discussion here is not part of this PR.

petr-balogh:
> > > Since the cluster is going to be broken in 24 hours, I _propose_ that instead of this change, you take --installer-version and --client-version as params and make it more flexible.
> > 
> > 
> > We don't need additional parameters to enable this, we can already pass a custom config with `--ocsci-conf` which sets the client and installer versions.
> 
> Do you see the context I am trying to say here? I know you have the conf file, but for CI system(_nightly_ runs) they have to edit the file which is not trivial compared to modifying the CLI in jenkins.

@vasukulkarni  the job we are using is consumed from jenkins file (pipeline) defined in our repository. And it's fetch once the job start from the master. So it doesn't matter if we will need to change the version in config for ocs-ci or if we will need send PR to jenkins repository to change the version there.

vasukulkarni:
> > > > Since the cluster is going to be broken in 24 hours, I _propose_ that instead of this change, you take --installer-version and --client-version as params and make it more flexible.
> > > 
> > > 
> > > We don't need additional parameters to enable this, we can already pass a custom config with `--ocsci-conf` which sets the client and installer versions.
> > 
> > 
> > Do you see the context I am trying to say here? I know you have the conf file, but for CI system(_nightly_ runs) they have to edit the file which is not trivial compared to modifying the CLI in jenkins.
> 
> @vasukulkarni the job we are using is consumed from jenkins file (pipeline) defined in our repository. And it's fetch once the job start from the master. So it doesn't matter if we will need to change the version in config for ocs-ci or if we will need send PR to jenkins repository to change the version there.

so what is the reason for this change? maybe we should file github issues and describe if its solving an issue or an enhancement?

petr-balogh:
With previous version the deployment didn't work at all cause of some issue with pull secret.  So all jenkins job failed to deploy the OCP.   With this version the deploy works, but after 24 hours the OCP get corrupted cause of known bug but at least for our jenkins runs it's working.  Anyway, we were sticked with some version in master as well. This change at least allow us to run CI jobs and not fail them cause of pull secret issue for previous used version.

PR #511, opened 2019-08-06
Generic fixtures for parametrize usage

RazTamir:
- Added a fixture to iterate over ceph interfaces
 - Created generic fixtures to create resources based on interface

PR #519, opened 2019-08-07
Add factory fixtures

fbalak:
Add factory fixtures from https://github.com/red-hat-storage/ocs-ci/pull/418 so they can be used in new tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,7 +21,9 @@ from ocs_ci.utility.utils import (
 )
 from ocs_ci.deployment import factory as dep_factory
 from tests import helpers
-from ocs_ci.ocs import constants
+from ocs_ci.ocs import constants, defaults, ocp
+from ocs_ci.ocs.resources.ocs import OCS
+from ocs_ci.ocs.resources.pvc import PVC
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -44,6 +46,523 @@ def pytest_logger_config(logger_config):
...
+
+@pytest.fixture()
+def pod_factory(request):
+    """
+    Create a Pod factory. Calling this fixture creates new Pod.
+    For custom Pods provide 'pvc' parameter.
+    """
+    instances = []
+
+    def factory(pvc=None, custom_data=None, status=constants.STATUS_RUNNING):
```
clacroix12: Are we sure we want to to default `pvc` to `None`? We aren't creating one if it isn't provided by the caller.

tests/conftest.py
```diff
@@ -44,6 +46,523 @@ def pytest_logger_config(logger_config):
...
+                instance.ocp.wait_for_delete(
+                    instance.name
+                )
+
+    request.addfinalizer(finalizer)
+    return factory
+
+
+@pytest.fixture()
+def rbd_pvc_factory(rbd_storageclass_factory, pvc_factory):
```
petr-balogh: Do I understand correctly that this one returns the same as one above when you use it as I see that in pvc_factory you are also using rbd_storageclass_factory? So maybe it's redundant and if we will document that pvc_factory is by default using rbc storage class we don't need to have this?  Or do I miss something?

tests/conftest.py
petr-balogh: I understand this as this is a general one and you have to pass pvc here if you don't provide custom_data.  So maybe worth to check on the line 470 that pvc objec is set and log the error. Anyway it will raise an exception AttributeError: 'NoneType' object has no attribute 'storageclass', right?

tests/conftest.py
```diff
@@ -44,6 +46,523 @@ def pytest_logger_config(logger_config):
...
+            object: helpers.create_pvc instance.
+        """
+        if custom_data:
+            pvc_obj = PVC(**custom_data)
+            pvc_obj.create(do_reload=False)
+        else:
+            nonlocal active_project
+            project = project or active_project or project_factory()
+            active_project = project
+            storageclass = storageclass or rbd_storageclass_factory()
```
petr-balogh: Will this create new storage class for each pvc?  Is this expected behavior to create for each pvc it's own storageclass? Or do we want to share one storage class for all of PVCs created in one object of pvc factory?

tests/conftest.py
clacroix12: I think if we are defaulting to `None` because we don't need it if we pass `custom_data` we should check for the existence of pvc if we don't provide `custom_data` and raise a proper exception if neither exist.

tests/conftest.py
clacroix12: I think you're right. It might be more clear to users if we only provide one `pvc_factory` and have the returned function take the interface as an argument. This can be propagated to the storageclass_factory as well. I have an example of how that would work here: https://github.com/clacroix12/ocs-ci/commit/80e8a384713fa9e497735a792a096fc4dcd26223

tests/conftest.py
clacroix12: It will create a new storage class per pvc through the `rbd_storageclass_factory` if one isn't provided, yes. If storageclass isn't something that needs to be created on a per pvc basis, we can always change the scope of the fixture to `class` instead of defaulting to `function`.

fbalak: I will create a similar construct to `active_project` so that if there is not provided custom storageclass and storageclass based on an interface exists then there will be not created new storageclass.

tests/conftest.py
fbalak: Ok, I will remove `cephfs_` and `rbd_` prefixes from all fixtures and make them universal with `interface` parameter set to rbd.

tests/conftest.py
```diff
@@ -44,6 +46,349 @@ def pytest_logger_config(logger_config):
...
+    Create a Ceph block pool factory.
+    Calling this fixture creates new block pool instance.
+    """
+    instances = []
+
+    def factory():
+        cbp_obj = helpers.create_ceph_block_pool()
+        assert cbp_obj, "Failed to create block pool"
+        instances.append(cbp_obj)
+        return cbp_obj
```
RazTamir: To make it more generic, we will need also the option to return the cephfilesystem object in case the interface is cephFileSystem

RazTamir:
Looks good @fbalak .
Is there anything blocking us from merging this? (I added 1 comment)

tests/conftest.py
fbalak: It seems to me from code that we are currently not creating pool for CephFS but that we just look for it with [helpers.get_cephfs_data_pool_name()](https://github.com/red-hat-storage/ocs-ci/blob/cacbd94e4bfb05124ce1520d08478194d3c5684d/tests/helpers.py#L405). I am not sure if this is a good use case.
I would keep this factory as is because block pool is relevant only to RBD.

RazTamir: We can change the name as well to make it more universal.
In case you prefer to keep it as is, let's merge this PR and I will make this more generic in different PR.
We really need the implementation of https://github.com/red-hat-storage/ocs-ci/pull/511 to enable the parametrize of RBD and cephFS

petr-balogh:
@fbalak, have you tried to execute some tests if it doesn't break anything if merged? Conftest files are loaded by pytest so just want to make sure it doesn't break anything here.  If tested (add verified label), we can merge.

fbalak:
> @fbalak, have you tried to execute some tests if it doesn't break anything if merged? Conftest files are loaded by pytest so just want to make sure it doesn't break anything here. If tested (add verified label), we can merge.

I tried just:
```
def test_it(pod_factory):
    pod = pod_factory()
```
I will do more testing when installation of my cluster is completed and update this PR but as I am not removing anything and those fixtures are not used anywhere, it should not break anything.

fbalak:
Testing with CephFileSystem interface passed too: https://paste.fedoraproject.org/paste/8z2GTdoro8a7w4AXQbA7ig and it seems that it doesn't break anything as `tests/e2e/test_jenkins_simulation.py` also passes.

PR #520, opened 2019-08-07
Add a Jenkins pipeline definition

zmc:
This is what we'll use to validate builds internally (it can also be used upstream).

### DIFF
--- a/.functional_ci_setup.py
+++ b/.functional_ci_setup.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python
+import base64
+import binascii
+import os
+import yaml
+
+from os import environ as env
+from configparser import ConfigParser
+
+
+def write_aws_creds():
+    aws_profile = env['AWS_PROFILE']
+    # Write the credentials file
+    creds = ConfigParser()
+    creds[aws_profile] = dict(
+        aws_access_key_id=env['AWS_ACCESS_KEY_ID'],
+        aws_secret_access_key=env['AWS_SECRET_ACCESS_KEY'],
+    )
+    creds_path = env['AWS_SHARED_CREDENTIALS_FILE']
+    os.makedirs(
+        os.path.dirname(creds_path),
+        exist_ok=True,
+    )
+    with open(creds_path, 'w') as creds_file:
+        creds.write(creds_file)
+
+    # Write the config file
+    conf = ConfigParser()
+    conf[aws_profile] = dict(
+        region=env['AWS_REGION'],
+        output='text',
+    )
+    conf_path = env['AWS_CONFIG_FILE']
+    os.makedirs(
+        os.path.dirname(conf_path),
+        exist_ok=True,
+    )
+    with open(conf_path, 'w') as conf_file:
+        conf.write(conf_file)
+
+
+def write_pull_secret():
+    secret_dir = os.path.join(env['WORKSPACE'], 'data')
+    os.makedirs(secret_dir, exist_ok=True)
+    secret = env['PULL_SECRET']
+    # In Jenkins, this is a JSON string. In GitLab CI, the JSON will be
+    # base64-encoded.
+    try:
+        secret = base64.b64decode(secret).decode()
+    except binascii.Error:
+        pass
+    with open(os.path.join(secret_dir, 'pull-secret'), 'w') as secret_file:
+        secret_file.write(secret)
+
+
+def get_ocsci_conf():
+    cluster_user = env['CLUSTER_USER']
+    pipeline_id = env['BUILD_ID']
+    conf_obj = dict(
+        RUN=dict(
+            log_dir=os.path.join(env['WORKSPACE'], 'logs'),
+        ),
+        ENV_DATA=dict(
+            platform='AWS',
+            cluster_name=f"{cluster_user}-ocs-ci-{pipeline_id}",
+            region=env['AWS_REGION'],
+            base_domain=env['AWS_DOMAIN'],
+        ),
+    )
+    # Apply image configuration if present
+    for image_type in ['rook', 'ceph']:
+        image_key = f"{image_type}_image"
+        image_value = env.get(image_key.upper())
+        if image_value is not None:
+            conf_obj['ENV_DATA'][image_key] = image_value
+    return conf_obj
+
+
+def write_ocsci_conf():
+    ocp_conf = get_ocsci_conf()
+    ocp_conf['ENV_DATA']['skip_ocs_deployment'] = True
+    ocp_conf_path = os.path.join(env['WORKSPACE'], 'ocs-ci-ocp.yaml')
+    with open(ocp_conf_path, 'w') as ocp_conf_file:
+        ocp_conf_file.write(yaml.safe_dump(ocp_conf))
+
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,94 @@
...
+      returnStdout: true,
+      script: 'echo $WORKSPACE/.aws/credentials'
+      )}"""
+    AWS_CONFIG_FILE = """${sh(
+      returnStdout: true,
+      script: 'echo $WORKSPACE/.aws/config'
+      )}"""
+    AWS_ACCESS_KEY_ID = credentials('openshift-dev-aws-access-key-id')
+    AWS_SECRET_ACCESS_KEY = credentials('openshift-dev-aws-secret-access-key')
+    PULL_SECRET = credentials('openshift-pull-secret')
```
petr-balogh: you should use withCredentials binding only on particular places where you really need to have those variables. See examples here:  https://jenkins.io/doc/pipeline/steps/credentials-binding/

I think that you need this pull secret only for deploy so no need to give this PULL_SECRET to all the other parts for example where someone can inject in code?

.functional_ci_setup.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
```
petr-balogh: Can we move this file in some folder for the jenkins related stuff? Probablhy we will need more files in future, so will be good have them in one place.  Is there specific reason to have this file as hidden prefixed with dot?

Jenkinsfile
```diff
@@ -0,0 +1,94 @@
...
+//   ROOK_IMAGE
+//   CEPH_IMAGE
+pipeline {
+  agent { node { label "rhel-8-large" }}
+  environment {
+    AWS_SHARED_CREDENTIALS_FILE = """${sh(
+      returnStdout: true,
+      script: 'echo $WORKSPACE/.aws/credentials'
+      )}"""
+    AWS_CONFIG_FILE = """${sh(
```
petr-balogh: WORKSPACE variable is directly available from pipeliony, Why not just AWS_CONFIG_FILE = "${WORKSPACE}/.aws/config" , the same above

.functional_ci_setup.py
zmc: Well, this file isn't just for Jenkins - so I'm not sure what I'd call the subdirectory to put this in. It also doesn't have to be a hidden file - I just didn't want people to be tempted to execute it manually, since it won't do anything useful.

.functional_ci_setup.py
```diff
@@ -0,0 +1,96 @@
...
+        exist_ok=True,
+    )
+    with open(conf_path, 'w') as conf_file:
+        conf.write(conf_file)
+
+
+def write_pull_secret():
+    secret_dir = os.path.join(env['WORKSPACE'], 'data')
+    os.makedirs(secret_dir, exist_ok=True)
+    secret = env['PULL_SECRET']
```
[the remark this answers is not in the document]
zmc: Not sure what you mean here. In Jenkins, this will just be a JSON string. In GitLab CI, it *must* be base64-encoded, as that's a requirement of theirs for secrets beyond a certain length.

Jenkinsfile
```diff
@@ -0,0 +1,94 @@
+// vim: et:ts=2:sw=2
```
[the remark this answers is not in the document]
zmc: https://vim.fandom.com/wiki/Modeline_magic

It's extremely rare for someone to have an editor setup which knows about Jenkins groovy scripts, so this at least tells vim how the file should be formatted.

Jenkinsfile
zmc: Does `withCredentials` work in declarative pipelines? If so, I agree that it should be used to scope this particular secret to the setup stage.

Jenkinsfile
zmc: I thought that didn't work with declarative pipelines? I could be mistaken here. Happy to change if that's the case.

Jenkinsfile
dahorak: Yes, `withCredentials` works in declarative pipeline.

.functional_ci_setup.py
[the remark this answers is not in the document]
petr-balogh: If it's cause of the length you can directly use jenkins credentials as secret file, you can directly upload this file and use it an copy it to proper location you need from the pipeline directly

.functional_ci_setup.py
petr-balogh: Ok, lets create some specific folder for those jenkins scripts called .jenkins , .jenkins-ci or whatever better in your mind for example ?

Jenkinsfile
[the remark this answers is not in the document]
petr-balogh: Yep, I know this syntax for vim, but no everyone is using vim so it was weird to me tho have this settings directly here in the file. I personally have this option in my vim configuration.

Jenkinsfile
petr-balogh: I think it should work also in declarative pipeline, maybe it's just ${env.WORKSPACE}

Jenkinsfile
petr-balogh: I think it should work

.functional_ci_setup.py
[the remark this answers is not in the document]
zmc: Decoding doesn't happen at all when this is run in Jenkins, though.

Jenkinsfile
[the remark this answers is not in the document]
zmc: I'll add an editorconfig file

.functional_ci_setup.py
zmc: Yeah, again, this won't just be useful for Jenkins. And if we choose to use GitLab CI in the future, its configuration file has to be in the project's root directory - similar to TravisCI.

As it stands right now, there are two files that we could put in this directory. Could we maybe decide how to organize things if and when we have more?

petr-balogh: I am not against config files have in root directory if some system counts on them to be in root directory. But this file is not any configuration file for any CI system and it's used directly in pipeline (So I think it can be easily moved to some specific folder like ci-scripts if we don't like to use jenkins in the name). So as @clacroix12  already tried to do some restructure/cleanup of the project and cleaned everything to specific folders I don't like to have any extra python file in root of the project.  @clacroix12 thoughts?   I won't block this PR cause of this but still think we don't need it in the root.

clacroix12: @petr-balogh The only thing I can think of is something like `.pipelines/` which would keep it arbitrary as far as what CI system is consuming it. A `.` dir to perhaps keep it out of the way and like @zmc said to keep people from trying to execute it manually.

Overall, I don't think it's a big deal to leave it as is but if we did want to have it in a directory that's how I would approach it.

clacroix12:
@zmc since this is for a pipeline, it's not important as far as packaging goes correct? So we wouldn't need any changes to `MANIFEST.in`?

zmc:
@clacroix12 yeah, I think these files will be irrelevant to packaged versions of the framework.

zmc:
@clacroix12 I like your idea - I can do a follow-up shortly

PR #529, opened 2019-08-08
Fixtures modification

ebenahar:
I'd like to propose one more option for fixtures implementation 

### DIFF
--- a/tests/fixtures_new.py
+++ b/tests/fixtures_new.py
@@ -1,0 +1,240 @@
+import pytest
+from tests import helpers
+from ocs_ci.ocs import constants, ocp
+
+
+@pytest.fixture()
+def secret(request, interface_type):
+    """
+    Create a secret
+
+    Returns:
+        OCS: An instance of secret OCS
+
+    """
+    secrets = list()
+
+    def finalizer():
+        """
+        Delete the RBD secret
+        """
+        for secret in secrets:
+            secret.delete()
+            secret.ocp.wait_for_delete(secret.name)
+
+    request.addfinalizer(finalizer)
+    secrets.append(helpers.create_secret(interface_type=interface_type))
+
+    return secrets
+
+
+@pytest.fixture()
+def pool(request, interface):
+    """
+    Create a pool
+
+    Returns:
+
+    """
+    pools = list()
+
+    def finalizer():
+        """
+        Delete the Ceph block pool
+        """
+        if interface == constants.CEPHBLOCKPOOL:
+            for pool in pools:
+                pool.delete()
+                pool.ocp.wait_for_delete(pool.name)
+
+    request.addfinalizer(finalizer)
+
+    if interface == constants.CEPHBLOCKPOOL:
+        pools.append(helpers.create_ceph_block_pool())
+    elif interface == constants.CEPHFILESYSTEM:
+        pools.append(helpers.get_cephfs_data_pool_name())
+    return pools
+
+
+@pytest.fixture()
+def storageclass(request, secret, pool, interface_type):
+    """
+    Create storage class
+
+    Returns:
+        OCS: An instance of storageclass OCS
+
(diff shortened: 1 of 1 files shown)


tests/fixtures_new.py
```diff
@@ -0,0 +1,240 @@
+import pytest
+from tests import helpers
+from ocs_ci.ocs import constants, ocp
+
+
+@pytest.fixture()
+def secret(request, interface_type):
```
petr-balogh: is interface_type other fixture? Or it's parameter here? If it's parameter I don't think you can pass parameter to fixture like this. You will have to create fixture factory to have this capability

tests/fixtures_new.py
```diff
@@ -0,0 +1,240 @@
...
+        Delete the RBD secret
+        """
+        for secret in secrets:
+            secret.delete()
+            secret.ocp.wait_for_delete(secret.name)
+
+    request.addfinalizer(finalizer)
+    secrets.append(helpers.create_secret(interface_type=interface_type))
+
+    return secrets
```
petr-balogh: It should return single resource here I guess as have single name in the fixture, look at Filips pr how those fixture factories are defined if you would like to achieve something similar.

tests/fixtures_new.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+@pytest.fixture()
+def pool(request, interface):
```
petr-balogh: same

tests/fixtures_new.py
```diff
@@ -0,0 +1,240 @@
...
+        """
+        for sc in storageclasses:
+            sc.delete()
+            sc.ocp.wait_for_delete(sc.name)
+
+    request.addfinalizer(finalizer)
+    storageclasses.append(
+        helpers.create_storage_class(
+            interface_type=interface_type,
+            interface_name=pool[0].name,
```
petr-balogh: The pool is by default function level scope, so it means that if you will use it in more tests in test class it will in your approach create pool for each storageclass but only first one will be used right?  But as both are function level scope you shouldn't have any other pool than [0] if I understand correctly.

Anyway, what if you would like to create 5 storageclass? I don't think it will work here for one test.

tests/fixtures_new.py
```diff
@@ -0,0 +1,240 @@
...
+    """
+    storageclasses = list()
+
+    def finalizer():
+        """
+        Delete the storage class
+
+        """
+        for sc in storageclasses:
+            sc.delete()
```
petr-balogh: What if storageclass got deleted from test intentionally? Than your attempt to delete here will fail, I think you have to check if the sc is not already deleted.

ebenahar: If the storageclass was deleted in the test then here storageclass will be an empty list and nothing will happen

petr-balogh: Can you give me example of delete from test?

if you will have 
```python
def test_sc(storageclass):
    storageclass[0].delete()
    #this won't delete the object from list provided by fixture.
    # This case you have to also do del myl[0] or something like:
    my_sc = storageclass.pop(0)
    my_sc.delete()
```

ebenahar:
@petr-balogh I wanted to propose this way because I think it's easier to use than the factory method.
We can look at it as one more option for fixtures for us to pick from

tests/fixtures_new.py
clacroix12: @petr-balogh @ebenahar this seems like something we would like the `delete()` call to handle instead of having to perform the check in all. utilities / fixtures. https://github.com/red-hat-storage/ocs-ci/issues/557

tests/fixtures_new.py
clacroix12: Yeah, the only way I know of to allow fixtures to use parameters is to use the factory approach.

PR #536, opened 2019-08-08
Add nodes operations

ebenahar:
Add node operations

  - Nodes maintenance related functions: 
    Scheduling, unscheduling, drain and full maintenance
  - Currently, due to the fact that pods are still using local storage, these operations won't work
    with OCS. See BZ 1734162

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -1,0 +1,117 @@
+from __future__ import print_function
+
+import logging
+from ocs_ci.ocs.exceptions import TimeoutExpiredError
+from ocs_ci.ocs.ocp import OCP
+from ocs_ci.ocs import constants
+from ocs_ci.utility.utils import TimeoutSampler
+
+
+log = logging.getLogger(__name__)
+
+
+def get_node_objs(node_names=None):
+    """
+    Get node objects by node names
+
+    Args:
+        node_names (list): The node names to get their objects for.
+            If None, will return all cluster nodes
+
+    Returns:
+        list: Cluster node OCP objects
+
+    """
+    nodes_obj = OCP(kind='node')
+    node_objs = nodes_obj.get()['items']
+    if not node_names:
+        return node_objs
+    else:
+        return [
+            node_obj for node_obj in node_objs if (
+                node_obj.get('metadata').get('name') in node_names
+            )
+        ]
+
+
+def wait_for_nodes_status(node_names, status=constants.NODE_READY):
+    """
+    Wait until all nodes are in Ready status
+
+    Args:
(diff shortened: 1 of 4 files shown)


vasukulkarni:
Please add more description to commit message, otherwise LGTM.

ebenahar:
Verified the TC that uses the existing functions that their location was changed
[test_logs.zip](https://github.com/red-hat-storage/ocs-ci/files/3482578/test_logs.zip)


=========================================================================================== 1 passed in 621.55 seconds ============================================================================================

ocs_ci/ocs/node.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Returns:
+        bool: True if all nodes reached status Ready, False otherwise
+
+    """
+    try:
+        for sample in TimeoutSampler(120, 3, get_node_objs, node_names):
+            for node in sample:
+                if not node_names:
+                    return True
```
jilju: Just for clarification. What is the expected result here?
If node_names=None (waiting for all nodes to be in status "status"), won't this give a false pass ?

ocs_ci/ocs/node.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    Wait until all nodes are in Ready status
+
+    Args:
+        node_names (list): The node names to wait for to be in Ready state
+        status (str): The node status to wait for
+            (e.g. 'Ready', 'NotReady', 'SchedulingDisabled')
+
+    Returns:
+        bool: True if all nodes reached status Ready, False otherwise
```
jilju:
```
        bool: True if all nodes reached status Ready, False otherwise
```
Consider other status as mentioned in
status (str): The node status to wait for
            (e.g. 'Ready', 'NotReady', 'SchedulingDisabled')

ocs_ci/ocs/node.py
ebenahar: the default value for status is Ready so in case you don't pass status, the function will wait for Ready

ocs_ci/ocs/node.py
ebenahar: good catch

ocs_ci/ocs/node.py
jilju: It will be good to enhance this function to wait for all nodes to reach a desired status without the need of passing the node_names while calling this function. 
node_names = None
get_node_objs function will return all nodes if node_names=None. So status of all nodes can be checked.

ebenahar: @jilju good idea, added

ebenahar:
=========================================================================================== 1 passed in 624.28 seconds ============================================================================================

PR #560, opened 2019-08-13
use uuid4 in hex as random suffix for ocs objects

vasukulkarni:
Datetime suffix doesn't sufficiently serve as random number when using threadpools where numerous objects are created within millisecond and high possibility of name collision occurs.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -3,6 +3,7 @@ Helper functions file for OCS QE
 """
 import logging
 import re
+import datetime
 
 from uuid import uuid4
 from ocs_ci.ocs.exceptions import TimeoutExpiredError
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -1,9 +1,10 @@
 """
 Helper functions file for OCS QE
 """
-import datetime
 import logging
 import re
+
+from uuid import uuid4
 from ocs_ci.ocs.exceptions import TimeoutExpiredError
 from ocs_ci.ocs import constants, defaults, ocp
 from ocs_ci.utility import templating
@@ -18,8 +19,8 @@ logger = logging.getLogger(__name__)
 
 def create_unique_resource_name(resource_description, resource_type):
     """
-    Creates a unique object name by using the object_description
-    and object_type, as well as the current date/time string.
+    Creates a unique object name by using the object_description,
+    object_type and a random uuid(in hex) as suffix
 
     Args:
         resource_description (str): The user provided object description
@@ -29,10 +30,7 @@ def create_unique_resource_name(resource_description, resource_type):
     Returns:
         str: A unique name
     """
-    current_date_time = (
-        datetime.datetime.now().strftime("%d%H%M%S%f")
-    )
-    return f"{resource_type}-{resource_description[:23]}-{current_date_time[:10]}"
+    return f"{resource_type}-{resource_description[:23]}-{uuid4().hex}"
 
 
 def create_resource(


tests/helpers.py
```diff
@@ -19,7 +19,7 @@
 def create_unique_resource_name(resource_description, resource_type):
     """
     Creates a unique object name by using the object_description
-    and object_type, as well as the current date/time string.
+    and object_type, and a random number between 100 to 1000000
```
petr-balogh: random number is not between 100 and 1000000 as you compose from 2 numbers from random range  100 - 1000000 . and  1000 - 10000 . And then you multiply them so lowest possible is 99 * 999 = 98901. I am proposing some diff approach, se my comment below.

tests/helpers.py
```diff
@@ -29,10 +29,10 @@ def create_unique_resource_name(resource_description, resource_type):
...
         str: A unique name
     """
-    current_date_time = (
-        datetime.datetime.now().strftime("%d%H%M%S%f")
-    )
-    return f"{resource_type}-{resource_description[:23]}-{current_date_time[:10]}"
+    randnum1 = randrange(100, 1000000)
+    randnum2 = randrange(1000, 10000)
+    randnum = randnum1 * randnum2
+    return f"{resource_type}-{resource_description[:23]}-{randnum}"
```
petr-balogh: I am proposing to keep the same length of the random number and keep the date timestamp as it's sometime useful for easy identification when the object got created and you can directly watch the log with this timestamp.

```python
randnum = (
    f'{datetime.datetime.now().strftime("%d%H%M%S")}'
    f'{randrange(1, 10**5):05}'
)
```
Example of 10 numbers generated in for loop:
1309275981707
1309275933379
1309275951637
1309275946251
1309275968797
1309275955706
1309275989516
1309275972782
1309275994294
1309275964369

clacroix12: I would add to this that we should separate the timestamp from the random numbers with `_` or `-`. This way the timestamp can easily be identified.

tests/helpers.py
vasukulkarni: Lowest possible would be 100*1000,  I can update the comment which I remembered later, but having high number is digits is required to avoid collision for scale tests.

tests/helpers.py
vasukulkarni: when running using threadpool the first 8 digits are same which is not what we want, in less than a second we can invoke 1000's of pvc creation rather a truly random number is needed.

clacroix12: if we need a truly unique string why don't we use `uuid`? If using the timestamp isn't a requirement I think this would be a better option for us.

```
In [30]: import uuid

In [31]: uuid.uuid4().hex
Out[31]: 'a0e3c31cbb6641aaa30432f59340d59f'
```

shylesh: Agree with @clacroix12 , If we are not really concerned about timestamp, UUID would be better rather than random number.

vasukulkarni: uuid depends on random program, but I really don't have a preference where it comes from(as long as its thread safe), from the python's own documentation 

```
Python uses the Mersenne Twister as the core generator. It produces 53-bit precision floats and has a period of 2**19937-1. The underlying implementation in C is both fast and threadsafe
```

clacroix12: You can check to see if your uuid was generated in a thread safe way with the `.is_safe` attribute.

```
my_uuid = uuid.uuid4()
my_uuid.is_safe
```

I did some local testing with threads and never hit any conflicts, however the `is_safe` attribute was always unknown for me. Not sure if that is a system issue or what, but you might be able to get different results?

vasukulkarni: I will use uuid4() and rerun the test with 500+ threads, if I dont hit issues I will just change it.

vasukulkarni: I see this is working fine, there are parts in our framework which is causing some issues while testing > 100 but nothing related to uuid part.

tests/helpers.py
```diff
@@ -29,10 +31,7 @@ def create_unique_resource_name(resource_description, resource_type):
     Returns:
         str: A unique name
     """
-    current_date_time = (
-        datetime.datetime.now().strftime("%d%H%M%S%f")
-    )
-    return f"{resource_type}-{resource_description[:23]}-{current_date_time[:10]}"
```
clacroix12: Is there a specific reason we were using the first 10 chars of the `current_date_time`? I just want to make sure with the change to `uuid4().hex` (which is 32 chars) we don't run into name length issues.

vasukulkarni: @ebenahar do you know if there is a restriction on PVC name length? In that case I will cut down the resource_description length to include more of uuid4.

vasukulkarni: I just checked my last run output, I dont see any restrictions actually

```
pvc-test-04557629fd824bafbc188ef52a8ec8fd   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-1cc55a0abd13400db053cdab45f678a1   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   91s
pvc-test-22458113cc4849a3aa63465100ed6fa5   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   96s
pvc-test-280f48a17ea7459bb85498b289ac8c2f   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   91s
pvc-test-31b1a9f6a4b84b7089ef180946c9bdd1   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   96s
pvc-test-3753d2744cce4e569d96e1ba767b5836   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-4ed240960ae94836bd98346e893a069a   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   96s
pvc-test-4efb6e4fc4ca43808fd4f4494b85b248   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-4f7ac597304449cea78dc46b666c457d   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   89s
pvc-test-56b7321cde4448fbba02c2c748abe4fd   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   95s
pvc-test-5afc8f6e37e14f099a3b42b720a70fcf   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   92s
pvc-test-692e5c670b4c45d59e9cd1f20b23ae23   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   88s
pvc-test-6edd02f38369434dae84ffe4ce27c486   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   96s
pvc-test-710aa9626a4d4906809b8e3443ee8ed7   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   91s
pvc-test-7295b78112e64f239f32eae34f45aa16   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   89s
pvc-test-7301a3db1ca443b2a1079b81e667b75d   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   96s
pvc-test-74dfa359ab284147a8cc939fe79f7121   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   93s
pvc-test-7975a8e519ee4704b6b7bec107fe3f09   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-7beb3fb9de83455bb273f97f5087df38   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-88da27cfa62a41abb71daa671952a413   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   96s
pvc-test-8fa57489ff5a417ea99017c1bc07def3   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-91f17acb89ea4b11b1ecc691cb365ebd   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-95b521491a3041b2a7f4839a8d4b3805   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   93s
pvc-test-973e1bfafba64fa482ce69fb9af4d0db   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   89s
pvc-test-9ba6d4946a0e4da5b36aafc662f05328   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   91s
pvc-test-9eac917dc1bd465294c2b7e32703af78   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   96s
pvc-test-9f863a2b1f5b4a87a339fb717ab9112a   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-a0d02b46212342e2a15387b6c8740819   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   92s
pvc-test-aaa056c52f31426f8311ee59cac91d71   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-b14bbc334734466595fcd56fc3e6f70f   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   93s
pvc-test-c53ba3e332e84fe98fcb8c77e3c4c3a0   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-c55dd701e40f410e814cad51c7caf47a   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   96s
pvc-test-cec6d1d2f5434eb3ae112e4a662d447d   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   91s
pvc-test-cfc0dda858cc4f0c9f28780f4ad66a81   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   89s
pvc-test-d09307fff5844769980ecea351d0332b   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   93s
pvc-test-df270741050441fa9ba4f57564ad662c   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-e171095ab86d49c787ef3f75c1d34fe7   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   95s
pvc-test-ea4e898f41d24a5a97a3eb53a9722430   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-f09ca2916c4b4703bf2b17f7d434cec2   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   93s
pvc-test-f149495db2a049199577be40d05f5eb3   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   91s
pvc-test-f71ccd33cdae49a9bdbb282bfa813f91   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
pvc-test-fb5833a637e34f25ac0b85291e95fa2f   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   93s
pvc-test-fd13e17aff034044bb46b669701efaa7   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   96s
pvc-test-fe9b25b7ebcb4df49e00ff387f30b2fc   Pending                                      storageclass-test-3adab1b9a13e4614ac48d922ecf356c9   94s
```

shylesh: > Is there a specific reason we were using the first 10 chars of the `current_date_time`? I just want to make sure with the change to `uuid4().hex` (which is 32 chars) we don't run into name length issues.

I saw some name length problems exceeding 32 chars during UPI but it was not for PVC though. It was related to name generated by CloudFormation templates but this can be handled as and when we hit using automation.

PR #565, opened 2019-08-14
Add teardown factory

RazTamir:
- This will help tearing down resources that were created as part of the test

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -257,7 +257,6 @@ def pvc_factory(
         if custom_data:
             pvc_obj = PVC(**custom_data)
             pvc_obj.create(do_reload=False)
-            instances.append(pvc_obj)
         else:
             project = project or project_factory()
             storageclass = storageclass or storageclass_factory(interface)
@@ -270,11 +269,11 @@ def pvc_factory(
                 wait=False
             )
             assert pvc_obj, "Failed to create PVC"
-            instances.append(pvc_obj)
-            pvc_obj.storageclass = storageclass
-            pvc_obj.project = project
         if status:
             helpers.wait_for_resource_state(pvc_obj, status)
+        pvc_obj.storageclass = storageclass
+        pvc_obj.project = project
+        instances.append(pvc_obj)
 
         return pvc_obj
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -358,8 +358,8 @@ def pod_factory(request, pvc_factory):
 def teardown_factory(request):
     """
     Tearing down a resource that was created during the test
-    To use this factory, you'll need to pass 'teardown_factory' to you test
-    function and call it in you test when a new resource was created and you
+    To use this factory, you'll need to pass 'teardown_factory' to your test
+    function and call it in your test when a new resource was created and you
     want it to be removed in teardown phase:
     def test_example(self, teardown_factory):
         pvc_obj = create_pvc()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -101,8 +101,7 @@ def ceph_pool_factory(request):
             ).get(defaults.CEPHFILESYSTEM_NAME)
             ceph_pool_obj = OCS(**cfs)
         assert ceph_pool_obj, f"Failed to create {interface} pool"
-        if interface == constants.CEPHBLOCKPOOL:
-            instances.append(ceph_pool_obj)
+        instances.append(ceph_pool_obj)
         return ceph_pool_obj
 
     def finalizer():
@@ -360,6 +359,13 @@ def pod_factory(request, pvc_factory):
 def teardown_factory(request):
     """
     Tearing down a resource that was created during the test
+    To use this factory, you'll need to pass 'teardown_factory' to you test
+    function and call it in you test when a new resource was created and you
+    want it to be removed in teardown phase:
+    def test_example(self, teardown_factory):
+        pvc_obj = create_pvc()
+        teardown_factory(pvc_obj)
+
     """
     instances = []
 
@@ -373,7 +379,7 @@ def teardown_factory(request):
 
     def finalizer():
         """
-        Delete the Pod
+        Delete the resources created in the test
         """
         for instance in instances:
             if not instance.is_deleted:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -101,7 +101,8 @@ def ceph_pool_factory(request):
             ).get(defaults.CEPHFILESYSTEM_NAME)
             ceph_pool_obj = OCS(**cfs)
         assert ceph_pool_obj, f"Failed to create {interface} pool"
-        instances.append(ceph_pool_obj)
+        if interface == constants.CEPHBLOCKPOOL:
+            instances.append(ceph_pool_obj)
         return ceph_pool_obj
 
     def finalizer():
@@ -257,6 +258,7 @@ def pvc_factory(
         if custom_data:
             pvc_obj = PVC(**custom_data)
             pvc_obj.create(do_reload=False)
+            instances.append(pvc_obj)
         else:
             project = project or project_factory()
             storageclass = storageclass or storageclass_factory(interface)
@@ -269,12 +271,12 @@ def pvc_factory(
                 wait=False
             )
             assert pvc_obj, "Failed to create PVC"
+            instances.append(pvc_obj)
+            pvc_obj.storageclass = storageclass
+            pvc_obj.project = project
         if status:
             helpers.wait_for_resource_state(pvc_obj, status)
-        pvc_obj.storageclass = storageclass
-        pvc_obj.project = project
 
-        instances.append(pvc_obj)
         return pvc_obj
 
     def finalizer():
@@ -338,6 +340,36 @@ def pod_factory(request, pvc_factory):
 
         instances.append(pod_obj)
         return pod_obj
+
+    def finalizer():
+        """
+        Delete the Pod
+        """
+        for instance in instances:
+            if not instance.is_deleted:
+                instance.delete()
+                instance.ocp.wait_for_delete(
+                    instance.name
+                )
+
+    request.addfinalizer(finalizer)
+    return factory
+
+
+@pytest.fixture()
+def teardown_factory(request):
+    """
+    Tearing down a resource that was created during the test
+    """
+    instances = []
+
+    def factory(resource_obj):
+        """
+        Args:
+            resource_obj (OCS object): Object to teardown after the test
+
+        """
+        instances.append(resource_obj)
 
     def finalizer():
         """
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -269,12 +271,12 @@ def factory(
                 wait=False
             )
             assert pvc_obj, "Failed to create PVC"
+            instances.append(pvc_obj)
+            pvc_obj.storageclass = storageclass
```
fbalak: I would keep this on its original place so it can create reference if object is provided in parameter. It provides consistency for calling the factory. Also if we want to remove this reference for custom_data it should be done in all fixtures.

tests/conftest.py
```diff
@@ -101,7 +101,8 @@ def factory(interface=constants.CEPHBLOCKPOOL):
             ).get(defaults.CEPHFILESYSTEM_NAME)
             ceph_pool_obj = OCS(**cfs)
         assert ceph_pool_obj, f"Failed to create {interface} pool"
-        instances.append(ceph_pool_obj)
+        if interface == constants.CEPHBLOCKPOOL:
```
fbalak: Why is this change needed?

RazTamir: We want to delete the pool only if that's cephblockpool.
cephfilesystem shouldn't be removed

tests/conftest.py
RazTamir: I'm not sure it ever worked like that cause if you providing custom_data, you expect that all information would be there and therefore, you don't have storageclass initialized or project.

Is that correct?

RazTamir:
Verified:

tests/e2e/test_pvc_creation_performance.py::TestPVCCreationPerformance::test_pvc_creation_measurement_performance[CephFS] 
-------------------------------------------------------------------------------------- live log setup ---------------------------------------------------------------------------------------
18:47:47 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig get Pod --selector=app=rook-ceph-tools -o yaml
18:47:48 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig rsh rook-ceph-tools-5f5dc75fd5-ht2k9 ceph auth get-key client.admin --format json-pretty
18:47:51 - MainThread - ocs_ci.ocs.resources.ocs - INFO - Adding Secret with name secret-test-cephfs-1418475146
18:47:51 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig create -f /tmp/Secretzp3cgfzt -o yaml
18:47:52 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig get CephFileSystem ocsci-cephfs -o yaml
18:47:54 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig get CephFileSystem  -o yaml
18:47:55 - MainThread - ocs_ci.ocs.resources.ocs - INFO - Adding StorageClass with name storageclass-test-cephfs-1418475577
18:47:55 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig create -f /tmp/StorageClassbwf4jj67 -o yaml
--------------------------------------------------------------------------------------- live log call ---------------------------------------------------------------------------------------
18:47:56 - MainThread - tests.e2e.test_pvc_creation_performance - INFO - Start creating new PVC
18:47:56 - MainThread - ocs_ci.ocs.resources.ocs - INFO - Adding PersistentVolumeClaim with name pvc-test-1418475688
18:47:56 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig create -f /tmp/PersistentVolumeClaim73zgb1mu -o yaml
18:47:58 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig get Pod --selector=app=csi-rbdplugin-provisioner -o yaml
18:47:59 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig logs csi-rbdplugin-provisioner-0 -c csi-provisioner
FAILED
------------------------------------------------------------------------------------- live log teardown -------------------------------------------------------------------------------------
18:48:03 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig delete PersistentVolumeClaim pvc-test-1418475688
18:48:07 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig get PersistentVolumeClaim pvc-test-1418475688 -o yaml
18:48:08 - MainThread - ocs_ci.ocs.ocp - INFO - PersistentVolumeClaim pvc-test-1418475688 got deleted successfully
18:48:08 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig delete StorageClass storageclass-test-cephfs-1418475577
18:48:10 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig get StorageClass storageclass-test-cephfs-1418475577 -o yaml
18:48:12 - MainThread - ocs_ci.ocs.ocp - INFO - StorageClass storageclass-test-cephfs-1418475577 got deleted successfully
18:48:12 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig delete Secret secret-test-cephfs-1418475146
18:48:13 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/ratamir/ocs_cluster1/auth/kubeconfig get Secret secret-test-cephfs-1418475146 -o yaml


You can see here the creation of PVC in the test call and deletion in teardown call

clacroix12:
This might be a silly question but why does this need to be a fixture? We're basically creating a `delete_resource()` function here, right?

RazTamir:
> This might be a silly question but why does this need to be a fixture? We're basically creating a `delete_resource()` function here, right?

Right but ... :)
the creation of that resource (PVC in this case) is part of the test, and the deletion should be done in the teardown

clacroix12:
> the creation of that resource (PVC in this case) is part of the test, and the deletion should be done in the teardown

Isn't that exactly what the factory fixture is for? What's stopping you from using the pvc_factory in this use case. You can create the pvc in the test using the factory and the finalizer will handle cleanup of the resource in 'teardown.' Not sure what use-case this teardown factory is trying to cover.

RazTamir:
The factory fixtures we currently have are not covering all resources we can create (daemon sets, deployments, monitoring resources, etc.) and that teardown fixture comes to help us.
In addition, it can provide great flexibility to keep using the helpers.py functions and still get the benefit of having the teardown clean up everything in the teardown

clacroix12:
@RazTamir That makes sense, thanks! I think if we have a certain resource that we're doing this with a lot it's probably a good sign that the resource needs a fixture of it's own, but allowing resources we don't have fixtures for yet benefit from being cleaned up in the teardown is nice.

tests/conftest.py
```diff
@@ -354,6 +356,36 @@ def finalizer():
     return factory
 
 
+@pytest.fixture()
+def teardown_factory(request):
+    """
+    Tearing down a resource that was created during the test
```
vasukulkarni: Can we add more docstring here on where/how to use this? I am not clear on the usecase

tests/conftest.py
```diff
@@ -354,6 +356,36 @@ def finalizer():
...
+        """
+        Args:
+            resource_obj (OCS object): Object to teardown after the test
+
+        """
+        instances.append(resource_obj)
+
+    def finalizer():
+        """
+        Delete the Pod
```
vasukulkarni: If I understand you are talking about generic cleanup of uncleaned objects? not just Pod right?

tests/conftest.py
vasukulkarni: @RazTamir this should be a different PR in itself, its like add on additional fix along with teardown factory PR :)

tests/conftest.py
RazTamir: done.

Please let me know if I need to add more

tests/conftest.py
RazTamir: Thanks!

tests/conftest.py
fbalak: You can have reference to name of some object in custom_data. You may want to keep the reference to object with that name also in object using it. However, it's not mandatory. If you don't provide parameter with the reference then the reference to object is empty.

If you want to change this functionality then I propose to do it in different PR as this PR is about a new fixture. Also it would need to be changed like this in all fixtures.

tests/conftest.py
RazTamir: fixed in https://github.com/red-hat-storage/ocs-ci/pull/572

tests/conftest.py
RazTamir: ok reverting this change and let's discuss this in new PR

tests/e2e/test_pvc_creation_performance.py
```diff
@@ -18,26 +18,30 @@ class TestPVCCreationPerformance(E2ETest):
     pvc_size = '1Gi'
 
     @pytest.fixture()
-    def base_setup(self, request, storageclass_factory):
+    def base_setup(
+        self, request, interface_iterate, storageclass_factory
```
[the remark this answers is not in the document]
RazTamir: fixed

tests/conftest.py
```diff
@@ -354,6 +354,43 @@ def finalizer():
     return factory
 
 
+@pytest.fixture()
+def teardown_factory(request):
+    """
+    Tearing down a resource that was created during the test
+    To use this factory, you'll need to pass 'teardown_factory' to you test
```
vasukulkarni: thanks for the docstring, one minor - s/you/your and one more instance below.  When you push new changes, I will use previous approval and merge it.

RazTamir: done

PR #569, opened 2019-08-14
Jenkinsfile: Send email and change node label

zmc:


Jenkinsfile
```diff
@@ -22,6 +23,10 @@ pipeline {
     stage("Setup") {
       steps {
         sh """
+          if [ ! -z '${env.EMAIL}' ]; then
+            sudo yum install -y /usr/sbin/postfix
```
petr-balogh: Don't you have pre-prepared jenkins slaves where you should have installed all the dependency needed for running jobs? I think that this shouldn't be part of each run if you have some static jenkins slaves but should be part of some ansible playbook which you are using for setup slave or part of image if you are using containerized slave, same for line bellow.

Jenkinsfile
```diff
@@ -76,10 +81,16 @@ pipeline {
       }
     }
     stage("Tier 1") {
+      environment {
```
petr-balogh: You can define variable above pipeline definition so above line 16 like:
`def email_arg = ""`

and here you can use:
```groovy
script {
  if(params.EMAIL != '') {
    email_arg = "--email=${params.EMAIL}"
  }
}
```

then on line 93 you can directly use variable from groovy:
`${email_arg}`

Jenkinsfile
zmc: No - we use vanilla images, so there's no ansible playbook at all.

Jenkinsfile
zmc: Thanks for the tip!

zmc:
@clacroix12 mind taking a look?

PR #572, opened 2019-08-15
Fix ceph_pool_factory fixture

RazTamir:
With this fix, the cephfilesystem pool will not be deleted

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -101,7 +101,7 @@ def ceph_pool_factory(request):
             ).get(defaults.CEPHFILESYSTEM_NAME)
             ceph_pool_obj = OCS(**cfs)
         assert ceph_pool_obj, f"Failed to create {interface} pool"
-        if interface == constants.CEPHBLOCKPOOL:
+        if interface != constants.CEPHFILESYSTEM:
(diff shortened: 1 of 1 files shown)


RazTamir:
Verified as part of https://github.com/red-hat-storage/ocs-ci/pull/565

tests/conftest.py
```diff
@@ -101,7 +101,8 @@ def factory(interface=constants.CEPHBLOCKPOOL):
             ).get(defaults.CEPHFILESYSTEM_NAME)
             ceph_pool_obj = OCS(**cfs)
         assert ceph_pool_obj, f"Failed to create {interface} pool"
-        instances.append(ceph_pool_obj)
+        if interface == constants.CEPHBLOCKPOOL:
```
dahorak: Just cosmetic proposal, if we want to avoid deleting cephfilesystem, I would maybe use "negative" condition:
```suggestion
        if interface != constants.CEPHFILESYSTEM:
```
(If I understood the code correctly, it should have basically the same effect, but from my POV it will be slightly more readable, WDYT?)

RazTamir: ack - fixing now

PR #566, opened 2019-08-14
Separate wait for resource from resource creation

ebenahar:
In order to be able to delete a resource in the teardown, we should get the resource object, even if the wait for this resource has ended with a  failure.

PR #574, opened 2019-08-15
Performance testing: Creation of 120 PVCs in 60 seconds

RazTamir:
https://jira.coreos.com/browse/KNIP-627 item #3

### DIFF
--- a/ocs_ci/ocs/exceptions.py
+++ b/ocs_ci/ocs/exceptions.py
@@ -46,3 +46,7 @@ class DeploymentPlatformNotSupported(Exception):
 
 class UnavailableBuildException(Exception):
     pass
+
+
+class PerformanceException(Exception):
+    pass
--- a/tests/e2e/test_pvc_creation_performance.py
+++ b/tests/e2e/test_pvc_creation_performance.py
@@ -3,11 +3,10 @@ Test to verify PVC creation performance
 """
 import logging
 import pytest
+import ocs_ci.ocs.exceptions as ex
 
 from ocs_ci.framework.testlib import tier1, E2ETest, polarion_id, bugzilla
-from tests.helpers import (
-    create_pvc, measure_pvc_creation_time, create_multiple_pvcs,
-    wait_for_resource_state)
+from tests import helpers
 from ocs_ci.ocs import defaults, constants
 
 log = logging.getLogger(__name__)
@@ -44,11 +43,15 @@ class TestPVCCreationPerformance(E2ETest):
         """
         log.info('Start creating new PVC')
 
-        pvc_obj = create_pvc(sc_name=self.sc_obj.name, size=self.pvc_size)
+        pvc_obj = helpers.create_pvc(
+            sc_name=self.sc_obj.name, size=self.pvc_size
+        )
         teardown_factory(pvc_obj)
-        create_time = measure_pvc_creation_time(self.interface, pvc_obj.name)
+        create_time = helpers.measure_pvc_creation_time(
+            self.interface, pvc_obj.name
+        )
         if create_time > 1:
-            raise AssertionError(
+            raise ex.PerformanceException(
                 f"PVC creation time is {create_time} and greater than 1 second"
             )
         logging.info("PVC creation took less than a 1 second")
@@ -65,7 +68,7 @@ class TestPVCCreationPerformance(E2ETest):
         number_of_pvcs = 120
         log.info('Start creating new 120 PVCs')
 
-        pvc_objs = create_multiple_pvcs(
+        pvc_objs = helpers.create_multiple_pvcs(
             sc_name=self.sc_obj.name,
             namespace=defaults.ROOK_CLUSTER_NAMESPACE,
             number_of_pvc=number_of_pvcs,
@@ -75,18 +78,17 @@ class TestPVCCreationPerformance(E2ETest):
         for pvc_obj in pvc_objs:
             teardown_factory(pvc_obj)
         for pvc_obj in pvc_objs:
-            wait_for_resource_state(pvc_obj, constants.STATUS_BOUND)
-        start_time = measure_pvc_creation_time(
-            self.interface, pvc_objs[0].name, return_start_time=True
+            helpers.wait_for_resource_state(pvc_obj, constants.STATUS_BOUND)
+        start_time = helpers.get_start_creation_time(
+            self.interface, pvc_objs[0].name
         )
-        end_time = measure_pvc_creation_time(
+        end_time = helpers.get_end_creation_time(
             self.interface, pvc_objs[number_of_pvcs - 1].name,
-            return_end_time=True
         )
         total = end_time - start_time
         total_time = total.total_seconds()
         if total_time > 60:
-            raise AssertionError(
+            raise ex.PerformanceException(
                 f"{number_of_pvcs} PVCs creation time is {total_time} and "
                 f"greater than 60 seconds"
             )
--- a/tests/e2e/test_pvc_creation_performance.py
+++ b/tests/e2e/test_pvc_creation_performance.py
@@ -5,7 +5,10 @@ import logging
 import pytest
 
 from ocs_ci.framework.testlib import tier1, E2ETest, polarion_id, bugzilla
-from tests.helpers import create_pvc, measure_pvc_creation_time
+from tests.helpers import (
+    create_pvc, measure_pvc_creation_time, create_multiple_pvcs,
+    wait_for_resource_state)
+from ocs_ci.ocs import defaults, constants
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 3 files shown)


RazTamir:
Verified:

_______________________________________________________ TestPVCCreationPerformance.test_multiple_pvc_creation_measurement_performance _______________________________________________________

self = <tests.e2e.test_pvc_creation_performance.TestPVCCreationPerformance object at 0x7fc421e92cc0>

    @polarion_id('OCS-1620')
    # @bugzilla('1740139')
    def test_multiple_pvc_creation_measurement_performance(self):
        """
        Measuring PVC creation time of 120 PVCs in 60 seconds
        """
        number_of_pvcs = 120
        log.info('Start creating new 120 PVCs')
    
        pvc_objs = create_multiple_pvcs(
            sc_name=self.sc_obj.name,
            namespace=defaults.ROOK_CLUSTER_NAMESPACE,
            number_of_pvc=number_of_pvcs,
            size=self.pvc_size,
            wait=False
        )
        for pvc_obj in pvc_objs:
            wait_for_resource_state(pvc_obj, constants.STATUS_BOUND)
        start_time = measure_pvc_creation_time(
            'CephBlockPool', pvc_objs[0].name, return_start_time=True
        )
        end_time = measure_pvc_creation_time(
            'CephBlockPool', pvc_objs[number_of_pvcs - 1].name,
            return_end_time=True
        )
        total = end_time - start_time
        total_time = total.total_seconds()
        if total_time > 60:
            raise AssertionError(
>               f"{number_of_pvcs} PVCs creation time is {total_time} and "
                f"greater than 60 seconds"
            )
E           AssertionError: 120 PVCs creation time is 108.804983 and greater than 60 seconds

tests/e2e/test_pvc_creation_performance.py:81: AssertionError

Will submit a bug and add Bugzilla decorator

tests/e2e/test_pvc_creation_performance.py
```diff
@@ -44,3 +47,40 @@ def test_pvc_creation_measurement_performance(self):
...
+            number_of_pvc=number_of_pvcs,
+            size=self.pvc_size,
+            wait=False
+        )
+        for pvc_obj in pvc_objs:
+            wait_for_resource_state(pvc_obj, constants.STATUS_BOUND)
+        start_time = measure_pvc_creation_time(
+            'CephBlockPool', pvc_objs[0].name, return_start_time=True
+        )
+        end_time = measure_pvc_creation_time(
```
vasukulkarni: Do you want to aggregate the total time taken by all PVC's instead? There is a possibility that last PVC was created > 60 seconds?

RazTamir: If I understand you correctly so not really

tests/e2e/test_pvc_creation_performance.py
```diff
@@ -43,9 +46,50 @@ def test_pvc_creation_measurement_performance(self, teardown_factory):
...
         if create_time > 1:
             raise AssertionError(
                 f"PVC creation time is {create_time} and greater than 1 second"
             )
         logging.info("PVC creation took less than a 1 second")
+
+    @pytest.mark.usefixtures(base_setup.__name__)
+    @polarion_id('OCS-1620')
+    @bugzilla('1741612')
+    def test_multiple_pvc_creation_measurement_performance(
```
vasukulkarni: needs tier1 or 2 ?

tests/e2e/test_pvc_creation_performance.py
vasukulkarni: Resolving this based on JIRA ticket

vasukulkarni:
@ramkiperiy FYI

tests/e2e/test_pvc_creation_performance.py
RazTamir: the class marked with tier 1

tests/e2e/test_pvc_creation_performance.py
```diff
@@ -43,9 +46,50 @@ def test_pvc_creation_measurement_performance(self, teardown_factory):
...
+        start_time = measure_pvc_creation_time(
+            self.interface, pvc_objs[0].name, return_start_time=True
+        )
+        end_time = measure_pvc_creation_time(
+            self.interface, pvc_objs[number_of_pvcs - 1].name,
+            return_end_time=True
+        )
+        total = end_time - start_time
+        total_time = total.total_seconds()
+        if total_time > 60:
```
clacroix12: I think an explicit assertion would be clearer here, but otherwise LGTM.
```python
assert total_time <= 60, f"{number_of_pvcs} PVCs creation time is ..."
```

tests/e2e/test_pvc_creation_performance.py
```diff
@@ -43,9 +46,50 @@ def test_pvc_creation_measurement_performance(self, teardown_factory):
...
+        pvc_objs = create_multiple_pvcs(
+            sc_name=self.sc_obj.name,
+            namespace=defaults.ROOK_CLUSTER_NAMESPACE,
+            number_of_pvc=number_of_pvcs,
+            size=self.pvc_size,
+            wait=False
+        )
+        for pvc_obj in pvc_objs:
+            teardown_factory(pvc_obj)
+        for pvc_obj in pvc_objs:
```
keesturam: If the PVCs don't get bound for some reason, this test case is going to take 120 * 60 seconds(2 hours!) before marking it as failed. we should arrive at a better way to check for bound status for bulk creations.

tests/helpers.py
```diff
@@ -734,13 +734,20 @@ def get_worker_nodes():
...
+    interface, pvc_name, return_start_time=False, return_end_time=False
+):
     """
     Measure PVC creation time based on logs
 
     Args:
         interface (str): The interface backed the PVC
         pvc_name (str): Name of the PVC for creation time measurement
+        *** The defaults below should be used only from dedicated functions ***
+        *** to get the start time and end time of PVC creation ***
```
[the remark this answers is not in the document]
RazTamir: Fixed

tests/e2e/test_pvc_creation_performance.py
RazTamir: I don't have a smart solution for it right now.
That's a limitation

ramkiperiy: May be using threads would help to reduce the time.

RazTamir: not really,
If I'm creating 120 PVCs without waiting for each one, I can assume that when finishing the creation of the last PVC, many of the first PVCs already bounded so the "wait" will return immediately.
When reaching to the first PVC the is still Pending, I will wait for the Bound status and in that same time, few more PVCs will reach Bound state.

So threads are not really helpful here cause the operation is not in sequence

ocs_ci/ocs/exceptions.py
```diff
@@ -42,3 +42,7 @@ class MDSCountException(Exception):
 
 class DeploymentPlatformNotSupported(Exception):
     pass
+
+
+class PerformanceException(Exception):
```
keesturam: This is nice :+1:

tests/e2e/test_pvc_creation_performance.py
```diff
@@ -41,11 +43,55 @@ def test_pvc_creation_measurement_performance(self, teardown_factory):
...
+            helpers.wait_for_resource_state(pvc_obj, constants.STATUS_BOUND)
```
keesturam: how about an assert here? it will fail on the first attempt. Any pvc creation that hasn't bound for more than 60 seconds should be considered a failure?

tests/helpers.py
```diff
@@ -760,13 +760,51 @@ def measure_pvc_creation_time(interface, pvc_name):
...
+    else:
+        pod_name = pod.get_cephfs_provisioner_pod().name
+
+    # get the logs from the csi-provisioner container
+    logs = pod.get_pod_logs(pod_name, 'csi-provisioner')
+    logs = logs.split("\n")
+    # Extract the starting time for the PVC provisioning
     end = [
-        i for i in logs if re.search(f"provision.*{pvc_name}.*succeeded", i)
+        i for i in logs if re.search(f"provision.*{pvc_name}.*started", i)
```
keesturam: This should be 'succeeded', right?

tests/helpers.py
```diff
@@ -734,16 +734,16 @@ def get_worker_nodes():
...
-    Measure PVC creation time based on logs
+    Get the starting creation time of a PVC based on provisioner logs
 
     Args:
         interface (str): The interface backed the PVC
         pvc_name (str): Name of the PVC for creation time measurement
 
     Returns:
-        float: Creation time for the PVC
+        datetime object: Creation time for the PVC
```
keesturam: datetime object: Start time of PVC creation

tests/helpers.py
```diff
@@ -760,13 +760,51 @@ def measure_pvc_creation_time(interface, pvc_name):
...
+def get_end_creation_time(interface, pvc_name):
+    """
+    Get the ending creation time of a PVC based on provisioner logs
+
+    Args:
+        interface (str): The interface backed the PVC
+        pvc_name (str): Name of the PVC for creation time measurement
+
+    Returns:
+        datetime object: Creation time for the PVC
```
keesturam: datetime object: end time for the PVC creation

tests/e2e/test_pvc_creation_performance.py
RazTamir: See https://github.com/red-hat-storage/ocs-ci/pull/566.
It will ensure to raise an exception instead of returning True/False

tests/helpers.py
RazTamir: Thanks!
Fixing

tests/helpers.py
RazTamir: done

tests/helpers.py
RazTamir: done

tests/e2e/test_pvc_creation_performance.py
```diff
@@ -41,11 +43,55 @@ def test_pvc_creation_measurement_performance(self, teardown_factory):
...
+        start_time = helpers.get_start_creation_time(
+            self.interface, pvc_objs[0].name
+        )
+        end_time = helpers.get_end_creation_time(
+            self.interface, pvc_objs[number_of_pvcs - 1].name,
```
nehaberry: @RazTamir is this sequential creation where we wait for one PVC to be BOUND before sending 2nd request.

If not, then, in scenarios where multiple pvc creations are sent in parallel, the order of them getting BOUND may vary. i.e. PVC97 may get BOUND before PVC95 and so on. So end_time may not always be the time when the 120th PVC got BOUND here. 

Just an observation. Please correct me if I am wrong.

RazTamir: You are right @nehaberry .
For now, this test is blocked on BZ - we'll need to figure out how to handle it as soon it is fixed

PR #585, opened 2019-08-17
Rook repo download

petr-balogh:
Proposing method to clone rook repository for consuming yaml files directly. This will allow us keep with all changes quite easier if we will convert everything to this new approach.

I just updated for example usage of RBD and CEPHFS CSI storage classes as I found occurrences of them.

Fixes: #596

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,7 @@ config.yaml
 .vscode
 *.bak
 openshift-install*
-data/pull-secret
 bin/
 dist
 .coverage
+data/*
--- a/ocs_ci/utility/templating.py
+++ b/ocs_ci/utility/templating.py
@@ -134,9 +134,48 @@ def dump_to_temp_yaml(src_file, dst_file, **kwargs):
         yaml.dump(data, yaml_file)
 
 
-def load_yaml_to_dict(file):
+def load_yaml_to_dict(file, multi_document=False):
+    """
+    Load yaml file to the dictionary
+
+    Args:
+        file (str): Path to yaml file to load
+        multi_document (bool): True if yaml contains more documents
+
+    Returns:
+        dict: If multi_document == False, returns loaded data from yaml file
+            with one document.
+        generator: If multi_document == True, returns generator which each
+            iteration returns dict from one loaded document from a file.
+
+    """
     template = os.path.join(file)
-    return yaml.safe_load(open(template, 'r'))
+    if not multi_document:
+        return yaml.safe_load(open(template, 'r'))
+    else:
+        return yaml.safe_load_all(open(template, 'r'))
+
+
+def get_n_document_from_yaml(yaml_generator, index=0):
+    """
+    Returns n document from yaml generator loaded by load_yaml_to_dict with
+    multi_document = True.
+
+    Args:
+        yaml_generator (generator): Generator from yaml.safe_load_all
+        index (int): Index of document to return. (0 - 1st, 1 - 2nd document)
+
+    Returns:
+        dict: Data from n document from yaml file.
+
+    Raises:
+        IndexError: In case that yaml generator doesn't have such index.
+
+    """
+    for idx, document in enumerate(yaml_generator):
+        if index == idx:
+            return document
+    raise IndexError(f"Passed yaml generator doesn't have index {index}")
 
 
 def dump_dict_to_temp_yaml(data, temp_yaml):
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -1137,3 +1137,31 @@ def ceph_health_check(namespace=None):
         raise CephHealthException(
             f"Ceph cluster health is not OK. Health: {health}"
         )
+
+
+def get_rook_repo(branch='master', to_checkout=None):
+    """
+    Clone and checkout the rook repository to specific branch/commit.
+
+    Args:
+        branch (str): Branch name to checkout
+        to_checkout (str): Commit id or tag to checkout
+
+    """
+    cwd = constants.ROOK_REPO_DIR
+    if not os.path.isdir(cwd):
+        log.info(f"Cloning rook repository into {cwd}.")
+        run_cmd(f"git clone {constants.ROOK_REPOSITORY} {cwd}")
+    else:
+        log.info(
+            f"The rook directory {cwd} already exists, ocs-ci will skip the "
+            f"clone of rook repository."
+        )
+        log.info("Fetching latest changes from rook repository.")
+        run_cmd("git fetch --all", cwd=cwd)
+    log.info(f"Checkout rook repository to specific branch: {branch}")
+    run_cmd(f"git checkout {branch}", cwd=cwd)
+    log.info(f"Reset branch: {branch} with latet changes")
+    run_cmd(f"git reset --hard origin/{branch}", cwd=cwd)
+    if to_checkout:
+        run_cmd(f"git checkout {to_checkout}", cwd=cwd)
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -24,6 +24,10 @@ RUN:
   client_version: '4.2.0-0.nightly'
   bin_dir: './bin'
   google_api_secret: '~/.ocs-ci/google_api_secret.json'
+  rook_branch: "master"
+  # We can also specify the tag or specific commit id to checkout by changin
+  # following parameter in custom config file:
+  # rook_to_checkout: "commit_id or tag_name"
 
 # In this section we are storing all deployment related configuration but not
 # the environment related data as those are defined in ENV_DATA section.
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -37,6 +37,14 @@ TEMPLATE_DEPLOYMENT_EO = os.path.join(
 TEMPLATE_DEPLOYMENT_CLO = os.path.join(
     TEMPLATE_DEPLOYMENT_LOGGING, "clusterlogging_operator"
 )
+DATA_DIR = os.path.join(TOP_DIR, 'data')
+ROOK_REPO_DIR = os.path.join(DATA_DIR, 'rook')
+ROOK_EXAMPLES_DIR = os.path.join(
+    ROOK_REPO_DIR, "cluster", "examples", "kubernetes", "ceph"
+)
+ROOK_CSI_RBD_DIR = os.path.join(ROOK_EXAMPLES_DIR, "csi", "rbd")
+ROOK_CSI_CEPHFS_DIR = os.path.join(ROOK_EXAMPLES_DIR, "csi", "cephfs")
+
 
 # Statuses
 STATUS_PENDING = 'Pending'
@@ -72,6 +80,7 @@ SECRET = "Secret"
 NAMESPACE = 'Namespace'
 IGNORE_SC_GP2 = "gp2"
 IGNORE_SC_FLEX = "rook-ceph-block"
+ROOK_REPOSITORY = "https://github.com/rook/rook.git"
 
 
 DEFAULT_BLOCKPOOL = 'rbd'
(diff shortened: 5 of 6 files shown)


vasukulkarni:
👍to this method!

ocs_ci/utility/utils.py
```diff
@@ -1137,3 +1137,27 @@ def ceph_health_check(namespace=None):
...
+
+def get_rook_repo(to_checkout):
+    """
+    Clone and checkout the rook repository to specific branch/commit.
+
+    Args:
+        to_checkout (str): Commit id or branch to checkout on
+
+    """
+    fresh_clone = False
```
vasukulkarni: This can cause stale issues, Rather age of dir can be used to decided freshness.

tests/manage/storageclass/test_create_storageclass_with_same_name.py
```diff
@@ -51,9 +52,27 @@ def create_storageclass(sc_name, expect_fail=False):
...
+        constants.ROOK_CSI_RBD_STORAGECLASS_YAML, multi_document=True
+    )
+    sc_data = templating.get_n_document_from_yaml(rook_example_sc_data, 1)
+    namespace = config.ENV_DATA['cluster_namespace']
+    sc_data['provisioner'] = f'{namespace}.rbd.csi.ceph.com'
     sc_data['metadata']['name'] = sc_name
-    sc_data['parameters']['clusterID'] = defaults.ROOK_CLUSTER_NAMESPACE
+    sc_data['parameters']['clusterID'] = namespace
+    sc_data['metadata']['pool'] = 'rbd'
+    sc_data['parameters'][
```
vasukulkarni: why is below assignment needed? doesn't it come from dict?

ocs_ci/utility/utils.py
petr-balogh: Not sure which stale issues you mean @vasukulkarni ? Can you please elaborate?
Basically if the directory data/rook doesn't exist it means that we don't have the rook repo cloned yet and then we clone it on line 1155.  As we just cloned the repository there is no need to do git fetch --all on line 1159 cause we just cloned the repository so we can directly checkout where we need.

The next run we won't clone the repository and we will just git fetch --all to make sure we have all recent changes fetched so we should be up to date every time I think.

tests/manage/storageclass/test_create_storageclass_with_same_name.py
petr-balogh: If I compared what we currently have in our template it's differ from what is currently in the template from rook. You can see diff here:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/templates/CSI/rbd/storageclass.yaml#L22 . we have csi-rbd-secret . 

But in rook example they have:
https://github.com/rook/rook/blob/master/cluster/examples/kubernetes/ceph/csi/rbd/storageclass.yaml#L32
rook-ceph-csi .

As I don't want to break anything by this PR and just want to make sure that everything is as before we need to have this change here to keep it as it was.

.gitignore
```diff
@@ -16,3 +16,4 @@ data/pull-secret
 bin/
 dist
 .coverage
+data/*
```
dahorak: With this addition, we can probably remove `data/pull-secret` (line 15)... just cosmetic change though...

ocs_ci/utility/utils.py
```diff
@@ -1137,3 +1137,27 @@ def ceph_health_check(namespace=None):
...
+    cwd = constants.ROOK_REPO_DIR
+    if not os.path.isdir(constants.ROOK_REPO_DIR):
```
dahorak: What is the reason for using `constants.ROOK_REPO_DIR` instead of `cwd`? Wouldn't be `cwd` slightly more clear since it is assigned just line above?

ocs_ci/utility/utils.py
```diff
@@ -1137,3 +1137,27 @@ def ceph_health_check(namespace=None):
...
+        log.info("Cloning rook repository.")
```
dahorak: I would maybe include the `cwd` (`ROOK_REPO_DIR`) into the log message.

ocs_ci/utility/utils.py
vasukulkarni: @petr-balogh I meant instead of using flag we can use datetime.timedelta to figure out if the dir created is older than 7 days and if so pull the rook files again.

tests/manage/storageclass/test_create_storageclass_with_same_name.py
vasukulkarni: I think since the files are coming from rook repo, we should use them as is and if they fail its a bug?

ocs_ci/utility/templating.py
```diff
@@ -134,9 +134,44 @@ def dump_to_temp_yaml(src_file, dst_file, **kwargs):
...
+    else:
+        return yaml.safe_load_all(open(template, 'r'))
+
+
+def get_n_document_from_yaml(yaml_generator, index=0):
+    """
+    Returns n document from yaml generator loaded by load_yaml_to_dict with
+    multi_document = True.
+
+    Args:
```
clacroix12: Need to add an entry for `yaml_generator`

ocs_ci/utility/utils.py
clacroix12: Isn't `fresh_clone` just determining whether we `clone` or `fetch` the latest before checking out the branch/commit? In either case we should be getting the latest files so I don't think there's an issue with stale files here. However, is `fresh_clone` even necessary? Seems we can get by with a simple if/else:

```python
if not os.path.isdir(constants.ROOK_REPO_DIR):
    # clone
else:
    # fetch
# checkout
```

ocs_ci/utility/templating.py
```diff
@@ -134,9 +134,44 @@ def dump_to_temp_yaml(src_file, dst_file, **kwargs):
...
+        index (int): Index of document to return. (0 - 1st, 1 - 2nd document)
+    Returns:
+        dict: Data from n document from yaml file.
+
+    """
+    x = 0
+    for document in yaml_generator:
+        if index == x:
+            return document
+        x += 1
```
clacroix12: We can simplify a bit if we use `enumerate`.

```python
for idx, document in enumerate(yaml_generator):
    if index == idx:
        return document
```

.gitignore
petr-balogh: done

ocs_ci/utility/templating.py
petr-balogh: done

ocs_ci/utility/templating.py
petr-balogh: done

ocs_ci/utility/utils.py
petr-balogh: Actually I realized that if we will have already cloned repo and will just git fetch --all we won't apply the latest changes, so I think that if we will reset --hard to specific branch will be to safest solution.  Also separated this to_commit just for commit id or tag so we can then checkout to the tag or commit once we have fetched latest changes.

ocs_ci/utility/utils.py
petr-balogh: done, forgot to change it here, thanks

ocs_ci/utility/utils.py
petr-balogh: I didn't put it here cause run_cmd is writing the command which is running to the logger. But added it here as well if you find it useful.

tests/manage/storageclass/test_create_storageclass_with_same_name.py
petr-balogh: Not really, we cannot rely only on the files which are on examples. We have more complex tests which needs to change a lot of parameters there starting with the namespace as in rook repo they use rook-ceph and we use openshift-storage to more complex parameters which needs to be set for test needs.  You can see the example of this helper method: https://github.com/red-hat-storage/ocs-ci/blob/master/tests/helpers.py#L239 . This deal with a lot of parameters.

petr-balogh:
Fixing issue #585

ocs_ci/utility/utils.py
clacroix12: Would [git pull](https://git-scm.com/docs/git-pull) be more appropriate?

petr-balogh: If there will be some changes done by user from some reason I thin it can lead to some conflict. With reset --hard you will make sure you are reseted exactly to the same code of origin without need of deal with any conflict or so.

tests/manage/storageclass/test_create_storageclass_with_same_name.py
vasukulkarni: Lets take one assignment example: 
```
  sc_data['parameters'][
        'csi.storage.k8s.io/provisioner-secret-namespace'
    ] = 'default'
```
Why is this needed? we are not assigning any dynamic value here just a static 'default' ,  Upstream repo should already have static values assigned properly?

ocs_ci/utility/utils.py
clacroix12: Oh I didn't realize that you were doing `git reset --hard origin/{branch}`. I was thinking just `git reset --hard` which would only reset you to your local branch's HEAD which would not update the files, but the way you have it works.

tests/manage/storageclass/test_create_storageclass_with_same_name.py
petr-balogh: As I wrote @vasukulkarni I am not author of this test and didn't want to change any behavior here. As I just compared the files already mentioned in comment above and changed values accordingly. 

>If I compared what we currently have in our template it's differ from what is currently in the template from rook. You can see diff here:
>https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/templates/CSI/rbd/storageclass.yaml#L22 . we have csi-rbd-secret .
>
>But in rook example they have:
>https://github.com/rook/rook/blob/master/cluster/examples/kubernetes/ceph/csi/rbd/storageclass.yaml#L32
rook-ceph-csi .

Just take a look in those files and differences. I don't really know what is the logic behind this but `csi.storage.k8s.io/provisioner-secret-namespace: rook-ceph` doesn't sounds good to me as we don't use `rook-ceph` namespace at all.

ocs_ci/utility/templating.py
```diff
@@ -134,9 +134,48 @@ def dump_to_temp_yaml(src_file, dst_file, **kwargs):
...
 
 
-def load_yaml_to_dict(file):
+def load_yaml_to_dict(file, multi_document=False):
+    """
+    Load yaml file to the dictionary
+
+    Args:
+        file (str): Path to yaml file to load
+        multi_document (boolean): True if yaml contains more documents
```
RazTamir: bool

ocs_ci/utility/templating.py
```diff
@@ -134,9 +134,48 @@ def dump_to_temp_yaml(src_file, dst_file, **kwargs):
...
 
 
-def load_yaml_to_dict(file):
+def load_yaml_to_dict(file, multi_document=False):
+    """
+    Load yaml file to the dictionary
+
+    Args:
+        file (str): Path to yaml file to load
+        multi_document (bool): True if yaml contains more documents
```
RazTamir: do you mean here that a single yaml contain few CRs inside it?
if yes, please add this to the description

tests/manage/storageclass/test_create_storageclass_with_same_name.py
vasukulkarni: @petr-balogh I think we can file an issue for this one and the owner of the test case can take care of it later,  It does sound like a bug if we are deviating from the standard templates.

petr-balogh:
Verification job running here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/472/

ocs_ci/utility/templating.py
petr-balogh: The description follows yaml conventions:
https://yaml.org/spec/1.2/spec.html

YAML uses three dashes (“---”) to separate directives from document content.

CR is openshift specific but here we are working with pure yaml.

tests/manage/storageclass/test_create_storageclass_with_same_name.py
petr-balogh: The problem is that the examples in rook repository are not official templates, those are just examples of the manifests and not exact files which needs or should be used.  User can change parameters, add new or do whatever they need, they just have provided files.  There are also real templates in examples here:  https://github.com/rook/rook/tree/master/cluster/examples/kubernetes/ceph/csi/template

which are real openshift templates, for those files I agree there shouldn't be any changes and should be created like they are and param should be passed through OC.

petr-balogh:
After yesterday discussion with @vasukulkarni on staff mtg I decided to just add the functionality of cloning rook repository and don't change usage of manifests yet in this PR.

Verification running here:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/495/

Issue #592, opened 2019-08-19
[CephFS] Volume provision fails with "InvalidValue: [Errno 22] error in setxattr"


sidhant-agrawal:
Multiple CephFS tests are failing with the following error:

`"InvalidValue: [Errno 22] error in setxattr"`

This is due to wrong data pool name in storageclass when using factory fixtures

PR #593, opened 2019-08-19
Fix for Issue #592 : "InvalidValue: [Errno 22] error in setxattr" error while creating CephFS based PVC

sidhant-agrawal:
Fixes: #592

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -155,6 +155,8 @@ def storageclass_factory(
             secret = secret or secret_factory(interface=interface)
             ceph_pool = ceph_pool_factory(interface)
             interface_name = ceph_pool.name
+            if interface == constants.CEPHFILESYSTEM:
+                interface_name += '-data0'
 
             sc_obj = helpers.create_storage_class(
                 interface_type=interface,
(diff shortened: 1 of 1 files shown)


sidhant-agrawal:
> @sidhant-agrawal I think it would be better if we don't workaround this bug

In fixtures.py [L124](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/fixtures.py#L124) we are using helpers function to get the CephFS data pool name dynamically, here in factory fixtures we are not getting the CephFS name instead of data pool name through ceph_pool_factory name, so need to append '-data0' in the cephfs name

vasukulkarni:
@sidhant-agrawal do you want to call the same function here as well helpers.get_cephfs_data_pool_name() for assignment?

vasukulkarni:
good find!

sidhant-agrawal:
@vasukulkarni 
`helpers.get_cephfs_data_pool_name()` seems more dynamic as it directly fetches name from the  Ceph.
For now, according to current convention appending '-data0' in the CephFS name also works.

clacroix12:
I agree with Vasu, I think we should be consistent and use `helpers.get_cephfs_data_pool_name()`. This also makes it easier to fix if appending `-data0` stops working for some reason, we can change the function logic to apply it everywhere as opposed to fixing multiple places in the code.

sidhant-agrawal:
Done.
Now using `interface_name = helpers.get_cephfs_data_pool_name()` for fetching CephFS data pool

PR #601, opened 2019-08-20
Fix PVC disruptive TCs

ebenahar:
Remove wait=False from create_pvc calls done with a TheadPoolExecutor


ebenahar:
[test_logs.zip](https://github.com/red-hat-storage/ocs-ci/files/3519963/test_logs.zip)

PR #603, opened 2019-08-20
update pvc_factory

fbalak:
Store reference for last project and storageclass. If there is not provided custom project or storageclass and there is already created project or storageclass from previous calling of factory function then it can be reused.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -227,6 +227,9 @@ def pvc_factory(
     PVC. For custom PVC provide 'storageclass' parameter.
     """
     instances = []
+    active_project = None
+    active_rbd_storageclass = None
+    active_cephfs_storageclass = None
 
     def factory(
         interface=constants.CEPHBLOCKPOOL,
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -263,8 +266,20 @@ def factory(
...
-            project = project or project_factory()
-            storageclass = storageclass or storageclass_factory(interface)
+            nonlocal active_project
+            nonlocal active_rbd_storageclass
+            nonlocal active_cephfs_storageclass
+
+            project = project or active_project or project_factory()
+            active_project = project
+            if interface == constants.CEPHBLOCKPOOL:
+                storageclass = storageclass or active_rbd_storageclass \
```
RazTamir:
```python
storageclass = (
    storageclass or active_rbd_storageclass or 
    storageclass_factory(interface)
)
```

Please fix below as well

tests/conftest.py
```diff
@@ -263,8 +266,24 @@ def factory(
             pvc_obj = PVC(**custom_data)
             pvc_obj.create(do_reload=False)
         else:
-            project = project or project_factory()
-            storageclass = storageclass or storageclass_factory(interface)
+            nonlocal active_project
+            nonlocal active_rbd_storageclass
+            nonlocal active_cephfs_storageclass
+
+            project = project or active_project or project_factory()
```
petr-balogh: Cannot we squash this and line below into one line? active_project = project or active_project or project_factory()
And then use just active project later on? I think only one reference on line to 291 project.namespace.

fbalak: Right, good point

Issue #591, opened 2019-08-19
helpers.wait_for_resource_state() is broken for PV resources


sidhant-agrawal:
helpers.wait_for_resource_state() is broken for PV resources.
wait_for_resource() is not able to fetch and check the correct status of PV resources

mbukatov:
It's possible that this is either related to directly causing failure of OCS-530 test case.

mbukatov:
The root cause seems to be in commit 2753e2cebaa5fe1738065d37f889639f2c046f57 introduced by https://github.com/red-hat-storage/ocs-ci/pull/573, which creates new method `get_resource_status` which instead of parsing yaml, tries to parse human readable command line output.

When I add some logging into the new code there like this:

```
# Only 1 resource expected to be returned
if resource_name:
    logging.info("waiting for a single resource")
    status = self.get_resource_status(resource_name)
    if status == condition:
        return True
    logging.warning("status of %s was %s, waiting for %s", resource_name, status, condition)
```

And run test case `tests/manage/pv_services/test_dynamic_pvc_accessmodes_with_reclaim_policies.py::TestRWODynamicPvc::test_rwo_dynamic_pvc[CephBlockPool-Retain]`, I see:

```
19:47:17 - MainThread - root - INFO - waiting for a single resource                                                                                            
19:47:17 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig /home/ocsqe/data/cluster-2019-08-19.1/auth/kubeconfig get PersistentVolume pvc-f5d5b4d8-c370-11e9-86b3-0a2d85acd846
19:47:18 - MainThread - root - WARNING - status of pvc-f5d5b4d8-c370-11e9-86b3-0a2d85acd846 was 11m, waiting for Released
```

which means that instead of status of given object, the function works with age instead.

When the state of the object in question is rechecked manually, we see that the status is indeed already `Released` as expected (I checked this both during and after the test run):

```
$ oc get pv pvc-f5d5b4d8-c370-11e9-86b3-0a2d85acd846
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS     CLAIM                                                                                       STORAGECLASS                                             REASON   AGE
[...]

PR #608, opened 2019-08-21
Fix issue #591

RazTamir:

### DIFF
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -6,6 +6,7 @@ import logging
 import time
 import yaml
 import shlex
+import re
 
 from ocs_ci.ocs.exceptions import CommandFailed
 from ocs_ci.utility.utils import TimeoutSampler
@@ -358,15 +359,21 @@ class OCP(object):
             str: The status returned by 'oc get' command not in the 'yaml'
                 format
         """
-        status_index = None
+        # Get the resource in str format
         resource = self.get(resource_name=resource_name, out_yaml_format=False)
+        # get the list of titles
+        titles = re.sub('\s{2,}', ',', resource)  # noqa: W605
+        titles = titles.split(',')
+        # Get the index of 'STATUS'
+        status_index = titles.index('STATUS')
         resource = shlex.split(resource)
-        for idx, i in enumerate(resource):
-            if i.isupper() and i == 'STATUS':
-                status_index = idx
-            if not i.isupper():
-                break
-        resource_info = [i for i in resource if not i.isupper()]
+        # Get the values from the output including access modes in capital
+        # letters
+        resource_info = [
+            i for i in resource if (
+                not i.isupper() or i in ('RWO', 'RWX', 'ROX')
+            )
+        ]
 
         return resource_info[status_index]
 


ocs_ci/ocs/ocp.py
```diff
@@ -358,15 +359,21 @@ def get_resource_status(self, resource_name):
...
-            if i.isupper() and i == 'STATUS':
-                status_index = idx
-            if not i.isupper():
-                break
-        resource_info = [i for i in resource if not i.isupper()]
+        # Get the values from the output including access modes in capital
+        # letters
+        resource_info = [
+            i for i in resource if (
+                not i.isupper() or i in ('RWO', 'RWX', 'RXO')
```
sidhant-agrawal: `RXO` should be changed to `ROX`

sidhant-agrawal: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

RazTamir: fixing - thanks!

ocs_ci/ocs/ocp.py
```diff
@@ -358,15 +359,21 @@ def get_resource_status(self, resource_name):
...
+        status_index = titles.index('STATUS')
         resource = shlex.split(resource)
-        for idx, i in enumerate(resource):
-            if i.isupper() and i == 'STATUS':
-                status_index = idx
-            if not i.isupper():
-                break
-        resource_info = [i for i in resource if not i.isupper()]
+        # Get the values from the output including access modes in capital
+        # letters
```
mbukatov: Could we add explicit note that we basically assume that all content lines from the output are not using all caps (with exception of access mode), because that is the way we figure out what is a content and what a title? This won't work if we use capital letters only in a name of storageclass, pv or a namespace.

mbukatov:
@RazTamir I have a related question: why `exec_oc_cmd()` method strips out newlines so that stdout is placed into a single line?

RazTamir:
@mbukatov , the issue you raised is valid but considered as an enhancement and not a bug.
This PR is a high priority that blocks tier1 execution.
If that's the only issue in this PR, please approve , we'll merge, and propose a fix for the mentioned issue you raised

RazTamir:
Verification link:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/470/console

mbukatov:
> @mbukatov , the issue you raised is valid but considered as an enhancement and not a bug.

I don't agree with this.

I tried argue that fixing this properly now would be cheaper for us, and that it will also increase efficiency of the whole team in the long term. This attitude to hack a fix immediately at all cost dismissing or not considering effect on the whole team is what created issue #591 in the first place,  made us uncertain about status of OCS-530 tier1 test in ci for almost a week, and slowed my progress this week.

You are ocs-ci maintainer and manager, fully responsible for this repo and the team. This means that you can have it merged no matter that this is a hack and that  #612 is a bug. Eg. if you argue that it make sense from the business perspective to have this resolved immediately at all costs, you can do that. You don't have to claim that this is not a bug to do this. And you can do that no matter if I agree with it. I fully respect your authority to argue like this and execute accordingly.

That said, I don't know how to read your claim that this is not a bug and the fact that you cleared result of my review. You don't need to lower yourself like that.

PR #614, opened 2019-08-21
Removing print() and using logger.info() in test_pod_exec.py

kshithijiyer:
It's not good to have print() statements in a testcase hence changing print() to logger.info() instead.

### DIFF
--- a/tests/libtest/test_pod_exec.py
+++ b/tests/libtest/test_pod_exec.py
@@ -1,8 +1,11 @@
 import os
+import logging
 os.sys.path.append(os.path.dirname(os.getcwd()))
 
 from ocs_ci.framework.testlib import libtest
 from ocs_ci.ocs.resources import pod
+
+logger = logging.getLogger(__name__)
 
 
 @libtest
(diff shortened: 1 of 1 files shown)


tests/libtest/test_pod_exec.py
```diff
@@ -12,7 +15,7 @@ def test_main():
 
     out, err, ret = tools_pod.exec_ceph_cmd(ceph_cmd=cmd)
     if out:
-        print(out)
+        logger.info(out)
     if err:
-        print(err)
-    print(ret)
+        logger.info(err)
```
clacroix12: should this be error level instead?

kshithijiyer: Done.

kshithijiyer:
@clacroix12 is there anything else to be changed?

tests/libtest/test_pod_exec.py
```diff
@@ -1,9 +1,12 @@
...
 os.sys.path.append(os.path.dirname(os.getcwd()))
 
 from ocs_ci.framework.testlib import libtest
 from ocs_ci.ocs.resources import pod
 
+logger = logging.getLogger(__name__)
+
 
 @libtest
 def test_main():
```
clacroix12: I would rename this to `test_pod_exec`. I know that the module is named that but `test_main` is too generic.

ebenahar:
Closing old PRs. Please re-open if needed

PR #616, opened 2019-08-21
Implement silent logging option

MeridianExplorer:
Fixes: #621


ocs_ci/utility/utils.py
```diff
@@ -366,7 +366,9 @@ def run_cmd(cmd, **kwargs):
         (str) Decoded stdout of command
 
     """
-    log.info(f"Executing command: {cmd}")
+    silence_logs = kwargs.pop('silent', False)
```
vasukulkarni: kwargs.get() instead,  How about *quite* as an option?

ocs_ci/utility/utils.py
```diff
@@ -366,7 +366,9 @@ def run_cmd(cmd, **kwargs):
...
+    if not silence_logs:
```
RazTamir: Could you please create a function to silent the logs?
This approach is very hard to maintain with if/esle all over:

ipdb> import sys 
ipdb> logging.info("Before disable")                                                                                                                                                    
08:33:00 - MainThread - root - INFO - Before disable                                                                                                                                                              
ipdb> logging.disable(sys.maxsize)                                                                                                                                                           
ipdb> logging.info("silent mode")                                                                                                                                                    
ipdb> logging.disable(logging.NOTSET)                                                                                                                                                        
ipdb> logging.info("Enable logging again")                                                                                                                                                    
08:35:32 - MainThread - root - INFO - Enable logging again

ocs_ci/utility/utils.py
MeridianExplorer: Using `.get()` creates a problem later on, because `kwargs` are passed over to `subprocess.run()` which doesn't know how to handle the `silent` option and crashes. Pop gives the same safety (if it doesn't exist, it returns False), while also returning `kwargs` to their original state.
Regarding `quiet` - sounds good, I'll implement that.

ocs_ci/ocs/resources/pod.py
```diff
@@ -124,7 +124,7 @@ def get_fio_results(self):
             logger.exception(f"Found Exception: {ex}")
             raise
 
-    def exec_cmd_on_pod(self, command, out_yaml_format=True):
+    def exec_cmd_on_pod(self, command, out_yaml_format=True, **kwargs):
```
RazTamir: Few comments that I believe will make this PR more simple.
See below

ocs_ci/utility/utils.py
```diff
@@ -352,6 +353,12 @@ def custom_ceph_config(suite_config, custom_config, custom_config_file):
     return full_custom_config
 
 
+def quiet_logs(quiet):
+    if quiet:
+        logging.disable(sys.maxsize)
+    return quiet
```
RazTamir: 1) no need to return this value

ocs_ci/utility/utils.py
```diff
@@ -366,25 +373,33 @@ def run_cmd(cmd, **kwargs):
...
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            stdin=subprocess.PIPE,
+            **kwargs
         )
-    return r.stdout.decode()
+        log.debug(f"Command output: {r.stdout.decode()}")
+        if r.stderr and not r.returncode:
+            log.warning(f"Command warning:: {r.stderr.decode()}")
+        logging.disable(logging.NOTSET)
```
RazTamir: 2. 'Try:' clause should end here

ocs_ci/utility/utils.py
```diff
@@ -366,25 +373,33 @@ def run_cmd(cmd, **kwargs):
...
+        if r.returncode:
+            raise CommandFailed(
+                f"Error during execution of command: {'quiet command' if quiet_state else cmd}."
+                f"\nError is {r.stderr.decode()}"
+            )
+        return r.stdout.decode()
+    except Exception as e:
```
RazTamir: 3. This will catch the CommandFailed in line 393 so needs to move earlier - see comment 2
4. No need for except - just finally is enough

ocs_ci/utility/utils.py
```diff
@@ -366,25 +373,33 @@ def run_cmd(cmd, **kwargs):
...
+        logging.disable(logging.NOTSET)
+        log.error(f"Exception while executing command - {e}")
+    finally:
```
RazTamir: In overall it should look something like:
```python
try:
    all_code_goes_here
finally:
    logging.disable(logging.NOTSET)
```

ocs_ci/utility/utils.py
MeridianExplorer: The problem is that I need to keep it in case a CommandFailed is raised. By default, it prints out the command that failed. If we'll simply silence it, we won't know which command failed.
Alternatively I could replace .pop with .get, but in that case, `quiet` would be sent over to `subprocess.run` and would crash

ocs_ci/utility/utils.py
MeridianExplorer: Any reason not to end it directly after `subprocess.run`? 
I don't think `if r.stderr and not r.returncode...` can cause an exception

ocs_ci/utility/utils.py
MeridianExplorer: Regarding 3 - I commented on the respective comment.
Regarding 4 - wouldn't we like to log the failure?

ocs_ci/utility/utils.py
RazTamir: ack

ocs_ci/utility/utils.py
RazTamir: ok

ocs_ci/utility/utils.py
RazTamir: 3. ack
4. yes we do

petr-balogh:
@MeridianExplorer do we have issue opened for this?  Please see second bullet here:  https://github.com/red-hat-storage/ocs-ci/blob/master/docs/code_review.md

Would be good to be descriptive here in PR or at least in the issue and describe the issue, enhancement or why it's need or so.

This will help everyone to understand what it's gonna be fixed, solved, enhanced and why.

Thanks.

ocs_ci/utility/utils.py
```diff
@@ -358,6 +358,9 @@ def run_cmd(cmd, **kwargs):
 
     Args:
         cmd (str): command to run
+        quiet (bool, optional, kw): Whether to log the current cmd execution
```
petr-balogh: As this is not Args I would use some method of documentation described here with Keyword arguments, WDYT?
https://stackoverflow.com/questions/1137161/what-is-the-correct-way-to-document-a-kwargs-parameter

When you need to break line, please use on second line just one indention.

ocs_ci/utility/utils.py
```diff
@@ -381,7 +386,7 @@ def run_cmd(cmd, **kwargs):
         log.warning(f"Command warning:: {r.stderr.decode()}")
     if r.returncode:
         raise CommandFailed(
-            f"Error during execution of command: {cmd}."
+            f"Error during execution of command: {'quiet command' if quiet_state else cmd}."
```
petr-balogh: Hmm, if we raise the exception anyway and there will be:
Error during execution of command: quiet command  , why not write also whole command if we do have such message in exception. I don't see any benefit to not to have mentioned cmd.

If I would like to avoid of raising the exception then I would add another argument ignore_error and than raise this exception only when not ignore_error.

ocs_ci/utility/utils.py
MeridianExplorer: I agree. Also, noted regarding the line break. Implementing now :)

ocs_ci/utility/utils.py
MeridianExplorer: So here's the thing - even if the command failed, I wouldn't want the credentials to be exposed. However, completely muting all error logs for quiet commands will be confusing and will not reflect what truly happens behind the scenes.
For example:
```Writing testfile to bucket...
Error during execution of command: quiet command```
VS.
```Writing testfile to bucket...```
We have no way of knowing whether the command even succeeded, and if it fails - we have a rough idea of what the command **should've** been.

Comment (petr-balogh) [reply ocs_ci/utility/utils.py]: Ah I see now what is the purpose here now.

I would choose different approach here, WDYT about some another parameter which will be called masked_secrets (list), you will need to define some another variable like masked_cmd inside this function and do something like this:

```python
masked_cmd = cmd
for secret in masked_secrets:
    masked_cmd.replace(secret, '*****')

then everywhere you will log or raise exception you will use this masked_cmd.  This case you will not leak secret data.   You can also make sure that you will mask you secret while returning value on line 393.   So the best way will be to cover the mentioned code in function ;) and just reuse it here and allow to use to use this function on different places where they need to mask their secret data.

Comment (MeridianExplorer) [reply ocs_ci/utility/utils.py]: Great idea, done :)

File: ocs_ci/utility/utils.py [hunk 10]
```diff
@@ -352,21 +352,42 @@ def custom_ceph_config(suite_config, custom_config, custom_config_file):
     return full_custom_config
 
 
+def mask_secrets(cmd, secrets):
```
petr-balogh: it's not cmd specific and can be used for whatever string, so can be called just like text_to_mask or just text. Update documentation accordingly please

ocs_ci/utility/utils.py
```diff
@@ -352,21 +352,42 @@ def custom_ceph_config(suite_config, custom_config, custom_config_file):
...
+
+    Args:
+        cmd (str): The command/plaintext to remove the secrets from
+        secrets (list): List of secret strings to replace in the cmd
+
+    Returns:
+        cmd (str): The censored version of cmd
+    """
+    for secret in secrets:
+        cmd = cmd.replace(secret, '*'*len(secret))
```
petr-balogh: I think that a few asterisks is enough.  For example 5 as you will never know the right length of secret which is more secure :)

ocs_ci/utility/utils.py
```diff
@@ -352,21 +352,42 @@ def custom_ceph_config(suite_config, custom_config, custom_config_file):
...
+    """
+    Replace secrets in plaintext with asterisks
+
+    Args:
+        cmd (str): The command/plaintext to remove the secrets from
+        secrets (list): List of secret strings to replace in the cmd
+
+    Returns:
+        cmd (str): The censored version of cmd
+    """
```
petr-balogh: one empty line after Returns section

ocs_ci/utility/utils.py
```diff
@@ -381,7 +402,7 @@ def run_cmd(cmd, **kwargs):
         log.warning(f"Command warning:: {r.stderr.decode()}")
     if r.returncode:
         raise CommandFailed(
-            f"Error during execution of command: {cmd}."
+            f"Error during execution of command: {masked_cmd}."
             f"\nError is {r.stderr.decode()}"
```
petr-balogh: You would like to probably also mask the stderr here

ocs_ci/utility/utils.py
```diff
@@ -381,7 +402,7 @@ def run_cmd(cmd, **kwargs):
...
         )
     return r.stdout.decode()
```
petr-balogh: maybe also worth to mask stdout?

ocs_ci/utility/utils.py
MeridianExplorer: I actually thought about that, thought that secret are fixed-length, but then found out they actually aren't :)
Hoped it'd be good for debugging purposes (for example, secrets that are obviously too long/short), but the security is more important.

Replaced! Important catch. Thank you :)

ocs_ci/utility/utils.py
```diff
@@ -352,21 +352,44 @@ def custom_ceph_config(suite_config, custom_config, custom_config_file):
...
 def run_cmd(cmd, **kwargs):
     """
     Run an arbitrary command locally
 
     Args:
         cmd (str): command to run
 
+    Keyword Args:
+        secrets (list): Whether to log the current cmd execution.
+          This kwarg is popped in order to not interfere with subprocess.run(**kwargs)
```
petr-balogh: 4 spaces indetion instead of 2

ocs_ci/utility/utils.py
```diff
@@ -352,21 +352,44 @@ def custom_ceph_config(suite_config, custom_config, custom_config_file):
...
+
 def run_cmd(cmd, **kwargs):
     """
     Run an arbitrary command locally
 
     Args:
         cmd (str): command to run
 
+    Keyword Args:
+        secrets (list): Whether to log the current cmd execution.
```
petr-balogh: Whether to log the current cmd execution. Doesn't sound like proper doc for this

ocs_ci/utility/utils.py
```diff
@@ -352,21 +352,44 @@ def custom_ceph_config(suite_config, custom_config, custom_config_file):
...
+def mask_secrets(plaintext, secrets):
+    """
+    Replace secrets in plaintext with asterisks
+
+    Args:
+        plaintext (str): The plaintext to remove the secrets from
+        secrets (list): List of secret strings to replace in the plaintext
+
+    Returns:
+        plaintext (str): The censored version of plaintext
```
petr-balogh:
```
Returns:
    str: The censored ....

File: ocs_ci/utility/utils.py [hunk 18]
```diff
@@ -352,21 +352,44 @@ def custom_ceph_config(suite_config, custom_config, custom_config_file):
...
+def mask_secrets(plaintext, secrets):
+    """
+    Replace secrets in plaintext with asterisks
+
+    Args:
+        plaintext (str): The plaintext to remove the secrets from
+        secrets (list): List of secret strings to replace in the plaintext
+
+    Returns:
+        (str): The censored version of plaintext
```
petr-balogh: no need for brackets in Returns section, just str: The censored

ocs_ci/utility/utils.py
```diff
@@ -352,21 +352,44 @@ def custom_ceph_config(suite_config, custom_config, custom_config_file):
...
 def run_cmd(cmd, **kwargs):
     """
     Run an arbitrary command locally
 
     Args:
         cmd (str): command to run
 
+    Keyword Args:
+        secrets (list): The list of secrets to be masked with asterisks
+            This kwarg is popped in order to not interfere with subprocess.run(**kwargs)
```
petr-balogh: I would keep this under 79 chars, so please break the line

ocs_ci/utility/utils.py
```diff
@@ -352,21 +352,44 @@ def custom_ceph_config(suite_config, custom_config, custom_config_file):
...
+    Args:
+        plaintext (str): The plaintext to remove the secrets from
+        secrets (list): List of secret strings to replace in the plaintext
+
+    Returns:
+        (str): The censored version of plaintext
+
+    """
+    for secret in secrets:
+        plaintext = plaintext.replace(secret, '*' * 5)
```
RazTamir: replace 5 with len(secret) if that's the intention

ocs_ci/utility/utils.py
```diff
@@ -352,21 +352,44 @@ def custom_ceph_config(suite_config, custom_config, custom_config_file):
...
     Raises:
         CommandFailed: In case the command execution fails
 
     Returns:
         (str) Decoded stdout of command
 
     """
-    log.info(f"Executing command: {cmd}")
+    secrets = kwargs.pop('secrets', [])
+    masked_cmd = mask_secrets(cmd, secrets)
```
RazTamir: perfect!

ocs_ci/utility/utils.py
```diff
@@ -378,13 +401,13 @@ def run_cmd(cmd, **kwargs):
     )
     log.debug(f"Command output: {r.stdout.decode()}")
     if r.stderr and not r.returncode:
-        log.warning(f"Command warning:: {r.stderr.decode()}")
+        log.warning(f"Command warning:: {mask_secrets(r.stderr.decode(), secrets)}")
```
RazTamir: double ::

ocs_ci/utility/utils.py
MeridianExplorer: len(secret) might become a security issue, so I and Petr agreed it'd be best to mask it with a static length

ocs_ci/utility/utils.py
MeridianExplorer: Sharp catch!

ocs_ci/utility/utils.py
petr-balogh: I suggested in some previous comment to not to do that and just keep with some static length here which doesn't tell anything about length of secret Raz

petr-balogh:
@MeridianExplorer I think you can squash all the commits to 1 commit now.  Or if someone will merge this one please Squash and Merge!

ocs_ci/utility/utils.py
RazTamir: ack

ocs_ci/utility/utils.py
```diff
@@ -352,21 +352,45 @@ def custom_ceph_config(suite_config, custom_config, custom_config_file):
...
 def run_cmd(cmd, **kwargs):
     """
     Run an arbitrary command locally
 
     Args:
         cmd (str): command to run
 
+    Keyword Args:
+        secrets (list): The list of secrets to be masked with asterisks
+            This kwarg is popped in order to not interfere with
```
vasukulkarni: I find this is hidden thing for users but..

clacroix12: This seems very prone to error. Can we be 100% sure that we pop the correct `kwarg` off so we don't interfere with what might be being passed to subprocess? What is stopping `secrets` from being it's own arg so we don't hit this issue?

MeridianExplorer: Not reason at all to prefer **kwargs. Fixed, thank you :)

MeridianExplorer:
Jenkins PR job seems to have failed, but because of irrelevant issues - http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/pr616/pr616_20190825T064055/logs/test_report.html

RazTamir:
The results seem identical to product execution

PR #618, opened 2019-08-22
Fix create_pvc() calls in test_resource_deletion_during_pvc_pod_creat…

ebenahar:
…ion_and_io.py

### DIFF
--- a/tests/manage/pv_services/test_resource_deletion_during_pvc_pod_creation_and_io.py
+++ b/tests/manage/pv_services/test_resource_deletion_during_pvc_pod_creation_and_io.py
@@ -92,9 +92,11 @@ class OperationsBase(ManageTest):
         # Create pods for running IO
         io_pods = helpers.create_pods(
             pvc_objs_list=pvc_objs_for_io_pods, interface_type=self.interface,
-            desired_status=constants.STATUS_RUNNING, wait=True,
             namespace=self.namespace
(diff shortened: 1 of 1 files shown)


tests/manage/pv_services/test_resource_deletion_during_pvc_pod_creation_and_io.py
```diff
@@ -92,9 +92,11 @@ def operations_base(self, resource_to_delete):
         # Create pods for running IO
         io_pods = helpers.create_pods(
             pvc_objs_list=pvc_objs_for_io_pods, interface_type=self.interface,
-            desired_status=constants.STATUS_RUNNING, wait=True,
             namespace=self.namespace
         )
+        for pod in io_pods:
```
RazTamir: always take care of adding the resources to the list for teardown before any other operation after the creation itself.
In this case, if wait_for_resource_state() will faill, you will have all those pods as leftovers

ebenahar: this PR won't change to factory fixtures, it's just to fix the separation between resource creation and wait. 
We'll change this file to factory later

ebenahar:
Changes will be addressed as part of https://github.com/red-hat-storage/ocs-ci/pull/693

Issue #628, opened 2019-08-23
Add check for collection of tests


petr-balogh:
Add --collect-only to tox to prevent issues like we had today #626

zmc:
Would it be enough to check "Include Administrators" [here](https://github.com/red-hat-storage/ocs-ci/settings/branch_protection_rules/5068680) ?

petr-balogh:
@zmc How does it help in case I am fixing? The ci passed for case mentioned in issue. This will ensure we are still able to collect all tests and there are no critical failures in imports and so on. See the example of failure in comment.

PR #629, opened 2019-08-23
Allow collect only without running cluster

petr-balogh:
Fixes issue #628
This allow us to run collect-only option with pytest to check some import errors and collection issues.

### DIFF
--- a/ocs_ci/framework/pytest_customization/reports.py
+++ b/ocs_ci/framework/pytest_customization/reports.py
@@ -45,5 +45,5 @@ def pytest_sessionfinish(session, exitstatus):
     """
     send email report
     """
-    if ocsci_config.RUN['cli_params']['email']:
+    if ocsci_config.RUN['cli_params'].get('email'):
         email_reports()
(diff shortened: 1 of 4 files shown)


petr-balogh:
I created one testing branch to test some import issue here: 
https://github.com/petr-balogh/ocs-ci/commit/a89d79062c51bfa981dbb03b82683b46aa375334
Results which are failing our CI:
https://travis-ci.org/petr-balogh/ocs-ci/builds/575738884

ocs_ci/framework/pytest_customization/reports.py
```diff
@@ -45,5 +45,5 @@ def pytest_sessionfinish(session, exitstatus):
     """
     send email report
     """
-    if ocsci_config.RUN['cli_params']['email']:
+    if ocsci_config.RUN['cli_params'].get('email'):
```
vasukulkarni: This seem like "add-on" commit,  please remove this line.

petr-balogh: This is required otherwise failing with keyerror

vasukulkarni: weird but we should revisit this one to find out more

PR #630, opened 2019-08-23
Automates TC OCS-965 : Test shared Mount Point is Writeable even after some pods are deleted

nehaberry:
This PR automates the following Test case:
OCS-965 - CEPHFS RWX: While two app pods access same mount, delete one pod. Confirm second pod is still able to write

Please review and suggest changes.

Note: I have re-used "verify_pod_node" from "test_pvc_assign_pod_node.py". Please suggest if this should be moved to helper.py for easy re-usability.

Signed-off-by: Neha Berry <nberry@redhat.com>


sidhant-agrawal:
> LGTM.
> Please move the verify_pod_node to helpers.py

@RazTamir @nehaberry how about moving to pod.py

nehaberry:
Tested by moving to helpers. Let me know if its needed to be moved to pod.py. will re-run and send the PR again.

@raz do i need to run this in jenkins as well?

RazTamir:
Ack for having this in pod.py.
Yes please, for execution in jenkins

tests/manage/pv_services/test_pvc_rwx_writeable_after_pod_deletions.py
```diff
@@ -0,0 +1,166 @@
...
+from concurrent.futures import ThreadPoolExecutor
+from ocs_ci.framework.testlib import ManageTest, tier1
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.resources import pod
+from tests import helpers
+
+logger = logging.getLogger(__name__)
+
+
+def verify_pod_node(pod_obj, node_name):
```
vasukulkarni: s/verify_pod_name/verify_node_name and as commented in earlier, please move to pod class.

```python
pod.verify_node_name(node_name)
```

ebenahar: I think you can use ocs_ci.ocs.resources.pod.get_pod_node and check whether the node name returned by the function is the particular node or not from the test itself

tests/manage/pv_services/test_pvc_rwx_writeable_after_pod_deletions.py
```diff
@@ -0,0 +1,166 @@
...
+        pvc_obj = pvc_factory(
+            interface=interface, access_mode=constants.ACCESS_MODE_RWX,
+            size=10, status=constants.STATUS_BOUND
+        )
+        logger.info(
+            f"Creating pods on all worker nodes backed"
+            f"with same pvc {pvc_obj.name}"
+        )
+
+        pod_list = [helpers.create_pod(
```
ebenahar: Please avoid using list comprehension here.
The reason is that in case one of the pods creation will fail, pod_list won't get filled with the ones that got created, and won't be deleted on teardown

tests/manage/pv_services/test_pvc_rwx_writeable_after_pod_deletions.py
```diff
@@ -0,0 +1,166 @@
...
+                    pod_obj.run_io, storage_type='fs', size='512M',
+                    runtime=30, fio_filename=pod_obj.name
+                )
+
+        # Check IO from all pods
+        for pod_obj in pod_list:
+            pod.get_fio_rw_iops(pod_obj)
+
+        # Calculate md5sum of each file
+        md5sum_pod_data = []
```
ebenahar: nice!

tests/manage/pv_services/test_pvc_rwx_writeable_after_pod_deletions.py
ebenahar: instead, append to a regular list and in each iteration, call teardown_factory(pod_obj)

nehaberry:
> Ack for having this in pod.py.
Moved to pod.py but outside the class, similar to other helper functions.
> Yes please, for execution in jenkins
Currently due to OCP installer download issue , unable to verify the test case on jenkins. verified on local setup.

tests/manage/pv_services/test_pvc_rwx_writeable_after_pod_deletions.py
nehaberry: Done

```
pod_list = []

        for each_node in worker_nodes_list:
            pod_obj = helpers.create_pod(
                interface_type=interface, pvc_name=pvc_obj.name,
                namespace=pvc_obj.namespace, node_name=each_node,
                pod_dict_path=constants.NGINX_POD_YAML
            )
            pod_list.append(pod_obj)
            teardown_factory(pod_obj)
```

Please review

tests/manage/pv_services/test_pvc_rwx_writeable_after_pod_deletions.py
sidhant-agrawal: It appears ocs_ci.ocs.resources.pod.get_pod_node is returning node object, won't that be an overhead and cause code duplication everytime just for checking the nodename of a pod?

nehaberry: @vasukulkarni the function is re-named as "verify_node_name".

Also, moved it to pod.py but not within the class. Mostly, all helper type functions were present outside the class, hence added this outside as well. 

Please let me know if its not ok.
```
 pod.verify_node_name(pod_obj, worker_nodes_list[node_count])
```

nehaberry:
[ocs-965-run-results.zip](https://github.com/red-hat-storage/ocs-ci/files/3540915/ocs-965-run-results.zip)
The Run result is attached here

```

================================================ PASSES =================================================
======================================== short test summary info ========================================
PASSED tests/manage/pv_services/test_pvc_rwx_writeable_after_pod_deletions.py::TestRWXMountPoint::test_pvc_rwx_writeable_after_pod_deletions[CephFileSystem]
====================================== 1 passed in 946.03 seconds =======================================
```

tests/manage/pv_services/test_pvc_rwx_writeable_after_pod_deletions.py
```diff
@@ -0,0 +1,147 @@
...
+
+class TestRWXMountPoint(ManageTest):
+    """
+    Automates the following test cases:
+    OCS-965	CEPHFS RWX: While two app pods access same mount,
+        delete one pod. Confirm second pod is still able to write
+    """
+
+    @tier1
+    @pytest.mark.parametrize(
```
RazTamir: if you don't plan to add also CephFilesystem interface, please remove parametrize

nehaberry: @RazTamir done. Thanks

Earlier we thought we need to cover for RBD RWX as well. But we don't need to test RBD + RWX with all scenarios as its a special usecase. Hence removed the parameterize as suggected by you.

If in future we want to include RBD, we can again add the parameterize.

nehaberry:
Verified on local setup due to issues in creating setups in Jenkins.

PR #584, opened 2019-08-16
Testing basics of RBD RWX raw block pv support:OSC-750 and OCS-751

shreekarSS:
tests/manage/pv_services/test_raw_block_pv.py
Testing basics of RBD RWX raw block pv feature

ocs_ci/ocs/constants.py
Added pod template for rbd raw block pv

tests/helpers.py
Added parameter volume_mode to create_pvc() for creating raw block pv
Modified number of retries of function validate_pv_delete() due to bz1726266

ocs_ci/ocs/resources/pod.py
For block IOs, added path of block device

PR #636, opened 2019-08-25
Wait for PV to delete in pvc_factory finalizer

jilju:
Ensure that PV is deleted after deleting PVC to avoid errors caused by PV leftover.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -307,10 +307,21 @@ def pvc_factory(
         """
         Delete the PVC
         """
+        pv_objs = []
+
+        # Get PV form PVC instances and delete PVCs
         for instance in instances:
-            instance.delete()
-            instance.ocp.wait_for_delete(
-                instance.name
+            if not instance.is_deleted:
+                pv_objs.append(instance.backed_pv_obj)
+                instance.delete()
+                instance.ocp.wait_for_delete(
+                    instance.name
+                )
+
+        # Wait for PVs to delete
+        for pv_obj in pv_objs:
+            pv_obj.ocp.wait_for_delete(
+                resource_name=pv_obj.name, timeout=180
             )
 
     request.addfinalizer(finalizer)


ebenahar:
@jilju Please run tier1 for verification

jilju:
> @jilju Please run tier1 for verification

[Build](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/540/) started.

tests/conftest.py
```diff
@@ -307,10 +307,21 @@ def finalizer():
...
+            if not instance.is_deleted:
+                pv_objs.append(instance.backed_pv_obj)
+                instance.delete()
+                instance.ocp.wait_for_delete(
+                    instance.name
+                )
+
+        # Wait for PVs to delete
+        for pv_obj in pv_objs:
+            pv_obj.ocp.wait_for_delete(
```
RazTamir: If the reclaim policy is retain it will fail

ebenahar: I think we can expect that in this teardown phase, all PVs should be deleted, even if the PVCs reclaim policy is retain (by the test itself or with a teardwon that should take place before this teardown).

RazTamir: not sure I get it.
if this is the expectation that all PVs will be deleted by now (this point in code), why do we need to add this wait_for_delete?

ebenahar: but in the case of 'delete' reclaim policy, in the tests, we don't explicitly delete the PVs.
Also, the main reason for this PR is https://bugzilla.redhat.com/show_bug.cgi?id=1726266

ebenahar: @jilju 
You can check the reclaim policy the backed_reclaim_policy added here https://github.com/red-hat-storage/ocs-ci/pull/584/files#diff-31d76ef98c9184ea833bdd9302eba288R104
and validate PV deletion only in case the policy is 'delete'

jilju: Or PV should be explicitly deleted here ?

jilju: If reclaim policy is retain and PV is not deleted, then  teardown will fail in leftover check.

jilju:
> > @jilju Please run tier1 for verification
> 
> [Build](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/540/) started.

29 passed, 10 skipped, 3 failed, 3 errors, 0 exp ected failures, 0 unexpected passes

tests/conftest.py
ebenahar: @jilju I agree. In case of retain policy, PV deletion should be taken care of in the test or in a teardwon that takes place before pvc_factory

RazTamir: ack.
We will need to have issue on enhancing the log in ocp.wait_for_delete() to describe the object when fail to delete and also add the reclaim policy after @shreekarSS PR is in.
Lste's merge this one

PR #652, opened 2019-08-27
Upgrade the default worker instance type from m4.large to m4.xlarge

MeridianExplorer:
For NooBaa deployment (requires a minimum of m4.xlarge worker machine)

### DIFF
--- a/ocs_ci/templates/ocp-deployment/install-config-aws-ipi.yaml.j2
+++ b/ocs_ci/templates/ocp-deployment/install-config-aws-ipi.yaml.j2
@@ -4,7 +4,7 @@ compute:
   - name: worker
     platform:
       aws:
-        type: {{ worker_instance_type | default('m4.large') }}
+        type: {{ worker_instance_type | default('m4.xlarge') }}
 {% if worker_availability_zones %}
         zones:
   {% for zone in worker_availability_zones %}
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -77,3 +77,4 @@ ENV_DATA:
   # but only once when move to new version like v15.
   ceph_image: 'ceph/ceph:v14.2.2-20190826'
   rook_image: 'rook/ceph:master'
+  worker_instance_type: 'm4.xlarge'


MeridianExplorer:
[Jenkins PR build](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/84/)

nehaberry:
@MeridianExplorer as per KNIP-821, the minimum expected instance type for ocs is m5.4xlarge.

But are we sticking with m4.xlarge to save cost? Just a question.

MeridianExplorer:
> @MeridianExplorer as per KNIP-821, the minimum expected instance type for ocs is m5.4xlarge.
> 
> But are we sticking with m4.xlarge to save cost? Just a question.

@nehaberry - Indeed :)

petr-balogh:
@MeridianExplorer is this verified? Can we merge?

vasukulkarni:
@MeridianExplorer ping?

MeridianExplorer:
Currently rebasing + trying to rerun the validation job

ocs_ci/templates/ocp-deployment/install-config-aws-ipi.yaml.j2
```diff
@@ -4,7 +4,7 @@ compute:
   - name: worker
     platform:
       aws:
-        type: {{ worker_instance_type | default('m4.large') }}
+        type: {{ worker_instance_type | default('m4.xlarge') }}
```
petr-balogh: Just thinking about the limit issue we've just hit in AWS about 105 instances. If currently we are using two different and we will switch master and worker to the same it means we can install just 17 clusters right?  @RazTamir I think this will be problematic cause of mentioned issue.

RazTamir: Good point,
Let me ask for increase this instance too

MeridianExplorer: @petr-balogh 
In this case, do we have a green light to merge this once it's approved and passes deployment testing?

RazTamir: based on our discussion I think this PR is ready?

petr-balogh:
Yep, we can merge once verified.

PR #667, opened 2019-08-28
Add factory fixture for deploymentconfig pod

prsurve:
1- Add factory fixture for deployment config pod
2- Move script from fixture to factory fixture

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -465,9 +465,9 @@ def service_account_factory(request):
         if active_service_account_obj and not service_account:
             return active_service_account_obj
         elif service_account:
-            sa_obj = helpers.get_serviceaccount_obj(sa_name=service_account, namespace=pvc.namespace.project)
-            if not helpers.validate_scc_policy(sa_name=service_account, namespace=pvc.namespace.project):
-                helpers.add_scc_policy(sa_name=service_account, namespace=pvc.namespace.project)
+            sa_obj = helpers.get_serviceaccount_obj(sa_name=service_account, namespace=project.namespace)
+            if not helpers.validate_scc_policy(sa_name=service_account, namespace=project.namespace):
+                helpers.add_scc_policy(sa_name=service_account, namespace=project.namespace)
             sa_obj.project = project
             active_service_account_obj = sa_obj
             instances.append(sa_obj)
@@ -536,7 +536,7 @@ def dc_pod_factory(
         else:
 
             pvc = pvc or pvc_factory(interface=interface, size=size)
-            sa_obj = service_account_factory(project=pvc.project,service_account=service_account)
+            sa_obj = service_account_factory(project=pvc.project, service_account=service_account)
             dc_pod_obj = helpers.create_pod(
                 interface_type=interface, pvc_name=pvc.name, do_reload=False,
                 namespace=pvc.namespace, sa_name=sa_obj.name, dc_deployment=True,
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -448,18 +448,30 @@ def service_account_factory(request):
     instances = []
     active_service_account_obj = None
 
-    def factory(project=None):
+    def factory(
+        project=None, service_account=None
+    ):
         """
         Args:
             project (object): ocs_ci.ocs.resources.ocs.OCS instance
                 of 'Project' kind.
+            service_account (str): service_account_name
 
         Returns:
-            object: helpers.create_serviceaccount instance.
+            object: serviceaccount instance.
         """
         nonlocal active_service_account_obj
-        if active_service_account_obj:
+
+        if active_service_account_obj and not service_account:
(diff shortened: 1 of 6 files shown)


prsurve:
[Logs](https://github.com/red-hat-storage/ocs-ci/files/3551181/factory_pod_logs.log)

tests/fixtures.py
```diff
@@ -298,7 +298,8 @@ def finalizer():
     class_instance.dc_pod_objs = [
         helpers.create_pod(
             interface_type=class_instance.interface, pvc_name=pvc_obj.name, do_reload=False,
```
ebenahar: do we still need this fixture? maybe it would be better to use the new factory fixture for deployment config

prsurve: This is an old fixture if someone needs they can use this but for my script, I am using factory fixture

tests/conftest.py
```diff
@@ -429,6 +429,112 @@ def finalizer():
...
+                by using these data. Parameter `pvc` is not used but reference
+                is set if provided.
+            replica_count (int): Replica count for deployment config
+        """
+        if custom_data:
+            dc_pod_obj = helpers.create_resource(**custom_data)
+        else:
+
+            pvc = pvc or pvc_factory(interface=interface, size=size)
+            sa_obj = service_account_factory(project=pvc.project)
```
fbalak: User should have option to add service_account via parameter (as with pvc)

tests/conftest.py
```diff
@@ -429,6 +429,112 @@ def finalizer():
...
+            )
+            instance.delete()
+            instance.ocp.wait_for_delete(resource_name=instance.name)
+
+    request.addfinalizer(finalizer)
+    return factory
+
+
+@pytest.fixture()
+def dc_pod_factory(
```
fbalak: Would it make sense to enhance `pod_factory` with new parameters like `service_account` and `dc_deployment` instead of creating this new fixture?

tests/conftest.py
prsurve: @fbalak sorry i didn't get you

tests/conftest.py
prsurve: personally, I feel we should have separate dc_pod_factory it will be easy for everyone to use if we suppose to make a change in pod_factory users  need to add additional data to pvc_factory for setting up dc_pods

tests/conftest.py
fbalak: I am sorry that I did not make myself clear. I meant to add `service_account` to factory function like:
```
   def factory(
        interface=constants.CEPHBLOCKPOOL,
        pvc=None,
        service_account=None,
        size=None,
        custom_data=None,
        replica_count=1,
    ):
```
and then fill this variable like:
```
sa_obj = service_account or service_account_factory(project=pvc.project)
```
This would allow user that uses `dc_pod_factory` to control what service account is used in created pod object.

prsurve: @fbalak service_account should be created under newly created namespace be cannot add service account from other namespace dc pod won't come in running state

tests/conftest.py
```diff
@@ -429,6 +429,112 @@ def finalizer():
...
+        """
+        Args:
+            project (object): ocs_ci.ocs.resources.ocs.OCS instance
+                of 'Project' kind.
+
+        Returns:
+            object: helpers.create_serviceaccount instance.
+        """
+        nonlocal active_service_account_obj
+        if active_service_account_obj:
```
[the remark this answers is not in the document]
vasukulkarni: please ignore, its initialized in closure function second time

tests/conftest.py
fbalak: Right, but if you add the parameter then you will have option to test adding service_account from different namespace it as negative testcase or create two pods with different user accounts in same namespace.

prsurve: Done

tests/manage/pv_services/test_run_io_multiple_dc_pods.py
```diff
@@ -1,41 +1,24 @@
...
+        pytest.param(
+            constants.CEPHFILESYSTEM, marks=pytest.mark.polarion_id("OCS-1285")
+        )
+    ]
 )
-class BaseRunIOMultipleDcPods(ManageTest):
+class TestRunIOMultipleDcPods(ManageTest):
     """
-    Run IO on multiple dc pods in parallel
+    Run IO on multiple dc pods in parallel - RBD
```
RazTamir: just RBD?

prsurve: Rbd and ceph fs both will update doc string

ebenahar:
@prsurve Following new commits, please re-verify

prsurve:
Logs:- https://paste.fedoraproject.org/paste/z6xgOQIue5jgqcrlCnHXCg

PR #683, opened 2019-09-03
Add option to create a bucket on a specific region

nimrod-becker:
Add an option to create a bucket on a specific region. 
When running against S3, you can specify the specific region for the bucket.

### DIFF
--- a/ocs_ci/ocs/resources/noobaa.py
+++ b/ocs_ci/ocs/resources/noobaa.py
@@ -52,7 +52,7 @@ class NooBaa(object):
         """
         Args:
             bucketname: Name of the bucket to be created
-            region: Name of the region for the bucket to be created on (defualt is us-east-1, as the SDK default)
+            region: Name of the region for the bucket to be created on
 
         Returns:
             s3.Bucket object
--- a/ocs_ci/ocs/resources/noobaa.py
+++ b/ocs_ci/ocs/resources/noobaa.py
@@ -48,16 +48,22 @@ class NooBaa(object):
             aws_secret_access_key=self.access_key
         )
 
-    def s3_create_bucket(self, bucketname):
+    def s3_create_bucket(self, bucketname, region='us-east-1'):
         """
         Args:
             bucketname: Name of the bucket to be created
+            region: Name of the region for the bucket to be created on (defualt is us-east-1, as the SDK default)
 
         Returns:
             s3.Bucket object
 
         """
-        return self.s3_resource.create_bucket(Bucket=bucketname)
+        return self.s3_resource.create_bucket(
+            Bucket=bucketname,
+            CreateBucketConfiguration={
+                'LocationConstraint': region
+            }
+        )
 
     def s3_delete_bucket(self, bucket):
         """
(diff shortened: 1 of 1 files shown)


mykaul:
I'm not sure the default should be set and not None. If it's not None, then it should be used.

nimrod-becker:
us-east-1 is the default of the sdk if you don't provide any region. This just makes it more explicit and visible.

ocs_ci/ocs/resources/noobaa.py
```diff
@@ -48,16 +48,22 @@ def __init__(self):
...
+            region: Name of the region for the bucket to be created on (defualt is us-east-1, as the SDK default)
 
         Returns:
             s3.Bucket object
 
         """
-        return self.s3_resource.create_bucket(Bucket=bucketname)
+        return self.s3_resource.create_bucket(
+            Bucket=bucketname,
+            CreateBucketConfiguration={
```
ebenahar: Can we pass multiple regions here? Seems like 'LocationConstraint' key can get only a string accroding to https://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Types/CreateBucketConfiguration.html

ebenahar: Sorry, this one:
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-example-creating-buckets.html

nimrod-becker: A bucket is always associated with 1 region only.

ocs_ci/ocs/resources/noobaa.py
```diff
@@ -48,16 +48,22 @@ def __init__(self):
             aws_secret_access_key=self.access_key
         )
 
-    def s3_create_bucket(self, bucketname):
+    def s3_create_bucket(self, bucketname, region='us-east-1'):
```
ebenahar: The default region should be taken from the env vars:
```python

self.region = config.ENV_DATA['region']
```

ocs_ci/ocs/resources/noobaa.py
ebenahar: It's clear now, after an offline discussion.
Thanks @nimrod-becker

ocs_ci/ocs/resources/noobaa.py
ebenahar: @MeridianExplorer @parikshithb Please change the region here to be as written above, in a following PR.
Thanks

RazTamir: Our main region is us-east-2 (Ohio).

ebenahar: @RazTamir right!

ebenahar: fixed

ebenahar: need to re-set to region=config.ENV_DATA['region']

MeridianExplorer:
Verified locally because NooBaa isn't deployed yet

========================== 3 passed in 237.39 seconds ==========================

PR #694, opened 2019-09-05
Enable OCS deployment via operator by default

petr-balogh:


dahorak:
I think, that we should change also the _default_ in [deployment.py on line 36](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/deployment.py#L36).
```python
        self.ocs_operator_deployment = config.DEPLOYMENT.get(
            'ocs_operator_deployment', False
        )
```

petr-balogh:
@RazTamir do we want to move to the deployment via operator as default one now? If so we need to merge this one. Thanks for your input.

RazTamir:
Yes please

Issue #696, opened 2019-09-05
Create config for production with m5.4xlarge


petr-balogh:
It was discussed on staff mtg that we should create such config and use it for production runs.
Also align with: https://jira.coreos.com/browse/KNIP-821

clacroix12:
Is this for workers only or both master and worker nodes?

petr-balogh:
In the ticket that mentioned nodes where OCS is running, which are worker nodes if I understand correctly.

PR #697, opened 2019-09-05
Add produciton config for AWS IPI with m5.4xlarge

petr-balogh:
Fixes: #696 

This config will be used for production runs


petr-balogh:
I guess no need for verification as this is just simple config file which is not used anywhere yet.

vasukulkarni:
why not test it once with deployment to confirm?

vasukulkarni:
Verified locally ?

petr-balogh:
yep

PR #702, opened 2019-09-05
AWS UPI Deployment

shylesh:
PR for aws-upi deployment

### DIFF
--- a/conf/ocsci/aws_upi.yaml
+++ b/conf/ocsci/aws_upi.yaml
@@ -1,3 +1,4 @@
 ---
 ENV_DATA:
   deployment_type: 'upi'
+  rhcos_ami: 'ami-06c85f9d106577272'
--- a/conf/ocsci/aws_upi.yaml
+++ b/conf/ocsci/aws_upi.yaml
@@ -1,0 +1,7 @@
+---
+ENV_DATA:
+  platform: 'aws'
+  deployment_type: 'upi'
+  region: 'us-east-2'
+  skip_ocp_deployment: false
+  skip_ocs_deployment: false
--- a/conf/ocsci/aws_upi.yaml
+++ b/conf/ocsci/aws_upi.yaml
@@ -1,7 +1,3 @@
 ---
 ENV_DATA:
-  platform: 'aws'
   deployment_type: 'upi'
-  region: 'us-east-2'
-  skip_ocp_deployment: false
-  skip_ocs_deployment: false
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -291,6 +291,7 @@ class AWSUPI(AWSBase):
 
             sys.path.append(self.upi_script_path)
             full_path = os.getcwd()
+            logger.info("Executing UPI install script")
             proc = Popen(
                 [os.path.join(full_path, 'upi_on_aws-install.sh')],
                 stdout=PIPE, stderr=PIPE,
@@ -304,6 +305,8 @@ class AWSUPI(AWSBase):
                 logger.error(stderr)
                 raise exceptions.CommandFailed("upi install script failed")
             logger.info(stdout)
+
+            self.test_cluster()
 
     def deploy_ocp(self, log_cli_level='DEBUG'):
         """
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -2,25 +2,26 @@
 This module contains platform specific methods and classes for deployment
 on AWS platform
 """
-import boto3
 import json
 import logging
 import os
+import shutil
 import sys
 import traceback
-import shutil
 from subprocess import Popen, PIPE
+
+import boto3
 
 from ocs_ci.deployment.ocp import OCPDeployment as BaseOCPDeployment
 from ocs_ci.framework import config
 from ocs_ci.ocs import defaults, constants
+from ocs_ci.ocs import exceptions
 from ocs_ci.ocs.parallel import parallel
 from ocs_ci.utility.aws import AWS as AWSUtil
 from ocs_ci.ocs.exceptions import SameNamePrefixClusterAlreadyExistsException
 
 from ocs_ci.utility.utils import run_cmd, clone_repo
 from .deployment import Deployment
-from ocs_ci.ocs import exceptions
 
 logger = logging.getLogger(__name__)
 
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -5,7 +5,12 @@ on AWS platform
 import json
 import logging
 import os
+import sys
 import traceback
+import shutil
+from subprocess import Popen, PIPE
+
+import pytest
 
 from ocs_ci.deployment.ocp import OCPDeployment as BaseOCPDeployment
 from ocs_ci.framework import config
@@ -16,6 +21,7 @@ from ocs_ci.ocs.exceptions import SameNamePrefixClusterAlreadyExistsException
 
 from ocs_ci.utility.utils import run_cmd, clone_repo
 from .deployment import Deployment
+from ocs_ci.ocs import exceptions
 
 logger = logging.getLogger(__name__)
 
@@ -194,6 +200,24 @@ class AWSUPI(AWSBase):
     class OCPDeployment(BaseOCPDeployment):
         def __init__(self):
             super(AWSUPI.OCPDeployment, self).__init__()
+            self.upi_repo_path = os.path.join(
+                constants.EXTERNAL_DIR, 'openshift-misc',
+            )
+            self.upi_unique_path = os.path.join(
+                constants.EXTERNAL_DIR,
+                f"aws_upi_deploy_{config.RUN['run_id']}"
+            )
+            try:
+                os.mkdir(self.upi_unique_path)
+            except OSError:
+                logger.error("Already a cluster exists with same runid")
+                pytest.fail("Cluster with same runid exists")
+
+            self.upi_script_path = os.path.join(
+                self.upi_unique_path,
+                f'openshift-misc/v3-launch-templates/functionality-testing'
+                f'/aos-4_2/hosts/'
+            )
 
         def deploy_prereq(self):
             """
@@ -210,7 +234,7 @@ class AWSUPI(AWSBase):
                 'route53_domain_name': config.ENV_DATA['base_domain'],
                 'vm_type_masters': config.ENV_DATA['master_instance_type'],
                 'vm_type_workers': config.ENV_DATA['worker_instance_type'],
-                'num_workers': 3
+                'num_workers': "3"
             }
             for key, value in upi_env_vars.items():
                 os.environ[key] = value
@@ -221,8 +245,42 @@ class AWSUPI(AWSBase):
 
             # git clone repo from openshift-qe repo
             clone_repo(
-                constants.OCP_QE_MISC_REPO,
-                os.path.join(constants.EXTERNAL_DIR, 'openshift-misc')
+                constants.OCP_QE_MISC_REPO, self.upi_repo_path
+            )
+
+            # copy openshift-misc cloned repos to this run dir
+            shutil.copytree(
+                self.upi_repo_path,
+                os.path.join(self.upi_unique_path, "openshift-misc")
+            )
+
+            # create install-dir inside upi_script_path
+            os.mkdir(os.path.join(self.upi_script_path, "install-dir"))
+
+            # NOT A CLEAN APPROACH: copy openshift-install and oc binary to
+            # script path because upi script expectes it to be present in
+            # script dir
+            bindir = os.path.join(os.getcwd(), 'bin')
+            shutil.copy2(
+                os.path.join(bindir, 'openshift-install'),
+                self.upi_script_path,
+            )
+            shutil.copy2(
+                os.path.join(bindir, 'oc'), self.upi_script_path
+            )
+            # copy install-config.yaml from cluster_path to
+            # upi_script_path/install-dir
+            shutil.copy2(
+                os.path.join(self.cluster_path, "install-config.yaml"),
+                os.path.join(self.upi_script_path, "install-dir/")
+            )
+            # Remove current cluster_path dir and make it as symlink to
+            # upi_script_path/install-dir/ so that further operations
+            # operates transparently
+            shutil.rmtree(self.cluster_path, ignore_errors=True)
+            os.symlink(
+                os.path.join(self.upi_script_path, "install-dir/"),
+                self.cluster_path.rstrip('/'),
             )
 
         def deploy(self, log_cli_level='DEBUG'):
@@ -238,7 +296,37 @@ class AWSUPI(AWSBase):
                 f"Openshift-installer will be using loglevel:{log_cli_level}"
             )
 
-            #TODO: call "upi_on_aws-install.sh" here
+            # Invoke upi_on_aws-install.sh
+            cidir = os.getcwd()
+            logger.info("Changing CWD")
+            try:
+                os.chdir(self.upi_script_path)
+            except OSError:
+                logger.exception(
+                    f"Failed to change CWD to {self.upi_script_path} "
+                )
+            logger.info(f"CWD changed to {self.upi_script_path}")
+
+            with open("./upi_on_aws-install.sh", "r") as fd:
+                buf = fd.read()
+            data = buf.replace("openshift-qe-upi", "ocs-qe-upi")
+            with open("./upi_on_aws-install.sh", "w") as fd:
+                fd.write(data)
+
+            sys.path.append(self.upi_script_path)
+            proc = Popen(
+                [f'{self.upi_script_path}/upi_on_aws-install.sh'],
+                stdout=PIPE, stderr=PIPE,
+            )
+            stdout, stderr = proc.communicate()
+
+            # Change dir back to ocs-ci dir
+            os.chdir(cidir)
+
+            if proc.returncode:
+                logger.error(stderr)
+                raise exceptions.CommandFailed("upi install script failed")
+            logger.info(stdout)
 
     def deploy_ocp(self, log_cli_level='DEBUG'):
         """
@@ -250,6 +338,8 @@ class AWSUPI(AWSBase):
         """
         super(AWSUPI, self).deploy_ocp(log_cli_level)
         volume_size = config.ENV_DATA.get('DEFAULT_EBS_VOLUME_SIZE', 100)
+        # TODO: Implement add_volume for UPI
+        # existing function looks for terraform files
         self.add_volume(volume_size)
 
     def destroy_cluster(self, log_level="DEBUG"):
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -199,8 +199,9 @@ class AWSUPI(AWSBase):
     class OCPDeployment(BaseOCPDeployment):
         def __init__(self):
             super(AWSUPI.OCPDeployment, self).__init__()
+            upi_repo_name = f'openshift-misc-{config.RUN["run_id"]}'
             self.upi_repo_path = os.path.join(
-                os.path.abspath(self.cluster_path), 'openshift-misc',
+                constants.EXTERNAL_DIR, upi_repo_name,
             )
 
             self.upi_script_path = os.path.join(
@@ -238,17 +239,13 @@ class AWSUPI(AWSBase):
                 constants.OCP_QE_MISC_REPO, self.upi_repo_path
             )
 
-            # create install-dir inside upi_script_path
-            # cluster_path is 6 directories higher than the sym-link
-            # path must be relative to the UPI script location
-            relative_cluster_path = "../" * 6
-            os.symlink(
-                os.path.join(
-                    relative_cluster_path,
-                    os.path.basename(self.cluster_path.rstrip('/'))
-                ),
-                os.path.join(self.upi_script_path, "install-dir")
+            # Sym link install-dir to cluster_path
+            install_dir = os.path.join(self.upi_script_path, "install-dir")
+            absolute_cluster_path = os.path.abspath(self.cluster_path)
+            logger.info(
+                "Sym linking %s to %s", install_dir, absolute_cluster_path
             )
+            os.symlink(absolute_cluster_path, install_dir)
 
             # NOT A CLEAN APPROACH: copy openshift-install and oc binary to
             # script path because upi script expects it to be present in
@@ -311,6 +308,13 @@ class AWSUPI(AWSBase):
 
             self.test_cluster()
 
+            # Delete openshift-misc repository
+            logger.info(
+                "Removing openshift-misc directory located at %s",
+                self.upi_repo_path
+            )
+            shutil.rmtree(self.upi_repo_path)
+
     def deploy_ocp(self, log_cli_level='DEBUG'):
         """
         OCP deployment specific to AWS UPI
@@ -363,14 +367,6 @@ class AWSUPI(AWSBase):
             logger.info("Destroying stack: %s", stack_name)
             cf.delete_stack(StackName=stack_name)
             verify_stack_deleted(stack_name)
-
-        # Delete openshift-misc repository from cluster_path
-        openshift_misc_path = os.path.join(self.cluster_path, 'openshift-misc')
-        logger.info(
-            "Removing openshift-misc directory located at %s",
-            openshift_misc_path
-        )
-        shutil.rmtree(openshift_misc_path)
 
 
 def get_infra_id(cluster_path):
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -22,7 +22,7 @@ logger = logging.getLogger(__name__)
 
 # As of now only IPI
 # TODO: Introduce UPI once we have proper doc
-__all__ = ['AWSIPI']
+__all__ = ['AWSIPI', 'AWSUPI']
 
 
 class AWSBase(Deployment):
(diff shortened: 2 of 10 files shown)


ocs_ci/deployment/aws.py
```diff
@@ -134,18 +150,190 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+
+class AWSUPI(AWSBase):
+    """
+    A class to handle AWS UPI specific deployment
+    """
+    def __init__(self):
+        self.name = self.__class__.__name__
+        super(AWSUPI, self).__init__()
+
+    class OCPDeployment(BaseOCPDeployment):
```
vasukulkarni: why is the class inside UPI class?  why can't it be just methods for UPI and use OCP deployment as  composition? I am also failing to understand what the *isa* relation between the classes, can you docstring them clearly please. Thanks

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +150,190 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+
+            # setup necessary env variables
+            upi_env_vars = {
+                'INSTANCE_NAME_PREFIX': config.ENV_DATA['cluster_name'],
+                'AWS_REGION': config.ENV_DATA['region'],
+                'rhcos_ami': 'ami-06c85f9d106577272',
+                'route53_domain_name': config.ENV_DATA['base_domain'],
+                'vm_type_masters': config.ENV_DATA['master_instance_type'],
+                'vm_type_workers': config.ENV_DATA['worker_instance_type'],
+                'num_workers': "3"  # TODO: read in num_workers
```
vasukulkarni: is num_master supported?

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +150,190 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+            }
+            for key, value in upi_env_vars.items():
+                os.environ[key] = value
+
+            # ensure environment variables have been set correctly
+            for key, value in upi_env_vars.items():
+                assert os.getenv(key) == value, f"{os.getenv(key)} != {value}"
+
+            # git clone repo from openshift-qe repo
+            clone_repo(
```
vasukulkarni: is something clean this up?

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +150,190 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+                os.chdir(self.upi_script_path)
+            except OSError:
+                logger.exception(
+                    f"Failed to change CWD to {self.upi_script_path} "
                 )
-        except Exception:
-            logger.error(traceback.format_exc())
+            logger.info(f"CWD changed to {self.upi_script_path}")
+
+            with open("./upi_on_aws-install.sh", "r") as fd:
```
vasukulkarni: Appreciate if one of you can document in Mojo the high level process. Thanks

ocs_ci/deployment/aws.py
petr-balogh: This is implementation of the OCPDeployment class for specific platform/deployment type. If it's like that than we can in each type of deployment instantiate the class with the same name which is already done from the base class of the deployment.

It's documented a bit here: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/deployment.py#L46

Base class is here: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/ocp.py#L22

ocs_ci/deployment/aws.py
```diff
@@ -79,6 +79,22 @@ def add_node(self):
         # TODO: Implement later
         super(AWSBase, self).add_node()
 
+    def destroy_volumes(self):
```
petr-balogh: Hope that once we will use just operator we will not have to deal with the volumes at all.

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +150,190 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+    class OCPDeployment(BaseOCPDeployment):
+        def __init__(self):
+            super(AWSUPI.OCPDeployment, self).__init__()
+            self.upi_repo_path = os.path.join(
+                self.cluster_path, 'openshift-misc',
+            )
+
+            self.upi_script_path = os.path.join(
+                self.upi_repo_path,
+                f'v3-launch-templates/functionality-testing'
```
petr-balogh: why used formatting when you don't format anything here :)

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +150,190 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+            Overriding deploy_prereq from parent. Perform all necessary
+            prerequisites for AWSUPI here.
+            """
+            super(AWSUPI.OCPDeployment, self).deploy_prereq()
+
+            # setup necessary env variables
+            upi_env_vars = {
+                'INSTANCE_NAME_PREFIX': config.ENV_DATA['cluster_name'],
+                'AWS_REGION': config.ENV_DATA['region'],
+                'rhcos_ami': 'ami-06c85f9d106577272',
```
petr-balogh: Can we also parametrize AMI as it can probably change in the future?

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +150,190 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+            for key, value in upi_env_vars.items():
+                assert os.getenv(key) == value, f"{os.getenv(key)} != {value}"
+
+            # git clone repo from openshift-qe repo
+            clone_repo(
+                constants.OCP_QE_MISC_REPO, self.upi_repo_path
+            )
+
+            # create install-dir inside upi_script_path
+            relative_cluster_path = "../../../../../../"
```
petr-balogh: this is terrible, cannot we go from TOP_DIR from constants if it needs to be relative I would go from TOP_DIR of the osc-ci repo

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +150,190 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+            os.symlink(
```
petr-balogh: We need to make sure we have data available in NFS share, cause if it will be sym-linked there we won't see them under URL we provide for kubeconfig for example.

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +150,190 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+            )
+
+            # Invoke upi_on_aws-install.sh
+            cidir = os.getcwd()
+            logger.info("Changing CWD")
+            try:
+                os.chdir(self.upi_script_path)
+            except OSError:
+                logger.exception(
+                    f"Failed to change CWD to {self.upi_script_path} "
```
petr-balogh: I think that this is problematic to change the CWD and leads why the oc is not visible as we assume its from ./bin of cwd I guess. Also maybe leads to problems with KUBECONFIG if cluster dir is set relatively?

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +150,190 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+            with open("./upi_on_aws-install.sh", "r") as fd:
+                buf = fd.read()
+            data = buf.replace("openshift-qe-upi", "ocs-qe-upi")
+            with open("./upi_on_aws-install.sh", "w") as fd:
+                fd.write(data)
+
+            sys.path.append(self.upi_script_path)
+            full_path = os.getcwd()
+            proc = Popen(
+                [f'{full_path}/'
```
petr-balogh: cannot be this and next line be on one line?

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +150,190 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+            log_level (str): log level for openshift-installer (
+                default:DEBUG)
+        """
+        cluster_name = get_cluster_name(self.cluster_path)
+        self.destroy_volumes()
+        suffixes = ['vpc', 'inf', 'sg', 'bs', 'ma']
+        for i in range(3):  # TODO: read in num_workers
+            suffixes.append(f'no{i}')
+        stack_names = [f'{cluster_name}-{suffix}' for suffix in suffixes]
+        cf = boto3.client('cloudformation')
```
petr-balogh: Can we move this to ocs_ci/utility/aws.py and create function there like delete_claudformation_stack ?

ocs_ci/deployment/aws.py
clacroix12: We have a google doc currently, but we can translate to mojo if necessary. https://docs.google.com/document/d/1eFWMPhhku113-uH8NZgS6Uu_5s1jtRiECVR2CR7pXN0/edit?pli=1

ocs_ci/deployment/aws.py
clacroix12: The script expects `install-dir` to be a directory at the script location. Our framework expects the contents of `install-dir` to be at the `cluster_path`. This is trying to keep both of those two requirements happy. TOP_DIR doesn't help us here since this is all based on the provided `cluster_path` which may or may not be in our repo. 

I agree it's not ideal, but using TOP_DIR doesn't fix this. We need a relative path from the script location to the `cluster_path` here.

ocs_ci/deployment/aws.py
clacroix12: The repo is being cloned in the `cluster_path` which should be on the NFS share. The sym link is being used by the install script, but the true contents exist at the `cluster_path`

ocs_ci/deployment/aws.py
clacroix12: We are changing CWD because the script uses CWD to calculate different paths. We keep track of the initial CWD before we change, and are changing back to that original CWD after we execute the script.

ocs_ci/deployment/aws.py
clacroix12: We can, however this was intended to be temporary as we will be calling the script in the repo that removes the stacks.

ocs_ci/deployment/aws.py
clacroix12: Yes, we were going to move this to the config so it can be overridden.

ocs_ci/deployment/aws.py
clacroix12: It seems as though the cloudformation template dictates that there are 3 master nodes.

ocs_ci/deployment/aws.py
clacroix12: Not currently, however once destroy is called I don't see any reason to keep the repo around. I think we could remove it at that point.

ocs_ci/deployment/aws.py
clacroix12: We should use `os.path.join` here too instead of an f-string. `[os.path.join(full_path, 'upi_on_aws-install.sh')]`

ocs_ci/deployment/aws.py
petr-balogh: I see that it's needed by script but I giess it's the problem of why you don't have oc in yor path. So we should make this .bin dir exported to the path with its full path from  root /

clacroix12: Two issues:
1. It's not only about oc - the script calculates other variables based on cwd such as the `working_dir` and `script_dir` which other variables depend on.
2. The PATH used by the shell script will be the PATH as it was when **python** was called. Even if we alter the path in our code, it won't apply to the PATH that the shell script will use. This means that `oc` needs to be on the path BEFORE we execute `run-ci`, as well as the other dependencies like `jq` and `aws`.

ocs_ci/deployment/aws.py
vasukulkarni: OCP Deployment is different from OCS Deployment so they dont really have to inherit and should be kept as different classes, Also the use of inner classes seems like overkill now. we could just use simple classes and composition for better readability here and maintain relationship based on strict inheritance eg:  https://www.w3resource.com/java-tutorial/inheritance-composition-relationship.php

clacroix12: @vasukulkarni I agree, and think we can simplify these classes a lot by not having them nested. It's not something we are going to address in this PR, but i think it's something we should absolutely create an issue for and discuss.

ocs_ci/deployment/aws.py
clacroix12: Fixed

ocs_ci/deployment/aws.py
```diff
@@ -2,26 +2,30 @@
 This module contains platform specific methods and classes for deployment
 on AWS platform
 """
-import os
-import logging
+import boto3
```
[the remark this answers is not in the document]
clacroix12: Done

conf/ocsci/aws_upi.yaml
```diff
@@ -0,0 +1,7 @@
+---
+ENV_DATA:
+  platform: 'aws'
+  deployment_type: 'upi'
+  region: 'us-east-2'
```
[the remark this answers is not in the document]
clacroix12: Fixed

ocs_ci/deployment/aws.py
clacroix12: Done

ocs_ci/deployment/aws.py
vasukulkarni: @clacroix12 sure, I will try to add it in issues today or monday.

ocs_ci/deployment/aws.py
vasukulkarni: I will take this with Anjana. Thanks

ocs_ci/deployment/aws.py
shylesh: I guess number of master nodes are controlled by install_config.yaml. In aws_upi script they will remove all the worker node entries and then bring up only master initially. Later they will add worker based on num_workers passed to the aws_upi script.

ocs_ci/deployment/aws.py
clacroix12: Done

ocs_ci/deployment/aws.py
clacroix12: We are now deleting the repo after we destroy the cluster in teardown.

clacroix12:
Rebased on latest master

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +152,247 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+            verify_stack_deleted(stack_name)
+
+        # Call openshift-installer destroy cluster
+        super(AWSUPI, self).destroy_cluster(log_level)
+
+        # Delete inf and vpc stacks
+        suffixes = ['inf', 'vpc']
+        stack_names = [f'{cluster_name}-{suffix}' for suffix in suffixes]
+        for stack_name in stack_names:
+            logger.info("Destroying stack: %s", stack_name)
```
petr-balogh: Can you please check if it also delete s3 buckets? When I did today's cleanup I saw many of them are with UPI inside name, so worth to check if we don't miss something when destroying the cluster.

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +152,247 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+        cluster_path: path to cluster install directory
+
+    Returns:
+        str: metadata.json['infraID']
+
+    """
+    metadata_file = os.path.join(cluster_path, "metadata.json")
+    with open(metadata_file) as f:
+        metadata = json.loads(f.read())
+    return metadata.get("infraID")
```
petr-balogh: I would rather use direct access of the key with metadata['infraID'] and in case of some issue you will fail with the KeyError exception. If there is no such key you will return None which will lead to some confusion when debugging the issue.

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +152,247 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+        cluster_path: path to cluster install directory
+
+    Returns:
+        str: metadata.json['clusterName']
+
+    """
+    metadata_file = os.path.join(cluster_path, "metadata.json")
+    with open(metadata_file) as f:
+        metadata = json.loads(f.read())
+    return metadata.get("clusterName")
```
petr-balogh: same

ocs_ci/deployment/aws.py
```diff
@@ -258,12 +259,16 @@ def deploy(self, log_cli_level='DEBUG'):
             with open("./upi_on_aws-install.sh", "w") as fd:
                 fd.write(data)
 
-            sys.path.append(self.upi_script_path)
-            full_path = os.getcwd()
+            # ensure oc is on PATH for UPI install script
+            test_env = os.environ.copy()
+            path = test_env.get('PATH')
+            oc_path = os.path.join(constants.TOP_DIR, 'bin')
```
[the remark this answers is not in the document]
clacroix12: @petr-balogh if we do that we will have to move where we determine the `oc_path` to before we start changing the `cwd` to satisfy the install shell script. We can pull it from config and use the absolute path to cover the jenkins case, that's not a problem. We were just seeing issues with that previously because it is a relative path and we are changing the `cwd` to allow the shell script to work properly.

ocs_ci/deployment/aws.py
clacroix12: I think the buckets that are still in s3 are due leftovers from testing, and not from our successful deployments/teardowns. I will verify that this is the case and if it's not look into a fix.

ocs_ci/deployment/aws.py
clacroix12: Using the `metadata` attribute directly was causing issues, and this method just works.

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +152,247 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+    def deploy_ocp(self, log_cli_level='DEBUG'):
+        """
+        OCP deployment specific to AWS UPI
+
+        Args:
+             log_cli_level (str): openshift installer's log level
+                (default: 'DEBUG')
+        """
+        super(AWSUPI, self).deploy_ocp(log_cli_level)
+        volume_size = config.ENV_DATA.get('DEFAULT_EBS_VOLUME_SIZE', 100)
```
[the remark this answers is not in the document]
clacroix12: @petr-balogh I don't think that link is related to this code, but I'm pretty sure I see what you mean. We should be using this code now:
```
        if not self.ocs_operator_deployment:
            volume_size = int(
                config.ENV_DATA.get('device_size', defaults.DEVICE_SIZE)
            )
            self.add_volume(volume_size)
```
Correct?

ocs_ci/deployment/aws.py
clacroix12: Just ran a local test and no s3 buckets were left after teardown.

ocs_ci/deployment/aws.py
clacroix12: Since this is no longer temporary, I'd like to address this again. I don't see the value in creating a wrapper function around what is essentially two lines of code. Creating the client, then executing boto's `delete_stack` method.

I'd like to avoid creating unnecessary wrapper utilities since we tend to run into many issues around supporting every use case the underlying methods support natively. `run_cmd` being a prime example of this.

shylesh:
Rebased on latest master

ocs_ci/deployment/aws.py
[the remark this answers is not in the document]
clacroix12: @petr-balogh you're actually right, after adding the `bin_dir`'s absolute path to the PATH, I no longer needed to add a custom env to subprocess. This will be in the next update.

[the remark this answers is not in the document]
clacroix12: Done

ocs_ci/deployment/aws.py
clacroix12: Done

ocs_ci/deployment/aws.py
clacroix12: Done

ocs_ci/deployment/aws.py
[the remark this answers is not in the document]
petr-balogh: Sorry, probably bad copy paste, I edited the link to line I meant.  But you got it right

ocs_ci/deployment/aws.py
petr-balogh: cool, thx

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -72,6 +72,10 @@ ENV_DATA:
   platform: 'AWS'
   deployment_type: 'ipi'
   region: 'us-east-2'
+  base_domain: 'qe.rh-ocs.com'
+  master_instance_type: 'm4.xlarge'
+  worker_instance_type: 'm4.xlarge'
+  rhcos_ami: 'ami-06c85f9d106577272'
```
petr-balogh: can we move conf/ocsci/aws_upi.yaml

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +152,241 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+                    f"Failed to change CWD to {self.upi_script_path} "
                 )
-        except Exception:
-            logger.error(traceback.format_exc())
+            logger.info(f"CWD changed to {self.upi_script_path}")
+
+            with open("./upi_on_aws-install.sh", "r") as fd:
+                buf = fd.read()
+            data = buf.replace("openshift-qe-upi", "ocs-qe-upi")
+            with open("./upi_on_aws-install.sh", "w") as fd:
```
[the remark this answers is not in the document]
clacroix12: As per our conversation, we can use `git diff` to show any differences in the file since the clone.

ocs_ci/framework/conf/default_config.yaml
clacroix12: Done

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +152,241 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+
+            # setup necessary env variables
+            upi_env_vars = {
+                'INSTANCE_NAME_PREFIX': config.ENV_DATA['cluster_name'],
+                'AWS_REGION': config.ENV_DATA['region'],
+                'rhcos_ami': config.ENV_DATA['rhcos_ami'],
+                'route53_domain_name': config.ENV_DATA['base_domain'],
+                'vm_type_masters': config.ENV_DATA['master_instance_type'],
+                'vm_type_workers': config.ENV_DATA['worker_instance_type'],
+                'num_workers': "3"  # TODO: read in num_workers
```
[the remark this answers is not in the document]
clacroix12: Done

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +152,241 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+                os.chdir(self.upi_script_path)
+            except OSError:
+                logger.exception(
+                    f"Failed to change CWD to {self.upi_script_path} "
                 )
-        except Exception:
-            logger.error(traceback.format_exc())
+            logger.info(f"CWD changed to {self.upi_script_path}")
+
+            with open("./upi_on_aws-install.sh", "r") as fd:
```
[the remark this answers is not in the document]
clacroix12: Done

ocs_ci/deployment/aws.py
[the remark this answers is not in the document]
clacroix12: Done

shylesh:
Rebased on latest master

ocs_ci/ocs/constants.py
```diff
@@ -15,6 +15,7 @@
     os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 )
 TEMPLATE_DIR = os.path.join(TOP_DIR, "ocs_ci", "templates")
+EXTERNAL_DIR = os.path.join(TOP_DIR, "external")
```
dahorak: Is the `EXTERNAL_DIR` used anywhere? I would expect, that it will be used for the _openshift-misc_ repo, but it seems to be cloned to `<cluster_path>/openshift-misc`, which is quite problematic on Jenkins (_cluster_path_ is on NFS share, and there is a problem with cloning git repository on NFS share exported by nfs-ganesha - [RHBZ#1735480](https://bugzilla.redhat.com/show_bug.cgi?id=1735480))

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +152,245 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+    Args:
+        cluster_path: path to cluster install directory
+
+    Returns:
+        str: metadata.json['infraID']
+
+    """
+    metadata_file = os.path.join(cluster_path, "metadata.json")
+    with open(metadata_file) as f:
+        metadata = json.loads(f.read())
```
dahorak: I'm just curious, why we don't use `json.load()` method?
```suggestion
        metadata = json.load(f)
```

vasukulkarni:
Based on the age of this PR, I think its better we merge and take any comments and address in future PR.

@shylesh @clacroix12 is this ready for merge?  can you guys update the subject and remove wip?

clacroix12:
@vasukulkarni We have been rebasing on master regularly and most of the code is in a file untouched by others so I'm not really concerned about the age. I don't see the point in merging before this is working.

shylesh:
> Based on the age of this PR, I think its better we merge and take any comments and address in future PR.
> 
> @shylesh @clacroix12 is this ready for merge? can you guys update the subject and remove wip?

@vasukulkarni 
We just discovered an issue with Jenkins run where deployment is failing.

ocs_ci/deployment/aws.py
clacroix12: No good reason :) will test this out

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +152,245 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+    """
+    def __init__(self):
+        self.name = self.__class__.__name__
+        super(AWSUPI, self).__init__()
+
+    class OCPDeployment(BaseOCPDeployment):
+        def __init__(self):
+            super(AWSUPI.OCPDeployment, self).__init__()
+            self.upi_repo_path = os.path.join(
+                os.path.abspath(self.cluster_path), 'openshift-misc',
```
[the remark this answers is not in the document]
vasukulkarni: @dahorak what is the ENV variable that is specific to jenkins, I think its better to distinguish between jenkins and local run, either through one ENV or through cli.

ocs_ci/deployment/aws.py
clacroix12: Done

ocs_ci/ocs/constants.py
clacroix12: That was the original purpose of the external_dir, but we decided to go with cloning into the cluster_path to keep things together. We will go back to cloning into the external dir to work around this issue in jenkins.

ocs_ci/deployment/aws.py
```diff
@@ -2,26 +2,32 @@
...
-
+from ocs_ci.utility.retry import retry
+from ocs_ci.utility.utils import run_cmd, clone_repo
+from .deployment import Deployment
 
 logger = logging.getLogger(__name__)
 
 
 # As of now only IPI
 # TODO: Introduce UPI once we have proper doc
```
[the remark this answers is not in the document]
clacroix12: Done

ocs_ci/ocs/constants.py
clacroix12: We are now cloning the repo into the `external` dir, sym-linking the `install-dir` to `cluster_path`, and removing the repo after deployment is done.

ocs_ci/deployment/aws.py
```diff
@@ -134,18 +150,241 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+            absolute_cluster_path = os.path.abspath(self.cluster_path)
+            logger.info(
+                "Sym linking %s to %s", install_dir, absolute_cluster_path
+            )
+            os.symlink(absolute_cluster_path, install_dir)
+
+            # NOT A CLEAN APPROACH: copy openshift-install and oc binary to
+            # script path because upi script expects it to be present in
+            # script dir
+            bindir = os.path.join(os.getcwd(), 'bin')
```
dahorak: The `bindir` shouldn't be taken from current working directory (`os.getcwd()`) but from configuration parameter `RUN['bin_dir']`. Something like this:

`bin_dir = os.path.abspath(os.path.expanduser(config.RUN['bin_dir']))`

The `os.path.expanduser(...)` is important here, because on Jenkins it is set to:
```
RUN:
  bin_dir: "~/bin"
```

ocs_ci/deployment/aws.py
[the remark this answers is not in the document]
petr-balogh: Not sure I follow here. In the jenkins we are just using some configuration like we change the bin_dir or other path which user do not usually change and using the default values. Cluster path is pointing to NFS share and so on.. But otherwise I don't see much differences between jenkins/local runs. It's just about the configuration we provide in config file and if you run on jenkins you store data like cluster dir and logs to NFS share.

ocs_ci/deployment/aws.py
petr-balogh: +1 Yep, this one approach was mentioned already in some of my previous comment, maybe it was missed. But if I remember correctly I suggested to use this approach directly in config itself so everywhere we will read `RUN['bin_dir']` we will read already path which will be absolute with expanded user.

ocs_ci/deployment/aws.py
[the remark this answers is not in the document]
dahorak: I think, this was "resolved" [above](https://github.com/red-hat-storage/ocs-ci/pull/702#discussion_r326077070).

ocs_ci/deployment/aws.py
clacroix12: @petr-balogh Agreed, I think it's best if any paths coming from our config are absolute paths and not relative. This would solve issues like this from arising when we are doing something different in jenkins vs. local environments.

For now though, this has been updated to @dahorak 's suggestion.

clacroix12:
I have rebased on latest master. Due to the issues with the ocs-operator, a successful deployment of IPI hasn't been possible without passing a custom configuration. That being said, with the config below I was able to deploy AWS IPI successfully locally and have our deployment test pass. I think it's safe to merge this now and address any subsequent issues in a follow-up PR.

```
---
DEPLOYMENT:
  ocs_operator_image: "quay.io/rhceph-dev/ocs-registry:4.2-77.bed1df0.master"
  ocs_operator_deployment: True
  ocs_operator_storage_cluster_cr: "http://pkgs.devel.redhat.com/cgit/containers/ocs-registry/plain/ocs_v1_storagecluster_cr.yaml?h=ocs-4.2-rhel-8"
  ocs_operator_olm: "http://pkgs.devel.redhat.com/cgit/containers/ocs-registry/plain/deploy-with-olm.yaml?h=ocs-4.2-rhel-8"
```

PR #705, opened 2019-09-06
WIP - cleanup destroy cluster test

dahorak:
- do not try to destroy cluster twice in `test_destroy_cluster` test (once as part of the test and second time during teardown)
- remove old unused `destroy_cluster()` function
- doc update: suggest `test_destroy_cluster` (marked as `destroy`) for cluster teardown instead of `test_cluster_is_running` (marked as `deployment`). The later require properly running cluster, otherwise it will fail on the check, despite the fact that the cluster is properly destroyed during the teardown phase.

### DIFF
--- a/tests/ecosystem/destroy/test_destroy_cluster.py
+++ b/tests/ecosystem/destroy/test_destroy_cluster.py
@@ -1,8 +1,7 @@
 import logging
 
-from ocs_ci.framework import config
 from ocs_ci.framework.testlib import EcosystemTest, destroy
-from ocs_ci.utility.utils import destroy_cluster
+from ocs_ci.framework import config
 
 log = logging.getLogger(__name__)
 
@@ -10,5 +9,13 @@ log = logging.getLogger(__name__)
 @destroy
 class TestDestroy(EcosystemTest):
     def test_destroy_cluster(self, log_cli_level):
-        log.info("Running OCS cluster destroy")
-        destroy_cluster(config.ENV_DATA['cluster_path'], log_cli_level)
+        teardown = config.RUN['cli_params'].get('teardown')
+        if teardown:
+            log.info(
+                "Cluster will be destroyed during teardown part of this test."
+            )
+        else:
+            log.warning(
+                "Command line parameter --teardown was not provided, "
+                "cluster will not be destroyed!"
+            )
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -132,7 +132,7 @@ run-ci tests/
 
 If you would like to destroy existing cluster you can run following command:
 ```bash
-run-ci -m deployment \
+run-ci -m destroy \
     --cluster-name kerberos_ID-ocs-deployment \
     --cluster-path /home/my_user/my-ocs-dir tests/ \
     --teardown
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -7,7 +7,6 @@ import shlex
 import string
 import subprocess
 import time
-import traceback
 import requests
 import yaml
 import re
@@ -19,7 +18,6 @@ from ocs_ci.ocs.exceptions import (
     UnavailableBuildException,
 )
 from ocs_ci.framework import config
-from ocs_ci.utility.aws import AWS
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
 from ocs_ci.ocs import constants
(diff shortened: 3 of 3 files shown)


tests/ecosystem/destroy/test_destroy_cluster.py
```diff
@@ -1,14 +1,21 @@
...
 
 log = logging.getLogger(__name__)
 
 
 @destroy
 class TestDestroy(EcosystemTest):
     def test_destroy_cluster(self, log_cli_level):
-        log.info("Running OCS cluster destroy")
-        destroy_cluster(config.ENV_DATA['cluster_path'], log_cli_level)
+        teardown = config.RUN['cli_params']['teardown']
```
vasukulkarni: .get instead?

dahorak:
Example execution of test_destroy_cluster test without `--teardown` parameter -
see the warning message:

{{{
$ run-ci tests/ecosystem/ -m destroy --cluster-name test-cluster --cluster-path ../clusters/test-cluster
Collecting Cluster versions
=============================== test session starts ===============================
platform linux -- Python 3.7.4, pytest-5.1.2, py-1.8.0, pluggy-0.12.0
rootdir: /ocs-ci, inifile: pytest.ini
plugins: logger-0.5.1, reportportal-1.0.5, html-1.22.0, marker-bugzilla-0.9.1.dev2, metadata-1.8.0
collected 2 items / 1 deselected / 1 selected                                     

tests/ecosystem/destroy/test_destroy_cluster.py::TestDestroy::test_destroy_cluster 
--------------------------------- live log setup ----------------------------------
  <<truncated>>
---------------------------------- live log call ----------------------------------
08:42:44 - MainThread - tests.ecosystem.destroy.test_destroy_cluster - WARNING - Command line parameter --teardown was not provided, cluster will not be destroyed!
PASSED                                                                      [100%]

========================= 1 passed, 1 deselected in 2.12s =========================
}}}

dahorak:
> PR validation [Job FAILED](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/177//testReport) (installation failed, tests not executed).

The PR validation job failed on installation, but the important part is the respective [teardown job](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-destroy-ocs-cluster/661/console).

tests/ecosystem/destroy/test_destroy_cluster.py
```diff
@@ -1,14 +1,21 @@
...
 
 log = logging.getLogger(__name__)
 
 
 @destroy
 class TestDestroy(EcosystemTest):
     def test_destroy_cluster(self, log_cli_level):
-        log.info("Running OCS cluster destroy")
-        destroy_cluster(config.ENV_DATA['cluster_path'], log_cli_level)
+        teardown = config.RUN['cli_params'].get('teardown')
```
RazTamir: the cli param is 'destroy' or 'teardown'?

dahorak: @RazTamir the CLI param is `--teardown`[1] 

[1] https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/pytest_customization/ocscilib.py#L58,L64

RazTamir: so I think this https://github.com/red-hat-storage/ocs-ci/pull/705/files#diff-4be2078a125fe364f7198675d51bc688R135 should stay 'deployment' and not 'destroy'.
We are tearing down the deployment

dahorak: @RazTamir I understand your linguistic point here, the problem is, that with `deployment` it will launch the `test_cluster_is_running` test, which is quite ok for properly running cluster (the test passes and during teardown phase it will destroy the cluster), but for partially working cluster, the *test part* of the `test_cluster_is_running` fails and even though the cluster is properly destroyed in the teardown phase, the overall status is failed. This is not big problem for manual work, because it is easily visible, that the cluster was clearly destroyed and almost nobody is checking the return code, but it is real problem in CI, because it is not possible to distinguish if the cluster was properly destroyed and we can continue with the cleanup of the env. or not.

So I'm not sure, what would be the best solution here? Or would be enough to change the `destroy` marker to `teardown`, or on the other hand change the `--teardown` parameter to `--destroy`? The later one will be slightly bigger change affecting whole team, as everybody is using it.

RazTamir: ok.
Let's not complicate things now.
We can revisit this in the future

dahorak: Here is another Idea, how to do it and for me it seems to be the best and cleanest one:

- rename the `test_cluster_is_running` from `test_ocs_basic_install.py` to something like `test_deployment`, so it will not be directly connected only with installation (it will stay marked as `deployment` as it is now)
- The `test_deployment` will run the `assert is_cluster_running(...)` only if `--destroy` parameter will not be specified. So there will be no change in the behaviour for installation as it is now (except that the test will not be named `test_cluster_is_running` but `test_deployment`). But for teardown, it will not check the cluster availability and thus not fails on cluster which is not fully installed.

- Then we can completely remove the `test_destroy_cluster` test.

petr-balogh: Yep, sounds like good solution. ACK from my side to go with this approach.

RazTamir: +1,
looks much better

dahorak: Will close this PR and create new one with the new approach.

dahorak:
Closing this PR, will create new one as discussed here: https://github.com/red-hat-storage/ocs-ci/pull/705#discussion_r322686460

PR #711, opened 2019-09-06
Jenkins: Avoid using operator for downstream

zmc:


RazTamir:
Why we want to avoid using operator for downstream?

zmc:
@RazTamir According to @petr-balogh, it's not implemented yet.

petr-balogh:
> @RazTamir According to @petr-balogh, it's not implemented yet.

That's true. Just submitted PR #713 . Consider as late night submitted PR, so maybe will need some polish on Monday.

RazTamir:
Thanks.
LGTM

PR #715, opened 2019-09-08
Test for profiling RBD performance

keesturam:
I'm sending this WIP PR to seek feedback on the perf charts being generated.  
Link to perf report: https://docs.google.com/spreadsheets/d/1jqypI0I4Ym1lvjSKMEcWPw8mFgvgQ2cTcGFMq4u0TKs/edit#gid=0

Once the format is finalized, we can do similar profiling against all IO profiles and volume types. 

To do:
- Validate any performance deviation from baseline

### DIFF
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -275,11 +275,17 @@ class Pod(OCS):
                 constants.FIO_IO_RW_PARAMS_YAML
             )
             self.io_params['rwmixread'] = rw_ratio
-        else:
+        elif io_direction == 'wo':
             self.io_params = templating.load_yaml(
                 constants.FIO_IO_PARAMS_YAML
             )
+        elif io_direction == 'ro':
+            self.io_params = templating.load_yaml(
+                constants.FIO_IO_PARAMS_YAML
+            )
+            self.io_params['rw'] = 'randread'
         self.io_params['runtime'] = runtime
+
         size = size if isinstance(size, str) else f"{size}G"
         self.io_params['size'] = size
         if fio_filename:
--- a/ocs_ci/utility/spreadsheet/spreadsheet_api.py
+++ b/ocs_ci/utility/spreadsheet/spreadsheet_api.py
@@ -28,18 +28,45 @@ class GoogleSpreadSheetAPI(object):
         client = gspread.authorize(creds)
         self.sheet = client.open(sheet_name).get_worksheet(sheet_index)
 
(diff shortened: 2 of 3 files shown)


ocs_ci/ocs/resources/pod.py
```diff
@@ -198,20 +184,15 @@ def exec_ceph_cmd(self, ceph_cmd, format='json-pretty'):
             return [item for item in out if item]
         return out
 
-    def get_storage_path(self, storage_type='fs'):
+    def get_mount_path(self):
```
ebenahar: why have you removed the option to return the block device path? This is being used in some tests

keesturam: I had tried this out and seem to have sent it without reverting to the original function. I'll fix this.

tests/e2e/test_performance_fio_randread_profile.py
```diff
@@ -0,0 +1,105 @@
...
+                *['10k', 'ro', 2, 300, 4, 1]
+            )
+        ]
+    )
+    @pytest.mark.usefixtures(base_setup.__name__)
+    def test_perf_read_latency(
+        self, size, io_direction, jobs, runtime, depth, sheet_index
+    ):
+        """
+        Test IO
```
SvetlanaRedHat: Function does not have enough description,i see it does a lot  of actions and it  will be better to have description and pass and fail criteria's

ocs_ci/utility/spreadsheet/spreadsheet_api.py
```diff
@@ -28,18 +28,45 @@ def __init__(self, sheet_name, sheet_index):
...
+        """
+        Updates the sheet name
+        """
+        return self.sheet.update_title(title)
+
+    def append_row(self, value):
+        """
+        Appends a row with content
+        """
+        return self.sheet.append_row(value)
```
RazTamir: I think that insert row is doing append of a row

tests/e2e/test_performance_fio_randread_profile.py
keesturam: I've added a description now. These tests as of today just generate graphs for the IO profiles. I'll introduce a check to fail the test once we have a benchmark to compare. I've added a todo for now.

ebenahar:
@keesturam Would like to close old PRs, is this one still relevant?

keesturam:
Closing this PR. We now use Ripsaw with FIO to run the workload.

PR #718, opened 2019-09-09
CMG OC and CLI tests

MeridianExplorer:


ocs_ci/ocs/resources/mcg.py
```diff
@@ -63,15 +64,59 @@ def s3_create_bucket(self, bucketname, region=config.ENV_DATA['region']):
             CreateBucketConfiguration={
                 'LocationConstraint': region
             }
-        )
+        ).name
+    s3_create_obc = s3_create_bucket
+
+    def oc_create_obc(self, bucketname):
```
ebenahar: Since most of the methods in this class get bucketname, would it make sense to instantiate the class with a bucketname?

ocs_ci/ocs/resources/mcg.py
```diff
@@ -63,15 +64,59 @@ def s3_create_bucket(self, bucketname, region=config.ENV_DATA['region']):
...
+        # TODO: Implement
+        logger.info(f"Deleting bucket: {bucketname}")
+        raise NotImplementedError()
+
+    def cli_delete_obc(self, bucketname):
+        """
+        Args:
+            bucketname: Name of bucket to be deleted
+
+        Returns:
```
ebenahar: Please delete 'Returns'

ocs_ci/ocs/resources/mcg.py
```diff
@@ -96,38 +141,20 @@ def s3_get_all_buckets(self):
         """
         return [bucket for bucket in self.s3_resource.buckets.all()]
 
-    def s3_verify_bucket_exists(self, bucket):
+    def s3_verify_bucket_exists(self, bucketname):
         """
         Verifies whether the Bucket exists
         Args:
-            bucket(S3 object) : The bucket object to be verified
+            bucketname(S3 object) : The bucket object to be verified
```
ebenahar: Isn't 'bucketname' a string?

tests/manage/mcg/conftest.py
```diff
@@ -60,38 +60,39 @@ def bucket_factory(request, mcg_obj):
...
-            if bucket.name in all_existing_buckets:
-                logger.info(f'Deleting bucket {bucket.name}')
-                bucket.object_versions.delete()
-                mcg_obj.s3_delete_bucket(bucket)
-                logger.info(
-                    f"Verifying whether bucket: {bucket.name} exists after deletion"
-                )
-                assert not mcg_obj.s3_verify_bucket_exists(bucket)
+        for bucket_name in set(created_bucket_names).intersection(all_existing_buckets):
+            logger.info(f'Cleanin up bucket {bucket_name}')
```
ebenahar: Cleaning

ocs_ci/utility/utils.py
```diff
@@ -1188,3 +1204,7 @@ def get_rook_repo(branch='master', to_checkout=None):
     run_cmd(f"git reset --hard origin/{branch}", cwd=cwd)
     if to_checkout:
         run_cmd(f"git checkout {to_checkout}", cwd=cwd)
+
+
+def check_if_executable_in_path(exec_name):
```
ebenahar: Impressive! (1)
Please add docstring

tests/manage/mcg/conftest.py
```diff
@@ -60,38 +60,39 @@ def bucket_factory(request, mcg_obj):
     """
     created_bucket_names = []
 
-    def _create_buckets(amount=1):
+    def _create_buckets(amount=1, interface='S3'):
```
ebenahar: Nice abstraction

ocs_ci/ocs/resources/mcg.py
MeridianExplorer: I agree with the premise, but it's impossible to require it - since you can create multiple buckets with different names under the same MCG.
Maybe it'll be better to move all those functions to a third, `mcg_bucket` class?

tests/manage/mcg/conftest.py
MeridianExplorer: TYVM :)

ocs_ci/ocs/resources/mcg.py
ebenahar: Having an mcg_bucket class makes sense. 
Let's consider this (doesn't have to be implemented in this PR)

tests/manage/mcg/test_bucket_deletion.py
```diff
@@ -4,31 +4,50 @@
...
+    # TODO: remove skipif
     @pytest.mark.polarion_id("OCS-1299")
     def test_s3_bucket_delete(self, mcg_obj, bucket_factory):
         """
         Test deletion of bucket using the S3 SDK
         """
+        for bucketname in bucket_factory(3, 'S3'):
+            logger.info(f"Deleting bucket: {bucketname}")
+            mcg_obj.s3_delete_bucket(bucketname)
+            assert mcg_obj.s3_verify_bucket_exists(bucketname) is False, \
```
ebenahar: Please don't break lines with '\'
Also, use 'assert not..' instead of 'assert... is False'
In this case, you can do:
```python
assert not mcg_obj.s3_verify_bucket_exists(bucketname), (
    f"Found {bucketname} that should've been removed"
)
```

tests/manage/mcg/test_bucket_deletion.py
```diff
@@ -4,31 +4,50 @@
...
-            mcg_obj.s3_delete_bucket(bucket)
-            assert mcg_obj.s3_verify_bucket_exists(bucket) is False
+    @pytest.mark.skipif(condition=check_if_executable_in_path('noobaa') is False, reason='MCG CLI was not found')
+    def test_cli_bucket_delete(self, mcg_obj, bucket_factory):
+        """
+        Test deletion of buckets using the MCG CLI
+        """
+        for bucketname in bucket_factory(3, 'CLI'):
+            mcg_obj.cli_delete_obc(bucketname)
+            assert mcg_obj.cli_verify_bucket_exists(bucketname) is False, \
```
ebenahar: The same

ocs_ci/ocs/resources/mcg.py
```diff
@@ -48,38 +50,31 @@ def __init__(self):
...
-        Args:
-            bucket: The bucket object to be deleted
+        Returns:
+            list: A list of all bucket names
 
         """
-        bucket.delete()
+        return [bucket.get('spec').get('bucketName')
+                for bucket
+                in OCP(namespace='openshift-storage', kind='obc').get().get('items')]
```
RazTamir: Please optimize this call.
OCP(namespace='openshift-storage', kind='obc').get().get('items')] should be called once (1 call to api) and list comprehension on the returned list

ocs_ci/ocs/resources/mcg.py
```diff
@@ -48,38 +50,31 @@ def __init__(self):
...
 
-    def s3_list_all_bucket_names(self):
+    def cli_list_all_bucket_names(self):
         """
         Returns:
             list: A list of all bucket names
 
         """
-        return [bucket.name for bucket in self.s3_resource.buckets.all()]
+        return [row.split()[1] for row in run_mcg_cmd('obc list').split('\n')[1:-1]]
```
RazTamir: Same,
You are calling a command every iteration which returns the same value

ocs_ci/ocs/resources/mcg.py
```diff
@@ -96,38 +91,52 @@ def s3_get_all_buckets(self):
...
-        raise NotImplementedError()
+        try:
+            OCP(namespace='openshift-storage', kind='obc').get(bucketname)
+            logger.info(f"{bucketname} exists")
+            return True
+        except CommandFailed as e:
+            if 'NotFound' in repr(e):
+                logger.info(f"{bucketname} does not exist")
+                return False
+            raise e
```
RazTamir: raise without 'e' is enough

ocs_ci/ocs/resources/mcg_bucket.py
```diff
@@ -0,0 +1,78 @@
...
+
+class OCBucket(MCGBucket):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        obc_data = templating.load_yaml_to_dict(constants.MCG_OBC_YAML)
+        if self.name is None:
+            self.name = create_unique_resource_name('oc', 'obc')
+        obc_data['metadata']['name'] = self.name
+        obc_data['spec']['bucketName'] = self.name
+        create_resource(**obc_data)
```
RazTamir: You can keep the resource here - self.data = create_resource()

ocs_ci/ocs/resources/mcg_bucket.py
```diff
@@ -0,0 +1,78 @@
...
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.ocp import OCP
+from ocs_ci.utility import templating
+from ocs_ci.utility.utils import run_mcg_cmd
+from tests.helpers import create_unique_resource_name, create_resource
+
+logger = logging.getLogger(name=__file__)
+
+
+class MCGBucket(ABC):
```
RazTamir: Why not inheriting from OCS to get all the methods come with it?

tests/manage/mcg/test_bucket_creation.py
```diff
@@ -4,30 +4,54 @@
...
         """
-
         assert set(
-            bucket.name for bucket in bucket_factory(3)
+            bucket.name for bucket in bucket_factory(3, 'S3')
         ).issubset(
             mcg_obj.s3_list_all_bucket_names()
         )
+
+    @pytest.mark.skipif(
```
petr-balogh: Can you specify the skipif marker here: 

https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/pytest_customization/marks.py#L55

with name nooba_cli_required or something like that? If you will have more such tests you can just reuse the same skipif marker imported from there.

tests/manage/mcg/test_bucket_creation.py
```diff
@@ -4,30 +4,54 @@
...
+        condition=check_if_executable_in_path('noobaa') is False,
```
petr-balogh: `not check_if_executable_in_path('noobaa')`

tests/manage/mcg/test_bucket_deletion.py
```diff
@@ -4,31 +4,54 @@
...
+from ocs_ci.utility.utils import check_if_executable_in_path
 
 logger = logging.getLogger(__name__)
 
 
-@pytest.mark.skipif(condition=True, reason="MCG is not deployed")
 @pytest.mark.filterwarnings(
     'ignore::urllib3.exceptions.InsecureRequestWarning'
 )
+@pytest.mark.skipif(
```
petr-balogh: define also marker form aws_platform_required and reuse it here and I see already was used somewhere above once.

ocs_ci/ocs/resources/mcg_bucket.py
```diff
@@ -0,0 +1,78 @@
...
+    def delete(self):
+        logger.info(f"Deleting bucket: {self.name}")
+        self.internal_delete()
+
+    @abstractmethod
+    def internal_delete(self):
+        pass
+
+
+class S3Bucket(MCGBucket):
```
vasukulkarni: missing docstring, same for other class below.

ocs_ci/ocs/resources/mcg_bucket.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.mcg.s3_resource.create_bucket(Bucket=self.name)
+
+    def internal_delete(self):
+        self.mcg.s3_resource.Bucket(self.name).object_versions.delete()
+        self.mcg.s3_resource.Bucket(self.name).delete()
+
+    def list_all_buckets(self):
```
vasukulkarni: docstring

ocs_ci/utility/utils.py
```diff
@@ -408,6 +410,21 @@ def run_cmd(cmd, secrets=None, **kwargs):
     return mask_secrets(r.stdout.decode(), secrets)
 
 
+def run_mcg_cmd(cmd, namespace='openshift-storage'):
```
vasukulkarni: can you enhance the run_cmd to handle noobaa cmds too, thanks

ocs_ci/ocs/resources/mcg_bucket.py
MeridianExplorer: I considered this option and preferred to leave it out - considering that the CLI doesn't return an object of any kind, and that S3 returns a different kind of object, I worried it might cause more confusion than help.

Do you think it's preferable to add a `self.data` field and implement if for S3 and OC?

ocs_ci/ocs/resources/mcg_bucket.py
MeridianExplorer: Seeing as it's not relevant in the S3 and CLI cases (and it created a clash with the existing `self.name` that can't be refactored easily), and due to the simplicity of the class, I chose to not inherit it. I also don't see many immediate uses to inheriting it - would be glad to have some examples if it's critical, might've missed something :)

ocs_ci/utility/utils.py
MeridianExplorer: Handle in what sense? :)

vasukulkarni: yeah if you can enhance `run_cmd` function to handle noobaa as well.

petr-balogh: I disagree here, the run_cmd is the general one function to run whatever command and shouldn't be bound with nooba or whatever else.  This function is just wrapper which will automatically pass the nooba command at the start of cmd + namespace.

So I would keep this in separate wrapper function like it is.

petr-balogh: namespace=None

and in the function please use:

namespace = nameaspace if namespace else config.ENV_DATA['cluster_namespace']

MeridianExplorer: I agree with Petr. I think adding complexity to `run_cmd` might cause unexpected complications in the future, and it means I'll have to add some sort of an MCG parameter to each and every function call (in which case it'd still be safer and easier to call `run_cmd('noobaa...')`

ocs_ci/ocs/resources/mcg_bucket.py
RazTamir: ack

ocs_ci/ocs/resources/mcg_bucket.py
RazTamir: I think that eventually, the desired state is to handle object exactly like file and block (from the testing perspective of course) so if we will follow the same guidelines for S3, MCG, object, etc. it will be much easier to onboard more ppl to the object storage world and help with the testing

MeridianExplorer: I agree that it'd be best to make the object storage world as accessible and straight-forward as possible - yet I worry that making `self.data` accessible will do the opposite.
If the bucket is an S3Bucket - `self.data` would contain an s3.Bucket object,
In case of a CLIBucket - `self.data` wouldn't contain anything (there is no inherent data rather than the name which has its own field)
And in case of an OCBucket, `self.data` would contain an OCP object.
Since the API is unified behind the scenes, and the user is never even exposed to the different kinds of buckets, I think that it'll be very confusing to hold different kinds of objects in `self.data` while on the surface the buckets all appear the same. (That is, for the user, all buckets are just 'MCGBuckets'. For us, they can be more specific like S3, OC or CLI, but that's strictly behind the scenes.)
In addition, I didn't run into any issues that required exposing the data, thus I didn't implement it yet.

Would be glad to have a deeper discussion regarding this

ocs_ci/utility/utils.py
vasukulkarni: The only difference right now I see is - it appends noobaa, I dont have a strong preference here *as long as it does the same and we dont duplicate code*, if its part of run_cmd, the duplication can never happen.

petr-balogh: this function is wrapping run_cmd, it adds noobaa to the cmd + it also add namespace which is related to that command. So for usage it will be easier to use this one wrapper function in my opinion.

And probably later on add some more functionality/parameters here.  I don't like the idea to put any other command specific stuff to general run_cmd function as it is suppose to run whatever command. So if Vasu is not strongly against this I would resolve this conversation if @vasukulkarni  is ok with this.

vasukulkarni: Resolving this, I can wait and watch how much noobaa cli differs.

MeridianExplorer:
Reopening after mistaken close

PR #722, opened 2019-09-10
Move Ceph to v14.2.3-20190904 in default config

petr-balogh:


petr-balogh:
This is just for old type of deployment. The operator deployment should handle this automatically by CSV

clacroix12:
Does this require deployment testing at least?

PR #713, opened 2019-09-06
Deploy via downstream ocs operator

petr-balogh:
Fixes: #712 

Implement missing functionality for DS OCS operator deployment.

PR #726, opened 2019-09-10
Change name of load_yaml_to_dict method to load_yaml

petr-balogh:
This was requested by @vasukulkarni to move this commit from PR #713 .

Changes the name of the method load_yaml_to_dict to just load_yaml.
This is because the method doesn't return just the dict but also generator in case of multi_document=True passed.

### DIFF
--- a/ocs_ci/utility/templating.py
+++ b/ocs_ci/utility/templating.py
@@ -136,7 +136,7 @@ def dump_to_temp_yaml(src_file, dst_file, **kwargs):
         yaml.dump(data, yaml_file)
 
 
-def load_yaml_to_dict(file, multi_document=False):
+def load_yaml(file, multi_document=False):
     """
     Load yaml file (local or from URL) and convert it to dictionary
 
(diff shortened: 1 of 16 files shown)


petr-balogh:
Discussion:  https://github.com/red-hat-storage/ocs-ci/pull/713#issuecomment-529606200

ocs_ci/utility/templating.py
```diff
@@ -136,7 +136,7 @@ def dump_to_temp_yaml(src_file, dst_file, **kwargs):
         yaml.dump(data, yaml_file)
 
 
-def load_yaml_to_dict(file, multi_document=False):
+def load_yaml(file, multi_document=False):
```
vasukulkarni: the name sounds like synonym for 'yaml.load' , I think we should use better internal name that represents it returning dict or dict generator, eg: `load_yaml_to_dict_gen`, Frankly speaking the current name doesnt have to change , there isn't that much confusion.

vasukulkarni: If we are not using jinja2 templating now, maybe we could revisit this with yaml.load() as is.

petr-balogh:
verification job here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/856/

petr-balogh:
Was just rebased and ran the verification. So the approve from Vasu was removed but I am consider it as approved as no change made and no error in test results is related to this change. I will send the email about this change.

Issue #687, opened 2019-09-04
WA for BZ 1747388 with WARN too few PGs


petr-balogh:
We need to create the WA and once it will be fixed we need to remove this WA. Please don't close this issue till the WA is deleted.

mbukatov:
Upstream references:
* https://github.com/rook/rook/issues/560
* https://github.com/rook/rook/pull/3745
* https://github.com/rook/rook/pull/3769

vasukulkarni:
Can be closed now?

petr-balogh:
yes, I thought it will be closed automatically as I mentioned this issue in PR message.

PR #727, opened 2019-09-10
Revert "workaround for BZ 1747388 with WARN too few PGs"

petr-balogh:
This reverts commit 250b6ec09be8dc62c6057e32d8b269ad23f27cf9.

Fixes issue: #687

The commit which is fixing it in the operator is this:
https://github.com/openshift/ocs-operator/commit/8c491c23972beab6088d7301397365ebb965887e

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -15,7 +15,6 @@ from ocs_ci.utility.utils import (
 )
 from ocs_ci.ocs.exceptions import CommandFailed, UnavailableResourceException
 from ocs_ci.ocs import constants, ocp, defaults
-from ocs_ci.ocs.cluster import CephCluster
 from ocs_ci.ocs.resources.ocs import OCS
 from ocs_ci.ocs.resources.csv import CSV
 from tests import helpers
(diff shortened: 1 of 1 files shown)


petr-balogh:
Looks like verification passed without any interaction.

I've just connected to the env to check the real status:
```
oc -n openshift-storage rsh rook-ceph-tools-5f5dc75fd5-255t9
sh-4.2# ceph health
HEALTH_OK
sh-4.2# ceph status
  cluster:
    id:     a3a5e0fa-fe7c-46d6-8e51-14e4bef19249
    health: HEALTH_OK

  services:
    mon: 3 daemons, quorum a,b,c (age 9m)
    mgr: a(active, since 9m)
    mds: rook-ceph-cephfilesystem:1 {0=rook-ceph-cephfilesystem-a=up:active} 1 up:standby-replay
    osd: 3 osds: 3 up (since 7m), 3 in (since 7m)

  data:
    pools:   4 pools, 32 pgs
    objects: 31 objects, 2.3 KiB
    usage:   3.0 GiB used, 294 GiB / 297 GiB avail
    pgs:     32 active+clean

  io:
    client:   853 B/s rd, 1 op/s rd, 0 op/s wr

sh-4.2# ceph osd pool ls
rook-ceph-cephblockpool
rook-ceph-cephfilesystem-metadata
rook-ceph-cephfilesystem-data0
.rgw.root
```

but when I connected I saw it's still pg_num 8 but see there are 4 pools so I think that rgw.root is created now by operator as they told me that now it should create all default storage classes, but din't check that as the env got destroyed right after.

PR #728, opened 2019-09-10
Implemention of OCS-278 Test Case

hadarsharon:
Implemented test case OCS-278 (tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py)

The purpose of this test case is to create at least 2 projects:
One project with just pvcs and the other with pvcs attached to a pod,
And then delete both projects entirely (along with their pvcs)

This might be a work in progress as the test is implemented sequentially according to the steps, but probably could be better encapsulated within functions or fixtures.


tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,165 @@
...
+
+        log.info("Creating Project 1")
+        project_1 = project_factory()
+
+        log.info("Creating {} CephFS PVCs".format(self.cephfs_pvcs_num))
+        # Generate a given number of CephFS PVCs, randomly assigned RWO or RWX
+        cephfs_pvcs = [pvc_factory(interface=constants.CEPHFILESYSTEM,
+                                   project=project_1,
+                                   storageclass=cephfs_sc,
+                                   access_mode=random.choice(
```
mykaul: Instead of this random call, randomize the number of PVs between 0 and self.cephfs_pvcs_num that will use RWO, and the rest will use RWX. it's a single call to random.choice().

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,165 @@
...
+from ocs_ci.ocs.resources import pod
+from ocs_ci.ocs.resources.ocs import OCS
+from ocs_ci.utility.retry import retry
+from tests import helpers
+
+log = logging.getLogger(__name__)
+
+
+@retry(UnexpectedBehaviour, tries=10, delay=3, backoff=1)
+def check_ceph_used_space():
```
mykaul: This looks like an infra material, not just for this test?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,165 @@
...
+    )
+
+@tier1
+@pytest.mark.polarion_id("OCS-278")
+class TestDeleteProjectWithPVCAndPods(ManageTest):
+    """
+    Create at least 2 projects, one with just pvcs and the other with pvcs
+    attached to pod, and delete them.
+    """
+    cephfs_pvcs_num = 100
```
mykaul: Read from ENV variables perhaps? We may want in sanity to use 10, on nightly to use 100, on scale to use 1000?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,165 @@
...
+            "Project 1 deletion time: {} seconds".format(project_deletion_time))
+        # PVs delete might still be running, check until all PVs are deleted
+        # Check time between first PV delete and last PV delete
+        time_between_pvcs_deletion_success = 0
+        # Also check delete time per each PV
+        individual_pv_delete_times = []
+        for pv in pvs:
+            pv_start_time = time.time()
+            pv_name = pv.backed_pv
+            #pv.delete()
```
mykaul: ?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,165 @@
...
+            pv.ocp.wait_for_delete(pv_name)
+            helpers.validate_pv_delete(pv_name)
+            pv_delete_time = time.time() - pv_start_time
+            individual_pv_delete_times.append({pv_name: pv_delete_time})
+            time_between_pvcs_deletion_success += pv_delete_time
+        log.info("{} seconds between first and last PVC deletion".format(
+            time_between_pvcs_deletion_success))
+        # Verify space has been reclaimed
+        space_used_after_deletion = check_ceph_used_space()
```
mykaul: If we delete lazily (and we will), this is not going to work well.

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,165 @@
...
+    attached to pod, and delete them.
+    """
+    cephfs_pvcs_num = 100
+    rbd_pvcs_num = 100
+
+    def test_delete_project_with_pvc_and_pods(self, project_factory,
+                                              storageclass_factory, pvc_factory,
+                                              multi_pvc_factory, pod_factory):
+        # Start with Project 1
+        log.info("Creating OCP+OCS Setup")
```
RazTamir: This block code L53-L55 should be in setup fixture.

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,165 @@
...
+        assert space_used_after_deletion < space_used_before_deletion
+        log.info("Space reclaimed successfully.")
+
+        # Project 1 done, moving on to Project 2
+        log.info("Creating Project 2")
+        project_2 = project_factory()
+
+        log.info("Creating {} CephFS PVCs "
+                 "(each bound to an app pod)".format(self.cephfs_pvcs_num))
+        cephfs_pvcs_pods = \
```
RazTamir:
```
cephfs_pvcs_pods = [
    ...
]
```

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,165 @@
...
+        rbd_pvcs_pods = \
+            [pod_factory(pvc=pvc_obj, interface=constants.CEPHBLOCKPOOL)
+             for pvc_obj in multi_pvc_factory(interface=constants.CEPHBLOCKPOOL,
+                                              project=project_2,
+                                              num_of_pvc=self.rbd_pvcs_num)]
+        # Delete the entire Project 2 (along with all of its PVCs)
+        pvs = helpers.get_all_pvs()
+        space_used_before_deletion = check_ceph_used_space()
+        start_time = time.time()
+        # ocp.switch_to_default_rook_cluster_project() #TODO: Is this needed?
```
RazTamir: yes, you will need to switch back to the default project

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: OK.

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: No problem, will do.

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: Will be removed now

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: OK - I will implement it as a single call outside of the list comprehension

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: Would using pytest.parametrize be a feasible approach?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: OK.

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: Does it mean step 3c in this test case ("Check space is reclaimed") is redundant?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
ebenahar: Indeed. This function could be useful for other tests. Could you please move it to be a method in the CephCluster class here in ocs_ci.ocs.cluster.CephCluster?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
ebenahar: @hadarsharon it will.
You can mark the 100 case with pytest.param(100, marks=pytest.mark.tier1)
and the 1000 one with pytest.param(100, marks=pytest.mark.scale)

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,174 @@
...
+
+@retry(UnexpectedBehaviour, tries=10, delay=3, backoff=1)
+def check_ceph_used_space():
+    """
+    Check for the used space in cluster
+    """
+    ct_pod = pod.get_ceph_tools_pod()
+    ceph_status = ct_pod.exec_ceph_cmd(ceph_cmd="ceph status")
+    assert ceph_status is not None
+    used = ceph_status['pgmap']['bytes_used']
```
ebenahar: Replace with:
```python
ceph_status.get('pgmap').get('bytes_used')
```

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+@retry(UnexpectedBehaviour, tries=10, delay=3, backoff=1)
+def check_ceph_used_space():
+    """
+    Check for the used space in cluster
+    """
+    ct_pod = pod.get_ceph_tools_pod()
+    ceph_status = ct_pod.exec_ceph_cmd(ceph_cmd="ceph status")
+    assert ceph_status is not None
```
ebenahar: You can do:
```python
assert ceph status
```

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,174 @@
...
+    used = ceph_status['pgmap']['bytes_used']
+    used_in_gb = used / constants.GB
+    global used_space
```
ebenahar: please have the used_space as a return value instead of using global

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,174 @@
...
+    rbd_pvcs_num = 100
+
+    @pytest.fixture()
+    def setup(self, storageclass_factory):
+        """
+        Create an OCP+OCS Setup, as well as two Storage classes:
+        one for CephFS and one for RBD.
+        Scale the setup to at-least 800 pvcs/pods.
+        """
+        log.info("Creating OCP+OCS Setup")
```
ebenahar: this logging is not needed as you're just initializing test objects

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,174 @@
...
+        self.ocp_setup = ocp.OCP()
+        self.ocs_setup = OCS()  # TODO: Scale the setup to at-least 800 pvcs/pods
+
+        log.info("Creating CephFS Storage class")
+        self.cephfs_sc = storageclass_factory(interface=constants.CEPHFILESYSTEM)
+        log.info("Creating RBD Storage class")
+        self.rbd_sc = storageclass_factory(interface=constants.CEPHBLOCKPOOL)
+
+    def test_delete_project_with_pvc_and_pods(self, project_factory,
```
ebenahar: please fix parentheses according to convention:

```python
def test_delete_project_with_pvc_and_pods(
    self, project_factory, pvc_factory, 
    multi_pvc_factory, pod_factory
)
```

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,174 @@
...
+                                              pvc_factory, multi_pvc_factory,
+                                              pod_factory):
+        # Start with Project 1
+        log.info("Creating Project 1")
+        project_1 = project_factory()
```
ebenahar: isn't project creation part of the test setup?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,174 @@
...
+
+        log.info("Creating {} CephFS PVCs".format(self.cephfs_pvcs_num))
+        # Generate a given number of CephFS PVCs, randomly assigned RWO or RWX
+        rwo_pvcs = random.randint(0, self.cephfs_pvcs_num)
```
ebenahar: please avoid randomization as this test will be executed as part of a regression suite

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,174 @@
...
+        log.info("Creating CephFS Storage class")
+        self.cephfs_sc = storageclass_factory(interface=constants.CEPHFILESYSTEM)
+        log.info("Creating RBD Storage class")
+        self.rbd_sc = storageclass_factory(interface=constants.CEPHBLOCKPOOL)
+
+    def test_delete_project_with_pvc_and_pods(self, project_factory,
+                                              pvc_factory, multi_pvc_factory,
+                                              pod_factory):
+        # Start with Project 1
+        log.info("Creating Project 1")
```
ebenahar: 'Project 1' isn't the actual project name right? If so, please don't log its var name

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,174 @@
...
+                                     project=project_1,
+                                     storageclass=self.rbd_sc,
+                                     access_mode=constants.ACCESS_MODE_RWO,
+                                     num_of_pvc=self.rbd_pvcs_num)
+
+        # Delete the entire Project 1 (along with all of its PVCs)
+        pvs = helpers.get_all_pvs()
+        space_used_before_deletion = check_ceph_used_space()
+        project_deletion_start_time = time.time()
+        ocp.switch_to_default_rook_cluster_project()
```
ebenahar: @RazTamir @fbalak don't we have something similar for time measurement?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: OK, I will just alternate between the two then.

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: I can move it to the setup fixture, however after deleting this project a new project has to be created. Would it still be better to move this one to the setup fixture?

ebenahar: I see, good point. Let's keep it as is

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,163 @@
...
+
+log = logging.getLogger(__name__)
+
+
+@tier1
+@pytest.mark.polarion_id("OCS-278")
+@pytest.mark.parametrize(
+    argnames=["pvcs_num"],
+    argvalues=[
+        pytest.param(
```
ebenahar: Looks good. 
We'll need each params permutation to be reported with a different polarion_id.
@keesturam @nehaberry wdyt, should we have a polarion case for the scale permutation too?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @pytest.fixture()
+    def setup(self, storageclass_factory):
+        """
+        Create an OCP+OCS Setup, as well as two Storage classes:
+        one for CephFS and one for RBD.
+        Scale the setup to at-least 800 pvcs/pods.
+        """
+        self.ocp_setup = ocp.OCP()
+        self.ocs_setup = OCS()  # TODO: Scale the setup to at-least 800 pvcs/pods
```
ebenahar: is the todo comment still relevant?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,163 @@
...
+        pvc_factory, pod_factory, pvcs_num
+    ):
+        # Start with Project 1
+        log.info("Creating Project")
+        project_1 = project_factory()
+
+        log.info("Creating {} CephFS PVCs".format(pvcs_num))
+        # Generate a given number of CephFS PVCs, some RWO and some RWX
+        rwo_rwx = cycle([constants.ACCESS_MODE_RWO, constants.ACCESS_MODE_RWX])
+        cephfs_pvcs = [pvc_factory(interface=constants.CEPHFILESYSTEM,
```
ebenahar: Seems like 1600 pvcs will be created (800*2).
I think

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,163 @@
...
+        assert space_used_after_deletion < space_used_before_deletion
+        log.info("Space reclaimed successfully.")
+
+        # Project 1 done, moving on to Project 2
+        log.info("Creating Project 2")
+        project_2 = project_factory()
+
+        log.info("Creating a mix of {} CephFS & RBD PVCs "
+                 "(each bound to an app pod)".format(2 * pvcs_num))
+        cephfs_pvcs_pods = [
```
ebenahar: same, 1600

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,163 @@
...
+            helpers.validate_pv_delete(pv_name)
+            pv_delete_time = time.time() - pv_start_time
+            individual_pv_delete_times.append({pv_name: pv_delete_time})
+            time_between_pvcs_deletion_success += pv_delete_time
+        log.info("{} seconds between first and last PVC deletion".format(
+            time_between_pvcs_deletion_success))
+        # Verify space has been reclaimed
+        space_used_after_deletion = self.ceph_cluster.check_ceph_used_space()
+        log.info("Verifying space has been reclaimed...")
+        assert space_used_after_deletion < space_used_before_deletion
```
ebenahar: please add an assertion message:

```python
assert space_used_after_deletion < space_used_before_deletion, (
    f"Space hasn't been reclaimed after PVs deletion./nUsed space before deletion: "
    f"{space_used_before_deletion}/nUsed space after deletion: {space_used_after_deletion}"
)
```

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: Isn't pvcs_num either 100 or 1000 (Depending on the scale)? In that case I fail to see how that'd happen?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: See above discussion

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: Yes - I will discuss it with you and apply accordingly

ocs_ci/ocs/cluster.py
```diff
@@ -383,3 +384,25 @@ def remove_mon_from_cluster(self):
         )
         logging.info(f"Removed the mon {random_mon} from the cluster")
         return remove_mon
+
+    @retry(exceptions.UnexpectedBehaviour, tries=10, delay=3, backoff=1)
```
ebenahar: you'll need to rebase as this function was already moved to cluster.py with a different PR

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
ebenahar: yes, my mistake

ocs_ci/ocs/cluster.py
hadarsharon: Done

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+    ]
+)
+class TestDeleteProjectWithPVCAndPods(ManageTest):
+    """
+    Create at least 2 projects, one with just pvcs and the other with pvcs
+    attached to pod, and delete them.
+    """
+
+    @pytest.fixture()
```
ebenahar: please add autouse=True [1] or inject the fixture name to the test function in line #54 [2] for the fixture to take place


[1]
```python
@pytest.fixture(autouse=True)
```

[2]
```python
def test_delete_project_with_pvc_and_pods(self, setup, ....)
```

hadarsharon: Cheers

Issue #740, opened 2019-09-12
Storage classes are now created by default via operator.


petr-balogh:
Storage classes for cephfs and rbd are now created by default via operator.
So this code should be moved to condition and process only when deploying by old way https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/deployment.py#L303-L324 .

PR #743, opened 2019-09-12
Not create cephfs when deploying via operator

petr-balogh:
The cpehfs is now created by operator, so finishing this TODO.

Fixes: #740


petr-balogh:
Merging this one, if some other stuff needs to be done we can do in other PR.

PR #750, opened 2019-09-13
Respin Ceph pods while running pgsql workload

tiffanyn108:
Test cases: OCS-802, OCS-803, OCS-804
Respin Ceph pods (mon, osd, mgr) while running pgsql workload.

Signed-off-by: Tiffany Nguyen <tunguyen@redhat.com>

### DIFF
--- a/tests/e2e/test_pgsql_pod_respin.py
+++ b/tests/e2e/test_pgsql_pod_respin.py
@@ -1,0 +1,146 @@
+"""
+Module to perform PGSQL workload
+"""
+import logging
+import pytest
+from ocs_ci.ocs.resources.ocs import OCS
+from ocs_ci.ocs.ocp import OCP
+from ocs_ci.utility import templating, utils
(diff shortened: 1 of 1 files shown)


tests/e2e/test_pgsql_pod_respin.py
```diff
@@ -0,0 +1,146 @@
...
+
+    # Create storage class
+    log.info("Creating a Storage Class")
+    storageclass_factory(sc_name='pgsql-workload')
+
+    # Create RipSaw Operator
+    ripsaw = RipSaw()
+
+    def teardown():
+        ripsaw.cleanup()
```
RazTamir: Are we reverting back to the default namespace in the cleanup?

vasukulkarni: The other PR will fix this in the class method, so cleanup() will revert the namespace.

tiffanyn108: Reverting to default namespace was fixed in PR  746.

Issue #737, opened 2019-09-12
Resources leftovers is shown when the pod is respinned "Resources added and resources removed"


Shrivaibavi:
Observed that when pods are respinned the environmental check is failing with error "resources are added and resources are removed". 

Few of the scenarios where we hit the issue. 
1. When setting RBAC permission the pod gets respinned, that is shown in environmental check(when installing openshift-logging)
2. When rebooting the master
3. When draining the worker nodes the pods are respinned.

For now we can use @ignore_leftovers markers, But It is good to have an enhancement for this.

Please share your inputs.

vasukulkarni:
It makes sense to disable this by default and only enable for tests that need the verification.

petr-balogh:
> It makes sense to disable this by default and only enable for tests that need the verification.

What about opposite way? Have it enabled by default but for problematic tests like mentioned in the issue have the way to disable this check?

clacroix12:
> What about opposite way? Have it enabled by default but for problematic tests like mentioned in the issue have the way to disable this check?

Isn't that the current functionality with `@ignore_leftovers`?

petr-balogh:
Yep, it is. So do not think that we should by default skip this leftover check.

PR #753, opened 2019-09-16
Fix for testcase-657

Shrivaibavi:


ebenahar:
@Shrivaibavi Why do we want to ignore the leftovers here? Is it a workaround for some bug? If so, let's consider adding a BZ decorator instead.

Shrivaibavi:
@ebenahar This is not a bug, When i set the RBAC permissions the Prometheus pod is getting respinned. That being showed in the resource leftovers saying "Resources removed resources added".
I opened an issue #737

This issue can be hit very often in case of any node reboot or pod respin. Can these cases be handled, please suggest.

Shrivaibavi:
Closing this PR as it cab be tracked by PR #837

PR #731, opened 2019-09-11
Adding test cases OCS-1009 and OCS-1018

jilju:
OCS-1009 - RBD: Delete csi-rbdplugin while deleting app pods along with their mount points
OCS-1018 - CEPHFS: Delete csi-cephfsplugin while deleting app pods along with their mount points

Signed-off-by: Jilju Joy <jijoy@redhat.com>

PR #766, opened 2019-09-18
test_ocs_1012_1015

jilju:
OCS-1012
CEPHFS: Delete csi-cephfsplugin while PVC deletion, Pod deletion and IO are progressing

OCS-1015
RBD: Delete csi-rbdplugin while PVC deletion, Pod deletion and IO are progressing


jilju:
Dependent on #731

PR #774, opened 2019-09-19
Add deploymentconfig to sanity_helpers

prsurve:

### DIFF
--- a/tests/sanity_helpers.py
+++ b/tests/sanity_helpers.py
@@ -1,5 +1,7 @@
 import logging
 
+from tests import helpers
+
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants, node
 from ocs_ci.ocs.resources.pod import get_fio_rw_iops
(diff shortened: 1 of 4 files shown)


tests/sanity_helpers.py
```diff
@@ -36,13 +39,14 @@ def health_check(self):
         )
         self.ceph_cluster.cluster_health_check(timeout=60)
 
-    def create_resources(self, pvc_factory, pod_factory, run_io=True):
+    def create_resources(self, pvc_factory, pod_factory, dc_pod_factory, run_io=True):
```
ebenahar: I don't think we need to use pod (not DC) anymore

prsurve: will replace pod_factory with dc_pod_factory in next commit

RazTamir:
I'm missing some background about this change.
Why we abandoned the pod_factory in favor of dc?

prsurve:
@RazTamir I have just replaced pod_factoy with dc_pod_factory, if needed i can revert the last commit and keep both factory_fixtures

vasukulkarni:
@prsurve can you add the verified tag?

PR #781, opened 2019-09-20
tests/e2e/test_pv_scale_ocs_create_delete_pvcs.py -- Enhanced

ramkiperiy:
Updated TC to skip memory leak function due to bz 1750328
Removed TODO for creating RAW block pod creation
Removed executor and implemented threads for parallel execution of function
Removed self variables which doesn't require across Class

Signed-off-by: ramkiperiy <rperiyas@redhat.com>

### DIFF
--- a/tests/e2e/test_pv_scale_ocs_create_delete_pvcs.py
+++ b/tests/e2e/test_pv_scale_ocs_create_delete_pvcs.py
@@ -8,8 +8,8 @@ import logging
 import pytest
 import random
 import time
+import threading
 
-from concurrent.futures import ThreadPoolExecutor
 from tests import helpers
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.resources import pod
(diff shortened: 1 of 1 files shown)


ramkiperiy:
Test local run result pastebin link: https://paste.fedoraproject.org/paste/yPxNrDr~SXEIuyum9b7sYQ

tests/e2e/test_pv_scale_ocs_create_delete_pvcs.py
```diff
@@ -138,41 +152,44 @@ def storageclass(self, storageclass_factory):
...
+        scale_pod_count = 120
+        size = '10Gi'
         test_run_time = 180
         self.all_pvc_obj, self.all_pod_obj = ([] for i in range(2))
         self.delete_pod_count = 0
 
         # Identify median memory value for each worker node
-        median_dict = helpers.get_memory_leak_median_value()
-        log.info(f"Median dict values for memory leak {median_dict}")
+        # TODO: Skipping memory leak median calculate because of bz 1750328
```
RazTamir: do we have an issue to track this TODO?

tests/e2e/test_pv_scale_ocs_create_delete_pvcs.py
```diff
@@ -138,41 +152,44 @@ def storageclass(self, storageclass_factory):
...
-                with ThreadPoolExecutor() as executor:
-                    log.info(f"Create {self.pvc_count_each_itr} and "
-                             f"in parallel delete {self.delete_pod_count} "
-                             f"pods & pvc")
-                    thread_list = [self.delete_pvc_pod(), self.create_pvc_pod(
-                        self.rbd_sc_obj, self.cephfs_sc_obj, self.pvc_count_each_itr,
-                        self.size, start_io)]
-                    for thread in thread_list:
-                        executor.submit(thread)
+                log.info(f"Create {pvc_count_each_itr} and "
```
RazTamir:
```
log.info(
    f"Create {pvc_count_each_itr} and in parallel delete {self.delete_pod_count}"
    "pods & pvc"
)
```

tests/e2e/test_pv_scale_ocs_create_delete_pvcs.py
```diff
@@ -138,41 +152,44 @@ def storageclass(self, storageclass_factory):
...
+                         f"in parallel delete {self.delete_pod_count} "
+                         f"pods & pvc")
+                thread1 = threading.Thread(target=self.delete_pvc_pod, args=())
+                thread2 = threading.Thread(target=self.create_pvc_pod, args=(
+                    self.rbd_sc_obj, self.cephfs_sc_obj, pvc_count_each_itr, size, start_io))
```
RazTamir: ) in new line

tests/e2e/test_pv_scale_ocs_create_delete_pvcs.py
ramkiperiy: If bz 1750328 is resolved then this function works as expected, so there is no separate bug. But requires following enhancements.
1. For now memory leak function runs on all workers but in future we will have worker nodes for OCS and app pods, so function needs enhancement to collect data only on OCS nodes.
2. If a worker nodes has 3 OSD's for now function collects all the ceph-osd daemon stats but needs enhancement in analysis, may be based on process id, but should be intelligent to identify restart of process id(which will change).

tests/e2e/test_pv_scale_ocs_create_delete_pvcs.py
ramkiperiy: ack

tests/e2e/test_pv_scale_ocs_create_delete_pvcs.py
ramkiperiy: ack

tests/e2e/test_pv_scale_ocs_create_delete_pvcs.py
ramkiperiy: github issue: https://github.com/red-hat-storage/ocs-ci/issues/811

ramkiperiy:
@vasukulkarni @RazTamir @ebenahar Comments are addressed, if the PR looks in good shape please approve

vasukulkarni:
LGTM

Issue #700, opened 2019-09-05
Consolidate custom configuration CLI arguments


clacroix12:
We currently have two CLI arguments for updating the configuration with non-default values. Both of these options effectively do the same thing. One of them (--ocsci-conf) is specifically used by jenkins which makes it unusable by users for jenkins executions. The other (--cluster-conf) is available for users. The name `--cluster-conf` is a bit disingenuous as well since values can be added to the file you pass here to effect all areas of the configuration, not just values specific to the cluster.

After some discussion with a few team members we thought it would be better to consolidate to having only one option, but make it able to be repeated so multiple configuration files may be passed that will override the default config values. We think sticking with `--ocsci-conf`, making it repeatable, and removing `--cluster-conf` is the right way to do this.

This approach will have to take order into consideration:
```
run-ci --ocsci-conf file1.yml --ocsci-conf file2.yml --ocsci-conf file3.yml
```
In this scenario the default config will be updated with `file1.yml` first, followed by `file2.yml`, and then finally `file3.yml`. If any of the custom config files share keys, the latter file's value will be used.

Objectives:
- [x] Remove `--cluster-conf`
- [x] Make `--ocsci-conf` a repeatable argument
- [x] Remove `--cluster-conf` field from jenkins job definitions

vasukulkarni:
👍

petr-balogh:
This needs to be well organized and after the PR is prepared we need to align also our jenkins job definition for it, test together and once both changes are verified we need to merge together.  Otherwise the definition above looks fine to me. Thanks

petr-balogh:
@clacroix12 I didn't make --ocsci-conf repeatable, instead I am allowing to take list of config files separated by comma which was much easier for implementation in current state as we don't use argparse and parsing those parameters ourself. 

Starting with this parameter ocsci-conf so once we will merge we can start changing our jobs to use this one parameter only and once we and engineering is done with this change in the jenkins jobs we can remove cluster-conf parameter and use just this one.

clacroix12:
@petr-balogh it's actuall easy to change `--ocsci-conf` from storing one value to being repeatable. We can do this by changing the `action` to `append`. I believe the default action is `store`.

```python
    parser.addoption(
        '--ocsci-conf',
        action="append"
        dest='ocsci_conf',
        help="Path to config file of OCS CI",
    )
```

This will change `ocsci_conf` from a single value to a list of values.

I tested this locally with some dummy code:
```python
[...]

PR #786, opened 2019-09-20
Allow ocsci-conf read multiple conf files

petr-balogh:
Initial step for this issue #700.

### DIFF
--- a/ocs_ci/framework/main.py
+++ b/ocs_ci/framework/main.py
@@ -39,14 +39,23 @@ def init_ocsci_conf(arguments=None):
     custom_config = get_param('--ocsci-conf', arguments)
     cluster_config = get_param('--cluster-conf', arguments)
     if custom_config:
-        with open(os.path.expanduser(custom_config)) as file_stream:
-            custom_config_data = yaml.safe_load(file_stream)
-            framework.config.update(custom_config_data)
+        config_files = [cc.strip() for cc in custom_config.split(",")]
+        for config_file in config_files:
+            with open(
+                os.path.abspath(os.path.expanduser(config_file))
+            ) as file_stream:
+                custom_config_data = yaml.safe_load(file_stream)
+                framework.config.update(custom_config_data)
     if cluster_config:
         with open(os.path.expanduser(cluster_config)) as file_stream:
             cluster_config_data = yaml.safe_load(file_stream)
             framework.config.update(cluster_config_data)
     framework.config.RUN['run_id'] = int(time.time())
+    bin_dir = framework.config.RUN.get('bin_dir')
+    if bin_dir:
+        framework.config.RUN['bin_dir'] = os.path.abspath(
+            os.path.expanduser(framework.config.RUN['bin_dir'])
+        )
 
 
 def main(arguments):
@@ -59,6 +68,7 @@ def main(arguments):
         '-p', 'ocs_ci.framework.pytest_customization.reports',
         '--logger-logsdir', pytest_logs_dir,
     ])
-    utils.add_path_to_env_path(os.path.expanduser(
-        framework.config.RUN['bin_dir']))
+    bin_dir = framework.config.RUN.get('bin_dir')
+    if bin_dir:
+        utils.add_path_to_env_path(bin_dir)
     return pytest.main(arguments)
--- a/ocs_ci/framework/main.py
+++ b/ocs_ci/framework/main.py
@@ -1,3 +1,4 @@
+import argparse
 import os
 import time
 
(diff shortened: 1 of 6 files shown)


ocs_ci/framework/main.py
```diff
@@ -39,14 +39,23 @@ def init_ocsci_conf(arguments=None):
     custom_config = get_param('--ocsci-conf', arguments)
     cluster_config = get_param('--cluster-conf', arguments)
     if custom_config:
-        with open(os.path.expanduser(custom_config)) as file_stream:
-            custom_config_data = yaml.safe_load(file_stream)
-            framework.config.update(custom_config_data)
+        config_files = [cc.strip() for cc in custom_config.split(",")]
```
vasukulkarni: I think we meant repeating --ocsci-conf , the docopt equivalent would be [ocsci-conf=conf]...

petr-balogh: Please see my comments about it here: https://github.com/red-hat-storage/ocs-ci/issues/700 . I don't see much additional value to have the parameter repeatable like `--ocscu-conf conf/1.yaml --ocsci-con conf/2.yaml`   . The one parameter `--ocscu-conf conf/1.yaml,conf/2.yaml` is also easy to use and does not require write `--ocscu-conf` twice.

vasukulkarni: I can understand if its hard to fix due to our own param handling, but repeatable args are standard and better than comma separated for file names.  It also helps with auto completion on bash shell.

petr-balogh: yep for auto completion of files is cool thing. I changed the behavior of function to do it like that. The rest is explained in the issue.

vasukulkarni:
LGTM (would have liked if argparse had worked but I think its kind of chicken/egg who takes over)

ocs_ci/framework/main.py
```diff
@@ -36,17 +48,24 @@ def init_ocsci_conf(arguments=None):
...
+            os.path.abspath(os.path.expanduser(config_file))
+        ) as file_stream:
             custom_config_data = yaml.safe_load(file_stream)
             framework.config.update(custom_config_data)
     if cluster_config:
         with open(os.path.expanduser(cluster_config)) as file_stream:
             cluster_config_data = yaml.safe_load(file_stream)
             framework.config.update(cluster_config_data)
     framework.config.RUN['run_id'] = int(time.time())
+    bin_dir = framework.config.RUN.get('bin_dir')
```
vavuthu: since we are already loading bin-dir in line 81, can we append bin_dir  in the same place ( after line 82 ).

dahorak:
> LGTM (would have liked if argparse had worked but I think its kind of chicken/egg who takes over)

Wouldn't be possible to use argparse and [disable the help output](https://docs.python.org/3/library/argparse.html#add-help)?
```
parser = argparse.ArgumentParser(add_help=False)
```
(I didn't try it together with the pytest code, so not sure if it will work or not.)

ocs_ci/framework/main.py
```diff
@@ -59,6 +78,7 @@ def main(arguments):
         '-p', 'ocs_ci.framework.pytest_customization.reports',
         '--logger-logsdir', pytest_logs_dir,
     ])
-    utils.add_path_to_env_path(os.path.expanduser(
-        framework.config.RUN['bin_dir']))
+    bin_dir = framework.config.RUN.get('bin_dir')
+    if bin_dir:
+        utils.add_path_to_env_path(bin_dir)
```
clacroix12: Why do we have to add `bin_dir` to the path here when we are already adding the absolute path in `init_ocsci_conf`?

petr-balogh: We always added bin dir in the PATH. So didn't changed the logic here. Just moved it all to the function init config for now to have it in one place.

petr-balogh:
@dahorak thanks for sharing this functionality, I didn't know about this prase_known_args and option to not process help, aligned the code

ocs_ci/framework/main.py
```diff
@@ -8,45 +9,36 @@
...
-        return
-    custom_config = get_param('--ocsci-conf', arguments)
-    cluster_config = get_param('--cluster-conf', arguments)
-    if custom_config:
-        with open(os.path.expanduser(custom_config)) as file_stream:
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument('--ocsci-conf', action='append', default=[])
+    # cluster-conf parameter will be deleted once we will update all the jobs
+    parser.add_argument('--cluster-conf')
+    args, unknown = parser.parse_known_args()
```
dahorak: This should fix the failing unit tests.
```suggestion
    args, unknown = parser.parse_known_args(args=arguments)
```

ocs_ci/framework/main.py
petr-balogh: Moved everything here to be on one place

petr-balogh:
Verifying in my temp job with modification to provide config with --ocsci-conf: 
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster-pbalogh-test/8/console

looks like it's running deployment OK with 3 provided config files as you can see from console output.

ocs_ci/framework/main.py
clacroix12: @petr-balogh right, we were doing it in two places when I had commented this.

PR #787, opened 2019-09-20
Stop osd test case

fbalak:
Add test_ceph_osd_stopped test case.

### DIFF
--- a/tests/manage/monitoring/conftest.py
+++ b/tests/manage/monitoring/conftest.py
@@ -118,7 +118,6 @@ def measure_operation(
         stop_time = time.time()
         info['run'] = False
         logging_thread.join()
-        logger.info(f"Alerts found during measurement: {alert_list}")
         results = {
             'start': start_time,
             'stop': stop_time,
@@ -126,6 +125,7 @@ def measure_operation(
             'metadata': metadata,
             'prometheus_alerts': alert_list
         }
+        logger.info(f"Results of measurement: {results}")
         with open(result_file, 'w') as outfile:
             logger.info(f"Dumping results of measurement into {result_file}")
             json.dump(results, outfile)
--- a/tests/manage/monitoring/conftest.py
+++ b/tests/manage/monitoring/conftest.py
@@ -291,11 +291,11 @@ def workload_stop_ceph_osd(measurement_dir):
     """
     oc = ocp.OCP(
         kind=constants.DEPLOYMENT,
-        namespace=config.ENV_DATA['cluster_namespace']
+        namespace=config.ENV_DATA.get('cluster_namespace')
     )
-    osd_deployments = oc.get(selector=constants.OSD_APP_LABEL)['items']
+    osd_deployments = oc.get(selector=constants.OSD_APP_LABEL).get('items')
     osds = [
-        deployment['metadata']['name']
+        deployment.get('metadata').get('name')
         for deployment in osd_deployments
     ]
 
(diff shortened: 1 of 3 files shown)


tests/manage/monitoring/conftest.py
```diff
@@ -277,3 +277,59 @@ def stop_mon():
...
+
+
+@pytest.fixture
+def workload_stop_ceph_osd(measurement_dir):
+    """
+    Downscales Ceph osd deployment, measures the time when it was
+    downscaled and alerts that were triggered during this event.
+
+    Returns:
+        dict: Contains information about `start` and `stop` time for stopping
```
RazTamir: Will be good to add example of that dict

tests/manage/monitoring/conftest.py
```diff
@@ -277,3 +277,59 @@ def stop_mon():
         assert check_old_mons_deleted, msg
 
     return measured_op
+
+
+@pytest.fixture
+def workload_stop_ceph_osd(measurement_dir):
```
vasukulkarni: name should be measure_ceph_stop_osd for better representing what it does.

tests/manage/monitoring/conftest.py
```diff
@@ -277,3 +277,59 @@ def stop_mon():
...
+    osd_to_stop = osds[-1]
+    logger.info(f"osd disks to stop: {osd_to_stop}")
+    logger.info(f"osd disks left to run: {osds[:-1]}")
+
+    def stop_osd():
+        """
+        Downscale Ceph osd deployments for 11 minutes. First 1 minutes
+        the alert CephOSDDiskNotResponding should be in 'Pending'. After 1 minute
+        the alert turns into 'Firing' state.        This configuration of osd can be observed in ceph-mixins which
+        are used in the project:
```
vasukulkarni: docstring needs formatting.

tests/manage/monitoring/conftest.py
```diff
@@ -277,3 +277,59 @@ def stop_mon():
...
+            https://github.com/ceph/ceph-mixins/blob/d22afe8c0da34490cb77e52a202eefcf4f62a869/config.libsonnet#L21
+        There should be also CephClusterWarningState alert that takes 10
+        minutest to be firing.
+
+        Returns:
+            str: Names of downscaled deployments
+        """
+        # run_time of operation
+        run_time = 60 * 11
```
vasukulkarni: could be parameterized?

tests/manage/monitoring/conftest.py
```diff
@@ -277,3 +277,59 @@ def stop_mon():
...
+        dict: Contains information about `start` and `stop` time for stopping
+            Ceph osd pod
+    """
+    oc = ocp.OCP(
+        kind=constants.DEPLOYMENT,
+        namespace=config.ENV_DATA['cluster_namespace']
+    )
+    osd_deployments = oc.get(selector=constants.OSD_APP_LABEL)['items']
+    osds = [
+        deployment['metadata']['name']
```
ebenahar: use .get()

fbalak: done

tests/manage/monitoring/conftest.py
fbalak: I don't think that it should be parametrized as it is specific for the alert to be triggered. What json did you mean? If you mean ceph-mixins configuration then I wouldn't do that as it is external upstream project. Also if we don't load the value dynamically and if the test starts to fail in CI due to change in times for status change then we can easy notice that change and resolve it.

tests/manage/monitoring/conftest.py
fbalak: Good idea! I will change it next PR so it is changed for all alerting tests at once. Fixtures for all alerting tests are currently named this way but this PR is only about one test case.

PR #789, opened 2019-09-20
Allow running node related tests on any platform type

ebenahar:
This PR prepares the classes needed for the abstraction layer of nodes operations used in node related tests.
Adjusted disk failures related tests to this change. Will adjust other tests in following PRs

### DIFF
--- a/ocs_ci/ocs/platform_nodes.py
+++ b/ocs_ci/ocs/platform_nodes.py
@@ -1,6 +1,11 @@
 import logging
+
+from ocs_ci.ocs.exceptions import TimeoutExpiredError
+from ocs_ci.utility.utils import TimeoutSampler
 from ocs_ci.framework import config
 from ocs_ci.utility import aws
+from ocs_ci.ocs.node import get_node_objs
+from ocs_ci.ocs.resources.pvc import get_deviceset_pvs
 
 
 logger = logging.getLogger(__name__)
@@ -25,7 +30,10 @@ class NodesBase(object):
     Should be inherited by specific platform classes
 
     """
-    def get_data_volume(self, node):
+    def get_data_volumes(self):
+        pass
+
+    def get_node_by_attached_volume(self, volume):
         pass
 
     def stop_nodes(self, nodes):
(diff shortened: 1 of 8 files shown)


ocs_ci/ocs/platform_nodes.py
```diff
@@ -0,0 +1,106 @@
...
+    def get_data_volume(self, node):
+        pass
+
+    def stop_nodes(self, nodes):
+        pass
+
+    def start_nodes(self, nodes):
+        pass
+
+    def restart_nodes(self, nodes):
```
hnallurv: Is this 'reboot' of nodes?

vasukulkarni:
👍  like the flow of class/methods.

ocs_ci/ocs/platform_nodes.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    def __init__(self):
+        self.cls_map = {'AWS': AWSNodes, 'VMWare': VMWareNodes}
+
+    def get_nodes_platform(self):
+        platform = config.ENV_DATA['platform']
+        return self.cls_map[platform]()
+
+
+class PlatformNodes(object):
```
vasukulkarni: we could simply call this Nodes too?

ebenahar: yes

ocs_ci/ocs/platform_nodes.py
ebenahar: yes. I kept consistency with aws.restart_instances

ocs_ci/ocs/platform_nodes.py
```diff
@@ -0,0 +1,276 @@
...
+    """
+    def __init__(self):
+        self.cls_map = {'AWS': AWSNodes, 'VMWare': VMWareNodes}
+
+    def get_nodes_platform(self):
+        platform = config.ENV_DATA['platform']
+        return self.cls_map[platform]()
+
+
+class NodesBase(object):
```
dahorak: I'm just curious, wouldn't be better to rise the `NotImplementedError` for the methods in this `NodeBase` class instead of in the child classes?

ebenahar:
@dahorak Added NotImplementedError to both base and VMWare classes. Thanks for noticing

PR #790, opened 2019-09-20
Fixes the ocs-ci tier2 failure due to leftover PVs in Test case test_pvc_concurrent_deletion_creation.py

nehaberry:
The Test case  test_pvc_concurrent_deletion_creation.py was intermittently failing in teardown due to PV leftovers. 
E.g. https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/865/

The failure is due to leftover PVs in Released state. There are some knows bugs around it, e.g. BZ#1729853 and it is an intermittent issue.

Plan - test will be updated to use pvc_factory to get rid of this issue - Won't call it as a automation bug fix because it is not automation bug. 

Why we update the test instead of marking some related bug to skip the test ?
 - It is because the bug(PV in released state for long time) is not affecting the test case. The test case is passing. Only teardown is failing.

Also, added check for underlying PV delete once all PVCs are deleted.
@jilju thanks for the help :) 

Signed-off-by: Neha Berry <nberry@redhat.com>

### DIFF
--- a/tests/manage/pv_services/test_pvc_concurrent_deletion_creation.py
+++ b/tests/manage/pv_services/test_pvc_concurrent_deletion_creation.py
@@ -5,56 +5,52 @@ import logging
 from concurrent.futures import ThreadPoolExecutor
 import pytest
 
-from ocs_ci.ocs import exceptions, constants
+from ocs_ci.ocs import constants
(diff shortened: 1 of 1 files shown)


nehaberry:
The Test case passed in local setup.

Trying to run in jenkins.

PR #796, opened 2019-09-23
Remove fixed version for DS deployment

petr-balogh:


petr-balogh:
Verified here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/1056/console

nehaberry:
The install worked as seen in [[1]](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/1056/console) , hence approving.

zmc:
Is there some background on why this was needed?

petr-balogh:
> Is there some background on why this was needed?
@zmc sorry that I wasn't much talkative here this was discussed on Monday on our ocs-qe google hangout channel that even if the downstream operator build supposed to be fixed we were failing cause I forgot remove this from config when I tested the PR https://github.com/red-hat-storage/ocs-ci/pull/713 if it also work with specified image.

So was just quick fix on Monday morning to enable our jobs to run with really latest downstream build to verify the BZ with LVM issue was really fixed.

We don't want to have specific tag in our default config as if it's not specified it will take automatically the specific tag from latest tag.

We will specify it in our configuration only when triggering with UMB to make sure that we are running against built which sent UMB. But if someone would like to just run the latest DS build no need to specify any image.

Issue #798, opened 2019-09-23
Check searching for already existing clusters with the same prefix fails when old cluster has been just  destroyed


mbukatov:
When I execute a teardown a cluster and then immediately deploy a new one, new check searching for already existing clusters with the same prefix fails:

```
E               ocs_ci.ocs.exceptions.SameNamePrefixClusterAlreadyExistsException: Cluster with name prefix mbukatov* already exists. Please destroy the existing cluster for a new cluster deployment
```

ebenahar:
@mbukatov we can check for the status of the instances and in case they are 'Terminating', allow deployment. WDYT?

mbukatov:
@ebenahar That could help.

PR #801, opened 2019-09-23
Allow deployment for existing name prefix cluster if the instances are Terminated

ebenahar:
Addresses issue https://github.com/red-hat-storage/ocs-ci/issues/798


ocs_ci/deployment/aws.py
```diff
@@ -89,7 +89,13 @@ def check_cluster_existence(self, cluster_name_prefix):
...
 
         """
-        if self.aws.get_instances_by_name_pattern(cluster_name_prefix):
+        instances = self.aws.get_instances_by_name_pattern(cluster_name_prefix)
+        instance_objs = [self.aws.get_ec2_instance(ins.get('id')) for ins in instances]
+        non_terminated_instances = [
+            ins for ins in instance_objs if ins.state
+            .get('Code') != constants.INSTANCE_TERMINATED
+        ]
+        if non_terminated_instances:
```
mbukatov: I would rather used an explicit check like `len(non_terminated_instances) > 0` ...

mbukatov: You could also consider logging the non terminated instances.

ebenahar:
Deployment finished successfully here:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/1065/console

We're using a different config in Jenkins, that allows multiple clusters with the same name prefix.

As for local runs:

[deployment_logs.zip](https://github.com/red-hat-storage/ocs-ci/files/3642926/deployment_logs.zip)

This was executed while having a cluster with running instances, therefore, this PR doesn't break anything.

ocs_ci/deployment/aws.py
ebenahar: Sorry @mbukatov, just now see your comments.
Regarding the first one, how is it different? if non_terminated_instances is basically the same as if the len of it is 0.
Regarding the second one, I'll raise a PR for improving the logging

ebenahar: https://github.com/red-hat-storage/ocs-ci/pull/804

PR #824, opened 2019-09-26
skip leftover checks for tier3 and tier4 by default and add disruptive mark

vasukulkarni:
skip leftover checks for tier3 and tier4 by default and also add disruptive marker that will skip environment checker

Fixes: https://github.com/red-hat-storage/ocs-ci/issues/737

### DIFF
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -43,6 +43,9 @@ bugzilla = pytest.mark.bugzilla
 # mark the test class with marker below to ignore leftover check
 ignore_leftovers = pytest.mark.ignore_leftovers
 
+# mark the test that are disruptive in nature and avoid leftover_checks
+disruptive = pytest.mark.disruptive
+
 # testing marker this is just for testing purpose if you want to run some test
 # under development, you can mark it with @run_this and run pytest -m run_this
 run_this = pytest.mark.run_this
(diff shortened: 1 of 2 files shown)


RazTamir:
Why rhis PR?

vasukulkarni:
I have linked this to an issue above: Fixes: #737

RazTamir:
Ok,

This will disable the environment check completely. 
Please don't merge it cause we do need to ensure that after each test, the environment is "clean"

vasukulkarni:
> Ok,
> 
> This will disable the environment check completely.
> Please don't merge it cause we do need to ensure that after each test, the environment is "clean"

Its not disabling for all tests, only for disruptive tests by default so that we dont have to mark those for all such testcases.

RazTamir:
I see tier2/3/4 are marked for skip (also described in the PR description)

vasukulkarni:
> I see tier2/3/4 are marked for skip (also described in the PR description)

I visiting the mojo for definition again, I see tier2 is still functional, but tier3 and above are disruptive so should be skipped. I will fix that in pr subject and test.

RazTamir:
Tier3 is not disruptive but negative.
I really prefer that every test will add the decorator to skip the check

vasukulkarni:
> Tier3 is not disruptive but negative.
> I really prefer that every test will add the decorator to skip the check

We can discuss this in call, I really want to understand what is the reason to force add when we know the disruptive tests are going to fail the check?

ebenahar:
@vasukulkarni Would like to close old PRs, is this one still relevant?

PR #780, opened 2019-09-19
Prevent deployment according to cluster name prefix

ebenahar:
Addresses issue https://github.com/red-hat-storage/ocs-ci/issues/779 - Prevent cluster deployment in case a cluster with the same kerberos ID already exists

Signed-off-by: Elad Ben Aharon <ebenahar@redhat.com>

PR #828, opened 2019-09-26
downstream-ci: fix OCP deployment

andrewschoen:
This changes our config to match the downstream_config.yaml
file in ocs-ci. We also pass the original downstream_config.yaml
as an extra --ocsci-conf argument.

The only ocs-operator build I was able to get working was:

http://quay.io/rhceph-dev/ocs-registry:4.2-77.bed1df0.master

Signed-off-by: Andrew Schoen <aschoen@redhat.com>

### DIFF
--- a/.functional_ci_setup.py
+++ b/.functional_ci_setup.py
@@ -92,33 +92,36 @@ def get_ocsci_conf():
             region=env['AWS_REGION'],
             base_domain=env['AWS_DOMAIN'],
             worker_instance_type='m5.4xlarge',
+            cluster_namespace="openshift-storage",
+
         ),
     )
     if env.get("DOWNSTREAM") == "true":
         conf_obj['REPORTING'] = dict(us_ds='DS')
     if env.get("OCS_OPERATOR_DEPLOYMENT") == "true":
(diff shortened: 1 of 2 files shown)


.functional_ci_setup.py
```diff
@@ -92,33 +92,36 @@ def get_ocsci_conf():
...
         ),
     )
     if env.get("DOWNSTREAM") == "true":
         conf_obj['REPORTING'] = dict(us_ds='DS')
     if env.get("OCS_OPERATOR_DEPLOYMENT") == "true":
         conf_obj['DEPLOYMENT'] = dict(
-            ocs_operator_deployment=True,
             ocs_operator_image=env['OCS_OPERATOR_IMAGE'],
-            ocs_registry_image=env['OCS_REGISTRY_IMAGE'],
+            ocs_operator_deployment=True,
```
petr-balogh: @zmc not sure if you are naming clusters the same before first dash, but will be worth if you will add also this parameter:
force_deploy_multiple_clusters: True

It's recent changed introduced here:
https://github.com/red-hat-storage/ocs-ci/pull/780/files

For preventing ppl to create new clusters if they already have one.

Issue #830, opened 2019-09-27
Stuck tier1 test in downstream CI


andrewschoen:
We've got a test that ran for ~ 20 hours in our downstream tier1 testing before I cancelled the job. Here is the last bit of relevant log output:

```
tests/manage/cluster/nodes/test_nodes_maintenance.py::TestNodesMaintenance::test_node_maintenance[worker] 
-------------------------------- live log setup --------------------------------
12:54:32 - Dummy-11 - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig cluster/auth/kubeconfig get Pod  -A -o yaml
12:54:32 - Dummy-9 - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig cluster/auth/kubeconfig get StorageClass  -A -o yaml
12:54:32 - Dummy-13 - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig cluster/auth/kubeconfig get CephFileSystem  -A -o yaml
12:54:32 - Dummy-8 - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig cluster/auth/kubeconfig get CephBlockPool  -A -o yaml
12:54:32 - Dummy-10 - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig cluster/auth/kubeconfig get PersistentVolume  -A -o yaml
12:54:32 - Dummy-1 - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig cluster/auth/kubeconfig get PersistentVolumeClaim  -A -o yaml
12:54:32 - Dummy-12 - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig cluster/auth/kubeconfig get Namespace  -A -o yaml
12:54:41 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig cluster/auth/kubeconfig get CephCluster  -n openshift-storage -o yaml
12:54:41 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig cluster/auth/kubeconfig get CephFilesystem  -n openshift-storage -o yaml
12:54:41 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig cluster/auth/kubeconfig get Pod  -n openshift-storage -o yaml
[...]

PR #831, opened 2019-09-27
Add default timeout(600 seconds) for run_cmd

vasukulkarni:
Avoids: https://github.com/red-hat-storage/ocs-ci/issues/830

### DIFF
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -84,7 +84,9 @@ class OCP(object):
         """
         self._data = self.get()
 
-    def exec_oc_cmd(self, command, out_yaml_format=True, secrets=None, **kwargs):
+    def exec_oc_cmd(
+        self, command, out_yaml_format=True, secrets=None, timeout=600, **kwargs
+    ):
         """
         Executing 'oc' command
 
@@ -99,6 +101,8 @@ class OCP(object):
                 This kwarg is popped in order to not interfere with
                 subprocess.run(**kwargs)
 
+            timeout (int): timeout for the oc_cmd, defaults to 600 seconds
+
         Returns:
             dict: Dictionary represents a returned yaml file
         """
@@ -111,7 +115,7 @@ class OCP(object):
             oc_cmd += f"--kubeconfig {kubeconfig} "
 
         oc_cmd += command
-        out = run_cmd(cmd=oc_cmd, secrets=secrets, **kwargs)
+        out = run_cmd(cmd=oc_cmd, secrets=secrets, timeout, **kwargs)
 
         try:
             if out.startswith('hints = '):
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -115,7 +115,7 @@ class OCP(object):
             oc_cmd += f"--kubeconfig {kubeconfig} "
 
         oc_cmd += command
-        out = run_cmd(cmd=oc_cmd, secrets=secrets, timeout, **kwargs)
+        out = run_cmd(cmd=oc_cmd, secrets=secrets, timeout=timeout, **kwargs)
 
         try:
             if out.startswith('hints = '):
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -384,6 +384,8 @@ def run_cmd(cmd, secrets=None, **kwargs):
             This kwarg is popped in order to not interfere with
             subprocess.run(**kwargs)
 
+        timeout (int): Timeout for the command, defaults to 600 seconds.
+
     Raises:
         CommandFailed: In case the command execution fails
 
@@ -395,6 +397,9 @@ def run_cmd(cmd, secrets=None, **kwargs):
     log.info(f"Executing command: {masked_cmd}")
     if isinstance(cmd, str):
         cmd = shlex.split(cmd)
+    if not kwargs.get('timeout'):
+        # set a default timeout for any command to timeout within 600 seconds
+        kwargs['timeout'] = 600
     r = subprocess.run(
         cmd,
         stdout=subprocess.PIPE,
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -373,7 +373,7 @@ def mask_secrets(plaintext, secrets):
     return plaintext
 
 
-def run_cmd(cmd, secrets=None, **kwargs):
+def run_cmd(cmd, secrets=None, timeout=600, **kwargs):
     """
     Run an arbitrary command locally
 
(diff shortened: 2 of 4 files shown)


clacroix12:
@vasukulkarni `run_cmd` is used in a ton of places. Some usages will not succeed in under 600 seconds, namely when we execute the `openshift-installer`. These instances will need to be updated to pass a reasonable timeout otherwise they will fail too soon.

vasukulkarni:
> @vasukulkarni `run_cmd` is used in a ton of places. Some usages will not succeed in under 600 seconds, namely when we execute `openshift-install`. These instances will need to be updated to pass a reasonable timeout otherwise they will fail too soon.

Yeah I know that, its base for every oc cmd, but probably very few need to cross 600 seconds mark.  Anyway its not going to get merged without testing :)

ocs_ci/utility/utils.py
```diff
@@ -381,6 +381,8 @@ def run_cmd(cmd, secrets=None, **kwargs):
             This kwarg is popped in order to not interfere with
             subprocess.run(**kwargs)
 
+        timeout (int): Timeout for the command, defaults to 600 seconds.
```
petr-balogh: This parameter is actually not in the arguments in the function so I guess we should document it under the kwargs or Keword arguments? 

Also there is one empty line above which should't be there.

ocs_ci/utility/utils.py
```diff
@@ -392,6 +394,9 @@ def run_cmd(cmd, secrets=None, **kwargs):
     log.info(f"Executing command: {masked_cmd}")
     if isinstance(cmd, str):
         cmd = shlex.split(cmd)
+    if not kwargs.get('timeout'):
+        # set a default timeout for any command to timeout within 600 seconds
+        kwargs['timeout'] = 600
```
petr-balogh: I would email ocs-qe mailing list to ask ppl to identify those commands. Test owners will probably know better in which places the 600 seconds won't be enough. WDYT?

ocs_ci/utility/utils.py
vasukulkarni: I will fix that,  **kwargs is essentially a named param but the syntactic sugar hides the param.

mbukatov:
You are breaking our guidelines by not creating an issue where you explain what is broken with our default timeout configuration.

vasukulkarni:
> You are breaking our guidelines by not creating an issue where you explain what is broken with our default timeout configuration.

@mbukatov its right there in the first message of github? Avoids: #830

mbukatov:
> @mbukatov its right there in the first message of github? Avoids: #830

Ok, so could you use `Fixes: #XYZ` so that your intentions are clear and the issue will be understood this way by github?

ocs_ci/utility/utils.py
mbukatov: I disagree. In some cases the author will likely be aware of that (eg. as Coady noted for the installation), but in most cases I doubt owners will be able to identify most such cases without  rerun and retesting.

mbukatov: I would suggest for Vasu to present xml junit files run of all (not just tier1) cases with and without the pr, analyze the differences and only then go to the people asking for help.

petr-balogh: @mbukatov if Vasu will run all the tiers, compare the results it will take some time and we don't cover all the test cases cause of BZ skips . So I just thought that before even executing and spend this time would be worth to just ask ppl if they remember which commands they are executing in their test cases take more then 600 timeout just in case they remember to give some rough identification. So we can do some alignment int this PR even before the execution all tiers.

mbukatov: > compare the results it will take some time and we don't cover all the test cases cause of BZ skips

I understand that you want to optimize the work you need to do to push parts of it to rest of the team. That said, this is the case for every change in the framework which could affect any test. If someone else did a change like that, you would require that the person does all this work. 

That said, do what you consider the best way forward. I just want to point this difference in approach clearly.

petr-balogh: Anyway, Vasu already sent the email yestarday so he will get maybe some feedback from someone and now it's up to him to do the rest. ;)

ocs_ci/ocs/ocp.py
```diff
@@ -94,7 +98,7 @@ def exec_oc_cmd(self, command, out_yaml_format=True, secrets=None, **kwargs):
             oc_cmd += f"--kubeconfig {kubeconfig} "
 
         oc_cmd += command
-        out = run_cmd(cmd=oc_cmd, secrets=secrets, **kwargs)
+        out = run_cmd(cmd=oc_cmd, secrets=secrets, timeout, **kwargs)
```
dahorak:
```
     ...
  File "/home/jenkins/workspace/qe-deploy-ocs-cluster/ocs-ci/ocs_ci/ocs/ocp.py", line 101
    out = run_cmd(cmd=oc_cmd, secrets=secrets, timeout, **kwargs)
                                              ^
SyntaxError: positional argument follows keyword argument
 ```
From [validation job logs](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/1205/console).

vasukulkarni: thanks :) fixing.

vasukulkarni:
Surprisingly all the tests passed!

petr-balogh:
What about other tears ? Do we need to run also with other tears as this affect not just tier1?

vasukulkarni:
> LGTM but prefer to run more tiers as well to see if we see some timeouts ;) In PR verification job you can chose test suite tier1 or tier2 or tier3 for example to test more tiers ;)

I will run tier2 and tier3 as well, just in case something comes up..

vasukulkarni:
I think its better to merge and we can use this conf file for testing when needed, the default is not changed anyway and the tier1/workload tests are passing. If there are port changes(which I doubt) we can take it as follow on PR.

dahorak:
@vasukulkarni what is the status of this PR? ... I'm for merging it, because we are still hitting issues that some command gets stuck for hours ... and especially in CI it is quite painful.
I can also implement timeout on the job level, but this will be definitely better, because one stuck command will not break whole test suite execution.

vasukulkarni:
Well have problems with installation today :)

vasukulkarni:
I think we can merge this one now, I pushed after rebase so lost the previous approvals,  VMware and UPI , I am not sure how to run, but VIjay/Shylesh can add the timeout for installs in follow on PR.

Test report for tier1 / 2/ 3 - No issues seen.

https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/QE/job/qe-trigger-test-pr/546/testReport/

PR #835, opened 2019-09-29
Implementation of  MCG storage class

parikshithb:


tests/helpers.py
```diff
@@ -374,6 +374,30 @@ def create_storage_class(
     return create_resource(**sc_data)
 
 
+def create_mcg_storageclass(reclaim_policy=constants.RECLAIM_POLICY_DELETE, sc_name=None):
```
ebenahar: I think you can utilize the already existing create_storageclass() function we have in helpers.py and tp pass it the MCG SC yaml

tests/manage/mcg/conftest.py
```diff
@@ -145,3 +150,38 @@ def awscli_pod(mcg_obj, created_pods):
     helpers.wait_for_resource_state(awscli_pod_obj, constants.STATUS_RUNNING)
     created_pods.append(awscli_pod_obj)
     return awscli_pod_obj
+
+
+@pytest.fixture()
+def mcg_storageclass(request):
```
ebenahar: the same, try to utilize the storage class factory fixture we have

tests/helpers.py
parikshithb: create_storageclass() in helpers has certain default parameters like secret_name, interface_name which is not required to create a mcg storage class.

tests/manage/mcg/conftest.py
parikshithb: Similary, storage class factory calls ceph_pool_factory_class and secret_factory_class which is not required for mcg sc unlike rbd and cephfs sc.

parikshithb:
@ebenahar True, will wait on the resolution of this bz, if there is default storage class for OBCs created then there is no need in creating a separate storageclass.

ebenahar:
@parikshithb ACK, thanks!

parikshithb:
There is no need for creating a separate storageclass for obc as it is created by default and only reclaim policy delete is supported. Closing this PR.

Issue #818, opened 2019-09-25
Make downstream deployment a default choice


mbukatov:
So far when one deploys OCS cluster via ocs-ci, the upstream ocs images are used. But since we have operator driven deployment available downstream, we should be using downstream deployment as a default already or very soon.

petr-balogh:
It was also decided in mail thread: `Please use OCS-4.2 downstream builds for testing` in ocs-qe mailing list.

PR #839, opened 2019-09-30
Move to downstream

petr-balogh:
Fixes: #818

### DIFF
--- a/conf/ocsci/upstream_config.yaml
+++ b/conf/ocsci/upstream_config.yaml
@@ -1,0 +1,20 @@
+---
+ENV_DATA:
+  # --------------------------------------------------------------------------
+  # Following parameters are not used when deploy via operator and will removed
+  # soon as part of cleanup: https://projects.engineering.redhat.com/browse/OCSQE-73
+  ceph_image: 'ceph/ceph:v14.2.3-20190904'
+  rook_image: 'rook/ceph:master'
+  # End of the unused parameters for operator deployment
+  # --------------------------------------------------------------------------
+  cluster_namespace: 'openshift-storage'
+REPORTING:
+  # Upstream: 'US' or Downstream: 'DS', used only for reporting (Test Run Name)
+  us_ds: 'US'
+  ocs_must_gather_image: "quay.io/ocs-dev/ocs-must-gather"
+DEPLOYMENT:
+  ocs_operator_deployment: true
+  ocs_operator_storage_cluster_cr: "https://raw.githubusercontent.com/openshift/ocs-operator/master/deploy/crds/ocs_v1_storagecluster_cr.yaml"
+  ocs_operator_olm: "https://raw.githubusercontent.com/openshift/ocs-operator/master/deploy/deploy-with-olm.yaml"
+  # you can overwrite the image for ocs operator catalog souce by following parameter:
+  # ocs_operator_image: "quay.io/ocs-dev/ocs-registry:latest"
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -60,6 +60,7 @@ REPORTING:
     project_id: "OpenShiftContainerStorage"
   # Upstream: 'US' or Downstream: 'DS', used only for reporting (Test Run Name)
   us_ds: 'DS'
+  ocp_must_gather_image: "quay.io/openshift/origin-must-gather"
   ocs_must_gather_image: "quay.io/rhceph-dev/ocs-must-gather"
   gather_on_deploy_failure: true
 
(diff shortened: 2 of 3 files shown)


conf/ocsci/upstream_config.yaml
```diff
@@ -0,0 +1,20 @@
+---
+ENV_DATA:
```
vasukulkarni: why upstream_config.yaml changes?

petr-balogh: As we move to DS by default I would like to provide the config if someone would like to still deploy with US for some bug verification earlier than it will be in DS for example. So ppl can achieve with this config

vasukulkarni:
LGTM

petr-balogh:
Need fix CI, already left office, will do at evening

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -82,12 +82,22 @@ ENV_DATA:
...
-  rook_image: 'rook/ceph:master'
+  ceph_image: 'quay.io/rhceph-dev/rhceph-4.0-rhel-8:latest'
+  rook_image: 'quay.io/rhceph-dev/rook:latest'
+  # DS CSI sidecars images mentioned in https://jira.coreos.com/browse/RHSTOR-150
+  rook_csi_registrar_image: 'quay.io/openshift/origin-csi-node-driver-registrar:4.2'
+  rook_csi_provisioner_image: 'quay.io/openshift/origin-csi-external-provisioner:4.2'
+  rook_csi_snapshotter_image: 'quay.io/openshift/origin-csi-external-snapshotter:4.2'
+  rook_csi_attacher_image: 'quay.io/openshift/origin-csi-external-attacher:4.2'
+  # End of the unused parameters for operator deployment
+  # --------------------------------------------------------------------------
```
RazTamir: One line below.
Should be m5.4xlarge?

petr-balogh: https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/production-aws-ipi.yaml
For production jobs we are using this config. We can have one general production config which will not be bounded to AWS-IPI later on but we don't want to all ppl to have this bigger instance type as default IIRC, just for production jobs as we don't have enough resources to allow everyone to deploy on m5.4xlarge

clacroix12:
@petr-balogh Should we also remove `conf/ocsci/downstream_config.yaml` as a part of this? Seems it's probably not necessary since we are defaulting to downstream now.

petr-balogh:
> @petr-balogh Should we also remove `conf/ocsci/downstream_config.yaml` as a part of this? Seems it's probably not necessary since we are defaulting to downstream now.

I was thinking about this as well but then I thought that it will be better to keep also DS config, so we can easily compare what difference we have between US/DS and also in the jobs would be better if I will pass DS config to be explicit.

clacroix12:
@petr-balogh I see your point, but that also means that if anything changes on the config side for downstream we now need to update two separate places. I think we should try to avoid this scenario as the explicit downstream config and the default config might become disconnected.

dahorak:
Since the `downstream_config.yaml` config file is removed, we will have to align the merge of this PR with update in our CI jobs.
Another option will be preserving the `downstream_config.yaml` config file in this PR and once it will be merged, update our CI job and then remove the `downstream_config.yaml` file in separated PR? (I'm ok with both options here, do not have strong opinion which one is better/easier...)

petr-balogh:
> Since the `downstream_config.yaml` config file is removed, we will have to align the merge of this PR with update in our CI jobs.
> Another option will be preserving the `downstream_config.yaml` config file in this PR and once it will be merged, update our CI job and then remove the `downstream_config.yaml` file in separated PR? (I'm ok with both options here, do not have strong opinion which one is better/easier...)

Decided to keep the downstream_config here till we will be sure we can delete it.
Created issue here:  #855

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -59,9 +59,8 @@ REPORTING:
   polarion:
     project_id: "OpenShiftContainerStorage"
   # Upstream: 'US' or Downstream: 'DS', used only for reporting (Test Run Name)
-  us_ds: 'US'
-  ocp_must_gather_image: "quay.io/openshift/origin-must-gather"
```
vavuthu: why we are removing ocp image from default. we are still using ocp image for collecting logs from failed test cases

https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/utils.py#L735

petr-balogh: Ah, good catch @vavuthu.  Removed by mistake when moving to upstream config.  Fixed in next commit

petr-balogh:
PR verification job passed crucial part, failed with deployment of OSDs cause of resource limit.

Issue #846, opened 2019-10-02
Separate subscription from deploy-with-olm.yaml


petr-balogh:
Description of problem is here:
https://bugzilla.redhat.com/show_bug.cgi?id=1757013

Issue #782, opened 2019-09-20
ocs_ci.ocs.exceptions.ResourceInUnexpectedState: CSV: ocs-operator.v0.0.1 is not in expected phase: Succeeded


shreekarSS:
ocs-operator is taking little longer to get into Succeeded. 15 seconds after ocs_ci.ocs.exceptions.ResourceInUnexpectedState, ocs-operator.v0.0.1 got into Succeded for me. Should increase timeout

petr-balogh:
Reopening since new update in BZ https://bugzilla.redhat.com/show_bug.cgi?id=1757013 . 

Will try to increase the timeout and we will see, also raised the question in BZ about what is the reasonable timeout here.

PR #847, opened 2019-10-02
Wait for catalog source before subscription

petr-balogh:
Fixes: #782 - increased timeout for 100 seconds for CSV
Fixes: #846 - separate subscriptin from deploy-with-olm.yaml

Recommended in bug: https://bugzilla.redhat.com/show_bug.cgi?id=1757013

### DIFF
--- a/ocs_ci/ocs/resources/catalog_source.py
+++ b/ocs_ci/ocs/resources/catalog_source.py
@@ -1,0 +1,93 @@
+"""
+CatalogSource related functionalities
+"""
+import logging
+
+from ocs_ci.ocs.ocp import OCP
+from ocs_ci.ocs.exceptions import CommandFailed, ResourceInUnexpectedState
+from ocs_ci.utility.utils import TimeoutSampler
+from ocs_ci.utility.retry import retry
+
+
+logger = logging.getLogger(__name__)
+
+
+class CatalogSource(OCP):
+    """
+    This class represent CatalogSource and contains all related
+    methods we need to do with it.
+    """
+
+    def __init__(self, resource_name="", *args, **kwargs):
+        """
+        Initializer function for CatalogSource class
+
+        Args:
+            resource_name (str): Name of catalog source
+
+        """
+        super(CatalogSource, self).__init__(
+            resource_name=resource_name, kind='CatalogSource', *args, **kwargs
+        )
+
+    def check_state(self, state):
+        """
+        Check state of catalog source
+
+        Args:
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/resources/catalog_source.py
```diff
@@ -0,0 +1,93 @@
...
+        Raises:
+            ResourceInUnexpectedState: In case the catalog source is not in
+                expected state.
+
+        """
+        self.check_name_is_specified()
+        sampler = TimeoutSampler(
+            timeout, sleep, self.check_state, state=state
+        )
+        if not sampler.wait_for_func_status(True):
```
dahorak: For better readability, I would use named argument here:
```suggestion
        if not sampler.wait_for_func_status(result=True):
```

petr-balogh: I don't think it's blocking this PR and don't see any issue for using positional arguments when calling function.

petr-balogh:
Verified here:  https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/1226/console . The crucial part passed but finally failed cause of BZ  https://bugzilla.redhat.com/show_bug.cgi?id=1757486 , so considering as verified

RazTamir:
Is this considered as a workaround?
I see the comments in the bug this PR solves but not sure

petr-balogh:
> Is this considered as a workaround?
> I see the comments in the bug this PR solves but not sure

If this will work I think it's not considered as workaround but as the solution of the problem.

ocs_ci/ocs/resources/catalog_source.py
dahorak: I don't want to block this PR because of this... I was just slightly confused what that means (without knowing much about the `wait_for_func_status` method and `True` is quite general value here, to distinguish it's meaning blindly).
So otherwise LTGM and fell free to merge it as it is now.

RazTamir:
So do we expect to have this from the product side if this will be the fix for the issue?

petr-balogh:
> So do we expect to have this from the product side if this will be the fix for the issue?

Customer won't experience this issue from product side as they will install directly from the operator hub. But as we need to deploy some unreleased content we have to add catalog source pointing to our internal registry. So this is just for our deployment in our automation.

Once we will agreed that this solved the issue we will require the upstream to separate the subscription from deploy-with-olm.yaml and update the documentation. Then we will need to align the code on new approach.

Issue #848, opened 2019-10-02
Print consolidated config file


petr-balogh:
It was requested by engineering and will be useful for all of us to see what is the final config consolidated from all configs we pass via ocsci-conf parameter.

Infra ticket:
https://projects.engineering.redhat.com/browse/OCSQE-84

PR #850, opened 2019-10-02
Print consolidated config to the file

petr-balogh:
Fixes: #848

Infra ticket:
https://projects.engineering.redhat.com/browse/OCSQE-84

### DIFF
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -102,9 +102,11 @@ def pytest_configure(config):
     """
     if not (config.getoption("--help") or config.getoption("collectonly")):
         process_cluster_cli_params(config)
-        config_file = os.path.join(
-            ocsci_config.RUN['log_dir'],
-            f"run-{ocsci_config.RUN['run_id']}-config.yaml"
+        config_file = os.path.expanduser(
+            os.path.join(
+                ocsci_config.RUN['log_dir'],
+                f"run-{ocsci_config.RUN['run_id']}-config.yaml",
+            )
         )
         dump_config_to_file(config_file)
         log.info(
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -16,12 +16,12 @@ from ocs_ci.framework import config as ocsci_config
 from ocs_ci.framework.exceptions import ClusterPathNotProvidedError
 from ocs_ci.ocs.exceptions import CommandFailed
 from ocs_ci.utility.utils import (
+    dump_config_to_file,
     get_cluster_version,
     get_ceph_version,
     get_rook_version,
     get_csi_versions,
     get_testrun_name,
-    print_cofig_to_file,
 )
 from ocs_ci.ocs.utils import collect_ocs_logs
 
@@ -106,7 +106,7 @@ def pytest_configure(config):
             ocsci_config.RUN['log_dir'],
             f"run-{ocsci_config.RUN['run_id']}-config.yaml"
         )
-        print_cofig_to_file(config_file)
+        dump_config_to_file(config_file)
         log.info(
             f"Dump of the consolidated config file is located here: "
             f"{config_file}"
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -21,6 +21,7 @@ from ocs_ci.utility.utils import (
     get_rook_version,
     get_csi_versions,
     get_testrun_name,
+    print_cofig_to_file,
 )
 from ocs_ci.ocs.utils import collect_ocs_logs
 
@@ -101,6 +102,15 @@ def pytest_configure(config):
     """
     if not (config.getoption("--help") or config.getoption("collectonly")):
         process_cluster_cli_params(config)
+        config_file = os.path.join(
+            ocsci_config.RUN['log_dir'],
+            f"run-{ocsci_config.RUN['run_id']}-config.yaml"
+        )
+        print_cofig_to_file(config_file)
+        log.info(
+            f"Dump of the consolidated config file is located here: "
+            f"{config_file}"
+        )
         # Add OCS related versions to the html report and remove extraneous metadata
         markers_arg = config.getoption('-m')
         if ocsci_config.RUN['cli_params'].get('teardown') or (
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -7,6 +7,7 @@ import shlex
 import string
 import subprocess
 import time
+from copy import deepcopy
 from shutil import which
 
 import requests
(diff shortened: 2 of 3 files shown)


ocs_ci/utility/utils.py
```diff
@@ -1303,3 +1304,39 @@ def wait_for_co(operator):
     from ocs_ci.ocs.ocp import OCP
     ocp = OCP(kind='ClusterOperator')
     ocp.get(operator)
+
+
+def censore_values(data_to_censore, keys):
```
zmc: *censor

ocs_ci/utility/utils.py
```diff
@@ -1303,3 +1304,39 @@ def wait_for_co(operator):
...
+        data_to_censore (dict): data to censore
+        keys (list): keys where to censore the values in data dictionary
+
+    """
+    for key in keys:
+        if key in data_to_censore:
+            data_to_censore[key] = '*****'
+
+
+def print_cofig_to_file(file_path):
```
zmc: *config

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -101,6 +102,15 @@ def pytest_configure(config):
     """
     if not (config.getoption("--help") or config.getoption("collectonly")):
         process_cluster_cli_params(config)
+        config_file = os.path.join(
+            ocsci_config.RUN['log_dir'],
+            f"run-{ocsci_config.RUN['run_id']}-config.yaml"
+        )
+        print_cofig_to_file(config_file)
```
zmc: *config

ocs_ci/utility/utils.py
```diff
@@ -1303,3 +1304,39 @@ def wait_for_co(operator):
...
+    """
+    This function censores values in dictionary for passed keys
+
+    Args:
+        data_to_censore (dict): data to censore
+        keys (list): keys where to censore the values in data dictionary
+
+    """
+    for key in keys:
+        if key in data_to_censore:
```
zmc: *censor

ocs_ci/utility/utils.py
```diff
@@ -1303,3 +1304,39 @@ def wait_for_co(operator):
     from ocs_ci.ocs.ocp import OCP
     ocp = OCP(kind='ClusterOperator')
     ocp.get(operator)
+
+
+def censore_values(data_to_censore, keys):
+    """
+    This function censores values in dictionary for passed keys
```
zmc: *censor

ocs_ci/utility/utils.py
```diff
@@ -1303,3 +1304,39 @@ def wait_for_co(operator):
...
+
+    Args:
+        data_to_censore (dict): data to censore
+        keys (list): keys where to censore the values in data dictionary
```
zmc: *censor

ocs_ci/utility/utils.py
```diff
@@ -1303,3 +1304,39 @@ def wait_for_co(operator):
...
+    """
+    Print the config to the yaml file sith censored secret values.
+
+    Args:
+        file_path (str): Path to file where to write the configuration
+
+    """
+    config_copy = deepcopy(config.to_dict())
+    if config_copy.get('ENV_DATA'):
+        censore_values(
```
zmc: *censor

ocs_ci/utility/utils.py
zmc: Also, "print" doesn't seem like the correct word to use here. "write"?

ocs_ci/utility/utils.py
```diff
@@ -1303,3 +1304,39 @@ def wait_for_co(operator):
...
+            config_copy['ENV_DATA'],
+            [
+                'vsphere_password',
+                'ipam_token',
+                'httpd_server_password',
+            ]
+        )
+    with open(file_path, "w+") as fs:
+        yaml.safe_dump(config_copy, fs)
```
zmc: This function doesn't return anything

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -101,6 +102,15 @@ def pytest_configure(config):
...
+        log.info(
+            f"Dump of the consolidated config file is located here: "
+            f"{config_file}"
```
zmc: Writing the config contents to a file and then logging the filename doesn't seem like it'll be enough; most of us don't have access to the filesystem your builds are using.

ocs_ci/utility/utils.py
dahorak: > Also, "print" doesn't seem like the correct word to use here. "write"?

What about "dump" instead of print or write? (just suggestion)

ocs_ci/framework/pytest_customization/ocscilib.py
dahorak: If it will be in the `log_dir`, it should be accessible on the common NFS share available via http and linked in each build.
We can also consider to print it to the console in Jenkins job, if it will be helpful.

ocs_ci/utility/utils.py
```diff
@@ -1303,3 +1304,39 @@ def wait_for_co(operator):
...
+
+    """
+    for key in keys:
+        if key in data_to_censore:
+            data_to_censore[key] = '*****'
+
+
+def print_cofig_to_file(file_path):
+    """
+    Print the config to the yaml file sith censored secret values.
```
dahorak:
```suggestion
    Print the config to the yaml file with censored secret values.
```

ocs_ci/utility/utils.py
```diff
@@ -1303,3 +1304,39 @@ def wait_for_co(operator):
...
+
+    Args:
+        file_path (str): Path to file where to write the configuration
+
+    """
+    config_copy = deepcopy(config.to_dict())
+    if config_copy.get('ENV_DATA'):
```
dahorak: Are we sure, that "any" secret data will be in `ENV_DATA` and not in other sections?
Wouldn't be better to process the whole config structure recursively?

ocs_ci/utility/utils.py
dahorak: What do you think about some "automatic" process - hide any values for keys containing `passw` or `token`?

ocs_ci/utility/utils.py
petr-balogh: It doesn't and it's not suppose to return, just dump the config file

ocs_ci/utility/utils.py
petr-balogh: Thanks, I didn't check the dictionary and just blindly thought from word censored that it's censore but obviously it's not ;) Will correct on all places.

ocs_ci/utility/utils.py
```diff
@@ -1303,3 +1304,39 @@ def wait_for_co(operator):
...
+        censore_values(
+            config_copy['ENV_DATA'],
+            [
+                'vsphere_password',
+                'ipam_token',
+                'httpd_server_password',
+            ]
```
vavuthu: Could we define these keys in constants something like keys_to_censor

ocs_ci/utility/utils.py
petr-balogh: I did change of this function @dahorak @vavuthu , please take a look if you like this approach more.

ocs_ci/utility/utils.py

ocs_ci/utility/utils.py
```diff
@@ -1303,3 +1304,33 @@ def wait_for_co(operator):
     from ocs_ci.ocs.ocp import OCP
     ocp = OCP(kind='ClusterOperator')
     ocp.get(operator)
+
+
+def censor_values(data_to_censor, keys=None):
```
vavuthu: why do we need "keys=None" parameter. we can this since we are using constants.config_keys_patterns_to_censor

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -101,6 +102,15 @@ def pytest_configure(config):
     """
     if not (config.getoption("--help") or config.getoption("collectonly")):
         process_cluster_cli_params(config)
+        config_file = os.path.join(
+            ocsci_config.RUN['log_dir'],
+            f"run-{ocsci_config.RUN['run_id']}-config.yaml"
+        )
```
dahorak: the `config_file` should accept `~` in the path. (`os.path.expanduser(...)`)

petr-balogh:
I see file got created now here:
http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/jenkins-pr850/jenkins-pr850_20191003T082133/logs/run-1570107260-config.yaml

In this execution:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/1251/console

So marking as verified

petr-balogh:
@dahorak @vavuthu   can you take a look and approve if you are OK with that?

ocs_ci/utility/utils.py
petr-balogh: right, will do, sorry, missed it

PR #852, opened 2019-10-03
ceph_health_check: Accept HEALTH_WARN

zmc:
I understand we may or may not want this behavior, but this might be a good place to have the discussion.


RazTamir:
Hi Zack,

This is a workaround - could you please add a label to not merge this?

vavuthu:
what is affect on cluster in case of HEALTH_WARN/Degraded data redundancy . How much affect it has on our test cases

Issue #851, opened 2019-10-02
ValueError when iterating over generator returned from load_yaml with multi_document=True


clacroix12:
When attempting to iterate over the generator returned from `templating.load_yaml`, I am seeing the following error:

` ValueError: I/O operation on closed file.`

This is only happening when `multi_document=True` and when we are loading a local file (not from a url).

I was able to circumvent this issue by casting the generator to a list before returning it to the caller. Iterating over the list provides the dictionaries properly and does not return the error.

Output of local testing of load_yaml with a multi_document yaml file:
```
In [1]: from ocs_ci.utility.templating import load_yaml

In [2]: data = load_yaml('admission_test.yaml', multi_document=True)

In [3]: data
Out[3]: <generator object load_all at 0x1027d0b10>

In [4]: for i in data:
   ...:     print(i)
   ...:
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-4-278b5b14aaab> in <module>
----> 1 for i in data:
      2     print(i)
      3

~/workspace/ocs-ci/venv/lib/python3.7/site-packages/yaml/__init__.py in load_all(stream, Loader)
    125         Loader = FullLoader
    126
--> 127     loader = Loader(stream)
    128     try:
    129         while loader.check_data():

~/workspace/ocs-ci/venv/lib/python3.7/site-packages/yaml/loader.py in __init__(self, stream)
     32
     33     def __init__(self, stream):
---> 34         Reader.__init__(self, stream)
     35         Scanner.__init__(self)
     36         Parser.__init__(self)

~/workspace/ocs-ci/venv/lib/python3.7/site-packages/yaml/reader.py in __init__(self, stream)
     83             self.eof = False
     84             self.raw_buffer = None
---> 85             self.determine_encoding()
     86
     87     def peek(self, index=0):

~/workspace/ocs-ci/venv/lib/python3.7/site-packages/yaml/reader.py in determine_encoding(self)
    122     def determine_encoding(self):
    123         while not self.eof and (self.raw_buffer is None or len(self.raw_buffer) < 2):
--> 124             self.update_raw()
    125         if isinstance(self.raw_buffer, bytes):
    126             if self.raw_buffer.startswith(codecs.BOM_UTF16_LE):

~/workspace/ocs-ci/venv/lib/python3.7/site-packages/yaml/reader.py in update_raw(self, size)
    176
    177     def update_raw(self, size=4096):
[...]

PR #858, opened 2019-10-03
Read file before passing to loader for local yaml files

clacroix12:
Fixes: #851 

### DIFF
--- a/ocs_ci/utility/templating.py
+++ b/ocs_ci/utility/templating.py
@@ -156,7 +156,7 @@ def load_yaml(file, multi_document=False):
         return loader(get_url_content(file))
     else:
         with open(file, 'r') as fs:
-            return loader(fs)
+            return loader(fs.read())
 
 
 def get_n_document_from_yaml(yaml_generator, index=0):


clacroix12:
Local testing output with fix:

```
In [1]: from ocs_ci.utility.templating import load_yaml

In [2]: y = load_yaml('test.yaml', multi_document=True)

In [3]: type(y)
Out[3]: generator

In [4]: for i in y:
   ...:     print(i)
   ...:
{'apiVersion': 'v1', 'kind': 'Namespace', 'metadata': {'name': 'openshift-storage'}, 'spec': {}}
{'apiVersion': 'v1', 'kind': 'Namespace', 'metadata': {'name': 'local-storage'}, 'spec': {}}
{'apiVersion': 'operators.coreos.com/v1', 'kind': 'OperatorGroup', 'metadata': {'name': 'openshift-storage-operatorgroup', 'namespace': 'openshift-storage'}, 'spec': {'serviceAccount': {'metadata': {'creationTimestamp': None}}, 'targetNamespaces': ['openshift-storage']}}
{'apiVersion': 'operators.coreos.com/v1', 'kind': 'OperatorGroup', 'metadata': {'name': 'local-operator-group', 'namespace': 'local-storage'}, 'spec': {'serviceAccount': {'metadata': {'creationTimestamp': None}}, 'targetNamespaces': ['openshift-storage']}}
{'apiVersion': 'operators.coreos.com/v1alpha1', 'kind': 'CatalogSource', 'metadata': {'name': 'local-storage-manifests', 'namespace': 'openshift-marketplace'}, 'spec': {'description': 'An operator to manage local volumes', 'displayName': 'Local Storage Operator', 'icon': {'base64data': '', 'mediatype': ''}, 'image': 'quay.io/gnufied/local-registry:v4.2.0', 'publisher': 'Red Hat', 'sourceType': 'grpc'}}
{'apiVersion': 'operators.coreos.com/v1alpha1', 'kind': 'CatalogSource', 'metadata': {'name': 'ocs-catalogsource', 'namespace': 'openshift-marketplace'}, 'spec': {'displayName': 'Openshift Container Storage', 'icon': {'base64data': '', 'mediatype': ''}, 'image': 'quay.io/ocs-dev/ocs-registry:latest', 'publisher': 'Red Hat', 'sourceType': 'grpc'}}
{'apiVersion': 'operators.coreos.com/v1alpha1', 'kind': 'Subscription', 'metadata': {'name': 'ocs-subscription', 'namespace': 'openshift-storage'}, 'spec': {'channel': 'alpha', 'name': 'ocs-operator', 'source': 'ocs-catalogsource', 'sourceNamespace': 'openshift-marketplace'}}
```

vasukulkarni:
The examples I looked at show usage of with both file - open and open + read() , but if read is working better than just open so be it :)

clacroix12:
> The examples I looked at show usage of with both file - open and open + read()

Do you have a link to an example of passing the file directly? The ones I looked at were iterating over the generator inside of the `with` block, so the file was still open. Since we are returning the generator to the caller in our code, we are outside of the context manager and the file is closed before we try to iterate.

vasukulkarni:
> > The examples I looked at show usage of with both file - open and open + read()
> 
> Do you have a link to an example of passing the file directly? The ones I looked at were iterating over the generator inside of the `with` block, so the file was still open. Since we are returning the generator to the caller in our code, we are outside of the context manager and the file is closed before we try to iterate.

Here: https://www.programcreek.com/python/example/60272/yaml.safe_load_all

clacroix12:
@vasukulkarni yeah it looks like in any scenario where the raw file is passed the generator is used while the file is open. Since we are returning the generator to the caller, reading the content is simple enough and seems to work.

clacroix12:
I should clarify that this fixes an issue specifically with `multi_document=True` but the change applies to all local files passed to `load_yaml`. I will update the title.

Issue #866, opened 2019-10-04
must gather for OCP fails to run during deploy


mirekdlugosz:
`oc adm must-gather` requires `KUBECONFIG` environment variable to properly run. It is set by `OCP.set_kubeconfig()`, but **only** if path to file exists. This file is created early in `openshift-install create cluster` execution, but earliest `OCP.set_kubeconfig()` call is usually near the end of `deploy` method (through `test_cluster()`).

In other words, errors in `openshift-install` will trigger must gather, but without KUBECONFIG variable set. This causes must gather to fail.

How to reproduce:
1. In config, set `gather_on_deploy_failure` to True and `installer_version` to `4.2.0-0.nightly`
2. Ensure your pull-secret **does not** have `registry.svc.ci.openshift.org` entry (this is sure way to make installer fail)
3. Run `run-ci -m deployment <options> --deploy`

```
$ run-ci -m deployment --ocsci-conf ocs_ci/framework/conf/config.yaml --cluster-conf conf/ocs_basic_install.yml --cluster-name mzalewsk-test --cluster-path /home/mzalewsk/ocs-test/ --deploy
# snip
18:59:49 - MainThread - ocs_ci.utility.utils - INFO - Executing command: /home/mzalewsk/sources/ocs-ci/bin/openshift-install create cluster --dir /home/mzalewsk/ocs-test/ --log-level INFO
19:05:55 - MainThread - ocs_ci.ocs.utils - INFO - Must gather image: quay.io/openshift/origin-must-gather will be used.
19:05:55 - MainThread - ocs_ci.ocs.utils - INFO - OCS logs will be placed in location /tmp/failed_testcase_ocs_logs_1570208363/deployment_ocs_logs/ocp_must_gather
19:05:55 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc adm must-gather --image=quay.io/openshift/origin-must-gather --dest-dir=/tmp/failed_testcase_ocs_logs_1570208363/deployment_ocs_logs/ocp_must_gather
19:05:55 - MainThread - ocs_ci.ocs.utils - ERROR - Failed during must gather logs! Error: Error during execution of command: oc adm must-gather --image=quay.io/openshift/origin-must-gather --dest-dir=/tmp/failed_testcase_ocs_logs_1570208363/deployment_ocs_logs/ocp_must_gather.
                                                   Error is error: Missing or incomplete configuration info.  Please login or point to an existing, complete config file:
                                                   
[...]

PR #873, opened 2019-10-07
collect_ocs_logs(): Skip if there is no kubeconfig

zmc:
Fixes: #866

### DIFF
--- a/ocs_ci/ocs/utils.py
+++ b/ocs_ci/ocs/utils.py
@@ -720,6 +720,16 @@ def collect_ocs_logs(dir_name, ocp=True, ocs=True):
         ocs (bool): Whether to gather OCS logs
 
     """
+    if not (
+        'KUBECONFIG' in os.environ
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/utils.py
```diff
@@ -720,6 +720,12 @@ def collect_ocs_logs(dir_name, ocp=True, ocs=True):
         ocs (bool): Whether to gather OCS logs
 
     """
+    if not ('KUBECONFIG' in os.environ
```
clacroix12: If we are breaking into multiple lines the parenthesis should be on their own lines.

```python
if not (
    'KUBECONFIG' in os.environ
    or os.path.exists(os.path.expanduser('~/.kube/config'))
):
```

same for the log statement
```python
log.warn(
    "Cannot find $KUBECONFIG or ~/.kubeconfig/config; "
    "skipping log collection"
)
```

Issue #844, opened 2019-10-01
Teardown fails on missing metadata.json file


mbukatov:
## Description

When I tried to run teardown of a cluster, the process failed on `FileNotFoundError: [Errno 2] No such file or directory: '/home/ocsqe/data/cluster-2019-10-01.1/metadata.json'`.

This is a regression.

## Version

I used ocs-ci commit e070975

## Reproducer

1) I recreated python virtualenv and reinstalled ocs-ci
2) I deployed a downstream cluster via run-ci as usual:

```
$ run-ci --cluster-conf conf/ocsci/downstream_config.yaml --cluster-path ~/data/cluster-2019-10-01.1 --cluster-name mbukatov-ocsqe --deploy -m deployment | tee ~/data/cluster-2019-10-01.1.deploy.log
```
3) I used the cluster for a while.
4) I started a teardown of the cluster via:

```
$ run-ci --cluster-path ~/data/cluster-2019-10-01.1 --teardown -m deployment | tee ~/data/cluster-2019-10-01.1.teardown.log
```

## Actual Results

The process failed on:

```
20:23:28 - MainThread - ocs_ci.deployment.aws - ERROR - Traceback (most recent call last):
                                                          File "/home/ocsqe/projects/ocs-ci-deploy/ocs_ci/deployment/aws.py", line 110, in destroy_volumes
                                                            cluster_name = get_cluster_name(self.cluster_path)
                                                          File "/home/ocsqe/projects/ocs-ci-deploy/ocs_ci/deployment/aws.py", line 399, in get_cluster_name
                                                            with open(metadata_file) as f:
                                                        FileNotFoundError: [Errno 2] No such file or directory: '/home/ocsqe/data/cluster-2019-10-01.1/metadata.json'


================ 1 passed, 190 deselected in 458.77s (0:07:38) =================
```

The cluster directory ~is~ looks now empty, there is just a ocp install log left:

```
$ ls -la ~/data/cluster-2019-10-01.1
total 196
drwxrwxr-x. 2 ocsqe ocsqe     36 Oct  1 20:23 .
drwxrwxr-x. 4 ocsqe ocsqe  16384 Oct  1 20:47 ..
-rw-rw-r--. 1 ocsqe ocsqe 179249 Oct  1 20:23 .openshift_install.log
```

## Expected results

The teardown run finishes with success.

Cluster directory `~/data/cluster-2019-10-01.1` is not completely empty.

mbukatov:
[...]

PR #900, opened 2019-10-14
Read clusterName from deployment metadata instead of from file

clacroix12:
Fixes: #844 


petr-balogh:
I see that destroy job passed here:  https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-destroy-ocs-cluster/1126/console so marking as verified

PR #925, opened 2019-10-22
Add config for us-east-1 region

petr-balogh:
As this issue was found:
https://github.com/noobaa/noobaa-operator/issues/90
We would like to test with us-east-1 region to see if we have other
issues.

### DIFF
--- a/conf/ocsci/aws_region_us-east-1.yaml
+++ b/conf/ocsci/aws_region_us-east-1.yaml
@@ -1,0 +1,4 @@
+---
+ENV_DATA:
+  platform: 'aws'
+  region: 'us-east-1'


conf/ocsci/aws_region_us-east-1.yaml
```diff
@@ -0,0 +1,4 @@
+---
+ENV_DATA:
+  platform: 'aws'
```
dahorak: In the [default_config.yaml](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L80) and also in [marks.py](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/pytest_customization/marks.py#L62:L65), the platform value is upper-case `AWS` so even though in other config files it is used with lower-case, it might be better/safer to set it upper-case here.

dahorak: Since issue https://github.com/red-hat-storage/ocs-ci/issues/926 was fixed (PR https://github.com/red-hat-storage/ocs-ci/pull/927 is merged), we can probably ignore this note.

PR #932, opened 2019-10-23
Allow custom SMTP server

zmc:


.functional_ci_setup.py
```diff
@@ -101,6 +101,8 @@ def get_ocsci_conf():
     )
     if env.get("DOWNSTREAM") == "true":
         conf_obj['REPORTING']['us_ds'] = 'DS'
+    if env.get("SMTP_SERVER"):
+        conf_obj['REPORTING']['email'] = dict(smtp_server=env['SMTP_SERVER'])
```
clacroix12: Won't this essentially remove `conf_obj['REPORTING']['email']['address']`?

zmc: Nope, the config system doesn't work that way (unit tests prove that)

zmc: If you want to verify:
```

my_conf = dict(REPORTING=dict(email=dict(smtp_server='my_server')))
config.update(my_conf)
assert config.REPORTING['email']['smtp_server'] == 'my_server'
assert config.REPORTING['email']['address'] == 'ocs-ci@redhat.com'
```

clacroix12: Oh I understand now, this is for a custom config that's updating the defaults. I was thinking this was manipulating the actual config object.

PR #939, opened 2019-10-24
Corrupted Placement Group alert test case

fbalak:
- Add `test_corrupt_pg_alerts` test case.
- Add `create_dummy_osd` function. This can be used for Ceph storage manipulation as by using osd created by this function there is no lock on osd data.

### DIFF
--- a/tests/manage/monitoring/conftest.py
+++ b/tests/manage/monitoring/conftest.py
@@ -2,11 +2,8 @@ import json
 import logging
 import os
 import pytest
-from subprocess import TimeoutExpired
 import threading
-import tempfile
 import time
-import yaml
 
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants, ocp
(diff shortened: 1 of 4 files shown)


fbalak:
Logs: https://paste.fedoraproject.org/paste/uBBiD0MOG0C816j6sRJcmQ

tests/manage/monitoring/conftest.py
```diff
@@ -342,3 +347,150 @@ def stop_osd():
     oc.exec_oc_cmd(f"scale --replicas=1 deployment/{osd_to_stop}")
 
     return measured_op
+
+
+def create_dummy_osd(deployment):
```
ebenahar: Since it's not a fixture, thinking it would be better to have this function in another file, like helpers.py

mbukatov: No, please don't add functions into helpers file withing a test directory tree.

mbukatov: You could add it into ocsci module.

fbalak:
@ebenahar Yes, it's done on lines starting on https://github.com/red-hat-storage/ocs-ci/pull/939/files#diff-a650626eeae1cbc843005d30e3a6c7feR415. It is done after the workload that is tested is done.

vasukulkarni:
Nice to see this test!

Issue #942, opened 2019-10-25
ocs-ci Checks for OCS even when skip_ocs_deployment: true


nehaberry:
I installed OCP using CI by editing default_config.yaml to have  "skip_ocs_deployment: true" .

OCP installed successfully and OCS install was skipped. But the check for OCS was still ON, due to which the ocs-ci run failed eventually with an assert.

```
11:04:59 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/nberry/aws-install/oct25-1/auth/kubeconfig get StorageCluster  -n openshift-storage -o yaml
FAILED 

E           ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc -n openshift-storage --kubeconfig /home/nberry/aws-install/oct25-1/auth/kubeconfig get StorageCluster  -n openshift-storage -o yaml.
E           Error is error: the server doesn't have a resource type "StorageCluster"

ocs_ci/utility/utils.py:410: CommandFailed

```

Complete Log

```
10:36:54 - MainThread - ocs_ci.deployment.aws - INFO - Deploying OCP cluster
10:36:54 - MainThread - ocs_ci.deployment.aws - INFO - Openshift-installer will be using loglevel:INFO
10:36:54 - MainThread - ocs_ci.utility.utils - INFO - Executing command: /home/nberry/ocs-ci-scripts-master/ocs-ci-scripts/ocs-ci/bin/openshift-install create cluster --dir /home/nberry/aws-install/oct25-1/ --log-level INFO
11:04:51 - MainThread - ocs_ci.utility.utils - WARNING - Command warning: level=info msg="Consuming \"Install Config\" from target directory"
                                                         level=info msg="Creating infrastructure resources..."
                                                         level=info msg="Waiting up to 30m0s for the Kubernetes API at https://api.nberry-oct25-1.qe.rh-ocs.com:6443..."
                                                         level=info msg="API v1.14.6+868bc38 up"
                                                         level=info msg="Waiting up to 30m0s for bootstrapping to complete..."
                                                         level=info msg="Destroying the bootstrap resources..."
                                                         level=info msg="Waiting up to 30m0s for the cluster at https://api.nberry-oct25-1.qe.rh-ocs.com:6443 to initialize..."
                                                         level=info msg="Waiting up to 10m0s for the openshift-console route to be created..."
                                                         level=info msg="Install complete!"
[...]

PR #944, opened 2019-10-25
check for skip_ocs_installation in deployment

vavuthu:
check for skip_ocs_installation in deployment

Fixes: #942

### DIFF
--- a/tests/ecosystem/deployment/test_deployment.py
+++ b/tests/ecosystem/deployment/test_deployment.py
@@ -16,7 +16,11 @@ def test_deployment():
     deploy = config.RUN['cli_params'].get('deploy')
     teardown = config.RUN['cli_params'].get('teardown')
     if not teardown or deploy:
-        ocs_install_verification()
+        if config.ENV_DATA['skip_ocs_deployment']:
+            # Verify cluster is running
+            assert is_cluster_running(config.ENV_DATA['cluster_path'])
+        else:
+            ocs_install_verification()
 
     if teardown:
         log.info(


tests/ecosystem/deployment/test_deployment.py
```diff
@@ -16,7 +16,11 @@ def test_deployment():
     deploy = config.RUN['cli_params'].get('deploy')
     teardown = config.RUN['cli_params'].get('teardown')
     if not teardown or deploy:
-        ocs_install_verification()
+        if config.ENV_DATA['skip_ocs_deployment']:
+            # Verify cluster is running
+            assert is_cluster_running(config.ENV_DATA['cluster_path'])
```
clacroix12: Since we are also doing this in `ocs_install_verification` would it make sense for us to check `config.ENV_DATA['skip_ocs_deployment']` after that has been done and return if true? Thus skipping the rest of the verification which seems like the goal here.

For example in `ocs_install_verification`:
```python
log.info("Verifying OCP installation")
assert is_cluster_running(config.ENV_DATA['cluster_path'])
if config.ENV_DATA['skip_ocs_deployment']:
    return
log.info("Verifying OCS installation")
...
```

zmc: IMO even adding wording that suggests OCP verification happens in `ocs_install_verification` can lead to confusion; another way to avoid the redundancy is to remove the call to `is_cluster_running` from `ocs_install_verification`, and replace this `if/else` block with:
```
assert is_cluster_running(config.ENV_DATA['cluster_path'])
if not config.ENV_DATA['skip_ocs_deployment']:
    ocs_install_verification
```

zmc:
Marked as high priority because the issue is blocking our build verification.

tests/ecosystem/deployment/test_deployment.py
clacroix12: @zmc forgot to mention in my suggestion we could also rename to just `install_verification` but I like your approach more.

vasukulkarni:
> Marked as high priority because the issue is blocking our build verification.

@zmc I dont think you are skipping ocs installation, I think this PR addresses only when ocs installation is skipped

zmc:
@vasukulkarni we are.

vasukulkarni:
@vavuthu sorry, I guess due to comments and priority, the PR is now overtaken by https://github.com/red-hat-storage/ocs-ci/pull/946

clacroix12:
Closing since we merged #946

PR #949, opened 2019-10-29
Must gather tier1 tests

AviadP:


tests/manage/cluster/test_must_gather.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        Check if directory to store must gather logs already exist
+        and use new directory if so.
+
+        Returns:
+            str: Logs directory
+
+        """
+        index = 1
+        directory = f"/tmp/mg_test{index}"
```
vasukulkarni: please import config and use config.RUN['log_dir']

tests/manage/cluster/test_must_gather.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @pytest.fixture(autouse=True)
+    def teardown(self, request):
+
+        def finalizer():
+            must_gather_pods = self.ocs.get_pods(label_selector='app=must-gather')
+            logger.info(f"must_gather_pods: {must_gather_pods} ")
+            if must_gather_pods:
+                for pod_to_del in must_gather_pods:
+                    self.ocp_obj.wait_for_delete(resource_name=pod_to_del)
```
vasukulkarni: We are not setting up the pods, why delete them?

tests/manage/cluster/test_must_gather.py
AviadP: changed to  utilsocsci_log_path() instead - based on config.RUN['log_dir']

tests/manage/cluster/test_must_gather.py
AviadP: it seems that TestManage tear down does not delete this pod in time, so this is a double check

ebenahar: @vasukulkarni if I'm not mistaken, a pod is being created for must gather logs collection. It's being destroyed when must gather is done. So here, we wait for that pod to get deleted so leftover checks will be happy

tests/manage/cluster/test_must_gather.py
```diff
@@ -0,0 +1,182 @@
...
+                    logger.info(f"deleted pods: {pod_to_del}")
```
ebenahar: since you're not actively deleting the must gather pod, this log message and the pod variable name can be confusing. I'd rephrase the pod var to 'must_gather_pod' and the log msg to "Pod {must_gather_pod} got deleted successfully"

tests/manage/cluster/test_must_gather.py
```diff
@@ -0,0 +1,182 @@
...
+                    time.sleep(3)
```
ebenahar: please avoid sleep. Try to understand why we need to wait here

tests/manage/cluster/test_must_gather.py
vasukulkarni: Thanks, I see that now.

tests/manage/cluster/test_must_gather.py
```diff
@@ -0,0 +1,181 @@
...
+        self.ocp_obj = ocp.OCP()
+
+    @pytest.fixture(autouse=True)
+    def teardown(self, request):
+
+        def finalizer():
+            must_gather_pods = self.ocs.get_pods(label_selector='app=must-gather')
+            logger.info(f"must_gather_pods: {must_gather_pods} ")
+
+            while must_gather_pods:
```
ebenahar: this could lead to an infinite loop.
Please use the TimeoutSampler form instead of while loop

AviadP:
verified locally due to jenkins PR tester being unstable

PR #950, opened 2019-10-29
Cleanup of old non OCS operator deployment code

clacroix12:
- Removed ocs_operator_deployment config
- Removed non ocs_operator_deployment code paths from deployments
- Removed now unused ocs-deployment yaml files
- Removed now unused monitoring yaml file

Jira: https://projects.engineering.redhat.com/browse/OCSQE-73


ocs_ci/deployment/deployment.py
```diff
@@ -160,7 +155,8 @@ def label_and_taint_nodes(self):
             )
             _ocp.exec_oc_cmd(command=taint_cmd)
 
-    def get_olm_and_subscription_manifest(self):
+    @staticmethod
```
vasukulkarni: If you make it static now, usage will change how its called now, but I dont see that code change?

clacroix12: Usage doesn't change, just that it doesn't take `self` as the first argument. Since this method isn't actually using that reference anyways, this shouldn't change anything.

clacroix12:
@zmc @andrewschoen I added you guys on here to make sure my changes to `.functional_ci_setup.py` make sense given the rest of the changes in this PR.

ocs_ci/deployment/deployment.py
clacroix12: I'm fairly certain it's fine as it is. You can test this locally with a dummy class that has a staticmethod and another method that calls it.

```python
In [1]: class MyClass:
   ...:     @staticmethod
   ...:     def do_something():
   ...:         print('hello from static method')
   ...:     def test_method(self):
   ...:         self.do_something()
   ...:
   ...:

In [2]: myclass = MyClass()

In [3]: myclass.test_method()
hello from static method
```

vasukulkarni: I would rather modify to make it clear its a class ref and be explicit about it

```python
class MyClass:
    @staticmethod
    def do_something():
        print('hello from static method')
    def test_method(self):
        MyClass.do_something()


myclass = MyClass()
myclass.test_method()
```

clacroix12: I think you might be confusing `@staticmethod` with `@classmethod`. But regardless, I only added the decorator because we don't reference `self` at all in the method and pycharm was complaining. I figured I might as well change it while I was changing other code in that file. It has no real relevance to the actual goal of the PR and if it's going to cause issues I'd rather just remove the change entirely.

vasukulkarni: They are almost same, one which can access class state other cannot and they both differ from normal methods.
But if you added due to pycharm complain maybe you didn't intend to use it as static(or a standalone method independent of instance method), then we can remove the staticmethod, Otherwise its LGTM.

clacroix12: It was intentional. I changed it to a `staticmethod` because the method doesn't utilize the instance of class in any way. I also didn't want to change the usage, otherwise I could've just as easily moved it to a module level function.

That being said I'm still not sure why we would want to change the usage from `self.method_name()` to `Class.method_name()`. Especially since we are calling it from within that very class. In ocs-ci we are already calling static methods through `self` and through objects, not the `Class` itself. Ex: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/cluster.py#L139-L158

Other references show the same thing:
https://stackoverflow.com/questions/136097/what-is-the-difference-between-staticmethod-and-classmethod

vasukulkarni: @clacroix12 In that thread you linked, it says neither self. nor cls. is passed as an argument and my intention to use cls is atleast denote its a *standalone* function and has no link to self. 

Since you are modifying this function, I think its fair to assume to fix the references where this function is used to properly indicate what kind of that method is. Personally I dont use *staticmethods*(and It play no role in inheritance) and prefer them in different module,  There are various examples I found where they use class to invoke a staticmethod

https://realpython.com/instance-class-and-static-methods-demystified/
```python
>>> p = Pizza(4, ['mozzarella', 'tomatoes'])
>>> p
Pizza(4, ['mozzarella', 'tomatoes'])
>>> p.area()
50.26548245743669
>>> Pizza.circle_area(4)
50.26548245743669
```

clacroix12: Yes, you absolutely _can_ use the class to invoke a static method. In your example you are calling `Pizza.circle_area(4)` from outside of the classes definition, so it makes sense that you would reference the method through the Class itself. Where we are calling `self.get_olm_and_subscription_manifest()` is from _inside_ of the class definition. This is consistent with how we are already calling static methods in our code from the same scope. The only places where we use the class to invoke a static method are from outside of the class definition.

vasukulkarni: What I was trying to get there is since its independent, self has no meaning (nor does cls -but it needs a namespace for the function to be called and cls makes more sense IMHO). we might have missed it in previous review but you cannot claim consistency when we also have https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/ocp.py#L162 :)

vasukulkarni: Also I dont understand what difference it makes where its called (inside class vs outside).

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -51,7 +51,6 @@ DEPLOYMENT:
   # This is described as a WA for minimal configuration 3/3 worker/master
   # nodes. See: https://github.com/openshift/ocs-operator
   ocs_operator_nodes_to_taint: 0
-  ocs_operator_deployment: True
   ssh_key: "~/.ssh/openshift-dev.pub"
   force_deploy_multiple_clusters: False
   allow_lower_instance_requirements: true
```
petr-balogh: Cannot comment on lower lines, but line: 96 ceph_image, do we still need it there?

vasukulkarni: @clacroix12 Are you planning to address this? we can still use old approvals?

clacroix12: @vasukulkarni @petr-balogh I checked for usages of `ceph_image` and it doesn't look like we need it anymore. I removed it locally and am running a local deployment test. If all goes well I will update the PR.

clacroix12: Fixed in latest commit

clacroix12:
@andrewschoen @zmc I made some additional changes to `. functional_ci_setup.py` after removing some more configuration. Please let me know if this doesn't look right.

Issue #952, opened 2019-10-30
check whether osd/devicesets are distrubuted in all AZ's


vavuthu:
check whether osd/devicesets are distrubuted in all AZ's

PR #953, opened 2019-10-30
check to verify OSD's are distributed among worker nodes

vavuthu:
check to verify OSD's are distributed among worker nodes

Fixes: #952

### DIFF
--- a/tests/ecosystem/deployment/test_deployment.py
+++ b/tests/ecosystem/deployment/test_deployment.py
@@ -2,6 +2,7 @@ import logging
 
 from ocs_ci.framework import config
 from ocs_ci.framework.testlib import deployment, destroy
+from ocs_ci.ocs import defaults
 from ocs_ci.ocs import constants, ocp
 from ocs_ci.utility.utils import is_cluster_running, ceph_health_check
 
@@ -162,3 +163,13 @@ def ocs_install_verification():
         item['metadata']['name'] for item in storage_classes['items']
     }
     assert required_storage_classes.issubset(storage_class_names)
+
+    # Verify OSD's are distributed
+    log.info("Verifying OSD's are distributed evenly across worker nodes")
+    ocp_pod_obj = ocp.OCP(kind=constants.POD, namespace=defaults.ROOK_CLUSTER_NAMESPACE)
+    osds = ocp_pod_obj.get(selector='app=rook-ceph-osd')['items']
+    node_names = [osd['spec']['nodeName'] for osd in osds]
+    for node in node_names:
+        assert not node_names.count(node) > 1, (
+            "OSD's are not distributed evenly across worker nodes"
+        )
--- a/tests/ecosystem/deployment/test_deployment.py
+++ b/tests/ecosystem/deployment/test_deployment.py
@@ -2,7 +2,6 @@ import logging
 
 from ocs_ci.framework import config
 from ocs_ci.framework.testlib import deployment, destroy
-from ocs_ci.ocs import defaults
 from ocs_ci.ocs import constants, ocp
 from ocs_ci.utility.utils import is_cluster_running, ceph_health_check
 
--- a/tests/ecosystem/deployment/test_deployment.py
+++ b/tests/ecosystem/deployment/test_deployment.py
@@ -166,8 +166,8 @@ def ocs_install_verification():
 
     # Verify OSD's are distributed
     log.info("Verifying OSD's are distributed evenly across worker nodes")
-    ocp_pod_obj = ocp.OCP(kind=constants.POD, namespace=defaults.ROOK_CLUSTER_NAMESPACE)
-    osds = ocp_pod_obj.get(selector='app=rook-ceph-osd')['items']
+    ocp_pod_obj = ocp.OCP(kind=constants.POD, namespace=namespace)
+    osds = ocp_pod_obj.get(selector=constants.OSD_APP_LABEL)['items']
     node_names = [osd['spec']['nodeName'] for osd in osds]
     for node in node_names:
         assert not node_names.count(node) > 1, (


tests/ecosystem/deployment/test_deployment.py
```diff
@@ -162,3 +163,13 @@ def ocs_install_verification():
...
     }
     assert required_storage_classes.issubset(storage_class_names)
+
+    # Verify OSD's are distributed
+    log.info("Verifying OSD's are distributed evenly across worker nodes")
+    ocp_pod_obj = ocp.OCP(kind=constants.POD, namespace=defaults.ROOK_CLUSTER_NAMESPACE)
+    osds = ocp_pod_obj.get(selector='app=rook-ceph-osd')['items']
+    node_names = [osd['spec']['nodeName'] for osd in osds]
+    for node in node_names:
+        assert not node_names.count(node) > 1, (
```
vasukulkarni: This check needs to change, I can setup a cluster with 2 or more OSDs, so we should handle that.

vavuthu: > This check needs to change, I can setup a cluster with 2 or more OSDs, so we should handle that.

I was under impression that installation will comes with 1 OSD per node unless we do cluster expansion expansion ( not part of instalation).
@nehaberry : Is my assumption correct

tests/ecosystem/deployment/test_deployment.py
```diff
@@ -162,3 +163,13 @@ def ocs_install_verification():
         item['metadata']['name'] for item in storage_classes['items']
     }
     assert required_storage_classes.issubset(storage_class_names)
+
+    # Verify OSD's are distributed
+    log.info("Verifying OSD's are distributed evenly across worker nodes")
+    ocp_pod_obj = ocp.OCP(kind=constants.POD, namespace=defaults.ROOK_CLUSTER_NAMESPACE)
+    osds = ocp_pod_obj.get(selector='app=rook-ceph-osd')['items']
+    node_names = [osd['spec']['nodeName'] for osd in osds]
+    for node in node_names:
```
prsurve: we need to check with worker node that is labeled for ocs, I guess this might fail if we have more then 3 worker

tests/ecosystem/deployment/test_deployment.py
```diff
@@ -162,3 +163,13 @@ def ocs_install_verification():
         item['metadata']['name'] for item in storage_classes['items']
     }
     assert required_storage_classes.issubset(storage_class_names)
+
+    # Verify OSD's are distributed
+    log.info("Verifying OSD's are distributed evenly across worker nodes")
+    ocp_pod_obj = ocp.OCP(kind=constants.POD, namespace=defaults.ROOK_CLUSTER_NAMESPACE)
```
petr-balogh: Instead of defaults.ROOK_CLUSTER_NAMESPACE use cluster name from config.ENV_DATA section

tests/ecosystem/deployment/test_deployment.py
```diff
@@ -162,3 +163,13 @@ def ocs_install_verification():
...
+    osds = ocp_pod_obj.get(selector='app=rook-ceph-osd')['items']
```
petr-balogh: Selector is defined already in ocs_ci/ocs/constants.py OSD_APP_LABEL = "app=rook-ceph-osd"

vavuthu: Done

tests/ecosystem/deployment/test_deployment.py
vavuthu: Done

tests/ecosystem/deployment/test_deployment.py
vavuthu: If we have more than 3 workers, and extra worker is part OCS cluster, only OSD will be created in 4th node. here we are making sure each osd is created on unique node for deployments. So this logic will apply for all AZ's as well as for all platforms.

tests/ecosystem/deployment/test_deployment.py
vasukulkarni: @vavuthu I couldnt find how to define in the config as well, but I guess its basic and we will have that.  If you output osd tree from ceph-tools-pod in json format, its easy to check no matter how many osd's each node will have.

vasukulkarni:
```
# ceph osd tree --format json

{"nodes":[{"id":-1,"name":"default","type":"root","type_id":11,"children":[-5]},{"id":-5,"name":"us-east-2","type":"region","type_id":10,"pool_weights":{},"children":[-14,-4,-10]},{"id":-10,"name":"us-east-2a","type":"zone","type_id":9,"pool_weights":{},"children":[-9]},{"id":-9,"name":"example-deviceset-2-0-85zb5","type":"host","type_id":1,"pool_weights":{},"children":[1]},{"id":1,"device_class":"ssd","name":"osd.1","type":"osd","type_id":0,"crush_weight":0.0966949462890625,"depth":4,"pool_weights":{},"exists":1,"status":"up","reweight":1,"primary_affinity":1},{"id":-4,"name":"us-east-2b","type":"zone","type_id":9,"pool_weights":{},"children":[-3]},{"id":-3,"name":"example-deviceset-0-0-bfdc9","type":"host","type_id":1,"pool_weights":{},"children":[0]},{"id":0,"device_class":"ssd","name":"osd.0","type":"osd","type_id":0,"crush_weight":0.0966949462890625,"depth":4,"pool_weights":{},"exists":1,"status":"up","reweight":1,"primary_affinity":1},{"id":-14,"name":"us-east-2c","type":"zone","type_id":9,"pool_weights":{},"children":[-13]},{"id":-13,"name":"example-deviceset-1-0-nht7p","type":"host","type_id":1,"pool_weights":{},"children":[2]},{"id":2,"device_class":"ssd","name":"osd.2","type":"osd","type_id":0,"crush_weight":0.0966949462890625,"depth":4,"pool_weights":{},"exists":1,"status":"up","reweight":1,"primary_affinity":1}],"stray":[]}
```

vavuthu: > If you output osd tree from ceph-tools-pod in json format, its easy to check no matter how many osd's each node will have

If we use ceph osd tree, logic will be more complicated because, 1st we need to find how many AZ are there in cluster and check whether all AZ exists in osd tree. Again traverse each region to check osd is created or not. If its vsphere platform flow will be different since its has single AZ.

nehaberry: @vavuthu +1  you are correct. As part of deployment, we are only allowed to create 3 OSDs( 1 per node). Any extra OSD per node will only be possible as expansion workflow(which is after a successful deployment)

See mail **No option to select #OSDs during OCS install** where it was confirmed by engg.

vasukulkarni: I mean osd tree from json logic, (eg: just getting osd's per zone and storing in list and checking if all the numbers in the list are same) , I was thinking VMware should also use the same zone logic?

Bur since we dont have this in UI, I am not going to block you on this :) , I am resolving this.

tests/ecosystem/deployment/test_deployment.py
vasukulkarni: Is this really resolved? @prsurve ?

prsurve: yes

PR #959, opened 2019-10-31
Setting up the environment variable of KUBECONFIG, only if it is not

Avilir:
defined by the user, outside the script.

### DIFF
--- a/ocs_ci/ocs/openshift_ops.py
+++ b/ocs_ci/ocs/openshift_ops.py
@@ -142,7 +142,9 @@ class OCP(object):
                 "The kubeconfig file %s doesn't exist!", kubeconfig_path
             )
             return False
-        os.environ['KUBECONFIG'] = kubeconfig_path
+        # Setting environment variable on ly if not exist
+        if 'KUBECONFIG' not in os.environ:
+            os.environ['KUBECONFIG'] = kubeconfig_path
         try:
             run_cmd("oc cluster-info")
         except CommandFailed as ex:


ocs_ci/ocs/openshift_ops.py
```diff
@@ -142,7 +142,9 @@ def set_kubeconfig(kubeconfig_path):
                 "The kubeconfig file %s doesn't exist!", kubeconfig_path
             )
             return False
-        os.environ['KUBECONFIG'] = kubeconfig_path
+        # Setting environment variable on ly if not exist
+        if 'KUBECONFIG' not in os.environ:
```
petr-balogh: We don't want probably this behavior as this kubeconfig is taken from cluster-dir, so it means that you can have by mistake exported kubeconfig from some previous cluster and than ocs-ci will use this one instead one from cluster dir which is supposed to be used. So I would rather use the behavior as we have and set it all the time with kubeconfig from cluster dir and as it's mandatory parameter we will be sure that we are running against correct cluster.

So please close this PR, if you think that it's needed, please open issue with proper description. We should link the issue to the PR than if it will be decided to do it like that with some good reason.

Thanks

Avilir: Yes, but it must be full path and not relative, since when we are deploying test operators like ripsaw, it install from different directory and the the KUBEConfig pas is incorrect.

vasukulkarni: I agree, we probably dont want this behaviour, If you want to use environment based kubeconfig, you can create a cluster dir and place that kubeconfig in that dir(as a workaround). But by default using environment will have unintended side affects when using with ocs-ci.

PR #960, opened 2019-11-01
Upgrade test

petr-balogh:
Adding upgrade test + all related changes.

Ecosystem ticket:
https://projects.engineering.redhat.com/browse/OCSQE-67


conf/ocsci/upgrade.yaml
```diff
@@ -0,0 +1,8 @@
+---
+# This config file is for the upgrade purpose
+DEPLOYMENT:
+  upgrade: true
+  # In teh case you need to upgrade to the latest build, the following
```
dahorak:
```suggestion
  # In the case you need to upgrade to the latest build, the following
```

petr-balogh: done

petr-balogh:
Running verification job here:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/1846/console

docs/upgrade.md
```diff
@@ -0,0 +1,51 @@
...
+## Upgrade execution
+
+For upgrade we have few pytest marks defined [here](/ocs_ci/framework/pytest_customization/marks.py):
+
+* **before_upgrade** - mark tests which are suppose to run before upgrade
+* **upgrade** - this is mark for upgrade itself
+* **after_upgrade** - mark tests which are suppose to run after upgrade
+
+Those marks has to be imported from mentioned
+[modul](/ocs_ci/framework/pytest_customization/marks.py) and your test have to
```
dahorak:
```suggestion
[module](/ocs_ci/framework/pytest_customization/marks.py) and your test have to
```

docs/upgrade.md
```diff
@@ -0,0 +1,51 @@
...
+
+For this purpose we have this config file: [upgrade.yaml](/conf/ocsci/upgrade.yaml)
+which you can pass via `--ocsci-conf` when you are running the deployment. This will
+automatically take not latest build but one before.
+
+## Upgrade execution
+
+For upgrade we have few pytest marks defined [here](/ocs_ci/framework/pytest_customization/marks.py):
+
+* **before_upgrade** - mark tests which are suppose to run before upgrade
```
vasukulkarni: I think we should remove both `before_upgrade` and `after_upgrade` until there is full clarity on what the usecase is,  Right now most of us will need -m upgrade to upgrade from existing cluster and run our tests.  The additional marks will create confusions. We can add them later when really needed.

ocs_ci/ocs/constants.py
```diff
@@ -364,3 +364,9 @@
 
 # Config related constants
 config_keys_patterns_to_censor = ['passw', 'token', 'secret']
+
+# Upgrade related constants, keeping some space between, so we can add
+# additional order.
+ORDER_BEFORE_UPGRADE = 1
+ORDER_UPGRADE = 10
+ORDER_AFTER_UPGRADE = 20
```
dahorak: Wouldn't be slightly nicer to use `10`, `20` and `30`?
```python
ORDER_BEFORE_UPGRADE = 10
ORDER_UPGRADE = 20
ORDER_AFTER_UPGRADE = 30
```

docs/upgrade.md
```diff
@@ -0,0 +1,51 @@
...
+    --cluster-path /home/my_user/my-ocs-dir \
+    -m 'before_upgrade or upgrade or after_upgrade'
+ ```
+
+Of course if you would like to run just the upgrade and not any other tests
+before and after upgrade you can use just `-m upgrade`
+
+## TODO
+
+Once there will be 4.3 build and we will have different branch for OCS-CI we
```
vasukulkarni: As said in previous comment, I dont see the usecase for separate branch or before and after_upgrade marks, It adds complexity.

docs/upgrade.md
petr-balogh: In RHV we had much more complex ordering for upgrade tests. This is is basically prepared infrastructure which allow ppl to write upgrade related tests which will be executed in one jenkins upgrade job. As before upgrade test you can create some pod and after upgrade you can check if it survived the upgrade and so on.

I guess that @RazTamir will want ppl write such tests later on.

vasukulkarni: And later its better to add some cli to fit our usecase ex:  `--ocs-version some_GA_version -m upgrade`(to upgrade to latest nightly)

docs/upgrade.md
petr-balogh: Once we will have 4.3. 4.3 will be master For this repo and for 4.2 we will have separate branch for sure.

docs/upgrade.md
vasukulkarni: @petr-balogh I am not sure I got all the point, with cli option, its easier to do whatever upgrade one wants to run and whatever sequence they prefer. We can definitely discuss on this one before just following what you did in RHV.

RazTamir: Both of you correct.
As we still don't have an actual (and official) way to test upgrades, it might be too early to introduce those new marks.  But it will come fast so it may be worth doing it in this PR.

One of the most important items for upgrade testing is to test that the existing cluster was not impacted by the upgrade process.
So we can have a small portion of tests dedicated to upgrading so we will be able to tell if something bad happened after the upgrade

For those tests we will need to ensure:
- Every upgrade test is combined from 2 tests - before upgrade and after upgrade
   * Before upgrade - create some entities with specific names (maybe IDs?)
   * After upgrade - use those entities by their names (maybe IDs?) and verify all good post upgrade
- Environment check is disabled
- NOT to teardown after the before upgrade test finished
- Teardown the after upgrade test


Every suggestion is welcomed of course as this approach might fit good for RHV like products but in K8s style, we might need to test more aspects

docs/upgrade.md
vasukulkarni: Are you talking about ocs or ocs-ci here?

docs/upgrade.md
vasukulkarni: Here is my input using cli and which will achieve the above use case without any additional marks:

Approach1: Using 2 pipelines:
pipeline1:
`run-ci -m deployment,some_upgrade_tests -ocs-version 4.x_GA`

pipeline2:
`run-ci -m upgrade,verify_upgrade_tests --ocs-version latest_nightly`

Approach 2(Just one pipeline):
 The test cases knows all about upgrades and does the job - better approach
Add upgrade tests in folder `/upgrade`
`run-ci -m upgrade, --ocs-version 4.2 --ocs-upgrade-version --nightly tests/upgrade`

docs/upgrade.md
petr-balogh: yes, as the product will evolve, we will have changes in deployment, will have new tests and so on which will be written for latest version which means that to keep everything aligned with version we will have master branch of ocs-ci for the latest supported version we have (for example 4.3 once we will have). For previous versions we will have separate branch and if some change will be needed to be backported to 4.2 for example we will cherry-pick those specific changes but development of new tests will be done just for latest version.

Or how else you would like to solve all those changes between versions?

docs/upgrade.md
petr-balogh: Let's discuss this on stuff/automation mtg, we currently have already the way how we specify the build which is deployed. It's via config [ocs_registry_image](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L49) . 

Currently upgrades cannot be performed to the latest nightly from some random previous version but it hast to be just one follow up build as described in the [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1767400)

So now the upgrade is done just to one build above the current. If it will be supported now we are able to do just with changing one parameter [here](https://github.com/red-hat-storage/ocs-ci/pull/960/files#diff-3c73c8d9493c132ee1cabd09de37c01aR8) 

Anyway we still have just one version 4.2 and upgrading between internal builds and if user want just upgrade they can easily run just `run-ci -m upgrade` and that's it and it will upgrade the cluster.

I think that we are going to use branches for ocs-ci when we will have 4.3, this we also have to discuss which also changing the approach and you won't be able to perform upgrade just in one execution if you need to checkout to different branch between two stages 1) before upgrade,  2) upgrade and after upgrade tests.

Will be better if we will discuss all those things in the call.

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -0,0 +1,69 @@
...
+        else:
+            log.info(f"CSV now upgraded to: {csv_name_after_upgrade}")
+            break
+
+    csv = CSV(
+        resource_name=csv_name_after_upgrade, kind="csv",
+        namespace=namespace
+    )
+    log.info(
+        f"Wating for CSV {csv_name_after_upgrade} to be in succeeded state"
```
vavuthu: Nit: s/Wating/Waiting/

docs/upgrade.md
clacroix12: So we have a couple issues here regarding versioning of the product (using 4.2 and 4.3 as examples):
- Test cases that do not apply to both 4.2 and 4.3
- Fixes that will need to be back ported to 4.2 branch
- Deployment logic that differs from 4.2 to 4.3

These problems can be solved using the following (without a separate branch):
- Tags (or module?) for tests that do not apply to the old version, 4.2 in our example. Or we can use the inverse and have a tag that denotes tests that _can_ run on the old version. We can do whatever is easiest.
- Conditionals in our deployment code to change between 4.2 and 4.3, hook into config to determine which we are deploying. Difficulty depends on how much changes between versions.
- Fixes do not need to be back ported if we remain on one branch.

In my opinion, I think we should stay clear of creating a separate branch if at all possible. I think maintaining another branch is more trouble than it will ultimately be worth. However, this entirely depends on how much goes into the deployment code to ensure it works for both versions, how many test cases will not be applicable to the old version, and how many fixes we plan to back port into the 'old' branch.

vasukulkarni: We should definitely discuss in the meeting before going forward with assumption that we need 2 branches to test upgrades, The *right* thing is to discuss what the upgrade usecases are and why *branching* is required to *achieve* that.  Branching is a costly operation and we cannot put a put a strain on users to maintain multiple branches.

vasukulkarni: As Coady suggested there are multiple ways to achieve upgrade logic in single branch, Infact just folder separate eg: upgrade_4.2, upgrade_4.3, can also be added to above bullet points for specific tests taht needs separation.

vasukulkarni:
High level points on discussion on this one.

a) Continue with single branch approach for 4.3 and use marks for 4.3 specific test.
b) before_upgrade and after_upgrade marks applies only when running in stages (discreet runs)
c) still unclear on single upgrade session run (  ex: upgrade scenario that kicks off specific version, continues running test and initiates upgrade on its own using function call)

docs/upgrade.md
vasukulkarni: I am resolving this for now because still unclear at this stage with only one version.

docs/upgrade.md
vasukulkarni: resolved in meeting.

conf/ocsci/upgrade.yaml
```diff
@@ -0,0 +1,8 @@
+---
+# This config file is for the upgrade purpose
+DEPLOYMENT:
+  upgrade: true
+  # In the case you need to upgrade to the latest build, the following
+  # parameter has to be set. Current builds does not support that yet, see BZ:
+  # https://bugzilla.redhat.com/show_bug.cgi?id=1767400
```
dahorak: The bug is already VERIFIED, is the second sentence still valid?

docs/upgrade.md
```diff
@@ -0,0 +1,58 @@
+# Upgrade
+
+Currently the upgrade can be performed just between follow up downstream builds
+e.g. 171 -> 172.
+There is opened [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1767400) which
+is tracking the issue which prevents us to upgrade to the latest available build.
```
dahorak: Same as above, since the bug is already VERIFIED, is this paragraph still valid?

docs/upgrade.md
```diff
@@ -0,0 +1,58 @@
...
+    -m 'before_upgrade or upgrade or after_upgrade'
+ ```
+
+## TODO
+
+Once there will be 4.3 build and we will have different branch for OCS-CI we
+will need to run upgrade in two steps. First run just with `-m before_upgrade`
+from 4.2 branch if we will have such tests and then with
+`-m 'upgrade or after_upgrade'` from 4.3 (master branch) to run everything from
+proper branches designed for version of OCS.
```
dahorak: This TODO section needs some update based on our yesterday discussion about not using branches right now.

vasukulkarni: please update the doc to reflect the changes.

nehaberry: @petr-balogh as already discussed offline, we would also need to fulfil all the Entry and exit criteria for each upgrade scenario. Once that is dome, only then can we run the tier1 test cases on the cluster.

We would be coming up with a plan on how to combine these validations with the actual upgrade process(as covered in this PR)

petr-balogh: I am currently depending on other changes in the deployment like this PR: https://github.com/red-hat-storage/ocs-ci/pull/983  and I am working on some changes with local storage operator validation but cause of this bug https://bugzilla.redhat.com/show_bug.cgi?id=1770183 I am unable to check the version of this operator so I will have to workaround this. Also there are issues with deployment since last week so it's problematic with verification now. Once I have changes verified locally I will push to this PR!

@nehaberry the tier1 execution on upgraded environment will be done in second execution of ocs-ci on the environment after the upgrade.

docs/upgrade.md
petr-balogh: Yep it's resolved.

petr-balogh:
I ran verification here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/2143/console . 

I will need to improve the verification of the upgrade but this will happen in the follow up PR.
@vasukulkarni can you please approve, I will ask @dahorak to re-approve tomorrow and will merge if you are OK with that.

Thanks

conf/ocsci/upgrade_to_next_build.yaml
```diff
@@ -0,0 +1,7 @@
+---
+# This config file is for the upgrade to the next one DS build follownig the
+# current installed build. Mainly used for testing upgrades, so you can install
+# som older version and then ran upgrade one by one version.
```
dahorak:
```suggestion
# some older version and then run upgrade one by one version.
```

ocs_ci/utility/utils.py
```diff
@@ -1230,9 +1230,10 @@ def clone_repo(url, location, branch='master', to_checkout=None):
         run_cmd(f"git checkout {to_checkout}", cwd=location)
 
 
-def get_latest_ds_olm_tag():
+def get_latest_ds_olm_tag(upgrade=False):
```
vavuthu: Can we add Args

ocs_ci/utility/utils.py
```diff
@@ -1241,11 +1242,54 @@ def get_latest_ds_olm_tag():
...
+        if not upgrade:
+            if tag['name'] != 'latest' and tag['image_id'] == latest_image:
+                return tag['name']
+        if upgrade:
+            if tag['name'] != 'latest' and tag['image_id'] != latest_image:
+                return tag['name']
+    raise TagNotFoundException(f"Couldn't find any desired tag!")
+
+
+def get_next_version_available_for_upgrade(current_tag):
```
vavuthu: Can we add Args

ocs_ci/ocs/resources/ocs.py
```diff
@@ -144,3 +149,170 @@ def add_label(self, label):
 
     def delete_temp_yaml_file(self):
         utils.delete_file(self.temp_yaml.name)
+
+
+def ocs_install_verification(timeout=0):
+    """
+    Perform steps necessary to verify a successful OCS installation
+    """
```
dahorak: Missing `Args:` section with description of `timeout` parameter.

ocs_ci/utility/utils.py
```diff
@@ -1230,9 +1230,10 @@ def clone_repo(url, location, branch='master', to_checkout=None):
...
 
-def get_latest_ds_olm_tag():
+def get_latest_ds_olm_tag(upgrade=False):
     """
-    This function returns latest tag of OCS downstream registry
+    This function returns latest tag of OCS downstream registry or one before
+    latest if upgrade parameter is True
 
     Returns:
         str: latest tag for downstream image from quay registry
```
dahorak: Missing `Args:` section with description of `upgrade` parameter.

ocs_ci/utility/utils.py
```diff
@@ -1241,11 +1242,54 @@ def get_latest_ds_olm_tag():
...
+                return tag['name']
+    raise TagNotFoundException(f"Couldn't find any desired tag!")
+
+
+def get_next_version_available_for_upgrade(current_tag):
+    """
+    This function returns the tag built after the current_version
+
+    Returns:
+        str: tag for downstream image from quay registry built after
```
dahorak: Missing `Args:` section with description of `current_tag` parameter.

PR #964, opened 2019-11-04
Jenkins: Use acceptance mark; drop old params

zmc:


Jenkinsfile
```diff
@@ -112,26 +103,17 @@ pipeline {
     success {
       script {
         if( env.UMB_MESSAGE in [true, 'true'] ) {
-          def operator_image = "${env.OCS_OPERATOR_IMAGE}"
+          def registry_image = "${env.OCS_REGISTRY_IMAGE}"
           // quay.io/rhceph-dev/ocs-operator:4.2-58.e59ca0f.master -> 4.2
           def operator_version = operator_image.split(':')[-1].split('-')[0]
```
petr-balogh: Where is the operator_image defined? I see it's still used on line 108:
```python
def operator_version = operator_image.split(':')[-1].split('-')[0]
```
Shouldn' you change it also there? And the comment below refer to ocs-operator image and not registry image so maybe also change this if we will use registry image instead.

zmc: Whoops, thanks for the catch!

PR #977, opened 2019-11-07
Try to retrieve the MCG region from the classic addresses

MeridianExplorer:
NooBaa has changed the product behavior to include a new address scheme that doesn't include the region in its name.
This PR aims to retrieve and use the old address regardless of the NooBaa version.
Signed-off-by: Ben <belimele@redhat. com>


MeridianExplorer:
I'd like to replace the entire region detection algo to something more correct and robust.
Closed.

PR #984, opened 2019-11-11
Adding yaml for AWS UPI production environment

vavuthu:

### DIFF
--- a/conf/ocsci/production-aws-upi.yaml
+++ b/conf/ocsci/production-aws-upi.yaml
@@ -1,0 +1,7 @@
+---
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
+ENV_DATA:
+  platform: 'aws'
+  deployment_type: 'upi'
+  worker_instance_type: 'm5.4xlarge'


vavuthu:
> I believe we will need `rhcos_ami` under `ENV_DATA` since this isn't set in the default configuration.



> I believe we will need `rhcos_ami` under `ENV_DATA` since this isn't set in the default configuration.



> I believe we will need `rhcos_ami` under `ENV_DATA` since this isn't set in the default configuration.

For AWS UPI, user has to pass the conf/ocsci/aws_upi.yaml ( which includes rhcos_ami ) to run-ci and if someone needs the production deployment, we need pass conf/ocsci/production-aws-upi.yaml along with conf/ocsci/aws_upi.yaml so that we are keeping rhcos_ami image in a single place

petr-balogh:
> > I believe we will need `rhcos_ami` under `ENV_DATA` since this isn't set in the default configuration.
> 
> For AWS UPI, user has to pass the conf/ocsci/aws_upi.yaml ( which includes rhcos_ami ) to run-ci and if someone needs the production deployment, we need pass conf/ocsci/production-aws-upi.yaml along with conf/ocsci/aws_upi.yaml so that we are keeping rhcos_ami image in a single place

Since we don't have any changes in we could have just one production config for AWS, but let's have it separated as it's now as we might need to add some changes later on.

PR #989, opened 2019-11-12
Add 1az Config for default region

prsurve:

### DIFF
--- a/conf/ocsci/aws_ipi_1az.yaml
+++ b/conf/ocsci/aws_ipi_1az.yaml
@@ -1,0 +1,7 @@
+---
+ENV_DATA:
+  platform: 'aws'
+  worker_availability_zones:
+    - 'us-east-2a'
+  master_availability_zones:
+    - 'us-east-2a'
--- a/conf/ocsci/aws_ipi_1az.yaml
+++ b/conf/ocsci/aws_ipi_1az.yaml
@@ -1,7 +1,0 @@
----
-ENV_DATA:
-  platform: 'aws'
-  worker_availability_zones:
-    - 'us-east-2a'
-  master_availability_zones:
-    - 'us-east-2a'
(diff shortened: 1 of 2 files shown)


conf/ocsci/aws_ipi_1az.yaml
```diff
@@ -0,0 +1,7 @@
+---
```
petr-balogh: I know it's default value for our deployment but when you are explicit in the name of the IPI deployment type. What about to add it also here:
https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/production-aws-ipi.yaml#L6

Or second option to not base it on deployment type, rename it to aws_1az.yaml and for UPI will will need to align to use the same if possible. I prefer second option.

prsurve: will check with UPI deployment as well and update file name

petr-balogh: BTW if the UPI deployment won't support it, we will need to open issue for it and you can just put comment to this config file at the start of this file that UPI is not supported and link to the issue.

petr-balogh: Also please put some comment at the start of the config file like:
`# This config is for setup cluster in 1 Availability Zone for AWS`

Thanks

prsurve: - Addressed Comments
- Updated yaml with issue url

prsurve: Done

vasukulkarni: if this is default for aws-ipi, why not just add it to existing aws-ipi file?

prsurve: @vasukulkarni this config can be also used for aws-upi so removed aws-ipi file, currently aws-upi installation with this config is not working so i created an issue https://github.com/red-hat-storage/ocs-ci/issues/991

Issue #992, opened 2019-11-13
ERROR - Wait for Pod resource  to reach desired condition Running failed, last actual status was ['Running']


wusui:
I got this message (which does not make sense (Running is the desired condition and current state -- why fail?))

This was when I tried deploying the latest ocs-ci (e4c1ca5164b9adc85933a81f79e51b3f9a85c044)

19:08:11 - MainThread - ocs_ci.ocs.ocp - ERROR - Wait for Pod resource  to reach desired condition Running failed, last actual status was ['Running']
FAILED                                                                   [100%]

=================================== FAILURES ===================================
_______________________________ test_deployment ________________________________

    @destroy
    @deployment
    def test_deployment():
        deploy = config.RUN['cli_params'].get('deploy')
        teardown = config.RUN['cli_params'].get('teardown')
        if not teardown or deploy:
            log.info("Verifying OCP cluster is running")
            assert is_cluster_running(config.ENV_DATA['cluster_path'])
            if not config.ENV_DATA['skip_ocs_deployment']:
>               ocs_install_verification()

tests/ecosystem/deployment/test_deployment.py:22: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/ecosystem/deployment/test_deployment.py:78: in ocs_install_verification
    assert pod.wait_for_resource(
ocs_ci/ocs/ocp.py:435: in wait_for_resource
    raise(ex)
ocs_ci/ocs/ocp.py:385: in wait_for_resource
    timeout, sleep, self.get, resource_name, True, selector
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ocs_ci.utility.utils.TimeoutSampler object at 0x7f48ae9384a8>

    def __iter__(self):
        if self.start_time is None:
            self.start_time = time.time()
        while True:
            self.last_sample_time = time.time()
            try:
                yield self.func(*self.func_args, **self.func_kwargs)
            except Exception:
                pass
    
            if self.timeout < (time.time() - self.start_time):
>               raise self.timeout_exc_cls(*self.timeout_exc_args)
E               ocs_ci.ocs.exceptions.TimeoutExpiredError: Timed Out: (0,)

[...]

PR #993, opened 2019-11-13
Fix noobaa pod checking after deployment

vasukulkarni:
Fixes: https://github.com/red-hat-storage/ocs-ci/issues/992

there is only one noobaa pod after deployment, fix the check for verification

### DIFF
--- a/tests/ecosystem/deployment/test_deployment.py
+++ b/tests/ecosystem/deployment/test_deployment.py
@@ -78,7 +78,7 @@ def ocs_install_verification():
     assert pod.wait_for_resource(
         condition=constants.STATUS_RUNNING,
         selector=constants.NOOBAA_APP_LABEL,
-        resource_count=2,
+        resource_count=1,
         timeout=timeout
     )
     # local-storage-operator


nehaberry:
@vasukulkarni is the change in count of nooba pod expected ? Or is it a Bug ?

petr-balogh:
> @vasukulkarni is the change in count of nooba pod expected ? Or is it a Bug ?

I see we have bug for it here:  https://bugzilla.redhat.com/show_bug.cgi?id=1771844 . so do not think that we should merge this change as we shouldn't miss this noobaa core pod I guess.

jilju:
> > @vasukulkarni is the change in count of nooba pod expected ? Or is it a Bug ?
> 
> I see we have bug for it here: https://bugzilla.redhat.com/show_bug.cgi?id=1771844 . so do not think that we should merge this change as we shouldn't miss this noobaa core pod I guess.

Shall we consider this as a workaround to unblock deployment.

vasukulkarni:
Closing, Invalid.

PR #995, opened 2019-11-13
Implement Multi-region MCG test

MeridianExplorer:


ocs_ci/ocs/resources/mcg.py
```diff
@@ -272,3 +296,289 @@ def _retrieve_reduction_data():
...
+
+        def _retrieve_credreq_provision_state():
+            return run_cmd(
+                f"oc get credentialsrequests {creds_request.name} -o jsonpath={{.status.provisioned}}"
+            )
+
+        try:
+            for provisioned_state in TimeoutSampler(30, 5, _retrieve_credreq_provision_state):
+                if provisioned_state == "true":
+                    sleep(10)
```
ebenahar: Not sure how TimeoutSampler behaves if you're waiting here for 10 seconds while its polling rate is 5 seconds.

MeridianExplorer: If I reach provisioned_state, I sleep for 10 seconds and then break - meaning it has no effect over `TimeoutSampler`. This sleep is needed because there is absolutely no way to check whether the IAM was actually provisioned except trying to use it (which causes an exception). This is just a safety measure that allows it to truly finish setting up. After the sleep, no further samplers are used.

ocs_ci/ocs/resources/mcg.py
```diff
@@ -272,3 +296,289 @@ def _retrieve_reduction_data():
...
+            "name": conn_name,
+            "secret": self.aws_access_key
+        }
+
+        try:
+            for resp in TimeoutSampler(
+                30, 3, self.send_rpc_query, 'account_api', 'add_external_connection', params
+            ):
+                if 'error' not in resp.text:
+                    logger.info(f'Connection {conn_name} created successfully')
```
ebenahar: this function returns False only. Please adjust it to return True instead of breaking

ocs_ci/ocs/resources/mcg.py
```diff
@@ -272,3 +296,289 @@ def _retrieve_reduction_data():
...
+            logger.error(f'Could not create connection {conn_name}')
+            return False
+
+    def create_new_backingstore_bucket(self, backingstore_info):
+        """
+        Creates an S3 target bucket for NooBaa to use as a backing store
+
+        Args:
+            backingstore_info: A tuple containing the BS information
+                               to be used in its creation.
```
ebenahar: too many indentations

ocs_ci/ocs/resources/mcg.py
```diff
@@ -272,3 +296,289 @@ def _retrieve_reduction_data():
...
+        bc_data['metadata']['name'] = name
+        bc_data['metadata']['namespace'] = self.namespace
+        tiers = bc_data['spec']['placementPolicy']['tiers'][0]
+        tiers['backingStores'] = backingstores
+        tiers['placement'] = placement
+        return create_resource(**bc_data)
+
+    def toggle_bucket_readwrite(self, bucketname, block=True):
+        """
+
```
ebenahar: Please add a description here

ocs_ci/ocs/resources/mcg.py
ebenahar: ok. Thanks for the explanation. 
Do we expect more than 1 iteration of TimeoutSampler?

ocs_ci/ocs/resources/mcg.py
```diff
@@ -272,3 +296,289 @@ def _retrieve_reduction_data():
...
+            for written_object in obj_list:
+                object_chunks = self.send_rpc_query('object_api', 'read_object_mapping', params={
+                    'bucket': bucket_name,
+                    'key': written_object.get('key'),
+                    'obj_id': written_object.get('obj_id')
+                }).json().get('reply').get('chunks')
+
+                for object_chunk in object_chunks:
+                    mirror_blocks = object_chunk.get('frags')[0].get('blocks')
+                    mirror_nodes = [mirror_blocks[i].get('block_md').get('node') for i in range(len(mirror_blocks))]
```
ebenahar: line too long

tests/manage/mcg/test_multi_region.py
```diff
@@ -0,0 +1,141 @@
...
+
+            for aws_bucket in aws_buckets:
+                mcg_obj.aws_s3_resource.Bucket(aws_bucket).objects.all().delete()
+                mcg_obj.aws_s3_resource.Bucket(aws_bucket).delete()
+
+        request.addfinalizer(resource_cleanup)
+
+        return aws_buckets, bs_secrets, bs_objs, bucketclasses
+
+    @pytest.mark.polarion_id("OCS-1298")
```
ebenahar: this polarion id is already used in tests/manage/mcg/test_bucket_creation.py:20

ocs_ci/ocs/resources/mcg.py
MeridianExplorer: Theoretically - it's possible.
Practically - I never ran into a case in which it happened.

ebenahar: So maybe the TimeoutSampleris redundant here?

MeridianExplorer: I think it'd be better to go on the safer side of things - if it won't be used - then no harm is done. If it'll be used - removing it will cause an exception.

It's mostly here because I saw that the NooBaa sanity test relied upon it

ocs_ci/ocs/resources/mcg.py
```diff
@@ -73,20 +77,33 @@ def __init__(self):
...
+        )
+
     def s3_list_all_bucket_names(self):
         """
         Returns:
             list: A list of all bucket names
 
         """
-        return [bucket.name for bucket in self.s3_resource.buckets.all()]
+        return {bucket.name for bucket in self.s3_resource.buckets.all()}
```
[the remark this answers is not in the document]
MeridianExplorer: Forgot to refactor the docs and names - that's a mistake on my side.
I wanted to update it to a set to keep the performance as good as possible. Bucket names are, by definition, unique across the entire service. Changing this from a list to a set allows us to have an access efficiency of O(1) instead of O(n).

tests/manage/mcg/helpers.py
```diff
@@ -0,0 +1,35 @@
+import boto3
+
+from ocs_ci.ocs import constants
+from tests.helpers import logger, craft_s3_command
+
+
+def retrieve_test_objects_to_pod(podobj, target_dir):
+    # Download test objects from the public bucket
```
ebenahar: Please add docstring

tests/manage/mcg/helpers.py
```diff
@@ -0,0 +1,35 @@
...
+        podobj.exec_cmd_on_pod(
+            command=f'sh -c "'
+                    f'wget -P {target_dir} '
+                    f'https://{constants.TEST_FILES_BUCKET}.s3.amazonaws.com/{obj.key}"'
+        )
+        downloaded_objects.append(obj.key)
+    return downloaded_objects
+
+
+def sync_object_directory(podobj, src, target, mcg_obj=None):
```
ebenahar: Pease add docstring

tests/manage/mcg/test_multi_region.py
```diff
@@ -0,0 +1,129 @@
...
+                mirror_tier_name = bucket.get('tiering').get('tiers')[0].get('tier')
+                break
+
+        # Retrieved the pools attached to the tier
+        for tier in system_state.get('tiers'):
+            if tier.get('name') == mirror_tier_name:
+                mirror_attached_pools = tier.get('attached_pools')
+                break
+
+        assert len(mirror_attached_pools) == 2
```
ebenahar: please add an assertion massage

tests/manage/mcg/test_multi_region.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @tier2
+    @pytest.mark.polarion_id("OCS-1784")
+    def test_multiregion_multiregion_mirror(self, mcg_obj, awscli_pod, multiregion_mirror_setup):
+        """
+        Test bucket creation using the S3 SDK
```
ebenahar: change description to:
Test multi-region bucket creation using the S3 SDK

tests/manage/mcg/test_multi_region.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # Bring bucket A down
+        mcg_obj.toggle_bucket_readwrite(backingstore1['name'])
+        mcg_obj.check_backingstore_state('backing-store-' + backingstore1['name'], BS_AUTH_FAILED)
+
+        # Verify integrity of B
+        # Retrieve all objects from MCG bucket to result dir in Pod
+        sync_object_directory(awscli_pod, mcg_bucket_path, local_temp_path, mcg_obj)
+
+        # Checksum is compared between original and result object
```
ebenahar: Really nice test logic!!

ocs_ci/ocs/resources/mcg.py
MeridianExplorer: Took out that TimeoutSampelr - resolved :)

tests/manage/mcg/test_multi_region.py
MeridianExplorer: Fixed :)

ocs_ci/ocs/resources/mcg.py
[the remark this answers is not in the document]
waynesun09: I didn't find the updated functions been used with list get, so it did not break anything, so update to set is fine.

tests/manage/mcg/test_multi_region.py
MeridianExplorer: Thank a lot :)

PR #996, opened 2019-11-13
Increase timeout for CSV and PackageManifest

petr-balogh:
Fixes: #782

The BZ appeared again:
https://bugzilla.redhat.com/show_bug.cgi?id=1757013

So trying to increase timeouts to wait a bit more and maybe it will
stabilize the deployment.


petr-balogh:
The verification job passed the stage which I am changing here but failed on other known noobaa missing pod issue. Consider it as verified.

PR #999, opened 2019-11-13
OCS-278 Test Case Implementation

hadarsharon:
Implementation of OCS-278 Test Case


tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,171 @@
...
+        multi_pvc_factory(interface=constants.CEPHBLOCKPOOL,
+                          project=project_1,
+                          storageclass=self.rbd_sc,
+                          access_mode=constants.ACCESS_MODE_RWO,
+                          num_of_pvc=pvcs_num)
+
+        # Delete the entire Project 1 (along with all of its PVCs)
+        pvs = helpers.get_all_pvs()
+        used_before_deletion = self.ceph_cluster.check_ceph_pool_used_space(
+            self.ceph_cluster.cluster_name)
```
jilju:
```suggestion
            self.ceph_cluster.cluster_name
    )
```

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,171 @@
...
+        project_deletion_start_time = time.time()
+        ocp.switch_to_default_rook_cluster_project()
+        log.info("Deleting Project 1")
+        project_1.delete(resource_name=project_1.namespace)
+        log.info("Waiting for Project 1 deletion success...")
+        project_1.wait_for_delete(resource_name=project_1.namespace)
+        log.info("Project 1 deleted successfully.")
+        project_deletion_time = time.time() - project_deletion_start_time
+        log.info(
+            "Project 1 deletion time: {} seconds".format(project_deletion_time))
```
jilju:
```suggestion
            "Project 1 deletion time: {} seconds".format(project_deletion_time)
    )
```

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,171 @@
...
+        for pv in pvs:
+            pv_start_time = time.time()
+            pv_name = pv.backed_pv
+            pv.ocp.wait_for_delete(pv_name)
+            helpers.validate_pv_delete(pv_name)
+            pv_delete_time = time.time() - pv_start_time
+            individual_pv_delete_times.append({pv_name: pv_delete_time})
+            time_between_pvcs_deletion_success += pv_delete_time
+        log.info("{} seconds between first and last PVC deletion".format(
+            time_between_pvcs_deletion_success))
```
jilju:
```
log.info(
            "{} seconds between first and last PVC deletion".format(
            time_between_pvcs_deletion_success
            )
)
```

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,171 @@
...
+        # Verify space has been reclaimed
+        used_after_deletion = self.ceph_cluster.check_ceph_pool_used_space(
+            self.ceph_cluster.cluster_name)
```
jilju:
```suggestion
            self.ceph_cluster.cluster_name
    )
```

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,171 @@
...
+        # Check time between first PV delete and last PV delete
+        time_between_pvcs_deletion_success = 0
+        # Also check delete time per each PV
+        individual_pv_delete_times = []
+        for pv in pvs:
+            pv_start_time = time.time()
+            pv_name = pv.backed_pv
+            pv.ocp.wait_for_delete(pv_name)
+            helpers.validate_pv_delete(pv_name)
+            pv_delete_time = time.time() - pv_start_time
```
jilju: The order of actual pv deletion need not be in the same order of pv in the list 'pvs'. So this value will not be accurate.
eg: The 10th one in this pvs list might be the last one in actual queue.

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,171 @@
...
+            individual_pv_delete_times.append({pv_name: pv_delete_time})
```
jilju: This info is not used anywhere.

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,171 @@
...
+            time_between_pvcs_deletion_success += pv_delete_time
```
jilju: time_between_pvcs_deletion_success += pv_delete_time

This value will be appropriate if it is the time difference  before and after this loop.

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,171 @@
...
+        log.info("{} seconds between first and last PVC deletion".format(
+            time_between_pvcs_deletion_success))
+        # Verify space has been reclaimed
+        used_after_deletion = self.ceph_cluster.check_ceph_pool_used_space(
+            self.ceph_cluster.cluster_name)
```
jilju: Fix parenthesis in all places

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,171 @@
...
+            f"/nUsed space before deletion: " f"{used_before_deletion}"
+            f"/nUsed space after deletion: {used_after_deletion}"
+        )
+        log.info("Space reclaimed successfully.")
+
+        # Project 1 done, moving on to Project 2
+        log.info("Creating Project 2")
+        project_2 = project_factory()
+
+        log.info("Creating a mix of {} CephFS & RBD PVCs "
```
jilju: use f-sting instead of str.format() in all places.

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,171 @@
...
+@tier1
+@pytest.mark.polarion_id("OCS-278")
+@pytest.mark.parametrize(
+    argnames=["pvcs_num"],
+    argvalues=[
+        pytest.param(
+            100, marks=pytest.mark.tier1
+        ),
+        pytest.param(
+            1000, marks=pytest.mark.scale
```
jilju: With this marker, this test will be picked for tier1 and scale. Should these tests be separate ?

hadarsharon: @ebenahar your take on that?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: Applying with f-string as well...

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: @jilju Which approach would you suggest in that case? Randomization...?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: @jilju So basically move it outside the loop?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
hadarsharon: True - I forgot to implement the "log deletion time per each PV" part... I don't see any reason to use this list anymore, as I can just log each individual deletion time from within the loop - and so I will implement it this way.

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
jilju: Checking for random pv from the list will not help. I am not getting any way to find the exact time taken for a pv deletion, without checking logs.

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
```diff
@@ -0,0 +1,175 @@
...
+                        storageclass=self.cephfs_sc, access_mode=next(rwo_rwx))
+        log.info(f"Creating {pvcs_num} RBD PVCs")
+        multi_pvc_factory(interface=constants.CEPHBLOCKPOOL,
+                          project=project_1,
+                          storageclass=self.rbd_sc,
+                          access_mode=constants.ACCESS_MODE_RWO,
+                          num_of_pvc=pvcs_num)
+
+        # Delete the entire Project 1 (along with all of its PVCs)
+        pvs = helpers.get_all_pvs()
```
jilju: Isn't this giving the list of all PVs in the cluster ?

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
ebenahar: this looks fine. 1000 should be tested as part of scale suite and 100 as part of tier1. We are good

jilju: Fine with that. Then this tier1 annotation should be removed from line 19. Otherwise, 1000 will be selected for both scale and tier1

hadarsharon: Understandable

ebenahar:
@hadarsharon please verify

tests/manage/pv_services/test_delete_project_with_pvc_and_pods.py
RazTamir: @hadarsharon this seems like risky

ebenahar:
Closing old PRs. Please re-open if needed

Issue #1001, opened 2019-11-14
Update production values for vSphere


vavuthu:
Update the conf/ocsci/production-vsphere-upi.yaml with actual production values.

And also due to hardware limitation in existing lab, provide user to use maximum values and allow allow_lower_instance_requirements

PR #1002, opened 2019-11-14
Adding Production values to vSphere platform

vavuthu:
Adding Production values to vSphere platform

Also, added separate production yaml file for ocs4qe lab which has
hardware limitations.

Fixes: #1001


conf/ocsci/production-vsphere-upi-ocs4qe-lab.yaml
```diff
@@ -0,0 +1,8 @@
+---
+# This config will provide the 12 vCPU's to VM.
+# We are restricting to 12 due to hardware limitation in the lab
+DEPLOYMENT:
+  allow_lower_instance_requirements: true
+ENV_DATA:
+  worker_num_cpus: '12'
+  master_num_cpus: '4'
```
petr-balogh: Do we have enough memory to increase on our lab env or not really?

petr-balogh:
@clacroix12 FYI for our lab we will have to use lab production config as we have lack of resources in our lab

conf/ocsci/production-vsphere-upi-ocs4qe-lab.yaml
vavuthu: We cannot accommodate 64GB

conf/ocsci/production-vsphere-upi-ocs4qe-lab.yaml
```diff
@@ -0,0 +1,8 @@
+---
```
clacroix12: If these are not true production values should we remove `production-` from the filename?

vavuthu: Done

PR #1009, opened 2019-11-14
MCG region fix

parikshithb:
Quick fix to invalid mcg region issue which was supposed to be fixed in #995. 
Noobaa had changed address scheme which resulted in fetching invalid region name.


vasukulkarni:
Please also file an github issue for a proper fix.

PR #1010, opened 2019-11-14
Add more bucket and OBC deletion cases

waynesun09:
Add test case for delete 100 and 1000 buckets and OBCs, delete                                                                                                                                                                            
non-empty bucket and OBC, delete non-existant bucket and OBC.                                                                                                                                                                             
                                                                                                                                                                                                                                              
Signed-off-by: Wayne Sun <gsun@redhat.com>

### DIFF
--- a/tests/manage/mcg/test_bucket_deletion.py
+++ b/tests/manage/mcg/test_bucket_deletion.py
@@ -50,16 +50,6 @@ class TestBucketDeletion:
                 marks=[tier2, pytest.mark.polarion_id("OCS-1866")]
             ),
             pytest.param(
-                *[100, 'CLI'],
-                marks=[tier2, noobaa_cli_required,
-                       pytest.mark.polarion_id("OCS-1865")]
-            ),
-            pytest.param(
-                *[1000, 'CLI'],
-                marks=[tier2, noobaa_cli_required,
-                       pytest.mark.polarion_id("OCS-1866")]
-            ),
-            pytest.param(
                 *[100, 'OC'],
                 marks=[tier2, pytest.mark.polarion_id("OCS-1865")]
             ),
--- a/tests/manage/mcg/test_bucket_deletion.py
+++ b/tests/manage/mcg/test_bucket_deletion.py
@@ -1,11 +1,19 @@
 import logging
 
+import boto3
+import botocore
 import pytest
 
 from ocs_ci.framework.pytest_customization.marks import (
     tier1, noobaa_cli_required, acceptance,
     filter_insecure_request_warning
 )
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.exceptions import CommandFailed
+from ocs_ci.ocs.ocp import OCP
+from ocs_ci.ocs.resources.mcg_bucket import S3Bucket, OCBucket, CLIBucket
+from ocs_ci.utility.utils import run_mcg_cmd
+from tests.helpers import craft_s3_command, create_unique_resource_name
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


tests/manage/mcg/test_bucket_deletion.py
```diff
@@ -1,51 +1,202 @@
...
-    @pytest.mark.polarion_id("OCS-1299")
-    def test_s3_bucket_delete(self, mcg_obj, bucket_factory):
+    @pytest.mark.parametrize(
+        argnames="amount,interface",
+        argvalues=[
+            pytest.param(
+                *[3, 'S3'],
+                marks=[pytest.mark.polarion_id("OCS-1299"), tier1, acceptance]
+            ),
+            pytest.param(
```
ebenahar: I think it is redundant to execute all sizes over all the different APIs. Let's focus on OC for all sizes and for s3 and CLI, test only the functional (amount=3)

MeridianExplorer: CLI + OC might be a bit redundant, but S3 is very important IMO. It's an entirely different control plane, and one that'll most definitely be used by client.
I suggest we do OC+S3 for all sizes, and keep CLI functional-only.

MeridianExplorer:
Looks absolutely great.
I'd just remove the additional sizes that Elad and I talked about, other than that - approved :)

waynesun09:
additional sizes removed

tests/manage/mcg/test_bucket_deletion.py
```diff
@@ -1,51 +1,192 @@
...
+                *[1000, 'S3'],
+                marks=[tier2, pytest.mark.polarion_id("OCS-1866")]
+            ),
+            pytest.param(
+                *[100, 'OC'],
+                marks=[tier2, pytest.mark.polarion_id("OCS-1865")]
+            ),
+            pytest.param(
+                *[1000, 'OC'],
```
RazTamir: should we create a new Polarion ID to distinguish OC and S3?
@ebenahar @hnallurv

tests/manage/mcg/test_bucket_deletion.py
```diff
@@ -1,51 +1,192 @@
...
+                marks=[pytest.mark.polarion_id("OCS-1867"), tier3]
+            ),
+            pytest.param(
+                *['CLI'],
+                marks=[tier1, noobaa_cli_required,
+                       pytest.mark.polarion_id("OCS-1868")]
+            ),
+            pytest.param(
+                *['OC'],
+                marks=[tier1, pytest.mark.polarion_id("OCS-1868")]
```
RazTamir: same

tests/manage/mcg/test_bucket_deletion.py
waynesun09: @RazTamir updated

tests/manage/mcg/test_bucket_deletion.py
waynesun09: updated

ebenahar:
@waynesun09 Please rebase

waynesun09:
Updated add one new case and also use the new helper functions.

tests/manage/mcg/test_bucket_deletion.py
```diff
@@ -1,51 +1,215 @@
...
+        argnames="amount,interface",
+        argvalues=[
+            pytest.param(
+                *[3, 'S3'],
+                marks=[pytest.mark.polarion_id("OCS-1299"), tier1, acceptance]
+            ),
+            pytest.param(
+                *[3, 'CLI'],
+                marks=[tier1, acceptance, noobaa_cli_required,
+                       pytest.mark.polarion_id("OCS-1299")]
```
ebenahar: multiple cases are marked with OCS-1299. Does it reflects the polarion test cases?
The same for 1400

waynesun09: 1299 was the original case here cover both OBCs and s3 buckets, as I parametrize it with add more new cases I left it unchanged.
1400 was a new case I created though, and it's also for both OBCs and s3 buckets.

waynesun09: updated all with 1:1 mapping

MeridianExplorer:
Does this PR include deletion of buckets that contain large datasets? I see that we're testing deletion with objects in `test_bucket_delete_with_objects`, but the dataset seems to be synced only once, which results in a (relatively) very small dataset

tests/manage/mcg/test_bucket_deletion.py
```diff
@@ -1,50 +1,214 @@
...
+            pytest.param(
+                *[100, 'S3'],
+                marks=[tier2, pytest.mark.polarion_id("OCS-1865")]
+            ),
+            pytest.param(
+                *[100, 'OC'],
+                marks=[tier2, pytest.mark.polarion_id("OCS-1915")]
+            ),
+            pytest.param(
+                *[1000, 'S3'],
```
MeridianExplorer: I think it'd be best to update it to conform with #1176 by skipping the test (skipif True) and stating the erratic behavior issue

waynesun09: Updated accordingly

waynesun09:
The large dataset case is:

test_s3_bucket_delete_1t_objects

PR #1012, opened 2019-11-15
Adding production device size as '2TB' for perf and scale testing

keesturam:
For all performance and scale testing, we will need to run the tests with the actual device size planned for OCS. i.e., 2TB. This PR adds a config to create devices with 2 TB. 

p.s., We are not updating production_device_size.yaml to 2 TB as running CI jobs (which are mostly functional tests) with 2TB OSDs would be an overkill.

### DIFF
--- a/conf/ocsci/production_device_size_e2e.yaml
+++ b/conf/ocsci/production_device_size_e2e.yaml
@@ -1,0 +1,5 @@
+---
+#This config file holds the disk size in GB for production.
+#This device size is to be used for all performance and scale testing.
+ENV_DATA:
+  device_size: '2048'


ramkiperiy:
Please resole DCO error.

petr-balogh:
@keesturam isn't time to change rather our whole production config: https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/production_device_size.yaml  to have this size? Cause this config is automatically used in our jenkins jobs since this week so maybe we can directly update our production config @hnallurv @RazTamir thoughts?  Or do we want to keep running our jobs with 1 TB?

vasukulkarni:
@petr-balogh I think we talked about running only performance and scale testing with 2TB and using different config for those, since tier1 and other tests are not really utilizing the full disk size(there aren't tests). So things that use 2TB can be configured in jenkins to use this file instead of changing production.

petr-balogh:
> @petr-balogh I think we talked about running only performance and scale testing with 2TB and using different config for those, since tier1 and other tests are not really utilizing the full disk size(there aren't tests). So things that use 2TB can be configured in jenkins to use this file instead of changing production.

The product will officially support just 2 TB right? My point is that I wanted to avoid that we will have some bug opened and they will tell us that we didn't run against production config which should be 2 TB.  But if no need I am fine with this but we will need to utilize our jobs to take this config for performance and workload jobs.

vasukulkarni:
> > @petr-balogh I think we talked about running only performance and scale testing with 2TB and using different config for those, since tier1 and other tests are not really utilizing the full disk size(there aren't tests). So things that use 2TB can be configured in jenkins to use this file instead of changing production.
> 
> The product will officially support just 2 TB right? My point is that I wanted to avoid that we will have some bug opened and they will tell us that we didn't run against production config which should be 2 TB. But if no need I am fine with this but we will need to utilize our jobs to take this config for performance and workload jobs.

I think that is what we discussed not to have 2TB, if we see issues because of size we could merge it to same production file, Infact I am thinking 100G for tier1 is fine too and Daniel talked about that for jenkins job as well.

dahorak:
Initially, there were request to have even smaller disks for tests utilizing the whole capacity - to get it filled quicker and don't wait hours just for creating/copying the testing data... But of course, if it might affect the tests validity (because of unsupported sizes), we need to use the supported sizes and find other ways how to utilize them quicker...

vasukulkarni:
As is there is no harm with this PR, we can merge one more after our discussion on Wed/Thursday and combine this files.

PR #1013, opened 2019-11-15
Add session scoped factory fixtures and pod IO test for upgrade

fbalak:
I propose to use files `test_resources.py` and `test_configuration.py` for upgrade tests as order is already distinguished by `order_pre_upgrade` and `order_post_upgrade` markers.


tests/ecosystem/upgrade/conftest.py
```diff
@@ -0,0 +1,49 @@
+import logging
+import pytest
+
+from ocs_ci.ocs import constants
+
+log = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope='session')
```
ebenahar: I think that it would be better to create both cephfs and rbd pods in the same fixture. You are doing that in the 'post_upgrade_pods' fixture

fbalak: ok, done

tests/conftest.py
```diff
@@ -53,6 +53,11 @@ def secret_factory_class(request):
     return secret_factory_fixture(request)
 
 
+@pytest.fixture(scope='session')
+def secret_factory_session(request):
+    return secret_factory_fixture(request)
+
```
nehaberry: @fbalak @petr-balogh are we not going to use the default Secret, SC and Blockpool for all IO before and after upgrade ? 
If yes & we agree to that, then do we really need these **session fixtures** , as we would anyways not delete default resources? Only PVC deletions would suffice

tests/conftest.py
```diff
@@ -151,6 +161,19 @@ def storageclass_factory_class(
...
+def storageclass_factory_session(
+    request,
+    ceph_pool_factory_session,
+    secret_factory_session
+):
+    return storageclass_factory_fixture(
+        request,
+        ceph_pool_factory_session,
+        secret_factory_session
+    )
```
nehaberry: same

tests/conftest.py
```diff
@@ -288,6 +316,19 @@ def pvc_factory_class(
...
+@pytest.fixture(scope='session')
+def pvc_factory_session(
+    request,
+    storageclass_factory_session,
+    project_factory_session
+):
+    return pvc_factory_fixture(
+        request,
+        storageclass_factory_session,
+        project_factory_session
```
nehaberry: @fbalak for new test cases, IMO, it would be better to use the default sc and Blockpool.

tests/conftest.py
fbalak: Default storage class is now used. What default secret should be used? Shouldn't customer create a new project with secret for her/his pods?

nehaberry: @fbalak  If we are using the default SC, then we dont need to worry about the secret. The secret is already created and specified in the default RBD and CEPHFS SCs.

tests/conftest.py
```diff
@@ -894,6 +945,19 @@ def multi_pvc_factory_class(
...
+@pytest.fixture(scope='session')
+def multi_pvc_factory_session(
+    storageclass_factory_session,
+    project_factory_session,
+    pvc_factory_session
+):
+    return multi_pvc_factory_fixture(
+        storageclass_factory_session,
+        project_factory_session,
+        pvc_factory_session
```
nehaberry: @fbalak just to be sure, we will be creating all kinds of PVCs.. right ? RWO and RWX and attach them to PODs accordingly ?

fbalak: Right, but for exact configuration I will probably need your help.

tests/conftest.py
nehaberry: @fbalak 
 Lets agree to use default CBP and secrets.. If required, we can create new SCs, but with default Secret and CBP

fbalak:
Some pod is failing for me in test:
```
17:59:22 - MainThread - ocs_ci.ocs.ocp - ERROR - timeout expired: Timed Out: (120,)
17:59:22 - MainThread - ocs_ci.ocs.ocp - ERROR - Wait for Pod resource pod-test-rbd-d42adaa4bc2b40839187614d034845d3 to reach desired
 condition Running failed, last actual status was ContainerCreating
17:59:22 - MainThread - tests.helpers - ERROR - Pod pod-test-rbd-d42adaa4bc2b40839187614d034845d3 failed to reach Running
```
[describe pod](https://paste.fedoraproject.org/paste/rsf2Yq1TEyFygnSvRBD9bQ)

fbalak:
The configurations should be covered but there is one issue: We probably don't do cleanup for `Reclaim Policy:  Retain` PVCs well as they stay in our system in `Released` state.

tests/conftest.py
```diff
@@ -223,7 +246,7 @@ def factory(
 
     def finalizer():
         """
-        Delete the Ceph block pool
+        Delete the storageclass
```
nehaberry: @fbalak if we are using default SC, we dont need Delete Storageclass.. right ?

fbalak: Yes, but this is generic factory fixture that is not used for default storageclass. I made this change just to fix a docstring that I noticed is wrong.

tests/conftest.py
vasukulkarni: I agree with @nehaberry here, I dont think we have to deal with fixtures here.

tests/ecosystem/upgrade/conftest.py
```diff
@@ -0,0 +1,309 @@
...
+    interface,
+    pvc_factory,
+    pod_factory,
+    storageclass,
+    count,
+    access_mode,
+    volume_mode=None
+):
+    """
+    Create pods for upgrade testing. pvc_factory and pod_factory have to be
```
vasukulkarni: Not clear why the fixture has to be specific to upgrade testing, IMHO the choice should be left to users to do what they want to do in upgrade as per testcases, eg

```python
      with parallel() as p:
                 do_upgrade()
                 create_pods()
                 run_io()
                 destroy_pods()
      post_verification()
```

tests/ecosystem/upgrade/conftest.py
```diff
@@ -0,0 +1,309 @@
...
+        sc.data['metadata']['name'] += '-retain'
+        sc._name = sc.data['metadata']['name']
+        sc.create()
+        teardown_factory_session(sc)
+        scs[constants.RECLAIM_POLICY_RETAIN].append(sc)
+    return scs
+
+
+@pytest.fixture(scope='session')
+def pre_upgrade_filesystem_pods(
```
vasukulkarni: I suggest, just provide the fixtures as is and let users use it where they want it.

tests/conftest.py
fbalak: In the PR I used pods in factory scope so I added ability to call pod_factory in session scope. After that I added ability to use session scope for all factory fixtures (like pvc_factory_session) as it didn't make sense to me to add only one session scoped factory. In test code I use default storageclasses.

tests/ecosystem/upgrade/conftest.py
fbalak: This is a function that helps to prepare pods for my test case. You mean pre_upgrade_block_pods and pre_upgrade_filesystem_pods fixtures? These fixtures should be only upgrade specific as there are test cases with markers @order_pre_upgrade and @order_post_upgrade that use them and control what they do during upgrade (eg. test_start_pre_upgrade_pod_io).

tests/ecosystem/upgrade/conftest.py
fbalak: You mean to move them to tests/conftest.py?
I don't think that these fixtures should be used in other places. These are designed specially for upgrade and should be used only in upgrade test cases. 
Where do you want to use them?

tests/conftest.py
fbalak: Ok

tests/conftest.py
fbalak: Ok

tests/ecosystem/upgrade/test_resources.py
```diff
@@ -0,0 +1,94 @@
...
+
+from concurrent.futures import ThreadPoolExecutor
+from ocs_ci.framework.pytest_customization.marks import (
+    ignore_leftovers, order_pre_upgrade, order_post_upgrade
+)
+
+log = logging.getLogger(__name__)
+
+
+@order_pre_upgrade
```
petr-balogh: `@pre_upgrade` should be used

tests/ecosystem/upgrade/test_resources.py
```diff
@@ -0,0 +1,94 @@
...
+@ignore_leftovers
+def test_start_pre_upgrade_pod_io(pre_upgrade_pods_running_io):
+    """
+    Confirm that there are pods created before upgrade.
+    """
+    assert pre_upgrade_pods_running_io
+
+
+@order_post_upgrade
```
petr-balogh: Just `@post_upgrade` should be used

tests/ecosystem/upgrade/conftest.py
```diff
@@ -0,0 +1,309 @@
+import logging
+
+from ocs_ci.ocs import constants, resources
+import pytest
+import threading
```
petr-balogh: Pep8 imports. So threading in first category, pytst in second,  imports from ocs in third category and empty line between them.
```
Standard library imports.

Related third party imports.

Local application/library specific imports.
```

tests/ecosystem/upgrade/conftest.py
```diff
@@ -0,0 +1,309 @@
...
+            pvc=pvc,
+            raw_block_pv=raw_block_pv,
+            pod_dict_path=pod_dict,
+        ) for pvc in pvcs
+    ]
+    return pods
+
+
+@pytest.fixture(scope='session')
+def default_storageclasses(request, teardown_factory_session):
```
petr-balogh: Is there reason have such general fixture available just for upgrade? Maybe worth to move it to /tests/conftest.py WDYT?

tests/ecosystem/upgrade/test_resources.py
```diff
@@ -0,0 +1,94 @@
+import logging
+import pytest
```
petr-balogh: one empty line above to follow pep8 imports

tests/ecosystem/upgrade/conftest.py
fbalak: Sure

tests/ecosystem/upgrade/conftest.py
fbalak: Done

tests/ecosystem/upgrade/test_resources.py
```diff
@@ -1,7 +1,8 @@
 import logging
-import pytest
 
 from concurrent.futures import ThreadPoolExecutor
+import pytest
```
petr-balogh: This should be second section

tests/ecosystem/upgrade/test_resources.py
```diff
@@ -1,7 +1,8 @@
 import logging
-import pytest
 
 from concurrent.futures import ThreadPoolExecutor
```
petr-balogh: This is stil standart lib so first section

fbalak:
Jenkins verification job: Build #2739 - fbalak-pr1013

petr-balogh:
@fbalak BTW no need to verify everything again when you just moved imports if you have already verified

tests/conftest.py
```diff
@@ -53,6 +53,11 @@ def secret_factory_class(request):
     return secret_factory_fixture(request)
 
 
+@pytest.fixture(scope='session')
+def secret_factory_session(request):
+    return secret_factory_fixture(request)
```
nehaberry: @fbalak i think we should not use the secret_factory fixture anymore. the helper function (create_secret) uses admin_key and id which are no longer suggested for both CephFS and RBD.

Its ok to still have a CBP facotry session, but I dont think we need a secret factory session. We should always use the secret which is created by default or create secrets which do not use admin ID and key.

tests/conftest.py
```diff
@@ -151,6 +161,19 @@ def storageclass_factory_class(
     )
 
 
+@pytest.fixture(scope='session')
+def storageclass_factory_session(
+    request,
+    ceph_pool_factory_session,
+    secret_factory_session
```
nehaberry: Same. Please try and used the non-admin secrets created in openshift-storage for adding in SC. 

$ oc get secret -n openshift-storage|grep csi|grep 'kubernetes.io/rook'
rook-csi-cephfs-node                                        kubernetes.io/rook                    2      25h
rook-csi-cephfs-provisioner                                 kubernetes.io/rook                    2      25h
rook-csi-rbd-node                                           kubernetes.io/rook                    2      25h
rook-csi-rbd-provisioner                                    kubernetes.io/rook                    2      25h

tests/conftest.py
nehaberry: ok

tests/conftest.py
```diff
@@ -1197,3 +1272,38 @@ def nodes():
...
     return nodes
+
+
+@pytest.fixture(scope='session')
+def default_storageclasses(request, teardown_factory_session):
+    """
+    Returns dictionary with storageclasses. Keys represent reclaim policy of
+    storageclass. There are two storageclasses for each key. First is RBD based
+    and the second one is CephFS based. Storageclasses with Retain Reclaim
+    Policy are created from default storageclasses.
```
nehaberry: If possible, we can make it:
 Storageclasses with ReclaimPolicy: Retain  are created from default storageclasses.

tests/conftest.py
```diff
@@ -1197,3 +1272,38 @@ def nodes():
...
+        )
+        sc.reload()
+        scs[constants.RECLAIM_POLICY_DELETE].append(sc)
+        sc.data['reclaimPolicy'] = constants.RECLAIM_POLICY_RETAIN
+        sc.data['metadata']['name'] += '-retain'
+        sc._name = sc.data['metadata']['name']
+        sc.create()
+        teardown_factory_session(sc)
+        scs[constants.RECLAIM_POLICY_RETAIN].append(sc)
+    return scs
```
nehaberry: Looks great!!

tests/libtest/test_retain_pvc.py
```diff
@@ -0,0 +1,17 @@
+from ocs_ci.framework.testlib import libtest
+from ocs_ci.ocs import constants, resources
+
+
+@libtest
+def test_main(teardown_factory, pvc_factory):
+    sc = resources.ocs.OCS(
+        kind=constants.STORAGECLASS,
+        metadata={'name': 'ocs-storagecluster-ceph-rbd'}
```
nehaberry: @fbalak Sorry I got confused, why only RBD strageclass here ?

tests/ecosystem/upgrade/conftest.py
```diff
@@ -0,0 +1,276 @@
...
+        Run IO by executing FIO and deleting the file created for FIO on
+        the pod, in a while true loop. Will be running as long as
+        the upgrade_fio_file contains string 'running'.
+        """
+        while upgrade_fio_file.read_text() == 'running':
+            for pod in pre_upgrade_filesystem_pods:
+                log.warning(f"Running fio on fs pod {pod.name}")
+                pod.run_io(
+                    storage_type='fs',
+                    size='1GB'
```
nehaberry: @fbalak how do we make sure FIO keeps running for extended period?  Don't we need the **runtime** param  ? Ignore if it's already handled and I misunderstood.

tests/conftest.py
fbalak: Ok, I added this factory session fixture for the sake of consistency with the rest of the code.

tests/conftest.py
fbalak: Ok, but this should be changed in a different pull request as this is a thing that can affect more test cases and shouldn't be part of a upgrade test case PR.

tests/conftest.py
fbalak: In the pull request I I clone default storageclasses and change the name and reclaim policy to Retain in order to get a storageclass with Retain reclaim policy.

tests/conftest.py
fbalak: Thanks :)

tests/libtest/test_retain_pvc.py
fbalak: This is just a libtest that is not a part of a test suite that is run periodically (tests with marker @libtest are not run in CI). I created this test for us to make sure that we teardown PVCs with Retain Reclaim Policy correctly.

tests/ecosystem/upgrade/conftest.py
fbalak: `run_io` has a default value of runtime set to 60 seconds and the function is triggered repeatedly.

tests/ecosystem/upgrade/conftest.py
```diff
@@ -0,0 +1,276 @@
...
+            raw_block_pv=raw_block_pv,
+            pod_dict_path=pod_dict,
+        ) for pvc in pvcs
+    ]
+    return pods
+
+
+@pytest.fixture(scope='session')
+def pre_upgrade_filesystem_pods(
+    request,
```
ebenahar: @fbalak Looks really good!
Please consider using https://github.com/red-hat-storage/ocs-ci/blob/dcf872c69993d39b77a4efe97d02d36f40ce9775/tests/conftest.py#L791.
It's a session scoped fixture will take care of both pods creation on block and file and run IO in BG.

fbalak: Thanks, in future it would be good to merge this fixture and workload from this PR as this PR covers more configurations including Retain Reclaim Policy and different access and volume modes. For now I would keep it separate as there is created a lot of pods for the utilization during upgrade and it would be good plan and decide if this is a kind of workload that we want to run during execution of all test cases.

fbalak: There is also created ten pods for each configuration which can be quite a big number for normal test run.

ebenahar: @fbalak thanks for looking into this.
Currently, run_io_in_bg fixture is not being used at all. 
Feel free to modify it so it will create 10 pods for each configuration

petr-balogh:
@clacroix12 @shylesh @ebenahar @nehaberry @RazTamir  can we get some second approve to merge this and let it part of upgrade tests and jenkins jobs for upgrade?  Actually I thought this one got already merged but seems I was wrong so the results which we got just now from upgrade doesn't contain results from this :(

Thanks

PR #1014, opened 2019-11-15
Add bugzilla marker in test cases

jilju:
The expected error message is updated
Added bugzilla marker in tests

### DIFF
--- a/tests/manage/pv_services/test_dynamic_pvc_accessmodes_with_reclaim_policies.py
+++ b/tests/manage/pv_services/test_dynamic_pvc_accessmodes_with_reclaim_policies.py
@@ -153,14 +153,18 @@ class BaseDynamicPvc(ManageTest):
             marks=[
                 pytest.mark.polarion_id("OCS-525"),
                 pytest.mark.bugzilla("1751866"),
-                pytest.mark.bugzilla("1750916")]
+                pytest.mark.bugzilla("1750916"),
+                pytest.mark.bugzilla("1772990")
+            ]
         ),
         pytest.param(
             *[constants.CEPHFILESYSTEM, constants.RECLAIM_POLICY_DELETE],
             marks=[
                 pytest.mark.polarion_id("OCS-526"),
                 pytest.mark.bugzilla("1751866"),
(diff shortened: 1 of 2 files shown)


tests/manage/pv_services/test_dynamic_pvc_accessmodes_with_reclaim_policies.py
```diff
@@ -179,7 +183,7 @@ class TestRWODynamicPvc(BaseDynamicPvc):
 
     access_mode = constants.ACCESS_MODE_RWO
     storage_type = 'fs'
-    expected_pod_failure = 'Multi-Attach error for volume'
+    expected_pod_failure = 'MountVolume.MountDevice failed for volume'
```
sidhant-agrawal: The change in error message seems due to effect of the BZ 1772990.
Is the change the failure string needed then?

vasukulkarni: I think we dont need this error change. thanks for pointing out

jilju: The bugzilla 1772990 was raised for cephfilesystem only. RWO is working with cephblockpool. The bugzilla marker is added accordingly.
The error message is changed based on observation. Not sure whether it will be reverted back to the previous message along with the fix for bz  1772990

sidhant-agrawal: AFAIU with the fix for the bz 1772990, Multi-Attach error will be thrown for both RBD and CephFS based RWO PVC.
However I have asked for clarification here: https://bugzilla.redhat.com/show_bug.cgi?id=1772990#c14

jilju: The fix is done for box cephfs and rbd, so the old error message is expected to be back.
https://github.com/rook/rook/pull/4332
Reverted the error message back to previous one.

PR #1023, opened 2019-11-19
Extended waiting_time for alertmanager from 30 to 45 seconds

gklein:


vasukulkarni:
This needs to be root caused, are we sure its the timeout ?

gklein:
I agree we need to understand the root cause here, and I wasn't able to pinpoint it (yet).  I also don't know why we choose to wait 30 seconds for this pod to start.

I can only say that a few seconds after this failure I'v tried to query the pod status and it was in a running state.

So I'm suggesting we try it just so we can eliminate those statistics failures, which also breaks CI once in a while.

I'll try to investigate it further in case I see it again.

PR #1034, opened 2019-11-21
Added AWS IPI production deployment configurations

clacroix12:


clacroix12:
Initial deployment test passed using the 1 AZ config. Second has been kicked off that is using the 3AZ config.

PR #1035, opened 2019-11-21
When initializing set the config to environment set by ocs-ci

vasukulkarni:
Fixes: https://github.com/red-hat-storage/ocs-ci/issues/1036

### DIFF
--- a/ocs_ci/ocs/openshift_ops.py
+++ b/ocs_ci/ocs/openshift_ops.py
@@ -18,7 +18,7 @@ class OCP(object):
 
     """
     def __init__(self):
-
+        config.load_kube_config(config_file=os.getenv('KUBECONFIG'))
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/openshift_ops.py
```diff
@@ -18,7 +18,7 @@ class OCP(object):
 
     """
     def __init__(self):
-
+        config.load_kube_config(config_file=os.getenv('KUBECONFIG'))
```
petr-balogh: We have this path also available in our config: config.RUN['kubeconfig_location'] but I don't mind to take it also from env var as we expose there at the start of the execution.

ebenahar:
@vasukulkarni Would like to close old PRs, is this one still relevant?

Issue #270, opened 2019-06-21
Move code from `run_ocsci.py` to `ocsci.main`


mbukatov:
When is #207 resolved, drop `run_ocsci.py` from root repo, moving the code into `ocsci.main` module entirely.

mbukatov:
@dahorak originally planned to do this in https://github.com/red-hat-storage/ocs-ci/pull/261#discussion_r295866932, but we concluded that it's better to wait for #207 for us to resolve it properly with less work.

gklein:
Is there any more work required here or can we close this issue?

mbukatov:
I see that this hasn't been directly done, and now we has kind of pointless wrapper here:

https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/run_ocsci.py

To resolve this issue, we should drop this file.

PR #1042, opened 2019-11-21
Removed need for run_ocsci.py as main entry point

clacroix12:
Fixes: #270 

Note: Users will need to execute `pip install -r requirements.txt` to fix their `run-ci` entry points.


RazTamir:
LGTM..

Please verify and remember to send mail as soon as this is merge to remind about the need for pip install.

Thanks

clacroix12:
@zmc I didn't notice until now that our unit tests for the main function are passing arguments directly and testing the configuration according to what is being passed. My question is now that the main method does not actually accept arguments and instead just pulls directly from `sys` how should we go about testing functionality in these unit tests?

mbukatov:
> @zmc I didn't notice until now that our unit tests for the main function are passing arguments directly and testing the configuration according to what is being passed. My question is now that the main method does not actually accept arguments and instead just pulls directly from `sys` how should we go about testing functionality in these unit tests?

Good point, this is usually done like this:

```
def main(argv=None):
    argv = argv or sys.argv
```

zmc:
@clacroix12 I'd go with what @mbukatov said!

clacroix12:
Thanks @mbukatov and @zmc! Looks like that works perfectly. Going to kick off another deployment test just to officially verify on the latest changes.

PR #1044, opened 2019-11-22
Increase catalog source wait timeout to 15 minutes

obnoxxx:
Based on the discussion in https://bugzilla.redhat.com/show_bug.cgi?id=1770192

8 minutes may not be enough. According to this:

https://github.com/operator-framework/operator-lifecycle-manager/issues/947
https://github.com/operator-framework/operator-lifecycle-manager/issues/1098
https://github.com/operator-framework/operator-lifecycle-manager/pull/974
https://bugzilla.redhat.com/show_bug.cgi?id=1737081

More than 10 minutes have been seen.
Trying 15 minutes. We may need even more?

### DIFF
--- a/ocs_ci/ocs/resources/catalog_source.py
+++ b/ocs_ci/ocs/resources/catalog_source.py
@@ -108,7 +108,7 @@ class CatalogSource(OCP):
         return False
 
     @retry(ResourceInUnexpectedState, tries=4, delay=5, backoff=1)
-    def wait_for_state(self, state, timeout=480, sleep=5):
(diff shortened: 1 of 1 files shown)


RazTamir:
Actually it will now wait for 1 hour (900 * 4) because of this decorator
@retry(ResourceInUnexpectedState, tries=4, delay=5, backoff=1)

Which have 4 retries

vasukulkarni:
Dont think we need this one then?

clacroix12:
If we do need to wait for longer I suggest we simply increase the retry count instead of changing the duration we wait for.

obnoxxx:
sure

PR #1052, opened 2019-11-25
Fetch reclaim policy from PV instead of getting it from storage class

### LINKED ISSUE #1051: Fetch reclaim policy from PV instead of getting it from storage class
reclaim_policy of class PVC is fetching reclaim policy from storage class. While creating a PV, the reclaim policy will be set as provided in sc. This value can be modified in pv without modifying sc. So the value of reclaim policy has to be obtained from PV itself.

jilju:
Fixes #1051

### DIFF
--- a/ocs_ci/ocs/resources/pvc.py
+++ b/ocs_ci/ocs/resources/pvc.py
@@ -113,21 +113,14 @@ class PVC(OCS):
     @property
     def reclaim_policy(self):
         """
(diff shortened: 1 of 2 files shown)


jilju:
> PR validation [Job UNSTABLE](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/623//testReport) (some or all tests failed).

The result is matching with latest run.

Issue #991, opened 2019-11-12
AWS-UPI doesn't consider Availability zone while deployment


prsurve:
Param used in default_config.yaml

```
ENV_DATA:
  worker_availability_zones:
    - 'us-east-2a'
  master_availability_zones:
    - 'us-east-2a' 
```

https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/2089/

![Screenshot (13)](https://user-images.githubusercontent.com/32088367/68677772-ae841300-0582-11ea-82fe-ad49c571d7a5.png)

clacroix12:
This is because we use an external shell script to perform the UPI deployment. If we want to read in the availability zones properly we either need to make a change [here](http://git.app.eng.bos.redhat.com/git/openshift-misc.git/tree/v3-launch-templates/functionality-testing/aos-4_2/hosts/upi_on_aws-install.sh) or move away from using this shell script in favor of our own code that will perform the UPI deployment which reads in the availability zones.

prsurve:
If we change Availability_Zone_Count from the script that will bring up upi env on 1 az setup, Currently UPI for AWS is using just 2 Availability_Zone but not 3 changing [this](http://git.app.eng.bos.redhat.com/git/openshift-misc.git/tree/v3-launch-templates/functionality-testing/aos-4_2/hosts/upi_on_aws-install.sh#n20) will value to 1 or 3 will deploy cluster in a different zone

hnallurv:
Testing on 1 AZ and 3 AZs is top priority for 4.2. Many of our tests are
planned to be executed on these two environments. Fixing this issue will
help us move ahead faster with testing.

clacroix12:
> If we change Availability_Zone_Count from the script that will bring up upi env on 1 az setup, Currently UPI for AWS is using just 2 Availability_Zone but not 3 changing [this](http://git.app.eng.bos.redhat.com/git/openshift-misc.git/tree/v3-launch-templates/functionality-testing/aos-4_2/hosts/upi_on_aws-install.sh#n20) will value to 1 or 3 will deploy cluster in a different zone

[...]

Issue #1030, opened 2019-11-20
AWS-UPI deployment fail for worker with m5.4xlarge


prsurve:
worker create cmd failed with:- An error occurred (ValidationError) when calling the CreateStack operation: Parameter 'WorkerInstanceType' must be one of AllowedValues

cmd:- aws --region us-east-2 cloudformation create-stack --stack-name pratik-42-ocs-13-no0 --template-body file:///tmp/06_cluster_worker_nodes_2019-11-20-15-10-01_0.yaml --parameters ParameterKey=Index,ParameterValue=0 ParameterKey=InfrastructureName,ParameterValue=pratik-42-ocs-13-lhshw ParameterKey=RhcosAmi,ParameterValue=ami-06c85f9d106577272 ParameterKey=IgnitionLocation,ParameterValue=https://api-int.pratik-42-ocs-13.qe.rh-ocs.com:22623/config/worker ParameterKey=Subnet,ParameterValue=subnet-0707d79791122dec4 ParameterKey=WorkerSecurityGroupId,ParameterValue=sg-09dbae4b41477caeb ParameterKey=WorkerInstanceProfileName,ParameterValue=pratik-42-ocs-13-sg-WorkerInstanceProfile-1EQ6LCRLTWIT7 ParameterKey=WorkerInstanceType,ParameterValue=m5.4xlarge --capabilities CAPABILITY_NAMED_IAM

m5.4xlarge Instance type is not included in AllowedValues in worker_cloudformation_templates
template link [here](http://git.host.prod.eng.bos.redhat.com/git/openshift-misc.git/tree/v3-launch-templates/functionality-testing/aos-4_2/hosts/upi_on_aws-cloudformation-templates/06_cluster_worker_node.yaml#n37l)

vavuthu:
@clacroix12 , I almost submitted PR for this issue and I see issue is assigned to you. let me know if you if you are occupied with high priority work.

clacroix12:
@vavuthu I have a local change that passed testing so I was going to push my branch and create a PR, however I seem to be hitting authentication issues for gerrit. Not sure why as I've pushed branches to this repo before. If I'm unable to resolve this please feel free to open the PR yourself. It was a very small change, I just added `m5.4xlarge` to the `AllowedValues` list in the worker template.

clacroix12:
PR opened here: https://code.engineering.redhat.com/gerrit/#/c/187031/

clacroix12:
[...]

PR #1057, opened 2019-11-25
Add support to pass availability zone count to aws upi script

clacroix12:
Fixes: #991 

### DIFF
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -291,18 +291,23 @@ class AWSUPI(AWSBase):
             upi_env_vars = {
                 'INSTANCE_NAME_PREFIX': config.ENV_DATA['cluster_name'],
                 'AWS_REGION': config.ENV_DATA['region'],
-                'rhcos_ami': config.ENV_DATA['rhcos_ami'],
+                'rhcos_ami': config.ENV_DATA.get('rhcos_ami'),
                 'route53_domain_name': config.ENV_DATA['base_domain'],
                 'vm_type_masters': config.ENV_DATA['master_instance_type'],
                 'vm_type_workers': config.ENV_DATA['worker_instance_type'],
-                'num_workers': str(config.ENV_DATA['worker_replicas'])
+                'num_workers': str(config.ENV_DATA['worker_replicas']),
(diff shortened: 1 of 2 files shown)


ocs_ci/deployment/aws.py
```diff
@@ -289,20 +289,25 @@ def deploy_prereq(self):
...
             # setup necessary env variables
             upi_env_vars = {
-                'INSTANCE_NAME_PREFIX': config.ENV_DATA['cluster_name'],
-                'AWS_REGION': config.ENV_DATA['region'],
-                'rhcos_ami': config.ENV_DATA['rhcos_ami'],
-                'route53_domain_name': config.ENV_DATA['base_domain'],
-                'vm_type_masters': config.ENV_DATA['master_instance_type'],
-                'vm_type_workers': config.ENV_DATA['worker_instance_type'],
-                'num_workers': str(config.ENV_DATA['worker_replicas'])
+                'INSTANCE_NAME_PREFIX': config.ENV_DATA.get('cluster_name'),
```
petr-balogh: If some of those variables are mandatory I would not use get function as than the job will continue and you will have None values set on the places you don't probably want if mandatory parameters. So if some of those parameters are mandatory for UPI I would keep old logic just accessing data by key as it will directly fail here at the start with clear error what is missing to be passed in the config.

clacroix12: If they are truly mandatory the shell script is performing checks for them first thing so functionally it should be fine. However, the output from our code is easier to read than the output of the script so I'll revert some of these and only use `.get()` where necessary.

prsurve:
ocs-ci Jenkins job will fail because of m5.4xlarge is not included in AllowedValues of cloud formation templates for worker_node I already created an issue #1030

clacroix12:
> ocs-ci Jenkins job will fail because of m5.4xlarge is not included in AllowedValues of cloud formation templates for worker_node I already created an issue #1030

Yeah, this is apparently because we are automatically assuming AWS IPI with the PR test job. See the cluster conf:

Screen Shot 2019-11-27 at 10 37 07 AM

I'll raise an issue regarding this to see if we can allow for AWS UPI PR testing. In the meantime I will execute and verify locally that the AWS UPI deployment is working as intended with these changes.

PR #1072, opened 2019-11-27
Remove AWS requirement markers from all MCG tests, optimize imports

MeridianExplorer:


MeridianExplorer:
Closed because some tests are not ready to be ran on VMWare

PR #1074, opened 2019-11-27
Added deployment configuration for vsphere upi

clacroix12:
Jira: https://projects.engineering.redhat.com/browse/OCSQE-100

### DIFF
--- a/conf/deployment/vsphere/upi_1az_coreos_vmfs_3m_3w.yaml
+++ b/conf/deployment/vsphere/upi_1az_coreos_vmfs_3m_3w.yaml
@@ -1,0 +1,13 @@
+---
+DEPLOYMENT:
+  # TODO: update to false when supported
+  allow_lower_instance_requirements: true
+ENV_DATA:
+  platform: 'vsphere'
+  deployment_type: 'upi'
+  worker_replicas: 3
+  master_replicas: 3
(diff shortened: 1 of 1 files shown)


conf/deployment/vsphere/upi_1az_coreos_vmfs_3m_3w.yaml
```diff
@@ -0,0 +1,11 @@
+---
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
+ENV_DATA:
+  platform: 'vsphere'
+  deployment_type: 'upi'
+  worker_replicas: 3
+  master_replicas: 3
+  worker_num_cpus: '16'
```
petr-balogh: As we do not have such hardware yet we need to have as a WA at least one configuration file like this: https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/vsphere-upi-ocs4qe-lab.yaml and for now we will need to run against this one configuration file

clacroix12: Do we have a ticket tracking the work to support this so I can link it in the workaround TODO?

clacroix12: For now I've updated the values and will create an issue to address them when the production values are supported.

PR #1084, opened 2019-11-29
Test to check crush map before and after upgrade

fbalak:

### DIFF
--- a/tests/ecosystem/upgrade/test_configuration.py
+++ b/tests/ecosystem/upgrade/test_configuration.py
@@ -1,8 +1,9 @@
 import logging
 
 import pytest
+
 from ocs_ci.framework.pytest_customization.marks import (
-    order_pre_upgrade, order_post_upgrade
+    pre_upgrade, post_upgrade
 )
 from ocs_ci.ocs.resources import pod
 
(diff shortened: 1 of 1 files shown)


tests/ecosystem/upgrade/test_configuration.py
```diff
@@ -0,0 +1,54 @@
...
+    Loads CRUSH map before upgrade by `test_load_crush_map` test case.
+
+    Returns:
+        str: String consisting of CRUSH map before upgrade
+    """
+    crash_map = get_crush_map()
+    log.info(f"Pre upgrade CRASH map: {crash_map}")
+    return crash_map
+
+
```
ebenahar: Polarion ID?

hnallurv: Please change CRASH to CRUSH wherever applicable

fbalak: done

petr-balogh:
Is that verified @fbalak ? If so please put verifed label

ebenahar:
@petr-balogh can we merge this or you think we need to verify something before?

Issue #1047, opened 2019-11-24
Async output in the run_cmd()


Avilir:
When running in Debug level, the outputs of the command output (only  in debug) are written to the log  not exactly after the command that ran, so it hard to debug.
the command should be printed with the output.

gklein:
Is it possible that the we are collecting stdout/stderr separately and this is causing an out of order log file?

I've implemented a version run_cmd() with live output, which also captures stdout and stderr in order.  You may try the [PR](https://github.com/red-hat-storage/ocs-ci/pull/1096/) locally to see if it might help solve this issue too.

Avilir:
closing on lack of interest

Issue #350, opened 2019-07-03
ocs_ci.utility.utils.run_cmd() should log output as it is generated


zmc:
The current implementation of `run_cmd()` uses `subprocess.run()` instead of `subprocess.Popen`. As a result, it has to wait for the command to finish executing in order to log any output.

For short-running commands, this is hardly noticeable; for longer ones like `openshift-install`, it's problematic. We should write the output of the command to stdout/stderr as normal, while buffering internally.

zmc:
An example that also shows how the logging ends up out-of-order:
```
$ cat cmd_test.sh
#!/bin/bash

out () {
  echo "stdout"
  echo "stderr" >&2
}

out
sleep 1
out sleep 2
out
sleep 1
out
```

```
$ ./cmd_test.sh
stdout
stderr
stdout
stderr
stdout
stderr
stdout
stderr
```

```
$ python cmd_test.py
INFO:ocs_ci.utility.utils:Executing command: ./cmd_test.sh
DEBUG:ocs_ci.utility.utils:Command output: stdout
stdout
stdout
stdout

WARNING:ocs_ci.utility.utils:Command warning:: stderr
stderr
stderr
stderr
```

gklein:
It seems we have a [run_async() method](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L778) that can be used for the similar case [as described](https://github.com/red-hat-storage/ocs-ci/issues/350#issue-463936265).

Does it answers this request ? Can we close the issue?

clacroix12:
I don't think `run_async` solves this issue. Unless I'm missing something, executing `run_async` doesn't wait for the command to finish, but it seems that `run_async.async_communicate()` still does. What we want to get out of this is the live output from long running commands in real time without having to wait for the process to end in order to log it.

gklein:
> I don't think `run_async` solves this issue. Unless I'm missing something, executing `run_async` doesn't wait for the command to finish, but it seems that `run_async.async_communicate()`(which is how we get the return code, stdout, and stderr of the cmd) still does. What we want to get out of this is the live output from long running commands in real time without having to wait for the process to end in order to log it.

Yes, you are right.  run_async()  does not solve the live output requirement, which might be handy for real time debugging especially for long running tests and the OCP/OCS deployment phase.

Here is a possible implementation for live output which captures both stdout  and stderr.

https://github.com/red-hat-storage/ocs-ci/pull/1096/

PR #1096, opened 2019-12-02
Added live output of stdout/stderr to run_cmd()

gklein:

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -1,3 +1,4 @@
+import io
 import json
 import logging
 import os
@@ -5,8 +6,10 @@ import platform
 import random
 import shlex
 import string
+import select
 import subprocess
 import time
+import timeit
 from copy import deepcopy
 from shutil import which
 
@@ -383,7 +386,6 @@ def run_cmd(cmd, secrets=None, timeout=600, **kwargs):
 
     Args:
         cmd (str): command to run
-
         secrets (list): A list of secrets to be masked with asterisks
             This kwarg is popped in order to not interfere with
             subprocess.run(``**kwargs``)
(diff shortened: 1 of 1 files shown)


gklein:
This should resolve #350 and might also be able to resolve #1047

ocs_ci/utility/utils.py
```diff
@@ -397,27 +399,43 @@ def run_cmd(cmd, secrets=None, timeout=600, **kwargs):
...
-            f"\nError is {mask_secrets(r.stderr.decode(), secrets)}"
-        )
-    return mask_secrets(r.stdout.decode(), secrets)
+    ) as p:
+        while True:
+            now = timeit.default_timer()
+            if (now - start_time) > timeout:
+                raise subprocess.TimeoutExpired(masked_cmd, timeout)
+            reads = [p.stdout.fileno(), p.stderr.fileno()]
+            ret = select.select(reads, [], [])
```
vasukulkarni: @zmc since you raised original request with subprocess.Popen, do we need select here to achieve this?

ocs_ci/utility/utils.py
```diff
@@ -397,27 +399,43 @@ def run_cmd(cmd, secrets=None, timeout=600, **kwargs):
...
+            for fd in ret[0]:
+                if fd == p.stdout.fileno():
+                    read = p.stdout.readline()
+                    full_output.write(read)
+                    log.info(f"Command output: {read}")
+                if fd == p.stderr.fileno():
+                    read = p.stderr.readline()
+                    full_output.write(read)
+                    log.warning(f"Command warning: {mask_secrets(read, secrets)}")
```
clacroix12: Is it necessary for us to have "Command output:" and "Command warning:" in these log messages? I feel like simply logging the message at info/warning level is enough without having to preface it with these strings.

I know this is carried over from what `run_cmd` has been doing but I'm bringing it up since we are here looking at this code.

ocs_ci/utility/utils.py
gklein: In case we want the live output of both stdout and stderr I guess we will need to use either select or poll. For now, all of my attempts to implement it with poll run into a deadlock, but if we agree we need live output of both pipes with poll, I can put some more time into it.

ocs_ci/utility/utils.py
```diff
@@ -397,27 +399,43 @@ def run_cmd(cmd, secrets=None, timeout=600, **kwargs):
...
-        raise CommandFailed(
-            f"Error during execution of command: {masked_cmd}."
-            f"\nError is {mask_secrets(r.stderr.decode(), secrets)}"
-        )
-    return mask_secrets(r.stdout.decode(), secrets)
+    ) as p:
+        while True:
+            now = timeit.default_timer()
+            if (now - start_time) > timeout:
+                raise subprocess.TimeoutExpired(masked_cmd, timeout)
```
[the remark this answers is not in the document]
gklein: Agree, I will fix it in case we continue with this approach

ocs_ci/utility/utils.py
gklein: You have raised a very good point here. I've tried to keep the the the original functionality as it was before this attempt to implement a live output. If we agree to redirect both stdout and stderr through a single pipe, it will be much easier to implement and will require a smaller change in the code.

https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.stdout

clacroix12: That sounds great, however the only thing I am curious about is the log level. Right now we are logging at different levels for stdout and stderr. If they are combined, I imagine this becomes more difficult to do.

ocs_ci/utility/utils.py
vasukulkarni: @zmc ping? , I know we can do with select, but i guess then it can be achieved with subprocess.run as well.

vasukulkarni:
@gklein can you rebase, I think Zack is on PTO

ebenahar:
Closing. please re-open if needed

Issue #1097, opened 2019-12-02
Use RC build as the default build for GA testing


petr-balogh:
We need to make sure that the RC build is the default and everyone will test with RC build till GA.

PR #1098, opened 2019-12-02
Set RC build as the default

petr-balogh:
Fixes: #1097

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -45,7 +45,7 @@ DEPLOYMENT:
   # ocs_csv_channel: "alpha"
   # you can overwrite the image for ocs operator catalog source by following parameter:
   # ocs_registry_image: "quay.io/rhceph-dev/ocs-registry:4.2-32.9b6c93e.master"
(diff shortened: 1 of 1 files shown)


ocs_ci/framework/conf/default_config.yaml
```diff
@@ -45,7 +45,7 @@ DEPLOYMENT:
   # ocs_csv_channel: "alpha"
   # you can overwrite the image for ocs operator catalog source by following parameter:
   # ocs_registry_image: "quay.io/rhceph-dev/ocs-registry:4.2-32.9b6c93e.master"
-  default_ocs_registry_image: "quay.io/rhceph-dev/ocs-registry:latest"
+  default_ocs_registry_image: "quay.io/rhceph-dev/ocs-registry:4.2-rc1"
```
vasukulkarni: would have been better with rc-latest label.

petr-balogh: Probably it would but we don't have such tag https://quay.io/repository/rhceph-dev/ocs-registry?tag=latest&tab=tags  I can talk to Boris about this tag, but let's merge this one to make sure we installing RC as default and then we can change to some different tag once we will have.

PR #1107, opened 2019-12-04
Fix test_pvc_no_space_left

### LINKED ISSUE #1106: Test test_pvc_no_space_left fails with assertion error
The test TestPVCFullWithIORWX.test_pvc_no_space_left is failing with error -
"AssertionError: No error raised during FIO to fill the device with data assert False"

fio is not failing when tried to utilize complete available space with cephfs based volume. Instead, it writes completely to utilize 100%.

jilju:
Removed check for fio failure. fio need not throw error when trying to utilize 100% space.
The intention of test is to validate full utilization of available space.

Fixes #1106 

### DIFF
--- a/tests/manage/pv_services/test_no_space_left.py
+++ b/tests/manage/pv_services/test_no_space_left.py
@@ -112,14 +112,12 @@ class TestPVCFullWithIORWX(ManageTest):
             'fs', size=self.pvc_size_gb, io_direction='write', runtime=60
         )
         log.info("Waiting for IO results")
-        catch_error = False
         try:
             self.pod_obj1.get_fio_results()
         except ex.CommandFailed as cf:
             if "No space left on device" not in str(cf):
                 raise
             else:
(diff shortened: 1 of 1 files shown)


tests/manage/pv_services/test_no_space_left.py
```diff
@@ -112,14 +112,12 @@ def test_pvc_no_space_left(self, base_setup):
             'fs', size=self.pvc_size_gb, io_direction='write', runtime=60
         )
         log.info("Waiting for IO results")
-        catch_error = False
         try:
             self.pod_obj1.get_fio_results()
         except ex.CommandFailed as cf:
             if "No space left on device" not in str(cf):
```
RazTamir: This will still fail on Cephfs right?

jilju: fio is passing with cephfs, so CommandFailed error won't be raised.

PR #1115, opened 2019-12-05
Noobaa bucket quota alerts test

fbalak:
Add test case for NooBaa bucket quota alerts


tests/manage/monitoring/conftest.py
```diff
@@ -863,3 +865,101 @@ def workload_storageutilization_85p_cephfs(
...
+        Returns:
+            str: Name of utilized bucket
+        """
+        nonlocal mcg_obj
+        nonlocal bucket_name
+        nonlocal awscli_pod
+        # run_time of operation
+        run_time = 60 * 11
+        awscli_pod.exec_cmd_on_pod(
+            'dd if=/dev/zero of=/tmp/testfile bs=1M count=500'
```
ebenahar: will the fact that the files are filled with zeros have an effect on the amount of data written to the bucket?

fbalak: I am not sure but I think that it shouldn't. Otherwise it would be bug.

Issue #1120, opened 2019-12-05
For the UI deployment we need to skip the OCS deployment via ocs-ci


petr-balogh:
For UI deployment which will be handled by openshift-console we need to deploy just the catalog source and create the namespace and operator group based on documentation.

Ticket: https://projects.engineering.redhat.com/browse/OCSQE-96

The UI deployment will be done outside of ocs-ci for now.

PR #1122, opened 2019-12-05
Prepare for UI deployment

petr-balogh:
Fixes: #1120

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -213,7 +213,7 @@ class Deployment(object):
         )
         run_cmd(f"oc create -f {catalog_source_manifest.name}", timeout=2400)
         catalog_source = CatalogSource(
-            resource_name='ocs-catalogsource',
+            resource_name=constants.OPERATOR_CATALOG_SOURCE_NAME,
             namespace='openshift-marketplace',
         )
         # Wait for catalog source is ready
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -173,14 +173,10 @@ class Deployment(object):
             )
             _ocp.exec_oc_cmd(command=taint_cmd)
 
-    @staticmethod
-    def get_olm_and_subscription_manifest():
+    def create_catalog_source(self):
         """
-        This method prepare manifest for deploy OCS operator and subscription.
-
-        Returns:
-            tuple: Path to olm deploy and subscription manifest
-
+        This prepare catalog source manifest for deploy OCS operator from
+        quay registry.
         """
         image = config.DEPLOYMENT.get('ocs_registry_image', '')
         upgrade = config.DEPLOYMENT.get('upgrade', False)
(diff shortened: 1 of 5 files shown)


petr-balogh:
I also split deployment to more parts and moved catalog source outside of deploy-with-olm as this will not be used when deploy from stage. Verified locally and worked, so running pr verification job here again with normal deployment.

This is prerequisite for running openshift-console out of ocs-ci, I will reconsider tomorrow and maybe I will decide to execute from ocs-ci but this will be handled by another PR as this change will be used also by @dahorak for stage deployment.

ocs_ci/deployment/deployment.py
```diff
@@ -237,38 +243,46 @@ def get_olm_and_subscription_manifest():
...
+        ui_deployment = config.DEPLOYMENT.get('ui_deployment')
+        if ui_deployment:
+            logger.info("Preparing stuff for deployment by openshift-console")
+        else:
+            logger.info("Deployment of OCS via OCS operator")
+            self.label_and_taint_nodes()
+        run_cmd(f"oc create -f {constants.OLM_YAML}")
+        self.create_catalog_source()
+        if ui_deployment:
+            logger.info(
```
vasukulkarni: This code can move into after line 265: as they are same if statement.

petr-balogh: Cannot cause I need to do all the operation above

vasukulkarni: There are two if statements and one can go away?

petr-balogh: Look at the code @vasukulkarni , I need to print some message at the start, in case of UI deployment I want to inform different way that preparing stuff for deployment via openshift-console, in case of non UI deployment I need to also call function for labeling nodes, than I need to run command on line 270 and 271 in both UI and non UI deployment and than here in case of UI deployment I need to exit (return here).
So here need to be the condition once more I cannot move the return or info message above cause I still need to do what is done on lines 270 and 271.   So what is the issue here?

ocs_ci/deployment/deployment.py
```diff
@@ -193,31 +189,41 @@ def get_olm_and_subscription_manifest():
...
+        catalog_source_manifest = tempfile.NamedTemporaryFile(
+            mode='w+', prefix='catalog_source_manifest', delete=False
         )
         templating.dump_data_to_temp_yaml(
-            olm_yaml_data, olm_manifest.name
+            catalog_source_data, catalog_source_manifest.name
         )
+        run_cmd(f"oc create -f {catalog_source_manifest.name}", timeout=2400)
+        catalog_source = CatalogSource(
+            resource_name='ocs-catalogsource',
```
shylesh: do you want to take resource_name from constants.OPERATOR_CATALOG_SOURCE_NAME

petr-balogh: done

ocs_ci/deployment/deployment.py
vasukulkarni: I see it merged, but this is what I am suggesting for better reaadability (the first *if* as I read did nothing but print some info and was handled in second *if*) https://github.com/red-hat-storage/ocs-ci/pull/1129

PR #1127, opened 2019-12-06
Remove securityContext from container section 

prsurve:
Having this privileged will not allow us to catch this type of [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1777384) 

Because making this privileged dc pod will be able to write on mount point even when normal pod doesn't have permission to write on that mount point

### DIFF
--- a/ocs_ci/templates/app-pods/fedora_dc.yaml
+++ b/ocs_ci/templates/app-pods/fedora_dc.yaml
@@ -10,6 +10,8 @@ spec:
       labels:
         name: fedorapod
     spec:
+      securityContext:
+        fsGroup: 2000
       serviceAccountName: admin
       restartPolicy: Always
       volumes:
@@ -25,9 +27,6 @@ spec:
             cpu: "150m"
         command: ["/bin/bash", "-ce", "tail -f /dev/null" ]
         imagePullPolicy: IfNotPresent
-        securityContext:
-          capabilities: {}
-          privileged: true
         volumeMounts:
         - mountPath: /mnt
           name: fedora-vol


prsurve:
As expected test case failed for running io on fs mount

```ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc -n namespace-test-4f6aa6ae2b924ad7bfcd3b370c4bea13 --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig rsh pod-test-rbd-2af03433baf146158cc5ddbbee0afcd1-1-bm9sl fio --name=fio-rand-readwrite --filename=/mnt/fio-rand-readwrite --readwrite=randrw --bs=4K --direct=1 --numjobs=1 --time_based=1 --runtime=60 --size=4G --iodepth=4 --invalidate=1 --fsync_on_close=1 --rwmixread=75 --ioengine=libaio --output-format=json. Error is fio: pid=0, err=13/file:filesetup.c:162, func=open, error=Permission denied command terminated with exit code 1```

ebenahar:
@prsurve @gklein @clacroix12 do we still want to merge this now that https://github.com/red-hat-storage/ocs-ci/pull/1180 is in?

gklein:
> @prsurve @gklein @clacroix12 do we still want to merge this now that #1180 is in?

If this patch was created in order to tackle #1151 , I think we can drop it as we agreed this will be [documented](https://bugzilla.redhat.com/show_bug.cgi?id=1777384#c23). 

#1180 is applying that workaround exactly as it appears in the docs.

prsurve:
@ebenahar reason for this patch was if we don't apply this #1180  patch dc pod were able to right on that mount point

PR #1129, opened 2019-12-06
Fix redundant if's for ui_deployment

vasukulkarni:
Unresolved comment from: https://github.com/red-hat-storage/ocs-ci/pull/1122

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -264,17 +264,18 @@ class Deployment(object):
         ui_deployment = config.DEPLOYMENT.get('ui_deployment')
         if ui_deployment:
             logger.info("Preparing stuff for deployment by openshift-console")
-        else:
-            logger.info("Deployment of OCS via OCS operator")
-            self.label_and_taint_nodes()
-        run_cmd(f"oc create -f {constants.OLM_YAML}")
-        self.create_catalog_source()
-        if ui_deployment:
+            run_cmd(f"oc create -f {constants.OLM_YAML}")
+            self.create_catalog_source()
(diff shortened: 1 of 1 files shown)


ocs_ci/deployment/deployment.py
```diff
@@ -264,17 +264,18 @@ def deploy_ocs_via_operator(self):
...
             logger.info(
                 "Skipping the rest of the deployment because it will be done "
                 "by UI deployment!"
             )
             return
+        else:
+            logger.info("Deployment of OCS via OCS operator")
+            self.label_and_taint_nodes()
+        run_cmd(f"oc create -f {constants.OLM_YAML}")
+        self.create_catalog_source()
```
petr-balogh: I don't se any extra readability added here by this change and you duplicate the lines from 267 and 268 here.

vasukulkarni: Line 268 is a function, hence not duplicated its inside *if* , I haven't seen the coding style where the first *if* does nothing but just print info and logical block is handled somewhere down the line with same *if* statement.

petr-balogh: At least store this cmd: oc create -f {constants.OLM_YAML}
In variable an use on both places if you insist on this change.

petr-balogh: I updated as you want in my follow up PR for UI deployment here: https://github.com/red-hat-storage/ocs-ci/pull/1130  this pr is still WIP

vasukulkarni: sure, thanks, will close this one then.

PR #1136, opened 2019-12-08
Yaml file for catalog source for ui testing

shyRozen:
As decided in ux group to implement creating the namespace in test code itself and the yaml will be deployed via infra before the install test run.


petr-balogh:
We already have function in our deployment and This catalog source definition here: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/templates/ocs-deployment/catalog-source.yaml no meed to have it again @shyRozen

shyRozen:
As Petr comment there is already a template yaml for this closing the PR

PR #1140, opened 2019-12-08
Added a check for a min/max cluster name length

### LINKED ISSUE #1133: Limit aws resource with 32 chars
'Some resource name (such as: api-int-jenkins-aws-upi-1az-rhel7-3m-3w-tier1) would excceed aws limitation (32), pls adjust your cluster name for a shorter name'\n+ exit 1\n"

See more details here:  https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/2838//console

gklein:
Fixes #1133 by enforcing a min/max cluster name length


ocs_ci/ocs/constants.py
```diff
@@ -464,3 +464,7 @@
 REPO_MAPPING = {
     '4.2.0': OCP4_2_REPO
 }
+
+# Cluster name limits
+CLUSTER_NAME_MIN_CHARACTERS = 5
```
shylesh: @gklein ,

Any reason behind min char length limit ?

gklein: I think we need to use some minimal text to identify each cluster and to help prevent duplicate names.

I don't mind to reduce it, but I think it would be risky to allow 1 or 2 character long cluster names.

shylesh: hmm makes sense.

ocs_ci/framework/exceptions.py
```diff
@@ -3,3 +3,15 @@ class ClusterPathNotProvidedError(Exception):
...
+
+class ClusterNameLengthError(Exception):
+
+    def __init__(self, name, min, max):
+        self.name = name
+        self.min = min
+        self.max = max
+
+    def __str__(self):
+        return (f"Cluster Name '{self.name}' is {len(self.name)} characters long "
```
petr-balogh:
```python
return (
    f"......"
    f"...."
)
```

ocs_ci/ocs/constants.py
```diff
@@ -464,3 +464,7 @@
...
+CLUSTER_NAME_MAX_CHARACTERS = 17
```
petr-balogh: So cluster name should be between 6-16 chars If I understand correctly right?  Please can this be added to some of our README? For example here:  https://github.com/red-hat-storage/ocs-ci/blob/master/docs/usage.md . thanks

ocs_ci/framework/exceptions.py
gklein: Just to clarify, should I fix the formatting here?

ocs_ci/ocs/constants.py
gklein: It will only raise the ClusterNameLengthError() when the cluster name is < 5 or > 17 characters, so the range valid range would be 5-17 chars.

gklein: Will be documented with #1143

ocs_ci/framework/exceptions.py
petr-balogh: just brackets https://ocs-ci.readthedocs.io/en/latest/docs/coding_guidelines.html

You have return (f"fsadfsa"

But we rather see the 
```
return (
    f"content on other line if it's too long above like you have"
) # bracket at the new line
```

petr-balogh:
Please also wait with the merge this PR. We need to align the jenkins jobs with this requirement or handle by some cut after 16 chars.

Opened the issue in our jenkins job repository here: https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/issues/23 . So this one can be merged after we have merged PR in the jenkins for the mentioned issue.

I added DO NOT MERGE label for now here.

Thanks

ocs_ci/framework/exceptions.py
gklein: Make sense. Fixed with the last commit.

petr-balogh:
> Please also wait with the merge this PR. We need to align the jenkins jobs with this requirement or handle by some cut after 16 chars.
> 
> Opened the issue in our jenkins job repository here: https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/issues/23 . So this one can be merged after we have merged PR in the jenkins for the mentioned issue.
> 
> I added DO NOT MERGE label for now here.
> 
> Thanks

Adding WIP to the subject of the PR to really prevent to be merged now

ocs_ci/framework/exceptions.py
```diff
@@ -3,3 +3,17 @@ class ClusterPathNotProvidedError(Exception):
 
     def __str__(self):
         return "Please provide a --cluster-path that is empty or non-existant."
+
+
+class ClusterNameLengthError(Exception):
+
+    def __init__(self, name, min, max):
+        self.name = name
+        self.min = min
```
clacroix12: Would it be useful to default to using `CLUSTER_NAME_MIN_CHARACTERS` and `CLUSTER_NAME_MAX_CHARACTERS` if `min`/`max` are not passed in to the `__init__`? This way we don't need pass the values through to the error init if we are using our defaults anyways.

petr-balogh:
@gklein we merged: https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/merge_requests/120 , so if verified please add verified label and we can merge. Thanks

ocs_ci/framework/exceptions.py
gklein: Yes, make sense. I'll adapt it accordingly.

ocs_ci/ocs/constants.py
gklein: I'll add the relevant documentation under this PR now that it was dismissed

gklein:
> @gklein we merged: https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/merge_requests/120 , so if verified please add verified label and we can merge. Thanks

Ok, I'll reverify now with the all of the  latest suggestions.

 It seems I can't add labels, do I need any special permission ?

petr-balogh:
@gklein  added you to ocs-qe team, you should get invitation from github in email and you need to approve I think.

vasukulkarni:
Remvoing "needs testing" flag.

vasukulkarni:
I think you will hit a conflict with other PR since both are changing --cluster-name

ocs_ci/framework/exceptions.py
```diff
@@ -1,10 +1,30 @@
...
 
 class ClusterPathNotProvidedError(Exception):
 
     def __str__(self):
         return "Please provide a --cluster-path that is empty or non-existant."
 
 
+class ClusterNameLengthError(Exception):
+    def __init__(
+        self, name, min=CLUSTER_NAME_MIN_CHARACTERS, max=CLUSTER_NAME_MAX_CHARACTERS
```
clacroix12: Apologies for catching this so late, but I think we might want to use names other than min/max for these arguments as they shadow the `min` and `max` builtin functions. I don't want to hold up this PR so I'm going to approve anyways, but perhaps we can change them to `minimum`/`maximum`, `min_length`/`max_length`, or something like this in a follow up PR.

vasukulkarni: good point!, luckily within this class scope so bit safe.

gklein: Good catch! I'll submit another PR to fix it.

gklein: See #1156 for a fix

Issue #1138, opened 2019-12-08
Fix mon leftover from test_add_capacity


RazTamir:

RazTamir:
the issue is being analyzed from ocs-ci results for OCS_Downstream_4-2_tier1-AWS-IPI (BUILD ID: rc2 RUN ID: 1575680073)

PR #1142, opened 2019-12-08
Fix mon as a leftover from add node and capacity test cases

RazTamir:
Run them last

Fixes: #1138


ebenahar:
@RazTamir Looks like the tests are being executed in alphabetic order and the name starts in the name of the root dir (tests/). So, for example, in order to have the cluster expansion test cases run last, we'll need to change their dir to be under /tests/z/manage/cluster/nodes/cluster_expension... (notice the 'z' dir)/

Found this solution which could be good - https://pytest-ordering.readthedocs.io/en/develop/

vasukulkarni:
> https://pytest-ordering.readthedocs.io/en/develop/

@ebenahar I think we should keep that ordering complexity out of ocs-ci, I understand how that ordering works, but i find it neither intuitive to follow nor easily maintainable with growing number of tests and trying to do ordering that way.

We assumed our tests are independent with each other(with fixtures setting up and tearing down the resources they created), I am still not understanding why one test failure should affect another, maybe I am missing the complete issue.

vasukulkarni:
> @RazTamir Looks like the tests are being executed in alphabetic order and the name starts in the name of the root dir (tests/). So, for example, in order to have the cluster expansion test cases run last, we'll need to change their dir to be under /tests/z/manage/cluster/nodes/cluster_expension... (notice the 'z' dir).
> 
> Found this solution which could be good - https://pytest-ordering.readthedocs.io/en/develop/

I read some more in the email thread, I see why its required

vasukulkarni:
I read some more in the mail thread and understand the issue now.

vasukulkarni:
As is the PR looks fine to me, but we can do numbering too inside folders, eg: 
tests/manage/1/basic_tests
tests/manage/2/cluster

RazTamir:
Not sure why it was merged :)
It wasn't verified

vasukulkarni:
Is this not fixed yet?  I saw the jenkins job run before merge, Please add "needs testing label" or wip branch and we can avoid any confusion.

RazTamir:
it was fixed but you were faster than me merging before I added the Verified label :)

vasukulkarni:
relief :) will keep an eye on verified.

PR #1146, opened 2019-12-10
Fix tier1 must-gather test

RazTamir:

Fixes: #1147

### DIFF
--- a/ocs_ci/ocs/openshift_ops.py
+++ b/ocs_ci/ocs/openshift_ops.py
@@ -11,6 +11,7 @@ import ocs_ci.ocs.defaults as default
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


vasukulkarni:
I have a similar issue for this one https://github.com/red-hat-storage/ocs-ci/issues/1036,  Personally I see there is nothing wrong with the script, It works in local environment and had also worked in jenkins before, we have to figure out what changed is causing this failure in jenkins(the recent cleanup issue might be affecting it as well but want to see where the context is getting reset)

RazTamir:
I fixed it to not use the OCP dynamic client.
The issue came from there and I don't want to even start understand what is going on there so just refactored to use ocs-ci infrastructure

Issue #1152, opened 2019-12-11
Method `exec_ceph_cmd()` silently drops list items from the output


mbukatov:
## Description of the problem

Method `exec_ceph_cmd()` silently drops list items from the output if an item evaluates to False.

## Steps to reproduce

Run:

```
ceph_osd_ls = ct_pod.exec_ceph_cmd(ceph_cmd="ceph osd ls")
```

## Actual Results

```
(Pdb) p ceph_osd_ls
[1, 2]
```

## Expected Results

```
(Pdb) p ceph_osd_ls
[0, 1, 2]
```

## Additional Information

When run from command line, output of `ceph osd ls` looks like this:

```
# ceph osd ls --format json-pretty
[
    0,
    1,
    2
]
#
```

But because `exec_ceph_cmd()` method contains following code:

https://github.com/red-hat-storage/ocs-ci/blob/47ffc61c6146b2f9a04c0e515ca828a9bb7aa630/ocs_ci/ocs/resources/pod.py#L206

The first item from the list is dropped silently. I don't think this makes any sense.

Suggested fix includes:
- logging of the raw output (`out` variable) in DEBUG level
- dropping the filtering altogether

## List of workaround

When this BZ is fixed by dropping the filtering, the following workarounds should be removed:

- [ ] https://github.com/red-hat-storage/ocs-ci/blob/f7b31ee7e1c90430bb8a18dc008981962c269f93/tests/manage/monitoring/conftest.py#L990

This is may not be complete.

gklein:
The reason that first item is drop is because "0" is being evaluated to False under the existing filter. 

```
In: out = [0, 1, 2, 3]

#With a filter:
In: [item for item in out if item]
Out: [1, 2, 3]

#Without any filter:
In: [item for item in out]
Out: [0, 1, 2, 3]

#Filter only an empty string (but not a zero number, empty list or empty dictionary) 
In: [item for item in out if not item == '' ]
Out: [0, 1, 2, 3]
```

mbukatov:
> The reason that first item is drop is because "0" is being evaluated to False under the existing filter.

I understand that, which is why I listed the code line in question. What I don't understand is why is it a good idea to do such filtering in a general method like `exec_ceph_cmd()`.

gklein:
> > The reason that first item is drop is because "0" is being evaluated to False under the existing filter.
> 
[...]

PR #1157, opened 2019-12-11
Change exec_ceph_cmd() so it won't filter anything of the command output

gklein:
A zero int or float which evaluates to false with the current condition, should not be filtered out.

Fixes: #1152 

### DIFF
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -203,7 +203,7 @@ class Pod(OCS):
 
         # For some commands, like "ceph fs ls", the returned output is a list
         if isinstance(out, list):
-            return [item for item in out if item]
+            return [item for item in out if not item == ""]
         return out
 
     def get_storage_path(self, storage_type='fs'):
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -203,7 +203,7 @@ class Pod(OCS):
 
         # For some commands, like "ceph fs ls", the returned output is a list
         if isinstance(out, list):
-            return [item for item in out if not item == ""]
+            return [item for item in out]
         return out
 
     def get_storage_path(self, storage_type='fs'):
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -188,6 +188,7 @@ class Pod(OCS):
 
         Returns:
             dict: Ceph command output
+            (For some commands, like "ceph fs ls", the returned output is a list)
 
         Raises:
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/resources/pod.py
```diff
@@ -203,7 +203,7 @@ def exec_ceph_cmd(self, ceph_cmd, format='json-pretty'):
 
         # For some commands, like "ceph fs ls", the returned output is a list
         if isinstance(out, list):
-            return [item for item in out if item]
+            return [item for item in out if not item == ""]
```
vasukulkarni: I am up for dropping this line completely, let the test program handle what it needs rather than preprocess here?

gklein: Fine with me, I thought there was an intention here to filter empty strings, but let remove it and run some testing to make sure it doesn't break anything.

gklein:
> PR validation [Job UNSTABLE](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/729//testReport) (some or all tests failed).

There was only one failure in a must-gather test:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/729/testReport/tests.manage.z_cluster.test_must_gather/TestMustGather/test_must_gather/

Seems to be due to #1036

ocs_ci/ocs/resources/pod.py
```diff
@@ -203,7 +203,7 @@ def exec_ceph_cmd(self, ceph_cmd, format='json-pretty'):
 
         # For some commands, like "ceph fs ls", the returned output is a list
         if isinstance(out, list):
-            return [item for item in out if item]
+            return [item for item in out]
```
petr-balogh: ok this is just copy of the out list here, this case I guess we can remove completely this line and line above and just return out on line 207 as we will return list anyway. Or do I miss something here?

petr-balogh: and if we return list in some cases we should put it to returns section in the docstring. So comment in line 204 should be moved there to docstring.

ocs_ci/ocs/resources/pod.py
```diff
@@ -188,6 +188,7 @@ def exec_ceph_cmd(self, ceph_cmd, format='json-pretty'):
 
         Returns:
             dict: Ceph command output
+            (For some commands, like "ceph fs ls", the returned output is a list)
```
petr-balogh: You can just start line like:

´´´
list: For some commands...
´´´

ebenahar:
Closing. Please re-open if needed

PR #1158, opened 2019-12-12
Add health check to the upgrade test

petr-balogh:
Ticket:
https://projects.engineering.redhat.com/browse/OCSQE-145

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -6,11 +6,13 @@ rook cluster. This works with assumptions that an OCP cluster is already
 functional and proper configurations are made for interaction.
 """
 
-import logging
 import base64
+import logging
 import random
-import yaml
 import re
+import threading
+import yaml
+from time import sleep
 
 import ocs_ci.ocs.resources.pod as pod
 from ocs_ci.ocs.exceptions import UnexpectedBehaviour
@@ -100,7 +102,9 @@ class CephCluster(object):
         self.mon_count = 0
         self.mgr_count = 0
         self.osd_count = 0
-
+        self.health_error_status = None
+        self.health_monitor_enabled = False
+        self.health_monitor = None
         self.scan_cluster()
         logging.info(f"Number of mons = {self.mon_count}")
         logging.info(f"Number of mds = {self.mds_count}")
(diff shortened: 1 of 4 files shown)


petr-balogh:
Running verification here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/2977/

ocs_ci/ocs/cluster.py
```diff
@@ -427,6 +431,86 @@ def check_ceph_pool_used_space(self, cbp_name):
...
+
+    def run(self):
+        self.ceph_cluster.health_monitor_enabled = True
+        while self.ceph_cluster.health_monitor_enabled and (
+            not self.ceph_cluster.health_error_status
+        ):
+            sleep(self.sleep)
+            health_status = self.ceph_cluster.get_ceph_health(detail=True)
+            if "HEALTH_ERROR" in health_status:
+                self.ceph_cluster.health_error_status = (
```
ebenahar: This means that only 1 error occurrence will be assigned to self.ceph_cluster.health_error_status IIUC.
What if we want more then 1 error occurrence to be saved?

petr-balogh:
Verified:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/2982/console

ocs_ci/ocs/cluster.py
petr-balogh: I was thinking about adding more as well but then I realized that if I run this health check each 5 second and I will hit HEALTH_ERROR and this one will repeat all the time that I will spam this list with the same error all the time and no more additional error.  I thought about concentrating for first occurrence of HEALTH_ERROR only and stop monitoring after that, let process the rest of the upgrade which will most likely fail later on anyway and collect must gather and raise the exception with that first error.

@nehaberry or do you think that we should try to collect all the errors again and again?

Thanks

ebenahar: @petr-balogh OK. Make sense to have it saved only once. Thanks!

ocs_ci/ocs/cluster.py
```diff
@@ -427,6 +431,86 @@ def check_ceph_pool_used_space(self, cbp_name):
...
+                    self.ceph_cluster.ceph_status()
```
shylesh: @petr-balogh , is that call ```self.ceph_cluster.get_ceph_status() ```?

petr-balogh: ops, I changed the name of function with prefix get_ and forgot to change it here. Thanks for noticing. Submitting fix in a minute.

PR #1164, opened 2019-12-12
Fix add MDS tier1 test case

RazTamir:
Removed the usage of ignore_leftovers as it cause following test to fail


RazTamir:
> @RazTamir thinking maybe this case should not be part of tier1

I agree.
Will take care of it in a different PR

vasukulkarni:
LGTM

PR #1172, opened 2019-12-13
Setting redundancy & curator values to default values for logging

Shrivaibavi:


keesturam:
Please change the commit message.
1) I believe 'single' redundancy is the default configuration and that is why we are changing to single?
2) schedule change is to increase the frequency of curator

Shrivaibavi:
http://pastebin.test.redhat.com/821731

PR #1176, opened 2019-12-15
Change MCG bucket creation tests markers

MeridianExplorer:
* Replace all 'tier2' markers with 'performance'
* Reorder param calls
* Add an always skipif marker

Signed-off-by: Ben <belimele@redhat.com>

### DIFF
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -82,8 +82,6 @@ aws_platform_required = pytest.mark.skipif(
     reason="Tests are not running on AWS deployed cluster"
 )
 
-skip_always = pytest.mark.skipif(True, reason="Test was marked for intentional skipping")
-
 # Filter warnings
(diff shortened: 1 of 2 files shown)


ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -82,6 +82,8 @@
     reason="Tests are not running on AWS deployed cluster"
 )
 
+skip_always = pytest.mark.skipif(True, reason="Test was marked for intentional skipping")
```
ebenahar: please try passing the reason from outside

MeridianExplorer:
> Please create a workaround issue to track the skipped test cases.

#1218

ocs_ci/framework/pytest_customization/marks.py
MeridianExplorer: For now, I replaced this with a constant message saved in `ERRATIC_TIMEOUTS_SKIP_REASON`

PR #1182, opened 2019-12-16
Define the craft_s3_command AWSCLI connection timeout as 60 seconds

MeridianExplorer:
Should help in cases of unstable connections while retrieving objects


MeridianExplorer:
@ebenahar - It's not a specific issue as much as 'unstable connections' are. I don't think there's any way for us to improve the condition other than adding this workaround. Thus, I'm not sure that opening an issue will contribute. What do you think?

MeridianExplorer:
After further testing, I found out this command didn't help in our case. PR closed.

Issue #1190, opened 2019-12-16
Fix OSD count number


petr-balogh:
I've hit the issue during the upgrade where we check exact number of OSDs:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/3110/consoleFull

this is after this change:
https://github.com/red-hat-storage/ocs-ci/pull/1173

So we need to count number of OSDs from count*replica

PR #1191, opened 2019-12-16
Fix OSD count number verification

petr-balogh:
Fixes: #1190

### DIFF
--- a/ocs_ci/ocs/resources/ocs.py
+++ b/ocs_ci/ocs/resources/ocs.py
@@ -253,7 +253,10 @@ def ocs_install_verification(timeout=600, skip_osd_distribution_check=False):
         timeout=timeout
     )
     # osds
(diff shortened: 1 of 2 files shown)


petr-balogh:
Running verification here:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/3117/console

Got confirmation from Travis that this is right way how to calculate the number of OSDs from strorageCluster device set.

petr-balogh:
I see the deployment stage passed and it's looking now for correct number of 3 osds. So marking as verified now as I do not expect any other issue when gather this in upgrade when they are just copy paste the same lines.

petr-balogh:
@vasukulkarni @RazTamir @ebenahar can you please approve if you are still online? :) thanks

Issue #1185, opened 2019-12-16
Run-ci ignores "region" parameter in config files when using RHEL UPI


AviadP:
workaround: using .aws/config file with default region

annagitel:
![noRegion](https://user-images.githubusercontent.com/57480476/70920662-a14bbf80-202b-11ea-8361-a67bc802bfb1.png)

clacroix12:
This is due to how we are instantiating our cloudformation client. We are not passing the `region_name` so it was defaulting to whatever was configured in `~/.aws/config`. I'll have a fix shortly.

PR #1192, opened 2019-12-17
Pass region to boto clients properly

clacroix12:
Fixes: #1185 

### DIFF
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -285,7 +285,7 @@ class AWSUPI(AWSBase):
             self.client = boto3.client(
                 'ec2', region_name=config.ENV_DATA['region']
             )
-            self.cf = boto3.client('cloudformation')
+            self.cf = boto3.client('cloudformation', region_name=self.region)
 
     class OCPDeployment(BaseOCPDeployment):
         def __init__(self):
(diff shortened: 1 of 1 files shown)


ocs_ci/deployment/aws.py
```diff
@@ -438,7 +438,7 @@ def gather_worker_data(self, suffix='no0'):
         stack_name = f'{self.cluster_name}-{suffix}'
         resource = self.cf.list_stack_resources(StackName=stack_name)
         worker_id = self.get_worker_resource_id(resource)
-        ec2 = boto3.resource('ec2')
+        ec2 = boto3.resource('ec2', region_name=self.region)
```
shylesh: I just have a question here, self.region will be fixed through out the execution and workers can be in 3 different regions in case of 3AZ. So would it still work?

gklein: Availability zones (AZ) are not the same as regions. Each AZ is a logical data center living in one specific region. While region is a geographical location that can have one ore more AZ.

As longs as we don't need to support multiple regions in the same test (only multiple AZ in the same region) , I goes this will work.

shylesh: Ah got it. Thanks for the clarification.

PR #1198, opened 2019-12-17
Adding Memory and CPU option for RHEL in vSphere platform

vavuthu:

### DIFF
--- a/conf/deployment/vsphere/upi_1az_rhel_3m_3w.yaml
+++ b/conf/deployment/vsphere/upi_1az_rhel_3m_3w.yaml
@@ -17,3 +17,5 @@ ENV_DATA:
   rhel_worker_prefix: 'rhel1'
   rhel_user: 'test'
   mixed_cluster: False
+  rhel_num_cpus: '16'
+  rhel_memory: '65536'


petr-balogh:
> @petr-balogh ,
> 
> Do we need this for AWS RHEL workers as well ?

@shylesh that's the good question.

How we do in AWS rhel? Do we set the same as we have for normal workers?  @vavuthu why do we differentiate between rhel and normal (CoreOS) workers?  

Thinking about why we just do not use the same values from https://github.com/red-hat-storage/ocs-ci/pull/1198/files#diff-c2468e5c96b4d141e508c8f926f3ab10R11 and https://github.com/red-hat-storage/ocs-ci/pull/1198/files#diff-c2468e5c96b4d141e508c8f926f3ab10R13 ?

vavuthu:
> > @petr-balogh ,
> > Do we need this for AWS RHEL workers as well ?
> 
> @shylesh that's the good question.
> 
> How we do in AWS rhel? Do we set the same as we have for normal workers? @vavuthu why do we differentiate between rhel and normal (CoreOS) workers?
> 
> Thinking about why we just do not use the same values from https://github.com/red-hat-storage/ocs-ci/pull/1198/files#diff-c2468e5c96b4d141e508c8f926f3ab10R11 and https://github.com/red-hat-storage/ocs-ci/pull/1198/files#diff-c2468e5c96b4d141e508c8f926f3ab10R13 ?

I would like to differentiate the between CoreOS and RHEL. and provide option to user to change for only RHEL if needs.

shylesh:
> > @petr-balogh ,
> > Do we need this for AWS RHEL workers as well ?
> 
> @shylesh that's the good question.
> 
> How we do in AWS rhel? Do we set the same as we have for normal workers? @vavuthu why do we differentiate between rhel and normal (CoreOS) workers?
> 
> Thinking about why we just do not use the same values from https://github.com/red-hat-storage/ocs-ci/pull/1198/files#diff-c2468e5c96b4d141e508c8f926f3ab10R11 and https://github.com/red-hat-storage/ocs-ci/pull/1198/files#diff-c2468e5c96b4d141e508c8f926f3ab10R13 ?

@petr-balogh ,

For RHEL workers on AWSUPI we are going with whatever m5.4xlarge instances have in terms of cpus and other resources, only storage part we have parameterized. Same case with RHCOS workers as well.

petr-balogh:
@vavuthu what values are used if those two vars are not set in the config:
```
  rhel_num_cpus: '16'
  rhel_memory: '65536' 
```
?

Do we have somewhere set default values or they are somewhere in the template?
I think that the best approach is when we do not specify those values we should use the values from:
 `worker_num_cpus` and `memory` as this will use the same values as we have for current worker nodes WDYT?

vavuthu:
> @vavuthu what values are used if those two vars are not set in the config:
> 
> ```
>   rhel_num_cpus: '16'
>   rhel_memory: '65536' 
> ```
> 
> ?
> 
> Do we have somewhere set default values or they are somewhere in the template?
> I think that the best approach is when we do not specify those values we should use the values from:
> `worker_num_cpus` and `memory` as this will use the same values as we have for current worker nodes WDYT?

@petr-balogh If we don't specify the values, it will take from https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/templates/ocp-deployment/scale_up_terraform.tfvars.j2#L14

PR #1199, opened 2019-12-17
Extend the timeout length of some of the MCG tests

MeridianExplorer:
To prevent false-negatives on slower machines.


ocs_ci/ocs/resources/mcg.py
```diff
@@ -550,7 +550,7 @@ def _check_mirroring():
             return all(results)
 
         try:
-            for mirroring_is_complete in TimeoutSampler(140, 5, _check_mirroring):
+            for mirroring_is_complete in TimeoutSampler(140, 10, _check_mirroring):
```
MeridianExplorer: From what I observed the cmd wouldn't execute quickly enough sometimes (>5 secs) - thus extended the wait just for the sake of cleaner logs and easier tracking.

ocs_ci/ocs/resources/mcg.py
```diff
@@ -587,7 +587,7 @@ def _check_state():
             return False
 
         try:
-            for reached_state in TimeoutSampler(180, 10, _check_state):
+            for reached_state in TimeoutSampler(600, 40, _check_state):
```
RazTamir: 3 minutes changed to 10 minutes.
why?

The reason I'm asking is to ensure we don't have a regression in performance of reaching the desired state

MeridianExplorer: The change is because we block and enable IO on the S3 bucket by using BucketPolicies. Even after drilling down the NB code and verifying that it refreshes in the intervals it should, we noticed it sometimes just takes a very long time for the policy to apply, so basically - waiting for AWS.

ebenahar: @MeridianExplorer is this PR still needed then?

MeridianExplorer: @ebenahar - Yep, there are still two changes for other timings in the MCG tests. Just needs further research and testing. There's also a bug regarding this specific part (backingstore refuses to update its state) - waiting for a response from NB

ebenahar: ack

MeridianExplorer:
Now assigned to @bernerhat

PR #1206, opened 2019-12-18
Deployment from stage

dahorak:
* Configure OperatorSource and related secret for deployment from stage.
* enhance `censor_values` function to process dictionaries recursively
* censor values in data printed to log in `dump_data_to_temp_yaml`

* Deployment now works only in VMWare environment.

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -256,8 +256,8 @@ class Deployment(object):
         # generate quay token
         credentials = {
             "user": {
-                "username": config.DEPLOYMENT.get("stage_quay_username"),
-                "password": config.DEPLOYMENT.get("stage_quay_password"),
+                "username": config.DEPLOYMENT["stage_quay_username"],
+                "password": config.DEPLOYMENT["stage_quay_password"],
             }
         }
         token = requests.post(
@@ -265,7 +265,7 @@ class Deployment(object):
             data=json.dumps(credentials),
             headers={'Content-Type': 'application/json'},
         ).json()['token']
-        stage_ns = config.DEPLOYMENT.get("stage_namespace")
+        stage_ns = config.DEPLOYMENT["stage_namespace"]
 
         # create Secret
         stage_os_secret = templating.load_yaml(
@@ -339,12 +339,9 @@ class Deployment(object):
         else:
             logger.info(f"Default channel will be used: {default_channel}")
             subscription_yaml_data['spec']['channel'] = default_channel
-        if (
-            config.DEPLOYMENT.get('stage')
-            and config.DEPLOYMENT.get('stage_namespace')
-        ):
+        if config.DEPLOYMENT.get('stage'):
             subscription_yaml_data['spec']['source'] = (
-                config.DEPLOYMENT.get('stage_namespace')
+                config.DEPLOYMENT['stage_namespace']
             )
         subscription_manifest = tempfile.NamedTemporaryFile(
             mode='w+', prefix='subscription_manifest', delete=False
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -6,6 +6,9 @@ import logging
 import os
 import tempfile
 import time
+
+import json
+import requests
 from copy import deepcopy
 
 from ocs_ci.deployment.ocp import OCPDeployment as BaseOCPDeployment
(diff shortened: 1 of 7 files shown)


ocs_ci/deployment/deployment.py
```diff
@@ -225,6 +228,68 @@ def create_catalog_source(self):
...
 
+    def create_stage_operator_source(self):
+        """
+        This prepare operator source for OCS deployment from stage.
+        """
+        logger.info("Adding Stage Secret")
+        # generate quay token
+        credentials = {
+            "user": {
+                "username": config.DEPLOYMENT.get("stage_quay_username"),
```
petr-balogh: I wouldn't use .get method as this is mandatory parameter and this case we will silently continue with None value in username.

I would rather fail with key error here and user will directly see that those parameters has to be set.

Same below

ocs_ci/deployment/deployment.py
```diff
@@ -225,6 +228,68 @@ def create_catalog_source(self):
...
+                "password": config.DEPLOYMENT.get("stage_quay_password"),
+            }
+        }
+        token = requests.post(
+            url='https://quay.io/cnr/api/v1/users/login',
+            data=json.dumps(credentials),
+            headers={'Content-Type': 'application/json'},
+        ).json()['token']
+        stage_ns = config.DEPLOYMENT.get("stage_namespace")
```
petr-balogh: What about to have some default stage_namespace defined and used here:
`config.DEPLOYMENT.get("stage_namespace", defaults.STAGE_NAMESPCE)`
something like this. Or if we don't want to have default stage namespace I would also directly access key without get method for the reason mentioned above.

ocs_ci/deployment/deployment.py
```diff
@@ -243,6 +308,13 @@ def subscribe_ocs(self):
         channel = config.DEPLOYMENT.get('ocs_csv_channel')
         if channel:
             subscription_yaml_data['spec']['channel'] = channel
+        if (
+            config.DEPLOYMENT.get('stage')
+            and config.DEPLOYMENT.get('stage_namespace')
```
petr-balogh: you can save the stage_namespace to some variable above line 311 and use it on both places instead of call get method twice

ocs_ci/templates/ocs-deployment/stage-operator-source-secret.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: SECRET-NAME
```
petr-balogh: add -PLACEHOLDER as we discussed.

Same for others placeholders

dahorak: Fixed by e8b56d6.

ocs_ci/deployment/deployment.py
dahorak: Fixed by 6ab36d8.

ocs_ci/deployment/deployment.py
dahorak: Since the namespace is tightly connected to the used credentials (or vice versa), I think it is better to make the `stage_namespace` also mandatory. Fixed by 6ab36d8.

ocs_ci/deployment/deployment.py
dahorak: Since the setting of `stage_namespace` is also mandatory (see previous comment), I've removed it from the condition and used it without the `get(...)` method. Fixed by 6ab36d8.

ocs_ci/utility/utils.py
```diff
@@ -1430,17 +1430,22 @@ def wait_for_co(operator):
 
 def censor_values(data_to_censor):
     """
-    This function censor values in dictionary keys that match pattern defined
-    in config_keys_patterns_to_censor in constants.
+    This function recursively censor string and numeric values in dictionary
```
vasukulkarni: I think sentence needs correction.

ocs_ci/utility/utils.py
```diff
@@ -1430,17 +1430,22 @@ def wait_for_co(operator):
...
-        for pattern in constants.config_keys_patterns_to_censor:
-            if pattern in key:
-                data_to_censor[key] = '*' * 5
+        if isinstance(data_to_censor[key], dict):
+            censor_values(data_to_censor[key])
+        elif isinstance(data_to_censor[key], (str, int, float)):
+            for pattern in constants.config_keys_patterns_to_censor:
+                if pattern in key.lower():
+                    data_to_censor[key] = '*' * 5
+    return data_to_censor
```
vasukulkarni: missing in doc string, probably from old PR, but you can fix it :)

dahorak: Doc string updated.

ocs_ci/utility/utils.py
dahorak: Doc string updated.

dahorak:
Validation jobs:
* [installation in VMWare environment](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/3197/),
* [tier1 test suite in VMWare environment](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/3201)

PR #1216, opened 2019-12-22
Increase 'oc login' timeout

RazTamir:
The default is 24h which is not enough for tier4 tests.
 Increased to 1 week

### DIFF
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -9,7 +9,7 @@ import tempfile
 import time
 import yaml
 
-from ocs_ci.ocs.constants import RSYNC_POD_YAML, STATUS_RUNNING
+from ocs_ci.ocs.constants import RSYNC_POD_YAML, STATUS_RUNNING, WEEK2SEC
 from ocs_ci.ocs.exceptions import (
     CommandFailed,
     NotSupportedFunctionError,
@@ -356,7 +356,9 @@ class OCP(object):
         Returns:
             str: output of login command
         """
-        command = f"oc login -u {user} -p {password}"
+        command = (
+            f"oc login -u {user} -p {password} --request-timeout={WEEK2SEC}"
+        )
         status = run_cmd(command)
         return status
 
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -132,6 +132,7 @@ IMAGE_REGISTRY_RESOURCE_NAME = "cluster"
 ADMIN_USER = 'admin'
 GB = 1024 ** 3
 GB2KB = 1024 ** 2
+WEEK2SEC = '604800s'
 
 # Reclaim Policy
 RECLAIM_POLICY_RETAIN = 'Retain'


ocs_ci/ocs/ocp.py
```diff
@@ -356,7 +356,9 @@ def login(self, user, password):
         Returns:
             str: output of login command
         """
-        command = f"oc login -u {user} -p {password}"
+        command = (
+            f"oc login -u {user} -p {password} --request-timeout={WEEK2SEC}"
+        )
```
gklein: It seems this approach won't work, as this option was intended for a connection timeout.

From `oc options`:
`--request-timeout='0': The length of time to wait before giving up on a single server request. Non-zero values`

Quick check:
```
▶ date
Mon Dec 23 11:44:36 IST 2019
▶ oc login -u kubeadmin -p *******************  --request-timeout=604800s
Login successful.

You have access to 53 projects, the list has been suppressed. You can list all projects with 'oc projects'

Using project "default".
▶ oc get oauthaccesstokens -A
NAME                                          USER NAME    CLIENT NAME                    CREATED                         EXPIRES                         REDIRECT URI                                                                 SCOPES
__m2eO8BPygdFVZB-5Bl8v9UGzxrv-I80bX4UxP7Ua4   kube:admin   openshift-challenging-client   2019-12-23 09:44:59 +0000 UTC   2019-12-24 09:44:59 +0000 UTC   https://oauth-openshift.apps.ocs.qe.rh-ocs.com/oauth/token/implicit   user:full
```

gklein:
I think the best approach to tackle this problem is to use a [service account token](https://cookbook.openshift.org/accessing-an-openshift-cluster/why-do-i-keep-having-to-login-from-the-command-line.html)

ocs_ci/ocs/ocp.py
RazTamir: ahh thanks for checking...

Couldn't find a way to change this default session

gklein:
@RazTamir Here is another possible solution on how to extend the default token expiration configuration:

https://docs.openshift.com/container-platform/4.2/authentication/configuring-internal-oauth.html#oauth-configuring-internal-oauth_configuring-internal-oauth

ocs_ci/ocs/ocp.py
shylesh: How about running oc login command during resource leftover check after every test , Just a dirty approach not sure if it works.

gklein: It might be possible, but maybe we can handle it very specifically by returning back to "system:admin" only at the end of the tests that changes the logged-in user.

Here is a possible fix #1219

RazTamir: system:admin doesn't have a default timeout for session?

gklein: I don't think so, as it does not create a token. 

I think this is also the reason why we didn't saw this issue before the new prometheus test was added (oc is defaulting to system:admin if you are not logging in explicitly).

```
▶ oc login -u system:admin
Logged into "https://api.gklein-ocs.qe.rh-ocs.com:6443" as "system:admin" using existing credentials.

You have access to 52 projects, the list has been suppressed. You can list all projects with 'oc projects'

Using project "default".

▶ oc get oauthaccesstokens
No resources found.
```

PR #1217, opened 2019-12-23
Increase memory limit for dc pods

prsurve:
Fixes: #1252 


prsurve:
> @prsurve Please link this PR to an issue so we'll know the reason for it

Updated with issue

PR #1237, opened 2019-12-31
Adds PVC cretion timeout values for 4.2 and params for different volume sizes

keesturam:


tests/e2e/test_pvc_creation_performance.py
```diff
@@ -39,38 +39,55 @@ def base_setup(
...
+            sc_name=self.sc_obj.name, size=pvc_size
         )
         helpers.wait_for_resource_state(pvc_obj, constants.STATUS_BOUND)
         pvc_obj.reload()
         teardown_factory(pvc_obj)
         create_time = helpers.measure_pvc_creation_time(
             self.interface, pvc_obj.name
         )
-        if create_time > 1:
+        logging.info(f"PVC created in {create_time} seconds")
```
vavuthu: NIT: above logging should be after if ( line 79 ) condition

Issue #929, opened 2019-10-23
Installation fails on pods "alertmanager-main-0" not found


mbukatov:
When I use ocs-ci to deploy OCS cluster in AWS (with default downstream configuration), the deployment fails on  `pods "alertmanager-main-0" not found` error.

That said, I see that the alermanager pods are running:

```
$ oc get pods --all-namespaces | grep alertmanager
openshift-monitoring                                    alertmanager-main-0                                                   3/3     Running     0          49m
openshift-monitoring                                    alertmanager-main-1                                                   3/3     Running     0          49m
openshift-monitoring                                    alertmanager-main-2                                                   3/3     Running     0          49m
```

We need to look into this to see if this is a ocs-ci or a product bug. This problem could also affects results in CI in a negative way.

See full traceback in gist: https://gist.github.com/mbukatov/9287daffc1bb79d225946d45b64feb27

ocs-ci version: commit fba3174

mbukatov:
This could be a product bug, but so far there is no evidence for this hypothesis.

gklein:
It was reproduced on my system today

```12:16:55 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-monitoring --kubeconfig /Users/gklein/cluster-path/auth/kubeconfig get ConfigMap cluster-monitoring-config -n openshift-monitoring -o yaml
12:16:56 - MainThread - ocs_ci.ocs.monitoring - INFO - Successfully created configmap cluster-monitoring-config
12:16:56 - MainThread - ocs_ci.deployment.deployment - INFO - Waiting 30 seconds...
12:17:26 - MainThread - ocs_ci.ocs.ocp - INFO - Waiting for a resource(s) of kind Pod identified by name 'alertmanager-main-0' and selector None to reach desired condition Running
12:17:26 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-monitoring --kubeconfig /Users/gklein/cluster-path/auth/kubeconfig get Pod alertmanager-main-0 -n openshift-monitoring -o yaml
12:17:26 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-monitoring --kubeconfig /Users/gklein/cluster-path/auth/kubeconfig get Pod alertmanager-main-0 -n openshift-monitoring
12:17:27 - MainThread - ocs_ci.ocs.ocp - INFO - status of alertmanager-main-0 was Terminating, but we were waiting for Running
[...]

PR #1253, opened 2020-01-03
Add retry in get function for wait for resource

petr-balogh:
Fixes: #929
During attempts for waiting for resource, it can happen that the
resource is deleted beteween the quieries and then the get function fail.
This fix adding additional retry to get function which is used in
wait_for_resource function.

This makes function more robust and proof for such issues whene the
resource is deleted and recreated between the calls.

The origin behavior for get function remain the same with no extra
attempt/retry or wait.

### DIFF
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -162,7 +162,7 @@ class OCP(object):
 
     def get(
         self, resource_name='', out_yaml_format=True, selector=None,
-        all_namespaces=False
+        all_namespaces=False, retry=0, wait=3
     ):
         """
         Get command - 'oc get <resource>'
@@ -172,6 +172,8 @@ class OCP(object):
             out_yaml_format (bool): Adding '-o yaml' to oc command
             selector (str): The label selector to look for
             all_namespaces (bool): Equal to oc get <resource> -A
+            retry (int): Number of attempts to retry to get resource
+            wait (int): Number of seconds to wait between attempts for retry
 
         Example:
             get('my-pv1')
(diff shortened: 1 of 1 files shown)


petr-balogh:
Running verification job here:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/3400/console

petr-balogh:
Another attempt of verification with upgrade here: 
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/3415/console

From previous execution I opened bug to OCP:
https://bugzilla.redhat.com/show_bug.cgi?id=1787597

petr-balogh:
Verification job with upgrade:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/3425/console

Executed also tier1, it's running in PR verification job.

petr-balogh:
I see tier1 PR verification job has 100% success rate.

ocs_ci/ocs/ocp.py
```diff
@@ -189,7 +191,24 @@ def get(
             command += f" --selector={selector}"
         if out_yaml_format:
             command += " -o yaml"
-        return self.exec_oc_cmd(command)
+        retry += 1
```
RazTamir: I don't think this will be desired.
If you pass retry = 0, it will do retry once which is not what you wanted

ocs_ci/ocs/ocp.py
```diff
@@ -428,7 +447,10 @@ def wait_for_resource(
 
                 # Only 1 resource expected to be returned
                 if resource_name:
-                    status = self.get_resource(resource_name, column)
+                    retry = int(timeout / sleep if sleep else timeout / 1)
```
RazTamir: I'm not sure what is executed first but please check that timeout / sleep if sleep, will not fail if sleep is 0 as it will be dividing by 0

petr-balogh: This line is here exactly for preventing this issue ;)

ocs_ci/ocs/ocp.py
petr-balogh: We need to run the code below once. If retry is set to 0 we will not not entry to while loop. The behavior is really same. Retry=0 will execute command once, of retry is 1 will execute once and will retry once more.

PR #1254, opened 2020-01-03
Add OCP 4.2 config file

petr-balogh:
For this ticket: https://projects.engineering.redhat.com/browse/OCSQE-166
We need to have also static file for OCP 4.2 to allow us to easily pass
from job parameter which one OCP config file to use for specific
version.

This will be also needed if we decide later on to have OCP 4.3 as a
default.


vasukulkarni:
Does this need verification?

clacroix12:
@vasukulkarni this is simply a new file that as of right now matches our default values so I don't think verification is necessary.

PR #1262, opened 2020-01-06
Fetch PV reclaim policy before deleting the PVC in the pv deletion perf testcase

keesturam:
Fixes: https://github.com/red-hat-storage/ocs-ci/issues/1266

### DIFF
--- a/tests/e2e/test_pvc_deletion_performance.py
+++ b/tests/e2e/test_pvc_deletion_performance.py
@@ -68,11 +68,12 @@ class TestPVCDeletionPerformance(E2ETest):
         helpers.wait_for_resource_state(pvc_obj, constants.STATUS_BOUND)
         pvc_obj.reload()
         pv_name = pvc_obj.backed_pv
+        pvc_reclaim_policy = pvc_obj.reclaim_policy
         teardown_factory(pvc_obj)
         pvc_obj.delete()
         logging.info('Start deletion of PVC')
         pvc_obj.ocp.wait_for_delete(pvc_obj.name)
-        if pvc_obj.reclaim_policy == constants.RECLAIM_POLICY_DELETE:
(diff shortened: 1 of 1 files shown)


vasukulkarni:
@keesturam please add description explaining the issue or link to github

tests/e2e/test_pvc_deletion_performance.py
```diff
@@ -68,11 +68,12 @@ def test_pvc_deletion_measurement_performance(self, teardown_factory, pvc_size):
         helpers.wait_for_resource_state(pvc_obj, constants.STATUS_BOUND)
         pvc_obj.reload()
         pv_name = pvc_obj.backed_pv
+        pvc_reclaim_policy = pvc_obj.reclaim_policy
         teardown_factory(pvc_obj)
         pvc_obj.delete()
         logging.info('Start deletion of PVC')
         pvc_obj.ocp.wait_for_delete(pvc_obj.name)
-        if pvc_obj.reclaim_policy == constants.RECLAIM_POLICY_DELETE:
+        if pvc_reclaim_policy == constants.RECLAIM_POLICY_DELETE:
```
vasukulkarni: If its part of object, why should the test fail before? the object is deleted from ocs perspective?

keesturam: Each time reclaim_policy is fetched by running `oc get pvc`, but the pvc is deleted already in this case which is why it fails.

vasukulkarni: This is a python object, I am not sure how the attribute is getting deleted, its just a reference to internal object, the oc get pvc is external and shouldn't matter.

vasukulkarni: I looked at the traceback and linked the issue, I see now how they are linked. Resolving my comment.

vasukulkarni:
Please rerun the same perf tests.

PR #1268, opened 2020-01-07
 Include dc_pod creation in Sanity Helpers

prsurve:
Signed-off-by: prsurve prsurve@redhat.com


tests/manage/z_cluster/nodes/test_nodes_maintenance.py
```diff
@@ -182,7 +182,7 @@ def test_2_nodes_maintenance_same_type(self, nodes_type):
 
     @tier2
     @pytest.mark.polarion_id("OCS-1274")
-    def test_2_nodes_different_types(self, pvc_factory, pod_factory):
+    def test_2_nodes_different_types(self, pvc_factory, pod_factory, dc_pod_factory):
```
ebenahar: wouldn't it be better to replace pod_factory with dc_pod_factory?

prsurve: it would be good we keep both as they act differently from each other

prsurve:
will create new pull request i messed up while rebase

Issue #1272, opened 2020-01-08
We should skip version report for teardown


petr-balogh:
If we fail in deployment of the cluster and it's not properly deployed the teardown job is currently failing cause it's trying to gather version info.

This leads that the teardown job is not successful and we end up with jenkins slave leftover cause of this even the OCP cluster and resources are deleted.

Logs:
```
        log.debug(f"Command output: {r.stdout.decode()}")
        if r.stderr and not r.returncode:
            log.warning(f"Command warning: {mask_secrets(r.stderr.decode(), secrets)}")
        if r.returncode and not ignore_error:
            raise CommandFailed(
>               f"Error during execution of command: {masked_cmd}."
                f"\nError is {mask_secrets(r.stderr.decode(), secrets)}"
            )
E           ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig get clusterversion version -o yaml.
E           Error is Unable to connect to the server: dial tcp: lookup api.jnk-vu1cs33-t1.qe.rh-ocs.com on 10.11.5.19:53: no such host
```

Job:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-destroy-ocs-cluster/2418/console

PR #1273, opened 2020-01-08
Do not report version when teardown

petr-balogh:
Fixes: #1272

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -123,6 +123,10 @@ def log_ocs_version(cluster):
      * ocs_version file in cluster path directory (for copy pasting into bug
        reports)
     """
+    teardown = config.RUN['cli_params']['teardown']
+    if teardown:
+        log.info("Skipping version reporting for teardown.")
+        return
     cluster_version, image_dict = get_ocs_version()
     file_name = os.path.join(
         config.ENV_DATA['cluster_path'],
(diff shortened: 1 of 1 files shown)


petr-balogh:
Verified here:  https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-destroy-ocs-cluster/2422/console

tests/conftest.py
```diff
@@ -123,6 +123,10 @@ def log_ocs_version(cluster):
      * ocs_version file in cluster path directory (for copy pasting into bug
        reports)
     """
+    teardown = config.RUN['cli_params']['teardown']
```
shylesh: @petr-balogh ,
What happens if cli doesn't supply 'teardown' keyword ? keyerror ?

tests/conftest.py
```diff
@@ -123,6 +123,10 @@ def log_ocs_version(cluster):
...
+    if teardown:
```
dahorak: It is not typical usage, but what if the `run-ci` will be executed with both `--deploy` and `--teardown` at once? Then it will not report the versions even though it might be expected.
What about something like:
```python
if teardown and not deploy:
    ...
```

petr-balogh:
Running verification job once more on some other leftover jenkins slave: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-destroy-ocs-cluster/2423/console

petr-balogh:
Verified

Issue #1274, opened 2020-01-08
Recently the OCP extraction takes longer time, we should increase the timeout here.


petr-balogh:
For example here:

https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/3535//console

If we increase timeout here it will also stabilize CI for such cases that it takes sometime more time cause some OCP infra issues.

PR #1275, opened 2020-01-08
Increase timeout for OCP installer/client donwload

petr-balogh:
Fixes: #1274


petr-balogh:
Marking as verified as it's just timeout increase

Issue #1276, opened 2020-01-08
FIO thread is failing because of race


RazTamir:
Same Intermittent Error with getting FIO state.

>           f"Read: {fio_result.get('jobs')[0].get('read').get('iops')}"
        )
E       AttributeError: 'NoneType' object has no attribute 'get'

ocs_ci/ocs/resources/pod.py:564: AttributeError

PR #1277, opened 2020-01-08
Fix FIO issues

RazTamir:
Fixes: https://github.com/red-hat-storage/ocs-ci/issues/1276

### DIFF
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -22,7 +22,7 @@ from ocs_ci.ocs.exceptions import CommandFailed, NonUpgradedImagesFoundError
 from ocs_ci.ocs.utils import setup_ceph_toolbox
 from ocs_ci.ocs.resources.ocs import OCS
 from ocs_ci.utility import templating
-from ocs_ci.utility.utils import run_cmd, TimeoutSampler, check_timeout_reached
+from ocs_ci.utility.utils import run_cmd, check_timeout_reached
 
 logger = logging.getLogger(__name__)
 FIO_TIMEOUT = 600
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -110,14 +110,14 @@ class Pod(OCS):
             Exception: In case of exception from FIO
         """
         try:
-            if self.fio_thread.running():
-                for sample in TimeoutSampler(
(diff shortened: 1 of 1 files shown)


RazTamir:
Verified locally several times

ocs_ci/ocs/resources/pod.py
```diff
@@ -110,14 +110,14 @@ def get_fio_results(self):
...
-            if self.fio_thread.running():
-                for sample in TimeoutSampler(
-                    timeout=FIO_TIMEOUT, sleep=3, func=self.fio_thread.done
-                ):
-                    if sample:
-                        return yaml.safe_load(self.fio_thread.result())
-            if self.fio_thread and self.fio_thread.done():
-                return yaml.safe_load(self.fio_thread.result())
+            for sample in TimeoutSampler(
+                timeout=FIO_TIMEOUT, sleep=3, func=self.fio_thread.result
```
shylesh: @RazTamir ,
According to https://docs.python.org/3/library/concurrent.futures.html , for result() function "If timeout is not specified or None, there is no limit to the wait time." so Technically result() never returns in our case unless job is Done.

Is that what expected here ? Correct me if I am missing something.

RazTamir: That's true but we have the TimeoutSampler for taking care of it

shylesh: According to my understanding , this line https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L732 will not yield until call is finished.

Am I right ?

RazTamir: You probably do.
Let me run some tests and see. 
I will add timeout in case that's true

RazTamir:
Tested multiple times locally - all passed

vasukulkarni:
@RazTamir are you running jenkins test pr as well?

RazTamir:
> @RazTamir are you running jenkins test pr as well?

yes

vasukulkarni:
CI is failing due to tox issue: ocs_ci/ocs/resources/pod.py:25:1: F401 'ocs_ci.utility.utils.TimeoutSampler' imported but unused

RazTamir:
Not sure why Jenkins didn't report execution result but it passed - https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/3805/console

PR #1278, opened 2020-01-08
Fix positional param issue for pvc_factory for session fixture

vasukulkarni:


vasukulkarni:
@petr-balogh @clacroix12 I think we can merge this one, we have to provide upgrade option in test pr as well later on.

vasukulkarni:
Had discussion with Coady, its safe to merge since the signature mismatch is obvious.

PR #1280, opened 2020-01-09
Update multi_pvc_factory_fixture to accept non- default sc

### LINKED ISSUE #1279: test_change_reclaim_policy_of_pv failing
Test test_change_reclaim_policy_of_pv is failing because multi_pvc_factory_fixture is not using the  parameter storageclass which is passed from the test. So the storage class and PV will be always created with reclaim policy Delete. 
AssertionError: Reclaim policy of pvc-086b3f68-9984-4202-80f6-f4400f9c92fe is Delete instead of Retain

jilju:
Fixes #1279 

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1103,10 +1103,9 @@ def multi_pvc_factory_fixture(
             status_tmp = ""
 
         project = project or project_factory()
-        if interface == constants.CEPHBLOCKPOOL:
-            storageclass = helpers.default_storage_class(interface_type=interface)
-        elif interface == constants.CEPHFILESYSTEM:
-            storageclass = helpers.default_storage_class(interface_type=interface)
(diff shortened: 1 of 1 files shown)


jilju:
pvc_factory_fixture also need update

vasukulkarni:
@jilju why are you not fixing the testcase instead fixing it in conftest.py? somewhat similar to https://github.com/red-hat-storage/ocs-ci/blob/master/tests/conftest.py#L1383 , we just finished verifying all tests are working fine except 1 or 2 that needs fix at test case level, same is being done in workloads.

vasukulkarni:
the test also seems pretty old so might be worth updating https://github.com/red-hat-storage/ocs-ci/blob/master/tests/manage/pv_services/test_reclaim_policy.py#L57-L61

Issue #1282, opened 2020-01-09
Assert in get_node_objs() and get_typed_nodes() in case nodes objects not found


ebenahar:
Will improve catching these situations in case node objects are not found for some reason.

See an example for failure without this fix:

self = <tests.manage.z_cluster.nodes.test_nodes_maintenance.TestNodesMaintenance object at 0x7fdc80ad0550>
node_type = 'worker'
pvc_factory = <function pvc_factory_fixture.<locals>.factory at 0x7fdc802848c8>
pod_factory = <function pod_factory_fixture.<locals>.factory at 0x7fdc80284b70>

    @tier1
    @pytest.mark.parametrize(
        argnames=["node_type"],
        argvalues=[
            pytest.param(*['worker'], marks=pytest.mark.polarion_id("OCS-1269")),
            pytest.param(*['master'], marks=pytest.mark.polarion_id("OCS-1272"))
        ]
    )
    def test_node_maintenance(self, node_type, pvc_factory, pod_factory):
        """
        OCS-1269/OCS-1272:
        - Maintenance (mark as unscheduable and drain) 1 worker/master node
        - Check cluster functionality by creating resources
          (pools, storageclasses, PVCs, pods - both CephFS and RBD)
        - Mark the node as scheduable
        - Check cluster and Ceph health
    
        """
        # Get a list of 2 nodes. Pick one of them after checking
        # which one does't have the rook operator running on
>       typed_nodes = get_typed_nodes(node_type=node_type, num_of_nodes=2)

tests/manage/z_cluster/nodes/test_nodes_maintenance.py:73: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ocs_ci/ocs/node.py:54: in get_typed_nodes
    n for n in nodes if node_type in n.get().get('metadata')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

.0 = <list_iterator object at 0x7fdc81293470>

        n for n in nodes if node_type in n.get().get('metadata')
>       .get('annotations').get('machineconfiguration.openshift.io/currentConfig')
    ]
E   TypeError: argument of type 'NoneType' is not iterable

ocs_ci/ocs/node.py:55: TypeError

[...]

PR #1283, opened 2020-01-09
Assert in case node objects are not found and few fixes in drain tests

ebenahar:
Addresses https://github.com/red-hat-storage/ocs-ci/issues/1282

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -30,13 +30,15 @@ def get_node_objs(node_names=None):
     nodes_obj = OCP(kind='node')
     node_dicts = nodes_obj.get()['items']
     if not node_names:
-        return [OCS(**node_obj) for node_obj in node_dicts]
+        nodes = [OCS(**node_obj) for node_obj in node_dicts]
     else:
-        return [
+        nodes = [
             OCS(**node_obj) for node_obj in node_dicts if (
                 node_obj.get('metadata').get('name') in node_names
             )
         ]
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/node.py
```diff
@@ -26,13 +26,17 @@ def get_node_objs(node_names=None):
...
+        return nodes
     else:
-        return [
+        nodes = [
             OCS(**node_obj) for node_obj in node_dicts if (
                 node_obj.get('metadata').get('name') in node_names
             )
         ]
+        assert nodes, "Failed to get the nodes OCS objects"
+        return nodes
```
sidhant-agrawal: These lines can be moved outside if-else and then L30 & L31 can be removed.

ocs_ci/ocs/node.py
```diff
@@ -49,7 +53,7 @@ def get_typed_nodes(node_type='worker', num_of_nodes=None):
 
     """
     nodes = get_node_objs()
-
+    assert nodes, "Failed to get the nodes OCS objects"
```
sidhant-agrawal: As we are already adding assert in `get_node_objs()` itself, do we still need this one?

ocs_ci/ocs/node.py
ebenahar: good catch

ocs_ci/ocs/node.py
ebenahar: no, again, good catch

PR #1287, opened 2020-01-10
Check whether RHEL nodes are accepting ssh connections

shylesh:

### DIFF
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -534,7 +534,7 @@ class AWSUPI(AWSBase):
                 InstanceId=inst_id,
             )
 
-    @retry(exceptions.CommandFailed, tries=10, delay=60, backoff=2)
+    @retry(exceptions.CommandFailed, tries=15, delay=30, backoff=1)
     def check_connection(self, rhel_pod_obj, host, pem_dst_path):
         cmd = 'ls'
         rhel_pod_obj.exec_cmd_on_node(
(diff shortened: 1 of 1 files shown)


shylesh:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/3651/consoleFull

Tested here

ocs_ci/deployment/aws.py
```diff
@@ -534,6 +534,13 @@ def create_rhel_instance(self):
                 InstanceId=inst_id,
             )
 
+    @retry(exceptions.CommandFailed, tries=10, delay=60, backoff=2)
```
petr-balogh: Do we really need to backoff=2? If I understand this parameter correctly it will double the delay each attempt. So if something will happen you will wait for a lot of time to fail.
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/retry.py#L16

I think that reasonable time to wait and retry will be about 5 max 10 mins WDYT?
so
```
@retry(exceptions.CommandFailed, tries=15, delay=30, backoff=1)
```
Sounds more reasonable to me

petr-balogh:
@shylesh next time if you will create PR please to the commit message or PR message you can write:

Fixes: #1288

So it's better to have link from PR to the issue and it automatically closes the issue once this one is closed.

ocs_ci/deployment/aws.py
shylesh: Agree

shylesh: Done

Issue #1303, opened 2020-01-14
Enable live testing in ocs-ci


petr-balogh:
We need to skip creation of catalog source when testing from live.
Let's see if something else needs to be done.

PR #1304, opened 2020-01-14
Live install

petr-balogh:
Fixes: #1303

### DIFF
--- a/conf/ocsci/live_deployment.yaml
+++ b/conf/ocsci/live_deployment.yaml
@@ -1,0 +1,5 @@
+---
+# This config file can be used for live deployment
+DEPLOYMENT:
+  live_deployment: True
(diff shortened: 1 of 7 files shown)


vasukulkarni:
Awesome!

petr-balogh:
> Awesome!

@vasukulkarni 
Triggered PR verification job just with the passed config file. Let's see if it will finally work at least for live deployment now ;)

If it will work we can update jenkins job to have checkbox LIVE_DEPLOYMENT and in case it will be selected we will use cli parameter.

Hope it will, then we will need to finish with other issue to allow us to deploy for internal builds.

petr-balogh:
For documentation part I opened this issue: https://github.com/red-hat-storage/ocs-ci/issues/1306

PR #1308, opened 2020-01-15
Reduce wait time between iterations to 10 secs

mykaul:
Generally, we should never wait more than 10 seconds per iteration for anything and if needed, iterate more times.

Signed-off-by: Yaniv Kaul ykaul@redhat.com

### DIFF
--- a/ocs_ci/ocs/registry.py
+++ b/ocs_ci/ocs/registry.py
@@ -58,8 +58,8 @@ def get_registry_pod_obj():
     # Sometimes when there is a update in config crd, there will be 2 registry pods
     # i.e. old pod will be terminated and new pod will be up based on new crd
     # so below loop waits till old pod terminates
-    wait_time = 30
-    for iteration in range(10):
+    wait_time = 10
+    for iteration in range(30):
         pod_data = pod.get_pods_having_label(
             label='docker-registry=default', namespace=constants.OPENSHIFT_IMAGE_REGISTRY_NAMESPACE
         )
@@ -68,10 +68,10 @@ def get_registry_pod_obj():
             break
         elif len(pod_obj) == 0:
             raise UnexpectedBehaviour("Image-registry pod not present")
-        elif iteration > 5:
+        elif iteration == 18:  # 18*10 = 180 secs.
             raise UnexpectedBehaviour("Waited for 3 mins Image-registry pod is not in Running state")
         else:
-            logger.info(f"Waiting for 30 sec's for registry pod to be up iteration {iteration}")
+            logger.info(f"Waiting for {wait_time} seconds for registry pod to be up - iteration {iteration}")
             time.sleep(wait_time)
     return pod_obj
 


ocs_ci/ocs/registry.py
```diff
@@ -68,10 +68,10 @@ def get_registry_pod_obj():
             break
         elif len(pod_obj) == 0:
             raise UnexpectedBehaviour("Image-registry pod not present")
-        elif iteration > 5:
+        elif iteration == 18: # 18*10 = 180 secs.
```
petr-balogh: Please fix flake8 issue: ocs_ci/ocs/registry.py:71:30: E261 at least two spaces before inline comment

mykaul:
> You need to sign commit @mykaul . Please do `git commit --amend -s`

Done.

ocs_ci/ocs/registry.py
```diff
@@ -68,10 +68,10 @@ def get_registry_pod_obj():
             break
         elif len(pod_obj) == 0:
             raise UnexpectedBehaviour("Image-registry pod not present")
-        elif iteration > 5:
+        elif iteration == 18:  # 18*10 = 180 secs.
             raise UnexpectedBehaviour("Waited for 3 mins Image-registry pod is not in Running state")
         else:
-            logger.info(f"Waiting for 30 sec's for registry pod to be up iteration {iteration}")
+            logger.info(f"Waiting for 10 sec's for registry pod to be up iteration {iteration}")
```
clacroix12:
```suggestion
            logger.info(f"Waiting for {wait_time} sec's for registry pod to be up iteration {iteration}")
```

ocs_ci/ocs/registry.py
```diff
@@ -68,10 +68,10 @@ def get_registry_pod_obj():
             break
         elif len(pod_obj) == 0:
             raise UnexpectedBehaviour("Image-registry pod not present")
-        elif iteration > 5:
+        elif iteration == 18:  # 18*10 = 180 secs.
```
clacroix12: Why are we looping over `range(30)` if we stop at 18?

ocs_ci/ocs/registry.py
mykaul: Sure, I can change that. I was just going what whatever was there before.

ocs_ci/ocs/registry.py
mykaul: No idea. Again, followed previous code. Should I change that?

clacroix12: It appears that the logic here is simply waiting for the pod_obj list to be of length 1 within 180 seconds checking 10 times. We have a retry decorator that we can use to simplify this entire chunk of code.

Something like this perhaps:

```python
@retry(UnexpectedBehaviour, tries=18, delay=10, backoff=1)
def get_registry_pod_obj():
    pod_data = pod.get_pods_having_label(
        label='docker-registry=default',
        namespace=constants.OPENSHIFT_IMAGE_REGISTRY_NAMESPACE
    )
    pod_obj = [pod.Pod(**data) for data in pod_data]
    if len(pod_obj) != 1:
        raise UnexpectedBehaviour("Old registry pod not fully terminated")
    return pod_obj
```

The only difference here is that the original code will raise the `UnexpectedBehavior` if the `pod_obj` length is 0 immediately. In my code here, it will simply be caught by the retry decorator until all attempts are tried. If we want to maintain the functionality of raising the error immediately when the length is 0 we can simply raise a slightly different error that is not caught by the retry decorator.

I'm not saying we NEED to change this all right now, I just wanted to call out that this logic isn't all that clear at first glance and could be made much more simple with tools we already have in our utility module.

mykaul: All I wanted is to reduce the wait time. I think your suggestions are good, but should not hold this PR?

clacroix12: Sure, that's fine. But you are changing the amount of times this loop is iterating and that count doesn't match the logic that stops the loop at 18. I think this is the right time to at least address that and figure out a way to make it consistent.

mykaul: I was aiming NOT to change the amount of time, but the amount of iterations and reduce the amount of time between iterations... Did I make a mistake here?

clacroix12: You didn't since the logic was already weird to begin with. I'll just create an issue and handle it another time.

ocs_ci/ocs/registry.py
```diff
@@ -68,10 +68,10 @@ def get_registry_pod_obj():
...
             raise UnexpectedBehaviour("Waited for 3 mins Image-registry pod is not in Running state")
         else:
-            logger.info(f"Waiting for 30 sec's for registry pod to be up iteration {iteration}")
+            logger.info(f"Waiting for {wait_time} seconds for registry pod to be up - iteration {iteration}")
```
petr-balogh: This line is pretty long as I can see it's breaking in other line in github UI view. Can you please split like:
```python
logger.info(
    f"Waiting for {wait_time} seconds for registry pod to be up - iteration {iteration}
")
```

clacroix12:
@petr-balogh yes I did: https://github.com/red-hat-storage/ocs-ci/issues/1326

vasukulkarni:
This also has to go through jenkins testing, not sure how Yaniv can run.

ocs_ci/ocs/registry.py
mykaul: <rant>
Really hate it when I need to break lines because of *Github* UI
</rant>

mykaul:
> @mykaul I see that DCO is still complaining about your signed off commit.

No idea why - please see https://github.com/red-hat-storage/ocs-ci/pull/1308/commits

> 
> This actually block the PR to be merged.
> 
> Not sure what can be wrong but can you try to run both command mentioned here for sure:
> https://github.com/red-hat-storage/ocs-ci/pull/1308/checks?check_run_id=391802831
> 
> Otherwise LGTM, BTW @clacroix12, have you opened mentioned issue? If so, please link it here. Thanks

ocs_ci/ocs/registry.py
petr-balogh: It's not just cause of github UI but also cause of our conventions described here:
https://ocs-ci.readthedocs.io/en/latest/docs/coding_guidelines.html
*Line length should be maximum 79 characters!*

I know it's not 100% followed and we do not prohibit it pep8 check in the tox as some people just not believe in 79 chars. So I am trying to at least keep eyes when I see it's too long and I see it overflows in UI.

mykaul: flake8 did not complain on it. I don't mind changing it of course. I really just wanted to change the timeout...

petr-balogh: I triggered verification job with tier1 in our jenkins so it will update status here how it passed to be sure we are OK.

Yes, please change. I also sent you invitation to red-hat-storage organization on github, you will need to accept from email most likely. But that's probably not the case of complaining DCO.

Did you have such issue also in different projects or just this one PR?

Maybe if you will do the change in the commit and commit --amend your changes and for sure once more:

git commit --amend --signoff
Now your commits will have your sign off. Next run

git push --force-with-lease origin mykaul-registry_iterate
Commit sha: fad6a04, Author: Yaniv Kaul, Committer: Yaniv Kaul; The sign-off is missing.

Maybe DCO will be satisfied.

petr-balogh:
Based on verification job it's seems passed well, only one failure which was also in last tier1 execution when I checked. Marking as verified.

mykaul:
Apologies, I've moved over and won't be able to rebase right now.

ebenahar:
@mykaul Would like to close old PRs, is this one still relevant?

ebenahar:
I see this change is not very relevant anymore as the code has changed significantly. Closing

Issue #1315, opened 2020-01-15
Check Ceph health before each test case


ebenahar:
Create fixture for this and in case this will fail we should skip the test case.
We shouldn't have this for deployment.

vasukulkarni:
Right now should be only for tier4 tests

PR #1318, opened 2020-01-15
Add health check for tier4 tests before they are run

vasukulkarni:
Fixes: https://github.com/red-hat-storage/ocs-ci/issues/1315

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -1188,6 +1188,17 @@ def get_az_count():
 @retry((CephHealthException, CommandFailed), tries=20, delay=30, backoff=1)
 def ceph_health_check(namespace=None):
     """
+    Args:
+        namespace (str): Namespace of OCS
+            (default: config.ENV_DATA['cluster_namespace'])
+
+    Returns: ceph_health_check_base with default retries of 20, delay of 30 seconds
+    """
+    return ceph_health_check_base(namespace)
+
+
+def ceph_health_check_base(namespace=None):
+    """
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -798,6 +799,22 @@ def polarion_testsuite_properties(record_testsuite_property, pytestconfig):
...
+    tier4_marks = [m.mark for m in [tier4a, tier4b, tier4c]]
+    for mark in node.iter_markers():
+        if mark in tier4_marks:
+            log.info("Checking for Ceph Health OK ")
+            status = ceph_health_check()
+            if status:
+                log.info("Health check passed")
+                return
+            else:
+                pytest.skip("Ceph Health check failed")
```
vavuthu: why would we need to skip the test if ceph health is not OK. I believe we need to raise assert here.

ocs_ci/utility/utils.py
```diff
@@ -1186,7 +1186,14 @@ def get_az_count():
 
 
 @retry((CephHealthException, CommandFailed), tries=20, delay=30, backoff=1)
-def ceph_health_check(namespace=None):
+def ceph_health_check(namespace=None, tries=20, delay=30, backoff=1):
```
clacroix12: The arguments `tries`, `delay`, and `backoff` aren't used, can we remove them? We will also want to add `namespace` to the docstring.

vasukulkarni: I will fix both

vasukulkarni:
Working as expected: 2 failures (+2) , 48 skipped (+48)

Issue #988, opened 2019-11-12
measure_pvc_creation_time Needs enahancement


ramkiperiy:
measure_pvc_creation_time function is failing, if we try to measure time for 500+ pvc's.
Either this function needs enhancement or need a new function in-case of more than 500+ PVC's

Akarsha-rai:
And also we need to update the code to pick the right provisioner plugin. 
We have two provisioner plugins out of which one is active and other act as idle.

In code we are selecting only the first plugin:
```
ceph_pod = Pod(**rbd_provision_pod_items[0])
    return ceph_pod
```

gklein:
> measure_pvc_creation_time function is failing, if we try to measure time for 500+ pvc's.
> Either this function needs enhancement or need a new function in-case of more than 500+ PVC's

As a best practice, I think we need to reuse the same function in order to eliminate code duplication. It would be better if we can fix it or add extra parameters to function in order to support this "special case".

Could you please share some more details about the failure. What's the exception you are receiving when running with 500+ PVCs?

ramkiperiy:
Closing this Issue, since code is merged.

PR #1320, opened 2020-01-16
Added new function to get pvc creation & deletion time for pvc_list

ramkiperiy:
Added new function to get pvc creation & deletion time for pvc_list
#988

Signed-off-by: rperiyas <rperiyas@redhat.com>

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -949,6 +949,86 @@ def measure_pvc_creation_time(interface, pvc_name):
     return total.total_seconds()
 
 
+def measure_pvc_creation_time_bulk(interface, pvc_obj_list):
+    """
+    Measure PVC creation time of bulk PVC based on logs.
+
+    Args:
+        interface (str): The interface backed the PVC
+        pvc_obj_list (list): List of PVC Names for measuring creation time
+
+    Returns:
+        pvc_dict (dict): Dictionary of pvc_name with creation time.
+
+    """
+    # Get the correct provisioner pod based on the interface
+    pod_name = pod.get_csi_provisioner_pod(interface)
+    # get the logs from the csi-provisioner containers
+    logs = pod.get_pod_logs(pod_name[0], 'csi-provisioner')
+    logs += pod.get_pod_logs(pod_name[1], 'csi-provisioner')
+
+    pvc_dict = dict()
+    format = '%H:%M:%S.%f'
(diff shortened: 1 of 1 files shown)


tests/helpers.py
```diff
@@ -949,6 +949,86 @@ def measure_pvc_creation_time(interface, pvc_name):
...
+    # Get the correct provisioner pod based on the interface
+    pod_name = pod.get_csi_provisioner_pod(interface)
+    # get the logs from the csi-provisioner containers
+    logs = pod.get_pod_logs(pod_name[0], 'csi-provisioner')
+    logs += pod.get_pod_logs(pod_name[1], 'csi-provisioner')
+
+    pvc_dict = dict()
+    format = '%H:%M:%S.%f'
+    for pvc_obj in pvc_obj_list:
+        logs = logs.split("\n")
```
keesturam: can you move this outside for loop? We don't want to split the static log file every time.

tests/helpers.py
```diff
@@ -949,6 +949,86 @@ def measure_pvc_creation_time(interface, pvc_name):
     return total.total_seconds()
 
 
+def measure_pvc_creation_time_bulk(interface, pvc_obj_list):
+    """
+    Measure PVC creation time of bulk PVC based on logs.
+
+    Args:
+        interface (str): The interface backed the PVC
+        pvc_obj_list (list): List of PVC Names for measuring creation time
```
keesturam: `pvc_obj_list` might be misleading as they are not objects and just names. can you change the name of this parameter?

ramkiperiy: It's pvc_obj_list only, if we expect pvc_name_list then while passing need to get name of each pvc object and store in list. I will change the args description, in next commit.

tests/helpers.py
ramkiperiy: ack

tests/helpers.py
ramkiperiy: changed it back to pvc_name_list, resolving comment.

ramkiperiy:
Local run logs for measuring creation and deletion of all 4 types of PVCs http://pastebin.test.redhat.com/829431
Local run logs for measuring creation and deletion time of 500 PVCs is in attachment 
[logs.txt](https://github.com/red-hat-storage/ocs-ci/files/4095966/logs.txt)

PR #1330, opened 2020-01-16
More fixes for avoiding conflict with live versions

zmc:


zmc:
I've tested this in our Jenkins, and it fixed the issue I was hitting. The issue was that ocs-ci was looking for a csv called ocs-operator.v4.2.0 when it should have been looking for ocs-operator.v0.0.287 .

PR #1331, opened 2020-01-17
Include dc_pod creation in Sanity Helpers

prsurve:
Signed-off-by: prsurve prsurve@redhat.com


tests/sanity_helpers.py
```diff
@@ -36,13 +40,14 @@ def health_check(self):
         )
         self.ceph_cluster.cluster_health_check(timeout=60)
 
-    def create_resources(self, pvc_factory, pod_factory, run_io=True):
+    def create_resources(self, pvc_factory, pod_factory, dc_pod_factory, run_io=True):
```
ebenahar: why not replacing pod with DC?

prsurve: It would be good if we keep both as these 2 act's differently with their behavior

tests/sanity_helpers.py
```diff
@@ -52,23 +57,42 @@ def create_resources(self, pvc_factory, pod_factory, run_io=True):
...
+            for dc_pod in self.dc_pod_objs:
+                get_fio_rw_iops(dc_pod)
 
-    def delete_resources(self):
+    def delete_resources(self, ignore_delete=False):
         """
         Sanity validation - Delete resources (FS and RBD)
 
+        Args:
+            ignore_delete (bool): skip deletion of pod if deleted already
```
fbalak: Why is this needed? Wouldn't it be better to just check if pod exists before deleting it? With implementation from this PR, there can be missed some unexpected errors of pod deletion that might not be intended behaviour as this option ignores all command errors for all pod objects created in `Sanity` class.

prsurve: Reason of this param is when doing node replacement some nginx pod will get deleted so at script teardown fails for deleting pod obj and script will never get a pass

fbalak: Wouldn't it be better to just check that the pod exists and if not then don't initiate delete? It would reduce a possibility that we will miss an error that we don't want to miss during deleting all pods.

prsurve: Okay this can be done will update in next commit

PR #1346, opened 2020-01-20
Verify osd tree output and ensure it have device set PVC names

jilju:
Updated ocs_install_verification function to verify ceph osd tree output.

ceph osd tree output structure and values are verified.
Using JSON Schema to validate the output.
Presence of device set PVC names is also verified along with this. (Verifies bug https://bugzilla.redhat.com/show_bug.cgi?id=1765154)
Implemented in this way because there are more checks to be done
in the output. Added as TODO.

### DIFF
--- a/ocs_ci/ocs/resources/ocs.py
+++ b/ocs_ci/ocs/resources/ocs.py
@@ -383,15 +383,3 @@ def ocs_install_verification(timeout=600, skip_osd_distribution_check=False):
             f"ceph osd tree output does not contain pvc name {pvc}"
         )
     log.info("Verified device set PVC names in ceph osd tree output.")
-
-    # Verify OSDs are listed as sdd
-    log.info("Verify OSDs are listed as sdd in ceph osd tree output.")
-    for num in range(len(deviceset_pvcs)):
-        re_exp = f'ssd [0-9]*.[0-9]*[ ]*osd.{num}'
-        match = re.search(re_exp, osd_tree)
-        log.info(match)
-        assert match, (
-            f"One or more OSDs are not listed as sdd in ceph osd tree output."
-            f" No match for expression {re_exp}"
-        )
-    log.info("Verified OSDs are listed as sdd in ceph osd tree output.")
--- a/ocs_ci/ocs/resources/ocs.py
+++ b/ocs_ci/ocs/resources/ocs.py
@@ -373,6 +373,6 @@ def ocs_install_verification(timeout=600, skip_osd_distribution_check=False):
     osd_tree = ct_pod.exec_ceph_cmd(ceph_cmd='ceph osd tree', format='')
     for pvc in deviceset_pvcs:
         assert f'host {pvc.name}' in osd_tree, (
-            f"ceph osd tree output does not contain pvc name {pvc}"
+            f"ceph osd tree output does not contain pvc name {pvc.name}"
         )
     log.info("Verified device set PVC names in ceph osd tree output.")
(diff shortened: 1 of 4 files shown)


ocs_ci/ocs/resources/ocs.py
```diff
@@ -349,3 +352,32 @@ def ocs_install_verification(timeout=600, skip_osd_distribution_check=False):
     assert {defaults.CEPHFS_PROVISIONER, defaults.RBD_PROVISIONER} == (
         {item['metadata']['name'] for item in csi_driver.get()['items']}
     )
+
+    # Verify ceph osd tree have device set PVC names specified
+    log.info("Checking for device set PVC names in ceph osd tree output.")
+    deviceset_pvcs = [
```
ebenahar: please use ocs_ci.ocs.resources.pvc.get_deviceset_pvs

ocs_ci/ocs/resources/ocs.py
```diff
@@ -349,3 +352,32 @@ def ocs_install_verification(timeout=600, skip_osd_distribution_check=False):
...
+    for pvc in deviceset_pvcs:
+        assert f'host {pvc}' in osd_tree, (
+            f"ceph osd tree output does not contain pvc name {pvc}"
+        )
+    log.info("Verified device set PVC names in ceph osd tree output.")
+
+    # Verify OSDs are listed as sdd
+    log.info("Verify OSDs are listed as sdd in ceph osd tree output.")
+    for num in range(len(deviceset_pvcs)):
+        re_exp = f'ssd [0-9]*.[0-9]*[ ]*osd.{num}'
```
vasukulkarni: this is dependent on hardware we run and can easily fail if we move out of aws or different systems?

jilju:
circular dependencies to be resolved

ocs_ci/ocs/resources/ocs.py
jilju: get_deviceset_pvs is giving list of PV. Names of PVCs are required here.

ocs_ci/ocs/resources/ocs.py
jilju: This part is removed now. It will be added in another PR after getting this clarity.

jilju:
> circular dependencies to be resolved

resolved

ocs_ci/ocs/resources/ocs.py
ebenahar: got you. Maybe would be good to create a function for this implementation in pvc.py

vasukulkarni:
I see couple for changes and its worth testing again.

ocs_ci/ocs/resources/ocs.py
jilju: Added new function get_deviceset_pvcs in pvc.py

ocs_ci/ocs/resources/ocs.py
```diff
@@ -363,3 +366,32 @@ def ocs_install_verification(timeout=600, skip_osd_distribution_check=False):
...
+    for item in osd_tree['nodes']:
+        validate(instance=item, schema=schemas[item['type']])
+        if item['type'] == 'host':
+            deviceset_pvcs.remove(item['name'])
+    log.info(
+        "Verified ceph osd tree output. Device set PVC names are given in the "
+        "output."
+    )
+
+    # TODO: Verify ceph osd tree output have osd listed as ssd
```
jilju: Verifies bug https://bugzilla.redhat.com/show_bug.cgi?id=1770910

PR #1367, opened 2020-01-22
Fix in get_typed_nodes()

ebenahar:

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -55,7 +55,6 @@ def get_typed_nodes(node_type='worker', num_of_nodes=None):
 
     """
     nodes = get_node_objs()
-    from pdb import set_trace; set_trace()
     typed_nodes = [
         node for node in nodes if node
         .ocp.get_resource(resource_name=node.name, column='ROLES') == node_type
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/node.py
```diff
@@ -55,9 +55,10 @@ def get_typed_nodes(node_type='worker', num_of_nodes=None):
 
     """
     nodes = get_node_objs()
+    from pdb import set_trace; set_trace()
```
prsurve: do we need set_trace here? it will need user input to proceed

ebenahar: no :)

ocs_ci/ocs/node.py
```diff
@@ -56,8 +56,8 @@ def get_typed_nodes(node_type='worker', num_of_nodes=None):
     """
     nodes = get_node_objs()
```
MeridianExplorer: I believe we can skip the assignment and use get_nodes_objs()` inline in the list comprehension below

MeridianExplorer:
LGTM, it's great that we're improving readability of the code, very important in my opinion

ocs_ci/ocs/node.py
ebenahar: fixed

Issue #1368, opened 2020-01-22
Add marker for vdbench tests


vasukulkarni:

vasukulkarni:
Also please mark one of the test as workload so that we can find any issues sooner.

Avilir:
Closed by merging : https://github.com/red-hat-storage/ocs-ci/pull/1372

PR #1372, opened 2020-01-23
Adding another test for the vdbench according to issue #1368

Avilir:
This is to add another test with "@workloads" marker.
This additional test will run for only one hour.


Avilir:
Was tested locally.

vasukulkarni:
I think in jenkins you have to select the correct cluster config for it to work.

Issue #1352, opened 2020-01-21
We should fail the destroy on vmware when CredentialFile is not passed properly


petr-balogh:
Job:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-destroy-ocs-cluster/2742/console

Pytest passed when failed in destroy of VmWare cluster cause of forgotten credential file passed to the job.  Then it started destroying the jenkins slave but the cluster still exist.

```

22:34:34 - MainThread - ocs_ci.deployment.ocp - INFO - Destroying the cluster
22:34:34 - MainThread - ocs_ci.deployment.ocp - INFO - Destroying cluster defined in /home/jenkins/current-cluster-dir/openshift-cluster-dir
22:34:34 - MainThread - ocs_ci.utility.utils - INFO - Executing command: /home/jenkins/bin/openshift-install destroy cluster --dir /home/jenkins/current-cluster-dir/openshift-cluster-dir --log-level INFO
22:34:34 - MainThread - ocs_ci.deployment.ocp - ERROR - Traceback (most recent call last):
  File "/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/ocs_ci/deployment/ocp.py", line 191, in destroy
    run_cmd(destroy_cmd)
  File "/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/ocs_ci/utility/utils.py", line 418, in run_cmd
    f"Error during execution of command: {masked_cmd}."
ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: /home/jenkins/bin/openshift-install destroy cluster --dir /home/jenkins/current-cluster-dir/openshift-cluster-dir --log-level INFO.
Error is level=fatal msg="Failed while preparing to destroy cluster: no platform configured in metadata"


22:34:34 - MainThread - botocore.credentials - INFO - Found credentials in shared credentials file: ~/.aws/credentials


=============================== warnings summary ===============================
.venv/lib/python3.6/site-packages/_pytest/junitxml.py:436
  /home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/.venv/lib/python3.6/site-packages/_pytest/junitxml.py:436: PytestDeprecationWarning: The 'junit_family' default value will change to 'xunit2' in pytest 6.0.
  Add 'junit_family=xunit1' to your pytest.ini file to keep the current format in future versions of pytest and silence this warning.
[...]

PR #1396, opened 2020-01-28
do not ignore failed cluster destroy command

dahorak:
* (re)raise exception of failed openshift-install destroy cluster command
* fixes: https://github.com/red-hat-storage/ocs-ci/issues/1352

### DIFF
--- a/ocs_ci/deployment/ocp.py
+++ b/ocs_ci/deployment/ocp.py
@@ -9,6 +9,7 @@ import traceback
 import pytest
 import yaml
 
+from ocs_ci.ocs.exceptions import CommandFailed
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.openshift_ops import OCP
@@ -190,5 +191,8 @@ class OCPDeployment:
             logger.info(f"Destroying cluster defined in {self.cluster_path}")
             run_cmd(destroy_cmd)
 
+        except CommandFailed:
+            logger.error(traceback.format_exc())
+            raise
         except Exception:
             logger.error(traceback.format_exc())


ocs_ci/deployment/ocp.py
```diff
@@ -190,5 +191,8 @@ def destroy(self, log_level="DEBUG"):
             logger.info(f"Destroying cluster defined in {self.cluster_path}")
             run_cmd(destroy_cmd)
 
+        except CommandFailed:
```
clacroix12: Is the intention to "ignore" all other types of Exceptions and just log them? If we want to raise those as well should we just add the `raise` to the `except Exception` block (line 197)?

dahorak: Initially, the behaviour was exactly like that - ignore all exceptions and just log them, so I don't want to change it too much (as https://github.com/red-hat-storage/ocs-ci/issues/1352 was the only issue we faced till now).

I'm aware of `FileNotFoundError` exception raised from line 186, when `metadata.json` file does not exist and this exception is expected and we want to catch it without failing the destroy process (this case usually means, that the cluster wasn't installed at all, so there is nothing to destroy - but we want to clean up other auxiliary resources - e.g. jenkins slave).

I agree, that it might be safer to explicitly catch only that *known/expected* exception and re-raise any other - but this proposed change is closer to the previous behaviour (but I don't have strong opinion about this, so if you think that it would be better to change it, I can do that).

clacroix12: I agree with your take on it. If this is the only issue we have encountered let's simply address that without changing the logic here drastically.

PR #1403, opened 2020-01-29
pre-pull container images on all nodes during pod creation

### LINKED ISSUE #1404: pre-pull container images before pod creation
pod creation fails due to time out occasionally, so it is better to pre-pull the images before starting the pod.

keesturam:
Fixes #1404

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -133,13 +133,26 @@ def create_pod(
         AssertionError: In case of any failure
     """
     if interface_type == constants.CEPHBLOCKPOOL:
-        pod_dict = pod_dict_path if pod_dict_path else constants.CSI_RBD_POD_YAML
         interface = constants.RBD_INTERFACE
+        if pod_dict_path:
+            pod_dict = pod_dict_path
+        else:
+            pod_dict = constants.CSI_RBD_POD_YAML
+            pull_images('nginx')
     else:
-        pod_dict = pod_dict_path if pod_dict_path else constants.CSI_CEPHFS_POD_YAML
         interface = constants.CEPHFS_INTERFACE
+        if pod_dict_path:
+            pod_dict = pod_dict_path
+        else:
+            pod_dict = constants.CSI_CEPHFS_POD_YAML
+            pull_images('nginx')
+
     if dc_deployment:
-        pod_dict = pod_dict_path if pod_dict_path else constants.FEDORA_DC_YAML
+        if pod_dict_path:
+            pod_dict = pod_dict_path
+        else:
+            pod_dict = constants.FEDORA_DC_YAML
+            pull_images('fedora')
     pod_data = templating.load_yaml(pod_dict)
     if not pod_name:
         pod_name = create_unique_resource_name(
@@ -708,6 +721,25 @@ def get_cephfs_name():
     )
     result = cfs_obj.get()
     return result['items'][0].get('metadata').get('name')
+
+
+def pull_images(image_name):
+    """
+    Function to pull images on all nodes
+
+    Args:
+        image_name (str): Name of the container image to be pulled
+
+    Returns: None
+
+    """
+
+    node_objs = node.get_node_objs(get_worker_nodes())
+    for node_obj in node_objs:
+        logging.info(f'pulling image "{image_name}  " on node {node_obj.name}')
+        assert node_obj.ocp.exec_oc_debug_cmd(
+            node_obj.name, cmd_list=[f'podman pull {image_name}']
+        )
 
 
 def run_io_with_rados_bench(**kw):


tests/helpers.py
```diff
@@ -710,6 +723,25 @@ def get_cephfs_name():
...
+    Function to pull images on all nodes
+    Args:
+        image_name (str): Name of the container image to be pulled
+
+    Returns:
+        bool: True if deletion of image pull is successful
+    """
+    nodes = get_worker_nodes()
+    for worker_node in nodes:
+        node_obj = ocp.OCP(kind=constants.NODE)
```
Shrivaibavi: Can we use get_node_objs() function here ?

tests/helpers.py
```diff
@@ -710,6 +723,25 @@ def get_cephfs_name():
     return result['items'][0].get('metadata').get('name')
 
 
+def pull_images(image_name):
+    """
+    Function to pull images on all nodes
+    Args:
+        image_name (str): Name of the container image to be pulled
+
```
Shrivaibavi: Please remove this empty line

keesturam: Done

tests/helpers.py
keesturam: Thanks for the pointer. I've used `get_node_objs` now.

tests/helpers.py
```diff
@@ -710,6 +723,24 @@ def get_cephfs_name():
...
+    Returns:
+        bool: True if deletion of image pull is successful
+    """
+
+    node_objs = node.get_node_objs(get_worker_nodes())
+    for node_obj in node_objs:
+        logging.info(f'pulling image "{image_name}" on node {node_obj.name}')
+        assert node_obj.ocp.exec_oc_debug_cmd(
+            node_obj.name, cmd_list=[f'podman pull {image_name}']
+        )
```
Shrivaibavi: I actually meant this way, node_objs = get_typed_node(nodetype="worker"), Never mind your implementation also looks good.

tests/helpers.py
```diff
@@ -133,13 +133,26 @@ def create_pod(
...
+            pod_dict = constants.CSI_CEPHFS_POD_YAML
+            pull_images('nginx')
+
     if dc_deployment:
-        pod_dict = pod_dict_path if pod_dict_path else constants.FEDORA_DC_YAML
+        if pod_dict_path:
+            pod_dict = pod_dict_path
+        else:
+            pod_dict = constants.FEDORA_DC_YAML
+            pull_images('fedora')
```
clacroix12: We have a little bit of duplication here that we can reduce. It seems like in every conditional if we specify `pod_dict_path` then that takes priority. If this is the case we can specify logic first and then if it isn't set we can check our other conditionals:

```
if pod_dict_path:
    pod_dict = pod_dict_path
else:
    if interface_type == constants.CEPHBLOCKPOOL:
        interface = constants.RBD_INTERFACE
        pod_dict = constants.CSI_RBD_POD_YAML
        pull_images('nginx')
    else:
        interface_type = constants.CEPHFS_INTERFACE
        pod_dict = constants.CSI_CEPHFS_POD_YAML
        pull_images('nginx')
    if dc_deployment:
        pod_dict = constants.FEDORA_DC_YAML
        pull_images('fedora')

```

One thing to note here is that I kept the same pattern of if/else/if. I'm not sure if this is intentional or not but if we  have an `interface_type` of `constants.CEPHBLOCKPOOL` and `dc_deployment` is true, we will overwrite the values for `CEPHBLOCKPOOL` and instead use the ones for `dc_deployment` since it is not an `elif`. Again, I'm not sure if this is intentional so I left it as is.

tests/helpers.py
```diff
@@ -710,6 +723,24 @@ def get_cephfs_name():
     return result['items'][0].get('metadata').get('name')
 
 
+def pull_images(image_name):
+    """
```
clacroix12: Leave an empty line after each section in the docstring:

```
    """
    Function to pull images on all nodes

    Args:
        image_name (str): Name of the container image to be pulled

    Returns:
        bool: True if deletion of image pull is successful

    """
```

tests/helpers.py
```diff
@@ -710,6 +723,24 @@ def get_cephfs_name():
...
+        bool: True if deletion of image pull is successful
+    """
+
+    node_objs = node.get_node_objs(get_worker_nodes())
+    for node_obj in node_objs:
+        logging.info(f'pulling image "{image_name}" on node {node_obj.name}')
+        assert node_obj.ocp.exec_oc_debug_cmd(
+            node_obj.name, cmd_list=[f'podman pull {image_name}']
+        )
+    return True
```
clacroix12: Do we need to return `True` here? The callers aren't doing anything with the return value.

tests/helpers.py
keesturam: Thanks Coady. I thought about this part i.e., to have a check for `pod_dict_path` before sending this PR. 

I thought of something like this,

```python
    if not pod_dict_path:
        if interface_type == constants.CEPHBLOCKPOOL:
            interface = constants.RBD_INTERFACE
            pod_dict = constants.CSI_RBD_POD_YAML
            pull_images('nginx')
        elif interface_type == constants.CEPHFS_INTERFACE:
            interface = constants.CEPHFS_INTERFACE
            pod_dict = constants.CSI_CEPHFS_POD_YAML
            pull_images('nginx')
    else:
        pod_dict = pod_dict_path
        if interface_type == constants.CEPHBLOCKPOOL:
            interface = constants.RBD_INTERFACE
        elif interface_type == constants.CEPHFS_INTERFACE:
            interface = constants.CEPHFS_INTERFACE
```

The reason I sent it in the current form is that the interface type needs to be set irrespective of whether `pod_dict_path` is defined or not. `pod_dict_path` doesn't define the interface type. So,  checking the `interface type` made more sense to me. pls, let me know what you think.

clacroix12: Yes that's a good point and something I missed. Your suggestion works for me!

tests/helpers.py
keesturam: Removed `true`

tests/helpers.py
keesturam: Done

tests/helpers.py
keesturam: Thanks. Resolving the comment.

tests/helpers.py
```diff
@@ -710,6 +723,27 @@ def get_cephfs_name():
...
+
+    """
+
+    node_objs = node.get_node_objs(get_worker_nodes())
+    for node_obj in node_objs:
+        logging.info(f'pulling image "{image_name}  " on node {node_obj.name}')
+        assert node_obj.ocp.exec_oc_debug_cmd(
+            node_obj.name, cmd_list=[f'podman pull {image_name}']
+        )
+    return
```
Shrivaibavi: Just a return statement with no value here won't return true, Can we change here ? I don't think we need a return statement here, Do we need ?

keesturam: Is there any difference? I don't see any. 

```
>>> def fn1():   
... 	return 
... 
>>> def fn2():
... 	pass
... 
>>> a = fn1()
>>> a
>>> type(a)
<class 'NoneType'>
>>> b = fn2()
>>> b
>>> type(b)
<class 'NoneType'>
```
I'm not sure what is the best practice here, so went with a similar function in ocs-ci.

Shrivaibavi: Best practice is ```return return_value```. If the function do not have a return value, no need to write return statement. So the return can be removed. Only return does not serve any purpose.

keesturam: Okay, removed `return` statement and fixed docstring.

vasukulkarni:
Please link jenkins logs.

keesturam:
I used an existing test to verify this PR.  Snippet from the Jenkins log where the image is being pulled. 

```
09:58:08 - MainThread - root - INFO - pulling image "nginx  " on node compute-0
09:58:08 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig debug nodes/compute-0 -- chroot /host /bin/bash -c "podman pull nginx || echo 'CMD FAILED'; "
09:58:32 - MainThread - ocs_ci.utility.utils - WARNING - Command warning: Starting pod/compute-0-debug ...
To use host binaries, run `chroot /host`

Removing debug pod ...

09:58:32 - MainThread - root - INFO - pulling image "nginx  " on node compute-1
09:58:32 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig debug nodes/compute-1 -- chroot /host /bin/bash -c "podman pull nginx || echo 'CMD FAILED'; "
09:58:49 - MainThread - ocs_ci.utility.utils - WARNING - Command warning: Starting pod/compute-1-debug ...
To use host binaries, run `chroot /host`

Removing debug pod ...

09:58:49 - MainThread - root - INFO - pulling image "nginx  " on node compute-2
09:58:49 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig debug nodes/compute-2 -- chroot /host /bin/bash -c "podman pull nginx || echo 'CMD FAILED'; "
09:59:08 - MainThread - ocs_ci.utility.utils - WARNING - Command warning: Starting pod/compute-2-debug ...
To use host binaries, run `chroot /host`

Removing debug pod ...

```

PR #1408, opened 2020-01-30
Verify pool crush rule is with "type": "zone"

jilju:
Verify pool crush rule is with "type": "zone"
Verifies bug https://bugzilla.redhat.com/show_bug.cgi?id=1763070


jilju:
rebased

PR #1413, opened 2020-01-31
Test to measure pod start time

keesturam:
Adds a helper function to measure pod start time and a simple test case.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -1111,6 +1111,37 @@ def measure_pvc_deletion_time(interface, pv_name):
     return total.total_seconds()
 
 
+def pod_start_time(pod_obj):
+    """
+    Function to measure time taken for container(s) to get into running state
+    by measuring the difference between container's start time (when container
+    went into running state) and started time (when container was actually
+    started)
+
+    Args:
+        pod_obj(obj): pod object to measure start time
+
+    Returns:
+        containers_start_time(dict):
+        Returns the name and start time of container(s) in a pod
+
+    """
+    time_format = '%Y-%m-%dT%H:%M:%SZ'
+    containers_start_time = {}
+    start_time = pod_obj.data['status']['startTime']
+    start_time = datetime.datetime.strptime(start_time, time_format)
+    for container in range(len(pod_obj.data['status']['containerStatuses'])):
+        started_time = pod_obj.data[
+            'status']['containerStatuses'][container]['state'][
+            'running']['startedAt']
+        started_time = datetime.datetime.strptime(started_time, time_format)
+        container_name = pod_obj.data[
+            'status']['containerStatuses'][container]['name']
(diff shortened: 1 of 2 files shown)


keesturam:
Note: This PR depends on #1403 for the `pull_imaes` helper function.

tests/helpers.py
```diff
@@ -1111,6 +1111,44 @@ def measure_pvc_deletion_time(interface, pv_name):
...
+
+    Function to measure time taken for container(s) to get into running state
+
+    Args:
+        pod_obj(obj): pod object to measure start time
+
+    Returns:
+        containers_start_time(dict):
+        Returns the name and start time of container(s) in a pod
+
```
SvetlanaRedHat: Remove  spaces please

tests/helpers.py
```diff
@@ -1111,6 +1111,44 @@ def measure_pvc_deletion_time(interface, pv_name):
     return total.total_seconds()
 
 
+def pod_start_time(pod_obj):
+    """
+
+    Function to measure time taken for container(s) to get into running state
```
SvetlanaRedHat: Please leave a little bit more comments how function works, i did not get the logic straightaway when I read the code.

tests/helpers.py
```diff
@@ -1111,6 +1111,44 @@ def measure_pvc_deletion_time(interface, pv_name):
...
+    Returns:
+        containers_start_time(dict):
+        Returns the name and start time of container(s) in a pod
+
+    """
+    time_format = '%Y-%m-%dT%H:%M:%SZ'
+    containers_start_time = {}
+    start_time = pod_obj.data['status']['startTime']
+    start_time = datetime.datetime.strptime(start_time, time_format)
+    containers_in_pod = len(pod_obj.data['status']['containerStatuses'])
```
fbalak: I suggest to remove (and fix indentation):
```
   containers_in_pod = len(pod_obj.data['status']['containerStatuses'])
   if containers_in_pod == 1:
        started_time = pod_obj.data['status']['containerStatuses'][0][
            'state']['running']['startedAt']
        started_time = datetime.datetime.strptime(started_time, time_format)
        container_name = pod_obj.data['status']['containerStatuses'][0]['name']
        container_start_time = (started_time - start_time).seconds
        containers_start_time[container_name] = container_start_time
    else:
```
The function should work correctly without this even when there is only one container. This should remove unnecessary code duplication.

keesturam: > The function should work correctly without this even when there is only one container. This should remove unnecessary code duplication.

oh yeah! :) I'll fix this.

tests/helpers.py
keesturam: Added some more description.

tests/helpers.py
keesturam: I guess empty lines here is a guideline we follow in this project?

PR #1424, opened 2020-02-03
Added --ocs-version argument for specifying ocs version from CLI

clacroix12:

### DIFF
--- a/conf/ocs_version/ocs-4.3.yaml
+++ b/conf/ocs_version/ocs-4.3.yaml
@@ -1,0 +1,10 @@
+---
+RUN:
+  client_version: '4.3.0-0.nightly'
+DEPLOYMENT:
+  installer_version: "4.3.0-0.nightly"
+  default_ocs_registry_image: "quay.io/rhceph-dev/ocs-registry:latest-4.3"
+  default_latest_tag: 'latest-4.3'
+ENV_DATA:
+  logging_version: '4.3'
+  ocs_version: '4.3'
--- a/conf/ocs_version/ocs-4.2.yaml
+++ b/conf/ocs_version/ocs-4.2.yaml
@@ -1,0 +1,10 @@
+---
+RUN:
+  client_version: '4.2.0-0.nightly'
+DEPLOYMENT:
+  installer_version: "4.2.0-0.nightly"
+  default_ocs_registry_image: "quay.io/rhceph-dev/ocs-registry:4.2-rc"
+  default_latest_tag: '4.2-rc'
+ENV_DATA:
+  logging_version: '4.2'
+  ocs_version: '4.2'
--- a/ocs_ci/framework/main.py
+++ b/ocs_ci/framework/main.py
@@ -7,8 +7,9 @@ import pytest
 import yaml
 
 from ocs_ci import framework
-from ocs_ci.utility import utils
+from ocs_ci.ocs.constants import CONF_DIR
 from ocs_ci.ocs.exceptions import MissingRequiredConfigKeyError
+from ocs_ci.utility import utils
 
 
 def check_config_requirements():
@@ -45,6 +46,9 @@ def init_ocsci_conf(arguments=None):
     """
     parser = argparse.ArgumentParser(add_help=False)
     parser.add_argument('--ocsci-conf', action='append', default=[])
+    parser.add_argument(
+        '--ocs-version', action='store', choices=['4.2', '4.3']
+    )
     args, unknown = parser.parse_known_args(args=arguments)
     for config_file in args.ocsci_conf:
         with open(
@@ -52,6 +56,13 @@ def init_ocsci_conf(arguments=None):
         ) as file_stream:
             custom_config_data = yaml.safe_load(file_stream)
             framework.config.update(custom_config_data)
+    if args.ocs_version:
+        version_config_file = os.path.join(
+            CONF_DIR, 'ocs_version', f'ocs-{args.ocs_version}.yaml'
+        )
+        with open(version_config_file) as file_stream:
+            version_config_data = yaml.safe_load(file_stream)
+            framework.config.update(version_config_data)
     framework.config.RUN['run_id'] = int(time.time())
     bin_dir = framework.config.RUN.get('bin_dir')
     if bin_dir:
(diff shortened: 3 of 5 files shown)


conf/ocs_version/ocs_4_2.yaml
```diff
@@ -0,0 +1,9 @@
+---
```
vasukulkarni: we can remove the old ocs_4.2 conf file ?

clacroix12: I pulled these from @shylesh 's WIP PR, but I think I actually need to update them. I will pull in the values from the file @dahorak added for 4.3 and create a similar one for 4.2 (using our default values).

clacroix12: We have one for 4.3 that we can remove. I just wanted to verify that the values in the files I created are correct before removing, but I will do that as a part of this PR.

clacroix12: Removed the old ocs 4.3 file

clacroix12: Actually holding off on deleting this until we are sure it's not being used in any automation.

clacroix12:
> PR validation [Job FAILED](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/1003//testReport) (installation failed, tests not executed).

Opened an issue regarding the PR validation failure: https://github.com/red-hat-storage/ocs-ci/issues/1425

conf/ocs_version/ocs-4.3.yaml
```diff
@@ -0,0 +1,10 @@
+---
+RUN:
+  client_version: '4.3.0-0.nightly'
+DEPLOYMENT:
+  installer_version: "4.3.0-0.nightly"
```
dahorak: I have one concern/question here: do we want to tightly connect OCP and OCS versions? Like should OCS 4.3 be always installed on OCP 4.3 and similarly can OCS 4.2 be installed only on OCP 4.2?
And how we will deal with the configuration files for OCP version: `ocp-4.2-config.yaml` and `ocp-4.3-config.yaml` (and related Jenkins parameter  `OCP_VERSION`)?

petr-balogh: Agree with Daniel that this will break our OCP_VERSION parameter as you read ocs version config after all config files it's not possible to change the value by passed config.

Solution 1)
Move this block: https://github.com/red-hat-storage/ocs-ci/pull/1424/files/66898fab3c530666936b9e7822c2b72a00efb9b0#diff-6e3f74e4d27052278e5be98b99e36c4eR59-R65 above line 53.

Solution 2)
above line 53 do just:
```python
if args.ocs_version:
    version_config_file = os.path.join(
        CONF_DIR, 'ocs_version', f'ocs-{args.ocs_version}.yaml'
    )
    args.ocsci_conf.insert(0, version_config_file)
```

This will read this config as first one and allow us to overwrite it with additional configs. This way we can keep also OCP version defined in this config files but I think we can just remove ocp versions from OCS version config files and use default from ocs-ci or the one passed by user. WDYT @clacroix12 @dahorak

dahorak: The question I have here is, _if particular version of OCS will be tightly connected to particular version of OCP_?
* if YES, than I'm voting for Solution 2
* if NO, I'm voting for Solution 3

And if the answer is somewhere between... then I don't have strong preference for neither Solution 2 or 3 - I see pros and cons for both solutions...

petr-balogh: @ebenahar do you have any input here? Are we going at some point to stick 4.2 version to some specific version of OCP or will we always have as default the default version of ocs-ci for all versions of OCS?   I am fixing it here: https://github.com/red-hat-storage/ocs-ci/pull/1427 as I think that maybe at some point we would like to stick OCP version for specific version of OCS this solution will allow us to do that. We can later on do append instead of insert to change the order of loading ocs-version file as last one.

clacroix12: @petr-balogh I see that you decoupled the OCP and OCS versions in #1427 . I think decoupling them is the right call, however you also moved the code which updates the configuration based on the ocs version to before the code where we update with custom config files. I don't think we should do this. In the even that a user provides`--ocs-version` AND a custom config file with `--ocsci-conf` and the values in those two things conflict, I think that the explicit `--ocs-version` should "win." Therefore we should update the configuration with the `--ocs-version` values last. Thoughts?

vasukulkarni: Yep the CLI should "win" and override what comes from config file, I dont see where the decoupling of ocs/ocp version happens in that PR, The ocp_version also has a config file, so not sure how they are mixed up?

clacroix12: @vasukulkarni client/installer versions were pulled out of the config files

vasukulkarni: so providing a newer --ocsci-conf with 4.3 installer could have merged new versions right?

vasukulkarni: ex: ocs-version 4.3 --ocsci-conf 4.3_config/4.2_config

shylesh: @clacroix12 , I feel the order in which Petr's PR is updating the config is correct. Because, when user is passing --ocsci-conf option that means he wants those config params to be applied in place of defaults effectively overriding defaults, hence --ocsci-conf wins over defaults. Regarding conflicts , its upto the user to make sure that there are no conflicting variables which might fail deployment. So I feel more semantically we can call ```--ocsci-conf``` as  ```--overrides``` which puts the responsibility on the user to make sure that whatever he is overriding over defaults should be legitimate. It's just my thought.

dahorak: I agree, that particular command line parameter (e.g. `--ocs-version`) should have precedence before common `--ocsci-conf`, but in the same time, we should ensure, that the configuration files loaded using the `--ocs-version` parameter should affect *only* OCS Version and do not change anything else (e.g. OCP version).

clacroix12: @shylesh the conflict isn't with defaults but rather with the value the user provides to `--ocs-version`. We think that the explicit version specified to `--ocs-version` should take priority over the values in the config files provided to `--ocsci-conf`.

@dahorak agreed.

PR #1428, opened 2020-02-04
Add config for 4.3 GA OCP version

petr-balogh:


vasukulkarni:
I think we should add --ocp-version so that its easier to take any z/point version with ocs, which can be a follow on PR.

clacroix12:
> I think we should add --ocp-version so that its easier to take any z/point version with ocs, which can be a follow on PR.

I agree. I think it is also worth it to create a directory under `conf` similar to what we did for `ocs_versions`. We have a decent amount of configurations under `conf/ocsci` and I think we can start to organize a bit, especially for configurations like this where we can expect to see many more with slightly different values.

I vote we merge this and address `--ocp-version` and the new directory in a follow up PR.

PR #1430, opened 2020-02-04
Replicas check before creating dummy osd

fbalak:
Check that there are all replicas up before creating a dummy osd.

### DIFF
--- a/tests/manage/monitoring/conftest.py
+++ b/tests/manage/monitoring/conftest.py
@@ -378,6 +378,14 @@ def measure_corrupt_pg(measurement_dir):
         namespace=config.ENV_DATA.get('cluster_namespace')
     )
     osd_deployments = oc.get(selector=constants.OSD_APP_LABEL).get('items')
+    msg = (
+        f"Some OSDs have no replicas. This is wrong state for this test.\n"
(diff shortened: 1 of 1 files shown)


fbalak:
Hi @nehaberry, this is based on our previous discussion that some test failed because some osd was down at the start of the test. This should help to simplify debugging of that situation. Do you think that this check is enough or would you expand it (for example to do this check before each test)?

nehaberry:
> Hi @nehaberry, this is based on our previous discussion that some test failed because some osd was down at the start of the test. This should help to simplify debugging of that situation. Do you think that this check is enough or would you expand it (for example to do this check before each test)?

@fbalak AFAIR, the issue was in **"test_ceph_health"** where after a dummy OSD is created and deleted, the original OSD never came back UP.

Its good to have the OSD check before bringing down one OSD, but we still need to analyse why the OSD fails to recover in "test_ceph_health" .

Also, atleast for all OSD related tests where we are disrupting OSDs, IMHO, we should have a check for OSD count before failure is introduced.

Thanks. We can discuss offline if required :)

ebenahar:
@fbalak Would like to close old PRs, is this one still relevant?

fbalak:
No, it can be closed.

Issue #1434, opened 2020-02-05
Different capacity messages for different OCS versions


fbalak:
OCS 4.3 adds mention about data deletion into message

PR #1435, opened 2020-02-05
add capacity message for ocs 4.3

fbalak:
fixes https://github.com/red-hat-storage/ocs-ci/issues/1434

### DIFF
--- a/tests/manage/monitoring/prometheus/test_capacity.py
+++ b/tests/manage/monitoring/prometheus/test_capacity.py
@@ -45,7 +45,14 @@ def test_capacity_workload_alerts(
     # TODO(fbalak): it seems that CephFS utilization triggers only firing
     # alerts. This needs to be more investigated.
 
-    if config.ENV_DATA.get('ocs_version') == '4.3':
+    if config.ENV_DATA.get('ocs_version') == '4.2':
+        nearfull_message = (
+            'Storage cluster is nearing full. Expansion is required.'
+        )
+        criticallfull_mesage = (
+            'Storage cluster is critically full and needs immediate expansion'
+        )
+    else:
         nearfull_message = (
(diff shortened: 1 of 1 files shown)


tests/manage/monitoring/prometheus/test_capacity.py
```diff
@@ -44,16 +44,34 @@ def test_capacity_workload_alerts(
     alerts = workload_storageutilization_95p.get('prometheus_alerts')
     # TODO(fbalak): it seems that CephFS utilization triggers only firing
     # alerts. This needs to be more investigated.
+
+    if config.ENV_DATA.get('ocs_version') == '4.2':
```
mbukatov: Would that work for both 4.2.0 and 4.2.1?

fbalak: For OCS 4.2 we have currently only this config:
https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocs_version/ocs-4.2.yaml

tests/manage/monitoring/prometheus/test_capacity.py
```diff
@@ -44,17 +45,36 @@ def test_capacity_workload_alerts(
...
-            ['pending', 'firing'] if interface == 'rbd' else ['firing'],
+            nearfull_message,
+            ['pending', 'firing'],
             'warning'
         ),
         (
             constants.ALERT_CLUSTERCRITICALLYFULL,
-            'Storage cluster is critically full and needs immediate expansion',
-            ['pending', 'firing'] if interface == 'rbd' else ['firing'],
+            criticallfull_mesage,
```
mbukatov: This looks like removal of hidden workaround ... but I'm assuming that since you are removing it now, it's no longer a problem. In the future, it would be better to track this as a workaround from the start.

PR #1444, opened 2020-02-07
Decorator to skip tests based on ocs version

shylesh:

### DIFF
--- a/docs/writing_tests.md
+++ b/docs/writing_tests.md
@@ -41,6 +41,18 @@ For more details what else is possible with `pytest_marker_bugzilla` plugin
 look at the
 [project documentation](https://github.com/eanxgeek/pytest_marker_bugzilla).
 
+## Skipping tests based on ocs version
+
+You can skip a test which is not applicable for specific ocs version using
+skipif_ocs_version('expr') decorator.
+example:
+```python
+from tests.helpers import skipif_ocs_version
+
+@skipif_ocs_version('<4.3')
+def test_fio_workload():
+    pass
+```
 
 ## Base test classes for teams
 
--- a/docs/writing_tests.md
+++ b/docs/writing_tests.md
@@ -44,12 +44,12 @@ look at the
 ## Skipping tests based on ocs version
 
 You can skip a test which is not applicable for specific ocs version using
-skipif_ocs_version('expr') decorator.
+skipif_ocs_version('expr_list') decorator.
 example:
 ```python
-from tests.helpers import skipif_ocs_version
+from ocs_ci.framework.testlib import skipif_ocs_version
 
-@skipif_ocs_version('<4.3')
+@skipif_ocs_version(['>4.1','<4.3'])
 def test_fio_workload():
     pass
 ```
--- a/docs/writing_tests.md
+++ b/docs/writing_tests.md
@@ -44,12 +44,21 @@ look at the
 ## Skipping tests based on ocs version
 
 You can skip a test which is not applicable for specific ocs version using
-skipif_ocs_version('expr_list') decorator.
+
+```@skipif_ocs_version([expression1, expression2, ...])``` decorator.
+
 example:
 ```python
 from ocs_ci.framework.testlib import skipif_ocs_version
 
 @skipif_ocs_version(['>4.1','<4.3'])
+def test_fio_workload():
+    pass
+```
+
+You can also specify a single expression in a string like:
+```python
+@skipif_ocs_version('<4.3')
 def test_fio_workload():
     pass
 ```
--- a/ocs_ci/framework/testlib.py
+++ b/ocs_ci/framework/testlib.py
@@ -1,4 +1,5 @@
 from ocs_ci.framework.pytest_customization.marks import *  # noqa: F403
+from functools import wraps
 
 
 @pytest.mark.usefixtures(  # noqa: F405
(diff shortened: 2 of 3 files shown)


shylesh:
Verified locally, ran with 
```run-ci ./tests --cluster-name shmohantl --cluster-path ~/skiptest/  -k test_fio_workload --ocs-version 4.2 --verbose```

got the results

```-------------------------------------------------------------------- live log collection ---------------------------------------------------------------------
17:55:31 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig /home/shylesh/skiptest/kubeconfig get Node  --selector=node-role.kubernetes.io/worker -o yaml
collected 357 items / 356 deselected / 1 selected                                                                                                            

tests/libtest/test_fio_workload.py::TestFIOWorkload::test_fio_with_block_storage SKIPPED                                                               [100%]
```

with decorator
```python
@skipif_ocs_version('<4.3')
def test_fio_workload():
    pass
```

tests/helpers.py
```diff
@@ -1907,3 +1908,27 @@ def modify_osd_replica_count(resource_name, replica_count):
     params = f'{{"spec": {{"replicas": {replica_count}}}}}'
     resource_name = '-'.join(resource_name.split('-')[0:4])
     return ocp_obj.patch(resource_name=resource_name, params=params)
+
+
+def skipif_ocs_version(expr):
```
petr-balogh: What about to move it here:  https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/testlib.py

From here we import all the markers, base class and so on so testlib sounds good to me for such functionality. WDYT?

tests/helpers.py
```diff
@@ -1907,3 +1908,27 @@ def modify_osd_replica_count(resource_name, replica_count):
...
+    """
+    This is a decorator to skip the test if condition evaluates to
+    true based on expression
+
+    Args:
+        expr (str): condition for which we need to check,
+            eg: '<4.3', '>4.3'
```
petr-balogh: add also example for `>=` .

What about if we will need to do something like . VERSION >=4.3 and VERSION < 4.6.

I can imagine that you can allow do something like  `> = 4.3 and {version} < 4.6` if you will do before expr.format(version=config.ENV_DATA['ocs_version'])

tests/helpers.py
```diff
@@ -1907,3 +1908,27 @@ def modify_osd_replica_count(resource_name, replica_count):
...
+
+    Return:
+        decorated function
+    """
+    comparision_str = config.ENV_DATA['ocs_version'] + expr
+    # skip_this will be either True or False after eval
+    skip_this = eval(comparision_str)
+
```
petr-balogh: What about to use: @functools.wraps(func)

You can read a bit about it here:
https://lerner.co.il/2019/05/05/making-your-python-decorators-even-better-with-functool-wraps/

clacroix12: @shylesh we use what Petr is talking about in our [retry module](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/retry.py) if you want another example.

vasukulkarni: I think the better place for this is in conftest with pytest.fixture so we can avoid decorating it.

tests/helpers.py
shylesh: @petr-balogh ,

What would be the best way to specify ranges during decoration in tests, 
how does this looks
```python
@skipif_ocsversion(">4.2 and <= 4.3")
def test_xyz()
    pass
```
is this ok ? or any other suggestions.

dahorak: Another possible syntax might be, to expect string or list of strings containing the simple expression like what it is documented (`<4.3`) - and then simply iterate over the list and ensure, that all the expressions are true... Not sure, which option is more clear and readable though.

clacroix12: I agree with @dahorak's suggestion. I think we can start off iterating over a list of expressions (essentially only using `and`). This approach is readable and makes it clear to anyone reading this code what versions the test supports. If we feel the need for more operators like `or` we can revisit the logic.

shylesh: @dahorak @clacroix12 
sounds like good idea, it would be easy to iterate over list and build expression to eval.

tests/helpers.py
petr-balogh: I think that decorator is easier to use and easy to send parameter to it. How will you send the versions you would like to avoid to the fixture? Can you please elaborate more? Thanks

tests/helpers.py
shylesh: done

tests/helpers.py
shylesh: Done, user can pass a list

ocs_ci/framework/testlib.py
```diff
@@ -35,3 +36,32 @@ class EcosystemTest(BaseTest):
...
+
+
+def skipif_ocs_version(expr_list):
+    """
+    This is a decorator to skip the test if condition evaluates to
+    true based on expression
+
+    Args:
+        expr_list (list): condition for which we need to check,
+            eg: ['<4.3', '>4.2'], ['<=4.3', '>=4.2']
```
dahorak: Would it make sense to handle expressions argument both as simple string and as a list? (e.g. if I want to specify only one condition  like `'>4.2'`, using `@skipif_ocs_version('>4.1')` is slightly more clear and easier than ` @skipif_ocs_version(['>4.1'])`.
And it might be more error prone, because with the current implementation if somebody will provide just string, it will iterate over each character of that string and might produce slightly confusing errors.

My suggestion is, to add following line at the beginning (and rename `expr_list` to `expressions`):
```python
expressions = [expressions] if isinstance(expressions, str) else expressions
```

docs/writing_tests.md
```diff
@@ -41,6 +41,18 @@ For more details what else is possible with `pytest_marker_bugzilla` plugin
 look at the
 [project documentation](https://github.com/eanxgeek/pytest_marker_bugzilla).
 
+## Skipping tests based on ocs version
+
+You can skip a test which is not applicable for specific ocs version using
+skipif_ocs_version('expr_list') decorator.
```
dahorak: I would remove the apostrophes from around the argument `skipif_ocs_version('expr_list')` ... as it can be list.

petr-balogh: or put the list directly there the same as you have in example below

ocs_ci/framework/testlib.py
shylesh: done

petr-balogh:
@shyles please send email to ocs-qe about this skip decorator. You can directly link the proper section from: https://ocs-ci.readthedocs.io/en/latest/docs/writing_tests.html

Issue #1446, opened 2020-02-07
4.2-rc build is not used anymore, we have to stop using it till it will be used again


petr-balogh:
Talking to Boris to start using again 4.2-rc tag for whatever 4.2.Z-rc build to not have to update config each for each Z-stream.

For now I would use latest-4.2.

```
upgrade = False, latest_tag = '4.2-rc'

    def get_latest_ds_olm_tag(upgrade=False, latest_tag=None):
        """
        This function returns latest tag of OCS downstream registry or one before
        latest if upgrade parameter is True
    
        Args:
            upgrade (str): If True then it returns one version of the build before
                the latest.
            latest_tag (str): Tag of the latest build. If not specified
                config.DEPLOYMENT['default_latest_tag'] or 'latest' will be used.
    
        Returns:
            str: latest tag for downstream image from quay registry
    
        Raises:
            TagNotFoundException: In case no tag found
    
        """
        latest_tag = latest_tag or config.DEPLOYMENT.get(
            'default_latest_tag', 'latest'
        )
        _req = requests.get(
            constants.OPERATOR_CS_QUAY_API_QUERY.format(tag_limit=20)
        )
        latest_image = None
        tags = _req.json()['tags']
        for tag in tags:
            if tag['name'] == latest_tag:
                latest_image = tag['image_id']
                break
        if not latest_image:
>           raise TagNotFoundException(f"Couldn't find latest tag!")
E           ocs_ci.ocs.exceptions.TagNotFoundException: Couldn't find latest tag!
```

We should also increase number of tags to gather from quay if 4.2 builds won't be produced that often.

PR #1447, opened 2020-02-07
Update 4.2 config to use latest-4.2 tag

petr-balogh:
Fixes: #1446


petr-balogh:
Please ignore the message above I verified MR which allow us to test PR over existing cluster

PR #1451, opened 2020-02-10
Function to get osd utilization and validate the osd utilization

Akarsha-rai:
This includes function:
- to get osd utilization 
- to validate the osd utilization matches with the given one

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -659,6 +659,6 @@ def validate_osd_utilization(osd_used=80):
             logger.info(f"{osd} used value {value}")
         else:
             _rc = False
-            logger.error(f"{osd} used value {value}")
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/cluster.py
```diff
@@ -619,3 +619,46 @@ def validate_pdb_creation():
...
+
+    """
+    _rc = True
+    osd_filled = get_osd_utilization()
+    for osd, value in osd_filled.items():
+        if int(value) >= osd_used:
+            logger.info(f"{osd} used value {value}")
+        else:
+            _rc = False
+            logger.error(f"{osd} used value {value}")
```
vavuthu: Do we need to log error here if utilization is greater than given value instead of warning ?

Akarsha-rai: Agreed, will change to warn in next PR.

PR #1458, opened 2020-02-11
Ability to cleanup AWS clusters in a periodic way

ebenahar:
The intention is to run this script with a cron job from a persistent server.


ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -35,7 +38,36 @@ def cleanup(cluster_name, cluster_id):
...
+    Args:
+         time_to_delete (int): The maximum time in seconds that is allowed
+            for clusters to continue running
+
+    Returns:
+        list: The cluster names (e.g ebenahar-cluster-gqtd4) to be provided to the
+            ci-cleanup script
+
+    """
+    conn = boto3.resource('ec2', region_name='us-east-2')
```
petr-balogh: I see region_name hardcoded here and not configurable. I would like to see this configurable.

ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -35,7 +38,35 @@ def cleanup(cluster_name, cluster_id):
...
+    instances = conn.instances.filter(Filters=[{'Name': 'tag:Name', 'Values': ['*master-0']}])
```
petr-balogh: I think we should select clusters for delete from VPCs as for each cluster we have one VPC. From VPC you can read tag:
 kubernetes.io/cluster/CLUSTER-name-ktzdr   where ktzdr is auto generated suffix.  But most  resources created with cluster should be tagged with such tag kubernetes.io/cluster/CLUSTER-name-ktzdr so you can find all ec2 instances for such cluster by that tag.

Haven't checked all the data provided from VPC object but I guess there will be also some info about creation datetime so you can directly take from vpc maybe.

ocs_ci/cleanup/aws/cleanup.py
ebenahar: fixed

ocs_ci/cleanup/aws/cleanup.py
ebenahar: done

ebenahar:
Verified locally

ebenahar:
[cleanup_log.zip](https://github.com/red-hat-storage/ocs-ci/files/4188275/cleanup_log.zip)

ebenahar:
Please do not merge yet as I would like to consider how we make sure this new capability is not being used in the wrong way.

ocs_ci/ocs/constants.py
```diff
@@ -629,3 +629,6 @@
 
 # gather bootstrap
 GATHER_BOOTSTRAP_PATTERN = 'openshift-install gather bootstrap --help'
+
+DEFAULT_AWS_REGION = 'us-east-2'
```
clacroix12: Should this be in `defaults.py` instead of `constants.py`?

ocs_ci/ocs/constants.py
```diff
@@ -629,3 +629,6 @@
...
+CLUSTER_PREFIXES_TO_EXCLUDE_FROM_DELETION = ['jnk']
```
clacroix12: Does this need to be in `constants.py`? This seems very specific to the aws cleanup code. I'm not sure it has to live in the constants module, as that was designed to hold values that are shared across multiple other modules.

ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -55,3 +97,40 @@ def main():
...
+
+def aws_cleanup():
+    parser = argparse.ArgumentParser(description='Cleanup AWS Resource')
+    parser.add_argument(
+        '--hours',
+        type=int,
+        nargs=1,
+        action='append',
+        required=True,
+        help="Cluster running time in hours"
```
clacroix12: This is sort of vague, and doesn't mention that it is the maximum age of the cluster you are allowing. Maybe we can be more specific here:

"Maximum running time of the cluster (in hours). Clusters older than this will be deleted"

WDYT?

ocs_ci/ocs/constants.py
ebenahar: Thanks, just changed to a new defaults file under cleanup dir. The reason for the new file is that these defaults are not related to the OCS, but rather to the cleanup CLI

ocs_ci/ocs/constants.py
ebenahar: same as the above comment. Please let me know if any change is needed

ocs_ci/cleanup/aws/cleanup.py
ebenahar: thanks! changed

ocs_ci/cleanup/aws/defaults.py
```diff
@@ -0,0 +1,7 @@
+"""
+Constants module for AWS cleanup
+"""
+DEFAULT_AWS_REGION = 'us-east-2'
```
clacroix12: This can probably just be `AWS_REGION` now that it's in a defaults module, but not a big deal :)

ebenahar:
[cleanup_log.zip](https://github.com/red-hat-storage/ocs-ci/files/4189022/cleanup_log.zip)

PR #1466, opened 2020-02-12
Function to get resource utilization(CPU,memory) of a node

Shrivaibavi:

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -743,3 +743,35 @@ def validate_osd_utilization(osd_used=80):
             logger.warn(f"{osd} used value {value}")
 
     return _rc
+
+
+def get_node_resource_utilization(nodename=None):
+    """
+    Gets the node's cpu and memory utilization in percentage
+
+    Args:
+        nodename (str): The node name
+
+    Returns:
+        cpu_utilization (int) : if nodename is present
+        memory_utilization (int) : if nodename is present
+        resource_utilization_all_nodes (str) : if nodename is None
+
+    """
+    obj = ocp.OCP()
+    resource_utilization_all_nodes = obj.exec_oc_cmd(
+        command='adm top nodes', out_yaml_format=False
+    )
+    if nodename:
+        for value in resource_utilization_all_nodes.split("\n"):
+            if nodename in value:
+                value = re.findall(r'\d+', value.strip())
+                cpu_utilization = value[2]
+                logger.info("The CPU utilized by the node "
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/cluster.py
```diff
@@ -619,3 +619,35 @@ def validate_pdb_creation():
...
+
+def get_node_resource_utilization(nodename=None):
+    """
+    Gets the node's cpu and memory utilization in percentage
+
+    Args:
+        nodename (str): The node name
+
+    Returns:
+        cpu_utilization (int) : if nodename is present
```
clacroix12: I don't really like the idea of the return type changing based on arguments provided. I think it would be better to keep the return type and format consistent regardless of if `nodename` is provided. This way the user can call the function and handle the return value in the same way for both cases.

What if we instead return a dict that includes data for each node? Where if the `nodename` is provided the dict length will just be 1.

Something like this:
```python
resource_utilization_data = {
    'example-node-1': {
        'cpu': <cpu_utilization>,
        'memory': <memory_utilization>
    },
    'example-node-2': {
        'cpu': <cpu_utilization>,
        'memory': <memory_utilization>
    }
}
```
What do you think?

Shrivaibavi: ACK, I will send in next commit

Shrivaibavi: @coady Thanks for the pointer, Modified

ocs_ci/ocs/cluster.py
```diff
@@ -743,3 +744,41 @@ def validate_osd_utilization(osd_used=80):
...
+        nodename (str) : The node name
+
+    Returns:
+        dict : Node name and its cpu and memory utilization in
+               percentage
+
+    """
+
+    node_names = [nodename] if nodename else [
+        node.name for node in get_typed_nodes(node_type='worker')
```
ramkiperiy: Add param for node_type, can use this function to get utilization of `master` nodes too, default can be `worker`

ocs_ci/ocs/cluster.py
```diff
@@ -746,7 +746,7 @@ def validate_osd_utilization(osd_used=80):
     return _rc
 
 
-def get_node_resource_utilization(nodename=None):
+def get_node_resource_utilization(nodename=None, node_type='worker'):
```
RazTamir: add node type to docstring

Shrivaibavi: @RazTamir done

tiffanyn108:
LGTM.

PR #1467, opened 2020-02-12
AWS cleanup: skip cloudformation VPCs

ebenahar:

### DIFF
--- a/ocs_ci/cleanup/aws/cleanup.py
+++ b/ocs_ci/cleanup/aws/cleanup.py
@@ -50,17 +50,25 @@ def get_clusters_to_delete(time_to_delete, region_name, prefixes_to_spare):
         prefixes_to_spare (list): The cluster prefixes to spare
 
     Returns:
-        list: The cluster names (e.g ebenahar-cluster-gqtd4) to be provided to the
-            ci-cleanup script
+        tuple: List of the cluster names (e.g ebenahar-cluster-gqtd4) to be provided to the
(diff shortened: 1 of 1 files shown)


ebenahar:
[cleanup_log.zip](https://github.com/red-hat-storage/ocs-ci/files/4194988/cleanup_log.zip)

clacroix12:
I'm not quite sure what we are trying to do here. Title says we are handling cloudformation VPCs but the code is just logging which ones we found. Are we supposed to be deleting these cloudformation stacks?

I'm assuming this is meant to support cleaning up after UPI deployments since that's the only time we use cloudformation to create the infrastructure. If this is the case then we will also need to handle several other cloudformation stacks before we can delete the VPC stack. You can see what we are doing to destroy the cluster in the [AWSUPI destroy_cluster method](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/aws.py#L847-L885)

ebenahar:
@clacroix12 the handling in the title is indeed misleading. 
I have observed that when VPCs that are of cloudfomation exist, aws-cleanup fails as there is no tag of 'Name' for that VPC. 
Handling here is done in a sense of not failing the program, but skipping this VPC deletion with a warning message at the end detailing the names of those VPCs not deleted. 
The reason is that AFIAK, we don't have a way of removing this kind of resources using our cleaup

clacroix12:
@ebenahar got it! Yeah, we handle the cleanup in our UPI destroy method but not in our `ci-cleanup` yet.

Issue #1469, opened 2020-02-12
Only append master_ip to list if it exists for gather bootstrap


clacroix12:
While building out the data we use to construct the gather bootstrap command, we are appending the IP of the master node even in a `None` case. We should only be appending to the `master_ips` list if the IP exists.

Example job: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/4541/console

We see logs where we found a control plane node with `None` for the IP:
```
20:37:50 - MainThread - ocs_ci.utility.bootstrap - INFO - Running gather bootstrap
20:37:50 - MainThread - ocs_ci.utility.bootstrap - INFO - Retrieving bootstrap node data
20:38:17 - MainThread - ocs_ci.utility.bootstrap - INFO - Found bootstrap node bootstrap-0 with IP 10.8.130.31
20:38:38 - MainThread - ocs_ci.utility.bootstrap - INFO - Found master node control-plane-0 with IP 10.8.130.37
20:38:58 - MainThread - ocs_ci.utility.bootstrap - INFO - Found master node control-plane-1 with IP 10.8.130.36
20:39:15 - MainThread - ocs_ci.utility.bootstrap - INFO - Found master node control-plane-2 with IP None
20:39:15 - MainThread - ocs_ci.utility.utils - INFO - Executing command: /home/jenkins/bin/openshift-install gather bootstrap --bootstrap 10.8.130.31 --dir /home/jenkins/current-cluster-dir/logs/gather_bootstrap --log-level debug --key /home/jenkins/.ssh/openshift-dev.pem --master 10.8.130.37 --master 10.8.130.36 --master None 
```

Which leads to:
```
level=debug msg="Collecting info from None"
level=debug msg="lost connection"
level=debug msg="ssh: Could not resolve hostname none: Name or service not known\r"
```

PR #1470, opened 2020-02-12
Ensure Nonetype values are not added to the master_ips list for gathering bootstrap data

clacroix12:
Fixes: #1469 


vasukulkarni:
I think this can solve the problem we have with bootstrap failure in jenkins?

clacroix12:
> I think this can solve the problem we have with bootstrap failure in jenkins?

This won't solve the issue, but it will reduce cluttering the logs with additional errors when we can't reach the 'None' IP of the master nodes. It will also make it clearer when we aren't able to get data for all of the master nodes.

petr-balogh:
Is this verified?

PR #1471, opened 2020-02-13
Rely on CSV for fetching ocs version info

shylesh:

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -941,6 +941,7 @@ def get_ocs_build_number():
         str: build number for ocs operator version
 
     """
+    # Importing here to avoid circular dependency
     from ocs_ci.ocs.resources.csv import get_csvs_start_with_prefix
 
     build_num = ""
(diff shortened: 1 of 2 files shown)


shylesh:
image

ocs_ci/utility/utils.py
```diff
@@ -940,17 +941,15 @@ def get_ocs_build_number():
         str: build number for ocs operator version
 
     """
-    from ocs_ci.ocs.resources.catalog_source import CatalogSource
+    from ocs_ci.ocs.resources.csv import get_csvs_start_with_prefix
```
clacroix12: Is there a reason we are importing here as opposed to at the top of the file with all other imports? If so please add a comment explaining why we are importing here.

shylesh: Sorry I forgot to put it in comments, there was circular dependency hence I imported it here. I will update the comment

shylesh: done

shylesh: done

shylesh:
image

shylesh:
Verified with live deployment. 
Version appears as expected in subject and ocs operator

PR #1472, opened 2020-02-13
Update Upi script 4 3  and repo for RHEL workers

shylesh:
1. Updated with proper UPI scripts for 4.2 and 4.3
2. Updated repo for RHEL workers


ocs_ci/deployment/aws.py
```diff
@@ -319,8 +328,11 @@ def deploy_prereq(self):
                 'num_workers': str(config.ENV_DATA['worker_replicas']),
                 'AVAILABILITY_ZONE_COUNT': str(config.ENV_DATA.get(
                     'availability_zone_count', ''
-                ))
+                )),
```
clacroix12: Is this intentional?

ocs_ci/ocs/constants.py
```diff
@@ -522,6 +523,11 @@
     '4.2.0': OCP4_2_REPO
 }
 
+# UPI script paths
```
clacroix12: Do these need to be in constants since they are only used in AWSUPI?

ocs_ci/deployment/aws.py
```diff
@@ -291,15 +291,24 @@ class OCPDeployment(BaseOCPDeployment):
...
             self.upi_repo_path = os.path.join(
                 constants.EXTERNAL_DIR, upi_repo_name,
             )
 
-            self.upi_script_path = os.path.join(
-                self.upi_repo_path,
-                'v3-launch-templates/functionality-testing'
-                '/aos-4_2/hosts/'
-            )
+            if self.ocp_version == '4.3':
```
clacroix12: Instead of having separate constants for each version could we make this a bit more simple and just use the ocp version to construct the filepath directly?

```python
upi_common_base = 'v3-launch-templates/functionality-testing'
self.upi_script_path = os.path.join(
    self.upi_repo_path,
    upi_common_base,
    self.ocp_version.replace('.','_'),
    'hosts'
)
```
If we _do_ want to go with setting explicit paths for certain versions, I would suggest we use a dict instead of individual constants. This way we don't need an `if`/`elif` block for each separate version and instead we can do it once. This also has the benefit of scaling better when we inevitably have more versions to account for.

```python
upi_common_base = 'v3-launch-templates/functionality-testing'
ocp_version_paths = {
    '4.2': os.path.join(upi_common_base, 'aos-4_2/hosts/'),
    '4.3': os.path.join(upi_common_base, 'aos-4_3/hosts/')
}
self.upi_script_path = os.path.join(
    self.upi_repo_path,
    ocp_version_paths[self.ocp_version],
)
```

ocs_ci/ocs/constants.py
shylesh: VMware also uses same constant.

shylesh: oh sorry, you meant line 527 - 529 ? that's not required

shylesh:
1. Deployment test pass for OCP4.3 (RHEL workers on UPI)with OCS4.2. https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/4605/.

I will run OCP4.2 (RHEL workers on UPI ) run is pending, I will update the results once done.

shylesh:
2. OCP-4.2 with OCS-4.2 run here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/4617.

ocs_ci/deployment/aws.py
```diff
@@ -291,14 +291,21 @@ class OCPDeployment(BaseOCPDeployment):
         def __init__(self):
             super(AWSUPI.OCPDeployment, self).__init__()
             upi_repo_name = f'openshift-misc-{config.RUN["run_id"]}'
+            self.ocp_version = ".".join(
+                config.DEPLOYMENT.get('installer_version').split('.')[:-2]
```
petr-balogh: I see this: config.DEPLOYMENT.get('installer_version').split('.')[:-2] on 2 places. If I remember correctly you said that you just copy pasted this from some other place as well.

WDYT about have some func in utils like get_ocp_version() which will return this value and you will replace also on other places where it is used as well?

Maybe you can add also parameter like separator and in case it's defined you can replace for example with this separator like _ so it will return "4_2" which you can use on different places where you do replace

ocs_ci/deployment/aws.py
```diff
@@ -557,9 +567,13 @@ def run_ansible_playbook(self):
         pem_dst_path = "/openshift-dev.pem"
         pod.upload(rhel_pod_obj.name, pem_src_path, pem_dst_path)
         repo_dst_path = "/etc/yum.repos.d/"
-        repo_file = os.path.basename(constants.OCP4_2_REPO)
+        ocp_version = ".".join(
+            config.DEPLOYMENT.get('installer_version').split('.')[:-2]
+        )
+        repo_file = os.path.basename(constants.OCP_REPO_MAP[ocp_version])
```
petr-balogh: I thought that it will be here also dynamically and no need to add each version to the constants.
```python
repo = os.path.join(constants.REPO_DIR, f"ocp_{version_with_underscore}.repo")
# where version_with_underscore can be from the mentioned function with separator _
repo_file = os.path.basename(repo)
```

One more thing you can add here is check if file exist or not. If not raise the exception that for this version repo file doesn't exist. So we will directly see that we have to add one.

Anyway I think this same thing has to be done for VmWare and in the consolidation of your and Vijay's PR both should use the same function to do this.  But for VmWare it can be handled in separate PR but when fixing for vmware as well we need to make sure to use the same function and not have this logic on both places.

ocs_ci/deployment/aws.py
```diff
@@ -291,14 +291,21 @@ class OCPDeployment(BaseOCPDeployment):
...
+                config.DEPLOYMENT.get('installer_version').split('.')[:-2]
+            )
+
+            self.upi_common_base = 'v3-launch-templates/functionality-testing'
             self.upi_repo_path = os.path.join(
                 constants.EXTERNAL_DIR, upi_repo_name,
             )
-
+            version_str = (
+                'aos-' + self.ocp_version.replace('.', '_') + '/hosts/'
```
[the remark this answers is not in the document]
clacroix12: I agree with petr. We should always be using os.path.join instead of string concatenation  when building filepaths.

ocs_ci/deployment/aws.py
```diff
@@ -321,6 +328,9 @@ def deploy_prereq(self):
                     'availability_zone_count', ''
                 ))
             }
+            if self.ocp_version == '4.3':
```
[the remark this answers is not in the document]
clacroix12: I would actually like to ask if setting this _hurts_ us in 4.2. If the environment variable isn't being used in the 4.2 script I don't see the harm in setting it for all versions and doing away with the `if`.

shylesh:
Rebased on latest master

ocs_ci/ocs/constants.py
```diff
@@ -522,6 +523,11 @@
...
+UPI_COMMON_BASE = 'v3-launch-templates/functionality-testing'
+UPI_4_2_BASE_PATH = os.path.join(UPI_COMMON_BASE, 'aos-4_2/hosts/')
+UPI_4_3_BASE_PATH = os.path.join(UPI_COMMON_BASE, 'aos-4_3/hosts/')
```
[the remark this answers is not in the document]
shylesh: Done

ocs_ci/ocs/constants.py
```diff
@@ -471,6 +471,7 @@
 
 # repos
 OCP4_2_REPO = os.path.join(REPO_DIR, "ocp_4_2.repo")
+OCP4_3_REPO = os.path.join(REPO_DIR, "ocp_4_3.repo")
```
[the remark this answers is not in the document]
shylesh: done

ocs_ci/deployment/aws.py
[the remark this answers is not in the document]
shylesh: done

ocs_ci/deployment/aws.py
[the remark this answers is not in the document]
shylesh: done, Since extra environment variable doesn't matter to the UPI script I just removed the if condition.

ocs_ci/deployment/aws.py
shylesh: Done, introduced a new function get_ocp_version().

ocs_ci/deployment/aws.py
shylesh: done.

ocs_ci/utility/utils.py
```diff
@@ -1039,6 +1039,29 @@ def get_csi_versions():
     return csi_versions
 
 
+def get_ocp_version(seperator=None):
+    """
+    Get current ocp version
+
+    Args:
+        seperator (str): String that would seperate major and
+        minor version nubers
```
petr-balogh: one indent right as it's still belong to `separator` arg.

ocs_ci/utility/utils.py
```diff
@@ -1039,6 +1039,29 @@ def get_csi_versions():
...
+    Returns:
+        string : If seperator is 'None', version string will be returned as is
+        eg: '4.2', '4.3'.
+        If seperator is provided then '.' in the version string would be
+        replaced by seperator and resulting string will be returned.
+        eg: If seperator is '_' then string returned would be '4_2'
+
+    """
+    char = seperator if seperator else '.'
+    log.info(config.DEPLOYMENT['installer_version'])
```
petr-balogh: I do not think we need to log anything here and at least not with info level. If you insist from some reason on logging the version you should use debug level but I would remove this line.

ocs_ci/deployment/aws.py
```diff
@@ -569,9 +574,12 @@ def run_ansible_playbook(self):
         pem_dst_path = "/openshift-dev.pem"
         pod.upload(rhel_pod_obj.name, pem_src_path, pem_dst_path)
         repo_dst_path = "/etc/yum.repos.d/"
-        repo_file = os.path.basename(constants.OCP4_2_REPO)
+        repo = os.path.join(
+            constants.REPO_DIR, f"ocp_{get_ocp_version('_')}.repo"
+        )
+        repo_file = os.path.basename(repo)
```
petr-balogh: Don't we want to check if `repo` file exists? Something like:
```python
assert os.path.exists(repo), f"Required repo file {repo} doesn't exist!"
```

ocs_ci/repos/ocp_4_2.repo
```diff
@@ -1,6 +1,6 @@
 [openshift]
 name=RHopenshift
-baseurl=https://mirror.openshift.com/enterprise/all/4.2/v4.2.0-201910111801_2019-10-11.2/RH7-RHAOS-4.2/x86_64/os/
+baseurl=https://mirror.openshift.com/enterprise/all/4.2/latest/RH7-RHAOS-4.2/x86_64/os/
```
petr-balogh: 👍

shylesh:
OCP4.3 + OCS4.2 Jenkins run success:- https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/4725/

shylesh:
I will update the results for OCP4.2+OCS4.2 once run is finished.

ocs_ci/deployment/aws.py
shylesh: done

ocs_ci/utility/utils.py
shylesh: sorry , I had it for debugging forgot to cleanup

ocs_ci/utility/utils.py
shylesh: done

shylesh:
> I will update the results for OCP4.2+OCS4.2 once run is finished.
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/4731. OCP4.2+OCS4.2 run finished successfully.

Issue #1477, opened 2020-02-14
One of OSDs is not up after test with measure_corrupt_pg fixture


fbalak:
Once fixed we will need to revert this one PR: #1478

PR #1478, opened 2020-02-14
skip test_ceph.py test cases

fbalak:
Skip tests in test_ceph.py because of https://github.com/red-hat-storage/ocs-ci/issues/1477 until further investigation is done.


petr-balogh:
@fbalak created new label Skip test, so maybe we can use this one to easily identify similar issues/PRs. I also added description to the issue about reverting this.

PR #1479, opened 2020-02-14
Fixed PGSQL Master node drain

tiffanyn108:
Signed-off-by: Tiffany Nguyen <tunguyen@redhat.com>


vasukulkarni:
@tiffanyn108 if its verified then remove the "needs testing" label, we can't have both.

PR #1496, opened 2020-02-18
Allow upgrade between versions like 4.2 -> 4.3

petr-balogh:
Infra ticket:
https://projects.engineering.redhat.com/browse/OCSQE-227

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -104,3 +104,7 @@ ENV_DATA:
   # measurement_dir: '/tmp/ocs_ci_monitoring_measurement/'
   # Default RHCOS image to be used for VmWare deployment
   vm_template: 'rhcos-4.3.0-x86_64-vmware'
+
+# This section is related to upgrade
+UPGRADE:
+  upgrade_to_latest: true
--- a/tests/ecosystem/upgrade/test_upgrade.py
+++ b/tests/ecosystem/upgrade/test_upgrade.py
@@ -2,6 +2,7 @@ import os
 import logging
 import yaml
 from copy import deepcopy
+from pkg_resources import parse_version
 from tempfile import NamedTemporaryFile
 from time import sleep
 
@@ -135,7 +136,14 @@ def test_upgrade():
     upgrade_version = config.UPGRADE.get(
         "upgrade_ocs_version", version_before_upgrade
     )
-    version_change = version_before_upgrade != upgrade_version
+    parsed_version_before_upgrade = parse_version(version_before_upgrade)
+    parsed_upgrade_version = parse_version(upgrade_version)
+    assert parsed_upgrade_version >= parsed_version_before_upgrade, (
+        f"Version you would like to upgrade to: {parsed_upgrade_version} "
+        f"is not higher or equal to the version you currently running: "
+        f"{parsed_version_before_upgrade}"
+    )
+    version_change = parsed_version_before_upgrade > parsed_upgrade_version
     if version_change:
         version_config_file = os.path.join(
             constants.CONF_DIR, 'ocs_version', f'ocs-{upgrade_version}.yaml'
--- a/tests/ecosystem/upgrade/test_upgrade.py
+++ b/tests/ecosystem/upgrade/test_upgrade.py
@@ -1,4 +1,6 @@
+import os
 import logging
+import yaml
 from copy import deepcopy
 from tempfile import NamedTemporaryFile
 from time import sleep
@@ -129,9 +131,29 @@ def test_upgrade():
         resource_name=constants.OPERATOR_CATALOG_SOURCE_NAME,
         namespace=constants.MARKETPLACE_NAMESPACE,
     )
+    version_before_upgrade = config.ENV_DATA.get("ocs_version")
+    upgrade_version = config.UPGRADE.get(
+        "upgrade_ocs_version", version_before_upgrade
+    )
+    version_change = version_before_upgrade != upgrade_version
+    if version_change:
+        version_config_file = os.path.join(
+            constants.CONF_DIR, 'ocs_version', f'ocs-{upgrade_version}.yaml'
+        )
+        assert os.path.exists(version_config_file), (
+            f"OCS version config file {version_config_file} doesn't exist!"
+        )
+        with open(
+            os.path.abspath(os.path.expanduser(version_config_file))
+        ) as file_stream:
+            custom_config_data = yaml.safe_load(file_stream)
+            config.update(custom_config_data)
     image_url = ocs_catalog.get_image_url()
     image_tag = ocs_catalog.get_image_name()
-    if config.DEPLOYMENT.get('upgrade_to_latest', True):
+    ocs_registry_image = config.UPGRADE.get('upgrade_ocs_registry_image')
+    if ocs_registry_image:
+        image_url, image_tag = ocs_registry_image.split(':')
+    elif config.UPGRADE.get('upgrade_to_latest', True) or version_change:
         new_image_tag = get_latest_ds_olm_tag()
     else:
         new_image_tag = get_next_version_available_for_upgrade(image_tag)
--- a/tests/ecosystem/upgrade/test_upgrade.py
+++ b/tests/ecosystem/upgrade/test_upgrade.py
@@ -139,11 +139,11 @@ def test_upgrade():
     parsed_version_before_upgrade = parse_version(version_before_upgrade)
     parsed_upgrade_version = parse_version(upgrade_version)
     assert parsed_upgrade_version >= parsed_version_before_upgrade, (
-        f"Version you would like to upgrade to: {parsed_upgrade_version} "
+        f"Version you would like to upgrade to: {upgrade_version} "
         f"is not higher or equal to the version you currently running: "
-        f"{parsed_version_before_upgrade}"
+        f"{version_before_upgrade}"
     )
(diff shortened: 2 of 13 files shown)


petr-balogh:
Running verification for upgrade 4.2 -> 4.3 here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/4764/console

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -129,9 +131,29 @@ def test_upgrade():
         resource_name=constants.OPERATOR_CATALOG_SOURCE_NAME,
         namespace=constants.MARKETPLACE_NAMESPACE,
     )
+    version_before_upgrade = config.ENV_DATA.get("ocs_version")
+    upgrade_version = config.UPGRADE.get(
+        "upgrade_ocs_version", version_before_upgrade
+    )
+    version_change = version_before_upgrade != upgrade_version
```
clacroix12: Should we ensure that `upgrade_version > version_before_upgrade` here to stop us from trying to "upgrade" to a lower version?

vasukulkarni: +1 on that

vasukulkarni:
overall LGTM.

tests/ecosystem/upgrade/test_upgrade.py
petr-balogh: I added assert in next commit. But need to check with >= cause in case we upgrade from 4.2.1 to 4.2.2 the x.y version is still 4.2 so I don't want to fail on such upgrade.

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -129,9 +132,36 @@ def test_upgrade():
         resource_name=constants.OPERATOR_CATALOG_SOURCE_NAME,
         namespace=constants.MARKETPLACE_NAMESPACE,
     )
+    version_before_upgrade = config.ENV_DATA.get("ocs_version")
+    upgrade_version = config.UPGRADE.get(
+        "upgrade_ocs_version", version_before_upgrade
+    )
+    parsed_version_before_upgrade = parse_version(version_before_upgrade)
+    parsed_upgrade_version = parse_version(upgrade_version)
+    assert parsed_upgrade_version >= parsed_version_before_upgrade, (
```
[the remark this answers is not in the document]
petr-balogh: I answered this to Coady's question. In case of 4.2.1 upgrade to 4.2.2 we still are using same OCS version config file for 4.2 and hence OCS version is just X.Y there so we are still on 4.2 and upgrading just Z stream, so no need to fail for that case.

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -129,9 +132,36 @@ def test_upgrade():
...
+        f"Version you would like to upgrade to: {parsed_upgrade_version} "
```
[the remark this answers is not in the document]
petr-balogh: parsed_upgrade_version can be formatted as string as well, but if we have it directly in string I will use that as you suggested.

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -129,9 +132,36 @@ def test_upgrade():
...
+        f"is not higher or equal to the version you currently running: "
+        f"{parsed_version_before_upgrade}"
+    )
+    version_change = parsed_version_before_upgrade > parsed_upgrade_version
```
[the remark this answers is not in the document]
petr-balogh: Mistake, I thought I have same order like here: https://github.com/red-hat-storage/ocs-ci/pull/1496/files#diff-40ca86338b56ecd56fb458ca2ea36c3bR141

tests/ecosystem/upgrade/test_upgrade.py
clacroix12: @petr-balogh can we keep track of the full version as well as the standard version we are using now? For example:

```python
ocs_version = '4.2'
full_ocs_version = '4.2.1'

upgrade_version = '4.2'
full_upgrade_version = '4.2.2'
```
If we were able to use the full versions in this check we would be able to use `>` and still support upgrading between minor version increases. If there is a reason why we are limited here then i support sticking with `>=` for now, but I think being able to discern between these minor versions will be a big help going forward. I assume we will need to test upgrade between minor versions quite frequently as more are released.

vavuthu: As @clacroix12 suggested, we need to provide precise full_ocs_version for upgrade

petr-balogh: You can pass 4.2.1 to the --ocs-version parameter or --upgrade-ocs-version parameter but We need to have config file for each z version. So I think that approach I have is more flexible and usually we will upgrade from latest 4.2 to latest 4.3 so between latest X versions. If needed to test specific version than we will create config file wit x.y.z and than we will load it. No need to add any extra full version to the config as internal builds have csv version like 0.0.310 for example. I have opened bz for it today. I will add it to the agenda for automation mtg and we can discuss but I don't want to block this PR on it.

Another disadvantage for keeping full version in config makes impossible to test automatically on new Z build by UMB cause before we will need to update config.

petr-balogh: And one more thing. It's mainly about reload config for version and we also have configs just for 4.x via --ocs-version . In case we switch between version we just need reload proper config. 

In case you would like to use specific version that's why we have that upgrade-ocs-registry-image where you specify exact image of specific build.

Makes sense @clacroix12 @vavuthu ?

clacroix12: I don't think we were necessarily talking about having config files for each z version, but rather having some way of determining the full version that is installed as well as the one we want to upgrade to. This way we can ensure that we aren't trying to upgrade to the same version that is installed.

> I will add it to the agenda for automation mtg and we can discuss but I don't want to block this PR on it.

Agreed, I don't want to either and I think that using `>=` is totally sufficient for now.

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -103,3 +103,7 @@ ENV_DATA:
   # measurement_dir: '/tmp/ocs_ci_monitoring_measurement/'
   # Default RHCOS image to be used for VmWare deployment
   vm_template: 'rhcos-4.3.0-x86_64-vmware'
+
+# This section is related to upgrade
+UPGRADE:
+  upgrade_to_latest: true
```
fbalak: won't this affect any not upgrade related scenarios?

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -179,6 +211,9 @@ def test_upgrade():
     log.info(
         f"Waiting for CSV {csv_name_post_upgrade} to be in succeeded state"
     )
+    if version_before_upgrade == '4.2' and upgrade_version == '4.3':
+        log.info("Force creating cep toolbox after upgrade 4.2 -> 4.3")
```
fbalak:
```suggestion
        log.info("Force creating ceph toolbox after upgrade 4.2 -> 4.3")
```

petr-balogh:
Running verification here:  https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/4841/console

petr-balogh:
I see that from verification job we correctly picked:
```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  labels:
    ocs-operator-internal: 'true'
  name: ocs-catalogsource
  namespace: openshift-marketplace
spec:
  displayName: Openshift Container Storage
  icon:
    base64data: ''
    mediatype: ''
  image: quay.io/rhceph-dev/ocs-olm-operator:4.2.2-rc6
  publisher: Red Hat
  sourceType: grpc
```
and upgrading to correct latest 4.3
```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  creationTimestamp: '2020-02-21T14:01:33Z'
  generation: 1
  labels:
    ocs-operator-internal: 'true'
  name: ocs-catalogsource
  namespace: openshift-marketplace
  resourceVersion: '17980'
  selfLink: /apis/operators.coreos.com/v1alpha1/namespaces/openshift-marketplace/catalogsources/ocs-catalogsource
  uid: 2c521f2f-562d-4b49-bb6f-7b31d29a6e80
spec:
  displayName: Openshift Container Storage
  icon:
    base64data: ''
    mediatype: ''
  image: quay.io/rhceph-dev/ocs-olm-operator:4.3-323.8ec8be3.release_4.3
  publisher: Red Hat
  sourceType: grpc
status:
```

petr-balogh:
Hit issue with nooba pod numbers cause of it's changed in 4.3 . Fixing and running another verification here:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/4845/

petr-balogh:
Yay! Upgrade from 4.2 -> 4.3 passed well.  http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/pbalogh-upgr/pbalogh-upgr_20200221T171734/logs/upgrade_test_report.html
Marking as verified.

ocs_ci/framework/conf/default_config.yaml
petr-balogh: no it won't

PR #1505, opened 2020-02-21
WIP: Initial pass at adding image tag logic that works with the docker registry

clacroix12:
Infra ticket:
https://projects.engineering.redhat.com/browse/OCSQE-231

### DIFF
--- a/ocs_ci/utility/docker.py
+++ b/ocs_ci/utility/docker.py
@@ -1,0 +1,145 @@
+import logging
+
+import requests
+
+from ocs_ci.framework import config
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.exceptions import TagNotFoundException
+
+logger = logging.getLogger(__name__)
+
+
+def get_latest_ds_olm_tag_docker(upgrade=False, latest_tag=None):
+    """
+    This function returns latest tag of OCS downstream registry or one before
+    latest if upgrade parameter is True
+
+    Args:
+        upgrade (str): If True then it returns one version of the build before
+            the latest.
+        latest_tag (str): Tag of the latest build. If not specified
+            config.DEPLOYMENT['default_latest_tag'] or 'latest' will be used.
+
+    Returns:
+        str: latest tag for downstream image from docker registry
+
+    Raises:
+        TagNotFoundException: In case no tag found
+
+    """
+    manifests = get_ocs_registry_docker_image_manifests()
+
+    latest_tag = latest_tag or config.DEPLOYMENT.get(
+        'default_latest_tag', 'latest'
+    )
+    latest_image = None
+    for manifest in manifests:
+        tag = manifest['tag']
+        image = manifest['fsLayers']
+        if tag == latest_tag:
+            latest_image = image
+            break
+    if not latest_image:
+        raise TagNotFoundException("Couldn't find latest tag!")
+    latest_tag_found = False
+    for manifest in manifests:
+        tag = manifest['tag']
+        image = manifest['fsLayers']
+        if not upgrade:
+            if (
+                tag not in constants.LATEST_TAGS
+                and image == latest_image
+            ):
+                return tag
+        if upgrade:
+            if not latest_tag_found and tag == latest_tag:
+                latest_tag_found = True
+                continue
+            if latest_tag_found:
+                if (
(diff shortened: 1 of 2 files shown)


ocs_ci/utility/docker.py
```diff
@@ -0,0 +1,145 @@
...
+        if manifest['tag'] == current_tag:
+            if index < 2:
+                raise TagNotFoundException("Couldn't find tag for upgrade!")
+            current_tag_index = index
+            break
+    sliced_reversed_tags = manifests[:current_tag_index]
+    sliced_reversed_tags.reverse()
+    for manifest in sliced_reversed_tags:
+        tag = manifest['tag']
+        if tag not in constants.LATEST_TAGS and "rc" in tag:
```
petr-balogh: we will need to fix this the same as I did in : https://github.com/red-hat-storage/ocs-ci/pull/1496/files but it's mostly related to upgrade so I will fix once we merge mentioned PR ^

ocs_ci/utility/docker.py
```diff
@@ -0,0 +1,145 @@
...
+                return tag
+        if upgrade:
+            if not latest_tag_found and tag == latest_tag:
+                latest_tag_found = True
+                continue
+            if latest_tag_found:
+                if (
+                    tag not in constants.LATEST_TAGS
+                    and image != latest_image
+                    and "rc" in tag
```
petr-balogh: the same, maybe open the issue to fix this please

ocs_ci/utility/docker.py
```diff
@@ -0,0 +1,145 @@
...
+def get_ocs_registry_docker_image_manifests():
+    """
+    Retrieves ocs-registry image manifest data for all tags.
+
+    Returns:
+        list: list of manifest dicts
+
+    """
+    # retrieve auth token
+    docker_user = ''  # todo: retrieve from config, probably RUN section
```
petr-balogh: the credentials we are used to take from ./data folder - do you think that we should try to automatically read some file there so people will not have to always pass the config file?

clacroix12: As the token comes from a one-time-viewable window in the docker UI I think we will most likely need to have users create a file and place it in the `data` folder so that we can read the credentials in. This avoids having to require passing a custom config file all the time, but still requires users to do some additional setup if they are going to execute deployments using images from docker. I will go this route and if we are unable to read the credentials I will raise an exception that will point users to the doc describing how to setup their docker credentials.

clacroix12:
> otherwise looks good to me. BTW have you tried the deployment?

I have not as it's not fully functional. I sent details in an email but there are still some items left to do. I need to update our deployment/upgrade calls to the old functions to use the new ones as well as make the changes regarding docker auth. These are both straight forward and shouldn't take long so I can start testing deployment shortly.

ocs_ci/utility/docker.py
petr-balogh: Yep, this sounds good to me. Answered in mail and put all to the infra ticket:
https://projects.engineering.redhat.com/browse/OCSQE-231

petr-balogh:
> > otherwise looks good to me. BTW have you tried the deployment?
> 
> I have not as it's not fully functional. I sent details in an email but there are still some items left to do. I need to update our deployment/upgrade calls to the old functions to use the new ones as well as make the changes regarding docker auth. These are both straight forward and shouldn't take long so I can start testing deployment shortly.


The call of the functions will be based on the image url as I saw the discussion with Zack in infra room. This sounds good to me. If the params will be the same for both functions you can do something like:

```python
if "quay_url" in image:
    get_latest_tag_func = function_we_have_for_quay
elif "docker_url" in image:
    get_latest_tag_func = function_we_have_for_docker

get_latest_tag_func(param1, param2)
```

Or we can wrap it in some general function which will have call of specific function covered inside and then we will call just that.

Thanks

ocs_ci/utility/docker.py
clacroix12: Credentials are now being pulled from a new file `auth.yaml` in the `data` dir that users will need to create. Doc has been added describing how to set this file up.

clacroix12:
Closing as we no longer need to use docker hub for private repos

PR #1506, opened 2020-02-21
supported_configuration fixture

fbalak:
Add supported_configuration fixture and use it for capacity workloads

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -53,6 +53,48 @@ def pytest_logger_config(logger_config):
     logger_config.set_formatter_class(OCSLogFormatter)
 
 
+@pytest.fixture()
+def supported_configuration():
+    """
+    Check that cluster nodes have enough CPU and Memory as described in:
+    https://access.redhat.com/documentation/en-us/red_hat_openshift_container_storage/4.2/html-single/planning_your_deployment/index#infrastructure-requirements_rhocs
+    This fixture is intended as a prerequisite for tests or fixtures that
+    run flaky on configurations that don't meet minimal requirements.
+
+    Minimum requirements for each starting node (OSD+MON):
+        16 CPUs
+        64 GB memory
+    Last documentation check: 2020-02-21
+    """
+    min_cpu = 16
+    min_memory = 64 * 10**9
+
+    node_obj = ocp.OCP(kind=constants.NODE)
+    log.debug('Checking if system meets minimal requirements')
+    nodes = node_obj.get().get('items')
+    for node in nodes:
+        real_cpu = int(node.get('status').get('capacity').get('cpu'))
+        real_memory = node.get('status').get('capacity').get('memory')
+        if real_memory.endswith('Ki'):
+            real_memory = int(real_memory[0:-2]) * 2**10
+        elif real_memory.endswith('Mi'):
+            real_memory = int(real_memory[0:-2]) * 2**20
+        elif real_memory.endswith('Gi'):
+            real_memory = int(real_memory[0:-2]) * 2**30
+        elif real_memory.endswith('Ti'):
+            real_memory = int(real_memory[0:-2]) * 2**40
+        else:
+            real_memory = int(real_memory)
+
+        if (real_cpu <= min_cpu or real_memory <= min_memory):
+            pytest.xfail(
+                f"Node {node.get('metadata').get('name')} doesn't have "
+                f"minimum of required reasources for running the test:\n"
+                f"{min_cpu} CPU and {min_memory} Memory\nIt has:\n{real_cpu} "
+                f"CPU and {real_memory} Memory"
+            )
+
+
 @pytest.fixture(scope='class')
 def secret_factory_class(request):
     return secret_factory_fixture(request)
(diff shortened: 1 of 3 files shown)


fbalak:
Example logs:
```
---------------------------------------------------------- live log setup -----------------------------------------------------------
15:58:28 - MainThread - tests.conftest - INFO - All logs located at /tmp/ocs-ci-logs-1582297090
15:58:28 - MainThread - ocs_ci.deployment.factory - INFO - Deployment key = aws_ipi
15:58:28 - MainThread - ocs_ci.deployment.factory - INFO - Current deployment platform: AWS,deployment type: ipi
(...)
15:58:45 - MainThread - ocs_ci.utility.utils - INFO - Fetching latest changes from rook repository.
15:58:45 - MainThread - ocs_ci.utility.utils - INFO - Executing command: git fetch --all
15:58:48 - MainThread - ocs_ci.utility.utils - INFO - Checkout rook repository to specific branch: master
15:58:48 - MainThread - ocs_ci.utility.utils - INFO - Executing command: git checkout master
15:58:48 - MainThread - ocs_ci.utility.utils - WARNING - Command warning: Already on 'master'
15:58:48 - MainThread - ocs_ci.utility.utils - INFO - Reset branch: master with latest changes
15:58:48 - MainThread - ocs_ci.utility.utils - INFO - Executing command: git reset --hard origin/master
15:58:48 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig /tmp/kubeconfig get Node  -o yaml
XFAIL

====================================================== short test summary info ======================================================
XFAIL tests/libtest/test_requirements.py::test_requirements
  reason: Node compute-0 doesn't have minimum of required reasources for running the test:
16 CPU and 64000000000 Memory
It has:
4 CPU and 16815443968 Memory
==================================================== 1 xfailed in 29.55 seconds =====================================================
```

tests/conftest.py
```diff
@@ -53,6 +53,48 @@ def pytest_logger_config(logger_config):
...
+        64 GB memory
+    Last documentation check: 2020-02-21
+    """
+    min_cpu = 16
+    min_memory = 64 * 10**9
+
+    node_obj = ocp.OCP(kind=constants.NODE)
+    log.debug('Checking if system meets minimal requirements')
+    nodes = node_obj.get().get('items')
+    for node_info in nodes:
```
mbukatov: You should be checking only worker nodes, which are hosting storage components (OSDs or MONs).

mbukatov: When you add filtering nodes to get only storage ones, it would be useful to log by which label you do it.

tests/conftest.py
```diff
@@ -53,6 +53,48 @@ def pytest_logger_config(logger_config):
...
+        real_cpu = int(node_info.get('status').get('capacity').get('cpu'))
+        real_memory = node_info.get('status').get('capacity').get('memory')
```
mbukatov: Using `.get().get()` chains like this works, but in case it fails could only lead to misleading errors when the field is not present (None has not ...). Compared to that, unhandled `[key][key]` will at least provide information about the missing key.

tests/conftest.py
```diff
@@ -53,6 +53,48 @@ def pytest_logger_config(logger_config):
     logger_config.set_formatter_class(OCSLogFormatter)
 
 
+@pytest.fixture()
```
ebenahar: shouldn't this be a session scoped fixture? Or is the intention to use this fixture in specific tests?

tests/conftest.py
```diff
@@ -53,6 +53,48 @@ def pytest_logger_config(logger_config):
...
+def supported_configuration():
```
ebenahar: for some purposes, like automation development, we'd want to deploy clusters with less than the required minimum.

ebenahar: but if the intention is to use it only for running specific test cases, then this should not be a problem

tests/conftest.py
fbalak: This should be called just for a specific subset of tests. I am adding this for now because tests related to capacity alerts are flaky for configurations that don't meet minimal requirements. I am will add another PR that will improve logging and debugging of these tests (and other) if they have sufficient configuration.

tests/conftest.py
fbalak: done

tests/conftest.py
fbalak: done

tests/conftest.py
fbalak: This will be now part of only 2 tests and it can be added to additional tests if needed.

tests/conftest.py
```diff
@@ -71,7 +71,7 @@ def supported_configuration():
 
     node_obj = ocp.OCP(kind=constants.NODE)
     log.debug('Checking if system meets minimal requirements')
-    nodes = node_obj.get().get('items')
+    nodes = node_obj.get(selector=constants.WORKER_LABEL).get('items')
```
mbukatov: In general, I don't think that we can assume that every worker node is a storage node, but as long as our CI env is concerned, this probably holds.

tests/conftest.py
ebenahar: Thanks for the explanation @fbalak

Issue #1445, opened 2020-02-07
must-gather failing for 4.2


petr-balogh:
@clacroix12 Actually I see issue with 4.2 as we are using RC build and there is no such tag propagated to must-gather image:  https://quay.io/repository/rhceph-dev/ocs-must-gather?tag=latest&tab=tags
```
dir/openshift-cluster-dir/auth/kubeconfig adm must-gather --image=quay.io/rhceph-dev/ocs-must-gather:4.2-rc --dest-dir=/home/jenkins/current-cluster-dir/logs/failed_testcase_ocs_logs_1581068201/deployment_ocs_logs/ocs_must_gather.
Error is error: gather did not start for pod must-gather-7zd26: unable to pull image: ErrImagePull: rpc error: code = Unknown desc = Error reading manifest 4.2-rc in quay.io/rhceph-dev/ocs-must-gather: manifest unknown: manifest unknown
```

petr-balogh:
For now it will be fixed if we will fix this issue: #1446 but long term we should ask engineering to propagate our default latest tag to other images or at least to must-gather. Or we will have to have defined special tag for must-gather

mbukatov:
I run into this today, using `OCP_VERSION=4.2-ga`, `OCS_VERSION=4.2`.

petr-balogh:
Oh, I forgot about this. 
Will need to fix that. Adding to my todo list and will fix.

PR #1507, opened 2020-02-21
Fix must gather image tag

petr-balogh:
Fixes: #1445

### DIFF
--- a/conf/ocs_version/ocs-4.3.yaml
+++ b/conf/ocs_version/ocs-4.3.yaml
@@ -5,3 +5,5 @@ DEPLOYMENT:
 ENV_DATA:
   logging_version: '4.3'
   ocs_version: '4.3'
+REPORTING:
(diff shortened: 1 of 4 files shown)


conf/ocs_version/ocs-4.3.yaml
```diff
@@ -5,3 +5,5 @@ DEPLOYMENT:
 ENV_DATA:
   logging_version: '4.3'
   ocs_version: '4.3'
+REPORTING:
+  default_ocs_must_gather_latest_tag: 'latest-4.2'
```
petr-balogh: 4.3

petr-balogh:
must-gather didn't trigger during verification job but I think that it shouldn't bigger problem than now because we are failing on non existing tag anyway. So worth to merge ASAP and in case something wrong fix in other PR.

PR #1460, opened 2020-02-12
Function to create custom image locally and test to validate

keesturam:
Fixes #403

PR #1508, opened 2020-02-21
Add build image factory

keesturam:
This PR adds a factory to allow creation of custom container images. When a pod is created from pod factory with default params, a fio image is created and saved in the local registry from which fio pods will be created each time. 

Depends on #1460

### DIFF
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -234,6 +234,10 @@ CSI_RBD_POD_YAML = os.path.join(
     TEMPLATE_CSI_RBD_DIR, "pod.yaml"
 )
 
+CSI_RBD_FIO_POD_YAML = os.path.join(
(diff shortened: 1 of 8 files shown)


ocs_ci/templates/CSI/rbd/pod.yaml
```diff
@@ -6,11 +6,11 @@ metadata:
   namespace: default
 spec:
   containers:
-   - name: web-server
-     image: nginx
+   - name: fio
+     image: fio
```
ramkiperiy: instead of changing image and name in every yaml, why not change this param by getting the yaml in dict.

ebenahar:
@keesturam Would like to close old PRs, is this one still relevant?

keesturam:
closing this PR for now. 
@hnallurv @PrasadDesala Please have a look at this PR. I was trying to implement a change so that all tests would use the registry backed by OCS volume. (In case you need to take it up or look for a simpler way)

Issue #1513, opened 2020-02-24
test_nodes_maintenance[worker] fails with timeout


prsurve:
Reason for failure

stdout = b'node/ip-10-0-136-135.us-east-2.compute.internal cordoned\nevicting pod "rook-ceph-osd-prepare-ocs-deviceset-1-8-vkkr...8-658cf6c4b-d86rx"\nevicting pod "rook-ceph-osd-2-bf8c564b4-qfx9h"\nevicting pod "rook-ceph-osd-20-65757678bb-7b9hm"\n'
stderr = b'WARNING: ignoring DaemonSet-managed Pods: openshift-cluster-node-tuning-operator/tuned-ccw55, openshift-dns/dns-defa...h-osd-20-65757678bb-7b9hm" (will retry after 5s): Cannot evict pod as it would violate the pod\'s disruption budget.\n'

PR #1515, opened 2020-02-24
Add ceph health check before doing node drain

prsurve:
Fixes: #1513 

Signed-off-by: Pratik Surve <pratiksurve2130@gmail.com>

### DIFF
--- a/tests/manage/z_cluster/nodes/test_nodes_maintenance.py
+++ b/tests/manage/z_cluster/nodes/test_nodes_maintenance.py
@@ -3,8 +3,8 @@ import pytest
 
 from subprocess import TimeoutExpired
 
-from ocs_ci.framework import config
-from ocs_ci.utility.utils import ceph_health_check
+from ocs_ci.ocs.exceptions import CephHealthException
+from ocs_ci.utility.utils import ceph_health_check_base
 
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.node import (
@@ -53,6 +53,16 @@ class TestNodesMaintenance(ManageTest):
         """
         self.sanity_helpers = Sanity()
 
+    @pytest.fixture(autouse=True)
+    def health_checker(self):
+        try:
+            status = ceph_health_check_base()
+            if status:
+                logger.info("Health check passed")
+        except CephHealthException:
+            # skip because ceph is not in good health
+            pytest.skip("Ceph Health check failed")
+
     @tier1
     @pytest.mark.parametrize(
         argnames=["node_type"],
@@ -75,11 +85,6 @@ class TestNodesMaintenance(ManageTest):
         typed_nodes = get_typed_nodes(node_type=node_type, num_of_nodes=1)
         assert typed_nodes, f"Failed to find a {node_type} node for the test"
         typed_node_name = typed_nodes[0].name
-
-        # Ceph Health check
-        assert ceph_health_check(namespace=config.ENV_DATA['cluster_namespace']), (
-            "Failed with ceph status is not Health Ok"
-        )
 
         # Maintenance the node (unschedule and drain)
         drain_nodes([typed_node_name])
--- a/tests/manage/z_cluster/nodes/test_nodes_maintenance.py
+++ b/tests/manage/z_cluster/nodes/test_nodes_maintenance.py
@@ -2,6 +2,10 @@ import logging
 import pytest
 
 from subprocess import TimeoutExpired
+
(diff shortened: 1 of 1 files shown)


tests/manage/z_cluster/nodes/test_nodes_maintenance.py
```diff
@@ -72,6 +76,11 @@ def test_node_maintenance(self, node_type, pvc_factory, pod_factory):
         assert typed_nodes, f"Failed to find a {node_type} node for the test"
         typed_node_name = typed_nodes[0].name
 
+        # Ceph Health check
```
ebenahar: IMO, Ceph health should be checked during test setup.

tests/manage/z_cluster/nodes/test_nodes_maintenance.py
```diff
@@ -49,6 +53,16 @@ def init_sanity(self):
         """
         self.sanity_helpers = Sanity()
 
+    @pytest.fixture(autouse=True)
```
ebenahar: Please use the fixture from https://github.com/red-hat-storage/ocs-ci/blob/1d55e2b6e8f052773837f3de7102c9a1a1d2c602/tests/conftest.py#L853

ebenahar: Please ignore. the mentioned fixture will take place only for specific tiers.

tests/manage/z_cluster/nodes/test_nodes_maintenance.py
```diff
@@ -53,6 +53,16 @@ def init_sanity(self):
...
+    def health_checker(self):
```
clacroix12: Please add a docstring

tests/manage/z_cluster/nodes/test_nodes_maintenance.py
```diff
@@ -53,6 +53,16 @@ def init_sanity(self):
...
+        try:
+            status = ceph_health_check_base()
+            if status:
+                logger.info("Health check passed")
+        except CephHealthException:
```
clacroix12: We are losing some information here from the original exception. Since the exception message is `f"Ceph cluster health is not OK. Health: {health}"` can we simply use that as the skip message? This way when we skip this test we know what health state the cluster is in.

We can do something like this:
```python
except CephHealthException as e:
    # skip because ceph is not in good health
    pytest.skip(e)
```

prsurve: Done

tests/manage/z_cluster/nodes/test_nodes_maintenance.py
prsurve: Done

RazTamir:
Hi Pratik,
Why this ceph health check is needed for the test? 
Because of leftover from different test?

prsurve:
> Hi Pratik,
> Why this ceph health check is needed for the test?
> Because of leftover from different test?

Hi Raz 
we are adding health check here because to make drain successful we need ceph HEALTH_Ok otherwise drain will be stuck saying (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.\n' it will keep retrying

RazTamir:
@prsurve but what might cause the ceph health not to be OK?
I'm asking because,
If the test itself does something to change the health status, that's ok to have such verification.
If not, that means the previous test is doing something that changes the ceph healthiness and therefore the other test should be stabilized.
The idea behind it is that every test should not do unneeded verifications to ensure the cluster is in shape for it to run but assume it is the only test run - I hope that statement makes sense

prsurve:
>@prsurve but what might cause the ceph health not to be OK?
@RazTamir it happens because of some leftover from the previous test case or something might have made cluster not unhealthy

tests/manage/z_cluster/nodes/test_nodes_maintenance.py
```diff
@@ -49,6 +53,20 @@ def init_sanity(self):
         """
         self.sanity_helpers = Sanity()
 
+    @pytest.fixture(autouse=True)
+    def health_checker(self):
+        """
+        Check Ceph health
+
+        """
+        try:
```
fbalak: I would add:
```
logger.info('Checking Ceph health. HEALTH_OK status is a prerequisite for this test.')
```
at the beginning of the fixture so it's clear from log why it fails if Ceph is not healthy.

PR #1520, opened 2020-02-24
Vdbench add node

Avilir:
vdbench: adding new app worker node for the test.


ocs_ci/ocs/machine.py
```diff
@@ -262,6 +262,24 @@ def add_node(machine_set, count):
...
+        count (int): Count to decrease by
+    Returns:
+        bool: True if commands executes successfully
+    """
+    replica_count = get_replica_count(machine_set) - count
+    ocp = OCP(namespace=constants.OPENSHIFT_MACHINE_API_NAMESPACE)
+    ocp.exec_oc_cmd(
+        f'scale --replicas={replica_count} machinesets {machine_set}'
+    )
+    return True
```
prsurve: If provided machine_set  has more then 1 replica count it might happen that it will delete ocs node and cluster will go into degraded state

Avilir: this will decrease the number of machines in the machineses by delete the non-OCS machines.

RazTamir:
As discussed offline,

I don't see a reason to add the 4th node as this is not mimic a customer use case.
Our customers will run on OCP cluster containing 6 nodes - 3 masters, 3 workers.
OCS node will run on the 3 worker nodes along with the customer applications.
VDbench is a customer application in our scenario and therefore should run on 1 of the 3 worker nodes.

If I got something wrong here, please reply

keesturam:
> As discussed offline,
> 
> I don't see a reason to add the 4th node as this is not mimic a customer use case.
> Our customers will run on OCP cluster containing 6 nodes - 3 masters, 3 workers.
> OCS node will run on the 3 worker nodes along with the customer applications.
> VDbench is a customer application in our scenario and therefore should run on 1 of the 3 worker nodes.
> 
> If I got something wrong here, please reply

There are 2 ways applications can be run by customers. 
1) Running applications along with OCS in converged mode
2) Running applications in other nodes from OCS nodes. i.e., OCS runs on dedicated nodes 
(There are other options like running OCS on master nodes, but let's keep it aside for now)

Option 1 is typically chosen by small customers AFAIU who don't want to have a higher number of nodes. But the catch here is that the resources for `m5.4xlarge` i.e., 64 GB RAM and 32  CPU are to be made available for OCS and depending on the number of applications, more resources need to be made available. So, if we are to run performance tests on just 3 nodes, we should be using a bigger instance type. 

On the other hand, Option 2 will be the widely used model, where OCS runs on dedicated nodes with m5.4xlarge instance type and applications will run on other nodes. This allows customers to scale up app nodes when there is demand and scale down when the demand is less. This will be a widely used model and it is important we run scale and performance tests this way. 

So, running vdbench or FIO on the 4th node is a realistic scenario IMO. Please let me know if this makes sense.

tests/e2e/performance/test_vdbench_workload.py
```diff
@@ -57,6 +63,20 @@ def teardown():
     worker_nodes = helpers.get_worker_nodes()
     # Getting list of free nodes
     free_nodes = list(set(worker_nodes) - set(ocs_nodes))
+    if not free_nodes:
+        # Getting list of current machinesets
+        log.info('Adding new worker node for the application to run on.')
+        mset = machine.get_machinesets()
```
keesturam: Does this hold good for UPI or VMware configurations? If not, we need a check. I don't see any machineset defined in my vmware cluster?!
```
(.venv) [kramdoss@kramdoss ocs-ci]$ oc get machineset --all-namespaces
No resources found.
(.venv) [kramdoss@kramdoss ocs-ci]$ oc get machinesets -n openshift-machine-api
No resources found.

```

tests/e2e/performance/test_vdbench_workload.py
```diff
@@ -57,6 +63,20 @@ def teardown():
...
+        replica_count = machine.get_replica_count(mset[0])
```
keesturam: Does `mset[0]` always represent worker node?

ocs_ci/ocs/machine.py
keesturam: I think Pratik's point is true that there is a chance that one of the OCS nodes is removed. Maybe, you should create a new machine set to avoid this issue.

tests/e2e/performance/test_vdbench_workload.py
```diff
@@ -42,12 +42,18 @@ def label_nodes(request):
...
+    m_set = ''  # this will hold machine_set name that added
+
     def teardown():
         log.info('Clear label form worker (Application) nodes')
         # Getting all Application nodes
         app_nodes = machine.get_labeled_nodes(constants.APP_NODE_LABEL)
         helpers.remove_label_from_worker_node(app_nodes,
                                               constants.APP_NODE_LABEL)
+        if m_set:
+            log.info(f'Decreas {m_set} by 1')
```
keesturam: typo in decrease

ocs_ci/ocs/machine.py
keesturam: Please check - please check - https://docs.openshift.com/container-platform/4.1/machine_management/creating-machineset.html for creating new machine set.

tests/e2e/performance/test_vdbench_workload.py
Avilir: This implementation is for AWS IPI.
Yes, i will work on solution for UPI and VMware as well.

tests/e2e/performance/test_vdbench_workload.py
Avilir: > > As discussed offline,
> > I don't see a reason to add the 4th node as this is not mimic a customer use case.
> > Our customers will run on OCP cluster containing 6 nodes - 3 masters, 3 workers.
> > OCS node will run on the 3 worker nodes along with the customer applications.
> > VDbench is a customer application in our scenario and therefore should run on 1 of the 3 worker nodes.
> > If I got something wrong here, please reply
> 
> There are 2 ways applications can be run by customers.
> 
> 1. Running applications along with OCS in converged mode
> 2. Running applications in other nodes from OCS nodes. i.e., OCS runs on dedicated nodes
>    (There are other options like running OCS on master nodes, but let's keep it aside for now)
> 
> Option 1 is typically chosen by small customers AFAIU who don't want to have a higher number of nodes. But the catch here is that the resources for `m5.4xlarge` i.e., 64 GB RAM and 32 CPU are to be made available for OCS and depending on the number of applications, more resources need to be made available. So, if we are to run performance tests on just 3 nodes, we should be using a bigger instance type.
> 
> On the other hand, Option 2 will be the widely used model, where OCS runs on dedicated nodes with m5.4xlarge instance type and applications will run on other nodes. This allows customers to scale up app nodes when there is demand and scale down when the demand is less. This will be a widely used model and it is important we run scale and performance tests this way.
> 
> So, running vdbench or FIO on the 4th node is a realistic scenario IMO. Please let me know if this makes sense.

I will add a parameter for selecting if the test need to run on different node, or on the same as the OCS.

tests/e2e/performance/test_vdbench_workload.py
Avilir: fixed

ocs_ci/ocs/machine.py
```diff
@@ -262,6 +262,24 @@ def add_node(machine_set, count):
     return True
 
 
+def delete_node(machine_set, count):
+    """
+    Delete nodes from the cluster by decreasing the number of replicas in the
+    machine set.
```
ebenahar: please add an empty line between the docstring's sections

Avilir: done.

Avilir:
Closing this PR and will open new one to resolve conflicts.

PR #1528, opened 2020-02-24
Added options to skip interactive prompts during AWS cleanup

clacroix12:

### DIFF
--- a/ocs_ci/cleanup/aws/cleanup.py
+++ b/ocs_ci/cleanup/aws/cleanup.py
@@ -131,28 +131,46 @@ def aws_cleanup():
         required=False,
         help="The name of the AWS region to delete the resources from"
     )
+    parser.add_argument(
+        '--force',
+        action='store_true',
+        required=False,
+        help="Skip confirmation prompt"
+    )
+    parser.add_argument(
+        '--skip-prefixes',
+        action='store_true',
+        required=False,
+        help="Skip prompt for additional prefixes to spare"
+    )
     logging.basicConfig(level=logging.DEBUG)
     args = parser.parse_args()
 
-    confirmation = input(
-        'Careful! This action could be highly destructive. Are you sure you want to proceed? '
-    )
-    prefixes_hours = input(
-        "Press Enter if there are no cluster prefixes to spare.\n"
-        "If you would like the cleanup to spare specific cluster prefixes, "
-        "please enter them along with the time allowed for these to be kept "
-        "running, in a dictionary representation.\nAn example: "
-        "{\'prefix1\': 36, \'prefix2\': 48}\" "
-    )
-    assert confirmation == defaults.CONFIRMATION_ANSWER, "Wrong confirmation answer. Exiting"
+    if not args.force:
+        confirmation = input(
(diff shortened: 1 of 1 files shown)


clacroix12:
@ebenahar this change supports [OCSQE-221](https://projects.engineering.redhat.com/browse/OCSQE-221). The interactive prompts for confirming the cleanup as well as for entering the additional prefixes stopped this from being automated easily. This change allows us to skip those prompts and simply run the cleanup.

We might not always want to skip the prefixes in the future, so I will create an issue to take in the additional prefix map as another CLI arg instead of an interactive prompt. This way we can pass that value through from the jenkins job as well if necessary.

ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -131,28 +131,46 @@ def aws_cleanup():
...
-        "{\'prefix1\': 36, \'prefix2\': 48}\" "
-    )
-    assert confirmation == defaults.CONFIRMATION_ANSWER, "Wrong confirmation answer. Exiting"
+    if not args.force:
+        confirmation = input(
+            'Careful! This action could be highly destructive. '
+            'Are you sure you want to proceed? '
+        )
+        assert confirmation == defaults.CONFIRMATION_ANSWER, (
+            "Wrong confirmation answer. Exiting"
```
vavuthu: is it good to display what is the confirmation answer. By seeing the Question, I prefer "y" or "yes" but defaults.CONFIRMATION_ANSWER is different than what I think. 
WDYT?

ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -131,28 +131,46 @@ def aws_cleanup():
         required=False,
         help="The name of the AWS region to delete the resources from"
     )
+    parser.add_argument(
+        '--force',
+        action='store_true',
+        required=False,
+        help="Skip confirmation prompt"
```
petr-balogh: Maybe worth to mention also about destructive operation.

ocs_ci/cleanup/aws/cleanup.py
petr-balogh: Coady is just indenting this in the new condition for force param. Doesn't change the logic here. If there is issue with this it should be addressed in different PR. You can open the issue for this.

ocs_ci/cleanup/aws/cleanup.py
clacroix12: Done

ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -131,28 +131,47 @@ def aws_cleanup():
         required=False,
         help="The name of the AWS region to delete the resources from"
     )
+    parser.add_argument(
+        '--force',
+        action='store_true',
+        required=False,
+        help="Skip confirmation prompt. WARNING: this utility is destructive, "
```
shylesh: @clacroix12 ,
For me its not clear about what confirmation prompt will be skipped! Can this be more explicit ?

clacroix12: Done

PR #1534, opened 2020-02-26
Add ability to read in access token for quay private repo authentication

clacroix12:
Jira: [OCSQE-237](https://projects.engineering.redhat.com/browse/OCSQE-237)

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -21,7 +21,7 @@ There are additional prerequisites if you plan to execute AWS UPI deployments
 
 1. Install the `jq` and `awscli` system packages
 
-#### AWS UPI with RHEL workers
+##### AWS UPI with RHEL workers
 Along with AWS UPI prerequisites we need following
 
 1. openshift-dev.pem needs to be availavle to ocs-ci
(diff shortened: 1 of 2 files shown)


ocs_ci/utility/utils.py
```diff
@@ -1426,7 +1422,56 @@ def get_next_version_available_for_upgrade(current_tag):
...
+
+def get_ocs_olm_operator_tags(limit=100):
+    """
+    Query the OCS OLM Operator repo and retrieve a list of tags.
+
+    Args:
+        limit: the number of tags to limit the request to
+
+    Raises:
+        requests.RequestException: if the response return code is not ok
```
dahorak: `FileNotFoundError` and `KeyError` exceptions should be also listed in this doc string.

ocs_ci/utility/utils.py
```diff
@@ -1426,7 +1422,56 @@ def get_next_version_available_for_upgrade(current_tag):
...
+
+    Returns:
+        requests.request: response of the request
```
dahorak: Wouldn't be better to return directly the list of tags, instead of the whole response object? (like:
```python
return resp.json()['tags']
```
... with possibility to better handle any other exception there)?

ocs_ci/utility/utils.py
clacroix12: Done

ocs_ci/utility/utils.py
clacroix12: I originally didn't want to change any of the logic and just wanted to cut down on duplication but this is a very small change and makes sense. Done!

docs/getting_started.md
```diff
@@ -122,6 +135,28 @@ If you don't want to use the shared key, you can change this value to
...
+Create a file under `ocs-ci/data/` named `auth.yaml`.
+
+#### Quay
+
+To authenticate with quay you will need to have an access token. You can
+generate one yourself by following [the API doc](https://docs.quay.io/api/) or
+you may use the one QE has generated already. Ask people on ocs-qe mailing list
+or chat room if you don't know where to find the access token.
+
+To enable ocs-ci to use this token, add the following to your `auth.yaml`:
```
petr-balogh: Can you please share the token on some of mojo page like we have for pull secret? And add link here?  You can do in different PR if you would like to merge this as you have approvals.

clacroix12: @petr-balogh We have the `auth.yaml` in our ecosystem private files directory as well as the token being documented in our passwords file with the other quay information. I can also put this in a mojo page if it will help users, but I was trying to avoid putting it in a publicly available place as for now it's legitimately _my_ token.

Issue #1415, opened 2020-01-31
move out the code to create/configure logging from 'test_openshift-logging.py' so other tests can call and configure logging


hnallurv:
moving the code to create/configure logging from 'test_openshift-logging.py' to ocs_ci/utility/deployment_openshift_logging will help in calling this function from other tests which need logging to be configured as part of entry criteria.

Shrivaibavi:
This issue is addressed in #1558 and got merged. Hence closing

PR #1558, opened 2020-02-28
Moving deployment of logging to conftest.py

Shrivaibavi:


Shrivaibavi:
Fixes #1415

tests/conftest.py
```diff
@@ -1505,3 +1508,59 @@ def default_storageclasses(request, teardown_factory_session):
         teardown_factory_session(sc)
         scs[constants.RECLAIM_POLICY_RETAIN].append(sc)
     return scs
+
+
+@pytest.fixture(scope='class')
```
fbalak: Why is there `scope='class'`? Shouldn't it be `session`?

PR #1550, opened 2020-02-27
add capacity 

annagitel:
Signed-off-by: annagitel <asandler@redhat.com>

PR #1559, opened 2020-02-28
WIP: Helpers functions for scale testing

ramkiperiy:
These functions will help in understanding the cluster env and scale the pods further.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -11,21 +11,20 @@ import tempfile
 import time
 import yaml
 import threading
+import random
 
 from ocs_ci.ocs.ocp import OCP
-
 from uuid import uuid4
 from ocs_ci.ocs.exceptions import (
-    TimeoutExpiredError,
-    UnexpectedBehaviour,
-    UnavailableBuildException
+    TimeoutExpiredError, UnexpectedBehaviour,
+    UnavailableBuildException, UnavailableResourceException,
+    CommandFailed, ResourceWrongStatusException
 )
 from concurrent.futures import ThreadPoolExecutor
-from ocs_ci.ocs import constants, defaults, ocp, node
+from ocs_ci.ocs import constants, defaults, ocp, node, machine, cluster
 from ocs_ci.utility import templating
-from ocs_ci.ocs.resources import pod, pvc
+from ocs_ci.ocs.resources import pod, pvc, storage_cluster
 from ocs_ci.ocs.resources.ocs import OCS
-from ocs_ci.ocs.exceptions import CommandFailed, ResourceWrongStatusException
 from ocs_ci.utility.retry import retry
 from ocs_ci.utility.utils import TimeoutSampler, ocsci_log_path, run_cmd
 from ocs_ci.framework import config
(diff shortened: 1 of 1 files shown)


tests/helpers.py
```diff
@@ -2056,3 +2057,200 @@ def modify_osd_replica_count(resource_name, replica_count):
...
+
+    """
+    # Make sure setup has required number of OSDs
+    actual_osd_count = cluster.count_cluster_osd()
+    ocs_nodes = node.get_typed_nodes(node_type='worker')
+    expected_osd_count = len(ocs_nodes) * total_osd_nos
+    if actual_osd_count == expected_osd_count:
+        logging.info(f"Setup has OSD count as per OCS workers")
+    else:
+        storage_cluster.add_capacity(int((expected_osd_count - actual_osd_count) / 3))
```
keesturam: There is no check to see if the OSD count after adding OSDs.

tests/helpers.py
```diff
@@ -2056,3 +2057,200 @@ def modify_osd_replica_count(resource_name, replica_count):
     params = f'{{"spec": {{"replicas": {replica_count}}}}}'
     resource_name = '-'.join(resource_name.split('-')[0:4])
     return ocp_obj.patch(resource_name=resource_name, params=params)
+
+
+def check_required_osd_count(total_osd_nos=3):
```
keesturam: can you rename the function to `add_n_osds` or something similar? You are doing more than `check` here.

tests/helpers.py
```diff
@@ -2056,3 +2057,200 @@ def modify_osd_replica_count(resource_name, replica_count):
...
+    # Make sure setup has required number of OSDs
+    actual_osd_count = cluster.count_cluster_osd()
+    ocs_nodes = node.get_typed_nodes(node_type='worker')
+    expected_osd_count = len(ocs_nodes) * total_osd_nos
+    if actual_osd_count == expected_osd_count:
+        logging.info(f"Setup has OSD count as per OCS workers")
+    else:
+        storage_cluster.add_capacity(int((expected_osd_count - actual_osd_count) / 3))
+        logging.info(f"Now setup has expected osd count {expected_osd_count}")
+    return cluster.count_cluster_osd()
```
keesturam: The intention of this function seems to add osds, so why do you want to return the count. You should expect this count to be there or else the function should assert. wdyt?

tests/helpers.py
ramkiperiy: add_capacity function has the validation, also it waits for the OSD to be up and running. PR #1550

tests/helpers.py
ramkiperiy: ack

tests/helpers.py
ramkiperiy: Since it's library function added return. If test case expects a particular number of OSD to be in cluster then it can be checked with the return value after adding new osd's.

ramkiperiy:
Closing this PR, Integrated scale helper functions in class format which is part of https://github.com/red-hat-storage/ocs-ci/pull/1976

Issue #1567, opened 2020-03-02
aws-cleanup doesn't download openshift installer


clacroix12:
The aws-cleanup entry point does not download the openshift installer. This presents an issue with our cron job that will periodically call this to cleanup resources in aws. We need to download the installer during the `aws_cleanup` function so this may run on jenkins slaves.

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib64/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/jenkins/workspace/qe-aws-cleanup/ocs-ci/ocs_ci/cleanup/aws/cleanup.py", line 38, in cleanup
    run_cmd(f"{oc_bin} destroy cluster --dir {cleanup_path} --log-level=debug")
  File "/home/jenkins/workspace/qe-aws-cleanup/ocs-ci/ocs_ci/utility/utils.py", line 412, in run_cmd
    **kwargs
  File "/usr/lib64/python3.6/subprocess.py", line 423, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib64/python3.6/subprocess.py", line 729, in __init__
    restore_signals, start_new_session)
  File "/usr/lib64/python3.6/subprocess.py", line 1364, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: './bin/openshift-install': './bin/openshift-install'
```

PR #1568, opened 2020-03-02
Download openshift installer before destroying clusters

clacroix12:
Fixes: #1567 

### DIFF
--- a/ocs_ci/cleanup/aws/cleanup.py
+++ b/ocs_ci/cleanup/aws/cleanup.py
@@ -7,7 +7,7 @@ import os
 
 from ocs_ci.framework import config
 from ocs_ci.ocs.constants import CLEANUP_YAML, TEMPLATE_CLEANUP_DIR
-from ocs_ci.utility.utils import run_cmd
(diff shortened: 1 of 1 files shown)


ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -183,6 +183,8 @@ def aws_cleanup():
 
     if not clusters_to_delete:
         logger.info("No clusters to delete")
+    else:
+        get_openshift_installer()
```
shylesh: @clacroix12 ,
Will there be any scenario where cluster is bit old and openshift-installer is not compatible ? Just a thought. Otherwise LGTM.

clacroix12: Potentially, but I think that's more a question for the openshift devs. If the latest installer won't work for some versions of 4.x then we should know about that and try to do some analysis of the cluster before destroying. Currently though, we don't do any such thing and this change is just to ensure that we have the openshift installer ready for the destroy command we try to execute. If we do hit any issues like that then we can open a ticket and come up with a fix.

PR #1416, opened 2020-02-01
Add tests for KNIP-677: OCS support for Automated fencing/unfencing RWO PV

sidhant-agrawal:
Add tests/manage/pv_services/test_rwo_pvc_fencing_unfencing.py: TCs  for KNIP-677: OCS support for Automated fencing/unfencing RWO PV

Signed-off-by: Sidhant Agrawal <sagrawal@redhat.com>

PR #1580, opened 2020-03-04
Function to make node network down

sidhant-agrawal:
This function will help to bring the node network interface down, making the node unresponsive

Creating this separate PR from #1416 because it is needed by other node failure and node replacement tests

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -2,6 +2,7 @@ import copy
 import logging
 import re
 
+from subprocess import TimeoutExpired
 from ocs_ci.framework import config
 from ocs_ci.ocs.exceptions import TimeoutExpiredError
 from ocs_ci.ocs.ocp import OCP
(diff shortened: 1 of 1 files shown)


sidhant-agrawal:
@PrasadDesala fyi

ocs_ci/ocs/node.py
```diff
@@ -321,3 +322,36 @@ def get_node_resource_utilization(nodename=None, node_type='worker'):
...
+                node=node_name, cmd_list=[fail_nw_cmd], timeout=15
+            )
+        except TimeoutExpired:
+            pass
+
+    if wait:
+        wait_for_nodes_status(
+            node_names=node_names, status=constants.NODE_NOT_READY
+        )
+    return True
```
RazTamir: In case you will not return True, it will return None by default which is ok for calling this with assert or in 'if' statement but I suggest to add the 'return False' just to make it more readable

sidhant-agrawal: @RazTamir  I'm expecting a  'CommandFailed' exception from ocp.exec_oc_debug_cmd() or 'ResourceWrongStatusException' from wait_for_nodes_status() will be raised in case something goes wrong, Otherwise the function will return True
Should we handle all exceptions and just return False

RazTamir: ok. thanks for clarifying

sidhant-agrawal:
Rebased and resolved conflicts

ocs_ci/ocs/node.py
```diff
@@ -359,3 +360,36 @@ def get_node_resource_utilization_from_oc_describe(nodename=None, node_type='wor
...
+        wait (bool): True in case wait for status is needed, False otherwise
+
+    Returns:
+        bool: True if node network fail is successful
+    """
+    if not isinstance(node_names, list):
+        node_names = [node_names]
+
+    ocp = OCP(kind='node')
+    fail_nw_cmd = "ifconfig $(route | grep default | awk '{print $(NF)}') down"
```
vasukulkarni: hopefully this works fine across platforms

PR #1593, opened 2020-03-04
Fixes: #1594 - Enable FIPS

ebenahar:

### DIFF
--- a/conf/ocsci/fips.yaml
+++ b/conf/ocsci/fips.yaml
@@ -1,0 +1,3 @@
+---
+ENV_DATA:
+  fips: 'true'
--- a/ocs_ci/templates/ocp-deployment/install-config-aws-ipi.yaml.j2
+++ b/ocs_ci/templates/ocp-deployment/install-config-aws-ipi.yaml.j2
@@ -35,6 +35,9 @@ networking:
   networkType: OpenShiftSDN
   serviceNetwork:
     - 172.30.0.0/16
+{% if fips %}
+fips: {{ fips | default('false') }}
+{% endif %}
 platform:
   aws:
     region: {{ region | default('us-east-2') }}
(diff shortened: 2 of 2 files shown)


clacroix12:
Please add `Fixes: #1594` to description but otherwise LGTM

conf/ocsci/fips.yaml
```diff
@@ -0,0 +1,3 @@
+---
+ENV_DATA:
+  fips: 'true'
```
petr-balogh:
```suggestion
  fips: true
```
use boolean value and not string

ocs_ci/templates/ocp-deployment/install-config-aws-ipi.yaml.j2
```diff
@@ -35,6 +35,9 @@ networking:
   networkType: OpenShiftSDN
   serviceNetwork:
     - 172.30.0.0/16
+{% if fips %}
+fips: {{ fips | default('false') }}
```
petr-balogh:
```suggestion
fips: {{ fips }}
```
no need for default value as if not defined we don't  add anything

conf/ocsci/fips.yaml
ebenahar: this is what I tried before and ended up the upper case with fips: True 
See the previous commit

ocs_ci/templates/ocp-deployment/install-config-aws-ipi.yaml.j2
ebenahar: ack. Will fix

ebenahar:
Can't verify with FIPS enabled due to https://bugzilla.redhat.com/show_bug.cgi?id=1807225. The deployment with FIPS disabled passes as seen above

Issue #1490, opened 2020-02-17
AWS cleanup script


RazTamir:
The AWS cleanup script should have the capability to do the following:
- [x] Remove instances after X hours
- [x] Ignore instances by name prefix
- [x] Avoid deleting instances from unneeded regions 
- [x] treat different prefixes with different uptime hours - with a dictionary of prefixes 
   * a prefix of 'lgvy' for longevity, will remove the instances after 48 hours
   * a prefix of 'dnd' for Do Not Delete, will completely ignore for deletion
   * a prefix for 'scl' for scale will delete after 36 hours 
.... and so on
- [ ] Ability to delete leftovers of resources like volumes and snapshots that are not associated to a cluster
- [ ] Add aws-cleanup usage to OCS CI docs

vasukulkarni:
I think we should not delete longevity and scale suites as well, they can take more time and it would be costly to delete them halfway.

ebenahar:
AWS cleanup addition was done in https://github.com/red-hat-storage/ocs-ci/pull/1458 and  https://github.com/red-hat-storage/ocs-ci/pull/1467

PR https://github.com/red-hat-storage/ocs-ci/pull/1481 will enable the deletion of spared clusters after a larger amount of hours.

RazTamir:
> I think we should not delete longevity and scale suites as well, they can take more time and it would be costly to delete them halfway.

I agree. 
To ensure they will get deleted eventually, maybe we can set the longevity runtime to the maximum time it is currently running + an extra few hours

ebenahar:
Updated https://github.com/red-hat-storage/ocs-ci/pull/1481 to handle treat different prefixes with different uptime hours - with a dictionary of prefixes

clacroix12:
@ebenahar do we have any docs on usage for `aws-cleanup` other than what is in `aws-cleanup --help`? I'm specifically looking for what to expect as far as output goes. What log files are generated and where are they placed?

ebenahar:
@clacroix12 right. Added this to the check list above

clacroix12:
@ebenahar I was going to open another issue but I figure it might just be worth to add to this list since we are already tracking it. It would be useful to convert the interactive prompt for the prefixes to spare into a CLI argument. This way we can easily provide a value from jenkins as well as not making automation of this command difficult due to interactive prompts.

PR #1595, opened 2020-03-05
AWS cleanup enhancements

clacroix12:
Addressing enhancements brought up in #1490 as well as some other small fixes.

### DIFF
--- a/ocs_ci/cleanup/aws/cleanup.py
+++ b/ocs_ci/cleanup/aws/cleanup.py
@@ -127,9 +127,11 @@ def aws_cleanup():
         nargs=1,
         action='append',
         required=True,
-        help="Maximum running time of the cluster (in hours). "
-             "Clusters older than this will be deleted. "
-             "The minimum is 10 hours"
+        help="""
+            Maximum running time of the cluster (in hours).
+            Clusters older than this will be deleted.
+            The minimum is 10 hours
+            """
     )
     parser.add_argument(
         '--region',
@@ -142,20 +144,24 @@ def aws_cleanup():
         '--prefix',
         action='append',
         required=False,
-        help="Additional prefix:hour combo to treat as a special rule. "
-             "Clusters starting with this prefix will only be cleaned up if "
-             "their runtime exceeds the provided hour(this takes precedence"
-             "over the value provided to --hours). "
-             "Example: --prefix foo:24 --prefix bar:48"
+        help="""
+            Additional prefix:hour combo to treat as a special rule.
+            Clusters starting with this prefix will only be cleaned up if
+            their runtime exceeds the provided hour(this takes precedence
+            over the value provided to --hours).
+            Example: --prefix foo:24 --prefix bar:48
+            """
     )
     parser.add_argument(
         '--force',
         action='store_true',
         required=False,
-        help="Force cluster cleanup. "
-             "User will not be prompted for confirmation. "
-             "WARNING: this utility is destructive, only use this option if "
-             "you know what you are doing."
+        help="""
+            Force cluster cleanup.
+            User will not be prompted for confirmation.
+            WARNING: this utility is destructive, only use this option if
+            you know what you are doing.
+            """
     )
     logging.basicConfig(level=logging.DEBUG)
     args = parser.parse_args()
--- a/ocs_ci/cleanup/aws/cleanup.py
+++ b/ocs_ci/cleanup/aws/cleanup.py
@@ -83,7 +83,7 @@ def get_clusters_to_delete(time_to_delete, region_name, prefixes_hours_to_spare)
             allowed_running_time = time_to_delete
             if instance.state["Name"] == "running":
                 for prefix, hours in prefixes_hours_to_spare.items():
-                    if prefix in cluster_name:
+                    if cluster_name.startswith(prefix):
                         allowed_running_time = int(hours) * 60 * 60
                         break
                 launch_time = instance.launch_time
--- a/ocs_ci/cleanup/aws/cleanup.py
+++ b/ocs_ci/cleanup/aws/cleanup.py
@@ -118,15 +118,18 @@ def cluster_cleanup():
 
 
 def aws_cleanup():
-    parser = argparse.ArgumentParser(description='AWS overall resources cleanup according to running time')
+    parser = argparse.ArgumentParser(
+        description='AWS overall resources cleanup according to running time'
+    )
     parser.add_argument(
         '--hours',
         type=int,
         nargs=1,
         action='append',
         required=True,
-        help="Maximum running time of the cluster (in hours). Clusters older than this "
-             "will be deleted. The minimum is 10 hours"
+        help="Maximum running time of the cluster (in hours). "
+             "Clusters older than this will be deleted. "
+             "The minimum is 10 hours"
     )
     parser.add_argument(
         '--region',
@@ -136,6 +139,16 @@ def aws_cleanup():
         help="The name of the AWS region to delete the resources from"
     )
     parser.add_argument(
+        '--prefix',
+        action='append',
+        required=False,
+        help="Additional prefix:hour combo to treat as a special rule. "
+             "Clusters starting with this prefix will only be cleaned up if "
+             "their runtime exceeds the provided hour(this takes precedence"
+             "over the value provided to --hours). "
+             "Example: --prefix foo:24 --prefix bar:48"
+    )
+    parser.add_argument(
         '--force',
         action='store_true',
         required=False,
@@ -143,12 +156,6 @@ def aws_cleanup():
              "User will not be prompted for confirmation. "
              "WARNING: this utility is destructive, only use this option if "
              "you know what you are doing."
-    )
-    parser.add_argument(
-        '--skip-prefixes',
-        action='store_true',
-        required=False,
-        help="Skip prompt for additional prefixes to spare"
     )
     logging.basicConfig(level=logging.DEBUG)
     args = parser.parse_args()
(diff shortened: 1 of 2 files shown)


clacroix12:
ocs4-jenkins MR: https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/merge_requests/194

ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -131,6 +138,16 @@ def aws_cleanup():
         required=False,
         help="The name of the AWS region to delete the resources from"
     )
+    parser.add_argument(
+        '--prefix',
+        action='append',
+        required=False,
+        help="Additional prefix:hour combo to treat as a special rule. "
```
petr-balogh: what about to use multi line  string with """ ?

ocs_ci/cleanup/aws/defaults.py
```diff
@@ -2,7 +2,7 @@
 Defaults module for AWS cleanup
 """
 AWS_REGION = 'us-east-2'
-CLUSTER_PREFIXES_TO_EXCLUDE_FROM_DELETION = {'jnk': '36'}
-MINIMUM_CLUSTER_RUNNING_TIME_FOR_DELETION = 10
+CLUSTER_PREFIXES_SPECIAL_RULES = {'jnk': 36, 'dnd': 36}
```
petr-balogh: DND is stand for do not delete so I am curious if this should be really treated like we don't want to delete such resources at all.  @RazTamir @ebenahar  thoughts?

ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -114,15 +118,18 @@ def cluster_cleanup():
...
+    )
     parser.add_argument(
         '--hours',
         type=int,
         nargs=1,
         action='append',
         required=True,
-        help="Maximum running time of the cluster (in hours). Clusters older than this "
-             "will be deleted. The minimum is 10 hours"
+        help="Maximum running time of the cluster (in hours). "
```
petr-balogh: same multi line string with """

ocs_ci/cleanup/aws/defaults.py
dahorak: The prefix `dnd` suggest *do not delete*... so *36* hours is quite short time IMHO. What about something like the `never` value mentioned above?

ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -158,20 +169,20 @@ def aws_cleanup():
...
-        prefixes_hours = input(
-            "Press Enter if there are no cluster prefixes to spare.\n"
-            "If you would like the cleanup to spare specific cluster prefixes, "
-            "please enter them along with the time allowed for these to be kept "
-            "running, in a dictionary representation.\nAn example: "
-            "{\'prefix1\': 36, \'prefix2\': 48}\" "
+    for prefix in args.prefix:
+        _prefix, _hours = prefix.split(':')
+        if not _prefix or not _hours:
+            raise
```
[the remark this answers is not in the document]
dahorak: Additionally it will be good to check the provided values slightly more better - this check only catch following situation: `--prefix foo:` or `--prefix :10` or `--prefix :`, but lot's of other type of values will cause unprocessed exception:
* `--prefix foo`
* `--prefix foo:bar`
* `--prefix foo:bar:123`

ocs_ci/cleanup/aws/defaults.py
ebenahar: I think LR (long running) is better wording

petr-balogh: we can have more such prefixes for dnd we can have some high number like 1000, for `LR` actually `lr` we can have other timeframe defined like 10 days for example? Or not sure what long running time frame should be?

I guess we cannot use capital letter as it's not propagated to AWS instance names and DNS record so I would stick with small letters . 

Also not sure if cluster name can start with the number, we can have defined later on also something like s5d - Stay 5 days.

ebenahar: Also not sure if cluster name can start with the number, we can have defined later on also something like s5d - Stay 5 days. -> this sounds like the best option. We will need also to read this name and translate it to the number of hours this cluster should remain running.

clacroix12: @ebenahar all we need to do is include the rule in this dict, we won't need to do any evaluation of the prefix past checking if the cluster name starts with it. I think using `lr` rather than `s` is cleared as well, and we can incorporate the number of days that we wish to keep this long running cluster up. For example:

```python
RULES = {'jnk': 36, 'DND': never, 'LR1': 24, 'LR2': 48, 'LR3': 72, 'LR4': 96, 'LR5': 120}

File: ocs_ci/cleanup/aws/cleanup.py (hunk 1 above)
Comment (clacroix12) [reply ocs_ci/cleanup/aws/cleanup.py]: Done

File: ocs_ci/cleanup/aws/cleanup.py (hunk 3 above)
Comment (clacroix12) [reply ocs_ci/cleanup/aws/cleanup.py]: Done

File: ocs_ci/cleanup/aws/cleanup.py [hunk 5]
```diff
@@ -131,6 +138,16 @@ def aws_cleanup():
...
+    parser.add_argument(
+        '--prefix',
+        action='append',
+        required=False,
+        help="Additional prefix:hour combo to treat as a special rule. "
+             "Clusters starting with this prefix will only be cleaned up if "
+             "their runtime exceeds the provided hour(this takes precedence"
+             "over the value provided to --hours). "
+             "Example: --prefix foo:24 --prefix bar:48"
+    )
```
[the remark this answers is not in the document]
clacroix12: I agree and added support for this

ocs_ci/cleanup/aws/cleanup.py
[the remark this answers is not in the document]
clacroix12: I got interrupted while trying to determine what error this should raise and forgot to add it in :) This check here has been moved to the validation function argparse uses for the type as you suggested.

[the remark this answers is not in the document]
clacroix12: Done! Using the type param and a custom validation function was a very good suggestion.

ocs_ci/cleanup/aws/defaults.py
clacroix12: Done!

ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -132,21 +153,29 @@ def aws_cleanup():
...
-             "WARNING: this utility is destructive, only use this option if "
-             "you know what you are doing."
+        type=prefix_hour_mapping,
+        help="""
+            Additional prefix:hour combo to treat as a special rule.
+            Clusters starting with this prefix will only be cleaned up if
+            their runtime exceeds the provided hour(this takes precedence
+            over the value provided to --hours).
+            Example: --prefix foo:24 --prefix bar:48
+            """
```
dahorak: I would mention the `never` keyword also here in the help for the `--prefix` argument.

ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -158,20 +187,18 @@ def aws_cleanup():
...
-            "running, in a dictionary representation.\nAn example: "
-            "{\'prefix1\': 36, \'prefix2\': 48}\" "
-        )
-        if prefixes_hours:
-            prefixes_hours_to_spare = eval(prefixes_hours)
     time_to_delete = args.hours[0][0]
-    assert time_to_delete > defaults.MINIMUM_CLUSTER_RUNNING_TIME_FOR_DELETION, (
+    assert time_to_delete > defaults.MINIMUM_CLUSTER_RUNNING_TIME, (
         "Number of hours is lower than the required minimum. Exiting"
     )
```
dahorak: I would also move the check for minimum number of hours to the `type` parameter of `parser.add_argument('--hours',...)`. It will be slightly more consistent and fails the command immediately and with better error. (Now it fails even after the confirmation question *Careful! This action could be highly destructive. Are you sure you want to proceed?* and it fails directly with full traceback - not only the clean error message.

clacroix12: Done

ocs_ci/cleanup/aws/cleanup.py
clacroix12: Done

clacroix12:
Example output for new hours validation:

```
$ aws-cleanup --hours AA --region us-west-1 --force
usage: aws-cleanup [-h] --hours HOURS [--region REGION] [--prefix PREFIX]
                   [--force]
aws-cleanup: error: argument --hours: AA is not an int, please provide an int value

$ aws-cleanup --hours 9 --region us-west-1 --force
usage: aws-cleanup [-h] --hours HOURS [--region REGION] [--prefix PREFIX]
                   [--force]
aws-cleanup: error: argument --hours: Number of hours (9) is lower than the required minimum (10).
```

ocs_ci/cleanup/aws/defaults.py
```diff
@@ -2,7 +2,10 @@
 Defaults module for AWS cleanup
 """
 AWS_REGION = 'us-east-2'
-CLUSTER_PREFIXES_TO_EXCLUDE_FROM_DELETION = {'jnk': '36'}
-MINIMUM_CLUSTER_RUNNING_TIME_FOR_DELETION = 10
+CLUSTER_PREFIXES_SPECIAL_RULES = {
+    'jnk': 36, 'DND': 'never', 'LR1': 24, 'LR2': 48, 'LR3': 72, 'LR4': 96,
```
vasukulkarni: even LR1 could have been in lower case? to keep it consistent

petr-balogh: I pointed this in my comment: https://github.com/red-hat-storage/ocs-ci/pull/1595#discussion_r388837693

I think we should have the lowercases here or we need to do the check for prefixes with lowercase of those keys.

vasukulkarni: I see it, +1 on that or lowercase match to ignore case.

clacroix12: The prefix check is done case-insensitive now so functionally it doesn't matter. I figured `LR` was easier to read than `lr`

clacroix12: https://github.com/red-hat-storage/ocs-ci/pull/1595/files#diff-3cab31ec2f837c8a4ce76ffd7a784444R88-R89

clacroix12:
Verified here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-aws-cleanup/40/

PR #1597, opened 2020-03-05
WIP: Scale Test case-Polarion ID: 1357

ramkiperiy:
Scale test case to add PVC + POD in multiple projects.

### DIFF
--- a/tests/e2e/scale/test_scale_pvc_pod_multi_project.py
+++ b/tests/e2e/scale/test_scale_pvc_pod_multi_project.py
@@ -23,7 +23,6 @@ log = logging.getLogger(__name__)
 
 
 @scale
-@ignore_leftovers
 @pytest.mark.parametrize(
     argnames="start_io",
     argvalues=[
@@ -49,9 +48,6 @@ class TestMultiProjectScalePVCPOD(E2ETest):
     def test_multi_project_scale_pvcs_pods(self, setup_fixture, start_io):
         """
         Test case to scale PVC+POD with multi projects and reach expected PVC count
-        :param setup_fixture:
-        :param start_io:
-        :return:
         """
 
         # Set PVC count based on number of OCS workers
--- a/tests/e2e/scale/test_scale_pvc_pod_multi_project.py
+++ b/tests/e2e/scale/test_scale_pvc_pod_multi_project.py
@@ -17,7 +17,7 @@ from tests import helpers
 from ocs_ci.ocs import constants, cluster, machine, node
 from ocs_ci.ocs.resources import pod, pvc
 from ocs_ci.ocs.exceptions import UnexpectedBehaviour, CephHealthException
-from ocs_ci.framework.testlib import scale, E2ETest, ignore_leftovers
+from ocs_ci.framework.testlib import scale, E2ETest
 
 log = logging.getLogger(__name__)
 
--- a/tests/e2e/scale/test_scale_pvc_pod_multi_project.py
+++ b/tests/e2e/scale/test_scale_pvc_pod_multi_project.py
@@ -1,0 +1,145 @@
+"""
+Testcase to Create 200 PVC+POD scale in one namespace with and without IO
+This will be Scale upto 1500 PVC+POD
+Scale PVC and POD Configuration
+---------------------------------------------------------
+Number of PVC+POD Creation: 1500
+PVC Type: All 4
+POD Type: Nginx basic configuration, no DC POD
+Cluster Capacity to be Utilized: 80%
+Network or IO saturation: 80%
+"""
+import logging
+import pytest
+import random
+
+from tests import helpers
+from ocs_ci.ocs import constants, cluster, machine, node
+from ocs_ci.ocs.resources import pod, pvc
+from ocs_ci.ocs.exceptions import UnexpectedBehaviour, CephHealthException
+from ocs_ci.framework.testlib import scale, E2ETest, ignore_leftovers
+
+log = logging.getLogger(__name__)
+
+
+@scale
+@ignore_leftovers
+@pytest.mark.parametrize(
+    argnames="start_io",
+    argvalues=[
+        pytest.param(
+            *[False], marks=pytest.mark.polarion_id("OCS-1357")
+        ),
+        pytest.param(
+            *[True], marks=pytest.mark.polarion_id("OCS-1357")
+        )
+    ]
+)
+class TestMultiProjectScalePVCPOD(E2ETest):
+    """
+    Scale the OCS cluster to reach 1500 PVC+POD
+    """
+    @pytest.fixture()
+    def setup_fixture(self, request):
+        def finalizer():
+            self.cleanup()
+
+        request.addfinalizer(finalizer)
+
+    def test_multi_project_scale_pvcs_pods(self, setup_fixture, start_io):
+        """
+        Test case to scale PVC+POD with multi projects and reach expected PVC count
+        :param setup_fixture:
+        :param start_io:
+        :return:
+        """
+
+        # Set PVC count based on number of OCS workers
+        ocs_nodes = node.get_typed_nodes(node_type='worker')
+        pvc_count = {3: 1500, 6: 3000, 9: 4500}
+        scale_count = pvc_count[len(ocs_nodes)]
+
+        # Other variables
+        rbd_sc_obj = helpers.default_storage_class(constants.CEPHBLOCKPOOL)
+        cephfs_sc_obj = helpers.default_storage_class(constants.CEPHFILESYSTEM)
+        pvc_count_each_itr = 5
+        size = f"{random.randrange(5, 105, 5)}Gi"
+        fio_size_param = '2G'
+        fio_rate_param = '16k'
+        self.namespace_list, all_pod_obj = ([] for i in range(2))
+
+        # Pre-requisite check for number of OSD's and app worker nodes.
+        helpers.add_required_osd_count(total_osd_nos=3)
+
+        # Create machineset for app worker nodes, which will create one app worker node
+        self.ms_name = machine.create_custom_machineset(instance_type='m5.4xlarge', zone='a')
+        machine.wait_for_new_node_to_be_ready(self.ms_name)
+        self.app_worker_nodes = machine.get_machine_from_machineset(self.ms_name)
+
+        # Create namespace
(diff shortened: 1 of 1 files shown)


ramkiperiy:
Depends on #1559 & #1550

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
```diff
@@ -0,0 +1,145 @@
...
+from ocs_ci.ocs import constants, cluster, machine, node
+from ocs_ci.ocs.resources import pod, pvc
+from ocs_ci.ocs.exceptions import UnexpectedBehaviour, CephHealthException
+from ocs_ci.framework.testlib import scale, E2ETest, ignore_leftovers
+
+log = logging.getLogger(__name__)
+
+
+@scale
+@ignore_leftovers
```
RazTamir: why?

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
```diff
@@ -0,0 +1,145 @@
...
+    def setup_fixture(self, request):
+        def finalizer():
+            self.cleanup()
+
+        request.addfinalizer(finalizer)
+
+    def test_multi_project_scale_pvcs_pods(self, setup_fixture, start_io):
+        """
+        Test case to scale PVC+POD with multi projects and reach expected PVC count
+        :param setup_fixture:
```
RazTamir: this is not the doc style we are using

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
```diff
@@ -0,0 +1,145 @@
...
+        size = f"{random.randrange(5, 105, 5)}Gi"
+        fio_size_param = '2G'
+        fio_rate_param = '16k'
+        self.namespace_list, all_pod_obj = ([] for i in range(2))
+
+        # Pre-requisite check for number of OSD's and app worker nodes.
+        helpers.add_required_osd_count(total_osd_nos=3)
+
+        # Create machineset for app worker nodes, which will create one app worker node
+        self.ms_name = machine.create_custom_machineset(instance_type='m5.4xlarge', zone='a')
```
RazTamir: please don't hard code the instance type and zone.
Will it work on VMware?

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
```diff
@@ -0,0 +1,145 @@
...
+        while True:
+            if scale_count <= len(all_pod_obj):
+                log.info(f"Scaled {scale_count} pvc and pods")
+                break
+            else:
+                log.info(f"Create {pvc_count_each_itr} pods & pvc")
+                pod_obj, pvc_obj = helpers.create_multi_pvc_pod(
+                    self.namespace_list[-1].name, rbd_sc_obj, cephfs_sc_obj, pvc_count_each_itr,
+                    size, fio_rate=fio_rate_param, start_io=start_io, fio_size=fio_size_param,
+                    fio_runtime=3600, node_selector=constants.SCALE_NODE_SELECTOR
```
RazTamir: isn't it too much to run 1 hour?

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
ramkiperiy: ack, will remove this.

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
ramkiperiy: ack.

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
ramkiperiy: For now the machineset function works only for IPI, and existing machineset.yaml will work for only AWS, will update this name more generic.

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
ramkiperiy: Expecting test case to run for more than 3hrs, since the config is going to create 1500/3000/4500 PVC+POD sequentially, but I should reduce the value for IO's at the end of test.

ramkiperiy:
Will be updating this PR to address this concern https://github.com/red-hat-storage/ocs-ci/issues/1639 
This will help to add more TC with existing Scale env of pod and pvcs.

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
```diff
@@ -0,0 +1,141 @@
...
+                )
+                all_pod_obj.extend(pod_obj)
+                try:
+                    # Check enough resources available in the dedicated app workers
+                    if helpers.add_worker_based_on_cpu_utilization(
+                        machineset_name=self.ms_name, node_count=1, expected_percent=75,
+                        role_type='app,worker'
+                    ):
```
Akarsha-rai, line 96: Observed even when node went to NotReady state and pods where not able to mount, this function didn't add the node.
I guess need to add some additional condition.

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
```diff
@@ -0,0 +1,141 @@
...
+        self.namespace_list.append(helpers.create_project())
+
+        # Continue to iterate till the scale pvc limit is reached
+        while True:
+            if scale_count <= len(all_pod_obj):
+                log.info(f"Scaled {scale_count} pvc and pods")
+                break
+            else:
+                log.info(f"Create {pvc_count_each_itr} pods & pvc")
+                pod_obj, pvc_obj = helpers.create_multi_pvc_pod(
```
Akarsha-rai, line 85: better to create dc pod than the nginx

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
```diff
@@ -0,0 +1,141 @@
...
+                    self.namespace_list[-1].name, rbd_sc_obj, cephfs_sc_obj, pvc_count_each_itr,
```
Akarsha-rai, line 86: replace self.namespace_list[-1].name with self.namespace_list[-1].namespace

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
```diff
@@ -0,0 +1,141 @@
...
+        ocs_nodes = node.get_typed_nodes(node_type='worker')
+        pvc_count = {3: 1500, 6: 3000, 9: 4500}
+        scale_count = pvc_count[len(ocs_nodes)]
+
+        # Other variables
+        rbd_sc_obj = helpers.default_storage_class(constants.CEPHBLOCKPOOL)
+        cephfs_sc_obj = helpers.default_storage_class(constants.CEPHFILESYSTEM)
+        pvc_count_each_itr = 5
+        size = f"{random.randrange(5, 105, 5)}Gi"
+        fio_size_param = '2G'
```
Akarsha-rai, line 63: it would be good to set fio_size and fio_rate according to the number of the osd's and it's size.

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
```diff
@@ -0,0 +1,141 @@
...
+                    fio_rate_param = helpers.suggest_fio_rate_based_on_cls_iops()
+
+                    # Check for pg_balancer
+                    if cluster.validate_pg_balancer():
+                        logging.info("OSD consumption and PG distribution is good to continue")
+                    else:
+                        raise UnexpectedBehaviour("Unequal PG distribution to OSDs")
+
+                    # Check for 200 pods per namespace
+                    pod_objs = pod.get_all_pods(namespace=self.namespace_list[-1].name)
```
Akarsha-rai, line 118: same here, replace self.namespace_list[-1].name with self.namespace_list[-1].namespace

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
```diff
@@ -0,0 +1,141 @@
...
+
+                    # Change fio_size and fio_rate param according to env
+                    fio_size_param = helpers.suggest_io_size_based_on_cls_usage()
+                    fio_rate_param = helpers.suggest_fio_rate_based_on_cls_iops()
+
+                    # Check for pg_balancer
+                    if cluster.validate_pg_balancer():
+                        logging.info("OSD consumption and PG distribution is good to continue")
+                    else:
+                        raise UnexpectedBehaviour("Unequal PG distribution to OSDs")
```
Akarsha-rai, line 115: it would be good to move this condition at the end. When new osd's are added backfilling starts and it takes some time and the test fails in between.

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
ramkiperiy: ack

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
ramkiperiy: ack, this is for initial testing but there is plan to have DC based pod.

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
ramkiperiy: Yes, ack will update it.

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
ramkiperiy: ok, let me check how to implement it.

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
ramkiperiy: ack

tests/e2e/scale/test_scale_pvc_pod_multi_project.py
ramkiperiy: ahh correct, need to skip this if there is rebalance going on in cluster.

ramkiperiy:
Closing this PR, Integrated scale helper functions in class format which is part of #1976
Comments in this PR will be addressed in latest 1976

PR #1601, opened 2020-03-05
Test to verify ceph block pool is not blocked by another invalid pool…

jilju:
… creation

Test to verify new ceph block pool can be created without deleting
ceph block pool having invalid parameters
Verifies bug https://bugzilla.redhat.com/show_bug.cgi?id=1711814


jilju:
Closing this because the reproducer steps of bug and this test case itself seems to be irrelevant in ocs 4.3.

Issue #1615, opened 2020-03-09
WA for default channel BZ 1811625


petr-balogh:
https://bugzilla.redhat.com/show_bug.cgi?id=1811625

Once the above is fixed we can remove the channel definitions from the config.

Once we will release 4.3 or for builds of 4.2 which are going to be released after 4.3 is out we will need to probably still have the default channel defined in 4.2 config cause the default channel will be 4.3.

petr-balogh:
Actually we do not want to have dynamic channel used for specific version cause this approach we used here can make sure we will not have such issue/bz in the future.

PR #1616, opened 2020-03-09
Propagate default channel for specific version

petr-balogh:
Fix for BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1811625

Workaround issue: #1615
After BZ is fixed we can remove channel from config files.

### DIFF
--- a/conf/ocs_version/ocs-4.2.yaml
+++ b/conf/ocs_version/ocs-4.2.yaml
@@ -2,6 +2,7 @@
 DEPLOYMENT:
   default_ocs_registry_image: "quay.io/rhceph-dev/ocs-olm-operator:4.2-rc"
   default_latest_tag: '4.2-rc'
+  ocs_csv_channel: 'stable-4.2'
 ENV_DATA:
   logging_version: '4.2'
   ocs_version: '4.2'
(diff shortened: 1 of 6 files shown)


vasukulkarni:
LGTM, but I would like to see internal wiki or mojo as well for the procedure, @andrewschoen @zmc FYI

petr-balogh:
> LGTM, but I would like to see internal wiki or mojo as well for the procedure, @andrewschoen @zmc FYI

@vasukulkarni what would you like to document exactly?

We have this issue opened:
https://github.com/red-hat-storage/ocs-ci/issues/1615

And this one this is fixed in the BZ we would like to just remove the default channel from the 4.3 and default config. For 4.2 config we will need to keep the default channel cause once 4.3 is out this will be the default in the product I think.  I think this can be updated in the product documentation for 4.2.

vasukulkarni:
> > LGTM, but I would like to see internal wiki or mojo as well for the procedure, @andrewschoen @zmc FYI
> 
> @vasukulkarni what would you like to document exactly?
> 
> We have this issue opened:
> #1615
> 
> And this one this is fixed in the BZ we would like to just remove the default channel from the 4.3 and default config. For 4.2 config we will need to keep the default channel cause once 4.3 is out this will be the default in the product I think. I think this can be updated in the product documentation for 4.2.

@petr-balogh the document should be from RCM, I want to see how the various versions are available from different channels from olm-operator and what settings should the user need for z/cve/point release.

PR #1624, opened 2020-03-10
test_az_failure: added skip marker

AviadP:


RazTamir:
@AviadP , this is considered as a workaround until the issue will be fixed.
Do we have such an issue reported?
Let's consider this with high priority as the tests will be skipped until got fixed

AviadP:
> 
> 
> @AviadP , this is considered as a workaround until the issue will be fixed.
> Do we have such an issue reported?
> Let's consider this with high priority as the tests will be skipped until got fixed

issue #1630

PR #1625, opened 2020-03-10
Delete folder from templates as post-check

vavuthu:

### DIFF
--- a/ocs_ci/deployment/vmware.py
+++ b/ocs_ci/deployment/vmware.py
@@ -283,6 +283,10 @@ class VSPHEREBASE(Deployment):
                 f"cluster {self.cluster}"
             )
 
+        # check whether folder exists in templates
+        if self.vsphere.is_folder_exists(pool, self.cluster, self.datacenter):
+            self.vsphere.destroy_folder(pool, self.cluster, self.datacenter)
+
 
 class VSPHEREUPI(VSPHEREBASE):
     """
--- a/ocs_ci/deployment/vmware.py
+++ b/ocs_ci/deployment/vmware.py
@@ -283,9 +283,8 @@ class VSPHEREBASE(Deployment):
                 f"cluster {self.cluster}"
             )
 
-        # check whether folder exists in templates
-        if self.vsphere.is_folder_exists(pool, self.cluster, self.datacenter):
-            self.vsphere.destroy_folder(pool, self.cluster, self.datacenter)
+        # destroy the folder in templates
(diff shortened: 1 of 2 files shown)


ocs_ci/utility/vsphere.py
```diff
@@ -543,3 +543,44 @@ def get_used_unit_number(self, vm):
             device.unitNumber for device in vm.config.hardware.device
             if hasattr(device.backing, 'fileName') and device.unitNumber != 0
         ]
+
+    def is_folder_exists(self, name, cluster, dc):
```
clacroix12: I would rename this to `folder_exists` or `check_folder_exists`

ocs_ci/deployment/vmware.py
```diff
@@ -283,6 +283,10 @@ def post_destroy_checks(self):
                 f"cluster {self.cluster}"
             )
 
+        # check whether folder exists in templates
+        if self.vsphere.is_folder_exists(pool, self.cluster, self.datacenter):
```
clacroix12: I'm not sure of any other usages of `destroy_folder` but would it make sense to perform this check within `destroy_folder` instead? This way we won't need to do this check before calling it if we have to call it in other places. This is of course assuming we wouldn't ever want to call `destroy_folder` on a folder that doesn't exist. 

I'd also like to see a log message saying that we couldn't find the folder if deletion is skipped due to that reason.

ocs_ci/utility/vsphere.py
vavuthu: Done

ocs_ci/deployment/vmware.py
vavuthu: Done

vavuthu:
recreated same scenario where folder exists in templates and called destroy job 
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-destroy-ocs-cluster/3891/console

``` 
11:08:30 - MainThread - ocs_ci.utility.utils - INFO - Executing command: terraform destroy '-var-file=/home/jenkins/current-cluster-dir/openshift-cluster-dir/terraform_data/terraform.tfvars' -auto-approve /home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/external/installer/upi/vsphere/
11:10:03 - MainThread - ocs_ci.deployment.vmware - INFO - Resource pool vavuthu-dc25 does not exist in cluster Cluster-1
11:10:26 - MainThread - ocs_ci.utility.vsphere - INFO - Destroying folder vavuthu-dc25 in templates
```

Issue #1644, opened 2020-03-12
pkg_resources.ContextualVersionConflict: kubernetes 11.0.0


clacroix12:
Looks like we're pulling `kubernetes` version `11.0.0` while `openshift` requires `~=10.0.1`

```
+ run-ci tests/ecosystem/ -m deployment --ocs-version 4.3 --ocsci-conf=**** '--ocsci-conf=~/current-cluster-dir/ocsci_conf.yaml' --ocsci-conf conf/ocsci/ocp-4.3-config.yaml --ocsci-conf=conf/deployment/vsphere/upi_1az_rhel_vmfs_3m_3w.yaml --ocsci-conf=conf/ocsci/production_device_size.yaml --cluster-name jnk-vu1rv33-t4a --cluster-path /home/jenkins/current-cluster-dir/openshift-cluster-dir --junit-xml /home/jenkins/current-cluster-dir/logs/destroy_results.xml -o junit_suite_name=jnk-vu1rv33-t4a --html=/home/jenkins/current-cluster-dir/logs/destroy_report.html --self-contained-html --teardown
Traceback (most recent call last):
  File "/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/.venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 584, in _build_master
    ws.require(__requires__)
  File "/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/.venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 901, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/.venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 792, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (kubernetes 11.0.0 (/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/.venv/lib/python3.6/site-packages), Requirement.parse('kubernetes~=10.0.1'), {'openshift'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/.venv/bin/run-ci", line 6, in <module>
    from pkg_resources import load_entry_point
[...]

PR #1645, opened 2020-03-12
Require kubernetes 10.0.1 to satisfy latest openshift package requirement

clacroix12:
Fixes: #1644

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
         'paramiko==2.4.2',
         'pyyaml>=4.2b1',
         'jinja2==2.10.1',
-        'kubernetes',
+        'kubernetes==10.0.1',  # latest openshift requires this version
         'openshift',
         'boto3',
         'munch',


clacroix12:
Current fix is to hard code the version that's required by `openshift` however I'm now wondering why we specify the `kubernetes` package in our requirements at all. If it's a dependency of `openshift`, why don't we omit it from our requirements and have whatever version is supported by `openshift` get installed automatically? @petr-balogh @RazTamir @ebenahar Is there a good reason why we have `kubernetes` specified in our requirements or can we simply remove it?

RazTamir:
Hi @Coady, if I remember correctly, we added this dependency in the early days of ocs where we explored the python-rest-client instead of OCP class.
I guess it is not needed anymore

clacroix12:
@RazTamir alright I figured it was leftover from the earlier stages of development. I think removing it is the better call, as we can ensure that we won't have a disconnect like this since we'll just pull whatever `openshift` needs. 

Unrelated, I noticed the unit tests are failing for some reason. It looks like previous PRs didn't hit this issue so I'm not really sure what the cause of it is yet but I'm in the process of investigating.

clacroix12:
Unit tests were broken because of a new pytest version. I have a workaround for it here: https://github.com/red-hat-storage/ocs-ci/pull/1647

Once the fix for the unit tests is merged I will update this PR and instead remove the `kubernetes` package from our requirements.

vasukulkarni:
pod exec still uses kubernetes so you will have to run those tests as well, have you looked into fixing unit test with latest pytest, the CI errors actually looks bit misleading to me.

clacroix12:
> pod exec still uses kubernetes so you will have to run those tests as well

we still get the `kubernetes` package if we remove it from our requirements because it's a dependency of `openshift`. However, if we delegate the version to whatever `openshift` supports we ensure that we don't hit this issue anymore.

> have you looked into fixing unit test with latest pytest, the CI errors actually looks bit misleading to me.

I wanted to get a workaround in first so things weren't entirely broken while we were looking into the fix.

clacroix12:
@vasukulkarni for now I'm going to leave it in but hardcode the version to get around the dependency error. If we want to remove it to delegate the version to whatever `openshift` needs we can do it in another PR.

vavuthu:
@clacroix12 this issue breaks all our jobs. have u run deployment checks local ?

clacroix12:
@vavuthu I haven't run a deployment but I have created a new venv and installed our requirements successfully which is what is failing in our jobs.

vavuthu:
I have kickstarted the destroy job with this job and it passed the stage where its failed before

Job link:  https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-destroy-ocs-cluster/3985/console

PR #1649, opened 2020-03-13
setup: be more generous in version requirement for kubernetes module

obnoxxx:
kubernetes 11.0.0 broke openshift. It requires `~=10.0.1`.
In the previous patch #1645 , the exact version 10.0.1 was required.
This change is relaxing the setup requirement to be `~=10.0.1`.
Not strictly required but may be more future-proof in case 10.0.2 is released and we need it or 10.0.1 would become unavailable.

See https://github.com/red-hat-storage/ocs-ci/issues/1644

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
         'paramiko==2.4.2',
         'pyyaml>=4.2b1',
         'jinja2==2.10.1',
-        'kubernetes==10.0.1',  # latest openshift requires this version
+        'kubernetes~=10.0.1',  # latest openshift requires this version
         'openshift',
         'boto3',
         'munch',


obnoxxx:
OK. The requirement of `kubernetes~=10.0.1` only matches versions of the form `10.0.x`. Not satisfied by `10.1.0`... But `10.0.1` does not seem to be available any more.

obnoxxx:
@petr-balogh - now this worked for the jenkins test. Not sure we would want to take it. Not needed, but might be more future-proof.

setup.py
```diff
@@ -22,7 +22,7 @@
         'paramiko==2.4.2',
         'pyyaml>=4.2b1',
         'jinja2==2.10.1',
-        'kubernetes==10.0.1',  # latest openshift requires this version
+        'kubernetes~=10.0.1',  # latest openshift requires this version
```
vasukulkarni: >= for better readability but one and the same.

clacroix12: >= would give us 11.x.x which was causing the original issue.

obnoxxx: No, that's not the same:
* `~=10.0.1` will match 10.0.1, 10.0.2 etc, but not 10.1.0 not 11.0.0, etc.
* `>=10.0.1` will match 10.0.1, 10.0.2, 10.1.0, 11.0.0.  ==> This fails here, because we get 11.0.0 in this case. And openshift requires `~= 10.0.1`.

obnoxxx: see #1644

vasukulkarni: I see, my initial thought was with major fixed >=10.0.* it would still stick with version, but i guess we need both '>=10.0.1,<=11'

vasukulkarni:
Only deployment has been run, please run few tier1. Thanks

obnoxxx:
@vasukulkarni  wrote:
> Only deployment has been run, please run few tier1. Thanks

how would I do this?

vasukulkarni:
> @vasukulkarni wrote:
> 
> > Only deployment has been run, please run few tier1. Thanks
> 
> how would I do this?

From ocs-jenkins we have the option to invoke the PR jobs, but if you have a setup locally you can run with your new venv and using '-m tier1' ?

obnoxxx:
> From ocs-jenkins we have the option to invoke the PR jobs, but if you have a setup locally you can run with your new venv and using '-m tier1' ?

I currently don't have a setup. I would like to get it and learn how to run locally, but can't do it right now.

vasukulkarni:
sorry I think this is now invalid due to recent PR that removed the kubernetes dependency, i think we can close this one.

PR #1651, opened 2020-03-13
Move smallfile workload tests into performance tier

keesturam:


vasukulkarni:
So the github merge is disabled and we dont have any jenkins/checks that is running that could have caused this, so needs to be checked further.

Issue #1652, opened 2020-03-13
We should parse the ocs version from ocs-registry-image parameter if passed


petr-balogh:
As was agreed on automation sync mtg this should fix the engineering job and not require to pass `--ocs-version` parameter if registry with specific image passed.

PR #1653, opened 2020-03-13
Take OCS version from registry image if passed

petr-balogh:
Fixes: #1652

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -50,9 +50,13 @@ to the pytest.
 ### Additional arguments:
 
 * `--cluster-name <name>` - name of cluster (always required for deployment) must be 5-17 characters long.
-* `--ocs-version` - version of OCS to be used (e.g. 4.2 or 4.3). If not specified, the default from ocs-ci will be used.
+* `--ocs-version` - version of OCS to be used (e.g. 4.2 or 4.3). If not
+    specified, the default from ocs-ci will be used. If --ocs-registry-image passed
+    than this parameter is ignored and version is parsed from registry image.
 * `--upgrade-ocs-version` - version of OCS to be used for upgrade (e.g. 4.2 or 4.3). If not specified, the default from ocs-ci will be used.
-* `--ocs-registry-image` - ocs registry image to be used for deployment (e.g quay.io/rhceph-dev/ocs-olm-operator:latest-4.2)
+* `--ocs-registry-image` - ocs registry image to be used for deployment
+    (e.g quay.io/rhceph-dev/ocs-olm-operator:latest-4.2). In case this parameter
+    is passed it overwrite the --ocs-version.
 * `--upgrade-ocs-registry-image` - ocs registry image to be used for upgrade (e.g quay.io/rhceph-dev/ocs-olm-operator:latest-4.3)
 * `--ocsci-conf` - with this configuration parameter you can overwrite the
     default OCS-CI config parameters defined in
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -56,8 +56,8 @@ to the pytest.
 * `--upgrade-ocs-version` - version of OCS to be used for upgrade (e.g. 4.2 or 4.3). If not specified, the default from ocs-ci will be used.
 * `--ocs-registry-image` - ocs registry image to be used for deployment
     (e.g quay.io/rhceph-dev/ocs-olm-operator:latest-4.2). In case this parameter
-    is passed, the registry image name and it overwrites any version passed in
-    via --ocs-version.
+    is passed, the version is parsed from the registry image name and it overwrites
+    any version passed in via --ocs-version.
 * `--upgrade-ocs-registry-image` - ocs registry image to be used for upgrade (e.g quay.io/rhceph-dev/ocs-olm-operator:latest-4.3)
 * `--ocsci-conf` - with this configuration parameter you can overwrite the
     default OCS-CI config parameters defined in
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -52,11 +52,12 @@ to the pytest.
 * `--cluster-name <name>` - name of cluster (always required for deployment) must be 5-17 characters long.
 * `--ocs-version` - version of OCS to be used (e.g. 4.2 or 4.3). If not
     specified, the default from ocs-ci will be used. If --ocs-registry-image passed
-    than this parameter is ignored and version is parsed from registry image.
+    then this parameter is ignored and the version is parsed from the registry image.
 * `--upgrade-ocs-version` - version of OCS to be used for upgrade (e.g. 4.2 or 4.3). If not specified, the default from ocs-ci will be used.
 * `--ocs-registry-image` - ocs registry image to be used for deployment
     (e.g quay.io/rhceph-dev/ocs-olm-operator:latest-4.2). In case this parameter
-    is passed it overwrite the --ocs-version.
+    is passed, the registry image name and it overwrites any version passed in
+    via --ocs-version.
 * `--upgrade-ocs-registry-image` - ocs registry image to be used for upgrade (e.g quay.io/rhceph-dev/ocs-olm-operator:latest-4.3)
 * `--ocsci-conf` - with this configuration parameter you can overwrite the
     default OCS-CI config parameters defined in
--- a/ocs_ci/framework/main.py
+++ b/ocs_ci/framework/main.py
@@ -52,19 +52,22 @@ def init_ocsci_conf(arguments=None):
     parser.add_argument('--ocs-registry-image')
     args, unknown = parser.parse_known_args(args=arguments)
     ocs_version = args.ocs_version
-    if args.ocs_registry_image:
-        ocs_version = utils.get_ocs_version_from_tag(args.ocs_registry_image)
-    if ocs_version:
-        version_config_file = os.path.join(
-            CONF_DIR, 'ocs_version', f'ocs-{ocs_version}.yaml'
-        )
-        args.ocsci_conf.insert(0, version_config_file)
     for config_file in args.ocsci_conf:
         with open(
             os.path.abspath(os.path.expanduser(config_file))
         ) as file_stream:
             custom_config_data = yaml.safe_load(file_stream)
             framework.config.update(custom_config_data)
+    ocs_registry_image = args.ocs_registry_image if (
+        args.ocs_registry_image
+    ) else framework.config.DEPLOYMENT.get('ocs_registry_image')
+    if ocs_registry_image:
+        ocs_version = utils.get_ocs_version_from_tag(args.ocs_registry_image)
+    if ocs_version:
+        version_config_file = os.path.join(
+            CONF_DIR, 'ocs_version', f'ocs-{ocs_version}.yaml'
+        )
+        args.ocsci_conf.insert(0, version_config_file)
     framework.config.RUN['run_id'] = int(time.time())
     bin_dir = framework.config.RUN.get('bin_dir')
     if bin_dir:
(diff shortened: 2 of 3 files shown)


docs/usage.md
```diff
@@ -50,9 +50,13 @@ to the pytest.
 ### Additional arguments:
 
 * `--cluster-name <name>` - name of cluster (always required for deployment) must be 5-17 characters long.
-* `--ocs-version` - version of OCS to be used (e.g. 4.2 or 4.3). If not specified, the default from ocs-ci will be used.
+* `--ocs-version` - version of OCS to be used (e.g. 4.2 or 4.3). If not
+    specified, the default from ocs-ci will be used. If --ocs-registry-image passed
+    than this parameter is ignored and version is parsed from registry image.
```
obnoxxx:
```suggestion
    then this parameter is ignored and the version is parsed from the registry image.
```

docs/usage.md
```diff
@@ -50,9 +50,13 @@ to the pytest.
...
 * `--upgrade-ocs-version` - version of OCS to be used for upgrade (e.g. 4.2 or 4.3). If not specified, the default from ocs-ci will be used.
-* `--ocs-registry-image` - ocs registry image to be used for deployment (e.g quay.io/rhceph-dev/ocs-olm-operator:latest-4.2)
+* `--ocs-registry-image` - ocs registry image to be used for deployment
+    (e.g quay.io/rhceph-dev/ocs-olm-operator:latest-4.2). In case this parameter
+    is passed it overwrite the --ocs-version.
```
obnoxxx:
```suggestion
    is passed, the version is parsed from the registry image name and it overwrites any version passed in via --ocs-version.
```

docs/usage.md
```diff
@@ -52,11 +52,12 @@ to the pytest.
 * `--cluster-name <name>` - name of cluster (always required for deployment) must be 5-17 characters long.
 * `--ocs-version` - version of OCS to be used (e.g. 4.2 or 4.3). If not
     specified, the default from ocs-ci will be used. If --ocs-registry-image passed
-    than this parameter is ignored and version is parsed from registry image.
+    then this parameter is ignored and the version is parsed from the registry image.
 * `--upgrade-ocs-version` - version of OCS to be used for upgrade (e.g. 4.2 or 4.3). If not specified, the default from ocs-ci will be used.
 * `--ocs-registry-image` - ocs registry image to be used for deployment
     (e.g quay.io/rhceph-dev/ocs-olm-operator:latest-4.2). In case this parameter
-    is passed it overwrite the --ocs-version.
+    is passed, the registry image name and it overwrites any version passed in
```
obnoxxx:
```suggestion
    is passed, the version is parsed from the registry image name and it overwrites any version passed in
```

This was missing a few words from the suggestion. Like this it is not a complete sentence. ;-)

ocs_ci/utility/utils.py
```diff
@@ -1796,3 +1797,31 @@ def skipif_ocs_version(expressions):
...
+
+    Returns
+        str: Version in x.y format
+
+    Raises:
+        ValueError: In case tag which we cannot parse passed.
+
+    """
+    try:
+        version = Version(tag.split(':')[1].lstrip("latest-"))
```
vavuthu: Above will give ValueError if we pass quay.io/rhceph-dev/ocs-olm-operator:latest-4.2

ocs_ci/framework/main.py
```diff
@@ -49,18 +49,25 @@ def init_ocsci_conf(arguments=None):
...
-        )
-        args.ocsci_conf.insert(0, version_config_file)
+    ocs_version = args.ocs_version
     for config_file in args.ocsci_conf:
         with open(
             os.path.abspath(os.path.expanduser(config_file))
         ) as file_stream:
             custom_config_data = yaml.safe_load(file_stream)
             framework.config.update(custom_config_data)
+    ocs_registry_image = args.ocs_registry_image if (
```
b-ranto: I might be nitpicking here but can we use standard multiline if here. Alternatively we could use `or` instead of the multiline if, i.e.
`ocs_registry_image = args.ocs_registry_image or framework.config.DEPLOYMENT.get('ocs_registry_image')`

ocs_ci/utility/utils.py
vavuthu:
``` >>> tag = "quay.io/rhceph-dev/ocs-olm-operator:latest-4.2"
>>> Version(tag.split(':')[1].lstrip("latest-"))
  File "<stdin>", line 1, in <module>
  File "/home/vavuthu/VJ/Gluster/automation/commits/MASTER/venv1/lib64/python3.7/site-packages/semantic_version/base.py", line 105, in __init__
    major, minor, patch, prerelease, build = self.parse(version_string, partial)
  File "/home/vavuthu/VJ/Gluster/automation/commits/MASTER/venv1/lib64/python3.7/site-packages/semantic_version/base.py", line 311, in parse
    raise ValueError('Invalid version string: %r' % version_string)
ValueError: Invalid version string: '4.2'
>>> 
>>> 
>>> Version.coerce(tag.split(':')[1].lstrip("latest-"))
Version('4.2.0')
>>> 
```

obnoxxx:
> This should work, but it creates code duplication. It would be better (if possible) to do the reconciliation of the ocs-version and ocs-registry-image variables at a later stage, after cmdline args and config file have both been parsed.

Forget that comment. After looking at the change with more context, I realized that it is already doing what I suggested. :-)

ocs_ci/framework/main.py
petr-balogh: done

obnoxxx: I agree, while this might be idiomatic python, I would usually prefer something like this because it's more explicit and less error prone:

```
ocs_registry_image = framework.config.DEPLOYMENT.get('ocs_registry_image')
if args.ocs_registry_image:
    ocs_registry_image = args.ocs_registry_image
...
```

Anyway, this is a style/taste discussion (which I enjoy) rather than about correctness or functionality.

ocs_ci/utility/utils.py
obnoxxx: If you treat `latest-` in the exception below, you can probably just skip lstrip here, right?
```suggestion
        version = Version(tag.split(':')[1])
```

ocs_ci/utility/utils.py
```diff
@@ -1796,3 +1797,34 @@ def skipif_ocs_version(expressions):
...
+            major=version.major, minor=version.minor
+        )
+    except ValueError as ex:
+        if version.endswith("-rc"):
+            # In case of the tag: 4.2-rc we are not able use semantic version.
+            log.info("X.Y-RC build detected")
+            return version.rstrip("-rc")
+        if version.startswith("latest-"):
+            log.info("latest-X.Y build detected")
+            return version.lstrip("latest-")
```
vavuthu: Above all can be handled by uisng Version.coerce

e.g:
```
>>> tag = "quay.io/rhceph-dev/ocs-olm-operator:latest-4.2-rc"
>>> Version.coerce(tag.split(':')[1].lstrip("latest-"))
Version('4.2.0-rc')
>>> "{major}.{minor}".format(major=version.major, minor=version.minor)
'4.2'
>>> 
```

ocs_ci/utility/utils.py
```diff
@@ -1796,3 +1797,28 @@ def skipif_ocs_version(expressions):
...
+
+    Returns
+        str: Version in x.y format
+
+    Raises:
+        ValueError: In case tag which we cannot parse passed.
+
+    """
+    try:
+        version = tag.split(':')[1].lstrip("latest-")
```
vavuthu: can we remove this

petr-balogh: this value is used in log

ocs_ci/framework/main.py
```diff
@@ -58,9 +58,9 @@ def init_ocsci_conf(arguments=None):
         ) as file_stream:
             custom_config_data = yaml.safe_load(file_stream)
             framework.config.update(custom_config_data)
-    ocs_registry_image = args.ocs_registry_image if (
-        args.ocs_registry_image
-    ) else framework.config.DEPLOYMENT.get('ocs_registry_image')
+    ocs_registry_image = args.ocs_registry_image or (
+        framework.config.DEPLOYMENT.get('ocs_registry_image')
+    )
```
obnoxxx: I personally prefer something less python-idiomatic but more explicit and possibly less error-prone like this:
```suggestion
    ocs_registry_image = framework.config.DEPLOYMENT.get('ocs_registry_image')
    if args.ocs_registry_image:
        ocs_registry_image = args.ocs_registry_image
```
But that's more a taste thing. ;-)

ocs_ci/utility/utils.py
obnoxxx: no. this *should* be removed. it is overwritten in the next line... ;-)

petr-balogh: Not sure how it got deleted but I am almost sure I had updated line bellow to what is coming in next commit

vavuthu: If we want log the error in str format in https://github.com/red-hat-storage/ocs-ci/pull/1653/commits/c8f7641eac7ba52d201835347bd44d6861e1e98d#diff-68bc5ed905b5a4cefbac287573f4f60fR1823 , then we need to take this into another variable.

petr-balogh: Why? if exception happen we just have the string from line 1817 not the version with Version.coerce

ocs_ci/utility/utils.py
```diff
@@ -1796,3 +1797,28 @@ def skipif_ocs_version(expressions):
...
+        version = Version.coerce(version)
```
vasukulkarni: why coerce? should work without it right?

ocs_ci/utility/utils.py
vavuthu: agreed

ocs_ci/utility/utils.py
b-ranto: See the previous discussions. this allows us to support `latest-4.Y` as well as `4.2.Z-<release>` tags.

vasukulkarni: But the 'latest-' is already stripped in line 1817 leaving with only supported version format?

vavuthu: @vasukulkarni '4.2' is invalid string in Version. To overcome that either we have to use coerce or pass "partial=True" as parameter to Version like 
``` Version('4.2', partial=True)```



https://pypi.org/project/semantic-version/

Comment (vasukulkarni) [reply ocs_ci/utility/utils.py]: I see, thanks!

File: ocs_ci/framework/main.py [hunk 10]
```diff
@@ -49,18 +65,20 @@ def init_ocsci_conf(arguments=None):
...
+            CONF_DIR, 'ocs_version', f'ocs-{ocs_version}.yaml'
         )
-        args.ocsci_conf.insert(0, version_config_file)
-    for config_file in args.ocsci_conf:
-        with open(
-            os.path.abspath(os.path.expanduser(config_file))
-        ) as file_stream:
-            custom_config_data = yaml.safe_load(file_stream)
-            framework.config.update(custom_config_data)
+        load_config([version_config_file])
```
obnoxxx: Would be possible in principle, that loading this config file overwrites the previously parsed `ocs_registry_image` or calculated/parsed `ocs_version` value? 

In this case, we should reconcile the version again..

obnoxxx: (or prevent this in order to be looping...)

petr-balogh: This will not be overwritten as we do not have `ocs_registry_image` in any of the config file.  Only if user will pass this one this has precedence over default value which is defined in the config files like  `default_ocs_registry_image`

obnoxxx: Not having it in any example file is one thing. But we *could* have it in a config file, right? So in theory it could have an effect...

Issue #1667, opened 2020-03-17
Deployment tests should check whether the right version of OCS is installed or not 


hnallurv:
Creating this issue as per the discussion in last QE staff meeting. It's very important that the Deployment tests check whether the right version of OCS is installed or not. If wrong version is installed and no checking done then subsequent tests will be run on wrong version.

PR #1668, opened 2020-03-17
Add check of OCS version is correct in the CSV

petr-balogh:
Fixes: #1667

### DIFF
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -35,7 +35,9 @@ class StorageCluster(OCP):
         )
 
 
-def ocs_install_verification(timeout=600, skip_osd_distribution_check=False):
+def ocs_install_verification(
+    timeout=600, skip_osd_distribution_check=False, ocs_registry_image=None
+):
     """
     Perform steps necessary to verify a successful OCS installation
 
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -66,6 +70,29 @@ def ocs_install_verification(timeout=600, skip_osd_distribution_check=False):
...
+        f"Check if OCS version: {ocs_version} matches with CSV: {csv_version}"
+    )
+    assert ocs_version in csv_version, (
+        f"OCS version: {ocs_version} mismatch with CSV version {csv_version}"
+    )
+    # Verify if OCS CSV has the same version in provided CI build.
+    ocs_registry_image = ocs_registry_image or config.DEPLOYMENT.get(
+        'ocs_registry_image'
+    )
+    if ocs_registry_image and ocs_registry_image.endswith(".ci"):
```
vasukulkarni: I think we can get rid of this if statement so that assert is always checked, if for some reason it doesn't end with .ci, corrections can be made to catch up.

petr-balogh: Then you wont be able to deploy like tag like 4.2.3-rc for example and other tags. Do we want to strick users that much?

vasukulkarni: maybe we can get rid of .ci or .rc before any check happens?

petr-balogh: But the version of image tag and version of CSV matches only for this specific CI tag ends wit .ci.

PR #1675, opened 2020-03-19
Fix quay-tag-image usage

b-ranto:
The quay-tag-image Jenkins job now requires full destionation url
instead of image name and tag string.


b-ranto:
@zmc Can you link the other PR? I guess we could drop the `DESTINATION_URL` patch altogether now when we don't actually need it.

Issue #1696, opened 2020-03-23
Get Storage capacity


Avilir:
This issue is to be able running performance test on different platforms and with different storage capacity automaticly.
the test need to find the storage capacity and modify its parameters accordingly.

mbukatov:
I have a fio job for monitoring testing which can target percentage of user accesible utilization based on MAX AVAIL data. Right now, I'm polishing the code so that this could be used for data verification after a cluster disruptions like shutdown or upgrade.

Avilir:
closed by PR #1968

PR #1698, opened 2020-03-23
Adding get_ceph_capacity function

Avilir:
This PR is to add function that retrieve the Storage capacity in GB to support Issue #1696

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -485,23 +485,23 @@ class CephCluster(object):
 
     def get_ceph_capacity(self):
         """
-        The function gets the total storage capacity of the ocs cluster
+        The function gets the total usable storage capacity of the ocs cluster.
 
         Returns:
             int : Total storage capacity in GiB (GiB is for development environment)
 
         """
+        storage_cluster_obj = storage_cluster.StorageCluster(
+            resource_name=config.ENV_DATA['storage_cluster_name'],
+            namespace=config.ENV_DATA['cluster_namespace'],
+        )
+        replica = int(storage_cluster_obj.data['spec']['storageDeviceSets'][0]['replica'])
 
-        ceph_status = self.get_ceph_status()
-        for item in ceph_status.split("\n"):
-            if 'usage:' in item:
-                total_storage = re.findall(r'\d+\.+\d+|\d\d*', item.strip())
-                total_storage = float(total_storage[2])
-                unit = item.split(' ')[-2]
-                conversion = {'GiB': 1, 'TiB': 1024}
-                total_storage = total_storage * conversion[unit]
-                logging.info(f"Total Storage capacity of the cluster is {total_storage}")
-                return int(total_storage)
+        ceph_pod = pod.get_ceph_tools_pod()
+        ceph_status = ceph_pod.exec_ceph_cmd(ceph_cmd="ceph df")
+        usable_capacity = int(ceph_status['stats']['total_bytes']) / replica / constant.GB
+
+        return usable_capacity
 
     def get_ceph_cluster_iops(self):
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/cluster.py
```diff
@@ -483,6 +483,26 @@ def get_ceph_status(self):
             "ceph status", out_yaml_format=False,
         )
 
+    def get_ceph_capacity(self):
+        """
+        The function gets the total storage capacity of the ocs cluster
+
+        Returns:
+            Total storage capacity in GiB (GiB is for development environment)
```
RazTamir: str? int?

mbukatov:
Also note that this capacity has nothing to do with space one can actually utilize.

ocs_ci/ocs/cluster.py
Avilir: fixed.

Avilir:
> I see multiple problems with such functionality:
> 
>     * parsing of human readable plain text output
> 
>     * ceph storage utilization is a tricky concept, but the functions doesn't seem to indicate what it is reporting neither via name nor docstring
> 
> 
> I believe that for getting information about ceph storage utilisation, one should use `ct_pod.exec_ceph_cmd(ceph_cmd="ceph df")` which returns dictionary created from a yaml, where you can find:
> 
>     * `['stats']['total_bytes']`
> 
>     * `['stats']['total_used_bytes']`
> 
>     * `['stats']['total_used_raw_bytes']`
> 
> 
> and others.

ctualy, this output line is the summery of "ceph df", and it is more simple to get.

Avilir:
> Also note that this capacity is not the space one can actually utilize, which depends on a ceph pool configuration, and also a little bit on a volume type.

I know that this is the total RAW capacity and not the usable capacity (which depend on the replica number).

mbukatov:
> ctualy, this output line is the summery of "ceph df", and it is more simple to get.

I don't see how parsing human readable output via regexps is easier than asking for particular value from a python dictionary structure.

Also, Is the value you are returning `total_used_bytes` or `total_used_raw_bytes`?

Avilir:
> > ctualy, this output line is the summery of "ceph df", and it is more simple to get.
> 
> I don't see how parsing human readable output via regexps is easier than asking for particular value from a python dictionary structure.
> 
> Also, Is the value you are returning `total_used_bytes` or `total_used_raw_bytes`?

Changed the return value to the : Usable Capacity (consider the replication).

Avilir:
> I consider this way of getting ceph capacity as needlessly complex and prone to error.
> 
> That said, since the rest of the code in the same module is implemented this way, feel free to ignore me.

Changed to pull the information from the JSON output of 'ceph df'

ocs_ci/ocs/cluster.py
```diff
@@ -483,6 +483,26 @@ def get_ceph_status(self):
             "ceph status", out_yaml_format=False,
         )
 
+    def get_ceph_capacity(self):
+        """
+        The function gets the total usable storage capacity of the ocs cluster.
```
annagitel: did you mean returns instead of gets?

shreekarSS:
For cluster expansion, we have this function 

def get_percent_used_capacity():
    ct_pod = pod_helpers.get_ceph_tools_pod()
    output = ct_pod.exec_ceph_cmd(ceph_cmd='ceph df')
    total_used = (output.get('stats').get('total_used_raw_bytes'))
    total_avail = (output.get('stats').get('total_bytes'))
    return 100.0 * total_used / total_avail

mbukatov:
Note that as stated in the docstring, this is an approximation, which may not work in all cluster configurations, use cases or in the future, but is probably fine in most cases.

PR #1704, opened 2020-03-25
Support deployments using LSO for AWS IPI on i3 instance types

clacroix12:
Jira: https://projects.engineering.redhat.com/browse/OCSQE-190

### DIFF
--- a/conf/ocsci/aws_i3_local_storage.yaml
+++ b/conf/ocsci/aws_i3_local_storage.yaml
@@ -5,4 +5,4 @@ DEPLOYMENT:
 ENV_DATA:
   platform: 'aws'
   worker_instance_type: 'i3en.2xlarge'
-  device_size: 2300
+  device_size: 2300  # Value pulled from here: https://bugzilla.redhat.com/show_bug.cgi?id=1812557#c27
--- a/conf/ocsci/aws_i3_local_storage.yaml
+++ b/conf/ocsci/aws_i3_local_storage.yaml
@@ -1,0 +1,8 @@
+---
+DEPLOYMENT:
+  local_storage: true
+
+ENV_DATA:
+  platform: 'aws'
+  worker_instance_type: 'i3en.2xlarge'
+  device_size: 2300
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -43,6 +43,7 @@ from ocs_ci.utility.retry import retry
 from ocs_ci.utility.utils import (
     ceph_health_check,
     get_latest_ds_olm_tag,
+    get_ocp_version,
     is_cluster_running,
     run_cmd,
 )
@@ -722,7 +723,7 @@ def setup_local_storage():
     # Update local-storage-operator subscription data with channel
     for data in lso_data:
         if data['kind'] == 'Subscription':
-            data['spec']['channel'] = config.ENV_DATA.get('ocs_version')
+            data['spec']['channel'] = get_ocp_version()
 
     # Create temp yaml file and create local storage operator
     logger.info(
@@ -773,8 +774,6 @@ def setup_local_storage():
         lv_data, lv_data_yaml.name
     )
     logger.info("Creating LocalVolume CR")
-    with open(lv_data_yaml.name) as f:
-        logger.info(f.read())
     run_cmd(f"oc create -f {lv_data_yaml.name}")
     logger.info("Waiting 30 seconds for PVs to create")
     verify_pvs_created(len(worker_names))
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -14,7 +14,9 @@ from ocs_ci.deployment.ocp import OCPDeployment as BaseOCPDeployment
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants, ocp, defaults, registry
 from ocs_ci.ocs.cluster import validate_cluster_on_pvc, validate_pdb_creation
-from ocs_ci.ocs.exceptions import CommandFailed, UnavailableResourceException
+from ocs_ci.ocs.exceptions import (
+    CommandFailed, UnavailableResourceException, UnsupportedPlatformError
+)
 from ocs_ci.ocs.monitoring import (
     create_configmap_cluster_monitoring_pod,
     validate_pvc_created_and_bound_on_monitoring_pods,
@@ -746,38 +748,22 @@ def setup_local_storage():
     ), "Local storage operator did not reach running phase"
 
     # Retrieve NVME device path ID for each worker node
-    dev_paths = []
-    for worker in worker_names:
-        logger.info("Retrieving device path for node: %s", worker)
-        cmd = (
-            f"oc debug nodes/{worker} "
-            f"-- chroot /host ls -la /dev/disk/by-id/"
-        )
-        out = run_cmd(cmd)
-        out_lines = out.split('\n')
-        pattern = 'nvme-nvme.1d0f-'
-        nvme_line = [line for line in out_lines if pattern in line][0]
-        device_path = [
-            part for part in nvme_line.split(' ') if pattern in part
-        ][0]
-        logger.info("Adding %s to device paths", device_path)
-        dev_paths.append(f'/dev/disk/by-id/{device_path}')
+    device_paths = get_device_paths(worker_names)
 
     # Pull local volume yaml data
-    logger.info("Pulling local volume data from yaml")
+    logger.info("Pulling LocalVolume CR data from yaml")
     lv_data = templating.load_yaml(
         constants.LOCAL_VOLUME_YAML
     )
 
     # Update local volume data with NVME IDs
     logger.info(
-        "Updating local volume data with device paths: %s",
-        dev_paths
+        "Updating LocalVolume CR data with device paths: %s",
+        device_paths
     )
     lv_data['spec']['storageClassDevices'][0][
         'devicePaths'
-    ] = dev_paths
-    logger.debug(lv_data)
+    ] = device_paths
 
     # Create temp yaml file and create local volume
     lv_data_yaml = tempfile.NamedTemporaryFile(
@@ -786,7 +772,9 @@ def setup_local_storage():
     templating.dump_data_to_temp_yaml(
         lv_data, lv_data_yaml.name
     )
-    logger.info("Creating local volume")
+    logger.info("Creating LocalVolume CR")
+    with open(lv_data_yaml.name) as f:
+        logger.info(f.read())
     run_cmd(f"oc create -f {lv_data_yaml.name}")
     logger.info("Waiting 30 seconds for PVs to create")
     verify_pvs_created(len(worker_names))
@@ -817,7 +805,41 @@ def verify_pvs_created(num_workers):
         )
     num_pvs = len(pv_json['items'])
     logger.debug("PVs, Workers: %s, %s", num_pvs, num_workers)
-    assert (num_pvs == num_workers), (
-        f'Number of PVs ({num_pvs}) does not match number of workers '
-        f'({num_workers})'
-    )
+
+
+def get_device_paths(worker_names):
+    """
+    Retrieve a list of the device paths for each worker node
+
+    Args:
+        worker_names (list): worker node names
+
+    Returns:
+        list: device path ids
+    """
+    device_paths = []
+    platform = config.ENV_DATA.get('platform').lower()
+    if platform == 'aws':
+        pattern = 'nvme-Amazon_EC2_NVMe_Instance_Storage'
+    # TODO: add patterns for vsphere and bare metal
+    else:
+        raise UnsupportedPlatformError(
+            'LSO deployment is not supported for platform: %s', platform
+        )
+    for worker in worker_names:
+        logger.info("Retrieving device path for node: %s", worker)
+        cmd = (
+            f"oc debug nodes/{worker} "
+            f"-- chroot /host ls -la /dev/disk/by-id/"
+        )
+        out = run_cmd(cmd)
+        out_lines = out.split('\n')
+        nvme_lines = [line for line in out_lines if pattern in line]
+        for nvme_line in nvme_lines:
+            device_path = [
+                part for part in nvme_line.split(' ') if pattern in part
+            ][0]
+            logger.info("Adding %s to device paths", device_path)
+            device_paths.append(f'/dev/disk/by-id/{device_path}')
+
+    return device_paths
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -20,7 +20,7 @@ from ocs_ci.ocs.monitoring import (
     validate_pvc_created_and_bound_on_monitoring_pods,
     validate_pvc_are_mounted_on_monitoring_pods
 )
-from ocs_ci.ocs.node import get_typed_worker_nodes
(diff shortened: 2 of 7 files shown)


clacroix12:
This has been locally tested and verified with IPI deployments. I have a [change to the openshift-misc repo](https://code.engineering.redhat.com/gerrit/#/c/195496/) that should allow testing for UPI but that has not been merged yet.

ocs_ci/deployment/deployment.py
```diff
@@ -652,3 +667,131 @@ def create_catalog_source(image=None, ignore_upgrade=False):
...
+    for worker in worker_names:
+        logger.info("Retrieving device path for node: %s", worker)
+        cmd = (
+            f"oc debug nodes/{worker} "
+            f"-- chroot /host ls -la /dev/disk/by-id/"
+        )
+        out = run_cmd(cmd)
+        out_lines = out.split('\n')
+        pattern = 'nvme-nvme.1d0f-'
+        nvme_line = [line for line in out_lines if pattern in line][0]
```
Avilir: This will take the OS device too. i think you can look in the pattern for EC2, this will give you the Free devices only.

ocs_ci/deployment/deployment.py
```diff
@@ -652,3 +667,131 @@ def create_catalog_source(image=None, ignore_upgrade=False):
...
+    local_storage_operator = ocp.OCP(
+        kind=constants.POD, namespace='local-storage'
+    )
+    assert local_storage_operator.wait_for_resource(
+        condition=constants.STATUS_RUNNING,
+        selector=constants.LOCAL_STORAGE_OPERATOR_LABEL,
+        timeout=600
+    ), "Local storage operator did not reach running phase"
+
+    # Retrieve NVME device path ID for each worker node
```
RazTamir: I wonder if from this point on is not too specific for i3?
If we can make this method more generic for ALL LSO deployments, it will be great

ocs_ci/deployment/deployment.py
```diff
@@ -652,3 +667,131 @@ def create_catalog_source(image=None, ignore_upgrade=False):
...
+
+    # Pull local volume yaml data
+    logger.info("Pulling local volume data from yaml")
+    lv_data = templating.load_yaml(
+        constants.LOCAL_VOLUME_YAML
+    )
+
+    # Update local volume data with NVME IDs
+    logger.info(
+        "Updating local volume data with device paths: %s",
```
RazTamir: local volume --> localVolume CR

conf/ocsci/aws_i3_local_storage.yaml
```diff
@@ -0,0 +1,8 @@
+---
+DEPLOYMENT:
+  local_storage: true
+
+ENV_DATA:
+  platform: 'aws'
+  worker_instance_type: 'i3en.2xlarge'
+  device_size: 2300
```
petr-balogh: Can you please comment from which place this value came?

ocs_ci/deployment/deployment.py
```diff
@@ -652,3 +667,131 @@ def create_catalog_source(image=None, ignore_upgrade=False):
...
+    )
+
+    logger.info("Retrieving local-storage-operator data from yaml")
+    lso_data = list(templating.load_yaml(
+        constants.LOCAL_STORAGE_OPERATOR, multi_document=True
+    ))
+    # Update local-storage-operator subscription data with channel
+    for data in lso_data:
+        if data['kind'] == 'Subscription':
+            data['spec']['channel'] = config.ENV_DATA.get('ocs_version')
```
petr-balogh: This is probably related to version of OCP and not OCS. Please ask in the KNIP and for example for un GAed OCP version like 4.4.

ocs_ci/deployment/deployment.py
clacroix12: Here's the full output I get from this command:

```
drwxr-xr-x. 2 root root 340 Mar 25 15:27 .
drwxr-xr-x. 8 root root 160 Mar 25 15:27 ..
lrwxrwxrwx. 1 root root  10 Mar 25 15:27 dm-name-coreos-luks-root-nocrypt -> ../../dm-0
lrwxrwxrwx. 1 root root  13 Mar 25 15:27 nvme-Amazon_EC2_NVMe_Instance_Storage_AWS6428258570A1B4DC5 -> ../../nvme1n1
lrwxrwxrwx. 1 root root  13 Mar 25 15:27 nvme-Amazon_EC2_NVMe_Instance_Storage_AWSB9CF180FB5F23849F -> ../../nvme2n1
lrwxrwxrwx. 1 root root  13 Mar 25 15:27 nvme-Amazon_Elastic_Block_Store_vol088bf4676b265b3ba -> ../../nvme0n1
lrwxrwxrwx. 1 root root  15 Mar 25 15:27 nvme-Amazon_Elastic_Block_Store_vol088bf4676b265b3ba-part1 -> ../../nvme0n1p1
lrwxrwxrwx. 1 root root  15 Mar 25 15:27 nvme-Amazon_Elastic_Block_Store_vol088bf4676b265b3ba-part2 -> ../../nvme0n1p2
lrwxrwxrwx. 1 root root  15 Mar 25 15:27 nvme-Amazon_Elastic_Block_Store_vol088bf4676b265b3ba-part3 -> ../../nvme0n1p3
lrwxrwxrwx. 1 root root  15 Mar 25 15:27 nvme-Amazon_Elastic_Block_Store_vol088bf4676b265b3ba-part4 -> ../../nvme0n1p4
lrwxrwxrwx. 1 root root  13 Mar 25 15:27 nvme-nvme.1d0f-4157533634323832353835373041314234444335-416d617a6f6e20454332204e564d6520496e7374616e63652053746f72616765-00000001 -> ../../nvme1n1
lrwxrwxrwx. 1 root root  13 Mar 25 15:27 nvme-nvme.1d0f-4157534239434631383046423546323338343946-416d617a6f6e20454332204e564d6520496e7374616e63652053746f72616765-00000001 -> ../../nvme2n1
lrwxrwxrwx. 1 root root  13 Mar 25 15:27 nvme-nvme.1d0f-766f6c3038386266343637366232363562336261-416d617a6f6e20456c617374696320426c6f636b2053746f7265-00000001 -> ../../nvme0n1
lrwxrwxrwx. 1 root root  15 Mar 25 15:27 nvme-nvme.1d0f-766f6c3038386266343637366232363562336261-416d617a6f6e20456c617374696320426c6f636b2053746f7265-00000001-part1 -> ../../nvme0n1p1
lrwxrwxrwx. 1 root root  15 Mar 25 15:27 nvme-nvme.1d0f-766f6c3038386266343637366232363562336261-416d617a6f6e20456c617374696320426c6f636b2053746f7265-00000001-part2 -> ../../nvme0n1p2
lrwxrwxrwx. 1 root root  15 Mar 25 15:27 nvme-nvme.1d0f-766f6c3038386266343637366232363562336261-416d617a6f6e20456c617374696320426c6f636b2053746f7265-00000001-part3 -> ../../nvme0n1p3
lrwxrwxrwx. 1 root root  15 Mar 25 15:27 nvme-nvme.1d0f-766f6c3038386266343637366232363562336261-416d617a6f6e20456c617374696320426c6f636b2053746f7265-00000001-part4 -> ../../nvme0n1p4
```

The devices with EC2 in them:
```
lrwxrwxrwx. 1 root root  13 Mar 25 15:27 nvme-Amazon_EC2_NVMe_Instance_Storage_AWS6428258570A1B4DC5 -> ../../nvme1n1
lrwxrwxrwx. 1 root root  13 Mar 25 15:27 nvme-Amazon_EC2_NVMe_Instance_Storage_AWSB9CF180FB5F23849F -> ../../nvme2n1
```

The ID my code is grabbing:
```
nvme-nvme.1d0f-4157533634323832353835373041314234444335-416d617a6f6e20454332204e564d6520496e7374616e63652053746f72616765-00000001
```

The device-id I'm grabbing (and the one similar to it that links to `nvme2n1`) seem to the same as the two that have EC2 in them. How can I differentiate between the OS device and the other from this output?

vasukulkarni: I think can use `blkid` to figure out fs on partitions on skip them, ceph-disk has used variety of things before, it is quite complex for all scenarios but for us its a good start.

clacroix12: I see what @AviadP means now. If we use a pattern like `nvme-Amazon_EC2_NVMe_Instance_Storage` then we will only get the two free NVME devices and not the OS device in addition to the free ones. What I have works because the ones we want are first in the list but if the list only contains the free devices that is better.

ocs_ci/deployment/deployment.py
```diff
@@ -652,3 +667,131 @@ def create_catalog_source(image=None, ignore_upgrade=False):
...
+    dev_paths = []
+    for worker in worker_names:
+        logger.info("Retrieving device path for node: %s", worker)
+        cmd = (
+            f"oc debug nodes/{worker} "
+            f"-- chroot /host ls -la /dev/disk/by-id/"
+        )
+        out = run_cmd(cmd)
+        out_lines = out.split('\n')
+        pattern = 'nvme-nvme.1d0f-'
```
prsurve: The pattern varies in baremetal some case it will be sda or any other disk

ocs_ci/ocs/constants.py
```diff
@@ -644,3 +644,14 @@
...
+
+# local storage
+LOCAL_STORAGE_OPERATOR = os.path.join(
+    TEMPLATE_DEPLOYMENT_DIR, "local-storage-operator.yaml"
+)
+LOCAL_VOLUME_YAML = os.path.join(
+    TEMPLATE_DEPLOYMENT_DIR, "local-volume.yaml"
+)
+LOCAL_STORAGE_CLUSTER_YAML = os.path.join(
+    TEMPLATE_DEPLOYMENT_DIR, "local-storage-cluster.yaml"
```
vavuthu: local-storage-cluster.yaml is missing in PR

clacroix12: Good catch, I was originally using this file but removed it in favor of updating the data we normally pull for storage cluster. I removed this constant in the latest update.

conf/ocsci/aws_i3_local_storage.yaml
clacroix12: Done

ocs_ci/deployment/deployment.py
clacroix12: Done

ocs_ci/deployment/deployment.py
clacroix12: I've moved the device path code to its own function and added a TODO for determining patterns for vmware and bare metal. Once we know how to find the device paths for those platforms I can add that logic in a follow up PR.

ocs_ci/deployment/deployment.py
clacroix12: I'm now using `Amazon_EC2_NVMe_Instance_Storage` as the pattern we use to find the free NVME device paths. I also changed the logic to add both drives this was brought up as something we should be doing in our platform testing syncup.

ocs_ci/deployment/deployment.py
clacroix12: I haven't received confirmation of this yet but I swapped to using the ocp version instead.

conf/ocsci/aws_i3_local_storage.yaml
```diff
@@ -0,0 +1,8 @@
+---
+DEPLOYMENT:
+  local_storage: true
+
+ENV_DATA:
+  platform: 'aws'
+  worker_instance_type: 'i3en.2xlarge'
+  device_size: 2300  # Value pulled from here: https://bugzilla.redhat.com/show_bug.cgi?id=1812557#c27
```
ebenahar: according to https://aws.amazon.com/ec2/instance-types/, i3en.2xlarge instances has 2 x 2,500 NVMe SSD

ocs_ci/deployment/deployment.py
```diff
@@ -678,3 +696,149 @@ def create_catalog_source(image=None, ignore_upgrade=False):
...
+        pv_name = pv['metadata']['name']
+        logger.info("%s is %s", pv_name, pv_state)
+        assert pv_state == 'Available', (
+            f"{pv_name} not in 'Available' state. Current state is {pv_state}"
+        )
+    num_pvs = len(pv_json['items'])
+    logger.debug("PVs, Workers: %s, %s", num_pvs, num_workers)
+
+
+def get_device_paths(worker_names):
```
ebenahar: @clacroix12 are you planning on creating classes for the different LSO types, like we have for AWSIPI, VSPHEREUPI, etc. (maybe AWSLSO, VSPHERELSO, BMLSO?). 
And, as raised by Raz in https://github.com/red-hat-storage/ocs-ci/pull/1704/files#r397939609, having this function being implemented in the corresponding deployment class

clacroix12: I think it will be overkill. If the only thing that changes is how we retrieve the device path we don't really need separate classes to handle it. This function can simply have conditionals. I think it depends on how much changes between the platform types.

If the logic changes enough to warrant breaking it out into separate class methods then that can absolutely be done but right now I don't have enough to go off of to make that determination.

conf/ocsci/aws_i3_local_storage.yaml
clacroix12: Yes, they do. I wasn't sure if we weren't using the entire device or some reason so I went with what was suggested. We can change this to 2500 if there isn't a good reason to have it at 2300.

ocs_ci/deployment/deployment.py
ebenahar: Yes, sounds good. For now, conditions in this function will do the job

sidhant-agrawal: @clacroix12 We will need one more change for VMware.
For Mon either we have to create and attach a 10GB VMDK and create the Filesystem LocalVolume CR 
Or use default thin storageclass for automatic creation and attaching VMDKs

ocs_ci/deployment/deployment.py
```diff
@@ -678,3 +696,149 @@ def create_catalog_source(image=None, ignore_upgrade=False):
...
+        worker_names (list): worker node names
+
+    Returns:
+        list: device path ids
+    """
+    device_paths = []
+    platform = config.ENV_DATA.get('platform').lower()
+    if platform == 'aws':
+        pattern = 'nvme-Amazon_EC2_NVMe_Instance_Storage'
+    # TODO: add patterns for vsphere and bare metal
```
sidhant-agrawal: One suggestion which can be taken up in follow up PR:
we can use one or all of the free device available and not just disk matching the pattern having `nvme-..`
In this way we will be able to do LSO based deployment even if we don't have nvme disks in that setup.

clacroix12: Do you have a reliable (and automatable) way to determine which disks are free and which are not?

ocs_ci/deployment/deployment.py
```diff
@@ -678,3 +696,149 @@ def create_catalog_source(image=None, ignore_upgrade=False):
...
+    logger.info("Creating LocalVolume CR")
+    run_cmd(f"oc create -f {lv_data_yaml.name}")
+    logger.info("Waiting 30 seconds for PVs to create")
+    verify_pvs_created(len(worker_names))
+
+
+@retry(AssertionError, 12, 10, 1)
+def verify_pvs_created(num_workers):
+    """
+    Verify that PVs were created and are in the Available state
```
vasukulkarni: we could add the retry info here as well - 12 retries with wait times before it gives up.

PR #1705, opened 2020-03-25
Add OCP ugprade related markers

petr-balogh:
Fixes: #1700

### DIFF
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -501,8 +501,13 @@ PEM_PATH = "/etc/pki/ca-trust/source/anchors/"
 # Upgrade related constants, keeping some space between, so we can add
 # additional order.
 ORDER_BEFORE_UPGRADE = 10
-ORDER_UPGRADE = 20
-ORDER_AFTER_UPGRADE = 30
+ORDER_BEFORE_OCP_UPGRADE = 20
+ORDER_OCP_UPGRADE = 30
+ORDER_AFTER_OCP_UPGRADE = 40
+ORDER_BEFORE_OCS_UPGRADE = 50
+ORDER_OCS_UPGRADE = 60
+ORDER_AFTER_OCS_UPGRADE = 70
+ORDER_AFTER_UPGRADE = 80
 
 # Deployment constants
 OCS_CSV_PREFIX = 'ocs-operator'
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -7,9 +7,14 @@ import pytest
 from funcy import compose
 
 from ocs_ci.ocs.constants import (
-    ORDER_AFTER_UPGRADE,
+    ORDER_BEFORE_OCS_UPGRADE,
+    ORDER_BEFORE_OCP_UPGRADE,
     ORDER_BEFORE_UPGRADE,
-    ORDER_UPGRADE,
+    ORDER_OCP_UPGRADE,
+    ORDER_OCS_UPGRADE,
+    ORDER_AFTER_OCP_UPGRADE,
+    ORDER_AFTER_OCS_UPGRADE,
+    ORDER_AFTER_UPGRADE,
 )
 from ocs_ci.framework import config
 from ocs_ci.utility.utils import check_if_executable_in_path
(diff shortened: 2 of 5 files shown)


petr-balogh:
Verification for OCS upgrade only here:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/5919/console

Once this will finish I will try to run OCP upgrade as well on same env.

ocs_ci/ocs/constants.py
```diff
@@ -501,8 +501,13 @@
...
 ORDER_BEFORE_UPGRADE = 10
-ORDER_UPGRADE = 20
-ORDER_AFTER_UPGRADE = 30
+ORDER_BEFORE_OCS_UPGRADE = 20
+ORDER_BEFORE_OCP_UPGRADE = 30
+ORDER_OCP_UPGRADE = 40
+ORDER_OCS_UPGRADE = 50
+ORDER_AFTER_OCP_UPGRADE = 60
+ORDER_AFTER_OCS_UPGRADE = 70
+ORDER_AFTER_UPGRADE = 70
```
petr-balogh: When I am looking at this I think that more better ordering will be 
```python
ORDER_BEFORE_UPGRADE = 10
ORDER_BEFORE_OCP_UPGRADE = 20
ORDER_OCP_UPGRADE = 30
ORDER_AFTER_OCP_UPGRADE = 40
ORDER_BEFORE_OCS_UPGRADE = 50
ORDER_OCS_UPGRADE = 60
ORDER_AFTER_OCS_UPGRADE = 70
ORDER_AFTER_UPGRADE = 8
```

WDYT @ebenahar @fbalak @AviadP

fbalak: How about having the order configurable in ENV conf with default set to OCP first?

AviadP: lgtm

petr-balogh: this are defined constants and this change is not going to be part of this PR.  This we discussed some time back but will be part of different PR and we need to figure out the approach which will work here. Anyway those are constants and can be used as default values if not provided by config. Then I can make it like read config values first if provided if not we will use those values.

petr-balogh:
Latest change tested with --colloect-only and I see that the tests are collected in proper order:
```
collected 383 items / 375 deselected / 8 selected
<Module tests/ecosystem/upgrade/test_configuration.py>
  <Function test_load_crush_map>
<Module tests/ecosystem/upgrade/test_noobaa.py>
  <Function test_fill_bucket>
<Module tests/ecosystem/upgrade/test_resources.py>
  <Function test_start_pre_upgrade_pod_io>
<Module tests/ecosystem/upgrade/test_upgrade_ocp.py>
  <Class TestUpgradeOCP>
      <Function test_upgrade_ocp>
<Module tests/ecosystem/upgrade/test_upgrade.py>
  <Function test_upgrade>
<Module tests/ecosystem/upgrade/test_configuration.py>
  <Function test_crush_map_unchanged>
<Module tests/ecosystem/upgrade/test_noobaa.py>
  <Function test_noobaa_postupgrade>
<Module tests/ecosystem/upgrade/test_resources.py>
  <Function test_pod_io>```

ocs_ci/ocs/constants.py
fbalak: Ok

vasukulkarni:
Isn't is just better to write that one test cases that requires ordering as a separate test or organize it in a folder where the ascii sort order will yield the required ordering like we have for maintenance tests in manage?

petr-balogh:
> Isn't is just better to write that one test cases that requires ordering as a separate test or organize it in a folder where the ascii sort order will yield the required ordering like we have for maintenance tests in manage?

The upgrade tests can be defined anywhere in the directory structure and as they are not related no reason have all of them in one folder. Even one pre upgrade test can be general for both OCP or OCS upgrade. We can even mark some other tier test as post upgrade if needed. 

I didn't change the logic here it's just about add OCP related upgrade stuff and will allow us to change the order dynamically as well. Like OCP upgrade will run after OCS and vice versa if we will make this ordering parametrized.

petr-balogh:
Verification job failed cause of problem with RC3 build I ran upgrade to but everything got executed fine. Marking as verified and will merge.

vasukulkarni:
> > Isn't is just better to write that one test cases that requires ordering as a separate test or organize it in a folder where the ascii sort order will yield the required ordering like we have for maintenance tests in manage?
> 
> The upgrade tests can be defined anywhere in the directory structure and as they are not related no reason have all of them in one folder. Even one pre upgrade test can be general for both OCP or OCS upgrade. We can even mark some other tier test as post upgrade if needed.
> 
> I didn't change the logic here it's just about add OCP related upgrade stuff and will allow us to change the order dynamically as well. Like OCP upgrade will run after OCS and vice versa if we will make this ordering parametrized.

Here is my 2 cents on this, A upgrade test has to be independent of other tests and is free to make calls to utilize them as needed by the testcase. if we try to decide the order by pymark it might work for 1 or 2 cases and still one has to look into those numbers to figure out what all is happening making debugging harder, As we add more cases it just becomes a spaghetti code with no clear structure.

Instead we have one folder for upgrade tests under manage and keep things clear.
  -> upgrade/
           v4.2/
               1_test_ocp_ocs_upgrade.py
               2_test_ocp_ocs_ocp_upgrade.py
               3_test ocs_ocp_upgrade.py
               4_test_ocs_ocp_upgrade_workload.py

Issue #1711, opened 2020-03-25
4.4 not supported yet?


zmc:
We tried to test a 4.4 build:
```
Traceback (most recent call last):
  File "/home/jenkins-build/workspace/ocs-ci/venv/bin/run-ci", line 11, in <module>
    load_entry_point('ocs-ci', 'console_scripts', 'run-ci')()
  File "/home/jenkins-build/workspace/ocs-ci/ocs_ci/framework/main.py", line 94, in main
    init_ocsci_conf(arguments)
  File "/home/jenkins-build/workspace/ocs-ci/ocs_ci/framework/main.py", line 81, in init_ocsci_conf
    load_config([version_config_file])
  File "/home/jenkins-build/workspace/ocs-ci/ocs_ci/framework/main.py", line 50, in load_config
    os.path.abspath(os.path.expanduser(config_file))
FileNotFoundError: [Errno 2] No such file or directory: '/home/jenkins-build/workspace/ocs-ci/conf/ocs_version/ocs-4.4.yaml'
```

I guess a file specifically for 4.4 needs to be added.

zmc:
https://github.com/red-hat-storage/ocs-ci/pull/1713

PR #1713, opened 2020-03-25
Add config for OCS 4.4

zmc:
Fixes: #1711

### DIFF
--- a/conf/ocs_version/ocs-4.4.yaml
+++ b/conf/ocs_version/ocs-4.4.yaml
@@ -1,0 +1,9 @@
+---
+DEPLOYMENT:
+  default_ocs_registry_image: "quay.io/rhceph-dev/ocs-olm-operator:latest-4.4"
+  default_latest_tag: 'latest-4.4'
+  ocs_csv_channel: "stable-4.4"
+ENV_DATA:
+  ocs_version: '4.4'
+REPORTING:
+  default_ocs_must_gather_latest_tag: 'latest-4.4'


conf/ocs_version/ocs-4.4.yaml
```diff
@@ -0,0 +1,10 @@
+---
+DEPLOYMENT:
+  default_ocs_registry_image: "quay.io/rhceph-dev/ocs-olm-operator:latest-4.4"
+  default_latest_tag: 'latest-4.4'
+  ocs_csv_channel: "stable-4.4"
+ENV_DATA:
+  logging_version: '4.4'
```
petr-balogh: I think this is related to the OCP version .
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/templates/openshift-infra/logging-deployment/elasticsearch_operator/eo-sub.yaml

But would be worth to check with someone who did this logging.

And if it's related to the OCP version I guess it should be deleted from this OCS related config and should be moved to the OCP related config file.

So I think that maybe if you will just delete this and let the default value use cause I guess you will fail with deployment cause of missing 4.4 channel. But it's just my guess.

vasukulkarni: I think you are right, maybe keep it as is which is '4.3' so that the tests pass.

petr-balogh: Please remove this value from here. This is also going to be deleted here:  https://github.com/red-hat-storage/ocs-ci/pull/1720/files

Issue #1659, opened 2020-03-16
MCG object integrity tests object sizes are too large and cause the tests to fail because of a timeout


Neon-White:
</title>

All object integrity tests failed because the sync commands all took longer than 10 minutes. We need to change the object sizes, or provide a longer timeout.
I support the first option because I think those tests should test functionality over velocity/scale.

PR #1717, opened 2020-03-26
Temporarily skip tests that utilize datasets that are too big 

MeridianExplorer:
...and take too long to run

A temporary workaround for #1659

### DIFF
--- a/tests/manage/mcg/test_object_integrity.py
+++ b/tests/manage/mcg/test_object_integrity.py
@@ -17,6 +17,7 @@ LARGE_FILE_KEY = "1000G_2504_high_coverage/data/ERR3239276/NA06985.final.cram"
 FILESIZE_SKIP = pytest.mark.skip('Current test filesize is too large.')
 RUNTIME_SKIP = pytest.mark.skip('Runtime is too long; Code needs to be parallelized')
 
+
 @filter_insecure_request_warning
 class TestObjectIntegrity(ManageTest):
     """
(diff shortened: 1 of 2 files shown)


tests/manage/mcg/test_object_integrity.py
```diff
@@ -63,23 +65,23 @@ def test_check_object_integrity(self, mcg_obj, awscli_pod, bucket_factory):
...
             ),
             pytest.param(
                 *[1000, 'small'],
-                marks=[pytest.mark.polarion_id("OCS-1951"), tier3]
+                marks=[pytest.mark.polarion_id("OCS-1951"), tier3, RUNTIME_SKIP]
             ),
             pytest.param(
                 *[100, 'large_small'],
-                marks=[pytest.mark.polarion_id("OCS-1952"), tier3]
+                marks=[pytest.mark.polarion_id("OCS-1952"), tier3, FILESIZE_SKIP]
```
RazTamir: Does this mean that all those tests will be skipped?

MeridianExplorer: Those tests will be skipped. This is not really a regression, because the filesize is impractically big. The solution is to change the amount of small files to a few tens of thousands, and have one "big" file (1-10GB, not 1.6TB like now).

fbalak:
I think that this would make sense for `test_s3_bucket_delete_1t_objects` in https://github.com/red-hat-storage/ocs-ci/blob/master/tests/manage/mcg/test_bucket_deletion.py. This test runs for a very long time (7+ hours). Or maybe to push it out of tier tests.

MeridianExplorer:
@fbalak - You're correct. I'll merge this (just to not lose the approval and runs) and add a new PR in a few minutes.

PR #1725, opened 2020-03-27
Update test_reclaim_policy to verify PV deletion

### LINKED ISSUE #1724: PV deletion is not verified in test_reclaim_policy.py
Test cases in  test_reclaim_policy.py is not verifying that PV is deleted which will lead to PV leftover issue. The leftover check is done immediately after PVC is deleted. There is chance that PV will take one or two seconds to get deleted. So it is important to verify PV deletion in test case.

jilju:
Fixes #1724 


vasukulkarni:
Please mark as verified

PR #1737, opened 2020-03-29
Fix MCG multi-region test

MeridianExplorer:
The `multiregion_mirror_setup` code was changed to return an object instead of a name. Updating the test code to comply with the change.


MeridianExplorer:
Test teardown failed because of [BZ 1817978](https://bugzilla.redhat.com/show_bug.cgi?id=1817978)
Otherwise - the test ran successfully.

PR #1742, opened 2020-03-29
Fixes: #1743 -  Fix vSphere get_vms() method for RHEL workers

ebenahar:

### DIFF
--- a/ocs_ci/ocs/platform_nodes.py
+++ b/ocs_ci/ocs/platform_nodes.py
@@ -111,7 +111,11 @@ class VMWareNodes(NodesBase):
             self.cluster_name, self.datacenter, self.cluster
         )
         node_names = [node.get().get('metadata').get('name') for node in nodes]
-        return [[vm for vm in vms_in_pool if vm.name in node] for node in node_names]
+        vms = []
+        for node in node_names:
+            node_vms = [vm for vm in vms_in_pool if vm.name in node]
+            vms.extend(node_vms)
+        return vms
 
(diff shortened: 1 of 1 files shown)


vasukulkarni:
@ebenahar can you test this locally if jenkins is going to take time?

ocs_ci/ocs/platform_nodes.py
```diff
@@ -111,7 +111,7 @@ def get_vms(self, nodes):
             self.cluster_name, self.datacenter, self.cluster
         )
         node_names = [node.get().get('metadata').get('name') for node in nodes]
-        return [vm for vm in vms_in_pool if vm.name in node_names]
+        return [[vm for vm in vms_in_pool if vm.name in node] for node in node_names]
```
clacroix12: Won't this return a list that contains a list per node we want to return? 
i.e. `[['node1'], ['node2']]`

I think what we want is something like this:

```python
vms = []
for node in node_names:
    node_vms = [vm for vm in vms_in_pool if vm.name in node]
    vms.extend(node_vms)
return vms
```

ebenahar: @clacroix12 great catch, thanks for the suggestion!

ebenahar:
Verified locally due to temp issue with:

https://openshift-release.svc.ci.openshift.org latest build information unavailable

[logs.zip](https://github.com/red-hat-storage/ocs-ci/files/4417117/logs.zip)

PR #1750, opened 2020-03-30
Fix max attempt check, range will return integers before final stop.

vasukulkarni:
Fixes: fixes possibility of false return when CSV is not yet upgraded.

### DIFF
--- a/tests/ecosystem/upgrade/test_upgrade.py
+++ b/tests/ecosystem/upgrade/test_upgrade.py
@@ -232,10 +232,8 @@ def test_upgrade():
         )
         if subscription_plan_approval == 'Manual':
             wait_for_install_plan_and_approve(namespace)
-        attempts = 145
-        for attempt in range(1, attempts):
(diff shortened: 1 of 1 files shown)


vasukulkarni:
Verified locally

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -232,10 +232,8 @@ def test_upgrade():
         )
         if subscription_plan_approval == 'Manual':
             wait_for_install_plan_and_approve(namespace)
-        attempts = 145
-        for attempt in range(1, attempts):
-            if attempts == attempt:
-                raise TimeoutException("No new CSV found after upgrade!")
+        attempts = 10
```
clacroix12: Why are we going from 145 attempts to 10?

vasukulkarni: good catch, was testing with lower value :), will fix.

PR #1754, opened 2020-03-31
Stabilize the MCG data reduction test

MeridianExplorer:
- Add a highly compressible text file that consistently compresses well
- Adjust the check threshold
- Improve log prints

Signed-off-by: Ben <belimele@redhat.com>

### DIFF
--- a/ocs_ci/ocs/resources/mcg.py
+++ b/ocs_ci/ocs/resources/mcg.py
@@ -295,7 +295,7 @@ class MCG(object):
 
         try:
             for total_size, total_reduced in TimeoutSampler(140, 5, _retrieve_reduction_data):
-                if total_size - total_reduced > 100*1024*1024:
+                if total_size - total_reduced > 100 * 1024 * 1024:
                     logger.info(
                         'Data reduced:' + str(total_size - total_reduced)
                     )
--- a/ocs_ci/ocs/resources/mcg.py
+++ b/ocs_ci/ocs/resources/mcg.py
@@ -288,23 +288,22 @@ class MCG(object):
             bucket_data_reduced = resp.json().get('reply').get('data').get('size_reduced')
 
             logger.info(
-                'Overall bytes stored: ' + str(bucket_data) + '. Amount reduced: ' + str(bucket_data_reduced)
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/resources/mcg.py
```diff
@@ -288,23 +288,22 @@ def _retrieve_reduction_data():
...
-                'Overall bytes stored: ' + str(bucket_data) + '. Amount reduced: ' + str(bucket_data_reduced)
+                'Overall bytes stored: ' + str(bucket_data) + '. Reduced size: ' + str(bucket_data_reduced)
             )
 
             return bucket_data, bucket_data_reduced
 
         try:
             for total_size, total_reduced in TimeoutSampler(140, 5, _retrieve_reduction_data):
-                if total_size - total_reduced > 80000000:
+                if total_size - total_reduced > 100 * 1024 * 1024:
```
RazTamir, line 298: I prefer not to use hardcoded for such things.
What will happen when using different clusters on different platforms?
2 suggestions:
1) use % instead of fixed number
2) if we are looking for data reduction to happen, we can sample the data reduction 3 times (for example) and in case in all 3 samples we can see reduction, it means it is happening

MeridianExplorer: AFAIK, different platforms should not affect the scale of "how much is compressed", but rather "how long it takes to reach finish compressing the data".

Regarding the suggestions:
1. Compression is extremely dependent on the type of data. Some types may not be compressed at all, some may compress to 20% of their size. Thus, the percentage will have to be hardcoded anyway, and on top of it, will have to be changed accordingly if we change the dataset (like I did in this PR).

2. I think that's a great idea, but it has a certain problem. Currently, the test looks for deduplication *and* compression. If either fails - the test fails. If we'll only check for reduction, we might miss a malfunction in either. 

Also, the reduction is mostly performed in one or two "pulses", so sampling over time isn't really meaningful. Usually, once you have two samples, it's already fully compressed.

In this case, I'd like to offer to rewrite the test - write one test dedicated for dedupe, another for compression, and both will look for *any* amount of data reduction. 

What do you think?

RazTamir: ack on all.
Continue as you see fit

PR #1755, opened 2020-03-31
Remove the 1TB object deletion test from Tier3 because of a long execution time

MeridianExplorer:
The test takes around 8 hours to execute because of the time it takes to retrieve and write all the objects to the NooBucket.
It should still be run outside of tiered runs in order to verify we don't experience a regression.


ebenahar:
@keesturam @ramkiperiy should we have this tested as part of scale tier?

ebenahar:
@MeridianExplorer let's have an issue tracking this removal

PR #1767, opened 2020-04-02
Change the data reduction threshold from 100MiB to 90MiB, add docs

MeridianExplorer:
The MCG data reduction test yields inconsistent compression results.
This threshold should be pretty sound. 86MiB is the mere minimum to test deduplication, and along with compression should be between 93-110MiBs.


MeridianExplorer:
PR closed; We shouldn't lower the bar to pass the test - we should find a consistent way of anticipating the compression and deduplication results.

PR #1770, opened 2020-04-02
Enable internal reportportal host

vasukulkarni:

### DIFF
--- a/ocs_ci/framework/main.py
+++ b/ocs_ci/framework/main.py
@@ -7,9 +7,10 @@ import pytest
 import yaml
 
 from ocs_ci import framework
+from getpass import getuser
+from ocs_ci.utility import utils
 from ocs_ci.ocs.constants import CONF_DIR
 from ocs_ci.ocs.exceptions import MissingRequiredConfigKeyError
-from ocs_ci.utility import utils
 
 
 def check_config_requirements():
@@ -94,10 +95,12 @@ def main(argv=None):
     init_ocsci_conf(arguments)
     pytest_logs_dir = utils.ocsci_log_path()
     utils.create_directory_path(framework.config.RUN['log_dir'])
+    launch_name = utils.get_testrun_name() + getuser()
     arguments.extend([
         '-p', 'ocs_ci.framework.pytest_customization.ocscilib',
         '-p', 'ocs_ci.framework.pytest_customization.marks',
         '-p', 'ocs_ci.framework.pytest_customization.reports',
         '--logger-logsdir', pytest_logs_dir,
+        '--rp-launch', launch_name,
     ])
     return pytest.main(arguments)
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ setup(
         'apache-libcloud',
         'docopt==0.6.2',
         'gevent==1.4.0',
-        'reportportal-client<4.0.0',
+        'reportportal-client==3.2.3',
         'requests==2.21.0',
         'paramiko==2.4.2',
         'pyyaml>=4.2b1',
@@ -26,7 +26,7 @@ setup(
         'boto3',
         'munch',
         'pytest==5.3.5',
-        'pytest-reportportal==1.0.5',
+        'pytest-reportportal==1.10.0',
         'pytest-logger',
         'dataclasses',  # For compatibility with python 3.6
         'pytest-html',
--- a/ocs_ci/framework/pytest_customization/reports.py
+++ b/ocs_ci/framework/pytest_customization/reports.py
@@ -2,6 +2,7 @@ import pytest
 import logging
 from py.xml import html
 from ocs_ci.utility.utils import email_reports
+from pytest_reportportal import RPLogHandler
 from ocs_ci.framework import config as ocsci_config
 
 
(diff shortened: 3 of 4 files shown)


ocs_ci/framework/main.py
```diff
@@ -94,10 +95,16 @@ def main(argv=None):
...
     utils.create_directory_path(framework.config.RUN['log_dir'])
+    launch_name = f"Tier run from {getuser()} RunId: {framework.config.RUN['run_id']}"
     arguments.extend([
         '-p', 'ocs_ci.framework.pytest_customization.ocscilib',
         '-p', 'ocs_ci.framework.pytest_customization.marks',
         '-p', 'ocs_ci.framework.pytest_customization.reports',
         '--logger-logsdir', pytest_logs_dir,
+        '--rp-launch', launch_name,
     ])
+    if '--reportportal' not in arguments:
```
clacroix12, line 106: Doesn't this make running without posting to report portal impossible? I'm not sure I like the idea of forcing users to post to report portal if they don't want to use this functionality.

ocs_ci/framework/main.py
```diff
@@ -94,10 +95,16 @@ def main(argv=None):
     init_ocsci_conf(arguments)
     pytest_logs_dir = utils.ocsci_log_path()
     utils.create_directory_path(framework.config.RUN['log_dir'])
+    launch_name = f"Tier run from {getuser()} RunId: {framework.config.RUN['run_id']}"
```
clacroix12, line 98: We don't know if this is a Tier run or a simple deployment so I think using a launch name like this could be misleading. Furthermore if we include the `run_id` in the launch name we won't be able to utilize the history feature in report portal. The launch name should be the same as other executions of the same nature. Report portal will append a number to the launch name as we execute several launches with the same name. This allows us to view the history and compare results from iteration to iteration.

I can see us writing a function that will use several items from the configuration and from the provided arguments in order to determine exactly what kind of deployment/tier test this is and then choose the correct launch name to provide to `--rp-launch`

vasukulkarni: This is just an initial version, I want to have something similar to what our email subject says but that part of the code is all across repo.  If you have the code handy send me :)

ocs_ci/framework/main.py
vasukulkarni: I would like to see this enabled for all runs, we currently dont have visibility outside jenkins. For accurate failure rate we should see results from across. The setting ignores failures due to reportportal issues but if there are issues its not hard to disable.

ocs_ci/framework/main.py
clacroix12: I don't have anything handy at the moment. I just wanted to raise this point early as it's fairly important to being able to fully leverage report portal.

ocs_ci/framework/main.py
clacroix12: This forces every execution to push to report portal, even local executions. Would it be acceptable if we had `--report-portal` configured by default for all of our jenkins jobs instead?

ocs_ci/framework/main.py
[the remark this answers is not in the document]
vasukulkarni: Yes I understand but that is available only for jenkins and we need something here in ocs-ci so it has to be written fresh if i am not mistaken. If its easier for you to add a function by copy/pasting the code in ocs-ci please do so, then we will have it in only one place. The tags and other things can be added as needed with follow on PRs.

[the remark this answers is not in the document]
clacroix12: We could try to pull that value (which would work in the event of a jenkins execution) and if it doesn't exist we could construct the name. We do something like this for email subject/polarion testrun name: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L1181-L1230

This won't work entirely as is because it includes information we would rather tag launches with than include in the launch name, but this at least gives you a good idea of what we do to retrieve this data.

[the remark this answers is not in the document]
vasukulkarni: +1 I am picking up the above function and putting in RUN config.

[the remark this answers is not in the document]
clacroix12: Doesn't it make more sense to stay in the `REPORTING` section?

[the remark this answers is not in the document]
vasukulkarni: I didn't have to use anything.

ocs_ci/framework/main.py
vasukulkarni: I will say, lets load our DB, we can revisit this in a weeks time , there is nothing wrong in having dashboard integrated by default.

setup.py
```diff
@@ -17,7 +17,7 @@
         'apache-libcloud',
         'docopt==0.6.2',
         'gevent==1.4.0',
-        'reportportal-client<4.0.0',
+        'reportportal-client==3.2.3',
```
clacroix12, line 20: Are we limited here due to the version of report portal running on cistatus?

vasukulkarni: @clacroix12 that is correct, the newer version has moved to pgsql and we are at 4.x

vasukulkarni:
Verified with: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/6749/ ,  I see the entries in RP

vasukulkarni:
Screen Shot 2020-04-16 at 11 27 01 AM

ocs_ci/framework/main.py
mbukatov: And would you be able to filter out the local runs in the report portal? I can imagine that my local runs which uses not yet comited code for test development purposes could mess with the data we see there.

vasukulkarni: yeah we can have multiple filters, so its easy to filter things out, Personally played with tons of data in ceph. Right now this is disabled by default so it requires explicit option for now.

PR #1781, opened 2020-04-06
Added configuration option to preserve the bootstrap node

clacroix12:
Jira: [OCSQE-293](https://projects.engineering.redhat.com/browse/OCSQE-293)

### DIFF
--- a/conf/ocsci/preserve_bootstrap_node.yaml
+++ b/conf/ocsci/preserve_bootstrap_node.yaml
@@ -1,0 +1,2 @@
+DEPLOYMENT:
+  preserve_bootstrap_node: True
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -198,6 +198,18 @@ class AWSIPI(AWSBase):
         def __init__(self):
             super(AWSIPI.OCPDeployment, self).__init__()
 
+        def deploy_prereq(self):
+            """
+            Overriding deploy_prereq from parent. Perform all necessary
+            prerequisites for AWSIPI here.
+            """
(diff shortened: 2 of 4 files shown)


clacroix12:
I have tested this locally with AWS IPI and UPI, as well as tested with vsphere. I was able to confirm that the bootstrap node remains up when configured to preserve.

clacroix12:
> @clacroix12 ,
> 
> Consider the case of AWSUPI deployment, suppose if there is any failure during deployment and we invoke teardown from Jenkins, so in this case I assume bootstrap node will be preserved , is my assumption correct?

The default configuration is to not preserve the bootstrap node, this is only done if you explicitly pass `conf/ocsci/preserve_bootstrap_node.yaml` or another similar config file at execution. Furthermore, this only affects the deployment of the cluster. Teardown remains unaffected and should (and does) clean up the bootstrap node as it normally would. Specifically for AWS UPI we run the `destroy cluster` cmd as well as explicitly destroy the bootstrap cloudformation stack during teardown.

clacroix12:
> PR validation [Job FAILED](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/1439//testReport) (installation failed, tests not executed).

OCS install failed but I can verify from the logs that we set the env var to preserve the bootstrap node due to the provided config file. Verifying.

clacroix12:
The travis CI issue (unrelated to this PR) has also been reported here: #1782 
With a fix for it here: #1784

clacroix12:
@shylesh @vasukulkarni rebased on master to fix Travis CI failure so your approvals have been dismissed.

PR #1801, opened 2020-04-08
First pass at scale helper function changes for pvc creation and

wusui:
deletion tests


wusui:
A better version of this has been pushed (see #2053)

PR #1804, opened 2020-04-08
Disable rhcos_ami

shylesh:


conf/ocp_version/ocp-4.2-config.yaml
```diff
@@ -6,3 +6,6 @@ DEPLOYMENT:
   installer_version: "4.2.0-0.nightly"
 ENV_DATA:
   vm_template: 'rhcos-4.2.0-x86_64-vmware'
+
+# rhcos_ami version for aws upi deployments will be fetched from
```
petr-balogh, line 10: I would keep example like:
```
# rhcos_ami: 'ami-06c85f9d106577272'
# If rhcos_ami is not specified the ID for aws upi deployments will be fetched from
# https://raw.githubusercontent.com/openshift/installer/release-4.2/data/data/rhcos.json
```

PR #1807, opened 2020-04-08
Add Ceph health check in environment_check as a finalizer for all test

RazTamir:

### DIFF
--- a/ocs_ci/utility/environment_check.py
+++ b/ocs_ci/utility/environment_check.py
@@ -8,6 +8,7 @@ import yaml
 from gevent.threadpool import ThreadPoolExecutor
 
 from ocs_ci.ocs import ocp, defaults, constants, exceptions
+from ocs_ci.utility.utils import ceph_health_check
 
 log = logging.getLogger(__name__)
 
@@ -184,10 +185,15 @@ def get_status_after_execution():
                 'Leftovers removed'
             ].append({f"***{kind}***": kind_diff[1]})
             leftover_detected = True
-    if leftover_detected:
-        raise exceptions.ResourceLeftoversException(
-            f"\nThere are leftovers in the environment after test case:"
-            f"\nResources added:\n{yaml.dump(leftovers['Leftovers added'])}"
-            f"\nResources "
-            f"removed:\n {yaml.dump(leftovers['Leftovers removed'])}"
-        )
+    try:
+        log.info("Checking Ceph health")
+        ceph_health_check(tries=6, delay=10)
+        log.info("Ceph cluster health is OK")
+    finally:
+        if leftover_detected:
+            raise exceptions.ResourceLeftoversException(
+                f"\nThere are leftovers in the environment after test case:"
+                f"\nResources added:\n{yaml.dump(leftovers['Leftovers added'])}"
+                f"\nResources "
+                f"removed:\n {yaml.dump(leftovers['Leftovers removed'])}"
+            )
--- a/ocs_ci/utility/environment_check.py
+++ b/ocs_ci/utility/environment_check.py
@@ -8,7 +8,6 @@ import yaml
 from gevent.threadpool import ThreadPoolExecutor
 
 from ocs_ci.ocs import ocp, defaults, constants, exceptions
-from ocs_ci.utility.utils import ceph_health_check
 
 log = logging.getLogger(__name__)
 
@@ -185,15 +184,10 @@ def get_status_after_execution():
                 'Leftovers removed'
             ].append({f"***{kind}***": kind_diff[1]})
             leftover_detected = True
-    try:
-        log.info("Checking Ceph health")
-        ceph_health_check(tries=6, delay=10)
-        log.info("Ceph cluster health is OK")
-    finally:
-        if leftover_detected:
-            raise exceptions.ResourceLeftoversException(
-                f"\nThere are leftovers in the environment after test case:"
-                f"\nResources added:\n{yaml.dump(leftovers['Leftovers added'])}"
-                f"\nResources "
-                f"removed:\n {yaml.dump(leftovers['Leftovers removed'])}"
-            )
+    if leftover_detected:
+        raise exceptions.ResourceLeftoversException(
+            f"\nThere are leftovers in the environment after test case:"
+            f"\nResources added:\n{yaml.dump(leftovers['Leftovers added'])}"
+            f"\nResources "
+            f"removed:\n {yaml.dump(leftovers['Leftovers removed'])}"
+        )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -956,7 +956,7 @@ def cluster(request, log_cli_level):
         deployer.deploy_cluster(log_cli_level)
 
 
-@pytest.fixture(scope='class')
+@pytest.fixture()
 def environment_checker(request):
     node = request.node
     # List of marks for which we will ignore the leftover checker
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,7 +12,9 @@ from datetime import datetime
 import random
 from math import floor
 
-from ocs_ci.utility.utils import TimeoutSampler, get_rook_repo, get_ocp_version
+from ocs_ci.utility.utils import (
+    TimeoutSampler, get_rook_repo, get_ocp_version, ceph_health_check
+)
 from ocs_ci.ocs.exceptions import TimeoutExpiredError, CephHealthException
 from ocs_ci.utility.spreadsheet.spreadsheet_api import GoogleSpreadSheetAPI
 from ocs_ci.utility import aws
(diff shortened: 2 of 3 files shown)


jilju:
This check will run one time for each test class, right ? Not after each parametrized test case in the test class.

RazTamir:
Nice catch!

Here is the fixture:
```
@pytest.fixture(scope='class')
def environment_checker(request):
    node = request.node
    # List of marks for which we will ignore the leftover checker
    marks_to_ignore = [m.mark for m in [deployment, ignore_leftovers]]
    for mark in node.iter_markers():
        if mark in marks_to_ignore:
            return

    request.addfinalizer(get_status_after_execution)
    get_status_before_execution()
```
So class indeed.
Will send a PR to fix it

ocs_ci/utility/environment_check.py
```diff
@@ -184,10 +185,15 @@ def get_status_after_execution():
...
-        raise exceptions.ResourceLeftoversException(
-            f"\nThere are leftovers in the environment after test case:"
-            f"\nResources added:\n{yaml.dump(leftovers['Leftovers added'])}"
-            f"\nResources "
-            f"removed:\n {yaml.dump(leftovers['Leftovers removed'])}"
-        )
+    try:
+        log.info("Checking Ceph health")
+        ceph_health_check(tries=6, delay=10)
+        log.info("Ceph cluster health is OK")
```
ebenahar, line 191: ceph_health_check returns bool, hence not sure if try is needed. 
how about:

```python
if not ceph_health_check(tries=6, delay=10):
    log.info("Ceph cluster health is not OK")
```

RazTamir: The try is to ensure to run the leftover_detected part.

vasukulkarni: I think try is not going to fail here because we are not really checking the return status of ceph_health_check, so I agree try is not needed here.

tests/conftest.py
```diff
@@ -956,7 +956,7 @@ def cluster_teardown_finalizer():
         deployer.deploy_cluster(log_cli_level)
 
 
-@pytest.fixture(scope='class')
+@pytest.fixture()
```
vasukulkarni: Not sure if its good idea to change this now, it will require all tier testing. Also, we want the scope to be class since we will have dependency across test within the same class, so some test may fail unnecessarily.

RazTamir: realized this after the job verification - thanks.
Fixed

tests/conftest.py
```diff
@@ -910,6 +912,7 @@ def tier_marks_name():
 @pytest.fixture(scope='function', autouse=True)
 def health_checker(request, tier_marks_name):
     node = request.node
+    request.addfinalizer(ceph_health_check)
```
ebenahar, line 915: ceph_health_check doesn't throw an exception, just returns bool, so here, it will have no effect in case ceph health if not OK.
A solution would be to create a finalizer function that calls ceph_health_check and asserts its result

RazTamir: ceph_health_check return what is being returned from ceph_health_check_base:
```
def ceph_health_check_base(namespace=None):
    """
    Exec `ceph health` cmd on tools pod to determine health of cluster.

    Args:
        namespace (str): Namespace of OCS
            (default: config.ENV_DATA['cluster_namespace'])

    Raises:
        CephHealthException: If the ceph health returned is not HEALTH_OK
        CommandFailed: If the command to retrieve the tools pod name or the
            command to get ceph health returns a non-zero exit code
    Returns:
        boolean: True if HEALTH_OK

    """
```
Which means a CephHealthException is thrown when the state is not OK

ebenahar: maybe I'm wrong but IIUC, ocs_ci.utility.retry.retry catches the exception raised by ocs_ci.utility.utils.ceph_health_check_base and returns False in case the cluster health is not OK

RazTamir:
Example from the console output:

------------------------------ live log teardown -------------------------------
20:38:41 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc wait --for condition=ready pod -l app=rook-ceph-tools -n openshift-storage --timeout=120s
20:38:41 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage get pod -l 'app=rook-ceph-tools' -o jsonpath='{.items[0].metadata.name}'
20:38:42 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage exec rook-ceph-tools-57fd596bb4-9bb4p ceph health
20:38:43 - MainThread - ocs_ci.utility.retry - WARNING - Ceph cluster health is not OK. Health: HEALTH_WARN Degraded data redundancy: 304/19470 objects degraded (1.561%), 4 pgs degraded
, Retrying in 30 seconds...
20:39:13 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc wait --for condition=ready pod -l app=rook-ceph-tools -n openshift-storage --timeout=120s
20:39:13 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage get pod -l 'app=rook-ceph-tools' -o jsonpath='{.items[0].metadata.name}'
20:39:13 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage exec rook-ceph-tools-57fd596bb4-9bb4p ceph health
20:39:14 - MainThread - ocs_ci.utility.retry - WARNING - Ceph cluster health is not OK. Health: HEALTH_WARN Degraded data redundancy: 203/19473 objects degraded (1.042%), 3 pgs degraded
, Retrying in 30 seconds...
20:39:44 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc wait --for condition=ready pod -l app=rook-ceph-tools -n openshift-storage --timeout=120s
20:39:44 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage get pod -l 'app=rook-ceph-tools' -o jsonpath='{.items[0].metadata.name}'
20:39:45 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage exec rook-ceph-tools-57fd596bb4-9bb4p ceph health
20:39:46 - MainThread - ocs_ci.utility.retry - WARNING - Ceph cluster health is not OK. Health: HEALTH_WARN Degraded data redundancy: 43/19479 objects degraded (0.221%), 1 pg degraded, 1 pg undersized
, Retrying in 30 seconds...
20:40:16 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc wait --for condition=ready pod -l app=rook-ceph-tools -n openshift-storage --timeout=120s
20:40:16 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage get pod -l 'app=rook-ceph-tools' -o jsonpath='{.items[0].metadata.name}'
20:40:17 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage exec rook-ceph-tools-57fd596bb4-9bb4p ceph health
20:40:17 - MainThread - ocs_ci.utility.utils - INFO - Ceph cluster health is HEALTH_OK.

RazTamir:
The verification is in the logs.
The job finished unstable as this is exactly the intention for this PR --> to improve tests teardown on health warning

tests/conftest.py
RazTamir:
```
def retry(exception_to_check, tries=4, delay=3, backoff=2):
    """
    """
    def deco_retry(f):

        @wraps(f)
        def f_retry(*args, **kwargs):
            mtries, mdelay = tries, delay
            while mtries > 1:   <<--- while there are still tries, it will catch the exception and the call is ** return f(*args, **kwargs) **
                try:
                    return f(*args, **kwargs)
                except exception_to_check as e:
                    logger.warning("%s, Retrying in %d seconds..." % (str(e), mdelay))
                    time.sleep(mdelay)
                    mtries -= 1
                    mdelay *= backoff
            return f(*args, **kwargs)  <<--- when no tries left, it will call it once more with no try: except: which means it will trow the exception 

        return f_retry

    return deco_retry
```

ebenahar: @RazTamir Thanks! just validated it myself and you are right :)

tests/conftest.py
```diff
@@ -909,18 +911,29 @@ def tier_marks_name():
...
+    def finalizer():
+        try:
+            ceph_health_check_base()
+            log.info("Ceph health check passed at teardown")
+        except CephHealthException:
+            log.info("Ceph health check failed at teardown")
+            # Retrying to increase the chance the cluster health will be OK
+            # for next test
+            ceph_health_check()
+            raise
```
vasukulkarni, line 923: Just thinking if we can suppress 'raise' here in case above check passed?

RazTamir: The idea is to fail the test if we are in the 'except' block

PR #1817, opened 2020-04-12
Add capacity test: Fix Ceph health check

ebenahar:

### DIFF
--- a/tests/manage/z_cluster/cluster_expansion/test_add_capacity.py
+++ b/tests/manage/z_cluster/cluster_expansion/test_add_capacity.py
@@ -4,7 +4,7 @@ from ocs_ci.framework.testlib import ignore_leftovers, ManageTest, tier1
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs.resources import storage_cluster
-from ocs_ci.ocs.cluster import CephCluster
+from ocs_ci.utility.utils import ceph_health_check
 
 
 @ignore_leftovers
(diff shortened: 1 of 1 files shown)


tests/manage/z_cluster/cluster_expansion/test_add_capacity.py
```diff
@@ -30,4 +29,6 @@ def test_add_capacity(self):
             selector='app=rook-ceph-osd',
             resource_count=result * 3
         )
-        self.ceph_cluster.cluster_health_check(timeout=1200)
+        assert ceph_health_check(
+            namespace=config.ENV_DATA['cluster_namespace'], tries=80
+        ), f"Ceph health is not OK"
```
RazTamir, line 34: also taking care of in the function call:
```
raise CephHealthException(
            f"Ceph cluster health is not OK. Health: {health}"
        )
```

tests/manage/z_cluster/cluster_expansion/test_add_capacity.py
```diff
@@ -30,4 +29,6 @@ def test_add_capacity(self):
             selector='app=rook-ceph-osd',
             resource_count=result * 3
         )
-        self.ceph_cluster.cluster_health_check(timeout=1200)
+        assert ceph_health_check(
```
RazTamir, line 32: no need assert imo.
This call will raise 'CephHealthException' in case not healthy and return True when does healthy so no chance to get False for the assert

tests/manage/z_cluster/cluster_expansion/test_add_capacity.py
ebenahar: ack

Issue #1818, opened 2020-04-12
[MCG] Status test needs redesigning


MeridianExplorer:
Right now, the `test_verify_noobaa_status` test fails because the amounts and names of resources change between different versions. The test needs to be redesigned in a way that is agnostic to the amounts, or knows them in advance.

MeridianExplorer:
Addressed in #2011

PR #1828, opened 2020-04-13
[MCG] Exec CLI commands through the NooBaa operator pod's binary

MeridianExplorer:
<Title>

Removes the unnecessary dependency on a local binary that's also dependent on the system of the code runner.

Also had to change an existing test's code (`test_registry_image_push_pull.py`) to comply with the changes.


ocs_ci/ocs/resources/mcg.py
```diff
@@ -35,6 +37,10 @@ def __init__(self):
         """
         Constructor for the MCG class
         """
+
+        # To clean the logs from warnings related to the lack of SSL certs
```
fbalak, line 41: Can we address these warnings by generating and configuring correct certificate? I am not sure if dropping warnings about not using SSL is a customer scenario.
If we want to suppress these messages for now, I would suggest to create issue for this, todo comment into code and a warning log message that SSL warnings are suppressed.

MeridianExplorer: I talked with NooBaa, and it's possible to generate a certificate, and NB should even automatically pull and use it. 
The comment is inaccurate - it doesn't stem from a lack of a cert, but rather by an inability to verify it, because it's self-signed.
However, I honestly don't know how easy it is to perform this per-cluster, or whether it's even possible to automate this, or do it quickly enough (generating a cert that isn't self-signed).

I chose to drop the warnings in order to reduce the log size and make it cleaner. How may it affect our actual testing in regards to mimicking customer scenarios?

MeridianExplorer: Regarding CLI usage - we won't be using any local binary anymore. All commands will run on the binary inside the operator's pod.

fbalak: Ok, thank you.

vasukulkarni:
How does this ensure the correct mcg binary is used from rhel repo?

vasukulkarni:
> How does this ensure the correct mcg binary is used from rhel repo?, The test should be using the binary that comes from rhel repo.

Please ignore, got answer from Petr.

MeridianExplorer:
Blocked by #1877 since the status test needs the `stderr` output to function with `exec_cmd`

ocs_ci/ocs/defaults.py
```diff
@@ -40,6 +40,6 @@
 NOOBAA_STATUS_CONTENT_COUNT = {
     'CustomResourceDefinition': 5, 'Namespace': 1, 'ServiceAccount': 1,
     'Role': 1, 'RoleBinding': 1, 'ClusterRole': 1, 'ClusterRoleBinding': 1,
-    'Deployment': 1, 'NooBaa': 1, 'StatefulSet': 1, 'Service': 2, 'Secret': 3,
+    'Deployment': 2, 'NooBaa': 1, 'StatefulSet': 2, 'Service': 3, 'Secret': 4,
```
MeridianExplorer, line 43: #1818 Regarding "why is it designed this way"
The code change here is merely to make it compatible with 4.3

MeridianExplorer:
`test_verify_noobaa_status` tested locally - 
```
=============================== warnings summary ===============================
myvenv/lib/python3.7/site-packages/_pytest/config/__init__.py:891
  /home/newuser/repos/ocs-ci-personal-fork/myvenv/lib/python3.7/site-packages/_pytest/config/__init__.py:891: PytestAssertRewriteWarning: Module already imported so cannot be rewritten: pytest_reportportal
    self._mark_plugins_for_rewrite(hook)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
======================== 1 passed, 1 warning in 45.35s =========================
```

MeridianExplorer:
Travis is falsely stuck on `pending`, when entering the link, it reports a complete success

vasukulkarni:
@MeridianExplorer can you force push again? thanks

Issue #1830, opened 2020-04-14
deploy cluster with skip OCS fails on health_check


vavuthu:
>               f"Error during execution of command: {masked_cmd}."
                f"\nError is {masked_stderr}"
            )
E           ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc wait --for condition=ready pod -l app=rook-ceph-tools -n openshift-storage --timeout=120s.
E           Error is error: no matching resources found

ocs_ci/utility/utils.py:430: CommandFailed

> since its only OCP deployment, toolbox didn't exist

https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/6639/console

PR #1832, opened 2020-04-14
Fixing skip_ocs_deployment teardown

vavuthu:
When we use skip_ocs_deployment, teardown is checking for
health_check of OCS cluster. In this PR, we are skipping health_check
for skip_ocs_deployment.

Fixes: #1830

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -915,7 +915,7 @@ def health_checker(request, tier_marks_name):
         try:
             teardown = config.RUN['cli_params']['teardown']
             skip_ocs_deployment = config.ENV_DATA['skip_ocs_deployment']
-            if not ( teardown or skip_ocs_deployment ):
+            if not (teardown or skip_ocs_deployment):
                 ceph_health_check_base()
                 log.info("Ceph health check passed at teardown")
         except CephHealthException:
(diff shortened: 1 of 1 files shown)


petr-balogh:
@vavuthu did you run some verification of this? If so, can you please link the job?

tests/conftest.py
```diff
@@ -914,7 +914,8 @@ def health_checker(request, tier_marks_name):
     def finalizer():
         try:
             teardown = config.RUN['cli_params']['teardown']
-            if not teardown:
+            skip_ocs_deployment = config.ENV_DATA['skip_ocs_deployment']
```
petr-balogh, line 917:
```suggestion
            skip_ocs_deployment = config.ENV_DATA.get('skip_ocs_deployment', False)
```

petr-balogh: I double checked the default config and seems we have those default values in the config so in this case I will not block this PR as it's probably can block engineering as they have separated pipeline of OCP/OCS deployment. So I will go ahead and merge this one as I see the verification job passed.  But anyway it will be better to change it to get method as this will not rely on this config value if not specified.

vavuthu: Deploy job passed with skip_ocs_deployment: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/6660/consoleFull

By default, skip_ocs_deployment will be False. Even for engineering I believe all the default values ( default_config.yaml ) will be loaded.

Issue #1729, opened 2020-03-27
Post upgrade verification should wait more time for Ceph HEALTH_OK


petr-balogh:
Based on discussion in BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1817727

We should wait more time if we run FIO during the upgrade.

petr-balogh:
Still happening in upgrade from time to time. We need to increase timeout more.

PR #1839, opened 2020-04-15
Increase timeout for upgrade

petr-balogh:
Fixes: #1729

### DIFF
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -342,7 +342,7 @@ def ocs_install_verification(
         # In case of upgrade with FIO we have to wait longer time to see
         # health OK. See discussion in BZ:
         # https://bugzilla.redhat.com/show_bug.cgi?id=1817727
-        health_check_tries = 60
+        health_check_tries = 180
     assert utils.ceph_health_check(
         namespace, health_check_tries, health_check_delay
     )


ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -211,7 +211,7 @@ def ocs_install_verification(
         # In case of upgrade with FIO we have to wait longer time to see
         # health OK. See discussion in BZ:
         # https://bugzilla.redhat.com/show_bug.cgi?id=1817727
-        health_check_tries = 60
+        health_check_tries = 180
```
RazTamir, line 214: @keesturam, do you think we need to report a bug on post upgrade "recovery" time?

petr-balogh: You can se BZ one line above:  https://bugzilla.redhat.com/show_bug.cgi?id=1817727

And discussion can be added there.

The recovery time depends on the how we are saturating the cluster during the FIO I guess.
And I can see this is more likely happening on VmWare.

As I am not workload specialist and I have no clue how much do we saturating the cluster and if the FIO is still running during the upgrade how long it can take that OSD will be recovered.

Maybe worth to someone will take look at the cluster while we are running the upgrade. But this is not 100% reproducible, sometime we passed without any issue and sometime not.

RazTamir: Thanks @petr-balogh, missed that

shylesh: @petr-balogh ,
How about coming up with an adaptive delay based on the progress made during recovery. 
I guess we can calculate the diff in the recovery progress (may be using ceph status) during fixed intervals and if there is a progress then increase the timeout else leave the timeout as is.

thoughts?

shylesh: This is the output from the above bug you mentioned.

"HEALTH_WARN Degraded data redundancy: 300/69912 objects degraded (0.429%), 4 pgs degraded, 2 pgs undersized"

The quickly changing item should be number of objects, so we can capture that in fixed intervals and find the diff which can tell us if there is any progress or not(this interval should be reasonable assuming each object size would be 4MB in the worst case) and based on this we can change the timeout value per iteration.

May not be for this PR but just a thought.

vasukulkarni: @shylesh good point, probably file an issue, we also need to experiment with VMware DC's more as they are varying in performance.

vasukulkarni:
Worth running one test on a DC?

Issue #1850, opened 2020-04-16
aws-cleanup failing with ocs_ci.ocs.exceptions.ClusterIOTagNotFound: Unable to find valid cluster IO tag from ec2 instance tags


dahorak:
aws-cleanup script sometimes fails with following error ([full log](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-aws-cleanup/901//consoleFull)):
```
[...]

PR #1858, opened 2020-04-16
Log warning instead of raising Error if cluster isn't an OCS cluster

clacroix12:
Fixes: #1850 


clacroix12:
I verified this works locally and that we see the error message for non OCS clusters / clusters that we are unable to find this tag for.

```
2020-04-16 15:01:48,798 - MainThread - ocs_ci.cleanup.aws.cleanup - WARNING - Unable to find valid cluster IO tag from ec2 instance tags for VPC dahorak-vpc. This is probably not an OCS cluster VPC!
```

Issue #1873, opened 2020-04-20
Validation of registry pod is backed by OCS is failing with: 'NoneType' is not iterable


petr-balogh:
Jenkins job: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/6756//console

```
        """
        pod_objs = get_registry_pod_obj()
        for pod_obj in pod_objs:
            mount_point = pod_obj.exec_cmd_on_pod(command="mount")
>           assert "/registry" in mount_point, (
                f"pvc is not mounted on pod {pod_obj.name}"
            )
E           TypeError: argument of type 'NoneType' is not iterable
```

PR #1875, opened 2020-04-20
Fix issue with mount command

petr-balogh:
Fixes: #1873


clacroix12:
@petr-balogh deployment test passed, was there additional testing you wanted to do with this or can we mark as verified / merge?

petr-balogh:
Deployment test was enough ;) merging

PR #1877, opened 2020-04-21
Replace run_cmd with exec_cmd

MeridianExplorer:
Up until today, `run_cmd` didn't return anything but a command's resulting `stdout`.
However, after discussion in the team chat, it was agreed that it'd be more beneficial to return the entirety of the `CompletedProcess` object, for the developers to dissect and use as they wish afterwards.

Thus, `exec_cmd` was created as a replacement that returns the entire object. Backwards compatibility is achieved by calling the new function inside `run_cmd` and then extracting the `stdout` and returning it.

I tried to use `warnings` in order to deprecate the old function, but this resulted in a lot of messages filling up the log, and I'm not sure whether it's a good idea to filter all deprecation warnings. I tried filtering by module, but it didn't seem to have effect.

So far I tested the change on a few tests locally, and everything seems to do well. Any idea what further testing will verify that the changes are correct?

*Also optimized imports in our `utils.py`

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -423,10 +423,10 @@ def exec_cmd(cmd, secrets=None, timeout=600, ignore_error=False, **kwargs):
     Returns:
         (CompletedProcess) A CompletedProcess object of the command that was executed
         CompletedProcess attributes:
-            args: The list or str args passed to run().
-            returncode (str): The exit code of the process, negative for signals.
-            stdout     (str): The standard output (None if not captured).
-            stderr     (str): The standard error (None if not captured).
+        args: The list or str args passed to run().
+        returncode (str): The exit code of the process, negative for signals.
+        stdout     (str): The standard output (None if not captured).
+        stderr     (str): The standard error (None if not captured).
 
     """
     masked_cmd = mask_secrets(cmd, secrets)
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -421,7 +421,7 @@ def exec_cmd(cmd, secrets=None, timeout=600, ignore_error=False, **kwargs):
         CommandFailed: In case the command execution fails
 
     Returns:
-        (str) Decoded stdout of command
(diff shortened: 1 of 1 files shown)


ocs_ci/utility/utils.py
```diff
@@ -406,31 +428,32 @@ def run_cmd(cmd, secrets=None, timeout=600, ignore_error=False, **kwargs):
...
-    log.debug(f"Command return code: {r.returncode}")
-    if r.returncode and not ignore_error:
+    log.debug(f"Command return code: {completed_process.returncode}")
+    if completed_process.returncode and not ignore_error:
         raise CommandFailed(
             f"Error during execution of command: {masked_cmd}."
             f"\nError is {masked_stderr}"
         )
-    return masked_stdout
+    return completed_process
```
RazTamir, line 456: please fix the returned value in the doc

MeridianExplorer: Already done 👍

RazTamir: Also, as we need to mask the output, I wonder if it will be better to return a tuple of (rc, out, err) all masked

MeridianExplorer: I think it can actually be valuable for us to have programmatic access to the unmasked variables. It's still masked in the logs, and thus, the only way to "leak" any values would be to intentionally log again what we already logged in `exec_cmd`. My mindset is to keep the object as untouched as possible, because every processing we choose to do causes a loss of information, which we might potentially need down the road (which is the reason for this PR).

ebenahar: @MeridianExplorer how about adding an example for how to unpack completed_process?

MeridianExplorer: @ebenahar Added :+1:

vasukulkarni:
> > I am expecting atleast couple of tests will fail :)
> 
> @vasukulkarni
> Anything in particular, besides the regular flakiness? ;)

will wait and watch for tier1 run to complete, the last one looks like it failed during installation.

PR #1899, opened 2020-04-23
Test-Verify RWO volumes are exclusively mounted by adding replica 2 i…

jilju:
…n pod dc

OCS-896 - RBD: Verify RWO volumes are exclusively mounted by adding replicas - 2 in pod dc
OCS-897 - CephFS: Verify RWO volumes are exclusively mounted by adding replicas - 2 in pod dc

### DIFF
--- a/tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
+++ b/tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
@@ -1,0 +1,122 @@
+import logging
+import pytest
+from itertools import cycle
+
+from ocs_ci.framework.testlib import ManageTest, tier2
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.exceptions import (
+    ResourceWrongStatusException, UnexpectedBehaviour
+)
+from ocs_ci.ocs.resources.ocs import OCS
+from ocs_ci.ocs.resources import pod
+from ocs_ci.ocs.ocp import OCP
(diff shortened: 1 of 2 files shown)


tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
```diff
@@ -0,0 +1,122 @@
...
+                break
+
+        # Verify the other pod is not coming up Running
+        try:
+            self.pod_not_running = next(pods_iter)
+            wait_for_resource_state(
+                resource=self.pod_not_running, state=constants.STATUS_RUNNING,
+                timeout=10
+            )
+            raise UnexpectedBehaviour(
```
ebenahar, line 101: it seems that this exception will be thrown in any case. Could you please clarify?

jilju: Expecting ResourceWrongStatusException from wait_for_resource_state.
If that exception did not occur, it is considered as failure and  UnexpectedBehaviour .

In other words, in the normal flow as expected by the test case, line 101 won't be executed.

ebenahar: got you :)
Please add this explanation to comment in the code

jilju: Done

Issue #1902, opened 2020-04-23
CSRs are approving twice.


vavuthu:
CSRs are approving twice.

approve_pending_csr() - approving all the existing CSR .
Instead we need to approve only pending CSRs

https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/6938/consoleFull

PR #1903, opened 2020-04-23
Fix in Approving CSRs

vavuthu:
* Current CSR functionality approves the alreday
  approved CSR.

Fixes: #1902

### DIFF
--- a/ocs_ci/utility/csr.py
+++ b/ocs_ci/utility/csr.py
@@ -2,6 +2,7 @@ import logging
 
 from ocs_ci.ocs import constants, exceptions, ocp
 from ocs_ci.utility.retry import retry
+from ocs_ci.utility.utils import run_cmd
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


ocs_ci/utility/csr.py
```diff
@@ -60,3 +63,18 @@ def get_csr_resource():
...
+    Gets the pending CSRs
+
+    Returns:
+        list: list of pending CSRs
+
+    """
+    csr_conf = get_csr_resource()
+    return [
+        item['metadata']['name'] for item in csr_conf.data.get('items')
+        if not item.get('status')
```
petr-balogh, line 79: I was also thinking about handle this in my pr but wanted to have quick fix and I was not sure about possible statuses and so on. Are you sure there cannot be any other status reported? Like bad one? This case this condition will not work.

vavuthu: I didn't see any other status. but pending CSRs are always empty dictionary

petr-balogh:
If we are sure no other status can appear which will lead to not approved CSR I am fine :)

Issue #1643, opened 2020-03-12
Multiple clusters unable to use new long running prefixes


clacroix12:
If multiple people want to use our [new prefixes](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/cleanup/aws/defaults.py#L5) which change how cleanup is done they will not be able to deploy. Currently we stop deploying if the prefix matches any other instances up to avoid having people spin up multiple clusters on accident.

```13:50:59 - MainThread - ocs_ci.deployment.aws - ERROR - Non terminated EC2 instances with the same name prefix were found: ['i-07934b17107d2d829', 'i-030a46f0be692a58f', 'i-088fb7b5710414a69', 'i-04d9467ac4b6fe95c', 'i-0588738b29361f6e8', 'i-009cf9267564c476b']```

PR #1906, opened 2020-04-23
Cluster prefix duplication fix

clacroix12:
Fixes: #1643

### DIFF
--- a/ocs_ci/cleanup/aws/defaults.py
+++ b/ocs_ci/cleanup/aws/defaults.py
@@ -3,8 +3,8 @@ Defaults module for AWS cleanup
 """
 AWS_REGION = 'us-east-2'
 CLUSTER_PREFIXES_SPECIAL_RULES = {
-    'jnk': 36, 'DND': 'never', 'LR1': 24, 'LR2': 48, 'LR3': 72, 'LR4': 96,
-    'LR5': 120
+    'jnk': 36, 'dnd': 'never', 'lr1': 24, 'lr2': 48, 'lr3': 72, 'lr4': 96,
+    'lr5': 120
 }
 MINIMUM_CLUSTER_RUNNING_TIME = 10
 AWS_CLOUDFORMATION_TAG = 'aws:cloudformation:stack-name'
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -9,7 +9,7 @@ from subprocess import PIPE, Popen
 
 import boto3
 
-from ocs_ci.cleanup.aws import defaults
+from ocs_ci.cleanup.aws.defaults import CLUSTER_PREFIXES_SPECIAL_RULES
 from ocs_ci.deployment.ocp import OCPDeployment as BaseOCPDeployment
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants, exceptions, ocp
@@ -253,7 +253,7 @@ class AWSIPI(AWSBase):
             cluster_name = config.ENV_DATA['cluster_name']
             cluster_name_parts = cluster_name.split("-")
             prefix = cluster_name_parts[0]
-            if prefix in defaults.CLUSTER_PREFIXES_SPECIAL_RULES.keys():
+            if prefix.lower() in CLUSTER_PREFIXES_SPECIAL_RULES.keys():
                 # if the prefix is a cleanup special rule, use the next part of
                 # the cluster name as the prefix
                 prefix = cluster_name_parts[1]
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -9,6 +9,7 @@ from subprocess import PIPE, Popen
 
 import boto3
 
+from ocs_ci.cleanup.aws import defaults
 from ocs_ci.deployment.ocp import OCPDeployment as BaseOCPDeployment
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants, exceptions, ocp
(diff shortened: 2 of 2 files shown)


clacroix12:
@petr-balogh verified that this doesn't affect our deployment, however this does add a caveat that if we are using one of the special aws prefixes that the second part of the cluster name is used as the prefix for determining uniqueness. for example:

cluster_name = lr1-clacroix-test

since `lr1` is a special rule, use the next part of the cluster_name so..

prefix = `clacroix`

vasukulkarni:
> @petr-balogh verified that this doesn't affect our deployment, however this does add a caveat that if we are using one of the special aws prefixes that the second part of the cluster name is used as the prefix for determining uniqueness. for example:
> 
> cluster_name = lr1-clacroix-test
> 
> since `lr1` is a special rule, use the next part of the cluster_name so..
> 
> prefix = `clacroix`

We can check for the special rule to exist in any part of the cluster name as discussed?

clacroix12:
@vasukulkarni i think i'd like to keep it as prefixes because they are referenced as prefixes in the aws code as well as here. Also to avoid the chance of someone hitting one of our 'prefixes' in the middle of the cluster name and causing the cleanup to miss resources. Even if we did use `in` as opposed to `startswith` i think we'd still have to determine the actual "prefix" here since we don't want to check uniqueness based on any of the special rule prefixes.

vasukulkarni:
> @vasukulkarni i think i'd like to keep it as prefixes because they are referenced as prefixes in the aws code as well as here. Also to avoid the chance of someone hitting one of our 'prefixes' in the middle of the cluster name and causing the cleanup to miss resources. Even if we did use `in` as opposed to `startswith` i think we'd still have to determine the actual "prefix" here since we don't want to check uniqueness based on any of the special rule prefixes.

'jnk', 'dnd' are the only one that one could hit, and I dont recall anyone having that prefix as their username, we could just expand and make it 'jenkins', 'no-delete', I rather like the flexibility than stick to the prefix rule that is making other things harder.

clacroix12:
@vasukulkarni i think we are keeping the prefixes short because we can only use so many characters for the resource names in aws. We end up having to truncate and we want most of the name to be descriptive. Its also not about having those specific prefixes as the cluster name but it could be within someones username. I've seen instances of some peoples usernames including their middle initial as well so something like `jnkennedy` would end up getting read in as a special use case and passed over for cleanup. Point is we avoid all of that if we explicitly match the actual prefix. I don't think using prefixes to define these rules makes anything more difficult now that this change will address the issue we had with uniqueness.

vasukulkarni:
> @vasukulkarni i think we are keeping the prefixes short because we can only use so many characters for the resource names in aws. We end up having to truncate and we want most of the name to be descriptive. Its also not about having those specific prefixes as the cluster name but it could be within someones username. I've seen instances of some peoples usernames including their middle initial as well so something like `jnkennedy` would end up getting read in as a special use case and passed over for cleanup. Point is we avoid all of that if we explicitly match the actual prefix. I don't think using prefixes to define these rules makes anything more difficult now that this change will address the issue we had with uniqueness.

As I said, expanding is not really a big issue like 'jenkins' and 'no-delete',  even for the hypotical case like username 'jnkennedy', the current pr may strip the first part and use 'jnkkennedy' as prefix causing it to match the standard prefix 'jnk', we can worry about the corner case with 3 letter prefix or expand and solve it to make it more flexible.

clacroix12:
The fix in this PR splits with `"-"` so unless it was `jnk-ennedy` it wouldn't be matched to a special prefix.

vasukulkarni:
> The fix in this PR splits with `"-"` so unless it was `jnk-ennedy` it wouldn't be matched to a special prefix.

So I am now missing how does the PR once it removes the prefix is still able to cleanup?

clacroix12:
> > The fix in this PR splits with `"-"` so unless it was `jnk-ennedy` it wouldn't be matched to a special prefix.
> 
> So I am now missing how does the PR once it removes the prefix is still able to cleanup?

The prefix isn't being removed, we are simply using the next part of the cluster name to detect other clusters using the same 'prefix'.

For example:

A cluster exists in aws named `dnd-vasu-test`

I'm trying to spin up a cluster named `dnd-clacroix-test`

Currently, we would grab the prefix `dnd` and check AWS for other clusters using that prefix to determine if there already exists a cluster using the prefix and then stops the user from creating a new one. This functionality is intended to stop people from spinning up multiple clusters with the same username to conserve AWS resources. The issue is that the prefix in my example isn't the username, it's the special prefix to control how the cluster is cleaned up by our aws cleanup job.

With my changes here, if the prefix (`dnd`) is found in the list of aws cleanup prefixes, we will grab the next part of the name (after splitting on '-') so the prefix would be `clacroix`. It would then check AWS for other clusters with `clacroix` in the name and let me create the cluster.

We don't actually remove the prefix from the cluster name, we simply use the next part of the name to determine if we are creating a cluster using the same "prefix" as another cluster in aws to stop us from creating multiple clusters per user.

Hopefully that makes sense.

vasukulkarni:
> > > The fix in this PR splits with `"-"` so unless it was `jnk-ennedy` it wouldn't be matched to a special prefix.
> > 
> > 
> > So I am now missing how does the PR once it removes the prefix is still able to cleanup?
> 
> The prefix isn't being removed, we are simply using the next part of the cluster name to detect other clusters using the same 'prefix'.
> 
> For example:
> 
> A cluster exists in aws named `dnd-vasu-test`
> 
> I'm trying to spin up a cluster named `dnd-clacroix-test`
> 
> Currently, we would grab the prefix `dnd` and check AWS for other clusters using that prefix to determine if there already exists a cluster using the prefix and then stops the user from creating a new one. This functionality is intended to stop people from spinning up multiple clusters with the same username to conserve AWS resources. The issue is that the prefix in my example isn't the username, it's the special prefix to control how the cluster is cleaned up by our aws cleanup job.
> 
> With my changes here, if the prefix (`dnd`) is found in the list of aws cleanup prefixes, we will grab the next part of the name (after splitting on '-') so the prefix would be `clacroix`. It would then check AWS for other clusters with `clacroix` in the name and let me create the cluster.
> 
> We don't actually remove the prefix from the cluster name, we simply use the next part of the name to determine if we are creating a cluster using the same "prefix" as another cluster in aws to stop us from creating multiple clusters per user.
> 
> Hopefully that makes sense.

Now makes sense! 👍

PR #1911, opened 2020-04-24
Noobaa workload during upgrade

fbalak:
- Add tests `test_start_upgrade_mcg_io` and `test_start_upgrade_mcg_io` - these tests are using fio to utilize s3 interface.
- Implement configparser handling of fio configuration loading. I propose to save fio configuration files on `TEMPLATE_FIO_DIR` path.


fbalak:
There is a problem with using [http](https://github.com/axboe/fio/blob/master/engines/http.c) engine. It is currently not part of fio downstream release. I will try to find a solution.

fbalak:
The problem with http ioengine should now be resolved by https://github.com/red-hat-storage/ocs-ci/pull/2082.

fbalak:
Jobs https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/1764//testReport and https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/1735//testReport failed because upgrade failed but it seems that noobaa workload worked as expected. Problems with upgrade itself should be resolved in another PR.

tests/ecosystem/upgrade/conftest.py
```diff
@@ -157,6 +160,18 @@ def fio_project(project_factory_session):
     return project_factory_session()
 
 
+@pytest.fixture(scope='session')
```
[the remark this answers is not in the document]
fbalak: You are right. The fixture would be useful if it was used on other places for some checks but your suggested solution is currently better.

ocs_ci/utility/utils.py
```diff
@@ -1997,6 +1997,34 @@ def destroy_cluster(installer, cluster_path, log_level="DEBUG"):
...
+    Convert ConfigParser object to string in INI format.
+
+    Args:
+        config (obj): ConfigParser object
+
+    Returns:
+        str: Config in one string
+
+    """
+    string = ""
```
AviadP, line 2011: i would use meaningful name like "cfg_str" or something else that can imply about string purpose

ocs_ci/utility/utils.py
```diff
@@ -1997,6 +1997,34 @@ def destroy_cluster(installer, cluster_path, log_level="DEBUG"):
...
+    sections = [config.default_section] + config.sections()
+
+    first = True
```
AviadP, line 2014: same

tests/ecosystem/upgrade/conftest.py
```diff
@@ -516,6 +543,66 @@ def pre_upgrade_pods_running_io(
...
+        fio_project,
+        tmp_path
+    )
+
+    # deploy the Job to the cluster and start it
+    job_file.create()
+    log.info(f"Job {job_name} created")
+
+    # get job object
+    ocp_job_obj = ocp.OCP(kind=constants.JOB, namespace=fio_project.namespace)
```
AviadP, line 591: are u using this object anywhere outside the teardown function? if not, may be better to create it within function scope

tests/ecosystem/upgrade/test_noobaa.py
```diff
@@ -1,6 +1,7 @@
 import logging
 
 import pytest
+import time
```
AviadP, line 4: its very minor, but time should be together with logging (built-in lib) and not with pytest(3rd party lib)

tests/ecosystem/upgrade/test_noobaa.py
```diff
@@ -157,7 +158,7 @@ def test_noobaa_postupgrade(
         LOCAL_TEMP_PATH,
         mcg_obj_session
     )
-    assert bucket.phase == constants.STATUS_BOUND
+    assert bucket.status == constants.STATUS_BOUND
```
AviadP, line 161: plz add assert message (if assert fails)

tests/ecosystem/upgrade/test_noobaa.py
```diff
@@ -181,3 +182,39 @@ def test_buckets_after_upgrade(upgrade_buckets, mcg_obj_session):
     """
     for bucket in mcg_obj_session.read_system().get('buckets'):
         assert bucket.get('mode') == BS_OPTIMAL
+    assert bucket.status == constants.STATUS_BOUND
```
AviadP, line 185: plz add assert message (if assert fails)

tests/ecosystem/upgrade/test_noobaa.py
```diff
@@ -181,3 +182,39 @@ def test_buckets_after_upgrade(upgrade_buckets, mcg_obj_session):
...
+        job = mcg_workload_job.ocp.get(
+            resource_name=mcg_workload_job.ocp.resource_name,
+            out_yaml_format=True
+        )
+        # 'active' attribute holds information about how many pods are running
+        job_status = job['items'][0]['status']['active']
+        if job_status == 1:
+            break
+        time.sleep(3)
+    assert job_status == 1
```
AviadP, line 205: plz add assert message (if assert fails)

tests/ecosystem/upgrade/test_noobaa.py
```diff
@@ -181,3 +182,39 @@ def test_buckets_after_upgrade(upgrade_buckets, mcg_obj_session):
...
+    """
+    Confirm that there is MCG workload job running after upgrade.
+    """
+    job = mcg_workload_job.ocp.get(
+        resource_name=mcg_workload_job.ocp.resource_name,
+        out_yaml_format=True
+    )
+    # 'active' attribute holds information about how many pods are running
+    job_status = job['items'][0]['status']['active']
+    assert job_status == 1
```
AviadP, line 220: plz add assert message (if assert fails)

tests/ecosystem/upgrade/conftest.py
fbalak: The fixture returns job object which is used in 2 tests: test_start_upgrade_mcg_io and test_upgrade_mcg_io and is kept active during upgrade.

ocs_ci/utility/utils.py
fbalak: What name do you propose? I's part of the algorithm and it should contain information if the iteration is first or not.

ocs_ci/templates/workloads/fio/config_s3.fio
```diff
@@ -0,0 +1,34 @@
+# Example test for the HTTP engine's S3 support against Amazon AWS.
```
mbukatov, line 1: I see one problem with placing this fio config file here, which is that this directory `ocs_ci/templates/workloads/fio` contains yaml file which is processed by some ocs-ci code and pushed to command line arguments of fio run, and so existing yaml files are neither complete nor actual fio config file, while `config_s3.fio` is complete and actual fio config file.

Maybe you can address this by adding a README file, where you explain this difference. Nothing that `config_s3.fio` is ini config file of fio, not jinja template nor yaml dump of option for ocs-ci code, while the other files are.

ocs_ci/utility/utils.py
```diff
@@ -1997,6 +1997,34 @@ def destroy_cluster(installer, cluster_path, log_level="DEBUG"):
...
+        if config.items(section):
+            if first:
+                first = False
+            else:
+                string += "\n\n"
+            string += f"[{section}]\n"
+            string += "\n".join(
+                [f"{item[0]}={item[1]}" for item in config.items(section)]
+            )
+    return string
```
mbukatov, line 2025: Are you implementing this because configparser object does allow to write string representation to file object only? If yes, it's would be better to just use `io.StringIO` object in `configparser.write(fileobject)` call instead of re-implementing string dump of configparser ini file.

Something like (this is just a draft, it's not a finalized code :):

```
strio = io.StringIO()
cp.write(strio)
string = strio.getvalue()
```

tests/ecosystem/upgrade/test_noobaa.py
mbukatov: > plz add assert message (if assert fails)

My personal opinion is that assert message is not necessary in such clear cases, and don't understand the urge to add them everywhere. That said, adding a logging above the assert could help.

tests/ecosystem/upgrade/test_noobaa.py
mbukatov: >  its very minor, but time should be together with logging (built-in lib) and not with pytest(3rd party lib)

+1

ocs_ci/utility/utils.py
AviadP: could be `first_iteration` then

tests/ecosystem/upgrade/conftest.py
AviadP: my question is, why not create `ocp_job_obj` within fixture scope

fbalak: This is just a wrapper to access `get` function implemented in our framework. It does not create anything so I don't think that it needs to be accessible from outside as it's own fixture. Also, it is not possible to use function scope fixture in session scope fixture.

ocs_ci/templates/workloads/fio/config_s3.fio
fbalak: I will add a README because I don't know any other better place to keep it in.
I can create some other directory under `ocs_ci/templates` like `ocs_ci/templates/fioconfig/` but it seems worse because then there will be two directories with fio configs under `ocs_ci/templates`.
I can create a new directory in `conf` directory but `conf` directory is currently used for cluster and OCSCI config files which are more related to CI.
But if you think it should be in a new directory then please suggest the path that the files should be placed in.

fbalak:
There seems to be a problem with getting `persistentVolumeClaim` of pod volumes in spec. This PR didn't alter the functionality of this or any of the failing test cases. The problem should be fixed in additional PR.

tests/ecosystem/upgrade/test_noobaa.py
```diff
@@ -159,3 +169,48 @@ def test_buckets_after_upgrade(upgrade_buckets, mcg_obj_session):
...
+@pre_upgrade
+def test_start_upgrade_mcg_io(mcg_workload_job):
+    """
+    Confirm that there is MCG workload job running before upgrade.
+    """
+    job_status = None
+    job_name = mcg_workload_job.ocp.resource_name
+    # wait a few seconds for fio job to start
+    logger.info(f"Checking number of active pods for job {job_name}")
+    for i in range(0, 5):
```
AviadP, line 183: plz consider use TimeoutSampler instead this loop

AviadP:
LGTM

tests/ecosystem/upgrade/test_noobaa.py
ebenahar: +1

fbalak: Thanks for the tip. I considered it but I would rather keep it as it is. There would have to be created special function inside test function and that would be passed to TimeoutSampler which would be in my opinion less readable. Also there would have to be a new call to the new function in `assert job_status == 1` in assert message which would lead to reporting wrong number of pods. It would look something like:
```
     assert TimeoutSampler(                                                                                                                                                                                                                   
         timeout=20, sleep=3, func=new_function, state=1                                                                                                                                                      
     ), (                                                                                                                                                                                                                                     
         f"Number of running pods for job {job_name} is not 1, "                                                                                                                                                                              
         f"it is {new_function()}"                                                                                                                                                                                                          
     )    
```

tests/ecosystem/upgrade/conftest.py
```diff
@@ -201,6 +204,30 @@ def fio_conf_block():
         """)
 
 
+@pytest.fixture(scope='session')
+def fio_conf_mcg(mcg_obj_session, bucket_factory_session):
+    """
+    Basic fio configuration for upgrade utilization for AWS S3 bucket.
```
MeridianExplorer, line 210: From what I understand, the FIO configuration is for a NooBaa bucket, and not for an AWS S3 bucket - no?
I see you create the workload buckets by using `bucket_factory_session()` and also use the `mcg_obj_session.s3_endpoint`, which is the S3 endpoint of NooBaa - not AWS

tests/ecosystem/upgrade/conftest.py
```diff
@@ -566,3 +653,28 @@ def upgrade_buckets(
         )
 
     return buckets
+
+
+@pytest.fixture(scope='session')
```
MeridianExplorer, line 658: Just for easier readability - I think it might be better to move it above `mcg_workload_job`. 
Was surprised to see it defined _after_ an invocation :)

tests/ecosystem/upgrade/test_noobaa.py
```diff
@@ -57,7 +58,10 @@ def test_fill_bucket(
         )
 
     mcg_obj_session.check_if_mirroring_is_done(bucket.name)
-    assert bucket.status == constants.STATUS_BOUND
```
MeridianExplorer, line 60: Regarding the three assertion messages (lines 60, 70, 80) - I feel they're rather redundant, considering `bucket.status` already logs the retrieved status, and the value of `constants.STATUS_BOUND` will be logged as part of the assertion error (while it is also accessible by checking the constants file, of course)
image

tests/ecosystem/upgrade/test_noobaa.py
```diff
@@ -159,3 +169,48 @@ def test_buckets_after_upgrade(upgrade_buckets, mcg_obj_session):
...
+    job_name = mcg_workload_job.ocp.resource_name
+    # wait a few seconds for fio job to start
+    logger.info(f"Checking number of active pods for job {job_name}")
+    for i in range(0, 5):
+        job = mcg_workload_job.ocp.get(
+            resource_name=job_name,
+            out_yaml_format=True
+        )
+        # 'active' attribute holds information about how many pods are running
+        job_status = job['items'][0]['status']['active']
```
MeridianExplorer, line 189: Have you considered replacing the direct access with several `.get()` calls?
Ditto for line 212

tests/ecosystem/upgrade/test_noobaa.py
MeridianExplorer: I tend to agree with Filip's point - but considering that lines 189 + 193-196, and 212-16 are outright code duplication - I'd reconsider moving the logic into a helper function that the two tests can later use.
Also, it's possible to give up on the one-liner `assert TimeoutSampler`, and instead save the return value on the side
i.e. - 
```
def check_job_state(job, desired_state):
    job_name = job.ocp.resource_name
    job_obj = job.ocp.get(resource_name=job_name, out_yaml_format=True)
    def _retrieve_job_state():
        return job.get('items').get(0).get('status').get('active')
    try:
        for state in TimeoutSampler(timeout=20, sleep=3, func=_retrieve_job_state):
            if state == desired_state:
                return True
    except TimeoutExpiredError:
        log.error(
            f"Number of running pods for job {job_name} is not 1, "
        ) 
        return False
```
Pretty much turns both of the tests into one-liners

tests/ecosystem/upgrade/conftest.py
fbalak: Changed

tests/ecosystem/upgrade/conftest.py
fbalak: Ok

tests/ecosystem/upgrade/test_noobaa.py
fbalak: Good point, changed

tests/ecosystem/upgrade/test_noobaa.py
fbalak: Ok, I changed it based on Ben's comment. The code duplication is a good point.

Issue #1912, opened 2020-04-24
AWS Cleanup vpc_tags NoneType issue


clacroix12:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-aws-cleanup/1109//console

```
Traceback (most recent call last):
  File "/home/jenkins/workspace/qe-aws-cleanup/ocs-ci/.venv/bin/aws-cleanup", line 11, in <module>
    load_entry_point('ocs-ci', 'console_scripts', 'aws-cleanup')()
  File "/home/jenkins/workspace/qe-aws-cleanup/ocs-ci/ocs_ci/cleanup/aws/cleanup.py", line 309, in aws_cleanup
    prefixes_hours_to_spare=prefixes_hours_to_spare,
  File "/home/jenkins/workspace/qe-aws-cleanup/ocs-ci/ocs_ci/cleanup/aws/cleanup.py", line 164, in get_clusters_to_delete
    tag['Value'] for tag in vpc_tags if tag['Key'] == defaults.AWS_CLOUDFORMATION_TAG
TypeError: 'NoneType' object is not iterable
```

PR #1913, opened 2020-04-24
Handle case where vpc_obj tags returns a NoneType

clacroix12:
Fixes: #1912 


clacroix12:
@shylesh @ebenahar there were other cases where that empty list caused issues down the line so I just put the whole thing behind an `if` to ensure we have tags to work with.

Verified here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-aws-cleanup/1112/

Issue #1925, opened 2020-04-26
move health check to the end of install verification


vavuthu:
move health check to the end of install verification.

Sometimes Health check of cluster is OK during install verification and went to bad in teardown and in retry of teardown again it went back to good state.

So its good to have health check at the end of install verification instead of middle.

```
------------------------------ live log teardown -------------------------------
20:26:28 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc wait --for condition=ready pod -l app=rook-ceph-tools -n openshift-storage --timeout=120s
20:26:28 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage get pod -l 'app=rook-ceph-tools' -o jsonpath='{.items[0].metadata.name}'
20:26:28 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage exec rook-ceph-tools-74494c7b4b-44dkv ceph health
20:26:29 - MainThread - tests.conftest - INFO - Ceph health check failed at teardown
20:26:29 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc wait --for condition=ready pod -l app=rook-ceph-tools -n openshift-storage --timeout=120s
20:26:29 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage get pod -l 'app=rook-ceph-tools' -o jsonpath='{.items[0].metadata.name}'
20:26:29 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage exec rook-ceph-tools-74494c7b4b-44dkv ceph health
20:26:30 - MainThread - ocs_ci.utility.retry - WARNING - Ceph cluster health is not OK. Health: HEALTH_WARN Reduced data availability: 1 pg peering
, Retrying in 30 seconds...
20:27:00 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc wait --for condition=ready pod -l app=rook-ceph-tools -n openshift-storage --timeout=120s
20:27:00 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage get pod -l 'app=rook-ceph-tools' -o jsonpath='{.items[0].metadata.name}'
20:27:00 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage exec rook-ceph-tools-74494c7b4b-44dkv ceph health
[...]

PR #1926, opened 2020-04-26
Fixing health_checker exception

vavuthu:

Fixes: #1925


RazTamir:
The 'raise' is the reason for this ceph health check.
Let me check what I did wrong and update

vavuthu:
> The 'raise' is the reason for this ceph health check.
> Let me check what I did wrong and update

had a quick chat and I agree that here "raise" is for the test cases to find out ocs is in good shape even after test case execution.

Closing this PR and I will raise to include health check in another PR

PR #1933, opened 2020-04-27
Scale - Add ceph_health_check in ceph pod respin test

ramkiperiy:
Updated Scale script to validate ceph status after ceph pod respin
Removed bugzilla markers in pytest params
Updated script to perform memory leak test.

### DIFF
--- a/tests/e2e/scale/test_pv_scale_and_respin_ceph_pods.py
+++ b/tests/e2e/scale/test_pv_scale_and_respin_ceph_pods.py
@@ -183,5 +183,5 @@ class TestPVSTOcsCreatePVCsAndRespinCephPods(BasePvcCreateRespinCephPods):
 
         # Added sleep for test case run time and for capturing memory leak after scale
         time.sleep(test_run_time)
-        utils.ceph_health_check(delay=180)
+        utils.ceph_health_check()
         helpers.memory_leak_analysis(median_dict)
--- a/tests/e2e/scale/test_pv_scale_and_respin_ceph_pods.py
+++ b/tests/e2e/scale/test_pv_scale_and_respin_ceph_pods.py
@@ -12,6 +12,7 @@ import time
 from tests import helpers, disruption_helpers
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.resources import pod
+from ocs_ci.utility import utils
 from ocs_ci.framework.testlib import scale, E2ETest, ignore_leftovers
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


tests/e2e/scale/test_pv_scale_and_respin_ceph_pods.py
```diff
@@ -183,5 +183,5 @@ def test_pv_scale_out_create_pvcs_and_respin_ceph_pods(
 
         # Added sleep for test case run time and for capturing memory leak after scale
         time.sleep(test_run_time)
-        # TODO: Skipping memory leak analysis because of bz 1750328
-        # helpers.memory_leak_analysis(median_dict)
+        assert utils.ceph_health_check(delay=180), "Failed, Ceph health NOT_OK after ceph pod respin"
```
RazTamir, line 186: no need assert as this function returns nothing but raise exception in case of failure

ramkiperiy: Updated the TC

tests/e2e/scale/test_pv_scale_and_respin_ceph_pods.py
```diff
@@ -183,5 +183,5 @@ def test_pv_scale_out_create_pvcs_and_respin_ceph_pods(
 
         # Added sleep for test case run time and for capturing memory leak after scale
         time.sleep(test_run_time)
-        # TODO: Skipping memory leak analysis because of bz 1750328
-        # helpers.memory_leak_analysis(median_dict)
+        utils.ceph_health_check(delay=180)
```
ebenahar, line 186: delay is "Delay in seconds between retries". Is there a reason to have a polling rate of 180 seconds?

ramkiperiy: @ebenahar Setup will have 120 pods with IO and in parallel there will be OSD pod respin, added delay to make sure rebalance is complete and cluster is back to OK state.

ebenahar: @ramkiperiy yes ok but with delay=180, it means that we wait for 3 minutes between checks and this is a very low poll rate. Also, with delay=180 and tries=20 (default value), we are waiting for 180*20 seconds, which is 1 hour

ramkiperiy: ack, This TC doesn't require that much time, will use the default 30sec.

PR #1936, opened 2020-04-27
helpers for uninstall OCS automation

annagitel:
Signed-off-by: annagitel <asandler@redhat.com>
-PVC
-constants
-storage cluster
-POD

### DIFF
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -114,6 +114,7 @@ CLUSTER_SERVICE_VERSION = 'csv'
 
 # Other
 SECRET = "Secret"
+TEST = "test"
 NAMESPACE = 'Namespace'
 IGNORE_SC_GP2 = "gp2"
 IGNORE_SC_FLEX = "rook-ceph-block"
@@ -766,3 +767,8 @@ NOOBAA_SERVICE_ACCOUNT = "system:serviceaccount:openshift-storage:noobaa"
 
 # Miscellaneous
 NOOBAA_OPERATOR_POD_CLI_PATH = "/usr/local/bin/noobaa-operator"
+
+# Storage classes provisioners
+OCS_PROVISIONERS = ['openshift-storage.rbd.csi.ceph.com',
+                    'openshift-storage.cephfs.csi.ceph.com',
+                    'openshift-storage.noobaa.io/obc']
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -769,6 +769,8 @@ NOOBAA_SERVICE_ACCOUNT = "system:serviceaccount:openshift-storage:noobaa"
 NOOBAA_OPERATOR_POD_CLI_PATH = "/usr/local/bin/noobaa-operator"
 
 # Storage classes provisioners
-OCS_PROVISIONERS = ['openshift-storage.rbd.csi.ceph.com',
-                    'openshift-storage.cephfs.csi.ceph.com',
-                    'openshift-storage.noobaa.io/obc']
+OCS_PROVISIONERS = [
+    'openshift-storage.rbd.csi.ceph.com',
+    'openshift-storage.cephfs.csi.ceph.com',
+    'openshift-storage.noobaa.io/obc'
+]
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -18,7 +18,7 @@ from tests import helpers
 from ocs_ci.ocs import workload
 from ocs_ci.ocs import constants, defaults, node
 from ocs_ci.framework import config
-from ocs_ci.ocs.exceptions import CommandFailed, NonUpgradedImagesFoundError
+from ocs_ci.ocs.exceptions import CommandFailed, NonUpgradedImagesFoundError, UnavailableResourceException
 from ocs_ci.ocs.utils import setup_ceph_toolbox
 from ocs_ci.ocs.resources.ocs import OCS
 from ocs_ci.utility import templating
@@ -1032,14 +1032,16 @@ def get_pvc_name(pod_obj):
         pod_obj (str): The pod object
 
     Returns:
-        (str): The pvc name of a given pod_obj,
-                None if there is no pvc attached
+        str: The pvc name of a given pod_obj,
+
+    Raises:
+        UnavailableResourceException: If no pvc attached
 
     """
     pvc = pod_obj.get().get('spec').get('volumes')[0].get('persistentVolumeClaim')
-    if pvc:
-        return pvc.get('claimName')
-    return pvc
+    if not pvc:
+        raise UnavailableResourceException
+    return pvc.get('claimName')
 
 
 def get_used_space_on_mount_point(pod_obj):
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -1032,11 +1032,14 @@ def get_pvc_name(pod_obj):
         pod_obj (str): The pod object
 
     Returns:
-        pvc_name (str): The pvc_name on a given pod_obj
+        (str): The pvc name of a given pod_obj,
+                None if there is no pvc attached
+
     """
-    return pod_obj.get().get(
-        'spec'
-    ).get('volumes')[0].get('persistentVolumeClaim').get('claimName')
+    pvc = pod_obj.get().get('spec').get('volumes')[0].get('persistentVolumeClaim')
+    if pvc:
+        return pvc.get('claimName')
+    return pvc
 
 
 def get_used_space_on_mount_point(pod_obj):
--- a/ocs_ci/ocs/resources/pvc.py
+++ b/ocs_ci/ocs/resources/pvc.py
@@ -216,9 +216,7 @@ def get_all_pvcs(namespace=None, selector=None):
     Returns:
          dict: Dict of all pvc in namespaces
     """
-    all_ns = False
-    if namespace == 'all-namespaces':
-        all_ns = True
+    all_ns = True if namespace == 'all-namespaces' else False
     if not namespace:
         namespace = config.ENV_DATA['cluster_namespace']
     ocp_pvc_obj = OCP(
--- a/ocs_ci/ocs/resources/pvc.py
+++ b/ocs_ci/ocs/resources/pvc.py
@@ -5,6 +5,7 @@ import logging
 from concurrent.futures import ThreadPoolExecutor
 
 from ocs_ci.ocs import constants
+from ocs_ci.ocs.exceptions import UnavailableResourceException
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs.resources.ocs import OCS
 from ocs_ci.framework import config
@@ -169,16 +170,17 @@ class PVC(OCS):
         Returns:
             list: A list of pod objects attached to the PVC
 
-        Note:
-             # Importing from pod inside, because of unsolvable import loop
-             from ocs_ci.ocs.resources.pod import get_all_pods, get_pvc_name
-
         """
+        # Importing from pod inside, because of unsolvable import loop
         from ocs_ci.ocs.resources.pod import get_all_pods, get_pvc_name
         attached_pods = []
         all_pods = get_all_pods()
         for pod_obj in all_pods:
-            if get_pvc_name(pod_obj) == self.name:
+            try:
+                pvc = get_pvc_name(pod_obj)
+            except UnavailableResourceException:
+                continue
+            if pvc == self.name:
                 attached_pods.append(pod_obj)
         return attached_pods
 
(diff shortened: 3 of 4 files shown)


ocs_ci/ocs/resources/pvc.py
```diff
@@ -162,6 +162,22 @@ def resize_pvc(self, new_size, verify=False):
             )
         return True
 
+    def get_attached_pods(self):
+        """
+        Get the pods attached to the PVC represented by this object instance
+
+        Returns:
+            list: A list of pod objects attached to the PVC
```
ebenahar, line 170: add an empty line below

ocs_ci/ocs/resources/pvc.py
```diff
@@ -162,6 +162,22 @@ def resize_pvc(self, new_size, verify=False):
...
+        """
+        from ocs_ci.ocs.resources.pod import get_all_pods, get_pvc_name
```
ebenahar, line 172: please add a comment above for the reason behind importing here

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -431,3 +431,27 @@ def get_deviceset_count():
...
+        kind=constants.STORAGECLASS,
+        namespace=defaults.ROOK_CLUSTER_NAMESPACE
+    )
+    result = sc_obj.get()
+    sample = result['items']
+
+    storageclass = [
+        item for item in sample if (
+            (item.get('metadata').get('name') not in constants.IGNORE_SC_GP2)
+            and (item.get('metadata').get('name') not in constants.IGNORE_SC_FLEX)
```
OdedViner, line 454: you can change:
    storageclass = [
        item for item in sample if (
            item.get('metadata').get('name') not in (constants.IGNORE_SC_GP2,constants.IGNORE_SC_FLEX)
        )
    ]

ocs_ci/ocs/resources/pod.py
```diff
@@ -1032,8 +1032,14 @@ def get_pvc_name(pod_obj):
         pod_obj (str): The pod object
 
     Returns:
-        pvc_name (str): The pvc_name on a given pod_obj
+        pvc_name (str): The pvc_name on a given pod_obj,
```
RazTamir, line 1035: there's no such variable pvc_name - please remove it.
Also, under 'Returns':
str: The pvc_name on a given pod_obj ....

ocs_ci/ocs/resources/pod.py
```diff
@@ -1032,8 +1032,14 @@ def get_pvc_name(pod_obj):
...
+                        Node if there is no pvc attached
+
     """
+    if pod_obj.get().get(
```
RazTamir, line 1039: please refactor the entire function:
```
pvc = pod_obj.get().get('spec').get('volumes')[0].get('persistentVolumeClaim')
if pvc:
    return pvc.get('claimName')
return pvc
```

ocs_ci/ocs/resources/pod.py
```diff
@@ -1032,11 +1032,14 @@ def get_pvc_name(pod_obj):
         pod_obj (str): The pod object
 
     Returns:
-        pvc_name (str): The pvc_name on a given pod_obj
+        (str): The pvc name of a given pod_obj,
+                Node if there is no pvc attached
```
jilju, line 1036:
```suggestion
                None if there is no pvc attached
```

ocs_ci/ocs/resources/pvc.py
```diff
@@ -162,6 +162,25 @@ def resize_pvc(self, new_size, verify=False):
...
 
+    def get_attached_pods(self):
+        """
+        Get the pods attached to the PVC represented by this object instance
+
+        Returns:
+            list: A list of pod objects attached to the PVC
+
+        Note:
+             importing from POD inside, because of unsolvable import loop
```
jilju, line 173: Add as a comment just before the import statement.
```suggestion
             # Importing from pod inside, because of unsolvable import loop
             from ocs_ci.ocs.resources.pod import get_all_pods, get_pvc_name
```

ocs_ci/ocs/constants.py
```diff
@@ -114,6 +114,7 @@
 
 # Other
 SECRET = "Secret"
+TEST = "test"
```
MeridianExplorer, line 117: Is this used somewhere?

ocs_ci/ocs/constants.py
```diff
@@ -766,3 +767,8 @@
 
 # Miscellaneous
 NOOBAA_OPERATOR_POD_CLI_PATH = "/usr/local/bin/noobaa-operator"
+
+# Storage classes provisioners
+OCS_PROVISIONERS = ['openshift-storage.rbd.csi.ceph.com',
```
MeridianExplorer, line 772: We usually don't put parentheses in the same lines as their contents - 
```
OCS_PROVISIONERS = [
                                           'openshift-storage.rbd.csi.ceph.com',
                                           'openshift-storage.cephfs.csi.ceph.com',
                                           'openshift-storage.noobaa.io/obc'
]
```

ocs_ci/ocs/resources/pod.py
```diff
@@ -1032,11 +1032,14 @@ def get_pvc_name(pod_obj):
...
-        pvc_name (str): The pvc_name on a given pod_obj
+        (str): The pvc name of a given pod_obj,
+                None if there is no pvc attached
+
     """
-    return pod_obj.get().get(
-        'spec'
-    ).get('volumes')[0].get('persistentVolumeClaim').get('claimName')
+    pvc = pod_obj.get().get('spec').get('volumes')[0].get('persistentVolumeClaim')
+    if pvc:
```
MeridianExplorer, line 1040: Is this check needed?
Since you already use `.get()`, I think it's safe to just call it directly - 
`return pod_obj.get().get('spec').get('volumes')[0].get('persistentVolumeClaim').get('claimName')`
If a PVC attached, it'll return it, otherwise - None. Unless I'm missing something?

ocs_ci/ocs/resources/pvc.py
```diff
@@ -162,6 +162,26 @@ def resize_pvc(self, new_size, verify=False):
...
+
+        Note:
+             # Importing from pod inside, because of unsolvable import loop
+             from ocs_ci.ocs.resources.pod import get_all_pods, get_pvc_name
+
+        """
+        from ocs_ci.ocs.resources.pod import get_all_pods, get_pvc_name
+        attached_pods = []
+        all_pods = get_all_pods()
+        for pod_obj in all_pods:
```
MeridianExplorer, line 180: Can be reduced to `for pod_obj in get_all_pods()`, I don't think a variable is needed

MeridianExplorer: Can also be highly compacted with some list comprehension - 
`return [pod_obj for pod_obj in get_all_pods() if get_pvc_name(pod_obj) == self.name]`

ocs_ci/ocs/constants.py
annagitel: yep

ocs_ci/ocs/resources/pod.py
annagitel: was changed because it threw errors

annagitel: get('persistentVolumeClaim') is a None, when you do  get('claimName') on it, it fails

ocs_ci/ocs/resources/pvc.py
annagitel: made it long for readability :)

ocs_ci/ocs/resources/pvc.py
```diff
@@ -190,18 +210,21 @@ def get_all_pvcs(namespace=None, selector=None):
...
     Args:
-        namespace (str): Name of namespace
+        namespace (str): Name of namespace  ('all-namespaces' to get all namespaces)
         selector (str): The label selector to look for
 
     Returns:
          dict: Dict of all pvc in namespaces
     """
+    all_ns = False
+    if namespace == 'all-namespaces':
```
MeridianExplorer, line 220: Can be compacted - 
`all_ns = True if lower(namespace) == 'all-namespaces' else False`

annagitel: again, readability

MeridianExplorer, line 220: Can be compacted
```suggestion
    all_ns = True if namespace.lower() == 'all-namespaces' else False
```

ocs_ci/ocs/resources/pod.py
```diff
@@ -1032,11 +1032,14 @@ def get_pvc_name(pod_obj):
...
+        return pvc.get('claimName')
+    return pvc
```
ebenahar, line 1042: what's the point of this condition? 

It would be better to return the pvc and assert outside, where you call this method

ocs_ci/ocs/resources/pod.py
annagitel: @ebenahar

ocs_ci/ocs/resources/pod.py
annagitel: tagged you in a comment above

ebenahar: thanks. You can do this instead:

```python
pvc = pod_obj.get().get('spec').get('volumes')[0].get('persistentVolumeClaim')
assert pvc, "Failed to find the PVC for pod {pod_obj.get('metadata').get('name')}"
return return pvc.get('claimName')
```

ebenahar: and if you do that, the function will not return None, so please remove the line about None from the returned value in the docstring

annagitel: but its not a failure, there will be places where wee want to get None and see that the pod doesent have a pvc attached to it.

annagitel: much easier to check 'None' then trying to catch an assertion error

ebenahar: the place that calls this method will not have to check for the assert as the assert is an exception which will be thrown in case the pvc object doesn't exist. 
We can go with either approach, although I don't see the use case in which we'll want to return None

annagitel: when deleting pvcs and pods, we want to check that we don't leave any unused pvc behind

annagitel: or a pvc that will make a pod that needs to be deleted come up again and again

ocs_ci/ocs/resources/pvc.py
```diff
@@ -162,6 +163,27 @@ def resize_pvc(self, new_size, verify=False):
...
+
+        """
+        # Importing from pod inside, because of unsolvable import loop
+        from ocs_ci.ocs.resources.pod import get_all_pods, get_pvc_name
+        attached_pods = []
+        all_pods = get_all_pods()
+        for pod_obj in all_pods:
+            try:
+                pvc = get_pvc_name(pod_obj)
+            except UnavailableResourceException:
```
RazTamir, line 181: Saw the conversation with @ebenahar.
I'm not sure why raising exception is the way we prefer. 
This is not the approach in all resources and not aligning (iiuc) with this function usecase - to get the pvc attached to a pod.
In most cases we are just returning whatever exists in the object or trying to align with oc commands. In this case, we don't have oc command so we have the freedom to use whatever we want. 

Also, when considering an API call (REST), you'll most likely get an empty object instead of resource not found as you are not searching for a resource here (not specific one)

ebenahar: @RazTamir the reason for raising an exception is that if pod_obj does not have spec.volumes[0].persistentVolumeClaim, the failure will be of ```None type doesn't have...``` so we'd like to catch such a thing and to raise a proper exception

ebenahar: if we'll return None in case pvc is None it will affect the usage of this function (ocs_ci.ocs.resources.pod.get_pvc_name). If we won't raise an exception here, this test will falsely pass

RazTamir: This comment was supposed to be in line 1043 where we explicitly raised the exception.

Look at this https://www.restapitutorial.com/httpstatuscodes.html.
The implementation here is similar to 404 but that's not right as we are not asking to GET for specific resource. 

Your call, not blocking this PR

RazTamir: Approving the PR

ebenahar: @RazTamir what is the better option then?

RazTamir: Returning the pvc name (str) or None if doesn't exist

ebenahar: let's continue the discussion here https://github.com/red-hat-storage/ocs-ci/issues/1946.
Merging this one

Issue #811, opened 2019-09-24
memory leak function/analysis needs enhancement to support multiple osd's on same worker node


ramkiperiy:
For now memory leak function runs on all workers but in future we will have worker nodes for OCS and app pods, so function needs enhancement to collect data only from OCS nodes or OSD service running nodes.

If a worker node has 3 OSD's for now function collects all the ceph-osd daemon stats but needs enhancement in analysis, may be based on process id, but should be intelligent to identify restart of process (because of daemon process id change).

ramkiperiy:
Change /tmp reference in the memory leak function code, use tempdir Module instead.
This change needs to be done for all the respective functions with respect to memory leak.

PR #1939, opened 2020-04-27
Copy worker node top output to log path

ramkiperiy:
Copy worker node top output to log path in memory leak test.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,6 +9,7 @@ from itertools import chain
 from math import floor
 from random import randrange
 from time import sleep
+from shutil import copyfile
 
 import pytest
 import yaml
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -1347,8 +1348,13 @@ def finalizer():
             )
         if thread:
             thread.join()
+        log_path = ocsci_log_path()
         for worker in helpers.get_worker_nodes():
             if os.path.exists(f"/tmp/{worker}-top-output.txt"):
+                copyfile(
+                    f"/tmp/{worker}-top-output.txt",
```
vasukulkarni, line 1355: can you change '/tmp' reference to tempfile.mkdtemp(prefix='perf_')

ramkiperiy: @vasukulkarni added code just to move the file from local /tmp to logs_dir. For /tmp/ reference change it required to be changed in lot of places, like fixture in conftest.py and 2 memory_leak related functions in helpers.

Anyway I have plans to fix issue https://github.com/red-hat-storage/ocs-ci/issues/811 during this fix will handle this change.

PR #1964, opened 2020-04-30
vdbench | Adding teardown ceph_health_check to the test

Avilir:

### DIFF
--- a/tests/e2e/performance/test_vdbench_workload.py
+++ b/tests/e2e/performance/test_vdbench_workload.py
@@ -7,7 +7,7 @@ import time
 import os
 
 from ocs_ci.ocs.ocp import OCP
-from ocs_ci.utility.utils import TimeoutSampler, ocsci_log_path
+from ocs_ci.utility.utils import TimeoutSampler, ocsci_log_path, ceph_health_check
 from ocs_ci.ocs.resources.ocs import OCS
 from ocs_ci.utility import templating
 from ocs_ci.ocs.utils import get_pod_name_by_pattern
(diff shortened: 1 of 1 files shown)


tests/e2e/performance/test_vdbench_workload.py
```diff
@@ -48,6 +48,9 @@ def label_nodes(request, with_ocs):
 
     def teardown():
 
+        if not ceph_health_check():
```
RazTamir, line 51: no need if and raise.
Just call ceph_health_check().  it will raise in case a failure

Avilir: Fixed it.

tests/e2e/performance/test_vdbench_workload.py
```diff
@@ -48,6 +48,8 @@ def label_nodes(request, with_ocs):
 
     def teardown():
 
+        ceph_health_check()
```
ebenahar, line 51: IMO, it would make more sense to check health as part of the test and not part of the teardown. 
The reason is that if the cluster is not healthy after the test step is done, then the test itself should fail and not the teardown

Avilir: When the test is run with more then 50% capacity (70% is defined), in the end it will be in WARN state, and this is OK, so after teardown i want to verify it came back to OK state.

ebenahar: why is it ok that the cluster is in WARN state when the test is done? It means that we are not validating that the test has succeeded as part of the test phase

Avilir: When the cluster is filled up with data of more then 50% in the pool capacity, it start to WARN the user (defined by target_pool_ration - default is 0.49) and this is OK and it should warn and it will stay in warn state until the user will add capacity or free some space.
In the teardown we are deleting the date, so it need to return to OK mode.

PR #1970, opened 2020-04-30
OCP upgreade: fixed indentation

AviadP:
Follow up PR#1921, fixed 1 intenatation


ebenahar:
as this is a docstring change, no need to actual verification. @AviadP please mark as verified

PR #1997, opened 2020-05-05
Increase the time normal jobs run in AWS cluster to 40 hours(5 days)

vasukulkarni:

### DIFF
--- a/ocs_ci/cleanup/aws/cleanup.py
+++ b/ocs_ci/cleanup/aws/cleanup.py
@@ -252,7 +252,7 @@ def aws_cleanup():
         help="""
             Maximum running time of the cluster (in hours).
             Clusters older than this will be deleted.
-            The minimum is 10 hours
+            The maximum is 40 hours
             """
     )
     parser.add_argument(
--- a/ocs_ci/cleanup/aws/cleanup.py
+++ b/ocs_ci/cleanup/aws/cleanup.py
@@ -252,7 +252,7 @@ def aws_cleanup():
         help="""
             Maximum running time of the cluster (in hours).
             Clusters older than this will be deleted.
-            The maximum is 40 hours
+            The maximum is 120 hours
             """
     )
     parser.add_argument(
--- a/ocs_ci/cleanup/aws/defaults.py
+++ b/ocs_ci/cleanup/aws/defaults.py
@@ -6,6 +6,6 @@ CLUSTER_PREFIXES_SPECIAL_RULES = {
     'jnk': 36, 'dnd': 'never', 'lr1': 24, 'lr2': 48, 'lr3': 72, 'lr4': 96,
     'lr5': 120
 }
-MINIMUM_CLUSTER_RUNNING_TIME = 10
+MINIMUM_CLUSTER_RUNNING_TIME = 40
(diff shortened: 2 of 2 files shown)


ocs_ci/cleanup/aws/cleanup.py
```diff
@@ -252,7 +252,7 @@ def aws_cleanup():
         help="""
             Maximum running time of the cluster (in hours).
             Clusters older than this will be deleted.
-            The minimum is 10 hours
+            The maximum is 40 hours
```
clacroix12, line 255: This is not a maximum, it is a minimum value. This stops us from accidentally cleaning up every cluster older than say '4' hours.

clacroix12:
If we do decide to increase the amount of time we allow clusters to exist in AWS by default we will need to also change the job definition in [ocs4-jenkins](https://gitlab.cee.redhat.com/ocs/ocs4-jenkins).

The special prefix rules can probably get entirely changed as well. If our default is 40 hours (or 120 if you meant 5 days) then it doesn't make sense to have "long running" prefixes that are less than that.

ocs_ci/cleanup/aws/defaults.py
```diff
@@ -6,6 +6,6 @@
     'jnk': 36, 'dnd': 'never', 'lr1': 24, 'lr2': 48, 'lr3': 72, 'lr4': 96,
     'lr5': 120
 }
-MINIMUM_CLUSTER_RUNNING_TIME = 10
+MINIMUM_CLUSTER_RUNNING_TIME = 120
```
RazTamir, line 9: Why?

vasukulkarni: 10 hours is way to less to do meaningful testing

RazTamir: 12 hours sound like a nice to have fix.
If you are working more than that - why? (-:
if not and this is just for ease of use - this is the intent to clean the clusters

vasukulkarni: It takes around ~30 mins to setup, I wont be awake in night to test cluster :) but having the cluster running for long time is needed for many tests rather than setup new everytime, the age of the cluster is also important overtime.

ebenahar: @vasukulkarni this should not be changed. It used for determine what is the minimum running time in hours any cluster is allowed to have. With 120 it means that aws-cleanup will not work as it passed 12 hours as the allowed running time for any cluster that doesn't have the exceptions in its name (like jnk or lr)

vasukulkarni: I think i am missing how the code works, but I am closing this anyway now and going to use the prefix for now

ocs_ci/cleanup/aws/cleanup.py
vasukulkarni: The way I am interpreting the semantic is "> 10 hours" is the maximum time the cluster can live in AWS?

vasukulkarni: The minimum can be anything less than 10 hours?

vasukulkarni:
Closing based on discussion in Chat, topic for another time!

PR #2018, opened 2020-05-08
Production AWS configs for LSO deployments

clacroix12:
Jira: [OCSQE-340](https://projects.engineering.redhat.com/browse/OCSQE-340)


clacroix12:
Locally verified IPI and UPI w/ RHEL worker deployments were successful with these configurations.

PR #2025, opened 2020-05-10
Skip TestRemoveMonFromCluster

ebenahar:
This test case should be disabled, as it was marked as inactive in Polarion (OCS-355) and also, this scenario should not be supported.
Thanks @sidhant-agrawal for raising this 

### DIFF
--- a/tests/manage/z_cluster/test_remove_mon_from_cluster.py
+++ b/tests/manage/z_cluster/test_remove_mon_from_cluster.py
@@ -55,6 +55,7 @@ def run_io_on_pool(pool_obj):
 
 @tier4
 @tier4c
+@pytest.mark.skip("Test case is disabled, as per requirement not to support this scenario")
 @ignore_leftovers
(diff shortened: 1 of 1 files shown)


tests/manage/z_cluster/test_remove_mon_from_cluster.py
```diff
@@ -55,6 +55,7 @@ def run_io_on_pool(pool_obj):
 
 @tier4
 @tier4c
+@pytest.mark.skip("Test case is disabled, as per requirement not to support this scenario")
```
RazTamir, line 58: maybe better to really disable it so it will not affect the reporting as skipped?
We can do this by removing all tier decorators

ebenahar: you're right. Will do that

ebenahar:
Although the PR validation job failed, all executed tests passed and the disabled test was skipped

PR #2028, opened 2020-05-10
Improve auth config loading mechanism

MeridianExplorer:
- Take out the existing code from `get_ocs_olm_operator_tags` and put it in a dedicated function in order to allow other functions to use it
- Add a constant leading to the authentication config dogs

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -1522,8 +1522,6 @@ def load_auth_config():
 
     Raises:
         FileNotFoundError: if the auth config is not found
-        KeyError: if the auth config isn't setup properly
-        requests.RequestException: if the response return code is not ok
 
     Returns:
         dict: A dictionary reprensenting the YAML file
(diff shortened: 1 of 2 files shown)


ocs_ci/utility/utils.py
```diff
@@ -1516,44 +1516,51 @@ def get_next_version_available_for_upgrade(current_tag):
...
-def get_ocs_olm_operator_tags(limit=100):
+def load_auth_config():
     """
-    Query the OCS OLM Operator repo and retrieve a list of tags.
-
-    Args:
-        limit: the number of tags to limit the request to
+    Load the authentication config YAML from /data/auth.yaml
 
     Raises:
```
petr-balogh, line 1523: This meant to be part of get_ocs_olm_operator_tags and not this load_auth_config  function I guess

ocs_ci/utility/utils.py
```diff
@@ -1516,44 +1516,51 @@ def get_next_version_available_for_upgrade(current_tag):
...
         raise
+
+
+def get_ocs_olm_operator_tags(limit=100):
+    """
+    Query the OCS OLM Operator repo and retrieve a list of tags.
+
+    Args:
+        limit: the number of tags to limit the request to
+
```
petr-balogh, line 1551: missing raises section which forgot to be moved here as I pointed above

ocs_ci/utility/utils.py
MeridianExplorer: Yep, you're right - `FileNotFoundError` _should_ be here, the other two - shouldn't :)

PR #2034, opened 2020-05-11
Add OCS 4.5 config

petr-balogh:
Infra ticket:
https://projects.engineering.redhat.com/browse/OCSQE-339


petr-balogh:
I see that deployment failed cause of BZ I've just opened:
https://bugzilla.redhat.com/show_bug.cgi?id=1834301

I think we are safe to merge this one as it looks like it does what is suppose and started deploy OCS 4.5 but issue is in the build now. So marking as verified.

PR #2037, opened 2020-05-11
 Functionality to determine the cluster throughput limit and run IO at a given percentage of it

ebenahar:
Reach the target throughput load percentage, and keep it running throughout the test suite execution. 
The number of pods needed for the desired target percentage is determined by creating pods one by one, while examining if the cluster throughput is increased by more than 10%. When it doesn't increased by more than 10% anymore after the new pod started running IO, it means that the cluster throughput limit is reached. Then, the function deletes the pods that are not needed as they are the difference between the limit (100%) and the target percentage (the default target percentage is 30%). This leaves the number of pods needed running IO for cluster throughput to be around the desired percentage.

To be used by https://github.com/red-hat-storage/ocs-ci/blob/5cee08d83ea59d0acae5255c04ef9d51995f61fd/tests/conftest.py#L1042

Also fixes https://github.com/red-hat-storage/ocs-ci/issues/2081

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -105,7 +105,7 @@ to the pytest.
    the cluster throughput limit.
 * `--io-load` - IOs throughput target percentage. The value should be
    between 0 to 100. The default is 30 (30%)
-* `--enable-bg-io-logs` - If passed, background IO log messages, will be printed
+* `--enable-bg-io-logs` - If passed, background IO log messages will be printed
 
 ## Examples
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -188,7 +188,7 @@ run-ci tests/
     --email=<emailid1>,<emailid2>,<emailid3>
  ```
 
-#### Running tests on deployed environment and sending reports
+#### Running tests with background IO load
 
 If you would like to run tests with IO load of 50% in the tests background,
 while background IO log messages are printed, append these arguments to the
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -99,6 +99,13 @@ to the pytest.
   bugzilla_password = yourPassword
   ```
 * `--collect-logs` - to collect OCS logs for failed test cases.
+* `--io-in-bg` - If passed, IO will be running in the test background.
+   The amount of IO load will be determined by the `--io-load` argument, if
+   provided. In case `--io-load` is not provided, IO load is set to 30% of
+   the cluster throughput limit.
+* `--io-load` - IOs throughput target percentage. The value should be
+   between 0 to 100. The default is 30 (30%)
+* `--enable-bg-io-logs` - If passed, background IO log messages, will be printed
 
 ## Examples
 
@@ -180,6 +187,12 @@ run-ci tests/
     --html=report.html --self-contained-html \
     --email=<emailid1>,<emailid2>,<emailid3>
  ```
+
+#### Running tests on deployed environment and sending reports
+
+If you would like to run tests with IO load of 50% in the tests background,
+while background IO log messages are printed, append these arguments to the
+`run-ci` command: `--io-in-bg --io-load 50 --enable-bg-io-logs`
 
 #### Destroy of cluster
 
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -92,11 +92,22 @@ def pytest_addoption(parser):
         help="Collect OCS logs when test case failed",
     )
     parser.addoption(
-        '--io_in_bg',
+        '--io-in-bg',
         dest='io_in_bg',
         action="store_true",
         default=False,
         help="Run IO in the background",
+    )
+    parser.addoption(
+        '--io-load',
+        dest='io_load',
+        help="IOs throughput target percentage. Value should be between 0 to 100",
+    )
+    parser.addoption(
+        '--enable-bg-io-logs',
+        dest='enable_bg_io_logs',
+        action="store_true",
+        help="Enable background IO logs, if --io-in-bg is passed",
     )
     parser.addoption(
         '--ocs-version',
@@ -299,7 +310,15 @@ def process_cluster_cli_params(config):
     ocsci_config.DEPLOYMENT['live_deployment'] = live_deployment or (
         ocsci_config.DEPLOYMENT.get('live_deployment', False)
     )
-    ocsci_config.RUN['cli_params']['io_in_bg'] = get_cli_param(config, "io_in_bg", default=False)
+    io_in_bg = get_cli_param(config, 'io_in_bg')
+    if io_in_bg:
+        ocsci_config.RUN['io_in_bg'] = True
+        io_load = get_cli_param(config, 'io_load')
+        if io_load:
+            ocsci_config.RUN['io_load'] = io_load
+        io_bg_logs = get_cli_param(config, 'enable_bg_io_logs')
+        if io_bg_logs:
+            ocsci_config.RUN['io_bg_logs'] = True
     upgrade_ocs_version = get_cli_param(config, "upgrade_ocs_version")
     if upgrade_ocs_version:
         ocsci_config.UPGRADE['upgrade_ocs_version'] = upgrade_ocs_version
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -19,7 +19,7 @@ from ocs_ci.ocs.exceptions import UnexpectedBehaviour
 from ocs_ci.ocs.resources import ocs, storage_cluster
 import ocs_ci.ocs.constants as constant
 from ocs_ci.utility.retry import retry
-from ocs_ci.utility.utils import TimeoutSampler, run_cmd
+from ocs_ci.utility.utils import TimeoutSampler, run_cmd, convert_device_size
 from ocs_ci.ocs.utils import get_pod_name_by_pattern
 from ocs_ci.framework import config
 from ocs_ci.ocs import ocp, constants, exceptions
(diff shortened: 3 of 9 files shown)


ocs_ci/ocs/cluster.py
```diff
@@ -590,6 +589,22 @@ def get_throughput_percentage(self):
...
+
+        Args:
+            num_of_samples (int): The number of samples to take
+
+        Returns:
+            float: The average cluster throughput
+
+        """
+        throughput_vals = [
+            self.get_cluster_throughput() for _ in range(num_of_samples)
```
keesturam, line 604: I see the following issues with the way throughput and throughput percentage is calculated. But they are out of scope for this PR. I'm still mentioning it here so that it is recorded. 
1) I believe constants.THROUGHPUT_LIMIT_OSD of 250 is not the right value even for m5.4xlarge. 
2) The throughput assumes that the environment is AWS with instance type m5.4xlarge, so this function cannot be used for other environments.

ocs_ci/ocs/cluster.py
```diff
@@ -924,3 +939,219 @@ def validate_pg_balancer():
...
+        # 'runtime' is set with a large value of seconds to
+        # make sure that the pods are running
+        io_file_size = f"{pvc_size - 1}G"
+        pod_obj.run_io(
+            storage_type='block', size=io_file_size,
+            runtime=io_run_time, rate='200M', bs='128K', rw_ratio=25
+        )
+
+        logger.info(f"Waiting for IO to kick in on pod {pod_obj.name}")
+        time.sleep(20)
```
keesturam, line 1071: pre-fill of `fio` may take more than 20 seconds for a large pvc size. During this time, it uses most of the bandwidth and is not restricted by the 'rate' option. 
Maybe, we need a function to check if pre-fill part of fio is completed.

ocs_ci/ocs/cluster.py
ebenahar: Hi @keesturam actually, here I'm not calculating the TP based on a constant, but rather based on the dynamic value collected by `ceph status`. The limit is being determined as explained in the below `reach_cluster_load_percentage_in_throughput()`. 
`get_cluster_throughput()` is not the method that uses the 250MB constant. The one that does it is `get_throughput_percentage()`

ocs_ci/ocs/cluster.py
```diff
@@ -924,3 +939,219 @@ def validate_pg_balancer():
...
+    io_run_time = 100**3
+
+    if target_percentage < 0.1:
+        logger.info(
+            f"Target percentage is {target_percentage} which is less "
+            f"than 10%. Hence, IO will not start"
+        )
+        return None
+
+    pvc_size = int(get_osd_pods_memory_sum() * 0.75)
```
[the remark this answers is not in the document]
ebenahar: Thanks @keesturam, I also got an imformation about the need to overcome the cache effect from @Avilir.
Here, the OSD memory is the sum of all OSD pods. So in a 3 worker cluster, with 8GB RMA for each OSD pod, the disk size that will be created is `24GB * 0.75`.
The multiplication by `0.75` is to lower the chances of choking smaller clusters in terms of storage/CPU/mem and in any case, the next created PVC will consume the rest of the 0.25

ocs_ci/ocs/cluster.py
```diff
@@ -924,3 +939,219 @@ def validate_pg_balancer():
...
+        raw_block_pv=True
+    )
+    pod_objs.append(pod_obj)
+
+    io_file_size = f"{pvc_size-1}G"
+    # 'runtime' is set with a large value of seconds
+    # to make sure that the pods are running
+    pod_obj.run_io(
+        storage_type='block', size=io_file_size,
+        runtime=io_run_time, rate='200M', bs='128K', rw_ratio=25
```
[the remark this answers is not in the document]
ebenahar: used 200M for rate based on a conversation with @Avilir

ocs_ci/ocs/cluster.py
ebenahar: yes, we can create such a function, maybe in a follow up PR. I think it is less urgent as the PVC are being create with a size of ~6GB, see line 1044:
`pvc_size = int(get_osd_pods_memory_sum() * 0.25)`

ocs_ci/ocs/cluster.py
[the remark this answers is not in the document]
ebenahar: and also the same for `bs=128K`

ebenahar:
Screenshot from 2020-05-16 20-23-51
[io_bg.log](https://github.com/red-hat-storage/ocs-ci/files/4638829/io_bg.log)

ocs_ci/ocs/cluster_load.py
```diff
@@ -0,0 +1,278 @@
...
+
+class ClusterLoad:
+    """
+    A class for cluster load functionalities
+
+    """
+
+    def __init__(self, log=True):
+        """
+        Initializer for ClusterLimit
```
ebenahar, line 23: `ClusterLimit` will be fixed to `ClusterLoad`

ebenahar:
Tier1 execution over AWS IPI with 1GB OSD volume size, while the cluster is loaded at 50% of its throughput limit:

Screenshot from 2020-05-17 20-22-47

ebenahar:
And here is tier1 execution over vSphere (results are in 1 comment above):

Screenshot from 2020-05-18 08-37-59

ocs_ci/ocs/cluster.py
```diff
@@ -924,3 +938,37 @@ def validate_pg_balancer():
             return False
     else:
         logging.info("pg_balancer is not active")
+
+
+def get_percent_used_capacity():
+    """
+    ***This function was taken from PR 1996***
```
AviadP, line 945: is this line intended ?

ocs_ci/ocs/cluster_load.py
```diff
@@ -0,0 +1,278 @@
...
+import time
+
+import ocs_ci.ocs.constants as constant
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.cluster import (
+    CephCluster, get_osd_pods_memory_sum, get_percent_used_capacity
+)
+
+
+class ClusterLoad:
```
AviadP, line 15:
```suggestion
class ClusterLoad(object):
```

ocs_ci/ocs/cluster.py
ebenahar: I can remove it

ocs_ci/ocs/cluster_load.py
ebenahar: I don't think it is necessary. See https://stackoverflow.com/questions/4015417/python-class-inherits-object/9448136

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -92,12 +92,23 @@ def pytest_addoption(parser):
...
     parser.addoption(
-        '--io_in_bg',
+        '--io-in-bg',
         dest='io_in_bg',
         action="store_true",
         default=False,
         help="Run IO in the background",
     )
+    parser.addoption(
+        '--io-load',
```
vasukulkarni, line 102: please update readme docs on when to use and when not to use and maybe one or two usecase.

ocs_ci/ocs/cluster_load.py
```diff
@@ -0,0 +1,278 @@
...
+        """
+        self.logger = logging.getLogger(__name__)
+        self.logger.propagate = propagate_logs
+        self.cl_obj = CephCluster()
+
+    def reach_cluster_load_percentage_in_throughput(
+        self, pvc_factory, pod_factory, target_percentage=0.3, cluster_limit=None
+    ):
+        """
+        This function determines how many pods, that are running FIO, needed in order to
```
vasukulkarni, line 37: can you reword the sentence? is it adjusting the fio or the number of pods or both?

ocs_ci/ocs/cluster_load.py
```diff
@@ -0,0 +1,278 @@
...
+                storage_type='block', size=io_file_size,
+                runtime=io_run_time, rate='200M', bs='128K', rw_ratio=25
+            )
+
+            self.logger.info(
+                f"While IO kicks-in on the previously created pod ({pod_obj.name}), "
+                f"creating a new pod and PVC for the next iteration"
+            )
+            pvc_size = int(get_osd_pods_memory_sum() * 0.25)
+            pvc_obj = pvc_factory(
```
vasukulkarni, line 155: Not blocking for thie one but It would be nice if we can restrict to < 10 pvc/pods and adjust bs to be mixed config with 265k,1m,4m + rate options to achieve desired throughput, probably we can achieve sooner.

ocs_ci/framework/pytest_customization/ocscilib.py
ebenahar: good idea! will add

ocs_ci/ocs/cluster_load.py
ebenahar: Ack

ocs_ci/ocs/cluster_load.py
ebenahar: Thanks for the suggestions!
- Num of pods is restricted (line 221 - `if len(pod_objs) > 8:)` in case the throughput is lower than 20Mb/s after 8 pods and PVCs have been created and IO was running on them. 
Also, timeout of 15 minutes is set below in case throughput didn't reach more than 20Mb/s.
And, a timeout of 30 in case throughput limit was not determined.

Regarding BS and rate, I set these after experimenting with different values. I can try 256K as BS

docs/usage.md
```diff
@@ -181,6 +188,12 @@ run-ci tests/
     --email=<emailid1>,<emailid2>,<emailid3>
  ```
 
+#### Running tests on deployed environment and sending reports
+
+If you would like to run tests with IO load of 50% in the tests background,
+while background IO log messages are printed, append these arguments to the
+`run-ci` command: `--io-in-bg --io-load 50 --enable-bg-io-logs`
```
vasukulkarni, line 195: 👍

vasukulkarni:
LGTM, nice work!

PR #2042, opened 2020-05-12
Added function to fetch all device paths irrespective of platform

anubhav-here:
Since there was a requirement to fetch device paths from the worker nodes irrespective of the platform (Aws/Bare Meral etc.)
Added a helper function that does the same.
Signed-off-by: Anubhav Deep <adeep@redhat.com>

### DIFF
--- a/ocs_ci/utility/localstorage.py
+++ b/ocs_ci/utility/localstorage.py
@@ -1,0 +1,34 @@
+"""
+This module contains local-storage related methods
+"""
+import logging
+import os
+import shutil
+import yaml
+
+from ocs_ci.utility.utils import clone_repo, run_cmd
+from ocs_ci.ocs import constants
+
+logger = logging.getLogger(__name__)
+
+
+def fetch_all_device_paths():
+    """
+    Return all device paths inside worker nodes
+
+    Returns:
+        list : List containing all device paths
+
+    """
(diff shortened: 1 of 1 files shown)


tests/helpers.py
```diff
@@ -2233,3 +2234,27 @@ def validate_pods_are_running_and_not_restarted(
     logger.error(f"Pod is in {pod_state} state and restart count of pod {restart_count}")
     logger.info(f"{pod_obj}")
     return False
+
+
+def fetch_device_paths():
+    """
+        Return all device paths inside worker nodes
+
+        Args:
```
prsurve, line 2243: No need of this

tests/helpers.py
```diff
@@ -2233,3 +2234,27 @@ def validate_pods_are_running_and_not_restarted(
...
+
+        Returns:
+            list : List containing all device paths
```
prsurve, line 2246: indentation Missing

tests/helpers.py
anubhav-here: Fixed

tests/helpers.py
anubhav-here: Fixed

anubhav-here:
@clacroix12 Added in a new module, as requested!

PR #2053, opened 2020-05-13
Pvc scale testing

wusui:
Make helper functions for pvc threading, pvc deletion, and pvc
reportiing.  Add routine where actions can be clumped together
once the pvcs have been brought up.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -2,6 +2,7 @@
 Helper functions file for OCS QE
 """
 import logging
+import csv
 import re
 import datetime
 import statistics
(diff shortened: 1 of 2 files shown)


tests/helpers.py
```diff
@@ -2233,3 +2234,58 @@ def validate_pods_are_running_and_not_restarted(
...
+def delete_objs(pvc_objs):
+    """
+    Delete pvc objects
+
+    Args:
+        pvc_objs (list): pvc objects to delete
+    """
+    for obj in pvc_objs:
+        obj.delete()
+        obj.ocp.wait_for_delete(obj.name)
```
mykaul, line 2248: You should not wait for deletion to complete if you are running at scale. At the end, you should ensure all objects were deleted.

wusui: Thanks.  I will do this in PR 2236.

PR #2080, opened 2020-05-14
Use official AWSCLI Docker image for the AWSCLI pod

MeridianExplorer:
Until this PR, we used an unofficial, 3rd party AWSCLI container image, since an official one didn't exist.
Since AWS released an official image on April, we're changing the image to the official one.

Also changed `craft_s3_command` to allow anonymous usage


MeridianExplorer:
Ran MCG tiers 1-3 locally, all passed successfully except flaky issues that pass after a rerun.

<details>
  <summary>Tier 1 - failure was flaky, passed after a rerun</summary>

image
</details>

<details>
  <summary>Tier 2 - failure fixed in #2084</summary>

image
</details>

<details>
  <summary>Tier 3 - flaky failure, passed after a rerun</summary>

image

</details>

Tier 4 wasn't run because the cluster was deployed on VMWare, but I believe tiers 1-3 are a safe enough indication.

tests/manage/mcg/test_object_integrity.py
```diff
@@ -9,6 +9,7 @@
 from ocs_ci.ocs import constants
 from tests.manage.mcg import helpers
 from tests.manage.mcg.helpers import retrieve_anon_s3_resource
```
AviadP, line 11: already imported 'tests.manage.mcg import helpers' so next line is not needed

tests/manage/mcg/test_object_integrity.py
```diff
@@ -9,6 +9,7 @@
...
+from tests.helpers import craft_s3_command
```
AviadP, line 12: already imported ```tests.manage.mcg import helpers```, so next line is not needed

ocs_ci/templates/app-pods/awscli.yaml
```diff
@@ -6,7 +6,9 @@ metadata:
 spec:
   containers:
     - name: awscli
-      # TODO: Replace the image with an official one once Amazon  publishes theirs
-      image: mikesir87/aws-cli:1.16.220
+      image: amazon/aws-cli:2.0.13
```
fbalak, line 9: Shouldn't we use `latest` image to test that we are in sync with the most recent aws-cli?
I am not sure about this - maybe this is ok if it's stable.

MeridianExplorer: That's a great point. I am honestly not entirely sure. 
Using a specific tag promises us stability, and assures us that everything will work as expected. It might limit us from things that can be added in future revisions, but at the moment, we don't seem to need those. That's why I chose to go with the most futureproof way - and if we'll need to use a feature added in a specific CLI version, we can always update the YAML and verify that everything is working well.

What do you think?

MeridianExplorer: Also, the description of the image clearly states that [no backwards compatibility is promised when using the `latest` tag](https://hub.docker.com/r/amazon/aws-cli#notes-on-interfaces-and-backwards-compatibility-of-docker-image)

fbalak: I will leave this up to you or if anyone else wants to decide this.
> Also, the description of the image states that [no backwards compatibility is promised when using the `latest` tag](https://hub.docker.com/r/amazon/aws-cli#notes-on-interfaces-and-backwards-compatibility-of-docker-image).

If we would use `latest` tag then we wouldn't miss changes in aws-cli related to our use cases. We would see that tests that were passing are not passing because of some change in aws-cli.

MeridianExplorer: You're right; I'll have some more discussions regarding this. Thanks for bringing it up!

ebenahar: @MeridianExplorer should we wait with merging until we reach a final decision?

MeridianExplorer: @ebenahar 
I think it'd be best to merge the PR ASAP, and change it down the road in case we choose to.

vasukulkarni: @MeridianExplorer please file an issue to track filip's comment on using latest tag for the image.

PR #2100, opened 2020-05-17
Remove the ceph_health_check from the vdbench test

Avilir:

### DIFF
--- a/tests/e2e/performance/test_vdbench_workload.py
+++ b/tests/e2e/performance/test_vdbench_workload.py
@@ -7,7 +7,7 @@ import time
 import os
 
 from ocs_ci.ocs.ocp import OCP
-from ocs_ci.utility.utils import TimeoutSampler, ocsci_log_path, ceph_health_check
+from ocs_ci.utility.utils import TimeoutSampler, ocsci_log_path
 from ocs_ci.ocs.resources.ocs import OCS
 from ocs_ci.utility import templating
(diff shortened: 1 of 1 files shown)


tests/e2e/performance/test_vdbench_workload.py
```diff
@@ -48,8 +48,6 @@ def label_nodes(request, with_ocs):
 
     def teardown():
```
keesturam, line 49: @Avilir simply removing the health check may lead to the subsequent tests to fail or skip until the cluster is healthy. Maybe we should wait until the cluster becomes healthy with a timer. WDYT?

Avilir: In case of CephFS, i am not sure how long it will take to return to HEALTH_OK status, so i do not think "sleep" can be a solution.
I think we need to have a "clean-up" that make sure to delete all back-end objects as well.

ramkiperiy: @Avilir don't add sleep but you can add this function which does wait for cluster to become OK state till some iterations(can change the iteration count accordingly)
https://github.com/red-hat-storage/ocs-ci/blob/master/tests/e2e/scale/test_ceph_pod_respin_in_scaled_cluster.py#L72

keesturam:
closing this PR as we are not using vdbench test anymore.

PR #2110, opened 2020-05-19
Changed restart_nodes() implementation to directly call reboot

### LINKED ISSUE #2103: restart_ec2_instances() should directly call reboot_instances instead of doing stop and start
We should directly call reboot_instances instead of stop and start.
AWS i3 based OCS deployment, cannot tolerate Node shutdown but can hold node restarts.
Due to the current way of AWS reboot, we are kind of doing shutdown(stop/start) due to which some of the test cases which involves node reboot are failing on AWS i3.

    def restart_ec2_instances(self, instances, wait=False, force=True):
        """
        Stop and start ec2 instances

        Args:
            instances (dict): A dictionary of instance IDs and names to restart
            wait (bool): True in case wait for status is needed,
                False otherwise
            force (bool): True for force instance stop, False otherwise

        """
        logger.info(f"Restarting instances {list(instances.values())}")
        self.stop_ec2_instances(instances=instances, wait=wait, force=force)
        self.start_ec2_instances(instances=instances, wait=wait)

ocs-ci results for OCS4-4-Downstream-OCP4-4-AWS-IPI-3AZ-RHCOS-LSO-3M-3W-workloads (BUILD ID: v4.4.0-420.ci RUN ID: 1589409630

PrasadDesala:
Fixes #2103
Created two different functions for nodes restart.
1) restart_nodes - which directly calls restart
2) restart_nodes_by_stop_and_start - which does restart by stopping and starting the instances/Vms

### DIFF
--- a/ocs_ci/ocs/platform_nodes.py
+++ b/ocs_ci/ocs/platform_nodes.py
@@ -20,7 +20,7 @@ from ocs_ci.utility.utils import (
     get_ocp_version, TimeoutSampler,
     download_file, delete_file, AZInfo
 )
-
+from ocs_ci.ocs.node import wait_for_nodes_status
 
 logger = logging.getLogger(__name__)
 
@@ -73,6 +73,11 @@ class NodesBase(object):
     def restart_nodes(self, nodes, force=True):
         raise NotImplementedError(
             "Restart nodes functionality is not implemented"
+        )
+
+    def restart_nodes_by_stop_and_start(self, nodes, force=True):
+        raise NotImplementedError(
+            "Restart nodes by stop and start functionality is not implemented"
         )
 
     def detach_volume(self, volume, node=None, delete_from_backend=True):
@@ -244,6 +249,21 @@ class VMWareNodes(NodesBase):
         )
         self.vsphere.restart_vms(vms, force=force)
 
+    def restart_nodes_by_stop_and_start(self, nodes, force=True):
+        """
+        Restart vSphere VMs
+
+        Args:
+            nodes (list): The OCS objects of the nodes
+            force (bool): True for force VM stop, False otherwise
+
+        """
+        vms = self.get_vms(nodes)
+        assert vms, (
+            f"Failed to get VM objects for nodes {[n.name for n in nodes]}"
+        )
+        self.vsphere.restart_vms_by_stop_and_start(vms, force=force)
+
     def detach_volume(self, volume, node=None, delete_from_backend=True):
         """
         Detach disk from a VM and delete from datastore if specified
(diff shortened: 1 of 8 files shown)


PrasadDesala:
I have tested the AWS restart_node changes and it is working fine.
Failed test [1] in run 1589409630 ran successfully on AWS i3 based setup.

[1] tests/e2e/monitoring/test_monitoring_on_negative_scenarios.py::TestMonitoringBackedByOCS::test_monitoring_when_one_of_the_prometheus_node_down 

Have made similar changes for VMware as well. I have not yet tested the code yet on VMware. Hence, marking this PR with WIP.

ocs_ci/ocs/platform_nodes.py
```diff
@@ -394,24 +414,64 @@ def start_nodes(self, nodes, wait=True):
...
+            Once the OCP node detects that the node is not reachable then the
+            node reaches status NotReady.
+            When the reboot operation is completed and the instance is up and 
+            reachable the OCP node reaches status Ready.
+            """
+            nodes_names = [n.name for n in nodes]
+            logger.info(
+                f"Waiting for nodes: {nodes_names} to reach not ready state"
+            )
+            wait_for_nodes_status(
```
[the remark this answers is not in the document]
PrasadDesala: 1) When a node is restart from AWS/Vmware the instance/VM stays at running state throughout the reboot operation and I think due to this there is no return wait conditions (wait_until/WaitForTasks). 
2) When we initiated reboot it will take few mins for the node to reach NotReady state and once the instance is reachable then the node reaches Ready state.

When we call restart_nodes() and immediately check for wait_for_nodes_status=Ready it will succeed but actually the OCP node has not detect the failure of node yet, after few mins the OCP node will detect the failure and reach NotReady state, which is misleading.

So, I am first waiting for the node to reach NotReady and then Ready.
I tried it even manually and I see always the node reaching NotReady state first and then Ready state.

Please suggest if there is any other better way/solution to handle this.

ocs_ci/ocs/platform_nodes.py
```diff
@@ -230,13 +236,15 @@ def start_nodes(self, nodes, wait=True):
         )
         self.vsphere.start_vms(vms)
 
-    def restart_nodes(self, nodes, force=True):
+    def restart_nodes(self, nodes, force=False, wait=True):
```
RazTamir, line 239: this changes functionality.
Are we sure this will not impact any existing operation?

PrasadDesala: *. In the current implementation, since we are setting the force arg to True by default it stops the VM by direct poweroff.
*. With direct VMware reboot, reboot can be done in two ways,
1)Hard reboot(VM Reset)
 2) Soft reboot(Guest Reboot)
I thought soft reboot is most common scenario compared to VM reset so set False by default, this anyways can be changed according to the test case.

I ran existing testcases which are using this function and didn't see any failure and I am also planning to run multiple tiers/workloads to make sure nothing is breaking because of this change.

Please let me know if I am missing anything here and you think setting to false breaks existing operations here.

PrasadDesala:
> PR validation [Job UNSTABLE](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/1770//testReport) (some or all tests failed).

These are test results for Tier 4a. 
There were  6 failures (+6) , 11 skipped (+11). 
The failed testcases doesn't touch the code patch in this PR.
Will run the individual tescases which involves restart_nodes.

PrasadDesala:
Ran the tests cases which are using restart_nodes() on both Vmware[1] and AWS[2] platform.
There were few failures in the runs and are not due to the changes made in this PR

[1] https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/8202/
[2] https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/8057/

So, marking with label Verified

Issue #1967, opened 2020-04-30
use 2 test functions instead of parametrization for test_capacity_workload_alerts 


mbukatov:
Test function `test_capacity_workload_alerts()` implements 2 test cases (OCS-899 for RBD, OCS-1934 for CephFS) via parametrization, while and using both RBD and CephFS workload fixtures in both parametrization cases.

To make it more clear which workload is necessary in which case from reporting point of view, we need to use 2 test functions instead of parametrization here.

PR #2111, opened 2020-05-19
Split capacity tests

fbalak:
This is done for reporting purpose. It is easier to analyse the results because execution of workload fixtures is done separately.

fixes https://github.com/red-hat-storage/ocs-ci/issues/1967

### DIFF
--- a/tests/manage/monitoring/prometheus/test_capacity.py
+++ b/tests/manage/monitoring/prometheus/test_capacity.py
@@ -20,7 +20,7 @@ log = logging.getLogger(__name__)
 )
 def test_rbd_capacity_workload_alerts(workload_storageutilization_95p_rbd):
     """
-    Test that there are appropriate alerts when ceph cluster is utilizedi
+    Test that there are appropriate alerts when ceph cluster is utilized
     via RBD interface.
     """
     api = prometheus.PrometheusAPI()
--- a/tests/manage/monitoring/prometheus/test_capacity.py
+++ b/tests/manage/monitoring/prometheus/test_capacity.py
@@ -12,55 +12,88 @@ from ocs_ci.ocs.ocp import OCP
 log = logging.getLogger(__name__)
 
 
-@pytest.mark.parametrize(
-    argnames='interface',
-    argvalues=[
-        pytest.param(
-            'rbd',
-            marks=[
-                pytest.mark.polarion_id("OCS-899"),
-                pytest.mark.bugzilla('1809248'),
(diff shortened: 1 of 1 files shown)


tests/manage/monitoring/prometheus/test_capacity.py
```diff
@@ -12,55 +12,88 @@
...
-            ]
+@pytest.mark.polarion_id("OCS-899")
+@pytest.mark.bugzilla('1809248')
+@tier2
+@gather_metrics_on_fail(
+    'ceph_cluster_total_used_bytes', 'cluster:memory_usage_bytes:sum'
+)
+def test_rbd_capacity_workload_alerts(workload_storageutilization_95p_rbd):
+    """
+    Test that there are appropriate alerts when ceph cluster is utilizedi
```
ebenahar, line 23: fix `utilizedi`

fbalak: done

fbalak:
The tests were skipped because BZ 1809248 was not fixed yet. I propose to merge this change now and decide if bugzilla marker for capacity tests should be removed in another PR as it might take a while before the problem is fixed and it might not affect the test in every case.

mbukatov:
> The tests were skipped because BZ 1809248 was not fixed yet. I propose to merge this change now and decide if bugzilla marker for capacity tests should be removed in another PR as it might take a while before the problem is fixed and it might not affect the test in every case.

Agreed, the bugzilla marker should not be removed at this point. But I would suggest to drop the marker (in a different branch based on this PR) just for the purpose of running CI validation job and post the results there.

tests/manage/monitoring/prometheus/test_capacity.py
```diff
@@ -12,55 +12,88 @@
...
+
+    if config.ENV_DATA.get('ocs_version') == '4.2':
+        nearfull_message = (
+            'Storage cluster is nearing full. Expansion is required.'
+        )
+        criticallfull_mesage = (
+            'Storage cluster is critically full and needs immediate expansion'
+        )
+    else:
+        # since OCS 4.3
```
vasukulkarni, line 40: can be 4.4 or in future 4.5, so better to use 'elif == 4.3'

fbalak: I would change it only if the message wasn't valid for 4.3, 4.4 and 4.5. This way it works as expected.

Issue #2115, opened 2020-05-19
Use latest-stable-version tags


petr-balogh:
Based on conversation in infra room

PR #2116, opened 2020-05-19
Use latest-stable

petr-balogh:
Fixes: #2115

### DIFF
--- a/conf/ocs_version/ocs-4.2.yaml
+++ b/conf/ocs_version/ocs-4.2.yaml
@@ -1,7 +1,7 @@
 ---
 DEPLOYMENT:
   default_ocs_registry_image: "quay.io/rhceph-dev/ocs-olm-operator:4.2-rc"
-  default_latest_tag: '4.2-rc'
+  default_latest_tag: 'latest-stable-4.2'
   ocs_csv_channel: 'stable-4.2'
   default_ocp_version: '4.2'
 ENV_DATA:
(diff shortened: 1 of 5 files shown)


conf/ocs_version/ocs-4.2.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 DEPLOYMENT:
   default_ocs_registry_image: "quay.io/rhceph-dev/ocs-olm-operator:4.2-rc"
-  default_latest_tag: '4.2-rc'
+  default_latest_tag: 'latest-stable-4.2'
```
vasukulkarni, line 4: should be just latest

ocs_ci/ocs/constants.py
```diff
@@ -602,8 +602,11 @@
 OCS_CSV_PREFIX = 'ocs-operator'
 LOCAL_STORAGE_CSV_PREFIX = 'local-storage-operator'
 LATEST_TAGS = (
-    'latest', 'latest-stable', '4.2-rc', 'latest-4.2', 'latest-stable-4.2',
+    'latest', 'latest-stable',
+    '4.2-rc', 'latest-4.2', 'latest-stable-4.2',
```
vasukulkarni, line 606: do we need x.y-rc ? @andrewschoen

petr-balogh: I would keep it here for now if we will use them at some point as we were used to. It doesn't matter here, it's just used to identify such build and then ocs-ci will replace such tag with some specific.

vasukulkarni: @petr-balogh if it has usage then its better to keep it but if latest-4.x will also point to rc then its better to get rid of this to avoid confusion and users setting this value and expecting rc build?

vasukulkarni:
Have to hold until we have latest-stable tags for older ocs version

vasukulkarni:
@andrewschoen is working on tagging 4.x rc builds to latest-stable in the format of x.y

vasukulkarni:
@andrewschoen @zmc do we have the stable tags to use now?

Issue #2049, opened 2020-05-12
Need a mark to skip test on AWS platform


vasukulkarni:
Usecase: https://github.com/red-hat-storage/ocs-ci/pull/1984

keesturam:
@vasukulkarni I have included a function `get_platform` and I skip the test if the platform is `AWS` in https://github.com/red-hat-storage/ocs-ci/pull/1895/files

Please see if #1895 will fix this issue.

PR #2119, opened 2020-05-19
Add mark to run tests only in red hat internal lab

vasukulkarni:
Fixes: https://github.com/red-hat-storage/ocs-ci/issues/2049

### DIFF
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -98,7 +98,7 @@ aws_platform_required = pytest.mark.skipif(
     reason="Tests are not running on AWS deployed cluster"
 )
 
-rh_internal_lab_only = pytest.mark.skipif(
+rh_internal_lab_required = pytest.mark.skipif(
     (config.ENV_DATA['platform'].lower() == 'aws'
         or config.ENV_DATA['platform'].lower() == 'azure'),
     reason="Tests will not run in AWS or Azure Cloud"
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -98,6 +98,12 @@ aws_platform_required = pytest.mark.skipif(
     reason="Tests are not running on AWS deployed cluster"
 )
 
+rh_internal_lab_only = pytest.mark.skipif(
+    (config.ENV_DATA['platform'].lower() == 'aws'
+        or config.ENV_DATA['platform'].lower() == 'azure'),
+    reason="Tests will not run in AWS or Azure Cloud"
+)
+
 vsphere_platform_required = pytest.mark.skipif(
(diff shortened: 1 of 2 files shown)


ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -98,6 +98,12 @@
     reason="Tests are not running on AWS deployed cluster"
 )
 
+rh_internal_lab_only = pytest.mark.skipif(
+    (config.ENV_DATA['platform'].lower() == 'aws'
```
petr-balogh, line 102: Won't be easier to have only:
config.ENV_DATA['platform'].lower() != "vsphere"
?

vasukulkarni: there will be openstack and having position checks is clearer.

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -98,6 +98,12 @@
     reason="Tests are not running on AWS deployed cluster"
 )
 
+rh_internal_lab_only = pytest.mark.skipif(
```
clacroix12, line 101: Small nitpick but our other similar marks all have the term `required`, can we keep that convention here and name this `rh_internal_lab_required`?

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -98,6 +98,12 @@
     reason="Tests are not running on AWS deployed cluster"
 )
 
+rh_internal_lab_required = pytest.mark.skipif(
+    (config.ENV_DATA['platform'].lower() == 'aws'
```
keesturam, line 102: There is a problem with this approach. This is a static value and for baremetal deployments, config shows up as `AWS` as that is the default value. Maybe, we should also default `config.ENV_DATA['platform']` to 'None'?

vasukulkarni: This mark will skip if its aws or gcloud, so anything else should work, for baremetal we should fix it to use 'baremetal' instead of 'none'

clacroix12: We definitely don't want to default to `None` as the expected default behavior is to deploy on aws. We should have a "platform" for baremetal that is set in the baremetal configuration file. This way the platform does not appear as configured for aws when it is baremetal.

keesturam: As long as we don't have Baremetal deployment automated, `config.ENV_DATA['platform']` will be set to `AWS` isn't it?

clacroix12: @keesturam you can still pass a custom config file through when running tests that changes the platform to 'baremetal' or something other than 'aws'.

keesturam: @clacroix12 Thanks, that clarifies.

PR #2120, opened 2020-05-20
downstream ci: tag successful builds with latest-stable-X.Y

andrewschoen:
Signed-off-by: Andrew Schoen <aschoen@redhat.com>

I'm still waiting to get this tested in the downstream CI.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -131,13 +131,18 @@ pipeline {
     success {
       script {
         def registry_image = "${env.OCS_REGISTRY_IMAGE}"
-        // quay.io/rhceph-dev/ocs-registry:4.2-58.e59ca0f.master -> 4.2-58.e59ca0f.master
+        // quay.io/rhceph-dev/ocs-registry:4.2.2-255.ci -> 4.2.2-255.ci
         def registry_tag = registry_image.split(':')[-1]
         def ocs_version = registry_tag.split('-')[0]
         // tag ocs-registry container as 'latest-stable-$ocs_version'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -131,13 +131,18 @@ pipeline {
...
         def ocs_version = registry_tag.split('-')[0]
         // tag ocs-registry container as 'latest-stable-$ocs_version'
         build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-registry:latest-stable-${ocs_version}")]
         // tag ocs-olm-operator container as 'latest-stable'
         build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-olm-operator:latest-stable-${ocs_version}")]
+        // ocs version with only X.Y
+        def short_ocs_version = ocs_version.tokenize(".").take(2).join(".")
+        build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-registry:latest-stable-${short_ocs_version}")]
+        // tag ocs-olm-operator container as 'latest-stable'
+        build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-olm-operator:latest-stable-${short_ocs_version}")]
```
petr-balogh, line 145: cannot this job quay-tag-image take multiple tags to apply in one run? If so it will be good to do like that WDYT?
Anyway for now I am OK and can be done in future.

andrewschoen: Right now that job only supports adding a single tag at a time. That is something we could add in the future though if we find it necessary.

PR #2121, opened 2020-05-20
Run MCG CLI locally instead via operator pod

MeridianExplorer:
Instead of running the commands directly via the operator pod (which requires the NooBaa serviceaccount to have a `cluster-admin` role), we now copy the binary to the local ...ocs-ci/data/ folder, and run everything from there.

### DIFF
--- a/ocs_ci/ocs/resources/mcg.py
+++ b/ocs_ci/ocs/resources/mcg.py
@@ -11,7 +11,7 @@ import urllib3
 from botocore.client import ClientError
 
 from ocs_ci.framework import config
-from ocs_ci.ocs import constants, registry
+from ocs_ci.ocs import constants
 from ocs_ci.ocs.exceptions import CommandFailed, TimeoutExpiredError
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs.resources import pod
@@ -19,6 +19,7 @@ from ocs_ci.ocs.resources.pod import get_pods_having_label, Pod
 from ocs_ci.utility import templating
 from ocs_ci.utility.utils import TimeoutSampler, exec_cmd
 from tests.helpers import create_unique_resource_name, create_resource
+import subprocess
 
 logger = logging.getLogger(name=__file__)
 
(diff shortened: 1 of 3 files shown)


ocs_ci/ocs/resources/mcg.py
```diff
@@ -102,16 +103,28 @@ def __init__(self):
...
-            **get_pods_having_label(constants.NOOBAA_OPERATOR_POD_LABEL, self.namespace)[0]
+            **get_pods_having_label(
+                constants.NOOBAA_OPERATOR_POD_LABEL, self.namespace
+            )[0]
         )
 
+        # TODO: Add version verification to see if an additional cp isn't needed
+        # If the MCG CLI binary isn't found, copy it from the operator pod
+        if not os.path.isfile(constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH):
+            cmd = (
```
petr-balogh, line 115: @MeridianExplorer can we use `oc cp` or `oc rsync` <- for that I know we already have function for rsync.

https://www.mankier.com/1/oc-cp
https://www.mankier.com/1/oc-rsync

MeridianExplorer: Nope, sadly, we can't use either. The pod doesn't contain `tar`, nor `rsync`. There was a long discussion about this in Slack

MeridianExplorer: Also, I had to work around `shlex` splitting the command in a way that wouldn't let it run in the way I wanted it to, thus the separate use in `subprocess_run` instead of `exec_cmd`

vasukulkarni: +1 on using rsync and fixing pod in future, looks clean.

MeridianExplorer: I believe that'd up to @nimrod-becker and the team - they're the owners of the image, so it's up to them to include tar/rsync in it, if they choose :)
I'd definitely support it.

petr-balogh: ok, thanks for info. I thought that this is the case as I experienced missing tar/rsync in some of the pod in the past as well.

 Thanks

petr-balogh:
Please run pr verification with noobaa tests. Thanks

MeridianExplorer:
All tier 1 MCG tests passed successfully.
Ran locally in order to allow comfortable debugging in case of an issue - gladly, I didn't need it :)
image

ocs_ci/ocs/resources/mcg.py
```diff
@@ -102,16 +103,28 @@ def __init__(self):
...
+                f" -- cat {constants.NOOBAA_OPERATOR_POD_CLI_PATH}"
+                f"> {constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH}"
+            )
+            subprocess.run(cmd, shell=True)
+            # Add an executable bit in order to allow usage of the binary
+            exec_cmd(f'chmod +x {constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH}')
+            # Make sure the binary was copied properly and has the correct permissions
+            assert os.path.isfile(constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH)
+            assert 'ELF' in exec_cmd(f'file {constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH}').stdout.decode()
+            assert os.access(constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH, os.X_OK)
```
ebenahar, line 126: good to add assert messages

ebenahar: can be done in separate PR

PR #2122, opened 2020-05-20
separate function for updating machine configuration

vavuthu:

### DIFF
--- a/ocs_ci/deployment/vmware.py
+++ b/ocs_ci/deployment/vmware.py
@@ -406,54 +406,8 @@ class VSPHEREUPI(VSPHEREBASE):
                 f.write(terraform_config_str)
             self.terraform_var = convert_yaml2tfvars(terraform_var_yaml)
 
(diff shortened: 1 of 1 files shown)


ocs_ci/deployment/vmware.py
```diff
@@ -751,3 +705,81 @@ def clone_openshift_installer():
             constants.VSPHERE_INSTALLER_REPO, upi_repo_path,
             f'release-{ocp_version}'
         )
+
+
+def change_mem_and_cpu():
```
petr-balogh, line 710: Do you think it's worth to have it parametrized with params like worker_num_cpus, master_num_cpus, .. ?

By default set them to None and if not specified you will just load default value from config as you do below? If so we can do the same for other functions if you think it can be useful.

vavuthu: Ack. I will take it up in another PR.

PR #2136, opened 2020-05-21
Refactor tests to work on OCS 4.5, work around missing `service-ca.crt`

MeridianExplorer:
<Title> 

Also added better handling of CLI binary and `service-ca.crt` copying.

### DIFF
--- a/ocs_ci/ocs/resources/mcg.py
+++ b/ocs_ci/ocs/resources/mcg.py
@@ -766,14 +766,11 @@ class MCG(object):
             assert os.path.isfile(constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH), (
                 f'MCG CLI file not found at {constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH}'
             )
-            assert 'ELF' in exec_cmd(
-                f'file {constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH}'
-            ).stdout.decode(), (
-                f"{constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH } doesn't seem like an ELF file",
-                " and was probably not copied properly"
-            )
             assert os.access(constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH, os.X_OK), (
                 "The MCG CLI binary does not have execution permissions"
+            )
+            assert _compare_cli_hashes(), (
+                "Binary hash doesn't match the one on the operator pod"
             )
 
     @property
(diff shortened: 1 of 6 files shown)


MeridianExplorer:
@ebenahar I actually pushed it kind of urgently since it caused issues on my machines (ran OCS-CI in front of several different machines, which caused it to use an inappropriate certificate with each machine) - but it's easily fixable, and AFAIK we don't really mix `ocs-ci` environments with different clusters outside of personal use, so anything goes

MeridianExplorer:
The multiregion test failed because it shouldn't run on OCS <4.5 (fixed in #2163), teardown failed because of the must-gather.
Thus, labeling as verified.

MeridianExplorer:
The cluster was torn down by the cleanup job in the middle of the run. New runs will be executed tomorrow.

ocs_ci/ocs/resources/mcg.py
```diff
@@ -740,3 +720,49 @@ def exec_mcg_cmd(self, cmd, namespace=None, **kwargs):
...
+            # Add an executable bit in order to allow usage of the binary
+            current_file_permissions = os.stat(constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH)
+            os.chmod(
+                constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH,
+                current_file_permissions.st_mode | stat.S_IEXEC
+            )
+            # Make sure the binary was copied properly and has the correct permissions
+            assert os.path.isfile(constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH)
+            assert 'ELF' in exec_cmd(f'file {constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH}').stdout.decode()
+            assert os.access(constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH, os.X_OK)
```
ebenahar, line 768: please add assert messages

tests/conftest.py
```diff
@@ -1726,7 +1729,10 @@ def bucket_factory_fixture(request, mcg_obj):
         'cli': CLIBucket
     }
 
-    def _create_buckets(amount=1, interface='S3', *args, **kwargs):
+    def _create_buckets(
```
ebenahar, line 1732: make sure that this function usages are not affected

MeridianExplorer: Doesn't seem like we ran into any unknown bugs/issues in the run, so I believe it is safe 👍

mbukatov:
For the record, this is related to BZ 1845188.

PR #2139, opened 2020-05-21
Work around copying the MCG cert to the AWSCLI pod

MeridianExplorer:
The workaround was needed after we changed the AWSCLI pod's image *and* started verifying ther MCG certificates. We tested each change separately, but not together - and thus they were merged.

Also fix the signature of `craft_s3_command` that wasn't updated properly in a #2080

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,6 +10,7 @@ from math import floor
 from random import randrange
 from time import sleep
 from shutil import copyfile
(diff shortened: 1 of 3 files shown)


MeridianExplorer:
One test failed because of an incorrect call to `craft_s3_command`. Fixed in e6e1091. Rerun is undergroing.

Issue #2140, opened 2020-05-21
number of PVs created is not verified in verify_pvs_created func


vavuthu:
number of PVs created is not verified in verify_pvs_created func

2020-05-21 23:23:33,840 - INFO - ocs_ci.deployment.deployment.verify_pvs_created.831 - local-pv-548b0d83 is in Available state
2020-05-21 23:23:33,844 - DEBUG - ocs_ci.deployment.deployment.verify_pvs_created.836 - PVs, Workers: 1, 3
2020-05-21 23:23:33,845 - INFO - ocs_ci.deployment.deployment.deploy_ocs_via_operator.410 - Deployment of OCS via OCS operator

PR #2141, opened 2020-05-21
Assert when the number of PVs created doesnot match with expected number

vavuthu:
Assert when the number of PVs created doesnot match with expected number

Fixes: #2140


vavuthu:
Closing this as this created wrongly on top of another branch

Issue #2147, opened 2020-05-24
resourrces/{pod,ocs}.py do not free NamedTemporaryFile file descriptors.


wusui:
The file descriptors created by NamedTemporaryFile calls in pod.py and ocs.py are never closed.

It appear that the only field used by the code is the file object name field.  self.temp_yaml should be modified to only save the name field, and the NamedTemporaryFile calls should be made from a
with statement.

Without this change, some scale tests will hit the ulimit value for open file descriptors.

PR #2148, opened 2020-05-24
Fix issue 2147

wusui:
Fixes: #2147

Previous NamedTemporaryFile calls saved the entrire file
object and did not release the file descriptor.  Now the
calls only save the name of the file (the only data the
code used from the file object).  The NamedTemporaryFile
call is made within a with clause to insure that the
file descriptor is released.


vasukulkarni:
@wusui to run tier1 for this pr.

PR #2150, opened 2020-05-24
Cluster load: Determine cluster limit by latency growth + use DC pods

ebenahar:
A follow-up on https://github.com/red-hat-storage/ocs-ci/pull/2037 - improvements and fixes for cluster load functionality:

- Detect cluster limit by an exponential increase of the latency
- Instead of background thread to watch the IO load, use DeploymentConfig pods of FIO, which will spin up automatically in case they are stopped during the execution 
- Track cluster latency in the background. Actively adjust the background load in case the latency increases
- Added a few backup mechanisms to determine the cluster limit, in case the latency does not grow exponentially 
- Change the default load from 30% to 50%
- Use Prometheus queries to get the different metrics
- Improve logging
- Change PVC size to be 50% of the sum of the OSD memory pods memory, for all created PVCs
- Examine IOPS instead of throughput 
- Calculate different metrics based on trimmed mean average instead of mean average
- cli param, `--log-cluster-utilization`, to log cluster utilization metrics (latency, IOPS, throughput, used space and number of background FIO pods) every 10 seconds during test execution
- Few other minor fixes

### DIFF
--- a/ocs_ci/ocs/cluster_load.py
+++ b/ocs_ci/ocs/cluster_load.py
@@ -4,12 +4,19 @@ A module for cluster load related functionalities
 """
 import logging
 import time
+from datetime import datetime
+from uuid import uuid4
 
-import ocs_ci.ocs.constants as constant
-from ocs_ci.ocs import constants
+from ocs_ci.utility.prometheus import PrometheusAPI
+from ocs_ci.utility.utils import get_trim_mean
+from ocs_ci.utility import templating
+from ocs_ci.ocs import constants, defaults
 from ocs_ci.ocs.cluster import (
-    CephCluster, get_osd_pods_memory_sum, get_percent_used_capacity
+    get_osd_pods_memory_sum, get_percent_used_capacity
 )
+
+
+logger = logging.getLogger(__name__)
 
 
 class ClusterLoad:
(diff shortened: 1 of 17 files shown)


ebenahar:
In the PR validation job execution above (https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/7956/consoleFull), while tier1 tests were running, the throughput load dropped twice below 60% of the target percentage (50%). When that happened, the thread that keeps watching the load kicked-in and re-triggered the IO load.
It happened here:

```
11:12:27 - Thread-1 - tests.conftest - WARNING - 
====================================================================================================
Cluster throughput dropped below 60% of the target throughput for background IO load. Re-loading IOs
====================================================================================================
```

And here:
```
11:16:03 - Thread-1 - tests.conftest - WARNING - 
====================================================================================================
Cluster throughput dropped below 60% of the target throughput for background IO load. Re-loading IOs
====================================================================================================
```

Which correspond to the graph below (with 3 hours timezone difference):
Screenshot from 2020-05-24 15-54-28

ocs_ci/ocs/cluster_load.py
```diff
@@ -58,218 +58,204 @@ def reach_cluster_load_percentage_in_throughput(
...
+            "to reach cluster throughput utilization that is more or less the "
+            "target throughput percentage"
+        )
+        while current_throughput > (target_throughput * 1.1) and len(pod_objs) > 1:
+            pod_name = pod_objs[-1].name
+            pod_objs[-1].delete()
+            pod_objs[-1].ocp.wait_for_delete(pod_objs[-1].name)
+            pod_objs.remove(pod_objs[-1])
+            pvc_objs[-1].delete()
+            pvc_objs[-1].ocp.wait_for_delete(pvc_objs[-1].name)
```
shyRozen, line 249: I think you need to take the name again for the wait_for_delete

ocs_ci/ocs/cluster_load.py
```diff
@@ -18,263 +19,257 @@ class ClusterLoad:
...
         bs = '128K'
+        rw_ratio = 25
+        pvc_size = int(get_osd_pods_memory_sum() * 0.5)
+        io_file_size = f"{pvc_size - 1}G"
 
-        if 0.1 < target_percentage > 0.95:
-            self.logger.warning(
-                f"The target percentage is {target_percentage * 100}% which is not "
-                f"within the accepted range. Therefore, IO will not be started"
+        def create_pvc_and_pod():
```
shyRozen, line 74: This is a suggestion only. From my experience creating pods will tend to stick to 2 nodes out of 3 for most of them. I think for better throughput results divide them manually between nodes.

ebenahar: That's a very good input!
I will need to check if indeed pods are scheduled only on 2 nodes out of 3, but pinning a pod to a specific node could be a bit problematic as sometimes a node cannot schedule new pods due to lack of resources

ocs_ci/ocs/cluster_load.py
```diff
@@ -18,263 +19,257 @@ class ClusterLoad:
...
         self.cl_obj = CephCluster()
+        self.pvc_factory = pvc_factory
+        self.pod_factory = pod_factory
+        self.target_percentage = target_percentage
+        self.target_throughput = None
 
-    def reach_cluster_load_percentage_in_throughput(
-        self, pvc_factory, pod_factory, target_percentage=0.3, cluster_limit=None
-    ):
+    def reach_cluster_load_percentage_in_throughput(self, ):
```
AviadP, line 41:
```suggestion
    def reach_cluster_load_percentage_in_throughput(self):
```

ebenahar: fixed

ocs_ci/utility/environment_check.py
```diff
@@ -93,9 +93,11 @@ def assign_get_values(
 
     for item in items:
         ns = item.get('metadata', {}).get('namespace')
+        if item.get('kind') == constants.PV:
+            ns = item.get('spec').get('claimRef').get('namespace')
```
yitzhak12, line 97: I think it will be better to change it to this: `ns = item.get('spec', {}).get('claimRef', {}).get('namespace')`

ebenahar: +1, will do

fbalak: I don't think that this is useful here. There needs to be returned `namespace` value from a given place in the dictionary. This way if some of the keys is not present, the fail will be silent and it will fail on `.get('namespace')` because there will be no 'namespace' key in empty dictionary.
We will not know which key was missing (if `spec` or `claimRef`) because on one or two of those steps was returned {} without error and Python won't provide information what key was missing right away.

ocs_ci/ocs/cluster_load.py
```diff
@@ -4,277 +4,335 @@
...
+        # FIO pods with large FIO 'rate' param
+        rate = '15M'
+        self.increase_load(rate=rate)
+        target_iops = self.cluster_limit * self.target_percentage
+        current_iops = self.get_query(query=constants.IOPS_QUERY)
+        logger.info(f"Target IOPS: {target_iops}")
+        logger.info(f"Current IOPS: {current_iops}")
+
+        while current_iops < target_iops * 0.95:
+            logger.info(
```
fbalak, line 244: This log message should be probably before first load is started (line 237)

ebenahar:
Hi Filip,

Thanks for the great feedback!


> if this will be the case then in future PR we can add some marker like @nobackgroundworkload to turn it off for some tests.

I actually had a conversation about this with @shyRozen and we thought about skipping certain tests, in case background IO is running.
We can add this decorator in a following PR

ocs_ci/ocs/cluster_load.py
ebenahar: correct, will fix

tests/conftest.py
```diff
@@ -1093,37 +1146,54 @@ def finalizer():
...
-                                cluster_limit=cluster_limit
+                    cl_load_obj.print_metrics()
+                    if config.RUN.get('io_in_bg'):
+                        latency = cl_load_obj.calc_trim_metric_mean(
+                            constants.LATENCY_QUERY
+                        )
+
+                        if latency > 1 and len(cl_load_obj.dc_objs) > 0:
+                            log.warning(
+                                "Latency is higher than 1 second. Lowering IO load"
```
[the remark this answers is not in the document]
ebenahar: will do

tests/conftest.py
```diff
@@ -1093,37 +1146,54 @@ def finalizer():
...
+                                "less than 0.5 seconds, FIO pod will be re-spawned"
+                            )
+                            cl_load_obj.decrease_load()
+
+                        diff = initial_num_of_pods - len(cl_load_obj.dc_objs)
+                        while latency < 0.5 and diff > 0 and (
+                            get_test_status() == 'running'
+                        ):
+                            log.info(
+                                "Latency is lower than 0.5 seconds. "
```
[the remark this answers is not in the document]
ebenahar: will do

ocs_ci/utility/environment_check.py
ebenahar: reverted

vasukulkarni:
👍

Issue #2165, opened 2020-05-26
test_automated_recovery_from_failed_nodes_IPI_proactive should validate also '-1-deploy', 'ocs-deviceset' pods


ebenahar:
Here - https://github.com/red-hat-storage/ocs-ci/blob/8eb4a1b26c1d686fa66ddd25cec1e28f54616329/tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_proactive_IPI.py#L137, instead of skipping the validation for '-1-deploy', 'ocs-deviceset' pods if they reached `Running` state, we should validate they reached `Completed` state

PR #2166, opened 2020-05-26
Fix node replacement tests to validate also OSD prepare and DC pods

ebenahar:
- Fixes https://github.com/red-hat-storage/ocs-ci/issues/2165
- Also move the pods validation section from the different test files to a common helpers file

### DIFF
--- a/tests/manage/z_cluster/nodes/helpers.py
+++ b/tests/manage/z_cluster/nodes/helpers.py
@@ -1,0 +1,42 @@
+import logging
+
+from ocs_ci.ocs import constants, defaults, ocp
+from ocs_ci.ocs.resources import pod
+from ocs_ci.ocs.exceptions import ResourceWrongStatusException
+from tests import helpers
+
+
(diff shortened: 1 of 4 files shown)


tests/manage/z_cluster/nodes/helpers.py
```diff
@@ -0,0 +1,42 @@
...
+from tests import helpers
+
+
+log = logging.getLogger(__name__)
+
+
+def wait_for_all_pods():
+    """
+    Check all OCS pods status, they should be in running state
+
```
MeridianExplorer, line 15: Missing `Raises:` docstring?

ebenahar: it raises any exception which is not ResourceWrongStatusException. So not sure if we can add this to the docstring

ebenahar:
Failures are not related to the change done in this PR. They were seen also in recent tier4b runs (like RUN ID: 1590390361). Hence, PR is verified

PR #2171, opened 2020-05-27
change method of workload_storageutilization_checksum_rbd to use static size

fbalak:
This change is done because utilization tests that should utilize cluster for a small percentage can become flaky because of utilization that is already run.


fbalak:
Upgrade test failed because of Ceph health and noobaa tests that now use certificate for communication failed because of error that should be resolved by https://github.com/red-hat-storage/ocs-ci/pull/2136. Test test_workload_with_checksum that should be affected by this change passed correctly.

PR #2177, opened 2020-05-27
Upgrade logging post OCP upgrade

Shrivaibavi:

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -125,6 +125,7 @@ UPGRADE:
   ocp_channel: "candidate-4.5"
   ocp_upgrade_path: "quay.io/openshift-release-dev/ocp-release"
   ocp_arch: "x86_64"
+  upgrade_logging_channel: "4.4"
 
 # This section stores secret and uploaded from home dir or s3
 # for entry into this section, please email ecosystem team
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -125,6 +125,8 @@ UPGRADE:
   ocp_channel: "candidate-4.5"
   ocp_upgrade_path: "quay.io/openshift-release-dev/ocp-release"
   ocp_arch: "x86_64"
+  # upgrade_logging_channel param needs change from 4.4 to 4.5 once OCP 4.5 GAed
+  # Issue to track https://github.com/red-hat-storage/ocs-ci/issues/2390
   upgrade_logging_channel: "4.4"
 
 # This section stores secret and uploaded from home dir or s3
--- a/tests/ecosystem/upgrade/test_logging_upgrade.py
+++ b/tests/ecosystem/upgrade/test_logging_upgrade.py
@@ -1,0 +1,232 @@
+import logging
+
+import pytest
+
+from ocs_ci.ocs.cluster import CephCluster, CephHealthMonitor
+from ocs_ci.ocs.resources.csv import CSV
+from ocs_ci.utility.deployment_openshift_logging import (
+    check_health_of_clusterlogging, get_clusterlogging_subscription
+)
+from ocs_ci.ocs.exceptions import CommandFailed
+from ocs_ci.framework import config
+from ocs_ci.ocs.resources.packagemanifest import PackageManifest
+from ocs_ci.framework.pytest_customization.marks import (
+    post_ocp_upgrade
+)
+from ocs_ci.ocs.resources.pod import get_all_pods
+from ocs_ci.ocs import constants, ocp
+from ocs_ci.utility.retry import retry
+from ocs_ci.utility.utils import (
+    get_ocp_version, run_cmd
+)
+
+logger = logging.getLogger(__name__)
+
+
+def check_cluster_logging():
+    """
+    Few checks to assert for before and after upgrade
+    1. Checks for pods in openshift-logging project
+    2. Checks for the health of logging cluster before upgrade
+
+    """
+
+    assert get_all_pods(
+        namespace=constants.OPENSHIFT_LOGGING_NAMESPACE
+    ), 'Some or all pods missing in namespace'
+
+    assert check_health_of_clusterlogging(), "Cluster is not Healthy"
+
+
+@retry(CommandFailed, 5, 30, 2)
+def get_csv_version(channel):
+    """
+    Get the cluster-logging and Elasticsearch CSV version
+    and Images of the pods
+
+    Args:
+        channel (int) : Logging Channel
+
+    Returns:
+        cluster_logging_csv (str) : Name of the cluster-logging CSV
+        elasticsearch_csv (str) : Name of the elasticsearch CSV
+        dict: Images dict like: {'image_name': 'image.url.to:tag', ...}
+
+    """
+
+    clo_package_manifest = PackageManifest(
+        resource_name='cluster-logging'
+    )
+    cluster_logging_csv = clo_package_manifest.get_current_csv(channel)
+    es_package_manifest = PackageManifest(
+        resource_name='elasticsearch-operator'
+    )
+    elasticsearch_csv = es_package_manifest.get_current_csv(channel)
+    logging_csv = CSV(
+        resource_name=cluster_logging_csv,
+        namespace=constants.OPENSHIFT_LOGGING_NAMESPACE
+    )
+    images = ocp.get_images(logging_csv.get())
+    return cluster_logging_csv, elasticsearch_csv, images
+
+
+def check_csv_version_post_upgrade(channel):
+    """
+    This function is to check the CSV version post upgrade
+
+    Args:
+        channel (int) : logging channel
+
+    Returns:
+        bool : Returns True if version matches, False otherwise
+
+    """
+
+    cluster_logging_csv, elasticsearch_csv, images = get_csv_version(channel)
+    if config.UPGRADE['upgrade_logging_channel'] in (
+        cluster_logging_csv and elasticsearch_csv
+    ):
+        logger.info(
+            f"Upgraded version of Cluster-logging "
+            f"operator {cluster_logging_csv}"
+        )
+        logger.info(
+            f"Upgraded version of Elastic-search "
+            f"operator {elasticsearch_csv}"
+        )
+        return True
+    return False
+
+
+@retry(AttributeError, 5, 30, 2)
+def check_csv_logging_phase(channel):
+    """
+    The function checks in the CSV for the phase "Succeeded"
+    for cluster-logging operator and Elastic-search
+    operator
+
+    Args:
+        channel (int) : Logging channel
+
+    Returns:
+        bool: True if CSV status is Succeeded, False otherwise
+
+    """
+
+    cluster_version_csv, elasticsearch_csv, images = get_csv_version(channel)
+    logging_csv = CSV(
+        resource_name=cluster_version_csv,
+        namespace=constants.OPENSHIFT_LOGGING_NAMESPACE
+    )
+    clo_phase = logging_csv.get().get('status').get('phase')
+    logging_csv.get().get('status').get('phase')
+
+    es_csv = CSV(
+        resource_name=elasticsearch_csv,
+        namespace=constants.OPENSHIFT_OPERATORS_REDHAT_NAMESPACE
+    )
+    eso_phase = es_csv.get().get('status').get('phase')
+
+    if clo_phase and eso_phase == 'Succeeded':
+        logger.info("CSV phase Succeeded")
+        return True
+
+    logger.info("Upgrade not completed yet")
+    return False
+
+
+def upgrade_info(channel):
+    """
+     Function to provide CSV names and images pre and post upgrade
+
+     Args:
+         channel (int): Logging channel
+
+    """
+
+    logging_csv, elasticsearch_csv, images = get_csv_version(channel)
+    logger.info(
+        f"Cluster-logging CSV for channel {channel} "
+        f"{logging_csv}"
+    )
+    logger.info(
+        f"Elastic-search CSV for channel {channel} "
+        f"{elasticsearch_csv}"
+    )
+    logger.info(
+        f"Images of cluster-logging components for channel {channel} "
+        f"{images}"
+    )
+
+
+@post_ocp_upgrade
+@pytest.mark.polarion_id("OCS-2201")
+class TestUpgradeLoggingAfterOCPUpgrade():
+    """
+    This class contains test for upgrade openshift-logging
+    after OCP upgrade
+    1. Monitors OCS health
+    2. Checks OCP version
+    3. Checks logging version, if mismatched then upgrades logging
+    4. Checks Elasticsearch and clusterlogging operator
+    5. Checks logging health post upgrade
+
+    """
+
+    def test_upgrade_logging(self):
+        """
+        This function contains test to upgrade openshift-logging
+        with Entry and Exit criteria for checks
+
+        """
+
+        ceph_cluster = CephCluster()
+        with CephHealthMonitor(ceph_cluster):
+
+            #  Pre-check
+            check_cluster_logging()
+
+            # Matching the OCP version and cluster-Logging version
+            ocp_version = get_ocp_version()
+            logger.info(f"OCP version {ocp_version}")
+            subscription = get_clusterlogging_subscription()
+            logging_channel = subscription.get('spec').get('channel')
+            logger.info(
+                f"Current Logging channel {logging_channel}"
+            )
+            upgrade_info(logging_channel)
+            upgrade_channel = config.UPGRADE['upgrade_logging_channel']
+            if ocp_version > logging_channel:
+                # Upgrade Elastic search operator Subscription
+                es_subscription_cmd = (
+                    'oc patch subscription elasticsearch-operator '
+                    f'-n {constants.OPENSHIFT_OPERATORS_REDHAT_NAMESPACE} '
+                    '--type merge -p \'{{"spec":{{"channel": '
+                    f'"{upgrade_channel}"}}}}\''
+                )
+                # Upgrade Cluster-logging operator subscription
+                clo_subscription_cmd = (
+                    'oc patch subscription cluster-logging '
+                    f'-n {constants.OPENSHIFT_LOGGING_NAMESPACE} '
+                    '--type merge -p \'{{"spec":{{"channel": '
+                    f'"{upgrade_channel}"}}}}\''
+                )
+                run_cmd(es_subscription_cmd)
+                run_cmd(clo_subscription_cmd)
+                assert check_csv_logging_phase(
+                    upgrade_channel
+                ), "Logging upgrade not completed yet!"
+                logger.info(
+                    "Logging upgrade completed!"
+                )
+                assert check_csv_version_post_upgrade(
+                    upgrade_channel
+                ), "Unable to get version"
+                logger.info("Version Matches!")
+                upgrade_info(upgrade_channel)
+                check_cluster_logging()
+            else:
+                logger.info(
+                    "Logging Version matches the OCP version, "
+                    "No upgrade needed"
+                )
--- a/tests/ecosystem/upgrade/test_logging_upgrade.py
+++ b/tests/ecosystem/upgrade/test_logging_upgrade.py
@@ -38,14 +38,14 @@ def check_cluster_logging():
     assert check_health_of_clusterlogging(), "Cluster is not Healthy"
 
 
-@retry(CommandFailed, 5, 30, 2)
+@retry(CommandFailed, tries=5, delay=120, backoff=1)
 def get_csv_version(channel):
     """
     Get the cluster-logging and Elasticsearch CSV version
     and Images of the pods
 
     Args:
-        channel (int) : Logging Channel
+        channel (str) : Logging Channel
 
     Returns:
         tuple: Tuple containing three elements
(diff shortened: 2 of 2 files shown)


tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,165 @@
...
+                        f"{'completed!' if sampler else 'did not completed yet!'}"
+                    )
+                    if sampler:
+                        break
+            else:
+                logger.info(
+                    "Logging Version matches the OCP version, No upgrade needed"
+                )
+
+            check_cluster_before_after_upgrade()
```
hnallurv, line 165: Add a check to see the versions of the logging are correct after the upgrade.

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,165 @@
...
+)
+from ocs_ci.ocs import constants, ocp
+from ocs_ci.utility.utils import (
+    TimeoutSampler, get_ocp_version, run_cmd
+)
+
+logger = logging.getLogger(__name__)
+
+
+def check_cluster_before_after_upgrade():
```
hnallurv, line 21: Suggest changing the name to "check_cluster_logging" as this can be called before and after the upgrade.

tests/ecosystem/upgrade/test_logging_upgrade.py
Shrivaibavi: Added

tests/ecosystem/upgrade/test_logging_upgrade.py
Shrivaibavi: Done

Shrivaibavi:
Passed Logs http://pastebin.test.redhat.com/869610

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,194 @@
...
+    """
+
+    assert get_all_pods(
+        namespace=constants.OPENSHIFT_LOGGING_NAMESPACE
+    ), 'Some or all pods missing in namespace'
+
+    assert check_health_of_clusterlogging(), "Cluster is not Healthy"
+
+
+def check_csv_logging_upgrade_phase():
```
AviadP, line 35: What happen if upgrade did not start at all? i dont see any version validation here - can status be "succeeded" before upgrade even starts?
also, it seems that this function is more general, checking just the status of CSV, regardless if upgrade is triggered or not. function  and variables name should reflect that - my suggestion here is to rename this function and call it every time when this check needed

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+    if clo_phase_after_upgrade and eso_phase_after_upgrade == 'Succeeded':
+        logger.info("CSV phase Succeeded")
+        return True
+
+    logger.debug("Upgrade not completed yet")
+    return False
+
+
+def check_version_after_upgrade():
```
AviadP, line 62: same comments as i wrote for ```check_csv_logging_upgrade_phase```

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,194 @@
...
+            f"Upgraded version of Elastic-search "
+            f"operator {eso_version_after_upgrade}"
+        )
+        return True
+    return False
+
+
+@post_ocp_upgrade
+@pytest.mark.polarion_id("OCS-2201")
+class TestUpgradeLoggingAfterOCPUpgrade():
```
AviadP, line 95: Can logging upgrade be performed without upgrading OCP  first?

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -126,6 +126,7 @@ UPGRADE:
   ocp_channel: "candidate-4.4"
   ocp_upgrade_path: "quay.io/openshift-release-dev/ocp-release"
   ocp_arch: "x86_64"
+  upgrade_logging_channel: "4.4"
```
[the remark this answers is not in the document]
Shrivaibavi: So in logging we only change the channel from 4.3 to 4.4. The channel changing is suffice as OCP 4.4 will have all necessary images for logging to upgrade. In logging the channel is just named as 4.4 or 4.3 , Please see link below
https://docs.openshift.com/container-platform/4.4/logging/cluster-logging-upgrading.html

tests/ecosystem/upgrade/test_logging_upgrade.py
Shrivaibavi: No logging upgrade cannot be performed before the OCP upgrade, Hence i used the decorator @post_ocp_upgrade. is it not enough ?

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,194 @@
...
+from ocs_ci.ocs import constants, ocp
+from ocs_ci.utility.utils import (
+    TimeoutSampler, get_ocp_version, run_cmd
+)
+
+logger = logging.getLogger(__name__)
+
+
+def check_cluster_logging():
+    """
```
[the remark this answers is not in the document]
Shrivaibavi: Done

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,194 @@
...
+
+    assert check_health_of_clusterlogging(), "Cluster is not Healthy"
+
+
+def check_csv_logging_upgrade_phase():
+    """
+    The function checks in the CSV for the phase "Succeeded"
+    for cluster-logging operator and Elastic-search
+    operator
+
```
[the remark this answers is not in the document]
Shrivaibavi: done

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,194 @@
...
+        csv_after_upgrade.get('items')[0].get('status').get('phase')
+    )
+    eso_phase_after_upgrade = (
+        csv_after_upgrade.get('items')[1].get('status').get('phase')
+    )
+    if clo_phase_after_upgrade and eso_phase_after_upgrade == 'Succeeded':
+        logger.info("CSV phase Succeeded")
+        return True
+
+    logger.debug("Upgrade not completed yet")
```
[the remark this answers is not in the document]
Shrivaibavi: done

tests/ecosystem/upgrade/test_logging_upgrade.py
Shrivaibavi: I already check for version before and after upgrade and also check for the Succeeded phase after the upgrade. Also using timeout sampler to check versions and phase after upgrade. Hope this checks are enough. Please let me know.
check_version_after_upgrade(): -----Will have updated CSV after upgrade.

tests/ecosystem/upgrade/test_logging_upgrade.py
Shrivaibavi: Done

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,201 @@
...
+    The function checks in the CSV for the phase "Succeeded"
+    for cluster-logging operator and Elastic-search
+    operator
+
+    Returns:
+        bool: True if CSV status is Succeeded, False otherwise
+
+    """
+
+    csv = ocp.OCP(
```
petr-balogh, line 47: Do you know we have CSV class directly in the resources?
https://github.com/red-hat-storage/ocs-ci/blob/045b0a5821a2ce6dc54956ddac532df8425bb8d3/ocs_ci/ocs/resources/csv.py

So please use directly CSV if you want to have CSV resource.
Also we have already implemented some functionality related to phase under OCP class here:

https://github.com/red-hat-storage/ocs-ci/blob/045b0a5821a2ce6dc54956ddac532df8425bb8d3/ocs_ci/ocs/ocp.py#L744

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,201 @@
...
+
+def check_version_after_upgrade():
+    """
+    Checks the version after upgrade
+
+    Returns:
+        bool: True if CSV version is correct after upgrade, False otherwise
+
+    """
+    csv = ocp.OCP(
```
petr-balogh, line 74: use CSV object instead of OCP

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,201 @@
...
+
+    """
+
+    csv = ocp.OCP(
+        kind=constants.CLUSTER_SERVICE_VERSION,
+        namespace=constants.OPENSHIFT_LOGGING_NAMESPACE
+    )
+    csv_after_upgrade = csv.get(out_yaml_format=True)
+    clo_phase_after_upgrade = (
+        csv_after_upgrade.get('items')[0].get('status').get('phase')
```
petr-balogh, line 53: I am concerned here with indexes of items as it can take some time till there are 2 CSVs when upgrade start rolling.

Also the second one CSV will disappear once the upgrade finish AFAIK.

Also not sure if we can rely that 1st item will be always the first CSV as I don't know how OCS returns those. 

As in the packageManifest you should be able to get current (default) CSV name for specific channel you should access CSV directly with it's name IMO.

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,201 @@
...
+    Returns:
+        bool: True if CSV version is correct after upgrade, False otherwise
+
+    """
+    csv = ocp.OCP(
+        kind=constants.CLUSTER_SERVICE_VERSION,
+        namespace=constants.OPENSHIFT_LOGGING_NAMESPACE
+    )
+    csv_after_upgrade = csv.get(out_yaml_format=True, wait=30)
+    clo_version_after_upgrade = (
```
petr-balogh, line 79: Same concern with indexes.

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,201 @@
...
+                namespace=constants.OPENSHIFT_LOGGING_NAMESPACE
+            )
+            csv_info = csv.get(
+                out_yaml_format=True
+            )
+            clo_version_before_upgrade = (
+                csv_info.get('items')[0].get('spec').get('version')
+            )
+            eso_version_before_upgrade = (
+                csv_info.get('items')[1].get('spec').get('version')
```
petr-balogh, line 141: After OCP upgrade we directly see both CSVs? Do you have the output of
```
oc get csv -n OPENSHIFT_LOGGING_NAMESPACE
```
Please replace the OPENSHIFT_LOGGING_NAMESPACE with relevant one.
Just curious how they are looking and don't have such cluster after upgrade or in upgrade process.

If I will see it for what we are looking for exactly I will maybe better understand it.

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,201 @@
...
+            subscription = get_clusterlogging_subscription()
+            logging_channel = subscription.get('spec').get('channel')
+            logger.info(
+                f"Logging channel before upgrade {logging_channel}"
+            )
+            upgrade_channel = config.UPGRADE['upgrade_logging_channel']
+            if ocp_version > logging_channel:
+                # Upgrade Elastic search operator Subscription
+                es_subscription_cmd = (
+                    f'oc patch subscription elasticsearch-operator '
```
petr-balogh, line 160: you do not expand any variable here so `f` should be deleted

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,201 @@
...
+                    f'-n {constants.OPENSHIFT_OPERATORS_REDHAT_NAMESPACE} '
+                    f'--type merge -p \'{{"spec":{{"channel": '
+                    f'"{upgrade_channel}"}}}}\''
+                )
+                # Upgrade Cluster-logging operator subscription
+                clo_subscription_cmd = (
+                    f'oc patch subscription cluster-logging '
```
petr-balogh, line 167: same

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,201 @@
...
+            logger.info(
+                f"Logging channel before upgrade {logging_channel}"
+            )
+            upgrade_channel = config.UPGRADE['upgrade_logging_channel']
+            if ocp_version > logging_channel:
+                # Upgrade Elastic search operator Subscription
+                es_subscription_cmd = (
+                    f'oc patch subscription elasticsearch-operator '
+                    f'-n {constants.OPENSHIFT_OPERATORS_REDHAT_NAMESPACE} '
+                    f'--type merge -p \'{{"spec":{{"channel": '
```
petr-balogh, line 162: here as well, so if you will remove `f` you will probably won't need to escape `{`

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,201 @@
...
+                    f'"{upgrade_channel}"}}}}\''
+                )
+                # Upgrade Cluster-logging operator subscription
+                clo_subscription_cmd = (
+                    f'oc patch subscription cluster-logging '
+                    f'-n {constants.OPENSHIFT_LOGGING_NAMESPACE} '
+                    f'--type merge -p \'{{"spec":{{"channel": '
```
petr-balogh, line 169: same

tests/ecosystem/upgrade/test_logging_upgrade.py
Shrivaibavi: Elasticsearch  is deployed on all namespaces by default so you will see 2 CSVs in openshift-logging as well.

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,232 @@
...
+@retry(CommandFailed, 5, 30, 2)
+def get_csv_version(channel):
+    """
+    Get the cluster-logging and Elasticsearch CSV version
+    and Images of the pods
+
+    Args:
+        channel (int) : Logging Channel
+
+    Returns:
```
AviadP, line 50: it actually return tuple that contains this 3 components

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+        return True
+    return False
+
+
+@retry(AttributeError, 5, 30, 2)
+def check_csv_logging_phase(channel):
+    """
+    The function checks in the CSV for the phase "Succeeded"
+    for cluster-logging operator and Elastic-search
```
AviadP, line 105:
```suggestion
    for cluster-logging operator and Elastic-search operator
```

tests/ecosystem/upgrade/test_logging_upgrade.py
AviadP: using `post_ocp_upgrade` marker is ok, question was about Class name
the test itself does not trigger ocp upgrade, marker means that it will run if this marker will be selected as run-ci parameter

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,232 @@
...
+        """
+        This function contains test to upgrade openshift-logging
+        with Entry and Exit criteria for checks
+
+        """
+
+        ceph_cluster = CephCluster()
+        with CephHealthMonitor(ceph_cluster):
+
+            #  Pre-check
```
AviadP, line 186: you may want to add some logging information here, if test failed in this stage it would me easier to debug, something like:
```logger.info("Checking cluster logging before starting to upgrade")```

File: tests/ecosystem/upgrade/test_logging_upgrade.py [hunk 22]
```diff
@@ -0,0 +1,232 @@
...
+        channel (int) : logging channel
+
+    Returns:
+        bool : Returns True if version matches, False otherwise
+
+    """
+
+    cluster_logging_csv, elasticsearch_csv, images = get_csv_version(channel)
+    if config.UPGRADE['upgrade_logging_channel'] in (
+        cluster_logging_csv and elasticsearch_csv
```
fbalak, line 87: This check is probably wrong:
```
("somestring" and "someotherstring")
```
returns `"someotherstring"`.

That means
```
(
    cluster_logging_csv and elasticsearch_csv
)
```
returns `elasticsearch_csv` so `cluster_logging_csv` is omitted entirely.

Shouldn't this be something like this?
```
    if (
        config.UPGRADE['upgrade_logging_channel'] == cluster_logging_csv and
        config.UPGRADE['upgrade_logging_channel'] == elasticsearch_csv
    ):
```

fbalak:
Looks good. I just have one question.

tests/ecosystem/upgrade/test_logging_upgrade.py
Shrivaibavi: done

tests/ecosystem/upgrade/test_logging_upgrade.py
Shrivaibavi: done

Shrivaibavi:
Passed logs http://pastebin.test.redhat.com/877359

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+@retry(CommandFailed, 5, 30, 2)
+def get_csv_version(channel):
+    """
+    Get the cluster-logging and Elasticsearch CSV version
+    and Images of the pods
+
+    Args:
+        channel (int) : Logging Channel
```
petr-balogh, line 48: I see that you are passing value from upgrade_logging_channel from config file https://github.com/red-hat-storage/ocs-ci/pull/2177/files#diff-4f80328bffaf5599157179f9e1a09464R126 which is string

`upgrade_logging_channel: "4.4"`

So please change it properly to `str` and maybe add some example of value what channel value can be.

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,234 @@
...
+    images = ocp.get_images(logging_csv.get())
+    return cluster_logging_csv, elasticsearch_csv, images
+
+
+def check_csv_version_post_upgrade(channel):
+    """
+    This function is to check the CSV version post upgrade
+
+    Args:
+        channel (int) : logging channel
```
petr-balogh, line 79: The same?

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -123,6 +123,7 @@ UPGRADE:
   ocp_channel: "candidate-4.5"
   ocp_upgrade_path: "quay.io/openshift-release-dev/ocp-release"
   ocp_arch: "x86_64"
+  upgrade_logging_channel: "4.4"
```
petr-balogh, line 126: I guess this "4.4" here is cause of OCP 4.5 is not GAed yet, so please can you open W/A issue to change this back once 4.5 is GAed? And add link to this issue to comment above this line?

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,234 @@
...
+        )
+        logger.info(
+            f"Upgraded version of Elastic-search "
+            f"operator {elasticsearch_csv}"
+        )
+        return True
+    return False
+
+
+@retry(AttributeError, 8, 30, 2)
```
petr-balogh, line 103: number 2 -  backoff: backoff multiplier e.g. value of 2 will double the delay each retry

Which will leads to about 2 hours - is this intention here?

The problem I see that if it will fail after delay will be increased to 3840 seconds it's a lot of time of waiting and doing nothing.

I would rather have backoff set to 1, and extend delay to some reasonable time to be able to check once a let say 1 or few minutes. It can save a lot of time and the calculation of maximum wait time will be more straight forward. 

For example 90 min and retry each 60 seconds = 
`@retry(AttributeError, 90, 60, 1)` .

tests/ecosystem/upgrade/test_logging_upgrade.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+
+    assert get_all_pods(
+        namespace=constants.OPENSHIFT_LOGGING_NAMESPACE
+    ), 'Some or all pods missing in namespace'
+
+    assert check_health_of_clusterlogging(), "Cluster is not Healthy"
+
+
+@retry(CommandFailed, 5, 30, 2)
```
petr-balogh, line 41: same comment with retry bellow. I prefer to avoid backoff to 2 and rather use 1 here.

ocs_ci/framework/conf/default_config.yaml
Shrivaibavi: so can we track with this issue itself, If @petr-balogh would suggest I would add a point in the same issue https://github.com/red-hat-storage/ocs-ci/issues/2256 or you would want me track this with new issue ?

petr-balogh: I am for open new issue with the title like:   `Fix issues after OCP 4.5 GA` and we can add it there as one of the bullet point.   I actually opened issue here:  https://github.com/red-hat-storage/ocs-ci/issues/2390  so you can add the link to this issue in the comment one line above this, so if someone will take a look in this config they will know the reason why in 4.5 config we have 4.4.

PR #2191, opened 2020-05-28
Load default ocp version conf if ocs version or registry image is provided

clacroix12:
Jira: [OCSQE-348](https://projects.engineering.redhat.com/browse/OCSQE-348)


conf/ocs_version/ocs-4.2.yaml
```diff
@@ -3,6 +3,7 @@ DEPLOYMENT:
   default_ocs_registry_image: "quay.io/rhceph-dev/ocs-olm-operator:4.2-rc"
   default_latest_tag: '4.2-rc'
   ocs_csv_channel: 'stable-4.2'
+  default_ocp_version: '4.2-ga'
```
petr-balogh, line 6: Not sure if we really want to go with GA version as default one. We should hit regression as soon as possible so I would go with nightlies anyway as default as we did till now.  Same for all below.

clacroix12: Done

Issue #2196, opened 2020-06-01
RHEL repo for 4.3 has changed


petr-balogh:
```07:03:29 - MainThread - ocs_ci.utility.utils - WARNING - Command stderr: https://mirror.openshift.com/enterprise/all/4.3/latest/RH7-RHAOS-4.3/x86_64/os/repodata/repomd.xml: [Errno 14] HTTPS Error 404 - Not Found```

Job:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/8271/console

I've just paused the job and verified that new repo is 
```
[openshift]
name=RHopenshift
baseurl=https://mirror.openshift.com/enterprise/all/4.3/latest/x86_64/os/
enabled=1
gpgcheck=0
sslverify=0
sslclientcert=/etc/pki/ca-trust/source/anchors/ops-mirror.pem```

PR #2197, opened 2020-06-01
Fix OCP 4.3 repo for RHEL

petr-balogh:
Fixes: #2196


petr-balogh:
Connecte to rhelpod and changed URL of repo to this new one and it was able to install all required packages.

Marking as verified, and after merge we can re-trigger the job

Issue #2233, opened 2020-06-02
fix get csv version for test_run_io


vasukulkarni:
Seeing following issue during test_run_io

```
IndexError: list index out of range
Stacktrace
self = <test_io_performance.TestIOPerformance object at 0x7f157e466470>
size = '1GB', io_direction = 'rw', jobs = 1, runtime = 60, depth = 4

    @pytest.mark.parametrize(
        argnames=[
            "size", "io_direction", "jobs", "runtime", "depth"
        ],
        argvalues=[
            pytest.param(
                *['1GB', 'rw', 1, 60, 4],
                marks=pytest.mark.polarion_id("OCS-676")
            ),
            pytest.param(
                *['1GB', 'rw', 6, 60, 16],
                marks=pytest.mark.polarion_id("OCS-677")
            ),
            pytest.param(
                *['1GB', 'rw', 12, 60, 32],
                marks=pytest.mark.polarion_id("OCS-678")
            ),
        ]
    )
    @pytest.mark.usefixtures(base_setup.__name__)
    def test_run_io(self, size, io_direction, jobs, runtime, depth):
        """
        Test IO
        """
        logging.info(
            f"Running FIO with:\nsize: {size}\njobs: {jobs}\n"
            f"runtime: {runtime}\nIO depth: {depth}\n"
        )
        self.pod_obj.run_io(
            'fs', size=size, io_direction=io_direction, jobs=jobs,
            runtime=runtime, depth=depth
        )
        logging.info("Waiting for results")
        fio_result = self.pod_obj.get_fio_results()
        logging.info("IOPs after FIO:")
        reads = fio_result.get('jobs')[0].get('read').get('iops')
        writes = fio_result.get('jobs')[0].get('write').get('iops')
        w_bw = fio_result.get('jobs')[0].get('write').get('bw')
        r_bw = fio_result.get('jobs')[0].get('read').get('bw')
        logging.info(f"Read: {reads}")
        logging.info(f"Write: {writes}")
    
>       push_perf_dashboard(self.interface, reads, writes, r_bw, w_bw)

tests/e2e/performance/test_io_performance.py:76: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ocs_ci/utility/performance_dashboard.py:77: in push_perf_dashboard
    data = initialize_data()
[...]

PR #2234, opened 2020-06-02
Fix get version for ocs to push to dashboard

vasukulkarni:
Fixes: https://github.com/red-hat-storage/ocs-ci/issues/2233

### DIFF
--- a/ocs_ci/utility/performance_dashboard.py
+++ b/ocs_ci/utility/performance_dashboard.py
@@ -10,16 +10,15 @@ environment: The platform we are testing. E.g, AWS
 result_value: The value of this benchmark.
 
 """
+import re
 import requests
 import json
-from datetime import datetime
-from ocs_ci.framework import config
-from ocs_ci.ocs import ocp, defaults
-from ocs_ci.ocs.node import get_typed_nodes
-from ocs_ci.ocs import constants
 
-now = datetime.now()
-date_time = now.strftime("%m.%d.%Y")
+from ocs_ci.ocs import constants
+from ocs_ci.framework import config
+from ocs_ci.ocs.node import get_typed_nodes
+from ocs_ci.ocs.version import get_ocs_version
+
 
 data_template = {
     "commitid": None,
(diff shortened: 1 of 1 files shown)


SvetlanaRedHat:
Trying to run https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/8445/ and deplow vmware cluuster, 
Previews run failed

ocs_ci/utility/performance_dashboard.py
```diff
@@ -39,21 +38,21 @@ def initialize_data():
...
-    csv = ocp.OCP(kind='csv', namespace=defaults.ROOK_CLUSTER_NAMESPACE)
     worker_type = get_typed_nodes(num_of_nodes=1)[0].data['metadata'][
         'labels'
     ]['beta.kubernetes.io/instance-type']
 
-    csv_vers = csv.get()['items'][0]['spec']['version'][:-3].split("-")
-    min_version = csv_vers[0]
-    build_id = csv_vers[1]
-    ocs_ver = ".".join(min_version.split(".")[:-1])
+    (ocs_ver_info, _) = get_ocs_version()
```
clacroix12, line 45: We don't need parenthesis here.

```suggestion
    ocs_ver_info, _ = get_ocs_version()
```

ocs_ci/utility/performance_dashboard.py
```diff
@@ -39,21 +38,21 @@ def initialize_data():
...
+    ocs_ver_full = ocs_ver_info['status']['desired']['version']
+    m = re.match(r"(\d.\d).(\d)-", ocs_ver_full)
+    if m.group(1) is not None:
+        ocs_ver = m.group(1)
```
clacroix12, line 49: If we do not enter this `if` block `ocs_ver` is never initialized and we will hit an error when using this variable below (lines 54 and 56). We should instead raise an error with a clear message if we are unable to initialize `ocs_ver`.

vasukulkarni: It is intentional so that the test can fail and we can check why the version is mismatched and take corrective steps if it fails.

ocs_ci/utility/performance_dashboard.py
vasukulkarni: @clacroix12 I am going to fix this in next version, want to unblock some testing for now.

ocs_ci/utility/performance_dashboard.py
clacroix12: I agree that we should fail, but we should raise an informative error message (referring to us being unable to retrieve the ocs version) instead of the user hitting `NameError ocs_ver is not defined` or something similar. This can be taken up in another PR as well, but I think it's worth doing.

PR #2243, opened 2020-06-03
Adding wait time after the restart/stop nodes

Akarsha-rai:


vavuthu:
@Akarsha-rai There are so many places where waiting_time is initialized. Do you want to take it as constant ?

Akarsha-rai:
> @Akarsha-rai There are so many places where waiting_time is initialized. Do you want to take it as constant ?

Removed waiting_time expect in one place instead added more retries.

PR #2249, opened 2020-06-04
DO NOT MERGE: Smallfile change log parsing

Avilir:
Changing the way results are pars. instead of reading results from elasticsearch (since it is not pushing results to ES when the test run in the public cloud), the results are parse from the log(s) file(s).
It also support running with more then one pod (for scaling).

One more change is the samples deviation results from 5% to 20% - this will changed when we will have hardware and network infra. that will run the test properly.

The test results are print in the log and also push to Internal elasticsearch server, in the future it will be push to codespeed server.

### DIFF
--- a/tests/e2e/performance/test_small_file_workload.py
+++ b/tests/e2e/performance/test_small_file_workload.py
@@ -250,6 +250,46 @@ class SmallFileResultsAnalyse(object):
         Aggregation results from all hosts in single sample
         (each host is actually a pod)
 
+        Example of log to parse :
+
+                                     version : 3.1
+                               hosts in test : None
+                            launch by daemon : False
+                       top test directory(s) : ['/mnt/pvc/smallfile_test_data']
+                                   operation : create
+                                files/thread : 50000
+                                     threads : 4
+               record size (KB, 0 = maximum) : 0
+                              file size (KB) : 4
+                      file size distribution : fixed
+                               files per dir : 100
+                                dirs per dir : 10
+                  threads share directories? : N
+                             filename prefix :
+                             filename suffix :
+                 hash file number into dir.? : N
+                         fsync after modify? : N
+              pause between files (microsec) : 50
+                 minimum directories per sec : 50
+                        finish all requests? : Y
+                                  stonewall? : Y
+                     measure response times? : Y
+                                verify read? : Y
+                                    verbose? : False
+                              log to stderr? : False
+            total threads = 4
+            total files = 200000
+            total IOPS = 12980
+            total data =     0.763 GiB
+            100.00% of requested files processed, warning threshold is  70.00
+            elapsed time =    17.079
+            files/sec = 12980.622480
+            IOPS = 12980.622480
+            MiB/sec = 50.705557
+            2020-05-04T13:50:01Z - INFO     - MainProcess - trigger_smallfile:
+                completed sample 1 for operation create , results in /var/tmp/RESULTS/1/create.json
+
+
         Args:
             host_res (dict): dictionary of host results - user for multi hosts
                              test
--- a/tests/e2e/performance/test_small_file_workload.py
+++ b/tests/e2e/performance/test_small_file_workload.py
@@ -307,14 +307,14 @@ class SmallFileResultsAnalyse(object):
         else:
 
             """
-            Each test can do some operations, so i am running loop on all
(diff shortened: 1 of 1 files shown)


tests/e2e/performance/test_small_file_workload.py
```diff
@@ -17,11 +17,138 @@
...
 from ocs_ci.ocs.exceptions import CommandFailed
-from elasticsearch import (Elasticsearch, exceptions as ESExp)
+from elasticsearch import (Elasticsearch, exceptions as esexp)
 import numpy as np
 
 log = logging.getLogger(__name__)
 
+"""
+Defining dictionary of keywords to look in the logs, foreach keyword we have the
+key-name in the final results output, the seperate charecter to split the line
```
fbalak, line 27:
```suggestion
key-name in the final results output, the separate character to split the line
```

tests/e2e/performance/test_small_file_workload.py
```diff
@@ -17,11 +17,138 @@
...
+    'total data': {
+        'name': 'Total_Data_GB', 'sep': ' ', 'pos': -2, 'op': np.sum
+    },
+}
+
+
+class SmallFileLogParser(object):
+
+    """
+    This class is reading the benchmark pod logs, and pars it
```
fbalak, line 67:
```suggestion
    This class is reading the benchmark pod logs, and parse it
```

tests/e2e/performance/test_small_file_workload.py
```diff
@@ -105,152 +235,51 @@ def add_key(self, key, value):
...
-
-        results = {}
-
-        for op in self.results['operations']:
-            for smp in range(self.results['samples']):
-                sample = smp + 1
-                if op in self.results['full-res'].keys():
-                    self.results['full-res'][op][sample] = self.combine_results(
-                        self.results['full-res'][op][sample], True)
+            for op in self.results['operations']:
```
fbalak, line 271: Can you please add docstring explaining the process here?

tests/e2e/performance/test_small_file_workload.py
Avilir: done.

tests/e2e/performance/test_small_file_workload.py
Avilir: done

tests/e2e/performance/test_small_file_workload.py
Avilir: done

tests/e2e/performance/test_small_file_workload.py
```diff
@@ -424,64 +439,93 @@ def test_smallfile_workload(self, ripsaw, file_size, files, threads,
...
-                if bench_pod[0] is not None:
-                    small_file_client_pod = bench_pod[0]
-                    break
-            except IndexError:
-                log.info("Bench pod not ready yet")
+        clients = sf_data['spec']['workload']['args']['clients']
+        log.info(f'Going to run on {clients} pods, waiting for pods to start')
+        count = 10  # total number of retry for all benchmark pods to start
+        small_file_client_pods = {}
+        while count > 0:
```
vasukulkarni, line 462: do we need the count, how about increasing the values in timeout sampler? will that help?

tests/e2e/performance/test_small_file_workload.py
```diff
@@ -105,152 +235,77 @@ def add_key(self, key, value):
...
+            operations that we have in this particular test.
 
-        Returns:
-            dict : dictionary of results records
+            """
+            for op in self.results['operations']:
 
-        """
+                """
+                For each operation, I am looking for all interesting lines
```
vasukulkarni, line 281: I think better to add small snipet of items we are parsing here.

tests/e2e/performance/test_small_file_workload.py
Avilir: changed the timeout, instead using the count.

tests/e2e/performance/test_small_file_workload.py
Avilir: added example of the log file.

Avilir:
I am working on different approach, so, pleas wait with this PR

Avilir:
replaced by PR #2397

Issue #2253, opened 2020-06-04
Default to OCS 4.5


clacroix12:
Now that 4.4 is released we want to default our deployments to 4.5.

PR #2254, opened 2020-06-04
Default OCP and OCS versions to 4.5

clacroix12:
Fixes: #2253 

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -120,7 +120,7 @@ ENV_DATA:
 # This section is related to upgrade
 UPGRADE:
   upgrade_to_latest: true
-  ocp_channel: "stable-4.5"
+  ocp_channel: "candidate-4.5"
   ocp_upgrade_path: "quay.io/openshift-release-dev/ocp-release"
   ocp_arch: "x86_64"
 
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -19,7 +19,7 @@ RUN:
   cli_params: {}  # this will be filled with CLI parameters data
   # If the client version ends with .nightly, the version will be exposed
   # to the latest accepted OCP nightly build version
-  client_version: '4.4.0-0.nightly'
+  client_version: '4.5.0-0.nightly'
   bin_dir: './bin'
   google_api_secret: '~/.ocs-ci/google_api_secret.json'
   rook_branch: "master"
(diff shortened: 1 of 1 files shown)


ocs_ci/framework/conf/default_config.yaml
```diff
@@ -120,7 +120,7 @@ ENV_DATA:
 # This section is related to upgrade
 UPGRADE:
   upgrade_to_latest: true
-  ocp_channel: "stable-4.4"
+  ocp_channel: "stable-4.5"
```
petr-balogh, line 123: I don't think such channel exists when OCP 4.5 is not GAed yet.

@AviadP can tell you more details what to use but I guess we should use something similar what we had here:
https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocp_version/ocp-4.4-candidate-upgrade.yaml 

Most likely there will be already candidate build for 4.5 but haven't checked here.

Other option to use nightly like in this config:
https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocp_version/ocp-4.4-upgrade.yaml

Which will be probably better option, but for this you will need to change also one line below.

petr-balogh: Please check the code how it is used or wait for @AviadP explanation. I do not remember it either. Just brought this up for discussion as we need to set it correctly for OCP upgrade.

AviadP: @petr-balogh, channel "cnadidate-4.5" should be ready for use. as far as i understand, "stable-4.5" isnt
i will create config files for 4.5-nightly or 4.5 candidate

clacroix12: `ocp_channel` is now set to `candidate-4.5`.

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -101,12 +101,12 @@ ENV_DATA:
...
   skip_ocp_deployment: false
   skip_ocs_deployment: false
-  ocs_version: '4.4'
+  ocs_version: '4.5'
   # uncomment to use custom directory for storing measurement data related to
   # monitoring tests, otherwise generate temporary directory for each test run
   # measurement_dir: '/tmp/ocs_ci_monitoring_measurement/'
   # Default RHCOS image to be used for VmWare deployment
-  vm_template: 'rhcos-4.4.0-rc.1-x86_64-vmware.x86_64'
+  vm_template: 'rhcos-4.5.0-rc.1-x86_64-vmware.x86_64'
```
vavuthu, line 109:
```suggestion
  vm_template: ' rhcos-4.5.0-0.nightly-2020-03-18-115438-x86_64-vmware.x86_64'
```

clacroix12: @vavuthu is the space at the start of the template string necessary or a typo?

vasukulkarni: i think just a typo

clacroix12: I updated the vm_template without the space at the start, assuming it was a typo.

vavuthu: its typo

PR #2255, opened 2020-06-05
Export all pgbench pods run outputs to google spreadsheet

PrasadDesala:

### DIFF
--- a/ocs_ci/ocs/pgsql.py
+++ b/ocs_ci/ocs/pgsql.py
@@ -412,15 +412,21 @@ class Postgresql(RipSaw):
             sheet_name=sheet_name, sheet_index=sheet_index
         )
         log.info("Exporting pgoutput data to google spreadsheet")
-        for i in range(len(pg_output)):
-            for j in range(len(pg_output[i][0])):
-                run_id = list(pg_output[i][0][j].keys())[0]
-                lat_avg = pg_output[i][0][j][run_id]['latency_avg']
-                lat_stddev = pg_output[i][0][j][run_id]['lat_stddev']
-                tps_incl = pg_output[i][0][j][run_id]['lat_stddev']
-                tps_excl = pg_output[i][0][j][run_id]['tps_excl']
+        for pgbench_pod in range(len(pg_output)):
+            for run in range(len(pg_output[pgbench_pod][0])):
+                run_id = list(pg_output[pgbench_pod][0][run].keys())[0]
+                lat_avg = pg_output[
+                    pgbench_pod
+                ][0][run][run_id]['latency_avg']
+                lat_stddev = pg_output[
+                    pgbench_pod
+                ][0][run][run_id]['lat_stddev']
+                tps_incl = pg_output[
+                    pgbench_pod
+                ][0][run][run_id]['lat_stddev']
+                tps_excl = pg_output[pgbench_pod][0][run][run_id]['tps_excl']
                 g_sheet.insert_row(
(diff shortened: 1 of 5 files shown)


PrasadDesala:
Attached output screenshot of collected data in spreadsheet:
pgbench_spreadsheet

ocs_ci/ocs/pgsql.py
```diff
@@ -338,6 +340,44 @@ def respin_pgsql_app_pod(self):
...
+        Collect pgbench output to google spreadsheet
+
+        Args:
+            pg_output (list):  pgbench outputs in list
+            sheet_name (str): Name of the sheet
+            sheet_index (int): Index of sheet
+
+        """
+        # Collect data and export to Google doc spreadsheet
+        g_sheet = GoogleSpreadSheetAPI(
```
hnallurv, line 354: How are we going to store the results of multiple runs over multiple builds/releases? Is it:
1) one sheet per release with many tabs - each tab for a particular run
 - looks compact
2) one sheet for all the releases with many tabs - each tab for a particular run
 - will have have too many tabs
3) one sheet for each run?
 - will create too many individual files - difficult to browse through and compare
4) any other option?

ocs_ci/ocs/pgsql.py
```diff
@@ -338,6 +340,44 @@ def respin_pgsql_app_pod(self):
...
+            sheet_name='E2E Workloads - PGSQL', sheet_index=sheet_index
+        )
+        log.info("Exporting pgoutput data to google spreadsheet")
+        for i in range(len(pg_output)):
```
hnallurv, line 358: I suggest following:
1) All details like versions of OCP, OCS, ceph etc., to be shown in table format
2) have pgbench pod name in first column and rest of the columns like lat_avg etc., from 2nd column

ocs_ci/ocs/pgsql.py
PrasadDesala: In the spreadsheet "E2E Workloads - PGSQL", I am maintaining individual tabs for each pgsql test. For every run the results of each test gets appended to its own tab. 

Below is the snap of pg_out data of test_pgsql.py for two runs,
multiple_runs

We can collect this way for every run in the interim builds and once we have the final benchmark results we can move the final run details to a new tab and lock(this has to be done manually).

PrasadDesala: I have updated the patch to only export test_pgsql.

ocs_ci/ocs/pgsql.py
PrasadDesala: Resolved in the new commit

PrasadDesala:
pgbench_data

ocs_ci/ocs/pgsql.py
```diff
@@ -366,21 +378,66 @@ def get_pgbech_pod_status_table(self, pgbench_pods):
...
+
+        """
+        # Collect data and export to Google doc spreadsheet
+        g_sheet = GoogleSpreadSheetAPI(
+            sheet_name=sheet_name, sheet_index=sheet_index
+        )
+        log.info("Exporting pgoutput data to google spreadsheet")
+        for i in range(len(pg_output)):
+            for j in range(len(pg_output[i][0])):
+                run_id = list(pg_output[i][0][j].keys())[0]
```
OdedViner, line 417: Can use different variable names? not (i,j)

PrasadDesala: Done

PrasadDesala:
new

PR #2261, opened 2020-06-07
Add capacity when one of the osd pods gets deleted

yitzhak12:


tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
```diff
@@ -0,0 +1,84 @@
...
+class TestAddCapacityWithOSDPodDelete(ManageTest):
+    """
+    Test add capacity when one of the osd pods gets deleted
+    in the middle of the process.
+    """
+
+    def test_add_capacity_osd_pod_delete(
+        self, nodes, multi_pvc_factory, pvc_factory, pod_factory, workload_storageutilization_rbd
+    ):
+        """ Test add capacity when one of the osd pods gets deleted
```
ebenahar, line 37: move the text to a new line

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
```diff
@@ -0,0 +1,84 @@
...
+        in the middle of the process.
+        """
+        logging.info("Condition 1 to start the test is met: storageutilization is completed")
+        # Please notice: When the branch 'wip-add-capacity-e_e' will be merged into master
+        # the test will include more much data both before, and after calling 'add_capacity'function.
+
+        max_osds = 12
```
ebenahar, line 44: why 12?

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
```diff
@@ -0,0 +1,84 @@
...
+        osd_pods_before = pod_helpers.get_osd_pods()
+        assert len(osd_pods_before) < max_osds, (
+            "Condition 2 to start test failed: We have maximum of osd's in the cluster")
+        logging.info("All start conditions are met!")
+
+        logging.info("Perform some IO operations...")
+        cluster_load = ClusterLoad()
+        cluster_load.reach_cluster_load_percentage_in_throughput(pvc_factory, pod_factory, target_percentage=0.3)
```
ebenahar, line 52: better not to use this in the test itself. Should be used from https://github.com/red-hat-storage/ocs-ci/blob/3e6e6121fdb8e92e0ccae3aaa62a259b037b35f2/tests/conftest.py#L1044

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
yitzhak12: I think we can change it to 15.

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
yitzhak12: fixed

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
yitzhak12: fixed

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
```diff
@@ -0,0 +1,74 @@
...
+    Test add capacity when one of the osd pods gets deleted
+    in the middle of the process.
+    """
+
+    def test_add_capacity_osd_pod_delete(self, workload_storageutilization_rbd):
+        """
+        Test add capacity when one of the osd pods gets deleted
+        in the middle of the process.
+        """
+        logging.info("Condition 1 to start the test is met: storageutilization is completed")
```
ebenahar, line 38: is there any variable to print here? maybe the storage utilization of the cluster?

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
```diff
@@ -0,0 +1,74 @@
...
+
+        max_osds = 15
+        osd_pods_before = pod_helpers.get_osd_pods()
+        assert len(osd_pods_before) < max_osds, (
+            "Condition 2 to start test failed: We have maximum of osd's in the cluster")
+        logging.info("All start conditions are met!")
```
ebenahar, line 44: the test should not fail in this case. Better to skip

yitzhak12: Yes, that makes sense.

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
yitzhak12: Added the used capacity

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
```diff
@@ -0,0 +1,75 @@
...
+        d = Disruptions()
+        d.set_resource('osd')
+
+        osd_size = storage_cluster.get_osd_size()
+        logging.info("Calling add_capacity function...")
+        result = storage_cluster.add_capacity(osd_size)
+        if result:
+            logging.info("add capacity finished successfully")
+        else:
+            logging.info("add capacity failed")
```
mbukatov, line 56: Return value from `storage_cluster.add_capacity(osd_size)` seems to be an nonzero integer, so this check will always go with "finished successfully" branch. The doc string of the function is (sadly) not aligned with the code.

To detect a failure, it seems you need to catch some exception.

mbukatov, line 56: Here we info log message about a failure. I would suggest to log errors at least with warning log level, so that a possible issue is easier to see in the logs.

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
```diff
@@ -0,0 +1,75 @@
...
+            pytest.skip("We have maximum of osd's in the cluster")
+
+        d = Disruptions()
+        d.set_resource('osd')
+
+        osd_size = storage_cluster.get_osd_size()
+        logging.info("Calling add_capacity function...")
+        result = storage_cluster.add_capacity(osd_size)
+        if result:
+            logging.info("add capacity finished successfully")
```
mbukatov, line 54: When `storage_cluster.add_capacity(osd_size)` finishes, the add capacity procedure just started, it's unlikely it will be finished immediately. Since this is done on purpose here, let's  update the log message.

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
```diff
@@ -0,0 +1,75 @@
...
+        else:
+            logging.info("add capacity failed")
+
+        logging.info("Delete osd resource")
+        d.delete_resource(1)
+
+        pod = OCP(
+            kind=constants.POD, namespace=config.ENV_DATA['cluster_namespace']
+        )
+        pod.wait_for_resource(
```
mbukatov, line 64: I would suggest to place an info log message about what we are actually going to wait for.

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
```diff
@@ -0,0 +1,75 @@
...
+        osd_size = storage_cluster.get_osd_size()
+        logging.info("Calling add_capacity function...")
+        result = storage_cluster.add_capacity(osd_size)
+        if result:
+            logging.info("add capacity finished successfully")
+        else:
+            logging.info("add capacity failed")
+
+        logging.info("Delete osd resource")
+        d.delete_resource(1)
```
mbukatov, line 59: Do you really want to delete the OSD immediately after asking OCS to add capacity (OSD)? I'm asking because it's not clear at which state of adding OSD would OCS cluster be at this point.

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
mbukatov: Is it possible that the test case goes through this wait without OCS to finish neither removal or addition of OSDs?

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
yitzhak12: from https://polarion.engineering.redhat.com/polarion/#/project/OpenShiftContainerStorage/workitem?id=OCS-1203: "Delete an OSD pod while the storage capacity is getting increased". Maybe we need to wait a little bit, but it seems to me that this is the purpose of the test.

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
yitzhak12: Yes, we need to change the docstring of this function. I think in this case I will just log a message with something like "finished executing 'add_capacity' function".

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
```diff
@@ -0,0 +1,76 @@
...
+        if len(osd_pods_before) >= max_osds:
+            pytest.skip("We have maximum of osd's in the cluster")
+
+        d = Disruptions()
+        d.set_resource('osd')
+
+        osd_size = storage_cluster.get_osd_size()
+        logging.info("Calling add_capacity function...")
+        result = storage_cluster.add_capacity(osd_size)
+        logging.info("finished executing 'add_capacity' function")
```
mbukatov, line 54: I would suggest to name these variables and tweak logging messages in more descriptive, high level way so that it's more obvious what is going on, or at least so that one can match your intentions with what is actually happening:

```
capacity_increment = storage_cluster.get_osd_size()
logging.info("Adding few OSDs to increase user visible capacity by %d UNIT??", capacity_increment)
storagedeviceset_count = storage_cluster.add_capacity(capacity_increment)
logging.info("request to increase capacity was issued without problems")
```

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
mbukatov: Ok.

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
```diff
@@ -0,0 +1,79 @@
...
+
+        osd_size = storage_cluster.get_osd_size()
+        logging.info(f"Adding one new set of OSDs. osd size = {osd_size}")
+        storagedeviceset_count = storage_cluster.add_capacity(osd_size)
+        logging.info("Adding one new set of OSDs was issued without problems")
+
+        # OSD number go down by one and then gradually go up by 1
+        # and finally the OSD number will be storagedeviceset_count*3
+        time_to_wait_before_delete_osd_pod = 20
+        time.sleep(time_to_wait_before_delete_osd_pod)
```
ebenahar, line 60: is there anything we can sample and wait for instead of waiting 20 seconds?

yitzhak12: I thought to check whether the first new osd pod is in running state, but the process happens really fast, and there is a chance that all the 3 new osd pods will be in running state before I had the chance to delete an osd pod.

ebenahar: so in that case, doesn't 20 seconds sleep mean that you could miss the opportunity to delete one of the OSDs before all reach Running? It's not deterministic

ocs_ci/ocs/resources/pod.py
```diff
@@ -1360,3 +1360,23 @@ def delete_deploymentconfig_pods(pod_obj):
                 dc_ocp_obj.wait_for_delete(
                     resource_name=pod_obj.get_labels().get('name')
                 )
+
+
+def wait_for_new_osd_pods_to_come_up(number_of_osd_pods_before):
```
ebenahar, line 1365: very nice!!

ocs_ci/ocs/resources/pod.py
```diff
@@ -1360,3 +1360,23 @@ def delete_deploymentconfig_pods(pod_obj):
...
+    status_options = ['Init:1/4', 'Init:2/4', 'Init:3/4', 'PodInitializing', 'Running']
+    try:
+        for osd_pods in TimeoutSampler(
+            timeout=180, sleep=3, func=get_osd_pods
```
ebenahar, line 1369: maybe sample more frequently for better chances not to miss 1 OSD coming up before all do? how about every 2 seconds instead of 3?

yitzhak12: I think 3 seconds is the minimum, cause it takes like 1-2 seconds to execute the 'oc' command

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
yitzhak12: Fixed

tests/manage/z_cluster/cluster_expansion/test_delete_osd_pod.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+
+        pod.wait_for_resource(
+            timeout=420,
+            condition=constants.STATUS_RUNNING,
+            selector='app=rook-ceph-osd',
+            resource_count=storagedeviceset_count * 3
+        )
+
+        logging.info("Finished verifying add capacity osd storage with node restart")
```
sidhant-agrawal, line 74: Is the log message correct?

yitzhak12: Yes, right. I need to change the log message.

PR #2263, opened 2020-06-08
Flexy Integration and PSIBaremetal deployment:

shylesh:
1. Flexy integration with OCS-CI
2. BaremetalPSI deployment with Flexy
3. PSI openstack libraries


shylesh:
Rebased on latest master and resolved conflicts

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+        config_parser.read_string(file_content)
+        # Iterate over config_parser keys, if same key is present
+        # in user supplied dict update config_parser
+        for ele in config_parser.items('root'):
+            if ele[0] in config.ENV_DATA:
+                # For LAUNCHER_VARS we need to merge the
+                # user provided dict with default obtained
+                # from env file
+                if ele[0] == 'LAUNCHER_VARS':
+                    merge_dict(config.ENV_DATA[ele[0]], ele[1])
```
dahorak, line 201: If I understand it correctly, the configuration of `LAUNCHER_VARS` in *ocs-osp.env* have precedence here before the onfiguration in the `config.ENV_DATA`, right? In other words if I'll provide my own dict in `config.ENV_DATA['LAUNCHER_VARS']`, existing keys in the `LAUNCHER_VARS` variable in *ocs-osp.env* file will not be updated?
If the above statement is correct, wouldn't be better to make it vice versa - give higher priority to the configuration provided by user in the `config.ENV_DATA['LAUNCHER_VARS']`?

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+
+        """
+        config_parser = configparser.ConfigParser()
+        # without this,config parser would convert all
+        # uppercase keys to lowercase
+        config_parser.optionxform = str
+        env_file = os.path.join(
+            self.flexy_host_private_conf_dir_path, constants.FLEXY_DEFAULT_ENV_FILE
+        )
+
```
dahorak, line 185: Couldn't this be omitted in favour of `self.flex_env_file` configured in `__init__()`?

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+                # from env file
+                if ele[0] == 'LAUNCHER_VARS':
+                    merge_dict(config.ENV_DATA[ele[0]], ele[1])
+                config_parser['root'][ele[0]] = config.ENV_DATA[ele[0]]
+                logger.info(f"env updated {ele[0]}:{config.ENV_DATA[ele[0]]}")
+
+        # write the updated config_parser content back to the env file
+        tmp_file = os.path.join(
+            self.flexy_host_private_conf_dir_path, f"{constants.FLEXY_DEFAULT_ENV_FILE}.tmp"
+        )
```
dahorak, line 208: Similarly to the above, couldn't this be simplified with `self.flex_env_file` configured in `__init__()`?

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+        self.flexy_openshift_misc_url = config.ENV_DATA.get(
+            'flexy_openshift_misc_url', constants.OCP_QE_MISC_REPO
+        )
+        self.flexy_private_conf_url = config.ENV_DATA.get(
+            'flexy_private_conf_url', constants.FLEXY_PRIVATE_CONF_URL
+        )
+        self.flexy_img_url = config.ENV_DATA.get('flexy_img_url')
+        self.flex_env_file = os.path.join(
+            self.flexy_host_private_conf_dir_path, constants.FLEXY_DEFAULT_ENV_FILE
+        )
```
dahorak, line 57: I think it might be worth to make the env file configurable and use `constants.FLEXY_DEFAULT_ENV_FILE` (pointing to `ocs-osp.env`) only as default value. For some configurations it might be worth to prepare new env file, instead of update all the required keys through another configuration file, WDYT?

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+            'flexy_mnt_private_conf_dir', constants.FLEXY_MNT_PRIVATE_CONF_DIR
+        )
+        self.flexy_openshift_misc_url = config.ENV_DATA.get(
+            'flexy_openshift_misc_url', constants.OCP_QE_MISC_REPO
+        )
+        self.flexy_private_conf_url = config.ENV_DATA.get(
+            'flexy_private_conf_url', constants.FLEXY_PRIVATE_CONF_URL
+        )
+        self.flexy_img_url = config.ENV_DATA.get('flexy_img_url')
+        self.flex_env_file = os.path.join(
```
dahorak, line 55:
```suggestion
        self.flexy_env_file = os.path.join(
```
`flex_env_file` > `flexy_env_file` (similarly on other places with this property)

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+            # we need to add section to make
+            # config parser happy, this will be removed
+            # while writing back to file
+            file_content = "[root]\n" + fp.read()
+
+        config_parser.read_string(file_content)
+        # Iterate over config_parser keys, if same key is present
+        # in user supplied dict update config_parser
+        for ele in config_parser.items('root'):
+            if ele[0] in config.ENV_DATA:
```
dahorak, line 196: Wouldn't be better to provide new section in the configuration, like `FLEXY`, instead of mixing those parameters used to update the flexy env file with other variables in *ENV_DATA* section?

ocs_ci/utility/psiutils.py
```diff
@@ -0,0 +1,151 @@
...
+            if v.status == 'in-use':
+                v.detach()
+                v.get()
+                if v.status == 'available':
+                    logger.info(f"Volume {v.name} detached successfully ")
+                else:
+                    for res in TimeoutSampler(60, 1, v.status == 'available'):
+                        if not res:
+                            logger.info(f"Waiting for {v.name} to detach")
+                            v.get()
```
dahorak, line 146: The message `Volume {v.name} detached successfully ` will be printed only when the volume will be detached immediately - when the detach process will take few seconds, it will not print that message even when the volume will be detached correctly before timeout. Or did I miss something?

ocs_ci/utility/psiutils.py
```diff
@@ -0,0 +1,151 @@
...
+            v.delete()
+            try:
+                v.get()
+            except Exception:
+                logger.info(f"Volume {v.name} deleted successfully")
```
dahorak, line 151: It will be better to handle one particular exception there, this general `except Exception` might hide some other issues (e.g. API outage,...).
And what if the volume will not be deleted correctly or if it will some time?

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+            if self.exitThread:
+                break
+        logger.info("Logger thread done printing flexy logs")
+        # exit only this thread
+        sys.exit()
+
+    def build_container_args(self):
+        """
+        Builds most commonly used arguments for flexy container
+
```
petr-balogh, line 134:
```
Returns:
    list: ....
```

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+                logger.info("Waiting for logger thread to join")
+
+    def build_install_cmd(self):
+        """
+        Build flexy command line for 'deploy' operation
+
+        """
+        cmd = "sudo podman run --rm=true"
+        flexy_container_args = self.build_container_args()
+        cmd_string = " ".join([cmd, " ".join(flexy_container_args)])
```
petr-balogh, line 97: You can use this from subprocess:  
from subprocess import list2cmdline
```suggestion
        cmd = shlex.split("sudo podman run --rm=true")
        flexy_container_args = self.build_container_args()
        cmd_string = list2cmdline(cmd + flexy_container_args)
```

This deal with few issues in string arguments and so on and you are reusing the functionality which is already implemented for this in subprocess module.

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+        return cmd_string
+
+    def build_destroy_cmd(self):
+        """
+        Build flexy command line for 'destroy' operation
+
+        """
+        cmd = "sudo podman run --rm=true"
+        flexy_container_args = self.build_container_args()
+        cmd_string = " ".join([cmd, " ".join(flexy_container_args), 'destroy'])
```
petr-balogh, line 107: same

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+        return cmd_string
+
+    def get_log(self):
```
petr-balogh, line 110: Just curious as I remember we had this kind of PR opened for long time : https://github.com/red-hat-storage/ocs-ci/pull/1096/files

This could actually solve the whole issue you would like to achieve here with live log right?

But the way how it's implemented in the PR looks more clean to me.

ocs_ci/utility/psiutils.py
```diff
@@ -0,0 +1,151 @@
+"""
+A module for all PSI-openstack related utilities
+"""
+
+import logging
+
+from keystoneauth1 import loading
+from keystoneauth1 import session
```
clacroix12, line 8: consolidate these lines

ocs_ci/utility/psiutils.py
```diff
@@ -0,0 +1,151 @@
...
+        A function to create openstack volumes
+
+        Args:
+            name (str): display name of the volume
+            size (int): of the volume in GBs
+            volume_type (str): type of volume to be created
+            metadata (dict): Any {k,v} to be associated with volume
+
+        Returns:
+            Instance of "Volume" class
```
clacroix12, line 59: You may have to import but this should show the actual type

```suggestion
            Volume: PSI openstack volume
```

ocs_ci/utility/psiutils.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        matching_vols = []
+        for vol in self.cinder_clnt.volumes.list():
+            for k, v in tag.items():
+                if vol.metadata.get(k) == v:
+                    matching_vols.append(vol)
+
+        return matching_vols
+
+    def detach_and_delete_vols(self, vollist):
```
clacroix12, line 128: Can we rename `vollist` to `volumes`?

ocs_ci/deployment/flexy.py
shylesh: oops, I missed it, good catch. I will change that.

ocs_ci/deployment/flexy.py
shylesh: ok, this scenario I didn't thought about. We can make it configurable. I will try this one, because if we pass env file in that case we might have to skip cloning of the ocs-private repo all together. I will update you how it works out.

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+        )
+        self.flexy_private_conf_url = config.ENV_DATA.get(
+            'flexy_private_conf_url', constants.FLEXY_PRIVATE_CONF_URL
+        )
+        self.flexy_img_url = config.ENV_DATA.get('flexy_img_url')
+        self.flex_env_file = os.path.join(
+            self.flexy_host_private_conf_dir_path, constants.FLEXY_DEFAULT_ENV_FILE
+        )
+        self.flexy_log_file = os.path.expanduser(
+            config.ENV_DATA.get('flexy_log_file')
```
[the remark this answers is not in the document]
shylesh: We can make default path for this, I will make it as contant.

ocs_ci/deployment/flexy.py
shylesh: ack, I will do it.

ocs_ci/deployment/flexy.py
shylesh: ack

ocs_ci/deployment/flexy.py
shylesh: Agree, that PR looks neater. I am not really familiar with ```select``` call, I will try to come up with this.

ocs_ci/deployment/flexy.py
shylesh: ack

ocs_ci/deployment/flexy.py
shylesh: Sorry, I missed to notice that. I will change it.

ocs_ci/deployment/flexy.py
shylesh: ack, I am fine with FLEXY section.

ocs_ci/deployment/flexy.py
shylesh: You are right, I visualized it in wrong way, I will change that.

ocs_ci/deployment/flexy.py
shylesh: ack

ocs_ci/ocs/constants.py
```diff
@@ -832,6 +832,24 @@
...
+FLEXY_DEFAULT_ENV_FILE = "ocs-osp.env"
+OPENSHIFT_MISC_BASE = "private-openshift-misc/v3-launch_templates/functionality-testing"
+FLEXY_BAREMETAL_UPI_TEMPLATE = "upi-on-baremetal/versioned-installer-openstack"
+FLEXY_PRIVATE_CONF_URL = "git@gitlab.cee.redhat.com:ocs/flexy-ocs-private.git"
+FLEXY_GIT_CRYPT_KEYFILE = "~/git_crypt_keyfile"
+NTP_CHRONY_CONF = os.path.join(
+    TEMPLATE_DIR, "ocp-deployment", "ntp_chrony.yaml"
+)
+FLEXY_SERVICE_CONF = os.path.join(
+    DATA_DIR, "flexy_service_config.conf"
```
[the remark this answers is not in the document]
shylesh: It is yaml, this has PSI auth info and all, I will move it to auth.yaml.

ocs_ci/utility/psiutils.py
shylesh: ack

ocs_ci/utility/psiutils.py
shylesh: ack

ocs_ci/utility/psiutils.py
shylesh: ack

ocs_ci/utility/psiutils.py
shylesh: Agree, I was bit confused with HTTP kind of message thrown instead of direct python exception. I will work on this.

ocs_ci/utility/psiutils.py
shylesh: ack

ocs_ci/deployment/baremetal.py
```diff
@@ -11,3 +21,90 @@ class BAREMETALUPI:
...
+            """
+            super().deploy_prereq()
+            self.flexy_instance.deploy_prereq()
+
+        def deploy(self):
+            self.flexy_instance.deploy()
+            # Point cluster_path to cluster-dir created by flexy
+            abs_cluster_path = os.path.abspath(self.cluster_path)
+            flexy_cluster_path = os.path.join(
+                self.flexy_instance.flexy_mnt_host_dir,
```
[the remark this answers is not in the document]
shylesh: This was fun!!
Actually  having flexy directory on ~/current-cluster-dir is a problem, container will not be able to start because it tries to do some modifications for flexy dir on nfs share and the op fails on nfs.

I tried having flexy dir on Jenkins slave's local dir and making ~/current-cluster-dir/openshift-cluster-dir as a softlink to this dir and container works fine and it installed.

Only think I need to check is ,this soft link doesn't cause any problem for OCS deployment.

@petr-balogh , Are you ok with this approach ?

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+        )
+        self.exitThread = False
+
+    def run_container(self, cmd_string):
+        """
+        Actualy container run happens here, a thread will be
```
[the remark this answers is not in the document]
shylesh: ack

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+        spawned to asynchronously print flexy container logs
+
```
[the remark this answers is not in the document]
shylesh: ack

ocs_ci/deployment/ocp.py
```diff
@@ -26,7 +26,8 @@ def __init__(self):
         self.metadata = {}
         self.deployment_platform = config.ENV_DATA['platform'].lower()
         self.deployment_type = config.ENV_DATA['deployment_type'].lower()
-        self.installer = self.download_installer()
+        if self.deployment_platform != 'baremetalpsi':
```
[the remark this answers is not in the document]
shylesh: ack

ocs_ci/deployment/ocp.py
```diff
@@ -112,7 +113,8 @@ def deploy_prereq(self):
                 "stored at: %s",
                 self.cluster_path
             )
-        self.create_config()
+        if self.deployment_platform != 'baremetalpsi':
```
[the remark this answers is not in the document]
shylesh: ack

ocs_ci/utility/psiutils.py
```diff
@@ -0,0 +1,151 @@
...
+            )
+
+        return vol
+
+    def attach_volume(self, vol, instance_id, mnt_pnt):
+        """
+        Attach the given volume to the specific PSI openstack instance
+
+        Args:
+            vol_id (Volume): cinder volume object
```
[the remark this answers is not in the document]
shylesh: ack

ocs_ci/utility/psiutils.py
```diff
@@ -0,0 +1,151 @@
...
+                    matching_vols.append(vol)
+
+        return matching_vols
+
+    def detach_and_delete_vols(self, vollist):
+        """
+        Detach and delete volumes from the list
+
+        Args:
+            list: of Volume objects
```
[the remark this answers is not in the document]
shylesh: ack

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+        Clone ocs_private_conf (flexy env and config) repo into
+        flexy_mnt_host_dir
+
+        """
+        clone_repo(self.flexy_private_conf_url, self.flexy_host_private_conf_dir_path)
+        # git-crypt unlock /path/to/keyfile
+        old_cwd = os.getcwd()
+        os.chdir(self.flexy_host_private_conf_dir_path)
+        cp = subprocess.run(
+            f'git-crypt unlock {constants.FLEXY_GIT_CRYPT_KEYFILE}',
```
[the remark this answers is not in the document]
shylesh: ack

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+    def clone_and_unlock_ocs_private_conf(self):
+        """
+        Clone ocs_private_conf (flexy env and config) repo into
+        flexy_mnt_host_dir
+
+        """
+        clone_repo(self.flexy_private_conf_url, self.flexy_host_private_conf_dir_path)
+        # git-crypt unlock /path/to/keyfile
+        old_cwd = os.getcwd()
+        os.chdir(self.flexy_host_private_conf_dir_path)
```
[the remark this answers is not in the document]
shylesh: ack

ocs_ci/utility/psiutils.py
```diff
@@ -0,0 +1,151 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class PSIUtils(object):
+    """
+    A class for handling PSI functionalities
+    """
+    def __init__(self, psi_conf):
+        self.auth_url = psi_conf.get('auth_url')
```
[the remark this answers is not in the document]
shylesh: ack

ocs_ci/utility/psiutils.py
```diff
@@ -0,0 +1,151 @@
+"""
+A module for all PSI-openstack related utilities
+"""
+
+import logging
+
+from keystoneauth1 import loading
```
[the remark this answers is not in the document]
shylesh: ack

ocs_ci/utility/psiutils.py
```diff
@@ -0,0 +1,151 @@
...
+
+        self.loader = loading.get_plugin_loader('password')
+        self.auth = self.loader.load_from_options(
+            auth_url=self.auth_url,
+            username=self.username,
+            password=self.password,
+            project_id=self.project_id,
+            user_domain_name=self.user_domain_name
+        )
+        self.sess = session.Session(auth=self.auth)
```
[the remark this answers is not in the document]
shylesh: ack, I ll check this one.

ocs_ci/deployment/baremetal.py
```diff
@@ -11,3 +21,90 @@ class BAREMETALUPI:
...
+            # Get all instances and for each instance add
+            # one disk
+            pattern = config.ENV_DATA['cluster_name']
+            for instance in self.utils.get_instances_with_pattern(pattern):
+                vol = self.utils.create_volume(
+                    name=f'disk0',
+                    size=config.ENV_DATA['volume_size'],
+                )
+                # wait till volume is available
+                for res in TimeoutSampler(300, 1, vol.status == "available"):
```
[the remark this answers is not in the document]
shylesh: ack

ocs_ci/deployment/baremetal.py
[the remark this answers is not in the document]
dahorak: What about cloning the private repo to `ocs-ci/data` directory? (There is already `rook`...)

Having `~/current-cluster-dir/openshift-cluster-dir` softlink  pointing to the clusterdir generated by flexy is not ideal, because it will not be backed up on the NFS. From my POV, it will be better to have the `flexy` directory (somewhere) in `data` directory of the `ocs-ci` repo directory and prospectively copy all the content somewhere to the defined cluster dir.

shylesh:
rebased on latest master

ocs_ci/deployment/flexy.py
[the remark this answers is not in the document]
petr-balogh: after discussion with @dahorak we were thinking about approach to use this variable for FLEXY_GIT_CRYPT_KEYFILE from config if defined which will allow the user to specify key file on it's own via config file. If it will not be defined it will use the gpg key. So something like:
`f'git-crypt unlock {config.RUN.get('FLEXY_GIT_CRYPT_KEYFILE', '')}',`

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+        args.append(f"{self.flexy_img_url}")
+        return args
+
+    def clone_and_unlock_ocs_private_conf(self):
+        """
+        Clone ocs_private_conf (flexy env and config) repo into
+        flexy_mnt_host_dir
+
+        """
+        clone_repo(self.flexy_private_conf_url, self.flexy_host_private_conf_dir_path)
```
petr-balogh, line 157: We were discussing with @dahorak that we would like to:
1) Have possibility to change those info by config which you can already define URL. But we are missing also the custom branch. So please have variable also for branch name.

2) The repository does not have to be cloned and you can use just env file from local source as Daniel pointed out in different comment maybe. So this way we can have variable for env file passed also via config, and if defined it will have higher priority and then we don't need to clone the private repository via OCS-CI at all and just reuse specific file?

3) We thought that maybe will be better to use multiple env files for different approaches like PSI / Disconnected on AWS and so on, and change just specific variables we need instead of using only one env file.  @dahorak can you maybe elaborate more on this maybe on proper place and drop a comment there. Thanks

dahorak: My points are, that different templates might require different variables to be set and it might be quite messy to configure them all through the normal *ocs-ci* configuration files. And also it might be good to have good overview with what configuration the particular run was executed, but on the other hand some of the information in the env file might be considered as private, so it is probably not good to print it to the log. If it will be directly the env file from private git repository with only few minor changes (e.g. cluster name, OCP image version), than anybody with the rights to access the git repo can review the configuration but it will not be "leaked" widely.

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+        # Host dir path which will be mounted inside flexy container
+        # This will be root for all flexy housekeeping
+        self.cluster_name = config.ENV_DATA['cluster_name']
+        self.flexy_mnt_host_dir = os.path.expanduser(
+            config.ENV_DATA['flexy_mnt_host_dir']
+        )
+        # Host dir path for private_conf dir
+        self.flexy_host_private_conf_dir_path = os.path.join(
+            self.flexy_mnt_host_dir, config.ENV_DATA['flexy_host_private_conf_dir']
+        )
```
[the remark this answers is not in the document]
shylesh: ack, Initial idea was to by default download private-conf and then make the changes and mount it inside container .  I will default this behaviour to just use a default env file so that no clone is necessary by us instad flexy does it itself inside container

ocs_ci/deployment/flexy.py
shylesh: @dahorak ,
How do you want this config file to be passed ? through --ocsci-conf ?

ocs_ci/deployment/flexy.py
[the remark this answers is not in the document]
dahorak: I think, that if you would like to use the env file from the private repository, than you will have to clone it. In other words, if my understanding is correct, the env file have to be accessible to the `podman run ...` command before it even start to clone any repositories inside of the container.

ocs_ci/deployment/flexy.py
dahorak: I would use some configuration variable in configuration (e.g. `flexy_env_file` in `ENV_DATA` section, or somewhere else?).

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,274 @@
...
+        """
+        Common flexy prerequisites like cloning the private-conf
+        repo locally and updating the contents with user supplied
+        values
+
+        """
+        if not os.path.exists(self.flexy_mnt_host_dir):
+            os.mkdir(self.flexy_mnt_host_dir)
+            os.chmod(self.flexy_mnt_host_dir, mode=0o777)
+        self.clone_and_unlock_ocs_private_conf()
```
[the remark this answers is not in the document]
shylesh: @dahorak @petr-balogh ,
By default we are referring to template file which lives inside openshift-misc repo, so here you are saying that either user has already checkout openshift-misc repo and modified template file ?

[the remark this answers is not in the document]
petr-balogh: @shylesh we've discussed this yesterday.  This comment refer to line above I guess for ` self.clone_and_unlock_ocs_private_conf()` , and in case the user will provide the absolute path to `flexy env` file  then no need to clone the repository locally for private repo.

ocs_ci/deployment/baremetal.py
```diff
@@ -11,3 +21,80 @@ class BAREMETALUPI:
     def __init__(self):
         logger.info("BAREMETAL UPI")
         # TODO: OCP,OCS deployment
+
+
+class BAREMETALPSIUPI(Deployment):
```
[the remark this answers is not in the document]
shylesh: done

ocs_ci/deployment/baremetal.py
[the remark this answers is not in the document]
shylesh: Making ```ocs-ci/data``` as the initial flexy working dir and then copying to nfs mount works. 
@petr-balogh I remember from our discussion that copying to NFS mount should be done from Jenkins code rather than OCS-CI , shall we stick to same plan ? I haven't added any copying code in updated PR .

ocs_ci/deployment/baremetal.py
```diff
@@ -11,3 +21,90 @@ class BAREMETALUPI:
...
+            run_cmd(f"oc create -f {constants.NTP_CHRONY_CONF}")
+            # add disks to instances
+            self.build_psi_conf()
+            self.utils = psiutils.PSIUtils(self.psi_conf)
+            # Get all instances and for each instance add
+            # one disk
+            pattern = config.ENV_DATA['cluster_name']
+            for instance in self.utils.get_instances_with_pattern(pattern):
+                vol = self.utils.create_volume(
+                    name=f'disk0',
```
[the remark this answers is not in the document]
shylesh: done

ocs_ci/deployment/baremetal.py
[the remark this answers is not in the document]
shylesh: done, made it as 10 sec

ocs_ci/deployment/flexy.py
shylesh: done, through CLI option ```--flexy-env-file```

ocs_ci/deployment/flexy.py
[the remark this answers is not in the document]
shylesh: removed log file altogether, instead directly getting from stdout PIPE.

ocs_ci/deployment/flexy.py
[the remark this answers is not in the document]
shylesh: done

ocs_ci/deployment/flexy.py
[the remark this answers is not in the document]
shylesh: done

ocs_ci/deployment/flexy.py
shylesh: done

ocs_ci/deployment/flexy.py
shylesh: Using ```select``` we had a side effect of ```out of order logs``` hence redirected stderr to stdout itself and reading directly from stdout pipe continuously.

ocs_ci/deployment/flexy.py
shylesh: done

ocs_ci/deployment/flexy.py
shylesh: done, @petr-balogh one thing I couldn't address is private-conf branch. Flexy conf file doesn't have option to get particular branch but only URL for the repo can be specified in env file.

ocs_ci/deployment/flexy.py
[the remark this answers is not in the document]
shylesh: done

ocs_ci/deployment/flexy.py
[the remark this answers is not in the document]
shylesh: @dahorak @petr-balogh 
This one is still pending, I am bit confused about this crypt file requirement,
we are having 3 options to specify flexy env 

1.Directly pass env file's absolute path to ocs-ci
2. Pass private-conf repo through yaml to ocs-ci
3. If nothing is specified we will go with default flexy-ocs-private conf repo.

In all the above cases we expect crypt key to be present in env file so that we don't have to handle it in ocs-ci ?

WDYT ?

ocs_ci/deployment/flexy.py
shylesh: done

ocs_ci/deployment/flexy.py
shylesh: Done, did it in the correct order now.

ocs_ci/deployment/flexy.py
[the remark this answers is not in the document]
shylesh: Addressed in the above comments

ocs_ci/deployment/ocp.py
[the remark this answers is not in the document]
shylesh: done

ocs_ci/utility/psiutils.py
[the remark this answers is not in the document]
shylesh: done

ocs_ci/utility/psiutils.py
shylesh: done

ocs_ci/utility/psiutils.py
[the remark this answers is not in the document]
shylesh: done

ocs_ci/utility/psiutils.py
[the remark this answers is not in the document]
shylesh: As we already had discussion , there is no timeout param for session.

ocs_ci/utility/psiutils.py
shylesh: done

ocs_ci/utility/psiutils.py
shylesh: done

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,277 @@
...
+                raise CalledProcessError(p.returncode, p.args)
+            logger.info("Flexy run finished successfully")
+
+    def build_install_cmd(self):
+        """
+        Build flexy command line for 'deploy' operation
+
+        """
+        cmd = shlex.split("sudo podman run --rm=true")
+        flexy_container_args = self.build_container_args()
```
dahorak, line 117: `self.build_container_args()` expects  `purpose` argument (there is no default value)

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,277 @@
...
+        cmd_string = list2cmdline(cmd + flexy_container_args)
+        return cmd_string
+
+    def build_destroy_cmd(self):
+        """
+        Build flexy command line for 'destroy' operation
+
+        """
+        cmd = shlex.split("sudo podman run --rm=true")
+        flexy_container_args = self.build_container_args()
```
dahorak, line 127: Same as above, `self.build_container_args()` expects  `purpose` argument (there is no default value).

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,277 @@
...
+        args = list()
+        args.append(f"--env-file={self.flexy_env_file}")
+        args.append(f"-w={self.flexy_mnt_container_dir}")
+        # For destroy on NFS mount, relabel=shared will not work
+        # with podman hence we will keep 'relable=shared' only for
+        # deploy which happens on Jenkins slave's local fs
+        if purpose == 'destroy':
+            args.append(
+                f"--mount=type=bind,source={self.flexy_host_dir},"
+                f"destination=f{self.flexy_mnt_container_dir}"
```
dahorak, line 152:
```suggestion
                f"destination={self.flexy_mnt_container_dir}"
```

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,277 @@
...
+            )
+        else:
+            args.append(
+                f"--mount=type=bind,source={self.flexy_host_dir},"
+                f"destination=f{self.flexy_mnt_container_dir},relabel=shared"
```
dahorak, line 157:
```suggestion
                f"destination={self.flexy_mnt_container_dir},relabel=shared"
```

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,277 @@
...
+        if not config.ENV_DATA.get('flexy-env-file'):
+            self.flexy_private_conf_url = config.ENV_DATA.get(
+                'flexy_private_conf_url',
+                constants.FLEXY_DEFAULT_PRIVATE_CONF_URL
+            )
+            # Host dir path for private_conf dir where private-conf will be
+            # cloned
+            self.flexy_host_private_conf_dir_path = os.path.join(
+                self.flexy_host_dir, 'flexy-ocs-private'
+            )
```
dahorak, line 55: I think, that it is not required (and maybe not expected) to have the private repository cloned to the flexy working directory or did I miss something? (If I try to run `podman run ...` command pointing to working  directory where private repo is already cloned into flexy directory, the command even fails with following error: `fatal: destination path 'flexy' already exists and is not an empty directory.`

shylesh: @dahorak ,
The idea of cloning private-repo to working dir so that when we backup this whole flexy dir to NFS we can do it in one shot.

Here is the flow that I have thought about how it would work on Jenkins
1. Create a directory called ```flexy``` in ```/path/to/ocs-ci/data/```
2. Clone ```flexy-ocs-private``` (assuming user wants defaults) into ```/path/to/ocs-ci/data/flexy/``` 
3. After modifying any private-conf variables we will run podman
```podman run --env=/path/to/ocs-ci/data/flexy/flexy-ocs-private/ocs-osp.env -w=/mnt  --mount=type=bind,source=/path/to/ocs-ci/data/flexy,destination=/mnt,relabel=shared   docker-registry.upshift.redhat.com/aosqe/flexy:poc ```

When container runs it will create another called ```flexy``` inside ```/path/to/ocs-ci/data/flexy```, this has worked for me while testing manually, I will update you how it goes with automation.

Please suggest me if I am missing something or any other requirements are missing. 
I can change this approach in case if it doesn't work. Let me know your thoughts.

**Comment by shylesh** [comment]
Rebased and conflict resolved for latest master

File: ocs_ci/deployment/flexy.py (hunk 40 above)
Comment (dahorak) [reply ocs_ci/deployment/flexy.py:55]: Ok, now I understand the `flexy` directory from the container will be in `flexy` directory cloned by us. I'm thinking if it is worth to mixup those two repositories together - or if it is even worth to backup the private repository (or any repository with the env file) cloned by us. If it wouldn't be enough to backup only the `flexy` directory created and used by the container?

File: ocs_ci/deployment/baremetal.py (hunk 18 above)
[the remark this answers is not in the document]
Comment (petr-balogh) [reply ocs_ci/deployment/baremetal.py:54]: @shylesh what I remember that we discussed you will cover copy here from ocs-ci .  In the jenkins we can do some second check if the execution was aborted, then we can do the copy from jenkins as well if it was not completed by ocs-ci.    So please go with copy from ocs-ci and from the jenkins job we can do later on to be sure we did not loose any data.

File: ocs_ci/deployment/flexy.py (hunk 31 above)
Comment (petr-balogh) [reply ocs_ci/deployment/flexy.py:169]: @shylesh I think that you can by using this variable:  GIT_PRIVATE_BRANCH .  Flexy is processing variable names and it should work similar like with repo: `GIT_PRIVATE_OPENSHIFT_MISC_URI` where branch can be defined by just have _BRANCH at the end of variable instead of _URI.  Like:  `GIT_PRIVATE_OPENSHIFT_MISC_BRANCH`

File: ocs_ci/deployment/flexy.py (hunk 25 above)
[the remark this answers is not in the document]
Comment (petr-balogh) [reply ocs_ci/deployment/flexy.py:172]: > @dahorak @petr-balogh
> This one is still pending, I am bit confused about this crypt file requirement,
> we are having 3 options to specify flexy env
> 
> 1.Directly pass env file's absolute path to ocs-ci
If you pass absolute path you expect that this file is not encrypted and is fully readable and contains key inside. (This case you don't clone the repo - so no need for unlock)
> 2. Pass private-conf repo through yaml to ocs-ci
If you pass the repository then ocs-ci will clone it, but to be able to pass any env file from this repo you have to unlock the repository. So this case the repo should be unlocked by `FLEXY_GIT_CRYPT_KEYFILE` - which we will generate to jenkins config file or can be done by CLI parameter.   If `FLEXY_GIT_CRYPT_KEYFILE` not passed - we will try to unlock with GPG key so the only command `git-crypt unlock` will be executed (and expect user to have gpg key)
> 3. If nothing is specified we will go with default flexy-ocs-private conf repo.
This will be the same approach as above.
> 
> In all the above cases we expect crypt key to be present in env file so that we don't have to handle it in ocs-ci ?
As you need to unlock the repository before you pass env file to flexy you need to handle unlock in ocs-ci.
> 
> WDYT ?

File: ocs_ci/deployment/flexy.py (hunk 31 above)
Comment (shylesh) [reply ocs_ci/deployment/flexy.py:169]: @petr-balogh Ok I will try this manually, Daniel also told the same thing about branch . I will update my findings.

File: ocs_ci/deployment/baremetal.py (hunk 18 above)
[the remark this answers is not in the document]
Comment (shylesh) [reply ocs_ci/deployment/baremetal.py:54]: @petr-balogh sorry I might have missed that part, ack, I will do it in ocs-ci as well.

File: ocs_ci/deployment/flexy.py (hunk 36 above)
Comment (shylesh) [reply ocs_ci/deployment/flexy.py:164]: done

File: ocs_ci/deployment/flexy.py (hunk 37 above)
Comment (shylesh) [reply ocs_ci/deployment/flexy.py]: done, now we have empty string as default which means 'deploy' for flexy

File: ocs_ci/deployment/flexy.py (hunk 38 above)
Comment (shylesh) [reply ocs_ci/deployment/flexy.py]: done

File: ocs_ci/deployment/flexy.py (hunk 39 above)
Comment (shylesh) [reply ocs_ci/deployment/flexy.py]: done

File: ocs_ci/deployment/flexy.py (hunk 25 above)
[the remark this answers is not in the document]
Comment (shylesh) [reply ocs_ci/deployment/flexy.py:236]: > > @dahorak @petr-balogh
> > This one is still pending, I am bit confused about this crypt file requirement,
> > we are having 3 options to specify flexy env
> > 1.Directly pass env file's absolute path to ocs-ci
> > If you pass absolute path you expect that this file is not encrypted and is fully readable and contains key inside. (This case you don't clone the repo - so no need for unlock)
> > 2. Pass private-conf repo through yaml to ocs-ci
> > If you pass the repository then ocs-ci will clone it, but to be able to pass any env file from this repo you have to unlock the repository. So this case the repo should be unlocked by `FLEXY_GIT_CRYPT_KEYFILE` - which we will generate to jenkins config file or can be done by CLI parameter.   If `FLEXY_GIT_CRYPT_KEYFILE` not passed - we will try to unlock with GPG key so the only command `git-crypt unlock` will be executed (and expect user to have gpg key)

@petr-balogh @dahorak  ,
In this case where user wants to pass his/her own private-conf repo How we want to pass location of keyfile ? through URL on Jenkin's job page ? OR a facility to upload keyfile in the Jenkin's job file ?

> > 3. If nothing is specified we will go with default flexy-ocs-private conf repo.
> > This will be the same approach as above.
> > In all the above cases we expect crypt key to be present in env file so that we don't have to handle it in ocs-ci ?
> > As you need to unlock the repository before you pass env file to flexy you need to handle unlock in ocs-ci.
> > WDYT ?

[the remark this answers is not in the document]
Comment (petr-balogh) [reply ocs_ci/deployment/flexy.py:236]: @shylesh from jenkins side we will use jenkins credentials while we are executing ocs-ci and will pass this location to the ocs-ci in way how you will integrate it CLI pointing to the file location or will update config file we generate in jenkins to provide this in proper section in conf file.

Do not think about it much, for now we will not use any other keyfile than only one which we have in our current private repo - do not expect that user will change that in their own fork for now.

Once it will be needed we will manage some way we will agree on but for now it should be passed from jenkins credentials like we use already with auth.yaml file and so on.

File: ocs_ci/deployment/baremetal.py [hunk 41]
```diff
@@ -11,3 +20,72 @@ class BAREMETALUPI:
...
+                'flexy/workdir/install-dir'
+            )
+            logger.info(
+                "Symlinking %s to %s", abs_cluster_path, flexy_cluster_path
+            )
+            os.symlink(abs_cluster_path, flexy_cluster_path)
+            self.test_cluster()
+            # We need NTP for OCS cluster to become clean
+            logger.info("creating ntp chrony")
+            run_cmd(f"oc create -f {constants.NTP_CHRONY_CONF}")
```
prsurve, line 64: do we need to wait here to reflect changes in all worker nodes, as I have seen worker nodes will be marked shedulingdisable for sometime

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,342 @@
...
+        self.flexy_mnt_container_dir = config.ENV_DATA.get(
+            'flexy_mnt_container_dir', constants.FLEXY_MNT_CONTAINER_DIR
+        )
+        self.flexy_img_url = config.ENV_DATA.get('flexy_img_url')
+
+        # If user has provided an absolute path for env file
+        # then we don't clone private-conf repo else
+        # we will look for 'flexy_private_conf_url' , if not specified
+        # we will go ahead with default 'flexy-ocs-private' repo
+        if not config.ENV_DATA.get('flexy-env-file'):
```
dahorak, line 88: I think, that for all other configuration variables, we are using underscores instead of dash in the key name, so no sure if it wouldn't be better to rename it to `ENV_DATA['flexy_env_file']` (but I'm not sure, if it is mandatory or not).

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,342 @@
...
+            # Host dir path for private_conf dir where private-conf will be
+            # cloned
+            self.flexy_host_private_conf_dir_path = os.path.join(
+                self.flexy_host_dir, 'flexy-ocs-private'
+            )
+            self.flexy_env_file = os.path.join(
+                self.flexy_host_private_conf_dir_path, constants.FLEXY_DEFAULT_ENV_FILE
+            )
+        else:
+            self.flexy_env_file = config.ENV_DATA['flexy-env-file']
```
dahorak, line 106: I'm thinking about the logic here and the comment and I think that we should slightly change the logic somehow - in this case it will not be possible to have multiple env files in the flexy private repository - because when you will provide different env file name, it will skip the cloning of the private repository.

I would suggest following approach:
* set `self.flexy_env_file` to `ENV_DATA['flexy_env_file']`, if configured or use default from  `constants.FLEXY_DEFAULT_ENV_FILE`
* check if `self.flexy_env_file` exists
* if `self.flexy_env_file` does not exists, expects that it is part of the private flexy repo and clone that repo
* finally once more check, that the `self.flexy_env_file` exists - if not raise exception

ocs_ci/deployment/flexy.py
```diff
@@ -0,0 +1,342 @@
...
+
+        Returns:
+            list: of flexy container args
+
+        """
+        args = list()
+        args.append(f"--env-file={self.flexy_env_file}")
+        args.append(f"-w={self.flexy_mnt_container_dir}")
+        # For destroy on NFS mount, relabel=shared will not work
+        # with podman hence we will keep 'relable=shared' only for
```
dahorak, line 195:
```suggestion
        # with podman hence we will keep 'relabel=shared' only for
```

ocs_ci/framework/conf/flexy_default_config.yaml
```diff
@@ -0,0 +1,15 @@
+ENV_DATA:
+  platform: 'baremetalpsi'
+  deployment_type: 'upi'
+  flexy_img_url: 'docker-registry.upshift.redhat.com/aosqe/flexy:poc'
```
dahorak, line 4: Is it plausible, that we will need to change this image? I think, that it might be saved in constants and consider this `flexy_img_url` configuration key only as optional.

ocs_ci/framework/conf/flexy_default_config.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+# Override flexy params here
+FLEXY:
+  default_private_conf_repo: 'git@gitlab.cee.redhat.com:ocs/flexy-ocs-private.git'
+  flexy_private_conf_url: ''
+  # Volume which will be attached to instance,
+  # which will be consumed for osd. (size in GB)
+  volume_size: 500
```
dahorak, line 12: Wouldn't be better/sufficient to use existing `device_size` configuration variable?

e.g. https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/production_device_size.yaml#L4

ocs_ci/framework/conf/flexy_default_config.yaml
```diff
@@ -0,0 +1,15 @@
+ENV_DATA:
+  platform: 'baremetalpsi'
+  deployment_type: 'upi'
+  flexy_img_url: 'docker-registry.upshift.redhat.com/aosqe/flexy:poc'
+
+# Override flexy params here
+FLEXY:
+  default_private_conf_repo: 'git@gitlab.cee.redhat.com:ocs/flexy-ocs-private.git'
+  flexy_private_conf_url: ''
```
dahorak, line 9: Did I understand it correctly, that the `flexy_private_conf_url` is for custom private repository url and `default_private_conf_repo` is *default* value for it? In other words, if `flexy_private_conf_url` is not specified, `default_private_conf_repo` is used? If yes, then `default_private_conf_repo` should be probably stored in *constants*, rather than here in configuration file.

Also I would suggest to rename the *flexy_private_conf_url* to indicate that it should point to repository (because this sounds more like direct link to configuration file).

conf/ocsci/flexy_baremetalpsi_upi.yaml
```diff
@@ -0,0 +1,8 @@
+DEPLOYMENT:
+  local_storage: true
+ENV_DATA:
+  platform: 'baremetalpsi'
+  deployment_type: 'upi'
+
+FLEXY:
+  LAUNCHER_VARS: {"num_workers":"3", "openstack_network":"provider_net_cci_9","rhel_vm_template":"ql-rhel-7.6","qe_additional_repos":"[{'id': 'rhel7', 'name': 'rhel7', 'baseurl': 'http://pulp.dist.prod.ext.phx2.redhat.com/content/dist/rhel/server/7/7Server/x86_64/os', 'enabled': 1, 'gpgcheck': 0}]","vm_type_masters":"m1.large","vm_type_workers":"ci.m1.xlarge.ocs03","ssh_key_name":"openshift-dev","rhel_ansible_user":"cloud-user","installer_payload_image":"registry.svc.ci.openshift.org/ocp/release:4.3","num_nodes": 1}
```
petr-balogh, line 8: @shylesh does this LAUNCHER_VARS needs to be in json format?
I guess it is but for our configuration it's not good for merging purpose with other config file if needed as it will overwrite whole LAUNCHER_VARS data.

I am thinking about to have those variables defined in pure yaml here, to allow us to merge data together from multiple config files and then when those data are used from here, transfer them to json.

so the yaml data will look like (just converted your data to yaml via python):

```yaml
installer_payload_image: registry.svc.ci.openshift.org/ocp/release:4.3
num_nodes: 1
num_workers: '3'
openstack_network: provider_net_cci_9
qe_additional_repos: '[{''id'': ''rhel7'', ''name'': ''rhel7'', ''baseurl'': ''http://pulp.dist.prod.ext.phx2.redhat.com/content/dist/rhel/server/7/7Server/x86_64/os'',
  ''enabled'': 1, ''gpgcheck'': 0}]'
rhel_ansible_user: cloud-user
rhel_vm_template: ql-rhel-7.6
ssh_key_name: openshift-dev
vm_type_masters: m1.large
vm_type_workers: ci.m1.xlarge.ocs03
```

```
In [13]: yaml_data = yaml.dump(data, default_flow_style=False)

In [14]: print(yaml_data)
installer_payload_image: registry.svc.ci.openshift.org/ocp/release:4.3
num_nodes: 1
num_workers: '3'
openstack_network: provider_net_cci_9
qe_additional_repos: '[{''id'': ''rhel7'', ''name'': ''rhel7'', ''baseurl'': ''http://pulp.dist.prod.ext.phx2.redhat.com/content/dist/rhel/server/7/7Server/x86_64/os'',
  ''enabled'': 1, ''gpgcheck'': 0}]'
rhel_ansible_user: cloud-user
rhel_vm_template: ql-rhel-7.6
ssh_key_name: openshift-dev
vm_type_masters: m1.large
vm_type_workers: ci.m1.xlarge.ocs03


In [15]: python_data = yaml.load(yaml_data)


In [16]: print(json.dumps(python_data))
{"installer_payload_image": "registry.svc.ci.openshift.org/ocp/release:4.3", "num_nodes": 1, "num_workers": "3", "openstack_network": "provider_net_cci_9", "qe_additional_repos": "[{'id': 'rhel7', 'name': 'rhel7', 'baseurl': 'http://pulp.dist.prod.ext.phx2.redhat.com/content/dist/rhel/server/7/7Server/x86_64/os', 'enabled': 1, 'gpgcheck': 0}]", "rhel_ansible_user": "cloud-user", "rhel_vm_template": "ql-rhel-7.6", "ssh_key_name": "openshift-dev", "vm_type_masters": "m1.large", "vm_type_workers": "ci.m1.xlarge.ocs03"}
```

What do you think? Does not have to be in this PR, but we can open issue for it if you think it is worth.

Maybe our current functionality will cover this merging as well as it will be loaded as dictionary anyway, but for me the pure yaml format in the config is more readable than this one.

ocs_ci/deployment/flexy.py
petr-balogh: @shylesh fix this typo please

ocs_ci/templates/ocp-deployment/ntp_chrony.yaml
```diff
@@ -0,0 +1,18 @@
+# This example MachineConfig replaces /etc/chrony.conf
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: worker
+  name: 50-examplecorp-chrony
```
prsurve, line 7: Can we name this like `worker-chrony-configuration`

petr-balogh:
Please rebase @shylesh I see some conflicts ;)

petr-balogh:
And also see some failure in the travis:

```
    def test_defaults(self):
243        config_sections = framework.config.to_dict().keys()
244        for section_name in config_sections:
245            section = getattr(framework.config, section_name)
246>           assert section == framework.config.get_defaults()[section_name]
247E           KeyError: 'FLEXY'
248
249ocs_ci/framework/tests/test_config.py:16: KeyError```

**Comment by shylesh** [comment]
Rebased on latest master and conflicts resolved

File: conf/ocsci/flexy_baremetalpsi_upi.yaml (hunk 48 above)
Comment (shylesh) [reply conf/ocsci/flexy_baremetalpsi_upi.yaml:8]: ack,https://github.com/red-hat-storage/ocs-ci/issues/2489

File: ocs_ci/deployment/flexy.py (hunk 44 above)
Comment (shylesh) [reply ocs_ci/deployment/flexy.py:163]: done

File: ocs_ci/framework/conf/flexy_default_config.yaml (hunk 45 above)
Comment (shylesh) [reply ocs_ci/framework/conf/flexy_default_config.yaml:4]: done

File: ocs_ci/framework/conf/flexy_default_config.yaml (hunk 47 above)
Comment (shylesh) [reply ocs_ci/framework/conf/flexy_default_config.yaml:9]: done

File: ocs_ci/templates/ocp-deployment/ntp_chrony.yaml (hunk 49 above)
Comment (shylesh) [reply ocs_ci/templates/ocp-deployment/ntp_chrony.yaml:7]: done

File: ocs_ci/deployment/baremetal.py [hunk 50]
```diff
@@ -11,3 +20,72 @@ class BAREMETALUPI:
...
+            # Point cluster_path to cluster-dir created by flexy
+            abs_cluster_path = os.path.abspath(self.cluster_path)
+            flexy_cluster_path = os.path.join(
+                self.flexy_instance.flexy_mnt_host_dir,
+                'flexy/workdir/install-dir'
+            )
+            logger.info(
+                "Symlinking %s to %s", abs_cluster_path, flexy_cluster_path
+            )
+            os.symlink(abs_cluster_path, flexy_cluster_path)
```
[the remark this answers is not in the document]
shylesh: Removed here, now its present only in base class

ocs_ci/deployment/flexy.py
shylesh: done

ocs_ci/deployment/flexy.py
shylesh: How about this:
We will check ENV_DATA['flexy_env_file'] whether it's absolute path or relative path (whether it starts with '/' OR not), if its absolute path then we will not do anything (assuming that its run on local machine), if its relative path then we assume its inside private repo. So we can have multiple env files in repository (like independent-conf.env, baremetal-conf.env etc) and just that name will be passed to flexy_env_file. WDYT ?

ocs_ci/deployment/baremetal.py
shylesh: @prsurve  Any suggestion on  what's the exact condition that we have to wait on ?

petr-balogh:
@shylesh can you remove WIP to execute travis please?  And can we have one PR verification job with this PR which will just run the deployment over the AWS to be sure we didn't break anything?

Thanks

ocs_ci/deployment/baremetal.py
prsurve: yes, i have added condition check-in my pr https://github.com/red-hat-storage/ocs-ci/blob/60e344e7fa4a5ecf54d590e952ca64f26ef0959a/ocs_ci/utility/utils.py#L2378-L2392

ocs_ci/deployment/flexy.py
dahorak: @shylesh I agree with your suggestion, if we will need some enhancement, we can do it in separated PR later.

ocs_ci/deployment/baremetal.py
shylesh: Reply from Pratik,

======================================================
you can check for machineconfigpools.machineconfiguration.openshift.io
def wait_for_machineconfigpool_status(resource_name, timeout=800):
    """
    Check for Machineconfigpool status

    Args:
        resource_name (str): The name of the resource to wait (master,worker)
        timeout (int): Time in seconds to wait:


    """
    from ocs_ci.ocs import ocp
    ocp_obj = ocp.OCP(kind=constants.MACHINECONFIGPOOL, resource_name=resource_name)
    machine_count = ocp_obj.get()['status']['machineCount']

    ocp_obj.wait_for_resource(condition=str(machine_count), column="READYMACHINECOUNT", timeout=timeout)
the above code will wait for all machine to come in ready state after ntp config is applied

================================================================

shylesh:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-destroy-ocs-cluster/7536/ . Flexy deployment worked, but job failed at FIO test

Issue #1992, opened 2020-05-04
Fixture measure_stop_ceph_osd should wait for osd to be back


mbukatov:
Fixture `measure_stop_ceph_osd` should wait for osd to be back so that cluster health is not affected.

While the osd is back quite fast, it's not consistent enough for other tests nor automatic health check, eg.:

- 2020-04-25 00:22:20,888 Upscaling deployment rook-ceph-osd-2 back to 1 (fixture)
- 2020-04-25 00:22:24,803 HEALTH_WARN 1 osds down ...                           
- 2020-04-25 00:22:26,136 HEALTH_WARN 1 osds down ...                           
- 2020-04-25 00:22:26,137 Ceph cluster health is not OK. Health: HEALTH_WARN 1 osds down ...
- 2020-04-25 00:22:57,484 Ceph cluster health is HEALTH_OK.

PR #2287, opened 2020-06-10
add ceph health check to alerts tests

fbalak:
Fixes https://github.com/red-hat-storage/ocs-ci/issues/1992

### DIFF
--- a/tests/manage/monitoring/conftest.py
+++ b/tests/manage/monitoring/conftest.py
@@ -8,10 +8,10 @@ import pytest
 
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants, ocp
-from ocs_ci.ocs.cluster import CephCluster
 from ocs_ci.ocs.fiojob import workload_fio_storageutilization
 from ocs_ci.ocs.resources import pod
 from ocs_ci.ocs.resources.mcg_bucket import S3Bucket
+from ocs_ci.utility.utils import ceph_health_check
 from ocs_ci.utility.workloadfixture import measure_operation
 from tests import helpers
 from tests.helpers import create_unique_resource_name
@@ -66,8 +66,7 @@ def measure_stop_ceph_mgr(measurement_dir):
 
     # wait for ceph to return into HEALTH_OK state after mgr deployment
     # is returned back to normal
-    ceph_cluster = CephCluster()
-    ceph_cluster.cluster_health_check()
+    ceph_health_check(tries=20, delay=15)
 
     return measured_op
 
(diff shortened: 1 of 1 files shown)


tests/manage/monitoring/conftest.py
```diff
@@ -62,6 +63,12 @@ def stop_mgr():
     measured_op = measure_operation(stop_mgr, test_file)
     logger.info(f"Upscaling deployment {mgr} back to 1")
     oc.exec_oc_cmd(f"scale --replicas=1 deployment/{mgr}")
+
+    # wait for ceph to return into HEALTH_OK state after mgr deployment
+    # is returned back to normal
+    ceph_cluster = CephCluster()
+    ceph_cluster.cluster_health_check()
```
mbukatov, line 70: I must be missing something. How come this won't fail?

I see that the original problem was that after issuing a request to scale a ceph deployment back to the original number, it takes some time for the cluster to recover after the missing component is back. So while the cluster recovered, a check for ceph health in teardown failed because the recovery was not fast enough. So how come this healt check would not fail and instead wait for the cluster to sort it out? Would the default timeout in `cluster_health_check()` choice work every time this is executed?

That said, if you demonstrate that it works when executed multiple times in a PR validation job, I'm ok with this as it would be still improvement, preventing false failures.

fbalak: As described in docstring for `cluster_health_check` function, if no timeout parameter is provided then the timeout will be scaled based on number of ceph pods in the cluster (which is `10 * number of ceph pods`). This gives cluster enough time to recover.

mbukatov: I'm bit reluctant to rely on a function which on one hand is called `cluster_health_check()`, but on the other hand is trying health check multiple times (based on some magic heuristic) and only fails if all these retries fail. Besides that, it also checks number of mons, mds and some noobaa stuff. The main functionality we need here is just a side effect of a design choice burried in this function (only when I actually check the source code I can understand why it could work, looking at the function name or a docstring makes me confused).

Have you tried to run it multiple times as I suggested? Have you tried to see what happens when this  `cluster_health_check()` function fails?

fbalak: I tried it 3x times. I didn't see it fail. Do you suggest to change the function in this PR or create a new `cluster_health_check` function?

ebenahar: @fbalak how about waiting for Ceph health to be OK, with a fixed timeout, regardless of validating different resources like MDS and Noobaa? 
You can use ocs_ci.utility.utils.ceph_health_check

fbalak: Ok, ocs_ci.utility.utils.ceph_health_check looks better.

mbukatov: I agree, usage of the other health check function is more explicit and should be easier to debug if something goes wrong.

PR #2288, opened 2020-06-10
Remove aws_platform_required marker from test_node_maintenance_restart_activate

### LINKED ISSUE #2286: Modify test_node_maintenance_restart_activate to run on different platforms
Modify the below test case to run on other platforms. Currently the test case is running only on AWS platform.

tests/manage/z_cluster/nodes/test_nodes_maintenance.py::TestNodesMaintenance::test_node_maintenance_restart_activate

Enabling the test case to run on BM is tracked separately - https://github.com/red-hat-storage/ocs-ci/issues/2679

This will partially fix https://github.com/red-hat-storage/ocs-ci/issues/1922

jilju:
Modify the below test case to run on different platforms. Currently the test case is running only on AWS platform.

tests/manage/z_cluster/nodes/test_nodes_maintenance.py::TestNodesMaintenance::test_node_maintenance_restart_activate
Enabling the test case to run on BM is tracked here - https://github.com/red-hat-storage/ocs-ci/issues/2679

Closes #2286

### DIFF
--- a/tests/manage/z_cluster/nodes/test_nodes_maintenance.py
+++ b/tests/manage/z_cluster/nodes/test_nodes_maintenance.py
@@ -154,28 +154,37 @@ class TestNodesMaintenance(ManageTest):
         assert typed_nodes, f"Failed to find a {node_type} node for the test"
         typed_node_name = typed_nodes[0].name
 
+        reboot_events_cmd = (
+            f"get events -A --field-selector involvedObject.name="
+            f"{typed_node_name},reason=Rebooted -o yaml"
+        )
+
+        # Find the number of reboot events in 'typed_node_name'
+        num_events = typed_nodes[0].ocp.exec_oc_cmd(reboot_events_cmd)['items']
+
         # Maintenance the node (unschedule and drain). The function contains logging
         drain_nodes([typed_node_name])
 
         platform = config.ENV_DATA['platform']
 
         # Restarting the node
-        if platform == 'vsphere':
-            nodes.restart_nodes(nodes=typed_nodes, force=True, wait=False)
-        elif platform == 'baremetal':
+        if platform == 'baremetal':
             nodes.restart_nodes(nodes=typed_nodes)
         else:
             nodes.restart_nodes(nodes=typed_nodes, wait=False)
 
         wait_for_nodes_status(
             node_names=[typed_node_name],
-            status=constants.NODE_NOT_READY_SCHEDULING_DISABLED
+            status=constants.NODE_READY_SCHEDULING_DISABLED, timeout=360
         )
 
-        wait_for_nodes_status(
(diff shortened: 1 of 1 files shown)


tests/manage/z_cluster/nodes/test_nodes_maintenance.py
```diff
@@ -154,21 +153,49 @@ def test_node_maintenance_restart_activate(
...
         drain_nodes([typed_node_name])
 
+        platform = config.ENV_DATA['platform']
+
         # Restarting the node
-        nodes.restart_nodes(nodes=typed_nodes, wait=False)
+        if platform == 'baremetal':
+            nodes.restart_nodes(nodes=typed_nodes)
+        else:
+            nodes.restart_nodes(nodes=typed_nodes, wait=False)
```
ebenahar, line 175:
```suggestion
        nodes.restart_nodes(nodes=typed_nodes, wait=(platform == 'baremetal'))
```

jilju: platform_nodes.BaremetalNodes.restart_nodes does not have wait argument. So this will fail  if platform is BM.

ebenahar: would it be good to add wait to the `restart_nodes()` method also for BM? 
If so, `ocs_ci.utility.baremetal.BAREMETAL.stop_baremetal_machines` should accept a `wait` argument so ocs_ci/utility/baremetal.py:120-125 will happen only if `wait=True`

jilju: It is good to have parameter for platform_nodes.BaremetalNodes.restart_nodes, but adding wait parameter will need redesign of ocs_ci.utility.baremetal.BAREMETAL.stop_baremetal_machines. Unlike VMware and AWS restart_nodes function BM restart_nodes does not validate node NotReady status. Also, if force is set to False in  stop_baremetal_machines, the power status check will be done outside the code segment ocs_ci/utility/baremetal.py:120-125 . This will also be needed to handle separately. 
Moreover, any change in this regard is not required for this test case because reboot events count is used to ensure that the node actually rebooted. The system status check in stop_baremetal_machines will just be an additional verification.

ebenahar: Ack. The needed change is indeed not in the scope of this test. 
How about maintaining an issue for the needed change?

jilju: Created issue https://github.com/red-hat-storage/ocs-ci/issues/2535

vavuthu (approved):

ebenahar (approved):

PR #2291, opened 2020-06-11
Fixes minor issues around FIO test

keesturam:
Fixes the following two issues. 

   a. it reports about sequential pattern, but it ran random read/write -> fixes with modifying the cr file. 
   b. it report on deviation > 5% also when it was less than 5% -> For now, we simply log the variance.

### DIFF
--- a/tests/e2e/performance/test_fio_benchmark.py
+++ b/tests/e2e/performance/test_fio_benchmark.py
@@ -65,7 +65,7 @@ def analyze_regression(io_pattern, es_username):
         if io_pattern == 'sequential':
             std_dev = 'std-dev-' + object_size
             variance = test_data[std_dev]
-            log.warning(f'variance - {variance} is greater than 5%')
+            log.info(f'variance - {variance}')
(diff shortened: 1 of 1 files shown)


tests/e2e/performance/test_fio_benchmark.py
```diff
@@ -110,8 +110,10 @@ def test_fio_workload_simple(self, ripsaw, interface, io_pattern):
         # Todo: have pvc_size set to 'get_osd_pods_memory_sum * 5'
         #  once pr-2037 is merged
         fio_cr['spec']['clustername'] = config.ENV_DATA['platform'] + get_build() + get_ocs_version()
-        fio_cr['spec']['test_user'] = interface + io_pattern
+        fio_cr['spec']['test_user'] = get_ocs_version() + interface + io_pattern
```
Avilir, line 113: I think it will be better as :
`        fio_cr['spec']['test_user'] = '_'.joine([get_ocs_version(), interface, io_pattern])`

PR #2294, opened 2020-06-11
Convert OCS Upgrade to OOP

AviadP:
This PR is preparation for ocs upgrade tier4a tests 
1. split upgrade into helper file and test file
2. create OCS upgrade class with methods

### DIFF
--- a/ocs_ci/ocs/ocs_upgrade.py
+++ b/ocs_ci/ocs/ocs_upgrade.py
@@ -1,0 +1,387 @@
+import os
+import logging
+from copy import deepcopy
+from pkg_resources import parse_version
+from tempfile import NamedTemporaryFile
+
+from ocs_ci.deployment.deployment import create_catalog_source
+from ocs_ci.framework import config
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.defaults import OCS_OPERATOR_NAME
+from ocs_ci.ocs.node import get_typed_nodes
+from ocs_ci.ocs.ocp import get_images, OCP
+from ocs_ci.ocs.resources.catalog_source import CatalogSource
+from ocs_ci.ocs.resources.csv import CSV
+from ocs_ci.ocs.resources.install_plan import wait_for_install_plan_and_approve
+from ocs_ci.ocs.resources.pod import verify_pods_upgraded
+from ocs_ci.ocs.resources.packagemanifest import (
+    get_selector_for_ocs_operator,
+    PackageManifest,
+)
+from ocs_ci.ocs.resources.storage_cluster import get_osd_count
+from ocs_ci.ocs.utils import setup_ceph_toolbox
+from ocs_ci.utility.utils import (
+    get_latest_ds_olm_tag,
+    get_next_version_available_for_upgrade,
+    get_ocs_version_from_image,
+    load_config_file,
+)
+from ocs_ci.utility.templating import dump_data_to_temp_yaml
+
+log = logging.getLogger(__name__)
+
+
+def get_upgrade_image_info(old_csv_images, new_csv_images):
+    """
+    Log the info about the images which are going to be upgraded.
+
+    Args:
+        old_csv_images (dict): dictionary with the old images from pre upgrade CSV
+        new_csv_images (dict): dictionary with the post upgrade images
+
+    Returns:
+        tuple: with three sets which contain those types of images
+            old_images_for_upgrade - images which should be upgraded
+            new_images_to_upgrade - new images to which upgrade
+            unchanged_images - unchanged images without upgrade
+
+    """
+    old_csv_images = set(old_csv_images.values())
+    new_csv_images = set(new_csv_images.values())
+    old_images_for_upgrade = old_csv_images - new_csv_images
+    log.info(
+        f"Old images which are going to be upgraded: "
+        f"{sorted(old_images_for_upgrade)}"
+    )
+    new_images_to_upgrade = new_csv_images - old_csv_images
+    log.info(
+        f"New images for upgrade: "
+        f"{sorted(new_images_to_upgrade)}"
+    )
+    unchanged_images = old_csv_images.intersection(new_csv_images)
+    log.info(
+        f"Unchanged images after upgrade: "
+        f"{sorted(unchanged_images)}"
+    )
+    return (
+        old_images_for_upgrade,
+        new_images_to_upgrade,
+        unchanged_images,
+    )
+
+
+def verify_image_versions(old_images, upgrade_version):
+    """
+    Verify if all the images of OCS objects got upgraded
+
+    Args:
+        old_images (set): set with old images
+        upgrade_version (packaging.version.Version): version of OCS
+
+    """
+    number_of_worker_nodes = len(get_typed_nodes())
+    osd_count = get_osd_count()
+    verify_pods_upgraded(old_images, selector=constants.OCS_OPERATOR_LABEL)
+    verify_pods_upgraded(old_images, selector=constants.OPERATOR_LABEL)
+    # in 4.3 app selector nooba have those pods: noobaa-core-ID, noobaa-db-ID,
+    # noobaa-operator-ID but in 4.2 only 2: noobaa-core-ID, noobaa-operator-ID
+    nooba_pods = 2 if upgrade_version < parse_version('4.3') else 3
+    verify_pods_upgraded(
+        old_images, selector=constants.NOOBAA_APP_LABEL, count=nooba_pods
+    )
+    verify_pods_upgraded(
+        old_images, selector=constants.CSI_CEPHFSPLUGIN_LABEL,
+        count=number_of_worker_nodes,
+    )
+    verify_pods_upgraded(
+        old_images, selector=constants.CSI_CEPHFSPLUGIN_PROVISIONER_LABEL,
+        count=2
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -0,0 +1,376 @@
...
+            bool: True if upgrade version is newer than current version
+
+        """
+        assert self.parsed_upgrade_version >= self.parsed_version_before_upgrade, (
+            f"Version you would like to upgrade to: {upgrade_version} "
+            f"is not higher or equal to the version you currently running: "
+            f"{self.version_before_upgrade}"
+        )
+
+        return True
```
ebenahar, line 184: the function doesn't return False. Hence, `return` is not needed

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -0,0 +1,376 @@
...
+        'upgrade_in_current_source', False
+    )
+    upgrade_version = config.UPGRADE.get(
+        "upgrade_ocs_version", version_before_upgrade
+    )
+    if ocs_registry_image:
+        upgrade_version = get_ocs_version_from_image(
+            ocs_registry_image
+        )
+    parsed_upgrade_version = parse_version(upgrade_version)
```
ebenahar, line 149: I think you need to initialize these vars in an `__init__()` method

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -0,0 +1,376 @@
...
+
+        subscription_plan_approval = config.DEPLOYMENT.get(
+            'subscription_plan_approval'
+        )
+        if subscription_plan_approval == 'Manual':
+            wait_for_install_plan_and_approve(self.namespace)
+
+    def check_if_upgrade_completed(self, channel, csv_name_pre_upgrade):
+        """
+        Checks if OCS operator finishes it's upgrade
```
ebenahar, line 277: add an empty line below

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -0,0 +1,376 @@
...
+        Args:
+            channel: (str): OCS subscription channel
+            csv_name_pre_upgrade: (str): OCS operator name
+
+        Returns:
+            bool: True if upgrade completed
```
ebenahar, line 283: `bool: True if upgrade completed, False otherwise`

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -0,0 +1,376 @@
...
+        )
+        csv_name_post_upgrade = package_manifest.get_current_csv(channel)
+        if csv_name_post_upgrade == csv_name_pre_upgrade:
+            log.info(f"CSV is still: {csv_name_post_upgrade}")
+            return False
+        else:
+            log.info(f"CSV now upgraded to: {csv_name_post_upgrade}")
+            return True
+
+    def check_if_images_updated(self, channel, pre_upgrade_images, upgrade_version):
```
ebenahar, line 298: the method name doesn't reflect its return value, which is not bool

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -0,0 +1,376 @@
...
+        """
+        Checks if all images of OCS cluster upgraded
+
+        Args:
+            channel: (str): OCS subscription channel
+            pre_upgrade_images: (dict): Contains all OCS cluster images
+            upgrade_version: (str): version to be upgraded
+
+        Returns:
+            set: Contains full path of OCS cluster old images
```
ebenahar, line 308: is `set` a type?

AviadP: yes

ocs_ci/ocs/ocs_upgrade.py
AviadP: fixed

ocs_ci/ocs/ocs_upgrade.py
AviadP: fixed

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -0,0 +1,359 @@
...
+            upgrade_version: (str): version to be upgraded
+
+        """
+
+        version_change = self.parsed_upgrade_version > self.parsed_version_before_upgrade
+        if version_change:
+            version_config_file = os.path.join(
+                constants.CONF_DIR, 'ocs_version', f'ocs-{upgrade_version}.yaml'
+            )
+            load_config_file(version_config_file)
```
ebenahar, line 165: what happens else? Should we throw an exception or at least log a warning?

AviadP: will add a warning

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -0,0 +1,363 @@
...
+        upgrade_version (packaging.version.Version): version of OCS
+
+    """
+    namespace = config.ENV_DATA['cluster_namespace']
+    number_of_worker_nodes = len(get_typed_nodes())
+    storage_cluster = StorageCluster(
+        resource_name=config.ENV_DATA['storage_cluster_name'],
+        namespace=namespace
+    )
+    osd_count = (
```
petr-balogh, line 88: This was probably missed in some of your rebase but the logic here were changed to use the function `get_osd_count` see here: https://github.com/red-hat-storage/ocs-ci/commit/4132546cd260a80aa85b335299ee8038d0dd228f#diff-40ca86338b56ecd56fb458ca2ea36c3b so please make sure we did not loose any content of that commit here, thanks.

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -0,0 +1,363 @@
...
+    )
+    verify_pods_upgraded(
+        old_images, selector=constants.CSI_RBDPLUGIN_PROVISIONER_LABEL,
+        count=2
+    )
+    verify_pods_upgraded(old_images, selector=constants.MON_APP_LABEL, count=3)
+    verify_pods_upgraded(old_images, selector=constants.MGR_APP_LABEL)
+    verify_pods_upgraded(
+        old_images, selector=constants.OSD_APP_LABEL, count=osd_count,
+        timeout=1800,
```
petr-balogh, line 120: This timeout should be counted dynamically now based on OSD count, see here:  https://github.com/red-hat-storage/ocs-ci/commit/0374fbe43436018e2772feed1c188b250f175e8d#diff-40ca86338b56ecd56fb458ca2ea36c3b

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -0,0 +1,363 @@
...
+        version_change = self.parsed_upgrade_version > self.parsed_version_before_upgrade
+        if version_change:
+            version_config_file = os.path.join(
+                constants.CONF_DIR, 'ocs_version', f'ocs-{upgrade_version}.yaml'
+            )
+            load_config_file(version_config_file)
+        else:
+            log.warning(f"Upgrade version {upgrade_version} is not higher "
+                        f"than old version: {self.version_before_upgrade},"
+                        f" config file will not be loaded")
```
petr-balogh, line 169:
```suggestion
            log.warning(
                f"Upgrade version {upgrade_version} is not higher than old version:"
                f" {self.version_before_upgrade}, config file will not be loaded"
                
            )
```

ocs_ci/ocs/ocs_upgrade.py
petr-balogh: This is expected as if it's upgrade between Z stream the X.Y version will not change hence no need to reload config for specific version like between 4.3 and 4.4 . So not sure if it should be warning level as nothing wrong when upgrade from 4.4.0 and 4.4.1 for example as X.Y version will not change.

ocs_ci/ocs/ocs_upgrade.py
AviadP: fixed

ocs_ci/ocs/ocs_upgrade.py
AviadP: fixed

ocs_ci/ocs/ocs_upgrade.py
AviadP: fixed

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -0,0 +1,387 @@
...
+
+    """
+    def __init__(self, namespace, version_before_upgrade, ocs_registry_image, upgrade_in_current_source):
+        self.namespace = namespace
+        self.version_before_upgrade = version_before_upgrade
+        self.ocs_registry_image = ocs_registry_image
+        self.upgrade_in_current_source = upgrade_in_current_source
+
+    @property
+    def get_version_before_upgrade(self):
```
fbalak, line 136: As this is a property and not a getter, this should not be named `get_version_before_upgrade` but `version_before_upgrade`. It is not called like a function but as an attribute so I don't think `get_` prefix should be used. It could look like this:
```
    @property
    def version_before_upgrade(self):
        return self._version_before_upgrade
```
The same for property `get_ocs_registry_image`.

PR #2297, opened 2020-06-15
Update: tests/e2e/workloads/couchbase/test_couchbase.py

shreekarSS:
- Fixed wrong polarion id

### DIFF
--- a/tests/e2e/workloads/couchbase/test_couchbase.py
+++ b/tests/e2e/workloads/couchbase/test_couchbase.py
@@ -19,7 +19,7 @@ def couchbase(request):
 
 
 @workloads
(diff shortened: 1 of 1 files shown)


tests/e2e/workloads/couchbase/test_couchbase.py
```diff
@@ -19,7 +19,7 @@ def teardown():
 
 
 @workloads
-@pytest.mark.polarion_id("OCS-807")
+@pytest.mark.polarion_id("OCS-785")
```
PrasadDesala, line 22: This test case is more of talking about performance and in this test we are not doing any steps related to perf.
Do you think the test case ID has to be changed to 784?

PR #2300, opened 2020-06-15
Ignore rook-ceph-drain-canary pod leftover for the test cases which d…

### LINKED ISSUE #2209: Consider rook-ceph-drain-canary pod re-spin as expected if the test case deletes OSD pod.
rook-ceph-drain-canary pod re-spin should be considered as expected if the test case deletes OSD pod.
It is possible that rook-ceph-drain-canary pod's deployment was removed when the OSD was down, and recreated when the OSD came back up. So the rook-ceph-drain-canary pod, which is present on the pod where the killed osd resides, will respin with a different name.

Currently this is not handled in environment check so it will throw error due to this. We will need to fix this in ocs-ci/ocs_ci/utility/environment_check.py. 

Applicable for these tests -

ocs-ci/tests/manage/pv_serices/test_pvc_disruptive.py
ocs-ci/tests/manage/pv_services/test_resource_deletion_during_pod_pvc_deletion.py
ocs-ci/tests/manage/pv_services/test_resource_deletion_during_pvc_pod_creation_and_io.py
ocs-ci/tests/manage/pv_services/test_resource_deletion_during_pvc_pod_deletion_and_io.py

jilju:
…eletes OSD pod

Closes #2209 

### DIFF
--- a/ocs_ci/utility/environment_check.py
+++ b/ocs_ci/utility/environment_check.py
@@ -75,7 +75,7 @@ def compare_dicts(before, after):
 
 
 def assign_get_values(
-    env_status_dict, key, kind=None, get_environment_status=None
+    env_status_dict, key, kind=None, exclude_labels=None
 ):
     """
     Assigning kind status into env_status_dict
--- a/ocs_ci/utility/environment_check.py
+++ b/ocs_ci/utility/environment_check.py
@@ -110,7 +110,7 @@ def assign_get_values(
     env_status_dict[key] = items_filtered
 
 
-def get_environment_status(env_dict, exclude_labels):
+def get_environment_status(env_dict, exclude_labels=None):
     """
     Get the environment status per kind in KINDS and save it in a dictionary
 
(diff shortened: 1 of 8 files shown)


ocs_ci/utility/environment_check.py
```diff
@@ -112,35 +110,42 @@ def assign_get_values(
...
+
+    Args:
+        exclude_labels (list): App labels to ignore leftovers
     """
-    get_environment_status(ENV_STATUS_PRE)
+    get_environment_status(ENV_STATUS_PRE, exclude_labels=exclude_labels)
 
 
-def get_status_after_execution():
+def get_status_after_execution(exclude_labels):
```
ebenahar, line 136: I think `exclude_labels` should be set to `None` so existing usage will not break

jilju: Thanks !.
The existing usage will not break because the function is called only from environment_checker where exclude_labels contain at least one value and it is passed to this function.
But I will set the default value to None because the function get_status_after_execution should be callable without exclude_labels param.

jilju: Done

ebenahar: nice, thanks

PR #2305, opened 2020-06-16
WIP: Update test_pvc_resize to align with OCS test case

jilju:
Updated and renamed test case test_pvc_resize to align with an OCS test case in polarion.
Added OCS-302 test marker.


tests/manage/pv_services/test_pvc_resize.py
```diff
@@ -1,64 +0,0 @@
-import logging
```
ebenahar, line 1: Don't we still want to test the happy path of PV resize?

jilju: I feel it is not important to keep a test case which just expands the PVC.
We will add more tests around resize feature so the resize part will be automatically covered in all those tests.

Planning to move this test case to a separate directory tests/manage/pv_services/pvc_resize.

ebenahar: ok. Please consider that the happy path will be tested as part of tier1 so we will need this basic test to exist

Issue #2306, opened 2020-06-16
Fix few issues for enable run on IBM cloud


petr-balogh:
IBM cloud folks contacted me to help them with execution on IBM cloud and there are several issues.

- [x] Missing config file with platform specification
- [x] Ceph toolbox is not created when manual setup
- [x] Issue with check_local_volume function which throws exception instead of returning boolean value
- [x] Do not create deployer object if running tests as it's not needed and failing with undefined platform
- [x] `oc get clusterversion version -o yaml` missing `channel` in `spec` and failing all test cases, we should report `unknow` instead of failing whole execution
- [x] Add config for usage of live must-gather image for OCS

PR #2307, opened 2020-06-16
Fix few issues for IBM cloud execution

petr-balogh:
Fixes: #2306

### DIFF
--- a/ocs_ci/utility/localstorage.py
+++ b/ocs_ci/utility/localstorage.py
@@ -1,20 +1,20 @@
 """
 This module contains local-storage related methods
 """
+import json
 import logging
 import os
 import shutil
 import yaml
-import json
 
-from ocs_ci.utility.utils import clone_repo, get_ocp_version, run_cmd
+from ocs_ci.framework import config
+from ocs_ci.ocs import constants, defaults
+from ocs_ci.ocs.exceptions import CommandFailed
 from ocs_ci.ocs.node import get_typed_nodes
-from ocs_ci.utility.retry import retry
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs.resources import csv
-from ocs_ci.utility import utils
-from ocs_ci.ocs import constants, defaults
-from ocs_ci.framework import config
+from ocs_ci.utility.retry import retry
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -2507,3 +2512,15 @@ def measurement_dir(tmp_path):
         )
         os.mkdir(measurement_dir)
     return measurement_dir
+
+
+@pytest.fixture(scope="session", autouse=True)
+def ceph_toolbox(request):
+    """
+    This fixture initiates ceph toolbox pod for manually created deployment
```
vasukulkarni, line 2520: I thought we had this already?

vasukulkarni: If it doesn't exist it does create it on the fly

ocs_ci/utility/localstorage.py
```diff
@@ -1,20 +1,21 @@
...
+from ocs_ci.ocs import constants, defaults
 from ocs_ci.ocs.node import get_typed_nodes
-from ocs_ci.utility.retry import retry
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs.resources import csv
 from ocs_ci.utility import utils
-from ocs_ci.ocs import constants, defaults
-from ocs_ci.framework import config
+from ocs_ci.utility.retry import retry
+from ocs_ci.utility.utils import clone_repo, run_cmd
```
clacroix12, line 18: We are already importing the entire `utils` module on line 16. Can we not use `utils.clone_repo` and `utils.run_cmd` or change the `utils.<utility>` calls to be explicitly imported? We probably shouldn't be doing both.

tests/conftest.py
petr-balogh: @vasukulkarni we didn't have it.  We do deploy cep_toolbox as part of our deployment but if the cluster was deployed manually and not via ocs-ci then no one deployed ceph toolbox pod and this fixture taking care of it on such clusters.

ocs_ci/utility/localstorage.py
petr-balogh: done

petr-balogh:
After merge of some PR there was conflict in conftest.py as my new fixture is added at the end of file.

I consider this as verified as no code change done from last run.

PR #2314, opened 2020-06-16
Enable Noobaa auto scale post deployment

ebenahar:


ocs_ci/framework/conf/default_config.yaml
```diff
@@ -68,6 +68,7 @@ DEPLOYMENT:
   # Following options are defaults for deployment from live content
   live_deployment: False
   preserve_bootstrap_node: False
+  noobaa_endpoints: 1
```
fbalak, line 71: It would be good to add a docstring or move it a little as docstring above this parameter is:
```
# Following options are defaults for deployment from live content
```
and this is not an option specific for live content deployment.

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -498,3 +498,17 @@ def get_all_storageclass():
...
+def change_noobaa_endpoints_count(nb_eps):
+    """
+    Scale up or down the number of maximum NooBaa emdpoints
+
+    Args:
+        nb_eps (int): The number of required Noobaa endpoints
+
+    """
+    log.info(f"Scaling up Noobaa endpoints to a maximum of {nb_eps}")
+    params = f'{{"spec":{{"endpoints":{{"maxCount":{nb_eps},"minCount":1}}}}}}'
```
fbalak, line 512: If `minCount` is set to 1 then wouldn't the number of endpoints vary during testing (it might be noticeable when set e.g. to `5`)? Is this expected behaviour for setting `noobaa_endpoints` options?

ebenahar:
> LGTM, I would just like to clarify behaviour when `minCount` is set to `1`.

Hi @fbalak, we want minCount to be 1, as in case no of load, 1 endpoint is enough.

ocs_ci/framework/conf/default_config.yaml
ebenahar: ack, will add

ocs_ci/ocs/resources/storage_cluster.py
ebenahar: yes, the number could vary during testing. We will do an experiment of triggering the scale up of endpoints with loading the endpoint and see what happens during the tiers executions.

fbalak: Ok

tests/ecosystem/deployment/test_deployment.py
```diff
@@ -23,6 +23,11 @@ def test_deployment(pvc_factory, pod_factory):
                 'ocs_registry_image'
             )
             ocs_install_verification(ocs_registry_image=ocs_registry_image)
+
+            nb_eps = config.DEPLOYMENT.get('noobaa_endpoints')
+            if nb_eps > 1:
```
RazTamir, line 28: maybe better to change:
'>' to '!=' in case we would like to test negative?

ebenahar: this is in the deployment test. If we will want to test negative, it could be done in other tests

petr-balogh, line 28: Hey @ebenahar ,

Sorry to be late here, but I don't think it should be part of this test_deployment test as it's mainly should care about deployment verification and should not do any extra changes in the deployment.

If this change should be part of our deployment this could should be called from deploy_ocs function I think:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/deployment.py#L439

ebenahar: Hi @petr-balogh thanks for the input.
I consider putting this change in the deploy_ocs function function but the problem with that is that changes will be required in https://github.com/red-hat-storage/ocs-ci/blob/ea0417e5aad2d4c27328b34f23f830c06d55a9f4/ocs_ci/ocs/resources/storage_cluster.py#L39 as we are expecting a fixed number of Noobaa resources and if we change the max of Noobaa endpoints before this verification without somehow adjusting it, the verification will fail. 

I will try to think how to solve this issue and send a PR

petr-balogh: @ebenahar I see you are patching this:
https://github.com/red-hat-storage/ocs-ci/pull/2314/files#diff-9829dc4eedf17bed5719295e70c7fa84R512
``` params = f'{{"spec":{{"endpoints":{{"maxCount":{nb_eps},"minCount":1}}}}}}'```

So by default if possible - no value should be changed and we should have default behavior, only if specific config value is provided, than you can compare the value with what is in the config?

Otherwise I guess you can read this value dynamically from that object instead of for verification to read it from config?
From value: `maxCount`. IIUC ?

WDYT?

ebenahar: In PR https://github.com/red-hat-storage/ocs-ci/pull/2366 I changed the default to be max of 2 endpoints.

Out of the box, `maxCount` is 1. this condition of checking the value from the config is in order to change the NooBaa CR.

ebenahar: Reported https://github.com/red-hat-storage/ocs-ci/issues/2376 to track the change needed here

PR #2316, opened 2020-06-17
Enhanced get_ceph_cluster_iops() function

ramkiperiy:
Enhanced function to get fetch output from dict than string
Updated function get_rebalance_status() similarly to get ceph status output in dict than string
Updated log.warning in validate_osd_utilization()


ramkiperiy:
Jenkins Validation passed for most of the tier1 tests and there is failure in 2 TCs which related to removal of PVC from backend.

Issue #2324, opened 2020-06-17
local-storage-operator installation failed over OCP 4.5 and OCP 4.6  when used channel 4.5 | 4.6


prsurve:
creating temp yaml file with local-storage-operator data:
 [{'apiVersion': 'v1', 'kind': 'Namespace', 'metadata': {'name': 'local-storage'}, 'spec': {}}, {'apiVersion': 'operators.coreos.com/v1', 'kind': 'OperatorGroup', 'metadata': {'annotations': {'olm.providedAPIs': 'LocalVolume.v1.local.storage.openshift.io'}, 'name': 'local-storage', 'namespace': 'local-storage'}, 'spec': {'targetNamespaces': ['local-storage']}}, {'apiVersion': 'operators.coreos.com/v1alpha1', 'kind': 'Subscription', 'metadata': {'name': 'local-storage-operator', 'namespace': 'local-storage'}, 'spec': {'channel': '4.5', 'installPlanApproval': 'Automatic', 'name': 'local-storage-operator', 'source': 'redhat-operators', 'sourceNamespace': 'openshift-marketplace'}}]
23:27:54 - MainThread - ocs_ci.utility.templating - INFO - ['apiVersion\n--- kind\n--- metadata\n--- spec\n...\n', 'apiVersion\n--- kind\n--- metadata\n--- spec\n...\n', 'apiVersion\n--- kind\n--- metadata\n--- spec\n...\n']

LSO installation with channel 4.5 will fail as 4.5 is not yet GA so 4.5 channel is not yet opened 
as a WA we need to use n-1 as channel for LSO deployment

`Current Opened Channel`
![Screenshot (56)](https://user-images.githubusercontent.com/32088367/84935139-ed201400-b0f5-11ea-8f9f-9a003158de7e.png)

https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/deployment.py#L675-L677

prsurve:
@petr-balogh @clacroix12 fyi

clacroix12:
This will be addressed in my changes for #2242

clacroix12:
A workaround exists for this [here](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/localstorage.py#L172-L174). When the 4.5 channel exists for the LSO we should remove this workaround.

ebenahar:
Hi @clacroix12 @prsurve don't we need a BZ for this issue?

[...]

PR #2326, opened 2020-06-18
Add Support to install OCS over BM

prsurve:


prsurve:
OCS installation will fail because of  issue #2324

tests/helpers.py
```diff
@@ -2330,3 +2330,24 @@ def retrieve_default_ingress_crt():
 
     with open(constants.DEFAULT_INGRESS_CRT_LOCAL_PATH, 'w') as crtfile:
         crtfile.write(decoded_crt)
+
+
+def get_pv_size(storageclass=None):
```
ebenahar, line 2335: please place this function in a file under ocs_ci/ 
maybe under ocs_ci/ocs/resources/pvc.py

Also, I see this function returns a list with the sizes of all the storageclass PVs.
What's the use case of such functionality? How about changing this function to return the size of a single PV and you'll create a list of the PVs outside. This will make this function more generic and usable

prsurve: Here I'm expecting to return PV size so that I can make use of PV size in storagecluster cr while creating it for BM env. As per doc, we are requesting customers to provide PV size as requested storage size that will help to avoid picking up lower size PV for creating OSD.

prsurve: in some cases might happen that all pv size are different so while requestion storage size in storage cluster cr we need to provide lower pv size volume then only all osd will come up if we provide big pv size small size pv will not be included and that might lead osd stuck in pending case

prsurve:
Verification job:- https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/9161

prsurve:
Verification job:- https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/9222/

prsurve:
Tier1 job:- https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/9233/

PR #2330, opened 2020-06-18
Fix #2329 with get provider name

### LINKED ISSUE #2329: get_provider faild on BM
Since on BM there is no Provider, the function failed and cause tests that use this function to failed.

Avilir:


Avilir:
Tested and verified locally on BM environment.

PR #2345, opened 2020-06-19
Pods check after upgrade

fbalak:


ocs_ci/ocs/resources/pod.py
```diff
@@ -1246,6 +1246,55 @@ def download_file_from_pod(pod_name, remotepath, localpath, namespace=None):
     run_cmd(cmd)
 
 
+def verify_pods_status(
```
ebenahar, line 1249: I think that similar implementation already exists in tests.manage.z_cluster.nodes.helpers.wait_for_all_pods
Could you please check and if so, move it to pod.py
Also, note that some pods, like '-1-deploy' and 'ocs-deviceset', are expected to be in `Completed` state

fbalak: Thank you for bringing this to my notice. I looked into `tests.manage.z_cluster.nodes.helpers.wait_for_all_pods`. There are some differences in our functions:

- `wait_for_all_pods` waits for pods to be in correct state. In `verify_pods_status` function I don't wait for correct status and check that pods are in correct status at the moment.
- `verify_pods_status` can be parametrized to select specific pods based on namespace or selector.
- There is logging for pods that failed the status check with DEBUG information in `verify_pods_status`.

I can move it and use `wait_for_all_pods` but I would rename it to something like `wait_for_storage_pods_status`.

ebenahar: the suggestion of `wait_for_storage_pods_status` sounds good

fbalak: Done

Issue #2250, opened 2020-06-04
Add capacity tests should be disabled over AWS LSO


ebenahar:
All tests that involve add capacity should be disabled from running over AWS LSO i3 instance based deployment, as these instances already use the 2 available storage devices post deployment

anubhav-here:
+1

Issue #2242, opened 2020-06-03
AWS LSO - re-confirm deployment in ocs-ci matches with official deployment documentation


nehaberry:
Official docs for [OCS 4.4](https://access.redhat.com/documentation/en-us/red_hat_openshift_container_storage/4.4/html-single/deploying_openshift_container_storage/index?lb_target=preview#creating-openshift-container-storage-cluster-on-amazon-ec2_rhocs) section **1.2.4. Creating OpenShift Container Storage cluster on Amazon EC2 storage optimized - i3en.2xlarge instance type**

- [ ] 1. ocs-ci uses the storageclassname as "local-block"  but official docs have **localblock** . -  in Step 1 of Procedure

- [ ] 2. Confirm that the storagecluster CR mentioned in the doc matches(almost) with the ocs-ci StorageclusterCR we create, including the CPU and memory limits (if not already)

- [ ] 3. In official docs, as part of deployment itself, we are recommending to use "count:2" for storagedeviceSets, hence both the disks in an instance are used for OSDs. But ocs-ci currently has count:1 

**From docs**
```
  storageDeviceSets:
    - count: 2
```

- [ ] 4. If we use all the disks, add capacity test for AWS LSO should be skipped OR we need to perform node add instead (as officially documented)

Please let me know if any clarification is needed. Thanks

ebenahar:
Once done, let's continue with https://github.com/red-hat-storage/ocs-ci/issues/2250

petr-balogh:
@clacroix12 maybe good if you can take this up as you did all lso related stuff on I3.  Thanks

clacroix12:
There are a couple differences still in the CR we use and the example given in the official doc. The differences being the limits/requests/memory for a few of the resources. I'm not 100% sure if these are something we want to align with so I wanted to see what you think @nehaberry.

Our CR:
```
apiVersion: ocs.openshift.io/v1
kind: StorageCluster
metadata:
  name: ocs-storagecluster
  namespace: openshift-storage
spec:
  manageNodes: false
  monDataDirHostPath: /var/lib/rook
  resources:
    mds:
      Limits: null
      Requests: null
    mgr:
      Limits: null
      Requests: null
    mon:
      Limits: null
      Requests: null
    noobaa-core:
      Limits: null
      Requests: null
    noobaa-db:
      Limits: null
      Requests: null
    rgw:
      Limits: null
      Requests: null
  storageDeviceSets:
  - count: 2
[...]

PR #2350, opened 2020-06-19
AWS LSO alignment with doc

clacroix12:
Fixes: #2242 
Fixes: #2250

### DIFF
--- a/conf/deployment/aws/ipi_1az_rhcos_lso_3m_3w.yaml
+++ b/conf/deployment/aws/ipi_1az_rhcos_lso_3m_3w.yaml
@@ -13,7 +13,7 @@ ENV_DATA:
   worker_replicas: 3
   master_replicas: 3
   worker_instance_type: 'i3en.2xlarge'
-  device_size: 2300  # Value pulled from here: https://bugzilla.redhat.com/show_bug.cgi?id=1812557#c27
+  device_size: 2328
 REPORTING:
   polarion:
     deployment_id: 'OCS-2196'
--- a/tests/e2e/scale/test_scale_osds_fill_75%_reboot_workers.py
+++ b/tests/e2e/scale/test_scale_osds_fill_75%_reboot_workers.py
@@ -15,6 +15,7 @@ from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs import constants, platform_nodes
 from ocs_ci.ocs.resources.pod import wait_for_dc_app_pods_to_reach_running_state
 from ocs_ci.ocs.resources import storage_cluster
+from ocs_ci.framework.pytest_customization.marks import skipif_aws_i3
 from ocs_ci.framework.testlib import scale, E2ETest, ignore_leftovers
 from ocs_ci.utility.utils import ceph_health_check
 
@@ -35,6 +36,7 @@ logger = logging.getLogger(__name__)
         )
     ]
 )
+@skipif_aws_i3
 class TestScaleOSDsRebootNodes(E2ETest):
     """
     Add first set of OSD to a minimum cluster with 50%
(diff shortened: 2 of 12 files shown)


tests/e2e/scale/test_scale_osds_fill_75%_reboot_workers.py
```diff
@@ -35,6 +35,11 @@
         )
     ]
 )
+@pytest.mark.skipif(
```
vasukulkarni, line 38: since we are using it in few places, i think this can move to custom marks.

clacroix12: @vasukulkarni Turns out we already have one actually, `skipif_aws_i3`. I'll use this instead!

clacroix12: Done!

clacroix12:
The PR test failed because of an issue cloning the ocs4-jenkins repo during the destroy phase. The actual deployment was successful and you can also see from the test results that `test_add_capacity` was [skipped](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/2112/testReport/tests.manage.z_cluster.cluster_expansion.test_add_capacity/TestAddCapacity/test_add_capacity/) with the reason `Test will not run on AWS i3`. Marking as verified.

conf/deployment/aws/ipi_1az_rhcos_lso_3m_3w.yaml
```diff
@@ -13,7 +13,7 @@ ENV_DATA:
   worker_replicas: 3
   master_replicas: 3
   worker_instance_type: 'i3en.2xlarge'
-  device_size: 2300  # Value pulled from here: https://bugzilla.redhat.com/show_bug.cgi?id=1812557#c27
+  device_size: 2328
```
petr-balogh, line 16: @clacroix12 I am afraid that this value is getting overwritten by cli parameter in our jenkins job https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/9240 I see storageCluster has 2Ti. http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/jnk-ai3cl33-t1/jnk-ai3cl33-t1_20200629T085616/logs/failed_testcase_ocs_logs_1593421657/deployment_ocs_logs/ocs_must_gather/quay-io-rhceph-dev-ocs-must-gather-sha256-4db6344cc517f57df8ca261314aeba25d5348d3f033a14e90267b75b905efc8a/storagecluster.yaml

clacroix12: Alright, the way I see it we have two options. We can change the behavior of the `OSD_SIZE` param in jenkins to default to an empty string in the case of our LSO trigger jobs. This has the downside of not being caught if someone is deploying a one-off LSO deployment without using the trigger job though. The other way is we can change the default behavior to not set a value in jenkins (default to empty string) and rely on the default in ocs-ci which can be set to 2Ti (overwritten by the config value). WDYT @petr-balogh ?

petr-balogh: We currently do report of this value in our spreadsheet, it it will be empty we will not report that value there I guess. So we need to align the job to report some value like DEFAULT if not specified.

We can for AWS LSO jobs to use empty value in param - or we need to define this size 2328 in our OSD_SIZE parameter and propagate this value from LSO jobs as default one.   I am for second option to have it in the report as well.

clacroix12: @petr-balogh https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/-/merge_requests/294

Issue #2351, opened 2020-06-19
Freeze requirement versions


clacroix12:
Now that we are cutting releases of ocs-ci we should have specific versions set for our dependencies. This will ensure that people who check out a release will install the exact environment that was used when the release was cut.

mbukatov:
This would also require fixing other releng issues with the ocs-ci.

PR #2352, opened 2020-06-19
Freeze requirement versions

clacroix12:
Fixes: #2351 


clacroix12:
I tested pip installation of all requirements in `setup.py`, `requirements-docs.txt`, and `requirements-dev.txt` into a newly created venv. This confirms that all versions are able to be installed. If functional testing of ocs-ci is required to verify this PR please add the label and I will kick off a test.

Issue #2364, opened 2020-06-23
Fails to get pgsql pods when namespace is switched. ( ie: when not in my-ripsaw namepsace)


parikshithb:
This happens when setting up postgresql, creating benchmark and while validating the pgbench run.

Failure 1:
Executing command: oc --kubeconfig /home/pbyregow/VMware/others/auth/kubeconfig get pod  --selector=app=postgres -o yaml
11:54:29 - MainThread - ocs_ci.ocs.ocp - INFO - status of  at column STATUS - item(s) were [], but we were waiting for all 1 of them to be Running
11:54:29 - MainThread - ocs_ci.ocs.ocp - ERROR - timeout expired: Timed Out: (3600,)
11:54:29 - MainThread - ocs_ci.ocs.ocp - ERROR - Wait for pod resource  at column STATUS to reach desired condition Running failed, last actual status was []

Failure 2: 
E           ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc --kubeconfig /home/pbyregow/Music/VMware/others/auth/kubeconfig get Pod pgbench-1-dbs-client-1-27196c72-hb5w2 -o yaml.
E           Error is Error from server (NotFound): pods "pgbench-1-dbs-client-1-27196c72-hb5w2" not found

ocs_ci/utility/utils.py:461: CommandFailed

PR #2365, opened 2020-06-23
Fix: Fails to get pgsql pods when namespace is switched.

parikshithb:
Fixes: #2364 


parikshithb:
Verified. Failures are not related to this PR.

Issue #2372, opened 2020-06-24
Update of subscription should happen after updating catalogSource


petr-balogh:
In this PR: https://github.com/red-hat-storage/ocs-ci/pull/2294/files

the order got changed and for manual upgrade it makes trouble to wait for subscription plan which will not appear cause of images of catalogSource are not changed yet.

@AviadP  FYI.

PR #2373, opened 2020-06-24
Fix order in upgrade for subscription

petr-balogh:
Fixes: #2372


petr-balogh:
As I triggered already production job which just finishing deployment, I will go ahead and merge this one which should fix the current running execution.

Running here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/9078/console

Issue #2368, opened 2020-06-23
Add root access via securityContext to pods that require it


MeridianExplorer:
OCP 4.5 changes the default pod user from root to an underprivileged one.
This causes two known failures:
1. The `awscli_relay_pod` cannot create the `/cert/` dir required for the workaround. Fixed in #2272 
2. `test_deployment` and other tests that rely on our default NGINX pod fail because it enters a `CrashLoopBackOff` since it required root permissions. When I replaced the image with `nginxinc/nginx-unprivileged` it was not sufficient because we try to install `fio` and `git` on the pod as part of `fio.py` and `jenkins.py`.

Using the solution offered by @petr-balogh allows root access once more, and thus the issue title was changed.

petr-balogh:
Can we fix this by adding this:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/templates/app-pods/rhel-7_7.yaml#L15-L17 to the pod CR?

petr-balogh:
Updated question also in BZ.

MeridianExplorer:
@petr-balogh 
I edited `awscli.yaml`, `fedora_dc.yaml`, `nginx.yaml`, `raw_block_pod.yaml` under `ocs_ci/templates/app-pods`, and `pod.yaml` under `CSI/cephfs` and `CCSI/rbd` - `test_deployment` passed successfully.

petr-balogh:
Cool @MeridianExplorer , do you heve PR you can link here?   I think that as quick fix for our CI we should go with this fix and then see status in the BZ and think of better change. But if we need root access in those pods I think it's only one solution if they did some different permission as default one in OCP 4.5.  Thanks

MeridianExplorer:
The SCC change was reverted, rendering this issue obsolete

PR #2375, opened 2020-06-24
[WIP] Add a securityContext field for regaining root access on app pods

MeridianExplorer:
Fixes: #2368

### DIFF
--- a/ocs_ci/templates/CSI/cephfs/pod.yaml
+++ b/ocs_ci/templates/CSI/cephfs/pod.yaml
@@ -7,6 +7,9 @@ spec:
   containers:
    - name: web-server
      image: nginx
+     securityContext:
+       privileged: true
+       runAsUser: 0
      volumeMounts:
        - name: mypvc
          mountPath: /var/lib/www/html
--- a/ocs_ci/templates/CSI/rbd/pod.yaml
+++ b/ocs_ci/templates/CSI/rbd/pod.yaml
@@ -8,6 +8,9 @@ spec:
   containers:
    - name: web-server
      image: nginx
+     securityContext:
+       privileged: true
+       runAsUser: 0
      volumeMounts:
        - name: mypvc
          mountPath: /var/lib/www/html
--- a/ocs_ci/templates/app-pods/awscli.yaml
+++ b/ocs_ci/templates/app-pods/awscli.yaml
@@ -7,6 +7,9 @@ spec:
   containers:
     - name: awscli
       image: amazon/aws-cli:2.0.13
+      securityContext:
+        privileged: true
+        runAsUser: 0
       # Override the default `aws` entrypoint in order to 
       # allow the pod to run continuously and act as a relay
       command: ['/bin/sh']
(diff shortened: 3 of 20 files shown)


ebenahar:
@MeridianExplorer please also add this change to the yamls under ocs_ci/templates/workloads/fio

MeridianExplorer:
@ebenahar - I added all YAMLs across the project.

petr-balogh:
Updated first comment to Fixes: #issueNumber instead of Works around.  See https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue ;) @MeridianExplorer 

LGTM .

Do we need to open WA issue for this to track this issue still open with the link to the BZ to see if this is the solution or need to revert with other fix?

MeridianExplorer:
Hey @petr-balogh - thanks for the review!
I actually didn't use `fixes` on purpose, since I wanted to use #2368 as a tracker for this solution, which might be a solution, or might be just a workaround. This is why I think we should leave the issue open, and track this solution (and the BZ) with it. What do you think?

petr-balogh:
I think we should open new WA issue with the workaround label to track workaround and let this pr to close one issue we have now. I think we have this approach documented in our docs.

petr-balogh:
https://github.com/red-hat-storage/ocs-ci/blob/master/docs/workarounds.md

MeridianExplorer:
Deployment passed successfully.

After completing 30% of tier1 tests, we ran into a known MCG bug. While our code was trying to save the NooBaa DB dump, the command exited with code 137 and caused the entire test suite to fail. No idea why this would happen. I started another verification job.

vasukulkarni:
we should change the label to verified after tier1 tests are clean.

MeridianExplorer:
@vasukulkarni - Maybe we should run a more thorough check? I thought about adding at least tier1 of performance and workloads.
IIUC, tier1 doesn't include many of the pods changed in this PR, which might lead to unexpected results down the road.

vasukulkarni:
> @vasukulkarni - Maybe we should run a more thorough check? I thought about adding at least tier1 of performance and workloads.
> IIUC, tier1 doesn't include many of the pods changed in this PR, which might lead to unexpected results down the road.

Sounds good to me, e2e should do as well.

MeridianExplorer:
`level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply Terraform: failed to complete the change"`

An unrelated issue that we also hit on Jun 12 - retriggering.

MeridianExplorer:
The one test that seemed to actually fail was `test_workload_rbd_cephfs_10g`.
Apparently, the added securityContext prevented the job from being created successfully.
Thus, I removed all securityContexts that were added to Jobs.

ebenahar:
> @ebenahar - I added all YAMLs across the project.

I don't see changes under ocs_ci/templates/workloads

MeridianExplorer:
@ebenahar - I reverted the change via d3ee063. It caused several failures, and cannot be used as part of those pods.

MeridianExplorer:
Closed due to the NB SCC change being reverted

PR #2383, opened 2020-06-25
Update to check running pods are on same node

jilju:
tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
1. Update to check whether running pods are on same node.
2. Increase replica count to verify  using more pods.

### DIFF
--- a/tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
+++ b/tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
@@ -33,7 +33,7 @@ log = logging.getLogger(__name__)
         )
     ]
 )
-class TestVerifyRwoUsingReplica2DcPod(ManageTest):
+class TestVerifyRwoUsingReplicatedPod(ManageTest):
     """
     This test class consists of tests to verify RWO volume is exclusively
     mounted.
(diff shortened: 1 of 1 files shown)


tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
```diff
@@ -75,40 +73,50 @@ def wait_for_pods_and_verify(self):
         """
         # Wait for pods
         for pods in TimeoutSampler(
-            180, 2, func=pod.get_all_pods, namespace=self.namespace,
+            360, 2, func=pod.get_all_pods, namespace=self.namespace,
             selector=[self.name], selector_label='name'
         ):
-            if len(pods) == 2:
+            if len(pods) == 5:
```
AviadP, line 79: if meaning of this condition is `if len(pods) == replica_count` that defined in another function, plz consider using constant instead

tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
```diff
@@ -75,40 +73,50 @@ def wait_for_pods_and_verify(self):
...
+                    f"Pods which are running are on the same node "
+                    f"{pod_running_node}"
+                )
+            except ResourceWrongStatusException:
+                log.info(
+                    f"Verified: Pod {pod_obj.name} is not in "
+                    f"running state."
+                )
 
     def test_verify_rwo_using_replica_2_dc_pod(self):
```
sidhant-agrawal, line 121: shouldn't it be `test_verify_rwo_using_replica_5_dc_pod`?

tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
jilju: Done

tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
jilju: Modified the test name

ebenahar (approved):

PR #2395, opened 2020-06-26
AMQ fixture and testcase amq workload when spinning ceph pods

Akarsha-rai:
- Created fixture for running amq workload
- AMQ workload test cases : spinning ceph pods and restarting amq pods

### DIFF
--- a/ocs_ci/ocs/amq.py
+++ b/ocs_ci/ocs/amq.py
@@ -9,6 +9,7 @@ import json
 from subprocess import run, CalledProcessError
 from prettytable import PrettyTable
 from threading import Thread
+from queue import Queue
 
 from ocs_ci.ocs.exceptions import (ResourceWrongStatusException, CommandFailed)
 from ocs_ci.ocs.ocp import OCP, switch_to_default_rook_cluster_project
@@ -572,13 +573,14 @@ class AMQ(object):
         log.info(f"Run benchmark and running command {cmd} inside the benchmark pod ")
 
         if run_in_bg:
+            queue = Queue()
             thread = Thread(
                 target=self.run_amq_workload_in_bg, args=(
-                    cmd, benchmark_pod_name, tiller_namespace, timeout
+                    cmd, benchmark_pod_name, tiller_namespace, timeout, queue
                 ))
             thread.start()
             time.sleep(60)
-            return thread
+            return thread, queue
 
         pod_obj = get_pod_obj(name=f"{benchmark_pod_name}-driver", namespace=tiller_namespace)
         result = pod_obj.exec_cmd_on_pod(command=cmd, out_yaml_format=False, timeout=timeout)
@@ -586,7 +588,7 @@ class AMQ(object):
         return result
 
     def run_amq_workload_in_bg(
-        self, command, benchmark_pod_name, tiller_namespace, timeout
+        self, command, benchmark_pod_name, tiller_namespace, timeout, queue
     ):
         """
         Runs amq workload in bg
@@ -595,14 +597,16 @@ class AMQ(object):
              command (str): Command to run on pod
              benchmark_pod_name (str): Pod name
              tiller_namespace (str): Namespace of pod
-             timeout (int)
+             timeout (int): Time to complete the run
+             queue (obj): Queue obj
 
         Returns:
             Thread: A thread of the amq workload execution
 
         """
         pod_obj = get_pod_obj(name=f"{benchmark_pod_name}-driver", namespace=tiller_namespace)
-        return pod_obj.exec_cmd_on_pod(command=command, out_yaml_format=False, timeout=timeout)
+        result = pod_obj.exec_cmd_on_pod(command=command, out_yaml_format=False, timeout=timeout)
+        return queue.put(result)
 
     def validate_amq_benchmark(self, result, amq_workload_yaml, benchmark_pod_name="benchmark"):
         """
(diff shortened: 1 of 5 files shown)


tests/e2e/workloads/amq/test_amq_pod_respin.py
```diff
@@ -0,0 +1,106 @@
...
+    pod = ocp.OCP(kind=constants.POD, namespace=kafka_namespace)
+    pod_obj_list = get_all_pods(namespace=kafka_namespace)
+    for pod_obj in pod_obj_list:
+        pod_obj.delete()
+        assert pod.wait_for_resource(
+            condition='Running', resource_count=len(pod_obj_list), timeout=300
+        )
+
+
+@ignore_leftovers
```
PrasadDesala, line 34: Any reason for ignoring leftovers?

ocs_ci/ocs/amq.py
```diff
@@ -569,11 +571,39 @@ def run_amq_benchmark(
...
+                cmd, benchmark_pod_name, tiller_namespace, timeout
+            )
+            return result
+
         pod_obj = get_pod_obj(name=f"{benchmark_pod_name}-driver", namespace=tiller_namespace)
         result = pod_obj.exec_cmd_on_pod(command=cmd, out_yaml_format=False, timeout=timeout)
 
         return result
 
+    def run_amq_workload_in_bg(
```
PrasadDesala, line 588: The function name says we are running the amq workload in bg. where are we using threading inside the function?

ocs_ci/ocs/amq.py
```diff
@@ -569,11 +571,39 @@ def run_amq_benchmark(
...
+        Runs amq workload in bg
+
+        Args:
+             command (str): Command to run on pod
+             benchmark_pod_name (str): Pod name
+             tiller_namespace (str): Namespace of pod
+             timeout (int): Time to complete the run
+
+        Returns:
+            Thread: A thread of the amq workload execution
```
PrasadDesala, line 601: returning exec_cmd_on_pod which will not return thread. 
Am I missing anything?

tests/conftest.py
```diff
@@ -2519,6 +2520,65 @@ def finalizer():
...
+        )
+
+        return amq, result
+
+    def finalizer():
+        """
+        Clean up
+
+        """
+
```
PrasadDesala, line 2574: Delete empty line

tests/e2e/workloads/amq/test_amq_pod_respin.py
```diff
@@ -0,0 +1,106 @@
...
+from tests.helpers import default_storage_class
+from tests.disruption_helpers import Disruptions
+from ocs_ci.utility import templating
+from ocs_ci.ocs.resources.pod import get_all_pods
+
+
+log = logging.getLogger(__name__)
+
+
+def respin_amq_app_pod(kafka_namespace):
```
PrasadDesala, line 17: Do you think it is a good idea to move this function in amq class so that it can be used in other tests as well?

tests/e2e/workloads/amq/test_amq_pod_respin.py
```diff
@@ -0,0 +1,106 @@
...
+    @pytest.mark.usefixtures(amq_setup.__name__)
+    def test_run_amq_respin_pod(self, pod_name):
+        """
+        Test amq workload when spinning ceph pods
+        and restarting amq pods
+
+        """
+        # Respin relevant pod
+        if pod_name == 'amq':
+            respin_amq_app_pod(kafka_namespace=constants.AMQ_NAMESPACE)
```
PrasadDesala, line 93: You are calling the fixture in setup which may be called only once, how are we sure that the pods are deleted while amq benchmark is running?
We will have to delete the pods while the benchmark is running

Akarsha-rai: No, fixture will be called for each test case. AMQ fixture will create kafka cluster and run amq workloads in bg if we set run_in_bg as True then delete or respin pods. Once the new pod is up and running waits and validates IOs in bg.

tests/e2e/workloads/amq/test_amq_pod_respin.py
Akarsha-rai: Since this function is more specific for one test case so defined here.

PrasadDesala: Ack. If there are new test cases which requires to respin app pod then we can consider moving this function to amy class.

tests/e2e/workloads/amq/test_amq_pod_respin.py
PrasadDesala: Correct, missed that parametrize  part.

ocs_ci/ocs/amq.py
Akarsha-rai: changed accordingly

ocs_ci/ocs/amq.py
Akarsha-rai: No, changed accordingly

tests/e2e/workloads/amq/test_amq_pod_respin.py
Akarsha-rai: yes, Respin pod creates new pod so at the end of the test fails with leftovers

tests/conftest.py
Akarsha-rai: done

Issue #2422, opened 2020-07-02
donot create ceph tool box for OCP ONLY deployments


vavuthu:
Below Job failed in creating cephtool box because skip_ocs yaml is passed

```
tests/ecosystem/deployment/test_deployment.py::test_deployment 
-------------------------------- live log setup --------------------------------
07:31:26 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig get pods -o name
07:31:27 - MainThread - ocs_ci.ocs.utils - INFO - starting ceph toolbox pod
07:31:27 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc patch ocsinitialization ocsinit -n openshift-storage --type json --patch  '[{ "op": "replace", "path": "/spec/enableCephTools", "value": true }]'
07:31:28 - MainThread - ocs_ci.utility.utils - WARNING - Command stderr: error: the server doesn't have a resource type "ocsinitialization"

ERROR                                                                    [100%]---------------------------- live log sessionfinish ----------------------------
07:31:29 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig get csv  -n openshift-storage -o yaml
07:31:29 - MainThread - root - WARNING - No version info found for OCS operator
07:31:29 - MainThread - ocs_ci.utility.utils - INFO - testrun_name: OCS4-5-Downstream-OCP4-5-VSPHERE-UPI-1AZ-RHCOS-3M-3W-deployment
07:31:29 - MainThread - ocs_ci.utility.utils - INFO - Results have been emailed to ['null@redhat.com']


==================================== ERRORS ====================================
______________________ ERROR at setup of test_deployment _______________________

request = <SubRequest 'ceph_toolbox' for <Function test_deployment>>

    @pytest.fixture(scope="session", autouse=True)
    def ceph_toolbox(request):
        """
        This fixture initiates ceph toolbox pod for manually created deployment
        and if it does not already exist.
        """
        deploy = config.RUN['cli_params']['deploy']
        teardown = config.RUN['cli_params'].get('teardown')
        if not deploy and not teardown:
            # Creating toolbox pod
>           setup_ceph_toolbox()

tests/conftest.py:2630: 
[...]

PR #2423, opened 2020-07-02
Skip deployment of tool box for OCP ONLY deployments

vavuthu:
Fixes: #2422 

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2625,6 +2625,7 @@ def ceph_toolbox(request):
     """
     deploy = config.RUN['cli_params']['deploy']
     teardown = config.RUN['cli_params'].get('teardown')
-    if not deploy and not teardown:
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -2625,6 +2625,7 @@ def ceph_toolbox(request):
     """
     deploy = config.RUN['cli_params']['deploy']
     teardown = config.RUN['cli_params'].get('teardown')
-    if not deploy and not teardown:
+    skip_ocs = config.ENV_DATA['skip_ocs_deployment']
+    if not (deploy or teardown or skip_ocs):
```
petr-balogh, line 2629: Just thinking, if we do deploy of ocp with skip ocs deploy I guess we have condition satisfied with is not deployment, why do we need this second one?

vavuthu: Once install stage is completed and As part test stage --deploy is not passed

PR #2426, opened 2020-07-02
Changes for test disk failures to support on Azure platform

kesavanvt:
**Description:**
This PR provides the modification for the below tests in tests/manage/z_cluster/nodes/test_disk_failures.py to run on Azure platform

- test_detach_attach_worker_volume 
- test_detach_attach_2_data_volumes 

### DIFF
--- a/ocs_ci/ocs/platform_nodes.py
+++ b/ocs_ci/ocs/platform_nodes.py
@@ -1668,42 +1668,89 @@ class AZURENodes(NodesBase):
             "Start nodes functionality is not implemented"
         )
 
-    def restart_nodes(self, nodes, wait=True):
-        raise NotImplementedError(
-            "Restart nodes functionality is not implemented"
-        )
+    def restart_nodes(self, nodes, timeout=540, wait=True):
+        """
+        Restart Azure vm instances
+
+        Args:
+            nodes (list): The OCS objects of the nodes / Azure Vm instance
+            wait (bool): True if need to wait till the restarted node reaches
+                READY state. False otherwise
+            timeout (int): time in seconds to wait for node to reach 'not ready' state,
+                and 'ready' state.
+
+        """
+        if not nodes:
+            logger.error("No nodes found for restarting")
+            raise ValueError
+        node_names = [n.name for n in nodes]
+        for node_name in node_names:
+            self.azure.restart_az_vm_instance(node_name)
+
+        if wait:
+            """
+            When reboot is initiated on an instance from the Azure, the
+            instance stays at "Running" state throughout the reboot operation.
+
+            Once the OCP node detects that the node is not reachable then the
+            node reaches status NotReady.
+            When the reboot operation is completed and the instance is
+            reachable the OCP node reaches status Ready.
+            """
+            logger.info(
+                f"Waiting for nodes: {node_names} to reach not ready state"
+            )
+            wait_for_nodes_status(
+                node_names=node_names, status=constants.NODE_NOT_READY,
+                timeout=timeout
+            )
+            logger.info(
+                f"Waiting for nodes: {node_names} to reach ready state"
+            )
+            wait_for_nodes_status(
+                node_names=node_names, status=constants.NODE_READY,
+                timeout=timeout
+            )
 
     def get_data_volumes(self):
         """
-        Get the data Azure volume name
+        Get the data Azure disk objects
 
         Returns:
-            list: azure volume names
+            list: azure disk objects
 
         """
         pvs = get_deviceset_pvs()
-        return azure_utils.get_data_volumes(pvs)
+        return self.azure.get_data_volumes(pvs)
 
     def get_node_by_attached_volume(self, volume):
         """
-        Get the Azure Vm instance that has the volume attached to
+        Get node OCS object of the Azure vm instance that has the volume attached to
 
         Args:
-            volume (str): The volume to get the Azure Vm according to
+            volume (Disk): The disk object to get the Azure Vm according to
 
         Returns:
-            Vm: An Azure Vm instance
+            OCS: The OCS object of the Azure Vm instance
 
         """
-        return self.azure.get_node_by_attached_volume(volume)
+        vm = self.azure.get_node_by_attached_volume(volume)
+        all_nodes = get_node_objs()
+        nodes = [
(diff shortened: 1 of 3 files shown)


ocs_ci/ocs/platform_nodes.py
```diff
@@ -1716,9 +1757,10 @@ def attach_volume(self, volume, node):
         )
 
     def wait_for_volume_attach(self, volume):
-        raise NotImplementedError(
-            "Wait for volume attach functionality is not implemented"
-        )
+        # TODO: To check whether the given volume has been attached for the given timeout
+        # TODO: replace time.sleep with the attach check
+        time.sleep(300)
```
ebenahar, line 1762: Please create an issue to track the change

kesavanvt: Now the TODOs have been implemented!

ebenahar:
hi @kesavanvt @mbukatov do we have an Azure cluster we can use in order to verify this PR?

kesavanvt:
> hi @kesavanvt @mbukatov do we have an Azure cluster we can use in order to verify this PR?

Hi Elad, The cluster is not up right now, i can spin it up shortly. 
**Note** : Before running we need to populate the conf/default_config.yaml for the Azure credentials

ocs_ci/utility/azure_utils.py
```diff
@@ -71,23 +93,80 @@ def get_node_by_attached_volume(self, volume):
...
+
+        Args:
+            deviceset_pvs (list): PVC objects of the deviceset PVs
+
+        Returns:
+            list: Azure Vm disk objects
+
+        """
+        volume_names = [pv.get().get('spec').get('azureDisk').get('diskName') for pv in deviceset_pvs]
+        return [self.compute_client.disks.get(self._resourcegroup, volume_name) for volume_name in volume_names]
```
mbukatov, line 159: What is data type of `self.compute_client.disks` object?

Could you replace `.get('foo')` calls with `['foo']` when get is called on a dictionary? In case of a missing entry, it will raise an exception anyway, but exception caused by calling get() on a none object is less useful compared to missing key exception.

ocs_ci/ocs/platform_nodes.py
```diff
@@ -1668,42 +1668,91 @@ def start_nodes(self, nodes):
...
+
+        Args:
+            nodes (list): The OCS objects of the nodes / Azure Vm instance
+            wait (bool): True if need to wait till the restarted node reaches
+                READY state. False otherwise
+            timeout (int): time in seconds to wait for node to reach 'not ready' state,
+                and 'ready' state.
+
+        """
+        vms = [n.name for n in nodes]
```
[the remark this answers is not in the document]
kesavanvt: Thanks, dropped the nodes_names and renamed vms to node_names.

ocs_ci/ocs/platform_nodes.py
```diff
@@ -1668,42 +1668,91 @@ def start_nodes(self, nodes):
...
+        assert vms, (
+            f"Failed to get the Azure Vm instances for "
+            f"nodes {[n.name for n in nodes]}"
```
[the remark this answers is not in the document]
kesavanvt: Thanks for pointing it out Martin,

- Dropping the below statements (missed to remove this earlier), as we are dealing with node object and not the azure vm objects in this method

```
        assert vms, (
            f"Failed to get the Azure Vm instances for "
            f"nodes {[n.name for n in nodes]}"
```

- Included the check on nodes variable and raised ValueError in case of empty list being passed

- I don't think any exception will occur when on accessing name attribute of node object, so check for this is avoided

ocs_ci/utility/azure_utils.py
kesavanvt: **type of self.compute_client.disks :** `<azure.mgmt.compute.v2019_11_01.operations._disks_operations.DisksOperations object at 0x7f101b058c10>`

- Used index for accessing the dictionary instead of `.get('foo')`

ocs_ci/ocs/platform_nodes.py
```diff
@@ -1668,42 +1668,89 @@ def start_nodes(self, nodes):
             "Start nodes functionality is not implemented"
         )
 
-    def restart_nodes(self, nodes, wait=True):
-        raise NotImplementedError(
-            "Restart nodes functionality is not implemented"
-        )
+    def restart_nodes(self, nodes, timeout=540, wait=True):
```
kesavanvt, line 1671: while restarting 2 nodes, one node tends to take longer to reach ready state (takes around 7 to 8 mins sometimes). Thus provided the timeout as 540 secs

kesavanvt:
@ebenahar Attaching the logs for the test carried out 
[test_disk_failures_logs.zip](https://github.com/red-hat-storage/ocs-ci/files/4889975/test_disk_failures_logs.zip)

ocs_ci/utility/azure_utils.py
```diff
@@ -71,23 +93,81 @@ def get_node_by_attached_volume(self, volume):
...
+        result.wait()
+        for x in range(timeout):
+            disk_state = self.get_disk_state(volume.name)
+            logger.debug(
+                "Volume id: %s has status: %s", volume.name,
+                disk_state
+            )
+            if disk_state == 'Unattached':
+                break
+            if x == timeout - 1:
```
mbukatov, line 128: I would rather use more forgiving check, like `x >= timeout - 1`, so that if last get disk state operation takes too long, we don't end up in an endless loop.

kesavanvt: `timeout` is the range value in the for loop.

- x iterates from 0 to `timeout-1` and x exceeding `timeout-1` is not possible
- At the end of each iteration `time.sleep(1)` is used to sleep for one second.

ebenahar: please consider using ocs_ci.utility.utils.TimeoutSampler

mbukatov: ah, ok - you are just checking that the loop is in it's last iteration

kesavanvt: @ebenahar Thanks, I have replaced with TimeoutSampler wherever possible.

Issue #1861, opened 2020-04-17
Test ceph *full ratio values


mbukatov:
To make sure that ceph *full ration values are within expectations, we should consider test case which check it's values.

It would basically cover BZ 1775432.

PR #2429, opened 2020-07-02
Test to check ceph values on the cluster with the default values

PrasadDesala:
Fixes: #1861 
Corrective measure for BZ 1805674

### DIFF
--- a/tests/manage/z_cluster/test_ceph_default_values_check.py
+++ b/tests/manage/z_cluster/test_ceph_default_values_check.py
@@ -19,21 +19,31 @@ class TestCephDefaultValuesCheck(ManageTest):
 
         """
         # The default ceph osd full ratio values
-        osd_full_ratios = {
+        expected_full_ratios = {
             'full_ratio': 0.85,
             'backfillfull_ratio': 0.8,
             'nearfull_ratio': 0.75
         }
+        actual_full_ratios = {}
         ct_pod = pod.get_ceph_tools_pod()
         log.info("Checking the values of ceph osd full ratios in osd map")
         osd_dump_dict = ct_pod.exec_ceph_cmd('ceph osd dump')
-        for ratio_parm, value in osd_full_ratios.items():
+        for ratio_parm, value in expected_full_ratios.items():
             ratio_value = osd_dump_dict.get(ratio_parm)
-            assert float(round(ratio_value, 2)) == value, (
-                f"Cluster {ratio_parm} value is {ratio_value:.2f} NOT "
-                f"matching the default value {value}"
-            )
-        log.info("Ceph osd full ratio in osd map matches the default values")
+            actual_full_ratios.update({ratio_parm:ratio_value})
+            if not float(round(ratio_value, 2)) == value:
+                log.error(
+                    f"Actual {ratio_parm} value is {ratio_value:.2f} NOT "
+                    f"matching the expected value {value}"
+                )
+        assert expected_full_ratios == actual_full_ratios, (
+            f"Actual {actual_full_ratios} values does not match "
+            f"expected full ratio values {expected_full_ratios}"
+        )
+        log.info(
+            f"Actual full ratio {actual_full_ratios} values MATCHES expected "
+            f"full ratio values {expected_full_ratios}"
+        )
 
         # Check if the osd full ratios satisfies condition
         #  "nearfull < backfillfull < full"
--- a/tests/manage/z_cluster/test_ceph_default_values_check.py
+++ b/tests/manage/z_cluster/test_ceph_default_values_check.py
@@ -27,33 +27,35 @@ class TestCephDefaultValuesCheck(ManageTest):
         ct_pod = pod.get_ceph_tools_pod()
         log.info("Checking the values of ceph osd full ratios in osd map")
         osd_dump_dict = ct_pod.exec_ceph_cmd('ceph osd dump')
-        for full_ratio in osd_full_ratios:
-            ratio_value = osd_dump_dict.get(full_ratio)
-            assert float(
-                round(ratio_value, 2)
-            ) == osd_full_ratios[full_ratio], (
-                f" {full_ratio} value is {ratio_value} NOT matching the "
-                f"default value {osd_full_ratios[full_ratio]}"
+        for ratio_parm, value in osd_full_ratios.items():
+            ratio_value = osd_dump_dict.get(ratio_parm)
+            assert float(round(ratio_value, 2)) == value, (
+                f"Cluster {ratio_parm} value is {ratio_value:.2f} NOT "
+                f"matching the default value {value}"
             )
         log.info("Ceph osd full ratio in osd map matches the default values")
 
         # Check if the osd full ratios satisfies condition
         #  "nearfull < backfillfull < full"
         assert (
-            osd_full_ratios[
+            osd_dump_dict[
                 'nearfull_ratio'
-            ] < osd_full_ratios[
+            ] < osd_dump_dict[
                 'backfillfull_ratio'
-            ] < osd_full_ratios[
+            ] < osd_dump_dict[
                 'full_ratio'
             ]
         ), (
             "osd full ratio values does not satisfy condition "
-            "nearfull < backfillfull < full"
+            f"{osd_dump_dict['nearfull_ratio']:.2f} < "
+            f"{osd_dump_dict['backfillfull_ratio']:.2f} < "
+            f"{osd_dump_dict['full_ratio']:.2f}"
         )
         log.info(
             "osd full ratio values satisfies condition "
-            "nearfull < backfillfull < full"
+            f"{osd_dump_dict['nearfull_ratio']:.2f} < "
+            f"{osd_dump_dict['backfillfull_ratio']:.2f} < "
+            f"{osd_dump_dict['full_ratio']:.2f}"
         )
 
         # Check if PG balancer is active
--- a/tests/manage/z_cluster/test_ceph_default_values_check.py
+++ b/tests/manage/z_cluster/test_ceph_default_values_check.py
@@ -30,7 +30,7 @@ class TestCephDefaultValuesCheck(ManageTest):
         for full_ratio in osd_full_ratios:
             ratio_value = osd_dump_dict.get(full_ratio)
             assert float(
-                round(ratio_value,2)
+                round(ratio_value, 2)
             ) == osd_full_ratios[full_ratio], (
                 f" {full_ratio} value is {ratio_value} NOT matching the "
                 f"default value {osd_full_ratios[full_ratio]}"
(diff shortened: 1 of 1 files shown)


tests/manage/z_cluster/test_ceph_default_values_check.py
```diff
@@ -0,0 +1,60 @@
...
+        log.info("Checking the values of ceph osd full ratios in osd map")
+        osd_dump_dict = ct_pod.exec_ceph_cmd('ceph osd dump')
+        for full_ratio in osd_full_ratios:
+            ratio_value = osd_dump_dict.get(full_ratio)
+            assert float(
+                round(ratio_value, 2)
+            ) == osd_full_ratios[full_ratio], (
+                f" {full_ratio} value is {ratio_value} NOT matching the "
+                f"default value {osd_full_ratios[full_ratio]}"
+            )
```
jilju, line 37:
```suggestion
         for ratio_parm, value in osd_full_ratios.items():
             ratio_value = osd_dump_dict.get(ratio_parm)
             assert float(round(ratio_value, 2)) == value, (
                 f" {ratio_parm} value is {ratio_value} NOT matching the "
                 f"default value {value}"
    )
```

tests/manage/z_cluster/test_ceph_default_values_check.py
```diff
@@ -0,0 +1,60 @@
...
+        log.info("Ceph osd full ratio in osd map matches the default values")
+
+        # Check if the osd full ratios satisfies condition
+        #  "nearfull < backfillfull < full"
+        assert (
+            osd_full_ratios[
```
jilju, line 43: Ckeck in osd_dump_dict ?
```suggestion
            osd_dump_dict[
```

tests/manage/z_cluster/test_ceph_default_values_check.py
```diff
@@ -0,0 +1,60 @@
...
+                'nearfull_ratio'
+            ] < osd_full_ratios[
+                'backfillfull_ratio'
+            ] < osd_full_ratios[
+                'full_ratio'
+            ]
+        ), (
+            "osd full ratio values does not satisfy condition "
```
jilju, line 51: Good to add actual values to get it from error message itself.

tests/manage/z_cluster/test_ceph_default_values_check.py
PrasadDesala: Done

tests/manage/z_cluster/test_ceph_default_values_check.py
PrasadDesala: Once the default values check phase is completed both osd_dump_dict and osd_full_ratios values will remain same.
Changed it to osd_dump_dict in the new commit.

tests/manage/z_cluster/test_ceph_default_values_check.py
PrasadDesala: Done

tests/manage/z_cluster/test_ceph_default_values_check.py
jilju: Okay, understood the point.
It will never reach this part if this condition is actually not satisfied. Because values in osd_full_ratios satisfies this and  its values are compared with each value in osd_dump_dict .
So this check is not really important.

PrasadDesala: Correct. This check will be helpful to catch issues where the osd_fill values are modified but didn't satisfy the condition (nearfull < backfillfull < full)
Eg: https://github.com/openshift/ocs-operator/pull/419

ebenahar:
Hi @PrasadDesala the validation done in this test is very important to do, I think we need to have it as in the verifications we do as part of the deployment

mbukatov:
I agree that this is an important test case, but I don't believe it belongs to deployment phase. If we really want to make sure this is executed during deployment every time, we could consider adding some appropriate marker to this test case.

vasukulkarni:
@ebenahar it is asserting if the default values of ceph config are not accurate, I think failure in tier1 is appropriate here rather than deployment. It shouldn't really affect testing.

ebenahar:
@mbukatov @vasukulkarni thanks for the input. I'm convinced to have it in tier1

tests/manage/z_cluster/test_ceph_default_values_check.py
```diff
@@ -0,0 +1,62 @@
...
+            f"{osd_dump_dict['nearfull_ratio']:.2f} < "
+            f"{osd_dump_dict['backfillfull_ratio']:.2f} < "
+            f"{osd_dump_dict['full_ratio']:.2f}"
+        )
+        log.info(
+            "osd full ratio values satisfies condition "
+            f"{osd_dump_dict['nearfull_ratio']:.2f} < "
+            f"{osd_dump_dict['backfillfull_ratio']:.2f} < "
+            f"{osd_dump_dict['full_ratio']:.2f}"
+        )
```
mbukatov, line 59: I have few suggestions to improve this.

1) Rename `osd_full_ratios` to `expected_full_ratios`
2) Based on values from `osd_dump_dict` create a dict with the same structure as the previous one and call it `actuall_full_ratios`.
3) Log both actual and expected values (in a for loop go through eg. osd_full_ratios and log the results, report error when they don't match but don't use assert)
4) In the very end, just use sheer `assert osd_full_ratios == actuall_full_ratios`

This way, based on a test log will always know, even if the test fails:

- what are the expected values
- what were actual values

PrasadDesala: Done

tests/manage/z_cluster/test_ceph_default_values_check.py
```diff
@@ -0,0 +1,72 @@
...
+            'backfillfull_ratio': 0.8,
+            'nearfull_ratio': 0.75
+        }
+        actual_full_ratios = {}
+        ct_pod = pod.get_ceph_tools_pod()
+        log.info("Checking the values of ceph osd full ratios in osd map")
+        osd_dump_dict = ct_pod.exec_ceph_cmd('ceph osd dump')
+        for ratio_parm, value in expected_full_ratios.items():
+            ratio_value = osd_dump_dict.get(ratio_parm)
+            actual_full_ratios.update({ratio_parm: ratio_value})
```
mbukatov: You could just write this as `actual_full_ratios[ratio_parm] = ratio_value` to improve readability.

tests/manage/z_cluster/test_ceph_default_values_check.py
```diff
@@ -0,0 +1,74 @@
...
+        }
+        actual_full_ratios = {}
+        ct_pod = pod.get_ceph_tools_pod()
+        log.info("Checking the values of ceph osd full ratios in osd map")
+        osd_dump_dict = ct_pod.exec_ceph_cmd('ceph osd dump')
+        for ratio_parm, value in expected_full_ratios.items():
+            ratio_value = osd_dump_dict.get(ratio_parm)
+            actual_full_ratios.update(
+                {ratio_parm: float(round(ratio_value, 2))}
+            )
```
mbukatov, line 35: You could just write this as `actual_full_ratios[ratio_parm] = float(round(ratio_value, 2))` to improve readability.

PrasadDesala: Done

tests/manage/z_cluster/test_ceph_default_values_check.py
PrasadDesala: Done

tests/manage/z_cluster/test_ceph_default_values_check.py
```diff
@@ -0,0 +1,74 @@
...
+            if not float(round(ratio_value, 2)) == value:
+                log.error(
+                    f"Actual {ratio_parm} value is {ratio_value:.2f} NOT "
+                    f"matching the expected value {value}"
+                )
+        assert expected_full_ratios == actual_full_ratios, (
+            f"Actual {actual_full_ratios} values does not match "
+            f"expected full ratio values {expected_full_ratios}"
+        )
```
mbukatov, line 42: Dumping dictionaries into the message of this assert is not necessary. Pytest provides quite clear report already, and having dictionaries again in the message doesn't add any value and could be actually puzzling at first:

```
>       assert expected_full_ratios == actual_full_ratios, (
            f"Actual {actual_full_ratios} values does not match "
            f"expected full ratio values {expected_full_ratios}"
        )
E       AssertionError: Actual {'full_ratio': 0.8500000238418579, 'backfillfull_ratio': 0.800000011920929, 'nearfull_ratio': 0.75} values does not match expected full ratio values {'full_ratio': 0.85, 'backfillfull_ratio': 0.8, 'nearfull_ratio': 0.75}
E       assert {'backfillful..._ratio': 0.75} == {'backfillful..._ratio': 0.75}
E         Omitting 1 identical items, use -vv to show
E         Differing items:
E         {'backfillfull_ratio': 0.8} != {'backfillfull_ratio': 0.800000011920929}
E         {'full_ratio': 0.85} != {'full_ratio': 0.8500000238418579}
E         Full diff:
E           {
E         -  'backfillfull_ratio': 0.8,
E         +  'backfillfull_ratio': 0.800000011920929,
E         ?                           ++++++++++++++
E         -  'full_ratio': 0.85,
E         +  'full_ratio': 0.8500000238418579,
E            'nearfull_ratio': 0.75,
E           }
```

See that esp. the full diff report is quite nice.

PrasadDesala: Done.

tests/manage/z_cluster/test_ceph_default_values_check.py
```diff
@@ -0,0 +1,72 @@
...
+log = logging.getLogger(__name__)
+
+
+@tier1
+@pytest.mark.polarion_id("OCS-2231")
+class TestCephDefaultValuesCheck(ManageTest):
+
+    def test_ceph_default_values_check(self):
+        """
+        This test checks ceph default values taken from OCS 4.5 with the
```
ebenahar, line 17: should OCS 4.3 be the minimum version in which these values apply?

shyRozen: I think you are right. When full disk handling was introduced (4.3), the values were change to the values in this PR. Till then the values were not exposed to OCS and were ceph default (full ratio 95% nearfull 85% as far as I remember)

PrasadDesala: yes, these values are exposed in OCS from 4.3.
I will replace the version in this comment to 4.3

PrasadDesala: Done

PrasadDesala:
No code changes are made in the new commit. So the previous PR validation job is still valid. 
Adding label Verified.
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/2390/testReport/

PR #2431, opened 2020-07-02
Adding ocs46 config files

vavuthu:


ocs_ci/framework/main.py
```diff
@@ -64,7 +64,7 @@ def init_ocsci_conf(arguments=None):
     parser = argparse.ArgumentParser(add_help=False)
     parser.add_argument('--ocsci-conf', action='append', default=[])
     parser.add_argument(
-        '--ocs-version', action='store', choices=['4.2', '4.3', '4.4', '4.5']
+        '--ocs-version', action='store', choices=['4.2', '4.3', '4.4', '4.5', '4.6']
```
petr-balogh, line 67: Just remembered that we have some condition for LSO deployment. It can be handled in different PR.
@clacroix12 I am talking about this:
https://github.com/red-hat-storage/ocs-ci/blob/c2f8c19380f89356c3cde1b22120e8619887a8e2/ocs_ci/utility/localstorage.py#L174

I think that we can maybe add this channel to config file to be able to set this via config? Or we will need to update the condition to follow unreleased content or we will need to find the way how to add catalog for unreleased LSO. If I remember correctly we did something like this in the past in deploy with OLM but we need to check with engineering if this is the way we should do.

clacroix12: @petr-balogh so the LSO channel is more based on the OCP version rather than the OCS version but your points still stand. I think we could easily have the channel set in the config file but that still relies on us manually updating the value as newer channels become available. If there is a better way for us to programmatically determine which channels are available that may be the better, more "future proof" way to go about it.

vavuthu:
> PR validation
> 
> Cluster Name:
> Cluster Configuration:
> PR Test Suite: deployment
> PR Test Path: tests/
> Additional Test Params:
> 
> [Job FAILED](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/2228//testReport) (installation failed, tests not executed).

Job failed due operator in pending state which is a know bug https://bugzilla.redhat.com/show_bug.cgi?id=1851328

Hence marking as Verified

PR #2433, opened 2020-07-02
[WIP] use kubeadmin as default for login function

fbalak:

### DIFF
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -401,19 +401,28 @@ class OCP(object):
             f"{project_name} was not deleted"
         )
 
-    def login(self, user, password):
+    def login(self, user, password=None):
         """
         Logs user in
 
         Args:
             user (str): Name of user to be logged in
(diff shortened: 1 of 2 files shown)


ebenahar:
Fixes https://github.com/red-hat-storage/ocs-ci/issues/2391?

fbalak:
@ebenahar Yes, this is done because of IBM cloud deployment have different cluster dir structure and doesn't contain `kubeadmin-password` file but it seems to be more complicated than I originally thought: It seems that `kubeadmin` user has to be logged in via password, `system:admin`has no token generated for API calls and service account can not be logged in via kubeconfig. I will probably have to generate a proper user for API communication.

ebenahar:
Hi @fbalak what's the current status with this change?

fbalak:
Hi @ebenahar. I am closing this PR as this is not a way (as described in my previous comment). I opened https://github.com/red-hat-storage/ocs-ci/pull/2766 to add user management into our framework. With this I will add a new Prometheus user in another PR.

PR #2442, opened 2020-07-05
Don't use the Ingress cert when running on IBM Cloud

### LINKED ISSUE #2454: Don't use Ingress certificate for external MCG communications when running on IBM Cloud
The default root certificates are sufficient. The Ingress cert leads to a verification failure.

MeridianExplorer:
Using the Ingress certificate causes an SSL verification failure.
The root certificates are enough, and thus, we rely on them instead.

Also improved the `_retrieve_data_reduction` flow.

Closes #2454.


ebenahar:
@MeridianExplorer Thanks for fixing! shouldn't we maintain an issue for the problem you solved here?

MeridianExplorer:
@ebenahar Done

MeridianExplorer:
[Verified](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/10832/)

ocs_ci/ocs/resources/mcg.py
```diff
@@ -349,7 +329,8 @@ def _retrieve_reduction_data():
                     )
         except TimeoutExpiredError:
             logger.error(
-                'Not enough data reduction. Something is wrong.'
+                'Data reduction is insufficient. '
+                f'{total_size - total_reduced} bytes reduced.'
```
bernerhat, line 333: i would have added the amount expected as well

PR #2445, opened 2020-07-06
Jenkins export data spreadsheet

OdedViner:


ocs_ci/ocs/jenkins.py
```diff
@@ -134,7 +136,9 @@ def wait_for_build_status(self, status, timeout=900):
                         wait_for_resource_state(
                             resource=jenkins_build, state=status, timeout=timeout
                         )
-                        self.build_completed.append((jenkins_build.name, project))
+                        self.get_build_duration_time(
```
PrasadDesala, line 139: We can wait for the build status to reach running/completed state using the same function.
how will the get_build_duration_time works if we pass wait for status running,  here we are trying to get the build duration time when the build creation is still running

OdedViner: no, build reach to 'complete' state and then "get build duration time"

PR #2452, opened 2020-07-07
DNM: Use earlier OCP 4.5 nightly build

ebenahar:

### DIFF
--- a/conf/ocp_version/ocp-4.5-config.yaml
+++ b/conf/ocp_version/ocp-4.5-config.yaml
@@ -3,7 +3,7 @@
 RUN:
   client_version: '4.5.0-0.nightly'
 DEPLOYMENT:
-  installer_version: "4.5.0-0.nightly-2020-06-19-043844"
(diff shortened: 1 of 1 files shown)


conf/ocp_version/ocp-4.5-config.yaml
```diff
@@ -3,7 +3,7 @@
 RUN:
   client_version: '4.5.0-0.nightly'
 DEPLOYMENT:
-  installer_version: "4.5.0-0.nightly-2020-06-19-043844"
+  installer_version: "4.5.0-rc.2-x86_64"
```
vasukulkarni, line 6: FYI: you can also use --ocp-version to override this.

ebenahar: I think that `--ocp-version` will work with `4.5` or `4.6` and not with the exact build

vasukulkarni: you are right, it just picks the config file based on version match.

PR #2453, opened 2020-07-07
Tier4a ocs upgrade

AviadP:
- moved test_ocs_upgrade code to helper file
- added dictionary with tier4 tests and arguments
- create class for tier4 tests with worker_node_shutdown function
- added 2 test function: worker node shutdown - abrupt and permanent
- some minor fixes

### DIFF
--- a/ocs_ci/ocs/disruptive_operations.py
+++ b/ocs_ci/ocs/disruptive_operations.py
@@ -2,7 +2,7 @@ import logging
 import time
 
 from ocs_ci.ocs.node import get_typed_nodes, get_node_name
-from ocs_ci.ocs.resources.pod import (get_ocs_operator_pod, get_pod_node)
+from ocs_ci.ocs.resources.pod import get_ocs_operator_pod, get_pod_node
 from ocs_ci.ocs.platform_nodes import PlatformNodesFactory
 from tests.helpers import wait_for_ct_pod_recovery
 
--- a/ocs_ci/ocs/disruptive_operations.py
+++ b/ocs_ci/ocs/disruptive_operations.py
@@ -29,7 +29,10 @@ def worker_node_shutdown(abrupt):
     Shutdown worker node that running ocs-operator pod
 
     Args:
-        abrupt: bool: True if abrupt shutdown, False for permanent shutdown
+        abrupt: (bool): True if abrupt shutdown, False for permanent shutdown
+
+    Raises:
+        AssertionError: in case the ceph-tools pod was not recovered
 
     """
 
--- a/ocs_ci/ocs/disruptive_operations.py
+++ b/ocs_ci/ocs/disruptive_operations.py
@@ -1,0 +1,59 @@
+import logging
+import time
+
+from ocs_ci.ocs.node import get_typed_nodes, get_node_name
+from ocs_ci.ocs.resources.pod import (get_ocs_operator_pod, get_pod_node)
+from ocs_ci.ocs.platform_nodes import PlatformNodesFactory
+from tests.helpers import wait_for_ct_pod_recovery
+
+log = logging.getLogger(__name__)
+
+
+def get_ocs_operator_node_name():
+    """
+    Getting node's name that running ocs-operator pod
+
+    Returns:
+        str: node's name that running ocs-operator pod
+
+    """
+    ocs_operator_pod = get_ocs_operator_pod()
+    log.debug(f"ocs operator pod info: {ocs_operator_pod}")
+    ocs_operator_node = get_pod_node(ocs_operator_pod)
+
+    return get_node_name(ocs_operator_node)
+
+
+def worker_node_shutdown(abrupt):
+    """
+    Shutdown worker node that running ocs-operator pod
+
+    Args:
+        abrupt: bool: True if abrupt shutdown, False for permanent shutdown
+
+    """
+
+    nodes = PlatformNodesFactory().get_nodes_platform()
+    log.info(f"Abrupt {abrupt}")
+    # get ocs-operator node:
+    ocs_operator_node_name = get_ocs_operator_node_name()
+
+    # get workers node objects:
+    node_to_shutdown = list()
+    for node in get_typed_nodes():
+        node_name = get_node_name(node)
+        log.info(f"node: {node_name}, ocs operator node: {ocs_operator_node_name}")
+        if node_name == ocs_operator_node_name:
+            node_to_shutdown.append(node)
+            log.info(f"node to shutdown: {get_node_name(node_to_shutdown[0])}")
+            nodes.stop_nodes(node_to_shutdown)
+            log.info("stop instance - done!")
+            break
+
+    log.info("Sleeping 5 minutes")
+    time.sleep(320)
+    assert wait_for_ct_pod_recovery(), "Ceph tools pod failed to come up on another node"
+    if abrupt:
+        log.info("Abrupt Shutdown")
+        if node_to_shutdown:
+            nodes.start_nodes(nodes=node_to_shutdown)
--- a/ocs_ci/ocs/ocs_upgrade.py
+++ b/ocs_ci/ocs/ocs_upgrade.py
@@ -3,30 +3,33 @@ import logging
 from copy import deepcopy
 from pkg_resources import parse_version
 from tempfile import NamedTemporaryFile
+import time
 
-from ocs_ci.deployment.deployment import create_catalog_source
 from ocs_ci.framework import config
+from ocs_ci.deployment.deployment import create_catalog_source
 from ocs_ci.ocs import constants
+from ocs_ci.ocs.cluster import CephCluster, CephHealthMonitor
 from ocs_ci.ocs.defaults import OCS_OPERATOR_NAME
-from ocs_ci.ocs.node import get_typed_nodes
 from ocs_ci.ocs.ocp import get_images, OCP
+from ocs_ci.ocs.node import get_typed_nodes
 from ocs_ci.ocs.resources.catalog_source import CatalogSource
 from ocs_ci.ocs.resources.csv import CSV
 from ocs_ci.ocs.resources.install_plan import wait_for_install_plan_and_approve
 from ocs_ci.ocs.resources.pod import verify_pods_upgraded
 from ocs_ci.ocs.resources.packagemanifest import (
-    get_selector_for_ocs_operator,
-    PackageManifest,
+    get_selector_for_ocs_operator, PackageManifest,
 )
-from ocs_ci.ocs.resources.storage_cluster import get_osd_count
+from ocs_ci.ocs.resources.storage_cluster import (
+    get_osd_count, ocs_install_verification,
+)
 from ocs_ci.ocs.utils import setup_ceph_toolbox
 from ocs_ci.utility.utils import (
-    get_latest_ds_olm_tag,
-    get_next_version_available_for_upgrade,
-    get_ocs_version_from_image,
-    load_config_file,
+    get_latest_ds_olm_tag, get_next_version_available_for_upgrade,
+    get_ocs_version_from_image, load_config_file, TimeoutSampler,
 )
 from ocs_ci.utility.templating import dump_data_to_temp_yaml
+from ocs_ci.ocs.exceptions import TimeoutException
+
 
 log = logging.getLogger(__name__)
 
@@ -357,9 +360,6 @@ class OCSUpgrade(object):
         """
         Set images for upgrade
 
-        Args:
-            upgrade_version: (str): version to be upgraded
-
         """
         ocs_catalog = CatalogSource(
             resource_name=constants.OPERATOR_CATALOG_SOURCE_NAME,
@@ -392,3 +392,65 @@ class OCSUpgrade(object):
             with NamedTemporaryFile() as cs_yaml:
                 dump_data_to_temp_yaml(cs_data, cs_yaml.name)
                 ocs_catalog.apply(cs_yaml.name)
+
+
+def run_ocs_upgrade(operation=None, *operation_args, **operation_kwargs):
+    """
+    Run upgrade procedure of OCS cluster
+
+    Args:
+        operation: (function): Function to run
+        operation_args: (iterable): Function's arguments
+        operation_kwargs: (map): Function's keyword arguments
+
+    """
+
+    ceph_cluster = CephCluster()
+    upgrade_ocs = OCSUpgrade(
+        namespace=config.ENV_DATA['cluster_namespace'],
+        version_before_upgrade=config.ENV_DATA.get("ocs_version"),
+        ocs_registry_image=config.UPGRADE.get('upgrade_ocs_registry_image'),
+        upgrade_in_current_source=config.UPGRADE.get('upgrade_in_current_source', False)
+    )
+    upgrade_version = upgrade_ocs.get_upgrade_version()
+    assert upgrade_ocs.get_parsed_versions()[1] >= upgrade_ocs.get_parsed_versions()[0], (
+        f"Version you would like to upgrade to: {upgrade_version} "
+        f"is not higher or equal to the version you currently running: "
+        f"{upgrade_ocs.version_before_upgrade}"
+    )
+    csv_name_pre_upgrade = upgrade_ocs.get_csv_name_pre_upgrade()
+    pre_upgrade_images = upgrade_ocs.get_pre_upgrade_image(csv_name_pre_upgrade)
+    upgrade_ocs.load_version_config_file(upgrade_version)
+    with CephHealthMonitor(ceph_cluster):
+        channel = upgrade_ocs.set_upgrade_channel()
+        upgrade_ocs.set_upgrade_images()
+        upgrade_ocs.update_subscription(channel)
+        if operation:
+            log.info(f"Calling test function: {operation}")
+            _ = _ = operation(*operation_args, **operation_kwargs)
+            # Workaround for issue #2531
+            time.sleep(30)
+            # End of workaround
+
+        for sample in TimeoutSampler(
+            timeout=725,
+            sleep=5,
+            func=upgrade_ocs.check_if_upgrade_completed,
+            channel=channel,
+            csv_name_pre_upgrade=csv_name_pre_upgrade,
+        ):
+            try:
+                if sample:
+                    log.info("Upgrade success!")
+                    break
+            except TimeoutException:
+                raise TimeoutException("No new CSV found after upgrade!")
+        old_image = upgrade_ocs.get_images_post_upgrade(
+            channel, pre_upgrade_images, upgrade_version
+        )
+    verify_image_versions(old_image, upgrade_ocs.get_parsed_versions()[1])
+    ocs_install_verification(
+        timeout=600, skip_osd_distribution_check=True,
+        ocs_registry_image=upgrade_ocs.ocs_registry_image,
+        post_upgrade_verification=True,
+    )
--- a/ocs_ci/ocs/ocs_upgrade.py
+++ b/ocs_ci/ocs/ocs_upgrade.py
@@ -427,7 +427,7 @@ def run_ocs_upgrade(operation=None, *operation_args, **operation_kwargs):
         upgrade_ocs.update_subscription(channel)
         if operation:
             log.info(f"Calling test function: {operation}")
-            _ = _ = operation(*operation_args, **operation_kwargs)
+            _ = operation(*operation_args, **operation_kwargs)
             # Workaround for issue #2531
             time.sleep(30)
             # End of workaround
--- a/tests/ecosystem/upgrade/test_upgrade.py
+++ b/tests/ecosystem/upgrade/test_upgrade.py
@@ -10,6 +10,19 @@ from ocs_ci.ocs.ocs_upgrade import run_ocs_upgrade
 log = logging.getLogger(__name__)
 
 
+@pytest.fixture(autouse=True)
+def teardown(request, nodes):
+
+    def finalizer():
+        """
+        Make sure all nodes are up again
+
+        """
+        nodes.restart_nodes_by_stop_and_start_teardown()
+
+    request.addfinalizer(finalizer)
+
+
 @pytest.mark.polarion_id("OCS-1579")
 def test_worker_node_abrupt_shutdown():
     """
--- a/tests/ecosystem/upgrade/test_upgrade.py
+++ b/tests/ecosystem/upgrade/test_upgrade.py
@@ -1,19 +1,41 @@
 import logging
 
+import pytest
+
 from ocs_ci.framework.testlib import ocs_upgrade
-from ocs_ci.framework import config
-from ocs_ci.ocs.cluster import CephCluster, CephHealthMonitor
-from ocs_ci.ocs.resources.storage_cluster import (
-    ocs_install_verification
-)
-from ocs_ci.utility.utils import TimeoutSampler
-from ocs_ci.ocs.ocs_upgrade import (
-    verify_image_versions,
-    OCSUpgrade,
-)
-from ocs_ci.ocs.exceptions import TimeoutException
+from ocs_ci.ocs.disruptive_operations import worker_node_shutdown
+from ocs_ci.ocs.ocs_upgrade import run_ocs_upgrade
+
 
 log = logging.getLogger(__name__)
+
+
+@pytest.mark.parametrize(
+    argnames=["operation"], argvalues=[
+        pytest.param(
+            worker_node_shutdown, marks=pytest.mark.polarion_id("OCS-1579")
+        ),
+    ]
+)
+def test_worker_node_abrupt_shutdown(operation):
+    log.info("Starting disruptive test function: "
+             "test_worker_node_abrupt_shutdown")
+    abrupt = True
+    run_ocs_upgrade(operation, abrupt)
+
+
+@pytest.mark.parametrize(
+    argnames=["operation"], argvalues=[
+        pytest.param(
+            worker_node_shutdown, marks=pytest.mark.polarion_id("OCS-1575")
+        ),
+    ]
+)
+def test_worker_node_permanent_shutdown(operation):
+    log.info("Starting disruptive test function:"
+             " test_worker_node_permanent_shutdown")
+    abrupt = False
+    run_ocs_upgrade(operation, abrupt)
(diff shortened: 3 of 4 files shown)


tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -1,66 +1,31 @@
...
-    OCSUpgrade,
-)
-from ocs_ci.ocs.exceptions import TimeoutException
+from ocs_ci.ocs.ocs_upgrade import run_ocs_upgrade, tier4_functions_to_run
+
 
 log = logging.getLogger(__name__)
 
 
+@pytest.mark.polarion_id("OCS-1579")
```
ebenahar, line 12: consider using pytest.parametrize so you do something like:


```python

@pytest.mark.parametrize(
    argnames=["operation"], argvalues=[
        pytest.param(
            "worker_abrupt_shutdown", marks=pytest.mark.polarion_id("OCS-1579")
        ),
        pytest.param(
            "worker_perm_shutdown", marks=pytest.mark.polarion_id("OCS-1575")
        ),
    ]    
)
def test_worker_node_abrupt_shutdown(nodes):
    run_ocs_upgrade(nodes, operation)

```

Then, you will not need https://github.com/red-hat-storage/ocs-ci/pull/2453/files#diff-1ba036d13bfcf92845199eee581b5e65R451-R462

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +391,132 @@ def set_upgrade_images(self):
             with NamedTemporaryFile() as cs_yaml:
                 dump_data_to_temp_yaml(cs_data, cs_yaml.name)
                 ocs_catalog.apply(cs_yaml.name)
+
+
+class Tier4aUpgrade(EcosystemTest):
```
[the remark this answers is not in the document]
AviadP: fixed

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +391,132 @@ def set_upgrade_images(self):
...
+    def get_ocs_operator_node_name(self):
+        """
+        Getting node's name that running ocs-operator pod
+
+        Returns:
+            str: node's name that running ocs-operator pod
+
+        """
+        ocs_operator_pod = get_ocs_operator_pod()
+        log.debug(f"ocs operator pod info: {ocs_operator_pod}")
```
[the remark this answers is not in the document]
AviadP: thats big yaml file, in my opinion it is needed for debug only

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +391,132 @@ def set_upgrade_images(self):
...
+        # get rook-ceph-tools node:
+        rook_ceph_tools_pod = get_ceph_tools_pod()
+        rook_ct_node = get_pod_node(rook_ceph_tools_pod)
+        rook_ct_node_name = get_node_name(rook_ct_node)
+        if rook_ct_node_name == ocs_operator_node_name:
+            delete_pods([get_ocs_operator_pod()])
+            time.sleep(120)
+            ocs_operator_node_name = self.get_ocs_operator_node_name()
+            assert not (ocs_operator_node_name == rook_ct_node_name),\
+                "rook-ceph-tools and ocs-operator pods running on same node"
```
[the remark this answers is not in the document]
AviadP: fixed

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +391,132 @@ def set_upgrade_images(self):
...
+        log.info(f"Abrupt {abrupt}")
+        # get ocs-operator node:
+        ocs_operator_node_name = self.get_ocs_operator_node_name()
+        # get rook-ceph-tools node:
+        rook_ceph_tools_pod = get_ceph_tools_pod()
+        rook_ct_node = get_pod_node(rook_ceph_tools_pod)
+        rook_ct_node_name = get_node_name(rook_ct_node)
+        if rook_ct_node_name == ocs_operator_node_name:
+            delete_pods([get_ocs_operator_pod()])
+            time.sleep(120)
```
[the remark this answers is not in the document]
AviadP: i will, but need some sleep time for pod to be created on another node

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +391,132 @@ def set_upgrade_images(self):
...
+            if node_name == ocs_operator_node_name:
+                if node_name != rook_ct_node_name:
+                    node_to_shutdown.append(node)
+                    log.info(f"node to shutdown: {get_node_name(node_to_shutdown[0])}")
+                    nodes.stop_nodes(node_to_shutdown)
+                    log.info(f"stop instance - done!")
+                    break
+
+        log.info("Sleeping 5 minutes")
+        time.sleep(300)
```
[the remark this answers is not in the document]
AviadP: it should be depends on how we want the test to be designed. sleep here give oc time to move components from node that shut down to another node. @nehaberry , what do u think ?

tests/ecosystem/upgrade/test_upgrade.py
AviadP: fixed

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -3,22 +3,27 @@
...
-def test_worker_node_permanent_shutdown(nodes):
-    run_ocs_upgrade(nodes, dictionary_name=tier4_functions_to_run,
-                    function_runner="worker_perm_shutdown")
+dist = UpgradeDisruptiveOps()
+
+
+@pytest.mark.parametrize(
+    argnames=["operation", "operation_args"], argvalues=[
+        pytest.param(
+            [dist.worker_node_shutdown], [True], marks=pytest.mark.polarion_id("OCS-1575")
```
ebenahar, line 17: I think it should be `[dist.worker_node_shutdown, True]`

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -3,22 +3,27 @@
...
+        ),
+        pytest.param(
+            [dist.worker_node_shutdown], [False], marks=pytest.mark.polarion_id("OCS-1579")
```
ebenahar, line 20: same here

tests/ecosystem/upgrade/test_upgrade.py
AviadP: it could be as it is now, or this way: 
```@pytest.mark.parametrize(
    argnames=["operation", "operation_args"], argvalues=[
        pytest.param(
            dist.worker_node_shutdown, True, marks=pytest.mark.polarion_id("OCS-1575")
        ),
        pytest.param(
            dist.worker_node_shutdown, False, marks=pytest.mark.polarion_id("OCS-1579")
        ),
    ]
)
```
i will fix it

ocs_ci/ocs/ocs_upgrade.py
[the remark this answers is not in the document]
ebenahar: so let's poll these components movement to the other node

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +391,124 @@ def set_upgrade_images(self):
...
+                "rook-ceph-tools and ocs-operator pods running on same node"
+            )
+
+        # get workers node objects:
+        node_to_shutdown = list()
+        for node in get_typed_nodes():
+            node_name = get_node_name(node)
+            log.info(f"node: {node_name}, ocs operator node: {ocs_operator_node_name}")
+            if node_name == ocs_operator_node_name:
+                if node_name != rook_ct_node_name:
```
ebenahar, line 447: consider using tests.helpers.wait_for_ct_pod_recovery so you will not need to exclude the node running the CT pod.
See reference in tests/manage/z_cluster/nodes/test_disk_failures.py:138

ocs_ci/ocs/ocs_upgrade.py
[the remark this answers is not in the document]
ebenahar: please poll the creation of a new pod instead of wait. You can do it with ocp.wait_for_resource with label selector

ocs_ci/ocs/ocs_upgrade.py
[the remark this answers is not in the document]
nehaberry: @AviadP +1 IMO too.. 

In case of node shutdown : The default k8s timeout before pods move to another node is 5 mins... So till the node is OFF for 5 mins, all the pods will stay in the same state. Hence, we might need to wait for minimum 5 mins before we observe any change in the cluster.

if we poll within this 5 mins.. all the pods would already be showing as Running.. So we need to handle it properly .

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +390,112 @@ def set_upgrade_images(self):
...
+        log.info("Sleeping 5 minutes")
+        time.sleep(320)
+        assert wait_for_ct_pod_recovery(), "Ceph tools pod failed to come up on another node"
+        if abrupt:
+            log.info("Abrupt Shutdown")
+            if node_to_shutdown:
+                nodes.start_nodes(nodes=node_to_shutdown)
+
+
+@pytest.mark.parametrize()
```
petr-balogh, line 449: I am aware of parametrize of tests or fixtures, but here it looks like some hack where you are trying to parametrize regular function here. Can you please explain the logic behind that as this doesn't look like the proper use case of parametrize here. I see you already parametrize test_worker_node_shutdown TC, so why it's needed here like this?

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +390,112 @@ def set_upgrade_images(self):
             with NamedTemporaryFile() as cs_yaml:
                 dump_data_to_temp_yaml(cs_data, cs_yaml.name)
                 ocs_catalog.apply(cs_yaml.name)
+
+
+class UpgradeDisruptiveOps(EcosystemTest):
```
petr-balogh, line 395: There is currently no usage of class as no actual reference to self is really needed here.  Maybe in the future you are going to use it but for now I see that two bellow can just live in upgrade module and as you really do not store any object member which needs to be shared it will simplify the logic here.  Or you can create those class as class methods and people can use them without creating an instance.

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +390,112 @@ def set_upgrade_images(self):
...
+            str: node's name that running ocs-operator pod
+
+        """
+        ocs_operator_pod = get_ocs_operator_pod()
+        log.debug(f"ocs operator pod info: {ocs_operator_pod}")
+        ocs_operator_node = get_pod_node(ocs_operator_pod)
+
+        return get_node_name(ocs_operator_node)
+
+    def worker_node_shutdown(self, nodes, abrupt):
```
petr-balogh, line 415: Is this function only upgrade specific? Or can this live directly in the node related module  `ocs_ci.ocs.node` ? Maybe also the function above is not really upgrade specific and can live directly in node module and used from there?

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +390,112 @@ def set_upgrade_images(self):
...
+            log.info("Abrupt Shutdown")
+            if node_to_shutdown:
+                nodes.start_nodes(nodes=node_to_shutdown)
+
+
+@pytest.mark.parametrize()
+def run_ocs_upgrade(nodes, operation=None, operation_args=None):
+    """
+    Tests upgrade procedure of OCS cluster
+
```
petr-balogh, line 453: missing docstring for all parameters

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +390,112 @@ def set_upgrade_images(self):
...
+        f"{upgrade_ocs.version_before_upgrade}"
+    )
+    csv_name_pre_upgrade = upgrade_ocs.get_csv_name_pre_upgrade()
+    pre_upgrade_images = upgrade_ocs.get_pre_upgrade_image(csv_name_pre_upgrade)
+    upgrade_ocs.load_version_config_file(upgrade_version)
+    with CephHealthMonitor(ceph_cluster):
+        channel = upgrade_ocs.set_upgrade_channel()
+        upgrade_ocs.set_upgrade_images()
+        upgrade_ocs.update_subscription(channel)
+        if operation and operation_args:
```
petr-balogh, line 475: What if the operation does not take the arguments?

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +390,112 @@ def set_upgrade_images(self):
...
+            log.info("**** tier4a test starting here ****")
+            log.info(f"running test function: {operation}")
+            _ = operation(nodes, operation_args)
```
petr-balogh, line 478: The call of operation should be done more generic to accept args and kwargs.  And in this call you should do something like 
```python
operation(*operation_args, **operation_kwargs)
```
Where `operation_args` will be list of all arguments and `operation_kwargs` will be dictionary of named arguments.

BTW why we need to pass nodes here? It will basically require each operation needs to take this nodes as first parameter where we close the door to have all this more general and reusable for other operations.

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +390,112 @@ def set_upgrade_images(self):
...
+    )
+    csv_name_pre_upgrade = upgrade_ocs.get_csv_name_pre_upgrade()
+    pre_upgrade_images = upgrade_ocs.get_pre_upgrade_image(csv_name_pre_upgrade)
+    upgrade_ocs.load_version_config_file(upgrade_version)
+    with CephHealthMonitor(ceph_cluster):
+        channel = upgrade_ocs.set_upgrade_channel()
+        upgrade_ocs.set_upgrade_images()
+        upgrade_ocs.update_subscription(channel)
+        if operation and operation_args:
+            log.info("**** tier4a test starting here ****")
```
petr-balogh, line 476: I wouldn't print anything about tier4a test here, as this suppose to be general place which will allow us to trigger whatever operations here.

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -1,66 +1,36 @@
...
+    ]
+)
+def test_worker_node_shutdown(nodes, operation, operation_args):
+    log.info("Starting disruptive test function: test_worker_node_shutdown")
+    run_ocs_upgrade(nodes, operation, operation_args)
+
 
 @ocs_upgrade
-def test_upgrade():
+def test_upgrade(nodes):
```
petr-balogh, line 30: Usage of nodes fixture here looks like overkill to me which adding a lot of complexity where you need just to pass it everywhere else.

Instead of passing it everywhere. Please use directly call of:
platform_nodes.PlatformNodesFactory().get_nodes_platform()
Just on the place where you really need those nodes which can be for example your operation?

ocs_ci/ocs/ocs_upgrade.py
AviadP: i need to pass 'nodes' fixture as parameter to run_ocs_upgrade function, i would love to find another way to implement it

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +390,112 @@ def set_upgrade_images(self):
...
+        ocs_operator_node = get_pod_node(ocs_operator_pod)
+
+        return get_node_name(ocs_operator_node)
+
+    def worker_node_shutdown(self, nodes, abrupt):
+        """
+        Shutdown worker node that running ocs-operator pod
+
+        Args:
+            abrupt: bool: True if abrupt shutdown, False for permanent shutdown
```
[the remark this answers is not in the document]
petr-balogh: it does not matter if it's or not a fixture, here it's just anothe parameter as it's a regular function.  If the nodes fixture returns list of the nodes, than here it's regular parameter with type list which contains node objects. But as I told you in other comment, IMO it's not right approach here.

BTW, nodes fixture can be considered as fixture only in test or another fixture, this case you don't need to document. But in this case as I described you should as it's not direct use of fixture.

ocs_ci/ocs/ocs_upgrade.py
petr-balogh: Yes, please set up mtg with me for afternoon ;) like 3pm your time

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -385,3 +390,112 @@ def set_upgrade_images(self):
...
+        # get ocs-operator node:
+        ocs_operator_node_name = self.get_ocs_operator_node_name()
+
+        # get workers node objects:
+        node_to_shutdown = list()
+        for node in get_typed_nodes():
+            node_name = get_node_name(node)
+            log.info(f"node: {node_name}, ocs operator node: {ocs_operator_node_name}")
+            if node_name == ocs_operator_node_name:
+                # if node_name != rook_ct_node_name:
```
petr-balogh, line 433: ?

ocs_ci/ocs/ocs_upgrade.py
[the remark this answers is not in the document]
AviadP: as we spoke, i will avoid using nodes fixture here

ocs_ci/ocs/ocs_upgrade.py
AviadP: left over, will be removed

ocs_ci/ocs/ocs_upgrade.py
AviadP: There is more functions like this planned, and we can use it for other things then upgrade. maybe new module for disruptive function would be better?

petr-balogh: Sounds good

ocs_ci/ocs/ocs_upgrade.py
AviadP: as discussed, i will move it to another module for disruptive operations, and no class will be needed to gather them all in same place.

ocs_ci/ocs/ocs_upgrade.py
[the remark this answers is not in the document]
AviadP: with @nehaberry comment, will leave it without polling. and increase timeout by 20 sec, to be on safer side

tests/ecosystem/upgrade/test_upgrade.py
AviadP: after discussion with @petr-balogh, i will split it again into 2 functions as it will be easier to execute and track results

ocs_ci/ocs/disruptive_operations.py
```diff
@@ -0,0 +1,59 @@
...
+    return get_node_name(ocs_operator_node)
+
+
+def worker_node_shutdown(abrupt):
+    """
+    Shutdown worker node that running ocs-operator pod
+
+    Args:
+        abrupt: bool: True if abrupt shutdown, False for permanent shutdown
+
```
petr-balogh, line 33: You can also add Raises: section with AssertionError and describe when it's raised

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -392,3 +392,65 @@ def set_upgrade_images(self):
...
+    csv_name_pre_upgrade = upgrade_ocs.get_csv_name_pre_upgrade()
+    pre_upgrade_images = upgrade_ocs.get_pre_upgrade_image(csv_name_pre_upgrade)
+    upgrade_ocs.load_version_config_file(upgrade_version)
+    with CephHealthMonitor(ceph_cluster):
+        channel = upgrade_ocs.set_upgrade_channel()
+        upgrade_ocs.set_upgrade_images()
+        upgrade_ocs.update_subscription(channel)
+        if operation:
+            log.info(f"Calling test function: {operation}")
+            _ = _ = operation(*operation_args, **operation_kwargs)
```
petr-balogh, line 430:
```suggestion
            _ = operation(*operation_args, **operation_kwargs)
```

I think that one is extra here, right?

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -1,66 +1,48 @@
...
+    ]
+)
+def test_worker_node_abrupt_shutdown(operation):
+    log.info("Starting disruptive test function: "
+             "test_worker_node_abrupt_shutdown")
+    abrupt = True
+    run_ocs_upgrade(operation, abrupt)
+
+
+@pytest.mark.parametrize(
```
petr-balogh, line 27: I think you can omit parametrize here and above.

Just use directly marker for polarion:
@pytest.mark.polarion_id("OCS-1575")

the operation can be directly passed in run_ocs_upgrade call like I will comment below.

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -1,66 +1,48 @@
...
+        pytest.param(
+            worker_node_shutdown, marks=pytest.mark.polarion_id("OCS-1575")
+        ),
+    ]
+)
+def test_worker_node_permanent_shutdown(operation):
+    log.info("Starting disruptive test function:"
+             " test_worker_node_permanent_shutdown")
+    abrupt = False
+    run_ocs_upgrade(operation, abrupt)
```
petr-balogh, line 38:
```suggestion
    run_ocs_upgrade(operation=worker_node_shutdown, abrupt=False)
```

The parametrize is bringing here the complexity which we don't need at all I think.

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -1,66 +1,48 @@
...
+@pytest.mark.parametrize(
+    argnames=["operation"], argvalues=[
+        pytest.param(
+            worker_node_shutdown, marks=pytest.mark.polarion_id("OCS-1579")
+        ),
+    ]
+)
+def test_worker_node_abrupt_shutdown(operation):
+    log.info("Starting disruptive test function: "
+             "test_worker_node_abrupt_shutdown")
```
petr-balogh, line 22:
```suggestion
    log.info(
        "Starting disruptive test function: "
        "test_worker_node_abrupt_shutdown"
    )
```

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -1,66 +1,48 @@
...
+
+@pytest.mark.parametrize(
+    argnames=["operation"], argvalues=[
+        pytest.param(
+            worker_node_shutdown, marks=pytest.mark.polarion_id("OCS-1575")
+        ),
+    ]
+)
+def test_worker_node_permanent_shutdown(operation):
+    log.info("Starting disruptive test function:"
```
petr-balogh, line 35: The same indention as I mentioned above

ocs_ci/ocs/ocs_upgrade.py
AviadP: typo, yes. fixed

tests/ecosystem/upgrade/test_upgrade.py
AviadP: removed

tests/ecosystem/upgrade/test_upgrade.py
AviadP: done

ocs_ci/ocs/disruptive_operations.py
AviadP: added

ocs_ci/ocs/disruptive_operations.py
```diff
@@ -0,0 +1,62 @@
+import logging
+import time
+
+from ocs_ci.ocs.node import get_typed_nodes, get_node_name
+from ocs_ci.ocs.resources.pod import (get_ocs_operator_pod, get_pod_node)
```
ebenahar, line 5: `( )` are redundant

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -1,66 +1,45 @@
...
+from ocs_ci.ocs.disruptive_operations import worker_node_shutdown
+from ocs_ci.ocs.ocs_upgrade import run_ocs_upgrade
+
 
 log = logging.getLogger(__name__)
 
 
+@pytest.mark.polarion_id("OCS-1579")
+def test_worker_node_abrupt_shutdown():
+    """
```
ebenahar, line 15: text is not aligned

ocs_ci/ocs/disruptive_operations.py
```diff
@@ -0,0 +1,62 @@
+import logging
```
petr-balogh, line 1: I've just found that we have:  tests/disruption_helpers.py - what do you think to consolidate those two modules?

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -1,66 +1,58 @@
...
-)
-from ocs_ci.ocs.exceptions import TimeoutException
+from ocs_ci.ocs.disruptive_operations import worker_node_shutdown
+from ocs_ci.ocs.ocs_upgrade import run_ocs_upgrade
+
 
 log = logging.getLogger(__name__)
 
 
+@pytest.fixture(autouse=True)
```
petr-balogh, line 13: this autouse fixture will be used for each function here in this test module so also for normal upgrade where we don't deal with restart nodes.  I am not sure if we want to have this finalizer called even for regular upgrade?

Also it's by default in functions cope, so it will be called after each tests if we have more tests defined in this module.  So I would not add it as autouse but use this fixture only in tests where it's needed.

ocs_ci/ocs/disruptive_operations.py
AviadP: i can move my code into this file for now, but in my opinion we should discuss having helpers files within tests/ folder

tests/ecosystem/upgrade/test_upgrade.py
AviadP: ok

ocs_ci/ocs/disruptive_operations.py
petr-balogh: I agree with not have them in tests folder.  For now you can have them as you have, but open issue and in some next PR move maybe helper functions from tests/disruption_helpers.py  to your new module?

PR #2461, opened 2020-07-08
setup ceph client when cluster is in independent mode

vasukulkarni:

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -48,6 +48,8 @@ DEPLOYMENT:
   force_download_installer: True
   force_download_client: True
   default_latest_tag: 'latest-4.5'
+  # define if ceph is setup as independent mode, default is false
+  independent_mode: False
   # You can overwrite csv channel version by following parameter
   ocs_csv_channel: "stable-4.5"
   # you can overwrite the image for ocs operator catalog source by following parameter:
@@ -147,3 +149,8 @@ AUTH:
 # flexy_default_config.yaml
 FLEXY:
   flexy_deploy: false
+
+# Defines the configuration of Independent Ceph Cluster
+INDEPENDENT_MODE:
+  admin_keyring:
+     key: PLACE_HOLDER
--- a/ocs_ci/ocs/utils.py
+++ b/ocs_ci/ocs/utils.py
@@ -29,6 +29,7 @@ from ocs_ci.utility.prometheus import PrometheusAPI
 from ocs_ci.utility.retry import retry
 from ocs_ci.utility.utils import create_directory_path, mirror_image, run_cmd
 
+
 log = logging.getLogger(__name__)
 
 
@@ -647,13 +648,23 @@ def setup_ceph_toolbox(force_setup=False):
     """
     namespace = ocsci_config.ENV_DATA['cluster_namespace']
     ceph_toolbox = get_pod_name_by_pattern('rook-ceph-tools', namespace)
+    # setup toolbox for independent mode
+    # Refer bz: 1856982 - invalid admin secret
     if len(ceph_toolbox) == 1:
         log.info("Ceph toolbox already exists, skipping")
         if force_setup:
             log.info("Running force setup for Ceph toolbox!")
         else:
             return
-    if ocsci_config.ENV_DATA.get("ocs_version") == '4.2':
+    independent_mode = ocsci_config.DEPLOYMENT.get("independent_mode")
+    if independent_mode:
+        toolbox = templating.load_yaml(constants.TOOL_POD_YAML)
+        keyring_dict = ocsci_config.INDEPENDENT_MODE.get("admin_keyring")
+        env = [{'name': 'ROOK_ADMIN_SECRET', 'value': keyring_dict['key']}]
+        toolbox['spec']['template']['spec']['containers'][0]['env'] = env
+        rook_toolbox = OCS(**toolbox)
+        rook_toolbox.create()
+    elif ocsci_config.ENV_DATA.get("ocs_version") == '4.2':
         rook_operator = get_pod_name_by_pattern(
             'rook-ceph-operator', namespace
         )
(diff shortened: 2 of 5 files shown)


ocs_ci/framework/conf/default_config.yaml
```diff
@@ -138,3 +140,8 @@ UPGRADE:
 # or ping in ocs infra chat room.
 AUTH:
     test_quay_auth: test_secret
+
+# Defines the configuration of Independent Ceph Cluster
+INDEPENDENT_MODE:
+  admin_keyring:
+     key: 'xyz=='
```
shylesh, line 147: why we need another level key ```key```, can't we directly have keyring string for the key ```admin_keyring```

ocs_ci/ocs/utils.py
```diff
@@ -677,6 +680,24 @@ def setup_ceph_toolbox(force_setup=False):
             'json --patch  \'[{ "op": "replace", "path": '
             '"/spec/enableCephTools", "value": true }]\''
         )
+    # use the existing toolbox but fix the keyring to talk to
+    # external cluster
+    independent_mode = ocsci_config.DEPLOYMENT.get("independent_mode")
+    if independent_mode:
+        keyring_dict = ocsci_config.INDEPENDENT_MODE.get("admin_keyring")
+        log.info("Setting up keyring on toolbox for independent mode")
+        ceph_toolbox = get_pod_name_by_pattern('rook-ceph-tools', namespace)
```
shylesh, line 689: so if I understand correctly the control flow, when we are deploying OCS for external cluster , at this point in time we will not be having ```rook-ceph-tools``` existing right ? OR you assume that we deploy it first without keyring and then update ?

ocs_ci/ocs/utils.py
```diff
@@ -29,6 +30,8 @@
 from ocs_ci.utility.retry import retry
 from ocs_ci.utility.utils import create_directory_path, mirror_image, run_cmd
 
+
+CEPH_KEYRING_FILE = '/etc/ceph/keyring2'
```
shylesh, line 34: Dont we need the file name convention ```/etc/ceph/<cluster>.client.admin.keyring``` ?

ocs_ci/framework/conf/default_config.yaml
vasukulkarni: we can, i just used full line as template, if we want 'value' only here, I have to update the template, either of those choice looks fine to me.

ocs_ci/ocs/utils.py
vasukulkarni: yeah correct, the call here should setup toolbox, I haven't seen the yaml in detail but it should help setup the pod with client binaries and we update the keyring.

ocs_ci/ocs/utils.py
vasukulkarni: /etc/ceph/keyring should work as well( I will fix the '2' string), it supports few formats without client/username as well.

ebenahar:
@nehaberry @hnallurv @rachael-george @Shrivaibavi please review this PR so we will have it merged in order to move forward with running tier1

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -138,3 +140,8 @@ UPGRADE:
 # or ping in ocs infra chat room.
 AUTH:
     test_quay_auth: test_secret
+
+# Defines the configuration of Independent Ceph Cluster
+INDEPENDENT_MODE:
+  admin_keyring:
```
clacroix12, line 146: Does this info warrant an entirely new top level section for our config? It seems like it would be easy enough to add under the DEPLOYMENT section as `independent_mode_keyring` or something similar.

ocs_ci/templates/ceph/ceph-keyring.j2
```diff
@@ -0,0 +1,2 @@
+[client.admin]
+{{ admin_keyring }}
```
clacroix12, line 2: Add empty line at the end of the file

ocs_ci/ocs/utils.py
```diff
@@ -677,6 +680,24 @@ def setup_ceph_toolbox(force_setup=False):
...
             '"/spec/enableCephTools", "value": true }]\''
         )
+    # use the existing toolbox but fix the keyring to talk to
+    # external cluster
+    independent_mode = ocsci_config.DEPLOYMENT.get("independent_mode")
+    if independent_mode:
+        keyring_dict = ocsci_config.INDEPENDENT_MODE.get("admin_keyring")
+        log.info("Setting up keyring on toolbox for independent mode")
+        ceph_toolbox = get_pod_name_by_pattern('rook-ceph-tools', namespace)
+        from ocs_ci.ocs.resources.pod import upload
```
clacroix12, line 690: Is the import here purposeful or can we import at the top of the file?

ocs_ci/framework/conf/default_config.yaml
vasukulkarni: We actually have couple other variables that are specific to this mode, I think it makes sense to move them to its own section (ip/user/pass/pool config etc)

vasukulkarni:
As I looked more deep into the toolbox_pod.yaml it looks like an operator issue for now, I have raised bz 1856982 to track that.

ocs_ci/templates/ceph/ceph-keyring.j2
vasukulkarni: Resolving because the new changes removed this

ocs_ci/ocs/utils.py
vasukulkarni: Resolving because the new changes removed this

ocs_ci/templates/ocs-deployment/toolbox_pod.yaml
```diff
@@ -19,7 +19,7 @@ spec:
       dnsPolicy: ClusterFirstWithHostNet
       containers:
       - name: rook-ceph-tools
-        image: rook/ceph:v1.1.1
+        image: rook/ceph:v1.3.8
```
vasukulkarni, line 22: this version should be released today.

vasukulkarni:
Verified during my unit test and also by Shylesh's PR.

PR #2462, opened 2020-07-09
Enabling logging testcases to run on OCP 4.5 and lesser

Shrivaibavi:
From OCP 4.5, The usage of elasticsearch5 is deprecated and elasticsearch6 is in use. Hence it needs some changes in the testcases to run on version on OCP 4.5 and lesser than that.
There are few changes in the way we fetch the logs of a project and few command changes.

### DIFF
--- a/tests/e2e/logging/test_openshift-logging.py
+++ b/tests/e2e/logging/test_openshift-logging.py
@@ -9,8 +9,8 @@ import pytest
 import random
 
 from tests import helpers, disruption_helpers
-from ocs_ci.ocs import constants
-from ocs_ci.ocs.resources.pod import get_all_pods, get_pod_obj, delete_deploymentconfig_pods
+from ocs_ci.ocs import constants, defaults
+from ocs_ci.ocs.resources.pod import get_all_pods, delete_deploymentconfig_pods
 from ocs_ci.utility.retry import retry
 from ocs_ci.framework.pytest_customization.marks import skipif_aws_i3
 from ocs_ci.framework.testlib import E2ETest, workloads, tier1, ignore_leftovers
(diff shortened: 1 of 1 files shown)


tests/e2e/logging/test_openshift-logging.py
```diff
@@ -72,44 +72,82 @@ def finalizer():
         helpers.wait_for_resource_state(resource=pod_obj, state=constants.STATUS_RUNNING)
         return pod_obj, pvc_obj
 
-    @retry(ModuleNotFoundError, tries=5, delay=200, backoff=1)
-    def validate_project_exists(self, pvc_obj):
+    @retry(ModuleNotFoundError, tries=5, delay=400, backoff=1)
+    def validate_project_exists(self, project):
```
ebenahar, line 76: I see you changed all the usages of this function to send project instead of pvc. +1

tests/e2e/logging/test_openshift-logging.py
```diff
@@ -72,44 +72,82 @@ def finalizer():
...
         else:
-            raise ModuleNotFoundError
+            cmd = (
+                'es_util --query=*/_search?pretty -d \'{"query": {"match":'
+                f' {{"kubernetes.namespace_name": "{project}"}}}}}}\''
+            )
+            project_out = elasticsearch_pod_obj.exec_cmd_on_pod(
+                command=cmd, out_yaml_format=False
+            )
+            if project_out:
```
ebenahar, line 108: in which case `project_out` can be empty?

tests/e2e/logging/test_openshift-logging.py
```diff
@@ -72,44 +72,82 @@ def finalizer():
...
+        Check the files in the project
+
+        Args:
+            project (str): The project name
+
+        """
+
+        elasticsearch_pod_obj = self.get_elasticsearch_pod_obj()
+        cmd = f'es_util --query=project.{project}.*/_count'
+        if get_ocp_version() >= "4.5":
```
ebenahar, line 140: is it intentional that the check here is if the cluster version is greater or equal to 4.5?

tests/e2e/logging/test_openshift-logging.py
Shrivaibavi: Changed

tests/e2e/logging/test_openshift-logging.py
Shrivaibavi: yes it is, The logging commands and the way of checking for a project and files in EFK stack significantly changed from OCP 4.5. hence it is written in such a way that the test case should run on OCP 4.5 and versions lesser than that.

Issue #2502, opened 2020-07-14
Noobaa DB dump is collected 5 times


petr-balogh:
There was one fix for the never ending loop here:
https://github.com/red-hat-storage/ocs-ci/pull/2497/files

But it still collecting the 5 times because the condition is satisfied when count is not we don't have 5 collection done.

The condition has to be updated.

PR #2503, opened 2020-07-14
Fix db dump collection

petr-balogh:
Fixes: #2502

### DIFF
--- a/ocs_ci/ocs/utils.py
+++ b/ocs_ci/ocs/utils.py
@@ -831,12 +831,11 @@ def collect_ocs_logs(dir_name, ocp=True, ocs=True, mcg=False):
         )
     if mcg:
         counter = 0
-        mcg_db_collected = False
-        while not mcg_db_collected or counter < 5:
+        while counter < 5:
             counter += 1
             try:
                 collect_noobaa_db_dump(log_dir_path)
-                mcg_db_collected = True
+                break
             except CommandFailed as ex:
                 log.error(f"Failed to dump noobaa DB! Error: {ex}")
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/utils.py
```diff
@@ -831,12 +831,11 @@ def collect_ocs_logs(dir_name, ocp=True, ocs=True, mcg=False):
         )
     if mcg:
         counter = 0
-        mcg_db_collected = False
-        while not mcg_db_collected or counter < 5:
+        while counter < 5:
```
clacroix12, line 834: Why not use our [retry decorator](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/retry.py) for `collect_noobaa_db_dump`?

petr-balogh: This was done by @ebenahar , and I am just fixing the issue here. The retry will anyway the re-raise the exception after final retry. This is not the case here as we are just ignoring it here and continue without re-raising the exception.

petr-balogh:
For sure I've triggered PR verification for tier1, but not sure if we will hit some failing test which will invoke must gather here.

petr-balogh:
Only one dump done:
23:12:17 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig rsh noobaa-db-0 mongodump --archive=nbcore.gz --gzip --db=nbcore

Issue #2245, opened 2020-06-03
Health check for upgrade needs to be improved


petr-balogh:
In the cluster with more OSDs after scale up cluster it takes more time to recover the cluster and we can see that in the job for example:
```ocs-ci results for OCS4-4-Downstream-OCP4-4-AWS-IPI-3AZ-RHCOS-3M-3W-upgrade-ocs-auto (BUILD ID: v4.4.0-437.ci RUN ID: 1591139817)```

It took more than 90 mins to recover cluster after upgrade, but just a bit, so we need to change the logic to check if the PG degradation is slowly decreasing and if so we can increase the re-tries.

PR #2505, opened 2020-07-14
Add inteligent ceph health

petr-balogh:
Fixes: #2245
Fixes: #2519

### DIFF
--- a/ocs_ci/ocs/ocs_upgrade.py
+++ b/ocs_ci/ocs/ocs_upgrade.py
@@ -105,7 +105,10 @@ def verify_image_versions(old_images, upgrade_version):
         old_images, selector=constants.CSI_RBDPLUGIN_PROVISIONER_LABEL,
         count=2
     )
-    verify_pods_upgraded(old_images, selector=constants.MON_APP_LABEL, count=3)
+    # Increased timeout for BZ https://bugzilla.redhat.com/show_bug.cgi?id=1856254
+    verify_pods_upgraded(
+        old_images, selector=constants.MON_APP_LABEL, count=3, timeout=1200
+    )
     verify_pods_upgraded(old_images, selector=constants.MGR_APP_LABEL)
     # OSD upgrade have timeout 10mins for new attempt if cluster is not health.
     # https://bugzilla.redhat.com/show_bug.cgi?id=1840729 setting timeout for
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -1304,7 +1304,7 @@ def get_az_count():
 
 def ceph_health_check(
     namespace=None, tries=20, delay=30, intelligent_wait=False,
-    max_timeout_for_pg=1500,
+    pg_recovery_progress_timeout=1500,
 ):
     """
     Args:
@@ -1315,9 +1315,9 @@ def ceph_health_check(
         intelligent_wait (bool): If true, the ceph_health is waiting for
             HEALTH_OK if we have degraded or undersized PGs. In case the time
             between decrease of degraded/undersized PGs is less than
-            max_timeout_for_pg than we continue to wait.
-        max_timeout_for_pg (int): Max timout in seconds to wait for decrease in
-            degraded/undersized PGs change for intelligent wait mode.
+            pg_recovery_progress_timeout then we continue to wait.
+        pg_recovery_progress_timeout (int): Max timout in seconds to wait for
+            decrease in degraded/undersized PGs change for intelligent wait mode.
 
     Returns:
         bool: ceph_health_check_base return value with default retries of 20,
@@ -1357,17 +1357,19 @@ def ceph_health_check(
                     degraded_pgs_diff_times.append(decrease_time - last_decrease)
                     degraded_pgs_diffs.append(degraded_pgs - degraded_pgs_new)
                     degraded_pgs = degraded_pgs_new
-                    degraded_pgs_diff_len = len(degraded_pgs_diffs)
-                    average_pgs_diff = (
-                        sum(degraded_pgs_diff_times) / degraded_pgs_diff_len
-                    )
-                    average_pgs_time_diff = (
-                        sum(degraded_pgs_diff_times) / degraded_pgs_diff_len
-                    )
-                    log.info(
-                        f"Average decrease of degraded PGs is {average_pgs_diff}, "
-                        f"average decrease time is {average_pgs_time_diff}sec."
-                    )
+                    degraded_pgs_diff_len = len(degraded_pgs_diffs) - 1
+                    if degraded_pgs_diff_len > 0:
+                        average_pgs_diff = (
+                            sum(degraded_pgs_diffs[1:]) / degraded_pgs_diff_len
+                        )
+                        average_pgs_time_diff = (
+                            sum(degraded_pgs_diff_times[1:]) / degraded_pgs_diff_len
+                        )
+                        log.info(
+                            "Average decrease of degraded PGs is "
+                            f"{average_pgs_diff}, average decrease time is "
+                            f"{average_pgs_time_diff}sec."
+                        )
             undersized_pgs_regexp = r"(\d+)(?=\s*pgs undersized)"
             undersized_pgs_found = re.findall(undersized_pgs_regexp, str_ex)
             if undersized_pgs_found:
@@ -1378,7 +1380,7 @@ def ceph_health_check(
             if decrease_time:
                 last_decrease = decrease_time
             last_time_change_diff = time.time() - last_decrease
-            if last_time_change_diff < max_timeout_for_pg:
+            if last_time_change_diff < pg_recovery_progress_timeout:
                 if degraded_pgs or undersized_pgs:
                     log.warn(
                         f"Waiting for clean PGs. Degraded: {degraded_pgs} "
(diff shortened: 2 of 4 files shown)


petr-balogh:
Testing here:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Upgrade-OCS/job/qe-trigger-aws-ipi-3az-rhcos-3m-3w-upgrade-ocs-auto/31/console

ocs_ci/utility/utils.py
```diff
@@ -1302,25 +1302,104 @@ def get_az_count():
...
+            if degraded_pgs_found:
+                degraded_pgs_new = int(degraded_pgs_found[0])
+                if degraded_pgs == 0 or degraded_pgs_new < degraded_pgs:
+                    decrease_time = time.time()
+                    degraded_pgs_diff_times.append(decrease_time - last_decrease)
+                    degraded_pgs_diffs.append(degraded_pgs - degraded_pgs_new)
+                    degraded_pgs = degraded_pgs_new
+                    degraded_pgs_diff_len = len(degraded_pgs_diffs)
+                    average_pgs_diff = (
+                        sum(degraded_pgs_diff_times) / degraded_pgs_diff_len
```
shylesh, line 1362: I guess this should be ```sum(degaraded_pgs_diff)```

ocs_ci/utility/utils.py
```diff
@@ -1302,25 +1302,104 @@ def get_az_count():
...
+                        f"Waiting for clean PGs. Degraded: {degraded_pgs} "
+                        f"Undersized: {undersized_pgs}."
+                    )
+                    log.warn(ex)
+                    log.info(
+                        f"Waiting for {delay}sec for next attempt. Timeout "
+                        f"remaining: {max_timeout_for_pg - last_time_change_diff}"
+                    )
+                    time.sleep(delay)
+                else:
```
shylesh, line 1393: I am not able to visualize flow here, if we don't have any degraded or undersized PGs ( I assume pgs are  clean) why we have to sleep in this block ? why can't just loop back and wait in the ```retry``` of try block above? because if its already clean then we don't want to sleep here for ```delay``` period of time.  am I missing something ?

ocs_ci/utility/utils.py
```diff
@@ -1302,25 +1302,104 @@ def get_az_count():
...
     Args:
         namespace (str): Namespace of OCS
             (default: config.ENV_DATA['cluster_namespace'])
         tries (int): Number of retries
         delay (int): Delay in seconds between retries
+        intelligent_wait (bool): If true, the ceph_health is waiting for
+            HEALTH_OK if we have degraded or undersized PGs. In case the time
+            between decrease of degraded/undersized PGs is less than
+            max_timeout_for_pg than we continue to wait.
+        max_timeout_for_pg (int): Max timout in seconds to wait for decrease in
```
shylesh, line 1319: Somehow the name ```max_timeout_for_pg``` is bit misleading to me, it made me think like 'its timeout per pg' . How about ```pg_recovery_progress_timeout```. May be others would be having better name , wdyt?

ocs_ci/utility/utils.py
petr-balogh: If we got into this block we had to hit CephHealthException exception, that means that we had some issue with Ceph health.

We have two possibilities here:
1) degraded_pgs or undersized_pgs - so we gather those data from exception and we had degraded or undersized PGs.  The log message above this sleep print's that we are waiting for another time to see that it reduces if we didn't reach max timeout.
2) Ceph health error was caused by something else than degraded/undersized PGs.  Then we decrease number of tries below and continue in next cycle in loop. (As intelligent wait now works only for degraded/undersized PGs) but this still simulate retry as it was in non intelligent wait.

The block else with raise an exception is when we reach the timeout.  But I've just realized that we don't raise an exception in case we reach maximum re-tries, so I will change the logic here a bit to raise an exception when we move out of while loop which will solve the issue.

Hope it's clear here @shylesh ?

petr-balogh:
Verified here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/10004/consoleFull

vasukulkarni:
@petr-balogh I am not sure we need the logic to wait for PG degradation check, I believe we have to see how it works without IO and see what the bz leads to.  the upper timeout we have using retry should be tweaked so that will serve as baseline for future regression.

petr-balogh:
> @petr-balogh I am not sure we need the logic to wait for PG degradation check, I believe we have to see how it works without IO and see what the bz leads to. the upper timeout we have using retry should be tweaked so that will serve as baseline for future regression.

@vasukulkarni we already know that without heavy load done by @fbalak tests it recover much faster as I updated the last run link in the BZ.

With this change we can also run upgrade with heavy load and @nehaberry told me that on some their cluster they run upgrade the recover took about 33 hours or something like that.   I don't want to have one specific timeout in upgrade test to wait for recover as it really depends on the load in the cluster and if we are just expecting to wait after upgrade to see HEALTH_OK I don't want to have static timeout set because I will close the door to test other options here.

So this will help us to help to not fail the post upgrade check.  In our case the upgrade of OCS itself passed. But then we had degraded PGs.  We discussed it some time back on upgrade sync and was suggested that we will check if the numbers are decreasing that we will just wait till the cluster will recover to HEALTH_OK.

What if I will do massive add capacity just before upgrade?  Maybe it will take also some bigger amount of time to get cluster healthy (don't know). But I know that upgrade of OCS passed - that's what I am testing in OCS upgrade test, and last step in verification is to see CEPH HEALTH_OK .  So this logic is only used after upgrade currently, nowhere else.

vasukulkarni:
> > @petr-balogh I am not sure we need the logic to wait for PG degradation check, I believe we have to see how it works without IO and see what the bz leads to. the upper timeout we have using retry should be tweaked so that will serve as baseline for future regression.
> 
> @vasukulkarni we already know that without heavy load done by @fbalak tests it recover much faster as I updated the last run link in the BZ.
> 
> With this change we can also run upgrade with heavy load and @nehaberry told me that on some their cluster they run upgrade the recover took about 33 hours or something like that. I don't want to have one specific timeout in upgrade test to wait for recover as it really depends on the load in the cluster and if we are just expecting to wait after upgrade to see HEALTH_OK I don't want to have static timeout set because I will close the door to test other options here.
> 
> So this will help us to help to not fail the post upgrade check. In our case the upgrade of OCS itself passed. But then we had degraded PGs. We discussed it some time back on upgrade sync and was suggested that we will check if the numbers are decreasing that we will just wait till the cluster will recover to HEALTH_OK.
> 
> What if I will do massive add capacity just before upgrade? Maybe it will take also some bigger amount of time to get cluster healthy (don't know). But I know that upgrade of OCS passed - that's what I am testing in OCS upgrade test, and last step in verification is to see CEPH HEALTH_OK . So this logic is only used after upgrade currently, nowhere else.

I will take this discussion offline , if we want to have this code for testing purpose we can have it but the regression should be still fixed for specific timeout(with cushion - maybe extra 10% time) , somewhere the logic changed for the upgrade to fail and I believe either there is ocs bz or our pre_upgrade logic.

ocs_ci/utility/utils.py
shylesh: got it. thanks

ebenahar:
> Question - I am missing a global timeout in case `intelligent_wait` is True, for cases in which the recovery is extremely slow

@petr-balogh

ocs_ci/utility/utils.py
```diff
@@ -1302,25 +1302,104 @@ def get_az_count():
...
+        degraded_pgs_diff_times = []
+        try:
+            return retry(
+                (CommandFailed),
+                tries=tries,
+                delay=delay,
+                backoff=1
+            )(ceph_health_check_base)(namespace)
+        except CephHealthException as ex:
+            str_ex = str(ex)
```
phlogistonjohn, line 1350: This is a very large amount of code to place in an except block. I recommend moving the majority of the logic to a function that you can call from the except block.

ocs_ci/utility/utils.py
```diff
@@ -1302,25 +1302,104 @@ def get_az_count():
...
+                    )
+                    time.sleep(delay)
+                else:
+                    tries -= 1
+                    log.warn(f"Ceph health error: {ex}")
+                    log.info(f"Waiting for {delay}sec for next attempt.")
+                    time.sleep(delay)
+            else:
+                raise CephHealthException(
+                    "Ceph health didn't get fixed in timeout: "
```
phlogistonjohn: Perhaps: "Ceph health didn't improve within {}sec timeout." would be more grammatical and clearer.

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -105,7 +105,10 @@ def verify_image_versions(old_images, upgrade_version):
         old_images, selector=constants.CSI_RBDPLUGIN_PROVISIONER_LABEL,
         count=2
     )
-    verify_pods_upgraded(old_images, selector=constants.MON_APP_LABEL, count=3)
+    # Increased timeout for BZ https://bugzilla.redhat.com/show_bug.cgi?id=1856254
+    verify_pods_upgraded(
```
AviadP, line 109: with merge of this PR: https://github.com/red-hat-storage/ocs-ci/pull/2453, this part moved to ocs_upgrade.py

PR #2509, opened 2020-07-15
test_pvc_creation_measurement_performance improvement 

ypersky1980:
I've added deviation from average creation time condition.


tests/e2e/performance/test_pvc_creation_performance.py
```diff
@@ -62,24 +64,47 @@ def base_setup(
...
+        The test measures PVC creation times for sample volumes ( currently samples_num is 3)
+        makes sure create time is not greated than 3 seconds
+        calculates average creation time
+        and that the difference between creation time of each one of the PVCs and the average
+        is not more than 5%
         """
-        log.info('Start creating new PVC')
+        samples_num = 3
+        accepted_diff = 5
+        acceoted_create_time = 3
```
keesturam, line 75: typo in `accepted`

tests/e2e/performance/test_pvc_creation_performance.py
```diff
@@ -62,24 +64,47 @@ def base_setup(
...
+                raise ex.PerformanceException(
+                    f"PVC creation time is {create_time} and greater than {acceoted_create_time} second"
+                )
+            create_time_measures.append(create_time)
+            log.info(f"Current measures are {create_time_measures}")
+
+        average = statistics.mean(create_time_measures)
+        log.info(f"The average creation time for the sampled {samples_num} PVCs is {average}.")
+
+        for i in range(samples_num):
```
keesturam, line 101: maybe you could use statistics.variance instead? 
>>> statistics.variance([1.2,1.4,0.6])
0.1733333333333333

tests/e2e/performance/test_pvc_creation_performance.py
```diff
@@ -62,24 +64,47 @@ def base_setup(
...
+            helpers.wait_for_resource_state(pvc_obj, constants.STATUS_BOUND)
+            pvc_obj.reload()
+            teardown_factory(pvc_obj)
+            create_time = helpers.measure_pvc_creation_time(
+                self.interface, pvc_obj.name
+            )
+            logging.info(f"PVC created in {create_time} seconds")
+            if create_time > acceoted_create_time:
+                raise ex.PerformanceException(
+                    f"PVC creation time is {create_time} and greater than {acceoted_create_time} second"
```
keesturam, line 93: same as above

tests/e2e/performance/test_pvc_creation_performance.py
```diff
@@ -24,6 +25,7 @@ class TestPVCCreationPerformance(E2ETest):
     """
     pvc_size = '1Gi'
 
+
```
Avilir, line 28: This empty line need to be delete

tests/e2e/performance/test_pvc_creation_performance.py
```diff
@@ -62,24 +64,47 @@ def base_setup(
     @pytest.mark.usefixtures(base_setup.__name__)
     def test_pvc_creation_measurement_performance(self, teardown_factory, pvc_size):
         """
-        Measuring PVC creation time is less than 3 seconds
+        The test measures PVC creation times for sample volumes ( currently samples_num is 3)
+        makes sure create time is not greated than 3 seconds
+        calculates average creation time
+        and that the difference between creation time of each one of the PVCs and the average
+        is not more than 5%
```
Avilir, line 71: Pleas add to docstring explanation about the args. and empty line at the end

tests/e2e/performance/test_pvc_creation_performance.py
ypersky1980: @keesturam , 
Do you suggest that I calculate statistics.variance instead of statistics.mean ? 
Or statistics.variance instead of the diff value? 

Variance is the average of the squared differences from arithmetic mean. Therefore variance is calculated once per the whole group of create_time measures. 
Standard deviation is the square root of variance.
Diff as I though I should calculate is just the distance of each one of the time_create measures from the arithmetic mean of the whole group of measures. Diff as I calculate so far in the test is totally not standard deviation. 

Let's decide what should we calculate is the test and what should we compare to the mean ? Is is variance compared to the mean ( one comparison) of is if diff for each one of the create_time compared to the mean?

tests/e2e/performance/test_pvc_creation_performance.py
ypersky1980: @keesturam  - spelling fixed, thank you for pointing out!

tests/e2e/performance/test_pvc_creation_performance.py
ypersky1980: @keesturam - spelling fixed, thank you!

ypersky1980: @keesturam ,
Regarding the function that calls a certain operation N ( number of times) times and returns the statistics. 
Such function can be written and this is a matter of our decision. 
This idea has 2 disadvantages:
1) In case such function will only calculate statistics ( but not compare the statistics to the accepted values) - it will "do more job" than required in case one of the measured values is not valid. 
Let's say, if one of the create time values is 5 secs ( invalid) - the current code will through the exception, fail the test and will not calculate any statistics. 
The suggested function will still calculate average/variance/whatever statistics we decide and only then the test will be failed.
2) In case we want such function to calculate statistics ( and be able to fail the test) - this moment if will not be a general function since we will have to pass the accepted values to this function.  

So the question is ( in case I understand things right) is what we decide to do.

tests/e2e/performance/test_pvc_creation_performance.py
ypersky1980: @Avilir thank you, done!

tests/e2e/performance/test_pvc_creation_performance.py
ypersky1980: @Avilir , thank you! I will add description of the arguments to the docstring.

tests/e2e/performance/test_pvc_creation_performance.py
```diff
@@ -24,6 +24,32 @@ class TestPVCCreationPerformance(E2ETest):
     """
     pvc_size = '1Gi'
 
+    def samples_creation(self, samples_num, teardown_factory, pvc_size):
```
ebenahar, line 27: maybe change `samples_num` to `num_of_samples` to make the variable more descriptive

tests/e2e/performance/test_pvc_creation_performance.py
```diff
@@ -24,6 +24,32 @@ class TestPVCCreationPerformance(E2ETest):
...
+        time_measures = []
+        for i in range(samples_num):
+            log.info(f'Start creation of PVC number {i + 1}.')
+
+            pvc_obj = helpers.create_pvc(
+                sc_name=self.sc_obj.name, size=pvc_size)
+            helpers.wait_for_resource_state(pvc_obj, constants.STATUS_BOUND)
+            pvc_obj.reload()
+            teardown_factory(pvc_obj)
+            create_time = helpers.measure_pvc_creation_time(
```
ebenahar, line 45: Isn't the intention here to measure the PVC creation time? if so, why is the measurement being taken post `teardown_factory` call?

tests/e2e/performance/test_pvc_creation_performance.py
ebenahar, line 27: shouldn't the function be called 'measure_multiple_pvcs_creation_time`

tests/e2e/performance/test_pvc_creation_performance.py
```diff
@@ -62,23 +88,39 @@ def base_setup(
...
-        )
-        logging.info(f"PVC created in {create_time} seconds")
-        if create_time > 3:
+        create_measures = self.samples_creation(samples_num, teardown_factory, pvc_size)
+        log.info(f"Current measures are {create_measures}")
+
+        for i in range(samples_num):
+            if create_measures[i] > accepted_create_time:
+                raise ex.PerformanceException(
+                    f"PVC creation time is {create_measures[i]} and greater than {accepted_create_time} seconds."
```
ebenahar, line 112: ..and is greater than..

tests/e2e/performance/test_pvc_creation_performance.py
```diff
@@ -62,23 +88,39 @@ def base_setup(
...
+
+        average = statistics.mean(create_measures)
+        st_deviation = statistics.stdev(create_measures)
+        log.info(f"The average creation time for the sampled {samples_num} PVCs is {average}.")
+
+        st_deviation_percent = abs(st_deviation - average) / average * 100.0
+        if st_deviation > accepted_deviation_percent:
             raise ex.PerformanceException(
-                f"PVC creation time is {create_time} and greater than 3 second"
+                f"PVC create time deviation is {st_deviation_percent}%"
```
ebenahar, line 122: creation time

tests/e2e/performance/test_pvc_creation_performance.py
ebenahar: Also, can you please check if `tests.helpers.measure_pvc_creation_time` suites you? If so, maybe this function is not needed

tests/e2e/performance/test_pvc_creation_performance.py
ypersky1980: @ebenahar , good idea, done.

tests/e2e/performance/test_pvc_creation_performance.py
ypersky1980: @ebenahar  -
> Isn't the intention here to measure the PVC creation time? if so, why is the measurement being taken post `teardown_factory` call?
I was not sure whether the call for the teardown_factory should be counted as part of pvc creation time. 
In the test as it was before my modification, the code was exactly like this ( teardown_factory called before the measure was taken).

ypersky1980: @ebenahar 


> Also, can you please check if `tests.helpers.measure_pvc_creation_time` suites you? If so, maybe this function is not needed

tests.helpers.measure_pvc_creation_time definitely will not suit me, since I need to create and measure times for scalable number of PVCs. 
In case you referred to measure_pvc_creation_time function: 
It measures times for existing PVCs which is not the case here, since the function I've added also creates the PVCs.

tests/e2e/performance/test_pvc_creation_performance.py
ypersky1980: @ebenahar 
I agree that the name should be more descriptive. 
Let's do it "create_mutiple_pvcs_statistics "?
Otherwise it can be easily seen as a duplication of tests.helpers.measure_pvc_creation_time .

tests/e2e/performance/test_pvc_creation_performance.py
ypersky1980: @ebenahar Fixed, thank you!

tests/e2e/performance/test_pvc_creation_performance.py
ypersky1980: @ebenahar Thanks, fixed.

tests/e2e/performance/test_pvc_creation_performance.py
```diff
@@ -24,6 +24,32 @@ class TestPVCCreationPerformance(E2ETest):
     """
     pvc_size = '1Gi'
 
+    def samples_creation(self, samples_num, teardown_factory, pvc_size):
+        """
+        samples_creation creates number (samples_num) of PVCs, measures creation time for each PVC and returns list
```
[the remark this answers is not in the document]
ypersky1980: @ebenahar Done, in all the tests in the test_pvc_creation_performance.py.

tests/e2e/performance/test_pvc_creation_performance.py
keesturam: 1) pvc creation time is measured from the logs, so I don't think it matters whether it is done after the teardown
2) isn't `measure_pvc_creation_time` actually the function used here?

tests/e2e/performance/test_pvc_creation_performance.py
```diff
@@ -62,23 +93,40 @@ def base_setup(
...
-            self.interface, pvc_obj.name
-        )
-        logging.info(f"PVC created in {create_time} seconds")
-        if create_time > 3:
+        Args:
+            teardown factory: A fixture used when we want a new resource that was created during the tests
+                               to be removed in the teardown phase.
+            pvc_size: Size of the created PVC
+         """
+        num_of_samples = 3
```
ramkiperiy, line 108: @ypersky1980  Is this number is going to be fixed or we plan to change randomly, if fixed to minimal number like 3 PVCs then its good to use measure_pvc_creation_time() fun, but if the plan is to create n numbers i.e. more than 10 then would suggest to use https://github.com/red-hat-storage/ocs-ci/blob/master/tests/helpers.py#L1169

One more question, what kind of PVC are we testing here ? It would be good to capture the perf for all 4 types of PVC

tests/e2e/performance/test_pvc_creation_performance.py
ebenahar: sounds good

tests/e2e/performance/test_pvc_creation_performance.py
ypersky1980: @keesturam  I use measure_pvc_creation_time function inside the new create_mutiple_pvcs_statistics function. 
I think that Elad was asking why to write this function in the first place, when we have the measue_pvc_creation_time function. 
My reply was - we need a new function not only for measuring existing PVCs but to create number ( scaleable) of PVCs and then to return statistics on their creation times.

PR #2513, opened 2020-07-16
Couchbase workload with node reboot and node drain

shreekarSS:


wusui:
Other than the minor thing I mentioned in my previous comment, this lloks good to me.

shreekarSS:
Verified locally

PR #2514, opened 2020-07-16
Use OCP 4.4 stable channel for GA version

petr-balogh:


petr-balogh:
Travis is triggered here:  https://travis-ci.org/github/red-hat-storage/ocs-ci/builds/709078518 but it's already second time I see it's not reported back here

PR #2529, opened 2020-07-17
Fix return code mismatch for unit tests running on OSX

clacroix12:
Fixes: #2528 


clacroix12:
I locally verified this works on OSX and our travis-ci execution (that runs on Xenial) has passed so there is no impact to PR validation / non-osx executions.

Issue #2532, opened 2020-07-19
OCS4.4-->4.5 upgrade: does not create 2nd rook-ceph-rgw pod


AviadP:
fresh OCS 4.5 cluster created 2 rook-ceph-rgw pods, which fails ` verify_image_versions()` during OCS upgrade validation 
workaround might be needed, for more details see: https://bugzilla.redhat.com/show_bug.cgi?id=1857802

ebenahar:
Re opening as the fixes for https://bugzilla.redhat.com/show_bug.cgi?id=1857802 were reverted and we will have 1 RGW post upgrade

AviadP:
@ebenahar, im closing this issue as bug fixed already, and it refer to very OCS versions

PR #2533, opened 2020-07-19
W/A for bug 1857802 - only 1 RGW pod post upgrade to 4.5

ebenahar:
Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1857802.
To be tracked with https://github.com/red-hat-storage/ocs-ci/issues/2532

### DIFF
--- a/ocs_ci/ocs/ocs_upgrade.py
+++ b/ocs_ci/ocs/ocs_upgrade.py
@@ -116,7 +116,13 @@ def verify_image_versions(old_images, upgrade_version):
     )
     verify_pods_upgraded(old_images, selector=constants.MDS_APP_LABEL, count=2)
     if config.ENV_DATA.get('platform') in constants.ON_PREM_PLATFORMS:
-        rgw_count = 2 if float(config.ENV_DATA['ocs_version']) >= 4.5 else 1
+        # Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1857802 - RGW count is 1
+        # post upgrade to OCS 4.5. Tracked with
+        # https://github.com/red-hat-storage/ocs-ci/issues/2532
+        # TODO: uncomment the below 1 line:
+        # rgw_count = 2 if float(config.ENV_DATA['ocs_version']) >= 4.5 else 1
+        # TODO: Delete the below 1 line
+        rgw_count = 1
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -116,7 +116,13 @@ def verify_image_versions(old_images, upgrade_version):
     )
     verify_pods_upgraded(old_images, selector=constants.MDS_APP_LABEL, count=2)
     if config.ENV_DATA.get('platform') in constants.ON_PREM_PLATFORMS:
-        rgw_count = 2 if float(config.ENV_DATA['ocs_version']) >= 4.5 else 1
+        # Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1857802 - RGW count is 1
+        # post upgrade to OCS 4.5. Tracked with
+        # https://github.com/red-hat-storage/ocs-ci/issues/2532
+        # TODO: uncomment the below 1 line:
+        # rgw_count = 2 if float(config.ENV_DATA['ocs_version']) >= 4.5 else 1
```
nehaberry, line 123: @ebenahar since this is commented out, Upgrade from OCS 4.5 to OCS 4.6 is failing (as RGW cunt would be 2 )

I wanted few clarifiactions before sending the fix(sorry for lame questions)
1. here [ocs_version] is pre-upgrade version. right ?
   If yes, then why did we comment it out ?  wouldn't it have worked for OCS 4.4 to 4.5 upgrade too (where **else 1** would have been activated)
2. if we meant post_upgrade version in this loop, then we would have used **upgrade_version** instead. right ?

IMO, to fix all upgrade scenarios(be it 4.4 to 45, 4.5 to 4.6 and beyond), if **`ocs_version`** is indeed pre-upgrade version, we just need to un-comment this line.

Let me know if my understanding is incorrect, I did not want to send the fix before verifying the reason.

ebenahar: Hi Neha, very good catch!

1. The version here is post upgrade

Seems like we simply need to replace lines 119-125 with:
```python
rgw_count = 2 if float(config.ENV_DATA['ocs_version']) >= 4.6 else 1
```

PR #2543, opened 2020-07-21
Improved AWS cred-req flow, create cred-reqs only for tests that require them

### LINKED ISSUE #2542: CredentialsRequest check times out and leaves bucket after running
The current 40 seconds dedicated to checking the credentials do not seem enough, thus leading to test failures because of timeouts.
Also, sometimes, buckets are left over after the check ran - we need to find a better way to check the credentials (preferably one that is also not dependent on bucket creation)

MeridianExplorer:
Closes #2542

### DIFF
--- a/ocs_ci/ocs/resources/mcg.py
+++ b/ocs_ci/ocs/resources/mcg.py
@@ -28,7 +28,7 @@ import stat
 logger = logging.getLogger(name=__file__)
 
 
-class MCG(object):
+class MCG:
     """
     Wrapper class for the Multi Cloud Gateway's S3 service
     """
@@ -39,7 +39,7 @@ class MCG(object):
         namespace, noobaa_user, noobaa_password, noobaa_token
     ) = (None,) * 12
 
-    def __init__(self):
+    def __init__(self, *args, **kwargs):
         """
         Constructor for the MCG class
         """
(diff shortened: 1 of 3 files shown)


ocs_ci/ocs/resources/mcg.py
```diff
@@ -117,7 +117,10 @@ def __init__(self):
 
         self.s3_client = self.s3_resource.meta.client
 
-        if config.ENV_DATA['platform'].lower() == 'aws':
+        if (
+            config.ENV_DATA['platform'].lower() == 'aws'
+            and kwargs.get('create_aws_creds')
```
ebenahar, line 122: what if `create_aws_creds` doesn't exist in kwargs? Please add a default:

```python
kwargs.get('create_aws_creds', 'default value')
```

tests/conftest.py
```diff
@@ -1546,21 +1546,32 @@ def mcg_obj_session(request):
...
 
-    mcg_obj = MCG()
+    mcg_obj = MCG(*args, **kwargs)
 
     def finalizer():
         if config.ENV_DATA['platform'].lower() == 'aws':
             mcg_obj.cred_req_obj.delete()
 
-    request.addfinalizer(finalizer)
+    if kwargs.get("create_aws_creds"):
```
ebenahar, line 1573: same, add a default value

ocs_ci/ocs/resources/mcg.py
MeridianExplorer: `.get()`returns `None` by default, which will be evaluated as `False` in this scenario - so this is working as intended

tests/conftest.py
MeridianExplorer: Ditto

ocs_ci/ocs/resources/mcg.py
ebenahar: ack, thanks

PR #2548, opened 2020-07-22
consider cluster wide proxy configuration in tests

dahorak:
If cluster wide proxy is configured (in Proxy configuration object), use it's values for `http_proxy` and `https_proxy` values in tests. This should enable test execution on proxy environment without special manual configuration.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -2391,8 +2391,8 @@ def get_cluster_proxies():
     else:
         ocp_obj = ocp.OCP(kind=constants.PROXY, resource_name='cluster')
         proxy_obj = ocp_obj.get()
-        http_proxy = proxy_obj['spec'].get('httpProxy', '')
-        https_proxy = proxy_obj['spec'].get('httpsProxy', '')
+        http_proxy = proxy_obj.get('spec', {}).get('httpProxy', '')
+        https_proxy = proxy_obj.get('spec', {}).get('httpsProxy', '')
     logger.info("Using http_proxy: '%s'", http_proxy)
     logger.info("Using https_proxy: '%s'", https_proxy)
     return http_proxy, https_proxy
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -234,7 +234,8 @@ def create_pod(
 
     # configure http[s]_proxy env variable, if required
     try:
-        if 'http_proxy' in config.ENV_DATA:
+        http_proxy, https_proxy = get_cluster_proxies()
+        if http_proxy:
             if 'containers' in pod_data['spec']:
                 container = pod_data['spec']['containers'][0]
             else:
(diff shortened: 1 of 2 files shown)


dahorak:
[tier1 tests execution on cluster behind proxy](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/10177] - `http_proxy` and `https_proxy` were properly configured automatically:
```
...
12:44:47 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig get Proxy cluster -o yaml
12:44:47 - MainThread - tests.helpers - INFO - Using http_proxy: 'http://user:password@proxy:3128'
12:44:47 - MainThread - tests.helpers - INFO - Using https_proxy: 'http://user:password@proxy:3128'
...
```

tests/helpers.py
```diff
@@ -2370,3 +2366,33 @@ def get_pv_size(storageclass=None):
...
+    """
+    if 'http_proxy' in config.ENV_DATA:
+        http_proxy = config.ENV_DATA['http_proxy']
+        https_proxy = config.ENV_DATA.get(
+            'https_proxy', config.ENV_DATA['http_proxy']
+        )
+    else:
+        ocp_obj = ocp.OCP(kind=constants.PROXY, resource_name='cluster')
+        proxy_obj = ocp_obj.get()
+        http_proxy = proxy_obj['spec'].get('httpProxy', '')
```
petr-balogh, line 2394: Based on our offline conversation where I have asked about if we will have Proxy object even on non proxy env I've got answer from Daniel that he found in documentation:

```
When a cluster is installed or upgraded without the proxy configured, a Proxy object is still generated but it will have a nil spec.
```

So my understanding is that spec can be None in python, so to be on safe side I would rather go with something like:
```suggestion
        http_proxy = proxy_obj.get('spec', {}).get('httpProxy', '')
```

The same below.

Otherwise LGTM

petr-balogh:
Please run verification on non proxy cluster as well.

Thanks

dahorak:
This was verification job on non proxy cluster.

> PR validation
> 
> Cluster Name:
> Cluster Configuration:
> PR Test Suite: tier1
> PR Test Path: tests/
> Additional Test Params:
> OCP VERSION: 4.5
> OCS VERSION: 4.5
> 
> [Job UNSTABLE](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/2391//testReport) (some or all tests failed).

tests/helpers.py
dahorak: Fixed by 166c156.

mbukatov: Spec can't be `None`, neither it can be removed from the CRD.

dahorak: @mbukatov thanks for checking, this way it is safe...

Issue #2549, opened 2020-07-22
AWS UPI leftovers in claudformations 


petr-balogh:
I've noticed a lot of failed claudformations stacks cause of it failed to delete Hosted Zone.

<img width="1084" alt="Screen Shot 2020-07-22 at 10 53 24 AM" src="https://user-images.githubusercontent.com/16169617/88156329-92ed0400-cc09-11ea-8890-8f23c0027f9d.png">

<img width="1084" alt="Screen Shot 2020-07-22 at 10 53 55 AM" src="https://user-images.githubusercontent.com/16169617/88156393-a4cea700-cc09-11ea-99e0-866129300d72.png">

There are 3 recordsets and if *.apps.jnk-pr2007-b1889.qe.rh-ocs.com. exists, we cannot delete the the hosted zone.

After deleting this 1 recordset, I am able to delete stack

@clacroix12, can you please take a look? And I think we will need to improve the force cleanup script to deal with this.

Thanks

clacroix12:
@petr-balogh Yep I've definitely noticed this behavior in the past. I'll look into explicitly deleting that recordset before iterating through the cloudformation stacks for deletion.

petr-balogh:
thanks

petr-balogh:
Hey @clacroix12 , I've just noticed this one one of the teardown job:
```    
        if http.status_code >= 300:
            error_code = parsed_response.get("Error", {}).get("Code")
            error_class = self.exceptions.from_code(error_code)
>           raise error_class(parsed_response, operation_name)
E           botocore.errorfactory.NoSuchBucket: An error occurred (NoSuchBucket) when calling the ListObjects operation: The specified bucket does not exist

.venv/lib/python3.6/site-packages/botocore/client.py:635: NoSuchBucket
=============================== warnings summary ===============================```

I don't think we should fail the upgrade if bucket not found but just continue here.

So re-opening this.

Job:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-destroy-ocs-cluster/7945/console

clacroix12:
@petr-balogh The error is not related to this issue, but rather an issue with `delete_cluster_buckets` that was added [here](https://github.com/red-hat-storage/ocs-ci/pull/2420/files)

petr-balogh:
Sorry I didn't get to deep of root cause, just knew about this recent change and thought it's related. So you can close the issue and open the new one please @clacroix12

PR #2561, opened 2020-07-23
Delete apps record set before destroying security group stack for AWS UPI teardown

clacroix12:
Fixes: #2549 

### DIFF
--- a/ocs_ci/utility/aws.py
+++ b/ocs_ci/utility/aws.py
@@ -38,6 +38,7 @@ class AWS(object):
     _ec2_resource = None
     _region_name = None
     _s3_client = None
+    _route53_client = None
 
     def __init__(self, region_name=None):
         """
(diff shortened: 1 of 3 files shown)


ocs_ci/utility/aws.py
```diff
@@ -953,6 +970,47 @@ def create_stack(self, s3_url, index, params_list, capabilities):
         logger.info(f"Stackid = {stack_id}")
         return stack_name, stack_id
 
+    def delete_apps_record_set(self):
+        """
+        Delete apps record set that sometimes blocks sg stack deletion.
+            https://github.com/red-hat-storage/ocs-ci/issues/2549
+
+        """
+        cluster_name = config.ENV_DATA['cluster_name']
```
petr-balogh, line 979: How those values get's propagated when we run aws cleanup script? Don't we have the same issue like with stack name that cluster_name is not defined at this point in config?

clacroix12: Good point, I forgot to allow passing the cluster_name through as an optional parameter. We can simply pass that during the more general cleanup and pull it from the config during standard cleanup.

clacroix12: Done

Issue #2567, opened 2020-07-27
Require Parameter for hostPrefix in install.conf during cluster deployment.


ramkiperiy:
In Scale testing there is a need to create more than 500/1000 pods per node but after creation of 510 pods there is a IP limit problem on each node for more details please refer [1] point 8 related to hostPrefix. Due to the IP limits after creation of 500 pods new pod creation will be stuck in ContainerCreation state [2]

This change should be done at the start the cluster deployment i.e. it should be changed in install-config.yaml.

[1] https://docs.openshift.com/container-platform/4.1/installing/installing_bare_metal/installing-bare-metal.html#installation-bare-metal-config-yaml_installing-bare-metal
[2] pod error from describe output
`Warning  FailedCreatePodSandBox  10s (x245 over 58m)  kubelet, compute-2  (combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create pod network sandbox k8s_pod-test-cephfs-26636c33118d4c3095cbd37d91ffbdcd_namespace-test-04da39dd13384c368442462a0a46b146_d78b6a81-5a2e-47fd-becf-5baf3d787ecf_0(3aaed0634517bee401dba2ae417eba78ab66cd785cb5e05460cb334f5ae895b4): Multus: [namespace-test-04da39dd13384c368442462a0a46b146/pod-test-cephfs-26636c33118d4c3095cbd37d91ffbdcd]: error adding container to network "openshift-sdn": delegateAdd: error invoking confAdd - "openshift-sdn": error in getting result from AddNetwork: CNI request failed with status 400: 'failed to run IPAM for 3aaed0634517bee401dba2ae417eba78ab66cd785cb5e05460cb334f5ae895b4: failed to run CNI IPAM ADD: failed to allocate for range 0: no IP addresses available in range set: 10.130.2.1-10.130.3.254`

PR #2569, opened 2020-07-27
Include hostprefix in vSphere

vavuthu:
Fixes: #2567 

### DIFF
--- a/ocs_ci/templates/ocp-deployment/install-config-vsphere-upi.yaml.j2
+++ b/ocs_ci/templates/ocp-deployment/install-config-vsphere-upi.yaml.j2
@@ -4,8 +4,8 @@ metadata:
   name: '{{ cluster_name }}'
 networking:
   clusterNetwork:
-    - cidr: 10.128.0.0/14
-      hostPrefix: 23
+    - cidr: {{ cluster_network_cidr | default('10.128.0.0/14') }}
+      hostPrefix: {{ cluster_host_prefix | default(23) }}
   machineCIDR: '{{ machine_cidr }}'
 {% if fips %}
 fips: {{ fips }}
--- a/ocs_ci/templates/ocp-deployment/install-config-vsphere-upi.yaml.j2
+++ b/ocs_ci/templates/ocp-deployment/install-config-vsphere-upi.yaml.j2
@@ -3,6 +3,9 @@ baseDomain: {{ base_domain | default('qe.rh-ocs.com') }}
 metadata:
   name: '{{ cluster_name }}'
 networking:
+  clusterNetwork:
+    - cidr: 10.128.0.0/14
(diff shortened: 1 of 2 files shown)


ocs_ci/templates/ocp-deployment/install-config-vsphere-upi.yaml.j2
```diff
@@ -3,6 +3,9 @@ baseDomain: {{ base_domain | default('qe.rh-ocs.com') }}
 metadata:
   name: '{{ cluster_name }}'
 networking:
+  clusterNetwork:
+    - cidr: 10.128.0.0/14
+      hostPrefix: 23
```
AviadP, line 8: do we plan more than 254 hosts?

petr-balogh: What about have this value configurable via config?

```
    - cidr:  {{ cluster_cidr | default("10.128.0.0/14") }}
      hostPrefix: {{ cluster_host_prefix | default(23) }}
```

this will give us capability to change this via config value in ENV_DATA section. WDYT? But not sure if we will need to change this value?

If it's needed for scale, we can have specific config for them maybe, which they will use when deploying, so maybe we can have just condition if cluster_cidr data are in config and if so add this section in networking.

vavuthu: Done

vavuthu: > do we plan more than 254 hosts?

In recent commit , values are configurable.

ocs_ci/templates/ocp-deployment/install-config-vsphere-upi.yaml.j2
```diff
@@ -3,6 +3,9 @@ baseDomain: {{ base_domain | default('qe.rh-ocs.com') }}
 metadata:
   name: '{{ cluster_name }}'
 networking:
+  clusterNetwork:
+    - cidr: {{ cluster_network_cidr | default('10.128.0.0/14') }}
+      hostPrefix: {{ cluster_host_prefix | default(23) }}
```
ramkiperiy, line 8: do we have cluster_host_prefix defined in default_config.yaml ?

ramkiperiy: I guess its not required, can use param cluster_host_prefix in another file which can override value 23

petr-balogh: The default value in default_config is not needed as it will be taken from | default(23) .  But if we have some config file, it will be taken from there instead of it.

PR #2570, opened 2020-07-27
Wait for rebalance post add capacity test

ebenahar:

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -644,17 +644,33 @@ class CephCluster(object):
                 and states['count'] == total_pg_count
             )
 
-    def wait_for_rebalance(self, timeout=600):
+    def wait_for_rebalance(self, timeout=None):
         """
         Wait for re-balance to complete
 
         Args:
-            timeout (int): Time to wait for the completion of re-balance
+            timeout (int): Time to wait for the completion of re-balance.
+                In case IO is running in the test background (according to --io-in-bg
+                run-ci param, the timeout will be determined by the amount of load
+                (io_load run-ci param)
 
         Returns:
             bool: True if rebalance completed, False otherwise
 
         """
+        if config.RUN.get('io_in_bg'):
+            io_load = config.RUN.get('io_load')
+            if 10 < io_load <= 30:
+                timeout = 900
+            if 30 < io_load <= 50:
+                timeout = 1800
+            if 50 < io_load <= 70:
+                timeout = 2700
+            if io_load > 70:
+                timeout = 3600
+        else:
+            timeout = timeout if timeout else 600
+
         try:
             for rebalance in TimeoutSampler(
                 timeout=timeout, sleep=10, func=self.get_rebalance_status
@@ -682,6 +698,7 @@ class CephCluster(object):
 
         """
         start_time = time.time()
+        timeout = None if config.RUN.get('io_in_bg') else timeout
         assert self.wait_for_rebalance(timeout=timeout), (
             f"Data re-balance failed to complete within the given "
             f"timeout of {timeout} seconds"
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -644,7 +644,7 @@ class CephCluster(object):
                 and states['count'] == total_pg_count
             )
 
-    def wait_for_rebalance(self, timeout=None):
+    def wait_for_rebalance(self, timeout=600):
         """
         Wait for re-balance to complete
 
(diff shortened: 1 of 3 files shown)


petr-balogh:
@ebenahar  I see that verification which I've triggered failed:
```
        ceph_cluster_obj = CephCluster()
>       assert ceph_cluster_obj.wait_for_rebalance(), (
            "Data re-balance failed to complete"
        )
E       AssertionError: Data re-balance failed to complete
E       assert False
E        +  where False = <bound method CephCluster.wait_for_rebalance of <ocs_ci.ocs.cluster.CephCluster object at 0x7fa944e4dcf8>>()
E        +    where <bound method CephCluster.wait_for_rebalance of <ocs_ci.ocs.cluster.CephCluster object at 0x7fa944e4dcf8>> = <ocs_ci.ocs.cluster.CephCluster object at 0x7fa944e4dcf8>.wait_for_rebalance```

 https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Upgrade-OCS/job/qe-trigger-aws-ipi-3az-rhcos-3m-3w-upgrade-ocs-auto/33/console

I see that you've increased time, but will this be enough for upgrade to running IO in the background?

**Comment by ebenahar** [comment]
@petr-balogh Looks like the add capacity test succeeds with Ceph health reaching OK and PGs are active+clean, but for some reason, during teardown, Ceph health becomes HEALTH_WARN with some few degraded objects.

See in here:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/10303/console

Not sure if it's a temp issue or not..

**Comment by petr-balogh** [comment]
> @petr-balogh Looks like the add capacity test succeeds with Ceph health reaching OK and PGs are active+clean, but for some reason, during teardown, Ceph health becomes HEALTH_WARN with some few degraded objects.
> 
> See in here:
> https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/10303/console
> 
> Not sure if it's a temp issue or not..

That's sound to me like what we usually see in the upgrade as well. Like we start upgrade after  ~3 mins add_capacity finished, Ceph health is OK before we start upgrade, and once upgrade start we see degraded PGs.  So maybe there is some delay till everything start completing.  Maybe it's expected to see degraded PGs after adding capacity:

As they mentioning here:
https://docs.ceph.com/docs/master/rados/operations/add-or-rm-osds/#observe-the-data-migration
```
You should see the placement group states change from active+clean to active, some degraded objects, and finally active+clean when migration completes. (Control-c to exit.)```

Maybe we need to wait till we will really see some degraded objects and then finally see: active+clean when migration completes

Currently we don't wait for degraded objects right?  But not sure if it's always like that.

ocs_ci/ocs/cluster.py
```diff
@@ -644,6 +644,47 @@ def get_rebalance_status(self):
...
+                In case IO is running in the test background (according to --io-in-bg
+                run-ci param, the timeout will be determined by the amount of load
+                (io_load run-ci param)
+
+        Returns:
+            bool: True if rebalance completed, False otherwise
+
+        """
+        if config.RUN.get('io_in_bg'):
+            io_load = config.RUN.get('io_load')
```
petr-balogh, line 662: is this value by default 0? If not please use default value here in get to have 0

ocs_ci/ocs/cluster.py
```diff
@@ -644,6 +644,47 @@ def get_rebalance_status(self):
...
+            if 10 < io_load <= 30:
+                timeout = 900
+            if 30 < io_load <= 50:
+                timeout = 1800
+            if 50 < io_load <= 70:
+                timeout = 2700
+            if io_load > 70:
+                timeout = 3600
```
petr-balogh, line 670: With heavy load I think that even 1 hour is not enough.  But let's have some testing.  It maybe also depends on number of OSDs we have and amount of data.  Maybe worth to ask some Ceph guys if we can somehow calculate this more specific.

petr-balogh: Asked here:  https://bugzilla.redhat.com/show_bug.cgi?id=1856975#c9. So hope we will get some answer, maybe from Josh.

vasukulkarni:
I hope this test doesn't run along with upgrade, it is just messing up the full upgrade suite, As Josh pointed out the huge misplaced PGs is caused by mixing this test without insufficient wait and kicking off upgrade at the same time. We should tweak so that add osd's happen after upgrade not before.

ocs_ci/ocs/cluster.py
ebenahar: the suggested solution from Josh is to implement something similar to https://github.com/ceph/ceph/blob/master/qa/tasks/ceph_manager.py#L2576.

See http://post-office.corp.redhat.com/archives/rhocs-eng/2020-July/msg00325.html

ebenahar:
@vasukulkarni this PR has nothing to do with upgrade. The intention is to make sure the PGs are active+clean post cluster expansion tests.

ocs_ci/ocs/cluster.py
ebenahar: For now, in order to unblock the upgrade executions, I will set a static timeout of 1800 for add capacity and add node tests.
We can think about implementing a solution as suggested above, but this will take some time to get merged

ebenahar: changed to 3600 seconds timeout

ebenahar:
The add capacity test passes with all PGs clean. 
However, the upgrade test failed with what seems to be 1 mon that failed to be upgraded -https://bugzilla.redhat.com/show_bug.cgi?id=1856254

Will examine this and re-open the BZ if needed

PR #2579, opened 2020-07-27
Refine dataclasses dependency

zmc:
The previous workaround to dataclasses' requirement on python < 3.6 pins
to the version before the requirement was added; while pinning is of
course a good idea, I wanted to offer a way to pin to the latest version
while preventing the module from being installed at all on python 3.7+,
where the functionality is part of the standard library.


zmc:
@gabriellasroman here's the fix we had talked about

PR #2581, opened 2020-07-28
Test case to verify worker node restart will not hinder PVC expansion

jilju:
OCS-2235 - FT-PvcExpand-Worker node restart during PVC expansion

Test case to verify PVC expansion will succeed if a worker node is restarted during expansion

### DIFF
--- a/tests/manage/pv_services/pvc_resize/test_node_restart_during_pvc_expansion.py
+++ b/tests/manage/pv_services/pvc_resize/test_node_restart_during_pvc_expansion.py
@@ -1,0 +1,109 @@
+import logging
+import pytest
+from concurrent.futures import ThreadPoolExecutor
+
+from ocs_ci.ocs import constants, node
+from ocs_ci.framework.testlib import (
+    skipif_ocs_version, ManageTest, tier4, tier4b, ignore_leftovers,
+    polarion_id
+)
+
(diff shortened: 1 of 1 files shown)


tests/manage/pv_services/pvc_resize/test_node_restart_during_pvc_expansion.py
```diff
@@ -0,0 +1,111 @@
...
+
+        log.info("Wait for IO to complete on all pods")
+        for pod_obj in self.pods:
+            pod_obj.io_proc.result()
+            fio_result = pod_obj.get_fio_results()
+            err_count = fio_result.get('jobs')[0].get('error')
+            assert err_count == 0, (
+                f"IO error on pod {pod_obj.name}. "
+                f"FIO result: {fio_result}"
+            )
```
sidhant-agrawal, line 109: Does it make sense to move this to `get_fio_results()` itself?

jilju: This is just an additional verification. In most of the error cases, return value of fio command will be non zero. So it be caught in exec_cmd itself.

sidhant-agrawal: ok, thanks

PR #2584, opened 2020-07-28
downstream-ci: change stable tagging for versions that use bundles

andrewschoen:
OCS versions from 4.5 onward should use the ocs-bundle-operator instead of ocs-olm-operator. If the tests pass, make sure to tag ocs-bundle-operator as ``latest-stable-4.*`` instead of ocs-olm-operator for those versions. Also, the old ocs-registry image won't be tagged anymore at all.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -134,15 +134,20 @@ pipeline {
         // quay.io/rhceph-dev/ocs-registry:4.2.2-255.ci -> 4.2.2-255.ci
         def registry_tag = registry_image.split(':')[-1]
         def ocs_version = registry_tag.split('-')[0]
-        // tag ocs-registry container as 'latest-stable-$ocs_version'
-        build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-registry:latest-stable-${ocs_version}")]
-        // tag ocs-olm-operator container as 'latest-stable'
-        build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-olm-operator:latest-stable-${ocs_version}")]
         // ocs version with only X.Y
         def short_ocs_version = ocs_version.tokenize(".").take(2).join(".")
-        build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-registry:latest-stable-${short_ocs_version}")]
-        // tag ocs-olm-operator container as 'latest-stable'
-        build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-olm-operator:latest-stable-${short_ocs_version}")]
+        if (short_ocs_version.toFloat() < 4.5 ) {
+          // tag ocs-olm-operator container because versions < 4.5 use olm registry containers
+          build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-olm-operator:latest-stable-${ocs_version}")]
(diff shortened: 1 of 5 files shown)


conf/ocs_version/ocs-4.5.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 DEPLOYMENT:
-  default_ocs_registry_image: "quay.io/rhceph-dev/ocs-olm-operator:latest-4.5"
+  default_ocs_registry_image: "quay.io/rhceph-dev/ocs-bundle-operator:latest-4.5"
```
vasukulkarni, line 3: do we need this for 4.4 as well?

andrewschoen: No, 4.4 does not support bundles.

petr-balogh:
@andrewschoen we have also use of it here:

https://github.com/red-hat-storage/ocs-ci/blob/08377c3e2332a5206aae7acd4316c81fd35e8cad/ocs_ci/ocs/constants.py#L608-L610

Which is important for us as we change latest to some specific tag.

So we will need to deal with it as well.

conf/ocs_version/ocs-4.5.yaml
petr-balogh, line 3: The bundle image is not deployable itself. Per Boris Ranto's comment you need to use:
quay.io/rhceph-dev/ocs-registry:4.5.0-30.ci

Index image for deployment.

conf/ocs_version/ocs-4.6.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 DEPLOYMENT:
-  default_ocs_registry_image: "quay.io/rhceph-dev/ocs-olm-operator:latest-4.6"
+  default_ocs_registry_image: "quay.io/rhceph-dev/ocs-bundle-operator:latest-4.6"
```
petr-balogh, line 3: The same here.

Also we still need to somehow dynamically update value from constants which I pointed out in different general comment in this PR.

conf/ocs_version/ocs-4.5.yaml
b-ranto: btw: OCP 4.4 does include support for bundles but we (RH) don't support it.

b-ranto: Yes, OCP does not support the deployment of bundle images. Apparently, it was discussed but it was not included in the release.

Jenkinsfile
```diff
@@ -134,15 +134,18 @@ pipeline {
...
+        if (short_ocs_version.toFloat() < 4.5 ) {
+          // tag ocs-olm-operator container because versions < 4.5 use olm registry containers
+          build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-olm-operator:latest-stable-${ocs_version}")]
+          build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-olm-operator:latest-stable-${short_ocs_version}")]
+        }
+        else {
+          // tag ocs-bundle-operator container because versions > 4.5 use bundle index containers
+          build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-bundle-operator:latest-stable-${ocs_version}")]
+          build job: 'quay-tag-image', parameters: [string(name: "SOURCE_URL", value: "${registry_image}"), string(name: "QUAY_IMAGE_TAG", value: "ocs-bundle-operator:latest-stable-${short_ocs_version}")]
```
b-ranto, line 147: You may want to tag ocs-registry, here, too.

andrewschoen:
> @andrewschoen we have also use of it here:
> 
> https://github.com/red-hat-storage/ocs-ci/blob/08377c3e2332a5206aae7acd4316c81fd35e8cad/ocs_ci/ocs/constants.py#L608-L610
> 
> Which is important for us as we change latest to some specific tag.
> 
> So we will need to deal with it as well.

@petr-balogh I added 3e620739302815e04bff42faa7071245b4f3a54e to deal with this concern. I don't think this will work for upgrades will it? If we upgrade for 4.4 that does not use bundles to 4.5 which does will ocs-ci do the right thing?

I'm testing with a `latest-stable-4.5` here: https://ceph-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/OCS%204.X%20CI%20CD/job/ocs-ci/538/

Issue #2571, opened 2020-07-27
Add retry if oc debug nodes commaned failed with error: Back-off pulling image "registry.redhat.io/rhel7/support-tools"


jilju:
Failure of oc debug nodes command with error ```error: Back-off pulling image "registry.redhat.io/rhel7/support-tools"``` should have a retry mechanism. The error is temporary and therefore this can be retried after few seconds.

```
18:02:37 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc debug nodes/ip-10-0-149-60.us-east-2.compute.internal -- df
18:02:42 - ThreadPoolExecutor-381_0 - ocs_ci.utility.utils - WARNING - Command stderr: Error from server (NotFound): pods "pod-test-cephfs-2e71ee0092804860aad497c93d6c7f5d" not found

18:03:34 - MainThread - ocs_ci.utility.utils - WARNING - Command stderr: Starting pod/ip-10-0-149-60us-east-2computeinternal-debug ...
To use host binaries, run `chroot /host`

Removing debug pod ...
error: Back-off pulling image "registry.redhat.io/rhel7/support-tools"
```

rachael-george:
These errors were seen in the run mentioned below and could also benefit from retry options

**ocs-ci results for OCS4-8-Downstream-OCP4-8-VSPHERE6-UPI-FIPS-1AZ-RHCOS-VSAN-3M-6W-tier4a (BUILD ID: 4.8.7-2 RUN ID: 1641502355)**

```
2022-01-07 09:07:07,112 - MainThread - INFO - ocs_ci.utility.utils.exec_cmd.487 - Executing command: oc debug nodes/compute-7 -- chroot /host /bin/bash -c "ifconfig $(route | grep default | awk '{print $(NF)}') down || echo 'CMD FAILED'; "
2022-01-07 09:07:07,698 - MainThread - DEBUG - ocs_ci.utility.utils.exec_cmd.502 - Command stdout is empty
2022-01-07 09:07:07,698 - MainThread - WARNING - ocs_ci.utility.utils.exec_cmd.506 - Command stderr: Starting pod/compute-7-debug ...
To use host binaries, run `chroot /host`

Removing debug pod ...
error: unable to create the debug pod "compute-7-debug"
```

jilju:
@vavuthu Can we reopen #2587 ?

vavuthu:
i will reopen

Issue #1604, opened 2020-03-06
Add retry for oc command if it fails due to TLS handshake timeout


sidhant-agrawal:
Sometimes the test case fails due to TLS handshake timeout while running an oc command. 
```
oc -n namespace-test-1aabf5e952a449ee808817650af14812 --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig get PersistentVolumeClaim pvc-test-f9e4b40037e14bd1b22af8320f8675c1 -n namespace-test-1aabf5e952a449ee808817650af14812 -o yaml. Error is Unable to connect to the server: net/http: TLS handshake timeout
```
This seems to be a temporary issue.It might be solved if we add retries for oc commands.

One of the failed testcase - [here](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/5188/testReport/junit/tests.manage.pv_services.test_pvc_disruptive/TestPVCDisruption/test_pvc_disruptive_CephFileSystem_create_pod_cephfsplugin_provisioner_/)

sidhant-agrawal:
Issue not seen in recent runs. Closing this old issue.

PR #2587, opened 2020-07-29
Adding function for debug nodes

vavuthu:
Fixes: #2571 

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -2433,3 +2433,20 @@ def configure_chrony_and_wait_for_machineconfig_status(
         # sleep here to start update machineconfigpool status
         time.sleep(60)
         wait_for_machineconfigpool_status(role)
+
+
+@retry(CommandFailed)
+def exec_cmd_on_debug_node(node, cmd):
+    """
+    Execute command on a node using debug pod
+
+    Args:
+        node (str): Node to create debug pod
+        cmd (str): Command to run on debug pod
+
+    Returns:
+        str: stdout of command
+
+    """
+    command = f"oc debug nodes/{node} -- {cmd}"
+    return run_cmd(command)
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -33,7 +33,6 @@ from ocs_ci.utility.retry import retry
 from ocs_ci.utility.utils import (
     TimeoutSampler,
     ocsci_log_path,
-    run_cmd,
     update_container_with_mirrored_image,
     exec_cmd_on_debug_node,
 )
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -35,6 +35,7 @@ from ocs_ci.utility.utils import (
     ocsci_log_path,
     run_cmd,
     update_container_with_mirrored_image,
+    exec_cmd_on_debug_node,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


tests/helpers.py
```diff
@@ -1692,8 +1692,8 @@ def verify_pv_mounted_on_node(node_pv_dict):
     """
     existing_pvs = {}
     for node_name, pvs in node_pv_dict.items():
-        cmd = f'oc debug nodes/{node_name} -- df'
-        df_on_node = run_cmd(cmd)
+        cmd = "df"
+        df_on_node = exec_cmd_on_debug_node(node_name, cmd)
```
vasukulkarni, line 1696: run_cmd should work?

clacroix12: It does but this PR is adding `exec_cmd_on_debug_node` for explicitly running commands through the debug node. It's also decorated with a retry to address #2571

vasukulkarni: I missed the retry but then I believe its a test issue @jilju , shouldn't test retry when it knows command can fail? this will cause all commands to retry

jilju: The issue is not limited to any test case. The error could show up in any occurrence of "oc debug node" command.
Retry only if error is 'Back-off pulling image ..'.

petr-balogh: @jilju but current implementation cover all the cases of CommandFailed exception which is not anyhow restricted only for this case: Back-off pulling image right?

We have two possibilities:

- update run_cmd function to raise new kind of exception CommandFailedBackOffPullingImage which will inherit from CommandFailed and here do the retry only on that specific exception.
- update the retry function to take additional parameter like message_in_excepion which will look for specific message in the exception and do retry only when `message_in_excepion in str(exception)`

Just my thoughts on this when I saw the reply from Jilju about `Retry only if error is 'Back-off pulling image `

jilju: @petr-balogh 
Yes, the current implementation will do retry for all cases of CommandFailed exception.
The second possibility you mentioned will be a great enhancement in general.

sidhant-agrawal: It would be good if we also add a retry for `TLS handshake timeout`  during oc command execution.
https://github.com/red-hat-storage/ocs-ci/issues/1604

ocs_ci/utility/utils.py
```diff
@@ -2433,3 +2433,20 @@ def configure_chrony_and_wait_for_machineconfig_status(
...
+    Args:
+        node (str): Node to create debug pod
+        cmd (str): Command to run on debug pod
+
+    Returns:
+        str: stdout of command
+
+    """
+    command = f"oc debug nodes/{node} -- {cmd}"
+    return run_cmd(command)
```
dahorak, line 2452: Function `run_cmd(...)` is [deprecated](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L436-L438) in favor of [`exec_cmd(...)`](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L460-L514), so since this is a new code, I would propose to use `exec_cmd(...)` here.
```suggestion
    return exec_cmd(command)
```

petr-balogh:
Closing due to inactivity. Feel free to re-open or send the new PR.

Issue #2604, opened 2020-07-30
Fix issue with delete_cluster_buckets


clacroix12:
From #2549 

```    
        if http.status_code >= 300:
            error_code = parsed_response.get("Error", {}).get("Code")
            error_class = self.exceptions.from_code(error_code)
>           raise error_class(parsed_response, operation_name)
E           botocore.errorfactory.NoSuchBucket: An error occurred (NoSuchBucket) when calling the ListObjects operation: The specified bucket does not exist

.venv/lib/python3.6/site-packages/botocore/client.py:635: NoSuchBucket
=============================== warnings summary ===============================
```

I don't think we should fail the upgrade if bucket not found but just continue here.

So re-opening this.

Job:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-destroy-ocs-cluster/7945/console

This is actually an issue with `delete_cluster_buckets` that was added [in this PR](https://github.com/red-hat-storage/ocs-ci/pull/2420/files)

clacroix12:
@petr-balogh The only scenario I could reproduce this issue in is when we found the bucket name initially but the bucket did not exist when we went to delete the files/bucket. In this scenario we don't want to error out and stop further teardown as the bucket is already non-existant. I have added a try/except in and we simply log any ClientErrors that we encounter when trying to delete those buckets.

petr-balogh:
Yep, I think that was exactly the case I've hit. Thanks. Can we close this issue then? @clacroix12

clacroix12:
I didn't include 'fixes' in the PR so it didn't automatically close. Closing!

PR #2611, opened 2020-07-31
Except boto errors when deleting cluster buckets

clacroix12:
#2604 


clacroix12:
I have locally verified that this gracefully handles boto ClientErrors and logs them instead of halting execution.

PR #2626, opened 2020-08-04
Couchbase:  workload running for longer time

shreekarSS:
With increasing the number of items to be loaded and the number of threads, IOs will be running for a longer time(like 7-8 minutes each pod) which was very short previously

### DIFF
--- a/ocs_ci/ocs/pillowfight.py
+++ b/ocs_ci/ocs/pillowfight.py
@@ -60,11 +60,16 @@ class PillowFight(object):
         self.up_check = OCP(namespace=constants.COUCHBASE_OPERATOR)
         self.logs = tempfile.mkdtemp(prefix='pf_logs_')
 
-    def run_pillowfights(self, replicas=1):
+    def run_pillowfights(self, replicas=1, num_items=None, num_threads=None):
         """
         loop through all the yaml files extracted from the pillowfight repo
         and run them.  Run oc logs on the results and save the logs in self.logs
         directory
+
+        Args:
+            replicas (int): Number of pod replicas
+            num_items (str): Number of items to be loaded to the cluster
+            num_threads (str): Number of threads
 
         """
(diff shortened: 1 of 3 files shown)


PrasadDesala:
@shreekarSS As discussed please change the parms so that the workers run for 15-20 minutes

shreekarSS:
> @shreekarSS As discussed please change the parms so that the workers run for 15-20 minutes

@PrasadDesala workers are running for 10 minutes now . Will take further enhancements in the followup PR

ocs_ci/ocs/pillowfight.py
```diff
@@ -80,7 +85,12 @@ def run_pillowfights(self, replicas=1):
                 # for basic-fillowfight.yaml
                 pfight = templating.load_yaml(pf_fullpath)
                 pfight['metadata']['name'] = 'pillowfight-rbd-simple' + f"{i}"
-
+                # num of items
+                pfight['spec']['template']['spec']['containers'][0]['command'][4] = str(
```
ebenahar, line 89: `num_items` is already a string

ocs_ci/ocs/pillowfight.py
```diff
@@ -80,7 +85,12 @@ def run_pillowfights(self, replicas=1):
...
+                    num_items) if num_items else '20000'
+                # num of threads
+                pfight['spec']['template']['spec']['containers'][0]['command'][13] = str(
```
ebenahar, line 92: `num_threads` is already a string

ocs_ci/ocs/pillowfight.py
shreekarSS: ack

ocs_ci/ocs/pillowfight.py
shreekarSS: ack

PR #2628, opened 2020-08-04
update flexy intergration

dahorak:
- use predefined knonw name for updated env file
- always update env file (e.g. to update cluster name etc)
- add also new (missing) keys to env file from configuration
- add possibility to skip custom NTP chrony configuration


ocs_ci/deployment/baremetal.py
```diff
@@ -36,6 +36,7 @@ def __init__(self):
 
     class OCPDeployment(BaseOCPDeployment):
         def __init__(self):
+            self.flexy_deployment = True
```
clacroix12, line 39: Should we instantiate this attribute in `BaseOCPDeployment` (OCPDeployment class) and set to `False`? By doing this we can change the `getattr` calls to simple attribute checks.

```python
if not getattr(self, 'flexy_deployment', False):
```
becomes
```python
if not self.flexy_deployment:
```

WDYT?

dahorak: Yep, I agree... it will be surely nicer.

dahorak: Changed in dfbb9da925d505e1b6720008f79330ebae166cd2.

ocs_ci/deployment/flexy.py
```diff
@@ -252,11 +243,14 @@ def merge_flexy_env(self):
...
-        # write the updated config_parser content back to the env file
-        tmp_file = os.path.join(
-            self.flexy_host_private_conf_dir_path, f"{constants.FLEXY_DEFAULT_ENV_FILE}.tmp"
-        )
-        with open(tmp_file, "w") as fp:
+        # add new keys to env file
+        for key in config.FLEXY:
+            if key not in applied_keys:
+                logger.info(f"Adding {key}={config.FLEXY[key]}")
+                config_parser.set('root', key, f"{config.FLEXY[key]}")
```
dahorak, line 250: I'll change the logic of this function - simplify it with just one loop and changing the value or adding new key in one step.

dahorak: Updated by 89a44ee.

ocs_ci/deployment/flexy.py
```diff
@@ -87,8 +87,8 @@ def deploy_prereq(self):
         if not config.ENV_DATA.get('flexy_env_file'):
             self.clone_and_unlock_ocs_private_conf()
             config.FLEXY['VARIABLES_LOCATION'] = self.template_file
-            config.FLEXY['INSTANCE_NAME_PREFIX'] = self.cluster_name
-            self.merge_flexy_env()
+        config.FLEXY['INSTANCE_NAME_PREFIX'] = self.cluster_name
+        self.merge_flexy_env()
```
shylesh, line 91: @dahorak ,
Do we want to merge_flexy_env() even if user has passed the env file ?

For ex consider:
```
1. default_flexy.conf 
FLEXY:
  GIT_PRIVATE_URI: xyz..
  VARIABLES_LOCATION: abc..

2. User supplied env file where he has already updated the above variables:
FLEXY:   
  GIT_PRIVATE_URI: pqr..

Now if we run merge_flexy_env() , 
a. Config parser would have the contents of user supplied env file (which is the value we expect to be passed to flexy run)
b. config.FLEXY would be having contents from default_flexy.conf
c. During merging we iterate over keys of config.FLEXY , in this case we get { GIT_PRIVATE_URI: xyz, VARIABLES_LOCATION: abc}  and config_parser will have  {GIT_PRIVATE_URI: pqr}.
d. We are updating config_parser(key) with value we get in config.FLEXY , which would overwrite what user wanted to pass to flexy, i.e instead of  {GIT_PRIVATE_URI: pqr} , flexy will get { GIT_PRIVATE_URI: xyz}.

am I missing something ?

dahorak: @shylesh My intention and understanding was that any supplied `ocs-ci` configuration should have precedence and user should have under control, what configuration files is providing. There are also variables, which have to be updated anyway (at least `INSTANCE_NAME_PREFIX` with supplied cluster name). The `FLEXY` section in `ocs_ci/framework/conf/default_config.yaml` is empty and I don't see any reference in the code pointing to the `ocs_ci/framework/conf/flexy_default_config.yaml`, am I missing something?

shylesh: @dahorak , Thanks for the clarification.

dahorak:
> PR validation
> 
> Cluster Name:
> Cluster Configuration:
> PR Test Suite: deployment
> PR Test Path: tests/
> Additional Test Params:
> OCP VERSION: 4.5
> OCS VERSION: 4.5
> 
> [Job PASSED](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/2609//testReport).

This validation job is only on AWS, it might be worth to run also deployment on BaremetalPSI but I'm not sure, if it is possible?

PR #2635, opened 2020-08-05
properly handle empty section in default config

dahorak:
Allow empty section in default config.

Without this change, empty section in `default_config.yaml` cause unit test `TestConfig.test_defaults` fails this way:

```
self = <ocs_ci.framework.tests.test_config.TestConfig object at 0x7f14a9c35b80>

    def test_defaults(self):
        config_sections = framework.config.to_dict().keys()
        for section_name in config_sections:
            section = getattr(framework.config, section_name)
>           assert section == framework.config.get_defaults()[section_name]
E           assert {} == None

ocs_ci/framework/tests/test_config.py:16: AssertionError
```
because empty section returns `None` instead of `{}`.

### DIFF
--- a/ocs_ci/framework/__init__.py
+++ b/ocs_ci/framework/__init__.py
@@ -44,7 +44,10 @@ class Config:
         Return a fresh copy of the default configuration
         """
         with open(DEFAULT_CONFIG_PATH) as file_stream:
-            return yaml.safe_load(file_stream)
+            return {
+                k: (v if v is not None else {})
+                for (k, v) in yaml.safe_load(file_stream).items()
+            }
 
     def update(self, user_dict: dict):
         """
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -142,11 +142,9 @@ UPGRADE:
 AUTH:
     test_quay_auth: test_secret
 
-# This section is for FLEXY, this is just dummy variable to have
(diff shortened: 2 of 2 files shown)


dahorak:
> This would convert `None` into an empty dict, but I don't understand the reasoning behind it.
> 
> Why would we want to convert `None` values from default config into empty dictionaries?

It will convert `None` only on the section level (for empty sections) -> e.g. `DEPLOYMENT`, `ENV_DATA`,... not on the configuration key itself (like `ENV_DATA['platform']`).

For example if snipped from the default config will look like this:
```yaml
ENV_DATA:
  cluster_name: null  # will be changed in ocscilib plugin
  storage_cluster_name: 'ocs-storagecluster'
FLEXY:
AUTH:
  test_quay_auth: test_secret
```
then `FLEXY` section is empty, but still it is a section, so it make sense to return empty dictionary rather than `None`. And [in the configuration object](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/__init__.py#L29) each section is defined as dict.

mbukatov:
@dahorak Thanks for the explanation.

mbukatov:
You could consider adding a unit test for behavior you are changing in this merge request, which will be useful when someone changes this code again in the future.

dahorak:
@mbukatov I think, it is covered by the `TestConfig.test_defaults` unittest (that was the reason, why I created this patch, because this test starts failing when I've added empty section).

PR #2641, opened 2020-08-05
Added --ocp-installer-version cli option for deploying OCP with a specific build

clacroix12:
Jira: [OCSQE-368](https://projects.engineering.redhat.com/browse/OCSQE-368)

### DIFF
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -132,6 +132,15 @@ def pytest_addoption(parser):
         """
     )
     parser.addoption(
+        '--ocp-installer-version',
+        dest='ocp_installer_version',
+        help="""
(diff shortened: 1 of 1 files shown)


clacroix12:
@petr-balogh I went with the additional CLI option for specific OCP versions, similar to how we treat the OCS image registry. If we would rather add the capability to use a specific version with the existing `--ocp-version` option we can definitely do it. We could attempt to read in one of our existing ocp config files and if we are unable to find one with the given version we could simply set the installer/client versions directly. The complexity would then fall to our jenkins job configurations to handle two separate fields (the existing ocp version dropdown and the new text field for specific ocp versions) and pass the correct value to the CLI. 

Given all of that, I thought adding a new CLI option which we can pair with a new text field was the most straight forward approach to this enhancement so I went with that to start.

PR #2647, opened 2020-08-06
trigger collect_ocs_logs in deploy success

jimdevops19:
Allows to collect logs as well as if the OCS deployment succeeds.
That way we will have information about the cluster as well as if the deployment succeeded.


tests/conftest.py
```diff
@@ -2704,3 +2704,20 @@ def ns_resources_and_connections_cleanup():
...
+def collect_logs_fixture(request):
+    """
+    This fixture collects ocs logs after tier execution and this will allow
+    to see the cluster's status after the execution on all execution status options.
+    """
+    def finalizer():
+        """
+        Tracking both logs separately reduce changes of collision
+        """
+        if not config.RUN['cli_params'].get('deploy'):
```
petr-balogh, line 2719: We should also not do it for destroy

clacroix12, line 2719: Won't this stop us from collecting logs if the deployment and tier tests are executed in the same command?

jimdevops19: I'm not sure if it will, this was separated as well [here](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/deployment.py#L130-L131) in the deployment.

clacroix12: If we execute one command to deploy and run tier 1 like this: `run-ci --deploy -m tier1 ...` we won't enter this `if` statement because `config.RUN['cli_params'].get('deploy')` will return `True`. Is this the expected behavior?

PR #2654, opened 2020-08-07
added multiple pvc deletion test

SvetlanaRedHat:


keesturam:
@SvetlanaRedHat This looks pretty much the same as the one for creation time, which means deletion of volumes will be done as part of the teardown. What we should be doing is introducing a way to delete the volumes in bulk as we do in the creation and then measure the time for deletion. WDYT?

Issue #2568, opened 2020-07-27
Function to change Pods Per node limit


ramkiperiy:
Scale testing requires a function to update pods per node count in the cluster, this will help in updating the existing limit of 250 pods to create more pods like 500 or 1000 per node (Note: Creation of more pod is subject to resource available in the node)

Procedure to update Pods Per node count follow link [1]

[1] https://docs.openshift.com/container-platform/4.1/nodes/nodes/nodes-nodes-managing-max-pods.html

PR #2656, opened 2020-08-07
Function to change Pods Per node limit

ramkiperiy:
Added new feature to increase pods per node limit in OCP
This pr will fix https://github.com/red-hat-storage/ocs-ci/issues/2568

Function to increase pods per node count, default OCP supports 250 pods per node,
from OCP 4.6 limit is going to be 500, but using this function can override this param
to create more pods per worker nodes.
more detail: https://docs.openshift.com/container-platform/4.5/nodes/nodes/nodes-nodes-managing-max-pods.html


ramkiperiy:
Local run output can be found in http://pastebin.test.redhat.com/891505
console output info can be found in http://pastebin.test.redhat.com/891507

ocs_ci/ocs/scale_lib.py
```diff
@@ -649,3 +651,49 @@ def check_and_add_enough_worker(worker_count):
...
+    assert ocp.exec_oc_cmd(command=label_cmd)
+
+    # First wait for Updating status to become True, default it will be False &
+    # machine_count and ready_machine_count will be equal
+    get_cmd = "get machineconfigpools -o yaml"
+    while True:
+        output = ocp.exec_oc_cmd(command=get_cmd)
+        update_status = output.get('items')[1].get('status').get('conditions')[4].get('status')
+        if update_status == 'True':
+            break
```
vasukulkarni, line 689: Can you add another count variable that can break from loop after some count ?

ramkiperiy: ack

ocs_ci/ocs/scale_lib.py
```diff
@@ -649,3 +651,49 @@ def check_and_add_enough_worker(worker_count):
...
+    more detail: https://docs.openshift.com/container-platform/4.5/nodes/nodes/nodes-nodes-managing-max-pods.html
+    WARN: This function will perform Unscheduling of workers and back to Scheduled state
+    Please aware if there is any non-dc pods then expected to be terminated.
+
+    Args:
+        count (int): Required pods per node count
+
+    """
+    max_pods_template = templating.load_yaml(constants.PODS_PER_NODE_COUNT_YAML)
+    max_pods_template['spec']['kubeletConfig']['podsPerCore'] = 65
```
keesturam, line 670: just shouldn't `podsPerCore` be dynamic based on the hardware used?

ramkiperiy: I don't think so but it depends on what kind of pod we are planning to deploy, eg: if it's nginx(without IO) then we can create 60+. Added value 65 based on manual testing where created 500 pods with 16 CPU and resources are half consumed, its a paper value to reach this limit need to use good hardware and pod.
Once any one of the limit is hit, either PodsPerCore or PodsPerNode we can't create more pods. Just want to ensure PodsPerNode count is slightly higher here, so this limit is not reached.

ramkiperiy: @keesturam @Avilir Added param for PodsPerCore in function, resolving this comment.

Issue #2649, opened 2020-08-06
Document changes to auth.yaml to be used by CloudManager


fbalak:
Function CloudManager calls [cred_dict = load_auth_config().get('AUTH')](https://github.com/red-hat-storage/ocs-ci/blob/8e153c6be5d4e1449d31f71fe4f391c561fc4b8a/ocs_ci/ocs/resources/cloud_manager.py#L34) and expects that in output of `load_auth_config` is key `'AUTH'`. There should be documented how to set `AUTH` in `auth.yaml`. In [Getting Started guide](https://github.com/red-hat-storage/ocs-ci/blob/master/docs/getting_started.md) is no information about it.

mbukatov:
Also this option should not be called just `AUTH`, but it should capture it's purpose in it's name, eg. `MCG_AUTH_FOOBAR`.

clacroix12:
> Also this option should not be called just `AUTH`, but it should capture it's purpose in it's name, eg. `MCG_AUTH_FOOBAR`.

I agree. `auth.yaml` is intended to be a generic yaml file that contains sensitive information regarding authentication. It's outlined in our [getting started guide](https://github.com/red-hat-storage/ocs-ci/blob/master/docs/getting_started.md#authentication-config) with quay.io as an example. The top level keys we are reading in from this file should be explicit and clear to their purpose / the service they correspond to. `AUTH` is far too generic for this file as everything in the file should be related to some form of authentication.

That being said, I don't think we need to document _every_ key that code in ocs-ci expects. However, what we should be doing in this case is reading in the file and seeing if the key we need exists. If it does not we should be raising an appropriate error with message describing the missing information so the user knows what their `auth.yaml` is missing.

MeridianExplorer:
All,
Currently, no doc changes were made because the change was implemented silently in MCG only. We're currently redesigning many parts, and the design isn't final. One main reason is described in issue #2623 

The main key of `AUTH` was not picked by me - it's merely the main key that resides in the `auth.yaml` file under our  `ocs-ci-data` bucket. My personal `auth.yaml` did not have this entry, I just chose to align myself to the existing format.
Also, specific auth fields are elaborated on beneath this key - 

```
AUTH:
  quay:
    access_token: 

  ipmi:
  - hostname: hostname
    console: console
    username: username
    password: password

  AWS:
    AWS_ACCESS_KEY_ID:
    AWS_SECRET_ACCESS_KEY:
  GOOGLE:
[...]

Issue #2623, opened 2020-08-03
Pick a source of truth for auth.yaml, implement it into the multiregion test


MeridianExplorer:
Right now, OCS-CI has two possible for auth.yaml - 
1. From the local code runner, under `ocs-ci/data/auth.yaml`
2. From OCS-QE's private AWS bucket, where the file is then loaded into OCS-CI's `config`.

This is confusing, hard to manage, and may result in unexpected behaviors.

MeridianExplorer:
Was addressed in #2664

PR #2664, opened 2020-08-09
Try loading MCG credentials from all possible sources

MeridianExplorer:
Right now the tests rely on loading the credentials from the local `auth.yaml`.
This PR also tries retrieving them from our private `ocs-ci-data` bucket, and if it fails - the tests that rely on those credentials will be skipped. (also added missing skip markers that were missing from a few tests)


MeridianExplorer:
Thanks for the review, @fbalak :)

The auth.yaml saved in our cloud contains the `AUTH` key while the local doesn't (according to the docs) - true. At the moment, it's unclear what should serve as a source of truth, and I think we need to take care of it as part of #2649 and #2623

Issue #2670, opened 2020-08-10
disable cloning of rook repo 


vavuthu:
we are cloning rook repo with autouse true and with scope 'session'.

Sometime cloning fails due to n/w or githib issue which causes entire deploymnet fails. 

Currently we are not using rook repo, so we can disable the cloning of repo ( keeping the code intact )

job failed due to this issue: https://ceph-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/ocs-ci/549/console

PR #2671, opened 2020-08-10
Remove cloning of rook repo since we are not using this repo

vavuthu:
Fixes: #2670 

Remove cloning of rook repo since we are not using this repo
keep the code intact incase if we need for future use


petr-balogh:
Please remove also those lines:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L25-L28

vavuthu:
> Please remove also those lines:
> https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L25-L28

Done

PR #2677, opened 2020-08-11
Shutdown and recovery node and its impact on registry

Akarsha-rai:

### DIFF
--- a/tests/e2e/registry/test_registry_shutdown_and_recovery_node.py
+++ b/tests/e2e/registry/test_registry_shutdown_and_recovery_node.py
@@ -69,7 +69,7 @@ class TestRegistryShutdownAndRecoveryNode(E2ETest):
             nodes.stop_nodes(nodes=[node])
 
             # Validate node reached NotReady state
-            wait_for_nodes_status(node_names=[node], status=constants.NODE_NOT_READY)
+            wait_for_nodes_status(node_names=[node.name], status=constants.NODE_NOT_READY)
 
             # Pull and push images to registries
             log.info("Pull and push images to registries")
--- a/tests/e2e/registry/test_registry_shutdown_and_recovery_node.py
+++ b/tests/e2e/registry/test_registry_shutdown_and_recovery_node.py
@@ -9,7 +9,7 @@ from ocs_ci.ocs.registry import (
     image_pull, image_push, image_list_all, image_rm,
     check_image_exists_in_registry
 )
-from ocs_ci.framework.testlib import E2ETest, workloads
+from ocs_ci.framework.testlib import E2ETest, workloads, ignore_leftovers
 from ocs_ci.utility.svt import svt_setup, svt_cleanup
 from ocs_ci.ocs.node import (
     wait_for_nodes_status, get_typed_nodes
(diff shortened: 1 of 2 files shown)


tests/e2e/registry/test_registry_shutdown_and_recovery_node.py
```diff
@@ -0,0 +1,143 @@
...
+
+    @pytest.mark.parametrize(
+        argnames=["iterations"],
+        argvalues=[
+            pytest.param(
+                *[5], marks=pytest.mark.polarion_id("OCS-1800")
+            )
+        ]
+    )
+    def test_registry_shutdown_and_recovery_node(self, iterations, nodes):
```
OdedViner, line 104: You don't need to use "parametrize" because you have one test case.
you can change:
def test_registry_shutdown_and_recovery_node(self, iterations=5, nodes):

Akarsha-rai: done

tests/e2e/registry/test_registry_shutdown_and_recovery_node.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    Clean up the namespace
+    """
+
+    log.info("Clean up and remove namespace")
+    ns_obj.exec_oc_cmd(command=f'delete project {project_name}')
+
+    # Reset namespace to default
+    ocp.switch_to_default_rook_cluster_project()
+    ns_obj.wait_for_delete(resource_name=project_name)
```
PrasadDesala, line 31: project_factory_fixture does these things, can we use the fixture here?

tests/e2e/registry/test_registry_shutdown_and_recovery_node.py
```diff
@@ -0,0 +1,106 @@
...
+
+            # Validate image exists in registries path
+            validate_image_exists(namespace=self.project_name)
+
+            # Remove project
+            remove_project(project_name=self.project_name)
+
+        # Validate image registry pods
+        validate_registry_pod_status()
+
```
PrasadDesala, line 104: Also validate that all the storage pods are in running state post node recovery.

Akarsha-rai: below function sanity_helpers.health_check(), validates all nodes are in Ready state, all storage pods are in running and also ceph health is OK. Do we need to add another check for only storage pods ?

tests/e2e/registry/test_registry_shutdown_and_recovery_node.py
```diff
@@ -0,0 +1,100 @@
...
+    validate_registry_pod_status,
+    image_pull_and_push,
+    validate_image_exists,
+)
+from ocs_ci.framework.testlib import E2ETest, workloads, ignore_leftovers
+from ocs_ci.ocs.node import wait_for_nodes_status, get_nodes
+from ocs_ci.utility.retry import retry
+from ocs_ci.ocs.exceptions import CommandFailed, ResourceWrongStatusException
+from ocs_ci.helpers.sanity_helpers import Sanity
+from ocs_ci.ocs.resources.pod import wait_for_storage_pods
```
vavuthu, line 15: Could you follow the ordering here

vavuthu (approved):

shreekarSS (approved):

PR #2686, opened 2020-08-11
Fix converged RGW OBC SC name

MeridianExplorer:
Up until now, the independent mode SC name was always used. This PR is supposed to fix it by using `storagecluster_independent_check` in order to return the proper SC name that's supposed to be used by OBCs

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -2372,6 +2372,14 @@ def storagecluster_independent_check():
     )
 
 
+def retrieve_rgw_storageclass_name():
+    if storagecluster_independent_check():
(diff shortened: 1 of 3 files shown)


tests/helpers.py
```diff
@@ -2372,6 +2372,14 @@ def storagecluster_independent_check():
     )
 
 
+def retrieve_rgw_storageclass_name():
```
fbalak, line 2375: I recommend to move this method to `ocs_ci/ocs/resources/rgw.py` so that it is at one place with other rgw related things. Also this method is only helpful only for rgw tests so it shouldn't be in generic helpers module.

MeridianExplorer:
Alternative PR available at #2718

PR #2691, opened 2020-08-12
test_monitoring_shutdown_mgr_pod

OdedViner:
OCS-1535


ocs_ci/ocs/monitoring.py
```diff
@@ -18,6 +18,252 @@
...
+    "ceph_osd_op_w",
+    "ceph_paxos_commit_bytes_sum",
+    "ceph_bluefs_log_bytes",
+    "ceph_rocksdb_submit_sync_latency_count",
+    "ceph_pool_num_bytes_recovered",
+    "ceph_pool_num_objects_recovered",
+    "ceph_pool_recovering_bytes_per_sec",
+    "ceph_pool_recovering_keys_per_sec",
+    "ceph_pool_recovering_objects_per_sec"
+]
```
mbukatov, line 265: This duplicates code in https://github.com/red-hat-storage/ocs-ci/blob/19605e18c5a5cdc77ca454b01ea26343cf1ab7a4/tests/manage/monitoring/prometheusmetrics/test_monitoring_defaults.py#L128

ocs_ci/ocs/monitoring.py
```diff
@@ -227,3 +473,30 @@ def prometheus_health_check(name=constants.MONITORING, kind=constants.CLUSTER_OP
 
     logging.error(f"Prometheus cluster is degraded {health_conditions}")
     return False
+
+
+def check_ceph_metrics_available():
```
mbukatov, line 478: This function is reimplementing OCS-1268, which is already covered in https://github.com/red-hat-storage/ocs-ci/blob/19605e18c5a5cdc77ca454b01ea26343cf1ab7a4/tests/manage/monitoring/prometheusmetrics/test_monitoring_defaults.py#L128

But it:
- doesn't take into account that on a cloud platforms, not all metrics are available (because RGW is not deployed there) or it doesn't cover all metrics of the platform (I haven't checked which what is the case)
-  doesn't properly log all metrics it asks propetheus instance for (this could be very useful, when inspecting what was actually exactly tested)

mbukatov (changes requested):
I see few cases of code duplication/reimplementation. I suggest to tackle this via:
- I'm going to move core part of https://github.com/red-hat-storage/ocs-ci/blob/19605e18c5a5cdc77ca454b01ea26343cf1ab7a4/tests/manage/monitoring/prometheusmetrics/test_monitoring_defaults.py#L128 into ocs-ci module for others to use https://github.com/red-hat-storage/ocs-ci/issues/2773
- you can then rebase your pull request and use the functionality from there

This way, we other test cases will be able to reuse the code which is part of tier1 for some time, and it's logging for other purposes.

ocs_ci/ocs/monitoring.py
mbukatov: Witch changes from https://github.com/red-hat-storage/ocs-ci/pull/2798, you can use list of metrics and basic check function from metrics module to use standard logging and avoid duplication.

PR #2693, opened 2020-08-12
Adding retries to the ripsaw.get_uuid function

Avilir:
This is to fix the first issue in Jenkins build #10799

### DIFF
--- a/ocs_ci/ocs/ripsaw.py
+++ b/ocs_ci/ocs/ripsaw.py
@@ -4,6 +4,7 @@ RipSaw Class to run various workloads and scale tests
 """
 import logging
 import tempfile
+import time
 
 from ocs_ci.ocs.exceptions import CommandFailed
 from ocs_ci.ocs.ocp import OCP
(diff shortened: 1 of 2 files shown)


Avilir:
second commit fix the second issue in the same Job which found leftover even there is none.

ocs_ci/ocs/ripsaw.py
```diff
@@ -101,13 +102,24 @@ def get_uuid(self, benchmark):
             str: the UUID of the test
 
         """
-        output = self.pod_obj.exec_oc_cmd(f'exec {benchmark} -- env')
-        uuid = ''
-        for line in output.split():
-            if 'uuid=' in line:
-                uuid = line.split('=')[1]
+        count = 0
+        while count <= 5:
```
AviadP, line 106: why using while loop instead of for loop?

Avilir: no particular reason

PR #2694, opened 2020-08-12
Add ocs-monkey test

jilju:
Added a test to run ocs-monkey, a chaos testing suite for OCS.
https://github.com/red-hat-storage/ocs-monkey

### DIFF
--- a/tests/ocs-monkey/test_ocs_monkey.py
+++ b/tests/ocs-monkey/test_ocs_monkey.py
@@ -1,0 +1,61 @@
+import logging
+import shlex
+import subprocess
+import os
+import time
+
+from ocs_ci.framework.testlib import ignore_leftovers
+from ocs_ci.utility.utils import clone_repo, run_cmd
+from ocs_ci.ocs import constants
+
+log = logging.getLogger(__name__)
+
+
+@ignore_leftovers
+def test_ocs_monkey():
+    ocs_monkety_dir = '/tmp/ocs-monkey'
+    # ocs-monkey run time in seconds
+    run_time = 3200
+    clone_repo(constants.OCS_MONKEY_REPOSITORY, ocs_monkety_dir)
+    run_cmd(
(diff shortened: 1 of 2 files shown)


tests/ocs-monkey/test_ocs_monkey.py
```diff
@@ -0,0 +1,61 @@
...
+import time
+
+from ocs_ci.framework.testlib import ignore_leftovers
+from ocs_ci.utility.utils import clone_repo, run_cmd
+from ocs_ci.ocs import constants
+
+log = logging.getLogger(__name__)
+
+
+@ignore_leftovers
```
ebenahar, line 14: @hnallurv @PrasadDesala should this test be part of `workloads` tier?

RazTamir: IMO it should be in a new tier level.
As this is not a real workload we can try to integrate it with the other workloads that are running for a long time.
I'm thinking mostly about the BDI workload that can run for hours and with the OCS monkey in place, we are really taking the test to the next level

ebenahar: yes. I have discussed this with @jilju and he brought up the idea of having a dedicated Jenkins job specifically for ocs-monkey test

ebenahar (approved):

AviadP (approved):

tests/ocs-monkey/test_ocs_monkey.py
```diff
@@ -0,0 +1,61 @@
...
+def test_ocs_monkey():
+    ocs_monkety_dir = '/tmp/ocs-monkey'
+    # ocs-monkey run time in seconds
+    run_time = 3200
+    clone_repo(constants.OCS_MONKEY_REPOSITORY, ocs_monkety_dir)
```
mykaul, line 19: Why not get OCS monkey as a git submodule of OCS-CI?

PR #2700, opened 2020-08-12
Add Permission exception for update_config_from_s3

vavuthu:
Fixes: #2698 

Signed-off-by: vavuthu <vavuthu@redhat.com>

PR #2702, opened 2020-08-12
skipif_aws_creds_are_missing: Check os.environ

zmc:
PR #2664 broke our build validation because we use environment variables to
provide AWS credentials.

### DIFF
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -98,6 +98,7 @@ run_this = pytest.mark.run_this
 skipif_aws_creds_are_missing = pytest.mark.skipif(
     (
         load_auth_config().get('AUTH', {}).get('AWS', {}).get('AWS_ACCESS_KEY_ID') is None
+        and 'AWS_ACCESS_KEY_ID' not in os.environ
         and update_config_from_s3() is None
     ),
     reason=(


ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -98,6 +98,7 @@
 skipif_aws_creds_are_missing = pytest.mark.skipif(
     (
         load_auth_config().get('AUTH', {}).get('AWS', {}).get('AWS_ACCESS_KEY_ID') is None
+        and 'AWS_ACCESS_KEY_ID' not in os.environ
```
vasukulkarni, line 101: If you dont mind waiting for a day, we can fix it properly? but I will approve if its blocking

vasukulkarni: @MeridianExplorer if we can use exception and skip that will be nice without additional environment variables.

zmc: What's not proper about this?

vasukulkarni: I was thinking if we can avoid environment variable since its valid for many other aws environment, we just hit this today in other dev environment and I filed an issue too.

zmc:
Thanks for the reviews guys!

ocs_ci/framework/pytest_customization/marks.py
vavuthu: This issue is handled in PR https://github.com/red-hat-storage/ocs-ci/pull/2700

PR #2703, opened 2020-08-12
utils: fix unhandled exception type when using bad auth.yaml contents

phlogistonjohn:
I hit this case when I had an empty auth.yaml file. This fix will cause either a missing or empty file to print out the reminder to consult the docs.


mykaul:
@ebenahar - can anyone review this and get it merged if it's OK?

PR #2710, opened 2020-08-13
Skip PVC-Create-Deletion TC due to "Cannot allocate memory" Problem

ramkiperiy:
Added pytest marker to skip test case


tests/e2e/scale/test_pvc_creation_deletion_scale.py
```diff
@@ -16,6 +16,7 @@
 
 
 @scale
+@pytest.mark.skip("Skip this test due to Cannot allocate memory problem")
```
petr-balogh, line 19:
```suggestion
@pytest.mark.skip(reason="Skip this test due to Cannot allocate memory problem")
```

PR #2719, opened 2020-08-13
ocscilib: Log tracebacks instead of just exceptions

zmc:
We were seeing must-gather failures like this:
```
14:33:26 - MainThread - ocs_ci.framework.pytest_customization.ocscilib - ERROR - Failed to collect OCS logs. Error: list index out of range
```

Just not enough information to figure out what was going on. I went through the module and found all the instances of `log.error()` combined with an attempt to display something helpful about the exception, and modified them to use `log.exception()`, which gives you the entire traceback.

### DIFF
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -423,8 +423,8 @@ def pytest_runtest_makereport(item, call):
         mcg = True if any(x in item.location[0] for x in ['mcg', 'ecosystem']) else False
         try:
             collect_ocs_logs(dir_name=test_case_name, ocp=ocp_logs_collection, mcg=mcg)
-        except Exception as ex:
-            log.error(f"Failed to collect OCS logs. Error: {ex}")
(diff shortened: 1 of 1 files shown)


ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -266,17 +266,14 @@ def gather_version_info_for_report(config):
...
-    except Exception as ex:
-        log.error("Failed to gather version info! Error: %s", ex)
+    except ResourceNotFoundError:
+        log.exception("Problem occurred when looking for some resource!")
+    except FileNotFoundError:
+        log.exception("File not found!")
+    except CommandFailed:
+        log.exception("Failed to execute command!")
+    except Exception:
+        log("Failed to gather version info!")
```
clacroix12, line 276:
```suggestion
        log.exception("Failed to gather version info!")
```

zmc: Yikes, thanks for catching that. I've just repushed.

PR #2730, opened 2020-08-17
Failover RGW pods tests

ebenahar:


tests/manage/mcg/test_mcg_resources_disruptions.py
```diff
@@ -0,0 +1,93 @@
...
+                *['noobaa_core'], marks=pytest.mark.polarion_id("OCS-2232")
+            ),
+            pytest.param(
+                *['noobaa_db'], marks=pytest.mark.polarion_id("OCS-2233")
+            )
+        ]
+    )
+    def test_restart_noobaa_resources(self, resource_to_delete):
+        """
+        Test Noobaa resources restart and check Noobaa health
```
fbalak, line 43: NooBaa resources are not restarted but deleted. After they are deleted the test waits for a new one to be created but it is not the original one so it is not restarted in my opinion.

ebenahar: done

ebenahar:
> Looks good but it would be good to add docstring comments for test steps to improve readability.

Added

vavuthu:
LGTM.

Since we scale down RGW pods to 1 and 0, are we bringing back RGW pods to original state. If not will this test case doesn't effect the other test cases?

ebenahar:
> LGTM.
> 
> Since we scale down RGW pods to 1 and 0, are we bringing back RGW pods to original state. If not will this test case doesn't effect the other test cases?

yes, here https://github.com/red-hat-storage/ocs-ci/pull/2730/files#diff-4d34189c6531b72ebec1e27d784bed8dR95-R97

vavuthu:
> > LGTM.
> > Since we scale down RGW pods to 1 and 0, are we bringing back RGW pods to original state. If not will this test case doesn't effect the other test cases?
> 
> yes, here https://github.com/red-hat-storage/ocs-ci/pull/2730/files#diff-4d34189c6531b72ebec1e27d784bed8dR95-R97

Thanks. Missed that

PR #2739, opened 2020-08-17
Fetch the Noobaa min and max endpoints from the run config instead of the Noobaa CR

ebenahar:


ebenahar:
Will be addressed in https://github.com/red-hat-storage/ocs-ci/pull/2740. Closing

Issue #2704, opened 2020-08-12
Handling existing cluster outside qe


vasukulkarni:
We have seen couple of issues that needs to be addressed when existing cluster is used from outside qe 

1) Download of oc binary in bin/folder if not already present
2) Dont bailout on report version checks in operators and catalog namespace
   ex:

"""
-OPERATOR_CATALOG_SOURCE_NAME = "ocs-catalogsource"
+OPERATOR_CATALOG_SOURCE_NAME = "redhat-operators"
$ oc -n openshift-marketplace --kubeconfig /Users/kylebader/src/infra-nodes/kb/auth/kubeconfig get CatalogSource ocs-catalogsource -n openshift-marketplace -o yaml
Error from server (NotFound): catalogsources.operators.coreos.com "ocs-catalogsource" not found
"""

3) cluster-name should not be mandatory

4) mcg tests loads s3 config in pytest customization - add exception checks

vasukulkarni:
5.
```
Kyles-Mac-mini:ocs-ci kylebader$ run-ci -m "test" --cluster-name kb --cluster-path ~/src/infra-nodes/kb tests/libtest/test_pod_exec.py
Collecting Cluster versions
Problem occurred when looking for some resource! Error: Requested packageManifest: ocs-operator with selector: ocs-operator-internal=true not found!
Failed to gather version details! The report of version mightnot be complete!
======================================================= test session starts ========================================================
platform darwin -- Python 3.7.7, pytest-5.3.5, py-1.9.0, pluggy-0.13.1
rootdir: /Users/kylebader/src/ocs-ci, inifile: pytest.ini
plugins: metadata-1.10.0, html-2.1.1, ordering-0.6, logger-0.5.1, reportportal-1.10.0, marker-bugzilla-0.9.4
collected 1 item / 1 deselected
```

clacroix12:
@vasukulkarni I'm wondering if #2562 is what is being hit here.

vasukulkarni:
> > 1. cluster-name should not be mandatory
> 
> @vasukulkarni I'm wondering if #2562 is what is being hit here.

That is the one!!

clacroix12:
[...]

PR #2755, opened 2020-08-19
Fix hard failure when unable to get csv version for reporting

vasukulkarni:
Partial Fixes (5): https://github.com/red-hat-storage/ocs-ci/issues/2704

We dont have redhat-operator-internal selector when installed via GUI so the get_csv call fails, we can safely ignore this error.

### DIFF
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -223,12 +223,15 @@ def pytest_configure(config):
         config._metadata['Test Run Name'] = get_testrun_name()
         gather_version_info_for_report(config)
 
-        ocs_csv = get_ocs_csv()
-        ocs_csv_version = ocs_csv.data['spec']['version']
-        config.addinivalue_line(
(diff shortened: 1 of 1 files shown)


vavuthu (approved):

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -223,12 +223,15 @@ def pytest_configure(config):
...
-
+        try:
+            ocs_csv = get_ocs_csv()
+            ocs_csv_version = ocs_csv.data['spec']['version']
+            config.addinivalue_line(
+                "rp_launch_tags", f"ocs_csv_version:{ocs_csv_version}"
+            )
+        except ResourceNotFoundError:
+            # might be using exisitng cluster path using GUI installation
+            log.info("Unable to get CSV version for Reporting")
```
clacroix12, line 234: This should probably be a warning level since it's not something we want to go unnoticed.

vasukulkarni: I will fix that along with the linter issues

vavuthu (approved):

PR #2756, opened 2020-08-19
added mutiple pvc deletion case

SvetlanaRedHat:

### DIFF
--- a/tests/e2e/performance/test_pvc_deletion_performance.py
+++ b/tests/e2e/performance/test_pvc_deletion_performance.py
@@ -4,11 +4,14 @@ Test to verify PVC deletion performance
 import logging
 import pytest
 import ocs_ci.ocs.exceptions as ex
+import threading
 from ocs_ci.framework.testlib import (
     performance, E2ETest
 )
+
+from concurrent.futures import ThreadPoolExecutor
 from tests import helpers
-from ocs_ci.ocs import constants
+from ocs_ci.ocs import defaults, constants
 from ocs_ci.utility.performance_dashboard import push_to_pvc_time_dashboard
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


Avilir (approved):
LGTM, nice work @SvetlanaRedHat

tests/e2e/performance/test_pvc_deletion_performance.py
```diff
@@ -86,3 +89,70 @@ def test_pvc_deletion_measurement_performance(self, teardown_factory, pvc_size):
...
+        for obj in pvc_objs:
+            obj.delete()
+            obj.ocp.wait_for_delete(obj.name)
+
+        # Get PVC deletion time
+        pvc_deletion_time = helpers.measure_pv_deletion_time_bulk(
+            interface=self.interface, pv_name_list=pv_name_list
+        )
+        logging.info(
+            f"{number_of_pvcs} PVCs deletion time took {pvc_deletion_time}"
```
ramkiperiy, line 157: @SvetlanaRedHat are we capturing the data in some file ? as per code it will just display in console.

SvetlanaRedHat: I am planning to push this to code speed in my next PR

tests/e2e/performance/test_pvc_deletion_performance.py
```diff
@@ -86,3 +89,70 @@ def test_pvc_deletion_measurement_performance(self, teardown_factory, pvc_size):
...
+            process2.start()
+            threads.append(process1)
+            threads.append(process2)
+        for process in threads:
+            process.join()
+        log.info("Preparing to delete 120 PVC")
+        # Delete PVC
+        for obj in pvc_objs:
+            obj.delete()
+            obj.ocp.wait_for_delete(obj.name)
```
keesturam, line 150: I think we should delete all the objects in a loop and then check/wait for deletion in a separate loop. By combining these two, we'll be deleting 120 PVs in serial and not as a bulk operation.

keesturam (dismissed):

tests/e2e/performance/test_pvc_deletion_performance.py
SvetlanaRedHat: this is fixed

tiffanyn108 (approved):
LGTM.

vasukulkarni (approved):

Issue #2758, opened 2020-08-19
Change fixture scope to class instead of session, also break the scale pod creation numbers.


ramkiperiy:
TC: https://github.com/red-hat-storage/ocs-ci/blob/master/tests/e2e/scale/test_ceph_pod_respin_in_scaled_cluster.py#L16
Fixture is created at session level, during batch run it affects the testcase which is executed after this.

Modify the code to use the fixture at class level instead of session.

Break the scale pod creation numbers, instead of creating 1500 single time break it into 350 + 350 + 400 + 400 (This combination is already tested locally)

ramkiperiy:
PR merged

PR #2759, opened 2020-08-19
Fix-2758

ramkiperiy:
Fixed issues 2758, changed fixture to be at class level.
Commented ceph and resource checks which is increasing the scale execution time

### DIFF
--- a/tests/e2e/scale/test_ceph_pod_respin_in_scaled_cluster.py
+++ b/tests/e2e/scale/test_ceph_pod_respin_in_scaled_cluster.py
@@ -13,7 +13,7 @@ from ocs_ci.framework.pytest_customization.marks import (
 log = logging.getLogger(__name__)
 
 
-@pytest.fixture(scope='session')
+@pytest.fixture(scope='class')
 def fioscale(request):
     """
     FIO Scale fixture to create expected number of POD+PVC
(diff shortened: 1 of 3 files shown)


ramkiperiy:
Fixes issue https://github.com/red-hat-storage/ocs-ci/issues/2758

ramkiperiy:
Testing it locally, can't execute this tests in jenkins due to memory allocation problem.

tests/e2e/scale/test_ceph_pod_respin_in_scaled_cluster.py
```diff
@@ -25,9 +25,19 @@ def fioscale(request):
         node_selector=constants.SCALE_NODE_SELECTOR
     )
     fioscale.create_scale_pods(
-        scale_count=1500, pods_per_iter=10, io_runtime=36000,
+        scale_count=350, pods_per_iter=10, io_runtime=36000,
```
keesturam, line 28: @ramkiperiy We used to run IO on 1500 pods for 10 hours earlier and now we run,
IOs on 350 pods for 10 hours, followed by the creation of 350 pods without IOs and then 400 pods with ~1 hour and 400 pods without IO. 

We'll simply use 400 pods for each iteration and run IOs for 18000 seconds alternatively. IMO, It is okay to cross 1500 pods. (we anyway don't run IOs on all of them)

Also, is there any reason we run IO for such a long time?

keesturam (changes requested):

ramkiperiy:
Planning to implement kube jobs to scale pvc+pod, so it then holding on merging this PR.
Will create new PR for changing the fixture scope to "Class"

ramkiperiy:
Closing this PR since entire script needs re-work to use kube_job or other python method which doesn't use more Object creations.

PR #2779, opened 2020-08-21
Add conftest.py hook for vSphere (#2779)

wusui:
The test_scale_pvc_expand and test_ceph_pod_respin_in_scaled_cluster
tests should not be run on vms.

### DIFF
--- a/tests/e2e/scale/conftest.py
+++ b/tests/e2e/scale/conftest.py
@@ -15,7 +15,8 @@ def pytest_collection_modifyitems(items):
     """
     skip_list = [
         'test_scale_pvc_expand',
-        'test_ceph_pod_respin_in_scaled_cluster'
+        'test_ceph_pod_respin_in_scaled_cluster',
+        'test_scale_osds_fill_75%_reboot_workers'
     ]
     if config.ENV_DATA['platform'].lower() == constants.VSPHERE_PLATFORM:
         for item in items.copy():
--- a/tests/e2e/scale/conftest.py
+++ b/tests/e2e/scale/conftest.py
@@ -1,0 +1,29 @@
+import logging
+from ocs_ci.framework import config
+from ocs_ci.ocs import constants
+
+log = logging.getLogger(__name__)
+
+
+def pytest_collection_modifyitems(items):
+    """
+    A pytest hook to skip certain tests when running on vSphere.
+
+    Args:
+        items: list of collected tests
+
+    """
+    skip_list = [
+        'test_scale_pvc_expand',
+        'test_ceph_pod_respin_in_scaled_cluster'
+    ]
+    if config.ENV_DATA['platform'].lower() == constants.VSPHERE_PLATFORM:
+        for item in items.copy():
+            for testname in skip_list:
+                if testname in str(item.fspath):
+                    log.info(
+                        f"Test {item} is removed from the collected items"
+                        f" since it does not run on vSphere"
+                    )
+                    items.remove(item)
+                    break


tests/e2e/scale/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Args:
+        items: list of collected tests
+
+    """
+    skip_list = [
+        'test_scale_pvc_expand',
+        'test_ceph_pod_respin_in_scaled_cluster'
+    ]
+    if config.ENV_DATA['platform'].lower() == "vsphere":
```
petr-balogh, line 19: you can use VSPHERE_PLATFORM  from https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/constants.py#L686

tests/e2e/scale/conftest.py
```diff
@@ -0,0 +1,28 @@
+import logging
+from ocs_ci.framework import config
+
+log = logging.getLogger(__name__)
+
+
+def pytest_collection_modifyitems(items):
+    """
+    A pytest hook to skip certain tests when running on vms.
```
petr-balogh, line 9: what the vms is stand for here?   Virtual Machines?   If so it should be VMs or whole Virtual Machines.

tests/e2e/scale/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+        'test_scale_pvc_expand',
+        'test_ceph_pod_respin_in_scaled_cluster'
+    ]
+    if config.ENV_DATA['platform'].lower() == "vsphere":
+        for item in items.copy():
+            for testname in skip_list:
+                if testname in str(item.fspath):
+                    log.info(
+                        f"Test {item} is removed from the collected items"
+                        f" since it does not run on vms"
```
petr-balogh, line 25: Same question with vms here.

tests/e2e/scale/conftest.py
wusui: Changed to use constants.VSPHERE_PLATFORM

tests/e2e/scale/conftest.py
wusui: changed to "virtual machine"

tests/e2e/scale/conftest.py
wusui: changed to virtual machine

ramkiperiy:
@wusui have you tested this ?

wusui:
I tested this before I checked in the PR.  I will run the PR now too.

wusui:
This is failing with -- Trying to acquire lock on [Label: vSphere-e2e,
Quantity: 1]


It appears that there have been some changes related to this lately,
so I am going to rebase onto the latest and reinstall.


I will also see how this affects the local run-ci issue that I have been seeing.

ramkiperiy (dismissed):

Avilir (dismissed):

vasukulkarni (approved):

tiffanyn108 (dismissed):

tiffanyn108:
Warren, can you add "test_scale_osds_fill_75%_reboot_workers" to the list since this testcase also running adding capacity.

ramkiperiy:
@wusui How are you triggering this PR validation meaning would like to know the param's you are using. Looks like last PR validation run doesn't skip the tests. I suspect it might be due to the platform param (i.e. default_config.yaml) which is default to AWS, I don't know where to get this input from logs. But the expected tests not skipped.

ramkiperiy (approved):

wusui:
See Ram's comment.  The change works as expected but the failure is caused by other scale test failures.  That second failure was an invalid run on my part because I thought I could get this to pass by running one of the skipped files but the change does not affect things when an individual run is sepcified.

vasukulkarni (approved):

PR #2787, opened 2020-08-23
removed workaround for OSD timeout after upgrade

AviadP:
with this BZ https://bugzilla.redhat.com/show_bug.cgi?id=1840729 moved to verify, im removing the workaround we had in the code, time out set to 10 minutes again

### DIFF
--- a/ocs_ci/ocs/ocs_upgrade.py
+++ b/ocs_ci/ocs/ocs_upgrade.py
@@ -111,12 +111,9 @@ def verify_image_versions(old_images, upgrade_version, version_before_upgrade):
     )
     verify_pods_upgraded(old_images, selector=constants.MON_APP_LABEL, count=3)
     verify_pods_upgraded(old_images, selector=constants.MGR_APP_LABEL)
-    # OSD upgrade have timeout 10mins for new attempt if cluster is not health.
-    # https://bugzilla.redhat.com/show_bug.cgi?id=1840729 setting timeout for
-    # 12.5 minutes per OSD
     verify_pods_upgraded(
         old_images, selector=constants.OSD_APP_LABEL, count=osd_count,
-        timeout=750 * osd_count,
+        timeout=600 * osd_count,
     )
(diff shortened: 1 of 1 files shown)


ebenahar (approved):

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -110,12 +110,9 @@ def verify_image_versions(old_images, upgrade_version):
     )
     verify_pods_upgraded(old_images, selector=constants.MON_APP_LABEL, count=3)
     verify_pods_upgraded(old_images, selector=constants.MGR_APP_LABEL)
-    # OSD upgrade have timeout 10mins for new attempt if cluster is not health.
-    # https://bugzilla.redhat.com/show_bug.cgi?id=1840729 setting timeout for
-    # 12.5 minutes per OSD
     verify_pods_upgraded(
         old_images, selector=constants.OSD_APP_LABEL, count=osd_count,
-        timeout=750 * osd_count,
+        timeout=600 * osd_count,
```
petr-balogh, line 115: what about upgrade from 4.3 to 4.4.  If we need to still run that, then we can maybe have an issue and maybe we should change this value only when version >= 4.5

petr-balogh (commented):
This has to be verified with all pre/post upgrade tests together @AviadP + enable parameter RUN IO IN BACKGROUND.  Please do not merge till it's verified like this please.

ocs_ci/ocs/ocs_upgrade.py
AviadP: i can add it as constant and use it according to upgraded version

petr-balogh: Please do @AviadP

petr-balogh (approved):

wusui (approved):
LGTM

ebenahar (approved):

PR #2789, opened 2020-08-23
test monitoring OCS-636

OdedViner:

### DIFF
--- a/tests/e2e/monitoring/test_monitoring_on_negative_scenarios.py
+++ b/tests/e2e/monitoring/test_monitoring_on_negative_scenarios.py
@@ -150,6 +150,37 @@ class TestMonitoringBackedByOCS(E2ETest):
             )
         return pod_objs
 
+    @pytest.fixture()
+    def delete_configmap(self):
+        """
+        Delete config map and monitoring pvc
+
+        """
+        # Get 'cluster-monitoring-config' configmap
+        ocp_configmap = ocp.OCP(namespace=constants.MONITORING_NAMESPACE, kind='configmap')
+        configmap_dict = ocp_configmap.get(resource_name='cluster-monitoring-config')
+        dir_configmap = tempfile.mkdtemp(prefix='configmap_')
+        yaml_file = f'{dir_configmap}/configmap.yaml'
+        templating.dump_data_to_temp_yaml(configmap_dict, yaml_file)
+
+        # Get all pvc on monitoring namespace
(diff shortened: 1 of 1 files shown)


tests/e2e/monitoring/test_monitoring_on_negative_scenarios.py
```diff
@@ -150,6 +150,37 @@ def pods(self, multi_pvc_factory, dc_pod_factory):
             )
         return pod_objs
 
+    @pytest.fixture()
+    def delete_configmap(self):
```
Akarsha-rai, line 154: Can we rename the function delete_and_recreate_configmap() something like that , since you are recreating configmap in this function wdyt?

And also can we move this function under monitoring.py file?

mbukatov: I agree:
- renaming would make the purpose of the function immediately more clear
- moving this function to [monitoring.py](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/monitoring.py) could make sense (this module contains code handling monitoring reconfiguration for OCS)

But the most important detail here is the reason why we are dong this before the test. What will this achieve? What would a failure of this procedure mean? It would be useful to noted this down in docstring or a comment in the test code.

tests/e2e/monitoring/test_monitoring_on_negative_scenarios.py
```diff
@@ -552,3 +583,33 @@ def test_monitoring_delete_pvc(self):
...
+
+        log.info('Waiting for 20 seconds')
+        time.sleep(20)
+
+        # Powering on node
+        nodes.start_nodes(nodes=[prometheus_node_obj])
+
+        # check pvcdata collected on prometheus
+        for pod_obj in pods:
+            assert check_pvcdata_collected_on_prometheus(pod_obj.pvc.name), (
```
mbukatov, line 613: I would rather iterated over all pods using log.error() instead of assert, and then after the for loop, used assert to check if at least one pod failed the check. This way, we will have more evidence in the logs (how many pods were checked? how many of these failed? ...) which will be helpful during evaluation of the results.

mbukatov (commented):
I have 2 suggestions/questions, but otherwise it looks good.

petr-balogh:
Hey Oded, I see no activity here. Are you going to finish with this PR? Thanks

PR #2804, opened 2020-08-25
Skip mcg operations when deleting the cluster

vavuthu:


tests/conftest.py
```diff
@@ -1590,9 +1590,9 @@ def mcgcli_pod_fixture(request):
     and upload the binary to it.
 
     """
-    deploy = config.RUN['cli_params']['deploy']
+    teardown = config.RUN['cli_params'].get('teardown')
     skip_ocs = config.ENV_DATA['skip_ocs_deployment']
-    if deploy and skip_ocs:
+    if skip_ocs or teardown:
         log.info('Skipping mcg operations for OCP only deployments.')
```
clacroix12, line 1596: Should this log message be updated as well?

vavuthu: Done

Issue #2803, opened 2020-08-25
Add polarion id to test_upgrade


fbalak:
tests/ecosystem/upgrade/test_upgrade.py::test_upgrade doesn't have Polarion ID with valid test steps.

clacroix12:
These can be handled the same way we handle deployment polarion_ids: by adding a corresponding polarion_id to the config files ([example](https://github.com/red-hat-storage/ocs-ci/blob/master/conf/examples/polarion_id.yaml)) and have the value added via our [reporting utility](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/reporting.py#L14-L18).

petr-balogh:
@clacroix12 I thinks we didn't want to merge the PR linked here and close this one, right? If so, please reopen?

clacroix12:
@petr-balogh yeah by the time I was aware of the test case changes we had merged it. It doesn't hurt anything really being merged though. I'm reopening this issue and going to raise another PR that will address the config file changes we discussed.

PR #2805, opened 2020-08-25
Upgrade polarion ids

clacroix12:
Fixes: #2803

### DIFF
--- a/ocs_ci/utility/reporting.py
+++ b/ocs_ci/utility/reporting.py
@@ -1,9 +1,16 @@
+import logging
+
 from ocs_ci.framework import config
 
+log = logging.getLogger(__name__)
 
-def get_deployment_polarion_id():
+
+def get_polarion_id(upgrade=False):
     """
     Determine the polarion_id of the deployment or upgrade
+
+    Args:
+        upgrade (bool): get upgrade_id if true, else get deployment_id
 
     Returns:
         str: polarion_id of the deployment or upgrade
@@ -11,7 +18,11 @@ def get_deployment_polarion_id():
     """
     polarion_config = config.REPORTING.get('polarion')
     if polarion_config:
-        if config.UPGRADE.get('upgrade'):
-            return polarion_config.get('upgrade_id')
+        if upgrade:
+            upgrade_id = polarion_config.get('upgrade_id')
+            log.info('polarion upgrade_id: %s', upgrade_id)
+            return upgrade_id
         else:
-            return polarion_config.get('deployment_id')
+            deployment_id = polarion_config.get('deployment_id')
+            log.info('polarion deployment_id: %s', deployment_id)
+            return deployment_id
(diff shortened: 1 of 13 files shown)


fbalak (approved):

clacroix12:
Currently the upgrade_id isn't being utilized because `config.UPGRADE.get('upgrade'):` is evaluating to `False`. This is because we aren't passing a configuration when we execute the upgrade that sets that config value to `True`.

@petr-balogh is there a reason we aren't passing a config file that sets `config.UPGRADE['upgrade']` to `True` when we run the upgrade stage? I'm assuming we just don't have anything that is hooking onto that config value yet so we haven't needed to pass that config. I'm going to raise an MR that adds that config to the upgrade stage so we know from the config that we are executing an upgrade. If we aren't adding that value for a particular reason I will have to change how we determine we are running an upgrade.

clacroix12:
Verified from the upgrade xml that the polarion ID is being added as a custom property: http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/jnk-pr2805-b2757/jnk-pr2805-b2757_20200831T225536/logs/upgrade_test_results_1598914396.xml

ocs_ci/utility/reporting.py
```diff
@@ -1,20 +1,28 @@
...
     polarion_config = config.REPORTING.get('polarion')
     if polarion_config:
-        if config.UPGRADE.get('upgrade'):
-            if config.DEPLOYMENT.get('subscription_plan_approval') == 'Manual':
-                return polarion_config.get('upgrade_manual_id')
-            else:
-                return polarion_config.get('upgrade_auto_id')
+        if upgrade:
+            upgrade_id = polarion_config.get('upgrade_id')
+            log.info('polarion upgrade_id: %s', upgrade_id)
```
AviadP, line 23:
```suggestion
            log.info(f'polarion upgrade_id: {upgrade_id}')
```

ocs_ci/utility/reporting.py
```diff
@@ -1,20 +1,28 @@
...
+            return upgrade_id
         else:
-            return polarion_config.get('deployment_id')
+            deployment_id = polarion_config.get('deployment_id')
+            log.info('polarion deployment_id: %s', deployment_id)
```
AviadP, line 27:
```suggestion
            log.info(f'polarion deployment_id: {deployment_id}')
```

ocs_ci/utility/reporting.py
fbalak: Those log messages are in line with our code guidelines https://github.com/red-hat-storage/ocs-ci/blob/a94361d4f30b3a1901ce376f0e85048c26707e78/docs/coding_guidelines.md for Logging.

clacroix12: Correct, when it comes to logging we have decided to let the logger handle the message formatting.

vasukulkarni (approved):

mbukatov (approved):

petr-balogh (approved):

ebenahar (approved):

petr-balogh:
@clacroix12 based on our latest conversation I guess this we didn't want to be merged right?

PR #2806, opened 2020-08-26
AWS cleanup: disable dnd prefix and lr of more than 2 days

ebenahar:

### DIFF
--- a/ocs_ci/cleanup/aws/cleanup.py
+++ b/ocs_ci/cleanup/aws/cleanup.py
@@ -147,33 +147,22 @@ def get_clusters(time_to_delete, region_name, prefixes_hours_to_spare):
     def determine_cluster_deletion(ec2_instances, cluster_name):
         for instance in ec2_instances:
             allowed_running_time = time_to_delete
-            do_not_delete = False
             if instance.state["Name"] == "running":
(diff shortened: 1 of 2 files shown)


AviadP (approved):

dahorak:
I don't agree with this change, because for now, we are not able to directly install Disconnected or Proxy environment cluster anywhere else than in AWS. We have to surely find the correct way, how to install those configurations in vSphere and OpenStack, but it is not ready out of the box right now.
And for example installation of disconnected cluster takes hours, so if I'll have to install it every day (or once per two days), I'll not have time for test execution, failures debugging etc. on that cluster.

The automation of those configurations is in progress, so hopefully, we will be able to install it fully automatically soon, but right now, it still require manual steps (for proxy env. it is not too difficult, for disconnected cluster it require lot of manual steps and takes hours).

So I think it would be great to postpone this change at least till we will be able to install those configurations automatically.

ebenahar:
Thanks @dahorak for the feedback!
We can definitely wait with merging this PR. However, we will need to work on moving away from the dependency on AWS.

vasukulkarni:
@ebenahar should we close this based on email?

PR #2809, opened 2020-08-26
Skip multipart upload test

MeridianExplorer:
The test was not supposed to be part of our RGW testing, since it uses a Python S3 client, but we do not have access to the RGW cluster from the code executioner, which means this code cannot run successfully.

This will only be possible to use in a future case where RGW traffic is externalized with an endpoint and becomes accessible from outside the cluster.

Done because of #2832

### DIFF
--- a/tests/manage/rgw/test_multipart_upload.py
+++ b/tests/manage/rgw/test_multipart_upload.py
@@ -50,6 +50,7 @@ class TestS3MultipartUpload(ManageTest):
     """
     @tier1
     @pytest.mark.polarion_id("OCS-2245")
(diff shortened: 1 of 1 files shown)


fbalak:
Wouldn't be better to add skip marker on the test case with note about need for RGW traffic to be externalized? It looks like a fine test case and it might be useful in future.

MeridianExplorer:
Skipped the test instead of deleting it, per @fbalak's recommendation

fbalak (approved):

RazTamir (approved):

ebenahar (approved):

ebenahar (approved):

fbalak (approved):

PR #2810, opened 2020-08-26
WIP: cfg file for 6 workers setup created

AviadP:
Added config file for with 6 worker nodes

### DIFF
--- a/conf/ocsci/vsphere_upi.yaml
+++ b/conf/ocsci/vsphere_upi.yaml
@@ -2,6 +2,6 @@
 ENV_DATA:
   platform: 'vsphere'
   deployment_type: 'upi'
-  worker_replicas: 3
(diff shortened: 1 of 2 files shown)


conf/ocsci/vsphere_upi.yaml
```diff
@@ -2,6 +2,6 @@
 ENV_DATA:
   platform: 'vsphere'
   deployment_type: 'upi'
-  worker_replicas: 3
+  worker_replicas: 6
```
petr-balogh, line 5: Added WIP to really prevent this to be merged. I guess it's just for testing purpose, right? If not you will need to create another file with this change.

AviadP: its just for testing, yes

PR #2820, opened 2020-08-27
Fixed Vsphere and BareMetal naming in codespeed

SvetlanaRedHat:
The current keys in the codespeed for the environment variables are in uppercase, this fixes the results that were not getting updated due to case sensitive match.

### DIFF
--- a/ocs_ci/utility/performance_dashboard.py
+++ b/ocs_ci/utility/performance_dashboard.py
@@ -58,7 +58,7 @@ def initialize_data():
     if m and m.group(1) is not None:
         ocs_ver = m.group(1)
     log.info(f'ocs_ver is {ocs_ver_full}')
-    platform = config.ENV_DATA['platform']
+    platform = config.ENV_DATA['platform'].upper()
     if platform.lower() not in ['vsphere', 'baremetal']:
         platform = f'{platform.upper()} {worker_type}'
     data_template['commitid'] = ocs_ver_full


ocs_ci/utility/performance_dashboard.py
```diff
@@ -59,7 +59,7 @@ def initialize_data():
         ocs_ver = m.group(1)
     log.info(f'ocs_ver is {ocs_ver_full}')
     platform = config.ENV_DATA['platform']
-    if platform.lower() not in ['vsphere', 'baremetal']:
+    if platform.lower() not in ['VSPHERE', 'BAREMETAL']:
```
keesturam, line 62: I think the 'if condition' will never be met as `platform.lower()` is always lowercase?

SvetlanaRedHat: Thanks for your comment, I will re-visit the code

petr-balogh:
Can you please refer to the issue you are trying to fix here?  Thanks

ocs_ci/utility/performance_dashboard.py
```diff
@@ -59,7 +59,7 @@ def initialize_data():
         ocs_ver = m.group(1)
     log.info(f'ocs_ver is {ocs_ver_full}')
     platform = config.ENV_DATA['platform']
```
Avilir, line 61:
```suggestion
    platform = config.ENV_DATA['platform'].upper()
```

ocs_ci/utility/performance_dashboard.py
SvetlanaRedHat: this has been removed

Avilir (approved):
LGTM

wusui:
Looks okay.  I assume that data_template['environment']  should be in uppercase when this is done.
One minor issue:  the upper case call in line 63 is probably redundant since line 61 already sets it to upper

wusui (approved):

tiffanyn108 (approved):
LGTM

vasukulkarni:
@SvetlanaRedHat have you tested this ? also can you add the fixes in the first comment?

vasukulkarni (approved):

Issue #2821, opened 2020-08-28
Remove autouse from mcgcli_pod_fixture


fbalak:
[mcgcli_pod_fixture](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/conftest.py#L1586) has set autouse to True. This means that the pod is created for all tests. This fixture should be used only when mcg tests are run.

mbukatov:
Ah, I was wondering what the heck is happening with a non mcg run I started yesterday with mcg failing.

PR #2825, opened 2020-08-28
make mcgcli_pod_fixture a prerequisite for mcg_obj fixture

fbalak:
In this PR is dropped `autosuse` in `mcgcli_pod_fixture` and making the fixture a prerequisite for `mcg_obj` fixture.
@MeridianExplorer Does it make sense? `mcg_obj` fixture is used in almost all mcg tests. It seems that tests that don't use `mcg_obj` fixture wouldn't benefit from this.

Fixes https://github.com/red-hat-storage/ocs-ci/issues/2821

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1583,7 +1583,7 @@ def rgw_obj_fixture(request):
     return RGW()
 
 
-@pytest.fixture(scope='session', autouse=True)
+@pytest.fixture(scope='session')
 def mcgcli_pod_fixture(request):
     """
     Grab the MCG CLI tool from the NooBaa operator pod,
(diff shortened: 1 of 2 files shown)


mbukatov (approved):

vasukulkarni:
tests are failing, not sure

fbalak:
There was only one failure in teardown when deleting uls. Other mcg tests passed. I don't think that the failure was caused by this PR. Let's wait for input from @MeridianExplorer.
Also I added link to `mcgcli_pod_fixture` into `test_deployment` because deployment failed without it on sanity check.

MeridianExplorer:
The ULS deletion problem is known, and is unrelated to the PR.
The solution offered in this PR seems good - can we please have a run of all tiers over `tests/manage/mcg/`?

MeridianExplorer (approved):

ebenahar:
Marking as verified based on @fbalak and @MeridianExplorer observations

Issue #2839, opened 2020-08-31
Error is Error from server (NotFound): pods "mcgcli-pod" not found


petr-balogh:
This happens on manually installed cluster when we would like to run some tests:

```
ocs_ci/ocs/cluster.py:114: in __init__
    self.mcg_obj = MCG()
ocs_ci/ocs/resources/mcg.py:70: in __init__
    self.mcgcli_pod = pod.get_pod_obj(name='mcgcli-pod', namespace=self.namespace)
ocs_ci/ocs/resources/pod.py:985: in get_pod_obj
    ocp_dict = ocp_obj.get(resource_name=name)
ocs_ci/ocs/ocp.py:207: in get
    return self.exec_oc_cmd(command)
ocs_ci/ocs/ocp.py:126: in exec_oc_cmd
    out = run_cmd(
ocs_ci/utility/utils.py:408: in run_cmd
    completed_process = exec_cmd(cmd, secrets, timeout, ignore_error, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
 
cmd = ['oc', '-n', 'openshift-storage', '--kubeconfig', '/home/sraghave/Downloads/auth/kubeconfig', 'get', ...]
secrets = None, timeout = 600, ignore_error = False, kwargs = {}
masked_cmd = 'oc -n openshift-storage --kubeconfig /home/sraghave/Downloads/auth/kubeconfig get Pod mcgcli-pod -n openshift-storage -o yaml'
completed_process = CompletedProcess(args=['oc', '-n', 'openshift-storage', '--kubeconfig', '/home/sraghave/Downloads/auth/kubeconfig', 'g...orage', '-o', 'yaml'], returncode=1, stdout=b'', stderr=b'Error from server (NotFound): pods "mcgcli-pod" not found\n')
masked_stdout = '', masked_stderr = 'Error from server (NotFound): pods "mcgcli-pod" not found\n'
 
    def exec_cmd(cmd, secrets=None, timeout=600, ignore_error=False, **kwargs):
        """
        Run an arbitrary command locally
   
        Args:
            cmd (str): command to run
            secrets (list): A list of secrets to be masked with asterisks
                This kwarg is popped in order to not interfere with
                subprocess.run(``**kwargs``)
[...]

PR #2840, opened 2020-08-31
Fix mcg-cli pod issue

petr-balogh:
Fixes: #2839


Avilir (approved):

vavuthu (approved):

ebenahar:
@petr-balogh @MeridianExplorer How is this affected by the merging of https://github.com/red-hat-storage/ocs-ci/pull/2843?

petr-balogh:
> #2843

It's not affected but this is better approach and fixes some other issues as we don't need to instantiate MCG object at the start of StorageCluster object as there are cases the MCG is not needed at all.  And if it's needed it will be taken from property and instantiated.

PR #2841, opened 2020-08-31
Check if cluster was upgraded for RGW count validation

ebenahar:

### DIFF
--- a/ocs_ci/ocs/resources/ocs.py
+++ b/ocs_ci/ocs/resources/ocs.py
@@ -233,3 +233,14 @@ def get_ocs_csv():
     log.info(f"Check if OCS operator: {ocs_csv_name} is in Succeeded phase.")
     ocs_csv.wait_for_phase(phase="Succeeded", timeout=600)
     return ocs_csv
+
+
+def check_if_cluster_was_upgraded():
+    """
+    Check whether the OCS cluster went through upgrade
+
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/resources/ocs.py
```diff
@@ -233,3 +233,14 @@ def get_ocs_csv():
...
+
+def check_if_cluster_was_upgraded():
+    """
+    Check whether the OCS cluster went through upgrade
+
+    Returns:
+        bool: True if the OCS cluster went through upgrade, False otherwise
+
+    """
+    return True if 'replaces' in get_ocs_csv().get().get('spec') else False
```
RazTamir, line 246: if the cluster got upgraded before, it will be 'replaces' and you will fail to know if it was upgraded to 4.5 right?

ebenahar: that's the intention here, to return True if the cluster got upgraded. And the indication is the existence of `replaces` under the CSV `spec`.
I hope this answers the question

AviadP (approved):

ebenahar:
Verified locally
[test_log.zip](https://github.com/red-hat-storage/ocs-ci/files/5167647/test_log.zip)

RazTamir (approved):

PR #2866, opened 2020-09-03
Log an error instead of raising an exception if ULS is found after deletion

### LINKED ISSUE #2812: Report a warning instead of a failure in cases where the ULS isn't deleted within the time limit
As of now, multiple MCG tests are creating AWS S3 buckets. As part of their teardowns, they try to remove the buckets.
However, bucket removal is an asynchronous operation, and sometimes, the bucket remains healthy for several minutes/hours - and is only then actually deleted.

Thus, it'd be preferable that we don't report a failure in case we fail to delete the ULS, but rather to report a warning.

MeridianExplorer:
Sometimes, AWS takes longer than 3 minutes to actually clean up buckets.
It'd make no sense to arbitrarily wait for however long this may take, but we also shouldn't report a false failure in such a case.
Closes #2812

### DIFF
--- a/ocs_ci/ocs/resources/cloud_manager.py
+++ b/ocs_ci/ocs/resources/cloud_manager.py
@@ -12,7 +12,6 @@ from google.cloud import storage
 
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants
-from ocs_ci.ocs.exceptions import TimeoutExpiredError
 from ocs_ci.utility import templating
 from ocs_ci.utility.aws import update_config_from_s3
 from ocs_ci.utility.utils import TimeoutSampler, load_auth_config
--- a/ocs_ci/ocs/resources/cloud_manager.py
+++ b/ocs_ci/ocs/resources/cloud_manager.py
@@ -170,14 +170,16 @@ class S3Client(CloudClient):
         assert self.exec_uls_deletion(name)
         # Todo: rename client to resource (or find an alternative)
         sample = TimeoutSampler(
-            timeout=60, sleep=6, func=self.verify_uls_exists,
+            timeout=180, sleep=15, func=self.verify_uls_exists,
             uls_name=name
         )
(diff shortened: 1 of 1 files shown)


fbalak (commented):
Increasing timeout is a good idea but isn't it a bug if uls can not be deleted within the time and is stuck? If we just log warning then we might miss some product issues...

MeridianExplorer:
@fbalak - Actually, ULS deletion is entirely out of scope for our testing. All ULS providers are outsider products - GCP, Azure, AWS, IBM, etc.
Failing a test because of a failure to delete ULS will actually falsely cause OCS-CI/NooBaa to seem buggy, while in reality, it's unrelated

ocs_ci/ocs/resources/cloud_manager.py
```diff
@@ -170,14 +169,16 @@ def internal_delete_uls(self, name):
...
+            timeout=180, sleep=15, func=self.verify_uls_exists,
             uls_name=name
         )
         if not sample.wait_for_func_status(result=False):
             logger.error(
                 f'Deletion of Underlying Storage {name} timed out. Unable to delete {name}'
             )
-            raise TimeoutExpiredError
+            logger.warning(
+                f'AWS S3 bucket {name} still found after 3 minutes, and might require manual removal.'
```
ebenahar, line 180: is it ok not to fail the test case in such situation?

ebenahar: sorry for re-asking, now I see Filip's question and your answer

ebenahar (approved):

fbalak (approved):

MeridianExplorer:
Verifying this PR is a tad problematic because of the inconsistent nature of the bucket deletion timeout.
Thus, we choose to assume this is safe and simple enough to merge without actual verification.

Issue #2598, opened 2020-07-30
Collect must-gather even after a "Green/Successful" Tier runs/fresh deployment


nehaberry:
At times, there are no test failures in tier executions/deployment jobs and hence no logs are collected. But , many times, we still need some details about the cluster & the run. In the absence of any must-gather, it becomes difficult to get/provide those details

Some suggestions:
1. Collect ocs-must-gather after a successful deployment (this is a proof how cluster looked immediately after deployment and we would have deployment logs)

2. Collect ocs-must-gather & ocp-must-gather at the end of a tier execution, even in case when there were no failures. This helps in comparison, getting cluster info , etc.

nehaberry:
@petr-balogh ocs-ci issue raised for the discussion about collecting OCS logs even after Green deployments/Tier executions

petr-balogh:
@jimdevops19  can you take this up please?

Thanks

jimdevops19:
@petr-balogh Sure.

PR #2880, opened 2020-09-07
Collecting must gather after every status of deployment

jimdevops19:
Allows to collect logs as well as if the OCS deployment succeeds.
That way we will have information about the cluster as well as if the deployment succeeded.

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -91,6 +91,7 @@ REPORTING:
   ocs_must_gather_image: "quay.io/rhceph-dev/ocs-must-gather"
   default_ocs_must_gather_latest_tag: 'latest-4.6'
   gather_on_deploy_failure: true
+  gather_on_deploy_success: true
 
 # This is the default information about environment.
 ENV_DATA:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -37,7 +37,7 @@ from ocs_ci.ocs.mcg_workload import (
 )
 from ocs_ci.ocs.node import get_node_objs, schedule_nodes
 from ocs_ci.ocs.ocp import OCP
-from ocs_ci.ocs.utils import setup_ceph_toolbox
+from ocs_ci.ocs.utils import setup_ceph_toolbox, collect_ocs_logs
 from ocs_ci.ocs.resources.backingstore import (
     backingstore_factory as backingstore_factory_implementation
 )
@@ -2892,3 +2892,20 @@ def snapshot_restore_factory(request):
 
     request.addfinalizer(finalizer)
     return factory
+
+
+@pytest.fixture(scope="session", autouse=True)
+def collect_logs_fixture(request):
+    """
+    This fixture collects ocs logs after tier execution and this will allow
+    to see the cluster's status after the execution on all execution status options.
+    """
+    def finalizer():
+        """
+        Tracking both logs separately reduce changes of collision
+        """
+        if not config.RUN['cli_params'].get('deploy') and not config.RUN['cli_params'].get('teardown'):
+            collect_ocs_logs('testcases', ocs=False, status_failure=False)
+            collect_ocs_logs('testcases', ocp=False, status_failure=False)
+
+    request.addfinalizer(finalizer)
--- a/.functional_ci_setup.py
+++ b/.functional_ci_setup.py
@@ -97,6 +97,7 @@ def get_ocsci_conf():
         ),
         REPORTING=dict(
             gather_on_deploy_failure=True,
+            gather_on_deploy_success=True,
         )
     )
     if env.get("DOWNSTREAM") == "true":
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -123,6 +123,8 @@ class Deployment(object):
         if not config.ENV_DATA['skip_ocs_deployment']:
             try:
                 self.deploy_ocs()
+                if config.REPORTING['gather_on_deploy_success']:
+                    collect_ocs_logs('deployment', ocp=False, status_failure=False)
             except Exception as e:
                 logger.error(e)
                 if config.REPORTING['gather_on_deploy_failure']:
--- a/ocs_ci/ocs/utils.py
+++ b/ocs_ci/ocs/utils.py
@@ -783,7 +783,7 @@ def collect_noobaa_db_dump(log_dir_path):
     )
 
 
-def collect_ocs_logs(dir_name, ocp=True, ocs=True, mcg=False):
+def collect_ocs_logs(dir_name, ocp=True, ocs=True, mcg=False, status_failure=True):
     """
     Collects OCS logs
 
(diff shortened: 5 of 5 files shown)


jimdevops19 (commented):
In order to work from scratch, this PR is created, so the [2647]( https://github.com/red-hat-storage/ocs-ci/pull/2647) is closed.

tests/conftest.py
```diff
@@ -2621,3 +2621,19 @@ def ns_resources_and_connections_cleanup():
...
+def collect_logs_fixture(request):
+    """
+    This fixture collects ocs logs after tier execution and this will allow
+    to see the cluster's status after the execution on all execution status options.
+    """
+    def finalizer():
+        """
+        Tracking both logs separately reduce changes of collision
+        """
+        if not config.RUN['cli_params'].get('deploy') and not config.RUN['cli_params'].get('destroy'):
```
jimdevops19, line 2635: This is the expected behavior because we collect the must-gather for any deployment status as well before.

clacroix12: So I have two questions:

1. Where is this fixture being used?

2. What happens when we override the configuration and set `gather_on_deploy_success` to `False` yet we still want to gather logs for the tier executions. In that scenario, we won't gather logs for the tier execution if we combine the deployment and tier in one command.

My point here is that I don't think keying off of the `deploy`/`destroy` `cli_params` is the right way to determine if we should be retrieving must gather logs.

tests/conftest.py
```diff
@@ -2621,3 +2621,19 @@ def ns_resources_and_connections_cleanup():
     request.addfinalizer(ns_resources_and_connections_cleanup)
 
     return _create_ns_resources
+
+
+def collect_logs_fixture(request):
```
fbalak, line 2626: `@pytest.fixture` marker is missing.

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -91,6 +91,7 @@ REPORTING:
   ocs_must_gather_image: "quay.io/rhceph-dev/ocs-must-gather"
   default_ocs_must_gather_latest_tag: 'latest-4.5'
   gather_on_deploy_failure: true
+  gather_on_deploy_success: true
```
fbalak, line 94: Why should be must gather collected for successful deployment by default?

jimdevops19: When jobs and tests complete without failures, no logs are collected. And sometimes there are still some details about the cluster and the execution are needed.
So collecting must-gather after a successful deployment could be a good way to be aware of the cluster's situation immediately after deployment and we would have deployment logs.

It was requested and discussed [here](https://github.com/red-hat-storage/ocs-ci/issues/2598)

tests/conftest.py
jimdevops19: I understand, might that be a better idea to search for the value provided for **test mark expression** ? 
Considering the following execution: ```run-ci tests/ -m tier1```
But in that case, I feel like I will have to search for keying off the deploy and the destroy again as the value of **-m**.
Am I missing something ?

ocs_ci/framework/conf/default_config.yaml
fbalak: Ok :+1:

fbalak (changes requested):
It seems that in this PR is now a lot of commits from other people that are not relevant to must-gather and deployment. This should be fixed so that only relevant commits are added.

fbalak (commented):
LGTM but I think that on test failure would be must gather collected twice if I understand it correctly.

petr-balogh:
@jimdevops19 we need to verify also the the deployment of OCP only as engineering running deployment in two stages - one deploy OCP only with skip of OCS - second one which skipping deploy of OCP and deploy OCS only.  You can run just deploy job directly to cover this scenario to provide config for skip OCS deploy, run with your PR and then run on that same cluster install again with skip OCP deployment config.  

Also I think we will have some conflict with this PR:  https://github.com/red-hat-storage/ocs-ci/pull/2913

But let's solve the conflict once one of the PR will get merged.

jimdevops19:
> @jimdevops19 we need to verify also the the deployment of OCP only as engineering running deployment in two stages - one deploy OCP only with skip of OCS - second one which skipping deploy of OCP and deploy OCS only. You can run just deploy job directly to cover this scenario to provide config for skip OCS deploy, run with your PR and then run on that same cluster install again with skip OCP deployment config.
> 
> Also I think we will have some conflict with this PR: #2913
> 
> But let's solve the conflict once one of the PR will get merged.
I see. Working on that.

tests/conftest.py
```diff
@@ -2592,3 +2592,20 @@ def ns_resources_and_connections_cleanup():
...
+    """
+    This fixture collects ocs logs after tier execution and this will allow
+    to see the cluster's status after the execution on all execution status options.
+    """
+    def finalizer():
+        """
+        Tracking both logs separately reduce changes of collision
+        """
+        if not config.RUN['cli_params'].get('deploy') and not config.RUN['cli_params'].get('destroy'):
```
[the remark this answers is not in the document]
nehaberry: +1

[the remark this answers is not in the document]
fbalak: After reviewing logs provided by @jimdevops19 I see that this may not be a problem. The hook collects must gathers right after each failed test and this new fixture collects logs after test run finishes. It might be useful to have new must gathers right after all tests finish even if we already have some must gathers collected from failed test cases. @petr-balogh @clacroix12 What do you think?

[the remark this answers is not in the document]
petr-balogh: That's the case here (why we have the issue opened) to collect at least once the must gather after all tests finished.  Or I didn't get you @fbalak

petr-balogh, line 2607: destroy -> teardown as we don't have any destroy parameter but --teradown https://ocs-ci.readthedocs.io/en/latest/docs/usage.html that's why we see the must gather collected also for destroy job.
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-destroy-ocs-cluster/9714/console

tests/conftest.py
petr-balogh: @clacroix12 I think no one does deployment and tier1 in one execution so we don't have to deal with that case I think at the moment.

petr-balogh (approved):
@fbalak hey Filip, can you please approve or at least remove request for change? Or is any other change still requested? Thanks

fbalak (dismissed):
Ok, my requested changes were resolved.

tests/conftest.py
jimdevops19: Fix of that verified here: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/12626/
(Only one folder generated)

petr-balogh:
@jimdevops19 I see some conflict with tests/conftest.py now. Please rebase

petr-balogh (approved):

fbalak (approved):

tests/conftest.py
```diff
@@ -2892,3 +2892,20 @@ def finalizer():
...
+def collect_logs_fixture(request):
+    """
+    This fixture collects ocs logs after tier execution and this will allow
+    to see the cluster's status after the execution on all execution status options.
+    """
+    def finalizer():
+        """
+        Tracking both logs separately reduce changes of collision
+        """
+        if not config.RUN['cli_params'].get('deploy') and not config.RUN['cli_params'].get('teardown'):
```
vasukulkarni, line 2907: should this check if the deployment was run?

PR #2900, opened 2020-09-09
Extend OBC health verification timeout

### LINKED ISSUE #2899: Extend OBC status verification from 30 to 60 seconds
At the moment, we've seen several cluster where OBCs took longer than 30 seconds to reach a _Bound_ state.
Thus, it'd be sensible to extend the status check timeout from 30 to 60 seconds.

MeridianExplorer:
Closes #2899

### DIFF
--- a/ocs_ci/ocs/resources/objectbucket.py
+++ b/ocs_ci/ocs/resources/objectbucket.py
@@ -162,7 +162,7 @@ class ObjectBucket(ABC):
             )
             assert False, f'{self.name} was not deleted within {timeout} seconds.'
 
-    def verify_health(self, timeout=30, interval=5):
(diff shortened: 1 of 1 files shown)


fbalak (approved):
Do you think that it can be calculated dynamically from some metric or is it platform dependent?

MeridianExplorer:
Thanks for the review, @fbalak :)  
Sadly, I do not know of any metric that allows us to anticipate the time it takes to create OBCs in a given cluster. In the absolute most of the cases, it takes roughly between 5 and 15 seconds. But in some external runs, I saw it going up to even 5 minutes...

ebenahar (approved):

Issue #2916, opened 2020-09-10
Remove workaround: skip RGW tests when upgrading from OCS 4.4 to 4.5


MeridianExplorer:
Currently, because of #2826, we have to skip RGW tests in clusters that were upgraded from OCS 4.4 to 4.5.

PR #2917, opened 2020-09-10
Skip RGW tests when cluster is upgraded from OCS 4.4

MeridianExplorer:
Implements a workaround for #2916


tests/manage/rgw/conftest.py
```diff
@@ -16,7 +17,8 @@ def pytest_collection_modifyitems(items):
     """
     if (
         config.ENV_DATA['platform'].lower() not in ON_PREM_PLATFORMS
-        or float(config.ENV_DATA['ocs_version']) < 4.5
+        or skipif_ocs_version('<4.5')
+        or skipif_upgraded_from('4.4')
```
ebenahar, line 21: maybe would be good to add a skip reason. If agreed, it will probably require changes in skipif_upgraded_from to accept a reason

fbalak: Also it would be good if `skipif_upgraded_from` accepted expressions like `skipif_ocs_version`. This PR currently covers only upgrade from OCS 4.4 but if upgrade to OCS 4.5 is triggered from older version (<4.4) then it's not skipped. For now we can use probably `['4.2', '4.3', '4.4']` as argument for `skipif_upgraded_from`.

MeridianExplorer: @ebenahar - I think it'd be better to just independently log the reason for the skip. I'll add it.
@fbalak - Great point, thank you, I'll add this too

ebenahar (dismissed):

fbalak (dismissed):

docs/conf.py
```diff
@@ -44,7 +44,11 @@
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
-exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store', '.venv', 'venv', 'tests', '.tox', 'data', '.pytest_cache', 'usecases/README.rst']
+exclude_patterns = [
```
MeridianExplorer, line 47: Had to change that in order to pass the pre-commit hooks

MeridianExplorer:
This upgrade path is no longer tested.

PR #2930, opened 2020-09-13
test_pod_reattachtime_fix

ypersky1980:


ebenahar (approved):

wusui:
Is this pod always going to be a debian pod?  If it isn't, then the apt-get will fail.

I am also following this because I may need to install rsync on a pod too.

Avilir (approved):

PR #2935, opened 2020-09-14
Enhanced TC for External mode, fixed pvc capacity problem

ramkiperiy:
Fixed fio_file_size bigger than PVC_size problem
Enhanced TC to support external mode cluster.

### DIFF
--- a/tests/e2e/scale/test_pods_are_not_oomkilled_while_running_ios.py
+++ b/tests/e2e/scale/test_pods_are_not_oomkilled_while_running_ios.py
@@ -2,7 +2,7 @@ import logging
 import pytest
 
 from ocs_ci.ocs import constants, defaults, ocp
-from ocs_ci.ocs.resources import pod
+from ocs_ci.ocs.resources import pod as Pod
 from ocs_ci.framework import config
 from ocs_ci.ocs.resources.pod import get_all_pods
 from ocs_ci.framework.testlib import E2ETest, scale
@@ -47,7 +47,7 @@ class TestPodAreNotOomkilledWhileRunningIO(E2ETest):
 
         """
         # Setting the io_size_gb to 40% of the total PVC capacity
-        ceph_pod = pod.get_ceph_tools_pod()
+        ceph_pod = Pod.get_ceph_tools_pod()
         external = config.DEPLOYMENT['external_mode']
         if external:
             ocp_obj = ocp.OCP()
--- a/tests/e2e/scale/test_pods_are_not_oomkilled_while_running_ios.py
+++ b/tests/e2e/scale/test_pods_are_not_oomkilled_while_running_ios.py
@@ -1,7 +1,9 @@
 import logging
 import pytest
 
-from ocs_ci.ocs import constants, defaults
+from ocs_ci.ocs import constants, defaults, ocp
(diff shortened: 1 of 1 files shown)


tests/e2e/scale/test_pods_are_not_oomkilled_while_running_ios.py
```diff
@@ -45,9 +46,30 @@ def base_setup(
...
+        else:
+            pool = constants.DEFAULT_BLOCKPOOL if interface == constants.CEPHBLOCKPOOL else constants.DATA_POOL
+
+        ceph_replica = ceph_pod.exec_ceph_cmd(ceph_cmd=f"ceph osd pool get {pool} size")
+        replica = ceph_replica['size']
+        ceph_status = ceph_pod.exec_ceph_cmd(ceph_cmd="ceph df")
+        ceph_capacity = int(ceph_status['stats']['total_bytes']) / replica / constants.GB
+        pvc_size_gb = int(ceph_capacity * 0.5)
+        io_size_gb = int(pvc_size_gb * 0.4)
+        if io_size_gb >= 600:
```
keesturam, line 71: maybe, `io_size_gb = 400 if io_size_gb >= 400` would be better, as we are setting io_size_gb above 600GB to 400, but any value between 400 to 600 is left as is.

ramkiperiy: ack.

ramkiperiy:
PR needs validation on external mode cluster.

ramkiperiy:
There is no external cluster available for testing, Only change is to get couple of inputs from tools_pod, earlier it was collected using the OCS ceph pods, now made the changes to get those inputs from tools_pod.
Note: Using tools pod output TC already passed for Internal-mode cluster and tool_pod output is no difference in case of external-mode.

Akarsha-rai (approved):
LGTM

vasukulkarni (approved):

tests/e2e/scale/test_pods_are_not_oomkilled_while_running_ios.py
```diff
@@ -45,9 +46,29 @@ def base_setup(
...
+
+        else:
+            pool = constants.DEFAULT_BLOCKPOOL if interface == constants.CEPHBLOCKPOOL else constants.DATA_POOL
+
+        ceph_replica = ceph_pod.exec_ceph_cmd(ceph_cmd=f"ceph osd pool get {pool} size")
+        replica = ceph_replica['size']
+        ceph_status = ceph_pod.exec_ceph_cmd(ceph_cmd="ceph df")
+        ceph_capacity = int(ceph_status['stats']['total_bytes']) / replica / constants.GB
+        pvc_size_gb = int(ceph_capacity * 0.5)
+        io_size_gb = int(pvc_size_gb * 0.4)
```
vasukulkarni, line 70: @ramkiperiy this is a shared cluster in case of external mode, should we reduce this a bit more or is it fine?

ramkiperiy: Even if we get higher value, in next line there is a check to have io_size_gb not be more than 400G

PR #2945, opened 2020-09-15
Change the PVC multiple creation to work as bulk

Avilir:

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -589,11 +589,11 @@ def create_pvc(
 
 
 def create_multiple_pvcs(
-    sc_name, namespace, number_of_pvc=1, size=None,
-    access_mode=constants.ACCESS_MODE_RWO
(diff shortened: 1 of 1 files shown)


ebenahar (approved):

tests/helpers.py
```diff
@@ -589,7 +589,7 @@ def create_pvc(
 
 
 def create_multiple_pvcs(
-    sc_name, namespace, number_of_pvc=1, size=None, do_reload=False,
+    sc_name, namespace, number_of_pvc=1, size=None,
```
vasukulkarni, line 592: if we can add bulk option then we can keep original signature for other cases.

Avilir: done

vasukulkarni (approved):

ramkiperiy (approved):

PR #2961, opened 2020-09-16
Disable test case: add MDS to cluster

ebenahar:


vasukulkarni (approved):

vasukulkarni:
If we are not going to use it, we can very well delete it.

ebenahar:
> If we are not going to use it, we can very well delete it.

Maybe this will be supported in one day :)

ebenahar:
Verified locally 
[log.zip](https://github.com/red-hat-storage/ocs-ci/files/5234833/log.zip)

jilju (approved):

vavuthu (approved):

PR #2964, opened 2020-09-17
Extend MCG node drain cluster health check tries + reduce load

MeridianExplorer:
Seems that right now, when the tests run on a cluster under load, the timeout isn't long enough, and errors out on teardown.
This should resolve the issue by lightening the load + adding 10 additional tries.

Added a retry inside MCG instantiation that should address #3203


ebenahar (dismissed):

fbalak (approved):

fbalak (dismissed):

ebenahar (approved):

ebenahar (approved):

ebenahar (dismissed):

MeridianExplorer:
Verified on this run with IO in the background - 
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/13958/

ebenahar (dismissed):

bernerhat (dismissed):

ocs_ci/ocs/resources/mcg.py
```diff
@@ -173,6 +166,44 @@ def __init__(self, *args, **kwargs):
...
+            try:
+                rpc_response = self.send_rpc_query(
+                    'auth_api', 'create_auth', params={
+                        'role': 'admin',
+                        'system': 'noobaa',
+                        'email': self.noobaa_user,
+                        'password': self.noobaa_password
+                    }
+                )
+                return rpc_response.json().get('reply').get('token')
```
ebenahar, line 184: this method can return 2 different types - the one in this line, bool from https://github.com/red-hat-storage/ocs-ci/pull/2964/files#diff-1d32a061b0f98ee283c52d1f7e9ca822177982cee47f7daa76904f5d8e1c3bfdR192 and token object from https://github.com/red-hat-storage/ocs-ci/pull/2964/files#diff-1d32a061b0f98ee283c52d1f7e9ca822177982cee47f7daa76904f5d8e1c3bfdR197

This doesn't look like a good practice

MeridianExplorer: Why would this be problematic?
This function can only be called strictly from within `retrieve_nb_token`, and it either returns `False` or a token string, which is being handled [here](https://github.com/red-hat-storage/ocs-ci/pull/2964/files#diff-1d32a061b0f98ee283c52d1f7e9ca822177982cee47f7daa76904f5d8e1c3bfdR196)

MeridianExplorer: Discussed offline; Changing to return None instead of False

ebenahar: found a good explanation here https://stackoverflow.com/questions/1839289/why-should-functions-always-return-the-same-type#:~:text=Sure%2C%20like%20everything%20else.&text=I%20personally%20think%20it%20is,string%20and%20list%20for%20example.

ebenahar (approved):

bernerhat (approved):

PR #2970, opened 2020-09-18
Changes needed for OCP 4.6 on BM platform

prsurve:


prsurve:
Verification job
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/12513/console

ocs_ci/deployment/baremetal.py
```diff
@@ -190,6 +191,21 @@ def deploy_prereq(self):
                 assert self.helper_node_handler.exec_cmd(cmd=cmd), "Failed to Download required File"
             else:
                 raise RhcosImageNotFound
+
+            if ocp_version == "4.6":
+                # Download metal_bios
+                rootfs_image_path = constants.coreos_url_prefix + image_data['live_rootfs_url']
```
vavuthu, line 197: use os.path.join()

ocs_ci/deployment/baremetal.py
```diff
@@ -423,6 +424,43 @@ def update_bm_status(self, bm_status):
...
+                ocp_version (str): OCP version
+                role (str): Role of node eg:- bootstrap,master,worker
+
+            Returns:
+                str: temp file path
+
+            """
+            extra_data = ""
+            if ocp_version == "4.6":
+                extra_data = f"coreos.live.rootfs_url={self.helper_node_details['bm_install_files']}" \
```
vavuthu, line 441: avoid line breaks

zmc: could also make quite a few lines shorter by storing `self.helper_node_details['bm_install_files']` in its own variable

ocs_ci/deployment/baremetal.py
prsurve: Done

ocs_ci/deployment/baremetal.py
prsurve: Done

vasukulkarni (dismissed):

ocs_ci/deployment/baremetal.py
```diff
@@ -190,6 +191,21 @@ def deploy_prereq(self):
                 assert self.helper_node_handler.exec_cmd(cmd=cmd), "Failed to Download required File"
             else:
                 raise RhcosImageNotFound
+
+            if ocp_version == "4.6":
```
ebenahar, line 195: shouldn't this be `if ocp_version >= "4.6":`

ocs_ci/deployment/baremetal.py
```diff
@@ -423,6 +424,44 @@ def update_bm_status(self, bm_status):
...
+                role (str): Role of node eg:- bootstrap,master,worker
+
+            Returns:
+                str: temp file path
+
+            """
+            extra_data = ""
+            bm_install_files_loc = self.helper_node_details['bm_install_files']
+            extra_data_pxe = "rhcos-live-rootfs.x86_64.img coreos.inst.insecure"
+            if ocp_version == "4.6":
```
ebenahar, line 442: same

ocs_ci/deployment/baremetal.py
prsurve: here we are not sure how it's going to be in 4.7 so I have used only 4.6 if I get that it will be the same will use 4.6 or greater release

dahorak (dismissed):

ocs_ci/deployment/baremetal.py
prsurve: os.path.join doesn't work here
```import os
a="https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos"
b="/pre-release/latest-4.6/rhcos-live-rootfs.x86_64.img"
print(os.path.join(a,b))                                                                                                                                                                                                              
/pre-release/latest-4.6/rhcos-live-rootfs.x86_64.img
```

vavuthu: you need to remove slash in the beginning ( b="pre-release/latest-4.6/rhcos-live-rootfs.x86_64.img" )

ocs_ci/deployment/baremetal.py
petr-balogh: You cannot compare >= with string @ebenahar

clacroix12: > You cannot compare >= with string @ebenahar

I thought so as well but apparently you can:

```
In [1]: version = "4.5"

In [2]: version >= "4.6"
Out[2]: False

In [3]: version <= "4.6"
Out[3]: True
```

dahorak: It is [possible to compare strings](https://docs.python.org/3/reference/expressions.html#value-comparisons), but generally it is probably not good approach (in the meaning, that it might not return expectable result):
```python
In [1]: "a" < "Z"                                                                                                                                
Out[1]: False

In [2]: "A" < "z"                                                                                                                                
Out[2]: True
```
But in this particular usecase, I think it is quite safe to use it this way.

dahorak (approved):

petr-balogh (approved):

ocs_ci/deployment/baremetal.py
prsurve: Do I need to change anything here or we are good to merge the PR

prsurve: @petr-balogh @dahorak ^^

petr-balogh: you have == so you are fine from the start of the conversation here ;)

prsurve: Thanks

Issue #2908, opened 2020-09-10
Need to add catalog source for internal operator bundle registry image for LSO deployment over OCP 4.6


petr-balogh:
As OCP 4.6 is not released yet than the nightly build does not contain the LSO operator.

We see then the error like:
476 E           ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc -n openshift-marketplace --kubeconfig /home/cloud-user/ocp45vjp_install/auth/kubeconfig get packagemanifest local-storage-operator -n openshift-mar
477 E           Error is Error from server (NotFound): packagemanifests.packages.operators.coreos.com "local-storage-operator" not found

As I got answer that LSO already support new bundle format:
```
ART team is already generating images with new format for 4.6 -
http://dist-git.host.prod.eng.bos.redhat.com/cgit/containers/local-storage-operator-bundle/tree/?h=rhaos-4.6-rhel-7.

openshift-storage (led by Jan Safranek) are maintainers of LSO.
```

And there is also one more thread which can help here.
http://post-office.corp.redhat.com/archives/aos-devel/2020-August/msg00067.html

@shylesh or @clacroix12  can you please take this task with high priority to help Vimal to get successful deployment for interop testing?

ebenahar:
For reference, the same is being discussed in https://chat.google.com/room/AAAAREGEba8/_DBjdmhlcj4

clacroix12:
I was able to use the catalogsource below in order to have the LocalStorage Operator appear in the OperatorHub in the openshift UI. However, I am still having issues actually installing it once that catalogsource has been created. I'm able to successfully retrieve the channel (4.6) from the packagemanifest, however the creation of the LSO does not succeed.

CatalogSource:
```yaml
---
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: production-operators
  namespace: openshift-marketplace
spec:
  displayName: Production Operators
  icon:
    base64data: ""
    mediatype: ""
  image: quay.io/openshift-qe-optional-operators/ocp4-index:latest
  publisher: OpenShift QE
  sourceType: grpc
```

LSO CR:
```yaml
---
apiVersion: v1
kind: Namespace
metadata:
  name: local-storage
spec: {}

---

apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  annotations:
    olm.providedAPIs: LocalVolume.v1.local.storage.openshift.io
  name: local-storage
[...]

Issue #2640, opened 2020-08-05
Change default LSO namespace from local-storage to openshift-local-storage to align with OCS 4.6


mbukatov:
LSO changed suggested default namespace from `local-storage` to `openshift-local-storage`, see https://github.com/openshift/local-storage-operator/pull/127. Moreover the namespace has been specified as suggested default namespace for the LSO operator, so that users don't have to create it themselves anymore.

We need to do matching changes since OCS 4.6.

clacroix12:
Currently the LSO namespace is specified in our constants module. Since this is no longer a constant, I suggest that we move the definition to our default config and update the value in our config during LSO depoyment based on OCP version if necessary. This way we can ensure that all uses of this namespace are pulling from the same place while ensuring it is correct for OCP 4.6 and beyond.

Once we no longer need to support the OCP < 4.6 use case (`local-storage`) we can move back it back to constants if that is preferred.

clacroix12:
I tested an LSO deployment on OCP 4.5 with the new namespace (`openshift-local-storage`) and the deployment was successful. Assuming anything that uses/references the local storage namespace does so through the constant we have defined I think it's safe to use the new namespace for all OCP versions.

mbukatov:
I agree that we can just rename the default namespace from `local-storage` to `openshift-local-storage`.

Strictly speaking,we should use `local-storage` as a default namespace for OCP < 4.6 because that is what the documentation states. There are also few subtle differences in handling `openshift-` prefixed namespaces. But since our testing with OCP < 4.6 is already done, and there is no observable difference between these 2 namespace names, there is no point in trying to preserve `local-storage` namespace usage for < 4.6 test runs.

PR #2979, opened 2020-09-21
Use optional-operators catalogsource for LSO on non-GA OCP versions

clacroix12:
Fixes: #2908 
Fixes: #2640 

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -49,18 +49,20 @@ from ocs_ci.ocs.utils import (
     setup_ceph_toolbox, collect_ocs_logs
 )
 from ocs_ci.utility import templating
+from ocs_ci.utility.deployment import get_ocp_ga_version
 from ocs_ci.utility.localstorage import get_lso_channel
 from ocs_ci.utility.openshift_console import OpenshiftConsole
 from ocs_ci.utility.retry import retry
 from ocs_ci.utility.utils import (
     ceph_health_check,
     get_latest_ds_olm_tag,
+    get_ocp_version,
     is_cluster_running,
     run_cmd,
     set_selinux_permissions,
     set_registry_to_managed_state,
     add_stage_cert,
-    modify_csv
+    modify_csv, wait_for_machineconfigpool_status
 )
 from ocs_ci.utility.vsphere_nodes import update_ntp_compute_nodes
 from tests import helpers
@@ -876,6 +878,31 @@ def setup_local_storage(storageclass):
         f"{constants.OPERATOR_NODE_LABEL}"
     )
 
+    ocp_version = get_ocp_version()
+    ocp_ga_version = get_ocp_ga_version(ocp_version)
+    if not ocp_ga_version:
+        optional_operators_data = templating.load_yaml(
+            constants.LOCAL_STORAGE_OPTIONAL_OPERATORS, multi_document=True
+        )
+        logger.info(
+            "Creating temp yaml file with optional operators data:\n %s",
+            optional_operators_data
+        )
+        optional_operators_yaml = tempfile.NamedTemporaryFile(
+            mode='w+', prefix='optional_operators', delete=False
+        )
+        templating.dump_data_to_temp_yaml(
+            optional_operators_data, optional_operators_yaml.name
+        )
+        with open(optional_operators_yaml.name, 'r') as f:
+            logger.info(f.read())
+        logger.info(
+            "Creating optional operators CatalogSource and ImageContentSourcePolicy"
+        )
+        run_cmd(f"oc create -f {optional_operators_yaml.name}")
+
+        wait_for_machineconfigpool_status('all')
+
     logger.info("Retrieving local-storage-operator data from yaml")
     lso_data = list(templating.load_yaml(
         constants.LOCAL_STORAGE_OPERATOR, multi_document=True
@@ -884,6 +911,9 @@ def setup_local_storage(storageclass):
     for data in lso_data:
         if data['kind'] == 'Subscription':
             data['spec']['channel'] = get_lso_channel()
+        if not ocp_ga_version:
+            if data['kind'] == 'Subscription':
+                data['spec']['source'] = 'optional-operators'
 
     # Create temp yaml file and create local storage operator
     logger.info(
@@ -902,7 +932,7 @@ def setup_local_storage(storageclass):
     run_cmd(f"oc create -f {lso_data_yaml.name}")
 
     local_storage_operator = ocp.OCP(
-        kind=constants.POD, namespace='local-storage'
+        kind=constants.POD, namespace=constants.LOCAL_STORAGE_NAMESPACE
     )
     assert local_storage_operator.wait_for_resource(
         condition=constants.STATUS_RUNNING,
@@ -915,7 +945,7 @@ def setup_local_storage(storageclass):
     lso_type = config.DEPLOYMENT.get('type')
     if platform == constants.VSPHERE_PLATFORM:
         # Types of LSO Deployment
-        # Importing here to avoid circular dependancy
+        # Importing here to avoid circular dependency
         from ocs_ci.deployment.vmware import VSPHEREBASE
         vsphere_base = VSPHEREBASE()
 
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -932,7 +932,7 @@ def setup_local_storage(storageclass):
     run_cmd(f"oc create -f {lso_data_yaml.name}")
 
     local_storage_operator = ocp.OCP(
-        kind=constants.POD, namespace=constants.LOCAL_STORAGE_NAMESPACE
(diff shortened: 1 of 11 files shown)


ocs_ci/templates/ocs-deployment/local-storage-optional-operators.yaml
```diff
@@ -0,0 +1,44 @@
...
+  managedFields:
+  - apiVersion: operator.openshift.io/v1alpha1
+    fieldsType: FieldsV1
+    fieldsV1:
+      f:spec:
+        .: {}
+        f:repositoryDigestMirrors: {}
+    manager: oc
+    operation: Update
+    time: "2020-09-15T08:31:30Z"
```
prsurve, line 16: this lines can be removed

ocs_ci/templates/ocs-deployment/local-storage-optional-operators.yaml
```diff
@@ -0,0 +1,44 @@
...
+  name: olmcontentsourcepolicy
+  resourceVersion: "38674"
+  selfLink: /apis/operator.openshift.io/v1alpha1/imagecontentsourcepolicies/olmcontentsourcepolicy
+  uid: ff7612c6-edbc-451f-9ab7-9c67e04f3e9d
```
prsurve, line 20: same

ocs_ci/utility/deployment.py
```diff
@@ -0,0 +1,62 @@
...
+    mcp_data = json.loads(out)
+
+    for mcp in mcp_data["items"]:
+        logger.info(
+            "Verifying readyMachineCount for MachineConfigPool %s",
+            mcp["metadata"]["name"],
+        )
+        machine_count = mcp["status"]["machineCount"]
+        ready_machine_count = mcp["status"]["readyMachineCount"]
+        assert ready_machine_count == machine_count
```
prsurve, line 62: we already have a function for this 
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L2611-L2642

ocs_ci/ocs/constants.py
```diff
@@ -180,6 +180,7 @@
 RIPSAW_DROP_CACHE = os.path.join(TEMPLATE_FIO_DIR, "drop_cache_pod.yaml")
 OCP_QE_DEVICEPATH_REPO = "https://github.com/anubhav-here/device-by-id-ocp.git"
 LOCAL_STORAGE_NAMESPACE = 'local-storage'
+NON_GA_LOCAL_STORAGE_NAMESPACE = 'openshift-local-storage'
```
prsurve, line 183: The variable name should be changed from ocp 4.6 onwards we are going to use `openshift-local-storage`

ocs_ci/deployment/deployment.py
```diff
@@ -876,6 +880,30 @@ def setup_local_storage(storageclass):
...
+        )
+        optional_operators_yaml = tempfile.NamedTemporaryFile(
+            mode='w+', prefix='optional_operators', delete=False
+        )
+        templating.dump_data_to_temp_yaml(
+            optional_operators_data, optional_operators_yaml.name
+        )
+        with open(optional_operators_yaml.name, 'r') as f:
+            logger.info(f.read())
+        logger.info("Creating optional operators CatalogSource and ICSP")
```
ebenahar, line 902: what is ICSP?

prsurve: icsp stand for imagecontentsourcepolicy

clacroix12: ImageContentSourcePolicy

ocs_ci/ocs/constants.py
clacroix12: We really shouldn't be changing the values of constants since that defeats the purpose of it being a constant. If the local storage namespace is variable based on the OCP version maybe we should move this to the default config and change that value based on the OCP version when we are deploying local storage. WDYT?

prsurve: i would prefer to move local storage namespace variable based on the OCP version to the default config

prsurve: we have already issue for new lso namespace #2640

ocs_ci/utility/deployment.py
clacroix12: Removed this function and am now using `wait_for_machineconfigpool_status`

ocs_ci/ocs/constants.py
clacroix12: Swapped to using `openshift-local-storage` (regardless of OCP version) to address #2640

ocs_ci/templates/ocs-deployment/local-storage-optional-operators.yaml
clacroix12: Removed

ocs_ci/templates/ocs-deployment/local-storage-optional-operators.yaml
clacroix12: Removed

ocs_ci/deployment/deployment.py
clacroix12: I changed it to `ImageContentSourcePolicy` so there is no confusion

ocs_ci/templates/ocs-deployment/local-volume.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: local.storage.openshift.io/v1
 kind: LocalVolume
 metadata:
   name: local-block
-  namespace: local-storage
+  namespace: openshift-local-storage
```
petr-balogh, line 5: @clacroix12 is this local-storage namespace suppose still be called local-storage for OCP version < 4.6 ? If so, how do we change it?  Or are we OK to use this new namespace also for older versions of OCP?

clacroix12: @petr-balogh I commented about this on #2640 and wanted to test using the new namespace for OCP < 4.6. The previous PR test was OCS 4.5 on OCP 4.5 and it passed so I do not think we will run into any deployment issues if we are using the new namespace. As long as anything that needs to reference the local storage namespace is doing so through the constant we have defined, we should be good to go.

clacroix12:
> PR validation
> 
> Cluster Name:
> Cluster Configuration: conf/ocsci/aws_i3_local_storage.yaml
> PR Test Suite: deployment
> PR Test Path: tests/
> Additional Test Params:
> OCP VERSION: 4.6
> OCS VERSION: 4.5
> 
> [Job FAILED](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/2920//testReport) (installation failed, tests not executed).

The deployment was a success but reported as a FAILURE due to a known [issue](https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/-/issues/49) that I have just merged a fix for. Marking as verified as I have tested both OCP 4.5 and 4.6.

ocs_ci/templates/ocs-deployment/local-volume.yaml
petr-balogh: @clacroix12 only one issue I can think of is running against manually deployed cluster on OCP 4.5. Then if we somewhere query in that namespace, we will be pissed of ;)

Also was thinking about upgrade of such 4.5 cluster to 4.6 but for this kind of solution that we will use the same namespace and if we install directly to this namespace also 4.5 we should be ok to run post upgrade as we will use the same namespace.

So only one issue I see is with manual deployment but not sure how to solve this.  Some dynamic assignment of value based of current state could be the solution and store the value somewhere in the config.ENV section with our default value overwritten by old namespace if it's found and used .  But let's try to not overcomplicate the things now and see how it will go.  You can at least document this somewhere as known issue.

petr-balogh (dismissed):

clacroix12:
@petr-balogh as we discussed I have moved the definition of the local storage namespace from the constants module to our config to accommodate users that may be running tests  on an externally deployed cluster.

shylesh (approved):
+1

vasukulkarni (approved):

vasukulkarni:
Does it handle the upgrade scenario? or there is another PR to handle it? from older version where the namespace is not changed?

clacroix12:
> Does it handle the upgrade scenario? or there is another PR to handle it? from older version where the namespace is not changed?

If both the installation and upgrade are done with ocs-ci we will always use `openshift-local-storage` as the namespace so there's no translation that needs to be handled.

If someone is using ocs-ci to upgrade a cluster that was deployed separately, they can pass a configuration file (like the example added in this PR) to set what the local storage namespace is for the cluster.

If the install is

ocs_ci/templates/ocs-deployment/local-volume.yaml
vasukulkarni: is this resolved now for this PR to be merged?

clacroix12: This should be handled (or at least able to be worked around) through changing the local storage namespace from a constant to a configuration variable. Resolving.

PR #2984, opened 2020-09-22
Add minimum size of '256' GB for test purpose (non production use)

vasukulkarni:
This is basically to use in DC-CP for test purpose( the total size provided by vSAN is bottleneck and using reduced disk size will allow us to create more clusters in DC-CP)

### DIFF
--- a/conf/ocsci/minimum_test_device_size.yaml
+++ b/conf/ocsci/minimum_test_device_size.yaml
@@ -1,0 +1,5 @@
+---
+#This config file holds the disk size in GB for minimum test environments.
+ENV_DATA:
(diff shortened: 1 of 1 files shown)


Avilir (approved):

ebenahar (approved):

petr-balogh:
We have this parameter --osd-size 
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/pytest_customization/ocscilib.py#L167

And in the Jenkiins jobs we have: OSD_SIZE

So the config file for this is not needed I guess.

PR #3017, opened 2020-09-30
Fio fill up functionality

Avilir:
This PR add the functionality to fill up file on PVC.
It create a file on the PVC, and run 1MB Sequential IO up to the desire size (given as parameter)
this PR contain :
* template YAML file for the FIO parameters
* function that actually do the work
* simple test to test this function

### DIFF
--- a/ocs_ci/templates/workloads/fio/workload_io_fillup.yaml
+++ b/ocs_ci/templates/workloads/fio/workload_io_fillup.yaml
@@ -1,0 +1,10 @@
+name: fio-fillup
+filename: fio-fill-up
+rw: write
+bs: 1m
+direct: 1
+numjobs: 1
+time_based: 0
+runtime: 36000
+size: 100M
+ioengine: libaio
--- a/tests/internal/test_fio_fillup.py
+++ b/tests/internal/test_fio_fillup.py
@@ -1,0 +1,62 @@
+"""
+Module to perform IOs with several weights
+"""
+import pytest
+import logging
+
+from ocs_ci.framework.testlib import BaseTest
+from ocs_ci.ocs import constants
+
+
+logger = logging.getLogger(__name__)
+
+
+class TestPVCFillup(BaseTest):
+    """
+    Test PVC Fillup
+
+    """
+
+    @pytest.mark.parametrize(
+        argnames=["size", "percentage"],
+        argvalues=[pytest.param(*['10', '50'])]
+    )
+    def test_fillup_fs(self, size, percentage,
+                       teardown_factory, storageclass_factory, interface_iterate,
+                       pod_factory, pvc_factory):
+        """
+        Test Fill up the filesystem
+        """
+        pvc_size = int(size)
+        self.interface = interface_iterate
+        self.sc_obj = storageclass_factory(self.interface)
+
+        # Creating PVC
+        self.pvc_obj = pvc_factory(
+            interface=self.interface,
+            size=size,
+            status=constants.STATUS_BOUND
+        )
+        self.pvc_obj.reload()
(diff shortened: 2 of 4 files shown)


bengland2:
why isn't prefill: true in ripsaw CR sufficient, if indeed this is for ripsaw tests?

Avilir:
> why isn't prefill: true in ripsaw CR sufficient, if indeed this is for ripsaw tests?

@bengland2 this is for FIO as a single pod which not using ripsaw.

tests/internal/test_fio_fillup.py
```diff
@@ -0,0 +1,62 @@
...
+    Test PVC Fillup
+
+    """
+
+    @pytest.mark.parametrize(
+        argnames=["size", "percentage"],
+        argvalues=[pytest.param(*['10', '50'])]
+    )
+    def test_fillup_fs(self, size, percentage,
+                       teardown_factory, storageclass_factory, interface_iterate,
```
OdedViner, line 25: you don't need to use parametrize for one test case.
you can change:
def test_fillup_fs(self, size='10', percentage='50',....)

Avilir: @OdedViner This is true, but i used parameterize in case we would run with different set of parameters.

AviadP (approved):

tests/internal/test_fio_fillup.py
```diff
@@ -0,0 +1,62 @@
...
+        logging.info(f'Going to run on PVC of {pvc_size} GB and will fill up {percentage} %')
+        logging.info(f'the filesize will be {filesize} MB')
+        self.pod_obj.fillup_fs(size=filesize,)
+        logging.info("Waiting for results")
+
+        # Getting the FIO output and verify all data was written
+        fio_result = self.pod_obj.get_fio_results()
+        writes = int(fio_result.get('jobs')[0].get('write').get('io_kbytes') / 1024)
+        logging.info(f"Total Write: {writes} MB")
+        assert (filesize <= writes), "Not all required data was written"
```
ebenahar, line 62: parentheses are redundant

Avilir: fixed.

ocs_ci/templates/workloads/fio/workload_io_fillup.yaml
```diff
@@ -0,0 +1,10 @@
+name: fio-fillup
+filename: fio-fill-up
+rw: write
+bs: 1m
+direct: 1
+numjobs: 1
+time_based: 0
+runtime: 36000
```
keesturam, line 8: should we have such a time run time as default? If in case there are any failures, the test might take such a long time to report failure.

tests/internal/test_fio_fillup.py
```diff
@@ -0,0 +1,62 @@
...
+import logging
+
+from ocs_ci.framework.testlib import BaseTest
+from ocs_ci.ocs import constants
+
+
+logger = logging.getLogger(__name__)
+
+
+class TestPVCFillup(BaseTest):
```
keesturam, line 14: I'm not sure if we need this test except for the PR validation job, but I don't see a harm in having it either.

keesturam (approved):
LGTM, except for the two minor comments from my side

ocs_ci/templates/workloads/fio/workload_io_fillup.yaml
Avilir: The IO will run until all data was written **or** up to the time is finished, this runtime if high so all data will be written for sure.
If the is error while writing and the test failed, it will not continue this long time.

ebenahar (approved):

Issue #2992, opened 2020-09-23
`test_object_actions` behaves unexpectedly in OCS 4.6


MeridianExplorer:
One of the changes introduced in OCS 4.6 is that [the OBC owner account will have total control over the bucket](https://bugzilla.redhat.com/show_bug.cgi?id=1849771), regardless of any bucket policies.
However, many parts of the test rely on bucket policies blocking the account from performing various actions on the bucket.
This leads to a scenario in which an object is put in the bucket, then the object is deleted (the deletion should fail, but it succeeds because of the 4.6 change), and then an attempt to get it is made (which fails, because the object was deleted).

The logic will need to be rethought, or the test skipped in >4.5.

PR #3018, opened 2020-09-30
Fix for test_object_actions test case for OCS >=4.6

parikshithb:
Fixes: #2992 

### DIFF
--- a/tests/manage/mcg/test_bucket_policy.py
+++ b/tests/manage/mcg/test_bucket_policy.py
@@ -194,9 +194,14 @@ class TestS3BucketPolicy(ManageTest):
         obc = bucket_factory(amount=1, interface='OC')
         obc_obj = OBC(obc[0].name)
 
+        # Creating noobaa account to access bucket belonging to obc account
+        user_name = "noobaa-user" + str(uuid.uuid4().hex)
+        email = user_name + "@mail.com"
+        user = NoobaaAccount(mcg_obj, name=user_name, email=email, buckets=[obc_obj.bucket_name])
+
         # Admin sets policy on obc bucket with obc account principal
         bucket_policy_generated = gen_bucket_policy(
-            user_list=obc_obj.obc_account,
+            user_list=[obc_obj.obc_account, user.email_id],
             actions_list=['PutObject'],
             resources_list=[f'{obc_obj.bucket_name}/{"*"}']
         )
@@ -211,14 +216,24 @@ class TestS3BucketPolicy(ManageTest):
         get_policy = get_bucket_policy(mcg_obj, obc_obj.bucket_name)
         logger.info(f"Got bucket policy: {get_policy['Policy']}")
 
-        # Verifying whether obc account can put object
-        logger.info(f'Adding object on bucket: {obc_obj.bucket_name}')
+        # Verifying whether users can put object
+        logger.info(f'Adding object on bucket: {obc_obj.bucket_name} using user: {obc_obj.obc_account}')
         assert s3_put_object(obc_obj, obc_obj.bucket_name, object_key, data), "Failed: Put Object"
 
+        logger.info(f'Adding object on bucket: {obc_obj.bucket_name} using user: {user.email_id}')
+        assert s3_put_object(user, obc_obj.bucket_name, object_key, data), "Failed: Put Object"
+
         # Verifying whether Get action is not allowed
-        logger.info(f'Verifying whether user: {obc_obj.obc_account} is denied to Get object')
+        logger.info(
+            f'Verifying whether user: '
+            f'{user.email_id if float(config.ENV_DATA["ocs_version"]) >= 4.6 else obc_obj.obc_account}'
+            f' is denied to Get object'
+        )
         try:
-            s3_get_object(obc_obj, obc_obj.bucket_name, object_key)
+            if float(config.ENV_DATA['ocs_version']) >= 4.6:
+                s3_get_object(user, obc_obj.bucket_name, object_key)
+            else:
+                s3_get_object(obc_obj, obc_obj.bucket_name, object_key)
         except boto3exception.ClientError as e:
             logger.info(e.response)
             response = HttpResponseParser(e.response)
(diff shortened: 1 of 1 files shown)


parikshithb:
Verified on 4.6 and 4.5.

tests/manage/mcg/test_bucket_policy.py
```diff
@@ -264,6 +265,11 @@ def test_object_actions(self, mcg_obj, bucket_factory):
         get_policy = get_bucket_policy(mcg_obj, obc_obj.bucket_name)
         logger.info(f"Got bucket policy: {get_policy['Policy']}")
 
+        # Put object using OBC account(as an Admin)
+        if float(config.ENV_DATA['ocs_version']) >= 4.6:
+            logger.info(f'Putting an object on bucket: {obc_obj.bucket_name} using user: {obc_obj.obc_account}')
+            assert s3_put_object(obc_obj, obc_obj.bucket_name, object_key, data), "Failed: Put Object"
```
petr-balogh, line 271: Can you please cut the long lines, for example like:
```suggestion
            logger.info(
                f'Putting an object on bucket: {obc_obj.bucket_name} using user: {obc_obj.obc_account}'
            )
            assert s3_put_object(
                obc_obj, obc_obj.bucket_name, object_key, data
                ), "Failed: Put Object"
```

parikshithb: Ack

tests/manage/mcg/test_bucket_policy.py
```diff
@@ -264,6 +265,16 @@ def test_object_actions(self, mcg_obj, bucket_factory):
         get_policy = get_bucket_policy(mcg_obj, obc_obj.bucket_name)
         logger.info(f"Got bucket policy: {get_policy['Policy']}")
 
+        # Put object using OBC account(as an Admin)
+        if float(config.ENV_DATA['ocs_version']) >= 4.6:
```
MeridianExplorer, line 269: I feel that this fix tailors the test to work with 4.6, rather than fixing the fundamental issues of it.

The current flow is flawed:
https://github.com/red-hat-storage/ocs-ci/pull/3018/files#diff-4b4e7d3ba98ac96bbc8edf7387c285d2R218
In this part, there's an attempt to get the object. The get should be denied. However, if it succeeds, the test continues executing just the same, which shouldn't be the case - the test should fail.

Same here:
https://github.com/red-hat-storage/ocs-ci/pull/3018/files#diff-4b4e7d3ba98ac96bbc8edf7387c285d2R237
The deletion should fail, but it succeeds, and it doesn't fail the test - even though it should.

True - reuploading the object before trying to `get` it solves the problem - but on the other hand, the test simply doesn't represent the true state of things in 4.6. I think it'll be better to rething the flow, or at least add verifications that actions that should fail, have failed successfully.

WDYT, @parikshithb?

MeridianExplorer (changes requested):

tests/manage/mcg/test_bucket_policy.py
parikshithb: > I feel that this fix tailors the test to work with 4.6, rather than fixing the fundamental issues of it.
> 
> The current flow is flawed:
> https://github.com/red-hat-storage/ocs-ci/pull/3018/files#diff-4b4e7d3ba98ac96bbc8edf7387c285d2R218
> In this part, there's an attempt to get the object. The get should be denied. However, if it succeeds, the test continues executing just the same, which shouldn't be the case - the test should fail.
> 
> Same here:
> https://github.com/red-hat-storage/ocs-ci/pull/3018/files#diff-4b4e7d3ba98ac96bbc8edf7387c285d2R237
> The deletion should fail, but it succeeds, and it doesn't fail the test - even though it should.
> 

I am not sure I quite understand, both GET and DELETE(line 218 & 237) actions does fail right( i.e Access denied)? Both the actions are denied to obc.account principal according to the policy set. 

> True - reuploading the object before trying to `get` it solves the problem - but on the other hand, the test simply doesn't represent the true state of things in 4.6. I think it'll be better to rething the flow, or at least add verifications that actions that should fail, have failed successfully.
> 
The new code is just for noobaa s3 access principal (i.e user.email) where I am uploading an object from obc.account(who is the admin) for version > 4.6 and verifying the policy set to user.email. 

There might be better way than re uploading, so can you please explain more about the flow? 
> WDYT, @parikshithb?

parikshithb: My bad, i missed that for obc.account, actions were not actually denied. 

What about this flow: 
1. We provide **user** principal as well along with **obc_obj.obc_account**.

        bucket_policy_generated = gen_bucket_policy(
            user_list=[obc_obj.obc_account, user.email],
            actions_list=['PutObject'],
            resources_list=[f'{obc_obj.bucket_name}/{"*"}']
        )

2. Then try-catch like this for both **get** and **delete** actions: 

        try:
            if float(config.ENV_DATA['ocs_version']) >= 4.6:
                s3_get_object(user, obc_obj.bucket_name, object_key)
            else:
                s3_get_object(obc_obj, obc_obj.bucket_name, object_key)
        except boto3exception.ClientError as e:
            logger.info(e.response)
            response = HttpResponseParser(e.response)
            if response.error['Code'] == 'AccessDenied':
                logger.info('Get Object action has been denied access')
            else:
                raise UnexpectedBehaviour(f"{e.response} received invalid error code {response.error['Code']}")

parikshithb: @MeridianExplorer Changed the flow.

tests/manage/mcg/test_bucket_policy.py
```diff
@@ -226,14 +242,39 @@ def test_object_actions(self, mcg_obj, bucket_factory):
             else:
                 raise UnexpectedBehaviour(f"{e.response} received invalid error code {response.error['Code']}")
```
MeridianExplorer, line 244: else:
                    assert False, "Get object succeeded when it should have failed"

tests/manage/mcg/test_bucket_policy.py
```diff
@@ -226,14 +242,39 @@ def test_object_actions(self, mcg_obj, bucket_factory):
...
+            f'Verifying whether user: '
+            f'{user.email_id if float(config.ENV_DATA["ocs_version"]) >= 4.6 else obc_obj.obc_account}'
+            f'is denied to Delete object'
+        )
         try:
-            s3_delete_object(obc_obj, obc_obj.bucket_name, object_key)
+            if float(config.ENV_DATA['ocs_version']) >= 4.6:
+                s3_delete_object(user, obc_obj.bucket_name, object_key)
+            else:
+                s3_delete_object(obc_obj, obc_obj.bucket_name, object_key)
```
MeridianExplorer, line 277: Same idea with "try-except-else"

tests/manage/mcg/test_bucket_policy.py
```diff
@@ -242,11 +283,6 @@ def test_object_actions(self, mcg_obj, bucket_factory):
             else:
                 raise UnexpectedBehaviour(f"{e.response} received invalid error code {response.error['Code']}")
```
MeridianExplorer, line 285: Same idea with "try-except-else"

parikshithb: ack

tests/manage/mcg/test_bucket_policy.py
parikshithb: ack

MeridianExplorer (approved):
Looks great!

MeridianExplorer:
I'm not sure if a run is already underway, but is it possible to also verify this over 4.6 please?

parikshithb:
> I'm not sure if a run is already underway, but is it possible to also verify this over 4.6 please?
Yup, in about to initiate.

parikshithb:
Verified new commit on both ocs 4.6 and 4.5

shreekarSS (approved):

ebenahar (approved):

PR #3020, opened 2020-09-30
Use new way of installing OCS over LSO based cluster

prsurve:


ocs_ci/templates/ocs-deployment/local-volume-discovery.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: auto-discover-devices
+spec:
+  nodeSelector:
+    nodeSelectorTerms:
+      - matchExpressions:
+          - key: kubernetes.io/hostname
+            operator: In
+            values:
+              - PLACEHOLDER
```
clacroix12, line 12: Add an empty line to the end of these yaml files

clacroix12 (commented):
Just a couple small things but overall LGTM!

ocs_ci/deployment/deployment.py
```diff
@@ -68,6 +68,7 @@
 )
 from ocs_ci.utility.vsphere_nodes import update_ntp_compute_nodes
 from tests import helpers
+from tests.helpers import get_worker_nodes
```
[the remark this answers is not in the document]
prsurve: Done

ocs_ci/templates/ocs-deployment/local-volume-discovery.yaml
prsurve: Done

ocs_ci/deployment/deployment.py
```diff
@@ -1048,40 +1049,94 @@ def setup_local_storage(storageclass):
             raise NotImplementedError(
                 "LSO Deployment for VMDirectPath is not implemented"
             )
+    if (ocp_version >= "4.6") and (ocs_version >= "4.6"):
```
petr-balogh, line 1052: So in case one of OCP or OCS is less than 4.6 we will go with else block right? So this approach is applicable only when both OCP and OCS are >= 4.6 ?

petr-balogh (dismissed):
LGTM

ocs_ci/deployment/deployment.py
prsurve: Yes with lower version then 4.6 we will go with localvolume creation yes the appoach is applicable for 4.6 version only

ebenahar (approved):

ebenahar (dismissed):

ocs_ci/deployment/deployment.py
```diff
@@ -984,7 +985,8 @@ def setup_local_storage(storageclass):
             "Creating optional operators CatalogSource and ImageContentSourcePolicy"
         )
         run_cmd(f"oc create -f {optional_operators_yaml.name}")
-
+        # sleep here to start update machineconfigpool status
+        time.sleep(60)
```
petr-balogh, line 989: I would transform comment above to log message:
logger.info("sleep for 60 sec to start update machineconfigpool status")

prsurve:
Marking it as verified test reported failed because of know issue #3047

petr-balogh (approved):
LGTM just one comment to add log message about sleep. But approving this one as it's not blocking the PR

petr-balogh (approved):

ocs_ci/deployment/deployment.py
prsurve: removed the cherry-picked comment which was causing this

clacroix12 (approved):

Issue #3001, opened 2020-09-25
must-gather is run even when the testcase has passed 


vasukulkarni:
I am seeing in my environment, that must-gather is collected even when the test case has passed

```
:32:30 - MainThread - tests.manage.mcg.test_bucket_creation - INFO - Create duplicate bucket oc-bucket-ef71a5dfd3d449c89ae99ea96c2d5fb8 failed as expected
PASSED                                                                                                                                                                                                                                 [100%]
------------------------------------------------------------------------------------------------------------- live log teardown --------------------------------------------------------------------------------------------------------------
11:32:30 - MainThread - tests.conftest - INFO - Cleaning up bucket oc-bucket-59e227053fd94348b154cef08d7bdb7a
11:32:30 - MainThread - /home/vasu/ocs-ci/ocs_ci/ocs/resources/objectbucket.py - INFO - Deleting bucket: oc-bucket-59e227053fd94348b154cef08d7bdb7a
11:32:30 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/vasu/my-cluster-path/auth/kubeconfig delete obc oc-bucket-59e227053fd94348b154cef08d7bdb7a
11:32:30 - MainThread - /home/vasu/ocs-ci/ocs_ci/ocs/resources/objectbucket.py - INFO - Verifying deletion of oc-bucket-59e227053fd94348b154cef08d7bdb7a
11:32:30 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage --kubeconfig /home/vasu/my-cluster-path/auth/kubeconfig get obc  -n openshift-storage -o yaml
11:32:31 - MainThread - /home/vasu/ocs-ci/ocs_ci/ocs/resources/objectbucket.py - INFO - oc-bucket-59e227053fd94348b154cef08d7bdb7a was deleted successfuly
11:32:31 - MainThread - tests.conftest - INFO - Cleaning up bucket oc-bucket-ef71a5dfd3d449c89ae99ea96c2d5fb8
11:32:31 - MainThread - /home/vasu/ocs-ci/ocs_ci/ocs/resources/objectbucket.py - INFO - Deleting bucket: oc-bucket-ef71a5dfd3d449c89ae99ea96c2d5fb8
[...]

PR #3025, opened 2020-10-01
Optional parameter to gather logs when deployment runs successfully

jimdevops19:
Fixes: https://github.com/red-hat-storage/ocs-ci/issues/3001

We wanted to provide an option to gather logs on successful deployments, since we don't want to cause over waiting after successful deployments.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -102,6 +102,8 @@ to the pytest.
   bugzilla_password = yourPassword
   ```
 * `--collect-logs` - to collect OCS logs for failed test cases.
+* `--collect-logs-on-success-run` - Collect must gather logs at the end of the 
+   execution (also when no failure in the tests)
 * `--io-in-bg` - If passed, IO will be running in the test background.
 * `--io-load` - IOs throughput target percentage. The value should be
    between 0 to 100. If not provided, the default is 30 (30%)
@@ -110,8 +112,6 @@ to the pytest.
 * `--csv-change` - Allows changes in the OCS CSV. For example, usage of custom image,
    like MCG or RHCS. The format should be:
    <pattern_to_replace_from::pattern_to_replace_to>, while '::' is the delimiter
-* `--collect-logs-on-success-run` - Allows to have the option of gathering logs even,
-   when the cluster deployment executed successfully
 
 ## Examples
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -110,6 +110,8 @@ to the pytest.
 * `--csv-change` - Allows changes in the OCS CSV. For example, usage of custom image,
    like MCG or RHCS. The format should be:
    <pattern_to_replace_from::pattern_to_replace_to>, while '::' is the delimiter
+* `--collect-logs-on-success-run` - Allows to have the option of gathering logs even,
+   when the cluster deployment executed successfully
 
 ## Examples
 
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -32,6 +32,7 @@ RUN:
   # This config file disables scale app pods to use OCS workers
   use_ocs_worker_for_scale: False
   load_status: None
+  collect_logs_on_success_run: False
 
 # In this section we are storing all deployment related configuration but not
 # the environment related data as those are defined in ENV_DATA section.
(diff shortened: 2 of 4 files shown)


docs/usage.md
```diff
@@ -110,6 +110,8 @@ to the pytest.
 * `--csv-change` - Allows changes in the OCS CSV. For example, usage of custom image,
    like MCG or RHCS. The format should be:
    <pattern_to_replace_from::pattern_to_replace_to>, while '::' is the delimiter
+* `--collect-logs-on-success-run` - Allows to have the option of gathering logs even,
```
petr-balogh, line 113: Can you move it right after line 104 to have this log collecting parameters together?

+ thinking about change the text to something like - Collect must gather logs at the end of the execution (also when no failure in the tests)

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -32,6 +32,7 @@ RUN:
   # This config file disables scale app pods to use OCS workers
   use_ocs_worker_for_scale: False
   load_status: None
+  collect_logs_on_success_run: False
```
petr-balogh, line 35: I see here:
https://github.com/red-hat-storage/ocs-ci/pull/2880/files#diff-153ccf5916ca230718b5455c349afbcdR126

we have it in reporting section, should we consider it to have it in reporting section as well here instead of RUN section?

jimdevops19: I think yes, I just followed the rule from some other argument that utilizes the logs.
Will move

petr-balogh (approved):

ocs_ci/framework/conf/default_config.yaml
jimdevops19: I am thinking of taking off the `_run` from the argument name before we merge it. @petr-balogh

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -411,7 +418,9 @@ def process_cluster_cli_params(config):
         csv_change = csv_change.split("::")
         ocsci_config.DEPLOYMENT['csv_change_from'] = csv_change[0]
         ocsci_config.DEPLOYMENT['csv_change_to'] = csv_change[1]
-
+    collect_logs_on_success_run = get_cli_param(config, 'collect_logs_on_success_run')
+    if collect_logs_on_success_run:
+        ocsci_config.REPORTING['collect_logs_on_success_run'] = True
```
petr-balogh, line 423: You are missing one empty line behind this, I see travis is complaining about it for pep8 issue ;)

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -92,6 +92,7 @@ REPORTING:
   default_ocs_must_gather_latest_tag: 'latest-4.6'
   gather_on_deploy_failure: true
   gather_on_deploy_success: true
+  collect_logs_on_success_run: False
```
petr-balogh, line 95: Jim I think it's ok to have _run here. But I don't mind to remove _run from here as well.. but if you will remove it from here, we should remove also from CLI arg.  @vavuthu @dahorak @clacroix12 what do you like more? :)

jimdevops19: Felt like it sounded a little unnecessary once we removed it under the REPORTING from the RUN :)
But we can leave it.

petr-balogh (approved):

vasukulkarni (approved):

clacroix12 (approved):

ocs_ci/framework/conf/default_config.yaml
clacroix12: I think removing `-run` and `_run` sounds better but this can be done afterwards as a follow up.

As a related question, I'm wondering if we need two separate CLI args for the log collection. Would it be better to have one like `--collect-logs` that takes a string as an argument like `failure|success|both` in order to configure when we want to run must gather? We could keep the separate config variables that would be manipulated by the one CLI argument.

Full examples:

```
--collect-logs failure
--collect-logs success
--collect-logs both
```

WDYT?

jimdevops19: That sounds like a great solution for this

PR #3032, opened 2020-10-01
Skip namespace tests in case AWS credentials are unavailable

MeridianExplorer:
Right now, all the namespace tests fail in case AWS credentials aren't provided.
Since all NS tests rely on AWS credentials, we should skip them in those cases.

### DIFF
--- a/tests/manage/mcg/test_namespace.py
+++ b/tests/manage/mcg/test_namespace.py
@@ -6,10 +6,12 @@ from ocs_ci.ocs.bucket_utils import (
     verify_s3_object_integrity
 )
 from ocs_ci.framework import config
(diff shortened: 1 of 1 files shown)


fbalak (approved):
Looks good. Can you please also remove following lines from namespace tests:
```
    # Test is skipped for other platforms due to
    # https://github.com/red-hat-storage/ocs-ci/issues/2774
    @aws_platform_required
```
I wrongly understood the problem and these lines are not necessary.

MeridianExplorer:
@fbalak - IMO, the marker is still appropriate.
True - the tests _should_ run independent of platform, *but*, as it currently stands, the OCS-CI configs of other platforms collide with those tests, and lead to failures.
Thus, for now, I think it's actually appropriate to keep it in :)

petr-balogh (approved):

Issue #3047, opened 2020-10-04
Deployment failed if unchecking "collect logs"


AviadP:
if ocs:
            latest_tag = ocsci_config.REPORTING.get(
                'ocs_must_gather_latest_tag',
                ocsci_config.REPORTING.get(
                    'default_ocs_must_gather_latest_tag', ocsci_config.DEPLOYMENT[
                        'default_latest_tag'
                    ]
                )
            )
>           ocs_log_dir_path = os.path.join(log_dir_path, 'ocs_must_gather')
E           UnboundLocalError: local variable 'log_dir_path' referenced before assignment

PR #3048, opened 2020-10-04
Preventing call of collect ocs logs when status_failure=False is sent

jimdevops19:
Fixes: #3047
Follow-up discussion for this PR: #3056

In [pull-3025](https://github.com/red-hat-storage/ocs-ci/pull/3025), we were able to accept new ocs-ci parameter `collect_logs_on_success_run`, but the actions of it did not cover all the possible cases.

Besides checking where  `collect_ocs_logs` is being called, I assigned if statement before the variable creation `log_dir_path`, so when we deployed without sending `--collect-ocs-logs`, we received `Local variable is called without any assignment`.

This is reverted, and then this merge calls the `collect_ocs_logs` with status_failure=False, **only** if `--collect-logs-on-success-run` is sent on run-ci

### DIFF
--- a/.functional_ci_setup.py
+++ b/.functional_ci_setup.py
@@ -97,7 +97,6 @@ def get_ocsci_conf():
         ),
         REPORTING=dict(
             gather_on_deploy_failure=True,
-            gather_on_deploy_success=True,
         )
     )
     if env.get("DOWNSTREAM") == "true":
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -137,7 +137,7 @@ class Deployment(object):
         if not config.ENV_DATA['skip_ocs_deployment']:
             try:
                 self.deploy_ocs()
-                if config.REPORTING['gather_on_deploy_success']:
+                if config.REPORTING['collect_logs_on_success_run']:
                     collect_ocs_logs('deployment', ocp=False, status_failure=False)
             except Exception as e:
                 logger.error(e)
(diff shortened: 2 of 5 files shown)


AviadP (dismissed):

annagitel (approved):
LGTM

jimdevops19:
Part of my edit in the latest **ocs-ci** comment, [this case](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/13028/) verifies that whenever deploy succeeds, and `--collect-ocs-logs-in-success` is not passed through `run-ci`, we don't gather logs.

petr-balogh:
@jimdevops19 please open the issue where you will put description of the issue which you are trying to fix.  I see you commented in first comment but we should have issue for that opened and here have Fixes: @numberOfssue so we can have discussion on the issue itself as well.

petr-balogh:
As we discussed offline, please remove all occurrences of : 

gather_on_deploy_failure
gather_on_deploy_success

And use only your new CLI parameter for that

petr-balogh (approved):

petr-balogh:
> As we discussed offline, please remove all occurrences of :
> 
> gather_on_deploy_failure
> gather_on_deploy_success
> 
> And use only your new CLI parameter for that

We decided to remove only gather_on_deploy_success and replace it with new CLI param.

The logic behind the gather_on_deploy_failure will be discussed on next call with engineering as this was added by Zack.

So intention is to remove it and use only CLI options we have.  @jimdevops19 please add this topic to the agenda and open the new issue fore this you can refer there ;) Thanks

annagitel (approved):

PR #3063, opened 2020-10-06
NooBaa endpoint autoscale infra

nb-ohad:
- A fixture to ensures an endpoint count in the system
- A base test for all MCG tests that enforce a specific endpoint count before every test (using fixture from the last bullet)
- Remove current code that set up the number of endpoints when deploying


tests/conftest.py
```diff
@@ -2910,3 +2909,63 @@ def finalizer():
...
+            storage_cluster.patch(resource_name=resource_name, params=params, format_type='merge')
+            should_wait = True
+
+        if resource['spec']['multiCloudGateway']['endpoints']['maxCount'] != max_ep_count:
+            log.info(f"Changing maximum Noobaa endpoints to {max_ep_count}")
+            params = f'{{"spec":{{"multiCloudGateway":{{"endpoints":{{"maxCount":{max_ep_count}}}}}}}}}'
+            storage_cluster.patch(resource_name=resource_name, params=params, format_type='merge')
+            should_wait = True
+
+    if should_wait:
```
ebenahar, line 2961: we have discussed this offline and I understand why we don't want to poll the number of EPs and continue when the number reaches the desired one (because it can change later while we have continued, which is False positive situation).
With that, I would suggest that instead of just waiting for 5 minutes, we will do 10 iterations of validating the number of EPs, with 0.5 minute sleep in between. This way, logging will not get hanged for 5 minutes and we will have some indication of the change.

tests/conftest.py
```diff
@@ -2910,3 +2909,63 @@ def finalizer():
...
+        time.sleep(5 * 60) # 5 min
+
+    pods_info = get_pods_having_label(label=constants.NOOBAA_ENDPOINT_POD_LABEL, namespace=namespace):
+    ready_count = 0;
```
ebenahar, line 2965: remove `;`

tests/conftest.py
```diff
@@ -2910,3 +2909,63 @@ def finalizer():
...
+    for ep_info in pods_info:
+        if ep_info['pod_info']['containerStatuses'][0]['ready'] == 'true':
```
ebenahar, line 2967: is there only 1 container always?

nb-ohad: For the endpoint pod, yes. If you have another suggestion I can change

tests/conftest.py
```diff
@@ -2910,3 +2910,77 @@ def finalizer():
...
+            log.info(f"Changing maximum Noobaa endpoints to {max_ep_count}")
+            params = f'{{"spec":{{"multiCloudGateway":{{"endpoints":{{"maxCount":{max_ep_count}}}}}}}}}'
+            storage_cluster.patch(resource_name=resource_name, params=params, format_type='merge')
+            should_wait = True
+
+    if should_wait:
+        # loggin the ready endpoint count for 5 min in 30sec internvals
+        for i in range(10):
+            ready_count = get_ready_noobaa_endpoint_count(namespace)
+            log.info(f"[{i * 30}]: Waiting for noobaa endpoints to stabilize, current ready count: {ready_count}")
```
ebenahar, line 2981: what is `[{i * 30}]`?

nb-ohad: I am just printing the number of seconds passed since the start of the check for tracing purposes. 
I find it very helpful while diagnosing logs, it makes it easy to get a sense of how much time we really need if we decide to fine-tune the timeout in the future.

ebenahar: ok. Maybe add something like `Time passed - {i * 30}...`

nb-ohad: The linter is limiting me to 120 chars. I already minimized most of the sentence

ebenahar: You can do:

```python
log.info(
    f"Time passed - {i * 30} waiting for noobaa endpoints to stabilize, current ready "
    f"count: {ready_count}"
)
```

nb-ohad: Done, thanks

ebenahar (approved):

tests/conftest.py
```diff
@@ -2924,3 +2924,80 @@ def finalizer():
                 collect_ocs_logs('testcases', ocp=False, status_failure=False)
 
     request.addfinalizer(finalizer)
+
+
+def get_ready_noobaa_endpoint_count(namespace):
```
fbalak, line 2929: It would be better to not have functions in conftest.py. It can be in `ocs_ci/ocs/resources/mcg.py`.

fbalak (approved):

tests/conftest.py
nb-ohad: @fbalak This is not a general-purpose util function, it is there just because I hade to use the same logic in 2 diff places in the fixture.  Also, we already have some utils functions in the top of  `conftest.py`

fbalak: ok

PR #3077, opened 2020-10-07
Adding marker for skip depend on OCP version

Avilir:
Since there are some features, like snapshot which cannot be run on all OCP versions, we need this functionality 

using it by adding the marker : @skipif_ocp_version(<expression>) 
for example : @skipif_ocp_version('<4.6') 


Avilir:
verified locally , example output :
```
----------------------------- live log collection ------------------------------
12:41:24 - MainThread - tests.conftest - INFO - Test: <Function test_pvc_snapshot_performance[RBD-1]> will be skipped due to OCP ('<=4.6',)
 ```

petr-balogh (approved):

dahorak (approved):

PR #3082, opened 2020-10-08
Couchbase: export pillowfight pods log info to google sheet

shreekarSS:


Avilir (approved):
LGTM

tests/conftest.py
```diff
@@ -2436,6 +2436,9 @@ def factory(replicas=3, run_in_bg=False, skip_analyze=False):
         couchbase.run_workload(replicas=replicas, run_in_bg=run_in_bg)
         # Run sanity check on data logs
         couchbase.analyze_run(skip_analyze=skip_analyze)
+        # Export data to the google sheet
+        couchbase.export_pfoutput_to_googlesheet(sheet_name='E2E Workloads', sheet_index=2)
```
PrasadDesala, line 2440: Can we exclude importing the data to gsheet from this fixture?
This fixture can be called for any test(including failure tests) and IMO, there is no need to capture this data for failure tests.

tests/e2e/workloads/couchbase/test_couchbase.py
```diff
@@ -2,33 +2,18 @@
...
-    def test_cb_workload_simple(self, couchbase):
+    def test_cb_workload_simple(self, couchbase_factory_fixture):
         """
         Testing basic couchbase workload
         """
-        couchbase.setup_cb()
-        couchbase.create_couchbase_worker(replicas=5)
-        couchbase.run_workload(replicas=5)
-        couchbase.analyze_run()
+        couchbase_factory_fixture(replicas=3)
```
PrasadDesala, line 19: Please export data to the google sheet in this basic CB test.

vasukulkarni (dismissed):

tests/conftest.py
vasukulkarni: @PrasadDesala what index sheet is that?

vasukulkarni: Ignore I got it, If we can also post, pass/fail probably it would be easier to delete those failed runs.

shreekarSS: if the test is failed, I don't think so code will reach till export to the sheet,
a teardown would happen immediately

tests/e2e/workloads/couchbase/test_couchbase.py
shreekarSS: couchbase_factory_fixture() is handling export data as well

tests/conftest.py
PrasadDesala: @vasukulkarni the failure tests I am referring to here are tier-4 tests.

tests/e2e/workloads/couchbase/test_couchbase.py
PrasadDesala: Yes, this was my continuation comment to above :)
Remove the gsheet export from the fixture and handle it in this test case.

wusui (approved):
Looks good.

PrasadDesala (approved):

vasukulkarni (approved):

Issue #3098, opened 2020-10-09
Vmware LSO Installation fails to create PV's


prsurve:
In our current of VMware dc-cp, we don't have datastore with SSD so localvolumeset doesn't detect NonRotational disk as a workaround we have to add `Rotational` for localvolumediscovery

PR #3099, opened 2020-10-09
Use Rotational to detect HDD disk for LocalVolumeSet for Vmware LSO deployment

prsurve:
Fixes: #3098 

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -1091,6 +1091,11 @@ def setup_local_storage(storageclass):
             constants.LOCAL_VOLUME_SET_YAML
         )
 
+        # as a workaround we are setting Rotational to deviceMechanicalProperties to detect HDD disk
+        if platform == constants.VSPHERE_PLATFORM:
+            logger.info("Adding Rotational for deviceMechanicalProperties spec")
+            lvd_data['spec']['deviceInclusionSpec']['deviceMechanicalProperties'].append("Rotational")
+
         # Update local volume set data with Worker node Names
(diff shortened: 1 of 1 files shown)


wusui (dismissed):
Code looks good.  You may want to expand on the comment.  Something lie:

Since we don't have datastore with SSD on our current VMware machines, localvolumeset doesn't detect NonRotational disk.  As a workaround we are setting Rotational to device MechanicalProperties to detect HDD disk

ocs_ci/deployment/deployment.py
```diff
@@ -1090,6 +1090,11 @@ def setup_local_storage(storageclass):
             constants.LOCAL_VOLUME_SET_YAML
         )
 
+        # as a workaround we are setting Rotational to deviceMechanicalProperties to detect HDD disk
```
ebenahar, line 1093: do we have an issue to track the W/A?
also, do we need to set this value everytime?

prsurve: i have created  an issue here #3098, yes need to set that value of all the deployment of VMware LSO

prsurve:
> Code looks good. You may want to expand on the comment. Something lie:
> 
> Since we don't have datastore with SSD on our current VMware machines, localvolumeset doesn't detect NonRotational disk. As a workaround we are setting Rotational to device MechanicalProperties to detect HDD disk

Done

ramkiperiy (approved):

sidhant-agrawal (approved):

ebenahar (approved):

PR #3116, opened 2020-10-12
NooBaa endpoint autoscale tests

nb-ohad:
The test will do the following:
1. Setup the cluster with a 1-2 endpoint autoscale configuration
2. Assert that the cluster is only running 1 endpoint pod
2. Start a Job that will load the NooBaa s3 service
3. Wait and  assert that the cluster is only running 2 endpoint pods
4. Stop the Job
5. Wait and  assert that the cluster is only running 1 endpoint pod 

Signed-off-by: nb-ohad <mitrani.ohad@gmail.com>

### DIFF
--- a/tests/manage/mcg/test_endpoint_autoscale.py
+++ b/tests/manage/mcg/test_endpoint_autoscale.py
@@ -1,0 +1,41 @@
+import pytest
+
+from ocs_ci.framework.testlib import MCGTest, tier1, skipif_ocs_version
+from ocs_ci.ocs import constants, defaults, ocp
+
+
+@pytest.mark.polarion_id("OCS-XXXX")
+@tier1
+class TestEndpointAutoScale(MCGTest):
+    """
+    Test MCG endpoint auto-scaling
+
+    """
+
+    # This will ensure the test will start
+    # with an autoscaling conifguration of 1-2
+    MIN_ENDPOINT_COUNT = 1
+    MAX_ENDPOINT_COUNT = 2
+
+    @skipif_ocs_version('<4.5')
+    def test_scaling_under_load(self, mcg_job_factory):
+        self._assert_endpoint_count(1)
+
+        options = {}
+        job = mcg_job_factory(custom_options=options)
+        self._assert_endpoint_count(2)
+
+        job.delete()
+        job.ocp.wait_for_delete(resource_name=job.name, timeout=60)
+        self._assert_endpoint_count(1)
+
+    def _assert_endpoint_count(self, desired_count):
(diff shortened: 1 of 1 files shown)


nb-ohad:
The reason this PR is marked as WIP is that I still need to figure up what should be the proper fio options that will yield the desired results (the `options` dict) 

The logic will not change so this PR can be reviewed

tests/manage/mcg/test_endpoint_autoscale.py
```diff
@@ -0,0 +1,41 @@
+import pytest
+
+from ocs_ci.framework.testlib import MCGTest, tier1, skipif_ocs_version
+from ocs_ci.ocs import constants, defaults, ocp
+
+
+@pytest.mark.polarion_id("OCS-XXXX")
```
ebenahar, line 7: I will add polarion ids later. Will track it with https://github.com/red-hat-storage/ocs-ci/issues/3121

tests/manage/mcg/test_endpoint_autoscale.py
```diff
@@ -0,0 +1,41 @@
...
+        self._assert_endpoint_count(1)
+
+    def _assert_endpoint_count(self, desired_count):
+        pod = ocp.OCP(kind=constants.POD, namespace=defaults.ROOK_CLUSTER_NAMESPACE)
+
+        assert pod.wait_for_resource(
+            resource_count=desired_count,
+            condition=constants.STATUS_RUNNING,
+            selector=constants.NOOBAA_ENDPOINT_POD_LABEL,
+            dont_allow_other_resources=True,
```
ebenahar, line 39: Just to make sure - this will ensure only the desired number of pods with the given label selector, right?

nb-ohad: To the best of my understanding, (because it wasn't me that added it) this is the exact intention behind the flag

ebenahar (dismissed):

tests/manage/mcg/test_endpoint_autoscale.py
petr-balogh, line 7: Can you please comment this line as I think it can break polarion importer when this polarion id OCS-XXXX does not exists.

tiffanyn108:
Test case looks good.  However, when executing on OCS 4.6 + OCP 4.6, test case fails due the only 1 noobaa-endpoint is running but it is expecting 2 noobaa-endpoint.  Does FIO generating enough enough load to trigger autoscale?

ebenahar (approved):
I suggest we merge this PR even if the load we put on the cluster doesn't trigger the auto scaling

tiffanyn108 (approved):
Test case looks good.  We will need to update it with option values when we have it finalized.

PR #2985, opened 2020-09-22
Use Prometheus user for alerting tests

fbalak:
This should allow alerting tests to work on setups without kubeadmin-password file (IBM cloud).

PR #1219, opened 2019-12-24
Added support for logging in as sa and using it in the end of relevant tests

gklein:
Signed-off-by: Gil Klein <gklein@users.noreply.github.com>

PR #3124, opened 2020-10-13
RGW monitoring test

fbalak:

### DIFF
--- a/tests/manage/monitoring/conftest.py
+++ b/tests/manage/monitoring/conftest.py
@@ -712,3 +712,58 @@ def workload_idle(measurement_dir):
     test_file = os.path.join(measurement_dir, 'measure_workload_idle.json')
     measured_op = measure_operation(do_nothing, test_file)
     return measured_op
+
+
+@pytest.fixture
+def measure_stop_rgw(measurement_dir, request, rgw_deployments):
+    """
+    Downscales RGW deployments, measures the time when it was
+    downscaled and monitors alerts that were triggered during this event.
+
+    Returns:
+        dict: Contains information about `start` and `stop` time for stopping
+            RGW pods
+
+    """
+    oc = ocp.OCP(
+        kind=constants.DEPLOYMENT,
+        namespace=config.ENV_DATA['cluster_namespace']
+    )
+
+    def stop_rgw():
+        """
+        Downscale RGW interface deployments for 5 minutes.
+
+        Returns:
+            str: Name of downscaled deployment
+
+        """
+        # run_time of operation
+        run_time = 60 * 5
+        nonlocal oc
+        nonlocal rgw_deployments
+        for rgw_deployment in rgw_deployments:
+            rgw = rgw_deployment['metadata']['name']
+            logger.info(f"Downscaling deployment {rgw} to 0")
+            oc.exec_oc_cmd(f"scale --replicas=0 deployment/{rgw}")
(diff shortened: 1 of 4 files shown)


tests/manage/monitoring/conftest.py
```diff
@@ -712,3 +712,51 @@ def do_nothing():
...
+        # run_time of operation
+        run_time = 60 * 5
+        nonlocal oc
+        nonlocal rgw_deployments
+        for rgw_deployment in rgw_deployments:
+            rgw = rgw_deployment['metadata']['name']
+            logger.info(f"Downscaling deployment {rgw} to 0")
+            oc.exec_oc_cmd(f"scale --replicas=0 deployment/{rgw}")
+        logger.info(f"Waiting for {run_time} seconds")
+        time.sleep(run_time)
```
ebenahar, line 750: how about adding logging to happen every few seconds, so there will be some indication of what's going on with the RGW deployment?

tests/manage/monitoring/conftest.py
```diff
@@ -712,3 +712,51 @@ def do_nothing():
...
+        dict: Contains information about `start` and `stop` time for stopping
+            RGW pods
+
+    """
+    oc = ocp.OCP(
+        kind=constants.DEPLOYMENT,
+        namespace=config.ENV_DATA['cluster_namespace']
+    )
+
+    def stop_rgw():
```
ebenahar, line 733: won't this function be usable by other things? maybe we can take it out to be somewhere under `ocs_ci/`?

tests/manage/monitoring/prometheus/test_rgw.py
```diff
@@ -0,0 +1,48 @@
...
+        states=states,
+        severity='error'
+    )
+    api.check_alert_cleared(
+        label=target_label,
+        measure_end_time=measure_stop_rgw.get('stop')
+    )
+
+
+def teardown_module():
```
ebenahar, line 46: where is it being used?

tests/manage/monitoring/conftest.py
fbalak: This function just downscales rgw deployments and waits 5 minutes. This seems like a very specific scenario for monitoring. I would keep it here until similar functionality is needed in other tests.

tests/manage/monitoring/conftest.py
fbalak: During this time is running other [thread which logs coming alerts](https://github.com/red-hat-storage/ocs-ci/blob/de5d0abb024121dc82c28be26ce741ba94fb2f2b/ocs_ci/utility/workloadfixture.py#L65) and prints them every few seconds.

tests/manage/monitoring/prometheus/test_rgw.py
fbalak: This is a teardown function introduced in https://github.com/red-hat-storage/ocs-ci/pull/1219 to make sure that correct user is used after monitoring tests.
This will not be needed after https://github.com/red-hat-storage/ocs-ci/pull/2985 will be merged. I will investigate with Petr and IBM guys why this is not working correctly on IBM cloud this week.

ebenahar: ok, thanks for pointing this out

ebenahar (approved):

mbukatov (approved):

Issue #3125, opened 2020-10-13
Create configuration for OCS 4.7


petr-balogh:
We need to prepare for OCS 4.7 as engineering already setting up the pipeline.

PR #3130, opened 2020-10-13
Added config file for OCS 4.7

clacroix12:
Fixes: #3125 

### DIFF
--- a/conf/ocs_version/ocs-4.7.yaml
+++ b/conf/ocs_version/ocs-4.7.yaml
@@ -1,0 +1,10 @@
+---
+DEPLOYMENT:
+  default_ocs_registry_image: "quay.io/rhceph-dev/ocs-registry:latest-4.7"
+  default_latest_tag: 'latest-4.7'
+  ocs_csv_channel: "stable-4.7"
+  default_ocp_version: '4.7'
+ENV_DATA:
+  ocs_version: '4.7'
+REPORTING:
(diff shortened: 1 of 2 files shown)


conf/ocs_version/ocs-4.7.yaml
```diff
@@ -0,0 +1,10 @@
+---
+DEPLOYMENT:
+  default_ocs_registry_image: "quay.io/rhceph-dev/ocs-registry:latest-stable-4.7"
+  default_latest_tag: 'latest-stable-4.7'
```
petr-balogh, line 4: I would go with latest-4.7 for now, it can make some time till the build will pass the acceptance and we can change it once we have it

petr-balogh:
ocs_ci/framework/main.py look at this file and add 4.7 option for ocs-version parameter

conf/ocs_version/ocs-4.7.yaml
clacroix12: Would I need to change the `ocs_csv_channel` as well if we aren't going with latest-stable or does that remain `stable-4.7`?

petr-balogh: csv channel is always in that format stable-X.Y ;) so no need to change.

petr-balogh (approved):

vavuthu (approved):

PR #3131, opened 2020-10-13
Nodereplacement proactive vsphere lso

yitzhak12:


prsurve:
Also, cover for BareMetal LSO

ocs_ci/ocs/node.py
```diff
@@ -904,6 +904,53 @@ def delete_and_create_osd_node_vsphere_upi(
     return new_node_name
 
 
+def delete_and_create_osd_node_vsphere_lso(
```
PrasadDesala, line 907: Can we also perform ceph side verification post node replacement?
Like replaced host should not list from ceph osd, the rack labels are added correctly etc if any

ocs_ci/ocs/node.py
```diff
@@ -1027,3 +1074,27 @@ def node_replacement_verification_steps_ceph_side(old_node_name, new_node_name):
 
     log.info("Verification steps from the ceph side finish successfully")
     return True
+
+
+def scale_down_deployments(node_obj):
```
ebenahar, line 1079: please add docstring

ocs_ci/ocs/node.py
```diff
@@ -1027,3 +1074,27 @@ def node_replacement_verification_steps_ceph_side(old_node_name, new_node_name):
...
+    ocp = OCP(kind='node', namespace=defaults.ROOK_CLUSTER_NAMESPACE)
+    pods_to_scale_down = get_node_pods_to_scale_down(node_obj)
+    for p in pods_to_scale_down:
+        deployment_name = pod.get_deployment_name(p.name)
+        cmd = f"scale deployment {deployment_name} --replicas=0"
+        ocp.exec_oc_cmd(cmd)
+
+
+def get_node_pods_to_scale_down(node):
```
ebenahar, line 1088: please add docstring

ocs_ci/ocs/node.py
yitzhak12: Yes. Do you mean to perform ceph side verification post node replacement with **LSO**?

PrasadDesala: Yes

ocs_ci/ocs/node.py
```diff
@@ -908,6 +914,105 @@ def delete_and_create_osd_node_vsphere_upi(osd_node_name, use_existing_node=Fals
...
+        use_existing_node (bool): If False, create a new node and label it.
+            If True, use an existing node to replace the deleted node
+            and label it.
+
+    Returns:
+        str: The new node name
+
+    """
+    from ocs_ci.ocs.platform_nodes import PlatformNodesFactory
+    from ocs_ci.ocs.resources.storage_cluster import get_osd_size
```
ebenahar, line 935: can these be moved to the top of the file?

ocs_ci/ocs/node.py
```diff
@@ -908,6 +914,105 @@ def delete_and_create_osd_node_vsphere_upi(osd_node_name, use_existing_node=Fals
     return new_node_name
 
 
+def delete_and_create_osd_node_vsphere_lso(osd_node_name, use_existing_node=False):
```
ebenahar, line 917: can this function changed to be more generic, to delete and create osd on any platform and for both LSO and dynamically provisioned clusters?

ocs_ci/ocs/node.py
yitzhak12: When I tried to put it at the top of the file, it caused a circular loop.

ocs_ci/ocs/node.py
yitzhak12: The steps are different. In the LSO node, we need to perform additional steps. Maybe we can move it to a class like the `platform_nodes` file, but it will take some time. Anyway, if you have any idea how to make it more generic please let me know.

ebenahar: adding it to platform_nodes is an option, although, platform_nodes should be platform oriented and not OCP oriented. 

Another option would be to create a new class structure for OCP node operations, which can include node replacement, node tain, etc. This new class structure can will have child classed for each platform / deployment method combination. Something like:

yitzhak12: I think it will be better to merge this PR and then implement a new PR that will handle the new functionality. 
I created a new issue for that https://github.com/red-hat-storage/ocs-ci/issues/3648.

tests/manage/z_cluster/nodes/test_node_replacement_proactive.py
```diff
@@ -46,6 +46,9 @@ def delete_and_create_osd_node(osd_node_name):
```
hnallurv, line 46: does this test run with IOs in background during node replacement? If not, please add IOs

yitzhak12: We have 2 tests in this file: 
1. https://github.com/red-hat-storage/ocs-ci/blob/master/tests/manage/z_cluster/nodes/test_node_replacement_proactive.py#L100
2. https://github.com/red-hat-storage/ocs-ci/blob/master/tests/manage/z_cluster/nodes/test_node_replacement_proactive.py#L165

The first one is with IO in the background and the second one without IO in the background. 
Currently, with AWS UPI and vSphere, I ran it without IO cause the rebalancing took too long. But I added the param `--io-in-bg --io-load 30` in the test.

ocs_ci/ocs/node.py
```diff
@@ -908,6 +914,106 @@ def delete_and_create_osd_node_vsphere_upi(osd_node_name, use_existing_node=Fals
...
+    """
+    Unschedule, drain and delete osd node, and creating a new osd node.
+    At the end of the function there should be the same number of osd nodes as
+    it was in the beginning, and also ceph health should be OK.
+    This function is for vSphere UPI.
+
+    Args:
+        osd_node_name (str): the name of the osd node
+        use_existing_node (bool): If False, create a new node and label it.
+            If True, use an existing node to replace the deleted node
```
hnallurv, line 927: Where is use_existing_node=True covered for LSO? Is it not required?

yitzhak12: We have an issue to create a new worker node on vSphere from the cli before, so I used an existing worker node. Now Vijay fixed the issue, so I can remove this option if it's not necessary.

ocs_ci/ocs/node.py
```diff
@@ -908,6 +914,106 @@ def delete_and_create_osd_node_vsphere_upi(osd_node_name, use_existing_node=Fals
...
+        node_not_in_ocs = get_worker_nodes_not_in_ocs()[0]
+        log.info(
+            f"Preparing to replace the node {osd_node_name} "
+            f"with an existing node {node_not_in_ocs.name}"
+        )
+        if node_type == constants.RHEL_OS:
+            set_selinux_permissions(workers=[node_not_in_ocs])
+        label_nodes([node_not_in_ocs])
+        new_node_name = node_not_in_ocs.name
+
```
prsurve, line 970: This code can me replaced with `delete_and_create_osd_node_vsphere_upi` because it's doing same task

ocs_ci/ocs/node.py
```diff
@@ -908,6 +914,106 @@ def delete_and_create_osd_node_vsphere_upi(osd_node_name, use_existing_node=Fals
...
+        f"Create a new disk with size {osd_size}, and attach to node {new_node_name}"
+    )
+    node_util.create_and_attach_volume(node=new_node, size=osd_size)
+
+    log.info(
+        "Replace the old node with the new worker node in localVolumeDiscovery and localVolumeSet"
+    )
+    res = add_new_node_to_lvd_and_lvs(
+        old_node_name=osd_node_name, new_node_name=new_node_name
+    )
```
prsurve, line 986: here new_node_name should be named from label value `kubernetes.io/hostname`

ocs_ci/ocs/resources/pod.py
```diff
@@ -1664,3 +1669,133 @@ def get_osd_removal_pod_name(osd_id, timeout=60):
...
+        osd_id (str): The osd id
+
+    Returns:
+        bool: True, if the ocs-osd-removal job completed successfully. False, otherwise
+
+    """
+    logger.info("Getting the ocs-osd-removal pod name")
+    osd_removal_pod_name = get_osd_removal_pod_name(osd_id)
+    osd_removal_pod_obj = get_pod_obj(
+        osd_removal_pod_name, namespace="openshift-storage"
```
prsurve, line 1714: replace with default.ROOK_CLUSTER_NAMESPACE

ocs_ci/ocs/resources/pod.py
```diff
@@ -1664,3 +1669,133 @@ def get_osd_removal_pod_name(osd_id, timeout=60):
...
+    try:
+        osd_removal_job.ocp.wait_for_delete(resource_name=f"ocs-osd-removal-{osd_id}")
+    except TimeoutError:
+        logger.warning(f"ocs-osd-removal-{osd_id} job did not deleted successfully")
+        return False
+
+    return True
+
+
+def get_mon_deployments(osd_label=constants.OSD_APP_LABEL, namespace=None):
```
prsurve, line 1757: dont it should be get_osd_deployment

ocs_ci/ocs/node.py
yitzhak12: Yes, you right. Generally, I created this issue https://github.com/red-hat-storage/ocs-ci/issues/3648 so that we have a function for every platform and deployment, and we don't need to check which function to use every time.

ocs_ci/ocs/resources/pod.py
yitzhak12: Actually, I don't need this function for this PR. I will delete it.

ocs_ci/ocs/node.py
yitzhak12: We get the node name from here https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/node.py#L966. 
`node.get("metadata").get("name")`. So you think we need to change it here when passing to the function `add_new_node_to_lvd_and_lvs` to get the node name from `kubernetes.io/hostname` label?

ocs_ci/ocs/node.py
ebenahar: ack

tests/manage/z_cluster/nodes/test_node_replacement_proactive.py
ebenahar: will be done as part of https://projects.engineering.redhat.com/browse/OCSQE-578

ocs_ci/ocs/node.py
yitzhak12: Fixed

ocs_ci/ocs/node.py
yitzhak12: Fixed

ocs_ci/ocs/resources/pod.py
yitzhak12: Fixed

ocs_ci/ocs/node.py
```diff
@@ -908,6 +914,91 @@ def delete_and_create_osd_node_vsphere_upi(osd_node_name, use_existing_node=Fals
...
+
+    switch_to_project(defaults.ROOK_CLUSTER_NAMESPACE)
+
+    osd_removal_job = pod.run_osd_removal_job(osd_id)
+    assert osd_removal_job, "ocs-osd-removal failed to create"
+    is_completed = (pod.verify_osd_removal_job_completed_successfully(osd_id),)
+    assert is_completed, "ocs-osd-removal-job is not in status 'completed'"
+    log.info("ocs-osd-removal-job completed successfully")
+
+    delete_released_pvs_in_sc(sc_name)
```
PrasadDesala, line 987: may be a good to throw an assert here if this operation fails otherwise the test continues and we may not notice it

ocs_ci/ocs/node.py
```diff
@@ -908,6 +914,91 @@ def delete_and_create_osd_node_vsphere_upi(osd_node_name, use_existing_node=Fals
...
+
+    # Delete the crashcollector pod deployment
```
PrasadDesala, line 989: Does this step is required? I see above we are scaling down the crashcollector pod deployment as well

ocs_ci/ocs/node.py
```diff
@@ -1016,6 +1110,19 @@ def node_replacement_verification_steps_user_side(old_node_name, new_node_name):
         log.warning("Not all the pods in running state")
         return False
 
+    new_osd_pod = get_node_pods(new_node_name, pods_to_search=pod.get_osd_pods())[0]
+    if not new_osd_pod:
+        log.warning("Didn't find any osd pods running on the new node")
+        return False
```
PrasadDesala, line 1116: do you think we should throw exceptions in this function for failures instead of returning boolean and remove the assert in the test case?

ocs_ci/ocs/node.py
yitzhak12: Yes, I will add an assert

ocs_ci/ocs/node.py
yitzhak12: Yes, this is true. But from the docs, I see that we need to perform this step too.

prsurve: i guess we dont need to delete crashcollector pods

ocs_ci/ocs/node.py
```diff
@@ -908,6 +914,93 @@ def delete_and_create_osd_node_vsphere_upi(osd_node_name, use_existing_node=Fals
     return new_node_name
 
 
+def delete_and_create_osd_node_vsphere_lso(osd_node_name, use_existing_node=False):
```
ebenahar, line 917: this function is specific to vsphere UPI. Better to have its name reflecting that

ocs_ci/ocs/node.py
```diff
@@ -908,6 +914,93 @@ def delete_and_create_osd_node_vsphere_upi(osd_node_name, use_existing_node=Fals
...
+        new_node_name=new_node_hostname_label,
+    )
+    assert res, "Failed to add the new node to LVD and LVS"
+
+    log.info("Verify new pv is available...")
+    is_new_pv_available = verify_new_pv_available_in_sc(old_pv_objs, sc_name)
+    assert is_new_pv_available, "New pv is not available"
+    log.info("Finished verifying that the new pv is available")
+
+    switch_to_project(defaults.ROOK_CLUSTER_NAMESPACE)
```
ebenahar, line 982: why is this needed?
Generally, I think that switching to a project, rather than specifying the namespace in the `oc` commands, is quite risky, because the test can fail before the switch back to the previous project took place, which means that the rest of the execution will be while the project is wrong

ocs_ci/ocs/pv.py
```diff
@@ -0,0 +1,135 @@
+import logging
```
ebenahar, line 1: I think that the file should reside under ocs_ci/ocs/resources, to keep consistent with other resources

ocs_ci/ocs/node.py
yitzhak12: Yes, I get your point. This step is mentioned in the docs: `oc project openshift-storage`, 
so I am not sure if we can remove it. @PrasadDesala @prsurve, what do you think?

ocs_ci/ocs/node.py
yitzhak12: Fixed

ocs_ci/ocs/pv.py
yitzhak12: Fixed

ocs_ci/ocs/node.py
prsurve: this is required to make sure ocs-removal job run's on openshift-storage and not in other ns

yitzhak12: maybe in the teardown part of the test, we can switch back to the default project

ebenahar: why can't we use `-n openshift-storage` in the oc commands instead of switching to a project?

yitzhak12: Do you mean to change it just here in the test or also in the docs? We don't have to use `switch_to_project` in the test. But then we will have a tiny diff from the steps in the doc.

ebenahar: change just here. I don't think that this diff is important

yitzhak12: Sure, np

ebenahar: change just here. I don't think that this diff is important (docs wise)

ebenahar (dismissed):

PrasadDesala (dismissed):

tests/manage/z_cluster/nodes/test_node_replacement_proactive.py
```diff
@@ -38,6 +39,28 @@ def select_osd_node_name():
     return osd_node_name
 
 
+@retry(AssertionError, tries=2, delay=60)
+def check_verification_steps(old_node_name, new_node_name, old_osd_id):
```
ebenahar, line 43: the function name is a bit generic and doesn't reflect that for what the validation steps are

yitzhak12: Fixed

ebenahar (dismissed):

ebenahar (approved):

prsurve (approved):

OdedViner (approved):
LGTM

ocs_ci/ocs/node.py
hnallurv: @yitzhak12 is this comment addressed?

PR #3134, opened 2020-10-14
Fix getting Hostname for worker nodes

prsurve:
fixes: #3133 

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -35,7 +35,7 @@ from ocs_ci.ocs.monitoring import (
     validate_pvc_created_and_bound_on_monitoring_pods,
     validate_pvc_are_mounted_on_monitoring_pods
 )
-from ocs_ci.ocs.node import get_typed_nodes, get_worker_hostname
+from ocs_ci.ocs.node import get_typed_nodes, get_compute_node_names
 from ocs_ci.ocs.resources.catalog_source import CatalogSource
 from ocs_ci.ocs.resources.csv import CSV
 from ocs_ci.ocs.resources.install_plan import wait_for_install_plan_and_approve
@@ -1065,7 +1065,7 @@ def setup_local_storage(storageclass):
         lvd_data = templating.load_yaml(
             constants.LOCAL_VOLUME_DISCOVERY_YAML
         )
-        worker_nodes = get_worker_hostname()
+        worker_nodes = get_compute_node_names(no_replace=True)
 
         # Update local volume discovery data with Worker node Names
         logger.info(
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/node.py
```diff
@@ -1027,3 +1027,18 @@ def node_replacement_verification_steps_ceph_side(old_node_name, new_node_name):
...
+
+    Returns:
+        list: List of worker node hostnames
+
+    """
+    worker_node_objs = get_typed_nodes()
+    return [
+        worker_obj.get()['metadata']['labels'][constants.HOSTNAME_LABEL]
+        for worker_obj in worker_node_objs
+    ]
```
vavuthu, line 1044: Could you fix this in the same function "get_compute_node_names" or remove one function.

By looking into names "get_worker_hostname" and "get_compute_node_names" , both looks like similar behavior. We need to maintain single function if the behavior is same.

prsurve: if we go with get_compute_node_names for baremetal,it will replace hostname with "-" for eg:- hostname of node "bm.woker.one.test.openshift" return value from above function is "bm-worker-one-test-openshift" this return value does not match the node selector for host and that will cause ocs install failure

ebenahar: can we add a condition to `get_compute_node_names` to return in the desired format in case needed?

prsurve: if we add there it will get more complicated because the same function is used for verification of OSD's and hostname

prsurve: reason for creating a new function was the return value was we get from those function 

Return Value from get_compute_node_names
> INFO - ['argo005-ceph-redhat-com', 'argo006-ceph-redhat-com', 'argo007-ceph-redhat-com']

Return Value from get_worker_hostname
> INFO - ['argo005.ceph.redhat.com', 'argo006.ceph.redhat.com', 'argo007.ceph.redhat.com']

Expected return value for using as selector
```
$ oc get nodes -o json |jq -r '.items|.[].metadata.labels."kubernetes.io/hostname"'
argo002.ceph.redhat.com
argo003.ceph.redhat.com
argo004.ceph.redhat.com
argo005.ceph.redhat.com
argo006.ceph.redhat.com
argo007.ceph.redhat.com
```
get_worker_hostname will return the expected value that's the reason created the new function

prsurve: Switch to old function added extra param not to replace hostname

vasukulkarni (approved):

ocs_ci/deployment/deployment.py
```diff
@@ -1065,7 +1065,7 @@ def setup_local_storage(storageclass):
         lvd_data = templating.load_yaml(
             constants.LOCAL_VOLUME_DISCOVERY_YAML
         )
-        worker_nodes = get_compute_node_names()
+        worker_nodes = get_compute_node_names(no_replace=True)
```
vasukulkarni, line 1068: minor nit, if you want to change variable name.

ebenahar (approved):

PR #3142, opened 2020-10-14
Verify OSD is Encrypted [deployment test]

OdedViner:
a.Get node where the OSD runs
$ oc get pods -n openshift-storage -o wide | grep -i osd
b.Go to Node and run "lsblk" command
$ oc debug node/XYZ
sh-4.2# chroot /host /bin/bash
[root@ip-10-0-142-26 /]# lsblk
c.Verify OSD Encrypted

### DIFF
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -370,7 +370,9 @@ def osd_encryption_verification():
     """
     Verify if OSD encryption at rest if successfully deployed on OCS
 
-    :raise: UnsupportedFeatureError, EnvironmentError
+    Raises:
+        UnsupportedFeatureError: OCS version is smaller than 4.6
+        EnvironmentError: The OSD is not encrypted
     """
     ocs_version = float(config.ENV_DATA['ocs_version'])
     if ocs_version < 4.6:
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -362,17 +362,16 @@ def ocs_install_verification(
         # a verification of the installation of it will run
         # on all running state pods
         check_fips_enabled()
+    if config.ENV_DATA.get("encryption_at_rest"):
+        osd_encryption_verification()
 
 
 def osd_encryption_verification():
     """
-    osd_encryption_verification
+    Verify if OSD encryption at rest if successfully deployed on OCS
 
-    Returns:
-        bool: True in case no configure encryption mode.
+    :raise: UnsupportedFeatureError, EnvironmentError
     """
-    if not config.ENV_DATA.get("encryption_at_rest"):
-        return True
     ocs_version = float(config.ENV_DATA['ocs_version'])
     if ocs_version < 4.6:
         error_message = "Encryption at REST can be enabled only on OCS >= 4.6!"
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -2,6 +2,7 @@
 StorageCluster related functionalities
 """
 import logging
+import re
 
 from jsonschema import validate
 
@@ -13,6 +14,9 @@ from ocs_ci.ocs.ocp import get_images, OCP
 from ocs_ci.ocs.resources.ocs import get_ocs_csv
 from ocs_ci.ocs.resources.pod import get_pods_having_label
 from ocs_ci.utility import localstorage, utils
+from ocs_ci.ocs.node import get_osd_running_nodes
+from ocs_ci.ocs.exceptions import UnsupportedFeatureError
+from ocs_ci.utility.utils import run_cmd
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


tests/ecosystem/deployment/test_deployment.py
```diff
@@ -39,3 +46,38 @@ def test_deployment(pvc_factory, pod_factory):
...
+@skipif_no_encrypted
+@deployment
+def test_verify_osd_encryption():
+    """
+    Verify the OSD is encrypted
+    """
+    ocs_version = float(config.ENV_DATA['ocs_version'])
+    if ocs_version < 4.6:
+        error_message = "Encryption at REST can be enabled only on OCS >= 4.6!"
+        log.error(error_message)
```
clacroix12, line 60: Remove this log statement, you're already including this message in the raised error.

tests/ecosystem/deployment/test_deployment.py
```diff
@@ -39,3 +45,40 @@ def test_deployment(pvc_factory, pod_factory):
...
+    if ocs_version < 4.6:
+        error_message = "Encryption at REST can be enabled only on OCS >= 4.6!"
+        raise UnsupportedFeatureError(error_message)
+
+    osd_node_names = get_osd_running_nodes()
+    osd_size = get_osd_size()
+    lsblk_output_list = []
+    for worker_node in osd_node_names:
+        lsblk_cmd = 'oc debug node/' + worker_node + ' -- chroot /host lsblk'
+        log.info(lsblk_cmd)
```
ebenahar, line 66: this logging is not needed because `out = run_cmd(lsblk_cmd)` will print the executed command

tests/ecosystem/deployment/test_deployment.py
```diff
@@ -39,3 +45,40 @@ def test_deployment(pvc_factory, pod_factory):
...
+        out = run_cmd(lsblk_cmd)
+        log.info(out)
```
ebenahar, line 68:
```suggestion
        log.info(f"the output from lsblk command is {out}")
```

tests/ecosystem/deployment/test_deployment.py
```diff
@@ -39,3 +45,40 @@ def test_deployment(pvc_factory, pod_factory):
         log.info(
             "Cluster will be destroyed during teardown part of this test."
         )
+
+
+@skipif_no_encrypted
+@deployment
+def test_verify_osd_encryption():
```
petr-balogh, line 52: This is problematic to have it defined as separate test case I think..

1) using skipif we will see it in the list of skipped tests for deployment all the time when we don't run against environment encrypted at rest
2) I am afraid about the teardown, that this will run as well as we run teardown like:
```
run-ci -m deployment \
    --cluster-name kerberos_ID-ocs-deployment \
    --cluster-path /home/my_user/my-ocs-dir tests/ \
    --teardown
```

Which mean's that test will run after the environment will be destroyed and will fail IMO.

I would add this check directly to the block after line 42:
https://github.com/red-hat-storage/ocs-ci/pull/3142/files#diff-572009fd6960f0b3ddfbf7a115af85876bdd2331d90084d188f96d59e604c98bR42

So you can keep it as separate function, and call it from there, but I would avoid have it as separate test case cause of the issues mentioned above.

petr-balogh (changes requested):

tests/ecosystem/deployment/test_deployment.py
petr-balogh: As we discussed offline, we decided to move it to ocs_install_verification , just after FIPS verification. And not have it as separate test case.

Update about teardown - it will probably not fail the destroy as cluster is session scope autouse fixture, but it will run this verification test also in destroy which is not intention here. So with the move to ocs_install_verification we solve everything.

Thanks

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -360,6 +364,45 @@ def ocs_install_verification(
         check_fips_enabled()
 
 
+def osd_encryption_verification():
+    """
+    osd_encryption_verification
```
petr-balogh, line 369: This can be changed to something more readable?
Verify if OSD encryption at rest if successfully deployed on OCS

Or something like this

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -360,6 +364,45 @@ def ocs_install_verification(
         check_fips_enabled()
```
petr-balogh, line 364: I thought that you will add this verification here.

```python
if config.ENV_DATA.get("encryption_at_rest"):
    osd_encryption_verification()
```

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -360,6 +364,45 @@ def ocs_install_verification(
...
 
+def osd_encryption_verification():
+    """
+    osd_encryption_verification
+
+    Returns:
+        bool: True in case no configure encryption mode.
+    """
+    if not config.ENV_DATA.get("encryption_at_rest"):
+        return True
```
petr-balogh, line 375: This can be removed if you will move call of this function to other place.

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -360,6 +364,45 @@ def ocs_install_verification(
...
+    ocs_version = float(config.ENV_DATA['ocs_version'])
+    if ocs_version < 4.6:
+        error_message = "Encryption at REST can be enabled only on OCS >= 4.6!"
+        raise UnsupportedFeatureError(error_message)
```
petr-balogh, line 379: I see you are raising several Exceptions here but missing them in docstring in `Raises:` sections

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -360,6 +364,45 @@ def ocs_install_verification(
...
+        lsblk_cmd = 'oc debug node/' + worker_node + ' -- chroot /host lsblk'
+        out = run_cmd(lsblk_cmd)
+        log.info(f"the output from lsblk command is {out}")
+        lsblk_output_list.append(out)
+
+    for node_output_lsblk in lsblk_output_list:
+        node_lsb = node_output_lsblk.split()
+        # Search 'crypt' in node_lsb list
+        if 'crypt' not in node_lsb:
+            raise EnvironmentError('OSD is not encrypted')
```
petr-balogh, line 394: the same

tests/ecosystem/deployment/test_deployment.py
```diff
@@ -23,6 +25,7 @@ def test_deployment(pvc_factory, pod_factory):
                 'ocs_registry_image'
             )
             ocs_install_verification(ocs_registry_image=ocs_registry_image)
+            osd_encryption_verification()
```
petr-balogh, line 28: As I wrote, please call this from  osd_encryption_verification function

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -358,6 +362,44 @@ def ocs_install_verification(
...
         check_fips_enabled()
+    if config.ENV_DATA.get("encryption_at_rest"):
+        osd_encryption_verification()
+
+
+def osd_encryption_verification():
+    """
+    Verify if OSD encryption at rest if successfully deployed on OCS
+
+    :raise: UnsupportedFeatureError, EnvironmentError
```
petr-balogh, line 373:
```suggestion
    Raises:
      UnsupportedFeatureError: when
      EnvironmentError: when
```

Look at example here:
https://github.com/red-hat-storage/ocs-ci/blob/dc944d2de87bdde0dd7a39c1bda2843f1ea9feb4/ocs_ci/ocs/resources/packagemanifest.py#L45

petr-balogh (commented):
Otherwise looks OK to me

petr-balogh (approved):

vasukulkarni (approved):

clacroix12 (approved):

PR #3146, opened 2020-10-15
exec_oc_cmd: use kubeconfig from cluster dir if env. variable KUBECONFIG not set

dahorak:
Use kubeconfig from cluster dir for `oc` command executed via `exec_oc_cmd` method (if evn. variable `KUBECONFIG` not set).

When OCP cluster is installed via Flexy, `KUBECONFIG` env. variable is not directly set, but `kubeconfig` file is available in `auth` directory in cluster dir, so it make sense to check if it exists and use it from there. 

### DIFF
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -115,16 +115,14 @@ class OCP(object):
 
         """
         oc_cmd = "oc "
-        if os.getenv('KUBECONFIG'):
-            kubeconfig = os.getenv('KUBECONFIG')
-        elif os.path.exists(os.path.join(
+        cluster_dir_kubeconfig = os.path.join(
             config.ENV_DATA['cluster_path'],
             config.RUN.get('kubeconfig_location')
-        )):
-            kubeconfig = os.path.join(
-                config.ENV_DATA['cluster_path'],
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/ocp.py
```diff
@@ -115,7 +115,19 @@ def exec_oc_cmd(
 
         """
         oc_cmd = "oc "
-        kubeconfig = os.getenv('KUBECONFIG')
+        if os.getenv('KUBECONFIG'):
+            kubeconfig = os.getenv('KUBECONFIG')
+        elif os.path.exists(os.path.join(
+            config.ENV_DATA['cluster_path'],
+            config.RUN.get('kubeconfig_location')
+        )):
```
petr-balogh, line 123: I would define this:  
```
os.path.join(
            config.ENV_DATA['cluster_path'],
            config.RUN.get('kubeconfig_location')
        ) 
```
as a variable to avoid double constructing the same string line bellow.

dahorak: Addressed in 5063216.

petr-balogh (approved):

vavuthu (approved):

PR #3153, opened 2020-10-15
Set constant limits and requests for noobaa endpoints

### LINKED ISSUE #3154: NB endpoints have null in storagecluster CR for limits and requests, which harms HPA
As raised by @nb-ohad:

OCS clusters deployed with ocs-ci have the following configuration on the storagecluster under the spec.resources section: resources:
      mds: {}
      mgr: {}
      mon: {}
      noobaa-core: {}
      noobaa-db: {}
      noobaa-endpoint: {}
      rgw: {}I traced it back to deployment.py lines 532-542"            if config.DEPLOYMENT.get('allow_lower_instance_requirements'):
                none_resources = {'Requests': None, 'Limits': None}
                deviceset_data["resources"] = deepcopy(none_resources)
                resources = [
                    'mon', 'mds', 'rgw', 'mgr', 'noobaa-core', 'noobaa-db',
                ]
                if ocs_version >= 4.5:
                    resources.append('noobaa-endpoint')
                cluster_data['spec']['resources'] = {
                    resource: deepcopy(none_resources) for resource in resources
                }Can some please explain what is the reasoning behind this configuration, was it intended? a bug? 

I am asking as this seems to be the reason for the following BZs:
https://bugzilla.redhat.com/show_bug.cgi?id=1885313 (OCS BZ)
https://bugzilla.redhat.com/show_bug.cgi?id=1885320 (OCP BZ)

ebenahar:
Fixes #3154 

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -561,6 +561,11 @@ class Deployment(object):
                         }
                     cluster_data['spec']['resources'] = resources
 
+            cluster_data['spec']['resources']['noobaa-endpoint'] = {
+                'limits': {'cpu': 1, 'memory': '500Mi'},
+                'requests': {'cpu': 1, 'memory': '500Mi'}
+            }
+
         # Enable host network if enabled in config (this require all the
         # rules to be enabled on underlaying platform).
         if config.DEPLOYMENT.get('host_network'):
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -540,6 +540,11 @@ class Deployment(object):
                 cluster_data['spec']['resources'] = {
                     resource: deepcopy(none_resources) for resource in resources
                 }
+                if ocs_version >= 4.5:
+                    cluster_data['spec']['resources']['noobaa-endpoint'] = {
+                        'limits': {'cpu': 1, 'memory': '500Mi'},
+                        'requests': {'cpu': 1, 'memory': '500Mi'}
+                    }
             else:
                 local_storage = config.DEPLOYMENT.get('local_storage')
                 platform = config.ENV_DATA.get('platform', '').lower()
@@ -560,12 +565,6 @@ class Deployment(object):
                             'requests': {'cpu': 1, 'memory': '8Gi'}
                         }
                     cluster_data['spec']['resources'] = resources
-
-            cluster_data['spec']['resources']['noobaa-endpoint'] = {
-                'limits': {'cpu': 1, 'memory': '500Mi'},
-                'requests': {'cpu': 1, 'memory': '500Mi'}
-            }
-
         # Enable host network if enabled in config (this require all the
         # rules to be enabled on underlaying platform).
         if config.DEPLOYMENT.get('host_network'):


ocs_ci/deployment/deployment.py
```diff
@@ -561,6 +561,11 @@ def deploy_ocs_via_operator(self):
                         }
                     cluster_data['spec']['resources'] = resources
 
+            cluster_data['spec']['resources']['noobaa-endpoint'] = {
```
petr-balogh, line 564: Elad, this block should be moved to line 542.

I would change the logic: https://github.com/red-hat-storage/ocs-ci/pull/3153/files#diff-8571fc52f1040ac2cb01985da6d61cc93a535f812814c200b45d00daf28c2e3fR538-R539 here to:

```python
                cluster_data['spec']['resources'] = {
                    resource: deepcopy(none_resources) for resource in resources
                }
                if ocs_version >= 4.5:
                    cluster_data['spec']['resources']['noobaa-endpoint'] = {
                        'limits': {'cpu': 1, 'memory': '500Mi'},
                        'requests': {'cpu': 1, 'memory': '500Mi'}
                    }
```

ebenahar: @petr-balogh won't this change be relevant for clusters that don't fall under `allow_lower_instance_requirements`? 
The intention is to add constant values to the resource requests and limits, right?

ebenahar:
@nb-ohad

ocs_ci/deployment/deployment.py
petr-balogh: IIUC the issue discussed on our last mtg with @nb-ohad was related only on clusters deployed with lower resources as only if that's true we allow and run this block of the code:

https://github.com/red-hat-storage/ocs-ci/blob/b996cf9ab9cad58007255f3d603be1c7dd5665ac/ocs_ci/deployment/deployment.py#L532-L542

Which will set resources to empty dirs.  In all other cases we are just following documentation and setting up resources/limits based on the documentation like for example LSO deployment on AWS I3.

In any other case we should not set the limits - so the default from the product will be used and that should be OK.

nb-ohad: @ebenahar @petr-balogh 
The main takeaway from our mtg was that the values for the resources configuration (requests and limits) should always be set and should never be `null` or `{}`, at least for the noobaa-endpoint configuration, if not for all.

Now, I am not sure where this code should lay as I am not aware of all the details and the code structure of the deployment code.

petr-balogh: Yep, and this should happen in our code only in the block mentioned above :

```python
if config.DEPLOYMENT.get('allow_lower_instance_requirements'): 
```

For now, we will start with the noobaa endpoint as it will take more time to find the proper configuration for all other resources.

ebenahar: addressed now

petr-balogh (approved):

dahorak (approved):

petr-balogh:
@MeridianExplorer hello, as @ebenahar is on PTO, can you please take a look at tier1 results from verification job: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/3166/ once it will finish if the results looks OK for noobaa tests please?

It should finish before you will start your day ;)

Thanks

MeridianExplorer:
@petr-balogh - The results look great, I think this is safely mergeable

petr-balogh:
Thanks @MeridianExplorer merged. @nb-ohad  FYI

Issue #3165, opened 2020-10-19
Delete IO upgrade tests


petr-balogh:
Now we have IO in BG  option, so I think that we can remove the @fbalak  tests in upgrade which are usually skipped when we run io_in_bg, but if we need to run without this flag, then the IO upgrade tests written by Filip will be triggered which causing the issues we saw in the past with too big load on the cluster.

I think we can remove those tests now.
WDYT @ebenahar @fbalak.

petr-balogh:
@fbalak  I think that here we see upgrade didn't pass cause of too heavy load on cluster with IO tests.

https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/13693

As I forgot to run with IO in bg it triggered this test.

If we decide to not delete those IO tests we should disable them now at least to not run them even when IO in BG is disabled.

petr-balogh:
After the discussion we decided to not delete those tests IIRC

PR #3166, opened 2020-10-19
Remove IO upgrade tests

petr-balogh:
Fixes: #3165


petr-balogh:
After discussion with Filip, we will probably not delete those test but rather align

Issue #3186, opened 2020-10-22
ocs-ci returned wrong hostname from ceph osd tree 


prsurve:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/13627/console

PR #3187, opened 2020-10-22
Fix getting wrong value from get_compute_node_names

prsurve:
Fixes: #3186 

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -657,15 +657,17 @@ def get_compute_node_names(no_replace=False):
     """
     platform = config.ENV_DATA.get('platform').lower()
     compute_node_objs = get_typed_nodes()
-    if platform == constants.VSPHERE_PLATFORM or constants.AWS_PLATFORM:
+    if platform == (constants.VSPHERE_PLATFORM or constants.AWS_PLATFORM):
         return [
             compute_obj.get()['metadata']['labels'][constants.HOSTNAME_LABEL]
             for compute_obj in compute_node_objs
         ]
     elif (
-        platform == constants.BAREMETAL_PLATFORM
-        or platform == constants.BAREMETALPSI_PLATFORM
-        or platform == constants.IBM_POWER_PLATFORM
+        platform == (
+            constants.BAREMETAL_PLATFORM
+            or constants.BAREMETALPSI_PLATFORM
+            or constants.IBM_POWER_PLATFORM
+        )
     ):
         if no_replace:
             return [


PrasadDesala (approved):

jilju (approved):

ocs_ci/ocs/node.py
```diff
@@ -657,15 +657,17 @@ def get_compute_node_names(no_replace=False):
     """
     platform = config.ENV_DATA.get('platform').lower()
     compute_node_objs = get_typed_nodes()
-    if platform == constants.VSPHERE_PLATFORM or constants.AWS_PLATFORM:
+    if platform == (constants.VSPHERE_PLATFORM or constants.AWS_PLATFORM):
```
clacroix12, line 660: @prsurve I don't think this is doing what is intended here. `constants.VSPHERE_PLATFORM or constants.AWS_PLATFORM` will always return `"vsphere"` as that is a "truthful" string. I think what we want instead is to use `"in"`.

```python
if platform in [constants.VSPHERE_PLATFORM, constants.AWS_PLATFORM]:
```

This applies to the other changed code as well. We want to see if the platform is in a particular list of platforms and this use of `or` is not achieving that.

ocs_ci/ocs/node.py
```diff
@@ -657,15 +657,17 @@ def get_compute_node_names(no_replace=False):
...
         return [
             compute_obj.get()['metadata']['labels'][constants.HOSTNAME_LABEL]
             for compute_obj in compute_node_objs
         ]
     elif (
-        platform == constants.BAREMETAL_PLATFORM
-        or platform == constants.BAREMETALPSI_PLATFORM
-        or platform == constants.IBM_POWER_PLATFORM
+        platform == (
+            constants.BAREMETAL_PLATFORM
```
clacroix12, line 667: I believe verification passed because baremetal just happens to be the first in this list and that is what you tested on.

```python
In [1]: from ocs_ci.ocs import constants

In [2]: constants.BAREMETAL_PLATFORM or constants.BAREMETALPSI_PLATFORM or constants.IBM_POWER_PLATFORM
Out[2]: 'baremetal'

In [3]: platform = 'baremetal'

In [4]: platform == (constants.BAREMETAL_PLATFORM or constants.BAREMETALPSI_PLATFORM or constants.IBM_POWER_PLATFORM)
Out[4]: True
```

However if I change the order...
```python

In [5]: platform == (constants.BAREMETALPSI_PLATFORM or constants.IBM_POWER_PLATFORM or constants.BAREMETAL_PLATFORM)
Out[5]: False
```

Issue #3145, opened 2020-10-14
Error during setup - Yaml parser issue yaml.scanner.ScannerError: mapping values are not allowed here 


vasukulkarni:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/13491/

```
==================================== ERRORS ====================================
__________________ ERROR at setup of test_monitoring_enabled ___________________

request = <SubRequest 'cluster_load' for <Function test_monitoring_enabled>>
project_factory_session = <function project_factory_fixture.<locals>.factory at 0x7f986c0d9ea0>
pvc_factory_session = <function pvc_factory_fixture.<locals>.factory at 0x7f9867cda048>
service_account_factory_session = <function service_account_factory_fixture.<locals>.factory at 0x7f9867cda158>
pod_factory_session = <function pod_factory_fixture.<locals>.factory at 0x7f9867cda268>

    @pytest.fixture(scope="session", autouse=True)
    def cluster_load(
        request, project_factory_session, pvc_factory_session,
        service_account_factory_session, pod_factory_session
    ):
        """
        Run IO during the test execution
        """
        cl_load_obj = None
        io_in_bg = config.RUN.get('io_in_bg')
        log_utilization = config.RUN.get('log_utilization')
        io_load = config.RUN.get('io_load')
    
        # IO load should not happen during deployment
        deployment_test = True if 'deployment' in request.node.items[0].location[0] else False
        if io_in_bg and not deployment_test:
            io_load = int(io_load) * 0.01
            log.info(wrap_msg("Tests will be running while IO is in the background"))
            log.info(
                "Start running IO in the background. The amount of IO that "
                "will be written is going to be determined by the cluster "
                "capabilities according to its limit"
            )
            cl_load_obj = ClusterLoad(
                project_factory=project_factory_session,
                sa_factory=service_account_factory_session,
                pvc_factory=pvc_factory_session,
                pod_factory=pod_factory_session,
                target_percentage=io_load
            )
>           cl_load_obj.reach_cluster_load_percentage()

tests/conftest.py:1167: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ocs_ci/ocs/cluster_load.py:206: in reach_cluster_load_percentage
[...]

PR #3192, opened 2020-10-22
Fix cluster load fixture

petr-balogh:
Make the autouse session scope fixture for cluster load more robust.

Fixes: #3145

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1154,9 +1154,16 @@ def cluster_load(
     io_in_bg = config.RUN.get('io_in_bg')
     log_utilization = config.RUN.get('log_utilization')
     io_load = config.RUN.get('io_load')
+    cluster_load_error = None
+    cluster_load_error_msg = (
+        "Cluster load might not work correctly during this run, because "
+        "it failed with an exception: %s"
+    )
 
     # IO load should not happen during deployment
-    deployment_test = True if 'deployment' in request.node.items[0].location[0] else False
+    deployment_test = True if (
+        'deployment' in request.node.items[0].location[0]
+    ) else False
     if io_in_bg and not deployment_test:
         io_load = int(io_load) * 0.01
         log.info(wrap_msg("Tests will be running while IO is in the background"))
@@ -1165,18 +1172,26 @@ def cluster_load(
             "will be written is going to be determined by the cluster "
             "capabilities according to its limit"
         )
-        cl_load_obj = ClusterLoad(
-            project_factory=project_factory_session,
-            sa_factory=service_account_factory_session,
-            pvc_factory=pvc_factory_session,
-            pod_factory=pod_factory_session,
-            target_percentage=io_load
-        )
-        cl_load_obj.reach_cluster_load_percentage()
+        try:
+            cl_load_obj = ClusterLoad(
+                project_factory=project_factory_session,
+                sa_factory=service_account_factory_session,
+                pvc_factory=pvc_factory_session,
+                pod_factory=pod_factory_session,
+                target_percentage=io_load
+            )
+            cl_load_obj.reach_cluster_load_percentage()
+        except Exception as ex:
+            log.error(cluster_load_error_msg, ex)
+            cluster_load_error = ex
 
     if (log_utilization or io_in_bg) and not deployment_test:
         if not cl_load_obj:
-            cl_load_obj = ClusterLoad()
+            try:
+                cl_load_obj = ClusterLoad()
+            except Exception as ex:
+                log.error(cluster_load_error_msg, ex)
+                cluster_load_error = ex
 
         config.RUN['load_status'] = 'running'
 
@@ -1187,6 +1202,8 @@ def cluster_load(
             config.RUN['load_status'] = 'finished'
             if thread:
                 thread.join()
+            if cluster_load_error:
+                raise cluster_load_error
 
         request.addfinalizer(finalizer)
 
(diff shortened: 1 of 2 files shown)


vasukulkarni (approved):

petr-balogh:
Verification is running here:  https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/13787/console

wusui (approved):
LGTM

tests/conftest.py
```diff
@@ -1154,75 +1154,87 @@ def cluster_load(
...
-            "will be written is going to be determined by the cluster "
-            "capabilities according to its limit"
-        )
-        cl_load_obj = ClusterLoad(
-            project_factory=project_factory_session,
-            sa_factory=service_account_factory_session,
-            pvc_factory=pvc_factory_session,
-            pod_factory=pod_factory_session,
-            target_percentage=io_load
+    try:
```
ebenahar, line 1160: we don't need to wrap the entire implementation with try except. The only call needed to be wrapped is `cl_load_obj.reach_cluster_load_percentage()`

petr-balogh: @ebenahar I am not sure what all calls do we do in init of: ClusterLoad , 
I saw there is initialization of prometheusAPI and some others factories so I for sure wrapped everything as if the code will fail somewhere else this is more safe option that we will not fail the rest of the tests cause of the issue here.

So I can end the except at line  1186.

As this autouse session scope fixture is very sensitive and each error/exception means whole execution will fail - and also found out that in case of tier1 it's failing each test case and running must gather for it. So it waste a lot of time.

ebenahar (approved):

tests/conftest.py
```diff
@@ -1187,6 +1202,8 @@ def finalizer():
             config.RUN['load_status'] = 'finished'
             if thread:
                 thread.join()
+            if cluster_load_error:
+                raise cluster_load_error
```
vasukulkarni, line 1206: raise would still cause the tests to fail?

tests/conftest.py
```diff
@@ -1154,9 +1154,16 @@ def cluster_load(
     io_in_bg = config.RUN.get('io_in_bg')
     log_utilization = config.RUN.get('log_utilization')
     io_load = config.RUN.get('io_load')
+    cluster_load_error = None
+    cluster_load_error_msg = (
+        "Cluster might not work correctly during this run, because"
```
ebenahar, line 1159: "Cluster load.."

ebenahar (commented):
Missed the issue Vasu pointed out and one small log fix that is needed so revising my approval

tests/conftest.py
petr-balogh: This will fail at the end of the session as it's in the finalizer, and yep, we should see it as teardown failure I think but all tests should have been already finished.  So that case we will notice that there was an issue with this cluster load, otherwise we could miss that problem.

tests/conftest.py
petr-balogh: done

tests/conftest.py
```diff
@@ -1154,9 +1154,16 @@ def cluster_load(
     io_in_bg = config.RUN.get('io_in_bg')
     log_utilization = config.RUN.get('log_utilization')
     io_load = config.RUN.get('io_load')
+    cluster_load_error = None
+    cluster_load_error_msg = (
+        "Cluster load might not work correctly during this run, because "
+        "it failed with an exception: %s"
```
ebenahar, line 1160: variable is missing - `%s`

tests/conftest.py
ebenahar: this will cause the last test case in the execution to fail which is kind of false negative for this case since the real failure is in the cluster load

petr-balogh: Anyway we need to report this error and I cannot come with any other solution how make it visible in report.. I rather see the last test case fail with this and people who will did analysis will say that it failed cause of issue with cluster load, then we will see there was some issue with cluster load and we didn't maybe run with the load.   How else you would like to notice that error @ebenahar if we will ignore it at all?

tests/conftest.py
petr-balogh: it's not missing, it's formatted later on in logger itself:  https://github.com/red-hat-storage/ocs-ci/pull/3192/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R1185

tests/conftest.py
vasukulkarni: How about ending the complete execution since it is going to cause bunch of failures?

ebenahar (approved):

dahorak (approved):

tests/conftest.py
petr-balogh: Why it should cause bunch of failures?  Worst case it will not run the IO in the background if some problem with ClusterLoad fixture.   So when we spent the time on deploying the cluster and we triggered tests it's better to have results without IO in the background instead of fail whole execution and destroy the cluster.

vasukulkarni: because you are raising an exception in the finalizer, if we truly dont want the test to fail then we shoudn't even care about the exception at the end of the thread.join()

petr-balogh: But then we will never notice that we had some issue with the execution and clusterLoad fixture.  As I discussed with Elad, we really want to see if there was some issue with that and hence one error which will be easily saw in the report and from the analysis they can easily tell what was the cause of that 1 test to fail.

ebenahar: yes, in addition to that, the chances we will have this exception raised are lower since Petr also added retries to get_query() method of ClusterLoad in case of ScannerError exception, which adds one more protection layer

PR #3197, opened 2020-10-22
Fixed how we determine method for retrieving compute node names based on platform

clacroix12:
#3187 introduced some faulty logic in how we match the platform to how we retrieve the compute node names. This change aims to fix that logic and ensure that we are using the right method for each platform.


prsurve (approved):

clacroix12:
Ignore the failed verification, and existing BM cluster exists so deployment failed. I'll verify with another platform

vasukulkarni (approved):

vavuthu (approved):

PR #3209, opened 2020-10-27
Create configs for OCP 4.6 GA

petr-balogh:
As OCP 4.6 is GAed, we need to add the GA config for OCP 4.6.
This is done in this commit.

### DIFF
--- a/conf/ocp_version/ocp-4.6-ga-config.yaml
+++ b/conf/ocp_version/ocp-4.6-ga-config.yaml
@@ -1,19 +1,18 @@
 ---
-# Config file for OCP GA 4.5 stable channel
+# Config file for OCP GA 4.6 stable channel
 
 RUN:
-  client_version: '4.5-ga'
+  client_version: '4.6-ga'
 
 DEPLOYMENT:
   ocp_url_template: "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/{version}/{file_name}-{os_type}-{version}.tar.gz"
-  installer_version: "4.5-ga"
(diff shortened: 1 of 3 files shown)


vasukulkarni:
please add some description.

petr-balogh:
Running another verification as from some reason the fast-4.6 channel is returning 4.5.16 build in the list. So changed fast to stable now and we will see in the next verification job if the OCP was really installed in 4.6.0

AviadP (approved):

vasukulkarni (approved):

vasukulkarni:
> Running another verification as from some reason the fast-4.6 channel is returning 4.5.16 build in the list. So changed fast to stable now and we will see in the next verification job if the OCP was really installed in 4.6.0

I think we should also add asserts for ocp now (in a different PR)

ebenahar (approved):

petr-balogh:
Previous job failed on not enough m4 instances in our zone. Also I found out that the version just got propagated to the API we use. So there was just some delay after GA. So removed the logic I added to return 4.6.0 version and we should be fine.

clacroix12 (approved):

petr-balogh:
@vasukulkarni can you please re-approve? It's verified, so we can merge. Thanks

vasukulkarni (approved):

clacroix12:
@petr-balogh unable to merge because we're still waiting on travis-ci success. For some reason these have been spending a very long time in the queue lately.

vavuthu (approved):

PR #3216, opened 2020-10-29
Add infra nodes to ocp cluster - AWSIPI

shylesh:

### DIFF
--- a/conf/deployment/aws/ipi_3az_rhcos_3m_3w_3i.yaml
+++ b/conf/deployment/aws/ipi_3az_rhcos_3m_3w_3i.yaml
@@ -1,0 +1,26 @@
+---
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
+ENV_DATA:
+  platform: 'aws'
+  deployment_type: 'ipi'
+  region: 'us-east-2'
+  worker_availability_zones:
+    - 'us-east-2a'
+    - 'us-east-2b'
+    - 'us-east-2c'
+  infra_availability_zones:
+    - 'us-east-2a'
+    - 'us-east-2b'
+    - 'us-east-2c'
+  master_availability_zones:
+    - 'us-east-2a'
+    - 'us-east-2b'
+    - 'us-east-2c'
+  worker_replicas: 3
+  master_replicas: 3
+  infra_replicas: 3
+  worker_instance_type: 'm5.4xlarge'
+REPORTING:
+  polarion:
+    deployment_id: 'OCS-xxxx'
--- a/conf/deployment/aws/ipi_3az_rhcos_3m_3w_3i.yaml
+++ b/conf/deployment/aws/ipi_3az_rhcos_3m_3w_3i.yaml
@@ -1,6 +1,7 @@
 ---
 DEPLOYMENT:
   allow_lower_instance_requirements: false
+  ocs_operator_nodes_to_label: 0
 ENV_DATA:
   platform: 'aws'
   deployment_type: 'ipi'
--- a/conf/deployment/aws/ipi_3az_rhcos_3m_3w_3i.yaml
+++ b/conf/deployment/aws/ipi_3az_rhcos_3m_3w_3i.yaml
@@ -24,4 +24,4 @@ ENV_DATA:
   worker_instance_type: 'm5.4xlarge'
 REPORTING:
   polarion:
-    deployment_id: 'OCS-xxxx'
+    deployment_id: 'OCS-2430'
--- a/conf/ocsci/infra_nodes.yaml
+++ b/conf/ocsci/infra_nodes.yaml
@@ -14,3 +14,4 @@ ENV_DATA:
 
 DEPLOYMENT:
   infra_nodes: True
+  ocs_operator_nodes_to_label: 0
--- a/conf/ocsci/infra_nodes.yaml
+++ b/conf/ocsci/infra_nodes.yaml
@@ -1,0 +1,14 @@
+---
+# This config will provide you the environment with:
+#  * 3 master nodes
+#  * 6 worker nodes where:
+#    * 3 worker nodes used for OCS
+#    * 3 worker nodes used for workload
+#
+ENV_DATA:
+  master_replicas: 3
+  worker_replicas: 3
+  infra_replicas: 3
+
+DEPLOYMENT:
+  infra_nodes: True
--- a/conf/ocsci/infra_nodes.yaml
+++ b/conf/ocsci/infra_nodes.yaml
@@ -9,6 +9,8 @@ ENV_DATA:
   master_replicas: 3
   worker_replicas: 3
   infra_replicas: 3
+  infra_instance_type: 'm5.4xlarge'
+
 
 DEPLOYMENT:
   infra_nodes: True
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -183,7 +183,7 @@ class AWSIPI(AWSBase):
         """
         super(AWSIPI, self).deploy_ocp(log_cli_level)
         if config.DEPLOYMENT['infra_nodes']:
-            num_nodes = config.DEPLOYMENT['ocs_operator_nodes_to_taint']
+            num_nodes = config.ENV_DATA['infra_replicas']
             ms_list = machine.create_infra_nodes(num_nodes)
             for node in ms_list:
                 machine.wait_for_new_node_to_be_ready(node)
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -182,8 +182,8 @@ class AWSIPI(AWSBase):
                 (default: "DEBUG")
         """
         super(AWSIPI, self).deploy_ocp(log_cli_level)
-        if config.DEPLOYMENT['infra_nodes']:
-            num_nodes = config.ENV_DATA['infra_replicas']
+        if config.DEPLOYMENT.get('infra_nodes'):
+            num_nodes = config.ENV_DATA.get('infra_replicas', 3)
             ms_list = machine.create_infra_nodes(num_nodes)
             for node in ms_list:
                 machine.wait_for_new_node_to_be_ready(node)
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -12,7 +12,7 @@ import boto3
 from ocs_ci.cleanup.aws.defaults import CLUSTER_PREFIXES_SPECIAL_RULES
 from ocs_ci.deployment.ocp import OCPDeployment as BaseOCPDeployment
 from ocs_ci.framework import config
-from ocs_ci.ocs import constants, exceptions, ocp
+from ocs_ci.ocs import constants, exceptions, ocp, machine
 from ocs_ci.ocs.resources import pod
 from ocs_ci.utility import templating
 from ocs_ci.utility.aws import (
@@ -182,7 +182,12 @@ class AWSIPI(AWSBase):
                 (default: "DEBUG")
         """
         super(AWSIPI, self).deploy_ocp(log_cli_level)
-        if config.DEPLOYMENT.get("host_network"):
+        if config.DEPLOYMENT['infra_nodes']:
+            num_nodes = config.DEPLOYMENT['ocs_operator_nodes_to_taint']
+            ms_list = machine.create_infra_nodes(num_nodes)
+            for node in ms_list:
+                machine.wait_for_new_node_to_be_ready(node)
+        if config.DEPLOYMENT.get('host_network'):
             self.host_network_update()
 
     def destroy_cluster(self, log_level="DEBUG"):
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -184,7 +184,7 @@ class AWSIPI(AWSBase):
         super(AWSIPI, self).deploy_ocp(log_cli_level)
         if config.DEPLOYMENT.get("infra_nodes"):
             num_nodes = config.ENV_DATA.get("infra_replicas", 3)
-            ms_list = machine.create_infra_nodes(num_nodes)
+            ms_list = machine.create_ocs_infra_nodes(num_nodes)
             for node in ms_list:
                 machine.wait_for_new_node_to_be_ready(node)
         if config.DEPLOYMENT.get("host_network"):
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -182,12 +182,12 @@ class AWSIPI(AWSBase):
                 (default: "DEBUG")
         """
         super(AWSIPI, self).deploy_ocp(log_cli_level)
-        if config.DEPLOYMENT.get('infra_nodes'):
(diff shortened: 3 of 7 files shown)


ocs_ci/ocs/machine.py
```diff
@@ -275,6 +298,39 @@ def create_custom_machineset(
...
 
+def create_infra_nodes(num_nodes):
+    """
+    Create infra node instances
+
+    Args:
+        num_nodes (int): Number of instances to be created
+
+    Returns:
+        list : of instance names
```
vavuthu, line 309:
```suggestion
        list: list of instance names
```

shylesh: done

shylesh:
rebased on latest master

conf/ocsci/infra_nodes.yaml
```diff
@@ -0,0 +1,14 @@
+---
```
ebenahar, line 1: do we need a new conf file or we can use conf/ocsci/tainted_worker_nodes.yaml and update it?

ebenahar (commented):
looks good, just one comment

ocs_ci/ocs/machine.py
```diff
@@ -263,6 +266,26 @@ def create_custom_machineset(
                     'tags'
                 ][0]['name'] = f"kubernetes.io/cluster/{cls_id}"
 
+                # if we have infra nodes to be added then we need
+                # taint and extra label
+                if role == 'infra':
+                    taint_dict = {
```
ebenahar, line 272: once more thing - this is quite similar to https://github.com/red-hat-storage/ocs-ci/blob/5dd1743cff414d2de5555b2d20f30cfcb64511c9/ocs_ci/deployment/deployment.py#L234

Should we maybe create a common function and call it from both places?

conf/ocsci/infra_nodes.yaml
shylesh: @ebenahar ,
Earlier I used the same file you mentioned, but after discussion with @petr-balogh , we arrived at a separate conf file because infra count was different in tainted_worker_nodes.yaml and not sure what are all the tests depends on it. Moreover this has default count of 6 workers on top of which we will be adding 3 more infra nodes using machinesets. 

If you still feel this file will be sufficient then I can revert back to tainted_worker_nodes.yaml . WDYT ?

ocs_ci/ocs/machine.py
shylesh: The difference I see between those usages is 
1. In one case , we already have nodes and we are just applying lables 
2. Another case which this PR has the code, we create the instances using machinesets whose resource yaml already has the labels.

I am not sure if creating the machineset first and then applying the label and taints has same effect as creating machineset along with labels and taints ? Please clarify.

conf/ocsci/infra_nodes.yaml
ebenahar: I am ok with either way, so we can have the new file

ocs_ci/ocs/machine.py
ebenahar: oh I just meant that maybe we can move the code which is common to both places to be in a separated function

ebenahar: not a big deal, feel free to ignore

ocs_ci/ocs/machine.py
```diff
@@ -275,6 +298,39 @@ def create_custom_machineset(
         raise UnsupportedPlatformError("Functionality not supported in UPI")
 
 
+def create_infra_nodes(num_nodes):
```
Shrivaibavi, line 301: My understanding when we say infra we not only label the nodes as infra, we also taint the infra nodes to enable OCS services on nodes. so I feel the name of the function can be create_ocs_infra_nodes().

ocs_ci/ocs/machine.py
```diff
@@ -275,6 +298,39 @@ def create_custom_machineset(
...
+    )
+    for machine in machinesets_obj.get()['items']:
+        aws_zone = machine.get('spec').get('template').get('spec').get(
+            'providerSpec').get('value').get('placement').get('availabilityZone')
+        zone_list.append(aws_zone[-1])
+
+    ms_names.extend(
+        [
+            create_custom_machineset(
+                role='infra',
```
Shrivaibavi, line 325: don't we need to pass label here, as the default label i see in the function is "app-scale"

Shrivaibavi:
The PR looks good just a few small comments and questions to clarify

conf/ocsci/infra_nodes.yaml
```diff
@@ -0,0 +1,14 @@
...
+# This config will provide you the environment with:
+#  * 3 master nodes
+#  * 6 worker nodes where:
+#    * 3 worker nodes used for OCS
+#    * 3 worker nodes used for workload
+#
+ENV_DATA:
+  master_replicas: 3
+  worker_replicas: 3
+  infra_replicas: 3
```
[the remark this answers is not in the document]
shylesh: sure

ocs_ci/deployment/deployment.py
```diff
@@ -231,7 +231,10 @@ def label_and_taint_nodes(self):
                     f"{constants.OPERATOR_NODE_LABEL} --overwrite"
                 )
             ]
-            if config.DEPLOYMENT["infra_nodes"]:
+            if (
+                config.DEPLOYMENT["infra_nodes"]
+                and not config.ENV_DATA.get('infra_replicas')
```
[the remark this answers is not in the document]
shylesh: still there is bit of confusion here for me. if we don't have 'infra_replicas' then we will have 3M , 3W . 
if we have 'infra_replicas' then we will end up having 3M, 3W, 3Infra nodes (don't consider UPI with 'infra_nodes', which would make the setup as 3M, 3Worker,Infra,App, 3Infra).

Am I totally missing the config here ? please let me know

ocs_ci/ocs/machine.py
shylesh: sure

ocs_ci/ocs/machine.py
shylesh: The label necessary for 'infra' nodes has been handled as an extra 'if' condition, assuming that these are extra labels along with whatever default (app-scale) we want.

Don't we need those default labels ? I am not really aware of label requirements here so I request you to clarify on this.

PrasadDesala: AFAIK, app-scale label is a custom label which scale team created for running their scale pgsql test.
@ramkiperiy @tiffanyn108 Please correct me if I wrong.
This is not need for normal deployments (non-scale).

ocs_ci/ocs/machine.py
```diff
@@ -275,6 +298,47 @@ def create_custom_machineset(
         raise UnsupportedPlatformError("Functionality not supported in UPI")
 
 
+def create_infra_nodes(num_nodes):
```
PrasadDesala, line 301: create_custom_machineset takes m4.xlarge as default input. Are we creating infra worker nodes with "m5.4xlarge" ?

ocs_ci/deployment/aws.py
```diff
@@ -177,6 +177,11 @@ def deploy_ocp(self, log_cli_level='DEBUG'):
                 (default: "DEBUG")
         """
         super(AWSIPI, self).deploy_ocp(log_cli_level)
+        if config.DEPLOYMENT['infra_nodes']:
```
PrasadDesala, line 180: Post deployment we expect few set of ocs pods to run on the worker,infra tainted nodes. IMO, it will be good to have a verification step for this post OCS deployment.
WDYT?

ocs_ci/deployment/deployment.py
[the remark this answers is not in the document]
petr-balogh: I think that we should have more free hand what we really want to label/taint here.

So I can think of few scenarios:
* 3w - labeled
* 3w - labeled, tainted - 3 workers not labeled used for workload
* 3w/i nodes - labeled, tainted - this case we will need to have another 3 worker nodes to enable workload to run somewhere
* 3w/i nodes - labeled 
* 3i labeled, 3w
* 3i labeled/tainted , 3w

So for this purpose I suggest to change the logic for:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L56

`ocs_operator_nodes_to_label` to be not int value but a dict which will have number defined for workers and infra nodes:
```python
ocs_operator_nodes_to_label = {
    "worker": 3,
    "infra": 3,
}
```
or have another value for infra nodes like 
`ocs_operator_infra_nodes_to_label` 

and the same for `ocs_operator_nodes_to_taint` which we need transform to dict or have `ocs_operator_infra_nodes_to_taint `

For easier use will be better to go with `ocs_operator_infra_nodes_to_label` and `ocs_operator_infra_nodes_to_taint`

@sidhant-agrawal @Shrivaibavi @ypersky1980 @shylesh @ebenahar   WDYT about this? Does those combinations makes sense or they don't?

ocs_ci/ocs/machine.py
```diff
@@ -173,7 +173,10 @@ def delete_machine_and_check_state_of_new_spinned_machine(machine_name):
 
 
 def create_custom_machineset(
-    role='app', instance_type='m4.xlarge', label='app-scale', zone='a'
+    role='app',
+    instance_type='m4.xlarge',
+    label='app-scale',
+    zone='a',
```
petr-balogh, line 179: I am thinking about add two additional parameters which will be:
labels and taints with default with type list and default value None,

If some of them will be passed to this function then those extra labels, taints will be added to proper lists in machineset which will make this function more generic.

Then from the place where we would like to create those extra infra nodes we will pass those informations from there where we will read information from values like I mentioned `ocs_operator_infra_nodes_to_label` and `ocs_operator_infra_nodes_to_taint` And based on this info we will pass what we really want from that place.

WDYT?

I am nto sure about default label `app-scale` here, this I think should not be applied on infra nodes, please discuss with author of this function and check usage of that. I think it should be None as default and from places where it's really need to be used and intended to use app-sclae label it has to be passed form place where it's used.

ocs_ci/ocs/machine.py
petr-balogh: This should depend on config value, or it can be parametrized in the funciton with None as default value, and if passed by function parameter it will use this value, otherwise it will use value from config and if not defined in config it can use some default value - e.g. same as worker node.

ocs_ci/ocs/machine.py
petr-balogh: I commented on other place that this should not be default behavior of the function and should not be templated by default by this custom label.

ocs_ci/ocs/machine.py
Shrivaibavi: I agree on this, This can be parametrized according to the user

ocs_ci/deployment/aws.py
shylesh: https://github.com/red-hat-storage/ocs-ci/issues/3291

conf/ocsci/infra_nodes.yaml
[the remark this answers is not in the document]
shylesh: keeping this as infra_replicas

ocs_ci/deployment/aws.py
shylesh: created a github issue

ocs_ci/deployment/deployment.py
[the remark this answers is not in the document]
shylesh: @PrasadDesala @ebenahar @Shrivaibavi , could you guys please provide feedback on the possible combinations ?

ocs_ci/ocs/machine.py
```diff
@@ -263,6 +266,26 @@ def create_custom_machineset(
...
 
+                # if we have infra nodes to be added then we need
+                # taint and extra label
+                if role == 'infra':
+                    taint_dict = {
+                        'effect': 'NoSchedule',
+                        'key': 'node.ocs.openshift.io/storage',
+                        'value': 'true',
+                    }
+                    machineset_yaml['spec']['template']['spec'].update({'taints': list()})
```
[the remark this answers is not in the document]
shylesh: By introducing seperate yaml , we are no longer using ocs_operator_nodes_to_taint.

ocs_ci/deployment/deployment.py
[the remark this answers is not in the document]
AviadP: i dont think we should label infra nodes. only workers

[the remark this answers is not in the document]
AviadP: @petr-balogh , as far as i know, Infra nodes should not be have label

shylesh:
Rebased on latest master

ebenahar (dismissed):

petr-balogh (approved):

shylesh:
Verified:https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/14990

conf/deployment/aws/ipi_3az_rhcos_3m_3w_3i.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - 'us-east-2a'
+    - 'us-east-2b'
+    - 'us-east-2c'
+  worker_replicas: 3
+  master_replicas: 3
+  infra_replicas: 3
+  worker_instance_type: 'm5.4xlarge'
+REPORTING:
+  polarion:
+    deployment_id: 'OCS-xxxx'
```
clacroix12, line 27: Is there an actual polarion ID that should be here?  If not let's remove the whole reporting section from this file.

ocs_ci/ocs/machine.py
```diff
@@ -191,7 +195,8 @@ def create_custom_machineset(
     Args:
         role (str): Role type to be added for node eg: it will be app,worker
         instance_type (str): Type of aws instance
-        label (str): Label to be added to the node
+        label (list of tuples): List of Labels (key, val) to be added to the node
+        taints (list of dict): List of taints to be applied
```
clacroix12, line 199:
```suggestion
        label (list): List of Labels (key, val) to be added to the node
        taints (list): List of taints to be applied
```
The types we specify here need to be actual python types. If we want to elaborate on what kind of information is in the list we can do that in the description.

shylesh: ack

conf/deployment/aws/ipi_3az_rhcos_3m_3w_3i.yaml
shylesh: I will get try to get this added to polarion, if not I will remove this.

ocs_ci/ocs/machine.py
shylesh: done

conf/deployment/aws/ipi_3az_rhcos_3m_3w_3i.yaml
shylesh: done

shylesh:
Rebased on latest master

clacroix12 (approved):

vasukulkarni (approved):

PR #3219, opened 2020-10-29
WIP: Add UI function for storage cluster deletion and a basic UI test

ebondare:


ocs_ci/ocs/ui_navigation.py
```diff
@@ -0,0 +1,127 @@
...
+from selenium.webdriver.common.keys import Keys
+from selenium.common.exceptions import NoSuchElementException
+
+from ocs_ci.utility.utils import get_kubeadmin_password, run_cmd
+
+logger = logging.getLogger(__name__)
+
+
+def click_element_by_id(driver, element_id):
+    '''
```
ebenahar, line 16: please change to `"""`

petr-balogh:
Please provide some more info in the commit message. And we can copy back to first comment here as well which is created usually from commit message but here I see just: `No description provided.`

PR #3246, opened 2020-11-02
Use api key in Prometheus API communication on IBM Cloud

fbalak:
- For successful Prometheus API communication on IBM Cloud needs to be set `ibm_apikey` in `AUTH` configuration.

### DIFF
--- a/ocs_ci/utility/prometheus.py
+++ b/ocs_ci/utility/prometheus.py
@@ -301,7 +301,7 @@ class PrometheusAPI(object):
     _user = None
     _password = None
     _endpoint = None
-    _cacert = None
+    _cacert = False
 
     def __init__(self, user=None, password=None):
         """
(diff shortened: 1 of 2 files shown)


mbukatov (dismissed):

ocs_ci/utility/prometheus.py
```diff
@@ -310,17 +310,23 @@ def __init__(self, user=None, password=None):
...
-            filename = os.path.join(
-                config.ENV_DATA['cluster_path'],
-                config.RUN['password_location']
-            )
-            with open(filename) as f:
-                password = f.read()
-        self._password = password
+        if config.ENV_DATA['platform'].lower() == 'ibm_cloud':
+            self._user = user or 'apikey'
+            self._password = password or config.AUTH.get('ibm_apikey')
```
petr-balogh, line 315: I am automating IBM Cloud deployment and the auth data will be in this structure:
```yaml
---
AUTH:
  ibmcloud:
    api_key: "XXXX"
    account_id: "XXXXX"
```

So I suggest to change this line to:
```suggestion
            self._password = password or config.AUTH["ibmcloud"]['api_key']
```

As it's mandatory value for IBM cloud run, I suggest to use it directly through key instead of get.
And if the value will be missing we will see keyError and we will see what value in config we are missing.

mbukatov (approved):

petr-balogh (approved):

PR #3247, opened 2020-11-02
Retry fetching Noobaa endpoint pods data in case of KeyError

### LINKED ISSUE #3248: get_ready_noobaa_endpoint_count() sometimes fails with KeyError: 'containerStatuses'
Seen in RUN ID: 1604321062

ebenahar:
Fixes #3248

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2989,7 +2989,7 @@ def get_ready_noobaa_endpoint_count(namespace):
     ready_count = 0
     for ep_info in pods_info:
         container_statuses = ep_info.get('status', {}).get('containerStatuses')
-        if container_statuses and container_statuses is not []:
+        if container_statuses is not None and len(container_statuses) > 0:
             if container_statuses[0].get('ready'):
                 ready_count += 1
     return ready_count
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2988,8 +2988,10 @@ def get_ready_noobaa_endpoint_count(namespace):
     pods_info = get_pods_having_label(label=constants.NOOBAA_ENDPOINT_POD_LABEL, namespace=namespace)
     ready_count = 0
     for ep_info in pods_info:
-        if ep_info['status']['containerStatuses'][0]['ready']:
-            ready_count += 1
+        container_statuses = ep_info.get('status', {}).get('containerStatuses')
+        if container_statuses and container_statuses is not []:
+            if container_statuses[0].get('ready'):
+                ready_count += 1
     return ready_count
 
 
@@ -3010,7 +3012,6 @@ def nb_ensure_endpoint_count(request):
     if float(config.ENV_DATA['ocs_version']) < 4.6:
         noobaa = OCP(kind='noobaa', namespace=namespace)
         resource = noobaa.get()['items'][0]
-        resource_name = resource['metadata']['name']
         endpoints = resource.get('spec', {}).get('endpoints', {})
 
         if endpoints.get('minCount', -1) != min_ep_count:


tests/conftest.py
```diff
@@ -2981,6 +2981,7 @@ def finalizer():
     request.addfinalizer(finalizer)
 
 
+@retry.retry(KeyError)
```
nb-ohad, line 2984: This is bad, a retry already happens from the calling code in `nb_ensure_endpoint_count`, we just need to not throw and return False

nb-ohad (changes requested):

tests/conftest.py
ebenahar: good point. Fixed now

nb-ohad (approved):

nb-ohad (dismissed):

tests/conftest.py
```diff
@@ -2988,8 +2988,10 @@ def get_ready_noobaa_endpoint_count(namespace):
     pods_info = get_pods_having_label(label=constants.NOOBAA_ENDPOINT_POD_LABEL, namespace=namespace)
     ready_count = 0
     for ep_info in pods_info:
-        if ep_info['status']['containerStatuses'][0]['ready']:
-            ready_count += 1
+        container_statuses = ep_info.get('status', {}).get('containerStatuses')
+        if container_statuses and container_statuses is not []:
+            if container_statuses[0].get('ready'):
```
RazTamir, line 2993: the above 'if' statement is if the container_statusses is not a list and here you are accessing it as a list.

ebenahar: the condition is if it's not an empty list

mbukatov: If `container_statuses` is `[]`, then `bool(container_statuses)` is false, so that the 2nd check isn't necessary. I would rather used something like `len(container_statuses) > 0` if the intention is to process nonempty lists.

mbukatov (changes requested):

tests/conftest.py
OdedViner: You can change:
if len(container_statuses)>0:
    if container_statuses[0].get('ready'):

ebenahar: @mbukatov @OdedViner 
the second part of the condition (`if container_statuses`) is to for cases in which `container_statuses` is actually `None`.

but what if `container_statuses` is None?

```
>>> container_statuses = None
>>> len(container_statuses)>0
  File "<input>", line 1, in <module>
    len(container_statuses)>0
TypeError: object of type 'NoneType' has no len()
```

nb-ohad: @OdedViner 

`container_statuses` might be `None`, empty lists, or a non-empty list. 
The condition should filter only non-empty lists

`len(arr) > 0` will not fit for the condition because `len(None)` will throw.

Which mean we need: `if container_statuses and len(container_statuses) > 0` which is eqivulent to the current condition:
`if container_statuses and container_statuses is not []` in the current context.

I think The condition is OK the way it is.

ebenahar:
The failures in the above run of PR validation are not related to the change done here, marking as verified

tests/conftest.py
mbukatov: Ah, right. In such case, it would be more readable if written as `if container_statuses is not None and len(container_statuses) > 0:` ...

ebenahar: sure, will update

ebenahar: done

OdedViner (approved):
LGTM

mbukatov (approved):

petr-balogh (approved):

Issue #2643, opened 2020-08-05
Code in ocs_ci module should not use assert statements


mbukatov:
Instead of assert statements, code in ocs_ci module should raise appropriate exception.

This way, code in ocs_ci or in the test can appropriately react on a failure, instead of a uncontrolled failure.

PR #3249, opened 2020-11-02
Function to check compression and replication

shyRozen:
In order to test KNIP 1462 multiple storage class added function to validate compression and replication at pool level

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -29,6 +29,7 @@ from ocs_ci.utility.utils import (
 from ocs_ci.ocs.utils import get_pod_name_by_pattern
 from ocs_ci.framework import config
 from ocs_ci.ocs import ocp, constants, exceptions
+from ocs_ci.ocs.exceptions import PoolNotFound
 from ocs_ci.ocs.resources.pvc import get_all_pvc_objs
 
 logger = logging.getLogger(__name__)
@@ -901,6 +902,91 @@ def get_osd_utilization():
         osd_filled[osd["name"]] = osd["utilization"]
 
     return osd_filled
+
+
+def get_ceph_df_detail():
+    """
+    Get ceph osd df detail
+
+    Returns:
+         dict: 'ceph df details' command output
+
+    """
+    ceph_cmd = "ceph df detail"
+    ct_pod = pod.get_ceph_tools_pod()
+    return ct_pod.exec_ceph_cmd(ceph_cmd=ceph_cmd)
+
+
+def validate_replica_data(pool_name, replica):
+    """
+    Check if data is replica 2 or 3
+
+    Args:
+        replica (int): size of the replica(2,3)
+        pool_name (str): name of the pool to check replica
+
+    Returns:
+        Bool: True if replicated data size is meet rep config and False if dont
+
+    """
+
+    ceph_df_detail_output = get_ceph_df_detail()
+    pool_list = ceph_df_detail_output.get("pools")
+    for pool in pool_list:
+        if pool.get("name") == pool_name:
+            logger.info(f"{pool_name}")
+            stored = pool["stats"]["stored"]
+            byte_used = pool["stats"]["bytes_used"]
+            compress_bytes_used = pool["stats"]["compress_bytes_used"]
+            compress_under_bytes = pool["stats"]["compress_under_bytes"]
+            byte_used = byte_used + compress_under_bytes - compress_bytes_used
+            store_ratio = byte_used / stored
+            if (replica + 0.2) > store_ratio > (replica - 0.2):
+                logger.info(f"pool {pool_name} meet rep {replica} size")
+                return True
+            else:
+                logger.info(
+                    f"pool {pool_name} meet do not meet rep {replica}"
+                    f" size Store ratio is {store_ratio}"
+                )
+
+                return False
+    raise PoolNotFound(f"Pool {pool_name} not found on cluster")
+
+
+def validate_compression(pool_name):
+    """
+    Check if data was compressed
+
+    Args:
+        pool_name (str): name of the pool to check replica
+
+    Returns:
+        bool: True if compression works. False if not
+
+    """
+    ceph_df_detail_output = get_ceph_df_detail()
+    pool_list = ceph_df_detail_output.get("pools")
+    for pool in pool_list:
+        if pool.get("name") == pool_name:
+            logger.info(f"{pool_name}")
+            byte_used = pool["stats"]["bytes_used"]
+            compress_bytes_used = pool["stats"]["compress_bytes_used"]
+            compress_under_bytes = pool["stats"]["compress_under_bytes"]
+            all_byte_used = byte_used + compress_under_bytes - compress_bytes_used
+            compression_ratio = byte_used / all_byte_used
+            logger.info(f"this is the comp_ratio {compression_ratio}")
+            if 0.6 < compression_ratio:
+                logger.info(
+                    f"Compression ratio {compression_ratio} is " f"larger than 0.6"
+                )
+                return True
+            else:
+                logger.info(
+                    f"Compression ratio {compression_ratio} is " f"smaller than 0.6"
+                )
+                return False
+    raise PoolNotFound(f"Pool {pool_name} not found on cluster")
 
 
 def validate_osd_utilization(osd_used=80):
--- a/ocs_ci/ocs/exceptions.py
+++ b/ocs_ci/ocs/exceptions.py
@@ -221,6 +221,10 @@ class StorageNotSufficientException(Exception):
     pass
 
 
+class PoolNotFound(Exception):
+    pass
+
+
 class MemoryNotSufficientException(Exception):
     pass
 
--- a/tests/manage/storageclass/test_new_sc_rbd_replica2_3.py
+++ b/tests/manage/storageclass/test_new_sc_rbd_replica2_3.py
@@ -4,14 +4,22 @@ import pytest
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.resources.pod import get_fio_rw_iops
 from ocs_ci.framework.testlib import ManageTest, tier1
-from ocs_ci.framework.pytest_customization.marks import skipif_external_mode
-from ocs_ci.ocs.cluster import get_percent_used_capacity
+from ocs_ci.framework.pytest_customization.marks import (
+    skipif_external_mode,
+    skipif_ocs_version,
+)
+from ocs_ci.ocs.cluster import (
+    get_percent_used_capacity,
+    validate_compression,
+    validate_replica_data,
+)
 
 log = logging.getLogger(__name__)
 
 
 @tier1
 @skipif_external_mode
+@skipif_ocs_version("<4.6")
 class TestCreateNewScWithNeWRbDPool(ManageTest):
     """
     Create a new  Storage Class on a new rbd pool with
@@ -63,3 +71,7 @@ class TestCreateNewScWithNeWRbDPool(ManageTest):
             f"Cluster used space with replica size {replica}, "
             f"compression mode {compression}={cluster_used_space}"
         )
+        cbp_name = sc_obj.get().get("parameters").get("pool")
+        if compression != "none":
+            validate_compression(cbp_name)
+        validate_replica_data(cbp_name, replica)


ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +912,61 @@ def get_osd_utilization():
     return osd_filled
 
 
+def get_ceph_df_detail():
+    """
```
ebenahar, line 916: please change to sphynx docstring style - https://www.jetbrains.com/help/pycharm/using-docstrings-to-specify-types.html#example

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +912,61 @@ def get_osd_utilization():
...
+        if pool.get('name') == pool_name:
+            logger.info(f"{pool_name}")
+            stored = pool.get('stats').get('stored')
+            byte_used = pool.get('stats').get('bytes_used')
+            compress_bytes_used = pool.get('stats').get('compress_bytes_used')
+            compress_under_bytes = pool.get('stats').get('compress_under_bytes')
+            byte_used = byte_used + compress_under_bytes - compress_bytes_used
+            store_ratio = byte_used / stored
+            assert (replica+0.2) > store_ratio > (replica-0.2), f"pool {pool_name} do not meet rep {replica} size"
+            return True
```
ebenahar, line 945: since there is no return False in this function, I would remove the `return True` statement and change this function to not return anything

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +912,61 @@ def get_osd_utilization():
...
+            logger.info(f"{pool_name}")
+            byte_used = pool.get('stats').get('bytes_used')
+            compress_bytes_used = pool.get('stats').get('compress_bytes_used')
+            compress_under_bytes = pool.get('stats').get('compress_under_bytes')
+            all_byte_used = byte_used + compress_under_bytes - compress_bytes_used
+            assert compress_bytes_used != 0 or compress_under_bytes !=0, f"pool {pool_name} have stats with no compression"
+            compression_ratio = byte_used / all_byte_used
+            logger.info(f"this is the comp_ratio {compression_ratio}")
+            assert 0.6 < compression_ratio, f"Ratio {compression_ratio} is smaller than 0.6"
+            return True
```
ebenahar, line 967: same here with `return True`

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +912,61 @@ def get_osd_utilization():
...
+    for pool in output.get('pools'):
+        if pool.get('name') == pool_name:
+            logger.info(f"{pool_name}")
+            stored = pool.get('stats').get('stored')
+            byte_used = pool.get('stats').get('bytes_used')
+            compress_bytes_used = pool.get('stats').get('compress_bytes_used')
+            compress_under_bytes = pool.get('stats').get('compress_under_bytes')
+            byte_used = byte_used + compress_under_bytes - compress_bytes_used
+            store_ratio = byte_used / stored
+            assert (replica+0.2) > store_ratio > (replica-0.2), f"pool {pool_name} do not meet rep {replica} size"
```
mbukatov, line 944: I [advise against using assert statements in ocs_ci module](https://github.com/red-hat-storage/ocs-ci/issues/2643), so here I would rather:
- stored the validation result in a variable: `result = (replica+0.2) > store_ratio > (replica-0.2)`
- logged the result in info or debug log level (reusing the assert description "pool do not meet ..." as a log message)
- returned the result

Whether this should then be as a test failure via assert depends on a test case, where one could write eg. `assert validate_replica_data(pool_name, replica)` ...

mbukatov (changes requested):

ocs_ci/ocs/cluster.py
shyRozen: Fixed

ocs_ci/ocs/cluster.py
shyRozen: Made it return true or false because of Martin comment to have the assert in the testcase itself

ocs_ci/ocs/cluster.py
shyRozen: Fixed

ocs_ci/ocs/cluster.py
shyRozen: Added return true or false. When testcase will be submitted will use assert in the testcase body.

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +912,69 @@ def get_osd_utilization():
...
+    """
+    output = get_ceph_df_detail()
+    for pool in output.get('pools'):
+        if pool.get('name') == pool_name:
+            logger.info(f"{pool_name}")
+            byte_used = pool.get('stats').get('bytes_used')
+            compress_bytes_used = pool.get('stats').get('compress_bytes_used')
+            compress_under_bytes = pool.get('stats').get('compress_under_bytes')
+            all_byte_used = byte_used + compress_under_bytes - compress_bytes_used
+            assert compress_bytes_used != 0 or compress_under_bytes !=0, f"pool {pool_name} have stats with no compression"
```
mbukatov, line 967: I would rather raised an exception instead.

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +912,69 @@ def get_osd_utilization():
...
+            compression_ratio = byte_used / all_byte_used
+            logger.info(f"this is the comp_ratio {compression_ratio}")
+            if 0.6 < compression_ratio:
+                f"Compression ratio {compression_ratio} is larger than 0.6"
+                return True
+            else:
+                f"Compression ratio {compression_ratio} is smaller than 0.6"
```
mbukatov, line 974: I guess that these lines are missing `logger.info()` (or `debug`) calls.

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +912,69 @@ def get_osd_utilization():
...
+            compress_under_bytes = pool.get('stats').get('compress_under_bytes')
+            byte_used = byte_used + compress_under_bytes - compress_bytes_used
+            store_ratio = byte_used / stored
+            if (replica+0.2) > store_ratio > (replica-0.2):
+                logger.info(f"pool {pool_name} meet rep {replica} size")
+                return True
+            else:
+                logger.info(f"pool {pool_name} meet do not meet rep {replica} size"
+                            f"Store ratio is {store_ratio}")
+                return False
```
mbukatov, line 949: In both functions, we need to handle a case where the pool with given pool_name is not found. I would suggest to raise an exception noting that the pool was not found.

mbukatov (changes requested):

ocs_ci/ocs/cluster.py
shyRozen: This assert is a leftover. Removed it and raising or asserting will happen in the testcase body (as requested by you in another conversation)

ocs_ci/ocs/cluster.py
shyRozen: added

ocs_ci/ocs/cluster.py
shyRozen: Added exception when pool is not found

ocs_ci/ocs/cluster.py
ebenahar: still not fixed

mbukatov: @shyRozen See https://sphinxcontrib-napoleon.readthedocs.io/en/latest/ how the expected doc string format looks like.

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +912,74 @@ def get_osd_utilization():
...
+
+
+def validate_replica_data(pool_name, replica):
+    """
+    Check if data is replica 2 or 3
+    :param int replica: size of the replica(2,3)
+    :param string pool_name: name of the pool to check replica
+    :return: true if replica is verified, false if not
+    """
+    from ocs_ci.ocs.exceptions import PoolNotFound
```
mbukatov, line 933: Move this import in the beginning of the python file, among other imports.

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +912,74 @@ def get_osd_utilization():
...
+    for pool in output.get('pools'):
+        if pool.get('name') == pool_name:
+            logger.info(f"{pool_name}")
+            stored = pool.get('stats').get('stored')
+            byte_used = pool.get('stats').get('bytes_used')
+            compress_bytes_used = pool.get('stats').get('compress_bytes_used')
+            compress_under_bytes = pool.get('stats').get('compress_under_bytes')
+            byte_used = byte_used + compress_under_bytes - compress_bytes_used
+            store_ratio = byte_used / stored
+            if (replica+0.2) > store_ratio > (replica-0.2):
```
mbukatov, line 944: This will likely fail on a flake8 check.

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +912,74 @@ def get_osd_utilization():
...
+                logger.info(f"pool {pool_name} meet rep {replica} size")
+                return True
+            else:
+                logger.info(f"pool {pool_name} meet do not meet rep {replica} size"
+                            f"Store ratio is {store_ratio}")
+                return False
+        if pool == output.get('pools')[-1]:
```
mbukatov, line 951: What does `pool == output.get('pools')[-1]` condition mean? It would be more readable, if we just have a boolean variable about whether we found the pool with poo_name, and then later here checked if that is the case and raise an exception accordingly.

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +912,74 @@ def get_osd_utilization():
...
+    """
+    from ocs_ci.ocs.exceptions import PoolNotFound
+    output = get_ceph_df_detail()
+    for pool in output.get('pools'):
+        if pool.get('name') == pool_name:
+            logger.info(f"{pool_name}")
+            stored = pool.get('stats').get('stored')
+            byte_used = pool.get('stats').get('bytes_used')
+            compress_bytes_used = pool.get('stats').get('compress_bytes_used')
+            compress_under_bytes = pool.get('stats').get('compress_under_bytes')
```
mbukatov, line 941: This is a general complain, so feel free to ignore it, but I don't understand the approach of using `get()` method to fetch value in a dictionary everywhere. This will make sense only if we actually care to handle the case when it returns None. If we don't care (eg. because a test or function should fail because of the missing key), using `get` will make the traceback more opaque, noting "None doesn't have get method", instead of "dict such and such doesn't have "stats" key.

ocs_ci/ocs/cluster.py
shyRozen: It means if the pool = the last pool in the list. Why we have to keep another variable to track this after the loop when I can handle it with a simple statement. I can simplify before the loop the expression like:
```pool_list = output.get('pools')```
and then call
```if pool == pool_list[-1]```
will be much more readable
wdyt?

File: ocs_ci/ocs/cluster.py (hunk 11 above)
Comment (shyRozen) [reply ocs_ci/ocs/cluster.py:941]: Do you mean using 
pool['stats']['compress_bytes_used'] ? This will make more efficient to debug?

File: ocs_ci/ocs/cluster.py (hunk 9 above)
Comment (shyRozen) [reply ocs_ci/ocs/cluster.py]: Fixed the white space around operators

File: ocs_ci/ocs/cluster.py (hunk 8 above)
Comment (shyRozen) [reply ocs_ci/ocs/cluster.py:937]: moved the import to the start of the file.

File: ocs_ci/ocs/cluster.py (hunk 10 above)
Comment (shyRozen) [reply ocs_ci/ocs/cluster.py]: Change the code to my suggestion. Not resolving the issue till your thoughts on it.

File: ocs_ci/ocs/cluster.py (hunk 1 above)
Comment (shyRozen) [reply ocs_ci/ocs/cluster.py:917]: I think now it should be ok. Thanks @mbukatov for the clarification.

File: ocs_ci/ocs/cluster.py (hunk 10 above)
Comment (mbukatov) [reply ocs_ci/ocs/cluster.py]: Actually, the most readable and simple would be to just raise an exception without the if statement. Code after the if won't be executed if we found the pool.

File: ocs_ci/ocs/cluster.py (hunk 11 above)
Comment (mbukatov) [reply ocs_ci/ocs/cluster.py]: Yes.

File: ocs_ci/ocs/cluster.py (hunk 8 above)
Comment (mbukatov) [reply ocs_ci/ocs/cluster.py:937]: I still see the import there.

File: ocs_ci/ocs/cluster.py [hunk 12]
```diff
@@ -912,6 +913,85 @@ def get_osd_utilization():
...
+            if 0.6 < compression_ratio:
+                logger.info(f"Compression ratio {compression_ratio} is "
+                            f"larger than 0.6")
+                return True
+            else:
+                logger.info(f"Compression ratio {compression_ratio} is "
+                            f"smaller than 0.6")
+                return False
+        if pool == pool_list[-1]:
+            raise PoolNotFound(f"Pool {pool_name} not found on cluster")
```
mbukatov, line 992: As I noted in another thread, let's just omit the if and raise the exception directly. The code there won't be executed if the pool is found.

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +913,85 @@ def get_osd_utilization():
...
+            if (replica + 0.2) > store_ratio > (replica - 0.2):
+                logger.info(f"pool {pool_name} meet rep {replica} size")
+                return True
+            else:
+                logger.info(f"pool {pool_name} meet do not meet rep"
+                            f" {replica} size"
+                            f"Store ratio is {store_ratio}")
+                return False
+        if pool == pool_list[-1]:
+            raise PoolNotFound(f"Pool {pool_name} not found on cluster")
```
mbukatov, line 960: As I noted in another thread, let's just omit the if and raise the exception directly. The code there won't be executed if the pool is found.

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +913,85 @@ def get_osd_utilization():
...
+
+def validate_replica_data(pool_name, replica):
+    """
+    Check if data is replica 2 or 3
+    :param replica: size of the replica(2,3)
+    :type replica: int
+    :param pool_name: name of the pool to check replica
+    :type pool_name: str
+    :return: true if replica is verified, false if not
+    :rtype: bool
```
mbukatov, line 935: I see no change in the docstring. What we expect would look like:
```
Args:
    replica (int): size of the replica (eg. 2 or 3)
    pool_name (str): name of the pool to check replica

Returns:
    true if replica is verified, false if not
```

mbukatov: See [Napoleon - Marching toward legible docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/) for few examples.

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +913,85 @@ def get_osd_utilization():
...
+
+
+def validate_compression(pool_name):
+    """
+    Check if data was compressed
+    :param pool_name: name of the pool to check replica
+    :type pool_name: str
+    :return: True if compression ratio is right, false if compression
+    ratio is wrong
+    :rtype: bool
```
mbukatov, line 970: Docstring style, see above.

ocs_ci/ocs/cluster.py
```diff
@@ -912,6 +913,85 @@ def get_osd_utilization():
     return osd_filled
 
 
+def get_ceph_df_detail():
+    """
+    Get ceph osd df detail
+    :return: dict returned from cmd
```
mbukatov, line 919: Docstring style, see above.

ocs_ci/ocs/cluster.py
mbukatov: That is because when this fails, exception will look like `KeyError: 'bytes_used'` on given dict, which will immediately tell us that 'bytes_used' key is not present in given dict. Compare that with exception `AttributeError: 'NoneType' object has no attribute 'get'` which doesn't tell us much at first sight.

And off curse, if you can recover from the error, it makes sense to use `.get()` methods, but then you also need to check whether the value is None, and log and process the error properly.

Here I agree that it doesn't make much sense to check whether the keys are there or not, because this code can't work without these keys anyway, and if we notice these keys are missing, we will have to raise exception and halt everything anyway.

ocs_ci/ocs/cluster.py
shyRozen: But the loop have  states that it doesn't find the pool because it is iterating over a list to find the pool and most of the time the loop have state which it doesn't find the pool till it is found, so how can we drop the if statement?

ocs_ci/ocs/cluster.py
shyRozen: Done

ocs_ci/ocs/cluster.py
shyRozen: Finally understood. Remove if and moved it out of the loop

ocs_ci/ocs/cluster.py
shyRozen: Changed

ocs_ci/ocs/cluster.py
shyRozen: remove the if and raise outside loop

ocs_ci/ocs/cluster.py
shyRozen: remove the if and raise outside loop

mbukatov (approved):

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,80 @@ def get_osd_utilization():
...
```
ebenahar, line 910:
```suggestion
    """
    Get ceph osd df detail
    
    Returns: 
        dict: 'ceph df details' command output
    
    """
```

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,80 @@ def get_osd_utilization():
...
+    ceph_cmd="ceph df detail"
+    ct_pod = pod.get_ceph_tools_pod()
+    output = ct_pod.exec_ceph_cmd(ceph_cmd=ceph_cmd)
+    return output
+
+
+def validate_replica_data(pool_name, replica):
+    """
+    Check if data is replica 2 or 3
+    Args:
```
ebenahar, line 921: add an empty line below

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,80 @@ def get_osd_utilization():
...
+        replica (int): size of the replica(2,3)
+        pool_name: name of the pool to check replica
```
ebenahar, line 923: add an empty line below

ocs_ci/ocs/cluster.py
shyRozen: Added

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,84 @@ def get_osd_utilization():
     return osd_filled
 
 
+def get_ceph_df_detail():
+    """
+    Get ceph osd df detail
```
ebenahar, line 909: add an empty line  below

ocs_ci/ocs/cluster.py
shyRozen: Added

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,84 @@ def get_osd_utilization():
...
+    Returns:
+         dict returned from cmd
```
ebenahar, line 911: dict: The returned output from 'ceph df detail' command

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,84 @@ def get_osd_utilization():
...
+    """
+    ceph_cmd="ceph df detail"
+    ct_pod = pod.get_ceph_tools_pod()
+    output = ct_pod.exec_ceph_cmd(ceph_cmd=ceph_cmd)
+    return output
+
+
+def validate_replica_data(pool_name, replica):
+    """
+    Check if data is replica 2 or 3
```
ebenahar, line 922: add an empty line below

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,84 @@ def get_osd_utilization():
...
+    Args:
+        replica (int): size of the replica(2,3)
+        pool_name: name of the pool to check replica
+
+    Returns:
+        True if replicated data size is meet rep config and False if dont
```
ebenahar, line 928: bool: True if the replicated data size is meet rep config and False otherwise

ocs_ci/ocs/cluster.py
shyRozen: Done

ocs_ci/ocs/cluster.py
shyRozen: Done

ocs_ci/ocs/cluster.py
shyRozen: Done

ocs_ci/ocs/cluster.py
shyRozen: Done

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,87 @@ def get_osd_utilization():
...
+    return output
+
+
+def validate_replica_data(pool_name, replica):
+    """
+    Check if data is replica 2 or 3
+
+    Args:
+        replica (int): size of the replica(2,3)
+        pool_name: name of the pool to check replica
```
ebenahar, line 927: pool_name (str):

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,87 @@ def get_osd_utilization():
...
+                return False
+    raise PoolNotFound(f"Pool {pool_name} not found on cluster")
+
+
+def validate_compression(pool_name):
+    """
+    Check if data was compressed
+
+    Args:
+    pool_name (str): name of the pool to check replica
```
ebenahar, line 961:
```suggestion
    Args:
        pool_name (str): name of the pool to check replica
```

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,87 @@ def get_osd_utilization():
...
+
+    Returns:
+         bool: True if replica much. False if not
```
ebenahar, line 964: add an empty line below

Avilir (approved):

ocs_ci/ocs/cluster.py
shyRozen: Done

ocs_ci/ocs/cluster.py
shyRozen: Fixed

ocs_ci/ocs/cluster.py
shyRozen: fixed

ocs_ci/ocs/cluster.py
shyRozen: Fixed

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,88 @@ def get_osd_utilization():
...
+
+def validate_compression(pool_name):
+    """
+    Check if data was compressed
+
+    Args:
+        pool_name (str): name of the pool to check replica
+
+    Returns:
+        bool: True if replica much. False if not
```
ebenahar, line 964: this doesn't seem relevant here

shyRozen: Fixed to compressoin

ebenahar (approved):

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,88 @@ def get_osd_utilization():
...
+
+    Returns:
+         dict: 'ceph df details' command output
+         
+    """
+    ceph_cmd="ceph df detail"
+    ct_pod = pod.get_ceph_tools_pod()
+    output = ct_pod.exec_ceph_cmd(ceph_cmd=ceph_cmd)
+    return output
+
```
OdedViner, line 919: you can change:
return ct_pod.exec_ceph_cmd(ceph_cmd=ceph_cmd)

ocs_ci/ocs/cluster.py
OdedViner: same

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,88 @@ def get_osd_utilization():
...
+            byte_used = byte_used + compress_under_bytes - compress_bytes_used
+            store_ratio = byte_used / stored
+            if (replica + 0.2) > store_ratio > (replica - 0.2):
+                logger.info(f"pool {pool_name} meet rep {replica} size")
+                return True
+            else:
+                logger.info(f"pool {pool_name} meet do not meet rep"
+                            f" {replica} size"
+                            f"Store ratio is {store_ratio}")
+                return False
```
OdedViner, line 952: PEP8 standard:
You can change:

logger.info(f"pool {pool_name} meet do not meet rep"
                  f" {replica} size ,Store ratio is {store_ratio}"
)

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,88 @@ def get_osd_utilization():
...
+            byte_used = pool["stats"]["bytes_used"]
+            compress_bytes_used = pool["stats"]["compress_bytes_used"]
+            compress_under_bytes = pool["stats"]["compress_under_bytes"]
+            all_byte_used = byte_used + compress_under_bytes - compress_bytes_used
+            compression_ratio = byte_used / all_byte_used
+            logger.info(f"this is the comp_ratio {compression_ratio}")
+            if 0.6 < compression_ratio:
+                logger.info(f"Compression ratio {compression_ratio} is "
+                            f"larger than 0.6")
+                return True
```
OdedViner, line 981: PEP8:
you can change 
                logger.info(f"Compression ratio {compression_ratio} is "
                            f"larger than 0.6"
)

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,88 @@ def get_osd_utilization():
...
+            else:
+                logger.info(f"Compression ratio {compression_ratio} is "
+                            f"smaller than 0.6")
```
OdedViner, line 984: same

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,88 @@ def get_osd_utilization():
...
+    Check if data was compressed
+
+    Args:
+        pool_name (str): name of the pool to check replica
+
+    Returns:
+        bool: True if compression works. False if not
+            
+    """
+    output = get_ceph_df_detail()
```
OdedViner, line 967: you can change the name of var `output` . for example `df_detail_output`

ocs_ci/ocs/cluster.py
shyRozen: Changed

ocs_ci/ocs/cluster.py
shyRozen: Does the rows need to be aligned by PEP8 standards? The IDE is marks it as not PEP8 standrad when I change the rows to be aligned like that.

ocs_ci/ocs/cluster.py
```diff
@@ -903,6 +904,87 @@ def get_osd_utilization():
...
+    for pool in pool_list:
+        if pool.get("name") == pool_name:
+            logger.info(f"{pool_name}")
+            byte_used = pool["stats"]["bytes_used"]
+            compress_bytes_used = pool["stats"]["compress_bytes_used"]
+            compress_under_bytes = pool["stats"]["compress_under_bytes"]
+            all_byte_used = byte_used + compress_under_bytes - compress_bytes_used
+            compression_ratio = byte_used / all_byte_used
+            logger.info(f"this is the comp_ratio {compression_ratio}")
+            if 0.6 < compression_ratio:
```
OdedViner, line 977: Suggestion:
logger.info(f"Verify the comp_ratio {compression_ratio} larger than 0.6")
return True if 0.6 < compression_ratio else False

ocs_ci/ocs/cluster.py
shyRozen: Do you mean only about the parentheses? Like it should be:
```logger.info(f"pool {pool_name} meet do not meet rep"
                  f" {replica} size ,Store ratio is {store_ratio}"
)```

shyRozen: Do you mean only the parentheses in a new line?
like that?
```logger.info(f"pool {pool_name} meet do not meet rep"
                     f" {replica} size ,Store ratio is {store_ratio}"
)```

ocs_ci/ocs/cluster.py
ebenahar: same :)

ocs_ci/ocs/cluster.py
shyRozen: It is just the output of the last command and doesn't relates to the next command. After the if there are loggers related to it.

ocs_ci/ocs/cluster.py
shyRozen: Changed

ocs_ci/ocs/cluster.py
shyRozen: Changed

ocs_ci/ocs/cluster.py
shyRozen: Changed

ocs_ci/ocs/cluster.py
shyRozen: Changed

OdedViner (approved):
LGTM

ebenahar (approved):

PR #3253, opened 2020-11-03
Add OCS uninstall to cluster teardown 

annagitel:
Signed-off-by: annagitel <asandler@redhat.com>

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -47,6 +47,7 @@ from ocs_ci.ocs.resources.pod import (
     get_all_pods,
     validate_pods_are_respinned_and_running_state
 )
+from ocs_ci.ocs.uninstall import uninstall_ocs
 from ocs_ci.ocs.utils import (
     setup_ceph_toolbox, collect_ocs_logs
 )
@@ -181,7 +182,7 @@ class Deployment(object):
         nodes = ocp.OCP(kind='node').get().get('items', [])
         worker_nodes = [
             node for node in nodes if "node-role.kubernetes.io/worker"
-            in node['metadata']['labels']
+                                      in node['metadata']['labels']
         ]
         if not worker_nodes:
             raise UnavailableResourceException("No worker node found!")
@@ -600,7 +601,7 @@ class Deployment(object):
             _ocp = ocp.OCP(kind='node')
             _ocp.exec_oc_cmd(
                 command=f"annotate namespace {defaults.ROOK_CLUSTER_NAMESPACE} "
-                f"{constants.NODE_SELECTOR_ANNOTATION}"
+                        f"{constants.NODE_SELECTOR_ANNOTATION}"
             )
 
     def deployment_with_ui(self):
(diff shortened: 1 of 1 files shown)


ocs_ci/deployment/deployment.py
```diff
@@ -854,6 +855,12 @@ def destroy_cluster(self, log_level="DEBUG"):
                 logger.info("Destroy of OCP not implemented yet.")
         else:
             self.ocp_deployment = self.OCPDeployment()
+            try:
+                uninstall_ocs()
+                # TODO - add ocs uninstall validation function call
+                logger.info("OCS uninstalled succesfully")
+            except Exception:
+                logger.warning("Failed to uninstall OCS, resuming teardown")
```
ebenahar, line 863: please also log the exception

ocs_ci/deployment/deployment.py
```diff
@@ -854,6 +855,13 @@ def destroy_cluster(self, log_level="DEBUG"):
                 logger.info("Destroy of OCP not implemented yet.")
         else:
             self.ocp_deployment = self.OCPDeployment()
+            try:
+                uninstall_ocs()
+                # TODO - add ocs uninstall validation function call
+                logger.info("OCS uninstalled succesfully")
+            except Exception:
+                logger.warning(f"Failed to uninstall OCS: {Exception}")
```
ebenahar, line 863: I would change to:
```python
logger.error(f"Failed to uninstall OCS. Exception is: \n{Exception}")
```

jimdevops19 (approved):
LGTM

ebenahar (approved):

petr-balogh:
@ebenahar @annagitel this is breaking the destroy:
https://github.com/red-hat-storage/ocs-ci/issues/3274

Opened issue and will send quick fix FYI

ocs_ci/deployment/deployment.py
```diff
@@ -854,6 +855,13 @@ def destroy_cluster(self, log_level="DEBUG"):
                 logger.info("Destroy of OCP not implemented yet.")
         else:
             self.ocp_deployment = self.OCPDeployment()
+            try:
+                uninstall_ocs()
```
vasukulkarni, line 859: we discussed about this being in the deployment code path, It will affect all runs - jenkins, test runs and local test runs. I think it has to be separate test case but some challenges exists when someone has to pause the jobs based. The tier name should be decided or it has to be different stage, @petr-balogh said he is going to check on options.

petr-balogh: This approach is not good because it will make destroy job running longer time cause of this uninstall.
Here is my input:

1) this should be done as separate test case like test_uninstall so we will run it only when we want and when we use the marker for it.
2) we need to report the issues to user when the uninstall will fail, only this way we will notice it as this case it will just fail, log on error level but honestly, who will check console output of successful job?
3) It cannot run on tier execution as last test case as then it will not work for our option pause before teardown.
4) We would like to run uninstall test probably after all production jobs
5) We need to run in destroy stage as it's designed now but we have 2 options 
    - try to run in one ocs-ci execution together with destroy + send email report from this execution.
    - or run 2 executions of ocs-ci - one for uninstall test only when production job with email report. Continue to destroy run even if uninstall execution will fail - the job itself will pause and email us if destroy job failed - so this case we know that destroy didn't succeeded and require manual intervention.
6) Some alignment from Jenkins job will be needed for sure.

So if we can set mtg @ebenahar @annagitel to discuss my points above.

PR #3273, opened 2020-11-05
PVC snapshot, clone, expansion combination test

jilju:
This test verifies that the clone, snapshot and parent PVCs are independent and any operation in one will not impact the other.

OCS-2408 - FT-PVC snapshot, clone, expansion combination test

### DIFF
--- a/tests/manage/pv_services/test_expansion_snapshot_clone.py
+++ b/tests/manage/pv_services/test_expansion_snapshot_clone.py
@@ -358,18 +358,19 @@ class TestExpansionSnapshotClone(ManageTest):
         log.info("Verified: Pods reached Running state")
 
         # Expand PVCs
+        pvc_size_expand_4 = pvc_size_expand_3 + 2
         log.info(
-            f"Expanding restored and cloned PVCs to {pvc_size_expand_3 + 2}Gi"
+            f"Expanding restored and cloned PVCs to {pvc_size_expand_4}Gi"
         )
         for pvc_obj in restore_objs_new + clone_objs:
             log.info(
                 f"Expanding size of PVC {pvc_obj.name} to "
-                f"{pvc_size_expand_3 + 2}Gi"
+                f"{pvc_size_expand_4}Gi"
             )
-            pvc_obj.resize_pvc(pvc_size_expand_2, True)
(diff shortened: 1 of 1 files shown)


tests/manage/pv_services/test_expansion_snapshot_clone.py
```diff
@@ -0,0 +1,398 @@
...
+        log.info("IO started on all pods")
+
+        log.info("Wait for IO completion on pods")
+        for pod_obj in self.pods:
+            pod_obj.get_fio_results()
+            log.info(f"IO finished on pod {pod_obj.name}")
+            # Calculate md5sum
+            md5sum = pod.cal_md5sum(pod_obj, filename)
+            pod_obj.pvc.md5sum = md5sum
+        log.info("IO completed on all pods")
```
ebenahar, line 79: there are some other test cases under tests/manage/pv_services that implement the same.
The problem with that is if changes are needed, all implementations will need to be changed. 
For example, 
https://github.com/red-hat-storage/ocs-ci/blob/86ad959efefb89ab12ba9c16180636062c8c7d54/tests/manage/pv_services/pvc_snapshot/test_snapshot_at_different_pvc_utlilization_level.py#L65

Would it be possible to create a function that contains this code and call it from all places?

jilju: @ebenahar Are you suggesting a function which accepts a list of pods and check IO status in each pod ?
Checking md5sum depends on test design. Here md5sum is obtained and added an attribute of PVC instance. This is done for verification in a later stage. If pod is not created using factory, pvc attribute will not be present in pod_obj. Some tests cases needs md5sum comparison after IO where the value to compare will be present is some list or dict. Other test cases like https://github.com/red-hat-storage/ocs-ci/blob/86ad959efefb89ab12ba9c16180636062c8c7d54/tests/manage/pv_services/pvc_snapshot/test_snapshot_at_different_pvc_utlilization_level.py#L65 will need to add the md5sum value in a dict.

vasukulkarni (approved):

Akarsha-rai (approved):

vavuthu (approved):

PR #3234, opened 2020-10-30
WIP: Implement automatic code formatting with Black

clacroix12:
Fixes: #2952

PR #3277, opened 2020-11-05
Manually fix files that black could not automatically reformat

clacroix12:

### DIFF
--- a/ocs_ci/utility/lvm_utils.py
+++ b/ocs_ci/utility/lvm_utils.py
@@ -30,64 +30,82 @@ def osd_scenario1(osd, devices_dict, dmcrypt=False):
 
     pvcreate(osd, devices_dict.get('devices'))
     vgname = vgcreate(osd, osd.LvmConfig.vg_name % '1', devices_dict.get('device1'))  # dev is a single device
-    data_lv1 = lvcreate(osd,
-                        osd.LvmConfig.data_lv %
-                        '1',
-                        osd.LvmConfig.vg_name %
-                        '1',
-                        osd.LvmConfig.size.format(20))
-    data_lv2 = lvcreate(osd,
-                        osd.LvmConfig.data_lv %
-                        '2',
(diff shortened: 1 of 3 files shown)


clacroix12:
This PR is to set our repository up so we can use black to format everything in one go. This commit was cherry picked from #3234 which is no longer going to be merged and instead split into a few different PRs, this being the first.

vasukulkarni (approved):

mbukatov (approved):

ocs_ci/utility/lvm_utils.py
```diff
@@ -30,64 +30,82 @@ def osd_scenario1(osd, devices_dict, dmcrypt=False):
...
+        osd.LvmConfig.vg_name %
+        '1',
+        osd.LvmConfig.size.format(2)
+    )
+
+    scenario = (
+        "{{'data':'{datalv1}','data_vg':'{vg_name}'}},{{'data':'{datalv2}','data_vg':'{vg_name}',"
+        "'db':'{dblv1}','db_vg':'{vg_name}'}},"
+        "{{'data':'{datalv3}','data_vg':'{vg_name}','wal':'{wallv1}','wal_vg':'{vg_name}'}},"
+        "{{'data':'{datalv4}','data_vg':'{vg_name}','db':'{dblv2}','db_vg':'{vg_name}','wal':'{wallv2}',"
```
petr-balogh, line 102: I see this line is still quite long, can we maybe break this line as well?

petr-balogh (approved):
Approving and not blocking this one. Maybe this long line will be then fixed by black in next PR you are going to send ;)

ocs_ci/utility/lvm_utils.py
clacroix12: Black will organize the `.formats()` much better but not the strings if they are too long. I will eventually cover the long strings in subsequent PRs.

Issue #2478, opened 2020-07-10
ocs subscription name in CI and in product differ


fbalak:
In CI is subscription created from `ocs_ci/templates/ocs-deployment/subscription.yaml` and named `ocs-subscription`. Subscription installed without CI from UI is called `ocs-operator`.

PR #3279, opened 2020-11-06
Name ocs subscription the same like done by UI

petr-balogh:
In UI deployment the subscription is called ocs-operator,
so we should follow the same naming convention.

Fixes: #2478


petr-balogh:
We cannot merge this if we are running any upgrade job.  Than if someone would like to upgrade existing cluster deployed via ocs-ci some time before the merge they will need to change the name of the subscription before running the upgrade on the cluster.

vavuthu (approved):

petr-balogh:
Adding WIP to prevent this to be merged. We will merge once no upgrade job will run

dahorak (approved):

PR #3293, opened 2020-11-10
Test to verify PVC clone when rook-ceph, csi pods are deleted

jilju:
Verify PVC clone will succeeded if rook-ceph, csi pods are re-spun while creating the clone
OCS-2413 - FT-Resource deletion during PVC clone


vasukulkarni:
I think you should combine this test with pr ##3151 using params

ebenahar (approved):

clacroix12 (approved):

jilju:
> I think you should combine this test with pr ##3151 using params

@vasukulkarni All steps except the pod deletion, md5sum verification are different for both the test cases.

Issue #3297, opened 2020-11-11
When deploying with vSphere LSO all the worker nodes have OCS label instead of 3


yitzhak12:
When I deploy a cluster with the configurations: vSphere LSO 4.6(and also 4.5), 5 worker nodes - All the worker nodes are labeled with OCS label instead of just 3 worker nodes. 
Here is the link to the deployment: https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/14458/

And here is a snippet of the relevant logs:
`14:21:38 - MainThread - ocs_ci.deployment.deployment - INFO - Labeling worker nodes
14:21:38 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc label nodes compute-0 compute-1 compute-2 compute-3 compute-4 cluster.ocs.openshift.io/openshift-storage=''`

PR #3304, opened 2020-11-11
Label the nodes based on config value

vavuthu:
Label the nodes based on config value

Fixes: #3297 

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -47,13 +47,7 @@ from ocs_ci.ocs.resources.pod import (
 )
 from ocs_ci.ocs.uninstall import uninstall_ocs
 from ocs_ci.ocs.utils import setup_ceph_toolbox, collect_ocs_logs
-<<<<<<< HEAD
 from ocs_ci.utility import templating, ibmcloud
-from ocs_ci.utility.deployment import get_ocp_ga_version
-from ocs_ci.utility.localstorage import get_lso_channel
-=======
-from ocs_ci.utility import templating
->>>>>>> 9df9fd8e (moving lso code to deployment helpers)
 from ocs_ci.utility.openshift_console import OpenshiftConsole
 from ocs_ci.utility.retry import retry
 from ocs_ci.utility.utils import (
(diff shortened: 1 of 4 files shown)


vavuthu:
Above job is failed with no test resulst found which is not related to this PR.

Deploy job and destroy job was successfull.

https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/14671/consoleFull

As expected, it label 3 nodes
``` 
10:49:49 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc label nodes compute-0 compute-1 compute-2 cluster.ocs.openshift.io/openshift-storage=''
```

marking it as verified

ocs_ci/deployment/deployment.py
```diff
@@ -927,12 +927,18 @@ def setup_local_storage(storageclass):
...
     workers = get_typed_nodes(node_type="worker")
     worker_names = [worker.name for worker in workers]
-    worker_names_str = " ".join(worker_names)
-    logger.debug("Workers: %s", workers)
+    logger.debug("Workers: %s", worker_names)
 
     # Label the worker nodes
     logger.info("Labeling worker nodes")
-    run_cmd(f"oc label nodes {worker_names_str} " f"{constants.OPERATOR_NODE_LABEL}")
+    to_label = config.DEPLOYMENT.get("ocs_operator_nodes_to_label", 3)
```
petr-balogh, line 934: I am curious why we don't use function for labeling and taint here:
`label_and_taint_nodes`
https://github.com/red-hat-storage/ocs-ci/blob/41034ca4342ea092c91d2ebe4b877ec2585b73c4/ocs_ci/deployment/deployment.py#L176

Only one problem I see that this function is called after:
https://github.com/red-hat-storage/ocs-ci/blob/41034ca4342ea092c91d2ebe4b877ec2585b73c4/ocs_ci/deployment/deployment.py#L381

But if we will change the order and move the code:
https://github.com/red-hat-storage/ocs-ci/blob/41034ca4342ea092c91d2ebe4b877ec2585b73c4/ocs_ci/deployment/deployment.py#L370-L371

After the place where we call function for labeling I think it should be feasible to handle labeling on one place and if some differences are needed this should be handled in that function.

petr-balogh (dismissed):

clacroix12 (dismissed):

petr-balogh (dismissed):

PrasadDesala (approved):

jilju (approved):

PR #3305, opened 2020-11-11
Add namespace bucket test for azure

fbalak:

### DIFF
--- a/ocs_ci/ocs/resources/mcg.py
+++ b/ocs_ci/ocs/resources/mcg.py
@@ -455,16 +455,15 @@ class MCG:
 
         return creds_request, aws_access_key_id, aws_access_key
 
-    def create_connection(self, cld_mgr, conn_name=None, platform=constants.AWS_PLATFORM):
+    def create_connection(self, cld_mgr, platform, conn_name=None):
         """
         Creates a new NooBaa connection to an AWS backend
 
         Args:
             cld_mgr (obj): A cloud manager instance
+            platform (str): Platform to use for new connection
             conn_name (str): The connection name to be used
                 If None provided then the name will be generated
-            platform (str): Platform to use for new connection
-                AWS platform is used as default
 
         Returns:
             bool: False if the connection creation failed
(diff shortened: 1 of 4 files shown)


ocs_ci/ocs/resources/mcg.py
```diff
@@ -453,29 +455,45 @@ def _check_aws_credentials():
 
         return creds_request, aws_access_key_id, aws_access_key
 
-    def create_new_aws_connection(self, cld_mgr, conn_name=None):
+    def create_connection(self, cld_mgr, conn_name=None, platform=constants.AWS_PLATFORM):
```
mbukatov, line 458: I would expect that no platform would be default, and one will need to chose, esp. when we are turning old AWS only function into proper multi platform one.

tests/conftest.py
```diff
@@ -2867,21 +2868,33 @@ def ns_resource_factory(request, mcg_obj, cld_mgr, cloud_uls_factory):
 
     """
     created_ns_resources = []
-    created_ns_connections = []
+    created_connections = {}
 
-    def _create_ns_resources():
+    def _create_ns_resources(platform=constants.AWS_PLATFORM):
```
mbukatov, line 2873: Why does it make sense to select AWS as predefined platform for the tests?

mbukatov (changes requested):
This is step into the right direction, but I would rather not introduce new platform related choices with AWS as a default platform in the code.

tests/conftest.py
fbalak: Because I don't want to break any existing functionality and tests. I am not only one that uses these functions and I am not planning to update all tests to have explicitly selected platform in this PR.

ocs_ci/ocs/resources/mcg.py
fbalak: Ok, removed default platform from this.

tests/conftest.py
mbukatov: Makes sense.

mbukatov (approved):

parikshithb (approved):
Nice... LGTM

Issue #3307, opened 2020-11-11
LSO not being created in namespace defined in config


clacroix12:
We have the namespace for LSO set in the [base yaml](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/templates/ocs-deployment/local-storage-operator.yaml) used for creation. We are not overwriting this with the value from our configuration so users that pass [configurations that set the namespace](https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/local_storage_namespace.yaml) will [experience failures](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/14571/console).

PR #3308, opened 2020-11-12
Use namespace from config for LSO deployment

clacroix12:
Use LSO namespace defined in configuration when creating LSO and local volume

Fixes: #3307 


clacroix12:
I have verified the deployment locally and ensured that the temp yaml files are created with the value defined in the config rather than always going with the default value.

vavuthu (approved):

Avilir (approved):

prsurve (approved):

PR #3314, opened 2020-11-12
Black formatting preparation

clacroix12:
This PR fixes various issues that were causing pre-commit to fail after formatting the repo with black. These failures were coming up now because pre-commit runs against files in your changeset and not against the entire repository. These problems existed before black formatting was being done.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,6 @@ repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v2.5.0
   hooks:
-    - id: check-docstring-first
     - id: check-json
     - id: check-yaml
       args: [--multi]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,7 +13,6 @@
 import os
 import sys
 
-import sphinx_rtd_theme
 
 sys.path.insert(0, os.path.abspath('..'))
 
@@ -44,7 +43,10 @@ templates_path = ['_templates']
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
-exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store', '.venv', 'venv', 'tests', '.tox', 'data', '.pytest_cache', 'usecases/README.rst']
+exclude_patterns = [
+    '_build', 'Thumbs.db', '.DS_Store', '.venv', 'venv', 'tests', '.tox', 'data',
+    '.pytest_cache', 'usecases/README.rst'
+]
 
 
 # -- Options for HTML output -------------------------------------------------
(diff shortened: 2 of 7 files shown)


ocs_ci/utility/gcp.py
```diff
@@ -26,17 +26,16 @@
...
+# str: absolute filepath of json file with service account key
+#
+# This is json key file of ``sg-serv-account`` service account, which has full
+# admin rights in given GCP project. The same key file is used by openshift
+# installer during OCP installation to create all cluster resources from virtual
+# machines to hostnames. Modules from ocs-ci are using the same key to get full
+# cluster access as well.
+#
+# For more details, see `GCP documentation on ServiceAccountKey resource
+# <https://cloud.google.com/iam/docs/reference/rest/v1/projects.serviceAccounts.keys>`_
```
mbukatov, line 38: Would this change preserve this docstring in our sphinx build?

Right now, it's part of our documentation as expected: https://ocs-ci.readthedocs.io/en/latest/apidoc/ocs_ci.utility.html#ocs_ci.utility.gcp.SERVICE_ACCOUNT_KEY_FILEPATH

clacroix12: Hmm, this I'm not sure of. Right now it's causing the following error in our `check-docstring-first` hook:
```
Check docstring is first.................................................Failed
- hook id: check-docstring-first
- exit code: 1

ocs_ci/utility/gcp.py:29 Multiple module docstrings (first docstring on line 2).
```

vasukulkarni: I think we can check the docs from this build and resolve it if it looks good

clacroix12: There is actually an open issue for this https://github.com/pre-commit/pre-commit-hooks/issues/159 with a potential fix that has yet to be merged. It may be worth reconsidering the use of `check-docstring-first` until that is fixed if we want to use attribute docstrings.

clacroix12: Actually, looks like there may be a workaround for this: https://github.com/pre-commit/pre-commit-hooks/issues/159#issuecomment-559886109

mbukatov: This is a bug in `check-docstring-first`, likely https://github.com/pre-commit/pre-commit-hooks/issues/159 - as you can see in the html sphinx build, both module docstring and variable docstring are present as expected.

I suggest to disable this particular hook.

clacroix12: Yeah, I have tried to use the `#:` syntax workaround but it doesn't seem to work for me. I'm not seeing any attribute doc unless I go with the triple quote docstring method.

> I suggest to disable this particular hook.
I agree, I am going to revert the change to this docstring and disable the check entirely.

clacroix12:
@vasukulkarni @mbukatov I have reverted the change to that particular docstring and removed `check-docstring-first` from our pre-commit hooks. I don't think it offers enough for us to limit our documentation for it.

mbukatov (approved):

vasukulkarni (approved):

Issue #2952, opened 2020-09-15
Automatically format code in pre-commit with black


clacroix12:
Based on the investigation that was done in #2848 we are going to implement [black](https://github.com/psf/black) as an automatic code formatter for ocs-ci. There are a few things that need to be done in order to make this work:

- [x] Use black to format all existing ocs-ci code
- [x] Add black as a pre-commit hook
- [x] Add (minimal) configuration file (optional / not necessary)
- [x] Add black to tox
- [x] Ignore certain formatting checks from flake8
- [x] Add black to our travis-ci pipeline for PR verification
- [x] Update documentation to reflect black usage and new coding guidelines

clacroix12:
Documentation added in #4397

PR #3317, opened 2020-11-12
dev workflow updates for black

clacroix12:
Addresses a few of the outstanding tasks in #2952 by adding `black` to our developer requirements, pre-commit hooks, tox, and travis-ci. Also adds a file that can be used to clean up `git blame` output by ignoring changes done in the reformatting commit.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Format repository using black
+ee4dc48b7f9f5ce54fab47a3c3d9b3dea12c52f1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,11 @@
 repos:
 
+- repo: https://github.com/psf/black
+  rev: 20.8b1
+  hooks:
+    - id: black
+      language_version: python3.7
+
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v2.5.0
   hooks:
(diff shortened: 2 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -1,5 +1,11 @@
 repos:
 
+- repo: https://github.com/psf/black
+  rev: 20.8b1
+  hooks:
+    - id: black
+      language_version: python3.7
```
vasukulkarni, line 7: do we need language_version?

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Format repository using black
+ee4dc48b7f9f5ce54fab47a3c3d9b3dea12c52f1
```
vasukulkarni, line 2: ?

.pre-commit-config.yaml
clacroix12: This is how they have it in the [black documentation](https://github.com/psf/black#version-control-integration) so I stuck with it.

.git-blame-ignore-revs
clacroix12: I covered that a bit in the PR description. I'll also elaborate on it more when I update our doc but this should explain: https://github.com/psf/black#migrating-your-code-style-without-ruining-git-blame

vasukulkarni: you have to take the blame right? :-)

vasukulkarni (approved):

mbukatov (approved):

PR #3330, opened 2020-11-16
Test case for mcg namespace supported S3 operations

parikshithb:

### DIFF
--- a/ocs_ci/ocs/bucket_utils.py
+++ b/ocs_ci/ocs/bucket_utils.py
@@ -837,7 +837,7 @@ def s3_delete_bucket_website(s3_obj, bucketname):
     return s3_obj.s3_client.delete_bucket_website(Bucket=bucketname)
 
 
-def s3_put_bucket_versioning(s3_obj, bucketname, status="Enabled"):
+def s3_put_bucket_versioning(s3_obj, bucketname, status="Enabled", s3_client=None):
     """
     Boto3 client based Put Bucket Versioning function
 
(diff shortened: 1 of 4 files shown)


ebenahar (approved):

ocs_ci/ocs/bucket_utils.py
```diff
@@ -833,35 +833,45 @@ def s3_delete_bucket_website(s3_obj, bucketname):
...
         status (str): 'Enabled' or 'Suspended'. Default 'Enabled'
+        s3_client : Any s3 client resource
 
     Returns:
         dict : PutBucketVersioning response
     """
-    return s3_obj.s3_client.put_bucket_versioning(
-        Bucket=bucketname, VersioningConfiguration={"Status": status}
-    )
+    if s3_client:
```
romayalon, line 849: Versioning is out of scope for namespace buckets.

parikshithb: Ins't just list_object_versions supported on namespace buckets? Basically, I have enabled versioning on resource targets(aws here)  then uploaded object directly to the target and just list these objects using namespace bucket. Isn't this supported?

romayalon: ahh if you enabled versioning on the target bucket so it's ok, just notice that namespace buckets on top of Azure do not support list_object_versions.

parikshithb: Yup, will add a condition to skip object versioning if azure.

tests/e2e/lifecycle/mcg/test_mcg_namespace_s3_ops.py
```diff
@@ -0,0 +1,421 @@
...
+        bucket_utils.s3_put_object(
+            s3_obj=mcg_obj, bucketname=bucket_name, object_key=obj_key, data=OBJ_DATA
+        )
+        object_keys.append(obj_key)
+        object_prefixes.append(f"{prefix}-{i}/")
+    mid_index = len(object_keys) // 2
+    return object_keys, object_prefixes, mid_index
+
+
+def get_list_and_verify(response, keys, verify="Contents", prefix="", delimiter=""):
```
fbalak, line 46: Please add docstring with description what it does.

fbalak (commented):
Looks great!
Would it make sense to split the `test_mcg_namespace_s3_ops` into more test cases? This way are not tested all operations if one of operations at the beginning fails...

tests/e2e/lifecycle/mcg/test_mcg_namespace_s3_ops.py
parikshithb: Added doc string

parikshithb: Agree. Test is not split into 3 sub tests: Basic ops , object version and multipart.

fbalak (commented):
Please add Polarion ids.

parikshithb:
This test was created as single lifecycle testcase there are no subtest created for each. I have added lifecycle test polarion id above class.

fbalak (approved):
Ack

ebenahar (approved):

PrasadDesala (approved):

PR #3345, opened 2020-11-17
Uninstall OCS 4.6

annagitel:
Signed-off-by: annagitel <asandler@redhat.com>
this PR updates the uninstall procedure to OCS 4.7
- node encryption removal
- support of retain/delete annotation
- storage classes are removed from uninstall 
- snapshots deletion 
- updated LSO removal 
- taints removal
- new CRDs removal

PR #3346, opened 2020-11-17
Rewriting OCS destroy on IBM Power

zmc:
With these changes, deploying OCS after it has already been deployed and destroyed works *usually*; I plan on continuing this work but want to get this in as it's more functional than before.

Proof it doesn't regress: https://storage-jenkins-csb-ceph.cloud.paas.psi.redhat.com/job/ocs-ci-kvm/120/testReport/

### DIFF
--- a/ocs_ci/deployment/ibm.py
+++ b/ocs_ci/deployment/ibm.py
@@ -136,19 +136,51 @@ class IBMDeployment(Deployment):
                 ocp.OCP().exec_oc_cmd(f"oc delete pv {pv_name}")
 
         # Section 3.1 Step 9
-        try:
-            ocp.OCP().exec_oc_cmd(
-                "delete crd backingstores.noobaa.io bucketclasses.noobaa.io "
-                "cephblockpools.ceph.rook.io cephclusters.ceph.rook.io "
-                "cephfilesystems.ceph.rook.io cephnfses.ceph.rook.io "
-                "cephobjectstores.ceph.rook.io cephobjectstoreusers.ceph.rook.io "
-                "noobaas.noobaa.io ocsinitializations.ocs.openshift.io  "
-                "storageclusterinitializations.ocs.openshift.io "
-                "storageclusters.ocs.openshift.io cephclients.ceph.rook.io "
-                "--wait=true --timeout=5m"
-            )
-        except CommandFailed:
-            pass
+        # Note that the below process differs from the documentation slightly.
+        # Instead of deleting all CRDs at once and calling the job done, we
+        # iterate over a list of them, noting which ones don't delete fully and
+        # applying the standard workaround of removing the finalizers from any
+        # CRs and also the CRD.
+        crd_types = [
+            "backingstores.noobaa.io",
+            "bucketclasses.noobaa.io",
+            "cephblockpools.ceph.rook.io",
+            "cephclients.ceph.rook.io",
+            "cephclusters.ceph.rook.io",
+            "cephfilesystems.ceph.rook.io",
+            "cephnfses.ceph.rook.io",
+            "cephobjectstores.ceph.rook.io",
+            "cephobjectstoreusers.ceph.rook.io",
+            "noobaas.noobaa.io",
+            "ocsinitializations.ocs.openshift.io",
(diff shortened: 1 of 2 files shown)


ocs_ci/deployment/ibm.py
```diff
@@ -26,206 +30,174 @@ def destroy_ocs(self):
...
-        worker_node_names = [node["metadata"]["name"] for node in worker_nodes]
-        logger.warning(worker_node_names)
-
-        for worker_node in worker_node_names:
-            rookstring = (
-                "oc debug node/"
-                + worker_node
-                + " -- chroot /host rm -rfv /var/lib/rook"
+        pvc.delete_pvcs(rbd_pvcs)
+        cephfs_pvcs = pvc.get_all_pvcs_in_storageclass("ocs-storagecluster-cephfs")
```
vasukulkarni, line 47: can you use from constants here? https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/constants.py#L96

ocs_ci/deployment/ibm.py
```diff
@@ -26,206 +30,174 @@ def destroy_ocs(self):
...
-                )
-                .get()
-                .get("spec")
+        # https://access.redhat.com/documentation/en-us/red_hat_openshift_container_storage/4.5/html/deploying_openshift_container_storage_using_bare_metal_infrastructure/assembly_uninstalling-openshift-container-storage_rhocs
+
+        # Section 3.1 Step 1
+        # Deleting PVCs
+        rbd_pvcs = [
+            p
+            for p in pvc.get_all_pvcs_in_storageclass("ocs-storagecluster-ceph-rbd")
```
vasukulkarni, line 40: we can use constants here to get storageclassname https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/constants.py#L95

ocs_ci/ocs/constants.py
```diff
@@ -672,6 +672,7 @@
 
 # Deployment constants
 OCS_CSV_PREFIX = "ocs-operator"
+LOCAL_STORAGE_SUBSCRIPTION = "local-storage-operator"
```
clacroix12, line 675: Do we need this? We already have a constant for `local-storage-operator` on the next line. If the constant name is the issue we could always make this more generic. `LOCAL_STORAGE_OPERATOR` or something like that.

zmc: Oh, if `LOCAL_STORAGE_CSV_PREFIX` has the same effective meaning as `LOCAL_STORAGE_SUBSCRIPTION` I'm fine with that.

vasukulkarni (approved):

annagitel:
So apparently we worked on the same automation of the OCS uninstall process for the last couple of month
yours - supported on IBM
mine - meant to support all platforms (so far works on VMWARE and AWS, for 4.6 meant to support IBM,bare-metal, Azure)

https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/uninstall.py

open PR https://github.com/red-hat-storage/ocs-ci/pull/3345

ocs_ci/deployment/ibm.py
```diff
@@ -26,206 +30,210 @@ def destroy_ocs(self):
...
-            logger.info(out)
-
-            for devicepath in alldevicepaths:
-                sgdiskstring = (
-                    "oc debug node/"
-                    + worker_node
-                    + " -- chroot /host sgdisk --zap-all "
-                    + devicepath
+        except (CommandFailed, subprocess.TimeoutExpired):
+            pass
```
clacroix12, line 69: Should we log something here or do we get enough output from `exec_oc_cmd`? If we do decide to add some logging let's also address the other instances of `except` -> `pass`/`continue`.

zmc: I don't *think* it's necessary to log, since the step after that will delete the namespace and also remove finalizers from several object types including storageclusters if the namespace doesn't delete fully. It just won't be terribly helpful information, and could even be a red herring.

clacroix12 (approved):

PR #3355, opened 2020-11-19
fixed multiple pvc  deletion case and the way we capture the time

SvetlanaRedHat:
Originally  the way that multiple pvc got deleted was wrong, 
in this change i am trying to delete 120 pvc at once, and also wrote function that returned total time of 120 pvc duration

### DIFF
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -1420,6 +1420,58 @@ def measure_pvc_creation_time_bulk(interface, pvc_name_list, wait_time=60):
     return pvc_dict
 
 
+def get_multiple_pvc_start_or_end_time(interface, pvc_name, status):
+    """
+    Get the starting/ending deletion time of a PVC based on provisioner logs
+
+    Args:
+        interface (str): The interface backed the PVC
+        pvc_name (str / list): Name of the PVC(s) for creation time
+                               the list will be list of pvc objects
+        status (str): the status that we want to get - Start / End
+
+    Returns:
+        datetime object: Time of PVC(s) deletion
+
+    """
+    # Define the status that need to retrieve
+    operation = "started"
+    if status.lower() == "end":
+        operation = "succeeded"
+
+    format = "%H:%M:%S.%f"
+    # Get the correct provisioner pod based on the interface
+    pod_name = pod.get_csi_provisioner_pod(interface)
+    # due to some delay in CSI log generation added wait
+    if isinstance(pvc_name, list):
+        time.sleep(len(pvc_name))
+    else:
+        time.sleep(60)
+    # get the logs from the csi-provisioner containers
+    logs = pod.get_pod_logs(pod_name[0], "csi-provisioner")
+    logs += pod.get_pod_logs(pod_name[1], "csi-provisioner")
+
+    logs = logs.split("\n")
+    # Extract the time for the one PVC deletion
+    if isinstance(pvc_name, str):
+        stat = [i for i in logs if re.search(f"delete.*{pvc_name}.*{operation}", i)]
+        stat = stat[0].split(" ")[1]
+    # Extract the time for the list of PVCs deletion
+    if isinstance(pvc_name, list):
+        all_stats = []
+        for pv_name in pvc_name:
+            name = pv_name.name
+            stat = [i for i in logs if re.search(f"delete.*{name}.*{operation}", i)]
+            stat = stat[0].split(" ")[1]
+            all_stats.append(stat)
+        all_stats = sorted(all_stats)
+        if status.lower() == "end":
+            stat = all_stats[-1]  # return the highest time
+        elif status.lower() == "start":
+            stat = all_stats[0]  # return the lowest time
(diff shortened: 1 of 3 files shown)


SvetlanaRedHat:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/3388/

AviadP (approved):

ocs_ci/helpers/helpers.py
```diff
@@ -1392,6 +1392,55 @@ def measure_pvc_creation_time_bulk(interface, pvc_name_list, wait_time=60):
     return pvc_dict
 
 
+def measure_multiple_pvc_deletion_time(interface, pvc_name, status):
+    """
+    Get the starting/ending deletion time of a PVC based on provisioner logs
+
+    Args:
+        interface (str): The interface backed the PVC
+        pvc_name (str / list): Name of the PVC(s) for creation time
```
ramkiperiy, line 1401: Better rename it to pvc_name_list or pvc_list

ocs_ci/helpers/helpers.py
```diff
@@ -1392,6 +1392,55 @@ def measure_pvc_creation_time_bulk(interface, pvc_name_list, wait_time=60):
...
+                               the list will be list of pvc objects
+        status (str): the status that we want to get - Start / End
```
ramkiperiy, line 1403: Purpose of the function is to return start time or end time of PVC ? if so function name is misleading please change it to something like get_multiple_pvc_start_or_end_time()

ocs_ci/helpers/helpers.py
```diff
@@ -1392,6 +1392,55 @@ def measure_pvc_creation_time_bulk(interface, pvc_name_list, wait_time=60):
...
+    # Define the status that need to retrieve
+    operation = "started"
+    if status.lower() == "end":
+        operation = "succeeded"
+
+    format = "%H:%M:%S.%f"
+    # Get the correct provisioner pod based on the interface
+    pod_name = pod.get_csi_provisioner_pod(interface)
+    # due to some delay in CSI log generation added wait
+    time.sleep(60)
```
ramkiperiy, line 1418: Would recommend to change the sleep time according to the PVC numbers. Have faced this problem with just 100 PVC 60sec might be enough but with more PVC numbers like 500/1000 this 60sec won't be enough.

ramkiperiy: Is it good to create separate function to address this?

ocs_ci/helpers/helpers.py
```diff
@@ -2003,6 +2052,42 @@ def create_pods_parallel(
     return pod_objs
 
 
+def delete_multiple_pvcs(pvc_objs):
```
ramkiperiy, line 2055: Can we use this function? https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/helpers/helpers.py#L2006 
This will delete all the objects in parallel

Avilir (dismissed):
Nice work

tests/e2e/performance/test_pvc_deletion_performance.py
```diff
@@ -127,13 +127,19 @@ def test_multiple_pvc_deletion_measurement_performance(self, teardown_factory):
...
 
         # Get PVC deletion time
-        pvc_deletion_time = helpers.measure_pv_deletion_time_bulk(
-            interface=self.interface, pv_name_list=pv_name_list
-        )
-        logging.info(f"{number_of_pvcs} PVCs deletion time took {pvc_deletion_time}")
+        start_time = helpers.measure_multiple_pvc_deletion_time(self.interface, pvc_objs, status="start")
+        end_time = helpers.measure_multiple_pvc_deletion_time(self.interface, pvc_objs, status="end")
+        total = end_time - start_time
+        total_time = total.total_seconds()
```
keesturam, line 138: @SvetlanaRedHat 
having this in a separate function oud be great. 
get_pvc_deletion_status_time -> collects the start/end time
measure_pvc_deletion_time -> Does the diff between start time and end time. 

We don't need to expose get_pvc_deletion_status_time function and have it defined within measure_pvc_deletion_time. wdyt?

ocs_ci/helpers/helpers.py
SvetlanaRedHat: This does not do in paralell, but does in thread which is not exacly what we want, in my function i delete them at once and its more accurate

ocs_ci/helpers/helpers.py
SvetlanaRedHat: Addressed this conversation

ocs_ci/helpers/helpers.py
SvetlanaRedHat: this can be string or list, prefer to keep it similar to other variable names

tests/e2e/performance/test_pvc_deletion_performance.py
SvetlanaRedHat: Kartick I tried to make is similar to measure mutiple pvc creation function

PR #3402, opened 2020-11-24
update setup.py to match released version

vasukulkarni:
Updates release note to match released version and also adds the initial docs for release note process.

### DIFF
--- a/docs/release_guidelines.md
+++ b/docs/release_guidelines.md
@@ -1,0 +1,25 @@
+# OCS-CI Releases
+
+OCS-CI releases are generated in conjunction with OCS releases, A release is cut
+as soon as OCS is Generally Availabled and available in open market place.
+
+The release helps to checkout specific versions of tests that were run during
+the release and serves as baseline for the specific version of OCS.
+
+# Release Process
+1. Create a tag specific to commit that matches the latest ocs-ci that was used to run
+the regression tests. one can create the tag via the git command line or via github
+
+    CLI:
+       * `git tag -l` to list current tags
+       * `git tag -a v4.5.0 abceb01 -m "v4.5.0 tag"` to add tag to commit abceb01
+       * `git push --tags` to push the tag
+
+    Github GUI:
+      From the main project, use the release page to Draft a new release
+      Specify new tag name (eg: v4.5.1) and the commit in '@' target field
+      Describe the release notes and publish the draft (steps 3 and 4 below)
+
+2. Update the setup.py to match the ocs version. 
+3. Update the release notes to highlight changes that went in the ocs-ci repo
+4. Generate the release
--- a/docs/release_guidelines.md
+++ b/docs/release_guidelines.md
@@ -10,11 +10,11 @@ the release and serves as baseline for the specific version of OCS.
 1. Update the setup.py to match the ocs version and commit the change.
 2. Create a tag specific to commit that matches the latest ocs-ci that was used to run
 the regression tests. one can create the tag via the git command line or via github
-## CLI:
+  ### CLI:
        * `git tag -l` to list current tags
        * `git tag -a v4.5.0 abceb01 -m "v4.5.0 tag"` to add tag to commit abceb01
        * `git push --tags` to push the tag
-## Github GUI:
+  ### Github GUI:
       From the main project, use the release page to Draft a new release
       Specify new tag name (eg: v4.5.1) and the commit in '@' target field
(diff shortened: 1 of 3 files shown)


Avilir (dismissed):

petr-balogh (dismissed):

docs/release_guidelines.md
```diff
@@ -0,0 +1,14 @@
+# OCS-CI Releases
```
petr-balogh, line 1: You are missing to link this document in index to be part of the generated documentation.
https://github.com/red-hat-storage/ocs-ci/blob/master/docs/index.rst

docs/release_guidelines.md
```diff
@@ -0,0 +1,14 @@
...
+
+OCS-CI releases are generated in conjunction with OCS releases, A release is cut
+as soon as OCS is Generally Availabled and available in open market place.
+
+The release helps to checkout specific versions of tests that were run during
+the release and serves as baseline for the specific version of OCS.
+
+# Release Process
+1. Create a tag specific to commit that matches the latest ocs-ci that was used to run
+the regression tests.
```
petr-balogh, line 11: Can you please provide more specific example how to create a tag here? So anyone will know what exact command to run in this step?

vasukulkarni: Added the steps

docs/release_guidelines.md
```diff
@@ -0,0 +1,25 @@
...
+       * `git tag -l` to list current tags
+       * `git tag -a v4.5.0 abceb01 -m "v4.5.0 tag"` to add tag to commit abceb01
+       * `git push --tags` to push the tag
+
+    Github GUI:
+      From the main project, use the release page to Draft a new release
+      Specify new tag name (eg: v4.5.1) and the commit in '@' target field
+      Describe the release notes and publish the draft (steps 3 and 4 below)
+
+2. Update the setup.py to match the ocs version.
```
petr-balogh, line 23: Shouldn't we create first the pr/commit with version change to have release containing that version?

If we do first the tag and then the version change then we will not have it as par of release no?

Or do I miss something?

petr-balogh: @vasukulkarni have you noticed my comment here?

vasukulkarni: @petr-balogh fixed

docs/release_guidelines.md
```diff
@@ -0,0 +1,23 @@
...
+
+OCS-CI releases are generated in conjunction with OCS releases, A release is cut
+as soon as OCS is Generally Available and available in open market place.
+
+The release helps to checkout specific versions of tests that were run during
+the release and serves as baseline for the specific version of OCS.
+
+# Release Process
+1. Update the setup.py to match the ocs version and commit the change.
+2. Create a tag specific to commit that matches the latest ocs-ci that was used to run
```
petr-balogh, line 11: I think that anyway we need to tag the commit created above which will match the version. If we will tag the commit which we ran the regression cycle with then we will never have the proper version in the setup.py for this release.  I think that we should just do that release the same date as we do the release of OCS and we don't have to be that strict to have exact commit we ran the regression cycle with. WDYT?

So if that OK I would suggest to create the tag on the commit created in 1st step here @vasukulkarni

vasukulkarni: If we do the same day release I think we are safe, but in case we miss we still have in this document that works. The only thing the user has to skip the same day is to not provide commit id in that case.

petr-balogh: We can make it as part of our release process and just as we do live testing we will do the release of OCS-CI as well.  @ebenahar @RazTamir  @dahorak @jimdevops19 @clacroix12 @shylesh @vavuthu   FYI.  WDYT?

vasukulkarni: @petr-balogh we can continue to have that discussion, but in the mean time lets update the version and get this one merged if everything is looking fine?   soon there will be one more update to setup.py

petr-balogh (approved):

clacroix12 (approved):
LGTM! @vasukulkarni I'll let you squash the commits as you see fit and merge

vasukulkarni:
> LGTM! @vasukulkarni I'll let you squash the commits as you see fit and merge

Thanks!

docs/release_guidelines.md
shylesh: @petr-balogh Agree.

PR #3412, opened 2020-11-25
update integration with flexy

dahorak:
- as flexy working directory always use data/flexy-dir
- copy flexy working dir to the cluster path
- change flexy image version

### DIFF
--- a/ocs_ci/deployment/flexy.py
+++ b/ocs_ci/deployment/flexy.py
@@ -308,19 +308,7 @@ class FlexyBase(object):
         run_cmd(chown_cmd)
         chmod_cmd = f"sudo chmod -R a+rX {self.flexy_host_dir}"
         run_cmd(chmod_cmd)
-
-        # dirs_exist_ok parameter for shutil.copytree is available in Python
-        # 3.8 and newer, so workarounding it with rsync command
-        # cluster_path_flexy_dir = os.path.join(
-        #     self.cluster_path, constants.FLEXY_HOST_DIR
-        # )
-        # shutil.copytree(
-        #     self.flexy_host_dir,
-        #     cluster_path_flexy_dir,
-        #     symlinks=True,
-        #     ignore_dangling_symlinks=True,
-        #     dirs_exist_ok=True,
-        # )
+        # mirror flexy work dir to cluster path
         rsync_cmd = f"rsync -av {self.flexy_host_dir} {self.cluster_path}/"
         run_cmd(rsync_cmd)
 
--- a/ocs_ci/deployment/flexy.py
+++ b/ocs_ci/deployment/flexy.py
@@ -22,7 +22,6 @@ from ocs_ci.utility.utils import (
     expose_ocp_version,
     wait_for_machineconfigpool_status,
 )
-from ocs_ci.ocs import exceptions
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


dahorak:
> PR validation
> 
> Cluster Name:
> Cluster Configuration: conf/ocsci/aws_upi_flexy_rhcos_3m_3w_proxy.yaml
> PR Test Suite: deployment
> PR Test Path: tests/
> Additional Test Params:
> OCP VERSION: 4.5
> OCS VERSION: 4.6
> 
> [Job FAILED](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/3453//testReport) (installation failed, tests not executed).

In fact, the [underlying deploy](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/15141/) job passed - the failure is because of some problem with the jenkins job.

petr-balogh (approved):

ocs_ci/deployment/flexy.py
```diff
@@ -291,45 +270,69 @@ def merge_flexy_env(self):
...
+        """
+        Perform copying of flexy-dir to cluster_path.
+        """
+        # change ownership of flexy-dir back to current user
+        chown_cmd = f"sudo chown -R {os.getuid()}:{os.getgid()} {self.flexy_host_dir}"
+        run_cmd(chown_cmd)
+        chmod_cmd = f"sudo chmod -R a+rX {self.flexy_host_dir}"
+        run_cmd(chmod_cmd)
+
+        # dirs_exist_ok parameter for shutil.copytree is available in Python
```
shylesh, line 312: do we need this commented part ?

ocs_ci/deployment/flexy.py
```diff
@@ -379,25 +369,31 @@ def deploy(self, log_level=""):
...
         """
         cmd = self.build_destroy_cmd()
-        self.run_container(cmd)
+        try:
+            self.run_container(cmd)
+        except Exception as err:
+            logger.error(err)
+            raise
+        finally:
+            self.flexy_backup_work_dir()
```
shylesh, line 396: I might be missing something but why do we need to backup work dir during after destroy ?

ocs_ci/deployment/flexy.py
dahorak: I think it will be slightly nicer to use pure python whenever possible, but `shutil.copytree` fails, if (some of) the destination directory exists, so I have to use `rsync` instead. In python 3.8 `copytree` method have additional parameter `dirs_exist_ok`, which if it is set to `True`, it will ignore the existing dirs. But if you think, it will be better to remove this commented code here, I have no problem with that - the `rsync` works well...

ocs_ci/deployment/flexy.py
dahorak: My intention is, to have the _cluster dir_ consistent with the state of the cluster. (So if the cluster is destroyed, the _cluster dir_ should reflect it I think.)

ocs_ci/deployment/flexy.py
dahorak: Reconsidered it and removed the commented part, we don't probably want to change this since the `rsync` way works correctly.

petr-balogh (approved):

shylesh (approved):
+1

PR #3425, opened 2020-11-26
Namespace bucket tier2 and tier3 tests

fbalak:
* Add some tier2 and tier3 tests for mcg namespace feature.
* Remove @aws_platform_required marker. All tests should be available on all platforms.

### DIFF
--- a/ocs_ci/ocs/bucket_utils.py
+++ b/ocs_ci/ocs/bucket_utils.py
@@ -36,20 +36,28 @@ def craft_s3_command(cmd, mcg_obj=None, api=False, signed_request_creds=None):
     """
     api = "api" if api else ""
     if mcg_obj:
+        if mcg_obj.region:
+            region = f"AWS_DEFAULT_REGION={mcg_obj.region} "
+        else:
+            region = ""
         base_command = (
             f'sh -c "AWS_CA_BUNDLE={constants.SERVICE_CA_CRT_AWSCLI_PATH} '
             f"AWS_ACCESS_KEY_ID={mcg_obj.access_key_id} "
             f"AWS_SECRET_ACCESS_KEY={mcg_obj.access_key} "
-            f"AWS_DEFAULT_REGION={mcg_obj.region} "
+            f"{region}"
             f"aws s3{api} "
             f"--endpoint={mcg_obj.s3_internal_endpoint} "
         )
         string_wrapper = '"'
     elif signed_request_creds:
+        if signed_request_creds.get("region"):
+            region = f'AWS_DEFAULT_REGION={signed_request_creds.get("region")} '
+        else:
+            region = ""
         base_command = (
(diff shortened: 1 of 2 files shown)


fbalak:
PR verification failed due to CI issue but the actual job with deployment and tests passed:
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/15177/testReport/

ocs_ci/ocs/bucket_utils.py
```diff
@@ -36,20 +36,28 @@ def craft_s3_command(cmd, mcg_obj=None, api=False, signed_request_creds=None):
     """
     api = "api" if api else ""
     if mcg_obj:
+        if mcg_obj.region:
+            region = f"AWS_DEFAULT_REGION={mcg_obj.region} "
```
mbukatov, line 40: Is this AWS only, or would it work on Azure, GCP, OpenStack and other providers?

tests/manage/mcg/test_namespace.py
```diff
@@ -27,10 +26,10 @@ class TestNamespace(MCGTest):
 
     MCG_NS_RESULT_DIR = "/result"
     MCG_NS_ORIGINAL_DIR = "/original"
+    # TODO: fix this when https://github.com/red-hat-storage/ocs-ci/issues/3338
+    # is resolved
+    DEFAULT_REGION = "us-east-2"
 
-    # Test is skipped for other platforms due to
-    # https://github.com/red-hat-storage/ocs-ci/issues/2774
-    @aws_platform_required
```
mbukatov, line 33: I don't see how this will be able to run on Azure, GCP and AWS, as the region names are always bit different. What am I missing?

ocs_ci/ocs/bucket_utils.py
fbalak: This is done so that other cloud providers that don't use region during authentication (like RGW) don't need to specify it.

tests/manage/mcg/test_namespace.py
fbalak: Tests in this file currently use resources from AWS, Azure and RGW. AWS and Azure use same region and RGW is not using a region to authenticate.

ocs_ci/ocs/bucket_utils.py
ebenahar: It is indeed confusing but from what I see, this is for creating the s3 command on the AWS CLI pod which is used by all platforms

fbalak: Yes, aws cli is used for s3 utilization of all platforms. AWS_DEFAULT_REGION, AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are aws cli specific variables but credentials contained in them should work for all s3 compatible interfaces.

ebenahar (approved):

parikshithb (approved):
LGTM

PR #3426, opened 2020-11-26
Deploying ocs if storage disks increases

AaruniAggarwal:
Deploying OCS using ocs-ci on IBM Power Systems if there are more than 1 storage disk available in an OCP cluster.

### DIFF
--- a/conf/ocsci/production_powervs_upi.yaml
+++ b/conf/ocsci/production_powervs_upi.yaml
@@ -11,4 +11,4 @@ ENV_DATA:
   deployment_type: 'upi'
   monitoring_enabled: false
   number_of_storage_nodes: 3
-  number_of_storage_disks: 1
+
--- a/conf/ocsci/production_powervs_upi.yaml
+++ b/conf/ocsci/production_powervs_upi.yaml
@@ -11,3 +11,4 @@ ENV_DATA:
   deployment_type: 'upi'
   monitoring_enabled: false
   number_of_storage_nodes: 3
+  number_of_storage_disks: 1
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -1127,6 +1127,9 @@ def setup_local_storage(storageclass):
     storage_class_device_count = 1
     if platform == constants.AWS_PLATFORM:
         storage_class_device_count = 2
+    if platform == constants.IBM_POWER_PLATFORM:
+        numberofstoragedisks = config.ENV_DATA["number_of_storage_disks"]
+        storage_class_device_count = numberofstoragedisks
     verify_pvs_created(len(worker_names) * storage_class_device_count)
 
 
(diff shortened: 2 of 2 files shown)


conf/ocsci/production_powervs_upi.yaml
```diff
@@ -11,3 +11,4 @@ ENV_DATA:
   deployment_type: 'upi'
   monitoring_enabled: false
   number_of_storage_nodes: 3
+  number_of_storage_disks: 1
```
zmc, line 14: I think we should drop this line - see my other comment
```suggestion
```

ocs_ci/deployment/deployment.py
```diff
@@ -1127,6 +1127,9 @@ def setup_local_storage(storageclass):
     storage_class_device_count = 1
     if platform == constants.AWS_PLATFORM:
         storage_class_device_count = 2
+    if platform == constants.IBM_POWER_PLATFORM:
+        numberofstoragedisks = config.ENV_DATA['number_of_storage_disks']
```
zmc, line 1131:
```suggestion
        numberofstoragedisks = config.ENV_DATA.get("number_of_storage_disks", 1)
```

We don't want a `KeyError` if someone doesn't use `production_powervs_upi.yaml`. This also means we can drop the value from that file as well.

zmc (changes requested):
Hi @AaruniAggarwal - I've suggested some simple changes; let me know what you think!

ocs_ci/deployment/deployment.py
AaruniAggarwal: The motive behind doing this was , we can update our config file in our case its (production_powervs_upi.yaml) according to the type of cluster we are using. We are using both libvirt and powervs clusters , and in both case number_of_storage_disks varies .  ie. in libvirt cluster no. of storage disks is 1, whereas for powervs its 8. By default i set it to 1 , so that it will not effect our ocs-upi-kvm project . 
If we don't use number_of_storage_disks in production_powervs_upi.yaml then we have to update deployment.py each time depending on the type of cluster we use . So i think good approach is updating config file rather than updating python file(deployment.py)  . Do you still want me to make the changes that you suggested, if yes then please let me know.
And while running run-ci for power , we are specifying production_powervs_upi.yaml as our config file. 
run-ci command we are using is : 

`run-ci -m "tier4a and manage" --ocsci-conf=conf/ocsci/production_powervs_upi.yaml --cluster-name "ocstest" --cluster-path /root tests/`

AaruniAggarwal:
@zmc , I commited the changes as per your suggesttions.

zmc (approved):

vasukulkarni (approved):

vasukulkarni:
Needs a verified flag to ensure it is test run

AaruniAggarwal:
Verified it on one of my system

PR #3437, opened 2020-11-29
Parameterize multicloud tests

### LINKED ISSUE #2304: Support multicloud scenarios
This issue *does not* refer to testing NooBaa while it's deployed on one of those.
This issue *does* refer to testing backingstores that use those services as their *target storage* supplies.

Cloud providers and systems that need to be supported. prioritized:
- [x] AWS S3
- [x] PVPool
- [x] IBM COS (S3 compatible)
- [x] Azure Blob Storage
- [x] Google Cloud Storage

bernerhat:
Closes #2304


MeridianExplorer:
Looks and runs great.
Please add IDs for the parametrization

tests/conftest.py
```diff
@@ -1902,42 +1903,59 @@ def rgw_bucket_factory_session(request, rgw_obj_session):
 
 
 @pytest.fixture()
-def bucket_factory(request, mcg_obj):
+def bucket_factory(request, bucket_class_factory, mcg_obj):
```
ebenahar, line 1906: there are more than 50 places in which this fixture is being used. Please make sure this change new arg addition is incorporated into these usages

MeridianExplorer: This arg should not affect existing calls to `bucket_factory`, since the fixture ultimately returns `_create_buckets()`.
`_create_buckets()` was only changed by the addition of an optional `bucketclass` parameter that defaults to `None`, and thus the change should be invisible to older code.
We also tested all of MCG tier1-3, including tests that were and weren't parametrized, and all of them passed.
I and Elad talked about this in the chat, and so, I'm resolving this conversation.

tests/conftest.py
```diff
@@ -1902,42 +1903,59 @@ def rgw_bucket_factory_session(request, rgw_obj_session):
...
 
     """
     created_buckets = []
 
-    def _create_buckets(amount=1, interface="S3", verify_health=True, *args, **kwargs):
+    def _create_buckets(
+        amount=1,
+        interface="S3",
+        verify_health=True,
+        bucketclass: typing.Union[None, str, dict] = None,
```
fbalak, line 1943: Why is there explicitly specified type by `typing` module? It is not done in any other part of our framework.

tests/conftest.py
```diff
@@ -1902,42 +1903,59 @@ def rgw_bucket_factory_session(request, rgw_obj_session):
...
         Creates and deletes all buckets that were created as part of the test
 
         Args:
             amount (int): The amount of buckets to create
             interface (str): The interface to use for creation of buckets.
                 S3 | OC | CLI | NAMESPACE
+            verify_Health (bool): Whether to verify the created bucket's health
+                post-creation
+            bucketclass (None|str|dict): An existing bucketclass name to use,
+                or a dictionary describing a new bucketclass to be created.
```
fbalak, line 1957: I would prefer to have only bucket name string as parameter. It seems overly complicated to me to create new bucket class as part of this factory if dict provided.

tests/manage/mcg/test_bucket_deletion.py
```diff
@@ -128,22 +139,60 @@ def test_bucket_delete(
...
         ],
     )
-    def test_bucket_delete_with_objects(self, mcg_obj, interface, awscli_pod):
+    def test_bucket_delete_with_objects(
+        self, mcg_obj, awscli_pod, bucket_class_factory, interface, bucketclass_dict
+    ):
         """
         Negative test with deletion of bucket has objects stored in.
+        ***IMPORTANT***
+        S3 interface is not supporting different bucketclasses DO NOT param S3
```
fbalak, line 182: What does it mean? (I don' understand the commend)
Also please add an empty line before and after the comment.

fbalak (commented):
Looks great! I like how the parametrization is done but I have few comments.

tests/conftest.py
MeridianExplorer: All test cases but one actually rely on the flow in which a new bucketclass is created.
The aim is for it to either be `None` (and then the bucket will use the default bucketclass), or `dict` (and then a bucketclass will be created).
Removing the option to create a new bucketclass as part of creating a new bucket will force us to refactor all of the parametrized tests to be two-liners instead of one-liners as they are now; Since we'll always have to create the bucketclass, and then pass its name/object to the factory anyway.

The only test that needs to use an existing bucketclass is `test_multi_region`, which was written before the current infrastructure was in place. The test needs access to the backingstores that are created for the bucket, but those are not externalized at the moment. The only way to access them is by creating a bucketclass separately, and then creating a bucket that utilizes it.

The plan is to change the design so that the ObjectBucket object will point to a BucketClass object, which will then have a list of its backingstores. Then, the option for `str` will be removed - and the factory will either use the default bucketclass, or create a new one.

tests/conftest.py
MeridianExplorer: It's meant as an additional calrifying layer regarding the current state of the variable.
More on the reason it can be one of three options [in this comment](https://github.com/red-hat-storage/ocs-ci/pull/3437#discussion_r539238628)

tests/manage/mcg/test_bucket_deletion.py
MeridianExplorer: This comment is actually misplaced; Comment has been moved and rephrased.

tests/conftest.py
ebenahar: what is required in order to accept only dict here? if it's possible to remove the str option, then this arg can be set to None as default and accept only dict

MeridianExplorer: Changed the code to accept `dict` and `None`

tests/conftest.py
MeridianExplorer: Var was changed to be either `dict` or `None`, and thus `typing` was ommitted. Resolving conversation.

ebenahar (dismissed):

fbalak (dismissed):

ebenahar (dismissed):

MeridianExplorer (approved):

PR #3444, opened 2020-11-30
WIP: Use channel '4.6' when deploying LSO '4.7'

yitzhak12:
Use channel '4.6' when deploying LSO '4.7'. 
This is a temporary solution for deploying a cluster with LSO 4.7

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -982,7 +982,7 @@ def setup_local_storage(storageclass):
     # Update local-storage-operator subscription data with channel
     for data in lso_data:
         if data["kind"] == "Subscription":
-            data["spec"]["channel"] = get_lso_channel()
+            data["spec"]["channel"] = "4.6"
         if not ocp_ga_version:
             if data["kind"] == "Subscription":
                 data["spec"]["source"] = "optional-operators"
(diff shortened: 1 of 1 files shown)


ocs_ci/deployment/deployment.py
```diff
@@ -982,7 +982,7 @@ def setup_local_storage(storageclass):
     # Update local-storage-operator subscription data with channel
     for data in lso_data:
         if data["kind"] == "Subscription":
-            data["spec"]["channel"] = get_lso_channel()
+            data["spec"]["channel"] = "4.6"
```
prsurve, line 985: Please add a check before applying channel directly

yitzhak12: What exactly do I need to check?

petr-balogh: I think it's only for your testing but you cannot directly define 4.6 here as it will also break 4.5 deployment.

If it doesn't work with 4.7 and you would like to merge this please open workaround issue to not forget this change to revert and add comment to the issue in the comment here.

I added WIP in the subject of the PR to really prevent this to be merged.

prsurve: you should add 4.6 only if you are using OCP 4.6 or OCP 4.7

clacroix12: Why do we need to do this when we are using the optional-operators for non-ga versions of LSO?

yitzhak12: If the change doesn't help, and from what I see, it doesn't, maybe I had better close the PR.

yitzhak12:
This PR is not relevant anymore

PR #3448, opened 2020-11-30
WIP: cli: Add support for env variables

b-ranto:
This commit will look for an env variable with a matching name in upper case with '-' replaced by '_' for the cli params with `OCS_CSI_` prefix.

This adds support for overriding e.g. ocp version with an `OCS_CI_OCP_VERSION` env variable.

### DIFF
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -51,6 +51,7 @@ log = logging.getLogger(__name__)
 handler = logging.StreamHandler()
 handler.setFormatter(logging.Formatter(LOG_FORMAT))
 log.addHandler(handler)
+notset = object()
 
 
 def pytest_addoption(parser):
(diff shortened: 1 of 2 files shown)


ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -363,6 +363,12 @@ def get_cli_param(config, name_of_param, default=None):
         any: value of cli parameter or default value
 
     """
+
+    # Override the default value from environmental variables
+    env_var_name = name_of_param.lstrip("-").replace("-", "_").upper()
+    if env_var_name in os.environ:
+        default = os.environ.get(env_var_name)
```
clacroix12, line 370: I think we could run into an issue if this value overwrites the `default` set when `get_cli_param` is called. I think we should only set this using env vars if `default` is `None`.

b-ranto: Sadly, this didn't work as expected in my testing. It is possible that pytest is resetting the env so the variables are not accepted, here. :-/

andrewschoen (approved):

ocs_ci/framework/pytest_customization/ocscilib.py
b-ranto: Hmm, testing locally, I could get `OCP_VERSION` to set the `--ocp-version` argument, I may need to dig a little bit deeper here.

vasukulkarni: is there no way to pass the version other then env and consume via cli from run-ci? consuming via cli is clean.

b-ranto: I added some testing output and the option is indeed passed as an environmental variable so that is not a problem.

What happens is that the option gets recognized but it is given a default value from the parser, not the default value that is defined in this function as a parameter. That default value is pretty much always ignored nowadays (as long as the option was defined). This default value from parser can be `None` (and in case of `--ocp-version` it in fact is).

@vasukulkarni The env variables support makes it a lot easier to support the options via a Jenkins job, you just define the argument there and it is passed to `run-ci`, no additional changes are necessary to support it.

zmc: Couldn't this potentially cause any number of collisions of we're just mapping env vars directly to CLI flags? Most similar mechanisms use a prefix - here's [Terraform's approach](https://www.terraform.io/docs/configuration/variables.html#environment-variables)

b-ranto: I added the `OCS_CI_` prefix to the enviromental variables to avoid (most of) the collisions.

I am still not happy about the way the values are fetched via getoption though.

clacroix12: What's wrong with `config.getoption`?

b-ranto: It works differently than you (would) think. It uses the `default` value only if the option was not _defined_ (as in via `addoption`), there is no need to pass it as a cli argument, i.e. the `default` argument in `get_cli_param` was actually never used, it always returned the default value that was set when the option was added.

There is also another thing that is wrong with it -- you can't really distinguish between the default value and the value set via an argument with it (unless you use the `notset` hack, see my latest commit, here). The `config` object does not give you an option to check for a presence of an argument (well, at least not unless you dig deeper inside its actual implementation).

vasukulkarni:
I think this will just complicate more and you should refactor better so that it doesn't collide with normal way of using cli, If we need this we have to document more on how the precedence will work and also add some unit tests.

b-ranto:
Yeah, originally I was hoping this could be a simple feature but it is turning out to be overly complicated for what it does especially when you can use `PYTEST_ADDOPTS` to add cli options through an env variable instead. I'll close this.

b-ranto:
btw: I'd still recommend addressing the `getoption` issue with the default option somehow (i.e. the value defined as default when adding an option is used instead of the value passed as an argument to the `get_cli_param` function). It currently works because it is not really used anywhere but it could bite you pretty hard in the future if it is not addressed.

petr-balogh:
@b-ranto please open the issue with description of the problem. Thanks

Also this one was discussed in the automation meeting today where I invited you, you can watch recording if you wont but I see that the PR is closed anyway. https://drive.google.com/file/d/18KRU-M-AASCiVYVr0DnBJuUY-bsBdrhN/view  thanks

PR #3464, opened 2020-12-02
Jenkinsfile: Set from to ocs-ci

b-ranto:
The pipeline shouldn't send the e-mail from the EMAIL env. We want to
support multiple comma-separated addresses in the EMAIL env and this
blocks us from doing that.

Using ocs-ci@redhat.com in `from` field instead of EMAIL env seems like
a reasonable value for this.


b-ranto:
@petr-balogh We need this to be able to use multiple addresses for our ocs-ci runs.

petr-balogh (approved):

andrewschoen (approved):

PR #3031, opened 2020-10-01
Openstack IPI platform configuration

brainfunked:

PR #3489, opened 2020-12-07
Openstack IPI platform configuration

renan-campos:
To replace https://github.com/red-hat-storage/ocs-ci/pull/3031


ocs_ci/deployment/openstack.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class OPENSTACKIPI:
+    """
+    A class to handle OpenStack IPI specific deployment
+    """
+
+    def __init__(self):
+        logger.info("OPENSTACK IPI")
+        # TODO: OCP,OCS deployment
```
clacroix12, line 13: I take it this is just setting up the ground work for openstack IPI deployments. Should we expect the actual OCP/OCS deployment code in this PR or in a follow up?

renan-campos: That may be addressed in a follow up PR. This is being added just to get the Tier1 tests to run.

clacroix12: If that is the case then we may want to log that the deployment isn't actually happening, or raise an error here saying it isn't supported yet. That way we avoid confusion from anyone trying to deploy a cluster with openstack IPI.

petr-balogh:
@rachael-george hello, can you please rebase the PR? So we can move on here and get it merged?  Just update also the log message as Coady suggested as deployment is not implemented. And open issue for the deployment implementation will be also good. Which you can mention in the comment in deployment code.

petr-balogh (approved):

PR #3491, opened 2020-12-08
Update the AWSCLI pod creation to support disconnected clusters

MeridianExplorer:


dahorak (approved):

petr-balogh (dismissed):

tests/conftest.py
```diff
@@ -1782,6 +1783,8 @@ def awscli_pod_fixture(request, scope_name):
     )
     awscli_pod_dict["metadata"]["name"] = awscli_pod_name
 
+    update_container_with_mirrored_image(awscli_pod_dict)
```
petr-balogh, line 1786: BTW should not this be done with condition to run this only when we are running on disconnected cluster?

dahorak: The condition is incorporated into the `update_container_with_mirrored_image()` function: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L2413

petr-balogh: After discussion with Daniel I understand that condition is in function laready.

petr-balogh (approved):

tests/conftest.py
MeridianExplorer: Thanks for the review and help :) 
Resolved

MeridianExplorer:
[PR3491 verification.txt](https://github.com/red-hat-storage/ocs-ci/files/5667355/PR3491.verification.txt)
Verified locally on a regular cluster. Verified by @dahorak on a disconnected cluster.

Issue #3404, opened 2020-11-25
Scale run test case has hardcoded node name as compute-0


gitsridhar:
While running the following testcases:

<Module tests/e2e/scale/test_cephfs_many_files.py>
  <Class TestMillionCephfsFiles>
      <Function test_scale_million_cephfs_files[mgr]> 
      <Function test_scale_million_cephfs_files[mon]>
      <Function test_scale_million_cephfs_files[osd]>
      <Function test_scale_million_cephfs_files[mds]>

it failed to create csicephs-demo-pod:
[svenkat@nx124-ahv ~]$ oc describe pod csicephfs-demo-pod
Name:         csicephfs-demo-pod
Namespace:    openshift-storage
Priority:     0
**Node:         compute-0/**
Labels:       <none>
Annotations:  openshift.io/scc: anyuid
Status:       Pending
IP:           
IPs:          <none>
Containers:
  web-server:
    Image:        quay.io/ocsci/nginx:latest
    Port:         <none>
    Host Port:    <none>
    Environment:  <none>
    Mounts:
      /var/lib/www/html from mypvc (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-wdssw (ro)
Volumes:
  mypvc:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  cephfs-pvc
    ReadOnly:   false
  default-token-wdssw:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  default-token-wdssw
    Optional:    false
QoS Class:       BestEffort
Node-Selectors:  <none>
Tolerations:     node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                 node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:          <none>

The problem is with Node on which this pod is supposed to run is named hardcoded as 'compute-0'. In our case, the node names are different:
[svenkat@nx124-ahv ~]$ oc get nodes
NAME       STATUS   ROLES    AGE     VERSION
master-0   Ready    master   7h43m   v1.19.0+9f84db3
master-1   Ready    master   7h43m   v1.19.0+9f84db3
master-2   Ready    master   7h43m   v1.19.0+9f84db3
worker-0   Ready    worker   7h31m   v1.19.0+9f84db3
worker-1   Ready    worker   7h32m   v1.19.0+9f84db3
[...]

PR #3499, opened 2020-12-10
Invalid node name - compute-0

lukebrowning:
Node name does not have to be specified for the testcase.  It doesn't matter which worker node is used to run the pod, so don't specify a node name.  Tested on a power server running the ocs-ci scale testsuite.

Signed-off-by: Luke Browning <lukebrowning@us.ibm.com>


vavuthu (approved):

keesturam (approved):

keesturam:
Fixes issue #3404

Issue #3507, opened 2020-12-11
Move test results breakdown to be on the top of the execution email


dahorak:
Based on discussion on yesterday's Ecosystem weekly meeting, it will be helpful to have main test result summary on top of the email and not after squad analysis, to be immediately visible.

It might be also helpful to include percentage pass in subject.

PR #3508, opened 2020-12-11
email report enhancement

dahorak:
- move summary (passed, failed, skipped...) to the top of the email report
- add passed percentage to the subject of email report

fixes: https://github.com/red-hat-storage/ocs-ci/issues/3507

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -1024,7 +1024,7 @@ def add_squad_analysis_to_email(session, soup):
                         skipped_message = "--unknown--"
                     skipped["UNASSIGNED"].append((result.nodeid, skipped_message))
 
-    # no failed or skipped tests - exist the function
+    # no failed or skipped tests - exit the function
     if not failed and not skipped:
         return
 
@@ -1044,6 +1044,7 @@ def add_squad_analysis_to_email(session, soup):
         font-family: monospace;
         background-color: #eee;
         padding: 5px;
+        margin-top: 10px;
     }
     .squad-analysis h2 {
         margin: 0px;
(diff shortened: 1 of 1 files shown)


dahorak:
With this change, the email report subject and top part will looks like this:
email_report_enhancement

petr-balogh (dismissed):

vavuthu (approved):

ocs_ci/utility/utils.py
```diff
@@ -1124,11 +1125,41 @@ def add_squad_analysis_to_email(session, soup):
...
     Email results of test run
 
     """
+    # calculate percentage pass
+    reporter = session.config.pluginmanager.get_plugin("terminalreporter")
+    passed = len(reporter.stats.get("passed", []))
+    failed = len(reporter.stats.get("failed", []))
+    error = len(reporter.stats.get("error", []))
+    total = passed + failed + error
+    percentage_passed = (passed / total) * 100
```
vasukulkarni, line 1161: :+1:

vasukulkarni (approved):

dahorak:
Email from verification job:

pr3508_verification

ocs_ci/utility/utils.py
```diff
@@ -1139,7 +1170,8 @@ def email_reports(session):
     msg["Subject"] = (
         f"ocs-ci results for {get_testrun_name()} "
         f"({build_str}"
-        f"RUN ID: {config.RUN['run_id']})"
+        f"RUN ID: {config.RUN['run_id']}) "
+        f"Passed: {percentage_passed:.0f}%"
```
mbukatov, line 1174: I would rather not change the subject of the email. This is basically your API, and as such it should be kept stable.

mbukatov (changes requested):
I don't like change in email subject format.

ocs_ci/utility/utils.py
vasukulkarni: I think percentage pass is great indicator without opening email for high level overview, what will break with addition of pass rate?

mbukatov: I'm processing our email archive automatically, so I'm not a fan of changes which provides no additional value and requires changes in my script.

vasukulkarni: I am not sure why you want to block this one for minor addition here, we also have junit for script parsing and the subject was meant mostly for user consumption. we can discuss this in our call again to get +1 , -1 on this.

@dahorak if you want to comment this line for now I think we can take this up next week, instead of holding the PR for long.

dahorak: Commented out in ad6fb10.

vasukulkarni (approved):

mbukatov (approved):

PR #3543, opened 2020-12-16
Skipping mcg related tests for openshift dedicated acceptance testing

kesavanvt:
**Description**

- This PR has the changes to skip the MCG related tests as part of openshift dedicated acceptance testing.
- `conf/ocsci/osd_cluster.yaml` will be passed as a value for `--ocsci-conf` parameter

### DIFF
--- a/conf/ocsci/openshift_dedicated.yaml
+++ b/conf/ocsci/openshift_dedicated.yaml
@@ -1,0 +1,7 @@
+---
+ENV_DATA:
+  platform: 'openshiftdedicated'
+  deployment_type: 'managed'
+
+DEPLOYMENT:
+  ocs_csv_channel: "alpha"
--- a/tests/manage/mcg/test_bucket_deletion.py
+++ b/tests/manage/mcg/test_bucket_deletion.py
@@ -21,11 +21,13 @@ from ocs_ci.ocs.bucket_utils import (
 )
 from ocs_ci.helpers.helpers import create_unique_resource_name
 from ocs_ci.framework.testlib import MCGTest
+from ocs_ci.framework.pytest_customization.marks import skipif_openshift_dedicated
 
 logger = logging.getLogger(__name__)
 ERRATIC_TIMEOUTS_SKIP_REASON = "Skipped because of erratic timeouts"
 
 
+@skipif_openshift_dedicated
 class TestBucketDeletion(MCGTest):
     """
     Test bucket Creation Deletion of buckets
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -181,6 +181,11 @@ skipif_bm = pytest.mark.skipif(
     reason="Test will not run on Bare Metal",
 )
 
+skipif_openshift_dedicated = pytest.mark.skipif(
+    config.ENV_DATA["platform"].lower() == "openshiftdedicated",
+    reason="Test will not run on Openshift dedicated cluster",
+)
+
 skipif_external_mode = pytest.mark.skipif(
     config.DEPLOYMENT.get("external_mode") is True,
     reason="Test will not run on External Mode cluster",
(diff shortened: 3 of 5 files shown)


mykaul:
I would add a decorator for ```@skip_if_no_mcg``` as we expect in the future, no only in Dedicated, to have MCG as an optional component (deployed in a day 2 operation)
@RazTamir  - we discussed this today. Perhaps not in this patch, but surely we need this.

conf/ocsci/osd_cluster.yaml
```diff
@@ -0,0 +1,3 @@
+---
+ENV_DATA:
+  platform: 'osd'
```
vasukulkarni, line 3: not to be confused with ceph osd which is used in other places in the script, we should name it "OpenshiftDedicated"

kesavanvt: Agreed and modified!

tests/manage/mcg/test_bucket_deletion.py
```diff
@@ -21,11 +21,13 @@
 )
 from ocs_ci.helpers.helpers import create_unique_resource_name
 from ocs_ci.framework.testlib import MCGTest
+from ocs_ci.framework.pytest_customization.marks import skipif_openshift_dedicated
 
 logger = logging.getLogger(__name__)
 ERRATIC_TIMEOUTS_SKIP_REASON = "Skipped because of erratic timeouts"
 
 
+@skipif_openshift_dedicated
```
petr-balogh, line 30: Do we want to see this test case in skipped in each of execution run on OSD?

Other approach is to deselect this test case from collection and this case it will not be visible in collection and summary at all?

If we don't want to even collect this test case for OSD and not be part of skips in the execution I suggest to use different approach like we do here:

https://github.com/red-hat-storage/ocs-ci/blob/master/tests/conftest.py#L113

Which will deselect the test case from collection and it will not be visible.

@ebenahar @RazTamir @vasukulkarni  FYI

ebenahar: +1 Petr on deselcting instead of skipping. 
We can follow the same approach as done for RGW tests https://github.com/red-hat-storage/ocs-ci/blob/5b41b78a63433ac2ae8437a791a8a3644322a624/tests/manage/rgw/conftest.py#L8 on non on-prem platforms

conf/ocsci/openshift_dedicated.yaml
```diff
@@ -0,0 +1,3 @@
+---
+ENV_DATA:
+  platform: 'openshiftdedicated'
```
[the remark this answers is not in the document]
kesavanvt: Done, Thanks @petr-balogh 
Will update the other parameters once the we proceed with e2e testing for Openshift dedicated, this patch is only specific to run acceptance test on an already existing openshift dedicated cluster.

tests/manage/mcg/test_bucket_deletion.py
kesavanvt: @petr-balogh @ebenahar This patch is a temporary solution to skip the mcg related tests, in future openshift dedicated may support for mcg and its okay to see them skipped in the test runs.

ebenahar (dismissed):
Discussed offline, since this is a temp solution until OSD will support MCG, approving as is

mykaul:
> Discussed offline, since this is a temp solution until OSD will support MCG

Or it may never support it. Generally, we need to learn how to work with or without optional components deployed - MCG, RGW, CephFS are all optional components (or will be).

tests/manage/mcg/test_bucket_deletion.py
RazTamir: +1 to Petr suggestion.
Skips for such reasons are just polluting the reports

RazTamir:
> > Discussed offline, since this is a temp solution until OSD will support MCG
> 
> Or it may never support it. Generally, we need to learn how to work with or without optional components deployed - MCG, RGW, CephFS are all optional components (or will be).

This should come from the product side.
We should not implement future features in our automation

ebenahar (approved):

RazTamir:
Following the agreement as a temporary solution, approving this PR

RazTamir (approved):

ebenahar:
Verified here https://storage-jenkins-csb-ceph.cloud.paas.psi.redhat.com/job/ocs-ci/185/

PR #3544, opened 2020-12-16
Verify thorough PV backingstore deletion

MeridianExplorer:
Up until now, we did not verify that the PV was deleted after a PV backingstore was deleted.
We did confirm that the related pod and PVC were deleted, but sometimes, a PV is left for a few seconds after the PVC is gone, which has caused false failures in teardowns.
This PR verifies that the PV is also deleted thoroughly, before moving on to the rest of the teardown.

### DIFF
--- a/ocs_ci/ocs/resources/backingstore.py
+++ b/ocs_ci/ocs/resources/backingstore.py
@@ -17,7 +17,7 @@ from ocs_ci.ocs.bucket_utils import (
 from ocs_ci.ocs.exceptions import TimeoutExpiredError
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.framework import config
-from ocs_ci.helpers.helpers import create_unique_resource_name
+from ocs_ci.helpers.helpers import create_unique_resource_name, get_all_pvs
 from ocs_ci.ocs.resources.pod import get_pods_having_label
 from ocs_ci.ocs.resources.pvc import get_all_pvcs
 from ocs_ci.utility.utils import TimeoutSampler
@@ -141,9 +141,10 @@ def check_resources_deleted(backingstore_name, namespace=None):
         bool: True if pvc(s) were deleted
 
     """
+    pvs = get_all_pvs(namespace=namespace, selector=f"pool={backingstore_name}")
     pvcs = get_all_pvcs(namespace=namespace, selector=f"pool={backingstore_name}")
     pods = get_pods_having_label(namespace=namespace, label=f"pool={backingstore_name}")
-    return True if len(pvcs["items"]) == 0 and len(pods) == 0 else False
+    return len(pvcs["items"]) == 0 and len(pvs["items"]) == 0 and len(pods) == 0
 
 
 def backingstore_factory(request, cld_mgr, mcg_obj, cloud_uls_factory):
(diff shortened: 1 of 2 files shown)


MeridianExplorer:
[3544 verification.txt](https://github.com/red-hat-storage/ocs-ci/files/5701532/3544.verification.txt)
Verified locally.

ocs_ci/ocs/resources/backingstore.py
```diff
@@ -48,9 +49,16 @@ def __init__(
         self.mcg_obj = mcg_obj
         self.vol_num = vol_num
         self.vol_size = vol_size
+        self.pv_name = None
 
     def delete(self):
         log.info(f"Cleaning up backingstore {self.name}")
+        # If the backingstore utilizes a PV, save its PV name for deletion verification
+        if "pv-backingstore" in self.name.lower():
```
ebenahar, line 57: is name the best way to look for a PV backingstore?

MeridianExplorer: From an object perspective - yes. All PV BSs created through our factory necessarily contain this string.
Added a `type` attribute that's set upon creation by the factory.

MeridianExplorer:
[3544 verification.txt](https://github.com/red-hat-storage/ocs-ci/files/5702061/3544.verification.txt)
Verified locally

ocs_ci/ocs/resources/backingstore.py
ebenahar: would it be possible to get the information from the BS resource itself?

MeridianExplorer: Yes; Although that'd add a few seconds of overhead for every `get` operation performed per backingstore that needs to be cleaned up

ebenahar: that doesn't sound very bad and the advantage is a deterministic way of getting the information

MeridianExplorer: The `type` attribute that's set by the factory is also fully deterministic, since we pass it upon the backingstore creation alongside other vars, such as `interface`. There is absolutely no way to create a discrepancy between the `type` attribute and the type of the backingstore

ebenahar: got you. So if it's solid, I am ok with any of these otions

MeridianExplorer:
[3544 verification (test_bucket_creation).txt](https://github.com/red-hat-storage/ocs-ci/files/5707279/3544.verification.test_bucket_creation.txt)
Verified all of the non-skipped tests in `test_bucket_creation`.

ocs_ci/ocs/resources/backingstore.py
```diff
@@ -35,6 +36,7 @@ def __init__(
         self,
         name,
         method,
+        type,
```
ebenahar, line 39: this is a new arg added to this function which does not have a default value. 
Please make sure it doesn't break existing usages and also, please run tier1 for verification

MeridianExplorer: It doesn't - the only two are in lines 242 and 270 of this file, and were changed.
I ran `test_bucket_creation` which tests this thoroughly; I'll add a tier1 run soon.

bernerhat (approved):

ebenahar (approved):

PR #3550, opened 2020-12-17
Print OBC YAML and `oc describe` if it fails to bind on creation

MeridianExplorer:
Seems like OBCs are not always included in the `must-gather`, so this change allows us to have a better idea of what was their state at the time of the failure

### DIFF
--- a/ocs_ci/ocs/resources/objectbucket.py
+++ b/ocs_ci/ocs/resources/objectbucket.py
@@ -12,7 +12,7 @@ from ocs_ci.helpers.helpers import (
 
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants
-from ocs_ci.ocs.exceptions import CommandFailed, TimeoutExpiredError
+from ocs_ci.ocs.exceptions import CommandFailed, TimeoutExpiredError, UnhealthyBucket
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs.bucket_utils import retrieve_verification_mode
 from ocs_ci.ocs.utils import oc_get_all_obc_names
@@ -205,9 +205,6 @@ class ObjectBucket(ABC):
             timeout (int): Timeout for the check, in seconds
             interval (int): Interval to wait between checks, in seconds
 
-        Returns:
-            (bool): True if the bucket is healthy, False otherwise
-
         """
         logger.info(f"Waiting for {self.name} to be healthy")
         try:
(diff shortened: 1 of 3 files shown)


MeridianExplorer:
[3550 failure verification.txt](https://github.com/red-hat-storage/ocs-ci/files/5707632/3550.failure.verification.txt)
Verified a failure case by using an intentionally wrong storageclass

bernerhat (dismissed):

ocs_ci/ocs/resources/objectbucket.py
```diff
@@ -222,9 +223,14 @@ def verify_health(self, timeout=60, interval=5):
             logger.error(
                 f"{self.name} did not reach a healthy state within {timeout} seconds."
             )
-            assert (
-                False
-            ), f"{self.name} did not reach a healthy state within {timeout} seconds."
+            obc_obj = OCP(kind="obc", namespace=self.namespace, resource_name=self.name)
+            obc_yaml = obc_obj.get()
+            obc_description = obc_obj.describe(resource_name=self.name)
+            assert False, (
```
ebenahar, line 229: would be good not to raise an exception here but rather return a value and let the test deal with how to handle the returned value. Think of situations in which a negative result is expected.

MeridianExplorer: [We have a function parameter for that](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/conftest.py#L1960)

ebenahar: maybe I am missing something but in https://github.com/red-hat-storage/ocs-ci/blob/df461085939bbc7497c793bc40347f5ff598a0ab/tests/conftest.py#L1994 we assert in case verify_heath returns False but here we raise an assertion exception instead of returning bool

MeridianExplorer: Actually, we do not assert in case it returns `False` - either we assert it returns `True` and the assertion passes, or it already fails the assertion inside `verify_health` -
https://github.com/red-hat-storage/ocs-ci/blob/df461085939bbc7497c793bc40347f5ff598a0ab/ocs_ci/ocs/resources/objectbucket.py#L226-L228

MeridianExplorer: We had a call; I removed the return values entirely and added a custom exception instead of using `assert False`

ebenahar (approved):

MeridianExplorer:
[3550 failure verification.txt](https://github.com/red-hat-storage/ocs-ci/files/5754188/3550.failure.verification.txt)
[3550 success verification.txt](https://github.com/red-hat-storage/ocs-ci/files/5754189/3550.success.verification.txt)

Avilir (approved):
LGTM

PR #3558, opened 2020-12-18
Adapt the performance testing ripsaw small file creation test for scale

wusui:
testing.


wusui:
There appears to be some confusion in tox because the name of this new file matches a different file in tests/e2e/performance.  I will push a pull request with a different name

Issue #3562, opened 2020-12-18
cli parameter ( ocs-vesrion) is not updated in config 


vavuthu:
cli parameter ( ocs-vesrion) is not updated in config .  When we pass ocs-version as 4.7 as paramter, config has 4.6 by default

vavuthu:
Closing this issue and it is working as expected

PR #3563, opened 2020-12-18
updating ocs_version cli parameter in config

vavuthu:
Fixes: #3562 

### DIFF
--- a/ocs_ci/framework/main.py
+++ b/ocs_ci/framework/main.py
@@ -69,6 +69,8 @@ def init_ocsci_conf(arguments=None):
     args, unknown = parser.parse_known_args(args=arguments)
     ocs_version = args.ocs_version
     load_config(args.ocsci_conf)
+    if ocs_version:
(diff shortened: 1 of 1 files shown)


ocs_ci/framework/main.py
```diff
@@ -69,6 +69,8 @@ def init_ocsci_conf(arguments=None):
     args, unknown = parser.parse_known_args(args=arguments)
     ocs_version = args.ocs_version
     load_config(args.ocsci_conf)
+    if ocs_version:
+        framework.config.ENV_DATA["ocs_version"] = ocs_version
```
clacroix12, line 73: We are loading the ocs version file on line 85 (87 with this change) which has the ocs_version set under ENV_DATA already. Do we just need to move that code higher up here?

vavuthu: correct. I didn't see ocs_version in config file.

vavuthu:
Closing this PR as it is not needed

PR #3566, opened 2020-12-18
Adapt performance testing ripsaw small file creation test for scale t…

wusui:
…esting


Avilir (commented):
@wusui this is the same as the performance test, just with different number of files.
i think we can use the performance class (for preventing code duplication) for the test and just change the numbers.

keesturam:
@wusui can you please update the description with a meaningful message.

keesturam:
@wusui please update the description

wusui:
Avi is coorect about the code duplication.  I will try to refactor this with the performance code.

I will add more information to the bug report itself.

PR #3573, opened 2020-12-21
Implement methods of disruptive operations for External mode cluster

sidhant-agrawal:
This PR has the following changes related to disruptive operations utilities and tests:
    * Update disruption_helpers.py to use factory design and implement methods for external mode cluster
    * Update methods in utils.py to work with any Ceph daemon
    * Refactor code in existing tests to align with factory design of disruption_helpers.py

### DIFF
--- a/ocs_ci/helpers/disruption_helpers.py
+++ b/ocs_ci/helpers/disruption_helpers.py
@@ -1,7 +1,10 @@
 import logging
+import random
+import re
+import time
 
 from ocs_ci.ocs.resources import pod
-from ocs_ci.ocs import constants, ocp
+from ocs_ci.ocs import constants, ocp, utils
 from ocs_ci.framework import config
 from ocs_ci.utility.utils import TimeoutSampler, run_async, run_cmd
 from ocs_ci.utility.retry import retry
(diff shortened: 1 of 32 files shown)


ocs_ci/helpers/disruption_helpers.py
```diff
@@ -162,3 +165,70 @@ def kill_daemon(self, node_name=None, check_new_pid=True):
...
+            mds_stat = ct_pod.exec_ceph_cmd("ceph mds stat", format="plain")
+            self.resource_id = re.search(r"(?<=0=).*?(?==up:active)", mds_stat).group(0)
+        else:
+            self.resource_id = random.randint(0, self.resource_count)
+
+        log.info(f"Selected daemon: ceph-{self.resource}@{self.resource_id}")
+
+    def delete_resource(self, wait_time=60):
+        """
+        Simulate pod deletion by stop and start ceph daemon
```
ebenahar, line 207: I would change to something like:
```Stop and start Ceph daemon, which is the equivalent of pod deletion in internal mode```

ebenahar (commented):
Great job @sidhant-agrawal!!

ocs_ci/helpers/disruption_helpers.py
sidhant-agrawal: updated the docstring

ebenahar (dismissed):

jilju (dismissed):

ebenahar (approved):
Amazing work Sidhant

petr-balogh:
@sidhant-agrawal @ebenahar one question about disruptive operations on external cluster - if we are running parallel executions on external cluster will those be affected? If so we need to come up with some mechanism how we will prevent using external cluster in parallel executions.   Please if that's the case and it's problematic, ping me somewhere next week and we can discuss this.  Thanks

sidhant-agrawal:
@petr-balogh yes, parallel executions involving disruptive test might affect each other. 
For eg. The Ceph health status might change due to daemon failures in one execution and this will lead to Ceph health check failure in other execution.
IMO, we need some lockable resource for external RHCS, so that we don't run parallel jobs for external mode once we merge this PR.
@vavuthu fyi, Incase you already started working on this lockable resource

petr-balogh:
@vavuthu any update? Thanks @sidhant-agrawal for details. I think that this should not be merged before such logic is implemented, otherwise our production executions which running in parallel will get into the troubles.

vavuthu:
> @vavuthu any update? Thanks @sidhant-agrawal for details. I think that this should not be merged before such logic is implemented, otherwise our production executions which running in parallel will get into the troubles.

raised MR for the same

PR #3591, opened 2020-12-28
Handle GCP ULS deletion race condition

MeridianExplorer:
As part of `test_bucket_deletion`, we delete an OBC containing objects.
This means that NB has to delete all the objects by itself once the OBC is deleted.
However, we proceed to try and clean up + delete the ULS once the test is done, and it seems like we ran into a race condition where both we and NB try to delete the same blobs.
This PR should retry blob deletion until the bucket is completely empty and ready for removal.


bernerhat (approved):

ebenahar (approved):

MeridianExplorer:
All tests passed, the Jenkins pipeline failed because of a Jenkins bug. Verified

ebenahar (approved):

PR #3609, opened 2020-12-31
Test 2 storage class using same pool with replica 2 and compression

shyRozen:
Test 2 storage class using same pool with replica 2 and compression

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -365,6 +365,7 @@ def storageclass_factory_fixture(
         replica=3,
         compression=None,
         new_rbd_pool=False,
+        pool_name=None,
     ):
         """
         Args:
@@ -379,6 +380,8 @@ def storageclass_factory_fixture(
             replica (int): Replica size for a pool
             compression (str): Compression type option for a pool
             new_rbd_pool (bool): True if user wants to create new rbd pool for SC
+            pool_name (str): Existing pool name to create the storageclass other
+                then the default rbd pool.
 
         Returns:
             object: helpers.create_storage_class instance with links to
@@ -397,7 +400,10 @@ def storageclass_factory_fixture(
                     )
                     interface_name = pool_obj.name
                 else:
-                    interface_name = helpers.default_ceph_block_pool()
+                    if pool_name is None:
+                        interface_name = helpers.default_ceph_block_pool()
+                    else:
+                        interface_name = pool_name
             elif interface == constants.CEPHFILESYSTEM:
                 interface_name = helpers.get_cephfs_data_pool_name()
 
--- a/tests/manage/storageclass/test_create_2_sc_with_1_pool_comp_rep2.py
+++ b/tests/manage/storageclass/test_create_2_sc_with_1_pool_comp_rep2.py
@@ -1,0 +1,98 @@
+import logging
+import pytest
+from ocs_ci.ocs.resources.pod import get_fio_rw_iops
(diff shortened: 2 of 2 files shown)


tests/manage/storageclass/test_create_2_sc_with_1_pool_comp_rep2.py
```diff
@@ -0,0 +1,122 @@
...
+
+        log.info("Creating new pool with replica2 and compression")
+        pool_obj = ceph_pool_factory(
+            interface=CEPHBLOCKPOOL,
+            replica=self.replica,
+            compression="aggressive",
+        )
+
+        # loading storageclass yaml and changing parameters
+        storageclass_dict = templating.load_yaml(CSI_RBD_STORAGECLASS_YAML)
```
ebenahar, line 66: you can avoid changing this yaml here in the test by allowing `tests.conftest.storageclass_factory_fixture` to get the existing pool you created above

shyRozen: I don't see in the factory and option to pass `interface_name`. You can pass `new_rbd_pool =False` but then it will take the default rbd pool as I understand from the code. What I can do if I'm right in the above is change the function to deal with existing rbd pool.

ebenahar: I meant that we should add a new argument to the factory which is the pool name to use

shyRozen: Great. Will do

tests/conftest.py
```diff
@@ -379,6 +380,8 @@ def factory(
             replica (int): Replica size for a pool
             compression (str): Compression type option for a pool
             new_rbd_pool (bool): True if user wants to create new rbd pool for SC
+            pool_name (str): Existing pool name to create the storageclass other
+                             then the default rbd pool.
```
ebenahar, line 384:
```suggestion
            pool_name (str): Existing pool name to create the storageclass other
                then the default rbd pool.
```

ebenahar (approved):

tests/manage/storageclass/test_create_2_sc_with_1_pool_comp_rep2.py
```diff
@@ -0,0 +1,98 @@
...
+@skipif_external_mode
+@skipif_ocs_version("<4.6")
+@pytest.mark.polarion_id("OCS-2391")
+class TestMultipleScOnePoolRep2Comp(ManageTest):
+    """
+    Create new rbd pool with replica 2 and compression.
+    Attach it to 2 new storageclasses.
+    Create PVCs and PODs for each storageclass.
+    Run IO.
+    Delete PODs, PVDs, Storageclass and pool.
```
Avilir, line 29: Does PVDs need to be PVCs ?

Avilir (approved):
just one small comment, but looking good

tests/manage/storageclass/test_create_2_sc_with_1_pool_comp_rep2.py
shyRozen: For Sure. Will push a fix -Thanks!!!

Avilir (approved):

ebenahar (approved):

PR #3618, opened 2021-01-04
[Openshift-Dedicated] Exclude openshift-build-test namespace on teardown

kesavanvt:
**Description**
This fixes the tear down failure on various tests in openshift-dedicated, by ignoring the openshift-build-test namespace. As multiple namespace of openshift-build-test* pattern gets created even after the successful cluster creation which then interferes with the OCS-CI tear down process.

### DIFF
--- a/ocs_ci/utility/environment_check.py
+++ b/ocs_ci/utility/environment_check.py
@@ -112,6 +112,11 @@ def assign_get_values(env_status_dict, key, kind=None, exclude_labels=None):
             if name == "openshift-must-gather-":
                 log.debug(f"ignoring item: {constants.NAMESPACE} with name {name}")
                 continue
+        if item.get("kind") == constants.NAMESPACE:
+            name = item.get("metadata").get("name")
+            if name.startswith(defaults.SRE_BUILD_TEST_NAMESPACE):
+                log.debug(f"ignoring item: {constants.NAMESPACE} with name {name}")
+                continue
         items_filtered.append(item)
 
     ignored = len(items) - len(items_filtered)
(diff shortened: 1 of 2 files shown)


jimdevops19 (approved):
LGTM

ocs_ci/utility/environment_check.py
```diff
@@ -112,6 +112,11 @@ def assign_get_values(env_status_dict, key, kind=None, exclude_labels=None):
             if name == "openshift-must-gather-":
                 log.debug(f"ignoring item: {constants.NAMESPACE} with name {name}")
                 continue
+        if item.get("kind") == constants.NAMESPACE:
+            name = item.get("metadata").get("name")
+            if name.startswith("openshift-build-test"):
+                log.debug(f"ignoring item: {constants.NAMESPACE} with name {name}")
+                continue
```
ebenahar, line 119: for ignoring a namespace, you should be able to append `"openshift-build-test"` to:
https://github.com/red-hat-storage/ocs-ci/blob/6f1375c5f95dd0b882d7dd8557749baa9bf82e19/ocs_ci/utility/environment_check.py#L97

As follows:

```python
and ns.startswith(("openshift-", defaults.BG_LOAD_NAMESPACE, "openshift-build-test"))
```

Please also create a var in defaults.py for `"openshift-build-test"`

ebenahar: I actually wonder how come `"openshift-"` does not cover the also `"openshift-build-test"`

kesavanvt: @ebenahar the above `ns` represents the namespace belonging to the available resources and with this only the resources under namespace will be ignored and not the namespace itself.
` ns = item.get("metadata", {}).get("namespace") `
Thus for ignoring the namespace, separate check needs to be done for kind `namespace` 

and var creation for `openshift-build-test` is done

kesavanvt: @ebenahar , yes the check on `openshift- `will eliminate only the resources under `openshift-` and not the `openshift- `namespace itself

ebenahar (approved):

jilju (approved):

PR #3633, opened 2021-01-07
Return the MCG resources created for the current factory scope instead of all scopes

bernerhat:
we currently return all created buckets and backingstores in these factories instead we should only return the currently requested objects.


MeridianExplorer (approved):
🏭

ebenahar (dismissed):

shreekarSS (approved):

MeridianExplorer (approved):

ebenahar (approved):

PR #3636, opened 2021-01-07
flexy: update pulp URL

ktdreyer:
The pulp.dist URL is going away soon. Use the new RHSM Pulp URL.


ktdreyer:
By the way, if this code is still in use, it would be great to set `'gpgcheck': 1` here instead of `0`.

vavuthu (approved):

petr-balogh (approved):

PR #3640, opened 2021-01-08
[WIP] modifying timeouts

AaruniAggarwal:
Signed-off-by: Aaruni Aggarwal <aaruniagg@gmail.com>

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -100,6 +100,7 @@ def wait_for_nodes_status(node_names=None, status=constants.NODE_READY, timeout=
             reached the desired state
 
     """
+    timeout = 1800
     try:
         if not node_names:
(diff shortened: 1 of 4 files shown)


keesturam:
@AaruniAggarwal I see that the timeouts are changed to 1800 seconds in most of the cases. I don't think this should be the approach. can you please raise issues (or link if there is already one) on where you see failures. Tests taking more time may imply a problem in the underlying cluster.

AaruniAggarwal:
There is no opened issue as of now. 
After doing hugepage changes , now we are getting less timeout errors. I will close this PR and will create a new one with less timeout changes.

AaruniAggarwal:
This is my new PR that I created : 
https://github.com/red-hat-storage/ocs-ci/pull/3718

PR #3642, opened 2021-01-08
Check if new alerts are raised during test

fbalak:
- Print warning if new alerts are raised during test to help with test analysis.
- Most of the code is in try, except blocks to prevent fails if Prometheus is not available.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -66,6 +66,7 @@ from ocs_ci.utility.environment_check import (
     get_status_before_execution,
     get_status_after_execution,
 )
+from ocs_ci.utility.prometheus import PrometheusAPI
 from ocs_ci.utility.uninstall_openshift_logging import uninstall_cluster_logging
 from ocs_ci.utility.utils import (
     ceph_health_check,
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -2719,6 +2720,57 @@ def user_factory_session(request, htpasswd_identity_provider, htpasswd_path):
     return users.user_factory(request, htpasswd_path)
 
 
+@pytest.fixture(autouse=True)
+def log_alerts(request):
```
ebenahar, line 2724: great idea!! will help with both failure analysis and with alerts validations

tests/conftest.py
```diff
@@ -2719,6 +2720,57 @@ def user_factory_session(request, htpasswd_identity_provider, htpasswd_path):
...
+    prometheus = None
+
+    try:
+        prometheus = PrometheusAPI()
+    except Exception as e:
+        log.warning(f"There was a problem with connecting to Promeheus: {e.message}")
+
+    def _collect_alerts():
+        try:
+            alerts_response = prometheus.get(
```
ebenahar, line 2741: we have seen in the past that in rare occasions, `prometheus.get()` can fail with one of `IndexError` of `ScannerError`.
See https://github.com/red-hat-storage/ocs-ci/blob/c950799f545b3b84d3539302a4b3abd475a7fcce/ocs_ci/ocs/cluster_load.py#L339

Since we this fixture will run for each and every test case, will it make sense to add the same retry decorator here as well?

fbalak: The query slightly differs and I haven't seen this error for this query from monitoring tests. I would rather not put it there because there are times that this should fail (when monitoring is not enabled - although it would probably not fail on index error) and add 15*5 wait for failed query could slow down some test runs too much.

mbukatov: Since this is going to be executed for every test to make additional evidence available, I would rather tried once and just report an error in case or problems, instead of trying multiple times and possible extending runtime of the whole suite. Also unless monitoring is disabled, broken or a whole cluster under high load, it should work fine.

mbukatov (approved):

tests/conftest.py
ebenahar: ack

ebenahar (approved):

PR #3674, opened 2021-01-13
adding new marker for regression tests

SvetlanaRedHat:
This pr is to make new marker for scale regression cases.

### DIFF
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -67,6 +67,7 @@ workloads = pytest.mark.workloads
 flowtests = pytest.mark.flowtests
 performance = pytest.mark.performance
 scale = pytest.mark.scale
+scale_regression = pytest.mark.scale_regression
 deployment = pytest.mark.deployment
 polarion_id = pytest.mark.polarion_id
 bugzilla = pytest.mark.bugzilla
--- a/pytest.ini
+++ b/pytest.ini
@@ -22,6 +22,7 @@ markers =
     flowtests: flowbased related tests
     performance: performance related tests
     scale: scale related tests
+    scale_regression: scale regression related tests
     deployment: deployment related tests
     ocs_upgrade: marker for OCS upgrade test
     ocp_upgrade: marker for OCP upgrade test
--- a/tests/e2e/scale/test_ceph_pod_respin_in_scaled_cluster.py
+++ b/tests/e2e/scale/test_ceph_pod_respin_in_scaled_cluster.py
@@ -5,7 +5,7 @@ from ocs_ci.helpers import disruption_helpers
 from ocs_ci.ocs import constants
 from ocs_ci.utility import utils
 from ocs_ci.ocs.scale_lib import FioPodScale
-from ocs_ci.framework.testlib import scale, E2ETest, ignore_leftovers
+from ocs_ci.framework.testlib import scale, E2ETest, ignore_leftovers, scale_regression
 from ocs_ci.framework.pytest_customization.marks import skipif_external_mode
 
 log = logging.getLogger(__name__)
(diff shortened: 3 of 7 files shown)


Avilir (commented):
i think you need to add it to pytets.ini file as well

wusui:
I think Avi is correct, but it is pytest.ini

add "regression: regression marker" (or some such comment) to the end of marker =

I am not sure if it messes things up if it  is added somewhere else.

mbukatov:
Add explanation of exact meaning of this marker into the docs. Explanation why we need this could be also included here in the PR. The regression is a general term, and without further clarification, it won't be as useful as it could be.

vasukulkarni:
what is the purpose for this one?

tiffanyn108:
> what is the purpose for this one?

This marker will be used to run scale regression tests.  We should name as "scale_regression" to make it more clear about the purpose of the marker.

tiffanyn108:
@SvetlanaRedHat There should be more test cases which needs to add scale regression marker.  Please review the spreadsheet for more info.

tiffanyn108:
Need to validate this PR which Jenkins run.

vasukulkarni:
> > what is the purpose for this one?
> 
> This marker will be used to run scale regression tests. We should name as "scale_regression" to make it more clear about the purpose of the marker.

Just "scale" should be sufficient in that case

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -67,6 +67,7 @@
 flowtests = pytest.mark.flowtests
 performance = pytest.mark.performance
 scale = pytest.mark.scale
+regression = pytest.mark.regression
```
vasukulkarni, line 70: please rename this to just "scale" for clarity

tiffanyn108:
> > > what is the purpose for this one?
> > 
> > 
> > This marker will be used to run scale regression tests. We should name as "scale_regression" to make it more clear about the purpose of the marker.
> 
> Just "scale" should be sufficient in that case

We already have scale label for all scale tests.  We need a subset of scale test to run as scale regression.

ocs_ci/framework/pytest_customization/marks.py
vasukulkarni: > > > > what is the purpose for this one?
> > > 
> > > 
> > > This marker will be used to run scale regression tests. We should name as "scale_regression" to make it more clear about the purpose of the marker.
> > 
> > 
> > Just "scale" should be sufficient in that case
> 
> We already have scale label for all scale tests. We need a subset of scale test to run as scale regression.

I see that, I think we could use pytest filter to achieve what we need, I really dont see need for another marker here if its just one test?

tiffanyn108:
> > > > what is the purpose for this one?
> > > 
> > > 
> > > This marker will be used to run scale regression tests. We should name as "scale_regression" to make it more clear about the purpose of the marker.
> > 
> > 
> > Just "scale" should be sufficient in that case
> 
> We already have scale label for all scale tests. We need a subset of scale test to run as scale regression.

ocs_ci/framework/pytest_customization/marks.py
SvetlanaRedHat: its not just one test, i added small group for test for fast jenkins run, adding the rest of  tests

SvetlanaRedHat:
> > what is the purpose for this one?
> 
> This marker will be used to run scale regression tests. We should name as "scale_regression" to make it more clear about the purpose of the marker.

Done

SvetlanaRedHat:
@tiffanyn108  updated to scale_regression
@Avilir  thanks a lot for your comment, i missed that part in my PR

wusui (approved):
Looks good

Avilir (approved):
LGTM

jimdevops19:
I think it will be great to have the pr-validation executed here with `scale_regression` test mark already, should we add this value to the dropdown **before** this gets merged?

vasukulkarni:
> I think it will be great to have the pr-validation executed here with `scale_regression` test mark already, should we add this value to the dropdown **before** this gets merged?

We are running just one test with the mark, I think it can be directly used as textbox, We have scale as mark so this might lead to some confusion.

keesturam:
Closing this PR. Ram is addressing a similar change here - https://github.com/red-hat-storage/ocs-ci/pull/4543

PR #3690, opened 2021-01-14
rgw_count semantic version

clacroix12:
Use semantic version to calculate prior ocs version when determining rgw_count

### DIFF
--- a/ocs_ci/utility/rgwutils.py
+++ b/ocs_ci/utility/rgwutils.py
@@ -1,5 +1,7 @@
 import logging
 
+from semantic_version import Version
+
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


clacroix12:
Verified that this works as expected locally. The idea is to future-proof this code and use an actual version object rather than performing a hacky rounded float calculation.

```
In [23]: x = Version.coerce("4.12")

In [24]: y = f"{x.major}.{x.minor - 1}"

In [25]: y
Out[25]: '4.11'
```

vasukulkarni (approved):

vavuthu (approved):

petr-balogh (approved):

PR #3693, opened 2021-01-15
Implement Openshift Dedicated deployment

kesavanvt:


ebenahar (dismissed):

kesavanvt:
@petr-balogh

kesavanvt:
@jimdevops19

docs/supported_platforms.md
```diff
@@ -164,3 +164,44 @@ As a workaround we need to create those CRDs depends on OCP version. (this is
...
 OCP 4.5 https://github.com/openshift/csi-external-snapshotter/tree/release-4.5/config/crd
+
+## Openshift Dedicated
+
+This platform supports deployment of OCP cluster + OCS cluster on top of it.
+
+### Requirements
+
+* Have locally installed OCM CLI, follow [official documentation](https://github.com/openshift-online/ocm-cli).
+* Have locally installed osde2e CLI, follow [official documentation](https://github.com/openshift/osde2e).
```
petr-balogh, line 175: I think this is not needed anymore, right?

ocs_ci/framework/main.py
```diff
@@ -68,6 +68,8 @@ def init_ocsci_conf(arguments=None):
     parser.add_argument("--flexy-env-file", default="", help="Path to flexy env file")
     args, unknown = parser.parse_known_args(args=arguments)
     ocs_version = args.ocs_version
+    if framework.config.ENV_DATA["platform"].lower() == "openshiftdedicated":
```
petr-balogh, line 71: Can you please explain why we are setting ocs_version here to None? Maybe worth to add it also as comment here to explain why.

Anyway, later on we should load the ocs-config for specific version (non framework default) so if someone use --ocs-version parameter then we should load specific yaml file on lines 83-87.

I think that we should still load that version specific file no?  If I understand correctly user cannot actually pick the version of OCS and then they will just get what is available behind the scenes. But I think that many of tests just rely on version and we should have ability to set which version it's there no?

ocs_ci/utility/openshift_dedicated.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def login():
+    """
+    Login to OCM client
+    """
+    token = openshift_dedicated["token"]
+    cmd = f"ocm login --token={token} --url=staging"
+    logger.info("Logging in to OCM cli")
+    run_cmd(cmd, secrets=[token])
```
petr-balogh, line 27: you can use exec_cmd instead as we are trying to depricate run_cmd and rather use exec_cmd, but I will not block PR on that.

tests/conftest.py
```diff
@@ -2728,6 +2728,8 @@ def ceph_toolbox(request):
     deploy = config.RUN["cli_params"]["deploy"]
     teardown = config.RUN["cli_params"].get("teardown")
     skip_ocs = config.ENV_DATA["skip_ocs_deployment"]
+    if config.ENV_DATA["platform"].lower() == "openshiftdedicated":
+        skip_ocs = False
```
petr-balogh, line 2732: Can you please add some comment why we do that.  As we discussed offline I would like to see if we will run post deployment verification which we have in our code but I see that the OCS deployment is done behind the scenes for us and that we will probably not do the changes we do for deployment as well but as we discussed after the deployment we are running some extra code for monitoring and so on which in this case will not be executed.

petr-balogh (approved):

docs/supported_platforms.md
kesavanvt: yes, removed

ocs_ci/utility/openshift_dedicated.py
kesavanvt: Noted, will use exec_cmd in future cases.

tests/conftest.py
kesavanvt: Thanks, Added comments in the code base.

ocs_ci/framework/main.py
kesavanvt: passing --ocs-version in run-ci overrides the `ocs_csv_channel` and thus causing issues in `get_ocs_csv()` execution since the `ocs_csv_channel` in openshfit dedicated is "alpha". The above changes was to avoid the override of `ocs_csv_channel` and wasn't working as expected . I agree that making ocs_version to None will not actually pick the version of OCS and thus as an alternative approach extracted OCS CSV from the available CSVs instead of pulling from packagemanifests in `get_ocs_csv()`.

ocs_ci/utility/openshift_dedicated.py
```diff
@@ -0,0 +1,124 @@
...
+    logger.info("Successfully logged in to OCM")
+
+
+def create_cluster(cluster_name):
```
jimdevops19, line 31: I see that cluster_name parameter is not used here, do we need to receive this as a param?

ocs_ci/ocs/resources/ocs.py
```diff
@@ -224,7 +224,17 @@ def get_ocs_csv():
         subscription_plan_approval=subscription_plan_approval,
     )
     channel = config.DEPLOYMENT.get("ocs_csv_channel")
-    ocs_csv_name = ocs_package_manifest.get_current_csv(channel=channel)
```
jimdevops19, line 227: Can you explain why you moved ocs_csv_name under the else statement?
Because for a scenario where we  hit the `if` conditional, and we do not hit the conditional `item["metadata"]["name"].startswith(defaults.OCS_OPERATOR_NAME):` we could end up with `ocs_csv_name` not being assigned, and this line will receive an error:
`ocs_csv = CSV(resource_name=ocs_csv_name, namespace=namespace)`

jimdevops19 (changes requested):

ocs_ci/utility/openshift_dedicated.py
kesavanvt: Thanks, removed

ocs_ci/ocs/resources/ocs.py
kesavanvt: If the ocs_csv_name is not assigned, means the OCS csv is not available and have made the changes to raise for the same. Moving `ocs_csv_name` under else to make sure it only executes for non openshift dedicated platforms. In case of openshift dedicated the OCS CSV is pulled from available CSVs instead of packagemanifests.

ocs_ci/ocs/resources/ocs.py
```diff
@@ -224,7 +228,20 @@ def get_ocs_csv():
...
+    ocs_csv_name = None
+    # OCS CSV is extracted from the available CSVs in cluster namespace
+    # for Openshift dedicated platform
+    if config.ENV_DATA["platform"].lower() == "openshiftdedicated":
+        ocp_cluster = OCP(
+            namespace=config.ENV_DATA["cluster_namespace"], kind="", resource_name="csv"
```
petr-balogh, line 236: should not this be changed to:
```suggestion
            namespace=config.ENV_DATA["cluster_namespace"], kind="csv"
```
As you would like to iterate over the CSV objects?

This you should get all kind of csv, and then you can iterate over them.

Just a side note:
When the upgrade happens, you can have 2 CSVs at the same time, one is in replacing state and second one in installing.  But I do not expect that you will somehow test upgrade of OCS on openshift dedicated, right?

tests/conftest.py
petr-balogh: Ah, this case I would rather change the code of the condition to proposed in the other comment

tests/conftest.py
```diff
@@ -2728,6 +2728,12 @@ def ceph_toolbox(request):
     deploy = config.RUN["cli_params"]["deploy"]
     teardown = config.RUN["cli_params"].get("teardown")
     skip_ocs = config.ENV_DATA["skip_ocs_deployment"]
+    # In Openshift dedicated platform, OCS is installed via osde2e tool and
+    # it is not installed via OCS-CI. Thus skip_ocs_deployment is set True.
+    # However during the test execution, Ceph toobox is not deployed when
+    # skip_ocs_deployment is True, so skip_ocs is set False here.
+    if config.ENV_DATA["platform"].lower() == "openshiftdedicated":
+        skip_ocs = False
```
petr-balogh, line 2736:
```suggestion
    deploy_teradown = deploy or teardown
    ocp_dedicated = config.ENV_DATA["platform"].lower() == "openshiftdedicated"
    if not (deploy_teardown or skip_ocs) or (ocp_dedicated and not deploy_teradown):
```
Which will be more readable that we would like to deploy toolbox pod for ocp_dedicated for test execution and not for teardown/deploy. WDYT?

ocs_ci/ocs/resources/ocs.py
kesavanvt: Fixed.
Not expecting to test upgrade on Openshift dedicated right now.

tests/conftest.py
kesavanvt: I agree on the readability and fixed accordingly

ocs_ci/ocs/resources/ocs.py
jimdevops19: Thanks :)

ocs_ci/ocs/resources/ocs.py
```diff
@@ -224,7 +228,18 @@ def get_ocs_csv():
         subscription_plan_approval=subscription_plan_approval,
     )
     channel = config.DEPLOYMENT.get("ocs_csv_channel")
-    ocs_csv_name = ocs_package_manifest.get_current_csv(channel=channel)
+    ocs_csv_name = None
+    # OCS CSV is extracted from the available CSVs in cluster namespace
+    # for Openshift dedicated platform
+    if config.ENV_DATA["platform"].lower() == "openshiftdedicated":
```
ebenahar, line 234: please save `"openshiftdedicated"` in constants.py

ebenahar (commented):
LGTM

ocs_ci/ocs/resources/ocs.py
kesavanvt: Done

kesavanvt:
Verification job (Openshift-dedicated deployment + tier 1 execution): https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/23/console

petr-balogh (approved):
LGTM!

jimdevops19 (approved):

ebenahar (approved):

Issue #3694, opened 2021-01-15
get_rgw_count count in upgrade is failing with:  argument must be a string or a number, not 'Version'


petr-balogh:
```
        verify_pods_upgraded(old_images, selector=constants.MDS_APP_LABEL, count=2)
        if config.ENV_DATA.get("platform") in constants.ON_PREM_PLATFORMS:
            rgw_count = get_rgw_count(
>               float(upgrade_version), True, float(version_before_upgrade)
            )
E           TypeError: float() argument must be a string or a number, not 'Version'
```

https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/16325/consoleFull

PR #3697, opened 2021-01-15
Fix issue when casting to float in upgrade

clacroix12:
Fixes: #3694 

### DIFF
--- a/ocs_ci/ocs/ocs_upgrade.py
+++ b/ocs_ci/ocs/ocs_upgrade.py
@@ -125,7 +125,7 @@ def verify_image_versions(old_images, upgrade_version, version_before_upgrade):
     verify_pods_upgraded(old_images, selector=constants.MDS_APP_LABEL, count=2)
     if config.ENV_DATA.get("platform") in constants.ON_PREM_PLATFORMS:
         rgw_count = get_rgw_count(
-            float(upgrade_version), True, float(version_before_upgrade)
+            upgrade_version.base_version, True, version_before_upgrade
         )
         verify_pods_upgraded(
             old_images,
--- a/ocs_ci/utility/rgwutils.py
+++ b/ocs_ci/utility/rgwutils.py
@@ -14,9 +14,9 @@ def get_rgw_count(ocs_version, is_upgrade, version_before_upgrade):
     Otherwise, RGW Count is 1
 
     Args:
-        ocs_version (str): OCS Version
+        ocs_version (str, float): OCS Version
         is_upgrade (bool): If cluster was upgraded to current version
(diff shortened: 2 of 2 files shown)


clacroix12:
I have tested this locally to ensure that get_rgw_count can be called using this method:

```
>>> from ocs_ci.utility.rgwutils import get_rgw_count
>>> from pkg_resources import parse_version
>>> ver = parse_version('4.5')
>>> ver
<Version('4.5')>
>>> ver.base_version
'4.5'
>>> get_rgw_count(ver.base_version, True, ver.base_version)
2
```

vasukulkarni (approved):

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -125,7 +125,7 @@ def verify_image_versions(old_images, upgrade_version, version_before_upgrade):
     verify_pods_upgraded(old_images, selector=constants.MDS_APP_LABEL, count=2)
     if config.ENV_DATA.get("platform") in constants.ON_PREM_PLATFORMS:
         rgw_count = get_rgw_count(
-            float(upgrade_version), True, float(version_before_upgrade)
+            upgrade_version.base_version, True, version_before_upgrade
```
petr-balogh, line 128: Is base version float or str? As you accepting those in function now.

ocs_ci/utility/rgwutils.py
```diff
@@ -14,9 +14,9 @@ def get_rgw_count(ocs_version, is_upgrade, version_before_upgrade):
     Otherwise, RGW Count is 1
 
     Args:
-        ocs_version (str): OCS Version
```
petr-balogh, line 17: Maybe add new type to unit test as well?

ocs_ci/ocs/ocs_upgrade.py
clacroix12: The function accepted both already I just updated the docstring to reflect that. `base_version` is a string:

```shell
>>> ver = parse_version('4.5')
>>> type(ver)
<class 'pkg_resources.extern.packaging.version.Version'>
>>> type(ver.base_version)
<class 'str'>
>>> ver
<Version('4.5')>
>>> ver.base_version
'4.5'
```

petr-balogh (approved):

ocs_ci/utility/rgwutils.py
clacroix12: This isn't a change in functionality, just updating the pydoc to reflect that you can pass a float or a string. We would basically be testing the `float()` function but if we feel more comfortable with them then I can add tests in that use floats instead of strings.

PR #3623, opened 2021-01-05
Replace OBC operations with factory fixtures in sanity_helpers

ebenahar:
Fixes https://github.com/red-hat-storage/ocs-ci/issues/3621

Signed-off-by: Elad Ben Aharon <ebenahar@redhat.com>

PR #3703, opened 2021-01-17
Fix the RGW test_bucket_delete flow

MeridianExplorer:
[The OB deletion flow was changed](https://github.com/red-hat-storage/ocs-ci/pull/3623/files#diff-182979f19c5a44501b1e65af59e1067af0e96e8b479ca0eb150e208459b16934), and doesn't return a value anymore.
The return value was unneeded because the deletion assertion is performed internally as part of the delete function.
This should fix a test that relied on a return value.

`verify` is True by default, but I felt it'd be good to explicitly state it so people reading the test can know that the verification is performed internally.

### DIFF
--- a/tests/manage/rgw/test_bucket_deletion.py
+++ b/tests/manage/rgw/test_bucket_deletion.py
@@ -33,7 +33,7 @@ class TestBucketDeletion:
         """
         for bucket in rgw_bucket_factory(amount, interface):
             logger.info(f"Deleting bucket: {bucket.name}")
-            assert bucket.delete()
+            bucket.delete(verify=True)
 
     @pytest.mark.parametrize(
         argnames="interface",


Avilir (approved):

ebenahar (approved):

ebenahar:
validation job is running - https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/3748/

petr-balogh (approved):

fbalak (approved):

petr-balogh:
I think this PR is not needed anymore as it has verify set to True by default as I can see here:
https://github.com/red-hat-storage/ocs-ci/blob/d9b6ee59862c0617646d0b1d53e9f8f0f65b1dcb/ocs_ci/ocs/resources/objectbucket.py#L153

Closing

Issue #3711, opened 2021-01-19
Load auth.yaml config to ocs-ci configuration


petr-balogh:
On the jenkins we are passing auth config with our custom auth data but the content is currently not loaded in framework config.

It will be good if the auth.yaml file exist to add it to AUTH section in our config.
Currently I got request to add IBM cloud token, so this will be the easiest way how we can update our credential file in jenkins and let it propagate to AUTH section from anyone can reach the data.

PR #3712, opened 2021-01-19
Load auth.yaml file to AUTH framework config

petr-balogh:
Fixes: #3711

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -76,6 +76,7 @@ from ocs_ci.utility.utils import (
     get_openshift_client,
     get_system_architecture,
     get_testrun_name,
+    load_auth_config,
     ocsci_log_path,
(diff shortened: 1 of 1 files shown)


vavuthu (approved):

clacroix12:
@petr-balogh I'm not exactly sure where but I believe we may be logging the config at some point during execution. We probably want to ensure that we aren't printing any authentication details to the console.

petr-balogh:
> @petr-balogh I'm not exactly sure where but I believe we may be logging the config at some point during execution. We probably want to ensure that we aren't printing any authentication details to the console.

I think that we already censor the content when we print that to the file.

dahorak (approved):

jimdevops19 (approved):
LGTM

PR #3739, opened 2021-01-22
[Openshift-Dedicated] Added changes related to tier 1 tests 

kesavanvt:
**Description**
This consists of the changes related to Tier 1 tests for Openshift Dedicated

### DIFF
--- a/ocs_ci/helpers/sanity_helpers.py
+++ b/ocs_ci/helpers/sanity_helpers.py
@@ -75,9 +75,10 @@ class Sanity:
         if rgw_bucket_factory:
             self.obc_objs.extend(rgw_bucket_factory(1, "rgw-oc"))
 
-        self.obc_objs.extend(bucket_factory(amount=1, interface="OC"))
+        if bucket_factory:
+            self.obc_objs.extend(bucket_factory(amount=1, interface="OC"))
 
-        self.ceph_cluster.wait_for_noobaa_health_ok()
+            self.ceph_cluster.wait_for_noobaa_health_ok()
 
     def delete_resources(self):
         """
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1771,6 +1771,9 @@ def mcg_obj_fixture(request, *args, **kwargs):
     Returns:
         MCG: An MCG resource
     """
+    if config.ENV_DATA["platform"].lower() == constants.OPENSHIFT_DEDICATED_PLATFORM:
+        log.warning("As openshift dedicated is used, no MCG resource is returned")
+        return None
 
     mcg_obj = MCG(*args, **kwargs)
 
@@ -1959,19 +1962,27 @@ def rgw_bucket_factory_session(request, rgw_obj_session):
 @pytest.fixture()
 def bucket_factory(request, bucket_class_factory, mcg_obj):
     """
-    Returns an MCG bucket factory
+    Returns an MCG bucket factory.
+    If MCG object not found returns None
     """
-    return bucket_factory_fixture(request, bucket_class_factory, mcg_obj)
+    if mcg_obj:
+        return bucket_factory_fixture(request, bucket_class_factory, mcg_obj)
+    else:
+        return None
 
 
 @pytest.fixture(scope="session")
 def bucket_factory_session(request, bucket_class_factory_session, mcg_obj_session):
     """
-    Returns a session-scoped MCG bucket factory
+    Returns a session-scoped MCG bucket factory.
+    If session-scoped MCG object not found returns None
     """
-    return bucket_factory_fixture(
-        request, bucket_class_factory_session, mcg_obj_session
-    )
+    if mcg_obj_session:
+        return bucket_factory_fixture(
+            request, bucket_class_factory_session, mcg_obj_session
+        )
+    else:
+        return None
 
 
 def bucket_factory_fixture(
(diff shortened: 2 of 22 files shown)


kesavanvt:
@jimdevops19 @petr-balogh

fbalak (commented):
I suggest to not check if parameters were provided in all noobaa fixtures in tests/conftest.py. These parameters don't have any default value so user have to provide them. I see that there is missing docstring for those parameters so it would be better to just add docstring to those parameters to describe what should be provided.

kesavanvt:
> I suggest to not check if parameters were provided in all noobaa fixtures in tests/conftest.py. These parameters don't have any default value so user have to provide them. I see that there is missing docstring for those parameters so it would be better to just add docstring to those parameters to describe what should be provided.

@fbalak In openshift dedicated, noobaa is disabled and thus checks related to mcg needs to avoided. 
Based on my observation, its required to have a check on mcg object in the required fixtures.

kesavanvt:
Verification job (Openshift-dedicated deployment + tier 1 execution): https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/23/console

fbalak:
> > I suggest to not check if parameters were provided in all noobaa fixtures in tests/conftest.py. These parameters don't have any default value so user have to provide them. I see that there is missing docstring for those parameters so it would be better to just add docstring to those parameters to describe what should be provided.
> 
> @fbalak In openshift dedicated, noobaa is disabled and thus checks related to mcg needs to avoided.
> Based on my observation, its required to have a check on mcg object in the required fixtures.

@kesavanvt Right, but if you are skipping noobaa tests then these checks should never be accessible. If those fixtures are part of some generic health check then it would be better to skip the creation of the object from factory with logged message that this doesn't work on openshift dedicated.

tests/conftest.py
```diff
@@ -1771,6 +1771,8 @@ def mcg_obj_fixture(request, *args, **kwargs):
     Returns:
         MCG: An MCG resource
     """
+    if config.ENV_DATA["platform"].lower() == constants.OPENSHIFT_DEDICATED_PLATFORM:
+        return None
```
fbalak, line 1775: Please provide a warning log message that openshift dedicated is used and the object is not created.

kesavanvt:
> > > I suggest to not check if parameters were provided in all noobaa fixtures in tests/conftest.py. These parameters don't have any default value so user have to provide them. I see that there is missing docstring for those parameters so it would be better to just add docstring to those parameters to describe what should be provided.
> > 
> > 
> > @fbalak In openshift dedicated, noobaa is disabled and thus checks related to mcg needs to avoided.
> > Based on my observation, its required to have a check on mcg object in the required fixtures.
> 
> @kesavanvt Right, but if you are skipping noobaa tests then these checks should never be accessible. If those fixtures are part of some generic health check then it would be better to skip the creation of the object from factory with logged message that this doesn't work on openshift dedicated.

@fbalak Since `test_node_maintenance` has `bucket_factory` as a parameter and its value it determined before the test method execution, its difficult to handle in the sanity_helpers for excluding the mcg related checks. Thus those checks needs to be handled in the fixtures as well. As discussed, have added the necessary warning message.

tests/conftest.py
kesavanvt: Done, Thanks

ocs_ci/helpers/sanity_helpers.py
```diff
@@ -75,9 +75,10 @@ def create_resources(
         if rgw_bucket_factory:
             self.obc_objs.extend(rgw_bucket_factory(1, "rgw-oc"))
 
-        self.obc_objs.extend(bucket_factory(amount=1, interface="OC"))
+        if bucket_factory:
+            self.obc_objs.extend(bucket_factory(amount=1, interface="OC"))
 
-        self.ceph_cluster.wait_for_noobaa_health_ok()
+            self.ceph_cluster.wait_for_noobaa_health_ok()
```
petr-balogh, line 81: Should we always wait for noobaa health OK? If so it should be one indent left. 
Or should this be in other condition like you have below to avoid this on OCP Dedicated?

tests/conftest.py
```diff
@@ -1961,17 +1964,23 @@ def bucket_factory(request, bucket_class_factory, mcg_obj):
     """
     Returns an MCG bucket factory
```
petr-balogh, line 1965: Add to description when this fixture returns None

tests/conftest.py
```diff
@@ -2176,7 +2191,12 @@ def bucket_class_factory(request, mcg_obj, backingstore_factory):
             a bucketclass
```
petr-balogh, line 2191: Add info when it returns None to docstring.

tests/conftest.py
```diff
@@ -2142,9 +2151,12 @@ def backingstore_factory(request, cld_mgr, mcg_obj, cloud_uls_factory):
             a backingstore
```
petr-balogh, line 2151: Add to the docstring when it returns None

tests/conftest.py
```diff
@@ -2192,9 +2212,12 @@ def bucket_class_factory_session(
             a bucketclass
```
petr-balogh, line 2212: The same

ocs_ci/helpers/sanity_helpers.py
kesavanvt: No, In case of no bucket factory (i.e) if its is None, its not necessary to wait for noobaa health ok. as there will be no mcg objects.

tests/conftest.py
kesavanvt: Done

tests/conftest.py
kesavanvt: Done

tests/conftest.py
kesavanvt: Done

tests/conftest.py
kesavanvt: done

petr-balogh (approved):

fbalak (approved):

ebenahar (approved):

PR #3749, opened 2021-01-25
Add OCS Installation via UI to deployment procedure

OdedViner:


ocs_ci/ocs/ui/views.py
```diff
@@ -1,49 +1,120 @@
...
+
+locators = {
+    "4.7": {
+        "login": login,
+        "deployment": deployment_4_7,
+        "pvc": pvc,
+    },
+    "4.6": {
+        "login": login,
+        "deployment": deployment_4_7,
```
ebondare, line 117: Shouldn't it be 4.6?

ocs_ci/deployment/deployment.py
```diff
@@ -623,19 +625,19 @@ def deploy_ocs_via_operator(self):
...
+        This method will deploy OCS with openshift-console vi UI.
+        """
+        setup_ui = login_ui()
+        deployment_obj = DeploymentUI(setup_ui)
+        deployment_obj.mode = "internal"
+        deployment_obj.storage_class_type = "thin_sc"
+        deployment_obj.osd_size = "0.5T"
+        deployment_obj.is_encryption = False
+        deployment_obj.is_wide_encryption = False
+        deployment_obj.is_class_encryption = False
```
dahorak, line 637: There is already configuration parameter `encryption_at_rest` ([conf/README.md](https://github.com/red-hat-storage/ocs-ci/blob/master/conf/README.md)), isn't it related to some of those variables? (if yes, wouldn't be better to load the value from that configuration)

ocs_ci/deployment/deployment.py
```diff
@@ -623,19 +625,19 @@ def deploy_ocs_via_operator(self):
...
+        deployment_obj.is_use_kms = False
```
dahorak, line 638: I don't have deep knowledge what all those parameters means, but wouldn't make sense to make them configurable from configuration file?

ocs_ci/ocs/ui/base_ui.py
```diff
@@ -109,6 +132,23 @@ def take_screenshot(self):
         logger.info(f"Creating snapshot: {filename}")
         self.driver.save_screenshot(filename)
 
+    def get_text(self, locator, timeout=30, expected_text=None):
```
dahorak, line 135: The method name `get_text` is slightly misleading here - it sounds like it will return you some text from element, but it only compare it with some expected string.

Also it looks like the `expected_text` parameter should be mandatory (it doesn't make much sense to omit it when calling this method), but the `timeout` parameter is clearly optional. So it will probably be better to put the `expected_text` parameter before `timeout` parameter and do not set any default value for the `expected_text` parameter.

ocs_ci/ocs/ui/base_ui.py
```diff
@@ -138,10 +182,11 @@ def login_ui():
 
         # headless browsers are web browsers without a GUI
         headless = ocsci_config.UI_SELENIUM.get("headless")
-        if headless:
+        if not headless:
```
dahorak, line 185: Why this change?

ocs_ci/ocs/ui/base_ui.py
```diff
@@ -138,10 +182,11 @@ def login_ui():
...
             chrome_options.add_argument("--headless")
 
-        chrome_browser_type = ocsci_config.UI_SELENIUM.get("chrome_type")
+        # chrome_browser_type = ocsci_config.UI_SELENIUM.get("chrome_type")
+        chrome_browser_type = "google-chrome"
```
dahorak, line 189: isn't this some debugging leftovers?

ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -125,29 +201,118 @@ def install_storage_cluster(self):
...
 
         """
-        logger.info("Click On 'Internal'")
-        self.do_click(deployment["Internal"], type=By.XPATH)
+        logger.info("Click Internal")
+        self.do_click(locator=self.dep_loc["internal_mode"])
+
+        logger.info("Configure Storage Class (thin on vmware, gp2 on aws)")
+        self.do_click(locator=self.dep_loc["storage_class_dropdown"])
+        self.do_click(locator=self.dep_loc["thin_sc"])
```
dahorak, line 212: Shouldn't there be `self.storage_class_type` instead of `"thin_sc"`?

ocs_ci/deployment/deployment.py
petr-balogh: Agree with Daniel here, those params should not be set here out of thin air but should get value from config if possible and if they are not defined in config, then we can use some default value.

ocs_ci/ocs/ui/base_ui.py
OdedViner: I changed this for debugging the code.. I will change it..

ocs_ci/ocs/ui/base_ui.py
OdedViner: I changed this for debugging the code.. I will change it..

ebenahar (approved):

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -68,7 +68,7 @@ DEPLOYMENT:
   # For UI deployment you have to set following parameter ui_deployment to True
   # UI deployment is not part of this framework, it will just do preparation
   # steps for openshift-console.
-  ui_deployment: False
+  ui_deployment: True
```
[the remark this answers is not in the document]
OdedViner: I changed it only for testing,I will add new config file.

vasukulkarni (approved):

vasukulkarni (changes requested):
fixing default UI installation

ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -125,29 +204,118 @@ def install_storage_cluster(self):
...
+    def install_ocs_ui(self):
         """
-        Install LSO Cluster
+        Install OCS via UI
 
         """
-        pass
+        self.create_catalog_source_yaml()
+        self.install_ocs_operator()
+        self.verify_ocs_installation()
```
ebenahar, line 320: shouldn't verify_ocs_installation() be called after install_storage_cluster()?

OdedViner: On this method, I verify that ocs operator move `succeeded` state , (before create storage cluster) 
image

ebenahar: In that case, I think that the method name should be changed from verify_ocs_installation() to verify_ocs_operator_succeeded()

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -68,7 +68,7 @@ DEPLOYMENT:
   # For UI deployment you have to set following parameter ui_deployment to True
   # UI deployment is not part of this framework, it will just do preparation
```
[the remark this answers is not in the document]
OdedViner: @ebenahar I will change it on next pr.. 
After I will add the ui_deploy function to deployment.py file

ebenahar (approved):

vasukulkarni (approved):

Issue #3746, opened 2021-01-22
AttributeError: 'SanityExternalCluster' object has no attribute 'obc_objs'


clacroix12:
We are [expecting](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/helpers/sanity_helpers.py#L97-L98) `SanityExternalCluster` to have the attribute `obc_objs`, however this is [defined in the parent class](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/helpers/sanity_helpers.py#L31) `Sanity`. We do not define this in the [SanityExternalCluster init](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/helpers/sanity_helpers.py#L169-L176) or call the parent init so this attribute is never defined which results in the error below.

```
_______________________________ test_deployment ________________________________

pvc_factory = <function pvc_factory_fixture.<locals>.factory at 0x7f0ed23e3a60>
pod_factory = <function pod_factory_fixture.<locals>.factory at 0x7f0ed23e3d08>

    @deployment
    @polarion_id(get_polarion_id())
    def test_deployment(pvc_factory, pod_factory):
        deploy = config.RUN["cli_params"].get("deploy")
        teardown = config.RUN["cli_params"].get("teardown")
        if not teardown or deploy:
            log.info("Verifying OCP cluster is running")
            assert is_cluster_running(config.ENV_DATA["cluster_path"])
            if not config.ENV_DATA["skip_ocs_deployment"]:
                ocs_registry_image = config.DEPLOYMENT.get("ocs_registry_image")
                ocs_install_verification(ocs_registry_image=ocs_registry_image)
    
                # Check basic cluster functionality by creating resources
                # (pools, storageclasses, PVCs, pods - both CephFS and RBD),
                # run IO and delete the resources
                if config.DEPLOYMENT["external_mode"]:
                    sanity_helpers = SanityExternalCluster()
                else:
                    sanity_helpers = Sanity()
                sanity_helpers.health_check()
>               sanity_helpers.delete_resources()

tests/ecosystem/deployment/test_deployment.py:33: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ocs_ci.helpers.sanity_helpers.SanityExternalCluster object at 0x7f0ed2741438>

    def delete_resources(self):
[...]

PR #3753, opened 2021-01-25
Fix attribute error for SanityExternalCluster

jimdevops19:
Fixes: #3746

### DIFF
--- a/ocs_ci/helpers/sanity_helpers.py
+++ b/ocs_ci/helpers/sanity_helpers.py
@@ -171,6 +171,4 @@ class SanityExternalCluster(Sanity):
         Initializer for Sanity class - Init CephCluster() in order to
         set the cluster status before starting the tests
         """
-        self.pvc_objs = list()
-        self.pod_objs = list()
-        self.ceph_cluster = CephClusterExternal()
+        super(SanityExternalCluster, self).__init__()


ebenahar (approved):

shylesh (approved):
+1

ocs_ci/helpers/sanity_helpers.py
```diff
@@ -171,6 +171,4 @@ def __init__(self):
         Initializer for Sanity class - Init CephCluster() in order to
         set the cluster status before starting the tests
         """
-        self.pvc_objs = list()
-        self.pod_objs = list()
-        self.ceph_cluster = CephClusterExternal()
+        super(SanityExternalCluster, self).__init__()
```
clacroix12, line 174: The parent class initializes `self.ceph_cluster` differently (it is a `CephCluster` not a `CephClusterExternal`). We need to consider that here.

clacroix12 (changes requested):

ocs_ci/helpers/sanity_helpers.py
jimdevops19: Correct. Thanks for pointing this out, I am closing this issue temporarily.

Issue #3465, opened 2020-12-02
Configure monitoring and registry during OCS deployment in external mode


sidhant-agrawal:
Currently, monitoring and registry are not configured if OCS is deployed in external mode via ocs-ci. Hence other tests might fail which require monitoring/registry to be configured.

petr-balogh:
@clacroix12 if you have some time, maybe you can take a look at this one? I think it will be just about move some part of code to the function and call it from external deployment or maybe make the current logic better to not return after that and finish with common stuff in the deployment function.  Thanks

clacroix12:
@sidhant-agrawal @petr-balogh as far as I can tell the code which manages [monitoring](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/deployment.py#L733-L770) and [registry](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/deployment.py#L780-L781) is completely separate from where we [deploy with external mode](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/deployment.py#L683-L685).

Perhaps I'm not understanding the issue correctly or this isn't the right part of the code I need to be looking at, if so please clarify. Do we at least have an example deployment that I can take a look at which these tests fail on external mode?

sidhant-agrawal:
@clacroix12 yes, currently we are not calling those functions when deploying external mode but we can use same functions for external mode, provided we use correct storage class present in external mode.
Example failure in [test_must_gather](https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-trigger-test-pr/3491/testReport/tests.manage.z_cluster.test_must_gather/TestMustGather/test_must_gather_OTHERS_/) due to monitoring and registry related PVCs not present

clacroix12:
@sidhant-agrawal Sorry, what I meant was I'm not seeing anything that is stopping this code from executing during a deployment with external mode. Am I looking at the right part of the code or is there something I am missing?

The only condition for monitoring is this:

```python
        if config.ENV_DATA.get("monitoring_enabled") and config.ENV_DATA.get(
            "persistent-monitoring"
        ):
```

The call to change the registry backend isn't in an `if` at all, so I'm wondering what is stopping this from happening for external mode deployments?

petr-balogh:
[...]

PR #3769, opened 2021-01-27
Configure monitoring and registry during OCS deployment in external mode

clacroix12:
Fixes: #3465

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -822,8 +822,7 @@ class Deployment(object):
             )
 
         # Change registry backend to OCS CEPHFS RWX PVC
-        if not config.DEPLOYMENT["external_mode"]:
-            registry.change_registry_backend_to_ocs()
+        registry.change_registry_backend_to_ocs()
 
         # Verify health of ceph cluster
         logger.info("Done creating rook resources, waiting for HEALTH_OK")
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -822,7 +822,8 @@ class Deployment(object):
             )
 
         # Change registry backend to OCS CEPHFS RWX PVC
-        registry.change_registry_backend_to_ocs()
+        if not config.DEPLOYMENT["external_mode"]:
+            registry.change_registry_backend_to_ocs()
 
         # Verify health of ceph cluster
         logger.info("Done creating rook resources, waiting for HEALTH_OK")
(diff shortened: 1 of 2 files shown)


ocs_ci/deployment/deployment.py
```diff
@@ -855,10 +822,10 @@ def deploy_ocs(self):
             )
 
         # Change registry backend to OCS CEPHFS RWX PVC
-        registry.change_registry_backend_to_ocs()
+        if not config.DEPLOYMENT["external_mode"]:
+            registry.change_registry_backend_to_ocs()
```
sidhant-agrawal, line 826: why are we skipping registry change for external mode?

clacroix12: I added this `if` because of the failure in the first PR validation. Should we still be running this? If so it looks like we will need to change the behavior in this function.

sidhant-agrawal: Yes, we should be running this.
We need to use appropriate value of `sc_name` here [change_registry_backend_to_ocs()](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/registry.py#L28)
Similar to what we are doing in `setup_persistent_monitoring()`

clacroix12: Done in latest commit

ocs_ci/deployment/deployment.py
```diff
@@ -1045,3 +1011,42 @@ def create_catalog_source(image=None, ignore_upgrade=False):
...
+    # config file)
+    create_configmap_cluster_monitoring_pod(
+        sc_name=sc.name,
+        telemeter_server_url=config.ENV_DATA.get("telemeter_server_url"),
+    )
+
+    # Take some time to respin the pod
+    waiting_time = 45
+    logger.info(f"Waiting {waiting_time} seconds...")
+    time.sleep(waiting_time)
```
shylesh, line 1039: @clacroix12 , Instead of sleep can we wait for the resource? rest LGTM.

clacroix12: This logic hasn't changed from before, I just moved it into a new function to clean up a bit. Regardless, isn't that what `validate_pods_are_respinned_and_running_state` should be doing already?

shylesh: ok so if that is the case then do we need sleep here ?

shylesh: since it was already existing , I don't want to hold this PR. LGTM.

shylesh (approved):
+1

sidhant-agrawal (approved):
LGTM

clacroix12:
Raised an issue for the failure blocking PR validation here: #3789

clacroix12:
> Raised an issue for the failure blocking PR validation here: #3789

This is actually due to a known bug: https://bugzilla.redhat.com/show_bug.cgi?id=1921864

I am kicking off a deployment test to verify that non-external mode deployments are okay after these changes.

Issue #3578, opened 2020-12-21
Unit tests are failing on master


zmc:
See https://storage-jenkins-csb-ceph.cloud.paas.psi.redhat.com/job/ocs-ci/155/execution/node/40/log/

For some reason, unit tests are trying to access the network (login.microsoftonline.com specifically).

Are unit tests not gating PRs anymore? Why not?

zmc:
The offending test is here: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/tests/test_azure_utils.py#L49-L63

vasukulkarni:
@mbukatov can you look into this one?

mbukatov:
This test case actually sends (invalid) requests to azure api. It can be skipped when we don't want to communicate with outside world.

PR #3775, opened 2021-01-28
Remove test_azureutil_init_invalidcredentials

zmc:
Fixes: #3578

### DIFF
--- a/ocs_ci/utility/tests/test_azure_utils.py
+++ b/ocs_ci/utility/tests/test_azure_utils.py
@@ -4,9 +4,6 @@
 from unittest.mock import patch
 import copy
 
-import msrest.exceptions
-import pytest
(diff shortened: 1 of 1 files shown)


andrewschoen (approved):
Looks good to me. This test has been keeping us from using the master branch for some time now downstream. I'm a bit curious why we're requesting URLs like this within unit tests anyway.

vasukulkarni:
@mbukatov since its blocking the dev, we can move ahead with this, and you can later add skips?

vasukulkarni (approved):

zmc:
Thanks @vasukulkarni 

To be clear, unit tests should access any network.

mbukatov:
Right. We can move it to libtest.

PR #3783, opened 2021-01-29
downstream ci: use python3.8

andrewschoen:
I've tested this and python3.8 is installed and used: https://storage-jenkins-csb-ceph.cloud.paas.psi.redhat.com/job/ocs-ci/215/console

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -39,14 +39,16 @@ pipeline {
           fi
           sudo sysctl -w net.ipv6.conf.all.disable_ipv6=1
           sudo sysctl -w net.ipv6.conf.default.disable_ipv6=1
+          sudo yum install -y python38-devel
           python3 -V
           pip3 install --user virtualenv
-          python3 -m virtualenv venv
+          python3 -m virtualenv -p python3.8 venv
           source ./venv/bin/activate
+          python -V
           pip3 install tox
           pip3 install -r requirements.txt
-          python3 setup.py develop
-          python3 ./.functional_ci_setup.py --skip-aws
+          python setup.py develop
+          python ./.functional_ci_setup.py --skip-aws
           """
       }
     }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -39,14 +39,16 @@ pipeline {
           fi
           sudo sysctl -w net.ipv6.conf.all.disable_ipv6=1
           sudo sysctl -w net.ipv6.conf.default.disable_ipv6=1
+          sudo yum install -y python38-devel
```
petr-balogh, line 42: Cannot you add this to your image used for testing instead install it each pipeline run? I will approve it, but will be better if you have it as part of your image. WDYT?

petr-balogh (approved):

clacroix12 (approved):

Jenkinsfile
andrewschoen: We use some very generic cloud images and I don't know how easy it will be to modify those. I'd prefer to keep this install here for now.

ktdreyer: That would require us to maintain a tool that keeps those custom images up-to-date.

What do you recommend that we use for that?

clacroix12: That sounds a bit out of scope to me and more like something that could be taken up as it's own issue. If engineering is okay with adding a bit of time to their pipeline (I don't think python38-devel takes an insane amount of time to yum install) then I think we should go ahead and merge this to unblock QE on dropping python 3.6 support. Just my two cents :)

vavuthu (approved):

Jenkinsfile
petr-balogh: I approved this PR and just thought that engineering are using some custom image in PSI where they installed already some other dependencies. If that's not the case I don't mind to keep it this way.

PR #3793, opened 2021-02-02
Modify test_nodes_restart to run on all Cloud platforms

sidhant-agrawal:
Replaces aws_platform_required with cloud_platform_required in test_nodes_restart

### DIFF
--- a/tests/manage/z_cluster/nodes/test_nodes_restart.py
+++ b/tests/manage/z_cluster/nodes/test_nodes_restart.py
@@ -6,7 +6,7 @@ from ocs_ci.framework.testlib import (
     tier4a,
     ignore_leftovers,
     ManageTest,
-    aws_platform_required,
+    cloud_platform_required,
     bugzilla,
     skipif_no_lso,
 )
@@ -55,7 +55,7 @@ class TestNodesRestart(ManageTest):
             pytest.param(*[True], marks=pytest.mark.polarion_id("OCS-894")),
             pytest.param(
                 *[False],
-                marks=[pytest.mark.polarion_id("OCS-895"), aws_platform_required],
+                marks=[pytest.mark.polarion_id("OCS-895"), cloud_platform_required],
             ),
         ],
     )


ebenahar (approved):

tests/manage/z_cluster/nodes/test_nodes_restart.py
```diff
@@ -55,7 +55,7 @@ def finalizer():
             pytest.param(*[True], marks=pytest.mark.polarion_id("OCS-894")),
             pytest.param(
                 *[False],
-                marks=[pytest.mark.polarion_id("OCS-895"), aws_platform_required],
+                marks=[pytest.mark.polarion_id("OCS-895"), cloud_platform_required],
```
yitzhak12, line 58: The test `test_nodes_restart` uses the function `restart_nodes_by_stop_and_start`, which is implemented for AWS, AZURE platforms but not for GCP. If the test runs on GCP, it will raise **NotImplementedError** as you can see here https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/platform_nodes.py#L110. So we need to implement this method for GCP or consider that the test will not run on GCP right now.

wusui (approved):

tests/manage/z_cluster/nodes/test_nodes_restart.py
sidhant-agrawal: ack, I'll add skip for GCP

sidhant-agrawal: @yitzhak12 Other tests like "OCS-894" will still fail on GCP, adding skip only for `OCS-895` doesn't look useful to me.
I think we can consider that the test will not run on GCP right now.

yitzhak12: Yes, no problem. In the future, I think we need to add another class for GCP methods in the https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/platform_nodes.py file.

yitzhak12 (approved):

PR #3795, opened 2021-02-02
Metro DR cluster yaml

vavuthu:

### DIFF
--- a/conf/deployment/aws/ipi_3az_rhcos_3m_4w.yaml
+++ b/conf/deployment/aws/ipi_3az_rhcos_3m_4w.yaml
@@ -1,0 +1,18 @@
+---
+#  This configuration file is used for Metro DR cluster
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
+ENV_DATA:
+  platform: 'aws'
+  deployment_type: 'ipi'
+  region: 'us-east-2'
+  worker_availability_zones:
+    - 'us-east-2a'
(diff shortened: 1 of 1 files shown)


conf/deployment/aws/ipi_3az_rhcos_3m_4w.yaml
```diff
@@ -0,0 +1,18 @@
...
+  worker_availability_zones:
+    - 'us-east-2a'
+    - 'us-east-2b'
+  master_availability_zones:
+    - 'us-east-2a'
+    - 'us-east-2b'
+    - 'us-east-2c'
+  worker_replicas: 4
+  master_replicas: 3
+  worker_instance_type: 'm5.4xlarge'
```
prsurve, line 18: are we testing Metro DR on AWS ?

petr-balogh: @prsurve we are going to discuss this on mtg in 50 mins. The thing is that create such deployment on vSphere is not that easy and work on such deployment will require resources which we do not have much. 

Such deployment implementation on vSphere is not possible in range of 1 month I guess and if this will work and allow us to test such scenarios easily for disconnecting 1 AZ from the other I would go with this approach which we can do in timeframe of 4.7 GA.

But let's discuss it on the mtg.

vavuthu:
closing this as this PR is not relevant anymore

Issue #3765, opened 2021-01-27
Change Node replacement Ceph verification steps when we use more than 3 OCS worker nodes


yitzhak12:
When we run this test https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/16556/ the test of node replacement has failed in the verification steps at this line https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/node.py#L1156 - the new OCS worker node not in the "Ceph osd status". The reason for the failure is: When we have more than 3 OCS worker nodes and deleting one worker node, one of the existing OCS nodes will replace it, and therefore the new node will be associated with one of the OSD's and the new node will not be associated with the OSD's and will not find in "Ceph osd status". 

To fix it, we need to change the Node replacement Ceph verification steps to check if we have more than 3 OCS worker nodes, and then we need to check if one of the existing OCS worker nodes are associated with the OSD's instead of the new one.

PR #3804, opened 2021-02-03
Update Node replacement verification steps when we use more than 3 OCS worker nodes

yitzhak12:
When we use more than 3 OCS worker nodes, if we replace an existing OCS worker node, one of the **existing** OCS worker nodes will replace the old one. So in the verification steps, we should consider the option that the new OCS worker node **will not** be associated with one of the OSD's. Instead, one of the existing OCS worker nodes will be associated with the OSD's.

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -1087,7 +1087,7 @@ def get_worker_nodes_not_in_ocs():
 
 
 def node_replacement_verification_steps_user_side(
-    old_node_name, new_node_name, old_osd_id
+    old_node_name, new_node_name, new_osd_node_name, old_osd_id
 ):
     """
     Check the verification steps that the user should perform after the process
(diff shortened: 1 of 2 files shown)


yitzhak12:
fix https://github.com/red-hat-storage/ocs-ci/issues/3765

ebenahar (approved):

prsurve (approved):

sidhant-agrawal (approved):

yitzhak12:
The PR above has succeeded. There was an error in the copy of the results as you can see here https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-trigger-test-pr/34/console: 

> Error when copy artifacts or creating report: ERROR: hudson.AbortException: Failed to copy artifacts from qe-deploy-ocs-cluster with filter: test_results/*.xml
[Pipeline] echo
Error detail: Failed to copy artifacts from qe-deploy-ocs-cluster with filter: test_results/*.xml
[Pipeline] echo
currentBuild.result: FAILURE

That's why it marked the PR as failed, but it actually passed as you can see here https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/173/consoleFull

PR #3806, opened 2021-02-03
Remove acceptance marker from namespace bucket test cases

ebenahar:

### DIFF
--- a/tests/manage/mcg/test_namespace.py
+++ b/tests/manage/mcg/test_namespace.py
@@ -7,7 +7,6 @@ from ocs_ci.framework.testlib import (
     tier3,
     tier4,
     tier4a,
-    acceptance,
 )
 from ocs_ci.ocs.bucket_utils import sync_object_directory, verify_s3_object_integrity
 from ocs_ci.framework.pytest_customization.marks import skipif_aws_creds_are_missing
@@ -43,7 +42,6 @@ class TestNamespace(MCGTest):
         ns_resource_factory()
 
     @tier1
-    @acceptance
     @pytest.mark.parametrize(
         argnames=["platform"],
         argvalues=[


ebenahar:
verified - https://storage-jenkins-csb-ceph.cloud.paas.psi.redhat.com/job/ocs-ci/222/

MeridianExplorer:
I don't think that removing tests from acceptance is the way to go.
Testing the multicloud functionalities of the Multicloud Gateway is essential - and I think the right solution is to make sure all Jenkins instances are configured correctly and have access to an appropriate _auth.yaml_ file. I don't think it'd be right to skip tests just because they require this file.

MeridianExplorer (dismissed):
I don't think that removing tests from acceptance is the way to go.
Testing the multicloud functionalities of the Multicloud Gateway is essential - and I think the right solution is to make sure all Jenkins instances are configured correctly and have access to an appropriate auth.yaml file. I don't think it'd be right to skip tests just because they require this file.
If anything, we should be adding many more tests to acceptance runs, which will lead to many more occurrences of failed tests because of _auth.yaml_ issues.

vasukulkarni:
No reason for removal, If this was passing before we shouldn't remove it.

b-ranto:
The purpose of the acceptance tests is to deliver a build that is worth testing. We only tag a build as latest-stable if it passes acceptance tests.

We have been asked numerous times to override this and manually tag a build as latest-stable because the build pretty much worked (was worth testing) but failed in one of the acceptance tests. This is really not maintainable in the long run.

We can auto-schedule an ocs-ci run with another set of tests after it passes the acceptance tests but the purpose of the acceptance tests is to give you a build that is worth testing, not perfect, not a build that is "without a regression".

An alternative approach would be to create a smaller set of tests for the latest-stable tagging feature.

ebenahar:
> No reason for removal, If this was passing before we shouldn't remove it.

These tests were added recently and never passed acceptance before. 
The problem is with misconfigured auth.yaml file with credentials to cloud providers. 
Not sure how this should be fixed but these failures should not be a reason for not accepting a build

vasukulkarni (approved):

vasukulkarni:
@ebenahar makes sense not to block builds, we can fix the test and add it back when its working

MeridianExplorer (approved):

PR #3819, opened 2021-02-04
Skip node_replacement_proactive test for IBM as the logic doesn't handle the IBM specific instance

vasukulkarni:
delete_and_create_osd nodes only handle cloud and vsphere platforms, it does nothing for IBM and in the end verification fails, we should skip this test.  Seen during one of the dev failures https://storage-jenkins-csb-ceph.cloud.paas.psi.redhat.com/job/ocs-ci-kvm/152/testReport/


petr-balogh (approved):

zmc (approved):
Thanks!

PR #3859, opened 2021-02-15
Update node replacement process to be compatible with OCP 4.7

yitzhak12:


ocs_ci/ocs/resources/pod.py
```diff
@@ -1779,14 +1785,18 @@ def delete_osd_removal_job(osd_id):
         bool: True, if the ocs-osd-removal job deleted successfully. False, otherwise
 
     """
-    osd_removal_job = get_job_obj(
-        f"ocs-osd-removal-{osd_id}", namespace=defaults.ROOK_CLUSTER_NAMESPACE
-    )
+    ocp_version = float(get_ocp_version())
+    if ocp_version >= 4.7:
+        job_name = "ocs-osd-removal-job"
+    else:
```
prsurve, line 1791: We need a verification on ocs 4.6 build as will to make sure everything is working fine

ocs_ci/ocs/resources/pod.py
```diff
@@ -1660,12 +1660,18 @@ def get_osd_removal_pod_name(osd_id, timeout=60):
         str: The osd removal pod name
 
     """
+    ocp_version = float(get_ocp_version())
+    if ocp_version >= 4.7:
+        pattern = "ocs-osd-removal-job"
+    else:
+        pattern = f"ocs-osd-removal-{osd_id}"
```
prsurve, line 1667: We need a verification on ocs 4.6 build as will to make sure everything is working fine

ocs_ci/ocs/resources/pod.py
yitzhak12: Okay, I will run a PR validation also on OCS 4.6

ocs_ci/ocs/resources/pod.py
```diff
@@ -1779,14 +1786,19 @@ def delete_osd_removal_job(osd_id):
...
+        job_name = f"ocs-osd-removal-{osd_id}"
+
+    osd_removal_job = get_job_obj(job_name, namespace=defaults.ROOK_CLUSTER_NAMESPACE)
     osd_removal_job.delete()
     try:
-        osd_removal_job.ocp.wait_for_delete(resource_name=f"ocs-osd-removal-{osd_id}")
+        osd_removal_job.ocp.wait_for_delete(resource_name=job_name)
     except TimeoutError:
-        logger.warning(f"ocs-osd-removal-{osd_id} job did not deleted successfully")
+        logger.warning(f"{job_name} job did not deleted successfully")
```
rachael-george, line 1801:
```suggestion
        logger.warning(f"{job_name} job did not get deleted successfully")
```

ocs_ci/ocs/node.py
```diff
@@ -1464,3 +1456,25 @@ def wait_for_new_osd_node(old_osd_node_names, timeout=180):
...
+    """
+    Add a new disk to a node
+
+    Args:
+        node_obj (ocs_ci.ocs.resources.ocs.OCS): The node object
+        disk_size (int): The size of the new disk to attach. If not specified,
+            the disk size will be equal to the size of the previous disk.
+
+    """
+    from ocs_ci.ocs.platform_nodes import PlatformNodesFactory
```
clacroix12, line 1471: Why was this import moved to inside this function?

yitzhak12: I put it here and not at the top of the file because it causes a circular loop error.
You can also see the same import at this line https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/node.py#L357.

ebenahar (dismissed):

tests/manage/z_cluster/nodes/test_node_replacement_proactive.py
```diff
@@ -104,14 +104,25 @@ def delete_and_create_osd_node(osd_node_name):
...
     elif config.ENV_DATA["platform"].lower() == constants.VSPHERE_PLATFORM:
-        if is_lso_cluster():
-            new_node_name = node.delete_and_create_osd_node_vsphere_upi_lso(
-                osd_node_name, use_existing_node=False
-            )
+        wnodes_not_in_ocs = node.get_worker_nodes_not_in_ocs()
+        # We use for vSphere an existing worker node, due to issue
+        # https://github.com/red-hat-storage/ocs-ci/issues/3894
+        # Issue for tracking is https://github.com/red-hat-storage/ocs-ci/issues/3945
+        if wnodes_not_in_ocs:
```
ebenahar, line 111: is it always the case that for vsphere, an extra worker non-OCS node has to exist?

ebenahar (approved):

tests/manage/z_cluster/nodes/test_node_replacement_proactive.py
yitzhak12: It was in the past. Then the issue was fixed by Vijay, and we used a new node. We now had this issue again due to a change in the adding node function - We now use terraform for adding nodes.

ebenahar: is there any way to first try adding a new node and if it is not possible, look for non OCS worker nodes?

yitzhak12: Yes. But also in this case, it will not be the same process of the node replacement we want to simulate.

ocs_ci/ocs/node.py
```diff
@@ -1027,11 +1019,13 @@ def delete_and_create_osd_node_vsphere_upi_lso(osd_node_name, use_existing_node=
     assert is_completed, "ocs-osd-removal-job is not in status 'completed'"
     log.info("ocs-osd-removal-job completed successfully")
 
-    expected_num_of_deleted_pvs = 1
+    expected_num_of_deleted_pvs = [0, 1]
```
ebenahar, line 1022: shouldn't this be a dynamic variable that is being passed to the function?

yitzhak12: Do you mean that we should check inside the function how many PVs have been deleted and return `True` in case of success and `False` otherwise?

ebenahar: either that or maybe the function can get the expected number of deleted PVs

ebenahar (approved):

ocs_ci/ocs/cluster.py
```diff
@@ -1531,6 +1536,25 @@ def is_lso_cluster():
...
 
+def is_flexible_scaling_enabled():
+    """
+    Check if flexible scaling is enabled
+
+    Returns:
+        bool: True if failure domain is "host" and flexible scaling is enabled. False otherwise
+
+    """
+    ocs_storage_cluster = ocp.OCP(
```
OdedViner, line 1547: you can use this function
https://github.com/red-hat-storage/ocs-ci/blob/2c67b247688ac0fac7474cafa7490c98f22e6875/ocs_ci/ocs/resources/storage_cluster.py#L510

ocs_ci/ocs/resources/pod.py
```diff
@@ -1791,14 +1806,19 @@ def delete_osd_removal_job(osd_id):
         bool: True, if the ocs-osd-removal job deleted successfully. False, otherwise
 
     """
-    osd_removal_job = get_job_obj(
-        f"ocs-osd-removal-{osd_id}", namespace=defaults.ROOK_CLUSTER_NAMESPACE
-    )
+    ocp_version = float(get_ocp_version())
+    ocs_version = float(config.ENV_DATA["ocs_version"])
+    if ocp_version >= 4.6 and ocs_version >= 4.7:
```
OdedViner, line 1811: We should be using semantic_version according to Petr's email a couple weeks back. This will basically prevent comparisons like this from breaking once ocs_version reaches something like 4.10 or 4.11. This code expects an ocs_version of 4.10 to be greater than 4.7 but the float 4.10 is less than 4.7. By using semantic_version, we can avoid this.

@clacroix12 ,Cody comment:)

ocs_ci/ocs/node.py
yitzhak12: I think it will limit the function... cause the purpose of the function is to delete all the released PVs in a specific storage class.

ocs_ci/ocs/cluster.py
yitzhak12: Fixed

ocs_ci/ocs/resources/pod.py
yitzhak12: Fixed

ocs_ci/ocs/resources/pv.py
```diff
@@ -133,3 +133,18 @@ def delete_released_pvs_in_sc(sc_name):
             logger.info(f"Failed to delete pv {pv_name} after {timeout} seconds")
 
     return num_of_deleted_pvs
+
+
+def get_pv_size(pv_obj):
```
ebenahar, line 138: consider using https://github.com/red-hat-storage/ocs-ci/blob/6fe0a01ded95676c39005ac8261629a0a59e6039/ocs_ci/utility/utils.py#L2567

yitzhak12: Fixed

ebenahar (approved):

ocs_ci/ocs/cluster.py
```diff
@@ -1450,9 +1450,14 @@ def check_ceph_osd_tree_after_node_replacement():
...
     osd_tree = ct_pod.exec_ceph_cmd(ceph_cmd="ceph osd tree")
-    if not check_ceph_osd_tree():
-        logger.warning("Incorrect ceph osd tree formation found")
-        return False
+    # We need to add a function to check the ceph osd tree when flexible scaling is enabled.
+    # So currently, when flexible scaling is enabled don't check the Ceph osd tree.
+    # The issue for tracking is https://github.com/red-hat-storage/ocs-ci/issues/4052
+    if not is_flexible_scaling_enabled():
+        if not check_ceph_osd_tree():
+            logger.warning("Incorrect ceph osd tree formation found")
```
OdedViner, line 1458: you can use "and"
if (not is_flexible_scaling_enabled()) and (not check_ceph_osd_tree()):

yitzhak12: Yes. But in this specific case, I preferred to separate the lines to highlight that only when flexible scaling is not enabled do we need to check the ceph osd tree. When the issue https://github.com/red-hat-storage/ocs-ci/issues/4052 will be fixed, I will remove the line `if not is_flexible_scaling_enabled()`.

yitzhak12:
After looking at the last PR validation job console output https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/2111/consoleFull - I saw that that the error isn't related to the test. The error seems to be related to connecting to the cluster(which still I am not sure why we can't connect to the cluster (it may relate to the time the cluster exists). But as you can see in the console output, Ceph cluster health was Health OK at the end of the test, and the resources reached the desired condition: 

> 17:30:43 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage get Pod rook-ceph-mds-ocs-storagecluster-cephfilesystem-b-778594b9j5957 -n openshift-storage
17:30:43 - MainThread - ocs_ci.ocs.ocp - INFO - 2 resources already reached condition!
17:30:43 - MainThread - ocs_ci.ocs.cluster - INFO - Cluster HEALTH_OK

ebenahar (approved):

OdedViner (approved):

Issue #3863, opened 2021-02-15
Workaround: tests who use `rgw_deployments` for skipping force-skip on IBM Cloud


MeridianExplorer:
Because of [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1926831), RGW deployments can always be found in IBM Cloud clusters.
However, the deployments should only exist in case COS is not available.
Until the bug is fixed, `rgw_deployments` was changed to always skip tests over IBM Cloud, even if RGW deployments were found.

PR #3864, opened 2021-02-15
skip namespace tests with rgw in ibm cloud deployments

bernerhat:
workaround as described in issue #3863


tests/conftest.py
```diff
@@ -1709,6 +1709,10 @@ def rgw_deployments(request):
         label=constants.RGW_APP_LABEL, namespace=config.ENV_DATA["cluster_namespace"]
     )
     if rgw_deployments:
+        # Force-skipping in case of IBM Cloud -
+        # https://github.com/red-hat-storage/ocs-ci/issues/3863
+        if config.ENV_DATA["platform"].lower() == "ibm_cloud":
```
petr-balogh, line 1714: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/constants.py#L621 we have ibm_cloud platform in constants. Please use this.

MeridianExplorer (approved):

petr-balogh:
There was reply in mail from the execution:
Also, we will make COS an option on IBM Cloud - if the secret is not created on COS, it will default to rgw object storage (just like satellite)

ebenahar:
Adding do not merge label because it's still unclear whether or not namespace bucket is available on IBM COS

nimrod-becker:
Your mixing between what resource the NS bucket points to and where the system is running.

NS bucket behaves and is supported exactly the same regardless of the platform its running on.
Regardless of where the system runs, NS bucket supports using certain resources such as S3, Azure etc.

In general, yes COS is supported as a *resource* since it supports an S3 compatible API

Also , @petr-balogh AFAIK and according to the PR opened by and for ROKS (as well as a BZ requesting to backport to 4.6.z) if no secret is available for the default BS creation, we would default back to pv pools, just like any other platform.

Not sure if this helped or not, but more context would help me help more :)

MeridianExplorer:
@ebenahar - On top of what Nimrod said - NS buckets are available. This PR only skips the RGW permutation of NS tests. 
No RGW means no RGW-backed NS buckets.
Once we'll have a way to distinguish whether RGW is actually available on the current IBM cluster, we'll be able to decide whether to run these tests or not.
Removing the label

ebenahar (approved):

petr-balogh (approved):

PR #3866, opened 2021-02-15
Remove skip mark from `test_scaling_under_load`

### LINKED ISSUE #3720: Remove skip from `test_scaling_under_load`
We observed in recent runs over OCS 4.7 that all MCG tests running after `test_scaling_under_load` have failed.
From early observations, we suspect that the test causes a system malfunction in the product.
Thus, we'd like to test this theory by skipping the test for the time being.

bernerhat:
Closes #3720


MeridianExplorer (approved):

ebenahar (approved):

MeridianExplorer (approved):

ebenahar:
@bernerhat verified?

bernerhat:
@ebenahar its not a new test case or any change to the existing one, the issue was not relevant to the test in the end and was fixed.
i dont think we need to verify, we can merge

ebenahar (approved):

ebenahar:
@bernerhat still better to run the test to see if it passes now as maybe there were changes that were done in the code over the past few weeks, when the test was disabled, which affected it but we didn't know it was affected as it was skipped

PR #3868, opened 2021-02-15
Fix time measurement error where negative time durations occur

### LINKED ISSUE #3844: measure_pvc_deletion_time(interface, pv_name) measures negative time in case the deletion of pvc started on one day and ended on the following day
1) measure_pvc_deletion_time(interface, pv_name) measures negative time in case the deletion of pvc started on one day and ended on the following day . 

This needs to be fixed, otherwise we'll might have the following measure in performance tests : 

01:59:59 - MainThread - ocs_ci.ocs.ocp - INFO - PersistentVolumeClaim pvc-test-a8acc7233ee246038e738a6d6919c19 got deleted successfully
02:00:29 - MainThread - root - INFO - Inteface type: CephFileSystem: PVC number 10 deleted in -86398.566805 seconds

2) Probably the same will happen with measuring pvc creation time when the operation of creation starts on one day and ends on the following day.

wusui:
when tests run on midnight.

Fixes #3844 on github

### DIFF
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -46,6 +46,7 @@ from ocs_ci.utility.utils import (
 )
 
 logger = logging.getLogger(__name__)
+DATE_TIME_FORMAT = "%Y I%m%d %H:%M:%S.%f"
 
 
 def create_unique_resource_name(resource_description, resource_type):
@@ -1194,7 +1195,6 @@ def get_snapshot_time(interface, snap_name, status):
         return None
 
     logs = ""
-    format = "%Y I%m%d %H:%M:%S.%f"
 
     # the starting and ending time are taken from different logs,
     # the start creation time is taken from the snapshot controller, while
@@ -1233,7 +1233,7 @@ def get_snapshot_time(interface, snap_name, status):
         elif status.lower() == "start":
             stat = all_stats[0]  # return the lowest time
     if stat:
-        return datetime.datetime.strptime(stat, format)
+        return datetime.datetime.strptime(stat, DATE_TIME_FORMAT)
     else:
         return None
 
@@ -1277,7 +1277,6 @@ def get_provision_time(interface, pvc_name, status="start"):
     if status.lower() == "end":
         operation = "succeeded"
 
-    format = "%Y I%m%d %H:%M:%S.%f"
     this_year = str(datetime.datetime.now().year)
     # Get the correct provisioner pod based on the interface
     pod_name = pod.get_csi_provisioner_pod(interface)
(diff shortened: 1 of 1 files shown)


vasukulkarni (approved):

keesturam (commented):
Just a minor change suggested from my side.

ocs_ci/helpers/helpers.py
```diff
@@ -1146,6 +1146,20 @@ def set_image_lookup(image_name):
     return status
 
 
+#  The time and date extraction code below has been modified to read
```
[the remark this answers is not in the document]
clacroix12: The description already includes the issue which should be sufficient. However this is quite a large comment. I think this should be consolidated a bit and placed in the docstring of the relevant function.

ocs_ci/helpers/helpers.py
```diff
@@ -1146,6 +1146,20 @@ def set_image_lookup(image_name):
...
+#  time values are calculated when test runs cross midnight.  Also, previous
+#  calculations would not set the year, and so the calculations were done
+#  as if the year were 1900.  This is not a problem except that 1900 was
+#  not a leap year and so the next February 29th would throw ValueErrors
+#  for the whole day.  To avoid this problem, changes were made to also
+#  include the current year.
+#
+#  Incorrect times will still be given for tests that cross over from
+#  December 31 to January 1.  Also, there is a lot of code duplication
+#  here that should probably be cleaned up at some point.
```
clacroix12, line 1160: Please create an issue describing the duplication you're talking about here so we can track that.

ocs_ci/helpers/helpers.py
```diff
@@ -1174,13 +1188,14 @@ def get_pattern_time(log, snapname, pattern):
             str: string of the pattern timestamp in the log, if not found None
 
         """
+        this_year = str(datetime.datetime.now().year) + " "
         for line in log:
             if re.search(snapname, line) and re.search(pattern, line):
-                return line.split(" ")[1]
+                return this_year + " ".join(line.split(" ")[0:2])
```
clacroix12, line 1194: I think we should use f-strings instead of string concat. Maybe we can use a variable for `" ".join(line.split(" ")[0:2]` as well so it's more clear via the code what this represents. As it stands I don't really know what this is supposed to be. I would also address other similar lines in the same way.

clacroix12 (approved):
There is a good amount of duplication across a few of these functions that could be eliminated. `get_start_creation_time ` and `get_end_creation_time ` could call a generic function and pass 'start' or 'end' to it. Same thing goes with `get_start_deletion_time` and `get_end_deletion_time`. Furthermore if that generic function took 'provision' or 'delete' as an argument you could call the same function from all four of those to cut down on a lot of code duplication. This is not something that needs to happen in this PR as it has already passed testing and has approvals, but I would suggest creating an issue to address this in a subsequent PR.

wusui:
#3966 has been opened to address code duplication in helper/helpers.py

PR #3869, opened 2021-02-17
Implement node stop/start/restart for Power

gitsridhar:
This is an implementation of node stop/start/restart for
IBM Power environment other than libvirt. In libvirt
environment, the same is implemented using virsh stop/start etc.
For IBM Power environment other than Libvirt, the same is
attempted with stop/start of kubelet service inside the nodes.

Signed-off-by: gitsridhar <svenkat@us.ibm.com>

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -278,6 +278,30 @@ def get_node_ips(node_type="worker"):
         raise NotImplementedError
 
 
+def get_node_ip_addresses(ipkind):
+    """
+    Gets a dictionary of required IP addresses for all nodes
+
+    Args:
+        ipkind: ExternalIP or InternalIP or Hostname
+
+    Returns:
+        dict: Internal or Exteranl IP addresses keyed off of node name
+
+    """
+    ocp = OCP(kind=constants.NODE)
+    masternodes = ocp.get(selector=constants.MASTER_LABEL).get("items")
+    workernodes = ocp.get(selector=constants.WORKER_LABEL).get("items")
+    nodes = masternodes + workernodes
+
+    return {
+        node["metadata"]["name"]: each["address"]
+        for node in nodes
+        for each in node["status"]["addresses"]
+        if each["type"] == ipkind
+    }
+
+
 def add_new_node_and_label_it(machineset_name, num_nodes=1, mark_for_ocs_label=True):
     """
     Add a new node for ipi and label it
--- a/ocs_ci/ocs/platform_nodes.py
+++ b/ocs_ci/ocs/platform_nodes.py
@@ -1720,7 +1720,7 @@ class IBMPowerNodes(NodesBase):
 
     def __init__(self):
         super(IBMPowerNodes, self).__init__()
-        self.powernodes = powernodes.POWERNodes()
+        self.powernodes = powernodes.PowerNodes()
 
     def stop_nodes(self, nodes, force=True):
         """
@@ -1736,8 +1736,8 @@ class IBMPowerNodes(NodesBase):
                 nodes, timeout=900, wait=True, force=force
             )
         else:
-            raise NotImplementedError(
-                "This is not libvirt environment. Stop nodes not implemented"
+            self.powernodes.stop_powernodes_machines_powervs(
+                nodes, timeout=900, wait=True
             )
 
     def start_nodes(self, nodes, force=True):
@@ -1754,8 +1754,8 @@ class IBMPowerNodes(NodesBase):
                 nodes, timeout=900, wait=True, force=force
             )
         else:
-            raise NotImplementedError(
-                "This is not libvirt environment. Start nodes not implemented"
+            self.powernodes.start_powernodes_machines_powervs(
+                nodes, timeout=900, wait=True
             )
 
     def restart_nodes(self, nodes, timeout=540, wait=True, force=True):
@@ -1775,8 +1775,8 @@ class IBMPowerNodes(NodesBase):
                 nodes, timeout=900, wait=True, force=force
             )
         else:
-            raise NotImplementedError(
-                "This is not libvirt environment. Restart nodes not implemented"
+            self.powernodes.restart_powernodes_machines_powervs(
+                nodes, timeout=900, wait=True
             )
 
     def restart_nodes_by_stop_and_start(self, nodes, force=True):
@@ -1793,31 +1793,37 @@ class IBMPowerNodes(NodesBase):
                 nodes, timeout=900, wait=True, force=force
             )
         else:
-            raise NotImplementedError(
-                "This is not libvirt environment. Restart nodes by stop and start not implemented"
+            self.powernodes.restart_powernodes_machines_powervs(
+                nodes, timeout=900, wait=True
             )
 
     def restart_nodes_by_stop_and_start_teardown(self):
         """
         Make sure all PowerNodes are up by the end of the test
         """
-        if not self.powernodes.iskvm():
-            raise NotImplementedError(
-                "This is not libvirt environment. Restart nodes by stop and start teardown not implemented"
-            )
-
         self.cluster_nodes = get_node_objs()
-        stopped_powernodes = [
-            powernode
-            for powernode in self.cluster_nodes
-            if self.powernodes.verify_machine_is_down(powernode) is True
-        ]
+        if self.powernodes.iskvm():
+            stopped_powernodes = [
+                powernode
+                for powernode in self.cluster_nodes
+                if self.powernodes.verify_machine_is_down(powernode) is True
+            ]
+        else:
+            stopped_powernodes = [
+                powernode
+                for powernode in self.cluster_nodes
+                if powernode.ocp.get_resource_status(powernode.name)
+                == constants.NODE_NOT_READY
+            ]
 
         if stopped_powernodes:
             logger.info(
                 f"The following PowerNodes are powered off: {stopped_powernodes}"
             )
-            self.powernodes.start_powernodes_machines(stopped_powernodes)
+            if self.powernodes.iskvm():
+                self.powernodes.start_powernodes_machines(stopped_powernodes)
+            else:
+                self.powernodes.start_powernodes_machines_powervs(stopped_powernodes)
 
 
 class AZURENodes(NodesBase):
--- a/ocs_ci/utility/powernodes.py
+++ b/ocs_ci/utility/powernodes.py
@@ -1,4 +1,5 @@
 import logging
+import time
 
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.exceptions import UnexpectedBehaviour
@@ -6,11 +7,12 @@ from ocs_ci.ocs.node import wait_for_nodes_status, get_worker_nodes, get_master_
 from ocs_ci.ocs.ocp import wait_for_cluster_connectivity
 from ocs_ci.utility.utils import TimeoutSampler, exec_cmd
 from ocs_ci.utility.utils import get_ocp_version
+from ocs_ci.utility.service import KubeletService
 
 logger = logging.getLogger(__name__)
 
 
-class POWERNodes(object):
+class PowerNodes(object):
     """
     Wrapper for PowerNodes
     """
@@ -19,6 +21,7 @@ class POWERNodes(object):
         """
         Class Initialization.
         Run lsmod command to prepare isKVM variable.
+        Also, get a reference for KubeletService object.
         """
         cmd = "sudo /usr/sbin/lsmod"
         result = exec_cmd(cmd)
@@ -27,6 +30,8 @@ class POWERNodes(object):
         else:
             self.isKVM = False
         logger.info(f"iskvm check: {self.isKVM}")
+
+        self.service = KubeletService()
 
     def iskvm(self):
         """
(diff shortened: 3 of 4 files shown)


ocs_ci/utility/powernodes.py
```diff
@@ -56,6 +57,37 @@ def verify_machine_is_down(self, node):
         else:
             return True
 
+    def stop_powernodes_machines_powervs(
```
vasukulkarni, line 60: Instead of power class, it can be its own class `kube_service.py` , I think in general its applicable for all platforms and should be different category than node reboots.

gitsridhar: Vasu, thanks for your comments. But I am aware of only code plugged into platform_nodes. If I implement this into a separate class, how do I link it to platform_nodes? Test cases get an instance of platform node and invoke node restarts from platform_nodes interface.

vasukulkarni: I will check the test case  to see how we can keep the distinction clear.

shylesh: @gitsridhar ,
If I understand your requirements properly, you need control over node and service. Why can't we have a composition here. 
```python

class Service(object):
   self.service_type=None

class Kubelet(Service):
  def __init__(self):
    super().__init__("kubelet")
 
  def start():
  def stop():
  def kill(graceful=True):     

class IBMPowerNodes(NodeBase):
   
  def __init_(self): 
    service = None

  def init_service_obj(self, service_name):
     service = service_class()
```

From the test we can do:
```python
node = IBMPowerNodes()
node.init_service_obj('kubelet')
node.service.start()
node.service.stop()
node.service.kill()
```

We can rethink the class in case if you need multiple service types to be controlled.

Not sure if this helps in your case .
Thoughts ?

shylesh: @gitsridhar , sorry I didn't realize that you are already doing something similar.

petr-balogh: @gitsridhar I agree with Vasu to move this to some other file as stop/restart/start of service can be used not just on power nodes.

So you can move just parts from here about start/stop/restart to some service.py file and describe that this action happens via ssh.  The rest of the logic can stay in this file as you have.

ocs_ci/utility/powernodes.py
```diff
@@ -56,6 +57,37 @@ def verify_machine_is_down(self, node):
...
+        self, powernode_machines, timeout=900, wait=True, force=True
+    ):
+        """
+        Stop PowerNode Machines - PowerVS
+
+        Args:
+            powernode_machines (list): PowerNodes objects
+            timeout (int): time in seconds to wait for node to reach 'not ready' state
+            wait (bool): True if need to wait till the restarted node reaches timeout
+                         - for future use
```
petr-balogh, line 70: too big indention here, only 4 spaces please

ocs_ci/utility/powernodes.py
```diff
@@ -90,6 +122,34 @@ def stop_powernodes_machines(
...
+            powernode_machines (list): List of PowerNode machines
+            timeout (int): time in seconds to wait for node to reach 'not ready' state,
+                and 'ready' state.
+            wait (bool): Wait for PowerNodes to start - for future use
+            force (bool): True for PowerNode ungraceful power off, False for
+                graceful PowerNode shutdown - for future use
+        """
+        for node in powernode_machines:
+            result = exec_cmd(
+                f"ssh core@{node.name} sudo systemctl start kubelet.service"
```
petr-balogh, line 141: I think that node.name is not reliable value for IP address as this is not always IP address but can be also the regular name.   Node IP can be taken from status:
```
status:
  addresses:
  - address: 10.243.128.6
    type: InternalIP
  - address: 10.243.128.6
    type: ExternalIP
  - address: 10.243.128.6
    type: Hostname
```
This is other part which should be moved to service.py and you will just import it and call it here.

So I would go with ExternalIP type.

We have this function for that purpose:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/node.py#L251

So I think you should use this and get one of external IP to be used here. So maybe you can write new function in node.py which will be called get_node_ip which will return only one reachable IP.

ocs_ci/utility/powernodes.py
```diff
@@ -56,6 +57,37 @@ def verify_machine_is_down(self, node):
...
+                         - for future use
+            force (bool): True for PowerNode ungraceful power off, False for
+                graceful PowerNode shutdown - for future use
+
+        Raises:
+            UnexpectedBehaviour: If PowerNode machine is still up
+        """
+        for node in powernode_machines:
+            cmd = f"ssh core@{node.name} sudo systemctl stop kubelet.service"
+            result = exec_cmd(cmd)
```
petr-balogh, line 79: So basically this will be moved out of here to service.py module.  And I would then also check the real status of the service there if it's really stopped.

ocs_ci/utility/powernodes.py
gitsridhar: Petr and Shylesh, thanks for your quick comments. I will amend this PR shortly based on your comments.

I am doing this based on one observation: when I manually ssh into a storage node and stop kubelet service, the node moves to the NotReady state and in about 5 minutes, all pods running on that node get terminated and either relocated to another node or moves to the pending state. When kubelet service is started, pods get back to the original Ready state. I understand this is equivalent to shutting down the entire storage node and starting it back after few minutes by ocs-ci disruptive test cases. Please comment if this understanding is not correct.

gitsridhar: @shylesh I am doing only part of what you said (tie into IBMPowerNodes..) but I agree with you on class separation, so it can be used by anyone for any service.

gitsridhar (commented):
I tried to gather ExternalIP for node to use it instead of node name. Here is what I found in my OCP 4.6 environment. I do not see ExternalIP. I see only InternalIP. I am planning to use the InternalIP.

status:
  addresses:
  - address: 10.17.120.156
    type: InternalIP
  - address: master0.sridhar-ocp-ocs.cp.fyre.ibm.com
    type: Hostname
  allocatable:

@petr-balogh Can you comment on this?

ocs_ci/utility/powernodes.py
gitsridhar: @shylesh @petr-balogh @vasukulkarni I just pushed an amend to PR, I have addressed all of your comments.

ocs_ci/utility/powernodes.py
gitsridhar: @petr-balogh I used InternalIP and submitted an amendment to PR. Let me know if there are any concerns.

gitsridhar:
My latest PR amendment has an implementation of Service and KubeletService classes which can be re-used by any other module in OCS-CI.

ocs_ci/utility/powernodes.py
```diff
@@ -90,6 +119,23 @@ def stop_powernodes_machines(
...
+    ):
+        """
+        Start PowerNode Machines
+
+        Args:
+            powernode_machines (list): List of PowerNode machines
+            timeout (int): time in seconds to wait for node to reach 'not ready' state,
+                and 'ready' state.
+            wait (bool): Wait for PowerNodes to start - for future use
+            force (bool): True for PowerNode ungraceful power off, False for
```
vasukulkarni, line 133: instead of 'force', can you call this `using_method=kube-service`, I think that will give users who are reading code better insights into what method is used for disruption.  You will have to fix at few more places below.

ocs_ci/utility/service.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def __init__(self, service_name, force=False):
+        """
+        Class Initialization.
+
+        Initialize the service name local variable and collect a dictionary of Internel IP addresses
+        of nodes with node name as keys.
+        """
+        self.service_name = service_name
+        self.force = force
```
vasukulkarni, line 33: I think you can get rid this option in this class,  The class doesn't need to be aware of this parameter.

vasukulkarni: unless this `force` option is different from what is mentioned above.

ocs_ci/ocs/platform_nodes.py
```diff
@@ -1736,8 +1736,8 @@ def stop_nodes(self, nodes, force=True):
                 nodes, timeout=900, wait=True, force=force
             )
         else:
-            raise NotImplementedError(
-                "This is not libvirt environment. Stop nodes not implemented"
+            self.powernodes.stop_powernodes_machines_powervs(
+                nodes, timeout=900, wait=True, force=force
```
vasukulkarni, line 1740: `use_method=kube-service` is better option than `force`

ocs_ci/utility/powernodes.py
gitsridhar: @vasukulkarni I see your point, but the parameter force is used as a boolean value and specifically, in stop methods, its boolean value is used to add '-f' parameter to systemctl command. I have enhanced the logging in server.py to include service name (like kubelet.service) so it will be very clear on which component and service is invoking the code.

ocs_ci/utility/service.py
gitsridhar: Vasu, see my comment above, the force is used as a boolean parameter.

ocs_ci/ocs/platform_nodes.py
gitsridhar: Vasu, see my other comments above.

gitsridhar:
@vasukulkarni @petr-balogh @clacroix12 @shylesh  I have updated the code based on your comments and tested and removed the 'WIP' prefix as well. Please review and merge if everything is ok. This work will help any component of ocs-ci to use systemctl and control any service on any master or worker nodes.

gitsridhar:
@vasukulkarni @petr-balogh @clacroix12 @shylesh  Please take a look and merge if everything is ok?

ocs_ci/ocs/platform_nodes.py
vasukulkarni: @gitsridhar I see there are 2 force options, I meant changing of the the force of stop_power and start_power functions to use `use_method` to distinguish between service and real power restarts

ocs_ci/utility/powernodes.py
```diff
@@ -56,6 +61,30 @@ def verify_machine_is_down(self, node):
...
+
+        """
+        for pnode in powernode_machines:
+            self.service.stop(pnode, timeout)
+
+        # Wait for an additional 300+60 seconds (for pods to drain)
+        waiting_time = 360
+        logger.info(f"Waiting for {waiting_time} seconds")
+        time.sleep(waiting_time)
+        logger.info(f"Waiting over for {waiting_time} seconds")
```
clacroix12, line 86: This log message is similar to the one before and makes it seem like we're waiting again. I would make this a bit more clear("Done waiting" or "Continuing...") or simply remove this log statement entirely.

ocs_ci/utility/service.py
```diff
@@ -0,0 +1,247 @@
...
+        return result.stdout.lower().rstrip()
+
+
+class KubeletService(Service):
+    """
+    Kubelet Service class
+
+    The purpose of this class is to extend Service class to provide stop/start/restart etc operations on
+    kubelet service. Since kubelet service stop and start operations involve changing status of OCP node
+    obects, this class verifies the same.
```
clacroix12, line 191:
```suggestion
    objects, this class verifies the same.
```

clacroix12 (commented):
This looks great! Really like the separation of the generic service function from the extra verification done in the `KubeletService` class. I just had a few minor comments but nothing that should hold this PR up.

ocs_ci/ocs/platform_nodes.py
gitsridhar: @vasukulkarni I have simplified the service, There are no more two different 'force' parameters. There is only one now when the service object is initialized, the force parameter can be initialized.  Hope this clears the confusion over the force parameter.

ocs_ci/ocs/node.py
```diff
@@ -278,6 +278,30 @@ def get_node_ips(node_type="worker"):
...
 
+def get_node_ip_addresses(ipkind):
+    """
+    Gets a dictionary of required IP addresses for all nodes
+
+    Args:
+        ipkind: ExternalIP or InternalIP or Hostname
+
+    Returns:
+        dictionary: Internal or Exteranl IP addresses keyed off of node name
```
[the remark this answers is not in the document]
gitsridhar: Done.

ocs_ci/utility/service.py
gitsridhar: Done.

ocs_ci/utility/powernodes.py
gitsridhar: Done. I removed the second log parameter.

ocs_ci/utility/service.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        Stop the kubelet service using parent service class. After that, ensures the corresponding OCP node
+        moves to NotReady state.
+
+        Args:
+            node (object): Node objects
+            timeout (int): time in seconds to wait for service to stop.
+
+        """
+        super().stop(node, timeout)
```
vasukulkarni, line 207: The implementation looks reversed,  we might actually not need either parent or child class, since you are doing heavy lifting in parent and waiting in child.  I will approve the PR not to hold it further.

vasukulkarni (approved):

clacroix12 (approved):

ocs_ci/utility/service.py
gitsridhar: I will revisit this in the next iteration. I feel that parent class should do as much in a generic way and child class should do whatever is specific to themselves. I will initiate a discussion when we get time to revisit this design.

Issue #2744, opened 2020-08-18
Observing memory exhaust issue in Jenkins slave while running scale tests for 1500+ pod or pvc creations


ramkiperiy:
The real problem is when there is continuous execution of creating/deleting pvc or pod in parallel, run-ci process memory consumption is increasing and its not dropping down which leads to failed to allocate memory.
Provided top output from a BM machine.

04:47:39 - MainThread - ocs_ci.framework.pytest_customization.ocscilib - ERROR - Failed to collect OCS logs. Error: [Errno 12] Cannot allocate memory
04:47:39 - MainThread - ocs_ci.framework.pytest_customization.ocscilib - ERROR - Failed to collect performance stats. Error: [Errno 12] Cannot allocate memory

Executed the same test in BM machine observed below output on top, and the memory consumption is growing.
    PID  USER    PR  NI  VIRT    RES  SHR   S %CPU  %MEM  TIME+   COMMAND
 589649 root      20   0   22.9g   4.0g  32264 S  106.6     1.1     31:23.51    run-ci

Another Top Output from BM machine
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                                                               
 589649 root      20   0   33.8g  10.4g  32264 S 105.3   2.8 187:39.56 run-ci    

```

                    # This matches the behavior of os._execvpe().
                    executable_list = tuple(
                        os.path.join(os.fsencode(dir), executable)
                        for dir in os.get_exec_path(env))
                fds_to_keep = set(pass_fds)
                fds_to_keep.add(errpipe_write)
                self.pid = _posixsubprocess.fork_exec(
                        args, executable_list,
                        close_fds, tuple(sorted(map(int, fds_to_keep))),
                        cwd, env_list,
                        p2cread, p2cwrite, c2pread, c2pwrite,
                        errread, errwrite,
                        errpipe_read, errpipe_write,
>                       restore_signals, start_new_session, preexec_fn)
E                       OSError: [Errno 12] Cannot allocate memory
```

Observed this problem in multiple Test runs
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/10814//console
https://ocs4-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/qe-deploy-ocs-cluster/10971/console
[...]

PR #3871, opened 2021-02-17
Noobaa scale obc creation

tiffanyn108:
Creation of OBCs in MCG and RGW.

### DIFF
--- a/ocs_ci/ocs/scale_noobaa_lib.py
+++ b/ocs_ci/ocs/scale_noobaa_lib.py
@@ -4,7 +4,6 @@ import time
 from ocs_ci.helpers import helpers
 from ocs_ci.utility import templating
 from ocs_ci.ocs import constants
-
 from ocs_ci.ocs.utils import oc_get_all_obc_names
 from ocs_ci.utility.utils import run_cmd
 
@@ -41,7 +40,7 @@ def construct_obc_creation_yaml_bulk_for_kube_job(no_of_obc, sc_name, namespace)
 
 
 def check_all_obc_reached_bound_state_in_kube_job(
-    kube_job_obj, namespace, no_of_obc, timeout=300, wait_time=20
+    kube_job_obj, namespace, no_of_obc, timeout=300, no_wait_time=20
 ):
     """
     Function to check either bulk created OBCs reached Bound state using kube_job
@@ -51,7 +50,7 @@ def check_all_obc_reached_bound_state_in_kube_job(
         namespace (str): Namespace of OBC's created
         no_of_obc (int): Bulk OBC count
         timeout (second): a timeout for all the obc in kube job to reach bound state
-        wait_time (int): number of wait time to ensure all OCBs to reach bound state
+        no_wait_time (int): number of wait time to ensure all OCBs to reach bound state
 
     Returns:
         obc_bound_list (list): List of all OBCs which is in Bound state.
@@ -67,28 +66,27 @@ def check_all_obc_reached_bound_state_in_kube_job(
         # Get kube_job obj and fetch either all OBC's are in Bound state
         # If not bound adding those OBCs to obc_not_bound_list
         job_get_output = kube_job_obj.get(namespace=namespace)
-        for i in range(no_of_obc):
-            status = job_get_output["items"][i]["status"]["phase"]
-            time.sleep(5)
-            # while not status:
-            #     status = job_get_output["items"][i]["status"]["phase"]
-            #    time.sleep(5)
-            logging.info(
-                f"obc {job_get_output['items'][i]['metadata']['name']} status {status}"
-            )
-            if status != "Bound":
-                obc_not_bound_list.append(
-                    job_get_output["items"][i]["metadata"]["name"]
+        if job_get_output is not None and len(job_get_output) == no_of_obc:
+            for i in range(no_of_obc):
+                status = job_get_output["items"][i]["status"]["phase"]
+                time.sleep(5)
+                log.info(
+                    f"obc {job_get_output['items'][i]['metadata']['name']} status {status}"
                 )
+                if status != "Bound":
+                    obc_not_bound_list.append(
+                        job_get_output["items"][i]["metadata"]["name"]
+                    )
 
         # Check the length of obc_not_bound_list to decide either all OBCs reached Bound state
         # If not then wait for timeout secs and re-iterate while loop
         if len(obc_not_bound_list):
-            time.sleep(timeout)
+            log.info(f"Number of OBCs are not in Bound state {len(obc_not_bound_list)}")
             while_iteration_count += 1
-            if while_iteration_count >= wait_time:
-                assert logging.error(
-                    f" Listed OBCs took more than {timeout*wait_time} secs to be bounded {obc_not_bound_list}"
+            if while_iteration_count >= no_wait_time:
+                assert log.error(
+                    f" Listed OBCs took more than {timeout*no_wait_time} "
+                    f"secs to be bounded {obc_not_bound_list}"
                 )
                 break
             obc_not_bound_list.clear()
@@ -96,7 +94,7 @@ def check_all_obc_reached_bound_state_in_kube_job(
         elif not len(obc_not_bound_list):
             for i in range(no_of_obc):
                 obc_bound_list.append(job_get_output["items"][i]["metadata"]["name"])
-            logging.info("All OBCs in Bound state")
+            log.info("All OBCs in Bound state")
             break
     return obc_bound_list
 
--- a/ocs_ci/ocs/scale_noobaa_lib.py
+++ b/ocs_ci/ocs/scale_noobaa_lib.py
@@ -40,7 +40,7 @@ def construct_obc_creation_yaml_bulk_for_kube_job(no_of_obc, sc_name, namespace)
 
 
 def check_all_obc_reached_bound_state_in_kube_job(
-    kube_job_obj, namespace, no_of_obc, timeout=300, no_wait_time=20
+    kube_job_obj, namespace, no_of_obc, timeout=120, no_wait_time=20
 ):
     """
     Function to check either bulk created OBCs reached Bound state using kube_job
@@ -69,7 +69,6 @@ def check_all_obc_reached_bound_state_in_kube_job(
         if job_get_output is not None and len(job_get_output) == no_of_obc:
             for i in range(no_of_obc):
                 status = job_get_output["items"][i]["status"]["phase"]
-                time.sleep(5)
                 log.info(
                     f"obc {job_get_output['items'][i]['metadata']['name']} status {status}"
                 )
@@ -82,6 +81,7 @@ def check_all_obc_reached_bound_state_in_kube_job(
         # If not then wait for timeout secs and re-iterate while loop
         if len(obc_not_bound_list):
             log.info(f"Number of OBCs are not in Bound state {len(obc_not_bound_list)}")
+            time.sleep(timeout)
             while_iteration_count += 1
             if while_iteration_count >= no_wait_time:
                 assert log.error(
--- a/ocs_ci/ocs/scale_noobaa_lib.py
+++ b/ocs_ci/ocs/scale_noobaa_lib.py
@@ -20,11 +20,12 @@ def construct_obc_creation_yaml_bulk_for_kube_job(no_of_obc, sc_name, namespace)
         sc_name (str): storage class name using for obc creation
         namespace(str): Namespace uses to create bulk of obc
     Returns:
+
          obc_dict_list (list): List of all obc.yaml dicts
 
     """
 
-    # Construct obc.yaml for the no_of_required_pvc count
+    # Construct obc.yaml for the no_of_obc count
     # append all the obc.yaml dict to obc_dict_list and return the list
     obc_dict_list = list()
     for i in range(no_of_obc):
@@ -40,7 +41,7 @@ def construct_obc_creation_yaml_bulk_for_kube_job(no_of_obc, sc_name, namespace)
 
 
 def check_all_obc_reached_bound_state_in_kube_job(
-    kube_job_obj, namespace, no_of_obc, timeout=300
+    kube_job_obj, namespace, no_of_obc, timeout=300, wait_time=20
 ):
     """
     Function to check either bulk created OBCs reached Bound state using kube_job
@@ -49,13 +50,14 @@ def check_all_obc_reached_bound_state_in_kube_job(
         kube_job_obj (obj): Kube Job Object
         namespace (str): Namespace of OBC's created
         no_of_obc (int): Bulk OBC count
-        timeout: a timeout for all the obc in kube job to reach bound status
+        timeout (second): a timeout for all the obc in kube job to reach bound state
+        wait_time (int): number of wait time to ensure all OCBs to reach bound state
 
     Returns:
         obc_bound_list (list): List of all OBCs which is in Bound state.
 
-    Asserts:
-        If not all OBC reached to Bound state.
+    Raises:
+        AssertionError: If not all OBC reached to Bound state
 
     """
     # Check all OBCs to reach Bound state
@@ -67,9 +69,10 @@ def check_all_obc_reached_bound_state_in_kube_job(
         job_get_output = kube_job_obj.get(namespace=namespace)
         for i in range(no_of_obc):
             status = job_get_output["items"][i]["status"]["phase"]
-            while not status:
-                status = job_get_output["items"][i]["status"]["phase"]
-                time.sleep(5)
+            time.sleep(5)
+            # while not status:
+            #     status = job_get_output["items"][i]["status"]["phase"]
+            #    time.sleep(5)
             logging.info(
                 f"obc {job_get_output['items'][i]['metadata']['name']} status {status}"
             )
@@ -83,9 +86,9 @@ def check_all_obc_reached_bound_state_in_kube_job(
         if len(obc_not_bound_list):
             time.sleep(timeout)
             while_iteration_count += 1
-            if while_iteration_count >= 20:
+            if while_iteration_count >= wait_time:
                 assert logging.error(
-                    f" Listed OBCs took more than {timeout*20} secs to bound {obc_not_bound_list}"
+                    f" Listed OBCs took more than {timeout*wait_time} secs to be bounded {obc_not_bound_list}"
                 )
                 break
             obc_not_bound_list.clear()
@@ -101,8 +104,10 @@ def check_all_obc_reached_bound_state_in_kube_job(
 def cleanup(namespace):
     """
     Delete all OBCs created in the cluster
+
     Args:
         namespace (str): Namespace of OBC's deleting
+
     """
     obc_name_list = oc_get_all_obc_names()
     log.info(f"Deleting {len(obc_name_list)} OBCs")
--- a/ocs_ci/ocs/scale_noobaa_lib.py
+++ b/ocs_ci/ocs/scale_noobaa_lib.py
@@ -1,0 +1,111 @@
+import logging
+import time
+
+from ocs_ci.helpers import helpers
+from ocs_ci.utility import templating
+from ocs_ci.ocs import constants
+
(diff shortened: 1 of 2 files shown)


wusui (approved):
Looks good.  A couple of minor nits.

Line 88 of scale_noobaa_lib.py "secs to be bounded" may be more grammatically correct.

Lines 30, 62 in scale_nooba_lib.py and Lines 85 and 127 in test_scale_obc_creation.py use i as an index but do not  use that index value within the loop.  This probably is not an issue for us, but some linters may flag this as an issue.  There is a common idiom used where _ is used as a variable in this case.

I don't think either of these issues should slow down checking this in, however.

tests/e2e/scale/noobaa/test_scale_obc_creation.py
```diff
@@ -0,0 +1,184 @@
...
+    OBCs are created in the Multicloud Object Gateway and
+    Ceph Object Gateway (RGW)
+
+    """
+
+    namespace = constants.OPENSHIFT_STORAGE_NAMESPACE
+    sc_name = constants.NOOBAA_SC
+    sc_rgw_name = constants.DEFAULT_STORAGECLASS_RGW
+    scale_obc_count = 1000
+    num_obc_batch = 50
```
ramkiperiy, line 29: @tiffanyn108 any reason for lesser batch count ?, in PVC creation using kubejob I tried using 750 PVC per batch and it worked without any issues.

ramkiperiy:
@tiffanyn108 Nice work !! and It's a good start with Noobaa scale

tests/e2e/scale/noobaa/test_scale_obc_creation.py
tiffanyn108: obc creation in larger batch is having issue which causes obc in unbound state. According noobaa developer, we are supporting up to 1000 obcs but not creating them in a same time or in few seconds apart.

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def construct_obc_creation_yaml_bulk_for_kube_job(no_of_obc, sc_name, namespace):
+    """
+    Constructing obc.yaml file to create bulk of obc's using kube_job
+
+    Args:
+        no_of_obc(int): Bulk obc count
+        sc_name (str): storage class name using for obc creation
+        namespace(str): Namespace uses to create bulk of obc
```
ebenahar, line 21: add an empty line below

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,111 @@
...
+        kube_job_obj (obj): Kube Job Object
+        namespace (str): Namespace of OBC's created
+        no_of_obc (int): Bulk OBC count
+        timeout: a timeout for all the obc in kube job to reach bound status
+
+    Returns:
+        obc_bound_list (list): List of all OBCs which is in Bound state.
+
+    Asserts:
+        If not all OBC reached to Bound state.
```
ebenahar, line 58:
```python
"""

Raises:
    AssertionError: If not all OBC reached to Bound state

"""
```

tests/e2e/scale/noobaa/test_scale_obc_creation.py
```diff
@@ -0,0 +1,184 @@
...
+            )
+            log.info(f"Number of PVCs in Bound state {len(obc_bound_list)}")
+        # Delete obc on cluster
+        scale_noobaa_lib.cleanup(self.namespace)
+
+    @vsphere_platform_required
+    @pytest.mark.polarion_id("OCS-2479")
+    def test_scale_rgw_obc_creation(self, tmp_path, timeout=60):
+        """
+        MCG OBC creation using RGW storage class
```
MeridianExplorer, line 77:
```suggestion
        RGW OBC creation
```

OBCs can be provisioned by _either_ MCG, or RGW - it can never be both :)
It's possible to create an MCG OBC that utilizes RGW as a backingstore - but that's a different construct

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,111 @@
...
+    # Check all OBCs to reach Bound state
+    obc_bound_list, obc_not_bound_list = ([] for i in range(2))
+    while_iteration_count = 0
+    while True:
+        # Get kube_job obj and fetch either all OBC's are in Bound state
+        # If not bound adding those OBCs to obc_not_bound_list
+        job_get_output = kube_job_obj.get(namespace=namespace)
+        for i in range(no_of_obc):
+            status = job_get_output["items"][i]["status"]["phase"]
+            while not status:
```
MeridianExplorer, line 70: We've seen rare cases in which OBCs never entered any particular phase. 
This might lead to an infinite loop - I recommend adding a timeout mechanism

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,111 @@
...
+    kube_job_obj, namespace, no_of_obc, timeout=300
+):
+    """
+    Function to check either bulk created OBCs reached Bound state using kube_job
+
+    Args:
+        kube_job_obj (obj): Kube Job Object
+        namespace (str): Namespace of OBC's created
+        no_of_obc (int): Bulk OBC count
+        timeout: a timeout for all the obc in kube job to reach bound status
```
MeridianExplorer, line 52: I think it's important to note the timeout is in seconds and will be checked 20 times.
Alternatively, I think it'd be best to replace the hardcoded `*20` with a parameter given by the user

MeridianExplorer (commented):
Looks good :)
Was usage of the `bucket_factory()` fixture considered for those tests?
If yes, what is the reasoning behind not using it?

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,111 @@
...
+
+    """
+    # Check all OBCs to reach Bound state
+    obc_bound_list, obc_not_bound_list = ([] for i in range(2))
+    while_iteration_count = 0
+    while True:
+        # Get kube_job obj and fetch either all OBC's are in Bound state
+        # If not bound adding those OBCs to obc_not_bound_list
+        job_get_output = kube_job_obj.get(namespace=namespace)
+        for i in range(no_of_obc):
```
ebenahar, line 68: what if `no_of_obc` is not equal to `len(job_get_output["items"])`?
Maybe add validation for that above

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,111 @@
...
+            )
+            if status != "Bound":
+                obc_not_bound_list.append(
+                    job_get_output["items"][i]["metadata"]["name"]
+                )
+
+        # Check the length of obc_not_bound_list to decide either all OBCs reached Bound state
+        # If not then wait for timeout secs and re-iterate while loop
+        if len(obc_not_bound_list):
+            time.sleep(timeout)
```
ebenahar, line 84: this will hand the logger for 6 minutes (unless specified otherwise).
Is there anything to poll instead of waiting?
In case not, please add a massage with the current status, logged every few seconds (maybe 30 seconds)

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,111 @@
...
+            while_iteration_count += 1
+            if while_iteration_count >= 20:
+                assert logging.error(
```
ebenahar, line 87: not sure what will `assert logging..` will do. 
I think you need to raise some kind of a timeout expired exception instead of assert

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,111 @@
...
+                obc_bound_list.append(job_get_output["items"][i]["metadata"]["name"])
+            logging.info("All OBCs in Bound state")
+            break
+    return obc_bound_list
+
+
+def cleanup(namespace):
+    """
+    Delete all OBCs created in the cluster
+    Args:
```
ebenahar, line 104: add an empty line above

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,111 @@
...
+        namespace (str): Namespace of OBC's deleting
```
ebenahar, line 105: add an empty line below

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    obc_name_list = oc_get_all_obc_names()
+    log.info(f"Deleting {len(obc_name_list)} OBCs")
+    if obc_name_list:
+        for i in obc_name_list:
+            run_cmd(f"oc delete obc {i} -n {namespace}")
```
ebenahar, line 111: is there a way you can use the existing obc related factory based fixtures?
This will make the test more robust

ebenahar (changes requested):

tests/e2e/scale/noobaa/test_scale_obc_creation.py
```diff
@@ -0,0 +1,184 @@
...
+            obc_bound_list = (
+                scale_noobaa_lib.check_all_obc_reached_bound_state_in_kube_job(
+                    kube_job_obj=job_file,
+                    namespace=self.namespace,
+                    no_of_obc=self.num_obc_batch,
+                )
+            )
+            log.info(f"Number of PVCs in Bound state {len(obc_bound_list)}")
+        # Delete obc on cluster
+        scale_noobaa_lib.cleanup(self.namespace)
```
fbalak, line 71: Teardown should be in some finalizer to ensure that objects are cleaned up when test fails.

fbalak (changes requested):
Can `scale_noobaa_lib` be replaced with usage of [OBC class](https://github.com/red-hat-storage/ocs-ci/blob/dae09a2efc83bed6fcd6bb5ad9f3380eba044aeb/ocs_ci/ocs/resources/objectbucket.py) (maybe with some modifications)?

If I understand it correctly, now there are two modules used for OBC creation and manipulation that ignore each other. You can add functions that manipulate with OBCs into the OBC class.

You can also create `obc_factory` fixture in [tests/conftest.py](https://github.com/red-hat-storage/ocs-ci/blob/dae09a2efc83bed6fcd6bb5ad9f3380eba044aeb/tests/conftest.py) and use it in tests. That would solve call for teardown when tests fail.

tests/e2e/scale/noobaa/test_scale_obc_creation.py
```diff
@@ -0,0 +1,184 @@
...
+
+    """
+
+    namespace = constants.OPENSHIFT_STORAGE_NAMESPACE
+    sc_name = constants.NOOBAA_SC
+    sc_rgw_name = constants.DEFAULT_STORAGECLASS_RGW
+    scale_obc_count = 1000
+    num_obc_batch = 50
+
+    @skipif_aws_creds_are_missing
```
mbukatov, line 31: Does it mean that this will work on AWS cloud platform only? Afaik there is ongoing effort to make sure noobaa test cases work on all cloud platforms.

MeridianExplorer: The way I see it, there is no reason for the test to be limited to any one platform - cloud or on-prem.
Even if it needed other cloud providers, we already have working accounts for AWS, GCP, and Azure.
The IBM COS infra is in place, we're waiting for credentials

fbalak: I agree with @MeridianExplorer. Multi-cloud tests shouldn't be limited to only one platform.
That said, we should cover more cloud providers and don't focus only on aws (we can parametrize tests with cloud providers or create new test cases for other providers).

MeridianExplorer: The way the test is written right now, the created OBCs will use the default backingstore. So if the cluster runs on Azure, it'll create N buckets that all lead to one Azure container. If the cluster runs on GCP, the test will create N buckets that all lead to one Google bucket, etc.

Either way, `bucket_factory()` takes care of it all automagically. As Filip said - it's just about passing the right parameters to the fixture.

tests/e2e/scale/noobaa/test_scale_obc_creation.py
tiffanyn108: Missed type, thanks for catching this.  Fixed.

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,111 @@
...
+    Args:
+        no_of_obc(int): Bulk obc count
+        sc_name (str): storage class name using for obc creation
+        namespace(str): Namespace uses to create bulk of obc
+    Returns:
+         obc_dict_list (list): List of all obc.yaml dicts
+
+    """
+
+    # Construct obc.yaml for the no_of_required_pvc count
```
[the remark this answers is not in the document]
tiffanyn108: Fixed.

ocs_ci/ocs/scale_noobaa_lib.py
tiffanyn108: Acked, fixed.

ocs_ci/ocs/scale_noobaa_lib.py
tiffanyn108: Acked, fixed.

ocs_ci/ocs/scale_noobaa_lib.py
tiffanyn108: Acked.

ocs_ci/ocs/scale_noobaa_lib.py
tiffanyn108: Acked.

tests/e2e/scale/noobaa/test_scale_obc_creation.py
tiffanyn108: @MeridianExplorer @fbalak, For scale testing, we are experiencing memory issues when using default factory. 
When creating a large amount of PVCs/OBCs, there are many objects are created and they consume all memory on Jenkins slave which causes the test failing.  This issue you don't see in normal testing with small number of pvc/obc creation.
The recent incident that happened with PVC creation https://github.com/red-hat-storage/ocs-ci/issues/2744.
After multiple discussions, we decide to change the test case implementation by using kube_job instead.  We have seen the big improvement in terms of memory usage and shorten the time of PVC creation as well.

tests/e2e/scale/noobaa/test_scale_obc_creation.py
tiffanyn108: Acked.  This can be fixed.

ocs_ci/ocs/scale_noobaa_lib.py
tiffanyn108: Added "wait_time" parameter and updated the timeout in seconds.

tests/e2e/scale/noobaa/test_scale_obc_creation.py
tiffanyn108: @mbukatov I will remove aws creds check since we are supporting other cloud platforms.

MeridianExplorer: @tiffanyn108 - I understand, thanks for the elaboration.
Was `bucket_factory()` tested in scale, or was the decision based solely on PVC creation?
While I can understand why PVC objects can consume a lot of memory (being dependent on containing OCS and OCP objects within them) - OB was actually designed to be [extremely lightweight](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/resources/objectbucket.py#L132-L141), and only contains 5 short strings per object (unlike the heavy-duty OBC which would not be needed/used for those tests)

tiffanyn108: @MeridianExplorer I have done a comparison of memory usage between creating 1000 obcs using `bucket_factory()` and using `kube_job`.  Here is what I saw:
* `bucket_factory()` - Memory usage is ~2.2G - 2.5G 
* `kube_job` - Memory usage is ~500MB
Base on my observation, we can save ~5 times of memory used with kube_job implementation.  
In addition, with kube_job, the test execution time is 20mins faster.  I would refer to use kube_job for scale testing, especially with all the loads and large number of bucket creation.

MeridianExplorer: That's a very interesting and important benchmark. Thanks, Tifanny!

ramkiperiy (approved):
LGTM

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,114 @@
...
+    obc_bound_list, obc_not_bound_list = ([] for i in range(2))
+    while_iteration_count = 0
+    while True:
+        # Get kube_job obj and fetch either all OBC's are in Bound state
+        # If not bound adding those OBCs to obc_not_bound_list
+        job_get_output = kube_job_obj.get(namespace=namespace)
+        if job_get_output is not None and len(job_get_output) == no_of_obc:
+            for i in range(no_of_obc):
+                status = job_get_output["items"][i]["status"]["phase"]
+                time.sleep(5)
```
keesturam, line 72: why do we need this sleep? 
For 1000 OBC creation, we'd wait for 5000 seconds. 
Also, since this is about checking the status, we should parallelize this operation.

tiffanyn108: Remove timeout.

ramkiperiy (approved):

wusui (approved):
Looks good

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,114 @@
...
+        # Get kube_job obj and fetch either all OBC's are in Bound state
+        # If not bound adding those OBCs to obc_not_bound_list
+        job_get_output = kube_job_obj.get(namespace=namespace)
+        if job_get_output is not None and len(job_get_output) == no_of_obc:
+            for i in range(no_of_obc):
+                status = job_get_output["items"][i]["status"]["phase"]
+                log.info(
+                    f"obc {job_get_output['items'][i]['metadata']['name']} status {status}"
+                )
+                if status != "Bound":
```
MeridianExplorer, line 75: I'd recommend using the `STATUS_BOUND` or `HEALTHY_OBC` constants instead

ocs_ci/ocs/scale_noobaa_lib.py
```diff
@@ -0,0 +1,114 @@
...
+    # Check all OBCs to reach Bound state
+    obc_bound_list, obc_not_bound_list = ([] for i in range(2))
+    while_iteration_count = 0
+    while True:
+        # Get kube_job obj and fetch either all OBC's are in Bound state
+        # If not bound adding those OBCs to obc_not_bound_list
+        job_get_output = kube_job_obj.get(namespace=namespace)
+        if job_get_output is not None and len(job_get_output) == no_of_obc:
+            for i in range(no_of_obc):
+                status = job_get_output["items"][i]["status"]["phase"]
```
MeridianExplorer, line 71: The `phase` field might take some time to generate, and I've seen cases where it doesn't exist, so `.get()` might be safer than direct access (`["phase"]`)

MeridianExplorer (approved):

ocs_ci/ocs/scale_noobaa_lib.py
tiffanyn108: @MeridianExplorer Yes, this can be improved.  Will address this in later PR.

ocs_ci/ocs/scale_noobaa_lib.py
tiffanyn108: Yes, this can be improved.  Will address this in later PR.

fbalak (approved):

ebenahar (approved):

Issue #3641, opened 2021-01-08
Issues from the OCS-CI Gaps document 


vasukulkarni:
Some of the issues are, we can create a new sub-issue or use this as master list to strikedown sub-issue

- It does not make sense for me to replace my sed tool (or any other tool) with gnu-sed in PATH. I will suggest to just use gnu-sed inside the ci scripts instead
 

-  Unabling to deploy/teardown a cluster creates a problem when we are trying to contribute deployment/teardown code, we need to find a way to enable us to test these code paths an not wait for a verification job run to find bugs 

- It is not very easy to run the test suite(s) locally, aka trigger them locally or somewhere to run against some setup of openshift, with your own builds. 

> This is probably somehting that user is missing on how to setup locally and run,  It is quite easy to run locally, cluster setup is something different from running ocs-ci, those reqiurements have to be met by operator installer

- the need for the pull secret but the docs didn't make clear which of the other keys were required and where to get them, My ideal for the "Getting Started" doc is that it be more of a "one stop shop" for someone who simply wants to do a basic run of the tool.  

- There are many options in the YAML config that are unclear to me and many of the options are not documented inline. The subsections for "RUN", "DEPLOYMENT", and "ENV_DATA" are unclear and I wouldn't know what section to look for a particular option in 

- as a design point I think that module import should not trigger complex code or use resources outside of the python source files themselves. If network resources are needed to create decorators, they should be deferred until the function itself is going to run

- Fixtures should be reserved for resources that have an external dependency such as a persistent network connection, database, etc. or complex/expensive data structure. Even then I recommend the fixture be a minimal wrapper around library functions whenever possible

-  I would recommend the use of a tool such as "black". Many in the Python community now use "black" to unify style and avoid having to make developers pep8 conscious.

> Fixed by issue: https://github.com/red-hat-storage/ocs-ci/issues/2952

clacroix12:
> It does not make sense for me to replace my sed tool (or any other tool) with gnu-sed in PATH. I will suggest to just use gnu-sed inside the ci scripts instead

We don't control the scripts that are calling `sed` directly which is why I opted to have mac users deploying UPI configure a `sed` that is compatible with the scripts.

> There are many options in the YAML config that are unclear to me and many of the options are not documented inline. The subsections for "RUN", "DEPLOYMENT", and "ENV_DATA" are unclear and I wouldn't know what section to look for a particular option in

[...]

PR #3890, opened 2021-02-22
build and push ocs-ci container image to quay

vasukulkarni:
Partial Fixes one of the item in: https://github.com/red-hat-storage/ocs-ci/issues/3641

This builds the ocs-ci container image and pushes to quay.io/ocsci/ocs-ci based on push to latest branch.

### DIFF
--- a/.github/workflows/build_quay.yml
+++ b/.github/workflows/build_quay.yml
@@ -1,0 +1,27 @@
+name: "Build Quay image"
+on:
+  push:
+    branches:
+      - 'stable*'
+
+jobs:
+  quay_build:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Login to Quay
+        uses: docker/login-action@v1
+        with:
+          registry: ${{ secrets.QUAY_URI }}
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+      -
+        name: Build and push
+        id: quay_build
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: quay.io/ocsci/ocs-ci:latest
+      -
+        name: Image digest
+        run: echo ${{ steps.docker_build.outputs.digest }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+# set base image (host OS)
+FROM centos:latest
+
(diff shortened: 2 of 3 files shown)


docs/getting_started.md
```diff
@@ -63,6 +63,14 @@ necessary dependencies
    * `pip install -r requirements-dev.txt`
    * `pre-commit install --hook-type pre-commit --hook-type commit-msg`
 
+
+## OCS-CI Container Image
+  We also have ocs-ci container image available at quay.io/ocsci/ocs-ci, the image
+  is build using the latest branch. It is still experimental as the ci requires
```
clacroix12, line 69: This says 'latest branch' but it looks like we're pushing this image when the stable branch is updated. Should that be mentioned here instead?

vasukulkarni: that is correct, thanks I will fix it.

.github/workflows/build_quay.yml
```diff
@@ -0,0 +1,27 @@
+name: "Build Quay image"
+on:
+  push:
+    branches:
+      - 'stable*'
+
+jobs:
+  quay_build:
+    runs-on: ubuntu-latest
```
mbukatov, line 9: Would it make sense to run this on fedora instead?

Dockerfile
```diff
@@ -0,0 +1,18 @@
+# set base image (host OS)
+FROM centos:latest
```
mbukatov, line 2: Could we use fully qualified container image name?

.github/workflows/build_quay.yml
```diff
@@ -0,0 +1,27 @@
...
+          registry: ${{ secrets.QUAY_URI }}
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+      -
+        name: Build and push
+        id: quay_build
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: quay.io/ocsci/ocs-ci:latest
```
mbukatov, line 24: We will likely also want to tag a container image using git tag of coresponding stable build (eg. `ee also stable-ocs-4.7-202102180854`). See https://github.com/red-hat-storage/ocs-ci/tags

.github/workflows/build_quay.yml
mbukatov: Image name could likely be `registry.fedoraproject.org/fedora:latest`.

PR #3919, opened 2021-02-25
Test snapshot restore with volume mode and access modes different than parent PVC

jilju:
Test snapshot restore with volume mode and access modes different than parent PVC

### DIFF
--- a/tests/manage/pv_services/pvc_snapshot/test_restore_with_different_volume_mode.py
+++ b/tests/manage/pv_services/pvc_snapshot/test_restore_with_different_volume_mode.py
@@ -129,12 +129,10 @@ class TestSnapshotRestoreWithDifferentVolumeMode(ManageTest):
         access_modes_dict = {
             constants.VOLUME_MODE_FILESYSTEM: [
                 constants.ACCESS_MODE_RWO,
-                constants.ACCESS_MODE_ROX,
             ],
             constants.VOLUME_MODE_BLOCK: [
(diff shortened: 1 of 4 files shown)


jilju:
Rebased

petr-balogh:
@jilju are you going to finish with this PR? Needs rebase.

tests/manage/pv_services/pvc_snapshot/test_restore_with_different_volume_mode.py
```diff
@@ -0,0 +1,434 @@
...
+        )
+
+    def test_snapshot_restore_with_different_volume_mode(
+        self, pod_factory, snapshot_factory, snapshot_restore_factory
+    ):
+        """
+        Restore snapshot with volume mode - access mode combination different than the parent PVC
+
+        """
+        file_name = "fio_test"
```
petr-balogh, line 60: Can there be some conflict in file name when running the test for second time?
The same for file_name_new.

Should we add some random characters to the file name?

petr-balogh (approved):

tests/manage/pv_services/pvc_snapshot/test_restore_with_different_volume_mode.py
jilju: Entirely new PVCs are pods ill be created when the test case is running for the second time. So no conflicts will arise.

PR #3921, opened 2021-02-25
Update and split Namespace tests

fbalak:
- Split namespace feature tests to CRD and RPC
- Add CRD namespace feature tests
- Add namespace feature alerting test
- Addd `namespacestores` attribute to namespace store dict to allow usage of pre-created ns stores in buckets.


tests/manage/mcg/test_namespace_crd.py
```diff
@@ -0,0 +1,676 @@
+"""
+Tests for Namespace resources and buckets by using OpenShift CRDs only.
+These tests are valid only for OCS version lesser than 4.6 because in later
+versions are for Namespace bucket creation used CRDs instead of NooBaa RPC calls.
```
parikshithb, line 4: Need to re frame this sentence. 'valid only for OCS version 4.7 and above"

parikshithb (commented):
Looks really great. :)

parikshithb (approved):

tests/manage/mcg/test_namespace_rpc.py
```diff
@@ -36,203 +40,10 @@ class TestNamespace(MCGTest):
...
-            mcg_obj, awscli_pod, bucket_to_read=aws_target_bucket, s3_creds=s3_creds
-        )
-
-        # Compare between uploaded files and downloaded files
-        assert self.compare_dirs(awscli_pod, amount=3)
-
     @pytest.mark.polarion_id("OCS-2255")
     @tier1
-    def test_namespace_resource_creation(self, ns_resource_factory):
+    @skipif_ocs_version(">4.6")
```
MeridianExplorer, line 45: Any reason we'd like to skip those above 4.6?
I don't think it's bad to cover both RPC and CRD flows - it's free extra coverage :)

tests/manage/mcg/test_namespace_rpc.py
```diff
@@ -553,13 +368,12 @@ def test_resource_combinations_with_rgw(
     @pytest.mark.bugzilla("1900760")
     @skipif_ocs_version("<4.7")
```
MeridianExplorer, line 369:
```suggestion
    @skipif_ocs_version(">4.6")
```
If I'm not wrong?

tests/manage/mcg/test_namespace_rpc.py
```diff
@@ -27,7 +31,7 @@
 @skipif_ocs_version("<4.6")
```
MeridianExplorer, line 31: Are the test `skipifs` needed considering the entire class has a `skipif`?

tests/manage/mcg/test_namespace_crd.py
```diff
@@ -0,0 +1,676 @@
...
+            mcg_obj, awscli_pod, bucket_to_read=ns_store.uls_name, s3_creds=s3_creds
+        )
+
+        logger.info("Compare between uploaded files and downloaded files")
+        assert self.compare_dirs(awscli_pod, amount=3)
+
+    @pytest.mark.polarion_id("OCS-2293")
+    @tier4
+    @tier4a
+    @skipif_ocs_version(">4.6")
```
MeridianExplorer, line 507: I believe this might be a mistake?

tests/manage/mcg/test_namespace_rpc.py
fbalak: I think that we should not include them because we test with CRDs the exact same cases and because running all those tests again for NooBaa RPC calls can be too time consuming. CRDs are also preferred way how to manage bucket resources since 4.7.
@ebenahar @parikshithb WDYT?

parikshithb: I think the same way as Filip.  Because in later releases, there might be some major/minor changes to these API calls which might cause failures and should be analyzed and updated which is time consuming.  Apart for that, IMO, most or all the users will use either  CRDs or UI(4.8) for this feature so maintaining these tests as well seem redundant to me.

parikshithb (approved):

tests/manage/mcg/test_namespace_rpc.py
ebenahar: I don't see a reason why we want to test both CRDs and RPC with OCS > 4.6. IIUC, RPC was an intermediate API for NS bucket and now that we have CRDs, it should not be in use by customers

ebenahar (approved):

MeridianExplorer (dismissed):

fbalak:
Rebased to resolve a conflict.

parikshithb (dismissed):

mbukatov (approved):

parikshithb (approved):

ebenahar:
@fbalak @mbukatov the PR was verified before the latest changes

PR #3924, opened 2021-02-25
add capacity on internal cluster via ui (selenium)

OdedViner:

### DIFF
--- a/ocs_ci/ocs/ui/infra_ui.py
+++ b/ocs_ci/ocs/ui/infra_ui.py
@@ -14,11 +14,6 @@ class InfraUI(PageNavigator):
     """
     InfraUI Class
 
-    *Add capacity
-    *Node Replacement
-    *Device Replacement
-    *Add Node
-
     """
 
     def __init__(self, driver):
--- a/ocs_ci/ocs/ui/infra_ui.py
+++ b/ocs_ci/ocs/ui/infra_ui.py
@@ -1,0 +1,45 @@
+import logging
+
+from ocs_ci.ocs.ui.base_ui import PageNavigator
+from ocs_ci.ocs.ui.views import locators
+from ocs_ci.utility.utils import get_ocp_version
+from ocs_ci.framework import config
+from ocs_ci.ocs import constants
+
+
+logger = logging.getLogger(__name__)
+
+
+class InfraUI(PageNavigator):
+    """
+    InfraUI Class
+
+    *Add capacity
+    *Node Replacement
+    *Device Replacement
+    *Add Node
+
+    """
+
+    def __init__(self, driver):
+        super().__init__(driver)
+        ocp_version = get_ocp_version()
+        self.infra_loc = locators[ocp_version]["infra"]
+        if config.ENV_DATA["platform"].lower() == constants.VSPHERE_PLATFORM:
+            self.storage_class = "thin_sc"
+        elif config.ENV_DATA["platform"].lower() == constants.AWS_PLATFORM:
+            self.storage_class = "gp2_sc"
+
+    def add_capacity_ui(self):
+        """
+        Add Capacity via UI.
+
+        """
+        self.navigate_installed_operators_page()
+        self.do_click(self.infra_loc["ocs_operator"])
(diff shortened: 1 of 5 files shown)


ocs_ci/ocs/ui/infra_ui.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+
+    def __init__(self, driver):
+        super().__init__(driver)
+        ocp_version = get_ocp_version()
+        self.infra_loc = locators[ocp_version]["infra"]
+        if config.ENV_DATA["platform"].lower() == constants.VSPHERE_PLATFORM:
+            self.storage_class = "thin_sc"
+        elif config.ENV_DATA["platform"].lower() == constants.AWS_PLATFORM:
+            self.storage_class = "gp2_sc"
```
ebenahar, line 26: same logic is already part of https://github.com/red-hat-storage/ocs-ci/blob/f120044486631f49133c9f3a137842673d765a1c/ocs_ci/deployment/deployment.py#L675

Can you please create a function out of these checks to be used by both places?

ocs_ci/ocs/ui/infra_ui.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class InfraUI(PageNavigator):
+    """
+    InfraUI Class
+
+    """
+
+    def __init__(self, driver):
+        super().__init__(driver)
```
ebenahar, line 20: Could you please add explanation about the logic here of different classes?

tests/ui/test_add_capacity_ui.py
```diff
@@ -0,0 +1,53 @@
...
+        logging.info("Wait for osd pods to be in Running state")
+        for osd_pods in TimeoutSampler(
+            timeout=600,
+            sleep=10,
+            func=get_osd_pods,
+        ):
+            if len(osd_pods) == (osd_count + 3):
+                break
+
+        for osd_pod in osd_pods:
```
ebenahar, line 50: is there validation we can do from the UI?

ocs_ci/ocs/ui/infra_ui.py
OdedViner: yes, I will create `Common Function Class`
image

tests/ui/test_add_capacity_ui.py
hnallurv: Better to validate from UI. Newly added osd pods should be shown in UI too.

tests/ui/test_add_capacity_ui.py
```diff
@@ -0,0 +1,53 @@
...
+class TestAddCapacityUI(object):
+    """
+    Test Add Capacity on via UI
+
+    """
+
+    @ui
+    @skipif_lso
+    @skipif_bm
+    @skipif_bmpsi
```
vasukulkarni, line 27: instead of 2 above skips, you can use @cloud_platform_required (if it works only on aws, gcp ...)

vasukulkarni:
LGTM

tests/ui/test_add_capacity_ui.py
OdedViner: it is work on vmware (internal cluster) too

ocs_ci/ocs/ui/infra_ui.py
ebenahar: @OdedViner maybe add an explanation about this class in its docstring

ocs_ci/ocs/ui/infra_ui.py
```diff
@@ -0,0 +1,53 @@
...
+
+from ocs_ci.ocs.ui.base_ui import PageNavigator
+from ocs_ci.ocs.ui.views import locators
+from ocs_ci.utility.utils import get_ocp_version
+
+
+logger = logging.getLogger(__name__)
+
+
+class InfraUI(PageNavigator):
```
ebenahar, line 11: the class name does not reflect its purpose that is mentioned in the docstring below

vasukulkarni (approved):

ebenahar (approved):

Issue #3941, opened 2021-03-04
[External] Test failures during tier1 execution


sidhant-agrawal:
As observed in `RUN ID: 1614294169` tier1 execution, tests are failing due to ceph health check fail with the following error message:

```
E           ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc -n openshift-storage exec rook-ceph-tools-5cd4984bb8-ljmgf -- ceph health.
E           Error is [errno 5] error connecting to the cluster
E           command terminated with exit code 1
```

sidhant-agrawal:
Ceph health check via toolbox was working after the deployment as seen below:
```
22:57:48 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage exec rook-ceph-tools-798d89f477-llqqm -- ceph health
22:57:48 - MainThread - ocs_ci.utility.utils - INFO - Ceph cluster health is HEALTH_OK.
```
but later pod was recreated through the command started failing
```
23:05:41 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage exec rook-ceph-tools-5cd4984bb8-ljmgf -- ceph health
23:05:42 - MainThread - ocs_ci.utility.utils - WARNING - Command stderr: [errno 5] error connecting to the cluster
command terminated with exit code 1
```
I suspect this is due to addition of this [line](https://github.com/vavuthu/ocs-ci/blob/cc768f60eb1684fa14995bd69ae6539c5b4c8d66/ocs_ci/ocs/utils.py#L773) in a recent PR.
Without executing the patch command, the toolbox disappears after some time. Hence we need to remove the `return`

sidhant-agrawal:
On further analysis, I think there are two issues here:
1) toolbox pod disappearance
toolbox pod is removed from the cluster during reconciliation.
Here is simple steps to reproduce this quickly:
```
$ oc create -f ext-toolbox.yaml
deployment.apps/rook-ceph-tools created
$ oc exec rook-ceph-tools-798d89f477-lvxn6 -- ceph health
HEALTH_OK
$ oc delete pod ocs-operator-6fb7565d4c-2wkmt
pod "ocs-operator-6fb7565d4c-2wkmt" deleted
$ oc get pod | grep tool
rook-ceph-tools-798d89f477-lvxn6                1/1     Running   0          86s
$ oc get pod | grep tool
[...]

PR #3942, opened 2021-03-04
Fix for toolbox pod disappearance in External mode

sidhant-agrawal:
Fixes: #3941 

### DIFF
--- a/ocs_ci/ocs/utils.py
+++ b/ocs_ci/ocs/utils.py
@@ -770,7 +770,7 @@ def setup_ceph_toolbox(force_setup=False):
             toolbox["spec"]["template"]["spec"]["volumes"].append(ceph_volume)
             rook_toolbox = OCS(**toolbox)
             rook_toolbox.create()
-            return
+
         # for OCS >= 4.3 there is new toolbox pod deployment done here:
         # https://github.com/openshift/ocs-operator/pull/207/
         log.info("starting ceph toolbox pod")


ocs_ci/ocs/utils.py
```diff
@@ -770,7 +770,7 @@ def setup_ceph_toolbox(force_setup=False):
             toolbox["spec"]["template"]["spec"]["volumes"].append(ceph_volume)
             rook_toolbox = OCS(**toolbox)
             rook_toolbox.create()
-            return
+
```
vasukulkarni, line 773: It was created before and thats why we have return, the patch is only for non-external mode method

vavuthu: > It was created before and thats why we have return, the patch is only for non-external mode method

yes, I have added the return statement because if we patch again it with https://github.com/vavuthu/ocs-ci/blob/cc768f60eb1684fa14995bd69ae6539c5b4c8d66/ocs_ci/ocs/utils.py#L777 , it will override the external cluster settings and it ( tool box )  won't even connect to cluster

sidhant-agrawal: I have observed that if toolbox is created manually, it is removed later from the cluster during reconciliation.
afaik in previous version, running the patch once toolbox was created used to avoid pod removal.
However, as seen by Vijay and in PR job that method is no longer working for us now.

I noted my observations in https://github.com/red-hat-storage/ocs-ci/issues/3941#issuecomment-791307074
Closing this PR as the proposed solution to remove `return` doesn't work.

PR #3959, opened 2021-03-09
CNVCu0 Workload

ebenahar:
This test runs for around 5 minutes, during this time it does the following:
1) starts an FIO pod to run in the background throughout the test execution
2) starts measuring the latency in the background throughout the test execution
3) creates, deletes PVCs and pods of block and file, write IO with data validation, and measures the creation and deletion of each
4) once 5 minutes are over, it compares the p99 latency, pvc creation time and pvc deletion time to the accepted p99 latency, pvc creation and pvc deletion we define as the benchmark. In case any of them is higher, the test fails

### DIFF
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -298,6 +298,8 @@ class Pod(OCS):
         fio_filename=None,
         bs="4K",
         end_fsync=0,
+        verify=None,
+        do_verify=0,
     ):
         """
         Execute FIO on a pod
@@ -326,6 +328,9 @@ class Pod(OCS):
             bs (str): Block size, e.g. 4K
             end_fsync (int): If 1, fio will sync file contents when a write
                 stage has completed. Fio default is 0
+            verify (str): data verification type
+            do_verify (int): If 1, data verification is done, 0 otherwise
+
         """
         if not self.wl_setup_done:
             self.workload_setup(storage_type=storage_type, jobs=jobs)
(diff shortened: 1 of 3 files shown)


vasukulkarni:
@ebenahar I think you can reuse things from here with few additions instead of new: https://github.com/red-hat-storage/ocs-ci/blob/master/tests/e2e/scale/test_pv_scale_ocs_create_delete_pvcs.py#L174-L182

ocs_ci/ocs/resources/pod.py
```diff
@@ -298,6 +298,8 @@ def run_io(
         fio_filename=None,
         bs="4K",
         end_fsync=0,
+        verify=None,
```
mykaul, line 301: But we do wish to perform data verification...

ocs_ci/ocs/resources/pod.py
```diff
@@ -298,6 +298,8 @@ def run_io(
         fio_filename=None,
         bs="4K",
         end_fsync=0,
```
mykaul, line 300: If we are not using direct IO and we are not using fsync, we can't be sure data has landed fully. I suggest using direct IO.

ebenahar:
> @ebenahar I think you can reuse things from here with few additions instead of new: https://github.com/red-hat-storage/ocs-ci/blob/master/tests/e2e/scale/test_pv_scale_ocs_create_delete_pvcs.py#L174-L182

@vasukulkarni thanks for the reference, I will check that

ebenahar:
had to re-create the PR because of the branch name - https://github.com/red-hat-storage/ocs-ci/pull/3978

PR #3961, opened 2021-03-10
Changing parameter names for vsphere cleanup

vavuthu:
Changing parameter names for vsphere cleanup to align with the name convention of ocs-ci


vavuthu:
verified locally

http://pastebin.test.redhat.com/946034

petr-balogh:
Thanks you Vijay. Now we will be consistent with run-ci which is good I think.

petr-balogh (approved):

vasukulkarni (approved):

PR #3969, opened 2021-03-11
Pod reattach | adding sync after writing data

Avilir:
The main issue that this PR is fixing is, adding 'sync' command after we are writing data, so all data will be actually will be written to the PVC and not stay in the cache.

also, it replace the 'nginx' image (for the pod) with pre-loaded image with the rsync command, so we do not need to install the rsync on the pod during the test.

the test was create new SC and now it use the default SC

### DIFF
--- a/tests/e2e/performance/test_pod_reattachtime.py
+++ b/tests/e2e/performance/test_pod_reattachtime.py
@@ -19,13 +19,12 @@ class TestPVCCreationPerformance(E2ETest):
     """
 
     @pytest.fixture()
-    def base_setup(self, request, interface_iterate, storageclass_factory):
+    def base_setup(self, request, interface_iterate):
         """
         A setup phase for the test
         Args:
             interface_iterate: A fixture to iterate over ceph interfaces
-            storageclass_factory: A fixture to create everything needed for a
-                storageclass
+
         """
         self.interface = interface_iterate
 
--- a/tests/e2e/performance/test_pod_reattachtime.py
+++ b/tests/e2e/performance/test_pod_reattachtime.py
@@ -112,6 +112,8 @@ class TestPVCCreationPerformance(E2ETest):
             _ocp.exec_oc_cmd(rsh_cmd)
             rsh_cmd = f"exec {pod_name} -- cp -r {pod_path}/tmp {pod_path}/folder{x}"
             _ocp.exec_oc_cmd(rsh_cmd)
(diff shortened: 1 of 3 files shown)


wusui (approved):
Looks good

tests/e2e/performance/test_pod_reattachtime.py
```diff
@@ -19,21 +19,21 @@ class TestPVCCreationPerformance(E2ETest):
...
         """
         self.interface = interface_iterate
 
         if self.interface.lower() == "cephfs":
             self.interface = constants.CEPHFILESYSTEM
+            self.sc_obj = constants.DEFAULT_STORAGECLASS_CEPHFS
         if self.interface.lower() == "rbd":
             self.interface = constants.CEPHBLOCKPOOL
-        self.sc_obj = storageclass_factory(self.interface)
+            self.sc_obj = constants.DEFAULT_STORAGECLASS_RBD
```
keesturam, line 36: Any reason the test won't be run on cephFS?

tests/e2e/performance/test_pod_reattachtime.py
```diff
@@ -76,13 +76,13 @@ def test_pvc_reattach_time_performance(self, pvc_factory, teardown_factory):
         # Create a pod on one node
         logging.info(f"Creating Pod with pvc {pvc_obj.name} on node {node_one}")
 
-        helpers.pull_images("nginx")
+        helpers.pull_images("quay.io/ocsci/perf:latest")
```
keesturam, line 79: should this path be added to the constant file?

tests/e2e/performance/test_pod_reattachtime.py
Avilir: it will run on both (RBD and CephFS), it will not create new SC

tests/e2e/performance/test_pod_reattachtime.py
Avilir: will add it as constant.

Avilir: done

tests/e2e/performance/test_pod_reattachtime.py
Avilir: you can see it ran on RBD & CephFS in the Jenkins run results

keesturam (approved):
LGTM

ypersky1980 (approved):

PR #3970, opened 2021-03-11
Update multi-arch image pullspecs

yselkowitz:


ocs_ci/ocs/fio_artefacts.py
```diff
@@ -123,7 +123,7 @@ def get_job_dict():
     if arch.startswith("x86"):
         image = "quay.io/fbalak/fio-fedora:latest"
     else:
-        image = "quay.io/multiarch-origin-e2e/fio-fedora:latest"
+        image = "quay.io/multi-arch/fio-fedora:latest"
```
Avilir, line 126: try to use quay.io/ocsci/perf:latest
this is multi architecture image with FIO installed in it (based on alpine), and you can just consume it without checking the arch.

yselkowitz: is that a drop in replacement?  is there any reason not to use that for *all* architectures, including x86?

mbukatov:
During job validation, you need to run `/tests/manage/monitoring` tests for at least tier1 and tier4 on at least x86 and non x86 platform.

ocs_ci/ocs/fio_artefacts.py
mbukatov: Module `fio_artefacts` assumes the code is using recent fedora fio package, with http I/O engine enabled (which is enabled since [fio-3.19-3.fc32](https://bugzilla.redhat.com/show_bug.cgi?id=1836323)).

When this `if` statement was inserted about a year ago, Filip was able to build x86 container image only and the change haven't reached Fedora yet. As long as the container https://quay.io/repository/multi-arch/fio-fedora contains latest fedora rpm package, it should be fine to use single fio fedora multiarch image and drop this if statement.

We need to have full control over fio build for debugging purposes and for us to be able to patch fio if necessary (these will end up upstreamed eventually). I would like multi-arch/fio-fedora image has metadata specified properly though.

mbukatov: > try to use quay.io/ocsci/perf:latest

Not here.

mbukatov (approved):
The proposed change looks ok.

fbalak:
> During job validation, you need to run `/tests/manage/monitoring` tests for at least tier1 and tier4 on at some non x86 platform. @fbalak may be able to provide smaller more specific set of tests to try.

The fio container is used in `mcg_job_factory` fixture that utilizes s3 noobaa interface. You can try to run test https://github.com/red-hat-storage/ocs-ci/blob/master/tests/e2e/scale/noobaa/test_scale_endpoint.py#L73 as a verification step.

fbalak (approved):

vasukulkarni (approved):

vasukulkarni:
Please add verified label after running tests on required platforms

ocs_ci/ocs/fio_artefacts.py
mbukatov: I tried to use the multiarch image and I'm getting:

```
Failed to pull image "quay.io/multi-arch/fio-fedora:latest": rpc error: code = Unknown desc = Error choosing image instance: no image found in manifest list for architecture amd64, variant "", OS linux
```

So the `if` definitely needs to stay as it is for now. No changes in the pull request needed.

yselkowitz:
What's needed to get this in?

petr-balogh:
> What's needed to get this in?

Can you please rebase on latest master and force push changes here?
This will be required to trigger PR checks again to be able to even merge your PR.

Thanks

yselkowitz:
Repushed.

petr-balogh (dismissed):

petr-balogh:
@yselkowitz was some testing done as @fbalak suggested?

yselkowitz:
I am not setup to run OCS CI.  This is a simple change to update the pullspec of a multi-arch test image which I built, and anyone can verify that the new location exists and is publicly available.

petr-balogh:
@mbukatov can you please approve so we can proceed with merge?

yselkowitz:
Ping?

dahorak (approved):

PR #3971, opened 2021-03-11
Remove unsafe access to potentially-unbound cld_mgr client attributes

MeridianExplorer:
At the moment, a lack of either AWS or Azure credentials in `auth.yaml` would lead to a failure of tests that utilize `NSS_MAPPING`.
This PR fixes the issue by safely initializing the dictionary


ocs_ci/ocs/resources/mcg.py
```diff
@@ -506,26 +507,26 @@ def create_connection(self, cld_mgr, platform, conn_name=None):
                 "auth_method": "AWS_V4",
                 "endpoint": constants.MCG_NS_AWS_ENDPOINT,
                 "endpoint_type": "AWS",
-                "identity": cld_mgr.aws_client.access_key,
+                "identity": safe_get_attr(cld_mgr, "aws_client.access_key"),
```
fbalak, line 510:
```suggestion
                "identity": reduce(lambda client, access_key: getattr(client, access_key, None), "aws_client.access_key".split("."), cld_mgr),
```

fbalak (commented):
Nice use of reduce! You can also update implematation of `safe_get_attr` with something like this to drop if and try blocks:
```
reduce(lambda x, x: getattr(x, y, None), "aws_client.access_key".split("."), cld_mgr)
```
Also please make the docstring of `safe_get_attr` more verbose for better understanding (maybe add example of usage).

ocs_ci/helpers/helpers.py
```diff
@@ -2942,6 +2943,28 @@ def get_mon_pdb():
     return disruptions_allowed, min_available_mon
 
 
+def safe_get_attr(obj, attribute):
```
mbukatov, line 2946: This is clever hack :) I would suggest few tweaks:
- Rename this function to something more obvious, like: `get_attr_chain()`. Ignoring non existing attributes ans silently returning `None` is not something I would normally associate with being "safe" :)
- rename `attribute` function parameter, to `attributes` or `attr_chain` to make it obvious that you are expecting multiple attributes
- consider to change type of  the attribute parameter from string to list, that would make the API and the code of this function little bit more obvious

ocs_ci/helpers/helpers.py
```diff
@@ -2942,6 +2943,28 @@ def get_mon_pdb():
...
+    """
+    Safely retrieve object attributes
```
mbukatov, line 2948: Docstring needs little improvement:
- don't note that it's safe ...
- stress that it tries to get value of given attribute chain/list
- stress that it doesn't throw exception but it returns None in case an attribute doesn't exist

ocs_ci/helpers/helpers.py
```diff
@@ -2942,6 +2943,28 @@ def get_mon_pdb():
...
+    Returns:
+        The requested attribute if found, otherwise None
+    """
+    if obj is not None:
+        try:
+            return reduce(getattr, attribute.split("."), obj)
+        except AttributeError:
+            return None
+    else:
+        return None
```
mbukatov, line 2965: Also add few unit tests, with test cases for attribute chain/list len of 0, 1, 2 and 3 at minimum, with negative and positive cases.

ocs_ci/helpers/helpers.py
MeridianExplorer: Thank you very much for the comprehensive review and guidance!
This was originally a tiny internal function, thus the lacking documentation and naming. I assumed many things, most of them revolving around `cld_mgr` and the way its attributes are assigned ;)

Regarding the consideration to use a list instead of a string - I thought about it, and I felt that it'd be rather redundant to have to call `.split()` in every call to the function, instead of doing it internally.
Keeping the variable as a string also allows, in my opinion, easier code readability when used - i.e. -
`get_attr_chain(cld_mgr, "aws_client.access_key")`
as opposed to
`get_attr_chain(cld_mgr, ["aws_client", "access_key"])`
or
`get_attr_chain(cld_mgr, "aws_client.access_key".split("."))`

ocs_ci/helpers/helpers.py
```diff
@@ -2942,6 +2943,34 @@ def get_mon_pdb():
...
+        obj: An object
+        attr_chain (str): A string containing one attribute or several sub-attributes
+                          separated by dots
+                          (i.e. - "attr.sub_attr_a.sub_attr_b")
+
+    Returns:
+        The requested attribute if found, otherwise None
+    """
+    return reduce(
+        lambda _attr, _obj: getattr(_attr, _obj, None), attr_chain.split("."), obj
```
fbalak, line 2970:
```suggestion
        lambda _obj, _attr: getattr(_obj, _attr, None), attr_chain.split("."), obj
```

fbalak (dismissed):

ocs_ci/helpers/helpers.py
MeridianExplorer: 🙈

lsmcfadden:
What else is needed to get this one reviewed and merged?  belimele@redhat.com anyway needed from our IBM QE side?

MeridianExplorer:
@lsmcfadden - I'll need at least two people to review and approve this, preferaby @mbukatov and @fbalak that reviewed this in the past :)

ocs_ci/utility/utils.py
```diff
@@ -1736,6 +1737,7 @@ def ceph_health_check_base(namespace=None):
         boolean: True if HEALTH_OK
 
     """
+    return True
```
fbalak, line 1740: This shouldn't be here.

fbalak (changes requested):
Looks good, just the change in ceph_health_check_base shouldn't be there.

ocs_ci/utility/utils.py
MeridianExplorer: Great catch, that could be _really_ bad... It's a local workaround that shouldn't've been pushed

fbalak (approved):

ocs_ci/utility/utils.py
```diff
@@ -3159,6 +3160,34 @@ def get_ocp_upgrade_history():
...
+    get_attr_chain(object, "attr.sub_attr_a.sub_attr_b")
+
+    The function can be used to try and retrieve "sub_attribute_b" without an exception,
+    even in cases where "attr" or "sub_attr_a" might not exist.
+    In those cases, the function will return None.
+
+    Args:
+        obj: An object
+        attr_chain (str): A string containing one attribute or several sub-attributes
+                          separated by dots
```
petr-balogh, line 3180: This should be indented only once and you can put the it on one line with the line bellow as well.

petr-balogh (commented):
Otherwise it looks OK to me

petr-balogh:
Also there is a conflict, so please resolve

fbalak (approved):

lsmcfadden:
@mbukatov @ebenahar @petr-balogh @parikshithb @bernerhat can you review?

gitsridhar:
@petr-balogh can you review and approve?

lsmcfadden:
And @mbukatov  do you have cycles to review?

RazTamir:
@red-hat-storage/top-level-reviewers could you please review?
This block IBM team

ebenahar (approved):

lsmcfadden:
I also re-asked @mbukatov if there is any bandwidth to review/approve this PR since originator mention Martin as an original involved person.  There are 4 pending reviewers.  If they don't all have cycles to review, are there enough reviews/approvals to move forward?

clacroix12 (approved):

mbukatov (approved):

PR #3978, opened 2021-03-14
Cnvcu0 workload

ebenahar:
This test runs for around 5 minutes, during this time it does the following:

starts an FIO pod to run in the background throughout the test execution
starts measuring the latency in the background throughout the test execution
creates, deletes PVCs and pods of block and file, write IO with data validation, and measures the creation and deletion of each
once 5 minutes are over, it compares the p99 latency, pvc creation time and pvc deletion time to the accepted p99 latency, pvc creation and pvc deletion we define as the benchmark. In case any of them is higher, the test fails


petr-balogh:
@ebenahar any progress here?

PR #3985, opened 2021-03-15
skipif_no_kms marker and NooBaa KMS test

fbalak:

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -3121,7 +3121,9 @@ def skipif_no_kms():
     log.info("Checking if StorageCluster has configured KMS encryption")
     try:
         resource = cluster.get()["items"][0]
-        encryption = resource.get("spec").get("encryption").get("kms").get("enable")
+        encryption = (
+            resource.get("spec").get("encryption", {}).get("kms", {}).get("enable")
+        )
     except KeyError:
         log.warning("Cluster is not yet installed. Skipping skipif_no_kms check.")
         # Set encryption to True so that the test is not skipped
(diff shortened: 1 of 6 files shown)


shreekarSS:
LGTM!

ocs_ci/utility/utils.py
```diff
@@ -3106,6 +3106,35 @@ def skipif_upgraded_from(version_list):
...
+    This function evaluates the condition to skip a test if the cluster
+    doesn't have configured KMS encryption.
+
+    Return:
+        (bool): True if test needs to be skipped else False
+
+    """
+    from ocs_ci.ocs.resources import storage_cluster
+
+    cluster = storage_cluster.get_storage_cluster()
```
shyRozen, line 3120: Can also be checked against config.DEPLOYMENT.get("kms_deployment") without the further checks.

shyRozen: https://github.com/red-hat-storage/ocs-ci/blob/6adc5a4fb5139a801b5a84769fa31056170c83c3/ocs_ci/deployment/deployment.py#L657

fbalak: Right, do you think it should be checked in ocs-ci configuration instead of cluster configuration?
It would save one oc command call but this way it can be ran without specifying additional configuration file in test run to make sure that the test is not skipped if cluster have kms set. It also tests that the cluster has the correct value set in StorageCluster yaml.

ocs_ci/utility/utils.py
```diff
@@ -3106,6 +3106,35 @@ def skipif_upgraded_from(version_list):
...
+    log.info("Checking if StorageCluster has configured KMS encryption")
+    try:
+        resource = cluster.get()["items"][0]
```
shylesh, line 3123: @fbalak , can we make this kms check as an utility function something like `is_kms_enabled()` , so that even others can use it for checks.

shylesh (approved):
+1

vasukulkarni (approved):

PR #3995, opened 2021-03-16
Abandoned - WIP : Add interface and implementation for adding nodes and storage

gitsridhar:
This is an effort to standardize on an interface that can be used
by any component to implement add node and add storage to nodes to
an existing OCP/OCS cluster. This will work in conjunction with
the platform_nodes interface.

This PR introduces NodeManagement interface and also an implementation
of the same for IBM Power platform.

Signed-off-by: gitsridhar <svenkat@us.ibm.com>

### DIFF
--- a/ocs_ci/utility/nodemgmt.py
+++ b/ocs_ci/utility/nodemgmt.py
@@ -1,0 +1,106 @@
+from ocs_ci.framework import config
+from ocs_ci.utility import powernodes
+
+
+class NodeManagement(object):
+    """
+    This is a class to define node management functionality. This can
+    be used to implement adding and removing nodes dynamically and driven
+    via test-case code. If a specific deployer of OCS reqires dynamic addition
+    of nodes, this class needs to be sub-classed and required platform specific
+    functionlity needs to be implemented and consumed.
+    """
+
+    def __init__(self, consumer):
+        """
+        Class Initialization.
+
+        Initialize the class. Maintain the consumer who initialized this class for
+        logging purposes. Also, maintain a dictionary of node management objects
+        with platform string as key. This key should match the platform entry in the
+        ocs-ci configuration.
+        """
+
+        self.cls_map = {
+            "powervs": IBMPowerNodeManagement,
+        }
+        self.consumer = consumer
+
+    def get_nodemanagement_platform(self):
+        """
+        Retrieve the node management object for a given platform from config
+
+        Args: None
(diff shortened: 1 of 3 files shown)


gitsridhar:
This PR is for comments only at this time, this is not tested yet.

ocs_ci/utility/nodemgmt.py
```diff
@@ -0,0 +1,106 @@
+from ocs_ci.framework import config
+from ocs_ci.utility import powernodes
+
+
+class NodeManagement(object):
```
shylesh, line 5: Hi @gitsridhar , I might be missing something but how is this base class different from `class NodesBase`?

gitsridhar: @shylesh good question, NodesBase has methods to add and attach the node to cluster. But if you want to share or inheric logic to add the node, you cannot do it today. Your implementation of NodesBase cannot extend from another implementation of NodesBase (technically possible, but may not be relevant outside of add node - like stop/start nodes etc.). I believe this PR has a clean framework specific to adding and deleting nodes dynamically. NodesBase has no facility for deleting a node and also add and delete additional storage. I expected this comment.

shylesh: @gitsridhar , I still didn't understand completely. If you look at AWS node management in the platfrom_nodes.py: 
``` python
class NodeBase
|____class AWSNodes
         |____class AWSUPINode
```

Here based on the platform tests can instantiate AWSUPINode OR AWSIPINode and perform all the operations like add node, remove node, add_storage, remove_storage etc. This way most common things stay  in `NodeBase` , then `AWSNode` will have something specific to AWS but common to both IPI & UPI , then `AWSUPINode` will have things specific to AWS UPI.

In the same way we can have class for IBMPlatform as well , 
```python
class NodeBase
|____class IBMPlatformNode
        |____class IBMUPINode
```
I am not sure if `IBMUPI` makes sense but just an example, tests can instantiate `IBMUPI()` or `IBMIPI()` and perform operations.

am I missing something completely ?

gitsridhar:
I will submit a new PR to continue. I have removed the new interface proposed in this PR based on feedback from Shylesh.

petr-balogh:
This PR is abandoned as I see from title. New PR: https://github.com/red-hat-storage/ocs-ci/pull/4285

Closing this one.

PR #3997, opened 2021-03-16
Upgrade OCS version via UI

OdedViner:

### DIFF
--- a/conf/upgrade/upgrade_ui.yaml
+++ b/conf/upgrade/upgrade_ui.yaml
@@ -1,0 +1,3 @@
+---
+UPGRADE:
+  upgrade_ui: True
--- a/ocs_ci/ocs/ui/deployment_ui.py
+++ b/ocs_ci/ocs/ui/deployment_ui.py
@@ -8,6 +8,7 @@ from ocs_ci.utility.utils import get_ocp_version, TimeoutSampler, run_cmd
 from ocs_ci.utility import templating
 from ocs_ci.ocs.exceptions import TimeoutExpiredError
 from ocs_ci.framework import config
+from ocs_ci.ocs.ocp import get_ocs_version
 from ocs_ci.ocs import constants
 
 
@@ -321,3 +322,43 @@ class DeploymentUI(PageNavigator):
         self.install_ocs_operator()
         self.verify_ocs_operator_succeeded()
         self.install_storage_cluster()
+
+    def verify_channel(self, timeout_upgrade=600, channel="4.7"):
+        """
+        Verify Channel
+
+        timeout_install (int): Time in seconds to wait
+        channel (str): ocs y version
+
+        """
+        self.navigate_installed_operators_page()
+        self.do_click(self.dep_loc["ocs_operator_installed"])
+        self.do_click(self.dep_loc["subscription"])
+        sample = TimeoutSampler(
+            timeout=timeout_upgrade,
+            sleep=20,
+            func=self.check_element_text,
+            expected_text=f"stable-{channel}",
+        )
+        if not sample.wait_for_func_status(result=True):
+            logger.error(
+                f"OCS upgrade status is not Succeeded after {timeout_upgrade} seconds"
+            )
+            raise TimeoutExpiredError
+
+    def upgrade_ocs_ui(self, channel=None):
+        """
+        Upgrade OCS version via UI
+
+        channel (str): ocs y version
+        """
+        self.navigate_installed_operators_page()
+        self.do_click(self.dep_loc["ocs_operator_installed"])
+        self.do_click(self.dep_loc["subscription"])
+        ocs_version = get_ocs_version()
+        if ocs_version == "4.6" and channel == "4.7":
+            self.do_click(self.dep_loc["channel_4_6"])
+            self.do_click(self.dep_loc["stable-4.7"])
+        self.do_click(self.dep_loc["save"])
+        self.verify_channel(channel=channel)
+        self.verify_ocs_operator_succeeded()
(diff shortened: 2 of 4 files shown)


conf/upgrade/upgrade_ui.yaml
```diff
@@ -0,0 +1,3 @@
+---
```
ebenahar, line 1: @dahorak @petr-balogh 
will we need modifications is Jenkins?

ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -321,3 +322,43 @@ def install_ocs_ui(self):
...
+
+        channel (str): ocs y version
+        """
+        self.navigate_installed_operators_page()
+        self.do_click(self.dep_loc["ocs_operator_installed"])
+        self.do_click(self.dep_loc["subscription"])
+        ocs_version = get_ocs_version()
+        if ocs_version == "4.6" and channel == "4.7":
+            self.do_click(self.dep_loc["channel_4_6"])
+            self.do_click(self.dep_loc["stable-4.7"])
```
ebenahar, line 361: @OdedViner does this work only for 4.6 to 4.7 upgrade?

ocs_ci/ocs/ui/views.py
```diff
@@ -52,6 +52,13 @@
     "next_on_configure": ("//*[contains(text(), 'Next')]", By.XPATH),
     "create_on_review": ("//button[text()='Create']", By.XPATH),
     "search_ocs_installed": ('input[data-test-id="item-filter"]', By.CSS_SELECTOR),
+    "subscription": (
+        'a[data-test-id="horizontal-link-olm~Subscription"]',
+        By.CSS_SELECTOR,
+    ),
+    "channel_4_6": ('//button[text()="stable-4.6"]', By.XPATH),
+    "stable-4.7": ('input[data-test="stable-4.7-radio-input"]', By.CSS_SELECTOR),
```
ebenahar, line 60: same

conf/upgrade/upgrade_ui.yaml
dahorak: It is possible to supply this config file as `CLUSTER_CONF` parameter (in that case, no modification is required) ... but if we would like to have it slightly more smooth (like for UI deployment - via `UI_DEPLOY` checkbox) than yes, we will have to update the deployment and other jobs.

ocs_ci/ocs/ui/deployment_ui.py
petr-balogh: I think that those values should be added in config for particular ocs version.
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/ocs_version/ocs-4.7.yaml#L5

We already have channel defined in ocs_csv_channel.

I think that channel_4_6 should be defined in 4.6 config as this is loaded when it's start upgrade and this value can be propagated to this function.

conf/upgrade/upgrade_ui.yaml
dahorak: Created [issue#77](https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/-/issues/77) in our ocs4-jenkins repo for this.

ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -321,3 +322,43 @@ def install_ocs_ui(self):
...
+        Upgrade OCS version via UI
+
+        channel (str): ocs y version
+        """
+        self.navigate_installed_operators_page()
+        self.do_click(self.dep_loc["ocs_operator_installed"])
+        self.do_click(self.dep_loc["subscription"])
+        ocs_version = get_ocs_version()
+        if ocs_version == "4.6" and channel == "4.7":
+            self.do_click(self.dep_loc["channel_4_6"])
```
ebondare, line 360: https://bugzilla.redhat.com/show_bug.cgi?id=1843045 hasn't been fixed, so this test will be rather unstable unless you expect a possibility that the channel change button will not be there in the UI and refresh the page in this case.

ebondare (changes requested):

conf/upgrade/upgrade_ui.yaml
dahorak: Related to the `UI_UPGRADE` parameter for jenkins job - if this configuration file is used, should be together also set `UPGRADE["upgrade"] = True` parameter (used for normal upgrade)?
And in the Jenkins job - should user select both parameters `UPGRADE` and `UI_UPGRADE`  together (for UI upgrade) or only the second one?

ebenahar: @dahorak it sounds redundant but is there any other option?

dahorak: I've discussed it with Petr and we decided, that both parameters (`UPGRADE` and `UI_UPGRADE`) should be selected together and I've implemented it that way in [MR!596](https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/-/merge_requests/596) (with check which will fail the job with error message, if `UI_UPGRADE` is selected without `UPGRADE` parameters).

conf/upgrade/upgrade_ui.yaml
```diff
@@ -0,0 +1,3 @@
...
+UPGRADE:
+  upgrade_ui: True
```
dahorak, line 3: I'm thinking about renaming the `upgrade_ui` to `ui_upgrade` to have it aligned with [`ui_deployment`](https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/ui_deployment.yaml#L4), what do you thing?

PR #4007, opened 2021-03-18
Collecting more environment data for performance test report

Avilir:
This PR add more data which collected from the environment, to use in the performance report.

The data collection is done in the PASTest object at the setup of the test, so
each test that is based on this class, will have the data to in it's report.

### DIFF
--- a/ocs_ci/ocs/perftests.py
+++ b/ocs_ci/ocs/perftests.py
@@ -6,15 +6,16 @@ from elasticsearch import Elasticsearch
 
 from ocs_ci.framework.testlib import BaseTest
 
-from ocs_ci.ocs import defaults, constants
+from ocs_ci.ocs import defaults, constants, node
 from ocs_ci.framework import config
 from ocs_ci.framework.pytest_customization.marks import *  # noqa: F403
 from ocs_ci.ocs.version import get_environment_info
 from ocs_ci.ocs.resources.ocs import OCS
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs.utils import get_pod_name_by_pattern
-from ocs_ci.utility.utils import TimeoutSampler
+from ocs_ci.utility.utils import TimeoutSampler, get_running_cluster_id
 from ocs_ci.ocs.elasticsearch import elasticsearch_load
+from ocs_ci.ocs.resources import pod
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


Avilir:
verifyed locally, example of collection results can be seen at : http://10.0.78.167:9200/ripsaw-fio-fullres/_search?q=uuid:df8ccefa-5168-5c8f-a4a2-51d2a6798fe2

ocs_ci/ocs/perftests.py
```diff
@@ -41,9 +42,46 @@ def setup(self):
...
-        self.environment = get_environment_info()
         self.pod_obj = OCP(kind="pod")
 
+        # Collecting all Environment configuration Software & Hardware
+        # for the performance report.
+        self.environment = get_environment_info()
+        self.environment["clusterID"] = get_running_cluster_id()
+
+        ct_pod = pod.get_ceph_tools_pod()
+        osd_info = ct_pod.exec_ceph_cmd(ceph_cmd="ceph osd df")
```
keesturam, line 53: pls consider writing this as a separate function that returns osd related info

ocs_ci/ocs/perftests.py
```diff
@@ -41,9 +42,46 @@ def setup(self):
...
+        self.environment["osd_size"] = osd_info.get("nodes")[0].get("crush_weight")
+        self.environment["osd_num"] = len(osd_info.get("nodes"))
+        self.environment["total_capacity"] = osd_info.get("summary").get(
+            "total_kb_avail"
+        )
+        self.environment["ocs_nodes_num"] = len(node.get_ocs_nodes())
+
+        oc_cmd = OCP(namespace=defaults.ROOK_CLUSTER_NAMESPACE)
+
+        # Getting worker nodes information
```
keesturam, line 63: maybe separate functions for worker node info and maser node info too.

ocs_ci/ocs/perftests.py
Avilir: done

ocs_ci/ocs/perftests.py
Avilir: done

ocs_ci/ocs/perftests.py
```diff
@@ -41,9 +42,61 @@ def setup(self):
...
+        dictionary.
+
+        Args:
+            node_type (str): the node type to collect data about,
+              can be : master / worker - the default is master
+
+        """
+        if node_type == "master":
+            nodes = node.get_master_nodes()
+        elif node_type == "worker":
```
keesturam, line 84: This is assuming that the worker nodes are OCS nodes which might not be the case for scaled nodes. We might also have different instance types for nodes running app pods and OCS pods. So, it would be a good idea to look for OCS-labeled nodes. 
This can be handled in a subsequent PR too, as we are not using different worker nodes.

keesturam (approved):

ypersky1980 (approved):

PR #4015, opened 2021-03-19
Improve timeout/wrong-state reporting

zmc:
This PR makes some improvements to ocs-ci's error reporting; specifically, it should save some time when triaging failures related to timeouts or wrong-state errors.

If you're like me, you typically review failures inside Jenkins, viewing the console log (either for the entire build or just one pipeline stage), starting from the bottom of the traceback and scrolling up from there.

The things I wanted to improve in this PR are:
1. The lack of information in `TimeoutExpiredError`
2. Having to scroll past hundreds or thousands of lines of `describe` output before seeing the "wait for _ resource at column _ to reach..." message
3. `ResourceWrongStatusException`: Instead of only reporting `Resource <name>`, can report as much as `<kind> resource <name> in column <col> was in state <actual> but expected <expected>`. Each value is independently added if present.
4. Also fixed a bug in `ResourceWrongStatusException` related to how it is often called.

A job timing out before these changes:
```
14:07:13 - MainThread - ocs_ci.ocs.ocp - ERROR - timeout expired: Timed Out: (600,)
14:07:13 - MainThread - ocs_ci.ocs.ocp - ERROR - Wait for Pod resource  at column STATUS to reach desired condition Running failed, last actual status was ['Running', 'Terminating', 'Running', 'Pending']
14:07:13 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage describe Pod  --selector=app=rook-ceph-mon
<describe output>
14:07:14 - MainThread - ocs_ci.deployment.deployment - ERROR - Timed Out: (600,)
<skip to the middle of the py.test ERRORS section>
E               ocs_ci.ocs.exceptions.TimeoutExpiredError: Timed Out: (600,)
<skip to end of the py.test ERRORS section>
ocs_ci/utility/utils.py:926: TimeoutExpiredError
```

A job timing out with these changes:
```
16:50:20 - MainThread - ocs_ci.ocs.ocp - ERROR - timeout expired: Timed out after 900s running get("worker", True, None)
16:50:20 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc describe MachineConfigPool worker
<describe output>
16:50:20 - MainThread - ocs_ci.deployment.deployment - ERROR - Timed out after 900s running get("worker", True, None)
<skip to the middle of the py.test ERRORS section>
E               ocs_ci.ocs.exceptions.TimeoutExpiredError: Timed out after 900s running get("worker", True, None)
<skip to end of the py.test ERRORS section>
ocs_ci/utility/utils.py:928: TimeoutExpiredError
```

These builds didn't fail for exactly the same reason so it's not a perfect comparison, but you can see how we at least get some information about the failure in the exception messages. This could be further improved, but I wanted to get some feedback before I spent much more time on it.

### DIFF
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -563,8 +563,10 @@ class OCP(object):
                     actual_status = status
                     if error_condition is not None and status == error_condition:
                         raise ResourceInUnexpectedState(
-                            f"Status of '{resource_name}' at column {column}"
-                            f" is {status}."
+                            resource_name,
+                            column=column,
+                            expected=condition,
+                            got=status,
                         )
                 # More than 1 resources returned
                 elif sample.get("kind") == "List":
@@ -586,8 +588,10 @@ class OCP(object):
                                 and status == error_condition
                             ):
                                 raise ResourceInUnexpectedState(
-                                    f"Status of '{item_name}' "
-                                    f" at column {column} is {status}."
+                                    item_name,
+                                    column=column,
+                                    expected=condition,
+                                    got=status,
                                 )
                         except CommandFailed as ex:
                             log.info(
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -15,7 +15,7 @@ from ocs_ci.ocs.exceptions import (
     CommandFailed,
     NotSupportedFunctionError,
     NonUpgradedImagesFoundError,
-    ResourceInUnexpectedState,
+    ResourceWrongStatusException,
     ResourceNameNotSpecifiedException,
     TimeoutExpiredError,
 )
@@ -562,7 +562,7 @@ class OCP(object):
                     )
                     actual_status = status
                     if error_condition is not None and status == error_condition:
-                        raise ResourceInUnexpectedState(
+                        raise ResourceWrongStatusException(
                             resource_name,
                             column=column,
                             expected=condition,
@@ -587,7 +587,7 @@ class OCP(object):
                                 error_condition is not None
                                 and status == error_condition
                             ):
-                                raise ResourceInUnexpectedState(
+                                raise ResourceWrongStatusException(
                                     item_name,
                                     column=column,
                                     expected=condition,
@@ -646,7 +646,7 @@ class OCP(object):
                 )
             )
             raise (ex)
-        except ResourceInUnexpectedState:
+        except ResourceWrongStatusException:
             output = self.describe(resource_name, selector=selector)
             log.warning(
                 "Description of the resource(s) we were waiting for:\n%s", output
@@ -844,7 +844,7 @@ class OCP(object):
             )
         return False
 
-    @retry(ResourceInUnexpectedState, tries=4, delay=5, backoff=1)
+    @retry(ResourceWrongStatusException, tries=4, delay=5, backoff=1)
     def wait_for_phase(self, phase, timeout=300, sleep=5):
         """
         Wait till phase of resource is the same as required one passed in
@@ -856,7 +856,7 @@ class OCP(object):
             sleep (int): Time in seconds to sleep between attempts
 
         Raises:
-            ResourceInUnexpectedState: In case the resource is not in expected
+            ResourceWrongStatusException: In case the resource is not in expected
                 phase.
             NotSupportedFunctionError: If resource doesn't have phase!
             ResourceNameNotSpecifiedException: in case the name is not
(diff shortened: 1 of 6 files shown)


zmc:
I would like to test the `ResourceWrongStatusException` changes, but I haven't seen that error in the last couple days.

zmc:
I've been using these commits on a lot of builds since Friday and they haven't caused any problems.

ocs_ci/ocs/ocp.py
```diff
@@ -563,8 +563,10 @@ def wait_for_resource(
                     actual_status = status
                     if error_condition is not None and status == error_condition:
                         raise ResourceInUnexpectedState(
-                            f"Status of '{resource_name}' at column {column}"
-                            f" is {status}."
+                            resource_name,
```
clacroix12, line 566: We are raising `ResourceInUnexpectedState` here (and below), yet the Exception you've made changes to is `ResourceWrongStatusException`. Am I missing something here or should we be changing `ResourceInUnexpectedState` instead?

zmc: Nope, good catch! I had conflated the two. I'll push a separate commit to add similar changes to `ResourceInUnexpectedState`.

zmc: Actually, @clacroix12 - now that I look at this more closely, these exceptions really are the same concept. `ResourceInUnexpectedState` is currently just the base `Exception` by another name. I could make it inherit from `ResourceWrongStatusException`, or consolidate the codebase to use that instead of its less-informative cousin. Thoughts?

zmc: I _think_ it's best to move to a single "resource is in the wrong state/status" exception, so I'm testing a commit which does that. I can push it here if it works out and you like the idea.

clacroix12: Yeah, from the names they definitely seem like they describe the same error so I'm personally in favor of consolidating them.

Pinging @ebenahar and @petr-balogh as they are the authors of those exceptions according to `git blame` just in case they have any objections.

zmc: Okay - just pushed another commit that does that. Tested with this build https://storage-jenkins-csb-ceph.cloud.paas.psi.redhat.com/job/ocs-ci-kvm/231/ (an extremely rare green Power build!)

clacroix12 (approved):
LGTM!

zmc:
@vasukulkarni, got time for a review?

ocs_ci/utility/utils.py
```diff
@@ -882,33 +882,53 @@ class TimeoutSampler(object):
...
-        """ An args for __init__ of the timeout exception. """
+        # Arguments that will be passed to the exception
+        self.timeout_exc_args = [self.timeout]
+        try:
+            self.timeout_exc_args.append(
+                f"Timed out after {timeout}s running {self._build_call_string()}"
+            )
+        except Exception:
+            log.exception(
+                "Failed to assemble call string. Not necessarily a test failure."
```
vasukulkarni, line 919: what would cause this exception?

ocs_ci/utility/utils.py
```diff
@@ -882,33 +882,53 @@ class TimeoutSampler(object):
...
-
+        # The exception to raise
         self.timeout_exc_cls = TimeoutExpiredError
-        """ Class of exception to be raised.  """
-        self.timeout_exc_args = (self.timeout,)
-        """ An args for __init__ of the timeout exception. """
+        # Arguments that will be passed to the exception
+        self.timeout_exc_args = [self.timeout]
+        try:
+            self.timeout_exc_args.append(
```
vasukulkarni, line 914: I am just curious, how this append is not affecting the __iter__ funciton in that class.. I thought it just needs to be timeout value..(integer)

ocs_ci/utility/utils.py
zmc: Good question: hopefully nothing. But certain edge cases (like `func_kwargs` not being a `dict`) could happen. It would almost certainly be a regression introduced later on.

Often times in error reporting code, you want to be extra defensive against things like that. I wouldn't want a test run to die because someone slightly broke an error message. Make sense?

vasukulkarni: :+1:

ocs_ci/utility/utils.py
zmc: The affect is that when we `raise self.timeout_exc_cls(*self.timeout_exc_args)`, we are expanding a two-item list instead of a one-item tuple - that's it

vasukulkarni (approved):

mbukatov:
This is a good change. Most of ocs-ci modules deserves a complete redesign/rewrite though.

PR #4020, opened 2021-03-22
Install lso cluster + encryption via UI

OdedViner:
*Add Disks
*Install Local Storage Operator
*Install OCS Operator
*Install Storage Cluster (+ encryption)

### DIFF
--- a/ocs_ci/ocs/ui/deployment_ui.py
+++ b/ocs_ci/ocs/ui/deployment_ui.py
@@ -206,7 +206,7 @@ class DeploymentUI(PageNavigator):
             expected_text="Memory",
         )
         if not sample.wait_for_func_status(result=True):
-            logger.error(f"Nodes not found after 600 seconds")
+            logger.error("Nodes not found after 600 seconds")
             raise TimeoutExpiredError
 
         logger.info(f"Select {constants.LOCAL_BLOCK_RESOURCE} storage class")
--- a/ocs_ci/ocs/ui/deployment_ui.py
+++ b/ocs_ci/ocs/ui/deployment_ui.py
@@ -31,29 +31,30 @@ class DeploymentUI(PageNavigator):
         Add Disks Vmware LSO
 
         """
-        logger.info("Add RDM disk for vSphere platform")
-        platform = config.ENV_DATA.get("platform").lower()
-        lso_type = config.DEPLOYMENT.get("type")
-        if platform == constants.VSPHERE_PLATFORM:
-            from ocs_ci.deployment.vmware import VSPHEREBASE
+        if config.DEPLOYMENT.get("local_storage"):
+            logger.info("Add RDM disk for vSphere platform")
+            platform = config.ENV_DATA.get("platform").lower()
+            lso_type = config.DEPLOYMENT.get("type")
+            if platform == constants.VSPHERE_PLATFORM:
+                from ocs_ci.deployment.vmware import VSPHEREBASE
 
-            vsphere_base = VSPHEREBASE()
+                vsphere_base = VSPHEREBASE()
 
-            if lso_type == constants.RDM:
-                logger.info(f"LSO Deployment type: {constants.RDM}")
-                vsphere_base.add_rdm_disks()
+                if lso_type == constants.RDM:
+                    logger.info(f"LSO Deployment type: {constants.RDM}")
+                    vsphere_base.add_rdm_disks()
 
-            if lso_type == constants.VMDK:
-                logger.info(f"LSO Deployment type: {constants.VMDK}")
-                vsphere_base.attach_disk(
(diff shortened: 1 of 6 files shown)


ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -243,52 +263,25 @@ def configure_encryption(self):
...
-
-            logger.info(f"kms_token: {self.kms_token}")
-            self.do_send_keys(text=self.kms_token, locator=self.dep_loc["kms_token"])
-
-    def verify_ocs_operator_succeeded(self, timeout_install=300, sleep=20):
+    def verify_operator_succeeded(
+        self, operator="OpenShift Container Storage", timeout_install=300, sleep=20
+    ):
         """
         Verify OCS Installation
```
ebondare, line 282: Please update the description to "Verify Operator Installation"

ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -24,112 +25,57 @@ def __init__(self, driver):
...
-    def select_kms_token(self):
-        return self.kms_token
-
-    @select_kms_token.setter
-    def select_kms_token(self, kms_token):
-        if not isinstance(kms_token, str):
-            raise ValueError("kms_token arg must be a string")
-        self.kms_token = kms_token
+
+    def add_disks_vmware_lso(self):
```
ebenahar, line 29: I don't think you need to have this function implemented here. 
1) it is not specific for UI testing
2) it is specific for vsphere, while we already have https://github.com/red-hat-storage/ocs-ci/blob/8dcae7014a94abfc18ef8538f17c94bd98f31f2c/ocs_ci/deployment/deployment.py#L476 which is generic 

Please try to use the call to setup_local_storage() by doing the UI installation after it is called in deploy_ocs_via_operator()

yitzhak12: I think that here we don't need all that is written in the function `setup_local_storage`. We need only the part that Oded added here. However, there is no point in writing a code twice. So my suggestion is to add a new function in the file https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/helpers/lso_helpers.py - where the function `setup_local_storage` is located. We can call it something like `add_disk_for_vsphere_platform()` and copy-paste the code above to this function. Then we can use it for both in `setup_local_storage` and here.

yitzhak12 (commented):
The implementation with selenium looks really nice. Except for my one comment, the PR looks good.

ocs_ci/ocs/ui/deployment_ui.py
OdedViner: done

ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -24,112 +26,42 @@ def __init__(self, driver):
...
+            and config.ENV_DATA.get("platform").lower() == constants.VSPHERE_PLATFORM
+        ):
+            add_disk_for_vsphere_platform()
+
+    def verify_disks_lso_attached(self, timeout=600, sleep=20):
+        """
+        Verify Disks Attached
+
+        timeout (int): Time in seconds to wait
+        sleep (int): Sampling time in seconds
```
vavuthu, line 46: please align under Args: section

ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -243,52 +246,25 @@ def configure_encryption(self):
...
-
-    def verify_ocs_operator_succeeded(self, timeout_install=300, sleep=20):
+    def verify_operator_succeeded(
+        self, operator="OpenShift Container Storage", timeout_install=300, sleep=20
+    ):
         """
-        Verify OCS Installation
+        Verify Operator Installation
 
+        operator (str): type of operator
```
vavuthu, line 267: please align under Args section

ocs_ci/ocs/ui/deployment_ui.py
ebenahar: do we still need this function? can't we directly call `add_disk_for_vsphere_platform()` from whatever needed under the same condition of LSO?

ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -292,7 +284,8 @@ def install_ocs_ui(self):
         Install OCS via UI
 
         """
-        self.add_disks_lso()
+        if config.DEPLOYMENT.get("local_storage"):
+            add_disk_for_vsphere_platform()
```
ebenahar, line 288: I would add another condition for checking if the platform is vsphere. If not, skip the test

ebenahar (approved):

dahorak (approved):

PR #4021, opened 2021-03-22
Replace BUCKET_MAP call with bucket_factory() in test_bucket_deletion

MeridianExplorer:
We've seen a failure of `test_bucket_delete_with_objects` in one of our recent runs.
Upon closer inspection, it was discovered that the test uses `BUCKET_MAP` to create buckets, and because of that, *does not wait for them to bind*.
This resulted in us trying to sync objects to buckets that were not yet ready, and running into a `NoSuchBucket` error.
I replaced the logic with a call to `bucket_factory()`, that by default waits for the buckets to become healthy.


fbalak (approved):

bernerhat (approved):
Great catch!

ebenahar (approved):

Issue #4000, opened 2021-03-17
get_osd_size() failed to run on Vmware-LSO


Avilir:
while trying to use the function ocs_ci.ocs.resources.storage_cluster.get_osd_size()  on vmware-lso cluster i am getting :
self.environment["osd_size"] =  get_osd_size()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def get_osd_size():
        """
        Get osd size from Storage cluster
    
        Returns:
            int: osd size
    
        """
        sc = get_storage_cluster()
>       return int(
            sc.get()
            .get("items")[0]
            .get("spec")
            .get("storageDeviceSets")[0]
            .get("dataPVCTemplate")
            .get("spec")
            .get("resources")
            .get("requests")
            .get("storage")[:-2]
        )
E       ValueError: invalid literal for int() with base 10: ''

PR #4023, opened 2021-03-22
Fix validate_ocs_pods_on_pvc for UI deployment

petr-balogh:
Fixes: #4022

Signed-off-by: Petr Balogh <pbalogh@redhat.com>

PR #4028, opened 2021-03-23
Fixing get_osd_size for LSO Direct I/O cluster

vavuthu:
Fixes: #4000 

For LSO Direct I/O cluster, storage value is "1" unlike other cluster ( 100Gi ).

### DIFF
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -9,14 +9,14 @@ from jsonschema import validate
 
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants, defaults, ocp
-from ocs_ci.ocs.exceptions import ResourceNotFoundError
+from ocs_ci.ocs.exceptions import ResourceNotFoundError, UnsupportedFeatureError
 from ocs_ci.ocs.ocp import get_images, OCP
 from ocs_ci.ocs.resources.ocs import get_ocs_csv
 from ocs_ci.ocs.resources.pod import get_pods_having_label, get_osd_pods
-from ocs_ci.ocs.resources.pvc import get_all_pvc_objs
+from ocs_ci.ocs.resources.pvc import get_all_pvc_objs, get_deviceset_pvcs
 from ocs_ci.utility import localstorage, utils, templating, kms as KMS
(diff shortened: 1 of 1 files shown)


vavuthu:
Verified locally with LSO Direct I/O cluster and normal cluster

LSO Direct I/O:
```
>>> from ocs_ci.ocs.resources.storage_cluster import get_osd_size
>>> get_osd_size()
'1490'
>>> 
```

Normal cluster:
```
>>> from ocs_ci.ocs.resources.storage_cluster import get_osd_size
>>> get_osd_size()
'512'
>>> 

```

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -566,8 +565,13 @@ def get_osd_size():
         .get("spec")
         .get("resources")
         .get("requests")
-        .get("storage")[:-2]
+        .get("storage")
     )
+    if not size.isdigit():
```
petr-balogh, line 570: @vavuthu wasn't that fixed here:  https://github.com/red-hat-storage/ocs-ci/pull/4023

As you are talking about LSO deployment it will already go to first `if config.DEPLOYMENT.get("local_storage"):`

But when I am looking to more details to usage of get_deviceset_pvcs then we can improve and consolidate it to something like:

When this is not the case only of UI LSO deployment we should consolidate this code. 

We need to return int, so better to always convert string to int in every return:

```
    sc = get_storage_cluster()
    size = (
        sc.get()
        .get("items")[0]
        .get("spec")
	@@ -566,8 +565,13 @@ def get_osd_size():
        .get("spec")
        .get("resources")
        .get("requests")
        .get("storage")
    )
    if size.isdigit() or config.DEPLOYMENT.get("local_storage"):
        # In the case of UI deployment of LSO cluster, the value in StorageCluster CR
        # is set to 1, so we can not take OSD size from there. For LSO we will return
        # the size from PVC.
        pvc = get_deviceset_pvcs()[0]
        return int(pvc.get()["status"]["capacity"]["storage"][:-2])
    
    else:
        return int(size[:-2])
```

vavuthu: Done

petr-balogh (approved):

dahorak (approved):

Issue #3895, opened 2021-02-23
test_git_clone fails very often to install git package


petr-balogh:
https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/41/testReport/tests.e2e.workloads.test_jenkins_simulation/TestJenkinsSimulation/test_git_clone_RBD_/

```
  @acceptance
    @workloads
    def test_git_clone(self, pod):
        """
        git clones a large repository
        """
>       pod.run_git_clone()

tests/e2e/workloads/test_jenkins_simulation.py:26: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ocs_ci/ocs/resources/pod.py:388: in run_git_clone
    assert wl.setup(), "Setup up for git failed"
ocs_ci/ocs/workload.py:66: in setup
    return self.work_load_mod.setup(**setup_conf)
ocs_ci/utility/workloads/jenkins.py:43: in setup
    return io_pod.exec_cmd_on_pod(cmd, out_yaml_format=False)
ocs_ci/ocs/resources/pod.py:171: in exec_cmd_on_pod
    rsh_cmd, out_yaml_format, secrets=secrets, timeout=timeout, **kwargs
ocs_ci/ocs/ocp.py:144: in exec_oc_cmd
    **kwargs,
ocs_ci/utility/utils.py:453: in run_cmd
    completed_process = exec_cmd(cmd, secrets, timeout, ignore_error, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cmd = ['oc', '-n', 'namespace-test-67b5af834b9040959c2d00bdf', 'rsh', 'pod-test-rbd-cdcc52e812234e9b8da39464629', 'apt-get', ...]
secrets = None, timeout = 600, ignore_error = False, kwargs = {}
masked_cmd = 'oc -n namespace-test-67b5af834b9040959c2d00bdf rsh pod-test-rbd-cdcc52e812234e9b8da39464629 apt-get -y install git'
[...]

PR #4031, opened 2021-03-23
Make test_jenkins_simulation work on all architectures

### LINKED ISSUE #4027: arch-specific acceptance test: test_jenkins_simulation
This test always fails on non-amd64 platforms because `quay.io/ocsci/git-alpine-image` isn't multi-arch. We should either:

* Use a different image
* Build the image on all arches
* Exclude the test from platforms it won't run on
* Remove the acceptance marker from the test

zmc:
This fixes #4027. I just had tests on ppc64le pass.

### DIFF
--- a/ocs_ci/utility/workloads/helpers.py
+++ b/ocs_ci/utility/workloads/helpers.py
@@ -7,7 +7,7 @@ from ocs_ci.ocs import constants
 
 log = logging.getLogger(__name__)
 
-DISTROS = {"Debian": "apt-get", "RHEL": "yum"}
+DISTROS = {"Debian": "apt-get", "RHEL": "yum", "Alpine": "apk"}
 
 
 def find_distro(io_pod):
--- a/ocs_ci/utility/workloads/jenkins.py
+++ b/ocs_ci/utility/workloads/jenkins.py
@@ -39,7 +39,10 @@ def setup(**kwargs):
         cmd = f"{pkg_mgr} update"
         io_pod.exec_cmd_on_pod(cmd, out_yaml_format=False)
 
-    cmd = f"{pkg_mgr} -y install git"
+    if distro == "Alpine":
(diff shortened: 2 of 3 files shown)


zmc:
Before: https://storage-jenkins-csb-ceph.cloud.paas.psi.redhat.com/job/ocs-ci-kvm/220/testReport/
After: https://storage-jenkins-csb-ceph.cloud.paas.psi.redhat.com/job/ocs-ci-kvm/229/testReport/

vasukulkarni:
Prasad was planning to update the image in quay.io/ocsci, but I think we can consume from here until its done, The issue is fio install on x86 that keeps failing..

vasukulkarni (approved):

zmc:
The nice thing about this is it removes the need to maintain the image yourselves. Where can I find out more about the fio issue?

vasukulkarni:
> The nice thing about this is it removes the need to maintain the image yourselves. Where can I find out more about the fio issue?

Sorry the issue is with the git, https://github.com/red-hat-storage/ocs-ci/issues/3895 , The PR that closed it preinstalled git to avoid this one, I believe we will start seeing 3895 in x86..

zmc:
ah, thanks for linking that. I was unaware of that history. Well, this will use the Alpine image and not Debian, so it may just not fail in the same way. Try:
`podman run --rm mirror.gcr.io/library/alpine sh -c "apk add git fio && which git && which fio"`

I haven't seen it fail, and it takes only 2-10s where I've run it (my amd64 laptop and a ppc64le test machine)

vasukulkarni:
> `podman run --rm mirror.gcr.io/library/alpine sh -c "apk add git fio && which git && which fio"`

can confirm it works and its fast!, so lets see :+1:

clacroix12 (approved):

Issue #3999, opened 2021-03-17
ocs-ci does not upgrade the openshift client after an update of the config.yaml


Sarah-Kriesch:
I had an installation of OCP and OCS with the version 4.6 before the installation of 4.7.
I run ocs-ci with both versions. After an update of the config.yaml with the version 4.7, I expect that the openshift client will be used with 4.7 (or 4.7.1 in my case) will be used.

But ocs-ci is looking only for any existing openshift client and is using this one. Therefore, the openshift client with the version 4.6.11 is used in my case, what is wrong.

ocs-ci should check the existing version of the openshift client before using it. If that is the wrong one (not matching with the config.yaml), then it should be updated.

petr-balogh:
Currently the client is downloaded during the installation deployment.
https://github.com/red-hat-storage/ocs-ci/blob/9f2a231a3fdd8ae677a0b1b64620dcfb4964c470/tests/conftest.py#L1118-L1122

Or when it does not exist.

You can delete the client from bin folder so then it should download new client next run-ci execution.

But you are right that we could have better check for particular version of the client. So if someone have still 4.6 and running 4.7 that it will rather delete that binary and download new one matching the OCP version passed to ocs-ci.

@clacroix12 can you take a look at this please?

clacroix12:
@petr-balogh @Sarah-Kriesch Just wanted to update on this. I have a fix locally but I still need to do a bit of testing to make sure the scenario you described is covered correctly. I'll have a PR up once I verify that.

Sarah-Kriesch:
Thank you, @clacroix12! One hint for you. That was one of my last feature requests / bug reports as an IBMer. My colleagues have to pick my tasks up, because I can not live without own open-source contributions. I will leave IBM.

Good luck and continue with your good work!

PR #4034, opened 2021-03-23
Force client download if existing oc version does not match config

clacroix12:
Fixes: #3999 

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -696,15 +696,25 @@ def get_openshift_client(
 
     """
     version = version or config.RUN["client_version"]
+    version = expose_ocp_version(version)
     bin_dir = os.path.expanduser(bin_dir or config.RUN["bin_dir"])
     client_binary_path = os.path.join(bin_dir, "oc")
     if os.path.isfile(client_binary_path) and force_download:
         delete_file(client_binary_path)
     if os.path.isfile(client_binary_path):
-        log.debug(f"Client exists ({client_binary_path}), skipping download.")
-        # TODO: check client version
+        current_client_version = get_client_version()
+        if current_client_version != version:
+            log.info(
+                f"Existing client version ({current_client_version}) does not match "
+                f"configured version ({version}), removing file: {client_binary_path}"
+            )
+            delete_file(client_binary_path)
+    if os.path.isfile(client_binary_path):
+        log.debug(
+            f"Client exists ({client_binary_path}) and matches configured version, "
+            f"skipping download."
+        )
     else:
-        version = expose_ocp_version(version)
         log.info(f"Downloading openshift client ({version}).")
         prepare_bin_dir()
         # record current working directory and switch to BIN_DIR
@@ -3197,3 +3207,18 @@ def get_default_if_keyval_empty(dictionary, key, default_val):
     if not dictionary.get(key):
         return default_val
     return dictionary.get(key)
+
+
+def get_client_version():
+    """
+    Get version reported by `oc version`.
+
+    Returns:
+        str: version reported by `oc version`. None if oc is not on the path.
+
+    """
+    if which("oc"):
+        cmd = "oc version -o json"
+        resp = exec_cmd(cmd, ignore_error=True)
+        output = json.loads(resp.stdout.decode())
+        return output["releaseClientVersion"]
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -699,22 +699,28 @@ def get_openshift_client(
     version = expose_ocp_version(version)
     bin_dir = os.path.expanduser(bin_dir or config.RUN["bin_dir"])
     client_binary_path = os.path.join(bin_dir, "oc")
-    if os.path.isfile(client_binary_path) and force_download:
-        delete_file(client_binary_path)
-    if os.path.isfile(client_binary_path):
+    download_client = True
+
+    if force_download:
+        log.info("Forcing client download.")
+    elif os.path.isfile(client_binary_path):
         current_client_version = get_client_version()
         if current_client_version != version:
             log.info(
                 f"Existing client version ({current_client_version}) does not match "
-                f"configured version ({version}), removing file: {client_binary_path}"
+                f"configured version ({version})."
             )
+        else:
+            log.debug(
+                f"Client exists ({client_binary_path}) and matches configured version, "
+                f"skipping download."
+            )
+            download_client = False
+
+    if download_client:
+        if os.path.isfile(client_binary_path):
+            log.info(f"Deleting client that exists at {client_binary_path}")
             delete_file(client_binary_path)
-    if os.path.isfile(client_binary_path):
-        log.debug(
-            f"Client exists ({client_binary_path}) and matches configured version, "
-            f"skipping download."
-        )
-    else:
         log.info(f"Downloading openshift client ({version}).")
         prepare_bin_dir()
         # record current working directory and switch to BIN_DIR


clacroix12:
From a local test where the existing oc version was 4.7 but the configured OCP version was 4.8. This is from a deployment verification test (`-m deployment` without `--deploy`):
```
14:12:34 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc version
14:12:35 - MainThread - ocs_ci.utility.oc_utils - INFO - full version string: 4.7.0-0.nightly-2021-03-22-025559
Server Version
14:12:35 - MainThread - ocs_ci.utility.oc_utils - INFO - 4.7.0-0.nightly
14:12:36 - MainThread - ocs_ci.utility.utils - INFO - Downloading openshift client (4.8.0-0.nightly-2021-03-22-104536).
```

I have also tested these functions locally to ensure that they are working as expected. I might add some libtests around `compare_client_versions` but I will need to mock out `get_client_version` since the output of that could break what tests are expecting.

ocs_ci/utility/oc_utils.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    Get the major.minor version of the openshift client
+
+    Returns:
+        str: major.minor oc version
+
+    """
+    # get version info from oc
+    cmd = "oc version"
+    resp = exec_cmd(cmd, ignore_error=True)
```
vasukulkarni, line 20: I believe we shouldn't ignore error, the resp might not be accurate for further processing.

clacroix12: We are ignoring the error because `oc version` will return an error code if the server version is not returned properly:

```
$oc version
Client Version: 4.8.0-0.nightly-2021-03-22-104536
error: You must be logged in to the server (Unauthorized)
```

In this case, I still want to be able to parse out the client version so I am ignoring the error. We could do this differently though. We could except the `CommandFailed` and still parse the stdout if it exists but I think the result would be similar to passing `ignore_error=True`.

vasukulkarni: I think its fine in that case

vasukulkarni (dismissed):

jimdevops19 (approved):
LGTM

ocs_ci/utility/oc_utils.py
```diff
@@ -0,0 +1,52 @@
...
+def get_client_version():
+    """
+    Get the major.minor version of the openshift client
+
+    Returns:
+        str: major.minor oc version
+
+    """
+    # get version info from oc
+    cmd = "oc version"
```
dahorak, line 19: What about using `oc version -o json` or `oc version -o yaml`, wouldn't that be slightly nicer for processing and maybe more stable format (you will not rely on the order of the output lines etc.)? I see, that processing of the plain `oc version` is quite easy... so maybe it is not worth that.
```json
$ oc version -o json
{
  "clientVersion": {
    "major": "",
    "minor": "",
    "gitVersion": "openshift-clients-4.6.0-202006250705.p0-168-g02c110006",
    "gitCommit": "02c110006bfef4ba53fa5042bb9eae170dd3dc1c",
    "gitTreeState": "clean",
    "buildDate": "2020-12-16T08:02:28Z",
    "goVersion": "go1.15.2",
    "compiler": "gc",
    "platform": "linux/amd64"
  },
  "serverVersion": {
    "major": "1",
    "minor": "20",
    "gitVersion": "v1.20.0+bafe72f",
    "gitCommit": "bafe72fb05eddc8246040b9945ec242b9f805935",
    "gitTreeState": "clean",
    "buildDate": "2021-03-14T16:01:39Z",
    "goVersion": "go1.15.7",
    "compiler": "gc",
    "platform": "linux/amd64"
  },
  "releaseClientVersion": "4.6.9",
  "openshiftVersion": "4.7.0-0.nightly-2021-03-25-035231"
}
```

ocs_ci/utility/oc_utils.py
dahorak: How it will behave, if `oc` client doesn't exist in the time of execution of this function?

ocs_ci/utility/oc_utils.py
```diff
@@ -0,0 +1,52 @@
...
+    resp = exec_cmd(cmd, ignore_error=True)
+    stdout = resp.stdout.decode()
+    # get the full version from the output
+    full_version_str = stdout.split(":")[1].strip()
+    log.debug("full version string: %s", full_version_str)
+    # remove the date from the version string
+    version = "-".join(full_version_str.split("-")[:2])
```
dahorak, line 26: How it will behave for released version? e.g.:
```
$ oc version
Client Version: 4.6.9
Server Version: 4.7.0-0.nightly-2021-03-25-035231
Kubernetes Version: v1.20.0+bafe72f
```

clacroix12: I tested this case and because the way I'm using slices still works for the version returned for GA versions. However, I'm going to change this to use the json output like you suggested so it will most likely be more consistent anyways.

ocs_ci/utility/oc_utils.py
clacroix12: @dahorak good point, I didn't actually test without a client present I will have to consider that and make sure we continue gracefully if it doesn't exist at all.

clacroix12:
@dahorak I took a different approach with this based on your input. We are now handling this within the `get_openshift_client` function similarly to how we treat the `force_download` argument. If a client exists and the version doesn't match what is configured we will remove the existing client so a new one will be downloaded.

I realized we already have [expose_ocp_version](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L551-L581) which will translate what we have in our config to a value and format that will match what `oc version` returns.

ocs_ci/utility/oc_utils.py
clacroix12: I am now using `-o json`

ocs_ci/utility/utils.py
```diff
@@ -696,15 +696,25 @@ def get_openshift_client(
...
+            log.info(
+                f"Existing client version ({current_client_version}) does not match "
+                f"configured version ({version}), removing file: {client_binary_path}"
+            )
+            delete_file(client_binary_path)
+    if os.path.isfile(client_binary_path):
+        log.debug(
+            f"Client exists ({client_binary_path}) and matches configured version, "
+            f"skipping download."
+        )
```
vavuthu, line 716: can we move to else of https://github.com/red-hat-storage/ocs-ci/pull/4034/files#diff-c35e2a9e1e1a4756ebf78be919131918bb3242c321053979545deab4e487025aR706

clacroix12: This `if` is what the following `else` is based off of (where we download the client). If I move this to the `else` off of the new `if` I have added then this entire block needs to be reworked. I think it's worth doing because this is logically a bit confusing as it is. I'm not quite sure what the best way to do it is yet though. I'll spend some time on it today.

clacroix12: @vavuthu I moved that part to where you suggested but that also came with other changes. I've moved the `delete_file` call to right before we actually download the client (if it exists) and broke that portion out of the `else` and opted for a boolean `download_client`. This is set to `True` by default and is only set to `False` if `force_download` isn't set, the client exists at the path, and the version matches what is set in the config.

dahorak (approved):

petr-balogh (approved):

zmc:
Just want to mention that this entirely broke ppc64le for us. I don't have a fix yet.

clacroix12:
@zmc can you explain what broke / create an issue for it and ping me there? I'm curious to know why that particular scenario broke due to these changes.

zmc:
@clacroix12 yes, apologies for not including all this earlier: ocs-ci only knows how to download x86 `oc` binaries. In the ppc setup we use, OCP is deployed independently of ocs-ci and we get our own binary from that process.

Then ocs-ci comes in, detects a version mismatch (which in this case isn't always a reflection of reality), deletes the `oc` we have, downloads an x86 version, and then fails when it tries to use it.

I wonder if part of a reasonable fix would be that if a new client is needed, the framework at least tries to download/verify the new one before removing the existing one - which in very many cases will work just fine regardless of a version mismatch.

PR #4063, opened 2021-03-29
Adding function to find the default pull replica count

### LINKED ISSUE #4053: fio benchmark try to use more capacity then it have
for some reason (under investigatio) the test_fio_benchmark is trying to use more capacity then the storage have.

Avilir:
This PR add a function which find the default pool replica count from the `ceph status` command.

it also change the get_ceph_capacith() function, to use the new replica count function, to fix #4053 

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -529,6 +529,8 @@ class CephCluster(object):
             if mod["name"] == "localpool":
                 return mod.get("module_options").get("num_rep").get("default_value")
         logger.error("Replica count number did not found !")
+        # if there is an error in the output of `ceph status` command and localpool
+        # module does not exist, return 0 as number of replica.
         return 0
 
     def get_ceph_capacity(self):
@@ -553,6 +555,8 @@ class CephCluster(object):
 
             return usable_capacity
         else:
+            # if the replica number is 0, usable capacity can not be calculate
+            # so, return 0 as usable capacity.
             return 0
 
     def get_ceph_cluster_iops(self):
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -514,31 +514,46 @@ class CephCluster(object):
             cmd += f" -f {format}"
         return self.toolbox.exec_cmd_on_pod(cmd, out_yaml_format=False)
 
+    def get_ceph_default_replica(self):
+        """
+        The function return the default replica count in the system,
+        taken from 'ceph status'. in case no parameter found, return '0'.
+
+        Returns:
+             int : the default replica count - 0 if not found.
+        """
+        ceph_pod = pod.get_ceph_tools_pod()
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/cluster.py
```diff
@@ -514,31 +514,46 @@ def get_ceph_status(self, format=None):
...
+             int : the default replica count - 0 if not found.
+        """
+        ceph_pod = pod.get_ceph_tools_pod()
+        ceph_status = ceph_pod.exec_ceph_cmd(ceph_cmd="ceph status")
+        av_mod = ceph_status.get("mgrmap").get("available_modules")
+        for mod in av_mod:
+            if mod["name"] == "localpool":
+                return mod.get("module_options").get("num_rep").get("default_value")
+        logger.error("Replica count number did not found !")
+        return 0
```
keesturam, line 532: isn't the replica  '1'  if there is no replica set.

ocs_ci/ocs/cluster.py
```diff
@@ -514,31 +514,46 @@ def get_ceph_status(self, format=None):
...
+            logger.info(f"Number of replica : {replica}")
+            ceph_pod = pod.get_ceph_tools_pod()
+            ceph_status = ceph_pod.exec_ceph_cmd(ceph_cmd="ceph df")
+            usable_capacity = (
+                int(ceph_status["stats"]["total_bytes"]) / replica / constant.GB
+            )
 
-        return usable_capacity
+            return usable_capacity
+        else:
```
keesturam, line 555: This code path is never going to be executed. if the replica is 0 there is no point in going ahead. should we simply raise an exception here?

ocs_ci/ocs/cluster.py
Avilir: we have no deployment with replica-1, so i do not know what this code will return on system with replica-1
this `return 0` is in case there is a problem and the num_rep parameter is not define.

ocs_ci/ocs/cluster.py
Avilir: this is an error handling instead of raising an exception for better understanding the failure reason IMO.

ocs_ci/ocs/cluster.py
```diff
@@ -514,31 +514,50 @@ def get_ceph_status(self, format=None):
...
-        usable_capacity = (
-            int(ceph_status["stats"]["total_bytes"]) / replica / constant.GB
-        )
+        replica = int(self.get_ceph_default_replica())
+        if replica > 0:
+            logger.info(f"Number of replica : {replica}")
+            ceph_pod = pod.get_ceph_tools_pod()
+            ceph_status = ceph_pod.exec_ceph_cmd(ceph_cmd="ceph df")
+            usable_capacity = (
+                int(ceph_status["stats"]["total_bytes"]) / replica / constant.GB
```
vasukulkarni, line 553: I think we should file an issue to improve on this one and use the function ceph provides to capture the real usable capacity as it will be more accurate and takes into considerations various profiles.

vasukulkarni (approved):

PR #4075, opened 2021-03-30
Save Noobaa DB migration pods logs during OCS upgrade

ebenahar:
Noobaa DB replacement takes place as part of the OCS upgrade from 4.6 to 4.7.
The DB migration job, done as part of the upgrade has a few retries so it can fail up to 5 times and the upgrade will still be considered a success.
Currently, there is no way to know if there were such failed attempts, as the DB migration job pods get deleted once the DB migration is done. 

For that, this PR adds the capability to save the pods logs, from those DB migration job pods

In addition to that, there is a pattern to look for in the logs, which indicates a failed attempt. I added a search for that in the logs and we will be notified in the console log where to look for

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -32,6 +32,8 @@ RUN:
   # This config file disables scale app pods to use OCS workers
   use_ocs_worker_for_scale: False
   load_status: None
+  # This config parameter is not supposed to be changed by the user but by the test fixture itself. It is used for saving the pods' logs during the test execution
+  save_live_pod_logs: False
 
 # In this section we are storing all deployment related configuration but not
 # the environment related data as those are defined in ENV_DATA section.
--- a/ocs_ci/ocs/utils.py
+++ b/ocs_ci/ocs/utils.py
@@ -7,6 +7,8 @@ import re
 import time
 import traceback
 from subprocess import TimeoutExpired
+import threading
+from uuid import uuid4
 
 import yaml
 from gevent import sleep
@@ -709,15 +711,16 @@ def get_pod_name_by_pattern(
     namespace = namespace if namespace else ocsci_config.ENV_DATA["cluster_namespace"]
     ocp_obj = OCP(kind="pod", namespace=namespace)
     pod_names = ocp_obj.exec_oc_cmd("get pods -o name", out_yaml_format=False)
-    pod_names = pod_names.split("\n")
     pod_list = []
-    for name in pod_names:
-        if filter is not None and re.search(filter, name):
-            log.info(f"Pod name filtered {name}")
-        elif re.search(pattern, name):
-            (_, name) = name.split("/")
-            log.info(f"pod name match found appending {name}")
-            pod_list.append(name)
+    if pod_names:
+        pod_names = pod_names.split("\n")
+        for name in pod_names:
+            if filter is not None and re.search(filter, name):
+                log.info(f"Pod name filtered {name}")
+            elif re.search(pattern, name):
+                (_, name) = name.split("/")
+                log.info(f"pod name match found appending {name}")
+                pod_list.append(name)
     return pod_list
 
 
@@ -890,6 +893,48 @@ def collect_noobaa_db_dump(log_dir_path):
     )
 
 
+def save_pods_logs_to_file(
+    pod_patterns_containers, dir_name_prefix, patterns_to_log=None
+):
+    """
+    Save live logs of requested pods
+
+    Args:
+        pod_patterns_containers (dict): The patterns of the pods that
+            their logs should be saved as keys and the container names as values
+        dir_name_prefix (str): A suffix for the logs directory name. Should be the test case name
+        patterns_to_log (list): A pattern in the logs to look for and log
+
+    """
+    log_dir_path = os.path.join(
+        os.path.expanduser(ocsci_config.RUN["log_dir"]),
+        f"live_logs_{ocsci_config.RUN['run_id']}",
+    )
+    pods_containers = dict()
+    for pattern, containers in pod_patterns_containers.items():
+        pod_names = get_pod_name_by_pattern(pattern=pattern)
+        for pod_name in pod_names:
+            pods_containers.update({pod_name: containers})
+    live_logs_dir_path = os.path.join(
+        log_dir_path, f"{dir_name_prefix}_{uuid4().hex[:8]}"
+    )
+    create_directory_path(live_logs_dir_path)
+
+    if len(pods_containers) > 0:
+        for pod, container in pods_containers.items():
+            cmd = f"oc -n openshift-storage logs {pod} -c {container}"
+            logs_file = os.path.join(live_logs_dir_path, f"{pod}")
+            out = run_cmd(cmd)
+            if patterns_to_log:
+                for pattern_to_log in patterns_to_log:
+                    if pattern_to_log in out:
+                        log.info(
+                            f"The requested pattern {pattern_to_log} was found and it will be saved in {logs_file}"
+                        )
+            with open(logs_file, "w") as f:
+                f.write(out)
+
+
 def collect_ocs_logs(dir_name, ocp=True, ocs=True, mcg=False, status_failure=True):
     """
     Collects OCS logs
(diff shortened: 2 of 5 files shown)


ocs_ci/ocs/utils.py
```diff
@@ -890,6 +892,45 @@ def collect_noobaa_db_dump(log_dir_path):
...
+    )
+    pods_containers = dict()
+    for pattern, containers in pod_patterns_containers.items():
+        pod_names = get_pod_name_by_pattern(pattern=pattern)
+        for pod_name in pod_names:
+            pods_containers.update({pod_name: containers})
+
+    now = datetime.datetime.now()
+    dt_string = now.strftime("%d/%m/%Y %H:%M:%S")
+    live_logs_dir_path = os.path.join(log_dir_path, f"live_logs_{dt_string}")
```
petr-balogh, line 917: and one more thing for improvement. live_logs_31/ can you replace it maybe with the name of the current test case?

There doesn't have to be live_logs in the name again if you already have it in parent folder live_logs_1617148726 .

The path now looks like:
`live_logs_1617148726/live_logs_31/03/2021 00:09:34/`


So I think it should be something like:
`live_logs_1617148726/tc_name/210331_00:09:34/`

Not sure if we need to have date and too many folders

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -32,6 +32,7 @@ RUN:
   # This config file disables scale app pods to use OCS workers
   use_ocs_worker_for_scale: False
   load_status: None
+  save_logs: False
```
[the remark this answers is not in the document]
ebenahar: @petr-balogh although this key (`save_logs`) is talking about logs, I think we still should consider it as a RUN parameter since it's not being passed as an env param, but rather defined in the test itself, and is being dynamically changed and tracked by a background thread. 

We can maybe change the param name, so it will reflect better the purpose of it. I don't have a better naming though.

ocs_ci/ocs/utils.py
ebenahar: done

ebenahar:
Verified:
https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/1749
http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/jnk-pr4075-b442/jnk-pr4075-b442_20210331T204006/logs/live_logs_1617226307/

Screenshot from 2021-04-01 01-02-07

ocs_ci/ocs/utils.py
```diff
@@ -1003,6 +1048,41 @@ def collect_prometheus_metrics(
...
+
+    def save_logs():
+        while ocsci_config.RUN["save_logs"]:
+            time.sleep(10)
+            if ocsci_config.RUN["save_logs"]:
+                try:
+                    save_pods_logs_to_file(
+                        pods_containers_dict, tc_name, patterns_to_log
+                    )
+                except Exception:
```
petr-balogh, line 1079: This sounds like:
https://realpython.com/the-most-diabolical-python-antipattern/

please at least write the exception to debug log level.

ocs_ci/ocs/utils.py
```diff
@@ -1003,6 +1048,41 @@ def collect_prometheus_metrics(
...
+        ocsci_config.RUN["save_logs"] = False
+        if thread:
+            thread.join()
+
+    request.addfinalizer(finalizer)
+
+    ocsci_config.RUN["save_logs"] = True
+
+    def save_logs():
+        while ocsci_config.RUN["save_logs"]:
```
petr-balogh, line 1072: change it here to RUN.get("save_live_pod_logs")

ocs_ci/ocs/utils.py
```diff
@@ -1003,6 +1048,41 @@ def collect_prometheus_metrics(
...
+            time.sleep(10)
+            if ocsci_config.RUN["save_logs"]:
```
petr-balogh, line 1074: use .get here as well please

ocs_ci/framework/conf/default_config.yaml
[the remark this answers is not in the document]
petr-balogh: ok, but please please document this value anyway in README and write that his is not supposed to be changed by user but by the fixture itself.

Also I think you can rename to something like:
`save_live_pod_logs`

If it's not suppose to be changed by user I would even not put this value to default config and just make sure you are using .get mehod when you getting this value.

This case you don't have to maybe even document in readme as this value is not suppose to be changed by user at all.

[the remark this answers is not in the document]
ebenahar: Thanks, added documentation.

ocs_ci/ocs/utils.py
ebenahar: :) 
Sure, will do

ocs_ci/ocs/utils.py
ebenahar: done

ocs_ci/ocs/utils.py
ebenahar: done

ocs_ci/ocs/utils.py
```diff
@@ -890,6 +893,48 @@ def collect_noobaa_db_dump(log_dir_path):
...
+        for pod, container in pods_containers.items():
+            cmd = f"oc -n openshift-storage logs {pod} -c {container}"
+            logs_file = os.path.join(live_logs_dir_path, f"{pod}")
+            out = run_cmd(cmd)
+            if patterns_to_log:
+                for pattern_to_log in patterns_to_log:
+                    if pattern_to_log in out:
+                        log.info(
+                            f"The requested pattern {pattern_to_log} was found and it will be saved in {logs_file}"
+                        )
```
yitzhak12, line 933: Something is missing here... where is the save command in the file for every pattern? Or if it finds one occurrence of the `pattern_to_log`, it adds the logs to `log_file`? In that case, we can stop the for-loop after one occurrence and continue with the command:
`with open(logs_file, "w") as f:
                f.write(out)`

ocs_ci/ocs/utils.py
```diff
@@ -1003,6 +1048,46 @@ def collect_prometheus_metrics(
...
+        if thread:
+            thread.join()
+
+    request.addfinalizer(finalizer)
+
+    ocsci_config.RUN["save_live_pod_logs"] = True
+
+    def save_logs():
+        while ocsci_config.RUN.get("save_live_pod_logs"):
+            time.sleep(10)
```
yitzhak12, line 1073: Maybe we can add here a var with a name like `seconds_to_wait_between_save_logs = 10`. Currently, it's not so clear why we need to wait 10 seconds.

ocs_ci/ocs/utils.py
```diff
@@ -1003,6 +1048,46 @@ def collect_prometheus_metrics(
...
+        ocsci_config.RUN["save_live_pod_logs"] = False
+        if thread:
+            thread.join()
+
+    request.addfinalizer(finalizer)
+
+    ocsci_config.RUN["save_live_pod_logs"] = True
+
+    def save_logs():
+        while ocsci_config.RUN.get("save_live_pod_logs"):
```
yitzhak12, line 1072: Just for me to learn... when should this for-loop stop? I mean when the condition of `ocsci_config.RUN.get("save_live_pod_logs")` will change to `False`?

ocs_ci/ocs/utils.py
ebenahar: the idea is not to save the pattern in a file. It's rather to inform, in the test log, if the pattern was found and in which location of log files it resides. The reason is that this new functionality saves the pod logs every 10 seconds, which results in having many sub-directories, sometimes in hundreds. So this indication is a hint where to look for in those files

ocs_ci/ocs/utils.py
ebenahar: sure, I will add this var and also a comment explaining why we wait

ocs_ci/ocs/utils.py
ebenahar: it changes to `False` and stops in https://github.com/red-hat-storage/ocs-ci/pull/4075/files#diff-13cfcd3566979733be4860f4cc0db5af5382acbf7cf54fd94eb36fab45474c21R1063

Which is the test teardown.
The idea of this new functionality is to save the logs of pods, every few seconds (10 seconds), as long as the test is running. In this PR, the test where I added this is `test_upgrade`, in order to collect Noobaa DB migration job pods, as in 4.7 Noobaa changes its DB from Mongo to PGSQL. 
The new capability that is being added in this PR allows collecting the logs of these DB migration job pods, which are getting deleted before the test ends so we don't have their logs collected in the must gather upon test completion

ocs_ci/ocs/utils.py
ebenahar: In second thought, maybe it is not necessary to wait at all between iterations.
I will try without any sleep

ocs_ci/ocs/utils.py
yitzhak12: Ok, I understand.

ocs_ci/ocs/utils.py
yitzhak12: Yes, I see. Thanks for the explanation.

petr-balogh:
@ebenahar what is the status here?

PR #4089, opened 2021-04-01
RBD thick provisioning - verify PVC expansion, snapshot and clone

jilju:
This PR covers the following test cases:
1. PVC snapshot restore, clone and expansion using think provision storage class.
OCS-2502 - FT-ThickProvisioning-PVC snapshot, clone, expansion test

2. Use thick provision storage class while restoring the snapshot of a thin provisioned PVC. Perform  expansion and clone operations on the restored PVC.
OCS-2507 - FT-Thick provision storage class to restore the snapshot of a thin provisioned PVC

3. Use thin provision storage class while restoring the snapshot of a thick provisioned PVC. Perform  expansion and clone operations on the restored PVC.
OCS-2508 - FT- Thin provision storage class to restore snapshot of a thick provisioned PVC

Updated create_pvcs_and_pods fixture to enable the capability to create CephFS or RBD PVC alone with the storage class specified.

Signed-off-by: Jilju Joy jijoy@redhat.com

### DIFF
--- a/tests/manage/pv_services/conftest.py
+++ b/tests/manage/pv_services/conftest.py
@@ -24,6 +24,8 @@ def create_pvcs_and_pods(multi_pvc_factory, pod_factory, service_account_factory
         num_of_cephfs_pvc=None,
         replica_count=1,
         deployment_config=False,
+        sc_rbd=None,
+        sc_cephfs=None,
     ):
         """
         Args:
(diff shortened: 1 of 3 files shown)


tests/manage/pv_services/conftest.py
```diff
@@ -139,7 +154,7 @@ def factory(
         # pods_dc will be an empty list if deployment_config is False
         for pod_dc in pods_dc:
             pod_objs = pod.get_all_pods(
-                namespace=project.namespace,
+                namespace=pvcs[0].project.namespace,
```
yitzhak12, line 157: Should we check if `pvcs` is an empty list or assume that it always has at least one element?

jilju: ```pvcs``` will contain at least one element. Only possibility of empty ```pvcs``` list is if the value of both ```num_of_cephfs_pvc``` and ```num_of_rbd_pvc``` are passed as  0 which is not a correct use case.

tests/manage/pv_services/test_expansion_snapshot_clone.py
```diff
@@ -2,22 +2,43 @@
...
 
 @tier2
-@skipif_ocs_version("<4.6")
 @skipif_ocp_version("<4.6")
-@pytest.mark.polarion_id("OCS-2408")
+@pytest.mark.parametrize(
+    argnames=["pvc_create_sc_type", "restore_sc_type"],
+    argvalues=[
+        pytest.param(
+            *["thin", "thin"],
```
yitzhak12, line 25: The words "thin", "thick" I would replace with constants

yitzhak12 (commented):
The PR looks good

tests/manage/pv_services/conftest.py
yitzhak12: Yes, I see. It makes sense.

tests/manage/pv_services/test_expansion_snapshot_clone.py
jilju: These parameters are used to distinguish between the usage of thick and thin provision storage class in this test case. I don't think the value of these parameters can be made as a standard.

jilju:
> 
> Cluster Name: jijoy-apr14
> Cluster Configuration:
> PR Test Suite:
> PR Test Path: tests/manage/pv_services/pvc_snapshot/test_restore_snapshot_using_different_sc.py
> Additional Test Params:
> OCP VERSION: 4.8
> OCS VERSION: 4.8
> 
> [Job PASSED](https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-trigger-test-pr/532//testReport).

Executed this test to verify that the code change in the fixture ```create_pvcs_and_pods``` have no impact on existing usages

yitzhak12 (approved):

ebenahar (approved):

PR #4094, opened 2021-04-02
OC download fixes

clacroix12:
This PR will address a few of the issues raised in #4093

PR #4099, opened 2021-04-06
get_openshift_client: Avoid removing unnecessarily

zmc:
On some platforms (e.g. ppc64le), we deploy OCS on top of an OCP that
was deployed with a very different toolchain from ocs-ci. We don't have
quite the same level of control over the exact OCP version.
get_openshift_client's current behavior is problematic here, especially
since it only knows how to find amd64 binaries. If skip_ocp_deployment
is set, we ought to assume that the oc binary is safe to use.

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -710,6 +710,11 @@ def get_openshift_client(
                 f"Existing client version ({current_client_version}) does not match "
                 f"configured version ({version})."
             )
+            if config.ENV_DATA.get("skip_ocp_deployment"):
+                log.info(
+                    "ENV_DATA.skip_ocp_deployment is set; leaving client in place."
+                )
+                download_client = False
         else:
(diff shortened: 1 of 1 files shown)


clacroix12:
@zmc i have a PR open that fixes this (among other things): #4094

zmc:
@clacroix12 Whoops, I hadn't seen that! Fortunately this only took a few minutes to whip up. Your approach is obviously a better overall approach, but I do think the logic in this one is still appropriate. I can repush something based on yours if you want.

clacroix12:
I'm hesitant to have too many conditions here as I would like to keep it as straight forward as possible, especially when we will have an explicit config item for skipping this process entirely. Also, are we sure that we will never want to download the openshift client if we have skipped the OCP deploy?

zmc:
Hmm, yeah, on second thought let's just leave this part out for now.

PR #4106, opened 2021-04-07
Adding skip for one of the scale test for power

AaruniAggarwal:
Fixes : One of the scale test ie. `tests/e2e/scale/test_scale_osds_fill_75%_reboot_workers.py` is failing as we are not supporting  **add capacity** capability. So skipping the test.
Signed-off-by: Aaruni Aggarwal <aaruniagg@gmail.com>


prsurve (approved):

ramkiperiy (approved):

ramkiperiy:
@AaruniAggarwal you might need to add the IBM skip marker in https://github.com/red-hat-storage/ocs-ci/blob/master/tests/e2e/scale/test_pvc_creation_deletion_scale.py testcase too since there is PR #4112  to uncomment the memory_leak function usage

AaruniAggarwal:
Thanks Ram for the information. Will check that by running this testcase manually. If it also fails due to memory leak then will add this test to skip on our platform.

PR #4113, opened 2021-04-09
Explicitly set log level for reportportal_client logger

clacroix12:

### DIFF
--- a/conf/README.md
+++ b/conf/README.md
@@ -117,6 +117,7 @@ Reporting related config. (Do not store secret data in the repository!).
 * `gather_on_deploy_failure` - Run must-gather on deployment failure or not (Default: true)
 * `collect_logs_on_success_run` - Run must-gather on successful run or not (Default: false)
 * `must_gather_timeout` - Time (in seconds) to wait before timing out during must-gather
+* `rp_client_log_level` - Log level for the reportportal_client logger (Default: ERROR)
 
 #### ENV_DATA
 
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -102,6 +102,7 @@ REPORTING:
   default_ocs_must_gather_latest_tag: 'latest-4.7'
   gather_on_deploy_failure: true
   collect_logs_on_success_run: False
+  rp_client_log_level: "ERROR"
 
 # This is the default information about environment.
 ENV_DATA:
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -252,6 +252,7 @@ def pytest_configure(config):
 
     """
     set_log_level(config)
+    set_rp_client_log_level()
     # Somewhat hacky but this lets us differentiate between run-ci executions
     # and plain pytest unit test executions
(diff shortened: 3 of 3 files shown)


clacroix12:
@mbukatov I verified locally that by explicitly changing the log level of the reportportal_client logger we can filter out the debug statements in our log files. I was originally hoping to move these to a new file through an additional file handler however that is apparently not as easy as it sounds. I'm sure it has something to do with the combination of pytest and how we are configuring loggers currently. I can spend more time on it if we think it's necessary but for now this fix at least addresses the issue of unnecessary reportportal_client log messages cluttering up our log files. Let me know what you think of this fix.

petr-balogh (approved):

mbukatov:
Understood. I agree that a proper solution is to have dedicated reportportal log file, so I'm obviously not happy about it.

I wonder how much it helps though: could you run some set of tests (eg. all tier1 cases from single python test_ module) and compare the number of reportportal lines in the log?

Eg. in the issue, I observed that about 28% of log lines are coming from reportportal module:
 
```
$ wc -l logs
7906 logs
$ grep reportportal logs | wc -l
2208
```

Comparing that with the result with patch applied would be helpful to understand how much is the proposed hack practical.

Since we are talking about a hack instead of a proper solution, we should also consider:
- dropping all reportportal log lines altogether
- setting up a separate log level for reportportal

mbukatov (commented):
Requires further discussion. A proposed hack should also not claim to fix the original issue.

clacroix12:
> Comparing that with the result with patch applied would be helpful to understand how much is the proposed hack practical.

I can run a test to give you an exact number but you could just look at how many of those lines are ERROR level. That will be how many will remain after this fix is in place.

> Since we are talking about a hack instead of a proper solution, we should also consider:
> * setting up a separate log level for reportportal

We don't control what level things are logged at in the underlying reportportal python agent so I don't see how this would be feasible. We can however control what level logs we care about, which is what I'm doing in this PR.

vasukulkarni:
This doesn't have an issue link and I believe @mbukatov issue might not be related to this fix, This one is needed to keep the reportal instance clean from the debugging point of view.

vasukulkarni (approved):

clacroix12:
> This doesn't have an issue link and I believe @mbukatov issue might not be related to this fix, This one is needed to keep the reportal instance clean from the debugging point of view.

I removed the issue link because as martin pointed out it's not a full "fix" for the issue as it doesn't handle moving these logs to a separate file. However it is very much related to martin's issue. This fix is to keep our log files clean of reportportal_client debug statements, not keeping our report portal instance clean of debugging statements.

petr-balogh (approved):

mbukatov:
Ok, so I randomly grabbed another test log and rechecked:

```
$ wc -l logs
702392 logs
$ grep reportportal logs | wc -l
21723
$ grep "ERROR - reportportal" logs | wc -l
0
```
In this case, the reportportal log pullution wasn't that dramatic in relative numbers:
- 3% of log lines are comming from reportportal
- no log line comes with ERROR log level from report portal

It seems that this will be practical improvement indeed.

I wonder how does ERROR log levels from reportportal look like ans what would they mean for the test though.

PR #2697, opened 2020-08-12
Scale 12 OCS workers, 36 OSDs and 6000 PVCs attached to PODs and perform tests

ramkiperiy:
To create 12 OCS workers, 36 OSDs, 6000 PVCs
Further testing re-spin, reboot and shutdown of pods, nodes

Signed-off-by: Ramakrishnan <rperiyas@redhat.com>

PR #4133, opened 2021-04-14
Add tests/e2e/scale/test_osd_node_balancing.py

wusui:

This test checks osd distributions for 4.7 clusters where 3 nodes and 3 osds are added at a time.  This should work for
both flexible scaling and non-flexible situations.  This makes sure that one case of OCSQE-571 is tested, and also makes sure that comment #33 in bugzilla Bug 1817438 is checked.

### DIFF
--- a/tests/e2e/scale/test_osd_node_balancing.py
+++ b/tests/e2e/scale/test_osd_node_balancing.py
@@ -1,0 +1,145 @@
+"""
+Test osd node balancing by adding nodes and osds and checking their distribution
+"""
+import logging
+from uuid import uuid4
+from ocs_ci.framework.testlib import scale_changed_layout, ignore_leftovers
+from ocs_ci.ocs import constants, ocp
+from ocs_ci.ocs.node import get_nodes
+from ocs_ci.ocs.perfresult import PerfResult
+from ocs_ci.ocs.perftests import PASTest
+from ocs_ci.ocs.scale_lib import scale_capacity_with_deviceset, scale_ocs_node
+from ocs_ci.utility import templating
+from ocs_ci.utility.utils import ceph_health_check
+
+FINAL_REPORT = "Final Report"
+INITIAL_SETUP = "Initial Setup"
+MAX_NODE_COUNT = 9
+MAX_OSDS_PER_NODE = 3
+START_NODE_NUM = 3
+REPLICA_COUNT = 3
+OSD_LIMIT_AT_START = MAX_OSDS_PER_NODE * START_NODE_NUM
+MAX_TIMES_ADDED = 3
+
+
+def collect_stats(action_text, elastic_info):
+    """
+    Write the current configuration information into the REPORT file.
+    This information includes the osd, nodes and which osds are on which
+    nodes.  The minimum and maximum numbers of osds per node are also
+    computed and saved.  If this is the final call to collect_stats
+    (action_text parameter is FINAL_REPORT), then the data collected
+    in the REPORT file is also displayed in the log.
+
+    Args:
+        action_text -- Title of last action taken
+                (usually adding nodes or adding osds)
+        elastic_info -- ElasticData object
+    """
+    output_info = {"title": action_text}
+    pod_obj = ocp.OCP(
+        kind=constants.POD, namespace=constants.OPENSHIFT_STORAGE_NAMESPACE
+    )
+    osd_list = pod_obj.get(selector=constants.OSD_APP_LABEL)["items"]
+    node_stats = {}
+    for osd_ent in osd_list:
+        osd_node = osd_ent["spec"]["nodeName"]
+        if osd_node in node_stats:
+            node_stats[osd_node].append(osd_ent)
+        else:
+            node_stats[osd_node] = [osd_ent]
+    osds_per_node = []
+    for entry in node_stats:
+        osds_per_node.append(len(node_stats[entry]))
+    wnodes = get_nodes(constants.WORKER_MACHINE)
+    for wnode in wnodes:
+        if wnode.name not in node_stats:
+            osds_per_node.append(0)
+    maxov = max(osds_per_node)
+    minov = min(osds_per_node)
+    this_skew = maxov - minov
+    logging.info(f"Skew found is {this_skew}")
+    output_info["osds"] = osd_list
+    output_info["worker_nodes"] = wnodes
+    output_info["pairings"] = {}
+    for entry in osd_list:
+        output_info["pairings"][entry["metadata"]["name"]] = entry["spec"]["nodeName"]
+    output_info["maxov"] = maxov
+    output_info["minov"] = minov
+    output_info["skew_value"] = this_skew
+    balanced = True
+    if this_skew > 1 and maxov > MAX_OSDS_PER_NODE:
+        balanced = False
+    elastic_info.add_key(elastic_info.record_counter, output_info)
+    elastic_info.log_recent_activity()
+    elastic_info.record_counter += 1
+    if not balanced:
+        logging.info("OSDs are not balanced")
+        if action_text == FINAL_REPORT:
+            logging.info("FINAL RESULT -- OSDs are not balanced")
+            ceph_health_check(tries=30, delay=60)
+            return
+    if action_text == FINAL_REPORT:
+        logging.info("FINAL RESULT -- OSDs are balanced")
+    ceph_health_check(tries=30, delay=60)
+
+
+class ElasticData(PerfResult):
+    """
+    Wrap PerfResult and keep track of a counter to be used as
+    an index into the table data saved.
+    """
+
+    def __init__(self, uuid, crd):
+        super(ElasticData, self).__init__(uuid, crd)
+        self.index = "test_osd_node_balancing"
+        self.new_index = "test_osd_node_balancing_new"
+        self.record_counter = 0
+
+    def log_recent_activity(self):
+        new_data = self.results[self.record_counter]
+        logging.info(new_data["title"])
+        logging.info("pairings:")
+        for entry in new_data["pairings"]:
+            logging.info(f"     {entry} -- {new_data['pairings'][entry]}")
+        logging.info(f"maxov: {new_data['maxov']}")
+        logging.info(f"minov: {new_data['minov']}")
+        logging.info(f"skew_value: {new_data['skew_value']}")
+
+
+@ignore_leftovers
+@scale_changed_layout
+class Test_Osd_Balance(PASTest):
+    """
+    There is no cleanup code in this test because the final
+    state is much different from the original configuration
+    (several nodes and osds have been added)
+    """
+
+    def test_osd_balance(self, es):
+        """
+        Current pattern is:
+            add 6 osds (9 total, 3 nodes)
+            add 3 nodes
+            add 9 osds (18 total, 6 nodes)
+            add 3 nodes
+            add 9 osds (27 total, 9 nodes)
+        """
+        crd_data = templating.load_yaml(constants.OSD_SCALE_BENCHMARK_YAML)
+        our_uuid = uuid4().hex
+        self.elastic_info = ElasticData(our_uuid, crd_data)
+        self.elastic_info.es_connect()
+        collect_stats(INITIAL_SETUP, self.elastic_info)
+        for cntr in range(0, MAX_TIMES_ADDED):
+            num_nodes = len(get_nodes(constants.WORKER_MACHINE))
+            osd_incr = 3
+            if cntr == 0 and num_nodes == START_NODE_NUM:
+                osd_incr = 2
+            if osd_incr == 3:
+                scale_ocs_node()
+                collect_stats("Three nodes have been added", self.elastic_info)
+            cntval = 3 * osd_incr
+            logging.info(f"Adding {cntval} osds to nodes")
+            scale_capacity_with_deviceset(add_deviceset_count=osd_incr, timeout=900)
+            collect_stats("OSD capacity increase", self.elastic_info)
+        collect_stats(FINAL_REPORT, self.elastic_info)
--- a/tests/e2e/scale/test_osd_node_balancing.py
+++ b/tests/e2e/scale/test_osd_node_balancing.py
@@ -2,10 +2,21 @@
 Test osd node balancing by adding nodes and osds and checking their distribution
 """
 import logging
+import re
 from uuid import uuid4
+from ocs_ci.framework.pytest_customization.marks import (
+    skipif_aws_i3,
+    skipif_bm,
+    skipif_external_mode,
+    skipif_ibm_cloud,
+    skipif_ibm_power,
+    skipif_lso,
+    ipi_deployment_required,
+)
 from ocs_ci.framework.testlib import scale_changed_layout, ignore_leftovers
 from ocs_ci.ocs import constants, ocp
 from ocs_ci.ocs.node import get_nodes
+from ocs_ci.ocs.ocp import get_ocs_version
 from ocs_ci.ocs.perfresult import PerfResult
 from ocs_ci.ocs.perftests import PASTest
 from ocs_ci.ocs.scale_lib import scale_capacity_with_deviceset, scale_ocs_node
@@ -14,6 +25,7 @@ from ocs_ci.utility.utils import ceph_health_check
 
 FINAL_REPORT = "Final Report"
 INITIAL_SETUP = "Initial Setup"
+NOT_BALANCED = "OSDs are not balanced"
 MAX_NODE_COUNT = 9
 MAX_OSDS_PER_NODE = 3
 START_NODE_NUM = 3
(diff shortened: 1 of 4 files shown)


tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    pod_obj = ocp.OCP(
+        kind=constants.POD, namespace=constants.OPENSHIFT_STORAGE_NAMESPACE
+    )
+    osd_list = pod_obj.get(selector=constants.OSD_APP_LABEL)["items"]
+    node_stats = {}
+    for osd_ent in osd_list:
+        try:
+            osd_node = osd_ent["spec"]["nodeName"]
+        except KeyError:
```
keesturam, line 105: There should definitely be a value for nodeName for each of the pod - why is this within a try-except block?

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,193 @@
...
+        if osd_node in node_stats:
+            node_stats[osd_node].append(osd_ent)
+        else:
+            node_stats[osd_node] = [osd_ent]
+    osds_per_node = []
+    for entry in node_stats:
+        osds_per_node.append(len(node_stats[entry]))
+    wnodes = get_nodes(constants.WORKER_MACHINE)
+    for wnode in wnodes:
+        logging.info(wnode.name)
```
keesturam, line 116: `wnode.name` doesn't provide any valuable info. What is the intention of adding this to the logs?

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,193 @@
...
+        fd.write(f"\nCurrent Nodes: {len(wnodes)}\n")
+        for entry in wnodes:
+            fd.write(f"\t\t{entry.name}\n")
+        fd.write("\nOSD distribution:\n")
+        for entry in osd_list:
+            fd.write(f"\t\t{entry['metadata']['name']}")
+            try:
+                fd.write(f" is on {entry['spec']['nodeName']}\n")
+            except KeyError:
+                fd.write(" is on an unknown node\n")
```
keesturam, line 139: same as the above comment on keyerror. It would be really strange for a pod not to have nodename assigned.

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,193 @@
...
+        fd.write(f"\nMaximum number of OSDs on a node: {maxov}\n")
+        fd.write(f"Minimum number of OSDs on a node: {minov}\n")
+        fd.write(f"Skew (Difference in OSD counts): {this_skew}\n")
+        if this_skew > 1:
+            fd.write("\n*** ERROR *** OSDs are not balanced between nodes\n")
```
keesturam, line 144: shouldn't we assert here?

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,193 @@
...
+        with open(outfile, "r") as fd:
+            logged_info = fd.read()
+            for inline in logged_info.split("\n"):
+                logging.info(inline)
+        logging.info(f"Summary of test results output in {outfile}")
+
+
+@ignore_leftovers
+@scale
+class Test_Prototype(E2ETest):
```
keesturam, line 158: Let's use PASTest here

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,193 @@
...
+from ocs_ci.ocs import constants, machine, ocp
+from ocs_ci.ocs.node import add_new_node_and_label_it, get_osd_running_nodes, get_nodes
+from ocs_ci.ocs.resources import storage_cluster
+from ocs_ci.utility.utils import (
+    ceph_health_check,
+    TimeoutSampler,
+    ocsci_log_path,
+)
+
+REPORT = "Node_and_Osd_distribution_report.txt"
```
keesturam, line 15: I believe the intention here is to retain the stages of the test in a file, but it would be great if this is done in the form of logging and as part of the flow of the test. 
@vasukulkarni need your thoughts here pls.

keesturam (dismissed):

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,193 @@
...
+    wnodecnt = len(get_nodes(constants.WORKER_MACHINE))
+    if wnodecnt == MAX_NODE_COUNT:
+        return
+    osd_running_nodes = get_osd_running_nodes()
+    logging.info(f"OSDs are running on nodes {osd_running_nodes}")
+    # Get the machine name using the node name
+    machine_names = [
+        machine.get_machine_from_node_name(osd_node) for osd_node in osd_running_nodes
+    ]
+    logging.info(f"{osd_running_nodes} associated " f"machine are {machine_names}")
```
Avilir, line 39: can be one string `f"{osd_running_nodes} associated machine are {machine_names}"`

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,193 @@
...
+    # This next section makes a new node on which OSD pods can run
+    for mnames in range(0, NUM_NODE_ADDS):
+        # This next section makes a new node on which OSD pods can run
+        new_node = add_new_node_and_label_it(
+            machineset_names[mnames], mark_for_ocs_label=False
+        )
+        node_obj.add_label(
+            resource_name=new_node[0], label=constants.OPERATOR_NODE_LABEL
+        )
+    collect_stats("Three nodes have been added")
```
Avilir, line 56: should it be variable (`NUM_NODE_ADDS`) ?

tests/e2e/scale/test_osd_node_balancing.py
Avilir: IMO, it will be good idea to push the report as doc to the elastic-search server.

tests/e2e/scale/test_osd_node_balancing.py
wusui: I ran into a problem with this earlier so I added this.  I think this issue is now fixed and I forgot to remove it.  I will remove it now.

tests/e2e/scale/test_osd_node_balancing.py
wusui: Yes.  I'll change this.

tests/e2e/scale/test_osd_node_balancing.py
wusui: I will get rid of this try as well.

tests/e2e/scale/test_osd_node_balancing.py
wusui: I'll remove the log statement.

tests/e2e/scale/test_osd_node_balancing.py
wusui: Yes. I'll fix.

tests/e2e/scale/test_osd_node_balancing.py
wusui: okay. will do.

tests/e2e/scale/test_osd_node_balancing.py
wusui: Yes.  I will change this.

tests/e2e/scale/test_osd_node_balancing.py
wusui: Note: I think that this message should be changed to something other than ERROR.  I think that we should assert only when we add new nodes and when the test is finished.  The reason for this is when we have 3 nodes with 3 OSDs on each, when we add 3 new nodes we will have 3 nodes with 3 OSDs and 3 nodes with no OSDs.  Now when we evenly add an OSD to each of the nodes with 0 OSDs we will have 3 OSDs on 3 nodes and 1 OSD on 3 nodes.  This is the best that we can do in this case and we do not want to assert when this happens.

tests/e2e/scale/test_osd_node_balancing.py
wusui: Yes.  I'll fix this.

tests/e2e/scale/test_osd_node_balancing.py
wusui: I will change this.

tests/e2e/scale/test_osd_node_balancing.py
wusui: Assert when imbalance occurs has been added.

tests/e2e/scale/test_osd_node_balancing.py
wusui: Pushing the report to the elastic search server is now a new git issue that I will work on (4185).

vasukulkarni: We should definitely add the logs or it becomes difficult to debug,  i think i also need to go over with warren to understand it better, i am missing few parts of the logic.

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,187 @@
...
+INITIAL_SETUP = "Initial Setup"
+MAX_NODE_COUNT = 9
+MAX_OSDS_PER_NODE = 3
+START_NODE_NUM = 3
+REPLICA_COUNT = 3
+OSD_LIMIT_AT_START = MAX_OSDS_PER_NODE * START_NODE_NUM
+NUM_NODE_ADDS = MAX_NODE_COUNT // REPLICA_COUNT
+
+
+def add_some_worker_nodes():
```
vasukulkarni, line 27: Can be just add_worker_nodes and default to 3.

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,187 @@
...
+    if action_text == FINAL_REPORT or not balanced:
+        with open(outfile, "r") as fd:
+            logged_info = fd.read()
+            for inline in logged_info.split("\n"):
+                logging.info(inline)
+        logging.info(f"Summary of test results output in {outfile}")
+    assert balanced, "OSDs are not balanced"
+
+
+@ignore_leftovers
```
ramkiperiy, line 150: If we are expecting the cluster to be destroyed after this test, we should not use left_overs.

ramkiperiy:
@wusui script does scaling of OCS worker nodes and OSD's, we should add more tests like scaling of PVCs and filling up data may be we can take this up in next PR

wusui:
The following changes have been made since the previous reviews:

1. Instead of writing collected data to a tempfile, that data is now
   stored in the elastic search database.

2. The logic in test_osd_balance was cleaned up.

3. Wait_for_nodes and wait_for_storage_pods are now used to wait until
   nodes and osds are created.

4. A few calls that are now not necessary have been removed.

5. A bug where rook-ceph-detect-version pods were causing later failures
   is now handled with a retry loop (some versions of previous code just
   slept).

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,205 @@
...
+START_NODE_NUM = 3
+REPLICA_COUNT = 3
+OSD_LIMIT_AT_START = MAX_OSDS_PER_NODE * START_NODE_NUM
+MAX_TIMES_ADDED = 3
+
+
+def add_some_worker_nodes(elastic_info):
+    """
+    Add three worker nodes to the storage cluster.  Update the report file.
+    """
```
Avilir, line 37: add an args section to docstring

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,205 @@
...
+    for mnames in range(0, REPLICA_COUNT):
+        add_new_node_and_label_it(machineset_names[mnames], mark_for_ocs_label=True)
+        wait_for_nodes_status()
+    collect_stats(f"{REPLICA_COUNT} nodes have been added", elastic_info)
+
+
+def increase_osd_capacity(elastic_info):
+    """
+    Add three osds to the cluster.  Update the report file.
+    """
```
Avilir, line 60: same as above

wusui: Done

tests/e2e/scale/test_osd_node_balancing.py
wusui: Done

tests/e2e/scale/test_osd_node_balancing.py
wusui: This new version uses the elastic search database to save these results

tests/e2e/scale/test_osd_node_balancing.py
wusui: changes made

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,249 @@
...
+        logging.info("pairings:")
+        for entry in new_data["pairings"]:
+            logging.info(f"     {entry} -- {new_data['pairings'][entry]}")
+        logging.info(f"maxov: {new_data['maxov']}")
+        logging.info(f"minov: {new_data['minov']}")
+        logging.info(f"skew_value: {new_data['skew_value']}")
+
+
+@ignore_leftovers
+@scale
```
ramkiperiy, line 216: @wusui we should not execute this TC with scale marker. As we are planning to add one more marker(not started any work on that yet) especially for scale TC which we can't perform teardown at the end of test.
I see the TC is already completed and with scale marker next testcase is going to fail in suite execution
1. Adding the new marker to setup and teardown the cluster after every test execution
2. Is there a way to execute this at end which is temporary solution
3. Other option would be to merge this PR with some dummy marker and later once we have the marker to teardown the cluster, we can update this code.

ramkiperiy: @keesturam ^^

ramkiperiy:
@wusui would request to remove scale marker in this TC and add scale_tier4.
**Note: TC should be consuming the cluster as it is, i.e. if you want to scale a OSD, and if the setup is already have scaled OSD's then have to decide either you want to add one more or the existing available OSD's are enough for you testing. We have to make sure the code is designed in such a way.**

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,205 @@
...
+INITIAL_SETUP = "Initial Setup"
+MAX_NODE_COUNT = 9
+MAX_OSDS_PER_NODE = 3
+START_NODE_NUM = 3
+REPLICA_COUNT = 3
+OSD_LIMIT_AT_START = MAX_OSDS_PER_NODE * START_NODE_NUM
+MAX_TIMES_ADDED = 3
+
+
+def add_some_worker_nodes(elastic_info):
```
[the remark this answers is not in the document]
ramkiperiy: @wusui can you use "def scale_ocs_node(node_count=3):" in PR #2697, it does add OCS worker nodes and label it.

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,247 @@
...
+    retry(
+        (CephDetectVersionPodFound),
+        tries=18,
+        delay=10,
+        backoff=1,
+    )(find_problem_pods)()
+    collect_stats(f"{REPLICA_COUNT} nodes have been added", elastic_info)
+
+
+def increase_osd_capacity(elastic_info):
```
ramkiperiy, line 70: @wusui Please check function "def scale_capacity_with_deviceset(add_deviceset_count=2):" in PR #2697 for scaling cluster capacity, and use this helper function.

wusui: This was done in my last checkin

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,145 @@
...
+        for entry in new_data["pairings"]:
+            logging.info(f"     {entry} -- {new_data['pairings'][entry]}")
+        logging.info(f"maxov: {new_data['maxov']}")
+        logging.info(f"minov: {new_data['minov']}")
+        logging.info(f"skew_value: {new_data['skew_value']}")
+
+
+@ignore_leftovers
+@scale_changed_layout
+class Test_Osd_Balance(PASTest):
```
ramkiperiy, line 112: @wusui you might need to update the skip markers, i.e. scale_ocs_node() function supports only ipi deployment, i.e. external modes, vmware_upi, vmware_LSO, aws_LSO platform deployed cluster can't be scaled with more worker nodes and OSD's
Please note line no 37-43 in https://github.com/red-hat-storage/ocs-ci/blob/master/tests/e2e/scale/test_scale_12_OCS_worker_nodes_and_6000_PVCs.py#L37 for reference.

ramkiperiy: apart from above comment PR looks good, we can merge it, PR validation passed

tests/e2e/scale/test_osd_node_balancing.py
```diff
@@ -0,0 +1,145 @@
...
+    This information includes the osd, nodes and which osds are on which
+    nodes.  The minimum and maximum numbers of osds per node are also
+    computed and saved.  If this is the final call to collect_stats
+    (action_text parameter is FINAL_REPORT), then the data collected
+    in the REPORT file is also displayed in the log.
+
+    Args:
+        action_text -- Title of last action taken
+                (usually adding nodes or adding osds)
+        elastic_info -- ElasticData object
```
ramkiperiy, line 37: Could you change this doc string to below format
```
    Args:
        kube_job_obj (obj): Kube Job Object
        namespace (str): Namespace of PVC's created
        no_of_pod (int): POD count

    Return:
        pod_dict (dict): Dictionary of pod_name with creation time.
```

ramkiperiy (approved):

Avilir (approved):
LGTM

PR #4148, opened 2021-04-19
Flexible component deployment

shylesh:
Make components like rgw, cephfs, noobaa, cephblockpool
to be flexibly deplyed.

### DIFF
--- a/conf/ocsci/skip_blockpools.yaml
+++ b/conf/ocsci/skip_blockpools.yaml
@@ -1,3 +1,0 @@
-COMPONENTS:
-  enable_blockpools: False
-
--- a/conf/ocsci/skip_blockpools.yaml
+++ b/conf/ocsci/skip_blockpools.yaml
@@ -1,0 +1,3 @@
+COMPONENTS:
+  enable_blockpools: False
+
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -567,6 +567,22 @@ class Deployment(object):
             kms.deploy()
         cluster_data = templating.load_yaml(constants.STORAGE_CLUSTER_YAML)
 
+        # Figure out all the OCS modules enabled/disabled
+        # CLI parameter --disable-components takes the precedence over
+        # anything which comes from config file
+        if config.ENV_DATA.get("disable_components"):
+            for component in config.ENV_DATA["disable_components"]:
+                config.COMPONENTS[f"enable_{component}"] = False
+
+        # Update cluster_data with respective component enable/disable
+        for key in config.COMPONENTS.keys():
+            if "noobaa" in key:
+                base_dict = cluster_data["spec"]
+            else:
+                base_dict = cluster_data["spec"]["managedResources"]
+            if not config.COMPONENTS[key]:
+                base_dict[constants.OCS_COMPONENTS_MAP[key.split("_")[1]]] = "ignore"
+
         if arbiter_deployment:
             cluster_data["spec"]["arbiter"] = {}
             cluster_data["spec"]["nodeTopologies"] = {}
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -488,7 +488,6 @@ class Deployment(object):
                 create_ocs_secret(constants.MARKETPLACE_NAMESPACE)
         if not live_deployment:
             self.create_ocs_operator_source(image)
-            pass
         self.subscribe_ocs()
         operator_selector = get_selector_for_ocs_operator()
         subscription_plan_approval = config.DEPLOYMENT.get("subscription_plan_approval")
@@ -574,6 +573,7 @@ class Deployment(object):
         if config.ENV_DATA.get("disable_components"):
             for component in config.ENV_DATA["disable_components"]:
                 config.COMPONENTS[f"disable_{component}"] = True
+                logger.warning(f"disabling: {component}")
 
         # Update cluster_data with respective component enable/disable
         for key in config.COMPONENTS.keys():
@@ -910,15 +910,16 @@ class Deployment(object):
                 timeout=600,
             )
 
-            # Check for CephFilesystem creation in ocp
-            cfs_data = cfs.get()
-            cfs_name = cfs_data["items"][0]["metadata"]["name"]
+            if not config.COMPONENTS["disable_cephfs"]:
+                # Check for CephFilesystem creation in ocp
+                cfs_data = cfs.get()
+                cfs_name = cfs_data["items"][0]["metadata"]["name"]
 
-            if helpers.validate_cephfilesystem(cfs_name):
-                logger.info("MDS deployment is successful!")
-                defaults.CEPHFILESYSTEM_NAME = cfs_name
-            else:
-                logger.error("MDS deployment Failed! Please check logs!")
+                if helpers.validate_cephfilesystem(cfs_name):
+                    logger.info("MDS deployment is successful!")
+                    defaults.CEPHFILESYSTEM_NAME = cfs_name
+                else:
+                    logger.error("MDS deployment Failed! Please check logs!")
 
         # Change monitoring backend to OCS
         if config.ENV_DATA.get("monitoring_enabled") and config.ENV_DATA.get(
@@ -934,8 +935,9 @@ class Deployment(object):
                 telemeter_server_url=config.ENV_DATA["telemeter_server_url"]
             )
 
-        # Change registry backend to OCS CEPHFS RWX PVC
-        registry.change_registry_backend_to_ocs()
+        if not config.COMPONENTS["disable_cephfs"]:
+            # Change registry backend to OCS CEPHFS RWX PVC
+            registry.change_registry_backend_to_ocs()
 
         # Verify health of ceph cluster
         logger.info("Done creating rook resources, waiting for HEALTH_OK")
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -572,7 +572,7 @@ class Deployment(object):
         # anything which comes from config file
         if config.ENV_DATA.get("disable_components"):
             for component in config.ENV_DATA["disable_components"]:
-                config.COMPONENTS[f"enable_{component}"] = False
+                config.COMPONENTS[f"disable_{component}"] = True
 
         # Update cluster_data with respective component enable/disable
         for key in config.COMPONENTS.keys():
@@ -580,7 +580,7 @@ class Deployment(object):
                 base_dict = cluster_data["spec"]
             else:
                 base_dict = cluster_data["spec"]["managedResources"]
-            if not config.COMPONENTS[key]:
+            if config.COMPONENTS[key]:
                 base_dict[constants.OCS_COMPONENTS_MAP[key.split("_")[1]]] = "ignore"
 
         if arbiter_deployment:
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -16,7 +16,7 @@ import yaml
 from ocs_ci.deployment.ocp import OCPDeployment as BaseOCPDeployment
 from ocs_ci.deployment.helpers.lso_helpers import setup_local_storage
 from ocs_ci.deployment.disconnected import prepare_disconnected_ocs_deployment
-from ocs_ci.framework import config
+from ocs_ci.framework import config, merge_dict
 from ocs_ci.ocs import constants, ocp, defaults, registry
 from ocs_ci.ocs.cluster import (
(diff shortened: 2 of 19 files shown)


shylesh:
@clacroix12 @vavuthu @dahorak @vasukulkarni ,
With this PR we are introducing a new cli option `--disable-components` followed by component names to be disabled from deployment. Currently this PR implements a comma separated component names  like 
```python
--disable-components rgw,cephfs,blockpools
```

However @petr-balogh suggested we go with 
```python
--disable-component rgw --disable-components cephfs --disable-component blockpool
```

which one you guys would feel more intuitive and better approach

dahorak:
The first (currently implemented) option is shorter and I would vote for it. But if it will be (easily) possible to use both ways together (e.g. `--disable-components rgw,cephfs --disable-component blockpool`), it might be handy in some cases.

ocs_ci/framework/main.py
```diff
@@ -66,6 +66,15 @@ def init_ocsci_conf(arguments=None):
     )
     parser.add_argument("--ocs-registry-image")
     parser.add_argument("--flexy-env-file", default="", help="Path to flexy env file")
+    parser.add_argument(
+        "--disable-components",
+        nargs="+",
```
dahorak, line 71: If I remember it correctly, the `nargs="+"` means, that you can use multiple values for one option (space separated) - e.g.:
```
--disable-components rgw cephfs blockpool
```
I would not recommend this type of handling as it might be slightly unclear in our case (what is the parameter value and what is next parameter).

clacroix12: > @clacroix12 @vavuthu @dahorak @vasukulkarni ,
> With this PR we are introducing a new cli option `--disable-components` followed by component names to be disabled from deployment. Currently this PR implements a comma separated component names like
> 
> ```python
> --disable-components rgw,cephfs,blockpools
> ```
> 
> However @petr-balogh suggested we go with
> 
> ```python
> --disable-component rgw --disable-components cephfs --disable-component blockpool
> ```
> 
> which one you guys would feel more intuitive and better approach

We have other CLI options in which we use the append action (what petr suggested). I think we should be consistent and not introduce different means of adding multiple values to CLI options. One [example](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/main.py#L61) of this is `--ocsci-conf`. Because we configure it to use the append action, we can call `--ocsci-conf` multiple times and the values get added to a list automatically.

clacroix12:
@shylesh just for consistency i would rename the config files to `disable_<component>.yaml` instead of `skip_<component>.yaml` Since enable/disable is the terminology we are using here, we should use the same in the filenames.

shylesh:
rebased on latest master

conf/ocsci/skip_blockpools.yaml
```diff
@@ -0,0 +1,3 @@
+COMPONENTS:
+  enable_blockpools: False
```
vavuthu, line 2: If you are changing the filename to disable_blockpools.yaml, then it would be good have parameter disable_blockpools: True

ocs_ci/framework/main.py
dahorak: The consistency with `--ocsci-conf` is good point and I fully support this option - that's really good point. That means, that you can repeat the whole combination of parameter name and value - e.g. `--ocsci-conf config1.yaml --ocsci-conf config2.yaml`, which is enabled via `action="append"` parameter in `parser.add_argument(...)` ([python doc](https://docs.python.org/3/library/argparse.html#action), [usage in ocs-ci](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/main.py#L61)).
But my point in the comment above was against using `nargs="+"`, which will allow you to put multiple values to single parameter like this: `--ocsci-conf config1.yaml config2.yaml` ([python doc](https://docs.python.org/3/library/argparse.html#nargs). This syntax is might be slightly confusing to distinguish what is positional parameter and what is value for named parameter, so if it is not necessary, I would avoid it.

dahorak:
> The first (currently implemented) option is shorter and I would vote for it. But if it will be (easily) possible to use both ways together (e.g. `--disable-components rgw,cephfs --disable-component blockpool`), it might be handy in some cases.

Coady have good point to keep it consistent with other parameters used in `ocs-ci`, so I agree with having only the second option (repeat whole combination of named argument and it's value) as it is for `--ocsci-conf` parameter.

ocs_ci/framework/main.py
clacroix12: @dahorak I totally agree with you. I must have replied to your comment instead of making a new one myself.

ocs_ci/templates/ocs-deployment/storage-cluster.yaml
```diff
@@ -5,6 +5,15 @@ metadata:
   namespace: openshift-storage
   name: ocs-storagecluster
 spec:
+  managedResources:
```
petr-balogh, line 8: This is just from 4.8 and this file is used also in other version.

Please check if we deploy via UI if those are even defined, I think we are trying to keep this file as minimal as possible and this should use the default from rook operator if not defined in storage cluster CR

ocs_ci/deployment/deployment.py
```diff
@@ -486,6 +486,7 @@ def deploy_ocs_via_operator(self, image=None):
                 create_ocs_secret(constants.MARKETPLACE_NAMESPACE)
         if not live_deployment:
             self.create_ocs_operator_source(image)
+            pass
```
clacroix12, line 489: I assume this is leftover from local testing?

ocs_ci/framework/__init__.py
```diff
@@ -28,6 +28,7 @@ class Config:
     UPGRADE: dict = field(default_factory=dict)
     FLEXY: dict = field(default_factory=dict)
     UI_SELENIUM: dict = field(default_factory=dict)
+    COMPONENTS: dict = field(default_factory=dict)
```
clacroix12, line 31: We should document this new config section [here](https://github.com/red-hat-storage/ocs-ci/blob/master/conf/README.md)

ocs_ci/deployment/deployment.py
shylesh: oops, yes , I missed it.

ocs_ci/framework/__init__.py
shylesh: ack, I will do that

ocs_ci/ocs/constants.py
```diff
@@ -1283,3 +1283,11 @@
 
 WORKLOAD_STORAGE_TYPE_BLOCK = "block"
 WORKLOAD_STORAGE_TYPE_FS = "fs"
+# Components of OCS
+OCS_COMPONENTS = ["rgw", "cephfs", "noobaa", "blockpools"]
+OCS_COMPONENTS_MAP = {
+    "rgw": "cephObjectStores",
```
petr-balogh, line 1289: ah, so capital letter was the problem? :)

ocs_ci/framework/__init__.py
petr-balogh: still missing documentation

petr-balogh (commented):
Looks OK, after adding doc I will approve

ocs_ci/framework/__init__.py
shylesh: done

petr-balogh (approved):

ocs_ci/framework/main.py
```diff
@@ -66,6 +66,12 @@ def init_ocsci_conf(arguments=None):
     )
     parser.add_argument("--ocs-registry-image")
     parser.add_argument("--flexy-env-file", default="", help="Path to flexy env file")
+    parser.add_argument(
+        "--disable-components",
+        action="append",
+        choices=["rgw", "cephfs", "noobaa", "blockpools"],
+        help=("disable deployment of ocs components:rgw, cephfs, noobaa, blockpools."),
+    )
```
dahorak, line 74: Since it takes one _component_ peer argument, I would use singular in the argument name: `--disable-component`.
Also it will be good to mention in the help, that this argument can be used multiple times.

dahorak (approved):
Just small suggestion about the argument name and help, otherwise LGTM.

clacroix12 (dismissed):
LGTM! I agree with @dahorak about making the argument singular and adding to the help text but that can probably be done in a quick follow up PR. Nice work!

shylesh:
rebased on latest master

petr-balogh (dismissed):

clacroix12 (dismissed):

petr-balogh (dismissed):
Just thinking if we even need config files when we have cli argument? Maybe we can delete them and keep it cli only? Wdyt?

shylesh:
> Just thinking if we even need config files when we have cli argument? Maybe we can delete them and keep it cli only? Wdyt?

@petr-balogh yes we can do that but I was just thinking about jenkins jobs. Do we need check boxes for components ? for now we can use these files to skip components for testing and once we have jenkins jobs with selectors we can remove this file. Even now another way is free text box for the job where we can pass this config. Let me know what's the best approach.

petr-balogh:
> > Just thinking if we even need config files when we have cli argument? Maybe we can delete them and keep it cli only? Wdyt?
> 
> 
> 
> @petr-balogh yes we can do that but I was just thinking about jenkins jobs. Do we need check boxes for components ? for now we can use these files to skip components for testing and once we have jenkins jobs with selectors we can remove this file. Even now another way is free text box for the job where we can pass this config. Let me know what's the best approach.

We don't have jobs yet, so make jobs with new check boxes and do it this way directly. We have option to add additional pytest params for testing if we need it now

shylesh:
rebased on the latest master

petr-balogh (approved):

shylesh:
Verified locally with vsphere.

```html
16:28:04 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage get CephFilesystem  -n openshift-storage -o yaml
16:28:04 - MainThread - root - WARNING - list index out of range
16:28:04 - MainThread - root - WARNING - No CephFS found
16:28:04 - MainThread - ocs_ci.helpers.sanity_helpers - INFO - Deleting resources as a sanity functional validation
PASSED                                                                                                                           [100%]
---------------------------------------------------------- live log teardown -----------------------------------------------------------
16:28:05 - MainThread - tests.conftest - WARNING - During test were raised new alerts
16:28:05 - MainThread - tests.conftest - WARNING - [{'labels': {'alertname': 'KubeAPIErrorBudgetBurn', 'long': '3d', 'severity': 'warning', 'short': '6h'}, 'annotations': {'description': 'The API server is burning too much error budget.', 'summary': 'The API server is burning too much error budget.'}, 'state': 'pending', 'activeAt': '2021-05-18T23:28:05.333788348Z', 'value': '2.319061812430001e-02'}, {'labels': {'alertname': 'KubePodCrashLooping', 'container': 'prometheus', 'endpoint': 'https-main', 'job': 'kube-state-metrics', 'namespace': 'openshift-monitoring', 'pod': 'prometheus-k8s-0', 'service': 'kube-state-metrics', 'severity': 'warning'}, 'annotations': {'description': 'Pod openshift-monitoring/prometheus-k8s-0 (prometheus) is restarting 0.52 times / 10 minutes.', 'summary': 'Pod is crash looping.'}, 'state': 'pending', 'activeAt': '2021-05-18T23:25:46.474542648Z', 'value': '5.245166666666666e-01'}]
16:28:05 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc wait --for condition=ready pod -l app=rook-ceph-tools -n openshift-storage --timeout=120s
16:28:06 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage get pod -l 'app=rook-ceph-tools' -o jsonpath='{.items[0].metadata.name}'
16:28:06 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc -n openshift-storage exec rook-ceph-tools-5cfc5bbd87-h7cld -- ceph health
16:28:08 - MainThread - ocs_ci.utility.utils - INFO - Ceph cluster health is HEALTH_OK.
16:28:08 - MainThread - tests.conftest - INFO - Ceph health check passed at teardown
```

Due to LSO issue I couldn't verify this with LSO.

clacroix12 (approved):
LGTM!

shylesh:
rebased on latest master

petr-balogh (dismissed):

clacroix12 (approved):

petr-balogh:
@shylesh I see there is again some conflict :( so please rebase again. Once you will get approval please merge, so we will have it in. Thanks

shylesh:
rebased on latest master

clacroix12 (approved):

vasukulkarni (approved):

PR #4157, opened 2021-04-21
Skip on failure domain verification on upgraded lso cluster

OdedViner:

### DIFF
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -6,7 +6,6 @@ import logging
 import tempfile
 
 from jsonschema import validate
-from semantic_version import Version
 
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants, defaults, ocp
@@ -18,7 +17,7 @@ from ocs_ci.ocs.resources.pvc import get_deviceset_pvcs
 from ocs_ci.ocs.node import get_osds_per_node
 from ocs_ci.utility import localstorage, utils, templating, kms as KMS
 from ocs_ci.utility.rgwutils import get_rgw_count
-from ocs_ci.utility.utils import run_cmd, skipif_upgraded_from
+from ocs_ci.utility.utils import run_cmd
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -384,18 +383,15 @@ def ocs_install_verification(
...
-    zone_num = utils.get_az_count()
-    if (
-        config.DEPLOYMENT.get("local_storage")
-        and Version.coerce(ocs_version) >= Version.coerce("4.7")
-        and zone_num < 3
-    ):
-        storage_cluster_obj = get_storage_cluster()
+    storage_cluster_obj = get_storage_cluster()
+    is_flexible_scaling = (
+        storage_cluster_obj.get()["items"][0].get("spec").get("flexibleScaling", False)
```
nbalacha, line 388: Clusters with flexibleScaling disabled will not have this field in the yaml. Will is_flexible_scaling be set to false in such cases?

If this is meant to target the upgrade, we should also ensure that the flexible scaling field is not set on an upgraded cluster if it was not set in the earlier version.

OdedViner: Clusters with flexibleScaling disabled will not have this field in the yaml. Will is_flexible_scaling be set to false in such cases? --> yes
`.get("flexibleScaling", False)` if flexibleScaling param doesn't exist is_flexible_scaling=False

If this is meant to target the upgrade, the test should also ensure that the flexible scaling field is not set on an upgraded cluster if it was not set in the earlier version. --> verify if flexibleScaling= True the failure domain is host (on upgrade and regular deployment)

nbalacha: > If this is meant to target the upgrade, the test should also ensure that the flexible scaling field is not set on an upgraded cluster if it was not set in the earlier version. --> verify if flexibleScaling= True the failure domain is host (on upgrade and regular deployment)

What I mean is suppose the cluster did not have flexibleScaling = true before upgrade, it should not have flexibleScaling = true after upgrade. Is that being tested ?

petr-balogh (approved):

dahorak (approved):

PR #4191, opened 2021-04-25
Enable UI tests on OCP4.8

OdedViner:

### DIFF
--- a/ocs_ci/ocs/ui/views.py
+++ b/ocs_ci/ocs/ui/views.py
@@ -171,6 +171,13 @@ infra = {
 }
 
 locators = {
+    "4.8": {
+        "login": login,
+        "page": page_nav,
+        "deployment": deployment_4_7,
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/ui/views.py
```diff
@@ -171,6 +171,13 @@
 }
 
 locators = {
+    "4.8": {
+        "login": login,
+        "page": page_nav,
+        "deployment": deployment_4_7,
```
ebondare, line 177: Does this line mean that all the locators for 4.8 deployment are the same as 4.7 locators? Do we then need this special 4.8 dictionary?

OdedViner: @ebenahar I tested lso cluster +Internal cluster on OCS4.8 and OCP 4.8,and the locators for 4.8 deployment are the same as 4.7

PR #4194, opened 2021-04-26
Adding cache dropper pod to performance tests

Avilir:
This PR Add a cache dropper pod to the benchmark-operator based test, so the cache can be drop off between test samples for consistency results

this is supported and will be implemented in the FIO and SmallFiles tests


Avilir:
open by mistake

PR #4195, opened 2021-04-26
Add cache-dropper pod to the benchmark-operator base tests

Avilir:
This PR adds a cache dropper pod the benchmark-operator base tests (FIO / Smallfiles)

this will enable the benchmarks to drop off the kernel cache between samples for consistent results.

this feature supported by the FIO and Smallfiles benchmark and will integrate into them (updating the CRD files) after this PR will be merged.

### DIFF
--- a/ocs_ci/ocs/ripsaw.py
+++ b/ocs_ci/ocs/ripsaw.py
@@ -11,7 +11,8 @@ from ocs_ci.ocs.ocp import switch_to_default_rook_cluster_project
 from subprocess import run, CalledProcessError
 from ocs_ci.utility.utils import run_cmd
 from ocs_ci.ocs.constants import RIPSAW_NAMESPACE
-
+from ocs_ci.ocs.node import get_nodes
+from ocs_ci.helpers import helpers
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


ramkiperiy:
@Avilir @keesturam General question, can't we use/create unique name for namespace ? I see in most of ripsaw tests we have my-ripsaw namespace created and deleted in teardown.
Observed a problem in one of the test execution, my-ripsaw namespace failed to be deleted and subsequent tests failed that namesapce already exists.

keesturam (approved):
Looks good Avi
Did we see any change in results after introducing dropping cache?

keesturam:
> @Avilir @keesturam General question, can't we use/create unique name for namespace ? I see in most of ripsaw tests we have my-ripsaw namespace created and deleted in teardown.
> Observed a problem in one of the test execution, my-ripsaw namespace failed to be deleted and subsequent tests failed that namesapce already exists. Don't think so this is proper PR to address this.

how often do we see this issue?
my-ripsaw project not getting deleted could be an underlying issue in the cluster. I'm not sure if creating a new project is a good idea.

ramkiperiy (approved):

ramkiperiy:
> > @Avilir @keesturam General question, can't we use/create unique name for namespace ? I see in most of ripsaw tests we have my-ripsaw namespace created and deleted in teardown.
> > Observed a problem in one of the test execution, my-ripsaw namespace failed to be deleted and subsequent tests failed that namesapce already exists. Don't think so this is proper PR to address this.
> 
> how often do we see this issue?
> my-ripsaw project not getting deleted could be an underlying issue in the cluster. I'm not sure if creating a new project is a good idea.

During a testing I observed this problem twice out of 4 runs but still good to add more validation to remove my-ripsaw namespace before teardown.

Issue #4129, opened 2021-04-13
Noobaa autoscale failed to validate autoscale endpoint count


tiffanyn108:
Full log: 
https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/398/testReport/junit/tests.e2e.scale.noobaa.test_scale_endpoint/TestScaleEndpointAutoScale/test_scale_endpoint_and_respin_ceph_pods_mds_/

Error log:
# Validate autoscale endpoint count
>       self._assert_endpoint_count(desired_count=2)

tests/e2e/scale/noobaa/test_scale_endpoint.py:90: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/e2e/scale/noobaa/test_scale_endpoint.py:65: in _assert_endpoint_count
    assert pod.wait_for_resource(
ocs_ci/ocs/ocp.py:648: in wait_for_resource
    raise (ex)
ocs_ci/ocs/ocp.py:538: in wait_for_resource
    for sample in TimeoutSampler(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ocs_ci.utility.utils.TimeoutSampler object at 0x7fed06313f40>

    def __iter__(self):
        if self.start_time is None:
            self.start_time = time.time()
        while True:
            self.last_sample_time = time.time()
            if self.timeout <= (self.last_sample_time - self.start_time):
>               raise self.timeout_exc_cls(*self.timeout_exc_args)
E               ocs_ci.ocs.exceptions.TimeoutExpiredError: Timed out after 500s running get("", True, "noobaa-s3=noobaa")

ocs_ci/utility/utils.py:989: TimeoutExpiredError

ramkiperiy:
@tiffanyn108 its duplicate of issue https://github.com/red-hat-storage/ocs-ci/issues/4044, please close either of the one.

tiffanyn108:
[...]

PR #4225, opened 2021-04-29
Add additonal FIO workload to trigger Noobaa autoscale 

tiffanyn108:
Creating 10 FIO pods running in parallel to generate enough load to trigger Noobaa autoscale feature.
Fixing issue: https://github.com/red-hat-storage/ocs-ci/issues/4129

### DIFF
--- a/tests/e2e/scale/noobaa/test_scale_endpoint.py
+++ b/tests/e2e/scale/noobaa/test_scale_endpoint.py
@@ -99,8 +99,8 @@ class TestScaleEndpointAutoScale(MCGTest):
 
         # Delete mcg_job_factory
         for i in range(10):
-            f"job{i}".delete()
-            f"job{i}".ocp.wait_for_delete(resource_name=f"job{i}".name, timeout=60)
+            exec(f"job{i}.delete()")
+            exec(f"job{i}.ocp.wait_for_delete(resource_name=job{i}.name, timeout=60)")
 
         # Validate autoscale endpoint count
         self._assert_endpoint_count(desired_count=1)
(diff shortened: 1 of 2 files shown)


tests/e2e/scale/noobaa/test_scale_endpoint.py
```diff
@@ -84,7 +84,8 @@ def test_scale_endpoint_and_respin_ceph_pods(
         self._assert_endpoint_count(desired_count=1)
 
         # Create s3 workload using mcg_job_factory
-        job = mcg_job_factory(custom_options=options)
+        for i in range(10):
```
ramkiperiy, line 87: I tried similar code during noobaa db upgrade tests, please make sure test it once.
When I did this way buckets were created but object creations were slow might be due to lot of obj write to multiple buckets at same time.
Also ended up spending too much time for completion of obj write.
Would request to test this and if required sleep might work to avoid the problem

Avilir (approved):
LGTM

tests/e2e/scale/noobaa/test_scale_endpoint.py
tiffanyn108: I have tested this locally and it works fine. For Noobaa autoscale testing, the main thing is generating a high number of workload to ensure that nb endpoint CPU usage goes up to >80% for a period of time to trigger scale up (2 pods), then fio can be terminated to reduce the load so nb endpoint will be scaled down (back to 1 pod).  We don't need to wait for all FIO jobs to complete.

ramkiperiy (approved):

tests/e2e/scale/noobaa/test_scale_endpoint.py
```diff
@@ -84,7 +84,8 @@ def test_scale_endpoint_and_respin_ceph_pods(
...
+            exec(f"job{i} = mcg_job_factory(custom_options=options)")
```
keesturam, line 88: @tiffanyn108  Maybe we need a wrapper method to simply call the job 'n' times and we could use this method at multiple places depending on how we want to load the cluster.

tiffanyn108: This can be enhanced in mcg_job_factory() function which including the option to create multiple FIO pods.
It can be done in a separate PR.

keesturam (approved):

PR #4237, opened 2021-04-30
Increase CLUSTER_NAME_MAX_CHARACTERS value

deepshikhaaa:
OCP deployment fails with ClusterNameLengthError when cluster name
, for instance is 'dkhandel-ocsci-371' which is 18 characters long
while 5-17 characters long is expected

Signed-off-by: Deepshikha Khandelwal <dkhandel@redhat.com>


zmc (commented):
I just want to point out that https://github.com/red-hat-storage/ocs-ci/issues/1133#issuecomment-562930520 suggests this might not be a good idea, unfortunately

PR #4247, opened 2021-05-03
Fix RGW tests to run on external-mode clusters

MeridianExplorer:
Currently, we only check for whether there are RGW deployments on the cluster. In external mode clusters, the deployments are not found on the same cluster, and thus another check for the external RGW storageclass was added, to let the tests run on external-mode clusters.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1739,7 +1739,16 @@ def rgw_obj_fixture(request):
     rgw_deployments = get_deployments_having_label(
         label=constants.RGW_APP_LABEL, namespace=config.ENV_DATA["cluster_namespace"]
     )
-    if rgw_deployments:
+    try:
+        storageclass = OCP(
+            kind=constants.STORAGECLASS,
+            namespace=config.ENV_DATA["cluster_namespace"],
+            resource_name=constants.DEFAULT_EXTERNAL_MODE_STORAGECLASS_RGW,
+        )
+    except CommandFailed:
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -1739,7 +1739,16 @@ def rgw_obj_fixture(request):
     rgw_deployments = get_deployments_having_label(
         label=constants.RGW_APP_LABEL, namespace=config.ENV_DATA["cluster_namespace"]
     )
-    if rgw_deployments:
+    try:
+        storageclass = OCP(
```
fbalak, line 1743: [Constructor of OCP](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/ocp.py#L42) doesn't call any oc command. There should be added: `storageclass.get()` at the end of `try` block.

MeridianExplorer: Right! Missed it. Thanks

fbalak (approved):

bernerhat (approved):

parikshithb (approved):

MeridianExplorer:
The PR is still relevant, and pending verification

mbukatov (approved):

MeridianExplorer:
Skipping verification as discussed with @ebenahar, since no external mode cluster is available at the moment, and the PR doesn't seem to be able to hurt any other part of the test suite in an obvious way.

PR #4252, opened 2021-05-04
Extend the node replacement verification steps to work with more than 3 OSDs

yitzhak12:
Currently, the node replacement verification steps are working. Still, when we have many OSD's(for example, after adding capacity to the cluster), sometimes the verification steps can fail. So we need to extend the verification steps to work on any cluster - no matter the number of the OSD's and worker nodes.

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -1109,7 +1109,7 @@ def get_worker_nodes_not_in_ocs():
 
 
 def node_replacement_verification_steps_user_side(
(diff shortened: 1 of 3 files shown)


ebenahar (approved):

sidhant-agrawal (approved):

ebenahar (approved):

sidhant-agrawal (approved):

annagitel (approved):
LGTM

Issue #4231, opened 2021-04-29
Unavailable OCP version URL results in cluster fixture failures for all tests


clacroix12:
[expose_ocp_version](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L553) causes the [cluster](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/conftest.py#L1109) session scope fixture to fail for every test in the suite if the URL where we retrieve the version information from is unavailable. You can see this in the following jenkins execution:

Jenkins: https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/541/
Logs: http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/j018ai3c33-p/j018ai3c33-p_20210423T170302/logs/

This is happening because we are now using `expose_ocp_version` within [get_openshift_client](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L681) to convert the configured version to a full legitimate version string. We can make get_openshift_client more robust by excepting the `AssertionError` raised by `expose_ocp_version` and skip over the version comparison.

PR #4262, opened 2021-05-05
Skip version comparison and download if unable to expose the OCP version

clacroix12:
Fixes: #4231 

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -679,9 +679,7 @@ def get_ocm_cli(
 
 
 def get_openshift_client(
-    version=None,
-    bin_dir=None,
-    force_download=False,
+    version=None, bin_dir=None, force_download=False, skip_comparison=False
 ):
     """
     Download the OpenShift client binary, if not already present.
@@ -692,22 +690,30 @@ def get_openshift_client(
             (default: config.RUN['client_version'])
         bin_dir (str): Path to bin directory (default: config.RUN['bin_dir'])
         force_download (bool): Force client download even if already present
+        skip_comparison (bool): Skip the comparison between the existing OCP client
+            version and the configured one.
 
     Returns:
         str: Path to the client binary
 
     """
     version = version or config.RUN["client_version"]
-    version = expose_ocp_version(version)
     bin_dir = os.path.expanduser(bin_dir or config.RUN["bin_dir"])
     client_binary_path = os.path.join(bin_dir, "oc")
     kubectl_binary_path = os.path.join(bin_dir, "kubectl")
     download_client = True
     client_version = None
+    try:
+        version = expose_ocp_version(version)
+    except Exception:
+        log.exception("Unable to expose OCP version, skipping client download.")
+        skip_comparison = True
+        download_client = False
+        force_download = False
 
     if force_download:
         log.info("Forcing client download.")
-    elif os.path.isfile(client_binary_path):
+    elif os.path.isfile(client_binary_path) and not skip_comparison:
         current_client_version = get_client_version(client_binary_path)
         if current_client_version != version:
             log.info(
@@ -770,7 +776,6 @@ def get_openshift_client(
         os.chdir(previous_dir)
 
     log.info(f"OpenShift Client version: {client_version}")
-
     return client_binary_path
 
 


clacroix12:
@petr-balogh I've added logic to `get_openshift_client` to allow for skipping the version comparison. This is done automatically if we run into issues exposing the OCP version. I elected to also skip the download entirely if we encounter this scenario since the we pull the version info and the actual client from the same host. If we only want to skip the version comparison and continue trying the download if we can I can change this logic to support that. However, I figured we would most likely be encountering issues with that step anyways so it might be better to gracefully skip it. Let me know what you think.

ocs_ci/utility/utils.py
```diff
@@ -692,22 +690,30 @@ def get_openshift_client(
...
     version = version or config.RUN["client_version"]
-    version = expose_ocp_version(version)
     bin_dir = os.path.expanduser(bin_dir or config.RUN["bin_dir"])
     client_binary_path = os.path.join(bin_dir, "oc")
     kubectl_binary_path = os.path.join(bin_dir, "kubectl")
     download_client = True
     client_version = None
+    try:
+        version = expose_ocp_version(version)
+    except AssertionError:
```
petr-balogh, line 708: What about to have general Exception here? In the case the service will be in bad shape and will not return the proper json but some other output we will hit the same issue. 

I see that that get_url_content is raising the AssertionError but then we do json.loads so it can error also there.
So I am for make it more robust with handle Exception, we do log warning. Maybe you can print the exception message here to have all the details written in that warning.

petr-balogh:
> @petr-balogh I've added logic to `get_openshift_client` to allow for skipping the version comparison. This is done automatically if we run into issues exposing the OCP version. I elected to also skip the download entirely if we encounter this scenario since the we pull the version info and the actual client from the same host. If we only want to skip the version comparison and continue trying the download if we can I can change this logic to support that. However, I figured we would most likely be encountering issues with that step anyways so it might be better to gracefully skip it. Let me know what you think.

Looks ok, just one comment about general Exception and also adding it to warning log message.
Thanks

ocs_ci/utility/utils.py
clacroix12: I have broadened to `Exception` here as well as changed from logging a warning to logging the original exception that caused this to fail.

mbukatov (commented):
Looks ok.

ocs_ci/utility/utils.py
```diff
@@ -692,22 +690,30 @@ def get_openshift_client(
...
-    version = expose_ocp_version(version)
     bin_dir = os.path.expanduser(bin_dir or config.RUN["bin_dir"])
     client_binary_path = os.path.join(bin_dir, "oc")
     kubectl_binary_path = os.path.join(bin_dir, "kubectl")
     download_client = True
     client_version = None
+    try:
+        version = expose_ocp_version(version)
+    except Exception:
+        log.exception("Unable to expose OCP version, skipping client download.")
```
petr-balogh, line 709:
```suggestion
    except Exception as ex:
        log.exception(f"Unable to expose OCP version, skipping client download. Exception: {ex}")
```

Please log the exception. If we do the general exception here, I would like to see it in the log as I pointed out in my previous comment.

mbukatov: The [`exception()` method](https://docs.python.org/3/library/logging.html#logging.Logger.exception) will automatically include a traceback and the error in the logs, there is no need to explicitly push the exception into the message again:

>     Logs a message with level ERROR on this logger. The arguments are interpreted as for debug(). Exception info is added to the logging message. This method should only be called from an exception handler.


```
In [1]: import logging                                                                                                                  

In [2]: try: 
   ...:     print(1/0) 
   ...: except: 
   ...:     logging.exception("ops") 
   ...:                                                                                                                                 
ERROR:root:ops
  File "<ipython-input-2-2ba5b913f0db>", line 2, in <module>
    print(1/0)
ZeroDivisionError: division by zero

In [3]:
```

petr-balogh: Ah, I didn't realized and missed it's log.exception. I thought it was warning level in some of past comment hence I wanted to have this logged out. Sorry for confusion.

petr-balogh (approved):

shylesh (approved):
LGTM

PR #4264, opened 2021-05-06
fix noobaa cli version in upgrade tests

fbalak:
- fixes https://github.com/red-hat-storage/ocs-ci/issues/4171
- Log noobaa cli version

### DIFF
--- a/ocs_ci/ocs/resources/mcg.py
+++ b/ocs_ci/ocs/resources/mcg.py
@@ -968,6 +968,8 @@ class MCG:
             assert (
                 _compare_cli_hashes()
             ), "Binary hash doesn't match the one on the operator pod"
+        logger.info("Retrieving mcg cli version")
+        self.exec_mcg_cmd("version")
 
     @property
     def status(self):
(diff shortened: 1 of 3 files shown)


ocs_ci/ocs/resources/mcg.py
```diff
@@ -968,6 +968,8 @@ def _compare_cli_hashes():
             assert (
                 _compare_cli_hashes()
             ), "Binary hash doesn't match the one on the operator pod"
+        logger.info("Retrieving mcg cli version")
+        self.exec_mcg_cmd("version")
```
MeridianExplorer, line 972: This version is very often not sufficient; It only changes once or twice across an entire version.

The info that's critical to log is the contents of `commit_hash`, in the root of the NB operator pod - which contains the exact version that can be later matched with a particular binary for debugging purposes

MeridianExplorer (approved):

ocs_ci/ocs/resources/mcg.py
MeridianExplorer, line 972: This version is very often not sufficient; It only changes once or twice across an entire version.

The info that's critical to log is the contents of `commit_hash`, in the root of the NB operator pod - which contains the exact version that can be later matched with a particular binary for debugging purposes

MeridianExplorer (changes requested):

PR #4277, opened 2021-05-11
Implement batch creation for NSS and BS factories

### LINKED ISSUE #4248: NSS and BS factories ignore amount parameter
when inspecting the tuple representing the amount and region of the requested NSS/BS we are currently ignoring the amount requirement (first field of the tuple), easily fixed by looping over the logic with the amount as the range indicator

MeridianExplorer:
Up until now, the factories simply disregarded the amount that was passed in the tuple, and only created one BS/NSS.
This PR fixes the problem and creates the requested amount of BS/NSS.

Closes #4248

The diff is showing very complex changes, but all I did was add `for _ in range(uls_tup[0]):` to both factories, and indent all the logic that follows.

### DIFF
--- a/ocs_ci/ocs/resources/backingstore.py
+++ b/ocs_ci/ocs/resources/backingstore.py
@@ -228,46 +228,22 @@ def backingstore_factory(request, cld_mgr, mcg_obj, cloud_uls_factory):
             )
         for cloud, uls_lst in uls_dict.items():
             for uls_tup in uls_lst:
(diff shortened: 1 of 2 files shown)


bernerhat (approved):

ebenahar (approved):

MeridianExplorer:
Apparently, this was already handled in our code. This fix has led to n² creations instead.

Issue #4280, opened 2021-05-12
Disconnected cluster info is loaded after cluster_load is triggered


dahorak:
All OCS Upgrade on disconnected cluster jobs have following failures:

```
19:22:22 - MainThread - tests.conftest - ERROR - Cluster load might not work correctly during this run, because it failed with an exception: 'mirror_registry'
```

The reason seems to be, that the information about Disconnected cluster configuration (fixture `load_cluster_info_file`) is loaded after `cluster_load`:

```
19:22:22 - MainThread - tests.conftest - ERROR - Cluster load might not work correctly during this run, because it failed with an exception: 'mirror_registry'
19:22:22 - MainThread - root - INFO - File /home/jenkins/current-cluster-dir/openshift-cluster-dir/cluster_info.json exists, loading relevant info.
19:22:22 - MainThread - root - INFO - Loading configuration for disconnected cluster
```

PR #4281, opened 2021-05-12
run load_cluster_info_file before cluster_load

dahorak:
- fixes https://github.com/red-hat-storage/ocs-ci/issues/4280

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -42,6 +42,7 @@ from ocs_ci.ocs.exceptions import (
     UnexpectedImage,
     UnsupportedOSType,
 )
+from ocs_ci.utility.flexy import load_cluster_info
 from ocs_ci.utility.retry import retry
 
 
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -1190,6 +1190,16 @@ def log_cli_level(pytestconfig):
     return pytestconfig.getini("log_cli_level") or "DEBUG"
 
 
+@pytest.fixture(scope="session", autouse=True)
+def load_cluster_info_file(request):
```
petr-balogh, line 1194: If you would like to make sure that this fixture is called before cluster load you should use load_cluster_info_file as a parameter of cluster_load fixture otherwise we will not be sure in which order pytest will load the fixture.

dahorak: Thanks for the note, I reconsidered it and explicitly run the `load_cluster_info` function on the beginning of `login_to_mirror_registry` function (where it was failing because of missing data).

jimdevops19 (approved):
LGTM

vavuthu (approved):

ebenahar (approved):

PR #4285, opened 2021-05-12
WIP: Add implementation for adding nodes and storage for IBM Power

gitsridhar:
This is an effort to implement dynamic add nodes and lead to provide
additional capacity to ocs-ci test cases for IBM Power platform. An internal
tool called ocpdr is used to add the desired number of nodes to the underlying
OCP cluster.


ocs_ci/ocs/node.py
```diff
@@ -957,6 +965,96 @@ def delete_and_create_osd_node_vsphere_upi(osd_node_name, use_existing_node=Fals
     return new_node_name
 
 
+def delete_and_create_osd_node_ibm_power_upi(osd_node_name, use_existing_node=False):
```
yitzhak12, line 968: This function is the same as `delete_and_create_osd_node_vsphere_upi_lso`. So you can use it here or can export some of the code to make it more generic. If there are some special changes for IBM and the code can be different(now or in the future), we don't have to export the code and leave it as it is now.

ocs_ci/ocs/node.py
```diff
@@ -384,7 +384,15 @@ def add_new_node_and_label_upi(
...
     node_util.create_and_attach_nodes_to_cluster(node_conf, node_type, num_nodes)
-    for sample in TimeoutSampler(timeout=600, sleep=6, func=get_worker_nodes):
+
+    addnodetimeout = 600
+
+    if config.ENV_DATA["platform"].lower() == constants.IBM_POWER_PLATFORM:
+        addnodetimeout = 1200
+
+    for sample in TimeoutSampler(
+        timeout=addnodetimeout, sleep=6, func=get_worker_nodes
```
yitzhak12, line 394: I didn't think of it before... but why the time to sleep here is only 6 seconds? If we wait 600 or 1200 seconds, I thought it would be 20 or even 30 seconds.

gitsridhar: @yitzhak12 The code I am proposing is just doubling the timeout (600 to 1200) if the environment is IBM _POWER_PLATFORM. Is there an issue with the original code itself? Then I can fix it, let me know.

ocs_ci/ocs/node.py
gitsridhar: @yitzhak12 I borrowed code from delete_and_create_osd_node_vsphere_upi_lso. This method does quite a few things specific to vSphere. 1. It has facilities to support the usage of exiting nodes. 2. It adds a disk when new node is created. 3. Validation of the number of pvs, it is straightforward.

In the case of IBM Power, we will always add a node. When a node is created, a disk is created automatically. The number of pv validation is through a configurable value. 

Due to these reasons, I would like to retain the code as is. There are quite a few differences.

yitzhak12: Okay, np. Whatever you feel comfortable with.

ocs_ci/ocs/node.py
yitzhak12: Sorry for the late response. Yes, I know that you took the timeout from the previous code - That's why I mentioned that I didn't think of it before. Maybe we need to ask another one in the relevant field about the matter. I think that sleep for 20 or 30 seconds when we wait for 600 seconds or more would be enough. But maybe there was a reason for waiting only 6 seconds between the iterations.

yitzhak12:
@gitsridhar I will be happy if you can wait a few more days until I finish with my PR https://github.com/red-hat-storage/ocs-ci/pull/4468. Because I see we are working on the same files, and I think when I finish with the PR, it will be also easier for you to implement your changes. Also, feel free to review my PR

gitsridhar:
@yitzhak12 Thanks. I will proceed with this PR but I am waiting for https://bugzilla.redhat.com/show_bug.cgi?id=1977121 be resolved. When count in storage cluster is increased, new osd is not getting created.

yitzhak12:
@gitsridhar, my PR https://github.com/red-hat-storage/ocs-ci/pull/4468 is merged. I changed the node replacement process to work with multiple PVs and multiple OSDs per node. You can see the relevant changes in the PR.

petr-balogh:
@gitsridhar have you seen Itzhak comment?  This needs re-work + rebase for sure.

PR #4290, opened 2021-05-13
WIP: Update the "add_capacity" tests

yitzhak12:
According to the last failures of the "add_capacity" tests, we need to perform some changes in the test:

1. Increase timeout when running IO in the background in the "add_capacity" test.
2. Label the OCS worker nodes before the "add_capacity" the test `test_add_capacity_with_resource_delete` starts.
3. If we use the vSphere platform, we might need to add 3 extra worker nodes and label them before the test `test_add_capacity_with_resource_delete` starts.

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -1556,11 +1556,11 @@ def check_ceph_health_after_add_capacity():
     ceph_health_tries = 80
     ceph_rebalance_timeout = 1800
     if config.RUN.get("io_in_bg") and config.RUN.get("io_load"):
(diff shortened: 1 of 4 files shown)


ocs_ci/ocs/cluster.py
```diff
@@ -1551,6 +1552,25 @@ def is_flexible_scaling_enabled():
     return failure_domain == "host" and flexible_scaling
 
 
+def check_ceph_health_after_add_capacity():
```
prsurve, line 1555: make sure you use ceph_health_tries as a parameter data we might need more in some cases

yitzhak12: Okay, I will add it to the function.

PR #4292, opened 2021-05-14
OCS4.6/4.8 Installation via UI

OdedViner:

### DIFF
--- a/ocs_ci/ocs/ui/views.py
+++ b/ocs_ci/ocs/ui/views.py
@@ -197,6 +197,11 @@ infra = {
 }
 
 locators = {
+    "4.8": {
+        "login": login,
+        "page": page_nav,
+        "deployment": {**deployment, **deployment_4_7},
+    },
     "4.7": {
         "login": login,
         "page": page_nav,
--- a/ocs_ci/ocs/ui/views.py
+++ b/ocs_ci/ocs/ui/views.py
@@ -9,9 +9,8 @@ login = {
     "click_login": ("//button[text()='Log in']", By.XPATH),
 }
 
-deployment_4_7 = {
+deployment = {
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/ui/views.py
```diff
@@ -172,16 +197,22 @@
 }
 
 locators = {
+    "4.8": {
+        "login": login,
+        "page": page_nav,
+        "deployment": {**deployment, **deployment_4_7},
```
ebenahar, line 203: just to be sure, is `**deployment_4_7` intentional here?

OdedViner: yes. 
There is no difference between OCS4.8 and OCS4.7  installation [only MULTUS option]

ebenahar (approved):

ocs_ci/ocs/ui/views.py
ebondare, line 203: I'd add pvc and infra dictionaries here as well, just to make sure that all the tests that can pass without any changes would pass.

OdedViner: https://github.com/red-hat-storage/ocs-ci/blob/f2aadb01fcf596a0736caf8672d267da49399fd6/ocs_ci/ocs/ui/views.py#L174
there is not conflict on this pr.. I added a new key [4.8]

ebondare: Just for anyone who might also be interested, we discussed this issue with Oded separately and he explained that pvc tests are failing for 4.8 so they require a separate effort and thus the pvc dictionary was not included in this PR. As for infra dictionary, it needs to be tested as well.

ebondare:
LGTM

ebondare (approved):

Issue #4306, opened 2021-05-18
failed to get ignition_version during deployment


vavuthu:
Deployment on vsphere failed to get ignition version ( run-ci without ocs-version parameter )

```
>           ] = config.DEPLOYMENT["ignition_version"]
E           KeyError: 'ignition_version'

ocs_ci/utility/utils.py:3336: KeyError
```

PR #4309, opened 2021-05-18
Include ignition_version in default_config.yaml

vavuthu:
Fixes: #4306 


prsurve:
@vavuthu ignition verison are based on OCP version this version `3.2.0` might not work for old OCP version

vavuthu:
> @vavuthu ignition verison are based on OCP version this version `3.2.0` might not work for old OCP version

yes, if you want to use old ocp version you need to use parameter so that appropriate ignition version is loaded.

clacroix12 (approved):

jilju (approved):

PR #4330, opened 2021-05-24
Push pvc_create_delete perf. results into ES server

Avilir:
In this PR the results of the PVC_creation_deletion performance results are push into the ElasticSearch server, so we can display them in the QPAS Performance Dashboard.

### DIFF
--- a/tests/e2e/performance/test_pvc_creation_deletion_performance.py
+++ b/tests/e2e/performance/test_pvc_creation_deletion_performance.py
@@ -1,6 +1,7 @@
 """
 Test to verify PVC deletion performance
 """
+import time
 import logging
 import datetime
 import pytest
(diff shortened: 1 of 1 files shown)


vasukulkarni (approved):

tests/e2e/performance/test_pvc_creation_deletion_performance.py
```diff
@@ -1,30 +1,87 @@
 """
 Test to verify PVC deletion performance
```
ypersky1980, line 2: Please change to : Test to verify PVC creation and deletion performance

Avilir: will do in next PR with some test modifications

ypersky1980 (approved):

PR #4333, opened 2021-05-24
Remove cpu and memory limits for ceph pods

lukebrowning:
vcpus are platform specific:
-  1 vcpu is 1/2 of a processor core for x86
-  1 vcpu is 1/8 of a processor core for power

The ocs-ci suite includes a performance test and limits seem contrary to the goal
of measuring performance.  These limits would have to be scaled to make sense for
power.  But, is it desirable to have limits on a storage resource?  A limit on
storage performance is an indirect limit on workload performance.

The UI deployment of ocs does not impose cpu and memory limits.  Shouldn't ocs-ci
emulate customer configuration?   I don't think it is needed or desirable.

By default, the linux scheduler and virtual memory management provide fairness.
The use of limits makes sense when a subsystem is cpu intensive or memory intensive,
but ceph is neither.

Signed-off-by: Luke Browning <lukebrowning@us.ibm.com>

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -635,11 +635,6 @@ class Deployment(object):
                 and not lso_type == constants.AWS_EBS
             ):
                 deviceset_data["count"] = 2
-            if ocs_version >= 4.5:
-                deviceset_data["resources"] = {
-                    "limits": {"cpu": 2, "memory": "5Gi"},
-                    "requests": {"cpu": 1, "memory": "5Gi"},
-                }
             if (ocp_version >= 4.6) and (ocs_version >= 4.6):
                 cluster_data["metadata"]["annotations"] = {
                     "cluster.ocs.openshift.io/local-devices": "true"


vasukulkarni:
The code is triggered only when  "allow_lower_instance_requirements" is set to True in config.

vasukulkarni:
Example config: https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/aws_lower_resource_reqs.yaml#L3

lukebrowning:
@vasukulkarni There are two instances of setting limits in the code.   One is under allow_lower_* and the other is not.   I removed the latter.

lukebrowning:
What is the status of the reviews?   This negatively impacts power servers as the cpu limit is not scaled for the platform.

vasukulkarni (approved):

lukebrowning:
I have tested this change.

lukebrowning:
@vasukulkarni Are you looking for logs?   OCS deploys and we have run tier tests.

vasukulkarni:
@lukebrowning  I will be running on x86 as well on vsphere to check for regression and asked few other folks to jump in and provide context before merging. If we want to merge this sooner, maybe you can selectively enable for ibm platform ?

lukebrowning:
I'll wait.  Thanks

keesturam:
@lukebrowning what is the limit for OSD pods in IBM platforms? is it any different from the default ones we have for OCS? (5GB ram & 2 cpu) - providing more resources will sure improve performance, but we should be looking at what customers would use out of the box. so, maybe this needs a fix in OCS?

vasukulkarni:
@ebenahar can you please check one of the backingstore test failure in the above run and let me know if its ok to ignore?

ebenahar:
> @ebenahar can you please check one of the backingstore test failure in the above run and let me know if its ok to ignore?

@MeridianExplorer could you please take a look?

ocs_ci/deployment/deployment.py
```diff
@@ -635,11 +635,6 @@ def deploy_ocs_via_operator(self, image=None):
                 and not lso_type == constants.AWS_EBS
             ):
                 deviceset_data["count"] = 2
-            if ocs_version >= 4.5:
```
vasukulkarni, line 638: @lukebrowning you can add ```if ocs_version >= 4.5 and config.ENV_DATA["platform"] != constants.IBMCLOUD_PLATFORM``` we can move forward with this change faster, in general we dont need the if but until we get some tier1 failure sorted out we can have it for now.

nehaberry:
Shouldn't ocs-ci emulate customer configuration? I don't think it is needed or desirable.--> IIUC this limit is supposed to be set only for AWS I3 based deployments which do not have enough CPUs , hence we need to fall-back to minimal deployment
 
 We had added this based on the changes in the [deployment guides](https://access.redhat.com/documentation/en-us/red_hat_openshift_container_storage/4.6/html-single/deploying_openshift_container_storage_using_amazon_web_services/index#creating-openshift-container-storage-cluster-on-amazon-ec2_local-storage)
 
 We do not need this setting for any other normal LSO based deployments.

Also, with OCP 4.6 and above, even UI based install on AWS I3 will set this resource limit (due to fix of https://bugzilla.redhat.com/show_bug.cgi?id=1886541)

clacroix12 (approved):
Approving since I will be opening a follow up re-adding these limits for aws i3 deployments

ocs_ci/deployment/deployment.py
vasukulkarni: not needed, resolving.

ebenahar (approved):

PR #4345, opened 2021-05-27
Add --re-trigger-failed-tests parameter

petr-balogh:
This new parameter allows to re-trigger only failed tests from the
previous execution provided as path to junit xml file.

### DIFF
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -10,6 +10,7 @@ import logging
 import os
 
 import pytest
+from junitparser import JUnitXml
 
 from ocs_ci.framework import config as ocsci_config
 from ocs_ci.framework.exceptions import (
@@ -254,6 +255,16 @@ def pytest_addoption(parser):
         action="append",
         choices=["rgw", "cephfs", "noobaa", "blockpools"],
         help=("disable deployment of ocs component:rgw, cephfs, noobaa, blockpools."),
+    )
+    parser.addoption(
+        "--re-trigger-failed-tests",
+        dest="re_trigger_failed_tests",
+        help="""
+        Path to the xunit file for xml junit report from the previous execution.
+        If the file is provided, the execution will remove all the test cases
+        which passed and will run only those test cases which were skipped /
+        failed / or had error in the provided report.
+        """,
     )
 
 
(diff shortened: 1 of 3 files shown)


ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -513,13 +524,32 @@ def process_cluster_cli_params(config):
...
     Add Polarion ID property to test cases that are marked with one.
     """
-    for item in items:
+
+    re_trigger_failed_tests = ocsci_config.RUN.get("re_trigger_failed_tests")
+    if re_trigger_failed_tests:
+        junit_report = JUnitXml.fromfile(re_trigger_failed_tests)
+        cases_to_re_trigger = []
+        for suite in junit_report:
+            cases_to_re_trigger += [_case.name for _case in suite if _case.result]
```
vasukulkarni, line 544: anything other than pass will be executed based on if condition? (that is it includes skipped)

petr-balogh: We want to include skipped tests as if cluster is not healthy because some test broke the cluster the rest of the tests are skipped, those we would like to trigger again. If the test case should be skipped because some skipif in other code condition, those will be still skipped also in the new execution.

Avilir (approved):

petr-balogh:
Running verification job here:  https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-trigger-vsphere-upi-encryption-1az-rhcos-vsan-lso-vmdk-3m-3w-tier1/46

dahorak (approved):

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -513,13 +524,32 @@ def process_cluster_cli_params(config):
...
+    for item in items[:]:
```
jimdevops19, line 545: Why do we need a new copy of the items list and not just removing it from the original items? And also where is this new copy of the items list will be stored if it is not assigned to a variable like `duplicate_list = items[:]`

jimdevops19 (approved):

ocs_ci/framework/pytest_customization/ocscilib.py
petr-balogh: @jimdevops19 this is one of the bad think you can do. You shouldn't delete the item from the list you are iterating over as it can leads to weird situations like you will miss some of the item and so on..  e.g. this article:  https://thispointer.com/python-remove-elements-from-a-list-while-iterating/  so this is the safe way how to iterate over all values and do not change iterator .  You can find many more articles on this problematic , this is one of the first I just googled.

PR #4370, opened 2021-06-03
io - Remove performance marker

Avilir:
Removing the performance marker from the basic IO test since it is not performance test, just an I/O functionality test


keesturam:
@vasukulkarni @petr-balogh what would be the best way to retire a test?

keesturam (approved):

ramkiperiy (approved):

Issue #4373, opened 2021-06-04
vSphereIPI teardown is calling aws destroy method instead of its own


vavuthu:
vSphereIPI teardown is calling aws destroy method instead of its own

PR #4374, opened 2021-06-04
destroy_cluster and deploy_ocp methods of vSphereIPI should be under its own class

vavuthu:
Fixes: #4373

destroy_cluster and deploy_ocp methods of vSphere IPI should be under VSPHEREIPI class

Description: During teardown of cluster for vSphere IPI, destroy method
is calling the aws destroy method instead of it own


petr-balogh (dismissed):

divyashree15 (approved):

shylesh (approved):
LGTM

petr-balogh (approved):

PR #4383, opened 2021-06-06
Fix delete pv when it is not bounded

Avilir:
When PV does not belong to any namespace the command will faild, so i just wrap it with 
try - except block.

### DIFF
--- a/tests/e2e/performance/test_fio_benchmark.py
+++ b/tests/e2e/performance/test_fio_benchmark.py
@@ -345,9 +345,12 @@ class TestFIOBenchmark(PASTest):
         for line in pvs_list:
             pv, ns = line.split(" ")
             pv = pv.replace("'", "")
-            if ns == constants.RIPSAW_NAMESPACE:
-                log.info(f"Going to delete {pv}")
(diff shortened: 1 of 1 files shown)


ypersky1980 (approved):

tests/e2e/performance/test_fio_benchmark.py
```diff
@@ -345,9 +345,12 @@ def cleanup(self):
         for line in pvs_list:
             pv, ns = line.split(" ")
             pv = pv.replace("'", "")
-            if ns == constants.RIPSAW_NAMESPACE:
-                log.info(f"Going to delete {pv}")
-                run_command(f"oc delete pv {pv}")
+            try:
```
keesturam, line 348: iiuc, pv deletion doens't need to be bound by namespace. 
in any case, wouldn't this could simply be 
`oc delete pv {pv} -n constants.RIPSAW_NAMESPACE`

keesturam (dismissed):

keesturam (approved):

ramkiperiy (approved):

Issue #4387, opened 2021-06-07
Email with test results should be send even in case of failure in sessionfinish


dahorak:
When there is some failure during _sessionfinish_ stage, email with test results is not sent. This might happen for example, when the test execution is too long and the cluster is cleaned up by _AWS Cleanup Job_ - and then during the _live log sessionfinish_ phase `oc -n openshift-storage get csv  -n openshift-storage -o yaml` command fails to gather data.

This happen for example in [this job](https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/1011/console).

PR #4406, opened 2021-06-09
Ensure that email with test results is send even in case of failure in sessionfinish 

dahorak:
- this will ensure, that email with test results will be send even when
  the get_ocs_build_number() fails for example because some network
  issue or because the cluster was deleted

fixes https://github.com/red-hat-storage/ocs-ci/issues/4387

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -1375,9 +1375,9 @@ def email_reports(session):
 
     try:
         build_id = get_ocs_build_number()
-    except Exception as e:
+    except Exception:
         build_id = ""
-        log.exception(e)
+        log.exception("Getting OCS operator build number failed!")
     build_str = f"BUILD ID: {build_id} " if build_id else ""
     mailids = config.RUN["cli_params"]["email"]
     recipients = []
@@ -1409,8 +1409,8 @@ def email_reports(session):
         s.sendmail(sender, recipients, msg.as_string())
         s.quit()
         log.info(f"Results have been emailed to {recipients}")
-    except Exception as e:
-        log.exception(e)
+    except Exception:
+        log.exception("Sending email with results failed!")
 
 
 def get_cluster_version_info():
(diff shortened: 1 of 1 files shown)


ocs_ci/utility/utils.py
```diff
@@ -1373,7 +1373,11 @@ def email_reports(session):
     # total = passed + failed + error
     # percentage_passed = (passed / total) * 100
 
-    build_id = get_ocs_build_number()
+    try:
+        build_id = get_ocs_build_number()
+    except Exception as e:
+        build_id = ""
+        log.exception(e)
```
clacroix12, line 1380: I believe this will result in an empty log message followed by the stack trace. `log.exception` will automatically grab the stack trace for us so we don't need to pass `e` to it. We should instead include a more informative message about what we were doing when we hit that error.

```python
try:
    assert True == False
except AssertionError as e:
    log.exception(e)
```
```
ERROR:__main__:
  File "<ipython-input-9-28d4662577e9>", line 2, in <module>
    testing()
  File "<ipython-input-3-899b3bb10609>", line 2, in testing
    assert True == False
AssertionError
```
```python
try:
    assert True == False
except AssertionError:
    log.exception('some specific message about failing to get build number')
```
```
ERROR:__main__:some specific message about failing to get build number
  File "<ipython-input-10-c8c132c2e320>", line 2, in <module>
    testing()
  File "<ipython-input-3-899b3bb10609>", line 2, in testing
    assert True == False
AssertionError
```

dahorak: Thanks for the notice, I copied it from the following lines - will fix it there also.

dahorak: Fixed in 50315f6.

petr-balogh (approved):

vasukulkarni (approved):

clacroix12 (approved):

PR #4409, opened 2021-06-09
Cleanup old unused state deployment related code

dahorak:
Stage deployment is now done differently and this code is not used anymore, so removing it to cleanup the code.


clacroix12 (approved):

dahorak:
The last change was only removing `self` from `self.create_catalog_source(...)`.

PR #4414, opened 2021-06-09
Fix test case test_pv_creation to skip PVC and yaml file deletion based on its existence

### LINKED ISSUE #4403: Tear down error in test test_pv_creation
If the test case test_pv_creation is skipped due to degraded cluster, teardown will execute and throw the below error.
```
E ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc -n openshift-storage delete -f test.yaml.
E Error is error: the path "test.yaml" does not exist

ocs_ci/utility/utils.py:510: CommandFailed
```
The file test.yaml will be created only if the test case is executed.

jilju:
If the test case test_pv_creation is skipped due to degraded cluster, teardown will execute and try to delete the PVC and yaml file which is not actually present. Temporary yaml file will be present only if the test case is executed. 
Fixes #4403 

### DIFF
--- a/tests/manage/pv_services/test_pv_creation.py
+++ b/tests/manage/pv_services/test_pv_creation.py
@@ -40,9 +40,10 @@ def teardown(self):
     """
     Tearing down the environment
     """
-    assert delete_pv(self.pv_name)
-    assert not verify_pv_exist(self.pv_name)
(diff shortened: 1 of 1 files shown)


sidhant-agrawal (approved):

rachael-george (approved):
LGTM

PR #4436, opened 2021-06-14
Adding support for rbd thick provisioning to test_pvc_creation_deletion_performance.py

ypersky1980:
The current PR handles the following

1) Added RBD-Thick interface to the test
2) Changed tested PVC size ( from 25/50/100 GB to 5/15/25GB) to make the test run shorter , since it has been found that the PVC size does not affect the PVC creation time ( for RBD and CephFS).
The PVC size does indeed affect the RBD-Thick PVCs, therefore we run the test with 3 sizes (5/15/25 GB) to make sure that the creation time is linear. 
3) The data written to the PVCs ( prior to the deletion) was changed from 10GB to 1 GB for 2 reasons
 - to match the new PVC sizes
 - to make the test run shorter
 
4) The following markers were created ( in addition to the existing marker  "performance" :
performance_new_feature - to mark test cases related to new features
performance_regression - to mark tests that check features that existed prior to OCS 4.8

Therefore the current implementation is that if a test is run using the marker "performance" - all the test cases will run. 
If a test is run using the marker performance_new_feature - only new feature-related test cases will run.
If a test is run using the marker "performance_regression" - only RBD and CephFS test cases of ONLY test_pvc_creation_deletion.py test will run. 
If we'd like that other tests will run while using the "performance_regression" mark - we'll have to go over those tests and add this mark to relevant test cases. 

5) Accepted standerd deviation percent ( for both creation and deletion) value was increased.


tests/e2e/performance/test_pvc_creation_deletion_performance.py
```diff
@@ -83,18 +84,25 @@ def setup(self):
...
-        self.interface = interface_iterate
-        self.sc_obj = storageclass_factory(self.interface)
+        self.interface = interface_type
+        if self.interface == constants.CEPHBLOCKPOOL_THICK:
+            self.sc_obj = storageclass_factory(
+                interface=constants.CEPHBLOCKPOOL,
+                new_rbd_pool=True,
+                rbd_thick_provision=True,
+            )
+        else:
```
keesturam, line 104: `self.interface = interface_iterate` should be included in the else block to run tests with cephfs & rbd interface

tests/e2e/performance/test_pvc_creation_deletion_performance.py
```diff
@@ -115,11 +123,44 @@ def init_full_results(self, full_results):
         return full_results
 
     @pytest.mark.parametrize(
-        argnames=["pvc_size"],
+        argnames=["interface_type", "pvc_size"],
         argvalues=[
-            pytest.param(*["25Gi"], marks=pytest.mark.polarion_id("OCS-2007")),
-            pytest.param(*["50Gi"], marks=pytest.mark.polarion_id("OCS-2007")),
-            pytest.param(*["100Gi"], marks=pytest.mark.polarion_id("OCS-2007")),
+            pytest.param(
```
keesturam, line 128: please add polarion IDs.

tests/e2e/performance/test_pvc_creation_deletion_performance.py
```diff
@@ -150,12 +193,22 @@ def test_pvc_creation_deletion_measurement_performance(
         )
         self.full_results.add_key("pvc_size", pvc_size)
         num_of_samples = 5
-        accepted_creation_time = 1
-        accepted_deletion_time = 2 if self.interface == constants.CEPHFILESYSTEM else 1
+        accepted_creation_time = (
+            3600 if self.interface == constants.CEPHBLOCKPOOL_THICK else 1
```
keesturam, line 197: A time out of 1 hour for 25Gi volume? Do you plan to redefine the timeout based on our test results?

tests/e2e/performance/test_pvc_creation_deletion_performance.py
```diff
@@ -150,12 +193,22 @@ def test_pvc_creation_deletion_measurement_performance(
...
+        elif self.interface == constants.CEPHBLOCKPOOL:
+            accepted_deletion_time = 1
+        else:
+            accepted_deletion_time = 5
+
         self.full_results.add_key("samples", num_of_samples)
 
-        accepted_creation_deviation_percent = 50
-        accepted_deletion_deviation_percent = 50
+        accepted_creation_deviation_percent = 80
```
keesturam, line 210: We've had several conversations around the deviation. creation time for rbd and cephfs can simply be verified to be less than 1 second and we can skip the calculation of deviation for cephfs & rbd. WDYT?

tests/e2e/performance/test_pvc_creation_deletion_performance.py
```diff
@@ -165,7 +218,10 @@ def test_pvc_creation_deletion_measurement_performance(
             logging.info(f"{msg_prefix} Start creating PVC number {i + 1}.")
             start_time = datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
             pvc_obj = helpers.create_pvc(sc_name=self.sc_obj.name, size=pvc_size)
-            helpers.wait_for_resource_state(pvc_obj, constants.STATUS_BOUND)
+            timeout = 3600 if self.interface == constants.CEPHBLOCKPOOL_THICK else 60
```
keesturam, line 221: attaching an rbd thick provisioned PVC should take the same time as rbd PVC as the initialization is already done during the creation of the PVC.

keesturam (commented):
@ypersky1980 can we separate the change for performance label into a different PR? I'm sure there is going to be some discussion around it and we don't want to hold this PR until we have an answer for the label. wdyt?

ocs_ci/helpers/helpers.py
```diff
@@ -170,7 +170,10 @@ def create_pod(
         AssertionError: In case of any failure
 
     """
-    if interface_type == constants.CEPHBLOCKPOOL:
+    if (
+        interface_type == constants.CEPHBLOCKPOOL
```
Avilir, line 174: dont we need to use different pools for RBD and RBD-Thick ?

tests/e2e/performance/test_pvc_creation_deletion_performance.py
ypersky1980: @keesturam - regarding the label - no problem. I'll remove the new labels from this PR and leave only "performance" label for all the test cases. And in the next PR we can add new labels to the test cases that we decide to run with other than "performance" label.

ypersky1980: > `self.interface = interface_iterate` should be included in the else block to run tests with cephfs & rbd interface

@keesturam Currently self.interface=interface_type line appears before the if--else-- and will be executed for all 3 interfaces. Thanks

tests/e2e/performance/test_pvc_creation_deletion_performance.py
ypersky1980: @keesturam  I've changed it to 600 secs ( upper limit of actial results)

tests/e2e/performance/test_pvc_creation_deletion_performance.py
ypersky1980: I think that in case creation /deletion time is less then 1 second - calculating and checking standard deviation is not relevant. 
And if those times are bigger than 1 second - we'll fail the test ( for CephFS and for RBD). 

As for rbd thick - I think that Standard Deviation calculations are irrelevant as well. Actual results show that the deviation in this case is often more than 50%. 

So should I just remove all the calculations of standard deviation from this test ?

tests/e2e/performance/test_pvc_creation_deletion_performance.py
ypersky1980: @keesturam I'll check whether the test will pass with timeout =6- for reaching status Bound. 
If not - should I open a bug ?

ocs_ci/helpers/helpers.py
ypersky1980: @Avilir  - In this function  we are creating a pod. 

As for your question - when the test runs for RBD interface and RBD - thick interfaces those are separate runs and we will not have case in this the same pool contains pvcs of different types. 
What differs is storage class, not the pool.

tests/e2e/performance/test_pvc_creation_deletion_performance.py
ypersky1980: @keesturam  60 secs is not enough - all the RBD thick PVCs do not reach bound state within this time. Is it a bug? 
I'm going to test timeout = 600 now. 
Are you sure that time for reaching bound state is not part of the creation time ?

ypersky1980:
The full test log ( 12/12 passed) is available here ( run on VMware platform) : 10.70.39.233:/home/ypersky/ypersky_pr_validation/pr_4436/

tests/e2e/performance/test_pvc_creation_deletion_performance.py
```diff
@@ -115,11 +123,44 @@ def init_full_results(self, full_results):
...
+                marks=[pytest.mark.performance],
+            ),
+            pytest.param(
+                *[constants.CEPHFILESYSTEM, "25Gi"],
+                marks=[pytest.mark.performance],
+            ),
+            pytest.param(
+                *[constants.CEPHBLOCKPOOL_THICK, "25Gi"],
+                marks=[pytest.mark.performance],
+            ),
```
Avilir, line 163: pleas change the order, all test from the same interface type together :
*[constants.CEPHBLOCKPOOL, "5Gi"]
*[constants.CEPHBLOCKPOOL, "15Gi"]
*[constants.CEPHBLOCKPOOL, "25Gi"]

*[constants.CEPHFILESYSTEM, "5Gi"]
*[constants.CEPHFILESYSTEM, "15Gi"]
*[constants.CEPHFILESYSTEM, "25Gi"]

*[constants.CEPHBLOCKPOOL_THICK, "5Gi"]
*[constants.CEPHBLOCKPOOL_THICK, "15Gi"]
*[constants.CEPHBLOCKPOOL_THICK, "25Gi"]

This is for the dashboard report.

ypersky1980: @Avilir done, thank you!

ypersky1980:
Please note that the test has passed ( 12/12) on DC4 and the full log is available here : 

The full test log ( 12/12 passed) is available here ( run on VMware platform) : 10.70.39.233:/home/ypersky/ypersky_pr_validation/pr_4436/ypersky_test_pvc_creation_deletion_perf_multi_2806_full

Avilir (approved):
Nice work Yuli.

keesturam (approved):
LGTM

PR #4493, opened 2021-06-29
Rely on OCS version rather than OCP for MCG UI tests

MeridianExplorer:
Since MCG is entirely dependent on OCS version, we should run the tests based on the OCS version in the cluster, and not the OCP one.


bernerhat (commented):
wouldn't it be better to create a fixture with namedtuple for all of the resources and ocs version?

OdedViner (approved):
LGTM

MeridianExplorer:
@bernerhat - I think that the simplicity it introduces might be overshadowed by the implementational complexity and non-trivialness. For now, I personally think it's better like this; Especially considering we're working with OCS version, meaning something like `locators.4.8.obc` might be inevitable.

ebenahar (approved):

MeridianExplorer:
There was one failure in the verification job because the default backingstore is unhealthy at the moment; Verified

Issue #4497, opened 2021-06-30
ClusterObjectStoreState message is changed since OCS 4.7


fbalak:
Alert message needs to be updated for future versions.

PR #4498, opened 2021-06-30
Update ClusterObjectStoreState alert message for higher version

fbalak:
Fixes https://github.com/red-hat-storage/ocs-ci/issues/4497

### DIFF
--- a/tests/manage/monitoring/prometheus/test_rgw.py
+++ b/tests/manage/monitoring/prometheus/test_rgw.py
@@ -1,6 +1,7 @@
 import logging
 import pytest
 
+from ocs_ci.framework import config
 from ocs_ci.framework.testlib import tier4, tier4a
 from ocs_ci.ocs import constants
 from ocs_ci.utility import prometheus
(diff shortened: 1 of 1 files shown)


tests/manage/monitoring/prometheus/test_rgw.py
```diff
@@ -25,10 +26,14 @@ def test_rgw_unavailable(measure_stop_rgw):
     # get alerts from time when manager deployment was scaled down
     alerts = measure_stop_rgw.get("prometheus_alerts")
     target_label = constants.ALERT_CLUSTEROBJECTSTORESTATE
-    target_msg = (
-        "Cluster Object Store is in unhealthy state for more than 15s. "
-        "Please check Ceph cluster health or RGW connection."
-    )
+    # The alert message is changed since OCS 4.7
+    if float(config.ENV_DATA["ocs_version"]) < 4.7:
```
mbukatov, line 30: Could `ocs_version` here be also something like `4.7.2`? If yes, I don't think it will work.
Some code in ocs-ci uses https://pypi.org/project/semantic-version/ module (I just checked that it's installed as indirect dependency, so it will be available even though it's not directly listed in requirements of ocs-ci).

mbukatov: Moreover consider what would happen when we reach version 4.10 :)

ebenahar: right, please see https://github.com/red-hat-storage/ocs-ci/blob/76cf9dfa780f672fd95d78754f7fdcb19e633e66/ocs_ci/ocs/node.py#L991 as a reference

mbukatov (approved):

ebenahar (approved):

PR #4501, opened 2021-07-01
Updating Snapshot creation time log parsing

Avilir:
On OCS 4.8 the way that the snapshot creation report to the log was changed, and all performance test of snapshot creation was failed.

This PR is fixing this problem and in case the creation time did not found in the log in the previous format, it try to find it in the new format (for backward compatibility)


ypersky1980 (approved):
Thank you for a prompt fix!

vasukulkarni (approved):

PR #4522, opened 2021-07-05
Added RGW restart count fixture into first running Tier1 test

AviadP:
Automated BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1784255
As part of Tier1 tests (where RGW pod count should remain 0), verify that no unwanted restart accrued

### DIFF
--- a/tests/manage/monitoring/prometheusmetrics/test_rgw.py
+++ b/tests/manage/monitoring/prometheusmetrics/test_rgw.py
@@ -8,7 +8,7 @@ import logging
 
 import pytest
 
-from ocs_ci.framework.testlib import skipif_ocs_version, tier4, tier4a
+from ocs_ci.framework.testlib import skipif_ocs_version, tier1, tier4, tier4a
 from ocs_ci.ocs import constants, defaults, ocp
 from ocs_ci.ocs import metrics
 from ocs_ci.ocs.resources.ocs import OCS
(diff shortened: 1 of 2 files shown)


ebenahar (dismissed):

jilju (dismissed):

AviadP:
verified locally: http://pastebin.test.redhat.com/977298

mbukatov (changes requested):
It would be better do change this so that:
- fixture `verify_rgw_restart_count_session` is started by some empty tier1 test case
- check of the RGW restart cound is done in a tier1 test marked with `@pytest.mark.last`, so that when this check fails, it will fail a dedicated test

Problem with the current design is that when this fails, it will be reported as a teardown issue in unrelated test case, which is bad for debugging and visibility.

MeridianExplorer:
@mbukatov - I believe there are two problems with the offered solution:
1. `@pytest.mark.last` is already in use by `test_add_capacity()`
2. The restart count check is performed in the teardown of `verify_rgw_restart_count_session`; Meaning that if we add this fixture to a test marked with `last` it will always succeed - it'll check the number of restarts at the time of the last test, and then directly proceed to the teardown, see that no restarts were performed, and pass.

I have an alternative offer - let's create an empty tier1 test, mark it with `@pytest.mark.run(order=1)`, then mark `test_monitoring_enabled()` to run afterwards with `@pytest.mark.run(order=2)`.
Since the fixture is session-scoped, it'll save the number of restarts right in the beginning, and will check it only after all tests are done running (because the teardown will only occur at the end of the entire session).

mbukatov (approved):

Avilir (approved):

tests/manage/monitoring/prometheusmetrics/test_rgw.py
```diff
@@ -56,3 +56,14 @@ def test_ceph_rgw_metrics_after_metrics_exporter_respin(rgw_deployments):
         "so that the list of metrics without results is empty."
     )
     assert list_of_metrics_without_results == [], msg
+
+
+@tier1
+@pytest.mark.post_ocp_upgrade
+@pytest.mark.run(order=1)
+@pytest.mark.polarion_id("OCS-2584")
+def verify_rgw_pods_restart_count(verify_rgw_restart_count_session):
```
ebenahar, line 65: not sure it's a good practice to call a session scope fixture from a test function. Does it work?

AviadP: it does

Issue #4531, opened 2021-07-07
test_osd_heap_profile test in tier 2 fails with TypeError: argument of type 'NoneType' is not iterable


sameshai:
https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/view/Tier2/job/qe-trigger-ibmcloud-managed-1az-rhel-3w-tier2/9/testReport/tests.manage.z_cluster.test_osd_heap_profile/TestOSDHeapProfile/test_osd_heap_profile/

log.info(f"Verify osd.{osd_id}.profile log exist on /var/log/ceph/")
        out = osd_pod_profile.exec_cmd_on_pod(command="ls -ltr /var/log/ceph/")
        log.info(out)
>       assert (
            f"osd.{osd_id}.profile" in out
        ), f"osd.{osd_id}.profile log does not exist on /var/log/ceph/\n{out}"
E       TypeError: argument of type 'NoneType' is not iterable

PR #4535, opened 2021-07-08
Add TimeoutSampler generator for test_osd_heap_profile

OdedViner:
Add a generator to check the content of the folder /var/log/ceph/
#4531

### DIFF
--- a/tests/manage/z_cluster/test_osd_heap_profile.py
+++ b/tests/manage/z_cluster/test_osd_heap_profile.py
@@ -74,31 +74,30 @@ class TestOSDHeapProfile(ManageTest):
             if get_osd_pod_id(osd_pod) == osd_id:
                 osd_pod_profile = osd_pod
 
-        log.info(f"Verify osd.{osd_id}.profile log exist on /var/log/ceph/")
+        osd_profile_str = f"osd.{osd_id}.profile"
+        log.info(f"Verify {osd_profile_str} log exist on /var/log/ceph/")
         sample = TimeoutSampler(
             timeout=100,
             sleep=10,
             func=self.verify_output_command_osd_pod,
             command="ls -ltr /var/log/ceph/",
             pod_obj=osd_pod_profile,
-            osd_id=osd_id,
+            str_to_check=osd_profile_str,
         )
         if not sample.wait_for_func_status(result=True):
-            log.error(f"osd.{osd_id}.profile log does not exist on /var/log/ceph")
-            raise ValueError(
-                f"osd.{osd_id}.profile log does not exist on /var/log/ceph"
-            )
+            log.error(f"{osd_profile_str} log does not exist on /var/log/ceph")
+            raise ValueError(f"{osd_profile_str} log does not exist on /var/log/ceph")
 
         log.info(f"osd.{osd_id}.profile log exist on /var/log/ceph")
 
-    def verify_output_command_osd_pod(self, command, pod_obj, osd_id):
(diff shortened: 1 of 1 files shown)


tests/manage/z_cluster/test_osd_heap_profile.py
```diff
@@ -73,9 +75,43 @@ def test_osd_heap_profile(self):
...
+
+        """
+        try:
+            out = pod_obj.exec_cmd_on_pod(command=command)
+            if f"osd.{osd_id}.profile" not in out:
+                log.info(f"output is {out}")
+                return False
+            else:
+                log.info(f"the output of the command {command}: {out}")
+                return True
```
PrasadDesala, line 114: May be you can change these lines of codes to,
log.info(f"the output of the command {command}: {out}")
return True if f"osd.{osd_id}.profile" in out else False

tests/manage/z_cluster/test_osd_heap_profile.py
```diff
@@ -73,9 +75,43 @@ def test_osd_heap_profile(self):
...
+
+    def verify_output_command_osd_pod(self, command, pod_obj, osd_id):
+        """
+        Check the output of the command (from osd pod)
+
+        Args:
+            command (str): command run on osd pod
+            pod_obj (obj): pod object
+            osd_id (str): osd id
+
```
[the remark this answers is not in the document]
OdedViner: I dont raise exception on this function

tests/manage/z_cluster/test_osd_heap_profile.py
OdedViner: done

tests/manage/z_cluster/test_osd_heap_profile.py
```diff
@@ -73,9 +75,39 @@ def test_osd_heap_profile(self):
...
+        )
+        if not sample.wait_for_func_status(result=True):
+            log.error(f"osd.{osd_id}.profile log does not exist on /var/log/ceph")
+            raise ValueError(
+                f"osd.{osd_id}.profile log does not exist on /var/log/ceph"
+            )
+
         log.info(f"osd.{osd_id}.profile log exist on /var/log/ceph")
+
+    def verify_output_command_osd_pod(self, command, pod_obj, osd_id):
```
yitzhak12, line 94: I think it will be clearer if you add another param `str_to_check`, or a similar name. Then you can pass the string "osd.{osd_id}.profile" in the test to the function `verify_output_command_osd_pod` instead of writing it twice.

OdedViner: @yitzhak12 In the case of ERROR, I want to fail the test, and not use generator...

yitzhak12: But still, you can add the param `str_to_check` to the function. 
What I mean is to change the code to the following: 
`def verify_output_command_osd_pod(self, command, pod_obj, osd_id, str_to_check):`

And then:
`sample = TimeoutSampler(
            timeout=100,
            sleep=10,
            func=self.verify_output_command_osd_pod,
            command="ls -ltr /var/log/ceph/",
            pod_obj=osd_pod_profile,
            osd_id=osd_id,
            str_to_check="osd.{osd_id}.profile"
        )`

yitzhak12 (approved):

tests/manage/z_cluster/test_osd_heap_profile.py
OdedViner: done

ebenahar (approved):

PR #4540, opened 2021-07-08
OCP 4.8 signle host Deployment fix for disabling Affinitygroup

### LINKED ISSUE #4539: RHV OCP 4.8 deployment failing due to default setting of Affinitygroup
Problem: OCP-4.8 deployment is failing with error `failed to fetch Cluster: failed to fetch dependency of "Cluster": failed to generate asset "Platform Provisioning Check": Affinity Group compute cannot be fulfilled, oVirt cluster doesn'thave enough hosts: found 1 hosts but 3 replicas assigned to affinity group%!(EXTRA ovirt.AffinityGroup={compute 3 AffinityGroup for spreading each compute machine to a different host true})`

Analysis: 
As per the new OCP 4.8, Affinitygroup is getting enabled by default. 
It is recommended to add a non-enforcing affinity group to distribute the control plane and workers,  to use as much of the cluster as possible.
On the RHV platform we have 2 live setups,
     
1. cluster name `rhv1` i.e credential file `rhv-DC1` and 
2. cluster name `rhv2` i.e credential file `rhv-DC2`.
 Both these setups use a single host.
 
Since it doesn't have a different host, deployment is failing with error: 
`failed to fetch Cluster: failed to fetch dependency of "Cluster": failed to generate asset "Platform Provisioning Check": Affinity Group compute cannot be fulfilled, oVirt cluster doesn't have enough hosts: found 1 host but 3 replicas assigned to affinity group%!(EXTRA ovirt.AffinityGroup={compute 3 AffinityGroup for spreading each compute machine to a different host true})`

Fix: We need to disable Affinitygroup in install-config while provisioning the RHV cluster.

suchita-g:
Closes #4539

### DIFF
--- a/ocs_ci/templates/ocp-deployment/install-config-rhv-ipi.yaml.j2
+++ b/ocs_ci/templates/ocp-deployment/install-config-rhv-ipi.yaml.j2
@@ -13,6 +13,7 @@ compute:
       osDisk:
         sizeGB: 120
       vmType: server
+      affinityGroupsNames: []
   replicas: {{ worker_replicas | default(3) }}
 controlPlane:
   architecture: amd64
@@ -27,6 +28,7 @@ controlPlane:
       osDisk:
         sizeGB: 120
       vmType: server
+      affinityGroupsNames: []
   replicas: {{ master_replicas | default(3) }}
 metadata:
   creationTimestamp: null


suchita-g:
Verified: https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/4353/

ocs_ci/templates/ocp-deployment/install-config-rhv-ipi.yaml.j2
```diff
@@ -13,6 +13,7 @@ compute:
       osDisk:
         sizeGB: 120
       vmType: server
+      affinityGroupsNames: []
```
prsurve, line 16: I would suggest you can add this explicitly in the deployment code because this parm will not work in the case of OCP 4.7

suchita-g: Noted

petr-balogh: @suchita-g if it will break OCP 4.7 deployment shouldn't we fix this?

PR #4543, opened 2021-07-09
Introducing new tier scale_long_run and scale_changed_layout

ramkiperiy:
Added new tier scale_long_run and scale_changed_layout
scale_long_run: scale tests which has execution time in days
scale_changed_layout: scale capacity/node tests without teardown

### DIFF
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -59,6 +59,8 @@ workloads = pytest.mark.workloads
 flowtests = pytest.mark.flowtests
 performance = pytest.mark.performance
 scale = pytest.mark.scale
+scale_long_run = pytest.mark.scale_long_run
+scale_changed_layout = pytest.mark.scale_changed_layout
 deployment = pytest.mark.deployment
 polarion_id = pytest.mark.polarion_id
 bugzilla = pytest.mark.bugzilla
@@ -74,6 +76,8 @@ tier_marks = [
     tier_after_upgrade,
     performance,
     scale,
+    scale_long_run,
+    scale_changed_layout,
     workloads,
 ]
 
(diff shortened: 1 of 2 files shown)


vasukulkarni (dismissed):

tiffanyn108 (dismissed):
LGTM.

wusui (dismissed):
LGTM

wusui (commented):
Actually, I have one question.  I get that there are now two scale decorators, scale_tier1 and scale_tier2.  I assume that scale_tier1 is used for scale test with teardown cleanup expected
and scale_tier2 is not.  How is they different within the infrastucture in that I don't see anything to distinguish the two from each other except for the value number.  In other words, how does tier2 know not to expect that leftover changes will not be cleaned up?

ETA: I guess this will work if everything in tier2 does ignore_leftovers and we know that things are changed.  Then these classifications work to keep things separate.  I approve.

ramkiperiy:
> Actually, I have one question. I get that there are now two scale decorators, scale_tier1 and scale_tier2. I assume that scale_tier1 is used for scale test with teardown cleanup expected
> and scale_tier2 is not. How is they different within the infrastucture in that I don't see anything to distinguish the two from each other except for the value number. In other words, how does tier2 know not to expect that leftover changes will not be cleaned up?
> 
> ETA: I guess this will work if everything in tier2 does ignore_leftovers and we know that things are changed. Then these classifications work to keep things separate. I approve.

I would say only difference is in scale_tier2 expect the setup to be different shape before start of test execution.
ignore_leftovers marker is not useful here, it is to skip comparison of resources before and after test execution, teardown is handled either in fixture or in cleanup before exiting the tests. I tried to do that eg: refer PR #2697 there are multiple tests in same script but at the end of each test there is no teardown.

wusui (approved):
LGTM

keesturam (approved):

Issue #4489, opened 2021-06-29
image-registry pods are not coming up 


shreekarSS:
From https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/1173/testReport/junit/tests.e2e.registry.test_registry_by_increasing_num_of_image_registry_pods/TestRegistryByIncreasingNumPods/test_registry_by_increasing_num_of_registry_pods/

E               ocs_ci.ocs.exceptions.TimeoutExpiredError: Timed out after 300s running get_pod_name_by_pattern("image-registry", "openshift-image-registry")

PR #4557, opened 2021-07-13
On teardown validate replica count of registry pod

Akarsha-rai:
Fixes: #4489 

Failed due to bug 1981639 as a result one of the registry pod stuck in a pending state which caused other tests to fail. So at the teardown make sure the replica count is set to 2.

Signed-off-by: Akarsha-rai <akrai@redhat.com>


PrasadDesala (approved):

ebenahar (approved):

Akarsha-rai:
The changes are done only for test `tests/e2e/registry/test_registry_by_increasing_num_of_image_registry_pods.py` and it passed. The Jenkin runs failed on some other tests, based on marking it as verified.

PR #4566, opened 2021-07-14
Skip manage tests with partially deployed cluster 

bernerhat:
Skip tests relaying on a specific component when it is not deployed, as part of [KNIP-1609](https://issues.redhat.com/browse/KNIP-1609).


AviadP:
Hi @bernerhat , it is possible to split this PR and do it part by part? it would be very hard to review and verify such PR

Issue #4542, opened 2021-07-08
In the test "test_check_pods_status_after_node_failure" need to consider the case of "pod not found"


yitzhak12:
In the test "test_check_pods_status_after_node_failure" in the function `wait_for_change_in_rook_ceph_pods` in the function `wait_for_change_in_pods_statuses`, when getting the pod objects, one of the pod didn't found. The test is planned to avoid such cases, but it seems that it has a small chance of happening. We need to fix the function so we didn't hit this kind of error. If we have this error, we need to continue with the test and ignore it.

yitzhak12:
This issue was raised due to this failed test: https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/1291/testReport/tests.manage.z_cluster.nodes.test_check_pods_status_after_node_failure/TestCheckPodsAfterNodeFailure/test_check_pods_status_after_node_failure/

yitzhak12:
After another investigation, I figured out that the issue is in the function `get_rook_ceph_pod_names_not_in_node` in the function `get_node_pods(node_name)` in the line https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/node.py#L1351 `return [p for p in pods_to_search if pod.get_pod_node(p).name == node_name]` when calling the function `get_pod_node`. In the line https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/resources/pod.py#L1140 `node_name = pod_obj.get().get("spec").get("nodeName")` when executing the command `pod_obj.get()` the pod object not found, because it's already deleted when executing the command. This is a rare case, but we also need to consider it, and fix the test so this kind of error will not fail the test.

PR #4567, opened 2021-07-14
Fix the function "get_node_pods" to not throw an exception in case of pod not found

yitzhak12:
Regarding this issue https://github.com/red-hat-storage/ocs-ci/issues/4542, we see that the test failed because one of the pods has been deleted while the function "get_node_pods" was running - which cause a CommandFailed exception of "pod not found" error. To fix it, I added one more param `raise_pod_not_found_error` in the function "get_node_pods", with the default value of `False`, so now it will not throw an exception of "pod not found" error unless we pass this param with the value `True`.

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -10,7 +10,12 @@ from semantic_version import Version
 from ocs_ci.ocs.machine import get_machine_objs
 
 from ocs_ci.framework import config
-from ocs_ci.ocs.exceptions import TimeoutExpiredError, NotAllNodesCreated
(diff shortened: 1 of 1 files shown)


ebenahar:
@yitzhak12 in which cases will we throw the exception?

yitzhak12:
@ebenahar In the case of "pod not found" error and the param `raise_pod_not_found_error` is `False`.

ebenahar (approved):

jilju (approved):

PrasadDesala (approved):

OdedViner (approved):

PR #4568, opened 2021-07-14
disable ES from smallfile used in snapshot test

Avilir:
since the smallfile benchmark is used in this test only to create files on the volume and we are not interesting with the small file results, the ES configuration disabled in this test.

### DIFF
--- a/tests/e2e/performance/test_pvc_snapshot_performance.py
+++ b/tests/e2e/performance/test_pvc_snapshot_performance.py
@@ -24,6 +24,7 @@ from ocs_ci.framework.testlib import (
     E2ETest,
     performance,
 )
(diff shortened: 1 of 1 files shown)


ypersky1980 (approved):
LGTM

wusui (approved):
Looks good to me.  The ceph_health_check looks fine.  I assume that there was an issue if the small file elastic_search was not disabled?

ypersky1980 (approved):

Avilir:
> Looks good to me. The ceph_health_check looks fine. I assume that there was an issue if the small file elastic_search was not disabled?

yes it was, since it try to connect ES server, but it wasn't configured well - since this test don't relay need the ES

PR #4576, opened 2021-07-18
Fix UI infra

OdedViner:
- [ ] Change the name of `skipif_ui` decorator to `skipif_ui_not_support`
- [ ] Split the screenshot folder by test name
- [ ] Add Screenshot flag
- [ ] Add helpers_ui python file

### DIFF
--- a/ocs_ci/ocs/ui/base_ui.py
+++ b/ocs_ci/ocs/ui/base_ui.py
@@ -51,7 +51,7 @@ class BaseUI:
         elif config.ENV_DATA["platform"].lower() == constants.AWS_PLATFORM:
             self.storage_class = "gp2_sc"
 
-    def do_click(self, locator, timeout=30, enable_screenshot=True):
+    def do_click(self, locator, timeout=30, enable_screenshot=False):
         """
         Click on Button/link on OpenShift Console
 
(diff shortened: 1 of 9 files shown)


ocs_ci/ocs/ui/base_ui.py
```diff
@@ -40,27 +41,30 @@ def __init__(self, driver):
...
-            os.mkdir(self.screenshots_folder)
+            Path(self.screenshots_folder).mkdir(parents=True, exist_ok=True)
         logger.info(f"screenshots pictures:{self.screenshots_folder}")
         if config.ENV_DATA["platform"].lower() == constants.VSPHERE_PLATFORM:
             self.storage_class = "thin_sc"
         elif config.ENV_DATA["platform"].lower() == constants.AWS_PLATFORM:
             self.storage_class = "gp2_sc"
 
-    def do_click(self, locator, timeout=30):
+    def do_click(self, locator, timeout=30, enable_screenshot=True):
```
shyRozen, line 54: I'm thinking out load...Maybe the default for enable_screenshot should be false and when we do want we should send true. I think that in most cases except if we catch some timeout exception it should be false. We need to strive to have less screenshot taken. WDYT

OdedViner: @shyRozen @MeridianExplorer I changed the enable_screenshot to False by default..

shyRozen (approved):

ebenahar (approved):

PR #4600, opened 2021-07-21
Add multus prod configs

clacroix12:

### DIFF
--- a/conf/deployment/vsphere/upi_1az_rhcos_multus_vmfs_3m_3w.yaml
+++ b/conf/deployment/vsphere/upi_1az_rhcos_multus_vmfs_3m_3w.yaml
@@ -1,0 +1,18 @@
+---
+# This config is suppose to work on most of DCs we have.
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
+ENV_DATA:
+  platform: 'vsphere'
+  deployment_type: 'upi'
+  worker_replicas: 3
+  master_replicas: 3
+  worker_num_cpus: '16'
+  master_num_cpus: '4'
+  master_memory: '16384'
+  compute_memory: '65536'
+  fio_storageutilization_min_mbps: 10.0
+  is_multus_enabled: true
+REPORTING:
+  polarion:
+    deployment_id: 'OCS-2510'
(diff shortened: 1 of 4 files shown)


clacroix12:
@petr-balogh @vavuthu I only found two polarion IDs related to multus, one for UI deployment and the other for CLI. This is why the polarion IDs are identical across these deployment configs.

I have created vsphere upi configs for both RHEL and RHCOS, as well as VSAN and VMFS. Please let me know if I'm missing any scenarios or if we don't need all of these. I'm unsure exactly which scenarios we need to cover with prod configs.

vavuthu (approved):

Akarsha-rai (approved):

sidhant-agrawal:
@vavuthu @clacroix12 there are three configuration possible with multus
- If only the public selector is specified
- If only the cluster selector is specified
- If both public and cluster selectors are specified

Do we plan to include those configurations as well?

clacroix12:
@sidhant-agrawal currently we only have the one configuration option `enable_multus` so I created prod configs that use that. This only creates the NAD for the public network. If we also need to support the private network we may want to refactor `enable_multus` to `enable_multus_public` and create a new config option for `enable_multus_private`. We can then create the corresponding NADs using those config options and adjust verification using them as well. @vavuthu thoughts?

vavuthu:
public network is mandatory. either we have ONLY public network or public network + cluster network.

currently we have tested public network. I am not sure about public network + cluster network. @sidhant-agrawal could you please confirm here.

Once we have automation/infra ready, we need to refactor and have separate configs.

sidhant-agrawal:
>I don't think there is a "private cluster selector only" scenario according to this since the public network is mandatory. Can you confirm?

> public network is mandatory. either we have ONLY public network or public network + cluster network.
> 

@clacroix12 @vavuthu 
`public network is mandatory` -> this is no longer true.
Please refer to [multus section](https://rook.io/docs/rook/v1.6/ceph-cluster-crd.html#multus) in docs for more information and possible configurations.

> currently we have tested public network. I am not sure about public network + cluster network. @sidhant-agrawal could you please confirm here.
> 
We have tested `public network + cluster network` on BareMetal setup.

PR #4631, opened 2021-07-28
[Do not merge] Delete BZ marker 1943137

fbalak:


mbukatov:
Why do we want to remove the marker?

Issue #2832, opened 2020-08-30
Ceph RGW endpoint isn't externalized and is inaccessible from outside the cluster


MeridianExplorer:
This prevents us from using the Python S3 SDK to perform various actions on RGW, and is also the reason we're currently skipping the RGW multipart upload test (which relies on said SDK)

ebenahar:
@MeridianExplorer is there a reason to keep this test marked as skipped? tests/manage/rgw/test_multipart_upload.py::TestS3MultipartUpload::test_multipart_upload_operations
Is there anything to be done from the product side or from our CI in order to enable the test?

MeridianExplorer:
@ebenahar - Not anymore. This issue was also solved in #3353

PR #4643, opened 2021-07-29
Unskip RGW multipart upload test

MeridianExplorer:
The test was skipped due to #2832 which was since closed.


bernerhat (dismissed):

ebenahar (approved):

ocs_ci/ocs/constants.py
```diff
@@ -886,6 +886,10 @@
 CEPHFS_PROVISIONER_SECRET = "rook-csi-cephfs-provisioner"
 CEPHFS_NODE_SECRET = "rook-csi-cephfs-node"
 # OSU = ObjectStoreUser, shortened for compliance with flake8+black because of line length issues
+OSU_SECRET_BASE = "rook-ceph-object-user-ocs-{}storagecluster-cephobjectstore-{}-{}"
+CEPH_OBJECTSTOREUSER_SECRET = OSU_SECRET_BASE.format(
+    "", "ocs-storagecluster", "cephobjectstoreuser"
+)
 NB_OSU_SECRET_BASE = "rook-ceph-object-user-ocs-{}storagecluster-cephobjectstore-noobaa-ceph-objectstore-user"
 NOOBAA_OBJECTSTOREUSER_SECRET = NB_OSU_SECRET_BASE.format("")
```
bernerhat, line 894: looks like the new base could be of use in the noobaa secret maybe we can use it and save a const?

tests/conftest.py
```diff
@@ -2005,7 +2005,7 @@ def rgw_deployments(request):
         pytest.skip("There is no RGW deployment available for this test.")
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="session", autouse=True)
```
bernerhat, line 2008: dont like the autouse of the fixture, could lead to problems in mcg-only deployment

bernerhat (dismissed):

tests/conftest.py
petr-balogh: Also be aware that the autouse fixture is scheduled even in deployment itself which leads to the situation that we don't have any OCP/ODF installed yet. In the case something will fail here it will fail all the test cases as well.

MeridianExplorer: Thanks Amit, Petr - I removed the autouse

ebenahar (dismissed):

parikshithb (approved):

parikshithb (approved):

PR #4645, opened 2021-07-29
Block pool UI create use and delete

shyRozen:
Pool management [KNIP-1548](https://issues.redhat.com/browse/KNIP-1548) 
Added UI classes for pool and storageclass (basic RBD) and test that create blockpool use them and delete.
There will more tests added for this PR so review but not merge.

### DIFF
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -868,6 +868,45 @@ def validate_cephfilesystem(fs_name):
     return True if (ceph_validate and ocp_validate) else False
 
 
+def create_ocs_object_from_kind_and_name(
+    kind, resource_name, namespace=constants.OPENSHIFT_STORAGE_NAMESPACE
+):
+    """
+    Create OCS object from kind and name
+
+    Args:
+        kind (str): resource kind like CephBlockPool, pvc.
+        resource_name (str): name of the resource.
+        namespace (str) the namespace of the resource.
+
+    Returns:
+        ocs_ci.ocs.resources.ocs.OCS (obj): returns OCS object from kind and name.
+
+    """
+    ocp_object = OCP(kind=kind, resource_name=resource_name, namespace=namespace).get()
+    return OCS(**ocp_object)
+
+
+def remove_ocs_object_from_list(kind, resource_name, object_list):
+    """
+    Given a list of OCS objects, the function removes the object with kind and resource from the list
+
+    Args:
+        kind (str): resource kind like CephBlockPool, pvc.
+        resource_name (str): name of the resource.
+        object_list (array): Array of OCS objects.
+
+    Returns:
+        (array): Array of OCS objects without removed object.
+
+    """
+
+    for obj in object_list:
+        if obj.name == resource_name and obj.kind == kind:
+            object_list.remove(obj)
+            return object_list
+
+
 def get_all_storageclass_names():
     """
     Function for getting all storageclass
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -1021,6 +1021,23 @@ def get_all_resource_names_of_a_kind(kind):
     ]
 
 
+def get_all_resource_of_kind_containing_string(search_string, kind):
+    """
+    Return all the resource of kind which name contain search_string
+    Args:
+         search_string (str): The string to search in name of the resource
+         kind (str): Kind of the resource to search for
+    Returns:
+        (list): List of resource
+    """
+
+    resource_list = []
+    for resource in OCP(kind=kind).get().get("items"):
+        if search_string in resource["metadata"]["name"]:
+            resource_list.append(resource)
+    return resource_list
+
+
 def get_clustername():
     """
     Return the name (DNS short name) of the cluster
--- a/ocs_ci/ocs/ui/block_pool.py
+++ b/ocs_ci/ocs/ui/block_pool.py
@@ -1,0 +1,175 @@
+import logging
+import time
+
+from ocs_ci.ocs.ui.base_ui import PageNavigator
+from ocs_ci.ocs.ui.views import locators
+from ocs_ci.utility.utils import get_ocp_version
+from selenium.webdriver.common.by import By
+from ocs_ci.helpers.helpers import create_unique_resource_name
+from ocs_ci.ocs.exceptions import PoolStateIsUnknow
+import ocs_ci.ocs.resources.pod as pod
+
+logger = logging.getLogger(__name__)
+
+
+class BlockPoolUI(PageNavigator):
+    """
+    User Interface Selenium for Block Pools page
+
+    """
+
+    def __init__(self, driver):
+        super().__init__(driver)
+        ocp_version = get_ocp_version()
+        self.bp_loc = locators[ocp_version]["block_pool"]
+        self.sc_loc = locators[ocp_version]["storageclass"]
+
+    def create_pool(self, replica, compression):
+        """
+        Create block pool via UI
+
+        Args:
+            replica (int): replica size usually 2,3
+            compression (bool): True to enable compression otherwise False
+
+        Return:
+            array: pool name (str) pool status (bool) #pool can be created with failure status
+
+        """
+        pool_name = create_unique_resource_name("test", "rbd-pool")
+        self.navigate_block_pool_page()
+        self.do_click(self.bp_loc["create_block_pool"])
+        self.do_send_keys(self.bp_loc["new_pool_name"], pool_name)
+        self.do_click(self.bp_loc["first_select_replica"])
+        if replica == 2:
+            self.do_click(self.bp_loc["second_select_replica_2"])
+        else:
+            self.do_click(self.bp_loc["second_select_replica_3"])
+        if compression is True:
+            self.do_click(self.bp_loc["conpression_checkbox"])
+        self.do_click(self.bp_loc["pool_confirm_create"])
+        wait_for_text_result = self.wait_until_expected_text_is_found(
+            self.bp_loc["pool_state_inside_pool"], "Ready", timeout=15
+        )
+        if wait_for_text_result is True:
+            logger.info(f"Pool {pool_name} was created and it is in Ready state")
+            return [pool_name, True]
+        else:
+            logger.info(f"Pool {pool_name} was created but did not reach Ready state")
+            return [pool_name, False]
+
+    def check_pool_existence(self, pool_name):
+        """
+        Check if pool appears in the block pool list
+
+        Args:
+            pool_name (str): Name of the pool to check
+
+        Return:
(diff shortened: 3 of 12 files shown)


ocs_ci/ocs/ui/block_pool.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        Delete pool from pool page
+        Args:
+            pool_name (str): The name of the pool to be deleted
+        Return:
+            bool: True if pool is not found in pool list, otherwise false
+        """
+        self.navigate_overview_page()
+        self.navigate_block_pool_page()
+        self.page_has_loaded()
```
OdedViner, line 82: Is it mandatory to use this function? Does it not work without this function? we use 'WebDriverWait'.
All tests use the "navigate" function

shyRozen: WebDriverWait is good when you search for an element in the next page. This is good for knowing that the page was loaded without searching for a specific element.

ocs_ci/helpers/helpers.py
```diff
@@ -835,6 +835,30 @@ def validate_cephfilesystem(fs_name):
     return True if (ceph_validate and ocp_validate) else False
 
 
+def delete_all_resource_of_kind_containing_string(kind, search_string):
+    """
+    Delete all resource of kind specified with name containing search_string.
+    Args:
+        kind (str): kind of the resource to search and delete.
+        search_string (str): Search string for the resources to be searched and deleted.
```
ebenahar, line 843: please provide an example for such a string

ocs_ci/helpers/helpers.py
```diff
@@ -835,6 +835,30 @@ def validate_cephfilesystem(fs_name):
     return True if (ceph_validate and ocp_validate) else False
 
 
+def delete_all_resource_of_kind_containing_string(kind, search_string):
```
ebenahar, line 838: this function could be quite risky in case a certain string is passed while, and this string is also part of the name of other unrelated cluster resrouces. 
Is there any other way to delete the needed resources that is more explicit?

ocs_ci/ocs/ocp.py
```diff
@@ -999,6 +999,23 @@ def get_all_resource_names_of_a_kind(kind):
     ]
 
 
+def get_all_resource_of_kind_containing_string(search_string, kind):
+    """
+    Return all the resource of kind which name contain search_string
```
ebenahar, line 1004: Add 1 empty line in between each docstring section. 
See https://ocs-ci.readthedocs.io/en/latest/coding_guidelines.html?highlight=docstring#coding-guidelines-for-ocs-ci-project
Please apply in all other functions in this PR

ocs_ci/helpers/helpers.py
```diff
@@ -835,6 +835,30 @@ def validate_cephfilesystem(fs_name):
...
+        for resource in resource_dict_list:
+            OCS(**resource).delete()
+        resource_dict_after_list = get_all_resource_of_kind_containing_string(kind=kind, search_string=search_string)
+        if len(resource_dict_after_list) > 0:
+            for resource in resource_dict_after_list:
+                logger.error(f"Resource {resource.resource_name} was not deleted successfully")
+            return False
+        else:
+            return True
+    return True
```
ebenahar, line 859: if `len(resource_dict_list)` is 0 we return True. Is this intentional?

tests/ui/test_create_pool_block_pool.py
```diff
@@ -0,0 +1,165 @@
...
+class TestPoolUserInterface(ManageTest):
+    """
+    Test Pool User Interface
+
+    """
+
+    @pytest.fixture(scope="function", autouse=True)
+    def teardown(self, request):
+        def finalizer():
+            if not \
```
ebenahar, line 40: please break lines with parentheses

ocs_ci/helpers/helpers.py
shyRozen: As you can see `get_all_resource_of_kind_containing_string` is happening twice - one for getting the object and deleting them and the second one is to verify that there are no leftovers, so if there are no leftovers `len(resource_dict_list) is 0` I'm returning true.

ocs_ci/helpers/helpers.py
shyRozen: I saw now an implementation that teardown can use an array that can be appended in the middle of the test. This could be helpful and more explicit. I'll give it a shoot.

ocs_ci/helpers/helpers.py
ebenahar: ack

ocs_ci/helpers/helpers.py
shyRozen: Did a different implementation to teardown without using this function. Should I delete it from helpers?

am-agrawa: You already created a more explicit function, we can probably get rid of this one.

shyRozen: @ebenahar @am-agrawa What do you  think of making the string constant to "test".

ebenahar: discussed offline and since this function will not be in use, it will be removed

ocs_ci/helpers/helpers.py
```diff
@@ -838,6 +838,47 @@ def validate_cephfilesystem(fs_name):
...
+    Args:
+        kind (str): resource kind like CephBlockPool, pvc.
+        resource_name (str): name of the resource.
+        namespace (str) the namespace of the resource.
+
+    Returns:
+        ocs_ci.ocs.resources.ocs.OCS (obj): returns OCS object from kind and name.
+
+    """
+    ocp_block_object = OCP(
```
ebenahar, line 856: A few comments:
1) ocp_block_object falsely indicates that this function creates a specific kind of object 
2) why can't we use an existing function, such as ocs_ci.helpers.helpers.create_resource or one that creates an object of a specific kind?

tests/ui/conftest.py
```diff
@@ -0,0 +1,195 @@
+import logging
```
ebenahar, line 1: not all UI tests are under /tests/ui/ 
Maybe better to place these fixtures directly in tests/conftest.py

ebenahar (commented):
Great work Shay

ocs_ci/helpers/helpers.py
shyRozen: 1. Didn't understand quiet but in the end I'm returning OCS object as the function describes.
2. In ocs_ci.helpers.helpers.create_resource you have to pass a dictionary and since I've getting data from UI I prefer to have it this way with 3 arguments and not create a dict.

tests/ui/conftest.py
shyRozen: I agree. Will move. To tell you the true I think that having a UI directory is not a good idea. I think that even it's ui tests it should go to the regulars directory that we have now in the context of the test.
But do I need tier1 again??? :(

ebenahar: I believe deployment is sufficient for validation

ocs_ci/helpers/helpers.py
ebenahar: 1) what does `ocp_block_object` var name mean?
2) can't we extend ocs_ci.helpers.helpers.create_resource to allow it to get  a dict? I just think that maintaining another function that does pretty much the same is a redundancy

shyRozen: 1. Got it. Changing the name to ocp_object :)
2. ocs_ci.helpers.helpers.create_resource is different than create_ocs_object_from_kind_and_name:
a. ocs_ci.helpers.helpers.create_resource really creates a resource on the cluster while my function is creating an 
  object(python) and not really creating the resource on the cluster. It is for an already existing resource that was created from UI and is translating the name, kind and namespace into an python object. In cli tests you already have an object if you created a resource. Here the resource is created in UI but I we need to have a representation of it in the code so I'm creating an OBJECT.
 It is complex to merge the two function and will always have to do two steps in UI test. First create the dict with OCP and then decide if the user wants to just create a python object or really create resource on the cluster. I don't think the 2 function are the same.

shyRozen:
The 2 tests that failed are not related to code added.
One is awscli pod can't be deleted because it is not found.
E           ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc -n openshift-storage delete Pod session-awscli-relay-pod-de8a0969a9f44cf.
E           Error is Error from server (NotFound): pods "session-awscli-relay-pod-de8a0969a9f44cf" not found

Other is failure to delete object in MCG:
E           ocs_ci.ocs.exceptions.UnexpectedBehaviour: Object was not deleted from cache properly
Adding verified label.

ocs_ci/helpers/helpers.py
ebenahar: Ok, I understand. Please just check maybe we already have a function that does the same thing

shyRozen: image
I didn't find one with the relevant words

ocs_ci/ocs/ui/block_pool.py
```diff
@@ -0,0 +1,174 @@
...
+
+        Args:
+             replica (int): size of pool.
+             compression (bool): True for enabling compression. Otherwise False.
+
+        """
+        pool_list = []
+        ceph_pod = pod.get_ceph_tools_pod()
+
+        while True:
```
OdedViner, line 144: Is it possible not to use `while True`? maybe generator?
It's dangerous to use `while True`.. for example, `pool_status` is `True` every time...

shyRozen: I doubt that pool status will be forever true because it generates a pool every iteration and you can create pools till pg limit is reached which is like 16-20 pools max.
Anyway there is something in your saying but a generator is an overkill. I've added a counter with a limit of 50.

shyRozen:
@OdedViner @ebenahar please review and approve

ebenahar (approved):

tests/ui/test_create_pool_block_pool.py
```diff
@@ -0,0 +1,121 @@
...
+            size=self.pvc_size,
+        )
+
+    @pytest.fixture()
+    def pod(self, pod_factory):
+        self.pod_obj = pod_factory(pvc=self.pvc_obj)
+
+    @ui
+    @tier1
+    @skipif_ocs_version("<4.8")
```
OdedViner, line 67: Can you add `@skipif_ui_not_support("block_pool")` decorator?

OdedViner (approved):

ebenahar (approved):

tests/ui/test_create_pool_block_pool.py
shyRozen: Added

ebenahar (approved):

OdedViner (approved):

PR #4651, opened 2021-07-30
test OSD flapping alert

annagitel:
automation for bug 1935342
Trigger OSD flapping Prometheus alert buy deleting OSD 6 times


tests/manage/monitoring/prometheus/test_osd.py
```diff
@@ -0,0 +1,38 @@
...
+@pytest.mark.bugzilla("1935342")
+@tier4c
+def test_osd_flapping_alert(measure_stop_ceph_osd):
+    """
+    Prometheus alert for OSD restart. alert notifies if an OpenShift Container
+    Storage OSD restarts more than 5 times in 5 minutes.
+
+    """
+    expected_alert_msg = (
+        f"Storage daemon osd.0 has restarted 5 times in the last 5 minutes. Please check "
```
ebenahar, line 20: will it always be osd.0?

annagitel: yes, randomly picked to test the osd with id 0

tests/manage/monitoring/conftest.py
```diff
@@ -808,3 +808,42 @@ def delete_target_bucket():
...
+def measure_ceph_osd_flapping(measurement_dir):
+    """
+    Delete Ceph osd, measures the time when it was
+    deleted and alerts that were triggered during this event.
+
+    Returns:
+        dict: Contains information about `start` and `stop` time for stopping
+            Ceph osd pod
+    """
+
```
ebenahar, line 823: We need to have a solid teardown that will be called upon test failure. This test case is disruptive and in case it fails while OSD is down, the whole cluster is left broken for the subsequent test cases

tests/manage/monitoring/prometheus/test_osd.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+from ocs_ci.framework.pytest_customization.marks import tier4c
+from ocs_ci.ocs import constants
+from ocs_ci.utility import prometheus
+
+
+@pytest.mark.polarion_id("")  # TODO
```
ebenahar, line 7: Polarion I'd is mising

tests/manage/monitoring/conftest.py
annagitel: the test will fail if this alert won't show up 
for the teardown - I can add a check that new OSD came up after the last delete
but in a case it did not come up, what can we do?

fbalak (commented):
Nice usage of alerting fixtures.

tests/manage/monitoring/conftest.py
fbalak: I think that it would be a bug.

ebenahar: ok, so let's keep it as is

annagitel:
not able to proceed with this PR duo to a bug - this alert could not be triggered
when this bug will be fixed the PR is ready for merge

PR #4654, opened 2021-08-01
Create a new fixture to use in the "add capacity" tests setup 

yitzhak12:


tests/manage/z_cluster/cluster_expansion/conftest.py
```diff
@@ -107,3 +110,40 @@ def factory(ocs_nodes=False, node_count=3, taint_label=None):
...
+        logging.info(
+            f"We have {number_of_osd_pods_before} OSDs in the cluster, "
+            f"and {num_of_wnodes} worker nodes in the cluster"
+        )
+        if num_of_wnodes < num_of_expected_wnodes:
+            num_of_wnodes_to_add = num_of_expected_wnodes - num_of_wnodes
+            logging.info(
+                f"Adding more {num_of_wnodes_to_add} worker nodes to the cluster"
+            )
+            add_nodes(ocs_nodes=False, node_count=num_of_wnodes_to_add)
```
ebenahar, line 144: shouldn't we pass `True` doe `ocs_nodes`?

yitzhak12: I label all the worker nodes in line 149 `label_nodes(wnodes_not_in_ocs)`.

yitzhak12: Fixed

ebenahar (dismissed):

shreekarSS (approved):

tests/manage/z_cluster/cluster_expansion/conftest.py
```diff
@@ -107,3 +110,40 @@ def factory(ocs_nodes=False, node_count=3, taint_label=None):
...
+    if number_of_osd_pods_before >= constants.MAX_OSDS:
+        pytest.skip("We have maximum of OSDs in the cluster")
+
+    # If we use vSphere we may need to add more worker nodes
+    # to the cluster before starting the test
+    if (
+        config.ENV_DATA["platform"].lower() == constants.VSPHERE_PLATFORM
+        and number_of_osd_pods_before >= 9
+    ):
+        num_of_expected_wnodes = 6
```
prsurve, line 132: i feel this should not be hardcoded we can what if add capacity is done more then 1 time

prsurve:
If I'm not wrong this will always create new node and then do add capacity if yes then we are missing a test where we do add capacity on existing node

tests/manage/z_cluster/cluster_expansion/conftest.py
```diff
@@ -107,3 +110,40 @@ def factory(ocs_nodes=False, node_count=3, taint_label=None):
...
+            num_of_wnodes_to_add = num_of_expected_wnodes - num_of_wnodes
+            logging.info(
+                f"Adding more {num_of_wnodes_to_add} worker nodes to the cluster"
+            )
+            add_nodes(ocs_nodes=True, node_count=num_of_wnodes_to_add)
+
+        wnodes_not_in_ocs = get_worker_nodes_not_in_ocs()
+        if wnodes_not_in_ocs:
+            logging.info("Label the worker nodes that are not in OCS")
+            label_nodes(wnodes_not_in_ocs)
```
prsurve, line 149: Why we are adding label for non ocs node this step should be optional

tests/manage/z_cluster/cluster_expansion/conftest.py
yitzhak12: I don't think we need to run the "add capacity" test with more than 6 worker nodes.

tests/manage/z_cluster/cluster_expansion/conftest.py
yitzhak12: All the 6 worker nodes should be OCS labeled, so we can use them for the newly created OSDs.

yitzhak12:
@prsurve We add new worker nodes, only if we have more than 9 OSD pods. 3 OCS worker nodes will not be enough for more than 9 OSD pods, that's why we add more OCS worker nodes. It shouldn't change the process of the test. Maybe with LSO, we can add only 1 OCS worker node.

tests/manage/z_cluster/cluster_expansion/conftest.py
prsurve: dont 144 line do for us create new node and label it for OCS

ocs_ci/ocs/node.py
```diff
@@ -1686,3 +1688,19 @@ def get_node_mon_ids(node_name):
     mon_pods = pod.get_mon_pods()
     node_mon_pods = get_node_pods(node_name, pods_to_search=mon_pods)
     return [pod.get_mon_pod_id(mon_pod) for mon_pod in node_mon_pods]
+
+
+def add_new_disk_for_vsphere(sc_name):
```
annagitel, line 1693: is this function used only on vsphere deployments? if so, shouldn't be a check for deployment type? with a error or skip if the type is not vmware?

yitzhak12: My assumption is that if someone uses this function, he will use it only for vSphere.

tests/manage/z_cluster/cluster_expansion/conftest.py
yitzhak12: Line 144 adds new nodes and OCS labels them. But if we already have worker nodes(before starting the test) that are not labeled - then we need this line to label them.

yitzhak12: For example, let's say we have 3 OCS worker nodes, and 2 worker nodes not labeled. The fixture will add one worker node and to reach 6 worker nodes and OCS labeled the node. Then it will take the other 2 worker nodes that are not labeled, and OCS labeled them.

OdedViner (approved):

annagitel (approved):
LGTM

petr-balogh (approved):

OdedViner (approved):

yitzhak12:
I ran the test of `tests/manage/z_cluster/cluster_expansion/test_delete_pod.py` in my local, and it passed. Here is the link to the logs: http://rhsqe-repo.lab.eng.blr.redhat.com/OCS/ocs-qe-bugs/pr-4654-logs/

PR #4655, opened 2021-08-02
Added ipi_deployment_required marker in scale tests where node scaling is done for app pods.

ramkiperiy:
In scale tests like creation of 1000+ PVC's and PODs we need more worker nodes(other than OCS nodes)
for running app pods and all the relevant tests supports only IPI platform.

### DIFF
--- a/tests/e2e/scale/test_ceph_pod_respin_in_scaled_cluster.py
+++ b/tests/e2e/scale/test_ceph_pod_respin_in_scaled_cluster.py
@@ -6,7 +6,10 @@ from ocs_ci.ocs import constants
 from ocs_ci.utility import utils
 from ocs_ci.ocs.scale_lib import FioPodScale
 from ocs_ci.framework.testlib import scale, E2ETest, ignore_leftovers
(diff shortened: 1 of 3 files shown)


Avilir (approved):
LGTM

keesturam (approved):
LGTM.
@ramkiperiy  Should we also raise an issue to allow these tests to be run on upi deployments?

ramkiperiy:
PR validation not required for this PR, since added this marker to skip UPI deployment type run executions.
**Reason:** In scale tests like creation of 1000+ PVC's and PODs we need more worker nodes(other than OCS nodes)
for running app pods and all the relevant tests supports only IPI platform.

Without this marker tests will be executed in UPI and fails with "Unsupported platform" exception, I think it's better to skip the test at the start itself than failing with Unsupported platform exception.

ramkiperiy:
> LGTM.
> @ramkiperiy Should we also raise an issue to allow these tests to be run on upi deployments?

We should add support scaling nodes in UPI deployments, and UPI or IPI scaling of PVCs and PODs not going to differ much IMO.

PR #4656, opened 2021-08-02
Fix bucketclass UI tests

MeridianExplorer:
The tests failed because, although the checkbox component could be found in time, it seems that the click on it would not register without a slight wait before and after the action.


OdedViner (approved):
LGTM

ebenahar (approved):

bernerhat (approved):

Issue #4659, opened 2021-08-02
deprecate command in test: test_2_nodes_different_types


AviadP:
```
oc adm drain compute-0 control-plane-0 --force=true --ignore-daemonsets --delete-local-data
Flag --delete-local-data has been deprecated, This option is deprecated and will be deleted. Use --delete-emptydir-data.
```

AviadP:
also, need to remove `force=true` flag

PR #4662, opened 2021-08-03
made force flag optional, updated deprecate flag

AviadP:
fix issue #4659 

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -185,11 +185,12 @@ def schedule_nodes(node_names):
     wait_for_nodes_status(node_names)
 
 
-def drain_nodes(node_names):
+def drain_nodes(node_names, force="true"):
     """
     Drain nodes
 
     Args:
+        force:
         node_names (list): The names of the nodes
 
     Raises:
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/node.py
```diff
@@ -201,8 +202,8 @@ def drain_nodes(node_names):
     log.info(f"Draining nodes {node_names_str}")
     try:
         ocp.exec_oc_cmd(
-            f"adm drain {node_names_str} --force=true --ignore-daemonsets "
-            f"--delete-local-data",
+            f"adm drain {node_names_str} --force={force} --ignore-daemonsets "
+            f"--delete-emptydir-data",
```
prsurve, line 206: this has to be tested with OCP 4.6,4.7,4.8 to make sure this param is aviabale

ocs_ci/ocs/node.py
```diff
@@ -201,8 +202,8 @@ def drain_nodes(node_names):
     log.info(f"Draining nodes {node_names_str}")
     try:
         ocp.exec_oc_cmd(
-            f"adm drain {node_names_str} --force=true --ignore-daemonsets "
-            f"--delete-local-data",
+            f"adm drain {node_names_str} --force={force} --ignore-daemonsets "
```
prsurve, line 205: we have to use force=true always

ocs_ci/ocs/node.py
AviadP: tested with 4.8, and will do on 4.6 and 4.7

PR #4668, opened 2021-08-04
Use released RHCOS 4.7 image for BM deployment

prsurve:


clacroix12:
@prsurve @petr-balogh I assume we want to cherry pick this to our 4.7 branch as well?

petr-balogh:
Do we use this file only for 4.7 deployment? Or 4.8/4.9 as well? @prsurve ? @clacroix12 depends on the answer we need to cherry-pick maybe also to the release-4.8 branch.

prsurve:
> or 4.7 deployment?
we use this file for all the OCP deployments

petr-balogh:
> > or 4.7 deployment?
> > we use this file for all the OCP deployments

Yesterday I had this PR opened just on my cell phone so it was not quite good visible as is on laptop that here we have the key for `4.7` only. So this changed section is changing this only for 4.7 deployment - it is not used anywhere else, right?

PR #4674, opened 2021-08-05
[WIP]Test case to verify RBD RWX PVC is consumable after pod deletion

jilju:
Test case to verify RBD RWX PVC is consumable after pod deletion.
OCS-2599 - RBD RWX: While two app pods access same mount, delete one pod. Confirm second pod is still able to write


jilju:
Closing the PR based on bugzilla comment [1990491](https://bugzilla.redhat.com/show_bug.cgi?id=1990491#c2)

Issue #2070, opened 2020-05-13
Add Stale Bot for ocs-ci issues


vasukulkarni:
we can add stale bot to osc-ci to better maintain active issues and close the ones that dont get attention or updates for like 4 to 6 months

vasukulkarni:
The option to unstable is present in the bot.

clacroix12:
I think using this github action is the right call: https://github.com/marketplace/actions/close-stale-issues

We can configure days-before-stale for issues and pull requests independently, and also specify a label to use which will have this action ignore certain long living issues or pull requests.

Initial values for those were determined in the automation syncup today:

Default date for stale PR: 90 days? Another 30 days to close.
Default date for stale Issue: 365 days? Another 30 days to close.

Essentially the action will mark items with a `stale` label and notify people watching the issue or PR that there has been no activity. 30 days afterwards it will close the issue/PR out if there are still no updates. These values are intended to be very generous to start since we can always reduce them once we get things under control and the process is reliable.

PR #4676, opened 2021-08-05
Add github action for labeling and closing stale issues and pull requests

clacroix12:
Fixes: #2070 

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,28 @@
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: '0 * * * *'
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+    - uses: actions/stale@v4
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-issue-stale: 365
+        days-before-pr-stale: 90
+        days-before-close: 30
+        stale-issue-message: 'There has been no activity on this issue in 365 days. It will be marked as stale and closed in 30 days if no actions are taken.'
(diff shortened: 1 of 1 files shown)


clacroix12:
I have added this action config with `debug-only: true` so we don't take any action on items until we know this is executing properly. We should be able to review logs to verify prior to removing this from the config.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,28 @@
...
+
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+    - uses: actions/stale@v4
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
```
vavuthu, line 18: is GITHUB_TOKEN added into secrets?

clacroix12: This was the value in the yaml when I went to create a new workflow in the github UI so I kept it. Not sure if there is another way to do this but I figured it was a variable defined by github and not by us.

clacroix12: It looks like this will automatically be created when we enable this workflow.

> GitHub automatically creates a GITHUB_TOKEN secret to use in your workflow. You can use the GITHUB_TOKEN to authenticate in a workflow run.

https://docs.github.com/en/actions/reference/authentication-in-a-workflow

vavuthu (approved):

clacroix12:
I have changed the cron schedule to run this every hour to make testing easier. I still have `debug-only:true` set so we shouldn't see any actions taken on issues. We will need to merge this PR in order for me to check the logs and verify that the workflow is working as intended.

jimdevops19 (approved):

shylesh (approved):
LGTM

vasukulkarni:
This is great! 👍

PR #4686, opened 2021-08-08
Upgrading the Elasticsearch module to use ECK 1.6.0 and ES 7.14

Avilir:


ocs_ci/ocs/elasticsearch.py
```diff
@@ -53,21 +55,17 @@ def get_data_from_text_file(json_file):
             l.strip() for l in open(str(json_file), encoding="utf8", errors="ignore")
         ]
         log.info(f"String docs length: {len(docs)}")
-        doc_list = []
 
         for num, doc in enumerate(docs):
             try:
                 dict_doc = json.loads(doc)
-                doc_list += [dict_doc]
+                yield dict_doc
```
keesturam, line 62: Nice yield implementation @Avilir

ocs_ci/ocs/elasticsearch.py
```diff
@@ -271,7 +278,8 @@ def cleanup(self):
         log.info("Deleting the es resource")
         self.ocp.delete(yaml_file=self.crd)
         log.info("Deleting the es project")
-        self.ns_obj.delete_project(project_name=self.namespace)
+        # self.ns_obj.delete_project(project_name=self.namespace)
+        self.ocp.delete(yaml_file=self.eck_file)
```
keesturam, line 282: As part of the cleanup, you should also delete the PVC created in line 216

keesturam (commented):
great to see the changes & code optimization @Avilir

ocs_ci/ocs/elasticsearch.py
Avilir: @keesturam PVC can not be deleted before the pod which use it deleted, and the pod is deleted on line 279 which delete in the end the all project.
when the project deleted, there is no PVC to delete, so in the end, the cleanup delete also the PVC.

keesturam (approved):
LGTM

keesturam:
@Avilir can we also run a sanity run of all performance tests that needs elasticsearch. This is to make sure that the change doesn't affect any existing tests.

Avilir:
> @Avilir can we also run a sanity run of all performance tests that needs elasticsearch. This is to make sure that the change doesn't affect any existing tests.

yes we can, but it will take a lot of time.
will try to run in parallel, one PR validation with SF, and one with FIO

OdedViner (approved):
LGTM

PR #4705, opened 2021-08-10
Fix issue with retrieving ocp channel

Avilir:
This PR fixing issue with IBM Cloud (or any other platform) which the OCP channel is not configure.

in this case instead of the the test will failed because the we do not found the OCP channel in the output of the command `oc get clusterversion -o json` it will return "None" - this is the string None and not None value since elastic-search an not index the None value
 

keesturam (approved):
LGTM

ramkiperiy (approved):

PR #4741, opened 2021-08-19
Change the condition to wait for the new osd node to come up after node replacement process

yitzhak12:


tests/manage/z_cluster/nodes/test_node_replacement_proactive.py
```diff
@@ -58,8 +58,32 @@ def check_node_replacement_verification_steps(
...
+    min_osd_nodes = 3
+    ocs_nodes = node.get_ocs_nodes()
+
+    if len(old_osd_node_names) <= min_osd_nodes:
+        log.info(
+            "We have the minimum number of osd nodes in the cluster. "
+            "Wait for the new created worker node to appear in the osd nodes"
+        )
+        new_osd_node_name = node.wait_for_new_osd_node(old_osd_node_names, timeout=1500)
+        assert new_osd_node_name, "New osd node not found"
```
ebenahar, line 70: please make the assert messages more elaborative. 
Maybe something like:
```
"New osd node not found after <operation> while waiting for <timeout>"
```

yitzhak12: Yes, I will change it.

ocs_ci/ocs/node.py
```diff
@@ -1233,17 +1233,27 @@ def node_replacement_verification_steps_ceph_side(
 
     ct_pod = pod.get_ceph_tools_pod()
     ceph_osd_status = ct_pod.exec_ceph_cmd(ceph_cmd="ceph osd status")
-    if new_osd_node_name not in ceph_osd_status:
```
annagitel, line 1236: I don't understand this condition and the change in line 1209 
if no `new_osd_node_name` was passed it will be None by default and the func will return false by default 
so why is this `=None` in row 1209 needed?

annagitel: so I kept reading to the end of the if/if/else and I thing that the logic of the code is not right 
I thing that the if part from line 1236 to line 1238 should inside the if condition in row 1240

tests/manage/z_cluster/nodes/test_node_replacement_proactive.py
yitzhak12: Fixed

annagitel (approved):
accidentally reviewed the red rows 
besides LGTM

ebenahar (approved):

yitzhak12:
Okay, thanks @annagitel

PR #4747, opened 2021-08-21
Installation of vSphere UPI cluster behind proxy via Flexy

dahorak:
* implementation of UPI installation of cluster behind proxy on vSphere (OCP installation is done by Flexy)
* introduce new parameter `DEPLOYMENT["proxy"]` to indicate cluster installed on proxy environment
* introduce new parameter `ENV_DATA["client_http_proxy"]` which can be used as proxy configuration for clients accessing the OCP cluster

It is now available on _vSphere DC-ECO, DC-CP, DC-PS and vSphere7 DC-CP_ (depends also on _ocs4-jenkins MR!736_).


dahorak:
All _acceptance_ tests are passing, results from _tier1_ test suite looks like this: _136 passed, 2 failed, 16 errors_.
UI tests doesn't work, because of the requirement to connect to the console through proxy server - this will require more investigation, because it seems, like there is no easy way to configure authenticated proxy in chrome (in Selenium). There is some problem in tests using _vault_ and _terraform_, will try to investigate the failures and consider fix in this PR or in new one.

ebenahar (approved):

PR #4750, opened 2021-08-22
Deployment and Add capacity tests via UI on Azure Internal cluster

OdedViner:

### DIFF
--- a/ocs_ci/ocs/ui/base_ui.py
+++ b/ocs_ci/ocs/ui/base_ui.py
@@ -53,6 +53,8 @@ class BaseUI:
             self.storage_class = "thin_sc"
         elif config.ENV_DATA["platform"].lower() == constants.AWS_PLATFORM:
             self.storage_class = "gp2_sc"
+        elif config.ENV_DATA["platform"].lower() == constants.AZURE_PLATFORM:
+            self.storage_class = "managed-premium_sc"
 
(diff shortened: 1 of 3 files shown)


ocs_ci/ocs/ui/base_ui.py
```diff
@@ -53,6 +53,8 @@ def __init__(self, driver):
             self.storage_class = "thin_sc"
         elif config.ENV_DATA["platform"].lower() == constants.AWS_PLATFORM:
             self.storage_class = "gp2_sc"
+        elif config.ENV_DATA["platform"].lower() == constants.AZURE_PLATFORM:
+            self.storage_class = "managed-premium_sc"
```
ebenahar, line 57: can we save "managed-premium_sc" in a constants file? I guess this would imply saving all the above defaults SCs there

OdedViner: We use the string `managed-premium_sc` to find the element..
for example:  
https://github.com/red-hat-storage/ocs-ci/blob/5be780a7a7daafa391cd92686c8e2f661d34dc5d/ocs_ci/ocs/ui/views.py#L333

OdedViner: we discussed on it offline

petr-balogh (approved):

ebenahar (approved):

PR #4752, opened 2021-08-23
Do not wait for re-balance to complete if not needed

ebenahar:
In test cases that cause PGs rebalance, instead of waiting for rebalance to complete, this change suggests moving the next test case to be executed last and move on to a test case that does not require PGs to be active+clean


tests/manage/z_cluster/nodes/test_nodes_maintenance.py
```diff
@@ -80,6 +80,8 @@ class TestNodesMaintenance(ManageTest):
 
     """
 
+    REBALANCE_COMPLETED_NEEDED = True
```
yitzhak12, line 83: Why did you put it in this file?

tests/conftest.py
```diff
@@ -204,6 +205,30 @@ def pytest_collection_modifyitems(session, items):
...
+    completed, in case PG re-balance completion is needed for it
+    """
+    yield
+    idx = item.session.items.index(item)
+    remaining = item.session.items[idx + 1 :]
+    if nextitem.cls.REBALANCE_COMPLETED_NEEDED:
+        cluster = CephCluster()
+        rebalance_status = cluster.get_rebalance_status()
+        if not rebalance_status:
+            if config.RUN["first_pushed_to_last_item"] is not nextitem:
```
yitzhak12, line 221: Can you explain more about the meaning of this condition?

petr-balogh:
I think that this also has to be well documented - so please add it to some of the documentation file.

mbukatov (commented):
The ocs-ci test are neither classified nor isolated so that we know which test cases requires particular level of cluster access. Letting this in current state would only increase effort required for CI analysis, and created unnecessary test failures.

I'm sure I discussed this somewhere already, but I'm adding it here since it was brought up today.

ebenahar:
Re-opening for re-examination

petr-balogh:
@ebenahar can this be closed or still relevant?

PR #4759, opened 2021-08-24
fix benchmark test from run for indefinitely

Avilir:
there is a bug in the performance benchmarks tests which in case of Failed status, it will run indefinitely, since it is waiting for Completed state.
This PR is fixing this issue.

### DIFF
--- a/ocs_ci/ocs/perftests.py
+++ b/ocs_ci/ocs/perftests.py
@@ -5,19 +5,19 @@ import re
 
 from elasticsearch import Elasticsearch
 
-from ocs_ci.framework.testlib import BaseTest
-
-from ocs_ci.ocs import defaults, constants, node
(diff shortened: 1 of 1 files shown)


Avilir:
> PR validation
> 
> Cluster Name:
> Cluster Configuration:
> PR Test Suite: performance
> PR Test Path: tests/e2e/performance/test_small_file_workload.py
> Additional Test Params:
> OCP VERSION: 4.8
> OCS VERSION: 4.8
> 
> [Job UNSTABLE](https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-trigger-test-pr/1358//testReport) (some or all tests failed).

The failures are not related to the fix in this PR, but without the fix the test will fun indefinitely. so they just show that the fix is working.

PR #4768, opened 2021-08-26
Add VAULT_BACKEND parameter as a work around for BZ 1997624

rachael-george:
Adding VAULT_BACKEND parameter to the csi-kms-connection-details to allow creation of encrypted PVs with the vault namespace functionality
Eng. Bug: [1997624](https://bugzilla.redhat.com/show_bug.cgi?id=1997624)

### DIFF
--- a/ocs_ci/ocs/defaults.py
+++ b/ocs_ci/ocs/defaults.py
@@ -84,5 +84,6 @@ VAULT_CSI_CONNECTION_CONF = {
         "VAULT_CACERT_FILE": "fullchain.pem",
         "VAULT_CLIENT_CERT_FILE": "cert.pem",
         "VAULT_CLIENT_KEY_FILE": "privkey.pem",
+        "VAULT_BACKEND": "kv-v2",
     }
 }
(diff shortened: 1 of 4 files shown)


ocs_ci/ocs/defaults.py
```diff
@@ -84,5 +84,6 @@
         "VAULT_CACERT_FILE": "fullchain.pem",
         "VAULT_CLIENT_CERT_FILE": "cert.pem",
         "VAULT_CLIENT_KEY_FILE": "privkey.pem",
+        "VAULT_BACKEND": "kv-v2",
```
shyRozen, line 87: Wouldn't it be better to leave "1-vault" for v1 and create "2-vault" for v2?
By the way I used "VAULT_BACKEND: v2", does "kv-v2" works also in configmap?

rachael-george: In the test case, this is replaced by custom key names like `vault-test-xxxxxxxx`, so that there are no conflicts with any existing keys in the configmap. 
For PV encryption `kv` and `kv-v2` are the the values that ceph-csi recognises. (https://bugzilla.redhat.com/show_bug.cgi?id=1983931#c16)

ebenahar (approved):

shyRozen (approved):

Issue #4764, opened 2021-08-25
SameNamePrefixClusterAlreadyExistsException when using j- prefix locally for AWS UPI deployment


clacroix12:
```
self = <ocs_ci.deployment.aws.AWSUPI object at 0x7fcc98269ef0>, log_cli_level = 'INFO'

    def deploy_ocp(self, log_cli_level="DEBUG"):
        """
        Deployment specific to OCP cluster on a cloud platform.

        Args:
            log_cli_level (str): openshift installer's log level
                (default: "DEBUG")

        """
        if not config.DEPLOYMENT.get("force_deploy_multiple_clusters"):
            cluster_name_parts = self.cluster_name.split("-")
            prefix = cluster_name_parts[0]
            import pdb
            pdb.set_trace()
>           if prefix.lower() in self.cluster_prefixes_special_rules:
E           ocs_ci.ocs.exceptions.SameNamePrefixClusterAlreadyExistsException: Cluster with name prefix j already exists. Please destroy the existing cluster for a new cluster deployment

ocs_ci/deployment/cloud.py:66: SameNamePrefixClusterAlreadyExistsException
```

When trying to deploy a cluster locally using the cluster name `j-999au3r33-t4an` I am seeing the above error. This is due to two things:

1. The way we are parsing out the "prefix" that is in use (not to be confused with the CLUSTER_PREFIX from jenkins, but essentially the provided cluster name without the test tier addition at the end). We are splitting the string on hyphens and grabbing chunks. This is unreliable as we have prefixes with hyphens in them as well as the fact we can't know if the cluster name has hyphens in it either. It is also why the above error considers `j` to be the prefix while we really want to check for `j-\\d\\d\\d`. We should instead be checking if the provided cluster name starts with one of our exception stings/patterns.
2. The [dictionary](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/cloud.py#L39) we are checking to see if a cluster "prefix" is one of our defined cluster prefixes in order to dismiss it is instantiated as an empty dict. We are [updating](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/aws.py#L176) this value in the `AWSIPI` class, but not the `AWSUPI` class. We should probably move this to `AWSBase` or instantiate it in `CloudDeploymentBase` this way if we want the same rules to apply to all cloud providers.

PR #4771, opened 2021-08-26
Cluster prefix match fix

clacroix12:
Fixes: #4764


petr-balogh (dismissed):

vavuthu (approved):

clacroix12:
I moved the logic to a utility function and added unit tests for various cluster name scenarios. I think as long as the scenarios we are parametrizing the unit test with are sufficient for all of our use cases this can be considered verified, however I also kicked off a deployment test for some additional testing.

dahorak (approved):

petr-balogh (approved):

PR #4804, opened 2021-09-03
Test rolling shutdown and recovery of OCS worker nodes

PrasadDesala:
Automates BZ 1895819.


tests/manage/z_cluster/nodes/test_rolling_shutdown_and_recovery.py
```diff
@@ -0,0 +1,78 @@
...
+        log.info("ShutDown OCS worker")
+        for node_obj in ocs_node_objs:
+            nodes.stop_nodes(nodes=[node_obj])
+            log.info("Keeping node in stopped state for 3 mins")
+            time.sleep(180)
+            nodes.start_nodes(nodes=[node_obj])
+            self.sanity_helpers.health_check(cluster_check=False, tries=60)
+            log.info("Checking storage pods status")
+            # Validate storage pods are running
+            wait_for_storage_pods()
```
yitzhak12, line 73: You can use here also the function `def check_pods_in_running_state`.

tests/manage/z_cluster/nodes/test_rolling_shutdown_and_recovery.py
```diff
@@ -0,0 +1,78 @@
...
+
+        """
+        # Get OCS worker node objects
+        ocs_node_objs = get_ocs_nodes()
+
+        # Start rolling shutdown and recovery of OCS worker nodes
+        log.info("ShutDown OCS worker")
+        for node_obj in ocs_node_objs:
+            nodes.stop_nodes(nodes=[node_obj])
+            log.info("Keeping node in stopped state for 3 mins")
```
yitzhak12, line 67: It will be better to define a var with the timeout, like `seconds_to_wait = 180` instead of writing a number every time.

tests/manage/z_cluster/nodes/test_rolling_shutdown_and_recovery.py
PrasadDesala: check_pods_in_running_state function seems added recently and both these functions are serving the same purpose.

tests/manage/z_cluster/nodes/test_rolling_shutdown_and_recovery.py
PrasadDesala: Done

tests/manage/z_cluster/nodes/test_rolling_shutdown_and_recovery.py
yitzhak12: There is an issue with the function `wait_for_storage_pods`. If the pod names are changed in the waiting time, it will not get updated - and then we can miss new created or deleted pods in the waiting time. That's why I created this function `wait_for_pods_to_be_running` https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/resources/pod.py#L1639 that get all the pod objects in every iteration.

PrasadDesala: Done, replaced with  wait_for_storage_pods with wait_for_pods_to_be_running().

yitzhak12 (approved):

tests/manage/z_cluster/nodes/test_rolling_shutdown_and_recovery.py
```diff
@@ -0,0 +1,79 @@
...
+log = logging.getLogger(__name__)
+
+
+@tier4c
+@pytest.mark.polarion_id("OCS-2633")
+@bugzilla("1895819")
+@skipif_ibm_cloud
+@skipif_external_mode
+@ignore_leftovers
+class TestRollingWorkerNodeShutdownAndRecovery(ManageTest):
```
ebenahar, line 28: the test is quite similar to tests.manage.cluster.nodes.test_nodes_restart.TestNodesRestart.test_rolling_nodes_restart

Do we need both?

PrasadDesala: @ebenahar CU hit the SIGSEG on a mon node during node shutdown. Neha ojha from Dev who fixed the issue suggested us to have test cases for both shutdown and reboot case and hence I added this new test for rolling node shutdown.

ebenahar (approved):

PR #4808, opened 2021-09-07
Restore snapshot to different pool

jilju:
Test case test_snapshot_restore_using_different_sc is updated to use a storage class using different pool. This will verify the bug 1901954.


rachael-george (approved):
LGTM

Akarsha-rai (approved):

Issue #4809, opened 2021-09-07
PVC on cluster validation can fail when OSD pod get re-spun


petr-balogh:
```
tests/ecosystem/upgrade/test_upgrade.py:68: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ocs_ci/ocs/ocs_upgrade.py:568: in run_ocs_upgrade
    ocs_install_verification(
ocs_ci/ocs/resources/storage_cluster.py:415: in ocs_install_verification
    validate_cluster_on_pvc()
ocs_ci/ocs/cluster.py:889: in validate_cluster_on_pvc
    out = run_cmd(f"oc -n {ns} get pods {ceph_pod} -o yaml")
ocs_ci/utility/utils.py:456: in run_cmd
    completed_process = exec_cmd(cmd, secrets, timeout, ignore_error, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cmd = ['oc', '-n', 'openshift-storage', 'get', 'pods', 'rook-ceph-osd-0-6c86877fd6-kscjs', ...]
secrets = None, timeout = 600, ignore_error = False, kwargs = {}
masked_cmd = 'oc -n openshift-storage get pods rook-ceph-osd-0-6c86877fd6-kscjs -o yaml'
completed_process = CompletedProcess(args=['oc', '-n', 'openshift-storage', 'get', 'pods', 'rook-ceph-osd-0-6c86877fd6-kscjs', '-o', 'yaml'], returncode=1, stdout=b'', stderr=b'Error from server (NotFound): pods "rook-ceph-osd-0-6c86877fd6-kscjs" not found\n')
masked_stdout = ''
masked_stderr = 'Error from server (NotFound): pods "rook-ceph-osd-0-6c86877fd6-kscjs" not found\n'

    def exec_cmd(cmd, secrets=None, timeout=600, ignore_error=False, **kwargs):
        """
        Run an arbitrary command locally
    
        Args:
            cmd (str): command to run
            secrets (list): A list of secrets to be masked with asterisks
                This kwarg is popped in order to not interfere with
                subprocess.run(``**kwargs``)
[...]

PR #4810, opened 2021-09-07
Re-try cluster on PVC validation

petr-balogh:
Fixes: #4809

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -841,9 +841,9 @@ def validate_claim_name_match_pvc(pvc_names, validated_pods=None):
     Args:
         pvc_names (list): names of all PVCs you would like to validate with.
         validated_pods(set): set to store already validated pods - if you pass
-            empty set outside of this function, this will speed up the next
-            validation when re-tries as it will skip those already validated
-            pods stored in this set.
+            an empty set from outside of this function, it will speed up the next
+            validation when re-tries, as it will skip those already validated
+            pods added to this set by previous run of this function.
     Raises:
         AssertionError: when the claim name does not match one of PVC name.
 
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -16,7 +16,7 @@ import time
 from semantic_version import Version
 
 import ocs_ci.ocs.resources.pod as pod
-from ocs_ci.ocs.exceptions import UnexpectedBehaviour
+from ocs_ci.ocs.exceptions import CommandFailed, UnexpectedBehaviour
 from ocs_ci.ocs.resources import ocs, storage_cluster
 import ocs_ci.ocs.constants as constant
 from ocs_ci.ocs import defaults
@@ -833,6 +833,37 @@ def validate_ocs_pods_on_pvc(pods, pvc_names, pvc_label=None):
             ), f"No PVC {pvc_name} found for pod: {pod_name} in PVCs: {pvc_names}!"
 
 
+@retry(CommandFailed, tries=3, delay=10, backoff=1)
+def validate_claim_name_match_pvc(pvc_names, validated_pods=None):
+    """
+    Validate if OCS pods have mathching PVC and Claim name
+
+    Args:
+        pvc_names (list): names of all PVCs you would like to validate with.
+        validated_pods(set): set to store already validated pods - if you pass
+            empty set outside of this function, this will speed up the next
+            validation when re-tries as it will skip those already validated
+            pods stored in this set.
+    Raises:
+        AssertionError: when the claim name does not match one of PVC name.
+
+    """
+    if validated_pods is None:
+        validated_pods = set()
+    ns = config.ENV_DATA["cluster_namespace"]
+    mon_pods = get_pod_name_by_pattern("rook-ceph-mon", ns)
+    osd_pods = get_pod_name_by_pattern("rook-ceph-osd", ns, filter="prepare")
+    for ceph_pod in set(mon_pods + osd_pods) - validated_pods:
+        out = run_cmd(f"oc -n {ns} get pods {ceph_pod} -o yaml")
+        out_yaml = yaml.safe_load(out)
+        for vol in out_yaml["spec"]["volumes"]:
+            if vol.get("persistentVolumeClaim"):
+                claimName = vol.get("persistentVolumeClaim").get("claimName")
+                logger.info(f"{ceph_pod} backed by pvc {claimName}")
+                assert claimName in pvc_names, "Ceph Internal Volume not backed by PVC"
+        validated_pods.add(ceph_pod)
+
+
 def validate_cluster_on_pvc():
     """
     Validate creation of PVCs for MON and OSD pods.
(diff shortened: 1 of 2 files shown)


petr-balogh:
Verification job: https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-trigger-aws-ipi-3az-rhcos-3m-3w-upgrade-ocs-auto/72

ocs_ci/ocs/cluster.py
```diff
@@ -884,15 +884,33 @@ def validate_cluster_on_pvc():
...
-    osd_pods = get_pod_name_by_pattern("rook-ceph-osd", ns, filter="prepare")
-    for ceph_pod in mon_pods + osd_pods:
-        out = run_cmd(f"oc -n {ns} get pods {ceph_pod} -o yaml")
-        out_yaml = yaml.safe_load(out)
-        for vol in out_yaml["spec"]["volumes"]:
-            if vol.get("persistentVolumeClaim"):
-                claimName = vol.get("persistentVolumeClaim").get("claimName")
-                logger.info(f"{ceph_pod} backed by pvc {claimName}")
-                assert claimName in pvc_names, "Ceph Internal Volume not backed by PVC"
+    retry_validation = True
```
clacroix12, line 887: Is the [retry decorator](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/retry.py) not sufficient for this use case?

petr-balogh: I wanted to do like that but then I added the logic to speed up the process to skip already validated pods.  I was thinking about how to change it here and came up with the other solution where I will pass this validated_pods to the function from outside, so the next iteration of re-try should be quicker and skip those already validated pods.

petr-balogh:
Next validation:
https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/1776/console

ocs_ci/ocs/cluster.py
```diff
@@ -833,6 +833,37 @@ def validate_ocs_pods_on_pvc(pods, pvc_names, pvc_label=None):
...
+        validated_pods(set): set to store already validated pods - if you pass
+            empty set outside of this function, this will speed up the next
+            validation when re-tries as it will skip those already validated
+            pods stored in this set.
+    Raises:
+        AssertionError: when the claim name does not match one of PVC name.
+
+    """
+    if validated_pods is None:
+        validated_pods = set()
```
clacroix12, line 852:
```suggestion
    validated_pods = validated_pods or set()
```

ocs_ci/ocs/cluster.py
```diff
@@ -884,15 +915,8 @@ def validate_cluster_on_pvc():
...
-    for ceph_pod in mon_pods + osd_pods:
-        out = run_cmd(f"oc -n {ns} get pods {ceph_pod} -o yaml")
-        out_yaml = yaml.safe_load(out)
-        for vol in out_yaml["spec"]["volumes"]:
-            if vol.get("persistentVolumeClaim"):
-                claimName = vol.get("persistentVolumeClaim").get("claimName")
-                logger.info(f"{ceph_pod} backed by pvc {claimName}")
-                assert claimName in pvc_names, "Ceph Internal Volume not backed by PVC"
+    validated_pods = set()
+    validate_claim_name_match_pvc(pvc_names, validated_pods)
```
clacroix12, line 919: Since we default validated_pods to None and create a set in the function if it is None, can we skip initializing it here and passing through an empty set?

```suggestion
    validate_claim_name_match_pvc(pvc_names)
```

clacroix12 (commented):
Just a couple small things I noticed but functionally looks right to me. @petr-balogh just to verify we do not want to retry this function if we get an `AssertionError`. We only want to retry if we get a `CommandFailed` retrieving pod data, correct?

ocs_ci/ocs/cluster.py
petr-balogh: This will not work. As I would like to pass empty set outside of this function.
As it's a mutable object, it will still refer to the same one and each iteration it will adding pods to the same object.

Empty set will be validated as False so in this  case you will get always empty set from inside the function which will get each iteration re-created. 

see:
```
In [1]: set() or "You don't want this"
Out[1]: "You don't want this"
``` 
Hence I really need to validate if it's None.

ocs_ci/ocs/cluster.py
petr-balogh: I explained in comment above, as I would like to speed up process in next re-try I need to use the set from outside of the function.

petr-balogh:
> Just a couple small things I noticed but functionally looks right to me. @petr-balogh just to verify we do not want to retry this function if we get an `AssertionError`. We only want to retry if we get a `CommandFailed` retrieving pod data, correct?

Yes, that's correct. I would like to repeat only if it fails to get that pod because of pod got re-created with new name. The other errors are valid and we should not re-try in this case.

Hence I wanted to implement own re-try functionality to cover only that specific part :).

ocs_ci/ocs/cluster.py
clacroix12: Got it, that makes sense! LGTM then

clacroix12 (approved):

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -438,9 +438,11 @@ def set_upgrade_images(self):
         )
 
         if not self.upgrade_in_current_source:
-            if not ocs_catalog.is_exist() and not self.upgrade_in_current_source:
+            if not ocs_catalog.is_exist():
                 log.info("OCS catalog source doesn't exist. Creating new one.")
                 create_catalog_source(self.ocs_registry_image, ignore_upgrade=True)
+                # We can return here as new CatalogSource contans right images
```
dahorak, line 444:
```suggestion
                # We can return here as new CatalogSource contains right images
```

ocs_ci/ocs/cluster.py
```diff
@@ -833,6 +833,37 @@ def validate_ocs_pods_on_pvc(pods, pvc_names, pvc_label=None):
...
 
+@retry(CommandFailed, tries=3, delay=10, backoff=1)
+def validate_claim_name_match_pvc(pvc_names, validated_pods=None):
+    """
+    Validate if OCS pods have mathching PVC and Claim name
+
+    Args:
+        pvc_names (list): names of all PVCs you would like to validate with.
+        validated_pods(set): set to store already validated pods - if you pass
+            empty set outside of this function, this will speed up the next
```
dahorak, line 844:
```suggestion
            an empty set from outside of this function, it will speed up the next
```

ocs_ci/ocs/cluster.py
```diff
@@ -833,6 +833,37 @@ def validate_ocs_pods_on_pvc(pods, pvc_names, pvc_label=None):
...
+            validation when re-tries as it will skip those already validated
+            pods stored in this set.
```
dahorak, line 846: Wouldn't be this slightly more clear?
```suggestion
            pods added to this set by previous run of this function.
```

dahorak (approved):
Just small typo and comment update suggestion, otherwise LGTM.

ocs_ci/ocs/cluster.py
petr-balogh: fixed, thanks

ocs_ci/ocs/ocs_upgrade.py
petr-balogh: fixed, thanks

petr-balogh:
Verified here: https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/1782/

dahorak (approved):

clacroix12 (approved):

PR #4811, opened 2021-09-08
Skip test_verify_rwo_using_replicated_pod if all pods are created on the same node

### LINKED ISSUE #4431: Test case test_verify_rwo_using_replicated_pod failed
Test case
```tests.manage.pv_services.test_verify_rwo_using_dc_pod.TestVerifyRwoUsingReplicatedPod.test_verify_rwo_using_replicated_pod[CephBlockPool]```
failed due to the error given below.

```
>               raise self.timeout_exc_cls(*self.timeout_exc_args)
E               ocs_ci.ocs.exceptions.TimeoutExpiredError: Timed out after 360s running get("pod-test-rbd-75591c00bd8e480ebcc91d33c9a-1-deploy", True, None)

ocs_ci/utility/utils.py:1002: TimeoutExpiredError
```
The test case creates DC with replica value 5 and expects at least one replica pod to be Pending and deploy pod be in Running state. This expectation is set because an RWO PVC is used. So among the 5 replica pods, only the pods which are residing on the same node will reach the Running state. So the deploy pod should remain in Running state. 

But in this case, all the 5 replicas were scheduled on the same node and the deploy pod reached the Completed state.
```
12:40:01 - MainThread - ocs_ci.ocs.ocp - INFO - status of pod-test-rbd-75591c00bd8e480ebcc91d33c9a-1-deploy at column STATUS was Completed, but we were waiting for Running
```
This situation is not handled in the test case.

Failed in RUN ID: 1622715438

jilju:
The test case test_verify_rwo_using_replicated_pod cannot be continued if all the pods are created on the same node.
This will also fixes #4431

### DIFF
--- a/tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
+++ b/tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
@@ -4,7 +4,7 @@ from itertools import cycle
 
 from ocs_ci.framework.testlib import ManageTest, tier2
 from ocs_ci.ocs import constants
-from ocs_ci.ocs.exceptions import ResourceWrongStatusException
+from ocs_ci.ocs.exceptions import ResourceWrongStatusException, TimeoutExpiredError
 from ocs_ci.ocs.resources.ocs import OCS
 from ocs_ci.ocs.resources import pod
 from ocs_ci.ocs.ocp import OCP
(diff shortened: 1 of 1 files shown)


rachael-george:
Instead of skipping the test case, is there a way to make sure the pods come up on different nodes? For eg:
1. Create a dc with replica 1
2. Identify the node where the pod is created and cordon it 
3. Increase the replica count in the dc
4. Check if the pods come up
Would this flow make sense?

ebenahar (approved):

ebenahar:
Although approved, @rachael-george +1 on finding another solution other than skipping the test case if all pods come up on the same node.

jilju:
> Instead of skipping the test case, is there a way to make sure the pods come up on different nodes? For eg:
> 
> 1. Create a dc with replica 1
> 2. Identify the node where the pod is created and cordon it
> 3. Increase the replica count in the dc
> 4. Check if the pods come up
>    Would this flow make sense?
This flow will not satisfy the point 2 given below which is verified by this test.
1. Pods being created from the same dc will not override any conditions which satisfy RWO.
2. Only one node where the pods are running will win the race. This test case verify no two pods which resides on different nodes will have access to the volume when all the nodes are created at the same time.

rachael-george:
After a discussion with Jilju, it appears a test case similar to the scenario (without replicated pods) suggested above already [exists](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/manage/pv_services/test_verify_rwo_using_multiple_pods.py)

rachael-george (approved):

jilju:
> PR validation
> 
> Cluster Name:
> Cluster Configuration:
> PR Test Suite:
> PR Test Path: tests/manage/pv_services/test_verify_rwo_using_dc_pod.py
> Additional Test Params:
> OCP VERSION: 4.8
> OCS VERSION: 4.8
> 
> [Job UNSTABLE](https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-trigger-test-pr/1516//testReport) (some or all tests failed).

Test case completed. Tear down error when checking ceph health
```Ceph cluster health is not OK. Health: HEALTH_WARN clock skew detected on mon.c```

rachael-george (approved):

vavuthu (approved):

PR #4812, opened 2021-09-08
Adding bulk deletion functions and functionality, adding parametrization of bulk size 

ypersky1980:

The content of this PR: 

1) Adding Bulk Delete functionality 
2) Adding parametrization of bulk size 
Please note that It was decided to parametrize bulk size to 2 values: 120 and 720 . 
It is possible to add additional  bulk size test cases, but it will cause the test to run for additional hours. 
Since our goal is to have this test as rather short test ( in terms of run time) - it was chosen to have only 2 bulk sizes as a parameter. 
120 - for backward compatibility with Performance results in the previous versions ( for bulk creation) 
720 - since this is close to the maximal bulk size tested on 3 masters and 2 workers cluster and a multiplication of 120 ( 720 = 120*6). 
We would like to compare the 120 and 720 bulk size results and make sure that 720 bulk creation/time is not more than 6 time 
120 bulk creation /deletion time.

### DIFF
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -786,6 +786,7 @@ def delete_bulk_pvcs(pvc_yaml_dir):
     cmd = f"delete -f {pvc_yaml_dir}/"
     oc.exec_oc_cmd(command=cmd, out_yaml_format=False)
 
+
 def verify_block_pool_exists(pool_name):
     """
     Verify if a Ceph block pool exist
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -778,10 +778,10 @@ def create_bulk_pvcs(
 
 def delete_bulk_pvcs(pvc_yaml_dir):
     """
-       Deletes all the pvcs created from yaml file in a provided dir
-       Args:
-           pvc_yaml_dir (str): Directory in which yaml file resides
-       """
+    Deletes all the pvcs created from yaml file in a provided dir
+    Args:
+        pvc_yaml_dir (str): Directory in which yaml file resides
+    """
     oc = OCP(kind="pod", namespace=defaults.ROOK_CLUSTER_NAMESPACE)
     cmd = f"delete -f {pvc_yaml_dir}/"
     oc.exec_oc_cmd(command=cmd, out_yaml_format=False)
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -659,7 +659,9 @@ def create_multiple_pvcs(
         burst (bool): True for bulk creation, False ( default) for multiple creation
 
     Returns:
-         list: List of PVC objects
+         ocs_objs (list): List of PVC objects
+         tmpdir (str): The full path of the directory in which the yamls for pvc objects creation reside
+
     """
     if not burst:
         if access_mode == "ReadWriteMany" and "rbd" in sc_name:
@@ -1395,48 +1397,6 @@ def get_provision_time(interface, pvc_name, status="start"):
     return datetime.datetime.strptime(stat, DATE_TIME_FORMAT)
 
 
-def get_pvc_bulk_deletion_time(interface, pv_names, status="start"):
-    """
-    Get the deletion time of PVC bulk based on csi provisioner logs
-    Args:
-        interface (str): The interface backed the PVC
-        pv_names (list): Names of the PVs which are backend objects for PVCs deleted
-        status (str): the status that we want to get - start / end time of bulk deletion
-    Returns:
-        datetime object: The time of bulk pvc deletion
-    """
-    logging.info(f"Getting pvc bulk deletion time {status}")
-    # Define the operation that need to retrieve
-    operation = "started"
-    if status.lower() == "end":
-        operation = "succeeded"
-
-    this_year = str(datetime.datetime.now().year)
-    # Get the correct provisioner pod based on the interface
-    pod_name = pod.get_csi_provisioner_pod(interface)
-    # get the logs from the csi-provisioner containers
-    logs = pod.get_pod_logs(pod_name[0], "csi-provisioner")
-    logs += pod.get_pod_logs(pod_name[1], "csi-provisioner")
-
-    logs = logs.split("\n")
-
-    all_stats = []
-    for pv_name in pv_names:
-        logging.info(f"Looking for deletion time of pv {pv_name}")
-        stat = [i for i in logs if re.search(f'delete "{pv_name}".*{operation}', i)]
-        mon_day = " ".join(stat[0].split(" ")[0:2])
-        stat = f"{this_year} {mon_day}"
-        all_stats.append(stat)
-
-    all_stats = sorted(all_stats)
-    if status.lower() == "end":
-        stat = all_stats[-1]  # return the latest time
-    elif status.lower() == "start":
-        stat = all_stats[0]  # return the earliest time
-
-    return datetime.datetime.strptime(stat, DATE_TIME_FORMAT)
-
-
 def get_start_creation_time(interface, pvc_name):
     """
     Get the starting creation time of a PVC based on provisioner logs
@@ -1577,7 +1537,9 @@ def measure_pvc_creation_time_bulk(interface, pvc_name_list, wait_time=60):
     return pvc_dict
 
 
-def measure_pv_deletion_time_bulk(interface, pv_name_list, wait_time=60):
+def measure_pv_deletion_time_bulk(
+    interface, pv_name_list, wait_time=60, return_log_times=False
+):
     """
     Measure PV deletion time of bulk PV, based on logs.
 
@@ -1585,9 +1547,14 @@ def measure_pv_deletion_time_bulk(interface, pv_name_list, wait_time=60):
         interface (str): The interface backed the PV
         pv_name_list (list): List of PV Names for measuring deletion time
         wait_time (int): Seconds to wait before collecting CSI log
+        return_log_times (bool): Determines the return value -- if False, dictionary of pv_names with the deletion time
+                is returned; if True -- the dictionary of pv_names with the tuple of (srart_deletion_time,
+                end_deletion_time) is returned
 
     Returns:
-        pv_dict (dict): Dictionary of pv_name with deletion time.
+        pv_dict (dict): Dictionary where the pv_names are the keys. The value of the dictionary depend on the
+                return_log_times argument value and are either the corresponding deletion times (when return_log_times
+                is False) or a tuple of (start_deletion_time, end_deletion_time) as they appear in the logs
 
     """
     # Get the correct provisioner pod based on the interface
@@ -1633,15 +1600,18 @@ def measure_pv_deletion_time_bulk(interface, pv_name_list, wait_time=60):
         # Extract the deletion start time for the PV
         start = [i for i in logs if re.search(f'delete "{pv_name}": started', i)]
         mon_day = " ".join(start[0].split(" ")[0:2])
-        start = f"{this_year} {mon_day}"
-        start_time = datetime.datetime.strptime(start, DATE_TIME_FORMAT)
+        start_tm = f"{this_year} {mon_day}"
+        start_time = datetime.datetime.strptime(start_tm, DATE_TIME_FORMAT)
         # Extract the deletion end time for the PV
         end = [i for i in logs if re.search(f'delete "{pv_name}": succeeded', i)]
         mon_day = " ".join(end[0].split(" ")[0:2])
         end_tm = f"{this_year} {mon_day}"
         end_time = datetime.datetime.strptime(end_tm, DATE_TIME_FORMAT)
         total = end_time - start_time
-        pv_dict[pv_name] = total.total_seconds()
+        if not return_log_times:
+            pv_dict[pv_name] = total.total_seconds()
+        else:
+            pv_dict[pv_name] = (start_tm, end_tm)
 
     return pv_dict
 
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -728,7 +728,7 @@ def delete_bulk_pvcs(pvc_yaml_dir, pv_names_list):
     cmd = f"delete -f {pvc_yaml_dir}/"
     oc.exec_oc_cmd(command=cmd, out_yaml_format=False)
 
-    time.sleep(500)
+    time.sleep(len(pv_names_list) / 2)
 
     for pv_name in pv_names_list:
         validate_pv_delete(pv_name)
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -716,6 +716,76 @@ def create_multiple_pvcs(
     return ocs_objs
 
 
+def create_bulk_pvcs(
+    sc_name,
+    namespace,
+    number_of_pvc=1,
+    size=None,
+    access_mode=constants.ACCESS_MODE_RWO,
+):
+    """
+    Create one or more PVC as a bulk or one by one
+    Args:
+        sc_name (str): The name of the storage class to provision the PVCs from
+        namespace (str): The namespace for the PVCs creation
+        number_of_pvc (int): Number of PVCs to be created
+        size (str): The size of the PVCs to create
+        do_reload (bool): True for wait for reloading PVC after its creation,
+            False otherwise
+        access_mode (str): The kind of access mode for PVC
+    Returns:
+         list: List of PVC objects and tmp dir where the yamls are created
+    """
+
+    pvc_data = templating.load_yaml(constants.CSI_PVC_YAML)
+    pvc_data["metadata"]["namespace"] = namespace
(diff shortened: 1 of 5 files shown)


ypersky1980:
Please note that the validation run has passed and the test run log is available here: 10.70.39.233:/home/ypersky/ypersky_pr_validation/pr_4812/ypersky-lso2_test_pvc_multiple_creation_deletion_performance_1

ocs_ci/helpers/helpers.py
```diff
@@ -716,6 +716,76 @@ def create_multiple_pvcs(
     return ocs_objs
 
 
+def create_bulk_pvcs(
```
Avilir, line 719: why did you duplicate the create_multiple_pvc() function (on line 639) and not modifying it ?
all you need is to change the return value.

ocs_ci/helpers/helpers.py
```diff
@@ -716,6 +716,76 @@ def create_multiple_pvcs(
...
+def delete_bulk_pvcs(pvc_yaml_dir):
+    """
+    Deletes all the pvcs created from yaml file in a provided dir
+    Args:
+        pvc_yaml_dir (str): Directory in which yaml file resides
+    """
+    oc = OCP(kind="pod", namespace=defaults.ROOK_CLUSTER_NAMESPACE)
+    cmd = f"delete -f {pvc_yaml_dir}/"
+    oc.exec_oc_cmd(command=cmd, out_yaml_format=False)
+
```
Avilir, line 787: you need to wait and verify that all PVC's was deleted.

ypersky1980: will add time.sleep (100). Or more ?

ypersky1980: I've added sleep (500) and calidation of pv deletion. Will execute it to check whether this timeout is enough.

ocs_ci/helpers/helpers.py
ypersky1980: Agree. Added to the existing function returning directory path, and change calls in the relevant places in the project.

ocs_ci/helpers/helpers.py
keesturam: @ypersky1980 we should also check if the total rbd images in ceph system match with the PV count at OCS. This will ensure there are no leftover rbd images from the backend after deletion.

ocs_ci/helpers/helpers.py
```diff
@@ -1377,6 +1447,47 @@ def get_provision_time(interface, pvc_name, status="start"):
...
+    logs = pod.get_pod_logs(pod_name[0], "csi-provisioner")
+    logs += pod.get_pod_logs(pod_name[1], "csi-provisioner")
+
+    logs = logs.split("\n")
+
+    all_stats = []
+    for pv_name in pv_names:
+        stat = [i for i in logs if re.search(f'delete "{pv_name}".*{operation}', i)]
+        mon_day = " ".join(stat[0].split(" ")[0:2])
+        stat = f"{this_year} {mon_day}"
```
keesturam, line 1479: We need to be more efficient here. Looking through PVC deletion start and end time for 720 volumes means, parsing through the same log 1440 times. Maybe there is a better way?

tests/e2e/performance/test_pvc_multiple_creation_deletion_performance.py
```diff
@@ -17,58 +17,72 @@
...
             teardown_factory: A fixture used when we want a new resource that was created during the tests
                                to be removed in the teardown phase.
+            bulk_size: Size of the bulk to be tested
         Returns:
 
         """
-        number_of_pvcs = 120
-        multi_creation_time_limit = 60
-        log.info("Start creating new 120 PVCs")
+        bulk_creation_time_limit = bulk_size / 2
```
keesturam, line 75: Do we know what are the numbers we see today for 120 & 720?

ypersky1980: Yes we do: 

RBD:  

23:17:52 - MainThread - root - INFO - 120 Bulk PVCs creation time is 40.038341 seconds.
00:38:40 - MainThread - root - INFO - 720 Bulk PVCs creation time is 280.852776 seconds.

23:24:44 - MainThread - root - INFO - 120 Bulk PVCs deletion time is 46.113197 seconds.
00:57:15 - MainThread - root - INFO - 720 Bulk PVCs deletion time is 286.213385 seconds.

CephFS: 

01:10:04 - MainThread - root - INFO - 120 Bulk PVCs creation time is 40.424715 seconds.
02:40:01 - MainThread - root - INFO - 720 Bulk PVCs creation time is 282.832763 seconds.

01:18:54 - MainThread - root - INFO - 120 Bulk PVCs deletion time is 46.005821 seconds.
03:00:57 - MainThread - root - INFO - 720 Bulk PVCs deletion time is 286.005817 seconds.

ocs_ci/helpers/helpers.py
ypersky1980: @keesturam - I've tried to think about it and I don't see how we find the relevant records in the log more efficiently.

ocs_ci/helpers/helpers.py
```diff
@@ -716,6 +716,76 @@ def create_multiple_pvcs(
...
+    ocs_objs = []
+    for _ in range(number_of_pvc):
+        name = create_unique_resource_name("test", "pvc")
+        logger.info(f"Adding PVC with name {name}")
+        pvc_data["metadata"]["name"] = name
+        templating.dump_data_to_temp_yaml(pvc_data, f"{tmpdir}/{name}.yaml")
+        ocs_objs.append(pvc.PVC(**pvc_data))
+
+    logger.info("Creating all PVCs as bulk")
+    oc = OCP(kind="pod", namespace=defaults.ROOK_CLUSTER_NAMESPACE)
```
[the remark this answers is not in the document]
ypersky1980: This function is in a way duplicate of create_multiple_pvcs and following Avi's review, I've removed it. In the create_multiple_pvcs function, the correct namespace is used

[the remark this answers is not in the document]
ypersky1980: @keesturam  - I think that this validation is more accutate than comparind all the images to all the PVs, because there might be leftovers from the previous test runs. 
And check if the total rbd images in ceph system matchs with the PV count at OCS should be part of the improved flow of all the tests ( to measure images and pv count before the test and if not equal  - exit). 
But not an intergral part of this specific test.

ypersky1980:
Validation run log is available here: ypersky@10.70.39.233:/home/ypersky/ypersky_pr_validation/pr_4812/ypersky-lso2_test_pvc_bulk_creation_deletion_performance_2

tests/e2e/performance/test_pvc_multiple_creation_deletion_performance.py
keesturam: Great, thank you, Yuli for sharing the results. 
maybe in a subsequent PR, we should fine-tune the time limit based on these test results. i.e., 50 seconds for 120 PVC delete ops and 300 seconds for 720 PVC delete ops.

ocs_ci/helpers/helpers.py
ramkiperiy: @ypersky1980 have you tried kube_job for creation ?, if so deletion is just one line, i.e. `kube_job.delete(namespace)` which kind of does the same.
Eg: https://github.com/red-hat-storage/ocs-ci/blob/master/tests/e2e/scale/test_pvc_creation_deletion_scale.py#L138

ocs_ci/helpers/helpers.py
ramkiperiy: Don't we have similar function in performance or helpers ?, it's good to enhance these functions.
measure_pv_deletion_time_bulk(interface, pv_name_list, wait_time=60) -- https://github.com/red-hat-storage/ocs-ci/blob/abcced1c32c2341449a00612e27188fca56f86ca/ocs_ci/helpers/helpers.py#L1520

tests/e2e/performance/test_pvc_bulk_creation_deletion_performance.py
```diff
@@ -79,28 +94,62 @@ def test_multiple_pvc_creation_measurement_performance(self, teardown_factory):
...
+                storageclass
+        """
+        self.interface = interface_iterate
+        self.sc_obj = storageclass_factory(self.interface)
+
+    @pytest.mark.usefixtures(base_setup_creation_after_deletion.__name__)
     @polarion_id("OCS-1270")
     @bugzilla("1741612")
-    def test_multiple_pvc_creation_after_deletion_performance(self, teardown_factory):
+    def test_bulk_pvc_creation_after_deletion_performance(self, teardown_factory):
```
ramkiperiy, line 149: It's better to move this function to performance helper file than using in testcase

ocs_ci/helpers/helpers.py
ypersky1980: @ramkiperiy  no, I did not try this, I thought that it would be a good idea to implement bulk deletion in a similar way the bulk creation is implemented.  The current implementation is also rather short and the separate function can be useful as well for the future. But thank you for pointing this out, I'll keep it in mind for future.

tests/e2e/performance/test_pvc_multiple_creation_deletion_performance.py
ypersky1980: @keesturam I think that we should address this question to Eran Tamir and implement the limit that he decides. Do you agree?

tests/e2e/performance/test_pvc_bulk_creation_deletion_performance.py
ypersky1980: @ramkiperiy I am afraid I do not understand. 
test_bulk_pvc_creation_after_deletion_performance  is a test case and not a function, we need it as a test.

ocs_ci/helpers/helpers.py
ypersky1980: @ramkiperiy  Yes, we do have this function, but it measured different time , not what we need. 
 def measure_pv_deletion_time_bulk provides a dictionary with deletion time for EACH one of the PVCs in a bulk and the bulk_deletion in this PR is meant to measure deletion time starting from the start of the deletion of the first PVC and till the end of the deletion of the last deleted PVC. 

I started with trying to customise the existing code ( measure_pv_deletion_time_bulk) , but it became cumbersome. 
Therefore, even though I realize that part of the code is duplicated, I chose to implement it in a ceparate function.

ocs_ci/helpers/helpers.py
```diff
@@ -713,7 +714,24 @@ def create_multiple_pvcs(
     )
     time.sleep(number_of_pvc)
 
-    return ocs_objs
+    return ocs_objs, tmpdir
```
Avilir, line 717: pleas update the return argument  in the docstring to reflect this change

tests/e2e/performance/test_pvc_bulk_creation_deletion_performance.py
```diff
@@ -111,14 +160,14 @@ def test_multiple_pvc_creation_after_deletion_performance(self, teardown_factory
...
-        log.info("Start creating new 120 PVCs")
+        log.info(f"Start creating new {initial_number_of_pvcs} PVCs in a bulk")
         pvc_objs = helpers.create_multiple_pvcs(
             sc_name=self.sc_obj.name,
             namespace=defaults.ROOK_CLUSTER_NAMESPACE,
             number_of_pvc=initial_number_of_pvcs,
             size=self.pvc_size,
             burst=True,
-        )
+        )[0]
```
Avilir, line 170: the function return a tuple, so you need to change to something like :
(pvc_objs,  placholder) = helpers.create_multiple_pvcs(...)

tests/e2e/performance/test_pvc_bulk_creation_deletion_performance.py
```diff
@@ -139,7 +188,7 @@ def test_multiple_pvc_creation_after_deletion_performance(self, teardown_factory
             number_of_pvc=number_of_pvcs,
             size=self.pvc_size,
             burst=True,
-        )
+        )[0]
```
Avilir, line 191: same as above

tests/e2e/performance/test_pvc_creation_deletion_performance.py
```diff
@@ -421,7 +421,7 @@ def test_multiple_pvc_deletion_measurement_performance(self, teardown_factory):
             number_of_pvc=number_of_pvcs,
             size=pvc_size,
             burst=True,
-        )
+        )[0]
```
Avilir, line 424: same as above

tests/fixtures.py
```diff
@@ -240,7 +240,7 @@ def finalizer():
         number_of_pvc=class_instance.num_of_pvcs,
         size=class_instance.pvc_size,
         namespace=class_instance.namespace,
-    )
+    )[0]
```
Avilir, line 243: same as above

tests/e2e/performance/test_pvc_bulk_creation_deletion_performance.py
```diff
@@ -17,58 +17,73 @@
...
     @pytest.mark.usefixtures(base_setup.__name__)
     @polarion_id("OCS-1620")
-    def test_multiple_pvc_creation_measurement_performance(self, teardown_factory):
+    def test_bulk_pvc_creation_deletion_measurement_performance(
+        self, teardown_factory, bulk_size
+    ):
+
         """
-        Measuring PVC creation time of 120 PVCs in 180 seconds
+        Measuring PVC creation and deletion time of bulk_size (120/720) PVCs
```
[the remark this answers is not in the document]
ypersky1980: Done. Thanks

ocs_ci/helpers/helpers.py
ypersky1980: Done. Thanks

tests/e2e/performance/test_pvc_bulk_creation_deletion_performance.py
ypersky1980: Done, changed to pvc_objs, _ = helpers.create_multiple_pvcs(...)
Thanks

tests/e2e/performance/test_pvc_creation_deletion_performance.py
ypersky1980: Done

tests/fixtures.py
ypersky1980: Done

tests/e2e/performance/test_pvc_bulk_creation_deletion_performance.py
ypersky1980: Done

ypersky1980:
@Avilir , regarding your comment : 

> @ypersky1980, as @ramkiperiy mention, you can use the measure_pv_deletion_time_bulk function, and then just sort it to get the start and end of the bulk.

For sure I understand the need of not duplicating the code. However currently measure_pv_deletion_time_bulk function returns the dictionary where the key is pv_name and the value is the deletion time. 

To use this function for our purpose (measure the deletion time of the whole bulk) we need to modify it and make it return a dictionary with pv_name as key and the value will be  a 3-tuple of (deletion_time, start_deletion_time, end_deletion_time). 

The existing calls to this function will use the deletion_time as before, while our new test will calculate the bulk deletion time as the difference between the latest  end_deletion_time and the earliest start_deletion_time. 
Is that the kind of solution that you think will be good?

@ramkiperiy  what do you think? Should I implement it this way ?

tests/e2e/performance/test_pvc_bulk_creation_deletion_performance.py
ramkiperiy: ah ok!! sorry my bad.

ramkiperiy:
> @Avilir , regarding your comment :
> 
> > @ypersky1980, as @ramkiperiy mention, you can use the measure_pv_deletion_time_bulk function, and then just sort it to get the start and end of the bulk.
> 
> For sure I understand the need of not duplicating the code. However currently measure_pv_deletion_time_bulk function returns the dictionary where the key is pv_name and the value is the deletion time.
> 
> To use this function for our purpose (measure the deletion time of the whole bulk) we need to modify it and make it return a dictionary with pv_name as key and the value will be a 3-tuple of (deletion_time, start_deletion_time, end_deletion_time).
> 
> The existing calls to this function will use the deletion_time as before, while our new test will calculate the bulk deletion time as the difference between the latest end_deletion_time and the earliest start_deletion_time.
> Is that the kind of solution that you think will be good?
> 
> @ramkiperiy what do you think? Should I implement it this way ?

I think better to add a new param (kind of flag), if set then function can return the 3 tuple value, else it will default return existing values.
The reason we need both return values is existing testcase can still use this function with just a key and value.

ypersky1980:
PLease note that PR validation job failed in deployment stage, the full passed validation run log is available here: 

Validation run log is available here: ypersky@10.70.39.233:/home/ypersky/ypersky_pr_validation/pr_4812/ypersky-lso2_test_pvc_bulk_creation_deletion_performance_validation1

Avilir (approved):

ramkiperiy (approved):

PR #4817, opened 2021-09-09
VMware ipi support for scale tests

ramkiperiy:
Enhanced scale test cases to use vmware ipi platform using custom machineset
With this PR we can execute all the scale tests in vmware ipi platform

### DIFF
--- a/ocs_ci/ocs/machine.py
+++ b/ocs_ci/ocs/machine.py
@@ -560,40 +560,40 @@ def create_custom_machineset(
         )
         for machine in machinesets_obj.get()["items"]:
             # Get inputs from existing machineset config.
-            cls_id = machine.get["spec"]["selector"]["matchLabels"][
+            cls_id = machine.get("spec")["selector"]["matchLabels"][
                 "machine.openshift.io/cluster-api-cluster"
             ]
-            disk_size = machine.get["spec"]["template"]["spec"]["providerSpec"][
+            disk_size = machine.get("spec")["template"]["spec"]["providerSpec"][
                 "value"
             ]["diskGiB"]
-            memory = machine.get["spec"]["template"]["spec"]["providerSpec"]["value"][
+            memory = machine.get("spec")["template"]["spec"]["providerSpec"]["value"][
                 "memoryMiB"
             ]
-            network_name = machine.get["spec"]["template"]["spec"]["providerSpec"][
+            network_name = machine.get("spec")["template"]["spec"]["providerSpec"][
                 "value"
             ]["network"]["devices"][0]["networkName"]
-            num_cpu = machine.get["spec"]["template"]["spec"]["providerSpec"]["value"][
+            num_cpu = machine.get("spec")["template"]["spec"]["providerSpec"]["value"][
                 "numCPUs"
             ]
-            num_core = machine.get["spec"]["template"]["spec"]["providerSpec"]["value"][
+            num_core = machine.get("spec")["template"]["spec"]["providerSpec"]["value"][
                 "numCoresPerSocket"
             ]
-            vm_template = machine.get["spec"]["template"]["spec"]["providerSpec"][
+            vm_template = machine.get("spec")["template"]["spec"]["providerSpec"][
                 "value"
             ]["template"]
-            datacenter = machine.get["spec"]["template"]["spec"]["providerSpec"][
+            datacenter = machine.get("spec")["template"]["spec"]["providerSpec"][
                 "value"
             ]["workspace"]["datacenter"]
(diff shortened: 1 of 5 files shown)


Avilir (dismissed):
LGTM

ramkiperiy:
> 
> Cluster Name: jopinto-dcps
> Cluster Configuration:
> PR Test Suite: scale
> PR Test Path: tests/e2e/scale/test_scale_3_OCS_worker_nodes_and_1500_PVCs.py
> Additional Test Params:
> OCP VERSION: 4.8
> OCS VERSION: 4.8
> 
> [Job UNSTABLE](https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-trigger-test-pr/1464//testReport) (some or all tests failed).

Jenkins job unstable due to reboot failure in vmware as seen in image, the code worked for creating more vm's in vmware and we did used those vms for running app pods.
The reboot failure is due to environment issue/or the support to vmware node restarts.
image
image

ocs_ci/ocs/machine.py
```diff
@@ -552,6 +552,199 @@ def create_custom_machineset(
...
+            kind=constants.MACHINESETS,
+            namespace=constants.OPENSHIFT_MACHINE_API_NAMESPACE,
+        )
+        for machine in machinesets_obj.get()["items"]:
+            # Get inputs from existing machineset config.
+            cls_id = (
+                machine.get("spec")
+                .get("selector")
+                .get("matchLabels")
+                .get("machine.openshift.io/cluster-api-cluster")
```
petr-balogh, line 567: why we are using this get approach?
If those data should be there and we need to use them, I would go with direct access from dict like:
machine.get["spec"]["selector"]["matchLabels"]["machine.openshift.io/cluster-api-cluster"]
I think it is more effective and will raise the direct keyError if we will get some data which do not have this key which will be more clear here than: ```
AttributeError: 'NoneType' object has no attribute 'get'```
Which doesn't tell you anything.

Instead of:
```
In [3]: data["spec"]["selector"]
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-3-34fb83c900ad> in <module>
----> 1 data["spec"]["selector"]

KeyError: 'selector'
```
Is more precise on which data we are exactly missing.

ramkiperiy: updated the code. Thanks for the comment @petr-balogh

ocs_ci/ocs/scale_lib.py
```diff
@@ -878,14 +888,67 @@ def check_and_add_enough_worker(worker_count):
...
+        ):
+            # Create machineset for app worker nodes on vmware
+            ms_name = list()
+            labels = [("node-role.kubernetes.io/app", "app-scale")]
+            for obj in machine.get_machineset_objs():
+                if "app" in obj.name:
+                    ms_name.append(obj.name)
+            if not ms_name:
+                if len(machine.get_machineset_objs()) == 3:
+                    for zone in ["3", "4", "5"]:
```
petr-balogh, line 904: What are the default zones of already existing nodes?  Do we have new zones here? Or existing zone names are 3, 4, 5? Should we somehow dynamically get current zones and add nodes to the existing one dynamically instead of hardcoded numbers here?

Or try to read it from config? As for arbiter deployment we have zones defined in config:
https://github.com/red-hat-storage/ocs-ci/blob/master/conf/deployment/vsphere/upi_1az_rhcos_vsan_lso_vmdk_3m_6w_arbiter.yaml#L20-L22

But this is UPI, for IPI I don't see we define zones in config (maybe we should).

ramkiperiy: Scale tests are for IPI and yes it makes sense to get the zone name dynamically than hardcoding it, I will make this change.

ramkiperiy: fixed

ramkiperiy:
@petr-balogh updated code to remove the zone hardcoded values, for now there is no descriptive yaml to add different zone in vsphere, once after this clarification will add this support.

Avilir (approved):
LGTM

tests/e2e/scale/test_scale_3_OCS_worker_nodes_and_1500_PVCs.py
```diff
@@ -11,6 +11,7 @@
 from ocs_ci.framework.pytest_customization.marks import (
     skipif_external_mode,
     ipi_deployment_required,
+    skipif_vsphere_ipi,
```
ramkiperiy, line 14: Skipped vsphere IPI due to no support for reboot, there is JIRA card.

vasukulkarni (approved):

Issue #4807, opened 2021-09-07
test_fio_workload_simple-CephBlockPool-sequential stuck in loop for 7 days on status.phase with timeout of 600


petr-balogh:
https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/1719/

running for 7 days

and it's still trying to run this command over and over

07:14:15 - MainThread - ocs_ci.helpers.performance_lib - INFO - Going to run oc --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig -n benchmark-operator get pod --no-headers -o custom-columns=:metadata.name,:status.phase with timeout of 600

Problematic case is:
15:14:12 - MainThread - test_fio_benchmark - INFO - Logs file path name is : /home/jenkins/current-cluster-dir/logs/ocs-ci-logs-1630335194/tests/e2e/performance/test_fio_benchmark.py/TestFIOBenchmark/test_fio_workload_simple-CephBlockPool-sequential

Logs: http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/j-011vu1cs33-pn/j-011vu1cs33-pn_20210830T125630

petr-balogh:
The same happened on AWS cluster here:
https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/1725/consoleFull but here it also deleted the cluster later on and it's stuck there.

We should disable the test till it's fixed. And make sure we will not get to such situation again.

Avilir:
I found 2 reasons for this issue :

1. the function which identify benchmark restart  is not working as expected when more then one pod is used in the benchmark - PR is WIP
2. it is look like out of resources make the FIO benchmark to failed.

PR #4825, opened 2021-09-12
Fix function which wait for benchmark to finish

Avilir:
this PR is part of the fix for #4807 

### DIFF
--- a/ocs_ci/ocs/perftests.py
+++ b/ocs_ci/ocs/perftests.py
@@ -274,15 +274,27 @@ class PASTest(BaseTest):
         total_time = timeout
         while not Finished and total_time > 0:
             results = run_oc_command(
-                "get pod --no-headers -o custom-columns=:metadata.name,:status.phase",
+                f"get pod --no-headers -o custom-columns=:metadata.name,:status.phase",
                 namespace=benchmark_operator.BMO_NAME,
             )
-            fname = ""
+            (fname, status) = ["", ""]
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/perftests.py
```diff
@@ -274,26 +274,40 @@ def wait_for_wl_to_finish(self, timeout=18000, sleep=300):
...
+            (fname, status) = ["", ""]
             for name in results:
-                if re.search(self.client_pod_name, name):
-                    (fname, status) = name.split()
-                    continue
+                # looking for the pod which run the benchmark (not the IO)
+                # this pod contain the `client` in his name, and there is only one
+                # pod like this, other pods have the `server` in the name.
+                (fname, status) = name.split()
+                if re.search("client", fname):
```
keesturam, line 286: @Avilir should we also add `fio-` or `smallfile-` before the client so we are precisely looking for the right one? I understand we don't run fio and smallfile workload together today, but maybe we will in the future? as some sort of stress test. (This can be done in a subsequent PR as well).

keesturam (approved):

ocs_ci/ocs/perftests.py
Avilir: agree, i will do it in separate PR.

keesturam (approved):

ramkiperiy (approved):
LGTM

PR #4826, opened 2021-09-12
Verify 'noobaa-service-monitor' does not exist after OCS upgrade

OdedViner:
BZ link: https://bugzilla.redhat.com/show_bug.cgi?id=1973179  
Test Procedure:
  1.Upgrade OCS version [4.6->4.7]
  2.Check servicemonitors
  3.Verify 'noobaa-service-monitor' does not exist


liranmauda (dismissed):
LGTM.
One thing, the test will be correct when upgrading from 4.6 to 4.7.4 as it is the first z that contains that fix.

OdedViner:
> LGTM.
> One thing, the test will be correct when upgrading from 4.6 to 4.7.4 as it is the first z that contains that fix.

@liranmauda I added a decorator `@skipif_ocs_version("<4.7")`.. 
@petr-balogh @AviadP We are working with the latest Z after OCS upgrading..correct?

tests/ecosystem/upgrade/test_resources.py
```diff
@@ -123,3 +124,26 @@ def test_pod_log_after_upgrade():
...
+    """
+    test_noobaa_service_mon_after_ocs_upgrade
+
+    Test Procedure:
+    1.Upgrade OCS version [4.6->4.7]
+    2.Check servicemonitors
+    3.Verify 'noobaa-service-monitor' does not exist
+
+    """
+    ocp_obj = ocp.OCP(kind="servicemonitors", namespace=defaults.ROOK_CLUSTER_NAMESPACE)
```
Shrivaibavi, line 142: Can we add this in constants.py

Shrivaibavi (dismissed):
Just a small comment, Otherwise it looks good to me

petr-balogh:
> > LGTM.
> > One thing, the test will be correct when upgrading from 4.6 to 4.7.4 as it is the first z that contains that fix.
> 
> @liranmauda I added a decorator `@skipif_ocs_version("<4.7")`..
> @petr-balogh @AviadP We are working with the latest Z after OCS upgrading..correct?

Yep, we are upgrading to 

> > LGTM.
> > One thing, the test will be correct when upgrading from 4.6 to 4.7.4 as it is the first z that contains that fix.
> 
> @liranmauda I added a decorator `@skipif_ocs_version("<4.7")`..
> @petr-balogh @AviadP We are working with the latest Z after OCS upgrading..correct?

Yes @OdedViner we upgrade usually to latest Z-stream, but someone can still invoke upgrade between particular internal builds - but I would not care about this.

Only one think is that @skipif_ocs_version("<4.7") is evaluated during the collection I think. And when we are starting Y-stream upgrade like 4.6 -> 4.7 , the OCS_VERSION is 4.6 in this case so this test will be most likely skipped here and will be covered only in Z-Stream upgrade from 4.7.3 to 4.7.4 this time and in next z-streams like 4.7.4-> 4.7.5 and y-stream upgrades like 4.7 -> 4.8 and so on.  So if we still would like to have coverage in 4.6 -> 4.7 upgrade we need to use different approach here.

We are storing upgrade ocs version here:
https://github.com/red-hat-storage/ocs-ci/blob/8b674782866ac846d519b39be4d291531aae6aae/ocs_ci/framework/pytest_customization/ocscilib.py#L471

So you might utilize this value from ocsci_config.UPGRADE["upgrade_ocs_version"] and  ocsci_config.ENV_DATA["ocs_version "] .

tests/ecosystem/upgrade/test_resources.py
```diff
@@ -123,3 +124,26 @@ def test_pod_log_after_upgrade():
     logging.info(
         f"The log '{expected_log_after_upgrade}' appears in all relevant pods."
     )
+
+
+@post_upgrade
+@bugzilla("1973179")
+@skipif_ocs_version("<4.7")
+def test_noobaa_service_mon_after_ocs_upgrade():
```
parikshithb, line 132: Add test case.

parikshithb (commented):
Looks great

tests/ecosystem/upgrade/test_resources.py
```diff
@@ -123,3 +125,30 @@ def test_pod_log_after_upgrade():
...
+    test_noobaa_service_mon_after_ocs_upgrade
+
+    Test Procedure:
+    1.Upgrade OCS version [4.6->4.7]
+    2.Check servicemonitors
+    3.Verify 'noobaa-service-monitor' does not exist
+
+    """
+    upgrade_ocs_version = config.UPGRADE.get("upgrade_ocs_version")
+    if Version.coerce(upgrade_ocs_version) >= Version.coerce("4.7.4"):
```
ebenahar, line 144: From what I remember, we don't use the Z number in ocs-ci. So `.4` won't work or will be ignored I think

tests/ecosystem/upgrade/test_resources.py
OdedViner: done

tests/ecosystem/upgrade/test_resources.py
OdedViner: done

Shrivaibavi (approved):
LGTM

PrasadDesala (approved):

petr-balogh (dismissed):

dahorak (approved):

PR #4840, opened 2021-09-15
Added timeout sampler to wait volume method

Gangadhar-ibm-lab:
Added timeout sampler to wait volume method

### DIFF
--- a/ocs_ci/utility/ibmcloud.py
+++ b/ocs_ci/utility/ibmcloud.py
@@ -17,6 +17,7 @@ from ocs_ci.ocs.exceptions import (
     UnsupportedPlatformVersionError,
     UnexpectedBehaviour,
     NodeHasNoAttachedVolume,
+    TimeoutExpiredError,
 )
 from ocs_ci.utility.utils import get_ocp_version, run_cmd, TimeoutSampler
 from ocs_ci.ocs.node import get_nodes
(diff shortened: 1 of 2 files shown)


Avilir (dismissed):
LGTM

ocs_ci/utility/ibmcloud.py
```diff
@@ -413,16 +414,20 @@ def wait_for_volume_attach(self, volume):
                 instance, False otherwise
 
         """
-        cmd = f"ibmcloud is volume {volume} --output json"
-        out = run_ibmcloud_cmd(cmd)
-        out = json.loads(out)
 
-        if not out["volume_attachments"]:
-            logger.info("volume is not attached to node")
+        def is_volume_attached(volume):
```
petr-balogh, line 418: I prefer this nested function to be moved outside of this method this to be regular method as it can be used not just in this function but maybe others as well in future.

Also missing docstring and description of volume argument plus documentation of return value.

Gangadhar-ibm-lab: Moved the method outside as regular method and added docstring

ocs_ci/utility/ibmcloud.py
```diff
@@ -401,6 +402,24 @@ def get_data_volumes(self):
...
+    def is_volume_attached(self, volume):
+        """
+        Check if volume is attached to node or not.
+
+        Args:
+            volume (str): The volume to check for to attached
+
+        Returns:
+            bool: 'True' if volume is attached
+            else 'False'
```
petr-balogh, line 414:
```suggestion
            bool: 'True' if volume is attached otherwise 'False'
```

Make it one liner. Her you anyway are missing one indention to the right in the line where you start with `else`

Gangadhar-ibm-lab: Updated the dock string to one line

petr-balogh (approved):

sidhant-agrawal:
@Gangadhar-ibm-lab I see timeout value is being changed in test_resource_deletion_during_pod_pvc_deletion.py
Can you add a description describing why this is necessary?

Gangadhar-ibm-lab:
> @Gangadhar-ibm-lab I see timeout value is being changed in test_resource_deletion_during_pod_pvc_deletion.py
> Can you add a description describing why this is necessary?

We have increased the time out from default(20 seconds) to 50 seconds for plugin pod deletion

sidhant-agrawal (approved):

PR #4843, opened 2021-09-16
Add skip if marker for ROSA platform

suchita-g:
Add skip if marker for ROSA platform
Add lib test to verify the added code

### DIFF
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -21,6 +21,7 @@ from ocs_ci.ocs.constants import (
     ON_PREM_PLATFORMS,
     IBM_POWER_PLATFORM,
     IBMCLOUD_PLATFORM,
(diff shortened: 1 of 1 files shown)


fbalak (approved):

tests/libtest/test_rosa.py
```diff
@@ -0,0 +1,13 @@
+import logging
```
petr-balogh, line 1: I don't think we really need to have libtest for simple skipif, anyway you do not testing if it was really skipped or not so this test is not telling anything in the report.

suchita-g: I added it to verify it via a valid job. else I can only able to verify this locally. 
I agree this is not any function so I am removing it

suchita-g:
Verified Locally.

petr-balogh (approved):

fbalak (approved):

PR #4853, opened 2021-09-19
Add skip for nodes operation dependent test till implementation

suchita-g:

### DIFF
--- a/ocs_ci/ocs/must_gather/const_must_gather.py
+++ b/ocs_ci/ocs/must_gather/const_must_gather.py
@@ -739,7 +739,10 @@ GATHER_COMMANDS_OTHERS_EXTERNAL_EXCLUDE = [
     "ocs-storagecluster-rbdplugin-snapclass.yaml",
 ]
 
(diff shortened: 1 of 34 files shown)


ocs_ci/ocs/must_gather/const_must_gather.py
```diff
@@ -739,7 +739,10 @@
     "ocs-storagecluster-rbdplugin-snapclass.yaml",
 ]
 
-if config.ENV_DATA["platform"].lower() == constants.OPENSHIFT_DEDICATED_PLATFORM:
+if (
```
ebondare, line 742: I think it would be better to use constants.MANAGED_SERVICE_PLATFORMS here and is similar situations below

tests/e2e/flowtest/mcg/test_mcg_namespace_disruptions_crd.py
```diff
@@ -8,6 +8,7 @@
     skipif_aws_creds_are_missing,
     flowtests,
     skipif_openshift_dedicated,
```
ebondare, line 10: I think it would be better to have something like skipif_managed_service for skipping both rosa and osd, as they go together in so many places.

Issue #4864, opened 2021-09-21
Vmware RHEL deployment fails with error object does not have an attribute named "image_8".


Akarsha-rai:
Vmware UPI RHEL deployment fails with the error: `object does not have an attribute named "image_8".`
```
E           ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: /home/jenkins/bin/terraform apply   '-var-file=/home/jenkins/current-cluster-dir/openshift-cluster-dir/terraform_data/scaleup_terraform_data/scale_up_terraform.tfvars' -auto-approve  '/home/jenkins/workspace/qe-deploy-ocs-cluster/ocs-ci/external/v4-scaleup/ocp4-rhel-scaleup/aos-4_9/vsphere'.
E           Error is 
E           Error: Unsupported attribute
E           
E             on ../../../../workspace/qe-deploy-ocs-cluster/ocs-ci/external/v4-scaleup/ocp4-rhel-scaleup/aos-4_9/vsphere/variables.tf line 12, in locals:
E             12:   rhel_template           = var.RHEL_MAJOR_OS != "7" ? local.vsphere_rhel_properties.image_8 : local.vsphere_rhel_properties.image_7
E               |----------------
E               | local.vsphere_rhel_properties is object with 8 attributes
E           
E           This object does not have an attribute named "image_8".
E           

../../../../workspace/qe-deploy-ocs-cluster/ocs-ci/ocs_ci/utility/utils.py:510: CommandFailed
```
https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/6047/consoleFull

PR #4866, opened 2021-09-21
Adding RHEL_MAJOR_OS to scale_up_terraform.tfvars file

vavuthu:
Fixes: #4864 

### DIFF
--- a/ocs_ci/deployment/helpers/vsphere_helpers.py
+++ b/ocs_ci/deployment/helpers/vsphere_helpers.py
@@ -8,6 +8,7 @@ import os
 
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants
+from ocs_ci.utility import version
 from ocs_ci.utility.templating import (
     Templating,
     load_yaml,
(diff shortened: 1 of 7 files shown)


ocs_ci/deployment/vmware.py
```diff
@@ -188,6 +188,10 @@ def add_nodes(self):
             helpers.generate_config_yaml()
             rhel_module = "RHEL_WORKER_LIST"
 
+        if Version.coerce(self.ocp_version) >= Version.coerce("4.9"):
+            with open(self.scale_up_terraform_var, "a") as fd:
```
petr-balogh, line 192: Is there a chance that this variable will be already defined in that file? If so we might rather think of replacing that line, and if not defined we will add it. But if there is no way it will be part of that file in the future I am OK with this approach.

Also thinking about to have this configurable instead of hardcoded version 7 here.  At some point we will support RHEL 8 nodes and this will bee good to have defined in our configuration.

https://github.com/red-hat-storage/ocs-ci/blob/master/conf/deployment/vsphere/upi_1az_rhel_vsan_3m_3w.yaml#L15

So somewhere here we can define:
```
rhel_workers_major_os_version: 7
```

So here we can take this variable and if it will not be specified we will use 7 as default value.
If at some OCP version we will change RHEL version, then we can maybe change the default value in OCP config file.

So you don't need to care about specifying it in particular configuration files, just take default value 7 here but to have option to overwrite it at some point will be good.

prsurve:
As per the OCP doc we need to have RHEL 7.9 as Base OS https://docs.openshift.com/container-platform/4.8/machine_management/adding-rhel-compute.html#rhel-compute-requirements_adding-rhel-compute

ocs_ci/deployment/helpers/vsphere_helpers.py
```diff
@@ -144,7 +144,14 @@ def modify_scaleup_repo(self):
                 "root_disk_size", "120"
             )
 
-            dict_data["vsphere"]["vcsa-qe"]["image"] = config.ENV_DATA["rhel_template"]
+            if Version.coerce(get_ocp_version()) >= Version.coerce("4.9"):
```
petr-balogh, line 147: Please change to new approach

petr-balogh (approved):

clacroix12 (approved):

PR #4874, opened 2021-09-23
Add marker for aws based platforms only test 

suchita-g:
ROSA is Redhat openshift service on AWS. Hence the test
applicable on aws should be consider for ROSA platform as well

### DIFF
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -143,6 +143,13 @@ aws_platform_required = pytest.mark.skipif(
     reason="Test runs ONLY on AWS deployed cluster",
 )
 
+aws_based_platform_required = pytest.mark.skipif(
+    (
+        config.ENV_DATA["platform"].lower() != "aws"
+        and config.ENV_DATA["platform"].lower() != ROSA_PLATFORM
+    ),
+    reason="Test runs ONLY on AWS based deployed cluster",
+)
 azure_platform_required = pytest.mark.skipif(
     config.ENV_DATA["platform"].lower() != "azure",
     reason="Test runs ONLY on Azure deployed cluster",
--- a/tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_proactive_IPI.py
+++ b/tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_proactive_IPI.py
@@ -5,7 +5,7 @@ from ocs_ci.framework.testlib import (
     tier4,
     tier4b,
     ManageTest,
-    aws_platform_required,
+    aws_based_platform_required,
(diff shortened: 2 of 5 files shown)


tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_proactive_IPI.py
```diff
@@ -6,6 +6,7 @@
     tier4b,
     ManageTest,
     aws_platform_required,
+    rosa_platform_required,
```
fbalak, line 9: I don't think that this will work because the tests will be marked as skipped by `aws_platform_required` marker and the marker `rosa_platform_required` won't undo it.

vasukulkarni:
why do we need a separate marker? we can assume the aws_platform_required applies to rosa and document it in ocs-ci?

suchita-g:
> why do we need a separate marker? we can assume the aws_platform_required applies to rosa and document it in ocs-ci?
Fir now MCG/Nooba feature is applicable to ODF on AWS however same is not applicable for i.e ODF-managed service on ROSA. Also upgrade is a different mechanism for ODF-MS. So I think a separate marker is a superior option.   
e.g.https://github.com/red-hat-storage/ocs-ci/blob/master/tests/ecosystem/upgrade/test_noobaa.py#L30

Please suggest if you can think of any robust way to achieve it.

suchita-g:
Verification-1 on ROSA platform: https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/6158/

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -143,6 +143,13 @@
     reason="Test runs ONLY on AWS deployed cluster",
 )
 
+aws_based_platform_required = pytest.mark.skipif(
+    (
+        config.ENV_DATA["platform"].lower() != "aws"
+        and config.ENV_DATA["platform"].lower() != ROSA_PLATFORM
```
fbalak, line 149:
```suggestion
        or config.ENV_DATA["platform"].lower() != ROSA_PLATFORM
```

fbalak (changes requested):

ocs_ci/framework/pytest_customization/marks.py
suchita-g: If the platform is not aws AND not rosa then skip TC
if either platform then the condition will be false and will not skip the TCs.

fbalak: You are right. The condition looks ok.

fbalak (approved):

yitzhak12 (approved):

PR #4872, opened 2021-09-22
Fix version comparison

petr-balogh:
Partially fixes: #4066

PR #4886, opened 2021-09-24
Fix version comparison for more modules

petr-balogh:
I am opening the second PR with more modules changed here:
https://github.com/red-hat-storage/ocs-ci/pull/4872/

The first PR I will let only few commits related to deployment and version module


ramkiperiy (dismissed):
LGTM

clacroix12:
@petr-balogh Maybe I'm missing something but if the `version` utility module and its corresponding unit tests were added in #4872 why are they also in this PR?

petr-balogh:
> @petr-balogh Maybe I'm missing something but if the `version` utility module and its corresponding unit tests were added in #4872 why are they also in this PR?

@clacroix12 this PR was based on previous one where I didn't touch the original commits as this has to be dependent on that. As commits are same github/git should merge only relevant commits out of this PR. I can do rebase as the previous PR got merged so UI will not show the commits which are already merged from previous one.

petr-balogh:
I just did rebase on the master now, checked the verification job and doesn't look like it's related to the change.

dahorak (approved):

ramkiperiy (approved):
LGTM

Issue #4096, opened 2021-04-05
change_default_storageclass() fails if there is an existing default SC present in the cluster


rachael-george:
If there is an existing default storageclass present in the cluster, the [change_default_storageclass()](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/helpers/helpers.py#L1644-L1675) method fails while executing the `oc patch` command on the existing SC. This is because the list of  SCs from  [get_default_storage_class()](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/helpers/helpers.py#L1621-L1641) gets passed to the `oc patch` command  instead of the name of the SC itself. 

Example: 
Test result link: http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/sgatfane-rhv2/sgatfane-rhv2_20210322T180859/logs/test_report_1617115977.html
- tests/manage/storageclass/test_create_sc_and_make_it_as_a_default.py::TestCreateStorageClassandMakeItAsDefault::test_create_sc_and_make_it_as_a_default[CephBlockPool]
- tests/manage/storageclass/test_create_sc_and_make_it_as_a_default.py::TestCreateStorageClassandMakeItAsDefault::test_create_sc_and_make_it_as_a_default[CephFileSystem]

```
E           ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc patch storageclass ['ovirt-csi-sc'] -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}' .
E           Error is Error from server (NotFound): storageclasses.storage.k8s.io "[ovirt-csi-sc]" not found
```

PR #4899, opened 2021-10-01
Update change_default_storageclass() to iterate over a list of SC

rachael-george:
Fixes: #4096 

### DIFF
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -1805,13 +1805,14 @@ def change_default_storageclass(scname):
     ocp_obj = ocp.OCP(kind="StorageClass")
     if default_sc:
         # Change the existing default Storageclass annotation to false
-        patch = (
-            ' \'{"metadata": {"annotations":'
-            '{"storageclass.kubernetes.io/is-default-class"'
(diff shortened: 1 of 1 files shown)


ocs_ci/helpers/helpers.py
```diff
@@ -1805,13 +1805,14 @@ def change_default_storageclass(scname):
...
     if default_sc:
         # Change the existing default Storageclass annotation to false
-        patch = (
-            ' \'{"metadata": {"annotations":'
-            '{"storageclass.kubernetes.io/is-default-class"'
-            ':"false"}}}\' '
-        )
-        patch_cmd = f"patch storageclass {default_sc} -p" + patch
-        ocp_obj.exec_oc_cmd(command=patch_cmd)
+        for sc in default_sc:
```
ebenahar, line 1808: What if `default_sc` is not a list?

rachael-george: `get_default_storage_class()` in line 1770 returns a list of SCs and is assigned to `default_sc`. In case there are no default SCs in the cluster, this code will not be executed and it will only run the patch command to change the new SC as the default SC.

PR #4914, opened 2021-10-04
Add ReportPortal launch attributes to the JUnit report XML as custom properties

clacroix12:


clacroix12:
The goal of this PR is to start adding data we tag our ReportPortal launches with to our JUnit report XML. We want to do this so we can continue to tag the same information when [we move to utilizing rp_preproc](https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/-/merge_requests/768).

PR #4924, opened 2021-10-06
Enable console plugin

vavuthu:

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -979,9 +979,11 @@ class Deployment(object):
         ):
             logger.info("Enabling console plugin")
             ocp_obj = ocp.OCP()
-            patch = '\'[{"op": "add", "path": "/spec", "value": {"plugins": ["odf-console"]}}]\''
+            patch = (
+                '\'[{"op": "add", "path": "/spec/plugins", "value": ["odf-console"]}]\''
+            )
             patch_cmd = (
-                f"patch console.operator.openshift.io cluster -n {constants.OPENSHIFT_STORAGE_NAMESPACE}"
+                f"patch console.operator cluster -n {constants.OPENSHIFT_STORAGE_NAMESPACE}"
                 f" --type json -p {patch}"
             )
             ocp_obj.exec_oc_cmd(command=patch_cmd)
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -971,6 +971,21 @@ class Deployment(object):
             # Change registry backend to OCS CEPHFS RWX PVC
             registry.change_registry_backend_to_ocs()
 
+        # Enable console plugin
+        ocs_version = version.get_semantic_ocs_version_from_config()
+        if (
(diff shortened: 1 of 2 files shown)


ocs_ci/deployment/deployment.py
```diff
@@ -971,6 +971,21 @@ def deploy_ocs(self):
...
 
+        # Enable console plugin
+        ocs_version = version.get_semantic_ocs_version_from_config()
+        if (
+            ocs_version >= version.VERSION_4_9
+            and config.ENV_DATA["enable_console_plugin"]
+        ):
+            logger.info("Enabling console plugin")
+            ocp_obj = ocp.OCP()
+            patch = f'\'[{{"op": "add", "path": "/spec", "value": {"plugins": ["odf-console"]}}}]\''
```
petr-balogh, line 982: Maybe I am blind but do you expand some variable here with f-string? As I don't see any, I think that `f` should be deleted and if you escaping `{` by `{{` the whole command should be modified as well I think.

ocs_ci/deployment/deployment.py
```diff
@@ -971,6 +971,21 @@ def deploy_ocs(self):
...
+            patch_cmd = (
+                f"patch -n {constants.OPENSHIFT_STORAGE_NAMESPACE} --type json -p "
+                + patch
```
petr-balogh, line 985:
```suggestion
                f"patch -n {constants.OPENSHIFT_STORAGE_NAMESPACE} --type json -p {patch}"
```

petr-balogh (dismissed):

ocs_ci/deployment/deployment.py
vavuthu: Done

ocs_ci/deployment/deployment.py
vavuthu: we are not expanding variable. updated in latest commit

petr-balogh (dismissed):

petr-balogh (approved):

ocs_ci/deployment/deployment.py
```diff
@@ -971,6 +971,23 @@ def deploy_ocs(self):
...
+            logger.info("Enabling console plugin")
+            ocp_obj = ocp.OCP()
+            patch = (
+                '\'[{"op": "add", "path": "/spec/plugins", "value": ["odf-console"]}]\''
+            )
+            patch_cmd = (
+                f"patch console.operator cluster -n {constants.OPENSHIFT_STORAGE_NAMESPACE}"
+                f" --type json -p {patch}"
+            )
+            ocp_obj.exec_oc_cmd(command=patch_cmd)
```
clacroix12, line 989: Logically looks good to me but I'd like to suggest we move this to a function we can call here instead. `deploy_ocs` is a pretty big method already and I'd like to limit the amount of code we are adding to it directly. If you agree this could definitely be taken up as a follow-up, no need to hold this PR up.

vavuthu: ack. I will take it up as follow-up

Issue #4939, opened 2021-10-10
version collection failed on GA version


Avilir:
while running test on GA version (of OCP & OCS), test faild to run on the version collection phase :
```
Current date is 20211010-1457
Log file is : /home/avili/Testing/Clusters/AWSCL-M5/test_small_file_workload.py_20211010-1457.txt
[runtest] Running : run-ci  --log-level=DEBUG --cluster-name alayani-m5cl --cluster-path /home/avili/Testing/Clusters/AWSCL-M5 --ocsci-conf /home/avili/Testing/OCS-CI/conf/avili/aws-m5.yaml tests/e2e/performance/test_small_file_workload.py::TestSmallFileWorkload::test_smallfile_workload[16-200000-4-3-CephBlockPool] | tee /home/avili/Testing/Clusters/AWSCL-M5/test_small_file_workload.py_20211010-1457.txt
2021-10-10 14:58:09,064 - MainThread - ocs_ci.framework.pytest_customization.ocscilib - INFO - Dump of the consolidated config file is located here: /tmp/run-1633867075-config.yaml
2021-10-10 14:58:36,120 - MainThread - ocs_ci.framework.pytest_customization.ocscilib - ERROR - Failed to gather version info!
Traceback (most recent call last):
  File "/home/avili/Testing/OCS-CI/ocs_ci/framework/pytest_customization/ocscilib.py", line 375, in gather_version_info_for_report
    mods = get_version_info(namespace=ocsci_config.ENV_DATA["cluster_namespace"])
  File "/home/avili/Testing/OCS-CI/ocs_ci/ocs/resources/ocs.py", line 189, in get_version_info
    csv_name = package_manifest.get_current_csv(channel)
  File "/home/avili/Testing/OCS-CI/ocs_ci/ocs/resources/packagemanifest.py", line 167, in get_current_csv
    channels = self.get_channels()
  File "/home/avili/Testing/OCS-CI/ocs_ci/ocs/resources/packagemanifest.py", line 138, in get_channels
    return self.data["status"]["channels"]
TypeError: list indices must be integers or slices, not str
2021-10-10 14:58:36,124 - MainThread - ocs_ci.framework.pytest_customization.ocscilib - WARNING - Failed to gather version details! The report of version mightnot be complete!
Collecting Cluster versions
INTERNALERROR> Traceback (most recent call last):
[...]

PR #4940, opened 2021-10-10
Fix for #4939

Avilir:

### DIFF
--- a/ocs_ci/ocs/resources/packagemanifest.py
+++ b/ocs_ci/ocs/resources/packagemanifest.py
@@ -135,7 +135,10 @@ class PackageManifest(OCP):
         """
         self.check_name_is_specified()
         try:
-            return self.data["status"]["channels"]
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/resources/packagemanifest.py
```diff
@@ -135,7 +135,10 @@ def get_channels(self):
         """
         self.check_name_is_specified()
         try:
-            return self.data["status"]["channels"]
+            if isinstance(self.data, list):
+                return self.data[0]["status"]["channels"]
```
dahorak, line 139: Are we sure, that the expected one is always the first one?

Avilir: yes, i have tested it with few GA version, and the used channel was all the time the first one in the list.

OdedViner (approved):

PR #4956, opened 2021-10-13
Added log on get_ocp_version function

OdedViner:

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -1584,6 +1584,7 @@ def get_ocp_version(seperator=None):
     else:
         raw_version = config.DEPLOYMENT["installer_version"]
     version = Version.coerce(raw_version)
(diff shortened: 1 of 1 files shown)


ocs_ci/utility/utils.py
```diff
@@ -1584,6 +1584,7 @@ def get_ocp_version(seperator=None):
     else:
         raw_version = config.DEPLOYMENT["installer_version"]
     version = Version.coerce(raw_version)
+    log.info(f"The OCP Version is {str(version.major)}.{str(version.minor)}")
```
clacroix12, line 1587: We use this function is a lot of places and I think this will add to our already overly verbose logs at `INFO` level. I think it would be better to drop this to `DEBUG` or to add this message in the particular place in the code (after you call `get_ocp_version()`) where you want to see the actual version that was retrieved.

PR #4979, opened 2021-10-19
Skip the deletion flow for ULSs that were not created

MeridianExplorer:
Currently, CloudManager lists all the buckets in all of the storage providers for which credentials are supplied.
This takes a lot of time, and is not needed when no underlying storage was created in particular storage providers.

### DIFF
--- a/ocs_ci/ocs/resources/cloud_uls.py
+++ b/ocs_ci/ocs/resources/cloud_uls.py
@@ -105,16 +105,16 @@ def cloud_uls_factory(request, cld_mgr):
 
     def uls_cleanup():
         for cloud, uls_set in all_created_uls.items():
-            if len(uls_set) > 0:
-                client = ulsMap.get(cloud)
-                if client is not None:
-                    all_existing_uls = client.get_all_uls_names()
-                    for uls in uls_set:
-                        if uls in all_existing_uls:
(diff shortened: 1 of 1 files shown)


bernerhat (dismissed):

ocs_ci/ocs/resources/cloud_uls.py
```diff
@@ -105,15 +105,16 @@ def _create_uls(uls_dict):
...
-            client = ulsMap.get(cloud)
-            if client is not None:
-                all_existing_uls = client.get_all_uls_names()
-                for uls in uls_set:
-                    if uls in all_existing_uls:
-                        log.info(f"Cleaning up uls {uls}")
-                        client.delete_uls(uls)
-                    else:
-                        log.warning(f"Underlying Storage {uls} not found.")
+            if len(uls_set) > 0:
```
mbukatov, line 108: It would be more readable if you just do "if len(uls_set) == 0: continue" instead (avoiding too nested branching).

MeridianExplorer: Awesome and elegant suggestion, thank you!

bernerhat (approved):

ebenahar (approved):

PR #4991, opened 2021-10-20
Increased vmware IPI deployment timeout to 1.30hrs

ramkiperiy:
Increase vmware IPI deployment timeout to 1.30hrs for testing issues in DC_13 & DC_15


mbukatov:
@ramkiperiy If you needed this to test a particular change without any plans to have this reviewed and merged, don't create a pull request, but just a branch, which you can then use instead of master when configuring a cluster deploy job.

mbukatov (changes requested):
This seems to not be expected to be merged. If that is the case, please close the PR.

PR #4994, opened 2021-10-20
added option for different url login

### LINKED ISSUE #4993: Need to add UI login for other URLs
it will be needed for ACM, which use another URL

AviadP:
added function that getting default OCP URL, and login_ui will use it as its default
fix #4993 

### DIFF
--- a/ocs_ci/ocs/ui/base_ui.py
+++ b/ocs_ci/ocs/ui/base_ui.py
@@ -625,6 +625,8 @@ def login_ui(console_url=None):
     """
     Login to OpenShift Console
 
+    Args:
+        console_url (str): ocp console url
 
     return:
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/ui/base_ui.py
```diff
@@ -618,21 +618,30 @@ def take_screenshot(driver):
...
+    Returns:
+        str: OCP console URL
+
+    """
+    logger.info("Get URL of OCP console")
+    return run_cmd(
+        "oc get consoles.config.openshift.io cluster -o"
+        "jsonpath='{.status.consoleURL}'"
+    )
+
```
OdedViner, line 633: Can move this function to helpers file?

PR #5052, opened 2021-11-05
Fix `test_host_node_failure` to wait longer for OBC to bind after node restart

MeridianExplorer:
Also added general support for passing `timeout` and `interval` functions to the bucket factory.


ebenahar (approved):

parikshithb (approved):

MeridianExplorer:
Verified locally

PR #5055, opened 2021-11-05
add support for VM query on  VSPHERE IPI

annagitel:
Signed-off-by: Anna Sandler <asandler@redhat.com>
until today, VM query was using resource pools. In IPI deployments there are no pools
this PR checks for deployment type:
UPI - was not changed. using pools.
IPI - added query using DC

### DIFF
--- a/ocs_ci/ocs/platform_nodes.py
+++ b/ocs_ci/ocs/platform_nodes.py
@@ -217,14 +217,25 @@ class VMWareNodes(NodesBase):
             list: vSphere vm objects list
 
         """
-        vms_in_pool = self.vsphere.get_all_vms_in_pool(
-            self.cluster_name, self.datacenter, self.cluster
-        )
-        node_names = [node.get().get("metadata").get("name") for node in nodes]
+        vms_all = []
         vms = []
+        logger.info(f"deployment type: {self.deployment_type}")
+        if self.deployment_type == "upi":
+            vms_all = self.vsphere.get_all_vms_in_pool(
+                self.cluster_name, self.datacenter, self.cluster
+            )
+        elif self.deployment_type == "ipi":
+            vms_all = self.vsphere.get_all_vms_in_dc(self.datacenter)
+
+        else:
+            logger.error("Deployment type not detected")
+
+        node_names = [node.get().get("metadata").get("name") for node in nodes]
+
         for node in node_names:
-            node_vms = [vm for vm in vms_in_pool if vm.name in node]
+            node_vms = [vm for vm in vms_all if vm.name in node]
             vms.extend(node_vms)
+
         return vms
 
     def get_data_volumes(self, pvs=None):
--- a/ocs_ci/ocs/platform_nodes.py
+++ b/ocs_ci/ocs/platform_nodes.py
@@ -220,15 +220,16 @@ class VMWareNodes(NodesBase):
         vms_all = []
         vms = []
         logger.info(f"deployment type: {self.deployment_type}")
-        if self.deployment_type == "upi":
+        if self.deployment_type == constants.DEPLOYMENT_TYPE_UPI:
             vms_all = self.vsphere.get_all_vms_in_pool(
                 self.cluster_name, self.datacenter, self.cluster
             )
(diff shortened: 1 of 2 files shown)


ebenahar:
We still need to run this test on vsphere IPI. It's just that we need to find an alternative way for IPI which is not with resource pool

ocs_ci/ocs/platform_nodes.py
```diff
@@ -217,14 +217,25 @@ def get_vms(self, nodes):
...
 
         """
-        vms_in_pool = self.vsphere.get_all_vms_in_pool(
-            self.cluster_name, self.datacenter, self.cluster
-        )
-        node_names = [node.get().get("metadata").get("name") for node in nodes]
+        vms_all = []
         vms = []
+        logger.info(f"deployment type: {self.deployment_type}")
+        if self.deployment_type == "upi":
```
ebenahar, line 223: please create constants for deployment type IPI and UPI and use them here

ocs_ci/ocs/platform_nodes.py
```diff
@@ -217,14 +217,25 @@ def get_vms(self, nodes):
...
+            vms_all = self.vsphere.get_all_vms_in_pool(
+                self.cluster_name, self.datacenter, self.cluster
+            )
+        elif self.deployment_type == "ipi":
+            vms_all = self.vsphere.get_all_vms_in_dc(self.datacenter)
+
+        else:
+            logger.error("Deployment type not detected")
```
ebenahar, line 231: I think we should return None here if deployment type is not detected

ocs_ci/ocs/platform_nodes.py
```diff
@@ -217,14 +217,25 @@ def get_vms(self, nodes):
...
-        node_names = [node.get().get("metadata").get("name") for node in nodes]
+        vms_all = []
         vms = []
+        logger.info(f"deployment type: {self.deployment_type}")
+        if self.deployment_type == "upi":
+            vms_all = self.vsphere.get_all_vms_in_pool(
+                self.cluster_name, self.datacenter, self.cluster
+            )
+        elif self.deployment_type == "ipi":
+            vms_all = self.vsphere.get_all_vms_in_dc(self.datacenter)
```
ebenahar, line 228: Wouldn't it mean that if you are getting all the VMs in the datacenter, you will get not only the VMs in the relevant ODF cluster but all other VMs in the DC, of all other ODF clusters running in that DC?

ebenahar (changes requested):

ocs_ci/ocs/platform_nodes.py
annagitel: yes, but then I filter the once I need by the `nodes` variable passed to the function

ocs_ci/ocs/platform_nodes.py
annagitel: done

ocs_ci/ocs/platform_nodes.py
ebenahar: oh I see. Thanks for clarifting

ocs_ci/ocs/platform_nodes.py
ebenahar: does the test process None if passed?

annagitel: there are some tests that use this function indirectly, it's a really "inner" function. they should fail. 
but this less likely should happen because a stable and correct VMware deployment will have no problem returning deployment type. else is there just to match the if/elif/else

ebenahar: ack

ebenahar (approved):

yitzhak12:
I think you may miss the change in my PR https://github.com/red-hat-storage/ocs-ci/pull/5109/files#diff-478e5b871f475f1e312384c0eeb216726f11da23e3d3ac4df80aa3ee5d92e441R2574. As part of the pr, I implemented the function `get_vms` for vSphere ipi.

annagitel:
duplicate of function fixed in #5109

PR #5063, opened 2021-11-08
OCS 4.8 to ODF 4.9 UI Upgrade

am-agrawa:
Signed-off-by: Aman Agrawal <amagrawa@redhat.com>

The test is to upgrade OCS 4.8 to ODF 4.9 via UI. Post upgrade validations are also part of this test function.


ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -331,9 +340,14 @@ def search_operator_installed_operators_page(self, operator=OCS_OPERATOR):
             self.choose_expanded_mode(
                 mode=True, locator=self.dep_loc["drop_down_projects"]
             )
-            self.do_click(
-                self.dep_loc["enable_default_porjects"], enable_screenshot=True
-            )
+            default_projects_is_checked = self.driver.find_element_by_id(
+                "no-label-switch-on"
+            ).is_selected()
+            if default_projects_is_checked is False:
```
vavuthu, line 346:
```suggestion
            if not default_projects_is_checked:
```

tests/ui/test_ocs_odf_upgrade_ui.py
```diff
@@ -0,0 +1,84 @@
+from ocs_ci.ocs.ui.base_ui import logger
+from ocs_ci.ocs.ui.views import locators, ODF_OPERATOR
+from ocs_ci.utility.utils import get_ocp_version
+from ocs_ci.framework import config
+from ocs_ci.ocs.ocs_upgrade import OCSUpgrade
+from ocs_ci.ocs.ui.deployment_ui import DeploymentUI
+from ocs_ci.ocs.ui.validation_ui import ValidationUI
+from ocs_ci.ocs.resources.storage_cluster import ocs_install_verification
+from ocs_ci.framework.testlib import ManageTest, skipif_ocs_version, skipif_ocp_version
```
vavuthu, line 9: for readability, prioritize the imports using alphabetical order

e.g:
```
from Z.A import v
from Z.A.B import j
from Z.A.B.C import y
```

tests/ui/test_ocs_odf_upgrade_ui.py
```diff
@@ -0,0 +1,84 @@
...
+            upgrade_in_current_source=upgrade_in_current_source,
+        )
+        val_obj = ValidationUI(setup_ui)
+        pagenav_obj = ValidationUI(setup_ui)
+
+        channel = upgrade_odf.set_upgrade_channel()
+        csv_name_pre_upgrade = upgrade_odf.get_csv_name_pre_upgrade()
+
+        ocs_registry_image = config.DEPLOYMENT.get("ocs_registry_image")
+        logger.info(f"------------------------- {ocs_registry_image}")
```
vavuthu, line 45: can we log appropriate message or remove the line

tests/ui/test_ocs_odf_upgrade_ui.py
```diff
@@ -0,0 +1,84 @@
...
+from ocs_ci.framework import config
+from ocs_ci.ocs.ocs_upgrade import OCSUpgrade
+from ocs_ci.ocs.ui.deployment_ui import DeploymentUI
+from ocs_ci.ocs.ui.validation_ui import ValidationUI
+from ocs_ci.ocs.resources.storage_cluster import ocs_install_verification
+from ocs_ci.framework.testlib import ManageTest, skipif_ocs_version, skipif_ocp_version
+
+
+@skipif_ocp_version("<4.9")
+@skipif_ocs_version("<4.8")
```
petr-balogh, line 13: I think that this is also not valid when upgrading between z-streams upgrade of 4.9.0 to 4.9.1 and for 4.9->4.10 upgrade.
So this will require some other logic to cover those skips. Which can be actually done as first lines in test to be able to get version from CSV and play with z-stream version as well

tests/ui/test_ocs_odf_upgrade_ui.py
```diff
@@ -0,0 +1,84 @@
...
+class TestOcsOdfUpgrade(ManageTest):
+
+    ocp_version = get_ocp_version()
+    dep_loc = locators[ocp_version]["deployment"]
+    validation_loc = locators[ocp_version]["validation"]
+    operator = ODF_OPERATOR
+
+    def test_ocs_odf_upgrade(self, setup_ui):
```
petr-balogh, line 21: I would call such upgrade from our regular upgrade test in the case some config value to do UI upgrade is provided to ocs-ci and do UI upgrade instead of CLI as we have now.

We have upgrade job sets to call upgrade ocs-ci run with such markers:
`pre_upgrade or pre_ocs_upgrade or ocs_upgrade or post_ocs_upgrade or post_upgrade`

So the logic still has to be same in order to keep also specific order of the tests. O this logic here should not be defined as test but as regular class/methods and inside our regular upgrade test to decide if do CLI upgrade or the new UI. Then the same verification like we do for checking images and so on has to be invoked as well.

ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -75,8 +76,16 @@ def install_ocs_operator(self):
             self.do_click(self.dep_loc["enable_console_plugin"], enable_screenshot=True)
         self.do_click(self.dep_loc["click_install_ocs_page"], enable_screenshot=True)
         if self.operator is ODF_OPERATOR:
-            time.sleep(60)
-            self.refresh_page()
+            time.sleep(90)
+            refresh_web_console_popup = self.wait_until_expected_text_is_found(
```
OdedViner, line 80: Can you create new method on this class for `refresh_web_console_popup`?

ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -75,8 +76,16 @@ def install_ocs_operator(self):
             self.do_click(self.dep_loc["enable_console_plugin"], enable_screenshot=True)
         self.do_click(self.dep_loc["click_install_ocs_page"], enable_screenshot=True)
         if self.operator is ODF_OPERATOR:
-            time.sleep(60)
-            self.refresh_page()
+            time.sleep(90)
```
ebenahar, line 79: waiting for a fixed time period may result in either not having enough time to wait or redundant time we continue waiting even after the state is achieved.
Could you please change this to wait for the resource mechanism?

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -541,32 +553,43 @@ def run_ocs_upgrade(operation=None, *operation_args, **operation_kwargs):
...
+        if config.UPGRADE.get("ui_upgrade"):
+            ocp_version = get_ocp_version()
+
+            if (
+                Version.coerce(ocp_version) == Version.coerce("4.9")
+                and original_ocs_version == "4.8"
+                and upgrade_version == "4.9"
+            ):
+                ocs_odf_upgrade_ui()
+        else:
```
petr-balogh, line 569: If someone will run this on different combination than this one you will basically do not run upgrade at all and user will even get any input.
So for such case I prefer to change the logic here little bit different. 
```suggestion
        ui_upgrade_supported = False
        if config.UPGRADE.get("ui_upgrade"):
            ocp_version = get_ocp_version()

            if (
                Version.coerce(ocp_version) == Version.coerce("4.9")
                and original_ocs_version == "4.8"
                and upgrade_version == "4.9"
            ):
                ui_upgrade_supported = True
            # you can add more combinations later on here or update the condition above.
            log.warning("UI upgrade combination is not supported. It will fallback to CLI upgrade")
        if ui_upgrade_supported:
            ocs_odf_upgrade_ui()
        else:
```

Other approach will be to raise an exception that UI upgrade is not supported after the line 568 but indented out of the if block. But I think that better is to fallback to CLI without spending time on all pre upgrade tests and then fail with exception to upgrade.

petr-balogh (dismissed):

jilju (dismissed):

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -541,64 +553,112 @@ def run_ocs_upgrade(operation=None, *operation_args, **operation_kwargs):
...
+            ocp_version = get_ocp_version()
+            if (
+                Version.coerce(ocp_version) == Version.coerce("4.9")
+                and original_ocs_version == "4.8"
+                and upgrade_version == "4.9"
+            ):
+                ui_upgrade_supported = True
+            else:
+                log.warning(
+                    "UI upgrade combination is not supported. It will fallback to CLI upgrade"
```
vavuthu, line 571: why we need to fall back to CLI upgrade instead of raising exception? Use is expected to do upgrade through UI

ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -24,6 +25,11 @@ class DeploymentUI(PageNavigator):
     def __init__(self, driver):
         super().__init__(driver)
         self.dep_loc = locators[self.ocp_version]["deployment"]
+        self.validation_loc = locators[self.ocp_version]["validation"]
+        ocs_version = version.get_semantic_ocs_version_from_config()
+        self.operator = (
+            ODF_OPERATOR if ocs_version >= version.VERSION_4_9 else OCS_OPERATOR
+        )
```
OdedViner, line 32: Why did you add these lines?

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -541,64 +553,112 @@ def run_ocs_upgrade(operation=None, *operation_args, **operation_kwargs):
...
-                    "of linking secret/SAs"
-                )
-                time.sleep(60)
-                link_all_sa_and_secret_and_delete_pods(
-                    constants.OCS_SECRET, config.ENV_DATA["cluster_namespace"]
+        ui_upgrade_supported = False
+        if config.UPGRADE.get("ui_upgrade"):
+            ocp_version = get_ocp_version()
+            if (
+                Version.coerce(ocp_version) == Version.coerce("4.9")
```
vavuthu, line 564:
```suggestion
                version.get_semantic_ocp_version_from_config() == version.VERSION_4_9
```
we can use version module here : https://github.com/red-hat-storage/ocs-ci/blob/f1674710eb69117d9e1d5a262740c292f4ac6d67/ocs_ci/utility/version.py

OdedViner (changes requested):
We dont need to change __init__ contractor on deployment_ui class

vavuthu:
since this PR touches upgrade paths, I would like to test both UI upgrade and CLI upgrade as well

ocs_ci/ocs/ocs_upgrade.py
am-agrawa: @vavuthu Petr suggested this above, and it makes sense to me. What do you think?

ocs_ci/ocs/ocs_upgrade.py
vavuthu: @petr-balogh Its better to raise exception instead of falling back to CLI for upgrade if not supported, because user is expecting upgrade through UI and the results are misleading if we do CLI upgrade.

petr-balogh: @vavuthu the same approach we use in UI deployment. If UI deployment is not implemented we fall back to CLI. If we already have the same approach in deployment, I think we should follow the same in upgrade.

ocs_ci/ocs/ocs_upgrade.py
petr-balogh: Aman is getting actual version of the cluster from get_ocp_version which actually taking actual version not one which is passed via --ocp-version.  So to get it with this approach I suggest here to change to:

```
version.get_semantic_version(ocp_version, True) ==  version.VERSION_4_9
```
This should keep the logic and will work with compare to version.VERSION_4_9 - as this one is generated with only_major_minor = True.

ocs_ci/ocs/ocs_upgrade.py
petr-balogh: We have the same fallback also for deployment

ocs_ci/ocs/ocs_upgrade.py
vavuthu: synced with @petr-balogh and agreed with falling back to CLI

ocs_ci/ocs/ui/deployment_ui.py
am-agrawa: @OdedViner Updated, please check

vavuthu (approved):

conf/upgrade/upgrade_ui.yaml
```diff
@@ -0,0 +1,3 @@
+---
```
petr-balogh, line 1: You can add some description here in comment and in conf/README.md it should be described there as well as we used to document new config values.

petr-balogh (approved):

am-agrawa:
Hi @petr-balogh @vavuthu @OdedViner @ebenahar All the feedbacks are implemented and the PR validation for both CLI and UI upgrade has passed. Can I please get the final approval to merge this PR?

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -541,32 +549,48 @@ def run_ocs_upgrade(operation=None, *operation_args, **operation_kwargs):
...
+                ui_upgrade_supported = True
+            else:
+                log.warning(
+                    "UI upgrade combination is not supported. It will fallback to CLI upgrade"
                 )
+            if ui_upgrade_supported:
+                ocs_odf_upgrade_ui()
+        else:
+            if upgrade_version != "4.9":
+                # In the case of upgrade to ODF 4.9, the ODF operator should upgrade
```
ebenahar, line 572: won't it be the case also for upgrade_version > 4.9?

ebenahar: discussed offline and will be taken care of by a separate PR

ebenahar (dismissed):

ocs_ci/ocs/ocs_upgrade.py
am-agrawa: I was thinking that this condition is already meeting the criteria **upgrade_version > 4.9** because the code says , if **upgrade_version != "4.9":**, so not equal to means lesser or greater both. Or Am I wrong?

OdedViner (approved):

vavuthu (dismissed):

conf/README.md
```diff
@@ -222,6 +222,7 @@ Upgrade related configuration data.
 * `ocp_upgrade_path` - OCP image to upgrade with
 * `ocp_arch` - Architecture type of the OCP image
 * `upgrade_logging_channel` - OCP logging channel to upgrade with
+* `upgrade_ui` - Upgrade from OCS 4.8 to ODF 4.9 via UI (Default: true)
```
petr-balogh, line 225: Please remove Default: true from here as it's not part of default config as well: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml

It's true only when you provide config file ;)

Also please remove the versions from the descriptions, this will be I guess used also for later versions upgrade once it's implemented and it's not worth to mention versions here. 

```suggestion
* `upgrade_ui` - Perform upgrade via UI. (Not all the versions are supported, please look at the code)
```

petr-balogh (approved):

ocs_ci/ocs/ocs_upgrade.py
am-agrawa: As discussed offline, this looks fine. :)

PR #5095, opened 2021-11-15
Noobaa system testing

OdedViner:

### DIFF
--- a/ocs_ci/ocs/bucket_utils.py
+++ b/ocs_ci/ocs/bucket_utils.py
@@ -1424,3 +1424,66 @@ def write_random_test_objects_to_bucket(
         signed_request_creds=s3_creds,
     )
     return obj_lst
+
+
+def write_data_to_buckets_and_check_integrity(
+    mcg_obj, awscli_pod, bucket_factory, bucketclass_dict, number_of_buckets
+):
+    """
+    Write data on several buckets and check data integrity with md5sum
+
+    Returns:
+        dictionary: Data on buckets including downloaded_files, original_dir,
+        result_dir, awscli_pod
+
+    """
+    buckets_data_dic = dict()
+    for bucket_id in range(1, number_of_buckets + 1):
+        bucketname = bucket_factory(1, bucketclass=bucketclass_dict)[0].name
+        original_dir = f"/original{bucket_id}"
+        result_dir = f"/result{bucket_id}"
+        awscli_pod.exec_cmd_on_pod(command=f"mkdir {result_dir}")
+        # Retrieve a list of all objects on the test-objects bucket and
+        # downloads them to the pod
+        full_object_path = f"s3://{bucketname}"
+        downloaded_files = retrieve_test_objects_to_pod(awscli_pod, original_dir)
+        # Write all downloaded objects to the new bucket
+        sync_object_directory(awscli_pod, original_dir, full_object_path, mcg_obj)
+        # Retrieve all objects from MCG bucket to result dir in Pod
+        logger.info("Downloading all objects from MCG bucket to awscli pod")
+        sync_object_directory(awscli_pod, full_object_path, result_dir, mcg_obj)
+
+        # Checksum is compared between original and result object
+        for obj in downloaded_files:
+            assert verify_s3_object_integrity(
+                original_object_path=f"{original_dir}/{obj}",
+                result_object_path=f"{result_dir}/{obj}",
+                awscli_pod=awscli_pod,
+            ), "Checksum comparison between original and result object failed"
+
+        buckets_data_dic[bucketname] = {
+            "downloaded_files": downloaded_files,
+            "original_dir": original_dir,
+            "result_dir": result_dir,
+            "awscli_pod": awscli_pod,
+        }
+    return buckets_data_dic
+
+
+def verify_s3_buckets_integrity(buckets_data_dic):
+    """
+    Get number of buckets and check the data integrity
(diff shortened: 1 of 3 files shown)


tests/e2e/systemtest/test_noobaa_system.py
```diff
@@ -0,0 +1,39 @@
...
+from ocs_ci.framework.testlib import E2ETest
+from ocs_ci.ocs.bucket_utils import (
+    write_data_to_buckets_and_check_integrity,
+    verify_s3_buckets_integrity,
+)
+
+log = logging.getLogger(__name__)
+
+
+class TestNoobaaSystem(E2ETest):
```
hnallurv, line 12: Could we change this to TestNoobaaBackupRecoverySystemTest?

tests/e2e/systemtest/test_noobaa_system.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Test Noobaa System
+
+    """
+
+    def test_noobaa_system(
```
hnallurv, line 18: Please change name to test_noobaa_backup_recovery_st

tests/e2e/systemtest/test_noobaa_system.py
```diff
@@ -0,0 +1,39 @@
...
+        self,
+        bucket_factory,
+        mcg_obj,
+        awscli_pod,
+    ):
+        """
+        Test Procedure:
```
hnallurv, line 25: Please add test description. For example: "The objective of this test case is to verify that noobaa can be recovered successfully by following the steps mentioned in https://access.redhat.com/solutions/5843261, when db-noobaa-db-0 PVC is accidentally corrupted/deleted on a cluster which is configured with various noobaa/mcg feature like mcg bucket replication, noobaa caching, namespace obc etc.,"

tests/e2e/systemtest/test_noobaa_system.py
MeridianExplorer: Also, I think it'd be right to refactor the name to MCG throughout all files. 
Most of our code uses `MCG` and not NooBaa - MCG being the correct downstream terminology.

ocs_ci/ocs/bucket_utils.py
```diff
@@ -1424,3 +1424,66 @@ def write_random_test_objects_to_bucket(
...
+    buckets_data_dic = dict()
+    for bucket_id in range(1, number_of_buckets + 1):
+        bucketname = bucket_factory(1, bucketclass=bucketclass_dict)[0].name
+        original_dir = f"/original{bucket_id}"
+        result_dir = f"/result{bucket_id}"
+        awscli_pod.exec_cmd_on_pod(command=f"mkdir {result_dir}")
+        # Retrieve a list of all objects on the test-objects bucket and
+        # downloads them to the pod
+        full_object_path = f"s3://{bucketname}"
+        downloaded_files = retrieve_test_objects_to_pod(awscli_pod, original_dir)
```
MeridianExplorer, line 1449: `retrieve_test_objects_to_pod()` is no longer in use.
Thje objects are now integrated into the S3 CLI pod image, please replace with the `test_directory_setup` fixture (in the test's scope)

ocs_ci/ocs/bucket_utils.py
```diff
@@ -1424,3 +1424,66 @@ def write_random_test_objects_to_bucket(
         signed_request_creds=s3_creds,
     )
     return obj_lst
+
+
+def write_data_to_buckets_and_check_integrity(
```
MeridianExplorer, line 1429: Is this use case common enough to warrant a function of its own?
I think it'd probably be better to turn this into the test's logic.

tests/e2e/systemtest/test_noobaa_system.py
```diff
@@ -0,0 +1,39 @@
...
+        log.info("Write data on several buckets and check data integrity with md5sum")
+        buckets_data_dic = write_data_to_buckets_and_check_integrity(
+            mcg_obj=mcg_obj,
+            awscli_pod=awscli_pod,
+            bucket_factory=bucket_factory,
+            bucketclass_dict=None,
+            number_of_buckets=5,
+        )
+        log.info("Check the data integrity of buckets")
+        verify_s3_buckets_integrity(buckets_data_dic)
```
MeridianExplorer, line 39: Why is there a second integrity check after the integrity check built into `write_data_to_buckets_and_check_integrity`?

tests/e2e/systemtest/test_noobaa_system.py
```diff
@@ -0,0 +1,39 @@
...
+class TestNoobaaSystem(E2ETest):
+    """
+    Test Noobaa System
+
+    """
+
+    def test_noobaa_system(
+        self,
+        bucket_factory,
+        mcg_obj,
```
MeridianExplorer, line 21: `bucketclass_dict` is not passed as a parameter. The test needs to be parametrized and the signature needs to change in order to support it.

MeridianExplorer (changes requested):
This test is largely a duplication of `test_check_object_integrity`.
I suggest modifying `test_check_object_integrity` by including a `for` loop to enable parametrization of an amount of buckets, and then invoking it directly from `test_noobaa_system`.
Alternatively, it's possible to move the logic to a shared function that both tests will import and use.

The comments below are generally for better practices, but I think it'd be best to replace the code logic with a call to `test_check_object_integrity`

Issue #5080, opened 2021-11-10
Change of MCG operator CSV in build


petr-balogh:
See info here:
https://mailman-int.corp.redhat.com/archives/rhocs-eng/2021-November/msg00047.html

From build:
4.9.0-231 there should be mcg-operator name instead of noobaa-operator.

PR #5101, opened 2021-11-16
fix disconnected odf > 4.9.0-231 deployment

dahorak:
`noobaa-operator` package was renamed to `mcg-operator` (we can keep there both names, as it skips those which are not available)

Related to https://github.com/red-hat-storage/ocs-ci/issues/5080.


petr-balogh (approved):

vavuthu (approved):

dahorak:
The failed test in verification job seems to be because of some temporary network or external resource issue, the deployment itself passed correctly.

Issue #4488, opened 2021-06-29
Couchbase pods are not coming up 


shreekarSS:
From https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/1173/testReport/junit/tests.e2e.workloads.couchbase.test_couchbase/TestCouchBaseWorkload/test_cb_workload_simple/

E               ocs_ci.ocs.exceptions.TimeoutExpiredError: Timed out after 1800s running get_pod_name_by_pattern("cb-example", "couchbase-operator-namespace")

cb-example pods are not coming up

Akarsha-rai:
We are seeing couchbase workload failing to run from OCP 4.8 onward, i.e cb-example pod fails to come with Readiness probe failure as shown below:

From the events of cb-example-0000 pod in 4.8:
```
Normal   AddedInterface          22s   multus                   Add eth0 [10.129.2.153/23] from openshift-sdn
Normal   Pulled                  22s   kubelet                  Container image "registry.connect.redhat.com/couchbase/server:6.0.1-1" already present on machine
...
...
...
Normal   Created                 20s   kubelet                  Created container couchbase-server
Normal   Started                 20s   kubelet                  Started container couchbase-server
Warning  Unhealthy               7s    kubelet                  Readiness probe failed:
```

From cb-example-0000 pod log in 4.8:
```
[ns_server:warn,2021-07-23T06:58:29.126Z,ns_1@127.0.0.1:<0.549.0>:ns_memcached:connect:1227]Unable to connect: {error,{badmatch,{error,econnrefused}}}.
[ns_server:warn,2021-07-23T06:58:29.130Z,ns_1@127.0.0.1:<0.554.0>:ns_memcached:connect:1227]Unable to connect: {error,{badmatch,{error,econnrefused}}}
```
Seems like memcahed issue, not able to connect. Need to check with OCP and couchbase team on it
For more info: https://bugzilla.redhat.com/show_bug.cgi?id=2005281

PR #5108, opened 2021-11-17
Couchbase workload new implementation

PrasadDesala:
*. Couchbase workload is redesigned with new implementation. We use couchbase operator available in OperatorHub for Couchbase cluster deployment.
*. Fixes: #4488 

### DIFF
--- a/ocs_ci/ocs/couchbase_new.py
+++ b/ocs_ci/ocs/couchbase_new.py
@@ -1,0 +1,282 @@
+"""
+Couchbase workload class
+"""
+import logging
+import random
+from concurrent.futures.thread import ThreadPoolExecutor
+
+from ocs_ci.ocs.resources.packagemanifest import PackageManifest
+from ocs_ci.ocs.resources.csv import CSV
+from ocs_ci.ocs.resources.ocs import OCS
+from ocs_ci.ocs.ocp import OCP, switch_to_project
+from ocs_ci.utility import templating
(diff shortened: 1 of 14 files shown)


vasukulkarni (approved):

ocs_ci/ocs/couchbase_new.py
```diff
@@ -0,0 +1,282 @@
...
+        Args:
+            replicas (int): Number of pods
+            num_items (int): Number of items to be loaded to the cluster
+            num_threads (int): Number of threads
+            run_in_bg (bool) : Optional run IOs in background
+
+        """
+        self.result = None
+        logging.info("Running IOs using Pillow-fight")
+        if run_in_bg:
```
vasukulkarni, line 193: I think we can get rid of this if:  doens't add much value considering it runs in bg by default, only the oc delay is what we see.

PrasadDesala: run_pillowfights() reads the pillowfight yamls, waits for the pods to come up and wait till the pillowfight pods reaches completed state, this will take a good amount of time. This run_in_bg can be helpful for tests which needs a workload to run in background while the ocs operations are going on.  This is anyhow set to False by default and can be used whenever required in tests.

vasukulkarni: :+1:

Issue #5065, opened 2021-11-08
fix test_simultaneous_drain_of_two_ocs_nodes to run on VMWARE IPI


annagitel:
this test failes on vmware IPI runs in `nodes.terminate_nodes` because this method is built for AWS (terminating EC2). a suitable method needs to be created at the same file ocs_ci/ocs/platform_nodes.py under `class VSPHEREUPINode(VMWareNodes):` on order to pass.
and ofcourse the test itself should be updated to have a condition for the platform and therefore create the right node object

yitzhak12:
Just one note. I think the implementation of the function `terminate_nodes` should be under the class: `VMWareNodes`

yitzhak12:
Also, we have the same issue in this test https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/7393/testReport/tests.manage.z_cluster.nodes.test_automated_recovery_from_failed_nodes_reactive_IPI/TestAutomatedRecoveryFromFailedNodes/test_automated_recovery_from_failed_nodes_IPI_reactive_cephfs_terminate_/.

PR #5109, opened 2021-11-17
Create a new method 'terminate_nodes' for vSphere platform

yitzhak12:
Fix https://github.com/red-hat-storage/ocs-ci/issues/5065

### DIFF
--- a/ocs_ci/ocs/machine.py
+++ b/ocs_ci/ocs/machine.py
@@ -973,3 +973,50 @@ def add_annotation_to_machine(annotation, machine_name):
     command = f"annotate machine {machine_name} {annotation}"
     log.info(f"Adding annotation: {command} to machine {machine_name} ")
     ocp_obj.exec_oc_cmd(command)
+
+
+def set_replica_count(machine_set, count):
+    """
+    Change the replica count of a machine set.
+
+    Args:
+          machine_set (str): Name of the machine set
+          count (int): The number of the new replica count
+
+    Returns:
+        bool: True if the change was made successfully. False otherwise
+
+    """
+    ocp = OCP(namespace=constants.OPENSHIFT_MACHINE_API_NAMESPACE)
+    ocp.exec_oc_cmd(f"scale --replicas={count} machinesets {machine_set}")
+    return True
+
+
+def change_current_replica_count_to_ready_replica_count(machine_set):
+    """
+    Change the current replica count to be equal to the ready replica count
+    We may use this function after deleting a node or after adding a new node.
+
+    Args:
+        machine_set (str): Name of the machine set
+
+    Returns:
+        bool: True if the change was made successfully. False otherwise
+
+    """
+    res = True
+    current_replica_count = get_replica_count(machine_set)
+    ready_replica_count = get_ready_replica_count(machine_set)
+    log.info(
+        f"current replica count is: {current_replica_count}, "
+        f"ready replica count is: {ready_replica_count}"
+    )
+    if current_replica_count != ready_replica_count:
+        log.info(
+            "Change the current replica count to be equal to the ready replica count"
+        )
+        res = set_replica_count(machine_set, count=ready_replica_count)
+    else:
+        log.info("The current replica count is equal to the ready replica count")
+
+    return res
--- a/ocs_ci/ocs/platform_nodes.py
+++ b/ocs_ci/ocs/platform_nodes.py
@@ -76,12 +76,16 @@ class PlatformNodesFactory:
             "powervs": IBMPowerNodes,
             "rhv": RHVNodes,
             "ibm_cloud": IBMCloud,
+            "vsphere_ipi": VMWareIPINodes,
         }
 
     def get_nodes_platform(self):
         platform = config.ENV_DATA["platform"]
-        if cluster.is_lso_cluster() and platform == constants.VSPHERE_PLATFORM:
-            platform += "_lso"
+        if platform == constants.VSPHERE_PLATFORM:
+            if cluster.is_lso_cluster():
+                platform += "_lso"
+            elif config.ENV_DATA["deployment_type"] == "ipi":
+                platform += "_ipi"
         return self.cls_map[platform]()
 
 
@@ -221,37 +225,6 @@ class VMWareNodes(NodesBase):
         for node in node_names:
             node_vms = [vm for vm in vms_in_pool if vm.name in node]
             vms.extend(node_vms)
-        return vms
-
-    def get_vm_nodes_in_dc(self, nodes, dc=None):
-        """
-        Get vSphere vm objects list in the Datacenter(and not just in the cluster scope).
-        Note: If one of the nodes failed with an exception, it will not return his
-        corresponding VM object.
-
-        Args:
-            nodes (list): The OCS objects of the nodes
-            dc (str): The Datacenter name. If not specified, it will use the current dc.
-
-        Returns:
-            list: vSphere vm objects list in the Datacenter
-
-        """
-        dc = dc or self.datacenter
-        vms_in_dc = self.vsphere.get_all_vms_in_dc(dc)
-        node_names = set([node.get().get("metadata").get("name") for node in nodes])
-        vms = []
-        for vm in vms_in_dc:
-            try:
-                vm_name = vm.name
-                if vm_name in node_names:
-                    vms.append(vm)
-            except Exception as e:
-                logger.info(f"Failed to get the vm name due to exception: {e}")
-
-        if len(vms) < len(nodes):
-            logger.warning("Didn't find all the VM objects for all the nodes")
-
         return vms
 
     def get_data_volumes(self, pvs=None):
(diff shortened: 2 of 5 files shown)


ocs_ci/ocs/platform_nodes.py
```diff
@@ -220,6 +223,31 @@ def get_vms(self, nodes):
             vms.extend(node_vms)
         return vms
 
+    def get_vms_in_dc(self, nodes, dc=None):
```
ebenahar, line 226: The function name is similar to ocs_ci.utility.vsphere.VSPHERE.get_all_vms_in_dc.
I think its name should be OCP nodes oriented, maybe `get_vm_nodes_in_dc`

vavuthu (approved):

ocs_ci/ocs/platform_nodes.py
```diff
@@ -410,6 +438,19 @@ def create_and_attach_nodes_to_cluster(self, node_conf, node_type, num_nodes):
...
+        Terminate the VM's
+
+        Args:
+            nodes (list): The OCS objects of the nodes
+            wait (bool): True for waiting the instances to terminate,
+            False otherwise
+
+        """
+        vms = self.get_vm_nodes_in_dc(nodes)
+        self.vsphere.destroy_vms(vms)
```
PrasadDesala, line 452: will it also delete the attached data disk to the VM if any? or just remove the VM from inventory

yitzhak12: From what I know, it should delete the attached data disk to the VM.

prsurve (dismissed):

ebenahar (approved):

yitzhak12:
The pr validation job above https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/8374/consoleFull failed on the sanity check, but the test passed. It seems that we need to increase the timeout in the sanity check to wait for Ceph Health to be OK.

ocs_ci/ocs/platform_nodes.py
```diff
@@ -220,6 +223,31 @@ def get_vms(self, nodes):
...
+        """
+        vms_in_dc = self.vsphere.get_all_vms_in_dc(self.datacenter)
+        node_names = set([node.get().get("metadata").get("name") for node in nodes])
+        vms = []
+        for vm in vms_in_dc:
+            try:
+                vm_name = vm.name
+                if vm_name in node_names:
+                    vms.append(vm)
+            except Exception as e:
```
jilju, line 246: What is the expected exception here ? ```vms``` will be incomplete list of VMs in case of an Exception. Please add a validation before returning if ```len(nodes)``` should match ```len(vms)```

jilju (approved):

ocs_ci/ocs/platform_nodes.py
yitzhak12: There is a chance that one of the VM objects will not exist, so it will throw an exception when trying to get the VM name.

ocs_ci/ocs/platform_nodes.py
```diff
@@ -220,6 +223,31 @@ def get_vms(self, nodes):
             vms.extend(node_vms)
         return vms
 
+    def get_vm_nodes_in_dc(self, nodes, dc=None):
+        """
+        Get vSphere vm objects list in the Datacenter(and not just in the cluster scope)
+
+        Args:
+            nodes (list): The OCS objects of the nodes
+            dc (str): The Datacenter name. If not specified, it will use the current dc.
```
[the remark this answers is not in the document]
yitzhak12: Yes, you are right. I will fix it to use also the `dc` param in case we pass it to the function.

jilju (dismissed):

ocs_ci/ocs/platform_nodes.py
jilju: Resolved after a discussion.

PrasadDesala (dismissed):

ocs_ci/ocs/platform_nodes.py
```diff
@@ -410,6 +444,22 @@ def create_and_attach_nodes_to_cluster(self, node_conf, node_type, num_nodes):
...
+        Terminate the VMs
+
+        Args:
+            nodes (list): The OCS objects of the nodes
+            wait (bool): True for waiting the VMs to terminate,
+            False otherwise
+
+        """
+        vms = self.get_vm_nodes_in_dc(nodes)
+        self.vsphere.remove_vms_from_inventory(vms)
```
vavuthu, line 458: why do you want to remove all VM's in a DC?

ocs_ci/ocs/platform_nodes.py
```diff
@@ -220,6 +223,37 @@ def get_vms(self, nodes):
...
+                vm_name = vm.name
+                if vm_name in node_names:
+                    vms.append(vm)
+            except Exception as e:
+                logger.info(f"Failed to get the vm name due to exception: {e}")
+
+        if len(vms) < len(nodes):
+            logger.warning("Didn't find all the VM objects for all the nodes")
+
+        return vms
```
vavuthu, line 255: this function will get all the VM's instance of Datacenter with matching node names of cluster and then its calling in https://github.com/red-hat-storage/ocs-ci/pull/5109/files#diff-478e5b871f475f1e312384c0eeb216726f11da23e3d3ac4df80aa3ee5d92e441R458

So basically this will delete all the the nodes for all the clusters in Datacenter

ocs_ci/ocs/platform_nodes.py
vavuthu: Need to revisit this PR

vavuthu (changes requested):

ocs_ci/ocs/platform_nodes.py
vavuthu: If you looking for a VM instance of a particular cluster, you can use https://github.com/red-hat-storage/ocs-ci/blob/f2171eaf52a8fef11f3c523f3b427b7863cb5c9f/ocs_ci/ocs/platform_nodes.py#L203

yitzhak12: The function `get_vms` searches the node's objects in a specific pool, and not in all the Datacenter. And what we need right now with the IPI cluster is searching in all the Datacenter - that's why I created this new function.

yitzhak12: We will delete only the nodes specified in the `nodes` list.

vavuthu: > 
If it is for IPI type, can we have it separate class instead of in common class ( VMWareNodes  )

yitzhak12: Yes, I will create a new class for IPI.

prsurve (approved):

jilju (dismissed):

ocs_ci/ocs/platform_nodes.py
```diff
@@ -410,6 +417,22 @@ def create_and_attach_nodes_to_cluster(self, node_conf, node_type, num_nodes):
...
+            wait (bool): True for waiting the VMs to terminate,
+            False otherwise
+
+        """
+        vms = self.get_vms(nodes)
+        self.vsphere.remove_vms_from_inventory(vms)
+        if wait:
+            for vm in vms:
+                self.vsphere.wait_for_vm_delete(vm)
+
```
vavuthu, line 435: move function definition out of base class ( in this case VMWareNodes ). reason is if someone call terminate_nodes from inherited class ( UPI ), it gives unexpected results.

yitzhak12: Okay, I will do that.

ebenahar (approved):

ocs_ci/ocs/platform_nodes.py
ebenahar: Yes, this needs to be done

ebenahar: done here https://github.com/red-hat-storage/ocs-ci/pull/5109/files#diff-478e5b871f475f1e312384c0eeb216726f11da23e3d3ac4df80aa3ee5d92e441R186

ebenahar (approved):

jilju (approved):

ocs_ci/ocs/machine.py
```diff
@@ -973,3 +973,50 @@ def add_annotation_to_machine(annotation, machine_name):
...
+          machine_set (str): Name of the machine set
+          count (int): The number of the new replica count
+
+    Returns:
+        bool: True if the change was made successfully. False otherwise
+
+    """
+    ocp = OCP(namespace=constants.OPENSHIFT_MACHINE_API_NAMESPACE)
+    ocp.exec_oc_cmd(f"scale --replicas={count} machinesets {machine_set}")
+    return True
```
vavuthu, line 992: it always return True?

ebenahar (approved):

ocs_ci/ocs/machine.py
yitzhak12: TBH I took it from this function here https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/machine.py#L903. I think the return value is just an indication that the function was executed successfully. Maybe we can add `try` and `exception` to catch an error and return also the `False` value.

prsurve (approved):

PR #5013, opened 2021-10-27
Remove legacy reportportal

clacroix12:
Remove pytest-reportportal integration code now that we have moved to utilizing rp_preproc.

PR #5117, opened 2021-11-18
Stop adding rp_launch_tags to pytest.ini

clacroix12:


clacroix12:
`rp_launch_tags` is no longer necessary since we have moved away from the `pytest-reportportal` integration. This should have been removed as a part of the legacy cleanup in #5013 but was missed.

vavuthu (approved):

petr-balogh (approved):

PR #5122, opened 2021-11-21
Changing test_create_pool_block_pool.py to 4.9 and check compression efficiency in UI

shyRozen:
Update the testcase for 4.9 with different location for block_pool page, handling default binding mode for storageclass and checking compression efficiency from UI with prometheus values.

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -1182,6 +1182,25 @@ def validate_compression(pool_name):
     raise PoolNotFound(f"Pool {pool_name} not found on cluster")
 
 
+def get_pool_id(pool_name):
+    """
+    Get pool ID from pool name
+
+    Args:
+        pool_name (str): name of the pool to get the id.
+
+    Returns:
+        pool id if found else None
+
+    """
+    ceph_df_detail_output = get_ceph_df_detail()
+    pool_list = ceph_df_detail_output.get("pools")
+    for pool in pool_list:
+        if pool.get("name") == pool_name:
+            return pool.get("id")
+    return None
+
+
 def validate_osd_utilization(osd_used=80):
     """
     Validates osd utilization matches osd_used value
--- a/ocs_ci/ocs/ui/block_pool.py
+++ b/ocs_ci/ocs/ui/block_pool.py
@@ -6,8 +6,11 @@ from ocs_ci.ocs.ui.views import locators
 from ocs_ci.utility.utils import get_ocp_version
 from selenium.webdriver.common.by import By
 from ocs_ci.helpers.helpers import create_unique_resource_name
-from ocs_ci.ocs.exceptions import PoolStateIsUnknow
+from ocs_ci.ocs.exceptions import PoolStateIsUnknow, PoolIdNotFound
 import ocs_ci.ocs.resources.pod as pod
+from ocs_ci.utility.prometheus import PrometheusAPI
+from ocs_ci.ocs.cluster import get_pool_id
+from ocs_ci.utility.utils import convert_bytes_into_right_measure
 
 logger = logging.getLogger(__name__)
 
@@ -69,9 +72,9 @@ class BlockPoolUI(PageNavigator):
             bool: True if pool is in the list of pools page, otherwise False
 
         """
-        self.navigate_overview_page()
+        self.navigate_installed_operators_page()
         self.navigate_block_pool_page()
-        self.page_has_loaded()
+        self.page_has_loaded(retries=15)
         pool_existence = self.wait_until_expected_text_is_found(
             (f"a[data-test-operand-link={pool_name}]", By.CSS_SELECTOR), pool_name, 5
         )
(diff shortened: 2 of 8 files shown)


shyRozen:
Added the need testing although job passed it is for 4.9.
As the test is also for 4.8 to test it. There is already a job for it but it is waiting for psi
[validation for 4.8](https://ocs4-jenkins-csb-ocsqe.apps.ocp4.prod.psi.redhat.com/job/qe-trigger-test-pr/1862/)

shyRozen:
First validation is for 4.9. Second for 4.8 aborted because PSI retry. 
Third validation for 4.8 aws.

ocs_ci/ocs/cluster.py
```diff
@@ -1182,6 +1182,25 @@ def validate_compression(pool_name):
...
+    Returns:
+        pool id if found else None
+
+    """
+    ceph_df_detail_output = get_ceph_df_detail()
+    pool_list = ceph_df_detail_output.get("pools")
+    for pool in pool_list:
+        if pool.get("name") == pool_name:
+            return pool.get("id")
+    return None
```
ebenahar, line 1201: I think it isn't needed

ocs_ci/ocs/ui/block_pool.py
```diff
@@ -173,3 +176,130 @@ def reach_pool_limit(self, replica, compression):
...
+    ):
+        """
+        Get pool compression efficiency from UI and compare them against prometheus metrics
+
+        Args:
+            pool_name (str): The pool name to be checked.
+            expected_compression_saving (str): The expected compression saving in string like "1.5 GiB"
+
+        Returns:
+            (bool): True if UI and prometheus parameters are equal to expected_compression_saving - else False.
```
ebenahar, line 191: replace `(bool)` with `bool`

ocs_ci/ocs/ui/block_pool.py
```diff
@@ -173,3 +176,130 @@ def reach_pool_limit(self, replica, compression):
...
+            f" compression_saving {compression_saving}"
+        )
+
+        # Get pool id and query Prometheus
+        pool_id = get_pool_id(pool_name)
+        if pool_id is None:
+            raise PoolIdNotFound(f"Pool {pool_name} didn't return pool id")
+        new_prom = PrometheusAPI()
+        ceph_pool_compress_bytes_used_data = new_prom.query(
+            f'ceph_pool_compress_bytes_used{{pool_id="{pool_id}"}}'
```
ebenahar, line 228: please save the query in ocs_ci/ocs/constants.py:1227

ocs_ci/ocs/ui/block_pool.py
```diff
@@ -173,3 +176,130 @@ def reach_pool_limit(self, replica, compression):
...
+        )
+        ceph_pool_compress_under_bytes_data = new_prom.query(
+            f'ceph_pool_compress_under_bytes{{pool_id="{pool_id}"}}'
```
ebenahar, line 231: same

ocs_ci/ocs/ui/block_pool.py
```diff
@@ -173,3 +176,130 @@ def reach_pool_limit(self, replica, compression):
...
+            return_value_are_equal = False
+            logger.info(
+                f"UI compression saving are not equal to expected saving: "
+                f"UI compression value is {compression_saving_value} "
+                f"and expected compression value is {expected_compression_value}. "
+                f"UI compression unit is {compression_saving_unit} "
+                f"and prometheus compression unit is {expected_compression_unit}"
+            )
+
+        if (
```
ebenahar, line 278: shouldn't it be `elif`? do we want to have all these conditions checked? if the first sets the `return_value_are_equal` to False, isn't it enough?

ocs_ci/utility/utils.py
```diff
@@ -2653,6 +2654,26 @@ def convert_device_size(unformatted_size, units_to_covert_to):
     return conversion[units_to_covert_to][units]
 
 
+def convert_bytes_into_right_measure(size_bytes):
```
ebenahar, line 2657: isn't it similar to ocs_ci.utility.utils.convert_device_size?

ocs_ci/ocs/cluster.py
shyRozen: Correct

ocs_ci/ocs/ui/block_pool.py
shyRozen: I thought to get a wider picture on all value compare before returning False. It doesn't take time as data is not retrieved but just compared.

ocs_ci/utility/utils.py
shyRozen: In convert_device_size you have to decide the right measure unit like GI MI TI etc. In my function the measure unit is self calculate for the size. 
My gets only bytes which is what ceph gives in json and convert_device_size is getting any kind of measure. 
convert_bytes_into_right_measure is predicting the measure unit by the size of the input. Similar but different. 
If you don't know what units the outcome will be (GB TB PB etc) it's better to use mine.

ocs_ci/ocs/ui/block_pool.py
shyRozen: I saw the queries in constant. None of them have variables inside them and mine have pool id.
Whomever will use them will have to know that he needs to format the string to insert the variable. Are you ok with that?

ocs_ci/utility/utils.py
ebenahar: ok

ocs_ci/ocs/ui/block_pool.py
ebenahar: yes, it is ok

Issue #4980, opened 2021-10-19
Running tests over MCG-only clusters is not supported in OCS-CI


MeridianExplorer:
At the moment, when OCS-CI is run on a cluster installed in an "MCG-Only" configuration, it fails in several parts, namely - the installation of the Ceph toolbox and the RGW pod check under `mcg.py`.

MeridianExplorer:
Reopening the issue; #5040 does not address either of the two points of failure mentioned above, and any OCS-CI test runs still fail on MCG-only clusters.

PR #5144, opened 2021-11-26
Enables to run test cases on MCG only deployment

vavuthu:
Fixes: #4980

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -61,8 +61,9 @@ class CephCluster(object):
         Cluster object initializer, this object needs to be initialized
         after cluster deployment. However its harmless to do anywhere.
         """
+        if config.ENV_DATA["mcg_only_deployment"]:
+            return
         # cluster_name is name of cluster in rook of type CephCluster
-
         self.POD = ocp.OCP(kind="Pod", namespace=config.ENV_DATA["cluster_namespace"])
         self.CEPHCLUSTER = ocp.OCP(
             kind="CephCluster", namespace=config.ENV_DATA["cluster_namespace"]
--- a/ocs_ci/ocs/resources/mcg.py
+++ b/ocs_ci/ocs/resources/mcg.py
@@ -183,7 +183,10 @@ class MCG:
                     not pods
                 ), "RGW pods should not exist in the current platform/cluster"
 
-        elif config.ENV_DATA.get("platform") in constants.ON_PREM_PLATFORMS:
+        elif (
+            config.ENV_DATA.get("platform") in constants.ON_PREM_PLATFORMS
+            and not config.ENV_DATA["mcg_only_deployment"]
+        ):
             rgw_count = get_rgw_count(
                 config.ENV_DATA["ocs_version"], check_if_cluster_was_upgraded(), None
             )
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -369,13 +369,14 @@ def gather_version_info_for_report(config):
         config._metadata["Cluster Version"] = clusterversion
 
         # add ceph version
-        ceph_version = get_ceph_version()
-        config._metadata["Ceph Version"] = ceph_version
+        if not ocsci_config.ENV_DATA["mcg_only_deployment"]:
+            ceph_version = get_ceph_version()
+            config._metadata["Ceph Version"] = ceph_version
 
-        # add csi versions
-        csi_versions = get_csi_versions()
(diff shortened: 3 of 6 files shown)


vavuthu:
Theres no blockage for running test cases in MCG only mode. verified acceptance with normal deployment and sample test case in MCG only mode.

Job are failed due to test cases. Hence marking this PR as verified.

vavuthu:
Normal deployment- acceptance
```
Summary

28 tests ran in 4986.18 seconds.
27 passed, 0 skipped, 1 failed, 1 errors, 0 expe cted failures, 0 unexpected passes
Squad Analysis - please analyze:
Failures:
Red squad

    tests/manage/mcg/test_write_to_bucket.py::TestBucketIO::test_write_file_to_bucket[DEFAULT-BACKINGSTORE]
```

sample test case in MCG Only deployment
```
4 tests ran in 3324.54 seconds.
2 passed, 0 skipped, 2 failed, 1 errors, 0 expec ted failures, 0 unexpected passes
Squad Analysis - please analyze:
Failures:
Red squad

    tests/manage/mcg/test_host_node_failure.py::TestNoobaaSTSHostNodeFailure::test_noobaa_sts_host_node_failure[noobaa-core-False]
    tests/manage/mcg/test_host_node_failure.py::TestNoobaaSTSHostNodeFailure::test_noobaa_sts_host_node_failure[noobaa-db-pg-False]
```

ocs_ci/ocs/cluster.py
```diff
@@ -61,8 +61,9 @@ def __init__(self):
         Cluster object initializer, this object needs to be initialized
         after cluster deployment. However its harmless to do anywhere.
         """
+        if config.ENV_DATA["mcg_only_deployment"]:
+            return
```
bernerhat, line 65: tests dont validate the returned cluster object, i would suggest maybe instead of returning here we could skip the tests that init a cluster obj from within the instanciation if possible. wdyt?

ocs_ci/ocs/resources/mcg.py
```diff
@@ -183,7 +183,10 @@ def __init__(self, *args, **kwargs):
                     not pods
                 ), "RGW pods should not exist in the current platform/cluster"
 
-        elif config.ENV_DATA.get("platform") in constants.ON_PREM_PLATFORMS:
+        elif (
+            config.ENV_DATA.get("platform") in constants.ON_PREM_PLATFORMS
+            and not config.ENV_DATA["mcg_only_deployment"]
+        ):
```
bernerhat, line 189: should be checked in if lines 171-174 same as in [here](https://github.com/red-hat-storage/ocs-ci/pull/5120/files#diff-1d32a061b0f98ee283c52d1f7e9ca822177982cee47f7daa76904f5d8e1c3bfdv)

petr-balogh (approved):

ocs_ci/ocs/resources/mcg.py
vavuthu: we can't use lines 171-174 same as in here, since its not TC

ocs_ci/ocs/cluster.py
vavuthu: its called by sanity class but not TC. TC is valid but we need to skip cephcluster part.

ocs_ci/ocs/resources/mcg.py
bernerhat: in the if statement above there is a check to verify that there are no rgw related pods in the deployment (mcg-only shouldnt have any) and the addition to this if statement skips the whole check

vavuthu: @bernerhat I didn't see any addition checks in this block . Only rgw verification is there in this block which we are skipping for MCG Only mode

MeridianExplorer: Why would lines 171-174 be limited to a test case?
The current flow of the PR will mean we're not verifying that RGW pods are missing in MCG-only (which certainly doesn't hurt).
If we'll change the if in these lines, it'll mean we'll *always* verify RGW pods are missing in MCG-only deployments -
```
if (
  config.ENV_DATA["platform"].lower() in constants.CLOUD_PLATFORMS
   or storagecluster_independent_check()
   or config.ENV_DATA["mcg_only_deployment"]
):
```
It's essentially better coverage for free

vavuthu: @MeridianExplorer I think we can RGW pods missing checks in post install verification of MCG only deployment instead of here, so that we have all checks at one place. I will take it up in another PR.

Akarsha-rai (approved):
LGTM

PR #5147, opened 2021-11-27
Skip `test_scale_rgw_down` on external mode clusters

MeridianExplorer:
The test is being skipped due to OCS-CI having no access to the RGW pods when testing external mode clusters


MeridianExplorer:
@sidhant-agrawal - What do you think about this PR? Is it perhaps possible to add it as part of your existing tier4-over-external-mode PR?

sidhant-agrawal:
Sorry, I missed the ping.
I agree, the test should be skipped since there is no RGW pod running on ODF side.

The existing PR is for enhancing `disruption_helpers.py` for external mode. This test case doesn't use disruption_helpers so this cannot be part of existing PR & needs to be taken as a follow-up task.

ebenahar (approved):

ebenahar (approved):

parikshithb (approved):

MeridianExplorer:
I have no access to an external-mode cluster, thus merging without verification

PR #5161, opened 2021-12-01
[Close loop BZ: 1998680]: Object bucket with large number of objects

parikshithb:
Signed-off-by: Parikshith Byregowda <pbyregow@redhat.com>

### DIFF
--- a/tests/e2e/scale/noobaa/test_hsbench.py
+++ b/tests/e2e/scale/noobaa/test_hsbench.py
@@ -89,7 +89,7 @@ class TestHsBench(E2ETest):
         # Create an Object bucket
         object_bucket = bucket_factory(amount=1, interface="OC")[0]
 
-        # Write 1M objects to the object bucket
+        # Write 700k objects to the object bucket
         s3bench.run_benchmark(
             num_obj=700000,
             timeout=12000,
--- a/tests/e2e/scale/noobaa/test_hsbench.py
+++ b/tests/e2e/scale/noobaa/test_hsbench.py
@@ -3,7 +3,11 @@ import pytest
 from ocs_ci.ocs import hsbench
 from ocs_ci.utility import utils
 from ocs_ci.framework.testlib import E2ETest, scale
-from ocs_ci.framework.pytest_customization.marks import vsphere_platform_required, bugzilla, skipif_ocs_version
+from ocs_ci.framework.pytest_customization.marks import (
+    vsphere_platform_required,
+    bugzilla,
+    skipif_ocs_version,
+)
 
 log = logging.getLogger(__name__)
 
@@ -19,6 +23,20 @@ def hsbenchs3(request):
 
     request.addfinalizer(teardown)
     return hsbenchs3
+
+
+@pytest.fixture(scope="function")
+def s3bench(request):
+
+    s3bench = hsbench.HsBench()
+    s3bench.create_resource_hsbench()
+    s3bench.install_hsbench()
+
+    def teardown():
+        s3bench.cleanup()
+
+    request.addfinalizer(teardown)
+    return s3bench
 
 
 @scale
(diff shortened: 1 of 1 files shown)


tests/e2e/scale/noobaa/test_hsbench.py
```diff
@@ -57,3 +75,28 @@ def test_s3_benchmark_hsbench(self, hsbenchs3):
...
+        """
+        Tests large number of objects on a single bucket:
+        * Create an OBC
+        * Run hs S3 benchmark to create 700k objects on the object bucket
+        * Post writing objects verify OBC creation
+        """
+        # Create an Object bucket
+        object_bucket = bucket_factory(amount=1, interface="OC")[0]
+
+        # Write 1M objects to the object bucket
```
tiffanyn108, line 92: This test is to write 70000 objects, not 1M.

tests/e2e/scale/noobaa/test_hsbench.py
```diff
@@ -3,7 +3,11 @@
 from ocs_ci.ocs import hsbench
 from ocs_ci.utility import utils
 from ocs_ci.framework.testlib import E2ETest, scale
-from ocs_ci.framework.pytest_customization.marks import vsphere_platform_required
+from ocs_ci.framework.pytest_customization.marks import (
+    vsphere_platform_required,
```
tiffanyn108, line 7: This test also can be run on both aws and VMware clusters.

tests/e2e/scale/noobaa/test_hsbench.py
```diff
@@ -57,3 +75,28 @@ def test_s3_benchmark_hsbench(self, hsbenchs3):
...
+        object_bucket = bucket_factory(amount=1, interface="OC")[0]
+
+        # Write 1M objects to the object bucket
+        s3bench.run_benchmark(
+            num_obj=700000,
+            timeout=12000,
+            access_key=mcg_obj.access_key_id,
+            secret_key=mcg_obj.access_key,
+            end_point=f"http://s3.openshift-storage.svc/{object_bucket.name}",
+        )
```
Akarsha-rai, line 99: is it possible to write such that few are small and few are larger objects in 70000/1M objects?

tests/e2e/scale/noobaa/test_hsbench.py
parikshithb: ack

tests/e2e/scale/noobaa/test_hsbench.py
parikshithb: hs_benchmark is set with default values when initialized, if we want, need to modify the class.
Also for this particular test we don't need multiple ob sizes, we are only focused on object count.

Akarsha-rai: ack, thanks

tests/e2e/scale/noobaa/test_hsbench.py
```diff
@@ -57,3 +75,28 @@ def test_s3_benchmark_hsbench(self, hsbenchs3):
...
+
+        # Create new OBC and verify it is bound
+        bucket_factory(interface="OC")
```
hnallurv, line 102: How new OBCs get created here?

The https://bugzilla.redhat.com/show_bug.cgi?id=1998680#c38 mentions you created 100 OBCs.

tests/e2e/scale/noobaa/test_hsbench.py
parikshithb: True, I only added skip for existing RGW test for platforms other than vsphere. Skip was at class level.

tests/e2e/scale/noobaa/test_hsbench.py
parikshithb: Does not matter, OBC should be bound instead of going to LOST or PENDING state.

hnallurv (approved):

PrasadDesala (approved):

tiffanyn108 (approved):

PR #5164, opened 2021-12-01
Add ocs_build attribute to RP launches

clacroix12:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -87,6 +87,7 @@ from ocs_ci.utility.uninstall_openshift_logging import uninstall_cluster_logging
 from ocs_ci.utility.utils import (
     ceph_health_check,
     ceph_health_check_base,
+    get_ocs_build_number,
     get_running_ocp_version,
     get_openshift_client,
     get_system_architecture,
@@ -1223,7 +1224,7 @@ def health_checker(request, tier_marks_name):
 
 
 @pytest.fixture(scope="session", autouse=True)
-def cluster(request, log_cli_level):
+def cluster(request, log_cli_level, record_testsuite_property):
     """
     This fixture initiates deployment for both OCP and OCS clusters.
     Specific platform deployment classes will handle the fine details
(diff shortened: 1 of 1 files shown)


ocs_ci/utility/reporting.py
```diff
@@ -48,6 +48,7 @@ def get_rp_launch_attributes():
     rp_attrs["worker_instance_type"] = config.ENV_DATA.get("worker_instance_type")
     rp_attrs["ocp_version"] = get_ocp_version()
     rp_attrs["ocs_version"] = config.ENV_DATA.get("ocs_version")
+    rp_attrs["ocs_build"] = get_ocs_build_number()
```
petr-balogh, line 51: Is this triggered at the start of the execution? Or once we have ocs running?

Mainly asking about deployment scenario.

As I see you call get client function, just not sure why we don't have client yet, maybe in deployment start, hence asking.

petr-balogh (dismissed):

ocs_ci/utility/reporting.py
clacroix12: Yeah this is the main issue with this approach. I originally hit the issue where we didn't have the client so i added this call to ensure we do. However the real problem is that this entire process is called from within an autouse fixture prior to any deployment being done, so we won't be able to do things like [pull CSV data](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L1500) anyways since it doesn't exist. 

In order to utilize `get_ocs_build_number` we will need to introduce some other place (post deployment) where we can call this. However, I haven't tested whether we can update the testsuite custom properties at that time.

vavuthu (dismissed):

petr-balogh (dismissed):

petr-balogh (approved):

clacroix12:
I'm seeing an issue where we are getting `None` back from rp_preproc which breaks the redirect link to ReportPortal. I'm not sure how these changes would cause this to happen so I'm not 100% sure it's related to this however it has been happening while trying to verify these changes so it is holding things up. Just wanted to update here.

shylesh (approved):
+1

clacroix12:
Verified: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/8273/

Launch: https://reportportal-ocs4.apps.ocp4.prod.psi.redhat.com/ui/#ocs/launches/all/464

You can see the attribute on the launch: `ocs_build:4.8.5-2`

Issue #4974, opened 2021-10-18
Handle BZ 2006865 alerting changes: increase fire timeout


mbukatov:
Delay timing of following alerts were updated to 15m:

- CephMonHighNumberOfLeaderChanges
- CephOSDDiskNotResponding
- CephClusterWarningState

Negative test cases which disables mon or osd pod needs to be updated accordingly. This means to increase the timeout in a workload fixtures these test cases are using.

PR #5169, opened 2021-12-02
Reflect changes in Prometheus alert times

fbalak:
- Fixes https://github.com/red-hat-storage/ocs-ci/issues/4974


mbukatov (commented):
Are you sure about the run time updates? I see the verification still fails, and I don't immediately see connection between changes here and https://github.com/rook/rook/pull/8896/files

fbalak:
> Are you sure about the run time updates? I see the verification still fails, and I don't immediately see connection between changes here and https://github.com/rook/rook/pull/8896/files

You are right, the issue was that new monitor is recreated before the warning alert is turned from Pending state into Firing state.

mbukatov (approved):

petr-balogh (approved):

Issue #5186, opened 2021-12-06
 External cluster deployment faling with KeyError: 'status' for:  get_default_bs["status"]["phase"]


petr-balogh:
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/2446/consoleFull
http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/j-042vu1ce33-t1/j-042vu1ce33-t1_20211203T205636/logs/ocs-ci-logs-1638565283/by_outcome/failed/tests/ecosystem/deployment/test_deployment.py/test_deployment/logs

Here we see that backing store didn't have state from some reason and the code is failing with traceback linked below.

```
2021-12-03 21:29:45,615 - MainThread - INFO - ocs_ci.utility.utils.exec_cmd.487 - Executing command: oc -n openshift-storage get backingstore noobaa-default-backing-store -n openshift-storage -o yaml
2021-12-03 21:29:45,795 - MainThread - DEBUG - ocs_ci.utility.utils.exec_cmd.500 - Command stdout: apiVersion: noobaa.io/v1alpha1
kind: BackingStore
metadata:
  annotations:
    rgw: ""
  creationTimestamp: "2021-12-03T21:29:45Z"
  finalizers:
  - noobaa.io/finalizer
  generation: 1
  labels:
    app: noobaa
  name: noobaa-default-backing-store
  namespace: openshift-storage
  ownerReferences:
  - apiVersion: noobaa.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: NooBaa
    name: noobaa
    uid: 8606bb02-4242-487d-b6de-04bd9b903beb
  resourceVersion: "31030"
  uid: b717a056-086d-4db0-97f5-30056a76e20c
spec:
  s3Compatible:
    endpoint: http://rook-ceph-rgw-ocs-external-storagecluster-cephobjectstore.openshift-storage.svc:8080
    secret:
      name: rook-ceph-object-user-ocs-external-storagecluster-cephobjectstore-noobaa-ceph-objectstore-user
      namespace: openshift-storage
    signatureVersion: v4
    targetBucket: nb.1638566985095.apps.j-042vu1ce33-t1.qe.rh-ocs.com
  type: s3-compatible
  ```
stack trace:
```
2021-12-03 22:40:20          if deploy:
[...]

PR #5187, opened 2021-12-06
Fix for MCG verification deployment on external cluster

petr-balogh:
It seems that the MCG deployment is not ready at the time we check it
in the case of external deployment.

See BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2014026
For more details.

To adding re-try also for KeyError when OCP doesn't return state and
phase of the object when it's in early stages of deployment we need to
do re-try to get those data on other attempt instead of fail the whole
deployment. As those object exists few moments later then we are running
verification.

Fixes: #5186


petr-balogh:
Verification:
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/2458/

dahorak (approved):

jimdevops19 (approved):

PR #4966, opened 2021-10-17
Implement tests for NooBaa component rescheduling improvements

sidhant-agrawal:
- Add new tests in mcg/test_host_node_failure.py
 - Reduce timeout from 300 seconds to 90 seconds in existing mcg/test_mcg_resources_disruptions.py

Signed-off-by: Sidhant Agrawal <sagrawal@redhat.com>

PR #5196, opened 2021-12-07
Extend the MCG DB deletion recovery timeout on external clusters

MeridianExplorer:
We're currently seeing failures after waiting 90 seconds, but the pod seems to reach a Running state after 660~ seconds

### DIFF
--- a/tests/manage/mcg/test_mcg_resources_disruptions.py
+++ b/tests/manage/mcg/test_mcg_resources_disruptions.py
@@ -75,7 +75,8 @@ class TestMCGResourcesDisruptions(MCGTest):
             condition=constants.STATUS_RUNNING,
             selector=self.labels_map[resource_to_delete],
             resource_count=1,
-            timeout=90,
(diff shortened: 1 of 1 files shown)


parikshithb (approved):

tests/manage/mcg/test_mcg_resources_disruptions.py
```diff
@@ -75,7 +75,8 @@ def test_delete_noobaa_resources(self, resource_to_delete):
             condition=constants.STATUS_RUNNING,
             selector=self.labels_map[resource_to_delete],
             resource_count=1,
-            timeout=90,
+            timeout=800 if config.DEPLOYMENT.get("external_mode") else 90,
```
sidhant-agrawal, line 78: @MeridianExplorer Increasing the timeout , is this a temporary workaround due to a bug or are we going to keep this forever?
Previously the timeout was 300 which was reduced to 90 (via #4966 ), but now we are ging 800 which is even more than that of previous value.

sidhant-agrawal (approved):

PR #5200, opened 2021-12-07
Extend ceph object

Avilir:
This PR Add to the ceph object the functionality of : 

- Creating new pools (RBD / CephFS)
- Deleting pools (RBD / CephFS), while insuring that the defaults polls will not be deleted (accidently)
- Changing the PG number of a pool
- Changing the target_size_ratio of a pool

All of that is for using in heavy loaded test (filling up the cluster), new pools that can be deleted in the end, 
so the actual cleanup will be very fast.

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -765,6 +765,32 @@ class CephCluster(object):
         except Exception as ex:
             logger.error(f"Failed to change the ratio : {ex}")
 
+    def get_cephfilesystem_status(self, fsname=None):
+        res = self.CEPHFS.get(resource_name=fsname)
+        return res.get("status").get("phase") == constants.STATUS_READY
+
+    def create_new_filesystem(self, fs_name):
+        """
+        Creating new filesystem to use in the tests insted of the default one.
+        the new filesystem is identical (parameters wise) to the default filesystem
+
+        Args:
+            fs_name (str):  The name of the filesystem to create
+
+        """
+        # Creating the new filesystem using the default parameters
+        self.cephfs.data["metadata"]["name"] = fs_name
+        self.cephfs.apply(**self.cephfs.data)
+
+        # Verify that the filesystem was created and the cluster if healthy
+        sample = TimeoutSampler(
+            timeout=120, sleep=3, func=self.get_cephfilesystem_status, fsname=fs_name
+        )
+        if not sample.wait_for_func_status(result=True):
+            err_msg = "Can not create new filesystem"
+            logger.error(err_msg)
+            raise exceptions.CephHealthException(err_msg)
+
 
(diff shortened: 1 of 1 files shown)


Avilir:
Currently no test is using this functionality, so, no PR validation ran.
following PR will implement (and test) this functionality in the performance IO tests (FIO, FmallFiles, etc.)

ypersky1980 (dismissed):
LGTM, great work!

ocs_ci/ocs/cluster.py
```diff
@@ -82,6 +86,13 @@ def __init__(self):
             logging.warning("No CephFS found")
             self.cephfs_config = None
 
+        try:
+            self.rbd_config = self.RBD.get().get("items")[0]
+        except IndexError as e:
+            logging.warning(e)
+            logging.warning("No RBD found")
```
Avilir, line 93: @vasukulkarni we might hit this exception in case the cluster was deployed without RBD, and in that case we will not run tests against RBD, so it will not affect the creation of new RBD pool later on.

ocs_ci/ocs/cluster.py
```diff
@@ -730,6 +746,142 @@ def time_taken_to_complete_rebalance(self, timeout=600):
...
+
+    def delete_filesystem(self, fs_name):
+        """
+        Delete a ceph filesystem - not the default one - from the cluster
+
+        Args:
+            fs_name (str): the name of the filesystem to delete
+        """
+        # Make sure the the default filesystem is not deleted.
+        if fs_name == "ocs-storagecluster-cephfilesystem":
```
keemano, line 827: Regardless of pass/fail of "if" condition, line num 806, 807 will be executed. This could be changed accordingly.

keemano (changes requested):

ocs_ci/ocs/cluster.py
Avilir: Good catch @keemano , i replaced it with return

ocs_ci/ocs/cluster.py
```diff
@@ -730,6 +746,142 @@ def time_taken_to_complete_rebalance(self, timeout=600):
...
+
+    def delete_blockpool(self, pool_name):
+        """
+        Delete a ceph RBD pool - not the default one - from the cluster
+
+        Args:
+            pool_name (str): the name of the RBD pool to delete
+        """
+        # Make sure the the default RBD pool is not deleted.
+        if pool_name == "ocs-storagecluster-cephblockpool":
```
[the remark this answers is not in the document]
Avilir: same as above, replaced with `return`

ramkiperiy (approved):

vasukulkarni (approved):

PR #5201, opened 2021-12-07
workaround for MS Upgrade

suchita-g:

### DIFF
--- a/conf/upgrade/upgrade_3m_3w.yaml
+++ b/conf/upgrade/upgrade_3m_3w.yaml
@@ -3,4 +3,5 @@ REPORTING:
   polarion:
     upgrade_id: 'OCS-1484'
 UPGRADE:
+  skip_upgrade: true
   upgrade: true
--- a/tests/manage/z_cluster/conftest.py
+++ b/tests/manage/z_cluster/conftest.py
@@ -3,6 +3,8 @@
 import logging
 import pytest
 
+from ocs_ci.framework import config
+from ocs_ci.ocs import constants
 
 from ocs_ci.ocs.fiojob import workload_fio_storageutilization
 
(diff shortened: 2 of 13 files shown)


tests/manage/z_cluster/conftest.py
```diff
@@ -94,3 +96,38 @@ def workload_storageutilization_cephfs(
...
+    if config.ENV_DATA["platform"].lower() in constants.MANAGED_SERVICE_PLATFORMS:
+        for item in items.copy():
+            for testname in skip_till_node_implement:
+                if testname in str(item.fspath):
+                    logger.info(
+                        f"Test {item} is removed from the collected items"
+                        f" till node implentation is in place"
+                    )
+                    items.remove(item)
+                    break
```
petr-balogh, line 133: This piece of code repeats in different files.

One issue I see is that the info message talking about `till node implementation is in place` But I guess it should mention something about MS instead?

You can remove it as function which will take to params: `items` and `list_of_items_so_remove` .
Then you can just call this same function from multiple places. It can be even more generic to have also part of message as argument of that function. So it can be used no only for MS use case but we can update message based on parameter.

conf/upgrade/upgrade_3m_3w.yaml
```diff
@@ -3,4 +3,5 @@ REPORTING:
   polarion:
     upgrade_id: 'OCS-1484'
 UPGRADE:
+  skip_upgrade: true
```
petr-balogh, line 6: How and where is this skip_upgrade used?

As this one is gneric file we use in our production runs, I probably don't want to have this changed here.

The main purpose of this file is for have proper polarion ID from production jobs. So please remove it from here.

PR #5220, opened 2021-12-13
Add SendGrid capacity test

fbalak:


tests/manage/monitoring/conftest.py
```diff
@@ -415,6 +415,29 @@ def workload_storageutilization_95p_rbd(
...
+    fixture_name = "workload_storageutilization_99p_rbd"
+    measured_op = workload_fio_storageutilization(
+        fixture_name,
+        project,
+        fio_pvc_dict,
+        fio_job_dict,
+        fio_configmap_dict,
+        measurement_dir,
+        tmp_path,
+        target_percentage=0.99,
```
mbukatov, line 436: Wouldn't that basically kill the cluster? :)

tests/manage/monitoring/sendgrid/test_capacity.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Test that there are appropriate alert emails when ceph cluster is utilized
+    via RBD interface.
+
+    """
+    log.warning(
+        "Cluster utilization is completed. "
+        f"There should be proper emails sent to {notification_emails_required}"
+    )
+    # TODO(fbalak): automate checking of email content
```
mbukatov, line 25: Ah, I see, this is WIP. Could you flag the PR as a draft?

tests/manage/monitoring/conftest.py
fbalak: Right, existing fixture that utilizes 95 percent is used instead.

tests/manage/monitoring/sendgrid/test_capacity.py
fbalak: This won't be automated as part of this PR.

fbalak:
Verification run: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/10331/console

mbukatov (approved):

yitzhak12:
It looks good. Maybe we can add another test for checking that it's not sending email notifications if the cluster is less than the expected capacity(if for example, it should send after 90% capacity we can add another test with the fixture `workload_storageutilization_85p_rbd`  and check it's not sending email notifications)

fbalak:
> It looks good. Maybe we can add another test for checking that it's not sending email notifications if the cluster is less than the expected capacity(if for example, it should send after 90% capacity we can add another test with the fixture `workload_storageutilization_85p_rbd` and check it's not sending email notifications)

I don't want to put it into this PR. Checking of email is done semiautomatically so we can check timestamps when we are validating alerts and also utilization of cluster takes a lot of time (it can be hours). When we have automation for checking mailbox, we can check for this in one test run just before the capacity is utilized (or later by timestamps) to make sure that the mail was not sent early.

ebenahar (approved):

PR #5244, opened 2021-12-17
<Closed loop BZ: 1944513> CephMonQuorumLost Alert raised when mon quorum lost

Akarsha-rai:
Automates bug: https://bugzilla.redhat.com/show_bug.cgi?id=1944513

Signed-off-by: Akarsha-rai <akrai@redhat.com>


tests/manage/monitoring/conftest.py
```diff
@@ -92,7 +96,7 @@ def measure_stop_ceph_mon(measurement_dir):
     mons = [deployment["metadata"]["name"] for deployment in mon_deployments]
 
     # get monitor deployments to stop, leave even number of monitors
-    split_index = len(mons) // 2 if len(mons) > 3 else 2
+    split_index = split_index if split_index else len(mons) // 2 if len(mons) > 3 else 2
```
mbukatov, line 99: Could we write this in more readable way instead of creating bit confusing function fixture overloading a variable name?

tests/manage/monitoring/conftest.py
```diff
@@ -76,11 +76,15 @@ def stop_mgr():
...
+def split_index(split_index=None):
+    return split_index
+
+
+@pytest.fixture
+def measure_stop_ceph_mon(measurement_dir, split_index):
     """
     Downscales Ceph Monitor deployment, measures the time when it was
     downscaled and monitors alerts that were triggered during this event.
-
```
mbukatov, line 83: No need to drop this newline.

tests/manage/monitoring/conftest.py
```diff
@@ -123,7 +127,9 @@ def stop_mon():
         time.sleep(run_time)
         return mons_to_stop
 
-    test_file = os.path.join(measurement_dir, "measure_stop_ceph_mon.json")
+    test_file = os.path.join(
+        measurement_dir, f"measure_stop_ceph_mon_{split_index}.json"
+    )
```
mbukatov, line 132: Why do we need to have a split index in the filename?

tests/manage/monitoring/conftest.py
```diff
@@ -140,8 +146,9 @@ def stop_mon():
         for mon in mons_to_stop:
             logger.info(f"Upscaling deployment {mon} back to 1")
             oc.exec_oc_cmd(f"scale --replicas=1 deployment/{mon}")
-        msg = f"Downscaled monitors {mons_to_stop} were not replaced"
-        assert check_old_mons_deleted, msg
+        if not split_index == 1:
```
mbukatov, line 149: When would split_index be equal to 1?

mbukatov (changes requested):
Test code looks good, but I'm bit puzzled by the changes in the conftest code.

tests/manage/monitoring/conftest.py
Akarsha-rai: I do want to use the same function when mon quorum is lost/down, i.e 2 mons down out of 3. 
So passing specifically `spilt_index` as 1 here but not able to set `split_index` as a parameter so created fixture to override a value

tests/manage/monitoring/conftest.py
Akarsha-rai: Especially when we need mon quorum to be lost, i.e 2 mon down out of 3 mons.
The operator cannot do anything with ceph or starting a new mon since the quorum is down so there is no chance of replacing old mons. Hence skipping it

tests/manage/monitoring/conftest.py
Akarsha-rai: ack

tests/manage/monitoring/conftest.py
mbukatov: Ah, I see. This makes sense, we just need to make the intentions more clear.

Do I understand it right that we need separate `split_index` fixture for the parametrization to work? In other words, would it work if we set `split_index=None` in definition of `measure_stop_ceph_mon`?

Overall, I would suggest to:

- parametrize the `measure_stop_ceph_mon` fixture via number of mon to go down (instead of split index, which is an internal implementation detail)
- calculate split index based on the above
- if we really need separate `split_index` fixture, it's name would rather be something like `stop_ceph_mon_num`, with some brief explanation in a docstring

Akarsha-rai: ack, created `create_mon_quorum_loss` fixture and if it passed as true `split_index` will be set to 1 so that only one mon will be up and running otherwise leave even number of monitors to be up and running.

fbalak (commented):
LGTM but errors in verification runs should be addressed.

Akarsha-rai:
With unchecked `USE_OCSQE_INGRESS_CERT`, 3/3 times test passed so marking it has verified.

fbalak (approved):
Errors related to USE_OCSQE_INGRESS_CERT could be handled in a different PR.

PrasadDesala (approved):

mbukatov (approved):

PR #5254, opened 2021-12-20
Test MG on Disconnected environment

OdedViner:


ocs_ci/ocs/must_gather/must_gather.py
```diff
@@ -134,6 +137,32 @@ def compare_running_pods(self):
...
+            ),
+        )
+        ocs_must_gather_image = ocsci_config.REPORTING["ocs_must_gather_image"]
+        ocs_must_gather_image_and_tag = f"{ocs_must_gather_image}:{latest_tag}"
+        if ocsci_config.DEPLOYMENT.get("disconnected"):
+            ocs_must_gather_image_and_tag = mirror_image(ocs_must_gather_image_and_tag)
+        cmd = f"adm must-gather --image={ocs_must_gather_image_and_tag} --dest-dir={self.root}"
+        occli = OCP()
+        output = occli.exec_oc_cmd(command=cmd, out_yaml_format=False, timeout=900)
+        if "cannot stat 'jq'" in output:
```
hnallurv, line 161: Can't this be checked in the existing test_must_gather test?

OdedViner: @hnallurv done

hnallurv (approved):

ocs_ci/ocs/must_gather/must_gather.py
```diff
@@ -44,10 +46,29 @@ def collect_must_gather(self):
         """
         Collect ocs_must_gather and copy the logs to a temporary folder.
 
+        Verify the output of mg command does not contain "cannot stat 'jq'"
+        https://bugzilla.redhat.com/show_bug.cgi?id=1974959
```
PrasadDesala, line 50: I think we dont need to add these details here in the function doc string, maybe we can add these as log.info and comments inside the function

ocs_ci/ocs/must_gather/must_gather.py
```diff
@@ -44,10 +46,29 @@ def collect_must_gather(self):
...
+            ocsci_config.REPORTING.get(
+                "default_ocs_must_gather_latest_tag",
+                ocsci_config.DEPLOYMENT["default_latest_tag"],
+            ),
+        )
+        ocs_must_gather_image = ocsci_config.REPORTING["ocs_must_gather_image"]
+        ocs_must_gather_image_and_tag = f"{ocs_must_gather_image}:{latest_tag}"
+        if ocsci_config.DEPLOYMENT.get("disconnected"):
+            ocs_must_gather_image_and_tag = mirror_image(ocs_must_gather_image_and_tag)
+        cmd = f"adm must-gather --image={ocs_must_gather_image_and_tag} --dest-dir={self.root}"
```
PrasadDesala, line 65: collect_ocs_logs() handles all these code, can you please let know the reason why the collect_ocs_logs is removed here?

OdedViner: I want to verify that "cannot stat 'jq'" string is not included in the output of mg cmd

ocs_ci/ocs/must_gather/must_gather.py
OdedViner: @PrasadDesala with bz link?

ocs_ci/ocs/must_gather/must_gather.py
OdedViner: @petr-balogh Can I add this code to `collect_ocs_logs` function?
```
        output = occli.exec_oc_cmd(command=cmd, out_yaml_format=False, timeout=1500)
        if "cannot stat 'jq'" in output:
            raise ValueError(
                f"must-gather fails in an disconnected environment\n{output}"
            )
```

petr-balogh: Not sure if it can have some effect on existing execution. If not I think it can be added. 
Don't you want to add some specific exception for this instead of use general ValueError?

As you are referring to disconnected env in exception message, this check should be done only when it's disconnected deployment.

PR #5292, opened 2022-01-03
Deploy an AWS cluster with the new storage classes gp2-csi, gp3-csi

yitzhak12:
Add the functionality to deploy an AWS cluster with the new storage classes gp2-csi, gp3-csi from the ocs-ci project.


ocs_ci/templates/ocs-deployment/aws_gp2_csi_storage_class.yaml
```diff
@@ -0,0 +1,10 @@
+---
```
ebenahar, line 1: do we need the SCs definition saved in templates? asking since they are supposed to exist automatically with OCP 4.10 from what I understand

conf/deployment/aws/gp2_csi_storage_class.yaml
```diff
@@ -0,0 +1,4 @@
+---
+# This config is for AWS deployment with gp2 CSI storage class provisioning
+DEPLOYMENT:
+  gp2_csi: True
```
ebenahar, line 4: thinking maybe it would be better to call the new key `customized_deployment_storage_class` and set it to the name of the actual storage class, so in this case - `gp2-csi`

conf/deployment/aws/gp2_storage_class.yaml
```diff
@@ -0,0 +1,4 @@
+---
+# This config is for AWS deployment with gp2 CSI storage class provisioning
+DEPLOYMENT:
+  gp2: True
```
ebenahar, line 4: gp2 is the default SC so not sure we need this conf file

conf/deployment/aws/gp3_csi_storage_class.yaml
```diff
@@ -0,0 +1,4 @@
+---
+# This config is for AWS deployment with gp3 CSI storage class provisioning
+DEPLOYMENT:
+  gp3_csi: True
```
ebenahar, line 4: thinking maybe it would be better to call the new key customized_deployment_storage_class and set it to the name of the actual storage class, so in this case - `gp3-csi`

ocs_ci/deployment/aws.py
```diff
@@ -61,6 +61,20 @@ def __init__(self):
...
+        ocs_version = version.get_semantic_ocs_version_from_config()
+
+        if ocs_version >= version.VERSION_4_10 and ocp_version >= version.VERSION_4_9:
+            if config.DEPLOYMENT.get("gp2_csi"):
+                self.DEFAULT_STORAGECLASS = "gp2-csi"
+            elif config.DEPLOYMENT.get("gp2"):
+                self.DEFAULT_STORAGECLASS = "gp2"
+            # The default storage class is gp3-csi for OCS 4.10 and above
+            else:
+                self.DEFAULT_STORAGECLASS = "gp3-csi"
```
ebenahar, line 76: if we go with my above suggestions, the conditions here would look like:
```python
            if config.DEPLOYMENT.get("customized_deployment_storage_class"):
                self.DEFAULT_STORAGECLASS = config.DEPLOYMENT.get("customized_deployment_storage_class")

```

conf/deployment/aws/gp2_storage_class.yaml
yitzhak12: But from 4.10 gp3-csi will be the default SC - that's why I added this gp2 YAML file.

conf/deployment/aws/gp2_csi_storage_class.yaml
yitzhak12: Yes, sounds reasonable.

conf/deployment/aws/gp3_csi_storage_class.yaml
yitzhak12: Will do.

ocs_ci/deployment/aws.py
yitzhak12: Will do.

ocs_ci/templates/ocs-deployment/aws_gp2_csi_storage_class.yaml
yitzhak12: I wasn't sure about it, so I added them also. But I think we don't need them.

conf/deployment/aws/gp2_storage_class.yaml
ebenahar: oh didn't know that. So please ignore

conf/deployment/aws/gp2_csi_storage_class.yaml
yitzhak12: Will I rename the file to `gp2_csi_customized_deployment_storage_class.yaml`?

OdedViner (approved):

conf/deployment/aws/gp2_csi_storage_class.yaml
```diff
@@ -0,0 +1,4 @@
+---
+# This config is for AWS deployment with gp2 CSI storage class provisioning
+DEPLOYMENT:
+  customized_deployment_storage_class: 'gp2-csi'
```
ebenahar, line 4: please add `customized_deployment_storage_class` in conf/README.md under `DEPLOYMENT`

ocs_ci/deployment/aws.py
```diff
@@ -61,6 +61,17 @@ def __init__(self):
...
         self.cluster_prefixes_special_rules = CLUSTER_PREFIXES_SPECIAL_RULES
 
+    def deploy_ocp(self, log_cli_level="DEBUG"):
+        super(AWSBase, self).deploy_ocp(log_cli_level)
+        ocp_version = version.get_semantic_ocp_version_from_config()
+        ocs_version = version.get_semantic_ocs_version_from_config()
+
+        if ocs_version >= version.VERSION_4_10 and ocp_version >= version.VERSION_4_9:
+            # The default storage class is gp3-csi for OCS 4.10 and above
+            self.DEFAULT_STORAGECLASS = config.DEPLOYMENT.get(
```
ebenahar, line 71: please don't change the default in ocs-ci for AWS storageclass. We first need to do the evaluation with gp3-sci and then decide along with PM if that's the way to go

ebenahar: in case it will be decided to change the default to gp3-csi, we will need to do a wide communication to the QE and dev teams.
Apart from that, changing the default is not done here but assigning the storageclass name to DEFAULT_STORAGECLASS in ocs_ci/deployment/aws.py:52

yitzhak12: Will change the default SC to be `gp2`. About the second comment, we had better change the default SC here, because we don't want to change the deployments with OCS 4.9 and below. Just from ODF 4.10 and above with use the new SCs.

conf/deployment/aws/gp2_csi_storage_class.yaml
yitzhak12: Yes, that will do.

ebenahar (approved):

OdedViner (approved):

OdedViner (approved):

OdedViner (approved):

jimdevops19 (approved):

conf/deployment/aws/gp3_csi_storage_class.yaml
```diff
@@ -0,0 +1,4 @@
+---
+# This config is for AWS deployment with gp3 CSI storage class provisioning
+DEPLOYMENT:
+  customized_deployment_storage_class: 'gp3-csi'
```
petr-balogh, line 4: the `conf/deployment/{platform}` is the directory where we store whole production jobs configurations which are suppose to contains all details about deployment + usually polarion ID.

This is just a chagne in deployment which probably should be just under `conf/ocsci` folder.
The same for the above. @yitzhak12 can you please move those two files?

PR #5309, opened 2022-01-06
Use Basic auth for accessing OCP repos

dahorak:
Since the recent changes in authentication method for `mirror.openshift.com/enterprise`, we have to replace the original authentication via certificate by newly used basic auth.

### DIFF
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -533,12 +533,12 @@ class AWSUPI(AWSBase):
         repo_file = os.path.basename(repo)
         pod.upload(rhel_pod_obj.name, repo, repo_dst_path)
         # prepare credential files for mirror.openshift.com
-        (
+        with prepare_mirror_openshift_credential_files() as (
             mirror_user_file,
             mirror_password_file,
-        ) = prepare_mirror_openshift_credential_files()
-        pod.upload(rhel_pod_obj.name, mirror_user_file, constants.YUM_VARS_PATH)
-        pod.upload(rhel_pod_obj.name, mirror_password_file, constants.YUM_VARS_PATH)
+        ):
+            pod.upload(rhel_pod_obj.name, mirror_user_file, constants.YUM_VARS_PATH)
+            pod.upload(rhel_pod_obj.name, mirror_password_file, constants.YUM_VARS_PATH)
 
         # Install scp on pod
         rhel_pod_obj.install_packages("openssh-clients")
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -532,14 +532,6 @@ class AWSUPI(AWSBase):
         assert os.path.exists(repo), f"Required repo file {repo} doesn't exist!"
         repo_file = os.path.basename(repo)
         pod.upload(rhel_pod_obj.name, repo, repo_dst_path)
-        # copy the .pem file for our internal repo on all nodes
-        # including ansible pod
-        # get it from URL
-        mirror_pem_file_path = os.path.join(
-            constants.DATA_DIR, constants.INTERNAL_MIRROR_PEM_FILE
-        )
-        dst = "/etc/pki/ca-trust/source/anchors/"
-        pod.upload(rhel_pod_obj.name, mirror_pem_file_path, dst)
         # prepare credential files for mirror.openshift.com
         (
             mirror_user_file,
@@ -572,21 +564,6 @@ class AWSUPI(AWSBase):
                 host,
                 pem_dst_path,
                 f'sudo mv {os.path.join("/tmp", repo_file)} {repo_dst_path}',
-                user=self.rhel_worker_user,
-            )
-            rhel_pod_obj.copy_to_server(
-                host,
-                pem_dst_path,
-                os.path.join(dst, constants.INTERNAL_MIRROR_PEM_FILE),
-                os.path.join("/tmp", constants.INTERNAL_MIRROR_PEM_FILE),
-                user=self.rhel_worker_user,
-            )
-            rhel_pod_obj.exec_cmd_on_node(
-                host,
-                pem_dst_path,
-                f"sudo mv "
-                f'{os.path.join("/tmp", constants.INTERNAL_MIRROR_PEM_FILE)} '
-                f"{dst}",
                 user=self.rhel_worker_user,
             )
             for file_name in (
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -24,6 +24,7 @@ from ocs_ci.utility.aws import (
     terminate_rhel_workers,
 )
 from ocs_ci.utility.bootstrap import gather_bootstrap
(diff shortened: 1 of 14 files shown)


ocs_ci/deployment/aws.py
```diff
@@ -565,21 +566,25 @@ def run_ansible_playbook(self):
...
+                    user=self.rhel_worker_user,
+                )
+                rhel_pod_obj.exec_cmd_on_node(
+                    host,
+                    pem_dst_path,
+                    f"sudo mv "
+                    f"{os.path.join(constants.RHEL_TMP_PATH, file_name)} "
+                    f"{constants.YUM_VARS_PATH}",
+                    user=self.rhel_worker_user,
+                )
```
petr-balogh, line 587: If I see correctly, this code is used on 3 places and and only diff is the pod object and different user name.  So I would move it to some util function (in mirror_openshift.py e.g.) and just pass as argument pod object and user. WDYT?  BTW have you considered the option which Martin brought up yesterday using secret? I think it's good idea to create secret and just in the pod definition to use this secret.

ocs_ci/deployment/aws.py
```diff
@@ -531,14 +532,14 @@ def run_ansible_playbook(self):
...
-        mirror_pem_file_path = os.path.join(
-            constants.DATA_DIR, constants.INTERNAL_MIRROR_PEM_FILE
-        )
-        dst = "/etc/pki/ca-trust/source/anchors/"
-        pod.upload(rhel_pod_obj.name, mirror_pem_file_path, dst)
+        # prepare credential files for mirror.openshift.com
+        (
+            mirror_user_file,
+            mirror_password_file,
+        ) = prepare_mirror_openshift_credential_files()
```
petr-balogh, line 539: Proposed function below can handle also this on one place instead of repeat it.

ocs_ci/deployment/aws.py
dahorak: > If I see correctly, this code is used on 3 places and and only diff is the pod object and different user name. So I would move it to some util function (in mirror_openshift.py e.g.) and just pass as argument pod object and user. WDYT?

The main problem here is, that the following two functions [aws.py - run_ansible_playbook()](https://github.com/red-hat-storage/ocs-ci/blob/396377c7dfc9c5c7a5a23ea6afef896533a2f6d5/ocs_ci/deployment/aws.py#L516-L604) and [platform_nodes.py - attach_rhel_nodes_to_upi_cluster(...)](https://github.com/red-hat-storage/ocs-ci/blob/396377c7dfc9c5c7a5a23ea6afef896533a2f6d5/ocs_ci/ocs/platform_nodes.py#L864-L955) are basically the same, with just small differences and I'm changing just small part of that functions, so I think it make more sense to polish whole those functions and not only small part of that. But I don't want to do it in this PR.
Another point is, that the steps I'm doing there (copying the credential variable files) are closely tight to the copying of the repo files and it will make more sense to cover those together.
And because this part of code and functionality is quite unknown to me, I don't want to do too much big changes and break anything else, so I would propose to leave it this way (if it will work) and do some polishing later in new PR.

> BTW have you considered the option which Martin brought up yesterday using secret? I think it's good idea to create secret and just in the pod definition to use this secret.

It is surely good idea and worth to consider, but I didn't considered it too deeply, because of following reasons:
- I want to fix the issue with changed authentication method without many additional changes in the code and if possible without any changes to the existing logic. The affected code is quite unknown for me and I wasn't able quickly review all the possible impacts of bigger change what will the usage of secret require.
- Additionally the newly used credential files are tightly connected to the repo files, which are uploaded to the pod in some way, and it doesn't make much sense, to use too much different approach for the credentials.
- And also in the existing code - the same/similar approach is used for uploading the files to pods same as to nodes, so I'm not sure if it make much sense to differentiate this.

dahorak:
With the last commit, I changed the `prepare_mirror_openshift_credential_files` function to context manager, to handle deletion of the credential files from local storage.

ocs_ci/deployment/install_ocp_on_rhel.py
```diff
@@ -214,19 +220,25 @@ def prepare_rhel_nodes(self):
...
+            for file_name in (
+                constants.MIRROR_OPENSHIFT_USER_FILE,
+                constants.MIRROR_OPENSHIFT_PASSWORD_FILE,
+            ):
+                self.rhelpod.copy_to_server(
+                    node,
+                    self.pod_ssh_key_pem,
+                    os.path.join(constants.YUM_VARS_PATH, file_name),
+                    os.path.join(constants.RHEL_TMP_PATH, file_name),
+                    user=self.rhel_worker_user,
```
vavuthu, line 232: self.rhel_worker_user doesn't exist in this module

ocs_ci/deployment/install_ocp_on_rhel.py
```diff
@@ -214,19 +220,25 @@ def prepare_rhel_nodes(self):
...
+                )
+                self.rhelpod.exec_cmd_on_node(
+                    node,
+                    self.pod_ssh_key_pem,
+                    f"sudo mv "
+                    f"{os.path.join(constants.RHEL_TMP_PATH, file_name)} "
+                    f"{constants.YUM_VARS_PATH}",
+                    user=self.rhel_worker_user,
```
vavuthu, line 240: self.rhel_worker_user doesn't exist in this module

ocs_ci/utility/mirror_openshift.py
```diff
@@ -0,0 +1,59 @@
...
+        data/mirror_openshift_user,
+        data/mirror_openshift_password.
+
+    Those files are used as variable files for repo files (stored in /etc/yum/vars/
+    on the target RHEL nodes or pods).
+
+    Returns:
+        (tuple): tuple containing two strings - file names for username file and password file
+
+    Raises
```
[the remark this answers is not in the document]
dahorak: Fixed in 0442ac3.

ocs_ci/deployment/install_ocp_on_rhel.py
dahorak: Good catch! ... fixed in 0442ac3.

ocs_ci/deployment/install_ocp_on_rhel.py
dahorak: Good catch! ... fixed in 0442ac3.

vavuthu (approved):

ocs_ci/deployment/aws.py
clacroix12: I agree with @petr-balogh on reducing the duplication. I think even if we are going to tackle the duplication of the overarching functions in a later PR it would be worth it to reduce the duplication of the newly added code. That would leave us with less to clean up later. Unless you think this will make things more difficult to refactor later in which case I'm fine with the duplication as long as we have an action item created to address the larger issues here.

petr-balogh: Let's merge this as is for now and open issue to remove duplication please.

petr-balogh (approved):

PR #5328, opened 2022-01-11
Skip the test_mcg_db_backup_recovery if the version is lower than 4.9

OdedViner:
Object replication is supported from version 4.9 
#5329

### DIFF
--- a/tests/e2e/system/test_mcg_recovery.py
+++ b/tests/e2e/system/test_mcg_recovery.py
@@ -6,6 +6,7 @@ from ocs_ci.framework.pytest_customization.marks import (
     tier4,
     ignore_leftovers,
     polarion_id,
+    skipif_ocs_version,
 )
 from ocs_ci.framework.testlib import E2ETest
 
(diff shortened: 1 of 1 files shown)


PrasadDesala (approved):
Please create an issue for the test failure and link it to this PR

tests/e2e/system/test_mcg_recovery.py
```diff
@@ -16,6 +17,7 @@
 @tier4a
 @ignore_leftovers
 @polarion_id("OCS-2716")
+@skipif_ocs_version("<4.9")
```
Shrivaibavi, line 20: Since we changed the terms from 4.9 shouldn't it be skipif_odf_version("<4.9"), So this marker can be widely used by ocs-qe

OdedViner: @petr-balogh what do you think?

petr-balogh: We have OCS on a lot of places, we will not change this now, for sure not in this PR. Also whole project is named OCS-CI ;) we also didn't rename to ODF-CI

petr-balogh (approved):

MeridianExplorer (approved):

Shrivaibavi (approved):

PR #5337, opened 2022-01-12
improve the create_storage_class helper function

Avilir:
In This PR i add another parameter to the function - fs_name (default value of None for backward compatibility) so we can create the storage class on new ceph-filesystem and not only on the default one.

I also remove some code duplication from the function.


OdedViner (approved):

ypersky1980 (approved):
LGTM

ramkiperiy (approved):

PR #5341, opened 2022-01-12
Adding retry for exec_oc_debug_cmd

vavuthu:
Fixes: #2571 

### DIFF
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -13,6 +13,7 @@ import copy
 
 from ocs_ci.ocs.exceptions import (
     CommandFailed,
+    DebugPodCreateFailed,
     NotSupportedFunctionError,
     NonUpgradedImagesFoundError,
     ResourceWrongStatusException,
@@ -159,7 +160,7 @@ class OCP(object):
             return yaml.safe_load(out)
         return out
 
-    @retry(CommandFailed, tries=3, delay=5, backoff=1)
+    @retry(DebugPodCreateFailed, tries=3, delay=5, backoff=1)
     def exec_oc_debug_cmd(self, node, cmd_list, timeout=300):
         """
         Function to execute "oc debug" command on OCP node
@@ -174,6 +175,8 @@ class OCP(object):
 
         Raises:
             CommandFailed: When failure in command execution
+            DebugPodCreateFailed: in case of debug pod creation fails
+
         """
         # Appending one empty value in list for string manipulation
         create_cmd_list = copy.deepcopy(cmd_list)
@@ -186,6 +189,8 @@ class OCP(object):
         )
         if err_msg in out:
             raise CommandFailed
+        elif "error:" in out:
+            raise DebugPodCreateFailed
         else:
             return out
 
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -159,6 +159,7 @@ class OCP(object):
             return yaml.safe_load(out)
         return out
 
+    @retry(CommandFailed, tries=3, delay=5, backoff=1)
     def exec_oc_debug_cmd(self, node, cmd_list, timeout=300):
         """
         Function to execute "oc debug" command on OCP node
--- a/ocs_ci/ocs/exceptions.py
+++ b/ocs_ci/ocs/exceptions.py
@@ -446,3 +446,7 @@ class BenchmarkTestFailed(Exception):
 
 class WrongVersionExpression(ValueError):
     pass
+
+
+class DebugPodCreateFailed(Exception):
+    pass


Avilir (dismissed):
LGTM

petr-balogh (approved):

ocs_ci/ocs/ocp.py
```diff
@@ -159,6 +159,7 @@ def exec_oc_cmd(
             return yaml.safe_load(out)
         return out
 
+    @retry(CommandFailed, tries=3, delay=5, backoff=1)
```
jilju, line 162: Will retry work if the actual command (argument  cmd_list) fail ? 
Retry is needed only if the debug pod is not created due to some issue.

clacroix12:
`exec_oc_debug_cmd` is used in a lot of different places with different commands. Are we sure that none of these will cause issues if ran more than once due to errors?

vavuthu:
> `exec_oc_debug_cmd` is used in a lot of different places with different commands. Are we sure that none of these will cause issues if ran more than once due to errors?

It won't cause any harm if it ran twice bcoz if 1st time fails, we are just running same command ( 1s time cmd fail to execute )

ocs_ci/ocs/ocp.py
vavuthu: cmd fails dcoz debug pod was not created so we are retrying to create debug pod

mbukatov (changes requested):
I agree with @clacroix12 , this doesn't seem like a good idea.

Especially because `CommandFailed` exception could mean both problem with debug pod itself (so that no target command is executed) or a problem with execution of a target program. Moreover this command allows to run multiple commands, which complicates things further.

Besides that, there is a limitation of number of concurrent connections via oc debug for a single node (see https://bugzilla.redhat.com/show_bug.cgi?id=1943287) and automatic retry in multiple threads could actually make things worse.

clacroix12:
Something we do with the [ceph health check](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L1859-L1915) is have a wrapper function that implements the retry. This leaves the base function available to be called directly if the user does not want to retry.

Could we do something like this here? Implement a new function with the retry that just calls exec_oc_debug_cmd. This way tests that would benefit from retrying commands that failed can call that and those that do not want that functionality can remain as is.

vavuthu:
exec_oc_debug_cmd is a wrapper around exec_oc_cmd. First I would like to separate
1. creation of debug pod failure and 
2. command failed

There is no easy way to fix for 2. command failed since it takes the list of commands as input. ( which command is failed in the list exactly )

From the issue #2571 , mostly failures is on 1. creation of debug pod failure and we are running debug pods serially

vavuthu:
@mbukatov @clacroix12 Could you revisit this PR and above comment

ocs_ci/ocs/ocp.py
```diff
@@ -185,6 +189,8 @@ def exec_oc_debug_cmd(self, node, cmd_list, timeout=300):
         )
         if err_msg in out:
             raise CommandFailed
+        elif "error:" in out:
```
clacroix12, line 192: Is this sufficient to know that the debug pod was failed to create or is there a more explicit error message we can try to look for? Do you have an example of the `out` in this case?

vavuthu: "error" is sufficient in this case. I saw below 2 use cases for debug pod failures
1.
```
18:02:37 - MainThread - ocs_ci.utility.utils - INFO - Executing command: oc debug nodes/ip-10-0-149-60.us-east-2.compute.internal -- df
18:02:42 - ThreadPoolExecutor-381_0 - ocs_ci.utility.utils - WARNING - Command stderr: Error from server (NotFound): pods "pod-test-cephfs-2e71ee0092804860aad497c93d6c7f5d" not found

18:03:34 - MainThread - ocs_ci.utility.utils - WARNING - Command stderr: Starting pod/ip-10-0-149-60us-east-2computeinternal-debug ...
To use host binaries, run `chroot /host`

Removing debug pod ...
error: Back-off pulling image "registry.redhat.io/rhel7/support-tools"
```
and
2.
```
2022-01-07 09:07:07,112 - MainThread - INFO - ocs_ci.utility.utils.exec_cmd.487 - Executing command: oc debug nodes/compute-7 -- chroot /host /bin/bash -c "ifconfig $(route | grep default | awk '{print $(NF)}') down || echo 'CMD FAILED'; "
2022-01-07 09:07:07,698 - MainThread - DEBUG - ocs_ci.utility.utils.exec_cmd.502 - Command stdout is empty
2022-01-07 09:07:07,698 - MainThread - WARNING - ocs_ci.utility.utils.exec_cmd.506 - Command stderr: Starting pod/compute-7-debug ...
To use host binaries, run `chroot /host`

Removing debug pod ...
error: unable to create the debug pod "compute-7-debug"
```

PR #5359, opened 2022-01-18
Update fio server pod count in test_fio_benchmark.py

pintojoy:
Server pod count is reduced from 51 to 9 to achieve high iops in fio tests. As per comments on bug-2015520, multiple tests were run with different pod count and iodepths were run and pod count = 9 and iodepth = 16 was chosen.

### DIFF
--- a/tests/e2e/performance/io_workload/test_fio_benchmark.py
+++ b/tests/e2e/performance/io_workload/test_fio_benchmark.py
@@ -176,7 +176,7 @@ class TestFIOBenchmark(PASTest):
         # To make sure the number of App pods will not be more then 50, in case
         # of large data set, changing the size of the file each pod will work on
         if self.total_data_set > 500:
-            self.filesize = int(ceph_capacity * 0.008)
+            self.filesize = int(ceph_capacity * 0.0415)
             self.crd_data["spec"]["workload"]["args"][
                 "filesize"
             ] = f"{self.filesize}GiB"
@@ -199,7 +199,6 @@ class TestFIOBenchmark(PASTest):
         """
         if io_pattern == "sequential":
             self.crd_data["spec"]["workload"]["args"]["jobs"] = ["write", "read"]
-            self.crd_data["spec"]["workload"]["args"]["iodepth"] = 1
         if io_pattern == "random":
             self.crd_data["spec"]["workload"]["args"]["jobs"] = [
                 "randwrite",


tests/e2e/performance/io_workload/test_fio_benchmark.py
```diff
@@ -176,7 +176,7 @@ def setting_storage_usage(self):
         # To make sure the number of App pods will not be more then 50, in case
```
Avilir, line 176: please change this number as well, since the new value will reduce the number of servers to 24 ? 25 ?

tests/e2e/performance/io_workload/test_fio_benchmark.py
```diff
@@ -199,7 +199,6 @@ def setting_io_pattern(self, io_pattern):
         """
         if io_pattern == "sequential":
             self.crd_data["spec"]["workload"]["args"]["jobs"] = ["write", "read"]
-            self.crd_data["spec"]["workload"]["args"]["iodepth"] = 1
```
Avilir, line 202: why this line removed ?

pintojoy: sequential fio runs gave considerable degradaton upon running with iodepth=1  and pod count of 9 whereas with pod count 9 and iodepth =16 higher iops were achieved. Hence for both sequential and random workload pod count of 9 and iodepth of 16 is used. (https://docs.google.com/document/d/1vTnBv0j5-Tnle3KdG3PU36G_N1SbjGJRiwmBb07ZYlg/edit#)

Avilir (approved):
LGTM

tests/e2e/performance/io_workload/test_fio_benchmark.py
ramkiperiy: @pintojoy so are we removing this line? or we would like to increase the value from 1 to 16, iodepth default value is 1 in fio.

pintojoy: @ramkiperiy Previously we were running with following configuration:
 server pods =51 iodepth =16 for random workload and server pods =51 iodepth =1 for sequential workload.

But with server pods=51 we were over-stressing the system, Hence pod count 9 was chosen and with pod count =9 and iodepth =16 random workload gave improved iops value whereas with sequential(pod count =9 and iodepth=1) degradation was seen, which again goes away when iodepth is increased to 16. Hence to maintain uniformity, pod count =9 and iodepth =16 is chosen for both sequential and random workloads.

ramkiperiy: ok, My question was where are we setting it ? noticed it's in benchmark_fio.yaml param list, so default it would be 16

ramkiperiy (approved):

PR #5362, opened 2022-01-18
Return to const usage for noobaa scc test

bernerhat:
after a discussion with dev team we understood that the user used by the db is now the same across previous versions (backported) so the check for version is no longer needed 

Signed-off-by: Amit Berner <aberner@redhat.com>


mbukatov (approved):

mbukatov:
Is this part of public API of the product? If yes, I would feel bit uncomfortable with the way how it's managed.

mbukatov (approved):

MeridianExplorer (approved):

Issue #5365, opened 2022-01-19
Destroy job fails due to KMS cleanup failures


vavuthu:
```
2022-01-18 17:34:54              if check and retcode:
2022-01-18 17:34:54  >               raise CalledProcessError(retcode, process.args,
2022-01-18 17:34:54                                           output=stdout, stderr=stderr)
2022-01-18 17:34:54  E               subprocess.CalledProcessError: Command '['vault', 'token', 'lookup', 's.uTNFmMlsyckrvGJCi4mVRfQe.lxuns']' returned non-zero exit status 2.
2022-01-18 17:34:54  
2022-01-18 17:34:54  /usr/lib64/python3.8/subprocess.py:516: CalledProcessError
```
Job: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-destroy-ocs-cluster/10137/consoleFull

PR #5366, opened 2022-01-19
Resuming teardown incase KMS cleanup failures

vavuthu:
Fixes: #5365 

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1367,8 +1367,13 @@ def cluster(request, log_cli_level, record_testsuite_property):
             # If KMS is configured, clean up the backend resources
             # we are doing it before OCP cleanup
             if config.DEPLOYMENT.get("kms_deployment"):
-                kms = KMS.get_kms_deployment()
-                kms.cleanup()
+                try:
+                    kms = KMS.get_kms_deployment()
+                    kms.cleanup()
+                except Exception as ex:
+                    log.error(f"Failed to cleanup KMS. Exception is: {ex}")
+                    log.info("resuming teardown")
+
             deployer.destroy_cluster(log_cli_level)
 
         request.addfinalizer(cluster_teardown_finalizer)


petr-balogh (approved):

tests/conftest.py
```diff
@@ -1367,8 +1367,13 @@ def cluster_teardown_finalizer():
             # If KMS is configured, clean up the backend resources
             # we are doing it before OCP cleanup
             if config.DEPLOYMENT.get("kms_deployment"):
-                kms = KMS.get_kms_deployment()
-                kms.cleanup()
+                try:
+                    kms = KMS.get_kms_deployment()
+                    kms.cleanup()
+                except Exception as ex:
+                    log.error(f"Failed to cleanup KMS. Exception is: {ex}")
```
petr-balogh, line 1374: One downside of this is that we will never get this info if it will happen as we will continue the job and think that it was successful and we will never get this fixed if it was not temporary issue.

@shylesh any input on this? 
@vavuthu is this happening permanently? If so we will need to fix the root cause.
If it was just temp issue then I think we should probably continue failing the job to teardown as the cleanup is required in vault side IIUC and if it's not done here we have leftovers.

I would wait with @shylesh comment here before merging.

shylesh: @petr-balogh correct, vault will have stale entries.  Earlier you suggested me to do the cleanup script directl y on the backend side just like aws cleanup job, I looked at it for couple of times but since it involves combing jenkins for alive cluster and comparing to the entries in vault I couldn't comeup with any simple implementation and it's been a backlog. For now we can merge this and I will create a ticket for this  and might have to work if priority is high.

vavuthu: @petr-balogh teardown failing whenever we have issue with KMS ( currently for KMS, certs are expired )

PR #5368, opened 2022-01-19
Add RHEL 7.9 AMI and possibility to define RHEL 8

petr-balogh:

### DIFF
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -474,6 +474,8 @@ class AWSUPI(AWSBase):
         cluster_id = get_infra_id(self.cluster_path)
         num_workers = int(os.environ.get("num_workers", 3))
         logging.info(f"Creating {num_workers} RHEL workers")
+        rhel_version = config.ENV_DATA["rhel_version"]
+        rhel_worker_ami = config.ENV_DATA[f"rhel{rhel_version}_worker_ami"]
         for i in range(num_workers):
             self.gather_worker_data(f"no{i}")
             logging.info(f"Creating {i + 1}/{num_workers} worker")
@@ -488,7 +490,7 @@ class AWSUPI(AWSBase):
                         },
                     },
                 ],
-                ImageId=config.ENV_DATA["rhel_worker_ami"],
+                ImageId=rhel_worker_ami,
                 SubnetId=self.worker_subnet,
                 InstanceType=config.ENV_DATA["rhel_worker_instance_type"],
                 MaxCount=1,
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -1067,7 +1067,7 @@ RHCOS_WORKER_CONF = os.path.join(CONF_DIR, "ocsci/aws_upi_rhcos_workers.yaml")
 AWS_WORKER_NODE_TEMPLATE = "06_cluster_worker_node.yaml"
 AWS_S3_UPI_BUCKET = "ocs-qe-upi"
 AWS_WORKER_LOGICAL_RESOURCE_ID = "Worker0"
-RHEL_WORKERS_CONF = os.path.join(CONF_DIR, "ocsci/aws_upi_rhel_workers.yaml")
+RHEL_WORKERS_CONF = os.path.join(CONF_DIR, "ocsci/aws_upi_rhel{version}_workers.yaml")
 
 # Users
 NB_SERVICE_ACCOUNT_BASE = "system:serviceaccount:openshift-storage:{}"
--- a/conf/README.md
+++ b/conf/README.md
@@ -182,6 +182,10 @@ higher priority).
 * `ignition_data_encoding` - Encoding type used for the ignition config data
 * `device_size` - Size (in GB) to use for storage device sets
 * `rhel_workers` - Use RHEL workers instead of RHCOS, for UPI deployments (Default: false)
+* `rhel_version` - For AWS UPI deployment over RHEL. Based on this value we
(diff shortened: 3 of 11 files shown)


mbukatov (dismissed):

ocs_ci/deployment/aws.py
```diff
@@ -488,7 +492,7 @@ def create_rhel_instance(self):
                         },
                     },
                 ],
-                ImageId=config.ENV_DATA["rhel_worker_ami"],
+                ImageId=rhel_worker_ami,
```
shylesh, line 495: @petr-balogh , I was wrongly thought that we are setting some env var for imageid, but this is correct. I guess it should work smoothly unless we are using any rhel7 specific admin commands post deployment of instance.

shylesh (dismissed):
LGTM

dahorak (approved):

ocs_ci/framework/conf/ocp_version/ocp-4.10-config.yaml
```diff
@@ -9,4 +9,6 @@ DEPLOYMENT:
   # https://docs.openshift.com/container-platform/4.7/post_installation_configuration/machine-configuration-tasks.html#machine-config-overview-post-install-machine-configuration-tasks
   ignition_version: "3.2.0"
 ENV_DATA:
+  rhel8_worker_ami: 'ami-074bab065c112399f' # RHEL 8.4
```
clacroix12, line 12: Can't these be set in the default config or be set as constants instead of explicitly in each individual config?

petr-balogh: I moved it out of default config to be bounded to specific ocp version. E.g in 4.5 documentation there is mentioned rhel version 7.6 or something like this, from 4.6 till 4.9 ocp doc mentioning explicitly 7.9 and I guess it can change in futures versions as well, hence I would like to be explicit in each ocp config file without need to think in the future that if I will change the default image that it can break older versions deployment.

shylesh (approved):
LGTM

ocs_ci/framework/conf/ocp_version/ocp-4.10-config.yaml
clacroix12: Well for each version of RHEL we should only have one AMI that we use, correct? If that's the case then the configs can specify the _version_ of RHEL that they require and we can handle mapping that version to the actual AMI ID elsewhere so if that AMI ID ever changes we don't need to change it across a dozen config files. Instead it's a one line change. WDYT?

petr-balogh: We still need to have ability to change to specific version we want if we need to test specific version. I came up with the change which will be more fit to what you would like to see. 

Anyway it will require the entry to the ocp version config files once we will diverge from what is the default and we will need to e.g. to ocp 4.6 config files put: `rhel_version: "7.9"` it it will be the lates supported version.

Anyway I hope that if OCP supported on RHEL 7, every new update to minor version should probably still support also those older versions of OCP even they do not update the documentation for older version to keep version updated. Let's see how it will go.

@clacroix12 let me know if you rather like this approach. Before I went with explicit definition because we go with such approach also for the other values.

clacroix12: Yeah I think this works. We can absolutely specify `rhel_version` in individual config files if the supported version differs from the default. Having the AMI ID specified in one place and being mapped to from the version is cleaner and calls for fewer changes if we need to update the AMI for any reason.

clacroix12 (approved):

ocs_ci/ocs/constants.py
```diff
@@ -1067,7 +1067,7 @@
 AWS_WORKER_NODE_TEMPLATE = "06_cluster_worker_node.yaml"
 AWS_S3_UPI_BUCKET = "ocs-qe-upi"
 AWS_WORKER_LOGICAL_RESOURCE_ID = "Worker0"
-RHEL_WORKERS_CONF = os.path.join(CONF_DIR, "ocsci/aws_upi_rhel_workers.yaml")
+RHEL_WORKERS_CONF = os.path.join(CONF_DIR, "ocsci/aws_upi_rhel{version}_workers.yaml")
```
vavuthu, line 1070: where is ```version``` coming from

petr-balogh: this is formatted here:
https://github.com/red-hat-storage/ocs-ci/pull/5368/files#diff-478e5b871f475f1e312384c0eeb216726f11da23e3d3ac4df80aa3ee5d92e441R1306

petr-balogh:
I see that nodes are successfully RHEL 7.9 here:
http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/jnk-pr5368-b2258/jnk-pr5368-b2258_20220126T155147/logs/failed_testcase_ocs_logs_1643212716/deployment_ocs_logs/ocs_must_gather/quay-io-rhceph-dev-ocs-must-gather-sha256-25baa44ae655531e53be759b708e9eba284f7a908c7434d76d308ce42bfa8069/cluster-scoped-resources/oc_output/get_nodes_-o_wide_--show-labels

Then it failed with some weird issue with catalog source redhat-operators didn't get up.
Trying again, but marking as verified as the RHEL deployment passed as expected.

vavuthu (approved):

petr-balogh:
Job has successfully installed OCP with RHEL 7.9 so I will merge this one. If we will se the other issue we see in deployment verification job we will need to open other issue for it as it's related to this change I think

PR #5371, opened 2022-01-19
Verify worker nodes are tainted

OdedViner:

### DIFF
--- a/conf/ocsci/tainted_worker_nodes_ui.yaml
+++ b/conf/ocsci/tainted_worker_nodes_ui.yaml
@@ -1,0 +1,3 @@
+---
+DEPLOYMENT:
+    ocs_operator_nodes_to_taint: 3
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -433,6 +433,8 @@ def ocs_install_verification(
     if managed_service:
         verify_managed_service_resources()
 
+    taint_node_verification()
+
 
 def mcg_only_install_verification(ocs_registry_image=None):
     """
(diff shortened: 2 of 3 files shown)


ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -624,6 +626,23 @@ def osd_encryption_verification():
...
+    from ocs_ci.ocs.node import get_nodes
+
+    if config.DEPLOYMENT.get("ocs_operator_nodes_to_taint", 0) > 0:
+        log.info("Verify worker nodes are tainted")
+        node_objs = get_nodes()
+        for node_obj in node_objs:
+            if node_obj.get("data").get("spec").get("taints")[0]["value"] != "true":
+                error_msg = f"The worker node {node_obj.name} is not tainted"
+                logging.error(error_msg)
+                raise ValueError(error_msg)
```
mbukatov, line 643: What if `ocs_operator_nodes_to_taint != len(node_objs)` (assuming `get_nodes()` returns all worker nodes)?

OdedViner: assuming get_nodes() returns all worker nodes --> correct
On ui_deployment, I dont test this scenario because I choose all worker nodes
image

conf/ocsci/tainted_worker_nodes_ui.yaml
```diff
@@ -0,0 +1,3 @@
+---
+DEPLOYMENT:
+    ocs_operator_nodes_to_taint: 3
```
PrasadDesala, line 3: Not sure if we would need a new yaml for tainted worker nodes. Cant we use 
https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/tainted_worker_nodes.yaml ?

OdedViner: I want to test only taint node [no infra nodes]

Issue #5383, opened 2022-01-23
Need to avoid clusters automatic tests of ODF for usecase of ACM import clusters without ODF


AviadP:
Clean errors of ceph tool box, storage cluster check, ceph health etc.

petr-balogh:
@AviadP the subject or the description of the issue should mention that it's related to import scenario, not whole regional DR setup.

PR #5384, opened 2022-01-23
Add regional dr arg for run-ci

AviadP:
1. added new run-ci argument
2. skip some checks when this argument selected

fix [issue](https://github.com/red-hat-storage/ocs-ci/issues/5383)

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -11,18 +11,18 @@
 
 # in this RUN section we will keep default parameters for run of OCS-CI
 RUN:
-  username: "kubeadmin"
-  password_location: "auth/kubeadmin-password"
+  username: 'kubeadmin'
+  password_location: 'auth/kubeadmin-password'
   log_dir: "/tmp"
-  run_id: null # this will be redefined in the execution
-  kubeconfig_location: "auth/kubeconfig" # relative from cluster_dir
-  cli_params: {} # this will be filled with CLI parameters data
+  run_id: null  # this will be redefined in the execution
+  kubeconfig_location: 'auth/kubeconfig' # relative from cluster_dir
+  cli_params: {}  # this will be filled with CLI parameters data
   regional_dr: False # avoid some checks and fixtures related to ODF
   # If the client version ends with .nightly, the version will be exposed
   # to the latest accepted OCP nightly build version
-  client_version: "4.10.0-0.nightly"
-  bin_dir: "./bin"
-  google_api_secret: "~/.ocs-ci/google_api_secret.json"
+  client_version: '4.10.0-0.nightly'
+  bin_dir: './bin'
+  google_api_secret: '~/.ocs-ci/google_api_secret.json'
   # Following chrome params are for openshift console UI testing
   force_chrome_branch_base: "665006"
   force_chrome_branch_sha256sum: "a1ae2e0950828f991119825f62c24464ab3765aa219d150a94fb782a4c66a744"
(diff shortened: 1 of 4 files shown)


ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -305,6 +305,11 @@ def pytest_addoption(parser):
         loaded when run-ci starts
         """,
     )
+    parser.addoption(
+        "--regional-dr",
```
petr-balogh, line 309: Please change to something related to import as this is suppose to be only for import OCP clusters to ACM.

tests/conftest.py
```diff
@@ -3219,12 +3229,15 @@ def ceph_toolbox(request):
     deploy = config.RUN["cli_params"]["deploy"]
     teardown = config.RUN["cli_params"].get("teardown")
     skip_ocs = config.ENV_DATA["skip_ocs_deployment"]
+    regional_dr = config.RUN["regional_dr"]
```
petr-balogh, line 3232: You will probably need to also skip alerting autouse fixtures for import scenario for now  https://github.com/red-hat-storage/ocs-ci/pull/5384/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R3174-R3180

Based on what we see here:
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-multi-ocs-cluster-x3/224/console

it's problematic on cluster without ODF, not sure why exactly as I didn't dig in to the alerting much.  

For now let's skip it there for import ACM scenario and later on we can figure out if we need to even re-enable it.

ocs_ci/framework/pytest_customization/ocscilib.py
AviadP: done

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -17,6 +17,7 @@ RUN:
   run_id: null  # this will be redefined in the execution
   kubeconfig_location: 'auth/kubeconfig' # relative from cluster_dir
   cli_params: {}  # this will be filled with CLI parameters data
+  regional_dr: False # avoid some checks and fixtures related to ODF
```
petr-balogh, line 20: I think this should be changed to : import_without_odf .

You on some places directly access this key without .get so if it's not defined here it will fail with keyError.

tests/conftest.py
```diff
@@ -3163,11 +3175,14 @@ def log_alerts(request):
     """
     teardown = config.RUN["cli_params"].get("teardown")
     dev_mode = config.RUN["cli_params"].get("dev_mode")
+    import_without_odf = config.RUN["cli_params"].get("import-without-odf")
```
petr-balogh, line 3178: Why not go with ocsci_config.RUN["import_without_odf"] like on other places?  ANd if you go with cli_params are you sure about `import-without-odf` shouldnt this be: `import_without_odf`? I am not sure about it hence would rather go with the same approach as on other places.

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -305,6 +305,11 @@ def pytest_addoption(parser):
         loaded when run-ci starts
         """,
     )
+    parser.addoption(
+        "--import-without-odf",
+        dest="import_without_odf",
```
petr-balogh, line 310: default=False,

AviadP: added

PR #5385, opened 2022-01-24
PR to add more logging and validation in test scale_pvc_create_delete

ramkiperiy:
Added more logging to PVC create/delete function to identify the failure more speifically
Added more validation for kube_obj deletion is successful

### DIFF
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -1546,6 +1546,12 @@ def measure_pvc_creation_time_bulk(interface, pvc_name_list, wait_time=60):
             start = [i for i in logs if re.search(f"provision.*{name}.*started", i)]
             end = [i for i in logs if re.search(f"provision.*{name}.*succeeded", i)]
             if not start or not end:
+                logging.info(f"PVC {name} missing start or end time")
+                try:
+                    logging.info(f"PVC {name} creation start {start}")
+                    logging.info(f"PVC {name} creation complete {end}")
+                except UnexpectedBehaviour:
+                    logging.warning(f"PVC {name} creation data missing")
                 no_data_list.append(name)
 
         if no_data_list:
@@ -1622,6 +1628,12 @@ def measure_pv_deletion_time_bulk(
             start = [i for i in logs if re.search(f'delete "{pv}": started', i)]
             end = [i for i in logs if re.search(f'delete "{pv}": succeeded', i)]
             if not start or not end:
+                logging.info(f"PVC {pv} missing start or end time")
+                try:
+                    logging.info(f"PVC {pv} deletion start {start}")
+                    logging.info(f"PVC {pv} deletion complete {end}")
+                except UnexpectedBehaviour:
+                    logging.warning(f"PVC {pv} deletion data missing")
                 no_data_list.append(pv)
 
         if no_data_list:
--- a/tests/e2e/scale/test_pvc_creation_deletion_scale.py
+++ b/tests/e2e/scale/test_pvc_creation_deletion_scale.py
@@ -136,7 +136,13 @@ class TestPVCCreationDeletionScale(E2ETest):
 
         # Delete kube_job
         job_file1.delete(namespace=self.namespace)
+        job_file1.wait_for_delete(
+            resource_name=job_file1.name, namespace=self.namespace
+        )
         job_file2.delete(namespace=self.namespace)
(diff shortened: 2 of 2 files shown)


ocs_ci/helpers/helpers.py
```diff
@@ -1546,6 +1546,12 @@ def measure_pvc_creation_time_bulk(interface, pvc_name_list, wait_time=60):
             start = [i for i in logs if re.search(f"provision.*{name}.*started", i)]
             end = [i for i in logs if re.search(f"provision.*{name}.*succeeded", i)]
             if not start or not end:
+                logging.info(f"PVC {name} missing start or end time")
+                try:
+                    logging.info(f"PVC {name} creation start {start}")
+                    logging.info(f"PVC {name} creation complete {end}")
+                except UnexpectedBehaviour:
+                    logging.warning(f"PVC {name} creation data missing")
```
mbukatov, line 1554: What would raise `UnexpectedBehaviour` exception in the try block? At first sight, I don't see anything which would raise it. Maybe we just want to log this info without having the try except block in the first place.

ocs_ci/helpers/helpers.py
```diff
@@ -1622,6 +1628,12 @@ def measure_pv_deletion_time_bulk(
             start = [i for i in logs if re.search(f'delete "{pv}": started', i)]
             end = [i for i in logs if re.search(f'delete "{pv}": succeeded', i)]
             if not start or not end:
+                logging.info(f"PVC {pv} missing start or end time")
+                try:
+                    logging.info(f"PVC {pv} deletion start {start}")
+                    logging.info(f"PVC {pv} deletion complete {end}")
+                except UnexpectedBehaviour:
+                    logging.warning(f"PVC {pv} deletion data missing")
```
mbukatov, line 1636: The same point as in the previous case.

mbukatov (commented):
I have some questions, but overall this looks like a good improvement.

ocs_ci/helpers/helpers.py
ramkiperiy: @mbukatov we are observing logs missing from CSI when we do bulk of PVC creation and deletion. adding this log will help in to understand if any of the value/log is not present in the CSI logs.
So expectation behind the exception is either `start` or `end` will get null value and want to display this warning before failing the test.

ramkiperiy: We are suspecting CSI pod logs are getting cleaned up after a certain point, i.e. if we do create 1500 PVC creation and deletion 4 times CSI logs will be loaded, it can hold only particular size and further it's getting flushed (i.e. it's kind of all logs lost/missing), planning to use logging to resolve this problem but still in exploratory stage.

ocs_ci/helpers/helpers.py
ramkiperiy: ack, replied in above comment

tests/e2e/scale/test_pvc_creation_deletion_scale.py
```diff
@@ -18,6 +18,14 @@
 log = logging.getLogger(__name__)
 
 
+@pytest.fixture()
+def setup_fixture(install_logging):
+    """
+    Installs openshift-logging
+    """
+    logging.info("Testcases execution post deployment of openshift-logging")
+
```
ramkiperiy, line 27: @Shrivaibavi ^^

tests/e2e/scale/test_pvc_creation_deletion_scale.py
```diff
@@ -18,6 +18,14 @@
 log = logging.getLogger(__name__)
 
 
+@pytest.fixture()
+def setup_fixture(install_logging):
+    """
+    Installs openshift-logging
+    """
+    logging.info("Testcases execution post deployment of openshift-logging")
```
Shrivaibavi, line 26: Please change the log info according to your needs

tests/e2e/scale/test_pvc_creation_deletion_scale.py
```diff
@@ -18,6 +18,14 @@
...
+
+
```
Shrivaibavi, line 28: Please call the setup fixture

ramkiperiy: Yeah, I missed this `@pytest.mark.usefixtures(setup_fixture.__name__)`, let me update the code.

Issue #5386, opened 2022-01-24
--delete-local-data has been deprecated for oc drain command, need to update drain_nodes() accordingly


ebenahar:
while running `oc drain` in our tests we get:
```
2022-01-23 21:48:17  19:48:16 - MainThread - ocs_ci.utility.utils - WARNING - Command stderr: Flag --delete-local-data has been deprecated, This option is deprecated and will be deleted. Use --delete-emptydir-data.
```
seems like we need to update ocs_ci.ocs.node.drain_nodes()

ebenahar:
fixed with https://github.com/red-hat-storage/ocs-ci/pull/5387#pullrequestreview-860803439

jimdevops19:
@ebenahar Can you share the execution you've faced this warning, please? So I can take a look to the params.
As @petr-balogh mentioned in the PR, it might be worth checking if we need to condition the flag change depending on the oc version.

petr-balogh:
@jimdevops19 you can try download the oc client for specific OCP version https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/ and you can check help command for that: e.g.
`oc adm drain --help`

You can check for more versions to see where it changed.

ebenahar:
it is not depending on the oc version but rather on ODF.

agarwal-mudit:
> it is not depending on the oc version but rather on ODF. I don't really know starting which version it should take effect. I see there is a previous issue about it - #4659 @agarwal-mudit any idea?

This is an oc command, code for oc command is with OCP team. I am not sure if this has anything to do with ODF.
You need to check with maszulik@redhat.com (I don't know his github id)

jimdevops19:
After checking in some clusters in different versions, I found out it is deprecated in 4.7 and above. I'll change my PR accordingly.

PR #5387, opened 2022-01-24
Update flag of --delete-local-data in all locations since it's deprecated

jimdevops19:
Fixes: #5386

### DIFF
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -14,6 +14,7 @@ from ocs_ci.deployment.ocp import OCPDeployment as BaseOCPDeployment
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants, exceptions, ocp, machine
 from ocs_ci.ocs.resources import pod
+from ocs_ci.ocs.node import drain_nodes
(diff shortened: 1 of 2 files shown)


ebenahar (dismissed):

prsurve (dismissed):

petr-balogh:
@jimdevops19 added description `Fixes #5386` then the issue will be automatically closed.
Can you please update also commit message in the commit please  please with `git commit --amend` to add the same line about `Fixes: #5386`?

ocs_ci/deployment/aws.py
```diff
@@ -631,7 +631,7 @@ def remove_rhcos_workers(self):
             cordon = f"oc adm cordon {node}"
             run_cmd(cordon)
             drain = (
-                f"oc adm drain {node} --force --delete-local-data "
+                f"oc adm drain {node} --force --delete-emptydir-data "
```
petr-balogh, line 634: `--delete-emptydir-data` From which version is it deprecated and from which version of oc is this supported?

If there are some versions then we should make this based on some condition I guess.

petr-balogh (approved):

AviadP (approved):

PR #5390, opened 2022-01-25
Dashboard Integration for test_pvc_clone_performance

pintojoy:
In this PR test_pvc_clone_performance_results are pushed into ES server and dashboard. Also base class is changed from E2Etest to PASTest

### DIFF
--- a/tests/e2e/performance/csi_tests/test_pvc_clone_performance.py
+++ b/tests/e2e/performance/csi_tests/test_pvc_clone_performance.py
@@ -4,63 +4,61 @@ Performance is measured by collecting clone creation/deletion speed.
 """
 import logging
 import pytest
+import os
+from uuid import uuid4
+import statistics
 
 from ocs_ci.ocs import constants
-from ocs_ci.framework.testlib import skipif_ocs_version, performance, E2ETest
-from ocs_ci.ocs.resources import pvc
+from ocs_ci.framework.testlib import performance
+from ocs_ci.framework import config
 from ocs_ci.helpers import helpers, performance_lib
 from ocs_ci.utility.utils import convert_device_size
+from ocs_ci.ocs.perfresult import ResultsAnalyse
+from ocs_ci.helpers.helpers import get_full_test_logs_path
+from ocs_ci.ocs.perftests import PASTest
+from ocs_ci.ocs.resources import pvc
 
 logger = logging.getLogger(__name__)
 
 
-@skipif_ocs_version("<4.6")
-@pytest.mark.parametrize(
-    argnames=["interface_type", "pvc_size", "file_size"],
-    argvalues=[
-        pytest.param(
-            *[constants.CEPHBLOCKPOOL, "1", "600Mi"],
-            marks=pytest.mark.polarion_id("OCS-2356"),
-        ),
-        pytest.param(
-            *[constants.CEPHBLOCKPOOL, "25", "15Gi"],
-            marks=pytest.mark.polarion_id("OCS-2340"),
-        ),
-        pytest.param(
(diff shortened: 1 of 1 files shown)


tests/e2e/performance/csi_tests/test_pvc_clone_performance.py
```diff
@@ -203,4 +251,80 @@ def test_clone_create_delete_performance(
...
+        and reporting the full results (links in the ES) of previous tests (4)
+        """
+        self.number_of_tests = 8
+        self.results_path = get_full_test_logs_path(
+            cname=self, fname="test_clone_create_delete_performance"
+        )
+        self.results_file = os.path.join(self.results_path, "all_results.txt")
+        logger.info(f"Check results in {self.results_file}.")
+        self.check_tests_results()
+        self.push_to_dashboard(test_name="pvc clone performance")
```
ypersky1980, line 314: I'd name the test : "PVC Clone Performance"  with Caps lock - similar to the convention of other tables names.

ypersky1980 (dismissed):

tests/e2e/performance/csi_tests/test_pvc_clone_performance.py
pintojoy: Thanks Yuli. I made that change

ypersky1980 (approved):
LGTM, thank you!

tests/e2e/performance/csi_tests/test_pvc_clone_performance.py
```diff
@@ -203,4 +251,80 @@ def test_clone_create_delete_performance(
...
+            full_results (obj): an FIOResultsAnalyse object
+
+        Returns:
+            FIOResultsAnalyse (obj): the input object fill with data
+
+        """
+        for key in self.environment:
+            full_results.add_key(key, self.environment[key])
+        full_results.add_key("index", full_results.new_index)
+        return full_results
```
ramkiperiy, line 330: Can this function be moved to generic library file?

ramkiperiy: @Avilir @ypersky1980 @pintojoy ^^ if its more specific to test-case then ok with it but all the duplicate function do have same lines of codes.

ypersky1980: @pintojoy @ramkiperiy I think that we have to have this def init_full_results(self, full_results) in each one of our tests - this is initialization of the results obj.

ypersky1980: we do have a generic class ResultsAnalyse(PerfResult): in ocs-ci/ocs_ci/ocs/perfresult.py /

https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/perfresult.py#L163

ramkiperiy: @ypersky1980 yes, even my point was same why can't we add this function `init_full_results` in the class `class ResultsAnalyse(PerfResult):` and call it in each test script ? 
@Avilir WDYT about this ?

Avilir: @ramkiperiy , the reason for this object is to hendale the results, calculating average , summarize results etc.

Avilir (approved):
LGTM

tests/e2e/performance/csi_tests/test_pvc_clone_performance.py
ramkiperiy: As discussed in the meeting, let's have this tracked part of code optimization and reducing code duplicates.

ramkiperiy (approved):

PR #5424, opened 2022-02-04
Add RHEL8 repos

petr-balogh:
Also adding support for changing between RHEL 7 or RHEL8.
For now, the ansible plabook for adding RHEL 8 nodes is documented
to run on RHEL 7 node. So we need to keep both repos for now.

Also the function for creating RHEL pod is not parametrized to create
different version of RHEL than RHEL 7. Once we will need to move to
RHEL 8 pod, the function needs to be updated as well.


shylesh (approved):
LGTM

ocs_ci/deployment/aws.py
```diff
@@ -576,7 +591,7 @@ def run_ansible_playbook(self):
             rhel_pod_obj.exec_cmd_on_node(
                 host,
                 pem_dst_path,
-                f'sudo mv {os.path.join("/tmp", repo_file)} {repo_dst_path}',
+                f'sudo mv {os.path.join("/tmp", repo_file)} {constants.YUM_REPOS_PATH}',
```
dahorak, line 594: The `/tmp` here and few lines above could be probably replaced by constants `POD_UPLOADPATH` or `RHEL_TMP_PATH`, but I think it is not necessary to change it in this PR.

ocs_ci/ocs/platform_nodes.py
```diff
@@ -922,7 +937,7 @@ def attach_rhel_nodes_to_upi_cluster(self, node_list):
             rhel_pod_obj.exec_cmd_on_node(
                 host,
                 pem_dst_path,
-                f'sudo mv {os.path.join("/tmp", repo_file)} {repo_dst_path}',
+                f'sudo mv {os.path.join("/tmp", repo_file)} {constants.YUM_REPOS_PATH}',
```
dahorak, line 940: Same as above, the `/tmp` here and few lines above could be probably replaced by constants `POD_UPLOADPATH` or `RHEL_TMP_PATH` (but I'm ok to keep it as it is in this PR, if there will be no other changes).

ocs_ci/repos/ocp_4_9_rhel8.repo
```diff
@@ -0,0 +1,35 @@
...
+baseurl=https://mirror.openshift.com/enterprise/reposync/ci-deps/rhel-8-appstream-rpms/
+enabled=1
+gpgcheck=0
+username=$mirror_openshift_user
+password=$mirror_openshift_password
+module_hotfixes = 1
+
+[rhel-fast]
+name=RHfast
+baseurl=https://mirror.openshift.com/enterprise/reposync/4.10/rhel-8-fast-datapath-rpms/
```
dahorak, line 30: The `4.10` version here in `4.9` repo is intentional?
(I see that `https://mirror.openshift.com/enterprise/reposync/4.9/rhel-8-fast-datapath-rpms/` also exists.)

dahorak (approved):
The open question is the `4.10` repo url in `4.9` repo file - if there is reason for that, I'm ok with merging this PR.
The other comments about constants for `/tmp` are only just in case, that you will change something else in this PR, that it will be good to fix them too.

dahorak (approved):

vavuthu (approved):

PR #5427, opened 2022-02-04
Use default storage class in test test_change_reclaim_policy_of_pv if the reclaim policy is Delete

jilju:
Use the default storage class in the test test_change_reclaim_policy_of_pv if the reclaim policy is "Delete". New storage class should be created only if the reclaim policy is "Retain". This will help running the tests which require the reclaim policy "Delete" on the platforms where the creation of new storage class is not supported/recommended. 


nehaberry (approved):
Changes LGTM. This will enable some MS tests to use deffault SC too.

PrasadDesala (approved):

PR #5436, opened 2022-02-07
Fix ROSA deployment and pagerduty service creation

fbalak:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -383,33 +383,33 @@ def pagerduty_service(request):
         str: PagerDuty service json
 
     """
-    if config.ENV_DATA["platform"].lower() in constants.MANAGED_SERVICE_PLATFORMS:
-        pagerduty_api = pagerduty.PagerDutyAPI()
-        payload = pagerduty_api.get_service_dict()
-        service_response = pagerduty_api.create("services", payload=payload)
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -412,7 +412,7 @@ def teardown():
         return None
 
 
-# @pytest.fixture(scope="session", autouse=True)
+@pytest.fixture(scope="session", autouse=True)
```
mbukatov, line 415: If you want to have it autouse enabled by default, let's either mark it by appropriate skip marker, or add statement like this in the beggining of the fixture:
```
 if config.ENV_DATA["platform"].lower() not in constants.MANAGED_SERVICE_PLATFORMS:
    return
```
To make it clear that it won't do anything on non manged service platforms.

Having a quick if statement which ends a function is preferable over large if blocks, since one can easily miss when these block ends ...

mbukatov (changes requested):
Asking for a simple fix.

mbukatov (approved):

fbalak:
Verified: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/9841

mbukatov (approved):

dahorak (approved):

PR #5440, opened 2022-02-08
Import Clusters to ACM Hub including a separated test mark

jimdevops19:
Signed-off-by: Cem Erginbas <cerginba@cerginba.tlv.csb>
Changes that are required for the importing scenario.
Some of the changes:
 - Added a new mark of `acm_import` as it will be a good solution for skipping pytest auto-use calls when entering run-ci
 - Added a new config file `conf/ocsci/import_clusters_to_acm.yaml` that skips ocs deployment and configures import_clusters_to_acm=True
 - Bug fixes in `ocsci/ocs/acm/acm.py` file that were required for the execution of the UI automation successfully to import the clusters
 - in `ocs_ci/ocs/ui/base_ui.py`  added chrome option of: `--disable-dev-shm-usage` when we pass the import_clusters_to_acm configuration. In such a scenario, the kubeconfig needs to be copied and sent as a text via UI automation. Since kubeconfig is a large file, Selenium crashes while trying to send those keys.
 - In `test_acm.py` file - the calls for submariner installation are deleted. We only want to handle import clusters for now.

### DIFF
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -342,7 +342,6 @@ def pytest_configure(config):
             # Add OCS related versions to the html report and remove
             # extraneous metadata
             markers_arg = config.getoption("-m")
-
             # add logs url
             logs_url = ocsci_config.RUN.get("logs_url")
             if logs_url:
@@ -355,6 +354,12 @@ def pytest_configure(config):
                 log.info(
                     "Skipping versions collecting because: Deploy or destroy of "
                     "cluster is performed."
+                )
+                continue
+            elif markers_arg == "acm_import":
+                log.info(
+                    "Skipping auto pytest executions and version collecting because "
+                    "Import Clusters to ACM is performed."
                 )
                 continue
             elif ocsci_config.ENV_DATA["skip_ocs_deployment"]:
--- a/ocs_ci/ocs/acm/acm.py
+++ b/ocs_ci/ocs/acm/acm.py
@@ -1,5 +1,6 @@
 import logging
 import time
+import os
 
 from selenium.webdriver.support.wait import WebDriverWait
 from selenium.webdriver.support import expected_conditions as ec
@@ -11,7 +12,7 @@ from ocs_ci.ocs.acm.acm_constants import (
     ACM_PAGE_TITLE,
 )
 from ocs_ci.ocs.ocp import OCP
-from ocs_ci.framework import MultiClusterConfig
+from ocs_ci.framework import config
 from ocs_ci.ocs.ui.helpers_ui import format_locator
 from ocs_ci.utility.utils import TimeoutSampler
 from ocs_ci.ocs.ui.acm_ui import AcmPageNavigator
@@ -48,8 +49,6 @@ class AcmAddClusters(AcmPageNavigator):
         self.do_click(self.page_nav["choose_kubeconfig"])
         log.info(f"Coping Kubeconfig {kubeconfig_location}")
         kubeconfig_to_import = copy_kubeconfig(kubeconfig_location)
-        log.info(kubeconfig_to_import)
-        self.do_click(self.page_nav["Kubeconfig_text"])
         for line in kubeconfig_to_import:
             self.do_send_keys(self.page_nav["Kubeconfig_text"], text=f"{line}")
             time.sleep(2)
@@ -63,6 +62,8 @@ class AcmAddClusters(AcmPageNavigator):
         Args:
             cluster_name: (str): cluster name to import
             kubeconfig_location: (str): kubeconfig location
+        Returns:
+            None, but exits if sample object is not None using TimeoutSampler
 
         """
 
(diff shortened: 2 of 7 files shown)


tests/ecosystem/deployment/test_acm.py
```diff
@@ -1,14 +1,15 @@
...
 ####################################################################################################
 # This file is placeholder for calling import ACM as test, until full solution will be implimented #
 ####################################################################################################
 
 
-def test_import_acm(setup_acm_ui):
-    acm_obj = AcmAddClusters(setup_acm_ui)
+@acm_import
+def test_acm_import(setup_acm_ui):
+    AcmAddClusters(setup_acm_ui)
```
OdedViner, line 12: I dont understand this test file...

tests/ecosystem/deployment/test_acm.py
```diff
@@ -1,14 +1,15 @@
...
     import_clusters_with_acm()
```
dahorak, line 13: This looks little bit strange - because there is instantiated the class `AcmAddClusters` here, but didn't stored in any variable and then there is the call of `import_clusters_with_acm()` function, which instantiated the class `AcmAddClusters` again and uses it.

ocs_ci/ocs/acm/acm.py
```diff
@@ -369,12 +366,12 @@ def import_clusters_with_acm():
...
-    kubeconfig_a = clusters_env.get("kubeconfig_cl_a_location")
-    kubeconfig_b = clusters_env.get("kubeconfig_cl_b_location")
+    kubeconfig_a = clusters_env.get("kubeconfig_location_c1")
+    kubeconfig_b = clusters_env.get("kubeconfig_location_c2")
     global cluster_name_a
-    cluster_name_a = clusters_env.get("cl_a_cluster_name")
+    cluster_name_a = clusters_env.get("cluster_name_1")
     global cluster_name_b
-    cluster_name_b = clusters_env.get("cl_b_cluster_name")
+    cluster_name_b = clusters_env.get("cluster_name_2")
```
dahorak, line 374: This doesn't look much scalable, but I understand that it was not introduced in this PR, but was already in the previous code. Also the usage of `global` is not nice in this context I think, but the same - it is not part of this PR.

tests/ecosystem/deployment/test_acm.py
jimdevops19: Yes, this was a forgotten line to delete since some of its methods used to be called here. We only need the `import_clusters_with_acm()` with the instance created there. Since it assigns the selenium driver to the obj.

ocs_ci/ocs/acm/acm.py
jimdevops19: I am not a big fan of `global` as well but I think we should address it as a follow-up

ocs_ci/ocs/acm/acm.py
```diff
@@ -77,6 +76,7 @@ def import_cluster(self, cluster_name, kubeconfig_location):
         ):
             if sample:
                 log.info(f"Cluster: {cluster_name} successfully imported")
+                return True
```
annagitel, line 79: if you added  `return True` in the function, could you add `return False` after it? just so the function will return bool argument

annagitel: and please add `Returns: ` section in the docs

ocs_ci/ocs/acm/acm.py
```diff
@@ -332,30 +334,25 @@ def validate_cluster_import(cluster_name):
             dict_status.get("status") == "True"
         ), f"Status is not True, but: {dict_status.get('status')}"
 
+        # Return true if Assertion error was not raised:
+        return True
```
annagitel, line 338: same as previous comment -True/False

tests/ecosystem/deployment/test_acm.py
jimdevops19: This is the file for the ACM testing, we will include most likely functions that you want to execute after ACM is installed. Since it is inside the deployment parent directory, the functions here will execute deployment alike code, like the import scenario that is covered in this PR.

ocs_ci/ocs/acm/acm.py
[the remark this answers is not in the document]
jimdevops19: Thanks, missed this

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -357,6 +356,11 @@ def pytest_configure(config):
                     "cluster is performed."
                 )
                 continue
+            elif markers_arg == "acm_import":
+                log.info(
+                    "Skipping auto pytest executions and version collecting because "
+                    "Import Clusters to ACM is performed."
+                )
```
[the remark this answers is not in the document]
jimdevops19: Yes, I agree I should try that.

ocs_ci/ocs/acm/acm.py
jimdevops19: Good point. The goal is to get out of the loop once the sampler is not None so I'll try `return` first.

petr-balogh (approved):

petr-balogh:
@jimdevops19 please open issue for documentation.

jimdevops19:
> @jimdevops19 please open issue for documentation.

Done

OdedViner (approved):

clacroix12 (approved):

petr-balogh (approved):

PR #5443, opened 2022-02-08
Add WebDriver Data type hint in UI Base Class

jimdevops19:
Signed-off-by: Cem Erginbas <cerginba@cerginba.tlv.csb>

Motivation:
For future upcoming UI tests, it is beneficial to have auto completions on your IDEs. The child classes are not aware that `self.driver` is actually a WebDriver object.
By using this type-hint, we will see all the available methods no matter which classes you work on that inherits `BaseUI`

Please see the difference in those examples:

Example without the `WebDriver` type hint:
bad

Example with the `WebDriver` type hint:
good

### DIFF
--- a/ocs_ci/ocs/ui/base_ui.py
+++ b/ocs_ci/ocs/ui/base_ui.py
@@ -11,6 +11,7 @@ from selenium.common.exceptions import (
     WebDriverException,
     NoSuchElementException,
 )
+from selenium.webdriver.remote.webdriver import WebDriver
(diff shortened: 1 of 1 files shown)


OdedViner (approved):

am-agrawa (approved):
LGTM, but one thing to notice here is that we have most of the customized functions in the ocs_ci/ocs/ui/base_ui.py file and we recommend using them until possible.

petr-balogh (approved):

PR #5448, opened 2022-02-09
Handle webhook errors during store deletion

Neon-White:
Since admission control webhooks were added, the deletion of stores is sometimes blocked to the time it takes for buckets above them to fully delete.
This code handles the webhook error and continues to try.

### DIFF
--- a/ocs_ci/ocs/resources/backingstore.py
+++ b/ocs_ci/ocs/resources/backingstore.py
@@ -84,12 +84,13 @@ class BackingStore:
         def _oc_deletion_flow():
             try:
                 OCP(
-                    kind="backingstore", namespace=config.ENV_DATA["cluster_namespace"]
+                    kind=constants.BACKINGSTORE,
+                    namespace=config.ENV_DATA["cluster_namespace"],
                 ).delete(resource_name=self.name)
                 return True
             except CommandFailed as e:
                 if "not found" in e.args[0].lower():
-                    log.warning(f"Namespacestore {self.name} was already deleted.")
+                    log.warning(f"Backingstore {self.name} was already deleted.")
                     return True
                 elif all(
                     err in e.args[0]
--- a/ocs_ci/ocs/resources/namespacestore.py
+++ b/ocs_ci/ocs/resources/namespacestore.py
@@ -103,7 +103,7 @@ class NamespaceStore:
         if self.method == "oc":
             try:
                 OCP(
-                    kind="namespacestore",
+                    kind=constants.NAMESPACESTORE,
                     namespace=config.ENV_DATA["cluster_namespace"],
                     resource_name=self.name,
                 ).get()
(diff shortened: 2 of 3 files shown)


MeridianExplorer:
The verification job is still underway, but a few tests have already passed, thus marking it as verified.

parikshithb (dismissed):

ocs_ci/ocs/resources/backingstore.py
```diff
@@ -74,33 +81,59 @@ def delete(self):
...
-            ).delete(resource_name=self.name)
-        elif self.method == "cli":
-
-            def _cli_deletion_flow():
-                try:
-                    self.mcg_obj.exec_mcg_cmd(f"backingstore delete {self.name}")
+        def _oc_deletion_flow():
+            try:
+                OCP(
+                    kind="backingstore", namespace=config.ENV_DATA["cluster_namespace"]
```
ebenahar, line 87: please use constants.BACKINGSTORE instead of "backingstore"

ocs_ci/ocs/resources/backingstore.py
```diff
@@ -74,33 +81,59 @@ def delete(self):
...
+                if "not found" in e.args[0].lower():
+                    log.warning(f"Namespacestore {self.name} was already deleted.")
                     return True
-                except CommandFailed as e:
-                    if "being used by one or more buckets" in str(e).lower():
+                elif all(
+                    err in e.args[0]
+                    for err in ["cannot complete because pool", "in", "state"]
+                ):
+                    if retry:
```
ebenahar, line 98: seems strange. Why do we do this for loop?

ocs_ci/ocs/resources/namespacestore.py
```diff
@@ -31,50 +31,71 @@ def __init__(
...
-                try:
-                    self.mcg_obj.exec_mcg_cmd(f"namespacestore delete {self.name}")
                     return True
-                except CommandFailed as e:
-                    if "being used by one or more buckets" in str(e).lower():
+                elif all(
+                    err in e.args[0]
+                    for err in ["cannot complete because pool", "in", "state"]
+                ):
+                    if retry:
```
ebenahar, line 59: same

ocs_ci/ocs/resources/backingstore.py
MeridianExplorer: The for is internal to the `all()` creation - it checks that _all_ the strings in the list are also inside the error.
Alternatively, you can imagine it as:
` if all of ["cannot complete because pool", "in", "state"] is in the error message`

ocs_ci/ocs/resources/namespacestore.py
MeridianExplorer: [See comment](https://github.com/red-hat-storage/ocs-ci/pull/5448#discussion_r802585721)

bernerhat (approved):

Issue #5283, opened 2021-12-27
[Tracker] Testing suites needed changes 


ebenahar:
- [x] `flowtests` tier to be removed as it's not being executed. All its existing and future test cases to be placed in other tiers
- [x] `TestRGWAndKafkaNotifications` to be moved out of tier1 as it's not a product basic functionality 
- [x] `tests.manage.mcg.test_bucket_deletion.TestBucketDeletion.test_s3_bucket_delete_1t_objects` - Currently, not part of any tier
- [ ] `TestS3BucketPolicy` - all test cases in this test class, but `test_basic_bucket_policy_operations` and maybe `test_public_website`, to be moved out of tier1 as they don't seem to be about basic product functionality 
- [x] `tests.manage.mcg.test_bucket_replication.TestReplication.test_unidirectional_bucket_object_change_replication` to be moved out of tier1 as it doesn't seem to be about basic product functionality 
- [ ] `TestEndpointAutoScale` - currently skipped on all versions but 4.5 and 4.6
- [ ] Missing test cases in tier1 - OS node restart
- [x] `tests.manage.mcg.test_multicloud.TestMultiCloud` - keep only 1-2 test cases in tier1 and move the rest out to tier2
- [ ] `tests.manage.mcg.test_multipart_upload.TestS3MultipartUpload` - move from tier1 to tier2 as it's not about the product basic functionality
- [x] `tests.manage.mcg.test_namespace_crd.TestNamespace.test_distribution_of_objects_in_ns_bucket_crd` - seems negative and not functional, should be placed in tier3
- [x] `test_check_multi_object_integrity` to be moved out of tier1 as we already have `test_check_object_integrity`  in tier1
- [x] `test_mcg_data_deduplication` is not about basic product functionality, to be moved out of tier1
- [x] `test_mcg_data_compression` is not about basic product functionality, to be moved out of tier1
- [x] `TestSnapshotAtDifferentPvcUsageLevel.test_snapshot_at_different_usage_level` - to be moved out of tier1 as it's not about basic product functionality, it's more complex
- [ ] `tests.manage.rgw.test_multipart_upload.TestS3MultipartUpload.test_multipart_upload_operations` - doesn't seem to be basic product functionality, to be moved out of tier1

ebenahar:
Hi @MeridianExplorer @mbukatov @PrasadDesala @prsurve @fbalak @ramkiperiy please take a look

ebenahar:
@jilju

PR #5464, opened 2022-02-14
Flowtest marker removal

bernerhat:
remove flowtest marker as part of #5283


tests/e2e/flowtest/test_base_operation_node_drain.py
```diff
@@ -20,7 +20,7 @@
 
 
 @ignore_leftovers
-@flowtests
+@tier4b
```
parikshithb, line 23: @bernerhat Can we skip adding tier for this test for now? It is a bit complicated and has been a while since this test has been executed, needs to be stabilized else it will break the entire cluster.

@PrasadDesala

PR #5472, opened 2022-02-15
Applies huge pages and verifies cluster pods are running

parikshithb:
Signed-off-by: Parikshith Byregowda <pbyregow@redhat.com>


tests/manage/z_cluster/test_hugepages.py
```diff
@@ -0,0 +1,99 @@
...
+    tier2,
+)
+
+log = logging.getLogger(__name__)
+
+
+@tier2
+@skipif_external_mode
+@skipif_ocs_version("<4.8")
+@bugzilla("1995271")
```
PrasadDesala, line 32: the marker is added twice with the same BZ. Remove one

tests/manage/z_cluster/test_hugepages.py
```diff
@@ -0,0 +1,99 @@
...
+    ignore_leftovers,
+    E2ETest,
+    tier2,
+)
+
+log = logging.getLogger(__name__)
+
+
+@tier2
+@skipif_external_mode
```
PrasadDesala, line 30: We can configure hugepages on openshift nodes and then are consumed by apps. I think it should valid even for external mode, any reason for skipping the test to run on external mode?

tests/manage/z_cluster/test_hugepages.py
```diff
@@ -0,0 +1,99 @@
...
+@skipif_ocs_version("<4.8")
+@bugzilla("1995271")
+@bugzilla("1995271")
+@pytest.mark.polarion_id("OCS-2754")
+@ignore_leftovers
+class TestHugePages(E2ETest):
+    """
+    Enable huge pages post cluster deployment
```
PrasadDesala, line 38: Please make it clear that in this test, huge pages is enabled post ODF installation to avoid confusion as huge pages can also be enabled immediate after ocp cluster deployment.

tests/manage/z_cluster/test_hugepages.py
```diff
@@ -0,0 +1,99 @@
...
+            Removes huge pages on worker nodes and verifies all pods are up
+
+            """
+            disable_huge_pages()
+
+            wait_for_nodes_status(status=constants.NODE_READY, timeout=600)
+
+            log.info("Wait for all pods to be in running state")
+            wait_for_pods_to_be_running(timeout=600)
+
```
PrasadDesala, line 61: Should we also add a verification step that huge pages is indeed disabled on the cluster by checking its page size?

PrasadDesala, line 61: In a failure case, any chances that the node can go to not ready state while enabling huge pages? 
If yes, I would suggest you to add node_restart_teardown fixture

tests/manage/z_cluster/test_hugepages.py
parikshithb: ack, removed skip

tests/manage/z_cluster/test_hugepages.py
parikshithb: updated

tests/manage/z_cluster/test_hugepages.py
parikshithb: Yup, makes sense. Added.

parikshithb: Not in my testing, it looks stable. But yeah there are chances of failures like n/w etc. Added this teardown

prsurve (approved):

PrasadDesala (approved):

PR #5473, opened 2022-02-15
Add marker for managed service provider and consumer

suchita-g:

### DIFF
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -207,14 +207,18 @@ managed_service_required = pytest.mark.skipif(
 )
 
 ms_provider_required = pytest.mark.skipif(
-    (config.ENV_DATA["platform"].lower() not in MANAGED_SERVICE_PLATFORMS)
-    and config.ENV_DATA["cluster_type"].lower() == "provider",
+    not (
+        config.ENV_DATA["platform"].lower() in MANAGED_SERVICE_PLATFORMS
+        and config.ENV_DATA["cluster_type"].lower() == "provider"
+    ),
     reason="Test runs ONLY on managed service provider cluster",
 )
 
 ms_consumer_required = pytest.mark.skipif(
-    (config.ENV_DATA["platform"].lower() not in MANAGED_SERVICE_PLATFORMS)
-    and config.ENV_DATA["cluster_type"].lower() == "consumer",
+    not (
+        config.ENV_DATA["platform"].lower() in MANAGED_SERVICE_PLATFORMS
+        and config.ENV_DATA["cluster_type"].lower() == "consumer"
+    ),
     reason="Test runs ONLY on managed service consumer cluster",
 )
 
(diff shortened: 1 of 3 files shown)


ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -239,6 +251,18 @@
     reason="Test will not run on Openshift dedicated cluster",
 )
 
+skipif_msprovider = pytest.mark.skipif(
```
ebenahar, line 254: to keep consistency with https://github.com/red-hat-storage/ocs-ci/pull/5473/files#diff-82668ea1726338960f1271091bdb155e25bf053bf5f744a46784269810e5f0b8R215, please call the var `skipif_ms_provider`

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -239,6 +251,18 @@
...
+    config.ENV_DATA["platform"].lower() in MANAGED_SERVICE_PLATFORMS
+    and config.ENV_DATA["cluster_type"].lower() == "provider",
+    reason="Test will not run on Managed service provider cluster",
+)
+
+skipif_msconsumer = pytest.mark.skipif(
```
ebenahar, line 260: same here, please call the var `skipif_ms_consumer`

suchita-g: modified as suggested

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -206,6 +206,18 @@
     reason="Test runs ONLY on OSD or ROSA cluster",
 )
 
+ms_provider_required = pytest.mark.skipif(
+    (config.ENV_DATA["platform"].lower() not in MANAGED_SERVICE_PLATFORMS)
```
ebondare, line 210: Is it possible that platform would NOT be managed service AND cluster type would be provider? Don't you want to skip the test if platform is not managed service OR cluster type != provider? Sorry If I misunderstand the logic.

suchita-g: Good catch. Thanks, @ebondare . Modified the conditons

tests/manage/z_cluster/cluster_expansion/test_node_expansion.py
```diff
@@ -6,12 +6,14 @@
     skipif_managed_service,
     skipif_flexy_deployment,
     skipif_ibm_flash,
+    skipif_ms_consumer,
```
ebondare, line 9: For the test to run on provider you'll need to remove skipif_managed_service mark, and for the test below as well.

suchita-g: Yes I am thinking about how to accommodate both so that it should accommodate skip on rosa/osd tier1 for converged mode and v2 managed service solution as well

ebondare (approved):
LGTM

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -206,6 +206,22 @@
...
+        config.ENV_DATA["platform"].lower() in MANAGED_SERVICE_PLATFORMS
+        and config.ENV_DATA["cluster_type"].lower() == "provider"
+    ),
+    reason="Test runs ONLY on managed service provider cluster",
+)
+
+ms_consumer_required = pytest.mark.skipif(
+    not (
+        config.ENV_DATA["platform"].lower() in MANAGED_SERVICE_PLATFORMS
+        and config.ENV_DATA["cluster_type"].lower() == "consumer"
```
fbalak, line 220:
```suggestion
        and config.ENV_DATA.get("cluster_type", "").lower() == "consumer"
```
"platform" key should be part of all configs but "cluster_type" might not be set (not managed platforms or converged mode). This can result in a key error.

suchita-g: I think not. Define this to default config will take care of this.  https://github.com/red-hat-storage/ocs-ci/blob/8f806f709e8190d545ddb924723364f4c9c9d031/ocs_ci/framework/conf/default_config.yaml#L198

fbalak: Ok, in that case it could work. I just recommend to run TestAddNode test case also at not managed service platform as a verification job to be sure.

fbalak (approved):

yitzhak12 (approved):

ebenahar (approved):

PR #5477, opened 2022-02-15
Extend the MCG OBC health check timeout post upgrade

MeridianExplorer:
Currently, the tests time out due to a race condition of a few seconds. The OBCs are bound by the time of the log collection, but not quickly enough to be caught by the default timeout period.
For an example, see the logs of run 1644190416.


parikshithb (approved):

keesturam (approved):
LGTM!

vavuthu (approved):

PR #5478, opened 2022-02-15
Migrate S3 CLI relay pod to an STS to withstand node drains

MeridianExplorer:
At the moment, certain node drain tests have lead to the AWSCLI session pod being removed.
Now, with the S3 CLI pod being replaced with an STS, it's going to respin automatically regardless of drains.

As part of this change, *all* MCG tests were refactored to use the session-scope S3 CLI pod, in order to both improve performance, and avoid STS naming conflicts.

Also removed several outdated tests that were already skipped for a while.

### DIFF
--- a/ocs_ci/templates/mcg/s3cli-sts.yaml
+++ b/ocs_ci/templates/mcg/s3cli-sts.yaml
@@ -1,0 +1,35 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  namespace: openshift-storage
+  name: s3cli
+spec:
+  selector:
+    matchLabels:
+      app: s3cli
+  replicas: 1
+  updateStrategy:
+    type: RollingUpdate
+  podManagementPolicy: OrderedReady
+  volumeClaimTemplates: []
+  template:
+    metadata:
+      labels:
+        app: s3cli
+    spec:
+      securityContext:
+        runAsUser: 0
+      volumes:
+        - name: service-ca
+          configMap:
+            name: awscli-service-ca
+      containers:
+        - name: s3cli
+          image: quay.io/ocsci/s3-cli-with-test-objects-multiarch:1.0
+          command: ['/bin/sh']
+          stdin: true
+          tty: true
+          volumeMounts:
+            - name: service-ca
+              mountPath: /cert/service-ca.crt
+              subPath: service-ca.crt
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -92,7 +92,6 @@ from ocs_ci.utility.utils import (
     ceph_health_check_base,
     get_ocs_build_number,
     get_openshift_client,
-    get_system_architecture,
     get_testrun_name,
     load_auth_config,
     ocsci_log_path,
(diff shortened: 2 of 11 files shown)


MeridianExplorer:
All modified e2e CRD tests were verified locally.
A rerun of the tests that failed in the original verification job is available [here](https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/10055/)

ocs_ci/templates/mcg/s3cli-sts.yaml
```diff
@@ -0,0 +1,35 @@
...
+    type: RollingUpdate
+  podManagementPolicy: OrderedReady
+  volumeClaimTemplates: []
+  template:
+    metadata:
+      labels:
+        app: s3cli
+    spec:
+      securityContext:
+        runAsUser: 0
```
mbukatov, line 21: Why do you need to run this as root?

ocs_ci/templates/mcg/s3cli-sts.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  namespace: openshift-storage
+  name: s3cli
+spec:
+  selector:
+    matchLabels:
+      app: s3cli
+  replicas: 1
```
mbukatov, line 10: Since this is StatefulSet with a single replica, does it mean that when a new pod is deployed (eg. because of the original pod was running on a broken node), it will create a new pod with a new name and "identity", or would it deploy pod with a new name?

mbukatov (commented):
This is a change in the right direction.

ocs_ci/templates/mcg/s3cli-sts.yaml
MeridianExplorer: I believe that the new pod will have different identity features (the only thing I can think of is the UID, but still), but since it's a StatefulSet, all pods are bound to use the same base name, with a `-{n}` suffix, where `n` is the replica ID.
i.e. - in this case, we'll always have a pod named `s3cli-0`. 
If the replica amount was higher, it'd be `s3cli-1`, s3cli-2`, etc.

tests/conftest.py
```diff
@@ -2148,34 +2150,31 @@ def awscli_pod_fixture(request, scope_name):
...
 
-    arch = get_system_architecture()
-    if arch.startswith("x86"):
-        pod_dict_path = constants.AWSCLI_POD_YAML
-    else:
-        pod_dict_path = constants.AWSCLI_MULTIARCH_POD_YAML
-
-    awscli_pod_dict = templating.load_yaml(pod_dict_path)
-    awscli_pod_dict["spec"]["volumes"][0]["configMap"][
+    awscli_sts_dict = templating.load_yaml(constants.S3CLI_MULTIARCH_STS_YAML)
```
ebenahar, line 2153: will S3CLI_MULTIARCH_STS_YAML be good for x86?

MeridianExplorer: Affirmative

parikshithb (dismissed):
lgtm

parikshithb (approved):

petr-balogh (approved):

tests/conftest.py
```diff
@@ -2186,16 +2185,19 @@ def finalizer():
...
-    return awscli_pod_fixture(request, scope_name="function")
-
-
 @pytest.fixture(scope="session")
 def awscli_pod_session(request):
     return awscli_pod_fixture(request, scope_name="session")
 
 
+@pytest.fixture(scope="session")
+def awscli_pod(request, awscli_pod_session):
```
mashetty330, line 2194: wasn't this supposed to be function scoped?

mashetty330:
@MeridianExplorer why do you have verified label? were you able to run PR validation successfully on this?

tests/conftest.py
mashetty330: okay!

mashetty330: sounds good to me as long as there are no dependent tests that only needs function scoped fixture!

mashetty330 (approved):

PR #5495, opened 2022-02-21
Wait for machineConfigPool after applying SSL cert

petr-balogh:

### DIFF
--- a/ocs_ci/utility/ssl_certs.py
+++ b/ocs_ci/utility/ssl_certs.py
@@ -11,7 +11,11 @@ from OpenSSL import crypto
 
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants, exceptions
-from ocs_ci.utility.utils import download_file, exec_cmd
+from ocs_ci.utility.utils import (
+    download_file,
+    exec_cmd,
+    wait_for_machineconfigpool_status,
+)
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


ocs_ci/utility/ssl_certs.py
```diff
@@ -266,6 +270,7 @@ def configure_custom_ingress_cert():
         '--type=merge -p \'{"spec":{"defaultCertificate": {"name": "ocs-cert"}}}\''
     )
     exec_cmd(cmd)
+    wait_for_machineconfigpool_status("all", timeout=1800)
```
prsurve, line 273: we need to sleep here for 10 sec and then we can start for wait_for_machineconfigpool_status

clacroix12: @prsurve why is the sleep necessary? / If it's always necessary should the `wait_for_machineconfigpool_status` function handle that wait time for us?

prsurve: It's because mcp takes few sec to update with new config

petr-balogh: I added it directly to the function in next commit and removed from all other places as I noticed we are waiting 60 seconds before calling this function on all the places.

vavuthu (approved):

jimdevops19 (approved):

PR #5498, opened 2022-02-22
Repair db_scc test to support 4.7.z service account

bernerhat:
version 4.7.z still uses "noobaa" SA instead of the new "noobaa-endpoint" account which causes the test to fail, this pr verifies the cluster's version and send it back to the test. this pr is to revert the deletion of this check function since 4.7.z has not received the backport with the SA name change
Signed-off-by: Amit Berner <aberner@redhat.com>


MeridianExplorer (approved):

jimdevops19 (approved):
LGTM

PR #5500, opened 2022-02-22
Fix UI deployment

OdedViner:

### DIFF
--- a/ocs_ci/ocs/ui/base_ui.py
+++ b/ocs_ci/ocs/ui/base_ui.py
@@ -177,6 +177,19 @@ class BaseUI:
         )
         return len(element_list) > 0
 
+    def get_element_by_xpath(self, xpath=""):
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/ui/base_ui.py
```diff
@@ -177,6 +177,19 @@ def check_element_text(self, expected_text, element="*"):
         )
         return len(element_list) > 0
 
+    def get_element_by_xpath(self, xpath=""):
```
jimdevops19, line 180: Should we call the method `get_elements_by_xpath` if the goal is to return a list of elements that match the xpath ?

petr-balogh:
@OdedViner is going to close this one as the origin issue suppose to be fixed in https://github.com/red-hat-storage/ocs-ci/pull/5495

PR #5509, opened 2022-02-24
changing compressed FIO test timeout

Avilir:
currently all FIO test on compressed volume with 16 KiB block size are not completed on time.
This PR is increasing the time of the test, so it will finishe.


pintojoy (approved):
LGTM

PR #5514, opened 2022-02-24
Updating tier definitions[tier4a]

Shrivaibavi:
As part of the "Improve task efficiency" workgroup, one of the KRs was to have lighter test regression suites so as to allow day2 operation tests to be run more frequently and the rest of the failure tests less frequently. 

So, the tier4a, tier4b, and tier4c tests will be classified with a clear and better definition as described below. 
Tier4a - Any failure that is introduced as part of day 2 operations such as device replacement, node replacement, etc
Tier4b - Any node related disruptions are grouped as tier4b
Tier4c - Any pod related disruptions are grouped as tier4c

### DIFF
--- a/tests/manage/mcg/test_host_node_failure.py
+++ b/tests/manage/mcg/test_host_node_failure.py
@@ -8,8 +8,7 @@ from ocs_ci.framework.testlib import (
     ManageTest,
     skipif_ocs_version,
     skipif_vsphere_ipi,
-    tier4,
-    tier4a,
+    tier4b,
 )
 from ocs_ci.helpers.sanity_helpers import Sanity, SanityExternalCluster
 from ocs_ci.helpers.helpers import (
@@ -29,8 +28,7 @@ from ocs_ci.utility.utils import TimeoutSampler
 log = logging.getLogger(__name__)
 
 
-@tier4
-@tier4a
+@tier4b
 @bugzilla("1853638")
 @ignore_leftovers
 @skipif_vsphere_ipi
--- a/tests/manage/mcg/test_mcg_resources_disruptions.py
+++ b/tests/manage/mcg/test_mcg_resources_disruptions.py
@@ -9,8 +9,7 @@ from ocs_ci.framework.testlib import (
     ignore_leftovers,
     on_prem_platform_required,
     skipif_ocs_version,
-    tier4,
-    tier4a,
+    tier4c,
     tier3,
     skipif_managed_service,
 )
@@ -30,8 +29,6 @@ def setup(request):
     request.cls.cl_obj = cluster.CephCluster()
 
 
-@tier4
-@tier4a
 @ignore_leftovers()
 @pytest.mark.usefixtures(setup.__name__)
 class TestMCGResourcesDisruptions(MCGTest):
@@ -52,6 +49,7 @@ class TestMCGResourcesDisruptions(MCGTest):
         "noobaa_operator": constants.NOOBAA_OPERATOR_POD_LABEL,
     }
 
+    @tier4c
     @pytest.mark.parametrize(
         argnames=["resource_to_delete"],
         argvalues=[
@@ -81,6 +79,7 @@ class TestMCGResourcesDisruptions(MCGTest):
         )
         self.cl_obj.wait_for_noobaa_health_ok()
 
+    @tier4c
     @skipif_ocs_version("<4.5")
     @on_prem_platform_required
     @pytest.mark.parametrize(
@@ -116,6 +115,7 @@ class TestMCGResourcesDisruptions(MCGTest):
         )
         self.cl_obj.wait_for_noobaa_health_ok()
 
+    @tier3
     @pytest.mark.parametrize(
         argnames=["pod_to_drain"],
         argvalues=[
--- a/tests/manage/mcg/test_multi_region.py
+++ b/tests/manage/mcg/test_multi_region.py
@@ -4,8 +4,7 @@ import pytest
 
 from ocs_ci.framework.pytest_customization.marks import (
     tier1,
-    tier4,
-    tier4a,
+    tier2,
     bugzilla,
     skipif_ocs_version,
     skipif_aws_creds_are_missing,
@@ -74,8 +73,7 @@ class TestMultiRegion(MCGTest):
             len(mirror_attached_pools) == 2
         ), "Multiregion bucket did not have two backingstores attached"
 
-    @tier4
-    @tier4a
+    @tier2
     @bugzilla("1827317")
     @skipif_ocs_version("==4.4")
     @pytest.mark.polarion_id("OCS-1784")
--- a/tests/manage/mcg/test_multi_region.py
+++ b/tests/manage/mcg/test_multi_region.py
@@ -4,7 +4,7 @@ import pytest
 
 from ocs_ci.framework.pytest_customization.marks import (
     tier1,
-    tier2,
+    tier4a,
     bugzilla,
     skipif_ocs_version,
     skipif_aws_creds_are_missing,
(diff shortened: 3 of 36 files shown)


tests/manage/mcg/test_host_node_failure.py
```diff
@@ -8,8 +8,7 @@
     ManageTest,
     skipif_ocs_version,
     skipif_vsphere_ipi,
-    tier4,
-    tier4a,
+    tier4b,
```
ebenahar, line 11: I think we still want to keep the `tier4` marker

Shrivaibavi: IMO the whole purpose of segregating the tier4 testcases as node failures, pod failures and day2 operations is to achieve less run time . if we include tier4 we might not achieve the less run time and neatly segregated testcases. 
@keesturam at any point of time do we prefer running @tier4 testcases without specifying tier4a, tier4b or tier4c ? WDYT ?

tests/manage/mcg/test_mcg_resources_disruptions.py
```diff
@@ -116,6 +115,7 @@ def test_scale_down_rgw(self, scale_down_to):
         )
         self.cl_obj.wait_for_noobaa_health_ok()
 
+    @tier3
```
MeridianExplorer, line 118: I do not think this is the correct tier for this test - it checks that all MCG pods manage to migrate to another node when the one they rely on is drained. I do not see this as a negative test - it's more of tier 2 or 4

tests/manage/mcg/test_multi_region.py
```diff
@@ -74,8 +73,7 @@ def test_multiregion_bucket_creation(self, mcg_obj, multiregion_mirror_setup):
             len(mirror_attached_pools) == 2
         ), "Multiregion bucket did not have two backingstores attached"
 
-    @tier4
-    @tier4a
+    @tier2
```
MeridianExplorer, line 76: This test does not fit in tier2; It checks that bucket mirroring policies work by simulating an underlying storage failure which takes one of the backingstores offline. It's related to tier 4.

MeridianExplorer (changes requested):

tests/manage/mcg/test_host_node_failure.py
keesturam: I believe the point Elad makes is that having tier4 marker would still allow us to run *all* tier4 tests without affecting tier4a, 4b, or 4c. The pipeline jbs would still use 4a, 4b and 4c markers.

ebenahar: correct

tests/manage/mcg/test_mcg_resources_disruptions.py
ebenahar: The reason for moving it to tier3 is that this test case is not a disruptive one. Drain operation should not cause down time

tests/manage/mcg/test_multi_region.py
ebenahar: this test indeed involves storage failures but the reason for not keeping it in tier4 is that the storage failure not supposed to cause down time and this storage failure triggers a mechanism that we consider as a product feature

tests/manage/mcg/test_mcg_resources_disruptions.py
MeridianExplorer: Resolving after re-reading tier3's definition

tests/manage/mcg/test_multi_region.py
MeridianExplorer: According to the tier 4 definition in `OCS Automation by tiers - guidelines`, this seems to fit perfectly into tier 4. Downtime does not constitute a part of the definition. If it's supposed to, I think the document should be updated to reflect that.

The mirroring is being performed and checked regardless of, and before any failure presented -
https://github.com/red-hat-storage/ocs-ci/blob/ef8d8a2af4b092b85320a2dc2335993a71b00cce/tests/manage/mcg/test_multi_region.py#L115-L117
The failure is not required, and is performed purely as a means of achieving deeper, more "real-life" coverage.

I still think that this test fits best into tier 4, but in any other case, 3 would prolly fit better than 2

ebenahar: I am ok either way so we can keep it as is. @Shrivaibavi @keesturam what do you think?

tests/manage/z_cluster/nodes/test_nodes_restart.py
```diff
@@ -77,6 +76,7 @@ def test_nodes_restart(
             pvc_factory, pod_factory, bucket_factory, rgw_bucket_factory
         )
 
+    @tier4b
```
sidhant-agrawal, line 79: There is `tier4a` marker at L#27, wouldn't this test be considered as both tier4a and tier4b in that case?

sidhant-agrawal:
@Shrivaibavi Some of tests which in g-sheet, are not updated in this PR.
For eg. tests in `test_namespace_crd.py` , `test_nodes_restart.py` and few others. Pls recheck.

tests/manage/z_cluster/nodes/test_nodes_restart.py
Shrivaibavi: IMO I dont think the test will be considered as both tier4a and tier4b, Wont it override ?? @petr-balogh please correct me if i am wrong

tests/manage/mcg/test_multi_region.py
keesturam: I am okay to have this in tier4a too.

tests/manage/z_cluster/nodes/test_nodes_restart.py
petr-balogh: no it will not overwrite. Please remove it from class level and specify directly for each test case in that class.  This will really lead that the test will be tier4a and tier4b as well if not changed.

Shrivaibavi: Added them separately under different tests and removed it in class

tests/manage/monitoring/prometheus/test_noobaa.py
```diff
@@ -89,8 +87,7 @@ def test_noobaa_bucket_quota(measure_noobaa_exceed_bucket_quota):
         )
 
 
-@tier4
-@tier4a
```
fbalak, line 93: I think that this is a disruptive test that causes an error in NooBaa. It makes Namespace bucket broken and checks that proper messages are raised. I think that this qualifies as tier4a.

fbalak (changes requested):
`tests/manage/monitoring/` based tests reviewed. Looks ok except one marker.

tests/manage/monitoring/prometheus/test_noobaa.py
Shrivaibavi: tier4a is related tp day 2 operations like node replacement, device replacement kind of things. I dont think this fits in tier4a. Its neither node failure nor pod failure. Not sure if bucket deletion falls under tier4a.

mbukatov (commented):
Looks ok from blue team perspective, with exception of the Filip's point. Keeping not approved to make sure Filip's remark get's resolved.

tests/manage/monitoring/prometheus/test_noobaa.py
Shrivaibavi: Changed into tier4a

MeridianExplorer (dismissed):

Shrivaibavi:
> @Shrivaibavi Some of tests present in g-sheet, are not updated in this PR. For eg. tests in `test_namespace_crd.py` , `test_nodes_restart.py` and few others. Pls recheck.

DOne

MeridianExplorer (approved):

sidhant-agrawal (approved):

fbalak (approved):

ebenahar (approved):

ebenahar:
@Shrivaibavi Please trigger one tier4a run for verification

ebenahar (approved):

PR #5515, opened 2022-02-24
In the test 'automated recovery from failed nodes' get the node in the same rack/zone

yitzhak12:


ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -222,6 +222,11 @@
     reason="Test runs ONLY on managed service consumer cluster",
 )
 
+ipi_or_managed_deployment_required = pytest.mark.skipif(
```
ebenahar, line 225: Since there is no tight connection between IPI and managed service, please split into 2 decorators - one for IPI and one for managed service

ocs_ci/ocs/node.py
```diff
@@ -1995,3 +1995,35 @@ def add_new_nodes_and_label_after_node_failure_ipi(
     """
     machine.change_current_replica_count_to_ready_replica_count(machineset_name)
     return add_new_node_and_label_it(machineset_name, num_nodes, mark_for_ocs_label)
+
+
+def get_other_worker_nodes_in_same_rack_or_zone(failure_domain, node_obj):
+    """
+    Get other worker nodes in the same rack or zone of a given node.
```
ebenahar, line 2002: Args
Returns

PR #5520, opened 2022-02-28
LSO 4.10 Deployment via UI

OdedViner:

### DIFF
--- a/ocs_ci/deployment/helpers/lso_helpers.py
+++ b/ocs_ci/deployment/helpers/lso_helpers.py
@@ -42,7 +42,8 @@ def setup_local_storage(storageclass):
     ocp_version = version.get_semantic_ocp_version_from_config()
     ocs_version = version.get_semantic_ocs_version_from_config()
     ocp_ga_version = get_ocp_ga_version(ocp_version)
-    create_optional_operators_catalogsource_non_ga()
+    if not ocp_ga_version:
+        create_optional_operators_catalogsource_non_ga()
 
     logger.info("Retrieving local-storage-operator data from yaml")
     lso_data = list(
--- a/ocs_ci/deployment/helpers/lso_helpers.py
+++ b/ocs_ci/deployment/helpers/lso_helpers.py
@@ -207,60 +207,63 @@ def setup_local_storage(storageclass):
     verify_pvs_created(expected_pvs, storageclass)
 
 
-def create_optional_operators_catalogsource_non_ga():
+def create_optional_operators_catalogsource_non_ga(force=False):
     """
     Creating optional operators CatalogSource and ImageContentSourcePolicy
     for non-ga OCP.
 
+    Args:
+        force (bool): enable/disable lso catalog setup
+
     """
     ocp_version = version.get_semantic_ocp_version_from_config()
     ocp_ga_version = get_ocp_ga_version(ocp_version)
-    if not ocp_ga_version:
-        optional_operators_data = list(
-            templating.load_yaml(
-                constants.LOCAL_STORAGE_OPTIONAL_OPERATORS, multi_document=True
-            )
+    if ocp_ga_version and not force:
+        return
+    optional_operators_data = list(
+        templating.load_yaml(
+            constants.LOCAL_STORAGE_OPTIONAL_OPERATORS, multi_document=True
         )
-        optional_operators_yaml = tempfile.NamedTemporaryFile(
-            mode="w+", prefix="optional_operators", delete=False
+    )
+    optional_operators_yaml = tempfile.NamedTemporaryFile(
+        mode="w+", prefix="optional_operators", delete=False
+    )
+    if config.DEPLOYMENT.get("optional_operators_image"):
+        for _dict in optional_operators_data:
+            if _dict.get("kind").lower() == "catalogsource":
+                _dict["spec"]["image"] = config.DEPLOYMENT.get(
+                    "optional_operators_image"
+                )
+    if config.DEPLOYMENT.get("disconnected"):
+        # in case of disconnected environment, we have to mirror all the
+        # optional_operators images
+        icsp = None
+        for _dict in optional_operators_data:
+            if _dict.get("kind").lower() == "catalogsource":
+                index_image = _dict["spec"]["image"]
+            if _dict.get("kind").lower() == "imagecontentsourcepolicy":
+                icsp = _dict
+        mirrored_index_image = (
+            f"{config.DEPLOYMENT['mirror_registry']}/"
+            f"{index_image.split('/', 1)[-1]}"
         )
-        if config.DEPLOYMENT.get("optional_operators_image"):
-            for _dict in optional_operators_data:
-                if _dict.get("kind").lower() == "catalogsource":
-                    _dict["spec"]["image"] = config.DEPLOYMENT.get(
-                        "optional_operators_image"
-                    )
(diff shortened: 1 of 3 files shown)


Avilir (dismissed):
LGTM

ocs_ci/deployment/helpers/lso_helpers.py
```diff
@@ -252,6 +207,62 @@ def setup_local_storage(storageclass):
...
 
+def create_optional_operators_catalogsource_non_ga():
+    """
+    Creating optional operators CatalogSource and ImageContentSourcePolicy
+    for non-ga OCP.
+
+    """
+    ocp_version = version.get_semantic_ocp_version_from_config()
+    ocp_ga_version = get_ocp_ga_version(ocp_version)
+    if not ocp_ga_version:
```
mbukatov, line 218: I would suggest to write this in the opposite way:
```
if ocp_ga_version:
    return
.... the rest of the code originally in the if branch ...
```
So that we avoid one level of code nesting, and prevent any future issues when one doesn't follow this nesting by mistake.

mbukatov: It would be also nice to have an option to force lso catalog setup (ignoring result of `get_ocp_ga_version()`), eg. via optional argument of the new fuction, eg. `create_optional_operators_catalogsource_non_ga(force=False)` so that one can use it even when the original assumption (setup only on non GA OCP instances) doesn't hold.

mbukatov (commented):
I have few suggestion, but otherwise it looks good. Thank you for the changes.

ocs_ci/deployment/helpers/lso_helpers.py
```diff
@@ -252,6 +207,62 @@ def setup_local_storage(storageclass):
...
+        templating.dump_data_to_temp_yaml(
+            optional_operators_data, optional_operators_yaml.name
+        )
+        with open(optional_operators_yaml.name, "r") as f:
+            logger.info(f.read())
+        logger.info(
+            "Creating optional operators CatalogSource and ImageContentSourcePolicy"
+        )
+        run_cmd(f"oc create -f {optional_operators_yaml.name}")
+        logger.info("Sleeping for 60 sec to start update machineconfigpool status")
```
petr-balogh, line 262: this message can be removed as the log message was moved already to wait_for_machineconfigpool_status function so, please remove it from here.

ocs_ci/deployment/helpers/lso_helpers.py
petr-balogh: In the case you will implement also this functionality, basically you will be able to change the condition as:
```python
if ocp_ga_version and not force:
    return
```

ocs_ci/deployment/helpers/lso_helpers.py
OdedViner: done

ocs_ci/deployment/helpers/lso_helpers.py
OdedViner: done

ocs_ci/deployment/helpers/lso_helpers.py
```diff
@@ -42,52 +42,7 @@ def setup_local_storage(storageclass):
...
-        )
-        with open(optional_operators_yaml.name, "r") as f:
-            logger.info(f.read())
-        logger.info(
-            "Creating optional operators CatalogSource and ImageContentSourcePolicy"
-        )
-        run_cmd(f"oc create -f {optional_operators_yaml.name}")
-        logger.info("Sleeping for 60 sec to start update machineconfigpool status")
-        wait_for_machineconfigpool_status("all")
+    create_optional_operators_catalogsource_non_ga()
```
PrasadDesala, line 45: Shouldn't we call this function only for non ocp_ga_versions. 
```
if not ocp_ga_version:
  create_optional_operators_catalogsource_non_ga()
```

ocs_ci/ocs/ui/views.py
```diff
@@ -147,6 +147,10 @@
     "enable_taint_node": ('input[id="taint-nodes"]', By.CSS_SELECTOR),
     "gp2-csi_sc": ('a[id="gp2-csi-link"]', By.CSS_SELECTOR),
     "gp3-csi_sc": ('a[id="gp3-csi-link"]', By.CSS_SELECTOR),
+    "choose_local_storage_version": (
```
petr-balogh, line 150: Where this one is used?

petr-balogh (approved):

ocs_ci/deployment/helpers/lso_helpers.py
mbukatov: @OdedViner moved this check into the function, so that it won't do anything when run on GA version. I would not do that myself, but at the same time I decided to not object it in the review (as I wouldn't approve anything otherwise).

ocs_ci/ocs/ui/views.py
petr-balogh: after the discussion , this should not be changed when 4.10 is GAed. So this has to be in condition

ocs_ci/deployment/helpers/lso_helpers.py
PrasadDesala: Yes, I did see that Oded is handling it in the function itself. 
If we call setup_local_storage(), below two lines of code are executed twice, one in the setup_local_storage and again in create_optional_operators_catalogsource_non_ga.
```
ocp_version = version.get_semantic_ocp_version_from_config()
ocp_ga_version = get_ocp_ga_version(ocp_version)
```
In the setup_local_storage() itself we will come to know whether the ocp version is a GA one or non GA. So if the ocp version is GA version we can right away skip entering create_optional_operators_catalogsource_non_ga().

I felt good to have this handled , if you all think that is fine, feel free to ignore this comment.

mbukatov: I agree with @PrasadDesala

OdedViner: done

ocs_ci/ocs/ui/views.py
OdedViner: done

PrasadDesala (approved):

ocs_ci/deployment/helpers/lso_helpers.py
```diff
@@ -252,6 +208,65 @@ def setup_local_storage(storageclass):
...
+    for non-ga OCP.
+
+    Args:
+        force (bool): enable/disable lso catalog setup
+
+    """
+    ocp_version = version.get_semantic_ocp_version_from_config()
+    ocp_ga_version = get_ocp_ga_version(ocp_version)
+    if ocp_ga_version and not force:
+        return
```
mbukatov, line 223: But you no longer need to fetch version and skip the function based on that here again, right?

OdedViner: Correct.. But I need to add condition here too
https://github.com/red-hat-storage/ocs-ci/blob/0d0522b469e0980097fbaf8766fec66389003676/ocs_ci/ocs/ui/deployment_ui.py#L422

@mbukatov What do you think? remove the condition in `create_optional_operators_catalogsource_non_ga`?

petr-balogh:
When you tested this, have you covered both UI and non UI deployments?

PR #5527, opened 2022-03-01
get RHEL version from external RHCS cluster

vavuthu:

### DIFF
--- a/ocs_ci/deployment/helpers/external_cluster_helpers.py
+++ b/ocs_ci/deployment/helpers/external_cluster_helpers.py
@@ -4,6 +4,7 @@ external cluster deployment.
 """
 
 import logging
+import re
 import tempfile
 
 from ocs_ci.framework import config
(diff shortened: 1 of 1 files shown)


ocs_ci/deployment/helpers/external_cluster_helpers.py
```diff
@@ -113,6 +126,20 @@ def get_rgw_endpoint_api_port(self):
         logger.info(f"External cluster rgw endpoint api port: {out}")
         return out
 
+    def get_rhel_version(self):
```
shylesh, line 129: @vavuthu can't we fetch from platform
```python
>>> import platform
>>> print(platform.linux_distribution())
('Red Hat Enterprise Linux Server', '7.9', 'Maipo')

vavuthu: yes, but we need to get that from remote server not on local machine. intention is to identify which python version exists on RHCS cluster. so we cann't use above method to get rhel version from remote server

shylesh (approved):
LGTM

petr-balogh (approved):

PR #5530, opened 2022-03-01
Performance Dashboard - remove the check if cluster is compressed

Avilir:
when pushing data to the dashboard the test if the test ran on compressed volume is irrelevant, since this test is running after actual tests was ran and the volumes (compressed) was deleted already.


ramkiperiy (approved):

ypersky1980 (approved):
LGTM

PR #5533, opened 2022-03-01
Change the default imagepullpolicy for performance test

Avilir:
The default image pull policy is : Always
Which mean that in the pod attach / reattach we are measuring the pull image time in the total attach time.

This PR change the default image pull policy to : IfNotPresent
So, the image pull time will no calculate in the attach time, since we are pulling it separately in the beginning of the test and not as part as the pod creation process.

This PR also remove some code duplication from the pod_reattach test.

the Pulling image is also implements in the bulk_pod_attach test
some duplicated code remove from the bulk_pod_attach test.

### DIFF
--- a/tests/e2e/performance/csi_tests/test_bulk_pod_attachtime_performance.py
+++ b/tests/e2e/performance/csi_tests/test_bulk_pod_attachtime_performance.py
@@ -7,13 +7,11 @@ import os
 import pytest
 import pathlib
 import time
-from uuid import uuid4
 
 from concurrent.futures import ThreadPoolExecutor
 from ocs_ci.framework.testlib import performance, polarion_id
 from ocs_ci.helpers import helpers
 from ocs_ci.helpers.helpers import get_full_test_logs_path
-from ocs_ci.framework import config
 from ocs_ci.ocs import defaults, constants, scale_lib
 from ocs_ci.ocs.resources.pod import get_pod_obj
 from ocs_ci.ocs.perftests import PASTest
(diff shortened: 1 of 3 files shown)


keesturam (approved):
LGTM Avi.

tests/e2e/performance/csi_tests/test_bulk_pod_attachtime_performance.py
```diff
@@ -78,19 +67,15 @@ def base_setup(self, project_factory, interface_type, storageclass_factory):
...
                 *[constants.CEPHBLOCKPOOL, 240],
-                marks=[pytest.mark.performance],
             ),
             pytest.param(
                 *[constants.CEPHFILESYSTEM, 120],
-                marks=[pytest.mark.performance],
             ),
             pytest.param(
                 *[constants.CEPHFILESYSTEM, 240],
-                marks=[pytest.mark.performance],
```
ypersky1980, line 93: Avi, why did you remove the mark ?

tests/e2e/performance/csi_tests/test_pod_reattachtime.py
```diff
@@ -32,25 +30,6 @@ def setup(self):
...
-                },
-            }
-        }
-        # during development use the dev ES so the data in the Production ES will be clean.
-        if self.dev_mode:
-            self.crd_data["spec"]["elasticsearch"] = {
-                "server": config.PERF.get("dev_es_server"),
-                "port": config.PERF.get("dev_es_port"),
-                "url": f"http://{config.PERF.get('dev_es_server')}:{config.PERF.get('dev_es_port')}",
-            }
```
ypersky1980, line 33: Also  - why did you remove this section in both tests? Don't we need it to upload to the ES?

tests/e2e/performance/csi_tests/test_bulk_pod_attachtime_performance.py
Avilir: since we have a global performance marker on the class level, so we don't need this marker per test.

tests/e2e/performance/csi_tests/test_pod_reattachtime.py
Avilir: we have it as global assignment in the PASTest object setup.

ypersky1980 (approved):
Thanks for a prompt fix!

PR #5537, opened 2022-03-02
Managed Services - Run pv_services disruption on provider in the multi cluster scenario

jilju:
Adding a function "run_on_cluster_type" which can be used to select a cluster type (eg: 'provider' , 'consumer' cluster type in managed services) when running oc command in a multi cluster scenario.
"run_on_cluster_type" can be used as a decorator in functions which need to use the kubeconfig of a particular cluster type.
This will set a local variable "cluster_kubeconfig" with value as the kubeconfig path of the cluster type. This variable can be accessed in called functions when needed. Currently applicable for Managed Services platforms only.

The functions which run oc command are updated to check the presence of a local variable "cluster_kubeconfig" in any stack above the caller's frame. If the local variable "cluster_kubeconfig" is detected, the value of "cluster_kubeconfig" which is the kubeconfig path will be added to the oc command.

Updated the functions in ocs_ci/helpers/disruption_helpers.py to use this decorator to perform disruption on provider.

This approach of using the appropriate cluster without switching the context is currently applicable for Managed Services platforms only.

This PR will enable the tests given below to run in a multi cluster scenario in managed services platform where consumer cluster type is used as the primary cluster.
tests/manage/pv_services/test_daemon_kill_during_pvc_pod_creation_deletion_and_io.py
tests/manage/pv_services/test_resource_deletion_during_pvc_pod_creation_and_io.py
tests/manage/pv_services/test_resource_deletion_during_pvc_pod_deletion_and_io.py
tests/manage/pv_services/pvc_snapshot/test_resource_deletion_during_snapshot_restore.py
tests/manage/pv_services/pvc_resize/test_resource_deletion_during_pvc_expansion.py
tests/manage/pv_services/pvc_clone/test_resource_deletion_during_pvc_clone.py

Other functions(eg: ocs_ci.ocs.resources.pod.get_osd_pods,  get_mon_pods) which should run on provider cluster can be decorated with "run_on_cluster_type" to use the provider cluster in a multi cluster run on Managed Services platform.


jilju:
Highly appreciate if a better solution can be provided. Some test cases will need to run oc commands on provider and consumer simultaneously which makes context switching within the test not feasible.

ocs_ci/utility/utils.py
```diff
@@ -1232,6 +1253,27 @@ def run_async(command):
...
+    # Managed services multi cluster run - Use provider cluster when needed
+    if (
+        command.lower().startswith("oc ")
+        and "--kubeconfig " not in command
+        and config.multicluster
+        and get_primary_cluster_config().ENV_DATA["platform"].lower()
+        in constants.MANAGED_SERVICE_PLATFORMS
+        and get_primary_cluster_config().ENV_DATA.get("cluster_type") == "consumer"
+    ):
+        all_stack = stack()
```
shylesh, line 1268: Hi @jilju , why are we accessing the function stack? you just need 'consumer' cluster's kubeconfig path while you are in different cluster context right ?

ocs_ci/helpers/disruption_helpers.py
```diff
@@ -23,6 +24,7 @@ class Disruptions:
     selector = None
     daemon_pid = None
 
+    @run_on_cluster_type(cluster_type="provider")
```
ebenahar, line 27: What if `Disruptions` constructor method will set the cluster's details, to run the commands on, based on whether or not the relevant resources exist in it? 
For instance, it will check on which cluster the ODF pods exist and once it finds the cluster, which should be the provider, it will use its details within the `Disruptions` class's methods

jilju: This possibility was tested when trying out different ways of implementation. Switching to provider context will make other simultaneous commands (run on consumer) to fail. Passing arguments to the called functions are also tested.
eg:
```
self.resource_obj[resource_id].delete(force=True, cluster=self.cluster_config)
```
after making the delete command to accept additional argument (testes key word args also). Doing this across ocs-ci tests will need code changes in many places(within test cases, helpers, utils etc). All intermediate functions will need to accept the cluster argument an pass it until reaching the function which runs the command.
I have also tried setting cluster argument after creating the OCP object of kind pod.
```
eg:  
setattr(self.resource_obj, cluster, <cluster conf or kubeconfig>)
```
This function `run_on_cluster_type` is not just created for the mentioned tier4 tests. This can be used in other functions which need provider cluster.

ocs_ci/utility/utils.py
jilju: The difficulty is in identifying which oc command should run on which cluster. I tried context switching but many test cases failed because there are simultaneous operations. Some operations need provider cluster and other need consumer cluster at the same time.

shylesh: @jilju , you can access any cluster's config object from config.clusters list using index and don't have to switch context.
So will it be possible to write a wrapper around this run_async where you will build the command and just call this run_async?

yitzhak12:
What about the functions in the sanity helpers file https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/helpers/sanity_helpers.py `create_resources`, `delete_resources` and `health_check`? Cause I see them in lots of the tests. I think `create_resources`, `delete_resources` should run on the consumer, but I am not sure about the `health_check` method.

jilju:
> What about the functions in the sanity helpers file https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/helpers/sanity_helpers.py `create_resources`, `delete_resources` and `health_check`? Cause I see them in lots of the tests. I think `create_resources`, `delete_resources` should run on the consumer, but I am not sure about the `health_check` method.

This is the initial PR. If this approach is acceptable and can be merged we can update other tests/helpers to choose the correct cluster when running in multi cluster configuration on managed services platform.

ocs_ci/ocs/utils.py
```diff
@@ -1266,3 +1267,50 @@ def get_primary_cluster_config():
     for cluster in ocsci_config.clusters:
         if cluster.MULTICLUSTER["primary_cluster"]:
             return cluster
+
+
+def get_provider_cluster_config():
```
fbalak, line 1272: I suggest to create a dedicated module for multi cluster related code and don't reuse utils module for the whole functionality.

ocs_ci/helpers/disruption_helpers.py
fbalak: This could be solved by creating locks on oc command calls and making sure that oc command calls in other threads wait until their turn and are not executed at the same time when other call is active. I think that this can be accomplished by using Lock objects (https://docs.python.org/3/library/threading.html#lock-objects) in functions that handle command execution (maybe exec_cmd).

ocs_ci/ocs/utils.py
yitzhak12: +1

ocs_ci/ocs/utils.py
```diff
@@ -1266,3 +1267,50 @@ def get_primary_cluster_config():
...
+
+    Returns:
+        framework.config: provider cluster config object from config.clusters
+    """
+    for cluster in ocsci_config.clusters:
+        if cluster.ENV_DATA.get("cluster_type") == "provider":
+            return cluster
+
+
+def run_on_cluster_type(cluster_type):
```
yitzhak12, line 1284: If we want to use more than one consumer, we need to add another param like `consumer_num` or `consumer_name`. But I am not sure if it's mandatory for now.

ocs_ci/helpers/disruption_helpers.py
jilju: @fbalak The requirement is to run two or more threads together when needed. So one thread waiting for another won't help in such cases.

ocs_ci/ocs/utils.py
jilju: Should it be managed services module or multi cluster module ? Something like
ocs_ci/ocs/manages_services.py
OR 
ocs_ci/ocs/multicluster.py

jilju:
The scope of tests in managed services platform to be revised. The requirement addressed in this PR is no longer relevant.

PR #5551, opened 2022-03-07
Added FORCE_OSD_REMOVAL flag on ocs-osd-removal-job

OdedViner:
https://bugzilla.redhat.com/show_bug.cgi?id=2057359#c4

### DIFF
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -1819,7 +1819,9 @@ def run_osd_removal_job(osd_ids=None):
     """
     osd_ids_str = ",".join(map(str, osd_ids))
     ocp_version = get_ocp_version()
-    if Version.coerce(ocp_version) >= Version.coerce("4.6"):
+    if Version.coerce(ocp_version) >= Version.coerce("4.10"):
+        cmd = f"process ocs-osd-removal -p FORCE_OSD_REMOVAL=true -p FAILED_OSD_IDS={osd_ids_str} -o yaml"
+    elif Version.coerce(ocp_version) >= Version.coerce("4.6"):
         cmd = f"process ocs-osd-removal -p FAILED_OSD_IDS={osd_ids_str} -o yaml"
     else:
         cmd = f"process ocs-osd-removal -p FAILED_OSD_ID={osd_ids_str} -o yaml"
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -1750,6 +1750,7 @@ def get_osd_removal_pod_name(osd_id, timeout=60):
         "4.7": "ocs-osd-removal-job",
         "4.8": "ocs-osd-removal-",
         "4.9": "ocs-osd-removal-job",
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/resources/pod.py
```diff
@@ -1819,7 +1819,10 @@ def run_osd_removal_job(osd_ids=None):
     """
     osd_ids_str = ",".join(map(str, osd_ids))
     ocp_version = get_ocp_version()
-    if Version.coerce(ocp_version) >= Version.coerce("4.6"):
+    ocs_version = config.ENV_DATA["ocs_version"]
+    if Version.coerce(ocs_version) >= Version.coerce("4.10"):
+        cmd = f"process ocs-osd-removal -p FORCE_OSD_REMOVAL=true -p FAILED_OSD_IDS={osd_ids_str} -o yaml"
```
Akarsha-rai, line 1824: Is it good to add clusters with 3 OSDs or not? Since with 6 OSDs or more without `FORCE_OSD_REMOVAL` should work

yitzhak12: I see in the last comment in the BZ: 

> The FORCE_OSD_REMOVAL value must be changed to “true” in clusters that only have three OSDs, or clusters with insufficient space to restore all three replicas of the data after the OSD is removed.

So it seems that also "insufficient space to restore all three replicas of the data after the OSD is removed" can cause a problem. IMO it will be too complicated to check it. What we can do is to run the osd removal job one **without** the param `FORCE_OSD_REMOVAL` and see if it works, and if not to run it again with the param `FORCE_OSD_REMOVAL`

yitzhak12: But if it's too complicated, I think we can leave it for now as it is.

OdedViner: This issue reproduced on 6 OSDs:
```
sh-4.4$ ceph osd tree 
ID  CLASS  WEIGHT   TYPE NAME           STATUS     REWEIGHT  PRI-AFF
-1         0.78149  root default                                    
-7         0.39075      host compute-0                              
 2    hdd  0.09769          osd.2       destroyed         0  1.00000
 3    hdd  0.09769          osd.3       destroyed         0  1.00000
 4    hdd  0.09769          osd.4              up   1.00000  1.00000
 5    hdd  0.09769          osd.5              up   1.00000  1.00000
-5         0.19537      host compute-1                              
 0    hdd  0.09769          osd.0              up   1.00000  1.00000
 6    hdd  0.09769          osd.6              up   1.00000  1.00000
-3         0.19537      host compute-2                              
 1    hdd  0.09769          osd.1              up   1.00000  1.00000
 7    hdd  0.09769          osd.7              up   1.00000  1.00000
```

yitzhak12 (dismissed):

Akarsha-rai (dismissed):

PrasadDesala (approved):

yitzhak12 (approved):

prsurve (dismissed):

PR #5555, opened 2022-03-08
#bz2030291 checks label on luks header info for all the encrypted osds 

mashetty330:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2030291#c25

This PR checks for LUKS_HEADER info on all the encrypted osds for each worker nodes to see if the label is set, which in turn helps us identify if a disk is an osd and belongs to our cluster.


ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -623,6 +628,49 @@ def osd_encryption_verification():
...
+                if luks_device_name == "":
+                    continue
+                logging.info(
+                    "Checking luks header label on Luks device {} for node {}".format(
+                        luks_device_name, n
+                    )
+                )
+
+                logging.info("Waiting for 15 seconds")
+                time.sleep(15)
```
OdedViner, line 658: you can use generator.

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -623,6 +628,49 @@ def osd_encryption_verification():
             )
             raise ValueError("OSD is not encrypted")
 
+    if ocs_version != version.VERSION_4_6:
```
OdedViner, line 631: you need to change this condition..
for example 4.5!=4.6
```
if ocs_version >= version.VERSION_4_6:

```

mashetty330: I want it to skip for version 4.6, thats why I used that condition. since this feature isn't back ported to 4.6 or I can do is 
`if ocs_version > version.VERSION_4_6`

ocs_ci/ocs/resources/storage_cluster.py
mashetty330: use generator for the inner loop? or for the sleep?

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -623,6 +628,49 @@ def osd_encryption_verification():
...
+        failure_message = ""
+        node_obj = OCP(kind="node")
+
+        for n in worker_nodes:
+            luks_devices = node_obj.exec_oc_debug_cmd(
+                node=n,
+                cmd_list=[
+                    "lsblk -o NAME,TYPE,FSTYPE | grep -E 'disk.*crypto_LUKS' | awk '{print $1}'"
+                ],
+            ).split("\n")
```
PrasadDesala, line 645: Please create a new function and move it to some common lib.  This looks to be a useful one and can help in other tests

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -623,6 +628,49 @@ def osd_encryption_verification():
...
+
+            for luks_device_name in luks_devices:
+                luks_device_name = luks_device_name.strip()
+                if luks_device_name == "":
+                    continue
```
PrasadDesala, line 650: You can remove this if you just pick osd running worker nodes

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -623,6 +628,49 @@ def osd_encryption_verification():
...
+                logging.info(
+                    "Checking luks header label on Luks device {} for node {}".format(
```
PrasadDesala, line 652: Please use f-string, which is the best and preferred way at all places

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -623,6 +628,49 @@ def osd_encryption_verification():
...
+                        luks_device_name, n
+                    )
+                )
+
+                logging.info("Waiting for 5 seconds")
+                time.sleep(5)
```
PrasadDesala, line 658: why are waiting for 5 sec here?

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -623,6 +628,49 @@ def osd_encryption_verification():
...
 
+    if ocs_version > version.VERSION_4_6:
+        log.info("Verify luks header label for encrypted devices")
+        worker_nodes = get_worker_nodes()
+        logging.info("NODES: {}".format(worker_nodes))
+        failures = 0
+        failure_message = ""
+        node_obj = OCP(kind="node")
+
+        for n in worker_nodes:
```
PrasadDesala, line 639: 'n' replace it with some meaningful name

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -623,6 +628,49 @@ def osd_encryption_verification():
...
+                    failures += 1
+                    failure_message += "\nNo label found on Luks header information for node {}\n".format(
+                        n
+                    )
+
+        if failures != 0:
+            logging.error(failure_message)
+            raise ValueError("Luks header label is not found")
+        logging.info("Luks header info found for all the encrypted osds")
+
```
PrasadDesala, line 673: As part of the BZ we should also try deleting the OSD deployment and see if the osd comes up successfully, do you plan to cover that aswell?

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -623,6 +628,49 @@ def osd_encryption_verification():
             )
             raise ValueError("OSD is not encrypted")
 
+    if ocs_version > version.VERSION_4_6:
+        log.info("Verify luks header label for encrypted devices")
+        worker_nodes = get_worker_nodes()
```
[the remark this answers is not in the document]
mashetty330: @PrasadDesala yeah im only checking osd's running on worker nodes. check line 633
`worker_nodes = get_worker_nodes()`

ocs_ci/ocs/resources/storage_cluster.py
mashetty330: yeah, I only automated the steps of bz verification mentioned by rachael. I understand we need to cover osd deployment deletion as part of bz but i'm not sure if its mandatory to check that.. if so then it will become a test case right?

ocs_ci/ocs/resources/storage_cluster.py
mashetty330: after each oc debug command (within the loop) we exit the debug mode, this might take few seconds depending on the cluster. so I gave 5 seconds wait

ocs_ci/ocs/resources/storage_cluster.py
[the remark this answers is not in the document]
PrasadDesala: ```get_worker_nodes()``` returns all the OCP worker nodes. Suppose if the cluster has an additional worker node without any running OSD this check will fail as there wont be any encrypted device on it. So, we have to check the encryption only on the osd running worker nodes.
You can use the existing function https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/node.py#L621 to get the osd running nodes.

ocs_ci/ocs/resources/storage_cluster.py
PrasadDesala: I think we can cover this scenario in a separate PR

ocs_ci/ocs/resources/storage_cluster.py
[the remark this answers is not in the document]
mashetty330: Okay I understand.. as per the logic it will only fail if the osd doesn't have header info not if the node doesn't have an osd. But if there is already existing function which only gives me osd running nodes then its great

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -623,6 +628,33 @@ def osd_encryption_verification():
             )
             raise ValueError("OSD is not encrypted")
 
+    if ocs_version > version.VERSION_4_6:
```
PrasadDesala, line 631: We are skipping the encryption verification for < 4.6 versions here[1]. 
This condition can be removed,

[1] https://github.com/red-hat-storage/ocs-ci/pull/5555/files#diff-193776bad25880fee96c0b8075161c63d5172b2d518d974326c937672cf7dd1fR604-R605

mashetty330: @PrasadDesala  but that condition includes 4.6 but luks header info feature isn't backported to 4.6.. so I want to skip even for 4.6

PrasadDesala: ack

PrasadDesala (approved):

OdedViner (approved):

PR #5559, opened 2022-03-08
Move RHEL7 config files to be without version

petr-balogh:
RHEL version is now taken from config of OCP version or the default
one. So no need to have RHEL version name in production configuration
files.

Also jenkins job names will require change and there will be open
MR.


clacroix12 (approved):

Avilir (approved):
LGTM

PR #5563, opened 2022-03-09
Upgrade external cluster

vavuthu:
supports ODF upgrade external cluster dynamically

This PR supports below upgrade scenarios which was missed few steps in previous releases.
Apart from that, In this PR we are creating object store user dynamically if it doesn't exists.

1. upgrade from 4.8 to 4.9
2. upgrade from 4.7 to 4.8
3. upgrade from 4.6 to 4.7
4. upgrade from 4.5 to 4.6

### DIFF
--- a/ocs_ci/ocs/ocs_upgrade.py
+++ b/ocs_ci/ocs/ocs_upgrade.py
@@ -541,6 +541,14 @@ def run_ocs_upgrade(operation=None, *operation_args, **operation_kwargs):
         f"is not higher or equal to the version you currently running: "
         f"{upgrade_ocs.version_before_upgrade}"
     )
+    # create external cluster object
+    if config.DEPLOYMENT["external_mode"]:
+        host = config.EXTERNAL_MODE["external_cluster_node_roles"]["node1"][
+            "ip_address"
+        ]
+        user = config.EXTERNAL_MODE["login"]["username"]
+        password = config.EXTERNAL_MODE["login"]["password"]
+        external_cluster = ExternalCluster(host, user, password)
 
     # For external cluster , create the secrets if upgraded version is 4.8
     if (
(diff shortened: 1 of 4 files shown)


ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -631,6 +642,31 @@ def run_ocs_upgrade(operation=None, *operation_args, **operation_kwargs):
             time.sleep(30)
             # End of workaround
 
+        # update external secrets
+        if config.DEPLOYMENT["external_mode"] and not (
+            original_ocs_version == "4.7" and upgrade_version == "4.8"
+        ):
+            external_cluster.update_permission_caps()
```
petr-balogh, line 649: Does this require the bug with caps to be backported to 4.8? Or it's something different here?

I remember that last conversation was not finalized to which version the bug will be backported. @agarwal-mudit might know more, but on last PC I remember he told that we will need another meeting to decide.

petr-balogh (dismissed):

ocs_ci/ocs/ocs_upgrade.py
agarwal-mudit: Yes, we need to backport it to 4.9 and 4.8
@parth-gr can add more.

parth-gr: Ohh is the automatic upgrade we have for external cluster?

No, we don't need any backport in case upgrade.

I want to understand what the change is about ig the things are already present you are moving parts more modular, Or I will say please update the PR description.

petr-balogh (approved):

dahorak (approved):

ocs_ci/ocs/ocs_upgrade.py
vavuthu: this PR is to create and run the script dynamically ( previously we have harcoded values for keys where user send as input to automation ) and also added few missing steps post upgrade.

parth-gr: Thanks for the change, 
I want to discuss with you the points for 4.10, how we should upgrade in that scenario because the upgrade flag we have used till 4.9 is updated in 4.10 script.

parth-gr: We can take this offline referr:https://github.com/rook/rook/blob/a585a69843fff3a5731ec0c21f02cbca37f4e687/deploy/examples/create-external-cluster-resources.py#L220

Issue #5565, opened 2022-03-09
OCP/OCS MG collection on failure tests


OdedViner:
- As part of the "Improve testing suites" group one of the tasks is to optimize the tests that take more than 30 mins to execute.

- We found that we collect OCP Must-Gather and OCS Must-Gather when the test fails.
> 

For Example on this tier1 job, we collected 40 OCS MG + 40 OCP MG 
- For_ each failure test, we collect ocp_mg+ocs_mg and noobaa_db_dump [only for MCG tests].
-  Estimated time for must gather collection in this job[tier1] is:
Time=40_OCS_MG * 5 min + 40_OCP_MG * 5 min ~ 6.67 hours

**Tasks:**

- Create a mechanism to enable/disable OCS/OCP must-gather collection:

For example:
```
@disable_ocs_mg_collection_on_failure
@enable_ocp_mg_collection_on_failure
def test_abc():
  	pass

```

- Enable OCS mg collection on failure by default 

- Disable OCP mg collection on failure by default 

-  Collect the OCP MG only on the end of the suite..

ebenahar:
IIUC, setting OCP MG collection to False by default should be done in ocs_ci.framework.pytest_customization.ocscilib.pytest_runtest_makereport

mykaul:
- Any reason the OCP and OCS must gather are not running in parallel?
- Is it really taking 5 minutes(!) to collect logs? Wow. What is that time spent on?

OdedViner:
> * Any reason the OCP and OCS must gather are not running in parallel? --> I agree, we can use threads
> * Is it really taking 5 minutes(!) to collect logs? Wow. What is that time spent on? -->
$ time oc adm must-gather --image=quay.io/rhceph-dev/ocs-must-gather:latest-4.10
real	4m43.058s
user	0m11.509s
sys	0m5.029s

http://pastebin.test.redhat.com/1050439

@agarwal-mudit

PR #5566, opened 2022-03-09
Disable/Enable OCS/OCP mg collection

OdedViner:
#5565

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -110,6 +110,8 @@ REPORTING:
     project_id: "OpenShiftContainerStorage"
   # Upstream: 'US' or Downstream: 'DS', used only for reporting (Test Run Name)
   us_ds: 'DS'
+  ocs_must_gather_collection: true
+  ocp_must_gather_collection: false
   ocp_must_gather_image: "quay.io/openshift/origin-must-gather"
   default_ocs_must_gather_image: "quay.io/rhceph-dev/ocs-must-gather"
   overwrite_must_gather_image: true
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -110,8 +110,6 @@ REPORTING:
     project_id: "OpenShiftContainerStorage"
   # Upstream: 'US' or Downstream: 'DS', used only for reporting (Test Run Name)
   us_ds: 'DS'
-  ocs_must_gather_collection: true
-  ocp_must_gather_collection: false
   ocp_must_gather_image: "quay.io/openshift/origin-must-gather"
   default_ocs_must_gather_image: "quay.io/rhceph-dev/ocs-must-gather"
   overwrite_must_gather_image: true
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -623,7 +623,9 @@ def pytest_runtest_makereport(item, call):
     if rep.failed and ocsci_config.RUN.get("cli_params").get("collect-logs"):
         test_case_name = item.name
         ocp_logs_collection = (
-            True if any(x in item.location[0] for x in ["ecosystem"]) else False
+            True
+            if any(x in item.location[0] for x in ["ecosystem", "e2e/performance"])
+            else False
         )
         ocs_logs_collection = (
             False
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -622,8 +622,14 @@ def pytest_runtest_makereport(item, call):
     # we only look at actual failing test calls, not setup/teardown
     if rep.failed and ocsci_config.RUN.get("cli_params").get("collect-logs"):
         test_case_name = item.name
-        ocp_logs_collection = ocsci_config.REPORTING.get("ocp_must_gather_collection")
-        ocs_logs_collection = ocsci_config.REPORTING.get("ocs_must_gather_collection")
+        ocp_logs_collection = (
+            True if any(x in item.location[0] for x in ["ecosystem"]) else False
+        )
+        ocs_logs_collection = (
+            False
+            if any(x in item.location[0] for x in ["_ui", "must_gather"])
+            else True
+        )
         mcg_logs_collection = (
             True if any(x in item.location[0] for x in ["mcg", "ecosystem"]) else False
         )
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -622,14 +622,18 @@ def pytest_runtest_makereport(item, call):
     # we only look at actual failing test calls, not setup/teardown
     if rep.failed and ocsci_config.RUN.get("cli_params").get("collect-logs"):
         test_case_name = item.name
-        ocp_logs_collection = True if rep.when == "call" else False
-        mcg = (
+        ocp_logs_collection = ocsci_config.REPORTING.get("ocp_must_gather_collection")
+        ocs_logs_collection = ocsci_config.REPORTING.get("ocs_must_gather_collection")
+        mcg_logs_collection = (
             True if any(x in item.location[0] for x in ["mcg", "ecosystem"]) else False
         )
         try:
             if not ocsci_config.RUN.get("is_ocp_deployment_failed"):
                 collect_ocs_logs(
-                    dir_name=test_case_name, ocp=ocp_logs_collection, mcg=mcg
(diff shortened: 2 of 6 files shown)


tests/conftest.py
```diff
@@ -4523,3 +4523,19 @@ def finalizer():
 
     request.addfinalizer(finalizer)
     return factory
+
+
+@pytest.fixture(scope="function", autouse=True)
```
ebenahar, line 4528: I don't think we need to define new fixures. 
For setting the defaults, we can do that in setting OCP MG collection to False by default should be done in ocs_ci.framework.pytest_customization.ocscilib.pytest_runtest_makereport

OdedViner: correct. but I want to add option to change it for specific tests.. So I used `disable_ocs_mg_collection_on_failure`, `enable_ocp_mg_collection_on_failure` fixtures.
So I think for the next test we need to init them again.. I will run it locally

OdedViner: I tested on my PC
We need to add this fixture:
```import logging
class TestAB():
    def test_a(self, enable_ocp_mg_collection_on_failure):
        logging.info(f"OCS-MG={config.REPORTING.get('ocs_must_gather_collection')}")
        logging.info(f"OCP-MG={config.REPORTING.get('ocp_must_gather_collection')}")

    def test_b(self):
        logging.info(f"OCS-MG={config.REPORTING.get('ocs_must_gather_collection')}")
        logging.info(f"OCP-MG={config.REPORTING.get('ocp_must_gather_collectio```

File: ocs_ci/framework/pytest_customization/ocscilib.py [hunk 2]
```diff
@@ -622,14 +622,18 @@ def pytest_runtest_makereport(item, call):
...
+        ocs_logs_collection = ocsci_config.REPORTING.get("ocs_must_gather_collection")
+        mcg_logs_collection = (
             True if any(x in item.location[0] for x in ["mcg", "ecosystem"]) else False
         )
         try:
             if not ocsci_config.RUN.get("is_ocp_deployment_failed"):
                 collect_ocs_logs(
-                    dir_name=test_case_name, ocp=ocp_logs_collection, mcg=mcg
+                    dir_name=test_case_name,
+                    ocp=ocp_logs_collection,
```
petr-balogh, line 634: As you set it to False in default config then for failed deployment we will not collect OCP must gather logs.  For investigating deployment issues we need OCP must gather so I see it might be problematic for ecosystem team for deployment failures analysis.

[edit] Ah, I see you are using enable_ocp_mg_collection_on_failure in deployment test so it should be resolved by this.

tests/conftest.py
```diff
@@ -4523,3 +4523,19 @@ def finalizer():
...
+def collect_mg_rules():
+    config.REPORTING["ocs_must_gather_collection"] = True
```
petr-balogh, line 4530: With this approach you are overwriting the configuration from the config - basically you will remove whatever will user provide in custom configuration even the default one.

Which as I pointed out is the reason why to document it if you we want to go with this approach.

@shylesh this might be a problematic in multi cluster scenario I think. Right? As we don't know which one cluster config context we are on at the moment.

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -110,6 +110,8 @@ REPORTING:
     project_id: "OpenShiftContainerStorage"
   # Upstream: 'US' or Downstream: 'DS', used only for reporting (Test Run Name)
   us_ds: 'DS'
+  ocs_must_gather_collection: true
```
petr-balogh, line 113: Those values should be documented in the documentation config file:
https://github.com/red-hat-storage/ocs-ci/blob/master/conf/README.md - Currently what I see , those values are not configurable by user as if I will set both to false, it will be overwritten in autouse fixture collect_mg_rules .  So in this case we should also document it that those values are not suppose to be configured by user.

tests/conftest.py
```diff
@@ -4523,3 +4523,19 @@ def finalizer():
...
+    config.REPORTING["ocp_must_gather_collection"] = False
+
+
+@pytest.fixture()
+def disable_ocs_mg_collection_on_failure():
```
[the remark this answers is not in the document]
OdedViner: If I understand it correctly, this approach will change the configuration value and it will be changed also for all the following tests (until one of those fixtures are not used again in another test) - is my understanding correct? --> correct and I added a new fixture
```
@pytest.fixture(scope="function", autouse=True)
def collect_mg_rules():
    config.REPORTING["ocs_must_gather_collection"] = True
```

tests/conftest.py
petr-balogh: It doesn't change anything. But as this is REPORTING related variable I wold keep it under REPORTING section. Or in RUN section as it is RUN related variable which changing over the run.

clacroix12 (changes requested):
Hi @OdedViner, I think there are a couple issues with this approach. Currently the autouse fixture you created is simply overwriting the values in the configuration for whether or not we execute must gather. As @petr-balogh  mentioned this is problematic because the config is overwritten so users have no control over whether or not must gather is executed, also as @dahorak  mentioned this will change the configuration for all tests executed after the one that uses this fixture. This is highly problematic because it causes unexpected behavior in all subsequent tests.

If we think a fixture is the correct way to handle this operation, I think it would make more sense for the fixture itself to execute must gather based on what values are set in the configuration. If the config says to not run ocp must gather, then the fixture will skip that step. Same for ocs. The fixture could also run this as a part of it's teardown (to ensure it runs after the test case) and only if the test fails. WDYT?

tests/conftest.py
shylesh: > With this approach you are overwriting the configuration from the config - basically you will remove whatever will user provide in custom configuration even the default one.
> 
> Which as I pointed out is the reason why to document it if you we want to go with this approach.
> 
> @shylesh this might be a problematic in multi cluster scenario I think. Right? As we don't know which one cluster config context we are on at the moment.

@petr-balogh , Yes , you are right. In between the multicluster tests if context gets switched then this flag will have unexpected value.

tests/conftest.py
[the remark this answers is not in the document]
dahorak: I meant it generally for all the new fixtures - when they are used, they change the configuration for all following tests until another fixture is explicitly used in some test.

For example if I'll have following tests:

1. `test_a(disable_ocs_mg_collection_on_failure)`
1. `test_b()`
1. `test_c(collect_mg_rules)`
1. `test_d()`

* Then if I'll run them all and if they will be executed in this order, then `test_b` will have `ocs_must_gather_collection` set to `False`, because it will be set by the fixture in `test_a`. Then `test_c` will change it back to `True`.
* If I'll run only `test_a`, `test_b` and `test_d`, then both `test_b` and `test_d` will have the `ocs_must_gather_collection` set to `False`, even though they didn't required it.
* And if I'll run for example only `test_b`, `test_c` and `test_d`, all the tests will have `ocs_must_gather_collection` set to `True`.

This is what I meant when I wrote above, that one test changes behavior of the following tests., which is not good.

I think, that the approach should be, that if you want to change the defaults for one test, you should change it in the beginning of the test, but save the previous values and then restore the previous values once the test finish.

ebenahar (approved):

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -622,14 +622,24 @@ def pytest_runtest_makereport(item, call):
     # we only look at actual failing test calls, not setup/teardown
     if rep.failed and ocsci_config.RUN.get("cli_params").get("collect-logs"):
         test_case_name = item.name
-        ocp_logs_collection = True if rep.when == "call" else False
-        mcg = (
+        ocp_logs_collection = (
```
petr-balogh, line 625: Do we cover the scenario that we are collecting OCP must gather logs at least once at the end of the execution? With that collect logs on success?

clacroix12: I believe this is handled separately in a session scope fixture finalizer here: https://github.com/red-hat-storage/ocs-ci/blob/master/tests/conftest.py#L3710-L3726

clacroix12: However, with the change to the default value for `ocp` below in `utils.py` these calls will need to be updated as it will no longer work as intended. We are relying on the defaults to be True for those calls. @OdedViner

OdedViner: done

ebenahar (dismissed):

jilju (approved):

petr-balogh (dismissed):

clacroix12 (approved):

dahorak (approved):

dahorak (approved):

petr-balogh (approved):

PR #5575, opened 2022-03-10
Implement Submariner Deployment with UI [Downstream]

jimdevops19:
Signed-off-by: jimdevops19 <jimdevops@gmail.com>

### DIFF
--- a/conf/ocsci/submariner_downstream.yaml
+++ b/conf/ocsci/submariner_downstream.yaml
@@ -1,0 +1,2 @@
+ENV_DATA:
+  submariner_source: "downstream"
--- a/ocs_ci/deployment/acm.py
+++ b/ocs_ci/deployment/acm.py
@@ -17,7 +17,7 @@ from ocs_ci.ocs.exceptions import (
 from ocs_ci.ocs.utils import get_non_acm_cluster_config
 from ocs_ci.utility.utils import run_cmd, run_cmd_interactive
 from ocs_ci.ocs.node import get_typed_worker_nodes, label_nodes
-
+from ocs_ci.ocs.acm.acm import AcmAddClusters, login_to_acm
 
 logger = logging.getLogger(__name__)
 
@@ -61,6 +61,8 @@ class Submariner(object):
     def __init__(self):
         # whether upstream OR downstream
         self.source = config.ENV_DATA["submariner_source"]
+        # Deployment type:
+        self.deployment_type = config.ENV_DATA.get("submariner_deployment")
         # Designated broker cluster index where broker will be deployed
         self.designated_broker_cluster_index = self.get_primary_cluster_index()
         # sequence number for the clusters from submariner perspective
@@ -74,15 +76,22 @@ class Submariner(object):
     def deploy(self):
         if self.source == "upstream":
             self.deploy_upstream()
-        else:
+        elif self.source == "downstream":
             self.deploy_downstream()
+        else:
+            raise Exception(f"The Submariner source: {self.source} is not recognized")
 
     def deploy_upstream(self):
         self.download_binary()
         self.submariner_configure_upstream()
 
     def deploy_downstream(self):
-        raise NotImplementedError("Deploy downstream functionality not implemented")
+        config.switch_acm_ctx()
+        # Get the Selenium driver obj after logging in to ACM
+        get_driver = login_to_acm()
+        acm_obj = AcmAddClusters(get_driver)
+        acm_obj.install_submariner_ui()
+        acm_obj.submariner_validation_ui()
 
     def download_binary(self):
         if self.source == "upstream":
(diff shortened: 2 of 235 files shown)


conf/ocsci/submariner_downstream.yaml
```diff
@@ -0,0 +1,7 @@
+ENV_DATA:
+  submariner_source: "downstream"
+  submariner_url: "https://get.submariner.io"
```
petr-balogh, line 3: Is this used also in downstream deployment? I thought that this is upstream page, so curious what it's doing in DS deployment config.

ocs_ci/deployment/acm.py
```diff
@@ -74,15 +76,23 @@ def __init__(self):
...
+        else:
+            raise Exception(f"The Submariner source: {self.source} is not recognized")
 
     def deploy_upstream(self):
         self.download_binary()
         self.submariner_configure_upstream()
 
     def deploy_downstream(self):
-        raise NotImplementedError("Deploy downstream functionality not implemented")
+        if self.deployment_type == constants.UI_DEPLOYMENT:
```
petr-balogh, line 89: what if the UI deployment will not be enabled? Then it will not do anything. I would raise some NotImplementedError in the else block.

shylesh: @jimdevops19 , here we need to handle the case, we can have flag which tells what type of submariner deployment is this ? `submariner_ui_deployment` and set it to true. WDYT?

jimdevops19: Addressed in a way that we pass `ui_deployment` through DEPLOYMENT section, since it is a key consistently passed from this section and we should be consistent with it

ocs_ci/deployment/acm.py
```diff
@@ -74,15 +74,27 @@ def __init__(self):
     def deploy(self):
         if self.source == "upstream":
             self.deploy_upstream()
-        else:
+        elif self.source == "downstream":
```
petr-balogh, line 77: I think that we should go with the downstream deployment by default without need of the passing extra configuration file, which we can achieve to set the default value in the .get method here: 
https://github.com/red-hat-storage/ocs-ci/blob/de15cfe75c67fd20ad711770f424d8888629f320/ocs_ci/deployment/acm.py#L63

Or setting it as a default value in the default configuration file here:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml

@shylesh @prsurve @sidhant-agrawal @am-agrawa can you put your thoughts here? Is my assumption correct that by default we should go aways with downstream? And if from some reason we need upstream, we can always pass the config file for upstream?

Thanks

ocs_ci/deployment/acm.py
```diff
@@ -74,15 +74,27 @@ def __init__(self):
...
+        else:
+            raise Exception(f"The Submariner source: {self.source} is not recognized")
 
     def deploy_upstream(self):
         self.download_binary()
         self.submariner_configure_upstream()
 
     def deploy_downstream(self):
-        raise NotImplementedError("Deploy downstream functionality not implemented")
+        if config.DEPLOYMENT.get("ui_deployment"):
```
petr-balogh, line 87: For now I would go with removing this configuration value.

Once someone will implement UI deployment for upstream or CLI deployment for downstream I would start thinking and adding the logic based on config value. But as we don't have the UI deployment for upstream, and here we don't have CLI for downstream let's remove this condition and config value from config file.

Only once someone will implement other logic we will add some logic for it.

conf/ocsci/submariner_downstream.yaml
```diff
@@ -0,0 +1,4 @@
+ENV_DATA:
+  submariner_source: "downstream"
+DEPLOYMENT:
+  ui_deployment: true
```
petr-balogh, line 4: As discussed below, remove.

It might cause and affect OCS/ODF deployment and I am not sure we have the UI deployment done for it.

ocs_ci/deployment/acm.py
sidhant-agrawal: yes, by default we should use downstream. In specific case, for eg. when using internal builds for ACM, we may need to use upstream.

petr-balogh (approved):

ocs_ci/ocs/ui/views.py
```diff
@@ -844,6 +844,7 @@
             **deployment_4_9,
             **deployment_4_10,
         },
+        "acm_page": {**acm_page_nav, **acm_configuration},
```
petr-balogh, line 847: It's already on line 856 so it can be removed from here

petr-balogh (approved):

petr-balogh:
@jimdevops19 I see that the docs check from tox is failing with:
```
  cannot import name 'Submariner' from partially initialized module 'ocs_ci.deployment.acm' (most likely due to a circular import) (/home/runner/work/ocs-ci/ocs-ci/ocs_ci/deployment/acm.py)
  ```

petr-balogh (approved):

petr-balogh (approved):

vavuthu (approved):

PR #5580, opened 2022-03-11
delete csi users during cluster deletion

vavuthu:

### DIFF
--- a/ocs_ci/deployment/vmware.py
+++ b/ocs_ci/deployment/vmware.py
@@ -29,6 +29,7 @@ from ocs_ci.ocs.node import (
 )
 from ocs_ci.utility import templating, version
 from ocs_ci.ocs.openshift_ops import OCP
+from ocs_ci.ocs.resources import pod
 from ocs_ci.utility.aws import AWS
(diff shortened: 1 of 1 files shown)


ocs_ci/deployment/vmware.py
```diff
@@ -733,6 +734,15 @@ def destroy_cluster(self, log_level="DEBUG"):
         logger.debug(f"changing state from {str_to_modify} to {target_str}")
         replace_content_in_file(terraform_tfstate, str_to_modify, target_str)
 
+        # remove csi users in case of external deployment
+        if config.DEPLOYMENT["external_mode"]:
```
petr-balogh, line 738: Do we need to do that also for vsphere-cleanup script?

petr-balogh (approved):

ocs_ci/deployment/vmware.py
vavuthu: yes, we need to remove in cleanup script as well in case of external mode. I will raise github issue for the same

jilju (approved):

PR #5593, opened 2022-03-16
Fixes #5582; Removing Black Squad

### LINKED ISSUE #5582: Black squad scrapped; Update black squad test marker with respective squads
The black squad is scrapped. So, it is important to update markers in ocs-ci with the respective squads.

keesturam:
Signed-off-by: Krishnaram Karthick Ramdoss <kramdoss@redhat.com>

### DIFF
--- a/tests/ui/test_validation_ui.py
+++ b/tests/ui/test_validation_ui.py
@@ -6,6 +6,7 @@ from ocs_ci.framework.testlib import (
     skipif_ocs_version,
     ui,
 )
+from ocs_ci.framework.pytest_customization.marks import blue_squad
 from ocs_ci.ocs.ui.validation_ui import ValidationUI
 from ocs_ci.utility import version
 
@@ -20,6 +21,7 @@ class TestUserInterfaceValidation(object):
 
     @ui
     @tier1
+    @blue_squad
     @skipif_ui_not_support("validation")
     def test_dashboard_validation_ui(self, setup_ui):
         """
@@ -40,6 +42,7 @@ class TestUserInterfaceValidation(object):
     @tier1
     @skipif_ui_not_support("validation")
     @skipif_ocs_version("<4.9")
+    @blue_squad
     def test_odf_storagesystems_ui(self, setup_ui):
         """
         Validate User Interface for ODF Storage Systems Tab for ODF 4.9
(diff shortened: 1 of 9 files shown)


vavuthu (approved):

vavuthu (commented):
please remove from https://github.com/red-hat-storage/ocs-ci/blob/dffe18d502007544da8e88d0c7c3214a302273dd/pytest.ini#L53

and

https://github.com/red-hat-storage/ocs-ci/blob/f57e12c4049c409570a9103b8807c82e5630e1bf/ocs_ci/framework/pytest_customization/marks.py#L371

ebenahar (dismissed):

tests/ui/test_validation_ui.py
```diff
@@ -20,6 +21,7 @@ class TestUserInterfaceValidation(object):
 
     @ui
     @tier1
+    @blue_squad
```
fbalak, line 24: This is not a blue squad test case.

tests/ui/test_validation_ui.py
```diff
@@ -40,6 +42,7 @@ def test_dashboard_validation_ui(self, setup_ui):
     @tier1
     @skipif_ui_not_support("validation")
     @skipif_ocs_version("<4.9")
+    @blue_squad
```
fbalak, line 45: This is not a blue squad test case.

fbalak (changes requested):
Blue squad is focused on monitoring testing, not on UI validations.

tests/ui/test_validation_ui.py
keesturam: @ebenahar any thoughts on which squad could be a better fit? 
This is a basic test that checks if the dashboard works. so it is not blue squad as per our definitions. But, it doesn't fall under any other category as well. 
Maybe purple squad?

am-agrawa (approved):
LGTM

jilju (approved):

tests/ui/test_validation_ui.py
ebenahar: indeed no squad has this area by definition. 
@keesturam can you keep it under black for now (even though we would like to deprecate this squad), until we define which squad will take this area

fbalak (approved):

am-agrawa (approved):

jilju (approved):

PR #5595, opened 2022-03-16
Remove duplicated constants

jimdevops19:
This just had to be addressed some time as a fix from adding multiple variables with the same value for no reason.

Addressed the following:
 - Line 161 was emoved, since 133 has the value of CLUSTER_SERVICE_VERSION = "csv"

 - Line 117 was removed since 134 has the value of 
JOB = "Job" (Line 134 overrides "job", so left the one with the lowercase)

 - Line 93 was removed since 88 has the value of STATUS_CLBO = "CrashLoopBackOff"

@clacroix12 @petr-balogh This might be too much to handle in our tox.ini, but I believe there must be a way to provide a pre-commit hook that might take a look into values that are used on the fly, while they exist in our `ocs_ci/ocs/constants.py` file.

### DIFF
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -12,7 +12,7 @@ and with consideration of the entire project.
 import os
 
 # Logging
-LOG_FORMAT = f"%(asctime)s - %(threadName)s - %(name)s - %(levelname)s - %(message)s - [{}]"
+LOG_FORMAT = "%(asctime)s - %(threadName)s - %(name)s - %(levelname)s - %(message)s"
 
 # Directories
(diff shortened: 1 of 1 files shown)


vavuthu (approved):

jilju (approved):

clacroix12:
@jimdevops19 I'm not sure I understand the point about our tox.ini. What would the pre-commit check do, look for duplicate definitions?

jimdevops19:
> @jimdevops19 I'm not sure I understand the point about our tox.ini. What would the pre-commit check do, look for duplicate definitions?

Yes. @clacroix12 .

PR #5541, opened 2022-03-03
Adjust the 'Automated recovery from failed nodes' tests to run on managed service

yitzhak12:

PR #5651, opened 2022-03-29
[WIP] Run test case on provider only

jilju:
The test case given below should be executed on provider cluster in a Managed Service run.
Test case - test_validate_ceph_config_values_in_rook_config_override

Added a fixture to switch the context to provider at the beginning of the test and switch the context back to the previous value after the test case.


jilju:
Verified: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/16/

jilju:
Verified https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/23/

jilju:
Verified running tier1 on MS using this PR https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/28/

tests/conftest.py
```diff
@@ -4720,3 +4720,38 @@ def patch_consumer_toolbox_with_secret():
...
+        and current_cluster.ENV_DATA.get("platform", "").lower()
+        in constants.MANAGED_SERVICE_PLATFORMS
+    ):
+        for cluster in config.clusters:
+            if cluster.ENV_DATA.get("cluster_type") == "provider":
+                provider_cluster = cluster
+                log.debug("Switching to the provider cluster context")
+                config.switch_ctx(provider_cluster.MULTICLUSTER["multicluster_index"])
+                switched_to_provider = True
+                break
```
dahorak, line 4746: There is a new function `switch_to_provider` introduced in [PR#5541](https://github.com/red-hat-storage/ocs-ci/pull/5541/files#diff-1aebcb25bdbc9860fbc89fc474c3ddc4c6ac92c93071487c9cf610a35b773655R268) which might be used here, but the PR is not yet merged - so it is mainly suggestion for future improvement.

jilju: Thanks @dahorak . Enhancements will be done after merging #5541

tests/conftest.py
```diff
@@ -4720,3 +4720,38 @@ def patch_consumer_toolbox_with_secret():
...
+
+    def finalizer():
+        """
+        Switch context to the initial cluster
+
+        """
+        if switched_to_provider:
+            log.debug("Switching back to the previous cluster context")
+            config.switch_ctx(current_cluster.MULTICLUSTER["multicluster_index"])
```
fbalak, line 4755:
```suggestion
            config.reset_ctx()
```

jilju: @fbalak reset_ctx will reset the context to clusters[0]. Here the finalizer is  reverting the context switching which is done in the fixture. So the context will be back to the one which was present just before this fixture. This can be clusters[0] or any other cluster index(when we use more than one consumer)

tests/conftest.py
jilju: @fbalak Added # TODO.

tests/conftest.py
fbalak: Ok, maybe some function to return previous context would be useful. It would be also good to add docstrings to all functions.

fbalak (approved):

yitzhak12 (approved):

PR #5659, opened 2022-03-31
Patch rook-ceph-tools deployment in consumer cluster

jilju:
Patch the rook-ceph-tools deployment with ceph.admin key. Applicable for MS platform only to enable rook-ceph-tools to run ceph commands until we have the proper for rook-ceph-tools in consumer cluster.

Added a session scope fixture to perform this at the beginning of the test.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4659,8 +4659,6 @@ def patch_consumer_toolbox_with_secret():
 
         # Patch the rook-ceph-tools deployment of all consumer clusters
         for cluster in config.clusters:
-            from pdb import set_trace
-
             if cluster.ENV_DATA.get("cluster_type") == "consumer":
                 config.switch_ctx(cluster.MULTICLUSTER["multicluster_index"])
                 consumer_tools_deployment = OCP(
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4628,58 +4628,61 @@ def patch_consumer_toolbox_with_secret():
     to run ceph commands until we have the fix for rook-ceph-tools in consumer cluster
 
     """
-    current_cluster = config.cluster_ctx
-
     # Get the secret from provider if MS multicluster run
-    if (
+    if not (
         config.multicluster
         and config.ENV_DATA.get("platform", "").lower()
         in constants.MANAGED_SERVICE_PLATFORMS
+        and not config.RUN["cli_params"].get("deploy")
     ):
-        # Get the admin key if available
-        ceph_admin_key = os.environ.get("CEPHADMINKEY") or config.AUTH.get(
-            "external", {}
-        ).get("ceph_admin_key")
+        return
 
-        if not ceph_admin_key:
-            provider_cluster = ""
+    current_cluster = config.cluster_ctx
 
-            # Identify the provider cluster
-            for cluster in config.clusters:
-                if cluster.ENV_DATA.get("cluster_type") == "provider":
-                    provider_cluster = cluster
-                    break
-            assert provider_cluster, "Provider cluster not found"
+    # Get the admin key if available
+    ceph_admin_key = os.environ.get("CEPHADMINKEY") or config.AUTH.get(
+        "external", {}
+    ).get("ceph_admin_key")
 
-            # Switch context to provider cluster
-            log.info("Switching to the provider cluster context")
-            config.switch_ctx(provider_cluster.MULTICLUSTER["multicluster_index"])
+    if not ceph_admin_key:
+        provider_cluster = ""
 
-            # Get the key from provider cluster tools pod
-            provider_tools_pod = get_ceph_tools_pod()
-            ceph_admin_key = (
-                provider_tools_pod.exec_cmd_on_pod("grep key /etc/ceph/keyring")
-                .strip()
-                .split()[-1]
-            )
-
-        # Patch the rook-ceph-tools deployment of all consumer clusters
+        # Identify the provider cluster
         for cluster in config.clusters:
-            if cluster.ENV_DATA.get("cluster_type") == "consumer":
-                config.switch_ctx(cluster.MULTICLUSTER["multicluster_index"])
-                consumer_tools_deployment = OCP(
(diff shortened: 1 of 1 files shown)


OdedViner (dismissed):

tests/conftest.py
```diff
@@ -4618,3 +4619,61 @@ def finalizer():
...
+def patch_consumer_toolbox_with_secret():
+    """
+    Patch the rook-ceph-tools deployment with ceph.admin key. Applicable for MS platform only to enable rook-ceph-tools
+    to run ceph commands until we have the fix for rook-ceph-tools in consumer cluster
+
+    """
+    current_cluster = config.cluster_ctx
+
+    # Get the secret from provider if MS multicluster run
+    if (
```
sabose, line 4634: Can we add another if block to get the key from env_data if present? @jilju ?

jilju: Added check for the environment variable CEPHADMINKEY. The key can be added as a env variable or in ocs-ci/data/auth.yaml file.

Use the template given below to add the key in auth.yaml
```
external:
  ceph_admin_key: "key"
```

jilju: If the key is not present in env (CEPHADMINKEY) or in auth.yaml , the fixture will need to access the provider cluster. In this case provider cluster details should be passed in the run-ci command.

OdedViner (dismissed):

fbalak (dismissed):
Introduced changes could have been added to the [original tool box fixture](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/conftest.py#L3246) but we can keep it separate.

tests/conftest.py
```diff
@@ -4618,3 +4619,67 @@ def finalizer():
...
+        config.multicluster
+        and config.ENV_DATA.get("platform", "").lower()
+        in constants.MANAGED_SERVICE_PLATFORMS
+    ):
```
petr-balogh, line 4638:
```suggestion
    if not (
        config.multicluster
        and config.ENV_DATA.get("platform", "").lower()
        in constants.MANAGED_SERVICE_PLATFORMS
    ):
        return
```

You can save the one indention in the whole block below.

Also as it's session scope fixture and in the case it's the deployment there will not be any OCP/ODF installed. In such case you should skip this code as well.

tests/conftest.py
```diff
@@ -4618,3 +4619,67 @@ def finalizer():
...
+
+
+@pytest.fixture(scope="session", autouse=True)
+def patch_consumer_toolbox_with_secret():
+    """
+    Patch the rook-ceph-tools deployment with ceph.admin key. Applicable for MS platform only to enable rook-ceph-tools
+    to run ceph commands until we have the fix for rook-ceph-tools in consumer cluster
+
+    """
+    current_cluster = config.cluster_ctx
```
petr-balogh, line 4631: This can be moved also after the condition

petr-balogh (dismissed):

tests/conftest.py
jilju: Made the changes

tests/conftest.py
jilju: Done

tests/conftest.py
```diff
@@ -4618,3 +4619,70 @@ def finalizer():
...
+    # Get the secret from provider if MS multicluster run
+    if not (
+        config.multicluster
+        and config.ENV_DATA.get("platform", "").lower()
+        in constants.MANAGED_SERVICE_PLATFORMS
+        and not config.RUN["cli_params"].get("deploy")
+    ):
+        return
+
+    current_cluster = config.cluster_ctx
```
petr-balogh, line 4640: Isn't context directly stored in this variable:
https://github.com/red-hat-storage/ocs-ci/blob/bcd66b7bfd3366ec3c4f67eba37183218daff0bb/ocs_ci/framework/__init__.py#L135

tests/conftest.py
```diff
@@ -4618,3 +4619,70 @@ def finalizer():
...
+                f'[{{"op": "replace", "path": "/spec/template/spec/containers/0/env", '
+                f'"value":[{{"name": "ROOK_CEPH_USERNAME", "value": "client.admin"}}, '
+                f'{{"name": "ROOK_CEPH_SECRET", "value": "{ceph_admin_key}"}}]}}]'
+            )
+            assert consumer_tools_deployment.patch(
+                params=patch_value, format_type="json"
+            ), "Failed to patch rook-ceph-tools deployment in consumer cluster"
+
+    log.info("Switching back to the initial cluster context")
+    config.switch_ctx(current_cluster.MULTICLUSTER["multicluster_index"])
```
petr-balogh, line 4688:
```suggestion
    config.switch_ctx(current_cluster)
```
If you will apply change above than you can directly access index. Also I would rename to something like here:
https://github.com/red-hat-storage/ocs-ci/blob/bc284be38fb1dc725333253f0d481ac13d596a62/ocs_ci/utility/utils.py#L514

tests/conftest.py
jilju: Yes, correct. Made the change.

tests/conftest.py
jilju: Done

petr-balogh (approved):

jilju:
Verified using https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/12/console
The cluster was accidentally destroyed before the completion of the job. The fixture added in the PR was executed. The deployment rook-ceph-tools in the consumer was patched. Verified the change manually in the consumer cluster.
Marking the PR as verified.

jilju:
Verified in https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/13/console

nehaberry (approved):
Changes LGTM and we may need to keep this as in Managed services, the consumer toolbox is allowed to run very few commands and that too with extra params

PR #5665, opened 2022-04-01
Provider endpoint check

fbalak:

### DIFF
--- a/ocs_ci/utility/managedservice.py
+++ b/ocs_ci/utility/managedservice.py
@@ -108,7 +108,7 @@ def get_storage_provider_endpoint(wait=False, timeout=1080):
     """
     config.switch_to_provider()
 
-    def _get_provider_endpoint:
+    def _get_provider_endpoint():
         oc = ocp.OCP(namespace="openshift-config")
         oc.exec_oc_cmd(
             "get storagecluster -n openshift-storage -o=jsonpath="
--- a/ocs_ci/utility/managedservice.py
+++ b/ocs_ci/utility/managedservice.py
@@ -106,7 +106,14 @@ def get_storage_provider_endpoint(wait=False, timeout=1080):
         str: value of storage provider endpoint
 
     """
-    config.switch_to_provider()
+    # TODO(fbalak): use following command when https://github.com/red-hat-storage/ocs-ci/pull/5541
+    # is merged:
+    # config.switch_to_provider()
+    for i, cluster in enumerate(config.clusters):
+        if cluster.ENV_DATA["cluster_type"] == "provider":
+            provider_id = i
+            break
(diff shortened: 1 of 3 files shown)


dahorak:
I would propose to keep the possibility to pass the storage provider endpoint via configuration file and only if it will not be provided, then try to get it dynamically.

ocs_ci/utility/managedservice.py
```diff
@@ -92,29 +92,45 @@ def remove_header_footer_from_key(key):
...
         str: value of storage provider endpoint
 
     """
+    # TODO(fbalak): use following command when https://github.com/red-hat-storage/ocs-ci/pull/5541
+    # is merged:
+    # config.switch_to_provider()
+    for i, cluster in enumerate(config.clusters):
+        if cluster.ENV_DATA["cluster_type"] == "provider":
+            provider_id = i
+            break
```
mbukatov, line 115: What if you don't find the provider?

fbalak: This will be replaced with a function from config module that raises ClusterNotFoundException if provider not found.

ocs_ci/utility/managedservice.py
```diff
@@ -92,29 +92,48 @@ def remove_header_footer_from_key(key):
...
+            for result in TimeoutSampler(
+                timeout=timeout, sleep=5, func=_get_provider_endpoint
+            ):
+                if result:
+                    provider_endpoint = result
+                    break
+        else:
+            provider_endpoint = _get_provider_endpoint()
+    finally:
+        config.reset_ctx()
```
yitzhak12, line 137: Why do we need this method here? If we want the original cluster, I think we had better save the original cluster index and use the method `switch_ctx` when passing that index.

ocs_ci/utility/managedservice.py
```diff
@@ -92,29 +92,48 @@ def remove_header_footer_from_key(key):
...
     """
-    Get get_storage_provider_endpoint
+    Get storage provider endpoint from Provider storage cluster or from
+    configuration if configured.
 
     Args:
-        cluster (str): cluster name
+        wait (bool): If true then wait for the value to be available for
+            number of seconds defined in timeout parameter. If false then
+            trye to return value only once.
```
yitzhak12, line 103:
```suggestion
            try to return value only once.
```

ocs_ci/utility/managedservice.py
fbalak: If `reset_ctx` works then I would prefer it to longer code construct that saves current context id and switches it back. I added https://github.com/red-hat-storage/ocs-ci/issues/5831 to better document those methods.

PR #5682, opened 2022-04-06
rephrase pullsecret_workaround parameter

suchita-g:
The pullsecrete_workaround only if execution is
on non-gaed version so instead of modifying
configs after and before release.Hence seperating
this configuration from other generic configs

### DIFF
--- a/conf/deployment/rosa/managed_1az_consumer_qe_3m_3w_m52x.yaml
+++ b/conf/deployment/rosa/managed_1az_consumer_qe_3m_3w_m52x.yaml
@@ -21,7 +21,6 @@ DEPLOYMENT:
   rosa_cli_version: '1.1.12'
   force_download_installer: True
   force_download_client: True
-  pullsecret_workaround: True
 REPORTING:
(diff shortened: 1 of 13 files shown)


conf/deployment/rosa/pullsecrete_for_nonga.yaml
```diff
@@ -0,0 +1,4 @@
+DEPLOYMENT:
```
fbalak, line 1: pullsecrete_for_nonga.yaml -> pullsecret_for_nonga.yaml

fbalak (commented):
Will you also move the workaround to be done after post deployment checks?

conf/deployment/rosa/pullsecrete_for_nonga.yaml
suchita-g: Thank you for pointing this out. renamed file now.

fbalak (approved):

jilju (approved):

mbukatov (approved):

PR #5697, opened 2022-04-08
ACM Hub Downstream Installation

clacroix12:


clacroix12:
Raising this PR so you can see where I'm currently at @petr-balogh. As of right now I still don't understand why the shell script is behaving differently when being executed from python as opposed to being run manually. I am performing the same exact steps in my python automation as I am locally before running the script. The line that is first tripping up in the shell script is [here](https://github.com/stolostron/deploy/blob/master/start.sh#L215). Full output from the script is below. I'm going through each step in the process and trying to see if I'm seeing something incorrectly in my code but as far as I can tell I'm doing everything that I'm doing locally. 

1. Cloning the deploy repo
2. Exporting environment variables
3. Writing pull-secret.yaml
4. Creating the ImageContentSourcePolicy
5. Writing to snapshot.ver
6. Running the script with `--silent`

Usage Errors:
[start_script_usage_error.txt](https://github.com/red-hat-storage/ocs-ci/files/8455688/start_script_usage_error.txt)

Successful Execution for comparison:
[start_script_success.txt](https://github.com/red-hat-storage/ocs-ci/files/8455787/start_script_success.txt)

ocs_ci/deployment/deployment.py
```diff
@@ -1172,6 +1175,94 @@ def deploy_acm_hub(self):
...
+        ci_dir = os.getcwd()
+        os.chdir(acm_hub_downstream_deploy_dir)
+        logger.info("Running open-cluster-management deploy")
+        cmd = [os.path.join(acm_hub_downstream_deploy_dir, "start.sh"), "--silent"]
+        logger.info("Running cmd: %s", " ".join(cmd))
+        proc = Popen(
+            cmd,
+            stdout=PIPE,
+            stderr=PIPE,
+            encoding="utf-8",
```
petr-balogh, line 1251: Not sure if it can be the case but can you try with shell=True?

```
If shell is True, the specified command will be executed through the shell. This can be useful if you are using Python primarily for the enhanced control flow it offers over most system shells and still want convenient access to other shell features such as shell pipes, filename wildcards, environment variable expansion, and expansion of ~ to a user’s home directory. However, note that Python itself offers implementations of many shell-like features (in particular, [glob](https://docs.python.org/3/library/glob.html#module-glob), [fnmatch](https://docs.python.org/3/library/fnmatch.html#module-fnmatch), [os.walk()](https://docs.python.org/3/library/os.html#os.walk), [os.path.expandvars()](https://docs.python.org/3/library/os.path.html#os.path.expandvars), [os.path.expanduser()](https://docs.python.org/3/library/os.path.html#os.path.expanduser), and [shutil](https://docs.python.org/3/library/shutil.html#module-shutil)).
```

In the script itself they are using the pipes and so on, but not sure if it might be the case if those are not directly used in the command.

Have you tried the debug output with added:

set -x 

Or you can directly run command like:

 bash -x script-name.sh
 
 this will not need you to change the script which might be better option for you.
 
 I am really curious to see debug output to see what happens inside the script.

ocs_ci/deployment/deployment.py
```diff
@@ -1172,6 +1175,94 @@ def deploy_acm_hub(self):
...
+        tag_data = "2.4.3-DOWNSTREAM-2022-04-02-03-50-40"
+        with open(
+            os.path.join(acm_hub_downstream_deploy_dir, "snapshot.ver"), "w"
+        ) as f:
+            f.write(tag_data)
+
+        ci_dir = os.getcwd()
+        os.chdir(acm_hub_downstream_deploy_dir)
+        logger.info("Running open-cluster-management deploy")
+        cmd = [os.path.join(acm_hub_downstream_deploy_dir, "start.sh"), "--silent"]
```
petr-balogh, line 1245: do you need to `os.path.join(acm_hub_downstream_deploy_dir, "start.sh")` when you change dir to that path?

Shouldn't be enough just `./start.sh` ?
Also path can be set in Popen by cwd parameter.

ocs_ci/deployment/deployment.py
```diff
@@ -1172,6 +1175,94 @@ def deploy_acm_hub(self):
...
+        logger.info("Running cmd: %s", " ".join(cmd))
+        proc = Popen(
+            cmd,
+            stdout=PIPE,
+            stderr=PIPE,
```
petr-balogh, line 1250: I know that env should be taken from current process. But if you see still some issues that those env variable are not propagated well you can try to explicitly pass it here via env argument.

ocs_ci/deployment/deployment.py
```diff
@@ -1172,6 +1175,94 @@ def deploy_acm_hub(self):
...
         Handle ACM HUB deployment
         """
+        if config.ENV_DATA.get("acm_hub_downstream"):
+            self.deploy_acm_hub_downstream()
+        else:
+            self.deploy_acm_hub_upstream()
+
+    def deploy_acm_hub_downstream(self):
+        """
+        Handle AM HUB downstream deployment
```
jimdevops19, line 1185:
```suggestion
        Handle ACM HUB downstream deployment
```

ocs_ci/deployment/deployment.py
clacroix12: I tried passing `env` explicitly with no change to the behavior of the script. As its more work to pull all of the base env vars and add them to the dict that is passed to the command, I opted to stick with the default behavior.

ocs_ci/deployment/deployment.py
clacroix12: Moved to using `./start.sh` and specified `cwd`

clacroix12:
The issue ended up being that `yq` was specified in our python requirements for ocs-ci. This package of `yq` is apparently not updated as they only have much older versions of the package. This script requires `4.12+` and the python package only has `2.x` versions. I have removed this as a python requirement as it is primarily a system package that is a requirement of some shell scripts we execute and not a requirement of any of our python code. After removing this from my venv I was able to successfully execute this workflow and verification succeeded.

There is still one more task left for this PR, and that is to specify the downstream images in the OCP config files. I have reached out on slack to the acm team and will update this PR once I know which images map to which OCP versions.

clacroix12:
As per the discussion with @petr-balogh and @shylesh today, this will also be blocked by the MR to install yq on our jenkins agents. I will also update our prerequisites documentation to reflect the need to install this manually if you are running on OSX.

clacroix12:
Tested locally with `yq (https://github.com/mikefarah/yq/) version 4.24.2`

Snippets from local logs:
```
* multiclusterhub-operator is running

* Beginning deploy...
* Applying the multiclusterhub-operator to install Red Hat Advanced Cluster Management for Kubernetes
multiclusterhub.operator.open-cluster-management.io/multiclusterhub created

#####
Wait for multicluster-operators-application to reach running state (4min).
* multicluster-operators-application is running
Search will only be automaticaly enabled with '--watch' is set.
Run 'oc set env deploy search-operator DEPLOY_REDISGRAPH=true -n open-cluster-management' to enable search.

#####
* Red Hat ACM URL: https://multicloud-console.apps.clacroix-acm.qe.rh-ocs.com
#####
```
```
12:17:10 - MainThread - ocs_ci.deployment.deployment - INFO - MultiClusterHub Deployment Succeeded
```

I will test again through jenkins once I also implement the acm hub downstream image versions in the OCP version files (still waiting for a response on this).

clacroix12:
I have removed the change to setup.py and will instead raise this in a separate PR so we can merge it quickly along with its corresponding [MR](https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/-/merge_requests/1008).

prsurve:
> PR validation
> 
> Cluster Name: Cluster Configuration: conf/ocsci/acm_hub_downstream_deploy.yaml conf/ocsci/skip_ocs_deploy.yaml PR Test Suite: deployment PR Test Path: tests/ Additional Test Params: OCP VERSION: 4.10 OCS VERSION: 4.10 tested against branch: master
> 
> [Job PASSED](https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-trigger-test-pr/2656//testReport).

I see that no test were nor deployment was ran here

petr-balogh (dismissed):

conf/ocsci/acm_hub_downstream_image.yaml
```diff
@@ -0,0 +1,3 @@
+---
+ENV_DATA:
+  acm_downstream_image: "2.4.3-DOWNSTREAM-2022-04-13-07-05-00"
```
prsurve, line 3: Any reason for hardcoded values? Also, it's for 2.4.3 we need for 2.5

petr-balogh: @prsurve how we will match acm version? With OCP? Like in ocp 4.10 we will use 2.5 and in 4.11 different version? I thought about to add it in our ocp version config files.

clacroix12: This is really just an example config and not one that will be used by jenkins. I chose this version because this is the last RC version, where as 2.5 does not have one yet.

@petr-balogh I have not talked to them about adding a floating tag. I can bring this up with them but we probably can't count on having this in the short term. I expected that we would add a field to the jenkins deployment job to specify the tag that we wish to deploy. 

Also, I spoke to them about compatibility and was given some information on which versions are supported for which OCP versions. This version works for any version we are currently actively testing with acm hub (4.9 and up).

prsurve: For now, we can use the 2.5 build

clacroix12:
Verified with a deploy job here so I could skip the test phase since that is not supported on OCP+ACM only clusters:

https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/12180/console

conf/ocsci/acm_hub_downstream_image.yaml
clacroix12: I have move this file to the `examples` dir and updated to the latest 2.5 image. @petr-balogh I also brought up the topic of `latest-x.y` tags and have started that discussion. Will update you on it.

petr-balogh (approved):

shylesh (approved):
LGTM

vavuthu (approved):

dahorak (approved):

Issue #3688, opened 2021-01-14
Cleanup storage class naming code and place it into a separate ocs module


mbukatov:
Place storage class names and related code into new dedicated `ocs_ci.ocs.scnaming` module.

There are lot of variables related to storage class names in constants.py module, some are duplicated (eg. `CEPHBLOCKPOOL_SC`), and given rising complexity of storage class naming scheme, this could benefit from cleanup and dedicated module where both names and related code are maintained.

Example of storage class related code:

- https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/helpers/helpers.py#L457 (but we should not use OCS/OCP classess unless we have some reason to do so)
- <https://github.com/red-hat-storage/ocs-ci/pull/5698>

PR #5698, opened 2022-04-10
Fix code for running workload suite on Managed Service

OdedViner:

### DIFF
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -2898,6 +2898,8 @@ def storagecluster_independent_check():
         .get()
         .get("items")[0]
     )
+    if config.ENV_DATA["platform"].lower() in constants.MANAGED_SERVICE_PLATFORMS:
+        return False
 
     return bool(
         storage_cluster.get("spec", {}).get("externalStorage", {}).get("enable", False)


nehaberry:
> Do we need this check for consumer cluster ?

@OdedViner IMO, we need independent_mode check + managed services check for consumer cluster, 

So if independent + managed services, then use the SC and storagecluster name without "external" else it is a normal external mode

IMO, if we just return false, our tests might fail for consumers. Assuming I did not misunderstand

nehaberry:
IMO this could be a temp fix for workloads, but we need to find a permanent solution for MS

@jilju @yitzhak12 how did we handle it in tier1 and tier4?

ocs_ci/helpers/helpers.py
```diff
@@ -2898,6 +2898,8 @@ def storagecluster_independent_check():
         .get()
         .get("items")[0]
     )
+    if config.ENV_DATA["platform"].lower() in constants.MANAGED_SERVICE_PLATFORMS:
```
jilju, line 2901: Independent mode check  is needed for consumer cluster. I think the update is needed places where the selection of the storage class is done. 
Or something like this in constants.py
```
DEFAULT_CLUSTERNAME_EXTERNAL_MODE = "ocs-external-storagecluster" if config.ENV_DATA["platform"] not in MANAGED_SERVICE_PLATFORMS else DEFAULT_CLUSTERNAME
```

mbukatov:
Ah, this hack is necessary because while cluster in managed service is in fact deployed in external configuration, it's not using different storage class names unlike plain external cluster mode, which requires different storage class names.

mbukatov (approved):

ocs_ci/helpers/helpers.py
mbukatov: That is a different topic: we don't have a good abstraction for storage class in ocs-ci.

fbalak: I agree with Jilju. This PR would invalidate the check and although there are differences in classic independent mode and managed service, we should keep the check as is to let us know that the cluster is in independent mode if the test is ran on consumer. We can handle differences between platforms at other places.

fbalak (changes requested):

mbukatov:
> IMO this could be a temp fix for workloads, but we need to find a permanent solution for MS

The only real solution is <https://github.com/red-hat-storage/ocs-ci/issues/3688>

PR #5709, opened 2022-04-12
Modify cephfs workload to use in other tests

yitzhak12:
The pr is extended the use of the functionality in the test `tests/e2e/workloads/test_data_consistency.py` to be used in other tests in the project. Also, I added a new variable to check how many times the rsync command failed and an option to ignore several failures - because sometimes the error doesn't relate to the data validation but instead to a temporary error.
The original functionality is still the same. I just divided it into a class and methods so it can be used in other files.

For example, now we can use it as below:

1. Write and read logfile stored on cephfs volume from all cluster worker nodes via k8s Deployment. (method 1)
2. Perform some operations.
3. While the workload is running, we will try to fetch and validate data from the cephfs volume of the workload. (method 2)


yitzhak12:
Regarding the discussion with @mbukatov, I am closing the pr. Soon, he will implement new methods for using the cephfs workload more straightforwardly.

PR #5711, opened 2022-04-12
Do do fail the run if provider is not found

jilju:
The fixture _patch_consumer_toolbox_with_secret_ is currently failing if the provider cluster is not present. Made changes to continue the tests if provider cluster is not present assuming that the toolbox on consumers are already fixed before the start  of the test. 


jilju:
Verified without adding a provider cluster in multicluster job: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/43/

dahorak (approved):

fbalak (approved):

PR #5729, opened 2022-04-17
In the test 'get_ocs_nodes' get only the OCS worker nodes by default unless we specify otherwise

yitzhak12:


tests/manage/z_cluster/nodes/test_toleration.py
```diff
@@ -13,6 +18,8 @@
 logger = logging.getLogger(__name__)
 
 
+@skipif_external_mode
```
Shrivaibavi, line 21: why is this test skipped on ext mode ?? 
IMO the OCP cluster is going to stay intact and also there will be pods running on openshift-storage ns except mons, mgr and osds, I dont see the need to skip this testcase. Can you please recheck once

ocs_ci/ocs/node.py
```diff
@@ -769,10 +769,14 @@ def get_ocs_nodes(num_of_nodes=None):
             it returns all the ocs nodes.
 
     Returns:
-        list: List of ocs nodes
+        list: List of ocs nodes. If not found, it returns None.
 
     """
     ocs_node_names = machine.get_labeled_nodes(constants.OPERATOR_NODE_LABEL)
+    if not ocs_node_names:
+        log.warning("Didn't find any OCS nodes")
```
Shrivaibavi, line 777: Why do we need this, we are not going to run any testcase without OCS nodes, This list can never be empty. Please correct me if my understanding is wrong

yitzhak12: I saw that the list is empty with external mode(4.10) and managed service. See this failure here: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/3887/testReport/tests.manage.z_cluster.nodes.test_check_pods_status_after_node_failure/TestCheckPodsAfterNodeFailure/test_check_pods_status_after_node_failure/.

tests/manage/z_cluster/nodes/test_toleration.py
yitzhak12: The problem is with the `get_ocs_nodes` function. With external mode, it didn't find the OCS nodes.

sidhant-agrawal: I agree with Shrivaibavi, except for ceph related pods we still have other pods in openshift-storage ns.
Perhaps we can consider non-ocs nodes to be used in case of external mode.

yitzhak12:
I am closing the pr. 
I will open a new pr soon regarding fixing the `get_ocs_nodes` function.

PR #5766, opened 2022-04-22
Remove skipif_ms_provider from tests

jilju:
When running MultiCluster job with consumer as the primary cluster and provider as the secondary cluster, the test cases in _tests/manage/pv_services/test_resource_deletion_during_pvc_pod_creation_and_io.py_ and _tests/manage/pv_services/test_resource_deletion_during_pvc_pod_deletion_and_io.py_ are skipped due to the reason 
```
tests/manage/pv_services/test_resource_deletion_during_pvc_pod_deletion_and_io.py::TestResourceDeletionDuringMultipleDeleteOperations::test_disruptive_during_pod_pvc_deletion_and_io[CephBlockPool-mgr]
Reason: Test will not run on Managed service provider cluster
```
This is due to the presence of the marker _skipif_ms_provider_ which should skip these tests only when the cluster context is provider. But the test cases are skipped when the cluster context is consumer.
These test cases will run with multicluster configuration only in MS platform because access to both the consumer and provider clusters are required. Removing the marker _skipif_ms_provider_ from these tests because the marker is not relevant in a configuration with consumer as primary cluster. 
Build - https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/86/

### DIFF
--- a/tests/manage/pv_services/test_resource_deletion_during_pvc_pod_creation_and_io.py
+++ b/tests/manage/pv_services/test_resource_deletion_during_pvc_pod_creation_and_io.py
@@ -8,7 +8,6 @@ from ocs_ci.framework.testlib import (
     tier4,
     tier4c,
     ignore_leftover_label,
-    skipif_ms_provider,
 )
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.resources.pod import (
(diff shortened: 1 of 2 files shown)


jilju:
Verified: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/86/
The test cases are not skipping in the same multcluster configuration after removing the marker _skipif_ms_provider_.

rachael-george (approved):

fbalak (approved):

Issue #3135, opened 2020-10-14
clone_repo() should use a shallow checkout (--depth 1) for repos


mykaul:
It'll drastically lower time to checkout repos.

PR #5778, opened 2022-04-25
Use shallow clone and support clone types in clone_repo function.

vavuthu:
Fixes: #3135

There are 3 different types of clone to improve performance.
1. shallow clone
2. blobless clone
3. treeless clone

Above clone are very helpful for slow network and also it dramatically
reduce the space.

eg: To clone openshift installer

normal takes around ~10 min ( slow network ) and 2.05 GB space while shallow clone takes  31 sec and 1.27 GB space 

$ time git clone https://github.com/openshift/installer.git 
Cloning into ''...
remote: Enumerating objects: 200325, done.
remote: Counting objects: 100% (6/6), done.
remote: Compressing objects: 100% (6/6), done.
remote: Total 200325 (delta 0), reused 1 (delta 0), pack-reused 200319
Receiving objects: 100% (200325/200325), 870.72 MiB | 1.60 MiB/s, done.
Resolving deltas: 100% (130276/130276), done.
Updating files: 100% (83002/83002), done.

real	9m16.119s
user	0m29.385s
sys	0m8.768s

$ time git clone https://github.com/openshift/installer.git  --depth=1
Cloning into ''...
remote: Enumerating objects: 51248, done.
remote: Counting objects: 100% (51248/51248), done.
remote: Compressing objects: 100% (27183/27183), done.
remote: Total 51248 (delta 26173), reused 38408 (delta 21583), pack-reused 0
Receiving objects: 100% (51248/51248), 74.87 MiB | 5.25 MiB/s, done.
Resolving deltas: 100% (26173/26173), done.
Updating files: 100% (82999/82999), done.

real	0m31.998s
user	0m7.082s
sys	0m2.956s

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -2053,7 +2053,7 @@ def get_rook_repo(branch="master", to_checkout=None):
         run_cmd(f"git checkout {to_checkout}", cwd=cwd)
 
 
-def clone_repo(url, location, branch="master", to_checkout=None):
+def clone_repo(url, location, branch="master", to_checkout=None, clone_type=None):
     """
     Clone a repository or checkout latest changes if it already exists at
         specified location.
@@ -2063,10 +2063,23 @@ def clone_repo(url, location, branch="master", to_checkout=None):
         location (str): path where the repository will be cloned to
         branch (str): branch name to checkout
         to_checkout (str): commit id or tag to checkout
+        clone_type (str): type of clone ( shallow, blobless, treeless)
+
     """
+    git_params = ""
+    if clone_type == "shallow":
+        if branch != "master":
+            git_params = "--no-single-branch --depth=1"
+        else:
+            git_params = "--depth=1"
+    if clone_type == "blobless":
+        git_params = "--filter=blob:none"
+    if clone_type == "treeless":
+        git_params = "--filter=tree:0"
+
     if not os.path.isdir(location):
         log.info("Cloning repository into %s", location)
-        run_cmd(f"git clone {url} {location}")
+        run_cmd(f"git clone {git_params} {url} {location}")
     else:
         log.info("Repository already cloned at %s, skipping clone", location)
         log.info("Fetching latest changes from repository")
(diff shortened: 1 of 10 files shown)


ocs_ci/utility/utils.py
```diff
@@ -2053,7 +2053,7 @@ def get_rook_repo(branch="master", to_checkout=None):
         run_cmd(f"git checkout {to_checkout}", cwd=cwd)
 
 
-def clone_repo(url, location, branch="master", to_checkout=None):
+def clone_repo(url, location, branch="master", to_checkout=None, clone_type=None):
```
petr-balogh, line 2056: Do we want shallow to be default?

petr-balogh (dismissed):

ocs_ci/utility/utils.py
vavuthu: currently I am leaving that option to users because users has to decide which option is best fit for their use. May be in future we can default to shallow and I don't have exact word for normal clone

clacroix12 (dismissed):

ocs_ci/utility/utils.py
clacroix12: As the other types of clones are typically used in development, I think `shallow` would be a fine default. I think the general use case for `clone_repo` is to download a repo/branch, run a script, then delete the repo afterwards without utilizing the commit history or trees in any way.

We could change the default to `shallow` as long as we specify in the docstring that we default to a shallow clone and a normal clone will be performed if the value is `None`. That being said, either way works for me as this is a good change.

vavuthu: done

ocs_ci/utility/utils.py
```diff
@@ -2063,10 +2063,26 @@ def clone_repo(url, location, branch="master", to_checkout=None):
         location (str): path where the repository will be cloned to
         branch (str): branch name to checkout
         to_checkout (str): commit id or tag to checkout
+        clone_type (str): type of clone (shallow, blobless, treeless). If any
+            parameter is passed other than above 3 mentioned, its considered as normal clone.
```
petr-balogh, line 2067: I would rather be more specific and tell user to pass None if they want normal clone, or just specify normal as one of the option.

petr-balogh (approved):

PR #5814, opened 2022-04-28
Variable seperation

Avilir:
In this PR i am introducing a mechanism to separate the test variables from the test code.
I removed the **@pytest.mark.parametrize**  from the TC and replace them with parameters which i read from yaml file that provided via the `--ocsci-conf` parameter.
And instead several individual runs of the same test with different parameters, it will run only one test with loop(s) for the parameter(s).
each test can implement this mechanism  in its own way, no mandatory parameter except the **test name**.
the test in this PR is just an example on how to implement this.

this will give us the flexibility to run test with different values, without changing the code.


Avilir:
> PR validation
> 
> Cluster Name: Cluster Configuration: PR Test Suite: PR Test Path: tests/e2e/performance/test_pvc_attach_new.py Additional Test Params: OCP VERSION: 4.10 OCS VERSION: 4.10 tested against branch: master
> 
> [Job PASSED](https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-trigger-test-pr/2728//testReport).

This Job ran with parameters in default configuration file.

tests/e2e/performance/test_pvc_attach_new.py
```diff
@@ -0,0 +1,308 @@
...
+            pod.delete(wait=False)
+        performance_lib.wait_for_resource_bulk_status(
+            resource="pod",
+            resource_count=0,
+            namespace=self.namespace,
+            status=constants.STATUS_RUNNING,
+            timeout=timeout,
+            sleep_time=sleeptime,
+        )
+        log.info("All pods was deleted")
```
ypersky1980, line 155: small spelling correction - "were" instead of "was"

tests/e2e/performance/test_pvc_attach_new.py
```diff
@@ -0,0 +1,308 @@
...
+        # Push the test results into the performance dashboard - only if ALL tests (2) ran, if not, don't try to push
+        # the results into the Performance dashboard
+        if len(self.params["interfaces"]) == 2:
+            self.add_test_to_results_check(
+                test="test_pod_start_time", test_count=2, test_name="PVC Attach Time"
+            )
+            try:
+                self.check_results_and_push_to_dashboard()
+            except Exception as exp:
+                log.error(f"Can not push the results into the performance DB [{exp}]")
```
ypersky1980, line 308: small spelling fix - "Cannot" instead of "Can not".

ypersky1980 (approved):
LGTM

dahorak:
@Avilir This is interesting idea. I have two questions/notes:
* If I didn't miss anything, it is automatically loading all configuration files from the `ocs_ci/framework/conf/tests.d/` directory, right? I'm thinking that if it will be used by higher number of tests, it will significantly extend the size of loaded configuration. But I'm not sure, if it might introduce some issue or not, though. Also if you would like to override existing test configuration, you will have to be careful how to override the parameters you want and not anything else (but if I remember it correctly, by providing custom configuration for particular testcase, you will always override the whole testcase configuration - which is probably ok, just people will have to count with that).
* Second and more important think is, that since you are running all the configurations in a loop in one test, it will not report each configuration separately, right? With standard `@pytest.mark.parametrize` approach, each configuration have own result reported like this:
2022-05-02_individual_test_configuration_reporting

Avilir:
> @Avilir This is interesting idea. I have two questions/notes:
> 
> * If I didn't miss anything, it is automatically loading all configuration files from the `ocs_ci/framework/conf/tests.d/` directory, right? I'm thinking that if it will be used by higher number of tests, it will significantly extend the size of loaded configuration. But I'm not sure, if it might introduce some issue or not, though. Also if you would like to override existing test configuration, you will have to be careful how to override the parameters you want and not anything else (but if I remember it correctly, by providing custom configuration for particular testcase, you will always override the whole testcase configuration - which is probably ok, just people will have to count with that).
> * Second and more important think is, that since you are running all the configurations in a loop in one test, it will not report each configuration separately, right? With standard `@pytest.mark.parametrize` approach, each configuration have own result reported like this:
>   2022-05-02_individual_test_configuration_reporting

Thank you @dahorak for your comments.
* you didn't miss anything, i am reading the tests configuration before all other conf files (default and --ocsci-conf files) are read, so every file that provided via CLI is overwrite the default. regarding the size of the config object, which will be large as the amount of files growth, i am opening for suggestion to reduce it.
* I am aware to the fact that in this methodology, the test case will report all  **parametrize** as a single test and maybe this is suitable to performance tests and not all other tests, or can be implemented in a different way so all results of individual test will collected and report in the end of the test - i will try to implement it to show example of it.

Avilir:
In last commit, i add a global report - in the log - for all individual tests, example from this test case : 
```
13:12:18 - MainThread - test_pvc_attach_new - INFO - Full Test Repport.
13:12:18 - MainThread - test_pvc_attach_new - INFO - Keys are : SC_Interface[PVC_Size-Number_of_Samples]
13:12:18 - MainThread - test_pvc_attach_new - INFO - CephBlockPool[5-5]  :  PASS
13:12:18 - MainThread - test_pvc_attach_new - INFO - CephFileSystem[5-5]  :  PASS
PASSED
------------------------------ live log teardown -------------------------------
```

or 
```
13:03:08 - MainThread - test_pvc_attach_new - INFO - Full Test Repport.
13:03:08 - MainThread - test_pvc_attach_new - INFO - Keys are : SC_Interface[PVC_Size-Number_of_Samples]
13:03:08 - MainThread - test_pvc_attach_new - INFO - CephBlockPool[5-5]  :  FAILED
13:03:08 - MainThread - test_pvc_attach_new - INFO - CephFileSystem[5-5]  :  PASS
FAILED
------------------------------ live log teardown -------------------------------
```
so, yes, in the pytest report it will report this as a single test, and it will pass **only** if all individual tests pass, but in the end of the log we will have detailed report for individual tests.

ocs_ci/framework/main.py
```diff
@@ -128,6 +138,8 @@ def process_ocsci_conf(arguments):
     )
 
     args, unknown = parser.parse_known_args(args=arguments)
+    marker = arguments[arguments.index("-m") + 1] if "-m" in arguments else ""
+    load_config(get_tests_config(marker))
```
petr-balogh, line 142: This is very problematic part - as e.g. for upgrade we are using this:
```
-m pre_upgrade or pre_ocs_upgrade or ocs_upgrade or post_ocs_upgrade or post_upgrade
```

For post upgrade execution we are using:
```
-m tier1 or tier_after_upgrade
```

And it can be whatever expression user will use with and and or operators and brackets.

Avilir: This is a grate example to use configuration file.
You can use a different configuration for different markers on the same test.
We also can have a "global" directory, for test which uses the same parameters with different markers.

Avilir:
with latest commit i add the ability to have the tests as single tests wich will preserve the same reporting that we have.
i ran the collect-only locally and you can see in the output :
```
(ocs-ci) avili@vm-11-87 [OCS-CI] (variable_seperation) # run-ci --cluster-name avili-perfs --cluster-path /home/avili/Testing/Clusters/DCPSCL --ocsci-conf conf/avili/dcPS.yaml tests/e2e/performance/test_pvc_attach_new.py --co
2022-05-08 11:33:07,664 - MainThread - ocs_ci.framework.pytest_customization.ocscilib - INFO - Pytest configure switching to: cluster=0
========================================================================================================================= test session starts =========================================================================================================================
platform linux -- Python 3.9.12, pytest-6.2.5, py-1.11.0, pluggy-0.13.1
rootdir: /home/avili/Testing/OCS-CI, configfile: pytest.ini
plugins: metadata-1.11.0, ordering-0.6, marker-bugzilla-0.9.4, logger-0.5.1, html-3.1.1, flaky-3.7.0, repeat-0.9.1, reportportal-1.10.3
collected 5 items                                                                                                                                                                                                                                                     

<Module tests/e2e/performance/test_pvc_attach_new.py>
  <Class TestPodStartTime>
      <Function test_pod_start_time>
  <Class TestPodBulkStartTime>
      <Function test_bulk_start_time[CephBlockPool-120]>
      <Function test_bulk_start_time[CephBlockPool-240]>
      <Function test_bulk_start_time[CephFileSystem-120]>
      <Function test_bulk_start_time[CephFileSystem-240]>

===================================================================================================================== 5 tests collected in 0.17s ======================================================================================================================
(ocs-ci) avili@vm-11-87 [OCS-CI] (variable_seperation) # 
``` 
the first test class have only one test which will run with parameters, while in the second test class there are 4 individual tests.
in both cases the variables are in a configuration file.

ocs_ci/utility/utils.py
```diff
@@ -3647,3 +3650,44 @@ def decode(encoded_message):
     decoded_base64_bytes = base64.b64decode(encoded_message_bytes)
     decoded_message = decoded_base64_bytes.decode("ascii")
     return decoded_message
+
+
+def get_parameters(test_name="", args_list=None):
+    """
+    Converting test configuration data from file into pytest.parametrize variables
+    if the test_names dosent have variables in the configuration, raising an exception
```
ypersky1980, line 3658: small spelling fix - "dosen't instead of dosent"

ypersky1980 (approved):
LGTM

ypersky1980 (approved):
LGTM

PR #5823, opened 2022-05-01
WIP

yosibsh:
OMADOps stands for "OCS Monkey for Admin Operations".
This test suite runs for ~24h while executing
both Admin operations such as node reboot,
node replacement, add capacity and etc
as well as User operations such as running
different workloads (pgsql, bdi, couchbase, ...)

Our goal is to make sure that the cluster is healthy
and operational during the test and after.

This version is structural ONLY.

Signed-off-by: Yosi Ben Shimon <ybenshim@redhat.com>

### DIFF
--- a/ocs_ci/helpers/omadops_helpers.py
+++ b/ocs_ci/helpers/omadops_helpers.py
@@ -1,0 +1,165 @@
+import uuid
+import time
+import threading
+import random
+import logging
+from datetime import datetime
+
+from ocs_ci.ocs.platform_nodes import PlatformNodesFactory
+from ocs_ci.ocs.node import get_nodes
+import tests.OMADOps.conftest as omadops_config
+from ocs_ci.ocs.longevity import Longevity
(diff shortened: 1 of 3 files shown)


ocs_ci/helpers/omadops_helpers.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    return datetime.now() < omadops_config.END_TIME
+
+
+def run_user_ops():
+    """
+    User Ops function:
+      - Runs while in timeframe (configured in tests.OMADOps.config.EXECUTION_TIME_HOURS)
+      - Steps:
+        1. Loop through the list and check if elements are None or thread.is_alive()
```
hnallurv, line 33: app workloads and user operations have to be running in the background in parallel to the selected random admin ops. Longevity tests automates several user ops and workloads running in the backgrounds. Let's utilize them here as we discussed earlier.

PR #5840, opened 2022-05-05
move test_stop_worker_nodes_pd to tier1

fbalak:


mbukatov (changes requested):
Tier1 should not include negative test cases.

PR #5848, opened 2022-05-06
Managed Service pod disruption test

jilju:
RHSTOR-3176 - Managed Service specific pod disruption test

I/O will be running on all consumers when performing pod deletion.

Provider pods to delete:
ocs-osd-controller-manager
alertmanager-managed-ocs-alertmanager
ocs-provider-server
prometheus-operator
prometheus-managed-ocs-prometheus
ocs-operator

Consumer pods to delete:
alertmanager-managed-ocs-alertmanager
ocs-osd-controller-manager
prometheus-managed-ocs-prometheus
prometheus-operator
ocs-operator

### DIFF
--- a/ocs_ci/helpers/disruption_helpers.py
+++ b/ocs_ci/helpers/disruption_helpers.py
@@ -125,6 +125,9 @@ class Disruptions:
         if self.resource == "prometheus_operator":
             self.resource_obj = [pod.get_prometheus_operator_pod()]
             self.selector = constants.PROMETHEUS_OPERATOR_LABEL
+        if self.resource == "ocs_provider_server":
+            self.resource_obj = [pod.get_ocs_provider_server_pod()]
+            self.selector = constants.PROVIDER_SERVER_LABEL
 
         self.resource_count = resource_count or len(self.resource_obj)
 
--- a/ocs_ci/helpers/disruption_helpers.py
+++ b/ocs_ci/helpers/disruption_helpers.py
@@ -103,6 +103,28 @@ class Disruptions:
         if self.resource == "operator":
             self.resource_obj = pod.get_operator_pods()
             self.selector = constants.OPERATOR_LABEL
+        if self.resource == "ocs_operator":
+            self.resource_obj = [pod.get_ocs_operator_pod()]
+            self.selector = constants.OCS_OPERATOR_LABEL
+        if self.resource == "alertmanager_managed_ocs_alertmanager":
+            self.resource_obj = pod.get_alertmanager_managed_ocs_alertmanager_pods()
+            self.selector = constants.MANAGED_ALERTMANAGER_LABEL
+        if self.resource == "ocs_osd_controller_manager":
+            self.resource_obj = [pod.get_ocs_osd_controller_manager_pod()]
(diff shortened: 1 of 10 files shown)


jilju:
Verified: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/152/

jilju:
Verified - [test report](http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/jijoy-m11-c1/jijoy-m11-c1_20220511T065349/multicluster/logs/test_report_1652273313.html)

ocs_ci/helpers/disruption_helpers.py
```diff
@@ -93,6 +103,31 @@ def set_resource(self, resource, leader_type="provisioner"):
         if self.resource == "operator":
             self.resource_obj = pod.get_operator_pods()
             self.selector = constants.OPERATOR_LABEL
+        if self.resource == "ocs_operator":
+            self.resource_obj = [pod.get_ocs_operator_pod()]
+            self.selector = constants.OCS_OPERATOR_LABEL
+        if self.resource == "alertmanager_managed_ocs_alertmanager":
+            self.resource_obj = pod.get_alertmanager_managed_ocs_alertmanager_pods()
+            self.selector = constants.MANAGED_ALERTMANAGER_LABEL
+        if self.resource == "ocs_osd_controller_manager":
```
OdedViner, line 112: Can you use `elif`?

jilju: @OdedViner 
The current structure of conditional statements was followed when adding the new pod types. Would you like to change the previous ```if``` conditions to ```elif``` ?

OdedViner: got it... I dont think you need to change it `elif`..

ocs_ci/ocs/resources/pod.py
```diff
@@ -695,6 +701,107 @@ def get_ocs_operator_pod(ocs_label=constants.OCS_OPERATOR_LABEL, namespace=None)
...
+        Pod: Pod object of prometheus-operator pod
+
+    """
+    namespace = namespace or config.ENV_DATA["cluster_namespace"]
+    prometheus_operator = get_pods_having_label(label, namespace)
+    return Pod(**prometheus_operator[0])
+
+
+def get_ocs_provider_server_pod(label=constants.PROVIDER_SERVER_LABEL, namespace=None):
+    """
```
ebondare, line 789: Do we really need so many functions that are so similar? Can't we have a more general one that takes label as an argument?

jilju: @ebondare 
The function 'get_pods_having_label' which is used inside this function, is actually the generic one. The idea was to keep the conditional statements in the Disruptions class in the same way it is originally designed. If separate functions are not present, the function 'set_resource' in the class Disruptions will look bulky.
Still it can work without the separate functions. We can change the original structure(which is not part of this PR) of the function  'set_resource' if needed. WDYT ?

OdedViner (approved):

ebenahar (approved):

Issue #5851, opened 2022-05-09
destroy job failed on partially install deployment-external RHCS


vavuthu:
destroy job failed on partially install deployment-external RHCS

```
2022-05-09 15:11:03      
2022-05-09 15:11:03  >       assert running_ct_pods, "No running Ceph tools pod found"
2022-05-09 15:11:03  E       AssertionError: No running Ceph tools pod found
```
job: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-destroy-ocs-cluster/14030/console

mbukatov:
Does ceph toolbox pod feature have a support for external clusters?

vavuthu:
> Does ceph toolbox pod feature have a support for external clusters?

ceph tool box on ocp cluster

petr-balogh:
> Does ceph toolbox pod feature have a support for external clusters?

If I remember correctly it's not officially supported.
But we know how to set it that way it works.
You can see here:
https://github.com/red-hat-storage/ocs-ci/blob/35853ce0bb2a7efd9469267b3b1b2407eedc82f6/ocs_ci/ocs/utils.py#L775

Admin keyring needs to be set + few more things.

PR #5852, opened 2022-05-09
catch AssertionError for destroy job - External RHCS

vavuthu:
If the external cluster deployment is failed, then cluster doesn't contains
ceph tool box. During those cases, destroy job will fail with AssertionError

Fixes: #5851 

### DIFF
--- a/ocs_ci/deployment/vmware.py
+++ b/ocs_ci/deployment/vmware.py
@@ -752,11 +752,14 @@ class VSPHEREUPI(VSPHEREBASE):
         # remove csi users in case of external deployment
         if config.DEPLOYMENT["external_mode"]:
             logger.debug("deleting csi users")
-            toolbox = pod.get_ceph_tools_pod()
-            toolbox.exec_cmd_on_pod("ceph auth del client.csi-cephfs-node")
-            toolbox.exec_cmd_on_pod("ceph auth del client.csi-cephfs-provisioner")
-            toolbox.exec_cmd_on_pod("ceph auth del client.csi-rbd-node")
-            toolbox.exec_cmd_on_pod("ceph auth del client.csi-rbd-provisioner")
+            try:
+                toolbox = pod.get_ceph_tools_pod()
+                toolbox.exec_cmd_on_pod("ceph auth del client.csi-cephfs-node")
+                toolbox.exec_cmd_on_pod("ceph auth del client.csi-cephfs-provisioner")
+                toolbox.exec_cmd_on_pod("ceph auth del client.csi-rbd-node")
+                toolbox.exec_cmd_on_pod("ceph auth del client.csi-rbd-provisioner")
+            except AssertionError as ex:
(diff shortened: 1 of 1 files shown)


ocs_ci/deployment/vmware.py
```diff
@@ -752,11 +752,14 @@ def destroy_cluster(self, log_level="DEBUG"):
...
-            toolbox.exec_cmd_on_pod("ceph auth del client.csi-rbd-node")
-            toolbox.exec_cmd_on_pod("ceph auth del client.csi-rbd-provisioner")
+            try:
+                toolbox = pod.get_ceph_tools_pod()
+                toolbox.exec_cmd_on_pod("ceph auth del client.csi-cephfs-node")
+                toolbox.exec_cmd_on_pod("ceph auth del client.csi-cephfs-provisioner")
+                toolbox.exec_cmd_on_pod("ceph auth del client.csi-rbd-node")
+                toolbox.exec_cmd_on_pod("ceph auth del client.csi-rbd-provisioner")
+            except AssertionError as ex:
+                logger.debug(ex)
```
petr-balogh, line 762: What about to add some more log message before like:
logger.warning("Failed to setup the Ceph toolbox pod. Probably because the installation was not successful")

petr-balogh (dismissed):

jilju (dismissed):

ocs_ci/deployment/vmware.py
vavuthu: done

petr-balogh (dismissed):

vavuthu:
verified destroy job here on partially installed cluster: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-destroy-ocs-cluster/14032/console

ocs_ci/deployment/vmware.py
mbukatov: Wait, why do we even care to run these commands via ceph toolbox, when we ignore any errors which may happen during this process?

mbukatov (changes requested):
I'm confused about the goal of this fix, see also the original issue.

vavuthu:
> I'm confused about the goal of this fix, see also the original issue.



> I'm confused about the goal of this fix, see also the original issue.

If the deployment fails for external mode, it won't create tool box/csi users. During destroy job, we need to delete the csi users which was created during deployment. 

There are some cases where deployment fails ( https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/12355/console ) and ceph tool box is not created, Since deployment was failed, We don't want to fail the destroy job ( leaves OCP cluster in Datacenter ) if toolbox doesn't exist

mbukatov (approved):

petr-balogh (approved):

PR #5910, opened 2022-05-17
Including version 4.8 after fix for luks header info issue in 4.8.11

mashetty330:

[BZ](https://bugzilla.redhat.com/show_bug.cgi?id=2069722) was raised for the luks header info issue with OCS 4.8. Now its fixed and verified in 4.8.11. Hence removing the condition to skip 4.8 version under osd encryption verification module.

### DIFF
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -705,7 +705,7 @@ def osd_encryption_verification():
             raise ValueError("OSD is not encrypted")
 
     # skip OCS 4.8 as the fix for luks header info is still not available on it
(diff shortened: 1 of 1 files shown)


petr-balogh (approved):

PrasadDesala (approved):

PrasadDesala:
@mashetty330 Do we need this change to be back-ported to 4.9, 4.8 stable branches as well?

PrasadDesala:
> @mashetty330 Do we need this change to be back-ported to 4.9, 4.8 stable branches as well?

I just checked zstreams runs and they are triggered on ocs-ci/master. So zstream runs will have these changes. 
I think there is no need for backport

PR #5944, opened 2022-05-25
Fix OCP 4.3 installation on vSphere

vavuthu:
hcl2.load is returning list, so we need to loop over list to chnage the
target memory or cpus for modules


vavuthu:
closing this PR as this is not a requirement for now

Issue #6000, opened 2022-06-08
ACM Login is stuck in login_ui() function


jimdevops19:
In the latest ACM deployments, we encounter an issue that we need to confirm login to the ACM before reaching the console login page.
http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/cerginba-acm200/cerginba-acm200_20220516T075223/problemmatic_screenshots/screenshots_ui_1653985137/test_acm_import/

That's why, the expected conditions are failing in the ACM Login function. This one requires to create one more step before the ACM login.

PR #6005, opened 2022-06-09
Handle page appears before the login of UI Console if necessary

jimdevops19:
Fixes: #6000 

Steps added:
 - Calling a new function just after the `driver.get(console_url)`. There seems to be a page that appears just before login that a click on a button is required.
 - New function in `base_ui.py` - Currently added as a separated function since we need to call it before login.
 - Added three new locators to the common `login` dictionary.

### DIFF
--- a/ocs_ci/ocs/ui/base_ui.py
+++ b/ocs_ci/ocs/ui/base_ui.py
@@ -830,6 +830,8 @@ def login_ui(console_url=None):
     wait = WebDriverWait(driver, 60)
     driver.maximize_window()
     driver.get(console_url)
+    # Validate proceeding to the login console before taking any action:
+    proceed_to_login_console(driver)
     if config.ENV_DATA.get("flexy_deployment") or config.ENV_DATA.get(
         "import_clusters_to_acm"
     ):
@@ -877,3 +879,26 @@ def close_browser(driver):
     logger.info("Close browser")
     take_screenshot(driver)
     driver.close()
+
+
+def proceed_to_login_console(driver: WebDriver):
+    """
+    Proceed to the login console, if needed to confirm this action in a page that appears before.
+    This is required to be as a solo function, because the driver initializes in the login_ui function.
+    Function needs to be called just before login
+
+    Args:
+        driver (Selenium WebDriver)
+
+    Returns:
+        None
+
+    """
+    login_loc = locators[get_ocp_version()]["login"]
+    if driver.title == login_loc["pre_login_page_title"]:
+        proceed_btn = driver.find_element(
+            by=login_loc["proceed_to_login_btn"][1],
+            value=login_loc["proceed_to_login_btn"][0],
+        )
+        proceed_btn.click()
+        WebDriverWait(driver, 60).until(ec.title_is(login_loc["login_page_title"]))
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/ui/base_ui.py
```diff
@@ -830,6 +830,8 @@ def login_ui(console_url=None):
     wait = WebDriverWait(driver, 60)
     driver.maximize_window()
     driver.get(console_url)
+    # Validate proceeding to the login console before taking any action:
+    proceed_to_login_console(driver)
```
shylesh, line 834: @jimdevops19 , I think we don't need any change in login_ui, instead:https://github.com/red-hat-storage/ocs-ci/blob/f57e12c4049c409570a9103b8807c82e5630e1bf/ocs_ci/ocs/acm/acm.py#L280 , add here .

ocs_ci/ocs/ui/views.py
```diff
@@ -6,11 +6,14 @@
 ODF_OPERATOR = "OpenShift Data Foundation"
 
 login = {
+    "pre_login_page_title": "Log In",
+    "login_page_title": "Log in · Red Hat OpenShift Container Platform",
```
am-agrawa, line 10: Locator type is missing.

ocs_ci/ocs/ui/base_ui.py
am-agrawa, line 834: 1. This additional login page is applicable only for ACM login, not for OCP login.
2. Is it applicable for all the ACM versions?

am-agrawa:
PR Validation job needs to be run to verify the changes.

ocs_ci/ocs/ui/base_ui.py
jimdevops19: The login function is common across the UI deployment + ACM automation. So that's why I thought it is a good placement here in case we hit this page. @AviadP also mentioned to me that he saw this once in regular UI deployment.

I can only confirm that I faced this in ACM V2.4.4

ocs_ci/ocs/ui/views.py
jimdevops19: Not sure that I understood the problem as I wanted to store a string only and not a tuple of locator that includes By.SOMETHING. 
For example: [this line](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/ui/views.py#L9)
Should I move those to a constants file ?

am-agrawa: Got it, it's okay to keep it here as well. Not sure if we have a better alternative.

ocs_ci/ocs/ui/base_ui.py
am-agrawa: I disagree, I have never seen this additional page while logging into the OCP console.
This needs to be only used for ACM login. Not for OCP. You can validate the same locally or by running a PR validation job. I don't think it will pass.
Though I can confirm it appears for RHACM 2.5 (Downstream).

OdedViner (approved):

am-agrawa (approved):

shyRozen (approved):

AviadP (approved):

PR #6035, opened 2022-06-17
Longevity Testing: Stage3

PrasadDesala:
In Longevity testing we make sure that the ODF can continue responding to the user and admin operations reliably and consistently under
significant load for an extended period of time.

To achieve this the tests are categorized into below stages.

Stage-0: Creates the initial softconfiguration of all supported types
on the Longevity cluster and keep the created resources stay forever on the cluster
Stage-1: Configure and run OCP and Application workloads. These workloads
will be run continuously for a specified period of time
Stage-2: Sequential PVC, OBC, APP pod lifecycle operations for a specified period of time
Stage-3: Concurrent PVC, OBC, APP pod lifecycle operations for a specified period of time
Stage-4: Clone, Snapshot, Expand operations for a specified period of time

This PR automates Longevity Stage-3

### DIFF
--- a/ocs_ci/ocs/longevity.py
+++ b/ocs_ci/ocs/longevity.py
@@ -311,7 +311,7 @@ class Longevity(object):
             list: List of all PVCs in Bound state
 
         Raises:
-        AssertionError: If not all PVCs reached to Bound state
+            AssertionError: If not all PVCs reached to Bound state
 
         """
         return self.validate_pvc_in_kube_job_reached_bound_state(
@@ -732,8 +732,8 @@ class Longevity(object):
             ]
             resource_to_delete_job_file = []
             # waiting for the resource thread to complete
+            log.info("Waiting for the resource thread to complete")
             for resource in resource_to_delete:
-                log.info("Waiting for the resource thread to complete")
                 resource_to_delete_job_file.append(resource.result())
 
             log.info(
@@ -763,8 +763,8 @@ class Longevity(object):
 
             bulk_create_delete_job_file = []
             # waiting for the bulk create delete thread to complete
+            log.info("Waiting for the bulk create delete thread to complete")
             for thread in bulk_create_delete:
-                log.info("Waiting for the bulk create delete thread to complete")
                 bulk_create_delete_job_file.append(thread.result())
             # Delete all the created resources in the bulk create delete thread
             for job_file in bulk_create_delete_job_file:
--- a/ocs_ci/ocs/longevity.py
+++ b/ocs_ci/ocs/longevity.py
@@ -288,6 +288,8 @@ class Longevity(object):
         self.create_stage_builder_kube_job(
             kube_job_obj_list=pvc_job_file_list, namespace=namespace
         )
+        # Wait 60 secs to ensure the PVC on the list has status field populated
+        time.sleep(60)
         # Validate PVCs in kube job reached BOUND state
         self.validate_pvc_in_kube_job_reached_bound_state(
             kube_job_obj_list=pvc_job_file_list,
@@ -511,6 +513,8 @@ class Longevity(object):
         self.create_stage_builder_kube_job(
             kube_job_obj_list=obc_job_file, namespace=namespace
         )
+        # Wait 60 secs to ensure the obc on the list has status field populated
+        time.sleep(60)
         # Validate OBCs in kube job reached BOUND state
         self.validate_obcs_in_kube_job_reached_running_state(
             kube_job_obj=obc_job_file[0],
--- a/ocs_ci/ocs/longevity.py
+++ b/ocs_ci/ocs/longevity.py
@@ -634,7 +634,9 @@ class Longevity(object):
 
         while datetime.now() < end_time:
             cycle_no += 1
-            log.info(f"#################[STARTING CYCLE:{cycle_no}]#################")
+            log.info(
+                f"#################[STARTING STAGE2 CYCLE:{cycle_no}]#################"
+            )
 
             for bulk in (False, True):
                 current_ops = "BULK-OPERATION" if bulk else "SEQUENTIAL-OPERATION"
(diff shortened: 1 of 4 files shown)


ocs_ci/ocs/longevity.py
```diff
@@ -282,13 +283,36 @@ def create_stagebuilder_all_pvc_types(
...
+            pvc_job_file_list (list): List of all PVC.yaml dicts
+            namespace (str): Namespace where the resource has to be created
+            pvc_count (int): Bulk PVC count; If not specified the count will be
+            fetched from the kube job pvc yaml dict
+
+        Returns:
+            list: List of all PVCs in Bound state
+
+        Raises:
+        AssertionError: If not all PVCs reached to Bound state
```
yosibsh, line 308: Add spaces (or tab)

ocs_ci/ocs/longevity.py
```diff
@@ -618,6 +656,120 @@ def stage_2(
...
+                    namespace,
+                    pvc_size,
+                    pvc_kube_job_name="delete_all_pvc_for_pod_attach_job_profile",
+                    pod_kube_job_name="delete_all_pods_job_profile",
+                ),
+            ]
+            resource_to_delete_job_file = []
+            # waiting for the resource thread to complete
+            for resource in resource_to_delete:
+                log.info("Waiting for the resource thread to complete")
```
yosibsh, line 726: Please move this line to be before the for-loop.
It will just print:
Waiting for the resource thread to complete
Waiting for the resource thread to complete
Waiting for the resource thread to complete

ocs_ci/ocs/longevity.py
```diff
@@ -618,6 +656,120 @@ def stage_2(
...
+            ]
+            for job_file in resource_to_delete_job_file:
+                ThreadPoolExecutor(max_workers=1).submit(
+                    self.delete_stage_builder_kube_job, job_file, namespace
+                )
+
+            bulk_create_delete_job_file = []
+            # waiting for the bulk create delete thread to complete
+            for thread in bulk_create_delete:
+                log.info("Waiting for the bulk create delete thread to complete")
```
yosibsh, line 757: Same (before for-loop)

ocs_ci/ocs/longevity.py
PrasadDesala: Done

ocs_ci/ocs/longevity.py
PrasadDesala: Done

ocs_ci/ocs/longevity.py
```diff
@@ -629,6 +666,120 @@ def stage_2(
...
             time.sleep(delay)
 
+    def stage3(
+        self,
+        project_factory,
+        num_of_pvc,
+        num_of_obc,
+        pvc_size=None,
+        delay=60,
+        run_time=1440,
```
Shrivaibavi, line 676: any reason for this exact number ?

ocs_ci/ocs/longevity.py
```diff
@@ -629,6 +666,120 @@ def stage_2(
...
+            num_of_pvc (int): Bulk PVC count
+            num_of_obc (int): Bulk OBC count
+            pvc_size (str): size of all pvcs to be created with Gi suffix (e.g. 10Gi).
+            If None, random size pvc will be created
+            delay (int): Delay in seconds before starting the next cycle
+            run_time (int): The amount of time the particular stage has to run (in minutes)
+
+        """
+        end_time = datetime.now() + timedelta(minutes=run_time)
+        cycle_count = 1
```
Shrivaibavi, line 698: I see the cycle count being incremental in line 778, but Can we get cycle_count as parameter in stage3() ?? Better to pass it via testcase and have pre defined cycles per stage?

ocs_ci/ocs/longevity.py
```diff
@@ -629,6 +666,120 @@ def stage_2(
...
+                    namespace,
+                    obc_kube_job_name="delete_obc_job_profile",
+                ),
+                ThreadPoolExecutor(max_workers=1).submit(
+                    self.create_stagebuilder_pods_with_all_pvc_types,
+                    num_of_pvc,
+                    namespace,
+                    pvc_size,
+                    pvc_kube_job_name="delete_all_pvc_for_pod_attach_job_profile",
+                    pod_kube_job_name="delete_all_pods_job_profile",
```
Shrivaibavi, line 730: First delete pods and then PVCs, Is this order of deletion passed in local run ?

ocs_ci/ocs/longevity.py
PrasadDesala: It has been set to 3hrs for now and once we have a stabilized runs we will change the run time to a longer period. And this is not a strict value and can be changed based on the use case.

ocs_ci/ocs/longevity.py
PrasadDesala: Handling the stage via cycle may not help in our case as we are aiming to run the stage for a longer duration of time. 
If we want to consider it based on the cycles we may have to first get how many such cycles would be required for running the identified duration and then pass the value. This would be a tedious task as there could be a  frequent requirement change.
In the current implementation, the number of cycles per stage depends on the run_time here.

ramkiperiy:
@PrasadDesala Code looks good, I have one question, there are create, delete operations of PVC/POD hope there is validation to check if they are created or deleted.

ocs_ci/ocs/longevity.py
Shrivaibavi: I agree, but i am quite unsure of the deletion process, as i am worried that we should not leave any resources undeleted as we also make use of ignore_leftovers decorator.

PrasadDesala:
> @PrasadDesala Code looks good, I have one question, there are create, delete operations of PVC/POD hope there is validation to check if they are created or deleted.

This is handled [here](https://github.com/red-hat-storage/ocs-ci/pull/6035#issuecomment-1173362142)

ramkiperiy (approved):

Akarsha-rai (approved):

PR #6038, opened 2022-06-19
lvmo shapshot and clone base tests

shyRozen:
Volume snapshot and clone for lvmo base tests

### DIFF
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -28,6 +28,7 @@ from ocs_ci.ocs.exceptions import (
     CommandFailed,
     ResourceNotFoundError,
 )
+from ocs_ci.ocs.cluster import check_clusters
 from ocs_ci.ocs.resources.ocs import get_version_info
 from ocs_ci.ocs.utils import collect_ocs_logs, collect_prometheus_metrics
 from ocs_ci.utility.utils import (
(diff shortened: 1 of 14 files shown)


AviadP (dismissed):

mbukatov (changes requested):
I'm pointing out to the root cause of the tox failure.

Avilir (approved):
Grate work @shyRozen

OdedViner (approved):

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -397,7 +398,9 @@ def pytest_configure(config):
                     del config._metadata[extra_meta]
 
             config._metadata["Test Run Name"] = get_testrun_name()
-            gather_version_info_for_report(config)
+            check_clusters()
+            if ocsci_config.RUN["cephcluster"]:
```
mbukatov, line 402: Why do you assume that this "cephcluster" value will be there? Couldn't it fail on unknown key?

mbukatov: And where is that documented? Please update ocs-ci docs accordingly.

mbukatov (changes requested):

ebenahar (approved):

ocs_ci/framework/pytest_customization/ocscilib.py
shyRozen: Before the condition it set via in line above `check_clusters()` which handles situation where command failed or oc bin is not there. Because it set in run itself I don't think that it need to be documented as it is transparent to the end user.

mbukatov: Cool, I don't want to block it.

mbukatov (approved):

PR #6042, opened 2022-06-20
ODF4.11 UI deployment

OdedViner:
I created some WA because this issue https://github.com/red-hat-storage/odf-console/issues/264

Signed-off-by: odedviner <oviner@redhat.com>

### DIFF
--- a/ocs_ci/ocs/ui/base_ui.py
+++ b/ocs_ci/ocs/ui/base_ui.py
@@ -650,7 +650,7 @@ class PageNavigator(BaseUI):
                 == "false"
             ):
                 logger.info("Show default projects")
-                self.do_click(self.validation_loc["show-default-projects"])
+                self.do_click(self.page_nav["show-default-projects"])
 
         pvc_loc = locators[self.ocp_version]["pvc"]
         logger.info(f"Wait and select namespace {project_name}")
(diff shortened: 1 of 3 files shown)


ocs_ci/ocs/ui/base_ui.py
```diff
@@ -343,7 +343,8 @@ def __init__(self, driver):
         self.ocp_version = get_ocp_version()
         self.ocp_version_full = version.get_semantic_ocp_version_from_config()
         self.page_nav = locators[self.ocp_version]["page"]
-        self.validation_loc = locators[self.ocp_version]["validation"]
+        if self.ocp_version_full != version.VERSION_4_11:
```
dahorak, line 346: Do we know, how it will be in the following OCP releases (4.12,...)? Or is this and the other changes tightly related to OCP 4.11 only and no lower or upper version?
Because if there is a chance, that it will apply also to the future releases, I would suggest to use `<` operator or similar (in the other conditions in this PR.

am-agrawa: +1

ebenahar (approved):

ocs_ci/ocs/ui/base_ui.py
OdedViner: I dont think we need to use "validation" dict in `page_nav` or `deployment_ui` classes.
So I deleted it

dahorak (approved):

PrasadDesala (approved):

Issue #6055, opened 2022-06-22
Engineering upgrade pipeline failing with semantic version check


petr-balogh:
https://jenkins.ceph.redhat.com/job/ocs-ci/1127/consoleFull

```
ocs_ci/helpers/sanity_helpers.py:48: in health_check
    self.ceph_cluster.cluster_health_check(timeout=60)
ocs_ci/ocs/cluster.py:318: in cluster_health_check
    and version.get_semantic_version(
ocs_ci/utility/version.py:27: in get_semantic_version
    version = Version.coerce(version)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'semantic_version.base.Version'>, version_string = None
partial = False

    @classmethod
    def coerce(cls, version_string, partial=False):
        """Coerce an arbitrary version string into a semver-compatible one.
    
        The rule is:
        - If not enough components, fill minor/patch with zeroes; unless
          partial=True
        - If more than 3 dot-separated components, extra components are "build"
          data. If some "build" data already appeared, append it to the
          extra components
    
        Examples:
            >>> Version.coerce('0.1')
            Version(0, 1, 0)
            >>> Version.coerce('0.1.2.3')
            Version(0, 1, 2, (), ('3',))
            >>> Version.coerce('0.1.2.3+4')
            Version(0, 1, 2, (), ('3', '4'))
            >>> Version.coerce('0.1+2-3+4_5')
            Version(0, 1, 0, (), ('2-3', '4-5'))
        """
        base_re = re.compile(r'^\d+(?:\.\d+(?:\.\d+)?)?')
    
>       match = base_re.match(version_string)
E       TypeError: expected string or bytes-like object

venv/lib/python3.8/site-packages/semantic_version/base.py:234: TypeError
```

PR #6056, opened 2022-06-22
Fixes health check with no provided ugprade version

petr-balogh:
Fixes: #6055


vavuthu (approved):

ebenahar (approved):

b-ranto:
This helped us progress with the upgrade DS, we still hit a couple of mcg errors in the acceptance suite though.

petr-balogh:
@b-ranto this is all caused by this bug:  https://bugzilla.redhat.com/show_bug.cgi?id=2075422

We had to many workarounds to unblock the upgrade (and deployment) and ignore noobaa checks. But then whole noobaa is affected by this because of rejected phase. So till bug is fixed we cannot do much when we have starting point OCP 4.10 when we install ODF 4.9 and upgrading to 4.10.

PR #6058, opened 2022-06-24
PR to collect container rpm package details during ocs collect logs phase

pintojoy:
This PR intends to collect conatiner rpm package information during ocs collect logs phase which can be used to verify CVE bugs.

Signed-off-by: pintojoy <jopinto@redhat.com>

### DIFF
--- a/ocs_ci/ocs/utils.py
+++ b/ocs_ci/ocs/utils.py
@@ -1381,26 +1381,3 @@ def collect_pod_container_rpm_package(dir_name):
                 log_file_name = f"{package_log_dir_path}/{container_name}-rpm.log"
                 with open(log_file_name, "w") as f:
                     f.write(container_output)
-
-    def label_pod_security_admission(namespace=None, upgrade_version=None):
-        """
-        Label PodSecurity admission
-
-        Args:
-            namespace (str): Namespace name
-            upgrade_version (semantic_version.Version): ODF semantic version for upgrade
-                if it's an upgrade run, otherwise None.
-        """
-        namespace = namespace or constants.OPENSHIFT_STORAGE_NAMESPACE
-        log.info(f"Labelling namespace {namespace} for PodSecurity admission")
-        if version.get_semantic_ocp_running_version() >= version.VERSION_4_12 or (
-            upgrade_version and upgrade_version >= version.VERSION_4_12
-        ):
-            ocp_obj = OCP(kind="namespace")
-            label = (
-                "security.openshift.io/scc.podSecurityLabelSync=false "
-                f"pod-security.kubernetes.io/enforce={constants.PSA_PRIVILEGED} "
-                f"pod-security.kubernetes.io/warn={constants.PSA_BASELINE} "
-                f"pod-security.kubernetes.io/audit={constants.PSA_BASELINE} --overwrite"
-            )
-            ocp_obj.add_label(resource_name=namespace, label=label)
--- a/ocs_ci/ocs/utils.py
+++ b/ocs_ci/ocs/utils.py
@@ -1336,3 +1336,45 @@ def label_pod_security_admission(namespace=None, upgrade_version=None):
             f"pod-security.kubernetes.io/audit={constants.PSA_BASELINE} --overwrite"
         )
         ocp_obj.add_label(resource_name=namespace, label=label)
+
+
+def collect_pod_container_rpm_package(dir_name):
+    """
+    Collect information about rpm packages from all containers from
+    Args:
+        dir_name(str): directory to store container rpm package info
+    """
+    from ocs_ci.ocs.resources import pod
+
+    timestamp = time.time()
+    cluster_namespace = ocsci_config.ENV_DATA["cluster_namespace"]
+
+    log_dir_path = os.path.join(
+        os.path.expanduser(ocsci_config.RUN["log_dir"]),
+        f"{dir_name}_{ocsci_config.RUN['run_id']}",
+    )
+    package_log_dir_path = os.path.join(
+        log_dir_path, "rpm_package", f"rpm_list_{timestamp}"
+    )
+    create_directory_path(package_log_dir_path)
+    log.info(f"Directory path for rpm logs is{package_log_dir_path}")
+    pods = pod.get_all_pods(namespace=cluster_namespace)
+    for pod_name in pods:
+        pod_object = pod_name.get()
+        pod_containers = pod_object.get("spec").get("containers")
+        for container in pod_containers:
+            container_name = container["name"]
+            ocp_pod_obj = OCP(kind=constants.POD, namespace=cluster_namespace)
+            pod_status = ocp_pod_obj.get_resource_status(pod_name.name)
+            if pod_status == constants.STATUS_RUNNING:
+                command = f"exec -i {pod_name.name} -c {container_name} -- rpm -qa"
+                ocp_obj = OCP(namespace=cluster_namespace)
+                try:
+                    container_output = ocp_obj.exec_oc_cmd(command)
+                except:
+                    log.warning(
+                        f"failed to get rpm packages on pod {pod_name.name} container {container_name}"
+                    )
+                log_file_name = f"{package_log_dir_path}/{container_name}-rpm.log"
+                with open(log_file_name, "w") as f:
+                    f.write(container_output)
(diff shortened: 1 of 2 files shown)


pintojoy:
Going through PR validation log to see if deployment and log collection stage is successful, will remove "DO NOT MERGE" label if everything looks fine

pintojoy:
Latest PR validation job has passed with container rpm packages collected, Please refer https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/14157/

ocs_ci/ocs/utils.py
```diff
@@ -1266,3 +1271,39 @@ def get_primary_cluster_config():
...
+
+    Args:
+        dir_name(str): directory to store container rpm package info
+
+    """
+    from ocs_ci.ocs.resources import pod
+
+    create_directory_path(dir_name)
+    log.info(f"Directory path for rpm logs is{dir_name}")
+    pods = pod.get_all_pods(namespace=defaults.ROOK_CLUSTER_NAMESPACE)
```
petr-balogh, line 1288: Please rather then defaults.ROOK_CLUSTER_NAMESPACE use:
namespace = ocsci_config.ENV_DATA["cluster_namespace"]

And use this namespace in every place where you are bellow using it from defaults.

ocs_ci/ocs/utils.py
```diff
@@ -1023,6 +1023,11 @@ def collect_ocs_logs(dir_name, ocp=True, ocs=True, mcg=False, status_failure=Tru
             raise ValueError(
                 f"must-gather fails in an disconnected environment bz-1974959\n{mg_output}"
             )
+        package_log_dir_path = os.path.join(log_dir_path, "rpm_package")
+        try:
+            collect_pod_container_rpm_package(package_log_dir_path)
```
petr-balogh, line 1028: This will collect those RPMs for every must gather execution. Even it will try when we have OCP deployment and fail, where OCS deployment is not finished yet.

Also this will collect for every test failure - do you know how long does this takes?
And do we really need to run it for each test failure again and again?

My guess is that we need to collect it only once per whole run? Right?
For production runs we are using collect logs on success parameter and this case I think will be better to do it only once at the end of the execution.

So what I suggest is to remove the call from here, And move it to here:
https://github.com/red-hat-storage/ocs-ci/blob/6c232602d0bf5745f67c28dbcaaadba27c535eb6/tests/conftest.py#L3862 .

WDYT?

This means that it will be called only once at the end of the production run.

ocs_ci/ocs/utils.py
```diff
@@ -1023,6 +1023,11 @@ def collect_ocs_logs(dir_name, ocp=True, ocs=True, mcg=False, status_failure=Tru
             raise ValueError(
                 f"must-gather fails in an disconnected environment bz-1974959\n{mg_output}"
             )
+        package_log_dir_path = os.path.join(log_dir_path, "rpm_package")
+        try:
```
petr-balogh, line 1027: I would move the try block inside the function itself to wrap exact command on this line:

https://github.com/red-hat-storage/ocs-ci/pull/6058/files#diff-13cfcd3566979733be4860f4cc0db5af5382acbf7cf54fd94eb36fab45474c21R1301

container_output = ocp_obj.exec_oc_cmd(command)

As I can imagine the scenario that some pod you got info about at the start of the function will get deleted during the run. And there will not be such container anymore so you will loose information from everything.
But if it will just write warning inside the loop, it will try to continue to collect RPMs info from the rest of the pods/containers.

ocs_ci/ocs/utils.py
```diff
@@ -1266,3 +1271,39 @@ def get_primary_cluster_config():
...
+                ocp_obj = OCP(namespace=defaults.ROOK_CLUSTER_NAMESPACE)
+                container_output = ocp_obj.exec_oc_cmd(command)
+                log_file_name = f"{dir_name}/{container_name}-rpm.log"
+                try:
+                    with open(log_file_name, "w") as f:
+                        f.write(container_output)
+                        f.close()
+                except Exception:
+                    log.warning(f"Cannot write the log to the file {log_file_name}")
+    log.info("Retrived rpm package list from all containers successfully")
```
petr-balogh, line 1309: If you will move the try logic inside to this function you cannot say that here as some could fail.

Even now if you collected the logs and didn't write to the file it will write that it successfully retrieved list but didn't write it as it failed and you just logged warning.

I guess this is not typical that you should handle such exception here, or you are failing to write a log file from some rason?

If so, this should be probably covered in general wrap of try except block on the place where you will call that to not break the session scope fixture.

petr-balogh (dismissed):

ocs_ci/ocs/utils.py
pintojoy: Updated with changes

ocs_ci/ocs/utils.py
pintojoy: Updated with suggested changes

ocs_ci/ocs/utils.py
pintojoy: Updated with suggested changes

ocs_ci/ocs/utils.py
```diff
@@ -1284,3 +1284,45 @@ def thread_init_class(class_init_operations, shutdown):
...
             return
+
+
+def collect_pod_container_rpm_package(dir_name):
+    """
+    Collect information about rpm packages from all containers from
+    Args:
+        dir_name(str): directory to store container rpm package info
+    """
+    from ocs_ci.ocs.resources import pod
```
keesturam, line 1295: can you please move this import outside the function at the beginning of the file?

tests/conftest.py
```diff
@@ -3903,6 +3907,7 @@ def finalizer():
             if config.REPORTING["collect_logs_on_success_run"]:
                 collect_ocs_logs("testcases", ocs=False, status_failure=False)
                 collect_ocs_logs("testcases", ocp=False, status_failure=False)
+                collect_pod_container_rpm_package("testcases")
```
keesturam, line 3910: Would this run for all test cases? 
Maybe we could add an option to run it only for deployment tests?

ocs_ci/ocs/utils.py
pintojoy: Since utils.py is imported in pod.py, importing at the beginning causes deadlock. Hence i added import statement within specific function

vavuthu: > 
please add the comment for circular dependancy

ocs_ci/ocs/utils.py
```diff
@@ -1284,3 +1284,45 @@ def thread_init_class(class_init_operations, shutdown):
         if shutdown == 0:
             executor.shutdown(wait=True)
             return
+
+
+def collect_pod_container_rpm_package(dir_name):
+    """
+    Collect information about rpm packages from all containers from
```
vavuthu, line 1291: blank line below 1291 line

vavuthu, line 1291:
```suggestion
    Collect information about rpm packages from all containers
```

tests/conftest.py
pintojoy: Made the changes

ocs_ci/ocs/utils.py
```diff
@@ -1284,3 +1284,48 @@ def thread_init_class(class_init_operations, shutdown):
...
+    log.info(f"Directory path for rpm logs is{package_log_dir_path}")
+    pods = pod.get_all_pods(namespace=cluster_namespace)
+    for pod_name in pods:
+        pod_object = pod_name.get()
+        pod_containers = pod_object.get("spec").get("containers")
+        for container in pod_containers:
+            container_name = container["name"]
+            ocp_pod_obj = OCP(kind=constants.POD, namespace=cluster_namespace)
+            pod_status = ocp_pod_obj.get_resource_status(pod_name.name)
+            if pod_status == constants.STATUS_RUNNING:
```
vavuthu, line 1320: we need to check these before the for loop, there is no need to check for every container

ocs_ci/ocs/utils.py
```diff
@@ -1284,3 +1284,48 @@ def thread_init_class(class_init_operations, shutdown):
...
+                command = f"exec -i {pod_name.name} -c {container_name} -- rpm -qa"
+                ocp_obj = OCP(namespace=cluster_namespace)
```
vavuthu, line 1322: you can initiate this out side for loop

Avilir:
Please rebase to resolve conflicts.

petr-balogh (approved):

petr-balogh:
BTW, triggered verification job here for you: 
https://url.corp.redhat.com/556e4e2

petr-balogh:
Can mark as verified:  https://url.corp.redhat.com/8d4b792

petr-balogh (approved):

dahorak (approved):

Issue #5789, opened 2022-04-26
get_ocs_csv loops for 7 hours waiting for nonexistent OCS version


clacroix12:
I have encountered this issue while testing ACM Hub Downstream deployments where there is no OCS being deployed. We have a check on some of our tests to determine whether they will be skipped or not, one of which is [skipif_upgraded_from](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L3396-L3423). This check is calling `get_ocs_csv` which is failing due to the fact that OCS is not installed on this cluster, which then proceeds to loop for seven hours before the execution finally fails.

Not only should this check be taking into consideration the possibility that OCS doesn't exist in the test environment, but we should be moving on much sooner. In addition, these checks shouldn't even be performed unless they are on tests that are being specified to actually run. For this example I'm trying to run a `deployment` test, yet we have dozens of checks being performed to determine whether or not some tests can be skipped, even though they won't be executed anyways. This is a big area where we can make performance improvements as these checks take a non-trivial amount of time to execute. I have opened an [issue](https://github.com/red-hat-storage/ocs-ci/issues/5788) to track this as well.

You can see this behavior present in this execution: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/12044/consoleFull

petr-balogh:
@clacroix12 what is happening is that we have this pytest hook:
`pytest_collection_modifyitems`
https://github.com/red-hat-storage/ocs-ci/blob/master/tests/conftest.py#L188
We do not perform those check in severla cases as you can see for deployment, if OCS deployment is skipped and so on.

Where we actually are trying to delete some of the tests from collection so they will not be in the report at all, so we will not see them as skipped for example but they will not be even collected by pytest.

And we have several checks which we perform to decide when such test should be deleted. One of the check you mentioned is skipif_upgraded_from which actually in the hook delete those items from collection - not just skip. 

In your job, you introduced scenario which is currently not valid for OCS-CI as the test execution requires the OCS to be present, so you cannot at the moment run the test execution against the cluster which has only OCP and ACM installed.

[...]

PR #6071, opened 2022-06-27
Add role binding for BZ2101343

shyRozen:
Signed-off-by: Shay Rozen <shay.rozen@gmail.com>

PR #6082, opened 2022-06-28
Solve waiting for csv

shyRozen:
This issue [get_ocs_csv loops for 7 hours waiting for nonexistent OCS version](https://github.com/red-hat-storage/ocs-ci/issues/5789) is causing 7H delay in PR validation when odf-operator is not istalled. I cases like ACM, SNO, OCP (skip_ocs_deployment) between the run-ci which deploy and the run-ci which do teardown there is this:
`run-ci --color=yes tests/ -m deployment -k '' --ocs-version 4.11 --ocp-version 4.11 --ocsci-conf=/tmp/DCRKJQVFDIYB '--ocsci-conf=~/current-cluster-dir/ocsci_conf.yaml' --ocsci-conf=conf/ocsci/production-aws-ipi.yaml --ocsci-conf=conf/deployment/vsphere/upi_1az_rhcos_vsan_1mw.yaml --ocsci-conf=ocs_ci/templates/lvmo-deployment/lvmo_deployment_5_disk_X_150GB.yaml --cluster-name jnk-pr5715-b2963 --cluster-path /home/jenkins/current-cluster-dir/openshift-cluster-dir --junit-xml /home/jenkins/current-cluster-dir/logs/test_results_1655682703.xml -o junit_suite_name=testsuite '--html=~/current-cluster-dir/logs/test_report_1655682703.html' --self-contained-html --bugzilla --email=srozen@redhat.com --squad-analysis`
Is is iterating over testcases to check marks and they are some marks that call csv.
The problem is here:
`ocs_csv.wait_for_phase(phase="Succeeded", timeout=600)`

The function has also retry:
`    @retry(ResourceWrongStatusException, tries=4, delay=5, backoff=1)
    def wait_for_phase(self, phase, timeout=300, sleep=5):
`
So multiple the testcases*timeout*retry you'll get 7H
The solution is that everytime `def get_ocs_csv():` is called there is a check if cephcluster object can be initiated and if not it returns CSVNotFound which `def get_ocs_csv():` is declaring to raise.
It takes 2 minutes instead of 7H. There are some errors but maybe they are related to my local run-ci which tries to imitate what is going in the run-ci in jenkins run.


shyRozen:
Deployment failed because this is not in stable yet https://github.com/red-hat-storage/ocs-ci/pull/6079 and also this https://github.com/red-hat-storage/ocs-ci/pull/6071. Will run on 4.11-98 which was stable for LVMO

ocs_ci/ocs/resources/ocs.py
```diff
@@ -212,6 +212,12 @@ def get_ocs_csv():
         CSVNotFound: In case no CSV found.
 
     """
+
+    test = OCP(kind="operator", resource_name="odf-operator.openshift-storage")
+    data = test.__getattribute__("data")
```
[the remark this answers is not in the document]
shyRozen: Yes

ocs_ci/ocs/resources/ocs.py
```diff
@@ -212,6 +212,12 @@ def get_ocs_csv():
         CSVNotFound: In case no CSV found.
 
     """
+
+    odf_csv = OCP(kind="operator", resource_name="odf-operator.openshift-storage")
```
petr-balogh, line 216: I think we should make this name of operator based on the version.
So if version < 4.9 use 
OCS_OPERATOR_NAME from here
https://github.com/red-hat-storage/ocs-ci/blob/baeb91c20ca96fcb5f3d97350183b08ebec481a5/ocs_ci/ocs/defaults.py#L38-L39 .

Otherwise:
ODF_OPERATOR_NAME .

For version, please use this function

 https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/version.py#L49 

Comparing with e.g. version.VERSION_4_9

petr-balogh, line 216:
```suggestion
    odf_csv = OCP(kind="operator", resource_name="odf-operator", namespace=namespace
```

The namespace definition is on the line 221 so you will need to move it above this.

shyRozen: This is not namespace but part of the operator name. The operator can be get from any namespace. It doesn't have in metadata the namespace mentioned. The operator is creating the namespace when installing it and therefore it is not in the openshift .

shyRozen: You are relaying on ocs/odf version installed for the check which can be that ocs/odf is not installed at all (The case of the PR).
What we can do is check for both of the names and if both check fails then return csv not found. WDYT?

shyRozen: metadata:
  creationTimestamp: "2022-06-21T19:06:33Z"
  generation: 1
  name: odf-operator.openshift-storage

vavuthu (approved):

Issue #6066, opened 2022-06-26
When verifying the number of nodes created in MS deployment - there are only 6 nodes instead of 9


yitzhak12:
When verifying the number of nodes created in MS deployment - there are only 6 nodes instead of 9. 
For example this deployment of the consumer cluster: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/13927/console: 

> 2022-06-26 14:30:29  11:30:29 - MainThread - ocs_ci.utility.utils - INFO  - Executing command: oc get clusterversion version -o jsonpath='{.spec.clusterID}'
2022-06-26 14:30:30  11:30:29 - MainThread - ocs_ci.deployment.deployment - INFO  - clusterID (UUID): 66863a9d-e533-464f-a98d-3ff16a72c473
2022-06-26 14:30:30  11:30:29 - MainThread - ocs_ci.utility.utils - INFO  - Executing command: oc get Node  -o yaml
2022-06-26 14:30:30  11:30:30 - MainThread - ocs_ci.deployment.deployment - ERROR  - Expected number of nodes is 9 but created during deployment is 6
2022-06-26 14:30:30  11:30:30 - MainThread - ocs_ci.ocs.utils - INFO  - Must gather image: quay.io/openshift/origin-must-gather will be used.

The error happens in the deployment when checking the number of nodes in the function `verify_all_nodes_created`: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/node.py#L1614. 
After approximately 45 min when I checked the node in the CLI, I saw that all the nodes were up and ready: 

> (ocs-ci-venv) oc get nodes
NAME                           STATUS   ROLES          AGE   VERSION
ip-10-0-138-45.ec2.internal    Ready    master         78m   v1.23.5+3afdacb
ip-10-0-139-116.ec2.internal   Ready    worker         70m   v1.23.5+3afdacb
ip-10-0-141-60.ec2.internal    Ready    infra,worker   59m   v1.23.5+3afdacb
ip-10-0-144-113.ec2.internal   Ready    infra,worker   60m   v1.23.5+3afdacb
ip-10-0-149-80.ec2.internal    Ready    worker         70m   v1.23.5+3afdacb
ip-10-0-150-54.ec2.internal    Ready    master         79m   v1.23.5+3afdacb
[...]

PR #6095, opened 2022-06-30
ignoring node count missmatch error in deployment test for MS

suchita-g:
This PR is to ignore known is\sue and
reduce the fleakyness in the deployment pipeline
https://github.com/red-hat-storage/ocs-ci/issues/6066

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -1612,9 +1612,16 @@ def verify_all_nodes_created():
 
     existing_num_nodes = len(get_all_nodes())
     if expected_num_nodes != existing_num_nodes:
-        raise NotAllNodesCreated(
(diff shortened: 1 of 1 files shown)


suchita-g:
verified: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/413/

ebenahar (approved):

jilju (approved):

vavuthu:
the main intention of this check is to stop deployment of ODF and raise excepton if we don't see expected nodes, because ODF installation obviously fails if the expected nodes are not created.  

I am not sure how MS behaves if expected nodes are not created ( let say if worker nodes are not created, then MS installation succeeds ? ). If MS installation fails, I feel its better to raise exception instead of failing MS cluster deployment and then debugging the job with question why deployments fails.

vavuthu (approved):

PR #6101, opened 2022-06-30
Add marker in the test case test_rook_operator_restart_during_mon_failover to run on MS

jilju:
Add marker 'runs_on_provider' in the test case 'test_rook_operator_restart_during_mon_failover' to run on Managed Services provider cluster. If the platform is MS in a multicluster run, the cluster context will be switched to provider cluster before the start of the test case. The cluster context will be changed back to the initial cluster context after the test case.

### DIFF
--- a/tests/manage/z_cluster/test_rook_operator_restart_during_mon_failover.py
+++ b/tests/manage/z_cluster/test_rook_operator_restart_during_mon_failover.py
@@ -16,6 +16,7 @@ from ocs_ci.framework.testlib import (
     skipif_ocs_version,
     skipif_external_mode,
     ignore_leftovers,
+    runs_on_provider,
 )
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


jilju:
Verified http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/jijoy-bug-c1/jijoy-bug-c1_20220630T114214/multicluster/logs/test_report_1656662247.html

vavuthu (approved):

yitzhak12:
I have just a general comment. I am not quite sure if switching to the provider is what we are looking for in the run. 
If we want to test the functionality on the consumer, all the tests should run on the consumer, and this one should be skipped(means to add the `skipif_ms_consumer` marker). And if we run all the tests on the provider, we should ignore other tests. 
I think it will be easier for us to track the consumer and provider failures this way cause some tests are applicable for both consumer and provider.

PR #6104, opened 2022-07-01
[WIP] Modify the test case test_rolling_shutdown_and_recovery to run on MS

jilju:
Modified the test case given below to run on Managed services platform.

tests/manage/z_cluster/nodes/test_rolling_shutdown_and_recovery.py::TestRollingWorkerNodeShutdownAndRecovery::test_rolling_shutdown_and_recovery

Node operations should be done on provider. Added the marker "runs_on_provider" to switch the cluster context to provider before the start of the test case and switch back to  the initial cluster context after completing the test case.
Switching to consumer cluster context is done before creating PVC, pods etc.
The finalizer which is intended to recover the nodes should run on provider cluster.

### DIFF
--- a/tests/manage/z_cluster/nodes/test_rolling_shutdown_and_recovery.py
+++ b/tests/manage/z_cluster/nodes/test_rolling_shutdown_and_recovery.py
@@ -10,10 +10,13 @@ from ocs_ci.framework.testlib import (
     bugzilla,
     skipif_external_mode,
     skipif_ibm_cloud,
+    runs_on_provider,
 )
+from ocs_ci.ocs import constants
 from ocs_ci.ocs.node import get_ocs_nodes
 from ocs_ci.ocs.resources.pod import wait_for_pods_to_be_running
 from ocs_ci.helpers.sanity_helpers import Sanity
+from ocs_ci.framework import config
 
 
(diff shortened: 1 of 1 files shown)


jilju:
Verified : http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/jijoy-bug-c1/jijoy-bug-c1_20220630T114214/multicluster/logs/test_report_1656667689.html

jilju:
ROSA doesn't yet support shutdown of full ROSA cluster or shutdown of a single node. Decision pending on skipping this test.

ebenahar (approved):

yitzhak12:
Let us skip the test and create a similar test for the Managed Service, which will be as follow:
Shutdown all nodes one after the other, but in a controlled fashion(ceph health OK before shutting down next node).
Thread link: https://chat.google.com/room/AAAABfd8ivI/oTXGyoT8VrU

jilju:
> Let us skip the test and create a similar test for the Managed Service, which will be as follow: Shutdown all nodes one after the other, but in a controlled fashion(ceph health OK before shutting down next node). Thread link: https://chat.google.com/room/AAAABfd8ivI/oTXGyoT8VrU


Yes, we can close this PR.

PR #6116, opened 2022-07-03
Test nodes restart scenarios using the Managed Service

yitzhak12:
Create a new file to test the node restart scenarios using the Managed Service.

### DIFF
--- a/tests/manage/z_cluster/nodes/test_nodes_restart_ms.py
+++ b/tests/manage/z_cluster/nodes/test_nodes_restart_ms.py
@@ -10,7 +10,6 @@ from ocs_ci.framework.testlib import (
     ManageTest,
     bugzilla,
     managed_service_required,
-    runs_on_provider,
 )
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.node import (
@@ -28,7 +27,6 @@ from ocs_ci.ocs.resources import pod
 from ocs_ci.helpers.sanity_helpers import Sanity
 from ocs_ci.ocs.cluster import (
     ceph_health_check,
-    is_ms_provider_cluster,
     is_ms_consumer_cluster,
 )
 from ocs_ci.framework import config
@@ -65,9 +63,7 @@ class TestNodesRestartMS(ManageTest):
         else:
             consumer_indexes = config.get_consumer_indexes_list()
 
-        self.create_pods_and_pvcs_factory(
-            consumer_indexes=consumer_indexes, scale_count=20
-        )
+        self.create_pods_and_pvcs_factory(consumer_indexes=consumer_indexes)
 
     @pytest.fixture(autouse=True)
     def teardown(self, request, nodes):
@@ -81,25 +77,23 @@ class TestNodesRestartMS(ManageTest):
             for n in ocp_nodes:
                 recover_node_to_ready_state(n)
 
-            if is_ms_provider_cluster():
-                logger.info(
-                    "Verify the worker nodes security groups on the provider..."
-                )
-                assert verify_worker_nodes_security_groups()
-
             ceph_health_check()
 
         request.addfinalizer(finalizer)
 
     @tier4a
-    @runs_on_provider
     def test_osd_node_restart_and_check_osd_pods_status(self, nodes):
         """
         1) Restart one of the osd nodes.
         2) Check that the osd pods associated with the node should change to a Terminating state.
         3) Wait for the node to reach Ready state.
-        3) Check that the new osd pods with the same ids start on the same node.
+        4) Check that the new osd pods with the same ids start on the same node.
+        5) Check the worker nodes security groups
         """
+        # This is a workaround due to the issue https://github.com/red-hat-storage/ocs-ci/issues/6162
+        if is_ms_consumer_cluster():
+            pytest.skip("The test will not run on a consumer cluster")
+
         self.create_resources()
 
         osd_node_name = random.choice(get_osd_running_nodes())
@@ -126,6 +120,9 @@ class TestNodesRestartMS(ManageTest):
         logger.info(
             f"the osd ids {old_osd_pod_ids} Successfully come up on the node {osd_node_name}"
         )
+
+        logger.info("Verify the worker nodes security groups on the provider...")
+        assert verify_worker_nodes_security_groups()
 
     @tier4a
     @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


tests/manage/z_cluster/nodes/test_nodes_restart_ms.py
```diff
@@ -0,0 +1,136 @@
...
+        )
+        logger.info(
+            f"the osd ids {old_osd_pod_ids} Successfully come up on the node {osd_node_name}"
+        )
+
+    @tier4a
+    @pytest.mark.parametrize(
+        argnames=["force"],
+        argvalues=[
+            pytest.param(*[True], marks=pytest.mark.polarion_id("OCS-894")),
```
ebenahar, line 109: are these the correct polarion IDs for these test cases? I think we need new Polarion test cases for the MS scenarios

yitzhak12:
The pr validation job passed here: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/517/console.

tests/manage/z_cluster/nodes/test_nodes_restart_ms.py
```diff
@@ -0,0 +1,180 @@
...
+            node_names = get_worker_nodes()
+        else:
+            node_names = get_master_nodes()
+
+        ocp_nodes = get_node_objs(node_names=node_names)
+        for node in ocp_nodes:
+            nodes.restart_nodes(nodes=[node], wait=False)
+            self.sanity_helpers.health_check(cluster_check=False, tries=60)
+
+        self.create_resources()
```
ebenahar, line 180: the implementation seems quite similar to the existing test case of non-MS. Please move similar implementation to a common function in order to avoid code duplication

yitzhak12: Yes, but the created resources should be different when we use the MS. Also, it is slightly different because I roll restart the worker nodes the first time and the master nodes the second time when using MS. 
Anyway, I will change the first lines to be more clear.

tests/manage/z_cluster/nodes/test_nodes_restart_ms.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        1) Restart one of the osd nodes.
+        2) Check that the osd pods associated with the node should change to a Terminating state.
+        3) Wait for the node to reach Ready state.
+        4) Check that the new osd pods with the same ids start on the same node.
+        5) Check the worker nodes security groups
+        """
+        # This is a workaround due to the issue https://github.com/red-hat-storage/ocs-ci/issues/6162
+        if is_ms_consumer_cluster():
+            pytest.skip("The test will not run on a consumer cluster")
```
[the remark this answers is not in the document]
ebenahar: just so we will not skip the test too often

tests/manage/z_cluster/nodes/test_nodes_restart_ms.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def create_resources(self):
+        """
+        Create resources on the consumers and run IO
+
+        """
+        if is_ms_consumer_cluster():
+            consumer_indexes = [config.cur_index]
+        else:
+            consumer_indexes = config.get_consumer_indexes_list()
```
dahorak, line 63: I'm not sure, if I didn't miss anything here, but if I'll have following configuration: 1 provider cluster `pr` and 1 consumer clusters `c1` and `c1` and if I'll run this test with clusters specified in following order: `pr`, `c1`, `c2` (first one is the default/current cluster - provider) - then the `consumer_indexes` will contain index for both consumers (`[1, 2]`), but if I'll run the tests with clusters specified in order `c1`, `c2`, `pr` (default/current cluster will be first consumer), then `consumer_indexes` will contain only list with one item - the default cluster - first consumer. Is this expected behavior?

yitzhak12: Yes. The test applies to a consumer and provider. So if I run on a consumer, I will create the resources only on the current consumer, and if I run the test on a provider, I will create the resources on all the consumers.

tests/manage/z_cluster/nodes/test_nodes_restart_ms.py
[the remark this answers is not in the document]
jilju: Many tier4 test cases will require both the provider and consumer. So we run test cases using musticluster job with access to provider and consumer. I would suggest switching to provider cluster when needed.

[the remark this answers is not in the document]
ebenahar: still not convinced why we can't switch to the provider here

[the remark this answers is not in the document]
yitzhak12: Okay, I fixed it.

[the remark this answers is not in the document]
yitzhak12: Fixed in the commit: https://github.com/red-hat-storage/ocs-ci/pull/6116/commits/ab3427f6c58dc806a03b2aa5d30c146134f18f27

ebenahar (approved):

ebenahar (approved):

tests/manage/z_cluster/nodes/test_nodes_restart_ms.py
dahorak: I'm not sure, if it is the best approach to change the behavior based on the order of clusters configuration provided during execution. The question is, if it is sufficient to run the tests with one clusters configuration order?

yitzhak12:
pr validation job passed: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/595/console.

keesturam (approved):
LGTM. What is your next plan on Daniel's comment?

tests/manage/z_cluster/nodes/test_nodes_restart_ms.py
yitzhak12: Maybe I didn't get you. The test is applicable for both the MS consumer cluster and MS provider cluster. - this means there are two options here(like we can test with vSphere UPI and vSphere LSO).

So suppose we want to test the MS consumer cluster, we will create the resources only on the current consumer cluster(because the other consumers are not relevant).
And if we run the test on a provider, we will create the resources on all the consumer clusters.

Let me know if it makes sense.

dahorak: Thanks for the explanation.

dahorak (approved):

PR #6159, opened 2022-07-09
Add threads to finalizers that have list of objects.

shyRozen:
Today the finalizer are acting in serial if there are instances of the same object kind. The teardown takes as long as the test, and even more. This PR is intent to make a finalizer that has a list of objects of the same kind to delete them in parallel.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,7 +6,7 @@ import random
 import time
 import tempfile
 import threading
-from concurrent.futures.thread import ThreadPoolExecutor
+from concurrent.futures import as_completed, ThreadPoolExecutor
 from datetime import datetime
 from math import floor
 from shutil import copyfile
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -5682,3 +5719,20 @@ def teardown():
     workload.deploy_workload()
 
     return workload
+
+
+def thread_finalizer(operation, instances):
```
ebenahar, line 5724: not all places that call this function use it as a finalizer. 
I can see that for instance, [here](https://github.com/red-hat-storage/ocs-ci/pull/6159/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R845), you are calling it from a non teardown fixture function. 
Maybe woth representing that fact by changing the function name from  `thread_finalizer` to `thread_operation`

shyRozen: Sounds good

shyRozen: Changed

shyRozen:
To ambitious. Lowering the resource in #6250

Issue #6209, opened 2022-07-20
Encrypted cluster fails to upgrade because of problem when creating debug pod


petr-balogh:
After the upgrade we are doing the verification via debug pod.
This is failing with the error:
>               raise CommandFailed(
                    f"Error during execution of command: {masked_cmd}."
                    f"\nError is {masked_stderr}"
                )
E               ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc debug node/ip-10-0-170-186.us-east-2.compute.internal -- chroot /host lsblk.
E               Error is Warning: would violate PodSecurity "restricted:latest": host namespaces (hostNetwork=true, hostPID=true), privileged (container "container-00" must not set securityContext.privileged=true), allowPrivilegeEscalation != false (container "container-00" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "container-00" must set securityContext.capabilities.drop=["ALL"]), restricted volume types (volume "host" uses restricted volume type "hostPath"), runAsNonRoot != true (pod or container "container-00" must set securityContext.runAsNonRoot=true), runAsUser=0 (container "container-00" must not set runAsUser=0), seccompProfile (pod or container "container-00" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
E               Starting pod/ip-10-0-170-186us-east-2computeinternal-debug ...
E               To use host binaries, run `chroot /host`
E               
E               Removing debug pod ...
E               error: unable to create the debug pod "ip-10-0-170-186us-east-2computeinternal-debug"

See here:
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-trigger-aws-ipi-fips-encryption-3az-rhcos-3m-3w-3i-upgrade-ocs-auto/34/testReport/tests.ecosystem.upgrade/test_upgrade/test_upgrade/

PR #6210, opened 2022-07-20
Add retry for encryption verification

petr-balogh:
Fixes: #6209

### DIFF
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -12,6 +12,7 @@ from jsonschema import validate
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants, defaults, ocp, managedservice
 from ocs_ci.ocs.exceptions import (
+    CommandFailed,
     ResourceNotFoundError,
     UnsupportedFeatureError,
     PVNotSufficientException,
@@ -49,6 +50,7 @@ from ocs_ci.utility import (
     kms as KMS,
     version,
 )
+from ocs_ci.utility.retry import retry
 from ocs_ci.utility.rgwutils import get_rgw_count
 from ocs_ci.utility.utils import run_cmd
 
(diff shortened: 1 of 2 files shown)


ocs_ci/utility/retry.py
```diff
@@ -24,6 +25,16 @@ def f_retry(*args, **kwargs):
                 try:
                     return f(*args, **kwargs)
                 except exception_to_check as e:
+                    if text_in_exception:
+                        if text_in_exception in str(e):
+                            logger.debug(
+                                f"Text: {text_in_exception} found in exception: {e}"
+                            )
+                        else:
+                            logger.error(
```
clacroix12, line 34: Are we sure we want this at error level? This still feels like a debug level log entry to me.

shylesh (approved):
LGTM

ocs_ci/utility/retry.py
petr-balogh: Right after this we will anyway fail with exception raised. I thought that this error log will help us to easier identify the reason.

clacroix12: If we're raising the exception anyways then that seems like even more reason to not log this at error level. Since the user has specified a string they care about, I don't think we want to log at error level when that string wasn't matched.

That being said, the logic here looks good to me so I'm approving. We can always change the log level in a quick follow up PR.

clacroix12 (approved):

ocs_ci/utility/retry.py
petr-balogh: changed to debug level

shylesh (approved):
+1

clacroix12 (approved):

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -763,7 +765,15 @@ def osd_encryption_verification():
...
-        lsblk_out = run_cmd(lsblk_cmd)
+        # It happens from time to time that we see this error:
+        # error: unable to create the debug pod "node-name-internal-debug
+        # Hence we need to add some re-try logic here
+        lsblk_out = retry(
+            (CommandFailed),
+            tries=3,
+            delay=60,
+            text_in_exception="unable to create the debug pod",
+        )(run_cmd)(lsblk_cmd)
```
shyRozen, line 776: Please use https://github.com/red-hat-storage/ocs-ci/blob/f69e03df9dde1f07f3f6cf1671c46ef50ab2fad6/ocs_ci/ocs/ocp.py#L174 instead of run_cmd

petr-balogh: This function has a bit different logic and it doesn't return any message in the commandFailed exception here:
https://github.com/red-hat-storage/ocs-ci/blob/f69e03df9dde1f07f3f6cf1671c46ef50ab2fad6/ocs_ci/ocs/ocp.py#L199

So the logic I need will not probably work here.

Also I am not author of this code, just trying to improve the logic in the case the debug pod will fail to be created from some reason and re-try.

Also as part of this fix I don't want to update this exec_od_debug_cmd function at the moment as this is suppose to unblock one of our production upgrade jobs ASAP for 4.11 executions.

petr-balogh:
Verification job scheduled here:
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-trigger-aws-ipi-fips-encryption-3az-rhcos-3m-3w-3i-upgrade-ocs-auto/36/

Issue #6075, opened 2022-06-27
Some of the test items do not have squad attribute reported in XML junit file and then not in RP 


petr-balogh:
For example here:
http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/j-240vu1cs33-ua/j-240vu1cs33-ua_20220621T051522/logs/test_results_1655788203.xml

I don't see any Squad attribute added for example this test case:
tests.test_failure_propagator.TestFailurePropagator
or
tests.ui.test_pvc_ui.TestPvcUserInterface

clacroix12:
`tests.test_failure_propagator.TestFailurePropagator` is at the top level of the `tests` directory and therefore doesn't have a default squad marker assigned in the `SQUADS` constant. It also isn't decorated with a squad, so this doesn't appear to be a bug. If we want a test marker we should move this test to another appropriate location that has a squad assigned, or add a decorator to specify the squad.

`tests.ui.test_pvc_ui.TestPvcUserInterface` is decorated with a squad market at the class level but it is also parametrized which I think might be causing the issue here. I haven't pinned down the root cause yet but I'm testing a few things out regarding the test markers and parametrized tests. I will report back as soon as I have something.

clacroix12:
Another issue I'm seeing is that some test cases have multiple squad markers.

```
<testcase classname="tests.manage.mcg.ui.test_mcg_ui.TestStoreUserInterface" name="test_store_creation_and_deletion[namespacestore]" file="tests/manage/mcg/ui/test_mcg_ui.py" line="39" time="29.443">
<properties>
<property name="squad" value="Red"/>
<property name="squad" value="Black"/>
<property name="polarion-testcase-id" value="OCS-2547"/>
</properties>
</testcase>
```

In ReportPortal however, they only have one of the squad markers since we are moving on once we find one of them.
![Screen Shot 2022-07-26 at 11 02 50 AM](https://user-images.githubusercontent.com/1605721/181079051-f70c7d0b-4fe4-4751-ab27-467b87bd5ae5.png)

This is due to the filepaths we define in the `SQUADS` both being in the path for this test case (and several others).

```
SQUADS = {
    ...
    "Red": ["/mcg/", "/rgw/"],
    ...
    "Black": ["/ui/"],
    ...
}
```

This should get resolved when we change how we are assigning ownership, but I wanted to call it out now that I'm noticing it.

clacroix12:
[...]

PR #6233, opened 2022-07-26
Add squad marker decorator values to junit item properties

clacroix12:
Fixes: #6075 

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -185,18 +185,21 @@ def pytest_collection_modifyitems(session, items):
     # Add squad markers to each test item based on filepath
     for item in items:
         # check, if test already have squad marker manually assigned
+        skip_path_squad_marker = False
         for marker in item.iter_markers():
             if "_squad" in marker.name:
                 squad = marker.name.split("_")[0]
                 item.user_properties.append(("squad", squad.capitalize()))
-        for squad, paths in constants.SQUADS.items():
-            for _path in paths:
-                # Limit the test_path to the tests directory
-                test_path = os.path.relpath(item.fspath.strpath, constants.TOP_DIR)
-                if _path in test_path:
-                    item.add_marker(f"{squad.lower()}_squad")
-                    item.user_properties.append(("squad", squad))
-                    break
+                skip_path_squad_marker = True
+        if not skip_path_squad_marker:
+            for squad, paths in constants.SQUADS.items():
+                for _path in paths:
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -185,8 +185,10 @@ def pytest_collection_modifyitems(session, items):
     # Add squad markers to each test item based on filepath
     for item in items:
         # check, if test already have squad marker manually assigned
-        if any(map(lambda x: "_squad" in x.name, item.iter_markers())):
-            continue
+        for marker in item.iter_markers():
+            if "_squad" in marker.name:
+                squad = marker.name.split("_")[0]
+                item.user_properties.append(("squad", squad))
```
ebenahar, line 191: what if `_squad` not in the item's markers?

petr-balogh (dismissed):

tests/conftest.py
clacroix12: If there is no `_squad` in the item's existing markers we move onto the code below that will add the marker and property based on the [constant map](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/constants.py#L1540-L1558) we have defined for squads.

clacroix12:
Took a look at the report from the PR test and I noticed that we capitalize the squad value when applying from the map. I want to match this behavior so we will need to call capitalize() when adding from the markers.

We also probably want to break out of the loop once we find the `_squad` marker as well as skip the for loop that adds markers/properties based on the path pattern.

vavuthu (approved):

petr-balogh (approved):

clacroix12:
Tier1 junit xml: http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/jnk-pr6233b3212/jnk-pr6233b3212_20220728T173533/logs/test_results_1659029636.xml

All of the squad markers appear to be accurate. I checked a few specific test cases to make sure the correct marker appeared and went through each squad marker to make sure that there were no duplicates or multiple markers on a single test case. If there are no issues with the results here I think we can consider this verified. 

```
<testcase classname="tests.ui.test_pvc_ui.TestPvcUserInterface" name="test_create_resize_delete_pvc[ocs-storagecluster-cephfs-ReadWriteMany-2-Filesystem]" file="tests/ui/test_pvc_ui.py" line="29" time="398.792">
  <properties>
    <property name="squad" value="Green"/>
  </properties>
  ...
</testcase>
```

PR #6251, opened 2022-07-29
[WIP] PR to collect performence metrics data

pintojoy:
This PR is intended to collect CPU and memory performance data from Promotheus

Signed-off-by: pintojoy <jopinto@redhat.com>

### DIFF
--- a/ocs_ci/ocs/perftests.py
+++ b/ocs_ci/ocs/perftests.py
@@ -34,6 +34,9 @@ from ocs_ci.ocs.version import get_environment_info
 from ocs_ci.utility import templating
 from ocs_ci.utility.perf_dash.dashboard_api import PerfDash
 from ocs_ci.utility.utils import TimeoutSampler, get_running_cluster_id, ocsci_log_path
+from ocs_ci.utility.prometheus import PrometheusAPI
+from ocs_ci.framework import config as ocsci_config
+from ocs_ci.ocs.node import get_all_nodes
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


pintojoy:
Please find output in http://pastebin.test.redhat.com/1069814 , Ran "get_performance_metrics" on "tests/e2e/performance/csi_tests/test_pvc_snapshot_performance.py::TestPvcSnapshotPerformance::test_pvc_snapshot_performance[RBD-1]" test

jtaleric:
hey @pintojoy - if you are going to run this in conjunction with the benchmark-operator, we have a mechanism to collect prom data. 
additionally, you can use kube-burner to simply collect prom for a time interval if you wanted...

ocs_ci/ocs/perftests.py
```diff
@@ -1023,3 +1027,114 @@ def cleanup_testing_pod(self):
         except Exception:
             log.warning("The POD failed to delete")
             pass
+
+    def get_performance_metrics(self, start_time):
+        # This function is used to determine CPU and memory utilization of openshift storage pods and nodes
+        # start_time (secs) : start time of the test
+        from ocs_ci.ocs.resources import pod as pod1
```
keesturam, line 1034: Could you please move the import to the top of the file? 
we use import within a method only in case of circular dependency. If that is the case, please add a comment for the same

ocs_ci/ocs/perftests.py
```diff
@@ -1023,3 +1027,114 @@ def cleanup_testing_pod(self):
         except Exception:
             log.warning("The POD failed to delete")
             pass
+
+    def get_performance_metrics(self, start_time):
```
keesturam, line 1031: Can we also have an option to run this on a selected pod instead of all pods? 
it provides the ability to choose the metric one likes to collect as opposed to all by default. Maybe pod can also take `all` in which case it could collect metrics from all the pods of a namespace.  
```
def get_performance_metrics(self, start_time, pod):
```
(It may be a good idea to have the namespace included as an option too, that can go in a subsequent PR)

pintojoy: Sure. I will make changes to run for specific pod or all pods
Also i will add "step" parameter as an option too, so that tests that take longer time will not result in data flooding
i.e. if test completes in 2 mins we can have step size as 10seconds
Whereas if test takes few hours we can provide step size as 10 mins

ocs_ci/ocs/perftests.py
```diff
@@ -1023,3 +1027,114 @@ def cleanup_testing_pod(self):
...
+                    ),
+                )
+                list_val_m = mem_result[0][0]["values"]
+                mem_val = [float(el[1]) for el in list_val_m]
+                m_result[podd.name] = mem_val
+                min_mem[podd.name] = min(mem_val)
+                max_mem[podd.name] = max(mem_val)
+                mem_percentile_99[podd.name] = np.percentile(mem_val, 99)
+                mem_percentile_95[podd.name] = np.percentile(mem_val, 95)
+
```
Avilir, line 1114: please, create one dictionary, with all of the collected data, and then dump it to a  JSON file, so in the next step we ca push this file to the ES, and use the data in the Performance dashboard.

ocs_ci/ocs/perftests.py
ramkiperiy, line 1031: @pintojoy does this data available for nodes? is it possible to capture memory data of all OCS nodes?

Avilir (commented):
Can you add to the resuluts dic. the cluster_id ?

ebenahar:
@pintojoy should we keep the PR open?

PR #6255, opened 2022-07-31
Change reporting to odf-lvm-operator

shyRozen:


shyRozen:
Check locally and now it has reporting for lvmo. Running 2 validation. One tests for odf-operator and one test of lvm operator.

shyRozen:
The solution works for the test tun part when run locally. When running PR validation the mail comes with no version and my guess is that the versions are taken after deployment and this need to debug locally. Although all the PASSED still not verified.

ebenahar (approved):

PR #6258, opened 2022-07-31
Tier2 snapshot with node reboot and tier4 multiple reboot to test catalogsource presence

shyRozen:
Tier 2 testcase Iterates over volume mode and volume binding mode for snapshot with node reboot.
Tier 4 is doing multiple reboot, check lvm init (and this checks catalog). In case of failure with catalogsource it repairs it and count the failure within 10 reboot. Raise if there was a failure.

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -50,10 +50,12 @@ from ocs_ci.framework import config
 from ocs_ci.ocs import ocp, constants, exceptions
 from ocs_ci.ocs.exceptions import PoolNotFound
 from ocs_ci.ocs.resources.pvc import get_all_pvc_objs
+from ocs_ci.ocs.resources.pod import get_pods_having_label, wait_for_pods_to_be_running
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs.resources.ocs import OCS
 from ocs_ci.ocs.resources.pvc import PVC
 from ocs_ci.utility.connection import Connection
+from paramiko.ssh_exception import SSHException
 
 logger = logging.getLogger(__name__)
 
@@ -2193,6 +2195,7 @@ class LVM(object):
         Return:
             (LVM) object
         """
+        self.image = None
         self.lv_data = None
         self.lvmcluster = None
         self.pv_data = None
@@ -2201,6 +2204,8 @@ class LVM(object):
         self.vg_data = None
         self.node_ssh = None
         self.new_prom = None
+        self.lvm_pods = None
+
         func_list = [
             self.cluster_ip(),
             self.get_lvmcluster(),
@@ -2208,6 +2213,7 @@ class LVM(object):
             self.get_and_parse_pvs(),
             self.get_and_parse_lvs(),
             self.get_and_parse_vgs(),
+            self.get_lvm_pods_names(),
         ]
         extend_func_list = []
         if fstrim and fail_on_thin_pool_not_empty:
@@ -2287,6 +2293,7 @@ class LVM(object):
             **redhat_operators_catalogesource_ocp.data
         )
         image = getattr(redhat_operators_catalogesource_ocs, "data")["spec"]["image"]
+        self.image = image
         full_version = image.split(":")[1]
         pattern = re.compile(r"([^\.]*\.[^\.]*)")
         result = pattern.search(full_version)
@@ -2621,6 +2628,28 @@ class LVM(object):
                 f"✅ Pvc {pvc_obj.name} utilization is {pvc_data_percent} and expected is {pvc_expected_data_percent}"
             )
 
+    def get_lvm_pods_names(self):
+        """
+        Get lvm pods name (without operator), sets self.lvm_pods
+        """
+
+        self.lvm_pods = get_pods_having_label(
+            label="app.kubernetes.io/managed-by=lvm-operator",
+            namespace=constants.OPENSHIFT_STORAGE_NAMESPACE,
+        )
+
+    def wait_for_lvm_pod_running(self):
+        """
+        Wait for all lvm pods to be running.
+        """
+
+        pod_name_list = []
+        for lvm_pod in self.lvm_pods:
+            pod_name_list.append(lvm_pod["metadata"]["name"])
+        wait_for_pods_to_be_running(
+            pod_names=pod_name_list, namespace=constants.OPENSHIFT_STORAGE_NAMESPACE
+        )
+
     def cluster_ip(self):
         """
         Get cluster ip address for ssh connections, sets self.ip
@@ -2649,6 +2678,11 @@ class LVM(object):
             (str) output from server.
         """
         if not self.node_ssh:
+            self.create_ssh_object()
+        try:
+            return_output = self.node_ssh.exec_cmd(cmd=cmd)
+        except SSHException as er:
+            logger.info(f"Restoring ssh connection because {er}, ")
             self.create_ssh_object()
         return_output = self.node_ssh.exec_cmd(cmd=cmd)
         return_stdout = return_output[1]
--- a/tests/lvmo/test_lvm_node_reboot_catalogsource.py
+++ b/tests/lvmo/test_lvm_node_reboot_catalogsource.py
@@ -1,0 +1,78 @@
+import logging
+
+from ocs_ci.framework.pytest_customization.marks import tier4a, skipif_lvm_not_installed
+from ocs_ci.framework.testlib import skipif_ocs_version, ManageTest, bugzilla
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.cluster import LVM
+from ocs_ci.ocs.resources.pod import wait_for_lvm_pod_running
+from ocs_ci.ocs.exceptions import CommandFailed, CatalogSourceNotFoundAfterReboot
+from ocs_ci.ocs.node import wait_for_nodes_status
+from ocs_ci.deployment.deployment import create_catalog_source
+from ocs_ci.utility.utils import run_cmd
+from ocs_ci.ocs.node import get_node_objs
+
+logger = logging.getLogger(__name__)
+
+
+@bugzilla("2122785")
+@tier4a
+@skipif_lvm_not_installed
+@skipif_ocs_version("<4.11")
+class TestLvmCatalogSourceNodeReboot(ManageTest):
+    """
+    Test lvm snapshot bigger than disk
+
+    """
+
+    number_of_reboots = 10
+
+    def test_node_reboot_sno_catalogsource(
+        self,
+        nodes,
+    ):
+        """
+        test reboot multiple time to check catalogsource existence
+        .* Reboot node
+        .* Check that catalogsource exists
+        .* repeat number_of_reboots
+
+        """
+        lvm = LVM()
+
+        image = lvm.image
+        catalogsource_not_found_occurrences = 0
+        nodes_ocs = get_node_objs()
+        nodes_names = []
+        for reboot in range(1, (self.number_of_reboots + 1)):
+            for node in nodes_ocs:
+                nodes_names.append(node.data["metadata"]["name"])
+            nodes.restart_nodes(nodes_ocs, force=False, wait=False)
+
+            wait_for_nodes_status(node_names=nodes_names)
+
+            wait_for_lvm_pod_running()
+            logger.info(f"++++++++++++++++++ In reboot {reboot} ++++++++++++++++++")
+            try:
+                LVM()
+            except CommandFailed as er:
+                logger.error(
+                    f"CatalogSource "
+                    f"{constants.OPERATOR_CATALOG_SOURCE_NAME} not found in reboot {reboot}. {er}"
+                    f" Re-creating and continuing for statistics. Gathering events"
+                )
+                retrieve = run_cmd(
+                    cmd="oc get events --all-namespaces --sort-by='.metadata.creationTimestamp'"
+                )
+                splitted = retrieve.split("\n")
+                for line in splitted:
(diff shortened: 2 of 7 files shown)


shyRozen:
There are two runs because there 2 testcases for different tears (2,4). Tier2 passed. Tier4 had some teardown problem not related to the test. Maybe network disconnection to vcenter.
[2022-08-02T22:04:20.280Z] .venv/lib64/python3.8/site-packages/pyVmomi/SoapAdapter.py:1647: vmodl.fault.ManagedObjectNotFound

shyRozen:
Running again. Did some 2 lines cleanup.

ocs_ci/ocs/utils.py
```diff
@@ -1284,3 +1284,20 @@ def thread_init_class(class_init_operations, shutdown):
         if shutdown == 0:
             executor.shutdown(wait=True)
             return
+
+
+def get_catalogsource(name):
```
ebenahar, line 1289: please check if we can use functionalities from ocs_ci.ocs.resources.catalog_source.CatalogSource

tests/lvmo/test_lvm_node_reboot_catalogsource.py
```diff
@@ -0,0 +1,71 @@
...
+from ocs_ci.ocs.exceptions import CommandFailed, CatalogSourceNotFoundAfterReboot
+from ocs_ci.utility.vsphere_nodes import VSPHERENode
+from ocs_ci.ocs.node import wait_for_nodes_status
+from ocs_ci.deployment.deployment import create_catalog_source
+from ocs_ci.utility.utils import run_cmd
+
+logger = logging.getLogger(__name__)
+
+
+@tier4
```
ebenahar, line 16: We don't execute `tier4` jobs. Please add to one of tier4a/b/c.

tests/lvmo/test_lvm_node_reboot_catalogsource.py
```diff
@@ -0,0 +1,71 @@
...
+        .* Check that catalogsource exists
+        .* repeat number_of_reboots
+
+        """
+        lvm = LVM()
+
+        image = lvm.image
+        catalogsource_not_found_occurrences = 0
+        for reboot in range(1, (self.number_of_reboots + 1)):
+            node = VSPHERENode(host=lvm.ip, user="core")
```
ebenahar, line 42: VSPHERENode is not supposed to be used directly from the test. For such operations, we should use ocs_ci.ocs.platform_nodes.VMWareNodes.

A reference for such usage could be found in tests/manage/z_cluster/nodes/test_nodes_restart.py

ocs_ci/ocs/utils.py
shyRozen: I need to drop that function, I'm initiating LVM and he does also check catalogsource for LVM version so I'll drop it.

shyRozen:
image
Jenkins is not running. Did local runs. Tier2 passed
Tier4 catalogsource at some point was not found which is the test about. I don't want to open a BZ for the reason that I tried it on a new OCP version with 200 sequential reboots and it didn't fail. Moving to verified.

tests/lvmo/test_lvm_snapshot_node_reboot.py
```diff
@@ -0,0 +1,307 @@
...
+    ):
+        """
+        test create snapshot reboot
+        .* Create PVC
+        .* Create POD
+        .* Run IO PVC
+        .* Create Snapshot
+        .* Create pvc from Snapshot
+        .* Attach pod
+        .* Reboot node
```
ebenahar, line 76: please note that the current implementation of node reboot is cold power off. IMO, that is considered a disruption, as it's ungraceful. 
In https://github.com/red-hat-storage/ocs-ci/pull/5382, I wanted to add the option to gracefully restart a node

tests/lvmo/test_lvm_snapshot_node_reboot.py
```diff
@@ -0,0 +1,307 @@
...
+        fio_size = (
+            int(str(int(float(pvc_size)) * first_io_ratio).split(".")[0])
+            if "." in str(int(float(pvc_size)) * first_io_ratio)
+            else int(float(pvc_size) * first_io_ratio)
+        )
+        second_fio_size = (
+            int(str(int(float(pvc_size)) * second_io_ratio).split(".")[0])
+            if "." in str(int(float(pvc_size)) * second_io_ratio)
+            else int(float(pvc_size) * second_io_ratio)
+        )
```
ebenahar, line 102: seems like code duplication

tests/lvmo/test_lvm_snapshot_node_reboot.py
shyRozen: I think it is already implemented for Vsphere:
[https://github.com/red-hat-storage/ocs-ci/blob/55de5d72da977055339e07a4334759786823c121/tests/lvmo/test_lvm_snapshot_node_reboot.py#L218](https://github.com/red-hat-storage/ocs-ci/blob/55de5d72da977055339e07a4334759786823c121/tests/lvmo/test_lvm_snapshot_node_reboot.py#L218)
If you send force=True it will do VM reboot, False will do guest os reboot. I'm doing guest os reboot so I don't think it is disruptive.

tests/lvmo/test_lvm_node_reboot_catalogsource.py
```diff
@@ -0,0 +1,76 @@
...
+@tier4a
+@skipif_lvm_not_installed
+@skipif_ocs_version("<4.11")
+class TestLvmCatalogSourceNodeReboot(ManageTest):
+    """
+    Test lvm snapshot bigger than disk
+
+    """
+
+    number_of_reboots = 10
```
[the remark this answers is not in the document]
shyRozen: Just from running into the catalogsource disappearing problem I thought that 10 would catch the problem.
It was so funny, I had the problem and sent Jenkins run with 200 reboots to realize the statistics and it didn't happen :)) Probably it relates to the OCP version that it was fixed and because of that didn't open a BZ.

tests/lvmo/test_lvm_snapshot_node_reboot.py
shyRozen: The ratio are different. You mean to do it in a function and send the ratio?

tests/lvmo/test_lvm_node_reboot_catalogsource.py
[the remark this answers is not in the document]
shyRozen: I was wondering, didn't find an option to choose specific OCP image in ocs-ci code. Is there such option beside choosing 4.10, 4.11, ga etc?

AviadP (dismissed):

ocs_ci/ocs/cluster.py
```diff
@@ -2611,6 +2617,28 @@ def compare_percent_data_from_pvc(self, pvc_obj, data_size):
                 f"✅ Pvc {pvc_obj.name} utilization is {pvc_data_percent} and expected is {pvc_expected_data_percent}"
             )
 
+    def get_lvm_pods_names(self):
```
ebenahar, line 2620: get_lvm_pod_names

ocs_ci/ocs/cluster.py
```diff
@@ -2706,3 +2739,29 @@ def get_lvm_full_version():
...
+    """
+    Wait for all lvm pods to be running.
+    """
+
+    pod_name_list = []
+    for lvm_pod in get_lvm_pods_names():
+        pod_name_list.append(lvm_pod["metadata"]["name"])
+    wait_for_pods_to_be_running(
+        pod_names=pod_name_list, namespace=constants.OPENSHIFT_STORAGE_NAMESPACE
+    )
```
ebenahar, line 2767: these 2 functions are already in https://github.com/red-hat-storage/ocs-ci/pull/6258/files#diff-a4d73d82c94d13634aded838ea2490c117996218f405a8da15f9ecea093f3f6dR2620-R2641

tests/lvmo/test_lvm_snapshot_node_reboot.py
```diff
@@ -0,0 +1,308 @@
...
+        thin_pool_size = lvm.get_thin_pool1_size()
+        first_io_ratio = 0.1
+        second_io_ratio = 0.05
+
+        def calc_fio_size(io_ratio, pvc__size):
+            return (
+                int(str(int(float(pvc__size)) * io_ratio).split(".")[0])
+                if "." in str(int(float(pvc__size)) * io_ratio)
+                else int(float(pvc__size) * io_ratio)
+            )
```
ebenahar, line 99: this function should be available for other use cases so should be taken out of the test file

tests/lvmo/test_lvm_snapshot_node_reboot.py
```diff
@@ -0,0 +1,308 @@
...
+            restore_pvc_yaml=constants.CSI_LVM_PVC_RESTORE_YAML,
+            access_mode=access_mode,
+            status=status,
+            timeout=300,
+        )
+        pvc_restore_second_end_time = time.time()
+        logger.info(
+            f"Pvc creation time is {(pvc_restore_second_end_time - pvc_restore_second_start_time)}"
+        )
+        pod_3_start_time = time.time()
```
ebenahar, line 298: Instead of pod1, pod2, pod3, why can't we iterate over 3 pods with a for loop?

ocs_ci/ocs/cluster.py
shyRozen: I need to access this data without initializing the class because the class will fail if LVM pods are not up like after reboot I can't initialize lvm unless I know that the pods are up.

tests/lvmo/test_lvm_snapshot_node_reboot.py
shyRozen: My mistake is to do it in a complex way. I should only multiple the ratio and then use round so I don't need any function. I'll convert it to a simple phrase and will drop the function.

tests/lvmo/test_lvm_snapshot_node_reboot.py
shyRozen: I didn't called the pod 1,2,3. Only the timing which is nice to know the time after reboot, before reboot and so on.
The pods are created in different stages of the test.
1. origin pod before reboot
2. pod with restored snapshot before reboot IO and after reboot more IO.
3. pod with restored snapshot created after reboot. 
So the loop is not possible.
In the logging I'm using pods names
`f"Pod {restored_pod_obj_second.name} got to running state after {pod_3_elapsed_time}"`

ocs_ci/ocs/cluster.py
shyRozen: I don't really understand your point. We have multiple pods check so why use pod? So I'm getting pods names. Logically the name seems right.

ocs_ci/ocs/cluster.py
shyRozen: And in the class init I'm checking also for the pods to be running.

tests/lvmo/test_lvm_snapshot_node_reboot.py
ebenahar: Ok

tests/lvmo/test_lvm_snapshot_node_reboot.py
shyRozen: Changed to a simple phrase.

ocs_ci/ocs/cluster.py
```diff
@@ -2788,3 +2822,29 @@ def get_lvm_full_version():
     image = getattr(redhat_operators_catalogesource_ocs, "data")["spec"]["image"]
     full_version = image.split(":")[1]
     return full_version
+
+
+def get_lvm_pods_names():
```
AviadP, line 2827: 1. function returns pods object, not pods name, so i suggest to change function name accordingly
2. such functions can be found [here](https://github.com/red-hat-storage/ocs-ci/blob/5192d7601f9e8bc709c70b3dedba09afa97d31c3/ocs_ci/ocs/resources/pod.py#L693), so consider move it there

shyRozen: You are right. Will change the function name and move it. Thanks!

shyRozen:
The test is failing because there is a real issue with catalogsource. The test is working, 2 out of 10 reboots catalogsource is missing:
[2022-08-30T18:10:50.401Z] [1m[31mE           ocs_ci.ocs.exceptions.CatalogSourceNotFoundAfterReboot: Catalogsource redhat-operators not foundin 2 out of 10 reboots[0m
Opened [https://bugzilla.redhat.com/show_bug.cgi?id=2122785](https://bugzilla.redhat.com/show_bug.cgi?id=2122785)
Added @bugzilla.
Please review @ebenahar @AviadP

tests/lvmo/test_lvm_snapshot_node_reboot.py
```diff
@@ -0,0 +1,301 @@
...
+            pytest.param(
+                *[
+                    constants.VOLUME_MODE_FILESYSTEM,
+                    constants.IMMEDIATE_VOLUMEBINDINGMODE,
+                ],
+                marks=pytest.mark.polarion_id("OCS-3991"),
+            ),
+            pytest.param(
+                *[constants.VOLUME_MODE_BLOCK, constants.IMMEDIATE_VOLUMEBINDINGMODE],
+                marks=pytest.mark.polarion_id("OCS-3991"),
```
ebenahar, line 52: all polarion IDs are the same

shyRozen: The tests that where suppose to check this parameters have multiple polarion ID. All my other tests are parameterize this way and I'm not using different polarion ID for each parameter as it is not stated in the test plan.

Issue #6162, opened 2022-07-11
Fix the MS markers to refer to the "default-cluster-context-index" instead of checking the current cluster


yitzhak12:
Currently, the MS markers such as `skipif_ms_consumer`, `skipif_ms_provider`, `ms_provider_required`, etc., do not work as expected - which means it sometimes checks the first cluster and sometimes it checks the last cluster. 
This can result in skipping or running unnecessary tests on the wrong cluster.

The solution is to refer to the "default-cluster-context-index" instead of checking the current cluster.
By default, it should be the first cluster as defined in the code here: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/main.py#L132. 

For example, in the marker `skipif_ms_consumer` instead of writing `config.ENV_DATA["cluster_type"].lower() == "consumer"`, we change it into `config.clusters["default-cluster-context-index"].ENV_DATA["cluster_type"].lower() == "consumer"` 

We can also define a new var like `config.default_cluster` to make it easier to check.

yitzhak12:
This may relate also to the issue https://github.com/red-hat-storage/ocs-ci/issues/5776.

fbalak:
It makes sense. Other way to fix it would be to ensure that context is always on the default cluster when those markers are evaluated.
I think that context should be changed to different cluster only during a test. Changed context shouldn't persist for multiple clusters (that being said, there is also a problem with fixtures and what context they are applied to).

Issue #5831, opened 2022-05-03
Add docstrings to MultiClusterConfig methods


fbalak:
Methods in MultiClusterConfig are not documented and some of those methods like `attr_init`, `reset_ctx` or `get_defaults` are not self explanatory: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/__init__.py#L128

PR #6271, opened 2022-08-03
Modify the Managed service markers to use the default cluster context

yitzhak12:
fix https://github.com/red-hat-storage/ocs-ci/issues/6162

### DIFF
--- a/ocs_ci/framework/__init__.py
+++ b/ocs_ci/framework/__init__.py
@@ -148,6 +148,15 @@ class MultiClusterConfig:
 
     @property
     def default_cluster_ctx(self):
+        """
+        Get the default cluster context.
+        The default cluster context will be defined by the default index as defined in the
+        'ENV DATA' param 'default_cluster_context_index'
+
+        Returns:
+            ocs_ci.framework.Config: The default cluster context
+
+        """
         # Get the default index. If not found, the default value is 0
         default_index = self.cluster_ctx.ENV_DATA.get(
             "default_cluster_context_index", 0
--- a/ocs_ci/framework/__init__.py
+++ b/ocs_ci/framework/__init__.py
@@ -148,7 +148,11 @@ class MultiClusterConfig:
 
     @property
     def default_cluster_ctx(self):
-        return self.clusters[self.cluster_ctx.ENV_DATA["default_cluster_context_index"]]
+        # Get the default index. If not found, the default value is 0
+        default_index = self.cluster_ctx.ENV_DATA.get(
+            "default_cluster_context_index", 0
+        )
+        return self.clusters[default_index]
 
     def _single_cluster_init_cluster_configs(self):
         self.clusters.insert(0, Config())
--- a/ocs_ci/framework/__init__.py
+++ b/ocs_ci/framework/__init__.py
@@ -146,6 +146,10 @@ class MultiClusterConfig:
         self.single_cluster_default = True
         self._single_cluster_init_cluster_configs()
 
+    @property
+    def default_cluster_ctx(self):
+        return self.clusters[self.cluster_ctx.ENV_DATA["default_cluster_context_index"]]
+
     def _single_cluster_init_cluster_configs(self):
         self.clusters.insert(0, Config())
         self.cluster_ctx = self.clusters[0]
--- a/ocs_ci/ocs/managedservice.py
+++ b/ocs_ci/ocs/managedservice.py
@@ -280,13 +280,9 @@ def get_admin_key_from_provider():
         return admin_key
 
 
-def check_default_cluster_context_index_equal_to_current_index(switch_to_default=True):
+def check_default_cluster_context_index_equal_to_current_index():
     """
     Check that the default index is equal to the current index
-
-    Args:
-        switch_to_default(bool): If true, it will switch to the default cluster in case the default index
-            is different from the current index. False, otherwise. The default value is true.
 
     Returns:
         bool: True, if the default index is equal to the current index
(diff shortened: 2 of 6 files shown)


tests/managed-service/test_ms_markers.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @pytest.fixture(autouse=True)
+    def teardown(self):
+        """
+        Check that the Ceph health is OK
+        """
+        logger.info("Check that the Ceph health is OK")
+        ceph_health_check()
+
+    def test_default_cluster_context_index_equal_to_current_index(self):
```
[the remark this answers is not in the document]
yitzhak12: Okay, I will change it.

ocs_ci/framework/__init__.py
```diff
@@ -146,6 +146,14 @@ def __init__(self):
         self.single_cluster_default = True
         self._single_cluster_init_cluster_configs()
 
+    @property
+    def default_cluster_ctx(self):
```
fbalak, line 150: Please add a docstring to this method. We are already missing docstrings for most of MultiClusterConfig (https://github.com/red-hat-storage/ocs-ci/issues/5831) but we shouldn't add more undocumented content.

ocs_ci/framework/__init__.py
```diff
@@ -146,6 +146,14 @@ def __init__(self):
...
+        # Get the default index. If not found, the default value is 0
```
fbalak, line 151: Shouldn't this function also change `cur_index`? It looks like it only returns cluster config but name and comment suggest that it is worked with cluster index.

ocs_ci/ocs/managedservice.py
```diff
@@ -278,3 +278,30 @@ def get_admin_key_from_provider():
     finally:
         config.switch_ctx(initial_cluster_index)
         return admin_key
+
+
+def check_default_cluster_context_index_equal_to_current_index(switch_to_default=True):
```
fbalak, line 283: The name of the function contains `check*` string. I suggest not to set `switch_to _default` to `True` as it could be confusing because the function name suggests only check and not switching or setting any config value.

tests/managed-service/test_ms_markers.py
```diff
@@ -0,0 +1,96 @@
...
+@tier1
+@managed_service_required
+@ignore_leftovers
+class TestManagedServiceMarkers(ManageTest):
+    """
+    Test that the managed service markers work as expected
+    """
+
+    @pytest.fixture(autouse=True)
+    def teardown(self):
```
fbalak, line 34: Ceph health should be checked automatically after all tests. If this is missing in case of multi cluster then this needs to be handled in a more general way that works for all tests. Please create an issue if it is missing.

tests/managed-service/test_ms_markers.py
```diff
@@ -0,0 +1,96 @@
...
+    def test_runs_on_provider_marker(self):
+        """
+        Test that the 'runs_on_provider' marker work as expected
+        """
+        assert (
+            is_ms_provider_cluster()
+        ), "The cluster is not a provider cluster, even though we have the marker 'runs_on_provider'"
+        logger.info("The cluster is a provider cluster as expected")
+
+    def test_current_index_not_change_after_using_runs_on_provider(self):
```
fbalak, line 86: Pytest uses it's own logic to execute tests that changes order of test cases. Just having tests written in some order in the module doesn't enforce them to be executed in that order. In some cases test_runs_on_provider_marker can be ran after test_current_index_not_change_after_using_runs_on_provider. There exists some discussions how to fix it: https://stackoverflow.com/questions/17571438/test-case-execution-order-in-pytest or https://discuss.dizzycoding.com/test-case-execution-order-in-pytest/. I am not sure about what should be best approach but in our case it would be probably best to add [some hook](https://docs.pytest.org/en/latest/reference/reference.html#hooks) for it or discuss with others use of some plugin (I see from https://github.com/red-hat-storage/ocs-ci/blob/master/setup.py#L44 that we are installing pytest-ordering that can help).

fbalak (changes requested):

tests/managed-service/test_ms_markers.py
yitzhak12: Yes, I can remove it.

ocs_ci/framework/__init__.py
yitzhak12: Okay, that will do.

ocs_ci/framework/__init__.py
yitzhak12: Yes, the comment is just for the following line. As you mentioned, I will add a docstring to be more precise.

ocs_ci/ocs/managedservice.py
yitzhak12: The thing is that I want to change the current index to the default index if they are not equal.

tests/managed-service/test_ms_markers.py
yitzhak12: I see, thanks. I will use it

ocs_ci/ocs/managedservice.py
yitzhak12: I split it into three small functions.

yitzhak12:
I ran a PR validation job in my local and it passed. I uploaded the logs to http://rhsqe-repo.lab.eng.blr.redhat.com/OCS/ocs-qe-bugs/pr-6271/test_provider_and_consumer.log.

yitzhak12:
The pr validation job passed: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/667/console.

fbalak (dismissed):

dahorak (approved):

dahorak (approved):

fbalak (approved):

yitzhak12:
Pr validation job passed: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/682/console.

PR #6290, opened 2022-08-08
Test LVM overprovisioning

shyRozen:

### DIFF
--- a/tests/lvmo/test_lvm_overprovisioning.py
+++ b/tests/lvmo/test_lvm_overprovisioning.py
@@ -1,0 +1,101 @@
+import logging
+import pytest
+
+from ocs_ci.framework.pytest_customization.marks import tier1, skipif_lvm_not_installed
+from ocs_ci.framework.testlib import skipif_ocs_version, ManageTest, acceptance
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.cluster import LVM
+from ocs_ci.ocs.exceptions import ResourceWrongStatusException
+
+logger = logging.getLogger(__name__)
+
+
+@tier1
+@acceptance
+@skipif_lvm_not_installed
+@skipif_ocs_version("<4.11")
+class TestLvmOverProvisioning(ManageTest):
+    """
+    Test lvm snapshot bigger than disk
+
+    """
+
+    @pytest.mark.parametrize(
+        argnames=["volume_mode", "volume_binding_mode"],
+        argvalues=[
+            pytest.param(
+                *[
+                    constants.VOLUME_MODE_FILESYSTEM,
+                    constants.IMMEDIATE_VOLUMEBINDINGMODE,
+                ],
+                marks=pytest.mark.polarion_id("OCS-4469"),
+            ),
+            pytest.param(
+                *[constants.VOLUME_MODE_BLOCK, constants.IMMEDIATE_VOLUMEBINDINGMODE],
+                marks=pytest.mark.polarion_id("OCS-4469"),
+            ),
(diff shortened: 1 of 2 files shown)


AviadP (dismissed):

tests/lvmo/test_lvm_overprovisioning.py
```diff
@@ -0,0 +1,109 @@
...
+        .* Check thin pool size
+        .* Create PVCs till you reach overprovisioning
+        .* Check for error
+
+        """
+
+        access_mode = constants.ACCESS_MODE_RWO
+
+        lvm = LVM(fstrim=True, fail_on_thin_pool_not_empty=True)
+        thin_pool_size = lvm.get_thin_pool1_size()
```
ebenahar, line 60: `1` in the function's name is not a good practice as it's unclear what `1` meaning is. 
How about we call it `get_default_thin_pool_size`?

tests/lvmo/test_lvm_overprovisioning.py
```diff
@@ -0,0 +1,109 @@
...
+
+        if "t" in thin_pool_size:
```
ebenahar, line 62: what if the think pool size is in GB or MB?

tests/lvmo/test_lvm_overprovisioning.py
shyRozen: The name of the thin pool is thin-pool-1. It is clear for anyone that is working with lvm operator. It can be changed but this will mean change it in some other tests.

tests/lvmo/test_lvm_overprovisioning.py
shyRozen: In MB? Not a real life scenario. The else is taking the GB option. What can be done is add to all OS commands (lvs, pvs, vgs) `--unit g` and then all of the outputs will have G for sure.

ebenahar: adding `--unit g` makes sense here

tests/lvmo/test_lvm_overprovisioning.py
ebenahar: what if the pool name will change tomorrow? it would mean you will need to change the function name.
The better practice IMO is to call the function with a generic naming and save the pool name as a constant

shyRozen: I understand. Because we have about 4 PRs in the air let's do it in a separate PR after they will be merged. I'm opening an issue.

shyRozen: Raised #6358 to handle it in another PR.

tests/lvmo/test_lvm_overprovisioning.py
shyRozen: Added --unit g to all os calls. Removed parsing the unit from the test.

AviadP (approved):

tests/lvmo/test_lvm_overprovisioning.py
```diff
@@ -0,0 +1,101 @@
...
+from ocs_ci.framework.testlib import skipif_ocs_version, ManageTest, acceptance
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.cluster import LVM
+from ocs_ci.ocs.exceptions import ResourceWrongStatusException
+
+logger = logging.getLogger(__name__)
+
+
+@tier1
+@acceptance
```
ebenahar, line 14: this doesn't seem like an acceptance test case

PR #6338, opened 2022-08-17
pvc_snapshot TC Improvements

Avilir:
This PR fix some issues in the PVC snapshot TC.

* Don't deploy ElasticSearch server if it is defined already.
* replacing the Nginx image with the Perf image so, no FIO installation is needed.
* reuse of existing code

The changes reduce the memory consumption and the running time of the test.


OdedViner (approved):

pintojoy (approved):
LGTM

keemano (approved):

PR #6362, opened 2022-08-22
Create a new sanity helper class for the Managed Sevice

yitzhak12:
I created a new sanity helper class to efficiently run IO on multiple consumers and check the functionality of the Managed Service cluster/s. To run IO I use the fixture `create_scale_pods_and_pvcs_using_kube_job_on_ms_consumers` defined here https://github.com/red-hat-storage/ocs-ci/blob/master/tests/conftest.py#L5647.
Before we create the resources on the MS consumers, we need to pass an instance of the fixture `create_scale_pods_and_pvcs_using_kube_job_on_ms_consumers` to the `init_create_resources_on_ms_factory` method.
Also, we can specify extra parameters in the method - if not, it uses the default parameters.

After adding this new class, we can quickly modify or create new tests with the Managed Service.

The implementation is a little complicated, but the usage is easy and flexible, as you can see in the added test https://github.com/red-hat-storage/ocs-ci/pull/6362/files#diff-69a3460a68a96af51b4b130b9501ec3e2c1e47f761960a686a62e5f1937e8e89.

### DIFF
--- a/ocs_ci/helpers/sanity_helpers.py
+++ b/ocs_ci/helpers/sanity_helpers.py
@@ -10,8 +10,7 @@ from ocs_ci.helpers import helpers
 from ocs_ci.ocs.bucket_utils import s3_delete_object, s3_get_object, s3_put_object
 from ocs_ci.helpers.pvc_ops import create_pvcs
 from ocs_ci.utility.utils import ceph_health_check
-from ocs_ci.ocs.cluster import CephCluster, CephClusterExternal
-
+from ocs_ci.ocs.cluster import CephCluster, CephClusterExternal, is_ms_consumer_cluster
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


jilju:
LGTM

ocs_ci/helpers/sanity_helpers.py
```diff
@@ -183,3 +182,156 @@ def __init__(self):
...
+        if consumer_indexes:
+            self.consumer_indexes = consumer_indexes
+        elif is_ms_consumer_cluster():
+            self.consumer_indexes = [config.cur_index]
+        else:
+            self.consumer_indexes = config.get_consumer_indexes_list()
+
+        self.create_resources_on_ms_consumers_factory = (
+            create_scale_pods_and_pvcs_using_kube_job_on_ms_consumers
+        )
```
ebenahar, line 253: why not including the initialization of create_resources_on_ms_factory withing the `__init__()` method? that way, you will always have create_resources_on_ms_consumers_factory set and ready

ocs_ci/helpers/sanity_helpers.py
```diff
@@ -183,3 +182,156 @@ def __init__(self):
...
+        """
+        Create resources on MS consumers.
+        This function uses the factory "create_scale_pods_and_pvcs_using_kube_job_on_ms_consumers"
+        for creating the resources - Create scale pods, PVCs, and run IO using a Kube job on MS consumers.
+        Note: Before using this method, you need to first init the 'Create resources on MS consumers' factory
+        using the method 'init_create_resources_on_ms_factory'.
+
+        """
+        orig_index = config.cur_index
+        if not self.create_resources_on_ms_consumers_factory:
```
ebenahar, line 271: same

ocs_ci/helpers/sanity_helpers.py
yitzhak12: I thought it would be more flexible to do it in another function.
For example, if we want to run two functions with different parameters. 
From my side, we can change it to the `init` method.

yitzhak12: In that case, you need first to init the `create_resources_on_ms_consumers_factory`
in the method 'init_create_resources_on_ms_factory'.

ocs_ci/helpers/sanity_helpers.py
```diff
@@ -183,3 +182,156 @@ def __init__(self):
...
+class SanityManagedService(Sanity):
+    """
+    Class for cluster health and functional validations for the Managed Service
+    """
+
+    def __init__(self):
+        """
+        Init the sanity managed service class
+        """
+        super(Sanity, self).__init__()
```
ebenahar, line 196: could be that you will need to replace with:
```python
super(SanityManagedService, self).__init__()
```

ocs_ci/helpers/sanity_helpers.py
yitzhak12: I changed the initialization of `create_resources_on_ms_factory` to be in the `__init__` method

ocs_ci/helpers/sanity_helpers.py
yitzhak12: Yes, fixed.

yitzhak12:
Pr validation passed: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/674/console

ocs_ci/helpers/sanity_helpers.py
```diff
@@ -183,3 +182,133 @@ def __init__(self):
...
+        self.scale_count = scale_count
+        self.pvc_per_pod_count = pvc_per_pod_count
+        self.start_io = start_io
+        self.io_runtime = io_runtime
+        self.pvc_size = pvc_size
+        self.max_pvc_size = max_pvc_size
+
+        if consumer_indexes:
+            self.consumer_indexes = consumer_indexes
+        elif is_ms_consumer_cluster():
```
fbalak, line 240: If the current cluster is consumer then only the current cluster index is saved. Why are not stored also indexes of other consumers at this step?

tests/managed-service/test_sanity_ms.py
```diff
@@ -0,0 +1,124 @@
...
+    ManageTest,
+    ignore_leftovers,
+    managed_service_required,
+)
+
+log = logging.getLogger(__name__)
+
+
+@libtest
+@ignore_leftovers
```
fbalak, line 18: Why are leftovers ignored? What is created and not deleted as part of those tests?

ocs_ci/helpers/sanity_helpers.py
yitzhak12: The assumption is that if the cluster is a consumer, we will prefer to create resources only for the consumer. And if it's a provider, we like to create resources for all the consumers.

tests/managed-service/test_sanity_ms.py
yitzhak12: Yes, you're right. I removed it.

yitzhak12:
Pr validation job passed: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/683/console.

ocs_ci/helpers/sanity_helpers.py
fbalak: This sounds inconsistent and complicate to me. In all other cases all indexes are stored in `self.consumer_indexes` variable and later used but in one case it is not stored depending on where the test is executed. I would prefer if this was consistent and the check was triggered on all available consumers no matter where it is executed. If we want an option to check health on one cluster then I would suggest to use only the `consumer_indexes` parameter of the function as you already have defined.

fbalak (commented):
LGTM, I just have one doubt.

ocs_ci/helpers/sanity_helpers.py
fbalak: After our discussion, I agree that your approach is ok. In most cases we don't have to check all consumers. Please just add more log messages into the code to make it easier to identify what indexes were passed and what should be checked.

yitzhak12: Okay, np.

ebenahar (dismissed):

yitzhak12:
The pr validation passed: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/696/console.

ebenahar (approved):

jilju (approved):

PR #6385, opened 2022-08-28
WIP: Workloads on MS

OdedViner:
Signed-off-by: oviner <oviner@redhat.com>


tests/e2e/conftest.py
```diff
@@ -418,3 +425,209 @@ def mcg_system_recovery_check(mcg_sys_setup_dict):
...
 
     return mcg_system_recovery_check
+
+
+@pytest.fixture(scope="class")
+def jenkins_factory(request):
+    return jenkins_factory_fixture(request)
+
+
+def jenkins_factory_fixture(request):
```
PrasadDesala, line 435: we already have fixtures for workloads with these names, is it possible to combine these two to  a single fixture?

tests/e2e/workloads/app/managed_srevice/test_workload_managed_service.py
```diff
@@ -0,0 +1,88 @@
+import logging
```
PrasadDesala, line 1: typo in the managed_service dir name, please correct it

tests/e2e/conftest.py
OdedViner: done

PR #6400, opened 2022-09-01
Added the flexibilty to skip waiting for the workload to complete to all workload fixtures

PrasadDesala:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2929,6 +2929,7 @@ def pgsql_factory_fixture(request):
         samples=None,
         timeout=None,
         sc_name=None,
+        wait_for_pgbench_to_complete=True,
     ):
         """
         Factory to start pgsql workload
@@ -2940,8 +2941,8 @@ def pgsql_factory_fixture(request):
             transactions (int): Number of transactions
             scaling_factor (int): scaling factor
             samples (int): Number of samples to run
-
             timeout (int): Time in seconds to wait
+            wait_for_pgbench_to_complete (bool): If set to True, the fixture will wait for the pgbench run to complete
 
         """
         # Setup postgres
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -3143,12 +3156,18 @@ def factory(
         log.info(f"Waiting for {waiting_time}sec to generate msg")
         time.sleep(waiting_time)
 
-        # Check messages are sent and received
-        threads = amq.run_in_bg(
-            namespace=kafka_namespace, value=value, since_time=since_time
-        )
+        if validate_messages:
```
mashetty330, line 3159: what if only run_in_bg=True and validate_messages=False ? This condition doesnt seems to be correct

tests/conftest.py
```diff
@@ -3143,12 +3156,18 @@ def factory(
...
+                threads = amq.run_in_bg(
+                    namespace=kafka_namespace, value=value, since_time=since_time
+                )
+                return amq, threads
+            else:
+                amq.validate_messages_are_produced()
+                amq.validate_messages_are_consumed()
 
-        return amq, threads
+        return amq
```
mashetty330, line 3171: you may as well have to check all the tests for the respective workloads and make necessary changes

tests/conftest.py
PrasadDesala: run_in_bg has to be called only when validate_messages is set to True. Now the code will throw exception for this case

tests/conftest.py
PrasadDesala: Not needed as the newly introduced flags are by default set to True. I have plans to trigger the entire suite if I find any regressions will make the appropriate changes

OdedViner (approved):

mashetty330 (approved):

vavuthu (approved):

PR #6425, opened 2022-09-08
Create new markers for checking if we use the Managed Service with provider and consumer clusters

yitzhak12:
I implement the following in the pr:
  1. Create new markers for checking if we use the Managed Service with provider and consumer clusters:
  
        - **ms_provider_and_consumer_required**
        - **skipif_ms_provider_and_consumer**
  
  2. Use these markers in the test "test_create_scale_pods_and_pvcs_using_kube_job".

### DIFF
--- a/ocs_ci/framework/__init__.py
+++ b/ocs_ci/framework/__init__.py
@@ -330,5 +330,27 @@ class MultiClusterConfig:
         """
         return self.ENV_DATA.get("cluster_name")
 
+    def is_provider_exist(self):
+        """
+        Check if the provider cluster exists in the clusters
+
+        Returns:
+            bool: True, if the provider cluster exists in the clusters. False, otherwise.
+
+        """
+        cluster_types = [cluster.ENV_DATA["cluster_type"] for cluster in self.clusters]
+        return "provider" in cluster_types
+
+    def is_consumer_exist(self):
+        """
+        Check if the consumer cluster exists in the clusters
+
+        Returns:
+            bool: True, if the consumer cluster exists in the clusters. False, otherwise.
+
+        """
+        cluster_types = [cluster.ENV_DATA["cluster_type"] for cluster in self.clusters]
+        return "consumer" in cluster_types
+
 
 config = MultiClusterConfig()
--- a/tests/e2e/workloads/test_create_scale_pods_and_pvcs_using_kube_job.py
+++ b/tests/e2e/workloads/test_create_scale_pods_and_pvcs_using_kube_job.py
@@ -12,7 +12,9 @@ from ocs_ci.framework.testlib import (
 from ocs_ci.framework.pytest_customization.marks import (
     skipif_external_mode,
     ipi_deployment_required,
-    managed_service_required,
+    ms_provider_and_consumer_required,
+    skipif_ms_provider_and_consumer,
+    ms_consumer_required,
 )
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs import constants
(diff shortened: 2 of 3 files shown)


yitzhak12:
I checked the pr with three scenarios:

1. **Test on provider and consumer** - test results: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/724/testReport/tests.e2e.workloads.test_create_scale_pods_and_pvcs_using_kube_job/TestCreateScalePodsAndPvcsUsingKubeJob/. The expected test run and passed: **test_create_scale_pods_and_pvcs_using_kube_job_ms**.
2. **Test on a consumer** - test results: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/16571/testReport/tests.e2e.workloads.test_create_scale_pods_and_pvcs_using_kube_job/TestCreateScalePodsAndPvcsUsingKubeJob/. The expected test run and passed: **test_create_scale_pods_and_pvcs_with_ms_consumer**.
3. **Test on a provider** - test results: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/16573/testReport/tests.e2e.workloads.test_create_scale_pods_and_pvcs_using_kube_job/TestCreateScalePodsAndPvcsUsingKubeJob/. **All the tests skipped as expected**.

tests/e2e/workloads/test_create_scale_pods_and_pvcs_using_kube_job.py
```diff
@@ -88,10 +90,32 @@ def test_create_scale_pods_and_pvcs_using_kube_job_ms(
         config.switch_to_consumer()
         log.info("The resources created successfully using the kube job")
 
+    @skipif_ms_provider_and_consumer
```
jilju, line 93: Why do we need to skip this when both the provider and consumer are present ? Instead, I think, this test case can use the consumer alone for running.

yitzhak12: The idea behind this is that we want to run some of the tests only with provider and consumer, but sometimes we want to cover the same tests above(with provider and consumer only) to run with a single cluster. 

For example, if we have one test with AWS and we want to create a similar test with vSphere - there is no need to run the second test with AWS, as we covered in the first test.

jilju (approved):

ocs_ci/framework/__init__.py
```diff
@@ -330,5 +330,27 @@ def current_cluster_name(self):
...
+    def is_provider_exist(self):
+        """
+        Check if the provider cluster exists in the clusters
+
+        Returns:
+            bool: True, if the provider cluster exists in the clusters. False, otherwise.
+
+        """
+        cluster_types = [cluster.ENV_DATA["cluster_type"] for cluster in self.clusters]
+        return "provider" in cluster_types
```
dahorak, line 342: I'm thinking if it will not be little bit more readable with using `any(...)`, something like this:

```python
return any([cluster.ENV_DATA["cluster_type"] == "provider" for cluster in self.clusters])
```
But I'm ok with leaving as it is.

tests/e2e/workloads/test_create_scale_pods_and_pvcs_using_kube_job.py
dahorak: Are we going to regularly (in CI) run MS related jobs in single cluster scenario?

dahorak (approved):
Only small note, but I'm ok with merging it as it is. LGTM

OdedViner (approved):

tests/e2e/workloads/test_create_scale_pods_and_pvcs_using_kube_job.py
yitzhak12: I saw that in tier1, we run the tests with one provider or consumer. I didn't see a run with multiple consumers and no provider - but we also should consider this scenario

Issue #6426, opened 2022-09-09
teardown fails due to missing metadatajson


vavuthu:
```
2022-09-08 12:37:12          # Retrieve cluster metadata
2022-09-08 12:37:12          metadata_file = os.path.join(self.cluster_path, "metadata.json")
2022-09-08 12:37:12  >       with open(metadata_file) as f:
2022-09-08 12:37:12  E       FileNotFoundError: [Errno 2] No such file or directory: '/home/jenkins/current-cluster-dir/openshift-cluster-dir/metadata.json'
```
job:  https://url.corp.redhat.com/78648aa

PR #6427, opened 2022-09-09
donot check for metadata json file for teardown of cluster

vavuthu:
Fixes: #6426 

### DIFF
--- a/ocs_ci/deployment/ocp.py
+++ b/ocs_ci/deployment/ocp.py
@@ -178,10 +178,6 @@ class OCPDeployment:
             log_level (str): log level openshift-installer (default: DEBUG)
 
         """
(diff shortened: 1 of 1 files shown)


petr-balogh (commented):
Don't we read some info like cluster name or uuid which we later on are using for destroying some related resources?

vavuthu:
> Don't we read some info like cluster name or uuid which we later on are using for destroying some related resources?

previously we used to read cluster name and destroy cluster and then destroy volumes (using cluster name ). A lot of things changed over time and below is the logic now
destroy volumes ---> destroy cluster buckets ---> uninstall ocs ---> destroy cluster

AviadP (approved):

Issue #6393, opened 2022-08-31
Automated LVMO deployment failed  -4.12


AviadP:
build url:  https://url.corp.redhat.com/a523f34 

```
2022-08-31 12:52:13          bundle_config_file = os.path.join(
2022-08-31 12:52:13              constants.TEMPLATE_DEPLOYMENT_DIR_LVMO, "lvm-bundle.yaml"
2022-08-31 12:52:13          )
2022-08-31 12:52:13          run_cmd(f"oc create -f {bundle_config_file} -n {self.namespace}")
2022-08-31 12:52:13          pod = ocp.OCP(kind=constants.POD, namespace=self.namespace)
2022-08-31 12:52:13          assert pod.wait_for_resource(
2022-08-31 12:52:13              condition="Running",
2022-08-31 12:52:13  >           selector=label_version[lvmo_version_without_period][
2022-08-31 12:52:13                  "controller_manager_label"
2022-08-31 12:52:13              ],
2022-08-31 12:52:13              resource_count=1,
2022-08-31 12:52:13              timeout=300,
2022-08-31 12:52:13          )
2022-08-31 12:52:13  E       KeyError: '412'

```

AviadP:
need to add key for "412" [here](https://github.com/red-hat-storage/ocs-ci/blob/7d0050c4e2281ddd577acb8651720935d21c8d3b/ocs_ci/ocs/constants.py#L1848):

PR #6457, opened 2022-09-13
Change pod label dictionary above 4.11 to default and add pod security label to openshift-storage

shyRozen:
Change pod label dictionary above 4.11 to default and add pod security label to openshift-storage
Fixes: #6393 

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -1269,8 +1269,9 @@ class Deployment(object):
             f"lvm-cluster-{file_version}.yaml",
         )
 
-        if int(lvmo_version_without_period) >= 411:
+        if version.get_semantic_ocs_version_from_config() >= version.VERSION_4_11:
             lvmo_version_without_period = "default"
+
         # this is a workaround for 2101343
         if 110 > int(minor) > 98 and major == "4.11.0":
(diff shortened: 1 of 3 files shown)


AviadP (approved):

Avilir (dismissed):

ocs_ci/deployment/deployment.py
```diff
@@ -1268,6 +1268,9 @@ def deploy_lvmo(self):
             constants.TEMPLATE_DEPLOYMENT_DIR_LVMO,
             f"lvm-cluster-{file_version}.yaml",
         )
+
+        if int(lvmo_version_without_period) >= 411:
```
ebenahar, line 1272: can we use  Version.coerce()?

shyRozen: If we have already the data available why add more actions? What is the benefit. If the data wasn't here I would do it with Version.corece but since I have already the data from other actions I do here, I don't understand the benefit.

ebenahar: My intention is that we will not convert to int but rather use a proper way of comparing versions, semantically

ebenahar (approved):

AviadP (approved):

PR #6475, opened 2022-09-16
Set AWS SC to gp2-csi from 4.12

petr-balogh:

### DIFF
--- a/ocs_ci/deployment/aws.py
+++ b/ocs_ci/deployment/aws.py
@@ -63,6 +63,9 @@ class AWSBase(CloudDeploymentBase):
         # dict of cluster prefixes with special handling rules (for existence
         # check or during a cluster cleanup)
         self.cluster_prefixes_special_rules = CLUSTER_PREFIXES_SPECIAL_RULES
+        ocp_version = version.get_semantic_ocp_version_from_config()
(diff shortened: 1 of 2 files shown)


petr-balogh:
I see it properly selected the gp2-csi SC:
```2022-09-16 23:16:39          storageClassName: gp2-csi```

File: ocs_ci/deployment/aws.py [hunk 1]
```diff
@@ -63,6 +63,9 @@ def __init__(self):
         # dict of cluster prefixes with special handling rules (for existence
         # check or during a cluster cleanup)
         self.cluster_prefixes_special_rules = CLUSTER_PREFIXES_SPECIAL_RULES
+        ocs_version = version.get_semantic_ocs_version_from_config()
```
vavuthu, line 66: since gp2-csi chnages are from OCP side, we can set DEFAULT_STORAGECLASS based on running OCP version

petr-balogh: makes sense to change it to OCP version, but not running one as this is the deployment class and we don't have the OCP running yet.

vavuthu (approved):

dahorak (approved):

PR #6483, opened 2022-09-19
Enhanced Pgsql workload to support running sql queries directly on Postgres pods (without Pgbench)

PrasadDesala:

### DIFF
--- a/ocs_ci/ocs/pgsql.py
+++ b/ocs_ci/ocs/pgsql.py
@@ -322,7 +322,6 @@ class Postgresql(BenchmarkOperator):
             output = run_cmd(f"oc logs {pgbench_pod.name} -n {BMO_NAME}")
             pg_output = utils.parse_pgsql_logs(output)
             log.info("*******PGBench output log*********\n" f"{pg_output}")
-            # for data in all_pgbench_pods_output:
             for data in pg_output:
                 run_id = list(data.keys())
                 latency_avg = data[run_id[0]]["latency_avg"]
@@ -719,8 +718,8 @@ class Postgresql(BenchmarkOperator):
             )
             result.append("-" * len(result[0]))
             log.info("\n".join(result))
-            return out
         log.info(out)
+        return out
 
     def run_insert_operation(self, num_of_ops, postgres_pod, row_index, table_name):
         """
--- a/ocs_ci/ocs/pgsql.py
+++ b/ocs_ci/ocs/pgsql.py
@@ -5,7 +5,9 @@ import logging
 import random
 import time
 from prettytable import PrettyTable
-from datetime import datetime
+import string
+import re
+from datetime import datetime, timedelta
 
 from ocs_ci.ocs.benchmark_operator import BenchmarkOperator, BMO_NAME
 from ocs_ci.utility.utils import TimeoutSampler, run_cmd
@@ -48,7 +50,6 @@ class Postgresql(BenchmarkOperator):
     def __init__(self, **kwargs):
         """
         Initializer function
-
         """
         super().__init__(**kwargs)
         BenchmarkOperator.deploy(self)
@@ -56,13 +57,10 @@ class Postgresql(BenchmarkOperator):
     def setup_postgresql(self, replicas, sc_name=None):
         """
         Deploy postgres sql server
-
         Args:
             replicas (int): Number of postgresql pods to be deployed
-
         Raises:
             CommandFailed: If PostgreSQL server setup fails
-
         """
         log.info("Deploying postgres database")
         try:
@@ -115,7 +113,6 @@ class Postgresql(BenchmarkOperator):
     ):
         """
         Create pgbench benchmark pods
-
         Args:
             replicas (int): Number of pgbench pods to be deployed
             pgbench_name (str): Name of pgbench bechmark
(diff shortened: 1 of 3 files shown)


ocs_ci/ocs/pgsql.py
```diff
@@ -319,9 +322,11 @@ def validate_pgbench_run(self, pgbench_pods, print_table=True):
             output = run_cmd(f"oc logs {pgbench_pod.name} -n {BMO_NAME}")
             pg_output = utils.parse_pgsql_logs(output)
             log.info("*******PGBench output log*********\n" f"{pg_output}")
+            # for data in all_pgbench_pods_output:
```
Avilir, line 325: Do we need this line ?

ocs_ci/ocs/pgsql.py
```diff
@@ -663,3 +664,193 @@ def pgsql_full(self):
...
+            ind = out.index("-")
+            result = list()
+            result.append("\n" + out[:ind])
+            result.append("-" * len(result[0]))
+            result.extend(
+                re.findall(r"\d+\s\|\s[A-Z0-9]+\s\|\s\d{4}-\d{2}-\d{2}\s", out)
+            )
+            result.append("-" * len(result[0]))
+            log.info("\n".join(result))
+            return out
```
Avilir, line 722: It is better to return the command output in any case and not for select only.

ocs_ci/ocs/pgsql.py
```diff
@@ -663,3 +664,193 @@ def pgsql_full(self):
...
+            postgres_pod (POD) : POD object on which postgres is running.
+            row_index (int) : Starting index of the rows to be inserted.
+            table_name (str) : Name of the table.
+
+        """
+        log.info(f"Running {num_of_ops} insert operations.")
+        res = ""
+        for row_num in range(row_index + 1, row_index + num_of_ops + 1):
+            random_username = self.generate_random_string(50)
+            res += f"({row_num}, '{random_username}', '{self.generate_random_date()}'),"
```
Avilir, line 740: Is it ok to have comma in the end of the command ?
maybe you should delete the last comma outside the loop.

also I think you should have a space after the comma.

ocs_ci/ocs/pgsql.py
```diff
@@ -663,3 +664,193 @@ def pgsql_full(self):
...
+
+        Returns:
+             str  : A string containing the last valid state of the table.
+
+        """
+        row_index = 0
+        last_valid_state = ""
+        self.run_pgsql_command(
+            postgres_pod,
+            f"CREATE TABLE {table_name} (row_id INT PRIMARY KEY, username VARCHAR (50) NOT NULL, date DATE NOT NULL);",
```
Avilir, line 819: Can you make it more generic, so the table structure will be take from an outside file (YAML / JSON) ?

this is relevant for all other functions where the table filed are constants.

ocs_ci/ocs/pgsql.py
```diff
@@ -663,3 +664,193 @@ def pgsql_full(self):
...
+                        (
+                            f"UPDATE {table_name} SET date='{self.generate_random_date()}' "
+                            f"WHERE EXTRACT(YEAR FROM date) = {str(random.randint(1900, datetime.now().year))};"
+                        ),
+                    )
+                else:
+                    self.run_delete_operation(
+                        num_of_ops, postgres_pod, row_index, table_name
+                    )
+                last_valid_state = self.run_pgsql_command(
```
Avilir, line 849: Since the last operation before this query is delete the row you add, this will return  an empty value, of error.
it should be in the loop before the delete.
also, it will be better to log the output which return.

ocs_ci/ocs/pgsql.py
PrasadDesala: Not need, removed.

ocs_ci/ocs/pgsql.py
PrasadDesala: Done

ocs_ci/ocs/pgsql.py
PrasadDesala: the res contains the randomly generated data to insert into the table. Each of the entry will be comma separated. 
Below is the output snip of it,

(1, 'YT00ARUZOC8NESHADPZ3SJSTQ42G4INYN43RZS6XLFHSY1YNX0', '1921-11-28'),(2, 'HRC4EZW0ZZIREEMUV6MIOZN12DJB5WB7CEDEF109QI6G4WXE90', '1906-11-19'),(3, 'ON6YVTDG74GVYCXC7659K9OU8UQMB7T76B24JT18TOP0S8FN3L', '1984-08-25'),

ocs_ci/ocs/pgsql.py
PrasadDesala: The table and its data cannot be generated running a YAML, we can read the content from the YAML/JSON and based on that we can construct the data to be inserted into the table but not sure if that is really needed for this use case and I do not see much benefit implementing that way.

ocs_ci/ocs/pgsql.py
PrasadDesala: here we are not deleting all the row. Only the num_of_ops number of rows will be deleted from the table. So last_valid_state will never be empty at this line. last_valid_state captures the snapshot of the last valid state of the table, this can be used for checking the data integrity after certain operations or failures.

PR #6492, opened 2022-09-21
Use the ingress cert for IBM Cloud deployments

MeridianExplorer:
Up until now the default CA bundle was used; Now, it leads to an SSL error.
This PR changes the behavior of MCG tests and fixtures running over IBM Cloud clusters to use the cluster's ingress certificate.


clacroix12 (approved):
LGTM but also looks like the DCO check is failing because the second commit isn't signed off.

clacroix12 (approved):

petr-balogh (approved):

dahorak (approved):

vavuthu (approved):

Issue #6462, opened 2022-09-13
Handle endpoint unavailability after upgrade


fbalak:
After upgrade there often happens that oc login creates invalid kubeconfig. The invalid kubeconfig contains all data but at the last line is dropped invalid string containing part of certificate used in the file. The kubeconfig is valid without this line. 

This is caused by cycle in a prometheus credentials refreshing loop. There is a fix for refreshing prometheus credentials when endpoint is unavailable. This usually fixes minor issues with endpoint availability but after upgrade after the endpoint is drained, the endpoint is unavailable for longer time and there is too many attempts to refresh the connection and during one of this attempts is corrupted kubeconfig (there is called oc login multiple times in a short time).

Code that refreshes connection after endpoint unavailability needs to be fixed: https://github.com/red-hat-storage/ocs-ci/blob/6357c5040735aecd22e6a5513f261432accee051/ocs_ci/utility/prometheus.py#L423

PR #6493, opened 2022-09-21
Use TimeoutIterator to fix too many failed Prometheus requests after upgrade

fbalak:
Fixes https://github.com/red-hat-storage/ocs-ci/issues/6462

### DIFF
--- a/ocs_ci/utility/prometheus.py
+++ b/ocs_ci/utility/prometheus.py
@@ -393,7 +393,7 @@ class PrometheusAPI(object):
         self._cacert = cert_file.name
         logger.info(f"Generated CA certification file: {self._cacert}")
 
-    def get(self, resource, payload=None):
+    def get(self, resource, payload=None, timeout=300):
         """
         Get alerts from Prometheus API.
 
@@ -403,6 +403,8 @@ class PrometheusAPI(object):
             payload (dict): Provide parameters to GET API call.
                 e.g. for `alerts` resource this can be
                 {'silenced': False, 'inhibited': False}
+            timeout (int): Number of seconds to wait for Prometheus endpoint to
+                get available if it is not available
 
         Returns:
             dict: Response from Prometheus alerts api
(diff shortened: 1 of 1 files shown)


ocs_ci/utility/prometheus.py
```diff
@@ -414,27 +415,28 @@ def get(self, resource, payload=None):
...
-        return response
+        for response in TimeoutIterator(
+            timeout=300,
+            sleep=15,
+            func=requests.get,
+            func_kwargs={
+                "url": self._endpoint + pattern,
+                "headers": headers,
+                "verify": self._cacert,
+                "params": payload,
```
mbukatov, line 426: Didn't you mention that you plan to tweak a timeout in the get call as well?

mbukatov (changes requested):

ocs_ci/utility/prometheus.py
fbalak: No, I mentioned only that both TimeoutSampler and requests.get functions have timeout parameter and it can cause a conflict. After that you advised me to use TimeoutIterator class that should solve this conflict in parameters because parameters are passed as dictionary.

ocs_ci/utility/prometheus.py
```diff
@@ -414,27 +415,28 @@ def get(self, resource, payload=None):
...
+            if response.status_code == 503:
+                logger.warning(f"There was an error in response: {response.text}")
+                logger.warning("Refreshing connection")
+                self.refresh_connection()
+                if not config.ENV_DATA["platform"].lower() == "ibm_cloud":
+                    logger.warning("Generating new certificate")
+                    self.generate_cert()
+                logger.warning("Connection refreshed")
+            else:
+                break
```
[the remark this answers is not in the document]
fbalak: The returned response is placed after the loop. This way the response is returned even when the function fails and a response with code 503 is returned. This is done for debugging purposes and to keep the function's return clause consistent no matter what is the response.

ocs_ci/utility/prometheus.py
```diff
@@ -414,27 +417,37 @@ def get(self, resource, payload=None):
...
+                else:
+                    break
+            return response
+        else:
+            return requests.get(
                 self._endpoint + pattern,
                 headers=headers,
                 verify=self._cacert,
                 params=payload,
             )
```
mbukatov, line 450: I would rather moved this section into the beginning so that you can handle this case in a separate code block, and don't have to think about it later.

mbukatov (approved):

fbalak:
Verified: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/16925/

Issue #6503, opened 2022-09-23
Get resource is broken for get node status


petr-balogh:
https://github.com/red-hat-storage/ocs-ci/pull/6144/files
This fix actually broke the other results for this command.

E.g.
resource =  'NAME            STATUS   ROLES           AGE   VERSION 10.243.128.20   Ready    master,worker   22h   v1.23.5+012e945'
at the end of the function we have:
ipdb> resource_info
['Ready', 'master,worker', '22h', 'v1.23.5+012e945']
ipdb> titles
['NAME', 'STATUS', 'ROLES', 'AGE', 'VERSION 10.243.128.20']

So when we want to get status we get `master,worker` instead.

Issue #6138, opened 2022-07-06
test_dynamic_pvc_accessmodes_with_reclaim_policies failed due to incorrect status check


rachael-george:
In `ocs-ci results for OCS4-11-Downstream-OCP4-11-VSPHERE6-UPI-1AZ-RHCOS-VSAN-3M-3W-acceptance (BUILD ID: 4.11.0-110 RUN ID: 1657098489)` the tests failed in teardown phase while checking if the status of the PV is in released state for PVCs with retain policy.

`Message: failed on teardown with "ocs_ci.ocs.exceptions.ResourceWrongStatusException: Resource pvc-029b590e-23d5-4b76-add2-39bc078885ee describe `

The PV was in released state but the `wait_for_resource` function returned the status as the name of the storageclass instead of the status. 

`2022-07-06 09:19:27,437 - MainThread - ERROR - ocs_ci.ocs.ocp.wait_for_resource.685 - Wait for PersistentVolume resource pvc-029b590e-23d5-4b76-add2-39bc078885ee at column STATUS to reach desired condition Released failed, last actual status was storageclass-test-rbd-644ae4125df64e7ca3`

This error wasn't seen in other runs with the same builds and will need further investigation

rachael-george:
The issue is being caused by this change made recently https://github.com/red-hat-storage/ocs-ci/pull/6059/files#diff-8cee422980dbcadd6a98fdb9c208fb16ff8e79bd064d549ebe9a3aece8f487feR770

```
       - titles = re.sub(r"\s{2,}", ",", resource)  # noqa: W605
       + titles = re.sub(r"\s", ",", re.sub(r"\s{2,}", ",", resource)) 
```
```
(Pdb) print(titles)
NAME,CAPACITY,ACCESS,MODES,RECLAIM,POLICY,STATUS,CLAIM,STORAGECLASS,REASON,AGE,pvc-04345643-00a2-48f8-9cb2-7b201f0e1d4d,10Gi,RWO,Retain,Bound,namespace-test-06f8ea0e6fee4a96ba51b5366/pvc-test-12b2db9d9b694792b7261fca6440188,storageclass-test-rbd-285b6008b008491f89,6m28s
```
It is causing `ACCESS MODE` and `RECLAIM POLICY` to be split into 4 items instead of 2, which is breaking column count further in the test.

PR #6504, opened 2022-09-23
Fix get resource for node status

petr-balogh:
Fixes: #6503

### DIFF
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -803,16 +803,14 @@ class OCP(object):
             selector=selector,
         )
         resource = re.split(r"\s{2,}", resource)
-
+        exception_list = ["RWO", "RWX", "ROX"]
         # get the list of titles
-        titles = [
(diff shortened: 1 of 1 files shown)


petr-balogh:
I don't like this function get_resource where we are trying to parse the text output and as we see we hit already several different issues with this function.

E.g. https://github.com/red-hat-storage/ocs-ci/issues/6138

Also some other issue introduced here:
https://github.com/red-hat-storage/ocs-ci/pull/6059/files#diff-8cee422980dbcadd6a98fdb9c208fb16ff8e79bd064d549ebe9a3aece8f487feR770

We should try to completely refactor this and if possible to try avoid text output and rather always process the yaml output.

The problem is that some values are not that easily accessible/visible from yaml structure, like the status column and maybe there are more, which we need to find out.

shylesh (approved):

clacroix12 (approved):
LGTM, but I agree about refactoring to utilize either yaml or json output. We can dump that data into a dictionary and utilize it much more consistently than the generic output assuming it contains all the info that we need.

PR #6511, opened 2022-09-26
removed obsolete file - lvm_utils.py

AviadP:
I cant see any use of this module in current ocs-ci repo
we currently support lvm via operator, so i would suggest to remove this file.

Signed-off-by: Aviadp <apolak@redhat.com>


dahorak (approved):

Avilir (approved):

shyRozen (commented):
Maybe move it and rename it instead of erasing the code

PR #6518, opened 2022-09-27
[DNM]: change ms_osd_pod_memory to 6Gi

vavuthu:


dahorak (approved):

fbalak (commented):
A new topology is currently only available with dev adddon. Would this change break testing of qe addon with original topology?

Issue #6561, opened 2022-10-13
rbd image metadata check fails 


rachael-george:
[test_rbd_image_metadata.py](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/manage/pv_services/test_rbd_image_metadata.py#L48) fails when getting metadata from the rbd image with the following error. 

```

2022-10-08 01:28:11,259 - MainThread - DEBUG - ocs_ci.utility.utils.exec_cmd.618 - Command stdout: There is 1 metadatum on this image:

Key                                 Value
rbd.csi.ceph.com/thick-provisioned  false

yaml.scanner.ScannerError: while scanning a simple key
  in "<unicode string>", line 3, column 1:
    Key                              ... 
    ^
could not find expected ':'
  in "<unicode string>", line 4, column 1:
    rbd.csi.ceph.com/thick-provision ... 
    ^
```
RUN ID: ocs-ci results for OCS4-10-Downstream-OCP4-10-IBM-Cloud-managed-1AZ-RHEL-3W-tier2 (BUILD ID: 4.10.6-6 RUN ID: 1665159532)
Report Portal [link](https://url.corp.redhat.com/31ec198)

PR #6562, opened 2022-10-13
Check metadata for rbd images created during the test only

rachael-george:
Fixes: #6561 

### DIFF
--- a/tests/manage/pv_services/test_rbd_image_metadata.py
+++ b/tests/manage/pv_services/test_rbd_image_metadata.py
@@ -1,7 +1,7 @@
 import logging
 
 from ocs_ci.ocs import constants
-from ocs_ci.helpers.helpers import create_unique_resource_name
+from ocs_ci.helpers.helpers import create_unique_resource_name, get_snapshot_content_obj
 from ocs_ci.ocs.resources.pod import get_ceph_tools_pod
 from ocs_ci.framework.pytest_customization.marks import tier2, polarion_id, bugzilla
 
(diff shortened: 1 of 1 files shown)


rachael-george:
The test case passed in the run above 
Console logs: https://url.corp.redhat.com/ab34883

ebenahar:
@rachael-george does the change keep backward compatibility of ODF and OCP?

rachael-george:
> 

I don't think backward compatibility should be an issue here. The setting of metadata was introduced in 4.11 in ceph-csi, which was then disabled as part of this bug https://bugzilla.redhat.com/show_bug.cgi?id=2099965, so there shouldn't be any metadata either way for regular rbd volumes.
The changes in this PR ensures that we check for metadata only for the RBD volumes created as part of the test and avoids checking metadata for all rbd images in the cluster which can have metadata, for eg: encrypted volumes.

@mashetty330 to keep me honest here

mashetty330:
> > 
> 
> I don't think backward compatibility should be an issue here. The setting of metadata was introduced in 4.11 in ceph-csi, which was then disabled as part of this bug https://bugzilla.redhat.com/show_bug.cgi?id=2099965, so there shouldn't be any metadata either way for regular rbd volumes. The changes in this PR ensures that we check for metadata only for the RBD volumes created as part of the test and avoids checking metadata for all rbd images in the cluster which can have metadata, for eg: encrypted volumes.
> 
> @mashetty330 to keep me honest here

correct @rachael-george

sidhant-agrawal (approved):
lgtm

vavuthu (approved):

Issue #6577, opened 2022-10-24
Wait for re-balance is showing health OK but 3 seconds after is Health Waning


petr-balogh:
https://url.corp.redhat.com/56fd990

Lately in acceptance add capacity test case we see it's failing with:
failed on teardown with "ocs_ci.ocs.exceptions.CephHealthException: Ceph cluster health is not OK. Health: HEALTH_WARN Reduced data availability: 4 pgs peering"

I found out that it completed the rebalance and after 3 seconds it's showing error above in teardown cluster health fixture.

PR #6578, opened 2022-10-24
Add repeat for wait_for_rebalance

petr-balogh:
Fixes: #6577

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -757,24 +757,39 @@ class CephCluster(object):
                 and states["count"] == total_pg_count
             )
 
-    def wait_for_rebalance(self, timeout=600):
+    def wait_for_rebalance(self, timeout=600, repeat=3):
         """
         Wait for re-balance to complete
 
         Args:
             timeout (int): Time to wait for the completion of re-balance
+            repeat (int): How many times to repeat the check to make sure, it's
+                really completed.
 
         Returns:
-            bool: True if rebalance completed, False otherwise
+            bool: True if re-balance completed, False otherwise
 
         """
         try:
-            for rebalance in TimeoutSampler(
-                timeout=timeout, sleep=10, func=self.get_rebalance_status
-            ):
-                if rebalance:
-                    logger.info("Re-balance is completed")
-                    return True
+            start_time = time.time()
+            for attempt in range(1, repeat + 1):
+                diff_time = int(time.time() - start_time)
+                logger.debug(f"Attempt {attempt} out of {repeat} repeats.")
+                for rebalance in TimeoutSampler(
+                    timeout=timeout - diff_time,
+                    sleep=10,
+                    func=self.get_rebalance_status,
+                ):
+                    if rebalance:
+                        logger.info(
+                            f"Attempt {attempt} for re-balance out of {repeat} is completed"
+                        )
+                        if repeat == attempt:
+                            return True
+                        else:
+                            logger.debug("Wait 10 seconds between next attempt")
+                            time.sleep(10)
+                            break
         except exceptions.TimeoutExpiredError:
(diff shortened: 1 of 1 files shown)


AviadP (approved):

dahorak (approved):

ebenahar (dismissed):

ocs_ci/ocs/cluster.py
```diff
@@ -757,24 +757,39 @@ def get_rebalance_status(self):
...
-            ):
-                if rebalance:
-                    logger.info("Re-balance is completed")
-                    return True
+            start_time = time.time()
+            for attempt in range(1, repeat + 1):
+                diff_time = int(time.time() - start_time)
+                logger.debug(f"Attempt {attempt} out of {repeat} repeats.")
+                for rebalance in TimeoutSampler(
+                    timeout=timeout - diff_time,
```
shylesh, line 779: @petr-balogh , I see that timeout value could become -ve at some point where diff_time grows larger during repeats. This could lead to quick successive attempts and TimeoutSampler exception. WDYT ?

ocs_ci/ocs/cluster.py
```diff
@@ -757,24 +757,39 @@ def get_rebalance_status(self):
...
+                    sleep=10,
+                    func=self.get_rebalance_status,
+                ):
+                    if rebalance:
+                        logger.info(
+                            f"Attempt {attempt} for re-balance out of {repeat} is completed"
```
shylesh, line 785: @petr-balogh ,This message needs more clarity that we are reattempting rebalance check even though previously we have seen clean status. Currently it sounds as if in the previous attempt cluster was still rebalancing where as that's not the case.

clacroix12 (approved):

ocs_ci/ocs/cluster.py
shylesh: @petr-balogh for ex:
let's say timeout=600 and in the first iteration diff_time will be very small because we started just now.
Hence timeout = ~600, assume that rebalance takes 600 seconds and finsihes successfully. since repeat =3 we will have 2 more attempts to be done even though cluster is clean at this point.

In the second iteration of outermost for loop where repeat=2, diff_time will be > 600 as we have already spent 600 seconds in the previous attempt, now timeout will be < 0 (some -ve value).  Its because of this line https://github.com/red-hat-storage/ocs-ci/blob/6ecc96ca640e96448a16e2271f5553a7a6248c12/ocs_ci/utility/utils.py#L938 , TimeoutSampler raises exception even before calling the `get_rebalance_status` and then `wait_for_rebalance` will fail even though there was no error.

Am I missing something?

petr-balogh: This is intentional @shylesh . As new behavior is that I would like to have the call of making really sure re-balance is complete don X many times as I have set the `repeat` value for in the mentioned timeout, in default value 600.

So in 600 seconds my function needs to prove rebalance completed 3 times.

So if it will take for first attempt more than 580 seconds, then it will not complete probably to do it 3 times and should fail which is intentional.

If user wants only one attempt, they need reduce repeat to 1.

ocs_ci/ocs/cluster.py
petr-balogh: I am mentioning that re banalance is completed and the number of the attempts and also out of how many I am going to perform. How you would like to re-phrase this?

petr-balogh:
@shylesh as current implementation fixes the test, and we have another issue with dependency which cannot be propagated to stable branch without this fix as add_capacity test is failing I am for merging as it is as I have already approve from Coady.

I made some improvements locally which might be satisfying you, but as I want to have new stable content promoted ASAP I will merge as is and send improvements in next PR if needed.

dahorak (approved):

Issue #6579, opened 2022-10-24
NFS tests make jenkins agent stuck.


petr-balogh:
As brought up in email with subject:
`NFS test and mount point in Jenkins agent makes job stuck`

Few points from the mail:

- Using sudo and mount directly on jenkins agent - or locally this will not work as we don't enable sudo for such commands without asking for a password.
- Mounting this to workspace folder of jenkins agent.  As in this case it didn't get unmounted and hence Wiping out workspace first was the problematic and got stuck. (Let's use /mnt folder for this case)
- Cannot we do this mounting test differently?  To avoid running sudo command locally I can imagine to run privileged RHEL pod on the cluster and try to run mount command there to avoid perfom this operation on system where ocs-ci executed.
- If the test fails in the middle the unmount even does not happen. This I think should be handled in the fixture.
- Do not use lazy option but really wait and make sure the mount point is unmounted.

For now we decided to disable those tests, and we will enable once the issue is fixed.

mbukatov:
> Using sudo and mount directly on jenkins agent - or locally this will not work as we don't enable sudo for such commands without asking for a password.

On the other hand, you can't avoid using sudo if you want to run a test with standard nfs client, as it runs in kernel space.

So the question we need to ask here is: Do we need to use `mount` command directly from outside of the cluster for these test cases? If the answer is yes, we won't avoid granting this test at least `CAP_SYS_ADMIN` on some machine.

> Cannot we do this mounting test differently? To avoid running sudo command locally I can imagine to run privileged RHEL pod on the cluster and try to run mount command there to avoid perfom this operation on system where ocs-ci executed.

Running this on the cluster may work, but if not done carefully, we could still mess with the cluster in a way which won't be immediately visible creating risk of failures for other tests (eg. wonder what would happen if the nfs server stops responding and one tries to run statfs on all mounted filesystems for monitoring purposes). That said depending on answer to the question above, we may try to use [openshift nfs client interface](https://docs.openshift.com/container-platform/4.8/storage/persistent_storage/persistent-storage-nfs.html). That would be easier to work with.

amr1ta:
Have opened https://github.com/red-hat-storage/ocs-ci/pull/6687 to fix this.

PR #6580, opened 2022-10-24
Disable NFS tests

petr-balogh:
Because of the issue: #6579

Disabling those tests which are using sudou mount on jenkins agent.


mbukatov (approved):

shylesh (approved):
LGTM

PR #6606, opened 2022-11-05
Add debug-mode entry_point

OdedViner:
This method allows you to develop individual functions without running pytest 

for example:
image


petr-balogh:
As discussed in email, I don't see general use case for this. And if you will not find any other supporter and update the code to be more general and be as general use case then I think we can close this.

PR #6599, opened 2022-11-03
Adjust the number of alertmanager pods

ebondare:
Signed-off-by: Elena Bondarenko <ebondare@redhat.com>

PR #6613, opened 2022-11-08
Check we have only one alertmanager pod instead of three

yitzhak12:
Check we have only one alertmanager pod instead of three.
See more details here: https://github.com/red-hat-storage/ocs-osd-deployer/pull/262#issue-1424994573.


yitzhak12:
Duplicate of https://github.com/red-hat-storage/ocs-ci/pull/6599.

PR #6616, opened 2022-11-09
Adding sleep to avoid flakiness in the MCG test

am-agrawa:
Signed-off-by: am-agrawa <amagrawa@redhat.com>

This test was repeatedly failing, so after debugging a couple of times, it seems the code is perfectly fine but it's flaky because sometimes the clicks are too fast and console is unable to match the timing. So as a temp. fix, I added a few sleeps and ran the PR validation which passed.

For the long term, we would need to design a function to handle such scenarios as using sleep is not recommended in UI automation.


keesturam (commented):
@aman can you please add a #todo to remove sleep once the code structure changes.

am-agrawa:
> @aman can you please add a #todo to remove sleep once the changes to function are made.

Created an issue- https://github.com/red-hat-storage/ocs-ci/issues/6618

keesturam (approved):

vavuthu (approved):

PR #6619, opened 2022-11-09
Test rolling terminate and recovery of the OCS worker nodes for the Managed Service and IPI deployments

yitzhak12:


tests/manage/z_cluster/nodes/test_rolling_terminate_and_recovery_ms.py
```diff
@@ -0,0 +1,83 @@
...
+            log.info("Waiting for all the worker nodes to be ready...")
+            wait_for_node_count_to_reach_status(
+                node_count=len(ocs_node_objs), timeout=900
+            )
+            log.info("Waiting for all the pods to be running")
+            assert check_pods_after_node_replacement(), "Not all the pods are running"
+            self.sanity_helpers.health_check_ms(cluster_check=False, tries=40)
+
+        # Check basic cluster functionality by creating some resources
+        self.sanity_helpers.create_resources_on_ms_consumers()
```
ebenahar, line 83: apart from these 2 functions, what else is different between this test case and a non MS environment? 
If no other differences, can you combine the 2 (MS and non MS) with `if` conditions?

yitzhak12: Currently, we don't have this test in a non-MS cluster. Do you want to apply this test also for a non-MS cluster? We need also to add `if` in the init function here: https://github.com/red-hat-storage/ocs-ci/pull/6619/files#diff-97907be6ea7ad97f58120078109a96b61fcc3f29c209bb30f017e9acf3137f2fR42.

ebenahar: yes, I think it's worth having this test cases applicable to both MS and non MS

yitzhak12: Okay, np.

ocs_ci/ocs/node.py
```diff
@@ -2556,6 +2556,13 @@ def consumers_verification_steps_after_provider_node_replacement():
         bool: True, if all the consumers verification steps finished successfully. False, otherwise.
 
     """
+    if version.get_semantic_ocs_version_from_config() >= version.VERSION_4_11:
+        log.info(
+            "We need to change the consumers verification steps when the ODF version is 4.11. "
+            "Currently, we can skip these steps when we use ODF 4.11"
```
ebenahar, line 2562: I think it's worth opening an issue to track this as we plan to adopt 4.11 in the near future.

ebenahar (approved):

ocs_ci/ocs/node.py
yitzhak12: Okay, I created an issue https://github.com/red-hat-storage/ocs-ci/issues/6649.

AviadP (approved):

PR #6635, opened 2022-11-15
Add arbiter deployment over 6 worker nodes in AWS

petr-balogh:

### DIFF
--- a/conf/deployment/aws/ipi_3az_rhcos_3m_6w_arbiter.yaml
+++ b/conf/deployment/aws/ipi_3az_rhcos_3m_6w_arbiter.yaml
@@ -1,0 +1,25 @@
+---
+#  This configuration file is used for Metro DR Arbiter cluster deployment
+#  TODO: Add polarion ID
+DEPLOYMENT:
+  arbiter_deployment: true
+  allow_lower_instance_requirements: false
+  type: 'EBS'
+  local_storage: true
+  ocs_operator_nodes_to_label: 6
+  arbiter_zone: 'us-east-2c'
+  arbiter_autodetect: false
+ENV_DATA:
+  platform: 'aws'
+  deployment_type: 'ipi'
+  region: 'us-east-2'
+  worker_availability_zones:
+    - 'us-east-2a'
+    - 'us-east-2b'
+  master_availability_zones:
+    - 'us-east-2a'
+    - 'us-east-2b'
+    - 'us-east-2c'
+  worker_replicas: 6
+  master_replicas: 3
+  worker_instance_type: 'm5.4xlarge'
--- a/conf/deployment/aws/ipi_3az_rhcos_3m_6w_arbiter.yaml
+++ b/conf/deployment/aws/ipi_3az_rhcos_3m_6w_arbiter.yaml
@@ -9,6 +9,8 @@ DEPLOYMENT:
   ocs_operator_nodes_to_label: 6
   arbiter_zone: 'us-east-2c'
   arbiter_autodetect: false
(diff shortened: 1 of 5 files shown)


conf/deployment/aws/ipi_3az_rhcos_3m_6w_arbiter.yaml
```diff
@@ -0,0 +1,25 @@
+---
+#  This configuration file is used for Metro DR Arbiter cluster deployment
+#  TODO: Add polarion ID
+DEPLOYMENT:
+  arbiter_deployment: true
+  allow_lower_instance_requirements: false
+  type: 'EBS'
+  local_storage: true
```
mbukatov, line 8: Don't we also need `local_storage_storagedeviceset_count: 3`?

conf/deployment/aws/ipi_3az_rhcos_3m_6w_arbiter.yaml
```diff
@@ -0,0 +1,25 @@
...
+  ocs_operator_nodes_to_label: 6
+  arbiter_zone: 'us-east-2c'
+  arbiter_autodetect: false
```
mbukatov, line 11: I would also add `network_split_setup: true` so that it's possible to try net splits if necessary.

conf/deployment/aws/ipi_3az_rhcos_3m_6w_arbiter.yaml
```diff
@@ -0,0 +1,25 @@
...
+  worker_availability_zones:
+    - 'us-east-2a'
+    - 'us-east-2b'
+  master_availability_zones:
+    - 'us-east-2a'
+    - 'us-east-2b'
+    - 'us-east-2c'
+  worker_replicas: 6
+  master_replicas: 3
+  worker_instance_type: 'm5.4xlarge'
```
mbukatov, line 25: I guess that we don't need `extra_disks: 2` here.

mbukatov (commented):
I did a quick comparison with `vsphere/upi_1az_rhcos_vsan_lso_vmdk_3m_6w_arbiter.yaml` trying to ignore platform based differences.

conf/deployment/aws/ipi_3az_rhcos_3m_6w_arbiter.yaml
petr-balogh: But actually I will use this value to add more than only one EBS volume in AWS node for this. So in next commit I will use extra_disks ;)

ocs_ci/utility/aws.py
```diff
@@ -1890,8 +1890,7 @@ def get_stack_name_from_instance_dict(instance_dict):
 
 
 def create_and_attach_ebs_volumes(
-    worker_pattern,
-    size=100,
+    worker_pattern, size=100, count=1, device_names=("sdx")
```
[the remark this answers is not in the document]
dahorak: Also I would maybe check, if the `device_names` is not only string and handle it differently (accept it when `count=1` or raise exception).

dahorak (approved):

ocs_ci/utility/aws.py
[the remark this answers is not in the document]
petr-balogh: > ("sdx",)

yeah, thanks for noticing. I had it as list but decided to put it as tuple as default value as it's more safe and forgot to add comma when I changed from [] to ().

petr-balogh:
Validation job running here:
https://url.corp.redhat.com/3207e17

vavuthu (approved):

PR #6639, opened 2022-11-16
Add configs for more VSphere IPI deployment types

clacroix12:


clacroix12:
@petr-balogh these configs don't include polarion_ids as I wasn't able to find any relevant polarion test cases for these scenarios. If we do have IDs for these deployment types I can add them to this PR.

vavuthu (dismissed):

conf/deployment/vsphere/ipi_1az_rhcos_lowerreq_vmfs_3m_3w.yaml
```diff
@@ -0,0 +1,13 @@
+---
```
petr-balogh, line 1: @vavuthu about VMFS, this is basically rely on which DC you run with, right? If I will run on DC ECO it will be VSAN, right?

vavuthu: yes. It will rely on whichDC we run, DC-ECO is vSAN

petr-balogh: @clacroix12 then we have two options, crate another copy of file for VSAN, or remove vmfs from name of the file.  As we will have the same polarion ID we can go with second option I guess.

clacroix12: Removed `vmfs` from the filename.

clacroix12:
@petr-balogh you mentioned that we would need to add multus configs as well. Could you clarify which deployment scenarios we need to cover with those configs? I can add them once I know which scenarios we want to cover.

petr-balogh (approved):

conf/deployment/vsphere/ipi_1az_rhcos_multus_vsan_3m_3w.yaml
```diff
@@ -0,0 +1,18 @@
+---
```
petr-balogh, line 1: Can we also add one more config for UPI please ?

clacroix12: [One already exists](https://github.com/red-hat-storage/ocs-ci/blob/master/conf/deployment/vsphere/upi_1az_rhcos_multus_vsan_3m_3w.yaml)

conf/deployment/vsphere/ipi_1az_rhcos_lowerreq_3m_3w.yaml
```diff
@@ -0,0 +1,16 @@
+---
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
```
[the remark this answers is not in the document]
clacroix12: I'm not sure as thats what we have it set to within the UPI counterpart [config](https://github.com/red-hat-storage/ocs-ci/blob/master/conf/deployment/vsphere/upi_1az_rhcos_lowerreq_vmfs_3m_3w.yaml)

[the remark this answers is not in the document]
clacroix12: @petr-balogh is setting this config value to true necessary for lowerreq configs? If so I think the UPI counterpart might be misconfigured.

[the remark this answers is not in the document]
petr-balogh: @jimdevops19 I remember that this requires UI deployment to be used and ODF does the trick itself. There is no way to mimic this behavior from CLI deployment the same way how ODF does that. @jimdevops19 might remember more about this.  If this is the case, not sure if we should contain the config value to enable UI deployment directly in this production configuration file. Our jenkins job has this option enabled by default for lowerreq jobs iirc.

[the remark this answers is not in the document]
petr-balogh: I think it's worth to add here the ui_deployment conf value like here:
https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/ui_deployment.yaml

Because this feature cannot be tested otherwise as we discussed offline.

petr-balogh (dismissed):

petr-balogh (approved):

shylesh (approved):
LGTM

PR #6643, opened 2022-11-16
WIP: Spot instances key change in machineset

jimdevops19:


petr-balogh:
As we discussed offline, this will not make the deployment with spot instances, you will need to generate manifests with installer and change the machine set there for worker nodes. This is not recommended for master nodes:
Some notes from our 1:1

https://github.com/red-hat-storage/ocs-ci/compare/master...jimdevops19:ocs-ci:fixed_spot_instances
https://github.com/openshift/installer/issues/5089
You will need to do something like this: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/vmware.py#L856 and edit worker nodes machineset

ebenahar (approved):

PR #6650, opened 2022-11-20
Split add capacity test on tier1 suite [ui] and acceptance test [cli]

OdedViner:
We are using `try exception`  method for add capacity test.
 https://github.com/red-hat-storage/ocs-ci/blob/master/tests/manage/z_cluster/cluster_expansion/test_add_capacity.py#L45-L51

When the add capacity process fails via UI, we perform the process via CLI.
In this situtaion, we cant catche bugs like this  https://bugzilla.redhat.com/show_bug.cgi?id=2137616 [there is an option to check the screenshots]

The developers change the locators frequtly, so the test fails even though there is no real bug. 

So I think the method is to split add capacity test to acceptance [cli] and tier1 [ui] 

### DIFF
--- a/ocs_ci/ocs/ui/add_replace_device_ui.py
+++ b/ocs_ci/ocs/ui/add_replace_device_ui.py
@@ -19,7 +19,7 @@ class AddReplaceDeviceUI(PageNavigator):
 
     def add_capacity_ui(self):
         """
-        Add Capacity via UI.
+        Add Capacity via UI
 
         """
         self.add_capacity_ui = locators[self.ocp_version]["add_capacity"]
--- a/ocs_ci/ocs/ui/add_replace_device_ui.py
+++ b/ocs_ci/ocs/ui/add_replace_device_ui.py
@@ -33,10 +33,6 @@ class AddReplaceDeviceUI(PageNavigator):
         self.do_click(self.add_capacity_ui["kebab_storage_cluster"])
         self.do_click(self.add_capacity_ui["add_capacity_button"])
         self.do_click(
-            self.add_capacity_ui["select_sc_add_capacity"], enable_screenshot=True
-        )
-        self.do_click(self.add_capacity_ui[self.storage_class], enable_screenshot=True)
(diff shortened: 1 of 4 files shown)


am-agrawa (approved):

ocs_ci/ocs/ui/add_replace_device_ui.py
```diff
@@ -31,11 +31,9 @@ def add_capacity_ui(self):
             self.do_click(self.add_capacity_ui["ocs_operator"])
             self.do_click(self.add_capacity_ui["storage_cluster_tab"])
         self.do_click(self.add_capacity_ui["kebab_storage_cluster"])
+        self.take_screenshot()
```
am-agrawa, line 34: It will be a good idea to simply print the name of the storage class visible on the UI, gp2 for AWS, thin for VMware, etc. Simply fetch the text.

ebenahar:
@OdedViner +1 on the proposed solution of keeping the CLI test in acceptance only and the UI one in tier1

ebenahar (approved):

am-agrawa (approved):

ocs_ci/ocs/ui/add_replace_device_ui.py
```diff
@@ -30,12 +31,11 @@ def add_capacity_ui(self):
         else:
             self.do_click(self.add_capacity_ui["ocs_operator"])
             self.do_click(self.add_capacity_ui["storage_cluster_tab"])
+        time.sleep(3)
```
am-agrawa, line 34: Pls use explicit wait

am-agrawa (dismissed):

ocs_ci/ocs/ui/add_replace_device_ui.py
OdedViner: It is problematic because the UI team changes the storage class name often.

am-agrawa (approved):

ebenahar (approved):

PR #6663, opened 2022-11-22
Ypersky test ceph capacity recover

ypersky1980:
This PR contains the following: 

1) A change to pod.yaml file so that fillup_fs function will be possible to use on both Performance and Non Performance pods. 

2) A new test tests/manage/pv_services/test_ceph_capacity_recovery.py  is added following https://bugzilla.redhat.com/show_bug.cgi?id=1910272 - a fix in ceph, which allows to delete pvs and corresponding pv when the cluster is in Read Only mode ( 85% capacity used). 

In the past this was not possible and stale PV were observed. 

The test flow is the following: 
- PVCs are created, partially filled with data. 
- Clones are created ( 1 clone per PVC) 
- some IO run on the system untill it reaches 85% used capacity. 
- PVCs are deleted one after another and Ceph Health and cluster capacity is monitored. 
- Ceph Capacity goes down and Ceph Health returns to OK ( Pass Criteria). 

3) A small change to tests/e2e/performance/csi_tests/test_pvc_bulk_creation_deletion_performance.py test in which the " creation after deletion time" is increased from 50 to 60 secs. The reason is that one of the test samples frequently measures slightly more than 50 secs. Therefore the test fails, while the average time is still good and does not require opening a bug. 
Therefore to my opinion this change will allow smoother Perf suite runs.


tests/manage/pv_services/test_ceph_capacity_recovery.py
```diff
@@ -0,0 +1,269 @@
...
+        status = ceph_health_check_base()
+        if status:
+            log.info("Health check passed")
+        else:
+            log.info("Health check failed")
+        return status
+    except CephHealthException as e:
+        # skip because ceph is not in good health
+        log.info(f"Ceph health exception received: {e}")
+        return False
```
Avilir, line 42: you can use the function : [ceph_health_check()](https://github.com/red-hat-storage/ocs-ci/blob/0e2db8e7846b08e1d621176fd1170b2e4b9fa3a2/ocs_ci/utility/utils.py#L2033) - this is code duplication

tests/manage/pv_services/test_ceph_capacity_recovery.py
```diff
@@ -0,0 +1,269 @@
...
+
+    Returns:
+         bool: True if used_capacity greater than expected_used_capacity, False otherwise
+
+    """
+    osds_utilization = get_osd_utilization()
+    log.info(f"osd utilization: {osds_utilization}")
+    for osd_id, osd_utilization in osds_utilization.items():
+        if osd_utilization > expected_used_capacity:
+            log.info(f"OSD ID:{osd_id}:{osd_utilization} greater than 85%")
```
Avilir, line 77: It is better to use a variable instead a number  (85), i assume  this is the  expected_used_capacity variable, you also can initialize it on line 62

tests/manage/pv_services/test_ceph_capacity_recovery.py
```diff
@@ -0,0 +1,269 @@
...
+            return True
+    return False
+
+
+class TestCephCapacityRecovery(PASTest):
```
Avilir, line 82: if this is a functional (Manage) test, why you are using the `PASTest` as base object and not `ManageTest` ?

tests/manage/pv_services/test_ceph_capacity_recovery.py
```diff
@@ -0,0 +1,269 @@
...
+        self.num_of_pvcs = 10
+        self.pvc_size = (
+            self.ceph_capacity * (1 - used_now / 100 - 0.15) / self.num_of_pvcs
+        )
+        self.pvc_size_str = str(self.pvc_size) + "Gi"
+        log.info(f"Creating pvs of {self.pvc_size_str} size")
+
+        pvc_list = []
+        pod_list = []
+        for i in range(5):
```
Avilir, line 150: dose this  need to be `range(self.num_of_pvcs)` "

ypersky1980: Yes, actually self.num_of_pvcs/2 , since on each loop cycel 1 pvc and 1 clone is created

tests/manage/pv_services/test_ceph_capacity_recovery.py
ypersky1980: Done

tests/manage/pv_services/test_ceph_capacity_recovery.py
ypersky1980: Sure, eventually I call to ceph_health_check_base library function, but I didn't want to call ceph_health_check since I wanted to enhance it with logging (and run the base only once , without tries and delays)

tests/manage/pv_services/test_ceph_capacity_recovery.py
ypersky1980: I wanted to  inherit from PasTest to pull the performance image and simplify the things

ramkiperiy: @ypersky1980 Perf image can be pulled without using PASTest.

tests/manage/pv_services/test_ceph_capacity_recovery.py
ramkiperiy: can we enhance the https://github.com/red-hat-storage/ocs-ci/blob/0e2db8e7846b08e1d621176fd1170b2e4b9fa3a2/ocs_ci/utility/utils.py#L2033 function itself to avoid duplication?

tests/manage/pv_services/test_ceph_capacity_recovery.py
```diff
@@ -0,0 +1,279 @@
...
+        else:
+            log.info("Health check failed")
+        return status
+    except CephHealthException as e:
+        # skip because ceph is not in good health
+        log.info(f"Ceph health exception received: {e}")
+        return False
+
+
+def get_used_capacity(msg):
```
ramkiperiy, line 48: This function can be moved to helpers library.

tests/manage/pv_services/test_ceph_capacity_recovery.py
```diff
@@ -0,0 +1,279 @@
...
+         bool: True if used_capacity greater than expected_used_capacity, False otherwise
+
+    """
+    log.info(f"{msg}")
+    used_capacity = get_percent_used_capacity()
+    log.info(f"Used Capacity is {used_capacity}%")
+    return used_capacity
+
+
+def verify_osd_used_capacity_greater_than_expected(expected_used_capacity):
```
ramkiperiy, line 65: Good to move this function to helpers, since it can be used for other tests as well

tests/manage/pv_services/test_ceph_capacity_recovery.py
ypersky1980: Thanks for your comment. I've moved it to the library (managed_services.py and not to helpers.py to avoid the circular dependency of imports).

tests/manage/pv_services/test_ceph_capacity_recovery.py
ypersky1980: Thanks! Similarly to the previous comment - I've moved it to the library (managed_services.py and not to helpers.py to avoid the circular dependency of imports).

petr-balogh (dismissed):

ramkiperiy (dismissed):

ocs_ci/helpers/managed_services.py
```diff
@@ -193,3 +194,42 @@ def verify_provider_topology():
...
+    """
+    log.info(f"{msg}")
+    used_capacity = ocs_ci.ocs.cluster.get_percent_used_capacity()
+    log.info(f"Used Capacity is {used_capacity}%")
+    return used_capacity
+
+
+def verify_osd_used_capacity_greater_than_expected(expected_used_capacity):
+    """
+    Verify OSD percent used capacity greate than ceph_full_ratio
```
jilju, line 218: Please fix typo "greate"

tests/manage/pv_services/test_ceph_capacity_recovery.py
```diff
@@ -0,0 +1,239 @@
...
+        else:
+            log.info("Health check failed")
+        return status
+    except CephHealthException as e:
+        # skip because ceph is not in good health
+        log.info(f"Ceph health exception received: {e}")
+        return False
+
+
+class TestCephCapacityRecovery(PASTest):
```
jilju, line 47: Please add skipif marker if the feature is available in a particular verision.

ocs_ci/helpers/managed_services.py
```diff
@@ -193,3 +194,42 @@ def verify_provider_topology():
...
             ), f"Pod {pod_name} is running on OSD running node {node_obj.name}"
     log.info("Verified that other pods are not running on OSD nodes")
+
+
+def get_used_capacity(msg):
+    """
+    Verify OSD percent used capacity greate than ceph_full_ratio
+
+    Args:
+        expected_used_capacity (float): expected used capacity
```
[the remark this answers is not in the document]
ypersky1980: Done. Thanks

ocs_ci/helpers/managed_services.py
```diff
@@ -193,3 +194,42 @@ def verify_provider_topology():
...
+
+    Returns:
+         bool: True if used_capacity greater than expected_used_capacity, False otherwise
```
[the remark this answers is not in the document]
ypersky1980: Done. Thanks!

ocs_ci/helpers/managed_services.py
ypersky1980: Fixed

tests/manage/pv_services/test_ceph_capacity_recovery.py
```diff
@@ -0,0 +1,239 @@
...
+
+        if self.benchmark_obj is not None:
+            self.benchmark_obj.cleanup()
+
+        # Deleting the namespace used by the test
+        self.delete_test_project()
+
+        super(TestCephCapacityRecovery, self).teardown()
+
+    def test_capacity_recovery(
```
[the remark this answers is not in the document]
ypersky1980: Also, added tier2 marker, similarly to the test_no_space_left.py.

tests/manage/pv_services/test_ceph_capacity_recovery.py
ypersky1980: The feature is available after fix of https://bugzilla.redhat.com/show_bug.cgi?id=1910272 ( fix in ceph). 
I 've tested it on ODF 4.12 and it worked. Adding a skipif marker for version < 4.12.

petr-balogh (approved):

jilju (approved):

ramkiperiy (approved):

pintojoy (approved):
LGTM

PR #6668, opened 2022-11-24
multus public interface config file for OCP 4.12

vavuthu:

### DIFF
--- a/conf/ocsci/multus_public_network_interface_412_vsphere.yaml
+++ b/conf/ocsci/multus_public_network_interface_412_vsphere.yaml
@@ -1,0 +1,5 @@
+---
+# Use this config for enabling multus deployment on vsphere for OCP 4.12
+ENV_DATA:
+  is_multus_enabled: true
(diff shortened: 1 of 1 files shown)


conf/ocsci/multus_public_network_interface_412_vsphere.yaml
```diff
@@ -0,0 +1,5 @@
+---
+# Use this config for enabling multus deployment on vsphere for OCP 4.12
+ENV_DATA:
+  is_multus_enabled: true
+  multus_public_network_interface: 'br-ex'
```
petr-balogh, line 5: Is this something we need to create the on the vSphere level manually in the specific DC before the deployment? Or this is automatically created?

petr-balogh (approved):

conf/ocsci/multus_public_network_interface_412_vsphere.yaml
vavuthu: it will be created automatically by openshift. we don't need to create anything

PR #6687, opened 2022-11-30
Updated nfs tests to fix failures

amr1ta:
Signed-off-by: Amrita Mahapatra <49347640+amr1ta@users.noreply.github.com>

### DIFF
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -10,8 +10,6 @@ and with consideration of the entire project.
 """
 
 import os
-from ocs_ci.framework import config
-
 
 # Logging
 LOG_FORMAT = "%(asctime)s - %(threadName)s - %(name)s - %(levelname)s - %(message)s"
@@ -1963,8 +1961,8 @@ ACM_CSV_LABEL = (
 )
 # Concatenated CA file for vcenter
 VSPHERE_CA_FILE_PATH = os.path.join(DATA_DIR, "vsphere_ca.crt")
-SSH_PRIV_KEY = config.DEPLOYMENT["ssh_key_private"]
-SSH_PUB_KEY = config.DEPLOYMENT["ssh_key"]
+SSH_PRIV_KEY = os.path.expanduser(os.path.join("~", ".ssh", "openshift-dev.pem"))
+SSH_PUB_KEY = os.path.expanduser(os.path.join("~", ".ssh", "openshift-dev.pub"))
 SPACE = " "
 
 # DR actions
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -1962,8 +1962,8 @@ ACM_CSV_LABEL = (
 )
 # Concatenated CA file for vcenter
 VSPHERE_CA_FILE_PATH = os.path.join(DATA_DIR, "vsphere_ca.crt")
-SSH_PRIV_KEY = os.path.expanduser(os.path.join(".ssh", "openshift-dev.pem"))
-SSH_PUB_KEY = os.path.expanduser(os.path.join(".ssh", "openshift-dev.pub"))
+SSH_PRIV_KEY = os.path.expanduser(os.path.join("~", ".ssh", "openshift-dev.pem"))
+SSH_PUB_KEY = os.path.expanduser(os.path.join("~", ".ssh", "openshift-dev.pub"))
 SPACE = " "
 
 # DR actions
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -1963,8 +1963,8 @@ ACM_CSV_LABEL = (
 )
 # Concatenated CA file for vcenter
 VSPHERE_CA_FILE_PATH = os.path.join(DATA_DIR, "vsphere_ca.crt")
-SSH_PRIV_KEY = config.DEPLOYMENT.get("ssh_key_private")
-SSH_PUB_KEY = config.DEPLOYMENT.get("ssh_key")
+SSH_PRIV_KEY = config.DEPLOYMENT["ssh_key_private"]
+SSH_PUB_KEY = config.DEPLOYMENT["ssh_key"]
 SPACE = " "
 
 # DR actions
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -10,6 +10,7 @@ and with consideration of the entire project.
 """
 
 import os
+from ocs_ci.framework import config
 
 
 # Logging
@@ -1962,8 +1963,8 @@ ACM_CSV_LABEL = (
 )
 # Concatenated CA file for vcenter
 VSPHERE_CA_FILE_PATH = os.path.join(DATA_DIR, "vsphere_ca.crt")
-SSH_PRIV_KEY = os.path.expanduser(os.path.join("~", ".ssh", "openshift-dev.pem"))
-SSH_PUB_KEY = os.path.expanduser(os.path.join("~", ".ssh", "openshift-dev.pub"))
+SSH_PRIV_KEY = config.DEPLOYMENT.get("ssh_key_private")
+SSH_PUB_KEY = config.DEPLOYMENT.get("ssh_key")
 SPACE = " "
 
 # DR actions
--- a/tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
+++ b/tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
@@ -2,6 +2,7 @@ import pytest
 import logging
 import time
 import yaml
+import os
 
 
 from ocs_ci.utility import nfs_utils
@@ -134,8 +135,9 @@ class TestNfsEnable(ManageTest):
         self.nfs_client_user = config.ENV_DATA.get("nfs_client_user")
         log.info(f"nfs_client_user is: {self.nfs_client_user}")
 
-        self.nfs_client_private_key = (
-            config.ENV_DATA.get("nfs_client_private_key") or constants.SSH_PRIV_KEY
+        self.nfs_client_private_key = os.path.expanduser(
+            config.ENV_DATA.get("nfs_client_private_key")
+            or config.DEPLOYMENT["ssh_key_private"]
         )
 
         self.con = None
--- a/tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
+++ b/tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
@@ -2,6 +2,7 @@ import pytest
 import logging
 import time
 import yaml
+import os
 
 
 from ocs_ci.utility import nfs_utils
@@ -90,10 +91,6 @@ class TestNfsEnable(ManageTest):
             private_key (str): Private key  to connect to load balancer
             password (password): Password for host
 
-        Prerequisites:
-            On the client host machine openshift-dev pub key in authorized_keys should be available
-            and nfs-utils package should be installed.
-
         Steps:
         ---Setup---
         1:- Create objects for storage_cluster, configmap, pod, pv, pvc, service and storageclass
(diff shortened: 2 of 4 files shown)


tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
```diff
@@ -121,6 +121,11 @@ def setup_teardown(self, request):
             self.storage_cluster_obj,
         )
 
+        # ssh to test-nfs-vm
+        log.info("Login to test vm")
+        self.con = Connection(
+            "10.0.151.68", user="root", private_key=constants.SSH_PRIV_KEY, stdout=False
```
petr-balogh, line 127: IIUC the 10.0.151.68 is some testing VM for this purpose.
If so, I think this value should be somehow configurable and not hardcoded here. If the value will not be provided, this test should be skipped.

I am thinking that someone out of the RH will try to run the test - then it will not be valid for them.

Also I would document somewhere in test what are the prerequisites which needs to be done/installed on this VM.

I can imagine that there needs to be uploaded openshift-dev pub key in authorized_keys ,  nfs-utils package needs to be installed and so on.

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
```diff
@@ -104,8 +105,7 @@ def setup_teardown(self, request):
         self.pv_obj = ocp.OCP(kind=constants.PV, namespace=self.namespace)
         self.nfs_sc = "ocs-storagecluster-ceph-nfs"
         self.sc = ocs.OCS(kind=constants.STORAGECLASS, metadata={"name": self.nfs_sc})
-        self.test_folder = "test_nfs"
-        utils.exec_cmd(cmd="mkdir -p " + self.test_folder)
+        self.test_folder = "/mnt/test_nfs"
```
petr-balogh, line 108: This is problematic. Imagine you have multiple parallel executions running, using the same VM.

Please expose the run_id into this path, something like:
```suggestion
        self.test_folder = f"/mnt/test_nfs_{run_id}"
```

petr-balogh (changes requested):

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
petr-balogh: so you can create custom_config.yaml which will look like:

```yaml
ENV_DATA:
    nfs_client_ip: "10.0.151.68"
```

then provide it to run-ci via `--ocsci-conf ./custom_config.yaml`

By default in our jenkins jobs we are creating one of our custom config file, so we can make sure this value is provided in all our automation runs in Jenkins.

But if someone will run locally, they will need to create such custom config file with proper VM IP. I guess not everyone will have access to this machine and we should describe how it should be setup.

If this VM IP is not in config, then you can skip the test with link to documentation that user needs to setup such VM and set this IP in config and provide via --ocsci-conf in order to run this test.

amr1ta: Addressed in latest commit. 
By default in our jenkins jobs we need to create one of our custom config file, so we can make sure
        ENV_DATA:
            nfs_client_ip: "10.0.151.68"
            nfs_client_user: "root"
            nfs_client_pwd: ""
these values are provided in all our automation runs in Jenkins.

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
```diff
@@ -106,8 +126,19 @@ def setup_teardown(self, request):
         self.pv_obj = ocp.OCP(kind=constants.PV, namespace=self.namespace)
         self.nfs_sc = "ocs-storagecluster-ceph-nfs"
         self.sc = ocs.OCS(kind=constants.STORAGECLASS, metadata={"name": self.nfs_sc})
-        self.test_folder = "test_nfs"
-        utils.exec_cmd(cmd="mkdir -p " + self.test_folder)
+        self.run_id = config.RUN.get("run_id")
+        self.test_folder = f"mnt/test_nfs_{self.run_id}"
+        log.info(f"nfs mount point out of cluster is----- {self.test_folder}")
+        self.nfs_client_ip = config.ENV_DATA["nfs_client_ip"]
```
petr-balogh, line 132: If the nfs_client_ip will not be defined it will fail here with KeyError exception.

You have to use 
```suggestion
        self.nfs_client_ip = config.ENV_DATA.get("nfs_client_ip")
```

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
```diff
@@ -106,8 +126,19 @@ def setup_teardown(self, request):
...
+        log.info(f"nfs_client_ip is: {self.nfs_client_ip}")
+
+        self.nfs_client_user = config.ENV_DATA["nfs_client_user"]
```
petr-balogh, line 135: Same here, use `.get(`

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
```diff
@@ -106,8 +126,19 @@ def setup_teardown(self, request):
...
+        log.info(f"nfs_client_user is: {self.nfs_client_user}")
+
+        self.nfs_client_pwd = config.ENV_DATA["nfs_client_pwd"]
```
petr-balogh, line 138: Same here, use `.get(`

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
```diff
@@ -106,8 +126,19 @@ def setup_teardown(self, request):
...
+        log.info(f"nfs_client_pwd is: {self.nfs_client_pwd}")
+
+        self.con = ""
```
petr-balogh, line 141:
```suggestion
        self.con = None
```
This looks better to me to use None instead of empty string


Or why not to init the connection directly here on one place instead of in each of test case?

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
```diff
@@ -267,6 +319,20 @@ def test_outcluster_nfs_export(
...
         """
+        if not self.nfs_client_ip:
+            pytest.skip(
+                "Skipped the test as a valid nfs client ip is required, "
+                " for nfs outcluster export validation. "
+            )
+
+        # ssh to test-nfs-vm
+        log.info("Login to test vm")
+        self.con = Connection(
```
petr-balogh, line 330: You are invoking this connection 4 times in this module, why not to have it in init of the class and re-use same connection in all places?

If the connection is not stable or there is some timeout, you can have slef.con define as property, and this property will check if connection exist, if so it will retun it from self._con e.g. and if it doesn't exist or it's expired or whatever it will create new connection and return new one.

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
```diff
@@ -441,6 +509,20 @@ def test_multiple_nfs_based_PVs(
         11:- Deletion of Pods and PVCs
 
         """
+        if not self.nfs_client_ip:
+            pytest.skip(
+                "Skipped the test as a valid nfs client ip is required, "
+                " for nfs outcluster export validation. "
```
petr-balogh, line 515: This is also repeated 4times, this can be defined in method and you can just call this method at the start of each test which require the connection to NFS client.

amr1ta: Addressed in latest commit.

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
amr1ta: Addressed in latest commit.

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
amr1ta: Addressed in latest commit.

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
amr1ta: Addressed in latest commit.

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
amr1ta: I have kept it like tha because I have incluster and out of cluster tests in the same test class and created the connection details for the out cluster tests. And  have added a check to skip the test in case valid ip address is not available for the out cluster mount validation tests.

ocs_ci/ocs/constants.py
```diff
@@ -1948,8 +1948,8 @@
 ACM_4_7_MULTICLUSTER_URL = "/multicloud/infrastructure/clusters/managed"
 # Concatenated CA file for vcenter
 VSPHERE_CA_FILE_PATH = os.path.join(DATA_DIR, "vsphere_ca.crt")
-SSH_PRIV_KEY = os.path.expanduser(os.path.join(".ssh", "openshift-dev.pem"))
-SSH_PUB_KEY = os.path.expanduser(os.path.join(".ssh", "openshift-dev.pub"))
+SSH_PRIV_KEY = os.path.expanduser(os.path.join("~", ".ssh", "openshift-dev.pem"))
```
petr-balogh, line 1951: I think we can rely on the config values we have here:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L62-L63

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
```diff
@@ -106,8 +125,19 @@ def setup_teardown(self, request):
...
+        self.test_folder = f"mnt/test_nfs_{self.run_id}"
+        log.info(f"nfs mount point out of cluster is----- {self.test_folder}")
+        self.nfs_client_ip = config.ENV_DATA.get("nfs_client_ip")
+        log.info(f"nfs_client_ip is: {self.nfs_client_ip}")
+
+        self.nfs_client_user = config.ENV_DATA.get("nfs_client_user")
+        log.info(f"nfs_client_user is: {self.nfs_client_user}")
+
+        self.nfs_client_pwd = config.ENV_DATA.get("nfs_client_pwd")
+        log.info(f"nfs_client_pwd is: {self.nfs_client_pwd}")
```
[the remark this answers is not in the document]
amr1ta: Updated in latest commit.

ocs_ci/ocs/constants.py
amr1ta: Updated in latest commit

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
```diff
@@ -107,8 +125,20 @@ def setup_teardown(self, request):
...
+        log.info(f"nfs mount point out of cluster is----- {self.test_folder}")
+        self.nfs_client_ip = config.ENV_DATA.get("nfs_client_ip")
+        log.info(f"nfs_client_ip is: {self.nfs_client_ip}")
+
+        self.nfs_client_user = config.ENV_DATA.get("nfs_client_user")
+        log.info(f"nfs_client_user is: {self.nfs_client_user}")
+
+        self.nfs_client_private_key = (
+            config.ENV_DATA.get("nfs_client_private_key") or constants.SSH_PRIV_KEY
+        )
```
petr-balogh, line 139:
```suggestion
        self.nfs_client_private_key = os.path.expanduser(
            config.ENV_DATA.get("nfs_client_private_key") or config.DEPLOYMENT["ssh_key_private"]
        )
```

amr1ta: Updated in latest commit

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
```diff
@@ -82,6 +83,24 @@ def setup_teardown(self, request):
         """
         Setup-Teardown for the class
 
+        Pre-Req:
+        By default in our jenkins jobs we are creating one of our custom config file,
+        so we can make sure
+        ENV_DATA:
+            nfs_client_ip: "10.0.151.68"
+            nfs_client_user: "root"
+            nfs_client_pwd: ""
```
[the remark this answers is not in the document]
amr1ta: Updated in latest commit

ocs_ci/ocs/constants.py
```diff
@@ -1962,8 +1963,8 @@
 )
 # Concatenated CA file for vcenter
 VSPHERE_CA_FILE_PATH = os.path.join(DATA_DIR, "vsphere_ca.crt")
-SSH_PRIV_KEY = os.path.expanduser(os.path.join(".ssh", "openshift-dev.pem"))
-SSH_PUB_KEY = os.path.expanduser(os.path.join(".ssh", "openshift-dev.pub"))
+SSH_PRIV_KEY = config.DEPLOYMENT["ssh_key_private"]
```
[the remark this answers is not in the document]
amr1ta: Updated in latest commit

tests/manage/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
amr1ta: By default in our jenkins jobs we need to create one of our custom config file, so we can make sure
ENV_DATA:
  nfs_client_ip: "10.0.151.68"
  nfs_client_user: "root"

these values are provided in all our automation runs in Jenkins.

petr-balogh (approved):

prsurve (approved):

Issue #6697, opened 2022-12-01
vsphere-cleanup ContextualVersionConflict google-api-core requires requires google-auth<3.0dev,>=2.14.1


clacroix12:
Hitting the following version conflict when running `vsphere-cleanup`:
```
2022-12-01 08:28:35  + set +e
2022-12-01 08:28:35  + vsphere-cleanup --cluster-name j-001vu1ce33-t2 --ocsci-conf=/tmp/5JSRBE6T0Y2I
2022-12-01 08:28:36  Traceback (most recent call last):
2022-12-01 08:28:36    File "/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/.venv/lib64/python3.8/site-packages/pkg_resources/__init__.py", line 583, in _build_master
2022-12-01 08:28:36      ws.require(__requires__)
2022-12-01 08:28:36    File "/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/.venv/lib64/python3.8/site-packages/pkg_resources/__init__.py", line 900, in require
2022-12-01 08:28:36      needed = self.resolve(parse_requirements(requirements))
2022-12-01 08:28:36    File "/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/.venv/lib64/python3.8/site-packages/pkg_resources/__init__.py", line 791, in resolve
2022-12-01 08:28:36      raise VersionConflict(dist, req).with_context(dependent_req)
2022-12-01 08:28:36  pkg_resources.ContextualVersionConflict: (google-auth 1.35.0 (/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/.venv/lib/python3.8/site-packages), Requirement.parse('google-auth<3.0dev,>=2.14.1'), {'google-api-core'})
2022-12-01 08:28:36  
2022-12-01 08:28:36  During handling of the above exception, another exception occurred:
2022-12-01 08:28:36  
2022-12-01 08:28:36  Traceback (most recent call last):
2022-12-01 08:28:36    File "/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/.venv/bin/vsphere-cleanup", line 6, in <module>
2022-12-01 08:28:36      from pkg_resources import load_entry_point
[...]

PR #6698, opened 2022-12-01
Update google package versions

clacroix12:
Fixes: #6697


clacroix12:
Rebuilt the failed job in the description with my branch and it successfully passed teardown:
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-destroy-ocs-cluster/21696/console

clacroix12:
@petr-balogh not sure if we are actually even using `google-cloud-storage` but I updated the version here as it was restricting what version of google-auth we could install. The only reference I see to `google-cloud-storage` is [here](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/bucket_utils.py#L556) in bucket_utils and we aren't even importing it, just creating a google-cloud-storage backingstore.

As the bug we are trying to fix with this change is important I'm not going to address this here but I think we might benefit from a review of our setup.py requirements list so we can remove any dependencies we might not need anymore.

petr-balogh (approved):

shylesh (approved):

PR #6708, opened 2022-12-05
Reduce memory consumption at utils.exec_cmd()

DanielOsypenko:
# Fixes: 6489

Changes provided with this PR called up to reduce memory consumption.
Assumed that utils.py:exec_cmd() produces a memory leakage.
utils.py:exec_cmd()  function is used explicitly all over the framework leaving, often in a loop cycle,
hence we have to optimize it to avoid OSError: [Errno 12] Cannot allocate memory error.

Changes were tested with loops ~1000 calls on a local machine and overall results show that peaks in memory usage both rss,vsz have been reduced. Required more tests on Jenkins slave before merging. 

AC will be that tests do not produce OSError: [Errno 12] Cannot allocate memory error.

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -447,6 +447,7 @@ def run_cmd(
     ignore_error=False,
     threading_lock=None,
     silent=False,
+    stdin=None,
     **kwargs,
 ):
     """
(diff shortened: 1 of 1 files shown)


ocs_ci/utility/utils.py
```diff
@@ -601,14 +616,16 @@ def exec_cmd(
     masked_cmd = mask_secrets(cmd, secrets)
     log.info(f"Executing command: {masked_cmd}")
     if isinstance(cmd, str):
-        cmd = shlex.split(cmd)
+        # use of shlex has been temporarily removed to fix issue #6489
+        # shlex.split() creates peaks in RSS+VSZ mem usage when called intensively
+        cmd = cmd.split()
```
petr-balogh, line 621: Shlex and split provides different outputs which will cause the regression in our code if we will change it here. It's very dangerous change here.

PR #6798, opened 2022-12-14
Added BlockPools tabs count verification to cover BZ #2096513

DanielOsypenko:
Fixes BZ #2096513

Signed-off-by: Daniel Osypenko <dosypenk@redhat.com>

### DIFF
--- a/ocs_ci/ocs/ui/base_ui.py
+++ b/ocs_ci/ocs/ui/base_ui.py
@@ -207,6 +207,18 @@ class BaseUI:
         """
         return self.driver.find_element(by=locator[1], value=locator[0]).text
 
+    def get_elements(self, locator):
+        """
+        Get an elements list. Useful to count number of elements presented on page, etc.
+
+        Args:
+            locator (set): (GUI element needs to operate on (str), type (By)).
+
+        Return:
+            list: The list of WebElements
+        """
+        return self.driver.find_elements(by=locator[1], value=locator[0])
+
     def page_has_loaded(self, retries=5, sleep_time=1):
         """
         Waits for page to completely load by comparing current page hash values.
--- a/ocs_ci/ocs/ui/validation_ui.py
+++ b/ocs_ci/ocs/ui/validation_ui.py
@@ -291,6 +291,11 @@ class ValidationUI(PageNavigator):
                     self.validation_loc["storage-system-status-card-hyperlink"],
                     enable_screenshot=True,
                 )
+            # verify that only one BlockPools tab is present. BZ #2096513
+            blockpools_tabs = self.get_elements(self.validation_loc["blockpools"])
+            if len(blockpools_tabs) > 1:
+                logger.critical("Multiple BlockPools tabs were found. BZ #2096513")
+                warnings.warn("Multiple BlockPools tabs were found. BZ #2096513")
             logger.info("Click on StorageSystems breadcrumb")
             self.do_click((self.validation_loc["storagesystems"]))
             logger.info("Navigate back to ODF Overview page")


OdedViner (approved):

ocs_ci/ocs/ui/base_ui.py
```diff
@@ -207,6 +207,18 @@ def get_element_text(self, locator):
...
+        """
+        Get an elements list. Useful to count number of elements presented on page, etc.
+
+        Args:
+            locator (set): (GUI element needs to operate on (str), type (By)).
+
+        Return:
+            list: The list of WebElements
+        """
+        return self.driver.find_elements(by=locator[1], value=locator[0])
```
am-agrawa, line 220: Can you pls fetch text of matching elements found and print as Info?

DanielOsypenko: If I understood correctly, that you want to print text of every elements, there are few reasons why not to do that: 
1) The elements not always have a text, e.g. navigation btns, icons, infographics, meaning you will see lots of empty lines
2) the rest general functions in base_ui.py such as get_element_text(), check_element_text() don't have print of the text, it will over-inflate logs and not so useful in case when we don't need the text
3) this function is intended to see a number of elements on the page, make assertion e.g.  len(webElementsList) == 0; iterate the list of elements, take a random element from the list, etc. If this will be a crucial logic of the test is
4) it may be so the list of elements has dozens of elements.. too much for printing, and in the same time you always can print their text implicitly, while iterating the list from the test/procedure

am-agrawa: > If I understood correctly, that you want to print text of every elements, there are few reasons why not to do that:
> 
> 1. The elements not always have a text, e.g. navigation btns, icons, infographics, meaning you will see lots of empty lines
> 2. the rest general functions in base_ui.py such as get_element_text(), check_element_text() don't have print of the text, it will over-inflate logs and not so useful in case when we don't need the text
> 3. this function is intended to see a number of elements on the page, make assertion e.g.  len(webElementsList) == 0; iterate the list of elements, take a random element from the list, etc. If this will be a crucial logic of the test is
> 4. it may be so the list of elements has dozens of elements.. too much for printing, and in the same time you always can print their text implicitly, while iterating the list from the test/procedure

Then why not use the existing function from base_ui?
`def check_element_text(self, expected_text, element="*"):`

ocs_ci/ocs/ui/validation_ui.py
```diff
@@ -291,6 +291,11 @@ def odf_overview_ui(
                     self.validation_loc["storage-system-status-card-hyperlink"],
                     enable_screenshot=True,
                 )
+            # verify that only one BlockPools tab is present. BZ #2096513
+            blockpools_tabs = self.get_elements(self.validation_loc["blockpools"])
+            if len(blockpools_tabs) > 1:
+                logger.critical("Multiple BlockPools tabs were found. BZ #2096513")
```
am-agrawa, line 297: Shouldn't we fail the test here?

DanielOsypenko: having warnings.warn next line the test will proceed until the end and fail in the end with proper output

ocs_ci/ocs/ui/base_ui.py
DanielOsypenko: because it is different:
self.driver.find_element != self.driver.find_elements

am-agrawa: > because it is different: self.driver.find_element != self.driver.find_elements

Pls check the function again
```
element_list = self.driver.find_elements_by_xpath(
            f"//{element}[contains(text(), '{expected_text}')]"
        )
        return len(element_list) > 0
```

ocs_ci/ocs/ui/validation_ui.py
am-agrawa: I don't think it's necessary to proceed if the test is failing at this point

ocs_ci/ocs/ui/base_ui.py
DanielOsypenko: please, specify what should I check again?

am-agrawa: Updated above comment

DanielOsypenko: I see this function, this is good function, but not what the test requires. We need to have a common routine to find element list (not verify that the number of elements is larger than 0). The func description is in a docstring.

am-agrawa (approved):

vavuthu (approved):

PR #6808, opened 2022-12-16
Ui automate bz2103975

mashetty330:
Validate if unprivileged user without any roles assigned can access to ODF dashboard under Storage in OCP console
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2103975


tests/ui/test_other_user.py
```diff
@@ -51,3 +59,21 @@ def test_create_storageclass_rbd(self, user_factory, login_factory):
         """Login using created user"""
         obc_ui_obj = ObcUi(login_factory(user[0], user[1]))
         obc_ui_obj.check_obc_option()
+
```
avd-sagare, line 62: Change the filename to like test_nonadmin_user_odfaccess

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -256,7 +256,7 @@ EXTERNAL_MODE:
 UI_SELENIUM:
   browser_type: "chrome"
   chrome_type: "google-chrome"
-  headless: True
```
OdedViner, line 259: Please change the `headless` value to `True`

ocs_ci/ocs/ui/views.py
```diff
@@ -1094,6 +1094,14 @@
         "//a[normalize-space()='noobaa.io~v1alpha1~BucketClass']",
         By.XPATH,
     ),
+    "Developer_dropdown": (
```
OdedViner, line 1097: Can you use lowercase on key? [ `Developer_dropdown` ->` `developer_dropdown`]

ocs_ci/ocs/ui/validation_ui.py
```diff
@@ -565,3 +565,18 @@ def check_capacity_breakdown(self, project_name, pod_name):
...
+
+    def validate_unprivileged_access(self):
+        """
+        Function to verify the unprivileged users can't access ODF dashbaord
+        """
+        self.do_click(self.validation_loc["developer_dropdown"])
+        self.do_click(self.validation_loc["select_administrator"], timeout=5)
+        try:
+            self.navigate_odf_overview_page()
+        except Exception:
```
am-agrawa, line 577: Can we raise any specific exception for this case?

tests/ui/test_non_admin_user.py
```diff
@@ -51,3 +59,21 @@ def test_create_storageclass_rbd(self, user_factory, login_factory):
...
         obc_ui_obj.check_obc_option()
+
+
+class TestUnprivilegedUserODFAccess(E2ETest):
+    """
+    Test if unprivileged user can see ODF dashboard
+    """
+
+    @ui
+    @tier2
```
am-agrawa, line 70: For adding it to tier2, let's seek clarity from Bipin Kunal or Harish

hnallurv: Since this is a security concern, IMO, better to add it in tier1 so we detect such loopholes much early

ocs_ci/ocs/ui/validation_ui.py
mashetty330: Here I expect `TimeoutException` from do_click() method inside self.navigate_odf_overview_page() but whats weird is if I use below piece of code to have specific exception handling

>         except TimeoutException:
            logger.info(
                "As expected, ODF dashboard is not available for the unprivileged user"
            )

I see that exception is not caught even though raised exception is TimeoutException. It only works when I keep the exception handling broader by using just Exception clause

OdedViner (approved):

tests/ui/test_non_admin_user.py
```diff
@@ -3,8 +3,17 @@
...
 from ocs_ci.ocs.ui.mcg_ui import ObcUi
+from ocs_ci.ocs.ui.validation_ui import ValidationUI
 from ocs_ci.ocs import ocp
-from ocs_ci.framework.testlib import ManageTest, ui, bugzilla, polarion_id, tier2
+from ocs_ci.framework.testlib import (
+    ManageTest,
+    ui,
+    bugzilla,
+    polarion_id,
+    tier2,
```
am-agrawa, line 13: Now it shows both tier 1 and 2

ocs_ci/ocs/ui/validation_ui.py
```diff
@@ -565,3 +566,18 @@ def check_capacity_breakdown(self, project_name, pod_name):
...
+
+    def validate_unprivileged_access(self):
+        """
+        Function to verify the unprivileged users can't access ODF dashbaord
+        """
+        self.do_click(self.validation_loc["developer_dropdown"])
+        self.do_click(self.validation_loc["select_administrator"], timeout=5)
+        try:
+            self.navigate_odf_overview_page()
+        except TimeoutException as ex:
```
am-agrawa, line 578: Pls create a custom exception for broader understanding

mashetty330: I feel custom exception is unnecessary if its not reusable, thats why `logger.info`

ocs_ci/ocs/ui/validation_ui.py
```diff
@@ -565,3 +566,18 @@ def check_capacity_breakdown(self, project_name, pod_name):
...
+            logger.info(
+                f"As expected, ODF dashboard is not available for the unprivileged user: {ex}"
+            )
+        else:
+            assert False, "Unexpected, unprivileged users can access ODF dashboard"
```
am-agrawa, line 583: if there is no exception in try block, else block will execute, which means navigation under try block worked, but then else block gets executed where you are asserting false, so the test will always fail in this scenario.

tests/ui/test_non_admin_user.py
mashetty330: @am-agrawa tier2 is being used by other test.

ocs_ci/ocs/ui/validation_ui.py
mashetty330: Yes correct that is what expected. Only criteria for test success is try block throws exception.

mashetty330: what I can do is basically, instead of assert here... like you suggested raise a custom exception which is then caught in the test function and that can be used to assert the test.

am-agrawa (approved):

am-agrawa:
Pls run the PR validation jobs for all versions from 4.9 to 4.12.

OdedViner (approved):

PrasadDesala (approved):

PR #6835, opened 2022-12-22
Updating provider deployment for using subnet_ids parmeter from config

suchita-g:


suchita-g:
Already handled in another merged PR

Issue #6838, opened 2022-12-22
test_namespacestore_creation_webhook[Empty secret name] fails to detect invalid secret error


sagihirshfeld:
The attempt at creating a resource without providing a valid secret name fails as expected, but the test fails in recognized the new error message and continues to raise it:

`Error is Error from server: error when creating "/tmp/NamespaceStorex19xuygu": admission webhook "admissionwebhook.noobaa.io" denied the request: Failed creating the NamespaceStore, please provide a valid ARN or secret name`

RP links:
- https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/6551/269495/269500/269503/log
- https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/6759/279024/279029/279037/log
- https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/7469/325315/325322/325328/log
- https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/all/6269/254781/254786/254790/log

PR #6858, opened 2022-12-27
Fix the expected error message when providing an empty secret name at TestAdmissionWebhooks::test_namespacestore_creation_webhook[Empty secret name]

sagihirshfeld:
Fixes: #6838 

Signed-off-by: Sagi Hirshfeld <shirshfe@redhat.com>

### DIFF
--- a/tests/manage/mcg/test_admission_control.py
+++ b/tests/manage/mcg/test_admission_control.py
@@ -155,7 +155,7 @@ class TestAdmissionWebhooks(MCGTest):
                             "secret": {"name": ""},
                         },
                     },
-                    "please provide a valid ARN or secret name",
+                    "secret name",
                 ],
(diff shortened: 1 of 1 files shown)


sagihirshfeld:
When trying to create a NamepsaceStore with an empty secret, the resulting error depends on the version:

- In ODF versions <= 4.11: 
`... Failed creating the namespacestore, please provide secret name`

- In ODF 4.12 it was modified to resemble the error we would get while trying to create a BackingStore in the same way:
`... Failed creating the NamespaceStore, please provide a valid ARN or secret name`

I changed the substring we are looking for in the error message to `secret name` in order to detect the error in both cases. The change was made in the PyTeset param arg to avoid specific what-if logic in the test itself

keesturam (approved):

mashetty330 (approved):

PR #6875, opened 2023-01-03
test object expiration with expiration period + buffer time

mashetty330:
https://bugzilla.redhat.com/show_bug.cgi?id=2039309


tests/manage/mcg/test_object_expiration.py
```diff
@@ -81,3 +81,81 @@ def test_object_expiration(self, mcg_obj, bucket_factory):
         assert s3_get_object(
             s3_obj=mcg_obj, bucketname=bucket, object_key=object_key
         ), "Failed: Get Object"
+
+    @tier1
```
PrasadDesala, line 85: As discussed, we will have this Test run separately in every release. This cannot be part of regression testing suite as it will take more than a day to complete. 
replace tier-1 marker with system_test and also move this test under system test so that we dont miss track of this test

PrasadDesala (approved):

parikshithb (approved):

PR #6499, opened 2022-09-22
Bug 2110504 List large number of objects with prefix and delimeter

mashetty330:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=2110504 

Fixes #6876 

Signed-off-by: mashetty <mashetty@redhat.com>

PR #6877, opened 2023-01-03
Fix backingstore deletion when we have huge number of objects to be deleted

### LINKED ISSUE #6876: Backingstore deletion shouldn't timeout too quick when there is huge number of objects to be deleted
When we have huge number of objects (> 1 Mil) the backingstore deletion must fail with the prompt "cannot complete because objects in Backingstore {backingstore_name} are still being deleted, Please try later". In such scenarios, teardown code for backingstore deletion is failing too quickly (~120sec) 

Here is the failed run: https://url.corp.redhat.com/722473d

mashetty330:
Fixes #6876 

### DIFF
--- a/ocs_ci/ocs/resources/backingstore.py
+++ b/ocs_ci/ocs/resources/backingstore.py
@@ -123,18 +123,41 @@ class BackingStore:
             "cli": _cli_deletion_flow,
         }
 
-        if retry:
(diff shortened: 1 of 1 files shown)


mashetty330:
covering here https://github.com/red-hat-storage/ocs-ci/pull/6499

Issue #6837, opened 2022-12-22
HTTPIterator has no len


sagihirshfeld:
`Message: failed on teardown with "TypeError: object of type 'HTTPIterator' has no len()"` on  ocs-ci/ocs_ci/ocs/resources/cloud_uls.py::uls_cleanup, apparently when trying to clean up GCP resources

Affected tests:
1. test_unidirectional_bucket_object_change_replication[AWStoAZURE-BS-OC]
2. test_unidirectional_bucketclass_replication[AZUREtoAWS-BC-CLI]
3. test_check_object_integrity[AWS-OC-Cache]

RP links:
1. https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/6942/289052/289099/289104/log
2. https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/6942/289052/289148/289150/log
3. https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/6949/289361/289413/289419/log
4. https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/6949/289361/289464/289471/log
5. https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/6959/290051/290105/290109/log
6. https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/6959/290051/290154/290160/log
7. https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/6976/290813/290865/290871/log
8. https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/6976/290813/290914/290922/log
9. https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/6998/291593/291640/291646/log
10. https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/7015/292448/292491/292494/log
11. https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/7015/292448/292531/292538/log
[...]

PR #6888, opened 2023-01-05
Fix the deletion of the underlying storage of GCP BackingStores and minor refactor regarding CloudClient.delete_uls

sagihirshfeld:
Fixes: #6837 

`GCPBucket.delete_blobs` in `GoogleClient.delete_uls` could not parse the blobs in HTTPIterator which is the default output of `GCPBucket.list_blobs`, causing an `HTTPIterator has no len` error. This issue is resolved by passing these blobs via a Python List instead.

In addition to solving this issue, this PR includes a minor refactor regarding the way that CloudClients and it's inheritors S3Client, GoogleClient, and AzureClient are deleting ULS':

The TimeoutSampler logic that was used in `S3Client.delete_uls` has been elevated to the abstract parent class and the abstract method `CloudClient.internal_delete_uls`, which is implemented by each inheritor, is the function that is used by the TimeoutSampler.

### DIFF
--- a/ocs_ci/ocs/resources/cloud_manager.py
+++ b/ocs_ci/ocs/resources/cloud_manager.py
@@ -5,7 +5,7 @@ from abc import ABC, abstractmethod
 
 import boto3
 import google.api_core.exceptions as GoogleExceptions
-from azure.core.exceptions import ResourceNotFoundError
+from azure.core.exceptions import ResourceNotFoundError, AzureError
 from azure.storage.blob import BlobServiceClient
 from botocore.exceptions import ClientError, EndpointConnectionError
 from google.auth.exceptions import DefaultCredentialsError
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/resources/cloud_manager.py
```diff
@@ -125,8 +124,25 @@ def delete_uls(self, name):
         the appropriate implementation
 
         """
+        if self.verify_uls_exists(name) is False:
+            logger.warning(
+                f"Underlying Storage {name} does not exist, and was not deleted."
```
keesturam, line 129: Could you please rephrase the warning to maybe `Underlying Storage {name} does not exist, hence it can't be deleted`

keesturam (approved):
LGTM

parikshithb (approved):

Issue #6909, opened 2023-01-12
The finalizer of user_factory does not delete the user at the end of the test


OdedViner:
[ocs_ci/utility/users.py](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/users.py#L142)

PR #6918, opened 2023-01-15
Fix user factory teardown

OdedViner:
#6909
Based on this doc:
https://docs.openshift.com/container-platform/4.11/authentication/identity_providers/configuring-htpasswd-identity-provider.html

### DIFF
--- a/ocs_ci/utility/users.py
+++ b/ocs_ci/utility/users.py
@@ -90,7 +90,7 @@ def delete_identity(user_name=None):
 
     """
     if user_name is None:
-        cmd = f"oc delete identity --all"
+        cmd = "oc delete identity --all"
     else:
         cmd = f"oc delete identity my_htpasswd_provider:{user_name}"
     try:
--- a/ocs_ci/utility/users.py
+++ b/ocs_ci/utility/users.py
@@ -6,6 +6,7 @@ from tempfile import NamedTemporaryFile
 
 from ocs_ci.ocs import constants, ocp
 from ocs_ci.utility.utils import exec_cmd
+from ocs_ci.ocs.exceptions import CommandFailed
 
 log = logging.getLogger(__name__)
 
@@ -77,8 +78,8 @@ def delete_user(user_name=None):
         cmd = f"oc delete user {user_name}"
     try:
         exec_cmd(cmd)
-    except Exception as e:
-        logging.error(e)
+    except CommandFailed as e:
+        log.error(e)
 
 
 def delete_identity(user_name=None):
@@ -95,8 +96,8 @@ def delete_identity(user_name=None):
         cmd = f"oc delete identity my_htpasswd_provider:{user_name}"
     try:
         exec_cmd(cmd)
-    except Exception as e:
-        logging.error(e)
+    except CommandFailed as e:
+        log.error(e)
 
 
 def create_htpasswd_idp():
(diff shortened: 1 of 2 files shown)


ocs_ci/utility/users.py
```diff
@@ -64,6 +64,41 @@ def delete_htpasswd_secret():
...
+
+    """
+    if user_name is None:
+        cmd = "oc delete identity --all"
+    else:
+        cmd = f"oc delete identity my_htpasswd_provider:{user_name}"
+    try:
+        exec_cmd(cmd)
+    except Exception as e:
+        logging.error(e)
```
fbalak, line 99: Tox fails because logging is used directly (https://github.com/red-hat-storage/ocs-ci/actions/runs/3924241160/jobs/6708369701).

ocs_ci/utility/users.py
```diff
@@ -64,6 +64,41 @@ def delete_htpasswd_secret():
...
+def delete_identity(user_name=None):
+    """
+    Delete identity
+
+    Args:
+        user_name (str): the user to delete
+
+    """
+    if user_name is None:
+        cmd = "oc delete identity --all"
```
fbalak, line 93: I am not sure if deleting all identities is a good default behaviour when no parameter provided. Especially when it is not documented in docstring.

ocs_ci/utility/users.py
```diff
@@ -64,6 +64,41 @@ def delete_htpasswd_secret():
...
 
+def delete_user(user_name=None):
+    """
+    Delete user
+
+    Args:
+        user_name (str): the user to delete
+    """
+    if user_name is None:
+        cmd = "oc delete user --all"
```
fbalak, line 75: Please update the docstring to reflect that all users will be deleted of no user name is provided or divide the function into two: `delete_user`  with mandatory parameter and `delete_all_users`.

ocs_ci/utility/users.py
```diff
@@ -144,13 +179,9 @@ def _finalizer():
         Delete all users created by the factory
 
         """
-        with open(htpasswd_path) as f:
-            htpasswd = f.readlines()
-        new_htpasswd = [line for line in htpasswd if not line.startswith(tuple(_users))]
-        with open(htpasswd_path, "w+") as f:
-            for line in new_htpasswd:
-                f.write(line)
-        create_htpasswd_secret(htpasswd_path, replace=True)
```
fbalak, line 153: I think that the secret needs to be updated too. According to documentation the step of updating the secret is in place and following commands to delete user and identity takes place only when the secret is updated.

fbalak (changes requested):

ocs_ci/utility/users.py
fbalak: https://access.redhat.com/documentation/en-us/openshift_container_platform/4.11/html-single/authentication_and_authorization/index#identity-provider-htpasswd-update-users_configuring-htpasswd-identity-provider

ocs_ci/utility/users.py
OdedViner: done

ocs_ci/utility/users.py
OdedViner: done

ocs_ci/utility/users.py
OdedViner: done

ocs_ci/utility/users.py
OdedViner: sorry... I did not get you...

fbalak: You are removing a step that updates the secret. It needs to stay to have users fully deleted - it is one of the steps from the documentation that is linked (step 3). You should not remove these lines.

Issue #6903, opened 2023-01-11
scale: test_s3_benchmark_object_bucket failed with ResourceLeftoversException on IBM Z


tuan-hoang1:
Hi,

For this test, I see that the actual test was running good : Object Bucket was created, benchmark run was executed, new OBC later was also created and passed. But ocs-ci seems to report left over PV ?

```
tests/e2e/scale/noobaa/test_hsbench.py::TestHsBench::test_s3_benchmark_object_bucket ERROR

=================================================================================================================== ERRORS ===================================================================================================================
______________________________________________________________________________________ ERROR at teardown of TestHsBench.test_s3_benchmark_object_bucket ______________________________________________________________________________________
exclude_labels = ['must-gather', 's3cli']

    def get_status_after_execution(exclude_labels=None):
        """
        Set the environment status and assign it into ENV_STATUS_PRE dictionary.
        In addition compare the dict before the execution and after using DeepDiff
    
        Args:
            exclude_labels (list): App labels to ignore leftovers
    
        Raises:
             ResourceLeftoversException: In case there are leftovers in the
                environment after the execution
        """
        get_environment_status(config.RUN["ENV_STATUS_POST"], exclude_labels=exclude_labels)
    
        pod_diff = compare_dicts(
            config.RUN["ENV_STATUS_PRE"]["pod"], config.RUN["ENV_STATUS_POST"]["pod"]
        )
        sc_diff = compare_dicts(
            config.RUN["ENV_STATUS_PRE"]["sc"], config.RUN["ENV_STATUS_POST"]["sc"]
        )
        pv_diff = compare_dicts(
            config.RUN["ENV_STATUS_PRE"]["pv"], config.RUN["ENV_STATUS_POST"]["pv"]
        )
        pvc_diff = compare_dicts(
            config.RUN["ENV_STATUS_PRE"]["pvc"], config.RUN["ENV_STATUS_POST"]["pvc"]
        )
        namespace_diff = compare_dicts(
            config.RUN["ENV_STATUS_PRE"]["namespace"],
            config.RUN["ENV_STATUS_POST"]["namespace"],
        )
        volumesnapshot_diff = compare_dicts(
            config.RUN["ENV_STATUS_PRE"]["vs"], config.RUN["ENV_STATUS_POST"]["vs"]
        )
        if config.RUN["cephcluster"]:
            cephfs_diff = compare_dicts(
                config.RUN["ENV_STATUS_PRE"]["cephfs"],
[...]

PR #6932, opened 2023-01-18
Fixed test_s3_benchmark_object_bucket resource leftover and test case…

tiffanyn108:
Fixing issue # [6903](https://github.com/red-hat-storage/ocs-ci/issues/6903)
Fixing test_s3_benchmark_object_bucket resource leftover and test case failure.

### DIFF
--- a/ocs_ci/ocs/hsbench.py
+++ b/ocs_ci/ocs/hsbench.py
@@ -2,6 +2,7 @@ import logging
 import os
 import csv
 import filecmp
+import time
 
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.helpers import helpers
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/hsbench.py
```diff
@@ -357,6 +358,7 @@ def validate_reshard_process(self):
             CommandFailed: If reshard process fails
 
         """
+        bucket_name = self.bucket_prefix + "000000000000"
```
keesturam, line 361: Why should we add ` "000000000000"`?

tiffanyn108: This is default naming convention which created by hsbench with user defined prefix.  In this test case, prefix is set as "bp01" and "000000000000" will be the first bucket created.

tiffanyn108: This test case is for a single bucket creation, will address this as enhancement in separate a PR. 
Enhance PR is tracked here [PR # 6947](https://github.com/red-hat-storage/ocs-ci/issues/6974)

keesturam (approved):

Issue #5463, opened 2022-02-14
[IBM Z] Test fails as bash executable missing in container


abdulkandathil:
Test `tests/manage/mcg/test_bucket_replication.py::TestReplication::test_unidirectional_bucket_object_change_replication[AWStoAZURE-BS-OC]` fails on IBM Z as bash executable missing in the container.

```
cmd = ['oc', '-n', 'openshift-storage', 'rsh', 'session-awscli-relay-pod-8ec657f065dd4a5', 'bash', ...]
secrets = None, timeout = 600, ignore_error = False, kwargs = {}
masked_cmd = 'oc -n openshift-storage rsh session-awscli-relay-pod-8ec657f065dd4a5 bash -c "md5sum test_unidirectional_bucket_object_change_replication[AWStoAZURE-BS-OC]/origin/ObjKey-0"'
completed_process = CompletedProcess(args=['oc', '-n', 'openshift-storage', 'rsh', 'session-awscli-relay-pod-8ec657f065dd4a5', 'bash', '-c...ontainer process caused: exec: \\"bash\\": executable file not found in $PATH"\ncommand terminated with exit code 1\n')
masked_stdout = ''
masked_stderr = 'time="2022-02-10T13:58:13Z" level=error msg="exec failed: container_linux.go:380: starting container process caused: exec: \\"bash\\": executable file not found in $PATH"\ncommand terminated with exit code 1\n'

```

petr-balogh:
@bernerhat @MeridianExplorer can you please take a look guys?

abdulkandathil:
We are still facing this error on s390x.

sagihirshfeld:
Hi @abdulkandathil, could you please check PR #6954 and confirm that the fix also works on IBM Z?

abdulkandathil:
@sagihirshfeld The test is now passing with your fix.. thank you.

PR #6954, opened 2023-01-19
Change pod.py::cal_md5sum to use oc exec instead of oc rsh

sagihirshfeld:
Fixes: #5463

This change allows running md5sum regardless of which type of shell is installed on the pod, as long as it's available in the desired path.

This should be compatible with all platforms, but testing on platforms other than vSphere is still needed to confirm.

Signed-off-by: Sagi Hirshfeld <shirshfe@redhat.com>


mashetty330 (approved):

keesturam (approved):
LGTM

Issue #6881, opened 2023-01-03
Run MS tests strategy to cover all the test cases


yitzhak12:
Some tier4 tests(tier4a, tier4b, and maybe tier4c) are applicable for provider and consumer clusters, like `test_nodes_restart` or `test_rolling_shutdown_and_recovery_in_controlled_fashion`. 
That means if we specify that the consumer cluster is first, those tests will run only on the consumer cluster and not on the provider. 

Moreover, if we switch to the provider cluster in some tests and the consumer cluster in other tests, in that case, we must also check that the Sanity helper function is configured correctly on the current running cluster.
For example, we have a run with one consumer and one provider. And the default cluster is the consumer. Let’s take the test `test_osd_node_restart_and_check_osd_pods_status` for this example.
In the setup function, it inits the Sanity helper for the consumer. When we switch to the provider cluster in the test, it checks the provider functionality, which can lead to validating the wrong cluster details.

In summary, if we want to cover all cases in tier4 we have two options:

1. Run one time when the consumer cluster is first(as we currently do) and another when the provider cluster is first.
2. Switch to the relevant cluster(consumer or provider) in every test by demand and init the sanity helper in the function. We can use something like [here](https://drive.google.com/file/d/1xYkJ0UriK1KGvO4zrKaKmOjDSxh7t23Y/view?usp=share_link)

I already discussed it with @ebenahar, and he prefers option 2(with a small addition).

yitzhak12:
Please Let me know WDYT

ebenahar:
+1 on option 2

nehaberry:
@jilju can you also share your input

jilju:
I'd prefer option 2.

yitzhak12:
Yes. @ebenahar and I also thought about it. But we still need to init the sanity function in the test itself and not in the `setup` method, right?

fbalak:
> Yes. @ebenahar and I also thought about it. But we still need to init the sanity function in the test itself and not in the `setup` method, right?

It's true that it would be problematic. Also the context would need to be used correctly in fixtures.

dahorak:
I don't have strong opinion here, because it will depends on how many such tests we have and how many tests we have which are tight to only one role (provider/consumer) and how the numbers are spread across the various testsuites etc.

[...]

PR #7037, opened 2023-02-02
Switch to the correct cluster index at setup

yitzhak12:
fix https://github.com/red-hat-storage/ocs-ci/issues/6881.
The purpose of the pr is to fix the problem described in the issue above.

In the pr, I implemented a functionality to switch to the correct cluster index at the test setup.
I had three goals in the implementation:

1. Stability - means that we can be sure that in every test, it switches to the correct cluster index in the setup method. That's why I preferred creating a function `check_switch_to_correct_cluster_at_setup`, rather than a fixture or decorator.
2. Clean - means that we have a minimum change in the tests. That's why the change is implemented only in the `setup` method and not in the test methods.
3. Easy to analyze - we can quickly look at the test runs and see if the cluster was a provider, consumer, or non-MS cluster type. I used the pytest `parametrize` with the `cluster_type` param.

I also added the test file `tests/managed-service/test_switch_to_correct_index_at_setup.py` to demonstrate the functionality mentioned above.

### DIFF
--- a/ocs_ci/framework/__init__.py
+++ b/ocs_ci/framework/__init__.py
@@ -366,28 +366,28 @@ class MultiClusterConfig:
         cluster_types = [cluster.ENV_DATA["cluster_type"] for cluster in self.clusters]
         return cluster_type in cluster_types
 
-    def get_cluster_type_indexes_list(self, cluster_type):
+    def get_cluster_type_indices_list(self, cluster_type):
         """
-        Get the cluster type indexes
+        Get the cluster type indices
 
         Returns:
-            list: the cluster type indexes
+            list: the cluster type indices
 
         Raises:
             ClusterNotFoundException: In case it didn't find any cluster with the cluster type
 
         """
-        cluster_type_indexes_list = []
+        cluster_type_indices_list = []
         for i, cluster in enumerate(self.clusters):
             if cluster.ENV_DATA["cluster_type"] == cluster_type:
-                cluster_type_indexes_list.append(i)
+                cluster_type_indices_list.append(i)
 
-        if not cluster_type_indexes_list:
+        if not cluster_type_indices_list:
             raise ClusterNotFoundException(
                 f"Didn't find any cluster with the cluster type '{cluster_type}'"
             )
 
-        return cluster_type_indexes_list
+        return cluster_type_indices_list
 
     def switch_to_cluster_by_cluster_type(self, cluster_type, num_of_cluster=0):
         """
@@ -403,7 +403,7 @@ class MultiClusterConfig:
 
         """
         self.switch_ctx(
-            self.get_cluster_type_indexes_list(cluster_type)[num_of_cluster]
+            self.get_cluster_type_indices_list(cluster_type)[num_of_cluster]
         )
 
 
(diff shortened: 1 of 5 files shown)


ocs_ci/framework/__init__.py
```diff
@@ -352,5 +352,59 @@ def is_consumer_exist(self):
...
+        Returns:
+            list: the cluster type indexes
+
+        Raises:
+            ClusterNotFoundException: In case it didn't find any cluster with the cluster type
+
+        """
+        cluster_type_indexes_list = []
+        for i, cluster in enumerate(self.clusters):
+            if cluster.ENV_DATA["cluster_type"] == cluster_type:
```
petr-balogh, line 382: Should we rather use get here?

yitzhak12: Thanks for revising the pr. From what I know, the param `cluster_type` should always exist in the run. And if not provided, the default value is empty string `''`.

ocs_ci/framework/__init__.py
```diff
@@ -352,5 +352,59 @@ def is_consumer_exist(self):
...
+            cluster_type (str): The cluster type
+            num_of_cluster (int): The cluster index to switch to. The default cluster number
+                is 0 - which means it will switch to the first cluster.
+                1 - is the second, 2 - is the third, and so on.
+        Raises:
+            ClusterNotFoundException: In case it didn't find any cluster with the cluster type
+
+        """
+        self.switch_ctx(
+            self.get_cluster_type_indexes_list(cluster_type)[num_of_cluster]
```
ebenahar, line 406: IIUC, `get_cluster_type_indexes_list()` returns a list of indices in a form of integers, e.g. `[2,4,5]`. 
But here you are switching context to the index in the list of integers.
So based on the above example of `[2,4,5]`, if `num_of_cluster` is 2, you will switch to cluster `5` and not `2` as `5` is the 3rd index in the list

yitzhak12: Yes, that should be the behavior—the same logic as the existing method `switch_to_consumer`: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/__init__.py#L296. Maybe the docstring is not clear enough.

ocs_ci/utility/utils.py
```diff
@@ -4104,3 +4105,60 @@ def string_chunkify(cstring, csize):
...
+            )
+            return
+        else:
+            pytest.skip(
+                f"The test will not run on a non-MS cluster with the cluster type '{cluster_type}'"
+            )
+
+    # If the cluster is an MS cluster
+    if not config.is_cluster_type_exist(cluster_type):
+        pytest.skip(f"The cluster type '{cluster_type}' is not exist in the run")
```
dahorak, line 4160:
```suggestion
        pytest.skip(f"The cluster type '{cluster_type}' does not exist in the run")
```

dahorak (approved):

jilju (approved):

yitzhak12:
The pr validation passed with provider and consumer clusters. The test results are https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/1508/testReport/tests.managed-service.test_switch_to_correct_index_at_setup/TestSwitchToCorrectIndexAtSetup/. You can see that all the tests with the `provider` or `consumer` params passed, and the tests with the `non_ms` param skipped.

jilju (approved):

suchita-g (approved):

dahorak (approved):

ebenahar (approved):

PR #7038, opened 2023-02-02
Update project version to 4.12

clacroix12:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ except ImportError:
 
 setup(
     name="ocs-ci",
-    version="4.6.0",
+    version="4.12.0",
     description="OCS CI tests that run in jenkins and standalone mode using aws provider",
     author="OCS QE",
     author_email="ocs-ci@redhat.com",


vavuthu (approved):

setup.py
```diff
@@ -9,7 +9,7 @@
 
 setup(
     name="ocs-ci",
-    version="4.6.0",
+    version="4.12.0",
```
vavuthu, line 12: isn't it 4.13 ?

vavuthu, line 12: isn't it 4.13 ?

petr-balogh: No, we want to have this in release 4.12 branch. 4.13 we will put in your pr to default to 4.13 Vijay

petr-balogh (approved):

setup.py
vavuthu: > No, we want to have this in release 4.12 branch. 4.13 we will put in your pr to default to 4.13 Vijay

got it

PR #7043, opened 2023-02-05
Fix test nodes restart for MS

yitzhak12:
The pr fixes the following in the nodes restart tests for MS:

- In the test `test_osd_node_restart_and_check_osd_pods_status`, check that the osd pods are running when we have more than three worker nodes instead of checking the osd pods on the same node.
- In the test `test_nodes_restart`, restart only one node for the provider, so we don't lose quorum, and pick the node/s randomly.
- Delete the test `test_rolling_provider_worker_nodes_restart` and include it in the existing test `test_rolling_nodes_restart`.

### DIFF
--- a/tests/manage/z_cluster/nodes/test_nodes_restart_ms.py
+++ b/tests/manage/z_cluster/nodes/test_nodes_restart_ms.py
@@ -32,6 +32,7 @@ from ocs_ci.helpers.sanity_helpers import SanityManagedService
 from ocs_ci.ocs.cluster import (
     ceph_health_check,
     is_ms_consumer_cluster,
+    is_ms_provider_cluster,
 )
 from ocs_ci.framework import config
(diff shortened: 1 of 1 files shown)


yitzhak12:
I ran two pr validation jobs:

- on the tier4a test "test_nodes_restart": https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/1324/console
-  on tier4b test "test_rolling_nodes_restart": https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/1325/console. 

Both passed successfully.

OdedViner (approved):

am-agrawa (approved):

fbalak (approved):

PR #7060, opened 2023-02-07
Non ocs taint and tolerations for NooBaa in standalone mode

vkathole:
Coverage for https://bugzilla.redhat.com/show_bug.cgi?id=2115613
Signed-off-by: vkathole <vkathole@redhat.com>


tests/manage/mcg/test_non_ocs_taint_and_toleration.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        Make sure all nodes are untainted
+
+        """
+
+        def finalizer():
+            assert untaint_nodes(
+                taint_label="xyz=true:NoSchedule",
+            ), "Failed to untaint"
+
```
sagihirshfeld, line 57: To fully revert to the previous state we should also remove the tolerations from the storagecluster and the subscription. Afterward, we should also respin the pods so that the new configuration will propagate to the pods.

mashetty330 (commented):
Does this test pass on non mcg standalone cluster?

tests/manage/mcg/test_non_ocs_taint_and_toleration.py
```diff
@@ -0,0 +1,123 @@
...
+from ocs_ci.helpers.sanity_helpers import Sanity
+
+logger = logging.getLogger(__name__)
+
+
+@ignore_leftovers
+@skipif_tainted_nodes
+@skipif_managed_service
+@bugzilla("2115613")
+class TestNonOCSTaintAndTolerations(E2ETest):
```
PrasadDesala, line 33: we already have test_non_ocs_taint_and_toleration.py  which covers all these steps. 
Can you check if you adjust the existing test to handle mcg-only by adding the relevant conditions?

PrasadDesala (approved):

sagihirshfeld (approved):

prsurve (approved):

PR #7092, opened 2023-02-09
add avoid_stale to do_click

### LINKED ISSUE #7090: test_add_capacity_ui has reference to stale element
Test test_add_capacity_ui fails on reference to stale element, see bellow.

```
Message: selenium.common.exceptions.StaleElementReferenceException: Message: stale element reference: element is not attached to the page document
  (Session info: headless chrome=109.0.5414.119)
Type: None

Text:
self = <tests.manage.z_cluster.cluster_expansion.test_add_capacity.TestAddCapacity object at 0x7f6faaebd5b0>
reduce_and_resume_cluster_load = None 
```

Add argument to do_click function **avoid_stale: bool**  to quickly resolve such issues.

* p.s. We cannot cover all do_click calls with avoid stale exception, since refresh page may distract test case, loosing data or linking to another webpage. Adding avoid_stale should be done explicitly.

DanielOsypenko:
fixes #7090

added parameter to ocs_ci/ocs/ui/base_ui.py::do_click() to refresh page if page is stale
add_capacity_ui now uses avoid_stale=True

### DIFF
--- a/ocs_ci/ocs/ui/base_ui.py
+++ b/ocs_ci/ocs/ui/base_ui.py
@@ -94,8 +94,8 @@ class BaseUI:
         enable_screenshot (bool): take screenshot
         copy_dom (bool): copy page source of the webpage
         avoid_stale (bool): if got StaleElementReferenceException, caused by reference to stale, cached element,
-            refresh the page once and try click again
-            * don't use when refreshed page expected to be different from initial page, or loose input values
+        refresh the page once and try click again
+        * don't use when refreshed page expected to be different from initial page, or loose input values
         """
 
         def _do_click(_locator, _timeout=30, _enable_screenshot=False, _copy_dom=False):
--- a/ocs_ci/ocs/ui/base_ui.py
+++ b/ocs_ci/ocs/ui/base_ui.py
@@ -100,7 +100,9 @@ class BaseUI:
 
         def _do_click(_locator, _timeout=30, _enable_screenshot=False, _copy_dom=False):
             self.page_has_loaded()
-            screenshot = ocsci_config.UI_SELENIUM.get("screenshot") and enable_screenshot
+            screenshot = (
+                ocsci_config.UI_SELENIUM.get("screenshot") and enable_screenshot
+            )
             if screenshot:
                 self.take_screenshot()
             if _copy_dom:
@@ -127,6 +129,7 @@ class BaseUI:
                 raise TimeoutException(
                     f"Failed to find the element ({locator[1]},{locator[0]})"
                 )
+
         try:
             _do_click(locator, timeout, enable_screenshot, copy_dom)
         except StaleElementReferenceException:
--- a/ocs_ci/ocs/ui/base_ui.py
+++ b/ocs_ci/ocs/ui/base_ui.py
@@ -78,7 +78,14 @@ class BaseUI:
             Path(self.dom_folder).mkdir(parents=True, exist_ok=True)
         logger.info(f"screenshots pictures:{self.dom_folder}")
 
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/ui/base_ui.py
```diff
@@ -78,44 +78,67 @@ def __init__(self, driver: WebDriver):
...
+        * don't use when refreshed page expected to be different from initial page, or loose input values
         """
-        if (
-            version.get_semantic_version(get_ocp_version(), True)
-            <= version.VERSION_4_11
-        ):
-            try:
-                wait = WebDriverWait(self.driver, timeout)
+
+        def _do_click(_locator, _timeout=30, _enable_screenshot=False, _copy_dom=False):
```
am-agrawa, line 101: Why referencing them as private?

ocs_ci/ocs/ui/base_ui.py
```diff
@@ -78,44 +78,67 @@ def __init__(self, driver: WebDriver):
...
         """
         Click on Button/link on OpenShift Console
 
         locator (set): (GUI element needs to operate on (str), type (By))
         timeout (int): Looks for a web element repeatedly until timeout (sec) happens.
         enable_screenshot (bool): take screenshot
         copy_dom (bool): copy page source of the webpage
+        avoid_stale (bool): if got StaleElementReferenceException, caused by reference to stale, cached element,
+        refresh the page once and try click again
+        * don't use when refreshed page expected to be different from initial page, or loose input values
```
am-agrawa, line 98: I did not get the meaning of this line from doc text, where are we checking that the refreshed page is same or different from the initial page? Isn't by default it's expected to be the same page?

ocs_ci/ocs/ui/base_ui.py
```diff
@@ -78,44 +78,67 @@ def __init__(self, driver: WebDriver):
...
+
+        try:
+            _do_click(locator, timeout, enable_screenshot, copy_dom)
+        except StaleElementReferenceException:
+            if avoid_stale:
+                logger.info("Refresh page to avoid reference to a stale element")
+                self.driver.refresh()
+                _do_click(locator, timeout, enable_screenshot, copy_dom)
+            else:
+                raise
```
am-agrawa, line 141: Any specific exception to be raised? Currently which exception is being raised here?

ocs_ci/ocs/ui/base_ui.py
DanielOsypenko: there are cases when you proceed with filling forms and refresh will clear your inputs. Or PopUp appears, it may disappear after refresh, or when we interacting with Topology and selecting the tab of chosen node, for instance Observe, it will be shown a default Resources after refresh. So shortly it depends on UX logic and must be validated by tester when he writes the test

ocs_ci/ocs/ui/base_ui.py
DanielOsypenko: raise itself will raise the same exception we got. This way we do refresh page and try _do_click once more, but this time do not pass avoid_stale and don't go into a new recursion and raise initial exception

ocs_ci/ocs/ui/base_ui.py
DanielOsypenko: this is internal function unavailable for use. To make this block work without changing name of the function I had to rename initial func.
"""
The underscore prefix is meant as a hint to another programmer that a variable or method starting with a single underscore is intended for internal use
"""

OdedViner (approved):

am-agrawa (approved):

dahorak (approved):

PR #7118, opened 2023-02-15
Update the test test_automated_recovery_from_failed_nodes_reactive_ms

jilju:
Update the test test_automated_recovery_from_failed_nodes_reactive_ms to run on  the clusters with the topology changes in Managed Services.
The check for intermediate state of the OSD pods is removed because the OSD pod can spin up on another node in the same zone on clusters with more than 3 worker node. Instead the expected state and the distribution of the OSD pods are verified.

### DIFF
--- a/tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_reactive_ms.py
+++ b/tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_reactive_ms.py
@@ -10,6 +10,7 @@ from ocs_ci.framework.testlib import (
 )
 
 from ocs_ci.framework import config
+from ocs_ci.helpers.managed_services import verify_osd_distribution_on_provider
 from ocs_ci.ocs import machine, constants
 from ocs_ci.ocs.cluster import is_ms_provider_cluster
 from ocs_ci.ocs.resources.pod import (
@@ -17,9 +18,9 @@ from ocs_ci.ocs.resources.pod import (
     check_pods_after_node_replacement,
     get_osd_pods_having_ids,
     delete_pods,
-    wait_for_osd_pods_having_ids,
 )
 from ocs_ci.utility.utils import ceph_health_check
+from ocs_ci.utility.retry import retry
 
 from ocs_ci.ocs.node import (
     get_osd_running_nodes,
@@ -31,7 +32,6 @@ from ocs_ci.ocs.node import (
     recover_node_to_ready_state,
     get_node_rook_ceph_pod_names,
     verify_worker_nodes_security_groups,
-    wait_for_osd_ids_come_up_on_node,
     unschedule_nodes,
     schedule_nodes,
     wait_for_new_worker_node_ipi,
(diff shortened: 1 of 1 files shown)


jilju:
Verified: http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/jm021mse2e7-c1/jm021mse2e7-c1_20230213T080906/multicluster/logs/test_report_1676470460.html

fbalak (approved):

tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_reactive_ms.py
```diff
@@ -109,10 +109,19 @@ def check_automated_recovery_from_stopped_node(nodes):
...
+    # Verify that OSD pods are running on the node
+    osd_ids_on_node = get_node_osd_ids(osd_node_name)
+    assert osd_ids_on_node, f"No OSD pod is running on the node {osd_node_name}."
+    log.info(f"OSD {osd_ids_on_node} are running on the node {osd_node_name}.")
+
+    # The OSD IDs that were present on the stopped should be in running state now
+    osds_with_old_id = get_osd_pods_having_ids(old_osd_pod_ids)
+    assert len(set(osds_with_old_id)) == len(
+        set(old_osd_pod_ids)
+    ), f"One or more of the OSD IDs {old_osd_pod_ids} which were running on the stopped node are not present now"
```
yitzhak12, line 123: This code can be replaced with `wait_for_osd_pods_having_ids(osd_ids=old_osd_pod_ids)`.

tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_reactive_ms.py
```diff
@@ -151,10 +160,19 @@ def check_automated_recovery_from_terminated_node(nodes):
...
+    # Verify that OSD pods are running on the new node
+    osd_ids_on_node = get_node_osd_ids(new_wnode.name)
+    assert osd_ids_on_node, f"No OSD pod is running on the node {new_wnode.name}."
+    log.info(f"OSD {osd_ids_on_node} are running on the node {new_wnode.name}.")
+
+    # The OSD IDs that were present on the terminated should be in running state now
+    osds_with_old_id = get_osd_pods_having_ids(old_osd_pod_ids)
+    assert len(set(osds_with_old_id)) == len(
+        set(old_osd_pod_ids)
+    ), f"One or more of the OSD IDs {old_osd_pod_ids} which were running on the terminated node are not present now"
```
yitzhak12, line 174: Same here.

tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_reactive_ms.py
```diff
@@ -151,64 +159,58 @@ def check_automated_recovery_from_terminated_node(nodes):
...
+    # Verify that OSD pods are running on the node
+    osd_ids_on_node = get_node_osd_ids(osd_node_name)
+    assert osd_ids_on_node, f"No OSD pod is running on the node {osd_node_name}."
+    log.info(f"OSD {osd_ids_on_node} are running on the node {osd_node_name}.")
+
+    # The OSD IDs that were present on the node should be in running state now
+    osds_with_old_id = get_osd_pods_having_ids(old_osd_pod_ids)
+    assert len(set(osds_with_old_id)) == len(
+        set(old_osd_pod_ids)
+    ), f"One or more of the OSD IDs {old_osd_pod_ids} which were running on the node are not present now"
```
yitzhak12, line 213: Same here.

tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_reactive_ms.py
jilju: @yitzhak12 Wait is not required at this point because the verification to check the running status of OSDs and its distribution is already done. At this point the OSDs will be in state Running. So wait is not needed. Without wait we can catch an issue if OSDs is going back to non- running state.

tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_reactive_ms.py
jilju: @yitzhak12 Wait is not required at this point because the verification to check the running status of OSDs and its distribution is already done. At this point the OSDs will be in state Running. So wait is not needed. Without wait we can catch an issue if OSDs is going back to non- running state.

tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_reactive_ms.py
jilju: Wait is not needed here due to the reason described above.

tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_reactive_ms.py
yitzhak12: Okay, that makes sense.

yitzhak12 (approved):

PR #7125, opened 2023-02-16
Ypersky multi clones used capacity enhancement

ypersky1980:
This PR contains the following enhancements: 

1) Free ceph capacity calculation is added and the PVC/clones sizes are calculated by diving the free capacity be the required number of maximal PVCs in the cluster ( instead of the total capacity) 

2) Teardown problems should be resolved  

3) Added retries to overcome network glitches during which the test is unable to read data from logs.

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -623,23 +623,17 @@ class CephCluster(object):
 
         Returns:
             int : Total storage capacity in GiB (GiB is for development environment)
-                  if the replica is '0', return 0.
 
         """
         replica = int(self.get_ceph_default_replica())
-        if replica > 0:
-            logger.info(f"Number of replica : {replica}")
-            ceph_pod = pod.get_ceph_tools_pod()
-            ceph_status = ceph_pod.exec_ceph_cmd(ceph_cmd="ceph df")
-            usable_capacity = (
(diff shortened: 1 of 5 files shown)


ocs_ci/ocs/cluster.py
```diff
@@ -641,6 +641,28 @@ def get_ceph_capacity(self):
...
+            logger.info(f"Number of replica : {replica}")
+            ct_pod = pod.get_ceph_tools_pod()
+            output = ct_pod.exec_ceph_cmd(ceph_cmd="ceph df")
+            total_avail = output.get("stats").get("total_bytes")
+            total_used = output.get("stats").get("total_used_raw_bytes")
+            total_free = total_avail - total_used
+            return total_free / replica / constants.BYTES_IN_GB
+        else:
+            # if the replica number is 0, usable capacity can not be calculate
+            # so, return 0 as usable capacity.
```
ramkiperiy, line 663: replica 0 doesn't exist, we can remove this comment or even remove the else IMO.

ypersky1980:
The fix is to increase teardown timeout to 3600 sec.

ocs_ci/ocs/cluster.py
ypersky1980: Removed it, also the "if - else".

ramkiperiy (approved):

pintojoy (approved):
LGTM

PR #7131, opened 2023-02-16
Skip the test 'test_check_pods_status_after_node_failure' when using MS

yitzhak12:
Skip the test 'test_check_pods_status_after_node_failure' when using MS because the functionality in the test is not relevant to the Managed Service.


pintojoy (approved):
LGTM

OdedViner (approved):

PR #7145, opened 2023-02-21
Fix test_nodereplacement_proactive by waiting for node deletion on AWS

### LINKED ISSUE #7069: Wait for node deletion before powering OFF the VM for test_nodereplacement_proactive
https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/407/8281/364646/364667/364669/log?item0Params=filter.eq.hasStats%3Dtrue%26filter.eq.hasChildren%3Dfalse%26filter.in.type%3DSTEP%26filter.in.status%3DFAILED%252CINTERRUPTED

Test fails bcz there are 3 CSR's in total, out of which 1 is for the old node which is getting deleted. The other 2 are for the new node which is fine.

Discussed with OCP team- https://redhat-internal.slack.com/archives/CJARLA942/p1675689335679539 and the suggestion is to wait for the node deletion before powering OFF the VM. Since the node is not deleted by this time, that would have generated a new CSR which is eventually failing the test.

In actual, there should only be 2 CSR's for the new node, one for client credentials, one for its serving certificate. However, if we don't approve those CSRs within 15 minutes, it will retry and create another.

This function `def remove_nodes(nodes):` doesn't wait for node deletion, so we will have to see if we still hit the isue when node is deleted before powering OFF the VM.

am-agrawa:
fixes #7069

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -922,7 +922,9 @@ def delete_and_create_osd_node_aws_upi(osd_node_name):
     remove_nodes([osd_node])
 
     log.info(f"Waiting for node {osd_node_name} to be deleted")
-    osd_node.ocp.wait_for_delete(osd_node_name, timeout=600), f"Node {osd_node_name} is not deleted"
+    osd_node.ocp.wait_for_delete(
+        osd_node_name, timeout=600
+    ), f"Node {osd_node_name} is not deleted"
 
     log.info(f"name of deleted node = {osd_node_name}")
     log.info(f"availability zone of deleted node = {az}")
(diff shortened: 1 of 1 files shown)


prsurve:
Test need's to be ran for AWS upi

ocs_ci/ocs/node.py
```diff
@@ -912,6 +912,7 @@ def delete_and_create_osd_node_aws_upi(osd_node_name):
     """
 
     osd_node = get_node_objs(node_names=[osd_node_name])[0]
+    log.info(f"OSD Node Object is {osd_node}")
```
prsurve, line 915: why do we want to print object here ?

ebenahar (dismissed):

amr1ta:
The changes are looking good only the log added to debug needs to be removed ,
 log.info(f"OSD Node Object is {osd_node}") 

But the last testrun had failed due to some command failure, can you please rerun.

ocs_ci/ocs/node.py
am-agrawa: Used it for debugging, removed

am-agrawa:
> Test need's to be ran for AWS upi

Already ran- https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-trigger-test-pr/4183/

am-agrawa:
> The changes are looking good only the log added to debug needs to be removed , log.info(f"OSD Node Object is {osd_node}")
removed
> 
> But the last testrun had failed due to some command failure, can you please rerun.
It's failing on VMware, and will be taken as part of another issue. Added a comment above with same info.

am-agrawa:
> The changes are looking good only the log added to debug needs to be removed , log.info(f"OSD Node Object is {osd_node}")
> 
> But the last testrun had failed due to some command failure, can you please rerun.
Debug log is now removed.
It's failing on VMware, and will be taken as part of another issue. Added a comment above with same info.

prsurve (approved):

ebenahar (approved):

PR #7164, opened 2023-02-27
Changes for disconnected ODF 4.13 deployment

dahorak:
* add required packages for ODF 4.13 disconnected deployment

### DIFF
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -1434,6 +1434,18 @@ DISCON_CL_REQUIRED_PACKAGES_PER_ODF_VERSION = {
         "odf-multicluster-orchestrator",
         "odf-operator",
     ],
+    "4.13": [
+        "cluster-logging",
+        "elasticsearch-operator",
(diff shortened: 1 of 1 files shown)


petr-balogh (dismissed):

ocs_ci/ocs/constants.py
```diff
@@ -1434,6 +1434,18 @@
         "odf-multicluster-orchestrator",
         "odf-operator",
     ],
+    "4.13": [
```
petr-balogh, line 1437: I see no change from 4.12. Thinking about if we can change it to prevent the duplications.

Maybe just something like:
DISCON_CL_REQUIRED_PACKAGES_PER_ODF_VERSION["4.13"] = DISCON_CL_REQUIRED_PACKAGES_PER_ODF_VERSION["4.12"]
?

dahorak: Nice idea, updated.

petr-balogh (approved):

OdedViner (approved):

PR #7170, opened 2023-02-28
[WIP]Automation for OCS-3887 : Verify storageclass is correctly attached/d…

paraggit:
UI Automation OCS-3887 : "Verify storageclass is correctly attached/deattach from pool."

1. Create rbd pool.
2. Create storageclass with the pool.
3. Check that in pool list and page the storageclass is there.
4. Delete the storageclass and check that it disappear from pool page and pool

### DIFF
--- a/tests/ui/test_rbd_pool_with_sc.py
+++ b/tests/ui/test_rbd_pool_with_sc.py
@@ -1,0 +1,83 @@
+import logging
+import pytest
+from ocs_ci.framework.pytest_customization.marks import (
+    tier1,
+)
+from ocs_ci.framework.testlib import skipif_ocs_version, ui
+
+from ocs_ci.ocs.ui.storageclass import StorageClassUI
+from ocs_ci.ocs.ui.block_pool import BlockPoolUI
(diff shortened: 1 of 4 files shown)


tests/ui/test_rbd_pool_with_sc.py
```diff
@@ -0,0 +1,83 @@
...
+        3. Check that in pool list and page the storageclass is there.
+        4. Delete the storageclass and check that it disapear from pool page and pool
+        """
+        self.sc_name = self.pool_name = None
+        self.bkp_obj = BlockPoolUI(setup_ui)
+        self.sc_obj = StorageClassUI(setup_ui)
+        self.ssn_obj = StorageSystemNavigator(setup_ui)
+
+        # create Storage Class with new Pool
+        self.sc_name, self.pool_name = self.sc_obj.create_storageclass(
```
shyRozen, line 60: Maybe you can use a fixture https://github.com/red-hat-storage/ocs-ci/blob/c05646fca74a911b9f3ac78e1ab608f02715d2e4/tests/conftest.py#L4893 which creates the storageclass and deletes it when the tests is in teardown.

Issue #7210, opened 2023-03-02
Destroy of vsphere cluster doesn't delete resources


petr-balogh:
Noticed that when we are using new terraform and destroying the cluster there is assumption that tfstate file is in the installer repository.

This repo is cleaned every time the new job starts. So we need to make sure we are storing the terraform tfstate file permanently and use this also for destroy and any other operation.

The current destroy in 4.13 looks like:
```
2023-02-21 03:35:29,445 - MainThread - INFO - ocs_ci.utility.utils.exec_cmd.605 - Executing command: /home/jenkins/bin/terraform -chdir=/home/jenkins/workspace/qe-destroy-ocs-cluster/ocs-ci/external/installer/upi/vsphere/ destroy -refresh=false '-var-file=/home/jenkins/current-cluster-dir/openshift-cluster-dir/terraform_data/terraform.tfvars' -auto-approve
2023-02-21 03:35:32,797 - MainThread - DEBUG - ocs_ci.utility.utils.exec_cmd.622 - Command stdout:
[0m[1m[32mNo changes.[0m[1m No objects need to be destroyed.[0m
 
[0mEither you have not created any objects yet or the existing objects were
already deleted outside of Terraform.
[33mâ•·[0m[0m
[33mâ”‚[0m [0m[1m[33mWarning: [0m[0m[1mValue for undeclared variable[0m
[33mâ”‚[0m [0m
[33mâ”‚[0m [0m[0mThe root module does not declare a variable named "folder" but a value was
[33mâ”‚[0m [0mfound in file
[33mâ”‚[0m [0m"/home/jenkins/current-cluster-dir/openshift-cluster-dir/terraform_data/terraform.tfvars".
[33mâ”‚[0m [0mIf you meant to use this value, add a "variable" block to the
[33mâ”‚[0m [0mconfiguration.
[33mâ”‚[0m [0m
[33mâ”‚[0m [0mTo silence these warnings, use TF_VAR_... environment variables to provide
[33mâ”‚[0m [0mcertain "global" settings to all configurations in your organization. To
[33mâ”‚[0m [0mreduce the verbosity of these warnings, use the -compact-warnings option.
[33mâ•µ[0m[0m
[0m[1m[32m
Destroy complete! Resources: 0 destroyed.
```

It thinks no resource exists because of missing tfstate file, hence it's continue and we have some backup solution in the destroy that the VMs in vSphere are deleted, but no AWS Route53 entries nor IPAM entries are properly cleaned.

PR #7211, opened 2023-03-02
Fix vsphere terraform destroy cluster issue for 4.13

petr-balogh:
Fixes: #7210


ocs_ci/deployment/terraform.py
```diff
@@ -17,31 +17,57 @@ class Terraform(object):
...
+        try:
+            with open(terraform_log_path, "r") as fd:
+                logger.debug(f"Reading terraform version from {terraform_log_path}")
+                for each_line in fd:
+                    if "Terraform version:" in each_line:
+                        terraform_version = each_line.split()[-1]
+                        logger.debug(
+                            f"Terraform version which will be use: {terraform_version}"
+                        )
+                        return terraform_version
```
dahorak, line 66: Just cosmetic note - this `return terraform_version` is probably not necessary here, as I think it will be used the return from `finally`.

dahorak (approved):

OdedViner (approved):

PR #7214, opened 2023-03-02
WIP: Add wait for crash collector in test_add_ocs_node

### LINKED ISSUE #7085: Add the "ignore_leftovers" decorator in the test "test_ceph_csidriver_runs_on_non_ocs_nodes"
We can see in the failed test teardown here https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster-prod/6613/testReport/tests.manage.z_cluster.cluster_expansion.test_verify_ceph_csidriver_runs_on_non_ocs_nodes/TestCheckTolerationForCephCsiDriverDs/test_ceph_csidriver_runs_on_non_ocs_nodes/ that the test "test_ceph_csidriver_runs_on_non_ocs_nodes" failed because leftovers. Also, I have seen it in one other test.
This is expected because we add three worker `app_nodes` in the test. 
So we need to add the `ignore_leftovers` decorator. 

Also, maybe we should delete the `app_nodes` after the test.

DanielOsypenko:
Fixes #7085

test_ceph_csidriver_runs_on_non_ocs_nodes fails with msg: failed on teardown with "ocs_ci.ocs.exceptions.ResourceLeftoversException" during environment_checker func

Test tests/manage/z_cluster/cluster_expansion/test_node_expansion.py::TestAddNode::test_add_ocs_node does not have leftovers check on finish and the next test immediately starts with collecting cluster resources configuration, **before all resources created on the new nodes**

Crashcollector pods produced in the middle of the test_ceph_csidriver_runs_on_non_ocs_nodes fail environment_checker on tearDown

* It was taken on consideration 2 options, either remove newly created nodes or wait until resources created
First option has been dropped because remove_node() may fail on ipi deployments.

### DIFF
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -2686,7 +2686,9 @@ def delete_all_osd_removal_jobs(namespace=defaults.ROOK_CLUSTER_NAMESPACE):
 
 
 def get_crashcollector_pods(
-    crashcollector_label=constants.CRASHCOLLECTOR_APP_LABEL, namespace=None
+    crashcollector_label=constants.CRASHCOLLECTOR_APP_LABEL,
+    namespace=None,
+    nodes: set = None,
 ):
     """
     Fetches info about crashcollector pods in the cluster
@@ -2696,6 +2698,7 @@ def get_crashcollector_pods(
             (default: defaults.CRASHCOLLECTOR_APP_LABEL)
         namespace (str): Namespace in which ceph cluster lives
             (default: defaults.ROOK_CLUSTER_NAMESPACE)
+        nodes (set): filter result by node names
 
     Returns:
         list : of crashcollector pod objects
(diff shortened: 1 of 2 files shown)


am-agrawa:
Also tagging @nehaberry to seek further inputs.

DanielOsypenko:
@amr1ta 
The proposed solution would not be effective because csi resources are already deployed and may be redeployed on a new nodes when the old ones will be disrupted. Exactly this happens in the test, next after test_add_ocs_node.
Because of the unexpected node drains we have crash collector pods deployed on previously (during test_add_ocs_node) added nodes. This happens in the middle of the test **test_ceph_csidriver_runs_on_non_ocs_nodes**.

This looks like a System issue.
More about it and proposed solution in https://github.com/red-hat-storage/ocs-ci/issues/7085#issuecomment-1486695495

PR #7222, opened 2023-03-02
removed log of ZombieProcess and NoSuchProcess exceptions

DanielOsypenko:
update release-4.6 branch with ZombieProcess & NoSuchProcess exception print in logs (prints removed)


petr-balogh:
I don't think we need to go to 4.6 release. Only to 4.8 and above please as 4.8 is last supported version.

PR #7231, opened 2023-03-03
Introduce wait time before creating resources for test_rolling_nodes_restart

am-agrawa:
Test fails with-
```
Message: ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc new-project namespace-test-a166037ce3714829bd8cb2e39.
Error is Error from server (ServiceUnavailable): the server is currently unable to handle the request (get projectrequests.project.openshift.io)
```

Run ID: 1676239998 || Suite: tier4b || ODF 4.12.1-11, OCP 4.12, vsphere

### DIFF
--- a/tests/manage/z_cluster/nodes/test_nodes_restart.py
+++ b/tests/manage/z_cluster/nodes/test_nodes_restart.py
@@ -1,5 +1,6 @@
 import logging
 import pytest
+import time
 
 
 from ocs_ci.framework.testlib import (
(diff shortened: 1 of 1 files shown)


am-agrawa:
Test passed here- https://url.corp.redhat.com/847aa76

amr1ta:
As we discussed, the waiting time varies so would it be possible to use wait() instead of sleep. For reference: https://realpython.com/python-sleep/#:~:text=Using%20Event.wait(),-The%20threading%20module&text=With%20time.sleep()%20%2C%20your,time.sleep()%20is%20blocking.

PR #7021, opened 2023-01-31
Fixing OBC bound list count mismatch

tiffanyn108:
Fixing scale [issue # 6942](https://github.com/red-hat-storage/ocs-ci/issues/6942):
* Using 2 different data files to store obc status
* Using separate namespace to create scaled obcs

PR #7244, opened 2023-03-08
Fixing OBC bound list count mismatch (#7021)

tiffanyn108:
Fixing scale issue # 6942:
    Using 2 different data files to store obc status
    Using separate namespace to create scaled obcs


ramkiperiy (approved):

mashetty330 (approved):
lgtm

Issue #7250, opened 2023-03-09
`test_deletion_of_store_with_attached_buckets` fails because it no longer receives the underlying error from `Backingstore:delete()`


sagihirshfeld:
`backingstore.py::Backingstore::delete` was enhanced in #6499 to catch errors when attempting to delete the Backingstore. However, `test_deletion_of_store_with_attached_buckets` is expecting the underlying error to be raised:

```Error from server: admission webhook "admissionwebhook.noobaa.io" denied the request: cannot complete because pool <backingstore_name> in "IN_USE" state```

PR #7251, opened 2023-03-09
Modify `Backingstore.py::delete` so it raises deletion errors if retrying is disabled

sagihirshfeld:
Limit of the error handling in `Backingstore.py::delete` only to when we intend to retry the deletion after the first failed attempt.

Fixes: #7250


tiffanyn108 (approved):

keesturam (approved):
LGTM

PR #7262, opened 2023-03-12
Added option to enable in_transit_encryption via UI

OdedViner:


ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -183,6 +183,17 @@ def install_mcg_only_cluster(self):
...
 
+    def configure_in_transit_encryption(self):
+        """
+        Configure in_transit_encryption
+
+        """
+        if config.ENV_DATA.get("in_transit_encryption"):
+            logger.info("Enable in-transit encryption")
+            self.select_checkbox_status(
+                status=True, locator=self.dep_loc["enable_in_transit_encryption"]
```
DanielOsypenko, line 194: @OdedViner  
Do we need to disable it in the end of the test / tearDown? 
Will it affect other test cases running on cluster? 
Is the feature "in-transit encryption" applicable for all deployments?

paraggit: @DanielOsypenko For UI this option ( to enable in-transit encryption) is available only during deployment not after that,  so we cant disable it form UI.  It should not affect the other tests cases , If it is then those are real issue we need to investigate.

paraggit (approved):

ocs_ci/ocs/ui/deployment_ui.py
DanielOsypenko: ack, thanks

DanielOsypenko (approved):

PR #7280, opened 2023-03-13
Automation of intransit encryption sanity test.

paraggit:
Added a tests case to sanity check of intransit encryption feature.

Test Steps:
        1. Verify intransit Encryption is Enable on setup.
        2. Disable Intransit Encryption from setup.
        3. Verify config is removed from the cluster.
        4. re-enable intransit encryption on setup.
        5. Again verify intransit encryption config is added.


ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -880,6 +880,52 @@ def in_transit_encryption_verification():
...
+        resource_name=constants.DEFAULT_CLUSTERNAME,
+        namespace=defaults.ROOK_CLUSTER_NAMESPACE,
+    )
+
+    patch_cmd = '[{ "op": "replace", "path": "/spec/network/connections/encryption/enabled", "value": false }]'
+    log.info("patching storageclass to disable in-transit encryption.")
+    ocp_obj.patch(params=patch_cmd, format_type="json")
+    ocp_obj.wait_for_phase("Progressing", timeout=60)
+    verify_storage_cluster()
+    return True
```
ebenahar, line 897: Seems redundant

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -880,6 +880,52 @@ def in_transit_encryption_verification():
...
+        resource_name=constants.DEFAULT_CLUSTERNAME,
+        namespace=defaults.ROOK_CLUSTER_NAMESPACE,
+    )
+
+    patch_cmd = '[{ "op": "replace", "path": "/spec/network/connections/encryption/enabled", "value": true }]'
+    log.info("patching storageclass to enable in-transit encryption.")
+    ocp_obj.patch(params=patch_cmd, format_type="json")
+    ocp_obj.wait_for_phase("Progressing", timeout=60)
+    verify_storage_cluster()
+    return True
```
ebenahar, line 914: Seems redundant. Please remove through this PR

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -880,6 +880,52 @@ def in_transit_encryption_verification():
...
+
+
+def verify_in_transit_encryption_keys_exists():
+    """
+    verify that "ms_client_mode", "ms_cluster_mode", "ms_service_mode" keys are exists
+    in 'ceph config dump output'.
+    """
+    try:
+        in_transit_encryption_verification()
+    except Exception:
```
ebenahar, line 924: `in_transit_encryption_verification()` raises ValueError. Please except this exception instead of `Exception`, as that is too wide.
Even better IMO, would be to modify in_transit_encryption_verification() so it will return True in case the verification succeeds and return False in case the verification fails.
That way, you won't need to think of which exception to catch here.
The exception then could be raised in the test case level with `assert`

tests/encryption/test_intransit_encryption_sanity.py
```diff
@@ -0,0 +1,40 @@
...
+    skipif_intransit_encryption_notset,
+    tier1,
+    skipif_ocs_version,
+)
+
+log = logging.getLogger(__name__)
+
+
+@skipif_intransit_encryption_notset
+class TestInTransitEncryptionSanity:
```
ebenahar, line 18: missing polarion id

paraggit: Added polarion ID

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -854,16 +854,41 @@ def osd_encryption_verification():
         log.info("Luks header info found for all the encrypted osds")
 
 
-def in_transit_encryption_verification():
+def ceph_config_dump():
+    """Get the 'ceph config dump' output.
+
+    Returns:
+        dict: The output of the 'ceph config dump' command as a dict.
     """
```
ebenahar, line 862:
```suggestion
    """
    Get the 'ceph config dump' output.

    Returns:
        dict: The output of the 'ceph config dump' command as a dict.
      
    """
```
Please apply to all places

tests/encryption/test_intransit_encryption_sanity.py
```diff
@@ -0,0 +1,51 @@
...
+        2. Disable Encryption
+        3. Verify in-transit encryption configuration is removed.
+        4. Enable encryption Again and verify it.
+        5. Verify in-transit encryption config is exists.
+        """
+        log.info("Verifying the in-transit encryption is enable on setup.")
+        assert in_transit_encryption_verification()
+
+        log.info("Disabling the in-transit encryption.")
+        set_in_transit_encryption(enabled=False)
```
ebenahar, line 37: there should be a fixture with a finalizer that in case of a test failure, would bring the cluster back to how it was before the test

ocs_ci/ocs/resources/storage_cluster.py
paraggit: Formatted the doc strings

tests/encryption/test_intransit_encryption_sanity.py
paraggit: Added Finalizer statement in test step.

ebenahar (dismissed):

avd-sagare (approved):

tests/encryption/test_intransit_encryption_sanity.py
```diff
@@ -0,0 +1,54 @@
...
+        The test does the following:
+        1. Verify in-transit Encryption is Enable on setup.
+        2. Disable Encryption
+        3. Verify in-transit encryption configuration is removed.
+        4. Enable encryption Again and verify it.
+        5. Verify in-transit encryption config is exists.
+
+        """
+        log.info("Verifying the in-transit encryption is enable on setup.")
+        assert in_transit_encryption_verification()
```
ebenahar, line 36: why not to place the `addfinalizer` in the `teardown` fixture itself?
For example https://github.com/red-hat-storage/ocs-ci/blob/6a2f2fa9c3c37bfd6f36dffd16896c35b19cdce6/tests/e2e/kcs/test_noobaa_db_backup_and_recovery.py#L86

paraggit: Moved addfinalizer in trardown fixture.

tests/encryption/test_intransit_encryption_sanity.py
```diff
@@ -0,0 +1,57 @@
...
+
+        request.addfinalizer(teardown)
+
+    @tier1
+    @skipif_ocs_version("<4.12")
+    @pytest.mark.polarion_id("OCS-4861")
+    def test_intransit_encryption_enable_disable_statetransition(self):
+        """
+        The test does the following:
+        1. Verify in-transit Encryption is Enable on setup.
```
shyRozen, line 32: Why we assume that in transit encryption will be enabled? I need to understand the test flow. Is it enabled at deployment stage? If yes maybe we need a mark to run it only when it is configured in deployment.
If not, let's check if it is enabled and if not enabled it.

tests/encryption/test_intransit_encryption_sanity.py
```diff
@@ -0,0 +1,57 @@
...
+class TestInTransitEncryptionSanity:
+    @pytest.fixture(autouse=True)
+    def set_encryption_at_teardown(self, request):
+        def teardown():
+            set_in_transit_encryption()
+
+        request.addfinalizer(teardown)
+
+    @tier1
+    @skipif_ocs_version("<4.12")
```
shyRozen, line 27: Shouldn't it be "<4.13"

tests/encryption/test_intransit_encryption_sanity.py
shyRozen: This probably relates to the question  why we enable it in teardown.

paraggit: All in-transit tests will be run using the deployment configuration [upi_1az_rhcos_vsan_3m_3w_intransit_encryption.yaml](https://github.com/red-hat-storage/ocs-ci/blob/master/conf/deployment/vsphere/upi_1az_rhcos_vsan_3m_3w_intransit_encryption.yaml). Therefore, I am assuming that intransit encryption is already enabled on the setup. I have added the marker skipif_intransit_encryption_notset so that the initial check can be skipped if intransit encryption is not set.

tests/encryption/test_intransit_encryption_sanity.py
paraggit: Corrected.

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -879,6 +910,88 @@ def in_transit_encryption_verification():
...
+        log.error("ceph mons are not configured with v2 protocol version. ")
+        raise ValueError(
+            "ceph mon protocol is not configured with 'v2' version which is recomended for in-transit encryption."
+        )
+
+    log.info("Ceph mons are configured with 'v2' protocol version.")
+    return True
+
+
+def get_in_transit_encryption_state():
```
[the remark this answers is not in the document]
paraggit: Corrected.

tests/encryption/test_intransit_encryption_sanity.py
shyRozen: Do we have a test that tests the in transit configuration without deployment?
I we have that's great and if we don't we can remove the mark and the flow test could be that we check at start if it enabled or not. If it is enabled then the same flow, if it is not enabled then enabled it and then the rest of the flow - that just in case we don't have cover for none deployment configuration. wdyt?

shyRozen: Maybe related to the scenario I've described - if it was i deployment config fail it if we don't have in-transit encryption.

paraggit: I have added the changes as you suggested.

tests/encryption/test_intransit_encryption_sanity.py
```diff
@@ -0,0 +1,64 @@
...
+        The test does the following:
+        1. Enable in-transit Encryption if not Enabled.
+        2. Verify in-transit Encryption is Enable on setup.
+        3. Disable Encryption
+        4. Verify in-transit encryption configuration is removed.
+        5. Enable encryption Again and verify it.
+        6. Verify in-transit encryption config is exists.
+
+        """
+        if not get_in_transit_encryption_config_state():
```
shyRozen, line 43: We need to know here if it is deployment config, and in-transit encryption is not enabled we should fail the test, else enable in-transit encryption.

tests/encryption/test_intransit_encryption_sanity.py
```diff
@@ -0,0 +1,64 @@
...
+        log.info("Disabling the in-transit encryption.")
+        set_in_transit_encryption(enabled=False)
+
+        # Verify that encryption is actually disabled by checking that a ValueError is raised.
+        log.info("Verifying the in-transit encryption is disabled.")
+        with pytest.raises(ValueError):
+            assert not in_transit_encryption_verification()
+
+        log.info("Re-enabling in-transit encryption.")
+        set_in_transit_encryption()
```
shyRozen, line 58: If it is deployment config we restore, if it is not deployment config, we should stay disabled.

tests/encryption/test_intransit_encryption_sanity.py
paraggit: Done

tests/encryption/test_intransit_encryption_sanity.py
paraggit: Done

shyRozen (approved):

ebenahar (approved):

PR #7360, opened 2023-03-22
Add Multi-Stage build Containerfile

rabin-io:
This PR allow us to build the tests images in a more optimized matter.

Note: The image is still very big, that is because of the azure package, which with a lot of legacy code, there is an issue about this opened in there repo, https://github.com/Azure/azure-sdk-for-python/issues/17801

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.git/
+.venv/
+venv/
--- a/Docker_files/Containerfile.ci
+++ b/Docker_files/Containerfile.ci
@@ -1,0 +1,43 @@
+FROM registry.access.redhat.com/ubi8/ubi:latest as BUILDER
+
+ENV OCS_CI_DIR=/opt/ocs-ci
+WORKDIR "${OCS_CI_DIR}"
+
+RUN dnf install --nodocs -y python38 python38-devel git libcurl-devel gcc openssl-devel libxml2-devel
+
+# Copy the entire source tree to the image
+# TODO: Check to see if there are any more files we can drop from the image
+# TODO: The azure package is very big (see issue: https://github.com/Azure/azure-sdk-for-python/issues/17801)
+
+COPY / "${OCS_CI_DIR}"
+
+RUN pushd "${OCS_CI_DIR}" \
+    && mkdir -p "${OCS_CI_DIR}/data" \
+    && python3 -m venv venv \
+    && source venv/bin/activate \
+    && pip3 install --upgrade pip \
+    && pip3 install setuptools==65.5.0 \
+    && pip3 install -r requirements.txt \
+    && rm -rf .git
+
+### Runner stage
+
+FROM registry.access.redhat.com/ubi8/ubi:latest as RUNNER
+
+ENV OCS_CI_DIR="/opt/ocs-ci" \
+    VIRTUAL_ENV="/opt/ocs-ci/venv" \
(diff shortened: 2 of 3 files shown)


petr-balogh (dismissed):

ebenahar (dismissed):

Containerfile
```diff
@@ -0,0 +1,42 @@
...
+WORKDIR "${OCS_CI_DIR}"
+
+RUN dnf install --nodocs -y python38 python38-devel git libcurl-devel gcc openssl-devel libxml2-devel
+
+# Copy the entire source tree to the image
+# TODO: check to see if there are any more files we can drop from the image
+# TODO: the azure package is very big, we should consider to see how we can compact it
+COPY / "${OCS_CI_DIR}"
+
+RUN pushd "${OCS_CI_DIR}" \
```
dkeler, line 13: Can you use `WORKDIR` instead?

Containerfile
```diff
@@ -0,0 +1,42 @@
...
+    && python3 -m venv venv \
+    && source venv/bin/activate \
+    && pip3 install --upgrade pip \
+    && pip3 install setuptools==65.5.0 \
+    && pip3 install -r requirements.txt \
+    && rm -rf .git
```
dkeler, line 19: Is it needed even if it's already in the .dockerignore?

Containerfile
```diff
@@ -0,0 +1,42 @@
...
+# Copy the entire source tree to the image
+# TODO: check to see if there are any more files we can drop from the image
+# TODO: the azure package is very big, we should consider to see how we can compact it
+COPY / "${OCS_CI_DIR}"
+
+RUN pushd "${OCS_CI_DIR}" \
+    && python3 -m venv venv \
+    && source venv/bin/activate \
+    && pip3 install --upgrade pip \
+    && pip3 install setuptools==65.5.0 \
```
dkeler, line 17: Why not to put setuptools in the requirements.txt?

Containerfile
```diff
@@ -0,0 +1,42 @@
...
+RUN dnf install -y --nodocs python38 git jq rsync \
+    && dnf clean all \
+    && rm -rf /var/cache/yum \
+    && curl -sL https://github.com/mikefarah/yq/releases/download/v4.32.2/yq_linux_amd64.tar.gz | tar -C /usr/local/bin -zxvf - ./yq_linux_amd64 \
+    && mv -v /usr/local/bin/yq_linux_amd64 /usr/local/bin/yq
+
+COPY --from=BUILDER "${OCS_CI_DIR}" "${OCS_CI_DIR}"
+COPY --chmod=755 /Docker_files/ocsci_container/scripts/run-ocs-ci.sh /usr/local/bin/
+
+CMD [ "run-ci", "--help" ]
```
dkeler, line 42: Can we have `run-ci` as entrypoint?

rabin-io: I had this before, and I removed it, as it limit the image execution options

Containerfile
rabin-io: Not sure, this was based on the original work @OdedViner did [here](https://github.com/red-hat-storage/ocs-ci/tree/master/Docker_files/ocsci_container), so I just adapted a bit.

But I do think your point is valid, @OdedViner what was the reason?

Containerfile
rabin-io: Probably

OdedViner (approved):

ebenahar (approved):

PR #7367, opened 2023-03-23
 Automation for BZ2142901

AviadP:
Checks if "liveness-prometheus" container exist / running on CSI pods


tests/e2e/lifecycle/test_no_liveness_probe.py
```diff
@@ -0,0 +1,71 @@
...
+    container_names = list()
+    for item in items:
+        containers = item.get("spec").get("containers")
+        container_names += [c.get("name") for c in containers]
+
+    log.debug(f"Containers: {container_names}")
+
+    return set(container_names)
+
+
```
PrasadDesala, line 43: Polarion ID, tier and Bugzilla markers are missing, please add

tests/e2e/lifecycle/test_no_liveness_probe.py
```diff
@@ -0,0 +1,71 @@
+# BZ2142901 Automated Test
```
PrasadDesala, line 1: Lifecycle may not be a relevant place for this test, should it be placed under z_cluster since it is a cluster-level pod check?
@prsurve wdyt?

tests/e2e/lifecycle/test_no_liveness_probe.py
```diff
@@ -0,0 +1,71 @@
...
+    return set(container_names)
+
+
+def test_no_liveness_container():
+    """
+    Automated test for BZ #2142901
+    Checks if "liveness-prometheus" container is running on CSI pods
+
+    """
+    csi_pod = OCP(
```
PrasadDesala, line 50: please change the var name to csi_plugin_pod to be more accurate

tests/e2e/lifecycle/test_no_liveness_probe.py
```diff
@@ -0,0 +1,71 @@
...
+        kind=POD, namespace=OPENSHIFT_STORAGE_NAMESPACE, selector=CSI_CEPHFSPLUGIN_LABEL
+    )
+    csi_prov_pod = OCP(
+        kind=POD,
+        namespace=OPENSHIFT_STORAGE_NAMESPACE,
+        selector=CSI_CEPHFSPLUGIN_PROVISIONER_LABEL,
+    )
+    rbd_pod = OCP(
```
PrasadDesala, line 58: please change the var name to rbd_plugin_pod to be more accurate

tests/e2e/lifecycle/test_no_liveness_probe.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+    rbd_prov_pod = OCP(
+        kind=POD,
+        namespace=OPENSHIFT_STORAGE_NAMESPACE,
+        selector=CSI_RBDPLUGIN_PROVISIONER_LABEL,
+    )
+    for pods in (csi_pod, rbd_pod, rbd_prov_pod, csi_prov_pod):
+        assert LIVENESS_CONTAINER not in get_containers_names_by_pod(
+            pods
+        ), "liveness-prometheus container found"
```
PrasadDesala, line 71: Please also add a success log message that the "liveness-prometheus container is not found in any of the csi pods as expected".

tests/e2e/lifecycle/test_no_liveness_probe.py
DanielOsypenko: also tier mark missed. If we want to run this test as a part of regression we need to define a tier mark

tests/e2e/lifecycle/test_no_liveness_probe.py
```diff
@@ -0,0 +1,71 @@
...
+    CSI_RBDPLUGIN_PROVISIONER_LABEL,
+)
+from ocs_ci.ocs.ocp import OCP
+
+log = getLogger("__name__")
+
+LIVENESS_CONTAINER = "liveness-prometheus"
+
+
+def get_containers_names_by_pod(pod: OCP) -> set:
```
DanielOsypenko, line 19: it may be helpful to move this routine func to the pod.py

tests/e2e/lifecycle/test_no_liveness_probe.py
```diff
@@ -0,0 +1,71 @@
...
+    POD,
+    CSI_CEPHFSPLUGIN_PROVISIONER_LABEL,
+    CSI_RBDPLUGIN_LABEL,
+    CSI_RBDPLUGIN_PROVISIONER_LABEL,
+)
+from ocs_ci.ocs.ocp import OCP
+
+log = getLogger("__name__")
+
+LIVENESS_CONTAINER = "liveness-prometheus"
```
DanielOsypenko, line 16: LIVENESS_CONTAINER may be moved to ocs_ci.ocs.constants for further use (not an issue, just recommendation)

AviadP: i cant see any other use case for this const, out of this test - it should not exist on ODF

DanielOsypenko (approved):

tests/e2e/lifecycle/test_no_liveness_probe.py
prsurve: Yes test can be moved to brown squad

yitzhak12 (approved):

PR #7380, opened 2023-03-27
selinux relabeling solution

vkathole:
Automation: ODF solution for SELinux issue for Experian

### DIFF
--- a/tests/e2e/flowtest/test_selinux_relabeling_solution.py
+++ b/tests/e2e/flowtest/test_selinux_relabeling_solution.py
@@ -15,7 +15,7 @@ from ocs_ci.helpers import helpers
 
 TARFILE = "cephfs.tar.gz"
 SIZE = "20Gi"
-TFILES = 40
+TFILES = 2000
 SAMPLE_TEXT = b"QWERTYUIOP ASDFGJKL ZXCVBNM"
 ntar_loc = mkdtemp()
 
@@ -27,12 +27,10 @@ class TestSelinuxrelabel(E2ETest):
         log.info(f"Creating {TFILES} files ")
         onetenth = TFILES / 10
         endoften = onetenth - 1
-
         tarfile = os.path.join(ntar_loc, TARFILE)
         one_dir = mkdtemp()
         log.info(f"dir crated{one_dir}")
         n = 0
-
         while n in range(0, 11):
             first_dir = mkdtemp(suffix="Fdir", dir=one_dir)
             log.info(f"created dir {first_dir}")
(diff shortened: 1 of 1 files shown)


tests/e2e/flowtest/test_selinux_relabeling_solution.py
```diff
@@ -0,0 +1,255 @@
...
+            selector=f"name=nfs-test-pod-2",
+            dont_allow_other_resources=True,
+            timeout=60,
+        )
+        pod_objs = pod.get_all_pods(
+            namespace=self.namespace, selector=["nfs-test-pod-2"], selector_label="name"
+        )
+
+        pod_obj = pod_objs[0]
+        log.info(f"pod obj name----{pod_obj.name}")
```
DanielOsypenko, line 139: very minor issue - make a space between "name----" and pod name

tests/e2e/flowtest/test_selinux_relabeling_solution.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def cleanup(self):
+        # Delete deployment config
+        cmd_delete_deployment_config = "delete dc nfs-test-pod"
+        self.storage_cluster_obj.exec_oc_cmd(cmd_delete_deployment_config)
+        # Delete pvc
+        self.nfs_pvc_obj.delete()
+        log.info("Teardown complete")
+
+    @ignore_leftovers
```
DanielOsypenko, line 223: what will be leftovers here? 
May it be hypothetically that resources may create after this test? if so it will fail the next test which is not labeled with @ignore_leftovers deco

tests/e2e/flowtest/test_selinux_relabeling_solution.py
```diff
@@ -0,0 +1,255 @@
...
+            size="10Gi",
+            do_reload=True,
+            access_mode=constants.ACCESS_MODE_RWO,
+            volume_mode="Filesystem",
+        )
+        log.info(f"{self.nfs_pvc_obj} created")
+
+        # Create deployment config for app pod
+        log.info("----create deployment config----")
+        deployment_config = """
```
DanielOsypenko, line 80: since this config is static, probably will be better to use a common approach to store it like a yaml file, to reuse afterwards. WDYT?

tests/e2e/flowtest/test_selinux_relabeling_solution.py
```diff
@@ -0,0 +1,255 @@
...
+        self.nfs_pvc_obj = helpers.create_pvc(
+            sc_name=self.nfs_sc,
+            namespace=self.namespace,
+            pvc_name="nfs-pvc-2",
+            size="10Gi",
+            do_reload=True,
+            access_mode=constants.ACCESS_MODE_RWO,
+            volume_mode="Filesystem",
+        )
+        log.info(f"{self.nfs_pvc_obj} created")
```
DanielOsypenko, line 76: don't see that PVC or its parent obj OCP have __str__ or __repr__ func implemented. What will be f"{self.nfs_pvc_obj} looking like?e

PR #7391, opened 2023-03-30
[Bug 2049081] Automate ReclaimSpaceJob is failing for RBD RWX PVC

pintojoy:
Bug https://bugzilla.redhat.com/show_bug.cgi?id=2049081

### DIFF
--- a/tests/manage/pv_services/space_reclaim/test_rbd_space_reclaim.py
+++ b/tests/manage/pv_services/space_reclaim/test_rbd_space_reclaim.py
@@ -33,12 +33,20 @@ class TestRbdSpaceReclaim(ManageTest):
     """
 
     @pytest.fixture(autouse=True)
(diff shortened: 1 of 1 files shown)


tests/manage/pv_services/space_reclaim/test_rbd_space_reclaim.py
```diff
@@ -47,69 +55,106 @@ def setup(self, project_factory, storageclass_factory, create_pvcs_and_pods):
...
+        3. Create two files of size 10GiB for RWWO and 20GB space for RWX
         4. Verify the increased used size of the RBD pool
-        5. Delete one file
+        5. Delete one file for RWO and erase 10GB space for RWX
         6. Create ReclaimSpaceJob
         7. Verify the decreased used size of the RBD pool
 
         """
-
+        size = 10
```
ramkiperiy, line 94: Good to make this a random value than 10, so it does varies for each run.

PR #7393, opened 2023-03-30
Use specific RGW endpoint from zone for MDR deployment

petr-balogh:
Ticket:
https://issues.redhat.com/browse/OCSQE-913

### DIFF
--- a/ocs_ci/deployment/helpers/external_cluster_helpers.py
+++ b/ocs_ci/deployment/helpers/external_cluster_helpers.py
@@ -161,7 +161,12 @@ class ExternalCluster(object):
             logger.info(ex)
             cmd = "grep -e '^rgw frontends' /etc/ceph/ceph.conf"
             rgw_node = get_rgw_endpoint()
-            rgw_conn = Connection(host=rgw_node, user=self.user, password=self.password)
+            rgw_conn = Connection(
+                host=rgw_node,
+                user=self.user,
+                private_key=self.ssh_key,
+                password=self.password,
+            )
             _, out, _ = rgw_conn.exec_cmd(cmd)
             port = out.split(":")[-1]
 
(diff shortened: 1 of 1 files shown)


petr-balogh:
Tested against regressions here that I didn't break old deployment without zone:
https://url.corp.redhat.com/eb723fd

The run failed but on different thing becasue previous execution didn't cleanup properly.

clacroix12 (approved):

ocs_ci/deployment/helpers/external_cluster_helpers.py
```diff
@@ -378,7 +383,12 @@ def get_rgw_endpoint():
 
     """
     rgw_endpoint = None
+    zone = config.ENV_DATA.get("zone")
```
shylesh, line 386: @petr-balogh , I am not able to visualize about the source of this zone info. the data center zone info from config.EXTERNAL_MODE["external_cluster_node_roles"] comes from the external cluster yaml that we will have but I was wondering from where this ENV_DATA.zone gets populated? I tried dumping node info from a vmware cluster and I don't see any, if we are passing this zone info from a config file through --ocsci-conf, which file is that ? Am I missing something totally!

petr-balogh: This is populated from multicluster job from this place: https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/-/blob/master/jobs/pipelines/qe_dr_setup/Jenkinsfile#L46 where we are creating the yaml text config.

This we looked at with @dahorak and should be created for each cluster.

petr-balogh: But for non mdr job it's not needed and I don't want to break it.

shylesh: @petr-balogh , got it, my mistake, I forgot that this is also related to single cluster as well.

shylesh (approved):
LGTM

PR #7441, opened 2023-04-06
Add notification mail with default value. Fix for issues#7440

suchita-g:


vavuthu (approved):

ebondare (approved):

suchita-g:
As per the discussion in QE channel thread https://chat.google.com/room/AAAABfd8ivI/W9MghuASTWM, we are avoiding using of mail id in public repo and moving the same change in credential-stage and credential-prod encrypted file

PR #7448, opened 2023-04-07
Install resolvelib package in case of RHEL8

shylesh:


shylesh:
Not a valid fix, issue is in the OCP itself.

PR #7501, opened 2023-04-20
Use fusion-storage as cluster namespace during deployment

fbalak:


fbalak:
Fusion deployment verified: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/23434.
Resources are being created in managed-fusion namespace.

fbalak:
Tier1 tests verified: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/23473/
The change seems to not breaking any tests.

jilju (approved):

suchita-g (dismissed):
LGTM

PrasadDesala (dismissed):

ocs_ci/helpers/helpers.py
```diff
@@ -984,7 +984,7 @@ def validate_cephfilesystem(fs_name):
 
 
 def create_ocs_object_from_kind_and_name(
-    kind, resource_name, namespace=constants.OPENSHIFT_STORAGE_NAMESPACE
+    kind, resource_name, namespace=config.ENV_DATA["cluster_namespace"]
```
dahorak, line 987: I'm not sure, if it might be problem or not in this case, but using value from dictionary key here means, that the default `namespace` value will be the one present in the `config` in the time of initialization of the function, not the value from `config` from the time when the function will be executed (I'm not sure, how to describe it better).
In other words, if this function will be initialized before whole configuration will be loaded (or if the `config.ENV_DATA["cluster_namespace"]` will be changed later for some other reason, which is probably not the case here), the default value for `namespace` parameter will be wrong.

fbalak: It seems that for deployment and regression runs is used correct value according to verification runs. I believe that it should contain correct value at the time of the execution but please let me know if you think that this needs to be changed. I can keep the argument as it is and change it in function calls.

dahorak: I was thinking about setting the default `namespace` to `None` and then in the beginning of the function set the actual default value:
```
if `namespace` is none`:
  namespace = config.ENV_DATA["cluster_namespace"]
```

But it is maybe over-complicated and not necessary in this case.

fbalak: Sounds good to me. I will make the change.

ocs_ci/helpers/performance_lib.py
```diff
@@ -108,7 +108,7 @@ def run_command(cmd, timeout=600, out_format="string", **kwargs):
     return output
 
 
-def run_oc_command(cmd, namespace=constants.OPENSHIFT_STORAGE_NAMESPACE):
+def run_oc_command(cmd, namespace=config.ENV_DATA["cluster_namespace"]):
```
dahorak, line 111: Same as above, change the default value to `None` and set it to the default from config in the code also here please.

ocs_ci/utility/kms.py
```diff
@@ -888,7 +888,7 @@ def validate_external_vault(self):
             raise NotFoundError("KMS flag not found")
 
     def create_vault_csi_kms_token(
-        self, namespace=constants.OPENSHIFT_STORAGE_NAMESPACE
+        self, namespace=config.ENV_DATA["cluster_namespace"]
```
dahorak, line 891: Same as above, change the default value to `None` and set it to the default from config in the code also here please.

ocs_ci/utility/kms.py
```diff
@@ -905,7 +905,7 @@ def create_vault_csi_kms_connection_details(
         self,
         kv_version,
         vault_auth_method=constants.VAULT_TOKEN,
-        namespace=constants.OPENSHIFT_STORAGE_NAMESPACE,
+        namespace=config.ENV_DATA["cluster_namespace"],
```
dahorak, line 908: and here

ocs_ci/helpers/helpers.py
dahorak: There are other similar cases (setting default value for function), so I would propose to change it there also.

conf/deployment/fusion_aas/managed_3az_provider_qe_3m_3w_m52x.yaml
```diff
@@ -3,6 +3,7 @@ ENV_DATA:
   platform: 'fusion_aas'
   deployment_type: 'managed'
   cluster_type: 'provider'
+  cluster_namespace: 'fusion-storage'
   multi_availability_zones: True
```
ebondare, line 7: If I understand correctly, namespace change to the config value isn't restricted to managed services but will happen to the whole ODF. Shouldn't we then include this value in the config files for other types of deployment?

fbalak: It is already part of a default config and on some places used for other platforms too. https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L135

dahorak (dismissed):

ebondare (dismissed):

yitzhak12 (approved):

jilju (dismissed):

fbalak:
New verification run: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/23648/

petr-balogh (approved):
Great. We need to then remove the value from the defaults to prevent others to still use this value and only rely on config value. But as we discussed offline we will do in separate PR to not break others if they have opened PRs using this value of namespace.

suchita-g (approved):
LGTM

jilju (approved):

dahorak (approved):

Issue #7240, opened 2023-03-07
test_worker_node_restart_during_pvc_expansion failed on setup with "ocs_ci.ocs.exceptions.TimeoutExpiredError:


keemano:
`test_worker_node_restart_during_pvc_expansion` fails in the setup with

```Message: failed on setup with "ocs_ci.ocs.exceptions.TimeoutExpiredError: Timed out after 360s running get("pod-test-cephfs-b048ddcef0f74a58868aab43-1-deploy", True, None)"
Type: None


Text:
self = <tests.manage.pv_services.pvc_resize.test_node_restart_during_pvc_expansion.TestNodeRestartDuringPvcExpansion object at 0x7fe5d877db20>
create_pvcs_and_pods = <function create_pvcs_and_pods..factory at 0x7fe618f750d0>
```

```
@pytest.fixture(autouse=True)
def setup(self, create_pvcs_and_pods):
    """
    Create PVCs and pods

    """
```

```
  self.pvcs, self.pods = create_pvcs_and_pods(
        pvc_size=4,
        num_of_rbd_pvc=12,
        num_of_cephfs_pvc=8,
        deployment_config=True,
    )

tests/manage/pv_services/pvc_resize/test_node_restart_during_pvc_expansion.py:53: 

```

```
tests/conftest.py:5768: in factory
    pod_obj = pod_factory(
tests/conftest.py:1063: in factory
    pod_obj = helpers.create_pod(
ocs_ci/helpers/helpers.py:391: in create_pod
    assert (ocp.OCP(kind="pod", namespace=namespace)).wait_for_resource(
ocs_ci/ocs/ocp.py:729: in wait_for_resource
    raise (ex)
ocs_ci/ocs/ocp.py:619: in wait_for_resource
    for sample in TimeoutSampler(

```

```
self = <ocs_ci.utility.utils.TimeoutSampler object at 0x7fe618866a00>


def __iter__(self):
    if self.start_time is None:
        self.start_time = time.time()
    while True:
        self.last_sample_time = time.time()
        if self.timeout <= (self.last_sample_time - self.start_time):

          raise self.timeout_exc_cls(*self.timeout_exc_args)
E               ocs_ci.ocs.exceptions.TimeoutExpiredError: Timed out after 360s running get("pod-test-cephfs-b048ddcef0f74a58868aab43-1-deploy", True, None)
```
[...]

PR #6124, opened 2022-07-04
Multi LVM snapshots and clones

shyRozen:
Multi snapshot test. Create 5 PVC, PODS, snapshots, restore PVC and attach to pod.
Need to merge [#6129](https://github.com/red-hat-storage/ocs-ci/pull/6129) to be able to execute PR validation

Signed-off-by: Shay Rozen <shay.rozen@gmail.com>

PR #7525, opened 2023-04-25
Fix deploy pod status value for DeploymentConfig workload in pod_factory_fixture

sidhant-agrawal:
Fixes: #7240

PR #6124 added `deploy_pod_status=status` when calling `create_pod()` in pod_factory_fixture
The status value which it was being set is same as other pods desired status (constants.STATUS_RUNNING).
This assumption that deploy pod status will be same as other pod status is not True and appears that it was added by mistake.
Also see this example output from ODF 4.13 setup where deploy pod goes to Completed state:
```
pod-test-cephfs-ed710c8134b6414ba84a2ff5-1-75rbh    1/1     Running     0          91s
pod-test-cephfs-ed710c8134b6414ba84a2ff5-1-deploy   0/1     Completed   0          92s
pod-test-rbd-2aaee60d8d0840d9a146af0e7d8-1-4p2pk    1/1     Running     0          29s
pod-test-rbd-2aaee60d8d0840d9a146af0e7d8-1-deploy   0/1     Completed   0          30s
```

Hence removing the param from inside the fixture. The `create_pod()` in helpers.py already uses correct value ,i.e., `deploy_pod_status=constants.STATUS_COMPLETED`

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1081,7 +1081,6 @@ def pod_factory_fixture(request, pvc_factory):
                 command=command,
                 command_args=command_args,
                 subpath=subpath,
-                deploy_pod_status=status,
             )
             assert pod_obj, "Failed to create pod"
         if deployment_config:


jilju (approved):

keesturam (approved):
LGTM

PR #7526, opened 2023-04-25
Add checks for managed-fusion resources

ebondare:


ebondare:
These checks were added in another PR

Issue #7498, opened 2023-04-19
Consume KUBEADMIN and KUBECONFIG environment variables by default


rabin-io:
Hi, 

As part of concertizing the tests, can you please refactor the code to use KUBEADMIN & KUBECONFIG environment variables by default if the cluster dir is not defined as cli argument?

This addition allows us to use the image in CI systems which inject this variables as part of the OCP installation process.

petr-balogh:
I will bring it out to tomorrow automation meeting. For now assigning to @OdedViner  :)

rabin-io:
@OdedViner What do you think?

PR #7538, opened 2023-04-27
Consume KUBEADMIN_PASSWORD and KUBECONFIG environment variables

OdedViner:
Added option to remove `cluster-path` parameter [from `run-ci` entrypoint], if there is an environment var named KUBECONFIG

#7498

### DIFF
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -497,7 +497,12 @@ def process_cluster_cli_params(config):
     suffix = ocsci_config.cur_index + 1 if ocsci_config.multicluster else ""
     cluster_path = get_cli_param(config, f"cluster_path{suffix}")
     if not cluster_path:
-        raise ClusterPathNotProvidedError()
+        if "KUBECONFIG" in os.environ:
+            cluster_path = os.path.dirname(os.environ["KUBECONFIG"])
+            ocsci_config.RUN["kubeconfig_location"] = "kubeconfig"
+            ocsci_config.RUN["password_location"] = "kubeadmin-password"
+        else:
+            raise ClusterPathNotProvidedError()
     cluster_path = os.path.expanduser(cluster_path)
     if not os.path.exists(cluster_path):
         os.makedirs(cluster_path)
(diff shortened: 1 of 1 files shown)


rabin-io:
The logic seems to be OK, we do assume both files will be on the same path, but that is reasonable.
My original thought was just setting the values from the ENV.

```python
            ocsci_config.RUN["kubeconfig_location"] = os.environ["KUBECONFIG"]
            ocsci_config.RUN["password_location"] = os.environ["KUBEADMIN_PASSWORD_FILE"]
```

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -497,7 +497,12 @@ def process_cluster_cli_params(config):
     suffix = ocsci_config.cur_index + 1 if ocsci_config.multicluster else ""
     cluster_path = get_cli_param(config, f"cluster_path{suffix}")
     if not cluster_path:
-        raise ClusterPathNotProvidedError()
+        if "KUBECONFIG" in os.environ:
```
dkeler, line 500: `if os.getenv("KUBECONFIG"):`

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -497,7 +497,12 @@ def process_cluster_cli_params(config):
...
+            cluster_path = os.path.dirname(os.environ["KUBECONFIG"])
```
dkeler, line 501: Do we need to check if `os.path.exists()`?

OdedViner:
> The logic seems to be OK, we do assume both files will be on the same path, but that is reasonable. My original thought was just setting the values from the ENV.
> 
> ```python
>             ocsci_config.RUN["kubeconfig_location"] = os.environ["KUBECONFIG"]
>             ocsci_config.RUN["password_location"] = os.environ["KUBEADMIN_PASSWORD_FILE"]
> ```

image
the password location and kubeconfig location are on the same path.. We need to change it in many places..

ocs_ci/framework/pytest_customization/ocscilib.py
OdedViner:
```oviner:~$ printenv | grep -i kube
KUBECONFIG=/home/oviner/ClusterPath/auth/kubeconfig
```
We are checking the value of `KUBECONFIG`

ocs_ci/framework/pytest_customization/ocscilib.py
OdedViner: done

rabin-io:
Also, you can't assume the filenames, as in the CI the files can be like `kubeadmin_<something>`, this why you need to use the reference from the environment variable as is.

petr-balogh (approved):

Issue #7516, opened 2023-04-24
Don't collect OCP must gathers when ODF must gather test fails


am-agrawa:
Discussed with Oded, reference- http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/j-090icm1r3-t1/j-090icm1r3-t1_20230404T212616/logs/failed_testcase_ocs_logs_1680647973/test_must_gather%5bOTHERS%5d_ocs_logs/

The test `test_must_gather[OTHERS]` failed but it collected OCP must gather which is of no use, and shouldn't be collected.

Message: Exception: Files don't exist:
['prometheus-k8s-scheduler-resources.yaml', 'prometheus-k8s-scheduler-resources.yaml']
Empty files:
[]

RUN ID: 1680647973

PR #7551, opened 2023-04-30
Don't collect OCP must gathers when ODF must gather test fails

OdedViner:
#7516


am-agrawa:
@OdedViner Could we merge this PR pls?

PR #7569, opened 2023-05-03
WIP: initial version of cumulative PVC acceptance tests

dahorak:
Related to [OCSQE-815](https://issues.redhat.com/browse/OCSQE-815) (and [AcceptanceSuiteImprovement sheet](https://docs.google.com/spreadsheets/d/1FFBbXOYej--m7n6ikm0nkdbix669wUB8VwQ8dNl5ZzI/edit#gid=0))

### DIFF
--- a/tests/functional/pv/pv_services/test_pvc_acceptance.py
+++ b/tests/functional/pv/pv_services/test_pvc_acceptance.py
@@ -6,8 +6,6 @@
 #     test_rwo_dynamic_pvc[CephFileSystem-Delete]
 #     test_rwx_dynamic_pvc[CephFileSystem-Retain]
 #     test_rwx_dynamic_pvc[CephFileSystem-Delete]
-#
-# ## partially implemented #########
 # tests/manage/pv_services/test_pvc_assign_pod_node.py::TestPvcAssignPodNode
 #     test_rwo_pvc_assign_pod_node[CephBlockPool]
 #     test_rwo_pvc_assign_pod_node[CephFileSystem]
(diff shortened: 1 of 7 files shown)


dahorak:
Before merging this, we have to remove `acceptance` marker from the original tests.

tests/functional/pv/pv_services/test_pvc_assign_pod_node.py
```diff
@@ -49,7 +48,6 @@ def verify_access_token_notin_odf_pod_logs(self):
             error_msg in pod_log
         ), f"Logs should not contain the error message '{error_msg}'"
 
-    @acceptance
```
petr-balogh, line 52: Should we mention in the comment that this test is also covered in acceptance test suite but in different approach and link to the test file?
Also would mention that if someone will modify this test case, to don't forgot to update the logic also in the acceptance suite test if needed.

dahorak: Good point, added comments for each testcase.

tests/functional/pv/pv_services/test_pvc_acceptance.py
```diff
@@ -0,0 +1,828 @@
...
+from ocs_ci.utility.utils import TimeoutSampler
+
+logger = logging.getLogger(__name__)
+
+
+@green_squad
+@acceptance
+class TestPvcAcceptance(ManageTest):
+    """
+    Acceptance tests for PVC
```
hnallurv, line 72: It would be great if you can share the improvement in the execution time compared to old implementation.

dahorak: I tried to gather some summary from last bunch of executions and the average execution time of the original tests covered by this one cumulative tests takes around 35 minutes (30 - 40 minutes), while few of the executions of the cumulative test takes around 17 minutes (15 - 19 minutes, I saw some executions with 12 and 13 minutes).

But I would not consider this as the fully representative statistics, we will need more executions and maybe also compare the full acceptance tests execution time.

PR #7573, opened 2023-05-03
Updating ManagedFusionOffering DFC template

jilju:
Removing unnecessary quotes which is creating an addition space in the value of dataFoundationProviderEndpoint.


jilju:
fixed in code.

PR #7584, opened 2023-05-04
Fix post install verification for FaaS consumer

jilju:
Update post installation checks to support FaaS consumer cluster.

1. Checks which require tools pods are skipped because tools pod is not available in consumer cluster. This will be updated when there is a solution to check ceph commands from consumer.
2. Updated deploy_ocs function in ocs_ci/deployment/fusion_aas.py to skip ocs install if managedfusionoffering is present instead of checking cehph cluster.
3. OCS operator, storagecluster, cephcluster will not be present in FaaS consumer cluster. Checks related to this will be skipped.
4. Added verification of pods in managed-fusion namespace.

### DIFF
--- a/ocs_ci/deployment/fusion_aas.py
+++ b/ocs_ci/deployment/fusion_aas.py
@@ -103,10 +103,12 @@ class FUSIONAAS(rosa_deployment.ROSA):
         """
         Deployment of ODF Managed Service addon on Fusion aaS.
         """
-        ceph_cluster = ocp.OCP(kind="CephCluster", namespace=self.namespace)
+        managed_fusion_offering = ocp.OCP(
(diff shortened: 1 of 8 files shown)


jilju:
Verified: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/24062

jilju:
Verified deployment on vmware platfrom to ensure that the code change if not affecting other platforms
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/24066/

suchita-g (approved):

dahorak (approved):

PR #7626, opened 2023-05-11
Add catsrc and subscription checks

ebondare:


ocs_ci/helpers/managed_services.py
```diff
@@ -456,6 +458,22 @@ def verify_faas_provider_resources():
     # Verify that the security groups are set up correctly
     assert verify_worker_nodes_security_groups()
 
+    # Check the presence of catalogsource and its state
+    catsrc = OCP(kind=constants.CATSRC, namespace=config.ENV_DATA["cluster_namespace"])
+    catsrc_info = catsrc.get().get("items")[0]
```
petr-balogh, line 463: How sure can we be that the Fusion Agent will be always as first item in those catalog sources?

Usually we use "openshift-maketplace" NS . Not sure how it is for the fusion, if it's in cluster_namespace and if only one catalog source is added at the same time?

If there is possibility that there will be more of them
I would iterate over all catalog sources and check if at least one of them starts with("Managed Fusion Agent") 
And if so then check if it's in ready state and do the check on the state via assert?

But if we are sure there is only one and in the future there will not be any other then we are OK probably.

ocs_ci/helpers/managed_services.py
```diff
@@ -456,6 +458,22 @@ def verify_faas_provider_resources():
...
+    log.info(f"Catalogsource: {catsrc_info}")
+    assert catsrc_info["spec"]["displayName"].startswith("Managed Fusion Agent")
+    assert catsrc_info["status"]["connectionState"]["lastObservedState"] == "READY"
+
+    # Check the presence of subscription
+    subscr = OCP(
+        kind=constants.CATSRC,
+        namespace=config.ENV_DATA["cluster_namespace"],
+        selector="operators.coreos.com/managed-fusion-agent.managed-fusion",
```
petr-balogh, line 472: Are we going to use this selector in other places as well?
If so, I suggest to move it to constants.

ocs_ci/helpers/managed_services.py
```diff
@@ -456,6 +458,28 @@ def verify_faas_provider_resources():
     # Verify that the security groups are set up correctly
     assert verify_worker_nodes_security_groups()
 
+    # Check the presence of catalogsource and its state
+    catsrc = OCP(kind=constants.CATSRC, namespace=config.ENV_DATA["service_namespace"])
+    catsrc_info = catsrc.get().get("items")[0]
```
jilju, line 463: Please try toe fetch the required catalogsource by name or other identifier. "Managed Fusion Agent" need not be the first item in the list. This works currenly becaue only one item will be present.

ebondare: Fixed

ebondare:
Successful test run: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/24905/console

jilju (approved):

suchita-g (approved):

PR #7638, opened 2023-05-16
Disregard the users array of noobaa-db SCC at test_db_scc

### LINKED ISSUE #7637: test_db_scc no longer needs to delete the user array in noobaa-db scc
It seems that in 4.13 the user array in the noobaa-db scc is empty by default, so this test currently fails while trying to delete (and later on restore) this array. 

According to https://url.corp.redhat.com/14a21df this step looks like a prerequisite for the noobaa-db pod to use the anyuid scc instead of the noobaa-db scc after restarting, and manual testing on 4.13 shows that the switch is possible without this step.

sagihirshfeld:
Fixes #7637

### DIFF
--- a/tests/manage/mcg/test_mcg_resources_disruptions.py
+++ b/tests/manage/mcg/test_mcg_resources_disruptions.py
@@ -197,17 +197,7 @@ class TestMCGResourcesDisruptions(MCGTest):
                     f'"value": "{service_account}"}}]',
                     format_type="json",
                 )
-            if not helpers.validate_scc_policy(
-                sa_name=scc_name,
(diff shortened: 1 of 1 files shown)


keesturam:
@sagihirshfeld I am assuming that this PR won't be backported to 4.12 and backwards but will be in 4.13 and beyond?

sagihirshfeld:
> @sagihirshfeld I am assuming that this PR won't be backported to 4.12 and backwards but will be in 4.13 and beyond?

Correct, in earlier versions we still need to modify the user array at the noobaa-db scc

ebenahar (approved):

PR #7653, opened 2023-05-18
Increased timeout to add node to 900sec after machineset creation

ramkiperiy:
Observed most of the Azure runs failed to add node in 600sec due to this increased the timeout to 900sec since part of machineset creation node add passed within 900sec.


pintojoy (approved):
LGTM

keesturam (approved):

Issue #7659, opened 2023-05-18
`test_noobaa_kms_validation` fails when Thales is the KM provider 


sagihirshfeld:
The noobaa-operator pod logs show logs of integration, but they're a bit different than those that are used with KMS:

```
time="2023-05-18T14:25:56Z" level=info msg="✅ Exists:  \"ocs-thales-kmip-c46b5b27fa9e4492982ead43\"\n"
time="2023-05-18T14:25:56Z" level=info msg="setKMSConditionType kmip" sys=openshift-storage/noobaa
``` 

Examples:
- https://url.corp.redhat.com/6b4f782

PR #7660, opened 2023-05-18
`test_noobaa_kms_validation`: Remove the use of RE and generalize the search based on environment config

sagihirshfeld:
Fixes: #7659

Looking for "setKMSConditionType <KM type>" should be sufficient, based on the printed logs from past runs and the following code from the noopaa-operator project:
https://github.com/noobaa/noobaa-operator/blob/b45e8b619bca06f91586fdbd523ecbcf55203824/pkg/system/phase2_creating.go#LL830C1-L852C1

And since we're specifying this type under ENV_DATA.KMS_PROVIDER in our conf files, looking for `setKMSConditionType + config.ENV_DATA["KMS_PROVIDER"]` should work for both of the current KM providers as well as future ones.


keesturam (dismissed):
LGTM

tiffanyn108 (approved):

keesturam (approved):

PR #7668, opened 2023-05-22
Set PagerDuty service for Fusion aaS use

fbalak:


fbalak:
Verified: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/24910/
The test failed but it is because of a bug. PagerDuty setting was in accordance with a guide.

ocs_ci/utility/pagerduty.py
```diff
@@ -247,15 +287,21 @@ def get_service_dict(self):
         cluster_name = config.ENV_DATA["cluster_name"]
         # timestamp is added to service name to ensure unique name of service
         timestamp = time.time()
-        default_policy = self.get_default_escalation_policy_id()
+        policy = None
+        policy_name = config.AUTH["pagerduty"].get("escalation_policy")
+        if policy_name:
```
ebondare, line 292: And if this condition is not satisfied, would it make sense to raise some meaningful error?

ocs_ci/utility/pagerduty.py
```diff
@@ -247,15 +287,21 @@ def get_service_dict(self):
...
+            policy = self.get_escalation_policy_id(policy_name)
+            if not policy:
+                raise ValueError(f"Policy {policy_name} not found")
+        policy = policy
```
ebondare, line 296: `policy = policy` line doesn't really make sense to me, could you please explain?

fbalak: You are correct. I will delete it.

ocs_ci/utility/pagerduty.py
fbalak: Indentation updated so that ValueError is raised even when the value is None

ocs_ci/utility/pagerduty.py
```diff
@@ -45,6 +47,41 @@ def set_pagerduty_integration_secret(integration_key):
...
+    """
+    Update managed-fusion-agent-config secret. This is valid only on Fusion aaS platform.
+    managed-fusion-agent-config secret is expected to be present prior to the update.
+
+    Args:
+        integration_key (str): Integration key taken from PagerDuty Prometheus integration
+
+    """
+    logger.info("Setting up PagerDuty")
+    kubeconfig = os.getenv("KUBECONFIG")
```
jilju, line 60: I am not sure, but does this work in a multicluster configuration ?
Alternative:
```
kubeconfig = os.path.join(
            config.ENV_DATA["cluster_path"], config.RUN.get("kubeconfig_location")
        )
```

fbalak: I am not sure but those tests are currently not working correctly for multicluster. I am planning to update those tests for multicluster and to be run with other tests from consumer which is not implemented yet in a future PR.

ocs_ci/utility/pagerduty.py
```diff
@@ -45,6 +47,41 @@ def set_pagerduty_integration_secret(integration_key):
...
+    secret_data["metadata"].pop("resourceVersion", None)
+    secret_data["metadata"].pop("uid", None)
+
+    with tempfile.NamedTemporaryFile(
+        mode="w+", prefix=f"{constants.FUSION_AGENT_CONFIG_SECRET}_"
+    ) as secret_file:
+        yaml.dump(secret_data, secret_file)
+        secret_file.flush()
+        exec_cmd(f"oc apply --kubeconfig {kubeconfig} -f {secret_file.name}")
+    logger.info("New PagerDuty service was set.")
```
yitzhak12, line 82: It could be helpful to add a comment or a log message that describes the process here.

yitzhak12 (commented):
Looks good

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -245,6 +245,8 @@ UPGRADE:
 # or ping in ocs infra chat room.
 AUTH:
   test_quay_auth: test_secret
+  pagerduty:
+    escalation_policy: "Default"
```
ebondare, line 249: Does it really belong in the AUTH section of the config file?

fbalak: It is a name of escalation policy related to our account (with rules who will get the pager). It is not a token but I think that it could be kept there as it is an account related info and we have a pagerduty section in the config.

ebondare (approved):

ebondare (dismissed):

ocs_ci/utility/pagerduty.py
fbalak: The log was expanded.

ebondare (approved):

dahorak (approved):

Issue #7667, opened 2023-05-22
Add peer ready check in RDR UI tests when regression BZ2208637 is fixed


am-agrawa:

am-agrawa:
Closing this issue as the BZ https://bugzilla.redhat.com/show_bug.cgi?id=2208637 has been closed without the fix.

am-agrawa:
Will send a fix by removing existing check

PR #7676, opened 2023-05-22
Remove peer ready check and allow option to select subscription if not auto-selected

am-agrawa:
This PR fixes issue #7667, the bug was closed as not a bug so the existing validation for Peer ready is being removed.
Also, ACM UI auto-selects subscription to trigger failover/relocate. ​If auto-selection is not enabled, the PR allows ability to first select the subscription and then triggers the operation.


am-agrawa:
Negative failover relocate- http://pastebin.test.redhat.com/1100664

PR #7710, opened 2023-05-29
Enhance the `ignore_leftover_label` to support labels other than app

sagihirshfeld:
This mark currently only ignores resources by their app labels. For example:
```
@ignore_leftover_label("some-app")
class TestSomeClass(BaseTest)
```
Will ignore only leftovers if `.metadata.labels.app == some-app` is in their YAML file.

This enhancement allows usage of other labels via the "label-key=label-value" syntax which we already use in other parts of our project. 

For example, the above example is now implemented via:
```
@ignore_leftover_label("app=some-app")
class TestSomeClass(BaseTest)
```

But now we can ignore other labels as well. For example, the following will ignore the noobaa-endpoint pods leftovers:
```
@ignore_leftover_label("noobaa-s3=noobaa")
class TestSomeClass(BaseTest)
```


sagihirshfeld:
Note that `tests/test_skip_labeled_leftover.py` is only here to demonstrate functionality for reviewers, and it should be removed before merging

sagihirshfeld:
Note that in [this validation run ](https://url.corp.redhat.com/140c0ea) both tests leave a leftover pod, but the test with the `ignore_leftover_label` mark successfully ignores the check and passes, while the unmarked test fails as expected

OdedViner (dismissed):

OdedViner (approved):

OdedViner (approved):

ocs_ci/ocs/constants.py
```diff
@@ -359,7 +359,7 @@
 S3CLI_LABEL = "app=s3cli"
 PROVIDER_SERVER_LABEL = "app=ocsProviderApiServer"
 PROMETHEUS_OPERATOR_LABEL = "app.kubernetes.io/name=prometheus-operator"
-MUST_GATHER_HELPER_LABEL = "must-gather-helper-pod"
+MUST_GATHER_HELPER_LABEL = "must-gather-helper-pod="
```
sagihirshfeld, line 362: This label appears as `"must-gather-helper-pod": ""` in the YAML of the pod

yitzhak12 (approved):

ebenahar (approved):

jilju (approved):

Issue #7716, opened 2023-05-30
test_nsfs.py: "Error is chmod: changing permissions of '/nsfs': Permission denied"


sagihirshfeld:
I recently noticed that this issue isn't related to https://bugzilla.redhat.com/show_bug.cgi?id=2165817. 

When an NSFS namespacestore is created, the noobaa-endpoint pods are restarting to mount the PVC. This issue occurred when attempting to access the PVC via the interrface pod while these pods still weren't running, and most likely in the process of mounting the PVC and handling permissions.

This would also explain why this error is not consistent: sometimes our NSFS tests fail during setup because of this issue, and when the race condition isn't met the test continues to fail due to the BZ

PR #7647, opened 2023-05-17
NSFS noobaa-endpoint SCC workaround 

sagihirshfeld:
This is a temporary workaround which changes the noobaa-endpoint scc back to the way it was before 4.12:

```
seLinuxContext:
  type: RunAsAny --> MustRunAs

users: [] --> users:
	             - system:serviceaccount:openshift-storage:noobaa-endpoint
```
The new fixture `modify_endpoint_scc` patches the noobaa-endpoint SCC before running any of the NSFS tests, and restores it back to the way it was after these tests. 

This PR also fixes: https://github.com/red-hat-storage/ocs-ci/issues/7716

PR #7717, opened 2023-05-30
test_nsfs.py: Wait for the nooba-endpoint pods to reset and mount the PVC at setup

sagihirshfeld:
Fixes: #7716

Adds a wait period for the noobaa-endpoint pods to stabilize before attempting to modify the PVC they're mounting


sagihirshfeld:
Moved the fix to #7647

Issue #7769, opened 2023-06-09
Engineering team is failing in upgrade scenario with applying ICSP


petr-balogh:
@b-ranto told me they see such issue in their upgrade pipeline.

```E               ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc image extract --registry-config /home/jenkins-build/workspace/ocs-ci/ocs-ci.git/data/pull-secret :4.13.0-218 --confirm --path /icsp.yaml:../cluster/icsp-1686294458.
E               Error is error: ":4.13.0-218" is not a valid image reference: invalid reference format
```

PR #7770, opened 2023-06-09
Fix issue with applying ICSP for upgrade scenario

petr-balogh:
Fixes: #7769


dahorak (approved):

b-ranto (approved):
It looks like this helped us resolve the upgrade issue in our CI.

petr-balogh:
Verified here:
 https://url.corp.redhat.com/c24a66d

PR #7836, opened 2023-06-16
Apply ICSP also for upgrade scenario

petr-balogh:


vavuthu (approved):

b-ranto (approved):
This was tested in our CI and the job was able to pass the upgrade so the new ICSP rules must have been applied properly this time.

PR #7845, opened 2023-06-19
Add post cluster validation to avoid failure in test_cephfs_many_files.py script

ramkiperiy:
Test run fails for 2 tests due to ceph health not in good shape but ceph recovers in some minutes, observed post TC there is no check or wait for ceph cluster to be back OK state, added those validations


pintojoy (approved):
LGTM

tiffanyn108 (approved):

PR #7847, opened 2023-06-19
PR to remove memory leak function

ramkiperiy:
Removed memory leak code from the test case, with memory leak function we haven't captured any issues, also due to environmen and threading it's very flaky.


pintojoy (approved):
LGTM

tiffanyn108 (approved):

PR #7865, opened 2023-06-20
Increase timeout for pods to reach running state in test_check_pods_status_after_node_failure

am-agrawa:
The test has been repeatedly failing with 
Message: AssertionError: Increased timeout from 480 to 600 seconds, The pods are not 'Running' even after 600 seconds
assert False
RP- https://url.corp.redhat.com/

But the Pods reach running state in collected must gather and ceph is also healthy. We should allow some more time for the checks to ensure Pods reaching running state before failing the test.


am-agrawa:
Product bug- https://bugzilla.redhat.com/show_bug.cgi?id=2216803

AviadP (approved):

am-agrawa:
Test is now stable, this PR is no longer needed.

PR #7795, opened 2023-06-13
Adjust send_rpc_query to send RPC requests via the NooBaa CLI

sagihirshfeld:
In 4.14 the URL that was used to make NooBaa RPC requests has been removed, so this PR adjusts the method that was used to make these calls, so it uses the NooBaa CLI instead.

See this google chat thread for details:
https://url.corp.redhat.com/c4c3508

PR #7875, opened 2023-06-21
Choose to use only the MCG-CLI for RPC calls elsewhere

sagihirshfeld:
In 4.14 RPCs are only possible via the MCG-CLI, but in #7795 we kept making RPCs against the noobaa-endpoint in lower versions just to keep covering the noobaa-mgmt route.

This PR separates the functionality issue from the coverage issue by introducing a dedicated test for the noobaa-mgmt route, and choosing to use the MCG-CLI for RPCs in all other cases.

### DIFF
--- a/tests/manage/mcg/test_nb_mgmt_endpoint.py
+++ b/tests/manage/mcg/test_nb_mgmt_endpoint.py
@@ -1,0 +1,50 @@
+import json
+import requests
+import logging
+
+from ocs_ci.framework.testlib import tier1
+
+from ocs_ci.framework.testlib import MCGTest
+from ocs_ci.framework.testlib import skipif_ocs_version
+
+logger = logging.getLogger(name=__file__)
+
+
+@tier1
+@skipif_ocs_version("<4.14")
+class TestNoobaaMgmtEndpoint(MCGTest):
+    """
+    Test the noobaa mgmt route functionality
+    """
+
+    def test_noobaa_mgmt_route(self, mcg_obj_session):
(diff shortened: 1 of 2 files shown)


tiffanyn108 (approved):

tests/manage/mcg/test_nb_mgmt_endpoint.py
```diff
@@ -0,0 +1,63 @@
...
+from ocs_ci.framework.testlib import MCGTest
+from ocs_ci.framework.testlib import skipif_ocs_version
+
+from ocs_ci.ocs.bucket_utils import retrieve_verification_mode
+
+logger = logging.getLogger(name=__file__)
+
+
+@tier1
+@skipif_ocs_version(">4.13")
```
mashetty330, line 16: why dont we raise this as seperate PR against release branches <= 4.12 ?

sagihirshfeld: Thanks for brining my attention to this - this PR was created when 4.14 and 4.13 used the same OCS-CI branch so the test's purpose was to keep covering the RPC route in 4.13. However, now that they're separated we no longer need this test in 4.14, just 4.13 (in 4.12 and under we still use HTTPS calls to the RPC route instead of the MCG-CLI, so don't need that test there).

I'll start by removing this test from this PR.

petr-balogh (approved):

tiffanyn108 (approved):

PR #7882, opened 2023-06-21
Update ACM version to 2.8 GAed

prsurve:


am-agrawa (approved):

prsurve:
marking as verifed as the acm deployment passed but when running deployment test it failed creating acm as it was already installed

petr-balogh (approved):

PR #7900, opened 2023-06-23
Automate bug 2168788

mashetty330:
bug: https://bugzilla.redhat.com/show_bug.cgi?id=2168788

**needs backport?**
yes release-4.13

### DIFF
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -878,6 +878,38 @@ def get_noobaa_operator_pod(
     return noobaa_operator_pod
 
 
+def get_noobaa_core_pod():
+    """
+    Fetches Noobaa core pod details
+
+    Returns:
+        Pod object: Noobaa core pod object
+
+    """
+    noobaa_core = get_pods_having_label(
+        label=constants.NOOBAA_CORE_POD_LABEL,
+        namespace=config.ENV_DATA["cluster_namespace"],
+    )
+    noobaa_core_pod = Pod(**noobaa_core[0])
+    return noobaa_core_pod
+
+
+def get_noobaa_endpoint_pods():
+    """
+    Fetches noobaa endpoint pod details
+
+    Returns:
+        List: List containing noobaa endpoint pod objects
+
+    """
+    noobaa_endpoints = get_pods_having_label(
+        label=constants.NOOBAA_ENDPOINT_POD_LABEL,
+        namespace=config.ENV_DATA["cluster_namespace"],
+    )
+    noobaa_endpoint_pods = [Pod(**pod) for pod in noobaa_endpoints]
+    return noobaa_endpoint_pods
+
+
 def get_odf_operator_controller_manager(
     ocs_label=constants.ODF_OPERATOR_CONTROL_MANAGER_LABEL, namespace=None
 ):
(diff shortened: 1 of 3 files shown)


tests/manage/mcg/test_bucket_replication.py
```diff
@@ -368,6 +409,76 @@ def test_unidirectional_bucketclass_replication(
...
+                f"stopped from {source_bucket_name} to {target_bucket_name}"
+            )
+            sample = TimeoutSampler(timeout=900, sleep=10, func=check_noobaa_core_logs)
+            assert sample.wait_for_func_status(
+                result=True
+            ), f"Replication hasn't stopped between {source_bucket_name} to {target_bucket_name}"
+            logger.info(
+                f"Replication has stopped from {source_bucket_name} to {target_bucket_name}"
+            )
+
```
sagihirshfeld, line 481: We should move this under a separate test method - even if it means to copy the setup from `test_unidirectional_bucketclass_replication`. 

The original method has about 30 lines of code, while the coverage of this BZ adds about 80 that are only specific to one of the params (`AWStoDefault-BS-Hybrid`). I think this justifies the separation for the sake of readability and maintainability.

ocs_ci/ocs/resources/pod.py
```diff
@@ -878,6 +878,24 @@ def get_noobaa_operator_pod(
...
+
+def get_noobaa_endpoint_pods(
+    ocs_label=constants.NOOBAA_ENDPOINT_POD_LABEL,
+    namespace=config.ENV_DATA["cluster_namespace"],
+):
+    noobaa_endpoints = get_pods_having_label(ocs_label, namespace)
+    noobaa_endpoint_pods = [Pod(**pod) for pod in noobaa_endpoints]
+    return noobaa_endpoint_pods
+
+
```
sagihirshfeld, line 898: Please add Docstrings - I would mainly be interested in the return types (instance/s of the Pod class) since they're different than the return type of `get_pods_having_label`.

Also, I suggest not using `ocs_label` as an argument in these methods because it allows this kind of confusing usage:
```
nb_db_pod = get_noobaa_core_pod(ocs_label=constants.NOOBAA_DB_LABEL_47_AND_ABOVE)
```

tests/conftest.py
```diff
@@ -6337,3 +6338,32 @@ def teardown():
...
+
+    def finalizer():
+        level = "default_level"
+        noobaa_cm.patch(
+            params=f'{{"data": {{"NOOBAA_LOG_LEVEL": "{level}"}}}}', format_type="merge"
+        )
+        wait_for_pods_to_be_running(pod_names=[pod.name for pod in get_noobaa_pods()])
+
+    request.addfinalizer(finalizer)
+    return factory
```
sagihirshfeld, line 6369: Nice! This will definitely prove useful in the future! 👍

tests/manage/mcg/test_bucket_replication.py
mashetty330: @sagihirshfeld The reason why i added this coverage to the existing code is to have this coverage for all the type of params not just `AWStoDefault-BS-Hybrid`. we check replication has stopped for all the types of buckets here.

sagihirshfeld: Oh I must've missed the "else" part here in the original review, this is ok then.

sagihirshfeld (dismissed):
LGTM

parikshithb (dismissed):

PrasadDesala (dismissed):

sagihirshfeld (approved):
LGTM

parikshithb (approved):

keesturam (approved):

PR #7906, opened 2023-06-25
Verify multus networks after add capacity

OdedViner:


OdedViner:
@ebenahar We dont test 'add capacity' on BM, and we don't run multus on vsphere.. So I can't test this PR...maybe run it on dev mode?

PR #7907, opened 2023-06-25
Verify multus on Disk Replacement

OdedViner:


OdedViner:
@ebenahar  We dont test disk replacement on BM, and we don't run multus on vsphere.. So I can't test this PR...maybe run it on dev mode?

PR #7908, opened 2023-06-25
Verify multus on Node Replacement proceudre

OdedViner:


OdedViner:
@ebenahar We dont test node replacement on BM, and we don't run multus on vsphere.. So I can't test this PR...maybe run it on dev mode

PR #7909, opened 2023-06-25
Add ceph health check for the Faas consumer cluster in the utility "ceph_health_check_base"

yitzhak12:
In the pr, I implement the following:

- Create two functions to check if the cluster is a Faas provider or consumer cluster.
- Divide the utility function `ceph_health_check_base` into small parts, making the code more clean and easy to modify. The basic ceph health check commands are still the same.
- Add ceph health check for the Faas consumer cluster because it behaves differently than a regular cluster.

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -50,6 +50,7 @@ from ocs_ci.ocs.exceptions import (
     UnsupportedOSType,
     InteractivePromptException,
     NotFoundError,
+    CephToolBoxNotFoundException,
 )
 from ocs_ci.utility import version as version_module
 from ocs_ci.utility.flexy import load_cluster_info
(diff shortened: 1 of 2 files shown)


yitzhak12:
I ran pr validation jobs with Faas consumer and provider clusters: One run when the provider is the first cluster: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/1777/, and another run when the consumer is the first cluster:  https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/1778/

ocs_ci/utility/utils.py
```diff
@@ -2262,6 +2247,27 @@ def ceph_health_check_base(namespace=None):
         raise CephHealthException(f"Ceph cluster health is not OK. Health: {health}")
 
 
+def wait_for_rook_ceph_tools_pod(namespace, timeout=300):
```
petr-balogh, line 2250: Is this function necessary? Shouldn't this be covered already in:
wait_for_rook_ceph_tools_pod 
?

ocs_ci/utility/utils.py
```diff
@@ -2282,6 +2288,113 @@ def create_ceph_health_cmd(namespace):
...
+        namespace: Namespace of OCS
+
+    Raises:
+        CommandFailed: In case the rook-ceph-tools pod failed to reach the Ready state.
+
+    Returns:
+        str: The output of the ceph health command
+
+    """
+    wait_for_rook_ceph_tools_pod(namespace)
```
petr-balogh, line 2305: Shouldn't this work?
```python
return = ct_pod.exec_ceph_cmd(ceph_cmd="ceph health", format="plain")
```

ocs_ci/utility/utils.py
```diff
@@ -2282,6 +2288,113 @@ def create_ceph_health_cmd(namespace):
...
+    )
+
+    wait_for_rook_ceph_tools_pod(namespace, timeout=120)
+    ceph_health_cmd = create_ceph_health_cmd(namespace)
+    try:
+        health = run_cmd(ceph_health_cmd)
+    except CommandFailed as ex:
+        if is_rados_connect_error_in_ex(ex):
+            log.info("Patch the consumer rook-ceph-tools deployment")
+            patch_consumer_toolbox()
```
petr-balogh, line 2388: I think that we should directly update the function:
https://github.com/red-hat-storage/ocs-ci/blob/d322ca8ea7d179e229319e6faaa9739432022c63/ocs_ci/ocs/resources/pod.py#L690
Which will do patch on consumer toolbox pod. WDYT?

If we can get info that the command is running on cluster which is consumer I think it can be without any additional parameter, or add one more parameter to be able to enable such patching for consumer cluster for toolbox pod.

ocs_ci/utility/utils.py
yitzhak12: Sorry, I didn't get the comment. Do you say that this function already exists somewhere? Or do we not need it?

ocs_ci/utility/utils.py
yitzhak12: Yes, it should work. The thing is that it will change the steps of the function `ceph_health_check_base`. Because in the process `ceph_health_check_base`, we use `oc` commands. From my side, we can replace it with the existing function `get_ceph_tools_pod`. The question is if there was a reason for using these `oc` commands, so we will not get regression failures.

ocs_ci/utility/utils.py
yitzhak12: It should be fine, but I am not 100% sure because the function returns the rook-ceph-tools pod and doesn't say anything about patching. IMO, it will be ok to add the patching as you suggested.

petr-balogh: In this function we call `setup_ceph_toolbox`

https://github.com/red-hat-storage/ocs-ci/blob/d322ca8ea7d179e229319e6faaa9739432022c63/ocs_ci/ocs/utils.py#L776

You can see we already do something similar for external mode deployment, so I think that the patch logic when we create new ceph_toolbox_pod should be done in this function.

ocs_ci/utility/utils.py
petr-balogh: Sorry, wrong copy paste, meant: `get_ceph_tools_pod`. Updated the comment.

ocs_ci/utility/utils.py
petr-balogh: Didn't get the thing with `oc` commands, can you please elaborate?

yitzhak12: Yes, I will be more clear. In the original function, we execute the code: 

> run_cmd(
        f"oc wait --for condition=ready pod "
        f"-l app=rook-ceph-tools "
        f"-n {namespace} "
        f"--timeout=300s"
    )
    ceph_health_cmd = create_ceph_health_cmd(namespace)
    run_cmd(ceph_health_cmd)

And now, we will use the function `get_ceph_tools_pod` instead. So, the question is if it will work correctly in all the tests.

ocs_ci/utility/utils.py
yitzhak12: We don't wait for the `rook-ceph-tool` pod to be ready in this function. But we may add this option.

yitzhak12:
I raised a new PR https://github.com/red-hat-storage/ocs-ci/pull/7946 since the changes discussed above in the current pr make it all different.

PR #7918, opened 2023-06-27
Use latest image tag for ACM unreleased Install

prsurve:


ocs_ci/deployment/deployment.py
```diff
@@ -2037,6 +2040,29 @@ def get_and_apply_icsp_from_catalog(image, apply=True, insecure=False):
...
+
+
+    """
+    response = requests.get(
+        "https://quay.io/api/v1/repository/acm-d/acm-custom-registry/tag/"
+    )
+    responce_data = response.json()
+    for data in responce_data["tags"]:
+        if version in data["name"] and "v" not in data["name"]:
+            return logger.info(data["name"])
```
petr-balogh, line 2061: In [4]: type(logger.info("Hello"))
Out[4]: NoneType

This doesn't look good to me.

You should separate the log and return the data["name"] on separate line.

prsurve: I had added for my testing fixing in next comment

ocs_ci/deployment/deployment.py
```diff
@@ -1732,9 +1734,12 @@ def deploy_acm_hub_unreleased(self):
...
 
         logger.info("Writing tag data to snapshot.ver")
-        image_tag = config.ENV_DATA.get(
-            "acm_unreleased_image", config.ENV_DATA.get("default_acm_unreleased_image")
-        )
+        acm_version = config.ENV_DATA.get("acm_version")
+
+        image_tag = get_latest_acm_tag_unreleased(version=acm_version)
+        if not image_tag:
+            raise TagNotFoundException("Couldn't find given tag!")
```
petr-balogh, line 1757: I would update the function `get_latest_acm_tag_unreleased` itself to raise an exception if it will not find a tag and document it in raises block in docstring.

ocs_ci/deployment/deployment.py
```diff
@@ -2037,6 +2042,29 @@ def get_and_apply_icsp_from_catalog(image, apply=True, insecure=False):
...
+    """
+    response = requests.get(
+        "https://quay.io/api/v1/repository/acm-d/acm-custom-registry/tag/"
+    )
+    responce_data = response.json()
+    for data in responce_data["tags"]:
+        if version in data["name"] and "v" not in data["name"]:
+            return logger.info(data["name"])
+
+    return False
```
petr-balogh, line 2081: You are describing you are returning string but here it's boolean. Please rather just return string and in case it fails to find the tag, raise an exception as suggested.

conf/ocsci/acm_2.8.yaml
```diff
@@ -0,0 +1,3 @@
+---
+ENV_DATA:
+  acm_version: "2.8"
```
petr-balogh, line 3: missing new line at the end of file, the same for the other config.

petr-balogh, line 3: Also I think we should move those values also to OCP config file and the here you should add also channel I think. Because if you will have different version in ocp config then you will have mismatch.

https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/ocp_version/ocp-4.13-config.yaml#L13C7-L13C31

prsurve: We already have in the OCP config but that represents GA values

ocs_ci/deployment/deployment.py
prsurve: Done

ocs_ci/deployment/deployment.py
prsurve: Done

conf/ocsci/acm_2.8.yaml
prsurve: done

ocs_ci/deployment/deployment.py
```diff
@@ -2053,6 +2056,33 @@ def get_and_apply_icsp_from_catalog(image, apply=True, insecure=False):
     return icsp_file_dest_location
 
 
+def get_latest_acm_tag_unreleased(version):
```
petr-balogh, line 2059: Please move it out of deployment.py module to ocs_ci/utility/utils.py where we have also get_latest_ds_olm_tag .

prsurve: I tried consuming get_latest_ds_olm_tag but the function did not worked for me

ocs_ci/utility/utils.py
```diff
@@ -4281,3 +4281,30 @@ def list_insert_at_position(lst, index, element):
...
+    response = requests.get(
+        "https://quay.io/api/v1/repository/acm-d/acm-custom-registry/tag/"
+    )
+    responce_data = response.json()
+    for data in responce_data["tags"]:
+        if version in data["name"] and "v" not in data["name"]:
+            log.info(f"Found Image Tag {data['name']}")
+            return data["name"]
+
+    raise TagNotFoundException("Couldn't find given tag!")
```
shylesh, line 4310: @prsurve , could you please add "ACM" keyword in the exception message so that its clear what tag we are looking for.

petr-balogh (approved):

prsurve:
Both the PR were able to deploy the ACM hub but failed because later

PR #7926, opened 2023-06-28
Automating GSS BZ https://bugzilla.redhat.com/show_bug.cgi?id=2179271

Shrivaibavi:


sagihirshfeld (approved):
LGTM

tests/manage/mcg/test_bucket_creation.py
```diff
@@ -178,3 +180,27 @@ def test_duplicate_bucket_creation(
...
+
+        bucket_factory(amount, interface)
+
+        # Check rook-operator pod logs for the error msg
+        unexpected_log = 'malformed BucketHost "s3.openshift-storage.svc": malformed subdomain name "s3"'
+        rook_op_pod = get_operator_pods()
+        pod_log = get_pod_logs(pod_name=rook_op_pod[0].name)
+        assert not (
+            unexpected_log in pod_log
+        ), "No malformed BucketHost, subdomain found"
```
parikshithb, line 206: Add correct assert message when log is found.

tests/manage/mcg/test_bucket_creation.py
```diff
@@ -178,3 +180,27 @@ def test_duplicate_bucket_creation(
...
                 )
+
+    @bugzilla("2179271")
+    @pytest.mark.parametrize(
+        argnames="amount,interface",
+        argvalues=[
+            pytest.param(*[1, "OC"], marks=[tier1]),
+        ],
+    )
+    def test_check_for_error_in_rook_op_logs(self, bucket_factory, amount, interface):
```
parikshithb, line 191: add polarion id

tests/manage/mcg/test_bucket_creation.py
```diff
@@ -178,3 +180,27 @@ def test_duplicate_bucket_creation(
...
+        """
+        Test to check for error 'malformed BucketHost "s3.openshift-storage.svc"
+        in rook operator logs post creation of noobaa obc
+
+        """
+
+        bucket_factory(amount, interface)
```
PrasadDesala, line 198: In the test, we are creating only one OBC. Are we sure that the error message appears in the logs during every OBC creation?
I see below comment in the BZ, 
`+ It has around 16 OBC's from NooBaa and RGW, but the error is only for 2 OBB's.`

tests/manage/mcg/test_bucket_creation.py
PrasadDesala, line 191: The test case name is too generic. we are checking for the bucket notification error, can we change the test name according to it?

tests/manage/mcg/test_bucket_creation.py
Shrivaibavi: The test failed in 4.12 (As expected) , from the logs I could see that even 1 OBC creation leads to that error.
http://pastebin.test.redhat.com/1104678

PrasadDesala: Thanks for testing this test on 4.12 as we discussed.

parikshithb (approved):

PR #7928, opened 2023-06-29
WIP: Make Ceph Health recovery faster

prsurve:


tests/conftest.py
```diff
@@ -1487,11 +1493,17 @@ def finalizer():
                     else:
                         ceph_health_check_base()
                         log.info("Ceph health check passed at teardown")
+                        log.info("Removing Faster recovery profile")
+                        ceph_cmd = "ceph config set osd osd_mclock_profile balanced"
```
petr-balogh, line 1497: You can rename it to `ceph_remove_cmd`, move above line 1470 and remove the duplicated line from 1505.

prsurve: sure will do

petr-balogh (approved):

prsurve:
Running Verification job here https://url.corp.redhat.com/58819bf

prsurve:
tier 4b:- https://url.corp.redhat.com/fdcf9c7
tier 4a:- https://url.corp.redhat.com/449cff4

sseshasa:
@prsurve From the changes, my understanding is that "high_recovery_ops" mclock profile is enabled for the duration of the recovery and once the health is reported 'OK', the profile is reverted back to the default. Is my understanding correct?

I don't see any issue with how the profile is being utilized.

PR #7933, opened 2023-06-30
Add throughput check

ebondare:


ebondare:
Not needed anymore as FAAS has been discontinued

PR #7936, opened 2023-06-30
delete PVC's before teardown

vavuthu:


ocs_ci/deployment/vmware.py
```diff
@@ -1142,6 +1153,49 @@ def destroy_cluster(self, log_level="DEBUG"):
             self.folder_structure = True
             config.ENV_DATA["folder_structure"] = self.folder_structure
 
+        # scale down mon pods
```
petr-balogh, line 1156: Maybe move this to some function outside of teardown.
Also thinking about to gave this in try/except block in case something happen here to not fail and process with destroy of the cluster via terraform as usual.

petr-balogh (approved):

shivamdurgbuns (approved):

PR #7941, opened 2023-07-02
ODF4.14, UI deployment

OdedViner:


ocs_ci/ocs/ui/helpers_ui.py
```diff
@@ -121,7 +121,16 @@ def ui_add_capacity_conditions():
     ):
         logger.info(f"Add capacity via UI is not supported on platform {platform}")
         return False
-    elif ocp_version not in ("4.7", "4.8", "4.9", "4.10", "4.11", "4.12", "4.13"):
+    elif ocp_version not in (
```
DanielOsypenko, line 124: I would recommend to remove the full list of OCP versions, because every new version we would need to remember to add a new version number here, WDYTH?

OdedViner: With this method, I check the ui deployment working as expected and then we run it in CI.  If the ui deployment fail all job fail and not only 1 test

DanielOsypenko: ack

petr-balogh (approved):

DanielOsypenko (approved):

PrasadDesala (approved):

PR #7969, opened 2023-07-06
Remove agent and offering from FaaS provider

jilju:
Add steps to remove agent and DF offering from FaaS provider.


jilju:
obsolete

Issue #7656, opened 2023-05-18
`test_disable_nb_lb` fails to disable MCG's load balancer


sagihirshfeld:
As can be seen in RP:
https://url.corp.redhat.com/a9cb4ac

The patch to the NooBaa CR gets reconciled and the S3 and STS services keep using the load balancer, because the way we're supposed to disable NooBaa's LB has changed in 4.13.

I verified locally that adding `disableLoadBalancerService: true` under spec.multiCloudGateway in the ocs-storagecluster does work as intended

Sravikaz:
Also observing this issue  on IBM Z in ODF 4.3

sagihirshfeld:
Seeing this error again on 4.14 and 4.13:

- 4.13: https://url.corp.redhat.com/e3e5349
- 4.14 https://url.corp.redhat.com/d8dd169

sagihirshfeld:
It's failing again 😞   
- https://url.corp.redhat.com/eaf8073
- https://url.corp.redhat.com/4368a06
- https://url.corp.redhat.com/ebf146e

sagihirshfeld:
Closing since it has been passing consistently on all versions since May 2025

PR #7979, opened 2023-07-07
Test KCS: Disabling Multicloud Object Gateway external service and make sure disableLoadBalancerService isn't reconciled

mashetty330:
Covers:

- KCS: https://access.redhat.com/articles/6970745
- Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2186171

**Need backports?**
Yes need to be backported to release-4.13

### DIFF
--- a/tests/e2e/kcs/test_disable_mcg_external_service.py
+++ b/tests/e2e/kcs/test_disable_mcg_external_service.py
@@ -3,7 +3,12 @@ import logging
 
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs import constants
-from ocs_ci.framework.pytest_customization.marks import tier2, polarion_id, bugzilla
+from ocs_ci.framework.pytest_customization.marks import (
+    tier2,
+    polarion_id,
+    bugzilla,
+    skipif_external_mode,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


tests/e2e/kcs/test_disable_mcg_external_service.py
```diff
@@ -0,0 +1,71 @@
...
+        noobaa_ocp_obj = OCP(
+            kind="noobaa",
+            namespace=constants.OPENSHIFT_STORAGE_NAMESPACE,
+            resource_name="noobaa",
+        )
+
+        # patch noobaa object
+        noobaa_ocp_obj.patch(
+            resource_name="noobaa",
+            params='{"spec": {"disableLoadBalancerService": true }}',
```
sagihirshfeld, line 25: Unless this was changed again in 4.14, the path that actually works is `/spec/multiCloudGateway/disableLoadBalancerService`.

See #7656 and #7657.

mashetty330: @sagihirshfeld please check this comment by liran:
```
Hi,

We are not supporting the disableLoadBalancerService flag via the storagecluster in 4.10.
This is probably the reason its overridden.
They can re-apply the disableLoadBalancerService flag on noobaa after changing the endpoints.

We are going to provide support for the disableLoadBalancerService flag via the storagecluster in 4.13

Best Regards,
Liran.
```
So we have the support for disableLoadBalancerService flag from Noobaa CRD from 4.13 onwards.

sagihirshfeld: I agree that it is supported in 4.13, but this specific path in the YAML did not work in `test_disable_nb_lb` in 4.13, as can be seen here: https://url.corp.redhat.com/a9cb4ac.

Maybe this was changed again since in 4.13 and I'm wrong - we'll have to see for sure in the PR validation.

mashetty330: IIRC it passed on 4.13

sagihirshfeld (approved):

keesturam (approved):

PR #8014, opened 2023-07-18
Remove test_az_failure

AviadP:
As discussed in brown squad team meeting, we decided to remove this test


tests/manage/z_cluster/nodes/test_az_failure.py
```diff
@@ -1,185 +0,0 @@
...
-                    f"Access to EC2 instances {self.instances_in_az} has been restored"
-                )
-
-            if self.security_group_id in aws_obj.get_all_security_groups():
-                logger.info(f"Deleting: {self.security_group_id}")
-                aws_obj.delete_security_group(self.security_group_id)
-
-        request.addfinalizer(finalizer)
-
-    def test_availability_zone_failure(
```
ebenahar, line 62: I would suggest keeping the test file and just replacing `test_availability_zone_failure` with `availability_zone_failure` which will cause this function not to be picked up by pytest collection

PR #8017, opened 2023-07-18
Enhance PVC creation to use kube_job which avoids memory issue

ramkiperiy:
Enhanced PVC creation and Ceph POD respin test to use kube_job obj way of creating PVCs and PODs, with factory objs observed script is using lot of memory also there were lot of kill messages in logs, with this enhancement memory issue should not be hit.


shivamdurgbuns (approved):

tiffanyn108 (approved):

ramkiperiy:
with kubejob TC execution time reduced to half. earlier each test used to run for 20mins, now it's reduced to 10mins and memory utilization reduced from 2G to 500M.

keesturam (approved):

PR #8018, opened 2023-07-18
fix removal of MS tests from skiped list

dahorak:
- this fixes issue in PR#7862 - `fspath` contains path to the file with tests, not the test name
- also the original condition in the code `if testname in str(item....)` would remove also tests with partial match - e.g. `test_foo` testname will match also with `test_foo_bar` test.


dahorak:
[This is the html report from the last verification job](https://url.corp.redhat.com/558be5a) - the job was triggered against `tier1` tests in `tests/e2e/workloads/test_create_scale_pods_and_pvcs_using_kube_job.py`, one test were executed and there is no skipped test listed, which is expected.

Shrivaibavi (approved):
LGTM

PrasadDesala (approved):

PR #8031, opened 2023-07-20
discon. cl.: ensure that metadata.json file exists

dahorak:
- if disconnected deployment using flexy fails before the OCP cluster deployment start (e.g. during images mirroring), the metadata.json file is not created and Destroy job is not properly triggered. This change ensures that at least empty metadata.json file exists and full Destroy job is properly triggerd and all resources are cleaned up.

### DIFF
--- a/ocs_ci/deployment/flexy.py
+++ b/ocs_ci/deployment/flexy.py
@@ -10,6 +10,7 @@ import yaml
 
 import io
 import configparser
+from pathlib import Path
 from semantic_version import Version
 import subprocess
 from subprocess import CalledProcessError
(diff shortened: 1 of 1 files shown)


petr-balogh (approved):

vavuthu (approved):

vavuthu:
@dahorak is this empty metadata.json will be overwrite by actual deployment job ? or will it error saying already file exists?

dahorak:
I'm waiting for the verification job, to be sure, that it didn't break somewhere unexpectedly, but it should overwrite the file (because when using Flexy, the original `metadata.json` file is created on different place and then copied to the cluster dir using `rsync -av ...` command, so it should overwrite it).

dahorak:
The verification job failed during ODF deployment (because of different issue), but OCP deployment passed correctly and also the `metadata.json` file was correctly created before the Containerized Flexy was executed and then overwritten with the correct content once the OCP deployment finished.

PR #8037, opened 2023-07-21
verify maximum openshift version allowed

vavuthu:


petr-balogh (approved):

Akarsha-rai (approved):

b-ranto:
The deployment is now failing if this is not set in the CSV. That should not be the case.

vavuthu:
> The deployment is now failing if this is not set in the CSV. That should not be the case.
> We are hitting this in upgrade testing where we live deploy ODF 4.13 first (no such annotation in the CSV there).

I am under impression that we are using older release branch for upgrades. Raised https://github.com/red-hat-storage/ocs-ci/pull/8043 to verify only for eligible version

Issue #8153, opened 2023-08-01
`MCG::retrieve_noobaa_cli_binary` fails with ArchitectureNotSupported error for arm64 Mac


sagihirshfeld:
In #8001 the following check was added after downloading the MCG-CLI binary from quay:

```
        elif system == "Darwin":  # Mac
            if machine == "arm64":
                raise ArchitectureNotSupported(
                    "Noobaa client binary is not compiled for arm64 architecture!"
                    " Please compile MCG binary yourself and store it "
                    f"{constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH}"
                )
            elif machine == "x86_64":
                path = os.path.join(path, "macosx", noobaa)
```
While this is true for versions 4.13 and under, on 4.14 I can attest that the same binary does work on my M1 Mac

PR #8155, opened 2023-08-01
Remove the Mac architecture check at `mcg.py::get_architecture_path`

sagihirshfeld:
Fixes: #8153


DanielOsypenko (approved):

keesturam (approved):

sagihirshfeld:
Since this can't be verified via any Jenkins job (they're not using MacOS), I'm marking this as verified since it seems to work to any Mac user that tried the fix.

Issue #8163, opened 2023-08-02
s3 sync of the /test_objects dir at the s3cli pod fails with "File not found" errors


sagihirshfeld:
As described in [this BZ](https://bugzilla.redhat.com/show_bug.cgi?id=2227162), many MCG tests are failing while trying to upload files from the /test_objects dir in the s3cli pod:

```
Error during execution of command: oc -n openshift-storage rsh s3cli-0 sh -c "AWS_CA_BUNDLE=/cert/service-ca.crt AWS_ACCESS_KEY_ID=***** AWS_SECRET_ACCESS_KEY=***** AWS_DEFAULT_REGION=us-east-2 aws s3 --endpoint=***** sync /test_objects/ s3://oc-bucket-71316705b4034a0b9296ec69854586".
2023-07-28 12:36:58  Error is warning: Skipping file /test_objects/airbus.jpg. File does not exist.
2023-07-28 12:36:58  warning: Skipping file /test_objects/apple.mp4. File does not exist.
2023-07-28 12:36:58  warning: Skipping file /test_objects/bolder.jpg. File does not exist.
2023-07-28 12:36:58  warning: Skipping file /test_objects/book.txt. File does not exist.
2023-07-28 12:36:58  warning: Skipping file /test_objects/canada.jpg. File does not exist.
2023-07-28 12:36:58  warning: Skipping file /test_objects/danny.webm. File does not exist.
2023-07-28 12:36:58  warning: Skipping file /test_objects/enwik8. File does not exist.
``` 

For some unclear reason this directory does not have the executable permissions, which evidently causes this error (local attempts to s3 sync/cp files from a dir with 777 permission for example have succeeded)

PR #8175, opened 2023-08-02
s3cli-sts: runAsUser 0 with image 2.0

sagihirshfeld:
Fixes: #8163

### DIFF
--- a/ocs_ci/templates/mcg/s3cli-sts.yaml
+++ b/ocs_ci/templates/mcg/s3cli-sts.yaml
@@ -18,14 +18,14 @@ spec:
         app: s3cli
     spec:
       securityContext:
-        runAsUser: 1000
+        runAsUser: 0
       volumes:
         - name: service-ca
           configMap:
             name: awscli-service-ca
       containers:
         - name: s3cli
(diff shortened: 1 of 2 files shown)


ebenahar (approved):

vavuthu (approved):

sagihirshfeld:
The first validation run shows that this change has fixed the original "File not found" issue while also allowing root-privileged operations needed in the TC. 

However, the validation has failed due to a potential BZ we still have to investigate further as a separate issue. 

Running a new PR validation on a different TC that uses s3 sync from the s3cli pod and also makes some of the aforementioned privileged operations.

mashetty330:
> The first validation run shows that this change has fixed the original "File not found" issue while also allowing root-privileged operations needed in the TC.
> 
> However, the validation has failed due to a potential BZ we still have to investigate further as a separate issue.
> 
> Running a new PR validation on a different TC that uses s3 sync from the s3cli pod and also makes some of the aforementioned privileged operations.

Thanks for validating this and about the connection closed issue, i as well have faced similar issue while running validation for https://github.com/red-hat-storage/ocs-ci/pull/7900 and it looks a potential bug to me as well

PR #8212, opened 2023-08-08
skip application of icsp from ocs-registry on disonnected cluster

dahorak:
- on Disconnected cluster ICSP from ocs-registry image is not required/valid (there is created another ICSP pointing directly to the mirror registry)


vavuthu (approved):

petr-balogh (approved):

dahorak:
Marking as verified based on verification here: https://github.com/red-hat-storage/ocs-ci/pull/8214

PR #8251, opened 2023-08-15
make icsp unique to avoid conflict during upgrade

dahorak:
ICSP from ocs-registry image is named `df-repo` and during upgrade between nightly versions, the older one is rewritten by the new one, which might cause issues that some pods went to `ImagePullBackOff` state.

### DIFF
--- a/ocs_ci/utility/deployment.py
+++ b/ocs_ci/utility/deployment.py
@@ -5,6 +5,7 @@ import logging
 import os
 import re
 import tempfile
+import yaml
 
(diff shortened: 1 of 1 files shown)


dahorak:
> [Job FAILED](https://url.corp.redhat.com/ba48d2d) (installation failed, tests not executed).

The verification job failed because of Bug 2232226, but the icsp were properly created with unique name.

vavuthu (approved):

PR #8349, opened 2023-08-28
BZ 2139037 Coverage - `test_rgw_routes`

sagihirshfeld:
Covers  [BZ #2139037](https://bugzilla.redhat.com/show_bug.cgi?id=2139037) by testing the following:

1. RGW's service is exposed by both http and https routes
2. Both endpoints use a TLS termination policy
3. Basic I.O functionality is working properly using both endpoints

### DIFF
--- a/tests/manage/rgw/test_rgw_routes.py
+++ b/tests/manage/rgw/test_rgw_routes.py
@@ -1,0 +1,105 @@
+import logging
+
+import boto3
+
+from ocs_ci.ocs import ocp, constants
+from ocs_ci.ocs.resources.objectbucket import OBC
+from ocs_ci.ocs.exceptions import CommandFailed, UnavailableResourceException
+from ocs_ci.framework.pytest_customization.marks import (
+    bugzilla,
+    skipif_external_mode,
+)
+from ocs_ci.ocs.bucket_utils import (
+    s3_put_object,
+    s3_get_object,
+    s3_delete_object,
+)
+
+log = logging.getLogger(__name__)
+
+
+class TestRGWRoutes:
+    """
+    Test the RGW routes in an ODF cluster
+
+    """
+
+    @skipif_external_mode
+    @bugzilla("2139037")
+    def test_rgw_routes(self, rgw_bucket_factory):
+        """
+        Test the availability of RGW routes in an ODF cluster
+
+        1. Assert that RGW's service is exposed by both http and https routes
+        2. Assert that both endpoints use a TLS termination policy
+        3. Test basic I.O functionality using both endpoints
+
+        """
+        route_obj = ocp.OCP(kind="Route", namespace=self.namespace)
+
+        log.info("Getting RGW's HTTP and HTTPS routes")
+        try:
+            http_route = route_obj.get(resource_name=constants.RGW_ROUTE_INTERNAL_MODE)
+        except CommandFailed as ex:
+            log.error("Failed to get RGW's HTTP route!")
+            raise UnavailableResourceException(ex)
+
+        try:
+            https_route = route_obj.get(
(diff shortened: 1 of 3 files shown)


sagihirshfeld:
The logic of the test case has been verified, but we still need to determine this test's `@tier` and polarion id

tests/manage/rgw/test_rgw_routes.py
```diff
@@ -0,0 +1,113 @@
...
+class TestRGWRoutes:
+    """
+    Test the RGW routes in an ODF cluster
+
+    """
+
+    @skipif_external_mode
+    @bugzilla("2139037")
+    @pytest.mark.polarion_id("OCS-5168")
+    @tier1
```
mashetty330, line 35: check with Harish/Karthick/Prasad if this should be tier1? I'm thinking tier2

tests/manage/rgw/test_rgw_routes.py
```diff
@@ -0,0 +1,113 @@
...
+    def test_rgw_routes(self, rgw_bucket_factory):
```
mashetty330, line 36: please use parametrisation for secure and insecure routes, instead of repeating the code

tests/manage/rgw/test_rgw_routes.py
```diff
@@ -0,0 +1,113 @@
...
+        assert http_route["spec"]["tls"]["insecureEdgeTerminationPolicy"] is not None
+        assert https_route["spec"]["tls"]["insecureEdgeTerminationPolicy"] is not None
+
+        log.info("Testing basic I.O functionality using both endpoints")
+        rgw_bucket_name = rgw_bucket_factory(amount=1, interface="RGW-OC")[0].name
+        rgw_obc = OBC(rgw_bucket_name)
+
+        for route in http_route, https_route:
+            # Apply the current route to the OBC s3_client
+            route_name = route["metadata"]["name"]
```
mashetty330, line 82: can we take this opportunity to improve 
```python
class RGW(object):
``` 
and implement this piece of code using RGW class objects hence enabling the re usability.

tests/manage/rgw/test_rgw_routes.py
sagihirshfeld: Good catch, done

tests/manage/rgw/test_rgw_routes.py
sagihirshfeld: Just to clarify, do you wish to capture the three basic S3 operations that I used here in a method with a name like `test_basic_io` in the RGW class?

mashetty330: No I mean.. if you see the current RGW() class its very basic.. we dont initialize class with s3 client or resources. and you do it here in the test code, can you move that part to RGW class init ? basically something similar to what happens in MCG()

tests/manage/rgw/test_rgw_routes.py
hnallurv: Since the steps covered in this test are for verifying the basic functionality, I recommend it to be part of tier1

tests/manage/rgw/test_rgw_routes.py
sagihirshfeld: I considered adding endpoint properties and "get_s3_resource" methods to the RGW class, but since the merge of this test is time sensitive and implementation depends on comparing the usage with an exernal cluster, I think its better to implement in a separate enhancement PR.

We can track and discuss further enhancements to the RGW class in this issue: https://github.com/red-hat-storage/ocs-ci/issues/8601

mashetty330 (commented):
lgtm please run validation with the latest changes

mashetty330 (approved):

parikshithb (approved):

PrasadDesala (approved):

PR #8356, opened 2023-08-28
[4.11] Support on external mode and increase timeout to allow pods to reach Running state in test_non_ocs_taint_and_tolerations after pod respin

am-agrawa:


pintojoy (approved):
LGTM

prsurve (approved):

PR #8370, opened 2023-08-29
Improve logging mechanism in OCS-CI

petr-balogh:
This PR suppose to reduce console log output in the jenkins jobs. Limit the logs to only INFO level from the /tests folder. Plus adding some usefull sperator info about the calls of fixtures and test names.

The log as we were used to have should be located under logs folder in file called: test-info-level.log

This contains INFO level logs from all the modules.

Related ticket:
https://issues.redhat.com/browse/OCSQE-882


petr-balogh:
> > run_cmd() or equivalent log level of `oc command` to be replaced from info to debug
> 
> If we would see only a fixture name in the log as it is in the acceptance run https://url.corp.redhat.com/2e52fff we will miss 90% of the test flow because not all the tests are based on fixtures.
> 
> All the existing tests where build based on existing logging, due to the logic that **every oc cmd is visible**. Redo this logging for existing tests will take serious effort for all the squads.
> 
> Convenient log output from the Jenkins console where we see the test and the prior tests together, with cmd's but without DEBUG lvl will now be useless.
> 
> It will make analysis more complicated and seriously more time consuming. @ebenahar @keesturam

@DanielOsypenko hello, don't be afraid of this change. I am keeping still all the logs as you were used to have in console output in the file: `test-info-level.log` where you have INFO level logs from all modules.

You can find it here:
https://url.corp.redhat.com/830dc6e

Current problem of jenkins output if you do consoleFull output, it has too many information which makes your browser crashing and is not usable. For high level overview which test is running and what is running should be visible only from test logs - if test do not provide enough logs - it's a problem of the test.

I have also added some [MD](https://github.com/red-hat-storage/ocs-ci/blob/f927f1c4f729067b1a0b69544467fd7a4da0ea5e/docs/logs.md) file where I am trying to describe it as well.

petr-balogh:
New job with some latest updates:
https://url.corp.redhat.com/9a968e2

dahorak (approved):
Small suggestion about usage of built-in python function, otherwise LGTM.

DanielOsypenko:
> > > run_cmd() or equivalent log level of `oc command` to be replaced from info to debug
> > 
> > 
> > If we would see only a fixture name in the log as it is in the acceptance run https://url.corp.redhat.com/2e52fff we will miss 90% of the test flow because not all the tests are based on fixtures.
> > All the existing tests where build based on existing logging, due to the logic that **every oc cmd is visible**. Redo this logging for existing tests will take serious effort for all the squads.
> > Convenient log output from the Jenkins console where we see the test and the prior tests together, with cmd's but without DEBUG lvl will now be useless.
> > It will make analysis more complicated and seriously more time consuming. @ebenahar @keesturam
> 
> @DanielOsypenko hello, don't be afraid of this change. I am keeping still all the logs as you were used to have in console output in the file: `test-info-level.log` where you have INFO level logs from all modules.
> 
> You can find it here: https://url.corp.redhat.com/830dc6e
> 
> Current problem of jenkins output if you do consoleFull output, it has too many information which makes your browser crashing and is not usable. For high level overview which test is running and what is running should be visible only from test logs - if test do not provide enough logs - it's a problem of the test.
> 
> I have also added some [MD](https://github.com/red-hat-storage/ocs-ci/blob/f927f1c4f729067b1a0b69544467fd7a4da0ea5e/docs/logs.md) file where I am trying to describe it as well.

Hello Petr,
After our discussion, it appears that we wouldn't lose much data by having a separate file with INFO-level logs that mimic the Jenkins console.
One request is to consider how to **include a stack trace in the log file** similar to what we currently have in the Jenkins console. This would significantly assist in debugging issues and eliminate the need to switch between the log file and the console. In this case, it would be preferable to add the stack trace after the teardown of each test, rather than as a list of stack traces as it is in the Jenkins console currently.

petr-balogh:
Another output https://url.corp.redhat.com/64b66f8

ocs_ci/utility/logging.py
```diff
@@ -0,0 +1,64 @@
...
+
+    Args:
+        symbol_ (str): Symbol to use as separator, e.g. "-"
+        val (str): Value which should be in the center of the output text
+
+    Returns:
+        str: string like ------------- VAL ----------------
+
+    """
+    terminal_width = shutil.get_terminal_size(fallback=(120, 40))[0]
```
dahorak, line 60: IIUC in jenkins job it uses the fallback, right? I would propose to set it to lower number than 120 (what about 80?), because with the timestamp and other information in the beginning of the line it is quite long and on normal monitor is split to two lines, which doesn't look clear.
obrazek

petr-balogh: This is because the logger adding additional characters like:
`17:46:53 - MainThread - console.logger - INFO  -`
And then if we are adding more text in terminal size it's overflowing :(

I came up with solution that by default I am adding new line character before separated text.
So now it should add it to new line to match the size of terminal correctly on new line.

petr-balogh:
https://url.corp.redhat.com/556dd00

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -413,6 +427,23 @@ def pytest_configure(config):
...
+        ocsci_config.switch_ctx(i)
+        if logs_url:
+            ocsci_config.RUN["info_logs_url"] = urljoin(
+                logs_url, os.path.basename(pytest_logs_dir)
+            )
+        ocsci_config.REPORTING["log_path_message"] = (
+            f"Logs from run-ci execution (RUN ID: {ocsci_config.RUN['run_id']})"
+            f" will be stored in: {ocsci_config.RUN.get('info_logs_url') or pytest_logs_dir}\n"
+        )
+    print(ocsci_config.REPORTING["log_path_message"])
```
dahorak, line 446: This is intentionally printed with the `print` command, not logged through `logger`, right?

ocs_ci/framework/pytest_customization/reports.py
```diff
@@ -174,3 +234,41 @@ def pytest_report_teststatus(report, config):
...
+    logger_config.add_loggers([""], stdout_level="info")
+    logger_config.set_log_option_default("")
+    logger_config.split_by_outcome()
+    logger_config.set_formatter_class(OCSLogFormatter)
+
+
+@pytest.hookimpl(tryfirst=True)
+def pytest_runtest_setup(item):
+    test_name = f"TEST NAME: {item.name}"
+    console_logger.info(f"\n{separator(symbol_='-', val=test_name)}")
```
dahorak, line 249: Is the `\n` new line expected here (when there will be second new line in the output from `separator(...)` function)?

ocs_ci/utility/logging.py
```diff
@@ -0,0 +1,67 @@
...
+        str: string like ------------- VAL ----------------
+
+    """
+    new_line_char = "\n"
+    if not new_line:
+        new_line_char = ""
+    terminal_width = shutil.get_terminal_size(fallback=(80, 40))[0]
+    if not val:
+        return f"{new_line_char}{symbol_ * terminal_width}"
+    return f"{new_line_char}{val.center(terminal_width, symbol_)}"
```
dahorak, line 67: I think that you can make it little bit simpler here with just the final `return` statement (without the `if not val` condition) - if you will set the default value for `val` to empty string instead of `None`.

dahorak (dismissed):
Overall LGTM!

dahorak:
> PR validation
> 
> Cluster Name: Cluster Configuration: PR Test Suite: acceptance PR Test Path: tests/ Additional Test Params: OCP VERSION: 4.14 OCS VERSION: 4.14 tested against branch: master
> 
> [Job PASSED](https://url.corp.redhat.com/7fe638a).

The original job were deleted, but all logs are available here: https://url.corp.redhat.com/251a2b4

ocs_ci/framework/pytest_customization/reports.py
petr-balogh: yes, in the output I wanted to have one extra empty line as it looked better in the output.

ocs_ci/framework/pytest_customization/ocscilib.py
petr-balogh: Yes, this is intentionally to print this out like the pytest is printing out the summary at the end of execution, so this needs to be just printed out without any logging information

ocs_ci/utility/utils.py
```diff
@@ -2145,8 +2145,34 @@ def ocsci_log_path():
...
+            additional_info += "-test"
+
+    jenkins_job_build = os.environ.get("BUILD_NUMBER")
+    if jenkins_job_build:
+        additional_info += f"-{jenkins_job_build}"
     return os.path.expanduser(
-        os.path.join(config.RUN["log_dir"], f"ocs-ci-logs-{config.RUN['run_id']}")
+        os.path.join(
+            config.RUN["log_dir"],
+            f"ocs-ci-logs{additional_info}-{config.RUN['run_id']}",
```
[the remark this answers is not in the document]
petr-balogh: Done

petr-balogh:
New verification job scheduled here:
https://url.corp.redhat.com/04adb3d

petr-balogh:
New one validation:
 https://url.corp.redhat.com/861d678

petr-balogh:
New verification:
 https://url.corp.redhat.com/2f2be2b

OdedViner (approved):

docs/logging.md
```diff
@@ -0,0 +1,48 @@
...
+User can change it by providing own location via config file passed via `--ocsci-conf`:
+
+```yaml
+---
+RUN:
+    log_dir: /my/custom/log/location
+```
+
+In the log_dir defined above, the run-ci will create new folder per execution
+named in format of: `ocs-ci-logs-$RUN_TYPE-$JENKINS_BUILD_ID-$RUN_ID`.
```
dahorak, line 15: I think that the format will be little bit shifted - first is the `RUN_ID` and then the additional info:
```
f"ocs-ci-logs-{config.RUN['run_id']}{additional_info}",
```
Also the `JENKINS_BUILD_ID` is valid only for Jenkins execution, so it might be worth to mention it that for non-jenkins executions, this will be empty.

dahorak (approved):
Just small note about documentation (usage of `JENKINS_BUILD_ID`), otherwise LGTM.

PR #8443, opened 2023-09-10
In case of high skip rate due to Ceph health not OK - Report test case and squad causing that

ebenahar:
This test runs second to last and examines the skipped test cases of the execution.
In case of a high rate of skipped tests due to Ceph health not OK, which indicates something went wrong with the cluster during the execution, it will fail and report the potential test case that caused this problematic state.

Below is an example for how it works:

```
2023-09-12 01:03:10 E Failed: This run had 71.875% of the tests skipped due to Ceph health not OK. The test that is likely 
to cause this is test_bucket_creation[3-CLI-DEFAULT-BACKINGSTORE] which is under red_squad's responsibility
```

In this example, since this logic identified the culprit to be a test case belonging to Red Squad, `test_report_skip_triggering_test` was automatically marked with the `red_squad` marker and it will be categorized under Red Squad so it will be examined by that squad

### DIFF
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -662,6 +662,7 @@ def pytest_collection_modifyitems(session, config, items):
 
 @pytest.hookimpl(tryfirst=True, hookwrapper=True)
 def pytest_runtest_makereport(item, call):
+    ocsci_config.RUN["number_of_tests"] = len(item.session.items)
     outcome = yield
     rep = outcome.get_result()
     # we only look at actual failing test calls, not setup/teardown
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -662,7 +662,6 @@ def pytest_collection_modifyitems(session, config, items):
 
 @pytest.hookimpl(tryfirst=True, hookwrapper=True)
 def pytest_runtest_makereport(item, call):
-    ocsci_config.RUN["number_of_tests"] = len(item.session.items)
     outcome = yield
     rep = outcome.get_result()
     # we only look at actual failing test calls, not setup/teardown
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -292,6 +292,18 @@ def pytest_collection_modifyitems(session, items):
                 items.remove(item)
 
 
+def pytest_collection_finish(session):
+    """
+    A pytest hook to get all collected tests post their collection modifications done in the varius
+    pytest_collection_modifyitems hook functions
+
+    Args:
+        session: pytest session
+
+    """
+    config.RUN["number_of_tests"] = len(session.items)
+
+
 @pytest.fixture()
 def supported_configuration():
     """
@@ -1486,6 +1498,10 @@ def health_checker(request, tier_marks_name, upgrade_marks_name):
 
     node = request.node
 
+    # ignore ceph health check for the TestFailurePropagator test cases
+    if "FailurePropagator" in node.cls:
+        return
+
     def finalizer():
         if not skipped:
             try:
@@ -1502,6 +1518,7 @@ def health_checker(request, tier_marks_name, upgrade_marks_name):
                     for mark in node.iter_markers():
                         if "ceph_health_retry" == mark.name:
                             ceph_health_retry = True
+                            break
                     if ceph_health_retry:
                         ceph_health_check(
                             namespace=config.ENV_DATA["cluster_namespace"]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1514,6 +1514,15 @@ def health_checker(request, tier_marks_name, upgrade_marks_name):
                         ceph_health_check_base()
                         log.info("Ceph health check passed at teardown")
             except CephHealthException:
+                if config.RUN["skip_reason_test_found"] is None:
+                    squad_name = None
+                    for marker in node.iter_markers():
+                        if "_squad" in marker.name:
+                            squad_name = marker.name
+                    config.RUN["skip_reason_test_found"] = {
+                        "test_name": node.name,
+                        "squad": squad_name,
+                    }
                 log.info("Ceph health check failed at teardown")
                 # Retrying to increase the chance the cluster health will be OK
                 # for next test
@@ -1532,6 +1541,7 @@ def health_checker(request, tier_marks_name, upgrade_marks_name):
                     log.info("Ceph health check passed at setup")
                     return
             except CephHealthException:
+                config.RUN["skipped_tests_ceph_health"] += 1
                 skipped = True
                 # skip because ceph is not in good health
                 pytest.skip("Ceph health check failed at setup")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1499,7 +1499,7 @@ def health_checker(request, tier_marks_name, upgrade_marks_name):
     node = request.node
 
     # ignore ceph health check for the TestFailurePropagator test cases
-    if "FailurePropagator" in node.cls:
+    if "FailurePropagator" in str(node.cls):
         return
 
     def finalizer():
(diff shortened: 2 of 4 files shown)


tests/conftest.py
```diff
@@ -1514,6 +1514,15 @@ def finalizer():
                         ceph_health_check_base()
                         log.info("Ceph health check passed at teardown")
             except CephHealthException:
+                if config.RUN["skip_reason_test_found"] is None:
+                    squad_name = None
+                    for marker in node.iter_markers():
+                        if "_squad" in marker.name:
+                            squad_name = marker.name
```
petr-balogh, line 1521: I think you can break from the loop here once you get squad marker to avoid looping over the other markers.

tests/test_failure_propagator.py
```diff
@@ -1,15 +1,59 @@
...
 
 
 @tier1
-@pytest.mark.last
+@acceptance
+@tier2
+@tier3
+@tier4a
+@tier4b
+@tier4c
```
petr-balogh, line 21: what about pre and post upgrade markers?

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -662,6 +662,7 @@ def pytest_collection_modifyitems(session, config, items):
 
 @pytest.hookimpl(tryfirst=True, hookwrapper=True)
 def pytest_runtest_makereport(item, call):
+    ocsci_config.RUN["number_of_tests"] = len(item.session.items)
```
petr-balogh, line 665: pytest_runtest_makereport is called after every item is executed IIUC.
```
Called to create a TestReport for each of the setup, call and teardown runtest phases of a test item.
```
[Documentation link](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.TestReport)

So for every test we run again and again right? I thought this should be one time operation done in different hook called:
pytest_collection_modifyitems

Where at the end you can do len on the items I think only once at the end of this hook.

But I might see a problem that we have this hook implementation in many different places:
```
tests/conftest.py:def pytest_collection_modifyitems(session, items):
tests/disaster-recovery/mdr/conftest.py:def pytest_collection_modifyitems(items):
tests/disaster-recovery/regional-dr/conftest.py:def pytest_collection_modifyitems(items):
tests/disaster-recovery/sc_arbiter/conftest.py:def pytest_collection_modifyitems(items):
tests/e2e/conftest.py:def pytest_collection_modifyitems(items):
tests/e2e/scale/conftest.py:def pytest_collection_modifyitems(items):
tests/manage/mcg/conftest.py:def pytest_collection_modifyitems(items):
tests/manage/pv_services/conftest.py:def pytest_collection_modifyitems(items):
tests/manage/rgw/conftest.py:def pytest_collection_modifyitems(items):
tests/manage/storageclass/conftest.py:def pytest_collection_modifyitems(items):
```

So maybe this is the reason why you haven't use it.

What about this hook:
https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.hookspec.pytest_collection_finish

? This might be after all collection is done.

tests/test_failure_propagator.py
```diff
@@ -1,15 +1,59 @@
...
 class TestFailurePropagator:
     """
     Test class for failure propagator test case. The test intention is to run last and propagate
     teardown failures caught during the test execution, so regular test cases won't false negatively fail
     """
 
+    @pytest.mark.second_to_last
+    def test_report_skip_triggering_test(self):
```
petr-balogh, line 29: How we avoid this test to be skipped because of ceph health issue as well?

tests/test_failure_propagator.py
```diff
@@ -1,15 +1,59 @@
...
+                    + f"The test that is likely to cause this is {config.RUN['skip_reason_test_found']['test_name']} "
+                )
+                if config.RUN["skip_reason_test_found"]["squad"]:
+                    message = (
+                        message
+                        + f"which is under {config.RUN['skip_reason_test_found']['squad']}'s responsibility"
+                    )
+
+            else:
+                message = message + "Couldn't identify the test case that caused this"
```
petr-balogh, line 53: I think better to add empty space before `Couldn't` world

```suggestion
                message = message + " Couldn't identify the test case that caused this"
```

So you don't need to care about if last `message` was with empty space at the end.

tests/test_failure_propagator.py
```diff
@@ -1,15 +1,59 @@
...
+            pytest.fail(message)
+        pass
```
petr-balogh, line 55: pass does nothing here, I think you can omit.

ocs_ci/framework/pytest_customization/ocscilib.py
ebenahar: Will use pytest_collection_finish

tests/conftest.py
ebenahar: +1

tests/test_failure_propagator.py
ebenahar: added

tests/test_failure_propagator.py
ebenahar: good point, I will add an exception to the health check setup fixture

tests/test_failure_propagator.py
ebenahar: done

tests/test_failure_propagator.py
ebenahar: ack

ebenahar:
> 
> Cluster Name: ebenahar2-s11 Cluster Configuration: PR Test Suite: acceptance PR Test Path: tests/ Additional Test Params: OCP VERSION: 4.14 OCS VERSION: 4.14 tested against branch: master
> 
> [Job UNSTABLE](https://url.corp.redhat.com/d257026) (some or all tests failed).

the newly added test case in this PR, called test_report_skip_triggering_test, is reported as failed under red squad. 
This is because during the early phase of the acceptance suite execution, I scaled down one of the OSDs deployment, so ceph health will be in WARN state. 
This happened during the execution of one of the red squad's test cases, and therefore, test_report_skip_triggering_test was dynamically marked with the red squad marker.

When examining the failure of test_report_skip_triggering_test, we can see the following message, which tells which red squad test case is the potential culprit which caused the rest of the test cases to get skipped on Ceph health not OK:

2023-09-12 01:03:10 E Failed: This run had 71.875% of the tests skipped due to Ceph health not OK. The test that is likely to cause this is test_bucket_creation[3-CLI-DEFAULT-BACKINGSTORE] which is under red_squad's responsibility

petr-balogh (dismissed):

ebenahar:
> PR validation
> 
> Cluster Name: Cluster Configuration: PR Test Suite: acceptance PR Test Path: tests/ Additional Test Params: OCP VERSION: 4.14 OCS VERSION: 4.14 tested against branch: master
> 
> [Job PASSED](https://url.corp.redhat.com/0cfb109).

also validated that the acceptance suite passes without having Ceph health issues injected during the execution

tests/test_failure_propagator.py
```diff
@@ -1,15 +1,83 @@
...
+        number_of_eligible_tests = config.RUN.get("number_of_tests") - 2
+
+        skipped_on_ceph_health_percent = (
+            config.RUN.get("skipped_tests_ceph_health") / number_of_eligible_tests
+        )
+        message = (
+            f"This run had {skipped_on_ceph_health_percent * 100}% of the "
+            f"tests skipped due to Ceph health not OK."
+        )
+        if 0.5 < skipped_on_ceph_health_percent < 1:
```
dahorak, line 62: What is the reason for the second part of the condition `skipped_on_ceph_health_percent < 1`?
I meant, that it could theoretically be `1` when all the tests will be skipped and we probably don't want to miss that situation.

Also I'm thinking if 50% skipped tests is ok?

dahorak (approved):
Just a question/doubt/note about the condition when to report the failure, otherwise LGTM.

tests/test_failure_propagator.py
ebenahar: you are right. Let me remove the `< 1` part and reduce the threshold to 25%

petr-balogh (approved):

jilju (approved):

ebenahar:
verified once again - https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/29349.

Issue #8437, opened 2023-09-08
UI deployment in 4.14 failing with  Failed to find the element (xpath,//button[text()='Create StorageSystem'])"


petr-balogh:
```
failed on setup with "selenium.common.exceptions.TimeoutException: Message: Failed to find the element (xpath,//button[text()='Create StorageSystem'])"
```
Failed job:
 https://url.corp.redhat.com/ca75845

PR #8449, opened 2023-09-11
UI Deployment, add a new step 'Data Protection'

OdedViner:
#8437
image

### DIFF
--- a/ocs_ci/ocs/ui/deployment_ui.py
+++ b/ocs_ci/ocs/ui/deployment_ui.py
@@ -349,7 +349,7 @@ class DeploymentUI(PageNavigator):
         Configure Data Protection
 
         """
-        if self.ocp_version_semantic >= version.VERSION_4_14:
+        if self.ocs_version_semantic >= version.VERSION_4_14:
             self.do_click(self.dep_loc["next"], enable_screenshot=True)
 
     def enable_taint_nodes(self):
--- a/ocs_ci/ocs/ui/deployment_ui.py
+++ b/ocs_ci/ocs/ui/deployment_ui.py
@@ -258,6 +258,8 @@ class DeploymentUI(PageNavigator):
 
         self.configure_encryption()
 
+        self.configure_data_protection()
+
         self.create_storage_cluster()
 
     def install_internal_cluster(self):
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -258,6 +258,8 @@ def install_lso_cluster(self):
 
         self.configure_encryption()
 
+        self.configure_data_protection()
```
petr-balogh, line 261: What about 4.13 case? I mean in the case we install 4.13 cluster out of master branch because of upgrade scenario.  Do we need to add condition to make sure that OCS version is >= 4.14 ?

am-agrawa: +1

ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -340,6 +344,13 @@ def configure_encryption(self):
             )
         self.do_click(self.dep_loc["next"], enable_screenshot=True)
 
+    def configure_data_protection(self):
+        """
+        Configure Data Protection
+
+        """
+        self.do_click(self.dep_loc["next"], enable_screenshot=True)
```
am-agrawa, line 352: Can we add a verifiication step to ensure it's actually enabled?

ocs_ci/ocs/ui/deployment_ui.py
OdedViner: We don't run UI deployment on upgrade cases because of this condition... https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/ui/helpers_ui.py#L50

ocs_ci/ocs/ui/deployment_ui.py
OdedViner: @prsurve @am-agrawa Do you want to deploy RDR seup via UI? the RDR checkbox is unchecked by default...

am-agrawa: I am not sure. RDR deployment is a mix of UI and CLI. All I am saying is that if someone enables it via UI, we should have some checks to verify DR is enabled. Having it disabled by default is fine for normal deployments.

ocs_ci/ocs/ui/deployment_ui.py
petr-balogh: I see this condition:
`ocs_version != ocp_version `

But our upgrade jobs do upgrade of both if it's Y-stream upgrade, we upgrade OCP,ODF 4.13 to OCP,ODF 4.14.

So in this case ODF and OCP is deployed from master - 4.14 branch the 4.13 version and then will do upgrade .

So the reply here is not considering this scenario and it's just one check to add here to avoid this code to run in 4.13 and below.

petr-balogh (approved):

vavuthu (approved):

PR #8466, opened 2023-09-12
Add the `skip_tls_verify` run config param

sagihirshfeld:
Adds the option to set the `skip_tls_verify` class member of the OCP object to true at its `__init__` method by having the following config param under RUN in an `ocsci-config ` file:

```
RUN:
  skip_tls_verify: true
```
External teams have expressed interest in this sort of config to overcome certification issues in their environments.

Required backports:
- [ ] 4.13
- [ ] 4.12
- [ ] 4.11


sagihirshfeld:
After further inspection, many OC commands are getting executed not via `OCP.exec_oc_cmd` so this would not solve the issue at hand, and modifying these commands is not realistic due to their amount.

PR #8481, opened 2023-09-13
A few improvements for skips Ceph health reporting

ebenahar:
* Added skips on Ceph health percentage to the results breakdown
* Added 2 ReportPortal attributes - `ceph_health_skips` and `ceph_health_skips_over_threshold`
* For acceptance suite, the value of config.RUN["skipped_on_ceph_health_threshold"] would be set to 0 so any skip on Ceph health during an acceptance suite execution would cause test_report_skip_triggering_test to fail

### DIFF
--- a/conf/README.md
+++ b/conf/README.md
@@ -57,6 +57,15 @@ run it belongs here.
 * `log_utilization` - Enable logging of cluster utilization metrics every 10 seconds. Set via --log-cluster-utilization
 * `use_ocs_worker_for_scale` - Use OCS workers for scale testing (Default: false)
 * `load_status` - Current status of IO load
+* `skip_reason_test_found` - In the case the cluster left unhealthy, this param is used to determine the
+  test case that is likely to cause that
+* `skipped_tests_ceph_health` - The number of tests that got skipped due to Ceph being unhealthy
+* `number_of_tests` - The number of tests being collected for the test execution
+* `skipped_on_ceph_health_ratio` - The ratio of tests skipped due to Ceph unhealthy against the
+  number of tests being collected for the test execution
+* `skipped_on_ceph_health_threshold` - The allowed threshold for the ratio of tests skipped due to Ceph unhealthy against the
+  number of tests being collected for the test execution. The default value is set to 0.1 (10%).
+  For acceptance suite, the value would be set to 0
 
 #### DEPLOYMENT
 
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -36,6 +36,8 @@ RUN:
   skip_reason_test_found: {}
   skipped_tests_ceph_health: 0
   number_of_tests: None
+  skipped_on_ceph_health_ratio: 0
+  skipped_on_ceph_health_threshold: 0.1
 
 
 # In this section we are storing all deployment related configuration but not
--- a/ocs_ci/utility/reporting.py
+++ b/ocs_ci/utility/reporting.py
@@ -73,6 +73,13 @@ def get_rp_launch_attributes():
         rp_attrs["fips"] = True
     if config.ENV_DATA.get("encryption_at_rest"):
         rp_attrs["encryption_at_rest"] = True
+    if config.RUN["skipped_on_ceph_health_ratio"] > 0:
+        rp_attrs["ceph_health_skips"] = True
+    if (
+        config.RUN["skipped_on_ceph_health_ratio"]
+        > config.RUN["skipped_on_ceph_health_threshold"]
+    ):
+        rp_attrs["ceph_health_skips_over_threshold"] = True
 
     return rp_attrs
 
--- a/tests/test_failure_propagator.py
+++ b/tests/test_failure_propagator.py
@@ -54,27 +54,46 @@ class TestFailurePropagator:
         """
         number_of_eligible_tests = config.RUN.get("number_of_tests") - 2
 
-        skipped_on_ceph_health_percent = (
-            config.RUN.get("skipped_tests_ceph_health") / number_of_eligible_tests
-        )
-        message = (
-            f"This run had {skipped_on_ceph_health_percent * 100}% of the "
-            f"tests skipped due to Ceph health not OK."
-        )
-        if skipped_on_ceph_health_percent > 0.25:
-            if config.RUN.get("skip_reason_test_found"):
-                test_name = config.RUN.get("skip_reason_test_found").get("test_name")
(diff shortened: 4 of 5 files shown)


ocs_ci/utility/reporting.py
```diff
@@ -73,6 +73,10 @@ def get_rp_launch_attributes():
         rp_attrs["fips"] = True
     if config.ENV_DATA.get("encryption_at_rest"):
         rp_attrs["encryption_at_rest"] = True
+    if config.RUN["skipped_on_ceph_health_ratio"] > 0:
+        rp_attrs["ceph_health_skips"] = True
+    if config.RUN["skipped_on_ceph_health_ratio"] > 0.25:
```
petr-balogh, line 78: Now we have this value on two places, I think that this value should be configurable from config. In default config we can have this set to 0.25 or any other number we will agree on and even add some CLI parameter which will increase this threshold. I saw the concerns in the email. This might even allow us to change default to 0% threshold, and by CLI or config change it to some other value if needed. WDYT?

dahorak: I agree with Petr, that it will be useful to make the threshold configurable and also tend to propose to set it for 0% by default.

ebenahar: added `config.RUN["skipped_on_ceph_health_threshold"]` which is set to 0.25 (25%) by default.
In addition to that, I decided that for `acceptance` suite, the value of `config.RUN["skipped_on_ceph_health_threshold"]` would be set to 0 so any skip on Ceph health during an `acceptance` suite execution would cause `test_report_skip_triggering_test` to fail

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -36,6 +36,8 @@ RUN:
   skip_reason_test_found: {}
   skipped_tests_ceph_health: 0
   number_of_tests: None
+  skipped_on_ceph_health_ratio: 0
+  skipped_on_ceph_health_threshold: 0.25
```
dahorak, line 40: Please add the new parameters to the [conf/README.md](https://github.com/red-hat-storage/ocs-ci/blob/master/conf/README.md) (I miss that in the previous PR).

tests/test_failure_propagator.py
```diff
@@ -52,27 +52,46 @@ def test_report_skip_triggering_test(self, request):
...
-                    message + f" The test that is likely to cause this is {test_name}"
-                )
-                squad = config.RUN.get("skip_reason_test_found").get("squad")
-                if squad:
-                    message = message + f" which is under {squad}'s responsibility"
-                    request.node.add_marker(squad)
+        # for acceptance suite, the value of config.RUN["skipped_on_ceph_health_threshold"] would be set to 0
+        # so any skip on Ceph health during an acceptance suite execution would cause
+        # test_report_skip_triggering_test to fail
+        if config.RUN.get("cli_params").get("-m") == "acceptance":
```
dahorak, line 58: Here it will ignore the case, if combination of `acceptance` and some other testsuite will be triggered (e.g. `acceptance or tier1`) - but I think it is not much likely production job configuration so it is ok to leave it like this.

dahorak (commented):
Please update the documentation with the new parameters, otherwise LGTM!

ocs_ci/framework/conf/default_config.yaml
petr-balogh: @ebenahar have you seen my input in the mail thread?  I think that 25% is too high. What about to lover to 10 or even 5%

tests/test_failure_propagator.py
petr-balogh:
```suggestion
        if "acceptance" in config.RUN.get("cli_params").get("-m", "") :
```
Should resolve it.

conf/README.md
```diff
@@ -57,6 +57,14 @@ run it belongs here.
 * `log_utilization` - Enable logging of cluster utilization metrics every 10 seconds. Set via --log-cluster-utilization
 * `use_ocs_worker_for_scale` - Use OCS workers for scale testing (Default: false)
 * `load_status` - Current status of IO load
+* `skip_reason_test_found` - In the case of cluster left unhealthy, used to determine the
+test case that is likely to cause that
```
petr-balogh, line 61: if the text belongs to point above you should indent it by two spaces here.

conf/README.md
```diff
@@ -57,6 +57,14 @@ run it belongs here.
...
+* `skipped_tests_ceph_health` - The number of tests that got skipped due to Ceph being unhealthy
+* `number_of_tests` - The number of tests being collected for the test execution
+* `skipped_on_ceph_health_ratio` - The ratio of tests skipped due to Ceph unhealthy against the
+number of tests being collected for the test execution
+* `skipped_on_ceph_health_threshold` - The allowed threshold for the ratio of tests skipped due to Ceph unhealthy against the
+number of tests being collected for the test execution
```
petr-balogh, line 67: Indent, also I would mention here the default value and maybe that in case of acceptance suite, this will be always set to 0.

conf/README.md
```diff
@@ -57,6 +57,14 @@ run it belongs here.
 * `log_utilization` - Enable logging of cluster utilization metrics every 10 seconds. Set via --log-cluster-utilization
 * `use_ocs_worker_for_scale` - Use OCS workers for scale testing (Default: false)
 * `load_status` - Current status of IO load
+* `skip_reason_test_found` - In the case of cluster left unhealthy, used to determine the
+test case that is likely to cause that
+* `skipped_tests_ceph_health` - The number of tests that got skipped due to Ceph being unhealthy
+* `number_of_tests` - The number of tests being collected for the test execution
+* `skipped_on_ceph_health_ratio` - The ratio of tests skipped due to Ceph unhealthy against the
+number of tests being collected for the test execution
```
petr-balogh, line 65: Indent

petr-balogh (dismissed):

dahorak (dismissed):

dahorak (approved):

petr-balogh (dismissed):

petr-balogh:
/cherry-pick release-4.13

This is just to test how it will look like if there will be a conflict.

ebenahar:
Verified - https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-trigger-test-pr/5767/console
Screenshot 2023-09-27 at 8 56 51

petr-balogh (approved):

vavuthu (approved):

PR #8483, opened 2023-09-13
Test Squad Ownership Decorators

clacroix12:
This is a large but fairly straight forward change set.

The first goal was to add validation in the pytest  collection process to verify that all collected tests were decorated with a squad owner marker. This was necessary to prepare for the new test directory structure proposal so we can move test locations without losing squad ownership of tests. The second goal was to add markers to all tests that were missing them.

Jira: https://url.corp.redhat.com/3be0f4e

### DIFF
--- a/tests/ecosystem/upgrade/test_upgrade.py
+++ b/tests/ecosystem/upgrade/test_upgrade.py
@@ -2,6 +2,7 @@ import logging
 
 import pytest
 
+from ocs_ci.framework.pytest_customization.marks import purple_squad
 from ocs_ci.framework.testlib import (
     ocs_upgrade,
     polarion_id,
@@ -25,6 +26,7 @@ def teardown(request, nodes):
     request.addfinalizer(finalizer)
 
 
+@purple_squad
 @pytest.mark.polarion_id("OCS-1579")
 def test_worker_node_abrupt_shutdown(teardown):
     """
@@ -36,6 +38,7 @@ def test_worker_node_abrupt_shutdown(teardown):
     run_ocs_upgrade(operation=worker_node_shutdown, abrupt=True)
 
 
+@purple_squad
 @pytest.mark.polarion_id("OCS-1575")
 def test_worker_node_permanent_shutdown(teardown):
     """
@@ -46,6 +49,7 @@ def test_worker_node_permanent_shutdown(teardown):
     run_ocs_upgrade(operation=worker_node_shutdown, abrupt=False)
 
 
+@purple_squad
 @pytest.mark.polarion_id("OCS-1558")
 def test_osd_reboot(teardown):
     """
@@ -57,6 +61,7 @@ def test_osd_reboot(teardown):
     run_ocs_upgrade(operation=osd_node_reboot)
 
 
+@purple_squad
 @ocs_upgrade
 @polarion_id(get_polarion_id(upgrade=True))
 def test_upgrade():
--- a/tests/test_failure_propagator.py
+++ b/tests/test_failure_propagator.py
@@ -15,12 +15,14 @@ from ocs_ci.framework.pytest_customization.marks import (
     pre_ocp_upgrade,
     post_ocp_upgrade,
     post_ocs_upgrade,
+    purple_squad,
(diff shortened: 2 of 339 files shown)


petr-balogh (dismissed):

tests/test_failure_propagator.py
```diff
@@ -15,12 +15,14 @@
...
     post_ocp_upgrade,
     post_ocs_upgrade,
+    purple_squad,
     workloads,
     performance,
     scale,
 )
 
 
+@purple_squad
```
dahorak, line 25: Shouldn't be the marker for the `test_report_skip_triggering_test` added automatically during the test execution based on which test probably caused the ceph healt issue? @ebenahar

tests/ecosystem/upgrade/test_upgrade.py
```diff
@@ -25,6 +26,7 @@ def finalizer():
     request.addfinalizer(finalizer)
 
 
+@purple_squad
```
dahorak, line 29: @AviadP @ebenahar Which squad should be the owner of those disruptive tests executed during upgrade?

dahorak (approved):
Overall LGTM, I have just doubt about the disruptive tests in tests_upgrade.py and  the `test_report_skip_triggering_test`.

OdedViner (dismissed):

tests/test_failure_propagator.py
clacroix12: Yeah marking this as `purple_squad` was really just to make the `--collectonly` test happy but after talking with Petr i'm going to introduce another test marker that we can use to ignore tests more directly. I can also update the test logic to mark this with the squad based on the failure.

DanielOsypenko:
> `tests/manage/mcg/ui/test_mcg_ui.py` and `tests/manage/mcg/ui/test_namespace_store.py` are currently under Black Squad ownership, and since analysis and maintenance require a lot of UI-specific knowledge, I think they should remain that way.
> 
> @DanielOsypenko wdyt?

agree, these two tests / modules are analyzed by the Black_squad. They are done mostly with UI automation via browser.

clacroix12:
> `tests/manage/mcg/ui/test_mcg_ui.py` and `tests/manage/mcg/ui/test_namespace_store.py` are currently under Black Squad ownership, and since analysis and maintenance require a lot of UI-specific knowledge, I think they should remain that way.
> 
> @DanielOsypenko wdyt?

I have updated those two modules to be owned by black squad, thanks!

tests/test_failure_propagator.py
clacroix12: I added a new marker: `ignore_owner` which we can apply to test cases to ignore them during the collection squad owner check. I've applied this to `test_failure_propagator`. We can update the logic in the test to apply the proper squad markers based on test errors in another PR to keep this one strictly focused on ownership.

tests/ui/test_creation_and_deletion_of_sc_and_rbdpool.py
```diff
@@ -21,6 +22,7 @@
 log = logging.getLogger(__name__)
 
 
+@black_squad
```
DanielOsypenko, line 25: The test was created and is maintained by Green Squad.
@shivamdurgbuns please confirm or reject.

ebenahar (approved):

tests/ui/test_creation_and_deletion_of_sc_and_rbdpool.py
clacroix12: Updated

DanielOsypenko: thanks

DanielOsypenko (approved):

tests/ecosystem/upgrade/test_upgrade.py
ebenahar: since this is part of upgrade testing, purple squad is the most logic fit IMO

ebenahar (approved):

petr-balogh (approved):

PR #8484, opened 2023-09-13
A new performance test - tests/e2e/performance/csi_tests/test_auto_reclaim_space_multi_clones.py

ypersky1980:
This PR contains the following: 

1) A new performance test tests/e2e/performance/csi_tests/test_auto_reclaim_space_multi_clones.py
that 
 - verifies creation of 513 reclaim space cron jobs for an rbd pvc and its 512 clones. 
 -  verifies that no reclaim space cron jobs are created for cephfs pvc and its clones
 
2) Changes to performance_lib ( a method wait_for_cronjobs was added there) 

3) Changes to tests/manage/pv_services/space_reclaim/test_auto_reclaim_space_cronjob.py
 - a method that is used in this test was moved to performance_lib

### DIFF
--- a/ocs_ci/helpers/performance_lib.py
+++ b/ocs_ci/helpers/performance_lib.py
@@ -1026,4 +1026,6 @@ def wait_for_cronjobs(namespace, cronjobs_num, msg, timeout=60):
             ):  # in the result one line is always a title
                 return sample
     except TimeoutExpiredError:
-        raise Exception(f"{msg} \n Cronjobs found: \n {sample}")
+        raise Exception(
+            f"{msg} \n Only {len(sample) -1} cronjobs found.\n This is the full list: \n {sample}"
+        )
--- a/ocs_ci/helpers/performance_lib.py
+++ b/ocs_ci/helpers/performance_lib.py
@@ -10,6 +10,9 @@ import re
 from ocs_ci.ocs.resources import pod
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants
+from ocs_ci.utility.utils import TimeoutSampler
+from ocs_ci.ocs.exceptions import TimeoutExpiredError
+
 
 logger = logging.getLogger(__name__)
 DATE_TIME_FORMAT = "%Y I%m%d %H:%M:%S.%f"
@@ -994,3 +997,33 @@ def pod_bulk_attach_csi_time(interface, pvc_objs, csi_start_time, namespace):
     )
 
     return total_time
+
+
+def wait_for_cronjobs(namespace, cronjobs_num, msg, timeout=60):
+    """
+    Runs 'oc get reclaimspacecronjob' with the TimeoutSampler
+
+    Args:
+        namespace(str): namespace in which cronjobs will be looked for
+        cronjobs_num (int): the exact number of cronjobs that should exist
+        msg (str): Error message to be printed if the desired condition is not reached
+        timeout (int): Timeout
+    Returns:
+
+        list : Result of 'oc get reclaimspacecronjob' command
+
+    """
+    try:
+        for sample in TimeoutSampler(
+            timeout=timeout,
+            sleep=5,
+            func=run_oc_command,
+            cmd="get reclaimspacecronjob",
+            namespace=namespace,
+        ):
+            if (
+                len(sample) == cronjobs_num + 1
+            ):  # in the result one line is always a title
+                return sample
+    except TimeoutExpiredError:
+        raise Exception(f"{msg} \n Cronjobs found: \n {sample}")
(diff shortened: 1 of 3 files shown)


pintojoy (approved):
LGTM

ypersky1980:
Please note that this PR contains 2 tests and I've ceparate validation run for each one of the tests. Both successfully passed.

ocs_ci/helpers/performance_lib.py
```diff
@@ -994,3 +997,35 @@ def pod_bulk_attach_csi_time(interface, pvc_objs, csi_start_time, namespace):
...
+        for sample in TimeoutSampler(
+            timeout=timeout,
+            sleep=5,
+            func=run_oc_command,
+            cmd="get reclaimspacecronjob",
+            namespace=namespace,
+        ):
+            if (
+                len(sample) == cronjobs_num + 1
+            ):  # in the result one line is always a title
```
ramkiperiy, line 1026: Is it line one is always a title?

tests/e2e/performance/csi_tests/test_auto_reclaim_space_multi_clones.py
```diff
@@ -0,0 +1,243 @@
...
+        super(TestReclaimSpaceCronJobMultiClones, self).setup()
+
+        self.num_of_clones = 512
+
+        # Getting the total Storage capacity
+        ceph_capacity = int(self.ceph_cluster.get_ceph_capacity())
+        # Getting the free Storage capacity
+        ceph_free_capacity = int(self.ceph_cluster.get_ceph_free_capacity())
+        # Use 70% of the free storage capacity in the test
+        capacity_to_use = int(ceph_free_capacity * 0.7)
```
ramkiperiy, line 40: General Question: Post 70% space utilisation there is still going to be any new IO in the cluster? or all the IO's or creation of new clones will be stopped immediately?

shivamdurgbuns (approved):

tests/e2e/performance/csi_tests/test_auto_reclaim_space_multi_clones.py
```diff
@@ -0,0 +1,243 @@
...
+            logger.info("Starting the test cleanup")
+
+            # Delete the created clones
+            try:
+                for clone in self.cloned_obj_list:
+                    performance_lib.run_oc_command(
+                        cmd=f"delete pvc {clone}", namespace=self.namespace
+                    )
+            except Exception:
+                logger.warning("Clones were not deleted")
```
ramkiperiy, line 92: since we have warning there is a possibility of printing this message and still continue to execute further cleanup i.e. before deleting clones PVC will be attempted to delete which might cause stale entries in PV.

ramkiperiy: Is there a option we loop back or re iterate to perform the deletion again? to make it successful and continue with further cleanup?

tests/e2e/performance/csi_tests/test_auto_reclaim_space_multi_clones.py
```diff
@@ -0,0 +1,243 @@
...
+            # Delete the test StorageClass
+            try:
+                logger.info(f"Deleting the test StorageClass : {self.sc_obj.name}")
+                self.sc_obj.delete()
+                logger.info("Wait until the SC is deleted.")
+                self.sc_obj.ocp.wait_for_delete(resource_name=self.sc_obj.name)
+            except Exception as ex:
+                logger.error(f"Can not delete the test sc : {ex}")
+
+            # Delete the test storage pool
```
ramkiperiy, line 106: again same possibility like my above comment, in-case if Clone and SC deletion fails but storage pool deletion is successful then we can't remove those clone, sc, PVC and PVs.
We have to handle it carefully.

tests/e2e/performance/csi_tests/test_auto_reclaim_space_multi_clones.py
ypersky1980: @ramkiperiy 
I see your point.
However, this is the way teardown works in all our tests - at first we delete the dependent entities ( like clones) and at the end we delete the PVC , pool and the namespace. The order of deletion is always opposite to the order of creation. 
Since this point is relevant not only to this test but to all the Performance tests - I'd suggest to leave it meanwhile as is and at some point to have a brainstorming to find the best way to cope with this issue.

tests/e2e/performance/csi_tests/test_auto_reclaim_space_multi_clones.py
ypersky1980: @ramkiperiy - Please see my previous comment.

keesturam (approved):

PR #8510, opened 2023-09-15
Ignore aro cluster operator in OCP upgrade

petr-balogh:


petr-balogh:
I did verification over upgrade OCP from 4.11 to 4.12 here:
https://url.corp.redhat.com/a45d1d0

As we don't have newer ARO than 4.12 I think we are OK to consider it as good enough as verified as this is quite straightforward change.

vavuthu (approved):

dahorak (approved):

PR #8521, opened 2023-09-18
Skip the test "test_toleration" when using MS

yitzhak12:
As per the discussion here: https://chat.google.com/room/AAAABfd8ivI/PItgYVWElNc, we decided that the test `test_toleration` is not relevant for MS. Therefore, I am skipping this test when using the Managed Service.


jilju (approved):

amr1ta (approved):
Looking good

am-agrawa (approved):

Issue #8328, opened 2023-08-23
`test_respin_mcg_pod_and_check_data_integrity_crd` sometimes respins the pod while objects are still being uploaded


sagihirshfeld:
In [this RP example](https://url.corp.redhat.com/6ae2531) we can see a "RPC CONNECTION CLOSED" error in the noobaa-endpoint logs while testfile1.txt is being uploaded via the PUT OBJECT method, and the time of the error is around the time of the respin of the pod.

This can easily be avoided by verifying that the upload has succeeded before proceeding to respin the pod.

PR #8552, opened 2023-09-21
Fixes object upload failure issue while respin is happening

udaysk23:
Fixes: #8328


tests/manage/mcg/test_namespace_crd.py
```diff
@@ -982,13 +983,28 @@ def test_respin_mcg_pod_and_check_data_integrity_crd(
...
+                s3_obj=mcg_obj,
+            )
+            if uploaded_obj_list == obj_list:
+                logger.info(f"Object list {obj_list}")
+                break
+            else:
+                logger.info("Objects are not synced yet retrying after 10 secs")
+                sleep(10)
+        else:
+            assert "Failed to sync objects"
```
sagihirshfeld, line 1007: This will never throw an exception, so the test won't fail here even if the upload wasn't complete. 

Python assertions throw an AssertError exception only if the expression evaluates as a "Falsey" value such as None, 0 or in this case, an empty string (check https://url.corp.redhat.com/f49ccda for info on "Falsey" and "Truthy" values in Python).

What you could do to fix this is to raise an exception instead:

```suggestion
            assert uploaded_obj_list == obj_list
            if uploaded_obj_list == obj_list:
                logger.info(f"Object list {obj_list}")
                break
            else:
                logger.info("Objects are not synced yet retrying after 10 secs")
                sleep(10)
        else:
            raise UnexpectedBehaviour("Failed to sync objects")
```

And while we're at it, adding a more descriptive Exception to ocs_ci.ocs.exceptions such as "SyncIncomplete" could be useful for future cases as well.

sagihirshfeld: Also, this kind of logic could be very useful in future test cases, so consider moving this to a new `bucket_utils.py` function. Something with a header like:

`assert_objects_in_bucket(bucket_name, objects_list, mcg_obj, s3pod,  tries=5, delay=10)`

sagihirshfeld (commented):
Besides for a small mistake - pretty solid. Please run an additional verification once its fixed.

Also, please make sure you include all the appropriate labels. In this case these would be:
`MCG`, `Squad/Red`, and since you have a passing validation job, `Verified`. Since you'll need to re-verify, removing the `Verified` label in favor of the `Needs Testing` would be the most transparent.

These labels are very helpful for keeping track of which PRs need our attention - especially the `Squad/Red` and the `Verified` labels.

sagihirshfeld:
> 
> Cluster Name: ukurundw-9 Cluster Configuration: PR Test Suite: tier4c PR Test Path: tests/manage/mcg/test_namespace_crd.py::TestNamespace::test_respin_mcg_pod_and_check_data_integrity_crd Additional Test Params: OCP VERSION: 4.14 OCS VERSION: 4.14 tested against branch: master
> 
> [Job UNSTABLE](https://url.corp.redhat.com/8ee34db) (some or all tests failed).

The failed verification could be the result of https://bugzilla.redhat.com/show_bug.cgi?id=2165907 due to the use of the `--force` option when we delete the noobaa-db pod in the first parameter of the TC, which then kept it stuck in CLBO. I can't tell for certain though because the must-gather logs weren't collected.

If you're using an existing cluster, you could follow Danny's instructions in this comment to fix the noobaa-db pod, so you could retry the verification:
https://bugzilla.redhat.com/show_bug.cgi?id=2165907#c27

sagihirshfeld (dismissed):
LGTM

ocs_ci/ocs/bucket_utils.py
```diff
@@ -139,6 +140,27 @@ def retrieve_test_objects_to_pod(podobj, target_dir):
     return downloaded_objects
 
 
+def check_objects_in_bucket(
+    bucket_name, objects_list, mcg_obj, s3pod, tries=5, delay=10
+):
+    """
+    Checks object list present in bucket and compare it with uploaded object Lists
+    """
+    for i in range(tries):
```
keesturam, line 149: Please see if you can use retry method instead - https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/retry.py#L8

ocs_ci/ocs/bucket_utils.py
```diff
@@ -139,6 +140,27 @@ def retrieve_test_objects_to_pod(podobj, target_dir):
...
+            f"s3://{bucket_name}",
+            s3_obj=mcg_obj,
+        )
+        if objects_list == obj_list:
+            logger.info(f"Object list {obj_list}")
+            break
+        logger.info(f"Objects are not synced yet retrying after {delay} secs")
+        sleep(delay)
+    else:
+        raise UnexpectedBehaviour("Failed to sync objects")
```
sagihirshfeld, line 161: Please return a bool value and move the exception raising to the test that uses this utility function. 

The need for this has became apparent when I found myself needing to verify that some objects are *not* in a bucket.

ocs_ci/ocs/bucket_utils.py
```diff
@@ -139,6 +140,23 @@ def retrieve_test_objects_to_pod(podobj, target_dir):
     return downloaded_objects
 
 
+@retry(UnexpectedBehaviour, tries=5, delay=10, backoff=1)
+def check_objects_in_bucket(bucket_name, objects_list, mcg_obj, s3pod):
```
sagihirshfeld, line 144: The check in `test_namespace_crd.py::test_respin_mcg_pod_and_check_data_integrity_crd` line 997 expects a potential false value that this won't return even if the bucket doesn't contain the objects - you need to return False instead of raising an exception here.  

However, the way you're currently using the retry function / decorator doesn't allow it. Also, because the timeout technicals are set in the decorator, this doesn't allow future users of this function to decide whether they want to wait 10 minutes instead of the ~1 minute that's currently set.

Please use retry (as a function and not a decorator) or the TimeoutSampler class on an anonymous function and return a boolean depending on the result, and add the necessary arguments to allow configuring the time.

ocs_ci/ocs/bucket_utils.py
```diff
@@ -139,6 +140,23 @@ def retrieve_test_objects_to_pod(podobj, target_dir):
...
+    """
+    Checks object list present in bucket and compare it with uploaded object Lists
+    """
+    obj_list = list_objects_from_bucket(
+        s3pod,
+        f"s3://{bucket_name}",
+        s3_obj=mcg_obj,
+    )
+    if objects_list == obj_list:
```
sagihirshfeld, line 153: The function is expected to return True if the bucket contains all the objects in a given list, but this only checks for exact matches this function will return False if the bucket has an additional object that is not in the given list.

Using python sets gives us an easy solution:

```suggestion
    if set(objects_list).issubset(obj_list):
```

ocs_ci/ocs/bucket_utils.py
sagihirshfeld: For example:

```
def is_remote_number_even(url, timeout=60):
    """
    Check if the number retrieved over HTTP is even.
     
    Args:
        timeout (int): Timeout in seconds
  
    Returns:
         bool: True if number is even, False otherwise
        
    """"
    def _is_remote_number_even(url_arg):
         return http_call_result(url_arg) % 2 == 0
     
    return any(result for result in TimeoutSampler(timetout, 10, _is_remote_number_even, url))          
```

ocs_ci/ocs/bucket_utils.py
udaysk23: Fixed

ocs_ci/ocs/bucket_utils.py
udaysk23: @sagihirshfeld Even if we try to use TimeoutSampler class, it will raise exception and it will not return false value once timeout gets completed

tiffanyn108 (approved):
LGTM

keesturam (approved):
LGTM

udaysk23:
Due to incorrect git filter issue, Closing this PR and reopening new PR for the same changes

PR #8556, opened 2023-09-22
Squad validation on collectonly

clacroix12:
To prevent the squad marker verification from blocking deployment or testing, we will only perform the validation if the pytest `--collectonly` option is set.

More changes here than I wanted but I had to change the name of the ocsci config object as it conflicted with the pytest `config` object that is automatically passed to this hook.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -274,7 +274,7 @@ def export_squad_marker_to_csv(items, filename=None):
     log.info("%s tests require action across %s files", num_tests, num_files)
 
 
-def pytest_collection_modifyitems(session, items):
+def pytest_collection_modifyitems(session, config, items):
     """
     A pytest hook to filter out skipped tests satisfying
     skipif_ocs_version, skipif_upgraded_from or skipif_no_kms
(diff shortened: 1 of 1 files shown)


dahorak (approved):

vavuthu (approved):

dahorak:
Deployment in the [PR verification job](https://url.corp.redhat.com/f9005f6) already passed, marking as verified.

tests/conftest.py
```diff
@@ -274,7 +274,7 @@ def export_squad_marker_to_csv(items, filename=None):
     log.info("%s tests require action across %s files", num_tests, num_files)
 
 
-def pytest_collection_modifyitems(session, items):
+def pytest_collection_modifyitems(session, config, items):
```
shivamdurgbuns, line 277: here it will be ocsci_config right?

tests/conftest.py
```diff
@@ -285,12 +285,13 @@ def pytest_collection_modifyitems(session, items):
...
-    teardown = config.RUN["cli_params"].get("teardown")
-    deploy = config.RUN["cli_params"].get("deploy")
-    skip_ocs_deployment = config.ENV_DATA["skip_ocs_deployment"]
+    teardown = ocsci_config.RUN["cli_params"].get("teardown")
+    deploy = ocsci_config.RUN["cli_params"].get("deploy")
+    skip_ocs_deployment = ocsci_config.ENV_DATA["skip_ocs_deployment"]
 
     # Verify tests are decorated with the correct squad owner
-    verify_squad_owners(items)
+    if config.option.collectonly:
```
shivamdurgbuns, line 293: here also ocsci_config instead of config?

PR #8619, opened 2023-10-03
Test OBCs UI access for regular user

ebondare:


DanielOsypenko:
hello @ebondare 
Do we need this PR? It was not tested before, right?

ebondare:
It wasn't tested before. I had some issue with this PR, I'll need to take a look

PR #8626, opened 2023-10-03
Warp - Added test cases for pv, aws, azure and gcp

tiffanyn108:
Increasing NooBaa Testing coverage:
* Added warp test run on pv, aws, azure and gcp

### DIFF
--- a/tests/e2e/scale/noobaa/test_warp.py
+++ b/tests/e2e/scale/noobaa/test_warp.py
@@ -55,25 +55,25 @@ class TestWarp(E2ETest):
             ),
             pytest.param(
                 *[1, "OC",
-                    {
-                        "interface": "OC",
-                        "backingstore_dict": {"aws": [(1, "eu-central-1")]}}],
+                  {
+                      "interface": "OC",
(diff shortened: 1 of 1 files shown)


ramkiperiy (dismissed):

tests/e2e/scale/noobaa/test_warp.py
```diff
@@ -38,9 +39,83 @@ class TestWarp(E2ETest):
...
+            ),
+            pytest.param(
+                *[
+                    1,
+                    "OC",
+                    {"interface": "OC", "backingstore_dict": {"gcp": [(1, None)]}},
+                ],
+                marks=[
+                    pytest.mark.skip(GCP_ACCOUNT_SKIP),
+                ],
```
sagihirshfeld, line 89: If this is here because of the missing / obsolete GCP credentials at auth.yaml?

If that's the case, then we can and should remove this mark because `bucket_factory` will skip the test when the ULS creation will fail. This way the test will resume running once the credentials are fixed without the need of an additional PR.

sagihirshfeld (approved):

udaysk23 (approved):

PR #8637, opened 2023-10-05
test_ceph_metrics_presence_when_osd_down avoid failures because of trying to reach OSD pods

DanielOsypenko:
failure case: https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/557/14972/694869/694945/694948/log

osd_to_stop = osds[-1]
E       IndexError: list index out of range

In an OCS external mode cluster, the OSD pods are managed by external ceph and may not be directly visible or manageable


DanielOsypenko:
@fbalak please check this case, if I am right

fbalak:
> @fbalak please check this case, if I am right

You are right but we should find a way how to access the external ceph cluster during a test run and shut down its components for alert testing and not just skip those tests.

PR #8657, opened 2023-10-09
A fixing to performance lib for correct calculation of pvc/clone deletion time in 4.14

ypersky1980:
This PR is a very much needed fix, since the deletion time notification has changed in the csi logs in ODF 4.14. 

Therefore we need to keep the old code for ODF <=4.13 and implement a new calculation starting 4.14.


keesturam (approved):

sagihirshfeld (approved):
LGTM

PR #8658, opened 2023-10-09
Fixes mirroring timeout issue

### LINKED ISSUE #7183: `test_multiregion_mirror` sometimes fails to complete mirroring in time
`MCG.check_if_mirroring_is_done` reaches its default timeout of 140 seconds before completing the mirroring, which might make sense on LOWERREQ environments due to the decreased resources available for the cluster. We should consider increasing the timeout on LOWERREQ deployments if we continue to see this issue.

RP Link for reference:
- https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/362/8945/395620/395621/395622/log

udaysk23:
Fixes #7183

### DIFF
--- a/ocs_ci/ocs/resources/mcg.py
+++ b/ocs_ci/ocs/resources/mcg.py
@@ -747,7 +747,7 @@ class MCG:
                 f"bucketclass create {placement_type}{name}{backingstores}{placement_policy}{replication_policy}"
             )
 
-    def check_if_mirroring_is_done(self, bucket_name, timeout=140):
+    def check_if_mirroring_is_done(self, bucket_name, timeout=300):
         """
         Check whether all object chunks in a bucket
         are mirrored across all backing stores.
@@ -756,12 +756,12 @@ class MCG:
             bucket_name: The name of the bucket that should be checked
             timeout: timeout in seconds to check if mirroring
 
-        Returns:
-            bool: Whether mirroring finished successfully
+        Raises:
+            AssertionError: In case mirroring is not done in defined time.
 
         """
 
-        def _check_mirroring():
+        def _get_mirroring_percentage():
             results = []
             obj_list = (
                 self.send_rpc_query(
@@ -798,21 +798,29 @@ class MCG:
                         results.append(True)
                     else:
                         results.append(False)
+            current_percentage = (results.count(True) / len(results)) * 100
+            return current_percentage
 
-            return all(results)
-
-        try:
-            for mirroring_is_complete in TimeoutSampler(timeout, 5, _check_mirroring):
-                if mirroring_is_complete:
-                    logger.info("All objects mirrored successfully.")
-                    return True
-                else:
-                    logger.info("Waiting for the mirroring process to finish...")
-        except TimeoutExpiredError:
-            logger.error(
-                "The mirroring process did not complete within the time limit."
-            )
-            assert False
+        mirror_percentage = _get_mirroring_percentage()
+        logger.info(f"{mirror_percentage}% mirroring is done.")
+        previous_percentage = 0
+        while mirror_percentage < 100:
+            previous_percentage = mirror_percentage
+            try:
+                for mirror_percentage in TimeoutSampler(
+                    timeout, 5, _get_mirroring_percentage
+                ):
+                    if previous_percentage == mirror_percentage:
+                        logger.warning("The mirroring process is stuck.")
+                    else:
+                        break
+            except TimeoutExpiredError:
+                logger.error(
+                    f"The mirroring process is stuck from last {timeout} seconds."
+                )
+                assert False
+            mirror_percentage = _get_mirroring_percentage()
+        logger.info("All objects mirrored successfully.")
 
     def check_backingstore_state(self, backingstore_name, desired_state, timeout=600):
         """


ocs_ci/ocs/resources/mcg.py
```diff
@@ -799,20 +799,41 @@ def _check_mirroring():
...
-
-        try:
-            for mirroring_is_complete in TimeoutSampler(timeout, 5, _check_mirroring):
-                if mirroring_is_complete:
+            return results
+
+        is_mirroring_complete = _check_mirroring()
+        current_percentage = (
+            is_mirroring_complete.count(True) / len(is_mirroring_complete)
+        ) * 100
```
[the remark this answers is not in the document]
sagihirshfeld: Also, both the names `_is_mirroring_complete` and `_check_mirroring()` are VERY misleading here:

- `is_mirroring_complete`: any variable name that starts with "is" is expected to be a boolean, but this is a list! A more fitting name for the existing implementation would be "mirror_status_flags" or even "mirror_complete_bool_list" to make it really obvious.

- `_check_mirroring`: this name was okay-ish for the original  implementation where it returned a bool, but now that it returns a list the name should give some sort of hint for this returned value. Something like `_get_mirroring_status` would suit better here.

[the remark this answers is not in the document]
sagihirshfeld: Finally, since `_check_mirroring()` is only used in this the scope of `check_if_mirroring_is_done()`, you can and should just modify it to return the result of this repeating logic, and change it's name to "_get_mirroring_percentage". Then you can call `_get_mirroring_percentage()` instead of repeating these 4 lines.

ocs_ci/ocs/resources/mcg.py
```diff
@@ -799,20 +799,41 @@ def _check_mirroring():
...
+        logger.info(f"{current_percentage}% mirroring is done.")
+        previous_percentage = 0
+        if current_percentage == 100:
+            logger.info("All objects mirrored successfully.")
+            return True
+        else:
```
sagihirshfeld, line 813: If current_percentage == 100 at line 810, we're exiting the method with the return True statement so there's no need to use an else: block that introduces an unnecessary additional level of indentation.

Also, the condition of the `while` statement at line 814 saves you the need for this if check entirely - exiting the while loop indicates that the mirroring has been complete, so just log the result and return True after it. This also applies for the `if` check at lines 834-836, so this helps to apply the DRY principle as well (let's say we want to change the logging info to include some data in the future here - you would have to change it in two places under this current implementation).

ocs_ci/ocs/resources/mcg.py
```diff
@@ -799,20 +799,41 @@ def _check_mirroring():
...
+            while current_percentage < 100:
+                previous_percentage = current_percentage
+                for i in range(timeout):
+                    is_mirroring_complete = _check_mirroring()
+                    current_percentage = (
+                        is_mirroring_complete.count(True) / len(is_mirroring_complete)
+                    ) * 100
+                    if previous_percentage == current_percentage:
+                        import time
```
sagihirshfeld, line 822: Imports should always be at the top of the file. Here are two reasons why this matters:

1.  Readability: we should strive to keep our code succinct to reduce the amount of time it takes to read and understand it. The use of the time module is obvious in line 824 so there's no need to include this import here - no new information of functionality is being added.

2. Unnecessary overhead that (slightly) affects performance.

sagihirshfeld: Also, I see that time.sleep is already imported at line 9 so you should just use `sleep` instead of `time.sleep`:
```
from time import sleep
```

ocs_ci/ocs/resources/mcg.py
```diff
@@ -799,20 +799,41 @@ def _check_mirroring():
...
+                        if i + 1 == timeout:
+                            logger.error("The mirroring process is stuck.")
+                            assert False
+                    else:
+                        break
+                is_mirroring_complete = _check_mirroring()
+                current_percentage = (
+                    is_mirroring_complete.count(True) / len(is_mirroring_complete)
+                ) * 100
+                if current_percentage == 100:
```
sagihirshfeld, line 834: This can be omitted by relying on the check of the while loop as described in a previous comment.

ocs_ci/ocs/resources/mcg.py
```diff
@@ -799,20 +799,41 @@ def _check_mirroring():
...
                     logger.info("All objects mirrored successfully.")
                     return True
```
sagihirshfeld, line 836: This should moved out of the while loop as described in a previous comment - exiting the loop means that the objects have been mirrored successfully.

sagihirshfeld (changes requested):

ocs_ci/ocs/resources/mcg.py
udaysk23: Fixed

ocs_ci/ocs/resources/mcg.py
udaysk23: Fixed

ocs_ci/ocs/resources/mcg.py
udaysk23: Fixed

ocs_ci/ocs/resources/mcg.py
udaysk23: Fixed

ocs_ci/ocs/resources/mcg.py
```diff
@@ -798,21 +798,26 @@ def _check_mirroring():
...
-                    return True
+            current_percentage = (results.count(True) / len(results)) * 100
+            return current_percentage
+
+        mirror_percentage = _get_mirroring_percentage()
+        logger.info(f"{mirror_percentage}% mirroring is done.")
+        previous_percentage = 0
+        while mirror_percentage < 100:
+            previous_percentage = mirror_percentage
+            for i in range(timeout):
```
keesturam, line 809: could we use retry here instead of a while loop?

udaysk23: @keesturam  We are resetting the values of timeout if we see some/any progress in sync. 
But with the retries, we will have fixed set of retries. In this case, even if there is sync in progress, it will probably fail after retries

ocs_ci/ocs/resources/mcg.py
```diff
@@ -798,21 +798,30 @@ def _check_mirroring():
...
+                    else:
+                        break
+            except TimeoutExpiredError:
+                logger.error(
+                    f"The mirroring process is stuck from last {timeout} seconds."
+                )
+                assert False
+            mirror_percentage = _get_mirroring_percentage()
+        logger.info("All objects mirrored successfully.")
+        return True
```
sagihirshfeld, line 824: Looking at the current usage of the method, we can remove this `return True` line. Please also update the docstring that this method now throws an exception in case mirroring was not done in time instead of returning a boolean value.

sagihirshfeld (commented):
Final minor suggestions - if you choose to incorporate the changes, a PR validation rerun won't be necessary as far as I'm concerned, because they don't affect the logic of the method.

sagihirshfeld (approved):
LGTM

tiffanyn108 (approved):

keesturam (approved):

PR #8660, opened 2023-10-10
Fix error with invalid XML character which breaks XML with report

petr-balogh:


petr-balogh:
https://url.corp.redhat.com/7737d15

vavuthu (approved):

dahorak (approved):

PR #8664, opened 2023-10-10
Increasing timeouts for bulk pod attach and bulk clone creation

ypersky1980:
This PR contains increasing timeouts for bulk pod attach and bulk clone creation.


keesturam (approved):

ramkiperiy:
@ypersky1980 I have couple of questions
1. This timeout is for complete while loop exit? am I correct?, while loop will be executed for every 30sec and will continue till 5400mins(1.30hrs)
2. Similar to creating PVC using kube_job could we enhance the script to validate the PVC bound status using kube_job? this will reduce lot of obj creation(memory issues) also it will reduce the script execution time. You can refer this code https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/scale_lib.py#L1251

ramkiperiy (approved):

PR #8673, opened 2023-10-12
test_pvc_snapshot_performance.py fix  - increasing get_fio_results timeout

ypersky1980:
For snapshot of 100GB size cephFS single snapshot test case was failing. 
I increased get_fio_results timeout from 600 ( default) to 1800.

### DIFF
--- a/tests/e2e/performance/csi_tests/test_pvc_snapshot_performance.py
+++ b/tests/e2e/performance/csi_tests/test_pvc_snapshot_performance.py
@@ -250,7 +250,7 @@ class TestPvcSnapshotPerformance(PASTest):
             self.pod_object.fillup_fs(size=file_size, fio_filename=file_name)
 
             # Wait for fio to finish
-            fio_result = self.pod_object.get_fio_results()
+            fio_result = self.pod_object.get_fio_results(timeout=1800)
             err_count = fio_result.get("jobs")[0].get("error")
             assert (
                 err_count == 0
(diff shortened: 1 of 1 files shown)


pintojoy (dismissed):
LGTM

tests/e2e/performance/csi_tests/test_pvc_snapshot_performance.py
```diff
@@ -250,7 +250,7 @@ def test_pvc_snapshot_performance(self, pvc_size):
             self.pod_object.fillup_fs(size=file_size, fio_filename=file_name)
 
             # Wait for fio to finish
-            fio_result = self.pod_object.get_fio_results()
+            fio_result = self.pod_object.get_fio_results(timeout=3600)
```
ramkiperiy, line 253: This is default fio profile right? hope we validated properly for the timeout. 
I have faced something similar in OOM kill TC fio didn't complete in expected time. Eventually I stopped using this func and instead used perf image and the custom profile so there will be more control on handling the IO, again this is just a suggestion will leave it to you on how you want it.

ypersky1980: @ramkiperiy  - yes, we run default FIO load, and with the increased timeout it works ok. 
This is now it works in all the previous versions as well, I wouldn't change the IO type since it might affect the performance comparison.

ypersky1980:
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-trigger-test-pr/5965/testReport/  - Please note that in this validation job all the tests passed, except the results which is expected in IBM cloud. Therefore this PR passed validation and can be merged.

pintojoy (approved):
LGTM

ramkiperiy (approved):
lgtm

Issue #8679, opened 2023-10-13
When no test run the time report prints traceback


petr-balogh:
```
------------------------------------- live log sessionfinish -------------------------------------
14:34:31 - MainThread - ocs_ci.framework.pytest_customization.reports - ERROR  - Failed to save Test Time report to logs directory
Traceback (most recent call last):
  File "/home/jenkins/workspace/qe-deploy-ocs-cluster/ocs-ci/ocs_ci/framework/pytest_customization/reports.py", line 104, in pytest_sessionfinish
    with open(time_report_file, "a") as fil:
FileNotFoundError: [Errno 2] No such file or directory: '/home/jenkins/current-cluster-dir/logs/ocs-ci-logs-1697207667/session_test_time_report_file.csv'
```

PR #8708, opened 2023-10-19
Adding FileNotFoundError exception

shivamdurgbuns:
Fixes Issue: https://github.com/red-hat-storage/ocs-ci/issues/8679

### DIFF
--- a/ocs_ci/framework/pytest_customization/reports.py
+++ b/ocs_ci/framework/pytest_customization/reports.py
@@ -114,8 +114,8 @@ def pytest_sessionfinish(session, exitstatus):
                 ]
                 c.writerow(row)
         logger.info(f"Test Time report saved to '{time_report_file}'")
-    except (Exception, FileNotFoundError) as e:
-        logger.exception(
+    except Exception as e:
+        logger.warning(
             f"Failed to save Test Time report to logs directory with exception. {e}"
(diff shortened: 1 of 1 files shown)


ocs_ci/framework/pytest_customization/reports.py
```diff
@@ -114,8 +114,10 @@ def pytest_sessionfinish(session, exitstatus):
                 ]
                 c.writerow(row)
         logger.info(f"Test Time report saved to '{time_report_file}'")
-    except Exception:
-        logger.exception("Failed to save Test Time report to logs directory")
+    except (Exception, FileNotFoundError) as e:
```
dahorak, line 117: The `FileNotFoundError` a child of the general `Exception`, so it is not necessary to mention it here:
```suggestion
    except Exception as e:
```

The traceback is printed to the logs because it is automatically added by the [`logger.exception(...)`](https://docs.python.org/3/library/logging.html#logging.Logger.exception) method. You probably want to change the log level to _debug_, _info_ or _warning_ (not sure if it is necessary to print it with `error` level).

dahorak (changes requested):

dahorak (approved):

petr-balogh (approved):

PR #8711, opened 2023-10-19
Wait for node deletion and verify original node count during teardown of test_automated_recovery_from_stopped_node_and_start[True]

### LINKED ISSUE #8671: Wait for node deletion and verify original node count during teardown of test_automated_recovery_from_stopped_node_and_start[True]
The next test **test_automated_recovery_from_stopped_node_and_start[False]** failed as it did not find the 3rd worker node due to which Pods scheduled on that node were stuck in Pending state.

This happened because we are deleting the newly added node in the previous test **test_automated_recovery_from_stopped_node_and_start[True]** during teardown but do not wait for it's deletion and ensure that the original node count is maintained.
The check has to be implemented with proper timeout to avoid hitting such issues with other tests.

Jenkins- https://url.corp.redhat.com/fe93967

However, the failure seems intermittent as it passed with ODF  4.14.0-145.stable but it's a bit strange as we were anyways expecting at least 3 worker nodes to be available when the next test was run as the only extra node which was added in the previous test was bring deleted, and not more.

am-agrawa:
fixes #8671

### DIFF
--- a/tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_reactive_IPI.py
+++ b/tests/manage/z_cluster/nodes/test_automated_recovery_from_failed_nodes_reactive_IPI.py
@@ -35,6 +35,7 @@ from ocs_ci.ocs.node import (
     get_node_pods,
     wait_for_nodes_racks_or_zones,
     wait_for_nodes_status,
+    wait_for_node_count_to_reach_status,
 )
 from ocs_ci.ocs.exceptions import ResourceWrongStatusException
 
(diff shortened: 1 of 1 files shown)


am-agrawa:
Both the parameterized test passed- https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/30680/console

ebenahar (approved):

pintojoy (approved):
LGTM

amr1ta (approved):

pintojoy (approved):
LGTM

PR #8736, opened 2023-10-25
added vsphere yaml for the Jenkins deployment

DanielOsypenko:
for Provider / Client deployment


dahorak (approved):

fbalak (approved):

conf/deployment/vsphere/upi_1az_rhcos_vsan_3m_3w_provider_client.yaml
```diff
@@ -0,0 +1,15 @@
+---
+# This config is suppose to work on DC-CP, DC-ECO, DC-PS
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
+ENV_DATA:
+  enable_hw_virtualization: true
+  platform: 'vsphere'
+  deployment_type: 'upi'
+  worker_replicas: 3
+  master_replicas: 3
```
petr-balogh, line 10: client doe't have masters right?
Do we still need to include them in config because of some compatibility?
Same for another lines with `master_` prefix.

DanielOsypenko: this is Provider configuration as a base for deploying both provider and kubevirt clients. Client would not have config file.

PR #8762, opened 2023-10-30
deployment yaml for provider-client with specs for master nodes added

DanielOsypenko:


conf/deployment/ibmcloud/ipi_3az_rhcos_3m_3w_provider_client.yaml
```diff
@@ -0,0 +1,29 @@
+---
+DEPLOYMENT:
+  openshift_install_timeout: 4800
+  allow_lower_instance_requirements: false
+ENV_DATA:
+  platform: 'ibm_cloud'
+  deployment_type: 'ipi'
+  region: 'us-south'
+  base_domain: 'ibmcloud2.qe.rh-ocs.com'
+  master_instance_type: 'bx2-16x64'
```
prsurve, line 10: This will not work as there is no IBM cloud template for master node it's empty [ see here](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/templates/ocp-deployment/install-config-ibm_cloud-ipi.yaml.j2#L17 )

DanielOsypenko: You are right, I already see it on deployed cluster. 
What should I do instead? I need similar 16x64 machines for both workers and masters

DanielOsypenko:
nor relevant because this type of deployments failed

PR #8782, opened 2023-11-02
Fix Multus Deployment

OdedViner:
https://bugzilla.redhat.com/show_bug.cgi?id=2245978
OSD YAML before fix:
```
 k8s.v1.cni.cncf.io/networks: openshift-storage/private-net, openshift-storage/public-net
```

OSD YAML after fix:
```
k8s.v1.cni.cncf.io/networks: '[{"name":"public-net","namespace":"openshift-storage"},{"name":"private-net","namespace":"openshift-storage"}]'

```

### DIFF
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -1844,7 +1844,8 @@ def verify_networks_in_ceph_pod(pod_networks, net_name, net_namespace):
         net_name (str): the network-attachment-definitions name
         net_namespace (str): the network-attachment-definitions namespace
 
-    bool: return True if net_name and net_namespce exist in pod_networks otherwise False
+    Returns:
+        bool: return True if net_name and net_namespce exist in pod_networks otherwise False
 
     """
     ocs_version = version.get_semantic_ocs_version_from_config()
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -1730,8 +1730,8 @@ def verify_multus_network():
             "k8s.v1.cni.cncf.io/networks"
         ]
         if public_net_created:
-            assert (
-                public_net_name in pod_networks
+            assert verify_networks_in_ceph_pod(
+                pod_networks, public_net_name, public_net_namespace
             ), f"{public_net_name} not in {pod_networks}"
 
         osd_id = get_osd_pod_id(_pod)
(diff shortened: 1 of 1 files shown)


petr-balogh (dismissed):

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -1731,8 +1731,8 @@ def verify_multus_network():
         ]
         if public_net_created:
             assert (
-                public_net_full_name in pod_networks
-            ), f"{public_net_full_name} not in {pod_networks}"
+                public_net_name in pod_networks
```
clacroix12, line 1734: Should we be a bit more explicit here? This no longer verifies that the namespace is correct. Since the new formatting is as such: 

```
k8s.v1.cni.cncf.io/networks: '[{"name":"public-net","namespace":"openshift-storage"},{"name":"private-net","namespace":"openshift-storage"}]'
```

We could do something like this:

```
networks = [{"name":"public-net","namespace":"openshift-storage"},{"name":"private-net","namespace":"openshift-storage"}]  # would need to convert pod_networks to an actual list like this

assert(any((network["name"] == public_net_name and network["namespace"] == public_net_namespace) for network in networks))

```
You'll need to convert the networks string into an actual list to parse it like that but this way we don't lose the ability to verify that the namespace is correct.

OdedViner: done. 
@clacroix12 Can you check my fix?

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -1835,6 +1835,32 @@ def verify_multus_network():
...
+def verify_networks_in_ceph_pod(pod_networks, net_name, net_namespace):
+    """
+    Verify network configuration on ceph pod
+
+    Args:
+        pod_networks (str): the value of k8s.v1.cni.cncf.io/networks param
+        net_name (str): the network-attachment-definitions name
+        net_namespace (str): the network-attachment-definitions namespace
+
+    bool: return True if net_name and net_namespce exist in pod_networks otherwise False
```
vavuthu, line 1847:
```suggestion
    Returns:
        bool: return True if net_name and net_namespce exist in pod_networks otherwise False
```

OdedViner: done

vavuthu (approved):

petr-balogh (approved):

PR #8823, opened 2023-11-09
Draft of new configuration files for Fusion HCI provider client solution

suchita-g:


conf/deployment/fusion_hci_pc/client_bm_upi_1az_rhcos_nvme_3m_3w.yaml
```diff
@@ -0,0 +1,13 @@
+---
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
+  local_storage: true
+ENV_DATA:
+  platform: 'baremetal'
```
fbalak, line 6: Should we use `baremetal` as platform? The deployment automation will differ from what we have for baremetal afaiu. Maybe we should introduce a new platform for fusion hci.

conf/deployment/fusion_hci_pc/provider_bm_upi_1az_rhcos_nvme_3m_3w.yaml
```diff
@@ -0,0 +1,13 @@
+---
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
+  local_storage: true
+ENV_DATA:
+  platform: 'baremetal'
+  cluster_type: 'provider'
```
ebondare, line 7: If we're going to have hci_client, why don't we introduce hci_provider to distinguish it from old provider cluster?

conf/deployment/fusion_hci_pc/client_bm_upi_1az_rhcos_nvme_3m_3w.yaml
```diff
@@ -0,0 +1,13 @@
...
+  cluster_type: 'hci_client'
+  cluster_namespace: "openshift-storage"
```
ebondare, line 8: What about the namespace change? Don't we want to specify here the new client namespace?

conf/deployment/fusion_hci_pc/provider_bm_upi_1az_rhcos_nvme_3m_3w.yaml
suchita-g: To minimise the changes for exiting working `runs_on_provider`

conf/deployment/fusion_hci_pc/client_bm_upi_1az_rhcos_nvme_3m_3w.yaml
suchita-g: The base platform/hardware where we can use to deploy this setup can be 

- vsphere 
- IBM BM instances
- baremetal

So for deployment we need either of above and that specific configuration then HCP/HCI specific configuration we need to update in above draft.  I am using the baremetal hardware on which this setup is deployed so just added this 2 configuration

conf/deployment/fusion_hci_pc/client_bm_upi_1az_rhcos_nvme_3m_3w.yaml
suchita-g: good catch. Copy paste mistake. Correcting it

DanielOsypenko (approved):

fbalak (approved):

conf/deployment/fusion_hci_pc/client_bm_upi_1az_rhcos_nvme_3m_3w.yaml
fbalak: I still think that this attribute should be different because based on this is selected installation class and platform specific checks. This will be different from standard baremetal deployment but we can keep it for now and change it when needed.

fbalak (dismissed):

ebondare (dismissed):

conf/deployment/fusion_hci_pc/client_bm_upi_1az_rhcos_nvme_3m_3w.yaml
ebenahar: Filip has a point. 
If set to `baremetal` here it could cause certain conditions in ocs-ci to be met.
Here is an example https://github.com/red-hat-storage/ocs-ci/blob/1e93e43af7cb6c44760e9c720f58c28740317bcb/ocs_ci/ocs/platform_nodes.py#L84

conf/deployment/fusion_hci_pc/client_bm_upi_1az_rhcos_nvme_3m_3w.yaml
```diff
@@ -0,0 +1,13 @@
...
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
+  local_storage: true
+ENV_DATA:
+  platform: 'baremetal'
+  cluster_type: 'hci_client'
+  cluster_namespace: "openshift-storage-client"
+  deployment_type: 'upi'
+  worker_replicas: 3
+  master_replicas: 3
```
ebenahar, line 11: Do we indeed have 3 masters? I remember it was mentioned that the node types differ on Fusion HCI from regular 3+3 deployment

conf/deployment/fusion_hci_pc/provider_bm_upi_1az_rhcos_nvme_3m_3w.yaml
```diff
@@ -0,0 +1,13 @@
+---
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
+  local_storage: true
+ENV_DATA:
+  platform: 'baremetal'
```
ebenahar, line 6: same comment regarding BM

fbalak (approved):

ebondare (approved):

fbalak (approved):

ebenahar (approved):

Issue #8834, opened 2023-11-13
RDR tests failed on setup with "ocs_ci.ocs.exceptions.UnexpectedDeploymentConfiguration: ACM not found" 


sidhant-agrawal:
RDR test failed on setup with "ocs_ci.ocs.exceptions.UnexpectedDeploymentConfiguration: ACM not found" 

Error details:
```
failed on setup with "ocs_ci.ocs.exceptions.UnexpectedDeploymentConfiguration: ACM not found"
```

```
request = <SubRequest 'setup_acm_ui' for <Function test_failover_and_relocate[primary_down_appset]>>

    @pytest.fixture(scope="function")
    def setup_acm_ui(request):
>       return setup_acm_ui_fixture(request)

tests/conftest.py:4779: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/conftest.py:4783: in setup_acm_ui_fixture
    driver = login_to_acm()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def login_to_acm():
        """
        Login to ACM console and validate by its title
    
        Returns:
            driver (Selenium WebDriver)
    
        """
    
        acm_version = ".".join(get_running_acm_version().split(".")[:2])
        if not acm_version:
>           raise UnexpectedDeploymentConfiguration("ACM not found")
E           ocs_ci.ocs.exceptions.UnexpectedDeploymentConfiguration: ACM not found

ocs_ci/ocs/acm/acm.py:429: UnexpectedDeploymentConfiguration
```

PR #8836, opened 2023-11-13
Switch context to ACM cluster before login to ACM console

sidhant-agrawal:
Fixes: #8834

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4780,10 +4780,14 @@ def setup_acm_ui(request):
 
 
 def setup_acm_ui_fixture(request):
+    restore_ctx_index = ocsci_config.cur_index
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -4780,6 +4780,7 @@ def setup_acm_ui(request):
 
 
 def setup_acm_ui_fixture(request):
+    ocsci_config.switch_acm_ctx()
```
shylesh, line 4783: @sidhant-agrawal , what happens after the test ? context will be still with ACM. I suggest you to store old context during setup and switch to old context in the finalizer so that ocs-ci works in that old context . WDYT ?

shylesh (approved):
LGTM

am-agrawa (approved):

PR #8895, opened 2023-11-20
Adding skipif_hci_client marker with reference to skipif_ms_consumer

suchita-g:


suchita-g:
Verification Job: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/2086/

ebenahar (approved):

amr1ta (approved):
Looking good.

yitzhak12 (approved):

petr-balogh (approved):

tests/manage/z_cluster/nodes/test_node_replacement_proactive.py
```diff
@@ -258,6 +259,7 @@ def test_nodereplacement_proactive_with_io_running(
 @skipif_bmpsi
 @skipif_external_mode
 @skipif_ms_consumer
+@skipif_hci_client
```
petr-balogh, line 262: Maybe we can just check if it's HCI muliticluster execution, if so, to enforce switching the context of config for provider, run tests on provider clutster and return back context of config back to origin after the test will finis.   But for now I am ok to disable those, but in future we can cover those cases like this I think.

petr-balogh (approved):

PR #8918, opened 2023-11-22
Allow re-try for ceph health for all cases

petr-balogh:


OdedViner (dismissed):

ebenahar (approved):

OdedViner (dismissed):

ebenahar (approved):

tests/conftest.py
```diff
@@ -1617,21 +1616,13 @@ def finalizer():
...
-                            namespace=ocsci_config.ENV_DATA["cluster_namespace"]
-                        )
-                        log.info(
-                            "Ceph health check passed at teardown. (After test "
-                            "marked with @ceph_health_retry. For such TC we allow more re-tries)"
-                        )
-                    else:
-                        ceph_health_check()
-                        log.info("Ceph health check passed at teardown")
+                    ceph_health_check(
```
vavuthu, line 1619: this will return True as soon as health is OK. but the issue we have is flip-flopping of ceph health. After upgrade health is OK and by the time it comes to teardown health is in HEALTH_WARN. either we can handle that in the test cases itself like ceph health should be stable after upgrade for some consecutive retries or in teardown.

tests/conftest.py
```diff
@@ -1617,21 +1616,13 @@ def finalizer():
...
+                        namespace=ocsci_config.ENV_DATA["cluster_namespace"]
+                    )
+                    # We are allowing 20 re-tries for health check, to avoid teardown failures for cases like:
+                    # "flip-floping ceph health OK and warn because of:
+                    # HEALTH_WARN Reduced data availability: 2 pgs peering
```
vavuthu, line 1624: could you move this comment before calling ceph_health_check

OdedViner (approved):

tests/conftest.py
ebenahar: How about we add a condition and a decorator that would re-check Ceph health again, even if the first check passed and health was OK?
We can add this marker to the upgrade test for example, which tend to have this flipping Ceph health status at the end

petr-balogh: I would merge this as is. If we see some issues and that it's not working I would implement it as new PR to add decorator and logic which will make sure several times ceph is OK,  and if we have some tests which really require ceph to be health OK those tests should be decorated

ebenahar: sounds good

ebenahar (approved):

tests/conftest.py
vavuthu: ack

PR #8919, opened 2023-11-22
Allow re-try for ceph health for all cases

petr-balogh:


petr-balogh:
Verification:
https://url.corp.redhat.com/07de908

tests/conftest.py
```diff
@@ -1492,21 +1491,11 @@ def finalizer():
...
-                            "Ceph health check passed at teardown. (After test "
-                            "marked with @ceph_health_retry. For such TC we allow more re-tries)"
-                        )
-                    else:
-                        ceph_health_check_base()
-                        log.info("Ceph health check passed at teardown")
+                    ceph_health_check(namespace=config.ENV_DATA["cluster_namespace"])
+                    # We are allowing 20 re-tries for health check, to avoid teardown failures for cases like:
+                    # "flip-floping ceph health OK and warn because of:
+                    # HEALTH_WARN Reduced data availability: 2 pgs peering
```
vavuthu, line 1497: could you move this comment before calling ceph_health_check

tests/conftest.py
```diff
@@ -1492,21 +1491,11 @@ def finalizer():
...
-                            namespace=config.ENV_DATA["cluster_namespace"]
-                        )
-                        log.info(
-                            "Ceph health check passed at teardown. (After test "
-                            "marked with @ceph_health_retry. For such TC we allow more re-tries)"
-                        )
-                    else:
-                        ceph_health_check_base()
-                        log.info("Ceph health check passed at teardown")
+                    ceph_health_check(namespace=config.ENV_DATA["cluster_namespace"])
```
vavuthu, line 1494: this will return True as soon as health is OK. but the issue we have is flip-flopping of ceph health. After upgrade health is OK and by the time it comes to teardown health is in HEALTH_WARN. either we can handle that in the test cases itself like ceph health should be stable after upgrade for some consecutive retries or in teardown.

OdedViner (approved):

tests/conftest.py
petr-balogh: the discussion is the same as for master PR: https://github.com/red-hat-storage/ocs-ci/pull/8918#discussion_r1403077647

ebenahar (approved):

PR #8933, opened 2023-11-23
Relocate functional MCG and RGW tests from `tests/manage` to `tests/functional/object`

sagihirshfeld:
This PR moves most of the Red Squad functional MCG and RGW tests to `tests/functional/object` in order to comply with the new test directory structure that was agreed upon in https://url.corp.redhat.com/d9003a4. 

The tests that were moved in this PR are still using the same conftest files, so they shouldn't require any adjustments in order to run as they did in their original location. The remaining tests do require some adjustments:

- MCGt tests under `tests/e2e/flowtest/mcg` and under `tests/e2e/lifecycle/mcg` depend on fixtures defined in `tests/e2e/conftest.py`, so we'll need to carefully move these fixtures to their own conftest files under the directories.

- MCG tests under `tests/ecosystem/upgrade` would also need some fixtures moved from `tests/ecosystem/conftest.py`, and `test_noobaa_service_mon_after_ocs_upgrade` would be need to be moved from `tests/ecosystem/upgrade/test_resources.py`


sagihirshfeld:
Verified locally after rebasing to the latest master by using pytest's `--setup-plan` option.

This option shows what fixtures and tests would be executed without running them, and if any test is attempting to call a fixture that's not available to it says so in the result, as demonstrated here by adding a call to `used_by_not_implemented_fixture` to one of the tests:
```
$ python3.8 -m pytest --setup-plan tests/functional/object/mcg/test_bucket_replication.py | grep used_by_not_implemented_fixture
E       fixture 'used_by_not_implemented_fixture' not found
E       fixture 'used_by_not_implemented_fixture' not found
E       fixture 'used_by_not_implemented_fixture' not found
E       fixture 'used_by_not_implemented_fixture' not found
E       fixture 'used_by_not_implemented_fixture' not found
E       fixture 'used_by_not_implemented_fixture' not found
``` 

Therefor, the following output that shows that the "fixture '<some_fixture>' not found" pattern is found in the result of the command on `ocs-ci/tests` is enough to show that all the tests still have access to their fixtures after the relocation:
```
$ python3.8 -m pytest --setup-plan ocs-ci/tests/ | grep "fixture '.*' not found" | wc -l
       0
```

keesturam (approved):

keesturam (approved):

udaysk23 (approved):

PrasadDesala (approved):

Issue #8967, opened 2023-11-30
test_dashboard_validation_ui some locators changed on ODF 4.15


OdedViner:
https://url.corp.redhat.com/bf9d983

DanielOsypenko:
it is one of the most stable tests in UI 
![image](https://github.com/red-hat-storage/ocs-ci/assets/61454420/2c5312f2-f7c6-4631-a03c-4d3d08a97bac)

This issue was not closed but the PR removing "acceptence" mark was merged long ago.
Do we want return acceptance? - if to vote, I am in favor of it.

@ebenahar

am-agrawa:
[...]

PR #8968, opened 2023-11-30
Skip test_dashboard_validation_ui from acceptance suite

OdedViner:
#8967

### DIFF
--- a/tests/ui/test_validation_ui.py
+++ b/tests/ui/test_validation_ui.py
@@ -17,7 +17,6 @@ from ocs_ci.framework.pytest_customization.marks import (
     skipif_external_mode,
     skipif_mcg_only,
     skipif_ibm_cloud_managed,
-    acceptance,
 )
 from ocs_ci.ocs.ui.validation_ui import ValidationUI
 from ocs_ci.utility import version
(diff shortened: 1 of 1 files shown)


DanielOsypenko:
hello @OdedViner. Copying suggestion from Petr on this topic here

cc @petr-balogh
> 
> I think that we should leverage the ui marker for ui tests.
> 
> As I see the test is already decorated with @ui mark then we can just update the mark expression when executing run-ci:
> from: 
> -m "acceptance"
> 
> change to:
> -m "acceptance and not ui"
> 
> This will avoid run ui test from your end without need changing anything from framework side.
> 
> [@Lital Alon](mailto:lalon@redhat.com) can you update it from your CI on line where you executing run-ci ?
> 
> Thanks,
> Petr

Did not it work out?

petr-balogh:
> Did not it work out?

Because of OCP 4.15 moved to PatternFly 5, (There is also one mail thread from Oded just added you to the loop), we will need to adjust a lot of locators for our tests - once ODF will move as well. I suggested to Oded to skip it now to unblock the stable pipeline and have all tests passing till we will fix all the UI tests in some of follow up PRs when we will enable this test back.
Also a bit relevant EPIC:  https://url.corp.redhat.com/aeade26 when Oded added comment. But the change came from OCP, not yet ODF based on discussion in mentioned email.

vavuthu (approved):

am-agrawa (approved):

PR #8975, opened 2023-12-03
Skip unsupported Feature Encryption, new SC  and NFS TC for HCI PC Platform

suchita-g:
Skip the Test after Tier1 and Tier2 analysis for unsupported Feature:
1. NFS
2. Encryption
3. TC which involves the creation of new storageclass
4. Adjust test step to skip verification with  for above unsupported feature


jilju (approved):

DanielOsypenko:
@suchita-g do you recall, why NFS tests were disabled on provider/client?
at least based on this story it is supported with ODF 4.19
https://issues.redhat.com/browse/RHSTOR-6922 
cc @amr1ta

Issue #8909, opened 2023-11-21
[HCI Provider-Client] In Tier1 Execution 54 TCs has Error: Error from server (AlreadyExists): error when creating "/tmp/StatefulSetpxyehdug": statefulsets.apps "s3cli" already exists"


suchita-g:
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/2105/testReport/

failed on setup with "ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc -n openshift-storage-client create -f /tmp/StatefulSetpxyehdug -o yaml.
Error is Error from server (AlreadyExists): error when creating "/tmp/StatefulSetpxyehdug": statefulsets.apps "s3cli" already exists"

`request = <SubRequest 'awscli_pod_session' for <Function test_bucket_delete_with_objects[CLI]>>

    @pytest.fixture(scope="session")
    def awscli_pod_session(request):
        awscli_pods = get_pods_having_label(
            constants.S3CLI_LABEL, ocsci_config.ENV_DATA["cluster_namespace"]
        )
        existing_pod = Pod(**awscli_pods[0]) if len(awscli_pods) > 0 else None
>       return existing_pod or awscli_pod_fixture(request, scope_name="session")

tests/conftest.py:2421: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/conftest.py:2465: in awscli_pod_fixture
    s3cli_sts_obj = helpers.create_resource(**awscli_sts_dict)
ocs_ci/helpers/helpers.py:89: in create_resource
    created_resource = ocs_obj.create(do_reload=do_reload)
ocs_ci/ocs/resources/ocs.py:117: in create
    status = self.ocp.create(yaml_file=self.temp_yaml)
ocs_ci/ocs/ocp.py:367: in create
    output = self.exec_oc_cmd(command)
ocs_ci/ocs/ocp.py:178: in exec_oc_cmd
    out = run_cmd(
ocs_ci/utility/utils.py:484: in run_cmd
    completed_process = exec_cmd(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[...]

PR #8980, opened 2023-12-03
Refactor awscli fixture to avoid "s3cli already exists" leftovers

sagihirshfeld:
Fixes: #8909

Two measures were taken to address the issue:
1. Setting up the teardown function so it'll actually get called if the setup fails at any point.
2. Making sure there are no leftovers from previous runs by calling said teardown function before attempting to create the resources.


DanielOsypenko (approved):

keesturam (approved):

keesturam:
Neatly done @sagihirshfeld

PR #8989, opened 2023-12-05
Add runs_on_provider marker in the tests test_monitoring_defaults

jilju:
Add runs_on_provider marker in the tests test_monitoring_defaults. This will make the test to run on the provider cluster while using a multicluster configuration.


fbalak (approved):

jilju:
Verified:

http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/hcp4142-s1/hcp4142-s1_20231127T065124/multicluster/logs/test_report_1701759251.html

The test cases ran on provider cluster with this code change. Some tests failed because the cluster health was not good(ran with --dev-mode parameter)

ebenahar (approved):

PR #8993, opened 2023-12-05
Refined in-transit encryption test cases.

paraggit:
Following changes done.

1.  reduce waiting time to confirm Ceph configs.
2. Removed check for Ceph protocol version v2 check since it's being by default.
3. Fixed tests cases error by initialising the empty list `self.mons = []` at the begning   of test case.


ebenahar (approved):

nehaberry (approved):
LGTM

PR #8994, opened 2023-12-05
Create assisted_installer low level api module

dahorak:
Creating low-level api module for interaction with Assisted Installer API

### DIFF
--- a/ocs_ci/utility/assisted_installer.py
+++ b/ocs_ci/utility/assisted_installer.py
@@ -1,0 +1,505 @@
+# -*- coding: utf8 -*-
+"""
+This module is for interaction with Assisted Installer API
+"""
+
+import logging
+import os
+import requests
+import time
+import re
+from urllib.parse import urljoin
+
+from ocs_ci.framework import config
+
+from ocs_ci.ocs.exceptions import OpenShiftAPIResponseException
+
+logger = logging.getLogger(__name__)
+
+
+class OpenShiftAPI(object):
+    """
+    Common class for interaction with various OpenShift APIs
+    """
+
+    def __init__(self, environment="production"):
+        """
+        Initialize required variables
+
+        Args:
+            environment (str): Environment defined in configuration (default: production)
+
+        """
+        # TODO: check required configuration
+        self.env = environment
+
+        self._token = None
+        self._token_exp = 0
+
+        # Load configuration
+        # SSO URL for generating API token
+        self.sso_url = config.AUTH["assisted_installer"][self.env]["sso_url"]
+        # Offline token used for generating API token
+        self.offline_token = config.AUTH["assisted_installer"][self.env][
+            "offline_token"
+        ]
+        # Assisted Installer Console API url
+        self.api_host_url = config.AUTH["assisted_installer"][self.env]["api_url"]
+
+    @property
+    def token(self):
+        """
+        Property for obtaining API token (based on OFFLINE_TOKEN)
+        """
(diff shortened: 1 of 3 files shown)


dahorak:
It is not possible to test this module via Jenkins job, but I've tested it together with changes from [PR#9042](https://github.com/red-hat-storage/ocs-ci/pull/9042/files) and other changes related to the deployment via Assisted Installer on vSphere locally and it worked as expected, so I would consider it as verified.

ocs_ci/utility/assisted_installer.py
```diff
@@ -0,0 +1,517 @@
...
+        # TODO: check required configuration
+        self.env = environment
+
+        self._token = None
+        self._token_exp = 0
+
+        # Load configuration
+        # SSO URL for generating API token
+        self.sso_url = config.AUTH["assisted_installer"][self.env]["sso_url"]
+        # Offline token used for generating API token
```
DanielOsypenko, line 42: for better understanding. Is it a private key, ".pem" file content?

ocs_ci/utility/assisted_installer.py
```diff
@@ -0,0 +1,517 @@
...
+            infra_env_id (str): Infra environment ID
+            host_id (str): Host ID
+            update_data (dict): the data to be updated
+        """
+        data = update_data
+        return self.patch_request(
+            f"infra-envs/{infra_env_id}/hosts/{host_id}", data=data
+        )
+
+    def get_discovery_iso_url(self, infra_env_id):
```
DanielOsypenko, line 499: for my understanding, can we use discovery to setup VM instead of boot iso?

DanielOsypenko (approved):

ocs_ci/utility/assisted_installer.py
```diff
@@ -0,0 +1,517 @@
...
+
+        Args:
+            endpoint (str): the final part of the endpoint URL
+
+        Returns:
+            str: full API endpoint URL
+        """
+        return urljoin(self.api_base_url, endpoint)
+
+    def get_request(self, endpoint, params=None, json=True, ignore_failure=False):
```
DanielOsypenko, line 114: I would suggest to add a retry decorator to all "_request" functions because of the bad connections and bad responses

ocs_ci/utility/assisted_installer.py
dahorak: No, this is _offline token_ from the OpenShift cluster Manager console used for accessing the API (in fact, it is used for generating another token, which is then used for accessing the API). See [the documentation](https://access.redhat.com/documentation/en-us/assisted_installer_for_openshift_container_platform/2023/html/assisted_installer_for_openshift_container_platform/installing-with-api#installing-the-openshift-cluster-manager-cli_installing-with-api).

ocs_ci/utility/assisted_installer.py
dahorak: Good point. I didn't faced any issue with accessing the API yet though, so not sure, if I should add the _retry_ right now or keep it without that, so we will know about potential issues and we could debug it if it is just network issue where the _retry_ makes sense or if it is another problem which we should fix differently. WDYT?

ocs_ci/utility/assisted_installer.py
dahorak: I'm not sure, what you meant?

petr-balogh (approved):

PR #8995, opened 2023-12-05
Disable test case: test_pv_creation

sidhant-agrawal:
This PR is to disable an outdated test case `test_pv_creation`.
Reason for Disabling:
 - No corresponding requirement or test case exists in Polarion.
 - Latest [OCP documentation](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.14/html-single/storage/index#reclaiming_understanding-persistent-storage) now classifies the used reclaim policy `Recycle` as deprecated. 

Also, discussed [here](https://url.corp.redhat.com/1acdac0) 


jilju (approved):

paraggit (approved):

keemano (approved):

nehaberry (approved):
LGTM

@sidhant-agrawal it would be good to add in the description of this PR the comment link /details copied from where this was discussed. So that we will have a have a history in the right place. WDYT?

PR #9072, opened 2023-12-14
Use COS as noobaa backing store for IBM Cloud IPI

clacroix12:
Issue: https://url.corp.redhat.com/1b7b3ca

### DIFF
--- a/ocs_ci/deployment/cloud.py
+++ b/ocs_ci/deployment/cloud.py
@@ -14,7 +14,6 @@ from ocs_ci.framework import config
 from ocs_ci.ocs import constants, exceptions
 from ocs_ci.utility.bootstrap import gather_bootstrap
 from ocs_ci.utility.deployment import get_cluster_prefix
-from ocs_ci.utility.ibmcloud import label_nodes_region
 from ocs_ci.utility.utils import get_cluster_name, run_cmd
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


vavuthu (approved):

petr-balogh (approved):

OdedViner (approved):

ocs_ci/deployment/cloud.py
```diff
@@ -119,3 +120,4 @@ def deploy(self, log_cli_level="DEBUG"):
                     logger.error(ex)
             raise e
         self.test_cluster()
+        label_nodes_region()
```
vavuthu, line 123: is this needed for other cloud platforms like aws, gcp ?

petr-balogh: This is IBM Cloud IPI specific because of we had issue to use COS as noobaa backing store.  @clacroix12 is this correct way to have it in the base IPI deployment class? Should we for now to only add this label for IBM Cloud IPI where we only see issue there?

Also the lable is only IBM Cloud specific, so it's not generic cloud specific.

petr-balogh (approved):

ocs_ci/deployment/cloud.py
clacroix12: Moved this function call to the `post_ocp_deploy` method in the base deployment class as well as added a platform/deployment_type check.

vavuthu (approved):

dahorak (approved):

PR #9077, opened 2023-12-15
Add new checks for provider/client setup

ebondare:
Some checks need to be disabled for client cluster, the way there were disabled for faas consumer. Noobaa checks needed to be disabled for provider/client deployment at this stage. I've added checks for StorageConsumers and adjusted check for StorageClient. Client cluster doesn't have ocs operator but ocs client operator, so this check needs adjustment as well.


jilju (approved):

amr1ta (dismissed):
Looking good, can you please trigger pr validation job.

dahorak (approved):

ocs_ci/helpers/managed_services.py
```diff
@@ -350,6 +352,21 @@ def verify_storageclient_storageclass_claims(storageclient):
...
 
+def verify_storageconsumers():
+    """
+    Verify that all storage consumers are in Ready state
+    """
+    sconsumer_ocp = OCP(
+        kind=constants.STORAGECONSUMER, namespace=config.ENV_DATA["cluster_namespace"]
+    )
+    sconsumers = sconsumer_ocp.get()["items"]
+    for sconsumer in sconsumers:
```
DanielOsypenko, line 363: one of the features of 4.16 is to change consumer names to client names on Client dashboard - https://issues.redhat.com/browse/RHSTOR-5500.
This test would probably fail if we don't change the names to their client names 
image

PR #9079, opened 2023-12-18
Update test_restart_mgr_while_two_mons_down for provider-client setup

jilju:
Update the test case test_restart_mgr_while_two_mons_down to run on provider-client setup


nehaberry (commented):
LGTM

However for checking the IO, it will switch back to Clients? Also, are we testing PVC create/delete too?

dahorak (approved):

PR #9111, opened 2024-01-04
[Provider/Client] deploy Hosted OCP Cluster 

DanielOsypenko:
**PR summary:**
Automation for the Hosted cluster deployment on IBM Bare Metal Hosting(Management) clusters.
**This PR includes:**
- MetalLB loadbalancer setup. Optionally reserve Ip addersses for vSphere nested virtualization deployments
- CNV installation and setup to create Hosted clusters
- hypershift and hcp cli tools setup on jenkins agent 
- helper utilities to operate hosted clusters with hypershift and hcp cli tools
- multiple simultaneous hosted cluster creation
- hosted clusters, nodes and pods verification from Hosting(Management) cluster
- ODF installation on Hosted clusters
- Storage provisioning on Hosted clusters (Storage Clients)

--------
vSphere is not supported officially, and Hosted cluster creation fails hence no successful testing is done


ocs_ci/deployment/metallb.py
```diff
@@ -0,0 +1,566 @@
...
+            ip_list_for_hosted_clusters = list()
+
+            # skip ip addresses reserved for machines, Network, Gateway, and Broadcast; first 10
+            for i, ip in enumerate(ip_list_by_cidr):
+                if i < 10 + 1 or i == len(ip_list_by_cidr) - 1:
+                    continue
+                ip_list_for_hosted_clusters.append(f"{ip}/32")
+            ipaddresspool_data.get("spec").update(
+                {"addresses": ip_list_for_hosted_clusters}
+            )
```
dahorak, line 313: I would propose to make the list of available IPs for Bare Metal deployment explicitly configurable via configuration file. We have to have anyway specific configuration file for each Baremetal cluster, so specifying the IPs for the ipaddresspool will be just small update there and it will be more generic for various environments.

ocs_ci/ocs/constants.py
```diff
@@ -1067,6 +1130,7 @@
 VSPHERE_PLATFORM = "vsphere"
 BAREMETAL_PLATFORM = "baremetal"
 IBM_POWER_PLATFORM = "powervs"
+IBM_CLOUD_BAREMETAL_PLATFORM = "ibm_cloud_baremetal"
```
dahorak, line 1133: I'm thinking if we have to separate the `baremetal` and `ibm_cloud_baremetal` platforms?

ocs_ci/ocs/constants.py
```diff
@@ -1154,7 +1218,7 @@
 SCALEUP_TERRAFORM_DATA_DIR = "scaleup_terraform_data"
 SCALEUP_VSPHERE_DIR = os.path.join(
     EXTERNAL_DIR,
-    f"v4-scaleup/ocp4-rhel-scaleup/aos-{get_ocp_version(seperator='_')}/vsphere",
+    # f"v4-scaleup/ocp4-rhel-scaleup/aos-{get_ocp_version(seperator='_')}/vsphere",
```
dahorak, line 1221: Is this change related to this PR?

ocs_ci/utility/utils.py
```diff
@@ -446,6 +446,8 @@ def mask_secrets(plaintext, secrets):
                 plaintext = [string.replace(secret, "*" * 5) for string in plaintext]
             else:
                 plaintext = plaintext.replace(secret, "*" * 5)
+    if secrets == ["*"]:
+        plaintext = "*" * 10
```
dahorak, line 450: What is the purpose of this change?

conf/ocsci/provider-client_ibm_cloud_baremetal.yaml
```diff
@@ -0,0 +1,12 @@
...
+DEPLOYMENT:
+  cnv_deployment: true
+  metallb_operator: true
+ENV_DATA:
+  deploy_acm_hub_cluster: true
+  skip_ocs_deployment: true
+  acm_version: "2.9"
+  acm_hub_channel: release-2.9
+  # we have 2 vlan public subnets, so we can use any of them
+  public_subnet_id: 1
```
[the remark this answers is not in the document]
DanielOsypenko: will remove, it is leftover, thanks

ocs_ci/deployment/cnv.py
```diff
@@ -233,47 +234,129 @@ def enable_software_emulation(self):
...
+        Args: allow_verification_failure: If True, allow the verification to fail. If False, raise an exception if the
+            verification fails.
 
         Raises:
-            TimeoutExpiredError: If the verification conditions are not met within the timeout.
-            HyperConvergedHealthException: If the HyperConverged cluster health is not healthy.
+            TimeoutExpiredError: If the verification conditions are not met within the timeout
+            and allow_verification_failure is False.
+            HyperConvergedHealthException: If the HyperConverged cluster health is not health
+            and allow_verification_failure is False.
```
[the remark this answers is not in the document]
DanielOsypenko: agree. adding it

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -0,0 +1,450 @@
...
+
+    def download_hosted_cluster_kubeconfig(self, name: str, auth_path: str = None):
+        """
+        Download HyperShift hosted cluster kubeconfig
+        :param name: name of the cluster
+        :param auth_path: path to download kubeconfig
+        :return: path to the downloaded kubeconfig, None if failed
+        """
+        if not auth_path:
+            auth_path = constants.AUTH_PATH_PATTERN.format(name)
```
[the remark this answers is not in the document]
DanielOsypenko: It is more complicated I think, so I wanted to avoid misunderstanding. 
For instance by the name in PATH we are picking the correct kubeconfig and performing "oc" commands to specific cluster. I wanted to make sure we have a standard in the path name. Other than that I don't see why user(qe) may want to have a custom path for the auth files.
We use "~/clusters/CLUSTER_NAME/openshift-cluster-dir/auth" as a standard here.

ocs_ci/deployment/metallb.py
DanielOsypenko: I thought I am following some convention. Because I am skipping first IPs that are usually for Network, Gateway, Broadcast and machine IPS. 
This will leave us a variable number of IPs that are left by cidr and we need to know only CIDR for all the job.
 
I did this by several reasons: 

- to not expose ipaddresses everywhere in docs, chats, deployment job 
- save User(qe) from the burden to find correct IPs 
- avoid potential issues when IPs are selected incorrectly.

ocs_ci/ocs/constants.py
DanielOsypenko: rename ibm_cloud_baremetal -> ibm_baremetal ?

ocs_ci/ocs/constants.py
DanielOsypenko: accidently left this. Thanks.

This should be removed by Vijay https://github.com/red-hat-storage/ocs-ci/issues/9432 it creates circular import when exporting any OCP related class to python console

ocs_ci/utility/utils.py
DanielOsypenko: I wanted to mask all the CR with IPs (IpAddressPool). I will probably remove this.

ocs_ci/deployment/cnv.py
```diff
@@ -233,47 +234,129 @@ def enable_software_emulation(self):
...
+                return False
+            else:
+                raise exceptions.ResourceNotFoundError(
+                    f"Namespace {self.namespace} does not exist"
+                )
+
+        try:
+            wait_for_nodes_status()
+            logger.info("All the nodes are in 'Ready' state")
+        except ResourceWrongStatusException:
```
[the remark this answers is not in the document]
DanielOsypenko: done

ocs_ci/deployment/cnv.py
```diff
@@ -233,47 +234,129 @@ def enable_software_emulation(self):
...
+        try:
+            health = (
+                ocp.get(resource_name=constants.KUBEVIRT_HYPERCONVERGED)
+                .get("status")
+                .get("systemHealthStatus")
             )
+            if health == "healthy":
+                logger.info("HyperConverged cluster health is healthy.")
+                return True
+            elif not health and allow_verification_failure:
```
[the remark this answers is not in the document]
DanielOsypenko: addressed

ocs_ci/deployment/cnv.py
```diff
@@ -233,47 +234,129 @@ def enable_software_emulation(self):
...
+                )
+            return False
+        except Exception as ef:
+            logger.error(
+                f"Failed to get the HyperConverged systemHealthStatus. Error: {ef}"
+            )
+            if allow_verification_failure:
+                raise exceptions.HyperConvergedHealthException(
+                    f"Failed to get the HyperConverged systemHealthStatus. Error: {ef}"
+                )
```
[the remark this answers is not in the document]
DanielOsypenko: addressed

ocs_ci/deployment/metallb.py
dahorak: > I thought I am following some convention. Because I am skipping first IPs that are usually for Network, Gateway, Broadcast and machine IPS. This will leave us a variable number of IPs that are left by cidr and we need to know only CIDR for all the job.

This is true for the two environment we have now, but it might not be for some other environment we will have in the future. For example the Primary IPs of the servers in IBM Cloud are assigned randomly from the subnet (not sequentially from the start.

> I did this by several reasons:
> 
>     * to not expose ipaddresses everywhere in docs, chats, deployment job

The "static" configuration for the particular environment will be part of the "credential" conf, so at least this part will be handled as other credentials. We will also have to make sure to mask it in the logs (if it is desirable).

>     * save User(qe) from the burden to find correct IPs

This will be one-time action for each BM environment - we have to prepare the configuration for all the other parameters for the cluster (configuration of each of the server and other network configurations), so this is just one more item in the existing config file for the particular environment.

>     * avoid potential issues when IPs are selected incorrectly.

That is the reason, why I would propose to set it statically once in configuration file.

ocs_ci/ocs/constants.py
dahorak: No I meant if we could use the existing `BAREMETAL_PLATFORM`? Or if there are some incompatible differences?

ocs_ci/deployment/helpers/hypershift_base.py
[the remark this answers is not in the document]
dahorak: The reason, why I would make it configurable is, that the `~/clusters/CLUSTER_NAME` convention is specific for our CI environment - jenkins jobs, but it might not be suitable for local execution and definitely it will not be suitable for other teams, if they would like to run our tests - so I would still vote for the option to make it configurable, while the default value could be the current path.

The reason why I previously suggested to place it to the _log_dir_ is, that this path is by default stored on the Shared storage on jenkins job, wile the `~/clusters/...` path has to be explicitly mounted for the particular agent (otherwise it will be stored only locally on the particular temp jenkins agent and will be lost whenever the jenkins agent will be destroyed or broken) - but we will anyway have to ensure that the cluster path will be stored on the Shared storage on proper location and mount the `~/clusters/...` directories properly, so the _log_dir_ is probably not good idea.

So to summarize it, I'm ok with the default value which you have implemented there, but I would propose to add option to change it from configuration file - if necessary.

ocs_ci/ocs/constants.py
DanielOsypenko: I think there maybe differences between baremetal and IBM baremetal

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -0,0 +1,450 @@
...
+    def get_hosted_cluster_progress(self, name):
+        """
+        Get HyperShift hosted cluster creation progress
+        :param name: name of the cluster
+        :return: progress status; 'Completed' is expected in most cases
+        """
+        cmd = f"oc get --namespace clusters hostedclusters | awk '$1==\"{name}\" {{print $4}}'"
+        return exec_cmd(cmd, shell=True).stdout.decode("utf-8").strip()
+
+    def get_ICSP_list(self):
```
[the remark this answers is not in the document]
DanielOsypenko: noted. For our tasks we are not configuring ICSP resource but only reading it and creating list with mirrors. I could not find if this resource will be removed from OCP cluster at 4.16 when depreciation will take an action. For now assuming that ICSP will be readable.

Will check on 4.16 deployment and update if needed

ocs_ci/deployment/metallb.py
DanielOsypenko: > This is true for the two environment we have now, but it might not be for some other environment we will have in the future.

The method create_ip_address_pool has if-else block to make an actions considering the platform. For now they are vSphere and IBM. Further we will prepare more and update the function if there will be deviation on IP's per setup protocol. 

Regarding to have a static IP list approach that designed to be universal for all the platforms. I am not sure we will have static IPs and no IP reservation functionality for all the new platforms. So in any case we most likely would need to alter the function `create_ip_address_pool()` for AWS or BareMetal with agent, or other new platforms.

---

If we still want to have files with all static IPs, can you pls prepare the configs, so I can see format of the static IP list on GitLab, if I understood you correctly. I will fetch and work with it within existing PR

ocs_ci/deployment/cnv.py
```diff
@@ -492,11 +575,26 @@ def _extract_virtctl_binary(
...
+            logger.info("CNV operator is already deployed, skipping the deployment")
+            return
+
+        if (
+            check_cnv_ready
+            and self.cnv_hyperconverged_installed()
+            and self.post_install_verification(allow_verification_failure=True)
+        ):
+            logger.info("CNV operator ready, skipping the deployment")
+            return
```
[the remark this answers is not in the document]
dahorak: I meant to call the `cnv_hyperconverged_installed` only once:

```python
        if self.cnv_hyperconverged_installed():
            if check_cnv_deployed:
                logger.info("CNV operator is already deployed, skipping the deployment")
                return

            if check_cnv_ready and self.post_install_verification(
                raise_exception=False
            ):
                logger.info("CNV operator ready, skipping the deployment")
                return
```

ocs_ci/deployment/metallb.py
dahorak: My note here was about the IBM BM platform, on vSphere the approach of reserving the IPs in IPAM server is correct.

About the format of the configuration file, it is basically up to you - what you will prefer/what will be easily re-usable in the code. For example something like this?
```yaml
ENV_DATA:
    metallb_ip_address_pool:
        - "192.0.2.12/32"
        - "192.0.2.13/32"
        - "192.0.2.16/28"
        ...
```

ocs_ci/ocs/constants.py
dahorak: The question will be, how big the differences will be? For example in my PR https://github.com/red-hat-storage/ocs-ci/pull/9451, I'm handling it as generic baremetal platform and the difference between the BM env in Octo lab  and the IBM Cloud BM is in the [`mgmt_provider`](https://github.com/red-hat-storage/ocs-ci/pull/9451/files#diff-05053122d6e5936ca528fcd1643379f792a8350f094e1c0aa5c62903f86a8746R1184), which controls if the BM machine is managed via `ipmitool` or `ibmcloud` command.

I don't know all the details there about the differences, so maybe it make sense to separate it like this - but if the differences are not too big, I would vote to keep it unified here.

DanielOsypenko: for node operations the differences are serious @yitzhak12 can confirm

ocs_ci/deployment/metallb.py
DanielOsypenko: I will update PR, could you please create this config files on gitlab repo. If I remember correctly I don't have permissions for MR.
The IPs are (if not changed):

baremetal1
subnet: 52.118.6.128/27
52.118.6.129  Gateway
52.118.6.130  provisioning VM
52.118.6.131  baremetal01
52.118.6.132  baremetal02
52.118.6.133  baremetal03
52.118.6.134  baremetal04
52.118.6.135  baremetal05
52.118.6.136  baremetal06
52.118.6.137  API
52.118.6.138  Ingress
52.118.6.139  - 52.118.6.158  metalLB ipaddress pool

baremetal2
subnet: 169.63.205.128/27
169.63.205.129  Gateway
169.63.205.130  provisioning VM
169.63.205.131  baremetal2-01
169.63.205.132  baremetal2-02
169.63.205.133  baremetal2-03
169.63.205.134  baremetal2-04
169.63.205.135  baremetal2-05
169.63.205.136  baremetal2-06
169.63.205.137  API
169.63.205.138  Ingress
169.63.205.139 - 169.63.205.158  metalLB ipaddress pool for baremetal2 cluster

dahorak: As discussed, I've added following configuration to the credentials file for each BM cluster:
```
ENV_DATA:
    ip_address_pool:
        - "..."
        - "..."
        ...
```

DanielOsypenko:
latest test-run:
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/35372/

ocs_ci/deployment/helpers/hypershift_base.py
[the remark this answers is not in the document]
DanielOsypenko: I've added `ENV_DATA.cluster_paths` dependency to store credentials and maybe other docs related to specific cluster. 
The cluster_paths should have the name of the cluster within the path, for example `~/clusters/CLUSTER_NAME/auth` and name should be one of the desired names from variable ENV_DATA.cluster_names

This is one of the cases when I see it will be correct to abort the deployment so I've added this block into HostedClients constructor:

` def __init__(self):
        HyperShiftBase.__init__(self)
        if not ("cluster_names" in config.default_cluster_ctx.ENV_DATA):
            raise ValueError("No 'cluster_names' set to ENV_DATA")
        if not ("cluster_paths" in config.default_cluster_ctx.ENV_DATA):
            raise ValueError("No 'cluster_paths' set to ENV_DATA")
        if not len(config.default_cluster_ctx.ENV_DATA.get("cluster_names")) != len(
            "cluster_paths"
        ):
            raise ValueError(
                "The number of 'ENV_DATA.cluster_names' and 'ENV_DATA.cluster_paths' should be the same"
            )`

ocs_ci/deployment/metallb.py
DanielOsypenko: added to the PR, all the logic to find IPs by using the cidr was removed.

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -0,0 +1,1027 @@
...
+            logger.info(
+                f"Setting up Storage client on hosted OCP cluster '{cluster_name}'"
+            )
+            hosted_odf = HostedODF(cluster_name)
+            client_setup.append(hosted_odf.setup_storage_client())
+
+        # stage 7 verify all hosted clusters are ready and print kubeconfig paths
+        logger.info("kubeconfig files for all hosted OCP clusters:\n")
+        for kubeconfig_path in kubeconfig_paths:
+            logger.info(f"kubeconfig path: {kubeconfig_path}\n")
```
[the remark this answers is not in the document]
DanielOsypenko: ack, removing

conf/ocsci/provider_mode_ibm_cloud_baremetal_kubevirt_clients.yaml
```diff
@@ -0,0 +1,20 @@
...
+  acm_hub_channel: "release-2.9" # this is an example, please provide the desired version
+  hcp_version: "4.14" # this is an example, please provide the desired version
+  hosted_odf_version: "latest-stable-4.14.5" # this is an example, please provide the desired version
+  metallb_version: "4.14" # this is an example, please provide the desired version
+  cpu_cores_per_hosted_cluster: 6 # minimal, tested value for the kubevirt cluster is 6
+  memory_per_hosted_cluster: "12Gi" # minimal, tested value for the kubevirt cluster is 12Gi
+  cluster_names:
+    - "hcp414-bm2-h" # this is an example, please provide the desired cluster_names
+  cluster_paths:
+    - "~/clusters/hcp414-bm2-h/openshift-cluster-dir/auth" # this is an example, provide the desired cluster_paths
```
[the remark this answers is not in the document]
DanielOsypenko: It is possible. I will redo this.

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -0,0 +1,1027 @@
...
+
+        cluster_names = get_hosted_cluster_names()
+        cluster_paths = config.default_cluster_ctx.ENV_DATA.get("cluster_paths")
+
+        name_to_path = {
+            name: path
+            for name in cluster_names
+            for path in cluster_paths
+            if name in path
+        }
```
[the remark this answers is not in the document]
DanielOsypenko: agree, I am going to redo this.

ocs_ci/deployment/metallb.py
```diff
@@ -0,0 +1,623 @@
...
+            if "ip_address_pool" not in config.ENV_DATA:
+                raise ValueError("IP address pool is not specified in the config file")
+
+            ipaddresspool_data.get("spec").update(
+                {"addresses": config.ENV_DATA["ip_address_pool"]}
+            )
+        else:
+            raise NotImplementedError(
+                f"Platform {config.ENV_DATA['platform']} is not supported yet"
+            )
```
dahorak, line 351: I think that if provided, we can use the values from `config.ENV_DATA["ip_address_pool"]` for any platform (not only the `HCI_BAREMETAL`).
Maybe even the values from this configuration could have precedence before the logic of reserving IPs from IPAM for the vSphere platform. In other words, if the `config.ENV_DATA["ip_address_pool"]` configuration will be provided, it will use the values from the configuration and if not, it will try to use the logic specific for the platform - e.g. reserve the IPs in IPAM server - and if no such logic is available, than raise the exception that the ip address pool is not provided and automatic assignment is not implemented for that platform. With this approach, it will be quite flexible for implementing new platforms and also for debugging purposes and similar cases. WDYT?

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -0,0 +1,1027 @@
...
+
+    def __init__(self):
+        HyperShiftBase.__init__(self)
+        if not ("cluster_names" in config.default_cluster_ctx.ENV_DATA):
+            raise ValueError("No 'cluster_names' set to ENV_DATA")
+        if not ("cluster_paths" in config.default_cluster_ctx.ENV_DATA):
+            raise ValueError("No 'cluster_paths' set to ENV_DATA")
+        if not len(config.default_cluster_ctx.ENV_DATA.get("cluster_names")) != len(
+            "cluster_paths"
+        ):
```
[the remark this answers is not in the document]
DanielOsypenko: I am dropping this block

conf/ocsci/provider_mode_ibm_cloud_baremetal_kubevirt_clients.yaml
```diff
@@ -0,0 +1,20 @@
...
+DEPLOYMENT:
+  cnv_deployment: true
+  metallb_operator: true
+ENV_DATA:
+  platform: "hci_baremetal"
+  cluster_type: "provider"
+  deploy_acm_hub_cluster: true
+  skip_ocs_deployment: true
+  acm_version: "2.9" # this is an example, please provide the desired version
+  acm_hub_channel: "release-2.9" # this is an example, please provide the desired version
```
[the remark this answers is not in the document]
DanielOsypenko: this example yaml has a full list of what we need in order to deploy ProviderMode, at least it was designed to have all necessary parts. We would like to have ACM version unbounded from OCP version. 

This is similar to "cnv_deployment: true" we may remove it but I'd prefer to show it to user.
I understand that we have documentation, but I'd rather add extra field than do force user to gather information from different resources.

conf/ocsci/provider_mode_ibm_cloud_baremetal_kubevirt_clients.yaml
```diff
@@ -0,0 +1,20 @@
...
+  hcp_version: "4.14" # this is an example, please provide the desired version
+  hosted_odf_version: "latest-stable-4.14.5" # this is an example, please provide the desired version
+  metallb_version: "4.14" # this is an example, please provide the desired version
```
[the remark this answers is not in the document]
DanielOsypenko: we have this file - conf/ocsci/metallb.yaml
same as with acm_version I would prefer to have everything needed for Provider mode deployment in one place, to show user all necessary dependencies.

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -0,0 +1,1027 @@
...
+        CNVInstaller.__init__(self)
+        self.name = name
+        if "cluster_paths" in config.ENV_DATA:
+            cluster_paths = config.ENV_DATA["cluster_paths"]
+            for path in cluster_paths:
+                if self.name in path:
+                    self.cluster_kubeconfig = os.path.expanduser(
+                        os.path.join(path, "kubeconfig")
+                    )
+                    break
```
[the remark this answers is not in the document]
DanielOsypenko: done, removed

conf/ocsci/provider_mode_ibm_cloud_baremetal_kubevirt_clients.yaml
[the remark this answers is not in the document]
DanielOsypenko: done. 
I'll tell right away that I think that it would be the best to store all related data for hosted `cluster` to the dict you proposed. 

For example: 
clusters:
  "cluster1": 
    "platform": "kubevirt"
    "cluster_path": "~/clusters/cluster1/openshift-cluster-dir"
    "cpu_cores": 6
    "memory": "12Gib"
    "odf_version": "4.16-latest"
  "cluster2": 
    "platform": "agent"
    "cluster_path": "~/clusters/cluster2/openshift-cluster-dir"
    "cpu_cores": 8
    "memory": "16Gib"
    "odf_version": "4.16.0-56"

In future it will return dividends, since we will be able to install hosted clusters with different multiple platforms, different ODF versions and different specs per cluster during one jenkins job.

I will leave this as an enhancement - https://github.com/red-hat-storage/ocs-ci/issues/9594

ocs_ci/helpers/helpers.py
```diff
@@ -4576,6 +4564,29 @@ def retrieve_cli_binary(cli_type="mcg"):
     ), f"The {cli_type} CLI binary does not have execution permissions"
 
 
+def download_pull_secret():
```
fbalak, line 4567: This can be added to ocs_ci/deployment/ocp.py. There is already a function for getting content of the pull secret file (https://github.com/red-hat-storage/ocs-ci/blob/503f071e820544889b09b9044ef507c74a2676f2/ocs_ci/deployment/ocp.py#L64)

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -0,0 +1,1127 @@
...
+                    raise AssertionError(
+                        f"Failed to install Hypershift on the cluster: {e}"
+                    )
+
+
+class HostedODF(HypershiftHostedOCP):
+    def __init__(self, name: str):
+        HyperShiftBase.__init__(self)
+        HypershiftHostedOCP.__init__(self, name)
+        self.namespace_client = constants.OPENSHIFT_STORAGE_CLIENT_NAMESPACE
```
fbalak, line 376: This value should be taken from config in case that client is installed in different namespace. It can be added to default config so that it is not defined again in other configs that need a different value. (e.g. add `client_namespace` to https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L149)

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -0,0 +1,1127 @@
...
+                os.path.join(cluster_path, "kubeconfig")
+            )
+        else:
+            # avoid throwing an exception if the cluster path is not found for some reason
+            # this way we can continue with the next cluster
+            logger.error(
+                f"Cluster path for desired cluster with name '{self.name}' was not found in ENV_DATA.clusters"
+            )
+
+    def kubeconfig_exists(self):
```
fbalak, line 283: This function is generic and can be helpful to others. I suggest to add it to parent class `Deployment`.

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -0,0 +1,1127 @@
...
+            bool: True if storage client exists, False otherwise
+        """
+        ocp = OCP(
+            kind=constants.STORAGECLIENTS,
+            namespace=self.namespace_client,
+            cluster_kubeconfig=self.cluster_kubeconfig,
+        )
+        return ocp.check_resource_existence(
+            timeout=self.timeout_check_resources_exist,
+            resource_name=constants.STORAGE_CLIENT_NAME,
```
fbalak, line 609: This should be taken from config. There is already a value for it in default config (https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L145).

ocs_ci/utility/utils.py
```diff
@@ -4739,3 +4739,21 @@ def exec_nb_db_query(query):
         output = output[2:-1]
 
     return output
+
+
+def get_latest_release_version():
```
fbalak, line 4744: This function is in generic utils module. Please add information about what latest release version is obtained by the function into function name and description.

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -0,0 +1,631 @@
...
+
+    def hcp_binary_exists(self):
+        """
+        Check if hcp binary exists
+        Returns:
+            bool: True if hcp binary exists, False otherwise
+        """
+        return os.path.isfile(self.hcp_binary_path)
+
+    def install_hcp_and_hypershift_from_git(self):
```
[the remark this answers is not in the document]
DanielOsypenko: 2 binaries are unpacked when one command is run `make hypershift product-cli` So each time when we need to have `hci` we download and unpack both of them. The only difference can be: do we want move unpacked 'hypershift' to bin directory or not. 

Considering 1 Jenkins agent may have only one OCP Managing cluster of one particular version I think that it does not worth separation on two utilities. There should be no situation when we would need `hypershift` of version 4.15 and `hcp` of version 4.16 on one agent

ocs_ci/deployment/helpers/icsp_parser.py
```diff
@@ -0,0 +1,119 @@
...
+Main task is to parse the ImageContentSourcePolicy cr in format, suitable for
+passing 'mirrors to sources' file to 'hcp' or 'hypershift' binaries while creating
+hosted cluster.
+"""
+
+logger = logging.getLogger(__name__)
+
+
+class ImageContentSourcePolicy:
+    def __init__(self, apiVersion=None, items=None, kind=None, metadata=None, **kwargs):
```
fbalak, line 14: Please add docstring to classes in this module.

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -0,0 +1,631 @@
...
+            f"git clone --single-branch --branch release-{hcp_version} "
+            f"--depth 1 {constants.HCP_REPOSITORY} {temp_dir}"
+        )
+
+        exec_cmd(f"cd {temp_dir} && make hypershift product-cli", shell=True)
+
+        exec_cmd(f"mv {temp_dir}/bin/hypershift {self.hypershift_binary_path}")
+        exec_cmd(f"mv {temp_dir}/bin/hcp {self.hcp_binary_path}")
+        hcp_version = exec_cmd("hcp --version").stdout.decode("utf-8").strip()
+        logger.info(f"hcp binary version: {hcp_version}")
```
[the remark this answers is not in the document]
DanielOsypenko: this is a critical point, if we fail here the job error is acceptable. Do you suggest to add a check on hcp and hypershift files have been moved into bin dir? 

Function has been tested multiple times and never give us failures

conf/ocsci/metallb.yaml
```diff
@@ -0,0 +1,3 @@
+---
+ENV_DATA:
+  metallb_version: "latest"
```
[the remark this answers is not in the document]
DanielOsypenko: you're right, this yaml is not independent. I will need to remove it.

ocs_ci/deployment/helpers/hypershift_base.py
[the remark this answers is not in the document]
dahorak: I think that my point was, that if the download or unpacking will be unsuccessful, it will fail here (or return invalid response), so it will be more clear to firstly do the check if it was correctly downloaded and unpacked and than check/print the version... But I agree that it is more like cosmetic thing.

conf/README.md
```diff
@@ -300,6 +301,17 @@ higher priority).
...
+  * `<cluster name>` - name of the cluster
+    * `<cluster path>` - path to the cluster directory to store credentials files, such as kubeconfig
+* `cpu_cores_per_hosted_cluster` - number of CPU cores per hosted cluster
+* `memory_per_hosted_cluster` - amount of memory per hosted cluster
+* `nodepool_replicas` - number of replicas of nodepool for each cluster
+* `hosted_odf_registry` - registry for hosted ODF
+* `hosted_odf_version` - version of ODF to be deployed on hosted clusters
+* `hcp_version` - version of HCP client to be deployed on machine running the tests
+* `metallb_version` - MetalLB operator version to install
+* `install_hypershift_upstream` - Install hypershift from upstream or not (Default: false). Necessary for unreleased OCP/CNV versions
```
amr1ta, line 315: Very helpful document

DanielOsypenko: thanks!

conf/ocsci/provider_mode_ibm_cloud_baremetal_kubevirt_clients.yaml
```diff
@@ -0,0 +1,25 @@
+# This yaml file contains the configuration for the deployment of the ACM hub cluster on IBM Cloud Bare Metal
+# with Hosted client clusters. The configuration is used by the HyperShiftBase.download_hcp_binary(),
+# HostedClients.deploy_hosted_ocp_clusters(), HypershiftHostedOCP.deploy_ocp(), HostedODF.do_deploy(),
+# MetallbOperator.deploy_metallb() and similar methods during the Hosted clusters deployment.
+# The configuration is for observation purposes only and is not intended to be executed. In order to execute
+# the configuration, the user must provide the desired values for the variables in the configuration.
```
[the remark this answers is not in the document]
DanielOsypenko: moved to `conf/examples`

ocs_ci/deployment/cnv.py
```diff
@@ -217,9 +226,10 @@ def enable_software_emulation(self):
         useEmulation.
 
         """
-        if config.ENV_DATA["platform"].lower() == "baremetal" and config.DEPLOYMENT.get(
-            "local_storage"
-        ):
+        if config.ENV_DATA["platform"].lower() in [
```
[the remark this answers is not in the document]
DanielOsypenko: addressed. 

```
        if config.ENV_DATA[
            "platform"
        ].lower() in constants.BAREMETAL_PLATFORMS and config.DEPLOYMENT.get(
            "local_storage"
        ):
            logger.info("Skipping enabling software emulation")
        else:
             ...
```

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -0,0 +1,631 @@
...
+
+    def install_hcp_and_hypershift_from_git(self):
+        """
+        Install hcp binary from git
+        """
+        if self.hcp_binary_exists():
+            logger.info(
+                f"hcp binary already exists {self.hcp_binary_path}, skipping download."
+            )
+            return
```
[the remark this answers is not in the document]
DanielOsypenko: corrected, now both binaries are checked and downloaded if one does not exist

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -0,0 +1,631 @@
...
+
+        exec_cmd(
+            f"git clone --single-branch --branch release-{hcp_version} "
+            f"--depth 1 {constants.HCP_REPOSITORY} {temp_dir}"
+        )
+
+        exec_cmd(f"cd {temp_dir} && make hypershift product-cli", shell=True)
+
+        exec_cmd(f"mv {temp_dir}/bin/hypershift {self.hypershift_binary_path}")
+        exec_cmd(f"mv {temp_dir}/bin/hcp {self.hcp_binary_path}")
```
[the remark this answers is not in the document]
DanielOsypenko: done

ocs_ci/deployment/helpers/hypershift_base.py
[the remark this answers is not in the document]
DanielOsypenko: addressed:

      ```
  if not (self.hcp_binary_exists() and self.hypershift_binary_exists()):
            raise Exception("Failed to download hcp binary from git")
```

File: ocs_ci/deployment/helpers/hypershift_base.py [hunk 35]
```diff
@@ -0,0 +1,631 @@
...
+        path_abs = os.path.expanduser(auth_path)
+        kubeconfig_path = os.path.join(path_abs, "kubeconfig")
+
+        os.makedirs(path_abs, exist_ok=True)
+
+        if os.path.isfile(kubeconfig_path):
+            logger.info(
+                f"Kubeconfig file for HyperShift hosted cluster {name} already exists at {path_abs}, removing it"
+            )
+            exec_cmd(f"rm -f {kubeconfig_path}")
```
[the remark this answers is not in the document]
DanielOsypenko: addressed

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -0,0 +1,450 @@
...
+
+    def get_ICSP_list(self):
+        """
+        Get list of ICSP clusters
+
+        """
+        if (
+            not os.path.getsize(self.icsp_mirrors_path)
+            and not os.path.getsize(self.icsp_mirrors_path) == 0
+        ):
```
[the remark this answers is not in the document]
DanielOsypenko: correct, thank you. 
now:
```
        if os.path.getsize(self.icsp_mirrors_path) > 0:
            logger.info(
                f"ICSP mirrors list already exists at '{self.icsp_mirrors_path}'"
            )
            return

```

ocs_ci/deployment/helpers/icsp_parser.py
DanielOsypenko: + added

ocs_ci/deployment/hosted_cluster.py
DanielOsypenko: the function does not have value if `self.cluster_kubeconfig` is not an instance attribute, which is exactly the case if we move it to Deployment.
Move attribute to method argument does not make sense either, in such case let's just use `os.path.exists`

I would leave it as is for now.

ocs_ci/deployment/hosted_cluster.py
DanielOsypenko: thanks, corrected

ocs_ci/deployment/hosted_cluster.py
DanielOsypenko: corrected, thanks

ocs_ci/deployment/metallb.py
```diff
@@ -0,0 +1,660 @@
...
+                    )
+
+            # due to circular import error, import is here
+            from ocs_ci.deployment.vmware import assign_ips
+
+            logger.info("Reserving IP addresses from IPAM and Creating IP address pool")
+            self.addresses_reserved = assign_ips(hosts=self.hostnames)
+
+            ip_addresses_with_mask = [ip + "/32" for ip in self.addresses_reserved]
+            ipaddresspool_data.get("spec").update({"addresses": ip_addresses_with_mask})
```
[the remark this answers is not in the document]
DanielOsypenko: added. thanks

ocs_ci/helpers/helpers.py
DanielOsypenko: due to the hierarchy, I mostly use it from other functions or helper objects like HypershiftBase I moved it to a file ocs-ci/ocs_ci/deployment/ocp.py but left it be a function and not a method of OCPDeployment

ocs_ci/utility/utils.py
DanielOsypenko: added, thanks

conf/examples/provider_mode_ibm_cloud_baremetal_kubevirt_clients.yaml
```diff
@@ -0,0 +1,36 @@
+# This yaml file contains the configuration for the deployment of the ACM hub cluster on IBM Cloud Bare Metal
+# with Hosted client clusters. The configuration is used by the HyperShiftBase.download_hcp_binary(),
+# HostedClients.deploy_hosted_ocp_clusters(), HypershiftHostedOCP.deploy_ocp(), HostedODF.do_deploy(),
+# MetallbOperator.deploy_metallb() and similar methods during the Hosted clusters deployment.
+# The configuration is for observation purposes only and is not intended to be executed. In order to execute
+# the configuration, the user must provide the desired values for the variables in the configuration.
+
```
dahorak, line 7: What about renaming this file and adding suffix `.example` (similarly to `conf/ocsci/vsphere_upi_vars.yaml.example`)?

dahorak (approved):
I tried to do a final review and overall it LGTM.

Did you tried to run verification OCP deployment also for vSphere IPI configuration against this PR (because of the changes in `ocs_ci/deployment/vmware.py`)?

DanielOsypenko:
> I tried to do a final review and overall it LGTM.
> 
> Did you tried to run verification OCP deployment also for vSphere IPI configuration against this PR (because of the changes in `ocs_ci/deployment/vmware.py`)?

no, I did not run deployment on vSphere since it was dropped from the scope. CNV is not stable on nested virtualization and most likely fail to create kubevirt cluster.
I did not remove related code because it may be supported in the future.

fbalak (approved):

dahorak (approved):

suchita-g (approved):

ebondare (dismissed):

DanielOsypenko:
>I meant normal OCP IPI deployment on vSphere, not CNV - just to verify that it was not affected by the changes in ocs_ci/deployment/vmware.py
I see, I will run the deployment job on vSphere IPI before req to merge

ebondare (approved):

suchita-g (dismissed):

ebondare (dismissed):

DanielOsypenko:
test run, deployment - Passed 
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/37850/console

fbalak (dismissed):

dahorak (dismissed):

jilju (approved):

suchita-g (approved):

ebondare (approved):

dahorak (approved):

ebondare (approved):

fbalak (approved):

PR #9124, opened 2024-01-08
Create a new file for the hci provider-client restart node tests

yitzhak12:
The PR is for the basic hci provider-client node restart test scenarios.
I haven't added IO operations yet because I am still unsure if we can use the regular sanity helper method or create a new sanity helper method as we did in the Managed Service.


yitzhak12:
We already implemented this test file in the PR https://github.com/red-hat-storage/ocs-ci/pull/9202

PR #9140, opened 2024-01-09
Adding encrypted PVC expansion test automation when PVC is 100% full

paraggit:
Test Steps:
            1. Create a encrypted storageclass.
            2. Create a encrypted PVC of size 4G using defined vault configuration.
            3. create a pod and mount the PVC inside the pod.
            4. Fill FS 100% using FIO workload.
            5. Expand PVC to size 10G
            6. continue IO workload and verify the IOs are succeeding
            7. Verify utilization alert will stop firing after volume expansion.

### DIFF
--- a/tests/functional/pv/pv_encryption/test_encrypted_pvc_expansion_when_full.py
+++ b/tests/functional/pv/pv_encryption/test_encrypted_pvc_expansion_when_full.py
@@ -63,7 +63,7 @@ else:
 @skipif_hci_provider_and_client
 @skipif_disconnected_cluster
 @skipif_proxy_cluster
-class TestPvcExpansionWhenFull(ManageTest):
+class TestEncryptedPvcExpansionWhenFull(ManageTest):
     """
     Tests to verify encrypted PVC expansion when the PVC is 100% utilized.
     Verify utilization alert will stop firing after volume expansion.
@@ -114,7 +114,7 @@ class TestPvcExpansionWhenFull(ManageTest):
             project=self.proj_obj,
         )
 
-    def test_pvc_expansion_when_full(self, threading_lock):
+    def test_encrypted_pvc_expansion_when_full(self, threading_lock):
         """
         Verify PVC expansion when the encrypted PVC is 100% utilized.
 
--- a/tests/functional/pv/pv_encryption/test_encrypted_pvc_expansion_when_full.py
+++ b/tests/functional/pv/pv_encryption/test_encrypted_pvc_expansion_when_full.py
@@ -1,0 +1,372 @@
+import logging
+import pytest
+
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.exceptions import CommandFailed
+from ocs_ci.ocs.resources.pod import get_used_space_on_mount_point
+from ocs_ci.framework.pytest_customization.marks import green_squad
+from ocs_ci.framework.testlib import (
+    skipif_ocs_version,
+    ManageTest,
+    tier2,
+    kms_config_required,
+    skipif_managed_service,
+    skipif_hci_provider_and_client,
+    skipif_disconnected_cluster,
+    skipif_proxy_cluster,
+    config,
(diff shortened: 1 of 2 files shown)


tests/functional/pv/pv_encryption/test_encrypted_pvc_expansion_when_full.py
```diff
@@ -0,0 +1,372 @@
...
+        kv_version,
+        kms_provider,
+        pv_encryption_kms_setup_factory,
+        use_vault_namespace,
+        project_factory,
+        storageclass_factory,
+        create_pvcs_and_pods,
+    ):
+        """
+        Setup csi-kms-connection-details configmap
```
ebenahar, line 85: this fixture does more than that

tests/functional/pv/pv_encryption/test_encrypted_pvc_expansion_when_full.py
```diff
@@ -0,0 +1,372 @@
...
+                jobs=8,
+                size=f"{self.pvc_size}G",
+                io_direction="write",
+                fio_filename=f"{pod_obj.name}_f1",
+            )
+        log.info("Started IO on all to utilise 100% of PVCs")
+        # Wait for IO to finish
+        log.info("Wait for IO to finish on pods")
+        for pod_obj in self.pods:
+            # for pod_obj in self.pod_objs:
```
ebenahar, line 149: should be deleted I guess

tests/functional/pv/pv_encryption/test_encrypted_pvc_expansion_when_full.py
```diff
@@ -0,0 +1,372 @@
...
+            try:
+                pod_obj.get_fio_results()
+            except CommandFailed as cfe:
+                if "No space left on device" not in str(cfe):
+                    raise
```
ebenahar, line 154: isn't it expected that the device will run out of space if we fill it to 100% capacity? If so, why is the exception being raised and not caught?

tests/functional/pv/pv_encryption/test_encrypted_pvc_expansion_when_full.py
```diff
@@ -0,0 +1,372 @@
...
+                    alert
+                    for alert in alerts
+                    if alert.get("labels", {}).get("persistentvolumeclaim")
+                    == pvc_obj.name
+                ]
+                # At least 2 alerts should be present
+                if len(alerts_pvc) < 2:
+                    break
+
+                # Verify 'PersistentVolumeUsageNearFull' alert is firing
```
shyRozen, line 180: Is there a reason to check both PersistentVolumeUsageNearFull (75%) and PersistentVolumeUsageCritical (85%). Just trying to understand why use both as we try to exhaust the capacity, maybe checking 85% alert is enough?

tests/functional/pv/pv_encryption/test_encrypted_pvc_expansion_when_full.py
```diff
@@ -0,0 +1,372 @@
...
+            5. Expand PVC to size 10G
+            6. continue IO workload and verify the IOs are succeeding
+            7. Verify utilization alert will stop firing after volume expansion.
+
+        """
+        pvc_size_expanded = 10
+
+        # Run IO to utilise 100% of volume
+        log.info("Run IO on all to utilise 100% of PVCs")
+        for pod_obj in self.pods:
```
shyRozen, line 135: If you'd like you can make them run parallel and save some time. Here is an example [https://github.com/paraggit/ocs-ci/blob/377ff1d814ba702eb1d870570f1ada12198cf654/tests/lvmo/test_lvm_multi_clone.py#L136-L162](https://github.com/paraggit/ocs-ci/blob/377ff1d814ba702eb1d870570f1ada12198cf654/tests/lvmo/test_lvm_multi_clone.py#L136-L162)

Issue #9155, opened 2024-01-15
Upgrade test is failing with verification of pods: app=rook-ceph-exporter


petr-balogh:
Related PR which introduced this check:

https://github.com/red-hat-storage/ocs-ci/pull/9109/files

Need to check if it's intentional behavior that on upgraded cluster we don't have those `rook-ceph-exporter` pods.

Failed job:
 https://url.corp.redhat.com/56bee03

petr-balogh:
@vavuthu  FYI

petr-balogh:
Checking if it's valid or not here:
https://chat.google.com/room/AAAAREGEba8/Hw_uSLmkrkE

PR #9156, opened 2024-01-15
Disable ceph exporter pods in case of upgrade from 4.14

petr-balogh:
Fixes: #9155


petr-balogh:
Verification job:
https://url.corp.redhat.com/1fd00dc

petr-balogh:
It's fixed in another PR by Vijay

PR #8873, opened 2023-11-17
CNV deployment

PrasadDesala:
OpenShift Virtualization (CNV) is an add-on to OCP that allows running and managing virtual machine workloads alongside container workloads. The ODF can be used in conjunction with CNV to provide persistent storage for virtual machines. 
This PR automates the entire CNV deployment process, including HyperConverged CR creation, after the ODF storage cluster creation. 

It also covers the following tasks after CNV deployment:
- [x] Post CNV installation verification
- [x] Enable software emulation - This is necessary on a cluster where the nodes do not support hardware emulation.
- [x]  Download and extract the virtctl binary to bin_dir - This additional binary provided by KubeVirt, allows quick access to the serial and graphical ports of a VM and also handles start/stop operations and various other VM operations.

PR #9192, opened 2024-01-19
install CNV only after ODF installation

vavuthu:

### DIFF
--- a/ocs_ci/deployment/cnv.py
+++ b/ocs_ci/deployment/cnv.py
@@ -497,18 +497,19 @@ class CNVInstaller(object):
         Installs CNV enabling software emulation.
 
         """
-        logger.info("Installing CNV")
-        # Create CNV catalog source
-        self.create_cnv_catalog_source()
-        # Create openshift-cnv namespace
-        self.create_cnv_namespace()
-        # create CNV subscription
-        self.create_cnv_subscription()
-        # Deploy the HyperConverged CR
-        self.deploy_hyper_converged()
-        # Post CNV installation checks
-        self.post_install_verification()
-        # Enable software emulation
-        self.enable_software_emulation()
-        # Download and extract the virtctl binary to bin_dir
-        self.download_and_extract_virtctl_binary()
+        if not config.ENV_DATA["skip_ocs_deployment"]:
+            logger.info("Installing CNV")
+            # Create CNV catalog source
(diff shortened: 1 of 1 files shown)


petr-balogh (dismissed):

ocs_ci/deployment/cnv.py
```diff
@@ -497,18 +497,19 @@ def deploy_cnv(self):
...
-        self.create_cnv_subscription()
-        # Deploy the HyperConverged CR
-        self.deploy_hyper_converged()
-        # Post CNV installation checks
-        self.post_install_verification()
-        # Enable software emulation
-        self.enable_software_emulation()
-        # Download and extract the virtctl binary to bin_dir
-        self.download_and_extract_virtctl_binary()
+        if not config.ENV_DATA["skip_ocs_deployment"]:
```
petr-balogh, line 500: Should we maybe just conditionalize the place where we call this function? 
This function is meant to install CNV, and it's not describing anywhere that it will not install the CNV in some of cases like now `if not config.ENV_DATA["skip_ocs_deployment"]:`

PrasadDesala: I'm not sure if this PR is relevant.  To provide a quick background - initially, I added the skip_ocs_deployment check in the PR for CNV deployment to ensure that the CNV deployment function is called only when OCS is deployed. However, it seems the CNV deployment is not being triggered during an MDR deployment job (could be that it was developed differently to handle multiple clusters). I later removed the check to allow CNV deployment on MDR clusters. Now it looks like the CNV deployment function is being called twice - once after the OCP install and again after the ODF install. We may need to handle this in a way that ensures the solution works for both non-MDR and MDR deployments

PrasadDesala: Just to clarify, earlier I add the skip in the deployment.py while calling the function
https://github.com/red-hat-storage/ocs-ci/pull/8873/files#diff-8571fc52f1040ac2cb01985da6d61cc93a535f812814c200b45d00daf28c2e3fR552-R555

petr-balogh: then we should bring that condition back on place but adjust for the other scenario as well there I think.

vavuthu: As per previous comments, looks like it is irrelevant PR in this case. So closing this

PR #9234, opened 2024-01-25
assisted installer api: removing deprecated keys

dahorak:
Newest version of Assisted Installer API deprecated `api_vip` and `ingress_vip` keys in favor of `api_vips` and `ingress_vips`.

PR #9279, opened 2024-02-05
update assisted_installer code base on recent changes in api

dahorak:
This fix is related/similar to PR https://github.com/red-hat-storage/ocs-ci/pull/9234, but in this case it is about response from Assisted installer API.


dahorak:
Triggered verification job here: https://url.corp.redhat.com/4da785c

petr-balogh (approved):

amr1ta (approved):
looking good

PR #9293, opened 2024-02-06
Extend timeout for ocp deployment via Assisted Installer

dahorak:
The timeout for OCP deployment via Assisted installer were set to 1 hour, but we spotted few cases when it takes little bit longer (e.g. 70 minutes or so) and the deployment was still successful. For example:
https://url.corp.redhat.com/8ee7238
image

It might be related to the data center occupation and performance.
I'm proposing to extend the timeout to two hours in this PR.


amr1ta (approved):
Looking good.

PR #9323, opened 2024-02-13
Multi-storagecluster deployment

shyRozen:
Multi-storagecluster deployment according to https://issues.redhat.com/browse/RHSTOR-4422

### DIFF
--- a/conf/ocsci/enable-multi-storagecluster.yaml
+++ b/conf/ocsci/enable-multi-storagecluster.yaml
@@ -1,0 +1,9 @@
+# config file enable multi-storagecluster deployment which install internal and external cluster in the same cluster
+DEPLOYMENT:
+  multi_storagecluster: True
+
+ENV_DATA:
+  storage_cluster_name: 'ocs-storagecluster'
+  external_storage_cluster_name: 'ocs-external-storagecluster'
+  external_storage_cluster_namespace: 'openshift-storage-extended'
+
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -38,6 +38,7 @@ from ocs_ci.ocs.cluster import (
     validate_pdb_creation,
     CephClusterExternal,
     get_lvm_full_version,
+    check_cephcluster_status,
 )
 from ocs_ci.ocs.exceptions import (
     CephHealthException,
@@ -96,6 +97,7 @@ from ocs_ci.ocs.utils import (
     get_all_acm_indexes,
     get_active_acm_index,
     enable_mco_console_plugin,
+    label_pod_security_admission,
 )
 from ocs_ci.utility.deployment import (
     create_external_secret,
@@ -134,6 +136,7 @@ from ocs_ci.utility.utils import (
     TimeoutSampler,
     get_latest_acm_tag_unreleased,
     get_oadp_version,
+    ceph_health_check_multi_storagecluster_external,
 )
 from ocs_ci.utility.vsphere_nodes import update_ntp_compute_nodes
 from ocs_ci.helpers import helpers
@@ -282,6 +285,7 @@ class Deployment(object):
         # Multicluster operations
         if config.multicluster:
             config.switch_acm_ctx()
+
             self.deploy_gitops_operator()
 
             logger.info("Creating GitOps CLuster Resource")
@@ -1507,30 +1511,33 @@ class Deployment(object):
         external/indpendent RHCS cluster
 
         """
-        live_deployment = config.DEPLOYMENT.get("live_deployment")
-        logger.info("Deploying OCS with external mode RHCS")
-        ui_deployment = config.DEPLOYMENT.get("ui_deployment")
-        if not ui_deployment:
-            logger.info("Creating namespace and operator group.")
-            run_cmd(f"oc create -f {constants.OLM_YAML}")
-        if not live_deployment:
-            create_catalog_source()
-        self.subscribe_ocs()
-        operator_selector = get_selector_for_ocs_operator()
-        subscription_plan_approval = config.DEPLOYMENT.get("subscription_plan_approval")
-        ocs_operator_names = get_required_csvs()
-        channel = config.DEPLOYMENT.get("ocs_csv_channel")
-        for ocs_operator_name in ocs_operator_names:
-            package_manifest = PackageManifest(
-                resource_name=ocs_operator_name,
-                selector=operator_selector,
-                subscription_plan_approval=subscription_plan_approval,
-            )
-            package_manifest.wait_for_resource(timeout=300)
-            csv_name = package_manifest.get_current_csv(channel=channel)
-            csv = CSV(resource_name=csv_name, namespace=self.namespace)
-            csv.wait_for_phase("Succeeded", timeout=720)
 
+        if not config.DEPLOYMENT.get("multi_storagecluster"):
+            live_deployment = config.DEPLOYMENT.get("live_deployment")
+            logger.info("Deploying OCS with external mode RHCS")
+            ui_deployment = config.DEPLOYMENT.get("ui_deployment")
+            if not ui_deployment:
+                logger.info("Creating namespace and operator group.")
+                run_cmd(f"oc create -f {constants.OLM_YAML}")
+            if not live_deployment:
+                create_catalog_source()
+            self.subscribe_ocs()
+            operator_selector = get_selector_for_ocs_operator()
+            subscription_plan_approval = config.DEPLOYMENT.get(
+                "subscription_plan_approval"
+            )
+            ocs_operator_names = get_required_csvs()
+            channel = config.DEPLOYMENT.get("ocs_csv_channel")
+            for ocs_operator_name in ocs_operator_names:
+                package_manifest = PackageManifest(
+                    resource_name=ocs_operator_name,
+                    selector=operator_selector,
+                    subscription_plan_approval=subscription_plan_approval,
+                )
+                package_manifest.wait_for_resource(timeout=300)
+                csv_name = package_manifest.get_current_csv(channel=channel)
+                csv = CSV(resource_name=csv_name, namespace=self.namespace)
+                csv.wait_for_phase("Succeeded", timeout=720)
         # Set rook log level
         self.set_rook_log_level()
 
@@ -1542,12 +1549,27 @@ class Deployment(object):
         # get admin keyring
         external_cluster.get_admin_keyring()
 
+        cluster_data = templating.load_yaml(constants.EXTERNAL_STORAGE_CLUSTER_YAML)
+
+        if config.DEPLOYMENT.get("multi_storagecluster"):
+            cluster_data["metadata"]["namespace"] = config.ENV_DATA[
+                "external_storage_cluster_namespace"
+            ]
+            cluster_data["metadata"]["name"] = config.ENV_DATA[
+                "external_storage_cluster_name"
+            ]
+            exec_cmd(
+                f"oc create -f {constants.MULTI_STORAGECLUSTER_EXTERNAL_NAMESPACE}"
+            )
+            label_pod_security_admission(
+                namespace=constants.OPENSHIFT_STORAGE_EXTENDED_NAMESPACE
+            )
+            exec_cmd(f"oc create -f {constants.STORAGE_SYSTEM_ODF_EXTERNAL}")
+        else:
+            cluster_data["metadata"]["name"] = config.ENV_DATA["storage_cluster_name"]
+
         # Create secret for external cluster
         create_external_secret()
-
-        cluster_data = templating.load_yaml(constants.EXTERNAL_STORAGE_CLUSTER_YAML)
-        cluster_data["metadata"]["name"] = config.ENV_DATA["storage_cluster_name"]
-
         # Use Custom Storageclass Names
         if config.ENV_DATA.get("custom_default_storageclass_names"):
             storageclassnames = config.ENV_DATA.get("storageclassnames")
@@ -1610,8 +1632,26 @@ class Deployment(object):
         # enable secure connection mode for in-transit encryption
         if config.ENV_DATA.get("in_transit_encryption"):
             external_cluster.enable_secure_connection_mode()
-
-        setup_ceph_toolbox()
+        if config.DEPLOYMENT.get("multi_storagecluster"):
+            logger.info("not setting toolbox in multi-storagecluster")
+        else:
+            setup_ceph_toolbox()
+        logger.info("Checking ceph health for external cluster")
+        if not config.DEPLOYMENT.get("multi_storagecluster"):
+            try:
+                ceph_health_check(
+                    tries=30,
+                    delay=10,
+                )
+            except CephHealthException:
+                raise CephHealthException("External ceph cluster not healthy")
+        else:
+            try:
+                ceph_health_check_multi_storagecluster_external()
+            except CephHealthException:
+                raise CephHealthException(
+                    "External multi-storagecluster external ceph cluster not healthy"
+                )
 
     def set_rook_log_level(self):
         rook_log_level = config.DEPLOYMENT.get("rook_log_level")
@@ -1717,6 +1757,17 @@ class Deployment(object):
                     defaults.CEPHFILESYSTEM_NAME = cfs_name
                 else:
                     logger.error("MDS deployment Failed! Please check logs!")
+            if config.DEPLOYMENT.get("multi_storagecluster"):
+                self.deploy_with_external_mode()
+                # Checking external cephcluster health
+                retry((CephHealthException, CommandFailed), tries=5, delay=20,)(
+                    check_cephcluster_status(
+                        desired_phase="Connected",
+                        desired_health="HEALTH_OK",
+                        name=constants.EXTERNAL_CEPHCLUSTER_NAME,
+                        namespace=constants.OPENSHIFT_STORAGE_EXTENDED_NAMESPACE,
+                    )
+                )
 
         # Change monitoring backend to OCS
         if config.ENV_DATA.get("monitoring_enabled") and config.ENV_DATA.get(
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -30,6 +30,7 @@ from ocs_ci.ocs.exceptions import (
     ThinPoolUtilityWrong,
     TimeoutExpiredError,
     ResourceWrongStatusException,
+    CephHealthException,
 )
 from ocs_ci.ocs.resources import ocs, storage_cluster
 import ocs_ci.ocs.constants as constant
@@ -2385,20 +2386,23 @@ class CephClusterExternal(CephCluster):
     """
 
     def __init__(self):
-        self.POD = ocp.OCP(kind="Pod", namespace=config.ENV_DATA["cluster_namespace"])
-        self.CEPHCLUSTER = ocp.OCP(
-            kind="CephCluster", namespace=config.ENV_DATA["cluster_namespace"]
-        )
+        if config.DEPLOYMENT.get("multi_storagecluster"):
+            namespace = constants.OPENSHIFT_STORAGE_EXTENDED_NAMESPACE
+        else:
+            namespace = config.ENV_DATA["cluster_namespace"]
+        self.POD = ocp.OCP(kind="Pod", namespace=namespace)
+        self.CEPHCLUSTER = ocp.OCP(kind="CephCluster", namespace=namespace)
 
         self.wait_for_cluster_cr()
         self._cluster_name = self.cluster_resource.get("metadata").get("name")
         self._namespace = self.cluster_resource.get("metadata").get("namespace")
         self.cluster = ocs.OCS(**self.cluster_resource)
-        # Decrease chance that we will hit issue:
-        # https://github.com/red-hat-storage/ocs-ci/issues/5186
-        logger.info("Sleep for 60 seconds before verifying MCG")
-        time.sleep(60)
-        self.wait_for_nooba_cr()
+        if not config.DEPLOYMENT.get("multi_storagecluster"):
+            # Decrease chance that we will hit issue:
+            # https://github.com/red-hat-storage/ocs-ci/issues/5186
+            logger.info("Sleep for 60 seconds before verifying MCG")
+            time.sleep(60)
+            self.wait_for_nooba_cr()
 
     @property
     def cluster_name(self):
@@ -2431,9 +2435,9 @@ class CephClusterExternal(CephCluster):
         sample = TimeoutSampler(timeout=timeout, sleep=3, func=self.is_health_ok)
         if not sample.wait_for_func_status(result=True):
             raise exceptions.CephHealthException("Cluster health is NOT OK")
-
-        self.wait_for_noobaa_health_ok()
-        self.validate_pvc()
+        if not config.DEPLOYMENT.get("multi_storagecluster"):
+            self.wait_for_noobaa_health_ok()
+            self.validate_pvc()
 
     def validate_pvc(self):
         """
@@ -3268,3 +3272,45 @@ def client_clusters_health_check():
         client_cluster_health_check()
 
     logger.info("The client clusters health check passed successfully")
+
+
+def check_cephcluster_status(
+    desired_phase="Connected",
+    desired_health="HEALTH_OK",
+    name=constants.EXTERNAL_CEPHCLUSTER_NAME,
+    namespace=constants.OPENSHIFT_STORAGE_EXTENDED_NAMESPACE,
+):
+    """
+    Check cephcluster health and phase.
+
+    Args:
+        desired_phase (string): The cephcluster desired phase.
+        desired_health (string): The cephcluster desired health.
+        name (string): name of the cephcluster.
+        namespace (string): namespace of the cephcluster.
+
+    Returns:
+        bool: True incase cluster is healthy and connected.
+
+    Raises:
+        CephHealthException incase phase or health are not as expected.
+
+    """
+    cephcluster = OCP(
+        kind=constants.CEPH_CLUSTER, resource_name=name, namespace=namespace
+    )
+    cc_resource = cephcluster.get()
+    if (
+        cc_resource["status"]["phase"] == desired_phase
+        and cc_resource["status"]["ceph"]["health"] == desired_health
+    ):
+        logger.info(
+            f"Cephcluster health is {desired_health} and phase is {desired_phase}"
+        )
+        return True
+    else:
+        logger.warning(
+            f'Cephcluster not healthy - phase is {cc_resource["status"]["phase"]} and health is'
+            f' {cc_resource["status"]["ceph"]["health"]}'
+        )
+        raise CephHealthException()
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -190,6 +190,7 @@ MACHINEHEALTHCHECK = "machinehealthcheck"
 STORAGECLIENT = "StorageClient"
 MANAGED_FUSION_OFFERING = "ManagedFusionOffering"
 CEPH_CLUSTER = "CephCluster"
+EXTERNAL_CEPHCLUSTER_NAME = "ocs-external-storagecluster-cephcluster"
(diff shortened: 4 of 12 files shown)


shyRozen:
Internal mode local verification https://privatebin.corp.redhat.com/?de7ab6195ab7ab91#Byk9bZ25HM189hmqmFv5ZQWqQRzrnkmoK3gw4m8dpWJm
Sent another PR validation for internal mode.

ocs_ci/deployment/deployment.py
```diff
@@ -1626,8 +1648,23 @@ def deploy_with_external_mode(self):
         # enable secure connection mode for in-transit encryption
         if config.ENV_DATA.get("in_transit_encryption"):
             external_cluster.enable_secure_connection_mode()
-
-        setup_ceph_toolbox()
+        if config.DEPLOYMENT.get("multi_storagecluster"):
+            logger.info("not setting toolbox in multi-storagecluster")
+            health_check_namespace = constants.OPENSHIFT_STORAGE_EXTENDED_NAMESPACE
```
petr-balogh, line 1653: Why we are not setting up toolbox pod for external cluster?
I think we should have two toolbox pods, one for internal mode - this will be created by default from non external cluster deployment. And we should have second toolbox pod - probably in this other namespace for external mode cluster - which will allow us to see status and health of external cluster.

ocs_ci/utility/utils.py
```diff
@@ -2356,16 +2358,27 @@ def run_ceph_health_cmd(namespace):
...
+            get_external_cluster_client,
+        )
 
-    try:
-        ct_pod = get_ceph_tools_pod(namespace=namespace)
-    except (AssertionError, CephToolBoxNotFoundException) as ex:
-        raise CommandFailed(ex)
+        host, user, password, ssh_key = get_external_cluster_client()
+        connection_to_cephcluster = Connection(host=host, user=user, password=password)
+        return connection_to_cephcluster.exec_cmd("ceph health")
```
petr-balogh, line 2369: If possible , I prefer to have toolbox pod for external mode as well.
This function is suppose to return only string, but you are returning tuple in this case, and you are checking only the health of external cluster here. I think we should do check on ceph health of both Internal and External cluster.

ocs_ci/utility/utils.py
```diff
@@ -2313,7 +2314,8 @@ def ceph_health_check_base(namespace=None):
     """
     namespace = namespace or config.ENV_DATA["cluster_namespace"]
     health = run_ceph_health_cmd(namespace)
-
+    if type(health) is tuple:
```
petr-balogh, line 2317: Dont' like this to be honest as you are not returning string as we were suppose to return from run_ceph_health_cmd.

I think this function should handle to run the health check two times on both namespaces to check status of both: internal/external mode.

ocs_ci/deployment/deployment.py
shyRozen: After I had problem in setting up toolbox in another namespace, Dev gave it a try too and after failure they asked me to work directly with the ceph cluster.

ocs_ci/utility/utils.py
shyRozen: No toolbox in multi-storagecluster. Check my comment above.

ocs_ci/utility/utils.py
shyRozen: I disagree. I don't want to handle every time the function is called both health checks as when I install the external I want to know external health and when internal - intenal health. What I can do is to return a string when I do ssh and drop this check.
Like change this https://github.com/red-hat-storage/ocs-ci/blob/3ff20a16395ddfe08a61214523fc099f9cabca2f/ocs_ci/utility/utils.py#L2369
to:
```
ssh_output = connection_to_cephcluster.exec_cmd("ceph health")
return ssh_output[1]
```
After this I can remove the tuple check.

conf/ocsci/enable-multi-storagecluster.yaml
```diff
@@ -0,0 +1,13 @@
...
+ENV_DATA:
+  storage_cluster_name: 'ocs-storagecluster'
+  external_storage_cluster_name: 'ocs-external-storagecluster'
+  external_storage_cluster_namespace: 'openshift-storage-extended'
+
+EXTERNAL_MODE:
+  # base64 encoded data of json output
+  # from exporter script
+  # https://github.com/rook/rook/blob/master/cluster/examples/kubernetes/ceph/create-external-cluster-resources.py
+  external_cluster_details: PLACE_HOLDER
```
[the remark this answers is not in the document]
shyRozen: I'll drop this but it was just copied from external yaml.

ocs_ci/deployment/deployment.py
shyRozen: I think the main problem is all rook resource are in another namespace and can't be shared like secret etc.

ocs_ci/utility/utils.py
shyRozen: What I meant is the first stage I'm installing internal mode and then I would like to check that internal is healthy and then install the external and check that the external is healthy. We can have a new function that call both. WDYT?

shyRozen: https://github.com/shyRozen/ocs-ci/blob/892b7caeb562ee873e91d2a5b5f87535045e727e/ocs_ci/utility/utils.py#L2328

ocs_ci/utility/utils.py
shyRozen: The function now return only strings and no need to check if tuple.

ocs_ci/utility/utils.py
```diff
@@ -2313,14 +2316,61 @@ def ceph_health_check_base(namespace=None):
...
     health = run_ceph_health_cmd(namespace)
-
     if health.strip() == "HEALTH_OK":
         log.info("Ceph cluster health is HEALTH_OK.")
         return True
     else:
         raise CephHealthException(f"Ceph cluster health is not OK. Health: {health}")
 
 
+def ceph_health_check_internal_and_external(cluster_platform):
```
petr-balogh, line 2326: I would tackle it differently.

Have separate function to perform only external cluster health check.
Add parameter `cluster_type`  (allowed values: internal_and_external, internal, external)to those functions:
`ceph_health_check_base` and `ceph_health_check`
Set default value for  `cluster_type="internal_and_external"`

Update those two function to tackle in the loop health check on internal, external, both based on the parameter.
Of course in case we don't run on multicluster type of deployment we will skip the external part.

**The reason:**
We already call many times in our tests the ceph health check function: `ceph_health_check`.
Where we are making sure the ceph is in healthy state - I would still except that in multicluster deployment, after any test - both internal and external cluster should be in healthy state.

I see you introduced some new exceptions, but I would do it this way if you want to introduce new CephHealthExceptions I would create:
`InternalCephHealthException` and `ExternalCephHealthException`
where both should inherit from original `CephHealthException` .

In case we already rely on ond except such exception `CephHealthException` in the code, then the generic
```python
except CephHealthException as ex:
```
Will cover also new exceptions which will tell us more precise where the health issue occurred.

shyRozen: **Have separate function to perform only external cluster health check** - The function support namespace which can serve us for internal or external, openshift-storage for internal and openshift-storage-extended for external. So no need for separate function. Just need a parameter for multi or act on config.DEPLOYMENT key.
In general I understand the need for before and after test. I'll will implement the function but leave me space to decide about the parameters. I'll do it in a manner that we will have the minimal changes to tests and framework.

ocs_ci/ocs/constants.py
```diff
@@ -222,6 +223,7 @@
 TEST_FILES_BUCKET = "ocsci-test-files"
 ROOK_REPOSITORY = "https://github.com/rook/rook.git"
 OPENSHIFT_STORAGE_NAMESPACE = "openshift-storage"
+OPENSHIFT_STORAGE_EXTENDED_NAMESPACE = "openshift-storage-extended"
```
petr-balogh, line 226: Can this one be something else than this one? Like we can have custom namespace?

If so, I would not rely on this one value from constants but rather have it defined in config - in past we had several issues because of using namespace from constants, mainly for managed cluster deployments - provider/client and so on, where we have different namespaces as well - in this case it should be configurable based on config which I prefer here as well in case it's possible to have custom namespace name for external cluster.

ocs_ci/utility/utils.py
petr-balogh: just summarizing the call we had.

- We wil add new parameter to ceph_health_check and ceph_health_check_base function to parametrize if it should do internal, external or both. (We can update namespace parameter docs - that this parameter is applicable only for internal mode deployment.) 
- This new parameter will be by default `internal` (Because all current use cases were only care of the internal mode - or regular external mode)
- All calls of ceph_health_check function in ceph_health_checker fixture: https://github.com/red-hat-storage/ocs-ci/blob/11cf368e1a02cb232b8d8dbf39c6e781638e44b4/tests/conftest.py#L1620 we will by default have `internal` , but in case it's multicluster deployment - we will check health of `both`.

shyRozen: Agree to all but **we will by default have internal , but in case it's multicluster deployment - we will check health of both** there will be no default because there will a variable in the ceph_health when calling the  function the parameter that will be decided upon all conditions checks.

shyRozen: In the function definition there will be default to "internal".

ocs_ci/deployment/deployment.py
```diff
@@ -1626,8 +1648,23 @@ def deploy_with_external_mode(self):
...
+            health_check_namespace = constants.OPENSHIFT_STORAGE_NAMESPACE
+        logger.info("Checking ceph health for external cluster")
+        try:
+            ceph_health_check(
+                namespace=health_check_namespace,
+                tries=30,
+                delay=10,
+            )
+        except CephHealthException as ex:
+            raise MultiStorageClusterExternalCephHealth(
```
ebenahar, line 1665: `MultiStorageClusterExternalCephHealth` is applicable only for the case of multi storage clusters. But here Ceph health check is done on any external mode cluster

shyRozen: Now the whole logic of ceph_health_check is changed and other exception are used. Fixed it to have CephHealthExceptionExternal

ocs_ci/utility/utils.py
```diff
@@ -2356,16 +2406,28 @@ def run_ceph_health_cmd(namespace):
...
+        from ocs_ci.deployment.helpers.external_cluster_helpers import (
+            get_external_cluster_client,
+        )
 
-    try:
-        ct_pod = get_ceph_tools_pod(namespace=namespace)
-    except (AssertionError, CephToolBoxNotFoundException) as ex:
-        raise CommandFailed(ex)
+        host, user, password, ssh_key = get_external_cluster_client()
+        connection_to_cephcluster = Connection(host=host, user=user, password=password)
```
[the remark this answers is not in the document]
shyRozen: Fixed to have both options

ocs_ci/ocs/exceptions.py
```diff
@@ -690,3 +690,15 @@ class NoRunningCephToolBoxException(Exception):
 
 class UsernameNotFoundException(Exception):
     pass
+
+
+class MultiStorageClusterExternalCephHealth(Exception):
```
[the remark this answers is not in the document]
shyRozen: Implement this for new exceptions created. This one was deleted.

ocs_ci/utility/utils.py
```diff
@@ -2313,14 +2316,61 @@ def ceph_health_check_base(namespace=None):
...
+    try:
+        ceph_health_check_base(namespace=constants.OPENSHIFT_STORAGE_NAMESPACE)
+    except CephHealthException as ex:
+        log.warning(f"Internal health problem {ex}")
+        internal_exception = True
+        err = str(ex)
+        log.warning(f"Ceph health check failed with {err}")
+        if "clock skew detected" in err:
+            log.info(f"Changing NTP on compute nodes to" f" {constants.RH_NTP_CLOCK}")
+            if cluster_platform == constants.VSPHERE_PLATFORM:
```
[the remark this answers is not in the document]
shyRozen: Whole function removed.

ocs_ci/utility/utils.py
```diff
@@ -2356,16 +2435,35 @@ def run_ceph_health_cmd(namespace):
...
-    return ct_pod.exec_ceph_cmd(
-        ceph_cmd="ceph health", format=None, out_yaml_format=False, timeout=120
-    )
+    else:
+        from ocs_ci.ocs.resources.pod import get_ceph_tools_pod
+
+        try:
+            ct_pod = get_ceph_tools_pod(namespace=namespace)
+        except (AssertionError, CephToolBoxNotFoundException) as ex:
+            raise CommandFailed(ex)
```
ebenahar, line 2462: why is try except needed here? Simply let the original exception to be raised

petr-balogh: If it's taken from other place I agree to keep it this way as it was probably like that for some reason and changing it brings risk of other regression.

ocs_ci/ocs/cluster.py
```diff
@@ -3188,3 +3192,42 @@ def get_mon_quorum_ranks():
...
+):
+    """
+    Check cephcluster health and phase.
+    Args:
+        desired_phase (string): The cephcluster desired phase.
+        desired_health (string): The cephcluster desired health.
+        name (string): name of the cephcluster.
+        namespace (string): namespace of the cephcluster.
+    Returns:
+        (bool) True incase cluster is healthy and connected.
```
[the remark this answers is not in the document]
petr-balogh:
```suggestion
        bool: True incase cluster is healthy and connected.
```
Those sections should be documented like this. The same for raises section.

ocs_ci/ocs/cluster.py
```diff
@@ -3188,3 +3192,42 @@ def get_mon_quorum_ranks():
...
+    Raise:
+        (CephHealthException) incase phase or health are not as expected.
```
petr-balogh, line 3213:
```suggestion
    Raises:
        CephHealthException: incase phase or health are not as expected.
```

ocs_ci/utility/utils.py
```diff
@@ -2356,16 +2432,35 @@ def run_ceph_health_cmd(namespace):
...
-        raise CommandFailed(ex)
+        host, user, password, ssh_key = get_external_cluster_client()
+        if not password and ssh_key:
+            connection_to_cephcluster = Connection(
+                host=host, user=user, private_key=ssh_key
+            )
+        elif password:
+            connection_to_cephcluster = Connection(
+                host=host, user=user, password=password
+            )
```
petr-balogh, line 2449: Connection itself handle already the ssh_key with higher priority as you can see here:
https://github.com/shyRozen/ocs-ci/blob/9ca9c4e26e8fd361962d410d2e297f3bfeddb609/ocs_ci/utility/connection.py#L66

```suggestion
        connection_to_cephcluster = Connection(
                host=host, user=user, password=password, private_key=ssh_key
         )
```

ocs_ci/utility/utils.py
```diff
@@ -2282,17 +2287,47 @@ def ceph_health_check(namespace=None, tries=20, delay=30):
...
+    ):
+        # checking if external mode is standalone or multi-storagecluster to determine namespace
+        if namespace is None:
+            if cluster_type == constants.CLUSTER_TYPE_EXTERNAL:
+                if config.DEPLOYMENT.get("multi_storagecluster"):
+                    namespace = constants.OPENSHIFT_STORAGE_EXTENDED_NAMESPACE
+                else:
+                    namespace = constants.OPENSHIFT_STORAGE_NAMESPACE
+            else:
+                namespace = constants.OPENSHIFT_STORAGE_NAMESPACE
```
petr-balogh, line 2303:
```suggestion
        if namespace is None:
            if cluster_type == constants.CLUSTER_TYPE_EXTERNAL:
                if config.DEPLOYMENT.get("multi_storagecluster"):
                    namespace = constants.OPENSHIFT_STORAGE_EXTENDED_NAMESPACE
```

The namespace is correctly taken from configuration in ceph_healht_check_base here:

https://github.com/shyRozen/ocs-ci/blob/9ca9c4e26e8fd361962d410d2e297f3bfeddb609/ocs_ci/utility/utils.py#L2349

For regular deployment, namespace should always be taken from config, as in fusion cluster deployment, there is different namespace for ODF used, and we are relaying on config value, and should not on constants.
@fbalak  tried to fix that on most of the places as part of his [PR](https://github.com/red-hat-storage/ocs-ci/pull/7501/files) some time back, this will break the logic.

ocs_ci/utility/utils.py
```diff
@@ -2313,14 +2348,55 @@ def ceph_health_check_base(namespace=None):
...
+        subprocess.TimeoutExpired: Incase timeout.
+        NoRunningCephToolBoxException: Incase toolbox doesn't exist - relevant only for internal cluster as
+                                       multi-storagecluster external uses ssh
+
+    Returns:
+        True: True if both ceph clusters are healthy.
+
+    """
+    try:
+        ceph_health_check_base(namespace=constants.OPENSHIFT_STORAGE_NAMESPACE)
```
petr-balogh, line 2375:
```suggestion
        ceph_health_check_base()
```
Let the function to determine the namespace from config itself.

ocs_ci/deployment/deployment.py
```diff
@@ -1626,8 +1649,24 @@ def deploy_with_external_mode(self):
         # enable secure connection mode for in-transit encryption
         if config.ENV_DATA.get("in_transit_encryption"):
             external_cluster.enable_secure_connection_mode()
+        if config.DEPLOYMENT.get("multi_storagecluster"):
+            logger.info("not setting toolbox in multi-storagecluster")
+            health_check_namespace = constants.OPENSHIFT_STORAGE_EXTENDED_NAMESPACE
+        else:
+            setup_ceph_toolbox()
+            health_check_namespace = constants.OPENSHIFT_STORAGE_NAMESPACE
```
petr-balogh, line 1657:
```suggestion
            health_check_namespace = None
```
Let the ceph_healt_check function dertermin itslef the namespace form config in this case

tests/conftest.py
```diff
@@ -1656,9 +1656,13 @@ def finalizer():
                     # We are allowing 20 re-tries for health check, to avoid teardown failures for cases like:
                     # "flip-flopping ceph health OK and warn because of:
                     # HEALTH_WARN Reduced data availability: 2 pgs peering
-                    ceph_health_check(
-                        namespace=ocsci_config.ENV_DATA["cluster_namespace"]
-                    )
+                    cluster_type = constants.CLUSTER_TYPE_INTERNAL
+                    if ocsci_config.DEPLOYMENT.get("external_mode"):
+                        cluster_type = constants.CLUSTER_TYPE_EXTERNAL
```
petr-balogh, line 1661: Classic external mode deployment will run the same ceph health as of now, we should not do any changes here, just in case of multi cluster we should make sure all ceph health in this fixture cover both clusters health.

```suggestion
```

petr-balogh: There is missing also the ceph_health_check call on the line 1681 and 1691 to be updated to pass the cluster_type = constants.CLUSTER_TYPE_MULTI_SC

So I would move the cluster type definition on top level of fixture, and then you can just propagate it to all places.

ocs_ci/utility/utils.py
```diff
@@ -2313,14 +2348,55 @@ def ceph_health_check_base(namespace=None):
...
+    Raises:
+        CephHealthExceptionInternal: Internal cluster ceph health problem.
+        CephHealthExceptionExternal: External cluster ceph health problem.
+        CommandFailed: Incase command failed.
+        subprocess.TimeoutExpired: Incase timeout.
+        NoRunningCephToolBoxException: Incase toolbox doesn't exist - relevant only for internal cluster as
+                                       multi-storagecluster external uses ssh
+
+    Returns:
+        True: True if both ceph clusters are healthy.
```
petr-balogh, line 2371:
```suggestion
        bool: True if both ceph clusters are healthy.
```

Docstring should contains the returns type here, not specific value

ocs_ci/ocs/cluster.py
```diff
@@ -3188,3 +3192,45 @@ def get_mon_quorum_ranks():
...
+    Check cephcluster health and phase.
+
+    Args:
+        desired_phase (str): The cephcluster desired phase.
+        desired_health (str): The cephcluster desired health.
+        name (str): name of the cephcluster.
+        namespace (str): namespace of the cephcluster.
+
+    Returns:
+        True: incase cluster is healthy and connected.
```
petr-balogh, line 3213:
```suggestion
        bool: True in case cluster is healthy and connected.
```


Docstring should contains the returns type here, not specific value

ocs_ci/deployment/deployment.py
shyRozen: Dropped the condition for health_check_namespace and sending ceph_health_check with no parameters.

ocs_ci/utility/utils.py
```diff
@@ -2267,13 +2270,15 @@ def get_az_count():
         return 1
 
 
-def ceph_health_check(namespace=None, tries=20, delay=30):
+def ceph_health_check(namespace=None, tries=20, delay=30, cluster_type="internal"):
```
vavuthu, line 2273: this breaks the existing flow. 
eg: ceph_health_check(tries=30, delay=60)

shyRozen: The function example you gave should work. What is the reason for it not to work?

petr-balogh:
Hello @shyRozen I asked @vavuthu to have a look to your PR, then we met and discussed together.

1) We need to simplify it and keep it simple as possible
2) As we discussed for first time, I thought that we will have the ceph_helth_check called by default for collecting ceph health status for both, after we discussed together @shyRozen , I tend to agree that we will do by default only internal mode.
3) As we don't want to change default behavior of current call of ceph_health to cover both cases, we should remove all logic from current functions for ceph_health_check to cover only ceph health of current internal/classic External mode which means, we want to remove all updates of current functions like: `ceph_health_check`, `ceph_health_check_base` and `run_ceph_health_cmd` - so those functions should be untouched to keep current behavior and not introduce more complexity to existing functions than needed as it's now confusing with those cluster_type and so on where we want to distinguish if it's classic external deployment or if it's external of multicluster deployment and so on...
4) The functionality to cover Ceph health check for multicluster scenario for external cluster as namespace is static - let's have the new function for only this scenario, we don't need any namespace parameter for it as it's static and defined in constants.
5) If we want, we can introduce wrapper function which will do check on both Ceph clusters in case of multicluster scenario.
6) In `health_checker` Function - we can call either wrapper function described from step 5), or if we won't have it, just call current ceph_health_check as is + in case of multicluster job `if ocsci_config.DEPLOYMENT.get("multi_storagecluster")`, will do extra call of separate function covering only health check of External cluster. I think wrapper function sound reasonable, as the logic of calling the ceph health check on external Ceph cluster can be conditionalized directly there based on config.


TL;DR:
* Don't touch current ceph health check functions
* Have only one extra function for external ceph health for multicluster scenario only
* update fixture to call those two functions in two separate calls

tests/conftest.py
```diff
@@ -1664,6 +1665,13 @@ def finalizer():
                         namespace=ocsci_config.ENV_DATA["cluster_namespace"]
                     )
                     log.info("Ceph health check passed at teardown!")
+                    if ocsci_config.DEPLOYMENT.get("multi_storagecluster"):
+                        ceph_health_check_multi_storagecluster_external()
+                        log.info(
+                            "Ceph health check for multi-storagecluster external cluster passed at teardown!"
+                        )
+                        multi_storagecluster_external_health_passed = True
```
petr-balogh, line 1673: I think that the variable should be initialized on the line above the `try` block line 1651. Because if the exception will be hit on the line 1669, then on the line 1690 you will fail on:
```
NameError: name 'multi_storagecluster_external_health_passed' is not defined
```

ocs_ci/deployment/deployment.py
```diff
@@ -1610,8 +1632,26 @@ def deploy_with_external_mode(self):
...
+            setup_ceph_toolbox()
+        logger.info("Checking ceph health for external cluster")
+        if not config.DEPLOYMENT.get("multi_storagecluster"):
+            try:
+                ceph_health_check(
+                    tries=30,
+                    delay=10,
+                )
+            except CephHealthException as ex:
+                raise CephHealthException(f"External ceph cluster not healthy - {ex}")
```
petr-balogh, line 1647:
```suggestion
                raise CephHealthException(f"External ceph cluster not healthy").with_traceback(ex.__traceback__)
```
I have found this solution here:
https://stackoverflow.com/questions/9157210/how-do-i-raise-the-same-exception-with-a-custom-message-in-python

WDYT?

The same for the line 1653

ocs_ci/ocs/cluster.py
petr-balogh: This still should be updated based on proper docstring practices to put the type and not the specific value.

ocs_ci/utility/utils.py
```diff
@@ -2387,6 +2389,54 @@ def run_ceph_health_cmd(namespace):
     )
 
 
+def ceph_health_multi_storagecluster_external_base():
+    """
+    Check ceph health for multi-storagecluster external implementation.
+
+    Returns:
+        True: If cluster health is ok.
```
petr-balogh, line 2397:
```suggestion
        bool: True if cluster health is ok.
```

petr-balogh:
Please rebase on latest master where we have fix for ping3 pin version to fix the PR check

ocs_ci/deployment/deployment.py
shyRozen: Did quite understand your intentions here. What is wrong in the first place?

ocs_ci/ocs/cluster.py
shyRozen: @ebenahar @petr-balogh each one of you is requesting different approach  https://github.com/red-hat-storage/ocs-ci/pull/9323#discussion_r1532067048 . I don't know which to follow.

ocs_ci/deployment/deployment.py
petr-balogh: We discussed offline we will do just:
```
except CephHealthException:
                raise CephHealthException(f"External ceph cluster not healthy")
                ```

File: ocs_ci/deployment/deployment.py [hunk 25]
```diff
@@ -1542,12 +1549,27 @@ def deploy_with_external_mode(self):
         # get admin keyring
         external_cluster.get_admin_keyring()
 
-        # Create secret for external cluster
-        create_external_secret()
-
         cluster_data = templating.load_yaml(constants.EXTERNAL_STORAGE_CLUSTER_YAML)
-        cluster_data["metadata"]["name"] = config.ENV_DATA["storage_cluster_name"]
 
+        if config.DEPLOYMENT.get("multi_storagecluster"):
```
ebenahar, line 1554: shouldn't it be `elif`?

petr-balogh (dismissed):

conf/ocsci/enable-multi-storagecluster.yaml
```diff
@@ -0,0 +1,7 @@
+DEPLOYMENT:
+  multi_storagecluster: True
```
petr-balogh, line 2: Can we document the new config values in https://github.com/red-hat-storage/ocs-ci/blob/master/conf/README.md

mainly:
multi_storagecluster
external_storage_cluster_name
external_storage_cluster_namespace

ebenahar (dismissed):

conf/ocsci/enable-multi-storagecluster.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+ENV_DATA:
+  storage_cluster_name: 'ocs-storagecluster'
+  external_storage_cluster_name: 'ocs-external-storagecluster'
+  external_storage_cluster_namespace: 'openshift-storage-extended'
+
+REPORTING:
+  polarion:
+    deployment_id: 'OCS-5503'
```
petr-balogh, line 12: Maybe as this is not production config, we should delete this section from here.

petr-balogh (approved):

vavuthu (approved):

ebenahar (approved):

PR #9357, opened 2024-02-22
Revert #8979 to patch `ocs-storagecluster` again instead of the noobaa CR at `test_disable_nb_lb` 

sagihirshfeld:
Fixes: #7656

This PR is just the appliance of `git revert` on #8979

I'm not sure why the back-and-forth changes were made in the first place, but this seems to be the procedure that works now and it matches the documentation: [Chapter 15. Disabling Multicloud Object Gateway external service after deploying OpenShift Data Foundation](https://access.redhat.com/documentation/en-us/red_hat_openshift_data_foundation/4.14/html/troubleshooting_openshift_data_foundation/disabling-multicloud-object-gateway-external-service-after-deploying-openshift-datat-foundation_rhodf#:~:text=When%20you%20deploy%20OpenShift%20Data,variable%20in%20the%20storagecluster%20CRD.)

Required backports:

- [ ] 4.15

### DIFF
--- a/tests/functional/object/mcg/test_s3_routes.py
+++ b/tests/functional/object/mcg/test_s3_routes.py
@@ -146,18 +146,18 @@ class TestS3Routes:
         """
 
         def finalizer():
-            noobaa_ocp_obj = ocp.OCP(
-                kind="noobaa",
+            storagecluster_obj = ocp.OCP(
+                kind=constants.STORAGECLUSTER,
                 namespace=config.ENV_DATA["cluster_namespace"],
-                resource_name="noobaa",
+                resource_name=constants.DEFAULT_CLUSTERNAME,
             )
             try:
-                if noobaa_ocp_obj.data["spec"]["disableLoadBalancerService"]:
-                    lb_param = (
-                        '[{"op": "remove", "path": "/spec/disableLoadBalancerService"}]'
-                    )
+                if storagecluster_obj.data["spec"]["multiCloudGateway"][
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_s3_routes.py
```diff
@@ -146,18 +146,18 @@ def revert_lb_service(self, request):
...
                 namespace=config.ENV_DATA["cluster_namespace"],
-                resource_name="noobaa",
+                resource_name=constants.DEFAULT_CLUSTERNAME,
             )
             try:
-                if noobaa_ocp_obj.data["spec"]["disableLoadBalancerService"]:
-                    lb_param = (
-                        '[{"op": "remove", "path": "/spec/disableLoadBalancerService"}]'
-                    )
+                if storagecluster_obj.data["spec"]["multiCloudGateway"][
```
DanielOsypenko, line 155: do you want to get KeyError if the StorageCluster does not have `multiCloudGateway` and nested `disableLoadBalancerService` ? if not we need to use .get()

sagihirshfeld: Getting a `KeyError` here from either would be ok since both would mean that `disableLoadBalancerService` is gone. However, we're expecting `storagecluster_obj.data["spec"]["multiCloudGateway"]` not to throw a KeyError since OCS-CI patches some nested properties to it during/right after deployment.

PR #9408, opened 2024-03-01
create s3cli in separate NS

vavuthu:
Currently awscli_pod_fixture is creating s3cli and statefulset in openshift-storage and cleaning the same resources irrespective which testcase is created. since resource are deleted during setup, when running testcases throuh ITR, test cases are failing.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -127,6 +127,7 @@ from ocs_ci.utility.utils import (
     get_default_if_keyval_empty,
     get_ocs_build_number,
     get_openshift_client,
+    get_random_str,
     get_testrun_name,
     load_auth_config,
     ocsci_log_path,
@@ -2511,6 +2512,7 @@ def awscli_pod_client_session(
 def awscli_pod_fixture(request, scope_name):
     """
     Creates a new AWSCLI pod for relaying commands
+
     Args:
         scope_name (str): The name of the fixture's scope,
         used for giving a descriptive name to the pod and configmap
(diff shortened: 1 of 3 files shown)


vavuthu:
thanks @prsurve for chnaging permissions to objects in s3-cli-with-test-objects-multiarch image and pushing with 4.0 tag

tests/conftest.py
```diff
@@ -2517,12 +2519,18 @@ def awscli_pod_fixture(request, scope_name):
         pod: A pod running the AWS CLI
 
     """
+    project = f"namespace-{get_random_str()}"
```
sagihirshfeld, line 2522: Please prefix the name with "s3cli" to make it easy to find

tests/conftest.py
```diff
@@ -2517,12 +2519,18 @@ def awscli_pod_fixture(request, scope_name):
...
+            ocp_obj.delete_project(project)
 
-    request.addfinalizer(awscli_pod_cleanup)
+    request.addfinalizer(lambda: delete_project(namespace=project))
+    request.addfinalizer(lambda: awscli_pod_cleanup(namespace=project))
 
-    log.info("Cleaning up any previous AWS CLI resources")
-    awscli_pod_cleanup()
-    return create_awscli_pod(scope_name)
+    ocp_obj.new_project(project)
```
sagihirshfeld, line 2532: This has a side effect of changing the default project to the newly created one. I don't think we're running any oc commands without specifying the namespace/project in ocs-ci, but please change the default project back to `ENV_DATA["cluster_namespace"]` to be on the safe side.

vavuthu: done

sagihirshfeld: Using `ocp.switch_to_default_rook_cluster_project` in the finalizer means that we'd only switch back to the default namespace after the tests that use this fixture are done, or at the end of the session when we're using this fixture with the session scope. 

Using this method right after the creation of the new project instead should guarantee we avoid this side effect:
```
    request.addfinalizer(lambda: delete_project(namespace=project))
    request.addfinalizer(lambda: awscli_pod_cleanup(namespace=project))

    ocp_obj.new_project(project)
    ocp.switch_to_default_rook_cluster_project()
```

vavuthu: done

sagihirshfeld (approved):
LGTM

petr-balogh (approved):

PR #9415, opened 2024-03-04
Deploy ODF BM cluster via UI

OdedViner:
```
--cluster-path ~/ClusterPath -m deployment --ocp-version 4.15 --ocs-version 4.15 --ocsci-conf conf/ocsci/ui_deployment.yaml --ocsci-conf conf/ocsci/myconfig.yaml --ocsci-conf conf/ocsci/skip_ocp_deploy.yaml --ocsci-conf  conf/deployment/baremetal/upi_1az_rhcos_nvme_intel_3m_3w.yaml --cluster-name cluster tests/ --deploy
```

### DIFF
--- a/ocs_ci/ocs/ui/deployment_ui.py
+++ b/ocs_ci/ocs/ui/deployment_ui.py
@@ -464,9 +464,6 @@ class DeploymentUI(PageNavigator):
                 self.dep_loc["choose_openshift-storage_project"], enable_screenshot=True
             )
 
-    def add_disk_for_bm_platform(self):
-        pass
-
     def install_ocs_ui(self):
         """
         Install OCS/ODF via UI.
@@ -474,9 +471,7 @@ class DeploymentUI(PageNavigator):
         """
         if config.DEPLOYMENT.get("local_storage"):
             create_optional_operators_catalogsource_non_ga()
-            if config.ENV_DATA.get("platform") == constants.BAREMETAL_PLATFORM:
-                self.add_disk_for_bm_platform()
-            elif config.ENV_DATA.get("platform") == constants.VSPHERE_PLATFORM:
+            if config.ENV_DATA.get("platform") == constants.VSPHERE_PLATFORM:
                 add_disk_for_vsphere_platform()
         self.install_local_storage_operator()
         self.install_ocs_operator()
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/ui/deployment_ui.py
```diff
@@ -224,8 +224,14 @@ def install_lso_cluster(self):
             self.do_click(
                 locator=self.dep_loc["all_nodes_create_sc"], enable_screenshot=True
             )
-        self.verify_disks_lso_attached()
-        self.do_click(self.dep_loc["next"], enable_screenshot=True)
+        if config.ENV_DATA.get("platform") != constants.BAREMETAL_PLATFORM:
+            self.verify_disks_lso_attached()
+            timeout_next = 60
+        else:
+            timeout_next = 600
```
ebenahar, line 231: this is a significant timeout. I suggest we poll the status of a certain resource that can indicate whether or not we can proceed to the next step

petr-balogh: Can you maybe elaborate why this is 10times more for BM deployment to go to next button?
Is it because it's not just clickable before and waiting till it's clickable or something like that?
Can we maybe wrap the line below with some retry on the specific exception with some e.g 30 seconds in between attempts to retry?

OdedViner: On BM we skip on `verify_disks_lso_attached` method because the disk size is not clear, and we can replace the disk on BM. [on LSO we add the disk "manually" and we can control the disk size. So my solution is waiting more time until we can click on the "next" button.

OdedViner: https://url.corp.redhat.com/70d7466

OdedViner: https://github.com/red-hat-storage/ocs-ci/pull/9415/files#diff-99d9a70366ab4f661fc2d565da8912b6664152141c29c0c6b5ac2da39c9d1c59L48

petr-balogh: Talked to Oded, and he explained me that the timeout is just timeout to wait maximum time for element to be clickable. Which means if it will be clickable in 30 seconds, it will continue in 30 seconds, not 600, but if the element is not clickable in 600 seconds (maximum timeout) it will fail the deployment.

petr-balogh (approved):

ebenahar (approved):

shyRozen (approved):

PR #9433, opened 2024-03-05
Censor additional data in MCG related commands

sagihirshfeld:


sagihirshfeld:
The OCS-CI logs from the above verification shows that all the targeted data has been censored as expected: https://url.corp.redhat.com/7776759

OdedViner (approved):

ocs_ci/ocs/constants.py
```diff
@@ -1205,6 +1205,14 @@
...
 # Config related constants
 config_keys_patterns_to_censor = ["passw", "token", "secret", "key", "credential"]
+config_keys_expressions_to_censor = [
+    "AccountKey",
+    "AccountName",
+    "ApplicationID",
+    "ApplicationSecret",
+    "LogsAnalyticsWorkspaceID",
+    "TenantID",
+]
```
udaysk23, line 1215: Cant we add these values as part of `config_keys_patterns_to_censor`

ocs_ci/helpers/helpers.py
```diff
@@ -4595,3 +4604,29 @@ def get_architecture_path(cli_type):
     elif system == "Darwin":  # Mac
         path = os.path.join(path, "macosx", image_prefix)
     return path
+
+
+def flatten_multilevel_dict(d):
```
ebenahar, line 4609: 💯

ebenahar (dismissed):

ocs_ci/ocs/constants.py
sagihirshfeld: That's the way I implemented it at first and it does technically work, but including full expressions along side pattern keywords like "secret" and "key" felt semantically wrong. This differentiation also prevents any potential side-effect censoring (if for example we include "AccountKey" under the patterns list then we would also censor a theoretical non-confidential key like "AccountKeyEncryptionType").

udaysk23: Thanks for the clarification

udaysk23 (approved):

petr-balogh (approved):

OdedViner (approved):

Issue #9446, opened 2024-03-10
`test_bucket_policy.py::test_object_actions` DeleteObject denial policy policy fails due a race condition 


sagihirshfeld:
Many of the regression runs of the TC are failing with the following error:
```
    # Verifying whether obc account is denied access to delete object
    logger.info(
        f"Verifying whether user: {obc_obj.obc_account} is denied to Delete object"
    )
    try:
        s3_delete_object(obc_obj, obc_obj.bucket_name, object_key)
    except boto3exception.ClientError as e:
        logger.info(e.response)
        response = HttpResponseParser(e.response)
        if response.error["Code"] == "AccessDenied":
            logger.info("Delete action has been denied access")
        else:
            raise UnexpectedBehaviour(
                f"{e.response} received invalid error code {response.error['Code']}"
            )
    else:

      assert False, "Delete object succeeded when it should have failed"
E           AssertionError: Delete object succeeded when it should have failed
E           assert False
``` 

By running the TC in a loop and counting failures vs passes, I observed that this error occurs about 20% of the time. However, when I added a 60s sleep period after the applying of the policy the fail percentage dropped to 0% (I also tried 30s and it didn't affect the result, so even if the optimum is somewhere between 30s and 60s it's good to have a buffer).

RP case for reference: https://url.corp.redhat.com/8b515be

PR #9447, opened 2024-03-10
Add hardcoded sleep after applying the policy in `test_bucket_policy.py::test_object_actions`

sagihirshfeld:
Fixes: #9446

Required backports:
- [ ] release-4.14

### DIFF
--- a/tests/functional/object/mcg/test_bucket_policy.py
+++ b/tests/functional/object/mcg/test_bucket_policy.py
@@ -1,4 +1,5 @@
 import logging
+import time
 
 import pytest
 import botocore.exceptions as boto3exception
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_bucket_policy.py
```diff
@@ -271,6 +272,12 @@ def test_object_actions(self, mcg_obj, bucket_factory):
         put_policy = put_bucket_policy(mcg_obj, obc_obj.bucket_name, bucket_policy)
         logger.info(f"Put bucket policy response from Admin: {put_policy}")
 
+        # Hardcoded sleep is needed because we lack a confirmation mechanism
+        # we could wait for - even the get-policy result has been observed to be
+        # unreliable in confirming whether they policy is actually taking effect
+        logger.info("Waiting for 60 seconds for the policy to take effect")
+        time.sleep(60)
```
udaysk23, line 279: Instead of waiting for 60 secs directly, can we break it to 12 retries. So that we don't need to wait 60 secs all the time. WDYT @sagihirshfeld

sagihirshfeld: Usually I prefer using retry where we can, but in this case if the policy fails to deny the deletion in time then the second retry attempt would fail regardless because the object would be missing.

We could modify the deletion part of the test so it generates and puts a new object before each attempt, but it's a bit of an over-complication of a simple test case in my opinion.

udaysk23 (approved):

keesturam (approved):

PR #9455, opened 2024-03-11
Create a new sanity helper class "SanityProviderMode" for the Provider Mode platform

yitzhak12:
The PR implements a new sanity helper class `SanityProviderMode` for the Provider Mode platform, that will extend the existing class `Sanity` with the following functionalities:

- Create and delete resources and run IO on multiple clients(It will use the fixture `create_scale_pods_and_pvcs_using_kube_job_on_hci_clients` to create the pods and PVCs and run IO on multiple clients).
- If the test runs on the provider, it will also check the cluster health of all the clients(Unless specified otherwise in the relevant param).

### DIFF
--- a/ocs_ci/helpers/sanity_helpers.py
+++ b/ocs_ci/helpers/sanity_helpers.py
@@ -10,9 +10,17 @@ from ocs_ci.helpers import helpers
 from ocs_ci.ocs.bucket_utils import s3_delete_object, s3_get_object, s3_put_object
 from ocs_ci.helpers.pvc_ops import create_pvcs
 from ocs_ci.utility.utils import ceph_health_check
-from ocs_ci.ocs.cluster import CephCluster, CephClusterExternal, is_ms_consumer_cluster
+from ocs_ci.ocs.cluster import (
+    CephCluster,
+    CephClusterExternal,
+    is_ms_consumer_cluster,
+    is_hci_client_cluster,
+    is_hci_provider_cluster,
+    client_clusters_health_check,
+)
 from ocs_ci.utility.retry import retry
 from ocs_ci.ocs.exceptions import CommandFailed
+from ocs_ci.utility.decorators import switch_to_orig_index_at_last
 
 logger = logging.getLogger(__name__)
 
@@ -31,7 +39,8 @@ class Sanity:
         self.pod_objs = list()
         self.obc_objs = list()
         self.obj_data = ""
-        self.ceph_cluster = CephCluster()
+        if not is_hci_client_cluster():
+            self.ceph_cluster = CephCluster()
 
     def health_check(self, cluster_check=True, tries=20):
         """
@@ -47,6 +56,7 @@ class Sanity:
                 config.ENV_DATA.get("platform") == constants.FUSIONAAS_PLATFORM
                 and config.ENV_DATA["cluster_type"].lower() == "consumer"
             )
+            or is_hci_client_cluster()
         ):
             ceph_health_check(
                 namespace=config.ENV_DATA["cluster_namespace"], tries=tries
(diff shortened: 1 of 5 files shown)


yitzhak12:
The PR validation job passed successfully: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/2461/.

ebondare (dismissed):

ocs_ci/helpers/sanity_helpers.py
```diff
@@ -31,7 +39,8 @@ def __init__(self):
         self.pod_objs = list()
         self.obc_objs = list()
         self.obj_data = ""
-        self.ceph_cluster = CephCluster()
+        if not is_hci_client_cluster():
+            self.ceph_cluster = CephCluster()
```
fbalak, line 43: ceph_cluster attribute looks like an important part of Sanity class. It is used for health_check for ceph and noobaa. This could make discrepancies for some functions like create_resources and health_check.

ocs_ci/ocs/cluster.py
```diff
@@ -3188,3 +3190,81 @@ def get_mon_quorum_ranks():
...
+    res = wait_for_pods_to_be_running(timeout=300, sleep=20)
+    if not res:
+        raise ResourceWrongStatusException("Not all the pods in running state")
+
+    logger.info("Checking that the storageclient is connected")
+    sc_obj = OCP(
+        kind=constants.STORAGECLIENT, namespace=config.ENV_DATA["cluster_namespace"]
+    )
+    sc_obj.wait_for_resource(
+        resource_name=constants.DEFAULT_STORAGE_CLIENT,
```
fbalak, line 3243: This would be better to take from config. There can be storage_client_name in ENV_DATA in https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L146 so that it doesn't have to be specified in platform configs. It can be obtained for correct context with `config.cluster_ctx.ENV_DATA.get("storage_client_name")`.
In future we could want to test other storageclient names and it can be allowed with this change.

fbalak (changes requested):
Looks good but I think that missing `ceph_cluster` can cause issues with some functions that are inherited from `Sanity` class.

ocs_ci/helpers/sanity_helpers.py
yitzhak12: When trying to run this code on a client, it couldn't get the client `Cephcluster` here: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/cluster.py#L90, as it doesn't have any.

ocs_ci/ocs/cluster.py
yitzhak12: Okay, I will add it to the config.

ocs_ci/helpers/sanity_helpers.py
fbalak: Ok, there will be probably some issues with health_check and validation of some resources in some tests but it can be taken care of in a different PR when those issues are discovered. You can add a note that this is missing so it is known when debugging.
In future there can be a new cluster class similar to `CephClusterExternal` (https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/cluster.py#L2378) that can be used on client and takes care of proper context (storage cluster and health checks are accessed on provider).

fbalak (approved):

ocs_ci/helpers/sanity_helpers.py
yitzhak12: Okay, that makes sense. I create a new issue for tracking and implementing in the future: https://github.com/red-hat-storage/ocs-ci/issues/9574.

yitzhak12:
The PR validation above checks that the regular `Sanity` class still works properly.

ebondare (approved):

PR #9469, opened 2024-03-13
Test automation for Default RBD storageclass for cluster

paraggit:
Test PVC creation without mentioning storageclass name in the spec.

        Steps:
            1. Verify RBD storageclass  is set as default.
            2. Create a PVC and don't provide any storage class name in the YAML file.
            3. Verify PVC has created and it has attached to the Default RBD  SC.
            4. Create a POD and attached the  above PVC to the Pod.
            5. Start IO on verify that IO is successful on the PV.


tests/functional/storageclass/test_rbd_default_storageclass.py
```diff
@@ -0,0 +1,60 @@
...
+
+log = logging.getLogger(__name__)
+
+
+@green_squad
+class TestRBDStorageClassAsDefaultStorageClass:
+    @tier1
+    @polarion_id("OCS-5459")
+    @pytest.mark.skipif(
+        not is_rbd_default_storage_class(),
```
shyRozen, line 19: You are running a test in the marker itself. We should check if it is (BM or Vsphere) and UI. Then in the test  run the check. Beside that it fails on collectony see PR38 checks. If we really wanted to this the way is to create a new mark and in the mark you need to put a condition for (BM or Vsphere) and UI and kubeconfig defined and only if this condition work we run `is_rbd_default_storage_class` . This way in the collectonly you'll not fail.

tests/functional/storageclass/test_rbd_default_storageclass.py
```diff
@@ -0,0 +1,67 @@
...
+    baremetal_deployment_required,
+    ui_deployment_required,
+)
+
+
+log = logging.getLogger(__name__)
+
+
+@green_squad
+@baremetal_deployment_required
```
shyRozen, line 20: We need to use @vsphere_platform_required and @baremetal_deployment_required but it wouldn't work for both so I'm suggesting creating a new mark `barmetal_or_vsphere_required` in marks.py

shyRozen: Maybe this will help `@on_prem_platform_required`.
It lead to this list in the end. The question is if the feature is supported for all those platform. We can check with DEV. If not let's create are own marker for vsphere and baremetal as described above.
`ON_PREM_PLATFORMS = [
    VSPHERE_PLATFORM,
    BAREMETAL_PLATFORM,
    BAREMETALPSI_PLATFORM,
    IBM_POWER_PLATFORM,
    RHV_PLATFORM,
    HCI_BAREMETAL,
    HCI_VSPHERE,
]
`

paraggit: I have created a marker `baremetal_or_vsphere_deployment_required ` in to the tests.

ocs_ci/helpers/helpers.py
```diff
@@ -4656,6 +4656,35 @@ def get_ceph_log_level(service, subsystem):
...
+    Args:
+        custom_sc: custom storageclass name.
+
+    Returns:
+        True : If RBD is set as Default storage class for the cluster else False.
+    """
+    default_rbd_sc = (
+        constants.DEFAULT_STORAGECLASS_RBD if custom_sc is None else custom_sc
+    )
+    kubeconfig = os.getenv("KUBECONFIG")
```
shyRozen, line 4672: No need to use kubeconfig. I meant when you use functions outside tests and collectonly will run into the function and invoke it we need to make sure kubeconfig exist. otherwise we don't need it. Delete this like and the kubeconfig command line switch.

paraggit: Removed KUBECONFIG from run_cmd

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -591,3 +591,13 @@
...
 # Marks to identify the cluster type in which the test case should run
 runs_on_provider = pytest.mark.runs_on_provider
+
+baremetal_or_vsphere_deployment_required = pytest.mark.skipif(
+    config.ENV_DATA["platform"].lower() not in ["baremetal", "vsphere"],
+    reason="Test required baremetal or vsphere deployment.",
+)
+
+ui_deployment_required = pytest.mark.skipif(
+    not config.DEPLOYMENT["ui_deployment"],
```
petr-balogh, line 601: this ui_deployment flag is mainly considered for deployment part, in the jenkins job we propagate it to all relevant stages - so test_stage should probably have it as well.  The question is if we should really rely on this when this config value is mainly used for deployment - considering it's part of DEPLOYMENT section ;)  Which means if someone will run it outside our jenkins - and if they do not propagate this value the test will be skipped. Even they run locally after they deployed via UI manually and they forgot to pass this value as it's mainly meant for deployment.

shyRozen: Hi @petr-balogh . I think who ever uses the test should know how to propagate the right values as our tests are full of marks. If someone doesn't know he should learn. beside that what are our options?

shyRozen:
@paraggit Please squash your commit.

shyRozen (dismissed):

ocs_ci/framework/pytest_customization/marks.py
petr-balogh: We have discussed offline with Parag and he is trying to get from UI dev team info if there is a way to dynamically find out if cluster was deployed via UI. If there is such a way to figure it out, then this bill be the best way to check and skip this test based on it.

Even though I do the deployment via UI, let's say manually, I can still decide to uncheck the che
ckbox to not set RBD as default SC. Then this test is also not relevant I think. Current, behavior is:
Screenshot 2024-03-21 at 10 54 30

That the checkbox is selected, and we don't do anything, then it will be used. Currently we don't have an automation to change it.

So this one is a bit tricky and what I am trying to say is that the config.DEPLOYMENT["ui_deployment"] value is mainly meant to be used for UI deployment test which does the deployment itself and can be sure that all values are correct for the deployment part. Hence discussed with @OdedViner to add some check for this feature directly to deployment itself, when this value can be used for sure, and it's a simple check to test this value is checked by default to True.  And we can at the end of deployment do additional check to really check the annotation of the SC is really set to be default.

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -591,3 +591,13 @@
 
 # Marks to identify the cluster type in which the test case should run
 runs_on_provider = pytest.mark.runs_on_provider
+
+baremetal_or_vsphere_deployment_required = pytest.mark.skipif(
+    config.ENV_DATA["platform"].lower() not in ["baremetal", "vsphere"],
+    reason="Test required baremetal or vsphere deployment.",
+)
```
ebenahar, line 598: @paraggit why can't we utilize 'on_prem_platform_required`?

ocs_ci/ocs/constants.py
```diff
@@ -1104,7 +1104,12 @@
     ROSA_PLATFORM,
     FUSIONAAS_PLATFORM,
 ]
-BAREMETAL_PLATFORMS = [BAREMETAL_PLATFORM, BAREMETALPSI_PLATFORM]
+BAREMETAL_PLATFORMS = [
+    BAREMETAL_PLATFORM,
+    BAREMETALPSI_PLATFORM,
+    HCI_BAREMETAL,
+    IBM_POWER_PLATFORM,
```
paraggit, line 1111: @suchita-g , @nehaberry   I wanted to inform you about these modifications. Please let me know if they are causing any trouble with your existing tests.

jilju: I could find only one usage which is https://github.com/paraggit/ocs-ci/blob/abb6e6792c0e6516baabd64726d58ddadc6a7c11/ocs_ci/deployment/factory.py#L71
The deployment automation on the platform which is mentioned as HCI_BAREMETAL is not completed. May be the the use of HCI_BAREMETAL is not causing a problem but I doubt if that is required in the list seeing the usage scenarios.

shyRozen (approved):

ocs_ci/helpers/helpers.py
```diff
@@ -4680,3 +4680,33 @@ def flatten_multilevel_dict(d):
...
+
+def is_rbd_default_storage_class(custom_sc=None):
+    """
+    Check if RDB is a default storageclass for the cluster
+
+    Args:
+        custom_sc: custom storageclass name.
+
+    Returns:
+        True : If RBD is set as Default storage class for the cluster else False.
```
ebenahar, line 4693:
```suggestion
        bool: True if RBD is set as the default storage class for the cluster, False otherwise.
```

ebenahar (approved):

shyRozen (approved):

ocs_ci/helpers/helpers.py
paraggit: Done

tests/functional/storageclass/test_rbd_default_storageclass.py
```diff
@@ -0,0 +1,72 @@
...
+
+        Steps:
+            1. Verify RBD storageclass  is set as default.
+            2. Create a PVC and don't provide any storage class name in the YAML file.
+            3. Verify PVC has created and it has attached to the Default RBD  SC.
+            4. Create a POD and attached the  above PVC to the Pod.
+            5. Start IO on verify that IO is successful on the PV.
+        """
+        if not is_ui_deployment():
+            pytest.skip("cluster is not deployed from UI. Skipping test.")
```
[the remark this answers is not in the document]
paraggit: `ui_deployment_required is an early quick check to verify the UI deployment via config, and is_ui_deployment thoroughly checks the UI deployment with multiple checks. Since we can't use the latter in the marker, I have put it into the test.

PR #9490, opened 2024-03-15
Test to automate validation of performance profiles and change the performance profiles post deployment

pintojoy:
This automates test cases from [RHSTOR-4547](https://issues.redhat.com//browse/RHSTOR-4547)

### DIFF
--- a/tests/functional/z_cluster/test_performance_profile_validation.py
+++ b/tests/functional/z_cluster/test_performance_profile_validation.py
@@ -131,28 +131,12 @@ class TestProfileDefaultValuesCheck(ManageTest):
             log.info("Performance profile is same as profile that is already present")
         else:
             ptch = f'{{"spec": {{"resourceProfile":"{self.perf_profile}"}}}}'
-            """
-            ptch_cmd = (
-                f"oc patch storagecluster/{storage_cluster.data.get('metadata').get('name')} "
-                f"-n openshift-storage  --type merge --patch {ptch}"
-            )
-            """
             ptch_cmd = (
                 f"oc patch storagecluster {storage_cluster.data.get('metadata').get('name')} "
                 f"-n {namespace}  --type merge --patch '{ptch}'"
             )
             run_cmd(ptch_cmd)
             log.info("Verify storagecluster on Ready state")
-            """
-            sample = TimeoutSampler(
-                timeout=2400,
-                sleep=40,
-                func=verify_storage_cluster,
-            )
-            if not sample.wait_for_func_status(result=True):
-                log.error("Storage Cluster did not reach Ready state after 2400s")
-                raise TimeoutExpiredError
-            """
 
             verify_storage_cluster()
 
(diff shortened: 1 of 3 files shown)


ramkiperiy:
Is this a common test to validate any of the profile change?, from TC name it looks like its common test or is it specific to one particular profile.

pintojoy:
> @pintojoy Is this a common test to validate any of the profile change?, from TC name it looks like its common test or is it specific to one particular profile.

@ramkiperiy Its a common test.. if the selected profile is differnet than existing profile, profile change takes place

tests/functional/z_cluster/test_performance_profile_validation.py
```diff
@@ -0,0 +1,159 @@
...
+
+log = logging.getLogger(__name__)
+
+
+@brown_squad
+@tier4a
+@skipif_external_mode
+@skipif_managed_service
+@skipif_ocs_version("<4.15")
+@pytest.mark.polarion_id("XXXX")
```
ramkiperiy, line 34: Planing to create TC in polrion or waiting for polarion data move?

pintojoy:
Both the test are verified via PR validation jobs. Adding verified marker

tests/functional/z_cluster/test_performance_profile_validation.py
pintojoy: Yes @ramkiperiy

ramkiperiy:
@pintojoy you need to resolve python 3.9 related checks, for the PR to get merged

tests/functional/z_cluster/test_performance_profile_validation.py
ramkiperiy: Do add the polarion ID, its generated and updated in test plan.

tests/functional/z_cluster/test_performance_profile_validation.py
```diff
@@ -0,0 +1,159 @@
...
+                ), f"Resource values arent reflecting actual values for {label} pod "
+        log.info("All the memory and CPU values are matching in the profile")
+
+    @pytest.mark.parametrize(
+        argnames=["perf_profile"],
+        argvalues=[
+            pytest.param(*["performance"], marks=pytest.mark.polarion_id("OCS-XXXX")),
+            pytest.param(*["lean"], marks=pytest.mark.polarion_id("OCS-XXXX")),
+            pytest.param(*["balanced"], marks=pytest.mark.polarion_id("OCS-XXXX")),
+        ],
```
ramkiperiy, line 112: Add polarion ID for the tests

ypersky1980 (approved):
LGTM

ramkiperiy (approved):

tests/functional/z_cluster/test_performance_profile_validation.py
```diff
@@ -0,0 +1,162 @@
...
+        else:
+            log.error("Does not match any performance profiles")
+
+        log.info(performance_profile)
+
+        label_selector = list(expected_cpu_limit_values.keys())
+
+        for label in label_selector:
+            for pod in get_pods_having_label(
+                label=label, namespace=config.ENV_DATA["cluster_namespace"]
```
yitzhak12, line 89: The logic does not seem to work correctly. The `label_selector` is the pod values: 'mgr', mon', 'osd', etc.
But the `label` values are different, as you can see here: https://github.com/pintojoy/ocs-ci/blob/ca45897ca2d0e41f1f51347950d45951fea79caf/ocs_ci/ocs/constants.py#L381. For example: `app=rook-ceph-mon`.

yitzhak12: Actually, I checked it in my local, and this will also work.

yitzhak12 (approved):

PR #9540, opened 2024-03-21
Fixes bi-directional bucket replication tests issue

### LINKED ISSUE #9131: Bidirectional bucket replication tests require different prefixes between the source and the destination
Due to the recent changes that were added in https://github.com/noobaa/noobaa-core/pull/7524, bidirectional bucket replication is no longer supported when the two policies use the same prefix for both the source and the destination. That includes when the two policies are using an empty prefix - as is the case in our OCS-CI tests.

Additional reference: https://url.corp.redhat.com/4b2558b

udaysk23:
Fixes #9131

### DIFF
--- a/tests/functional/object/mcg/test_bucket_replication.py
+++ b/tests/functional/object/mcg/test_bucket_replication.py
@@ -276,32 +276,41 @@ class TestReplication(MCGTest):
         """
 
         first_bucket_name = bucket_factory(bucketclass=first_bucketclass)[0].name
-        replication_policy = ("basic-replication-rule", first_bucket_name, None)
-        second_bucket_name = bucket_factory(
-            1, bucketclass=second_bucketclass, replication_policy=replication_policy
-        )[0].name
+        second_bucket_name = bucket_factory(bucketclass=second_bucketclass)[0].name
+        prefix_site1 = "site1"
+        prefix_site2 = "site2"
 
         patch_replication_policy_to_bucket(
-            first_bucket_name, "basic-replication-rule-2", second_bucket_name
+            second_bucket_name,
+            "basic-replication-rule-1",
+            first_bucket_name,
+            prefix_site2 + "/",
         )
 
-        standard_test_obj_list = awscli_pod_session.exec_cmd_on_pod(
-            f"ls -A1 {AWSCLI_TEST_OBJ_DIR}"
-        ).split(" ")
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_bucket_replication.py
```diff
@@ -276,32 +276,41 @@ def test_bidirectional_bucket_replication(
...
+            test_directory_setup.origin_dir,
+            amount=10,
+            prefix=prefix_site1,
+            mcg_obj=mcg_obj_session,
         )
-        first_bucket_set = set(standard_test_obj_list)
-        assert first_bucket_set == {
+        written_objects_with_prefix = list(
+            map(lambda x: prefix_site1 + "/" + x, written_objects)
+        )
```
sagihirshfeld, line 308: The use of `map` is clever but a bit unreadable IMO. How about this for simplicity?
```suggestion
        written_objects_with_prefix = [f"{prefix_site1}/{obj}" for obj in written_objects]
```

tests/functional/object/mcg/test_bucket_replication.py
```diff
@@ -310,10 +319,14 @@ def test_bidirectional_bucket_replication(
...
             test_directory_setup.origin_dir,
-            amount=5,
+            amount=10,
+            prefix=prefix_site2,
             mcg_obj=mcg_obj_session,
         )
-        second_bucket_set = set(written_objects)
+        written_objects_with_prefix = list(
+            map(lambda x: prefix_site2 + "/" + x, written_objects)
+        )
```
sagihirshfeld, line 328:
```suggestion
        written_objects_with_prefix = [f"{prefix_site2}/{obj}" for obj in written_objects]
```

sagihirshfeld:
Also please set the `Verified` label since the PR passed the verification - it helps bring attention for reviews and merging.
Other than that and the two semantic suggestions above - LGTM 👍 

Once you decide to reject or accept the suggestions I'll approve the PR

tests/functional/object/mcg/test_bucket_replication.py
udaysk23: Yes we can use both logic. I will go ahead and add for loop for simplicity

sagihirshfeld (approved):

tiffanyn108 (approved):

PR #9547, opened 2024-03-22
Global and custom timeout for space reclaim job

keemano:
This test adds the following
    1) Create a global config map with timeout value set, which will be used while creating reclaim space job.
    2) Adding custom timeout value for reclaim space job yaml and reclaim space cron job yaml.

### DIFF
--- a/ocs_ci/ocs/resources/pvc.py
+++ b/ocs_ci/ocs/resources/pvc.py
@@ -275,6 +275,9 @@ class PVC(OCS):
         """
         Create ReclaimSpaceJob to invoke reclaim space operation on RBD volume
 
+        Args:
+            global_timeout(bool): True, if global config map is going to be used, None otherwise
+
         Returns:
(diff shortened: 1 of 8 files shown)


ocs_ci/ocs/resources/pvc.py
```diff
@@ -271,7 +271,7 @@ def create_snapshot(self, snapshot_name=None, wait=False):
         snapshot_obj.parent_volume_mode = self.get()["spec"]["volumeMode"]
         return snapshot_obj
 
-    def create_reclaim_space_job(self):
+    def create_reclaim_space_job(self, global_timeout=None):
```
ramkiperiy, line 274: Does the param description added?

ramkiperiy:
@keemano Could you run jenkins validation and add verified label to the PR

ramkiperiy (dismissed):

PR #9553, opened 2024-03-25
[Close loop BZ: 2183444] Modify the schedule to run ReclaimSpaceJob accidentally disabled CronJob

vkathole:
https://bugzilla.redhat.com/show_bug.cgi?id=2183444

### DIFF
--- a/tests/functional/pv/space_reclaim/test_rbd_reclaimspace_cronjob.py
+++ b/tests/functional/pv/space_reclaim/test_rbd_reclaimspace_cronjob.py
@@ -306,7 +306,7 @@ class TestRbdSpaceReclaim(ManageTest):
         OCP(kind=constants.PVC, namespace=pvc_obj.project.namespace).annotate(
             '"reclaimspace.csiaddons.openshift.io/schedule=@midnight"',
             pvc_obj.name,
-            overwrite=False,
+            overwrite=True,
         )
 
         log.info("Wait for 2 hrs")


tests/functional/pv/space_reclaim/test_rbd_reclaimspace_cronjob.py
```diff
@@ -257,7 +265,98 @@ def test_reclaim_space_cronjob_with_annotation(
...
+
+        log.info("add reclaimspace.csiaddons.openshift.io/schedule label to PVC ")
+        OCP(kind=constants.PVC, namespace=pvc_obj.project.namespace).annotate(
+            '"reclaimspace.csiaddons.openshift.io/schedule=@midnight"',
+            pvc_obj.name,
+            overwrite=False,
+        )
+
+        log.info("Wait for 2 hrs")
+        time.sleep(6000)
```
PrasadDesala, line 312: Do we really need to wait for 2 hrs hear? can we set the initial time to something say 3 mins and then modify to 1 mins and check for the messages?

tests/functional/pv/space_reclaim/test_rbd_reclaimspace_cronjob.py
```diff
@@ -257,7 +265,98 @@ def test_reclaim_space_cronjob_with_annotation(
...
+        assert (
+            pvc_to_chron_job_dict[chron_job_name] == chron_job_schedule
+        ), "Reclaim space cron job does not exist, or schedule is not correct"
+
+        OCP(kind=constants.PVC, namespace=pvc_obj.project.namespace).annotate(
+            '"reclaimspace.csiaddons.openshift.io/schedule=*/1 * * * *"',
+            pvc_obj.name,
+            overwrite=True,
+        )
+        expected_log = "delete and recreate reclaimspacecronjob"
```
PrasadDesala, line 340: IIUC, the fix is to improve the error message. If that is the case, is there any possibility if this can be added to some existing test instead of a new test?

tests/functional/pv/space_reclaim/test_rbd_reclaimspace_cronjob.py
```diff
@@ -257,7 +265,98 @@ def test_reclaim_space_cronjob_with_annotation(
             pvc_to_chron_job_dict[chron_job_name] == chron_job_schedule
         ), "Reclaim space cron job does not exist, or schedule is not correct"
 
-    def wait_for_cronjobs(self, cronjobs_exist, timeout=60):
+    @tier1
```
PrasadDesala, line 268: I see you are waiting for 2 hrs in the test. If it is a must, then this should not be placed under tier1. Either the wait time has to be reduced or we need to change the test tier

ypersky1980 (approved):
LGTM

PR #9561, opened 2024-03-27
Rwop high priority pod

keemano:
Test RWOP with priority classes. 
This PR covers creating priority classes, and create pods using priority classes to set priority.

### DIFF
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -254,6 +254,7 @@ def create_pod(
     scc=None,
     volumemounts=None,
     pvc_read_only_mode=None,
+    priorityClassName=None,
 ):
     """
     Create a pod
@@ -435,6 +436,9 @@ def create_pod(
             ] = subpath
         else:
             pod_data["spec"]["containers"][0]["volumeMounts"][0]["subPath"] = subpath
+
+    if priorityClassName:
+        pod_data["spec"]["priorityClassName"] = priorityClassName
 
     # overwrite used image (required for disconnected installation)
     update_container_with_mirrored_image(pod_data)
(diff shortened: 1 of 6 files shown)


tests/functional/pv/space_reclaim/test_rbd_reclaimspace_global_timeout.py
```diff
@@ -0,0 +1,218 @@
...
+            if "No such file or directory" not in str(cmdfail):
+                raise
+            log.info(f"Verified: File {file_path} deleted.")
+
+        # Wait for 15 seconds after deleting the file
+        time.sleep(15)
+
+        # Create Global config map with required timeout set in openshift-storage namespace
+        log.info("Creating global config timeout")
+        global_timeout, self.cm_obj = create_csi_addons_global_timeout_configmap()
```
shivamdurgbuns, line 126: If the function call doesnt create object, it is returning only "False", this will result in unpacking variable error and one value can be passed to two variables. Please update the function return to return two appropriate value.

ocs_ci/helpers/helpers.py
```diff
@@ -4103,6 +4110,29 @@ def create_reclaim_space_cronjob(
...
+
+    # timeout in str format
+    logging.info(f"global_timeout_configmap is {global_timeout_configmap}")
+    timeout = global_timeout_configmap["data"]["reclaim-space-timeout"]
+    logging.info(f"timeout {timeout}")
+    cm_obj = create_resource(**global_timeout_configmap)
+    if cm_obj:
+        return timeout, cm_obj
+    else:
+        return False
```
shivamdurgbuns, line 4133: Should return two values to avoid unpacking error during function call eg. False, None

shivamdurgbuns (dismissed):

shivamdurgbuns:
@keemano  Could you please run the PR validation job as well.

ramkiperiy (approved):

PR #9583, opened 2024-03-28
Create new test file for testing resize OSD

yitzhak12:
The PR implements the basic test scenario of the resize osd feature. For more details about the feature, see https://issues.redhat.com/browse/RHSTOR-5251, and for the test details, see the test plan spreadsheet: https://docs.google.com/spreadsheets/d/1wCSwMk_L4SikzZ2UarUH1-VCXyGkHs-ZNdDEtWPG7f4/edit#gid=105186681&range=B8. 
In the test, we will perform the following:

- Prepare all the relevant data before the resizing(Save the old pods, PVCs, PVs, create pods and PVCs and run IO, save the file md5sum). 
- Resize the osd by multiplying 2(This will be done by patching the storagecluster storage param).
- Check that the resources are in the expected state and that the new size is reflected in the resources and Ceph. Also, check that the md5sum of the files didn't change after the resize osd. 
- Check the cluster and Ceph health

### DIFF
--- a/ocs_ci/ocs/resources/osd_resize.py
+++ b/ocs_ci/ocs/resources/osd_resize.py
@@ -1,0 +1,306 @@
+import logging
+
+from ocs_ci.ocs.exceptions import (
+    StorageSizeNotReflectedException,
+    ResourceWrongStatusException,
+    CephHealthException,
+)
+from ocs_ci.ocs.resources.pod import (
+    get_osd_pods,
+    delete_pods,
+    wait_for_pods_to_be_in_statuses,
+    get_ceph_tools_pod,
+)
+from ocs_ci.ocs.resources.pvc import get_deviceset_pvcs, get_deviceset_pvs
+from ocs_ci.ocs.resources.pv import get_pv_size
+from ocs_ci.ocs.resources.storage_cluster import (
+    get_storage_size,
+    get_device_class,
+    verify_storage_device_class,
+    verify_device_class_in_osd_tree,
+)
+from ocs_ci.ocs.cluster import check_ceph_osd_tree, CephCluster
+from ocs_ci.utility.utils import ceph_health_check, TimeoutSampler, convert_device_size
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.ocp import OCP
+from ocs_ci.framework import config
+
+
+logger = logging.getLogger(__name__)
+
+
+def check_resources_state_post_resize_osd(old_osd_pods, old_osd_pvcs, old_osd_pvs):
+    """
+    Check that the pods, PVCs, and PVs are in the expected state post resizing the osd.
+    It will perform the following steps:
+    1. Check that the old osd pods are in a terminating state or deleted
+    2. Check that the new osd pods running, and we have exactly the same number of osd pods as the old ones.
+    3. Check that the PVCs are in a Bound state
+    4. Check that the old PVC and PV names are equal to the current PVC and PV names
+
+    Args:
+        old_osd_pods (list): The old osd pod objects before resizing the osd
+        old_osd_pvcs (list): The old osd PVC objects before resizing the osd
+        old_osd_pvs (list): The old osd PV objects before resizing the osd
+
+    Raises:
+        ResourceWrongStatusException: If the following occurs:
+            1. The OSD pods failed to reach the status Terminated or to be deleted
+            2. The old PVC and PV names are not equal to the current PVC and PV names
+
+    """
+    old_osd_pods_count = len(old_osd_pods)
+    logger.info("Wait for the OSD pods to reach the status Terminated or to be deleted")
+    old_osd_pod_names = [p.name for p in old_osd_pods]
+    res = wait_for_pods_to_be_in_statuses(
+        expected_statuses=[constants.STATUS_TERMINATING],
+        pod_names=old_osd_pod_names,
+        timeout=300,
+        sleep=20,
+    )
+    if not res:
+        raise ResourceWrongStatusException(
+            "The OSD pods failed to reach the status Terminated or to be deleted"
+        )
+
+    logger.info("Check that the new OSD pods are running")
+    ocp_pod = OCP(kind=constants.POD, namespace=config.ENV_DATA["cluster_namespace"])
+    ocp_pod.wait_for_resource(
+        condition=constants.STATUS_RUNNING,
+        selector=constants.OSD_APP_LABEL,
+        resource_count=old_osd_pods_count,
+        timeout=300,
+        sleep=20,
+    )
+
+    logger.info(
+        f"Check that the number of the new OSD pods are exactly {old_osd_pods_count}"
+    )
+    for osd_pods in TimeoutSampler(timeout=180, sleep=10, func=get_osd_pods):
+        osd_pods_count = len(osd_pods)
+        logger.info(f"number of osd pods = {osd_pods_count}")
+        if old_osd_pods_count == osd_pods_count:
+            break
+
+    logger.info("Check that the PVCs are in a Bound state")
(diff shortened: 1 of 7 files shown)


fbalak (commented):
The functionality looks good but I have 2 suggestions:

 - New `*post_resize_osd ` functions shouldn't be in cluster.py module in my opinion. It would be better to have a new module (e.g. `osd_resize.py`) that would contain those checks and resize functionality.  Some of those functions use storagecluster cr but their functionality is focused on osds and resizing.

 - It is better to raise an `Exception` in module function instead of `assert`. `Asserts` should be used in tests but functions that could be reused on multiple places should raise an `Exception` that can be handled (e.g. negative scenario where you want it to fail).

yitzhak12:
Thanks @fbalak. I will modify the code accordingly.

ocs_ci/ocs/resources/osd_resize.py
```diff
@@ -0,0 +1,306 @@
...
+        condition=constants.STATUS_RUNNING,
+        selector=constants.OSD_APP_LABEL,
+        resource_count=old_osd_pods_count,
+        timeout=300,
+        sleep=20,
+    )
+
+    logger.info(
+        f"Check that the number of the new OSD pods are exactly {old_osd_pods_count}"
+    )
```
ebenahar, line 78: can we also add a validation that the Running OSD pod names are different than old_osd_pods names?

ocs_ci/ocs/resources/osd_resize.py
```diff
@@ -0,0 +1,306 @@
...
+        "Check that the PVC and PV sizes are equal to the expected storage size"
+    )
+    current_osd_pvcs = get_deviceset_pvcs()
+    expected_storage_size_in_gb = convert_device_size(expected_storage_size, "GB")
+    pvc_sizes = [pvc.size for pvc in current_osd_pvcs]
+    logger.info(f"PVC sizes = {pvc_sizes}")
+    if not all([p_size == expected_storage_size_in_gb for p_size in pvc_sizes]):
+        raise StorageSizeNotReflectedException(
+            f"The PVC sizes are not equal to the expected storage size {expected_storage_size_in_gb}"
+        )
```
ebenahar, line 158:
```suggestion
        raise StorageSizeNotReflectedException(
            f"The PVC sizes are not equal to the expected storage size {expected_storage_size_in_gb}. "
            f"PVC sizes: {[p_size for p_size in pvc_sizes]}
        )
```

ocs_ci/ocs/resources/osd_resize.py
```diff
@@ -0,0 +1,306 @@
...
+
+    current_osd_pvs = get_deviceset_pvs()
+    pv_sizes = [get_pv_size(pv.get()) for pv in current_osd_pvs]
+    logger.info(f"PV sizes {pv_sizes}")
+    if not all([p_size == expected_storage_size_in_gb for p_size in pv_sizes]):
+        raise StorageSizeNotReflectedException(
+            f"The PV sizes are not equal to the expected storage size {expected_storage_size_in_gb}"
```
ebenahar, line 165:
```suggestion
            f"The PV sizes are not equal to the expected storage size {expected_storage_size_in_gb}. "
            f"PV sizes: {[p_size for p_size in pv_sizes]}"
```

ocs_ci/ocs/resources/osd_resize.py
```diff
@@ -0,0 +1,306 @@
...
+    logger.info("Check the Ceph state post resize OSD")
+    check_ceph_state_post_resize_osd()
+    logger.info("All the Ceph verification steps post resize osd finished successfully")
+
+
+def ceph_verification_steps_post_resize_osd(
+    old_osd_pods, old_osd_pvcs, old_osd_pvs, expected_storage_size, num_of_tries=6
+):
+    """
+    Try to execute the function 'base_ceph_verification_steps_post_resize_osd' a number of tries
```
ebenahar, line 238: why are we doing that?

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -2669,3 +2650,56 @@ def validate_serviceexport():
...
+    Resize the OSD(e.g., from 512 to 1024, 1024 to 2048, etc.)
+
+    Args:
+        new_osd_size (str): The new osd size(e.g, 512Gi, 1024Gi, 1Ti, 2Ti, etc.)
+
+    Returns:
+        bool: True in case if changes are applied. False otherwise
+
+    """
+    sc = get_storage_cluster()
```
ebenahar, line 2696: please add a check that the new requested size is greater than the existing OSD size.
And should we add a test case of tyring to resize to a smaller size?

ocs_ci/ocs/resources/osd_resize.py
yitzhak12: Yes, I can add it.

ocs_ci/ocs/resources/osd_resize.py
yitzhak12: Here, it logs the PVC sizes anyway a few lines back. Should we still add this log message?

ocs_ci/ocs/resources/osd_resize.py
yitzhak12: Here, it logs the PV sizes anyway a few lines back. Should we still add this log message?

ocs_ci/ocs/resources/osd_resize.py
yitzhak12: Yes, right. This is regarding the comment here: https://bugzilla.redhat.com/show_bug.cgi?id=2257497#c14.

> 4. If the resize did not take effect yet, restart the OSD pods again, possibly even multiple times to get the resize to take effect. (But if it's AWS and the OSDs have already been resized previously, the resize may not be possible according to initial testing.)

ocs_ci/ocs/resources/storage_cluster.py
yitzhak12: Okay, np. 
We added a negative test in the test plan: https://docs.google.com/spreadsheets/d/1wCSwMk_L4SikzZ2UarUH1-VCXyGkHs-ZNdDEtWPG7f4/edit#gid=105186681&range=B24

ocs_ci/ocs/resources/osd_resize.py
ebenahar: ack, then no need

ocs_ci/ocs/resources/osd_resize.py
ebenahar: ack, thanks for the reference

ocs_ci/ocs/resources/storage_cluster.py
ebenahar: do you plan to add it in a follow up PR?

yitzhak12: I have implemented the osd size check. 
I will add the negative test in another PR. I need to check first manually what is the behavior

ocs_ci/ocs/resources/osd_resize.py
yitzhak12: Done

ebenahar (approved):

ebenahar (approved):

fbalak (approved):

petr-balogh:
If this PR is going to be cherry-picked we need to include fix from Vijay:
https://github.com/red-hat-storage/ocs-ci/pull/9678

As this code braking the add capacity test:
```
>       device_sets_required = int(osd_size_capacity_requested / osd_size_existing)
E       TypeError: unsupported operand type(s) for /: 'str' and 'str'

ocs_ci/ocs/resources/storage_cluster.py:1470: TypeError
```
@yitzhak12  FYI

Issue #8809, opened 2023-11-08
nsfs  failed cannot access '/nsfs': No such file or directory


tiffanyn108:
Test case: test_nsfs_object_integrity_with_wrong_permissions[CLI-20Gi] failed with error:
Logs:
https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/557/16404/785603/785668/785669/log
http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/j-013vud1cs33-t3/j-013vud1cs33-t3_20231103T065608/logs/ocs-ci-logs-1698998622/tests/manage/mcg/test_nsfs.py/TestNSFSObjectIntegrity/test_nsfs_object_integrity_with_wrong_permissions-CLI-20Gi/logs

```
Message: ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc -n openshift-storage rsh noobaa-endpoint-569b69b6d8-pjfmz chmod -R 777 /nsfs.
Error is chmod: cannot access '/nsfs': No such file or directory
command terminated with exit code 1
Type: None
```

sagihirshfeld:
Recently seen again in 4.14 runs that included the fix:
- https://url.corp.redhat.com/973a589
- https://url.corp.redhat.com/749b667
- https://url.corp.redhat.com/aefe7cd

PR #9607, opened 2024-04-01
Added retry when command failing if /nsfs not existing

tiffanyn108:
Fixed https://github.com/red-hat-storage/ocs-ci/issues/8809

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5720,9 +5720,9 @@ def nsfs_bucket_factory_fixture(
         )
 
         # Apply the necessary permissions on the filesystem
-        retry(CommandFailed, tries=3, delay=15)(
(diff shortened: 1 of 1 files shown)


udaysk23 (approved):

keesturam:
@tiffanyn108 could you please add information on how adding retry will fix the issue?

tiffanyn108:
> @tiffanyn108 could you please add information on how adding retry will fix the issue?

There is no issue with the code, the only thing causing the failure is timing.  /nsfs path is created as part of creating a endpoint pod.  When the pod is in ready state, it needs some extra time to completely come up and able to access /nsfs path.  Adding the retry command  with delay of 15s each will give it enough time to execute the next command for /nsfs.

sagihirshfeld (approved):

keesturam (approved):

PR #9618, opened 2024-04-02
Add Graviton type deployment

petr-balogh:
Ticket:
https://issues.redhat.com/browse/RHSTOR-2044

### DIFF
--- a/conf/deployment/aws/ipi_3az_rhcos_3m_3w_graviton.yaml
+++ b/conf/deployment/aws/ipi_3az_rhcos_3m_3w_graviton.yaml
@@ -1,0 +1,33 @@
+---
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
+ENV_DATA:
+  platform: 'aws'
+  deployment_type: 'ipi'
+  region: 'us-east-2'
+  worker_availability_zones:
+    - 'us-east-2a'
+    - 'us-east-2b'
+    - 'us-east-2c'
+  master_availability_zones:
+    - 'us-east-2a'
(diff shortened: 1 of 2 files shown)


vavuthu (approved):

petr-balogh:
OCP Deployment passed.
But failed on ODF as ocs-registry image is not build for arm64 arch.

Requires for now also to overwrite the image to arm64 one by passing extra config like:
```yaml
ENV_DATA:
  early_testing: true
  RELEASE_IMG: quay.io/openshift-release-dev/ocp-release@sha256:0c13de057d9f75c40999778bb924f654be1d0def980acbe8a00096e6bf18cc2a
  ```
  
  Which invokes the logic of exporting this env variable:
  ```
  export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=quay.io/openshift-release-dev/ocp-release:4.0.0-0.1
  ```
  Mentioned here: https://access.redhat.com/solutions/3880221
  Which we have already implemented to test early RHCOS builds in ocs-ci. Otherwise the installer will fail in bootstrap complaining about the image is amd64 and it requires arm64 in bootstrap and other control plane nodes if we are deploying via installer from x86_64 arch on our jenkins agent. (Deployment on mac Mx chip with arm64 built installer should work without it)
  
So we can merge for now and later on do the rest of things we will need, marking as verified.

vavuthu (dismissed):

dahorak (approved):

shivamdurgbuns (approved):

vavuthu (approved):

PR #9623, opened 2024-04-03
[Release 4.14]: Cherry pick stretch cluster PR

mashetty330:
Stretch cluster got GAed in 4.13. Its test cases are automated over the releases as it needed huge effort. This cherry pick aims to backport the tier1 test cases to respective release branches. So we include these tests as part of stretch cluster z stream production job run.


petr-balogh (approved):

ocs_ci/ocs/resources/pod.py
```diff
@@ -708,13 +710,14 @@ def get_all_pods(
     return pod_objs
 
 
-def get_ceph_tools_pod(skip_creating_pod=False, namespace=None):
+def get_ceph_tools_pod(skip_creating_pod=False, wait=False, namespace=None):
```
petr-balogh, line 713: Usually when we add new argument to function it's best practice to add new params at the end, then you are more safe that you don't break the current code if there is usage of positional arguments.

I did check that it's not the case in references so we are safe, just a note for next time, so we can keep it as it is as it's cherry-pick and was already merged like that to newer versions.

petr-balogh (approved):
There are some docstring related comments but as those are cherry-picks and already in newer versions of code, we can take them in new PR going to master and backport separately. Will not hold this PR as those commits are cherry-picks.

PrasadDesala (approved):

PR #9634, opened 2024-04-04
OCP AI deployment: enhance host id to mac mapping

dahorak:
There seems to be cases, when the first interface is not used and the actual provisioning interface is the second (or later) one.
So I'm changing the logic of `get_host_id_mac_mapping` method to prepare combination of all used mac addresses to host IDs and during updating of the name and role ignoring `KeyError`s of the mac addresses not used for the mapping.


dahorak:
Verification job: https://url.corp.redhat.com/51304da

petr-balogh (approved):

amr1ta (approved):

PR #9639, opened 2024-04-05
Fix required capacity card test

ebondare:


ocs_ci/ocs/ui/views.py
```diff
@@ -1284,19 +1284,24 @@
...
     "req_capacity_dropdown_btn_two": (
-        "(//span[@class='pf-c-select__toggle-arrow'])[2]",
+        "(//span[@class='pf-c-select__toggle-arrow'])[2] | "
+        "(//span[@class='pf-v5-c-select__toggle-arrow'])[2]",
         By.XPATH,
     ),
     "req_capacity_dropdown_list_option": (
-        "//*[@class='pf-c-select__menu-item' and contains(text(), '{}')]",
+        "//*[@class='pf-c-select__menu-item' and contains(text(), '{}')] | "
+        "//*[@class='pf-v5-c-select__menu-item' and contains(text(), '{}')]",
```
petr-balogh, line 1304: If I understand correctly the logic and `|` act as `or` here means it will try first version and if not found it will use second - so it is backward compatible with older version and new version I would prioritize the new one to be defined as first. 

As master/stable where it will get propagated will be mostly used on newer versions of OCP/ODF so it should first give a try what is more probable so:
```
"//*[@class='pf-v5-c-select__menu-item' and contains(text(), '{}')]
```
and then 
```
//*[@class='pf-c-select__menu-item' and contains(text(), '{}')]
```

so if my assumption is correct, can we adjust all of those like this?

Thanks.

DanielOsypenko: hello Petr, your understanding is correct. 
If an impact exists it counts in milliseconds. 
Additionally we don't test locators, tags or html elements themselves and the only requirement to locator is that it should be unique pointing on specific desired web-element, meaning that we don't care if the first element will work or second - it will not be counted as a bug.

More impact in this locator is from using "*", this way we are parsing all the html elements existing on dom. 
I know this was my peace of code, but if it is possible I would suggest to set a type of element we are looking - "div", "li", "input", or anything else?

ebondare: It's a button now. I'll need to check a 4.15 cluster to see how it was before the latest changes.

ebondare:
Successful test run: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/36209/consoleFull

DanielOsypenko (approved):

ocs_ci/ocs/ui/views.py
ebondare: I checked it in a 4.14 cluster, it was also a button.

nehaberry (approved):

ocs_ci/ocs/ui/views.py
```diff
@@ -1284,19 +1284,23 @@
         By.XPATH,
     ),
     "req_capacity_dropdown_selected": (
-        "//div[@id='breakdown-card-title']/following-sibling::*//*[@class = 'pf-c-select__toggle-text']",
+        "//div[@id='breakdown-card-title']/following-sibling::*//*[@class = 'pf-c-select__toggle-text'] | "
```
am-agrawa, line 1287: Can we use simple locators which are shorter?

am-agrawa (approved):

petr-balogh (approved):

ocs_ci/ocs/ui/views.py
```diff
@@ -1284,19 +1284,23 @@
...
     ),
     "req_capacity_dropdown_btn_two": (
-        "(//span[@class='pf-c-select__toggle-arrow'])[2]",
+        "(//span[@class='pf-c-select__toggle-arrow'])[2] | "
+        "(//span[@class='pf-v5-c-select__toggle-arrow'])[2]",
         By.XPATH,
     ),
     "req_capacity_dropdown_list_option": (
-        "//*[@class='pf-c-select__menu-item' and contains(text(), '{}')]",
+        "//button[contains(@class, 'c-select__menu-item') and contains(text(), '{}')]",
```
DanielOsypenko, line 1303: good way to resolve issue with "pf-<number>"  class changes, even better than to use "or" in locator.
If it is possible pls apply this approach to all locators above. Will be really more elegant and hypothetically will work with Patternfly 6

PR #9655, opened 2024-04-09
Test CLI tool for disk replacement proceudre

OdedViner:
https://bugzilla.redhat.com/show_bug.cgi?id=2234479
Test path:
```
tests/functional/z_cluster/nodes/test_disk_failures.py::TestDiskFailures::test_recovery_from_volume_deletion_cli_tool
```

### DIFF
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -75,6 +75,7 @@ TEMPLATE_CNV_VM_STANDALONE_PVC_DIR = os.path.join(
     TEMPLATE_CNV_VM_WORKLOAD_DIR, "vm-standalone-pvc"
 )
 DATA_DIR = os.getenv("OCSCI_DATA_DIR") or os.path.join(TOP_DIR, "data")
+BIN_DIR = os.path.join(TOP_DIR, "bin")
 ROOK_REPO_DIR = os.path.join(DATA_DIR, "rook")
 ROOK_EXAMPLES_DIR = os.path.join(
     ROOK_REPO_DIR, "cluster", "examples", "kubernetes", "ceph"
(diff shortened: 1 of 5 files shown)


ocs_ci/ocs/constants.py
```diff
@@ -75,6 +75,7 @@
     TEMPLATE_CNV_VM_WORKLOAD_DIR, "vm-standalone-pvc"
 )
 DATA_DIR = os.getenv("OCSCI_DATA_DIR") or os.path.join(TOP_DIR, "data")
+BIN_DIR = os.path.join(TOP_DIR, "bin")
```
petr-balogh, line 78: nope, bin dir is not from TOP_DIR directory, we already have bin dir define in config, so it should be directly used from config value

ocs_ci/ocs/constants.py
```diff
@@ -1602,7 +1603,7 @@
 # Miscellaneous
 NOOBAA_OPERATOR_POD_CLI_PATH = "/usr/local/bin/noobaa-operator"
 NOOBAA_OPERATOR_LOCAL_CLI_PATH = os.path.join(DATA_DIR, "mcg-cli")
-CLI_TOOL_LOCAL_PATH = os.path.join(DATA_DIR, "odf-cli")
+CLI_TOOL_LOCAL_PATH = os.path.join(BIN_DIR, "odf-cli")
```
petr-balogh, line 1606: you are not using it anywhere in this PR right?
But I see references in ocs_ci/helpers/helpers.py . 

So please change it there, to use directly path of bin dir (config.RUN['bin_dir']) from config we talked yesterday.

As it will be in the path, I think that you will really need anywhere else the full path as you will be able to just run directly `odf-cli`

ocs_ci/ocs/constants.py
OdedViner: done

ocs_ci/utility/utils.py
```diff
@@ -496,25 +496,32 @@ def run_cmd(
...
         timeout(int): Timeout in seconds, for pexpect to wait for prompt
-
+        string_answer (bool): string answer
+        raise_exception (bool): raise excption
     Raises:
         InteractivePromptException: in case something goes wrong
 
     """
     child = pexpect.spawn(cmd)
     for prompt, answer in prompts_answers.items():
```
shivamdurgbuns, line 516: can we add a condition when there are no prompts?

OdedViner: if there is no prompt we don need to use this funcion

tests/functional/z_cluster/nodes/test_disk_failures.py
```diff
@@ -236,3 +237,22 @@ def test_recovery_from_volume_deletion(
         self.sanity_helpers.create_resources(
             pvc_factory, pod_factory, bucket_factory, rgw_bucket_factory
         )
+
+    @bugzilla("2234479")
+    @vsphere_platform_required
+    @skipif_ocs_version("<4.15")
+    @pytest.mark.polarion_id("OCS-5502")
+    @skipif_external_mode
+    def test_recovery_from_volume_deletion_cli_tool(
```
petr-balogh, line 246: Do we want to still test both scenarios, mean without CLI Tool? As the TC test_recovery_from_volume_deletion will still get triggered as well.

petr-balogh (approved):

shivamdurgbuns (approved):

yitzhak12 (approved):

PR #9664, opened 2024-04-10
Addding a new test for cross sc clone and snapshot restore - in the same pool 

ypersky1980:
As part of work on [RHSTOR-4130](https://issues.redhat.com//browse/RHSTOR-4130) in 4.16 this PR is a new test that implements the following flow: 

1. Create two storage classes in the same pool
2. Create a pvc in the first sc
3. Create a clone on the same sc
4. Create a clone on another sc
5. Create pvc's shapshot
6. Restore the snapshot on the same sc
7. Restore the snapshot on another sc

Test cases with similar flow for sc in different than pvc pool will be handled ceparately in another PR.

### DIFF
--- a/tests/functional/storageclass/test_cross_sc_clone_snap_restore.py
+++ b/tests/functional/storageclass/test_cross_sc_clone_snap_restore.py
@@ -1,0 +1,130 @@
+import pytest
+import logging
+from ocs_ci.helpers import helpers
+from ocs_ci.ocs import constants
+from ocs_ci.framework.pytest_customization.marks import green_squad
+from ocs_ci.framework.testlib import ManageTest, tier2
+from ocs_ci.ocs.resources.pod import get_fio_rw_iops
+from tests.fixtures import create_project
+from ocs_ci.ocs.resources import pvc
+
+log = logging.getLogger(__name__)
+
+
+@green_squad
+@tier2
+@pytest.mark.usefixtures(
+    create_project.__name__,
+)
+class TestCrossScCloneSnapRestore(ManageTest):
+    """
+    The test verifies creation of clones and restore of snapshots to other than pvc storage class
+    """
+
+    @pytest.mark.parametrize(
+        argnames="interface_type",
+        argvalues=[
+            pytest.param(*[constants.CEPHBLOCKPOOL]),
+            pytest.param(*[constants.CEPHFILESYSTEM]),
+        ],
+    )
+    def test_cross_class_same_pool_clone_snap_restore(
+        self,
+        interface_type,
+        storageclass_factory,
+        pvc_factory,
+        pod_factory,
+        pvc_clone_factory,
+        snapshot_factory,
+    ):
+        """
+        Test function which validates operations accross different storage classes.
+        1. Create two storage classes in the same pool
+        2. Create a pvc in the first sc
+        3. Create a clone on the same sc
+        4. Create a clone on another sc
+        5. Create pvc's shapshot
+        6. Restore the snapshot on the same sc
+        7. Restore the snapshot on another sc
+        """
+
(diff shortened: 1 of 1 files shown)


tests/functional/storageclass/test_cross_sc_clone_snap_restore.py
```diff
@@ -0,0 +1,130 @@
...
+        pvc_factory,
+        pod_factory,
+        pvc_clone_factory,
+        snapshot_factory,
+    ):
+        """
+        Test function which validates operations accross different storage classes.
+        1. Create two storage classes in the same pool
+        2. Create a pvc in the first sc
+        3. Create a clone on the same sc
```
ramkiperiy, line 44: Is this steps correct bit confusing, after creating PVC are we creating the snapshot followed by clone and then snapshot restore?

tests/functional/storageclass/test_cross_sc_clone_snap_restore.py
```diff
@@ -0,0 +1,130 @@
...
+        4. Create a clone on another sc
+        5. Create pvc's shapshot
+        6. Restore the snapshot on the same sc
+        7. Restore the snapshot on another sc
+        """
+
+        # Create a Storage Class
+        sc_obj = storageclass_factory(interface=interface_type)
```
ramkiperiy, line 52: good to follow similar naming convention. I mean could you change this to sc_obj1 which would be good since there is sc_obj2 in code. having it common might confuse later.

tests/functional/storageclass/test_cross_sc_clone_snap_restore.py
```diff
@@ -0,0 +1,130 @@
...
+
+        snap_name = f"pvc-{interface_type.lower()}-snapshot-test-cross"
+        snap_obj = snapshot_factory(pvc_obj, snap_name)
+        log.info(f"Snapshot {snap_name} successfully created")
+
+        restore_pvc_yaml = constants.CSI_RBD_PVC_RESTORE_YAML
+        if interface_type == constants.CEPHFILESYSTEM:
+            restore_pvc_yaml = constants.CSI_CEPHFS_PVC_RESTORE_YAML
+
+        restore_pvc_name = f"{pvc_obj.name}-restored-same-sc"
```
ramkiperiy, line 98: we are restoring different snap right? could you update the naming convention here too?
good to have it something like restore_pvc1_yaml and restore_pvc2_yaml or restore_pvc_yaml_1 & 2.

tests/functional/storageclass/test_cross_sc_clone_snap_restore.py
ypersky1980: Fixed. Thanks.

tests/functional/storageclass/test_cross_sc_clone_snap_restore.py
ypersky1980: Thanks for your comment. restore_pvc_yaml is the same for both restored objects. However, following your advice, I defined for more clarity restore_pvc_name1 and restore_pvc_name2, as well as restore_pvc_obj1 and restore_pvc_obj2.

tests/functional/storageclass/test_cross_sc_clone_snap_restore.py
ypersky1980: @ramkiperiy actually we create a pvc, then create a clone in the same sc and different sc, and after this we create a snpashot and restore it twice - to the same sc and to a different sc. So it seems like the most logical flow to me, also the steps are described in the test for more clarity.

ramkiperiy (approved):

pintojoy (approved):
LGTM

Issue #9669, opened 2024-04-12
vmodl.fault.ManagedObjectNotFound during vsphere destroy


petr-balogh:
```console
2024-04-11 14:41:36  12:41:27 - MainThread - ocs_ci.framework - INFO  - Switched to cluster: jnkpr9552-c1
2024-04-11 14:41:36  12:41:27 - MainThread - ocs_ci.framework.pytest_customization.reports - INFO  - Dump of the consolidated config is located here: /home/jenkins/current-cluster-dir/logs/run-1712838405-cl0-config-end.yaml
2024-04-11 14:41:36  
2024-04-11 14:41:36  
2024-04-11 14:41:36  ==================================== ERRORS ====================================
2024-04-11 14:41:36  _____________________ ERROR at teardown of test_deployment _____________________
2024-04-11 14:41:36  def cluster_teardown_finalizer():
2024-04-11 14:41:36          # If KMS is configured, clean up the backend resources
2024-04-11 14:41:36          # we are doing it before OCP cleanup
2024-04-11 14:41:36          if ocsci_config.DEPLOYMENT.get("kms_deployment"):
2024-04-11 14:41:36              try:
2024-04-11 14:41:36                  kms = KMS.get_kms_deployment()
2024-04-11 14:41:36                  kms.cleanup()
2024-04-11 14:41:36              except Exception as ex:
2024-04-11 14:41:36                  log.error(f"Failed to cleanup KMS. Exception is: {ex}")
2024-04-11 14:41:36  >       deployer.destroy_cluster(log_cli_level)
2024-04-11 14:41:36  
2024-04-11 14:41:36  tests/conftest.py:1761: 
2024-04-11 14:41:36  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
2024-04-11 14:41:36  ocs_ci/deployment/vmware.py:1406: in destroy_cluster
2024-04-11 14:41:36      self.post_destroy_checks()
2024-04-11 14:41:36  ocs_ci/deployment/vmware.py:396: in post_destroy_checks
2024-04-11 14:41:36      self.vsphere.destroy_folder(template_folder, self.cluster, self.datacenter)
2024-04-11 14:41:36  ocs_ci/utility/vsphere.py:802: in destroy_folder
[...]

PR #9670, opened 2024-04-12
Fix issue in teardown when folder vsphere doesn't exist

petr-balogh:
Fixes: #9669

### DIFF
--- a/ocs_ci/utility/vsphere.py
+++ b/ocs_ci/utility/vsphere.py
@@ -785,8 +785,19 @@ class VSPHERE(object):
 
         vms = dc.vmFolder.childEntity
         for vm in vms:
-            if vm.name == name:
-                _rc = True
+            try:
+                if vm.name == name:
+                    _rc = True
+            except vmodl.fault.ManagedObjectNotFound as ex:
+                logger.exception(
+                    "There was an exception hit while attempting to check if folder exists!"
+                )
+                if (
+                    "has already been deleted or has not been completely create"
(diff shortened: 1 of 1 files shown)


petr-balogh:
Verification destroy job:
 https://url.corp.redhat.com/2413e8e

petr-balogh:
New verification job:
https://url.corp.redhat.com/77ea28d

ocs_ci/utility/vsphere.py
```diff
@@ -785,8 +785,19 @@ def check_folder_exists(self, name, cluster, dc):
...
+                    _rc = True
+            except vmodl.fault.ManagedObjectNotFound as ex:
+                logger.exception(
+                    "There was an exception hit while attempting to check if folder exists!"
+                )
+                if (
+                    "has already been deleted or has not been completely create"
+                    not in str(ex)
+                ):
+                    raise
```
shivamdurgbuns, line 799: it would be better if we describe which exception to raise here

dahorak (approved):

ocs_ci/utility/vsphere.py
petr-balogh: Whole exception details will be described as part of looger.exception as it automatically include exception details.
```
logger.exception(
```

The `raise` is just simple re-raise of exception caught in except block, so nothing to specify here IMO.

shylesh (approved):
LGTM

PR #9676, opened 2024-04-15
Test NotPrincipal, NotAction, NotResource policies

mashetty330:


sagihirshfeld (changes requested):
1. Looks like you're covering "standard" bucket policies here. The `NotPrincipal`, `NotAction`, and `NotResource` are properties with different behaviors of their own.

For example the following NotPrincipal AWS S3 bucket policy (generated by ChatGPT) denies all access to the bucket except for the specified user:

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Deny",
            "NotPrincipal": {
                "AWS": "arn:aws:iam::ACCOUNT-ID-WITHOUT-ACCESS:root"
            },
            "Action": "s3:*",
            "Resource": "arn:aws:s3:::YOUR-BUCKET-NAME/*"
        }
    ]
}
```

To test this you would need to make adjustments to `gen_bucket_policy`, and probably change the flow of the test cases themselves. Check the AWS documentation to be sure about the expected behaviors. 

2. Please separate this to three different test functions - there's little to no shared logic between the three since each param is basically handled in its own if-else block. I think you should be able to reuse the same polarion and bz decorators on each of them.

mashetty330:
> 1. Looks like you're covering "standard" bucket policies here. The `NotPrincipal`, `NotAction`, and `NotResource` are properties with different behaviors of their own.
> 
> For example the following NotPrincipal AWS S3 bucket policy (generated by ChatGPT) denies all access to the bucket except for the specified user:
> 
> ```
> {
>     "Version": "2012-10-17",
>     "Statement": [
>         {
>             "Effect": "Deny",
>             "NotPrincipal": {
>                 "AWS": "arn:aws:iam::ACCOUNT-ID-WITHOUT-ACCESS:root"
>             },
>             "Action": "s3:*",
>             "Resource": "arn:aws:s3:::YOUR-BUCKET-NAME/*"
>         }
>     ]
> }
> ```
> 
> To test this you would need to make adjustments to `gen_bucket_policy`, and probably change the flow of the test cases themselves. Please check the AWS documentation to be sure about the expected behaviors.
> 
> 2. Please separate this to three different test functions - there's little to no shared logic between the three since each param is basically handled in its own if-else block. I think you should be able to reuse the same polarion and bz decorators on each of them.

@sagihirshfeld the point 1, is you are correct absolutely. but then my test covers the exact mentioned scenario right? For example, in case of NotPrincipal accesss should be denied to every user except the one mentioned under NotPrincipal. What is wrong there? am i missing something?

mashetty330:
closing this PR because its been part of https://issues.redhat.com/browse/RHSTOR-5151 and covered here https://github.com/red-hat-storage/ocs-ci/pull/9688

PR #9679, opened 2024-04-16
Add changes for toleration of toolbox pod

vkathole:
The process for enabling the toolbox & adding tolerations to them had changed in ODF 4.15.
https://issues.redhat.com/browse/RHSTOR-5012
ODF 4.15 onwards we need to add an entry for "toolbox" in storagecluster CR 
spec:
  placement:
    toolbox:
      tolerations:
      - effect: NoSchedule
        key: xyz
        operator: Equal
        value: "true"

### DIFF
--- a/tests/functional/z_cluster/nodes/test_non_ocs_taint_and_toleration.py
+++ b/tests/functional/z_cluster/nodes/test_non_ocs_taint_and_toleration.py
@@ -27,6 +27,7 @@ from ocs_ci.ocs.node import (
     untaint_nodes,
     get_worker_nodes,
 )
+from ocs_ci.ocs.exceptions import CommandFailed
 from ocs_ci.ocs.resources import storage_cluster
 from ocs_ci.utility import version
 from ocs_ci.framework.pytest_customization.marks import (
@@ -35,6 +36,8 @@ from ocs_ci.framework.pytest_customization.marks import (
     skipif_ocs_version,
 )
 from ocs_ci.helpers.sanity_helpers import Sanity
+from ocs_ci.utility.retry import retry
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


tests/functional/z_cluster/nodes/test_non_ocs_taint_and_toleration.py
```diff
@@ -138,19 +150,22 @@ def test_non_ocs_taint_and_tolerations(self):
...
-                '"value": "true"}]}}'
-            )
-            ocsini_obj = ocp.OCP(
-                resource_name=constants.OCSINIT,
-                namespace=config.ENV_DATA["cluster_namespace"],
-                kind=constants.OCSINITIALIZATION,
-            )
-            ocsini_obj.patch(params=param, format_type="merge")
-            logger.info(f"Successfully added toleration to {ocsini_obj.kind}")
+            if version.get_semantic_ocs_version_from_config() < version.VERSION_4_15:
```
am-agrawa, line 153: We don't need this condition as we are working with release branches. Just backport the fix till ODF 4.15. The current master will become release branch for 4.16.

tests/functional/z_cluster/nodes/test_non_ocs_taint_and_toleration.py
```diff
@@ -114,9 +123,12 @@ def test_non_ocs_taint_and_tolerations(self):
             )
         else:
             param = (
-                f'{{"spec": {{"placement": {{"all": {tolerations}, "mds": {tolerations}, '
-                f'"noobaa-core": {tolerations}, "rgw": {tolerations}}}}}}}'
+                f'"all": {tolerations}, "mds": {tolerations}, '
+                f'"noobaa-core": {tolerations}, "rgw": {tolerations}, '
             )
+            if version.get_semantic_ocs_version_from_config() >= version.VERSION_4_15:
```
am-agrawa, line 129: We don't need this condition as we are working with release branches. Just backport the fix till ODF 4.15. The current master will become release branch for 4.16.

tests/functional/z_cluster/nodes/test_non_ocs_taint_and_toleration.py
```diff
@@ -70,6 +78,7 @@ def finalizer():
 
         request.addfinalizer(finalizer)
 
+    @skipif_ocs_version(">4.15")
```
am-agrawa, line 81: We shouldn't skip the test for lower versions IMO.

am-agrawa (changes requested):

tests/functional/z_cluster/nodes/test_non_ocs_taint_and_toleration.py
am-agrawa: Then the approach should be to make a branch out of release-4.15 and not from master, and then push the changes.

PR #9695, opened 2024-04-17
Add additional checks in the resize osd test

yitzhak12:
Check the resize osd pre-conditions:
    1. Check that the current storage size is less than the osd max size(See the comment [here](https://docs.google.com/spreadsheets/d/1wCSwMk_L4SikzZ2UarUH1-VCXyGkHs-ZNdDEtWPG7f4/edit#gid=105186681&range=B20)).
    2. If we use AWS, check that the osd resize count is no more than the AWS max resize count(Currently, there are no restrictions for the users, but for testing, it will be good to have a restriction so we don't waste resources).

If the conditions are not met, the test will be skipped.


suchita-g (dismissed):
LGTM

pintojoy (dismissed):
LGTM

tests/functional/z_cluster/cluster_expansion/test_resize_osd.py
```diff
@@ -78,6 +115,32 @@ def setup(self, create_pvcs_and_pods):
             pvc_size=pvc_size, num_of_rbd_pvc=5, num_of_cephfs_pvc=5
         )
 
+    @pytest.fixture(autouse=True)
+    def teardown(self, request):
+        """
+        Check that the new osd size has increased and increase the resize osd count
```
ebenahar, line 121: Logically, this validation should be considered a test step and not teardown

tests/functional/z_cluster/cluster_expansion/test_resize_osd.py
```diff
@@ -32,11 +37,42 @@
...
 from ocs_ci.ocs.resources.storage_cluster import resize_osd, get_storage_size
 from ocs_ci.helpers.sanity_helpers import Sanity
+from ocs_ci.utility.utils import convert_device_size
+from ocs_ci.framework import config
 
 
 logger = logging.getLogger(__name__)
 
 
+def check_resize_osd_pre_conditions():
```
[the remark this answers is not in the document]
yitzhak12: I will move it to the `osd_resize` file: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/resources/osd_resize.py.

tests/functional/z_cluster/cluster_expansion/test_resize_osd.py
yitzhak12: This validation should also be performed if the test fails in one of the steps so we have the information for the next test. Maybe I can move it to a function in the `osd_resize` file.

ebenahar: Agreed. Shouldn't we call it from the test as well?

yitzhak12: I don't think so, because then we call it twice. This check is more important for the following test cases.
If you are worried about the storage size check, it's already implemented here: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/resources/osd_resize.py#L154.

ebenahar (approved):

OdedViner (approved):

PR #9716, opened 2024-04-23
Automate bug 2219536: Check if CSI holder pods are updated after ODF upgrade

mashetty330:


mashetty330:
This PR becomes invalid as there is design change https://issues.redhat.com/browse/RHSTOR-5327

PR #9720, opened 2024-04-24
Test noobaa db activitylogs table for unnecessary logs when in default log level

mashetty330:
This PR automates
https://bugzilla.redhat.com/show_bug.cgi?id=2264480
https://bugzilla.redhat.com/show_bug.cgi?id=2259189

### DIFF
--- a/tests/functional/object/mcg/test_write_to_bucket.py
+++ b/tests/functional/object/mcg/test_write_to_bucket.py
@@ -460,6 +460,7 @@ class TestBucketIO(MCGTest):
             "Put object operation is preserving ContentEncoding as a object metadata"
         )
 
+    @tier2
     @bugzilla("2259189")
     @bugzilla("2264480")
     @pytest.mark.polarion_id("OCS-5773")
--- a/tests/functional/object/mcg/test_write_to_bucket.py
+++ b/tests/functional/object/mcg/test_write_to_bucket.py
@@ -20,6 +20,7 @@ from ocs_ci.framework.testlib import (
     acceptance,
     performance,
 )
+from ocs_ci.utility.utils import exec_nb_db_query
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.bucket_utils import (
     sync_object_directory,
@@ -27,7 +28,9 @@ from ocs_ci.ocs.bucket_utils import (
     craft_s3_command,
     s3_put_object,
     s3_head_object,
+    rm_object_recursive,
 )
+
 from ocs_ci.framework.pytest_customization.marks import (
     skipif_managed_service,
     bugzilla,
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_write_to_bucket.py
```diff
@@ -456,3 +459,42 @@ def test_content_encoding_with_write(
         logger.info(
             "Put object operation is preserving ContentEncoding as a object metadata"
         )
+
```
PrasadDesala, line 462: tier marker missing

tests/functional/object/mcg/test_write_to_bucket.py
```diff
@@ -456,3 +459,42 @@ def test_content_encoding_with_write(
...
+        """
+
+        bucket = bucket_factory()[0]
+        full_objectpath = f"s3://{bucket.name}"
+        logger.info("successfully created bucket")
+
+        sync_object_directory(
+            awscli_pod_session, AWSCLI_TEST_OBJ_DIR, full_objectpath, mcg_obj
+        )
+        logger.info("uploaded objects to the bucket")
```
PrasadDesala, line 484: how many objects will it upload?

tests/functional/object/mcg/test_write_to_bucket.py
```diff
@@ -456,3 +459,42 @@ def test_content_encoding_with_write(
...
+
+        assert "obj.uploaded" not in str(
+            nb_activitylogs
+        ), "Object upload event is being logged in activitylogs table"
+        assert "obj.deleted" not in str(
+            nb_activitylogs
+        ), "Object deletion event is being logged in activitylogs table"
+        logger.info(
+            "No object upload/deletion info is being updated in the activitylogs table"
+        )
```
PrasadDesala, line 500: there is no need to check for the DB activitylogs size?

tiffanyn108:
I would suggest to upload and delete a large number of objects and check the activitylogs size after the operation to ensure the activitylogs size doesn't grow intensively.

mashetty330:
> I would suggest to upload and delete a large number of objects and check the activitylogs size after the operation to ensure the activitylogs size doesn't grow intensively.

I dont think this is needed because the aim of the fix was to eliminate the activity logs pertaining to object upload, delete etc. It doesn't matter how many number of objects you upload.

tests/functional/object/mcg/test_write_to_bucket.py
mashetty330: Nope. we only need to make sure that the activity logs doesn't include the updates for simplest operations like object upload/delete when the log level is at default mode. size of the logs can be subjective and there is no parameter to benchmark it.

tests/functional/object/mcg/test_write_to_bucket.py
mashetty330: actually im not sure. its everything that is present in the /test_objects directory

tests/functional/object/mcg/test_write_to_bucket.py
```diff
@@ -456,3 +459,43 @@ def test_content_encoding_with_write(
...
+    @tier2
+    @bugzilla("2259189")
+    @bugzilla("2264480")
+    @pytest.mark.polarion_id("OCS-5773")
+    def test_nb_db_activity_logs_on_io(
+        self, bucket_factory, awscli_pod_session, mcg_obj
+    ):
+
+        """
+        This test checks if the activity logs are being logged
```
keesturam, line 472: Could you please add more detail to the description. i.e., for default log level, we don't expect to log creation/deletion operations.

tests/functional/object/mcg/test_write_to_bucket.py
```diff
@@ -456,3 +459,43 @@ def test_content_encoding_with_write(
...
+
+        bucket = bucket_factory()[0]
+        full_objectpath = f"s3://{bucket.name}"
+        logger.info("successfully created bucket")
+
+        sync_object_directory(
+            awscli_pod_session, AWSCLI_TEST_OBJ_DIR, full_objectpath, mcg_obj
+        )
+        logger.info("uploaded objects to the bucket")
+
```
sagihirshfeld, line 486: 1. Can we expect the activity logs to be immediately available after the I.O operations? The test might pass when it shouldn't due to a race condition, if the logs take a bit to get written in the DB. If that might be the case, please add a sleep period in between.

2. Other existing or potential future tests might do I.O with elevated noobaa logs. If such a test will run before this one, the test might fail when it shouldn't due to the old logs. You can prevent this by filtering out logs that don't mention the name of the bucket that's used here:
```
nb_activitylogs = [log for log in nb_activitylogs if bucket.name in log]
```

sagihirshfeld: Thanks for addressing my comments in https://github.com/red-hat-storage/ocs-ci/pull/9720/commits/d252c8df12b5dd6d77acf28d63253e92a1905fa3

sagihirshfeld:
Please add a validation run on for the new test

keesturam (approved):
LGTM

PrasadDesala (approved):

PR #9731, opened 2024-04-25
Added support for Deployment with Azure KMS

paraggit:
Added support for Deployment with Azure KMS.


ocs_ci/utility/kms.py
```diff
@@ -1797,7 +1802,284 @@ def cleanup(self):
...
+            raise ValueError(
+                f"ConfigMap {csi_kms_configmap.resource_name} Not found in the namespace {self.namespace}"
+            )
+
+        if self.azure_kms_connection_name not in csi_kms_configmap.data["data"]:
+            raise ValueError(
+                f"Azure Key vault connection {self.azure_kms_connection_name} not exists."
+            )
+
+        return True
```
ebenahar, line 1876: the function would either raise an exception or return True. It can never return False.
In order to keep the consistency of returning a bool value, suggesting to catch the exception and return False in case Azure KV connection is not found. And throw an exception from outside the function if needed.

ocs_ci/utility/kms.py
```diff
@@ -1797,7 +1802,284 @@ def cleanup(self):
...
+
+    def create_azure_kv_csi_kms_connection_details(self):
+        """
+        Create Azure specific csi-kms-connection-details
+        configmap resource
+        """
+
+        # Check is already configmap exists
+        csi_kms_configmap = ocp.OCP(
+            kind="ConfigMap",
```
ebenahar, line 1902: use https://github.com/red-hat-storage/ocs-ci/blob/d48844cabb7e3bbf41b3905617ecab49f80f77cc/ocs_ci/ocs/constants.py#L155

Please apply to other usages as well

ocs_ci/utility/kms.py
```diff
@@ -1797,7 +1802,284 @@ def cleanup(self):
...
+
+        # verifying ConfigMap is created or not.
+        try:
+            self.is_azure_kv_connection_exists()
+            return True
+        except ValueError as ex:
+            logger.info(
+                f"Error Creating resource {constants.AZURE_KV_CSI_CONNECTION_DETAILS} : {ex}"
+            )
+            return False
```
ebenahar, line 1947: as commented above, self.is_azure_kv_connection_exists() should return True or False and not throw an exception. With such a change, you will not need to try and except here

ocs_ci/utility/kms.py
paraggit: Done

ocs_ci/utility/kms.py
paraggit: Done

ebenahar (approved):

PR #9733, opened 2024-04-26
Use external mcg endpoint in bucket utils

fbalak:
Some tests including acceptance fail in provider client test runs executed from client due to use of internal endpoint.


ocs_ci/ocs/bucket_utils.py
```diff
@@ -55,13 +55,18 @@ def craft_s3_command(cmd, mcg_obj=None, api=False, signed_request_creds=None):
             region = f"AWS_DEFAULT_REGION={mcg_obj.region} "
         else:
             region = ""
+        endpoint = (
+            mcg_obj.s3_endpoint
+            if config.ENV_DATA["platform"] in constants.HCI_PC_OR_MS_PLATFORM
+            else mcg_obj.s3_internal_endpoint
+        )
```
udaysk23, line 62: Since This code is repeated in multiple places, I would suggest to create one separate method for capturing endpoint and call that method wherever required.

sagihirshfeld: I agree, please capture this under a method in the MCG class - something like `determine_endpoint`. We could leverage it in a potential future refactor where we only use `MCG::s3_endpoint` as a property.

udaysk23 (approved):

sagihirshfeld (approved):

ocs_ci/ocs/resources/mcg.py
```diff
@@ -379,6 +381,24 @@ def _retrieve_reduction_data():
...
+            string: S3 endpoint URI
+        """
+        return (
+            self.s3_internal_endpoint
+            if (
+                config.DEPLOYMENT.get("disconnected")
+                or config.DEPLOYMENT.get("proxy")
+                or config.ENV_DATA.get("private_link")
+            )
+            else self.s3_external_endpoint
```
udaysk23, line 399: @fbalak Can we just reverse the condition here. Something like
```
return (
            self.s3_external_endpoint
            if (
                config.DEPLOYMENT.get("Provider_client")
            )
            else self.s3_external_endpoint
            ```

**Comment by petr-balogh** [comment]
Please rebase.
See email from Coady: OCS-CI updated python version from Jun 29

Otherwise we will not get pr_check for 3.10

**Review by keesturam — DISMISSED** [review]

**Comment by sagihirshfeld** [comment]
@fbalak please rebase for the 3.10 pr_checks to pass

File: ocs_ci/ocs/resources/mcg.py [hunk 3]
```diff
@@ -379,6 +381,23 @@ def _retrieve_reduction_data():
...
+        """
+        return (
+            self.s3_external_endpoint
+            if (
+                config.ENV_DATA.get("platform", "").lower()
+                in constants.HCI_PROVIDER_CLIENT_PLATFORMS
+            )
+            else self.s3_internal_endpoint
+        )
+
```
sagihirshfeld, line 400: 1. Please update the Docstring
2. https://github.com/red-hat-storage/ocs-ci/pull/9733/commits/0faef55565b27346762f932b4c690a3169647ab9 changes `the self.s3_endpoint` class member to be the internal endpoint on most cases, which is a big change that we need to verify. 

While this endpoint should work on S3 operations that are made via the awscli pod which is within the cluster, some test cases are making s3 operations from the Jenkins agent via boto3 from outside of it. I fear this change will fail those kind of test cases.

PR #9742, opened 2024-04-30
WIP: Add option to add additional markexpression

petr-balogh:


shivamdurgbuns (approved):
Nice approach.

petr-balogh:
Verification job:
https://url.corp.redhat.com/c2fd9a5

mashetty330:
@petr-balogh can you briefly explain why do we need this approach?

tests/conftest.py
```diff
@@ -334,6 +335,10 @@ def pytest_collection_modifyitems(session, config, items):
                 item.user_properties.append(("squad", squad.capitalize()))
 
     if not (teardown or deploy or (deploy and skip_ocs_deployment)):
+        if ocsci_config.RUN.get("extra_markexpr"):
```
petr-balogh, line 338: We might still have a problem with upgrade scenario when we don't want to have this tests selected.

I am still not sure we want to have this kind of logic if it will be beneficial for us or not. Still trying to go with some pros and cons.

PR #9773, opened 2024-05-07
Run pvc resize and clone acceptance test on all clients

ebondare:


DanielOsypenko (approved):

tests/cross_functional/system_test/multicluster/test_acceptance.py
```diff
@@ -25,90 +26,95 @@ class TestAcceptance(ManageTest):
     @provider_client_ms_platform_required
     def test_acceptance(self, pvc_factory, pod_factory, teardown_factory):
         """
-        Acceptance test
+        Acceptance test on all clients
```
PrasadDesala, line 29: I have a question that's not specific to the changes in this PR. Is there a particular reason why this test has been added to the cross_functional/system_test directory? Looks at the test steps, I don't think it's a relevant place for it

fbalak: This have been added because general acceptance tests usually test multiple product components. It was moved here because of the test name but it is true that this test only tests PVCs so it can be renamed and moved to a directory with pvc tests.

PrasadDesala (approved):

fbalak (approved):

PR #9274, opened 2024-02-02
restructure baremetal configuration

dahorak:
Move BM configuraiton from `AUTH"[baremetal"]` and `AUTH["ipmi"]` to `ENV_DATA["baremetal"]` and `ENV_DATA["baremetal"]["servers"]` respectively.

This is just cosmetic change, but I think it will make the code and config little bit more readable and clean. (Right now for example the `AUTH["ipmi"]` section contains also information about the network configuration of the server (IP, MAC,...) and disks, so I think it make more sense to name the section differently and also have all the configuration related to the particular baremetal environment in one section (`ENV_DATA["baremetal"]`).

The configuration now looks like this:
```
ENV_DATA:
  baremetal:
    bm_private_nic: "..."
    bm_httpd_server: "..."
    bm_httpd_server_vm: "..."
    bm_path_to_upload: "..."
    bm_install_files: "..."
    bm_httpd_server_user: "..."
    bm_tftp_base_dir: "..."
    bm_tftp_dir: "..."
    bm_dnsmasq_dir: "..."
    bm_status_check: "..."
    bm_provisioner: "..."
    bm_provisioner_user: "..."
    servers:
      host001.example.com:
        mgmt_console: '...'
        mgmt_username: '...'
        mgmt_password: '...'
        role: '...'
        cluster_name: '...'
        public_mac: '...'
        private_mac: '...'
        ip: '...'
        root_disk_id: '...'
      host002.example.com:
        mgmt_console: '...'
        mgmt_username: '...'
        mgmt_password: '...'
        role: '...'
        cluster_name: '...'
        public_mac: '...'
        private_mac: '...'
        ip: '...'
        root_disk_id: '...'
      ...
```

*This PR has to be merged together with respective MR in ocs4-jenkins repo.*

PR #9777, opened 2024-05-08
Multus deployment with nmstate operator

OdedViner:


conf/deployment/baremetal/upi_1az_rhcos_multus_nvme_intel_3m_3w.yaml
```diff
@@ -13,6 +13,11 @@ ENV_DATA:
   is_multus_enabled: true
   multus_public_net_interface: 'enp1s0f1'
   multus_cluster_net_interface: 'enp1s0f1'
+  worker_node_network_configuration: [
+    ["ceph-public-net-shim-argo005","argo005.ceph.redhat.com","192.168.252.1","24","192.168.20.0/24"],
```
petr-balogh, line 17: I think we should move those names:
argo005","argo005.ceph.redhat.com"
to specific credential conf. Imagine someone else would like to run whit this BM configuration but on completely different env. So it will be confusing here and I think should be stored in specific credential conf (jenkins secret) for specific env.

@dahorak WDYT?

dahorak: @OdedViner as Petr suggested, this is specific for the _BM Octo Argo_ environment, so it should be configured in the respective "secrets" configuration file (`secrets/bm-octo-argo` in _ocs4-jenkins_ repo).

Also the configuration for BM environment looks like this (see also description for [PR#9274](https://github.com/red-hat-storage/ocs-ci/pull/9274)):
```
ENV_DATA:
  baremetal:
    ...
    servers:
      host001.example.com:
        <configuration-for-host001>
      host002.example.com:
        <configuration-for-host001>
      ...
```
which means, that each server already have its own section in the configuration file where is (beside other thinks) configured the role of the server, network configuration, etc. So I think those new parameters could be added to the configuration of each respective server, WDYT?

OdedViner: @dahorak @petr-balogh  done 
Opened a MR https://url.corp.redhat.com/e2f27e3

dahorak (dismissed):
Overall LGTM, just please describe the new parameters in [conf/README.md](https://github.com/red-hat-storage/ocs-ci/blob/master/conf/README.md#env_data):
* `node_network_configuration_policy_name`
* `node_network_configuration_policy_ip`
* `node_network_configuration_policy_prefix_length`
* `node_network_configuration_policy_destination_route`

ocs_ci/deployment/deployment.py
```diff
@@ -1012,6 +1012,49 @@ def deploy_ocs_via_operator(self, image=None):
...
         # Create Multus Networks
         if config.ENV_DATA.get("is_multus_enabled"):
+            from ocs_ci.deployment.nmstate import NMStateInstaller
+
+            logger.info("Install NMState operator and create an instance")
+            nmstate_obj = NMStateInstaller()
+            nmstate_obj.running_nmstate()
+            logger.info("Configure NodeNetworkConfigurationPolicy on all worker nodes")
+            worker_node_names = get_worker_nodes()
+            for worker_node_name in worker_node_names:
```
petr-balogh, line 1022: do we have add node operation running on BM env?  If so, we might think of configure also 4th node argo008.
And we need to then think of invoking those changes for that specific node during add node operation.

Doesn't need to be in this pr implemented for add node even though we use it, but if we are adding node we need to implement this later on I guess.

dahorak (dismissed):

OdedViner:
@petr-balogh @dahorak https://url.corp.redhat.com/c9d05af deployment pass

petr-balogh (approved):

dahorak (approved):

PR #9783, opened 2024-05-09
Fix test_nodes_restart test for timeout issue

pintojoy:
Test_nodes_restart fails at sanity_helpers.create_resources stage with error service unavailable error post node restart, Hence adding retries to fix the issue

### DIFF
--- a/tests/functional/z_cluster/nodes/test_nodes_restart.py
+++ b/tests/functional/z_cluster/nodes/test_nodes_restart.py
@@ -25,6 +25,7 @@ from ocs_ci.helpers.helpers import (
     storagecluster_independent_check,
 )
 from ocs_ci.ocs.ocp import OCP
+from ocs_ci.ocs.exceptions import ServiceUnavailable
 from ocs_ci.ocs.exceptions import CommandFailed
 from ocs_ci.utility.utils import retry
 from ocs_ci.ocs.cluster import is_vsphere_ipi_cluster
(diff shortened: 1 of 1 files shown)


pintojoy:
Jenkins job for the test_nodes_restart test case is passed here https://url.corp.redhat.com/bccffbd

prsurve (approved):

am-agrawa (approved):

tests/functional/z_cluster/nodes/test_nodes_restart.py
```diff
@@ -94,9 +95,9 @@ def test_nodes_restart(
             nodes.restart_nodes_by_stop_and_start(nodes=ocp_nodes, force=force)
 
         self.sanity_helpers.health_check()
-        self.sanity_helpers.create_resources(
-            pvc_factory, pod_factory, bucket_factory, rgw_bucket_factory
-        )
+        retry(ServiceUnavailable, tries=20, delay=30)(
+            self.sanity_helpers.create_resources
+        )(pvc_factory, pod_factory, bucket_factory, rgw_bucket_factory)
```
yitzhak12, line 100: Overall, the change looks good. But I have one concern: Do we see only the `ServiceUnavailable` exception? If there is a chance that there will also be a `CommandFailed` exception, I would also add it. 
I mean changing it to: 

> retry((ServiceUnavailable, CommandFailed), tries=20, delay=30)(
            self.sanity_helpers.create_resources
        )(pvc_factory, pod_factory, bucket_factory, rgw_bucket_factory)

pintojoy: As of now i have observed only service unavailable exception, as post disruption it takes some time for the cluster to except new requests. will add other exceptions if observed in later stage

yitzhak12: Okay, np.

PR #9797, opened 2024-05-13
Use full api path for subscription for CNV workload

prsurve:
Need for creating this PR is that when we choose cluster for DR then there are 2 diff subscription api ie `subscriptions.apps.open-cluster-management.io` and `subscriptions.operators.coreos.com` so it's better to use full api path


prsurve:
Verification job:-https://url.corp.redhat.com/pr9797

am-agrawa (approved):

PrasadDesala (approved):
This look specific to multicluster deployment where the subscription by default is using subscriptions.apps.open-cluster-management.io whereas in regular deployments it uses subscriptions.operators.coreos.com.  There is no harm in using the full api path for regular deployments as well.

PR #9813, opened 2024-05-19
Create two new test cases for resize osd and add verification for OSDs encryption

yitzhak12:
Create two new test cases: 

- Test resize osd when the cluster capacity is near full
- Test resize osd for large differences
- Add the MAX cluster size, and check that the expected size is less than the Max cluster size.
- Add verification steps for OSD encryption.

### DIFF
--- a/tests/functional/z_cluster/cluster_expansion/test_resize_osd.py
+++ b/tests/functional/z_cluster/cluster_expansion/test_resize_osd.py
@@ -20,6 +20,7 @@ from ocs_ci.framework.testlib import (
     tier1,
     tier4b,
     tier4c,
+    tier4a,
 )
 from ocs_ci.ocs.constants import VOLUME_MODE_BLOCK, OSD, ROOK_OPERATOR, MON_DAEMON
 from ocs_ci.helpers.osd_resize import (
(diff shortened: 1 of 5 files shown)


tests/functional/z_cluster/cluster_expansion/test_resize_osd.py
```diff
@@ -216,3 +226,44 @@ def test_resize_osd_with_resource_delete(self, resource_name, num_of_iterations)
...
+        logger.info(f"The current osd size is {self.old_storage_size}")
+        current_osd_size_in_gb = convert_device_size(self.old_storage_size, "GB", 1024)
+        max_osd_size_in_gb = 1024
+        if current_osd_size_in_gb > max_osd_size_in_gb:
+            pytest.skip(
+                f"The test will not run when the osd size is less than {max_osd_size_in_gb}Gi"
+            )
+
+        self.prepare_data_before_resize_osd()
+        self.new_storage_size = "4Ti"
```
OdedViner, line 267: 4Ti is constant? Do we have enough resources for it?

yitzhak12: It's one of the test cases: https://docs.google.com/spreadsheets/d/1wCSwMk_L4SikzZ2UarUH1-VCXyGkHs-ZNdDEtWPG7f4/edit#gid=105186681&range=B14. Maybe we can add a condition at setup to skip the test if we have more than 3 OSDs? What do you think should be the maximum resource limits?

am-agrawa (approved):

fbalak (approved):
I think that values for cluster capacity should be part of platform config (the default config and overridden when needed) but it can be changed in future if there will be a need for it.

dahorak (approved):

PR #9840, opened 2024-05-22
Refactor log-based replication testing

sagihirshfeld:
This PR decouples the MockupBucketLogger logic from LBR (log-based replication) tests, and introduces LbrHandlers as a way of reusing these tests for different setups: AWS LBR, Azure LBR, MCG-logs LBR and potential future setups.

This PR also allows writing random objects in these tests instead of the same test objects, and allows prefix specific operations.


sagihirshfeld:
Verification of `test_log_based_replication_with_disruptions`: https://url.corp.redhat.com/6fc5b8d

Note that the test has been passing until this part:
```
        # Remove replication policy and upload some objects to the bucket
        # make sure the replication itself doesn't take place
        source_bucket.replication_policy = ""
        update_replication_policy(source_bucket.name, None)
    
        logger.info("Uploading test objects and waiting for replication to complete")
        replication_handler.upload_random_objects_to_source(10)
       assert not replication_handler.wait_for_sync(
            timeout=300
        ), "Standard replication failed to complete even though replication policy is removed"
```

Which failed with the `AssertionError: Standard replication failed to complete even though replication policy is removed` error. This is due to [BZ 2266805](https://bugzilla.redhat.com/show_bug.cgi?id=2266805) which hasn't been fixed on 4.15 yet (running this on 4.16 was unstable due to other issues).

PR #9861, opened 2024-05-28
Delete bucket versioning

OdedViner:


ocs_ci/utility/aws.py
```diff
@@ -1994,17 +1994,30 @@ def delete_objects_in_bucket(self, bucket):
...
+
+            # Delete each version objects
+            versioning = self.s3_client.get_bucket_versioning(Bucket=bucket)
+            s3_resource = boto3.resource("s3")
+            if versioning.get("Status") == "Enabled":
+                version_objs = s3_resource.Bucket(bucket).object_versions.all()
+                for version_obj in version_objs:
+                    version_obj.delete()
+        except Exception as e:
+            logger.info(f"Cant delete objects {e}")
```
vavuthu, line 2018:
```suggestion
            logger.info(f"Failed to delete objects {e}")
```

vavuthu, line 2018: change info to warn/error. Also add Return to Args section.

is there any reason to suppress the exception because we are not dealing with failures in the function where we are calling this https://github.com/red-hat-storage/ocs-ci/pull/9861/files#diff-44938617241aefc7f12e348ee2af0cf229899d035156d229c9cf2b33f0b3b81eL2009

OdedViner: done

OdedViner: If you fail to delete all the objects in the bucket, you cannot delete the bucket. so we retun false and we dont try to delete the bucket.

vavuthu (approved):

OdedViner:
Running AWS cleanup job for s3:  https://url.corp.redhat.com/674826f

petr-balogh (approved):

ocs_ci/utility/aws.py
```diff
@@ -1994,17 +1994,30 @@ def delete_objects_in_bucket(self, bucket):
             bucket (str): Name of the bucket to delete objects
```
petr-balogh, line 1995: Returns:
    bool: True if bucket deleted, false otherwise.

PR #9144, opened 2024-01-10
WIP: Storage client deployment automation

amr1ta:
This pr is for automation of installation of ODF in provider mode and storage-client creation on the hosting cluster

PR #9865, opened 2024-05-29
Verify the owner of CSI deployments and daemonsets

jilju:
Verify the owner of CSI deployments and daemonsets. This is currently implemented for non-provider mode only. The verification function is not implemented for provider mode. It will be added separately following the PR #9144.
The OwnerReference will be different for provider mode and non-provider mode.
Verification is done to ensure that changes in owner for provider mode is not affected in internal mode.


jilju:
Verified https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/38100/

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -781,6 +781,45 @@ def ocs_install_verification(
         ), "Seems like MCG root secrets are public, please check"
         log.info("Noobaa root secrets are not public")
 
+    # Verify the owner of CSI deployments and daemonsets if not provider mode
+    storage_cluster_obj = get_storage_cluster()
```
ebondare, line 785: I think you need to add a condition that this code shouldn't run on a cluster client.

jilju: Added a new condition that is commonly used in the same function.

vavuthu (approved):

ebondare (approved):

PR #9868, opened 2024-05-29
ITR - Allow ocs-ci to run parallely

vavuthu:
This PR will allow us to run ocs-ci to run parrallely with ITR framework and it shouldn'tcause any functionality break of existing flow.
Introduces below 3 cli parameters

--disable-environment-checker: this will disable the left over checks in existing flow
--resource-checker: this will identify the leftover which was created by test cases. This is similiar to environment-checker, only difference is resource-checker will track the resources created during test case run where as environment-checker will track all resources in cluster irrespective of who created.
--kubeconfig: location of kubeconfig. 

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -157,6 +157,12 @@ to the pytest.
 * `--install-lvmo` - Deploy LVMCluster, will skip ODF deployment.
 * `--lvmo-disks` - Number of disks to add to SNO deployment.
 * `--lvmo-disks-size` - Size of disks to add to SNO deployment.
+* `--disable-environment-checker` - Disable the leftover checks in existing flow.
+* `--resource-checker` - This will identify the leftover which was created by test cases. This is
+    similar to environment-checker, only difference is resource-checker will track the resources
+    created during test case run whereas environment-checker will track all resources in
+    cluster irrespective of who created.
+* `--kubeconfig` - Location of kubeconfig.
 
 ## Examples
 
--- a/ocs_ci/framework/main.py
+++ b/ocs_ci/framework/main.py
@@ -154,6 +154,19 @@ def process_ocsci_conf(arguments):
 
     args, unknown = parser.parse_known_args(args=arguments)
     load_config(args.ocsci_conf)
+    if unknown:
+        for each_arg in unknown:
+            if each_arg.startswith("--cluster-path"):
+                if "=" in each_arg:
+                    framework.config.ENV_DATA["cluster_path"] = each_arg.split("=", 1)[
+                        1
+                    ]
+                else:
+                    cluster_path_position = unknown.index("--cluster-path")
+                    framework.config.ENV_DATA["cluster_path"] = unknown[
+                        cluster_path_position + 1
+                    ]
+                break
     ocs_version = args.ocs_version or framework.config.ENV_DATA.get("ocs_version")
     ocp_version = args.ocp_version or framework.config.ENV_DATA.get("ocp_version")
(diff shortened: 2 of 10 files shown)


docs/usage.md
```diff
@@ -157,6 +157,12 @@ to the pytest.
 * `--install-lvmo` - Deploy LVMCluster, will skip ODF deployment.
 * `--lvmo-disks` - Number of disks to add to SNO deployment.
 * `--lvmo-disks-size` - Size of disks to add to SNO deployment.
+* `--disable-environment-checker` - Disable the leftover checks in existing flow.
+* `--resource-checker` - This will identify the leftover which was created by test cases. This is
+    similar to environment-checker, only difference is resource-checker will track the resources
+    created during test case run whereas environment-checker will track all resources in
+    cluster irrespective of who created.
+* `--kubeconfig` - Location of kubeconfig.
```
petr-balogh, line 165: We should probably overwrite env variable KUBECONFIG pointing to this specific location as well like we do here:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/__init__.py#L208

But probably on the place in `ocs_ci/framework/pytest_customization/ocscilib.py`.

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -334,6 +334,27 @@ def pytest_addoption(parser):
...
+        action="store_true",
+        default=False,
+        help=(
+            "Resource checker checks for the left over resources which are added while running the test cases."
+        ),
+    )
+    parser.addoption(
+        "--kubeconfig",
+        dest="kubeconfig",
+        help=("Kubeconfig location which will be loaded as environmental variable"),
```
petr-balogh, line 356: 1. We should check on exporting evn variable to overwrite the default one.
2. check the multicluster scenario and context switching how it's handled per cluster.
3. Overwrite config value under config.RUN["kubeconfig"]
4. We also have kubeconfig_location under RUN section which looks like is a path from cluster dir - just check it's usage if it's OK and will not break anything.

docs/usage.md
vavuthu: done

ocs_ci/framework/pytest_customization/ocscilib.py
vavuthu: > 1. We should check on exporting evn variable to overwrite the default one.
> 
>     2. check the multicluster scenario and context switching how it's handled per cluster.
> 
>     3. Overwrite config value under config.RUN["kubeconfig"]
> 
>     4. We also have kubeconfig_location under RUN section which looks like is a path from cluster dir - just check it's usage if it's OK and will not break anything.

It won't break anything. Acceptance suite is clean and this PR didn't break any existing flow.

done apart from multicluster scenario. will take multicluster scenario later.

ocs_ci/framework/main.py
```diff
@@ -154,6 +154,11 @@ def process_ocsci_conf(arguments):
 
     args, unknown = parser.parse_known_args(args=arguments)
     load_config(args.ocsci_conf)
+    if unknown:
+        for each_arg in unknown:
+            if each_arg.startswith("--cluster-path="):
```
petr-balogh, line 159: what about scenario of --cluster-path /my/cluster path?

Then you will not split with = but space in line bellow, should we tackle also this scenario?

vavuthu: done in latest commit

petr-balogh (approved):

Akarsha-rai (approved):

PR #9883, opened 2024-05-31
Test Automation for validation of PV encryption with Azure KV

paraggit:
Verify Azure KV encryption operation with PV encryption.

        Steps:
            1. Set up Azure KV with the cluster.
            2. Create multiple PVCs and attach them to pods.
            3. Verify in Azure secrets that each PV has its corresponding secrets stored.
            4. Start IO from the pods on the encrypted PVCs.
            5. Verify that each PV's encrypted device is present on the node.
            6. Wait until IO is complete.
            7. Delete all the pods and PVCs that were created.
            8. Verify that after deleting the PVCs, the respective secrets are also removed from the Azure Vault.

### DIFF
--- a/ocs_ci/utility/kms.py
+++ b/ocs_ci/utility/kms.py
@@ -2070,6 +2070,9 @@ class AzureKV(KMS):
     def remove_kmsid(self):
         """
         Removing azure kmsid from the configmap `csi-kms-connection-details`.
+
+        Returns:
+            bool: True if KMS ID is successfully removed, otherwise False.
         """
         if not self.is_azure_kv_connection_exists():
             logger.info(
@@ -2083,7 +2086,7 @@ class AzureKV(KMS):
             namespace=self.namespace,
         )
 
-        if len(get_encryption_kmsid()) == 1:
+        if len(get_encryption_kmsid()) <= 1:
             # removing configmap csi-kms-connection-details.
             csi_kms_configmap.delete()
         else:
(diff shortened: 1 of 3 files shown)


ocs_ci/utility/kms.py
```diff
@@ -2059,6 +2067,44 @@ def verify_osd_keys_present_on_azure_kv(self):
...
+            )
+            return False
+
+        csi_kms_configmap = ocp.OCP(
+            kind=constants.CONFIGMAP,
+            resource_name=constants.VAULT_KMS_CSI_CONNECTION_DETAILS,
+            namespace=self.namespace,
+        )
+
+        if len(get_encryption_kmsid()) == 1:
```
ebenahar, line 2086: can there be situations there are  more than 1 KMS? How about covering the `>= 1` case?

ocs_ci/utility/kms.py
```diff
@@ -2059,6 +2067,44 @@ def verify_osd_keys_present_on_azure_kv(self):
         logger.info("All OSD keys are present in the Azure KV ")
         return True
 
+    def remove_kmsid(self):
+        """
+        Removing azure kmsid from the configmap `csi-kms-connection-details`.
```
ebenahar, line 2072: add `Returns`

ocs_ci/utility/kms.py
paraggit: If the cluster is already configured with the Vault KMS then it will have multiple KMS IDs.

ocs_ci/utility/kms.py
paraggit: Done

ebenahar (approved):

PR #9890, opened 2024-06-03
Skip the default backingstore override tests in deployments with limited connectivity

sagihirshfeld:
These tests attempt to create a backingstore of the same type as the original default noobaaa backingstore, which can result in failure when the platform is not on vSphere. 

In the case of a disconnected cluster on the AWS platform, the default backingstore is using an AWS bucket so the tests attempts to create such a backingstore and fails.


sagihirshfeld:
Adding the verified label since this PR only adds existing, verified decorators

udaysk23 (approved):

mashetty330 (approved):

petr-balogh:
Please rebase.
See email from Coady: OCS-CI updated python version from Jun 29

Otherwise we will not get pr_check for 3.10

keesturam (approved):

PR #9893, opened 2024-06-03
[RDR] [Co-situated hub] Hub recovery after site-failure and failover of DR workloads

am-agrawa:
When active hub cluster is co-situated with the primary managed cluster, during site-failure, both the clusters are expected to go down. In this case, we perform hub recovery and switch to passive hub. When the primary managed cluster is down, workloads become inaccessible and thus failover is performed on all the workloads to recover them on the surviving (secondary) managed cluster.


am-agrawa:
Opened new [PR](https://github.com/red-hat-storage/ocs-ci/pull/9912) created out of my fork. Hence closing this one.

PR #9908, opened 2024-06-06
DR must-gather modular test

OdedViner:


ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -354,6 +356,13 @@
     ),
     reason="Test runs ONLY on Fusion HCI provider and client clusters",
 )
+dr_setup_required = pytest.mark.skipif(
+    not (
+        config.default_cluster_ctx.MULTICLUSTER.get("multicluster_mode")
```
shylesh, line 361:
```suggestion
        config.MULTICLUSTER.get("multicluster_mode")
```
this should be enough

PR #9954, opened 2024-06-14
Setup host inventory for hosted clusters using agent

jilju:
Creating hosted cluster using agent have additional prerequisites. This PR creates resources for host inventory which is part of the prerequisite.
To run this , add a new parameter "hosted_cluster_platform" and value "agent" in the config.
```
DEPLOYMENT:
  hosted_cluster_platform: "agent"
```
This parameter is introduced to differentiate between kubevirt, agent and aws. Kubevirt is the current default. Aws is not supported.


ocs_ci/deployment/hosted_cluster.py
```diff
@@ -428,6 +428,82 @@ def deploy_dependencies(
...
+    template_yaml = os.path.join(
+        constants.TEMPLATE_DIR, "hosted-cluster", "agent_service_config.yaml"
+    )
+    agent_service_config_data = templating.load_yaml(template_yaml)
+    # TODO: Add custom OS image details
+    helpers.create_resource(**agent_service_config_data)
+
+    # Verify new pods that should be created
+    pod_obj = OCP(kind=constants.POD, namespace="multicluster-engine")
+    for pod_label in ["app=assisted-service", "app=assisted-image-service"]:
```
DanielOsypenko, line 464: hey, you can use a new utility for this to wait for statuses of the pods simultaneously
[wait_for_pods_to_be_in_statuses_concurrently() ](https://github.com/red-hat-storage/ocs-ci/blob/eb9a669e83c832e5d7374f98beef79758284f566/ocs_ci/ocs/resources/pod.py#L3228)

DanielOsypenko (dismissed):

ocs_ci/deployment/hosted_cluster.py
jilju: Done

ebondare (approved):

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -428,6 +428,82 @@ def deploy_dependencies(
...
+    )
+
+
+def create_host_inventory():
+    """
+    Create InfraEnv resource for host inventory
+
+    """
+    # Create new project
+    project_name = helpers.create_project(project_name="bm-agents").resource_name
```
fbalak, line 480: Shouldn't this namespace be created by operator?

fbalak (approved):

ocs_ci/deployment/hosted_cluster.py
jilju: No, this is manual step.

jilju:
This PR can be merged because this is a prerequisite for agent based hosted cluster creation. Hosted cluster creation using agent is not implemented. So this PR cannot be tested. The changes in the PR does not take effect unless the configuration hosted_cluster_platform: "agent" is used. This is used for agent only. So this will not cross path with any existing implementation.

PR #9974, opened 2024-06-23
S3 cleanup with new method

OdedViner:
The PR opened because some aws s3 cleanup jobs failed 
https://url.corp.redhat.com/0433241


OdedViner:
@petr-balogh  You were right! When I added a try/except block to bucket.objects.all().delete() and it fails, we can simply delete the entire bucket.
https://url.corp.redhat.com/22537d6

petr-balogh (approved):

sagihirshfeld (approved):

Issue #9978, opened 2024-06-24
Prometheus related tests fails because check_alert_list() function is broken


schaefm-ibm:
Many of tests related to evaluation of Prometheus alerts fail because the function `check_alert_list()` in `ocs_ci.utility.prometheus` is broken:

It initializes itself by setting the `target_alerts` array with all alerts carying the passed `label`. However, within the loop over the passed `states` it is overwriting the `target_alerts` array. As a consequence, if passing a `states` array, for example `[ 'pending', 'firing' ]`, the second pass through the loop, when `state` is `'firing'` will fail on assertion error, because after the previous pass, only the `'pending'` alerts are contained in the `target_alerts` array when it is inspected. Therefore the resulting `target_alerts` is quaranteed to be empty which is definitely wrong.

PR #9987, opened 2024-06-26
Fix alert list checking #9978

schaefm-ibm:
Checking alert list fails if filtering for more than two states. Reason was the loop on provided states filtered the examined alert list by the fist element in the states list. Therefore respective tests always fail when filtering for the second and further states.
Fixed by repeatedly filtering the original and not the filtered one.


petr-balogh:
I think this should be fixed by @fbalak here:
https://github.com/red-hat-storage/ocs-ci/pull/10018/files

Issue #10002, opened 2024-07-01
setup_mcg_bg_features() fails with a KeyError when the skip_any_provider arg is used


PrasadDesala:
```
 # skip any cloud providers if one wants to skip
        if skip_any_provider is not None:
            for provider in skip_any_provider:
                if provider not in cloud_providers.keys():
                    logger.error(
                        f"Bucket type {provider} you asked to skip is not valid type "
                        f"and valid are {list(cloud_providers.keys())}"
                    )
                else:
>                   bucket_types.pop(provider)
E                   KeyError: 'azure'
tests/cross_functional/conftest.py:1282: KeyError
```

PR #10011, opened 2024-07-01
Longevity enhancements

PrasadDesala:
This PR covers the following,
- [x] Include running MCG features in Longevity Stage 2 runs (setting up MCG features buckets, run IOs, validate IOs)
- [x] Stage2() can now be called to run mcg and non-mcg operations individually 
- [x] Fixes: #10002 
- [x] Cluster sanity checks can be collected at the start of the cycle and at the end of each iteration for stage 2, 3 and 4

### DIFF
--- a/tests/cross_functional/conftest.py
+++ b/tests/cross_functional/conftest.py
@@ -1279,7 +1279,7 @@ def setup_mcg_bg_features(
                         f"and valid are {list(cloud_providers.keys())}"
                     )
                 else:
-                    bucket_types.pop(provider)
+                    cloud_providers.pop(provider)
 
         all_buckets = list()
         feature_setup_map = dict()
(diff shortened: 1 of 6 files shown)


tests/cross_functional/conftest.py
```diff
@@ -1346,6 +1346,6 @@ def factory(
         )
         feature_setup_map["executor"]["event"] = event
         feature_setup_map["executor"]["threads"] = threads
-        return feature_setup_map
+        return feature_setup_map, all_buckets
```
mashetty330, line 1349: this change will break the existing system tests @PrasadDesala because of unpacking.

mashetty330: instead you can add all_buckets to the feature_setup_map dict with key "all_buckets"

mashetty330 (changes requested):

tests/cross_functional/conftest.py
PrasadDesala: Done

mashetty330 (approved):

PrasadDesala:
verification job - https://url.corp.redhat.com/stag2-verification

vavuthu (approved):

PR #10054, opened 2024-07-08
Adding pvc resize functionality to RWOP test

ypersky1980:
This PR adds the following flow to RWOP test : 

- after we make sure that only one pod with the same priority can be attached to RWOP PVC, we resize the PVC and try to create another pod with the same priority and attach it to the PVC, making sure the operation will still not success.

### DIFF
--- a/tests/functional/pv/pv_services/test_rwop_pvc.py
+++ b/tests/functional/pv/pv_services/test_rwop_pvc.py
@@ -59,6 +59,13 @@ class TestRwopPvc(ManageTest):
 
         # Verify that PVCs are reusable by creating new pods
         log.info(f"PVC obj {self.pvc_obj}")
+        self.create_pod_and_validate_pending(pod_factory, interface)
+
+        self.pvc_obj.resize_pvc(20, True)
+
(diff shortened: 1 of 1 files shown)


tests/functional/pv/pv_services/test_rwop_pvc.py
```diff
@@ -59,6 +59,13 @@ def test_pod_with_same_priority(self, pod_factory, interface):
 
         # Verify that PVCs are reusable by creating new pods
         log.info(f"PVC obj {self.pvc_obj}")
+        self.create_pod_and_validate_pending(pod_factory, interface)
+
+        self.pvc_obj.resize_pvc(20, True)
+
+        self.create_pod_and_validate_pending(pod_factory, interface)
+
+    def create_pod_and_validate_pending(self, pod_factory, interface):
```
jilju, line 68: Nested function is not actually a  necessity here. It will be nice to keep it outside the test function and pass the function arguments.

jilju (approved):

ramkiperiy (approved):

PR #10086, opened 2024-07-12
A new test for migration of RWO to RWOP pvc 

ypersky1980:
This test implements the following flow: 

1) Creating a RWO PVC for RWD/CephFS pvcs 
2) Implementing migration prodecure ( https://kubernetes.io/docs/tasks/administer-cluster/change-pv-access-mode-readwriteoncepod/#migrating-existing-persistentvolumes ) 
3) After PVC mode is changed from RWO to RWOP - verifying that after the first pod is attached , the second pod with the same priority is not attached to the PVC and is in status Pending.


ypersky1980:
This PR contains already merged content.

PR #10095, opened 2024-07-14
Deploy ODF4.17 via UI

OdedViner:


ebondare (approved):

DanielOsypenko (approved):

ocs_ci/ocs/ui/views.py
```diff
@@ -1858,6 +1858,55 @@
...
+        "login": {**login, **login_4_11, **login_4_14},
+        "page": {**page_nav, **page_nav_4_10, **page_nav_4_14},
+        "generic": generic_locators,
+        "add_capacity": {**add_capacity, **add_capacity_4_11, **add_capacity_4_12},
+        "deployment": {
+            **deployment,
+            **deployment_4_7,
+            **deployment_4_9,
+            **deployment_4_10,
+            **deployment_4_11,
```
petr-balogh, line 1871: Can we think of some other solution to avoid using such old versions in our code base?

Not needed for this PR but I think we should change the logic here of locators, to be able to merge some of previous and just do copy whole 4.17 from 4.16 and just update some specific values for 4.17 which are required.

petr-balogh (approved):

PR #10117, opened 2024-07-16
[4.17] Test STS support for MCG clients

mashetty330:


ocs_ci/ocs/bucket_utils.py
```diff
@@ -2739,3 +2794,89 @@ def map_objects_to_owners(mcg_obj, bucket_name, prefix=""):
...
+    signed_request_creds=None,
+):
+    """
+    Aws s3 assume role of an User
+
+    Args:
+        role_name (str): Role name of a role attached to the assumed user
+        access_key_id_assumed_user (str): Access key id of the assumed user
+        mcg_obj (MCG): MCG object
+        signed_request_creds (dict): a dictionary containing AWS S3 creds for a signed request
```
PrasadDesala, line 2814: add an empty line after args

ocs_ci/ocs/bucket_utils.py
```diff
@@ -2739,3 +2794,89 @@ def map_objects_to_owners(mcg_obj, bucket_name, prefix=""):
...
+
+def s3_create_bucket(s3_obj, bucket_name, s3_client=None):
+    """
+    AWS s3 create bucket
+
+    Args:
+        s3_obj (MCG): MCG object
+        bucket_name (str): Name of the bucket
+        s3_client (S3.Client): Any S3 client resource
+
```
PrasadDesala, line 2840: doc string for return is missing

ocs_ci/ocs/bucket_utils.py
```diff
@@ -2739,3 +2794,89 @@ def map_objects_to_owners(mcg_obj, bucket_name, prefix=""):
...
+    """
+    if s3_client:
+        return s3_client.create_bucket(Bucket=bucket_name)
```
PrasadDesala, line 2843: according to this code, s3_obj is a mandatory arg and the bucket can be created with it. If that is case, why do you need s3_client?

ocs_ci/ocs/bucket_utils.py
```diff
@@ -2739,3 +2794,89 @@ def map_objects_to_owners(mcg_obj, bucket_name, prefix=""):
...
+
+def s3_delete_bucket(s3_obj, bucket_name, s3_client=None):
+    """
+    AWS s3 delete bucket
+
+    Args:
+        s3_obj (MCG): MCG object
+        bucket_name (str): Name of the bucket
+        s3_client (S3.Client): Any s3 client resource
+
```
PrasadDesala, line 2856: doc string for return is missing

ocs_ci/ocs/bucket_utils.py
```diff
@@ -2739,3 +2794,89 @@ def map_objects_to_owners(mcg_obj, bucket_name, prefix=""):
...
+    """
+    if s3_client:
+        return s3_client.delete_bucket(Bucket=bucket_name)
```
PrasadDesala, line 2859: same as my above comment

ocs_ci/ocs/bucket_utils.py
```diff
@@ -2739,3 +2794,89 @@ def map_objects_to_owners(mcg_obj, bucket_name, prefix=""):
...
+def s3_list_buckets(s3_obj, s3_client=None):
+    """
+    AWS S3 list buckets
+
+    Args:
+        s3_obj (MCG): MCG object
+        s3_client (S3.Client): Any s3 client resource
+
+    Returns:
+        List of buckets
```
PrasadDesala, line 2873:
```suggestion
        list: List of buckets
```

ocs_ci/ocs/resources/bucket_policy.py
```diff
@@ -128,6 +133,16 @@ def __init__(
             aws_secret_access_key=self.access_key,
         )
 
+    def delete_account(self):
+        """
+        Delete the noobaa account
+
```
PrasadDesala, line 139: doc string for return is missing

tests/conftest.py
```diff
@@ -7350,6 +7350,115 @@ def finalizer():
...
+            f"Adding following new env vars to the noobaa-core sts: {added_env_vars}"
+        )
+        dep_obj.patch(
+            resource_name=constants.NOOBAA_ENDPOINT_DEPLOYMENT,
+            params=json.dumps(patch_ops),
+            format_type="json",
+        )
+
+        # # Reset the noobaa-core pod to apply the changes
+        # mcg_obj_session.reset_core_pod()
```
PrasadDesala, line 7419: you need this commented lines?

tests/functional/object/mcg/test_sts_client.py
```diff
@@ -0,0 +1,264 @@
...
+            logger.info(f"Deleting bucket {bucket.get('bucket')}")
+            s3_delete_bucket(
+                bucket.get("mcg"), bucket.get("bucket"), bucket.get("s3client")
+            )
+            logger.info(f"Deleted the bucket {bucket}")
+
+    request.addfinalizer(finalizer)
+    return factory
+
+
```
PrasadDesala, line 133: can this be run on any platforms or it is specific to only aws?

tests/functional/object/mcg/test_sts_client.py
```diff
@@ -0,0 +1,264 @@
...
+        add_env_vars_to_noobaa_endpoint_class(
+            [(constants.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS, 600000)]
+        )
+        logger.info("Changed the sts token expiration time to 10 minutes")
+
+        # hardcoded wait for 5 minute. test fails because noobaa takes time
+        # stabilise after pod restarts even though all pods are running
+        logger.info("Wait for 5 minute after patch operation")
+        time.sleep(300)
```
PrasadDesala, line 172: can we replace the hardcoded sleep time with polling the status of the resources or some other way?

tests/conftest.py
```diff
@@ -7350,6 +7350,115 @@ def finalizer():
...
+
+            # Patch the noobaa-core sts to remove the env var
+            dep_obj.patch(
+                resource_name=constants.NOOBAA_ENDPOINT_DEPLOYMENT,
+                params=json.dumps([remove_env_var_op]),
+                format_type="json",
+            )
+
+        # Reset the noobaa-core pod to apply the changes
+        # mcg_obj_session.reset_core_pod()
```
PrasadDesala, line 7456: you need this commented lines?

PrasadDesala, line 7456: Do you need this commented lines?

ocs_ci/utility/retry.py
```diff
@@ -74,3 +74,55 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return decorator
+
+
+def retry_until_exception(
```
PrasadDesala, line 79: Instead of creating a separate function can we handle this in the existing retry function? One way is to add a new arg until_exception defaulting to False and return True only if until_exception.
I am not sure if it can cause any regression to the existing function but it is something that we can try as it reduces the code redundancy.

tests/functional/object/mcg/test_sts_client.py
mashetty330: can run on any platform. Not platform specific.

ocs_ci/ocs/bucket_utils.py
```diff
@@ -88,6 +88,61 @@ def craft_s3_command(cmd, mcg_obj=None, api=False, signed_request_creds=None):
...
+        if mcg_obj.region:
+            region = f"AWS_DEFAULT_REGION={mcg_obj.region} "
+        else:
+            region = ""
+        base_command = (
+            f'sh -c "AWS_CA_BUNDLE={constants.SERVICE_CA_CRT_AWSCLI_PATH} '
+            f"AWS_ACCESS_KEY_ID={mcg_obj.access_key_id} "
+            f"AWS_SECRET_ACCESS_KEY={mcg_obj.access_key} "
+            f"{region}"
+            f"aws sts "
```
udaysk23, line 121: The only difference I can see in `craft_sts_command` and `craft_s3_command` is string `sts`. Can we pass `sts` string to `craft_s3_command` wherever needed

mashetty330: it wouldnt be s3 command if its sts. sts is completely different and would have sts specific parameters if we were to extend the functions

ocs_ci/ocs/bucket_utils.py
mashetty330: s3_obj consists of admin credentials. but if any function call passes s3_client (optional) parameter then we use that s3_client instead of s3_obj. s3_client will have credentials for any customer s3 user

ocs_ci/ocs/bucket_utils.py
mashetty330: added my reply in the previous comment

ocs_ci/utility/retry.py
mashetty330: Retry serves purpose of "retry until we dont see the given exception" and retry until exception is we want to retry until the exception is occured. Its two different purpose and managing it in single function doesnt sound good to me even though it reduces the number of lines of code.

tests/functional/object/mcg/test_sts_client.py
```diff
@@ -0,0 +1,258 @@
...
+
+        """
+        try:
+            mcg_obj_session.remove_sts_role(email)
+        except CommandFailed as e:
+            if "No such account email" not in e.args[0]:
+                raise
+
+    request.addfinalizer(teardown)
+    return factory
```
sagihirshfeld, line 59: Please relocate to `ocs-ci/tests/conftest.py` for future reuse

tests/functional/object/mcg/test_sts_client.py
```diff
@@ -0,0 +1,258 @@
...
+        logger.info(f"Assigned the assume role policy to the user {user_2}")
+
+        # noobaa admin assumes the above role
+        creds_generated = sts_assume_role(
+            awscli_pod_session,
+            role_name,
+            nb_user_access_key_id,
+            signed_request_creds=signed_request_creds,
+        )
+        mcg_obj_session.create_s3client_assumed_role(creds_generated)
```
sagihirshfeld, line 213: Please consider merging the logic of `create_s3client_assumed_role` into `sts_assume_role` to it just returns the s3 client. You can then pass that client instead of `mcg_obj_session.assumed_s3_client` for the rest of the test.

This would simplify the flow a bit, but more importantly it untangles the use of assumed roles from the MCG class. For example, you could let an unprivileged account assume the noobaa-admin's role to elevate access instead.

tests/functional/object/mcg/test_sts_client.py
```diff
@@ -0,0 +1,258 @@
...
+        try:
+            new_bucket(bucket_2, mcg_obj_session.assumed_s3_client)
+            assert (
+                False
+            ), "Bucket was created even though assumed role user doesnt have ability to create new bucket"
+        except Exception as err:
+            if "AccessDenied" not in err.args[0]:
+                raise
+            logger.info("Bucket creation failed as expected")
+        logger.info(s3_list_buckets(mcg_obj_session, mcg_obj_session.assumed_s3_client))
```
sagihirshfeld, line 227: Is this log a leftover from development? If not, please add content to the log that would clarify why we're listing the buckets when new no buckets were supposed to get created

tests/functional/object/mcg/test_sts_client.py
```diff
@@ -0,0 +1,258 @@
...
+        # change sts token expire time to 10 minute
+        add_env_vars_to_noobaa_core_class(
+            [(constants.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS, 600000)]
+        )
+        logger.info("Changed the sts token expiration time to 10 minutes")
+
+        add_env_vars_to_noobaa_endpoint_class(
+            [(constants.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS, 600000)]
+        )
+        logger.info("Changed the sts token expiration time to 10 minutes")
```
sagihirshfeld, line 166: Why do we need this step? At the final step of the test where we're waiting for the role-assumption to no longer work, it looks like we're expecting that because we removed the role, and not because the token expired

mashetty330: @sagihirshfeld once you remove the role config from a user.. it no longer allows new role assumption. But the old token would still be valid until the expiration is reached

sagihirshfeld: Ohhh ok thanks for the clarification 👍

PrasadDesala (approved):

sagihirshfeld (approved):

keesturam (approved):

PR #10123, opened 2024-07-16
Allow ssh jump host configuration for external rhcs cluster

dahorak:
This update should allow configuration of SSH Jump host for connecting to external RHCS cluster nodes (in cases, when the RHCS cluster is not directly accessible).

The configuration is done in the config `EXTERNAL_MODE` section, for example like this:
```yaml
EXTERNAL_MODE:
    ssh_jump_host:
        host: jump.host.example.com
        user: test
        private_key: /path/to/private/ssh/key.pem
```

### DIFF
--- a/ocs_ci/deployment/vmware.py
+++ b/ocs_ci/deployment/vmware.py
@@ -1444,7 +1444,13 @@ class VSPHEREUPI(VSPHEREBASE):
                 rbd_name = config.ENV_DATA.get("rbd_name") or defaults.RBD_NAME
                 # get external cluster details
                 host, user, password, ssh_key = get_external_cluster_client()
-                external_cluster = ExternalCluster(host, user, password, ssh_key)
+                external_cluster = ExternalCluster(
+                    host,
+                    user,
+                    password,
+                    ssh_key,
+                    jump_host=config.EXTERNAL_MODE.get("ssh_jump_host"),
+                )
                 external_cluster.remove_rbd_images(pvs_to_delete, rbd_name)
             except Exception as ex:
                 logger.warning(
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -637,7 +637,13 @@ def ocs_install_verification(
             object_store_user = defaults.EXTERNAL_CLUSTER_OBJECT_STORE_USER
             realm = config.ENV_DATA.get("rgw-realm")
             host, user, password, ssh_key = get_external_cluster_client()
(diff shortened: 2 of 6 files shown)


dahorak:
@Shrivaibavi would you have a chance to test this proposal?

ocs_ci/deployment/vmware.py
```diff
@@ -1386,7 +1386,13 @@ def destroy_cluster(self, log_level="DEBUG"):
...
                 # get external cluster details
                 host, user, password, ssh_key = get_external_cluster_client()
-                external_cluster = ExternalCluster(host, user, password, ssh_key)
+                external_cluster = ExternalCluster(
+                    host,
+                    user,
+                    password,
+                    ssh_key,
+                    jump_host=config.EXTERNAL_MODE.get("ssh_jump_host"),
+                )
```
Shrivaibavi, line 1395: why do we need this change for vmware and ocs_upgrade ?

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -625,7 +625,13 @@ def ocs_install_verification(
...
             realm = config.ENV_DATA.get("rgw-realm")
             host, user, password, ssh_key = get_external_cluster_client()
-            external_cluster = ExternalCluster(host, user, password, ssh_key)
+            external_cluster = ExternalCluster(
+                host,
+                user,
+                password,
+                ssh_key,
+                jump_host=config.EXTERNAL_MODE.get("ssh_jump_host"),
+            )
```
Shrivaibavi, line 634: if we are considering the changes according to IBM cloud, do we need changes according to IBM config too ?

ocs_ci/deployment/vmware.py
dahorak: I've applied it generally to all the places where we are using `ExternalCluster` class (to have this option for any environment).

ocs_ci/ocs/resources/storage_cluster.py
dahorak: This change requires following additional configuration for the environments where it will be required (probably in the external cluster config or some secret file). We don't want to add this to the platform related config in ocs-ci.

```
EXTERNAL_MODE:
  ssh_jump_host:
    host: "..."
    user: "..."
    private_key: "..."
    password: "..."
```

Shrivaibavi: Shouldn't i expect similar changes in ocs_ci/deployment/ibm.py ??

dahorak: And there is any usage of `ExternalCluster` class there?

PR #10129, opened 2024-07-17
reboot nfs_client_vm if not accessible and retry connection

dahorak:
The aim of this PR is automatic reboot of NFS Client VM in case it is not accessible.

To make this work properly, following two new configuration parameters are required:
```
ENV_DATA:
    nfs_client_vm_cloud: NAME_OF_CLOUD_DEFINED_IN_CONFIG
    nfs_client_vm_name: NFS_CLIENT_VM_NAME
```

The logic is, that it firstly try to connect to the NFS Client VM and if it fails, it reboots it, wait some time and try to connect again (with retry this time).


petr-balogh (approved):

amr1ta (approved):
Looking good

PR #10141, opened 2024-07-18
Fixed default sc tests case which was failing

paraggit:
test `test_pvc_creation_without_storageclass_name` was failing in case of custom sorageclass present in the cluster and there are multiple default RBD storage class in the cluster.

### DIFF
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -4760,7 +4760,7 @@ def flatten_multilevel_dict(d):
     return leaves_list
 
 
-def is_rbd_default_storage_class(custom_sc=None):
+def is_rbd_default_storage_class(sc_name=None):
     """
     Check if RDB is a default storageclass for the cluster
 
@@ -4770,9 +4770,7 @@ def is_rbd_default_storage_class(custom_sc=None):
     Returns:
         bool : True if RBD is set as the  Default storage class for the cluster, False otherwise.
     """
-    default_rbd_sc = (
-        constants.DEFAULT_STORAGECLASS_RBD if custom_sc is None else custom_sc
-    )
+    default_rbd_sc = constants.DEFAULT_STORAGECLASS_RBD if sc_name is None else sc_name
     cmd = (
         f"oc get storageclass {default_rbd_sc} -o=jsonpath='{{.metadata.annotations}}' "
     )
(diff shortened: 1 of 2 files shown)


tests/functional/storageclass/test_rbd_default_storageclass.py
```diff
@@ -52,12 +47,11 @@ def test_pvc_creation_without_storageclass_name(self, pvc_factory, pod_factory):
...
         sc_attached_to_pvc = pvc_obj.get().get("spec").get("storageClassName")
-        sc_default_in_cluster = default_storage_class(constants.CEPHBLOCKPOOL)
         log.info("Verifying the storageclass attached to PVC is correct.")
 
-        assert (
-            sc_attached_to_pvc == sc_default_in_cluster.name
-        ), "Storageclass attached to PVC is different from StorageClass set as default for BlockPool."
+        assert is_rbd_default_storage_class(
+            sc_name=sc_attached_to_pvc
+        ), "RBD is not default storageclass for Cluster."
```
nehaberry, line 54: If we have multiple RBD storageclasses and one is default, will this message be good enough even then ?

paraggit: Here, we are specifically checking the 'default' tag  on the storageclass which is attached to the PVC.  For more clarity, I will add the storage class name in the assert message.

ebenahar (approved):

jilju (approved):

ocs_ci/helpers/helpers.py
```diff
@@ -4770,9 +4770,7 @@ def is_rbd_default_storage_class(custom_sc=None):
     Returns:
         bool : True if RBD is set as the  Default storage class for the cluster, False otherwise.
     """
-    default_rbd_sc = (
-        constants.DEFAULT_STORAGECLASS_RBD if custom_sc is None else custom_sc
-    )
+    default_rbd_sc = constants.DEFAULT_STORAGECLASS_RBD if sc_name is None else sc_name
```
keemano, line 4773: @paraggit do we need this validation? because, in this case, this function is always called with `sc_name`. if condition to validate `sc_name` will always be not none until no sc is attached to PVC. Looks like always `default_rbd_sc` is always assigned with passed `sc_name`

nehaberry (approved):

PR #10149, opened 2024-07-19
GSS bug automation:  #2240714 and #2188032

mashetty330:
1. 2240714: https://bugzilla.redhat.com/show_bug.cgi?id=2240714
2. 2188032: https://bugzilla.redhat.com/show_bug.cgi?id=2188032

### DIFF
--- a/tests/functional/object/mcg/test_object_versioning.py
+++ b/tests/functional/object/mcg/test_object_versioning.py
@@ -142,6 +142,7 @@ class TestObjectVersioning:
         logger.info("Test succeeded!!")
 
 
+@bugzilla("2240714")
 class TestGetObjectByVersionID:
 
     object_name = "file_1"
@@ -218,7 +219,7 @@ class TestGetObjectByVersionID:
             response_body_v1 == self.object_data
         ), f"Object data doesnt match the version {ver_1} data. Response data: {response_body_v1}"
         logger.info(
-            f"Verified the get object by version id. \n Response data: {response_body_v1}"
+            f"Verified the get object by version id. \nResponse data: {response_body_v1}"
         )
 
         # cache the bucket name to pass it to post upgrade verification
@@ -274,5 +275,5 @@ class TestGetObjectByVersionID:
             response_body_v1 == self.object_data
         ), f"Object data doesnt match the version {ver_1} data. Response data: {response_body_v1}"
         logger.info(
-            f"Verified the get object by version id. \n Response data: {response_body_v1}"
+            f"Verified the get object by version id. \nResponse data: {response_body_v1}"
         )
--- a/tests/functional/object/mcg/test_object_versioning.py
+++ b/tests/functional/object/mcg/test_object_versioning.py
@@ -15,6 +15,7 @@ from ocs_ci.framework.pytest_customization.marks import (
     mcg,
     post_upgrade,
     pre_upgrade,
+    polarion_id,
 )
 from ocs_ci.ocs.bucket_utils import (
     s3_put_bucket_versioning,
@@ -147,6 +148,7 @@ class TestObjectVersioning:
 @mcg
 @red_squad
 @bugzilla("2240714")
+@polarion_id("OCS-6177")
 class TestGetObjectByVersionID:
 
     # Common consts between the
--- a/tests/functional/object/mcg/test_object_versioning.py
+++ b/tests/functional/object/mcg/test_object_versioning.py
@@ -19,6 +19,7 @@ from ocs_ci.ocs.bucket_utils import (
     s3_put_object,
     s3_delete_object,
     s3_head_object,
+    s3_get_object,
 )
 from ocs_ci.ocs.resources.pod import get_pods_having_label, Pod
 from ocs_ci.ocs import constants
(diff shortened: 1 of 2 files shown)


tests/functional/object/rgw/test_obc_quota.py
```diff
@@ -117,3 +121,80 @@ def test_obc_quota(
...
+            )
+            if alert.get("labels").get("objectbucketclaim") == bucket_name
+        ]
+
+        if len(alerts) == 0:
+            assert False, (
+                f"Alert {constants.ALERT_OBC_QUOTA_OBJECTS_ALERT} doesn't seem to occur "
+                f"despite the bucket being 90% full"
+            )
+
```
sagihirshfeld, line 186:
```suggestion
        assert len(alerts) > 0, (
            f"Alert {constants.ALERT_OBC_QUOTA_OBJECTS_ALERT} doesn't seem to occur "
            f"despite the bucket being 90% full"
        )
```

tests/functional/object/rgw/test_obc_quota.py
```diff
@@ -117,3 +121,80 @@ def test_obc_quota(
...
+        for alert in alerts:
+            if alert_desc in alert.get("annotations").get("description"):
+                logger.info(
+                    f"Verified the alert {constants.ALERT_OBC_QUOTA_OBJECTS_ALERT}"
+                )
+                break
+            else:
+                assert (
+                    False
+                ), f"Alert {constants.ALERT_OBC_QUOTA_OBJECTS_ALERT} doesn't seem have expected format"
```
sagihirshfeld, line 200: This loop either breaks or fails the test on the first iteration, thus only checking the first alert on the list.

If you wish to check all the alerts, the following would fix the bug:
```suggestion
        for alert in alerts:
            assert (
                False
            ), f"Alert {constants.ALERT_OBC_QUOTA_OBJECTS_ALERT} doesn't seem have expected format"     
        logger.info(
            f"Verified the alert {constants.ALERT_OBC_QUOTA_OBJECTS_ALERT}"
            )
```

And If your intention is to sample only one of the alerts, you can use the following instead for simplicity:
```suggestion

        assert alert_desc in alerts[0].get("annotations").get("description"), (
            f"Alert {constants.ALERT_OBC_QUOTA_OBJECTS_ALERT} doesn't seem have expected format"
        )
        logger.info(
            f"Verified the alert {constants.ALERT_OBC_QUOTA_OBJECTS_ALERT}"
        )       
```

I

tests/functional/object/mcg/test_object_versioning.py
```diff
@@ -139,3 +140,140 @@ def test_versioning_parallel_ops(
             query_out
         ), "[Test failed] There are more than one versions considered to be latest version !!"
         logger.info("Test succeeded!!")
+
+
+@bugzilla("2240714")
```
sagihirshfeld, line 145: The `red_squad` and `mcg` decorators are required here

tests/functional/object/mcg/test_object_versioning.py
```diff
@@ -139,3 +140,140 @@ def test_versioning_parallel_ops(
...
+class TestGetObjectByVersionID:
+
+    object_name = "file_1"
+    object_data = "version 1"
+
```
sagihirshfeld, line 150: Some clarification that these are consts, and why they are not just function-scoped vars:

```suggestion

    # Common consts between the 
    # pre-upgrade and post-upgrade tests
    OBJ_NAME = "file_1"
    V1_OBJ_DATA = "version 1"

```

tests/functional/object/mcg/test_object_versioning.py
```diff
@@ -139,3 +140,140 @@ def test_versioning_parallel_ops(
...
+            f"Verified the get object by version id. \nResponse data: {response_body_v1}"
+        )
+
+        # cache the bucket name to pass it to post upgrade verification
+        request.config.cache.set("versioning_enabled_bucket", bucket.name)
+        request.config.cache.set("object_version", ver_1)
+        logger.info(
+            "Cached the bucket name and object name for the post upgrade verification"
+        )
+
```
sagihirshfeld, line 231: Unless I'm missing someting, it looks like you forgot to use this at the upgrade test - either use it and make the class-scoped consts above function-scoped, or delete this.

tests/functional/object/mcg/test_object_versioning.py
```diff
@@ -139,3 +140,140 @@ def test_versioning_parallel_ops(
...
+    def test_get_object_post_upgrade(self, request, mcg_obj_session):
```
sagihirshfeld, line 232:
```suggestion
    @post_upgrade
    def test_get_object_post_upgrade(self, request, mcg_obj_session):
```

tests/functional/object/mcg/test_object_versioning.py
```diff
@@ -139,3 +140,140 @@ def test_versioning_parallel_ops(
...
+
+
+@bugzilla("2240714")
+class TestGetObjectByVersionID:
+
+    object_name = "file_1"
+    object_data = "version 1"
+
+    def test_get_object_pre_upgrade(
+        self, request, bucket_factory_session, mcg_obj_session
```
sagihirshfeld, line 152:
```suggestion
    @pre_upgrade
    def test_get_object_pre_upgrade(
        self, request, bucket_factory_session, mcg_obj_session
```

tests/functional/object/rgw/test_obc_quota.py
mashetty330: @sagihirshfeld I want to sample all the alerts but your suggestion 1 doesnt work right? it straight way asserts. To give more details, I want to check all the alert in the alerts list which is filtered to have OBC Quota reached alerts only. As part of check, i just verify that it adheres the expected format. 

To achieve this, i think i need to replace break with continue and move the log confirming the successful verification out of the for loop.

mashetty330: below should work:

>         for alert in alerts:
            assert alert_desc in alert.get("annotations").get("description"), f"Alert {constants.ALERT_OBC_QUOTA_OBJECTS_ALERT} doesn't seem have expected format"
        logger.info(
            f"Verified the alert {constants.ALERT_OBC_QUOTA_OBJECTS_ALERT}"
        )

tests/functional/object/mcg/test_object_versioning.py
mashetty330: @sagihirshfeld 

using it here right?

```python
  @post_upgrade
    def test_get_object_post_upgrade(self, request, mcg_obj_session):
        """
        Test get object by bucket versioning on a bucket
        which already has some versioned data

        """
        # fetch bucket name from the cache
        bucket_name = request.config.cache.get("versioning_enabled_bucket", None)
        ver_1 = request.config.cache.get("object_version", None)
```

tests/functional/object/rgw/test_obc_quota.py
sagihirshfeld: You're correct my first solution doesn't work - while editing and formatting I must've messed up and reused the original assertion by mistake. Please use the code in your last comment instead  👍

tests/functional/object/mcg/test_object_versioning.py
sagihirshfeld: Yep - must've missed it 😅

udaysk23 (approved):

tests/functional/object/rgw/test_obc_quota.py
```diff
@@ -117,3 +122,75 @@ def test_obc_quota(
...
+        bucket_name = rgw_bucket_factory(amount, interface, quota=quota)[0].name
+        logger.info(f"created rgw bucket {bucket_name} with quota {quota}")
+
+        # fill the bucket with about 90% of maxObjects
+        number_of_objects = int(quota["maxObjects"])
+        write_random_test_objects_to_bucket(
+            awscli_pod_session,
+            bucket_name,
+            test_directory_setup.origin_dir,
+            amount=(number_of_objects * 90) // 100,
```
ebenahar, line 166: this can result in less than 90%.
For example:
if `number_of_objects` is 31, `amount` will be:
(31*90) // 100 = 27
27 are 87% of 31

tests/functional/object/rgw/test_obc_quota.py
```diff
@@ -117,3 +122,75 @@ def test_obc_quota(
                 logger.error("Copy objects to bucket failed unexpectedly!!")
         else:
             logger.info(f"New quota {new_quota_str} got applied!!")
+
+    @bugzilla("2188032")
```
ebenahar, line 126: the bug is fixed and closed, the decorator is not needed

tests/functional/object/mcg/test_object_versioning.py
```diff
@@ -139,3 +143,147 @@ def test_versioning_parallel_ops(
             query_out
         ), "[Test failed] There are more than one versions considered to be latest version !!"
         logger.info("Test succeeded!!")
+
+
+@mcg
+@red_squad
+@bugzilla("2240714")
```
ebenahar, line 150: the bug is fixed and closed, the decorator is not needed

tests/functional/object/rgw/test_obc_quota.py
mashetty330: @ebenahar here we are not considering the decimal value. 90% of 31 will 27.9 not 27. but in this case we are ignoring the decimal value. We dont need 90%, we only need (80- 85%)

ebenahar (approved):

PR #10153, opened 2024-07-22
[GSS] Test object lists with millions of objects under single directory

mashetty330:
This PR Addresses:

1. [aws s3 ls takes a long time](https://bugzilla.redhat.com/show_bug.cgi?id=2277990)


ocs_ci/ocs/bucket_utils.py
```diff
@@ -2712,3 +2712,62 @@ def list_objects_in_batches(
...
+    sampler = TimeoutSampler(
+        timeout=timeout,
+        sleep=30,
+        func=_check_objs_deletion,
+    )
+
+    assert sampler.wait_for_func_status(
+        result=True
+    ), "Not all the objects are marked deleted"
+    logger.info("All the objects are marked expired")
```
sagihirshfeld, line 2773: To allow reusing the util function for the negative case, please consider returning the boolean and move the TimeoutSampling and assertion to the test itself.

tests/cross_functional/scale/noobaa/test_delete_objects.py
```diff
@@ -190,3 +201,96 @@ def test_delete_objects(
         # stop the io running in the background
         event.set()
         io_thread.result()
+
+    @bugzilla("2279742")
+    @bugzilla("2279964")
+    @polarion_id("OCS-6097")
+    @polarion_id("OCS-6096")
```
PrasadDesala, line 208: I don't think adding multiple polarion IDs to a test function is not supported by the polarion importer plugin. Can we combine these two tests?

tests/cross_functional/scale/noobaa/test_delete_objects.py
```diff
@@ -190,3 +201,96 @@ def test_delete_objects(
...
+            )
+        else:
+            # remove the objects in the bucket recursively
+            rm_object_recursive(
+                awscli_pod_session, bucket.name, mcg_obj_session, timeout=7200
+            )
+            log.info("Deleted objects from the bucket recursively")
+
+        # Verify that all the objects are marked as deleted
+        verify_objs_deleted_from_objmds(bucket.name, timeout=72000, sleep=90)
```
PrasadDesala, line 292: 20hr timeout is too much. why do we need to have a 20hr timeout? 
Is there anything that we can do to have a reasonable timeout?

tests/cross_functional/scale/noobaa/test_delete_objects.py
mashetty330: I have done these in the past. There are already many tests with multiple polarion ids

tests/cross_functional/scale/noobaa/test_delete_objects.py
mashetty330: As discussed in the weekly syncup, the only way to go forward is with this long timeout. About 1000 objects gets expired every 1 to 2 minutes timespan. Given 1 million object, 1000000/1000=1000 * 1.5 = 1500/60=25 hours. As per my two last runs we need atleast 20 hour for all the objects to expire. Currently we dont have support to increase the object expiration batch size above 1000.

tests/cross_functional/scale/noobaa/test_delete_objects.py
PrasadDesala: > I have done these in the past. There are already many tests with multiple polarion ids
There is no restriction on adding multiple polarion ids to a test func in ocs-ci. However, I doubt if the test execution status gets updated for all those polarion ids in polarion. At least, it was not supported by the polarion importer plugin the last time I checked, although that was a long time ago. Did you check if the test results were being updated for all those decorated polarion ids tests in polarion?

tests/cross_functional/scale/noobaa/test_delete_objects.py
PrasadDesala: ok, this test is going to be into scale suite. Let's get feedback from scale team on this.

udaysk23: 25 hours for single test seems too much. We should find alternative way to reduce it. CC: @ramkiperiy

mashetty330: @sagihirshfeld So i was under the same impression that we can use that env variable to modify the batch size. But as per my discussion with Ben we cant modify to a size greater than 1000, you can only modify the size within the limit of 1000 objects. It will work if you set the value 900 but wont work for 1500.
Bug: https://issues.redhat.com/browse/DFBUGS-472

ramkiperiy: +1, I would suggest to reduce the amt of objs, since its limitation in MCG and why we should consider the limitation in CI coverage. 
There is already concerns raised to reduce the runtime of Noobaa scale tests which @udaysk23 is looking for enhancements.
If we would like to have this part of CI scripts then I would suggest to have some scale_extended tier marker and don't add it part of scale tier, with scale_extended tier marker you can execute the particular TC on a need basis and it won't be executed part of production or regression suit runs.
@mashetty330 @PrasadDesala @hnallurv ^^

mashetty330:
I have moved out bug 2279742, 2279964 automation into a different PR as it is time consuming tests and would require different strategy to run in scale suite

udaysk23 (approved):

tests/cross_functional/scale/noobaa/test_delete_objects.py
mashetty330: I have moved long running tests to different PR. will address it seperately

sagihirshfeld (approved):

ramkiperiy (approved):

vavuthu (approved):

Issue #10156, opened 2024-07-22
Disable tests/cross_functional/performance/csi_tests/test_auto_reclaim_space_multi_clones.py as the feature is moved out 


pintojoy:
This test verifies automatic creation of Reclaim Space Cron Jobs for multiple clones of RBD PVCs in openshift-* namespace. It also verifies that no Reclaim Space Cron Job is created automatically any of the multiple clones of CephFS PVC in this namespace.  As there was an issue with ARO cluster, this feature was removed and will be implemented at storageclass level rather than at the namespace level in 4.17

4.17 EPIC to implement the same at storageclass level - https://issues.redhat.com/browse/RHSTOR-5876

PR #10170, opened 2024-07-24
Disabling test_auto_reclaim_space_multi_clones test

pintojoy:
Disabling this test as creating reclaim space cron job for namespace with prefix openshift-* has been deprecated since 4.16. This will be implemented at storageclass level rather than at the namespace level in 4.17

This fixes https://github.com/red-hat-storage/ocs-ci/issues/10156


ypersky1980 (approved):
LGTM

ramkiperiy (approved):

PR #10172, opened 2024-07-25
Set cluster_kubeconfig based on the type of resource available in provider cluster only

jilju:
In provider mode multicluster run, set 'cluster_kubeconfig' if the resource kind is available only in provider cluster.
* Created a primary list of 'Kind' which are available in provider cluster.
* Set the cluster_kubeconfig value to provider cluster kubeconfig path if the Kind is available in the list.

Advantages:
* When using the class OCP or OCS, need not switch cluster context to provider to get/create/delete/patch/decribe etc.. a resource that is in provider cluster.

Example: Need not bother about the current cluster context when using the function https://github.com/red-hat-storage/ocs-ci/blob/0b95c6fa7e9a190726079bfd3018daa190e0413b/ocs_ci/ocs/resources/storage_cluster.py#L1658 and also when using the returned object.

* Enables simultaneous usages of two instances of different kinds where one of them is in provider and the other one is in the client cluster.
 eg: When using threading (or a mechanism to run two oc commands at the same time) , get StorageCluster(from provider) and application pod(from client) at the same time. Switching context is not helpful in this situation.

Limitations:
* All config values are not set as the cluster where the resource is present. eg: ocs_version. This have less impact because these config values are not used in 'oc' commands. cluster_namespace will be used which is going to be same for provider and client going forward. Added a TODO to overcome this.

### DIFF
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -357,6 +357,43 @@ NATIVE_STORAGE_CLIENT_YAML = os.path.join(
     PROVIDER_CLIENT_DEPLOYMENT_DIR, "native_storage_client.yaml"
 )
 
+PROVIDER_CLUSTER_RESOURCE_KINDS = [
+    "cephblockpoolradosnamespaces",
+    "cephblockpoolradosnamespace",
+    "cephblockpools",
+    "cephblockpool",
+    "cephclients",
+    "cephclient",
+    "cephclusters",
+    "cephcluster",
+    "cephfilesystems",
+    "cephfilesystem",
+    "cephfilesystemsubvolumegroups",
+    "cephfilesystemsubvolumegroup",
+    "cephobjectstores",
+    "cephobjectstore",
+    "cephobjectstoreusers",
+    "cephobjectstoreuser",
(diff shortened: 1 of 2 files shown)


fbalak (approved):
I think that this is good and it can help to fix a lot of test cases without editing them. It would be good to have this included with cluster context too but it is enough to have it as TODO for now.
The note about simultaneous usage could be solved if each context have assigned static address to its kubeconfig - The static path to kubeconfig needs to exist for this PR to work. In this way the kubeconfig wouldn't have to be specifically provided but it would be used from the context and users would have access to other environment information like config parameters for the given context. It would also mitigate conflicts in different context and kubeconfig.
It would also mean not to provide kubeconfig in parameter but to add cluster_context as parameter (and use current context (config.cluster_ctx.MULTICLUSTER.get("multicluster_index")) as default).

dahorak (approved):

jilju:
Verified in provider-client multicluster run https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/2677/
The change does not break anything.

PR #10185, opened 2024-07-25
Test restart the csi hosted cluster pods

yitzhak12:


yitzhak12:
The test already exists in our ocs-ci.

PR #10258, opened 2024-08-07
[WIP] Rename hci to providermode

ebondare:
"HCI" naming isn't technically correct, so it needs to be replaced.


conf/deployment/fusion_hci_pc/client_bm_upi_1az_rhcos_nvme_2w.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 ENV_DATA:
   platform: 'hci_baremetal'
-  cluster_type: 'hci_client'
+  cluster_type: 'providermode_client'
```
DanielOsypenko, line 4: maybe `storage_client` ?

PR #10268, opened 2024-08-08
Provider client context manager

fbalak:
Simplify context usage for Provider client setups.

### DIFF
--- a/ocs_ci/framework/__init__.py
+++ b/ocs_ci/framework/__init__.py
@@ -481,9 +481,10 @@ class MultiClusterConfig:
             primary_index = primary_config.MULTICLUSTER.get("multicluster_index")
             super().__init__(primary_index)
 
-    class RunWithProviderConfigContext(RunWithConfigContext):
+    class RunWithProviderConfigContextIfAvailable(RunWithConfigContext):
         """
         Context manager that makes sure that a given code block is executed on Provider.
+        If Provider config is not available then run with current config context.
         """
 
         def __init__(self):
@@ -496,9 +497,10 @@ class MultiClusterConfig:
                 switch_index = config.cur_index
             super().__init__(switch_index)
 
-    class RunWithFirstConsumerConfigContext(RunWithConfigContext):
+    class RunWithFirstConsumerConfigContextIfAvailable(RunWithConfigContext):
         """
         Context manager that makes sure that a given code block is executed on First consumer.
+        If Consumer config is not available then run with current config context.
         """
 
         def __init__(self):
--- a/ocs_ci/framework/__init__.py
+++ b/ocs_ci/framework/__init__.py
@@ -494,8 +494,7 @@ class MultiClusterConfig:
                 # no switch happens and code runs on current cluster
                 logger.DEBUG("No provider was found - using current cluster")
                 switch_index = config.cur_index
-            switch_index = config
-            super().__init__(primary_index)
+            super().__init__(switch_index)
 
     class RunWithFirstConsumerConfigContext(RunWithConfigContext):
         """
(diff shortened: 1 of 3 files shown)


fbalak:
Verified: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-odf-multicluster/2630/

DanielOsypenko (approved):

jilju (dismissed):

ocs_ci/framework/__init__.py
```diff
@@ -479,6 +481,36 @@ def __init__(self):
             primary_index = primary_config.MULTICLUSTER.get("multicluster_index")
             super().__init__(primary_index)
 
+    class RunWithProviderConfigContext(RunWithConfigContext):
+        """
+        Context manager that makes sure that a given code block is executed on Provider.
```
petr-balogh, line 486: I would mention in docstring that if no provider cluster found, the code will run on current index.
And not sure we should change name of class to something like:
`RunWithProviderOrCurrnetConfigContext`

BTW This logic makes sense only if we would like to have this code using this context manager to work in provider/consumer as well as other type of executions, correct?

ocs_ci/framework/__init__.py
```diff
@@ -479,6 +481,36 @@ def __init__(self):
...
+            except ClusterNotFoundException:
+                # if no provider is available then set the switch to current index so that
+                # no switch happens and code runs on current cluster
+                logger.DEBUG("No provider was found - using current cluster")
+                switch_index = config.cur_index
+            super().__init__(switch_index)
+
+    class RunWithFirstConsumerConfigContext(RunWithConfigContext):
+        """
+        Context manager that makes sure that a given code block is executed on First consumer.
```
petr-balogh, line 501: Same as above comment applies here.

ocs_ci/ocs/resources/storageconsumer.py
```diff
@@ -86,53 +87,36 @@ def stop_heartbeat(self):
         """
         Suspend status reporter cron job.
         """
-        self._switch_consumer_cluster()
-        patch_param = '{"spec": {"suspend": true}}'
-        self.heartbeat_cronjob.ocp.patch(
-            resource_name=self.heartbeat_cronjob.name, params=patch_param
-        )
-        self._switch_provider_cluster()
+        with config.RunWithConfigContext(self.consumer_context):
```
petr-balogh, line 90: Do we hold some specific consumer context in `self.consumer_context`?
Or why not using new context manager `RunWithFirstConsumerConfigContext` here and other occurrences below?

fbalak: It uses `consumer_context` defined for the instance constructor of the `StorageConsumer` class. It doesn't have to be the first consumer in all cases.

ocs_ci/framework/__init__.py
fbalak: > I would mention in docstring that if no provider cluster found, the code will run on current index.

This makes sense.

> And not sure we should change name of class to something like:
> RunWithProviderOrCurrnetConfigContext

How about `RunWithProviderConfigContextIfAvailable`?

> BTW This logic makes sense only if we would like to have this code using this context manager to work in provider/consumer as well as other type of executions, correct?

Yes, this way people can add it to the code without a worry that it breaks other test cases. It allows to mark chunks of code with a context for provider/consumer in generic function e.g. in `ocs_ci/ocs/resources/*` or `tests/conftest.py`

petr-balogh (approved):

petr-balogh:
/cherry-pick release-4.16

I think we need to cherry pick to 4.16 at least, right @fbalak ?

DanielOsypenko (approved):

PR #10270, opened 2024-08-08
Introducing ConfigSafeThread

petr-balogh:
Refactoring MultiClusterConfig to support ConfigSafeThread

### DIFF
--- a/ocs_ci/framework/__init__.py
+++ b/ocs_ci/framework/__init__.py
@@ -14,7 +14,7 @@ import logging
 from collections.abc import Mapping
 from dataclasses import dataclass, field, fields
 from ocs_ci.ocs.exceptions import ClusterNotFoundException
-from threading import Thread, RLock, get_ident
+from threading import Thread, RLock, local, get_ident
 
 THIS_DIR = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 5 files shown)


ocs_ci/framework/__init__.py
```diff
@@ -130,9 +133,9 @@ class MultiClusterConfig:
     # multiple cluster contexts
     def __init__(self):
         # Holds all cluster's Config() object
+        self.thread_local_data = local()
```
DanielOsypenko, line 136: this way we create multiple threading.local objects of MultiClusterConfig. Is it the desired behavior? 
Maybe we want isolate MultiClusterConfig from multiple creations? Make it Singleton?

petr-balogh: Yes it's good candidate to define MultiClusterConfig as singleton but it's not a part of this POC and PR to do refactoring of current implementation from this point of view.

In whole framework we have only one instance of config object defined here:

https://github.com/red-hat-storage/ocs-ci/blob/c4b295570eddd7a0e61231738c8adbf2a2fd11ef/ocs_ci/framework/__init__.py#L503

And we are everywhere referring to the same object - so we don't actually doing in whole framework the other instance of this class.

PR #10284, opened 2024-08-13
[WIP] Updated the platform check condition

amr1ta:
Updated the platform check value from 'constants.HCI_PROVIDER_CLIENT_PLATFORMS' with 'constants.BAREMETAL_PLATFORM', and 'constants.VSPHERE_PLATFORM' as its mentioned in deployment config.

The Error it was causing during ocs installation is, due to unavailable of enough pvs the number of osds its trying to create was unsuccessful and the storage-cluster was not moving to 'Ready' status

### DIFF
--- a/ocs_ci/deployment/provider_client/storage_client_deployment.py
+++ b/ocs_ci/deployment/provider_client/storage_client_deployment.py
@@ -224,7 +224,10 @@ class ODFAndNativeStorageClientDeploymentOnProvider(object):
                     "replica"
                 ] = no_of_worker_nodes
 
-                if self.platform in constants.HCI_PROVIDER_CLIENT_PLATFORMS:
+                if self.platform in [
+                    constants.BAREMETAL_PLATFORM,
+                    constants.VSPHERE_PLATFORM,
+                ]:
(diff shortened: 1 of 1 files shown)


ocs_ci/deployment/provider_client/storage_client_deployment.py
```diff
@@ -224,7 +224,10 @@ def provider_and_native_client_installation(
                     "replica"
                 ] = no_of_worker_nodes
 
-                if self.platform in constants.HCI_PROVIDER_CLIENT_PLATFORMS:
+                if self.platform in [
```
DanielOsypenko, line 227: we need to use HCI_BAREMETAL until it is renamed on the decided new platform name, or until we find all dependencies and replace it with just "baremetal" all over the places

DanielOsypenko (commented):
I assume the error happens because we don't use "hci_baremetal" in deployment job, and for now, it is mandatory until renamed - [link to full yaml config](https://github.com/red-hat-storage/ocs-ci/blob/3ba76680571f7be0f2bcc78c169fdfe4aa0767aa/conf/examples/provider_mode_ibm_cloud_baremetal_kubevirt_clients.yaml#L15C3-L15C28)

PR #10297, opened 2024-08-13
Retry if rack or zone return value is None but assignment is correct for test_crashcollector_pod_existence_on_ceph_pods_running_nodes

am-agrawa:
We saw a few failure instances for `test_crashcollector_pod_existence_on_ceph_pods_running_nodes.py` over VMware where rack assignment is correct but function `get_node_rack_or_zone` returned None. The intention is to add retry to get the correct rack so as to stabilise the test.


prsurve (approved):

pintojoy (approved):
LGTM

ebenahar (approved):

PR #10304, opened 2024-08-15
Set the namespace dynamically in OC BS creation of GCP and Azure

sagihirshfeld:
When creating GCP and Azure Backingstores via OC + YAMLs, we've been using `openshift-storage` namespace regardless of the actual namespace we're using in a run. This PR simply sets the namespace from `config.ENV_DATA["cluster_namespace"]`.


sagihirshfeld:
Setting the `Verified` label due to the very limited scope of this PR, and since I've verified this locally when using a cluster that doesn't use the `openshift-storage` namespace.

udaysk23 (approved):

dahorak (approved):

PR #10344, opened 2024-08-22
Test Automation for feature PV key rotation.

paraggit:
Test Encrypted PVC keyrotation.
        Following steps will be verified with Vault KMS, using a combination of (v1, v2) and with and without Vault namespaces.
        
          Steps:

           1.	Create an encrypted RBD storage class.
            2.	Add an annotation to the encrypted storage class.
            "keyrotation.csiaddons.openshift.io/schedule='*/3 * * * *'"
            3.	Configure the Persistent Volume (PV) encryption settings with the Vault service.
            4.	Create a PVC using the encrypted storage class.
            5.	Deploy a Pod that utilizes the previously created PVC for storage.
            6.	Start an IO workload with verify=True option.
            7.	Wait for the key rotation to occur for the PV.
            8.	Check for any error messages that appear during the IO operation.

### DIFF
--- a/ocs_ci/helpers/keyrotation_helper.py
+++ b/ocs_ci/helpers/keyrotation_helper.py
@@ -8,6 +8,7 @@ from ocs_ci.framework import config
 from ocs_ci.ocs.resources.pvc import get_deviceset_pvcs
 from ocs_ci.ocs.exceptions import UnexpectedBehaviour
 from ocs_ci.utility.retry import retry
+from ocs_ci.utility.kms import get_kms_deployment
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


ocs_ci/helpers/keyrotation_helper.py
```diff
@@ -350,3 +351,48 @@ def compare_old_with_new_keys():
...
+            device_handle (str): The handle or identifier for the device whose key
+                rotation is to be checked.
+
+        Returns:
+            bool: True if the key rotation is successful, otherwise False.
+        """
+        current_key = self.kms.get_pv_secret(device_handle)
+
+        @retry(UnexpectedBehaviour, tries=tries, delay=delay)
+        def compare_keys():
```
ebenahar, line 383: this could be a separate function, which we can use in both `wait_till_keyrotation` and as a standalone one

ocs_ci/utility/kms.py
```diff
@@ -607,18 +607,32 @@ def gather_vault_config(self):
...
+            cm_data = cm_obj.get(constants.VAULT_KMS_CONNECTION_DETAILS_RESOURCE)
+            return cm_data.get().get("data")[resource_name]
+
+        elif cm_obj.is_exist(constants.VAULT_KMS_CSI_CONNECTION_DETAILS):
+            cm_data = cm_obj.get(constants.VAULT_KMS_CSI_CONNECTION_DETAILS)
+            for k, v in cm_data.get("data").items():
+                json_out = json.loads(cm_data.get("data")[k])
+                if json_out.get("KMS_SERVICE_NAME") == "vault":
+                    return json_out[resource_name]
+            else:
```
ebenahar, line 632: seems like wrong indentation for this `else` block

paraggit: This else block is associated only with the for loop. It executes when no KMS_SERVICE_NAME == "vault" is found in the list of connection strings being iterated over in the for loop.

ebenahar (dismissed):

AviadP (approved):

AviadP (approved):

ebenahar (approved):

PR #10357, opened 2024-08-27
<GSS Closed Loop BZ: 2264900> Tests to verify restore a pvc from snapshot when the parent PVC is deleted

Akarsha-rai:


tests/functional/pv/pvc_snapshot/test_restore_snapshot_when_parent_pvc_deleted.py
```diff
@@ -0,0 +1,228 @@
...
+)
+from ocs_ci.ocs.resources import pod
+from ocs_ci.helpers import helpers
+
+log = logging.getLogger(__name__)
+
+
+@green_squad
+@tier1
+@skipif_ocs_version("<4.16")
```
jilju, line 22: Isn't the test case supported in lower versions ?

Akarsha-rai: I see the [bug2264900](url) fixed in 4.16, so I believe it is not supported in lower version.

jilju (approved):

tests/functional/pv/pvc_snapshot/test_restore_snapshot_when_parent_pvc_deleted.py
```diff
@@ -0,0 +1,228 @@
...
+        for snap_obj in snap_objs:
+            access_modes = access_modes_dict[snap_obj.interface][
+                snap_obj.parent_volume_mode
+            ]
+            for access_mode in access_modes:
+                restore_obj = snapshot_restore_factory(
+                    snapshot_obj=snap_obj,
+                    volume_mode=snap_obj.parent_volume_mode,
+                    access_mode=access_mode,
+                    status="",
```
PrasadDesala, line 152: any reason for not checking the restored PVC status for bound state here ?

tests/functional/pv/pvc_snapshot/test_restore_snapshot_when_parent_pvc_deleted.py
```diff
@@ -0,0 +1,228 @@
...
+        restore_pvcs = []
+        for snap_obj in snap_objs:
+            access_modes = access_modes_dict[snap_obj.interface][
+                snap_obj.parent_volume_mode
+            ]
+            for access_mode in access_modes:
+                restore_obj = snapshot_restore_factory(
+                    snapshot_obj=snap_obj,
+                    volume_mode=snap_obj.parent_volume_mode,
+                    access_mode=access_mode,
```
PrasadDesala, line 151: how are you making sure that we are not restoring the snapshot with the same access mode? To me, it looks like with the current approach, there will be a few restores with same access mode.

tests/functional/pv/pvc_snapshot/test_restore_snapshot_when_parent_pvc_deleted.py
Akarsha-rai: It takes more time to reach Bound state, to save time, Bound status is verified as next step

PrasadDesala (approved):

jilju (approved):

PR #10382, opened 2024-09-01
Skipping in case of external mode setup

paraggit:
Skipping test `test_encrypted_pvc_no_expansion_option`  in case of external mode setup


nehaberry (approved):
If PVC Encryption and the related activities are not supported on external mode, Adding the approval for the PR

PR #10405, opened 2024-09-03
Adds test coverage of expanding PVC size for noobaa-db-pg pod

udaysk23:
Adds test coverage of expanding PVC size for noobaa-db-pg pod

KC article: https://access.redhat.com/solutions/6976547


sagihirshfeld (commented):
The test looks solid, but please also verify that the default noobaa backingstore is healthy after the rescaling of the noobaa-db statefulset 

(as instructed in step number 6 of the KCS)

udaysk23:
> The test looks solid, but please also verify that the default noobaa backingstore is healthy after the rescaling of the noobaa-db statefulset
> 
> (as instructed in the 6th step of the KCS)

@sagihirshfeld Thanks for the suggestion, I have added BS validation step in script now

sagihirshfeld (dismissed):
LGTM

mashetty330 (commented):
@udaysk23 we already have fixture to scale the noobaa db pv.  `scale_noobaa_db_pod_pv_size()` is the one. You can use that fixture here.

mashetty330 (changes requested):
use the existing KCS automation for the same. `def scale_noobaa_db_pod_pv_size(request):`

sagihirshfeld (dismissed):

tests/functional/object/mcg/test_noobaa_db_pg_expansion.py
```diff
@@ -0,0 +1,88 @@
...
+                avail_size = int(max_avai_bytes) / constants.GB
+                break
+        logger.info(f"Max avail size on ceph block pool is {avail_size} GB")
+
+        # record PVC capacity before expanding capacity
+        noobaa_db_pvc_obj = get_pvc_objs(pvc_names=[constants.NOOBAA_DB_PVC_NAME])[0]
+        pvc_capacity = get_pvc_size(noobaa_db_pvc_obj)
+        logger.info(f"Current PVC capacity is {pvc_capacity} GB")
+
+        # Validate PVC expansion against available RBD size and current PVC size
```
mashetty330, line 64: @udaysk23 why this step is neccessary?

udaysk23: @mashetty330 Script is increasing Capacity by averaging current size and available RBD size. Also script is using current pvc capacity to validate the real change happened or not

mashetty330: sorry i referenced the wrong code

mashetty330:
```python
        assert (
            pvc_capacity <= avail_size
        ), f"PVC expand operation is not valid, current PVC capacity {pvc_capacity} >= ceph RBD AVAIL size {avail_size}"

        # Calculate new PVC size by averaging MAX AVAIL of the RBD Pool and current PVC size
        new_pvc_size = round((avail_size + pvc_capacity) / 2)
```

mashetty330: This piece of code i meant to ask. Is it neccessary to use average of available size and current pvc capacity? 
because for example, if avail size 400GB and pvc size is 50GB and the average of that would be 225GB for only noobaa db pv and we know that we cant revert this back also. Cant we just increase it by small amount? like maybe add half of what current pv size is. something like new_pvc_size = current_size + (current_size//2)

udaysk23: Thanks for the suggestion @mashetty330. I have added one more condition to get valid PVC value

mashetty330 (approved):

sagihirshfeld (dismissed):

mashetty330 (dismissed):

keesturam (dismissed):

sagihirshfeld (approved):

keesturam (approved):

PR #10421, opened 2024-09-04
rosa hcp deployment

DanielOsypenko:
Deployment of ROSA HCP cluster with setting up ODF
* olm changes to deploy storage on custom namespace
* logs in deployment of OCP and ODF are improved
* templates refreshed with custom namespace
* post-deployment settings adjusted
* post-deployment verifications adjusted
* ROSAProdEnvCluster helper class has a sibling ROSAStageEnvCluster, module rosa_prod_cluster_helpers.py
 became rosa_cluster_helpers.py 
* ROSAProdAdminLoginFailedException renamed to ROSAAdminLoginFailedException
* create_and_attach_sts_role adjusted to work with clusters without infraID
* openshift_dedicated.py finalized with get_admin_name and cached secret data
* ROSA download binary, login, create_account_roles, wait_operator_roles are corrected, functions added: create oidc-config, get_machinepools, wait_machinepool_replicas_ready and oidc - helper methods
*  helper functions created and corrected, such as get_ocp_version


DanielOsypenko:
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/42938 - deployment job
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/42973 - acceptance suite

DanielOsypenko:
TODO’s: 
1. add delete for some resources, such as oidc-config to the tearDown (destroy cluster)
2. create_openshift_install_log_file

conf/deployment/aws/rosa_hcp_1az_3w_m5.12x.yaml
```diff
@@ -1,13 +1,26 @@
...
+  ocm_cli_version: '0.1.67'
+  rosa_cli_version: '1.2.45'
+  live_deployment: true
+  customized_deployment_storage_class: 'gp3-csi'
 RUN:
-  username: 'cluster-admin'
+  username: "cluster-admin"
 ENV_DATA:
   platform: 'rosa_hcp'
-  deployment_type: 'managed'
```
dahorak, line 7: I think you should preserve the `deployment_type` parameter in the config too, because otherwise it is [set by default to `ipi`](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L155), which is not correct in this case, so I think it will be better to have it explicitly configured. (It is quite heavily used parameter.)

dahorak (commented):
Overall it looks good to me, I have just one comment related to removing `deployment_type` from the deployment config.

And generally one or two notes:
I don't have strong opinion about using the `log_step` wrapper, I understood it's purpose, but I'm not sure if it is not overused in the deployment part (on the other hand I fully agree with adding log message before each logical step in the workflow, which will surely help with debugging). Better to get some input from others as well.

Also little bit related to the previous one, the PR is quite large and some of the changes are not directly related to each other, which makes the review little bit more difficult. So next time, I would propose to try to split the work to multiple logical parts and separated PRs, if possible - which will help with the reviews.

conf/deployment/aws/rosa_hcp_1az_3w_m5.12x.yaml
DanielOsypenko: Thank you for the review, and important question. 

The problem is that if deployment_type is set to 'managed', it will fail in a large number of 'if' blocks, both within the deployment and tests. The platform behaves as both managed and self-managed. From my perspective, it was presented as self-managed because the user will have access to metrics, alerts, and the ability to manage storage independently. Additionally to this, user can manage HCP nodes independently. So the support is for control-plane machines/nodes and k8s resources.

Epic: [ODF self-managed installable on ROSA HostedControlPlane (HCP)](https://issues.redhat.com/browse/RHSTOR-4858)
The experience should be as close to regular ODF on AWS as possible.

This left me with open question, should we present another deploment_type, such as "semi_managed"?

DanielOsypenko:
PR deployment verified on IBM Cloud as well:
https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/43027/

DanielOsypenko:
Hey Daniel, in resp to this comment
> I don't have strong opinion about using the log_step wrapper, I understood it's purpose, but I'm not sure if it is not overused in the deployment part 

I found it helpful to break down the steps in the deployment process. The code isn’t very well-organized, as we all tend to add new 'if-else' blocks for tuning specific platforms. By spending some time understanding what’s happening, I found it useful (at least for myself) to track progress in the console logs using 'step <num>' lines.

When it comes to logs, we often lack clear explanations and rely on the general knowledge of the PR owner. As a general rule, more detailed logs can save engineers time, which is the most valuable resource during the testing process.

Additionally, @petr-balogh started changes to minimize logs by removing 'exec_cmd'. I’m not sure if this is still in progress, but if we go this route, we will have to rely heavily on informative log.info lines.

Adding a few snippets to show that we are not over-polluting the log console of what we see from the console:
image

image

conf/deployment/aws/rosa_hcp_1az_3w_m5.12x.yaml
dahorak: Thanks for the explanation. It will be probably better to introduce new type, because otherwise it will be considered as `ipi` (due to the default value for this parameter), which is not correct and might be misleading.

dahorak (dismissed):
I would propose to add some note/comment also directly to those deployment config files.

Othervise LGTM

dahorak (approved):

Issue #10249, opened 2024-08-06
`rgw.py.RGW.get_credentials` fails with IndexError in RGW SSL deployments


sagihirshfeld:
The usual we in which we fetch RGW's endpoint on external clusters has to be modified for deployments where we deploy RGW in with SSL enabled:

```
    def get_credentials(self, secret_name=constants.NOOBAA_OBJECTSTOREUSER_SECRET):
        ...
        ...
        secret_ocp_obj = OCP(kind=constants.SECRET, namespace=self.namespace)

        if storagecluster_independent_check():
            secret_name = constants.EXTERNAL_MODE_NOOBAA_OBJECTSTOREUSER_SECRET
            cos_ocp_obj = OCP(
                kind=constants.CEPHOBJECTSTORE,
                namespace=config.ENV_DATA["cluster_namespace"],
            )
            cephobjectstore = cos_ocp_obj.get(
                resource_name=constants.RGW_ROUTE_EXTERNAL_MODE
            )
            endpoint = cephobjectstore["status"]["endpoints"]["insecure"][0]



E           IndexError: list index out of range
```

PR #10426, opened 2024-09-04
Use the secure external endpoint and the needed CA for RGW SSL external deployments

sagihirshfeld:
Fixes: #10249


sagihirshfeld:
The successful validations above demonstrate that the solution functions properly in both RGW and MCG environments, whether using an RGW SSL external deployment or a standard vSphere deployment.

dahorak (approved):

ocs_ci/ocs/awscli_pod.py
```diff
@@ -69,6 +72,17 @@ def create_awscli_pod(scope_name=None, namespace=None, service_account=None):
...
     wait_for_resource_state(awscli_pod_obj, constants.STATUS_RUNNING, timeout=180)
+
+    awscli_pod_obj.exec_cmd_on_pod(
+        f"cp {constants.SERVICE_CA_CRT_AWSCLI_PATH} {constants.AWSCLI_CA_BUNDLE_PATH}"
+    )
+
+    if storagecluster_independent_check() and config.EXTERNAL_MODE["rgw_secure"]:
+        log.info("Concatenating the RGW CA to the AWS CLI pod's CA bundle")
+        awscli_pod_obj.exec_cmd_on_pod(
+            f"bash -c 'wget -O - {config.EXTERNAL_MODE['rgw_cert_ca']} >> {constants.AWSCLI_CA_BUNDLE_PATH}'"
```
petr-balogh, line 83: @sagihirshfeld 
`>>` Question for this. Do we really want to just add content to existing content, so each run, it will add the data into same file, just want to confirm, it's intentional? Or if we wanted to just write it once `>` and replace previous content?

petr-balogh (approved):

fbalak (approved):

ocs_ci/ocs/awscli_pod.py
sagihirshfeld: Apologies for the late response here.

The use of `>>` here is intentional because I want to concatenate the rgw-ssl specific certificate to the existing one which was injected when the pod was created. 

This should only happen once per such bundle file  and not create a huge repeating file by accident, because `AWSCLI_CA_BUNDLE_PATH` is in ephemeral storage on the pod (under /tmp) and we're running this logic only once per such pod creation.

mashetty330: @sagihirshfeld storing this AWSCLI_CA_BUNDLE_PATH in ephemeral storage (/tmp) is causing the regression. After node restart tests, these ca-cert data is lost causing failure.

PR #10468, opened 2024-09-11
Fix direct use of logging module

clacroix12:
This was introduced because of changes I am making to support python 3.11 which has temporarily made it so our PR checks weren't verifying logging statements.


clacroix12:
@nagendra202 can you confirm that this doesn't break what you intended with your logging here? I see you added some configuration to the logging but never created an explicit logger to use. I also want to make sure that this doesn't impact the rest of the frameworks logging.

vavuthu (approved):

petr-balogh (approved):

PR #10525, opened 2024-09-18
[bz:2274165] Test to verify mgr does not crash after enabling rook module

parikshithb:
Covers bz: 2274165

### DIFF
--- a/tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
+++ b/tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
@@ -19,7 +19,7 @@ log = logging.getLogger(__name__)
 @tier2
 @brown_squad
 @skipif_external_mode
-@skipif_ocs_version("<4.16")
+@skipif_ocs_version("<4.15")
 @bugzilla("2274165")
 @pytest.mark.polarion_id("OCS-6240")
 class TestMgrRookModule(ManageTest):
@@ -78,7 +78,6 @@ class TestMgrRookModule(ManageTest):
 
         """
         toolbox = pod.get_ceph_tools_pod()
-
         log.info("Enabling rook module on mgr")
         toolbox.exec_ceph_cmd("ceph mgr module enable rook")
         log.info("Setting orchestrator backend to rook")
@@ -107,7 +106,7 @@ class TestMgrRookModule(ManageTest):
             timeout=600,
             sleep=30,
             func=run_cmd_verify_cli_output,
-            cmd="ceph crash ls",
+            cmd="ceph crash ls-new",
             expected_output_lst={"mgr"},
             cephtool_cmd=True,
         )
--- a/tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
+++ b/tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
@@ -69,7 +69,8 @@ class TestMgrRookModule(ManageTest):
 
     def test_mgr_enable_rook_backend_module(self):
         """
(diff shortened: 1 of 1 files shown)


nagendra202 (commented):
Code looks fine for me. Can you please run a validation job?

tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
```diff
@@ -0,0 +1,115 @@
...
+
+        log.info(
+            "Verify that there are no crash events for any of the ceph-mgr daemons by waiting upto 600s"
+        )
+        sample = TimeoutSampler(
+            timeout=600,
+            sleep=30,
+            func=run_cmd_verify_cli_output,
+            cmd="ceph crash ls",
+            expected_output_lst={"mgr"},
```
PrasadDesala, line 111: I think archived crashes will still be listed in the "ceph crash ls" output. The test can fail falsely if there are any existing archived mgr crashes. Instead of checking for a specific mgr, should we check for 
"expected_output_lst={"HEALTH_WARN", "daemons have recently crashed"}," and fail the test if we see this message?

tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def test_mgr_enable_rook_backend_module(self):
+        """
+        Test to verify there are no crashes post enabling rook module on mgr and
+        setting rook as an orchestrator backend.
+        BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2274165
+
+        """
+        toolbox = pod.get_ceph_tools_pod()
+
```
PrasadDesala, line 81: you can delete the empty line

tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
```diff
@@ -0,0 +1,115 @@
...
+from ocs_ci.framework.testlib import bugzilla, ManageTest
+from ocs_ci.utility.utils import TimeoutSampler
+
+log = logging.getLogger(__name__)
+
+
+@tier2
+@brown_squad
+@skipif_external_mode
+@skipif_ocs_version("<4.16")
```
PrasadDesala, line 22: the fix was backported to previous release as well

parikshithb: done

tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
parikshithb: removed

tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
parikshithb: ack, make sense. I changed to use `crash ls-new` instead of `crash ls` which suits here.
As per ceph docs: `Archive a crash report so that it is no longer considered for the RECENT_CRASH health check and does not appear in the crash ls-new output (it will still appear in the crash ls output).`
https://docs.ceph.com/en/quincy/mgr/crash/#commands

parikshithb: made changes as discussed.

PrasadDesala (approved):

Shrivaibavi (approved):
LGTM

prsurve (approved):

PR #10532, opened 2024-09-19
Switch to acm ctx before checking S3storeprofiles

shylesh:


ocs_ci/deployment/deployment.py
```diff
@@ -3752,7 +3752,7 @@ def deploy(self):
         config.switch_ctx(old_ctx)
         # Only on the active hub enable managedserviceaccount-preview
         acm_version = get_acm_version()
-
+        config.switch_acm_ctx()
```
petr-balogh, line 3755: Shouldn't we then return back to old_ctx later on in this function as we do on line 3752?

shylesh: @petr-balogh , I have rearragend the ctx switches between acm and old ctx.

@prsurve , please let me know if this is ok

prsurve: yes LGTM

prsurve (approved):

petr-balogh (approved):

ocs_ci/deployment/deployment.py
```diff
@@ -3749,10 +3749,9 @@ def deploy(self):
             config.switch_ctx(index)
             logger.info("Creating Resource DataProtectionApplication")
             run_cmd(f"oc create -f {constants.DPA_DISCOVERED_APPS_PATH}")
-        config.switch_ctx(old_ctx)
         # Only on the active hub enable managedserviceaccount-preview
         acm_version = get_acm_version()
-
+        config.switch_acm_ctx()
```
petr-balogh, line 3754: This line should be moved above line 3753 as get_acm_version require ACM Context

PR #10589, opened 2024-09-26
Add new conf file for 'openshift-storage' cluster namespace

yitzhak12:


jilju (approved):

amr1ta (approved):
looking good

fbalak (approved):
It looks ok. Why is this change needed? The same value of `cluster_namespace` is set in default config https://github.com/red-hat-storage/ocs-ci/blob/b0181e73c9871998705edceb97edba5e469ab9b6/ocs_ci/framework/conf/default_config.yaml#L148

ebenahar (approved):

PR #10600, opened 2024-09-30
WIP: Add option to work with other custom StorageCluster CR name

OdedViner:


ocs_ci/ocs/constants.py
```diff
@@ -450,6 +450,11 @@
 DEFAULT_STORAGECLASS_RBD_THICK = f"{DEFAULT_CLUSTERNAME}-ceph-rbd-thick"
```
petr-balogh, line 450: Did you make sure the usage of all of these above is replaced by your new approach using storagecluster name from config + suffix? I would delete those from constants to be used, and rather have only those suffixes and combined by suffixes.

petr-balogh (approved):

PR #10610, opened 2024-10-02
Ingress Node Firewall deployment on un-released OCP version (from QE App Registry catalog)

dahorak:
- create QE App Registry catalog source for Ingress Node Firewall
- create also ICSP for Brew registry

### DIFF
--- a/ocs_ci/deployment/ingress_node_firewall.py
+++ b/ocs_ci/deployment/ingress_node_firewall.py
@@ -2,10 +2,17 @@ import logging
 
 from ocs_ci.framework import config
 from ocs_ci.ocs import constants, exceptions
+from ocs_ci.ocs.resources.catalog_source import CatalogSource
 from ocs_ci.ocs.resources.csv import CSV, get_csvs_start_with_prefix
 from ocs_ci.ocs.resources.ocs import OCS
+from ocs_ci.ocs.ocp import OCP
+from ocs_ci.ocs.resources.packagemanifest import PackageManifest
 from ocs_ci.utility import templating
-from ocs_ci.utility.utils import TimeoutSampler
+from ocs_ci.utility.utils import (
+    get_ocp_version,
+    TimeoutSampler,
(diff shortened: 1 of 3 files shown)


dahorak:
Verification job: https://url.corp.redhat.com/a2d7e05

ocs_ci/deployment/ingress_node_firewall.py
```diff
@@ -106,6 +120,70 @@ class IngressNodeFirewallInstaller(object):
 
     def __init__(self):
         self.namespace = constants.INGRESS_NODE_FIREWALL_NAMESPACE
+        self.source = constants.OPERATOR_CATALOG_SOURCE_NAME
+
+    def icsp_brew_registry_exists(self):
```
petr-balogh, line 125: As we discussed offline, I suggest to open issue to move the logic of brew registry to different module and reuse it in other places as well to not repeat the code. This way we can move first code out of this modul to the other, and verify only this change here. The other places - we can replace in separate PRs and verify them separately to know we didn't break other things.

petr-balogh (dismissed):

ocs_ci/deployment/ingress_node_firewall.py
dahorak: For now, I've at least moved the code related to qe-app-registry catalog to new module: 6cad0f41c911dc41a1a14b2c9a6ced49b41897c1

dahorak:
> Also I guess we don't need backport it as lower versions are GAed and we don't need this logic, right? 
yes, no need to backport

petr-balogh (approved):

dahorak:
Another verification job: https://url.corp.redhat.com/89be95d (this verification includes the last commit).

fbalak (dismissed):
LGTM, icsp() and catalog_source() function names could be more informative (suffix _exists would be good but icsp_brew_registry_exists also exists so it would collide) but it could work as is.

petr-balogh (approved):

Issue #10586, opened 2024-09-26
Collection of go versions is stuck


petr-balogh:
exec -i ux-backend-server-7b6d74c56b-shsjb -c oauth-proxy -- rpm -qa
2024-09-24 21:42:15  15:42:15 - MainThread - ocs_ci.utility.utils - INFO  - Executing command: oc -n openshift-storage exec -i ux-backend-server-7b6d74c56b-shsjb -c oauth-proxy -- /bin/bash -c '[ -f /go.version ] && cat /go.version || exit 0'
2024-09-26 15:17:37  Aborted by [Petr Balogh](https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/user/pbalogh)
2024-09-26 15:17:37  Sending interrupt signal to process

Job:
https://url.corp.redhat.com/6b2c353
qe-deploy-ocs-cluster-prod/13403/

petr-balogh:
I see it again here:
```
2024-09-29 18:56:22  12:56:22 - MainThread - ocs_ci.utility.utils - INFO  - Executing command: oc -n openshift-storage exec -i ux-backend-server-54cf6c4f4-9wln6 -c ux-backend-server -- /bin/bash -c '[ -f /go.version ] && cat /go.version || exit 0'
2024-09-29 18:56:23  12:56:23 - MainThread - ocs_ci.utility.utils - INFO  - Executing command: oc -n openshift-storage exec -i ux-backend-server-54cf6c4f4-9wln6 -c oauth-proxy -- rpm -qa
2024-09-29 18:56:24  12:56:24 - MainThread - ocs_ci.utility.utils - INFO  - Executing command: oc -n openshift-storage exec -i ux-backend-server-54cf6c4f4-9wln6 -c oauth-proxy -- /bin/bash -c '[ -f /go.version ] && cat /go.version || exit 0'
```
https://url.corp.redhat.com/4dc4f09

petr-balogh:
From logs:
```
tail -n 25 logs
python3-rpm-4.16.1.3-24.el9_2.x86_64
python3-setuptools-53.0.0-12.el9_2.1.noarch
python3-urllib3-1.26.5-3.el9_2.2.noarch
python3-requests-2.25.1-7.el9_2.noarch
python3-cloud-what-1.29.33.2-1.el9_2.x86_64
python3-subscription-manager-rhsm-1.29.33.2-1.el9_2.x86_64
[...]

PR #10617, opened 2024-10-04
Release lock when exception hit in exec_cmd

petr-balogh:
This will prevent deadlocks when exception during subprocess.run hit.

Related issue:
https://github.com/red-hat-storage/ocs-ci/issues/10586

Hopefully it will fix this issue above.

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -597,6 +597,7 @@ def exec_cmd(
     silent=False,
     use_shell=False,
     cluster_config=None,
+    lock_timeout=7200,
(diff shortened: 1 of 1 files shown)


dahorak (approved):

vavuthu (approved):

petr-balogh:
Verification:
 https://url.corp.redhat.com/2fe6990

petr-balogh:
I did run several times with this PR, it doesn't break anything , but once I encountered the same issue, so it doesn't seems to resolve the issue but for sure it's better to merge as it prevents the deadlocks as well.

PR #10630, opened 2024-10-08
Adds test script for validating bucket delete operation using OBC generated credentials

udaysk23:
Adds test script for validating bucket delete operation using OBC generated credentials

GSS bug: https://bugzilla.redhat.com/show_bug.cgi?id=2301657#c16

### DIFF
--- a/tests/functional/object/mcg/test_bucket_delete_using_obc_creds.py
+++ b/tests/functional/object/mcg/test_bucket_delete_using_obc_creds.py
@@ -1,0 +1,83 @@
+import base64
+import boto3
+import logging
+
+from ocs_ci.framework.pytest_customization.marks import (
+    tier2,
+    bugzilla,
+    red_squad,
+    runs_on_provider,
+    skipif_ocs_version,
+    polarion_id,
+    mcg,
+)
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_bucket_delete_using_obc_creds.py
```diff
@@ -0,0 +1,82 @@
...
+@mcg
+@red_squad
+@runs_on_provider
+@skipif_ocs_version("<4.17")
+@polarion_id("OCS-6252")
+@bugzilla("2301657")
+def test_bucket_delete_using_obc_creds(mcg_obj, bucket_factory):
+    """
+    Test bucket deletion using OBC generated credentials and noobaa admin credentials
+
```
sagihirshfeld, line 36: Please add the expected behavior in the Docstring, especially since it's a new expectation that was just introduced in the fix of the covered BZ:
```suggestion
    Verify that deletion of an OBC's bucket is denied when using the
    OBC's credentials, but permitted with the noobaa admin credentials.    

```

udaysk23: Fixed

keesturam (approved):

PR #10641, opened 2024-10-09
Check that the ceph osd df tree weight and size are equal and reflect the current osd size

yitzhak12:
Add a new check for the resize osd tests: 
Check that the ceph osd df tree weight and size are equal and reflect the current osd size.
See more details in the thread: https://ibm-systems-storage.slack.com/archives/C06DWD30QTZ/p1724321789366579?thread_ts=1724316882.105889&cid=C06DWD30QTZ.
One more note regarding the implementation in the PR: When getting the `ceph osd df tree` output, I used the regular format and not the JSON format because the JSON format lacks the values of `WEIGHT` and `SIZE`.

### DIFF
--- a/ocs_ci/helpers/osd_resize.py
+++ b/ocs_ci/helpers/osd_resize.py
@@ -61,9 +61,7 @@ def check_resources_state_post_resize_osd(old_osd_pods, old_osd_pvcs, old_osd_pv
 
     Raises:
         StorageSizeNotReflectedException: If the OSD pods failed to restart
-        ResourceWrongStatusException: If the following occurs:
-            1. The OSD pods failed to reach the status Terminated or to be deleted
-            2. The old PVC and PV names are not equal to the current PVC and PV names
+        ResourceWrongStatusException: The old PVC and PV names are not equal to the current PVC and PV names
 
     """
     old_osd_pods_count = len(old_osd_pods)
--- a/ocs_ci/helpers/osd_resize.py
+++ b/ocs_ci/helpers/osd_resize.py
@@ -22,7 +22,7 @@ from ocs_ci.ocs.resources.storage_cluster import (
     get_deviceset_count,
     resize_osd,
 )
-from ocs_ci.ocs.cluster import check_ceph_osd_tree, CephCluster
+from ocs_ci.ocs.cluster import check_ceph_osd_tree, CephCluster, check_ceph_osd_df_tree
 from ocs_ci.ocs.ui.page_objects.page_navigator import PageNavigator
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -2929,3 +2929,94 @@ def get_csi_images_for_client_ocp_version(ocp_version=None):
     csi_ocp_version_images = csi_images.split(first_str)[1].split(last_str)[0]
     csi_ocp_version_images_urls = extract_image_urls(csi_ocp_version_images)
     return csi_ocp_version_images_urls
+
+
+def parse_ceph_osd_df_tree_weight_and_size():
```
ebenahar, line 2934: this has similarity to https://github.com/red-hat-storage/ocs-ci/pull/10359/files#diff-a4d73d82c94d13634aded838ea2490c117996218f405a8da15f9ecea093f3f6dR1331

Would it be possible to use the same function for both cases?

ebenahar (changes requested):

ocs_ci/ocs/resources/storage_cluster.py
yitzhak12: The values are not the same, but changing the `header` value may work in our case. I will check it.

ebenahar (dismissed):

ocs_ci/helpers/osd_resize.py
```diff
@@ -60,6 +60,7 @@ def check_resources_state_post_resize_osd(old_osd_pods, old_osd_pvcs, old_osd_pv
         old_osd_pvs (list): The old osd PV objects before resizing the osd
 
     Raises:
+        StorageSizeNotReflectedException: If the OSD pods failed to restart
         ResourceWrongStatusException: If the following occurs:
```
prsurve, line 64: if this raises is not used can we remove this ?

yitzhak12: I added it in line 79: https://github.com/red-hat-storage/ocs-ci/pull/10641/files#diff-6bcc223cc6becbbfac8729d997ca9a5c7b7cbe57ef19edecdfac9014c6e36749R79.

yitzhak12: Do you mean to remove the line: 

> 1. The OSD pods failed to reach the status Terminated or to be deleted

yitzhak12: fixed

ebenahar (approved):

prsurve (approved):

PR #10658, opened 2024-10-11
FDF Deployment support

clacroix12:


clacroix12:
I have successfully executed this deployment and verification locally a couple of times but I have also kicked off a PR verification job.

clacroix12:
The verification job will show as a failure but that's because I forgot to uncheck the `deploy OCS` checkbox when kicking the job off.

ebenahar (approved):

petr-balogh (approved):

clacroix12:
Just for clarity here in the PR, this supports deploying GA versions of FDF, similar to how we are currently deploying GA Fusion. I will work on pre-GA deployment support in a follow-up PR.

PR #10697, opened 2024-10-16
Use ephemeral ca cert path for RGW only

mashetty330:


mashetty330:
This PR is not needed as the issue is being fixed by https://github.com/red-hat-storage/ocs-ci/pull/10702

PR #10700, opened 2024-10-17
Support deployment of pre-GA versions of Fusion and FDF

clacroix12:


clacroix12:
Verified locally with updated pull secret.

dahorak (approved):

petr-balogh (approved):

vavuthu (approved):

clacroix12:
Currently working with Elad to get access to the shared user we will be using in our automation in order to update our pull secret. Once I'm able to do that I will test this PR in jenkins to ensure things are working properly with the updated pull secret and mark this PR as verified.

clacroix12:
The FDF install and verification passed but was reported as a failure because the PR test job has no option to disable the `RUN_INSTALL_OCS` step which is unnecessary in this scenario.

PR #10716, opened 2024-10-22
Fix configmap name to whitelist metrics needed for ACM observability on DR monitoring dashboard 

am-agrawa:
Since the existing configmap name is incorrect, metrics are not getting whitelisted and data is not reflecting on the graphs.


am-agrawa:
It's verified locally.

prsurve (approved):

keesturam (approved):

PR #10721, opened 2024-10-23
reboot server if not properly discovered by Assisted Installer

dahorak:
From time to time, OCP deployment via assisted installer fails, because some of the node didn't properly boot to the mode. Usually it is sufficient to reboot the server once more time, which is the aim of this PR.

Related to: [OCSQE-2955](https://url.corp.redhat.com/4e960e8)

### DIFF
--- a/ocs_ci/deployment/assisted_installer.py
+++ b/ocs_ci/deployment/assisted_installer.py
@@ -319,6 +319,13 @@ class AssistedInstallerCluster(object):
                 )
                 break
 
+    def get_infra_env_hosts(self):
+        """
+        Return:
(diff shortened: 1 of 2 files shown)


dahorak:
Verification job: https://url.corp.redhat.com/45dd411
Manually simulated the issue and the problematic server were correctly re-booted.
The deployment failed later on configuration validation issue.

The verification were common for both PR#10721 and PR#10741 (I tested it as part of one PR and later split to those two logical PRs).

vavuthu (approved):

petr-balogh (approved):

PR #10756, opened 2024-10-27
Stablize MCG object expiration tests by waiting after putting lifecycle policy

### LINKED ISSUE #10755: MCG expiration tests fail with "NoSuchBucket" errors
These errors are not consistent and can't be reproduced locally, so it might be a timing issue - perhaps MCG needs some time to process the lifecycle policy before we can perform I.O on the bucket?

sagihirshfeld:
Fixes #10755


sagihirshfeld:
Setting this to verified since the only changes are adding log messages and sleep durations of 10 seconds

OdedViner (approved):

DanielOsypenko (approved):

udaysk23 (approved):

PR #10782, opened 2024-10-30
fix external_cluster_helpers.upload_rgw_cert_ca

dahorak:
- add missing `key_file` parameter


dahorak:
Verified locally the `get_external_cluster_details()` function on existing cluster.

petr-balogh (approved):

PR #10801, opened 2024-11-05
Adding functionality to set same token value across cluster

shivamdurgbuns:
This is for Vault Configuration.
fixes: https://issues.redhat.com/browse/OCSQE-1636

### DIFF
--- a/ocs_ci/utility/kms.py
+++ b/ocs_ci/utility/kms.py
@@ -139,8 +139,10 @@ class Vault(KMS):
     def get_vault_token(self):
         """
         Get the Vault token value. Generate it first if it does not already exist.
+
         Returns:
             str: Vault token
+
         """
         if self.vault_path_token is None:
             Vault.vault_path_token = self.generate_vault_token()
--- a/ocs_ci/utility/kms.py
+++ b/ocs_ci/utility/kms.py
@@ -100,7 +100,13 @@ class Vault(KMS):
     A class which handles deployment and other
     configs related to vault
 
+    HCP is enterprise vault, clusters are deployed on AWS platform can communicate with no error.
+    for vsphere we use our community vault
+
     """
+
+    # creating class variable (shared across all instances)
+    vault_path_token = None
 
     def __init__(self):
         super().__init__("vault")
@@ -123,12 +129,22 @@ class Vault(KMS):
         )
         self.kmsid = None
         # Base64 encoded (with padding) token
-        self.vault_path_token = None
+        self.vault_path_token = Vault.vault_path_token
         self.vault_policy_name = None
         self.vault_kube_auth_path = "kubernetes"
         self.vault_kube_auth_role = constants.VAULT_KUBERNETES_AUTH_ROLE
         self.vault_kube_auth_namespace = None
         self.vault_cwd_kms_sa_name = constants.VAULT_CWD_KMS_SA_NAME
+
+    def get_vault_token(self):
+        """
+        Get the Vault token value. Generate it first if it does not already exist.
+        Returns:
+            str: Vault token
+        """
+        if self.vault_path_token is None:
+            Vault.vault_path_token = self.generate_vault_token()
+        return Vault.vault_path_token
 
     def deploy(self):
         """
@@ -578,7 +594,7 @@ class Vault(KMS):
             raise VaultOperationError(
                 f"Failed to create policy f{self.vault_policy_name}"
             )
-        self.vault_path_token = self.generate_vault_token()
+        self.vault_path_token = self.get_vault_token()
 
     def generate_vault_token(self):
         """
--- a/ocs_ci/utility/kms.py
+++ b/ocs_ci/utility/kms.py
@@ -106,7 +106,7 @@ class Vault(KMS):
     """
 
     # creating class variable (shared across all instances)
-    vault_path_token = None
+    _vault_path_token = None
 
     def __init__(self):
         super().__init__("vault")
@@ -128,25 +128,18 @@ class Vault(KMS):
             "VAULT_BACKEND", defaults.VAULT_DEFAULT_BACKEND_VERSION
         )
         self.kmsid = None
-        # Base64 encoded (with padding) token
-        self.vault_path_token = Vault.vault_path_token
         self.vault_policy_name = None
         self.vault_kube_auth_path = "kubernetes"
         self.vault_kube_auth_role = constants.VAULT_KUBERNETES_AUTH_ROLE
         self.vault_kube_auth_namespace = None
         self.vault_cwd_kms_sa_name = constants.VAULT_CWD_KMS_SA_NAME
 
-    def get_vault_token(self):
-        """
-        Get the Vault token value. Generate it first if it does not already exist.
-
-        Returns:
-            str: Vault token
-
-        """
-        if self.vault_path_token is None:
-            Vault.vault_path_token = self.generate_vault_token()
-        return Vault.vault_path_token
+    @property
+    def vault_path_token(self):
+        # Base64 encoded (with padding) token
+        if Vault._vault_path_token is None:
+            Vault._vault_path_token = self.generate_vault_token()
+        return Vault._vault_path_token
 
     def deploy(self):
         """
@@ -596,7 +589,6 @@ class Vault(KMS):
             raise VaultOperationError(
                 f"Failed to create policy f{self.vault_policy_name}"
             )
-        self.vault_path_token = self.get_vault_token()
 
     def generate_vault_token(self):
         """


dahorak (approved):

petr-balogh (dismissed):

vavuthu (approved):

shylesh (dismissed):
lgtm

ocs_ci/utility/kms.py
```diff
@@ -123,13 +126,22 @@ def __init__(self):
...
 
+    @staticmethod
+    def set_vault_token():
+        """
+        This is a static method fucntion that will set the token value.
+        """
+        if Vault.vault_path_token is None:
+            Vault.vault_path_token = Vault.generate_vault_token()
+        return Vault.vault_path_token
+
```
clacroix12, line 144: I think we should use @classmethod here instead:

```suggestion
    @classmethod
    def set_vault_token(cls):
        """
        This is a class method that will set the token value.
        """
        if cls.vault_path_token is None:
            cls.vault_path_token = cls.generate_vault_token()
        return cls.vault_path_token

```

shivamdurgbuns: Great idea Coady, made the changes.

petr-balogh (dismissed):

clacroix12 (dismissed):

shivamdurgbuns:
Job ran: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/view/DR/job/qe-rdr-setup/1360/

shivamdurgbuns:
deployment job: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/view/DR/job/qe-rdr-setup/1371/

shivamdurgbuns:
deployment job: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-rdr-setup/1389/console

petr-balogh (dismissed):

shivamdurgbuns:
Verification job: https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-rdr-setup/1391/

ocs_ci/utility/kms.py
```diff
@@ -123,13 +129,22 @@ def __init__(self):
...
 
+    @staticmethod
+    def set_vault_token(self):
+        """
+        This is a static method that will set the token value.
+        """
+        if self.vault_path_token is None:
+            self.vault_path_token = self.generate_vault_token()
+        return self.vault_path_token
+
```
clacroix12, line 147: A staticmethod won't have a `self` to reference, we want to use a `classmethod` here. 

```suggestion
    @classmethod
    def set_vault_token(cls):
        """
        This is a class method that will set the token value.
        """
        if cls.vault_path_token is None:
            cls.vault_path_token = cls.generate_vault_token()
        return cls.vault_path_token

```

ocs_ci/utility/kms.py
```diff
@@ -578,7 +593,7 @@ def vault_create_policy(self, policy_name=None):
             raise VaultOperationError(
                 f"Failed to create policy f{self.vault_policy_name}"
             )
-        self.vault_path_token = self.generate_vault_token()
+        self.vault_path_token = self.set_vault_token(self)
```
clacroix12, line 596: If you go the classmethod route (which we should), you wont need to pass self here.
```suggestion
        self.vault_path_token = self.set_vault_token()
```

clacroix12 (changes requested):

ocs_ci/utility/kms.py
shivamdurgbuns: Hello, we tried the class method route which ended up causing failure in the verification job thus has to reverse back to the static method here.

shivamdurgbuns: We are calling generate token function inside set token token function that is where the problem happens. This static method is being a simple solution here.

clacroix12: Why does it need to be a staticmethod? You are circumventing it by passing a `self` reference from another method anyways.

clacroix12: If I'm understanding the issue, it seems like calling `generate_vault_token()` from inside this as a classmethod is the issue. I'm assuming here, but that is most likely because we have it defined as a standard method (with a reference to self) and we might be calling this code before there is an object instantiated. Looking at `generate_vault_token`, it doesn't actually use the `self` reference, and as such could be a `@staticmethod`, which should fix that issue. 

Since we're using a class level variable here, and calling what is effectively a static method, I think this is a perfect use case to make `set_vault_token` a classmethod.

Please let me know if there was another issue I'm not seeing, but this solve the issue of calling `generate_vault_token` from within the `set_vault_token` classmethod.

clacroix12: I took another look at the code and I think I'm understanding the issue. We are calling `generate_vault_token` which is a standard method with a reference to an instance of the class. If we are calling this class level method without having instantiated an actual object, then calling this method will fail. In this case, a class method might make things more difficult.

That being said, this should not be a `@staticmethod`. A static method is something that could be defined outside of the class and work the same, without any references to the class or an instance of the class (self). If we are going to circumvent this and manually pass a reference to the instance of the class then this should simply be a standard method with no `@staticmethod` decorator.

I will update my requested changes.

ocs_ci/utility/kms.py
```diff
@@ -123,13 +129,22 @@ def __init__(self):
...
         self.vault_cwd_kms_sa_name = constants.VAULT_CWD_KMS_SA_NAME
 
+    @staticmethod
+    def set_vault_token(self):
+        """
+        This is a static method that will set the token value.
+        """
+        if self.vault_path_token is None:
+            self.vault_path_token = self.generate_vault_token()
+        return self.vault_path_token
```
clacroix12, line 146:
```suggestion
    def get_vault_token(self):
        """
        Get the Vault token value. Generate it first if it does not already exist.

        Returns:
            str: Vault token

        """
        if self.vault_path_token is None:
            self.vault_path_token = self.generate_vault_token()
        return self.vault_path_token
```

ocs_ci/utility/kms.py
clacroix12, line 596:
```suggestion
        self.vault_path_token = self.get_vault_token()
```

clacroix12 (changes requested):
Couple things here:
 - changed `set_vault_token` from a staticmethod to a standard method since we need an instance reference.
 - changed the name to `get_vault_token` and updated the docstring since we are returning the value

ocs_ci/utility/kms.py
```diff
@@ -578,7 +594,7 @@ def vault_create_policy(self, policy_name=None):
             raise VaultOperationError(
                 f"Failed to create policy f{self.vault_policy_name}"
             )
-        self.vault_path_token = self.generate_vault_token()
+        self.vault_path_token = self.get_vault_token(self)
```
clacroix12, line 597: You don't need to explicitly pass this to the function anymore.

```suggestion
        self.vault_path_token = self.get_vault_token()
```

petr-balogh (approved):

clacroix12 (dismissed):

ocs_ci/utility/kms.py
```diff
@@ -123,13 +129,23 @@ def __init__(self):
...
         self.vault_cwd_kms_sa_name = constants.VAULT_CWD_KMS_SA_NAME
 
+    def get_vault_token(self):
+        """
+        Get the Vault token value. Generate it first if it does not already exist.
+        Returns:
+            str: Vault token
+        """
+        if self.vault_path_token is None:
+            self.vault_path_token = self.generate_vault_token()
```
petr-balogh, line 146:
```suggestion
            Vault.vault_path_token = self.generate_vault_token()
```

You need to change class variable, not instance variable here otherwise it will not work.

I checked example you shared with me and little update to prove the uuid will be same in both calls and it's not:
```python
import uuid

class Vault():
    vault_path_token = None

    def __init__(self):
        self.vault_path_token = Vault.vault_path_token

    def get_vault_token(self):
        """
        Get the Vault token value. Generate it first if it does not already exist.
        Returns:
            str: Vault token
        """
        if self.vault_path_token is None:
            self.vault_path_token = self.generate_vault_token()
        return self.vault_path_token

    def deploy(self):
        self.vault_path_token = self.get_vault_token()

    def generate_vault_token(self):
        print("inside generate_vault_token")
        # Generate a unique random value
        return str(uuid.uuid4())


v1 = Vault()
print(v1.vault_path_token)
v1.deploy()
print(v1.vault_path_token)

v2 = Vault()
v2.deploy()
print(v2.vault_path_token)
```
Output:
```
None
inside generate_vault_token
368f386a-dce3-4370-b60a-4309acf26fed
inside generate_vault_token
393d3edc-46a7-493c-b45c-8037e731924c
```

You clearly see it's different number - but we need to achieve to have same for both instances

Updated working code

```python
import uuid

class Vault():
    vault_path_token = None

    def __init__(self):
        self.vault_path_token = Vault.vault_path_token

    def get_vault_token(self):
        """
        Get the Vault token value. Generate it first if it does not already exist.
        Returns:
            str: Vault token
        """
        if self.vault_path_token is None:
            Vault.vault_path_token = self.generate_vault_token()
        return Vault.vault_path_token

    def deploy(self):
        self.vault_path_token = self.get_vault_token()

    def generate_vault_token(self):
        print("inside generate_vault_token")
        # Generate a unique random value
        return str(uuid.uuid4())


v1 = Vault()
print(v1.vault_path_token)
v1.deploy()
print(v1.vault_path_token)

v2 = Vault()
v2.deploy()
print(v2.vault_path_token)
```

Output:
```
None
inside generate_vault_token
ba5e5c53-f49c-4fb4-b5c4-22c4f945ec1d
ba5e5c53-f49c-4fb4-b5c4-22c4f945ec1d
```
You see we have same here.

petr-balogh (changes requested):

ocs_ci/utility/kms.py
shivamdurgbuns: thanks for proof check, udpated

ocs_ci/utility/kms.py
```diff
@@ -123,13 +129,23 @@ def __init__(self):
...
         self.vault_policy_name = None
         self.vault_kube_auth_path = "kubernetes"
         self.vault_kube_auth_role = constants.VAULT_KUBERNETES_AUTH_ROLE
         self.vault_kube_auth_namespace = None
         self.vault_cwd_kms_sa_name = constants.VAULT_CWD_KMS_SA_NAME
 
+    def get_vault_token(self):
+        """
+        Get the Vault token value. Generate it first if it does not already exist.
+        Returns:
```
vavuthu, line 142: leave empty line above Retruns

ocs_ci/utility/kms.py
```diff
@@ -123,13 +129,23 @@ def __init__(self):
...
+            str: Vault token
+        """
```
vavuthu, line 144: leave empty line above

shivamdurgbuns:
PR validation job: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-rdr-setup/1455/

petr-balogh (approved):

ocs_ci/utility/kms.py
```diff
@@ -123,13 +129,25 @@ def __init__(self):
...
+        Get the Vault token value. Generate it first if it does not already exist.
+
+        Returns:
+            str: Vault token
+
+        """
+        if self.vault_path_token is None:
+            Vault.vault_path_token = self.generate_vault_token()
+        return Vault.vault_path_token
```
dahorak, line 149: If I didn't miss anything, every time you will call `get_vault_token` from the firstly created `Vault` object, you will get new token - because you are checking it in the object parameter`self.vault_path_token` but updating it in the class variable `Vault.vault_path_token`. Is that intentional? Or did I miss something there?

dahorak: The easiest solution here, would be to save the newly generated token to both `Vault.vault_path_token` and `self.vault_path_token` directly in the `get_vault_token` method, but it would be quite ugly solution I think.

The best solution might be to make `self.vault_path_token` a property, which would check and return the `Vault.vault_path_token` and prospectively call the `self.generate_vault_token()` if needed (the code of the property will be very similar as the `get_vault_token` - except it will use `Vault.vault_path_token` everywhere and the `get_vault_token` could be removed.)

Something like this:

```python
@property
def vault_path_token(self):
    if Vault.vault_path_token is None:
        Vault.vault_path_token = self.generate_vault_token()
    return Vault.vault_path_token
```

and prospectively:
```python
@vault_path_token.setter
def vault_path_token(self, value):
    Vault.vault_path_token = value
```
but the setter might not be needed.

And you will have to remove the line ~599:
```
self.vault_path_token = self.get_vault_token()
```

dahorak (changes requested):
As mentioned above, the existing `get_vault_token` method will not work consistently and in some circumstances it might generate new value for for each execution.

petr-balogh (approved):

shivamdurgbuns:
Verification Job: [#1471](https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-rdr-setup/1471/)

dahorak (approved):

PR #10836, opened 2024-11-11
install MetalLB as part of dependencies

dahorak:
Add possibility to install and configure MetalLB operator independently on ODF/hosted clusters deployment. This is required for Provider-Client pipeline where are all dependencies installed as part of separated stage.

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -611,6 +611,15 @@ class Deployment(object):
         ):
             CNVInstaller().deploy_cnv()
 
+    def do_deploy_metallb(self):
+        """
+        Deploy MetalLB
+        """
+        if config.DEPLOYMENT.get("metallb_operator"):
+            from ocs_ci.deployment.metallb import MetalLBInstaller
+
+            MetalLBInstaller().deploy_lb()
+
     def do_deploy_hosted_clusters(self):
         """
         Deploy Hosted cluster(s)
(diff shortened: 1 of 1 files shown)


ocs_ci/deployment/deployment.py
```diff
@@ -611,6 +611,15 @@ def do_deploy_cnv(self):
         ):
             CNVInstaller().deploy_cnv()
 
+    def do_deploy_metallb(self):
+        """
+        Deploy MetalLB
+        """
+        if config.DEPLOYMENT.get("metallb_operator"):
+            from ocs_ci.deployment.metallb import MetalLBInstaller
```
vavuthu, line 619: can we move it to import section or if there is any circular dependency issue, please add a comment

dahorak:
Verification job: https://url.corp.redhat.com/42b231f

ocs_ci/deployment/deployment.py
dahorak: I'll check that, I've placed it there similarly as it is in the following method `do_deploy_hosted_clusters`,  but I'll check if there actually is some circular dependency issue and update accordingly.

dahorak: Fixed by [33b3d46](https://github.com/red-hat-storage/ocs-ci/pull/10836/commits/33b3d4647eebb8d73d04f25b83c1a6b76c9a6b81)

vavuthu (approved):

PR #10862, opened 2024-11-14
Adding test automation for reclaimspace disable operation

paraggit:
Adding test automation for reclaimspace disable operation

        Steps:
            1.	Create a PVC from the CephBlockPool and attach it to a pod.
            2.	To disable the reclaimspace operation for the PVC, set suspend = true and change the state annotation
                to unmanaged.
            3.	Verify that the reclaimspace operation is disabled for the PVC by checking that suspend = true is set.
            4.	Re-enable the reclaimspace cronjob by removing suspend = true and changing the state annotation back
                to managed for each PVC.
            5.	Verify that all reclaimspace cronjobs are enabled.

### DIFF
--- a/tests/functional/pv/space_reclaim/test_disable_reclaimspace_operation.py
+++ b/tests/functional/pv/space_reclaim/test_disable_reclaimspace_operation.py
@@ -1,0 +1,137 @@
+import logging
+import pytest
+import math
+from ocs_ci.framework.pytest_customization.marks import green_squad
+from ocs_ci.ocs import constants
+from ocs_ci.framework.testlib import tier1
+from ocs_ci.helpers.helpers import (
+    change_reclaimspacecronjob_state_for_pvc,
+    get_rbd_image_info,
+    create_pods,
+    verify_reclaimspacecronjob_suspend_state_for_pvc,
+)
+from ocs_ci.ocs.exceptions import UnexpectedBehaviour
+from ocs_ci.utility.retry import retry
+from ocs_ci.ocs.resources.pod import delete_pods
+
+log = logging.getLogger(__name__)
+
+
+@green_squad
+@tier1
+class TestDisableReclaimSpaceOperation:
+    @pytest.fixture(autouse=True)
+    def setup(self, storageclass_factory, multi_pvc_factory):
+        """Setup the test environment by creating StorageClass and PVCs."""
+        reclaimspace_annotations = {
+            "reclaimspace.csiaddons.openshift.io/schedule": "* * * * *"
+        }
+        self.sc_obj = storageclass_factory(
+            interface=constants.CEPHBLOCKPOOL, annotations=reclaimspace_annotations
+        )
+        self.pvc_objs = multi_pvc_factory(
+            size=5,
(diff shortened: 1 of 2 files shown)


tests/functional/pv/space_reclaim/test_disable_reclaimspace_operation.py
```diff
@@ -0,0 +1,76 @@
...
+        for pvc in pvc_objs:
+            pod_obj = pod_factory(
+                pvc=pvc,
+                status=constants.STATUS_RUNNING,
+                raw_block_pv=True,
+            )
+            self.pod_objs.append(pod_obj)
+
+        # Disable Reclaimspace operation for the PVC
+        log.info("Changing reclaimspace cronjob status to disable for all PVC object.")
```
shyRozen, line 53: Maybe I'm missing something but if we want to disable the reclaim space, don't we need to enable it first?
You are disabling but then enabling. I think it should go the other way around.
Enable->disable and then you can enable it back if we want to see that it can be reversed again to enable.

tests/functional/pv/space_reclaim/test_disable_reclaimspace_operation.py
```diff
@@ -0,0 +1,76 @@
...
+            change_reclaimspacecronjob_state_for_pvc(pvc, enable=True)
+
+        # Verify Reclaimnspace cronjob is enable for the PVC
+        for pvc in pvc_objs:
+            reclaimspace_cronjob = get_reclaimspacecronjob_for_pvc(pvc)
+            suspend_state = reclaimspace_cronjob.data["spec"].get("suspend", False)
+
+            assert (
+                not suspend_state
+            ), "reclaimspace cronjob state is still in suspened state"
```
shyRozen, line 76: Please add the pvc name and the suspend state we got.

tests/functional/pv/space_reclaim/test_disable_reclaimspace_operation.py
paraggit: Reclaimspace is enabled by default on the cluster, so the flow starts with disabling and then enabling it to test reversibility.

tests/functional/pv/space_reclaim/test_disable_reclaimspace_operation.py
paraggit: Done

tests/functional/pv/space_reclaim/test_disable_reclaimspace_operation.py
shyRozen, line 1: Looking at the test it only checks configuration but there is no validation if the reclaim has stopped happening. Is there a way to expand the testcase to really check it the functionality has stopped working? Something like fill the PVC with data that will be seen in the whole storage capacity and delete the data within he PVC and see that the space in not reclaimed like it should be when reclaim space is enabled. From your experience how much time it takes for space to be reclaimed?

ebenahar (dismissed):

shyRozen (approved):

ebenahar (approved):

PR #10867, opened 2024-11-15
Use Let's Encrypt for certificate generation

dahorak:
Add possibility to sign custom ingress (and API) certificates by Let's Encrypt (using `certbot`).

This PR contains following main changes:
* split existing `OCSCertificate` class to generic `Certificate` and specific `OCSCertificate`
* implement `LetsEncryptCertificate` class
* update logic for generating new certificate to select proper class based on `custom_ssl_cert_provider` configuration parameter

For now this functionality is available only for clusters where DNS records are handled by AWS Route53 (those which are directly in domain `qe.rh-ocs.com`, which means platforms: AWS, vSphere and Baremetal).


dahorak:
Verification job: https://url.corp.redhat.com/6d6ee9e
The custom ingress certificate is correctly signed by Let's Encrypt.

petr-balogh (approved):

dahorak:
In the verification job, OCP&ODF deployment passed, but there are few failures in acceptance test suite. The failures are surely related to this new feature of using Let's Encrypt certificate, but since it will not be heavily used, I would consider fixing the test failures as lower priority and will handle it in separated PR later, if needed.

dahorak:
Update the PR to use certificate with full chain path, this will also hopefully solve the issue in some acceptance tests.
Verification job: https://url.corp.redhat.com/8268852

vavuthu (approved):

petr-balogh (approved):

PR #10876, opened 2024-11-18
Update disk cleanup to remove metadata at different locations

jilju:
With 4.18, bluestore replicates its metadata at multiple places on the device (at 0 / 1Gb / 10Gb / 100Gb / 1000Gb etc).
This has to be cleaned up before installing any other version of ODF once 4.18 is installed on the same machines.


jilju:
Verified on a setup that previously failed to do the additional metadata : https://ocs4-jenkins-csb-odf-qe.apps.ocp-c1.prod.psi.redhat.com/job/qe-deploy-ocs-cluster/44126/

dahorak (approved):

amr1ta (approved):
Looking good

PR #10886, opened 2024-11-19
Updated local storage template to accept different version

amr1ta:
This pr is about, updating the optional-operator catalogsource with different versioons. Earlier it had fixed value.

### DIFF
--- a/ocs_ci/deployment/helpers/lso_helpers.py
+++ b/ocs_ci/deployment/helpers/lso_helpers.py
@@ -261,9 +261,12 @@ def create_optional_operators_catalogsource_non_ga(force=False):
     if config.DEPLOYMENT.get("optional_operators_image"):
         for _dict in optional_operators_data:
             if _dict.get("kind").lower() == "catalogsource":
-                _dict["spec"]["image"] = config.DEPLOYMENT.get(
-                    "optional_operators_image"
-                )
+                if not config.DEPLOYMENT.get("optional_operators_image"):
+                    _dict["spec"]["image"] = ocp_ga_version
+                else:
+                    _dict["spec"]["image"] = config.DEPLOYMENT.get(
+                        "optional_operators_image"
+                    )
     if config.DEPLOYMENT.get("disconnected"):
         # in case of disconnected environment, we have to mirror all the
         # optional_operators images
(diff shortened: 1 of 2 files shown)


fbalak (approved):

dahorak (approved):

dahorak (approved):

ocs_ci/deployment/helpers/lso_helpers.py
```diff
@@ -256,9 +256,12 @@ def create_optional_operators_catalogsource_non_ga(force=False):
     if config.DEPLOYMENT.get("optional_operators_image"):
         for _dict in optional_operators_data:
             if _dict.get("kind").lower() == "catalogsource":
-                _dict["spec"]["image"] = config.DEPLOYMENT.get(
-                    "optional_operators_image"
-                )
+                if not config.DEPLOYMENT.get("optional_operators_image"):
```
dahorak, line 259: This "negative" condition will never be met - since the whole block is in `if config.DEPLOYMENT.get("optional_operators_image"):` condition.

ocs_ci/deployment/helpers/lso_helpers.py
```diff
@@ -256,9 +256,12 @@ def create_optional_operators_catalogsource_non_ga(force=False):
...
+                    _dict["spec"]["image"] = ocp_ga_version
```
dahorak, line 260: `ocp_ga_version` variable will probably contain just version string, not full image url

dahorak (changes requested):

ocs_ci/deployment/helpers/lso_helpers.py
amr1ta: Have addressed this in latest commit

ocs_ci/deployment/helpers/lso_helpers.py
amr1ta: yes it is value like 4.19. which is used in option-operator catalogsource creation for the image details.
like,
`image: quay.io/openshift-qe-optional-operators/aosqe-index:v4.19`

suchita-g (approved):

DanielOsypenko (approved):

dahorak (approved):

PR #10609, opened 2024-10-01
ipv6 single stack deployment

shyRozen:
ipv6 single stack deployment of ODF.
Most of the changes are in openshift installer git for terraform and ipam.
[https://gitlab.cee.redhat.com/srozen/installer_ipv6](https://gitlab.cee.redhat.com/srozen/installer_ipv6)
[https://gitlab.cee.redhat.com/srozen/ipv6_script](https://gitlab.cee.redhat.com/srozen/ipv6_script
)

PR #10890, opened 2024-11-19
[release-4.16] ipv6 single stack deployment (#10609)

shyRozen:
Cherry pick to 4.16
ipv6 single stack deployment of ODF.
Most of the changes are in openshift installer git for terraform and ipam.
https://gitlab.cee.redhat.com/srozen/installer_ipv6
https://gitlab.cee.redhat.com/srozen/ipv6_script

### DIFF
--- a/ocs_ci/deployment/vmware.py
+++ b/ocs_ci/deployment/vmware.py
@@ -6,6 +6,7 @@ import glob
 import json
 import logging
 import os
+import gzip
 from shutil import rmtree, copyfile
 from subprocess import TimeoutExpired
 import time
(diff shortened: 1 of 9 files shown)


OdedViner (approved):

ocs_ci/deployment/vmware.py
```diff
@@ -2046,10 +2080,17 @@ def clone_openshift_installer():
                     branch="release-4.12",
                 )
             else:
+                if config.DEPLOYMENT.get("ipv6"):
+                    constants.VSPHERE_INSTALLER_REPO = (
+                        "https://gitlab.cee.redhat.com/srozen/installer_ipv6.git"
```
petr-balogh, line 2085: As discussed offline, I prefer to use public repo from github.
Also maintain the same branches per release and not hardcode master should be a case here as we know from the past that there were some changes between OCP versions.

As it's just cherry-pick I am not requesting changes now in this PR , but will be good to do changes in some follow up and cherry-pick to all versions where we did cherry-picked this change.

petr-balogh (approved):

Issue #10859, opened 2024-11-14
In the test "test_rolling_nodes_restart" increase timeout waiting for the ceph health


yitzhak12:
See the failed test at https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/632/25507/1239361/1239400/log. This issue does not happen usually, but it would be better to increase the timeout at least for vSphere and IBMCloud managed platforms.

Issue #10834, opened 2024-11-10
In the test "test_rolling_terminate_and_recovery_in_controlled_fashion_ipi" the rook-ceph-tools pod failed to recover after worker node terminate and recovery with vSphere IPI


yitzhak12:
In the test "test_rolling_terminate_and_recovery_in_controlled_fashion_ipi," the ocs pods failed to recover after the worker node terminated and recovered. See the failed test here: https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/738/26643/1314478/1314530/log.

This test worked fine in the past. However, it seems to happen only with vSphere IPI and not other platforms. We need to investigate the issue to determine whether it's a bug or not.

yitzhak12:
After examining the failure again and looking at the logs: http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/j-074vi1cs33-t4b/j-074vi1cs33-t4b_20241105T115511/logs/failed_testcase_ocs_logs_1730830538/test_rolling_terminate_and_recovery_in_controlled_fashion_ipi_ocs_logs/j-074vi1cs33-t4b/ocs_must_gather/quay-io-rhceph-dev-ocs-must-gather-sha256-2d3fc39bc479fc5d0886f894b266ac0ed90464b17d9c85bd69f24f70b10e7e98/namespaces/openshift-storage/oc_output/all_-o_wide, I observed that all the pods were Running except the tools pod. 
Restarting the tool pod may have solved the problem, but I'm not sure. We need to determine whether it's a bug or not.

Issue #10833, opened 2024-11-10
In the test "test_rolling_nodes_restart" the rook-ceph-tool pod was unreachable after restarting the worker nodes


yitzhak12:
In the test "test_rolling_nodes_restart", the rook-ceph-tool pod was unreachable after restarting the worker nodes. 
See the test failure here: https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/738/25621/1242438/1242478/log.
I see that the error occurred because the rook-ceph-tool pod should restart after deleting the associated worker node.
This is probably a bug. One of the ways to fix it is to restart the rook-ceph-tool pod manually in the test and maybe other pods like the mds pods. See the logs: http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/j-072vi1cs33-t4b/j-072vi1cs33-t4b_20241008T034921/logs/failed_testcase_ocs_logs_1728379145/test_rolling_nodes_restart_ocs_logs/j-072vi1cs33-t4b/ocs_must_gather/quay-io-rhceph-dev-ocs-must-gather-sha256-587ef87b87a242f18a9a93e9951e034d02639e90823570edba21dfedae3a8ca5/namespaces/openshift-storage/oc_output/all_-o_wide.

PR #10915, opened 2024-11-24
Wait for the ceph command to execute successfully in the rolling nodes terminate and rolling nodes restart tests

yitzhak12:
The PR fixes the following issues:
- fix https://github.com/red-hat-storage/ocs-ci/issues/10859
- fix https://github.com/red-hat-storage/ocs-ci/issues/10834
- fix https://github.com/red-hat-storage/ocs-ci/issues/10833

### DIFF
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -757,7 +757,9 @@ def get_all_pods(
     return pod_objs
 
 
-def get_ceph_tools_pod(skip_creating_pod=False, wait=False, namespace=None):
+def get_ceph_tools_pod(
+    skip_creating_pod=False, wait=False, namespace=None, get_running_pods=True
+):
     """
     Get the Ceph tools pod
 
@@ -766,6 +768,8 @@ def get_ceph_tools_pod(skip_creating_pod=False, wait=False, namespace=None):
             if it doesn't exist
         wait (bool): True if you want to wait for the tool pods to be Running
         namespace: Namespace of OCS
+        get_running_pods (bool): If True, get only the ceph tool pods in a Running status.
+            If False, get the ceph tool pods even if they are not in a Running status.
 
     Returns:
         Pod object: The Ceph tools pod object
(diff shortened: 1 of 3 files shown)


amr1ta (dismissed):
Looking good

ocs_ci/ocs/resources/pod.py
```diff
@@ -3955,3 +3956,66 @@ def get_prometheus_pods(
     pods_with_label_match = get_pods_having_label(prometheus_label, namespace)
     prometheus_pod_objs = [Pod(**prometheus) for prometheus in pods_with_label_match]
     return prometheus_pod_objs
+
+
+def get_ceph_tool_pods(ceph_tool_label=constants.TOOL_APP_LABEL, namespace=None):
```
ebenahar, line 3961: similar function already exists - https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/resources/pod.py#L760

yitzhak12: Yes, I will change the name prefix. The problem with the function you mentioned is that it performs extra conditions and does not just get all the ceph tool pods.

AviadP (dismissed):

ocs_ci/ocs/resources/pod.py
ebenahar: why not fixing/adjusting the existing function? Having 2 functions for similar purpose isn't ideal

yitzhak12: We can do that. The problem is that we must be more careful as this function is widely used in our tests.

ebenahar: yes, changing it would require extensive validation. This doesn't mean we should have a dup function which does similar operation

yitzhak12: Okay, I will make the changes that are needed.

ebenahar (approved):

AviadP (approved):

PR #10916, opened 2024-11-24
WIP IPI disk removal when cluster is destroyed

shyRozen:
It seems when we do destroy IPI cluster we use openshift-install destroy cluster which leaves all vmdk files in the datastore. 
This PR should resolve the problem.

### DIFF
--- a/ocs_ci/deployment/vmware.py
+++ b/ocs_ci/deployment/vmware.py
@@ -1604,6 +1604,14 @@ class VSPHEREIPI(VSPHEREBASE):
             template_folder = get_infra_id(self.cluster_path)
         else:
             logger.warning("metadata.json file doesn't exist.")
+        vsphere = VSPHERE(
+            config.ENV_DATA["vsphere_server"],
+            config.ENV_DATA["vsphere_user"],
+            config.ENV_DATA["vsphere_password"],
+        )
+        all_vms = vsphere.get_vms_by_string(config.ENV_DATA["cluster_name"])
+        for vm in all_vms:
+            vsphere.remove_disks_with_main_disk(vm)
 
         try:
             run_cmd(
--- a/ocs_ci/utility/vsphere.py
+++ b/ocs_ci/utility/vsphere.py
@@ -818,6 +818,36 @@ class VSPHERE(object):
             if hasattr(device.backing, "fileName") and device.unitNumber != 0
         ]
 
+    def remove_disks_with_main_disk(self, vm):
+        """
+        Removes all disks for a VM
+
+        Args:
+            vm (vim.VirtualMachine): VM instance
+
+        """
+        extra_disk_unit_numbers = self.get_used_unit_number_with_all_unit_number(vm)
+        if extra_disk_unit_numbers:
+            for each_disk_unit_number in extra_disk_unit_numbers:
+                self.remove_disk(vm=vm, identifier=each_disk_unit_number)
+
+    def get_used_unit_number_with_all_unit_number(self, vm):
+        """
+        Gets the used unit numbers including main disk for a VM
+
+        Args:
+            vm (vim.VirtualMachine): VM instance
+
+        Returns:
+            list: list of unit numbers
+
+        """
+        return [
+            device.unitNumber
+            for device in vm.config.hardware.device
+            if hasattr(device.backing, "fileName")
+        ]
+
     def check_folder_exists(self, name, cluster, dc):
         """
         Checks whether folder exists in Templates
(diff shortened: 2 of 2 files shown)


vavuthu (dismissed):

ocs_ci/deployment/vmware.py
```diff
@@ -1626,6 +1626,15 @@ def destroy_cluster(self, log_level="DEBUG"):
...
         else:
             logger.warning("metadata.json file doesn't exist.")
+        vsphere = VSPHERE(
+            config.ENV_DATA["vsphere_server"],
+            config.ENV_DATA["vsphere_user"],
+            config.ENV_DATA["vsphere_password"],
+        )
+        all_vms = vsphere.get_vms_by_string(config.ENV_DATA["cluster_name"])
+        vsphere.stop_vms(all_vms)
+        for vm in all_vms:
```
ebenahar, line 1636: I think we should proceed with cluster teardown even in cases when disk removal fails. @shyRozen @vavuthu what do you think?

shyRozen: @ebenahar We can do that but there could be situation where un-associated objects will be created, and if we will not be aware of the issue of disks not being deleted, we can end up with the situation we have today.
If we really want to proceed with the destroy we can think of a way of knowing that a problem occurred like mail at least. Maybe we can set config flag to fail it at the end of the destroy and then the destroy will continue but will notify failed and will raise awareness.

ebenahar: an email sounds like a good idea. 
In addition, we can still consider running a user object cleanup job periodically, to collect all those leftovers

vavuthu (dismissed):

ocs_ci/deployment/vmware.py
```diff
@@ -1626,6 +1627,22 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+        vsphere.stop_vms(all_vms)
+        try:
+            for vm in all_vms:
+                vsphere.remove_disks_with_main_disk(vm)
+        except Exception as e:
+            # Failing the test but letting it to continue
+            try:
+                fail(f"Removing disks in destroy fail with the error {e}")
+            except:
+                pass
```
petr-balogh, line 1645: I would proceed with it different way to make sure it's removed for all  VMs - imagine it will fail in first VM but rest is untouched because of this error, you need to move try block into the for loop.

```suggestion
            vm_with_disk_removal_failure = []
            for vm in all_vms:
                try:
                     vsphere.remove_disks_with_main_disk(vm)
                except Exception as e:
                    # Failing the test but letting it to continue
                    vm_with_disk_removal_failure.append(vm)
                    logger.exception("Removing disks in destroy fail with the error")
```

Continue with the rest of the code for destroy - and if we still insist to fail the destroy test we should add some condition at the end of this function like:

```python
if vm_with_disk_removal_failure:
    pytest.fail(f"We have failed to remove disks for those VMs: {vm_with_disk_removal_failure}")
```

Maybe it needs to print just names with something like [vm.name for vm in vm_with_disk_removal_failure] - not sure how object of VM is represented in string.

Anyway, for teardown we are trying to be robust as much as possible - there are a lot of cases - the cluster is not properly even installed - and we run destroy on it - so not sure if this function handles this situation as well - it should.

We do not send emails of the teardown failure as no --email param is set in our jenkins jobs I think.

Also what we need to consider is that destroy job has 3 attempts - first from branch it was triggered  (e.g. pr validation) - second attempt - stable/release branch - as PR could break it. And third attempt is vsphere cleanup script - I think that if we just succeed in one of those destroy attempts- the deploy job is still considered as unstable and we don't send any email as we consider it was cleaned up either in one of those attempts. Email about teardown failure is only if even 3rd force vsphere-cleanup script is not success.


This was intentionally - as it bring a lot of overhead for ecosystem team to deal with all those issues we had for not successfully deployed clusters and so on. 


**So what will anyway happen here is this:**

1. First attempt will destroy the cluster - but might fail at the end
2. Second attempt will try to run - but because cluster will be already destroyed - there is no cluster on vsphere - no metadata.json file as cluster is destroyed - it will consider it as successful destroy 
3. Third attempt will not even trigger if we don't from some reason have failure in 2.nd attempt
4. it will consider destroy as unstable - but as all leftovers were deleted - at least it thinks of it as they were - it will not notify us.

So in this case I don't think it even make sense to deal with such try except and failing it

We can produce a file in log directory with such reported issues - and from jenkins job we can make a condition if this fail will exist we can send an email from it - which will work in this situation but will require also update from jenkins job.

ebenahar (approved):

vavuthu (approved):

petr-balogh (approved):

petr-balogh:
@shyRozen if we need cherry-picks please create them. Thanks

PR #10938, opened 2024-11-27
Fixes bucket policy NotAction issue

### LINKED ISSUE #9968: test_bucket_policy_elements_NotAction fails because correct expectations are not kept for effect=Allow
We expect the delete operation to fail at:

```python
        if effect == "Allow":
            # Put Object is allowed
            logger.info("Writing index data to the bucket")
            assert s3_put_object(
                s3_obj=obc_obj,
                bucketname=obc_obj.bucket_name,
                object_key="index.html",
                data=index,
                content_type="text/html",
            ), "Failed to put object."

            # Delete bucket get access denied.
            logger.info(f"Deleting bucket {obc_obj.bucket_name}")
            try:
                s3_delete_bucket_website(s3_obj=obc_obj, bucketname=obc_obj.bucket_name)
                raise UnexpectedBehaviour(
                    "Failed: Bucket got deleted, expect to get AccessDenied."
                )
``` 
This expectation doesnt seem to be correct as per my understanding of the NotAction. NotAction parameters will exclude the action from the effect. So in this case, by default bucket can be deleted. By excluding DeleteBucket from the action list with effect=Allow doesnt make it to have deny affect!

This needs more analysis.

udaysk23:
Fixes #9968


sagihirshfeld (approved):
LGTM

ypersky1980 (approved):
LGTM

mashetty330 (approved):

nishtha-mishra:
This fix is not working on HCP Cluster where ODF is installed.In release-4.17 branch I don't see this updated parameter     obc_bucket = bucket_factory(amount=1, interface="OC") .

PR #11003, opened 2024-12-08
Rename deprecated upgrade tests

AviadP:
Rename deprecated upgrade tests


suchita-g (approved):

AviadP:
we are not going to deprecate these test, but find another way to run them. closing it

PR #11015, opened 2024-12-09
Improve getting vrg name for Discovered Apps

prsurve:
If we have multiple vrg in openshift-dr-ops earlier, ocs-ci was picking up 0th index vrg but now it will pick the right one


prsurve:
Verifed 

==================================================================================================== 1 passed, 16 warnings in 2084.35s (0:34:44) ====================================================================================================

PR #11083, opened 2024-12-29
Testing new Object Storage Page[WIP]

AviadP:
4.18 new feature

### DIFF
--- a/ocs_ci/ocs/ui/page_objects/buckets_tab.py
+++ b/ocs_ci/ocs/ui/page_objects/buckets_tab.py
@@ -13,6 +13,36 @@ class BucketsTab(ObjectStorage, ConfirmDialog):
     def __init__(self):
         super().__init__()
 
+    def create_bucket_ui(self, method):
+        locators = {
+            "create_bucket_button": ("//*[@id='yaml-create']", By.XPATH),
+        }
+        self.do_click(locators["create_bucket_button"])
+        if method == "obc":
+            self.create_bucket_via_obc()
+        elif method == "s3":
+            self.create_bucket_via_s3()
+        else:
+            raise ValueError(f"Invalid method: {method}")
+
+    def create_bucket_via_obc(self):
+        locators = {
+            "create_bucket_button": (
+                "//*[@id='content-scrollable']/section/div[2]/div[1]/div/div[2]/div[1]/div[2]",
+                By.XPATH,
+            ),
+        }
+        self.do_click(locators["create_bucket_button"])
+
+    def create_bucket_via_s3(self):
+        locators = {
+            "create_bucket_button": (
+                "//*[@id='content-scrollable']/section/div[2]/div/div/div[2]/div[2]/div[2]",
+                By.XPATH,
+            ),
+        }
+        self.do_click(locators["create_bucket_button"])
+
     def delete_bucket_ui(self, delete_via, expect_fail, resource_name):
         """
         Delete an Object Bucket via the UI
--- a/ocs_ci/ocs/ui/page_objects/buckets_tab.py
+++ b/ocs_ci/ocs/ui/page_objects/buckets_tab.py
@@ -1,12 +1,16 @@
 import json
-
+import logging
 import requests
+
 from selenium.webdriver.common.by import By
+
 from ocs_ci.ocs.ocp import get_ocp_url
 from ocs_ci.ocs import exceptions
 from ocs_ci.ocs.ui.page_objects.confirm_dialog import ConfirmDialog
-from ocs_ci.ocs.ui.page_objects.object_storage import ObjectStorage, logger
+from ocs_ci.ocs.ui.page_objects.object_storage import ObjectStorage
 from ocs_ci.utility import version
+
+logger = logging.getLogger(__name__)
 
 
 class BucketsTab(ObjectStorage, ConfirmDialog):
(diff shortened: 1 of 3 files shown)


ocs_ci/ocs/ui/page_objects/buckets_tab.py
```diff
@@ -1,18 +1,146 @@
...
+logger = logging.getLogger(__name__)
+
 
 class BucketsTab(ObjectStorage, ConfirmDialog):
+    """
+    A class representation for abstraction of Buckets tab related OpenShift UI actions
+    """
+
     def __init__(self):
         super().__init__()
```
DanielOsypenko, line 28: since we do serious changes here, can you pls along this remove `super().__init__()` from the constructor? it is unnecessary and takes time to instantiate parents up to BaseUI.

ocs_ci/ocs/ui/page_objects/buckets_tab.py
```diff
@@ -1,18 +1,146 @@
...
+        if not folder_name:
+            folder_name = f"test-folder-{uuid.uuid4()}"
+
+        self.do_click(self.bucket_tab["create_folder_button"])
+        self.do_send_keys(self.bucket_tab["folder_name_input"], folder_name)
+        self.do_click(self.bucket_tab["submit_button_folder"])
+
+        folder_path = self.generate_folder_with_file()
+
+        time.sleep(2)  # Wait for folder navigation
```
DanielOsypenko, line 104: ideally we try to find wait-and-check solutions if we depend on timing

ocs_ci/ocs/ui/page_objects/buckets_tab.py
```diff
@@ -1,18 +1,146 @@
...
+                arguments[0].removeAttribute('webkitdirectory');
+                arguments[0].removeAttribute('directory');
+                arguments[0].style.display = 'block';
+                arguments[0].style.visibility = 'visible';
+                arguments[0].style.height = '1px';
+                arguments[0].style.width = '1px';
+                arguments[0].style.opacity = '1';
+            """,
+                file_input,
+            )
```
DanielOsypenko, line 127: we don't use js scripts. It is not checking UX, because all the web elements may be hidden or overlayed but still be available for script.
remove `webkitdirectory` changes the behavior of the product, s we check not management-console, but modified management-console. 
arguments[0].removeAttribute('directory') same as above; this way we make input element accept files, not directories. 
Question: is something in UI hidden or not interactable via locators that forces us to use this script?

ocs_ci/ocs/ui/views.py
```diff
@@ -1989,7 +1989,61 @@
...
+        By.CSS_SELECTOR,
+    ),
+    "bucket_name_input": (
+        "[data-test='bucket-name']",
+        By.CSS_SELECTOR,
+    ),
+    "submit_button_obc": (
+        "[data-test='obc-create']",
+        By.CSS_SELECTOR,
+    ),
```
DanielOsypenko, line 2016: apart from these locators: [data-test='bucket-name'], [data-test='obc-create'], //*[@id='yaml-create'], //button[contains(text(), 'Upload')]" that are unique other locators cant be considered good. HTML changes and even slight change of the DOM will shift indexes, relation between css elements, etc.

We need to find unique locators, not using indexes of common elements. Also this relation with multiple parents and childen elements`]/section/div[2]/div[1]/div/div[2]/div[1]/div[2]` may change even without changes from devs, but with different user inputs. 

In opposite, rigid locators, like "//*[@id='yaml-create'] or [data-test='bucket-name'] will continue work even with changes on page from development.

tests/functional/object/mcg/ui/test_mcg_ui.py
```diff
@@ -414,3 +414,17 @@ def test_obc_creation_and_deletion(
         obc_ui_obj.delete_obc_ui(obc_name, delete_via)
 
         assert test_obc.check_resource_existence(should_exist=False)
+
+
+@black_squad
+class TestBucketCreate:
+    def test_bucket_create(self, setup_ui_class_factory):
+        setup_ui_class_factory()
+        PageNavigator().nav_object_storage_page()
```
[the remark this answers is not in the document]
DanielOsypenko: PageNavigator().navigate_buckets_page() returns BucketsTab instance. 
With Fluent interface we can use it like this if we don't want to use breadcrumbs or other new elements of the object browser/ocp console:

```
page_navigator = PageNavigator()
bucket_ui = PageNavigator().navigate_buckets_page()

bucket_ui.create_bucket_ui("obc")
page_navigator.nav_bucket_class_tab() 
bucket_ui.create_bucket_ui("s3")
etc
```

this is not much navigation happen here, it most beneficial when we navigate and do many actions, but even here it simplifies and saves execution time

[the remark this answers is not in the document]
DanielOsypenko: create_bucket_ui should have return another Page Object with its own navigation. Then Fluent interface can continue with navigation, validation and creation

ocs_ci/ocs/ui/page_objects/buckets_tab.py
```diff
@@ -1,18 +1,146 @@
...
+            self.create_bucket_via_s3()
+        else:
+            raise ValueError(f"Invalid method: {method}")
+
+    def create_bucket_via_obc(self):
+
+        self.do_click(self.bucket_tab["create_bucket_button_obc"])
+        self.do_click(self.bucket_tab["storage_class_dropdown"])
+        self.do_click(self.bucket_tab["storage_class_noobaa_option"])
+        self.do_click(self.bucket_tab["submit_button_obc"])
```
DanielOsypenko, line 77: please make this function return Page Object. It should return the instance of the web-page where user left after clicking `submit_button_obc`

ocs_ci/ocs/ui/page_objects/buckets_tab.py
```diff
@@ -1,18 +1,146 @@
...
+        content = "".join(
+            random.choices(string.ascii_letters + string.digits + " \n", k=100)
+        )
+
+        with open(filepath, "w") as f:
+            f.write(content)
+
+        return folder_path
+
+    def create_bucket_ui(self, method):
```
DanielOsypenko, line 62: please make this function return Page Object. It should return the instance of the web-page where user left after clicking self.create_bucket_via_obc() or self.create_bucket_via_s3()

ocs_ci/ocs/ui/page_objects/buckets_tab.py
AviadP: > Question: is something in UI hidden or not interactable via locators that forces us to use this script?

As far as i could find, yes

tests/functional/object/mcg/ui/test_mcg_ui.py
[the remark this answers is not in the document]
DanielOsypenko:
```
        PageNavigator().nav_object_storage_page()
        bucket_ui.create_bucket_ui("s3")
        PageNavigator().nav_object_storage_page()
        bucket_ui.create_folder_in_bucket()
        PageNavigator().nav_object_storage_page()
   ```
  this way we create PageNavigator 3 times. It will 3 times create BaseUI with checking folders for UI.
  
  Please create `page_navigator = PageNavigator()` in the beginning and reuse object

DanielOsypenko:
@AviadP please add test execution res, so we can add label Verified.

DanielOsypenko:
@AviadP , is this a preparation PR? because no assertions or polarion ID

AviadP:
> @AviadP , is this a preparation PR? because no assertions and polarion ID

fixed

DanielOsypenko (approved):

PR #11122, opened 2025-01-07
Config thread safe updated

petr-balogh:


petr-balogh:
I have compared results from tier1 execution with PR:
https://url.corp.redhat.com/6c47caf
And without PR:
https://url.corp.redhat.com/57f8d01

Looks like that cluster got deleted or broken somewhere in the execution but from the failures - I check one by one and it doesn't look like any of this is relevant to this PR.

I see that clearly in commands we are adding `--kubeconfig` explicitly.

I have tested also ITR execution here, where only errors are relevant to ssl, but not this change:
https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-trigger-test-pr/8637/testReport/

Acceptance suite 100 % success rate:
https://url.corp.redhat.com/6ad107f

So I consider this as verified

dahorak (approved):

ocs_ci/framework/__init__.py
```diff
@@ -587,6 +590,63 @@ def remove_cluster_by_name(self, cluster_name):
...
+            f"Thread ID: {thread_id} is using config index: {self.config_index}"
+        )
+        try:
+            super(ConfigSafeThread, self).run()
+        finally:
+            if hasattr(self.thread_local_data, "config_index"):
+                del config.thread_local_data.config_index
+
+
+def config_safe_thread_pool_task(config_index, task, *args, **kwargs):
```
DanielOsypenko, line 628: config_safe_thread_pool_task  looks really useful!

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -587,19 +587,14 @@ def process_cluster_cli_params(config):
 
     # Importing here cause once the function is invoked we have already config
     # loaded, so this is OK to import once you sure that config is loaded.
+    kubeconfig_location = ocsci_config.RUN["kubeconfig_location"]
+    custom_kubeconfig_location = get_cli_param(config, "kubeconfig")
+    ocsci_config.RUN["custom_kubeconfig_location"] = custom_kubeconfig_location
```
DanielOsypenko, line 592: in what cases do we need custom_kubeconfig that overwrites default kubeconfig?

ocs_ci/framework/__init__.py
petr-balogh: Good catch @DanielOsypenko , added return on line 644

ocs_ci/utility/utils.py
```diff
@@ -643,20 +645,17 @@ def exec_cmd(
         stderr     (str): The standard error (None if not captured).
 
     """
-    masked_cmd = mask_secrets(cmd, secrets)
-    log.info(f"Executing command: {masked_cmd}")
+    _env = kwargs.pop("env", os.environ.copy())
```
DanielOsypenko, line 648: How do you think, will this brake logic when "--kubeconfig" is provided in string of the command? 
see: https://github.com/red-hat-storage/ocs-ci/blob/c3536dfc7fb98fef4b0d042eb5c0031c24dd8bb6/ocs_ci/deployment/hosted_cluster.py#L685-L710

petr-balogh: no `--kubeconfig` cli argument has precedence - but in case we do not provide, I still want to be sure we have KUBECONFIG also in env variable just to be on safe side in case some other command we used in the past relied on env var from some reason.

Before we had anyway KUBECONFIG exported in whole framework process - so we always had it exported

vavuthu (dismissed):

DanielOsypenko (approved):

fbalak (dismissed):

DanielOsypenko (approved):

dahorak (approved):

vavuthu (approved):

PR #11141, opened 2025-01-13
Update dr_helpers and dr_workload to support VRG preservation in ODF 4.18 and AppSet workload health check

sidhant-agrawal:
This PR includes updates to two files:
 -  dr_helpers.py: Skips waiting for VRG deletion starting from ODF 4.18, while retaining the wait behavior for ODF 4.17 and below.
 - dr_workload.py: Update AppSet workload deployment verification to use the health status instead of phase


keemano (approved):

shylesh (approved):
lgtm

Issue #11223, opened 2025-01-24
throw exception in the case someone is using retry with backoff >1 and it's bigger than some reasonable time


petr-balogh:
I think we were facing multiple times the issue with infinite loop of re-tries because if backoff is bigger than 1 (default value is 2), it will each attempt increase wait time by 2 - which then grows exponentially. 

We should agree on some reasonable limit - like 10 minutes maximum to increase the delay between checks - and not allow it to be longer - as then we can have very long waiting time before next attempt which will waste time in case the operation will complete in this 10 minutes but we will be waiting e.g. 30 minutes.

Second thing is that we should also throw an exception if we will have time higher than e.g. 4 hours.

To allow it to be longer for some specific cases like upgrade operations and so on which can take a time - we could set parameter max_time to retry function set to 4 hours  like 
`def retry(exception_to_check, tries=4, delay=3, backoff=2, text_in_exception=None, func=None, max_delay=600, max_time=14400):`

And at the start we can calculate if we will be over the specified max_time and if so - raise the exception for that.

PR #11224, opened 2025-01-24
Update retry logic with some limits

petr-balogh:
Fixes: #11223

### DIFF
--- a/ocs_ci/helpers/dr_helpers.py
+++ b/ocs_ci/helpers/dr_helpers.py
@@ -1216,7 +1216,7 @@ def get_fence_state(drcluster_name, switch_ctx=None):
     return state
 
 
-@retry(UnexpectedBehaviour, tries=40, delay=5, backoff=5)
+@retry(UnexpectedBehaviour, tries=25, delay=5, backoff=5)
 def verify_fence_state(drcluster_name, state, switch_ctx=None):
     """
     Verify the specified drcluster is in expected state
@@ -1288,7 +1288,7 @@ def restore_backup():
     config.switch_ctx(restore_index)
 
 
-@retry(UnexpectedBehaviour, tries=40, delay=5, backoff=5)
+@retry(UnexpectedBehaviour, tries=25, delay=5, backoff=5)
 def verify_restore_is_completed():
     """
     Function to verify restore is completed or finished
@@ -1309,7 +1309,7 @@ def verify_restore_is_completed():
     config.switch_ctx(restore_index)
 
 
-@retry(UnexpectedBehaviour, tries=60, delay=5, backoff=2)
+@retry(UnexpectedBehaviour, tries=25, delay=5, backoff=2)
 def verify_drpolicy_cli(switch_ctx=None):
     """
     Function to verify DRPolicy status
@@ -1336,7 +1336,7 @@ def verify_drpolicy_cli(switch_ctx=None):
         )
 
 
-@retry(UnexpectedBehaviour, tries=40, delay=5, backoff=5)
+@retry(UnexpectedBehaviour, tries=25, delay=5, backoff=5)
 def verify_backup_is_taken():
     """
     Function to verify backup is taken
--- a/ocs_ci/utility/retry.py
+++ b/ocs_ci/utility/retry.py
@@ -28,7 +28,7 @@ def validate_retry_params(tries, delay, backoff, max_timeout, max_delay):
             raise ValueError(
                 f"The retry configuration exceeds the maximum allowed timeout ({max_timeout} seconds) "
                 f"with the given parameters: tries={tries}, delay={delay}, backoff={backoff}, max_delay={max_delay}."
-                f"You can reach up to: {total_time} seconds! Please fix it or adjust parameters of retry."
+                f"You reached to: {max_timeout} seconds! Please fix it or adjust parameters of retry."
             )
         current_delay = min(current_delay * backoff, max_delay)
 
@@ -50,7 +50,8 @@ def retry(
         exception_to_check: The exception to check. May be a tuple of exceptions to check.
         tries: Number of times to try (not retry) before giving up.
         delay: Initial delay between retries in seconds.
-        backoff: Backoff multiplier e.g., value of 2 will double the delay each retry.
+        backoff: Backoff multiplier e.g., value of 2 will double the delay each retry. If the delay
+            is greater than max_delay after multiplier, than max_delay is used as delay
         text_in_exception: Retry only when text_in_exception is in the text of exception.
         func: Function for garbage collector.
         max_timeout: Maximum total time for retries in seconds (default: 4 hours).
@@ -127,7 +128,8 @@ def retry_until_exception(
         exception_to_check: the exception to check. may be a tuple of exceptions to check
         tries: number of times to try (not retry) before giving up
         delay: initial delay between retries in seconds
-        backoff: backoff multiplier e.g. value of 2 will double the delay each retry
+        backoff: Backoff multiplier e.g., value of 2 will double the delay each retry. If the delay
+            is greater than max_delay after multiplier, than max_delay is used as delay
         text_in_exception: Retry only when text_in_exception is in the text of exception
         func: function for garbage collector
     """
--- a/ocs_ci/utility/retry.py
+++ b/ocs_ci/utility/retry.py
@@ -5,20 +5,59 @@ from functools import wraps
 logger = logging.getLogger(__name__)
 
 
-def retry(
-    exception_to_check, tries=4, delay=3, backoff=2, text_in_exception=None, func=None
-):
+def validate_retry_params(tries, delay, backoff, max_timeout, max_delay):
     """
-    Retry calling the decorated function using exponential backoff.
+    Validate retry parameters to ensure the total retry time does not exceed max_timeout.
 
     Args:
-        exception_to_check: the exception to check. may be a tuple of exceptions to check
-        tries: number of times to try (not retry) before giving up
-        delay: initial delay between retries in seconds
-        backoff: backoff multiplier e.g. value of 2 will double the delay each retry
-        text_in_exception: Retry only when text_in_exception is in the text of exception
-        func: function for garbage collector
+        tries: Number of retry attempts.
+        delay: Initial delay between retries in seconds.
+        backoff: Multiplicative factor for backoff.
+        max_timeout: Maximum total time for retries in seconds.
+        max_delay: Maximum delay between retries in seconds.
+
+    Raises:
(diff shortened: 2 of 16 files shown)


dahorak (dismissed):

ocs_ci/utility/retry.py
```diff
@@ -5,20 +5,59 @@
...
+
 def retry(
-    exception_to_check, tries=4, delay=3, backoff=2, text_in_exception=None, func=None
+    exception_to_check,
+    tries=4,
+    delay=3,
+    backoff=2,
+    text_in_exception=None,
+    func=None,
+    max_timeout=14400,
```
clacroix12, line 43: Allowing up to 4 hours of potential delay for one decorator by default still seems like an issue that we should try our best to reduce. Maybe default to 1 hour and explicitly set it to a higher value in use-cases where it's necessary?

petr-balogh: In this initial PR I didn't want to be that much strict as it will bring me to the situation I will need to touch much more files and potentially breaking something.  So already this high number you see already found out a lot of those bad retries. 

I am also for reducing it more - but for this - we can do in next round - where I will probably just give people list of those by changing values in script and provide output for 1 hour and then let teams to incorporate changes and then reduce time. But I don't want to be stuck with this PR myself for long time and affecting code I am not familiar with.

scripts/python/checks/validate_retry_usage.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def _validate_retry_call(self, node):
+        """Validate the arguments of a retry call."""
+        if node.lineno in self.processed_lines:
+            return  # Skip already processed lines
+        self.processed_lines.add(node.lineno)
+        kwargs = {arg.arg: arg.value for arg in node.keywords}
+
+        # Retrieve and evaluate values
+        tries = self._eval_value(kwargs.get("tries"), default=4)
```
ebenahar, line 36: how about saving the default values in a conf file?

scripts/python/checks/validate_retry_usage.py
```diff
@@ -0,0 +1,109 @@
...
+        if total_time > max_timeout:
+            self.errors.append(
+                (
+                    node.lineno,
+                    f"Total retry time ({total_time}s or {total_time // 3600}h {total_time % 3600 // 60}m) "
+                    f"exceeds max_timeout ({max_timeout}s or {max_timeout // 3600}h {max_timeout % 3600 // 60}m).",
+                )
+            )
+
+    def _calculate_total_time(self, tries, delay, backoff, max_delay):
```
ebenahar, line 65: I assume it won't be straight forward to have a single implementation for this and the one in [ocs_ci/utility/retry.py](https://github.com/red-hat-storage/ocs-ci/pull/11224/files#diff-8853bdf565bbbd80748a64aeff9cc34510945363a6c88e4cbc89b81bd1ae6b8f), right?

ocs_ci/helpers/dr_helpers.py
```diff
@@ -1216,7 +1216,7 @@ def get_fence_state(drcluster_name, switch_ctx=None):
     return state
 
 
-@retry(UnexpectedBehaviour, tries=40, delay=5, backoff=5)
+@retry(UnexpectedBehaviour, tries=25, delay=5, backoff=5)
```
DanielOsypenko, line 1219: We had cases when backoff multiplier makes execution never end.
Discussion in gmail Subject "repeating" from @petr-balogh from Fri, Oct 11, 2024

Backoff multiplier increases delays between retries exponentially

```
delays = [initial_delay * (backoff_multiplier ** (i - 1)) for i in range(1, num_retries + 1)]
 ```
in case of `tries=25, delay=5, backoff=5` it will be

1st retry: 5 × (5^0) = 5 seconds
2nd retry: 5 × (5^1) = 25 seconds
3rd retry: 5 × (5^2) = 125 seconds
4th retry: 5 × (5^3) = 625 seconds
...
24 retry: 5 × (5^23) = 59604644775390625 seconds
25 retry: 5 × (5^24) = 298023223876953125 seconds

We may want to remove backoffs larger than 2 in most cases, because they are misused like here in ocs_ci/helpers/dr_helpers.py
and `ocs_ci/ocs/ui/acm_ui.py`:
```
 @retry(TimeoutException, tries=10, delay=10, backoff=5)
    def navigate_applications_page(self):
```

Also we may want to consider a default `backoff=1` ?

File: ocs_ci/helpers/dr_helpers.py [hunk 5]
```diff
@@ -1288,7 +1288,7 @@ def restore_backup():
     config.switch_ctx(restore_index)
 
 
-@retry(UnexpectedBehaviour, tries=40, delay=5, backoff=5)
+@retry(UnexpectedBehaviour, tries=25, delay=5, backoff=5)
```
DanielOsypenko, line 1291: same as above, lets reduce backoff to 1 in this case? because 25 tries with delay 5 and even with backoff = 2 it will last 2+ years

ocs_ci/helpers/dr_helpers.py
```diff
@@ -1309,7 +1309,7 @@ def verify_restore_is_completed():
     config.switch_ctx(restore_index)
 
 
-@retry(UnexpectedBehaviour, tries=60, delay=5, backoff=2)
+@retry(UnexpectedBehaviour, tries=25, delay=5, backoff=2)
```
DanielOsypenko, line 1312: please correct backoff or tries to make it last shorter time

ocs_ci/helpers/dr_helpers.py
```diff
@@ -1336,7 +1336,7 @@ def verify_drpolicy_cli(switch_ctx=None):
         )
 
 
-@retry(UnexpectedBehaviour, tries=40, delay=5, backoff=5)
+@retry(UnexpectedBehaviour, tries=25, delay=5, backoff=5)
```
DanielOsypenko, line 1339: please correct backoff or tries to make it last shorter time

ocs_ci/utility/retry.py
```diff
@@ -5,20 +5,59 @@
...
+            )
+        current_delay = min(current_delay * backoff, max_delay)
+
+
 def retry(
-    exception_to_check, tries=4, delay=3, backoff=2, text_in_exception=None, func=None
+    exception_to_check,
+    tries=4,
+    delay=3,
+    backoff=2,
```
DanielOsypenko, line 40: we may want to reduce default backoff 2 to 1 if tries  in case if retries are more than 12.

ocs_ci/helpers/dr_helpers.py
petr-balogh: @DanielOsypenko it won't, as I limited maximum delay time to be 10 minutes , so once it should go over 10 minutes it's limited to max 10 minutes.  Based on pre-commit hook check , this will new values will be under 4 hours max.

ocs_ci/utility/retry.py
clacroix12: That makes sense to me, I agree with that approach.

scripts/python/checks/validate_retry_usage.py
petr-balogh: This is simple script located in this path and completely independent on the framework. If it will require imports from framework it will require to install whole virtual env of ocs-ci which is not worth for this purpose and I would rather keep it separate here.

ocs_ci/utility/retry.py
petr-balogh: as I mentioned in other comment, we are limiting max delay to be 600 seconds, so we will never wait for more than 10 minutes in between attempts so we will not reach mentioned `3 hours and 24 minutes`

ocs_ci/helpers/dr_helpers.py
petr-balogh: I don't want to change much and come up with new calculations out of nowhere when I don't know the usage of those functions. As per my other comments , this is the change which will limit us to max 4 hours of retry and max 10 minutes delay between attempts. So, the plan will be probably in next round try to adjust values, run pre commit script - identify again problematic re-tries with even shorter time than 4 hours max time - and as part of some initiative let squads to fix their test cases - and not me fixing those as I don't know which test need what number of re-tries.

DanielOsypenko (dismissed):

ocs_ci/utility/retry.py
```diff
@@ -5,20 +5,59 @@
...
+        tries: Number of times to try (not retry) before giving up.
+        delay: Initial delay between retries in seconds.
+        backoff: Backoff multiplier e.g., value of 2 will double the delay each retry.
+        text_in_exception: Retry only when text_in_exception is in the text of exception.
+        func: Function for garbage collector.
+        max_timeout: Maximum total time for retries in seconds (default: 4 hours).
+        max_delay: Maximum delay between retries in seconds (default: 600 seconds or 10 minutes).
     """
+    # Validate parameters before proceeding
+    validate_retry_params(tries, delay, backoff, max_timeout, max_delay)
```
clacroix12, line 60: By calling this here does that mean any implementation where the retry configured for more than the maximum will result in an ValueError being raised?

Is the intention that we are raising those errors, or that we are limiting the delay/max time of the decorator in cases where its configured to exceed it?

petr-balogh: > By calling this here does that mean any implementation where the retry configured for more than the maximum will result in an ValueError being raised?
> 
Yes, we basically propagate all arguments to this validation function and checking how long it will take with calculating delays and re-tries (backoff ...) and if it will exceed with given configuration the limits set in max_timeout then we will throw an exception.

I thought that in case it will be called in some testing phase with the change - throwing reasonable exception from this check directly to the user will inform the user about he/she doing something wrong will make the problem immediately observable.

> Is the intention that we are raising those errors, or that we are limiting the delay/max time of the decorator in cases where its configured to exceed it?

We arre intentionally throwing exception if user set's the combination of args of retry decorator will take longer than provided max_timeout (default 4 hours). If user will set something longer, e.g. 5 hours if needed and it will be under that limit of custom set max_timeout 5 hours, it should not throw an exception here.


Hope I answered all you've asked :)

clacroix12: Got it, yeah that answers my question! I mostly wanted to avoid the situation where this is merged and existing tests are raiding errors due to this but I assumed that it would be covered by your updates to certain test cases. So at least at the time of merging this, any usage of the retry decorator should be passing validation and only future misconfiguration should be raising this ValueError. If that's the case then this looks great to me and gives us a way to enforce stricter limits in the future.

clacroix12 (approved):

vavuthu (approved):

shylesh (approved):
lgtm

PR #11239, opened 2025-01-28
Fix creating the RBD namespace command with external mode

yitzhak12:


yitzhak12:
I ran a test in my local, and it passed using the PR.

petr-balogh (approved):

PR #11273, opened 2025-01-30
Automatically resolve ceph daemon crash

fbalak:


ocs_ci/utility/utils.py
```diff
@@ -2489,13 +2489,15 @@ def check_ceph_health_not_ok():
     sampler.wait_for_func_status(True)
 
 
-def ceph_health_check(namespace=None, tries=20, delay=30):
+def ceph_health_check(namespace=None, tries=20, delay=30, resolve_daemon_crash=True):
```
prsurve, line 2492: I feel that the value resolve_daemon_crash should not be set as TRUE by default in some cases, we can actually miss a bug just by archiving it. Value should be set true for specific tests only and not default

ebenahar (approved):

ocs_ci/utility/utils.py
```diff
@@ -2514,16 +2516,18 @@ def ceph_health_check(namespace=None, tries=20, delay=30):
         tries=tries,
         delay=delay,
         backoff=1,
-    )(ceph_health_check_base)(namespace)
+    )(ceph_health_check_base)(namespace, resolve_daemon_crash)
```
petr-balogh, line 2519: I would go more generic like and name it like `fix_ceph_health` - I guess we might have multiple cases like this maybe in future we might have some fix for them.

ocs_ci/utility/utils.py
```diff
@@ -2536,6 +2540,15 @@ def ceph_health_check_base(namespace=None):
     namespace = namespace or config.ENV_DATA["cluster_namespace"]
     health = run_ceph_health_cmd(namespace)
 
+    if "daemons have recently crashed" in health and resolve_daemon_crash:
```
petr-balogh, line 2543: I would move this block under to separate function like ceph_health_resolve_daemon_crash. Than have function called `fix_ceph_health(health_status)` which will take the health as param and then it will try to fix that, so we can be adding multiple if and elif blocks under it. Which will decide which function to call to fix our issue.

ocs_ci/utility/utils.py
```diff
@@ -2536,6 +2540,15 @@ def ceph_health_check_base(namespace=None):
...
+        # importing here to avoid circular imports
+        from ocs_ci.ocs.resources.pod import get_ceph_tools_pod
+
+        log.error(health)
+        ct_pod = get_ceph_tools_pod()
+        ceph_crash_info_display(ct_pod)
```
petr-balogh, line 2549: in fucntion of fix_ceph_health I would throw new type of exception like:
CephHealthForciblyRecoveredException

And in ceph_health_check this exception will not be handled by re-try as we need to raise this exception to actually fail at least one test case for it to notice the issue.

Also I think that we should at the end of this function check ceph health status with: fix_ceph_health=False

We can reduce - re-tries if needed for this call - just to give a chance that next test case will be fine to run with Ceph Health OK

PR #11300, opened 2025-02-04
Delete leftover dns records when destroying ibmcloud clusters

clacroix12:
Issue: https://url.corp.redhat.com/db47959

### DIFF
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -2696,6 +2696,8 @@ SHA_SEPARATOR = "@sha256:"
 
 # ibmcloud related constants
 IBMCLOUD_VOLUME_NAME = "ibmvolume"
+IBMCLOUD_BASE_DOMAIN_ID = "a180c88cc263d58808a739e1fdfc268a"
+IBMCLOUD_CIS_INSTANCE_NAME = "odf-qe-cis"
 
 # manifest.json and background.js files used for Chrome extention configuring
 # authenticated proxy, see also:
--- a/ocs_ci/ocs/constants.py
+++ b/ocs_ci/ocs/constants.py
@@ -2696,8 +2696,6 @@ SHA_SEPARATOR = "@sha256:"
 
 # ibmcloud related constants
 IBMCLOUD_VOLUME_NAME = "ibmvolume"
-IBMCLOUD_BASE_DOMAIN_ID = "a180c88cc263d58808a739e1fdfc268a"
-IBMCLOUD_CIS_INSTANCE_NAME = "odf-qe-cis"
 
 # manifest.json and background.js files used for Chrome extention configuring
 # authenticated proxy, see also:
--- a/ocs_ci/utility/ibmcloud.py
+++ b/ocs_ci/utility/ibmcloud.py
@@ -8,6 +8,7 @@ import json
 import logging
 import os
 import re
+import traceback
 import requests
 import time
 import ibm_boto3
@@ -1250,10 +1251,14 @@ def delete_dns_records(cluster_name):
         cluster_name (str): Name of the cluster, used to filter DNS records
 
     """
-    dns_domain_id = constants.IBMCLOUD_BASE_DOMAIN_ID
-    cis_instance_name = constants.IBMCLOUD_CIS_INSTANCE_NAME
+    dns_domain_id = config.ENV_DATA["base_domain_id"]
+    cis_instance_name = config.ENV_DATA["cis_instance_name"]
     ids_to_delete = []
     page = 1
+
+    logger.info(f"Setting cis instance to {cis_instance_name}")
+    run_ibmcloud_cmd(f"ibmcloud cis instance-set {cis_instance_name}")
+
     while True:
         out = run_ibmcloud_cmd(
             f"ibmcloud cis dns-records {dns_domain_id} --per-page 1000 --page {page} --output json"
@@ -1274,9 +1279,12 @@ def delete_dns_records(cluster_name):
     logger.info(f"Records to delete: {records}")
     for record_id in ids_to_delete:
         logger.info(f"Deleting DNS record: {record_id}")
-        run_ibmcloud_cmd(
-            f"ibmcloud cis dns-record-delete {dns_domain_id} {record_id} -i {cis_instance_name}"
-        )
+        try:
+            run_ibmcloud_cmd(
+                f"ibmcloud cis dns-record-delete {dns_domain_id} {record_id}"
+            )
+        except CommandFailed:
+            logger.error(traceback.format_exc())
 
 
 class IBMCloudObjectStorage:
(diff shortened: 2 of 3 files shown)


vavuthu (approved):

ocs_ci/ocs/constants.py
```diff
@@ -2696,6 +2696,8 @@
 
 # ibmcloud related constants
 IBMCLOUD_VOLUME_NAME = "ibmvolume"
+IBMCLOUD_BASE_DOMAIN_ID = "a180c88cc263d58808a739e1fdfc268a"
+IBMCLOUD_CIS_INSTANCE_NAME = "odf-qe-cis"
```
dahorak, line 2700: Wouldn't be better to provide those parameters via configuration, so it will not be hardcoded just for our account? (Similarly as we have `base_domain` configuration for Azure in `azure_qe_rh_ocs_com` credentials conf.)

dahorak (commented):
Just one small suggestion, otherwise LGTM.

ocs_ci/utility/ibmcloud.py
```diff
@@ -1242,6 +1242,43 @@ def find_free_network_subnets(subnet_cidr, network_prefix=27):
...
+        cluster_name (str): Name of the cluster, used to filter DNS records
+
+    """
+    dns_domain_id = constants.IBMCLOUD_BASE_DOMAIN_ID
+    cis_instance_name = constants.IBMCLOUD_CIS_INSTANCE_NAME
+    ids_to_delete = []
+    page = 1
+    while True:
+        out = run_ibmcloud_cmd(
+            f"ibmcloud cis dns-records {dns_domain_id} --per-page 1000 --page {page} --output json"
```
dahorak, line 1259: Btw. shouldn't be the `-i {cis_instance_name}` argument used also in this command?

Because when I tried to run this command locally, it failed with following error:
```
$ ibmcloud cis dns-records a180c88cc263d58808a739e1fdfc268a --per-page 100 --page 1
FAILED
Context instance is not set yet. Use 'ibmcloud cis instance-set INSTANCE' to set the context instance before running commands.
```

ocs_ci/ocs/constants.py
petr-balogh: I prefer to have those set as part of credential conf in ocs4-jenkins repository for our IBM Cloud account and we can then read it from config.

ocs_ci/utility/ibmcloud.py
petr-balogh: We need to first run this command:
```
ibmcloud cis instance-set odf-qe-cis
```

And once this is set it will allow us to run CIS related operations - this has to be triggered only once for the session, so we should run this outside of the while loop - somewhere before while loop perheps.

I think I shared this command with you yesterday in DM Coady.

ocs_ci/utility/ibmcloud.py
```diff
@@ -1242,6 +1242,43 @@ def find_free_network_subnets(subnet_cidr, network_prefix=27):
...
+        )
+        records = json.loads(out)
+        if not records:
+            logger.info("Reached end of pagination")
+            break
+
+        filter_string = f".{cluster_name}."
```
petr-balogh, line 1266: Cool, I checked we usually have 3 entries:
```
api.rperiyas-scltest.ibmcloud2.qe.rh-ocs.com
api-int.rperiyas-scltest.ibmcloud2.qe.rh-ocs.com
*.apps.rperiyas-scltest.ibmcloud2.qe.rh-ocs.com
```

So as you are using `.rperiyas-scltest.` it should not by mistake delete other cluster which will have subset of string if person just add just extra number at the end of previous name.

👍

ocs_ci/utility/ibmcloud.py
```diff
@@ -1242,6 +1242,43 @@ def find_free_network_subnets(subnet_cidr, network_prefix=27):
...
+        for record in records:
+            if filter_string in record["name"]:
+                logger.info(f"Found {record['name']}, marking for deletion")
+                ids_to_delete.append(record["id"])
+        page += 1
+
+    logger.info(f"Records to delete: {records}")
+    for record_id in ids_to_delete:
+        logger.info(f"Deleting DNS record: {record_id}")
+        run_ibmcloud_cmd(
```
petr-balogh, line 1277: I would cover this to try except block and do print error message with traceback only in case this fail from some reason.  In case one entry will fail from some reason to be deleted - the next one iteration will try to delete the other at least.   In case of failure we can store exception in some list and then re-raise this exception - or we can just have this logged as error - as this doesn't look like critical leftover as we as of now had a lot of them.

ocs_ci/utility/ibmcloud.py
clacroix12: Good catch, this is being explicitly set now.

ocs_ci/utility/ibmcloud.py
clacroix12: That's the exact scenario I wanted to avoid here, since I do that a lot :)

ocs_ci/ocs/constants.py
clacroix12: This is being read in from the configuration. I've also opened an MR to add this data to the credentials

clacroix12:
Verified and confirmed the leftover record was deleted here: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-destroy-ocs-cluster/52685/consoleFull

The only issue was incorrectly referencing the wrong list in the log statement showing which IDs were found for deletion. I've pushed a fix addressing that.

clacroix12:
I should also mention that this shouldn't be merged prior to the MR adding the config details to our credentials.

ocs_ci/utility/ibmcloud.py
```diff
@@ -1242,6 +1243,50 @@ def find_free_network_subnets(subnet_cidr, network_prefix=27):
...
+
+    logger.info(f"Records to delete: {ids_to_delete}")
+    for record_id in ids_to_delete:
+        logger.info(f"Deleting DNS record: {record_id}")
+        try:
+            run_ibmcloud_cmd(
+                f"ibmcloud cis dns-record-delete {dns_domain_id} {record_id}"
+            )
+        except CommandFailed:
+            logger.error(traceback.format_exc())
```
petr-balogh, line 1287:
```suggestion
            logger.exception("Failed to delete CIS leftovers")
```

I think that logger.exception give us also details about traceback without need of import and use traceback:

```
In [7]: try:
   ...:     1/0
   ...: except Exception:
   ...:     logger.exception("This is the zero division error")
   ...:
This is the zero division error
  File "<ipython-input-7-8c23c22edfe2>", line 2, in <module>
    1/0
    ~^~
ZeroDivisionError: division by zero
```

clacroix12: Removed the import and went with your suggestion

petr-balogh (approved):

vavuthu (approved):

dahorak (approved):

clacroix12:
Manually cherry-picked for other release branches here:

#11314 
#11315 
#11316

PR #11303, opened 2025-02-05
Stabilize and adjust `test_bucket_notifications`

sagihirshfeld:
**Changes made:**

1. Setup and teardown of the AMQ cluster that is used to setup Kafka takes a long time (~5 mins). This hinders the debugging and development process. 

To fix this issue I've enabled local re-running of the test without destroying and re-creating the AMQ cluster - just use `run-ci` with the --`dev-mode` option. 

Note that this will affect PR verifications on an existing cluster the AMQ cluster is already set up, so to bypass this one needs to destroy the cluster manually:
```
$ oc delete namespace myproject
$ oc delete all -l app=strimzi
```

2. Added  a retry to the fetching of the PVC in `enable_bucket_notifs_on_cr` - the test would often fail on new clusters because MCG was still creating the PVC when we were trying to read the resource.

3. Adjust `put-bucket-notification` to use `<secret-name>/<config_json_file_name>` under `Topic` instead of just `<config_json_file_name>`. See https://github.com/noobaa/noobaa-operator/pull/1515/files for details.

4. Replace the `CephCluster().wait_for_noobaa_health_ok()` costly health check with the tighter `MCG.wait_for_ready_status()`. Note that this required a workaround to use the existing `MCG::status` method in a static manner.

5. Make the wait optional at `enable_bucket_notifs_on_cr` and `put_bucket_notification` (this is particularly useful in the more advanced tests that are still in draft).

6. Adjust `get_events` - MCG now sends one unstructured notification to the topic to test it (`test notification`), so this adjustments makes sure we're only reading the JSON lines.

7. Skip the test on environments where network access is limited since the setup requires cloning an external repo.

8. Improve the error logging in `test_bucket_notifications`

### DIFF
--- a/ocs_ci/ocs/amq.py
+++ b/ocs_ci/ocs/amq.py
@@ -1035,3 +1035,19 @@ class AMQ(object):
         switch_to_default_rook_cluster_project()
         run_cmd(f"oc delete project {kafka_namespace}")
         self.ns_obj.wait_for_delete(resource_name=kafka_namespace, timeout=90)
+
+    def check_amq_cluster_exists(self):
+        """
+        Check if amq cluster exists
+
+        Returns:
+            bool: True if amq cluster exists
+
+        """
+        try:
+            self.ns_obj.get(resource_name=constants.AMQ_NAMESPACE)
+            return True
+        except CommandFailed as cf:
+            if "NotFound" in str(cf):
+                return False
+            raise cf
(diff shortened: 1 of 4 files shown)


mashetty330 (approved):

ypersky1980 (approved):
LGTM

ebenahar (approved):

Issue #11311, opened 2025-02-07
Ceph health not OK aftertest_consumption_trend_after_osd_resize


ebondare:
Ceph health could not recover after running tests/functional/ui/test_odf_storage_consumption_trend.py::TestConsumptionTrendUI::test_consumption_trend_after_osd_resize

PR #11312, opened 2025-02-07
Skip resize test until bug is fixed

ebondare:
DFBUGS-1467 results in ceph health not being able to recover, so we need to skip resize test until it's fixed

Fixes https://github.com/red-hat-storage/ocs-ci/issues/11311


parth-gr:
@ebondare the fix is merge from the ceph side, can we check on how the build looking?

ebondare:
> @ebondare the fix is merge from the ceph side, can we check on how the build looking?
The test still results in unhealthy state of ceph. https://issues.redhat.com/browse/DFBUGS-1467 is in FailedQA

DanielOsypenko (approved):

parth-gr:
@ebondare the ceph version used was not correct, I still say to wait till another round of testing

suchita-g (approved):

PR #11329, opened 2025-02-09
`BucketLoggingManager::wait_for_logs_pvc_mount_status` - Increase the timeout from 2 minutes to 5

sagihirshfeld:
As seen [here](https://url.corp.redhat.com/2eeffd1), bucket logs tests often fail because the pods fail to mount the logging PVC in time.

There might be a more nuanced race-condition that causes this, but I wasn't able to reproduce it locally or via Jenkins. If increasing the timeout won't fix the issue, then that would at least be a hint we can further use to find the root cause.

### DIFF
--- a/ocs_ci/ocs/resources/bucket_logging_manager.py
+++ b/ocs_ci/ocs/resources/bucket_logging_manager.py
@@ -300,7 +300,7 @@ class BucketLoggingManager:
 
         return answers_dict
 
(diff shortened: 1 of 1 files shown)


sagihirshfeld:
Adding the Verified label due to the nature of the fix. The test still passes even with the lesser timeout: https://url.corp.redhat.com/f8b269a

udaysk23 (approved):

keesturam (approved):

Issue #10982, opened 2024-12-04
test_object_actions sometimes fails with AccessDenied on the put-object op


sagihirshfeld:
In this test we're setting a bucket policy that is equivalent to the following:
```
$ my_s3api_alias put-bucket-policy --bucket "$BUCKET" --policy '{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Deny",
            "Action": [
                "s3:GetObject",
                "s3:DeleteObject"
            ],
            "Principal": {
                "AWS": "obc-account.my-obc-2.675049c7@noobaa.io"
            },
            "Resource": "arn:aws:s3:::'"$BUCKET"'/*"
        }
    ]
}'
```
Although this operation only restricts the OBC's user from the GetObject and Delete Object ops, its PutObject operation fails about 60 seconds after applying it even though it should have that access by default. 

The hit ratio on this is pretty rare (about 5%), so it might be a timing issue.

RP links:
- https://url.corp.redhat.com/d4337a5
- https://url.corp.redhat.com/f816f84

PR #11340, opened 2025-02-10
`test_object_actions` - double the sleep from 60 seconds to 120

sagihirshfeld:
Fixes: https://github.com/red-hat-storage/ocs-ci/issues/10982

Since adding a 60 seconds sleep in https://github.com/red-hat-storage/ocs-ci/pull/9447 reduced the hit ratio from 20% (according to https://github.com/red-hat-storage/ocs-ci/issues/9446) to ~5%, then there's room to believe this is still a race condition issue that more timeout will solve.


sagihirshfeld:
Setting the Verified due to the nature of the fix - a single successful execution of the test won't tell us much in this case.

keesturam (approved):

udaysk23 (approved):

ypersky1980 (approved):
LGTM

PR #11343, opened 2025-02-10
Added ttl value to the vault token generation command

paraggit:
Added ttl value to the vault token generation command

### DIFF
--- a/ocs_ci/utility/kms.py
+++ b/ocs_ci/utility/kms.py
@@ -599,7 +599,7 @@ class Vault(KMS):
         # Printing out here the Vault Path Token
         logger.info(f"Vault Path Token is: ${self.vault_path_token}")
 
-    def generate_vault_token(self):
+    def generate_vault_token(self, ttl="768h"):
         """
         Generate a token for self.vault_policy_name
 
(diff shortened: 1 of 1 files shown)


petr-balogh (approved):

petr-balogh:
Verification job:
 https://url.corp.redhat.com/58955ad

ocs_ci/utility/kms.py
```diff
@@ -599,15 +599,19 @@ def vault_create_policy(self, policy_name=None):
...
         Generate a token for self.vault_policy_name
 
         Returns:
             str: vault token
 
         """
-        cmd = f"vault token create -policy={self.vault_policy_name} " f"--format=json"
+        cmd = (
+            f"vault token create -policy={self.vault_policy_name} "
+            f"--format=json "
```
dahorak, line 612: There is a f-string without any variable, which I think I think this should be catch by pre-commit hook.

dahorak (approved):
Not a big deal, but I think pre-commit hook is complaining about f-strings without any variables, so it might be worth to fix it now.

petr-balogh (approved):

dahorak (approved):

Issue #11344, opened 2025-02-10
Creating buckets via S3 sometimes results in NoSuchBucket errors, and then fails to get deleted on teardown


sagihirshfeld:
```
E           botocore.errorfactory.NoSuchBucket: An error occurred (NoSuchBucket) when calling the PutObject operation: The specified bucket does not exist.
```
```
E           AssertionError: s3-bucket-4ccb8adac3d84b328b2872db4c8908 was not deleted within 60 seconds.
```

- I couldn't reproduce this manually, but we're seeing this issue rather frequently in regression analysis. 
- This only happens with buckets that were created via S3. Look at the reason that tests fail in `test_bucket_policy.py` - only `test_anonymous_read_only` and `test_public_website` are showing these errors, and these are the only tests that are creating the bucket via S3 and not OC.

Examples:
- https://url.corp.redhat.com/f4044d6
- https://url.corp.redhat.com/a5bdb6a

sagihirshfeld:
Further investigation is required on whether this is due to an automation issue or a product bug.

abdulkandathil:
Observing this behavior on IBM Z as well.

PR #11345, opened 2025-02-10
`test_bucket_policy.py` - create all the buckets via OC instead of S3

sagihirshfeld:
Looking at many regression runs - tests in `test_bucket_policy.py` that create the bucket via OC instead of S3 seem to never face instability around the creation/deletion of the bucket.

This is simply a workaround for https://github.com/red-hat-storage/ocs-ci/issues/11344 in these specific tests that are not focused around bucket creation.


udaysk23 (approved):

keesturam (approved):
Good observation @sagihirshfeld :)

PR #11431, opened 2025-02-18
VolumeSynchronizationDelay alert on the DR dashboard due to sync delays

am-agrawa:
Test to verify that "VolumeSynchronizationDelay" warning level alert is fired on the DR dashboard when the lastGroupSyncTime is lagging behind 2x or critical level alert when this value is behind 3x the sync interval defined in DRPolicy applied to a particular workload.
      
No DR action is performed in this test case. We scale down rbd-mirror daemon deployment on the secondary cluster and mds daemons on the primary cluster and scale them up back to their original count.

Also added and updated the name to marker to `@rdr_ui` as per my discussion with @sidhant-agrawal

### DIFF
--- a/tests/functional/disaster-recovery/regional-dr/test_warning_and_alerting.py
+++ b/tests/functional/disaster-recovery/regional-dr/test_warning_and_alerting.py
@@ -18,6 +18,7 @@ from ocs_ci.ocs import constants
 from ocs_ci.ocs.acm.acm import AcmAddClusters
 from ocs_ci.helpers.dr_helpers_ui import (
     failover_relocate_ui,
+    verify_drpolicy_ui,
 )
 from ocs_ci.ocs.exceptions import UnexpectedBehaviour
 from ocs_ci.ocs.ui.validation_ui import ValidationUI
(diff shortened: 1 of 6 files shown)


tests/functional/disaster-recovery/regional-dr/test_warning_and_alerting.py
```diff
@@ -348,3 +349,202 @@ def test_rdr_inconsistent_data_warning_alert(
...
+        config.switch_to_cluster_by_name(primary_cluster_name)
+        helpers.modify_deployment_replica_count(
+            deployment_name=constants.MDS_DAEMON_DEPLOYMENT_ONE, replica_count=0
+        )
+        helpers.modify_deployment_replica_count(
+            deployment_name=constants.MDS_DAEMON_DEPLOYMENT_TWO, replica_count=0
+        )
+        config.switch_acm_ctx()
+        logger.info(f"Waiting for {wait_time * 60} seconds to allow alert to appear")
+        sleep(wait_time * 60)
```
keemano, line 440: Here, we are waiting for 2x of the scheduled interval but in the description it has been mentioned 3x.

keemano, line 440: is it enough to wait 2x of scheduled interval for the alerts to be triggered? In the description it has been mentioned as 3x of the scheduled interval.

am-agrawa: @keemano Updated multiple logics in the test, please re-check

tests/functional/disaster-recovery/regional-dr/test_warning_and_alerting.py
```diff
@@ -33,41 +36,59 @@
...
+        modify_deployment_count(status="up")
 
     request.addfinalizer(teardown)
 
 
 @rdr
 @tier1
 @turquoise_squad
-@skipif_ocs_version("<4.18")
 class TestRDRWarningAndAlerting:
```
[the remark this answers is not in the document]
am-agrawa: @sidhant-agrawal @sidhant-agrawal Added the marker and updated it's name to `@rdr_ui` as per our discussion

prsurve (dismissed):

prsurve (dismissed):

keesturam (dismissed):

tests/functional/disaster-recovery/regional-dr/test_warning_and_alerting.py
```diff
@@ -348,3 +346,227 @@ def test_rdr_inconsistent_data_warning_alert(
                         avoid_stale=True,
                     )
                     logger.info("Action modal closed successfully")
+
+    @polarion_id("OCS-5304")
```
DanielOsypenko, line 350: hi @dahorak , @petr-balogh , curious if it is ok to use multiple poparion-id's on one test. It can be handy, although, will reporting, ci and scripts in spreadsheets work for every item?

keemano (approved):

tests/functional/disaster-recovery/regional-dr/test_warning_and_alerting.py
DanielOsypenko: discussed here: https://url.corp.redhat.com/ef24633 
looks like having multiple polarion id's is not the solution that we may use in automation

dahorak: Technically this doesn't work, because it saves only the last polarion ID in the XML junit file.
I tried it with following dummy test case:
```
@polarion_id("AAA")
@polarion_id("BBB")
@polarion_id("CCC")
@polarion_id("DDD")
def test_dahorak():
    ...
```
And the junit xml output file looks like this:
```
    <testcase classname="tests.test_dahorak_dummy" name="test_dahorak" time="1.256">
      <properties>
        <property name="polarion-testcase-id" value="DDD"/>
      </properties>
    </testcase>
```

am-agrawa: Will handle as part of fixing https://github.com/red-hat-storage/ocs-ci/issues/11728

DanielOsypenko (approved):

prsurve (approved):

dahorak (approved):

PR #11480, opened 2025-02-24
Verify IPv6 Connectivity Between Rook-Ceph-Exporter and OCP Prometheus

OdedViner:
### **Description:**  
This PR introduces a test to verify the connectivity of `rook-ceph-exporter` pods to OpenShift Prometheus over **IPv6**.  

#### **Test Procedure:**  
1. Retrieve the **IPv6 addresses** of `rook-ceph-exporter` pods.  
2. Identify the `prometheus-k8s` pod in the **openshift-monitoring** namespace.  
3. Use `curl` from within the `prometheus-k8s` pod to access the `/metrics` endpoint of each exporter via IPv6.  
4. Verify that the response contains the expected Ceph metric: **`ceph_AsyncMessenger_Worker_msgr_connection`**.

### DIFF
--- a/tests/functional/monitoring/prometheus/metrics/test_monitoring_defaults.py
+++ b/tests/functional/monitoring/prometheus/metrics/test_monitoring_defaults.py
@@ -289,7 +289,6 @@ def test_provider_metrics_available(threading_lock):
 
 @blue_squad
 @tier1
-@bugzilla("2297285")
 @skipif_external_mode
 @pytest.mark.polarion_id("OCS-6796")
 def test_monitoring_ip_connectivity(threading_lock):
--- a/tests/functional/monitoring/prometheus/metrics/test_monitoring_defaults.py
+++ b/tests/functional/monitoring/prometheus/metrics/test_monitoring_defaults.py
@@ -306,9 +306,12 @@ def test_monitoring_ipv6(threading_lock):
     )
     prometheus_pod_obj = None
     for pod_obj in pod_obj_list:
-        if "prometheus-k8s" in pod_obj.name:
-            prometheus_pod_obj = pod_obj
-            break
+        try:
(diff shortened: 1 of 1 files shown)


tests/functional/monitoring/prometheus/metrics/test_monitoring_defaults.py
```diff
@@ -282,3 +283,43 @@ def test_provider_metrics_available(threading_lock):
         "so that the list of metrics without results is empty."
     )
     assert list_of_metrics_without_results == [], msg
+
+
+@blue_squad
+@tier1
+@bugzilla("2297285")
+@pytest.mark.polarion_id("OCS-XXX")
+def test_monitoring_ipv6(threading_lock):
```
DanielOsypenko, line 292: no `rook-ceph-exporter` on External mode clusters because rook does not manage ceph in this case. 
I think this test is irrelevant to External mode and we may add "skip_if" decorator

DanielOsypenko, line 292: should we check if ipv6 feature enabled on deployment before running this test? if not we may want to skip.
we'll probably need to add marker for `ipv6_enabled`

OdedViner: done

DanielOsypenko (dismissed):

tests/functional/monitoring/prometheus/metrics/test_monitoring_defaults.py
OdedViner: done `test_monitoring_ip_connectivity`

AviadP (approved):

DanielOsypenko (approved):

PrasadDesala (approved):

PR #11493, opened 2025-02-25
Skipping UI test cases which uses kubeadmin pw

vasudha786:
Skipping UI test cases which uses kubeadmin pw

### DIFF
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -39,6 +39,7 @@ from ocs_ci.ocs.constants import (
     ROSA_HCP_PLATFORM,
     VAULT_KMS_PROVIDER,
 )
+from ocs_ci.framework.pytest_customization.marks import skipif_ibm_cloud_managed
 from ocs_ci.utility import version
 from ocs_ci.utility.aws import update_config_from_s3
 from ocs_ci.utility.utils import load_auth_config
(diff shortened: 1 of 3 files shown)


DanielOsypenko:
we may want to fix issue with ibm_cloud instead of skipping. 
It requires generating login:password pair through credentials generator such as `htpasswd`  
we did this for rosa and other deployments.

petr-balogh:
> htpasswd

Hi Daniel, I have tried this on ROKS cluster - but it doesn't work.  IBM Managed platform enforce only the login through IBM Cloud - doesn't allow anything else - and if you set up other method - it doesn't work.

For UI deployment I have some work done - but it's using my personal account at the moment and we are in discussion to get it working for our functional ID instead of my credentials - so once that's done - we will be able to enable UI tests probably with that approach - but we will still miss the kubeadmin or whatever user password.

I can provide you such cluster - so you might try to help, but I am afraid you will end up with the same.

petr-balogh (dismissed):

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -726,3 +726,5 @@ def get_current_test_marks():
     not in [VAULT_KMS_PROVIDER, HPCS_KMS_PROVIDER],
     reason="This test requires both Vault or HPCS KMS deployment to be enabled and a valid KMS provider.",
 )
+
+ui = compose(skipif_ibm_cloud_managed, pytest.mark.ui)
```
vavuthu, line 730: ui is too generic name here. when we see ui marker for test cases, it gives impression that test case is related to UI.

petr-balogh: @vavuthu we already have this UI marker here:
https://github.com/red-hat-storage/ocs-ci/blob/25d9e7549cb32d8700e8f33f7f12ab2d830d2452/ocs_ci/framework/pytest_customization/marks.py#L74

So I guess @vasudha786 forgot to delete the one from line 74.

On ROKS we do not support UI testing as of now - so I suggested Vasudha to combine it to skip basically all UI decorated tests on ROKS platform instead of separately decorate new and new tests.

petr-balogh (approved):

ebenahar (approved):

PR #11522, opened 2025-02-27
Fix ceph health

petr-balogh:


petr-balogh:
https://url.corp.redhat.com/acaa85a

ocs_ci/utility/utils.py
```diff
@@ -2514,13 +2516,80 @@ def check_ceph_health_not_ok():
...
+                    namespace,
+                    tries=fix_dict.get("ceph_health_tries", 5),
+                    delay=fix_dict.get("ceph_health_delay", 30),
+                )
+            except Exception as ex:
+                log.error("")
+                raise CephHealthExceptionNotRecovered(
+                    f"Attempt to try to recover the Ceph Health failed! Exception: {ex}"
+                )
+            raise CephHealthExceptionRecovered(
```
clacroix12, line 2576: As we discussed I think we should handle this Exception if caught during the setup so that the test case can still execute. We can still raise this during the teardown of the test case so the information isn't lost but it also won't block the test case.

ocs_ci/ocs/exceptions.py
```diff
@@ -14,6 +14,14 @@ class CephHealthException(Exception):
     pass
 
 
+class CephHealthExceptionRecovered(Exception):
+    pass
+
+
+class CephHealthExceptionNotRecovered(Exception):
+    pass
```
dahorak, line 22: We didn't have any hierarchy used in the exceptions yet, but wouldn't make sense to place those two new exceptions under some covering one? (Maybe`CephHealthException`, if it will not break anything elsewhere?)

dahorak (commented):
LGTM, just two small notes.

ocs_ci/ocs/exceptions.py
petr-balogh: I need to distinguish between those two specific cases and act differently than we experience CephHealthException, hence I cannot inherit from it and make some hierarchy

ocs_ci/utility/utils.py
petr-balogh: I have updated the logic to not fail the test case at the setup - but rather store the exception and raise it in the teardown - so we will still have coverage and the test should have error in the teardown so it should be visible in RP and analyzed by owner of the test.

petr-balogh:
New verification after changes:
https://url.corp.redhat.com/02e6a39

tests/conftest.py
```diff
@@ -1863,6 +1867,20 @@ def finalizer():
...
+            """
+            Handling a rare scenario where the teardown of the previous test completed successfully,
+            but the setup of the next test detects an unhealthy Ceph state:
+            - If a test case encounters a Ceph health issue during setup and it later recovers,
+              we still need to fail the test during teardown to make the issue noticeable.
+            - However, if Ceph health recovered before running the test, we proceed with execution
+              to avoid losing test coverage.
+            """
+            log.error(
+                "We are failing the test case becasue it was not healthy in the setup phase"
```
vavuthu, line 1880:
```suggestion
                "We are failing the test case because it was not healthy in the setup phase"
```

tests/conftest.py
```diff
@@ -1863,6 +1867,20 @@ def finalizer():
                 ):
                     ceph_health_check_multi_storagecluster_external()
                 raise
+        if ceph_health_recovered_exception:
```
vavuthu, line 1870: where will this value assigned to something, to enter to this if block

ocs_ci/utility/utils.py
```diff
@@ -2539,10 +2607,10 @@ def ceph_health_check(namespace=None, tries=20, delay=30):
         tries=tries,
         delay=delay,
         backoff=1,
-    )(ceph_health_check_base)(namespace)
+    )(ceph_health_check_base)(namespace, fix_ceph_health)
 
 
-def ceph_health_check_base(namespace=None):
+def ceph_health_check_base(namespace=None, fix_ceph_health=False):
```
vavuthu, line 2613: Add the parameter in the docstring

ocs_ci/utility/utils.py
```diff
@@ -2514,13 +2516,79 @@ def check_ceph_health_not_ok():
...
+            "ceph_health_tries": 5,
+            "ceph_health_delay": 30,
+        },
+        # TODO: Add more patterns and fix functions
+    ]
+    for fix_dict in ceph_health_fixes:
+        pattern = fix_dict["pattern"]
+        if re.search(pattern, health_status):
+            log.info(
+                "Trying to fix Ceph Health becasue we found in Health status the matching pattern"
```
vavuthu, line 2559:
```suggestion
                "Trying to fix Ceph Health because we found in Health status the matching pattern"
```

dahorak (dismissed):

ocs_ci/utility/utils.py
petr-balogh: done

ocs_ci/utility/utils.py

tests/conftest.py
petr-balogh: done

dahorak (approved):

vavuthu (approved):

PR #11523, opened 2025-02-27
add-hyperconverged-into-dependencies

DanielOsypenko:
In past we had a problem with deploying ODF on Hosted Client clusters that were deployed with latest unreleased CVN and latest unreleased ACM.
The nature of the problem was not determined, but installing upstream hyperconverged along with upstream mce proved to be sustainable. 
This PR is also giving us a backdoor option, when CNV ureleased version is still not available and upstream HCO is already been populated via unstable index-tag of quay.io/kubevirt/hyperconverged-cluster-index

Link to the issue: https://issues.redhat.com/browse/OCPBUGS-32196  
Link to verification run: https://privatebin.corp.redhat.com/?08e39cfadafad0c6#E5JtjD7BpRWhLi22Jobta6U3FX6hJ3RfgUJZcAvqjsUT

Added int parameter in configuration `ENV_DATA.clusters.<cluster_name>.storage_quota`. Default value - no storage_quota, meaning storage for this client is unlimited

### DIFF
--- a/ocs_ci/deployment/hosted_cluster.py
+++ b/ocs_ci/deployment/hosted_cluster.py
@@ -632,6 +632,11 @@ class HostedODF(HypershiftHostedOCP):
             constants.PROVIDER_MODE_STORAGE_CLIENT
         )
         self.storage_client_name = storage_client_data["metadata"]["name"]
+        self.storage_quota = (
+            config.ENV_DATA.get("clusters", {})
+            .get(self.name, {})
+            .get("storage_quota", None)
+        )
 
     @kubeconfig_exists_decorator
     def exec_oc_cmd(self, cmd, timeout=300, ignore_error=False, **kwargs):
@@ -976,7 +981,11 @@ class HostedODF(HypershiftHostedOCP):
         )
 
         try:
-            token = generate_onboarding_token(private_key=decoded_key)
+            token = generate_onboarding_token(
+                private_key=decoded_key,
+                use_ticketgen_with_quota=True,
+                storage_quota=self.storage_quota,
+            )
         except Exception as e:
             logger.error(f"Error during onboarding token generation: {e}")
             token = ""
--- a/ocs_ci/deployment/hosted_cluster.py
+++ b/ocs_ci/deployment/hosted_cluster.py
@@ -5,6 +5,7 @@ import tempfile
 import time
 from concurrent.futures import ThreadPoolExecutor
 from ocs_ci.deployment.cnv import CNVInstaller
+from ocs_ci.deployment.hyperconverged import HyperConverged
 from ocs_ci.deployment.mce import MCEInstaller
 from ocs_ci.deployment.deployment import Deployment
 from ocs_ci.deployment.helpers.hypershift_base import (
@@ -22,6 +23,7 @@ from ocs_ci.ocs.exceptions import (
     CommandFailed,
     TimeoutExpiredError,
     ResourceWrongStatusException,
+    UnexpectedDeploymentConfiguration,
 )
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs.resources.catalog_source import CatalogSource
@@ -36,8 +38,6 @@ from ocs_ci.utility.retry import retry
 from ocs_ci.utility.utils import (
     exec_cmd,
     TimeoutSampler,
-    get_ocp_version,
-    get_latest_release_version,
 )
 from ocs_ci.utility.version import get_semantic_version
 from ocs_ci.ocs.resources.storage_client import StorageClient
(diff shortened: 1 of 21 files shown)


ocs_ci/deployment/hosted_cluster.py
```diff
@@ -384,36 +428,16 @@ def __init__(self, name):
...
-        deploy_metallb=True,
-        download_hcp_binary=True,
-        deploy_mce=True,
-    ):
+    def deploy_ocp(self, **kwargs) -> str:
         """
         Deploy hosted OCP cluster on provisioned Provider platform
 
         Args:
-            deploy_cnv: (bool) Deploy CNV
```
fbalak, line 399: It would be good to keep some arguments documented and have **kwargs after they are listed.

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -493,31 +534,6 @@ def deploy_dependencies(
...
-            self.validate_mce_deployment()
-
-        provider_ocp_version = str(
-            get_semantic_version(get_ocp_version(), only_major_minor=True)
-        )
-        latest_released_ocp_version = str(
-            get_semantic_version(get_latest_release_version(), only_major_minor=True)
-        )
-
-        if provider_ocp_version > latest_released_ocp_version:
```
fbalak, line 505: This looks ok to me. If the workaround is needed for older version then we can use release-4.17 branch.

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -260,12 +260,50 @@ def deploy_hosted_ocp_clusters(self, cluster_names_list=None):
...
+                    raise UnexpectedDeploymentConfiguration(
+                        "Conflict: Both 'deploy_acm_hub_cluster' and 'mce_deployment' are enabled. Choose one."
+                    )
+                if deploy_cnv and deploy_hyperconverged:
+                    raise UnexpectedDeploymentConfiguration(
+                        "Conflict: Both 'cnv_deployment' and 'deploy_hyperconverged' are enabled. Choose one."
+                    )
+
+            else:
+                deploy_acm_hub = False
```
fbalak, line 286: Shouldn't the values be collected from config even after first deployment?

DanielOsypenko: this function is driven by `ocs_ci/deployment/deployment.py:deploy_cluster` which has its own `deploy_cnv` call.
We have it here in case we don't want to deploy Provider and use fixture, or try to deploy clients directly from other place.

So in this case it passes the calls where we do install cnv - mandatory, when config saying this and now only check that with first client cnv and rest dependencies are installed.
Hope that answers on q.

ocs_ci/deployment/hosted_cluster.py
DanielOsypenko: we don't use kwargs in any call, this was removed entirely, because it was bad design. Now before calling `deploy_ocp` we need to call `deploy_dependencies`, that reduces shared dependencies and makes code more flexible

DanielOsypenko: kwargs were used because deploy_ocp is overloaded method and pylint urges on missing arguments, and kinda for future use

ocs_ci/deployment/hosted_cluster.py
DanielOsypenko: we don't need this workaround anymore, because it was not stable at all.
Now we have option to deploy Hyperconverged from the start, in aregular way we do it will all other dependencies, instead of installing ACM -> turning off imbedded Hyperconverged -> installing Hyperconverged upstream.
So this code is redundant and not stable, hence removed

ocs_ci/deployment/hosted_cluster.py
fbalak: Ok, sounds reasonable.

fbalak (approved):

dahorak (commented):
Overall LGTM, just small question about one condition.

About the Provider - Client jenkins job, should I add following two parameters also to the job UI (similarly as we have `DEPLOY_ACM_HUB` and `DEPLOY_CNV`?
```
* `mce_deployment` - Deploy MCE or not (Default: false)
* `deploy_hyperconverged` - Deploy hyperconverged operator or not (Default: false).  Necessary for Converged clusters with hosted clients with unreleased OCP version
```


obrazek

dahorak (dismissed):

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -260,12 +260,50 @@ def deploy_hosted_ocp_clusters(self, cluster_names_list=None):
...
+                deploy_cnv = False
+                deploy_hyperconverged = False
+                deploy_mce = False
+
+            if not config.ENV_DATA["platform"].lower() in HCI_PROVIDER_CLIENT_PLATFORMS:
+                raise ProviderModeNotFoundException()
+
+            if not config.ENV_DATA.get("deploy_acm_hub_cluster", True):
+                deploy_acm_hub = False
```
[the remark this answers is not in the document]
DanielOsypenko: removed

DanielOsypenko:
verified with https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/view/all/job/qe-odf-provider-client-multicluster/139/

dahorak (approved):

fbalak (approved):

PR #11539, opened 2025-03-02
Adding a new test case for NSS creation with Multi bucketclass type via CLI

ypersky1980:
This PR implements a new test case for creation of bucket over Multi NSS and bucket class of Multi type via CLI .

### DIFF
--- a/ocs_ci/ocs/resources/mcg.py
+++ b/ocs_ci/ocs/resources/mcg.py
@@ -751,8 +751,8 @@ class MCG:
             name (str): The name to be given to the bucket class
             namespacestores (list): The namespaces stores to use as part of the policy
             namespace_policy (dict): The namespace policy to be used. The supported namespace policy types are
-                Single, Cache and Multiple. For Cache NSS, default noobaa backingstore is used as the cache.
-                In the case of Multi namespace policy type, first namespace store is used as the write resource.
+            Single, Cache and Multi. For Cache NSS, default backing store is used as the cache.
+            In the case of Multi namespace policy type, first namespace store is used as the write resource.
 
         Returns:
             OCS: The bucket class resource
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/resources/mcg.py
```diff
@@ -773,10 +773,14 @@ def cli_create_bucketclass_over_namespacestores(
...
             self.exec_mcg_cmd(cmd)
+        elif namespace_policy_type == constants.NAMESPACE_POLICY_TYPE_MULTI.lower():
+            cmd += f" --read-resources='{namestores_name_str}'"
+            cmd += f" --write-resource='{namespacestores[0].name}'"
+            self.exec_mcg_cmd(cmd)
         else:
             raise NotImplementedError(
                 f"Cli creating of bucketclass on namespacestore "
-                f"with policy {namespace_policy_type} is not implemented yet"
+                f"with policy {namespace_policy_type} is supported"
```
udaysk23, line 783: We should remove this exception since we implemented it

ocs_ci/ocs/resources/mcg.py
```diff
@@ -773,10 +773,14 @@ def cli_create_bucketclass_over_namespacestores(
             if "ttl" in namespace_policy:
                 cmd += f" --ttl=={namespace_policy['ttl']}"
             self.exec_mcg_cmd(cmd)
+        elif namespace_policy_type == constants.NAMESPACE_POLICY_TYPE_MULTI.lower():
+            cmd += f" --read-resources='{namestores_name_str}'"
+            cmd += f" --write-resource='{namespacestores[0].name}'"
```
sagihirshfeld, line 778: Using the first NSS as the write resource is fine, but please add a note at the method's docstring to communicate this.

While you're at it, please also document that we're using the default noobaa backingstore as the cache for cache NSS.

ocs_ci/ocs/resources/mcg.py
ypersky1980: I prefer to leave this exception in case unsupported policy type is passed but you are right, the wording of the exception is confusing and I change it to "not supported".

udaysk23 (approved):

keesturam (approved):

PR #11550, opened 2025-03-04
remove RHEL worker nodes deployment config files

dahorak:
- we are not supporting this for long time and the presence of those configs sometime confuses users and leads to failed deployments


dahorak:
Closing this in favour of https://github.com/red-hat-storage/ocs-ci/pull/11551

PR #11562, opened 2025-03-05
Skipuitest with kubeadmin for IBM Cloud v4.17

vasudha786:
Skipuitest with kubeadmin for IBM Cloud v4.17

### DIFF
--- a/tests/functional/ui/test_alert_text.py
+++ b/tests/functional/ui/test_alert_text.py
@@ -9,6 +9,7 @@ from ocs_ci.framework.pytest_customization.marks import (
     polarion_id,
     black_squad,
     tier2,
+    ui,
 )
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.cluster import is_hci_provider_cluster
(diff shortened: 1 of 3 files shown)


tests/functional/ui/test_alert_text.py
```diff
@@ -79,6 +80,7 @@ def alerts_expected():
 @skipif_hci_client
 @polarion_id("OCS-5509")
 @skipif_disconnected_cluster
+@ui
```
am-agrawa, line 83: Please update the marker name as ui could be misleading, to me it looks like it's an UI test but the definition says something else.

am-agrawa (changes requested):

tests/functional/ui/test_alert_text.py
petr-balogh: @am-agrawa I think this test_runbooks is a UI test - so it's a valid marker here.

It's navigating to specific alerting page and check there.

petr-balogh (approved):

tests/functional/ui/test_alert_text.py
am-agrawa: That's exactly what I wanted to understand? Since we are skipping it and not running it, using `ui` marker whose definition is to skip the test on IBM is really confusing to me.
Or did I not understand the implementation?

am-agrawa: @petr-balogh ^^

am-agrawa (approved):

tests/functional/ui/test_alert_text.py
petr-balogh: We sorted out over the direct message in slack

PR #11568, opened 2025-03-06
Fixed Reclaimspace tests case by adding sync operation after dd command

paraggit:
Fixed Reclaimspace tests case by adding sync operation after dd command


tests/functional/storageclass/test_storageclass_reclaim_space.py
```diff
@@ -89,7 +89,7 @@ def test_storageclass_reclaimspace(
             storage_path = pod_obj.get_storage_path("block")
             log.info("Writing 2.0GiB of data to the block device")
             pod_obj.exec_cmd_on_pod(
-                f"dd if=/dev/zero of={storage_path} bs=1M count=2048 oflag=direct > /dev/null 2>&1 &",
```
jilju, line 92: Will this address the failure in two places of the test case where the expected size is 2.0 and 0.0 G respectively ?

PR #11573, opened 2025-03-06
SKip test_2_nodes_different_types for compact mode deployments

### LINKED ISSUE #11572: Skip test_2_nodes_different_types on compact mode
Discussed with Santosh from Rook team for failures in ODF 4.18.

Quoting him:
```
If the TC is to drain one master and one worker at a time, but compact mode only has 3 master nodes, then I don't think it makes sense to have this test.
For compact modes, the TC should be to drain one storage node at a time.
```

Hence this test should be skipped on compact mode deployments.

am-agrawa:
fixes #11572


am-agrawa:
Test skipped on the pull request, refer logs.

suchita-g (approved):

Issue #11129, opened 2025-01-08
Bucketclass NSS creation via the MCG-CLI: Add coverage for Single AZ / AWS/ IBMCOS / GCP Namespacestores


sagihirshfeld:
During the research for addressing https://github.com/red-hat-storage/ocs-ci/issues/11026 it was found out that we're missing coverage for other platforms as well.

ypersky1980:
PR with the fix: https://github.com/red-hat-storage/ocs-ci/pull/11597

ypersky1980:
PR#11597 was closed. The relevant  PR with the fix is :  https://github.com/red-hat-storage/ocs-ci/pull/11603 
It was merged, therefore closing this issue.

PR #11597, opened 2025-03-10
Adding 4 new test cases and a support for nss creation for gcp platform

ypersky1980:
This PR handles requirements specified in https://github.com/red-hat-storage/ocs-ci/issues/11129 and there are 4 new test cases which were added: 

1) CLI NSS creation over Azure
2) CLI NSS creation over IBMCOS
3) CLI NSS creation over AWS
4) CLI NSS creation over GCP

In order to support creation of NSS over GCP a support was added to namespacestore.py.


sagihirshfeld (commented):
Please remove the newly added `on_prem_platform_required` marks since they will cause a skip of these tests when they shouldn't. This is only applicable for `RGW-OC-Single` and `RGW-CLI-Single`.

PR #11621, opened 2025-03-12
Enhance StorageConsumer cr and class

DanielOsypenko:
following https://github.com/red-hat-storage/ocs-operator/pull/3062 with this PR we enhance StorageConsumer cr and python class

### DIFF
--- a/ocs_ci/ocs/version.py
+++ b/ocs_ci/ocs/version.py
@@ -279,6 +279,23 @@ def main():
 
 
 def if_version(expressions):
+    """
+    Decorator to skip the function if the OCS version does not match the required version
+
+    ! Important note: This decorator is silently skipping the function
+    if the OCS version does not match the required version. Only warning message will be added to the log.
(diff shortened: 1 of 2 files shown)


DanielOsypenko:
tested locally. 
This change is not directly affecting existing test, since it only added additional helper functions

fbalak (approved):
It looks ok. I would suggest to change functions that return client info to return a list of statuses from all clients.

ocs_ci/ocs/version.py
```diff
@@ -275,3 +276,20 @@ def main():
         return
 
     report_ocs_version(cluster_version, image_dict, file_obj=sys.stdout)
+
+
+def if_version(expressions):
```
dahorak, line 281: Could you please add some doc string with explanation here?

DanielOsypenko: done, thanks!

dahorak (approved):

PR #11652, opened 2025-03-18
Add cnv dployment retries and increase timeouts

petr-balogh:


vavuthu (approved):

petr-balogh:
It has passed critical stage here:
https://url.corp.redhat.com/13191e3

Before it has failed on that - later on job above failed because stretch ceph cluster deployment failed - which is unrelated to this PR. So considering as verified.

shylesh (approved):
lgtm

Issue #11716, opened 2025-03-25
Boto3 operations against RGW fail due to SSL validation error in external mode with SSL RGW deployments


sagihirshfeld:
RP reference from the cloud manager that tries to create a new RGW bucket via boto3: https://url.corp.redhat.com/3c66c2a

PR #11717, opened 2025-03-25
In external deployments that use RGW w/ SSL - download the CA and set it as an environment variable for boto3 operations

sagihirshfeld:
Fixes: #11716


ypersky1980 (approved):
LGTM

PR #11733, opened 2025-03-26
Automate BZ 2302507: Test upgrade Noobaa with some bucket policy in place

mashetty330:
This PR covers the scenarios mentioned in the BZ 2302507:
- Test upgrade Noobaa with some bucket policy in place

### DIFF
--- a/tests/functional/object/mcg/test_bucket_policy.py
+++ b/tests/functional/object/mcg/test_bucket_policy.py
@@ -6,6 +6,7 @@ import botocore.exceptions as boto3exception
 import json
 import uuid
 
+from ocs_ci.ocs.cluster import CephCluster
 from ocs_ci.ocs.exceptions import (
     NoBucketPolicyResponse,
     InvalidStatusCode,
@@ -53,7 +54,10 @@ from ocs_ci.framework.pytest_customization.marks import (
     runs_on_provider,
     mcg,
     provider_mode,
+    post_upgrade,
+    pre_upgrade,
 )
+from ocs_ci.ocs.resources.storage_cluster import verify_backing_store
 from ocs_ci.utility import version
 from ocs_ci.utility.retry import retry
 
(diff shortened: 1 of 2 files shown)


hnallurv (dismissed):

tests/functional/object/mcg/test_bucket_policy.py
```diff
@@ -1286,3 +1291,65 @@ def test_supported_bucket_policy_operations(self, mcg_obj, bucket_factory):
...
+        assert (
+            obc_name and bucket_policy
+        ), "Seem like either pre-upgrade test for this failed or unable to cache the bucket/bucket policy info"
+
+        # Check noobaa health
+        logger.info("Verifying noobaa health")
+        CephCluster().noobaa_health_check()
+
+        # Check backingstore health
+        verify_backing_store()
```
sagihirshfeld, line 1351: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/resources/storage_cluster.py/#L1323-L1334

Might be a misunderstanding on my part, but this function doesn't seem to query the status of a specific backingstore. Couldn't this fail the test if we have an unrelated backingstore leftover in a bad state?

OdedViner (approved):

suchita-g (dismissed):

tests/functional/object/mcg/test_bucket_policy.py
mashetty330: You are right, I think i need fix this

tests/functional/object/mcg/test_bucket_policy.py
```diff
@@ -1286,3 +1291,65 @@ def test_supported_bucket_policy_operations(self, mcg_obj, bucket_factory):
         assert (
             not missing_policies
         ), f"Some bucket_policies are not created : {missing_policies}"
+
+
```
PrasadDesala, line 1295: add the customer bug jira id

mashetty330: @PrasadDesala Recently all the bugzilla ids were removed from all the test cases. On my discussion with Elad, he said there is no need to add bugzilla or jira ids if the bug is closed with fix.

tests/functional/object/mcg/test_bucket_policy.py
```diff
@@ -1286,3 +1291,65 @@ def test_supported_bucket_policy_operations(self, mcg_obj, bucket_factory):
...
+@polarion_id("OCS-6540")
+class TestNoobaaUpgradeWithBucketPolicy:
+    """
+    Test noobaa status post upgrade when there is bucket
+    with some bucket policy
+
+    """
+
+    @pre_upgrade
+    def test_create_bucket_policy_before_upgrade(
```
vavuthu, line 1305: add squad decorator

tests/functional/object/mcg/test_bucket_policy.py
```diff
@@ -1286,3 +1291,65 @@ def test_supported_bucket_policy_operations(self, mcg_obj, bucket_factory):
...
+        bucket_policy = json.dumps(bucket_policy_generated)
+
+        logger.info(
+            "Caching the bucket and bucket policy info for post upgrade verification"
+        )
+        request.config.cache.set("bucket_policy_bucket", obc_name)
+        request.config.cache.set("bucket_policy", bucket_policy)
+
+    @post_upgrade
+    def test_verify_noobaa_after_upgrade(self, request, mcg_obj_session):
```
vavuthu, line 1334: add squad decorator

mashetty330:
Verified here: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/49266/testReport/tests.functional.object.mcg.test_bucket_policy/TestNoobaaUpgradeWithBucketPolicy/

tests/functional/object/mcg/test_bucket_policy.py
PrasadDesala: We have been following the practice of including GSS bug IDs in the test for tracking purposes. Due to the transition from Bugzilla to Jira, I think we may need to revisit this approach and decide on the best way to capture this bug metadata. Perhaps mentioning the bug ID in the test case docstring could be one way to do this but there may also be other approaches we can consider, which we need to discuss and decide on

sagihirshfeld (approved):

tests/functional/object/mcg/test_bucket_policy.py
mashetty330: ack. for now i have mentioned the bug id in the test case doc strings

PrasadDesala (approved):

sagihirshfeld (approved):
LGTM

PR #11749, opened 2025-03-26
allow login to OCP via provided OCP URL and kubeadmin password

dahorak:
If `kubeconfig` file is not provided and there is `OCP_URL` and `KUBEADMIN_PASSWORD` environment variables set, it tries to login to the OCP cluster via the provided OCP URL and `kubeadmin` password and create `kubeconfig`.

This will be helpful for tests execution in environments where the `kubeconfig` is not available and rest of the automation rely on OCP URL and `kubeadmin` password.

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -16,6 +16,10 @@ RUN:
   log_dir: "/tmp"
   run_id: null  # this will be redefined in the execution
   kubeconfig_location: 'auth/kubeconfig' # relative from cluster_dir
+  # kubeadmin_password: '' # kubeadmin password used as alternative way to
+  # login to the OCP cluster (if kubeconfig is not available)
+  # ocp_url: '' # OCP Cluster URL (api or console) used to login to OCP cluster
+  # if kubeconfig is not available
   cli_params: {}  # this will be filled with CLI parameters data
   # If the client version ends with .nightly, the version will be exposed
   # to the latest accepted OCP nightly build version
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -548,24 +548,27 @@ def process_cluster_cli_params(config):
     if not os.path.isfile(kubeconfig_path) and not get_cli_param(
         config, "deploy", default=False
     ):
-        if os.environ.get("KUBEADMIN_PASSWORD") and os.environ.get("OCP_URL"):
+        if ocsci_config.RUN.get("kubeadmin_password") and ocsci_config.RUN.get(
+            "ocp_url"
+        ):
             log.info(
                 "Generating kubeconfig file from provided kubeadmin password and OCP URL"
             )
             # check and correct OCP URL (change it to API url if console url provided and add port if needed
-            ocp_api_url = os.environ.get("OCP_URL").replace(
+            ocp_api_url = ocsci_config.RUN.get("ocp_url").replace(
                 "console-openshift-console.apps", "api"
             )
             if ":6443" not in ocp_api_url:
                 ocp_api_url = ocp_api_url.rstrip("/") + ":6443"
 
             cmd = (
-                f"oc login --username {ocsci_config.RUN['username']} --password {os.environ['KUBEADMIN_PASSWORD']} "
+                f"oc login --username {ocsci_config.RUN['username']} "
+                f"--password {ocsci_config.RUN['kubeadmin_password']} "
                 f"{ocp_api_url} "
                 f"--kubeconfig {kubeconfig_path} "
                 "--insecure-skip-tls-verify=true"
             )
-            result = exec_cmd(cmd, secrets=(os.environ["KUBEADMIN_PASSWORD"],))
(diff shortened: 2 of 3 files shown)


ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -539,6 +540,45 @@ def process_cluster_cli_params(config):
...
         os.makedirs(cluster_path)
+
+    # create kubconfig if doesn't exists and OCP url and kubeadmin password is provided
+    kubeconfig_path = os.path.join(
+        cluster_path, ocsci_config.RUN["kubeconfig_location"]
+    )
+    if not os.path.isfile(kubeconfig_path) and not get_cli_param(
+        config, "deploy", default=False
+    ):
+        if os.environ.get("KUBEADMIN_PASSWORD") and os.environ.get("OCP_URL"):
```
petr-balogh, line 551: can we have this option passed rather via config value?

I see an issue with multicluster execution when we will need multiple clusters like this it's not simply possible and better to have this value taken from config per specific cluster I think.

Also we will probably need to iterate over every cluster to get this.

dahorak: Good point, I was taking this approach as the easiest one for integration in jenkins, but you are right that for multicluster scenario this will not work and both creating config for that and loading it from it will be easy and directly multicluster ready.

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -539,6 +540,48 @@ def process_cluster_cli_params(config):
     cluster_path = os.path.expanduser(cluster_path)
     if not os.path.exists(cluster_path):
         os.makedirs(cluster_path)
+
+    # create kubconfig if doesn't exists and OCP url and kubeadmin password is provided
+    kubeconfig_path = os.path.join(
+        cluster_path, ocsci_config.RUN["kubeconfig_location"]
+    )
+    if not os.path.isfile(kubeconfig_path) and not get_cli_param(
```
petr-balogh, line 548: should we do that for all clusters?

e.g. if ocsci_config.multicluster is set - we should iterate over all clusters to do this?

dahorak: I didn't tried it explicitly, but the `process_cluster_cli_params` function is called in `pytest_configure` [in following loop](https://github.com/red-hat-storage/ocs-ci/pull/11749/files#diff-81dd75ca7912048f20e991916771e11b16203494498849a39b6189401b3ac974L388-L393):
```python
    for i in range(ocsci_config.nclusters):
        log.info(f"Pytest configure switching to: cluster={i}")
        ocsci_config.switch_ctx(i)
```
So my understanding is that it should work out of the box.

petr-balogh (dismissed):

vavuthu (dismissed):

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -539,6 +540,48 @@ def process_cluster_cli_params(config):
...
+                log.warning(f"returncode: {result.returncode}")
+                log.warning(f"stdout: {result.stdout}")
+                log.warning(f"stderr: {result.stderr}")
+            else:
+                log.warning(f"Kubeconfig file were created: {kubeconfig_path}.")
+        else:
+            log.warning(
+                "Kubeconfig doesn't exists and RUN['kubeadmin_password'] and RUN['ocp_url'] "
+                "environment variables were not provided."
+            )
```
vavuthu, line 583: if we don't have kubeconfig, kubeadmin password and ocp_url, can we exit the run? not sure whether we can able to run tests with out above. Also can we change from warning to error

dahorak: Initially I don't want to change the current behaviour, but you are right, that there is probably no reason to continue with the execution and fail later with not much clear error.
I'll change it to raise exception and exit the execution.

dahorak: Updated in [be234ec](https://github.com/red-hat-storage/ocs-ci/pull/11749/commits/be234ec3d8f184f4ea63d5b44cb27333a9b8b12e).

vavuthu (approved):

dahorak:
Re-based against latest master to resolve PR checks issue.

ocs_ci/framework/conf/default_config.yaml
```diff
@@ -16,6 +16,10 @@ RUN:
   log_dir: "/tmp"
   run_id: null  # this will be redefined in the execution
   kubeconfig_location: 'auth/kubeconfig' # relative from cluster_dir
+  # kubeadmin_password: '' # kubeadmin password used as alternative way to
```
petr-balogh, line 19: Update values in:
https://github.com/red-hat-storage/ocs-ci/blob/master/conf/README.md

dahorak: Updated in [657d28f](https://github.com/red-hat-storage/ocs-ci/pull/11749/commits/657d28ff0cdeea9a3fa9394ac701699f09b92bc5).

petr-balogh (approved):

vavuthu (approved):

PR #11759, opened 2025-03-27
class and template for client profile

DanielOsypenko:
New python class and utilities introduced in order to cover product change: https://github.com/ceph/ceph-csi-operator/pull/165


ebondare (approved):

ocs_ci/ocs/constants.py
```diff
@@ -253,6 +253,12 @@
 ENCRYPTIONKEYROTATIONJOB = "encryptionkeyrotationjobs.csiaddons.openshift.io"
 DEFAULT_CEPH_DEVICECLASS = "defaultCephDeviceClass"
 CRD_KIND = "CustomResourceDefinition"
+"""
+ClientProfileSpec defines the desired state of Ceph CSI
+configuration for volumes and snapshots configured to use
+this profile
+"""
```
dahorak, line 260: What is the reason for using bare string instead of comment?
```suggestion
# ClientProfileSpec defines the desired state of Ceph CSI
# configuration for volumes and snapshots configured to use
# this profile
```

ocs_ci/ocs/resources/clientprofile.py
```diff
@@ -0,0 +1,118 @@
...
+            self.provider_context = None
+
+    def get_ceph_connection_reference(self):
+        """
+        Get the CephConnectionReference name
+
+        Returns:
+            str: CephConnectionReference name
+        """
+        with config.RunWithConfigContext(self.consumer_context):
```
dahorak, line 40: What would happen, if the `self.consumer_context` will be `None`? (Not sure, if that is possible, but assuming that yes, since the argument in `__init__` is optional, default to `None`.)
The same apply for the following occurrences too.

dahorak (approved):
Just two questions/comments, otherwise LGTM.

suchita-g (approved):

PR #11761, opened 2025-03-27
Test scale object expiration and verify obj mds

mashetty330:


tests/cross_functional/scale/noobaa/test_delete_objects.py
```diff
@@ -190,3 +202,122 @@ def test_delete_objects(
         # stop the io running in the background
         event.set()
         io_thread.result()
+
+    @pytest.fixture(scope="session")
+    def create_bucket_verify_object_deletion(
+        self,
+        request,
+        reduce_expiration_interval_session,
+        change_lifecycle_schedule_min_session,
```
sagihirshfeld, line 211: Do we really need to use the `session` scope fixtures do adjust MCG's background workers configuration?

This means that other tests might get affected by these conditions we only used as a workaround for testing expiration in a reasonable time.

tests/cross_functional/scale/noobaa/test_delete_objects.py
```diff
@@ -190,3 +202,122 @@ def test_delete_objects(
...
+        ],
+    )
+    def test_delete_objs_by_expiration_and_recursive_deletion(
+        self,
+        create_bucket_verify_object_deletion,
+        mcg_obj_session,
+        awscli_pod_session,
+        test_directory_setup,
+        reduce_expiration_interval,
+        change_lifecycle_schedule_min,
```
sagihirshfeld, line 282: I think we can omit these two - we're already using their session-scoped versions in `create_bucket_verify_object_deletion`. 

Alternatively we can omit them from `create_bucket_verify_object_deletion` and keep them here.

tests/cross_functional/scale/noobaa/test_delete_objects.py
mashetty330: 1. The reason why i used session scoped fixture here is because i have a session scoped fixture that is used to create and verify object expiration. If I use function scoped fixture within the session scope fixture, it's not really possible to call function scoped within session scope fixture. if I use function scope fixture in the test itself, then the expiration intervals are reset back to default which will affect the session scoped fixture of object expiration verification and it is not useful for us. Hope I didnt confuse you 😃 

2.  I dont think it will affect any other tests as this test will not be executed part of other test suites. This will only get executed when we call `scale_long` marker

tests/cross_functional/scale/noobaa/test_delete_objects.py
mashetty330: good catch!

tests/cross_functional/scale/noobaa/test_delete_objects.py
```diff
@@ -190,3 +202,120 @@ def test_delete_objects(
...
+        # Generate and upload objects to the
+        # bucket parallely
+        gen_empty_file_and_upload(
+            mcg_obj_session,
+            awscli_pod_session,
+            test_directory_setup.origin_dir,
+            amount=500000,
+            bucket=bucket.name,
+            threads=10,
+            timeout=7200,
```
ramkiperiy, line 302: 1. Regarding Timeout --> are we expecting the expiration will complete after 120hrs which is 5 days? or How about using timeout_sampler to iteratively check if the object policy is expired?
2. Is the expiration time is configurable for object? if so could we simulate with less expiration time?

mashetty330: @ramkiperiy 
1. So we have two parametrisation, hence test executes 2 times. Total execution time was about 10 hours. This includes object expiration verification taking 70% time which is happening at the end of the session scoped fixture. so it wont block other scale tests from executing. 
2.  It is only configurable for bucket. So we have configured 1day (it cant be less than this) but we are tweaking the noobaa background worker configuration to fasten this. Hence object expiration cleanup starts within few minutes.

ramkiperiy: 1. So even with expiration starts in few minutes the reason for long execution time is due to no of objects? If so TC expectation is to create lot of objs?
2. 1day expiration policy in bucket --> Any specific reason we can't reduce this?
Would like to understand if there is any option to optimise the test run time.

sagihirshfeld (approved):

vavuthu (approved):

PR #11119, opened 2025-01-07
Adding support in creation of bucket class with mcg cli and a new test case for creation of rgw namespacestore using mcg-cli 

ypersky1980:
This PR implements the following enhancement to cli_create_bucketclass method : in the past it was creating only backingstore with mcg-cli and from now on it will be able to create a bucketclass with mcg-cli command also for namespacestore. 

Following this enhancement it was possible to add a new test case which implements creation of namespacestore and bucketclass for rgw using mcg-cli commands. 

In the next PR I will implement automation of https://issues.redhat.com/browse/DFBUGS-1035 ( creating namespacestore and bucketclass on rgw bucket and writing IO to this bucket).

PR #11804, opened 2025-04-01
Automate DFBUGS-700: Verify NS health when rgw bucket used as backend storage

mashetty330:


tests/functional/object/mcg/test_namespace_store.py
```diff
@@ -0,0 +1,95 @@
...
+        logger.info(f"Created bucket {bucket.name}")
+
+        # Write objects to the bucket one by one
+        obj_list = write_random_objects_in_pod(
+            awscli_pod, test_directory_setup.origin_dir, amount=20, bs="10K"
+        )
+        write_individual_s3_objects(
+            mcg_obj,
+            awscli_pod,
+            bucket_factory,
```
sagihirshfeld, line 86: We don't need to pass `bucket_factory` as an argument here since we're already passing a `bucket_name`:

https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/bucket_utils.py/#L640-L641

tests/functional/object/mcg/test_namespace_store.py
```diff
@@ -0,0 +1,95 @@
...
+
+        # Create the bucketclass using the namespacestore created
+        bucketclass_dict = {
+            "interface": "OC",
+            "namespace_policy_dict": {
+                "type": "Single",
+                "namespacestores": [nss_obj],
+            },
+        }
+        bucket = bucket_factory(interface="CLI", bucketclass=bucketclass_dict)[0]
```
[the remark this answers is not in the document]
sagihirshfeld: I've now noticed that this way allows you to create the RGW bucket via OBC credentials instead of the RGW account's credentials. That said, after looking at the bug, I wouldn't consider this flow of creation a significant enough element to justify the potential leftover errors this will cause.

ypersky1980:
@mashetty330 Could you please add a description for this PR ( edit first comment)?

suchita-g (dismissed):

tests/functional/object/mcg/test_namespace_store.py
mashetty330: @sagihirshfeld we need to pass because its not an optional argument. check the function definition

tests/functional/object/mcg/test_namespace_store.py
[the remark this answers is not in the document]
mashetty330: @sagihirshfeld Here I want to use OBC credentials not RGW credentials, hence I have used this method. Where as #11119 uses MCG credentials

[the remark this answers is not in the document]
mashetty330: @sagihirshfeld About the leftovers from NSS object, I can use teardown factory fixture for the cleanup.

tests/functional/object/mcg/test_namespace_store.py
sagihirshfeld: now I see it... my bad 😅

tests/functional/object/mcg/test_namespace_store.py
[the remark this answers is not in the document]
sagihirshfeld: @mashetty330 cool, I wasn't aware of the `teardown_factor` method until now.

However, you still need to manage the order of fixture teardown so you get the following order:
1. `bucket_factory` teardown - clear the OBC and bucketclass
2. `teardown_factory` teardown - clear the NSS
3. `rgw_factory` teardown - clear the underlying RGW bucket

Otherwise some deletions might get blocked - like the NSS that won't be deleted because it's still being used by the bucketclass.

To achieve this you just need to rearrange these fixtures in the reverse order at the test func arguments:
```
    def test_namespacestore_with_rgw(
        self,
        mcg_obj,
        rgw_bucket_factory,      --> created 1st deleted 3rd
        teardown_factory,          --> created 2nd deleted 2nd
        bucket_factory,               --> created 3rd deleted 1st
        test_directory_setup,
        awscli_pod,
    ):
```

tests/functional/object/mcg/test_namespace_store.py
```diff
@@ -0,0 +1,102 @@
...
+    cli_create_namespacestore,
+    NamespaceStore,
+)
+from ocs_ci.framework.pytest_customization.marks import polarion_id, mcg, red_squad
+from ocs_ci.ocs.ocp import OCP
+from ocs_ci.ocs.resources.ocs import OCS
+
+logger = logging.getLogger(__name__)
+
+
```
PrasadDesala, line 22: Do we need to include the tier marker as well?

tests/functional/object/mcg/test_namespace_store.py
```diff
@@ -0,0 +1,102 @@
...
+@red_squad
+@mcg
```
PrasadDesala, line 24: since the test is specific to rgw, should we skip it to run on cloud platforms?

tests/functional/object/mcg/test_namespace_store.py
```diff
@@ -0,0 +1,102 @@
...
+            },
+        }
+        bucket = bucket_factory(interface="CLI", bucketclass=bucketclass_dict)[0]
+        logger.info(f"Created bucket {bucket.name}")
+
+        # Write objects to the bucket one by one
+        obj_list = write_random_objects_in_pod(
+            awscli_pod, test_directory_setup.origin_dir, amount=20, bs="10K"
+        )
+        write_individual_s3_objects(
```
PrasadDesala, line 82: how many objects are we uploading to the MCG bucket?

mashetty330: 20 objects

tests/functional/object/mcg/test_namespace_store.py
mashetty330: yes will add that marker

tests/functional/object/mcg/test_namespace_store.py
mashetty330: yes will add that

tests/functional/object/mcg/test_namespace_store.py
```diff
@@ -0,0 +1,110 @@
...
+logger = logging.getLogger(__name__)
+
+
+@red_squad
+@mcg
+class TestNamespaceStore:
+
+    @tier2
+    @polarion_id("OCS-6550")
+    @baremetal_deployment_required
```
PrasadDesala, line 35: can't we use on_prem_platform_required? it has a broader list of platforms where RGW can run

PrasadDesala (approved):

sagihirshfeld (approved):
LGTM

Issue #9294, opened 2024-02-06
Boto3-related issue #9250 requires a permanent solution


sagihirshfeld:
As described in #9250, upgrading boto3 from `1.14.7` to `1.34.14` broke the current way we use AWS buckets across different regions. 

Due to urgency, in #9257 we downgraded to `1.24.96` which is the latest version that supports both our current use of OCS-CI and other recent package updates that were required due to security needs.

However, since we can expect a need to upgrade boto3 again in the future, we need to adjust OCS-CI so we can use multi-region AWS buckets with the latest version of boto3.

sagihirshfeld:
Fixed by [#13384](https://github.com/red-hat-storage/ocs-ci/pull/13384)

PR #11812, opened 2025-04-01
Upgrade boto3 version and add dynamic boto3 region change for AWS buckets to `cloud_manager.py`

sagihirshfeld:
Fixes: #9294

Changes:
- Upgrade boto3 to the latest version
- Under `cloud_manager.py::S3Client` - use a wrapper class that uses boto3 via composition to dynamically recreate the boto3 client/resource when commands are failing due to region issues.
- Standardize the use of the new wrapper class under `cloud_manager.py::S3Client`

Note that IBMCOS and RGW buckets are both also managed by instances of the `cloud_manager.py::S3Client` class, and that these changes were made so the way they are managed is not affected.

### DIFF
--- a/ocs_ci/ocs/resources/boto3_wrapper.py
+++ b/ocs_ci/ocs/resources/boto3_wrapper.py
@@ -1,0 +1,176 @@
+import logging
+
+import boto3
+from botocore.exceptions import ClientError
+
+
+logger = logging.getLogger(name=__file__)
+
+
+class BaseBoto3Wrapper:
+    """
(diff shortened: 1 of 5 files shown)


sagihirshfeld:
The verification of this PR requires we show the stability of both MCG tests that use various underlying storage buckets, as well as tests that make use of the new boto3 version - such as deployment.

sagihirshfeld:
Closing in favor of https://github.com/red-hat-storage/ocs-ci/pull/13384

PR #11838, opened 2025-04-03
Check odf prerequisites on client connect host network

DanielOsypenko:
following dev change https://github.com/red-hat-storage/ocs-operator/pull/3096 added check on network preferences

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -1563,6 +1563,7 @@ class Deployment(object):
         # rules to be enabled on underlaying platform).
         if config.DEPLOYMENT.get("host_network"):
             cluster_data["spec"]["hostNetwork"] = True
+            logger.info("Host network is enabled")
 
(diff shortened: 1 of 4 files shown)


DanielOsypenko:
Tested manually in python console.
**Need to check on fresh deployment of single client cluster.**

jilju (approved):
Please try to use a different name instead of "hub" cluster. This is to distinguish easily between ACM hub and the storage cluster.

ebondare (approved):

DanielOsypenko:
> Please try to use a different name instead of "hub" cluster. This is to distinguish easily between ACM hub and the storage cluster.

I am confused too, but I follow the terminology that comes from Ohad. Hub cluster and Application cluster. 
It is not Provider anymore, so should we call it Storage Cluster? It will cause confusion with StoratgeCluster cr.

DanielOsypenko:
verified: BUILD ID: 4.19.0-26 RUN ID: 1743918093

suchita-g (approved):

dahorak (approved):

PR #11851, opened 2025-04-06
Add check on csi drivers

DanielOsypenko:
This completes the Test case with title:
[ODF] CSI Driver is owned by client operator not by Rook after convergence in both internal and external mode


DanielOsypenko:
tested locally via python console

sagihirshfeld (approved):

suchita-g (approved):

dahorak (approved):

PR #11251, opened 2025-01-29
[4.18 system test] MCG bucket notification

mashetty330:
This PR covers the system test automation for the ODF 4.18 feature "Support bucket notification in Multicloud Object Gateway - GA"

PR #11870, opened 2025-04-08
[Release-4.18] System test automation for MCG bucket notification feature 

mashetty330:
This PR covers the system test automation for the ODF 4.18 feature "Support bucket notification in Multi-cloud Object Gateway - GA"

Original PR (#11251 ) raised against master (release-4.19 as of today) is closed because of the change in noobaa db HA implementation. This would break many of the MCG system test regression runs as the step to perform Noobaa db backup and recovery will be completely revamped.


mashetty330:
This is no longer needed

PR #11888, opened 2025-04-09
add skipif

### LINKED ISSUE #11887: skip test_mgr_enable_rook_backend_module test case
https://ibm-systems-storage.slack.com/archives/C06DWD30QTZ/p1743501807062239?thread_ts=1743500696.407249&cid=C06DWD30QTZ 

The rook mgr module is really only used by the upstream dashboard, which we don't use downstream. There is not currently a scenario to enable this module in the downstream. When it is enabled (though we are not sure the cause), it causes errors as we have seen in a couple customer clusters.

vkathole:
Fixes #11887

### DIFF
--- a/tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
+++ b/tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
@@ -20,8 +20,7 @@ log = logging.getLogger(__name__)
 @tier2
 @brown_squad
 @skipif_external_mode
-@skipif_ocs_version("<4.15")
-@skipif_ocs_version(">=4.18")
+@skipif_ocs_version(["<4.15", ">4.17"])
 @jira("DFBUGS-1284")
 @pytest.mark.polarion_id("OCS-6240")
 class TestMgrRookModule(ManageTest):


tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
```diff
@@ -20,7 +20,7 @@
 @tier2
 @brown_squad
 @skipif_external_mode
-@skipif_ocs_version("<4.15")
+@skipif_ocs_version("<4.15 or >=4.18")
```
am-agrawa, line 23: Is it supported on 4.16, 4.17 and 4.18?

tests/functional/pod_and_daemons/test_mgr_enable_rook_backend_module.py
```diff
@@ -21,6 +21,7 @@
 @brown_squad
 @skipif_external_mode
 @skipif_ocs_version("<4.15")
+@skipif_ocs_version(">=4.18")
```
PrasadDesala, line 24: can you check if it can be combined below way?
@skipif_ocs_version(["<4.15", ">4.17"])

PrasadDesala (approved):

avd-sagare (approved):

vavuthu (approved):

PR #11909, opened 2025-04-14
Fix disk size propagation in terraform ocp ai deployment vsphere

dahorak:


dahorak:
Verification job: https://url.corp.redhat.com/68fb54e
The deployment failed because of a bug https://url.corp.redhat.com/4f4cdce, but the VMs were created correctly with the expected size.

clacroix12 (approved):

PR #11926, opened 2025-04-16
new onboarding procedure for client

DanielOsypenko:
This PR covers changes provided with Epic https://issues.redhat.com/browse/RHSTOR-6664
StorageConsumer should be created before the StorageClient and there are changes in onboarding procedure.

### DIFF
--- a/ocs_ci/deployment/hosted_cluster.py
+++ b/ocs_ci/deployment/hosted_cluster.py
@@ -33,10 +33,12 @@ from ocs_ci.ocs.resources.csv import check_all_csvs_are_succeeded
 from ocs_ci.ocs.resources.ocs import OCS
 from ocs_ci.ocs.resources.pod import (
     wait_for_pods_to_be_in_statuses_concurrently,
+    get_pod_logs,
 )
 from ocs_ci.ocs.resources.storageconsumer import (
     create_storage_consumer_on_default_cluster,
 )
+from ocs_ci.ocs.utils import get_pod_name_by_pattern
 from ocs_ci.ocs.version import if_version
 from ocs_ci.utility import templating, version
 from ocs_ci.utility.deployment import get_ocp_ga_version
@@ -172,8 +174,20 @@ class HostedClients(HyperShiftBase):
                 ):
                     client_installed = hosted_odf.setup_storage_client()
                 else:
+                    start_time = time.time()
                     client_installed = hosted_odf.setup_storage_client_converged(
                         storage_consumer_name=f"consumer-{cluster_name}"
+                    )
+                    time_taken = time.time() - start_time
+                    time_sec = int(time_taken % 60) + 1
+                    provider_server_pod = get_pod_name_by_pattern(
+                        "ocs-provider-server"
+                    )[0]
+                    logs = get_pod_logs(
+                        pod_name=provider_server_pod, since=f"{time_sec}s"
+                    )
+                    logger.info(
+                        f"Logs from provider-server pod:\n******************\n{logs}\n******************\n"
                     )
                 client_setup_res.append(client_installed)
                 if client_installed:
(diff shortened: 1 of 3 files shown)


DanielOsypenko:
verification run:
https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/view/all/job/qe-odf-provider-client-multicluster/203/

Please consider this run successful. There is ongoing discussion that StorageClass distribution that does not happen and fail the last stage is a bug. 
https://ibm-systems-storage.slack.com/archives/C081L5H77H7/p1745251776651109?thread_ts=1744817098.195109&cid=C081L5H77H7

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -660,6 +745,25 @@ def create_host_inventory():
     logger.info("Created InfraEnv.")
 
 
+def get_onboarding_token_from_secret(secret_name):
```
fbalak, line 748: There could be some parameter with retry and wait so it can be used without the wait.

ocs_ci/templates/ocs-deployment/provider-mode/storage_consumer.yaml
```diff
@@ -1,15 +1,18 @@
 apiVersion: ocs.openshift.io/v1alpha1
 kind: StorageConsumer
 metadata:
+  namespace: openshift-storage
```
fbalak, line 4: Why are changes to this template necessary?

fbalak (approved):

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -660,6 +745,25 @@ def create_host_inventory():
...
+    """
+    ocp_obj = OCP(
+        kind="secret",
+        namespace=config.ENV_DATA["cluster_namespace"],
+        resource_name=secret_name,
+    )
+    secret_obj = ocp_obj.get(retry=6, wait=10, silent=True)
+    return secret_obj.get("data").get("onboarding-token")
+
+
```
amr1ta, line 766: So we will be able to only fetch the onboarding token from the secret ? Generate onboarding token option will not be available?

DanielOsypenko: Based on how I understand Leela's words this is by design starting from 4.17. We did onboarding incorrectly, generating secret by our own. But I did not check on older versions

ocs_ci/templates/ocs-deployment/provider-mode/storage_consumer.yaml
DanielOsypenko: this CR was created 7 of April. I missed the namespace initially, assuming it will be changed frequently and hence be set in python script. This default namespace removes couple lines from code

ocs_ci/deployment/hosted_cluster.py
DanielOsypenko: we have retry few lines bellow
`secret_obj = ocp_obj.get(retry=6, wait=10, silent=True)`

amr1ta (approved):

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -660,6 +745,25 @@ def create_host_inventory():
...
+    Returns:
+        str: Onboarding token
+    """
+    ocp_obj = OCP(
+        kind="secret",
+        namespace=config.ENV_DATA["cluster_namespace"],
+        resource_name=secret_name,
+    )
+    secret_obj = ocp_obj.get(retry=6, wait=10, silent=True)
+    return secret_obj.get("data").get("onboarding-token")
```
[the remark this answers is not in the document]
DanielOsypenko: replaced with `return secret_obj.get("data", {}).get("onboarding-token")
`

dahorak (approved):

PR #11952, opened 2025-04-20
generate and update multicluster config on flight

DanielOsypenko:
### Generic solution aligned task 

This PR does: 
1. rework for fixture create_hypershift_clusters, splitting it to 3 different fixtures:
a) create_hypershift_clusters_push_config - will create hosted cluster with default preferences, provision it with similar OCP and ODF versions that Provider has
b) create_config_of_existing_hosted_clusters - will add config only if cluster exist in deployment but multicluster config does not
c) refresh_hosted_config - will replace configs of existing clusters with newly generated

2.  rework for fixture destroy_hosted_cluster, splitting it to 2 different fixtures
a) destroy_hosted_cluster
b) hosted_cluster_remove_config
 **it allows us to remove config during test suit execution, when hitting unhealthy cluster** 
 
-----

Pr is a necessary rework that partially allows us to run all tests dynamically editing a MultiCluster config, updating it with desired state and prevent cascade skip of the execution when one of the clusters is unhealthy. 

**Combining use of these fixtures we can run the execution agnostic of the mode—whether it's Internal, Converged, or Converged with clients and even change mode from Internal to Converged with clients.**

* 20/4/2025 - tested in python console

### DIFF
--- a/ocs_ci/deployment/helpers/hypershift_base.py
+++ b/ocs_ci/deployment/helpers/hypershift_base.py
@@ -794,8 +794,12 @@ def create_cluster_dir(cluster_name):
     Returns:
         str: Path to the kubeconfig directory
     """
-    from ocs_ci.ocs.constants import CLUSTERS_PATH
 
-    path = os.path.join(CLUSTERS_PATH, cluster_name, "openshift-cluster-dir")
+    path = os.path.join(
+        config.ENV_DATA["cluster_path"],
+        "clusters",
+        cluster_name,
+        "openshift-cluster-dir",
+    )
     os.makedirs(path, exist_ok=True)
     return path
--- a/ocs_ci/deployment/helpers/hypershift_base.py
+++ b/ocs_ci/deployment/helpers/hypershift_base.py
@@ -50,6 +50,27 @@ def get_hosted_cluster_names():
     return exec_cmd(cmd, shell=True).stdout.decode("utf-8").strip().split()
 
 
+def get_available_hosted_clusters_to_ocp_ver_dict():
+    """
+    Get available HyperShift hosted clusters with their versions
+
+    Returns:
+        dict: hosted clusters available with their versions. Example: {'cl-418-x': '4.18.7', 'cl-418-c': '4.19.0'}
+    """
+    logger.info("Getting HyperShift hosted clusters available")
+    cmd = (
+        "oc get hostedclusters -n clusters -o json | "
+        "jq -r '.items[] | "
+        'select(.metadata.annotations["hypershift.openshift.io/HasBeenAvailable"] == "true") | '
+        '"\\(.metadata.name)|\\(.status.version.history[0].version)"\''
+    )
+    out = exec_cmd(cmd, shell=True).stdout.decode("utf-8").strip()
+    if not out:
+        return {}
+    result = dict(line.split("|") for line in out.splitlines())
+    return result
+
+
 def kubeconfig_exists_decorator(func):
     """
     Decorator to check if the kubeconfig exists before executing the decorated method
@@ -170,6 +191,19 @@ def get_hosted_cluster_type(cluster_name=None):
         kind=constants.HOSTED_CLUSTERS, namespace="clusters", resource_name=cluster_name
     )
     return ocp_hosted_cluster_obj.get()["spec"]["platform"]["type"].lower()
+
+
+def get_current_nodepool_size(name):
+    """
+    Get existing nodepool of HyperShift hosted cluster
+    Args:
+        name (str): name of the cluster
+    Returns:
+         str: number of nodes in the nodepool
+    """
+    logger.info(f"Getting existing nodepool of HyperShift hosted cluster {name}")
+    cmd = f"oc get --namespace clusters nodepools | awk '$1==\"{name}\" {{print $4}}'"
+    return exec_cmd(cmd, shell=True).stdout.decode("utf-8").strip()
 
 
 class HyperShiftBase:
@@ -490,20 +524,6 @@ class HyperShiftBase:
         )
         return validation_passed
 
-    def get_current_nodepool_size(self, name):
-        """
-        Get existing nodepool of HyperShift hosted cluster
-        Args:
-            name (str): name of the cluster
-        Returns:
-             int: number of nodes in the nodepool
-        """
-        logger.info(f"Getting existing nodepool of HyperShift hosted cluster {name}")
-        cmd = (
-            f"oc get --namespace clusters nodepools | awk '$1==\"{name}\" {{print $4}}'"
-        )
-        return exec_cmd(cmd, shell=True).stdout.decode("utf-8").strip()
-
     def worker_nodes_deployed(self, name: str):
         """
         Check if worker nodes are deployed for HyperShift hosted cluster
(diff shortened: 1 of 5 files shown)


fbalak (approved):

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -50,6 +50,27 @@ def get_hosted_cluster_names():
     return exec_cmd(cmd, shell=True).stdout.decode("utf-8").strip().split()
 
 
+def get_available_hosted_clusters_to_ocp_ver_dict():
+    """
+    Get HyperShift hosted clusters available
```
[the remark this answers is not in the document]
DanielOsypenko: thanks. Will address with following PR. I have additional functionality on my mind to implement

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -50,6 +50,27 @@ def get_hosted_cluster_names():
...
+    Returns:
+        dict: hosted clusters available with their versions. Example: {'cl-418-x': '4.18.7', 'cl-418-c': '4.19.0'}
+    """
+    logger.info("Getting HyperShift hosted clusters available")
+    cmd = (
+        "oc get hostedclusters -n clusters -o json | "
+        "jq -r '.items[] | "
+        'select(.metadata.annotations["hypershift.openshift.io/HasBeenAvailable"] == "true") | '
+        '"\\(.metadata.name)|\\(.status.version.history[0].version)"\''
+    )
```
dahorak, line 66: I'm thinking if it wouldn't be better to parse the output json directly in python. This is nice once-liner, but it might be little bit tricky to debug any issues there if there will be any change in the output or any other issue.

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -172,6 +193,19 @@ def get_hosted_cluster_type(cluster_name=None):
...
+def get_current_nodepool_size(name):
+    """
+    Get existing nodepool of HyperShift hosted cluster
+    Args:
+        name (str): name of the cluster
+    Returns:
+         str: number of nodes in the nodepool
+    """
+    logger.info(f"Getting existing nodepool of HyperShift hosted cluster {name}")
+    cmd = f"oc get --namespace clusters nodepools | awk '$1==\"{name}\" {{print $4}}'"
```
dahorak, line 205: Generally I would prefer to parse the json instead of the plain output as it will be more robust, but I know it is not possible in all the cases (not sure about this particular case). So take it more like suggestion for the future.

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -764,3 +782,20 @@ def install_hypershift_upstream_on_cluster(self):
...
+        cluster_name (str): Name of the cluster
+
+    Returns:
+        str: Path to the kubeconfig directory
+    """
+    from ocs_ci.ocs.constants import CLUSTERS_PATH
+
+    path = os.path.join(CLUSTERS_PATH, cluster_name, "openshift-cluster-dir")
+    os.makedirs(path, exist_ok=True)
+    return path
```
dahorak, line 801: I'm thinking about this approach of creating the cluster dirs dynamically in the user's home and I see couple of issues there - if it will local execution, the user might not want to create the directory and files there (completely outside of his/her working directories) and for Jenkins execution, any data in those directories will be lost after the cluster deletion, because those directories will not be mounted from the Shared storage.

What about placing the cluster directories somewhere to the main cluster directory?
E.g. create a `clusters` directory inside the main cluster directory (`config.ENV_DATA.get("cluster_path")`) - this will solve both the issues in the meaning, that all the data for the execution will be together in the working directory of the existing cluster and for Jenkins the main cluster path is already on the shared storage, so also the subdirectories will be preserved.

Is it clear, what I mean?

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -376,18 +384,35 @@ def verify_hosted_ocp_clusters_from_provider(self):
...
-    def download_hosted_clusters_kubeconfig_files(self):
+    def download_hosted_clusters_kubeconfig_files(self, cluster_names_paths_dict=None):
         """
-        Get HyperShift hosted cluster kubeconfig for multiple clusters
+        Get HyperShift hosted cluster kubeconfig for multiple clusters.
+        Provided cluster_names_paths_dict will always be a default source of cluster names and paths
+
+        Args:
+            cluster_names_paths_dict (dict): Optional argument. Function will download all kubeconfigs to specified in
+            configuration folders or download specific clusters kubeconfig to specified in arguments folder
```
dahorak, line 394: This sentence seems to require some update to make it clear.

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -1491,3 +1516,152 @@ def odf_csv_installed(self):
...
+
+    hosted_clients_obj = HostedClients()
+    logger.info(f"Factory duty is '{duty}'")
+
+    if duty == "create_hosted_cluster_push_config":
+        hosted_cluster_conf_on_provider = {"ENV_DATA": {"clusters": {}}}
+        for cluster_name in cluster_names:
+            # this configuration is necessary to deploy hosted cluster, but not for running tests with multicluster job
+            hosted_cluster_conf_on_provider["ENV_DATA"]["clusters"][cluster_name] = {
+                "hosted_cluster_path": f"~/clusters/{cluster_name}/openshift-cluster-dir",
```
dahorak, line 1552: See my comment under the `create_cluster_dir` function.

ocs_ci/deployment/helpers/hypershift_base.py
DanielOsypenko: yes, I think I understood. Thanks

ocs_ci/deployment/helpers/hypershift_base.py
DanielOsypenko: I will address it later if you allow. it is old method it was just refactored to become a function, for easier access.

ocs_ci/deployment/helpers/hypershift_base.py
DanielOsypenko: tbh if I can continue with this one-liner for now and focus on enhancements later it will be more effective time management. Overloaded with Converged testing.

ocs_ci/deployment/hosted_cluster.py
DanielOsypenko: done

DanielOsypenko:
https://github.com/user-attachments/assets/3f9b1113-49a7-469b-b8c8-80440a97e110

works as expected after fix.
@dahorak please check that cluster folder change does not affect pipeline

ocs_ci/ocs/constants.py
```diff
@@ -337,6 +337,8 @@
 NON_HOSTED_CLUSTER = "non_hosted"
 
 # provider mode constants
+CLUSTERS_PATH = os.path.join(os.path.expanduser("~"), "clusters")
+
```
dahorak, line 341: I think this constant is not used any more.

dahorak (approved):

dahorak (approved):

ocs_ci/ocs/constants.py
DanielOsypenko: removed

PR #11971, opened 2025-04-22
enable console plugin on clients via patch

DanielOsypenko:
This is not independent change.
Tested in python console.

```
import ocs_ci.framework
config = ocs_ci.framework.config
config.ENV_DATA["cluster_path"] = "/Users/danielosypenko/Work/automation_4/ocs-ci/cluster_path"
config.RUN["kubeconfig"] = "/Users/danielosypenko/Work/automation_4/ocs-ci/cluster_path/auth/kubeconfig"
from ocs_ci.ocs.resources.storageconsumer import *
config.RUN["resource_checker"] = False
from ocs_ci.deployment.hosted_cluster import (
    HypershiftHostedOCP,
    HostedODF,
    HostedClients,
)
config.ENV_DATA["platform"] = "hci_baremetal"
from ocs_ci.deployment.hosted_cluster import HostedODF
h = HostedODF("cl-418-x")
h.cluster_kubeconfig = "/Users/danielosypenko/clusters/cl-418-x/openshift-cluster-dir/auth/kubeconfig"
h.wait_console_plugin_pod_running()
True
h.enable_client_console_plugin()
True
```

### DIFF
--- a/ocs_ci/deployment/hosted_cluster.py
+++ b/ocs_ci/deployment/hosted_cluster.py
@@ -32,7 +32,9 @@ from ocs_ci.ocs.resources.csv import check_all_csvs_are_succeeded
 from ocs_ci.ocs.resources.ocs import OCS
 from ocs_ci.ocs.resources.pod import (
     wait_for_pods_to_be_in_statuses_concurrently,
+    wait_for_pods_to_be_running,
 )
+from ocs_ci.ocs.utils import get_pod_name_by_pattern
 from ocs_ci.ocs.version import if_version
 from ocs_ci.utility import templating, version
 from ocs_ci.utility.deployment import get_ocp_ga_version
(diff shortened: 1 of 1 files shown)


am-agrawa (approved):

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -1491,3 +1497,50 @@ def odf_csv_installed(self):
...
+    def enable_client_console_plugin(self):
+        """
+        Enable the ODF client console plugin by patching the console operator
+
+        Returns:
+            bool: True if the patch is applied successfully, False otherwise
+        """
+        try:
+            self.exec_oc_cmd(
+                "patch console.operator cluster --type json "
```
jilju, line 1511: I think this is already implemented in https://github.com/red-hat-storage/ocs-ci/blob/f7ca514dc95b78cd88be98df89674cfa366c5ed0/ocs_ci/ocs/resources/storage_client.py#L130

dahorak (approved):

ocs_ci/deployment/hosted_cluster.py
DanielOsypenko: I was thinking to reuse it, but in this case I need redo `enable_console_plugin` because it does not apply with the client kubeconfig. And Provider client deployment does not use multicluster config for deployment but just use modified `self.exec_oc_cmd` to solve issue with setting up new client

ebondare (approved):

PR #11994, opened 2025-04-25
[RDR] Neutral ACM hub failure and recovery

am-agrawa:
When the active ACM hub cluster is at a neutral site and goes down, we perform hub recovery and switch to the passive hub using backup and restore. After successful hub recovery, the primary managed cluster is brought down and workloads are failedover. 
Post successful failover, we recover the down managed cluster and ensure it undergoes cleanup of all the failedover workloads. Post cleanup, we relocate these workloads back to their original primary cluster.


am-agrawa:
Moved to tier2 after discussing with @hnallurv and @keesturam

keemano (approved):

PR #12007, opened 2025-04-27
Pr 11864

DanielOsypenko:
Based on solution designed by @fbalak 

Adding marker "@run_on_all_clients_push_missing_configs" make:
1. Run test on single cluster without spending additional time, performing and reporting same way
2. Run test on multiple storage client cluster, gathering configurations and pushing them to MultiClusterConfig. Running only on clients

### DIFF
--- a/ocs_ci/deployment/hosted_cluster.py
+++ b/ocs_ci/deployment/hosted_cluster.py
@@ -1824,13 +1824,33 @@ def hypershift_cluster_factory(
                 ] = running_ocp_version
 
             cluster_path = create_cluster_dir(cluster_name)
-            kubeconf_path = (
+            logger.info("Cluster path: %s", cluster_path)
+            def_client_config_dict["ENV_DATA"]["cluster_path"] = cluster_path
+
+            kubeconf_paths = (
                 hosted_clients_obj.download_hosted_clusters_kubeconfig_files(
                     {cluster_name: cluster_path}
                 )
             )
+            if not len(kubeconf_paths):
+                from ocs_ci.framework.exceptions import ClusterKubeconfigNotFoundError
 
-            logger.info(f"Kubeconfig path: {kubeconf_path}")
+                ClusterKubeconfigNotFoundError(
+                    f"Failed to download kubeconfig for cluster {cluster_name}"
+                )
+            else:
+                kubeconf_path = [
+                    path for path in kubeconf_paths if cluster_name in path
+                ][0]
+            logger.debug(f"Kubeconfig path: {kubeconf_path}")
+
+            logger.debug(
+                "Setting default context to config. Every config should have same default context"
+            )
+            def_client_config_dict["ENV_DATA"].setdefault(
+                "default_cluster_context_index", 0
+            )
+
             def_client_config_dict.setdefault("RUN", {}).update(
                 {"kubeconfig": kubeconf_path}
             )
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -395,7 +395,9 @@ hci_provider_and_client_required = pytest.mark.skipif(
 )
 run_on_all_clients = pytest.mark.run_on_all_clients
 try:
-    client_indexes = [pytest.param(idx) for idx in config.get_consumer_indexes_list()]
+    client_indexes = [
+        pytest.param(*[idx]) for idx in config.get_consumer_indexes_list()
+    ]
     run_on_all_clients = pytest.mark.parametrize(
         argnames=["cluster_index"], argvalues=client_indexes
     )
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -395,8 +395,11 @@ hci_provider_and_client_required = pytest.mark.skipif(
 )
 run_on_all_clients = pytest.mark.run_on_all_clients
 try:
+    client_indexes = [
+        pytest.param(*[idx]) for idx in config.get_consumer_indexes_list()
+    ]
     run_on_all_clients = pytest.mark.parametrize(
-        argnames=["cluster_index"], argvalues=config.get_consumer_indexes_list()
+        argnames=["cluster_index"], argvalues=client_indexes
     )
 except ClusterNotFoundException:
     pass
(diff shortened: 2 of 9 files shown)


DanielOsypenko:
execution one cluster (shortened version of test, but we see it runs on single cluster as previously was running on internal mode): 
[single_cluster_ruwith_mark.txt](https://github.com/user-attachments/files/19930423/single_cluster_ruwith_mark.txt)

execution on Provider with multiple clusters:
`run-ci: --cluster-path ${CLUSTER_PATH} --ocp-version 4.18 --ocs-version 4.18 --cluster-name ${CLUSTER_NAME}  tests/functional/pv/pv_services/test_rwop_pvc.py::TestRwopPvc::test_pod_with_same_priority -m tier1 -k "" --dev-mode --ocsci-conf /Users/danielosypenko/Work/automation_4/ocs-ci/conf/deployment/fusion_hci_pc/provider_bm_upi_1az_rhcos_nvme_3m_3w.yaml `

[console_log_info.txt](https://github.com/user-attachments/files/19930426/console_log_info.txt)

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -1824,13 +1824,33 @@ def hypershift_cluster_factory(
...
+            if not len(kubeconf_paths):
+                from ocs_ci.framework.exceptions import ClusterKubeconfigNotFoundError
+
+                ClusterKubeconfigNotFoundError(
+                    f"Failed to download kubeconfig for cluster {cluster_name}"
+                )
+            else:
+                kubeconf_path = [
+                    path for path in kubeconf_paths if cluster_name in path
+                ][0]
```
dahorak, line 1844: If I understand it correctly, this expects that the cluster name is used in the path for the kubeconfig, which we can probably assume, but there might be a problem with a situation when one cluster name will be a substring of another cluster name, right? For example `cluster-a` and `cluster-aa`.

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -1824,13 +1824,33 @@ def hypershift_cluster_factory(
...
+            logger.info("Cluster path: %s", cluster_path)
+            def_client_config_dict["ENV_DATA"]["cluster_path"] = cluster_path
+
+            kubeconf_paths = (
                 hosted_clients_obj.download_hosted_clusters_kubeconfig_files(
                     {cluster_name: cluster_path}
                 )
             )
+            if not len(kubeconf_paths):
+                from ocs_ci.framework.exceptions import ClusterKubeconfigNotFoundError
```
dahorak, line 1836: Could be this import done in the beginning of the file?

ocs_ci/deployment/hosted_cluster.py
DanielOsypenko: good point.
We can try to read kubeconfig.clusters.name
Kubeconfig has list of clusters and one of them have matching name, others are dedicated for routes

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -392,6 +393,46 @@
...
+        push_missing_configs: Boolean flag to push missing configs
+
+    Returns:
+        Parametrized marker or original marker if setup fails
+    """
+    try:
+        if push_missing_configs:
+            from ocs_ci.deployment.hosted_cluster import hypershift_cluster_factory
+
+            hypershift_cluster_factory(
```
ebenahar, line 415: probably unrelated to this PR, but the concept of ODF client clusters is not tightly coupled to HCP hosted cluster. ODF client can be installed on non HCP hosted OCP cluster

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -392,6 +393,46 @@
...
+                duty="use_existing_hosted_clusters_push_missing_configs",
+            )
+        client_indexes = [
```
ebenahar, line 418: is this function specific to client clusters or to multi clusters in general?

tests/conftest.py
```diff
@@ -577,6 +590,41 @@ def pytest_fixture_setup(fixturedef, request):
             for mark in request.node.iter_markers():
                 if mark.name == "config_index":
                     ocsci_config.switch_ctx(mark.args[0])
+    _switch_context_helper(request)
+
+
+def _switch_context_helper(request):
```
ebenahar, line 596: why is it defined as a private function? Can't it be used outside this file?

DanielOsypenko: code was extracted from the pytest_fixture_setup to reuse in finalizers. 
This function was to use it in module only in fixture functions and helpers where request arg is visible. Can't see now where it can be reused outside of the module.
But this can change in future, we can set it public when use case will appear.

ocs_ci/framework/pytest_customization/marks.py
DanielOsypenko: This concept was started few years back, I did not think about agent and external cluster having a role as client. Purpose then, when it was new was to make object entities of Cluster that aggregates capabilities of Hypershift OCP cluster, deployment helper functions and ODF client

ocs_ci/deployment/hosted_cluster.py
DanielOsypenko: fixed in https://github.com/red-hat-storage/ocs-ci/pull/12041

ocs_ci/framework/pytest_customization/marks.py
DanielOsypenko: setup_multicluster_marker is a helper function specific for multicluster scenarios, it does use significant time to execute, but we rely it. It makes possible during the execution to reduce number of client clusters in system from many to 0 or increase from 0 to many

DanielOsypenko: we have moved from this draft to https://github.com/red-hat-storage/ocs-ci/pull/12041, it is updated and verified

ocs_ci/framework/pytest_customization/marks.py
DanielOsypenko: we have moved from this draft to https://github.com/red-hat-storage/ocs-ci/pull/12041, it is updated and verified

tests/conftest.py
DanielOsypenko: we have moved from this draft to https://github.com/red-hat-storage/ocs-ci/pull/12041, it is updated and verified

DanielOsypenko:
merged https://github.com/red-hat-storage/ocs-ci/pull/12041 containing changes from current PR

PR #12032, opened 2025-04-29
Fix test_automated_recovery_from_stopped_node_and_start[True] on clusters with 6W

### LINKED ISSUE #11771: test_automated_recovery_from_stopped_node_and_start[True] fails on VSPHERE6-IPI-IN-TRANSIT-ENCRYPTION deployments
Failed on ODF 4.18 as it is unable to surpass `get_node_rack_or_zone`

```
  node_names = set([node.name for node in nodes])
E   AttributeError: 'NoneType' object has no attribute 'name'



ocs_ci/ocs/platform_nodes.py:3020: AttributeError
```

RP- https://url.corp.redhat.com/734f7ce

Quoting Itzhak from our offline discussion:

```
I see in the node output https://url.corp.redhat.com/734d1b9 there are 2 worker nodes without rack. This is because the run was with 6 worker nodes. The function get_node_rack_or_zone returns ValueError when it doesn't find the node rack. This will not happen if we run with 3 worker nodes. Anyway, one fix is to add a param in the function get_node_rack_or_zone so it will not raise an exception when it doesn't find the rack or zone.
```

am-agrawa:
fixes #11771

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -2705,7 +2705,7 @@ def get_another_osd_node_in_same_rack_or_zone(
     return osd_node_in_same_rack_or_zone
 
 
-def get_nodes_racks_or_zones(failure_domain, node_names):
+def get_nodes_racks_or_zones(failure_domain, node_names, raise_value_error=True):
     """
     Get the nodes racks or zones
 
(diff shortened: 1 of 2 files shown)


am-agrawa:
Passed on cluster with 6W- https://url.corp.redhat.com/bff279c

suchita-g (approved):

PR #12042, opened 2025-04-30
Refactor determination of storageclass for deployments

clacroix12:
Issue: https://url.corp.redhat.com/ce3ae0e

### DIFF
--- a/ocs_ci/deployment/helpers/storage_class.py
+++ b/ocs_ci/deployment/helpers/storage_class.py
@@ -34,10 +34,10 @@ def get_storageclass() -> str:
         "customized_deployment_storage_class"
     )
 
-    storage_class = DEFAULT_STORAGE_CLASS_MAP.get(platform)
-
     if customized_deployment_storage_class:
         storage_class = customized_deployment_storage_class
+    else:
+        storage_class = DEFAULT_STORAGE_CLASS_MAP[platform]
 
     logger.info(f"Using storage class: {storage_class}")
     return storage_class
--- a/ocs_ci/deployment/helpers/storage_class.py
+++ b/ocs_ci/deployment/helpers/storage_class.py
@@ -1,0 +1,64 @@
+import logging
+import yaml
+
+from ocs_ci.framework import config
+from ocs_ci.framework.logger_helper import log_step
+from ocs_ci.ocs import constants
+from ocs_ci.utility.utils import run_cmd
+
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_STORAGE_CLASS_MAP = {
+    constants.AWS_PLATFORM: "gp2-csi",
+    constants.IBMCLOUD_PLATFORM: "ibmc-vpc-block-10iops-tier",
+    constants.VSPHERE_PLATFORM: "thin-csi",
+    constants.AZURE_PLATFORM: "managed-csi",
+    constants.GCP_PLATFORM: None,
+    constants.ROSA_HCP_PLATFORM: None,
+    constants.RHV_PLATFORM: "ovirt-csi-sc",
+}
+
+
+def get_storageclass() -> str:
+    """
+    Retrieve the storageclass to use from the config or based on platform
+
+    Returns:
+        str: Name of the storageclass
+
+    """
+    logger.info("Getting storageclass")
+    platform = config.ENV_DATA.get("platform")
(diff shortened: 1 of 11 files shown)


ocs_ci/deployment/helpers/storage_class.py
```diff
@@ -0,0 +1,64 @@
...
+
+DEFAULT_STORAGE_CLASS_MAP = {
+    constants.AWS_PLATFORM: "gp2-csi",
+    constants.IBMCLOUD_PLATFORM: "ibmc-vpc-block-10iops-tier",
+    constants.VSPHERE_PLATFORM: "thin-csi",
+    constants.AZURE_PLATFORM: "managed-csi",
+    constants.GCP_PLATFORM: None,
+    constants.ROSA_HCP_PLATFORM: None,
+    constants.RHV_PLATFORM: "ovirt-csi-sc",
+}
```
dahorak, line 20: The `thin-csi` value is also available as `constants.THIN_CSI_STORAGECLASS`, but I'm not sure, if it is worth to change all the values and get them from constants, so maybe as it is now is ok.

ocs_ci/deployment/helpers/storage_class.py
```diff
@@ -0,0 +1,64 @@
...
+        str: Name of the storageclass
+
+    """
+    logger.info("Getting storageclass")
+    platform = config.ENV_DATA.get("platform")
+    customized_deployment_storage_class = config.DEPLOYMENT.get(
+        "customized_deployment_storage_class"
+    )
+
+    storage_class = DEFAULT_STORAGE_CLASS_MAP.get(platform)
```
dahorak, line 37: Wouldn't be better to use `DEFAULT_STORAGE_CLASS_MAP[platform]`, so it will directly raise a clear `KeyError:` exception in case we will implement new platform or is `None` a reasonable "default"?

In the case of raising `KeyError` exception for missing `platform`, it might be worth to combine this with the following condition for `customized_deployment_storage_class`, so in case the `platform` will be missing in the map, but `customized_deployment_storage_class` will be defined, it will not raise the exception.

ocs_ci/deployment/vmware.py
```diff
@@ -129,6 +118,11 @@ def __init__(self):
         self.cluster_launcer_repo_path = os.path.join(
             constants.EXTERNAL_DIR, "v4-scaleup"
         )
+        if version.get_semantic_ocp_version_from_config() >= version.VERSION_4_13:
```
dahorak, line 121: You can probably remove this condition and run the following code directly as this condition will be always evaluated to true for the supported versions.

dahorak (approved):
Just some small questions/suggestions otherwise LGTM.

petr-balogh (approved):

petr-balogh:
Daniel had few valuable comments to consider, but I am also OK with this change - but if you will incorporate Daniel's suggestions I will re-approve.

ocs_ci/deployment/helpers/storage_class.py
clacroix12: I originally had it this way but went with `.get` due to failures in [test_cloud.py](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/deployment/tests/test_cloud.py). I could revert back to using a direct call to the key and change the test class to just set the proper attribute so it doesn't hit any issues.

For the `customized_deployment_storage_class` case I did add keys for the platforms which leverage this but I do think it's a good idea to check both before raising the error so I will incorporate that as well.

clacroix12: I moved the check for `customized_deployment_storage_class` above setting the default value so we will check if that exists as a priority, and changed getting the default value to directly access the key.

ocs_ci/deployment/vmware.py
clacroix12: Removed

clacroix12:
A small change here that I want to mention: The default platform in our default_config is now `aws` as opposed to `AWS`. This was causing an issue with the `test_cloud` unit tests now that the platform is being used to retrieve the default storageclass. I suspect that nothing else in the framework was expecting the platform to be in all caps since our constant `AWS_PLATFORM` is lower case as well as the value in all of our aws specific config files, but I wanted to mention it since it's a change to the default config.

dahorak (approved):

PR #12044, opened 2025-05-02
Fixed metadata test case

paraggit:
IN 4.19 , metadata tests failing due to pod provisioner  pods name has changed , this fix will works for that.


paraggit:
This Fix will applicable after the FIX of https://issues.redhat.com/browse/DFBUGS-2395

paraggit:
Another PR opened for the same FIX

Issue #11244, opened 2025-01-28
[RDR] Do not fail RDR deployment when HEALTH_WARN 1 slow ops is hit. Mon restart and re-check health to continue deployment.


am-agrawa:
Sometimes during RDR deployment, we see slow ops issue which fails the Deployment and thus resulting in incomplete RDR deployment.
This happens on either of the managed clusters after ODF installation. We should add a workaround so as to have a successful deployment in this particular case.

```
2025-01-27 22:44:59  >           raise CephHealthException(f"Ceph cluster health is not OK. Health: {health}")
2025-01-27 22:44:59  E           ocs_ci.ocs.exceptions.CephHealthException: Ceph cluster health is not OK. Health: HEALTH_WARN 1 slow ops, oldest one blocked for 924 sec, mon.e has slow ops
2025-01-27 22:44:59  
2025-01-27 22:44:59  ocs_ci/utility/utils.py:2543: CephHealthException
```

PR #12049, opened 2025-05-04
Fix slow ops issue in rdr deployment

petr-balogh:
Fixes: #11244

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -2558,6 +2558,25 @@ def ceph_health_resolve_crash():
     archive_ceph_crashes(ct_pod)
 
 
+def ceph_health_resolve_mon_slow_ops(health_status):
+    """
+    Fix ceph health issue with mon slow ops
+    """
+    log.warning("Trying to fix the issue with mon slow ops by restarting mon pod")
+    mon_pattern = r"mon\.([a-z]) has slow ops"
+    match = re.search(mon_pattern, health_status)
+    mon_id = None
+    if match:
+        mon_id = match.group(1)
+        log.warning(f"Problematic MON ID with slow ops: {mon_id} will be restarted")
+    if mon_id:
+        from ocs_ci.ocs import ocp
+
+        ocp.OCP().exec_oc_cmd(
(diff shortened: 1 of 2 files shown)


am-agrawa (dismissed):

ocs_ci/utility/utils.py
```diff
@@ -2558,6 +2558,25 @@ def ceph_health_resolve_crash():
...
+    """
+    log.warning("Trying to fix the issue with mon slow ops by restarting mon pod")
+    mon_pattern = r"mon\.([a-z]) has slow ops"
+    match = re.search(mon_pattern, health_status)
+    mon_id = None
+    if match:
+        mon_id = match.group(1)
+        log.warning(f"Problematic MON ID with slow ops: {mon_id} will be restarted")
+    if mon_id:
+        from ocs_ci.ocs import ocp
```
vavuthu, line 2573: can we move this at the start of this function if it has circular dependency issue

petr-balogh: Yes we have circular dependency issue, and I prefer to keep it here, as in case we don't get mon_id we don't need to import as we will not need ocp module to be imported.

vavuthu (dismissed):

prsurve (dismissed):

ocs_ci/utility/utils.py
```diff
@@ -2558,6 +2558,25 @@ def ceph_health_resolve_crash():
     archive_ceph_crashes(ct_pod)
 
 
+def ceph_health_resolve_mon_slow_ops(health_status):
+    """
+    Fix ceph health issue with daemon crash
```
am-agrawa, line 2563: Do we have daemon crash when slow ops is seen?

dahorak (approved):

ocs_ci/utility/utils.py
petr-balogh: copy paste issue, fixed

vavuthu (approved):

am-agrawa (approved):

dahorak (approved):

PR #12066, opened 2025-05-05
Skip `TestDefaultBackingstoreOverride` in external deployments

sagihirshfeld:
This test suite attempts to clone the existing default backingstore and set the clone as the new default.

In external deployments, this process requires additional configuration steps that add complexity without meaningful benefit. This PR skips these scenarios to streamline the test flow and avoid spending time on a low-value edge case.


vavuthu (approved):

suchita-g (approved):

sagihirshfeld:
Adding the Verified label due to the limited nature of the change

udaysk23 (approved):

PR #12069, opened 2025-05-06
Skipping ibmcloud testcase for machine not found error: v4.17

vasudha786:
Skipping ibmcloud testcase for machine not found error: v4.17

### DIFF
--- a/tests/functional/z_cluster/nodes/test_node_replacement_proactive.py
+++ b/tests/functional/z_cluster/nodes/test_node_replacement_proactive.py
@@ -327,6 +327,7 @@ class TestNodeReplacementTwice(ManageTest):
       2. ceph side host still on the old rack
     """
 
+    @skipif_ibm_cloud_managed
     def test_nodereplacement_twice(self):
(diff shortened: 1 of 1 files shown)


tests/functional/z_cluster/nodes/test_node_replacement_proactive.py
```diff
@@ -327,6 +327,7 @@ class TestNodeReplacementTwice(ManageTest):
       2. ceph side host still on the old rack
     """
 
+    @skipif_ibm_cloud_managed
```
prsurve, line 330: Instead of skipping, why not fix it in ocs-ci?

petr-balogh: My guess is that in IBM ROKS managed platform the node operations are very limited and this might be the case here that this is not possible to implement on IBM CLOUD Managed AKA ROKS platform. But @vasudha786 know better.

petr-balogh (approved):

PR #12083, opened 2025-05-06
Make sure run_id is unique by using miliseconds instead of only seconds

dahorak:
When the `run_id` is taken just from seconds timestamp, there is a possibility that two runs will get the same `run_id`:
* https://url.corp.redhat.com/209791b
* https://url.corp.redhat.com/6cc73a3

This situation breaks the reporting to Report Portal and might cause another issues.
To minimize this risk, we can use miliseconds as run_id.

### DIFF
--- a/ocs_ci/utility/framework/fusion_fdf_init.py
+++ b/ocs_ci/utility/framework/fusion_fdf_init.py
@@ -201,7 +201,7 @@ def generate_run_id() -> int:
 
     """
     logger.debug("Generating run_id from timestamp")
-    run_id = int(time.time())
+    run_id = int(time.time() * 1000)
     config.RUN["run_id"] = run_id
     return run_id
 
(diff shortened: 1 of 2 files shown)


ocs_ci/utility/framework/fusion_fdf_init.py
```diff
@@ -200,8 +201,8 @@ def generate_run_id() -> int:
         int: Unique identifier for the run
 
     """
-    logger.debug("Generating run_id from timestamp")
-    run_id = int(time.time())
+    logger.debug("Generating run_id from timestamp and random number")
+    run_id = int(time.time()) * 1000 + random.randint(0, 1000)
```
petr-balogh, line 205:
```suggestion
    run_id = int(time.time() * 1000)
```
As time.time() returns something like:
1746544165.797612

You can make it quite easy with milliseconds included instead of using random. Then the chance we hit the same millisecond is probably impossible

petr-balogh (approved):

vavuthu (approved):

PR #12084, opened 2025-05-06
Update the noobaa_db_backup_and_recovery_locally fixture to use the new noobaa-db HA solution

sagihirshfeld:


udaysk23 (dismissed):

tests/cross_functional/conftest.py
```diff
@@ -215,18 +209,72 @@ def factory(
...
         # Login to the NooBaa DB pod and cleanup potential database clients to nbcore
-        query = "SELECT pg_terminate_backend (pid) FROM pg_stat_activity WHERE datname = 'nbcore';"
-        noobaa_db_pod.exec_cmd_on_pod(
-            command=f'psql -h 127.0.0.1 -p 5432 -U postgres -c "{query}"'
+        nb_db_pg_cluster_app_secret = OCP(
+            kind=constants.SECRET,
+            resource_name=constants.NB_DB_CNPG_APP_SECRET,
         )
-        logger.info("Cleaned up potential database clients to nbcore!")
+        pg_password = b64decode(
```
mashetty330, line 216: @sagihirshfeld what is this password? how different is this from running normal db query?

mashetty330 (commented):
Thanks @sagihirshfeld for addressing this quickly.

tests/cross_functional/conftest.py
sagihirshfeld: I got an authorization error for this specific command without it. This command wasn't a part of the old or new procedures, so I tried to leave it as-is in case it was a specific test requirement I wasn't aware of:
```
"SELECT pg_terminate_backend (pid) FROM pg_stat_activity WHERE datname = 'nbcore';"
```

sagihirshfeld: Apparently we didn't need this password to run this query - I just switched to using `exec_nb_db_query` in https://github.com/red-hat-storage/ocs-ci/pull/12084/commits/ffc97edfc0d891f4228f18d5cd7d33cd1dc1141e and it still works, as can be seen in the below passing verification job

PrasadDesala (approved):

PR #12085, opened 2025-05-06
install latest hypershift operator to support 4.19 guest OCP

DanielOsypenko:
To support unreleased versions of OCP we need follow one of three options

1. install MCE 2.9 from pre-release (need quay access for it)
2. use Latest hypershift operator [quay.io/hypershift/hypershift-operator:latest](http://quay.io/hypershift/hypershift-operator:latest)
3. put annotation on the HostedCluster to ignore this validation

Hypershift team recommended to follow the 2nd option as we can not access unreleased upstream MCE versions. 
This was implemented here.

New ENV_DATA config var introduced: `install_latest_hypershift_operator`, by default False/None

### DIFF
--- a/ocs_ci/deployment/helpers/hypershift_base.py
+++ b/ocs_ci/deployment/helpers/hypershift_base.py
@@ -211,6 +211,43 @@ def get_current_nodepool_size(name):
     return exec_cmd(cmd, shell=True).stdout.decode("utf-8").strip()
 
 
+def update_hypershift_operator(version="latest"):
+    """
+    Update the hypershift operator by extracting the latest image and installing it.
+    This function is also used to update the existing hypershift operator.
+    One of the suggested options to support unreleased guest OCP clusters
+
+    Args:
+        version (str): The version of the hypershift operator to install. Default is 'latest'.
+
+    """
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+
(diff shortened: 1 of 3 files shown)


fbalak (approved):

suchita-g (approved):

jilju (dismissed):

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -206,6 +206,43 @@ def get_current_nodepool_size(name):
...
+            logger.info("Installing the hypershift operator")
+            cmd = (
+                f"{hypershift_binary_path} install --hypershift-image {constants.HCP_REGISTRY}:{version}",
+            )
+            exec_cmd(
+                cmd,
+                shell=True,
+            )
+        except Exception as e:
+            logger.error(f"Failed to install/update hypershift operator: {e}")
```
dahorak, line 248: In case of any error, is it ok to just print the exception to the log and continue with the execution?

DanielOsypenko: yes, it is by design, since it will not bring any valuable bug on ODF, but can fail installation of other dependencies or fail installation of a kubevirt cluster n-1 which should pass even on non-upgraded  hypershift.

PR #12096, opened 2025-05-07
Fix issue with node labels[Replica-1]

AviadP:
This fix one of few issues in Replica1 test

### DIFF
--- a/tests/functional/storageclass/test_replica1.py
+++ b/tests/functional/storageclass/test_replica1.py
@@ -24,6 +24,9 @@ from ocs_ci.ocs.constants import (
     VOLUME_MODE_BLOCK,
     CSI_RBD_RAW_BLOCK_POD_YAML,
     DEFALUT_DEVICE_CLASS,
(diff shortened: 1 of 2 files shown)


AviadP:
Verified locally. it clear one issue, but test might be fail due to another issues. i want to keep it small focused PRs

paraggit (dismissed):

tests/functional/storageclass/test_replica1.py
```diff
@@ -39,32 +43,113 @@
...
     get_failure_domains,
 )
+from ocs_ci.ocs.node import get_worker_nodes, get_node_objs
 
 
 log = getLogger(__name__)
 
 
-def create_pod_on_failure_domain(project_factory, pod_factory, failure_domain: str):
+def _get_node_selector_for_failure_domain(
```
ebenahar, line 52: this function should be taken outside the test file into a helper / utils file

paraggit (approved):

PR #12098, opened 2025-05-07
Update noobaa-db queries at test_unidirectional_bucketclass_replication and add jira decorator for bug

sagihirshfeld:
- Update the noobaa-db queries to be made against the primary noobaa-db CNPG pod
- Add the Jira decorator for this regression bug that fails the test: https://issues.redhat.com/browse/DFBUGS-2353

### DIFF
--- a/tests/functional/object/mcg/test_bucket_replication.py
+++ b/tests/functional/object/mcg/test_bucket_replication.py
@@ -3,7 +3,6 @@ import gc
 
 import pytest
 
-from ocs_ci.framework import config
 from ocs_ci.framework.pytest_customization.marks import (
     tier1,
     tier2,
@@ -16,7 +15,7 @@ from ocs_ci.framework.pytest_customization.marks import (
 from ocs_ci.ocs.exceptions import CommandFailed
 from ocs_ci.framework.testlib import MCGTest
 from ocs_ci.utility.retry import retry
-from ocs_ci.utility.utils import TimeoutSampler
+from ocs_ci.utility.utils import TimeoutSampler, exec_nb_db_query
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.bucket_utils import (
     compare_bucket_object_list,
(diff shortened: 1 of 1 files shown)


sagihirshfeld:
The test will be skipped in any verification run. However, I can show that locally the test has passed two noobaa-db queries, and then failed at a later stage due to the bug:
```
        replication_id = exec_nb_db_query(
            f"SELECT data ->> 'replication_policy_id' FROM buckets WHERE data ->> 'name'='{source_bucket_name}';"
        )[0].strip()
    
        # write random objects to the bucket
        write_random_test_objects_to_bucket(
            awscli_pod_session,
            source_bucket_name,
            test_directory_setup.origin_dir,
            amount=5,
            mcg_obj=mcg_obj_session,
        )
    
        if target_bucketclass is None:
            # Delete the target bucket
            target_bucket.delete()
            logger.info(f"Deleted target bucket {target_bucket_name}")
    
            def check_nb_endpoint_logs():
                for pod in get_noobaa_endpoint_pods():
                    pod_logs = get_pod_logs(pod.name)
                    if (
                        "The AWS access key Id you provided does not exist in our records"
                        in pod_logs
                    ):
                        return True
                    del pod_logs
                    gc.collect()
                return False
    
            sample = TimeoutSampler(timeout=360, sleep=10, func=check_nb_endpoint_logs)
            assert sample.wait_for_func_status(
                result=False
            ), "Error 'The AWS access key Id you provided does not exist in our records' found in the endpoint logs"
        else:
            # delete the bucket
            source_bucket.delete()
            logger.info(f"Deleted source bucket {source_bucket_name}")
    
            # check in db that the replication config was deleted
            replication_conf_count = exec_nb_db_query(
                f"SELECT COUNT (*) FROM replicationconfigs WHERE _id='{replication_id}'"
            )[0].strip()
    
>           assert (
                int(replication_conf_count) == 0
            ), f"Replication config for {source_bucket_name} is not deleted!"
E           AssertionError: Replication config for oc-bucket-72b1deae0407403bb6c91554aa0c5f is not deleted!
E           assert 1 == 0
E             +1
E             -0
```

Based off of that, I'm setting the Verified label. Once the bug is fixed, the test will no longer be skipped and then should pass.

udaysk23 (approved):

shivamdurgbuns (approved):

suchita-g (approved):

PR #12107, opened 2025-05-08
Remove cephblockpoolradosnamespaces before cephblockpool

avd-sagare:
Followed Steps mentioned in below PR-
https://github.com/red-hat-storage/ocs-operator/pull/3196#issuecomment-2838980820

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -945,6 +945,16 @@ def ceph_pool_factory_fixture(request, replica=3, compression=None):
 
         for instance in instances:
             try:
+                radosns_obj = ocp.OCP(
+                    kind=constants.CEPHBLOCKPOOLRADOSNS, namespace=instance.namespace
+                )
+                radosnamespaces = [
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -945,7 +945,17 @@ def finalizer():
 
         for instance in instances:
             try:
-                instance.delete()
+                instance.delete(wait=False)
```
jilju, line 948: Should this be after deleting radosnamespaces ?

jilju (approved):

tests/conftest.py
avd-sagare: As per interim fix added by leela -
when cbp is marked for deletion in that case only Radosnamespace will get deleted otherwise it will get reconsiled. 
Ones  Radosnamespace get deleted cbp which is marked for deletion will  also get deleted.

dahorak (approved):

suchita-g (approved):

Issue #11143, opened 2025-01-14
stabilize test_ui_storage_size_post_resize_osd 4.16-4.18


DanielOsypenko:
test_ui_storage_size_post_resize_osd on assertion:

The total UI size 6144Gi is not in the expected total size range range(12276, 12300)Gi

```
def check_storage_size_is_reflected_in_ui():
    """
    Check that the current total storage size is reflected in the
    UI 'ocs-storagecluster-storagesystem' page.

    """
    block_and_file = (
        PageNavigator()
        .nav_odf_default_page()
        .nav_storage_systems_tab()
        .nav_storagecluster_storagesystem_details()
        .nav_block_and_file()
    )
    used, available = block_and_file.get_raw_capacity_card_values()
    # Get the used, available and total size in bytes
    used_size_bytes = human_to_bytes_ui(used)
    available_size_bytes = human_to_bytes_ui(available)
    total_size_bytes = used_size_bytes + available_size_bytes

    # Convert the used, available and total size to GB
    bytes_to_gb = 1024**3
    used_size_gb = used_size_bytes / bytes_to_gb
    available_size_gb = available_size_bytes / bytes_to_gb
    total_size_gb = round(total_size_bytes / bytes_to_gb)
    logger.info(f"Used size = {used_size_gb}Gi")
    logger.info(f"Available size = {available_size_gb}Gi")
    logger.info(f"Total size = {total_size_gb}Gi")

    ceph_cluster = CephCluster()
    ceph_capacity = int(ceph_cluster.get_ceph_capacity())
    ceph_total_size = ceph_capacity * len(get_osd_pods())

    # There could be a small gap between the total size in the UI and the actual Ceph total size.
    # So, instead of checking the accurate size, we check that the total size is within the expected range.
    max_gap = 6 if ceph_total_size < 1500 else 12
    expected_total_size_range_gb = range(
        ceph_total_size - max_gap, ceph_total_size + max_gap
    )
    logger.info(
        f"Check that the total UI size {total_size_gb}Gi is in the "
        f"expected total size range {expected_total_size_range_gb}Gi"
    )

  assert total_size_gb in expected_total_size_range_gb, (
        f"The total UI size {total_size_gb}Gi is not in the "
        f"expected total size range {expected_total_size_range_gb}Gi"
    )
E       AssertionError: The total UI size 6144Gi is not in the expected total size range range(12276, 12300)Gi
```

DanielOsypenko:
[...]

Issue #12109, opened 2025-05-08
[Bug]: check_storage_size_is_reflected_in_ui is miss calculating the storage total capacity


shyRozen:
### Description
In check_storage_size_is_reflected_in_ui the function using get_ceph_capacity which is getting from "ceph df" command the total_bytes, convert it to Gi and divided by replica. When it returns it is multiplied by the number of OSDs:
    ceph_total_size = ceph_capacity * len(get_osd_pods())
 Which is wrong and should be multiplied again with the replica number.
### Steps to Reproduce
1. 
2.
3.

### Actual Behavior
ceph total capacity is returned divided to replica and then multiplied by OSDs numbers
### Expected Behavior
ceph total capacity is returned divided to replica and should be multiplied again to replica
### Impact (likelihood of reproduction, impact on the cluster and on other tests, etc.)
test_ui_storage_size_post_resize_osd is failing to calculate.
### Environment
- Test Suite(s):
- Platform(s):
- Version(s):
- OS:

PR #12110, opened 2025-05-08
Fix calcualtion of check_storage_size_is_reflected_in_ui

shyRozen:
[https://github.com/red-hat-storage/ocs-ci/issues/12109](https://github.com/red-hat-storage/ocs-ci/issues/12109)

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -662,24 +662,27 @@ class CephCluster(object):
         # module does not exist, return 0 as number of replica.
         return 0
 
-    def get_ceph_capacity(self):
+    def get_ceph_capacity(self, replica_divide=True):
         """
         The function gets the total mount of storage capacity of the ocs cluster.
-        the calculation is <total bytes> / <replica number>
+        the calculation is <total bytes> / <replica number> depends if replia_divide is true
         it will not take into account the current used capacity.
-
+        Args:
+            replica_divide (bool): if true it will divide the capacity in to replica else return the capacity as it.
         Returns:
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/cluster.py
```diff
@@ -682,6 +682,22 @@ def get_ceph_capacity(self):
 
         return usable_capacity
 
+    def get_ceph_total_capacity(self):
```
ebenahar, line 685: isn't it similar to https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/cluster.py#L665?

shyRozen: The one you mentioned is for some reason dividing the total capacity into replica, I just dropped the replica so it will have the real total capacity. This has cause all the failures because when it got back to the function that called it, it multiplied it be osd count for some reason, so I thought why divide and then get again the replica (need to create again a cephcluster object), so just copied the function and dropped the /replica divide to simplify the testcase.

ebenahar: can't we use the existing function and add a condition for not deciding by replica if not needed?

shyRozen: We can, it just that it is used many times and I didn't want to harm any other testcase. I'll do it.

shyRozen: Fixed.

ebenahar (approved):

yitzhak12 (approved):

yitzhak12:
fix https://github.com/red-hat-storage/ocs-ci/issues/11143

PR #12130, opened 2025-05-12
Use delete_all_objects_in_batches wherever we used bucket.objects.all.delete()

sagihirshfeld:
Add the `delete_all_objects_in_batches` utility function, and use it wherever we previously used bucket.objects.all().delete() via an S3 resource. 

- bucket.objects.all() loads all the objects into memory
- deleting in batches avoids overwhelming the S3 backend


ocs_ci/ocs/bucket_utils.py
```diff
@@ -3280,3 +3282,46 @@ def _check_objs_deletion():
...
+    logger.info(
+        f"Starting batch deletion in bucket '{bucket_name}' with prefix '{prefix}'"
+    )
+
+    objects_to_delete = []
+    total_deleted = 0
+
+    for obj in bucket.objects.filter(Prefix=prefix):
+        objects_to_delete.append({"Key": obj.key})
+        if len(objects_to_delete) >= batch_size:
```
petr-balogh, line 3312:
```suggestion
        if objects_to_delete:
```

Is it returning maximum 1000 of objects when we do iteration over bucket.objects.filter(Prefix=prefix)?
Does this allow us to set the number of the entires per query? if so batch_size should be used there to only get that many objects.

Then till we still have some objects in next iteration it should just delete them as well, so you don't even need to hold objects to delete I think. And you can avoid the smae cod from line 3319 to again delete those objects.

Also we should probably use some try except block on delete , and in case we fail to delete some objects, try to add them to some list failed_to_delete - and then at the end check if this list is not empty then maybe do other attempt to delete those - or raise an exception. So it will at least try to delete as many objects it can and than fail at the end if we failed to remove some.

OdedViner (approved):

ocs_ci/ocs/bucket_utils.py
sagihirshfeld: Thanks for the review Petr - I switched to using pagination, which fetches a 1000 objects by default (we can only set it to fetch less, and I don't see the point).

I was also able to avoid repeating myself thanks to it, and I added a collection of failed uploads so we retry again once more at the end

petr-balogh (approved):

PR #12180, opened 2025-05-16
Pin oslo.utils version

clacroix12:
We are seeing issues with our destroy jobs due to this dependency. I suspect something weird happening with the version that was released today so I am pinning this to the last version we were using successfully.

Example failure: https://url.corp.redhat.com/8858210


clacroix12:
Verified we did not encounter the same issue: https://url.corp.redhat.com/54b028c

petr-balogh (approved):

vavuthu (approved):

dahorak (approved):

PR #12194, opened 2025-05-19
Test to verify that an existing PV is not used for a new PVC when PV name is not given in PVC yaml

jilju:
Test case to verify that an existing PV is not used for a new PVC when PV name is not given in PVC yaml


jilju:
Closing this PR because it is expected to use the available PV. This test case is not valid. The bug https://issues.redhat.com/browse/DFBUGS-2596 is closed as not a bug. The behavior is same in external and internal mode.

PR #12234, opened 2025-05-22
Enables machine autorepair with machine health checks

DanielOsypenko:
Added parameter to apply to a nodepool on kubevirt cluster creation.
"Enables machine autorepair with machine health checks, default True"

### DIFF
--- a/ocs_ci/deployment/helpers/hypershift_base.py
+++ b/ocs_ci/deployment/helpers/hypershift_base.py
@@ -385,6 +385,7 @@ class HyperShiftBase:
         ocp_version=None,
         cp_availability_policy=None,
         disable_default_sources=None,
+        auto_repair=True,
     ):
         """
         Create HyperShift hosted cluster. Default parameters have minimal requirements for the cluster.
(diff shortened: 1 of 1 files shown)


DanielOsypenko:
verified: https://url.corp.redhat.com/5a483fa

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -441,9 +443,12 @@ def create_kubevirt_ocp_cluster(
             f"--pull-secret {pull_secret_path} "
             f"--image-content-sources {self.icsp_mirrors_path} "
             "--annotations 'hypershift.openshift.io/skip-release-image-validation=true' "
-            "--olm-catalog-placement Guest"
+            "--olm-catalog-placement Guest "
         )
 
+        if auto_repair:
+            create_hcp_cluster_cmd += " --auto-repair"
```
dahorak, line 450: I see just small inconsistency, that for some parts of the `create_hcp_cluster_cmd` command, there is one space before the added "part" (like here^^^)  and for other parts it is after the "part" or on both places - before and after. But if I didn't miss anything, there shouldn't be any case where the two arguments will be combined without space, so it should be ok.

dahorak (approved):

dahorak (approved):

PR #12240, opened 2025-05-22
Skip clonesnap test on external mode setup

paraggit:
Previously, I had applied the `@skipif_external_mode` marker at the class level, but it was not working. After investigation, I found that `@pytest.mark.parametrize` creates new test instances dynamically, which can override or bypass class-level markers like `@skipif_external_mode` if they are not correctly inherited. I have moved marker at function level

### DIFF
--- a/tests/functional/storageclass/test_cross_sc_clone_snap_restore.py
+++ b/tests/functional/storageclass/test_cross_sc_clone_snap_restore.py
@@ -26,7 +26,6 @@ log = logging.getLogger(__name__)
 @tier2
 @post_upgrade
 @skipif_managed_service
-@skipif_external_mode
 @skipif_hci_provider_and_client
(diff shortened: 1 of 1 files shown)


tests/functional/storageclass/test_cross_sc_clone_snap_restore.py
```diff
@@ -182,6 +182,7 @@ def test_cross_class_same_pool_clone_snap_restore(
         ],
     )
     @skipif_mcg_only
+    @skipif_external_mode
     @polarion_id("OCS-5871")
```
jilju, line 186: I think the right way to add polarion id is inside the parametrize values.

jilju (approved):

ebenahar (approved):

PR #12245, opened 2025-05-22
Ypersky pv based bs data distribution

ypersky1980:
This PR contains 3 test cases for happy path testing of [RHSTOR-6635](https://issues.redhat.com//browse/RHSTOR-6635) - Qualify "NooBaa Backing Store Data Distribution and Rebalancing". 

The dataset is 20 files, 10MB each.
In each test cases the test verifies data distribution of the dataset on 2 pv-backed backinstores, each one has 1/5/10 backed pvs. 

The data should be distributed evenly between all the pvs, while the accepted standard deviation of the average data written to the PVs is 20%.

### DIFF
--- a/tests/functional/object/mcg/test_pv_pool.py
+++ b/tests/functional/object/mcg/test_pv_pool.py
@@ -1,5 +1,7 @@
 import json
 import logging
+import time
+
 import pytest
 
 from ocs_ci.framework import config
@@ -426,3 +428,32 @@ class TestPvPool:
             amount=5,
             mcg_obj=mcg_obj,
         )
+
+
+    @tier2
+    @ignore_leftovers
+    @polarion_id("OCS-6552")
+    def test_pv_data_dist(
+        self, setup_nfs, bucket_factory, awscli_pod, test_directory_setup, mcg_obj
+    ):
+
+        bucketclass_dict = {
+            "interface": "OC",
+            "backingstore_dict": {
+                "pv": [
+                    (
+                        3,
+                        MIN_PV_BACKINGSTORE_SIZE_IN_GB,
+                        constants.NFS_SC_NAME,
+                    ),
+                    (
+                        3,
+                        MIN_PV_BACKINGSTORE_SIZE_IN_GB,
+                        constants.NFS_SC_NAME,
+                    )
+                ]
+            },
+        }
+        bucket = bucket_factory(1, "OC", bucketclass=bucketclass_dict)[0]
+        logger.info(f"********** The bucket with name {bucket.name} was created")
+        time.sleep(600)
--- a/tests/functional/object/mcg/test_pv_pool.py
+++ b/tests/functional/object/mcg/test_pv_pool.py
@@ -437,29 +437,65 @@ class TestPvPool:
             pytest.param(
                 *[
                     1,
+                    "5K",
+                    1,
+                    50000,  # dataset contains 50000 small files of 5K each, 1 pv per backingstore
+                ],
+                marks=pytest.mark.polarion_id("OCS-6852"),
+            ),
+            pytest.param(
+                *[
+                    1,
+                    "5K",
+                    1,
+                    5000,  # dataset contains 5000 small files of 5K each, 1 pv per backingstore
+                ],
+                marks=pytest.mark.polarion_id("OCS-6853"),
+            ),
+            pytest.param(
+                *[
+                    1,
                     "1M",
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_pv_pool.py
```diff
@@ -426,3 +428,115 @@ def test_pv_pool_with_nfs(
...
+        block_size,
+        block_count,
+        file_count,
+        bucket_factory,
+        awscli_pod_session,
+        mcg_obj_session,
+    ):
+        """
+        The test checks even distribution of the data written on bucket with 2 pv-backed backingstores,several pvs each.
+            1) Create bucket with 2 pv-based backingstores, several pvs on each
```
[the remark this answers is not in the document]
ypersky1980: @keesturam we are using 2 pv-based backingstores, but we would like to parametrize the number of PVs per each one of the two backingstores. Therefore we have 3 test cases with the same dataset, while the number of the pvs per each one of the two backingstores is 1) 1 2) 5 3) 10.  We would like to make sure that the data is distributed evenly in each one of the cases.

shivamdurgbuns (dismissed):

tests/functional/object/mcg/test_pv_pool.py
```diff
@@ -426,3 +428,151 @@ def test_pv_pool_with_nfs(
...
+            ),
+            pytest.param(
+                *[
+                    1,
+                    "5K",
+                    1,
+                    5000,  # dataset contains 5000 small files of 5K each, 1 pv per backingstore
+                ],
+                marks=pytest.mark.polarion_id("OCS-6853"),
+            ),
```
udaysk23, line 452: The repeated parameters are making lot of confusion. Can you please add ids to the test which will show difference between same params. Also it will make test script more readable

tests/functional/object/mcg/test_pv_pool.py
```diff
@@ -426,3 +428,115 @@ def test_pv_pool_with_nfs(
...
+                    ),
+                ]
+            },
+        }
+        bucket = bucket_factory(1, "OC", bucketclass=bucketclass_dict)[0]
+        logger.info(
+            f"The bucket with name {bucket.name} was successfully created on {pv_in_bs*2} pvs."
+        )
+
+        for i in range(file_count):
```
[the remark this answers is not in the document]
udaysk23: @ypersky1980 Please use `copy_random_individual_objects` method to upload objects if possible

tests/functional/object/mcg/test_pv_pool.py
ypersky1980: @udaysk23 which kind of IDs you'd like me to add? I did added different Polarion IDs and commented out explanations about each set of parameters .

ypersky1980: @udaysk23 I've added IDs, Done.

tests/functional/object/mcg/test_pv_pool.py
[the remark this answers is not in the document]
ypersky1980: Tried it and it seems that in this case the current way is more simple, it's also used in other test cases of this class. Prefer to leave it as is.

tests/functional/object/mcg/test_pv_pool.py
```diff
@@ -426,3 +428,160 @@ def test_pv_pool_with_nfs(
             amount=5,
             mcg_obj=mcg_obj,
         )
+
+    @tier2
+    @if_version(">4.19")
```
udaysk23: I don't think we need this marker since we are pushing all code to latest branch which is pointing to 4.19 only and in future if master changed to 4.20 or so on, the code will be present on from 4.19 and higher version and not on older versions

ypersky1980: Agree, removing the marker.

tests/functional/object/mcg/test_pv_pool.py
```diff
@@ -14,7 +16,7 @@
     fips_required,
     ignore_leftovers,
 )
-
+from ocs_ci.ocs.version import if_version
```
udaysk23: We need to remove this line as well

ypersky1980: @udaysk23 Sure, thank you. Done.

udaysk23 (approved):

Issue #12249, opened 2025-05-26
[Bug]: GCP deployment failed with incorrect storageclass name


vavuthu:
### Description
GCP deployment failed with incorrect storageclass name

### Steps to Reproduce
1. Install ODF 4.19 on GCP platform
2.
3.

### Actual Behavior
```
2025-05-23 03:36:26  18:06:26 - MainThread - ocs_ci.utility.utils - INFO  - Going to sleep for 3 seconds before next iteration
2025-05-23 03:36:29  18:06:29 - MainThread - ocs_ci.utility.utils - INFO  - Executing command: oc --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig -n openshift-storage get Pod -n openshift-storage --selector=app=rook-ceph-mon -o yaml
2025-05-23 03:36:30  18:06:29 - MainThread - ocs_ci.ocs.ocp - INFO  - status of  at column STATUS - item(s) were [], but we were waiting for all 3 of them to be Running
```

### Expected Behavior
mons should be up

### Environment
- Test Suite(s):
- Platform(s):
- Version(s):
- OS:

### Additional Context

job: https://url.corp.redhat.com/7a5a11a

PR #12250, opened 2025-05-26
set storage class name for custom storage class

vavuthu:
Fixes: #12249 


vavuthu:
verified here https://url.corp.redhat.com/face28c

Issue #12258, opened 2025-05-27
test_nsfs is failing on storageclient pod search


vkathole:
### Description

https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/51134/testReport/junit/tests.cross_functional.system_test.test_nsfs_system/TestNSFSSystem/test_nsfs/

`self = <ocs_ci.ocs.ocp.OCP object at 0x7f069773a9a0>, condition = 'Running'
resource_name = 'storageclient-737342087af10580-status-reporter-29137577-pb6px'
column = 'STATUS', selector = None, resource_count = 0, timeout = 200, sleep = 3
dont_allow_other_resources = False, error_condition = None

    def wait_for_resource(
        self,
        condition,
        resource_name="",
        column="STATUS",
        selector=None,
        resource_count=0,
        timeout=60,
        sleep=3,
        dont_allow_other_resources=False,
        error_condition=None,
    ):
        """
        Wait for a resource to reach to a desired condition
    
        Args:
            condition (str): The desired state the resource that is sampled
                from 'oc get <kind> <resource_name>' command
            resource_name (str): The name of the resource to wait
                for (e.g.my-pv1)
            column (str): The name of the column to compare with
            selector (str): The resource selector to search with.
                Example: 'app=rook-ceph-mds'
            resource_count (int): How many resources expected to be
            timeout (int): Time in seconds to wait
            sleep (int): Sampling time in seconds
            dont_allow_other_resources (bool): If True it will not allow other
                resources in different state. For example you are waiting for 2
                resources and there are currently 3 (2 in running state,
                1 in ContainerCreating) the function will continue to next
                iteration to wait for only 2 resources in running state and no
                other exists.
            error_condition (str): State of the resource that is sampled
                from 'oc get <kind> <resource_name>' command, which makes this
                method to fail immediately without waiting for a timeout. This
                is optional and makes sense only when there is a well defined
                unrecoverable state of the resource(s) which is not expected to
                be part of a workflow under test, and at the same time, the
                timeout itself is large.
    
        Returns:
            bool: True in case all resources reached desired condition,
                False otherwise
    
        """
        if condition == error_condition:
            # when this fails, this method is used in a wrong way
            raise ValueError(
                f"Condition '{condition}' we are waiting for must be different"
                f" from error condition '{error_condition}'"
                " which describes unexpected error state."
            )
        log.info(
            (
[...]

PR #12275, opened 2025-05-28
Fix system test cases

vkathole:
Fixes https://github.com/red-hat-storage/ocs-ci/issues/12258

Storageclient pod is actually the result of a cronjob scheduled every minute
So we need to skip that pod from checking everytime as pod name will continuously changes every minute

And replace noobaa_db_backup_and_recovery with noobaa_db_backup_and_recovery_locally in system test cases


vkathole:
Closing as changes included in PR 12416

PR #12291, opened 2025-05-30
[RDR][Mulitclient]Get provider index based on config value if present

jilju:
When there are more more than one provider cluster in the config (RDR test , for example), the current implementation will fetch the lower index among the provider mode cluster indexes. This PR add an additional step in the function to get the provider index based on ENV_DATA parameter `provider_cluster_name`. The client cluster config can have this value to easily identify the associated provider cluster. If the parameter is not present, current implementation will execute.


jilju:
In future, the parameter `provider_cluster_name` can be set at the start of the test run by identifying the correct provider cluster.

amr1ta (approved):
Looking good

jilju:
The code is tested for getting the correct config value used. This code change will actually be consumed in a configuration where more than one provider cluster will be used.

PR #12299, opened 2025-06-03
Fix OCP deployment on vSphere via Assisted Installer

dahorak:
Recently vSphere deployments via Assisted Installer are failing, because the installation seems to count with EFI boot mode, while the VMs are configured to legacy BIOS mode. Changing the boot mode to EFI explicitly.


dahorak:
Verification job: https://url.corp.redhat.com/19bbb24

PR #12303, opened 2025-06-03
Ensure lvm tests are skipped when skipif marker is applied and valid

clacroix12:
This change fixes an issue where lvm tests are being executed against a cluster where lvmo is not installed because the skipif marker is only considered in cases where `lvm` is found in `config.RUN`.


clacroix12:
@shyRozen pinging you directly cause it looks like you added this marker originally. I just wanted to ensure that this change matches the expected functionalty while covering the case I described above.

shyRozen:
@clacroix12 Looks like my condition has a contradiction and your fix is legit.

shyRozen (approved):

petr-balogh:
@clacroix12 please cherry-pick

PR #12318, opened 2025-06-05
Add support for multiple image changes in CSV modifications

petr-balogh:


ebenahar (dismissed):

ocs_ci/deployment/deployment.py
```diff
@@ -1331,11 +1331,16 @@ def deploy_ocs_via_operator(self, image=None):
...
+                zipped_csv_changes = zip(
+                    config.DEPLOYMENT["csv_change_from"],
+                    config.DEPLOYMENT["csv_change_to"],
                 )
+                for csv_change in zipped_csv_changes:
+                    modify_csv(
+                        csv=csv_name,
+                        replace_from=csv_change[0],
+                        replace_to=csv_change[1],
+                    )
```
dahorak:
```suggestion
                for replace_from, replace_to in zipped_csv_changes:
                    modify_csv(
                        csv=csv_name,
                        replace_from=replace_from,
                        replace_to=replace_to,
                    )
```

dahorak (dismissed):
Generally I would propose to change the logic and not store the input values in two separated lists in configuration, because it will not be directly clear what belongs to what (so I would propose to store the "zipped" version directly in the config object). Also this implementation could lead to strange issues if somebody will use it with older configuration, but that is probably not likely (previously the `DEPLOYMENT["csv_change_from"]` and `DEPLOYMENT["csv_change_to"]` values were strings, while now it expects list, but if string will be provided, it will iterate over  the characters.)
But since it is just occasionally used feature it probably doesn't make sense to spend more time on that, so I'm ok with this implementation for now.

dahorak (approved):

dahorak (approved):

petr-balogh:
Verification:
https://url.corp.redhat.com/9621166

ebenahar (approved):

PR #12401, opened 2025-06-17
odf-cli: from 4.20 use odf-cli for noobaa commands

subhamkrai:
this commits replace noobaa root command as odf-cli noobaa as a root command from odf 4.20. The noobaa cli will embedded with odf-cli.

### DIFF
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -4681,10 +4681,13 @@ def retrieve_cli_binary(cli_type="mcg"):
 
     remote_path = get_architecture_path(cli_type)
     remote_cli_basename = os.path.basename(remote_path)
-    if cli_type == "mcg":
+    if cli_type == "mcg" and semantic_version < version.VERSION_4_20:
         local_cli_path = constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH
     elif cli_type == "odf":
         local_cli_path = os.path.join(config.RUN["bin_dir"], "odf-cli")
+    elif cli_type == "mcg" and semantic_version >= version.VERSION_4_20:
+        local_cli_path = os.path.join(config.RUN["bin_dir"], "odf-cli noobaa")
+
     local_cli_dir = os.path.dirname(local_cli_path)
     live_deployment = config.DEPLOYMENT["live_deployment"]
     if live_deployment and semantic_version >= version.VERSION_4_13:
--- a/ocs_ci/ocs/resources/mcg.py
+++ b/ocs_ci/ocs/resources/mcg.py
@@ -959,20 +959,24 @@ class MCG:
 
         namespace = f"-n {namespace}" if namespace else f"-n {self.namespace}"
 
+        local_cli_path = constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH
+        if version.get_semantic_ocs_version_from_config() >= version.VERSION_4_20:
+            local_cli_path = os.path.join(config.RUN["bin_dir"], "odf-cli noobaa")
+
         # Mask sensitive data
         if self.data_to_mask:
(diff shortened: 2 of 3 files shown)


OdedViner:
As part of a new feature that @subhamkrai  is working on, mcg-cli has been integrated into odf-cli. To support this change, we now need to test mcg-cli functionality via odf-cli.
Currently, the code uses the following path to download the mcg-cli binary:
```
NOOBAA_OPERATOR_LOCAL_CLI_PATH = os.path.join(DATA_DIR, "mcg-cli")
```
To align with the new approach, we need to update all relevant tests and functions to work with the odf-cli instead of calling mcg-cli directly.

Constant definition:
https://github.com/red-hat-storage/ocs-ci/blob/487b125cce0efdf081e537ff2d76899700c2c5f6/ocs_ci/ocs/constants.py#L2216

Example usages:
https://github.com/red-hat-storage/ocs-ci/blob/487b125cce0efdf081e537ff2d76899700c2c5f6/ocs_ci/ocs/resources/mcg.py#L98
https://github.com/red-hat-storage/ocs-ci/blob/487b125cce0efdf081e537ff2d76899700c2c5f6/ocs_ci/ocs/resources/mcg.py#L968
https://github.com/red-hat-storage/ocs-ci/blob/487b125cce0efdf081e537ff2d76899700c2c5f6/tests/functional/object/mcg/test_custom_credentials_using_mcg_cli.py#L20 

CC @sagihirshfeld @petr-balogh @dahorak

ocs_ci/helpers/helpers.py
```diff
@@ -4681,10 +4681,13 @@ def retrieve_cli_binary(cli_type="mcg"):
 
     remote_path = get_architecture_path(cli_type)
     remote_cli_basename = os.path.basename(remote_path)
-    if cli_type == "mcg":
+    if cli_type == "mcg" and semantic_version < version.VERSION_4_20:
         local_cli_path = constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH
     elif cli_type == "odf":
         local_cli_path = os.path.join(config.RUN["bin_dir"], "odf-cli")
+    elif cli_type == "mcg" and semantic_version >= version.VERSION_4_20:
+        local_cli_path = os.path.join(config.RUN["bin_dir"], "odf-cli noobaa")
```
petr-balogh, line 4689: I think this will be a problem here as you have a space `odf-cli noobaa`

This is mainly used to extract the binary - and later on in line 4707 we. do rename of that locsl_cli_path , which will not work as expected here.

We just need to update above condition I think to something like:
```
elif cli_type == "odf" or cli_type == "mcg" and semantic_version >= version.VERSION_4_20:
```
And this new elif block can be deleted as we will just have `odf-cli`.

ocs_ci/ocs/resources/mcg.py
```diff
@@ -959,20 +959,24 @@ def exec_mcg_cmd(
 
         namespace = f"-n {namespace}" if namespace else f"-n {self.namespace}"
 
+        local_cli_path = constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH
+        if version.get_semantic_ocs_version_from_config() >= version.VERSION_4_20:
+            local_cli_path = os.path.join(config.RUN["bin_dir"], "odf-cli noobaa")
```
petr-balogh, line 964:
```suggestion
        local_cli_path = os.path.join(config.RUN["bin_dir"], "odf-cli noobaa")
        if version.get_semantic_ocs_version_from_config() < version.VERSION_4_20:
            local_cli_path = constants.NOOBAA_OPERATOR_LOCAL_CLI_PATH
```

As in 4.20 this will be default option, I would change this logic like this

subhamkrai:
Not required anymore, closing

PR #12413, opened 2025-06-18
Changes in tier1 tests to align with its purpose

ebenahar:
We are reviewing the current tier1 functional test suite to ensure it aligns with its core purpose - validating release readiness.
 
 
Note that many tests are proposed to move to other test suites, and especially to tier2.
It’s reasonable to increase tier2’s run time as it’s execution frequency is significantly lower than tier1’s frequency.
 
### Goals of this exercise:

- Ensure only critical tests remain (failures mean release isn’t ready)
- Reduce failure analysis overhead. tier1 is executed very frequently so moving out test cases that don’t align with its definition will reduce the time spent on analysis
- Shorten test execution time and in turn reduce release qualification turnaround time
- Improve clarity for release readiness

### DIFF
--- a/tests/functional/storageclass/test_delete_rbd_pool_attached_to_sc.py
+++ b/tests/functional/storageclass/test_delete_rbd_pool_attached_to_sc.py
@@ -8,8 +8,7 @@ from ocs_ci.framework import config
 from ocs_ci.framework.pytest_customization.marks import (
     skipif_external_mode,
     ignore_resource_not_found_error_label,
-    tier1,
-    tier2,
+    tier3,
     green_squad,
     skipif_ibm_cloud_managed,
 )
@@ -101,7 +100,7 @@ class TestDeleteRbdPool(ManageTest):
                     constants.WFFC_VOLUMEBINDINGMODE,
                     constants.STATUS_PENDING,
                 ],
-                marks=pytest.mark.polarion_id("OCS-5134"),
+                marks=[tier3, pytest.mark.polarion_id("OCS-5134")],
             ),
             pytest.param(
                 *[
@@ -110,7 +109,7 @@ class TestDeleteRbdPool(ManageTest):
                     constants.IMMEDIATE_VOLUMEBINDINGMODE,
                     constants.STATUS_BOUND,
                 ],
-                marks=[tier2, pytest.mark.polarion_id("OCS-5135")],
+                marks=[tier3, pytest.mark.polarion_id("OCS-5135")],
             ),
             pytest.param(
                 *[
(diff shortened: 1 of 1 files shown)


tests/functional/storageclass/test_delete_rbd_pool_attached_to_sc.py
```diff
@@ -9,6 +9,7 @@
     skipif_external_mode,
     ignore_resource_not_found_error_label,
     tier1,
+    tier2,
```
sidhant-agrawal: These tests are proposed for tier3 in gsheet.

tests/functional/storageclass/test_delete_rbd_pool_attached_to_sc.py
```diff
@@ -89,7 +90,6 @@ def preconditions_rbd_pool_created_associated_to_sc(
 @green_squad
 @ignore_resource_not_found_error_label
 class TestDeleteRbdPool(ManageTest):
-    @tier1
```
sidhant-agrawal, line 92: The marker can be updated here because all 4 tests are proposed for same tier

tests/functional/z_cluster/cluster_expansion/test_resize_osd.py
```diff
@@ -313,10 +314,7 @@ def test_resize_osd_for_large_diff(self, size_to_increase):
         resize_osd(self.new_storage_size)
         self.verification_steps_post_resize_osd()
 
-    @tier1
-    @tier4a
-    @tier4b
-    @tier4c
+    @tier2
```
DanielOsypenko, line 317: removing other tiers will make this verify-only test run only in tier2, meaning very rare. 

Multiple marks were added by purpose.
As of my memory this test does not do resize but works in pair wether any preceding resize test was running.
@yitzhak12 please correct me if I am wrong

ebenahar: having all of tier4a, tier4b and tier4c markers for the same test cases means these tests run every time any of those tiers is executed. Is that really needed? 
IMO, the way to ensure frequent execution is not by including the tests in multiple suites but rather to make sure the suite they are part of is executed frequently enough.
In  terms of tier2, our new regression pipeline currently has the following jobs:
* tier2 over AWS every 2 weeks
* RDR tier2 every 2 weeks
Planned to be added:
* tier2 over Azure every 8 weeks

I plan to propose having at least one more tier2 job over vSphere

tests/functional/storageclass/test_delete_rbd_pool_attached_to_sc.py
ebenahar: the approach of decorating the class can lead to mistakes (if some test functions are decorated with one marker but the class with another one).
So I preferred going with decorating each test function instead, to avoid such potential mistakes

tests/functional/storageclass/test_delete_rbd_pool_attached_to_sc.py
ebenahar: thanks, good catch. Fixed

AviadP (approved):

shyRozen (approved):

petr-balogh (approved):

hnallurv (approved):

PR #12456, opened 2025-06-23
Add negative test cases for where the StorageAutoscaler is expected not to trigger scaling

yitzhak12:
Add negative test cases for where the StorageAutoscaler is expected not to trigger scaling.

This PR introduces a new test class, **TestStorageAutoscalerNoTrigger**, which covers expected no-scaling scenarios where the StorageAutoscaler should not be triggered. These tests ensure that the autoscaler behaves correctly when scaling is intentionally suppressed or blocked by platform state or configuration.

**New Tests Added:**

- **test_create_autoscaler_and_delete_before_threshold**
  - Creates a StorageAutoscaler and deletes it before the usage threshold is reached.
  - Fills the cluster beyond the threshold after deletion.
  - Validates that the autoscaler is not triggered.

- **test_create_autoscaler_and_shutdown_osd_node**
  - Shuts down a randomly selected OSD node before filling the cluster.
  - Validates that the autoscaler does not trigger while the OSD node is offline.

- **test_storage_capacity_limit_reached**
  - Creates an autoscaler with a capacity limit that has already been reached. 
  - Validates that the autoscaler does not trigger due to the configured capacity limit.
  
**Implementation Notes:**

- Added `@pytest.mark.skip_resize_pre_conditions` to skip OSD resize logic when not applicable.
- Introduced `verify_storage_not_change` method to ensure: 
  - Storage size, OSD count, and Ceph capacity remain unchanged.
  - Data integrity checks are performed when needed.
- All tests use `fill_up_cluster` method with `is_completed=False` to control autoscaler timing.
- Add the function `wait_for_percent_used_capacity_reached` that waits until the used capacity percentage reaches or exceeds a specified threshold.
- Add the function `safe_teardown_delete_all_autoscalers` to execute in the teardown to safely delete all StorageAutoScaler resources, ensuring Prometheus has reconciled since the last cleanup.


ocs_ci/ocs/cluster.py
```diff
@@ -4008,3 +4008,42 @@ def is_lower_requirements():
         "Cluster worker machines use standard or larger instance types. No lower requirements detected."
     )
     return False
+
+
+def wait_for_percent_used_capacity_reached(
```
AviadP: as general note, cluster.py is huge module, if you can find another place for new helper function it would be better in my opinion

ocs_ci/ocs/cluster.py
```diff
@@ -4008,3 +4008,42 @@ def is_lower_requirements():
...
+            sleep=sleep,
+            func=get_percent_used_capacity,
+        ):
+            logger.info(f"Current percent used capacity = {used_capacity}%")
+            if used_capacity >= expected_used_capacity:
+                logger.info(
+                    f"The expected percent used capacity {expected_used_capacity}% reached"
+                )
+                break
+    except TimeoutExpiredError:
```
AviadP: you raise same exception within the exception. you can do raise {message} instead, or if u want to keep it that way, add `from`

ocs_ci/helpers/storage_auto_scaler.py
```diff
@@ -100,25 +101,116 @@ def wait_for_auto_scaler_status(
...
+    """
+    namespace = namespace or config.ENV_DATA["cluster_namespace"]
+
+    cleanup_cluster_time = config.RUN.get("cleanup_cluster_time")
+    if not cleanup_cluster_time:
+        logger.info("No last cleanup cluster time recorded — skipping wait.")
+        delete_all_storage_autoscalers(namespace)
+        return
+
+    time_remaining = (
```
AviadP, line 210: could it be <0 in some edge cases?

tests/functional/z_cluster/conftest.py
```diff
@@ -98,3 +98,14 @@ def workload_storageutilization_cephfs(
...
+
+@pytest.fixture
+def skip_resize_pre_conditions(request) -> bool:
+    """
+    Fixture that returns True if the test is marked with @pytest.mark.skip_resize_pre_conditions.
+
+    Returns:
+        bool: True if the marker is present, False otherwise.
+    """
+    return request.node.get_closest_marker("skip_resize_pre_conditions") is not None
```
AviadP: this fixture is not being used directly, so it should not return anything, should be something like:
`request.node.skip_resize=(request.node.get_closest_marker("skip_resize_pre_conditions") is not None)

ocs_ci/ocs/cluster.py
```diff
@@ -4008,3 +4008,42 @@ def is_lower_requirements():
...
+    `get_percent_used_capacity()` until it meets or exceeds the `expected_used_capacity`
+    or until the timeout is reached.
+
+    Args:
+        expected_used_capacity (int or float): The percentage of used capacity to wait for.
+        timeout (int): Maximum time to wait in seconds. Defaults to 1800 seconds (30 minutes).
+        sleep (int): Time to wait between checks in seconds. Defaults to 20 seconds.
+
+    Raises:
+        TimeoutExpiredError: If the expected capacity is not reached within the timeout.
```
dahorak: There should be one empty line after a list (in markdown, which is used in our docstrings).
Applies also to few other places in this PR (not mentioned all of them).

tests/functional/z_cluster/cluster_expansion/test_storage_auto_scaling.py
```diff
@@ -200,8 +245,35 @@ def prepare_data_before_auto_scaling(
             )
             self.run_io_on_pods(self.pods_for_run_io, size="2G", runtime=runtime)
 
+    def verify_post_resize_osd_steps(self):
```
dahorak, line 246: Please add there some docstring.

tests/functional/z_cluster/cluster_expansion/test_storage_auto_scaling.py
```diff
@@ -222,6 +295,7 @@ def verify_autoscaler_post_threshold_steps(
                 If not provided, the first available autoscaler will be used.
             create_additional_resources (bool): If True, creates PVCs and pods post-scaling
                 and runs I/O to verify continued cluster functionality. Defaults to True.
+            cleanup_cluster (bool): Whether to call self.cleanup_cluster() as part of verification
```
dahorak, line 314: Same as above, there should be one empty line after the list.

tests/functional/z_cluster/cluster_expansion/test_storage_auto_scaling.py
```diff
@@ -278,6 +332,54 @@ def verify_autoscaler_post_threshold_steps(
...
+
+        This method performs the following checks:
+        - Ensures the current storage size has not changed since the test started.
+        - Ensures the OSD count remains unchanged.
+        - Ensures the Ceph cluster capacity has not increased beyond the original value.
+        - Optionally, verifies data integrity on test pods using md5sum if
+          'self.create_resources_for_integrity' is True.
+
+        Raises:
+            AssertionError: If any of the current cluster values differ from the initial values.
```
dahorak, line 367: Same as above, missing empty line.

ocs_ci/helpers/storage_auto_scaler.py
yitzhak12: Yes. If everything works as expected, the cleanup is executed at the end of the test, before the teardown. For example, if we collect logs before the teardown and it takes more than 10 minutes, the `time_remaining` will be less than 0.

dahorak (dismissed):

AviadP (dismissed):

dahorak (approved):

PR #11336, opened 2025-02-10
Adding get_and_apply_idms_from_catalog function

shivamdurgbuns:
Issue: https://issues.redhat.com/browse/OCSQE-3249
Subtask: https://issues.redhat.com/browse/OCSQE-3022

PR #12460, opened 2025-06-24
fix ACM Deployment IDMS yaml - apiVersion

dahorak:
Related to: https://github.com/red-hat-storage/ocs-ci/pull/11336


dahorak:
Verified manually on the failed cluster.

PR #12517, opened 2025-06-30
infra-availability-policy added HighlyAvaialble as a default

DanielOsypenko:
when this setting is not used the default parameter is used silently and PDB is to strict to run drain operation on nodes. It is affecting ODF upgrade, OCP upgrade and node operation tests.


dahorak (dismissed):

dahorak (approved):

DanielOsypenko:
verified - https://url.corp.redhat.com/97d5b9e

PR #12527, opened 2025-07-01
Adds support of md_blow tool

udaysk23:
Adds script for md_blow tool which is required for dumping IO directly into noobaa DB

### DIFF
--- a/ocs_ci/ocs/md_blow.py
+++ b/ocs_ci/ocs/md_blow.py
@@ -1,0 +1,159 @@
+import logging
+import threading
+from time import sleep
+
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.resources.pod import get_noobaa_core_pod, get_noobaa_db_pod
+from ocs_ci.ocs.resources.mcg import MCG
+from ocs_ci.utility.utils import run_cmd
+from ocs_ci.helpers import helpers
+from ocs_ci.framework import config
+
+
+logger = logging.getLogger(__name__)
+
+
+class MdBlow(object):
+    """
+    Upload s3 objects directly into database using MD_Blow script present in noobaa-core pod.
+    """
+
+    def __init__(self):
+        """
+        Assign default values to its parameter and patch required lines in core pod
+        """
+        self.db_percentage = None
+        self.obj_count = 1000
+        self.concurrency = 50
+        self.chunks = 1
+        self.chunk_size = 200
+        mcg_obj = MCG()
+        self.creds = mcg_obj.get_noobaa_admin_credentials_from_secret()
+        self.email = self.creds["email"]
+        self.password = self.creds["password"]
+        self.noobaa_core_pod = get_noobaa_core_pod()
+        self.noobaa_db_pod = get_noobaa_db_pod()
+        namespace = config.ENV_DATA["cluster_namespace"]
+        storage_cluster = constants.DEFAULT_STORAGE_CLUSTER
+
+        ptch = '{{"spec": {{"resources": {{"noobaa-core": {{"limits": {{"cpu": "6","memory": "4Gi"}}}}}}}}}}'
+        ptch_cmd = (
+            f"oc patch storagecluster {storage_cluster} "
+            f"-n {namespace}  --type merge --patch '{ptch}'"
+        )
+        run_cmd(ptch_cmd)
+        logger.info("Wait for noobaa-core pod move to Running state")
+        helpers.wait_for_resource_state(
+            self.noobaa_core_pod, state=constants.STATUS_RUNNING, timeout=300
+        )
+
+    def monitor_db_usage(self, threshold_pct):
+        """
+        Monitor database usage and raise a flag to stop dumping once the threshold is reached.
+
+        Args:
+            threshold_pct (int): Usage percentage threshold at which to stop dumping.
+        """
+        current_db_percentage = self.noobaa_db_pod.exec_cmd_on_pod(
+            "df -h | grep postgresql | awk '{print $5}'",
+            shell=True,
+        )
+        current_db_usage = int(current_db_percentage.strip().replace("%", ""))
+        logger.info(f"current DB usage: {current_db_usage}")
+        count = 0
+        while not self.stop_dumping.is_set():
+            if current_db_usage >= threshold_pct:
+                self.stop_dumping.set()
+                logger.info(f"DB is filled with {threshold_pct}")
+            prev_db_percentage = self.noobaa_db_pod.exec_cmd_on_pod(
+                "df -h | grep postgresql | awk '{print $5}'",
+                shell=True,
+            )
+            prev_db_usage = int(prev_db_percentage.strip().replace("%", ""))
+            if prev_db_usage == current_db_percentage:
+                count += 1
+            else:
+                count = 0
+            if count == 100:
+                logger.error(
+                    "DB Percentage is stuck from last 500 seconds to the same value"
+                )
+                self.stop_dumping.set()
+            current_db_percentage = self.noobaa_db_pod.exec_cmd_on_pod(
+                "df -h | grep postgresql | awk '{print $5}'",
+                shell=True,
+            )
+            current_db_usage = int(current_db_percentage.strip().replace("%", ""))
+            sleep(5)
+        logger.info("Exiting from monitor task")
+
+    def upload_obj_using_md_blow(
+        self,
+        bucket_name="first.bucket",
+        threshold_pct=None,
+        obj_count=1000,
+        concurrency=50,
+        chunks=200,
+        chunk_size=1,
+    ):
+        """
+        Upload objects directly into DB using MD Blow script present in core pod
+
+        Args:
+            bucket_name (str): Bucket name where objects need to be uploaded
+            db_percentage (int): Total percent of DB to be filled
+            obj_count (int): Number of objects to upload in given bucket
+            concurrency (int): Number of threads to be used while uploading data
+            chunks (int): Chunk numbers in each object
+            chunk_size (int): Chunk size in each object
+        """
+        base_cmd = (
+            "node /root/node_modules/noobaa-core/src/tools/md_blow.js --system=noobaa "
+            f"--email={self.email} "
+            f"--password={self.password} "
+            f"--bucket={bucket_name} "
+        )
+        if threshold_pct is None:
+            cmd = (
+                f"--count={obj_count} "
+                f"--concur={concurrency} "
+                f"--chunks={chunks} "
+                f"--chunk_size={chunk_size}"
+            )
+            self.noobaa_core_pod.exec_cmd_on_pod(base_cmd + cmd)
+            logger.info("Workload executed successfully")
+        else:
+            assert threshold_pct <= 100, f"Invalid value. Given {threshold_pct}"
+            current_db_percentage = self.noobaa_db_pod.exec_cmd_on_pod(
+                "df -h | grep postgresql | awk '{print $5}'",
+                shell=True,
+            )
+            current_db_usage = int(current_db_percentage.strip().replace("%", ""))
+            logger.info(f"current DB usage: {current_db_usage}")
+            if current_db_usage >= threshold_pct:
+                logger.info(
+                    f"DB is already filled with {current_db_percentage}. Exiting from script without dumping IOs"
+                )
+            else:
+                self.stop_dumping = threading.Event()
+                # Launch threads to monitor DB usage
+                t1 = threading.Thread(
+                    target=self.monitor_db_usage,
+                    args=(threshold_pct,),
+                    name="MonitorThread",
+                )
+                t1.start()
+                # Adding sleep to validate invalid percentage usage
+                sleep(10)
+                cmd = (
+                    f"--count={self.obj_count} "
+                    f"--concur={self.concurrency} "
+                    f"--chunks={self.chunks} "
+                    f"--chunk_size={self.chunk_size}"
+                )
+                logger.info("Initiating IO dump directly into DB")
+                while not self.stop_dumping.is_set():
+                    self.noobaa_core_pod.exec_cmd_on_pod(base_cmd + cmd)
+                    sleep(5)
+                t1.join()
+                logger.info(f"Stopping the IO... DB is filled with {threshold_pct}")


udaysk23:
This code is required for MCG DB tests cases where the requirement is to fill DB with certain percentage.
Currently those tests are not yet automated and due to this we can't trigger jenkins run for PR verification.
I have written a sample test script and triggered locally for verifying this and based on result, I am attaching `verified` label to this PR

```========================================================================= PASSES =========================================================================
___________________________________________________________ TestMd_blow.test_md_blow_benchmark ___________________________________________________________
18:01:08 - MainThread - ocs_ci.framework.pytest_customization.reports - INFO  - duration reported by tests/functional/object/mcg/test_md_blow.py::TestMd_blow::test_md_blow_benchmark immediately after test execution: 7134.01
================================================================ short test summary info =================================================================
PASSED tests/functional/object/mcg/test_md_blow.py::TestMd_blow::test_md_blow_benchmark
======================================================= 1 passed, 3 warnings in 7537.22s (2:05:37) =======================================================```

**Review by shivamdurgbuns — APPROVED** [review]

**Review by ypersky1980 — APPROVED** [review]
LGTM

**Review by ypersky1980 — DISMISSED** [review]
LGTM

File: ocs_ci/ocs/md_blow.py [hunk 1]
```diff
@@ -0,0 +1,149 @@
...
+        self.password = self.creds["password"]
+        self.noobaa_core_pod = get_noobaa_core_pod()
+        self.noobaa_db_pod = get_noobaa_db_pod()
+
+    def patch_lines_in_core_pod(self):
+        """
+        patch required lines in script to avoid failures
+        seed below line in js file at line no 32
+        await require('../server/system_services/master_key_manager.js').get_instance().load_root_keys_from_mount();
+        """
```
keesturam: why can't we have the updated script made available somewhere and access it instead of modifying the script everytime?

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,149 @@
...
+            shell=True,
+        )
+        current_db_usage = int(current_db_percentage.strip().replace("%", ""))
+        logger.info(current_db_usage)
+        if current_db_usage >= db_percentage:
+            logger.warn(
+                f"Provided fill percentage is invalid. Given {db_percentage}... Current usage of DB {current_db_usage}"
+            )
+            self.stop_dumping.set()
+        if db_percentage >= 100:
```
keesturam: You could simply assert `assert db_percentage < 100` in the beginning if the value is 100 or more?

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,149 @@
...
+        while not self.stop_dumping.is_set():
+            if current_db_usage >= db_percentage:
+                self.stop_dumping.set()
+                logger.info(f"DB is filled with {db_percentage}")
+            current_db_percentage = self.noobaa_db_pod.exec_cmd_on_pod(
+                "df -h | grep postgresql | awk '{print $5}'",
+                shell=True,
+            )
+            current_db_usage = int(current_db_percentage.strip().replace("%", ""))
+            sleep(5)
```
keesturam, line 87: why do we need this sleep? can TimeoutSampler be used instead of a static value set here.

ocs_ci/ocs/md_blow.py
udaysk23: This script is present inside the core pod

ocs_ci/ocs/md_blow.py
udaysk23: @keesturam The reason for using if condition is, Script needs to set threading flag and assert statement wont allow that kind of behaviour

ocs_ci/ocs/md_blow.py
udaysk23: We cant use TimeoutSampler here since this sleep is needed to add some kind of delay while checking and setting threading flag

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,149 @@
...
+
+from ocs_ci.ocs.resources.pod import get_noobaa_core_pod, get_noobaa_db_pod
+from ocs_ci.ocs.resources.mcg import MCG
+
+logger = logging.getLogger(__name__)
+
+
+class MdBlow(object):
+    """
+    Upload objects directly into DB using MD Blow script present in core pod
```
keesturam: could you pls add more details on this tool. if this is an external tool, maybe give a link to the tool?

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,149 @@
...
+        self.chunks = 1
+        self.chunk_size = 200
+        mcg_obj = MCG()
+        self.creds = mcg_obj.get_noobaa_admin_credentials_from_secret()
+        self.email = self.creds["email"]
+        self.password = self.creds["password"]
+        self.noobaa_core_pod = get_noobaa_core_pod()
+        self.noobaa_db_pod = get_noobaa_db_pod()
+
+    def patch_lines_in_core_pod(self):
```
keesturam: why is this needed?

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,149 @@
...
+        await require('../server/system_services/master_key_manager.js').get_instance().load_root_keys_from_mount();
+        """
+        f_path = "/root/node_modules/noobaa-core/src/tools/md_blow.js"
+        insert_at = 31
+        line = (
+            "        await require("
+            '"../server/system_services/master_key_manager.js"'
+            ").get_instance().load_root_keys_from_mount();"
+        )
+        cmd = f"sed -i '{insert_at}a\\{line}' {f_path}"
```
keesturam: what happens when any of these operation fails?

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,149 @@
...
+            )
+            self.noobaa_core_pod.exec_cmd_on_pod(base_cmd + cmd)
+            logger.info("Workload executed successfully")
+
+    def cleanup(self):
+        """
+        Removes patched line from md_blow.js file
+        """
+        f_path = "/root/node_modules/noobaa-core/src/tools/md_blow.js"
+        delete_from = 32
```
keesturam: what happens when this number changes for eg, in next release?

ocs_ci/ocs/md_blow.py
udaysk23: @keesturam I am checking with noobaa-core dev to add above mentioned line in source code itself.
Will provide update here once I hear anything from DEV on this

udaysk23: Dev is planning to push fix for this issue. Once fix is available in core pod, I will remove this part from script

udaysk23: Below PR will fix this issue https://github.com/noobaa/noobaa-core/pull/9170

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,155 @@
...
+        self.concurrency = 50
+        self.chunks = 1
+        self.chunk_size = 200
+        mcg_obj = MCG()
+        self.creds = mcg_obj.get_noobaa_admin_credentials_from_secret()
+        self.email = self.creds["email"]
+        self.password = self.creds["password"]
+        self.noobaa_core_pod = get_noobaa_core_pod()
+        self.noobaa_db_pod = get_noobaa_db_pod()
+        namespace = constants.OPENSHIFT_STORAGE_NAMESPACE
```
sagihirshfeld: Please add support for different default namespaces:
```suggestion
        namespace = config.ENV_DATA["cluster_namespace"]
```

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,155 @@
...
+        ptch = '{{"spec": {{"resources": {{"noobaa-core": {{"limits": {{"cpu": "6","memory": "4Gi"}}}}}}}}}}'
+        ptch_cmd = (
+            f"oc patch storagecluster {storage_cluster} "
+            f"-n {namespace}  --type merge --patch '{ptch}'"
+        )
+        run_cmd(ptch_cmd)
+        logger.info("Wait for noobaa-core pod move to Running state")
+        helpers.wait_for_resource_state(
+            self.noobaa_core_pod, state=constants.STATUS_RUNNING, timeout=300
+        )
```
sagihirshfeld, line 48: We should undo this optimization once we're done running the md_blow script. 

One safe approach to do so would be to serve the instance of this class via a fixture, and then add the necessary cleanup at its teardown.

Another could be to wrap `upload_obj_using_md_blow` in a method that would call it inside a try-finally block like this:
```
def upload_obj_using_md_blow(...):
    # Fetch default cpu and mem values to local variables
    ...
    ...
    # Patch high cpu and mem values to optimization
   ...
   ...
    try:
        self._actual_upload_obj_using_md_blow(...):
    finally:
        # Patch the original cpg and mem values
```

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def monitor_db_usage(self, db_percentage):
+        """
+        Get current DB percentage
+
+        Args:
+            db_percentage (int): Percentage of DB to be monitored
+        """
```
sagihirshfeld: For clarity:
```suggestion
def monitor_db_usage(self, threshold_pct):
    """
    Monitor database usage and raise a flag to stop dumping once the threshold is reached.

    Args:
        threshold_pct (int): Usage percentage threshold at which to stop dumping.
    """
```

ocs_ci/ocs/md_blow.py
sagihirshfeld: How about we move these validations to `upload_obj_using_md_blow`?

If `db_percentage` is lower than the current, we return since the target has been reached.
If it's invalid (not a valid percentage number) we can raise an exception before we start any threads.

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,155 @@
...
+        if db_percentage is not None:
+            self.stop_dumping = threading.Event()
+            # Launch threads to monitor DB usage
+            t1 = threading.Thread(
+                target=self.monitor_db_usage,
+                args=(db_percentage,),
+                name="MonitorThread",
+            )
+            t1.start()
+            # Addign sleep to validate invalid percentage usage
```
sagihirshfeld:
```suggestion
            # Adding sleep to validate invalid percentage usage
```

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,155 @@
...
+            chunks (int): Chunk numbers in each object
+            chunk_size (int): Chunk size in each object
+        """
+        base_cmd = (
+            "node /root/node_modules/noobaa-core/src/tools/md_blow.js --system=noobaa "
+            f"--email={self.email} "
+            f"--password={self.password} "
+            f"--bucket={bucket_name} "
+        )
+        if db_percentage is not None:
```
sagihirshfeld: We can avoid the giant indentation block by returning/failing early if the argument is invalid:
```suggestion
        if db_percentage is None:
            raise ValueError("db_percentage must be provided and in the range [0, 100].") 
```

udaysk23: @sagihirshfeld We have else block which will run if user wants to dump IOs without considering `threshold_pct`

ocs_ci/ocs/md_blow.py
udaysk23: @sagihirshfeld I have one more PR ready considering this tool which contains design of fixture in conftest and  cleanup method implementation in this script.

ocs_ci/ocs/md_blow.py
udaysk23: fixed

ocs_ci/ocs/md_blow.py
udaysk23: fixed

ocs_ci/ocs/md_blow.py
udaysk23: fixed

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Args:
+            threshold_pct (int): Usage percentage threshold at which to stop dumping.
+        """
+        current_db_percentage = self.noobaa_db_pod.exec_cmd_on_pod(
+            "df -h | grep postgresql | awk '{print $5}'",
+            shell=True,
+        )
+        current_db_usage = int(current_db_percentage.strip().replace("%", ""))
+        logger.info(current_db_usage)
```
shivamdurgbuns: little log info here would be helpful

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,161 @@
...
+            f"--password={self.password} "
+            f"--bucket={bucket_name} "
+        )
+        if threshold_pct is not None:
+            current_db_percentage = self.noobaa_db_pod.exec_cmd_on_pod(
+                "df -h | grep postgresql | awk '{print $5}'",
+                shell=True,
+            )
+            current_db_usage = int(current_db_percentage.strip().replace("%", ""))
+            logger.info(current_db_usage)
```
shivamdurgbuns: little log info here would be helpful

ocs_ci/ocs/md_blow.py
sagihirshfeld: Oh I missed the `else` part at the bottom... in this case, we can swap so if `db_percentage` is None we run the else logic and then return:
```
# Run md_blow with the given parameters without monitoring the db's capacity
if db_percentage is None:
    cmd = (
        f"--count={obj_count} "
        f"--concur={concurrency} "
        f"--chunks={chunks} "
        f"--chunk_size={chunk_size}"
    )
    self.noobaa_core_pod.exec_cmd_on_pod(base_cmd + cmd)
    logger.info("Workload executed successfully")
    return

# The rest of the original method, unindented

# Run md_blow until the used capacity reaches db_percentage
current_db_percentage = self.noobaa_db_pod.exec_cmd_on_pod(
    "df -h | grep postgresql | awk '{print $5}'",
    shell=True,
)
...
....
```

ocs_ci/ocs/md_blow.py
```diff
@@ -0,0 +1,161 @@
...
+            if current_db_usage >= threshold_pct:
+                logger.error(
+                    f"Invalid threshold percentage. Given {threshold_pct}... Current usage of DB {current_db_usage}"
+                )
+            elif threshold_pct >= 100:
+                logger.error(f"Invalid value. Given {threshold_pct}")
+            else:
```
sagihirshfeld: We can avoid the indentation of lines 130-152 by exiting the method in case these validation fail. Raising a ValueError exception with the existing messages would fit in both cases.

ocs_ci/ocs/md_blow.py
udaysk23: Fixed

ocs_ci/ocs/md_blow.py
udaysk23: fixed

ocs_ci/ocs/md_blow.py
udaysk23: fixed

ocs_ci/ocs/md_blow.py
udaysk23: fixed

PR #12551, opened 2025-07-02
Remove early testing key for Graviton deployment

petr-balogh:


petr-balogh:
Verification job:
https://url.corp.redhat.com/4086baf

PR #12556, opened 2025-07-03
disable consumer check to prevent build failures

DanielOsypenko:
Added catch for Assertion errror for storageconsumer and  config map verification on fresh deployments.
We expected to have these changes in secret names as a part of the fix of https://issues.redhat.com/browse/DFBUGS-2840.


DanielOsypenko:
created issue to return back verification when code will be adjusted: #12557

Issue #12361, opened 2025-06-11
In the node replacement verification steps add a retry to the "ceph osd status" check


yitzhak12:
In the node replacement verification steps, add a retry to the "ceph osd status" check. 
See, for example, the failure here: https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/852/31785/1557903/1557936/log.

ebenahar:
@yitzhak12 to work on the fix this week

PR #12561, opened 2025-07-03
Wait more time for verifying the node names in the Ceph OSD status

yitzhak12:
fix https://github.com/red-hat-storage/ocs-ci/issues/12361

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -1474,38 +1474,16 @@ def node_replacement_verification_steps_ceph_side(
     if not pod.check_pods_after_node_replacement():
         return False
 
-    node_names_from_ceph_osd_status = get_node_names_from_osd_status()
-    osd_node_names = get_osd_running_nodes()
-    log.info(f"osd node names: {osd_node_names}")
-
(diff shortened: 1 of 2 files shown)


ebenahar (approved):

AviadP (approved):

ocs_ci/ocs/node.py
```diff
@@ -3375,3 +3353,52 @@ def apply_node_affinity_for_noobaa_pod():
...
+        return False
+
+    if (
+        old_node_name in osd_node_names
+        and config.ENV_DATA["platform"].lower() != constants.VSPHERE_PLATFORM
+    ):
+        log.warning("Old node name found in OSD running node names")
+        return False
+
+    return True
```
am-agrawa, line 3404: Can we use this bool into assert?

yitzhak12: I think it would be better in this case to return a boolean value.
The function returns a boolean, so the caller has control over how to handle success/failure (e.g., log, retry, raise).

am-agrawa (approved):

PR #12588, opened 2025-07-07
Fix ceph_rocksdb_* metric names

fbalak:
As clarified in https://issues.redhat.com/browse/DFBUGS-2931

### DIFF
--- a/ocs_ci/ocs/metrics.py
+++ b/ocs_ci/ocs/metrics.py
@@ -153,12 +153,13 @@ ceph_metrics_healthy = (
     "ceph_bluefs_slow_used_bytes",
     "ceph_osd_op_r_latency_sum",
     "ceph_bluestore_kv_flush_lat_count",
-    # "ceph_rocksdb_compact_range" replaced with compact_running, compact_lasted, compact_completed
+    # "ceph_rocksdb_compact_range" replaced with:
+    # ceph_rocksdb_compact_running, ceph_rocksdb_compact_lasted, ceph_rocksdb_compact_completed
     # more info in https://issues.redhat.com/browse/DFBUGS-2760,
     # https://tracker.ceph.com/issues/67040 and https://github.com/ceph/ceph/pull/57107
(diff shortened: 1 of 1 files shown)


fbalak:
Verified: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/53187

petr-balogh (approved):

ocs_ci/ocs/metrics.py
```diff
@@ -153,12 +153,13 @@
...
     "ceph_bluestore_kv_flush_lat_count",
-    # "ceph_rocksdb_compact_range" replaced with compact_running, compact_lasted, compact_completed
+    # "ceph_rocksdb_compact_range" replaced with:
+    # ceph_rocksdb_compact_running, ceph_rocksdb_compact_lasted, ceph_rocksdb_compact_completed
     # more info in https://issues.redhat.com/browse/DFBUGS-2760,
     # https://tracker.ceph.com/issues/67040 and https://github.com/ceph/ceph/pull/57107
-    "compact_running",
-    "compact_lasted",
-    "compact_completed",
+    "ceph_rocksdb_compact_running",
```
petr-balogh, line 160: Are those metrics relevant to 4.19 only? As we still need to keep in mind to support in ocs-ci  ODF n-1 version which means also 4.18

fbalak: It is only relevant to 4.19 as per description of https://issues.redhat.com/browse/DFBUGS-2760

petr-balogh: should we then make some condition and append those new metrics only when OCS version is >= 4.19 ?

DanielOsypenko: @petr-balogh is this relevant only for pre-upgrade tests? I am not sure we are running this on pre-upgrade, but I will check if relevant

Issue #11479, opened 2025-02-24
test_update_strategy_config_change[csi-cephfsplugin-2-2] is failing on IBM Power


Arpanchak:
Below test case is failing on ODF 4.18 tier1 on IBM Power -
`tests/functional/upgrade/test_configuration.py::test_update_strategy_config_change[csi-cephfsplugin-2-2]`

Error being -
```
daemonset = 'csi-cephfsplugin', value_to_set = 2, expected_value = 2
rook_operator_configmap_cleanup = None

    @pytest.mark.parametrize(
        argnames=["daemonset", "value_to_set", "expected_value"],
        argvalues=[
            pytest.param(
                "csi-rbdplugin", 2, 2, marks=[tier1, pytest.mark.polarion_id("OCS-6276")]
            ),
            pytest.param(
                "csi-cephfsplugin", 2, 2, marks=[tier1, pytest.mark.polarion_id("OCS-6277")]
            ),
        ],
    )
    @brown_squad
    def test_update_strategy_config_change(
        daemonset, value_to_set, expected_value, rook_operator_configmap_cleanup
    ):
        """
        Test that tested value added to configmap rook-ceph-operator-config is
        reflected in respective daemonset.
        """
        if daemonset == "csi-rbdplugin":
            parameter_name = "CSI_RBD_PLUGIN_UPDATE_STRATEGY_MAX_UNAVAILABLE"
        elif daemonset == "csi-cephfsplugin":
            parameter_name = "CSI_CEPHFS_PLUGIN_UPDATE_STRATEGY_MAX_UNAVAILABLE"
    
        config_map_patch = f'\'{{"data": {{"{parameter_name}": "{value_to_set}"}}}}\''
        exec_cmd(
            f"oc patch configmap -n {config.ENV_DATA['cluster_namespace']} "
            f"{constants.ROOK_OPERATOR_CONFIGMAP} -p {config_map_patch}"
        )
        ds_obj = DaemonSet(
            resource_name=daemonset, namespace=config.ENV_DATA["cluster_namespace"]
        )
        results = ds_obj.get_update_strategy()
>       assert str(expected_value) == str(results["rollingUpdate"]["maxUnavailable"])
E       AssertionError: assert '2' == '1'
E         - 1
E         + 2

tests/functional/upgrade/test_configuration.py:140: AssertionError
```

Logs : 

[...]

PR #12616, opened 2025-07-09
Issue#11479 is uncertain issue while execution because slowness in update

suchita-g:
Just wait for few seconds before taking information related to update strategy. 
Fixes https://github.com/red-hat-storage/ocs-ci/issues/11479


suchita-g:
I verified the fix locally. For 4.19 it need another fix to pass this test which I pushed in another PR.

Issue #12132, opened 2025-05-12
test_namespace_bucket_creation_crd - various set of parameters -NSS creation does not reach a healthy state within 180 seconds - frequent problem


ypersky1980:
There is a frequent failure in NSS creation that does not reach a healthy state within 180 seconds. 
NSS creation timeout should be increased. 

Backtrace: 

self = <ocs_ci.ocs.resources.namespacestore.NamespaceStore object at 0x7facc831d100>
timeout = 180, interval = 5

def verify_health(self, timeout=180, interval=5):
    """
    Health verification function that tries to verify
    a namespacestores's health until a given time limit is reached

    Args:
        timeout (int): Timeout for the check, in seconds
        interval (int): Interval to wait between checks, in seconds

    Returns:
        (bool): True if the bucket is healthy, False otherwise

    """
    log.info(f"Waiting for {self.name} to be healthy")
    try:

      for health_check in TimeoutSampler(
            timeout, interval, getattr(self, f"{self.method}_verify_health")
        ):

ocs_ci/ocs/resources/namespacestore.py:174: 

self = <ocs_ci.utility.utils.TimeoutSampler object at 0x7facc831d220>

def __iter__(self):
    if self.start_time is None:
        self.start_time = time.time()
    while True:
        self.last_sample_time = time.time()
        if self.timeout <= (self.last_sample_time - self.start_time):

          raise self.timeout_exc_cls(*self.timeout_exc_args)
E               ocs_ci.ocs.exceptions.TimeoutExpiredError: Timed out after 180s running oc_verify_health()

ocs_ci/utility/utils.py:1493: TimeoutExpiredError

During handling of the above exception, another exception occurred:

self = <mcg.test_namespace_crd.TestNamespace object at 0x7facacdaf4c0>
bucket_factory = <function bucket_factory_fixture.._create_buckets at 0x7fac9c1c4940>
bucketclass_dict = {'interface': 'OC', 'namespace_policy_dict': {'namespacestore_dict': {'aws': [(1, None)]}, 'type': 'Single'}}

@pytest.mark.parametrize(
    argnames=["bucketclass_dict"],
    argvalues=[
        pytest.param(
            {
                "interface": "OC",
                "namespace_policy_dict": {
                    "type": "Single",
                    "namespacestore_dict": {"aws": [(1, None)]},
                },
            },
            marks=[tier1, pytest.mark.polarion_id("OCS-2256")],
        ),
        pytest.param(
            {
[...]

Issue #12112, opened 2025-05-08
test_namespace_bucket_creation_crd - various set of parameters - namespace deletion tims should be increased


ypersky1980:
There is a failure in 

def nss_cleanup():
        for nss in created_nss:

def nss_cleanup():
        for nss in created_nss:

      nss.delete()

ocs_ci/ocs/resources/namespacestore.py:476: 

self = <ocs_ci.ocs.resources.namespacestore.NamespaceStore object at 0x7f002fe735b0>
retry = True

def delete(self, retry=True):
    """
    Deletes the current namespacestore by using OC/CLI commands

    Args:
        retry (bool): Whether to retry the deletion if it fails

    """
    log.info(f"Cleaning up namespacestore {self.name}")

    def _oc_deletion_flow():
        try:
            OCP(
                kind="namespacestore",
                namespace=config.ENV_DATA["cluster_namespace"],
            ).delete(resource_name=self.name)
            return True
        except CommandFailed as e:
            if "not found" in str(e).lower():
                log.warning(f"Namespacestore {self.name} was already deleted.")
                return True
            elif all(
                err in e.args[0]
                for err in ["cannot complete because pool", "in", "state"]
            ):
                if retry:
                    log.warning(
                        f"Deletion of {self.name} failed due to its state; Retrying"
                    )
                    return False
                else:
                    raise
            else:
                raise

    def _cli_deletion_flow():
        try:
            self.mcg_obj.exec_mcg_cmd(f"namespacestore delete {self.name}")
            return True
        except CommandFailed as e:
            if "being used by one or more buckets" in str(e).lower():
                log.warning(
                    f"Deletion of {self.name} failed because it's being used by a bucket. "
                    "Retrying..."
                )
            else:
                log.warning(f"Deletion of self.name failed. Error:\n{str(e)}")
            return False

    cmdMap = {
        "oc": _oc_deletion_flow,
        "cli": _cli_deletion_flow,
    }
    if retry:
        sample = TimeoutSampler(
            timeout=120,
            sleep=20,
            func=cmdMap[self.method],
        )
        if not sample.wait_for_func_status(result=True):
            err_msg = f"Failed to delete {self.name}"
            log.error(err_msg)

          raise TimeoutExpiredError(err_msg)
E               ocs_ci.ocs.exceptions.TimeoutExpiredError: Timed Out: Failed to delete ibmcos-ns-store-c2bff41c20ca41deb4f099c1

[...]

PR #12644, opened 2025-07-13
Inscreasing timeouts for nss creation and nss deletion

ypersky1980:
This PR resolves issue  https://github.com/red-hat-storage/ocs-ci/issues/12132  and https://github.com/red-hat-storage/ocs-ci/issues/12112 . 

I've added a timeout parameter to namespacestore factory for waiting for NSS to become healthy (nss_creation_timeout) and added it to all the relevant test cases and parametrizations. 

I've also added nss_deletion_timeout and added it to all the relevant test cases and parametrizations.


udaysk23 (dismissed):

ocs_ci/ocs/resources/bucketclass.py
```diff
@@ -128,7 +123,23 @@ def _create_bucket_class(bucket_class_dict):
...
-                namespacestores = namespace_store_factory(interface, nss_dict)
+                if "nss_creation_deletion_timeout" in bucket_class_dict:
+                    nss_creation_timeout = bucket_class_dict[
+                        "nss_creation_deletion_timeout"
+                    ]
+                    nss_deletion_timeout = bucket_class_dict[
+                        "nss_creation_deletion_timeout"
+                    ]
+                else:
+                    nss_creation_timeout = 180
```
keesturam, line 134: could you please see if timeoutsampler could be used instead of static timeout?

ypersky1980: @keesturam 
The nss_creation_timeout is passed to namespace_store_factory, to _create_nss method and eventually to verify_health method . And in this method there is a TimeoutSampler, so nss_creation_timeout defined here is the upper limit in the worst case and not the real time the creation will take.
The same applies for nss_deletion_timeout. So we are already using a timeout sampler indeed.

tests/functional/object/mcg/test_namespace_crd.py
```diff
@@ -152,6 +156,7 @@ def deprecated_test_namespace_store_creation_crd(
                         "type": "Single",
                         "namespacestore_dict": {"aws": [(1, None)]},
                     },
+                    "nss_creation_deletion_timeout": 240,
```
udaysk23, line 159: Please change timeout in TimeoutSampler to 240 to avoid adding timeout for each case

PR #12650, opened 2025-07-14
Reduce noise from exec_cmd

AviadP:
- [ ] Removed empty output logging
- [ ] Added flexible stdout logging control (log_stdout parameter)
- [ ] Smart stderr level logic (ERROR for failures, configurable for success)
- [ ] Added Deprecation warning for `run_cmd`

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -500,6 +500,9 @@ def run_cmd(
         output_file=output_file,
         **kwargs,
     )
+    log.warning(
+        "You are using the deprecated form of exec_cmd. Please use exec_cmd instead."
+    )
     return mask_secrets(completed_process.stdout.decode(), secrets)
 
 
@@ -608,6 +611,8 @@ def exec_cmd(
     cluster_config=None,
     lock_timeout=7200,
     output_file=None,
+    log_stdout=True,
+    log_stderr_level=logging.DEBUG,
     **kwargs,
 ):
     """
@@ -632,6 +637,8 @@ def exec_cmd(
                 will be non-null
         lock_timeout (int): maximum timeout to wait for lock to prevent deadlocks (default 2 hours)
         output_file (str): path where to write output of stderr from command - apply only when silent mode is True
+        log_stdout (bool): If True will log stdout, default true
+        log_stderr_level (int): Log level for stderr on successful commands, default DEBUG
 
     Raises:
         CommandFailed: In case the command execution fails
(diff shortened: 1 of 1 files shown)


ocs_ci/utility/utils.py
```diff
@@ -680,15 +687,15 @@ def exec_cmd(
                 # If oc cmdline has plugin name then we need to push the
                 # --kubeconfig to next index
                 kube_index = 2
-                log.info(f"Found oc plugin {subcmd}")
+                log.info("Found oc plugin %s", subcmd)
```
ebenahar, line 1017: this change is unnecessary

ocs_ci/utility/utils.py
```diff
@@ -680,15 +687,15 @@ def exec_cmd(
...
         cmd = list_insert_at_position(cmd, kube_index, ["--kubeconfig"])
         cmd = list_insert_at_position(cmd, kube_index + 1, [kubeconfig_path])
     try:
         if kwargs.get("shell"):
             masked_cmd = mask_secrets(cmd, secrets)
         else:
             masked_cmd = shlex.join(mask_secrets(cmd, secrets))
-        log.info(f"Executing command: {masked_cmd}")
+        log.info("Executing command: %s", masked_cmd)
```
ebenahar: this change is unnecessary

ocs_ci/utility/utils.py
```diff
@@ -700,38 +707,39 @@ def exec_cmd(
...
-        log.debug("Command stdout is empty")
+    if log_stdout and len(completed_process.stdout) > 0:
+        log.debug("Command stdout: %s", masked_stdout.rstrip())
 
     masked_stderr = mask_secrets(completed_process.stderr.decode(), secrets)
     if len(completed_process.stderr) > 0:
         if not silent:
-            log.warning(f"Command stderr: {masked_stderr}")
+            # Use ERROR level for failed commands, configurable level for successful ones
+            level = logging.ERROR if completed_process.returncode else log_stderr_level
```
ebenahar, line 1106: Why ERROR if `if completed_process.returncode` is True?

ocs_ci/utility/utils.py
AviadP: it saves a bit of resources, the string formatting only happens if the message will actually be logged, it considered as logging best practice, at least in case there is only one argument (else it will harm readability)

ocs_ci/utility/utils.py
AviadP: t saves a bit of resources, the string formatting only happens if the message will actually be logged, it considered as logging best practice, at least in case there is only one argument (else it will harm readability)

ocs_ci/utility/utils.py
dahorak: Because return code other than `0` is considered as command failure, here it will be evaluated as `True`.

dahorak (approved):
Overall LGTM.
Would be good to have some example execution/comparison from previous state.

ocs_ci/utility/utils.py
AviadP: i may change it to `!=0` if it makes more sense

clacroix12: I believe we can also use `completed_process.ok`

ocs_ci/utility/utils.py
clacroix12: We technically prefer the formatting method that @AviadP implemented here for logging according to our [coding guidelines](https://github.com/red-hat-storage/ocs-ci/blob/master/docs/coding_guidelines.md), however I personally think f-strings are just fine since they tend to be more readable. Not sure if we want to update our guidelines. @ebenahar WDYT?

ebenahar: no specific preference. Both options would work

ebenahar (approved):

AviadP:
Need to re consider base64 trimming
Need to re consider what information is printed to log or returning to caller

AviadP:
closing this PR now, as discussed, we will split it into smaller PRs

PR #12652, opened 2025-07-14
Reduce noise from retry decorator

clacroix12:
Remove warnings after each exception in retry logic. Summarize intent and exceptions hit at debug level

### DIFF
--- a/ocs_ci/utility/retry.py
+++ b/ocs_ci/utility/retry.py
@@ -67,7 +67,7 @@ def retry(
             logger.debug(
                 f"Executing {f.__name__}. Tries: {mtries}. Delay: {mdelay}. Backoff: {backoff}"
             )
-            exception_summary = []
+            exception_summary = set()
             while mtries > 1:
                 try:
                     if func is not None:
@@ -84,7 +84,7 @@ def retry(
                                 f"Text: {text_in_exception} not found in exception: {e}"
                             )
                             raise
-                    exception_summary.append(e)
+                    exception_summary.add(repr(e))
                     time.sleep(mdelay)
                     mtries -= 1
                     mdelay = min(
@@ -92,7 +92,8 @@ def retry(
                     )  # Cap the delay to max_delay
                     if func is not None:
                         func()
-            logger.debug(f"Retry exception summary: {exception_summary}")
+            if exception_summary:
+                logger.debug(f"Retry exception summary: {exception_summary}")
             return f(*args, **kwargs)
 
         return f_retry


clacroix12:
Tested locally with a simple test which would fail some assertionerrors and the debug output we see looks like this:

```console
DEBUG    test_retry_decorator:test_retry_decorator.py:40 Executing do_something. Tries: 10. Delay: 0. Backoff: 1
DEBUG    test_retry_decorator:test_retry_decorator.py:67 Retry failure summary: [AssertionError('assert 1 == 11'), AssertionError('assert 8 == 11'), AssertionError('assert 5 == 11'), AssertionError('assert 4 == 11'), AssertionError('assert 1 == 11'), AssertionError('assert 2 == 11')
, AssertionError('assert 9 == 11'), AssertionError('assert 7 == 11'), AssertionError('assert 6 == 11')]
================================================================================================================================= short test summary info =================================================================================================================================
FAILED test_retry_decorator.py::test_retry_noise - assert 10 == 11

```

We summarize the basic retry configuration at the start as well as the exceptions that were _caught_ and triggered subsequent retries. Note that this does not include the final exception which is actually raised. That is logged as the actual failure for the test / logged as a legitimate error in the logs already.

ocs_ci/utility/retry.py
```diff
@@ -80,14 +84,15 @@ def f_retry(*args, **kwargs):
                                 f"Text: {text_in_exception} not found in exception: {e}"
                             )
                             raise
-                    logger.warning("%s, Retrying in %d seconds..." % (str(e), mdelay))
+                    exception_summary.append(e)
```
petr-balogh: As we discussed on our 1:1, use set or at the end print out just unique error messages - to avoid duplicate lines in output.

ocs_ci/utility/retry.py
```diff
@@ -64,6 +64,10 @@ def deco_retry(f):
         @wraps(f)
         def f_retry(*args, **kwargs):
             mtries, mdelay = tries, delay
+            logger.debug(
+                f"Executing {f.__name__}. Tries: {mtries}. Delay: {mdelay}. Backoff: {backoff}"
```
petr-balogh, line 68: we can probably print this one only on first retry as there can be case where we will not re-try at all

ocs_ci/utility/retry.py
clacroix12: I've updated to maintain a set instead of a list.

ocs_ci/utility/retry.py
clacroix12: This code doesn't really differentiate between the initial attempt and the retries so I just logged this before the while loop where the wrapped function is called. We could add a few lines to log it after the first failure if we think it's worth it to keep, but I could also just remove this log statement if we don't think it provides a lot of value to our debug logs.

vavuthu (approved):

ebenahar (approved):

PR #12670, opened 2025-07-16
Introduce lock for downloaing subctl binary: As there are multiple th…

shylesh:
…reads trying to download the binary we need a locking mechanism

### DIFF
--- a/ocs_ci/ocs/utils.py
+++ b/ocs_ci/ocs/utils.py
@@ -8,6 +8,7 @@ import time
 import traceback
 import subprocess
 import shlex
+import threading
 from subprocess import TimeoutExpired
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/utils.py
```diff
@@ -57,6 +58,15 @@
 log = logging.getLogger(__name__)
 
 
+mg_fail_count = 0
+mg_skip_count = 0
+mg_last_fail = None
+mg_collected_logs = 0
+mg_collected_types = set()
+mg_lock = threading.Lock()
```
petr-balogh: Delete as those are not relevant to this change and required change is not part of release-4.17 branch yet

shylesh: done

PR #12684, opened 2025-07-17
skip capacity_breakdown test if disconnected. Container not found problem. Cluster cant download image

DanielOsypenko:
We need to skip execution of the test on disconnected environment to avoid failure noise. This is cluster env limitation.


ebondare (approved):

dahorak:
I think we can fix the issue here and enable this test for disconnected environments: https://github.com/red-hat-storage/ocs-ci/pull/12745

DanielOsypenko:
Thanks @dahorak , deleting this PR

PR #12632, opened 2025-07-11
Add cli arg for report path and add reportportal related testsuite properties

clacroix12:
I've added a new CLI argument `--report <filepath>` to specify where the generated junit report will be placed. I've also the reportportal related testsuite properties to the junit report.

PR #12689, opened 2025-07-17
[release-4.18] Add report cli arg for fusion and fdf

clacroix12:
Add cli arg for report path and add reportportal related testsuite properties (#12632)

### DIFF
--- a/ocs_ci/utility/framework/fusion_fdf_init.py
+++ b/ocs_ci/utility/framework/fusion_fdf_init.py
@@ -14,7 +14,7 @@ from ocs_ci.framework.exceptions import (
 )
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.exceptions import CommandFailed
-from ocs_ci.utility import utils
+from ocs_ci.utility import reporting, utils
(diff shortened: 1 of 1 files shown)


vavuthu (approved):

dahorak (approved):

ocs_ci/utility/framework/fusion_fdf_init.py
```diff
@@ -104,6 +106,9 @@ def init_cli(self, args: list) -> list:
             default=[],
             help="Path to config file. Repeatable.",
         )
+        parser.add_argument(
+            "--report", default=None, help="Filepath for generated junit report"
```
dahorak, line 110: The `report` name seems to be quite generic and not directly clear what kind of report it provides (e.g. xml, html,...). In `ocs-ci` (or better in pytest) there is `--junit-xml`  parameter for that, what about using the same?

clacroix12: This is a cherry pick of https://github.com/red-hat-storage/ocs-ci/pull/12632 but we can change the param name and then backport that change as well.

Issue #11411, opened 2025-02-17
test_nonexist_bucket_delete[CLI] - Consistent test case failures - a fix of the test is needed


ypersky1980:
test_nonexist_bucket_delete[CLI] - Consistent test case failures ( on 4.17, also on 4.19 and 4.18 ) - a fix of the test is needed 

Relevant links: 

https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/795/29140/1435356/1435392/log
https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/795/29310/1448513/1448549/log
https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/795/29490/1456971/1457011/log
https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/795/29717/1468631/1468667/log

https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/852/30759/1512463/1512498/log
https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/852/30399/1499220/1499256/log

https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/795/27552/1358966/1359006/log
https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/795/27576/1359798/1359838/log
https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/795/27662/1364550/1364586/log
https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/795/28258/1391450/1391486/log
https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/795/28347/1394937/1394977/log
https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/795/28545/1401597/1401633/log
https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/795/28698/1408629/1408669/log

Backtrace : 

[...]

PR #12707, opened 2025-07-20
A fix to test_nonexist_bucket_delete test case

ypersky1980:
This PR fixes : https://github.com/red-hat-storage/ocs-ci/issues/11411.

The error message was updated, therefore we needed to catch the correct message in the test code.

### DIFF
--- a/tests/functional/object/mcg/test_bucket_deletion.py
+++ b/tests/functional/object/mcg/test_bucket_deletion.py
@@ -164,7 +164,7 @@ class TestBucketDeletion(MCGTest):
                 cli_del = mcg_obj.exec_mcg_cmd(f"obc delete {name}")
                 assert cli_del, "Unexpected cli delete non-exist OBC succeed"
             except CommandFailed as err:
-                assert "Could not delete OBC" in str(
+                assert "Could not delete. OBC" in str(
                     err
                 ), "Couldn't verify delete non-exist OBC with cli"
         logger.info(f"Delete non-exist OBC {name} failed as expected")


tests/functional/object/mcg/test_bucket_deletion.py
```diff
@@ -59,25 +59,41 @@ class TestBucketDeletion(MCGTest):
                     {
                         "interface": "OC",
                         "backingstore_dict": {"aws": [(1, "eu-central-1")]},
+                        "timeout": 1200,
```
keesturam: How are we arriving at this timeout value? maybe use timeoutsampler? 
if the test fails due to a automation issue, everytime we'd run the test and timout after 1200 seconds which seems to be inefficient.

ypersky1980: @keesturam The existing default timeout is 600 sec. The test occasionally fails ( in about 10%) since sometimes it takes longer . Anyway we have a timeout sampler, so in case the backingstore is ready before -the test will not wait 1200 secs. 
It is only for the rare cases when 600 secs is not enough. So that we won't have those failures. 
So can we leave it? The meaning is definitely not wait for 20 mins each time we create a backingstore.

keesturam: Do we know what causes the backingstore operation to take more than 10 minutes? is it delayed from cloud provider?

ypersky1980: @keesturam I don't know what cauyses the backingstore operation to take more than 10 mins. I'ts happens rarely and I do not see any reason. We might not change the timeout, and then we'll just see those occasional errors. 
So what do you prefer? 
I think that no harm will be done in case we increase the timeout to 1200 secs, since we have a timeout sampler anyway. Don't you think so?

keesturam: I think we should assert and exit the test in such case and log the reason for failure being the cloud provider being slow.

ocs_ci/ocs/resources/bucketclass.py
```diff
@@ -165,12 +165,18 @@ def _create_bucket_class(bucket_class_dict):
...
+            try:
+                backingstores = [
+                    backingstore
+                    for backingstore in backingstore_factory(
+                        interface,
+                        bucket_class_dict["backingstore_dict"],
+                        timeout=timeout,
+                    )
+                ]
+            except TimeoutExpiredError:
```
keesturam, line 177: we don't know if cloud provider is slow, no? The log message could be,
'Backingstore was not created after <> seconds'

ypersky1980: @keesturam ok, I'll change the error message to "Backingstore was not created after 600 seconds"

keesturam (approved):

ocs_ci/ocs/resources/bucketclass.py
```diff
@@ -165,12 +165,18 @@ def _create_bucket_class(bucket_class_dict):
...
+                log.info("Backingstore was not created after 600 secs")
```
udaysk23:
```suggestion
                log.info(f"Backingstore was not created after {timeout} secs")
```

keesturam (approved):

PR #12725, opened 2025-07-21
Adjust stage testing for konflux builds

petr-balogh:


ebenahar (approved):

petr-balogh:
Verified with normal deployment and disconnected deployment as well. Both worked

PR #12735, opened 2025-07-22
delete subscriptions webhook and bring down ocs-client-operator-controller-manager

vavuthu:

To upgrade from 4.19 to 4.20, bring down ocs-client-operator-controller-manager and delete subscriptions webhook.

https://issues.redhat.com/browse/DFBUGS-3572

### DIFF
--- a/ocs_ci/ocs/ocs_upgrade.py
+++ b/ocs_ci/ocs/ocs_upgrade.py
@@ -42,7 +42,11 @@ from ocs_ci.ocs.resources.csv import (
     get_csvs_start_with_prefix,
 )
 from ocs_ci.ocs.resources.install_plan import wait_for_install_plan_and_approve
-from ocs_ci.ocs.resources.pod import get_noobaa_pods, verify_pods_upgraded
+from ocs_ci.ocs.resources.pod import (
+    get_noobaa_pods,
+    verify_pods_upgraded,
+    wait_for_pods_deletion,
+)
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -787,6 +787,18 @@ def run_ocs_upgrade(
             time.sleep(120)
         else:
             ui_upgrade_supported = False
+            if original_ocs_version == "4.19" and upgrade_version == "4.20":
```
dahorak, line 794: I would suggest to use:
```suggestion
            if original_ocs_version == version.VERSION_4_19 and upgrade_version == version.VERSION_4_20:
```
(`version.VERSION_4_20` is probably not yet defined in the `version` module)

petr-balogh: the version here is taken directly from config which is string, so Vijay's implementation is ok, it's already used like this on line 740

vavuthu:
verified here https://url.corp.redhat.com/ac84e19
(BUILD ID: 4.20.0-43 RUN ID: 1753283468613)

Result 	Test 	Description 	Duration 	Links
Passed 	tests/functional/upgrade/test_upgrade.py::test_upgrade 	Tests upgrade procedure of OCS cluster 	2282.00

dahorak (approved):

PR #12770, opened 2025-07-23
Removed test TestDefaultMetadataDisabled since metadata is enabled by default from 4.19

paraggit:
Removed test `TestDefaultMetadataDisabled` since metadata is enabled by default from 4.19

Refer below feature  changes:

Bug: https://issues.redhat.com//browse/DFBUGS-2395
PR: https://github.com/red-hat-storage/ocs-client-operator/pull/378


paraggit:
> Hi Parag,
> 
> IIUC the feature was disabled from 4.12 till 4.18 and enabled from 4.19
> 
> So this test is expected to continue being executed for the versions less than 4.19. Am I correct in this understanding?
> 
> Are we just removing it from 4.19? where is that check added?

Yes Neha, the test is valid for releases < 4.19. We pull tests based on the release tag, so it will continue to run on earlier versions. Since it’s not valid for 4.19, I am removing it from this release.

nehaberry (approved):

PR #12805, opened 2025-07-27
adding noobaa admin test case

ypersky1980:
This PR implements noobaa admin access to a bucket with PublicAccessBlock settings applied to it.

### DIFF
--- a/tests/functional/object/mcg/test_bucket_policy.py
+++ b/tests/functional/object/mcg/test_bucket_policy.py
@@ -99,6 +99,178 @@ def delete_bucket_policy_verify(s3_obj, bucket_name):
             )
 
 
+def put_allow_bucket_policy(bucket_name, mcg_obj):
+    """
+    This utility function puts allow policy on the specified bucket
+    Args:
+        bucket_name (str): The bucket on which the policy should be put
+        mcg_obj (obj): An object representing the current state of the MCG in the cluster
+    """
+
+    bucket_policy_generated = gen_bucket_policy(
+        user_list="*",
+        actions_list=["*"],
+        resources_list=[bucket_name, f'{bucket_name}/{"*"}'],
+    )
+    bucket_policy = json.dumps(bucket_policy_generated)
+    # Put bucket policy
+    logger.info(f"Putting bucket policy {bucket_policy} on bucket: {bucket_name}")
+    put_bucket_policy(mcg_obj, bucket_name, bucket_policy)
+    # Hardcoded sleep is needed because we lack a confirmation mechanism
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_bucket_policy.py
```diff
@@ -1455,127 +1546,112 @@ def test_public_access_block_anonymous(
...
+            f"ls -A1 {AWSCLI_TEST_OBJ_DIR}"
+        ).split(" ")
+        file_name = standard_test_obj_list[0]
+        logger.info(f"Going to copy file {file_name} to the bucket {bucket_name}")
+
+        check_ls_command(mcg_obj, awscli_pod_session, bucket_name, file_name)
+
+        put_allow_bucket_policy(bucket_name, mcg_obj)
+
+        check_ls_command(mcg_obj, awscli_pod_session, bucket_name, file_name)
```
udaysk23, line 1638: Can't we use `check_commands` method in this test?

ypersky1980: @udaysk23 . Check_commands is rather a distinct method which tests mainly anonymous user access. It already has rather complicated logic. I thought about combining two methods in one, but I reached the conclusion that this  would create even more complicated code with a lot of 'if-else' and would produce almost unreadable code. That is why I preferred to write it this way

udaysk23 (approved):

udaysk23 (approved):

tests/functional/object/mcg/test_bucket_policy.py
```diff
@@ -99,6 +99,178 @@ def delete_bucket_policy_verify(s3_obj, bucket_name):
...
+    bucket_policy_generated = gen_bucket_policy(
+        user_list="*",
+        actions_list=["*"],
+        resources_list=[bucket_name, f'{bucket_name}/{"*"}'],
+    )
+    bucket_policy = json.dumps(bucket_policy_generated)
+    # Put bucket policy
+    logger.info(f"Putting bucket policy {bucket_policy} on bucket: {bucket_name}")
+    put_bucket_policy(mcg_obj, bucket_name, bucket_policy)
+    # Hardcoded sleep is needed because we lack a confirmation mechanism
```
keesturam, line 119: would it make sense to open a bug to introduce this mechanism?

keesturam (approved):

tests/functional/object/mcg/test_bucket_policy.py
ypersky1980: @keesturam we have a bug on this : https://issues.redhat.com/browse/DFBUGS-3654

PR #12808, opened 2025-07-28
use full paths for custom ingress and api certificate related files

dahorak:


dahorak:
Provider-Client verification job: https://url.corp.redhat.com/b0c8320
There is no test failed with the error:
```
failed on setup with "FileNotFoundError: [Errno 2] No such file or directory: 'data/ca.crt'"
```

PR #12828, opened 2025-08-04
Fix for RBD and CEPHPFS daemonset name as per 4.19 change

suchita-g:
Fixes issue #12831. and issue #12524


tests/functional/upgrade/test_configuration.py
```diff
@@ -10,14 +10,14 @@
     brown_squad,
     skipif_mcg_only,
     tier1,
-    jira,
+    # jira,
```
AviadP, line 13: plz remove it if not needed, dont leave commented code

suchita-g:
this test is getting skipped in newer ODF versions for "DFBUGS-3826". so closing this PR

Issue #11663, opened 2025-03-19
In the resize osd verification steps fix how we check the expected ceph capacity


yitzhak12:
See, for example, this failure: https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/795/30129/1489520/1489861/log

ebenahar:
According to @yitzhak12 seen only with ROSA. 
To be tried next week on ROSA (May 19th)

ebenahar:
on ROSA we have a different monitoring stack. Could this impact the way we validate Ceph capacity? @yitzhak12 @DanielOsypenko

ebenahar:
@yitzhak12 please update post the conversation with Daniel

DanielOsypenko:
* when we were validating ROSA HCP this inconsistency was hidden with another bug 
https://issues.redhat.com/browse/DFBUGS-965 it was fixed and verified and this issue we could not see.

* increase of the size should have give us 1024Gb of storage 

test_resize_osd: 
[2025-03-18T01:48:38.946Z] 21:48:38 - MainThread - ocs_ci.helpers.osd_resize - [32mINFO[0m  - The current osd size is 512Gi
[2025-03-18T01:48:38.946Z] 21:48:38 - MainThread - ocs_ci.helpers.osd_resize - [32mINFO[0m  - Increase the osd size to 1024Gi

* during the test before failing we had logs confirming that storage indeed has increased to 1024Gb. Data taken from PV sizes
[2025-03-18T02:04:55.757Z] 22:04:55 - MainThread - ocs_ci.helpers.osd_resize - [32mINFO[0m  - PV sizes [1024, 1024, 1024]

* right before Assertion we had confirmation that both expected and actual sizes are matching, meaning storage increase was successfl
```
[2025-03-18T02:13:18.961Z] 22:13:18 - MainThread - ocs_ci.helpers.osd_resize - [32mINFO[0m  - The expected storage size is 1024Gi
...
[2025-03-18T02:13:19.519Z] 22:13:19 - MainThread - ocs_ci.helpers.osd_resize - [32mINFO[0m  - The current storage size is 1024Gi
```
* nothing is done on cluster during the test itself before and after the logs ^^. Test has only calculations and assertions
https://github.com/red-hat-storage/ocs-ci/blob/1ddf17dcf8b89d51a40c2dcff45d13174e362ecb/ocs_ci/helpers/osd_resize.py#L161

[...]

PR #12841, opened 2025-08-05
Fix the Ceph capacity validation after the OSD resize test

yitzhak12:
fix https://github.com/red-hat-storage/ocs-ci/issues/11663
In this PR, we changed the Ceph capacity validation to be based on the expected storage size per OSD and the number of OSDs. This provides a direct and accurate way to verify that the Ceph raw capacity reflects the updated OSD sizes after a resize operation.

### DIFF
--- a/ocs_ci/helpers/osd_resize.py
+++ b/ocs_ci/helpers/osd_resize.py
@@ -19,7 +19,6 @@ from ocs_ci.ocs.resources.storage_cluster import (
     get_device_class,
     verify_storage_device_class,
     verify_device_class_in_osd_tree,
-    get_deviceset_count,
     resize_osd,
 )
 from ocs_ci.ocs.cluster import check_ceph_osd_tree, CephCluster, check_ceph_osd_df_tree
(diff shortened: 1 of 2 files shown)


ocs_ci/helpers/osd_resize.py
```diff
@@ -192,20 +192,40 @@ def check_storage_size_is_reflected(expected_storage_size):
             f"The PV sizes are not equal to the expected storage size {expected_storage_size_in_gb}"
         )
 
+    if expected_ceph_capacity:
+        check_ceph_capacity_increased(expected_ceph_capacity)
+
+
+def check_ceph_capacity_increased(
```
AviadP, line 199: do we have validation that all OSDs actually resized successfully before capacity check?

ocs_ci/helpers/osd_resize.py
```diff
@@ -192,20 +192,40 @@ def check_storage_size_is_reflected(expected_storage_size):
...
+    expected_ceph_capacity,
+    tolerance_percent=1.0,
```
AviadP, line 201: can you explain in why 1.0 in docstring?

ocs_ci/helpers/osd_resize.py
yitzhak12: Yes, we get the osd size from the storage cluster. See the line https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/helpers/osd_resize.py#L162

ocs_ci/helpers/osd_resize.py
yitzhak12: The expected and actual Ceph capacity should ideally be the same. However, in practice, minor mismatches can occur due to the way Ceph calculates capacity in bytes, which may introduce minor differences. Additionally, rounding issues can arise when converting between different size units (e.g., TiB to GiB), which can slightly affect the calculated values. To account for these variations, we apply a small tolerance in the validation.
A 1.0% difference seems reasonable, but we can consider a higher threshold if needed.

AviadP: maybe worth mentioning it in functions docstring

AviadP (approved):
LGTM

yitzhak12:
The test passed also with a ROSA cluster: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/54666/.

PR #12888, opened 2025-08-11
fix_ceph_health to True in install verification

vavuthu:

### DIFF
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -777,7 +777,7 @@ def ocs_install_verification(
                 namespace,
                 health_check_tries,
                 health_check_delay,
-                fix_ceph_health=rdr_run,
+                fix_ceph_health=True,
             )
         except CephHealthRecoveredException as ex:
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -777,7 +777,7 @@ def ocs_install_verification(
                 namespace,
                 health_check_tries,
                 health_check_delay,
-                fix_ceph_health=rdr_run,
+                fix_ceph_health=True,
```
petr-balogh, line 780: Do we have any issue linked to this one?
As this was as W/A for the issue mentioned below.

```
        except CephHealthRecoveredException as ex:
            if rdr_run and "slow ops" in str(ex):
                # Related issue: https://github.com/red-hat-storage/ocs-ci/issues/11244
                log.warning("For RDR run we ignore slow ops error as it was recovered!")
```

So we anyway should raise an exception , but we are ignoring it for slow ops error for rdr deployment. Otherwise it will anyway probably raise the exception and not ignore it.

dahorak (approved):

PR #12910, opened 2025-08-12
OCP 4.20 changes for RDR deployment

am-agrawa:


ocs_ci/ocs/acm/acm.py
```diff
@@ -566,16 +566,16 @@ def login_to_acm():
     driver = login_ui(url)
     page_nav = AcmPageNavigator()
     page_nav.page_has_loaded(retries=10, sleep_time=5)
-    look_for_local_cluster = page_nav.wait_until_expected_text_is_found(
+    look_for_administrator_dropdown = page_nav.wait_until_expected_text_is_found(
```
DanielOsypenko, line 569: Hi Aman, 
I checked our ACM setup with OCP 4.20 and MCE 2.9.0, and local-cluster is still there.
Please work out two scenarios so that, in case local-cluster does not work, an Administrator is checked.

Image

DanielOsypenko (changes requested):

am-agrawa:
UI code until submariner installation passed- https://url.corp.redhat.com/9bd4a8f

ocs_ci/ocs/acm/acm.py
```diff
@@ -566,16 +566,16 @@ def login_to_acm():
...
         locator=page_nav.acm_page_nav["click-local-cluster"],
-        expected_text="local-cluster",
+        expected_text="Administrator",
```
jilju, line 571: Usually this is the first page that will load. The options will change based on the installed operators. "Administrator", "Virtualization", "Fleet Manager" are the options we have seen. Instead of looking for the "Administrator", can click the button using some other property to avoid any failure if the loading does not have "Administrator" menu.

jilju: "local-cluster" , "All Clusters" option is also available in 4.20. In cluster where ACM is installed this was not seen.

am-agrawa:
Opened https://github.com/red-hat-storage/ocs-ci/pull/12933 out of local branch, and not master.

ocs_ci/ocs/acm/acm.py
am-agrawa: fixed in https://github.com/red-hat-storage/ocs-ci/pull/12933

PR #12923, opened 2025-08-12
Fix test_pv_data_dist

sagihirshfeld:
1. Create the pv-pool pods with increased CPU and MEM resources to prevent them from crashing due to OOM issue
2. Optimize the uploading of the objects:
- Use s3 sync to upload many small objects at once
- Upload objects with s3 cp one-by-one for bigger objects

### DIFF
--- a/tests/functional/object/mcg/test_pv_pool.py
+++ b/tests/functional/object/mcg/test_pv_pool.py
@@ -429,7 +429,6 @@ class TestPvPool:
         )
 
     @tier2
-    @pytest.mark.skip(reason="Skip due to issue #12717")
     @pytest.mark.parametrize(
         argnames=["pv_in_bs", "block_size", "block_count", "file_count"],
         argvalues=[
@@ -438,7 +437,7 @@ class TestPvPool:
                     1,
                     "5K",
                     1,
-                    50000,  # dataset contains 50000 small files of 5K each, 1 pv per backingstore
+                    10000,  # dataset contains 50000 small files of 5K each, 1 pv per backingstore
                 ],
                 marks=pytest.mark.polarion_id("OCS-6852"),
             ),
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_pv_pool.py
```diff
@@ -438,7 +437,7 @@ def test_pv_pool_with_nfs(
                     1,
                     "5K",
                     1,
-                    50000,  # dataset contains 50000 small files of 5K each, 1 pv per backingstore
+                    10000,  # dataset contains 50000 small files of 5K each, 1 pv per backingstore
```
udaysk23: Please change comment value to `10000`

sagihirshfeld: Fixed in https://github.com/red-hat-storage/ocs-ci/pull/12923/commits/d7ffaefd403ffac0e31f312155451eb82cf29866

sagihirshfeld:
Since the passing verification I added https://github.com/red-hat-storage/ocs-ci/pull/12923/commits/d7ffaefd403ffac0e31f312155451eb82cf29866

This commit is mostly semantic, with the exception of the increase to the number of files for the `OnePV_ManySmall_Files` case from 10,000 to 25,000. I halved the original 50,000 to reduce runtime.

tests/functional/object/mcg/test_pv_pool.py
```diff
@@ -533,34 +532,78 @@ def test_pv_data_dist(
 
         """
 
+        pv_backingstore_specs = {
+            "vol_num": pv_in_bs,
+            "size": MIN_PV_BACKINGSTORE_SIZE_IN_GB,
+            "storagecluster": CEPHBLOCKPOOL_SC,
```
udaysk23, line 538: I think `skip on mcg cluster` mark is needed for this test

sagihirshfeld: Do you mean the `skipif_mcg_only` decorator? It's already set on its class: 
https://github.com/red-hat-storage/ocs-ci/blob/master/tests/functional/object/mcg/test_pv_pool.py/#L50

PR #12937, opened 2025-08-14
<GSS Closed Loop BZ: 2231860> : Add a test to verify no CLBO seen on force deletion of noobaa-db pod repeatedly

Akarsha-rai:
This PR covers and test the scenarios mentioned in the BZ 2231860.

- The PR attached to this bug is also solving the bug 2165907#c35 so used these steps instead of bug 2231860#c15

- Test to verify no CLBO seen on force deletion of noobaa-db pod repeatedly and check data availability for MCG Namespace bucket

### DIFF
--- a/tests/functional/object/mcg/test_namespace_crd.py
+++ b/tests/functional/object/mcg/test_namespace_crd.py
@@ -1080,15 +1080,27 @@ class TestNamespace(MCGTest):
 
     @tier4c
     @pytest.mark.parametrize(
-        argnames=["mcg_pod"],
+        argnames=["mcg_pod", "respin_count"],
         argvalues=[
             pytest.param(
                 *["noobaa-db"],
+                respin_count=0,
                 marks=[pytest.mark.polarion_id("OCS-2291")],
             ),
-            pytest.param(*["noobaa-core"], marks=pytest.mark.polarion_id("OCS-2319")),
             pytest.param(
-                *["noobaa-operator"], marks=pytest.mark.polarion_id("OCS-2320")
+                *["noobaa-db"],
+                respin_count=10,
+                marks=[pytest.mark.polarion_id("OCS-6907")],
+            ),
+            pytest.param(
+                *["noobaa-core"],
+                respin_count=0,
+                marks=pytest.mark.polarion_id("OCS-2319"),
+            ),
+            pytest.param(
+                *["noobaa-operator"],
+                respin_count=0,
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_namespace_crd.py
```diff
@@ -1148,17 +1161,33 @@ def test_respin_mcg_pod_and_check_data_integrity_crd(
...
+            noobaa_pods = pod.get_noobaa_pods()
+            pod_obj = [pod for pod in noobaa_pods if pod.name.startswith(mcg_pod)][0]
+            pod_obj.delete(force=True)
+            logger.info("Wait for noobaa pods to come up")
+            assert pod_obj.ocp.wait_for_resource(
+                condition="Running",
+                selector="app=noobaa",
+                resource_count=len(noobaa_pods),
+                timeout=1000,
+            )
```
udaysk23: Since both condition has same code except iteration loop, I would suggest to create method inside this test and call it in if..else block

Akarsha-rai:
Test passed: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/55252/testReport/

tests/functional/object/mcg/test_namespace_crd.py
Akarsha-rai: done

tests/functional/object/mcg/test_namespace_crd.py
```diff
@@ -1080,15 +1080,27 @@ def test_create_ns_bucket_from_utilized_resources_crd(
 
     @tier4c
     @pytest.mark.parametrize(
-        argnames=["mcg_pod"],
+        argnames=["mcg_pod", "respin_count"],
```
udaysk23: `respin_count` is little misleading since we are passing boolean value. I would suggest you to change it to `respin_pod` or something similar to it

Akarsha-rai: By default, the test case respins the pod once and I agree the name `respin_count` can be a bit misleading since we're using it a Boolean. The idea was:

- False --> default 1 respin
- True --> respin 10 times

I'm thinking to rename it to `respin_multiple_times` to better reflect the Boolean nature, let me know your thoughts?

udaysk23: `respin_multiple_times` looks more readable

PrasadDesala (approved):

PR #12960, opened 2025-08-17
Add decorator to wrap functions with RunWithProviderConfigContextIfAvailable

sagihirshfeld:
### What
Introduce a new decorator to wrap functions or methods, as if their content were ran inside a `with config.RunWithProviderConfigContextIfAvailable():` block. The purpose of which is to make sure certain code runs only on the provider cluster and not on the clients.

### Why
We need to modify many functions and methods (especially on MCG relalted tests) so they run only on provider. The current solution works, but it results in messy code and big PRs where we'll indent hundreds if not thousands of rows to fit inside `with config.RunWithProviderConfigContextIfAvailable()` blocks.

This decorator would simplify this process.

### How
- Added @staticmethod MultiClusterConfig.run_with_provider_context_if_available
- Added a new libtest to verify its functionality

### DIFF
--- a/ocs_ci/framework/__init__.py
+++ b/ocs_ci/framework/__init__.py
@@ -9,6 +9,7 @@ under section PYTEST_DONT_REWRITE
 
 # Use the new python 3.7 dataclass decorator, which provides an object similar
 # to a namedtuple, but allows type enforcement and defining methods.
+import functools
 import os
 import yaml
 import logging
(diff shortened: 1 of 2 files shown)


sagihirshfeld:
Verified a run of the new libtest here on a provider-client deployment: https://url.corp.redhat.com/cec2911

PR #12964, opened 2025-08-18
OCSQECL-5384 corrective measures. test_prometheus_rule_failures

DanielOsypenko:
With this PR we expanded `test_prometheus_rule_failures` to check wether warning appears on failed 'many-to-many matching not allowed'. PrometheusRuleFailures can fire with this warn message or not, and in next case we are failing test
a) PrometheusRuleFailures query return message.Ok != True 
c) PrometheusRuleFailures query is fired and 'many-to-many matching not allowed'  error log is found in primetheus-k8s pod
d) PrometheusRuleFailures query is NOT fired, but 'many-to-many matching not allowed'  error log is found in primetheus-k8s pod 
* this may happen when custom query via Prometheus http req or via UI has violated 'many-to-many matching not allowed' requirement
 e) PrometheusRuleFailures query is fired, but 'many-to-many matching not allowed'  error log is NOT found in primetheus-k8s pod

### DIFF
--- a/tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
+++ b/tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
@@ -69,13 +69,6 @@ def test_prometheus_rule_failures(threading_lock):
         pod_logs.reverse()
         for log_line in pod_logs:
             if "many-to-many matching not allowed" in log_line.lower():
-                log.error(
-                    f"Found 'warn' msg in logs of pod {pod_name}:"
-                    f"\n\n****************** Log message: ******************\n{log_line}"
-                    f"\n****************** End of logs ******************"
-                    f"\n\nThis is a known issue. Follow OCSQECL-5384 and knowledge base article "
-                    f"solutions/7103336 for more details."
-                )
                 test_results[f"many-to-many-error-present-{pod_name}-check"] = False
                 break
         else:


DanielOsypenko:
verification run on Healthy cluster:
https://url.corp.redhat.com/174e720

DanielOsypenko:
test executed on pre-fix 4.17.5 ODF environment - failed, as expected
https://url.corp.redhat.com/5220982

fbalak (dismissed):

DanielOsypenko:
I doublechecked with Arun Mohan (leading dev on monitoring stack)

> Daniel Osypenko
> when I have this message in prometheus-k8s logs “many-to-many matching not allowed: matching labels must be unique on one side” will I always see that PrometheusRuleFailures alert fire or these are not related?
> 
> Arun Mohan
> Even though the logs is stating the same reason many-to-many matching not allowed: .... the query that caused this issue can vary
> You will always see a PrometheusRuleFailures alert
> which rule failed may vary (according to query)

tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
```diff
@@ -39,17 +40,48 @@ def test_alerting_works(threading_lock):
...
+            namespace=constants.MONITORING_NAMESPACE,
+        ).splitlines()
+        pod_logs.reverse()
+        for log_line in pod_logs:
+            if "many-to-many matching not allowed" in log_line.lower():
+                log.error(
+                    f"Found 'warn' msg in logs of pod {pod_name}:"
+                    f"\n\n****************** Log message: ******************\n{log_line}"
+                    f"\n****************** End of logs ******************"
+                    f"\n\nThis is a known issue. Follow OCSQECL-5384 and knowledge base article "
```
PrasadDesala: OCSQECL-5384? instead should be add DFBUG id?

tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
```diff
@@ -39,17 +40,48 @@ def test_alerting_works(threading_lock):
...
+        log.info(f"Checking logs of pod {pod_name}")
+        pod_logs = get_pod_logs(
+            pod_name=pod_name,
+            namespace=constants.MONITORING_NAMESPACE,
+        ).splitlines()
+        pod_logs.reverse()
+        for log_line in pod_logs:
+            if "many-to-many matching not allowed" in log_line.lower():
+                log.error(
+                    f"Found 'warn' msg in logs of pod {pod_name}:"
```
PrasadDesala: IIUC, the fix has already been backported to a few zstreams and is pending for the others. Once the fix is fully backported to all relevant zstream releases, there is no need for this code?

tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
```diff
@@ -39,17 +40,48 @@ def test_alerting_works(threading_lock):
 def test_prometheus_rule_failures(threading_lock):
     """
     There should be no PrometheusRuleFailures alert when OCS is configured.
+    This test is extended to check for many-to-many matching errors in Prometheus logs.
```
PrasadDesala: Please include the DFBUG number here to indicate that this scenario is added from a customer reported bug

tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
```diff
@@ -39,17 +40,48 @@ def test_alerting_works(threading_lock):
...
     log.info(f"Prometheus Alerts: {alerts}")
-    assert constants.ALERT_PROMETHEUSRULEFAILURES not in [
-        alert["labels"]["alertname"] for alert in alerts
-    ]
+    test_results[f"{constants.ALERT_PROMETHEUSRULEFAILURES}-present-check"] = (
+        constants.ALERT_PROMETHEUSRULEFAILURES
+        not in [alert["labels"]["alertname"] for alert in alerts]
+    )
+    prometheus_pods = get_pod_name_by_pattern(
+        "prometheus-k8s", constants.MONITORING_NAMESPACE
```
PrasadDesala: you can replace it with defaults.PROMETHEUS_ROUTE

tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
```diff
@@ -39,17 +40,48 @@ def test_alerting_works(threading_lock):
...
+    )
+    for pod_name in prometheus_pods:
+        log.info(f"Checking logs of pod {pod_name}")
+        pod_logs = get_pod_logs(
+            pod_name=pod_name,
+            namespace=constants.MONITORING_NAMESPACE,
+        ).splitlines()
+        pod_logs.reverse()
+        for log_line in pod_logs:
+            if "many-to-many matching not allowed" in log_line.lower():
```
PrasadDesala, line 71: we have an existing function [search_pattern_in_pod_log](https://github.com/red-hat-storage/ocs-ci/blob/14ffa2a8bdc83b057f08bc7aff64628f5ce31e71/ocs_ci/ocs/resources/pod.py#L3826). Please use it if it is suitable for your case here

tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
DanielOsypenko: I think we want to cover this scenario to not miss bug in future.
This issue can return when alerts are modified or added, and warn message will be logged again.

tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
DanielOsypenko: Why? 
This test improvement is a part of Corrective measures and Escape analysis initiative. It has link to a 'triggered by' bug. 
I could not find anything specific to what we need to write in comments in ocs-ci, but the closest one is 
* Both the analysis and corrective measures implementation is done and tracked in JIRA project "OCSQE Close Loop (OCSQECL)" in ODF QE closed loop process slides. I think we need to link them somehow.

tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
DanielOsypenko: done, thanks

tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
DanielOsypenko: [search_pattern_in_pod_log](https://github.com/red-hat-storage/ocs-ci/blob/14ffa2a8bdc83b057f08bc7aff64628f5ce31e71/ocs_ci/ocs/resources/pod.py#L3826)
this method has namespace limitation for some reason.

I'd want to search from the end of the log and not for the beginning. This will improve logi, efficiency and make PR lighter. 

Thanks for the suggestion, will have it in mind in future, maybe to extend this func, but for this PR I think it is out of scope.

tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
DanielOsypenko: added, thanks

tests/functional/monitoring/prometheus/alerts/test_alerting_works.py
PrasadDesala: I understand that this test is developed as part of the closed loop process, but the current OCSQECL JIRA does not contain any information about the escape analysis. So, I don’t think referencing the OCSQECL JIRA here would be helpful, as it lacks the necessary details.

IMO, since this issue is going to be fixed in the product, mentioning it as a known issue might cause confusion.
Also, I’m wondering if we really need to add this message. We've already included information about the scenario and DFBUG details in the description. The warning are not expected to be found according to the test and it would fail if it find any, eventually have to open a bug.

> Both the analysis and corrective measures implementation is done and tracked in JIRA project "OCSQE Close Loop (OCSQECL)" in ODF QE closed loop process slides. I think we need to link them somehow.

Once the PR is merged, we can add the below details in the Parent Story of this subtask
1) Add the PR links to your approved/merged github PRs.
2) Add polarion link to the test in "testing"

And we are also including the DFBUG number in the test description of the automated test to indicate that this scenario is added from a customer reported bug

DanielOsypenko: Will correct this according to your comment. Thanks

PrasadDesala (approved):

Akarsha-rai (approved):

suchita-g (approved):

PR #12978, opened 2025-08-19
Support RDR deployment on provider and client clusters

jilju:
Support RDR deployment on provider and client clusters.

* Added support to configure everything related to RDR for client clusters except submariner and backup.
* Use provider cluster for verification required, eg: verifying mirrorpeer, verify mirror pods.
* Use unique name for mirrorpeer and drpolicy.
* Updated the functions get_provider_index and get_consumer_indexes_list to avoid error if provider or client cluster is not available in config.
* Added a new parameter MULTICLUSTER.dr_cluster_relations to identify the pair of RDR clusters. There will be 1 pair for deployment and multiple pair for test execution. Updated README.md to add new parameter 'dr_cluster_relations'.
* Skip creating serviceexport for multiclient mode(provider mode) because hostNetwork is True.

### DIFF
--- a/ocs_ci/helpers/dr_helpers.py
+++ b/ocs_ci/helpers/dr_helpers.py
@@ -2354,7 +2354,7 @@ def create_service_exporter():
             config.ENV_DATA["platform"].lower()
             in constants.HCI_PROVIDER_CLIENT_PLATFORMS
         ):
-            logger.info(f"Skipping ServiceExport creation")
+            logger.info("Skipping ServiceExport creation for multiclient cluster")
             continue
         logger.info("Creating Service exporter")
         run_cmd(f"oc create -f {constants.DR_SERVICE_EXPORTER}")
--- a/ocs_ci/helpers/dr_helpers.py
+++ b/ocs_ci/helpers/dr_helpers.py
@@ -2293,9 +2293,16 @@ def validate_storage_cluster_peer_state():
     restore_index = config.cur_index
     managed_clusters = get_non_acm_cluster_config()
     for cluster in managed_clusters:
-        index = cluster.MULTICLUSTER["multicluster_index"]
+        if (
+            config.ENV_DATA["platform"].lower()
+            in constants.HCI_PROVIDER_CLIENT_PLATFORMS
+        ):
+            index = config.get_provider_index()
+        else:
+            index = cluster.MULTICLUSTER["multicluster_index"]
         config.switch_ctx(index)
         logger.info("Validating Storage Cluster Peer status")
+        # TODO: Check whether storageclusterpeer is created for each RDR pair of client clusters
         sample = TimeoutSampler(
             timeout=300,
             sleep=5,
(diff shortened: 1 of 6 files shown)


jilju:
Verified: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-rdr-setup/2709/

fbalak (dismissed):
Looks good. It would be good to try if it doesn't break standard provider mode deployment without RDR.

ocs_ci/utility/version.py
```diff
@@ -132,7 +132,9 @@ def get_ocs_version_from_csv(only_major_minor=False, ignore_pre_release=False):
     csvs = ocp.OCP(
         namespace=config.ENV_DATA["cluster_namespace"], kind="", resource_name="csv"
     )
-    if get_semantic_ocs_version_from_config() >= VERSION_4_9:
+    if config.ENV_DATA["cluster_type"].lower() == constants.HCI_CLIENT:
```
DanielOsypenko, line 135: version of an ODF client operator is not the same as OCS Operator version. In this case it will be right to switch to Provider and get version of ODF Operator, elsewise we are creating misleading logic here.

 Some reasoning
- if versions of ODF Provider and ODF Client on consumer cluster are different, we will hide a bug. 
- we can not always rely on fact that versions are the matching, this may change in future, or even not be true, for example, during upgrade.
- purpose of this func is to get ocs version from csv of ODF, this must correlate to version of ceph components, and dependency operators like rook-csi, etc.

ocs_ci/helpers/dr_helpers.py
```diff
@@ -2293,9 +2293,16 @@ def validate_storage_cluster_peer_state():
...
+        if cluster.ENV_DATA["cluster_type"].lower() == constants.HCI_CLIENT:
+            with config.RunWithConfigContext(
+                cluster.MULTICLUSTER["multicluster_index"]
+            ):
+                index = config.get_provider_index()
+        else:
+            index = cluster.MULTICLUSTER["multicluster_index"]
         config.switch_ctx(index)
         logger.info("Validating Storage Cluster Peer status")
+        # TODO: Check whether storageclusterpeer is created for each RDR pair of client clusters
```
DanielOsypenko, line 2305: not sure if it is leftover or a note for enhancement

ocs_ci/utility/version.py
jilju: Made the change to use provider

ocs_ci/helpers/dr_helpers.py
jilju: Yes, it is a leftover. Removed.

DanielOsypenko (approved):

jilju:
Single client deployment for RDR is done to ensure that no regression is caused by this change.
Build https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-rdr-setup/2731

fbalak (approved):

Issue #8270, opened 2023-08-16
NooBaa bucket S3 I.O fails with `InvalidBucketState`


sagihirshfeld:
In rare occasions we attempt to use buckets too soon after their creation. While we do currently attempt to verify the bucket's health by checking it's  status via mcg-cli or oc commands, we should check if it's available via S3 before declaring a bucket is ready for use.

udaysk23:
Re-opening this issue since we are hitting the same issue again
https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/465/16520/789606/789666/789667/log?logParams=history%3D789667%26page.page%3D1

sagihirshfeld:
This failed TC is from 4.13 where #8271 was yet to be backported to. Besides backporting the extra health check to earlier versions branches, we should keep an eye out to check whether it helped reduce the error in 4.14 and above.

keesturam:
@sagihirshfeld should we keep this issue open?

PR #13017, opened 2025-08-25
Increase retry max_attempts for MCG S3 resource and AWSCLI commands

### LINKED ISSUE #12924: [Bug]: test_bucket_list_command often fails at teardown
### Description
RP references:
- https://url.corp.redhat.com/4ba3ea8
- https://url.corp.redhat.com/ef2f504
- https://url.corp.redhat.com/9c389f0

sagihirshfeld:
### What
Both boto3 and AWSCLI’s S3 clients have a built-in retry mechanism using exponential backoff. Until now, we’ve relied on the default value of `max_attempts=4`, which results in only ~4 seconds of wait time in case of transient failures.

This PR increases the value to `max_attempts=8`, which allows retries for about one minute.

### Why
While OCS-CI has its own retry tools, those are mostly used for waiting on higher-level operations (e.g., bucket replication completion). Our tests generally assume that basic S3 commands succeed, so we don’t wrap them with `TimeoutSampler` or custom retries.

Occasionally, these basic operations fail unexpectedly due to transient issues (e.g., a noobaa-core pod restart). By leveraging the client’s built-in retry mechanism, we can handle these failures transparently and with minimal changes to OCS-CI.

In local experiments, deliberately restarting the noobaa-core pod before an S3 operation showed that the extended retry period was enough for the pod to recover and the operation to succeed. This should significantly reduce test flakiness caused by such transient errors.

### How

1. Update the creation of MCG::s3_resource to use a custom boto3 configuration with `max_attempts=8`.
2. Update the `craft_s3_command` helper to set AWS_MAX_ATTEMPTS={max_attempts} (defaulting to 8)

Fixes #12924 #8270

### DIFF
--- a/ocs_ci/helpers/helpers.py
+++ b/ocs_ci/helpers/helpers.py
@@ -2309,7 +2309,7 @@ def craft_s3_command(cmd, mcg_obj=None, api=False, max_attempts=8):
             f'sh -c "AWS_CA_BUNDLE={constants.SERVICE_CA_CRT_AWSCLI_PATH} '
             f"AWS_ACCESS_KEY_ID={mcg_obj.access_key_id} "
             f"AWS_SECRET_ACCESS_KEY={mcg_obj.access_key} "
-            f"AWS_MAX_ATTEMPTS={max_attempts}"
+            f"AWS_MAX_ATTEMPTS={max_attempts} "
             f"AWS_DEFAULT_REGION={mcg_obj.region} "
             f"aws s3{api} "
             f"--endpoint={mcg_obj.s3_internal_endpoint} "


keesturam (approved):
LGTM

shivamdurgbuns (dismissed):

sagihirshfeld:
[This verification run](https://url.corp.redhat.com/ec1d19a) revealed that I forgot to add a space in the AWSCLI command, and that I had to rebase off of master to pass a few of the tests.

Running the verification again now that I fixed the issue in commit https://github.com/red-hat-storage/ocs-ci/pull/13017/commits/b3cc696ecce9286cea2799b222fc4e891721100d and rebased off of master.

sagihirshfeld:
For further verification, I ran ocs-ci with -m `tier1 and mcg` twice: one with the change which is proposed in the PR, and one without. The failed test cases in both runs seem to be related to other issues we need to investigate:

- The run via master: https://url.corp.redhat.com/28d0b43
- The run via `pr/13017`: https://url.corp.redhat.com/442dffc

It's worth noting that one test ran 5 hours longer via pr/13017, but it seems to be due to an unrelated bug with the deletion of pv-pool backingstores: https://issues.redhat.com/browse/DFBUGS-3885

keesturam (approved):

PR #13045, opened 2025-08-26
Fusion and FDF CLI version args

clacroix12:


clacroix12:
Verified locally that the params are being read in properly and the correct config files are being loaded when they exist.

Some examples with extra output for verification:
```
❯ deploy-fusion --cluster-name cl-3637 --cluster-path ~/clusters/cl-3637 --conf conf/deployment/aws/ipi_3az_rhcos_3m_3w.yaml --fusion-version 2.8
fusion version: 2.8
file loaded: /home/clacroix/workspace/ocs-ci/ocs_ci/framework/conf/fusion_version/fusion-2.8.yaml
```
```
❯ deploy-fdf --cluster-name cl-3637 --cluster-path ~/clusters/cl-3637 --conf conf/deployment/aws/ipi_3az_rhcos_3m_3w.yaml --fdf-version 4.18
fdf version: 4.18
file loaded: /home/clacroix/workspace/ocs-ci/ocs_ci/framework/conf/fdf_version/fdf-4.18.yaml
```

Essentially this is doing exactly what we do for OCP/OCS version params where we take the provided value and find the corresponding config file for that version.

ocs_ci/utility/framework/fusion_fdf_init.py
```diff
@@ -85,6 +85,18 @@ def init_config(self, args: list) -> None:
 
         config.REPORTING["report_path"] = args.report
 
+        if self.deployment_type == "fusion" and args.fusion_version:
+            base_dir = os.path.join(constants.FRAMEWORK_CONF_DIR, "fusion_version")
+            version_file = f"fusion-{args.fusion_version}.yaml"
```
rh-aaleem, line 90: Rather than changing the file , can we do in this way 
- we will create a common file lets say fdf.yaml 
- and will read the common file and update the version in that yaml which is mentioned
- now will use the current created file for fusion or fdf deployment,

clacroix12: I thought about consolidating this but I thought it best to keep it handled the same way we are handling ocp-version and ocs-version params. Currently not much changes between the config files (for FDF) other than the image tag but that might not be the case moving forward.

This also helps us validate whether or not the version is supported since we fail basically right away if the file doesn't exist rather than failing when we get around to pulling the image.

rh-aaleem: for zstream validation the version changes frequently and for each version change we need to push the changes and need to run the pipeline using that branch, which is looking a bit cumbersome,

petr-balogh: I am on the call with Abdul now, and what he is suggesting is to have `fdf_image_tag` easily configurable by parameter in jenkins job - which also maybe worth to have CLI argument for the entrypoint like --fdf-image-tage where we can pass the very specific version like `v4.18.6-5`. 
So in jenkins if we have separate param: `FDF_IMAGE_TAG` it might be helpful instead of overwriting those in YAML_TEXT_CONFIG.

Also not sure if order of loading config files passed via `--ocsci-conf` and via new version parameter loads first and then we load those config files which will allow us to overwrite those values. I hope yes.

clacroix12: After discussing with Petr we have decided on implementing two other CLI args that will be used to overwrite the image tag for Fusion/FDF version. The args implemented here will remain as they compliment how we are handling things for OCP/OCS versions. The new params will compliment how we handle OCP_INSTALLER_VERSION and OCS_REGISTRY_IMAGE. These will be added in a follow up PR.

clacroix12: PR raised to add image tag args: https://github.com/red-hat-storage/ocs-ci/pull/13247

PR #13048, opened 2025-08-28
Adds tests for noobaa route authentication feature

udaysk23:
Adds tests for noobaa route authentication feature

### DIFF
--- a/tests/functional/object/mcg/test_internal_route_auth.py
+++ b/tests/functional/object/mcg/test_internal_route_auth.py
@@ -1,10 +1,19 @@
 import logging
 import requests
+import pytest
 
+from time import sleep
 from ocs_ci.ocs import constants
-from ocs_ci.helpers.helpers import get_noobaa_metrics_token_from_secret
-from ocs_ci.ocs.resources.pod import get_noobaa_endpoint_pods
+from ocs_ci.utility.utils import TimeoutSampler
 from ocs_ci.ocs.ocp import OCP
+from ocs_ci.framework import config
+from ocs_ci.helpers.helpers import get_noobaa_metrics_token_from_secret
+from ocs_ci.ocs.resources.pod import (
+    get_noobaa_endpoint_pods,
+    get_pod_logs,
+    wait_for_pods_to_be_running,
+    get_noobaa_core_pod,
+)
 from ocs_ci.framework.pytest_customization.marks import (
     tier2,
     red_squad,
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_internal_route_auth.py
```diff
@@ -15,13 +22,84 @@
...
+        """
+        Changes flag value to false for testing purpose
+        """
+        noobaa_configmap = OCP(
+            kind="configmap",
+            namespace=constants.OPENSHIFT_STORAGE_NAMESPACE,
+            resource_name=constants.NOOBAA_CONFIGMAP,
+        )
+        params = f'{{"data": {{"{flag_param}": "false"}}}}'
+        noobaa_configmap.patch(params=params, format_type="merge")
```
sagihirshfeld: 1. Please consider making this more flexible. It'll make the usage self explanatory (no need for the comment like the one in line 234), and you'll be able to use it to set other values for other test cases.
2. Use the `config.ENV_DATA["cluster_namespace"]` instead of `constants.OPENSHIFT_STORAGE_NAMESPACE`. IIRC using the existing one will fail the test when tested by teams outside of ODF-QE.

```suggestion
def set_nb_config_flag(self, flag_param, value):
    """
    Sets the value for a given config flag in noobaa-config.

    Args:
        flag_param (str): The config flag key to modify.
        value (str): Desired value for the flag.
    """
    noobaa_configmap = OCP(
        kind="configmap",
        namespace=config.ENV_DATA["cluster_namespace"],
        resource_name=constants.NOOBAA_CONFIGMAP,
    )
    params = f'{{"data": {{"{flag_param}": "{value}"}}}}'
    noobaa_configmap.patch(params=params, format_type="merge")
```

tests/functional/object/mcg/test_internal_route_auth.py
```diff
@@ -15,13 +22,84 @@
...
+        ).get()
+        return f"https://{s3_route_data['spec']['host']}/version"
+
+    def acccess_url(self, url, header=None, verify=False):
+        """
+        Executes curl request to the given url with or without header and returns status code
+        """
+        response = requests.get(url, headers=header, verify=verify)
+        logger.info(response)
+        return response.status_code
```
sagihirshfeld: 1. Note the typo in the function name: 3 c's in access instead of 2.
2. Please use this more detailed docstring to clarify usage and return value. I was specifically interested in whether the returned code is a string of an int because of the way this function is used in line 247.
```suggestion
def access_url(self, url, header=None, verify=False):
    """
    Executes curl request to the given url with or without header.

    Args:
        url (str): The target URL to request
        header (dict, optional): A dictionary of HTTP headers to include in the request.
            Defaults to None.
        verify (bool, optional): Whether to verify the server's TLS/SSL certificate.
            Defaults to False.

    Returns:
        int(?): The HTTP status code from the server response.
    """
    response = requests.get(url, headers=header, verify=verify)
    logger.info(response)
    return response.status_code
```

tests/functional/object/mcg/test_internal_route_auth.py
```diff
@@ -62,79 +140,137 @@ def test_mgmt_and_endpoint_route_authentication(self):
...
+            cm_data.get(flag) == "true"
+        ), f"Incorrect default value of {flag}. Received value {cm_data.get(flag)}"
+
+        # Validate respective routes using auth token
+        response_code = self.acccess_url(test_url, header=auth_headers, verify=False)
+        assert (
+            response_code == 200
+        ), f"Expected response code 200. Received {response_code}"
+
+        # Change flag value to "False"
```
sagihirshfeld, line 249: You can remove this comment if you choose to make the changes I suggested for the `change_config_flag` function

tests/functional/object/mcg/test_internal_route_auth.py
```diff
@@ -62,79 +140,137 @@ def test_mgmt_and_endpoint_route_authentication(self):
...
+        self.change_config_flag(flag)
+        import time
```
sagihirshfeld: Please move this import to the top of the file if possible

tests/functional/object/mcg/test_internal_route_auth.py
```diff
@@ -62,79 +140,137 @@ def test_mgmt_and_endpoint_route_authentication(self):
...
+
+        time.sleep(60)
```
sagihirshfeld: Can't we remove this hardcoded sleep period since you're using the TimeoutSampler anyway to wait for the desired state?

udaysk23: Thanks for point this, This code is not part of this test. I used it to manually validate the configMap

tests/functional/object/mcg/test_internal_route_auth.py
udaysk23: This line is from steps which I mentioned in doc string in this test function.

tests/functional/object/mcg/test_internal_route_auth.py
udaysk23: Fixed

tests/functional/object/mcg/test_internal_route_auth.py
udaysk23: Fixed

tests/functional/object/mcg/test_internal_route_auth.py
udaysk23: Fixed

sagihirshfeld (approved):
LGTM

sagihirshfeld (approved):

PR #13070, opened 2025-09-01
remove adm-d mirrors for metallb installation

DanielOsypenko:
With this PR we cover another case, when optional-operators catsrc exists but we still want to apply idms correction.
This function need to be called earlier, to make metallb stop pulling images from adm-d image which is propagated during  ACM/MCE deployment.

correct conflicting idms redirection should happen earlier, even in case if CatalogSource already exist.


DanielOsypenko:
verified manually on broken cluster, works as before, when catsrc did not exist

suchita-g (approved):

dahorak (approved):

PR #13085, opened 2025-09-02
Adding tests for RHSTOR-7208

shivamdurgbuns:
Feature link: https://issues.redhat.com/browse/RHSTOR-7208
QE Link: https://issues.redhat.com/browse/OCSQE-3523
Tests:
OCS-6916
OCS-6917
OCS-6918

### DIFF
--- a/tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
+++ b/tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
@@ -1,28 +1,26 @@
 import logging
-import base64
 import re
 import time
 
-import pytest
 
 from ocs_ci.framework.pytest_customization.marks import (
     mcg,
     polarion_id,
     red_squad,
     runs_on_provider,
-    tier1,
+    tier2,
 )
 from ocs_ci.framework.testlib import MCGTest
-from ocs_ci.ocs.ocp import OCP
-from ocs_ci.ocs import constants
+from ocs_ci.framework import config
 from ocs_ci.ocs.bucket_utils import (
     update_replication_policy,
     wait_for_object_versions_match,
     copy_random_individual_objects,
 )
+from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs.resources.mcg_replication_policy import McgReplicationPolicy
-from ocs_ci.utility.utils import run_cmd
 from ocs_ci.ocs.resources.pod import get_noobaa_core_pod
+from ocs_ci.helpers.helpers import get_noobaa_metrics_token_from_secret
 
 logger = logging.getLogger(__name__)
 
@@ -35,56 +33,17 @@ TIMEOUT = 300
 @runs_on_provider
 class TestReplicationAndItsMetrics(MCGTest):
     """
-    Test suite for MCG object replication policies
+    Test suite for MCG object replication policies and its metrics
+    Here we will be testing newly added metrices to expose replication per bucket:
+        bucket_last_cycle_total_objects_num,
+        bucket_last_cycle_replicated_objects_num,
+        bucket_last_cycle_error_objects_num
     """
 
-    @pytest.fixture(autouse=True, scope="class")
-    def reduce_replication_delay_setup(self, add_env_vars_to_noobaa_core_class):
-        """
-        Reduce the replication delay to one minute
-
-        Args:
-            new_delay_in_miliseconds (function): A function to add env vars to the noobaa-core pod
-        """
-        new_delay_in_miliseconds = 60 * 1000
-        new_env_var_tuples = [
-            (constants.BUCKET_REPLICATOR_DELAY_PARAM, new_delay_in_miliseconds),
-            (constants.BUCKET_LOG_REPLICATOR_DELAY_PARAM, new_delay_in_miliseconds),
-        ]
-        add_env_vars_to_noobaa_core_class(new_env_var_tuples)
-
-    @pytest.fixture()
-    def make_buckets(self, bucket_factory):
-        """
-        A factory that creates MCG buckets
-
-        Args:
-            bucket_factory: Fixture for creating new buckets
-
-        Returns:
-            function: The factory function
-        """
-
-        def _factory(amount):
-            """
-            Create buckets
-
-            Args:
-                amount (int): The number of buckets to create
-
-            Returns:
-                list(Bucket): The created buckets
-            """
-            # Using the OC interface allows patching a replication policy on the OBC
-            buckets = bucket_factory(amount, "OC")
-            return buckets
-
-        return _factory
-
-    @tier1
+    @tier2
     @polarion_id("OCS-6916")
     def test_bucket_replication_and_its_metrics(
-        self, awscli_pod, mcg_obj, make_buckets, test_directory_setup
+        self, awscli_pod, mcg_obj, test_directory_setup, bucket_factory
     ):
         """
         1. Create source and target buckets.
@@ -95,7 +54,7 @@ class TestReplicationAndItsMetrics(MCGTest):
         6. Verify the metrics
         """
         # 1. Create source and target buckets
-        bucket_1, bucket_2 = make_buckets(2)
+        bucket_1, bucket_2 = bucket_factory(2, "OC")
 
         # 2. Set bucket replication policies on the source buckets
         replication_policy = McgReplicationPolicy(destination_bucket=bucket_2.name)
@@ -123,14 +82,13 @@ class TestReplicationAndItsMetrics(MCGTest):
         logger.info("All the versions were replicated successfully")
 
         # 5. get JWT TOKEN
-        token_cmd = "oc get secret noobaa-metrics-auth-secret -n openshift-storage -o jsonpath='{.data.metrics_token}'"
-        jwt_token = base64.b64decode(run_cmd(cmd=token_cmd)).decode()
-        noobaa_core_pod = get_noobaa_core_pod()
+        jwt_token = get_noobaa_metrics_token_from_secret()
 
         # 6. Get the metrics for the bucket replication
+        noobaa_core_pod = get_noobaa_core_pod()
+        metrics_cmd = f"curl -k -H 'Authorization: Bearer {jwt_token}' localhost:8080/metrics/bg_workers"
         metrics_output = noobaa_core_pod.exec_cmd_on_pod(
-            f"curl -k -H 'Authorization: Bearer {jwt_token}' localhost:8080/metrics/bg_workers",
-            out_yaml_format=False,
+            metrics_cmd, out_yaml_format=False
         )
 
         # 7. Verify the metrics
@@ -140,17 +98,17 @@ class TestReplicationAndItsMetrics(MCGTest):
             "NooBaa_bucket_last_cycle_error_objects_num": 0,
         }
         for metric, expected_value in expected_metrics.items():
-            pattern = rf'^{metric}{{bucket_name="{bucket_1.name}"}} (\d)$'
+            pattern = rf"^{metric}{{bucket_name=\"{bucket_1.name}\"}} (\d)$"
             output_value = re.findall(pattern, metrics_output, re.MULTILINE)[-1]
             logger.info(f"Metrics {metric} : {output_value}")
             assert (
                 int(output_value) == expected_value
             ), f"Metric {metric} has unexpected value"
 
-    @tier1
+    @tier2
     @polarion_id("OCS-6917")
     def test_failing_bucket_replication_and_its_metrics(
-        self, awscli_pod, mcg_obj, make_buckets, test_directory_setup
+        self, awscli_pod, mcg_obj, test_directory_setup, bucket_factory
     ):
         """
         1. Create source and target buckets.
@@ -160,14 +118,15 @@ class TestReplicationAndItsMetrics(MCGTest):
         5. Verify the objects were replicated to the target buckets
         6. Wait before writing to the source bucket so that the replication
             policy will be applied and quota will get exceeded
-        7. Write an object to the source bucket so that the quota will get exceeded
-        8. Wait before getting the metrics so that the replication policy
-            will be applied and quota will get exceeded
+        7. Write an object to the source bucket so that the quota will get
+            exceeded
+        8. Wait before getting the metrics so that the replication
+        policy will be applied and quota will get exceeded
         9. Get the metrics for the bucket replication
         10. Verify the metrics
         """
         # 1. Create source and target buckets
-        bucket_1, bucket_2 = make_buckets(2)
+        bucket_1, bucket_2 = bucket_factory(2, "OC")
 
         # 2. Set bucket replication policies on the source buckets
         replication_policy = McgReplicationPolicy(destination_bucket=bucket_2.name)
@@ -175,7 +134,7 @@ class TestReplicationAndItsMetrics(MCGTest):
 
         # 3. Patch the target bucket to change the quota
         quota_str = '{"spec": {"additionalConfig":{"maxObjects": "4"}}}'
-        cmd = f"patch obc {bucket_2.name} -p '{quota_str}' -n openshift-storage --type=merge"
+        cmd = f"patch obc {bucket_2.name} -p '{quota_str}' -n {config.ENV_DATA['cluster_namespace']} --type=merge"
         OCP().exec_oc_cmd(cmd)
         logger.info(f"Patched quota to obc {bucket_2.name}")
         time.sleep(60)
@@ -203,8 +162,8 @@ class TestReplicationAndItsMetrics(MCGTest):
 
         time.sleep(120)
         logger.info(
-            "Waiting before writing to the source bucket so that the replication \
-            policy will be applied and quota will get exceeded"
+            "Waiting before writing to the source bucket so that the \
+            replication policy will be applied and quota will get exceeded"
         )
 
         # 6. Write an object to the source bucket so that the quota will get exceeded
@@ -219,19 +178,18 @@ class TestReplicationAndItsMetrics(MCGTest):
 
         time.sleep(120)
         logger.info(
-            "Waiting before getting the metrics so that the replication \
-            policy will be applied and quota will get exceeded"
+            "Waiting before getting the metrics so that the replication policy \
+            will be applied and quota will get exceeded"
         )
 
         # 7. get JWT TOKEN
-        token_cmd = "oc get secret noobaa-metrics-auth-secret -n openshift-storage -o jsonpath='{.data.metrics_token}'"
-        jwt_token = base64.b64decode(run_cmd(cmd=token_cmd)).decode()
-        noobaa_core_pod = get_noobaa_core_pod()
+        jwt_token = get_noobaa_metrics_token_from_secret()
 
         # 8. Get the metrics for the bucket replication
+        noobaa_core_pod = get_noobaa_core_pod()
+        metrics_cmd = f"curl -k -H 'Authorization: Bearer {jwt_token}' localhost:8080/metrics/bg_workers"
         metrics_output = noobaa_core_pod.exec_cmd_on_pod(
-            f"curl -k -H 'Authorization: Bearer {jwt_token}' localhost:8080/metrics/bg_workers",
-            out_yaml_format=False,
+            metrics_cmd, out_yaml_format=False
         )
 
         # 9. Verify the metrics, will get on error object in the last cycle
--- a/tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
+++ b/tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
@@ -1,6 +1,5 @@
 import logging
 import time
-import json
 import pytest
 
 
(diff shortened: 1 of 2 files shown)


tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,249 @@
...
+
+TIMEOUT = 300
+
+
+@mcg
+@red_squad
+@runs_on_provider
+class TestReplicationAndItsMetrics(MCGTest):
+    """
+    Test suite for MCG object replication policies
```
udaysk23: Doc string needs to be more clear about what script tests

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,249 @@
...
+
+        Args:
+            new_delay_in_miliseconds (function): A function to add env vars to the noobaa-core pod
+        """
+        new_delay_in_miliseconds = 60 * 1000
+        new_env_var_tuples = [
+            (constants.BUCKET_REPLICATOR_DELAY_PARAM, new_delay_in_miliseconds),
+            (constants.BUCKET_LOG_REPLICATOR_DELAY_PARAM, new_delay_in_miliseconds),
+        ]
+        add_env_vars_to_noobaa_core_class(new_env_var_tuples)
```
udaysk23: Please use existing fixture directly in test method
```suggestion
    add_env_vars_to_noobaa_core_class(
            [
                (constants.BUCKET_REPLICATOR_DELAY_PARAM, new_delay_in_miliseconds),
                (constants.BUCKET_LOG_REPLICATOR_DELAY_PARAM, new_delay_in_miliseconds),
            ]
        )
```

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,249 @@
...
+                amount (int): The number of buckets to create
+
+            Returns:
+                list(Bucket): The created buckets
+            """
+            # Using the OC interface allows patching a replication policy on the OBC
+            buckets = bucket_factory(amount, "OC")
+            return buckets
+
+        return _factory
```
udaysk23: Please use existing bucket factory fixture directly in test method

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,249 @@
...
+
+    @tier1
```
udaysk23: Change it to tier2

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,249 @@
...
+        """
+        1. Create source and target buckets.
+        2. Set bucket replication policies on the source buckets
+        3. Write some objects to the source buckets
+        4. Verify the objects were replicated to the target buckets
+        5. Get the metrics for the bucket replication
+        6. Verify the metrics
+        """
+        # 1. Create source and target buckets
+        bucket_1, bucket_2 = make_buckets(2)
```
udaysk23: use bucket factory directly here

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,249 @@
...
+            awscli_pod,
+            bucket_1.name,
+            bucket_2.name,
+            obj_key="test_obj-",
+        )
+        logger.info("All the versions were replicated successfully")
+
+        # 5. get JWT TOKEN
+        token_cmd = "oc get secret noobaa-metrics-auth-secret -n openshift-storage -o jsonpath='{.data.metrics_token}'"
+        jwt_token = base64.b64decode(run_cmd(cmd=token_cmd)).decode()
```
udaysk23: Please use `get_noobaa_metrics_token_from_secret` method from helper class

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,249 @@
...
+        noobaa_core_pod = get_noobaa_core_pod()
+
+        # 6. Get the metrics for the bucket replication
+        metrics_output = noobaa_core_pod.exec_cmd_on_pod(
+            f"curl -k -H 'Authorization: Bearer {jwt_token}' localhost:8080/metrics/bg_workers",
+            out_yaml_format=False,
+        )
```
udaysk23: Can we achieve it using `response` library?

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,249 @@
...
+        }
+        for metric, expected_value in expected_metrics.items():
+            pattern = rf'^{metric}{{bucket_name="{bucket_1.name}"}} (\d)$'
+            output_value = re.findall(pattern, metrics_output, re.MULTILINE)[-1]
+            logger.info(f"Metrics {metric} : {output_value}")
+            assert (
+                int(output_value) == expected_value
+            ), f"Metric {metric} has unexpected value"
+
+    @tier1
```
udaysk23: Please change it to tier2

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,249 @@
...
+        6. Wait before writing to the source bucket so that the replication
+            policy will be applied and quota will get exceeded
+        7. Write an object to the source bucket so that the quota will get exceeded
+        8. Wait before getting the metrics so that the replication policy
+            will be applied and quota will get exceeded
+        9. Get the metrics for the bucket replication
+        10. Verify the metrics
+        """
+        # 1. Create source and target buckets
+        bucket_1, bucket_2 = make_buckets(2)
```
udaysk23: Use bucket factory here as well

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,249 @@
...
+
+        # 2. Set bucket replication policies on the source buckets
+        replication_policy = McgReplicationPolicy(destination_bucket=bucket_2.name)
+        update_replication_policy(bucket_1.name, replication_policy.to_dict())
+
+        # 3. Patch the target bucket to change the quota
+        quota_str = '{"spec": {"additionalConfig":{"maxObjects": "4"}}}'
+        cmd = f"patch obc {bucket_2.name} -p '{quota_str}' -n openshift-storage --type=merge"
```
udaysk23: Use config.env to access namespace

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,249 @@
...
+
+        time.sleep(120)
+        logger.info(
+            "Waiting before getting the metrics so that the replication \
+            policy will be applied and quota will get exceeded"
+        )
+
+        # 7. get JWT TOKEN
+        token_cmd = "oc get secret noobaa-metrics-auth-secret -n openshift-storage -o jsonpath='{.data.metrics_token}'"
+        jwt_token = base64.b64decode(run_cmd(cmd=token_cmd)).decode()
```
udaysk23: Please use `get_noobaa_metrics_token_from_secret` method from helper class

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
shivamdurgbuns: as we need to fetch the data inside the noobaa_core pod, introducing response library here will just add to extra steps in the test, so i think we can skip that library here.

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,207 @@
...
+
+        # 4. Verify the objects were replicated to the target buckets
+        wait_for_object_versions_match(
+            mcg_obj,
+            awscli_pod,
+            bucket_1.name,
+            bucket_2.name,
+            obj_key="test_obj-",
+        )
+        logger.info("All the versions were replicated successfully")
```
sagihirshfeld: Nice use of the `wait_for_object_versions_match` function. I didn't even consider it can be used to check standard replication as well, even when versioning is disabled!

However, the proper term here is "objects" instead of "versions". The latter implies versioning is enabled or related when it's not the case.
```suggestion
        logger.info("All the objects were replicated successfully")
```

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,207 @@
...
+
+        # 5. get JWT TOKEN
+        jwt_token = get_noobaa_metrics_token_from_secret()
+
+        # 6. Get the metrics for the bucket replication
+        noobaa_core_pod = get_noobaa_core_pod()
+        metrics_cmd = f"curl -k -H 'Authorization: Bearer {jwt_token}' localhost:8080/metrics/bg_workers"
+        metrics_output = noobaa_core_pod.exec_cmd_on_pod(
+            metrics_cmd, out_yaml_format=False
+        )
```
sagihirshfeld: Since these metrics are available in Prometheus, it would be better to use get_prometheus_response calls here. That way the test covers the complete end-to-end flow.

At least via bash, the query goes like this:
```
$ TOKEN=$(oc whoami --show-token)
$ ROUTE=$(oc -n openshift-monitoring get route thanos-querier -o jsonpath='{.spec.host}')
$ curl -ks -H "Authorization: Bearer $TOKEN" \
  "https://$ROUTE/api/v1/query?query=NooBaa_bucket_last_cycle_total_objects_num" | jq -r ".data.result[0].value[-1]"
5
```

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,207 @@
...
+            "NooBaa_bucket_last_cycle_replicated_objects_num": 5,
+            "NooBaa_bucket_last_cycle_error_objects_num": 0,
+        }
+        for metric, expected_value in expected_metrics.items():
+            pattern = rf"^{metric}{{bucket_name=\"{bucket_1.name}\"}} (\d)$"
+            output_value = re.findall(pattern, metrics_output, re.MULTILINE)[-1]
+            logger.info(f"Metrics {metric} : {output_value}")
+            assert (
+                int(output_value) == expected_value
+            ), f"Metric {metric} has unexpected value"
```
sagihirshfeld: If you check prometheus instead of the local noobaa-core endpoint, you'll be able to simply check the response dictionaries (which you'll get from using json.loads() on the response) without having to use regex.

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,207 @@
...
+        # 2. Set bucket replication policies on the source buckets
+        replication_policy = McgReplicationPolicy(destination_bucket=bucket_2.name)
+        update_replication_policy(bucket_1.name, replication_policy.to_dict())
+
+        # 3. Patch the target bucket to change the quota
+        quota_str = '{"spec": {"additionalConfig":{"maxObjects": "4"}}}'
+        cmd = f"patch obc {bucket_2.name} -p '{quota_str}' -n {config.ENV_DATA['cluster_namespace']} --type=merge"
+        OCP().exec_oc_cmd(cmd)
+        logger.info(f"Patched quota to obc {bucket_2.name}")
+        time.sleep(60)
```
sagihirshfeld, line 141:
```suggestion
        # Wait a bit for the quota to take effect
        time.sleep(60)
```

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,207 @@
...
+            bucket_2.name,
+            obj_key="test_obj-",
+        )
+        logger.info("All the versions were replicated successfully")
+
+        time.sleep(120)
+        logger.info(
+            "Waiting before writing to the source bucket so that the \
+            replication policy will be applied and quota will get exceeded"
+        )
```
sagihirshfeld:
```suggestion
        logger.info(
            "Waiting for the quota limit to take effect on the target bucket"
            "now that its maxObjects has been reached after the replication"
        )
        time.sleep(120)
```

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,207 @@
...
+            pattern="test_obj-over-quota-",
+            s3_obj=mcg_obj,
+            amount=1,
+        )
+
+        time.sleep(120)
+        logger.info(
+            "Waiting before getting the metrics so that the replication policy \
+            will be applied and quota will get exceeded"
+        )
```
sagihirshfeld: Waiting 120 seconds might not be enough here. Even with the added `reduce_replication_delay_setup` (see suggestion below this one) fixture it might be too close.

 It will be safer to wait for the metrics via a TimeoutSampler.

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,207 @@
...
+@runs_on_provider
+class TestReplicationAndItsMetrics(MCGTest):
+    """
+    Test suite for MCG object replication policies and its metrics
+    Here we will be testing newly added metrices to expose replication per bucket:
+        bucket_last_cycle_total_objects_num,
+        bucket_last_cycle_replicated_objects_num,
+        bucket_last_cycle_error_objects_num
+    """
+
```
sagihirshfeld, line 39: Since these tests are waiting on replications, you can reduce their runtimes by reducing the interval in which noobaa looks for an operates on needed replications. You can do so by adding this type of fixture, like I used in `TestLogBasedBucketReplication`:
```suggestion
    @pytest.fixture(scope="class", autouse=True)
    def reduce_replication_delay_setup(self, add_env_vars_to_noobaa_core_class):
        """
        A fixture to reduce the replication delay to one minute.

        Args:
            new_delay_in_milliseconds (function): A function to add env vars to the noobaa-core pod

        """
        new_delay_in_milliseconds = 60 * 1000
        new_env_var_tuples = [
            (constants.BUCKET_REPLICATOR_DELAY_PARAM, new_delay_in_milliseconds),
        ]
        add_env_vars_to_noobaa_core_class(new_env_var_tuples)
```

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,207 @@
...
+        time.sleep(120)
+        logger.info(
+            "Waiting before getting the metrics so that the replication policy \
+            will be applied and quota will get exceeded"
+        )
+
+        # 7. get JWT TOKEN
+        jwt_token = get_noobaa_metrics_token_from_secret()
+
+        # 8. Get the metrics for the bucket replication
```
sagihirshfeld: Please consider querying Prometheus instead, like I suggested for the previous test.

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,208 @@
...
+        bucket_1, bucket_2 = bucket_factory(2, "OC")
+
+        # 2. Set bucket replication policies on the source buckets
+        replication_policy = McgReplicationPolicy(destination_bucket=bucket_2.name)
+        update_replication_policy(bucket_1.name, replication_policy.to_dict())
+
+        # 3. Patch the target bucket to change the quota
+        quota_str = '{"spec": {"additionalConfig":{"maxObjects": "4"}}}'
+        cmd = f"patch obc {bucket_2.name} -p '{quota_str}' -n {config.ENV_DATA['cluster_namespace']} --type=merge"
+        OCP().exec_oc_cmd(cmd)
```
udaysk23: I would request you to use OCP object to access OBC and then we can perform `patch` operation
```suggestion
        obc_obj = OCP(kind="obc", namespace=namespace, resource_name=name)
```

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,208 @@
...
+            file_dir=test_dir,
+            pattern="test_obj-over-quota-",
+            s3_obj=mcg_obj,
+            amount=1,
+        )
+
+        logger.info(
+            "Waiting for objects to get replicated and new metrics to be exposed"
+        )
+        time.sleep(120)
```
udaysk23, line 182: I can see script will wait for 5 mins apart from execution. Can we use `TimeoutSampler` class here to avoid defined 5 mins waiting time?

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,208 @@
...
+)
+from ocs_ci.ocs.ocp import OCP
+from ocs_ci.ocs.resources.mcg_replication_policy import McgReplicationPolicy
+from ocs_ci.utility.prometheus import PrometheusAPI
+from ocs_ci.ocs import constants
+
+logger = logging.getLogger(__name__)
+
+
+TIMEOUT = 300
```
udaysk23: I dont see script is using this value anywhere. Please remove it if not needed

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
shivamdurgbuns: AFAIU function of TimeoutSampler is to check for a function to achieve desired state under given timeout with sleep in between. But the case here is that 
1.  Reason for the first sleep is if the bucket has certain quota assigned to it and if add objects quickly the bucket takes up those object without raising any error, it seems we have to wait for sometime before uploading new objects so the noobaa gets to check on the number of objects, so if we upload the data after some time the objects fails to upload with error maxobjects achieved (as expected). So If I use a TimeoutSampler here it will continue to upload the objects if I add small sleep time. Large sleep time will make it behave the same way as it is now.
2.  Reason for the second sleep is that the metrics that are getting generated upon replication take some time to update, so if do multiple replication cycle the metrics take some time to get generated and reflect on Prometheus, it take between 1-2 minutes for the metrics to get updated. And adding a Timeout Sampler will behave the same and wait for our desired metrics with sleep in between.

This is what I feel, I can try to optimise the test case more and try to reduce the timing of the test, but TimeoutSampler is not suitable for this situation here. Please correct me if I am wrong. 😄

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
shivamdurgbuns: yep, I forgot to remove that, will update it

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
shivamdurgbuns: thank you, will update it.

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,292 @@
...
+    def test_bucket_replication_and_its_metrics(
+        self, awscli_pod, mcg_obj, test_directory_setup, bucket_factory, threading_lock
+    ):
+        """
+        1. Create source and target buckets.
+        2. Set bucket replication policies on the source buckets
+        3. Write some objects to the source buckets
+        4. Verify the objects were replicated to the target buckets
+        5. Get the metrics for the bucket replication
+        6. Verify the metrics from Prometheus
```
sagihirshfeld:
```suggestion
        5. Verify the bucket replication metrics from Prometheus
```

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,292 @@
...
+        wait_for_object_versions_match(
+            mcg_obj,
+            awscli_pod,
+            bucket_1.name,
+            bucket_2.name,
+            obj_key="test_obj-",
+        )
+        logger.info("All the objects were replicated successfully")
+
+        # # 7. Verify the metrics
```
sagihirshfeld:
```suggestion
        # 5. Verify the bucket replication metrics from Prometheus
```

ypersky1980 (approved):
LGTM

tests/functional/object/mcg/test_bucket_replication_and_its_metrics.py
```diff
@@ -0,0 +1,292 @@
...
+                if resp.ok:
+                    logger.debug(query)
+                    metrics_output = json.loads(resp.text)
+                    got_metrics_value = int(
+                        metrics_output["data"]["result"][0]["value"][1]
+                    )
+                    logger.info(f"Metrics {metric} : {got_metrics_value}")
+                    assert (
+                        got_metrics_value == expected_value
+                    ), f"Metric {metric} has unexpected value"
```
udaysk23: I can see this code of block repeated in all tests, It would be great if we create one separate method in the same class and call it wherever needed to avoid repeated block for future tests as well

udaysk23 (approved):

sagihirshfeld (approved):
LGTM

shivamdurgbuns:
Verification Job: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/56848/
Test: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/56848/testReport/

keesturam (approved):

PR #13087, opened 2025-09-02
deploy-w/o-kubeconfig-fix-temp

DanielOsypenko:
this is a temp fix until we find way to replace call to `setup_multicluster_marker(pytest.mark.run_on_all_clients, True) `
in ocs_ci/framework/pytest_customization/marks.py module

### DIFF
--- a/ocs_ci/deployment/helpers/hypershift_base.py
+++ b/ocs_ci/deployment/helpers/hypershift_base.py
@@ -35,7 +35,6 @@ Main tasks include:
 
 """
 
-
 logger = logging.getLogger(__name__)
 
 
@@ -76,7 +75,6 @@ def get_available_hosted_clusters_to_ocp_ver_dict():
         '"\\(.metadata.name)|\\(.status.version.history[0].version)"\''
     )
     with config.RunWithProviderConfigContextIfAvailable():
-
         out = OCP().exec_oc_cmd(
             command=cmd,
             cluster_config=config,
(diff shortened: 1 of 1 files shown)


ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -375,7 +373,26 @@ class HyperShiftBase:
...
+        # When the kubeconfig file does NOT exist AND console credentials ARE provided - do not run test_cluster()
+        # If we have password or ocp url we might want to log in to cluster and generate kubeconfig,
+        # which is done very early with process_cluster_cli_params; follow pr/11749
+        if not os.path.isfile(
+            os.path.join(
+                config.ENV_DATA["cluster_path"],
+                config.RUN.get("kubeconfig_location"),
+            )
+        ):
+            logger.warning("kubeconfig path is not set or file does not exist")
```
dahorak, line 386: Is this warning log necessary? I'm thinking if it will not create unnecessary noise.

DanielOsypenko: this warn msg must log only once, when process_cluster_cli_params is yet not called AND only on HCI deployments
once we have kubeconfig after it is generated with process_cluster_cli_params we will not get into this block
```
os.path.isfile(
            os.path.join(
                config.ENV_DATA["cluster_path"],
                config.RUN.get("kubeconfig_location"),
            )
        )

```

suchita-g (approved):

PR #10350, opened 2024-08-26
Ignore noobaa-endpoint pods leftovers in MCG tests

sagihirshfeld:
The checking for leftover resources after a test suite(=PyTest test class), comparisons are made between the list of pods in the cluster from before and after the tests . However, some MCG tests might remove original endpoint pods and create new ones due to scaling or mounting of PVCs, thus raising a false ResourceLeftoversException.

This PR aims to ignore these false "discrepancies" by adding `ignore_leftover_label` label to the base `MCGTest` class.

PR #13090, opened 2025-09-02
Adding block  external access - storagecluster test case

ypersky1980:
This PR contains first test case ( adding disableRoutes flag to storagecluster yaml) of https://issues.redhat.com/browse/RHSTOR-7195 ( "Add an option to disable all routes that enables external access"). 

The implemented flow is: 
- verify that disableRoutes flag in storagecluster yaml ither does not exist of equal to False 
- verify that once the flag does not exist or is equal to False, all the routes are recreated automatically once deleted.
- change disableRoutes value to True
- verify that s3/sts/noobaa-mgmt routes are not recreated automatically once deleted, and the other routes are still recreated automatically
- change disableRoutes value to False
- verify that s3/sts/noobaa-mgmt routes are recreated, and the other do exist as well.

### DIFF
--- a/tests/functional/object/mcg/test_block_external_access.py
+++ b/tests/functional/object/mcg/test_block_external_access.py
@@ -1,0 +1,126 @@
+import logging
+import time
+
+import yaml
+
+import pytest
+
+from ocs_ci.framework.testlib import MCGTest
+from ocs_ci.ocs import constants, ocp
+from ocs_ci.framework import config
+
+
+logger = logging.getLogger(__name__)
+
+
+def check_disable_routes(disable_routes_val):
+    """
+    Args:
+        disable_routes_val (bool) Value to be set to 'disableRoutes' parameter
+    """
+    non_deletable_routes = [
+        "ocs-storagecluster-cephobjectstore",
+        "ocs-storagecluster-cephobjectstore-secure",
+    ]
+    deletable_routes = ["noobaa-mgmt", "s3", "sts"]
+    all_routes = non_deletable_routes + deletable_routes
+    timeout = 60
+
+    ocp_obj = ocp.OCP()
+    result = ocp_obj.exec_oc_cmd(command="get routes -o yaml", out_yaml_format=False)
+    data = yaml.safe_load(result)
+    route_names = [item["metadata"]["name"] for item in data["items"]]
+    logger.info(f"Existing routes {route_names}")
+    assert set(all_routes).issubset(
+        route_names
+    ), "Some of the predefined routes don't exist"
+
+    for route in all_routes:
+        ocp_obj.exec_oc_cmd(command=f"delete route {route}")
+
+    time.sleep(timeout)
+
+    result = ocp_obj.exec_oc_cmd(command="get routes -o yaml", out_yaml_format=False)
+    data = yaml.safe_load(result)
+    route_names = [item["metadata"]["name"] for item in data["items"]]
+    logger.info(f"Routes after deletion {route_names}")
+
+    if not disable_routes_val:
+        assert set(all_routes).issubset(
+            set(route_names)
+        ), "Some of the predefined routes don't exist"
+    else:
+        assert set(non_deletable_routes).issubset(
+            set(route_names)
+        ), "Some of the predefined routes were deleted"
+        assert set(deletable_routes).isdisjoint(
+            set(route_names)
+        ), "Some routes were not deleted"
+
+
+def set_disable_routes_value(storagecluster_obj, val):
+    """
+    Args:
+        storagecluster_obj (obj): Storage cluster on which the value should be set
+        val (bool) Value to be set to 'disableRoutes' parameter
+    """
+    if not isinstance(val, bool):
+        raise TypeError("val argument should be boolean")
+
+    param_value = str(val).lower()
+    disable_routes_param = (
+        f'{{"spec": {{"multiCloudGateway": {{"disableRoutes": {param_value} }}}}}}'
+    )
+
+    storagecluster_obj.patch(
+        params=disable_routes_param,
+        format_type="merge",
+    )
+
+    # validate that the configuration really changed
+    sc_dict = storagecluster_obj.get()
+    logger.info(f"Updated configuration is {sc_dict}")
+
+    disable_routes = (
+        sc_dict.get("spec", {}).get("multiCloudGateway", {}).get("disableRoutes")
+    )
+
+    assert (
+        disable_routes == val
+    ), f"Disable routes is expected to be {val}, is {disable_routes}"
+
+
+class TestBlockExternalAccess(MCGTest):
+    def test_block_access_from_storagecluster(
+        self,
+    ):
+        """
+        This method validates that
+         - if disableRoutes flag in storagecluster yaml is set to False then all routes are recreated after deletion
+         - if disableRoutes flag in storagecluster yaml is set to True then noobaa-mgmt, s3 and sts routes are not
+        recreated after deletion
+        """
+        storagecluster_obj = ocp.OCP(
+            resource_name=constants.DEFAULT_CLUSTERNAME,
+            namespace=config.ENV_DATA["cluster_namespace"],
+            kind=constants.STORAGECLUSTER,
+        )
+
+        # validate that initial value of disableRoutes is false or missing (and is false by default)
+        sc_dict = storagecluster_obj.get()
+        logger.info(f"Initial storagecluster configuration is {sc_dict}")
+
+        disable_routes_status = (
+            sc_dict.get("spec", {}).get("multiCloudGateway", {}).get("disableRoutes")
+        )
+        assert (
+            not disable_routes_status or disable_routes_status == False
+        ), "Disable Routes should be not defined or false when test starts"
+
+        check_disable_routes(False)
+
+        set_disable_routes_value(storagecluster_obj, True)
+        check_disable_routes(True)
+
+        set_disable_routes_value(storagecluster_obj, False)
+        check_disable_routes(False)


tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,126 @@
...
+
+    assert (
+        disable_routes == val
+    ), f"Disable routes is expected to be {val}, is {disable_routes}"
+
+
+@red_squad
+@mcg
+class TestBlockExternalAccess(MCGTest):
+    def test_block_access_from_storagecluster(
```
sagihirshfeld, line 131: Please add a tier marker. I guess this is a tier2?
Also polarion_id if possible.

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,126 @@
...
+        # validate that initial value of disableRoutes is false or missing (and is false by default)
+        sc_dict = storagecluster_obj.get()
+        logger.info(f"Initial storagecluster configuration is {sc_dict}")
+
+        disable_routes_status = (
+            sc_dict.get("spec", {}).get("multiCloudGateway", {}).get("disableRoutes")
+        )
+        assert (
+            not disable_routes_status or disable_routes_status is False
+        ), "Disable Routes should be not defined or false when test starts"
```
sagihirshfeld: The second condition is redundant:
```suggestion
    assert not disable_routes_status, (
        "Disable Routes should be not defined or false when test starts"
    )
```

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,126 @@
...
+    timeout = 60
+
+    ocp_obj = ocp.OCP()
+    result = ocp_obj.exec_oc_cmd(command="get routes -o yaml", out_yaml_format=False)
+    data = yaml.safe_load(result)
+    route_names = [item["metadata"]["name"] for item in data["items"]]
+    logger.info(f"Existing routes {route_names}")
+    assert set(all_routes).issubset(
+        route_names
+    ), "Some of the predefined routes don't exist"
```
sagihirshfeld: In most deployments where RGW isn't available, this will fail because the RGW endpoints `ocs-storagecluster-cephobjectstore` and `ocs-storagecluster-cephobjectstore-secure` will be missing from `route_names`

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,126 @@
...
+        "ocs-storagecluster-cephobjectstore-secure",
+    ]
+    deletable_routes = ["noobaa-mgmt", "s3", "sts"]
+    all_routes = non_deletable_routes + deletable_routes
+    timeout = 60
+
+    ocp_obj = ocp.OCP()
+    result = ocp_obj.exec_oc_cmd(command="get routes -o yaml", out_yaml_format=False)
+    data = yaml.safe_load(result)
+    route_names = [item["metadata"]["name"] for item in data["items"]]
```
sagihirshfeld:
```suggestion
    ocp_routes_obj = ocp.OCP(kind=constants.ROUTE, namespace=config.ENV_DATA["cluster_namespace"])
    route_items = ocp_routes_obj.get().get("items")
    route_names = [item["metadata"]["name"]  for item in route_items)
```

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,126 @@
...
+from ocs_ci.framework import config
+
+
+logger = logging.getLogger(__name__)
+
+
+def check_disable_routes(disable_routes_val):
+    """
+    Args:
+        disable_routes_val (bool) Value to be set to 'disableRoutes' parameter
```
sagihirshfeld: Please summarize what this function checks. It's confusing to read the asserts at the end without the needed context.

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,126 @@
...
+    result = ocp_obj.exec_oc_cmd(command="get routes -o yaml", out_yaml_format=False)
+    data = yaml.safe_load(result)
+    route_names = [item["metadata"]["name"] for item in data["items"]]
+    logger.info(f"Existing routes {route_names}")
+    assert set(all_routes).issubset(
+        route_names
+    ), "Some of the predefined routes don't exist"
+
+    for route in all_routes:
+        ocp_obj.exec_oc_cmd(command=f"delete route {route}")
```
udaysk23: @ypersky1980 Once you are done with @sagihirshfeld's suggestion, Please check `ocp_routes_obj` has any inbuilt method which I suppose `delete()` method(Not completely sure) will delete routes instead of  using `exec_oc_cmd`

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,126 @@
...
+
+    time.sleep(timeout)
```
udaysk23: use `from time import sleep` instead of importing entire module

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,126 @@
...
+
+    result = ocp_obj.exec_oc_cmd(command="get routes -o yaml", out_yaml_format=False)
```
udaysk23: Use `reload_data()` method from OCP class to get the latest route

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,126 @@
...
+        assert set(deletable_routes).isdisjoint(
+            set(route_names)
+        ), "Some routes were not deleted"
+
+
+def set_disable_routes_value(storagecluster_obj, val):
+    """
+    Args:
+        storagecluster_obj (obj): Storage cluster on which the value should be set
+        val (bool) Value to be set to 'disableRoutes' parameter
```
udaysk23: Please provide the doc string about what this method does

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+
+    assert (
+        disable_routes == val
+    ), f"Disable routes is expected to be {val}, is {disable_routes}"
+
+
+@red_squad
+@mcg
+class TestBlockExternalAccess(MCGTest):
```
udaysk23, line 83: Considering logical grouping of the method, I would request you to add `check_disable_routes` and `set_disable_routes_value` method inside the class

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: Added Tier2 marker.

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: Fixed it.  Thanks

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: Done. Thanks

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: Done. Thanks

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: Done. Thanks

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: Done. Thanks

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: Done. Thanks

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: Done. Thanks

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: Fixed. Thank you

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: Fixed. Thank you

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -114,13 +36,185 @@ def test_block_access_from_storagecluster(
...
+        noobaa_dict = noobaa_obj.get()
+        logger.info(f"Updated configuration is {noobaa_dict}")
+
+        disable_routes = noobaa_dict.get("spec", {}).get("disableRoutes")
+
+        assert (
+            disable_routes == val
+        ), f"Disable routes is expected to be {val}, is {disable_routes}"
+
+    def delete_multiCLoudGateway_from_storagecluster(self):
```
sagihirshfeld: If I understand correctly, we only need to remove `/spec/multiCloudGateway/disableRoutes` - please update the method accordingly (and its very scary current name)

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -114,13 +36,185 @@ def test_block_access_from_storagecluster(
...
+
+        self.check_disable_routes(False)
+
+        self.set_storagecluster_disable_routes_value(storagecluster_obj, True)
+        self.check_disable_routes(True)
+
+        self.set_storagecluster_disable_routes_value(storagecluster_obj, False)
+        self.check_disable_routes(False)
+
+    def check_disable_routes(self, disable_routes_val):
```
sagihirshfeld, line 165: There's a core issue with this method - it assumes to predict what routes will be in an ODF deployment, when it can't always do that. Different types of deployments might have different routes, and some may get deleted or added in future versions. 

Luckily it doesn't have to make that prediction.

1. Add an auto-use and class-scoped fixture that will run before these tests run, so it fetches the names of the routes in the cluster and saves them in a class-member:
```
@pytest.fixture(scope="class", autouse=True)
def save_all_routes_set(self):
    """
    Fetch the ODF route names and store them in a class member
    """
    ocp_routes_obj = ocp.OCP(
        kind=constants.ROUTE, namespace=config.ENV_DATA["cluster_namespace"]
    )
    route_items = ocp_routes_obj.get().get("items")
    self.original_route_names = set([item["metadata"]["name"] for item in route_items])
```

2. Get noobaa's routes by using the `app=noobaa` label:
```
ocp_routes_obj = ocp.OCP(
    kind=constants.ROUTE,
    selector=constants.NOOBAA_APP_LABEL,
)
route_items = ocp_routes_obj.get().get("items")
noobaa_route_names = set([item["metadata"]["name"] for item in route_items])
```

3. Fetch all the routes and make sure that `self.original_route_names - noobaa_route_names` was not deleted, while `noobaa_route_names` stays deleted when it's supposed to.

tests/functional/object/mcg/test_block_external_access.py
sagihirshfeld: Also, please invoke this method in class-scoped teardown fixture:
```
@pytest.fixture(scope="class", autouse=True)
def cleanup(self, request):
    request.add_finalizer(self.clear_disable_routes_flag_from_storagecluster)
```
This way we won't accidentally fail subsequent noobaa tests that rely on these routes.

ypersky1980: @sagihirshfeld Regarding what we delete from storagecluster.yaml - we do delete the while multiCloudGateway section, not only the flag, since we would like to have the storagecluster.yaml same as on a freshly deployed cluster, on which this section is not present at all under spec. 

Therefore, I'll rename the method to delete_multiCloudGateway_section_from_storagecluster. 
If you have any better idea, please suggest.

ypersky1980: @sagihirshfeld Teardown fixture added to class, thank you!

ypersky1980: @sagihirshfeld Done. Thank you.

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: @sagihirshfeld Done, thank you.

tests/functional/object/mcg/test_block_external_access.py
sagihirshfeld: The `spec.multiCloudGateway` section does exist at the beginning of the test:
```
$ oc get storagecluster ocs-storagecluster -o jsonpath='{.spec.multiCloudGateway}'
{"endpoints":{"maxCount":2,"minCount":2}}
```

This is because of the [nb_ensure_endpoint_count](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/conftest.py/#L5021-L5101) fixture that applies to `MCGTest`s, including this one.

Deleting it will likely cause a teardown error on the last test in the class due to a `ResourceLeftoversException`:
1. Due to `nb_ensure_endpoint_count` you'll have two noobaa-endpoint pods before the environment checker runs before the first test in the class
2. The environment checker will note that you had two noobaa-endpoint pods before the first test runs
3. Deleting `spec.multiCloudGateway` will result in the deletion of one of the endpoint pods, because the default minCount is 1
4. After the last test in the class the environment checker will run again, and note the difference of the one missing pod, thus throwing a `ResourceLeftoversException`: [get_status_after_execution snippet](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/environment_check.py/#L293-L309)

So to avoid this (and other potential side effects) please tighten the cleanup to only what was added in the test.

sagihirshfeld: Adding a summary of our live discussion on this method and the PR in general:

1. I was wrong to be concerned about getting a `ResourceLeftoversException` error - we excluded the noobaa-endpoint pods from the environment checker (see https://github.com/red-hat-storage/ocs-ci/pull/10350)
2. A point that I missed in my original reviews of the PR: for `noobaa/noobaa.spec.disableRoutes` to take effects, `storagecluster/ocs-storagecluster.spec.multiCloudGateway` must be removed. Hence line 229 in `test_block_access_from_noobaa`. 
3. Still, deleting it can have side effects on other tests because we might decide to add other session-lasting configurations in the future.  The workaround we decided on is to save its original state (=dictionary) in an auto-use and class-scoped fixture at the beginning of the test suite, and patch it back via a class-scoped cleanup after the last test has ran.
4. We also identified that we need to revert the noobaa yaml to its original state at the class-scoped cleanup.

ypersky1980: @sagihirshfeld Done, thank you!

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,252 @@
...
+                self.original_route_names - self.original_noobaa_route_names
+            )
+            assert non_deletable_route_names.issubset(
+                route_names_after_deletion
+            ), "Some of the predefined routes were deleted"
+            assert self.original_noobaa_route_names.isdisjoint(
+                route_names_after_deletion
+            ), "Some routes were not deleted"
+
+    def set_storagecluster_disable_routes_value(self, storagecluster_obj, val):
```
udaysk23: @ypersky1980 is there any chance where we can integrate `set_noobaa_disable_routes_value` method in this method since the only difference I can see is spec path which we can achieve using if else block

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,252 @@
...
+            sc_dict.get("spec", {}).get("multiCloudGateway", {}).get("disableRoutes")
+        )
+
+        assert (
+            disable_routes == val
+        ), f"Disable routes is expected to be {val}, is {disable_routes}"
+
+    def set_noobaa_disable_routes_value(self, noobaa_obj, val):
+        """
+        This method sets the value of 'disableRoutes' flag of storagecluster_obj to val
```
udaysk23:
```suggestion
        This method sets the value of 'disableRoutes' flag of noobaa_obj to val
```

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+@tier2
+@red_squad
+@mcg
+@pytest.mark.usefixtures("save_all_routes_set")
+class TestBlockExternalAccess(MCGTest):
+    @pytest.fixture(scope="class", autouse=True)
+    def cleanup(self, request):
+        request.addfinalizer(self.delete_multiCloudGateway_section_from_storagecluster)
```
udaysk23: @ypersky1980 Are we suppose to delete `multiCloudGateway` section from storagecluster as part of clean up?

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: Done, thank you.

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: No, looks like the previous commit was not pushed. I did it now, please check @udaysk23

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: Done. Thank you.

PR #13111, opened 2025-09-05
Fix conflict with packagemanifest for mcg-operator from submariner CS

petr-balogh:


shylesh (approved):
lgtm

shylesh:
Verified here: https://url.corp.redhat.com/packagemanifest-fix

PR #13122, opened 2025-09-08
Add FDF 4.20 yaml

ebondare:


vavuthu (approved):

petr-balogh:
I think we already have this added by Coady:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/fdf_version/fdf-4.20.yaml

So we can close this one I think, right @ebondare

PR #13123, opened 2025-09-09
use numeric priority in the pytest order for "last"

vavuthu:
test_failure_propagator should be last test case to run and since we had many test cases with same order "last" which make different order which is not expected. use numeric priority in last order.

### DIFF
--- a/tests/cross_functional/kcs/test_monitor_recovery.py
+++ b/tests/cross_functional/kcs/test_monitor_recovery.py
@@ -53,7 +53,7 @@ logger = logging.getLogger(__name__)
 @magenta_squad
 @system_test
 @ignore_leftovers
-@pytest.mark.order("last")
+@pytest.mark.order("last-1")
 @pytest.mark.polarion_id("OCS-3911")
 @skipif_ocs_version("<4.6")
 @skipif_ocp_version("<4.10")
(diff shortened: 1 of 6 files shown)


tests/cross_functional/kcs/test_monitor_recovery.py
```diff
@@ -53,7 +53,7 @@
 @magenta_squad
 @system_test
 @ignore_leftovers
-@pytest.mark.order("last")
+@pytest.mark.order("last-1")
```
petr-balogh, line 56: What about:
@pytest.mark.order("second_to_last")
https://pytest-order.readthedocs.io/en/stable/usage.html#order-relative-to-other-tests

tests/functional/disaster-recovery/regional-dr/test_site_failure_recovery_and_failover.py
```diff
@@ -51,7 +51,7 @@
 @tier4a
 @turquoise_squad
 @dr_hub_recovery
-@pytest.mark.order("last")
+@pytest.mark.order("last-1")
```
am-agrawa, line 54: RDR hub recovery tests should run last as per the current implementation

tests/functional/disaster-recovery/regional-dr/test_neutral_hub_failure_and_recovery.py
```diff
@@ -47,7 +47,7 @@
 @tier2
 @turquoise_squad
 @dr_hub_recovery
-@pytest.mark.order("last")
+@pytest.mark.order("last-1")
```
am-agrawa, line 50: RDR hub recovery tests should run last as per the current implementation

PR #13124, opened 2025-09-09
Add Support for Mix Workload with Discovered Apps

prsurve:


tests/conftest.py
```diff
@@ -7485,6 +7487,13 @@ def factory(
                         "dr_workload_app_placement_name"
                     ],
                     discovered_apps_multi_ns=multi_ns,
+                    pvc_interface=pvc_interface,
+                    mix_workload_rbd_pvc_count=workload_details.get(
```
keemano, line 7491: I think this will raise exception when the `workload_key` is `dr_workload_discovered_apps_cephfs` or `dr_workload_discovered_apps_rbd`. You can make the get to default to None

tests/conftest.py
```diff
@@ -7485,6 +7487,13 @@ def factory(
...
+                        "mix_workload_rbd_pvc_count"
+                    ),
+                    mix_workload_cephfs_pvc_count=workload_details.get(
+                        "mix_workload_cephfs_pvc_count"
```
keemano, line 7495: same as above

ocs_ci/helpers/dr_helpers.py
```diff
@@ -799,16 +803,34 @@ def wait_for_replication_resources_creation(
...
         raise TimeoutExpiredError(error_msg)
 
+    if mix_workload:
+        resource_map = {
+            "rbd": (constants.VOLUME_REPLICATION, get_vr_count),
+            "cephfs": (constants.REPLICATION_SOURCE, get_replicationsources_count),
+        }
+
+        for key, vr_count in mix_workload_data.items():
+            if key not in resource_map:
```
keemano, line 812: @prsurve Do we really need this because `mix_workload_data` dict is in user control, as we create this dictionary during deployment of the workload.

ocs_ci/helpers/dr_helpers.py
```diff
@@ -799,16 +803,34 @@ def wait_for_replication_resources_creation(
...
+                logger.warning(
+                    f"Unsupported storage type '{key}' in mix_workload. Skipping..."
+                )
+                continue
+
+            resource_kind, count_function = resource_map[key]
+            wait_for_resource_creation_and_state(
```
keemano, line 825: In `wait_for_resource_creation_and_state`, there is no validation for replicationsources being done. So you we can skip passing cephfs args to that function

keemano, line 825: I think from line no 834 to 862 does what `wait_for_resource_creation_and_state` does. Looks like the existing code will call the function `wait_for_resource_creation_and_state`

prsurve: Yes now i have  removed duplicate code

prsurve: It's getting validate via count_function

ocs_ci/helpers/dr_helpers.py
prsurve: Yes we need this

PR #13133, opened 2025-09-09
Add test for ODF pod resource requests and limits

yitzhak12:
This PR introduces a new functional test to validate the resource requests and limits for OpenShift Data Foundation (ODF) pods. The test compares live cluster configurations against a predefined "source of truth" in a new JSON file, ensuring that critical ODF components have the expected CPU and memory allocations.


tests/functional/pod_and_daemons/test_odf_resources_requests_and_limits.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # Step 3: Compare the live data with the expected data
+        log.info("Comparing live pod resource data with expected configurations.")
+        has_mismatches = check_odf_resources_requests_and_limits(
+            pods_resources_details_dict
+        )
+
+        # Step 4: Assert that there are no differences
+        error_message = "Found resource mismatches. Please check the logs for details."
+        assert has_mismatches, error_message
```
ebenahar: if there are mismatches, this will not print what the mismatch is. I suggest printing the mismatch in the assert error message to make it easier to understand for debugging purposes

tests/functional/pod_and_daemons/test_odf_resources_requests_and_limits.py
```diff
@@ -0,0 +1,75 @@
...
+
+@tier1
+@skipif_ocs_version("<4.20")
+class TestLiveResourcesPresenceAndFormat(BaseTest):
+    """
+    Functional test to verify that live pod resource values (requests/limits)
+    exist and start with a digit (no None/'N/A'/'null'/non-numeric prefixes).
+    """
+
+    @brown_squad
```
ebenahar, line 32: polarion ID is missing

ebenahar (dismissed):

tests/functional/pod_and_daemons/test_odf_resources_requests_and_limits.py
yitzhak12: Fixed

ocs_ci/helpers/pod_helpers.py
```diff
@@ -0,0 +1,144 @@
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def get_pod_resource_details(pod_obj):
```
suchita-g: Can we modify the name of the Function with sense of "containers" details instead . Like get_pod_container_resource

suchita-g (commented):
rest looks good to me except one of the commented functions name is not really sounds like a what that function is achieving , plus it resembles similar to another function name just after that so It may be confusing in future. I strongly recommend to update the function get_pod_resource_details with suitable name

prsurve (approved):

ocs_ci/helpers/pod_helpers.py
yitzhak12: Fixed

Issue #13144, opened 2025-09-10
[Bug]: test_csi_addon_pod_security fails on ODF 4.20 due to CSI Addon pod restructuring (no csi-addons container in nodeplugin pod)


paraggit:
### Description
On ODF 4.20, `tests/functional/pod_and_daemons/test_csiaddon_pod_security.py::TestCSIAddonPodSecurity::test_csi_addon_pod_security` fails at the assertion that expects a `csi-addons` container inside the `RBD nodeplugin` pod selected by label `app=openshift-storage.rbd.csi.ceph.com-nodeplugin.`
From 4.20 onwards, CSI Addon containers are no longer embedded in the RBD nodeplugin pods. Instead, separate CSI Addon pods (e.g., openshift-storage.rbd.csi.ceph.com-nodeplugin-csi-addons-*) exist. The current test still targets the legacy pod layout and therefore cannot find the csi-addons container.

### Steps to Reproduce
1. Deploy ODF 4.20
2. Run pytest tests/functional/pod_and_daemons/test_csiaddon_pod_security.py::TestCSIAddonPodSecurity::test_csi_addon_pod_security.
3. The test fetches a pod using label app=openshift-storage.rbd.csi.ceph.com-nodeplugin and then tries to locate the csi-addons container inside it.

### Actual Behavior

The test fails because the selected RBD nodeplugin pod does not contain a csi-addons container on ODF 4.20.

```
self = <test_csiaddon_pod_security.TestCSIAddonPodSecurity object at 0x7ff2ebd09340>

def test_csi_addon_pod_security(self):
    ...
    pod_obj = Pod(**get_pods_having_label(CSI_RBDPLUGIN_LABEL_419)[0])
    csi_addon_container = pod_obj.get_container_data("csi-addons")

>   assert csi_addon_container, "No CSI Addon container found in the pod"
E   AssertionError: No CSI Addon container found in the pod
E   assert []

tests/functional/pod_and_daemons/test_csiaddon_pod_security.py:52: AssertionError
```
### Expected Behavior
The test should validate Pod Security Standards against the actual CSI Addon pod or container as laid out in ODF 4.20:
	•	Reachable securely over HTTPS on {localhost}:{port}/healthz.
	•	Reject insecure HTTP connections (non-TLS).
### Environment
- Test Suite(s): tests/functional/pod_and_daemons/test_csiaddon_pod_security.py
- Platform(s): ODF 4.20
- Version(s):
- OS:

[...]

PR #13145, opened 2025-09-10
fix(test_csiaddon_pod_security.py): adapt CSI addon pod security test for ODF 4.20 version

paraggit:
This is a Fix for the issue: https://github.com/red-hat-storage/ocs-ci/issues/13144

### Problem Statement
The issue was that in ODF 4.20, the CSI addon functionality has been restructured:
Before ODF 4.20: CSI RBD nodeplugin pods contained csi-addons containers directly
ODF 4.20+: CSI addon functionality is in separate dedicated pods with names like:
openshift-storage.rbd.csi.ceph.com-nodeplugin-csi-addons-*
The test was using the old approach (looking for csi-addons containers in traditional RBD nodeplugin pods), which no longer works in ODF 4.20.
### Solution Overview
New constants for ODF 4.20 CSI addon pod labels
Simplified, efficient pod discovery logic
Removed unnecessary version constraints

### DIFF
--- a/tests/functional/pod_and_daemons/test_csiaddon_pod_security.py
+++ b/tests/functional/pod_and_daemons/test_csiaddon_pod_security.py
@@ -1,23 +1,20 @@
 import pytest
 import logging
 
-from ocs_ci.ocs.resources.pod import get_pods_having_label
-from ocs_ci.ocs.resources.pod import Pod
 from ocs_ci.ocs.exceptions import CommandFailed
-from ocs_ci.ocs.constants import CSI_RBDPLUGIN_LABEL_419
 from ocs_ci.framework.pytest_customization.marks import (
     tier1,
     green_squad,
 )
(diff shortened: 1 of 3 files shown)


tests/functional/pod_and_daemons/test_csiaddon_pod_security.py
```diff
@@ -1,23 +1,75 @@
...
 )
-from ocs_ci.framework.pytest_customization.marks import polarion_id, skipif_ocs_version
+from ocs_ci.framework.pytest_customization.marks import polarion_id
+from ocs_ci.framework import config
 
-log = logging
+log = logging.getLogger(__name__)
+
+
+def get_csi_addon_pod():
```
ebenahar: please move this function into a util/helper file

paraggit: Moved function into helper file

PR #13146, opened 2025-09-10
hosted/spoke odf deployment separation

DanielOsypenko:
This PR changes the hierarchy of classes and follows the Mixins (lightweight base classes) design e.g. HyperShiftBase, MetalLBInstaller, CNVInstaller, MCEInstaller, HyperConverged) that each encapsulate a focused slice of behavior.

- separates common functionality from `HostedCluster` to new classes - parent class `SpokeODF` and child classes `ExternalODF`, `HostedODF`, also HostedOCP to new classes - parent class SpokeOCP and child classes HypershiftHostedOCP and ExternalOCP.
- introduces new flow for deployment of ExternalODF
- added iterative method call do_deploy_external_spoke_clusters to deploy multiple external clusters
- added verification flow  before deployment to test network and StorageCluster on Hub for multiple external clusters
- added verification flow  after deployment to test StorageConsumer, backing cephstore, Storage Classes on all clients

After merging this PR, considering we have Jenkins job that triggers External OCP deployment and provides kubeconfig to a cluster_path on agent of a hub cluster we can run **Deployments like vSphere storage Hub to vSphere Spoke(s)**


DanielOsypenko:
verified with Hosted cluster OCP and ODF deployment: https://url.corp.redhat.com/308e0e8

DanielOsypenko:
tested with external spoke: https://url.corp.redhat.com/a5fcf3f

ocs_ci/deployment/deployment.py
```diff
@@ -666,19 +666,46 @@ def do_deploy_metallb(self):
         if config.DEPLOYMENT.get("metallb_operator"):
             MetalLBInstaller().deploy_lb()
 
-    def do_deploy_hosted_clusters(self):
+    def do_deploy_hosted_spoke_clusters(self):
         """
         Deploy Hosted cluster(s)
```
dahorak, line 701: It might be worth to update also the docstring here.

ocs_ci/deployment/deployment.py
```diff
@@ -666,19 +666,46 @@ def do_deploy_metallb(self):
...
+            from ocs_ci.deployment.hub_spoke import HostedClients
 
             HostedClients().do_deploy()
 
+    def do_deploy_external_spoke_clusters(self):
+        """
+        Deploy External spoke cluster(s)
+        """
+        if config.ENV_DATA.get("clusters", False) and (
+            not config.ENV_DATA["skip_ocs_deployment"]
```
dahorak, line 724: I see similar condition used also above for hosted client clusters, but what is the reason for this part of the condition on both places?
```python
not config.ENV_DATA["skip_ocs_deployment"]
```

ocs_ci/deployment/hub_spoke.py
```diff
@@ -493,336 +1075,315 @@ def do_deploy(self, cluster_names=None):
...
 
-    def verify_client_cluster_storage(self, cluster_name):
+    def verify_hosted_ocp_clusters_from_provider(self):
         """
-        Verify storage connectivity for a single cluster by checking storage class existence
+        Verify multiple HyperShift hosted clusters from provider. If cluster_names is not provided at ENV_DATA,
+        it will get the list of hosted clusters from the provider to verify them all
+
+        Returns:
+            bool: True if all hosted clusters passed verification, False otherwise
```
dahorak, line 1085: There should be one empty line after lists in markdown.

ocs_ci/deployment/hub_spoke.py
```diff
@@ -493,336 +1075,315 @@ def do_deploy(self, cluster_names=None):
...
         Args:
-            cluster_name (str): Name of the cluster to verify
+            cluster_names_paths_dict (dict): Optional argument. The function will download all kubeconfigs
+            to the folders specified in the configuration, or download a specific cluster's kubeconfig
+            to the folder provided as an argument.
+            from_hcp (bool): If True, download kubeconfig from HCP, otherwise from the secret
 
         Returns:
-            bool: True if storage classes exist and are properly configured, False otherwise
+            list: the list of hosted cluster kubeconfig paths
```
dahorak, line 1129: Same as above, there should be one empty line after list in markdown. Similarly also on few other places updated in this PR.

ocs_ci/deployment/hub_spoke.py
```diff
@@ -493,336 +1075,315 @@ def do_deploy(self, cluster_names=None):
...
 
-        return True
+    def get_kubeconfig_path(self, cluster_name):
+        """
+        Get the kubeconfig path for the cluster
+
+        Args:
+            cluster_name (str): Name of the cluster
+        Returns:
+            str: Path to the kubeconfig file
```
dahorak, line 1167: Similarly as above, missing empty line after each list in markdown.

ocs_ci/deployment/hub_spoke.py
```diff
@@ -493,336 +1075,315 @@ def do_deploy(self, cluster_names=None):
...
-        """
-        self.update_hcp_binary()
-
-        hosted_cluster_names = get_hosted_cluster_names()
-
-        for cluster_name in hosted_cluster_names:
-            logger.info(f"Deploying ODF client on hosted OCP cluster '{cluster_name}'")
-            hosted_odf = HostedODF(cluster_name)
-            hosted_odf.do_deploy()
+        logger.info(f"Port {port} on {ip_address} is open")
```
dahorak: Since you are processing the return value from this function and generate another log error message in case the port is not open, I think both those logging messages could be just on `DEBUG` level.

ocs_ci/deployment/deployment.py
DanielOsypenko: It is another fuse on not running this code when we don't need, it is historically here. 
In any case we always have this True when we install OCP or ODF on spokes, because we are doing it via Deploy OCS stage in backing Deploy OCS Cluster job.

ocs_ci/deployment/deployment.py
DanielOsypenko: they are both spoke(s) in k8s/DF terms (a workload cluster)

ocs_ci/deployment/hub_spoke.py
DanielOsypenko: fixed

ocs_ci/deployment/hub_spoke.py
DanielOsypenko: fixed

ocs_ci/deployment/hub_spoke.py
DanielOsypenko: fixed

ocs_ci/deployment/hub_spoke.py
DanielOsypenko: messages are different; outside of this function we are emphasizing on skipping deploying the cluster because of latency. Here we print the result of this function. I think these few lines are very important and in future may save significant time on debugging, if engineer can observe them quickly. Also this functions presumably may be used outside of the deployment scenario.

ocs_ci/deployment/deployment.py
dahorak: That was my point - it is not only about Hosted clusters.

DanielOsypenko: method name was updated

dahorak: I meant to include the _spoke_ cluster(s) also in the doc string:
```suggestion
        Deploy Hosted and External cluster(s)
```
or something like that.

ocs_ci/utility/utils.py
```diff
@@ -4413,7 +4413,11 @@ def get_system_architecture():
 
 
 def wait_for_machineconfigpool_status(
-    node_type, timeout=1900, skip_tls_verify=False, force_delete_pods=False
+    node_type,
+    timeout=1900,
+    skip_tls_verify=False,
+    force_delete_pods=False,
+    cluster_kubeconfig=None,
```
dahorak: Parameter `cluster_kubeconfig` shouldn't be `None` but empty string `""`, similarly as defined here: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/ocp.py#L54 otherwise it leads to following error:
```
2025-09-11 19:52:36  ocs_ci/utility/utils.py:4460: in wait_for_machineconfigpool_status
2025-09-11 19:52:36      machine_count = ocp_obj.get()["status"]["machineCount"]
2025-09-11 19:52:36  ocs_ci/ocs/ocp.py:341: in get
2025-09-11 19:52:36      return self.exec_oc_cmd(
2025-09-11 19:52:36  ocs_ci/ocs/ocp.py:199: in exec_oc_cmd
2025-09-11 19:52:36      self.cluster_kubeconfig if os.path.exists(self.cluster_kubeconfig) else None
2025-09-11 19:52:36  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
2025-09-11 19:52:36  
2025-09-11 19:52:36  path = None
2025-09-11 19:52:36  
2025-09-11 19:52:36      def exists(path):
2025-09-11 19:52:36          """Test whether a path exists.  Returns False for broken symbolic links"""
2025-09-11 19:52:36          try:
2025-09-11 19:52:36  >           os.stat(path)
2025-09-11 19:52:36  E           TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```
https://url.corp.redhat.com/8d0996c

ocs_ci/deployment/deployment.py
DanielOsypenko: This method is for deploying Hosted clusters only. We can add "spoke" to the name of this method because it is not contradicting with the fact that Hosted cluster similarly to External, they are both spokes. But this method can not be used for deploying External clusters.
We have 2 of them:

* do_deploy_hosted_spoke_clusters
* do_deploy_external_spoke_clusters

I separated them to emphasize their purpose in https://github.com/red-hat-storage/ocs-ci/blob/25d7586d048ffe8b32fc9a54b6bfab832ad7e309/ocs_ci/deployment/deployment.py#L709

dahorak: Ah... thanks for detailed explanation... the `spoke` work mislead me little bit.

DanielOsypenko:
separate Hosted cluster OCP & ODF installation passed - https://url.corp.redhat.com/e657758

dahorak (approved):

jilju (approved):

dahorak:
Verification jobs (it was split to two follow-up jobs, because the first one has accidentally disabled client clusters deployment):
* https://url.corp.redhat.com/eac373f
* https://url.corp.redhat.com/85e10d9

dahorak (approved):

PR #13164, opened 2025-09-10
Enable automatic import of hosted clusters in ACM hub

jilju:
This PR is to enable automatic import of hosted clusters to ACM hub.
Reference : https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.14/html-single/multicluster_engine_operator_with_red_hat_advanced_cluster_management/index#hosted-discover
and
https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.14/html-single/multicluster_engine_operator_with_red_hat_advanced_cluster_management/index#auto-import-hcp

### DIFF
--- a/ocs_ci/ocs/acm/acm.py
+++ b/ocs_ci/ocs/acm/acm.py
@@ -892,6 +892,10 @@ def discover_hosted_clusters():
 
 
 def install_clusteradm():
+    """
+    Install clusteradm CLI
+
+    """
     try:
         run_cmd("clusteradm")
     except CommandFailed:
@@ -902,6 +906,10 @@ def install_clusteradm():
 
 
 def automate_import_of_hosted_clusters():
+    """
+    Enable automatic import of hosted clusters that are created on the imported multicluster engine clusters
+
+    """
     policy_mce_hcp_autoimport = create_resource(
         **templating.load_yaml(constants.POLICY_MCE_HCP_AUTOIMPORT_YAML)
     )


fbalak (approved):

ocs_ci/ocs/acm/acm.py
```diff
@@ -815,6 +820,103 @@ def import_clusters_with_acm():
...
+        run_cmd("clusteradm")
+    except CommandFailed:
+        # Install/re0install clusteradm
+        run_cmd(
+            "curl -L https://raw.githubusercontent.com/open-cluster-management-io/clusteradm/main/install.sh | bash"
+        )
+
+
+def automate_import_of_hosted_clusters():
+    policy_mce_hcp_autoimport = create_resource(
```
ebondare, line 913: Docstring?

ocs_ci/ocs/acm/acm.py
```diff
@@ -815,6 +820,103 @@ def import_clusters_with_acm():
...
+                label="app=hypershift-addon-agent",
+                expected_count=1,
+                namespace=agent_install_namespace,
+                timeout=600,
+                sleep=20,
+            )
+
+
+def install_clusteradm():
+    try:
```
ebondare, line 899: Docstring?

ocs_ci/ocs/acm/acm.py
jilju: Added.

ocs_ci/ocs/acm/acm.py
jilju: Added.

ebondare (approved):

jilju:
This code change is needed to unblock development of further stages. This cannot be verified alone and should be accompanied with further changes. Merging this PR to unblock development. The code change in the PR will work only if a parameter "configure_acm_to_import_mce" is present. So this change does not cause any regression.

PR #13169, opened 2025-09-11
drop "hosted" from config names

DanielOsypenko:
Drop "hosted" from config names allows to move away from "hosted"-only deployments for Hub/Spoke storage design and deploy External spoke as well as other in future.

### DIFF
--- a/ocs_ci/deployment/hosted_cluster.py
+++ b/ocs_ci/deployment/hosted_cluster.py
@@ -576,9 +576,7 @@ class HostedClients(HyperShiftBase):
 
         """
         version_exists = (
-            config.ENV_DATA.get("clusters")
-            .get(cluster_name)
-            .get("hosted_odf_version", False)
+            config.ENV_DATA.get("clusters").get(cluster_name).get("odf_version", False)
         )
 
         return version_exists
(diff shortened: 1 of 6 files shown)


ocs_ci/deployment/hosted_cluster.py
```diff
@@ -2073,12 +2069,12 @@ def hypershift_cluster_factory(
...
             cluster_path = create_cluster_dir(cluster_name)
             hosted_cluster_conf_on_provider["ENV_DATA"]["clusters"][cluster_name] = {
-                "hosted_cluster_path": cluster_path,
+                "cluster_path": cluster_path,
                 "ocp_version": ocp_version,
                 "cpu_cores_per_hosted_cluster": 8,
                 "memory_per_hosted_cluster": "12Gi",
-                "hosted_odf_registry": "quay.io/rhceph-dev/ocs-registry",
-                "hosted_odf_version": odf_version,
+                "odf_registry": "quay.io/rhceph-dev/ocs-registry",
```
dahorak: I know it is little bit unrelated to this PR, but what about using here also the `defaults.ODF_REGISTRY_DEFAULT`?

DanielOsypenko: definitely makes sense

PR #13204, opened 2025-09-15
Make 'disable_default_sources' default True for hcp clusters

jilju:
Set the default value of '`disable_default_sources`' True for hcp clusters.
This is needed because to enabled RDR on hcp clusters(used as client clusters) we need to have custom catalogsource redhat-operators created to install df operator automatically. If the default sources are not disabled, the custom catalogsource redhat-operators will be reconciled.


jilju:
This is needed because to enabled RDR on hcp clusters(used as client clusters) we need to have custom catalogsource redhat-operators created to install df operator automatically. If the default sources are not disabled, the custom catalogsource redhat-operators will be reconciled.

prsurve (approved):

fbalak (approved):

PR #13221, opened 2025-09-17
Test to verify that the CSI addon pods running on each worker node

suchita-g:
4.20 : RHSTOR- 7086: Verify that the CSI addon pods running on each worker node


suchita-g:
Duplicate of 13349

PR #13225, opened 2025-09-17
Fixing Bugs for FDF Deployment on some SDS cluster

rh-aaleem:
Following things are being fixed with these changes 

 **Not able to create a idms**
 If IDMS is with the same name is there on the cluster script will and stop the pipeline there itself. for that we are checking if the df-repo is already there we are applying the latest changes.
 
 **Not able to change FSD**
 In fusion Service Definition while patching we are hardcoding the ocp version that i am getting runtime and patching the FSD correctly in this branch.
 
 **Not able to install correct and mentioned version of the FDF**
 Its always installing the gaed version, its not installing the dev version always failing the reason was simple it was not able to edit the FSD correctly.
 
 **Backing Storage**
 By default we are going for local rather dynamic because mostly Z-stream as test on Local mode.
 
 
 **How Do I tested?**
 I tested by running the `Deploy-fdf-cluster` pipeline.

### DIFF
--- a/ocs_ci/deployment/fusion_data_foundation.py
+++ b/ocs_ci/deployment/fusion_data_foundation.py
@@ -49,24 +49,24 @@ class FusionDataFoundationDeployment:
         # self.setup_storage()
 
     def create_image_tag_mirror_set(self):
-    """
-    Create or update ImageTagMirrorSet.
-    """
-    logger.info("Creating or Updating FDF ImageTagMirrorSet")
+        """
+        Create or update ImageTagMirrorSet.
+        """
+        logger.info("Creating or Updating FDF ImageTagMirrorSet")
 
-    imagetag_file = constants.FDF_IMAGE_TAG_MIRROR_SET
-    resource_name = "isf-fdf" 
+        imagetag_file = constants.FDF_IMAGE_TAG_MIRROR_SET
+        resource_name = "isf-fdf" 
 
-    # Check if resource exists
-    cmd_check = f"oc --kubeconfig {self.kubeconfig} get imagetagmirrorset {resource_name}"
-    exists = run_cmd(cmd_check, ignore_error=True)
+        # Check if resource exists
+        cmd_check = f"oc --kubeconfig {self.kubeconfig} get imagetagmirrorset {resource_name}"
+        exists = run_cmd(cmd_check, ignore_error=True)
 
-    if exists:
-        logger.info("ImageTagMirrorSet exists applying changes")
-        run_cmd(f"oc --kubeconfig {self.kubeconfig} apply -f {imagetag_file}")
-    else:
-        logger.info("ImageTagMirrorSet not found creating with save-config")
-        run_cmd(f"oc --kubeconfig {self.kubeconfig} create --save-config -f {imagetag_file}")
+        if exists:
+            logger.info("ImageTagMirrorSet exists applying changes")
+            run_cmd(f"oc --kubeconfig {self.kubeconfig} apply -f {imagetag_file}")
+        else:
+            logger.info("ImageTagMirrorSet not found creating with save-config")
+            run_cmd(f"oc --kubeconfig {self.kubeconfig} create --save-config -f {imagetag_file}")
 
     def create_image_digest_mirror_set(self):
         """
--- a/ocs_ci/deployment/fusion_data_foundation.py
+++ b/ocs_ci/deployment/fusion_data_foundation.py
@@ -19,7 +19,7 @@ from ocs_ci.utility.retry import retry
 from ocs_ci.utility.utils import run_cmd
 
 logger = logging.getLogger(__name__)
-
+# 1
 
 class FusionDataFoundationDeployment:
     def __init__(self):
@@ -32,9 +32,10 @@ class FusionDataFoundationDeployment:
         """
         logger.info("Installing IBM Fusion Data Foundation")
         if self.pre_release:
-            self.create_image_tag_mirror_set()
+            # self.create_image_tag_mirror_set()
             self.create_image_digest_mirror_set()
             self.setup_fdf_pre_release_deployment()
+
         self.create_fdf_service_cr()
         self.verify_fdf_installation()
         # self.setup_storage()
--- a/ocs_ci/deployment/fusion_data_foundation.py
+++ b/ocs_ci/deployment/fusion_data_foundation.py
@@ -25,7 +25,6 @@ from ocs_ci.utility.utils import (
 ) 
 
 logger = logging.getLogger(__name__)
-# 1
 
 class FusionDataFoundationDeployment:
     def __init__(self):
@@ -48,7 +47,7 @@ class FusionDataFoundationDeployment:
 
         self.create_fdf_service_cr()
         self.verify_fdf_installation()
-        # self.setup_storage()
+        self.setup_storage()
 
     def create_image_tag_mirror_set(self):
         """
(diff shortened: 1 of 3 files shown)


ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -33,30 +38,54 @@ def deploy(self):
         logger.info("Installing IBM Fusion Data Foundation")
         if self.pre_release:
             self.create_image_tag_mirror_set()
+            time.sleep(60)
+            wait_for_machineconfigpool_status(node_type="all")
```
dahorak: Do we need to wait for the MCP status here? Shouldn't be enough to wait for the MCP status after the creation of Image Digest Mirror Set? (so we will save some time, because both ITMS and IDMS will be propagated during one MCP update)

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -33,30 +38,54 @@ def deploy(self):
...
+
+        cmd_check = f"oc --kubeconfig {self.kubeconfig} get imagetagmirrorset {resource_name}"
+        exists = run_cmd(cmd_check, ignore_error=True)
+
+        if exists:
+            logger.info("ImageTagMirrorSet exists applying changes")
+            run_cmd(f"oc --kubeconfig {self.kubeconfig} apply -f {imagetag_file}")
+        else:
+            logger.info("ImageTagMirrorSet not found creating with save-config")
+            run_cmd(f"oc --kubeconfig {self.kubeconfig} create --save-config -f {imagetag_file}")
```
dahorak: Do we need to differentiate between creating or updating the ITMS? I think you can use `oc apply ...` (like only changing the original command to `apply` instead of `create`) in both cases, or is there any issue with that?

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -33,30 +38,54 @@ def deploy(self):
...
+
+        cmd_check = f"oc --kubeconfig {self.kubeconfig} get imagedigestmirrorset {resource_name}"
+        exists = run_cmd(cmd_check, ignore_error=True)
+
+        if exists:
+            logger.info("ImageDigestMirrorSet exists applying changes")
+            run_cmd(f"oc --kubeconfig {self.kubeconfig} apply -f {image_digest_mirror_set}")
+        else:
+            logger.info("ImageDigestMirrorSet not found creating with save-config")
+            run_cmd(f"oc --kubeconfig {self.kubeconfig} create --save-config -f {image_digest_mirror_set}")
```
dahorak: Same as above, shouldn't be sufficient to just change the `create` to `apply`?

ocs_ci/deployment/fusion_data_foundation.py
rh-aaleem: if i was using `apply` it was raising error if the ITMS or IDMS was not there or being created,

2025-09-15 14:33:28 2025-09-15 05:03:27,509 - MainThread - ocs_ci.utility.utils - WARNING - Command stderr: Warning: resource imagetagmirrorsets/isf-fdf is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by oc apply. oc apply should only be used on resources created declaratively by either oc create --save-config or oc apply. The missing annotation will be patched automatically.

ocs_ci/deployment/fusion_data_foundation.py
rh-aaleem: Yeah you are right i can try that

ocs_ci/deployment/fusion_data_foundation.py
rh-aaleem: `apply` throwing an error if resources doesnt exist thats why i put a check on it,

dahorak (dismissed):

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -17,10 +17,15 @@
 from ocs_ci.utility import templating
 from ocs_ci.utility.retry import retry
 from ocs_ci.utility.utils import run_cmd
+import time
+from ocs_ci.utility.utils import (
+    get_ocp_version,
```
petr-balogh: Unused import

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -34,29 +39,51 @@ def deploy(self):
         if self.pre_release:
             self.create_image_tag_mirror_set()
             self.create_image_digest_mirror_set()
+            time.sleep(60)
+            wait_for_machineconfigpool_status(node_type="all")
```
clacroix12: If this is necessary for `self.setup_fdf_pre_release_deployment()` I would rather move it to that method as I would like to reserve this top level `deploy` method for higher level method calls and not direct functionality. This is to keep the logical flow of the deployment clear and concise without muddying it with lower level logic.

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -34,29 +39,51 @@ def deploy(self):
...
+        if exists:
+            logger.info("ImageTagMirrorSet exists applying changes")
+            run_cmd(f"oc --kubeconfig {self.kubeconfig} apply -f {imagetag_file}")
+        else:
+            logger.info("ImageTagMirrorSet not found creating with save-config")
+            run_cmd(f"oc --kubeconfig {self.kubeconfig} create --save-config -f {imagetag_file}")
 
     def create_image_digest_mirror_set(self):
         """
         Create ImageDigestMirrorSet.
```
clacroix12:
```suggestion
        Create or update ImageDigestMirrorSet.
```

ocs_ci/templates/fusion-data-foundation/data-foundation-instance.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     value: Fusion
   - name: backingStorageType
     provided: true
-    value: Dynamic
+    value: Local
```
clacroix12, line 18: Should this be configurable or do we only plan on testing local backingStorageType?

rh-aaleem: this should be configurable, mostly right now we are doing testing on local mode,

ocs_ci/deployment/fusion_data_foundation.py
rh-aaleem: Yeah thats looking good approach to make it clear and concise,

ocs_ci/templates/fusion-data-foundation/data-foundation-instance.yaml
petr-balogh: If this is the default - let's have this as default and then bring up the possibility to overwrite this value from config.

petr-balogh (approved):

ocs_ci/deployment/fusion_data_foundation.py
petr-balogh: @rh-aaleem , the apply error is actually just a warning. So it will not fail anything and we are already using that in many places, so you can just ignore that warning.

We can just run normal :
```
run_cmd(f"oc --kubeconfig {self.kubeconfig} apply -f {imagetag_file}", silent=True)
```
 and add `silent=True` there to surpass the error.

rh-aaleem: Sure @petr-balogh i will do the changes and test it and push it,

ocs_ci/templates/fusion-data-foundation/data-foundation-instance.yaml
clacroix12: FYI here is the documentation on which modes are supported for different providers: https://www.ibm.com/docs/en/fusion-software/2.11.0?topic=foundation-installing-fusion-data#tasksf_enable_odf_service__context__1

petr-balogh (approved):

clacroix12:
Hey @rh-aaleem, will you be able to cherry-pick this to `release-4.18` and `release-4.19`? Once you do I will do the same for my fix to the FDF image tag here: #13309. Thanks!

PR #13229, opened 2025-09-17
Create number of VMs based on resources availability

avd-sagare:
This PR introduces logic to determine the maximum number of VMs that can be provisioned on an OpenShift cluster based on available resource constraints.
The VM count is calculated considering:

- Available Storage
- CPU capacity
- Memory capacity

The final VM count is derived by taking the minimum value from these three resource-based calculations, ensuring that VM provisioning respects all cluster resource limits.

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -3495,10 +3495,10 @@ def get_pod_requests_per_node(worker_nodes=None):
         if not node or (worker_nodes and node not in worker_nodes):
             continue  # skip unscheduled pods and non-worker nodes
 
-        for container in pod["spec"].get("containers", []):
+        for container in pod_data["spec"].get("containers", []):
             requests = container.get("resources", {}).get("requests", {})
-            cpu_req = requests.get("cpu", "0")
-            mem_req = requests.get("memory", "0")
+            cpu_req = str(requests.get("cpu") or "0")
+            mem_req = str(requests.get("memory") or "0")
 
             # CPU conversion
             if cpu_req.endswith("m"):  # e.g. "500m"


ocs_ci/helpers/cnv_helpers.py
```diff
@@ -620,3 +620,21 @@ def check_fio_status(vm_obj, fio_service_name="fio_test"):
     """
     output = vm_obj.run_ssh_cmd(f"systemctl status {fio_service_name}")
     return "running" in output
+
+
+def compute_vm_count_from_storage_capacity():
+    """
+    Computes the number of VMs to create based on available storage capacity.
```
hnallurv, line 627: Shouldn't CPU cores, RAM and N/W bandwidth also be considered?

tests/conftest.py
```diff
@@ -7856,10 +7860,13 @@ def multi_cnv_workload_factory(request, storageclass_factory, cnv_workload):
 
     """
 
-    def factory(namespace=None, encrypted=False):
+    def factory(namespace=None, encrypted=False, vm_count=False):
         """
         Args:
             namespace (str, optional): The namespace to create the vm on.
+            vm_count (bool): If True, the function will create VMs based on available storage capacity.
```
hnallurv: somehow, vm_"count" is not a right name for a boolean type argument. Consider changing the name of the argument - something like 'create_from_config_yaml'....

ocs_ci/helpers/cnv_helpers.py
PrasadDesala: +1

ocs_ci/helpers/cnv_helpers.py
```diff
@@ -620,3 +620,21 @@ def check_fio_status(vm_obj, fio_service_name="fio_test"):
...
+
+    This function calculates the number of VMs to create by determining the usable storage
+    capacity after accounting for used storage and then dividing it by the VM size (30 GiB).
+
+    Returns:
+        int: The computed number of VMs to create.
+    """
```
PrasadDesala, line 634: please leave an empty line above this

ocs_ci/helpers/cnv_helpers.py
```diff
@@ -620,3 +620,21 @@ def check_fio_status(vm_obj, fio_service_name="fio_test"):
...
+    total_used_in_gibibytes, _ = cluster.get_used_and_total_capacity_in_gibibytes()
+    ceph_obj = cluster.CephCluster()
+    total_capacity_in_gibibytes = ceph_obj.get_ceph_capacity()
```
PrasadDesala, line 637: get_used_and_total_capacity_in_gibibytes returns a tuple of used capacity and total capacity.  You don't need to get the ceph capacity again calling a different function

ocs_ci/helpers/cnv_helpers.py
```diff
@@ -620,3 +620,21 @@ def check_fio_status(vm_obj, fio_service_name="fio_test"):
...
+    usable_capacity = (0.75 * total_capacity_in_gibibytes) - total_used_in_gibibytes
+    vm_count = int(usable_capacity / 30)  # 30 gib is size of 1 vm
```
PrasadDesala, line 639: how did we come to this number to divide by 30? does it remain same for any type of VM? (linux, fedora, windows)?

tests/conftest.py
```diff
@@ -7856,10 +7860,13 @@ def multi_cnv_workload_factory(request, storageclass_factory, cnv_workload):
 
     """
 
-    def factory(namespace=None, encrypted=False):
+    def factory(namespace=None, encrypted=False, vm_count=False):
```
PrasadDesala: What is the main purpose of having this argument to create those number of VMs based on free storage?

IIUC, calling multi_cnv_workload_factory creates 7 VMs with various access modes and types, which can be more than sufficient for functionality testing. Please be aware that it is not just storage, we also need to consider CPU and memory. There are high chances of the VMs not getting up because of insufficient CPU and memory resources and this can also impact other pods running on the worker nodes.

For instance, if we take 500GB as usable capacity, you are trying to create 16 VMs by repeating the cnv_vm_configs. I doubt if that many can run on a shared worker nodes or even dedicated nodes for that matter (leaving BM high config).

ocs_ci/helpers/cnv_helpers.py
avd-sagare: Yes. I am working on that. I will consider CPU and RAM. not sure about n/w bandwidth now.

ocs_ci/helpers/cnv_helpers.py
avd-sagare: It gives total capacity without replica. get_ceph_capacity() gives me total capacity with replica 3.

ocs_ci/helpers/cnv_helpers.py
avd-sagare: Yes. right now we are taking 30gb rootdisk .

tests/conftest.py
avd-sagare: This PR is basically create for resiliency testing. here we are getting VM count and passing to resiliency tool created by Parag.

tests/conftest.py
avd-sagare: correct. I will chang eit

ocs_ci/helpers/cnv_helpers.py
avd-sagare: Done

ocs_ci/ocs/node.py
```diff
@@ -3417,3 +3418,87 @@ def select_osd_node():
...
+    This function queries the OpenShift (okd) cluster's API using 'oc' command to retrieve the
+    available CPU and memory resources for each worker node. The results are then parsed
+    and returned as a dictionary with node names as keys, and nested dictionaries containing
+    CPU (in millicores) and Memory (in GB) values.
+
+    Returns:
+    dict: A dictionary where keys are worker node names and values are another dictionary
+          containing 'cpu' (int, in millicores) and 'mem' (float, in GB).
+    """
+    cmd = ["oc", "get", "nodes", "-l", "node-role.kubernetes.io/worker", "-o", "yaml"]
```
paraggit, line 3455: Can you please check with the command `oc adm top nodes`? It gives the CPU and memory utilization of all nodes in a single command.

avd-sagare: yes it will give used CPU and memory but we need to calculate what is requested cpu and memory bu pod because in as execution progress pod might use more cpu and memory than current. which will cause node hang like scenario.

ocs_ci/helpers/cnv_helpers.py
ebenahar: it would be more robust to fetch the root disk or, at the very least, save the size as a constant

ocs_ci/ocs/node.py
```diff
@@ -3417,3 +3418,87 @@ def select_osd_node():
...
+    result = run_cmd(cmd)
```
ebenahar, line 3456: I see that `node.status.allocatable` returns also `pods` - the maximum number of Pods that can run on the node. Wouldn't it be simpler to use that only?

ocs_ci/ocs/node.py
```diff
@@ -3417,3 +3418,87 @@ def select_osd_node():
...
+
+            # CPU conversion
+            if cpu_req.endswith("m"):  # e.g. "500m"
+                cpu_val = int(cpu_req[:-1]) / 1000
+            elif cpu_req:
+                cpu_val = int(cpu_req)
+            else:
+                cpu_val = 0
+
+            # Memory conversion
```
ebenahar, line 3511: can this function help? https://github.com/red-hat-storage/ocs-ci/blame/master/ocs_ci/utility/utils.py#L3779

avd-sagare: No i guess because function suggested by you converts MiB/GiB to MB/GB. I am here converting MiB/KiB to GiB

ocs_ci/ocs/node.py
avd-sagare: Sorry. I didn't get this if i understand correctly you are saying take number of pods from node.status.allocatable, but here I am collecting cpu and RAM information of each worker node only.

ebenahar: what I meant is that it seems like instead of calculating based on compute allocatable numbers, you can use the calculation done automatically and provides the number of allocatable pods directly. But feel free to ignore as the method of caluclating based on CPU and memory is just as good

ocs_ci/ocs/node.py
ebenahar: ok

ocs_ci/ocs/node.py
avd-sagare: Yeah I think lets ignore because  method of caluclating based on CPU and memory is just as good

paraggit (approved):

PrasadDesala (approved):

Issue #12753, opened 2025-07-22
[Bug]: KeyError 'selector' while creating DC pod with replica_count > 1 using create_pod() in test_verify_rwo_using_dc_pod.py


paraggit:
### Description
The test test_verify_rwo_using_dc_pod.py fails with a KeyError: 'selector' while trying to create a DC pod with replica count set to 5. This occurs during the setup phase of the test when the create_pod() helper is invoked with deployment=True. The underlying issue appears to be that the fedora DC YAML template (FEDORA_DC_YAML) used in this path lacks the spec.selector.matchLabels field, which is required by the code.

### Steps to Reproduce
	1.	Run the test: tests.functional.pv.pv_services.test_verify_rwo_using_dc_pod.py
	2.	Use interface 'CephBlockPool' and set replica_count=5
	3.	Observe the failure during setup while invoking create_pod() with deployment=True

### Actual Behavior
```
Test Failed With Following Error:
E           KeyError: 'selector'
```
### Expected Behavior
The pod (DC) should be created successfully with the specified replica count using the create_pod() helper function, and the test should proceed without encountering a KeyError.
### Environment
- Test Suite(s):
- Platform(s):
- Version(s):
- OS:

PR #13230, opened 2025-09-17
Fixed keyerror in the test test_verify_rwo_using_replicated_pod

paraggit:
This is Fix for the issue : https://github.com/red-hat-storage/ocs-ci/issues/12753 
Root Cause Identified

The test was failing with KeyError: 'selector' because the DeploymentConfig template  is being use and it doesn't has the "selector" key. So i have changed the yaml to use  Deployment instead of DeploymentConfig


paraggit:
These required  changes are addressed in another PR

PR #13263, opened 2025-09-23
Use prefix for the imported hosted clusters in ACM hub

jilju:
Use prefix for the discovered hosted clusters that will be automatically imported to in ACM hub. Current prefix is empty.
This change is based on a new node added in https://github.com/stolostron/hypershift-addon-operator/blob/main/docs/discovering_hostedclusters.md#configure-naming-convention
The note is
```
Important: Using an empty string as the custom prefix is not recommended as it can cause klusterlet naming collisions within the MCE cluster.
```

### DIFF
--- a/ocs_ci/ocs/acm/acm.py
+++ b/ocs_ci/ocs/acm/acm.py
@@ -873,7 +873,7 @@ def discover_hosted_clusters():
         resource_name="hypershift-addon-deploy-config",
         params=(
             '{"spec":{"customizedVariables":[{"name":"disableMetrics","value": "true"},'
-            '{"name":"disableHOManagement","value": "true"},{"name":"discoveryPrefix","value": "dr-"}]}}'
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/acm/acm.py
```diff
@@ -873,7 +873,7 @@ def discover_hosted_clusters():
         resource_name="hypershift-addon-deploy-config",
         params=(
             '{"spec":{"customizedVariables":[{"name":"disableMetrics","value": "true"},'
-            '{"name":"disableHOManagement","value": "true"},{"name":"discoveryPrefix","value": ""}]}}'
+            '{"name":"disableHOManagement","value": "true"},{"name":"discoveryPrefix","value": "dr"}]}}'
```
prsurve: it should be **dr-**

PR #13266, opened 2025-09-23
Skip acm brew for clients and check the presence of ENV_DATA.cluster_type if exists

jilju:
Use ENV_DATA.cluster_type to identify hcp cluster. Update the usage of cluster_type parameter considering this parameter may not exists in all configs.


jilju:
Verified: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-rdr-setup/2758/console
(the failure at a later stage is due to a missing config value)

PR #13283, opened 2025-09-26
4.18-correct-storage-client-name

DanielOsypenko:
correct storage client name; align it with 4.19

### DIFF
--- a/ocs_ci/deployment/hosted_cluster.py
+++ b/ocs_ci/deployment/hosted_cluster.py
@@ -913,7 +913,7 @@ class HostedODF(HypershiftHostedOCP):
 
             storage_client_data["spec"]["onboardingTicket"] = onboarding_key
 
-            self.storage_client_name = storage_client_data["metadata"]["name"]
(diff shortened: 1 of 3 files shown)


jilju (approved):

suchita-g (approved):

ocs_ci/deployment/hosted_cluster.py
```diff
@@ -913,7 +913,7 @@ def _apply_storage_client_cr():
 
             storage_client_data["spec"]["onboardingTicket"] = onboarding_key
 
-            self.storage_client_name = storage_client_data["metadata"]["name"]
```
fbalak, line 916: Why is this changed? Could we deploy storage client with a different name?

DanielOsypenko: we need to choose where to get name for storage-client. Ideally we should have one source (template or constants). This is visible especially now, when we had to change it to "ocs-storagecluster" - w/a for upgrade issue. I had to find multiple sources for that name (bad initial design). 
Considering that the template is more hidden way I thought better to have a constant name.

PR #13297, opened 2025-09-29
Open ports on IBM Cloud Hub cluster - utility funcs

DanielOsypenko:
This PR automates a portion HowTo document [Deploying provider cluster on IBM cloud and external client cluster]( https://url.corp.redhat.com/cc68443). 
Opening ports is a necessary part on IBM Cloud cluster to make it available to external spoke clusters (design known as Provider/Client)

Testing  - https://url.corp.redhat.com/08fe650

### DIFF
--- a/ocs_ci/utility/ibmcloud.py
+++ b/ocs_ci/utility/ibmcloud.py
@@ -796,7 +796,7 @@ class IBMCloudIPI(object):
             force (bool): True for force instance stop, False otherwise
             timeout (int): Timeout for the command, defaults to 300 seconds.
         """
-        logger.info(f"Stopping instances {list(node.name for node in nodes )}")
+        logger.info(f"Stopping instances {list(node.name for node in nodes)}")
         self.stop_nodes(nodes=nodes, force=force)
         if wait:
             for node in nodes:
@@ -808,7 +808,7 @@ class IBMCloudIPI(object):
                     node_status=constants.STATUS_STOPPED,
                 )
                 sample.wait_for_func_status(result=True)
-        logger.info(f"Starting instances {list(node.name for node in nodes )}")
+        logger.info(f"Starting instances {list(node.name for node in nodes)}")
 
         self.start_nodes(nodes=nodes)
         if wait:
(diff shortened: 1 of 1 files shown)


ocs_ci/utility/ibmcloud.py
```diff
@@ -1166,18 +1166,147 @@ def create_vpc(cluster_name, resource_group):
     )
 
 
-def get_used_subnets():
+def get_used_subnets(vpc_id=""):
     """
     Get currently used subnets in IBM Cloud
 
+    Args:
+        vpc_id (str): VPC ID to filter subnets, if empty it will. Empty string means all VPCs.
```
dahorak:
```suggestion
        vpc_id (str): VPC ID to filter subnets. Empty string means all VPCs.
```

ocs_ci/utility/ibmcloud.py
```diff
@@ -1166,18 +1166,147 @@ def create_vpc(cluster_name, resource_group):
...
+    """
+    Get security group ID by name
+
+    Args:
+        sg_name (str): security group name
+        vpc_id (str): VPC ID to filter security groups, if empty it will return security groups for all VPCs.
+        resource_group_id (str): Resource group ID to filter security groups,
+            if empty it will return security groups for all resource groups.
+        resource_group_name (str): Resource group name to filter security groups,
+            if empty it will return security groups for all resource groups.
```
dahorak: IIUC, the `vpc_id`, `resource_group_id` and `resource_group_name` parameters here just limit the number of returned security groups from the `ibmcloud` command,  which is then filtered for the particular `sg_name`. In other words, usage of this parameters didn't change the returned value, but have performance impact, right? If my understanding is correct, could you please update the description of those parameters, so it will not suggest that `...if empty it will return security groups for all ...`?

ocs_ci/utility/ibmcloud.py
```diff
@@ -1166,18 +1166,147 @@ def create_vpc(cluster_name, resource_group):
...
+    Add security group rule
+
+    Args:
+        security_group (str): security group ID or Name
+        direction (str): inbound or outbound
+        protocol (str): protocol, e.g. tcp, udp, icmp, all
+        port_min (int): minimum port number
+        port_max (int): maximum port number
+        **kwargs: other arguments to be passed to command, e.g.
+            --vpc ID or name of the VPC. It is required to specify only the unique resource by name inside this VPC.
```
dahorak, line 1246: The list here should end up with one empty line.

ocs_ci/utility/ibmcloud.py
```diff
@@ -1166,18 +1166,147 @@ def create_vpc(cluster_name, resource_group):
...
+    """
+    Get security group name by pattern
+
+    Args:
+        sg_name_pattern (str): security group name pattern (regular expression)
+        vpc_id (str): VPC ID to filter security groups, if empty it will return security groups for all VPCs.
+        resource_group_id (str): Resource group ID to filter security groups,
+            if empty it will return security groups for all resource groups.
+        resource_group_name (str): Resource group name to filter security groups,
+            if empty it will return security groups for all resource groups.
```
dahorak: Similarly as above.

ocs_ci/utility/ibmcloud.py
DanielOsypenko: `resource_group_name` and others are filters. If argument is not passed then the function will return all security group id's (from all `resource_group_name`(s))
The filtering happens via ibmcloud parameters, so passing any argument is reducing the number of security groups and if not passed it will return all w/o filtering. 

Could you pls clarify q, because I don't see inaccuracy in 
> "if empty it will return security groups for all"

ocs_ci/utility/ibmcloud.py
DanielOsypenko: f argument is not passed then the function will return all security_group_name(s) that are matching pattern. Filtering with `resource_group_id` and other args will not take place. Similarly to above `get_security_group_id`.

Could you pls clarify q, because I don't see inaccuracy

ocs_ci/utility/ibmcloud.py
dahorak: I meant, that this function is directly doing the filtering based on the passed `sg_name`, so in all cases it will return only one ID. The `vpc_id`, `resource_group_id` and `resource_group_name` only limit the list of security groups returned by the `ibmcloud` command which is then searched for the `sg_name` (so it is only internal change inside the function with performance effect, but do not affect the returned value).

So the  `...if empty it will return security groups for all...` is valid for the `get_security_groups` function, but not for `get_security_group_id` and `get_security_group_name_by_pattern` functions, because those two functions always return only one security ID (or empty string if not found).

ocs_ci/utility/ibmcloud.py
dahorak: See my explanation above.

ocs_ci/utility/ibmcloud.py
DanielOsypenko: ack ) I meant "it will return" ->  "ibmcloud cli" call will return. Confusing comment, you're right.

DanielOsypenko: removed confusing line from the docstring

ocs_ci/utility/ibmcloud.py
DanielOsypenko: removed confusing line from the docstring

PR #13306, opened 2025-10-01
Remove OCP MG image from default config

petr-balogh:
Fixes: #13308


petr-balogh:
Verified here:
https://url.corp.redhat.com/b25a69e

dahorak (approved):

PR #13327, opened 2025-10-06
WIP: Add support for simulating Ceph BlueStore label stamping on OSD disks

yitzhak12:
This change enables automated testing of BlueStore label simulation on test disks in a baremetal environment. It provides a safe and controlled way to mimic BlueStore disk preparation, useful for validating Ceph OSD detection logic without impacting real data.

This PR introduces a utility for simulating the stamping of a BlueStore label on a block device, primarily for testing and validation purposes. It includes:

**A new shell script `simulate_bluestore_label.sh` that:**

- Verifies the target device is empty (optional)
- Validates the device type and readiness
- Wipes existing signatures and partition tables
- Writes a mock BlueStore label with a generated UUID
- Verifies the label was written correctly 

**Two helper functions in baremetal.py:**

- `download_script_to_node`: Downloads a script to a worker node using oc debug
- `run_script_on_node`: Executes a script on a worker node using oc debug

**A new test case TestSimulateCephBlueStoreLabel in `test_simulate_ceph_bluestore_label.py` that:**

- Downloads the script to a worker node
- Executes the script on a specified device
- Asserts successful label verification

### DIFF
--- a/ocs_ci/deployment/baremetal.py
+++ b/ocs_ci/deployment/baremetal.py
@@ -1493,6 +1493,96 @@ def clean_disk(worker, namespace=constants.DEFAULT_NAMESPACE):
                 logger.info(out)
 
 
+def download_script_to_node(
+    worker,
+    script_url: str,
+    script_path: str,
+    namespace: str = "default",
+    timeout: int = 300,
+):
+    """
+    Download a shell script directly on a node (inside chroot /host)
+    using curl via oc debug.
+
+    Assumes the URL is valid and directly downloadable
+    (e.g., a raw GitHub or internal script URL).
+
+    Args:
+        worker (object): Worker node object.
+        script_url (str): Direct URL to the script.
+        script_path (str): Destination path on the node (default: /tmp/simulate_bluestore_label.sh).
+        namespace (str): Namespace for the debug pod.
+        timeout (int): Timeout for the oc debug command (in seconds).
+
+    Returns:
+        str: Output of the command execution.
+    """
+    ocp_obj = ocp.OCP()
+
+    # Single, atomic command string
+    download_cmd = [
+        (
+            f"set -euo pipefail; "
+            f'curl -fsSL "{script_url}" -o "{script_path}" && '
+            f'chmod 755 "{script_path}" && '
+            f'echo "Downloaded to {script_path}" && ls -l "{script_path}"'
+        )
+    ]
+
+    out = ocp_obj.exec_oc_debug_cmd(
+        node=worker.name,
+        cmd_list=download_cmd,
+        namespace=namespace,
+        use_root=True,
+        timeout=timeout,
+    )
+
+    return out
+
+
+def run_script_on_node(
+    worker,
+    script_path: str,
+    args: str = "",
+    namespace: str = "default",
+    timeout: int = 600,
+):
+    """
+    Execute a shell script directly on a node (inside chroot /host)
+    using oc debug.
+
+    Args:
+        worker (object): Worker node object.
+        script_path (str): Full path to the script on the node.
+        args (str): Optional arguments to pass to the script.
+        namespace (str): Namespace for the debug pod.
+        timeout (int): Timeout for the command execution (in seconds).
+
+    Returns:
+        str: Output of the script execution.
+    """
+    ocp_obj = ocp.OCP()
+
+    # Build the command — single element, no single quotes inside
+    run_cmd = [
+        (
+            f"set -euo pipefail; "
+            f'echo ">>> Running: {script_path} {args}"; '
+            f'bash "{script_path}" {args}'
+        )
+    ]
+
+    out = ocp_obj.exec_oc_debug_cmd(
+        node=worker.name,
+        cmd_list=run_cmd,
+        namespace=namespace,
+        use_root=True,
+        timeout=timeout,
+    )
+
+    return out
+
+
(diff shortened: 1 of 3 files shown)


ocs_ci/deployment/baremetal.py
```diff
@@ -1543,6 +1543,222 @@ def clean_disk(
...
+    Args:
+        wnode (ocs_ci.ocs.resources.ocs.OCS): The worker node object.
+        namespace (str): Namespace for the debug pod.
+        timeout (int): Timeout for the command execution.
+
+    Returns:
+        str or None: The detected disk path (e.g., "/dev/sdb") or None if not found.
+
+    """
+    logger.info("Attempting to auto-detect a suitable /dev/sd* disk.")
```
petr-balogh:
```suggestion
    logger.info(f"Attempting to auto-detect a suitable /dev/sd* disk on worker node: {wnode.name}.")
```
Is it worth to mention a worker node here?

ocs_ci/deployment/baremetal.py
```diff
@@ -1543,6 +1543,222 @@ def clean_disk(
...
+        namespace=namespace,
+        use_root=True,
+        timeout=timeout,
+    )
+
+    logger.info(f"Disk detection command output: {out}")
+    disk_name = out.strip()
+
+    if not disk_name:
+        logger.warning("No suitable disk found for BlueStore simulation.")
```
petr-balogh: also maybe worth to include worker node

ocs_ci/deployment/baremetal.py
```diff
@@ -1543,6 +1543,222 @@ def clean_disk(
...
+    if not disk_name:
+        disk_name = detect_simulation_disk_on_node(wnode, namespace, timeout=300)
+
+    if not disk_name:
+        logger.error("Disk detection failed. Aborting BlueStore simulation.")
+        return False
+
+    # Step 2: Define script location and source
+    script_path = "/tmp/simulate_bluestore_label.sh"
+    url_prefix = "https://raw.githubusercontent.com/yitzhak12/ocs-ci/add-simulate-ceph-bluestore-label/"
```
petr-balogh: Not good idea to use and download it from here or even from normal github URL as this will only use master branch for example and not possible to test.

We should maybe use somehow exec_oc_debug_cmd function to from local content print it to file somehow if possible, we will need to check but don't have time to do now. 

Or we should know the IP address of node, so maybe we can scp the file via ssh using openshift-dev.pem key. This might be also an option.

yitzhak12: The problem is that until the branch is merged, I can't take it from master. I thought of changing it after merging to the master branch. I tried to copy-paste the file content using the `exec_oc_debug_cmd` method, but it turned out to be too complicated. That's why I preferred to download the file.

yitzhak12: I have changed the code to take the script from the master branch.

yitzhak12: I have changed the method to use base64 to download the script.

ocs_ci/deployment/baremetal.py
```diff
@@ -1543,6 +1545,243 @@ def clean_disk(
...
+        wnode=wnode,
+        script_src_path=script_src_path,
+        script_dest_path=script_dest_path,
+        namespace=namespace,
+        timeout=300,
+    )
+
+    # Step 4: Verify script integrity using SHA256
+    logger.info("Verifying the script's SHA256 checksum")
+    script_expected_sha256sum = (
```
petr-balogh: Same here - this should be calculated also dynamically from current script in function `upload_script_to_node`
And can be checked after upload there as well.  But we I think don't need to be that scary about it and we can even avoid this check - but if you want to keep it, move it to function please and make it dynamic.

Which means, every time you would like to do some change to script you need to maintain this checksum here == doesn't make much sense to me.

ocs_ci/deployment/baremetal.py
```diff
@@ -1543,6 +1545,243 @@ def clean_disk(
...
+
+    if not disk_name:
+        logger.error("Disk detection failed. Aborting BlueStore simulation.")
+        return False
+
+    script_name = "simulate_bluestore_label.sh"
+    current_dir = Path(__file__).parent.parent.parent
+    script_src_path = os.path.join(current_dir, "scripts", "bash", script_name)
+    if not os.path.exists(script_src_path):
+        raise FileNotFoundError(f"Script not found at {script_src_path}")
```
petr-balogh: Lets define it in constants:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/constants.py#L18

We have TOP_DIR defined there. 

So check how we referring other directories and files using it there. So define constant with script path in constants using same mechanism.

Not sure you really need to check file existence here. I would rather move `if not os.path.exists(script_src_path):`

To the function upload_script_to_node itself ;) which makes more sense to me to check there.

ocs_ci/deployment/baremetal.py
yitzhak12: I removed the checksum verification, as we only copy-paste the file.

ocs_ci/deployment/baremetal.py
yitzhak12: I moved it to `upload_script_to_node`.

ebenahar (approved):

PR #13339, opened 2025-10-08
Fixing issue reconcile disableLoadBalancerService to True

posharma1:
If test execution platform is Ibmcloud managed then need to reconcile disableLoadBalancerService to True via storage cluster else for other platforms should work with noobaa cr

### DIFF
--- a/tests/cross_functional/kcs/test_disable_mcg_external_service.py
+++ b/tests/cross_functional/kcs/test_disable_mcg_external_service.py
@@ -26,12 +26,26 @@ class TestDisableMCGExternalService:
             resource_name="noobaa",
         )
 
-        # patch noobaa object
-        noobaa_ocp_obj.patch(
-            resource_name="noobaa",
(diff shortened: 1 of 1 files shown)


tests/cross_functional/kcs/test_disable_mcg_external_service.py
```diff
@@ -26,13 +26,27 @@ def patch_noobaa_object(self, request):
...
-            params='{"spec": {"disableLoadBalancerService": true }}',
-            format_type="merge",
+        # get storagecluster object
+        storagecluster_obj = OCP(
+            kind="storagecluster",
+            namespace=config.ENV_DATA["cluster_namespace"],
+            resource_name="ocs-storagecluster",
         )
 
+        if config.ENV_DATA["deployment_type"].lower() == "managed":
```
PrasadDesala, line 36: Given Dev's recommendation to patch the NooBaa CR rather than the StorageCluster CR, as discussed please implement this change for all deployments, not just managed ones

PR #13350, opened 2025-10-10
Ading new config with odf-qe vpc on IBM Cloud

shivamdurgbuns:
Ading new config with odf-qe vpc on IBM Cloud

### DIFF
--- a/ocs_ci/deployment/helpers/external_cluster_helpers.py
+++ b/ocs_ci/deployment/helpers/external_cluster_helpers.py
@@ -195,7 +195,7 @@ class ExternalCluster(object):
             self.user,
             self.password,
             self.ssh_key,
-            jump_host_connection=self.rhcs_conn if self.jump_host else None,
+            ssh_connection=self.rhcs_conn if self.jump_host else None,
         )
         return script_path
 
@@ -216,7 +216,7 @@ class ExternalCluster(object):
             self.user,
             self.password,
             self.ssh_key,
-            jump_host_connection=self.rhcs_conn if self.jump_host else self.ssh_key,
+            ssh_connection=self.rhcs_conn if self.jump_host else self.ssh_key,
         )
         return remote_rgw_cert_ca_path
 
@@ -352,7 +352,7 @@ class ExternalCluster(object):
             self.user,
             self.password,
             self.ssh_key,
-            jump_host_connection=self.rhcs_conn if self.jump_host else None,
+            ssh_connection=self.rhcs_conn if self.jump_host else None,
         )
         return script_path
 
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -3212,7 +3212,7 @@ def upload_file(
     user=None,
     password=None,
     key_file=None,
-    jump_host_connection=None,
+    ssh_connection=None,
 ):
     """
     Upload a file to remote server
@@ -3222,13 +3222,16 @@ def upload_file(
         localpath (str): Local file to upload
         remotepath (str): Target path on the remote server. filename should be included
         user (str): User to use for the remote connection
+        password (str): Password to use for the remote connection
+        key_file (str): Key file to use for the remote connection
+        ssh_connection (SSHClient): SSH connection to use for the remote connection
 
     """
     if not user:
         user = "root"
     try:
-        if jump_host_connection:
-            sftp = jump_host_connection.client.open_sftp()
+        if ssh_connection:
+            sftp = ssh_connection.client.open_sftp()
             log.info(f"uploading {localpath} to {user}@{server}:{remotepath}")
             sftp.put(localpath, remotepath)
             sftp.close()


ocs_ci/deployment/ibmcloud.py
```diff
@@ -774,3 +783,182 @@ def prepare_custom_vpc_and_network(self):
...
+                f"{', '.join(instance_names)}"
+            )
+
+            return instance_ids
+
+        except Exception as e:
+            logger.error(f"Failed to retrieve instances by prefix '{prefix}': {e}")
+            return []
+
+    def add_security_group_to_vsi(self, instance_id, security_group_name="odf-qe-sg"):
```
petr-balogh: security_group_name have this an None by default,

And in start of function read if from config if it's None.

```
if not security_group_name:
    security_group_name = config.ENV_DATA['custom_vpc_security_group_name']
```

ocs_ci/deployment/ibmcloud.py
```diff
@@ -774,3 +783,182 @@ def prepare_custom_vpc_and_network(self):
...
+            security_group = next(
+                (sg for sg in security_groups if sg["name"] == security_group_name),
+                None,
+            )
+
+            if not security_group:
+                logger.error(f"Security group '{security_group_name}' not found")
+                return False
+
+            new_security_group_id = security_group["id"]
```
petr-balogh:
```suggestion
            custom_vpc_security_group_id = security_group["id"]
```

ocs_ci/deployment/ibmcloud.py
```diff
@@ -774,3 +783,182 @@ def prepare_custom_vpc_and_network(self):
...
+            proc = exec_cmd(cmd)
+            network_interfaces = json.loads(proc.stdout)
+
+            if not network_interfaces:
+                logger.error(f"No network interfaces found for instance: {instance_id}")
+                return False
+
+            # Get the primary network interface (usually the first one)
+            primary_nic = network_interfaces[0]
+            nic_id = primary_nic["id"]
```
petr-balogh:
```suggestion
            primary_nic_id = primary_nic["id"]
```

shivamdurgbuns:
Verification jobs: 

1. https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-destroy-ocs-cluster/67314/console
2. https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/60639/console

ocs_ci/deployment/ibmcloud.py
```diff
@@ -774,3 +790,705 @@ def prepare_custom_vpc_and_network(self):
...
                     subnet_name, gateway_name, vpc_name
                 )
+
+    def prepare_existing_vpc_and_network(self):
+        """
+        Prepare to use existing VPC, resource group, and subnets for IBM Cloud IPI deployment.
+        This function allows you to use your own pre-existing VPC infrastructure.
+
+        Required ENV_DATA configuration:
+        - existing_vpc: true
```
petr-balogh, line 809: Make sure we have those values documented in:
https://github.com/red-hat-storage/ocs-ci/blob/master/conf/README.md

ocs_ci/deployment/ibmcloud.py
```diff
@@ -774,3 +790,705 @@ def prepare_custom_vpc_and_network(self):
...
+                elif resource_type == "load-balancer":
+                    cmd = f"ibmcloud is load-balancer {resource_id} --output json"
+                elif resource_type == "volume":
+                    cmd = f"ibmcloud is volume {resource_id} --output json"
+                elif resource_type == "security-group":
+                    cmd = f"ibmcloud is security-group {resource_id} --output json"
+                elif resource_type == "floating-ip":
+                    cmd = f"ibmcloud is floating-ip {resource_id} --output json"
+                else:
+                    return True
```
petr-balogh: We should rather raise an exception of unsupported resource type here

ocs_ci/deployment/ibmcloud.py
```diff
@@ -774,3 +790,705 @@ def prepare_custom_vpc_and_network(self):
...
+            except CommandFailed:
+                return True
+            if attempt < max_attempts:
+                time.sleep(wait_time)
+        logger.warning(
+            f"Timeout waiting for {resource_type} {resource_id} to be deleted"
+        )
+        return False
+
+    @retry(CommandFailed, tries=3, delay=5, backoff=2)
```
petr-balogh: You don't need wrapper function but directly call retry like here for example:
https://github.com/red-hat-storage/ocs-ci/blob/4f6240d3bacbd7075c338503af4523b65c46e672/ocs_ci/utility/lvmo_utils.py#L67C5-L72C32

petr-balogh (dismissed):

conf/README.md
```diff
@@ -378,6 +378,14 @@ higher priority).
...
   with specific Address prefixes to prevent /18 CIDR to be used.
 * `ip_prefix` - Applicable only for IMB Cloud IPI deployment when custom_vpc, if not specified: 27 prefix will be used.
+* `existing_vpc` - Set to true to use existing VPC, resource group, and subnets for IBM Cloud IPI deployment.
+* `resource_group_name` - Name of existing resource group for IBM Cloud IPI deployment when using existing VPC.
+* `network_resource_group_name` - Name of existing network resource group for IBM Cloud IPI deployment when using existing VPC (can be same as resource_group_name).
+* `vpc_name` - Name of existing VPC for IBM Cloud IPI deployment when using existing VPC.
+* `control_plane_subnets` - List of existing control plane subnet names for IBM Cloud IPI deployment when using existing VPC.
+* `compute_subnets` - List of existing compute subnet names for IBM Cloud IPI deployment when using existing VPC.
+*  `worker_instance_type` - Worker instance type in ibmcloud; example: 'bx2-16x64'
+*  `master_instance_type` - Worker instance type in ibmcloud; example: 'bx2-4x16'
```
petr-balogh: I see one extra space at start

ocs_ci/deployment/helpers/external_cluster_helpers.py
```diff
@@ -71,11 +72,20 @@ def __init__(self, host, user, password=None, ssh_key=None):
                 "No SSH Auth to connect to external RHCS cluster provided! "
                 "Either password or SSH key is missing in EXTERNAL_MODE['login'] section!"
             )
+        # adding jump host configuration to connect to external RHCS cluster on ibmcloud via jump host
+        self.jump_host = config.DEPLOYMENT.get("ssh_jump_host", {})
```
petr-balogh:
```suggestion
        self.jump_host = config.DEPLOYMENT.get("ssh_jump_host", None)
```

ocs_ci/deployment/helpers/external_cluster_helpers.py
```diff
@@ -71,11 +72,20 @@ def __init__(self, host, user, password=None, ssh_key=None):
...
+
+        if self.jump_host:
+            self.jump_host["private_key"] = self.jump_host.get(
+                "private_key"
+            ) or os.path.expanduser(config.DEPLOYMENT["ssh_key_private"])
+
```
petr-balogh:
```suggestion
        if self.jump_host and not self.jump_host.get("private_key") :
            self.jump_host["private_key"] = os.path.expanduser(config.DEPLOYMENT["ssh_key_private"])

```

ocs_ci/deployment/helpers/external_cluster_helpers.py
```diff
@@ -71,11 +72,20 @@ def __init__(self, host, user, password=None, ssh_key=None):
...
         self.rhcs_conn = Connection(
             host=self.host,
             user=self.user,
             password=self.password,
             private_key=self.ssh_key,
+            jump_host=None if not self.jump_host else self.jump_host,
```
petr-balogh:
```suggestion
            jump_host=self.jump_host,
```

ocs_ci/deployment/helpers/external_cluster_helpers.py
```diff
@@ -193,14 +219,25 @@ def upload_rgw_cert_ca(self):
...
+            )
+        else:
+            upload_file(
+                self.host,
+                rgw_cert_ca_path,
+                remote_rgw_cert_ca_path,
+                self.user,
+                self.password,
+                self.ssh_key,
+            )
```
petr-balogh:
```suggestion
            upload_file(
                self.host,
                rgw_cert_ca_path,
                remote_rgw_cert_ca_path,
                self.user,
                self.password,
                self.ssh_key,
                jump_host_connection=self.rhcs_conn if self.jump_host else None,,
            )
```

ocs_ci/deployment/helpers/external_cluster_helpers.py
```diff
@@ -328,9 +365,25 @@ def upload_config_ini_file(self, params):
...
+        else:
+            upload_file(
+                self.host,
+                script_path,
+                script_path,
+                self.user,
+                self.password,
+                self.ssh_key,
+            )
```
petr-balogh: Same as above

ocs_ci/deployment/ibmcloud.py
```diff
@@ -774,3 +791,691 @@ def prepare_custom_vpc_and_network(self):
...
+        logger.info(f"Using existing control plane subnets: {control_plane_subnets}")
+        logger.info(f"Using existing compute subnets: {compute_subnets}")
+
+        # Set the configuration for the installer
+        config.ENV_DATA["existing_vpc"] = True
+        config.ENV_DATA["resource_group_name"] = resource_group
+        config.ENV_DATA["network_resource_group_name"] = network_resource_group
+        config.ENV_DATA["vpc_name"] = vpc_name
+        config.ENV_DATA["control_plane_subnets"] = control_plane_subnets
+        config.ENV_DATA["compute_subnets"] = compute_subnets
```
petr-balogh: Is it needed? I guess not as those values are provided directly via config and are loaded , no?

ocs_ci/deployment/ibmcloud.py
```diff
@@ -774,3 +791,691 @@ def prepare_custom_vpc_and_network(self):
...
+        """
+        try:
+            logger.info(f"Fetching instances with prefix: {prefix}")
+            cmd = "ibmcloud is instances --output json"
+            proc = exec_cmd(cmd)
+            instances = json.loads(proc.stdout)
+
+            # Filter instances by name prefix
+            matching_instances = [
+                inst for inst in instances if inst.get("name", "").startswith(prefix)
```
petr-balogh, line 873: This might be danger if some cluster name is subset of some other. If we know there is always also some suffix like `-xxxxx`. Better to update here to remove last X characters from name and compare on `==` .

The same on other occurrences of `startswith(prefix)` in this PR

ocs_ci/deployment/helpers/external_cluster_helpers.py
```diff
@@ -178,9 +188,25 @@ def upload_exporter_script(self):
         if ocs_version <= version.VERSION_4_18:
             use_configmap = False
         script_path = generate_exporter_script(use_configmap=use_configmap)
-        upload_file(
-            self.host, script_path, script_path, self.user, self.password, self.ssh_key
-        )
+        if self.jump_host:
```
petr-balogh: Same as below comment

ocs_ci/deployment/ibmcloud.py
```diff
@@ -191,6 +194,13 @@ def deploy_ocp(self, log_cli_level="DEBUG"):
             "oc get clusterversion version -o jsonpath='{.spec.clusterID}'"
         )
         logger.info(f"clusterID (UUID): {cluster_id}")
+        # adding odf-qe security group to the instances
+        if config.ENV_DATA.get("existing_vpc"):
+            instance_names = self.get_instance_names_by_prefix(
+                f"{config.ENV_DATA['cluster_name']}"
```
petr-balogh: Here we are using only culster_name , so if one cluster name is subset of the other we will be getting instances from other cluster here.

ocs_ci/deployment/ibmcloud.py
```diff
@@ -774,3 +791,685 @@ def prepare_custom_vpc_and_network(self):
...
+        Required ENV_DATA configuration:
+        - existing_vpc: true
+        - resource_group_name: name of existing resource group
+        - network_resource_group_name: name of existing network resource group (can be same as resource_group_name)
+        - vpc_name: name of existing VPC
+        - control_plane_subnets: list of existing control plane subnet names
+        - compute_subnets: list of existing compute subnet names
+        """
+        cluster_id = get_random_str(size=5)
+        config.ENV_DATA["cluster_id"] = cluster_id
```
petr-balogh, line 817: I think this might be a problem as this cluster_id is populated only here when we configure VPC - means in deployment only.  So this is not injected anyhow into the cluster name. And this cluster_id is just randomly generated here and stored in ENV_DATA only here.

But the function destroy_cluster_from_existing_vpc
Is running in the teardown. Which is the new run of run-ci , for destroy only, this means we are not running there prepare_existing_vpc_and_network which is setting config.ENV_DATA["cluster_id"]  , and even though we will run it again, it will generate new random ID which will not match for what we have defined in different run for deployment.


In destroy function you are using:
        prefix = (
            f"{config.ENV_DATA.get('cluster_name')}-{config.ENV_DATA.get('cluster_id')}"
        )

In that case cluster_id will not be defined and will lead to `cluster_name-None`

Which is not intended here.

shivamdurgbuns:
latest verification job: https://url.corp.redhat.com/32d801f

ocs_ci/utility/utils.py
```diff
@@ -3334,7 +3205,15 @@ def check_if_executable_in_path(exec_name):
...
 
-def upload_file(server, localpath, remotepath, user=None, password=None, key_file=None):
+def upload_file(
+    server,
+    localpath,
+    remotepath,
+    user=None,
+    password=None,
+    key_file=None,
+    jump_host_connection=None,
```
dahorak: Please document this argument in the docstring (looks like, we are missing there also the `password` and `key_file` parameters, so please if you can include them too).

If I understand it correctly, the `jump_host_connection` is "an existing ssh connection" which is used for the sftp transfer, right? In that case, I would propose to change the name to something like `ssh_connection` or something like that (because it actually could be used for any connection, not only through jump host).

ocs_ci/deployment/helpers/external_cluster_helpers.py
```diff
@@ -330,7 +346,13 @@ def upload_config_ini_file(self, params):
...
         script_path = create_config_ini_file(params=params)
         upload_file(
-            self.host, script_path, script_path, self.user, self.password, self.ssh_key
+            self.host,
+            script_path,
+            script_path,
+            self.user,
+            self.password,
+            self.ssh_key,
+            jump_host_connection=self.rhcs_conn if self.jump_host else None,
```
dahorak: If I understanding it correctly (see the other comment under the `upload_file` function definition), we can use the existing ssh connection always, no matter if there is a jump host or not. So we could pass there only the connection (`self.rhcs_conn`) and remove the user, password and ssh_key.

PR #13355, opened 2025-10-10
Add --product-type CLI arg to choose ODF or FDF

clacroix12:


dahorak (approved):

petr-balogh:
please cherry-pick and update the job to propagate this to test stage into the run-ci in case of fdf execution. Thanks

PR #13365, opened 2025-10-13
add multus support for existing cluster

subhamkrai:


tests/functional/z_cluster/test_enable_multus_existing_cluster.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+    Test class for enabling Multus on an existing cluster.
+    """
+
+    @tier2
+    def test_enable_multus_on_existing_cluster(self):
+        """
+        Test enabling Multus networking on an existing OCS cluster.
+
+        Steps:
```
OdedViner, line 59: We need to install nmstate for vSphere. @shyRozen , correct me if I’m wrong. Nmstate installation instructions: https://github.com/red-hat-storage/ocs-ci/blob/763bd62012827f85d5ffaf4bcf0cfaa4f4131a07/ocs_ci/deployment/nmstate.py#L156

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
```diff
@@ -0,0 +1,244 @@
...
+
+            public_net_data = templating.load_yaml(nad_to_load)
+            public_net_data["metadata"]["name"] = config.ENV_DATA.get(
+                "multus_public_net_name"
+            )
+            public_net_data["metadata"]["namespace"] = config.ENV_DATA.get(
+                "multus_public_net_namespace"
+            )
+
+            public_net_config_str = public_net_data["spec"]["config"]
```
OdedViner, line 139: Since this feature is specific to vSphere and Bare Metal, we prefer not to add configuration files via `--ocsci-conf` to the `run-ci` command. Also, because we currently have only one Bare Metal setup, I think we should run it only on vSphere. We can add the `@vsphere_platform_required` marker to the test.

CC @petr-balogh @shyRozen

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
```diff
@@ -0,0 +1,250 @@
...
+        multus_cluster_net_interface = constants.VSPHERE_MULTUS_INTERFACE  # "ens224"
+        multus_cluster_net_range = "192.168.30.0/24"
+        multus_cluster_ipv6_net_range = "fd01:db8:2:1::/96"
+        multus_cluster_net_mode = "bridge"
+
+        # Create Public Network
+        if create_public_net:
+            logger.info("Creating Multus public network")
+            nad_to_load = constants.MULTUS_PUBLIC_NET_YAML
+            if config.DEPLOYMENT.get("ipv6"):
```
OdedViner: Can you remove ipv6?

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
```diff
@@ -0,0 +1,250 @@
...
+
+            public_net_data = templating.load_yaml(nad_to_load)
+            public_net_data["metadata"]["name"] = multus_public_net_name
+            public_net_data["metadata"]["namespace"] = multus_public_net_namespace
+
+            public_net_config_str = public_net_data["spec"]["config"]
+            public_net_config_dict = json.loads(public_net_config_str)
+            public_net_config_dict["master"] = multus_public_net_interface
+
+            if not config.DEPLOYMENT.get("ipv6"):
```
OdedViner: can remove ipv6? I think we need to test with ipv6 cluster @shyRozen correct?

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+@brown_squad
+@skipif_managed_service
+@skipif_external_mode
+@vsphere_platform_required
+@polarion_id(get_polarion_id())
```
OdedViner: Polarion-id is a number like https://github.com/red-hat-storage/ocs-ci/blob/8e5027681159ecf588c06be853e77887816e767d/tests/cross_functional/kcs/test_disable_mcg_external_service.py#L59

ebenahar: @subhamkrai please work with @shyRozen to create the polarion test cases. We can add them in a follow up PR to not block this one. Just please comment out this line

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
```diff
@@ -0,0 +1,272 @@
...
+        multus_public_net_interface = "enp1s0f1"
+        multus_public_net_range = "192.168.20.0/24"
+        multus_public_net_type = "macvlan"
+        multus_public_net_mode = "bridge"
+
+        multus_cluster_net_name = "private-net"
+        multus_cluster_net_namespace = "openshift-storage"
+        multus_cluster_net_interface = "enp1s0f1"
+        multus_cluster_net_range = "192.168.30.0/24"
+        multus_cluster_net_mode = "bridge"
```
ebenahar: please save these in https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/constants.py if don't exist there already

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
```diff
@@ -0,0 +1,272 @@
...
+        logger.info(f"Created CephFS PVC: {cephfs_pvc_obj.name}")
+
+        cephfs_pod_obj = pod_factory(
+            interface=constants.CEPHFILESYSTEM,
+            pvc=cephfs_pvc_obj,
+            status=constants.STATUS_RUNNING,
+        )
+        logger.info(f"Created pod with CephFS PVC: {cephfs_pod_obj.name}")
+
+        logger.info("Successfully enabled Multus on existing cluster")
```
ebenahar, line 299: need to disable Multus on the cluster at the end of the test to bring the cluster back to its original state. 
Please add that as a step in the test case and also as a teardown fixture, to ensure the cluster is brought back to be without Multus even in cases when the test ends/fails prematurely

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
```diff
@@ -0,0 +1,272 @@
...
+        logger.info("StorageCluster is Ready")
+
+        logger.info("Step 2: Verifying OSD pods are running")
+        wait_for_storage_pods()
+        osd_pods = get_osd_pods()
+        logger.info(f"All {len(osd_pods)} OSD pods are running")
+
+        sc = get_storage_cluster()
+
+        logger.info("Interface added to compute VMs successfully")
```
shyRozen: I don't see in the test any adding of an interface. If we follow Oded suggestion and going only with Vsphere then you should use this function
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/vsphere.py#L1850

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
shyRozen: Some of them are here https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L278

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
subhamkrai: @ebenahar can we do the `multus-to-non-multus` teardown in a separate PR as currently, the test is passing but teardown is failing. Also,  `multus-to-non-multus` is out-of-scope for my epic and changes.

subhamkrai:
```
2025-12-01 19:39:35  09:09:34 - MainThread - ocs_ci.framework.pytest_customization.reports - INFO  - duration reported by tests/functional/z_cluster/test_enable_multus_existing_cluster.py::TestEnableMultusExistingCluster::test_enable_multus_on_existing_cluster immediately after test execution: 646.8
2025-12-01 19:39:35  PASSED
```
@ebenahar Check above.

It will help proceed with PR and work on other item separately.

ebenahar: @subhamkrai if we merge this PR as is without disabling Multus at the end, it will have an effect on the subsequent test cases running in the tier2 suite, which we can't allow. Why is disabling multus failing? If It's due to a product bug then we will want to keep this test case skipped until the bug is fixed

subhamkrai: We have not tested, from `multus-to-non-multus` as this was not part of the epic. And the must-gather is failing so we can't debug the failures.

.secrets.baseline
```diff
@@ -477,7 +477,7 @@
         "hashed_secret": "de19c40310ad831a71524cd279e6f5f0577fb09a",
```
shyRozen, line 567: I don't think this file should be here.

ocs_ci/framework/conf/default_config.yaml
shyRozen, line 1: For some reason you deleting and adding the same lines although your change is minor.

OdedViner (dismissed):
LGTM

.secrets.baseline
```diff
@@ -567,7 +567,7 @@
...
         "is_secret": false,
         "is_verified": false,
-        "line_number": 370,
+        "line_number": 369,
```
OdedViner: What is this line?

subhamkrai: I think it is coming from rebase, I'll revert

OdedViner:
PR validation job pass https://url.corp.redhat.com/dc98cda

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
```diff
@@ -0,0 +1,361 @@
...
+
+
+@brown_squad
+@skipif_managed_service
+@skipif_external_mode
+@tier2
+@ignore_leftovers
+@pytest.mark.skipif(
+    config.ENV_DATA["platform"].lower() != constants.BAREMETAL_PLATFORM,
+    reason="Test runs ONLY on bare metal platform",
```
ebenahar, line 46: vSphere should be suitable as well. Don't want to block the merge here. @shyRozen can you please add vSphere in a follow up PR?

shyRozen: @ebenahar Sure thing.

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
```diff
@@ -0,0 +1,361 @@
...
+                        format_type="json",
+                    )
+                    logger.info("Multus disabled on StorageCluster")
+
+                    # Wait for OSD pods to restart without Multus
+                    logger.info(
+                        "Waiting for OSD pods to restart without Multus configuration"
+                    )
+                    wait_for_storage_pods()
+                    verify_storage_cluster()
```
shyRozen, line 386: I think we need a check that the OSD don't have ip from public or cluster.

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
```diff
@@ -0,0 +1,361 @@
...
+                    tool_pod = get_ceph_tools_pod()
+                    tool_pod.exec_ceph_cmd(
+                        ceph_cmd="ceph config rm global public_network", format=None
+                    )
+                    tool_pod.exec_ceph_cmd(
+                        ceph_cmd="ceph config rm global cluster_network", format=None
+                    )
+
+                    # Remove Multus configuration by removing the network spec using JSON patch
+                    patch = '[{"op": "remove", "path": "/spec/network"}]'
```
shyRozen, line 330: Is this process dependent on the ceph commands above? If yes I think this is happening to fast. If it is dependent I would add some check to ceph to see that it is globals are removed. If not it's legit.

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -2219,6 +2226,14 @@ def verify_multus_network():
         )
 
 
+def verify_multus_network():
```
shyRozen: The outcome of the change is all the place that the function was used will skip MDS check that till now was used in all of them. What is the reason for this change?

petr-balogh: Also didn't get why we need extra wrapping function if you have just param with default value why it cannot be added directly to  `verify_multus_network` without renaming it to : `verify_multus_network_base` if no other logic than just calling underlaying function with it's default value? This seems to me like redundant.

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
petr-balogh: @subhamkrai please comment this line out plus remove import of get_polarion_id otherwise it will complain of unused import.
```suggestion
# TODO: Add polarion ID
```

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
petr-balogh: As Elad said, if we keep enabled multus, than some of tests which were decorated with:
skipif_multus_enabled , will not be skipped as this is evaluated during test collection - so once pytest starts.
But we also have some logic in the code depending on this config value:
```yaml
ENV_DATA:
  is_multus_enabled: true
```
So maybe worth also as part of enabling multus to set this config value to True, and then if we disable it, set this value back to `false`.

But as I said, if there is some test to be skipped on multus env , and in the middle of the test we change the config value on which skipif marker depends on, this will not skip test if value got updated in the middle of the run.

There is a way I can imagine to enhance the logic to combine the skipif marker skipif_multus_enabled
with some other marker for this purpose and then have global autouse function scope fixture which will check if the value of is_multus_enabled is set to true and if test is decorated with that specific marker the check about skip will be invoked.

Combination of multiple markers you can see in action in marks module:
```
tier4a = compose(tier4, pytest.mark.tier4a)
```
https://github.com/red-hat-storage/ocs-ci/blob/b53e6d08eece43854df99380f5f6415ebb8244e2/ocs_ci/framework/pytest_customization/marks.py#L60C1-L60C44

So in case we will not be able to disable the test, let's update the skipif_multus_enabled and implement the fixture as well and make sure we are changing the config value `is_multus_enabled` once it's enabled.

I can explain specific implementation of this `compose` and how I mean in fixture, just DM on slack if need help.

petr-balogh: BTW when I wrote this comment , I thought there is an problem with disabling multus. If there is no issue with disabling mutlus , we don't need to implement the `compose` approach which I suggested, but still worth to enable config value once multus enabled and set to false once disabled.

ocs_ci/ocs/resources/storage_cluster.py
subhamkrai: In odf, we have small bug where mds is not moved to multus network, it requires a manual restart and restart should be done once osd are updated to multus network. This has already being discussed with Eran and conclusion was we'll document this as a extra step when moving from non-multus to multus network.

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
subhamkrai: No, it will not have any timing issue or at least in all the testing now we have not seen this as anyways after remove the network spec it takes time as each pod are move at one-by-one.So, we are good here.

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
subhamkrai: @petr-balogh so IIUC, we need to mark the config true once we moved to multus and false when move to non-multus?

ocs_ci/framework/conf/default_config.yaml
subhamkrai: @shyRozen so these changes are lint changes for yaml, so while adding the defualt config it was added by my linter which are good changes to have

.secrets.baseline
subhamkrai: I'm not touching this file in my pr

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
subhamkrai: Added

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
subhamkrai: I have add the env now

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
subhamkrai: added the comment

ocs_ci/framework/conf/default_config.yaml
shyRozen: @petr-balogh Do you agree to these changes?

petr-balogh (approved):

tests/functional/z_cluster/test_enable_multus_existing_cluster.py
```diff
@@ -0,0 +1,405 @@
...
+        wait_for_storage_pods()
+        osd_pods = get_osd_pods()
+        logger.info(f"All {len(osd_pods)} OSD pods are running")
+
+        sc = get_storage_cluster()
+        from ocs_ci.deployment.nmstate import NMStateInstaller
+
+        logger.info("Step 3: Installing NMState operator and creating an instance")
+        nmstate_obj = NMStateInstaller()
+        nmstate_obj.running_nmstate()
```
petr-balogh: You will need to update to something like:
```
nmstate_operator = NMStateOperator(
                    create_catalog=True,
                )
                nmstate_operator.deploy()
```

Of course on top of module to import it:
```
from ocs_ci.utility.operators import NMStateOperator
```

subhamkrai: Updated, should be fine now.

petr-balogh (approved):

shyRozen (approved):

Issue #9250, opened 2024-01-30
The bucket_factory fixture fails to verify AWS buckets creation via boto3 due to the upgrading of ocs-ci packages


sagihirshfeld:
https://github.com/red-hat-storage/ocs-ci/blob/bf2b7dc2ac41eaf8c9466e583c5b0c39bf654220/ocs_ci/ocs/resources/cloud_manager.py#L283-L301

Results in the following error due to the attempt to use head_bucket at line 296:

self.client.meta.client.head_bucket(Bucket=uls_name)
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/path/to/ocs-ci/venv/lib/python3.8/site-packages/botocore/client.py", line 553, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/private/path/to/ocs-ci//venv/lib/python3.8/site-packages/botocore/client.py", line 1009, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (301) when calling the HeadBucket operation: Moved Permanently
```

This seems to be in result of upgrading the virtual envs after the merging of https://github.com/red-hat-storage/ocs-ci/pull/9139 and https://github.com/red-hat-storage/ocs-ci/pull/9125 since at first I only saw this on Jenkins runs, and once I upgraded my local venv I also got the same error.

sagihirshfeld:
As a side effect that we might need to address as its own issue, AWS buckets that were created but failed to be verified in this manner are not cleaned up by OCS-CI because the cloud_uls_factory fixture crushes after their creation but before adding them to the "buckets to delete" list.

sagihirshfeld:
Downgrading boto3 from 1.34.14 to 1.20.2 seems to fix this, but might cause other side effects.

1.20.2 was chosen arbitrarily between the original 1.14.7 and the current 1.34.14 since the original was incompatible with other packages that were updated. 

If we can't find a workaround in our usage of boto3 and we must downgrade, then we should probably look for the latest boto3 version that fixes the issue.

sagihirshfeld:
After some binary-search like exploration, I found that the latest version that fixes this is 1.24.96

petr-balogh:
Are we passing correctly the region? I see they mention something related to region here:
https://stackoverflow.com/questions/45338550/unable-to-access-s3-from-ec2-instance-in-cloudformation-a-client-error-301

[...]

PR #13384, opened 2025-10-15
Update cloud_manager::S3Client's boto3 client to use virtual addressing style for AWS buckets

sagihirshfeld:
Fixes: #9250, #9294

### Background
In the latest boto3 versions releases path-style S3 addressing (https://s3.amazonaws.com/bucket) is no longer supported for most regions, and causes errors when there's a mismatch between the client's region and the region of the bucket we are trying to create or manage (e.g., client in us-east-2 creating a bucket in eu-central-1).

Virtual-hosted–style addressing (`https://bucket.s3.<region>.amazonaws.com`) embeds the bucket name directly in the hostname, allowing AWS to automatically route the request to the correct regional endpoint. 

This requires omitting the legacy global endpoint (`https://s3.amazonaws.com`), since that endpoint always resolves to `us-east-1` and would override region-based routing. We use us-east-1 as the default region because it’s the only one that can reliably create and manage buckets across other regions.

### Changes
Update cloud_manager::S3Client:
1. Create the boto3 client with `Config(s3={"addressing_style": "virtual"})`
2. Removes use of the legacy global endpoint (https://s3.amazonaws.com).
3. Sets us-east-1 as the region_name
4. Allow bucket creation in the us-east-1 region by adding a check to `internal_create_uls`

### DIFF
--- a/ocs_ci/ocs/resources/cloud_manager.py
+++ b/ocs_ci/ocs/resources/cloud_manager.py
@@ -4,6 +4,7 @@ import logging
 from abc import ABC, abstractmethod
 
 import boto3
+from botocore.client import Config
 import google.api_core.exceptions as GoogleExceptions
 from azure.core.exceptions import ResourceNotFoundError, AzureError
 from azure.storage.blob import BlobServiceClient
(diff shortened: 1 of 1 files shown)


sagihirshfeld:
This PR also fixes the creation of AWS Namespacestores that use Caching, but those tests are failing due to an unrelated issue: https://url.corp.redhat.com/411b96b

PR #13409, opened 2025-10-16
Refactored Resiliency tests

paraggit:
This PR refactors the resiliency test framework to improve maintainability and simplify workload management.
- Added a factory pattern (resiliency_workload_factory.py) and centralized configuration system with a new `resiliency_tests_config.yaml ` file to manage both workload parameters and framework parameters from a single config file, eliminating hardcoded values and improving test maintainability
- Simplified test implementations in  by using workload fixtures defined in conftest.py, reducing code duplication by 353 lines while adding comprehensive documentation for workload operations

### DIFF
--- a/ocs_ci/krkn_chaos/krkn_workload_factory.py
+++ b/ocs_ci/krkn_chaos/krkn_workload_factory.py
@@ -127,6 +127,51 @@ class WorkloadOps:
             log.error(f"Failed to start background cluster operations: {e}")
             self.background_cluster_ops = None
 
+    def validate_workload_operations(self):
+        """
+        Validate workload health without performing cleanup.
+
+        This method validates that workloads are still running correctly
+        after chaos testing, without stopping or cleaning them up.
+        Useful for intermediate validation during test execution.
+
+        Raises:
+            UnexpectedBehaviour: If workload validation fails
+            CommandFailed: If command execution fails during validation
+        """
+        log.info(f"Validating {len(self.workloads)} workloads...")
+
+        validation_errors = []
+        for i, workload in enumerate(self.workloads, 1):
+            try:
+                # Determine workload type for this specific workload
+                workload_type = self._get_workload_type_for_workload(workload)
+
+                if workload_type == KrknWorkloadConfig.VDBENCH:
+                    self._validate_vdbench_workload(workload)
+                elif workload_type == KrknWorkloadConfig.CNV_WORKLOAD:
+                    self._validate_cnv_workload(workload)
+                elif workload_type == KrknWorkloadConfig.RGW_WORKLOAD:
+                    self._validate_rgw_workload(workload)
+                else:
+                    log.warning(f"Unknown workload type: {workload_type}")
+
+            except Exception as e:
+                error_msg = f"Issue validating workload {i}: {e}"
+                log.warning(error_msg)
+                validation_errors.append(error_msg)
+
+        if validation_errors:
+            error_summary = "\n".join(validation_errors)
+            log.error(f"Workload validation errors:\n{error_summary}")
+            from ocs_ci.ocs.exceptions import UnexpectedBehaviour
+
+            raise UnexpectedBehaviour(
+                f"Workload validation failed for {len(validation_errors)} workload(s):\n{error_summary}"
+            )
+
+        log.info(f"✓ All {len(self.workloads)} workloads validated successfully")
+
     def validate_and_cleanup(self):
         """
         Validate workload health and perform cleanup.
--- a/ocs_ci/resiliency/resiliency_workload_config.py
+++ b/ocs_ci/resiliency/resiliency_workload_config.py
@@ -322,24 +322,37 @@ class ResiliencyWorkloadConfig:
             if "VDBENCH" in workloads:
                 vdbench_config = self.get_vdbench_config()
                 if not vdbench_config:
-                    log.warning("VDBENCH enabled but no vdbench_config found")
-                    return True
+                    raise ValueError(
+                        "VDBENCH workload is enabled but 'vdbench_config' section is missing. "
+                        "Please provide vdbench_config in your resiliency_tests_config.yaml. "
+                        "See conf/ocsci/resiliency_tests_config.yaml for configuration examples."
+                    )
 
                 # Check block and filesystem configs
                 block_config = vdbench_config.get("block", {})
                 filesystem_config = vdbench_config.get("filesystem", {})
 
                 if not block_config and not filesystem_config:
-                    log.warning("No block or filesystem workload configuration found")
+                    raise ValueError(
+                        "VDBENCH workload is enabled but neither 'block' nor 'filesystem' "
+                        "configuration is provided in vdbench_config. "
+                        "At least one of these must be configured. "
+                        "See conf/ocsci/resiliency_tests_config.yaml for configuration examples."
+                    )
 
-                # Validate patterns
-                for workload_type in ["block", "filesystem"]:
-                    config_section = vdbench_config.get(workload_type, {})
+                # Validate patterns for each configured workload type
+                configs = {
+                    "block": block_config,
+                    "filesystem": filesystem_config,
+                }
+                for workload_type, config_section in configs.items():
                     if config_section:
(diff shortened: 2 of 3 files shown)


ocs_ci/resiliency/resiliency_workload_config.py
```diff
@@ -0,0 +1,331 @@
...
+
+class ResiliencyWorkloadConfig:
+    """
+    Configuration loader for resiliency workload testing.
+
+    This class accesses configuration from the framework's config system
+    and provides methods to access workload-specific settings.
+
+    Configuration should be passed via --ocsci-conf parameter:
+        pytest --ocsci-conf conf/ocsci/resiliency_tests_config.yaml ...
```
shyRozen: run-ci?

ocs_ci/resiliency/resiliency_workload_config.py
```diff
@@ -0,0 +1,331 @@
...
+
+    def get_workloads(self) -> List[str]:
+        """
+        Get list of enabled workload types.
+
+        Returns:
+            list: List of workload type strings
+        """
+        resiliency_config = self.config.ENV_DATA.get("resiliency_config", {})
+        workloads = resiliency_config.get("workloads", ["VDBENCH"])
```
shyRozen: Maybe if no workloads specify we should fail the test and not go with VDBENCH. Could be that the user did some typo mistake like:
worloads:
  - RGW
And no one will know of as it will run with VDBENCH.

ocs_ci/resiliency/resiliency_workload_config.py
```diff
@@ -0,0 +1,331 @@
...
+        Returns:
+            bool: True if configuration is valid
+
+        Raises:
+            ValueError: If configuration is invalid
+        """
+        try:
+            # Check if resiliency_config exists
+            resiliency_config = self.config.ENV_DATA.get("resiliency_config", {})
+            if not resiliency_config:
```
shyRozen, line 287: I think we shouldn't run if no specified configuration. What can be the scenario that the configuration is not specified and we should run the test? 
The doc specify "bool: True if configuration is valid" but in this case not valid configuration was found.

paraggit: If the configuration is not specified, the code uses the default values defined inside it, so the test still runs.  Is this l okay.?

ocs_ci/resiliency/resiliency_workload_config.py
paraggit: I will make a change here like if no workload is defined then it will not run any workload , in case if we want to only inject failures only in that case it will work. and if there is typo in the workload name it wont get match with the existing workload implimantation and it wil raise that workload is not defined.

ocs_ci/resiliency/resiliency_workload_config.py
paraggit: Done

ocs_ci/resiliency/resiliency_workload_config.py
shyRozen: I don't know really. Again what could be the case that configuration is not specified? It like saying to run-ci if you don't have --ocsci-conf run with default. Please try to clarify what is the need? My opinion is that who ever runs this need to familiar with the configuration and not put it in drive and go :)

paraggit: Agree Raising Assert here if configuration is not defined.

ocs_ci/resiliency/resiliency_workload_config.py
```diff
@@ -0,0 +1,355 @@
...
+                    "No workloads configured. Running in failure-injection-only mode. "
+                    "No workload will be executed during the test."
+                )
+
+            # Validate VDBENCH config if enabled
+            if "VDBENCH" in workloads:
+                vdbench_config = self.get_vdbench_config()
+                if not vdbench_config:
+                    log.warning("VDBENCH enabled but no vdbench_config found")
+                    return True
```
shyRozen: Why return True (configuration is valid) when vdbench config is missing?

ocs_ci/resiliency/resiliency_workload_config.py
```diff
@@ -0,0 +1,355 @@
...
+
+                # Check block and filesystem configs
+                block_config = vdbench_config.get("block", {})
+                filesystem_config = vdbench_config.get("filesystem", {})
+
+                if not block_config and not filesystem_config:
```
shyRozen: May OR between them?

ocs_ci/resiliency/resiliency_workload_config.py
```diff
@@ -0,0 +1,355 @@
...
+
+                # Validate patterns
+                for workload_type in ["block", "filesystem"]:
+                    config_section = vdbench_config.get(workload_type, {})
+                    if config_section:
+                        patterns = config_section.get("patterns", [])
+                        if not patterns:
+                            log.warning(
+                                f"No patterns configured for {workload_type} workloads"
+                            )
```
shyRozen: All the check is for the log warning? in the end you are returning true if "VDBENCH" in workloads:
Will it work if none of the workloads is defined?

ocs_ci/resiliency/resiliency_workload_config.py
```diff
@@ -0,0 +1,355 @@
...
+
+                # Check block and filesystem configs
+                block_config = vdbench_config.get("block", {})
+                filesystem_config = vdbench_config.get("filesystem", {})
+
+                if not block_config and not filesystem_config:
+                    log.warning("No block or filesystem workload configuration found")
+
+                # Validate patterns
+                for workload_type in ["block", "filesystem"]:
```
shyRozen: Maybe put the for loop with an else to the condition above because if can run even there was not config for block and filesystem.
Other thing I'm thinking is that you already got the config few lines above:
                block_config = vdbench_config.get("block", {})
                filesystem_config = vdbench_config.get("filesystem", {})
What is the reason to get it again and not use the variable that we already have?

ocs_ci/resiliency/resiliency_workload_config.py
paraggit: Yes right , Raised  ValueError if vdbench_config is not defined.

ocs_ci/resiliency/resiliency_workload_config.py
paraggit: Raised ValueError when none of the config is defined.

ocs_ci/resiliency/resiliency_workload_config.py
paraggit: Yes Added suggested code changes

conf/ocsci/resiliency_tests_config.yaml
```diff
@@ -0,0 +1,133 @@
+# Resiliency testing configuration
+
+ENV_DATA:
+  resiliency_config:
+    # Workload types: VDBENCH, CNV_WORKLOAD, FIO, RGW_WORKLOAD
+    workloads:
+      - VDBENCH
+      - RGW_WORKLOAD
+
```
avd-sagare, line 9: cnv_workload?

paraggit: CNV_WORKLOAD is not enabled by default as it requires additional configuration during deployment. A separate PR will be submitted to add the required CNV-specific configuration.

ocs_ci/krkn_chaos/krkn_workload_factory.py
```diff
@@ -127,6 +127,51 @@ def _start_background_cluster_operations(self):
...
+        """
+        log.info(f"Validating {len(self.workloads)} workloads...")
+
+        validation_errors = []
+        for i, workload in enumerate(self.workloads, 1):
+            try:
+                # Determine workload type for this specific workload
+                workload_type = self._get_workload_type_for_workload(workload)
+
+                if workload_type == KrknWorkloadConfig.VDBENCH:
```
ebenahar, line 150: @paraggit have you considered implementing this as an Interface-like base class and child classes with implementation per workload? It's probably just another design pattern but maybe worth considering

PR #13414, opened 2025-10-17
add ClusterIP as allowed providerAPIServerServiceType

dahorak:
Allow `ClusterIP` as `providerAPIServerServiceType` (additionally to `NodePort`).


vavuthu (approved):

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -940,9 +940,9 @@ def ocs_install_verification(
     sc_obj = get_storage_cluster()
     sc_ob_spec = sc_obj.get().get("items")[0].get("spec")
     if sc_ob_spec.get("hostNetwork") is True:
-        assert (
-            sc_ob_spec.get("providerAPIServerServiceType")
-            == constants.SERVICE_TYPE_NODEPORT
+        assert sc_ob_spec.get("providerAPIServerServiceType") in (
```
DanielOsypenko, line 960: In case of `providerAPIServerServiceType == constants.SERVICE_TYPE_CLUSTERIP`. Provider mode deployment will not work, because external clients cannot reach the provider API. 
For my understanding, do we have such deployments or requirements?

dahorak:
Verification job: https://url.corp.redhat.com/b3730fd

ocs_ci/ocs/resources/storage_cluster.py
dahorak: It seems to be the configuration on HCI racks.

Issue #13429, opened 2025-10-20
[Bug]: odf noobaa calls fail in post 4.19->4.20 runs because older cli binary is kept after upgrade


sagihirshfeld:
### Description
Post-upgrade OCS-CI runs have been failing on MCG tests that call `odf noobaa`.

 In some cases it's clear there's something wrong with the binary:
```
cs_ci.ocs.exceptions.CommandFailed: Error during execution of command: odf noobaa api object_api list_objects '{"bucket":
"oc-bucket-81170022b9104887984948d0795f10"}' -ojson -n openshift-storage.
Error is Error: unknown shorthand flag: 'o' in -ojson
```

or
```
ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: ['yes | odf noobaa namespacestore create nsfs nsfs-ns-store-9f9a0daa3c5840ec8291fc3a3b --pvc-name nsfs-persistentvolumeclaim-0669f50abca44  -n openshift-storage'].
Error is Error: unknown flag: --pvc-name
```

But in others it's more subtle:
```
ocs_ci.ocs.exceptions.TimeoutExpiredError: Timed out after 600s running _check_state()
```
^ in such cases the TimeoutSampler hides the actual error, which looks like one of the other two above in the OCS-CI logs. 

In the OCS-CI logs of these runs we can also see that `noobaa` is not a sub-option for the `odf` binary, since it was only added in 4.20. We can also see that post-upgrade we're not downloading the new binary we only check the existence on the local agent in the expected path: [odf_cli.py lines 28-59](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/helpers/odf_cli.py/#L28-L59)

### RP links
- https://url.corp.redhat.com/a0a0fdb
- https://url.corp.redhat.com/9da3a44
- https://url.corp.redhat.com/7cdecac

PR #13430, opened 2025-10-20
Re-download the odf cli when needed to match the cluster's version

sagihirshfeld:
Fixes: #13429

### DIFF
--- a/ocs_ci/helpers/odf_cli.py
+++ b/ocs_ci/helpers/odf_cli.py
@@ -1,5 +1,6 @@
 import os
 import platform
+import xattr
 from stat import S_IEXEC
 from logging import getLogger
 from typing import Union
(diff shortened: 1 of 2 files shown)


sagihirshfeld:
Here's an upgrade run where we can see tests failing with the original issue: https://url.corp.redhat.com/887535e

And in contrast, we can see in a run which uses the branch of this PR that the original solution of always-downloading the CLI in upgrade runs fixes the issue so the same tests pass after an upgrade: https://url.corp.redhat.com/2d49a78

sagihirshfeld:
While the original approach worked, it had the downside of always downloading the CLI during upgrade runs.

Unfortunately, the ODF CLI doesn’t have a “version” option, so in [this commit](https://github.com/red-hat-storage/ocs-ci/pull/13430/commits/b0d29a065cabf45ff3ed8affff26e18b410c9ae1)
 I’m using extended file attributes to tag the binary file with the associated cluster version. This allows for a quick and simple version check, letting us re-download the binary only once after an upgrade.

sagihirshfeld:
Verified in an ugprade job from 4.19 to 4.20: https://url.corp.redhat.com/46e3c02

PR #13431, opened 2025-10-20
upgrade mce with provider upgrade

DanielOsypenko:
upgrade MCE
MCEInstall was extended with upgrade functionality. The flow mostly taken from the `ACMUpgrade`, but added modern way to run upgrade due to upgrade order and due to mce_version from the `ocp-<ver>-config.yaml`
`test_mce_upgrade` added and `UPGRADE_TEST_MARK_EXPRESSION` keyword(test marker) added - `mce_upgrade`

Tested locally.


jilju (approved):

ocs_ci/deployment/mce.py
```diff
@@ -356,3 +368,352 @@ def wait_mce_resources(self):
                 **depl_ocp_obj.get(retry=60, wait=10, dont_raise=True)
             )
             deployment_obj.wait_for_available_replicas(timeout=600)
+
+    def upgrade_mce(self):
```
fbalak: What is the difference between this method and ACMUpgrade class (https://github.com/red-hat-storage/ocs-ci/blob/33b598393dc677bc7a68b0a458742551222f1822/ocs_ci/ocs/acm_upgrade.py#L26)?

tests/functional/upgrade/test_upgrade.py
```diff
@@ -184,3 +186,15 @@ def test_acm_upgrade(zone_rank, role_rank, config_index):
     """
     acm_hub_upgrade_obj = ACMUpgrade()
     acm_hub_upgrade_obj.run_upgrade()
+
+
+@purple_squad
+@mce_upgrade
```
shylesh, line 192: You need to add this marker in the list https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/multicluster.py#L36 , otherwise this test will not be picked up by the pytest hook.

DanielOsypenko: Fixed. Thank you.

fbalak (approved):

DanielOsypenko:
@fbalak I verified it locally, partially. If you’re Ok with this, lets merge it and since it is a new functionality we can run and in case of issues, I will fix it anyways

PR #13438, opened 2025-10-21
Allow to use system available oc client when failed to download

petr-balogh:


DanielOsypenko (approved):

ocs_ci/utility/utils.py
```diff
@@ -916,7 +915,6 @@ def get_openshift_installer(
                     log.debug(f"version {previous_version} is not GA'ed")
 
         log.info(f"Downloading openshift installer ({version}).")
-        prepare_bin_dir()
         # record current working directory and switch to BIN_DIR
         previous_dir = os.getcwd()
         os.chdir(bin_dir)
```
clacroix12, line 921: Should this be moved after we actually call `prepare_bin_dir`? Otherwise we might try to change to it before it's created.

petr-balogh: Returning it back

DanielOsypenko (approved):

PR #13470, opened 2025-10-29
Ocp upgrade hosted cluster

DanielOsypenko:
* Upgrade for kubevirt client clusters
* post upgrade verification 
a) control plane resources
b) nodepool
c) storage consumer heartbeat timestamp is recent
* post deploy verification
a) storage consumer heartbeat

------ Tests ------
local upgrade passed
lib tests: **10 of 10 completed, 10 Pass, 0 Fail,**


DanielOsypenko:
verified: https://url.corp.redhat.com/b438152

ocs_ci/deployment/helpers/lso_helpers.py
```diff
@@ -307,7 +307,11 @@ def create_optional_operators_catalogsource_non_ga(force=False):
             "Creating optional operators CatalogSource and ImageContentSourcePolicy"
         )
         run_cmd(f"oc apply -f {optional_operators_yaml.name}")
-    wait_for_machineconfigpool_status("all")
+    if config.ENV_DATA.get("platform").lower() == constants.HCI_BAREMETAL:
```
fbalak, line 319: this behaviour could be added to a docstring

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -372,6 +377,39 @@ def delete_hcp_podman_container():
...
+                )
+            else:
+                logger.warning(
+                    "No supported versions returned from hub cluster. Falling back to config hcp_version."
+                )
+    except Exception as e:
+        logger.debug(
+            f"Could not determine Hypershift installation or supported versions due to: {e}"
+        )
+    return hcp_version
```
clacroix12, line 410: Is this safe to return `None` in the event we hit an exception? It looks like we continue on to pass this to the `git clone` command which I assume will cause issues.

ocs_ci/ocs/resources/storageconsumer.py
```diff
@@ -1206,3 +1208,92 @@ def wait_storage_classes_equal(given_classes):
...
+            config.ENV_DATA["cluster_namespace"],
+            config.get_provider_index(),
+        )
+        client_info = sc.get_client_status()
+        consumer_url = client_info.get("clusterName", "")
+
+        return consumer_url.split(".")[0] if consumer_url else ""
+
+
+def verify_last_heartbeat_timestamp(cluster_name):
```
clacroix12, line 1238: Add args and returns to docstring

ocs_ci/deployment/helpers/hypershift_base.py
DanielOsypenko: Originally helper methods are not designed to be called only once in a specific flow. This method can return None and we must evaluate it higher in the flow.  I added processing None and fallback to configured version higher. Thanks.
```
            if not hcp_version:
                logger.error("Falling back to configured hcp_version.")
                return self.install_hcp_and_hypershift_from_git(install_latest=False)
```

jilju (dismissed):

suchita-g (dismissed):

fbalak (dismissed):

ocs_ci/ocs/resources/storageconsumer.py
clacroix12: This is still missing the returns section but that can be updated in a follow-up so we don't hold this PR up

clacroix12 (dismissed):

fbalak (approved):

fbalak (approved):

suchita-g (approved):

DanielOsypenko:
although job overall is not successful, the kubevirt upgrade test has passed
https://url.corp.redhat.com/42e9f96

jilju (approved):

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -509,19 +564,23 @@ def _download_hcp_binary_with_podman(self):
...
+            If False, use the configured hcp_version.
         """
 
         if not config.ENV_DATA.get("hcp_version"):
             logger.error("hcp_version is not set in config.ENV_DATA")
             return
 
-        self.delete_hcp_and_hypershift()
-        self.install_hcp_and_hypershift_from_git()
+        self.delete_hcp_and_hypershift_bin()
```
dahorak, line 580: Wouldn't be possible to firstly check the installed version and compare it with the desired? (this will avoid unnecessary re-installation of the same version, but not sure, if that situation could happen)

DanielOsypenko: this is "update_hcp_binary" function, this utility never gave us issues it suppose to delete and install latest, even if this only a latest build update.
re-installing just gives us peace of mind, at least about this component in all the scenarios including upgrade, when version is bumped

suchita-g (approved):

PR #13485, opened 2025-10-31
Setup Storage for Fusion 2.11

rh-aaleem:
### Summary

This PR adds the setup and validation flow for Storage configuration for Fusion 2.11, ensuring that Data Foundation and StorageCluster components are deployed, initialized, and verified for health before proceeding with dependent services.

### Changes Introduced
Added logic for creating local volume set
Added logic to add label on worker node 
Added setup_storage() logic to handle storage setup for Fusion 2.11.
Added checks for validating the storage cluster

### DIFF
--- a/conf/deployment/baremetal/ai_1az_rhcos_nvme_intel_3m_3w_provider.yaml
+++ b/conf/deployment/baremetal/ai_1az_rhcos_nvme_intel_3m_3w_provider.yaml
@@ -1,6 +1,7 @@
 ---
 DEPLOYMENT:
   allow_lower_instance_requirements: false
+  customized_deployment_storage_class: 'localblock'
   local_storage: true
   host_network: true
 ENV_DATA:
--- a/ocs_ci/deployment/fusion_data_foundation.py
+++ b/ocs_ci/deployment/fusion_data_foundation.py
@@ -37,7 +37,9 @@ class FusionDataFoundationDeployment:
         self.pre_release = config.DEPLOYMENT.get("fdf_pre_release", False)
         self.kubeconfig = config.RUN["kubeconfig"]
         self.lso_enabled = config.DEPLOYMENT.get("local_storage", False)
-        self.storage_class = storage_class.get_storageclass()
+        self.storage_class = (
+            storage_class.get_storageclass() or constants.DEFAULT_STORAGECLASS_LSO
+        )
         self.custom_storage_class_path = None
 
     def deploy(self):
--- a/ocs_ci/deployment/helpers/storage_class.py
+++ b/ocs_ci/deployment/helpers/storage_class.py
@@ -44,7 +44,7 @@ def get_storageclass() -> str:
         storage_class = DEFAULT_STORAGE_CLASS_MAP.get(platform)
 
     logger.info(f"Using storage class: {storage_class}")
-    return storage_class or constants.DEFAULT_STORAGECLASS_LSO
+    return storage_class
 
 
 def create_custom_storageclass(storage_class_path: str) -> str:


rh-aaleem:
Dependent MR 
https://gitlab.cee.redhat.com/ocs/ocs4-jenkins/-/merge_requests/2300

ocs_ci/framework/conf/fdf_version/fdf-4.18.yaml
```diff
@@ -4,3 +4,4 @@ DEPLOYMENT:
   fdf_image_tag: v4.18
   fdf_pre_release_registry: docker-na-public.artifactory.swg-devops.com/hyc-abell-devops-team-dev-docker-local/df
   fdf_pre_release_image_digest: null
+  local_storage: true
```
clacroix12: I don't think we want to default to using local_storage in these configurations. These version configs shouldn't assume anything about the underlying cluster and only contain FDF version specifics.

ocs_ci/ocs/constants.py
```diff
@@ -3572,6 +3572,8 @@
 DAEMONSET_CSI_RBD_CSI_ADDONS = (
     "openshift-storage.rbd.csi.ceph.com-nodeplugin-csi-addons"
 )
+FDF_LSO_STORAGECLASS = "ibm-spectrum-fusion-local"
+FDF_LSO_DEVICE_SET_NAME = "ocs-deviceset-ibm-spectrum-fusion-local"
```
clacroix12: These aren't being used

ocs_ci/utility/storage_cluster_setup.py
```diff
@@ -0,0 +1,650 @@
...
+logger = logging.getLogger(__name__)
+
+
+# create custom storage class for StorageCluster CR if necessary
+class StorageClusterSetup(object):
+    def __init__(self):
+        self.custom_storage_class_path = None
+        self.namespace = constants.OPENSHIFT_STORAGE_NAMESPACE
+        self.platform = config.ENV_DATA["platform"]
+        self.storage_class = constants.LOCALSTORAGE_SC
```
clacroix12: Should we default to retrieving the default storageclass the same way we do for ODF deployments? Then we can override this value in the event we are deploying with local storage, or update `get_storageclass()` to return the proper value if we are configured for local storage.

```suggestion
        self.storage_class = storage_class.get_storageclass()
```

ocs_ci/utility/storage_cluster_setup.py
```diff
@@ -0,0 +1,650 @@
...
+
+    def setup_storage_cluster(self):
```
clacroix12, line 49: For now I think this function is okay as it's mostly directly taken from how we are handling things in the ODF deployment module. I would however like to see us revisit this and try to use something like a builder design to organize the many conditionals that go into constructing our storage cluster data prior to creation. I think it would go a long way in making it easier to adapt this function/class to both ODF and FDF deployments.

rh-aaleem: Sure, that we already discussed, may be will try in next iteration,

ocs_ci/framework/conf/fdf_version/fdf-4.18.yaml
rh-aaleem: Yeah right, i will remove it from here as this is already mentioned in the platform yaml.

ocs_ci/utility/storage_cluster_setup.py
```diff
@@ -0,0 +1,602 @@
+from ocs_ci.deployment.helpers import storage_class
```
ebenahar: I think the file name should be storage_cluster_setup_for_fdf.py

clacroix12: > I think the file name should be storage_cluster_setup_for_fdf.py

I believe the long term goal is to have ODF utilize the code in this module as well.

ocs_ci/utility/storage_cluster_setup.py
```diff
@@ -0,0 +1,686 @@
...
+                    resources["noobaa-core"] = {
+                        "limits": {"cpu": 2, "memory": "8Gi"},
+                        "requests": {"cpu": 1, "memory": "8Gi"},
+                    }
+                    resources["noobaa-db"] = {
+                        "limits": {"cpu": 2, "memory": "8Gi"},
+                        "requests": {"cpu": 1, "memory": "8Gi"},
+                    }
+                cluster_data["spec"]["resources"] = resources
+
```
petr-balogh:
```
        # Disable NFS when cluster resources are insufficient
        from ocs_ci.utility.nfs_utils import check_cluster_resources_for_nfs

        if not check_cluster_resources_for_nfs():
            logger.info(
                "Disabling NFS because cluster resources are insufficient "
                "(based on actual node CPU/memory measurement)"
            )
            cluster_data["spec"]["nfs"] = {"enable": False}

```
We are missing this recent change:
https://github.com/red-hat-storage/ocs-ci/commit/fa047c4430f600ad5f09d43e74fe52eac794d762
Check all changes from PR as there are more lines than above

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -30,6 +35,9 @@ class FusionDataFoundationDeployment:
     def __init__(self):
         self.pre_release = config.DEPLOYMENT.get("fdf_pre_release", False)
         self.kubeconfig = config.RUN["kubeconfig"]
+        self.lso_enabled = config.DEPLOYMENT.get("local_storage", False)
+        self.storage_class = storage_class.get_storageclass() or "localblock"
```
DanielOsypenko: not sure if "localblock" can be a default value here, I would rather remove it to reduce questions in debugging scenarios

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -166,9 +175,12 @@ def setup_storage(self):
             self.create_odfcluster()
             odfcluster_status_check()
         else:
-            logger.warning(
-                "Storage configuration for Fusion 2.11 or greater not yet implemented"
-            )
+            logger.info("Storage configuration for Fusion 2.11 or greater")
+            clustersetup = StorageClusterSetup(FusionDataFoundationDeployment=self)
+            create_lvs_resource(self.storage_class, self.storage_class)
```
DanielOsypenko, line 183: I think we don't need create lvs resource. It should probably created with `clustersetup.setup_storage_cluster()`, please check this

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -166,9 +175,12 @@ def setup_storage(self):
...
+            add_storage_label()
+            clustersetup.setup_storage_cluster()
+            storagecluster_health_check()
```
DanielOsypenko, line 186: I think we are doing this test already with setup_storage_cluster and post-deployment checks, it is redundant

conf/deployment/baremetal/ai_1az_rhcos_nvme_intel_3m_3w_provider.yaml
```diff
@@ -0,0 +1,12 @@
+---
```
DanielOsypenko, line 1: do we need this new configuration? it seems, it does not have unique settings if to compare to other provider configurations. Also there are more fields in env_data, like, for example cluster_type (there are more actually), that required to create hub cluster

ocs_ci/deployment/fusion_data_foundation.py
rh-aaleem: its needed, it was not creating local volume set, i tried this without create_lvs_resources,

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -292,3 +337,10 @@ def __init__(self, resource_name="", *args, **kwargs):
         super(OdfCluster, self).__init__(
             resource_name=resource_name, kind="OdfCluster", *args, **kwargs
         )
+
+
+class StorageCluster(OCP):
```
clacroix12: Is there a reason why we don't use the `StorageCluster` class from `ocs_ci/ocs/resources/storage_cluster.py`?

ocs_ci/framework/conf/fdf_version/fdf-4.19.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 DEPLOYMENT:
   fdf_pre_release: true
-  fdf_image_tag: v4.19
+  fdf_image_tag: 4.19
```
clacroix12: The `v` here is intentional and necessary to pull the image from artifactory.

ocs_ci/deployment/fusion_data_foundation.py
clacroix12: Last step of `setup_storage_cluster` is the `oc create` so I think the health check here is okay. This ensures we fail the FDF deployment stage in our jenkins pipeline prior to test execution if the storagecluster is not healthy.

ocs_ci/deployment/fusion_data_foundation.py
clacroix12: Ideally I think we should have `storage_class.get_storageclass()` return `localblock` when the correct conditions are met. This way we don't need to handle those scenarios anywhere else and the logic is consistent across deployment types.

ocs_ci/deployment/fusion_data_foundation.py
rh-aaleem: @clacroix12 you are right its just creating storage cluster and not verifying anything,

ocs_ci/deployment/fusion_data_foundation.py
petr-balogh: We don't even set this to localblock for LSO deployments. We are returning None in get_storageclass. So do we really need to set here the `localblock` ? But agree with Coady - it's should be handled and returned properly in:
`get_storageclass` if possible

conf/deployment/baremetal/ai_1az_rhcos_nvme_intel_3m_3w_provider.yaml
rh-aaleem: Yes we need to have this configuration for the simplicity, if we will not have this config we need to pass parameter through jenkins `yaml to text` which is little cumbersome,

ocs_ci/deployment/fusion_data_foundation.py
rh-aaleem: Just Providing another abstraction to make more readable , like we are doing for `odfCluster` and `FusionServiceInstance`.

ocs_ci/deployment/fusion_data_foundation.py
DanielOsypenko: in the post-deployment verification we do have `health_check` and `post_onboarding_verification` and the right place to do checks happened to be in `test_deployment`.
If we fail in cluster fixture that we will not see a failure with assigned polarion ID of the deployment, I guess that was the reason why all the checks are there.

DanielOsypenko: @petr-balogh WDYT?

ocs_ci/deployment/fusion_data_foundation.py
petr-balogh: @rh-aaleem as we discussed - please import and use the one class mentioned by Coady

rh-aaleem: done

ocs_ci/deployment/fusion_data_foundation.py
petr-balogh: @DanielOsypenko , the deployment of FDF is done a bit differently - not in test case - means we don't run pytest for this entrypoint.  And this will not affect anyhow the normal deployment.

In the FDF deployment - we need to actually fail the deployment in case it's not healthy at the end which is the purpose of this check, otherwise this will end up with 0 RC and will continue with test execution. 

So I am fine to keep the ceph health check here.

petr-balogh (dismissed):

ocs_ci/deployment/helpers/storage_class.py
```diff
@@ -44,7 +44,7 @@ def get_storageclass() -> str:
         storage_class = DEFAULT_STORAGE_CLASS_MAP.get(platform)
 
     logger.info(f"Using storage class: {storage_class}")
-    return storage_class
+    return storage_class or constants.DEFAULT_STORAGECLASS_LSO
```
clacroix12: Like we discussed, I think it's okay to revert this and go back to overriding the value in the data foundation module.

I think we need to take a look at the storageclass code (and deployment class heirarchy in general) outside the scope of this PR.

rh-aaleem:
This is the verification job https://url.corp.redhat.com/c9e161f

petr-balogh (approved):

PR #13494, opened 2025-11-02
prune and apply idms on upgrade

DanielOsypenko:
1. delete old df-repo idms to avoid problem with conflicting items when we have multiple sources like:
```
oc get imagedigestmirrorsets.config.openshift.io 
NAME AGE 
acm-idms 13d 
df-repo-4.18.12-1760620940436 16d 
df-repo-4.19.7-1761060439245 11d
```
2. refactor and use `apply_idms_to_hosted_clusters` to prevent ImagePullBackOff post ODF upgrade

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -5988,6 +5988,13 @@ def get_acm_mce_build_tag(quay_path, version):
 def parse_k8s_timestamp(ts: str) -> datetime:
     """
     Parse RFC3339 timestamp like '2024-10-30T12:34:56Z' to a datetime.
+
+    Args:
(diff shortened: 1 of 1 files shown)


ocs_ci/utility/utils.py
```diff
@@ -5983,3 +5983,14 @@ def get_acm_mce_build_tag(quay_path, version):
     ]
 
     return downstream_tag[0]
+
+
+def parse_k8s_timestamp(ts: str) -> datetime:
+    """
+    Parse RFC3339 timestamp like '2024-10-30T12:34:56Z' to a datetime.
```
fbalak, line 5990: please add docstring to the parameter and return value

PR #13505, opened 2025-11-03
Skipping NFS deployment and related tests for the lowreq deployment configuration.

paraggit:
As per the documentation, NFS deployment is not recommended for lean deployment configurations. Hence, this PR disables NFS and NFS-related tests when the lowreq configuration is used.

### DIFF
--- a/ocs_ci/framework/pytest_customization/marks.py
+++ b/ocs_ci/framework/pytest_customization/marks.py
@@ -838,4 +838,3 @@ skipif_lean_deployment = pytest.mark.skipif(
     or not check_cluster_resources_for_nfs(),
     reason="Test cannot run on lean profile or requires higher cluster resources (insufficient CPU/memory)",
 )
-


ocs_ci/deployment/deployment.py
```diff
@@ -1754,6 +1754,13 @@ def deploy_ocs_via_operator(self, image=None):
                     }
                 cluster_data["spec"]["resources"] = resources
 
+        # Disable NFS when allow_lower_instance_requirements is false
+        if not config.DEPLOYMENT.get("allow_lower_instance_requirements", False):
```
petr-balogh: allow_lower_instance_requirements this is deprecated config value and no production job should use this values.

So we should not rely on this. All production jobs should have this value set to `false` so it will not work as you would think of it.

That's why I've suggested to check the CPU and memory on nodes dynamically and decide if skip or not.

paraggit:
Verified the change in this run , and we can see the nfs tests are being skip as per the condition added
https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster-prod/18085

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -831,3 +832,8 @@ def get_current_test_marks():
 )
 
 ui = compose(skipif_ibm_cloud_managed, pytest.mark.ui)
+
+skipif_lean_deployment = pytest.mark.skipif(
+    not check_cluster_resources_for_nfs(),
+    reason="Test requires higher cluster resources (insufficient CPU/memory for NFS)",
+)
```
petr-balogh:
```suggestion
    config.ENV_DATA.get("performance_profile") == "lean" or not check_cluster_resources_for_nfs(),
    reason="Test cannot run on lean profile or requires higher cluster resources (insufficient CPU/memory)",
)
```

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -831,3 +832,8 @@ def get_current_test_marks():
 )
 
 ui = compose(skipif_ibm_cloud_managed, pytest.mark.ui)
+
+skipif_lean_deployment = pytest.mark.skipif(
+    config.ENV_DATA.get("performance_profile") == "lean" or not check_cluster_resources_for_nfs(),
+    reason="Test cannot run on lean profile or requires higher cluster resources (insufficient CPU/memory)",
+)
```
petr-balogh, line 840:
```suggestion
skipif_lean_deployment = pytest.mark.skipif(
    config.ENV_DATA.get("performance_profile") == "lean"
    or not check_cluster_resources_for_nfs(),
    reason="Test cannot run on lean profile or requires higher cluster resources (insufficient CPU/memory)",
)

```

PR #13523, opened 2025-11-04
Adding writing io to bucket before and after blocking external access

ypersky1980:
In this PR I added creation and writing a file to s3 bucket, to make sure that after  the routes are deleted and reconciled it's still possible to access and write to the bucket.

### DIFF
--- a/tests/functional/object/mcg/test_block_external_access.py
+++ b/tests/functional/object/mcg/test_block_external_access.py
@@ -5,7 +5,15 @@ import pytest
 
 from time import sleep
 
-from ocs_ci.framework.testlib import MCGTest, red_squad, mcg, tier2
+from ocs_ci.framework.testlib import (
+    MCGTest,
+    red_squad,
+    mcg,
+    tier2,
+    skipif_disconnected_cluster,
+    skipif_proxy_cluster,
+    post_upgrade,
+)
 from ocs_ci.ocs import constants, ocp
 from ocs_ci.framework import config
 from ocs_ci.ocs.bucket_utils import craft_s3_command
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -156,12 +158,43 @@ def test_block_access_from_storagecluster(
...
         self.check_disable_routes(False)
 
+        bucket_name = bucket_factory(
+            1,
+            interface="CLI",
+            bucketclass={
+                "interface": "CLI",
+                "backingstore_dict": {"aws": [(1, "eu-central-1")]},
+            },
+        )[0].name
```
sagihirshfeld, line 179: Would creating a bucket on the default noobaa backingstore work here?
```suggestion
        bucket = bucket_factory()[0].name
```

If not, then using an AWS backingstore bucket will require you add these marks so it doesn't fail on disconnected and proxy runs where the cluster can't access AWS:
```
@skipif_disconnected_cluster
@skipif_proxy_cluster
```

tests/functional/object/mcg/test_block_external_access.py
```diff
@@ -156,12 +158,43 @@ def test_block_access_from_storagecluster(
...
         self.check_disable_routes(False)
 
+        logger.info(f"Going to copy file {file_name} to the bucket {bucket_name}")
+        awscli_pod_session.exec_cmd_on_pod(
+            command=craft_s3_command(
+                f"cp {AWSCLI_TEST_OBJ_DIR}{file_name} s3://{bucket_name}/{file_name}",
+                mcg_obj=mcg_obj,
+            ),
+            out_yaml_format=False,
+        )
```
shivamdurgbuns: could we add assert statement to check if the expected results are acheived.
i.e. 
- if disableRoutes flag in storagecluster yaml is set to False then all routes are recreated after deletion
- if disableRoutes flag in storagecluster yaml is set to True then noobaa-mgmt, s3 and sts routes are not
        recreated after deletion

tests/functional/object/mcg/test_block_external_access.py
ypersky1980: @sagihirshfeld thank you. I prefer that it would be an AWS backingstore bucket, since we'd like to see that external access is working. Added the "skipifs".

keesturam (approved):

PR #13526, opened 2025-11-05
Fixed customstorageclass creation issue

paraggit:
Updated the `patch_storage_cluster_for_custom_storage_class` function to correctly handle JSON Patch operations.
 it now checks the current state of the `StorageCluster` and uses `add` for new fields or `replace` for existing ones.

The validation timeout was increased from 4 seconds to 120 seconds using TimeoutSampler, since creating a storage class usually takes 47–56 seconds on healthy clusters.

### DIFF
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -2824,6 +2824,39 @@ def get_storageclass_names_from_storagecluster_spec():
     return data
 
 
+def check_underlying_resource_exists(storage_class_type):
+    """
+    Check if the underlying Ceph resource exists for a given storage class type.
+
+    Args:
+        storage_class_type (str): The type of storage class (e.g., "cephFilesystems",
+                                 "cephBlockPools", "cephObjectStores", "nfs", "encryption")
+
+    Returns:
+        bool: True if the underlying resource exists or if it's a type that doesn't
+             need verification (nfs/encryption), False otherwise.
+    """
+    try:
+        namespace = config.ENV_DATA["cluster_namespace"]
+        resource_map = {
+            "cephFilesystems": "CephFilesystem",
+            "cephBlockPools": "CephBlockPool",
+            "cephObjectStores": "CephObjectStore",
+        }
+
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -2844,16 +2844,56 @@ def check_custom_storageclass_presence(interface=None):
...
+            }
+
+            if storage_class_type in resource_map:
+                kind = resource_map[storage_class_type]
+                resource_obj = OCP(kind=kind, namespace=namespace)
+                resources = resource_obj.get().get("items", [])
+                return len(resources) > 0
+            return True  # For nfs/encryption, assume resource exists
+        except Exception:
+            return True  # If we can't check, assume it exists
```
ebenahar: in what situation would that happen? Shouldn't we at least throw an error log?

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -2877,41 +2917,69 @@ def patch_storage_cluster_for_custom_storage_class(
...
+            if not component:
+                op, op_path, value = (
+                    "add",
+                    base_path,
+                    {"storageClassName": storage_class_name},
+                )
+            elif field_exists:
+                op, op_path, value = "replace", path, storage_class_name
+            else:
+                op, op_path, value = "add", path, storage_class_name
```
ebenahar, line 2989: this is a dup of lines 2944-2962. Please create one block and call it twice

paraggit: yes, created a common function `check_underlying_resource_exists()`  to remove duplicate code.

ocs_ci/ocs/resources/storage_cluster.py
paraggit: Added a error message 
```
            except Exception as e:
            log.error(
                f"Failed to check if underlying resource exists for storage class type '{storage_class_type}': {e}"
            )
            return True 
```

keemano (approved):

keesturam (approved):

Issue #12882, opened 2025-08-11
[Bug]: test_unnecessary_pods fails on external mode setup


prsurve:
### Description
rook-ceph-operator is also present when we have externl mode setup but testcases expect there should not be any pod named rook-ceph-operator 
### Steps to Reproduce
1.
2.
3.

### Actual Behavior
if invalid_pods_found:
>               raise InvalidPodPresent(
                    f"Pods {invalid_pods_found} should not be present because Ceph is not available"
                )
E               ocs_ci.ocs.exceptions.InvalidPodPresent: Pods ['rook-ceph-operator-98689f9cb-j49t4'] should not be present because Ceph is not available

ocs_ci/ocs/resources/storage_cluster.py:3380: InvalidPodPresent
### Environment
- Test Suite(s):  https://url.corp.redhat.com/db0e365
- Platform(s):
- Version(s): 4.19
- OS:

ebenahar:
@prsurve any update?

ebenahar:
@prsurve please update

ebenahar:
@prsurve to submit a PR shortly

PR #13531, opened 2025-11-05
Fix test failure for external mode cluster when wrong pods are checked

prsurve:
fixes: #12882

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -617,7 +617,7 @@
         "hashed_secret": "b7e9a6e2e04ed3edbf8b4e21def4beccbb6eb6b1",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 225,
+        "line_number": 229,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -3380,3 +3379,17 @@ def check_unnecessary_pods_present():
             raise InvalidPodPresent(
                 f"Pods {invalid_pods_found} should not be present because Ceph is not available"
             )
+    if no_ceph:
+        for invalid_pod_name in constants.CEPH_PODS_NAMES:
+            if invalid_pod_name != constants.ROOK_CEPH_OPERATOR:
```
yitzhak12, line 3468: We need to change this line to `if invalid_pod_name.startswith(constants.ROOK_CEPH_OPERATOR):`. Or maybe we can exclude the `constants.ROOK_CEPH_OPERATOR` from the ceph list before we start the pod names iteration.

prsurve: if we use if invalid_pod_name.startswith(constants.ROOK_CEPH_OPERATOR): then it fails with ocs_ci.ocs.exceptions.InvalidPodPresent: Pods ['rook-ceph-operator-5b7f454df-qrkwc'] should not be present because Ceph is not available which is wrong

yitzhak12: ack

PR #13539, opened 2025-11-06
Fixes for system tests

posharma1:
Changes Added :

1. if kms is enable on the system "noobaa-root-master-key-backend" secret will not be present on the system.
2. while deleting CNPG cluster need to pass namespace since previous test step related to kafka is changing namespace need to change back to openshift-storage namespace
3. updating noobaa pvc name with latest ones in 4.20. and changing fixture to  noobaa_db_backup_and_recovery_locally for noobaa db recovery

### DIFF
--- a/tests/cross_functional/conftest.py
+++ b/tests/cross_functional/conftest.py
@@ -153,23 +153,15 @@ def noobaa_db_backup_and_recovery_locally(
         ocp_secret_obj = OCP(
             kind="secret", namespace=config.ENV_DATA["cluster_namespace"]
         )
-        if is_kms_enabled():
-            secrets = [
-                "noobaa-root-master-key-volume",
-                "noobaa-admin",
-                "noobaa-operator",
-                "noobaa-server",
-                "noobaa-endpoints",
-            ]
-        else:
-            secrets = [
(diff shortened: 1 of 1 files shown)


tests/cross_functional/conftest.py
```diff
@@ -153,14 +153,23 @@ def factory(
...
+            ]
+        else:
+            secrets = [
+                "noobaa-root-master-key-volume",
+                "noobaa-root-master-key-backend",
+                "noobaa-admin",
+                "noobaa-operator",
+                "noobaa-server",
+                "noobaa-endpoints",
+            ]
```
PrasadDesala: `noobaa-root-master-key-backend` will not be present on kms enabled. Instead of repeating the list again, you could modify it to below,

secrets = [
        "noobaa-root-master-key-volume",
        "noobaa-admin",
        "noobaa-operator",
        "noobaa-server",
        "noobaa-endpoints",
    ]
if not is_kms_enabled():
        secrets.append("noobaa-root-master-key-backend")

PrasadDesala (approved):

vkathole (approved):

PR #13544, opened 2025-11-06
Stabilize test_ceph_osd_stopped

fbalak:
https://issues.redhat.com/browse/DFBUGS-360 was closed as Can't do. There is a small change in timestamp when osd alerts are raised and this needs to be reflected in the test.


fbalak:
Verified: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/59925/

PR #13560, opened 2025-11-09
Stop skipping AMQ setup in FIPS clusters

sagihirshfeld:
Bucket notifications tests (both MCG or RGW) are skipped on FIPS clusters because of this old bug which has been fixed for a while now: https://issues.redhat.com/browse/ENTMQST-3422

This PR just removes the skip, and increases the initial wait time in the `is_amq_pod_running` method in case we're in FIPS mode. 

The increase in timeout is required because the entity-operator pod of the Kafka cluster takes longer to initialize in FIPS mode, where stricter crypto routines slow down its TLS/SSL startup.


ypersky1980 (approved):
LGTM

udaysk23 (approved):

suchita-g (approved):

keesturam (approved):

PR #13587, opened 2025-11-11
Inject cluster name into the compute name to avoid OCPBUGS-64930

petr-balogh:
There is an issue when we have a tainted node because of bug: https://issues.redhat.com/browse/OCPBUGS-64930

Slack discussion:
https://ibm-systems-storage.slack.com/archives/C06DWD30QTZ/p1762789361407019

### DIFF
--- a/ocs_ci/cleanup/vsphere/cleanup.py
+++ b/ocs_ci/cleanup/vsphere/cleanup.py
@@ -157,7 +157,11 @@ class IPAM(object):
         # IPAM server. In that we have to delete the IP's in IPAM server
         # even though resource pool is not created in Datacenter
         if not nodes:
-            node_type = ["compute", "control-plane", "lb"]
+            node_type = [
+                f"compute-{cluster_name}",
+                f"control-plane-{cluster_name}",
+                "lb",
+            ]
             for each_type in node_type:
                 if each_type == "lb":
                     nodes.append(f"{each_type}-0")
--- a/ocs_ci/deployment/vmware.py
+++ b/ocs_ci/deployment/vmware.py
@@ -69,6 +69,7 @@ from ocs_ci.utility.utils import (
     create_directory_path,
     read_file_as_str,
     replace_content_in_file,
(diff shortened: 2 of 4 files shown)


petr-balogh:
Verification job:
https://url.corp.redhat.com/d84b975

petr-balogh:
New verification:
 https://url.corp.redhat.com/9d72352

petr-balogh:
New verification job:
 https://url.corp.redhat.com/e0a1f33

petr-balogh:
Screenshot 2025-11-11 at 20 24 38
Working as expected here:
 https://url.corp.redhat.com/7d81285

petr-balogh:
Tier-1 job:
https://url.corp.redhat.com/0400b86

ocs_ci/cleanup/vsphere/cleanup.py
```diff
@@ -157,7 +157,11 @@ def delete_ips_upi(self, cluster_name):
         # IPAM server. In that we have to delete the IP's in IPAM server
         # even though resource pool is not created in Datacenter
         if not nodes:
-            node_type = ["compute", "control-plane", "lb"]
+            node_type = [
+                f"compute-{cluster_name}",
+                f"control-plane-{cluster_name}",
+                "lb",
+            ]
```
dahorak, line 164: Shouldn't we preserve also the original values `"compute"` and `"control-plane"` there, so the cleanup script will work for clusters deployed both by the previous version and the new one?

dahorak (approved):
Overall LGTM, I have just two questions:
* for IPI deployment, the VM name schema is: `<cluster-name>-<node-type>-<sufix>` (e.g. `j359vi1cs33uba-ztkxh-master-0`), here we are changing the UPI deployment to use VM name shcema `<node-type>-<clsuter-name>-<suffix>`, wouldn't make sense to use the cluster name in the beginning same as it is used by IPI deployment?
* And second question is about the load balancer VM, would it be possible to rename also the `lb-*` vm?

PR #13588, opened 2025-11-11
AMQ_fix

tchaitan-web:


ocs_ci/ocs/amq.py
```diff
@@ -159,9 +159,45 @@ def setup_amq_cluster_operator(self, namespace=constants.AMQ_NAMESPACE):
...
+                            if not crds:
+                                log.info("All Strimzi/Kafka CRDs successfully deleted. Proceeding with retry.")
+                                break
+                            log.warning(f"CRDs still present (possibly terminating): {crds}")
+                            time.sleep(5)
+                        else:
+                            log.warning("Proceeding with creation despite lingering CRDs.")
+                        # Give etcd a moment to settle before recreating CRDs
+                        time.sleep(5)
+                        log.info("Retrying Strimzi CRD creation after cleanup confirmation...")
```
Akarsha-rai, line 187: Can we handle this inside the cleanup only?

ocs_ci/ocs/amq.py
```diff
@@ -1037,6 +1077,167 @@ def cleanup(
             switch_to_default_rook_cluster_project()
             run_cmd(f"oc delete project {kafka_namespace}")
             self.ns_obj.wait_for_delete(resource_name=kafka_namespace, timeout=90)
+        # NEW SECTION: Clean up leftover Strimzi CRDs, ClusterRoles, and RoleBindings
+        log.info("Checking and cleaning up Strimzi-related CRDs and RBAC roles...")
```
Akarsha-rai, line 1081: Could we implement a dedicated function to manage specific resources—such as CRDs, ClusterRoles, and ClusterRoleBindings? This function would be modified to check for the existence of these particular resources. If found, it would proceed to delete them and then wait until their deletion is confirmed.

ocs_ci/ocs/amq.py
```diff
@@ -1037,6 +1077,167 @@ def cleanup(
...
+            log.info("Waiting for namespace 'myproject' to terminate completely...")
+            run_cmd(
+                "oc delete ns myproject --ignore-not-found=true --wait=true --timeout=180s",
+                shell=True,
+                ignore_error=True
+            )
+        except Exception as e:
+            log.warning(f"Namespace termination wait failed: {e}")
+
+        # Ensure 'myproject' namespace recreation safety before retry
```
Akarsha-rai, line 1194: Namespace deletion is expected at line 1078. If deletion fails, maybe we add a step to catch the error and attempt a force delete, especially if there are pending resources. I don't believe recreating and verifying the namespace deletion is required.

tchaitan-web:
This is duplicate PR 13515/
@Akarsha-rai I have incorporated your review comments in 13515 PR

Issue #13591, opened 2025-11-11
[Bug]: RGW Backingstores and Namespacestores fail to start because of wrong keys in secret


sagihirshfeld:
### Description
https://github.com/red-hat-storage/ocs-ci/pull/13489 had a side effect of changing the prefix of the keys in the secret that stores the `access_key` and `secret_key` for the cloud manager. 

Instead of:
```
$ oc extract secret/secret-cldmgr-rgw --to=-
# AWS_SECRET_ACCESS_KEY
...
# AWS_ACCESS_KEY_ID
...
```
We're creating a secret with these keys:
```
$ oc extract secret/secret-cldmgr-rgw --to=-
# RGW_SECRET_ACCESS_KEY
...
# RGW_ACCESS_KEY_ID
...
```

The result is that the Backingstore/Namespacestore never becomes ready:
```
apiVersion: noobaa.io/v1alpha1
kind: NamespaceStore
metadata:
...
  name: rgw-ns-store-cb88cb456b674c04b9b39ba0d6d
...
spec:
  s3Compatible:
    endpoint: https://rook-ceph-rgw-ocs-storagecluster-cephobjectstore.openshift-storage.svc:443
    secret:
      name: secret-cldmgr-rgw-bbc56ebb6cb84553ba6e38
      namespace: openshift-storage
    signatureVersion: v2
    targetBucket: rgw-uls-0b965cf3e2fb4e0fb437fdc2476b0ce1
  type: s3-compatible
status:
  conditions:
...
  - lastHeartbeatTime: "2025-11-07T22:56:38Z"
    lastTransitionTime: "2025-11-07T22:49:00Z"
    message: CheckExternalConnection Status=UNKNOWN_FAILURE Error=CredentialsError
      Message=Missing credentials in config, if using AWS_CONFIG_FILE, set AWS_SDK_LOAD_CONFIG=1
    reason: TemporaryError
    status: "False"
    type: Upgradeable
  mode: {}
  phase: Creating
```

### Additional Context
RP references:
- https://url.corp.redhat.com/f6bd9a9

sagihirshfeld:
This issue together with https://issues.redhat.com/browse/DFBUGS-4549 kills many of our tests at the health check of noobaa since it kills the default backingstore.

PR #13592, opened 2025-11-11
Restore RGW client data prefix and refactor cloud manager

sagihirshfeld:
### Changes 
1. Revert the change that was made in https://github.com/red-hat-storage/ocs-ci/pull/13489
2. Decouple the management of S3 clients of AWS, IBM and RGW by creating a separate class for each one. Using the same S3Client class for all three is what caused some of the recent automation bugs.
3. Simplify the CloudManager constructor by changing the following:
- Remove the `obc_obj` argument and its usage. Only `test_namespacestore_with_rgw` used it, so this requires adjusting the test with an alternative solution.
- Move the RGW and AWS STS client creation logic to their specific classes

Fixes https://github.com/red-hat-storage/ocs-ci/issues/13591

### DIFF
--- a/ocs_ci/ocs/resources/cloud_manager.py
+++ b/ocs_ci/ocs/resources/cloud_manager.py
@@ -347,6 +347,7 @@ class AwsClient(S3Client):
 
     """
 
+    @config.run_with_provider_context_if_available
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


ypersky1980 (approved):
LGTM

PR #13597, opened 2025-11-12
`test_guaranteed_bucket_logs_management`: Increase the timeout of `wait_for_logs_pvc_mount_status` from 120 to the default 300

sagihirshfeld:
`wait_for_logs_pvc_mount_status` sometimes fails because one of the old noobaa-endpoint pods is still up and not mounting the needed volume:
```
15:27:21 - MainThread - ocs_ci.ocs.resources.bucket_logging_manager - WARNING  - One of the noobaa pods still doesn't mount the logs PVC: {'noobaa-core-0': True, 'noobaa-endpoint-5cf48746bb-d4kjk': False, 'noobaa-endpoint-6c68585f6b-92qmc': True, 'noobaa-endpoint-6c68585f6b-bqhg5': True}
```
This doesn't happen in `test_bucket_logs_integrity` where the same timeout is set to the default 300 seconds.

### DIFF
--- a/tests/functional/object/mcg/test_bucket_logs.py
+++ b/tests/functional/object/mcg/test_bucket_logs.py
@@ -119,7 +119,7 @@ class TestBucketLogs(MCGTest):
 
         # 3. Wait for the nb pods to have mounts to logs PVC
         assert logs_manager.wait_for_logs_pvc_mount_status(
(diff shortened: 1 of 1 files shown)


ypersky1980 (approved):
LGTM

keesturam (approved):

PR #13601, opened 2025-11-12
Update test_odf_resources_requests_and_limits.py

s-selwin:
updated test steps to ignore test pods as we don't need to check these pods for request and limit values.


s-selwin:
@yitzhak12 all DCO & PR checks passed

yitzhak12 (approved):

s-selwin:
test_run

PR #13611, opened 2025-11-12
Move annotation in create_service_exporter function

petr-balogh:
It's documented here:
https://docs.redhat.com/en/documentation/red_hat_openshift_data_foundation/4.19/html-single/configuring_openshift_data_foundation_disaster_recovery_for_openshift_workloads/index#connecting-storage-clusters-with-the-ocs-provider-server-serviceexport_rdr And this is needed to be done also for ROKS deployment so it can be moved to create_service_exporter function to do at once for all cases, also for those when we install ODF via addon.

### DIFF
--- a/ocs_ci/helpers/dr_helpers.py
+++ b/ocs_ci/helpers/dr_helpers.py
@@ -2554,9 +2554,13 @@ def check_storage_cluster_peer_state():
         return False
 
 
-def create_service_exporter():
+def create_service_exporter(annotate=True):
     """
     Create Service exporter
+
+    Args:
+        annotate (bool): If True - annotate the service exporter
+
     """
     restore_index = config.cur_index
     managed_clusters = get_non_acm_cluster_config()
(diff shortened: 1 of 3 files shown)


ocs_ci/deployment/deployment.py
```diff
@@ -1887,20 +1887,6 @@ def deploy_ocs_via_operator(self, image=None):
...
-            api_server_exported_address_annotation = {
-                "ocs.openshift.io/api-server-exported-address": (
-                    f'{config.ENV_DATA["cluster_name"]}.'
-                    f"ocs-provider-server.openshift-storage.svc.clusterset.local:50051"
-                )
-            }
-            merge_dict(
-                cluster_data,
-                {"metadata": {"annotations": api_server_exported_address_annotation}},
-            )
```
vavuthu: since this part is moved to create_service_exporter(), do we need to call create_service_exporter() or is it not needed here

petr-balogh: It's not needed here

ocs_ci/helpers/dr_helpers.py
```diff
@@ -2529,6 +2533,12 @@ def create_service_exporter():
             continue
         logger.info("Creating Service exporter")
         run_cmd(f"oc create -f {constants.DR_SERVICE_EXPORTER}")
+        if annotate:
+            run_cmd(
+                "ooc annotate storagecluster ocs-storagecluster -n openshift-storage"
```
petr-balogh: typo
```suggestion
                "oc annotate storagecluster ocs-storagecluster -n openshift-storage"
```

dahorak (approved):

PR #13640, opened 2025-11-16
`test_check_object_integrity` - skip if d/c in all pytest params

sagihirshfeld:
The test tries to download test objects from AWS, and while in some environments our "disconnected" clusters still have access to AWS, in others it doesn't. 

This PR makes sure the test is being skipped so we won't get a false failure in those environments.


sagihirshfeld:
Adding the `Verified` label since this PR just adds additional skips.

Issue #13671, opened 2025-11-18
[Bug]: test_csi_addon_pod_security failed on version less than 4.20


suchita-g:
### Description
CSI addon pods are available on ODF4.20 and above Hence test_csi_addon_pod_security failed on version less than 4.20

### Steps to Reproduce
1.
2.
3.
Tier1 https://hyc-icps-team-jenkins.swg-devops.com/job/odf-qe/job/hci-tests-execution/320/testReport/
### Environment
- Test Suite(s):
- Platform(s):
- Version(s):
- OS:

suchita-g:
PR approved . Closing this issue.

PR #13672, opened 2025-11-18
Fix: #13671 Skip the test for ocs version less than 4.20

suchita-g:
Signed-off-by: suchita-g <sgatfane@redhat.com>

CSI addons pods are avaialble on ODF4.20 and later so need to skip this test on ocs version lower than 4.20 
Fixes: issue #13671


anajain16 (approved):

suchita-g:
verified locally as per the issue #13671.

PR #13683, opened 2025-11-18
fix broken download hcp kubeconfig 420

DanielOsypenko:
This pr fixes the issue for running a job on provider with hosted clusters, when we do not provide client configurations.
This approach we use in local setups, but do not use on Multicluster Provider mode job.


fbalak (approved):

dahorak (approved):

DanielOsypenko:
pr verification: https://url.corp.redhat.com/dd95dee

PR #13699, opened 2025-11-19
Bump python-ipmi to 0.5.7

petr-balogh:
Avoiding issues like:
AttributeError: 'array.array' object has no attribute 'tostring'
In newer version of python , as this method was deprecated and dropped
already in python 3.9.


petr-balogh:
Verification job:
 https://url.corp.redhat.com/dd1ffa2

PR #13737, opened 2025-11-24
Disable old BM locking mechanism (bm_status)

dahorak:
One of the BM environment were still using some custom locking mechanism together with Resource Locker service. This causes occasional issues, when the locking fails for some reason even though the env. is correctly reserved in Resource locker.


dahorak:
Triggered verification job: https://url.corp.redhat.com/9e48a1d

PR #13748, opened 2025-11-24
WIP: Allow provide VIPs for IPI deployment to skip IPAM IP Reservation

petr-balogh:
For enabling vSphere9 testing on custom vSphere env


vavuthu:
can we update vips in README.md

PR #13780, opened 2025-11-26
Fix missing resource_name fallback in wait_for_resource_oc_wait()

AviadP:
Add resource_name = resource_name or self.resource_name fallback to match the existing pattern used for selector and the legacy path. Without this, in rare cases,  `oc wait` commands fail with "no name was specified" when resource_name is set on the OCP constructor but not passed to wait_for_resource().

### DIFF
--- a/ocs_ci/ocs/ocp.py
+++ b/ocs_ci/ocs/ocp.py
@@ -792,6 +792,7 @@ class OCP(object):
             )
 
         selector = selector or self.selector
+        resource_name = resource_name or self.resource_name
         command = f"wait {self.kind} {resource_name} --for=jsonpath='{{.status.phase}}'={condition}"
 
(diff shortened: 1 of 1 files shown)


AviadP:
> PR validation
> 
> Cluster Name: apolak-nov23 Cluster Configuration: PR Test Suite: tier1 PR Test Path: tests/ Additional Test Params: OCP VERSION: 4.20 OCS VERSION: 4.20 tested against branch: master
> 
> [Job UNSTABLE](https://url.corp.redhat.com/aef6f8b) (some or all tests failed).

failures is not  related to the change in this PR. safe to merge

DanielOsypenko (approved):

ebenahar (approved):

PR #13812, opened 2025-12-02
ISCSI configuration script

avd-sagare:


ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,286 @@
...
+
+log = logging.getLogger(__name__)
+
+
+USERNAME = "root"
+key_path = "~/.ssh/openshift-dev.pem"
+key_path = os.path.expanduser(key_path)
+
+TARGET_IQN = "iqn.2003-01.org.linux-iscsi.localhost.x8664:sn.d7a7c8437192"
+TARGET_IP = "10.1.161.239"
```
PrasadDesala: I would suggest not to hardcode the values and make them visible in public projects

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+USERNAME = "root"
+key_path = "~/.ssh/openshift-dev.pem"
+key_path = os.path.expanduser(key_path)
+
+TARGET_IQN = "iqn.2003-01.org.linux-iscsi.localhost.x8664:sn.d7a7c8437192"
```
PrasadDesala: Do we need to a maintain a separate python script for iscsi config? Do you think it is a good idea to move it under ocs-ci dir?

scripts/python/iscsi_config.py
```diff
@@ -0,0 +1,286 @@
...
+    cmd = [
+        "oc",
+        "--kubeconfig",
+        kubeconfig_path,
+        "get",
+        "nodes",
+        "-l",
+        "node-role.kubernetes.io/worker",
+        "-o",
+        "json",
```
PrasadDesala: How can the kubeconfig details be loaded into this Python script?

Please look at my above comment. This can be avoided moving this file into the ocs-ci directory and run using run-ci

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,286 @@
...
+            if addr["type"] == "InternalIP":
+                worker_ips.append(addr["address"])
+
+    return worker_ips
+
+
+# --------------------------------------------------------
+# STEP 1: Collect worker node initiator IQNs
+# --------------------------------------------------------
+def get_worker_iqns(worker_node_ips):
```
PrasadDesala: you can get the worker node iqns using our ocs-ci libs

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,286 @@
...
+        "mount -t configfs configfs /sys/kernel/config 2>/dev/null || true",
+        # Load kernel modules
+        "modprobe target_core_mod",
+        "modprobe iscsi_target_mod",
+        # Start and enable target service
+        "systemctl start target",
+        "systemctl enable target",
+        # Verify setup
+        "systemctl status target --no-pager",
+    ]
```
PrasadDesala: there is no need to load the modules, this can be removed

scripts/python/iscsi_config.py
```diff
@@ -0,0 +1,286 @@
...
+    for node_ip in worker_node_ips:
+        check_iscsi_command = "which iscsiadm"
+        iscsiadm_installed = ssh_run(node_ip, check_iscsi_command, username="core")
+        if not iscsiadm_installed:
+            log.info(f"[{node_ip}] iscsiadm not found. Installing...")
+            ssh_run(
+                node_ip,
+                "sudo yum install -y iscsi-initiator-utils || sudo apt install -y open-iscsi",
+                username="core",
+            )
```
PrasadDesala: it will be installed by default during the OCP cluster deployment. It is not required

scripts/python/iscsi_config.py
```diff
@@ -0,0 +1,286 @@
...
+        # Add ACL
+        acl_cmd = f"targetcli /iscsi/{target_iqn}/tpg1/acls create {worker_iqn}"
+        ssh_run(target_ip, acl_cmd, username)
+        log.info(f"AddedACL for worker IQN: {worker_iqn}")
+
+        # Get list of LUNs to map
+        lun_list_cmd = f"targetcli /iscsi/{target_iqn}/tpg1/luns ls"
+        success, stdout, stderr = ssh_run(target_ip, lun_list_cmd, username)
+
+        import re
```
PrasadDesala: move it to the top

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,286 @@
...
+    if success:
+        log.info("Configuration saved successfully")
+    else:
+        log.warning(f"Save warning: {stderr}")
+
+
+# --------------------------------------------------------
+# STEP 3: Configure iSCSI Initiator on each Worker
+# --------------------------------------------------------
+def configure_initiators(worker_ip):
```
PrasadDesala: you can use our ocs-ci libs

scripts/python/iscsi_config_removal.py
```diff
@@ -0,0 +1,187 @@
...
+)
+log = logging.getLogger(__name__)
+
+
+USERNAME = "root"
+key_path = "~/.ssh/openshift-dev.pem"
+key_path = os.path.expanduser(key_path)
+
+
+def remove_acls_from_target(target_vm_ip, target_iqn, worker_iqns, username):
```
PrasadDesala: same as my previous comment, you can create a file under ocs-ci and handle the teardown in the same file

scripts/python/iscsi_config_removal.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # Check ACLs on target
+    log.info("\nChecking target ACLs...")
+    check_cmd = f"targetcli /iscsi/{target_iqn}/tpg1/acls ls 2>&1"
+    success, stdout, stderr = ssh_run(target_vm_ip, check_cmd, username_target)
+
+    if "No ACLs" in stdout or not stdout:
+        log.info(" No ACLs remaining")
+    else:
+        log.info(f" ACLs: {stdout}")
```
PrasadDesala: you should fail the deplopyment if you see any teardown issues

ocs_ci/utility/iscsi_config.py
avd-sagare: removed

ocs_ci/utility/iscsi_config.py
avd-sagare: Done

ocs_ci/utility/iscsi_config.py
avd-sagare: Done

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,404 @@
...
+
+log = logging.getLogger(__name__)
+
+
+USERNAME = config.ENV_DATA.get("iscsi_target_username")
+key_path = "~/.ssh/openshift-dev.pem"
+key_path = os.path.expanduser(key_path)
+
+TARGET_IQN = config.ENV_DATA.get("iscsi_target_iqn")
+TARGET_IP = config.ENV_DATA.get("iscsi_target_ip")
```
dahorak: I would propose to not use those parameters as global variables (and load the values from config here), because it will process them in the time of loading the module (not when the particular function will be actually called), which might lead to a situation, when the config is not fully initialized or some other unexpected behaviour.
Instead load those parameters inside the functions.

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,404 @@
...
+
+BACKSTORES = ["disk0", "disk1", "disk2"]  # Already created on target
+MOUNT_BASE = "/mnt/iscsi_lun"
```
dahorak: Similarly to the above comment, not sure, if those parameters should be configured there or if it will be better to set them in `constants` module (or if there might be a need to change them, then in `defaults` module with an option to change it via configuration or function/command line parameters).

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,404 @@
...
+            log.info(out)
+        if err:
+            log.error("ERROR:", err)
+        return True, out, err
+
+    except Exception as e:
+        log.error(f"SSH ERROR on {host}: {e}")
+        return None
+    finally:
+        client.close()
```
dahorak: Wouldn't be possible to use the existing [`connection` module](https://github.com/red-hat-storage/ocs-ci/blob/c2c7a239626044e9edbb36a581065fa00848593d/ocs_ci/utility/connection.py#L14)?

ocs_ci/deployment/deployment.py
```diff
@@ -972,6 +973,17 @@ def post_ocp_deploy(self):
                 logger.warning(
                     f"Ingress Node Firewall deployment and SSH access to nodes restriction failed: {err}"
                 )
+        # Setup iSCSI if configured
+        if config.ENV_DATA.get("iscsi_target_ip") and config.ENV_DATA.get(
+            "iscsi_target_iqn"
+        ):
```
dahorak: Here I would propose to create one additional parameter (something like `iscsi_setup`, which will be by default `False` and this condition will be based on that new parameter (similarly as it is few lines above for configuring Ingress Node Firewall:
```python
        # configure Ingress Node Firewall and restrict SSH access to nodes
        if config.ENV_DATA.get("restrict_ssh_access_to_nodes", False):
```
or few more lines above for configuring custom ingress and api ssl certificate:
```python
            if config.DEPLOYMENT.get("use_custom_ingress_ssl_cert"):
                configure_custom_ingress_cert()
            if config.DEPLOYMENT.get("use_custom_api_ssl_cert"):
                configure_custom_api_cert()
```

Then you can create new configuration file for enabling this functionality, for example like this: https://github.com/red-hat-storage/ocs-ci/blob/master/conf/ocsci/custom-ingress-ssl-certificate.yaml
(this one is for configuring custom ingress ssl certificate and the main part here is the `use_custom_ingress_ssl_cert: true` configuration, but it also contain few other commented parameters, which could help with custom configuration.

The reason is, that we want to configure the `iscsi_target_ip`, `iscsi_target_iqn` (and the other) parameters as part of the jenkins job (so they will not be exposed to this public ocs-ci repository) - but that means, that those parameters will be configured for *every* triggered job. So the additional configuration (`iscsi_setup` or something like that) will actually control if the iscsi setup will be performed or not.
(Theoretically it is possible to pass the `iscsi_target_*` parameters from the jenkins job conditionally, but it will add more complexity and require much more changes in the jobs itself, which is I think unnecessary. So better solution is the proposal above.)

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,812 @@
...
+import argparse
+
+from ocs_ci.ocs import node, ocp, constants
+from ocs_ci.framework import config
+from ocs_ci.utility.connection import Connection
+
+log = logging.getLogger(__name__)
+
+
+key_path = "~/.ssh/openshift-dev.pem"
```
dahorak: This `key_path` variable is not used anywhere in the module.

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,812 @@
...
+                node_data = ocp_node_obj.get(resource_name=node_name)
+                node_ip = (
+                    node_data.get("status", {}).get("addresses", [{}])[0].get("address")
+                )
+
+                if node_ip:
+                    node_ssh = Connection(
+                        host=node_ip,
+                        user="core",
+                        private_key=f"{os.path.expanduser('~')}/.ssh/openshift-dev.pem",
```
dahorak: I think you can use `config.DEPLOYMENT["ssh_key_private"]` instead of hardcoding it here (see https://github.com/red-hat-storage/ocs-ci/blob/master/conf/README.md#deployment).
The [default value is this](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/conf/default_config.yaml#L72): `~/.ssh/openshift-dev.pem`

This applies for other occurrences of the same.

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,812 @@
...
+
+def get_worker_node_ips(kubeconfig_path):
+    """
+    Collects worker node IPs from the Kubernetes cluster using oc command.
+
+    Parameters:
+    kubeconfig_path (str): Path to the kubeconfig file.
+
+    Returns:
+    list: A list of IP addresses of worker nodes or an empty list on failure.
```
dahorak: The list of parameters and returned value should be moved one more level to the right and there should be an empty space after a list (which means also after the Return section):

```suggestion
    Parameters:
        kubeconfig_path (str): Path to the kubeconfig file.

    Returns:
        list: A list of IP addresses of worker nodes or an empty list on failure.

```

Similarly for other docstrings.

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,812 @@
...
+    This function ensures the iSCSI service is started before reading the IQN.
+
+    Parameters:
+    worker_node_names (list): List of worker node names (strings).
+
+    Returns:
+    list: A list of IQNs of worker nodes or an empty list on failure.
+    """
+    iqns = []
+    log.info("\n=== Collecting Worker IQNs ===")
```
dahorak:
```suggestion
    log.info("=== Collecting Worker IQNs ===")
```

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,812 @@
...
+    return iqns
+
+
+# --------------------------------------------------------
+# STEP 2: Configure Target VM
+# --------------------------------------------------------
+def configure_target(target_node_ssh, target_iqn, worker_iqns):
+    """
+    Configures the iSCSI target with given IQNs and IP.
+    This function is a placeholder and requires implementation.
```
dahorak: Is this comment still valid or is it just leftover?

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,812 @@
...
+    target_node_ssh (object): An established SSH connection to the target VM.
+    target_iqn (str): Target iSCSI IQN.
+    worker_iqns (list): List of IQNs of worker nodes.
+    """
+
+    # Setup environment first
+    # setup_target_environment(target_node_ssh)
+    # Check if target exists
+    check_cmd = f"targetcli ls /iscsi/{target_iqn} 2>/dev/null"
+    success, stdout, stderr = target_node_ssh(check_cmd)
```
dahorak: There seems to be missing `exec_cmd()` method (similarly on other places).

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,812 @@
...
+        log.error(f"Error during iSCSI verification: {e}")
+        import traceback
+
+        traceback.print_exc()
+        return {"error": str(e)}
+
+
+def main():
+    "main Function to setup iscsi target and initiators"
+    args = init_arg_parser()
```
dahorak: If you want to use the script from command line, how are you passing the command line arguments to the `iscsi_setup()` or `iscsi_teardown()` functions?

tests/functional/deployment/test_iscsi_setup.py
```diff
@@ -0,0 +1,232 @@
...
+
+from ocs_ci.framework import config
+from ocs_ci.framework.testlib import deployment
+from ocs_ci.utility.iscsi_config import verify_iscsi_setup
+
+log = logging.getLogger(__name__)
+
+
+@deployment
+def test_iscsi_setup_verification():
```
dahorak: I forgot this yesterday, there is missing squad decorator for those tests (that is the reason for PR check failure):
```
  INTERNALERROR> Missing squad decorator for the following test items: {
  INTERNALERROR>     "test_iscsi_setup_verification": "/home/runner/work/ocs-ci/ocs-ci/tests/functional/deployment/test_iscsi_setup.py",
  INTERNALERROR>     "test_iscsi_connectivity": "/home/runner/work/ocs-ci/ocs-ci/tests/functional/deployment/test_iscsi_setup.py",
  INTERNALERROR>     "test_iscsi_sessions": "/home/runner/work/ocs-ci/ocs-ci/tests/functional/deployment/test_iscsi_setup.py",
  INTERNALERROR>     "test_iscsi_devices": "/home/runner/work/ocs-ci/ocs-ci/tests/functional/deployment/test_iscsi_setup.py"
  INTERNALERROR> }
```

Shrivaibavi:
@avd-sagare Can you please add in the description of what this script is trying to achieve ?

conf/ocsci/iscsi_deployment.yaml
```diff
@@ -0,0 +1,2 @@
+DEPLOYMENT:
+  iscsi_setup: True
```
dahorak, line 2: please mention this parameter also in `conf/README.md`

ocs_ci/deployment/deployment.py
```diff
@@ -2204,6 +2214,20 @@ def destroy_cluster(self, log_level="DEBUG"):
         Args:
             log_level (str): log level for installer (default: DEBUG)
         """
+        # Teardown iSCSI if configured (before destroying cluster)
+        if config.ENV_DATA.get("iscsi_target_ip") and config.ENV_DATA.get(
+            "iscsi_target_iqn"
+        ):
```
dahorak: This condition should be similar to the above one:

```suggestion
        if config.ENV_DATA.get("iscsi_setup", False):
```

conf/ocsci/iscsi_deployment.yaml
avd-sagare: Done

vavuthu:
@avd-sagare I had a discussion with petr and he suggested below and I tried locally which is working

pytest.mark.skipif is not intended for testcase selection , so we need to use compose like below 

```
iscsi_setup_required = pytest.mark.iscsi_setup_required
iscsi_cluster_needed = pytest.mark.skipif(
    not config.ENV_DATA.get("iscsi_setup"),
    reason="Test runs only when iscsi_setup is enabled in ENV_DATA",
)
iscsi_setup_required = compose(iscsi_setup_required, iscsi_cluster_needed)
```

petr-balogh:
In case we really want to use this marker for test selection as well as for skipping.
Let's do it like this:
```
# Do not use directly _iscsi_cluster_needed, but use iscsi_setup_required!
_iscsi_cluster_needed = pytest.mark.skipif(
    not config.ENV_DATA.get("iscsi_setup"),
    reason="Test runs only when iscsi_setup is enabled in ENV_DATA",
)
iscsi_setup_required = compose(pytest.mark.iscsi_setup_required
, _iscsi_cluster_needed)
```

And also register only: `iscsi_setup_required` in pytest.ini

avd-sagare:
PR validation - (https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-fdf-cluster/556/)]

ocs_ci/framework/pytest_customization/marks.py
```diff
@@ -538,6 +538,7 @@ def setup_multicluster_marker(marker_base, push_missing_configs=False):
     reason="Test will run on External Mode cluster only",
 )
 
+
```
dahorak: This empty line here is probably leftover.

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,933 @@
...
+        if stdout:
+            log.debug(f"Output: {stdout.strip()}")
+
+
+def get_worker_node_ips():
+    """
+    Collects worker node IPs from the Kubernetes cluster using oc command.
+
+    Parameters:
+        kubeconfig_path (str): Path to the kubeconfig file.
```
dahorak: This function doesn't have any parameter defined, so this is probably leftover.

Also the Returns list (or any other lists in the docstring) should end up with one empty line - which applies also for the case, when it is the end of the docstring (same apply for other functions in this PR as well).

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,933 @@
...
+    for node_name in worker_node_names:
+        log.info(f"Getting IQN from worker node {node_name}...")
+
+        try:
+            # First, ensure iSCSI service is started
+            start_service_cmd = (
+                "systemctl start iscsid 2>/dev/null || "
+                "systemctl enable --now iscsid 2>/dev/null || "
+                "true"
+            )
```
dahorak: Since this variable have the same content as the following `start_cmd` variable, what about setting it up only once even before the `for` loop (and maybe the same apply for the follow-up `grep InitiatorName...` command)?

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,933 @@
...
+            # Fallback to SSH if oc debug fails
+            try:
+                # Get node IP from OCP API
+                ocp_node_obj = ocp.OCP(kind=constants.NODE)
+                node_data = ocp_node_obj.get(resource_name=node_name)
+                node_ip = (
+                    node_data.get("status", {}).get("addresses", [{}])[0].get("address")
+                )
+
+                if node_ip:
```
dahorak, line 259: If for some reason the `node_ip` will be `None` or empty, it will silently skip the following code without any notification/log message/exception/etc. Is that intentional? Wouldn't it make the debugging quite difficult?

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,933 @@
...
+                            log.debug(f"Command output: {stdout}")
+            except Exception as fallback_error:
+                log.error(f"SSH fallback also failed for {node_name}: {fallback_error}")
+
+
+def remove_acls_from_target(target_node_ssh, target_iqn, worker_iqns, username):
+    """
+    Remove ACLs and LUN mappings from target using targetcli.
+    """
+    log.info("\n" + "=" * 70)
```
dahorak: Please remove the new lines from the beginning of the log messages, because it breaks the log message schema and makes it harder to read/parse, etc.
Also I'm not a big fan of the "frames" (`=====...` and similar) around the messages, especially in the cases like here, when it is just few minor steps to set up or remove some configuration.

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,933 @@
...
+                            log.warning(f"Command failed on {node_ip}: {stderr}")
+                        else:
+                            log.debug(f"Command output: {stdout}")
+            except Exception as fallback_error:
+                log.error(f"SSH fallback also failed for {node_name}: {fallback_error}")
+
+
+def remove_acls_from_target(target_node_ssh, target_iqn, worker_iqns, username):
+    """
+    Remove ACLs and LUN mappings from target using targetcli.
```
dahorak, line 283: Missing list of parameters

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,933 @@
...
+    target_node_ssh.exec_cmd(save_cmd)
+    log.info("\n Configuration saved")
+
+
+def wipe_luns_on_target(target_node_ssh, target_iqn, username):
+    """
+    Wipe LUN data on target VM.
+    Also deletes IBM Spectrum Scale related resources.
+    Only wipes devices that are referenced in LocalDisks resources.
+    """
```
dahorak, line 333: Missing list of parameters (applies on other places as well).

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,933 @@
...
+    worker_node_names = get_worker_node_names()
+    if not worker_node_names:
+        log.warning("No worker nodes found, skipping validations")
+        return
+
+    log.info("Running iSCSI setup validations (connectivity, sessions, devices)...")
+    validate_iscsi_connectivity()
+    validate_iscsi_sessions()
+    validate_iscsi_devices()
+    log.info("All iSCSI setup validations passed.")
```
hnallurv, line 935: @avd-sagare I used Bob to get this file reviewed. Here are the comments from it:

**Positive Aspects ✅**

1. Good use of ocs-ci libraries - Uses node.get_worker_nodes(), ocp.OCP(), config.ENV_DATA
2. Comprehensive verification - Multiple validation functions ensure setup correctness
3. Detailed logging - Good use of log levels and informative messages
4. Fallback mechanisms - IP-based connection when oc debug fails
5. Cleanup functions - Proper teardown with verification
6. Configuration from ENV_DATA - Uses framework's configuration system

**Summary of Priority Fixes**
**Must Fix (Before Merge):**

1. Remove sys.exit() calls - use exceptions
2. Fix command injection vulnerabilities
3. Add input validation
4. Remove hardcoded credentials (already flagged by reviewer)
5. Consistent error handling strategy

Should Fix (High Priority):

1. Add type hints
2. Break down long functions
3. Extract magic numbers to constants
4. Add retry logic for network operations
5. Implement resource cleanup guarantees

Nice to Have:

1. Improve docstrings
2. Standardize logging format
3. Add dataclasses for complex return types
4. Reduce code duplication

The code shows good understanding of iSCSI concepts and ocs-ci framework, but needs refinement in error handling, security, and code organization before production use.

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,933 @@
...
+    Complete cleanup of iSCSI target.
+
+    Parameters:
+        target_vm_ip: Target VM IP address
+        target_iqn: Target IQN
+        worker_iqns: List of worker IQNs to remove
+        worker_ips: List of worker IP addresses
+        wipe_data: Whether to wipe data (default: True)
+        username_target: SSH user for target VM
+    """
```
dahorak, line 458: The docstring is not aligned with the actual list of parameters.

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,933 @@
...
+    Returns:
+        dict: Dictionary containing all verification results.
+    """
+    log.info("\n" + "=" * 70)
+    log.info("iSCSI SETUP VERIFICATION - START")
+    log.info("=" * 70)
+
+    if not config.ENV_DATA.get("iscsi_target_ip") or not config.ENV_DATA.get(
+        "iscsi_target_iqn"
+    ):
```
dahorak: This condition should be mainly based on `config.ENV_DATA.get("iscsi_setup")`, because the `iscsi_target_ip` and `iscsi_target_iqn` are automatically configured for all the jenkins jobs not only for those, where ISCSI setup should be applied.

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,933 @@
...
+    log.info(f"Device validation passed. Total devices found: {total_devices}")
+
+
+def run_iscsi_setup_validations():
+    """
+    Run all three iSCSI validations: connectivity, sessions, and devices.
+    Called at the end of iscsi_setup() and can be used by tests for post-deploy verification.
+    Raises AssertionError or ValueError on first validation failure.
+    """
+    if not config.ENV_DATA.get("iscsi_target_ip") or not config.ENV_DATA.get(
```
dahorak: Same as above, this condition should be mainly based on `config.ENV_DATA.get("iscsi_setup")`, because the `iscsi_target_ip` and `iscsi_target_iqn` are automatically configured for all the jenkins jobs not only for those, where ISCSI setup should be applied.

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,933 @@
...
+        return {"error": str(e)}
+
+
+# --------------------------------------------------------
+# Validation functions (run after iscsi_setup or from tests)
+# --------------------------------------------------------
+def validate_iscsi_connectivity():
+    """
+    Validate network connectivity to iSCSI target from all worker nodes.
+    Raises AssertionError if any worker node cannot connect.
```
dahorak: The `Raises` section should be formatted similarly as `Parameters` or `Returns` (list of the exceptions with description (applies also on other places).

ocs_ci/utility/iscsi_config.py
avd-sagare: Added else part.

ocs_ci/utility/iscsi_config.py
avd-sagare: Added parameters

ocs_ci/utility/iscsi_config.py
avd-sagare: Added parameters

ocs_ci/utility/iscsi_config.py
avd-sagare: aligned with the actual list of parameters

ocs_ci/utility/iscsi_config.py
avd-sagare: Incorporated Must Fix  and Should Fix comments

ocs_ci/utility/iscsi_config.py
```diff
@@ -304,14 +295,25 @@ def configure_initiators(worker_node_names):
                             log.warning(f"Command failed on {node_ip}: {stderr}")
                         else:
                             log.debug(f"Command output: {stdout}")
+                else:
+                    log.error(f"Node IP is none or empty for {node_name}")
+                    raise Exception(f"Node IP is none or empty for {node_name}")
```
dahorak: It is not a good practice to raise such generic `Exception`. Please try to use some [common python but more specific exception](https://docs.python.org/3.11/library/exceptions.html#exception-hierarchy) or [custom ocs-ci exception](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/exceptions.py).

avd-sagare:
Teardown is also completed successfully.
https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-destroy-ocs-cluster/71670/console

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,966 @@
...
+            # Fallback to IP-based connection if oc debug fails
+            try:
+                # Get node IP from OCP API
+                ocp_node_obj = ocp.OCP(kind=constants.NODE)
+                node_data = ocp_node_obj.get(resource_name=node_name)
+                node_ip = (
+                    node_data.get("status", {}).get("addresses", [{}])[0].get("address")
+                )
+
+                if node_ip:
```
dahorak, line 94: There is no `else` for this condition, which means that if `node_ip` will be None, it will silently skip this node. Is it expected?

avd-sagare: Added.

dahorak (dismissed):

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+                    if retcode == 0 and stdout.strip():
+                        iqns.append(stdout.strip())
+                        log.info(
+                            f"Worker {node_ip} IQN: {stdout.strip()} (via SSH fallback)"
+                        )
+                else:
+                    log.error(f"Node IP is none or empty for {node_name}")
+                    raise NodeNotFoundError(f"Node IP is none or empty for {node_name}")
+            except Exception as fallback_error:
+                log.error(f"SSH fallback also failed for {node_name}: {fallback_error}")
```
PrasadDesala, line 122: The else block is inside the inner try-except, but the exception will be caught by the outer except Exception as fallback_error block, making the error message misleading.

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+                        retcode, stdout, stderr = worker_node_ssh.exec_cmd(cmd)
+                        if retcode != 0:
+                            log.warning(f"Command failed on {node_ip}: {stderr}")
+                        else:
+                            log.debug(f"Command output: {stdout}")
+                else:
+                    log.error(f"Node IP is none or empty for {node_name}")
+                    raise NodeNotFoundError(f"Node IP is none or empty for {node_name}")
+            except Exception as fallback_error:
+                log.error(f"SSH fallback also failed for {node_name}: {fallback_error}")
```
PrasadDesala, line 278: same as my above comment.
The else block is inside the inner try-except, but the exception will be caught by the outer except Exception as fallback_error block, making the error message misleading.

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+from ocs_ci.utility.connection import Connection
+
+log = logging.getLogger(__name__)
+
+
+def setup_target_environment(target_node_ssh):
+    """
+    Set up target environment on remote target VM via SSH.
+
+    Parameters:
```
PrasadDesala: this should args

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+        # Verify setup
+        "systemctl status target --no-pager",
+    ]
+
+    for cmd in commands:
+        log.info(f"Executing on target VM: {cmd}")
+
+        retcode, stdout, stderr = target_node_ssh.exec_cmd(cmd)
+
+        if not retcode and "already" not in stderr.lower():
```
PrasadDesala: should it be retcode != 0?

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+    log.info("Configuration saved")
+
+
+def wipe_luns_on_target(target_node_ssh, target_iqn, username):
+    """
+    Wipe LUN data on target VM.
+    Also deletes IBM Spectrum Scale related resources.
+    Only wipes devices that are referenced in LocalDisks resources.
+
+    Parameters:
```
PrasadDesala: it should args

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+        target_node_ssh (object): An established SSH connection to the target VM.
+        target_iqn (str): Target iSCSI IQN.
+        username (str): Username for the target VM.
+
+    Returns:
+        None
```
PrasadDesala: If it is not returing anything, I think it can be removed from the doc string

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+
+        for sc in storageclasses.get("items", []):
+            provisioner = sc.get("provisioner", "")
+            if provisioner == "spectrumscale.csi.ibm.com":
+                sc_name = sc.get("metadata", {}).get("name")
+                if sc_name:
+                    try:
+                        sc_obj.delete(resource_name=sc_name, wait=False)
+                        log.info(f"Deleted StorageClass: {sc_name}")
+                    except Exception as e:
```
PrasadDesala, line 356: Silently swallowing exceptions is dangerous. Please handle specific exceptions or re-raise

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+                        log.warning(f"Failed to delete StorageClass {sc_name}: {e}")
+    except Exception as e:
```
PrasadDesala, line 358: Silently swallowing exceptions is dangerous. Please handle specific exceptions or re-raise

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+
+    Parameters:
+        target_node_ssh (object): An established SSH connection to the target VM.
+        target_iqn (str): Target iSCSI IQN.
+        worker_iqns (list): List of IQNs of worker nodes.
+
+    """
+
+    # Setup environment first
+    # setup_target_environment(target_node_ssh)
```
PrasadDesala: Please remove it if not required

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+    for node_name in worker_node_names:
+        log.info(f"Configuring initiator on worker node {node_name}...")
+
+        try:
+            for cmd in cmds:
+                stdout = ocp_obj.exec_oc_debug_cmd(
+                    node=node_name, cmd_list=[cmd], use_root=True, timeout=120
+                )
+                log.debug(f"Command output on {node_name}: {stdout}")
+        except Exception as e:
```
PrasadDesala: Please catch specific exceptions

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+                    node_data.get("status", {}).get("addresses", [{}])[0].get("address")
+                )
+
+                if node_ip:
+                    worker_node_ssh = Connection(
+                        host=node_ip,
+                        user="core",
+                        private_key=config.DEPLOYMENT["ssh_key_private"],
+                    )
+                    log.info(f"Using SSH fallback for {node_ip}...")
```
PrasadDesala, line 265: The SSH fallback pattern appears repeating, should we extract it to some helper function and use it here?

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+    """
+    Configures a worker node as an iSCSI initiator.
+    This function installs necessary packages, discovers targets, logs in to the target,
+    and enables the iSCSI service.
+
+    Parameters:
+        worker_node_names (list): List of worker node names (strings).
+
+    Returns:
+        None
```
PrasadDesala: please fix the doc strings at all the places in this file

ocs_ci/utility/iscsi_config.py
```diff
@@ -0,0 +1,967 @@
...
+            wipe_luns_on_target(target_node_ssh, target_iqn, username_target)
+
+        # Step 3: Verify cleanup
+        verify_cleanup(target_node_ssh, target_iqn, username_target, worker_iqns)
+
+        log.info("CLEANUP COMPLETED SUCCESSFULLY")
+
+    except Exception as e:
+        log.error(f"Cleanup failed: {e}")
+        import traceback
```
PrasadDesala: Please import it at the top of the file

ocs_ci/utility/iscsi_config.py
avd-sagare: It is required in enhancement. I will add todo.

ocs_ci/utility/iscsi_config.py
avd-sagare: It appeares only for twice. I guess  no need for seperate function.

ocs_ci/utility/iscsi_config.py
avd-sagare: Added at top

ocs_ci/utility/iscsi_config.py
avd-sagare: Fixed doc string

ocs_ci/utility/iscsi_config.py
avd-sagare: Added specific exception

ocs_ci/utility/iscsi_config.py
avd-sagare: Fixed doc string

ocs_ci/utility/iscsi_config.py
avd-sagare: changed

ocs_ci/utility/iscsi_config.py
avd-sagare: Fixed doc string

ocs_ci/utility/iscsi_config.py
avd-sagare: corrected exception

ocs_ci/utility/iscsi_config.py
avd-sagare: corrected exception

dahorak (approved):

avd-sagare:
configuration job - https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-fdf-cluster/621/ 
Teardown job - https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-destroy-ocs-cluster/72142/

PR #13832, opened 2025-12-03
verify_last_heartbeat_timestamp-skip-if-no-odf

DanielOsypenko:
remove heartbeat validation when ODF is not set on clients.\
By default we are passing same set of validations when ODF installed and not installed and originally we forgot to include skip logic for this validation.


dahorak (approved):

fbalak (approved):

dahorak:
Verified here: https://url.corp.redhat.com/e1c6aba

PR #13836, opened 2025-12-04
remove-rgw-check-on-aws-provider

DanielOsypenko:
rgw pod is not automatically deployed on cloud provider platforms


DanielOsypenko:
> The condition looks ok but I see that constants.ON_PREM_PLATFORMS (https://github.com/red-hat-storage/ocs-ci/blob/ee47d2e8d82b7d8eb1ecb023813199f1d1866f86/ocs_ci/ocs/constants.py#L1707C1-L1707C18) is usually used for rgw checks.

thank you, fixed that

PR #13838, opened 2025-12-04
Assisted instaler deployment: set platform

dahorak:
There is an option to explicitly pass `platform` parameter when creating cluster via Assisted Installer. It looks like it is by default set to `baremetal`, which breaks ODF deployment on vSphere.


vavuthu (approved):

DanielOsypenko (approved):

dahorak:
Verification job (failed during ODF deployment, but OCP deployment passed correctly): https://url.corp.redhat.com/1ee0dd9
```
$ oc get infrastructure cluster -o jsonpath='{.status.platform}'
VSphere
```

PR #13272, opened 2025-09-24
Create hosted cluster with agents

jilju:
Create hosted cluster with agents

PR #13858, opened 2025-12-07
Deploy HCP agent clusters on vSphere

DanielOsypenko:
This PR is a continuation of [the Create hosted cluster with agents #13272 PR ](https://github.com/red-hat-storage/ocs-ci/pull/13272)

With this PR we added:
- [x] get_ip_list_by_cluster_id from assistant installer
- [x] dns records modification
- [x] booting machines with specs based on hosted cluster configuration 
- [x] added a class VSPHEREAgentAI
- [x] some restructuring to align deployment classes relation with the logic of the product (1 Hosted Cluster attached to 1 Host Environment)
- [x] resolve MCE mirrors issues  
- [x] resolve problem with pull containers policy on machines (applied w/a)
- [x] create kubeconfig of Agent cluster on Jenkins agent
- [x] simple post-deployment check

TODO with the next PR:
- [ ] create kubadmin-password
- [ ] verify Provider cluster creates kubeconfig with ODF deployment phase
- [ ] deploy ODF Client (ensure existing code works)
- [ ] provision ODF Client (ensure existing code works)
- [ ] release IPs on IPAM - check wether it works

Bellow cluster utilization with 3 Agent clusters (all of them have 3 wn)
image


conf/deployment/vsphere/ai_1az_rhcos_vsan_3w_agent.yaml
```diff
@@ -0,0 +1,14 @@
+DEPLOYMENT:
+  allow_lower_instance_requirements: false
+ENV_DATA:
+  platform: "vsphere_agent"
```
jilju: Please use only "hosted_cluster_platform" value as an identifier for agent/kubevirt. 
We are working on changing the platform values to represent as platform only, not the type of cluster configuration.

ocs_ci/deployment/factory.py
```diff
@@ -32,15 +32,27 @@ def __init__(self):
                     "aws_upi_flexy": AWSUPIFlexy,
                 }
             )
-        elif self.deployment_platform == constants.VSPHERE_PLATFORM:
-            from .vmware import VSPHEREUPI, VSPHEREIPI, VSPHEREUPIFlexy, VSPHEREAI
+        elif self.deployment_platform in [
+            constants.VSPHERE_PLATFORM,
+            constants.VSPHERE_AGENT_PLATFORM,
+        ]:
```
DanielOsypenko: @jilju, referring to your comment. We must use a platform to identify deployer class with deployment factory

conf/deployment/vsphere/ai_1az_rhcos_vsan_3w_agent.yaml
DanielOsypenko: cluster still must have a platform, if we do not put it here it will be `aws` by a default. 

We continue use `hosted_cluster_platform` as a distinctive key-value pair to separate `kubevirt` from `agent` deployments.

dahorak: The platform should be only `vsphere` and the information that it is `agent` deployment should be specified somewhere else.
I think it will make more sense to set the `deployment_type` to `agent` or `ai_agent`, because it actually is not deployed via Assisted Installer service, event though the process is very similar.

ocs_ci/deployment/hub_spoke.py
```diff
@@ -1584,16 +1607,68 @@ def deploy_ocp(self, **kwargs) -> str:
...
+                return self.name
+
+            log_step("Create host inventory and wait for min image creation")
+            self.create_host_inventory()
+            self.wait_for_image_created_in_infraenv()
+
+            log_step("Boot machines for Agent hosted cluster with min image")
+            if not self.boot_machines_for_agent():
+                # this cluster will not be added to the list of deployed clusters and ODF installation will be skipped
+                return ""
```
dahorak, line 1673: Not sure, if I didn't miss anything there - but wouldn't this silently skip any issue with the creating of the machines?

ocs_ci/deployment/vmware.py
```diff
@@ -2011,6 +2016,387 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+                openshift_version=str(version.get_semantic_ocp_version_from_config()),
+                base_dns_domain=config.ENV_DATA["base_domain"],
+                api_vip=self.api_vip,
+                ingress_vip=self.ingress_vip,
+                ssh_public_key=self.get_ssh_key(),
+                pull_secret=self.get_pull_secret(),
+            )
+
+            log_step("create (register) cluster in Assisted Installer console")
+            self.ai_cluster.create_cluster()
```
dahorak: Is this part relevant for the Agent deployment?

ocs_ci/deployment/hub_spoke.py
DanielOsypenko: yes it will log the issue with error and move to deploy rest of the hosted clusters. In a later step we are checking if all desired clusters were actually created and failing the job if not -> https://github.com/red-hat-storage/ocs-ci/blob/3c4ad9090e37b49dbfbd280aca55e6027bacadf9/ocs_ci/deployment/hub_spoke.py#L628

conf/deployment/vsphere/ai_1az_rhcos_vsan_3w_agent.yaml
DanielOsypenko: addressed

ocs_ci/deployment/vmware.py
DanielOsypenko: not relevant, removed

ocs_ci/deployment/factory.py
jilju: Can we use only 
```suggestion
        elif self.deployment_platform in [
            constants.VSPHERE_PLATFORM
        ]:
```
?
The keys in self.cls_map.update is platform+_+deployment_type which will be vsphere_ai_agent.

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -716,6 +716,128 @@ def create_kubevirt_ocp_cluster(
...
+        agents_obj.wait_for_resource(
+            condition="Discovered",
+            resource_count=nodepool_replicas,
+            timeout=1800,
+            sleep=30,
+        )
+        for agent_info in agents_obj.get()["items"]:
+            agents_obj.patch(
+                resource_name=agent_info["metadata"]["name"],
+                params='{"spec":{"approved":true}}',
```
jilju: There is another function in this PR which does the approval

jilju (dismissed):

dahorak:
Based on yesterday's discussion, I would propose to handle the HCP Agent cluster deployment similarly as standalone OCP deployment via Assisted Installer - with the difference, that we will provide a reference to the Hub cluster, which will be used instead of the public Assisted Installer service.
In other words, each Agent cluster will be deployed using separated Deploy OCS Cluster job, from the perspective of the agent cluster and the Hub cluster will be referenced in the configuration as `DEPLOYMENT["hub_cluster_name"]` and `DEPLOYMENT["hub_cluster_path"]` and based on this it will dynamically create the configuration context for the Hub cluster and use that context for the tasks performed on the Hub cluster (infrastructure creation, nodes approval,....).

Here I created proof of concept update for creating the Hub cluster config context: https://github.com/red-hat-storage/ocs-ci/pull/13912/changes/86292e74e6aeb223d28cfa9e934efed2d75a4162.

@jilju WDYT?

Related MR for enabling this in the main Deploy OCS Cluster job: https://url.corp.redhat.com/ade84c0

ocs_ci/deployment/vmware.py
```diff
@@ -2011,6 +2017,348 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+
+        def generate_terraform_vars(self):
+            """
+            Generates the terraform.tfvars.json file
+            """
+            self.terraform_vars_file = os.path.join(
+                self.terraform_data_dir,
+                "terraform.tfvars.json",
+            )
+            # control plane specs removed by purpose. HCP spoke clusters can not have own control plane nodes
```
dahorak, line 2268: You will have to explicitly set:
```
"control_plane_count": 0,
```
because this variable is by default set to 3 in the terraform: https://github.com/red-hat-storage/ocs-ci/blob/732d87b7530492162a1653ae3f187626fbc6a830/terraform/ai/vsphere/variables.tf#L90-L93

ocs_ci/deployment/vmware.py
```diff
@@ -2011,6 +2017,348 @@ def destroy_cluster(self, log_level="DEBUG"):
...
+            if config.ENV_DATA.get("vsphere_storage_policy"):
+                tfvars["vsphere_storage_policy"] = config.ENV_DATA[
+                    "vsphere_storage_policy"
+                ]
+            with open(self.terraform_vars_file, "w") as tf_var_file:
+                json.dump(tfvars, tf_var_file, indent=4)
+
+        def assign_api_ingress_ips(self):
+            """
+            Request API and Ingress IPs from IPAM server
```
dahorak, line 2301: Do we need additional API IP for the hosted cluster? IIUC API is provided by the control plane/master nodes, which means that it is provided by the hosting cluster.

ocs_ci/deployment/vmware.py
```diff
@@ -2663,6 +3011,64 @@ def delete_dns_records():
             aws.delete_record(record, hosted_zone_id)
 
 
+def modify_dns_records(api_ips, apps_ips, ttl=60):
+    """
+    Update existing DNS records with multiple IP addresses.
+
+    Args:
+        api_ips (list(str)): IPs for api.<cluster>.
+        apps_ips (list(str)): IPs for *.apps.<cluster>.
```
dahorak: Better to use `ingress` instead of `apps` to align it with the other code and also with the naming from OCP.

ocs_ci/ocs/constants.py
```diff
@@ -1716,6 +1710,7 @@
 AZURE_WITH_LOGS_PLATFORM = "azure-with-logs"
 GCP_PLATFORM = "gcp"
 VSPHERE_PLATFORM = "vsphere"
+VSPHERE_AGENT_PLATFORM = "vsphere_agent"
```
dahorak: I think this is not used anywhere (`agent` is part of deployment type, not platform).

ocs_ci/utility/utils.py
```diff
@@ -790,6 +792,54 @@ def download_file(url, filename, **kwargs):
...
 
 
+def download_with_retries(url, boot_image_path, max_retries=3):
+    """
+    Download file with retries and proper error handling
+
+    Args:
+        url (str): URL of the file to download
+        boot_image_path (str): Path where to save the downloaded file
+        max_retries (int): Maximum number of retries for downloading the file
```
dahorak: Since the function is quite common/universal, I would propose to change the `boot_image_path` (and other references to ISO/boot image) to some common name (like `filename` in the above `download_file` function).

I'm also thinking, if it is worth to keep both the old `download_file` and new `download_with_retries` functions, while they do the same and this new one is just enhancement of the old one.

DanielOsypenko: makes sense

ocs_ci/deployment/factory.py
DanielOsypenko: correct, I fixed this part

ocs_ci/deployment/vmware.py
DanielOsypenko: You're likely right. I will ned to test it.

ocs_ci/deployment/vmware.py
DanielOsypenko: will try it, previously I had failures around `control_plane` or some checks, not remember particularly, so I had to exclude control_plane specs at all.

ocs_ci/utility/utils.py
DanielOsypenko: I added a deprecation note into old download_file. I don't want to replace this function completely with this PR, better do it in separate PR with more testing around

ocs_ci/deployment/helpers/hypershift_base.py
DanielOsypenko: this is redundant indeed. Removed it.

ocs_ci/deployment/vmware.py
DanielOsypenko: I think we need this IP in case if we need to pass it with terraform: https://github.com/red-hat-storage/ocs-ci/pull/13858/changes#diff-0f88aacdde0eb738c226e3b802c9f361ddb8127c345327bbda199cd23c4f9fa3R2279 
If terraform can manage booting machines without it we can remove it.

dahorak: If you will not add the `api_ip` variable to the terraform vars file, it will not create the DNS record for that, otherwise it should work. Another point is, how exactly is the API handled - if we can simply create DNS record pointing to the Provider cluster API IP?

ocs_ci/deployment/helpers/hypershift_base.py
```diff
@@ -1107,3 +1231,215 @@ def create_cluster_dir(cluster_name):
...
+    Returns:
+        Config: Config object for the HUB Cluster
+
+    """
+    cluster_config = Config()
+    cluster_path = os.path.expanduser(config.DEPLOYMENT.get("hub_cluster_path"))
+    def_client_config_dict = {
+        "DEPLOYMENT": {},
+        "ENV_DATA": {
+            "cluster_name": config.DEPLOYMENT.get("hub_cluster_name"),
```
dahorak, line 1250: It would be good to mention `DEPLOYMENT.hub_cluster_path` and DEPLOYMENT.hub_cluster_name` in `conf/README.md` and maybe also worth to mention it as comment in the `conf/deployment/vsphere/ai_1az_rhcos_vsan_3w_agent.yaml` config file.

dahorak (approved):

ocs_ci/deployment/vmware.py
DanielOsypenko: currently we are relying on this logic, that terraform is creating dns record with agent cluster name as root of the url. In later step we are modifying dns records, adding 3 ips of the master node. 
We should probably use loadbalancer in future.

ocs_ci/deployment/helpers/hypershift_base.py
DanielOsypenko: done

Issue #13458, opened 2025-10-28
Script needs to be adjusted based on new returned content from boto method


udaysk23:
Script needs to be adjusted based on new returned content from boto method

```
E       AssertionError: Put object operation doesn't store ContentEncoding!!
E       assert 'zip,aws-chunked' == 'zip'
E         - zip
E         + zip,aws-chunked

```

RP link: https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/954/39174/1844370/1844580/log

PR #13865, opened 2025-12-08
Fixes content check in boto response

udaysk23:
Fixes content check in boto response

Latest boto response has `zip,aws-chunked` in content type where as previously script was checking only for `zip` using == operator

Fixes https://github.com/red-hat-storage/ocs-ci/issues/13458


ypersky1980 (approved):
LGTM

PR #13866, opened 2025-12-08
AssistedInstaller: fix destroy process for old clusters

dahorak:
If a cluster exists for quite long time (maybe more than 2 months), it is automatically removed from Assisted Installer console (or moved to some archive), but some leftovers are left there (AddHostsCluster), which breaks the existing destroy workflow.
This PR should fix this issue and if the Cluster itself is missing, check also for the alone AddHostsCluster type.


DanielOsypenko (approved):

dahorak:
Verification Destroy job: https://url.corp.redhat.com/67c2f71
We are not able to easily fully test this fix, because the issue is related to clusters running for quite long time. So the verification just verifies that this PR didn't break the normal workflow.

Issue #13872, opened 2025-12-08
[Bug]: catch destroy appliance cluster error 404


DanielOsypenko:
### Description
A clear and concise description of the issue.

```
2025-12-08 18:18:11  E               ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: rosa list service.
2025-12-08 18:18:11  E               Error is WARN: The current version (1.2.51) is not up to date with latest rosa cli released version (1.2.53).
2025-12-08 18:18:11  E               WARN: It is recommended that you update to the latest version.
2025-12-08 18:18:11  E               ERR: Failed to retrieve list of managed services: cannot retrieve services list: expected response content type 'application/json' but received 'text/html' and content '<html>
2025-12-08 18:18:11  
2025-12-08 18:18:11  E               <head><title>404 Not Found</title></head>
2025-12-08 18:18:11  
2025-12-08 18:18:11  E               <body>
2025-12-08 18:18:11  
2025-12-08 18:18:11  E               <center><h1>404 Not Found</h1></center>
2025-12-08 18:18:11  
2025-12-08 18:18:11  E               <hr><center>nginx/1.22.1</center>
2025-12-08 18:18:11  
2025-12-08 18:18:11  E               </body>
2025-12-08 18:18:11  
2025-12-08 18:18:11  E               </html>
2025-12-08 18:18:11  
2025-12-08 18:18:11  E               '
```

### Steps to Reproduce
1.
2.
3.

### Environment
- Test Suite(s):
- Platform(s):
- Version(s):
- OS:

PR #13873, opened 2025-12-08
catch destroy applience cluster 404

DanielOsypenko:
Fixes: #13872

The function destroy_appliance_mode_cluster was a leftover on managed-service destroy job. We utilized that until 4.20 and this issue did not appear. The problem is that this api call is now either deprecated or broken. So the safer solution will be to preserve this function and try to call it and later remove it or adjust, by necessity.


DanielOsypenko:
https://url.corp.redhat.com/2a5ac8a - passed

PR #13895, opened 2025-12-09
[release-4.14] Restore RGW client data prefix and refactor cloud manager + Cloud Manager init: account for the case where AWS credentials are not provided 

sagihirshfeld:
This is a manual backport of #13842

The only difference between this and the 4.15 backport is the removal of the `AZURE_WITH_LOGS` key from the `cloud_map`, since the associated feature was only introduced in 4.15.


udaysk23 (commented):
@sagihirshfeld Since this is manual cherry pick, I would request you to run validation job once to validate nothing is broken

Issue #13623, opened 2025-11-13
[Bug]:  test_add_capacity: check_pods_in_statuses  function getting executed before pods stabilize


s-selwin:
### Description

When checked , a default timeout period can be included in this function as it is getting executed before the pods stabilize. The MG logs are clean with right pod statuses( Running/Completed) which explains the wait time needed.

Report Portal: 

Issue 1: https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/852/39780/1866556/1866916/log
Issue 2: https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/852/39501/1854802/1855179/log

### Steps to Reproduce
1.
2.
3.

### Environment
- Test Suite(s):
- Platform(s):
- Version(s):
- OS:

s-selwin:
Getting the error again in RP- https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/954/40299/1883113/1883494/log

s-selwin:
fix #13899

s-selwin:
skipping this testcase until https://issues.redhat.com/browse/DFBUGS-5191 is fixed. There is a PR present --> https://github.com/red-hat-storage/ocs-ci/pull/13899 but can't proceed in checking , as the ceph crash issue happens in the middle.

s-selwin:
sharing slack conversation ---> https://ibm-systems-storage.slack.com/archives/C06DWD30QTZ/p1767861247643149

PR #13899, opened 2025-12-10
Increased timeout for pod stabilize & ignoring storage client pod

s-selwin:
This PR resolves two key instability issues that were causing transient failures (false negatives) in the **test_add_capacity**

**Entry Criteria Race Condition:** Fixes the Error from server (NotFound) failures occurring in check_pods_in_statuses. This was due to a race condition where the check tried to query short-lived storageclient pods that had already been deleted by Kubernetes Garbage Collection. We now explicitly exclude these transient pods from the health check.

**Entry Check Timeout:** We were incorrectly failing the entry criteria because the test was trying to track deleted, temporary Job Pods (like storageclient-...). These are now excluded to eliminate the NotFound API error and the subsequent timeout.

### DIFF
--- a/tests/functional/z_cluster/cluster_expansion/test_add_capacity_entry_exit_criteria.py
+++ b/tests/functional/z_cluster/cluster_expansion/test_add_capacity_entry_exit_criteria.py
@@ -104,6 +104,7 @@ class TestAddCapacity(ManageTest):
             expected_statuses=expected_statuses,
             exclude_pod_name_prefixes=["demo-pod", "storageclient"],
             timeout=500,
+            sleep=20,
         ), "Entry criteria FAILED: one or more OCS pods are not in running state"
         # Create the namespace under which this test will execute:
         project = project_factory()


amr1ta:
Please run a pr validation job.

tests/functional/z_cluster/cluster_expansion/test_add_capacity_entry_exit_criteria.py
```diff
@@ -100,9 +100,10 @@ def test_add_capacity(
         # All OCS pods are in running state:
         # ToDo https://github.com/red-hat-storage/ocs-ci/issues/2361
         expected_statuses = [constants.STATUS_RUNNING, constants.STATUS_COMPLETED]
-        assert pod_helpers.check_pods_in_statuses(
+        assert pod_helpers.wait_for_pods_to_be_in_statuses(
             expected_statuses=expected_statuses,
-            exclude_pod_name_prefixes=["demo-pod"],
+            exclude_pod_name_prefixes=["demo-pod", "storageclient"],
+            timeout=500,
```
AviadP, line 106: 500 sec sounds like long timeout if problem is race condition, is it possible to use shorter timeout? and what would be sampling interval?

s-selwin: @AviadP You are absolutely right that a 500-second timeout is excessive for fixing a simple race condition. but in here,  the timeout is for a much broader purpose to make sure all pods stabilize before starting the test. 

As we know, the wait_for_pods_to_be_in_statuses validates the readiness of the pods before introducing  add_capacity operation.

I found a bug (https://github.com/red-hat-storage/ocs-ci/issues/13623) where in test failed with error "one or more ocs pods are not in running but when troubleshooting through MG logs (all -o wide) , I spotted all pods in right statuses but took time in stabilizing - just by the duration they were running. PFB

Screenshot 2025-12-11 at 11 06 16 AM


A shorter timeout (e.g., 180s) would lead to a false failure every time initialization takes longer than 3 minutes.

The 500 seconds provides the necessary safety buffer (over 8 minutes) required for the cluster to settle, preventing the test from starting in an unstable pod states.

There is no change in sampling interval (10 sec by default)

s-selwin:
PFA the PR verification logs
[test-add_capacity-execution.docx](https://github.com/user-attachments/files/24814771/test-add_capacity-execution.docx)

tests/functional/z_cluster/cluster_expansion/test_add_capacity_entry_exit_criteria.py
yitzhak12: Ack. Just one general comment: If you use a long timeout like `500`, it's better to increase the interval to `20`, so it will be more consistent, and we will have fewer dirty lines in the log.

s-selwin: done

s-selwin:
PFA the PR verification logs

[PR verification logs - test_add_capacity.docx](https://github.com/user-attachments/files/24904680/PR.verification.logs.-.test_add_capacity.docx)

PR #13919, opened 2025-12-12
New way to install unreleased operators

petr-balogh:
Based on this doc:
https://docs.google.com/document/d/1YpMZrKUHQG1QAVstVY4Z7RvM3jyZec6sMSCYoNOAaA8/edit?tab=t.0#heading=h.99jz7u40644


petr-balogh:
Initial testing triggered here:
https://url.corp.redhat.com/69b7f41

As it's first run I expect this will have some issues

petr-balogh:
Verified here:
 https://url.corp.redhat.com/12cfe56

ocs_ci/utility/deployment.py
```diff
@@ -27,6 +27,7 @@
 logger = logging.getLogger(__name__)
 
 
+# TODO: remove this function and use the one in version.py
 def get_ocp_ga_version(channel):
```
clacroix12, line 31: Should we decorate this (and other similar functions) as deprecated until it is removed?

```suggestion
from warnings import deprecated

@deprecated("Use function from version.py instead")
def get_ocp_ga_version(channel):
```

ocs_ci/utility/operators.py
```diff
@@ -0,0 +1,310 @@
...
+        if not self.unreleased_catalog_name:
+            raise ValueError(
+                "Child class must define attribute `unreleased_catalog_name`"
+            )
+        self.catalog_name = self.unreleased_catalog_name
+
+    def get_idms_data(self):
+        if not self.unreleased_images:
+            raise ValueError(
+                "Child class must define attribute `unreleased_images_source_mirrors`"
```
clacroix12: I suspect this attribute name changed after the error was written.

```suggestion
                "Child class must define attribute `unreleased_images`"
```

ocs_ci/utility/operators.py
```diff
@@ -0,0 +1,310 @@
...
+        return f"{self.name}-unreleased-idms"
+
+    @property
+    def unreleased_catalog_name(self):
+        return f"{self.name}-unreleased-catalog"
+
+    def set_unreleased_catalog(self):
+        if not self.unreleased_catalog_name:
+            raise ValueError(
+                "Child class must define attribute `unreleased_catalog_name`"
```
clacroix12: Is this true? We already have the property defined right above in the parent class

petr-balogh:
New verification job:
https://url.corp.redhat.com/19fb66b

petr-balogh:
Multus job verification:
https://url.corp.redhat.com/4965f46

petr-balogh:
```
pbalogh nmstate $ oc get csv -n openshift-nmstate
NAME                                              DISPLAY                       VERSION               REPLACES   PHASE
kubernetes-nmstate-operator.4.21.0-202512120820   Kubernetes NMState Operator   4.21.0-202512120820              Succeeded
pbalogh nmstate $ oc get csv -n openshift-local-storage
NAME                                          DISPLAY         VERSION               REPLACES   PHASE
local-storage-operator.v4.21.0-202512101620   Local Storage   4.21.0-202512101620              Succeeded
```

Marking as verified

ocs_ci/utility/operators.py
```diff
@@ -0,0 +1,548 @@
...
+            self.unreleased_catalog_full_image,
+            mirrored_index_image,
+            self.disconnected_required_packages,
+            idms=idms,
+        )
+        self._create_catalog(mirrored_index_image, self.unreleased_catalog_name)
+
+    def is_available(self):
+        """
+        Get the version of the operator
```
ebenahar: It should be something like:
```
Check if the operator is available
```

ocs_ci/utility/operators.py
```diff
@@ -0,0 +1,548 @@
...
+        """
+        Get the channel to use for installing the operator
+
+        Returns:
+            str: operator channel
+
+        """
+        ocp_version = get_ocp_version()
+        selector = self.catalog_selector
+        # Retrieve available channels for LSO
```
ebenahar: This class is the parent Operator class. LSO here seems a mistake

ocs_ci/utility/operators.py
```diff
@@ -0,0 +1,548 @@
...
+
+        # Ensure versions are sorted
+        versions.sort()
+        sorted_versions = [v.vstring for v in versions]
+
+        if len(sorted_versions) >= 1:
+            # Use latest channel
+            if stable_channel_found:
+                if stable_channel_version > get_semantic_version(sorted_versions[-1]):
+                    return "stable"
```
ebenahar: shouldn't we save `stable` as a constant?

ocs_ci/utility/operators.py
petr-balogh: Yes, I am reusing this logic from one place where it was defined only for LSO. So this is mostly moved code and forgot to remove it. Deleting now.

ocs_ci/utility/operators.py
petr-balogh: I don't think it's necessary to have everything in constants, but created and using from there.

ocs_ci/utility/deployment.py
petr-balogh: I am going to remove this function at least from here in follow up PR as moved same code to versions module. 
BTW `@deprecated` decorator seems to be only available from python 3.13. And we support max python 3.11 as of now

vavuthu (approved):

ocs_ci/utility/deployment.py
clacroix12: Ah, I thought I tested that with my ocs-ci venv but perhaps I didn't have it activated. For older python versions we would have to create a custom `deprecated` decorator ourselves which would log the warning.

PR #13951, opened 2025-12-16
Adds logic for installing the LSO operator if its not there

rh-aaleem:


ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -56,6 +58,17 @@ def deploy(self):
         self.verify_fdf_installation()
         self.setup_storage()
 
+    def install_lso(self):
+        """
+        Install LSO operator if its not being installed
```
petr-balogh:
```suggestion
        In the case of LSO is not available - bring catalog for unreleased version and install it.
```

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -56,6 +58,17 @@ def deploy(self):
...
+        """
+        from ocs_ci.utility.operators import LocalStorageOperator
```
petr-balogh: This import can be in the start of the modul please

PR #13959, opened 2025-12-17
Add ui market to black squad test cases

anajain16:
Added ui marker to black squad marked ui test cases


anajain16:
will create new pr

PR #13977, opened 2025-12-19
Ibmcloud ROKS Node reboot with hard option during cluster shutdown and restart

sikhlaqu:
**Fixes:**

1. During ODF-DR CI execution, the primary cluster All Nodes are shutdown and then restarted Later. 
The IBM ROKs node reboot often hangs and doesnt bring the Node to 'ready' state. 
Now "--hard" option has been added during Node reboot, which worked fine on the 'hung' cluster during the Test.

cc: @petr-balogh


petr-balogh:
Verification job:
https://url.corp.redhat.com/7efc3f8

PR #13978, opened 2025-12-19
W/A for multi platform deployment issue

petr-balogh:
We have serious issue in design and how the deployment is implemented for multicluster scenarios as only single deployer class object is initiated and then we cannot rely on it when we iterating over multiple clusters in underlaying deployment methods for do_deploy_rdr and so on.

Issue:
https://issues.redhat.com/browse/OCSQE-4214

I am tryin to W/A this on curcial parts to rely on config values which should be specific for concrete cluster.

But to fix this properly will require bigger refactoring and broader discussion with team.

### DIFF
--- a/conf/README.md
+++ b/conf/README.md
@@ -185,6 +185,7 @@ version.
 * `enable_nested_virtualization` - Enable nested virtualization for vSphere platform primarily. Used for kubevirt on HCP Clusters. It sets options kvm_intel nested=1 options kvm_amd nested=1 in MachineConfig
 * `host_network` - Enable host network in the storage cluster CR and to be able to connect to the storage cluster from the host network or other scenarios where host network is required.
 * `partitioned_disk_on_workers` - Create a partition for OSD on the OS disk on worker nodes.
+* `submariner_cli_deployment` - Enforce Submariner CLI deployment.
 
 #### REPORTING
 
--- a/conf/ocsci/submariner_cli_deployment.yaml
+++ b/conf/ocsci/submariner_cli_deployment.yaml
@@ -1,0 +1,4 @@
+# Use this file to enforce CLI Submariner deployment
+---
+DEPLOYMENT:
+  submariner_cli_deployment: true
--- a/ocs_ci/deployment/acm.py
+++ b/ocs_ci/deployment/acm.py
@@ -142,50 +142,54 @@ class Submariner(object):
             config.switch_ctx(old_ctx)
 
         global_net = get_primary_cluster_config().ENV_DATA.get("enable_globalnet", True)
+        # W/A for ROKS deployment
+        roks_deployment = (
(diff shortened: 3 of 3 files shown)


ocs_ci/deployment/deployment.py
```diff
@@ -207,13 +207,46 @@ class Deployment(object):
...
 
     def __init__(self):
-        self.platform = config.ENV_DATA["platform"]
-        self.ocp_deployment_type = config.ENV_DATA["deployment_type"]
-        self.cluster_path = config.ENV_DATA["cluster_path"]
-        self.namespace = config.ENV_DATA["cluster_namespace"]
         self.sts_role_arn = None
-        self.storage_class = storage_class.get_storageclass()
-        self.custom_storage_class_path = None
+        storage_class.set_custom_storage_class_path()
```
clacroix12, line 211: Does this need to be called from the deployment class or should this be handled fully within the storage_class module?

ocs_ci/deployment/deployment.py
```diff
@@ -4049,3 +4048,38 @@ def deploy_dr_policy(self):
     "regional-dr": RDRMultiClusterDROperatorsDeploy,
     "metro-dr": MDRMultiClusterDROperatorsDeploy,
 }
+
+
+def get_arbiter_location():
```
clacroix12, line 4053: In my WIP branch for 2AZ+Arbiter I've moved this to it's own utility module `utility/arbiter.py` to try to avoid any circular imports by leaving it in the deployment module. Not sure if you've hit these but if you are making any other changes to the branch perhaps this could be moved as well.

ocs_ci/deployment/deployment.py
petr-balogh: I am calling it here to make sure when creating deployer object for each cluster config in cluster fixture - that we have properly set right away right after init the storage_class in every config.

ocs_ci/deployment/deployment.py
petr-balogh: as I have passed deployment and it's kinda verified that not caused the regression and hopefully will fix also multi platform RDR deployment, then I prefer to not hold this PR for this change as I already started my holiday and it can be merged and people in india start using it as their feature testing will not be blocked if it's working.  

Probably move can be done as part of your 2az arbiter PR later on. WDYT @clacroix12 ?

petr-balogh:
Trying verification here:
 https://url.corp.redhat.com/82eae13

petr-balogh:
Multi platform RDR BM/vSphere deployment passed here:
 https://url.corp.redhat.com/0f26026
 
 So now it's fully verified ready to merge

sidhant-agrawal (approved):

shivamdurgbuns (approved):

PR #13987, opened 2025-12-22
`test_bucket_replication_with_versioning.py`: Verify health after applying a replication policy to the OBCs

sagihirshfeld:
Sometimes (in under ~25% of runs), the first operation on a bucket immediately after applying a replication policy fails with a NoSuchBucket error:

```
E ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc --kubeconfig ...
...

An error occurred (NoSuchBucket) when calling the PutObject operation: The specified bucket does not exist.
```

This PR stabilizes the tests by reusing the existing ObjectBucket::verify_health check after applying the replication policy, and waiting until the bucket is reported as healthy before proceeding.

To keep the scope contained, the change is currently limited to test_bucket_replication_with_versioning.py. Additional tests will be updated once we have sufficient signal that this improves overall stability.

### DIFF
--- a/ocs_ci/ocs/bucket_utils.py
+++ b/ocs_ci/ocs/bucket_utils.py
@@ -2212,6 +2212,25 @@ def update_replication_policy(bucket_name, replication_policy_dict):
 
 
 @config.run_with_provider_context_if_available
+def update_replication_policy_and_verify_health(bucket_obj, replication_policy_dict):
+    """
+    Updates the replication policy of a bucket and verifies the health of the bucket.
+    This is useful because the bucket may not be immediately available after the
+    replication policy is updated.
+
+    Args:
+        bucket_obj (ObjectBucket): The bucket object to update the replication policy of
+        replication_policy_dict (dict): A dictionary containing the new replication
+        policy
+
+    Raises:
+        TimeoutExpiredError: If the bucket does not reach a healthy state within 120 seconds
+    """
+    update_replication_policy(bucket_obj.name, replication_policy_dict)
+    bucket_obj.verify_health(timeout=120)
+
+
+@config.run_with_provider_context_if_available
 def get_replication_policy(bucket_name):
     """
     Get the replication policy on a bucket
--- a/ocs_ci/ocs/bucket_utils.py
+++ b/ocs_ci/ocs/bucket_utils.py
@@ -2184,7 +2184,7 @@ def patch_replication_policy_to_bucket(
 
 
 @config.run_with_provider_context_if_available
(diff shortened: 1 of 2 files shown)


ypersky1980 (dismissed):
LGTM

ocs_ci/ocs/bucket_utils.py
```diff
@@ -2211,6 +2211,25 @@ def update_replication_policy(bucket_name, replication_policy_dict):
...
+    Args:
+        bucket_obj (ObjectBucket): The bucket object to update the replication policy of
+        replication_policy_dict (dict): A dictionary containing the new replication
+        policy
+
+    Raises:
+        TimeoutExpiredError: If the bucket does not reach a healthy state within 120 seconds
+    """
+    update_replication_policy(bucket_obj.name, replication_policy_dict)
+    bucket_obj.verify_health(timeout=120)
```
udaysk23: We can add this line in `update_replication_policy` method with extra keyword parameter like `verify_health=False`

sagihirshfeld: Looking at this PR again, I see what you mean. 

The problem was that `update_replication_policy` is only accepting the bucket name and not the ObjectBucket instance that we need to access the right `verify_health` method, and that we can't easily create such an instance given only the name. Also that this function is already widely used in OCS-CI.

However, I think I found a solution. In https://github.com/red-hat-storage/ocs-ci/pull/13987/commits/ec5fc650f2b73a47c33b1343b289088ff5c15416 I've added `**kwargs` to the original function's arguments, and if one wants to run a health verification after the patch they would have to pass the ObjectBucket instance as well.

For example:
```
        update_replication_policy(
            bucket_a.name,
            replication_policy.to_dict(),
            verify_health=True,
            bucket_obj=bucket_a,
        )
```

ypersky1980 (approved):
LGTM

ocs_ci/ocs/bucket_utils.py
```diff
@@ -2210,6 +2218,17 @@ def update_replication_policy(bucket_name, replication_policy_dict):
         resource_name=bucket_name,
     ).patch(params=json.dumps(replication_policy_patch_dict), format_type="merge")
 
+    # Optional health verification
+    verify_health = kwargs.get("verify_health", False)
+    bucket_obj = kwargs.get("bucket_obj", None)
```
udaysk23, line 2223:
```suggestion
    bucket_obj = kwargs.get("bucket_obj")
```

PR #14001, opened 2025-12-23
Changed the fixture scope from "session" to "class" in the TestCephtoolboxPod class.

suchita-g:
Test test after test_nodeaffinity_to_ceph_toolbox_with_default_taints are failing to schedule pods due to the presence of taint in the worker nodes. The test case adds the node taints but the finalizer removes the taint which has session scope.
Changed the fixture scope from scope="session" to scope="class", The finalizer now runs after all tests in the TestCephtoolboxPod class complete


suchita-g:
Verified with http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/sgatfane-22docp/sgatfane-22docp_20251222T071237/logs/test_report_1766517983.html

DanielOsypenko (approved):

PR #14003, opened 2025-12-24
Metallb operator 421

DanielOsypenko:
add MetalLBOperator into MetalLBInstaller similarly to LocalStorageOperator.
Unnecessary functions will be removed later.

verification run: https://url.corp.redhat.com/f9ad672

### DIFF
--- a/ocs_ci/utility/operators.py
+++ b/ocs_ci/utility/operators.py
@@ -575,6 +575,8 @@ class MetalLBOperator(Operator):
         operatorgroup_data["metadata"]["annotations"] = {
             "olm.providedAPIs": "MetalLB.v1beta1.metallb.io",
         }
+        # metallb does not support InstallMode OwnNamespace
+        operatorgroup_data["spec"]["targetNamespaces"] = []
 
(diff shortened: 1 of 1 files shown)


fbalak (approved):

ocs_ci/utility/operators.py
```diff
@@ -545,4 +545,46 @@ def _deployment_verification(self):
...
+
+    def _customize_operatorgroup(self, operatorgroup_data: dict):
+        """
+        Hook for MetalLB to customize OperatorGroup YAML
+
+        Args:
+            operatorgroup_data (dict): the OperatorGroup YAML data
+        """
+        operatorgroup_data["metadata"]["annotations"] = {
+            "olm.providedAPIs": "MetalLB.v1beta1.metallb.io",
```
petr-balogh, line 576: Is this something new required for 4.21 ? As I don't see we used to have this before?
As I see beta mentioned there, not sure if this is something which will be also used for GAed one operator and if not, maybe we should somehow conditionalize this, but I dunno?

Issue #13284, opened 2025-09-26
[Bug]: test_get_mon_endpoint failed with AssertionError: Invalid mon endpoint format


suchita-g:
### Description
A clear and concise description of the issue.
The tier1 test test_get_mon_endpoint failed with AssertionError: Invalid mon endpoint format 
### Steps to Reproduce
1. [ODF 4.18](https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/795)
[OCS4-18-Downstream-OCP4-18-VSPHERE6-UPI-1AZ-RHCOS-EXTERNAL-3M-3W-tier1-jenkins #13](https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/795/37655)
[OCS4-18-Downstream-OCP4-18-VSPHERE6-UPI-1AZ-RHCOS-EXTERNAL-3M-3W-tier1-jenkins](https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/795/37655/1777244)
test_get_mon_endpoint
2. https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/795/37655/1777244/1777370/log
2.
3.

### Actual Behavior
test_get_mon_endpoint test failed 
### Expected Behavior
test_get_mon_endpoint should pass

### Environment
- Test Suite(s): tier1 
- Platform(s): external mode
- Version(s): 4.18
- OS:

### Additional Context
```
self = <test_get_commands.TestGetCommands object at 0x7f5002b626a0>


@polarion_id("OCS-6238")
def test_get_mon_endpoint(self):
    result = self.odf_cli_runner.run_get_mon_endpoint()
    output = result.stdout.decode().strip()
    assert output, "Mon endpoint not found in output"
    # Validate the format of the mon endpoint output
    endpoint_pattern = r"^(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}:\d+,?)+$"

  assert re.match(
        endpoint_pattern, output
    ), f"Invalid mon endpoint format: {output}"
E       AssertionError: Invalid mon endpoint format: 576133165110.0.66.64:3300,576133165210.0.66.66:6789,576133165310.0.65.164:6789,576133165610.0.66.93:6789,576133165510.0.65.239:6789
E       assert None
[...]

PR #14015, opened 2025-12-31
Updating the ip-address regex pattern to support external mode

suchita-g:
Fixes issue #13284
Existing test is adjusted to handles the external mode IP addresses vs all other ODF platforms. External mode format is like 576133165110.0.66.64:3300 where the first octet can be longer. Now code uses the appropriate regex pattern based on the mode


suchita-g:
skiping this test on external mode. Tried fix but test itself not valid further .closing PR.

PR #14018, opened 2026-01-01
Add truncate_long_lines() to reduce log volume from OLM metadata

AviadP:
- [x] Add truncate_long_lines() function to truncate lines exceeding 500 characters
- [x] Integrate line truncation in exec_cmd() stdout logging to reduce log volume
- [x] Preserve first 250 + last 50 chars with [...N chars truncated...] marker for debugging

this PR should trim another ~2% of unwanted prints

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -922,11 +922,12 @@ def truncate_long_lines(output: str, max_line_length: int = 500) -> str:
     - Adds truncation marker in middle
 
     Args:
-        output: Command output to process
-        max_line_length: Maximum line length before truncation
+        output (str): Command output to process.
+        max_line_length (int): Maximum line length before truncation.
 
     Returns:
-        str: Output with long lines truncated
+        str: Output with long lines truncated.
+
     """
     if not output:
         return output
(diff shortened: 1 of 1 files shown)


ocs_ci/utility/utils.py
```diff
@@ -771,6 +772,46 @@ def repl(matchobj: Match[str]) -> str:
...
+    - First N/2 chars (includes log prefix and key info)
+    - Last 50 chars (for context)
+    - Adds truncation marker in middle
+
+    Args:
+        output: Command output to process
+        max_line_length: Maximum line length before truncation
+
+    Returns:
+        str: Output with long lines truncated
```
dahorak: Similarly to the other PR, there should be one empty line (each list should end with empty line in markdown).

dahorak (dismissed):

ocs_ci/utility/utils.py
```diff
@@ -771,6 +772,46 @@ def repl(matchobj: Match[str]) -> str:
...
+    for line in lines:
+        if len(line) > max_line_length:
+            # Keep prefix and suffix, truncate middle
+            prefix_len = max_line_length // 2
+            suffix_len = 50
+            truncated_chars = max(0, len(line) - prefix_len - suffix_len)
+            if truncated_chars > 0:
+                result.append(
+                    f"{line[:prefix_len]}[...{truncated_chars} chars truncated...]{line[-suffix_len:]}"
+                )
```
clacroix12: I think instead of `0` here we want to use the length of `[...{truncated_chars} chars truncated...]` since there is an edge case where the amount of characters we truncate is greater than 0 but smaller than the length of the string we are replacing those characters with.


```suggestion
            truncated_text = f"[...{truncated_chars} chars truncated...]"
            if truncated_chars > len(truncated_text):
                result.append(
                    f"{line[:prefix_len]}{truncated_text}{line[-suffix_len:]}"
                )
```

AviadP: fixed

ocs_ci/utility/utils.py
AviadP: fixed

ebenahar (approved):

PR #14040, opened 2026-01-06
Only ensure LSO is installed if we are configured for local storage

clacroix12:


vavuthu (approved):

rh-aaleem (approved):

ocs_ci/deployment/fusion_data_foundation.py
petr-balogh: I think we will need to conditionalize the call of create_lvs_resource also only `if self.lso_enabled`.
My root analysis on issue seen by Avdhoot:
https://ibm-systems-storage.slack.com/archives/C06GGES4810/p1769515645359539

clacroix12:
Verified: https://url.corp.redhat.com/8733f91

PR #14053, opened 2026-01-07
AMQ::is_amq_pod_running - increase the time to wait for the pods to get ready

sagihirshfeld:
In a previous PR we've increased the timeout only to FIPs clusters, but since then I've seen more cases where AMQ's pods fail to setup in the given 300 seconds yet they appear in the must-gather logs.

### DIFF
--- a/ocs_ci/ocs/amq.py
+++ b/ocs_ci/ocs/amq.py
@@ -185,14 +185,9 @@ class AMQ(object):
         """
 
         _rc = True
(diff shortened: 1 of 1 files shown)


sagihirshfeld:
Adding the Verified label since the change is just an increase of the timeout.

vavuthu (approved):

ypersky1980 (approved):
LGTM

PR #14074, opened 2026-01-09
RHSTOR-7671:TC to verify ODF upgrade prevent status when Ceph Health WARN

suchita-g:
Test case to Verify ODF upgrade is blocked when Ceph cluster health is in WARN state. 
This Test Verifies:
1. OCS OperatorCondition CR: Upgradeable=False
2. ODF OperatorCondition CR: Upgradeable=False
3. Alert ODFOperatorNotUpgradeable


ocs_ci/helpers/upgrade_precheck_helpers.py
```diff
@@ -0,0 +1,140 @@
+"""
+Helper functions for ODF upgrade pre-check conditions.
```
fbalak: I don't think that this should be a helpers module. It would be better to add those methods to ocs_ci/ocs/resources/csv.py

ocs_ci/helpers/upgrade_precheck_helpers.py
```diff
@@ -0,0 +1,140 @@
...
+def get_operator_csv_names(namespace=None):
+    """
+    Get CSV names for OCS and ODF operators.
+
+    Args:
+        namespace (str): Namespace where CSVs are located.
+            Defaults to openshift-storage namespace.
+
+    Returns:
+        tuple: (ocs_csv_name, odf_csv_name)
```
dahorak: There should be one empty space after each list, which means also after the `Returns:...` section. (It is the markdown specification.)

Similarly on the other places in this PR.

tests/functional/upgrade/test_upgrade_precheck_ceph_health.py
```diff
@@ -0,0 +1,308 @@
...
+from ocs_ci.helpers.upgrade_precheck_helpers import (
+    get_operator_csv_names,
+    check_operatorcondition_upgradeable_false,
+)
+
+log = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope="function")
+def mon_pod_down(request):
```
fbalak: This can be in its tests/functional/upgrade/conftest.py file

dahorak (dismissed):
Overall LGTM, just one small suggestion related to docstring formatting.

tests/functional/upgrade/test_upgrade_precheck_ceph_health.py
```diff
@@ -0,0 +1,308 @@
...
+        # Verify health status contains WARN
+        assert (
+            "WARN" in health_status or "HEALTH_WARN" in health_status
+        ), f"Expected HEALTH_WARN but got: {health_status}"
+
+        # Check OperatorCondition CRs for OCS and ODF operators
+        log.info("Checking OperatorCondition CRs for OCS and ODF operators")
+
+        # Get operator namespace (CSVs are in openshift-storage namespace)
+        operator_namespace = constants.OPENSHIFT_STORAGE_NAMESPACE
```
fbalak: Installed cluster could have a different namespace from the default one. It should be taken from config.

PR #14081, opened 2026-01-09
Fix typo: DR_OPS_NAMESAPCE -> DR_OPS_NAMESPACE

RHdshelar:
Added correct constant DR_OPS_NAMESPACE and kept the old typo as a deprecated alias for backward compatibility.


prsurve:
@RHdshelar insted of creating new variable I would suggest fix typo in existing variable

PR #14083, opened 2026-01-09
Fix condition for drs deployment on client

fbalak:


fbalak:
Verified: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-odf-provider-client-multicluster/1013/

PR #14097, opened 2026-01-13
RHSTOR-7671:TC ODF upgrade prevent status when OCP and ODF Version mismatch

suchita-g:
Test case which verifies: ODF upgrade is blocked when ODF is at N+1 with respect to OCP Version. 
This Test Verifies:
1. ODF OperatorCondition CR: Upgradeable=False
2. Alert ODFOperatorNotUpgradeable

### DIFF
--- a/ocs_ci/ocs/resources/csv.py
+++ b/ocs_ci/ocs/resources/csv.py
@@ -224,3 +224,98 @@ def check_operatorcondition_upgradeable_false(
             f"or not found: {e}."
         )
         return False
+
+
+def check_operatorcondition_upgradeable_false_version_mismatch(
+    operator_name, csv_name, namespace, timeout=300
+):
+    """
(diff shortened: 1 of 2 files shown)


dahorak (approved):
Overall LGTM, I'm just not sure, how this test will be executed as part of regular test runs, because IIUC our production job are executed against the same OCP & ODF version (e.g. OCP 4.20, ODF 4.20) - so this test will be always skipped.

suchita-g:
As discussed with Petr, this automation is only useful for this Release as a part of [RHSTOR-7671](https://issues.redhat.com//browse/RHSTOR-7671) coverage. In addition to it, No Production Run has combination of OCP and ODF mismatch. So this Test will get skipped all the Time. So Whether it should be part of Regression run for a future point of view or not is conflicting.

clacroix12 (approved):

shylesh (approved):
lgtm

PR #14105, opened 2026-01-14
extend timeout for wait-for install-complete command

dahorak:
The `openshift-install wait-for install-complete ...` command have internal timeout 40 minutes, so it doesn't make much sense to have shorter timeout for the command execution. Changing the command execution timeout for this command to 3600s so it will be the same for all the occurrences of this command.


petr-balogh:
Do we want cherry-picks @dahorak ?

PR #14113, opened 2026-01-14
Add uv project requirements

clacroix12:
This PR serves as an introduction to UV as our new project and package manager. Initially I would like to add uv compatibility alongside our current dependency management to allow us to migrate smoothly.

This contains the following:
- `pyproject.toml` - Replacement for setup.py. Defines project information, dependencies, scripts, etc.
- `uv.lock` - Managed entirely by uv, lockfile which ensures that environments that run our code through `uv run` will be utilizing the correct dependency versions
- `.python-version` - Generated by uv as a part of the init process, specifies the version we wish uv to execute with

Initial performance comparison:
I wanted to display the difference between pip installing our requirements and installing them through uv. This was done by performing the following experiment:

- Created a lock of our dependency tree:
  - `uv pip compile pyproject.toml -o requirements.lock`
- Created new venvs for both pip and uv:
  - `python -m venv .venv-pip`
  - `python -m venv .venv-uv`
- Using their respective venvs, timed the installation of the requirements: 
  - `time pip install -r requirements.lock`
  - `time uv pip install -r requirements.lock`

One example of the results:
`pip install -r requirements.lock  41.09s user 2.82s system 86% cpu 50.728 total`
`uv pip install -r requirements.lock  0.12s user 1.09s system 333% cpu 0.362 total`

I performed this experiment several times and **on average uv was over 300 times faster than pip**.


clacroix12:
Fixed formatting issue black is complaining about in the PR checks here: #14114. I'm not sure why it's blocking this PR as I haven't touched those files but I fixed them nonetheless.

pyproject.toml
```diff
@@ -0,0 +1,141 @@
+[project]
+name = "ocs-ci"
+version = "4.21.0"
+description = "ODF QE test framework for qualifying Data Foundation"
+authors = [{name = "OCS QE", email = "ocs-ci@redhat.com"}]
+license="MIT"
+readme = "README.md"
+requires-python = ">=3.10,<3.12"
+dependencies = [
```
dahorak, line 9: In the original `install_requires` section in `setup.py ; we have few comments  for some of the dependencies, for example:
```
        # greenlet 1.0.0 is broken on ppc64le
        # https://github.com/python-greenlet/greenlet/issues/230
        # by default program attempting to load an x86_64-only library from a native arm64 process
        # Beginning with gevent 20.12.0, 64-bit ARM binaries are distributed on PyPI for aarch64 manylinux2014
        # compatible systems. Resolves problem for m1 Mac chips
        "greenlet==3.0rc3",
```
or
```
        # new version of marshmallow 4.0.0 seems to be broken, failing with error:
        # TypeError: __init__() got an unexpected keyword argument 'default'
        "marshmallow==3.26.2",
```
which might be worth to preserve for future updates (to be more careful when updating those modules).

dahorak (approved):
Overall LGTM, I have one suggestion related to the list of dependencies and also it will be good to add some documentation related to those changes (I know that the full switch will require bigger update of the existing documentation, but for this update it will be good to have some new readme/summary with explanation how it work and what are the basic steps to use it.)

pyproject.toml
clacroix12: Comments have been added

petr-balogh:
@clacroix12 asked people for approval, so we can get this merged and start using uv.

Just thinking, in follow up PR, can we update also https://github.com/red-hat-storage/ocs-ci/blob/master/docs/getting_started.md to document people should use uv as default option?

b-ranto:
This broke our CI job as we were running `python setup.py develop` in it to get `run-ci`. What is the new workflow/replacement for that? Are we expecting to get this merged all the wait to release-4.14 branch? (the last supported release)

petr-balogh:
@b-ranto 

Create venv with python version we support: python -m venv .venv 

Activate venv:
source .venv/bin/activate

pip install -r requirements.txt

This is old document flow which still works for us.

We will probably nit go with cherry picking to all supported versions yet. But it's still under consideration.

b-ranto:
OK, I was able to get the CI to work just by removing the `python setup.py develop` line from our job, it looks like that file was not used for a while now.

PR #14116, opened 2026-01-15
calculate number of nmstate pods dynamically

dahorak:
- the number of nmstate pods depends on number of nodes in cluster and have to be calculated dynamically to support compact mode clusters (end other sizes of clusters than 3m 3w).


dahorak:
Verification job: https://url.corp.redhat.com/dd3dbb6

suchita-g (approved):

PR #14117, opened 2026-01-15
add possibility to pack mg log files to one tarball

dahorak:
The aim of this PR is to allow packing all individual MG (and related) files to one tarball. There are two reasons for this approach:
* save space (the tarballed MG files are 10 times smaller than the original files)
* allow easy download for local examination

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -152,7 +152,7 @@ REPORTING:
   collect_logs_on_success_run: False
   rp_client_log_level: "ERROR"
   max_mg_fail_attempts: 3
-  tarball_mg_logs: false
+  tarball_mg_logs: true
   delete_packed_mg_logs: true
 
 # This is the default information about environment.


ocs_ci/ocs/utils.py
```diff
@@ -2022,6 +2034,16 @@ def collect_pod_container_rpm_package(dir_name):
                     with open(go_log_file_name, "w") as f:
                         f.write(go_output)
 
+    if config.REPORTING.get("tarball_mg_logs"):
```
petr-balogh, line 2039: Can you please check if we can also somehow add the MG tarball file path relatively from log dir into the junit xml file?  We might use it later on in reporting probably which might be usefull.

Thanks

dahorak:
Last two verification jobs were running tier1 test suite with similar configuration, the difference were in the `tarball_mg_logs: True/False` configuration.
The overall results are not 100% the same, so the comparison is not exactly precise, but the difference is still very significant:
* Logs from the job where the MG logs were packed consumes *3.2G* (from which the MG logs occupies only *~530MB*).
* Logs from the job where the MG logs were not packed consumes *30G* (from which the MG logs occupies *~27GB*).

Another example is one of the production executions (Azure IPI upgrade tier2 job), which one execution consumed *1.4TB!*, when I tarballed all the MG logs, it now consumes only *78GB*.

PR #14124, opened 2026-01-18
Remove excessive logging from ODFCLIRunner::run_noobaa

sagihirshfeld:
[ODFCLI::run_noobaa](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/helpers/odf_cli.py/#L294-L349) has been logging the *full* JSON output of every `mcg-cli` command.

Whether its while listing buckets, creating OBCs or checking on the health of NooBaa - whenever we use `run_noobaa` for these purposes, we're extracting the necessary sub-parts from the JSON and logging what's needed anyway. This makes the logging of the entire output excessive and often redundant.


sagihirshfeld:
Setting as `Verified` since this change  is only removing logging

PR #14131, opened 2026-01-19
Remove marker 'skip_resize_pre_conditions' from the autoscaler tests

yitzhak12:
Remove the marker 'skip_resize_pre_conditions' from the autoscaler tests, as we don't use it now.


yitzhak12:
The Jenkins job failure above was a temporary issue connecting to IBM Cloud.

Issue #14134, opened 2026-01-19
[Bug]: RDR test failed during teardown with Error from server (NotFound): persistentvolumeclaims "vs-busybox-pvc-1" not found"


sidhant-agrawal:
### Description
In one of the RDR test run, one CephFS test failed during teardown with Error from server (NotFound): persistentvolumeclaims "vs-busybox-pvc-1" not found"

Error Details:
```
failed on teardown with "ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc --kubeconfig /home/jenkins/clusters/j-rdr-i2-s43/openshift-cluster-dir/auth/kubeconfig -n appset-busybox-1-cephfs get PersistentVolumeClaim vs-busybox-pvc-1 -n appset-busybox-1-cephfs -o yaml.
Error is Error from server (NotFound): persistentvolumeclaims "vs-busybox-pvc-1" not found"
```

Testname: tests/functional/disaster-recovery/regional-dr/test_failover.py::TestFailover::test_failover[primary_down-cephfs-cli]
RP link: https://url.corp.redhat.com/21e6421

Issue #13538, opened 2025-11-06
[Bug]: RDR test failed during teardown with Error from server (NotFound): Unable to list "/v1, Resource=lists": the server could not find the requested resource


sidhant-agrawal:
### Description
In one of the test run, one CephFS test failed during teardown with Error from server (NotFound): Unable to list "/v1, Resource=lists": the server could not find the requested resource

Error details:
```
failed on teardown with "ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc --kubeconfig /home/jenkins/clusters/j-rdr-i1-s-146/openshift-cluster-dir/auth/kubeconfig get List -o yaml.
Error is Error from server (NotFound): Unable to list "/v1, Resource=lists": the server could not find the requested resource"
```

Stack trace:
```
def teardown():
        failed_to_delete = []
        for instance in instances:
            try:
>               instance.delete_workload(switch_ctx=ctx[0])

tests/conftest.py:7288: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ocs_ci/ocs/dr/dr_workload.py:476: in delete_workload
    backend_volumes = dr_helpers.get_backend_volumes_for_pvcs(
ocs_ci/helpers/dr_helpers.py:1197: in get_backend_volumes_for_pvcs
    backend_volume = pvc_obj.get_cephfs_subvolume_name
ocs_ci/ocs/resources/pvc.py:167: in get_cephfs_subvolume_name
    return self.backed_pv_obj.get()["spec"]["csi"]["volumeAttributes"][
ocs_ci/ocs/resources/ocs.py:105: in get
    return self.ocp.get(resource_name=self.name, out_yaml_format=out_yaml_format)
ocs_ci/ocs/ocp.py:342: in get
    return self.exec_oc_cmd(
ocs_ci/ocs/ocp.py:216: in exec_oc_cmd
    out = run_cmd(
ocs_ci/utility/utils.py:491: in run_cmd
    completed_process = exec_cmd(
ocs_ci/utility/retry.py:79: in f_retry
    return f(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[...]

PR #14135, opened 2026-01-19
Update filter to skip volsync PVCs with new naming pattern

sidhant-agrawal:
Update the filter to also skip PVCs that start with "vs-" in addition to "volsync". 

Fixes: #14134
Fixes: #13538


sidhant-agrawal:
Verified: https://url.corp.redhat.com/69358f4

keemano (approved):

PR #14189, opened 2026-01-23
[release-4.17] add possibility to pack mg log files to one tarball

dahorak:
This is backport of [PR#14117](https://github.com/red-hat-storage/ocs-ci/pull/14117).

The aim of this PR is to allow packing all individual MG (and related) files to one tarball. There are two reasons for this approach:
* save space (the tarballed MG files are 10 times smaller than the original files)
* allow easy download for local examination


dahorak:
Recent verification job failed on some tests cases, but the failures are not related to this PR and moreover it also verifies the must gather logs packing in case of test failure.

PR #14201, opened 2026-01-26
fix for getting multi arch image

petr-balogh:


petr-balogh:
Looks like in the case of 4.21, the right tag is found only on page 58, and it can be bigger page, so we cannot limit to 50 pages only here.

PR #14203, opened 2026-01-26
Add high priority tests for RHSTOR-7576 (NooBaa optimization utilizing secondary DB pod)

sagihirshfeld:
The high priority tests for https://issues.redhat.com/browse/RHSTOR-7576, which is about utilizing the secondary NooBaa DB pod for handling some of the read-only queries.

The added tests are:
-  `test_secondary_queries_after_cnpg_pod_respin[primary-db-pod]`
-  `test_secondary_queries_after_cnpg_pod_respin[secondary-db-pod]`

They both check the effect of deleting the CNPG pods on object data integrity and the persistence of the redirection of the read only queries to the secondary DB pod.

### DIFF
--- a/tests/functional/object/mcg/test_secondary_db_optimization.py
+++ b/tests/functional/object/mcg/test_secondary_db_optimization.py
@@ -1,7 +1,8 @@
+import pytest
 import json
 import logging
 from time import sleep
-import pytest
+from ocs_ci.framework.pytest_customization.marks import tier4
 
 from ocs_ci.framework import config
 from ocs_ci.framework.testlib import (
@@ -12,13 +13,21 @@ from ocs_ci.framework.testlib import (
     tier2,
 )
 from ocs_ci.ocs import constants
+from ocs_ci.ocs.bucket_utils import (
+    compare_directory,
+    sync_object_directory,
+    write_random_test_objects_to_bucket,
+)
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs.resources.pod import (
     get_pod_logs,
+    wait_for_pods_to_be_running,
 )
-from ocs_ci.utility.utils import get_secondary_nb_db_pod
+from ocs_ci.utility.utils import get_primary_nb_db_pod, get_secondary_nb_db_pod
 
 logger = logging.getLogger(__name__)
+
+WAIT_TIME = 90  # seconds
 
 
 @red_squad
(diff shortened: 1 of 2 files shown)


tests/functional/object/mcg/test_secondary_db_optimization.py
```diff
@@ -12,14 +13,22 @@
...
 from ocs_ci.ocs.resources.pod import (
     get_pod_logs,
+    wait_for_pods_to_be_running,
 )
-from ocs_ci.utility.utils import get_secondary_nb_db_pod
+from ocs_ci.utility.utils import get_primary_nb_db_pod, get_secondary_nb_db_pod
 
 logger = logging.getLogger(__name__)
 
+WAIT_TIME = 90  # seconds
```
udaysk23, line 30: Please remove the same varaible mentioned in `test_secondary_db_ro_queries`

tests/functional/object/mcg/test_secondary_db_optimization.py
```diff
@@ -104,59 +135,185 @@ def test_secondary_db_ro_queries(self, add_env_vars_to_noobaa_core):
...
+        second_bucket = bucket_factory(amount=1, interface="OC")[0].name
+
+        write_random_test_objects_to_bucket(
+            io_pod=awscli_pod,
+            bucket_to_write=second_bucket,
+            file_dir=test_directory_setup.origin_dir,
+            amount=OBJ_AMOUNT,
+            pattern="NbDbRespinB-",
+            mcg_obj=mcg_obj,
+        )
```
udaysk23, line 262: We can use existing data created for first bucket

sagihirshfeld: 1.`write_random_test_objects_to_bucket` is just as tried as tested an alternative way that would use the existing data. Runtime wise it doesn't too much either.
2. I'd rather not rely on `first.bucket` since it might be affected by prior tests, or assumed to be unused by future tests

tests/functional/object/mcg/test_secondary_db_optimization.py
sagihirshfeld: Fixed in https://github.com/red-hat-storage/ocs-ci/pull/14203/changes/cccb3ec49d898fa1f5eead4a468a3b092481d013

udaysk23 (approved):

PR #14062, opened 2026-01-08
Adds happy path coverage for noobaa db backup and recovery feature

udaysk23:
Adds happy path coverage for noobaa db backup and recovery process

PR #14206, opened 2026-01-27
Corrects volume snapshot class name for db backup and recovery process

udaysk23:
While testing MCG only deployment, I changed volumesnapshot class from `ocs-storagecluster-rbdplugin-snapclass` to `vpc-block-snapshot` and while doing some cosmetic changes in #14062 , Mistakenly I pushed the code with `vpc-block-snapshot` 

this value has been corrected in this PR

### DIFF
--- a/tests/functional/object/mcg/test_noobaa_db_backup_recovery.py
+++ b/tests/functional/object/mcg/test_noobaa_db_backup_recovery.py
@@ -114,7 +114,7 @@ class TestNoobaaDbBackupRecoveryOps:
         )
         # patch storagecluster object
         num_backups = 2
-        snapshot_class = "vpc-block-snapshot"
+        snapshot_class = constants.DEFAULT_VOLUMESNAPSHOTCLASS_RBD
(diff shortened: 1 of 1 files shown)


udaysk23:
Adding verified label since I already tested this code in #14062 PR and adding passed jenkins link below

https://url.corp.redhat.com/5074b42

tests/functional/object/mcg/test_noobaa_db_backup_recovery.py
```diff
@@ -114,7 +114,7 @@ def test_noobaa_db_backup_recovery_op(
         )
         # patch storagecluster object
         num_backups = 2
-        snapshot_class = "vpc-block-snapshot"
+        snapshot_class = "ocs-storagecluster-rbdplugin-snapclass"
```
prsurve: create a variable in constants.py and use that value jere

udaysk23: Fixed

Issue #9416, opened 2024-03-04
Option "output=avail" is not recognized by the 'df' command


jilju:
test_clone_when_full failing because option "output=avail" is not recognized by the 'df' command.

```
E               ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc -n namespace-test-69b8b89bb00b4abeba0047447 rsh pod-test-cephfs-317161ca73174be980eb720e df /var/lib/www/html -B M --output=avail.
E               Error is df: unrecognized option: output=avail
E               BusyBox v1.36.1 (2023-11-06 11:32:24 UTC) multi-call binary.
E
E               Usage: df [-PkmhTai] [-B SIZE] [-t TYPE] [FILESYSTEM]...
E
E               Print filesystem usage statistics
E
E                   -P    POSIX output format
E                   -k    1024-byte blocks (default)
E                   -m    1M-byte blocks
E                   -h    Human readable (e.g. 1K 243M 2G)
E                   -T    Print filesystem type
E                   -t TYPE    Print only mounts of this type
E                   -a    Show all filesystems
E                   -i    Inodes
E                   -B SIZE    Blocksize
E               command terminated with exit code 1



ocs_ci/utility/utils.py:685: CommandFailed
```
After installing "coreutils" using the command "apk add coreutils", the df command in the test case worked.

RP link - https://reportportal-ocs4.apps.ocp-c1.prod.psi.redhat.com/ui/#ocs/launches/632/18942/921411/921516/log

jilju:
Found in recent runs https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/994/41990/1933921/1934219/log

jilju:
Another instance https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#OCS/launches/954/42419/1952766/1953064/log

Platform is AWS in this run and in the failure mentioned in the previous comment.

jilju:
In 4.19.10 run https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/852/42405/1951170/1951447/log. (AWS)

https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/852/43329/1992224/1992515/log (VSPHERE)

This error is seen o AWS and VSphere

jilju:
[...]

PR #14265, opened 2026-02-01
add GNU coreutils to support advanced df flags

OdedViner:
#9416 
Issue: The current `nginx:alpine` image uses the BusyBox version of df, which lacks support for GNU-specific flags like `--output`. This causes scripts relying on these flags to fail with unrecognized option errors.

Solution: Installed the `coreutils` package via apk to provide the full GNU version of system utilities (including df).

https://quay.io/repository/ocsci/nginx?tab=tags&tag=latest


OdedViner:
@nehaberry @jilju Can we merge it?

PR #14280, opened 2026-02-03
Add foundation infrastructure for external cluster replica-1 

AviadP:
- [x] Add dataclasses and helpers for external cluster replica-1 provisioning    
- [x] Add ZoneConfig, TopologyReplica1Config and exec_external_ceph_cmd for replica-1

### DIFF
--- a/ocs_ci/deployment/helpers/external_cluster_helpers.py
+++ b/ocs_ci/deployment/helpers/external_cluster_helpers.py
@@ -3,6 +3,7 @@ This module contains helpers functions needed for
 external cluster deployment.
 """
 
+from dataclasses import dataclass
 import json
 import logging
 import re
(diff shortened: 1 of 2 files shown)


ocs_ci/deployment/helpers/external_cluster_helpers.py
```diff
@@ -44,6 +45,49 @@
...
+
+    zone_name: str
+    host_name: str
+    pool_name: str = ""
+
+    def __post_init__(self):
+        if not self.zone_name:
+            raise ValueError("zone_name cannot be empty")
+        if not self.host_name:
+            raise ValueError("host_name cannot be empty")
```
ebenahar, line 71: Should we also validate `pool_name`?

ocs_ci/deployment/helpers/external_cluster_helpers.py
```diff
@@ -44,6 +45,49 @@
...
+    Args:
+        zones (list[ZoneConfig]): List of zone configurations.
+        pool_prefix (str): Prefix for auto-generated pool names.
+        pg_num (int): Number of placement groups per pool.
+
+    """
+
+    zones: list[ZoneConfig]
+    pool_prefix: str = "rbd-zone"
+    pg_num: int = 32
```
ebenahar, line 88: is the 32 the default for replica 1 pools?

ocs_ci/deployment/helpers/external_cluster_helpers.py
```diff
@@ -44,6 +45,49 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass
+class ZoneConfig:
+    """
+    Configuration for a single zone in topology-based replica-1 setup.
```
ebenahar, line 51: what if the cluster is 3 zones? would it still be possible to configure replica 1?

ocs_ci/deployment/helpers/external_cluster_helpers.py
AviadP: its just a reasonable default. when test will be added, we can configure any other number, or several tests for each amount of pgs.

ocs_ci/deployment/helpers/external_cluster_helpers.py
AviadP: it (pool names) should be auto generated (will add it in next PR, i didnt want to submit huge PR so splitted it to 3)

ocs_ci/deployment/helpers/external_cluster_helpers.py
AviadP: `TopologyReplica1Config` accepts lists of zones

PR #14334, opened 2026-02-09
Reorder RDR upgrade tests

shylesh:


sidhant-agrawal (approved):

shylesh:
Verified: https://url.corp.redhat.com/verification

petr-balogh:
I think because of this change - I see:
2026-03-03 22:50:44  tests/functional/upgrade/test_upgrade.py::test_upgrade[200-2-1] 

Is triggered sooner than `test_dr_cluster_upgrade`, and then we see in `test_dr_cluster_upgrade`:
```
2026-03-03 23:42:25                      return csv_name
2026-03-03 23:42:25  >       raise CSVNotFound(f"No preupgrade CSV found for {resource_name}")
2026-03-03 23:42:25  E       ocs_ci.ocs.exceptions.CSVNotFound: No preupgrade CSV found for odr-cluster-operator
2026-03-03 23:42:25  
2026-03-03 23:42:25  ocs_ci/ocs/ocs_upgrade.py:573: CSVNotFound
```

Because it's already rolled out to 4.21.

Seen:
https://url.corp.redhat.com/5440a2f

AI Analysis:
```
Why this has failed?

⏺ Looking at this test failure, I can see the issue. The test is failing in the get_csv_name_pre_upgrade method when trying to find the pre-upgrade CSV for odr-cluster-operator.

  Let me examine the relevant code to understand what's happening:

⏺ Searched for 1 pattern, read 2 files (ctrl+o to expand)

⏺ Based on the logs and code analysis, here's why the test failed:

  Root Cause

  The test is failing because get_csv_name_pre_upgrade() cannot find a CSV that matches the pre-upgrade version.

  From the logs, I can see:
  - Current CSV: odr-cluster-operator.v4.21.0-rhodf (already upgraded version)
  - The method is looking for a CSV that contains both:
    a. odr-cluster-operator in the name ✓
    b. The pre-upgrade OCS version from config.PREUPGRADE_CONFIG.get("ENV_DATA").get("ocs_version") ✗

  Why This Happens

  The issue is that when run_upgrade() is called at line 84 of dr_upgrade.py, it's trying to capture the pre-upgrade CSV name, but:

  1. The cluster appears to already be upgraded - The only CSV present is v4.21.0 (the target upgrade version)
  2. The pre-upgrade CSV is missing - Either it was already replaced during an earlier upgrade step, or the cluster wasn't in the expected pre-upgrade state
  3. The PREUPGRADE_CONFIG doesn't match - The OCS version stored in config.PREUPGRADE_CONFIG.get("ENV_DATA").get("ocs_version") doesn't match "4.21.0", so the existing CSV is not recognized as the pre-upgrade version

  Likely Scenarios

  1. Upgrade already happened: The DR operator may have already been upgraded in a previous step, and this test is running on an already-upgraded cluster
  2. Config mismatch: The PREUPGRADE_CONFIG might not be properly set or might be referencing a different version than what's actually deployed
  3. CSV cleanup timing: The old CSV might have been removed before this check runs

  Let me check the test to see if there's any setup that should be happening:

⏺ Read 1 file (ctrl+o to expand)

⏺ The test is straightforward - it just instantiates DRClusterOperatorUpgrade() and calls run_upgrade(). The problem is in the timing/state.

  The fundamental issue: The code expects to find the pre-upgrade CSV still present when run_upgrade() starts, but only the post-upgrade CSV (v4.21.0) exists.

  Potential Fixes

  This could be addressed by:

  1. Checking if already upgraded: Before trying to get the pre-upgrade CSV, check if the current CSV version already matches the target upgrade version
  2. Better error handling: Provide more context about what CSVs were found and what version was expected
  3. Config validation: Ensure PREUPGRADE_CONFIG is properly initialized before the upgrade test runs

  Would you like me to:
  1. Investigate how PREUPGRADE_CONFIG gets initialized?
  2. Look at similar upgrade tests to see how they handle this?
  3. Propose a fix to handle the case where the cluster is already at the target version?
  ```

PR #14340, opened 2026-02-09
Skip Invoking setup ACM UI fixture when it's false

prsurve:


tests/functional/disaster-recovery/regional-dr/test_acm_kubevirt_dr_integration.py
```diff
@@ -42,17 +42,20 @@ class TestACMKubevirtDRIntergration:
     """
 
     @pytest.mark.parametrize(
-        argnames=["protection_type"],
+        argnames=["protection_type", "via_ui"],
```
keemano: This is not required for this test because all the tests in this test file require setup acm ui fixtures

tests/functional/disaster-recovery/regional-dr/test_warning_and_alerting.py
```diff
@@ -96,24 +96,24 @@ class TestRDRWarningAndAlerting:
     """
 
     @pytest.mark.parametrize(
-        argnames=[
-            "action",
-        ],
+        argnames=["action", "via_ui"],
```
keemano: Same. All the tests in this file need acm setup ui fixture.

tests/functional/disaster-recovery/regional-dr/test_rdr_monitoring_dashboard_ui.py
```diff
@@ -44,8 +44,17 @@ class TestRDRMonitoringDashboardUI:
     """
 
     @pytest.mark.polarion_id("OCS-5013")
+    @pytest.mark.parametrize(
+        argnames="via_ui",
```
keemano: Same as above. All UI specific tests need setu acm ui fixture. We can leave them as it is.

PR #14359, opened 2026-02-10
Use dr- prefix for hosted clusters in submariner installation stage

jilju:
Use dr- prefix for hosted clusters in submariner installation stage.
Update is_hosted_cluster to use the context of the given cluster name.


jilju:
Verified with the same change in another branch https://jenkins-csb-odf-qe-stage.dno.corp.redhat.com/job/qe-rdr-setup/179

PR #14336, opened 2026-02-09
Validating Node Availability (CPU & Memory) to Prevent Pod Scheduling Failures

s-selwin:
**Summary** 
Regression test (test_add_capacity specifically) is frequently failing due to Failed Pod Scheduling. Analysis identifies two primary blockers:

1. Resource Exhaustion: Insufficient Memory and CPU requests available on worker nodes.

2. Scheduling Constraints: Presence of Node Taints or missing Tolerations that prevent pod placement.

**Goal**
Implement a proactive "Gatekeeper" mechanism to validate cluster health and resource capacity before the test begins. This prevents long, expensive test runs from failing due to predictable infrastructure issues.

**Proposed Solution**
Auto-use Fixture: Develop a fixture (e.g., resource_check) that runs during the test setup phase.

Capacity Validation: Calculate "Real" available capacity (Allocatable minus current Reservations).

Pre-emptive Skipping: If the cluster cannot satisfy the test's resource requirements, the test should Skip immediately with a detailed breakdown of current resource "hogs" (e.g., leftover namespaces or ODF overhead).

Successfully implemented a parser that provides a namespace-level breakdown of RAM and CPU requests.

PR #14361, opened 2026-02-10
Validating Node Availability (CPU & Memory) to Prevent Pod Scheduling Failures

s-selwin:
Summary
Regression test (test_add_capacity specifically) is frequently failing due to Failed Pod Scheduling. Analysis identifies two primary blockers:

Resource Exhaustion: Insufficient Memory and CPU requests available on worker nodes.

Scheduling Constraints: Presence of Node Taints or missing Tolerations that prevent pod placement.

Goal
Implement a proactive "Gatekeeper" mechanism to validate cluster health and resource capacity before the test begins. This prevents long, expensive test runs from failing due to predictable infrastructure issues.

Proposed Solution
Auto-use Fixture: Develop a fixture (e.g., resource_check) that runs during the test setup phase.

Capacity Validation: Calculate "Real" available capacity (Allocatable minus current Reservations).

Pre-emptive Skipping: If the cluster cannot satisfy the test's resource requirements, the test should Skip immediately with a detailed breakdown of current resource "hogs" (e.g., leftover namespaces or ODF overhead).

Successfully implemented a parser that provides a namespace-level breakdown of RAM and CPU requests.

### DIFF
--- a/ocs_ci/ocs/node.py
+++ b/ocs_ci/ocs/node.py
@@ -7,7 +7,6 @@ from collections import defaultdict
 from operator import itemgetter
 import random
 import json
-import pytest
 import unicodedata
 from ocs_ci.ocs.ocp import exec_cmd
 import yaml
@@ -3539,7 +3538,6 @@ def get_pod_requests_per_node(worker_nodes=None):
     return usage
 
 
-def check_cluster_resource_availability(needed_ram_gb, needed_cpu_cores):
 def get_cluster_resource_capacity():
     """
     Parses 'oc describe node' for all worker nodes to calculate actual


s-selwin:
Closing old PR (https://github.com/red-hat-storage/ocs-ci/pull/14336) as I am moving the work to a fork per the team's contribution guidelines.

ocs_ci/ocs/node.py
```diff
@@ -3436,3 +3439,135 @@ def mark_masters_schedulable():
     assert scheduler_obj.patch(
         params=params, format_type="json"
     ), "Failed to run patch command to update control nodes as scheduleable"
+
+
+def check_cluster_resource_availability(needed_ram_gb, needed_cpu_cores):
```
AviadP: this function looks like a test function. i would suggest to split it into few functions with single responsibility, and aggregate everything as test function. also, i would suggest to avoid pytest.skip as part of the helper
this is not mandatory, up for your desicion

s-selwin: handled

tests/functional/z_cluster/cluster_expansion/test_add_capacity_entry_exit_criteria.py
```diff
@@ -48,6 +51,46 @@
 @skipif_managed_service
 @skipif_hci_provider_and_client
 class TestAddCapacity(ManageTest):
+    @pytest.fixture(autouse=True)
```
ebenahar: can we either move this to conftest.py or tranform it to a decorator, which we can use in other tests as well?

s-selwin: done

ocs_ci/ocs/node.py
```diff
@@ -3436,3 +3438,136 @@ def mark_masters_schedulable():
...
+                    "{:<45} | {:<12.2f} | {:<12.2f}".format(ns, ram_gb_used, cpu_c_used)
+                )
+
+        log.error(
+            f"RESOURCE GATEKEEPER FAILED. Available: {free_ram:.2f}GB RAM / {free_cpu:.2f} Cores. "
+            f"Required: {ram_gb}GB / {cpu_cores} Cores.\n" + "\n".join(table_rows)
+        )
+        return False
+
+    except Exception as e:
```
AviadP: plz avoid Exception - this might hide real issues

s-selwin: done

AviadP (approved):

ocs_ci/ocs/node.py
```diff
@@ -3534,3 +3536,138 @@ def get_pod_requests_per_node(worker_nodes=None):
...
+                    mem_usage = re.findall(r"(\d+[GMK]i)\s+\(", line)
+
+                    if cpu_usage:
+                        raw_c = cpu_usage[0]
+                        c_val = int(re.search(r"\d+", raw_c).group())
+                        if "m" not in raw_c:
+                            c_val *= 1000
+                        breakdown["cpu"][ns] += c_val
+
+                    if mem_usage:
```
ebenahar, line 3607: is it possible to take advantage of https://github.com/red-hat-storage/ocs-ci/blob/4d1122578f3281039b928db72c48020a29b0dbcf/ocs_ci/utility/utils.py#L4377?

s-selwin: Hi Elad

I actually looked into using convert_device_size, but I feel it’s safer to stick with the current manual parsing for this specific case. 

1. The issue is that the utility function doesn't support the "m" unit for CPU milli-cores, and as I'm parsing both CPU and RAM in the same loop, it keeps the logic a lot more consistent to handle them together. 
 
2. Also, convert_device_size seems a bit rigid because it expects exactly two-character units like "Mi"—if the oc describe output ever drops to a single-character unit or a raw integer, the utility would likely fail. 

My current regex is a bit more flexible with those mixed formats found in the pods section.

Issue #14333, opened 2026-02-08
test_bucket_policy_elements_NotAction[Deny] , test_bucket_policy_elements_NotAction[Allow] and test_bucket_policy_elements_NotPrincipal fail in 4.21


ypersky1980:
test_bucket_policy_elements_NotAction[Deny] , test_bucket_policy_elements_NotAction[Allow] and test_bucket_policy_elements_NotPrincipal fail in 4.21 with E           botocore.exceptions.ClientError: An error occurred (MalformedPolicy) when calling the PutBucketPolicy operation: Invalid principal in policy - a fix is needed. 
Relevant runs: 

https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/994/43966/2010661/2010740/log
https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/994/43966/2010661/2010739/log
https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/994/43966/2010661/2010738/log

Backtrace : 

self = <mcg.test_bucket_policy.TestS3BucketPolicy object at 0x7ff9cd825f50>
mcg_obj = <ocs_ci.ocs.resources.mcg.MCG object at 0x7ff9ed724650>
bucket_factory = <function bucket_factory_fixture.._create_buckets at 0x7ff9dd6a4e00>
effect = 'Deny'

@pytest.mark.parametrize(
    argnames="effect",
    argvalues=[
        pytest.param(
            *["Allow"], marks=[tier1, pytest.mark.polarion_id("OCS-5768")]
        ),
        pytest.param(*["Deny"], marks=[tier1, pytest.mark.polarion_id("OCS-5769")]),
    ],
)
def test_bucket_policy_elements_NotAction(self, mcg_obj, bucket_factory, effect):
    """
    Test bucket policy element of NotAction with Effect: Allow/Deny
    """

    # Creating obc and obc object to get account details, keys etc
    obc_bucket = bucket_factory(amount=2, interface="OC")
    obc_obj = OBC(obc_bucket[0].name)
    obc_obj1 = OBC(obc_bucket[1].name)

    # Set bucket policy for user
    bucket_policy_generated = gen_bucket_policy(
        user_list=obc_obj1.obc_account,
        action_property="NotAction",
        actions_list=["DeleteBucket"],
[...]

Issue #14332, opened 2026-02-08
test_bucket_policy_actions is constantly failing due to E           botocore.exceptions.ClientError: An error occurred (MalformedPolicy) when calling the PutBucketPolicy operation: Invalid principal in policy


ypersky1980:
test_bucket_policy_actions is constantly failing due to E           botocore.exceptions.ClientError: An error occurred (MalformedPolicy) when calling the PutBucketPolicy operation: Invalid principal in policy. 

Relevant runs: 

https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/994/43686/2004702/2004778/log
https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/994/43966/2010661/2010736/log

etc 

Backtrace: 
self = <mcg.test_bucket_policy.TestS3BucketPolicy object at 0x7fb7ce39c350>
mcg_obj = <ocs_ci.ocs.resources.mcg.MCG object at 0x7fb7d5bf0cd0>
bucket_factory = <function bucket_factory_fixture.._create_buckets at 0x7fb7d58c5080>

@provider_mode
@pytest.mark.polarion_id("OCS-2146")
@tier1
def test_bucket_policy_actions(self, mcg_obj, bucket_factory):
    """
    Tests user access to Put, Get, Delete bucket policy actions
    """
    # Creating obc and obc object to get account details, keys etc
    obc_name = bucket_factory(amount=1, interface="OC")[0].name
    obc_obj = OBC(obc_name)

    bucket_policy_generated = gen_bucket_policy(
        user_list=obc_obj.obc_account,
        actions_list=["PutBucketPolicy"],
        resources_list=[obc_obj.bucket_name],
    )
    bucket_policy = json.dumps(bucket_policy_generated)

    # Admin creates a policy on the user bucket, for Action: PutBucketPolicy
    logger.info(f"Creating policy by admin on bucket: {obc_obj.bucket_name}")

  put_policy = put_bucket_policy(mcg_obj, obc_obj.bucket_name, bucket_policy)

tests/functional/object/mcg/test_bucket_policy.py:409: 

ocs_ci/ocs/bucket_utils.py:1286: in put_bucket_policy
    return s3_obj.s3_client.put_bucket_policy(Bucket=bucketname, Policy=policy)
venv/lib64/python3.11/site-packages/botocore/client.py:595: in _api_call
[...]

Issue #14331, opened 2026-02-08
test_basic_bucket_policy_operations in 4.21 is constantly failing due to Invalid principal in policy - a fix is needed


ypersky1980:
test_basic_bucket_policy_operations in 4.21 is constantly failing due to Invalid principal in policy - a fix is needed

Relevant runs: 
https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/994/43966/2010661/2010735/log
https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/994/43775/2006856/2006938/log

etc 

backtrace : 

self = <mcg.test_bucket_policy.TestS3BucketPolicy object at 0x7f749137ec10>
mcg_obj = <ocs_ci.ocs.resources.mcg.MCG object at 0x7f7490160d50>
bucket_factory = <function bucket_factory_fixture.._create_buckets at 0x7f74915e7060>

@pytest.mark.polarion_id("OCS-2150")
@tier1
def test_basic_bucket_policy_operations(self, mcg_obj, bucket_factory):
    """
    Test Add, Modify, delete bucket policies
    """
    # Creating obc and obc object to get account details, keys etc
    obc_name = bucket_factory(amount=1, interface="OC")[0].name
    obc_obj = OBC(obc_name)

    bucket_policy_generated = gen_bucket_policy(
        user_list=obc_obj.obc_account,
        actions_list=["GetObject"],
        resources_list=[obc_obj.bucket_name],
    )
    bucket_policy = json.dumps(bucket_policy_generated)

    # Add Bucket Policy
    logger.info(f"Creating bucket policy on bucket: {obc_obj.bucket_name}")

  put_policy = put_bucket_policy(mcg_obj, obc_obj.bucket_name, bucket_policy)

/home/jenkins/workspace/qe-odf-provider-client-multicluster@2/ocs-ci/tests/functional/object/mcg/test_bucket_policy.py:304: 

/home/jenkins/workspace/qe-odf-provider-client-multicluster@2/ocs-ci/ocs_ci/ocs/bucket_utils.py:1286: in put_bucket_policy
    return s3_obj.s3_client.put_bucket_policy(Bucket=bucketname, Policy=policy)
/home/jenkins/workspace/qe-odf-provider-client-multicluster@2/ocs-ci/venv/lib64/python3.11/site-packages/botocore/client.py:595: in _api_call
[...]

PR #14364, opened 2026-02-10
Adding a fix to policy scenarios - adjust Principal to contain ARN

ypersky1980:
This is is a fix to https://github.com/red-hat-storage/ocs-ci/issues/14333, https://github.com/red-hat-storage/ocs-ci/issues/14332 and https://github.com/red-hat-storage/ocs-ci/issues/14331. 
In 4.21 format of Principal in policy changed, it no longer contains a an account name but ARN. 
The tests were adjusted accordingly.

### DIFF
--- a/ocs_ci/ocs/bucket_utils.py
+++ b/ocs_ci/ocs/bucket_utils.py
@@ -3587,20 +3587,3 @@ def get_bucket_status_value(mcg_obj, bucket_name, key):
     op = bucket_status.split("\n")
     value = next(item.split(":")[1].strip() for item in op if key in item)
     return value
-
-
-def get_bucket_policy_user_list(obc_obj):
-    """
-    Helper function returning user list used to construct bucket policy depending on oc version
-    Args:
-        obc_obj (obj): Obc object
-    Returns:
-        str: user list to be used in policy
-    """
-    ocs_version = version.get_semantic_ocs_version_from_config()
-
-    user_list = (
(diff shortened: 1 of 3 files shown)


ocs_ci/ocs/bucket_utils.py
```diff
@@ -3587,3 +3587,20 @@ def get_bucket_status_value(mcg_obj, bucket_name, key):
     op = bucket_status.split("\n")
     value = next(item.split(":")[1].strip() for item in op if key in item)
     return value
+
+
+def get_bucket_policy_user_list(obc_obj):
```
sagihirshfeld: This gets the job done, but its readability isn't great:
1. It accepts only OBCs but the name doesn't make it clear. Policy identification is also relevant for MCG accounts.
2. If this is relevant only to OBC, it's better to encapsulate this at the responsible class.

How about we add a  read-only property to the OBC class instead?
```
    @property
    def id_for_policies(self):
        """
        Returns the ID for the OBC account to be used in bucket policies,
        based on the OCS version.

        Returns:
            str: The ID for the OBC account to be used in bucket policies.
        """
        ocs_version = version.get_semantic_ocs_version_from_config()
        return self.obc_arn if ocs_version >= version.VERSION_4_21 else self.obc_account
```

ypersky1980: Fixed. The new name is : user_list=obc_obj.id_for_policy_principal

sagihirshfeld (approved):

PR #14375, opened 2026-02-11
Reduction of logging for timeout sampler

petr-balogh:


ocs_ci/utility/utils.py
```diff
@@ -1757,15 +1757,23 @@ def stringify(value):
...
+            attempt += 1
             try:
                 yield self.func(*self.func_args, **self.func_kwargs)
-            except Exception as ex:
-                msg = f"Exception raised during iteration: {ex}"
-                log.exception(msg)
+            except Exception:
+                log.info(
+                    f"TimeoutSampler attempt {attempt} for function '{self.func.__name__}' failed, "
+                    "see debug level logs for details"
```
clacroix12: This seems like a good reduction in the logs so far. I'd like to further analyze this and see if we could implement some sort of buffer to when we log. Many of the implementations of TimeoutSampler have sleep times of only a few seconds which means a lot of duplicate INFO statements. If we notice that this is creating more noise than value let's try to reduce it further by creating a minimum time to wait between these INFO level timeout messages.

clacroix12 (approved):

vavuthu (approved):

PR #14383, opened 2026-02-11
Default FusionServiceInstance backingStorageType

clacroix12:
1. Default FusionServiceInstance backingStorageType to Dynamic
2. Make the backingStorageType configurable
3. Set value to Local for VSphere and Baremetal deployments


clacroix12:
Verified the value is being set properly for vsphere: https://url.corp.redhat.com/d880209

PR #14405, opened 2026-02-13
Add possibility to stop execution and skip rest of test cases

petr-balogh:


docs/graceful_stop_feature.md
```diff
@@ -0,0 +1,214 @@
...
+### Configuration Flags
+
+The following flags are set in `ocsci_config.RUN`:
+- `stop_requested` (bool): True when any stop file is detected
+- `graceful_stop` (bool): True for `.stop_gracefully`, False for `.stop`
+
+## Best Practices
+
+1. **Use `.stop_gracefully` by default** unless you're in a hurry
+2. **Remove stop files** after use to avoid accidentally stopping future runs
```
dahorak, line 170: What about automate this stop file removal as part of the process?

petr-balogh: we can if this will bother us. But I don't think we will heavily use this option and every run we clean up the workspace so it will be deleted from job I think

PR #14414, opened 2026-02-13
Fix UI deployment

petr-balogh:


petr-balogh:
https://url.corp.redhat.com/10e9252
 
 I had this commit in different PR which I was testing with, so considering this as verified

ebondare (approved):

PR #14420, opened 2026-02-16
Reduce parametrized args in test_csi_addons_socket_creation_per_pods_node

clacroix12:
We are seeing OSErrors when attempting to upload this test case to report portal due to the file length of the test directory. The parametrized values in the test case name account for roughly 140 characters. This fix intends to reduce the file length enough to avoid the errors without changing the functionality of the test.


clacroix12:
@red-hat-storage/brown-squad please review and verify this change is acceptable

PR #14423, opened 2026-02-17
Add RGW bucket UI support with selector and tests

AviadP:
- [x] Add endpoint selection methods to ObjectStorage page object
- [x] Extend navigation and bucket operations to support provider param
- [x] Add 4 UI tests for RGW bucket lifecycle
- [x] Tests use runtime RGW pod detection to skip on clusters without RGW

## Summary by CodeRabbit

- **New Features**
  - Added support for selecting and navigating between NooBaa and RGW Internal S3 providers in the object-storage interface.
  - Added RGW provider availability detection.
  - Enabled bucket creation, listing, navigation, and deletion through the RGW Object Browser.

- **Bug Fixes**
  - Improved bucket navigation and deletion flows when multiple storage providers are available.

- **Tests**
  - Added functional coverage for RGW bucket browsing, creation, listing, and deletion.

### DIFF
--- a/tests/functional/object/mcg/ui/test_rgw_bucket_ui.py
+++ b/tests/functional/object/mcg/ui/test_rgw_bucket_ui.py
@@ -1,0 +1,152 @@
+import logging
+import time
+
+import pytest
+
+from ocs_ci.framework.pytest_customization.marks import (
+    red_squad,
+    rgw,
+    skipif_hci_provider_and_client,
+)
+from ocs_ci.framework.testlib import (
+    tier1,
+    tier2,
+    ui,
+    post_upgrade,
+)
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.resources.pod import get_rgw_pods
+from ocs_ci.ocs.ui.page_objects.buckets_tab import BucketsTab
+
+
+logger = logging.getLogger(__name__)
+
+
+@ui
+@rgw
+@red_squad
+@skipif_hci_provider_and_client
+class TestRGWBucketUI:
+    """
+    Test RGW bucket operations via the Object Browser UI.
+    Requires ODF 4.19+ with RGW provider support.
+    """
+
+    @tier1
+    @post_upgrade
+    @pytest.mark.polarion_id("OCS-7746")
+    def test_navigate_to_rgw_object_browser(self, setup_ui_class):
+        """
+        Verify navigation to the RGW object browser and bucket list loading.
+
+        Steps:
+        1. Navigate to Object Storage page
+        2. Verify RGW provider card is available and enabled
+        3. Select RGW provider
+        4. Verify bucket list loads
+
+        """
+        if not get_rgw_pods():
+            pytest.skip("No RGW pods running on this cluster")
+        bucket_ui = BucketsTab()
+        obj_storage = bucket_ui.nav_object_storage_page()
+
+        if not obj_storage.is_rgw_provider_available():
+            pytest.skip("RGW provider card is not available or is disabled")
+
+        obj_storage.select_storage_provider(constants.S3_PROVIDER_RGW_INTERNAL)
+        bucket_ui.page_has_loaded(retries=15)
+
+        buckets = bucket_ui.get_buckets_list()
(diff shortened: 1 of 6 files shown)


sagihirshfeld (dismissed):

tests/functional/object/mcg/ui/test_rgw_bucket_ui.py
```diff
@@ -0,0 +1,156 @@
...
+
+        Steps:
+        1. Navigate to RGW bucket list, create a test bucket
+        2. Delete the bucket using three_dots menu
+        3. Verify bucket no longer appears in the list
+
+        """
+        if not get_rgw_pods():
+            pytest.skip("No RGW pods running on this cluster")
+        setup_ui_class_factory()
```
DanielOsypenko: not high importance, but still does not make sense.
- why to use factory in this case? we have `setup_ui_class`

tests/functional/object/mcg/ui/test_rgw_bucket_ui.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+        Delete an RGW bucket via the UI.
+
+        Steps:
+        1. Navigate to RGW bucket list, create a test bucket
+        2. Delete the bucket using three_dots menu
+        3. Verify bucket no longer appears in the list
+
+        """
+        if not get_rgw_pods():
```
DanielOsypenko, line 132: I think for this we have @rgw marker

tests/functional/object/mcg/ui/test_rgw_bucket_ui.py
```diff
@@ -0,0 +1,156 @@
...
+        Steps:
+        1. Navigate to RGW bucket list
+        2. Create a test bucket if none exist
+        3. Verify bucket list is non-empty
+        4. Verify bucket names are displayed
+
+        """
+        if not get_rgw_pods():
+            pytest.skip("No RGW pods running on this cluster")
+        setup_ui_class_factory()
```
DanielOsypenko: please use @rgw and `setup_ui_class`

tests/functional/object/mcg/ui/test_rgw_bucket_ui.py
```diff
@@ -0,0 +1,156 @@
...
+        time.sleep(10)
+
+        bucket_ui.navigate_buckets_page(provider=constants.S3_PROVIDER_RGW_INTERNAL)
+        buckets = bucket_ui.get_buckets_list()
+        assert (
+            bucket_name in buckets
+        ), f"Bucket '{bucket_name}' not found in RGW bucket list: {buckets}"
+
+    @tier2
+    @pytest.mark.polarion_id("OCS-XXXX")
```
DanielOsypenko: probably better to add proper id before merge

tests/functional/object/mcg/ui/test_rgw_bucket_ui.py
```diff
@@ -0,0 +1,156 @@
...
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.resources.pod import get_rgw_pods
+from ocs_ci.ocs.ui.page_objects.buckets_tab import BucketsTab
+
+
+logger = logging.getLogger(__name__)
+
+
+@ui
+@rgw
```
DanielOsypenko, line 26: please check, if we use @rgw, should we check wethere there are rgw pods? Or maybe in this case it is better to fail, when no rgw pods are found

tests/functional/object/mcg/ui/test_rgw_bucket_ui.py
```diff
@@ -0,0 +1,156 @@
...
+@black_squad
```
DanielOsypenko: please check with Red squad, if this test should be under their ownership. 
@shyRozen what is your view on this?

tests/functional/object/mcg/ui/test_rgw_bucket_ui.py
AviadP: im not sure that this marker is useful, it doesnt seems like it skips clusters without rgw

Issue #13898, opened 2025-12-10
[Bug]: test_negative_custom_taint Failed inconsistently  with error pods "ceph-object-controller-detect-version-XXXX" not found


suchita-g:
### Description
[Bug]: test_negative_custom_taint Failed inconsistently  with error pods "ceph-object-controller-detect-version-XXXX" not found
[OCS4-20-Downstream-OCP4-20-VSPHERE6-UPI-FIPS-1AZ-RHCOS-VSAN-3M-6W-tier4b-post-upgrade-jenkins](https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/954/40405/1885931)
test_negative_custom_taint
### Steps to Reproduce
1.[OCS4-20-Downstream-OCP4-20-VSPHERE6-UPI-FIPS-1AZ-RHCOS-VSAN-3M-6W-tier4b-post-upgrade-jenkins](https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/954/40405/1885931)
test_negative_custom_taint
2.
3.

### Actual Behavior
test_negative_custom_taint Failed
### Expected Behavior
test_negative_custom_taint should pass
### Environment
- Test Suite(s):
- Platform(s):
- Version(s): 4.20.
- OS:

### Additional Context
self = <tests.functional.z_cluster.nodes.test_non_ocs_taint_and_toleration.TestNonOCSTaintAndTolerations object at 0x7f7ce3d240d0>
nodes = <ocs_ci.ocs.platform_nodes.VMWareUPINodes object at 0x7f7ce39aded0>

@pytest.mark.polarion_id("OCS-5983")
def test_negative_custom_taint(self, nodes):
    """
    Test runs the following steps
    1. Taint odf nodes with non-ocs taint
    2. Set toleration in storagecluster yaml.
    3. Set toleration in wrong subscription yaml.
    4. Check that toleration is not applied on all subscriptions and pods.
    5. Check that all pods are not in running state.

    """

    logger.info("Taint all odf nodes with custom taint")
    ocs_nodes = get_ocs_nodes()
    for node in ocs_nodes:
        taint_nodes(nodes=[node.name], taint_label="xyz=true:NoSchedule")
    resource_name = constants.DEFAULT_CLUSTERNAME
    if config.DEPLOYMENT["external_mode"]:
[...]

Issue #13663, opened 2025-11-18
[Bug]: test_negative_custom_taint Failed with Error CommandFailed :  rook-ceph-osd-0-xxxx" not found


suchita-g:
### Description
test_negative_custom_taint Failed with Error CommandFailed :  rook-ceph-osd-0-xxxx" not found

```

          raise CommandFailed(
                f"Error during execution of command: {masked_cmd}."
                f"\nError is {masked_stderr}"
            )
E               ocs_ci.ocs.exceptions.CommandFailed: Error during execution of command: oc --kubeconfig /home/jenkins/current-cluster-dir/openshift-cluster-dir/auth/kubeconfig -n openshift-storage delete Pod rook-ceph-osd-0-848d5b8d64-bsspn --wait=false.
E               Error is Error from server (NotFound): pods "rook-ceph-osd-0-848d5b8d64-bsspn" not found



ocs_ci/utility/utils.py:737: CommandFailed


```

### Steps to Reproduce
1.  Deploy ODF Cluster 
2. Run test test_negative_custom_taint
3.
[OCS4-20-Downstream-OCP4-20-VSPHERE6-UPI-FIPS-1AZ-RHCOS-VSAN-3M-6W-tier4b-post-upgrade-jenkins](https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/954/39880/1869628)
test_negative_custom_taint

https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/954/39880/1869628/1869674/log?logParams=history%3D1869673%26page.page%3D1

### Expected Behavior
test_negative_custom_taint should pass
### Environment
- Test Suite(s):
- Platform(s):
- Version(s):
- OS:

suchita-g:
As PR is merged Closing this issue

PR #14424, opened 2026-02-17
Fix test_negative_custom_taint handling of already-deleted pods

suchita-g:
Fixes: #13900 #13898 #13663
When deleting pods from a snapshot, some pods may already be gone (replaced by controllers) by the time we try to delete them, causing NotFound. Catch CommandFailed for NotFound/not found and skip instead of failing


vavuthu (approved):

amr1ta (approved):

shylesh (approved):
lgtm

PR #14429, opened 2026-02-18
workload_ui: fix retry usage, wait for all containers, add container names in logs

dgaikwad:
# PR: Fix and improve wait_for_container_status_ready in workload_ui

## Title (for PR header)

```
Fix and improve wait_for_container_status_ready: retry usage, all containers check, logging
```

---

### Summary

Updates `wait_for_container_status_ready()` in `ocs_ci/ocs/ui/workload_ui.py` to fix incorrect `retry()` usage, wait for **all** containers in a pod (not only the first), add explicit while-not-ready logic, improve logging with container names,and use safe access for pod status. Transient "can't read container state of deployment" errors are retried correctly.

### File changed

| File | Scope |
|------|--------|
| `ocs_ci/ocs/ui/workload_ui.py` | `wait_for_container_status_ready()` and inner `do_wait_for_container_status_ready()` |

---

#### 1. Fix `retry()` usage

- **Before:** `retry(..., func=do_wait_for_container_status_ready)(pod)` — the decorator was given `pod` (a Pod object) instead of the function, so the wrong callable was retried.
- **After:** `retry(CommandFailed, text_in_exception="can't read container state of deployment")(do_wait_for_container_status_ready)(pod)` — the function is passed to the decorator and the wrapped function is called with `(pod)`.
- Removed the incorrect `func=` argument (`func` in this codebase is a no-arg hook, not the retried function).

#### 2. Check all containers, not only the first

- **Before:** Only `containerStatuses[0]` was checked; multi-container pods could be considered ready when only one container was ready.
- **After:** Iterates over all `containerStatuses` and uses `all(cs.get("ready", False) for cs in container_statuses)`. Returns `True` only when every container has `ready=True`.

#### 3. Explicit "while not ready" check

- **Before:** Loop condition was only `time.time() - start_time < timeout`.
- **After:** `all_ready = False` and loop condition `while not all_ready and (time.time() - start_time < timeout)` so the loop explicitly runs while containers are not ready and within timeout.

#### 4. Safe access to pod status

- Uses `.get("status")`, `.get("containerStatuses")`, and `.get("ready", False)` so missing or incomplete status (e.g. pod still initializing) does not raise; the loop continues and retries.

#### 5. Container names in logs

- **Success:** Logs all container names when ready: `All container(s) in pod <name> are ready: ['c1', 'c2', ...]`.
- **Waiting:** Logs which containers are not ready: `Waiting for container(s) to be ready in pod <name>: ['not_ready_1', ...]`.
- **No status yet:** Logs `No container statuses yet, waiting...`.

#### 6. Docstring and return value

- Docstring updated to state that **all** containers are waited on and that the return value is `True` if all are ready within timeout, `False` otherwise.
- Inner function returns `True` when all containers are ready, `False` on timeout.

---

### Code flow (after changes)

1. Log that we are waiting for all containers of the pod.
2. Inner function `do_wait_for_container_status_ready(pod_obj, timeout=300)`:
   - Log wait message with pod name and timeout.
   - `all_ready = False`; loop `while not all_ready and (within timeout)`.
   - In loop: get `status.containerStatuses` (with safe `.get()`).
   - If no container statuses yet, log and sleep 5s, continue.
   - Compute `container_names` and `all_ready = all(ready for each container)`.
   - If `all_ready`: log container names, return `True`.
   - Else: compute `not_ready` names, log them, sleep 5s.
   - After loop: return `False` (timeout).
3. Call the inner function through `retry(CommandFailed, text_in_exception="...")` so transient "can't read container state of deployment" errors are retried.

---

### DIFF
--- a/ocs_ci/ocs/ui/workload_ui.py
+++ b/ocs_ci/ocs/ui/workload_ui.py
@@ -434,27 +434,32 @@ def compare_mem_usage(
     return deviation <= deviation_accepted
 
 
-def wait_for_container_status_ready(pod: Pod):
+def wait_for_container_status_ready(pod: Pod, timeout=300, interval=5):
     """
     Wait for all containers of the pod to move to ready state.
 
     :param pod: a pod object of the pod whose containers are to be checked
+    :param timeout: overall timeout in seconds
+    :param interval: polling interval in seconds
     :return: True if all containers are ready within timeout, False otherwise
     """
     logger.info(f"Wait for all containers of the pod {pod.name} to be ready")
 
-    def do_wait_for_container_status_ready(pod_obj: Pod, timeout=300):
+    def do_wait_for_container_status_ready(
+        pod_obj: Pod, wait_timeout=timeout, wait_interval=interval
+    ):
         logger.info(
-            f"Waiting for all containers in pod {pod_obj.name} to be ready for {timeout}s"
+            f"Waiting for all containers in pod {pod_obj.name} to be ready for {wait_timeout}s"
         )
         start_time = time.time()
         all_ready = False
-        while not all_ready and (time.time() - start_time < timeout):
-            pod_status = pod_obj.get().get("status") or {}
+        while not all_ready and (time.time() - start_time < wait_timeout):
+            pod_data = pod_obj.get()
+            pod_status = pod_data["status"] if isinstance(pod_data, dict) else {}
             container_statuses = pod_status.get("containerStatuses") or []
             if not container_statuses:
                 logger.info("No container statuses yet, waiting...")
-                time.sleep(5)
+                time.sleep(wait_interval)
                 continue
             container_names = [cs.get("name", "?") for cs in container_statuses]
             all_ready = all(cs.get("ready", False) for cs in container_statuses)
@@ -471,7 +476,7 @@ def wait_for_container_status_ready(pod: Pod):
             logger.info(
                 f"Waiting for container(s) to be ready in pod {pod_obj.name}: {not_ready}"
             )
-            time.sleep(5)
+            time.sleep(wait_interval)
         return False
 
     retry(


dgaikwad:
[PR CHECK](https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/64561/consoleFull)

dgaikwad:
https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/66620/testReport/

ocs_ci/ocs/ui/workload_ui.py
```diff
@@ -436,27 +436,49 @@ def string_to_bytes(s):
...
+                cs.get("name", "?")
+                for cs in container_statuses
+                if not cs.get("ready", False)
+            ]
+            logger.info(
+                f"Waiting for container(s) to be ready in pod {pod_obj.name}: {not_ready}"
+            )
             time.sleep(5)
-        return pod_obj.get()["status"]["containerStatuses"][0]["ready"]
+        return False
```
posharma1, line 480: why hardcoded False ?

dgaikwad: Previously readiness reflected one container; it now reflects all containers: True iff every containerStatuses[].ready is true, False otherwise.

ocs_ci/ocs/ui/workload_ui.py
```diff
@@ -436,27 +436,49 @@ def string_to_bytes(s):
...
-            and time.time() - start_time < timeout
-        ):
-            logger.info("Waiting for container status ready")
+        all_ready = False
+        while not all_ready and (time.time() - start_time < timeout):
+            pod_status = pod_obj.get().get("status") or {}
+            container_statuses = pod_status.get("containerStatuses") or []
+            if not container_statuses:
+                logger.info("No container statuses yet, waiting...")
+                time.sleep(5)
```
am-agrawa: What if containers are not up in 5 seconds? How are we handling that?

dgaikwad: Handled by the outer timeout loop. The sleep(5) is just the polling interval; we keep rechecking until the overall timeout=300 in do_wait_for_container_status_ready() is reached. If containerStatuses is still empty by then, the helper exits the loop and returns False.

am-agrawa: I don't think this timeout here inside the loop is acting as a polling interval. Polling interval as an arg to the function would be a better approach I guess.

dgaikwad: Expected changes are incorporated.

PR #14433, opened 2026-02-18
tests: log non-running pods on test_reboot_on_tainted_node failure

suchita-g:
Add diagnostics so that when wait_for_pods_to_be_running fails we log a sample of non-running pods (name + status). Helps triage automation vs product (e.g. Pending with "untolerated taint").
may Fix issues #14430

### DIFF
--- a/tests/functional/z_cluster/nodes/test_non_ocs_taint_and_toleration.py
+++ b/tests/functional/z_cluster/nodes/test_non_ocs_taint_and_toleration.py
@@ -27,6 +27,7 @@ from ocs_ci.ocs.exceptions import (
 )
 from ocs_ci.ocs.resources.pod import (
     get_all_pods,
+    get_not_running_pods,
     wait_for_pods_to_be_running,
     check_toleration_on_pods,
     check_toleration_on_subscriptions,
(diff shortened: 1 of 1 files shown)


suchita-g:
https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/64559/

suchita-g:
https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/65208/

tests/functional/z_cluster/nodes/test_non_ocs_taint_and_toleration.py
```diff
@@ -191,6 +192,30 @@ def test_reboot_on_tainted_node(self, nodes):
         logger.info("Apply custom taints and tolerations.")
         apply_custom_taint_and_toleration()
 
+        logger.info(
+            "After adding toleration wait for some time for pods to respin as expected"
+        )
+        time.sleep(300)
+
+        def _ensure_pods_running():
```
AviadP, line 200: this seems like local solution for broad problem. i would suggest to add this solution to `wait_for_pods_to_be_running` (if u want to skip wide range testing, can add argument if enable this check with false as default)

PR #14438, opened 2026-02-18
destroy client cluster with hcp

DanielOsypenko:
Destroy HCP cluster
Destroy AWS resources
Destroy client OC components

Utility function to destroy particular HCP cluster. Full destroy of Management and HCP.

Now destroy of a regular AWS sts ipi cluster destroys also associated Hosted CP clusters if they found. 
**No Multi-cluster job, just regular destroy.**

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -533,7 +533,7 @@
         "hashed_secret": "1348b145fa1a555461c1b790a2f66614781091e9",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 5215,
+        "line_number": 6771,
         "type": "Private Key",
         "verified_result": null
       }
@@ -738,6 +738,14 @@
         "is_secret": false,
         "is_verified": false,
         "line_number": 2139,
+        "type": "Secret Keyword",
+        "verified_result": null
+      },
+      {
+        "hashed_secret": "9ec53cfd9929c70c3f87c210b6a7b77fb6d79d43",
+        "is_secret": false,
+        "is_verified": false,
+        "line_number": 2718,
         "type": "Secret Keyword",
         "verified_result": null
       },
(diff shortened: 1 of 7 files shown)


DanielOsypenko:
Passed with 1st path - when destroying cluster with `hypershift destroy` command works as expected
https://url.corp.redhat.com/092bcff

DanielOsypenko:
Passed with 1st path - when destroying cluster with hypershift destroy command works as expected https://url.corp.redhat.com/d5232a6

DanielOsypenko:
Verified: Destroy of a regular AWS ipi cluster destroys also Hosted HCP clusters
https://url.corp.redhat.com/3bc6999

dahorak (approved):

DanielOsypenko:
> The current code does this:
> 
> > except (CommandFailed, TimeoutExpiredError) as e:
> > logger.error(...)
> > return False
> 
> > except Exception as e:
> > if "TimeoutExpired" in type(e).**name**:  # <-- checking by string name
> > logger.error(...)
> > return False
> > raise
> 
> The problem: it catches a broad Exception and then checks if the exception's class name contains "TimeoutExpired" as a string. This is fragile because:
> 
> * If the exception class gets renamed, it silently stops matching
> * It could accidentally match unrelated exceptions with "TimeoutExpired" in their name
> 
> The fix: figure out which specific timeout exception is leaking through (likely subprocess.TimeoutExpired) and add it to the first except clause directly:
> 
> except (CommandFailed, TimeoutExpiredError, subprocess.TimeoutExpired) as e: logger.error(...) return False
> 
> No need for the second except Exception block at all.

Thank Amrita for checking this. 
This code must be super reliable, because the general approach is - more resources we are able remove before failing the job is better. Even if one error occurs we'd want to move further and remove other resources.

Also there is no risk on hiding even a possible bug, because these are all aws components we'd want to remove them to not pay for their use, hence I'd rather catch exceptions with broad Exception and continue with the build run. 

If this is not a critical issue on your opinion, I'd rather continue with this code

Issue #14446, opened 2026-02-20
[Bug]: Recent PR#14363 break OCP deployment


dahorak:
### Description

Commit: 7157b35ac  (Feb 19, 2026) - "Fix report properties for FDF deployments (#14363)"

This commit changed the import in `ocs_ci/utility/reporting.py:4-5:

Before (working):
```
from ocs_ci.utility.utils import get_ocp_version, get_testrun_name
```

After (broken):
```
from ocs_ci.ocs.version import get_ocp_version
from ocs_ci.utility.utils import get_testrun_name
```

**The Key Difference**

These are two different functions with the same name:

1. ocs_ci.utility.utils.get_ocp_version() (old, working):
    - Gets version from config.DEPLOYMENT["installer_version"] by default
    - Only queries the cluster if skip_ocp_deployment is set
    - Works for deployment tests because it uses the config version (from --ocp-version 4.21)
    - Returns short version like "4.21"
2. ocs_ci.ocs.version.get_ocp_version() (new, broken):
    - Always queries the cluster: runs oc get clusterversion version -o yaml
    - Returns full version like "4.10.0-0.nightly-2022-02-09-111355"
    - Fails when kubeconfig doesn't exist (during deployment tests)

### Steps to Reproduce
1. Trigger OCP deployment from `master` branch.

### Actual Behavior
OCP Deployment broken

### Expected Behavior
OCP deployment should work

### Impact (likelihood of reproduction, impact on the cluster and on other tests, etc.)
100%

PR #14447, opened 2026-02-20
Fix get_ocp_version() to handle deployment scenarios without kubeconfig

dahorak:
Commit 7157b35ac changed reporting.py to use ocs_ci.ocs.version.get_ocp_version() instead of the deprecated ocs_ci.utility.utils.get_ocp_version(). However, the new function always queries the cluster, which fails during OCP deployment tests when the kubeconfig doesn't exist yet.

This fix enhances get_ocp_version() to:
- Try to get version from the running cluster first
- Catch CommandFailed exception when cluster is not accessible
- Fall back to config.DEPLOYMENT['installer_version'] from CLI args

This matches the behavior of the deprecated function while using the modern API.

Fixes https://github.com/red-hat-storage/ocs-ci/issues/14446

### DIFF
--- a/ocs_ci/ocs/version.py
+++ b/ocs_ci/ocs/version.py
@@ -122,7 +122,24 @@ def get_ocp_version(version_dict=None):
     Returns:
       str: full version string of OCP cluster
     """
-    version_dict = get_ocp_version_dict()
+    if version_dict is None:
+        try:
+            version_dict = get_ocp_version_dict()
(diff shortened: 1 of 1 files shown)


dahorak:
Verification job: https://url.corp.redhat.com/535255a

dahorak:
Second verification job (after changed log message): https://url.corp.redhat.com/9b89118

PR #14448, opened 2026-02-20
DRS: exclude clusterNetwork, serviceNetwork and link-local CIDRs from IP selection and return interface name

fbalak:


fbalak:
Verified: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-odf-provider-client-multicluster/1180/

PR #14490, opened 2026-02-25
checking if it is a hci in case of storage setup

rh-aaleem:
This PR is just putting check if it is a hci baremetal in that case it will setup storage like odf,

### DIFF
--- a/ocs_ci/deployment/fusion_data_foundation.py
+++ b/ocs_ci/deployment/fusion_data_foundation.py
@@ -478,4 +478,5 @@ class CustomResourceDefinition(OCP):
             resource_name=resource_name,
             kind="CustomResourceDefinition",
             *args,
-            **kwargs,        )
+            **kwargs,
+        )
--- a/ocs_ci/deployment/fusion_data_foundation.py
+++ b/ocs_ci/deployment/fusion_data_foundation.py
@@ -39,7 +39,7 @@ logger = logging.getLogger(__name__)
 
 class FusionDataFoundationDeployment:
     def __init__(self):
-        self.pre_release = config.DEPLOYMENT.get("fdf_pre_release", False)
+        self.pre_release = str(config.DEPLOYMENT.get("fdf_pre_release", False)).lower() == "true"
         self.kubeconfig = config.RUN["kubeconfig"]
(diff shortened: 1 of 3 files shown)


vavuthu (dismissed):

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -478,5 +478,4 @@ def __init__(self, resource_name="", *args, **kwargs):
             resource_name=resource_name,
             kind="CustomResourceDefinition",
             *args,
-            **kwargs,
-        )
```
dahorak, line 482: This change was probably unintentional, right?

petr-balogh (dismissed):

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -38,7 +39,9 @@
 
 class FusionDataFoundationDeployment:
     def __init__(self):
-        self.pre_release = config.DEPLOYMENT.get("fdf_pre_release", False)
+        self.pre_release = (
+            str(config.DEPLOYMENT.get("fdf_pre_release", False)).lower() == "true"
+        )
```
dahorak: Do we need this change? If the configuration contains `true` or `false`, it should be properly parsed as boolean `True` or `False` values in python.

ocs_ci/deployment/fusion_data_foundation.py
```diff
@@ -75,7 +78,16 @@ def deploy(self):
...
         self.create_fdf_service_cr()
         self.verify_fdf_installation()
+        try:
+            wait_for_install_plan_and_approve(
+                constants.OPENSHIFT_STORAGE_NAMESPACE, 5 * 60
+            )
+        except TimeoutExpiredError:
+            logger.warning(
+                "Timeout waiting for install plan approval. Continuing execution..."
+            )
```
clacroix12: I would like to see this moved to a class method and called from here instead. I've tried to keep the `deploy` function to a very minimal scope to keep the workflow of the deployment clear.

petr-balogh (approved):

dahorak (approved):

PR #14530, opened 2026-02-27
Handle resources or modules for terraform

petr-balogh:


petr-balogh:
The issue was initially hit here:
https://url.corp.redhat.com/133f91d


This was suggested by claude code as fix, but not sure it's really just a temp issue or we will really need to update like this.

I have re-triggered job again without this PR to check if it was just really temp issue.

petr-balogh:
Verification job:
https://url.corp.redhat.com/3a94472

PR #14533, opened 2026-02-28
WIP: POC- Live cluster AI analysis hook during test failure

shylesh:
- Introduces `ocs_ci/ocs/must_gather/ai_analyzer.py` — a new opt-in module that automatically invokes an AI in parallel with must-gather log collection whenever a test fails, performing a live read-only cluster investigation (ODF/OCS operator status, StorageCluster health, pod states, events).
- The AI classifies each failure into one of three categories — **Product Bug**, **Framework Issue**, or **Infrastructure Issue** — and writes a structured per-test AI summary file to the test log directory.
- At the end of the pytest session, all per-test results are aggregated into a single HTML report (`AI_Failure_Analysis_Report.html`) with a failure-category pie chart and cluster version metadata (OCP, ODF, ACM, Ceph).
- The feature is fully opt-in via `ENV_DATA.ai_live_analysis: true` with zero overhead when disabled, and credentials under `data/` are explicitly blocked from AI access.

### DIFF
--- a/ocs_ci/ocs/defaults.py
+++ b/ocs_ci/ocs/defaults.py
@@ -177,10 +177,12 @@ AI_MCP_SERVER_REPO_URL = "https://github.com/shivamdurgbuns/ocs-ci-server.git"
 AI_MCP_SERVER_DIR = os.path.join(AI_MCP_BASE_DIR, "ocs-ci-server")
 AI_OCS_CI_REPO_URL = "https://github.com/red-hat-storage/ocs-ci.git"
 AI_OCS_CI_DIR = os.path.join(AI_MCP_BASE_DIR, "ocs-ci")
+AI_MCP_VENV_DIR = os.path.join(AI_MCP_BASE_DIR, "venv")
 AI_MCP_TEMPLATE_PATH = os.path.expanduser("~/.mcp.json")
 AI_GCP_CREDENTIALS_PATH = os.path.expanduser("~/.gcp/gcp-auth.json")
 AI_MCP_SERVER_NAME = "ocs-ci-server"
-AI_PYTHON_PATH = "python3"
+# Full path to Python 3.12+ in MCP venv (will be created if doesn't exist)
+AI_PYTHON_PATH = os.path.join(AI_MCP_BASE_DIR, "venv/bin/python")
 
 # CrushDeviceClass
 CRUSH_DEVICE_CLASS = "ssd"
--- a/ocs_ci/ocs/defaults.py
+++ b/ocs_ci/ocs/defaults.py
@@ -171,7 +171,8 @@ MUST_GATHER_UPSTREAM_TAG = "latest"
 MUST_GATHER_TIMEOUT = 3600
 
 # AI Live Analysis with MCP Server:
-AI_MCP_BASE_DIR = "/opt/claude"
+# Use ~/.claude/mcp/ as base directory (user-writable, no sudo required)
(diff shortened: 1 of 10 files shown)


ocs_ci/ocs/must_gather/ai_analyzer.py
```diff
@@ -0,0 +1,1466 @@
...
+
+#### 7. Confidence Level
+Rate your overall confidence in this analysis:
+**Confidence**: <High | Medium | Low>
+**Justification**: <brief explanation — e.g. "clear stack trace pointing to operator",
+  or "limited log data available, cluster state ambiguous">
+
+Output ONLY the summary report text. Do not include any preamble or meta-commentary.
+"""
+    return prompt
```
shivamdurgbuns, line 824: I think we are using lots of token in our [prompt](https://github.com/red-hat-storage/ocs-ci/pull/14533/changes#diff-94e89f04006da3ca04275d96d52ac021e4e2036dec83bd11ee1ad9504fe75eb1R309-R418) and as I see this prompt is getting sent on each failed test call, so I feel we can divide the prompt in two, one is the role definition that we can cache  and one is failed test details prompt. This would save us significant amount of token atleast 60+% on each fail test analysis run.

ocs_ci/ocs/must_gather/ai_analyzer.py
```diff
@@ -0,0 +1,1466 @@
...
+```
+{failure_repr}
+```
+{log_section}
+## Cluster Access (Kubeconfigs)
+The following clusters are available for investigation:
+{kubeconfig_section}
+Use the appropriate --kubeconfig flag when running oc/kubectl commands.
+
+## OCS-CI Codebase Reference
```
shivamdurgbuns, line 746: I think instead of passing through the whole codebase as context in each failed run analysis cycle it would be better to create an mcp server and attach it to our agent this will also in turn reduce the token sizes. I use a mcp server for ocs-ci locally and this has resulted in fewer token usage for me.

shylesh: ack, that should save us USDs.

ocs_ci/ocs/defaults.py
```diff
@@ -170,6 +170,20 @@
...
+# AI Live Analysis with MCP Server:
+# Use ~/.claude/mcp/ as base directory (user-writable, no sudo required)
+AI_MCP_BASE_DIR = os.path.expanduser("~/.claude/mcp")
+AI_MCP_SERVER_REPO_URL = "https://github.com/shivamdurgbuns/ocs-ci-server.git"
+AI_MCP_SERVER_DIR = os.path.join(AI_MCP_BASE_DIR, "ocs-ci-server")
+AI_OCS_CI_REPO_URL = "https://github.com/red-hat-storage/ocs-ci.git"
+AI_OCS_CI_DIR = os.path.join(AI_MCP_BASE_DIR, "ocs-ci")
+AI_MCP_VENV_DIR = os.path.join(AI_MCP_BASE_DIR, "venv")
+AI_MCP_TEMPLATE_PATH = os.path.expanduser("~/.mcp.json")
+AI_GCP_CREDENTIALS_PATH = os.path.expanduser("~/.gcp/gcp-auth.json")
```
petr-balogh, line 182: Should not we have at least some of those values configurable from config so user can use own location if needed?

petr-balogh: Ah, I see you actually have it only as default here. Then ignore :)

PR #14538, opened 2026-03-01
Fix lifecycle UI locators for PF5/PF6 compatibility

shyRozen:
OCP 4.21 uses a mix of PatternFly 5 and 6 components. PF6 wraps button text in <span class="pf-v6-c-button__text">, which breaks XPath text() matching since text() only matches direct text nodes.

- Replace text()= with normalize-space()= in button locators so they match both PF5 (direct text) and PF6 (span-wrapped text)
- Use contains(@data-ouia-component-type, 'MenuToggle') instead of hardcoding PF5/MenuToggle for the kebab menu selector
- Add page_has_loaded() wait in navigate_to_bucket_lifecycle between clicking the bucket link and the Management tab
- Remove redundant empty-list guard in test_delete_lifecycle_rule


anajain16:
Please refer to https://github.com/red-hat-storage/ocs-ci/pull/14864/changes , point 1 and 2 are already taken care in this.

PR #14560, opened 2026-03-03
RHSTOR-7671:Add runbook link verification for ODFOperatorNotUpgradeable alert

suchita-g:
Code chnages:
- Add RUNBOOK_URL_ODFOPERATORNOTUPGRADABLE in constants for the
  openshift/runbooks ODFOperatorNotUpgradeable.md URL
- Extend verify_odf_not_upgradeable to return (bool, alerts) so
  the same alert list can be reused for runbook checks
- Add verify_odf_not_upgradeable_runbook_link(alerts) to assert
  annotations.runbook_url matches the expected runbook
- In test_ceph_health_warning_blocks_upgrade and
  test_storagecluster_not_ready_blocks_upgrade: verify runbook link
  after alert check and assert with a clear failure message

### DIFF
--- a/tests/functional/upgrade/test_upgrade_precheck_ceph_health.py
+++ b/tests/functional/upgrade/test_upgrade_precheck_ceph_health.py
@@ -28,13 +28,14 @@ log = logging.getLogger(__name__)
 
 def verify_odf_not_upgradeable(threading_lock):
     """
-    This Function Verifies ODFOperatorNotUpgradeable alert using PrometheusAPI
+    Verify ODFOperatorNotUpgradeable alert using PrometheusAPI.
 
     Args:
         threading_lock: Threading lock for Prometheus API calls
 
     Returns:
-        bool: True if ODFOperatorNotUpgradeable alert found otherwise False
+        tuple: (bool, list) - True and list of alerts if alert found,
+            False and empty list otherwise.
 
     """
     log.info("Verifying ODFOperatorNotUpgradeable alert is shown")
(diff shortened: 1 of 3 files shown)


tests/functional/upgrade/test_upgrade_precheck_ceph_health.py
```diff
@@ -183,6 +223,10 @@ def test_ceph_health_warning_blocks_upgrade(self, mon_pod_down, threading_lock):
         log.info(
             f"  ODFOperatorNotUpgradeable Alert: " f"{'✓' if alert_found else '⚠'}"
         )
+        log.info(
+            f"  ODFOperatorNotUpgradeable Runbook Link: "
```
dahorak, line 227: The _f-string_ here is not necessary, as there is no variable used inside, but I see it is similarly used also on the previous lines as well.

```suggestion
            "  ODFOperatorNotUpgradeable Runbook Link: "
```

(I thought this should be caught by flake8 pre-commit-hook, but maybe not.)

suchita-g: originally        
       
       ```log.info(
            f"  ODFOperatorNotUpgradeable Runbook Link: {'✓' if runbook_ok else '⚠'}"
        )
        ```
which then formatted to this while commit        
        
```
 log.info(
            f"  ODFOperatorNotUpgradeable Runbook Link: "
            f"{'✓' if runbook_ok else '⚠'}"
        )
```

PR #14567, opened 2026-03-04
Add Logging Guide for Contributors

clacroix12:
A comprehensive contributor guide for logging in ocs-ci. This is a practical, example-focused document complete with an overview of each log level, direction based on module type, usage patterns to follow, anti-patterns to avoid, more.

Note: The information here is assuming the presence of a custom logger class with custom levels defined. A proposal for this update has been raised with the logging enhancements initiative and is currently being implemented.

### DIFF
--- a/docs/logging_guide.md
+++ b/docs/logging_guide.md
@@ -29,6 +29,63 @@ logger = logging.getLogger(__name__)
 
 The framework automatically provides a custom `OCSCILogger` with extended log levels.
 
+**Important:** The logger creation statement should be placed **immediately after import statements**, before any other code.
+
+#### Best Practice
+
+```python
+# test_example.py
+import logging
+import pytest
+from ocs_ci.framework.testlib import tier1, ManageTest
+from ocs_ci.ocs.resources.pvc import create_pvc
+
+logger = logging.getLogger(__name__)  # First line after imports
+
+
+@tier1
+class TestExample(ManageTest):
+    """Example test class"""
+
+    def test_something(self):
+        logger.info("Test starting")
+        # Test code here
+```
+
+#### Why Use `__name__`?
+
+Using `logging.getLogger(__name__)` creates a logger named after the module (e.g., `tests.functional.test_pvc` or `ocs_ci.helpers.helpers`). This:
+- Creates a hierarchical logger structure
+- Makes it easy to filter logs by module
+- Helps identify the source of log messages
+- Follows Python logging best practices
+
+#### What NOT to Do
+
+```python
+# ❌ Wrong: Using a custom string
+logger = logging.getLogger("my_logger")
+
+# ❌ Wrong: Getting the root logger
+logger = logging.getLogger()
+
+# ❌ Wrong: Creating logger inside functions
+def my_function():
+    logger = logging.getLogger(__name__)  # Creates logger on every call
+    logger.info("Message")
+
+# ❌ Wrong: Placing logger creation after other code
+import logging
+from ocs_ci.framework.testlib import tier1
+
+CONSTANT = "value"
+
+def helper_function():
+    return "data"
+
+logger = logging.getLogger(__name__)  # Should be right after imports
+```
+
 ### Available Log Levels
 
 From highest to lowest severity:
--- a/docs/logging_guide.md
+++ b/docs/logging_guide.md
@@ -471,25 +471,6 @@ def scale_deployment(deployment_name, replicas):
 
     scale(deployment_name, replicas)
     logger.info(f"Deployment '{deployment_name}' scaled successfully")
-
-# Test execution flow
-def test_multiple_pvcs(self):
-    """Test creating multiple PVCs concurrently"""
-    pvc_count = 10
-
-    logger.info(f"Starting concurrent PVC creation test (count: {pvc_count})")
-
-    logger.info("Creating PVCs...")
-    pvcs = [create_pvc(f"test-pvc-{i}", size="1Gi") for i in range(pvc_count)]
-
-    logger.info("Waiting for all PVCs to be bound...")
-    all_bound = wait_for_all_pvcs_bound(pvcs, timeout=300)
-    assert all_bound, "Not all PVCs reached Bound state"
-
-    logger.info("Cleaning up PVCs...")
-    delete_pvcs(pvcs)
-
-    logger.info("Test completed successfully")
 ```
 
 **Output:**
--- a/docs/logging_guide.md
+++ b/docs/logging_guide.md
@@ -38,9 +38,9 @@ From highest to lowest severity:
 | **CRITICAL** | 50 | `logger.critical()` | System failures, framework crashes |
 | **ERROR** | 40 | `logger.error()` | Test failures, exceptions, errors |
 | **WARNING** | 30 | `logger.warning()` | Deprecations, retries, potential issues |
-| **ASSERTION** | 27 | `logger.assertion()` | Test validations and assertions |
 | **TEST_STEP** | 25 | `logger.test_step()` | Major test workflow phases |
 | **INFO** | 20 | `logger.info()` | General progress, operations |
+| **ASSERTION** | 15 | `logger.assertion()` | Test validations and assertions |
 | **DEBUG** | 10 | `logger.debug()` | Detailed diagnostic info |
 | **AI_DATA** | 5 | `logger.ai_data()` | ML predictions, metrics, analysis |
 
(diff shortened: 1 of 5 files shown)


docs/logging_guide.md
```diff
@@ -0,0 +1,1690 @@
...
+- Pass/fail validations
+
+**Examples:**
+
+```python
+# Basic assertion logging
+def test_pvc_creation(self):
+    pvc = create_pvc(size="10Gi")
+
+    logger.assertion(
```
ebenahar: isn't it redundant? Why do we need both the log message and the assertion message?

docs/logging_guide.md
```diff
@@ -0,0 +1,1690 @@
...
+ASSERTION - IOPS check: measured=1250, minimum=1000, passed=True
+ASSERTION - operators_ready: True
+ASSERTION - storage_cluster_healthy: True
+ASSERTION - Pod count: expected=12, actual=12, match=True
+ASSERTION - Capacity usage: measured=73.5%, expected=75%, tolerance=±5%, passed=True
+```
+
+---
+
+### STEP (25) - Test Workflow Phases
```
ebenahar: how about we call it TESTSTEP?

docs/logging_guide.md
```diff
@@ -0,0 +1,1690 @@
...
+
+    logger.info("Cleaning up PVCs...")
+    delete_pvcs(pvcs)
+
+    logger.info("Test completed successfully")
+```
+
+**Output:**
+```
+INFO - Creating PVC 'test-pvc-rbd': size=10Gi, storage_class=ocs-storagecluster-ceph-rbd
```
ebenahar, line 459: Should we expect all logging within a test case to be done with STEP (or TESTSTEP as I suggested)?
If so, would be good to describe where INFO still fits

ebenahar (commented):
As discussed offline, let's consider adding a skill / standard prompt to be used directly by team members who will adjust test cases' logging to be compatible with the new log levels introduced in this PR.

docs/logging_guide.md
clacroix12: The main reason for it is that the assertion is present even in the case where we succeed. For the majority of assertions this will be the case. We will be logging that a check was performed and that it was successful. It's really only redundant in the failure case where the AssertionError is raised.

docs/logging_guide.md
clacroix12: I'm fine with TESTSTEP, or TEST_STEP to be more precise in it's purpose.

docs/logging_guide.md
clacroix12: This is covered a bit further down in the Usage by Code Type section. For test cases:
```
Tests should use STEP and ASSERTION liberally to make test flow clear
```
I would say this is a general guideline but not a hard limitation. I think we probably want to limit the amount of INFO level logging tests are doing. I think INFO should mostly come directly from the utility functions tests are calling, or prior to calling them to describe individual actions being taken as a part of a larger test step.

That being said I do think the output here is from an example that should be revisited or removed, as it somewhat contradicts the presence of the STEP level.

docs/logging_guide.md
ebenahar: so this is applicable in negative scenarios, where we expect failure?

clacroix12: This is applicable in any scenario where a test case is making an assertion. The reason for implementing this is so that we can easily identify checks that our tests are performing. Since assertions are generally important factors in a test case, I thought it would be beneficial to include what we are asserting in our logs, even when those assertions are true.

Currently if we use an assert statement, we only see something in the log if that assertion fails. By following the suggestions in the guide (log the comparison using ASSERTION level prior to the actual assert statement) we can display what is being asserted in our logs even when those assertions pass.

docs/logging_guide.md
clacroix12: The STEP level has been changed to TEST_STEP. The logger method will be `test_step` to match.

clacroix12:
Custom Logger is being added here: #14668

docs/logging_guide.md
```diff
@@ -0,0 +1,1723 @@
...
+- Critical resource exhaustion
+- Infrastructure failures requiring immediate attention
+
+**Examples:**
+
+```python
+# Framework initialization
+try:
+    initialize_framework()
+except Exception as e:
```
AviadP, line 122: i dont like this example, as we trying to avoid `except Exception`. another thing, we may want to use `logger.exception` aswell as the other logs level

docs/logging_guide.md
```diff
@@ -0,0 +1,1723 @@
...
+    assert result_a == expected, "Operation A failed"
+
+    # Continue
+    logger.info("Starting sub-operation B")
+    result_b = sub_operation_b()
+
+    logger.info("Complex operation completed successfully")
+```
+
+### Pattern: Retry Logging
```
AviadP, line 1175: 2 comments about these sections (Retry and Iter):
1. we may want to refer to TimeoutSampler, as it is main tool for managing retries in ocs-ci
2. we should guide contributors to write repeated logs in a way it will not spam the logs. worth an offline discussion in my opinion

docs/logging_guide.md
```diff
@@ -0,0 +1,1723 @@
...
+logger.assertion(f"PVC status check passed")  # Never logs if assertion fails
+```
+
+**Why?** Logging before ensures the assertion context is captured even if the assert fails.
+
+### Exception Logging
+
+Python's logging module provides `logger.exception()` which automatically logs at ERROR level and includes the full traceback. This is the preferred method for logging exceptions.
+
+#### Using logger.exception()
```
AviadP, line 1604: Agree. worth mentioning at the head of this doc as one of the logging levels

clacroix12: I made sure to note that we should use `logger.exception` when it's applicable, however exception isn't it's own log level. The exception is logged at ERROR level thus there is no section in the log levels for it.

docs/logging_guide.md
clacroix12: 1. I definitely agree we should use TimeoutSampler in some examples as it's the most common use case for this type of behavior.
2. I thought about this as well. I'll try to come up with a better example that displays how to log a subset of iterations at warning level. I do think the level is appropriate, but we definitely want to include some info on how to limit the number of identical log messages.

docs/logging_guide.md
clacroix12: I think that's a good rule for test cases and utilities that perform isolated functions. However, we have framework code that has to catch exceptions like this because we don't want uncaught errors to interrupt certain processes by crashing the execution. As this is a framework initialization example, I don't think this goes against how we handle such cases in ocs-ci today.

docs/logging_guide.md
petr-balogh: I am also hesitant about we really need to have those might actually increase a log a lot.

I am seeing many times even when we do assert and it fails because people tend to use assert on big about of data - whole yaml of some object - it in the traceback from assertion error floods our console output.

petr-balogh: I am also thinking when we will be doing changes in the logging if we can have some control over the number of lines of code which is being printed out to console output at once.  The worst really is if someone is printing some very big amount of data you can scroll in jenkins job for ages and to have some max number of lines we can print at once will be good improvement in the logging. I think that we should not go over 500-1000 lines per one log.

docs/logging_guide.md
clacroix12: I've added an example using TimeoutSampler which also displays how to limit the log messages based on the context being checked.

clacroix12:
Rebased on master and moved ASSERTION to level 15 between DEBUG (10) and INFO (20)

petr-balogh (dismissed):

dahorak (dismissed):
Overall LGTM.
I'm just thinking about one more topic which is usage of various "frames", for example like this:
```
        log.info("=" * 80)
        log.info("BACKGROUND CLUSTER VALIDATION SUMMARY")
        log.info("=" * 80)
```
(from `ocs_ci/krkn_chaos/background_cluster_validator.py`) or
```

            logger.info(
                "\n====================================================\n"
                "Starting component failure tests after each iteration\n"
                "ALL components (MDS, MGR, MON, OSD) will be tested\n"
                "===================================================="
            )
```
(from `tests/cross_functional/stress/cephfs/test_cephfs_stress_with_component_failures.py`)

I'm not a big fan of this type of highlighting, because it breaks the log flow (similarly for one line logs started with new line, like `log.info("\nOperations by Type:")`) and makes "greping"/filtering the logs more difficult (for example based on the level or module), but I'm not sure what is the point of view to this for others?

clacroix12:
> Overall LGTM. I'm just thinking about one more topic which is usage of various "frames", for example like this:
> 
> ```
>         log.info("=" * 80)
>         log.info("BACKGROUND CLUSTER VALIDATION SUMMARY")
>         log.info("=" * 80)
> ```
> 
> (from `ocs_ci/krkn_chaos/background_cluster_validator.py`) or
> 
> ```
> 
>             logger.info(
>                 "\n====================================================\n"
>                 "Starting component failure tests after each iteration\n"
>                 "ALL components (MDS, MGR, MON, OSD) will be tested\n"
>                 "===================================================="
>             )
> ```
> 
> (from `tests/cross_functional/stress/cephfs/test_cephfs_stress_with_component_failures.py`)
> 
> I'm not a big fan of this type of highlighting, because it breaks the log flow (similarly for one line logs started with new line, like `log.info("\nOperations by Type:")`) and makes "greping"/filtering the logs more difficult (for example based on the level or module), but I'm not sure what is the point of view to this for others?

I agree with you and think this is something that we should try to avoid. We can discuss this at the next logging enhancements meeting to see if anyone disagrees. In the meantime I can work on an addition to the anti-patterns section for this.

clacroix12:
Rebased on master and added a section on framing messages to the anti-patterns. @dahorak FYI

petr-balogh (approved):

dahorak (approved):

Issue #14597, opened 2026-03-05
[Improvement]: In the test "test_patch_and_verify_mon_healthcheck" simplified the mon healthcheck timeout pattern search


yitzhak12:
### Feature Description
What do you want to be added or improved?
Simplify the mon healthcheck timeout pattern search.

### Use Case
Why is this needed?
Currently, some tests fail when searching for the mon healthcheck timeout pattern.

### Alternatives Considered
Any other solutions you thought of?
No.

### Additional Context
Screenshots, references, etc.
https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/65270/testReport/

PR #14623, opened 2026-03-09
Fix mon healthcheck log pattern to be more resilient to log format changes

yitzhak12:
fix https://github.com/red-hat-storage/ocs-ci/issues/14597


yitzhak12:
The error above is related to the drain node failure, and not to the test.

PR #14676, opened 2026-03-13
Fix NFS mount failure on IBM Cloud when client VM is in a different VPC

amr1ta:
## Problem
   On IBM Cloud, when the `rook-ceph-nfs-my-nfs-load-balancer` LoadBalancer service is recreated (e.g. after an ODF redeployment), IBM Cloud assigns a new VPC LB hostname (`*.lb.appdomain.cloud`) with new
   backend IPs. The NFS client VM is in a different VPC from the OpenShift cluster, so:

   - IBM Cloud VPC LB hostnames are only resolvable from within the same VPC — the client VM's DNS servers cannot resolve the new hostname
   - Any previous `/etc/hosts` entry pointing to the old LB IP becomes stale
   - NFS mounts hang indefinitely or fail with `Name or service not known`

   ## Fix

   `update_etc_hosts_on_nfs_client()` resolves the LB hostname by running `getent hosts` on the node where the NFS pod runs (via `oc debug node`), then removes the stale `/etc/hosts` entry and writes the fresh
   IP mapping on the client VM. This runs automatically on every connection (including after VM reboots), so it stays correct across redeployments.

   The fix is conditional — it only triggers when `hostname_add` is a hostname (IBM Cloud), not a raw IP (AWS/other platforms).

### DIFF
--- a/ocs_ci/utility/nfs_utils.py
+++ b/ocs_ci/utility/nfs_utils.py
@@ -210,7 +210,7 @@ def update_etc_hosts_on_nfs_client(con, hostname):
     Resolve an NFS LB hostname from within the cluster and update /etc/hosts
     on the NFS client VM.
 
-    IBM Cloud VPC Load Balancer hostnames (*.lb.appdomain.cloud) are only
+    IBM Cloud VPC Load Balancer hostnames (``*.lb.appdomain.cloud``) are only
     resolvable from within the same VPC. When the NFS client VM is in a
     different VPC, DNS resolution fails and mounts hang. This function resolves
     the hostname by exec-ing on the node where the NFS pod runs, then writes


amr1ta:
validation job: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/65590/console

tests/functional/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
```diff
@@ -335,7 +340,11 @@ def __make_connection():
                 private_key=self.nfs_client_private_key,
             )
 
-        return __make_connection()
+        con = __make_connection()
+        hostname_add = getattr(self, "hostname_add", None)
+        if hostname_add and not hostname_add[0].isdigit():
```
suchita-g: Use proper IP validation instead of isdigit()
```suggestion
        try:
            ipaddress.ip_address(hostname_add)
            is_ip = True
        except ValueError:
            is_ip = False
        if hostname_add and not is_ip:
```

tests/functional/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
suchita-g, line 7:
```suggestion
import ipaddress

```

ocs_ci/utility/nfs_utils.py
```diff
@@ -178,38 +177,128 @@ def create_nfs_load_balancer_service(
...
+        sleep=15,
+        func=exec_cmd,
+        command=(
+            f"oc debug node/{nfs_node} --to-namespace=default "
+            f"-- chroot /host getent hosts {hostname}"
+        ),
+        ignore_error=True,
+    ):
+        if sample and sample.stdout:
+            lb_ips = [
```
suchita-g: Add try-except for decode() operations
```suggestion
            try:
                lb_ips = [line.split()[0] for line in sample.strip().decode().strip().split("\n")]
            except (UnicodeDecodeError, AttributeError) as e:
                log.warning(f"Failed to decode output: {e}")
                continue
```

tests/functional/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
amr1ta: addressed in latest commit

tests/functional/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
amr1ta: addressed in latest commit

ocs_ci/utility/nfs_utils.py
amr1ta: addressed in latest commit

ocs_ci/utility/nfs_utils.py
```diff
@@ -178,38 +177,131 @@ def create_nfs_load_balancer_service(
...
+    )
+    if not nfs_pods:
+        log.warning("No NFS pods found, skipping /etc/hosts update on NFS client VM")
+        return
+
+    nfs_node = nfs_pods[0].get()["spec"]["nodeName"]
+    log.info("Resolving %s from cluster node %s", hostname, nfs_node)
+
+    lb_ips = []
+    for sample in TimeoutSampler(
```
AviadP: exec_cmd argument is `cmd` instead of `command` - that might cause typeError

ocs_ci/utility/nfs_utils.py
```diff
@@ -178,38 +177,131 @@ def create_nfs_load_balancer_service(
...
+                    for line in sample.stdout.decode().strip().splitlines()
+                    if line.strip()
+                ]
+            except (UnicodeDecodeError, AttributeError) as e:
+                log.warning("Failed to decode getent output: %s", e)
+        if lb_ips:
+            break
+        log.info("Could not resolve %s yet, retrying in 15s...", hostname)
+
+    if not lb_ips:
```
AviadP: im not sure it will ever reach this point.` if lb_ips` it will get out of timoutSampler, else it will fail there

ocs_ci/utility/nfs_utils.py
amr1ta: addressed in latest commit

ocs_ci/utility/nfs_utils.py
amr1ta: addressed in latest commit

amr1ta:
validation job: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/65760/testReport/

PR #14677, opened 2026-03-13
Added Krknctl random scenario execution

paraggit:
PR: Add test automation for krknctl chaos scenarios

- Krknctl random chaos (test_krknctl_random_chaos_run): Runs krknctl random chaos with a generated plan and workload validation; parametrized over max_parallel 2, 3, and 4.
- Krknctl service disruption (test_random_service_disruption): Runs comprehensive service disruption (application-outages, pod-scenarios, container-scenarios, service-disruption-scenarios) for all app labels in a single plan; parametrized over max_parallel 2, 3, and 4.

Both tests assert exit criteria: no Ceph crash and cluster not in HEALTH_ERR after chaos.

### DIFF
--- a/ocs_ci/krkn_chaos/background_cluster_operations.py
+++ b/ocs_ci/krkn_chaos/background_cluster_operations.py
@@ -31,6 +31,8 @@ from ocs_ci.ocs import constants, ocp
 from ocs_ci.ocs.resources import pod as pod_helpers
 from ocs_ci.ocs.resources import pvc as pvc_helpers
 from ocs_ci.ocs.resources import job as job_helpers
+from ocs_ci.ocs.resources.ocs import OCS
+from ocs_ci.ocs.resources.pvc import PVC
 from ocs_ci.ocs import node as node_helpers
 from ocs_ci.ocs.exceptions import (
     UnexpectedBehaviour,
@@ -908,7 +910,7 @@ class BackgroundClusterOperations:
             num_pvcs = min(3, len(self.workloads))
             log.info(f"Will check up to {num_pvcs} workloads for PVCs")
 
-            # Collect PVCs from workloads
+            # Collect PVCs from workloads (coerce OCS → PVC for snapshot APIs)
             workload_pvcs = []
             for idx, workload in enumerate(self.workloads[:num_pvcs]):
                 log.debug(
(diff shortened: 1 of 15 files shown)


ocs_ci/krkn_chaos/background_cluster_operations.py
```diff
@@ -1069,6 +1079,39 @@ def _longevity_operations(self):
     # Helper Methods
     # ==========================================================================
 
+    def _coerce_workload_pvc_to_pvc_instance(self, pvc_obj):
```
ebenahar, line 1082: it's unclear why this function is needed

ocs_ci/krkn_chaos/krkn_helpers.py
```diff
@@ -2135,6 +2151,58 @@ def check_ceph_crashes(self, component_label, chaos_type="chaos"):
             return True, ""
 
 
+def krkn_exit_criteria(chaos_context="krkn chaos", namespace=None):
+    """
+    Generic exit criteria: assert no Ceph crash and cluster not in HEALTH_ERR.
```
ebenahar: what about NooBaa status?

ocs_ci/krkn_chaos/krknclt_helper.py
```diff
@@ -0,0 +1,971 @@
+"""
```
ebenahar, line 1: shouldn't these functionalities be added within ocs_ci/ocs/platform_nodes.py?

ocs_ci/krkn_chaos/background_cluster_operations.py
paraggit: Workload PVC references are sometimes a generic OCS object that still represents the same PersistentVolumeClaim in the cluster. Snapshot and related helpers expect a real PVC instance so methods like create_snapshot exist. This helper turns those generic references into PVC when needed, and returns the object unchanged if it’s already a PVC.

ocs_ci/krkn_chaos/krkn_helpers.py
paraggit: Added NooBaa Health Check condition

ocs_ci/krkn_chaos/krknclt_helper.py
paraggit: This file is for krknctl plan generation and running chaos. It does not contain node-operation logic so these functions are not a good fit for platform_nodes.py

ocs_ci/krkn_chaos/background_cluster_operations.py
```diff
@@ -1069,6 +1079,39 @@ def _longevity_operations(self):
...
+            return pvc_obj
+        if isinstance(pvc_obj, OCS) and getattr(pvc_obj, "kind", None) == (
+            "PersistentVolumeClaim"
+        ):
+            try:
+                if hasattr(pvc_obj, "reload"):
+                    pvc_obj.reload()
+                return PVC(**pvc_obj.data)
+            except Exception as err:
+                log.warning(
```
fbalak, line 1102: Wouldn't it be better to raise the exception to not miss any relevant pvc?

ocs_ci/krkn_chaos/krkn_helpers.py
```diff
@@ -2135,6 +2155,99 @@ def check_ceph_crashes(self, component_label, chaos_type="chaos"):
             return True, ""
 
 
+def _krkn_should_run_noobaa_health_check():
```
fbalak, line 2158: This could be a generic reusable check - something like storagecluster_noobaa_health_available. Or even a returned value with specific platform exception if the condition is not met.

ocs_ci/krkn_chaos/background_cluster_operations.py
paraggit: In the background cluster operations (PVC create → restore → delete) are running. Hence we do not expect every PVC to be present in this output so we are logging this as a warning only.

ocs_ci/krkn_chaos/krkn_helpers.py
paraggit: This Comment I will incorporate in the Next Noobaa Specific chaos  scenario 
PR : https://github.com/red-hat-storage/ocs-ci/pull/14286

PR #14679, opened 2026-03-13
[4.21 system test] Test PV Key Rotation with all combinations

posharma1:
Test validates automatic key rotation for encrypted RBD PVCs with Vault KMS across multiple storage types and access modes.

## Summary by CodeRabbit

* **Tests**
  * Added system-level coverage for Vault-KMS key rotation of encrypted RBD PVCs across RBD, CephFS, and NFS storage.
  * Added reusable test setup for provisioning PVCs and pods across multiple storage backends, including cleanup of temporary resources.
  * Added coverage for creating PVC clones until cluster-full alerts are detected, including pending-clone handling.
  * Expanded validation across FIO workloads, snapshots, clones, resizing, restores, and Ceph full-ratio recovery.
* **Reliability**
  * Increased the wait period for NFS CSI provisioner pods during setup.

### DIFF
--- a/tests/cross_functional/conftest.py
+++ b/tests/cross_functional/conftest.py
@@ -8,6 +8,7 @@ from concurrent.futures import ThreadPoolExecutor
 from threading import Event
 from subprocess import TimeoutExpired
 
+from ocs_ci.ocs.cluster import change_ceph_full_ratio
 from ocs_ci.helpers import helpers
 from ocs_ci.helpers.odf_cli import odf_cli_setup_helper
 from ocs_ci.helpers.helpers import (
@@ -27,7 +28,6 @@ from ocs_ci.helpers.e2e_helpers import (
 from ocs_ci.ocs import constants
 from ocs_ci.ocs import ocp
 from ocs_ci.utility import templating
-from ocs_ci.utility.kms import is_kms_enabled
 from ocs_ci.ocs.amq import AMQ
 from ocs_ci.ocs.bucket_utils import (
     compare_object_checksums_between_bucket_and_local,
@@ -2977,6 +2977,7 @@ def vdbench_mixed_workload_config():
 def create_multiple_storage_pvcs_pods(
     pvc_factory,
     deployment_pod_factory,
+    request,
 ):
     """
     Fixture factory to create multiple PVCs and deployment pods across different storage types.
@@ -2984,40 +2985,43 @@ def create_multiple_storage_pvcs_pods(
         function: Factory function that accepts parameters and returns created resources
     """
 
-    def factory(proj_obj, sc_obj, skip_storage_types=None, total_pvcs=10):
+    def teardown():
+
+        change_ceph_full_ratio(85)
+
+    request.addfinalizer(teardown)
+
+    def factory(
+        proj_obj, sc_obj, skip_storage_types=None, total_pvcs=10, pvc_name_prefix=""
+    ):
         """
         Args:
             proj_obj: Project object for namespace
             sc_obj: Encrypted storage class object (for encrypted RBD)
             skip_storage_types (list): List of storage types to skip.
             total_pvcs (int): Total number of encrypted PVCs to create (default: 10)
+            pvc_name_prefix (str): Prefix to add to PVC names to avoid conflicts (default: "")
         """
 
-        # Initialize skip_storage_types if None
         if skip_storage_types is None:
             skip_storage_types = []
 
-        # Validate skip_storage_types
         valid_storage_types = ["encrypted_rbd", "non_encrypted_rbd", "cephfs", "nfs"]
         if not isinstance(skip_storage_types, list):
             raise ValueError(
                 f"skip_storage_types must be a list. Valid values: {valid_storage_types}"
             )
 
-        # Initialize all resource lists
         pvc_objs, pod_objs = [], []
         non_enc_pvc_objs, non_enc_pod_objs, non_enc_combinations = [], [], []
         cephfs_pvc_objs, cephfs_pod_objs = [], []
         nfs_pvc_objs, nfs_pod_objs = [], []
 
-        # Step 1: Create encrypted RBD PVCs and pods
-
+        # Create encrypted RBD PVCs and pods
         if "encrypted_rbd" not in skip_storage_types:
             logger.info(
                 f"Step 1: Creating {total_pvcs} encrypted RBD PVCs and deployment pods"
             )
-
-            # Prepare PVC combinations for encrypted RBD
             enc_base_combinations = [
                 (constants.VOLUME_MODE_FILESYSTEM, constants.ACCESS_MODE_RWO),
                 (constants.VOLUME_MODE_BLOCK, constants.ACCESS_MODE_RWO),
@@ -3031,7 +3035,6 @@ def create_multiple_storage_pvcs_pods(
             ]
 
             for idx, pvc_config in enumerate(pvc_combinations, start=1):
-                logger.info(f"\nCreating encrypted PVC {idx}/{total_pvcs}:")
                 pvc_obj = pvc_factory(
                     interface=constants.CEPHBLOCKPOOL,
                     project=proj_obj,
@@ -3047,15 +3050,10 @@ def create_multiple_storage_pvcs_pods(
                 is_block_volume = (
                     pvc_config["volume_mode"] == constants.VOLUME_MODE_BLOCK
                 )
-                sa_obj = helpers.create_serviceaccount(pvc_obj.project.namespace)
-                helpers.add_scc_policy(
-                    sa_name=sa_obj.name, namespace=pvc_obj.project.namespace
-                )
                 pod_obj = deployment_pod_factory(
                     interface=constants.CEPHBLOCKPOOL,
                     pvc=pvc_obj,
                     raw_block_pv=is_block_volume,
-                    sa_obj=sa_obj,
                 )
                 pod_objs.append(pod_obj)
 
@@ -3063,14 +3061,13 @@ def create_multiple_storage_pvcs_pods(
                 f" All {len(pod_objs)} encrypted RBD deployments created and running"
             )
 
-        # Step 2: Create non-encrypted RBD PVCs and pods
+        # Create non encrypted RBD PVCs and pods
         if "non_encrypted_rbd" not in skip_storage_types:
-            logger.info("Step 2: Creating non-encrypted RBD PVCs and deployment pods")
+            logger.info(" Creating non-encrypted RBD PVCs and deployment pods")
 
             non_enc_sc_obj = helpers.default_storage_class(
                 interface_type=constants.CEPHBLOCKPOOL
             )
-            # Non-encrypted combinations: Filesystem(RWO, RWOP) + Block(RWO, RWX, RWOP) = 5 base
             non_enc_volume_modes = [
                 constants.VOLUME_MODE_FILESYSTEM,
                 constants.VOLUME_MODE_FILESYSTEM,
@@ -3085,16 +3082,14 @@ def create_multiple_storage_pvcs_pods(
                 constants.ACCESS_MODE_RWX,
                 constants.ACCESS_MODE_RWOP,
             ]
-            total_non_enc_pvcs = 10
             non_enc_combinations = [
                 {"volume_mode": vm, "access_mode": am, "size": 1}
                 for vm, am in itertools.islice(
                     itertools.cycle(zip(non_enc_volume_modes, non_enc_access_modes)),
-                    total_non_enc_pvcs,
+                    total_pvcs,
                 )
             ]
             for idx, pvc_config in enumerate(non_enc_combinations, start=1):
-                logger.info(f"Creating non-encrypted PVC {idx}/{total_non_enc_pvcs}: ")
                 non_enc_pvc_obj = pvc_factory(
                     interface=constants.CEPHBLOCKPOOL,
                     project=proj_obj,
@@ -3107,40 +3102,30 @@ def create_multiple_storage_pvcs_pods(
                 non_enc_pvc_objs.append(non_enc_pvc_obj)
 
                 is_block = pvc_config["volume_mode"] == constants.VOLUME_MODE_BLOCK
-                sa_obj = helpers.create_serviceaccount(
-                    non_enc_pvc_obj.project.namespace
-                )
-                helpers.add_scc_policy(
-                    sa_name=sa_obj.name, namespace=non_enc_pvc_obj.project.namespace
-                )
                 non_enc_pod_obj = deployment_pod_factory(
                     interface=constants.CEPHBLOCKPOOL,
                     pvc=non_enc_pvc_obj,
                     raw_block_pv=is_block,
-                    sa_obj=sa_obj,
                 )
                 non_enc_pod_objs.append(non_enc_pod_obj)
             logger.info(
                 f" All {len(non_enc_pod_objs)} non-encrypted deployments created and running"
             )
 
-        # Step 3: Create CephFS PVCs and pods
+        # Create cephfs PVCs and pods
         if "cephfs" not in skip_storage_types:
-            logger.info("Step 3: Creating CephFS PVCs and deployment pods")
+            logger.info("Creating CephFS PVCs and deployment pods")
 
             cephfs_access_modes = [
                 constants.ACCESS_MODE_RWO,
                 constants.ACCESS_MODE_RWX,
                 constants.ACCESS_MODE_RWOP,
             ]
-            total_cephfs_pvcs = 10
-
             cephfs_sc_obj = helpers.default_storage_class(
                 interface_type=constants.CEPHFILESYSTEM
             )
-            for idx in range(1, total_cephfs_pvcs + 1):
+            for idx in range(1, total_pvcs + 1):
                 access_mode = cephfs_access_modes[(idx - 1) % len(cephfs_access_modes)]
-                logger.info(f"Creating CephFS PVC {idx}/{total_cephfs_pvcs}: ")
                 cephfs_pvc_obj = pvc_factory(
                     interface=constants.CEPHFILESYSTEM,
                     project=proj_obj,
@@ -3152,15 +3137,10 @@ def create_multiple_storage_pvcs_pods(
                 )
                 cephfs_pvc_objs.append(cephfs_pvc_obj)
 
-                sa_obj = helpers.create_serviceaccount(cephfs_pvc_obj.project.namespace)
-                helpers.add_scc_policy(
-                    sa_name=sa_obj.name, namespace=cephfs_pvc_obj.project.namespace
-                )
                 cephfs_pod_obj = deployment_pod_factory(
                     interface=constants.CEPHFILESYSTEM,
                     pvc=cephfs_pvc_obj,
                     raw_block_pv=False,
-                    sa_obj=sa_obj,
                 )
                 cephfs_pod_objs.append(cephfs_pod_obj)
 
@@ -3168,21 +3148,20 @@ def create_multiple_storage_pvcs_pods(
                 f" All {len(cephfs_pod_objs)} CephFS deployments created and running"
             )
 
-        # Step 4: Create NFS PVCs and pods
+        # Create nfs PVCs and pods
         if "nfs" not in skip_storage_types:
-            logger.info("Step 4: Creating NFS in-cluster PVCs and deployment pods")
+            logger.info(" Creating NFS in-cluster PVCs and deployment pods")
 
             nfs_access_modes = [constants.ACCESS_MODE_RWO, constants.ACCESS_MODE_RWX]
-            total_nfs_pvcs = 10
             nfs_sc_name = constants.NFS_STORAGECLASS_NAME
             nfs_pvc_objs, nfs_pod_objs = [], []
 
-            for idx in range(1, total_nfs_pvcs + 1):
+            for idx in range(1, total_pvcs + 1):
                 access_mode = nfs_access_modes[(idx - 1) % len(nfs_access_modes)]
                 nfs_pvc_obj = helpers.create_pvc(
                     sc_name=nfs_sc_name,
                     namespace=proj_obj.namespace,
-                    pvc_name=f"nfs-pvc-{idx}",
+                    pvc_name=f"{pvc_name_prefix}nfs-pvc-{idx}",
                     size="1Gi",
                     do_reload=True,
                     access_mode=access_mode,
@@ -3190,8 +3169,7 @@ def create_multiple_storage_pvcs_pods(
                 )
                 nfs_pvc_objs.append(nfs_pvc_obj)
 
-                # Create deployment for NFS pod
-                deployment_name = f"nfs-test-pod-{idx}"
+                deployment_name = f"{pvc_name_prefix}nfs-test-pod-{idx}"
                 deployment_template = templating.load_yaml(constants.NFS_APP_POD_YAML)
                 deployment_data = copy.deepcopy(deployment_template)
 
@@ -3231,18 +3209,28 @@ def create_multiple_storage_pvcs_pods(
                 f" All {len(nfs_pod_objs)} NFS deployment pods created and running"
             )
 
-        # Return all created resources
+        # Return resources in a nested structure for better organization
         resources = {
-            "pvc_objs": pvc_objs,
-            "pod_objs": pod_objs,
-            "pvc_combinations": pvc_combinations,
-            "non_enc_pvc_objs": non_enc_pvc_objs,
-            "non_enc_pod_objs": non_enc_pod_objs,
-            "non_enc_combinations": non_enc_combinations,
-            "cephfs_pvc_objs": cephfs_pvc_objs,
-            "cephfs_pod_objs": cephfs_pod_objs,
-            "nfs_pvc_objs": nfs_pvc_objs,
-            "nfs_pod_objs": nfs_pod_objs,
+            "encrypted_rbd": {
+                "pvcs": pvc_objs,
+                "pods": pod_objs,
+                "combinations": pvc_combinations,
+            },
+            "non_encrypted_rbd": {
+                "pvcs": non_enc_pvc_objs,
+                "pods": non_enc_pod_objs,
+                "combinations": non_enc_combinations,
+            },
+            "cephfs": {
+                "pvcs": cephfs_pvc_objs,
+                "pods": cephfs_pod_objs,
+            },
+            "nfs": {
+                "pvcs": nfs_pvc_objs,
+                "pods": nfs_pod_objs,
+            },
+            "all_pvcs": pvc_objs + non_enc_pvc_objs + cephfs_pvc_objs + nfs_pvc_objs,
+            "all_pods": pod_objs + non_enc_pod_objs + cephfs_pod_objs + nfs_pod_objs,
         }
 
         return resources
--- a/tests/cross_functional/conftest.py
+++ b/tests/cross_functional/conftest.py
@@ -2,10 +2,13 @@ import logging
 import boto3
 import pytest
 import time
+import itertools
+import copy
 from concurrent.futures import ThreadPoolExecutor
 from threading import Event
 from subprocess import TimeoutExpired
 
+from ocs_ci.helpers import helpers
 from ocs_ci.helpers.odf_cli import odf_cli_setup_helper
 from ocs_ci.helpers.helpers import (
     run_cmd_verify_cli_output,
@@ -22,6 +25,9 @@ from ocs_ci.helpers.e2e_helpers import (
     validate_mcg_nsfs_feature,
 )
 from ocs_ci.ocs import constants
+from ocs_ci.ocs import ocp
+from ocs_ci.utility import templating
+from ocs_ci.utility.kms import is_kms_enabled
 from ocs_ci.ocs.amq import AMQ
 from ocs_ci.ocs.bucket_utils import (
     compare_object_checksums_between_bucket_and_local,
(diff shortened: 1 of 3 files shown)


tests/cross_functional/conftest.py
```diff
@@ -3168,3 +3173,280 @@ def _factory(
...
+                pvc_objs.append(pvc_obj)
+
+                # Create deployment pod
+                is_block_volume = (
+                    pvc_config["volume_mode"] == constants.VOLUME_MODE_BLOCK
+                )
+                sa_obj = helpers.create_serviceaccount(pvc_obj.project.namespace)
+                helpers.add_scc_policy(
+                    sa_name=sa_obj.name, namespace=pvc_obj.project.namespace
+                )
```
PrasadDesala: deployment_pod_factory already takes care of creating the service account creation. It can be removed here

tests/cross_functional/conftest.py
```diff
@@ -3168,3 +3173,280 @@ def _factory(
...
+                non_enc_pod_obj = deployment_pod_factory(
+                    interface=constants.CEPHBLOCKPOOL,
+                    pvc=non_enc_pvc_obj,
+                    raw_block_pv=is_block,
+                    sa_obj=sa_obj,
+                )
+                non_enc_pod_objs.append(non_enc_pod_obj)
+            logger.info(
+                f" All {len(non_enc_pod_objs)} non-encrypted deployments created and running"
+            )
```
PrasadDesala: can we use multi_deployment_pods fixture?

tests/cross_functional/conftest.py
```diff
@@ -3168,3 +3173,280 @@ def _factory(
...
+                    namespace=proj_obj.namespace,
+                    pvc_name=f"nfs-pvc-{idx}",
+                    size="1Gi",
+                    do_reload=True,
+                    access_mode=access_mode,
+                    volume_mode=constants.VOLUME_MODE_FILESYSTEM,
+                )
+                nfs_pvc_objs.append(nfs_pvc_obj)
+
+                # Create deployment for NFS pod
```
PrasadDesala: we cant use pod_factory or create_pod?

tests/cross_functional/conftest.py
```diff
@@ -3168,3 +3173,280 @@ def _factory(
         return cfg
 
     return _factory
+
+
+@pytest.fixture(scope="function")
+def create_multiple_storage_pvcs_pods(
```
PrasadDesala, line 2982: The fixture creates resources but has no finalizer to clean them up

tests/cross_functional/conftest.py
```diff
@@ -3168,3 +3173,280 @@ def _factory(
...
+            raise ValueError(
+                f"skip_storage_types must be a list. Valid values: {valid_storage_types}"
+            )
+
+        # Initialize all resource lists
+        pvc_objs, pod_objs = [], []
+        non_enc_pvc_objs, non_enc_pod_objs, non_enc_combinations = [], [], []
+        cephfs_pvc_objs, cephfs_pod_objs = [], []
+        nfs_pvc_objs, nfs_pod_objs = [], []
+
```
PrasadDesala, line 3052: pvc_combinations is not initialized but is returned in the dictionary

tests/cross_functional/conftest.py
```diff
@@ -3168,3 +3173,280 @@ def _factory(
...
+                )
+                nfs_pod_obj = nfs_pod_list[0]
+                nfs_pod_objs.append(nfs_pod_obj)
+
+            logger.info(
+                f" All {len(nfs_pod_objs)} NFS deployment pods created and running"
+            )
+
+        # Return all created resources
+        resources = {
```
PrasadDesala, line 3251: I see some issues with this way,
1) Flat structure makes it hard to group related resources
2) Combinations are mixed with actual objects
3) No clear separation between storage types

Should we consider below structure? 

```
resources = {
    "encrypted_rbd": {
        "pvcs": pvc_objs,
        "pods": pod_objs,
        "combinations": pvc_combinations,
    },
    "non_encrypted_rbd": {
        "pvcs": non_enc_pvc_objs,
        "pods": non_enc_pod_objs,
        "combinations": non_enc_combinations,
    },
    "cephfs": {
        "pvcs": cephfs_pvc_objs,
        "pods": cephfs_pod_objs,
    },
    "nfs": {
        "pvcs": nfs_pvc_objs,
        "pods": nfs_pod_objs,
    },
    "all_pvcs": pvc_objs + non_enc_pvc_objs + cephfs_pvc_objs + nfs_pvc_objs,
    "all_pods": pod_objs + non_enc_pod_objs + cephfs_pod_objs + nfs_pod_objs,
}
```

tests/cross_functional/conftest.py
```diff
@@ -3168,3 +3173,280 @@ def _factory(
...
+                    raw_block_pv=False,
+                    sa_obj=sa_obj,
+                )
+                cephfs_pod_objs.append(cephfs_pod_obj)
+
+            logger.info(
+                f" All {len(cephfs_pod_objs)} CephFS deployments created and running"
+            )
+
+        # Step 4: Create NFS PVCs and pods
```
PrasadDesala: these step comments can be removed

tests/cross_functional/conftest.py
```diff
@@ -3168,3 +3173,280 @@ def _factory(
...
+    pvc_factory,
+    deployment_pod_factory,
+):
+    """
+    Fixture factory to create multiple PVCs and deployment pods across different storage types.
+    Returns:
+        function: Factory function that accepts parameters and returns created resources
+    """
+
+    def factory(proj_obj, sc_obj, skip_storage_types=None, total_pvcs=10):
```
PrasadDesala: would the same number of PVCs gets created for the individual storage types and varieties or it is the total number of PVCs this fixture creates?

tests/cross_functional/conftest.py
```diff
@@ -3168,3 +3173,280 @@ def _factory(
...
+
+
+@pytest.fixture(scope="function")
+def create_multiple_storage_pvcs_pods(
+    pvc_factory,
+    deployment_pod_factory,
+):
+    """
+    Fixture factory to create multiple PVCs and deployment pods across different storage types.
+    Returns:
```
PrasadDesala, line 2989:
```suggestion
    
    Returns:
```

tests/cross_functional/conftest.py
```diff
@@ -3168,3 +3173,280 @@ def _factory(
...
+    def factory(proj_obj, sc_obj, skip_storage_types=None, total_pvcs=10):
+        """
+        Args:
+            proj_obj: Project object for namespace
+            sc_obj: Encrypted storage class object (for encrypted RBD)
+            skip_storage_types (list): List of storage types to skip.
+            total_pvcs (int): Total number of encrypted PVCs to create (default: 10)
+        """
+
+        # Initialize skip_storage_types if None
```
PrasadDesala: these # comments can be removed

tests/cross_functional/conftest.py
```diff
@@ -3168,3 +3173,280 @@ def _factory(
...
+        pvc_objs, pod_objs = [], []
+        non_enc_pvc_objs, non_enc_pod_objs, non_enc_combinations = [], [], []
+        cephfs_pvc_objs, cephfs_pod_objs = [], []
+        nfs_pvc_objs, nfs_pod_objs = [], []
+
+        # Step 1: Create encrypted RBD PVCs and pods
+
+        if "encrypted_rbd" not in skip_storage_types:
+            logger.info(
+                f"Step 1: Creating {total_pvcs} encrypted RBD PVCs and deployment pods"
```
PrasadDesala, line 3056:
```suggestion
                f"Creating {total_pvcs} encrypted RBD PVCs and deployment pods"
```

tests/cross_functional/conftest.py
posharma1: deletion of all resources are taken care by factory fixtures and nfs resources cleanup is separately handled at the end of test case

tests/cross_functional/conftest.py
posharma1: can you elaborate this change

tests/cross_functional/conftest.py
posharma1: same no of pvc will be created for all storage_types, total count for each type will be the value assigned to variable total_pvcs

tests/cross_functional/conftest.py
posharma1: done

tests/cross_functional/conftest.py
posharma1: its initialised in create encrypted RBD PVCs and pods block

tests/cross_functional/conftest.py
posharma1: done

tests/cross_functional/conftest.py
posharma1: done

tests/cross_functional/conftest.py
posharma1: since we are looping in non_enc_volume_modes and non_enc_access_modes combinations so at a time single pod will be created for a particular combination

tests/cross_functional/conftest.py
posharma1: done

tests/cross_functional/conftest.py
posharma1: no since create_pod does not have check added for  type NFS it won't create that

tests/cross_functional/conftest.py
posharma1: updated code with this

dgaikwad:
Generated PR review from cursor


## PR review


### Summary

Adds a 4.21 system test for RBD encrypted PVC key rotation plus a large fixture `create_multiple_storage_pvcs_pods` (encrypted RBD, non-encrypted RBD, CephFS, NFS). Direction is good for cluster-wide key rotation coverage; the fixture needs hardening before merge.

**Verdict:** **Request changes**

---

### What’s working well

- Mixing encrypted RBD with CephFS / NFS / plain RBD matches a realistic system scenario.
- Encrypted RBD combinations (FS vs block, access modes) give useful matrix coverage without exploding PVC count.
- Single factory entry point is easier to reuse than duplicated setup in the test.

---

### Must fix

**1. Teardown / finalizer**  
The factory creates many PVCs and deployments but doesn’t clean them up. Failed runs can leave collisions (`nfs-pvc-1`, etc.) and quota noise. Please add a `request` finalizer (or equivalent) to delete resources in order (workloads → PVCs), or use a dedicated namespace the test always tears down.

**2. Return dict + `pvc_combinations`**  
If encrypted RBD is skipped, `pvc_combinations` may never be defined but still returned → risk of `NameError`. Initialize all lists / combination vars before branches, or build the return object only from executed paths.

**3. Duplicate service account**  
`deployment_pod_factory` already handles SA creation in many flows. Manual `create_serviceaccount` + `add_scc_policy` before every `deployment_pod_factory` call duplicates that unless block/NFS truly need a custom SA—please align with existing factory behavior or document the exception.

---

### Should fix / improve

**4. Return structure**  
A flat dict (`pvc_objs`, `non_enc_pvc_objs`, …) is easy to misuse. A nested shape per storage type (`encrypted_rbd: {pvcs, pods, combinations}`, same for others, plus optional `all_pvcs` / `all_pods`) would be clearer for the key rotation test.

**5. Docstring**  
Clarify that `total_pvcs` applies to **encrypted RBD count** only; other tiers use fixed counts (e.g. 10). Document return keys and types.

**6. NFS vs other backends**  
RBD/CephFS use `deployment_pod_factory`; NFS uses hand-rolled `NFS_APP_POD_YAML`. Fine if required—add a short comment why `pod_factory` / shared patterns aren’t used.

**7. `multi_deployment_pods`**  
If non-encrypted RBD can be expressed via `multi_deployment_pods`, that would reduce duplication; if not, a one-line comment is enough.

---

### Nice to have

- Parameterize per-tier counts instead of hardcoded `10` everywhere except encrypted RBD.
- Trim redundant step comments if logs already state the same.

---

### Merge checklist

- [ ] Cleanup / finalizer for created resources
- [ ] Safe return contract when `skip_storage_types` is used
- [ ] SA usage consistent with `deployment_pod_factory`
- [ ] Docstring updated (`total_pvcs` scope + return shape)
- [ ] CI green on a representative platform with KMS as required

Happy to re-review after updates.

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,384 @@
...
+            """
+            rotation_results = []
+            for idx, pvc_obj in enumerate(pvc_objs, start=1):
+                log.info(
+                    f"\nVerifying key rotation for PVC {idx}/{len(pvc_objs)}: {pvc_obj.name}"
+                )
+
+                # Get PV volume handle name
+                volume_handle = pvc_obj.get_pv_volume_handle_name
+                log.info(f"  - PV Volume Handle: {volume_handle}")
```
dgaikwad:
```suggestion
                log.info(f"PV Volume Handle: {volume_handle}")
```

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,384 @@
...
+                f"All {len(nfs_pod_objs)} NFS deployment pods deleted successfully"
+            )
+
+        if nfs_pvc_objs:
+            for nfs_pvc_obj in nfs_pvc_objs:
+                nfs_pvc_obj.delete()
+                nfs_pvc_obj.ocp.wait_for_delete(
+                    resource_name=nfs_pvc_obj.name, timeout=180
+                )
+            log.info(f" All {len(nfs_pvc_objs)} NFS PVCs deleted successfully")
```
dgaikwad: If the test case fails midway, the cleanup steps are never executed, leaving the cluster in an inconsistent or messy state. Additionally, if the test case fails during the cleanup phase, the entire test is marked as Failed, which is misleading. In this situation, the actual test scenario has already passed, and the failure is related only to cleanup, which is either outside the test scope or part of a different feature.
This results in a false positive failure.
Could we handle this more gracefully in the teardown activity—for example, by ensuring cleanup always runs, and by preventing cleanup failures from affecting the actual test result?

tests/cross_functional/conftest.py
```diff
@@ -3168,3 +3173,280 @@ def _factory(
...
+                constants.VOLUME_MODE_BLOCK,
+            ]
+            non_enc_access_modes = [
+                constants.ACCESS_MODE_RWO,
+                constants.ACCESS_MODE_RWOP,
+                constants.ACCESS_MODE_RWO,
+                constants.ACCESS_MODE_RWX,
+                constants.ACCESS_MODE_RWOP,
+            ]
+            total_non_enc_pvcs = 10
```
dgaikwad: Can we assign with `total_pvcs` value here instead of static 10?

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+        if all_nfs_pod_objs:
+            for nfs_pod_obj in all_nfs_pod_objs:
+                pod.delete_deployment_pods(nfs_pod_obj)
+
+        if all_nfs_pvc_objs:
+            for nfs_pvc_obj in all_nfs_pvc_objs:
+                nfs_pvc_obj.delete()
+                nfs_pvc_obj.ocp.wait_for_delete(
+                    resource_name=nfs_pvc_obj.name, timeout=180
+                )
```
dgaikwad, line 593: Can we move this code somewhere in the teardown activity as this will not execute if test get fail midway.

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+                    provisioner = sc_data.get("provisioner", "")
+
+                    if constants.RBD_PROVISIONER in provisioner:
+                        log.info(f" Flattening RBD clone: {clone_obj.name}")
+                        flatten_image(clone_obj)
+
+                    all_clones.append(clone_obj)
+                    log.info(f" Clone {len(all_clones)} created: {clone_obj.name}")
+
+                except Exception as e:
```
PrasadDesala: Generic exception catch masks specific errors. please throw only the appropriate exceptions

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+
+        if attempt == max_attempts:
+            log.error("Maximum attempts reached. Expected alerts were not detected.")
+            raise TimeoutExpiredError(
+                "Failed to reach 85% cluster full ratio - alerts not triggered"
+            )
+
+        # Count pending clones
+        pending_clones = 0
+        for clone in all_clones[-15:]:  # Check last 15 clones
```
PrasadDesala: how many clones are we creating?

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+            + list(zip(fixture_non_enc_pod_objs, fixture_non_enc_combinations))
+            + [
+                (pod, {**fs_config, "pod_type": "cephfs"})
+                for pod in fixture_cephfs_pod_objs
+            ]
+            + [(pod, {**fs_config, "pod_type": "nfs"}) for pod in fixture_nfs_pod_objs]
+        )
+        self.start_fio_on_pods(fixture_pods_with_config)
+
+        def finalizer():
```
PrasadDesala: NFS resource cleanup is in the test method instead of the fixture finalizer

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+        fs_config = {"volume_mode": constants.VOLUME_MODE_FILESYSTEM}
+        fixture_pods_with_config = (
+            list(zip(fixture_pod_objs, fixture_pvc_combinations))
+            + list(zip(fixture_non_enc_pod_objs, fixture_non_enc_combinations))
+            + [
+                (pod, {**fs_config, "pod_type": "cephfs"})
+                for pod in fixture_cephfs_pod_objs
+            ]
+            + [(pod, {**fs_config, "pod_type": "nfs"}) for pod in fixture_nfs_pod_objs]
+        )
```
PrasadDesala: Similar pod configuration logic is repeated in line 436-442. Consider extracting to a helper method

tests/cross_functional/conftest.py
PrasadDesala: shouldn't we handle nfs cleanup in this same fixture?

tests/cross_functional/conftest.py
```diff
@@ -2965,3 +2971,268 @@ def _factory(
...
+
+        if skip_storage_types is None:
+            skip_storage_types = []
+
+        valid_storage_types = ["encrypted_rbd", "non_encrypted_rbd", "cephfs", "nfs"]
+        if not isinstance(skip_storage_types, list):
+            raise ValueError(
+                f"skip_storage_types must be a list. Valid values: {valid_storage_types}"
+            )
+
```
PrasadDesala, line 3039: Add validation for invalid storage types

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+    skipif_hci_provider_and_client,
+    skipif_disconnected_cluster,
+    skipif_proxy_cluster,
+)
+from ocs_ci.framework.testlib import E2ETest, tier1
+from ocs_ci.ocs import constants
+from ocs_ci.helpers.keyrotation_helper import PVKeyrotation
+from ocs_ci.ocs import ocp
+from ocs_ci.ocs.resources import pod
+from ocs_ci.ocs.resources import ocs
```
PrasadDesala: these two imports can be combined

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+@skipif_hci_provider_and_client
+@skipif_disconnected_cluster
+@skipif_proxy_cluster
+class TestRBDEncryptedPVCKeyRotation(E2ETest):
+    """
+    Test class for RBD encrypted PVC key rotation with Vault KMS.
+    """
+
+    @pytest.fixture(scope="function")
+    def setup_nfs_feature_and_prerequisites(
```
PrasadDesala: Please create a fixture for nfs setup, including its teardown

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+                sc,
+                nfs_ganesha_pod_name,
+            )
+            log.info("NFS feature disabled successfully")
+
+        request.addfinalizer(finalizer)
+        return fixture_storage_resources
+
+    @pytest.fixture(scope="function")
+    def verify_pvc_key_rotation(self):
```
PrasadDesala, line 147: how is it different from wait_till_all_pv_keyrotation_on_vault_kms?

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+                    runtime=300,
+                )
+        log.info(f"FIO workload started on all {len(all_pods_with_config)} pods")
+
+        # Wait for IO completion on all pods
+        for idx, (pod_obj, pvc_config) in enumerate(all_pods_with_config, start=1):
+            pod_obj.get_fio_results()
+        log.info(f"FIO completed on all {len(all_pods_with_config)} pods")
+
+    def create_clones_until_full(
```
PrasadDesala: most of the code is redundant. test_clone_deletion_after_max_cluster_space_utilization.py 
Please create a common function and place it in e2e_helpers and then use it in both these tests

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+
+        # Step 5: Add key rotation annotation to storage class
+        self.pvk_obj = PVKeyrotation(self.sc_obj)
+        self.pvk_obj.annotate_storageclass_key_rotation(schedule="*/2 * * * *")
+        log.info(
+            "Added annotation: keyrotation.csiaddons.openshift.io/schedule='*/2 * * * *'"
+        )
+        log.info("SETUP COMPLETE: Ready to create PVCs and test key rotation")
+
+    @tier1
```
PrasadDesala, line 314: why is it added to tier1?

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+        6. Start FIO workload on all pods with verify=True option
+        7. Wait for 2 minutes for key rotation to occur
+        8. Verify key rotation happened for all encrypted PVs
+        9. Create snapshots, clones, expand PVCs, and restore from snapshots for test PVCs
+        10. Start cloning multiple PVCs of both RBD and CephFS till cluster reaches 85% full ratio,
+            ensuring at least 7 pending clone jobs
+
+        """
+
+        # Call the fixture factory to create all PVCs and deployment pods with "test-" prefix
```
PrasadDesala: this can be removed

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+            # This ALWAYS runs, even if above step fails
+            # Ensures cluster is recovered from full/read-only state
+            change_ceph_full_ratio(95)
+
+        log.info("Waiting for 2 minutes to ensure key rotation has occurred")
+        time.sleep(120)  # Wait 2 minutes for key rotation schedule
+        verify_pvc_key_rotation(pvc_objs, self.pvk_obj)
+
+        log.info("Cleaning up all NFS resources ")
+
```
PrasadDesala: move it to teardown

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+        if all_nfs_pvc_objs:
+            for nfs_pvc_obj in all_nfs_pvc_objs:
+                nfs_pvc_obj.delete()
+                nfs_pvc_obj.ocp.wait_for_delete(
+                    resource_name=nfs_pvc_obj.name, timeout=180
+                )
+
+
+# Made with Bob
```
PrasadDesala: it can be removed

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+                    log.error(f"Key rotation failed for PVC {pvc_obj.name}: {str(e)}")
+                    rotation_results.append(False)
+                    raise
+
+            log.info("Key rotation verified successfully for all PV's")
+            return rotation_results
+
+        return _verify_key_rotation
+
+    def start_fio_on_pods(self, all_pods_with_config):
```
dgaikwad: Can we write this function in common so other test also can use it if required?

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,596 @@
...
+
+                except Exception as e:
+                    log.warning(
+                        f"Clone creation failed (expected when cluster is full): {e}"
+                    )
+                    break
+
+            # Check if we've reached the target alerts
+            log.info("Checking for CephOSDNearFull and CephOSDCriticallyFull alerts")
+            expected_alerts = ["CephOSDCriticallyFull"]
```
dgaikwad: Can we define and get these types of constant alert globally?

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
posharma1: verify_pvc_key_rotation is a test-specific wrapper providing detailed per-PVC logging and error handling, while wait_till_all_pv_keyrotation_on_vault_kms is a batch verification method for validation of multiple PVCs at once.

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
posharma1: This method is specific to this test's multi-storage-type setup with custom FIO parameters. Other tests have different requirements, so keeping it as a test class method will be better

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
posharma1: can you please elaborate

tests/cross_functional/conftest.py
```diff
@@ -2968,3 +2976,409 @@ def _factory(
...
+            logger.info(
+                f" All {len(nfs_pod_objs)} NFS deployment pods created and running"
+            )
+
+        # Return resources in a nested structure for better organization
+        resources = {
+            "encrypted_rbd": {
+                "pvcs": pvc_objs,
+                "pods": pod_objs,
+                "combinations": pvc_combinations,
```
PrasadDesala, line 3255: It can fail with NameError when skip_storage_types includes "encrypted_rbd"

tests/cross_functional/conftest.py
```diff
@@ -2968,3 +2976,409 @@ def _factory(
...
+            },
+            "non_encrypted_rbd": {
+                "pvcs": non_enc_pvc_objs,
+                "pods": non_enc_pod_objs,
+                "combinations": non_enc_combinations,
```
PrasadDesala, line 3260: The same problem as abive exists for non_enc_combinations when "non_encrypted_rbd" is skipped

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,510 @@
...
+        )
+
+        log.info(" Waiting for key rotation to occur")
+        wait_time = 120  # 2 minutes
+        buffer_time = 30
+        total_wait = wait_time + buffer_time
+        time.sleep(total_wait)
+
+        log.info(" Verifying key rotation for all encrypted PVs")
+        verify_pvc_key_rotation(pvc_objs, self.pvk_obj)
```
PrasadDesala: Key rotation verification result is never asserted

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,510 @@
...
+        # Add key rotation annotation to storage class
+        self.pvk_obj = PVKeyrotation(self.sc_obj)
+        self.pvk_obj.annotate_storageclass_key_rotation(schedule="*/2 * * * *")
+        log.info(
+            "Added annotation: keyrotation.csiaddons.openshift.io/schedule='*/2 * * * *'"
+        )
+        log.info("SETUP COMPLETE: Ready to create PVCs and test key rotation")
+
+        def finalizer():
+
```
PrasadDesala: doc strings missing

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,510 @@
...
+            pod_objs=pod_objs,
+            pvc_combinations=pvc_combinations,
+            non_enc_pod_objs=non_enc_pod_objs,
+            non_enc_combinations=non_enc_combinations,
+            cephfs_pod_objs=cephfs_pod_objs,
+            nfs_pod_objs=nfs_pod_objs,
+        )
+
+        log.info(" Waiting for key rotation to occur")
+        wait_time = 120  # 2 minutes
```
PrasadDesala, line 382: instead of a harcoded wait, use TimeoutSampler which polls and exits as soon as the condition is met, making the test faster when rotation completes early and more robust when it takes longer than expected.

tests/cross_functional/conftest.py
```diff
@@ -2968,3 +2976,409 @@ def _factory(
...
+
+                except Exception as e:
+                    logger.warning(
+                        f"Clone creation failed (expected when cluster is full): {e}"
+                    )
+                    break
+
+            # Check if we've reached the target alerts
+            logger.info(f"Checking for expected alerts: {expected_alerts}")
+            prometheus = PrometheusAPI(threading_lock=threading_lock)
```
PrasadDesala: It creates a new PrometheusAPI instance on every loop iteration. This instantiates a fresh API object on every batch. Move the instantiation outside the loop.

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,510 @@
...
+
+            change_ceph_full_ratio(85)
+
+        request.addfinalizer(finalizer)
+
+    @tier1
+    def test_rbd_encrypted_pvc_keyrotation_all_combinations(
+        self,
+        create_multiple_storage_pvcs_pods,
+        setup_nfs_feature,
```
PrasadDesala, line 319: setup_prerequisites already declares setup_nfs_feature as a dependency, so NFS will be enabled when setup_prerequisites is used.  Do we still need it here?

tests/cross_functional/conftest.py
```diff
@@ -2968,3 +2976,409 @@ def _factory(
...
+                    selector=[deployment_name],
+                    selector_label="name",
+                )
+                nfs_pod_obj = nfs_pod_list[0]
+                nfs_pod_objs.append(nfs_pod_obj)
+                # Track NFS resources for cleanup in teardown
+                all_nfs_resources["pods"].append(nfs_pod_obj)
+
+            # Track NFS PVCs for cleanup in teardown
+            all_nfs_resources["pvcs"].extend(nfs_pvc_objs)
```
PrasadDesala: NFS pods are added to all_nfs_resources["pods"] on each loop iteration (inside the NFS block), but the NFS PVCs are batch extended after the loop. any reason?

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,510 @@
...
+            pvcs_to_clone = pvc_objs + non_enc_pvc_objs + cephfs_pvc_objs
+            log.info(f"Total PVCs available for cloning: {len(pvcs_to_clone)}")
+
+            full_ratio_clones = create_clones_until_cluster_full(
+                pvcs_to_clone=pvcs_to_clone,
+                clone_batch_size=10,
+                max_attempts=12,
+                expected_alerts=["CephOSDCriticallyFull"],
+                flatten_rbd_clones=True,
+                min_pending_clones=7,
```
PrasadDesala, line 461: The create_clones_until_cluster_full fixture counts pending clones and logs them, but there is no assertion that the minimum is met. The test doc string says "at least 7 pending"

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
PrasadDesala: this comment is still not fixed

tests/cross_functional/conftest.py
```diff
@@ -2968,3 +2976,409 @@ def _factory(
...
+    Fixture factory to create multiple PVCs and deployment pods across different storage types.
+    Returns:
+        function: Factory function that accepts parameters and returns created resources
+    """
+    # Track all NFS resources created across all factory calls
+    all_nfs_resources = {"pods": [], "pvcs": []}
+
+    def teardown():
+        """
+        Teardown to restore Ceph full ratio and clean up NFS resources.
```
PrasadDesala: Teardown docstring says "restore Ceph full ratio" but the teardown never calls ceph full ratio restore

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,510 @@
...
+                try:
+                    rotation_success = pvk_obj.wait_till_keyrotation(volume_handle)
+                    rotation_results.append(rotation_success)
+
+                except Exception as e:
+                    log.error(f"Key rotation failed for PVC {pvc_obj.name}: {str(e)}")
+                    rotation_results.append(False)
+                    raise
+
+            log.info("Key rotation verified successfully for all PV's")
```
PrasadDesala:
```suggestion
            log.info("Key rotation verified successfully for all PVs")
```

tests/cross_functional/conftest.py
```diff
@@ -2968,3 +2976,409 @@ def _factory(
...
+                (constants.VOLUME_MODE_BLOCK, constants.ACCESS_MODE_RWX),
+            ]
+            pvc_combinations = [
+                {"volume_mode": vm, "access_mode": am, "size": 1}
+                for vm, am in itertools.islice(
+                    itertools.cycle(enc_base_combinations), total_pvcs
+                )
+            ]
+
+            for idx, pvc_config in enumerate(pvc_combinations, start=1):
```
PrasadDesala: The idx loop variable used but never used in the loop body

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
PrasadDesala: this comment is not still not incorporated.

tests/cross_functional/conftest.py
```diff
@@ -2968,3 +2976,409 @@ def _factory(
...
+                pvc_obj = pvc_factory(
+                    interface=constants.CEPHBLOCKPOOL,
+                    project=proj_obj,
+                    storageclass=sc_obj,
+                    size=pvc_config["size"],
+                    access_mode=pvc_config["access_mode"],
+                    volume_mode=pvc_config["volume_mode"],
+                    status=constants.STATUS_BOUND,
```
PrasadDesala, line 3078: multi_pvc_factory already accepts an access_modes list with embedded volume mode (e.g., "ReadWriteMany-Block") and a num_of_pvc count. same for non NFS storage types.

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
posharma1: done

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
posharma1: there is no resource for which we can poll and wait 2 minutes is for key rotation schedule for rotation to happen and it will not finish rotation before 2 minutes

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
posharma1: setup_prerequisites already depends on setup_nfs_feature, so pytest won't run it twice. The duplicate listing is intentional — it makes the fixture ordering dependency explicit to readers and ensures setup_nfs_feature is always set up even if setup_prerequisites is ever refactored to drop its dependency.

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
posharma1: updated code

tests/cross_functional/conftest.py
posharma1: updated doc string

tests/cross_functional/conftest.py
posharma1: added

tests/cross_functional/conftest.py
posharma1: multi_pvc_factory may include all combination but here for each type (cephfs, nfs, rbd, encrypted red) explicitly we are picking supported combinations defined here                                        enc_base_combinations [
                (constants.VOLUME_MODE_FILESYSTEM, constants.ACCESS_MODE_RWO),
                (constants.VOLUME_MODE_BLOCK, constants.ACCESS_MODE_RWO),
                (constants.VOLUME_MODE_BLOCK, constants.ACCESS_MODE_RWX),
            ]

tests/cross_functional/conftest.py
posharma1: done

tests/cross_functional/conftest.py
posharma1: done

tests/cross_functional/conftest.py
posharma1: done

tests/cross_functional/conftest.py
posharma1: updated code

tests/cross_functional/conftest.py
posharma1: updated code

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
posharma1: done

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
posharma1: done

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,519 @@
...
+from ocs_ci.utility import nfs_utils
+from ocs_ci.ocs.cluster import change_ceph_full_ratio
+from ocs_ci.helpers import helpers
+
+log = logging.getLogger(__name__)
+
+
+@magenta_squad
+@system_test
+@ignore_leftovers
```
dgaikwad, line 30: I think we can not add entirely "@ignore_leftovers". Can you please check with @PrasadDesala

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,519 @@
...
+@skipif_hci_provider_and_client
+@skipif_disconnected_cluster
+@skipif_proxy_cluster
+class TestRBDEncryptedPVCKeyRotation(E2ETest):
+    """
+    Test class for RBD encrypted PVC key rotation with Vault KMS.
+    """
+
+    @pytest.fixture(scope="function")
+    def setup_nfs_feature(self, request):
```
dgaikwad, line 42: Our multiple testcases are using nfs enable feature so this should be common fixture so every testcase will it and have to move it somewhere common place

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,519 @@
...
+                storage_cluster_obj,
+                config_map_obj,
+                pod_obj,
+                sc,
+                nfs_ganesha_pod_name_holder[0],
+            )
+            log.info("NFS feature disabled successfully")
+
+        # Register the finalizer BEFORE enabling NFS so cleanup always runs
+        request.addfinalizer(finalizer)
```
dgaikwad, line 83: This finalizer should get added immediately after nfs_enable step complete not before the nfs_enable step.

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,519 @@
...
+                )
+        log.info(f"FIO workload started on all {len(all_pods_with_config)} pods")
+
+        # Wait for IO completion on all pods
+        for idx, (pod_obj, pvc_config) in enumerate(all_pods_with_config, start=1):
+            pod_obj.get_fio_results()
+        log.info(f"FIO completed on all {len(all_pods_with_config)} pods")
+
+    @pytest.fixture(autouse=True)
+    def setup_encrypted_storage(
```
dgaikwad, line 251: Do you think we might use this fixture in other test cases in the future? If so, I think we should move it to a common location and invoke it explicitly instead of using an autouse=True fixture.

tests/cross_functional/system_test/test_rbd_encrypted_pvc_keyrotation_system.py
```diff
@@ -0,0 +1,519 @@
...
+            non_enc_combinations=non_enc_combinations,
+            cephfs_pod_objs=cephfs_pod_objs,
+            nfs_pod_objs=nfs_pod_objs,
+        )
+
+        log.info(" Waiting for key rotation to occur")
+        wait_time = 120  # 2 minutes
+        buffer_time = 30
+        total_wait = wait_time + buffer_time
+        time.sleep(total_wait)
```
dgaikwad, line 385: why don't we add directly 120 + 30 that is 150 seconds, we are not using wait_time, buffer_time, total_wait other places as well.

Issue #14686, opened 2026-03-13
test_duplicate_bucket_creation[3-S3] fails in 4.21 while there is no bug in the product - fix the test


ypersky1980:
Relevant run : 

https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/994/42968/1979270/1979303/log

ypersky1980:
https://github.com/red-hat-storage/ocs-ci/pull/14688 PR with a fix

ypersky1980:
PR was merged, closing the issue.

PR #14688, opened 2026-03-14
Fix to creation of duplicate bucket negative test case

ypersky1980:
This PR contains a fix to creation of duplicate bucket negative test case (https://github.com/red-hat-storage/ocs-ci/issues/14686) 
When bucket creation failed because bucket with such name already existed , the exception was not propagated, and now it is.

### DIFF
--- a/ocs_ci/ocs/resources/objectbucket.py
+++ b/ocs_ci/ocs/resources/objectbucket.py
@@ -489,8 +489,7 @@ class MCGS3Bucket(ObjectBucket):
         except botocore.exceptions.ClientError as e:
             if e.response["Error"]["Code"] == "BucketAlreadyOwnedByYou":
                 logger.info(f"Bucket {self.name} already exists")
-            else:
-                raise e
+            raise e
 
     def internal_delete(self):
         """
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/resources/objectbucket.py
```diff
@@ -489,8 +489,7 @@ def __init__(self, *args, **kwargs):
         except botocore.exceptions.ClientError as e:
             if e.response["Error"]["Code"] == "BucketAlreadyOwnedByYou":
                 logger.info(f"Bucket {self.name} already exists")
-            else:
-                raise e
+            raise e
```
udaysk23, line 492: Why are we removing else part from here?

ypersky1980: @udaysk23 We need an exception to be propagated in any case, since otherwise the test logic is broken ( which we see in cinstant failures).

udaysk23: If we are raising exception directly then we are defeating the purpose of `try...catch` block

ypersky1980: @udaysk23 The purpose of try-catch block here is as following: to catch the exception and if the cause is "bucket already exists", write the corresponding message to the log and rethrow the exception
If we remove the try-catch block completely, as you suggest' the exception will sure be propagated and the test will work but the log message for "bucket already exists" will not be written here (and only in the test).

ypersky1980: Please let me know what do you prefer in this case.

udaysk23 (approved):

keesturam (approved):

PR #14708, opened 2026-03-17
Add Claude AI code review workflow

AviadP:
- [x] Add GitHub Actions workflow triggered by `claude-review` label or `@claude` comment
- [x] Add review guide with severity calibration (BLOCKER/HIGH/MEDIUM/LOW) for ocs-ci rules
- [x] Advisory only (comment mode) — never blocks PRs
- [x] Critical rules enforced: specific exceptions, centralized locators, no yield fixtures, tier marker conflicts, teardown required, version-agnostic selectors

to decide:

- where to place the reviewer tool (should be under our organization, in my opinion)
- make it mandatory or not
- rules is just a suggestion, open for discussion

[review tool link](https://github.com/AviadP/claude-review-action)

### DIFF
--- a/.github/claude-review-guide.md
+++ b/.github/claude-review-guide.md
@@ -1,0 +1,82 @@
+# Claude Review Guide for OCS-CI
+
+This guide complements `.claude/CLAUDE.md` — it defines **how to review** code, not how to write it.
+
+## Review Scope
+
+- Only review code added or modified by the PR — never flag existing code
+- If the codebase already uses a pattern, don't flag it in new code that follows the same pattern
+- Read docstrings and comments to understand the author's intent before flagging design decisions
+- Ask "is this a realistic failure scenario?" — not "could this theoretically fail?"
+
+## Severity Calibration
+
+Map findings to severity levels based on real impact, not theoretical risk.
+
+### BLOCKER
+
+These violate critical project rules and must always be flagged:
+
+- `except Exception` or bare `except:` — must use specific exceptions
+- UI locators defined outside `ocs_ci/ocs/ui/views.py` (unless the locator contains a variable)
+- Fixtures using `yield` — must use `request.addfinalizer()` instead
+- Tier marker at class level when `parametrize` marks have different tiers
+- Test classes that create resources without a `teardown()` method
+- Version-specific CSS selectors (`.pf-c-` instead of `[class*='c-']`)
+
+### HIGH
+
+- Missing type hints on new public method parameters or return values
+- Missing docstrings on new public methods
+- Magic numbers (unnamed timeout values like `time.sleep(2)`)
+- Use of `@pytest.mark.usefixtures` instead of passing fixtures as parameters
+- Global variables used for sharing state between tests
+
+### MEDIUM
+
+- Missing `log_step()` in UI test methods (non-UI tests use `logger.info()`)
+- Missing fallback selectors for UI elements known to be flaky
+- Inconsistent return types (method returns different types based on a flag)
+
+### LOW
+
+- Missing exception chaining (`from e`)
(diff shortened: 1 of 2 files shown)


.github/claude-review-guide.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Severity Calibration
+
+Map findings to severity levels based on real impact, not theoretical risk.
+
+### BLOCKER
+
+These violate critical project rules and must always be flagged:
+
+- `except Exception` or bare `except:` — must use specific exceptions
```
petr-balogh, line 20: l don't see this as blocker, sometime we use this.

We definitely should not allow diabolic anti pattern like:
```
except Exception:
    pass
```

Better to always use specific exception whenever it's possible, but sometimes there are so many possibilities that catching generic Exception is only one solution to avoid many different unpredictable exceptions.

.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,53 @@
...
+          anthropic-api-key: ${{ secrets.ANTHROPIC_API_KEY }}
+          model: claude-sonnet-4-6
+          review-authority: comment-only
+          review-guide-path: .github/claude-review-guide.md
+          max-diff-lines: '4000'
+          max-diff-bytes: '120000'
+          max-files: '40'
+          context-intro: >
+            You are reviewing a PR in ocs-ci — a Python/pytest testing framework
+            for Red Hat OpenShift Container Storage (OCS/ODF). The codebase uses
```
petr-balogh, line 37: I would change to something like this
```suggestion
            for Red Hat OpenShift Data Foundation (ODF) formerly known as OpenShift Container Storage OCS. The codebase uses
```

.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,53 @@
...
+            Selenium for UI tests, pytest fixtures for resource management, and
+            has strict conventions for locator placement, exception handling, and
+            test teardown. Read the project CLAUDE.md for full conventions.
+          critical-rules: |
+            1. Never use `except Exception` or bare `except:` — always use specific exceptions
```
petr-balogh, line 42: As pointed bellow , this probably should not be a blocker. But if it's used we should log the exception out:
```
logger.error("Exception hit:", exc_info=True)
```
something like this. So if something will debugged we see what is happening.

.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,53 @@
...
+            2. UI locators MUST be centralized in `ocs_ci/ocs/ui/views.py` (unless the locator contains a variable)
+            3. Fixtures MUST NOT use `yield` — use `request.addfinalizer()` instead
```
petr-balogh, line 44: why yield is not acceptable? I think it's a valid scenario which we use

.github/claude-review-guide.md
```diff
@@ -0,0 +1,82 @@
...
+- UI locators defined outside `ocs_ci/ocs/ui/views.py` (unless the locator contains a variable)
+- Fixtures using `yield` — must use `request.addfinalizer()` instead
```
petr-balogh, line 22: Pointed out above, this is valid case to use yield

.github/claude-review-guide.md
AviadP: the tool should only add comments, not change requests, so it is not real blocker. using braod exceptions makes debug much harder and sometimes lets test that should fail to pass

.github/workflows/claude-review.yml
AviadP: https://ocs-ci.readthedocs.io/en/latest/fixture_usage.html#avoid

AviadP:
we already have code rabbit, no reason to keep this up

PR #14727, opened 2026-03-20
Fix ARO cluster deletion failure when resources already deleted

petr-balogh:
Wraps DNS records and ARO cluster deletion in try-except blocks to handle
  ResourceNotFound errors gracefully. When resources don't exist (already deleted),
  logs an info message and continues with remaining cleanup operations instead of
  raising CommandFailed. Ensures all three cleanup steps (DNS, cluster, VNET)
  execute independently for idempotent best-effort cleanup.

  Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>


petr-balogh:
Verification job:
https://url.corp.redhat.com/0cb54de

PR #14733, opened 2026-03-23
fix acm submariner validation using ops ci framework

deepanjanacharya:
fix acm submariner validation using ops ci framework

## Summary by CodeRabbit

* **New Features**
  * Automatic configuration of Advanced Cluster Management (ACM) and Submariner during initialization and test setup, reducing manual steps
  * Automatic detection of ACM/Submariner release type based on OpenShift version for smoother deployments

* **Chores**
  * Updated component versions and environment defaults for OpenShift 4.22 support and related tooling configuration

[Review Change Stack](https://app.coderabbit.ai/change-stack/red-hat-storage/ocs-ci/pull/14733?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -2592,8 +2592,9 @@ class Deployment(object):
 
     def deploy_acm_hub(self):
         """
-        Handle ACM HUB deployment
+        Deploy ACM hub cluster
         """
+
         if self.acm_operator_installed():
             logger.info("ACM Operator is already installed")
             self.deploy_multicluster_hub()
@@ -2610,6 +2611,7 @@ class Deployment(object):
         else:
             self.deploy_acm_hub_released()
             self.deploy_multicluster_hub()
+
         if config.ENV_DATA.get("configure_acm_to_import_mce"):
             self.configure_acm_to_import_mce_clusters()
         from ocs_ci.ocs.acm.acm import verify_running_acm
--- a/ocs_ci/framework/conf/ocp_version/ocp-4.21-config.yaml
+++ b/ocs_ci/framework/conf/ocp_version/ocp-4.21-config.yaml
@@ -15,5 +15,4 @@ ENV_DATA:
   mce_version: "2.11"
   submariner_version: "0.23.1"
   # Below values needs to be changed when acm and submariner are GAed
-  submariner_source: "downstream"
   oadp_version: "1.5"
--- a/ocs_ci/framework/conf/ocp_version/ocp-4.22-config.yaml
+++ b/ocs_ci/framework/conf/ocp_version/ocp-4.22-config.yaml
@@ -22,7 +22,4 @@ ENV_DATA:
   submariner_version: "0.23.1"
   submariner_image: "quay.io/redhat-user-workloads/submariner-tenant/submariner-fbc-4-21@sha256:1585a2a6c372f8a66b7644e742d95356b577072dfc83492c14759f6b2372b609"
   # Below values needs to be changed when acm and submariner are GAed
-  submariner_source: "downstream"
-  submariner_release_type: "released"
-  acm_hub_unreleased: true
   oadp_version: "1.6"
--- a/ocs_ci/framework/pytest_customization/ocscilib.py
+++ b/ocs_ci/framework/pytest_customization/ocscilib.py
@@ -407,6 +407,7 @@ def pytest_configure(config):
     set_log_record_factory()
     # Ensure custom log levels are registered
     from ocs_ci.framework.custom_logger import TEST_STEP, ASSERTION, AI_DATA
+    from ocs_ci.utility.utils import auto_configure_acm, auto_configure_submariner
 
     logging.addLevelName(TEST_STEP, "TEST_STEP")
     logging.addLevelName(ASSERTION, "ASSERTION")
@@ -422,6 +423,8 @@ def pytest_configure(config):
 
         if not (config.getoption("--help") or config.getoption("collectonly")):
             process_cluster_cli_params(config)
+            auto_configure_acm()
+            auto_configure_submariner()
             config_file = os.path.expanduser(
                 os.path.join(
                     ocsci_config.RUN["log_dir"],
--- a/ocs_ci/ocs/defaults.py
+++ b/ocs_ci/ocs/defaults.py
@@ -195,5 +195,31 @@ API_SSL_CERT = os.path.join(constants.DATA_DIR, "api-cert.crt")
 API_SSL_KEY = os.path.join(constants.DATA_DIR, "api-cert.key")
 API_SSL_CA_CERT = os.path.join(constants.DATA_DIR, "ca.crt")
 
+ocp_to_acm_unreleased_mapping = {
+    "4.14": False,
+    "4.15": False,
+    "4.16": False,
+    "4.17": False,
+    "4.18": False,
+    "4.19": False,
+    "4.20": False,
+    "4.21": False,
+    "4.22": True,
+    "4.23": True,
+}
+
+ocp_to_submariner_unreleased_mapping = {
+    "4.14": False,
+    "4.15": False,
+    "4.16": False,
+    "4.17": False,
+    "4.18": False,
+    "4.19": False,
+    "4.20": False,
+    "4.21": False,
+    "4.22": False,
+    "4.23": True,
+}
+
 # Storage cluster related defaults
 DEVICE_TYPE = "SSD"
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -7613,3 +7613,65 @@ def genereate_cred_file_rack():
     )
 
     return rack_dict
+
+
+def auto_configure_acm():
+    """
+    Auto-configure ACM as released or unreleased based on OCP version
+    If explicitly configured by user, skips auto-configuration.
+    Uses OCP version mappings with safe defaults (unreleased) for unknown versions.
+    """
+    # If explicitly configured by user, skip auto-configuration
+    if config.ENV_DATA.get("acm_hub_unreleased") is not None:
+        log.info(
+            f"ACM explicitly configured: acm_hub_unreleased={config.ENV_DATA['acm_hub_unreleased']}"
+        )
+        return
+    # Get OCP version
+    ocp_version = version_module.get_semantic_ocp_version_from_config()
+    # Look up in mapping, default to True (unreleased) for unknown versions
+    acm_unreleased = defaults.ocp_to_acm_unreleased_mapping.get(f"{ocp_version}", True)
+    config.ENV_DATA["acm_hub_unreleased"] = acm_unreleased
+    if f"{ocp_version}" in defaults.ocp_to_acm_unreleased_mapping:
+        log.info(
+            f"OCP {ocp_version} → ACM: {'UNRELEASED' if acm_unreleased else 'RELEASED'}"
+        )
+    else:
+        log.info(f"OCP {ocp_version} not in mapping → ACM: UNRELEASED (safe default)")
+
+
+def auto_configure_submariner():
+    """
+    Auto-configure Submariner as released or unreleased based on OCP version.
+    If explicitly configured by user, skips auto-configuration.
+    Uses OCP version mappings with safe defaults (unreleased) for unknown versions.
+    """
+    # Set submariner_source default if not provided by user
+    if config.ENV_DATA.get("submariner_source") is None:
+        config.ENV_DATA["submariner_source"] = "downstream"
+        log.info("Submariner source not specified, defaulting to: downstream")
+
+    # If explicitly configured by user, skip auto-configuration
+    if config.ENV_DATA.get("submariner_release_type") is not None:
+        log.info(
+            f"Submariner explicitly configured: release_type={config.ENV_DATA['submariner_release_type']}"
+        )
+        return
+    # Get OCP version
+    ocp_version = version_module.get_semantic_ocp_version_from_config()
+    # Look up in mapping, default to True (unreleased) for unknown versions
+    submariner_unreleased = defaults.ocp_to_submariner_unreleased_mapping.get(
+        f"{ocp_version}", True
+    )
+    config.ENV_DATA["submariner_release_type"] = (
+        "unreleased" if submariner_unreleased else "released"
+    )
+
+    if f"{ocp_version}" in defaults.ocp_to_submariner_unreleased_mapping:
+        log.info(
+            f"OCP {ocp_version} → Submariner: {'UNRELEASED' if submariner_unreleased else 'RELEASED'}"
+        )
+    else:
+        log.info(
+            f"OCP {ocp_version} not in mapping → Submariner: UNRELEASED (safe default)"
+        )
--- a/conf/ocsci/acm_hub_released_deploy.yaml
+++ b/conf/ocsci/acm_hub_released_deploy.yaml
@@ -1,0 +1,5 @@
+---
+ENV_DATA:
+  deploy_acm_hub_cluster: true
+  skip_ocs_deployment: true
+  acm_hub_unreleased: false
--- a/conf/ocsci/submariner_downstream_released.yaml
+++ b/conf/ocsci/submariner_downstream_released.yaml
@@ -1,0 +1,5 @@
+ENV_DATA:
+  submariner_source: "downstream"
+  submariner_release_type: "released"
+  submariner_channel: ""
+  subctl_version: "subctl-rhel9:v0.23"
--- a/ocs_ci/framework/conf/ocp_version/ocp-4.22-ga-config.yaml
+++ b/ocs_ci/framework/conf/ocp_version/ocp-4.22-ga-config.yaml
@@ -17,7 +17,4 @@ ENV_DATA:
   submariner_version: "0.23.1"
   submariner_image: "quay.io/redhat-user-workloads/submariner-tenant/submariner-fbc-4-21@sha256:1585a2a6c372f8a66b7644e742d95356b577072dfc83492c14759f6b2372b609"
   # Below values needs to be changed when acm and submariner are GAed
-  submariner_source: "downstream"
-  submariner_release_type: "released"
-  acm_hub_unreleased: true
   oadp_version: "1.6"


ocs_ci/deployment/deployment.py
```diff
@@ -2448,31 +2448,33 @@ def acm_operator_installed(self):
...
-        else:
-            self.deploy_acm_hub_released()
-            self.deploy_multicluster_hub()
-        if config.ENV_DATA.get("configure_acm_to_import_mce"):
-            self.configure_acm_to_import_mce_clusters()
-        from ocs_ci.ocs.acm.acm import verify_running_acm
-
-        verify_running_acm()
+    def deploy_acm_hub(self):                                                                                                                                                                                                                         
+      """
```
petr-balogh: Wrong indetion

ocs_ci/utility/utils.py
```diff
@@ -6769,3 +6769,94 @@ def create_kubeconfig(kubeconfig_path):
                 "Kubeconfig doesn't exists and RUN['kubeadmin_password'] and RUN['ocp_url'] "
                 "environment variables were not provided."
             )
+
+def get_current_git_branch():
+      """
```
petr-balogh: Here it's overindented

ocs_ci/deployment/deployment.py
```diff
@@ -2448,31 +2448,33 @@ def acm_operator_installed(self):
...
-            else:
-                self.deploy_acm_hub_unreleased()
-        else:
-            self.deploy_acm_hub_released()
-            self.deploy_multicluster_hub()
-        if config.ENV_DATA.get("configure_acm_to_import_mce"):
-            self.configure_acm_to_import_mce_clusters()
-        from ocs_ci.ocs.acm.acm import verify_running_acm
-
-        verify_running_acm()
```
petr-balogh, line 2629: Was this removed intentionally `verify_running_acm` ?

ocs_ci/utility/utils.py
```diff
@@ -6769,3 +6769,94 @@ def create_kubeconfig(kubeconfig_path):
...
+      Returns:                                                                                                                                                                                                                                      
+          bool: True if branch matches release-* pattern                                                                                                                                                                                            
+      """                                                                                                                                                                                                                                           
+      return re.match(r'^release-\d+\.\d+$', branch_name) is not None
+
+def auto_configure_acm_submariner():                                                                                                                                                                                                              
+      """
```
petr-balogh: We should not anyhow rely on GIT and specific branch here - as we should not anyhow know from framework about git repor of ocs-ci as it needs to run independently on git.

What I suggest is to have kind of mapping - maybe in defaults.py or constants.py uder `https://github.com/red-hat-storage/ocs-ci/tree/master/ocs_ci/ocs`

That map can look like:

ocp_to_acm_unrelease_mapping = {
     "4.16": False,
     "4.18": False,
...
     "4.22": True,
}

Similarly to submariner mapping.

Once you will be getting the value from the mapping, if the version key is missing - use by default the value we want. 

E.g. for 4.23 - we don't want to fail on the missing 4.23 key - so use ocp_to_acm_unrelease_mapping.get(get_ocp_version_from_config, True) # Where by default if missing key we will get unreleased in this case.



2) We need to remove values from ocp version config files like `acm_hub_unreleased` and `submariner_release_type`.

3) If the value is propagated by user don't call this function. Hence will be good if we decouple this function to two to handle submariner and acm separately.

ocs_ci/deployment/deployment.py
```diff
@@ -2448,31 +2448,33 @@ def acm_operator_installed(self):
...
+    def deploy_acm_hub(self):                                                                                                                                                                                                                         
+      """                                                                                                                                                                                                                                           
+      Deploy ACM hub cluster                                                                                                                                                                                                                        
+      """                                                                                                                                                                                                                                           
+      # Auto-configure ACM and Submariner based on branch if not explicitly set
+      from ocs_ci.utility.utils import auto_configure_acm_submariner
```
petr-balogh: Here if user provided value like `acm_hub_unreleased` and `submariner_release_type` via config - do not call that function to set it automatically - as user passed value should take a prcedence.

ocs_ci/utility/utils.py
deepanjanacharya: Fixed the indentation.

ocs_ci/deployment/deployment.py
deepanjanacharya: Fixed the indentation

ocs_ci/deployment/deployment.py
deepanjanacharya: added back the function. it was removed unintentionally

ocs_ci/utility/utils.py
deepanjanacharya: Implemented two functions now. also created a mapping.

ocs_ci/ocs/defaults.py
```diff
@@ -194,3 +194,27 @@
...
+ocp_to_acm_unreleased_mapping = {
+"4.21": False,
+"4.20": False,
+"4.19": False,
+"4.18": False,
+"4.17": False,
+"4.16": False,
+"4.15": False,
+"4.14": False,
+"4.22": True,
```
petr-balogh: can you probablyu sort it out from latest to newest version to have 4.22 as last as you have but 4.14 as first?

Same below

ocs_ci/ocs/defaults.py
```diff
@@ -194,3 +194,27 @@
...
+}
```
petr-balogh: Missing new line at the end of file

ocs_ci/utility/utils.py
```diff
@@ -6769,3 +6769,84 @@ def create_kubeconfig(kubeconfig_path):
...
+    Args:
+    ocp_version_str (str): OCP version string (e.g., "4.22")
+    """                                                                                                                              
+    # If explicitly configured by user, skip auto-configuration
+    if config.ENV_DATA.get("submariner_release_type") is not None:                                                                   
+        log.info(f"Submariner explicitly configured: release_type={config.ENV_DATA['submariner_release_type']}")
+        return
+    # Get OCP version
+    ocp_version = version_module.get_semantic_ocp_version_from_config()                                                              
+    ocp_version_str = f"{ocp_version.major}.{ocp_version.minor}"
```
petr-balogh:
```
def get_semantic_ocp_version_from_config():
    """
    Returning OCP semantic version from config.

    Returns:
       semantic_version.base.Version: Object of semantic version for OCP.

    """
    return get_semantic_version(config.DEPLOYMENT["installer_version"], True)
```
As True in get_semantic_version is for `only_major_minor` it will only have major.minor so you can safely assume while doing string conversion that it will be in format X.Y

So you can avoid defining ocp_version_str and directly us as I will show below.

ocs_ci/utility/utils.py
```diff
@@ -6769,3 +6769,84 @@ def create_kubeconfig(kubeconfig_path):
...
+    # Look up in mapping, default to True (unreleased) for unknown versions
+    submariner_unreleased = defaults.ocp_to_submariner_unreleased_mapping.get(ocp_version_str, True)
```
petr-balogh:
```suggestion
    submariner_unreleased = defaults.ocp_to_submariner_unreleased_mapping.get(f"{ocp_version}", True)
```

ocs_ci/utility/utils.py
```diff
@@ -6769,3 +6769,84 @@ def create_kubeconfig(kubeconfig_path):
...
+    config.ENV_DATA["submariner_release_type"] = "unreleased" if submariner_unreleased else "released"
+                                                                                                                                       
+    if ocp_version_str in defaults.ocp_to_submariner_unreleased_mapping:
+        log.info(f"OCP {ocp_version_str} → Submariner: {'UNRELEASED' if submariner_unreleased else 'RELEASED'}")
```
petr-balogh:
```suggestion
        log.info(f"OCP {ocp_version} → Submariner: {'UNRELEASED' if submariner_unreleased else 'RELEASED'}")                     
```

ocs_ci/utility/utils.py
```diff
@@ -6769,3 +6769,84 @@ def create_kubeconfig(kubeconfig_path):
...
+    Returns:                                                                                                                                                                                                                                      
+        bool: True if branch matches release-* pattern                                                                                                                                                                                            
+    """                                                                                                                                                                                                                                           
+    return re.match(r'^release-\d+\.\d+$', branch_name) is not None
+
+def auto_configure_acm():
+    """         
+    Auto-configure ACM as released or unreleased based on OCP version                                                                                                                              
+    Args:
+    ocp_version_str (str): OCP version string (e.g., "4.22")
```
petr-balogh: There is no such argument for this function

ocs_ci/utility/utils.py
```diff
@@ -6769,3 +6769,84 @@ def create_kubeconfig(kubeconfig_path):
...
+    Args:
+    ocp_version_str (str): OCP version string (e.g., "4.22")                                                   
+    """
+    # If explicitly configured by user, skip auto-configuration
+    if config.ENV_DATA.get("acm_hub_unreleased") is not None:                                                                        
+        log.info(f"ACM explicitly configured: acm_hub_unreleased={config.ENV_DATA['acm_hub_unreleased']}")
+        return 
+    # Get OCP version                                                                                                                
+    ocp_version = version_module.get_semantic_ocp_version_from_config()
+    ocp_version_str = f"{ocp_version.major}.{ocp_version.minor}"
```
petr-balogh: We don't need this as I commented in function `auto_configure_submariner`

ocs_ci/utility/utils.py
```diff
@@ -6769,3 +6769,84 @@ def create_kubeconfig(kubeconfig_path):
...
+    if ocp_version_str in defaults.ocp_to_acm_unreleased_mapping:
+        log.info(f"OCP {ocp_version_str} → ACM: {'UNRELEASED' if acm_unreleased else 'RELEASED'}")
+    else:
+        log.info(f"OCP {ocp_version_str} not in mapping → ACM: UNRELEASED (safe default)")
+
+def auto_configure_submariner():
+    """         
+    Auto-configure Submariner as released or unreleased based on OCP version.
+    Args:
+    ocp_version_str (str): OCP version string (e.g., "4.22")
```
petr-balogh: This function doesn't take any arg

ocs_ci/framework/conf/ocp_version/ocp-4.21-config.yaml
```diff
@@ -13,7 +13,8 @@ ENV_DATA:
   acm_hub_channel: "release-2.16"
   acm_version: "2.16"
   mce_version: "2.11"
-  submariner_version: "0.23.1"
+  submariner_version: "0.22.1"
+  submariner_image: "quay.io/redhat-user-workloads/submariner-tenant/submariner-fbc-4-21@sha256:14b379cf94a7ee452a381055c35a58770068d659e263a98e67ff4e54250738cf"
   # Below values needs to be changed when acm and submariner are GAed
   submariner_source: "downstream"
```
petr-balogh: @prsurve or @sidhant-agrawal , I think we can remove also `submariner_source` from ocp config files - and have similar logic to have by default downstream in .get("submariner_source", "downstream") - on places where we are getting this value - this will allow us to have by default upstream, but in case we need to overwrite this - we will be able via config file or YAML_TEXT_CONFIG param in jenkins job.

Right?

ocs_ci/framework/conf/ocp_version/ocp-4.21-config.yaml
```diff
@@ -13,7 +13,8 @@ ENV_DATA:
   acm_hub_channel: "release-2.16"
   acm_version: "2.16"
   mce_version: "2.11"
-  submariner_version: "0.23.1"
+  submariner_version: "0.22.1"
+  submariner_image: "quay.io/redhat-user-workloads/submariner-tenant/submariner-fbc-4-21@sha256:14b379cf94a7ee452a381055c35a58770068d659e263a98e67ff4e54250738cf"
```
petr-balogh: Also I think we should not have this `submariner_image`  in  4.21 - are you rebased on latest master @deepanjanacharya ?

ocs_ci/utility/utils.py
```diff
@@ -6976,43 +6728,123 @@ def create_kubeconfig(kubeconfig_path):
...
-            "Kubeconfig doesn't exists and RUN['kubeadmin_password'] and RUN['ocp_url'] "
-            "environment variables were not provided."
-        )
+        else:
+            raise ConfigurationError(
+                "Kubeconfig doesn't exists and RUN['kubeadmin_password'] and RUN['ocp_url'] "
+                "environment variables were not provided."
+            )
+
+def get_current_git_branch():
```
petr-balogh: I think you can remove this function

ocs_ci/utility/utils.py
```diff
@@ -6976,43 +6728,123 @@ def create_kubeconfig(kubeconfig_path):
...
+        log.warning(f"Could not detect git branch: {e}")                                                                                                                                                                                       
+        # Fallback: check environment variable set by Jenkins                                                                                                                                                                                     
+        branch = config.ENV_DATA.get('ocs_ci_branch', 'master')                                                                                                                                                                                   
+        log.info(f"Using branch from ENV_DATA: {branch}")                                                                                                                                                                                      
+        return branch
+
+def is_release_branch(branch_name):
```
petr-balogh: I think you can remove this function

ocs_ci/deployment/deployment.py
```diff
@@ -2503,27 +2448,33 @@ def acm_operator_installed(self):
...
+        """                                                                                                                                                                                                                                           
+        Deploy ACM hub cluster                                                                                                                                                                                                                        
+        """                                                                                                                                                                                                                                           
+        # Auto-configure ACM and Submariner based on based on OCP version if not explicitly set
+        from ocs_ci.utility.utils import auto_configure_acm, auto_configure_submariner                                                                                                                                                                               
+        auto_configure_acm() 
+        auto_configure_submariner()
```
petr-balogh: Also not sure it will be enough to call this here only. 

Look at other places where e.g. we instantiate Submariner class:

```
ocs_ci/ocs/utils.py:                        submariner = Submariner()
tests/functional/disaster-recovery/regional-dr/conftest.py:        submariner = acm.Submariner()
```

So there are some function called for download binary and so on which also depend on properly loaded values even without deployment - just for test execution.

So we will need to make sure we are calling this post loading config files I think - so it will make sure we have properly loaded values in config for all the cases - not just for deployment.


**And one more important point - we need to do this for all the cluster configs - for every cluster config need to update this config values to be in config of all clusters.**

So call of those two functions:
```
        auto_configure_acm() 
        auto_configure_submariner()   
```
I would move right after this line here:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/framework/pytest_customization/ocscilib.py#L424

ocs_ci/framework/conf/ocp_version/ocp-4.21-config.yaml
deepanjanacharya: Done rebase.

ocs_ci/utility/utils.py
deepanjanacharya: Removed function

ocs_ci/utility/utils.py
deepanjanacharya: Removed function

ocs_ci/deployment/deployment.py
deepanjanacharya: Done the changes.

ocs_ci/deployment/deployment.py
```diff
@@ -2503,27 +2503,30 @@ def acm_operator_installed(self):
...
-            ):
-                self.deploy_acm_hub_unreleased_konflux()
-                self.deploy_multicluster_hub()
-            else:
-                self.deploy_acm_hub_unreleased()
-        else:
-            self.deploy_acm_hub_released()
-            self.deploy_multicluster_hub()
-        if config.ENV_DATA.get("configure_acm_to_import_mce"):
+    def deploy_acm_hub(self):
```
petr-balogh: I am curious why we see this function as changed - I believe there is now no require change needed - so you should probably not touch this code at all.  Or what is the change here? 

If no change required - probably make this file untouched so it will not be visible in diff at all.

As I see just some changes in new lines and in docstring  , but you don't need to touch this I think

ocs_ci/framework/conf/ocp_version/ocp-4.22-config.yaml
```diff
@@ -9,20 +9,12 @@ DEPLOYMENT:
   # https://docs.openshift.com/container-platform/4.7/post_installation_configuration/machine-configuration-tasks.html#machine-config-overview-post-install-machine-configuration-tasks
   ignition_version: "3.2.0" # TODO: We might want to update to 3.5 or later: https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/release_notes/ocp-4-19-release-notes
 ENV_DATA:
-  # Dictionary of available RHCOS templates by major version
-  # To use RHCOS 10, set rhcos_version: "10" in your custom config
-  vm_templates:
```
petr-balogh, line 14: This doesn't look like properly resolved conflict as you removed those lines and as you see versions of acm and mce also are changed in your PR which I guess you should not touch.


Only change by you in this file should be removed those lines:
```
submariner_source: "downstream"
submariner_release_type: "released"
acm_hub_unreleased: true
```
But if we remove :
`submariner_source: "downstream"`
We need to handle this in code as well to have this as default value loaded in config I think.

deepanjanacharya: handled in submariner function

ocs_ci/deployment/deployment.py
deepanjanacharya: fixed the extra stuff

petr-balogh (dismissed):

ocs_ci/framework/conf/ocp_version/ocp-4.21-config.yaml
sidhant-agrawal: @petr-balogh I think you meant `default downstream` not `default upstream` in your last sentence.
yes, we can remove `submariner_source` from ocp config files since it's now auto-configured and defaults to `downstream` if not set.

sidhant-agrawal (dismissed):

prsurve (dismissed):

shylesh (dismissed):
lgtm

deepanjanacharya:
Validation Checklist as suggested by [Sidhant.Agrawal@ibm.com]
**Case 1:**
Default ACM and Submariner for OCP 4.22

- [done]   Rebased ocs-ci on latest master
- [done]   Run DR deployment with ACM=blank(default)
- [done]   Run DR deployment with ACM=blank(default)

Jenkins build SUCCESS

Build:
https://jenkins-csb-odf-qe-stage.dno.corp.redhat.com/view/DR/job/qe-rdr-setup/282/console


**Case 2:**

ACM=released for OCP  4.21

- [done]   Rebased ocs-ci on latest master
- [done]   Run DR deployment with ACM=blank(default)
- [done]   Run DR deployment with ACM=blank(default)

Jenkins build SUCCESS

Build:
https://jenkins-csb-odf-qe-stage.dno.corp.redhat.com/view/DR/job/qe-rdr-setup/284/console
Remaining:

RDR 4.22 deployment with user provided values (released/downstream)

ocs_ci/framework/pytest_customization/ocscilib.py
```diff
@@ -422,6 +423,8 @@ def pytest_configure(config):
 
         if not (config.getoption("--help") or config.getoption("collectonly")):
             process_cluster_cli_params(config)
+            auto_configure_acm()
+            auto_configure_submariner()
```
shylesh, line 427: @petr-balogh I was having discussion with @clacroix12 on having these auto_configure_* in this pytest hook, Coady suggested that log object at this point in execution won't be instantiated with proper handler, so either we have to instantiate handler locally so that log.info() message gets printed else if don't have any compusion of having auto_configure_* methods in this hook , can we move it to a session scoped autouse fixture which gets called before any test gets executed. WDYT?

petr-balogh: Here we have broader issue which should not block this PR - if logic works let's keep it here and do not block the PR and we can have separate discussion what other impact it will have to call it in different place - this pytest_configure hook is for a configuration - as in other places we rely on those values - this needs to happen as soon as possible for all cluster configurations and here we actually modifying config values we want them also to be dumped in the output file to see them reflected as part of dump_config_to_file which happens on line 439 here.

With that said moving that logic somewhere else will not reflect real config values dumped to file if we move to some session scope fixture.

So if only logging is the issue I don't want to move code in this PR and we should discuss this maybe in ecosystem open hours what are the possibilities here.

petr-balogh (approved):

petr-balogh (approved):

vavuthu (approved):

PR #14754, opened 2026-03-24
storage monitoring for FDF

DanielOsypenko:
Add `openshift.io/cluster-monitoring=true` label to storage namespace in FDF deployment                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                 
  In the FDF deployment flow, the openshift-storage namespace (or custom cluster_namespace) is created implicitly by the Fusion operator — unlike the standard ODF flow where it is created from deploy-with-olm.yaml which already carries the monitoring label.                
                                                                                                                                                                                                                                                                                 
  This change applies the `openshift.io/cluster-monitoring=true` label at the start of setup_storage(), after FDF installation is verified and the namespace exists, but before the StorageCluster CR is created. This ensures Prometheus picks up ServiceMonitor and  PrometheusRule objects in the storage namespace from the moment they appear.

### DIFF
--- a/ocs_ci/deployment/fusion_data_foundation.py
+++ b/ocs_ci/deployment/fusion_data_foundation.py
@@ -241,6 +241,14 @@ class FusionDataFoundationDeployment:
         Setup storage
         """
         logger.info("Configuring storage.")
+        storage_namespace = config.ENV_DATA.get(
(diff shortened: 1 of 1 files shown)


DanielOsypenko:
I verified the changes manually: adding label `openshift.io/cluster-monitoring=true` always enables monitoring

dahorak (approved):

PR #14765, opened 2026-03-25
Fix test_node_drain_and_fault_tolerance_for_multiple_mds that caused cascading test failures on IBM Cloud 3-worker clusters.

amr1ta:
Fix test_node_drain_and_fault_tolerance_for_multiple_mds that caused cascading test failures on IBM Cloud 3-worker clusters.

 Root Cause (from log analysis)

  - Item 2042165: oc adm drain timed out after 1800s because a pre-existing unschedulable node + the drain-cordoned node left only 1 CPU-constrained node. PDB requirements for noobaa-endpoint,
  ocs-client-operator-console, etc. could not be satisfied. The node stayed cordoned.
  - Item 2042166: Auto-retried run inherited the cordoned node. Drain succeeded (2 schedulable nodes now), but teardown's adjust_active_mds_count(1) rolled the MDS-a ReplicaSet. Teardown then described the deleted
  pod by name → NotFound.

Fix: 
Wrap drain_nodes in try/finally to guarantee schedule_nodes always runs, even if drain fails or times out.
Replace the stale pod-object iteration with OCP.wait_for_resource(selector="app=rook-ceph-mds", resource_count=2). This waits for 2 Running MDS pods (1 active + 1 standby-replay) by label selector, so it automatically picks up newly created pods regardless of RS rollover.

### DIFF
--- a/tests/functional/z_cluster/test_multiple_mds.py
+++ b/tests/functional/z_cluster/test_multiple_mds.py
@@ -14,7 +14,6 @@ from ocs_ci.framework.pytest_customization.marks import (
     skipif_hci_client,
 )
 from ocs_ci.framework import config
-from ocs_ci.helpers import helpers
 from ocs_ci.ocs.cluster import (
     adjust_active_mds_count_storagecluster,
     get_active_mds_count_cephfilesystem,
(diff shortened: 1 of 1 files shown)


tests/functional/z_cluster/test_multiple_mds.py
```diff
@@ -147,9 +153,12 @@ def test_node_drain_and_fault_tolerance_for_multiple_mds(self, pod_factory):
...
         log.info("Drain active mds pod running node")
-        node.drain_nodes([active_mds_node_name])
-        # Make the node schedulable again
-        node.schedule_nodes([active_mds_node_name])
+        try:
+            node.drain_nodes([active_mds_node_name])
+        finally:
+            # Always uncordon, even if drain fails or times out, to prevent
+            # the node remaining unschedulable and affecting subsequent tests.
+            node.schedule_nodes([active_mds_node_name])
```
yitzhak12, line 161: Better to run this code in the teardown if it doesn't relate to the test.

Issue #14562, opened 2026-03-03
[Automation_issue]: Updating node maintenance test cases which needs to be skipped for client clusters


amr1ta:
### Description
test_pdb_check_simultaneous_node_drains and test_pdb_check_simultaneous_node_drains

### Steps to Reproduce
1.
2.
3.

### Environment
- Test Suite(s):
- Platform(s):
- Version(s):
- OS:

amr1ta:
fix, https://github.com/red-hat-storage/ocs-ci/pull/14766

amr1ta:
fixed pr is merged closing this issue

PR #14766, opened 2026-03-25
Fix skipif_hci_provider_and_client eager evaluation by using a fixture

amr1ta:
pytest.mark.skipif conditions are evaluated at module import time, before config.clusters is populated. This means config.hci_provider_exist() and config.hci_client_exist() always return False, causing the skip condition to freeze to False and the tests to always run on HCI provider+client setups.

Replace the @skipif_hci_provider_and_client decorator on test_node_maintenance_restart_activate and test_pdb_check_simultaneous_node_drains with a skip_on_hci_provider_client fixture that performs the same check at runtime (fixture setup time), when config.clusters is fully populated.

### DIFF
--- a/tests/functional/z_cluster/nodes/test_nodes_maintenance.py
+++ b/tests/functional/z_cluster/nodes/test_nodes_maintenance.py
@@ -131,22 +131,6 @@ class TestNodesMaintenance(ManageTest):
             # skip because ceph is not in good health
             pytest.skip(str(e))
 
-    @pytest.fixture()
-    def skip_on_hci_provider_client(self):
-        """
-        Skip the test at runtime when running on Fusion HCI provider+client
-        setup. Uses a fixture instead of pytest.mark.skipif to avoid the
-        eager-evaluation issue where config.clusters is not yet populated at
-        module import time.
-        """
-        if (
-            config.ENV_DATA["platform"].lower()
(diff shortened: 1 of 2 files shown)


amr1ta:
fix for https://github.com/red-hat-storage/ocs-ci/issues/14562

ebenahar (dismissed):

tests/functional/z_cluster/nodes/test_nodes_maintenance.py
```diff
@@ -132,6 +131,22 @@ def health_checker(self):
             # skip because ceph is not in good health
             pytest.skip(str(e))
 
+    @pytest.fixture()
+    def skip_on_hci_provider_client(self):
```
yitzhak12, line 135: I think this can be moved to `tests/conftest.py`.

suchita-g: +1

amr1ta: addressed

yitzhak12 (dismissed):

ebenahar (dismissed):

AviadP (approved):

ebenahar (approved):

Issue #14772, opened 2026-03-25
[Bug]: Missing list of packages for disconnected 4.22 deployment


dahorak:
### Description
Deployment of ODF 4.22 in disconnected mode is failing, because `const.DISCON_CL_REQUIRED_PACKAGES_PER_ODF_VERSION` dictionary is not updated for version 4.22.

### Steps to Reproduce
1. Run disconnected mode deployment.

### Actual Behavior
Failed deployment with error:
```
2026-03-25 17:36:39              mirrored_index_image = mirror_index_image_via_oc_mirror(
2026-03-25 17:36:39                  index_image,
2026-03-25 17:36:39  >               constants.DISCON_CL_REQUIRED_PACKAGES_PER_ODF_VERSION[f"{ocs_version}"],
2026-03-25 17:36:39                  idms=idms,
2026-03-25 17:36:39              )
2026-03-25 17:36:39  E           KeyError: '4.22'
2026-03-25 17:36:39  
2026-03-25 17:36:39  ocs_ci/deployment/disconnected.py:452: KeyError
```

### Expected Behavior
Disconnected deployment should work.

### Impact (likelihood of reproduction, impact on the cluster and on other tests, etc.)
100%

### Environment
- Platform(s): vSphere
- Version(s): 4.22

### Additional Context
https://url.corp.redhat.com/5e44281

PR #14773, opened 2026-03-25
add list of 4.22 packages for disconnected deployment

dahorak:
fixes: https://github.com/red-hat-storage/ocs-ci/issues/14772


dahorak:
There is one additional operator `ceph-volsync-plugin-operator` which wasn't present in 4.21. Based on [following discussion](https://url.corp.redhat.com/2131e4e), this operator is needed only for DR, but we might still want to mirror it as well (see also [RHSTOR-8719](https://url.corp.redhat.com/f2d20be)).

vavuthu (approved):

PR #14774, opened 2026-03-25
Adding a test for iam user actions

ypersky1980:
This PR contains a happy path scenario for IAM user actions: 

The scenario is as following:
        1. Create several users and verify that they are created
        2. Choose a random user, update its path and verify that the update worked
        3. Choose a random user, create, update and delete its access key, verify that everything works as expected
        4. Choose a random user, create multiple access keys and verify that the quota cannot be exceeded
        5. Choose a random user, tag/untag it and verify that works as expected
        6. Delete all the created users

### DIFF
--- a/tests/functional/object/mcg/test_iam_users.py
+++ b/tests/functional/object/mcg/test_iam_users.py
@@ -141,13 +141,20 @@ class TestIAMUsers(MCGTest):
             f"Two access keys for the user {user_for_multiple_access_keys} created successfully"
         )
 
+        quota_enforced = False
         try:
             run_iam_command(mcg_obj, awscli_pod_session, create_access_key_cmd)
         except CommandFailed as ex:
             if "Cannot exceed quota for AccessKeysPerUser" in str(ex):
                 logger.info("As expected, cannot create access key above the quota (2)")
+                quota_enforced = True
             else:
                 raise ex
+
+        assert quota_enforced, (
+            f"Expected CommandFailed when exceeding access key quota for {user_for_multiple_access_keys}, "
+            "but command succeeded"
+        )
 
         # Test tag-user, list-user-tags, untag-users commands
         user_for_tags = random.choice(new_users_list)
--- a/ocs_ci/utility/iam_utils.py
+++ b/ocs_ci/utility/iam_utils.py
@@ -41,7 +41,7 @@ def run_iam_command(mcg_obj, awscli_pod_session, cmd):
 
 def generate_random_iam_path(levels=3, length=8):
     """
-        Generates random path of iam user
+    Generates random path of iam user
 
     Args:
         levels (int): Number of levels (nesting) of the path
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,7 @@ import logging
 import os
 import pandas as pd
 import random
+import secrets
 import string
 import time
 import tempfile
@@ -11781,7 +11782,7 @@ def iam_users_factory_fixture(request, mcg_obj, awscli_pod_session):
         """
         for _ in range(num):
             random_name_part = "".join(
-                random.choices(string.ascii_lowercase + string.digits, k=8)
+                secrets.choice(string.ascii_lowercase + string.digits) for _ in range(8)
             )
             new_user_name = f"iam_user_{random_name_part}"
             new_user_path = generate_random_iam_path()
@@ -11799,9 +11800,8 @@ def iam_users_factory_fixture(request, mcg_obj, awscli_pod_session):
         """
         for user_name in created_iam_users_name_list:
             access_keys = get_user_access_keys(mcg_obj, awscli_pod_session, user_name)
-            for (
-                key
-            ) in access_keys:  # access keys should be deleted before the user deletion
+            # Access keys should be deleted before the user deletion
+            for key in access_keys:
                 access_key_id = key["AccessKeyId"]
                 delete_key_cmd = f"delete-access-key --user-name {user_name} --access-key-id {access_key_id}"
                 run_iam_command(mcg_obj, awscli_pod_session, delete_key_cmd)


tests/functional/object/mcg/test_iam_users.py
```diff
@@ -0,0 +1,307 @@
...
+    # iam_host of the iam is the endpoint of the iam command
+    routes = run_oc_command(f"get routes -n {config.ENV_DATA['cluster_namespace']}")
+
+    iam_host = None
+    for line in routes[1:]:  # skip header
+        parts = line.split()
+        if parts[0] == "iam":
+            iam_host = parts[1]
+            break
+    endpoint = "https://" + iam_host
```
udaysk23: I think we should write separate method in utils to get IAM route info so that other tests can also access the same.

tests/functional/object/mcg/test_iam_users.py
```diff
@@ -0,0 +1,307 @@
...
+    """
+
+    list_access_key_cmd = f"list-access-keys --user-name {user_name}"
+    list_access_key_result = run_iam_command(
+        mcg_obj, awscli_pod_session, list_access_key_cmd
+    )
+    return list_access_key_result.get("AccessKeyMetadata", [])
+
+
+@tier1
```
udaysk23: We should avoid using `tier 1` and it should go with tier 2

tests/functional/object/mcg/test_iam_users.py
sagihirshfeld: +1

We can put it in the MCG class, similar to how we're already storing the s3 and sts endpoints:
https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/resources/mcg.py#L131-L160

Apparently this route is also listed on the NooBaa CR's yaml, similar to the other ones, so you can even use a similar syntax.

In bash it's:
```
❯ oc get noobaa noobaa -o json | jq .status.services.serviceIam.externalDNS
[
  "https://iam-openshift-storage.apps.<CENSORED>.com:443",
  "https://<CENSORED>.appdomain.cloud:443"
]
```

So in Python it should be:
```
            self.iam_endpoint = (
                get_noobaa.get("items")[0]
                .get("status")
                .get("services")
                .get("serviceIam.")
                .get("externalDNS")[0]
            )
```

tests/functional/object/mcg/test_iam_users.py
```diff
@@ -0,0 +1,307 @@
...
+                random.choices(string.ascii_lowercase + string.digits, k=8)
+            )
+            new_user_name = f"iam_user_{random_name_part}"
+            new_user_path = generate_random_iam_path()
+            create_user_cmd = (
+                f"create-user --user-name {new_user_name} --path {new_user_path}"
+            )
+            run_iam_command(mcg_obj, awscli_pod_session, create_user_cmd)
+            logger.info(f"User {new_user_name} created")
+            new_users_list.append(new_user_name)
```
sagihirshfeld: Please move this creation logic and the deletion of the iam users to a factory fixture under [conftest.py](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/conftest.py). This change has two benefits:

1. Reusability in other tests 
2. It guarantees the deletion of the IAM users. Currently they might be let as leftovers if the test fails for any reason after this part by before their deletion.

tests/functional/object/mcg/test_iam_users.py
```diff
@@ -0,0 +1,307 @@
...
+    Returns:
+        dict: user-access-keys command result
+    """
+
+    list_access_key_cmd = f"list-access-keys --user-name {user_name}"
+    list_access_key_result = run_iam_command(
+        mcg_obj, awscli_pod_session, list_access_key_cmd
+    )
+    return list_access_key_result.get("AccessKeyMetadata", [])
+
```
sagihirshfeld: For reusability in other tests (and the suggested fixture for creating/deletion iam users), please move these functions to a different files. A new file named "iam_utils.py" woud be the cleanest solution IMO.

For clarity, I'm referring to `get_base_iam_command`, `run_iam_command`, `generate_random_iam_path` and `get_user_access_keys`.

tests/functional/object/mcg/test_iam_users.py
```diff
@@ -0,0 +1,307 @@
...
+            access_keys = get_user_access_keys(mcg_obj, awscli_pod_session, user_name)
+            for (
+                key
+            ) in access_keys:  # access keys should be deleted before the user deletion
+                access_key_id = key["AccessKeyId"]
+                delete_key_cmd = f"delete-access-key --user-name {user_name} --access-key-id {access_key_id}"
+                run_iam_command(mcg_obj, awscli_pod_session, delete_key_cmd)
+
+            delete_user_cmd = f"delete-user --user-name {user_name}"
+            run_iam_command(mcg_obj, awscli_pod_session, delete_user_cmd)
```
sagihirshfeld: The deletion logic of the IAM users should be managed in the factory fixture I suggested in a previous comment. This should guarantee their deletion in case the test fails between the creation and the deletion.

tests/functional/object/mcg/test_iam_users.py
ypersky1980: Following Sagi's suggestion, I've implemented iam_endpoint property in mcg.py

tests/functional/object/mcg/test_iam_users.py
ypersky1980: Following our conversation, moved to factory fixture. Thanks

tests/functional/object/mcg/test_iam_users.py
ypersky1980: Implemented as a finalizer of factory fixture

tests/functional/object/mcg/test_iam_users.py
ypersky1980: Done. Thanks

tests/functional/object/mcg/test_iam_users.py
ypersky1980: Done. Thanks

tests/functional/object/mcg/test_iam_users.py
```diff
@@ -0,0 +1,307 @@
...
+from ocs_ci.framework.testlib import MCGTest
+from ocs_ci.helpers.performance_lib import run_oc_command
+from ocs_ci.framework import config
+from ocs_ci.ocs.exceptions import CommandFailed
+
+
+logger = logging.getLogger(__name__)
+
+
+@lru_cache(maxsize=1)
```
udaysk23: The `@lru_cache` here is dangerous. It caches based on `mcg_obj` identity, but if different MCG objects with different credentials are passed, you'll get stale cached results.

**Suggested fix:**
```python
@lru_cache(maxsize=1)
def _get_iam_host():
    """Get IAM route hostname from cluster"""
    routes = run_oc_command(f"get routes -n {config.ENV_DATA['cluster_namespace']}")
    for line in routes[1:]:
        parts = line.split()
        if len(parts) >= 2 and parts[0] == "iam":
            return parts[1]
    raise ValueError("IAM route not found in cluster")

def get_base_iam_command(mcg_obj):
    """Builds base IAM command with credentials"""
    iam_host = _get_iam_host()
    endpoint = f"https://{iam_host}"
    
    return (
        f'sh -c "'
        f"AWS_ACCESS_KEY_ID={mcg_obj.access_key_id} "
        f"AWS_SECRET_ACCESS_KEY={mcg_obj.access_key} "
        f"{f'AWS_DEFAULT_REGION={mcg_obj.region} ' if mcg_obj.region else ''}"
        f"aws --endpoint={endpoint} iam --no-verify-ssl "
    )```


This way only the route lookup is cached, not the credentials.

File: tests/functional/object/mcg/test_iam_users.py [hunk 7]
```diff
@@ -0,0 +1,307 @@
...
+
+    # iam_host of the iam is the endpoint of the iam command
+    routes = run_oc_command(f"get routes -n {config.ENV_DATA['cluster_namespace']}")
+
+    iam_host = None
+    for line in routes[1:]:  # skip header
+        parts = line.split()
+        if parts[0] == "iam":
+            iam_host = parts[1]
+            break
```
udaysk23: `iam_host` could be `None` if the route isn't found, leading to `endpoint = "https://None"`.

**Suggested fix:**
```python
iam_host = None
for line in routes[1:]:
    parts = line.split()
    if len(parts) >= 2 and parts[0] == "iam":
        iam_host = parts[1]
        break

if not iam_host:
    raise ValueError("IAM route not found in cluster")

endpoint = f"https://{iam_host}"
This will fail fast with a clear error message instead of creating an invalid endpoint.

File: tests/functional/object/mcg/test_iam_users.py [hunk 8]
```diff
@@ -0,0 +1,307 @@
...
+        Args:
+            mcg_obj: An MCG class instance
+            awscli_pod_session (pod): A pod running the AWSCLI tools
+
+        """
+        new_users_num = 3
+        new_users_list = []
+
+        # Test create-user and list-users command
+        for _ in range(new_users_num):
```
udaysk23: You're using `random.choices` here but `secrets.choice` elsewhere (e.g., line 79). 

**Suggestion:** Use `secrets` consistently for better practice:
```python
random_name_part = "".join(
    secrets.choice(string.ascii_lowercase + string.digits) for _ in range(8)
)
While this is just test data, consistency is good and secrets is the modern approach.

File: tests/functional/object/mcg/test_iam_users.py [hunk 9]
```diff
@@ -0,0 +1,307 @@
...
+        user_for_multiple_access_keys = random.choice(new_users_list)
+        create_access_key_cmd = (
+            f"create-access-key --user-name {user_for_multiple_access_keys}"
+        )
+        run_iam_command(mcg_obj, awscli_pod_session, create_access_key_cmd)
+        run_iam_command(mcg_obj, awscli_pod_session, create_access_key_cmd)
+        access_keys = get_user_access_keys(
+            mcg_obj, awscli_pod_session, user_for_multiple_access_keys
+        )
+        assert len(access_keys) == 2, (
```
udaysk23, line 136: If the quota is NOT enforced (command succeeds), the test will silently pass. You need to assert that the exception was actually raised.

**Suggested fix:**
```python
quota_enforced = False
try:
    run_iam_command(mcg_obj, awscli_pod_session, create_access_key_cmd)
except CommandFailed as ex:
    if "Cannot exceed quota for AccessKeysPerUser" in str(ex):
        logger.info("As expected, cannot create access key above the quota (2)")
        quota_enforced = True
    else:
        raise ex

assert quota_enforced, (
    f"Expected CommandFailed when exceeding access key quota for {user_for_multiple_access_keys}, "
    "but command succeeded"
)

File: tests/functional/object/mcg/test_iam_users.py [hunk 10]
```diff
@@ -0,0 +1,307 @@
...
+        )
+        user_tags_dict = {
+            t["Key"]: t["Value"] for t in list_user_tags_result.get("Tags", [])
+        }
+
+        # assert the deleted tag is absent
+        assert (
+            first_key_name not in user_tags_dict
+        ), f"Tag {first_key_name} should be absent but is present"
+        logger.info("User tags work as expected")
```
udaysk23, line 200: If deletion fails for one user, the loop will break and remaining users won't be cleaned up, polluting the test environment.

**Suggested fix:**
```python
cleanup_errors = []
for user_name in new_users_list:
    try:
        # Delete access keys first
        access_keys = get_user_access_keys(mcg_obj, awscli_pod_session, user_name)
        for key in access_keys:
            access_key_id = key["AccessKeyId"]
            delete_key_cmd = f"delete-access-key --user-name {user_name} --access-key-id {access_key_id}"
            run_iam_command(mcg_obj, awscli_pod_session, delete_key_cmd)

        # Delete the user
        delete_user_cmd = f"delete-user --user-name {user_name}"
        run_iam_command(mcg_obj, awscli_pod_session, delete_user_cmd)
        logger.info(f"Successfully cleaned up user {user_name}")
    except Exception as e:
        error_msg = f"Failed to cleanup {user_name}: {e}"
        cleanup_errors.append(error_msg)
        logger.warning(error_msg)

if cleanup_errors:
    logger.warning(f"Cleanup completed with {len(cleanup_errors)} errors")
This ensures all users are attempted for cleanup even if some fail.

File: tests/functional/object/mcg/test_iam_users.py [hunk 11]
```diff
@@ -0,0 +1,307 @@
...
+
+    Returns:
+        dict: command result
+    """
+
+    full_command = get_base_iam_command(mcg_obj) + cmd + '"'
+    return awscli_pod_session.exec_cmd_on_pod(full_command)
+
+
+def generate_random_iam_path(levels=3, length=8):
```
udaysk23: Extra indentation on the first line of the docstring.

**Fix:**
```python
def generate_random_iam_path(levels=3, length=8):
    """
    Generates random path of iam user

File: tests/functional/object/mcg/test_iam_users.py [hunk 12]
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        # assert the deleted tag is absent
+        assert (
+            first_key_name not in user_tags_dict
+        ), f"Tag {first_key_name} should be absent but is present"
+        logger.info("User tags work as expected")
+
+        # Now delete all the created users to restore the state before the test
+        for user_name in new_users_list:
```
udaysk23: This parenthesis wrapping is awkward. Just use a normal loop with the comment above:

```python
# Access keys must be deleted before the user can be deleted
for key in access_keys:
    access_key_id = key["AccessKeyId"]

tests/functional/object/mcg/test_iam_users.py
ypersky1980: @udaysk23 I've noticed that I did not push all the fixes, did that now.

tests/functional/object/mcg/test_iam_users.py
ypersky1980: Fixed. Thanks

tests/functional/object/mcg/test_iam_users.py
ypersky1980: Done. Thanks

tests/functional/object/mcg/test_iam_users.py
ypersky1980: Now the users are deleted as a part of factory finalizer and teardown, this should solve the issue.

tests/functional/object/mcg/test_iam_users.py
ypersky1980: Done. Thanks

tests/functional/object/mcg/test_iam_users.py
ypersky1980: Fixed. Thanks

tests/functional/object/mcg/test_iam_users.py
ypersky1980: This code is obsolete; please look at the latest version at mcg.py

PR #14831, opened 2026-04-06
Fix NFS mount failure on IBM Cloud when client VM is in a different V…

amr1ta:
…PC (#14676)

* Fix NFS mount failure on IBM Cloud when client VM is in a different VPC

When the NFS LoadBalancer service is recreated, IBM Cloud assigns a new VPC LB hostname (*.lb.appdomain.cloud) that is only resolvable from within the cluster VPC. NFS client VMs in a different VPC cannot resolve this hostname via their DNS servers, causing mounts to hang indefinitely.

Add update_etc_hosts_on_nfs_client() to nfs_utils.py that resolves the LB hostname by exec-ing getent on the node running the NFS pod, then updates /etc/hosts on the client VM with the fresh IP mapping. Call this from get_nfs_client_connection() whenever the NFS endpoint is a hostname (IBM Cloud) rather than a raw IP (AWS). Stale entries are removed before writing so repeated redeployments remain safe.

* Improve NFS LB service create/delete reliability on IBM Cloud

- Replace fixed time.sleep(30) with TimeoutSampler polling loop (15s interval, 5min timeout) when waiting for LB ingress to be assigned after service creation
- Add retry loop with ignore_error=True when resolving LB hostname via oc debug node, to handle DNS propagation delay within the cluster
- Add existence check and wait_for_delete after deleting the LB service to ensure IBM Cloud VPC LB is fully deprovisioned before teardown continues
- Remove unused time import

* Address PR review: improve IP validation and decode error handling

- Replace hostname_add[0].isdigit() check with ipaddress.ip_address() for proper IP validation, correctly handling IPv4, IPv6 and hostnames
- Add import ipaddress to test file
- Wrap sample.stdout.decode() in try-except for UnicodeDecodeError and AttributeError in update_etc_hosts_on_nfs_client

* Refactor IP check to avoid exception-based flow control

Separate the ipaddress.ip_address() check from the update_etc_hosts_on_nfs_client() call to prevent potential .pyc cache issues where the except ValueError handler may not be present in stale bytecode. Also adds a log message when hostname resolution is triggered.

* Fix exec_cmd parameter name in update_etc_hosts_on_nfs_client

The exec_cmd function expects 'cmd' not 'command' as the parameter name. Using the wrong name caused hostname resolution via getent to silently fail.

* Handle TimeoutExpiredError gracefully in hostname resolution

Catch TimeoutExpiredError from TimeoutSampler when hostname resolution fails instead of letting it crash the test. Log a warning and skip the /etc/hosts update, allowing the test to continue.

* Fix RST docstring warning for wildcard in hostname pattern

---------

### DIFF
--- a/ocs_ci/utility/nfs_utils.py
+++ b/ocs_ci/utility/nfs_utils.py
@@ -5,16 +5,15 @@ Module that contains all operations related to nfs feature in a cluster
 
 import logging
 import yaml
-import time
 import pytest
-from ocs_ci.ocs import constants, resources
+from ocs_ci.ocs import constants, resources, ocp
 from ocs_ci.helpers import helpers
 from ocs_ci.ocs.resources import pod
 from ocs_ci.utility.retry import retry
-from ocs_ci.ocs.exceptions import CommandFailed
+from ocs_ci.ocs.exceptions import CommandFailed, TimeoutExpiredError
 from ocs_ci.framework import config
 from ocs_ci.utility import version as version_module
-from ocs_ci.utility.utils import convert_device_size
+from ocs_ci.utility.utils import convert_device_size, exec_cmd, TimeoutSampler
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


amr1ta:
verification job: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/67191/console

ebenahar (approved):

AviadP (approved):

PR #14857, opened 2026-04-09
Increase timeout for OCP IBM Cloud deployment on IBM Cloud

petr-balogh:
When we have setup with 6 Worker nodes we see sometime failure in OCP deployment, but when checked cluster later on it was OK and healthy.


vavuthu (approved):

petr-balogh:
Verified here:
https://url.corp.redhat.com/a8c401f

shivamdurgbuns (approved):

PR #14868, opened 2026-04-10
Prevent upgrade conflicts from unexpected catalog builds

petr-balogh:
During y-stream upgrades (e.g., 4.21 to 4.22), if the upgrade catalog contains an unreleased build of the current version (e.g., 4.21.2 when upgrading from 4.21.1), OLM detects it immediately after the catalog update and starts upgrading before the channel change occurs. This creates conflicting upgrade paths that are hard to debug.

This change adds validation to detect such scenarios:
- Before catalog update: capture current CSV from current channel
- After catalog update: wait for catalog ready, then re-check CSV
- If CSV changed: fail with clear error message identifying the issue

This prevents the race condition and provides clear diagnostics when the catalog contains unexpected builds.

### DIFF
--- a/ocs_ci/ocs/ocs_upgrade.py
+++ b/ocs_ci/ocs/ocs_upgrade.py
@@ -80,6 +80,8 @@ from ocs_ci.ocs.exceptions import (
     TimeoutException,
     ExternalClusterRGWAdminOpsUserException,
     CSVNotFound,
+    UnexpectedCatalogBuildException,
+    ChannelNotFound,
 )
 from ocs_ci.ocs.ui.base_ui import logger, login_ui
 from ocs_ci.ocs.ui.views import locators, ODF_OPERATOR
@@ -621,10 +623,7 @@ class OCSUpgrade(object):
             channel: (str): OCS subscription channel
 
         """
-        if version.get_semantic_ocs_version_from_config() >= version.VERSION_4_9:
-            subscription_name = constants.ODF_SUBSCRIPTION
-        else:
-            subscription_name = constants.OCS_SUBSCRIPTION
+        subscription_name = constants.ODF_SUBSCRIPTION
         kind_name = "subscription.operators.coreos.com"
         subscription = OCP(
             resource_name=subscription_name,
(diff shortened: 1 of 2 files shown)


petr-balogh:
Verification:
https://url.corp.redhat.com/3a61b2e

petr-balogh:
The latest build has already fix so job above should pass - and this one is with problematic build 4.22.0-48 which should fail:
https://url.corp.redhat.com/8b20e9a

If this will be case we can consider as verified

petr-balogh:
One of job failed as expected. The second failed on some other issue non related to this PR.
Re-triggered here:
https://url.corp.redhat.com/9aaa6af

Failure from one of the job:
```
ocs_ci.ocs.exceptions.UnexpectedCatalogBuildException: UPGRADE BLOCKED: Detected unexpected build in catalog. Current channel 'stable-4.21' had CSV 'ocs-operator.v4.21.1-rhodf' before catalog update, but now shows 'ocs-operator.v4.21.2-rhodf'. This would cause OLM to upgrade to 'ocs-operator.v4.21.2-rhodf' before the channel change, creating conflicting upgrade paths.
```

petr-balogh:
Z-stream upgrade verification triggered here:
https://url.corp.redhat.com/4857616

The rest of verification passed or failed as expected - so this run is just to make sure any regression caused for z-stream upgrade as well. If it will pass we can merge.

ocs_ci/ocs/ocs_upgrade.py
```diff
@@ -880,6 +887,67 @@ def run_ocs_upgrade(
...
+            # Y-stream upgrade: check for unexpected builds in current channel
+            log.info(
+                f"Y-stream upgrade ({original_ocs_version} -> {upgrade_version}). "
+                "Checking for unexpected builds in current channel."
+            )
+
+            if version.get_semantic_ocs_version_from_config() >= version.VERSION_4_9:
+                subscription_name = constants.ODF_SUBSCRIPTION
+            else:
+                subscription_name = constants.OCS_SUBSCRIPTION
```
dahorak: This condition is probably not necessary any more (we don't need to support versions older than 4.9).

dahorak (dismissed):
Overall LGTM, just one unnecessary condition related to old unsupported OCS/ODF versions.

dahorak (approved):

vavuthu (approved):

PR #14876, opened 2026-04-12
RHSTOR-8248-Happypath

AviadP:
- [x] Add generic csi-addons-config ConfigMap helpers (get/set any key, wait-for-ready after controller restart)
- [x] Add happy path test verifying UID-based stagger offsets for ReclaimSpaceCronJobs (8 steps: create PVCs, verify CronJob   schedule unchanged, compare Job timestamps for spread, disable stagger and verify survival)
- [x] Set `schedule-precedence: storageclass` in fixture setup so the controller reads SC annotations
- [x] Refactor `set_schedule_precedence` and get_schedule_precedance_value_from_csi_addons_configmap` in helpers.py to thin wrappers delegating to the new generic helpers 

Callers to validate (refactored helpers.py functions):
  - [ ] tests/functional/storageclass/test_enforce_storageclass_precedance_for_KRRS.py (set_schedule_precedence)
  - [ ] tests/functional/pv/pv_encryption/test_disable_pv_keyrotation.py (both functions)

### DIFF
--- a/tests/functional/pv/space_reclaim/test_staggered_cronjob_scheduling.py
+++ b/tests/functional/pv/space_reclaim/test_staggered_cronjob_scheduling.py
@@ -68,9 +68,7 @@ class TestStaggeredCronjobScheduling(ManageTest):
         )
 
         def finalizer():
-            logger.info(
-                "Restoring stagger window to: %s", self.original_stagger_window
-            )
+            logger.info("Restoring stagger window to: %s", self.original_stagger_window)
             update_csi_addons_config(
                 CRONJOB_STAGGER_WINDOW_KEY, self.original_stagger_window
             )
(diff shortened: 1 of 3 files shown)


yitzhak12 (dismissed):

tests/functional/pv/space_reclaim/test_staggered_cronjob_scheduling.py
```diff
@@ -0,0 +1,322 @@
...
+        update_csi_addons_config(
+            SCHEDULE_PRECEDENCE_KEY, SCHEDULE_PRECEDENCE_STORAGECLASS
+        )
+
+        def finalizer():
+            logger.info("Restoring stagger window to: %s", self.original_stagger_window)
+            update_csi_addons_config(
+                CRONJOB_STAGGER_WINDOW_KEY, self.original_stagger_window
+            )
+            if self.original_precedence:
```
ebenahar, line 77: if original_precedence is an empty string then this condition will not be met and the configMap will never be restored to its original values.

from Claude:

```
  The problem:

  # setup: original_precedence = "" (key didn't exist)
  update_csi_addons_config(SCHEDULE_PRECEDENCE_KEY, SCHEDULE_PRECEDENCE_STORAGECLASS)

  # teardown:
  if self.original_precedence:  # "" is falsy, so this block is SKIPPED
      update_csi_addons_config(SCHEDULE_PRECEDENCE_KEY, self.original_precedence)

  So if schedule-precedence wasn't set in the ConfigMap before this test ran, the test adds it with value "storageclass" — and then never removes it. The ConfigMap
  is a cluster-level resource that persists across tests. Any subsequent test that checks whether schedule-precedence exists, or behaves differently based on its
  value, will see a leftover "storageclass" that wasn't there before.

  The fix would be something like:

  def finalizer():
      if self.original_precedence:
          update_csi_addons_config(SCHEDULE_PRECEDENCE_KEY, self.original_precedence)
      else:
          # Remove the key we added, restoring the original state
          remove_csi_addons_config_key(SCHEDULE_PRECEDENCE_KEY)
```

yitzhak12 (approved):

ebenahar (approved):

PR #14884, opened 2026-04-14
Fix craft_s3_command to check mcg_obj.ssl for disabling SSL verification

sagihirshfeld:
craft_s3_command only checked signed_request_creds for the ssl flag, ignoring mcg_obj.ssl. This caused NSFS tests to fail with "SSL: CERTIFICATE_VERIFY_FAILED" because --no-verify-ssl was never added to the AWS CLI command when ssl=False was set on mcg_obj.

Also clarifies the operator precedence in the existing signed_request_creds check for readability — the original logic was functionally correct but relied on subtle Python `and` short-circuit behavior.


ypersky1980 (approved):
LGTM

udaysk23 (approved):

udaysk23 (approved):

PR #14893, opened 2026-04-14
Feature/must gather report generator

shivamdurgbuns:
Consolidates important commands that are run from the must-gather command and creates a report out of it, so that we can get a one glance status of the cluster during the failure.
This feature doesnt use AI, it just greps out command outputs from must-gather and pulls it out in a file.

## Release Notes

* **New Features**
  * Automatic must-gather report generation now runs after OCS log collection for deployments and failed tests, when enabled.
  * Generated report links are mapped to remote viewing URLs and attached to the corresponding deployment/test results (including additional report artifacts when available).
* **Configuration**
  * Added `REPORTING.generate_must_gather_report` (default: `true`) to control this post–log-collection processing.
* **Bug Fixes**
  * Improved multicluster targeting so the correct cluster configuration set is used for success vs failure paths.


scripts/python/must_gather_report_generator/outputs/xml_output.py
```diff
@@ -0,0 +1,547 @@
+"""Analysis functions for xml_output"""
```
petr-balogh, line 1: I see this XML report generation - do you have any specific reason to use XML ? As this is not that much readable by person - I think, that for readability it will be better to use format like YAML instead

petr-balogh: Ah, after we discussed offline this was probably a misunderstanding of what we disucssed. 

This is probably not needed at all - what I meant was to inject the details what you are generating in plain text to Junit XML file post traceback - similarly like @shylesh doing with his AI report.

shivamdurgbuns:
the must-gather report are also generating correctlly both in txt and xml: [must-gahter  report](https://url.corp.redhat.com/d335675 )
Job: [link](https://url.corp.redhat.com/95195bf)

shivamdurgbuns:
RP: https://url.corp.redhat.com/6e010a6
Verification job: https://url.corp.redhat.com/12e4931

ocs_ci/ocs/must_gather/report_generator_hook.py
```diff
@@ -0,0 +1,183 @@
...
+
+from ocs_ci.framework import config
+
+logger = logging.getLogger(__name__)
+
+# JUnit testcase property consumed by Data Router / Report Portal (property_filter: ".*")
+MUST_GATHER_ANALYSIS_URL_PROPERTY = "must-gather-analysis-url"
+
+# Magna HTTP base for NFS paths when logs_url is not configured (see ocs4-jenkins)
+MAGNA_BASE_URL = "http://magna002.ceph.redhat.com"
```
dahorak, line 16: I wouldn't expose this url here publicly and also with the upcoming change when we will migrate to the containerized agents we will use different services/URLs to access the logs.

pyproject.toml
```diff
@@ -138,7 +138,7 @@ rosa-ocp-version = "ocs_ci.utility.rosa:rosa_ocp_version_endpoint"
 get-rosa-version = "ocs_ci.utility.rosa:get_rosa_version"
 deploy-fusion = "ocs_ci.framework.fusion.main:main"
 deploy-fdf = "ocs_ci.framework.fusion_data_foundation.main:main"
-fdf-mirror = "ocs_ci.framework.fdf_mirror.main:main"
+must-gather-report = "must_gather_report_generator.main:main"
```
dahorak, line 142: The `must-gather-report` entry point should be added as new one, not replacing the `fdf-mirror` right?

shivamdurgbuns: must have happened during rebase. corrected it.

ocs_ci/ocs/must_gather/report_generator_hook.py
shivamdurgbuns: how to tackle this then?

PR #14905, opened 2026-04-14
Adding a test case for iam user policy actions

ypersky1980:
The new test case implements the following  iam user policy scenario: 
            1. Create an iam user add a policy to him, verify put_user_policy command
            2. Verify get_user_policy and list_user_policies commands
            3. Delete the policy and verify that the deletion was successful

## Summary by CodeRabbit

* **Tests**
  * Improved teardown to remove all IAM user-attached policies before deleting accounts, improving cleanup reliability.
  * Added functional tests covering IAM user policy lifecycle: create/attach, verify retrieval, delete, and confirm no remaining policies.
* **Chores**
  * Improved backend IAM handling for admin operations to support more reliable policy management.

[Review Change Stack](https://app.coderabbit.ai/change-stack/red-hat-storage/ocs-ci/pull/14905?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/functional/object/mcg/test_iam_users.py
+++ b/tests/functional/object/mcg/test_iam_users.py
@@ -2,6 +2,8 @@ import logging
 import random
 import string
 import secrets
+import json
+import boto3
 
 from ocs_ci.framework.pytest_customization.marks import tier2, red_squad, mcg
 from ocs_ci.framework.testlib import MCGTest
@@ -45,6 +47,7 @@ class TestIAMUsers(MCGTest):
         6. Delete all the created users
 
         Args:
+            iam_users_factory: calling this fixture creates new iam users
             mcg_obj: An MCG class instance
             awscli_pod_session (pod): A pod running the AWSCLI tools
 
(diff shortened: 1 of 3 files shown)


tests/functional/object/mcg/test_iam_users.py
```diff
@@ -198,3 +201,82 @@ def test_iam_user_actions(self, iam_users_factory, mcg_obj, awscli_pod_session):
...
+        )
+        policy_names = list_policy_result.get("PolicyNames")
+        assert (
+            isinstance(policy_names, list)
+            and len(policy_names) == 1
+            and policy_names[0] == policy_name
+        ), f'Expected one policy "{policy_name}", got: {policy_names}'
+        logger.info(f"List user policies for user {user_name} was successful")
+
+        delete_policy_cmd = (
```
ebondare, line 271: If one of the assertions above fails, the policy will not be deleted. Would it make sense to add teardown in case that happens, so that there are no leftovers? Or use pytest.fail instead of assert?

sagihirshfeld: @ypersky1980 if a user still has a policy attached to it and the teardown of `iam_users_factory` tries to delete the user, would it block the deletion of the user? 

If the answer is no - then no changes are needed IMO. It'll imply that deleting the user will result in the deletion of its policies (similar to the relationship of buckets and their bucket-policies).

Otherwise you'll have to adjust the `iam_users_factory` [teardown logic](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/conftest.py#L11942-L11955) to iterate over existing policies and delete them before deleting the user.

ypersky1980: The flow of successfully deleting a user is : 
  1. Delete the user policy:
  iam delete-user-policy --user-name <user_name> --policy-name <policy _name>
  2. Delete the access keys (get the key ID first with iam list-access-keys --user-name <user_name>):
  iam delete-access-key --user-name <user_name> --access-key-id <AccessKeyId>
  3. Delete the user:
  iam delete-user --user-name <user_name>

tests/functional/object/mcg/test_iam_users.py
```diff
@@ -198,3 +201,82 @@ def test_iam_user_actions(self, iam_users_factory, mcg_obj, awscli_pod_session):
...
+        policy_name = "test_policy"
+
+        iam_client = boto3.client(
+            "iam",
+            endpoint_url=mcg_obj.iam_endpoint,
+            aws_access_key_id=mcg_obj.access_key_id,
+            aws_secret_access_key=mcg_obj.access_key,
+            verify=False,
+            region_name=mcg_obj.region if mcg_obj.region else "",
+        )
```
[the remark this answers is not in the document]
ypersky1980: Done.

tests/functional/object/mcg/test_iam_users.py
ypersky1980: @ebondare @sagihirshfeld finalizer was updated with deletion of the policy before deleting the user.

sagihirshfeld (dismissed):
LGTM

tests/functional/object/mcg/test_iam_users.py
```diff
@@ -198,3 +201,82 @@ def test_iam_user_actions(self, iam_users_factory, mcg_obj, awscli_pod_session):
...
+        policy_name = "test_policy"
+
+        iam_client = boto3.client(
+            "iam",
+            endpoint_url=mcg_obj.iam_endpoint,
+            aws_access_key_id=mcg_obj.access_key_id,
+            aws_secret_access_key=mcg_obj.access_key,
+            verify=False,
+            region_name=mcg_obj.region or None,
+        )
```
udaysk23: How about creating IAM user in mcg class like what we have for s3 client

udaysk23 (approved):

sagihirshfeld (approved):

PR #14907, opened 2026-04-15
Update the managed cluster list to identify clusterset name in MDR test

jilju:
In the function get_cluster_set_name the managedclusters under test is identified by ENV_DATA.rbd_dr_scenario. This value is not preset in MDR case so the list will be empty. Update the list with the actual managedclusters name if it is empty.


shylesh (approved):
lgtm

ebondare (approved):

petr-balogh:
Verification MDR job:
https://url.corp.redhat.com/e680b08

petr-balogh (approved):

PR #14910, opened 2026-04-15
change default openshift_install_timeout to 2 hours

dahorak:
Change the timeout for waiting for `openshift-install create cluster ...` command to two hours.

Recently we saw bunch of OCP deployment failures because execution of `openshift-install create cluster ...` command takes longer than 1 hour which was the default timeout set in ocs-ci. (For example: https://url.corp.redhat.com/25bbb73 or https://url.corp.redhat.com/7c90ae8) This leads to situation, that the deployment were aborted even when it was progressing correctly or if there were any real issue, this external abort effectively hide any reasonable failure from the `openshift-install` command.

The `openshift-install` command itself have it's own timeout, so if there is any problem with the deployment, the command itself should fail and print proper error message, instead of being killed by ocs-ci without any reasonable message.

### DIFF
--- a/conf/deployment/ibmcloud/ipi_3az_rhcos_3m_3w.yaml
+++ b/conf/deployment/ibmcloud/ipi_3az_rhcos_3m_3w.yaml
@@ -1,6 +1,5 @@
 ---
 DEPLOYMENT:
-  openshift_install_timeout: 7200
   allow_lower_instance_requirements: false
 ENV_DATA:
   platform: 'ibm_cloud'
--- a/conf/deployment/ibmcloud/ipi_3az_rhcos_3m_3w.yaml
+++ b/conf/deployment/ibmcloud/ipi_3az_rhcos_3m_3w.yaml
@@ -1,6 +1,6 @@
 ---
 DEPLOYMENT:
-  openshift_install_timeout: 4800
+  openshift_install_timeout: 7200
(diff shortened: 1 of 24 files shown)


petr-balogh (dismissed):

conf/deployment/ibmcloud/ipi_3az_rhcos_3m_3w.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 DEPLOYMENT:
-  openshift_install_timeout: 4800
+  openshift_install_timeout: 7200
```
petr-balogh: I am thinking if we should then remove those values and rely only on default?

shivamdurgbuns (dismissed):

conf/deployment/ibmcloud/ipi_3az_rhcos_3m_3w.yaml
dahorak: Good point, updated in [c126c39](https://github.com/red-hat-storage/ocs-ci/commit/c126c39e11b73f0f021fd77fb381dd724f776091).

shivamdurgbuns:
fixes: https://github.com/red-hat-storage/ocs-ci/issues/14917

vavuthu (approved):

fbalak (approved):

PR #14915, opened 2026-04-16
deployment/baremetal: fix clean_disks skipping virtio-blk on KVM/Fyre

yitzhak12:
The previous clean_disks used `lsblk -e252` to exclude device-mapper devices, but on KVM/Fyre platforms major 252 is assigned to virtio-blk, so OSD disks (vdb, vdc) were silently skipped during cleanup.

Fix: drop the global lsblk scan with the major-number exclusion. Instead, rely on `disks_available_to_cleanup` (which already filters by boot-disk detection and nbd/loop exclusion) to get the eligible disk list, then query each disk's size individually via a targeted `lsblk /dev/<disk>` call.

Also pass namespace through to `disks_available_to_cleanup` so both functions are consistent, and add a libtest
(`test_disks_available_to_cleanup`) that verifies the per-disk size lookup works correctly on a live cluster without wiping any disks.

### DIFF
--- a/ocs_ci/deployment/baremetal.py
+++ b/ocs_ci/deployment/baremetal.py
@@ -1453,28 +1453,29 @@ def clean_disks(worker, namespace=constants.DEFAULT_NAMESPACE):
 
     """
     ocp_obj = ocp.OCP()
-    disks_available_on_worker_nodes_for_cleanup = disks_available_to_cleanup(worker)
-
-    # Get the name and size in bytes of the disks
-    out = ocp_obj.exec_oc_debug_cmd(
-        node=worker.name,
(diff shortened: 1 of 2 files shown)


yitzhak12:
The Pr validation error above is related to a Ceph health issue, and not to the test.

petr-balogh (approved):

dahorak (approved):
LGTM, but it should be verified also in the Baremetal environment UPI/AI deployment type (at least one of them, better both).

dahorak:
Verification job on BM UPI deployment passed: https://url.corp.redhat.com/7f411c6

PR #14948, opened 2026-04-22
Fix ceph health and do not fail in case Jira was updated

petr-balogh:


vavuthu (approved):

petr-balogh:
Deployment passed here:
https://url.corp.redhat.com/d78be8c

But it didn't hit the issue  , anyway no regression in normal deployment verified.

clacroix12 (approved):

PR #14952, opened 2026-04-22
Fix ARO deployment to clean up orphaned Azure resources

petr-balogh:
ARO deployments were leaving orphaned resources:
- Azure AD applications named 'aro-{cluster-name}'
- VNETs named 'aro-vnet-{cluster-name}'
- DNS records

These resources were not cleaned up during normal teardown, leading to resource accumulation.

Changes:
- Add delete_azure_ad_app() method to clean up AD applications
- Enhance destroy_cluster() to delete all resources (DNS, cluster, VNET, AD app)
- Use ignore_error=True for all delete operations

All cleanup operations are idempotent - they handle missing resources gracefully and will not fail if resources have already been deleted or were never created.

The destroy_cluster() method now:
1. Deletes DNS records
2. Deletes ARO cluster (which also deletes cluster resource group)
3. Deletes VNET
4. Deletes Azure AD application

Each operation logs whether the resource was found and deleted, or was already gone. This ensures cleanup works even when deployment partially fails or resources were manually deleted.

### DIFF
--- a/ocs_ci/utility/azure_utils.py
+++ b/ocs_ci/utility/azure_utils.py
@@ -631,13 +631,21 @@ class AzureAroUtil(AZURE):
         logger.info(f"Deleting service principal: {sp_name}")
 
         try:
-            # Get the app ID first
+            # Get the service principal by display name
+            # NOTE: --display-name does substring matching, so we need to filter for exact match
             list_cmd = f"az ad sp list --display-name {sp_name}"
             result = exec_cmd(list_cmd, timeout=60, ignore_error=True)
             if result.stdout:
                 sp_list = json.loads(result.stdout)
-                if sp_list:
-                    app_id = sp_list[0]["appId"]
+
+                # Filter for exact display name match to avoid deleting SPs from clusters
+                # where one cluster name is a substring of another (e.g., "aro-sp-j-002" vs "aro-sp-j-002zm3c33")
+                matching_sps = [
+                    sp for sp in sp_list if sp.get("displayName") == sp_name
+                ]
+
+                if matching_sps:
+                    app_id = matching_sps[0]["appId"]
                     delete_cmd = f"az ad sp delete --id {app_id}"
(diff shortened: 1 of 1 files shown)


petr-balogh:
Destroy verification:
 https://url.corp.redhat.com/3de8371

ocs_ci/utility/azure_utils.py
```diff
@@ -849,42 +850,43 @@ def get_kubeconfig(self, cluster_name, resource_group, path=None):
...
         try:
-            cmd = f"az network vnet delete --resource-group {resource_group} --name {vnet}"
-            out = exec_cmd(cmd, timeout=600).stdout
-            logger.info(f"VNET deletion output: {out}")
-        except CommandFailed as e:
-            logger.warning(f"Failed to delete VNET '{vnet}': {e}")
-            logger.info("VNET may have already been deleted or does not exist")
+            # First, try to find the app by display name
+            list_cmd = (
+                f"az ad app list --display-name {app_name} --query '[].appId' -o tsv"
```
dahorak: It looks like the `--display-name` parameter is not an exact match filter, but returns any app containing the provided string. For example:
```
$ az ad app list --display-name aro-j-002 --query '[].displayName' -o tsv
aro-j-002zm3c33-d
aro-j-002zm3c33-t1
aro-j-002zm3c33-uba
```
This could theoretically lead to deleting unrelated app, if the cluster_name of one cluster will be a substring of cluster_name of another cluster.

petr-balogh:
Verification here:
https://url.corp.redhat.com/10814bb

Petr

ocs_ci/utility/azure_utils.py
petr-balogh: Update logic as I also had to rebase on Vijay's fix

dahorak (approved):

petr-balogh:
Destroy is now triggered here:
https://url.corp.redhat.com/0276a0c

vavuthu (approved):

PR #14997, opened 2026-04-27
[4.18] Fix NFS CSI pod leftover false positives in `TestBucketCreationAndDeletion`

sagihirshfeld:
## Summary
  - `test_bucket_creation_deletion[1-OC-PVPOOL]` consistently fails with
    `ResourceLeftoversException` due to `csi-nfsplugin` and
    `csi-nfsplugin-provisioner` pods still terminating from the preceding
    NFS test class
  - This has only been observed in 4.17 and 4.18 — no NFS CSI pod
    leftover failures were found in 4.19 or above launches despite the 
    same test execution order
  - The `environment_checker` fixture is class-scoped, so its finalizer
    runs after the last test in the class — `1-OC-PVPOOL` is always that
    last tier1 parametrization, making it the consistent victim
  - Added `@ignore_leftover_label` with both NFS CSI pod labels to
    `TestBucketCreationAndDeletion` so these transient pods are excluded
    from the PRE/POST environment diff

### DIFF
--- a/tests/functional/object/mcg/test_bucket_creation_deletion.py
+++ b/tests/functional/object/mcg/test_bucket_creation_deletion.py
@@ -14,6 +14,7 @@ from ocs_ci.framework.pytest_customization.marks import (
     runs_on_provider,
     red_squad,
     mcg,
+    ignore_leftover_label,
 )
 from ocs_ci.ocs.bucket_utils import sync_object_directory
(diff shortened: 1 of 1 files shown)


ypersky1980 (approved):
LGTM

udaysk23 (approved):

PR #15005, opened 2026-04-28
Deprecate test_db_scc — NooBaa DB no longer uses custom SCC

sagihirshfeld:
In ODF 4.19, NooBaa migrated its database to CNPG, which removed the noobaa-db SCC entirely — the DB pods now run under restricted-v2 with no elevated privileges needed. This test validated SCC priority behavior for the old custom SCC, so its entire premise no longer applies.


sagihirshfeld:
Since this only deprecates a test, I'm adding the Verified label

vavuthu (approved):

ypersky1980 (approved):
LGTM

PR #15006, opened 2026-04-28
`test_write_to_full_bucket`: Increase PV-pool backingstore memory limit to 800Mi

sagihirshfeld:
## Summary
  - Increase PV-pool backingstore pod memory (request + limit) from 500Mi to 800Mi in `test_write_to_full_bucket`
  - The NooBaa agent consistently OOMKills at 500Mi when handling 500MB file writes — RSS climbs to ~494–506 MB
  with concurrent inflight writes, exceeding the 500Mi cgroup limit (524 MB)
  - Confirmed via must-gather across 3 independent failed runs (PV-pool pod exits with code 137 / SIGKILL)

## Summary by CodeRabbit

* **Tests**
  * Improved reliability of PV-backed storage tests by increasing requested PV capacity to `800Mi`.
  * When capacity is reached, the test now frees 5 uploaded objects (instead of 3).
  * Added a health check polling step (up to 600s, polling every 30s) to confirm the backing store returns to a healthy state before retrying uploads.

### DIFF
--- a/tests/functional/object/mcg/test_pv_pool.py
+++ b/tests/functional/object/mcg/test_pv_pool.py
@@ -77,9 +77,9 @@ class TestPvPool:
                         MIN_PV_BACKINGSTORE_SIZE_IN_GB,
                         "ocs-storagecluster-ceph-rbd",
                         "100m",
-                        "500Mi",
+                        "800Mi",
                         "100m",
-                        "500Mi",
+                        "800Mi",
                     )
                 ]
             },
--- a/tests/functional/object/mcg/test_pv_pool.py
+++ b/tests/functional/object/mcg/test_pv_pool.py
@@ -133,7 +133,7 @@ class TestPvPool:
         ), f"BackingStore {bs_name} did not recover to healthy status after freeing objects"
 
         # 4. Confirm that uploading is possible again now that there is space
-        retry((CommandFailed,), tries=5, delay=20, backoff=1)(
+        retry((CommandFailed,), tries=5, delay=30, backoff=1)(
             awscli_pod_session.exec_s3_cmd_on_pod
         )(
             f"cp /tmp/testfile s3://{bucket.name}/{uploaded_objs[0]}",
--- a/tests/functional/object/mcg/test_pv_pool.py
+++ b/tests/functional/object/mcg/test_pv_pool.py
@@ -114,13 +114,24 @@ class TestPvPool:
                 break
 
         # 3. Free up some space in the bucket
-        for obj in uploaded_objs[:3]:
+        for obj in uploaded_objs[:5]:
             awscli_pod_session.exec_s3_cmd_on_pod(
                 f"rm s3://{bucket.name}/{obj}", mcg_obj_session
             )
 
+        # Wait for the BackingStore to detect the freed space
+        bs_name = bucket.bucketclass.backingstores[0].name
+        sample = TimeoutSampler(
+            timeout=600,
+            sleep=30,
+            func=check_pv_backingstore_status,
+            backingstore_name=bs_name,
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_pv_pool.py
```diff
@@ -114,13 +114,24 @@ def test_write_to_full_bucket(
...
+            sleep=30,
+            func=check_pv_backingstore_status,
+            backingstore_name=bs_name,
+            namespace=config.ENV_DATA["cluster_namespace"],
+        )
+        sample.wait_for_func_status(result=True)
+
         # 4. Confirm that uploading is possible again now that there is space
-        retry((CommandFailed,), tries=10, delay=30, backoff=1)(
+        retry((CommandFailed,), tries=3, delay=10, backoff=1)(
```
ypersky1980: Retry window reduced too aggressively, it might cause flaky failures

The retry was changed from tries=10, delay=30 (5 minutes total) to tries=3, delay=10 (30 seconds total). The reasoning seems to be that the new TimeoutSampler before it (lines 122-131)
  already ensures the backingstore is healthy, so the retry can be shorter.

  But "backingstore reports healthy" does not mean "bucket is immediately writable." NooBaa needs to internally reconcile the capacity update from the backingstore status through to the  bucket's write path. This reconciliation gap is exactly why the original retry was generous. Cutting it from 5 minutes to 30 seconds is a 10x reduction that trades the old problem (slow test) for a new one (flaky test on slower or loaded clusters). 
I suggest that you consider something like tries=5, delay=20 — still faster than before but with enough buffer for NooBaa's internal reconciliation.

ypersky1980 (commented):
Left a comment.

tests/functional/object/mcg/test_pv_pool.py
```diff
@@ -114,13 +114,24 @@ def test_write_to_full_bucket(
...
+        bs_name = bucket.bucketclass.backingstores[0].name
+        sample = TimeoutSampler(
+            timeout=600,
+            sleep=30,
+            func=check_pv_backingstore_status,
+            backingstore_name=bs_name,
+            namespace=config.ENV_DATA["cluster_namespace"],
+        )
+        sample.wait_for_func_status(result=True)
+
```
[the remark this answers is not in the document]
sagihirshfeld: Fixed in 65a547a6707cee45589205e20d9e660bc365d89f

tests/functional/object/mcg/test_pv_pool.py
sagihirshfeld: Fixed in 65a547a6707cee45589205e20d9e660bc365d89f

ypersky1980 (dismissed):
LGTM

tests/functional/object/mcg/test_pv_pool.py
```diff
@@ -114,13 +114,26 @@ def test_write_to_full_bucket(
...
+            backingstore_name=bs_name,
+            namespace=config.ENV_DATA["cluster_namespace"],
+        )
+        assert sample.wait_for_func_status(
+            result=True
+        ), f"BackingStore {bs_name} did not recover to healthy status after freeing objects"
+
         # 4. Confirm that uploading is possible again now that there is space
-        retry((CommandFailed,), tries=10, delay=30, backoff=1)(
+        retry((CommandFailed,), tries=5, delay=20, backoff=1)(
```
udaysk23: We can keep tries to 5 but please keep delay to 30

sagihirshfeld: Fixed in c35cdacc04590e48b048880b9dfe8dc91e0ced52

udaysk23 (approved):

ypersky1980 (approved):
LGTM

PR #15013, opened 2026-04-29
Fix positional arg bug passing wait as topic_name in bucket notif test

sagihirshfeld:
## Summary
- `should_wait` was passed as the first positional arg to `create_and_register_kafka_topic_with_noobaa`, mapping to `topic_name` instead of `wait`
- When `True`, it passes through the `topic_name or generate_name()` guard and ends up as a boolean `metadata.name` in the KafkaTopic YAML, failing `oc create`


ypersky1980 (approved):
LGTM

ypersky1980 (approved):
LGTM

udaysk23 (approved):

PR #15035, opened 2026-04-30
Fix race condition in test_block_external_access disableRoutes propagation

sagihirshfeld:
## Summary
  - `set_disable_routes_value` patches the storagecluster but doesn't wait for
    the OCS operator to propagate `disableRoutes` to the noobaa CR. The noobaa
    operator reads its own CR, so it can reconcile with the stale value and
    recreate routes before the test checks them.
  - Add a `TimeoutSampler` poll (120s, 5s interval) after patching the
    storagecluster to wait for the noobaa CR's `spec.disableRoutes` to match
    the expected value before proceeding.


ypersky1980 (approved):

ypersky1980 (approved):
LGTM

udaysk23 (approved):

PR #15051, opened 2026-05-04
add test for GKLM kms over kmip

black-dragon74:
This patch adds a new test case for KMIP with `USE_CRYPTO_RPC` set to false in kms configmap.

Setting `USE_CRYPTO_RPC` uses local crypto instead of remote (which is not supported by GKLM) and can be tested with any KMIP compatible KMS(ciphertrust in this case).


tests/functional/pv/pv_encryption/test_kmip_rbd_pv_encryption.py
```diff
@@ -123,3 +123,105 @@ def test_rbd_pv_encryption_kmip(
...
+
+@green_squad
+@aws_platform_required
+@skipif_ocs_version("<4.12")
+@kms_config_required
+@skipif_managed_service
+@skipif_hci_provider_and_client
+@skipif_disconnected_cluster
+@skipif_proxy_cluster
+@polarion_id("OCS-XXXX")
```
black-dragon74, line 136: TODO

PR #15061, opened 2026-05-05
Print pods from submariner operator namespace if submariner install fails

prsurve:


ocs_ci/ocs/acm/acm.py
```diff
@@ -434,38 +434,164 @@ def submariner_validation_ui(self):
...
+                cluster_configs = get_non_acm_cluster_config()
+                for cluster in cluster_configs:
+                    with config.RunWithConfigContext(
+                        cluster.MULTICLUSTER["multicluster_index"]
+                    ):
+                        log.info(
+                            f"Getting pods from cluster: {cluster.ENV_DATA['cluster_name']}"
+                        )
+                        pod_status = run_cmd(
+                            f"oc get pods -n {constants.SUBMARINER_OPERATOR_NAMESPACE}"
```
RHdshelar, line 455: Functionally this looks good, but the same try/assert → collect pod status → raise pattern is repeated in six places. Could we extract a helper (for example, _assert_submariner_ui_status()) and a shared _log_submariner_pod_status() to reduce duplication and keep any future locator/message updates in one place?

PR #15066, opened 2026-05-06
kubeconfig should always create even if kubeconfig is there

rh-aaleem:
Always create kubeconfig for newer build even if the kubeconfig is present.

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -6945,7 +6945,7 @@ def create_kubeconfig(kubeconfig_path):
                 fd.write(config.RUN.get("kubeadmin_password"))
             log.info("Created kubeadmin-password file")
 
-    else:
+    elif not os.path.isfile(kubeconfig_path):
         raise ConfigurationError(
             "Kubeconfig doesn't exists and RUN['kubeadmin_password'] and RUN['ocp_url'] "
             "environment variables were not provided."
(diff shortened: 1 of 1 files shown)


ocs_ci/utility/utils.py
```diff
@@ -6910,44 +6910,43 @@ def create_kubeconfig(kubeconfig_path):
...
-        else:
-            raise ConfigurationError(
-                "Kubeconfig doesn't exists and RUN['kubeadmin_password'] and RUN['ocp_url'] "
-                "environment variables were not provided."
-            )
+    else:
+        raise ConfigurationError(
+            "Kubeconfig doesn't exists and RUN['kubeadmin_password'] and RUN['ocp_url'] "
+            "environment variables were not provided."
+        )
```
dahorak: This `else` part here doesn't make sense, because most of our executions are without `kubeadmin_password` and `ocp_url` (because we are relying on the presence of `kubeconfig` file).
If we want to do this change, we have to raise this exception conditionally only when `kubeconfig` file doesn't exists, something like this:
```
elif not os.path.isfile(kubeconfig_path):
```

petr-balogh, line 6952: This will raise an exception for the cases we do run with kubecofnig file no?

```suggestion
         not os.path.isfile(kubeconfig_path):
             raise ConfigurationError(
                "Kubeconfig doesn't exists and RUN['kubeadmin_password'] and RUN['ocp_url'] "
                "environment variables were not provided."
            )
```

I guess to avoid the exception I mentioned above to be raised in normal kubeconfig file case - we need to fix it this way

dahorak (changes requested):

dahorak (approved):

petr-balogh (approved):

PR #15108, opened 2026-05-10
Add NooBaa managed identity for Azure STS/WIF deployments

sagihirshfeld:
## Summary
- Create an Azure user-assigned managed identity with federated credentials for NooBaa service accounts (`noobaa`, `noobaa-core`, `noobaa-endpoint`) during ODF deployment on Azure STS clusters
- Pass the managed identity's client ID (instead of the service principal's) and resource group to the ODF Subscription env vars (`CLIENTID`, `RESOURCEGROUP`)
- Clean up the managed identity and its role assignments during cluster teardown
- Add `get_azure_sts_creds_from_sub()` utility for retrieving Azure STS credentials from the Subscription (for follow-up BackingStore/NamespaceStore work)

## Details
When deploying ODF on Azure with STS/WIF (`sts_enabled: true`), NooBaa needs a managed identity with federated credentials to authenticate to Azure Blob Storage via workload identity. Without this, NooBaa falls back to a PV-pool backing store.

This mirrors the existing AWS STS pattern (`create_and_attach_sts_role` / `get_role_arn_from_sub`).

**Changes:**
- `azure_utils.py`: `create_noobaa_managed_identity()` and `delete_noobaa_managed_identity()` on the `AZURE` class
- `deployment.py`: Wire MI creation into `deploy_ocs_via_operator()`, fix `subscribe_ocs()` to use MI client ID + add `RESOURCEGROUP` env var + fix `.append()` → `.extend()` bug
- `azure.py`: Wire MI deletion + role assignment cleanup into `AZUREIPI.destroy_cluster()`
- `utils.py`: Add `get_azure_sts_creds_from_sub()`

## Summary by CodeRabbit

* **New Features**
  * Azure STS deployments now create a NooBaa user-assigned managed identity and configure workload identity federation for NooBaa service accounts.
* **Bug Fixes**
  * Cluster destruction for Azure STS now performs managed-identity cleanup to remove NooBaa-related identities and role assignments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/red-hat-storage/ocs-ci/pull/15108?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -1350,12 +1350,11 @@ class Deployment(object):
             logger.test_step("Create Azure managed identity for NooBaa STS")
             azure_util = AzureUtil()
             azure_util.az_login()
-            mi_data = azure_util.create_noobaa_managed_identity(
+            self.azure_noobaa_mi_client_id = azure_util.create_noobaa_managed_identity(
                 cluster_name=config.ENV_DATA["cluster_name"],
                 resource_group=config.ENV_DATA["cluster_name"],
                 subscription_id=config.AUTH["azure_auth"]["subscription_id"],
             )
-            self.azure_noobaa_mi_client_id = mi_data["client_id"]
         stage_testing = config.DEPLOYMENT.get("stage_rh_osbs")
         konflux_build = config.DEPLOYMENT.get("konflux_build")
         upgrade = config.UPGRADE.get("upgrade", False)
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -18,6 +18,7 @@ import yaml
 from botocore.exceptions import EndpointConnectionError, BotoCoreError
 
 from ocs_ci.deployment.helpers import storage_class
+from ocs_ci.utility.azure_utils import AZURE as AzureUtil
 from ocs_ci.deployment.helpers.hypershift_base import is_hosted_cluster
 from ocs_ci.deployment.ocp import OCPDeployment as BaseOCPDeployment
 from ocs_ci.deployment.helpers.external_cluster_helpers import (
@@ -215,6 +216,7 @@ class Deployment(object):
 
     def __init__(self):
         self.sts_role_arn = None
+        self.azure_noobaa_mi_client_id = None
         storage_class.set_custom_storage_class_path()
         logger.info(
             f"Deployment platform {self.platform} initiated with storage class: {self.storage_class}"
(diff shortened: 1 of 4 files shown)


sagihirshfeld:
Verified in https://url.corp.redhat.com/1ea3d3b

ypersky1980 (approved):
LGTM

udaysk23 (dismissed):

ocs_ci/utility/utils.py
```diff
@@ -6326,6 +6263,41 @@ def get_role_arn_from_sub():
         raise ClusterNotInSTSModeException
 
 
+def get_azure_sts_creds_from_sub():
```
clacroix12: I don't see this function being used anywhere, is it still necessary?

Edit: I see it's intended for follow-up BackingStore/NamespaceStore work. I think this should at least be in it's own commit, but better-yet it's own PR if it isn't necessary for these changes. If we need to backport this PR to previous release branches, it's better to not include unrelated code.

ypersky1980 (dismissed):
LGTM

ocs_ci/utility/utils.py
sagihirshfeld: Gotcha - removed it in commit https://github.com/red-hat-storage/ocs-ci/pull/15108/changes/47c0e598147f669facfbaff8c56e6fe401b45ca7

I'll add it in the follow-up PR along with the support for AzureSTS backingstore and namespacestore creation

nehaberry (dismissed):

ocs_ci/deployment/deployment.py
```diff
@@ -1228,15 +1230,19 @@ def subscribe_ocs(self):
...
             azure_auth_data = config.AUTH["azure_auth"]
+            mi_client_id = (
+                self.azure_noobaa_mi_client_id or azure_auth_data["client_id"]
+            )
             azure_sub_data = [
-                {"name": "CLIENTID", "value": azure_auth_data["client_id"]},
+                {"name": "CLIENTID", "value": mi_client_id},
                 {"name": "TENANTID", "value": azure_auth_data["tenant_id"]},
                 {"name": "SUBSCRIPTIONID", "value": azure_auth_data["subscription_id"]},
+                {"name": "RESOURCEGROUP", "value": config.ENV_DATA["cluster_name"]},
```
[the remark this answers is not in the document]
sagihirshfeld: In Azure IPI (which is the only way we're deploying on Azure) we can always assume that its Azure resources (VMs, etc) are installed on a resource group of the same name as the cluster — it's set up this way by the installer. So this derivation is safe.

That said, the main item in that method's output is the managed identity's client ID, and the resource group name was just a pass-through of the input. I made a follow-up commit so the method now returns only the client ID.

udaysk23 (approved):

ypersky1980 (approved):
LGTM

sagihirshfeld:
Verified again after the lates changes: https://url.corp.redhat.com/65d7b66

clacroix12 (approved):

PR #15119, opened 2026-05-12
Fix NFS mount failure on IBM Cloud by fixing retry logic and opening port 2049

amr1ta:
# Summary

**Root cause of IBM Cloud failure:** 

`IBM Cloud VPC LoadBalancer security groups block inbound traffic by default. When the NFS LB service (rook-ceph-nfs-my-nfs-load-balancer) is created, port 2049 is not allowed through the security group, so mount requests from the external NFS client VM never reach the NFS server.`

**Fix:**

 Open inbound TCP port 2049 on IBM Cloud VPC LoadBalancer security group for the NFS LB service. IBM Cloud VPC LB security groups block inbound traffic by default — same issue that required configure_ingress_load_balancer_security_group() for ports 80/443 in #[15012(https://github.com/red-hat-storage/ocs-ci/pull/15012)]. The rule is added automatically in create_nfs_load_balancer_service() and cleaned up in delete_nfs_load_balancer_service().

Have updated, Stale subvolume assertion logic — the old assertion assert len(stale_volumes) == 0 fails if there are pre-existing stale subvolumes from other tests or previous runs. The test should only verify that its own subvolume (new_pvc[1]) was successfully deleted, not that the entire cluster has zero stale subvolumes. 

Changed to:
Log stale volumes before and after delete for debugging
Log the delete command output so failures are visible
Assert that the specific subvolume created by this test is no longer in the stale list

### DIFF
--- a/ocs_ci/utility/ibmcloud.py
+++ b/ocs_ci/utility/ibmcloud.py
@@ -1502,20 +1502,16 @@ def _get_lb_security_groups(svc_name, namespace):
     lb_ingress = svc_data.get("status", {}).get("loadBalancer", {}).get("ingress", [])
     if not lb_ingress:
         logger.warning(
-            "No LB ingress on service %s, cannot configure " "security group",
-            svc_name,
+            f"No LB ingress on service {svc_name}, cannot configure " f"security group"
         )
         return []
 
     lb_hostname = lb_ingress[0].get("hostname") or lb_ingress[0].get("ip")
     if not lb_hostname:
-        logger.warning(
-            "No hostname/IP in LB ingress for service %s",
-            svc_name,
-        )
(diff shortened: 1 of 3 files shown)


amr1ta:
verification jobs,
https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/67936/console
https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/67948/
https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/67953/

dahorak (approved):

ebenahar (approved):

PR #15129, opened 2026-05-13
Skip test_rgw_keda_ha in external mode

sagihirshfeld:
## Summary
  - The `enable_rgw_hpa` fixture annotates `ocs-storagecluster`, which does not exist in external mode (`ocs-external-storagecluster`)
  - Beyond the name mismatch, the test requires installing KEDA via Helm on the cluster where RGW runs. In external mode that cluster is not managed by OCS-CI, so
  setting up KEDA and its prerequisites there is out of scope
  - Added `@skipif_external_mode` to `TestKedaHA`, consistent with other RGW tests in the same directory

### DIFF
--- a/tests/functional/object/rgw/test_keda_ha.py
+++ b/tests/functional/object/rgw/test_keda_ha.py
@@ -8,6 +8,7 @@ from ocs_ci.framework.pytest_customization.marks import (
     red_squad,
     rgw,
     skipif_disconnected_cluster,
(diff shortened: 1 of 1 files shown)


sagihirshfeld:
Adding the Verified label to the limited nature of the change

ypersky1980 (approved):
LGTM

udaysk23 (approved):

PR #15184, opened 2026-05-19
Add FDF Image Mirroring Functionality

rh-aaleem:
**Summary**

Implements FDF (Fusion Data Foundation) image mirroring functionality to mirror unreleased FDF images from Mgen to disconnected mirror registries, following the same pattern as existing ODF mirroring.

**Features**

1. Automated Mirroring: Mirrors FDF catalog and all related operator images
2. IDMS Management: Automatically creates and applies ImageDigestMirrorSet
3. Registry Configuration: Optional configuration of /etc/containers/registries.conf for internal images
4. Pull Secret Support: Uses existing pull secret from cluster-path or ~/.docker/config.json
5. CLI Command: Easy-to-use fdf-mirror command or Python module invocation
6. Jenkins Integration: Works seamlessly in Jenkins pipelines

**Testing**

1. Tested in Jenkins pipeline with FDF 4.18.20-5 catalog
2. Successfully mirrors catalog and operator images
3. IDMS creation and application verified
4. MachineConfigPool updates confirmed

## Summary by CodeRabbit

* **New Features**
  * Added a CLI to mirror Fusion Data Foundation catalogs with mirror-registry override, optional on-host registries configuration, and improved validation/logging.
  * Mirroring supports whole-catalog or package-scoped operations and configurable naming prefix for generated metadata.

* **Chores**
  * Added ImageSet and registries configuration templates and extended init/CLI to accept mirror-related options and credentials.

### DIFF
--- a/docs/usecases/fdf_mirroring.md
+++ b/docs/usecases/fdf_mirroring.md
@@ -14,8 +14,7 @@ This guide covers mirroring FDF catalog images and their associated operator ima
 ### 1. Pull Secret (Required)
 
 A pull secret with authentication for both source and destination registries. The pull secret should be located at:
-- `${CLUSTER_PATH}/auth/pull-secret` (recommended), or
-- `~/.docker/config.json`
+- pull_secret_path = os.path.join(constants.DATA_DIR, "pull-secret") (recommended)
 
 The pull secret must include authentication for:
 
@@ -25,7 +24,7 @@ The pull secret must include authentication for:
 - `icr.io/cpopen` - FDF operator images
 
 **Destination registry** (where images are pushed to):
-- Your mirror registry (e.g., `r8-ru26-w-l.quay-service.fusion.tadn.ibm.com`)
+- Your mirror registry
 
 ```
 
@@ -55,8 +54,8 @@ DEPLOYMENT:
 
 ```bash
 python -m ocs_ci.framework.fdf_mirror.main \
-  --catalog-image cp.stg.icr.io/cp/df/isf-data-foundation-catalog:v4.20 \
-  --mirror-registry r8-ru26-w-l.quay-service.fusion.tadn.ibm.com/organization/df_team \
+  --catalog-image <cateloge name> \
+  --mirror-registry <mirror-registry> \
   --cluster-path ~/current-cluster-dir/openshift-cluster-dir \
   --cluster-name my-cluster \
   --configure-registries
@@ -65,7 +64,7 @@ python -m ocs_ci.framework.fdf_mirror.main \
 **With Config File:**
 ```bash
 python -m ocs_ci.framework.fdf_mirror.main \
-  --catalog-image cp.stg.icr.io/cp/df/isf-data-foundation-catalog:v4.20 \
+  --catalog-image <cateloge name> \
   --cluster-path ~/current-cluster-dir/openshift-cluster-dir \
   --ocsci-conf /path/to/config.yaml \
   --configure-registries
--- a/docs/usecases/fdf_mirroring.md
+++ b/docs/usecases/fdf_mirroring.md
@@ -52,7 +52,7 @@ DEPLOYMENT:
 
 **Basic Usage (with all CLI arguments):**
 ```bash
-python -m ocs_ci.framework.fdf_mirror.main \
+fdf-mirror \
   --catalog-image <catalog-image-url> \
   --mirror-registry <mirror-registry-url> \
   --mirror-registry-user <username> \
@@ -64,7 +64,7 @@ python -m ocs_ci.framework.fdf_mirror.main \
 
 **With Config File (credentials in config):**
 ```bash
-python -m ocs_ci.framework.fdf_mirror.main \
+fdf-mirror \
   --catalog-image <catalog-image-url> \
   --mirror-registry <mirror-registry-url> \
   --cluster-path <path-to-cluster-dir> \
@@ -74,7 +74,7 @@ python -m ocs_ci.framework.fdf_mirror.main \
 
 **Mixed (registry from CLI, credentials from config):**
 ```bash
-python -m ocs_ci.framework.fdf_mirror.main \
+fdf-mirror \
   --catalog-image <catalog-image-url> \
   --mirror-registry <mirror-registry-url> \
   --cluster-path <path-to-cluster-dir> \
@@ -118,7 +118,7 @@ results = mirror_fdf_catalog_via_oc_mirror(
 ```
 
 **Notes:**
-- Use `python -m ocs_ci.framework.fdf_mirror.main` for reliability in Jenkins
+- Use `fdf-mirror` command (installed entrypoint) or `python -m ocs_ci.framework.fdf_mirror.main` as alternative
 - KUBECONFIG must point to cluster kubeconfig
 - Pull secret at `${CLUSTER_PATH}/auth/pull-secret` is used automatically
 - oc-mirror tool is installed automatically if not present
@@ -132,7 +132,7 @@ results = mirror_fdf_catalog_via_oc_mirror(
 The tool supports generating JUnit XML format test reports for CI/CD integration:
 
 ```bash
-python -m ocs_ci.framework.fdf_mirror.main \
+fdf-mirror \
   --catalog-image <catalog-image-url> \
   --mirror-registry <mirror-registry-url> \
   --cluster-path <path-to-cluster-dir> \
@@ -425,7 +425,7 @@ cat /path/to/report.xml
 ### 6. CI/CD Integration
 - Use `--report` option for JUnit XML output
 - Set appropriate timeouts in CI pipelines (5+ hours recommended)
-- Use `python -m ocs_ci.framework.fdf_mirror.main` for module execution reliability
+- Use `fdf-mirror` command (installed entrypoint)
 - Check exit codes for success/failure status
 
 ### 7. Troubleshooting
--- a/docs/usecases/fdf_mirroring.md
+++ b/docs/usecases/fdf_mirroring.md
@@ -62,15 +62,24 @@ python -m ocs_ci.framework.fdf_mirror.main \
   --configure-registries
 ```
 
+**With Config File:**
+```bash
+python -m ocs_ci.framework.fdf_mirror.main \
+  --catalog-image cp.stg.icr.io/cp/df/isf-data-foundation-catalog:v4.20 \
+  --cluster-path ~/current-cluster-dir/openshift-cluster-dir \
+  --ocsci-conf /path/to/config.yaml \
+  --configure-registries
+```
+
 **Required Arguments:**
 - `--catalog-image`: FDF catalog image to mirror (required)
-- `--mirror-registry`: Target mirror registry URL (required)
+- `--mirror-registry`: Target mirror registry URL (required if not in config)
 - `--cluster-path`: Path to OCP cluster directory containing auth/kubeconfig (required)
 
 **Optional Arguments:**
 - `--cluster-name`: Name of the OCP cluster (optional if metadata.json exists in cluster-path)
 - `--configure-registries`: Configure /etc/containers/registries.conf for internal images
-- `--conf`: Path to config file (optional, only if overriding mirror registry credentials)
+- `--ocsci-conf` or `--conf`: Path to config file (optional, can be used to provide mirror_registry and credentials). Both arguments are supported and can be used interchangeably or together.
 - `--report`: Path for JUnit report output
 
 ### Python API
--- a/docs/usecases/fdf_mirroring.md
+++ b/docs/usecases/fdf_mirroring.md
@@ -40,8 +40,6 @@ DEPLOYMENT:
   mirror_registry_password: ''
 ```
 
-**Note:** If your pull secret already contains credentials for the mirror registry, you don't need to provide them in the config file.
-
 ### 3. Required Tools
 
 - `oc` CLI tool (OpenShift client)
@@ -94,10 +92,17 @@ python -m ocs_ci.framework.fdf_mirror.main \
 - `--mirror-registry-user`: Mirror registry username (optional, can be provided via CLI or config file)
 - `--mirror-registry-password`: Mirror registry password (optional, can be provided via CLI or config file)
 - `--configure-registries`: Configure /etc/containers/registries.conf for internal FDF images
-- `--ocsci-conf` or `--conf`: Path to config file (optional, can be used to provide mirror_registry and credentials). Both arguments are supported and can be used interchangeably or together.
-- `--report`: Path for JUnit report output
+- `--ocsci-conf` or `--conf`: Path to config file (optional, can be used to provide mirror_registry and credentials). Both arguments are supported and can be used interchangeably or together. Multiple config files can be specified and will be merged.
+- `--report`: Path for JUnit report output (generates JUnit XML format test results)
 
-**Note:** CLI arguments take precedence over config file values. If credentials are not provided via CLI or config, the tool will use credentials from the pull secret.
+### Credential Resolution Order
+
+The tool resolves mirror registry credentials using the following priority (highest to lowest):
+1. **CLI arguments** (`--mirror-registry-user`, `--mirror-registry-password`) - highest priority
+2. **Config file values** (`mirror_registry_user`, `mirror_registry_password` in DEPLOYMENT section)
+3. **Pull secret** (credentials extracted from pull-secret file) - fallback
+
+**Note:** CLI arguments always take precedence over config file values. If credentials are not provided via CLI or config, the tool will automatically extract and use credentials from the pull secret.
 
 ### Python API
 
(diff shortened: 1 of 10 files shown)


docs/usecases/fdf_mirroring.md
```diff
@@ -0,0 +1,260 @@
...
+- Mirrors the FDF catalog and all related images
+- Creates ImageDigestMirrorSet (IDMS) for cluster configuration
+- Optionally configures registries.conf for internal image mirrors
+
+## Prerequisites
+
+### 1. Pull Secret (Required)
+
+A pull secret with authentication for both source and destination registries. The pull secret should be located at:
+- `${CLUSTER_PATH}/auth/pull-secret` (recommended), or
```
dahorak: In ocs-ci the common place for `pull-secret` is `./data/pull-secret` location (for example see [this code](https://github.com/red-hat-storage/ocs-ci/pull/15184/changes#diff-2ec5dfe2075a6d031a8a009889826dd1fdc613c620671c1751906fb2215ce137R296):
```
    pull_secret_path = os.path.join(constants.DATA_DIR, "pull-secret")
```
So I think it might be worth to use the same location here (and on other places in this PR) as well.

docs/usecases/fdf_mirroring.md
```diff
@@ -0,0 +1,260 @@
...
+This guide covers mirroring FDF catalog images and their associated operator images using the `oc-mirror` tool. The process automatically:
+- Mirrors the FDF catalog and all related images
+- Creates ImageDigestMirrorSet (IDMS) for cluster configuration
+- Optionally configures registries.conf for internal image mirrors
+
+## Prerequisites
+
+### 1. Pull Secret (Required)
+
+A pull secret with authentication for both source and destination registries. The pull secret should be located at:
```
dahorak, line 16: ~~In `ocs-ci` we have an option to login to the "destination" mirror registry based on credentials provided via the config like this:~~
```
DEPLOYMENT:
  mirror_registry: 'registry.example.com:5000'
  mirror_registry_user: 'username'
  mirror_registry_password: 'password'
```
~~Would it be possible to add similar option to this code as well? So it will be compatible with the existing approach for OCP&ODF disconnected mode deployment.~~

Based on the bellow section, this is actually possible, so it might be worth to update this sentence that there is actually an option to login to the destination/mirror registry during the execution.

docs/usecases/fdf_mirroring.md
```diff
@@ -0,0 +1,260 @@
...
+
+The pull secret must include authentication for:
+
+**Source registries** (where images are pulled from):
+- `cp.stg.icr.io` - FDF catalog registry
+- `registry.redhat.io` - Red Hat base images
+- `icr.io/cpopen` - FDF operator images
+
+**Destination registry** (where images are pushed to):
+- Your mirror registry (e.g., `r8-ru26-w-l.quay-service.fusion.tadn.ibm.com`)
```
dahorak: Please do not put there real mirror registry url (it is public repo).

docs/usecases/fdf_mirroring.md
```diff
@@ -0,0 +1,260 @@
...
+- Access to OCP cluster with kubeconfig
+
+## Quick Start
+
+### CLI Command
+
+```bash
+python -m ocs_ci.framework.fdf_mirror.main \
+  --catalog-image cp.stg.icr.io/cp/df/isf-data-foundation-catalog:v4.20 \
+  --mirror-registry r8-ru26-w-l.quay-service.fusion.tadn.ibm.com/organization/df_team \
```
dahorak: It will be probably better to not use real internal links/urls here (and on other places in this PR).

docs/usecases/fdf_mirroring.md
```diff
@@ -0,0 +1,260 @@
...
+kind: ImageSetConfiguration
+apiVersion: mirror.openshift.io/v1alpha2
+mirror:
+  operators:
+    - catalog: cp.stg.icr.io/cp/df/isf-data-foundation-catalog:v4.20
+```
+
+### 2. Configure registries.conf (Optional - For Internal Images)
+
+If you need to mirror internal images, add to `/etc/containers/registries.conf`:
```
dahorak: Instead of updating the `/etc/containers/registries.conf` file, it might be worth to create new file in `/etc/containers/registries.conf.d/` with unique/ocs-ci specific name, so we will be able to easily track the changes we did and remove/update them if needed. (If this `registries.conf.d/` approach is available on all the platform, where we will need it.)

ocs_ci/deployment/disconnected.py
```diff
@@ -256,31 +309,59 @@ def mirror_index_image_via_oc_mirror(index_image, packages, idms=None):
...
+                registries_conf_path = "/etc/containers/registries.conf"
+                exec_cmd(
+                    f"sudo cp {registries_conf_path} {registries_conf_path}.backup"
+                )
+
+                temp_file = tempfile.NamedTemporaryFile(
+                    mode="w", delete=False, suffix=".conf"
+                )
+                temp_file.write(registries_content)
+                temp_file.close()
```
petr-balogh: Do inside with statement as suggested

ocs_ci/deployment/disconnected.py
```diff
@@ -37,6 +37,45 @@
...
+    Returns:
+        str: mirrored catalog image URL
+
+    """
+    logger.info(f"Mirroring FDF catalog: {catalog_image}")
+
+    # Set mirror_registry in config if provided
+    if mirror_registry:
+        config.DEPLOYMENT["mirror_registry"] = mirror_registry
+        logger.info(f"Set mirror_registry in config: {mirror_registry}")
```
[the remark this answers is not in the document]
dahorak: I agree with the _coderabbitai_ suggestion, that it might not be good/expected behaviour to overwrite the global `config.DEPLOYMENT["mirror_registry"]` parameter here.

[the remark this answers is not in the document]
dahorak: I agree with the _coderabbitai_ suggestion, that it might not be good/expected behaviour to overwrite the global `config.DEPLOYMENT["mirror_registry"]` parameter here (even though I'm not sure, if we will actually hit some case, when this could break the following workflow.

ocs_ci/deployment/disconnected.py
```diff
@@ -256,31 +309,59 @@ def mirror_index_image_via_oc_mirror(index_image, packages, idms=None):
...
+                    f"sudo cp {registries_conf_path} {registries_conf_path}.backup"
+                )
+
+                temp_file = tempfile.NamedTemporaryFile(
+                    mode="w", delete=False, suffix=".conf"
+                )
+                temp_file.write(registries_content)
+                temp_file.close()
+
+                exec_cmd(f"sudo cp {temp_file.name} {registries_conf_path}")
```
dahorak: We probably don't want to overwrite the whole `/etc/containers/registries.conf` file (see also the `/etc/containers/registries.conf.d/` directory option mentioned above).

ocs_ci/templates/fusion-data-foundation/registries.conf.template
prsurve, line 1: we also need one more entry

[[registry]]
location="[preprod.icr.io/cpopen/](http://preprod.icr.io/cpopen/)"
[[registry.mirror]]
location="[cp.stg.icr.io/cp/df](http://cp.stg.icr.io/cp/df)"
mirror-by-digest-only = false
pull-from-mirror = "all"
short-name-mode = "permissive"

ocs_ci/utility/framework/fusion_fdf_init.py
```diff
@@ -127,11 +149,17 @@ def init_cli(self, args: list) -> list:
...
+            "--ocsci-conf",
             action="append",
             default=[],
             help="Path to config file. Repeatable.",
         )
+        parser.add_argument(
+            "--conf",
+            action="append",
+            default=[],
+            help="Path to config file (alias for --ocsci-conf). Repeatable.",
```
dahorak: Do we really need the `--ocsci-conf` parameter here? IIRC there were discussion when the deploy-fdf/deploy fusion entry points were developed, that we will use the more generic names `--conf` parameters instead of the specific `--ocsci-conf`.

docs/usecases/fdf_mirroring.md
```diff
@@ -0,0 +1,435 @@
...
+- `oc-mirror` plugin (automatically installed if not present)
+- Access to OCP cluster with kubeconfig
+
+## Quick Start
+
+### CLI Command
+
+**Basic Usage (with all CLI arguments):**
+```bash
+python -m ocs_ci.framework.fdf_mirror.main \
```
dahorak: Why it is suggesting `python -m ocs_ci.framework.fdf_mirror.main` instead of the new entry point `fdf-mirror` (similarly on other places)?

docs/usecases/fdf_mirroring.md
```diff
@@ -0,0 +1,435 @@
...
+# Mirror FDF catalog and images
+results = mirror_fdf_catalog_via_oc_mirror(
+    catalog_image="cp.stg.icr.io/cp/df/isf-data-foundation-catalog:v4.20",
+    mirror_registry="your-registry.com:5000/fdf",
+    configure_registries=True  # Optional: configure registries.conf
+)
+```
+
+**Notes:**
+- Use `python -m ocs_ci.framework.fdf_mirror.main` for reliability in Jenkins
```
dahorak: What is the reason for using `python -m ocs_ci.framework.fdf_mirror.main` instead of the new entry point `fdf-mirror`?

ocs_ci/deployment/disconnected.py
```diff
@@ -256,31 +309,68 @@ def mirror_index_image_via_oc_mirror(index_image, packages, idms=None):
...
         os.symlink(pull_secret_path, os.path.expanduser(docker_config_file))
 
+    # Configure registries.conf if requested (for FDF internal images)
+    if configure_registries:
+        logger.info(
+            "Configuring registry mirrors for internal images using registries.conf.d/"
+        )
+        registries_template = os.path.join(
+            constants.FDF_TEMPLATE_DIR, "registries.conf.template"
+        )
```
dahorak: Here it would make sense to pass the template via the parameter from outside of this function (as the template is quite specific for FDF, while the function is generic).

ocs_ci/deployment/disconnected.py
```diff
@@ -256,31 +309,68 @@ def mirror_index_image_via_oc_mirror(index_image, packages, idms=None):
...
-    )
+    logger.info(f"Mirror required images to mirror registry {mirror_registry}")
 
     cmd = (
         f"oc mirror --config {imageset_config_file} "
-        f"docker://{config.DEPLOYMENT['mirror_registry']} "
-        "--workspace file://oc-mirror-workspace/results-files --v2"
+        f"docker://{mirror_registry} "
+        "--workspace file://oc-mirror-workspace/results-files --v2 "
+        "--dest-tls-verify=false --image-timeout 30m"
```
dahorak, line 380: Not sure, if we should add the `--dest-tls-verify=false` here without an option to enable/disable it. If there is an option to import the registry certificate, it will be better to not disable the tls verification.

ocs_ci/deployment/disconnected.py
```diff
@@ -329,10 +419,10 @@ def mirror_index_image_via_oc_mirror(index_image, packages, idms=None):
         "working-dir/cluster-resources/idms-oc-mirror.yaml",
     )
 
-    # make idms name unique - append run_id
+    # make idms name unique - append run_id with configurable prefix
```
dahorak, line 429: not sure, if the note `with configurable prefix` is necessary here - it suggests that the prefix is related to the run_id, but it is just cosmetic note... and it is quite clear from the code

ocs_ci/deployment/disconnected.py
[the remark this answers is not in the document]
petr-balogh: @rh-aaleem please use with as suggested

ocs_ci/deployment/disconnected.py
```diff
@@ -256,31 +311,66 @@ def mirror_index_image_via_oc_mirror(index_image, packages, idms=None):
...
+                temp_file = tempfile.NamedTemporaryFile(
+                    mode="w", delete=False, suffix=".conf"
+                )
+                temp_file.write(registries_content)
+                temp_file.close()
+
+                # Copy to registries.conf.d/ directory and set readable permissions
+                exec_cmd(f"sudo cp {temp_file.name} {ocs_ci_conf_file}")
+                exec_cmd(f"sudo chmod 644 {ocs_ci_conf_file}")
+                os.unlink(temp_file.name)
```
rh-aaleem: Remove the file at the end.

ocs_ci/utility/framework/fusion_fdf_init.py
```diff
@@ -32,6 +32,7 @@
 LOG_NAMES = {
     "fusion": "fusion_deployment",
     "fdf": "fusion_data_foundation_deployment",
+    "fdf-mirror": "fdf_catalog_mirroring",
```
clacroix12: Since this module is intended to be used for Fusion and FDF deployment initialization code, I don't think this is the appropriate place to include logic for the `fdf-mirror` entry point. If there is some functionality here that would be beneficial to be shared, we can consider refactoring. That being said, I think it would be best to create functions specific to the `fdf-mirror` entry points requirements in it's own main.py.

As this is deployment related code, it's likely that changes to this module will need to be backported as far back as release-4.18 so I would like to avoid including code that is not deployment related. This will also have the benefit of not being tied to assumptions being made by the deployment code, such as the report generation.

clacroix12 (changes requested):

ocs_ci/deployment/disconnected.py
```diff
@@ -256,31 +311,71 @@ def mirror_index_image_via_oc_mirror(index_image, packages, idms=None):
...
+        logger.info(
+            "Configuring registry mirrors for internal images using registries.conf.d/"
+        )
+        if registries_template and os.path.exists(registries_template):
+            try:
+                with open(registries_template, "r") as f:
+                    registries_content = f.read()
+
+                # Use registries.conf.d/ directory
+                registries_conf_d_dir = "/etc/containers/registries.conf.d"
```
petr-balogh: Let's try to use path from userspace which will not require sudo
```
from pathlib import Path

# 1. Define the path cleanly using the / operator
registries_dir = Path.home() / ".config" / "containers" / "registries.conf.d"

# 2. Pythonic way to ensure the directory exists before using it
registries_dir.mkdir(parents=True, exist_ok=True)
```

docs/usecases/fdf_mirroring.md
```diff
@@ -0,0 +1,435 @@
+# FDF Image Mirroring
+
+Guide for mirroring FDF (Fusion Data Foundation) unreleased images from Mgen to a disconnected mirror registry.
```
clacroix12: What is `Mgen`?

docs/usecases/fdf_mirroring.md
```diff
@@ -0,0 +1,435 @@
...
+The report includes:
+- Test suite properties (cluster name, OCP version, catalog image, etc.)
+- Test case results (success/failure status)
+- Execution time and timestamps
+- Error details if mirroring fails
+
+This is particularly useful for:
+- Jenkins/GitLab CI pipeline integration
+- Automated test result tracking
+- Failure analysis and debugging
```
clacroix12: I don't think this is true anymore since we've changed the parent class. Let's remove this section and other references to `--report` or the junit reporting.

Let's also ensure the rest of the documentation matches the change in parent class as this has impacted CLI params and capabilities of the entry-point.

ocs_ci/framework/fdf_mirror/main.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+
+        logger.info(f"FDF catalog successfully mirrored to: {mirrored_image}")
+        logger.info("Mirroring completed successfully!")
+
+        return 0
+
+    except Exception as e:
+        logger.exception(f"FDF catalog mirroring failed: {e}")
+        return 1
```
clacroix12: Since this is the main entry point, can't we remove the `try`/`except` and let the error propagate normally? Or are we doing this to preserve the execution of the pipeline?

clacroix12 (commented):
A few issues with the doc and a question about the main function, but overall looks good to me. Coderabbit also left a few comments that I think are worth considering.

ocs_ci/deployment/disconnected.py
```diff
@@ -37,6 +38,46 @@
...
+    # Call the generic function with FDF-specific parameters
+    # mirror_registry is passed directly - no need to modify global config
+    return mirror_index_image_via_oc_mirror(
+        index_image=catalog_image,
+        packages=None,  # Mirror entire FDF catalog (no package filtering)
+        idms=None,
+        idms_name_prefix="fdf",  # Use 'fdf' prefix for IDMS naming
+        configure_registries=configure_registries,  # FDF-specific feature
+        mirror_registry=mirror_registry,
+        registries_template=constants.FDF_REGISTRIES_CONF_TEMPLATE,
```
petr-balogh, line 77: Maybe in the future we can get this value from config - and document this value - and only if not specified we will use the one from constants

ocs_ci/framework/fdf_mirror/main.py
rh-aaleem: using the try/except for just better logging the error, if some error occurred it will still print the stack trace and cancel the execution of pipeline,

clacroix12: It seems all we're adding is `FDF catalog mirroring failed`, I would suggest to just let the error do it's thing instead of wrapping it in another string. We could leave the try/except if you want to include this so we can easily find if it did fail in the logs, but I think I would prefer it to be it's own log statement prior to raising the error.

This way if we do need something else to do something when we encounter that error, or a specific error that might be raised, we will be able to catch it properly instead of it being eaten by this catch all block.

clacroix12 (dismissed):

clacroix12 (approved):

petr-balogh (approved):

PR #15187, opened 2026-05-19
Wait for NooBaa DB readiness after endpoint max count change

sagihirshfeld:
## Motivation
Endpoint scaling (maxCount change) triggers CNPG to restart PostgreSQL
pods sequentially ([DFBUGS-6888](https://redhat.atlassian.net/browse/DFBUGS-6888)). The `nb_ensure_endpoint_count` fixture
sets maxCount from 1→2 on the first MCG test in a session, causing a DB
restart that often fails the next test — frequently seen as the first MCG
failure in acceptance launches.

## Changes
- Add `wait_for_noobaa_db_ready()` in `ocs_ci/ocs/resources/pod.py` that
  checks all DB pods are container-ready (1/1) for multiple consecutive
  checks, handling both the propagation delay and sequential pod restarts.
  Expected pod count is derived from the CNPG Cluster CR's `spec.instances`.
- Call it after endpoint scaling in `nb_ensure_endpoint_count` and in
  `test_disable_mcg_external_service`'s fixture and finalizer.

## Summary by CodeRabbit

* **New Features**
  * Added a helper to wait for NooBaa database pods to be ready and remain stable, with configurable timeout, polling interval, and required consecutive ready checks (includes an initial restart-grace phase).
* **Tests**
  * Updated test setup/fixtures to synchronize with NooBaa database pod readiness: now waits before/after endpoint scaling and during related patching/cleanup flows.

### DIFF
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -2417,7 +2417,8 @@ def wait_for_noobaa_db_ready(timeout=300, sleep=10, stability_count=3):
     Wait for all NooBaa DB pods to be container-ready (1/1) and stable.
 
     Operations like endpoint scaling can trigger CNPG to rolling-restart
-    DB pods. This function requires all DB pods to report ready for
+    DB pods. This function first waits for the restart to begin (at least
+    one pod goes not-ready), then requires all DB pods to report ready for
     multiple consecutive checks, ensuring the pods are genuinely stable
     and not mid-rolling-restart.
 
@@ -2453,8 +2454,28 @@ def wait_for_noobaa_db_ready(timeout=300, sleep=10, stability_count=3):
         for pod_info in pods:
             container_statuses = pod_info.get("status", {}).get("containerStatuses")
             if not container_statuses or not container_statuses[0].get("ready"):
+                pod_name = pod_info.get("metadata", {}).get("name", "unknown")
+                logger.info(f"Pod {pod_name} container not ready: {container_statuses}")
                 return False
         return True
+
+    # Wait for the rolling restart to begin before checking stability.
+    # Without this, the stability checks could pass before the CNPG
+    # operator even starts restarting pods.
+    restart_grace_timeout = 60
+    logger.info(f"Waiting up to {restart_grace_timeout}s for a DB pod restart to begin")
+    try:
+        for not_ready in TimeoutSampler(
+            restart_grace_timeout, sleep, _all_db_pods_ready
+        ):
+            if not not_ready:
+                logger.info("Detected DB pod restart, proceeding to stability checks")
+                break
+    except TimeoutExpiredError:
+        logger.info(
+            "No DB pod restart detected within the grace period, "
+            "proceeding to stability checks"
+        )
 
     try:
         for is_ready in TimeoutSampler(timeout, sleep, _all_db_pods_ready):
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -2465,10 +2465,10 @@ def wait_for_noobaa_db_ready(timeout=300, sleep=10, stability_count=3):
     restart_grace_timeout = 60
     logger.info(f"Waiting up to {restart_grace_timeout}s for a DB pod restart to begin")
     try:
-        for not_ready in TimeoutSampler(
+        for pods_ready in TimeoutSampler(
             restart_grace_timeout, sleep, _all_db_pods_ready
         ):
-            if not not_ready:
+            if not pods_ready:
                 logger.info("Detected DB pod restart, proceeding to stability checks")
                 break
     except TimeoutExpiredError:
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -2411,6 +2411,72 @@ def wait_for_noobaa_pods_running(timeout=300, sleep=10):
     sampler.wait_for_func_status(True)
 
 
+@config.run_with_provider_context_if_available
+def wait_for_noobaa_db_ready(timeout=300, sleep=10, stability_count=3):
+    """
+    Wait for all NooBaa DB pods to be container-ready (1/1) and stable.
+
+    Operations like endpoint scaling can trigger CNPG to rolling-restart
+    DB pods. This function requires all DB pods to report ready for
+    multiple consecutive checks, ensuring the pods are genuinely stable
+    and not mid-rolling-restart.
+
+    The expected pod count is derived from the CNPG Cluster CR's
+    spec.instances field.
+
+    Args:
+        timeout (int): Overall timeout in seconds
+        sleep (int): Seconds between checks
+        stability_count (int): Number of consecutive ready checks required
+
+    """
+    namespace = config.ENV_DATA["cluster_namespace"]
+    stable_checks = 0
+
+    cnpg_cluster = OCP(
+        kind=constants.CNPG_CLUSTER_KIND,
+        resource_name=constants.NB_DB_CNPG_CLUSTER_NAME,
+        namespace=namespace,
+    )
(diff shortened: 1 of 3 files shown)


tests/cross_functional/kcs/test_disable_mcg_external_service.py
```diff
@@ -47,6 +48,7 @@ def patch_storagecluster_object(self, request):
             params='{"spec": {"multiCloudGateway": {"endpoints": {"minCount": 2,"maxCount": 4}}}}',
```
posharma1, line 48: there is no key multiCloudGateway in noobaa CR this should be changed to params='{"spec":  {"endpoints": {"minCount": 2,"maxCount": 4}}}',

sagihirshfeld: You are right, and this makes me wonder if this test is even passing. However, this is not something I've added as part of this PR. The needed change (and its verification) is out of scope for this PR which is just about adding the needed safeguard in case the upscale/downscale does work.

posharma1: ok, np

ocs_ci/ocs/resources/pod.py
```diff
@@ -2458,6 +2411,72 @@ def _check_nb_pods_status():
...
+                f"Found {len(pods)} NooBaa DB pods, waiting for {expected_instances}"
+            )
+            return False
+        for pod_info in pods:
+            container_statuses = pod_info.get("status", {}).get("containerStatuses")
+            if not container_statuses or not container_statuses[0].get("ready"):
+                return False
+        return True
+
+    try:
```
ypersky1980, line 2480: Race condition: stability checks may pass before the CNPG restart even begins.

After the endpoint scaling patch is applied, the CNPG operator needs time to detect the change, reconcile, and initiate the rolling restart. But wait_for_noobaa_db_ready() starts stability checks immediately. With defaults (stability_count=3, sleep=10), the function needs only 30 seconds of consecutive "all pods ready" to declare success. If the CNPG operator takes longer than 30 seconds to begin the first pod restart (operator reconciliation delays, queue contention), the function will return "stable and ready" before any disruption has started — completely defeating its purpose.

Suggestion: Add an initial phase that waits for at least one pod to go not-ready (confirming the restart has actually begun), with a shorter timeout, before starting stability checks. Or at minimum, add a brief initial sleep/grace period and document the assumption about operator reconciliation latency.

ocs_ci/ocs/resources/pod.py
```diff
@@ -2458,6 +2411,72 @@ def _check_nb_pods_status():
...
+    Args:
+        timeout (int): Overall timeout in seconds
+        sleep (int): Seconds between checks
+        stability_count (int): Number of consecutive ready checks required
+
+    """
+    namespace = config.ENV_DATA["cluster_namespace"]
+    stable_checks = 0
+
+    cnpg_cluster = OCP(
```
ypersky1980, line 2437: No OCS version guard; will break on pre-4.19 clusters
  
  The function unconditionally uses CNPG-specific constants (CNPG_CLUSTER_KIND, NB_DB_CNPG_CLUSTER_NAME, NOOBAA_DB_LABEL_419_AND_ABOVE). Throughout the codebase, other code consistently gates these behind version checks, e.g. in test_host_node_failure.py:

  if ocs_version >= version.VERSION_4_19:
      noobaa_db_label = constants.NOOBAA_DB_LABEL_419_AND_ABOVE

  The nb_ensure_endpoint_count fixture in conftest.py can run on any OCS version. On pre-4.19 clusters (where NooBaa uses a non-CNPG DB), cnpg_cluster.get() will fail because the CNPG Cluster CR doesn't exist.

  Suggestion: Add a version check at the top of wait_for_noobaa_db_ready() to skip/return early on < VERSION_4_19, or add a guard in the callers.

ocs_ci/ocs/resources/pod.py
```diff
@@ -2458,6 +2411,72 @@ def _check_nb_pods_status():
...
+        )
+        if len(pods) < expected_instances:
+            logger.info(
+                f"Found {len(pods)} NooBaa DB pods, waiting for {expected_instances}"
+            )
+            return False
+        for pod_info in pods:
+            container_statuses = pod_info.get("status", {}).get("containerStatuses")
+            if not container_statuses or not container_statuses[0].get("ready"):
+                return False
```
ypersky1980, line 2459: Silent failure gives no diagnostic info when pods aren't ready, making timeouts hard to debug
  
  When _all_db_pods_ready() finds the right number of pods but a container isn't ready, it silently returns False with no log message. The only log is when the pod count is too low (line 2450).
  This means that when the function times out after 300 seconds, you get:
  NooBaa DB pods did not stabilize within 300s. Achieved 0/3 consecutive ready checks.
  ...but no indication of which pod is not ready or what its container status is. In a QE framework, this makes debugging significantly harder.

  Suggestion: Add a log line in the for pod_info in pods loop when a pod's container isn't ready:

  pod_name = pod_info.get("metadata", {}).get("name", "unknown")
  logger.info(f"Pod {pod_name} container not ready: {container_statuses}")
  return False

ypersky1980 (commented):
Left comments.

ocs_ci/ocs/resources/pod.py
sagihirshfeld: This function uses CNPG-specific constants which only exist on 4.19+, but that's fine here - this PR targets the master branch which only runs against 4.19 and above. If it were ever backported (unlikely, and 4.18 at the earliest wouldn't apply), a version guard would be added at that point. No need to add defensive code for a scenario that doesn't exist on this branch.

ocs_ci/ocs/resources/pod.py
```diff
@@ -2411,6 +2411,72 @@ def _check_nb_pods_status():
...
```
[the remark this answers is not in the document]
sagihirshfeld: NooBaa DB pods are managed by CNPG and consistently have a single container - they're straightforward PostgreSQL database pods. There's no realistic scenario where a sidecar or additional container would be injected into these pods, so checking `containerStatuses[0]` is sufficient. Adding a loop for a hypothetical multi-container case would be over-engineering.

ocs_ci/ocs/resources/pod.py
sagihirshfeld: Fixed in c3a4063ef9b66284a659ee1079f5010a33471a34

ocs_ci/ocs/resources/pod.py
sagihirshfeld: Fixed in c3a4063ef9b66284a659ee1079f5010a33471a34

ypersky1980 (dismissed):
LGTM

ocs_ci/ocs/resources/pod.py
```diff
@@ -2411,6 +2411,93 @@ def _check_nb_pods_status():
...
+    # Without this, the stability checks could pass before the CNPG
+    # operator even starts restarting pods.
+    restart_grace_timeout = 60
+    logger.info(f"Waiting up to {restart_grace_timeout}s for a DB pod restart to begin")
+    try:
+        for not_ready in TimeoutSampler(
+            restart_grace_timeout, sleep, _all_db_pods_ready
+        ):
+            if not not_ready:
+                logger.info("Detected DB pod restart, proceeding to stability checks")
```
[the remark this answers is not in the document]
udaysk23: @sagihirshfeld Are you planning to add fix for this?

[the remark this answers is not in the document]
sagihirshfeld: Fixed in 651b68eaa19bbc2ec3ebcef8c73953c5b2ddace4

udaysk23 (approved):

ypersky1980 (approved):
LGTM

posharma1 (approved):

PR #15188, opened 2026-05-19
Add Ceph health monitoring during RDR StorageCluster wait

petr-balogh:
During RDR deployment with globalnet, StorageCluster can get stuck in "Progressing" state due to Ceph monitor slow ops issues, preventing the cluster from reaching "Ready" phase. This causes deployment to timeout after waiting for the Ready state.

This adds a new function wait_for_storagecluster_ready_with_health_check() that wraps the wait_for_phase() call with periodic Ceph health checks (every 60s). When slow ops are detected, it automatically:
- Collects MustGather logs
- Restarts the affected MON pods (fixes Paxos commit deadlock)
- Updates Jira [DFBUGS-2456](https://redhat.atlassian.net/browse/DFBUGS-2456) with deployment details
- Continues waiting for Ready phase after recovery

Applied specifically to regional-dr deployment flow where slow ops issues are known to occur during monitor topology changes.

Fixes: [DFBUGS-2456](https://redhat.atlassian.net/browse/DFBUGS-2456)

### DIFF
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -3586,3 +3586,101 @@ def get_deviceset_name_per_count():
     """
     device_sets = get_all_device_sets()
     return {d.get("name"): d["count"] for d in device_sets}
+
+
+def wait_for_storagecluster_ready_with_health_check(
+    storage_cluster,
+    timeout=1000,
+    sleep=30,
+    fix_ceph_health=True,
+    update_jira=True,
+):
+    """
+    Wait for StorageCluster to reach Ready phase with periodic Ceph health checks.
+    This function is particularly useful for RDR deployments where slow ops issues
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -3586,3 +3586,101 @@ def get_deviceset_name_per_count():
...
+        current_time = time.time()
+        if current_time - last_health_check >= health_check_interval:
+            log.info("Performing periodic Ceph health check...")
+            try:
+                ceph_health_check(
+                    namespace=storage_cluster.namespace,
+                    tries=3,
+                    delay=30,
+                    fix_ceph_health=fix_ceph_health,
+                    update_jira=update_jira,
```
vavuthu, line 3651: It updates the jira for all the periodic checks which is not needed. We just need to update once in jira.

PR #15195, opened 2026-05-20
Adds test for s3 vector Happy path verification

udaysk23:
Adds test for s3 vector Happy path verification

## Summary by CodeRabbit

* **New Features**
  * S3 Vectors support: vector buckets, a vector-backed OBC/bucket type, and an OBC manifest template.
  * Client & endpoint initialization for S3 Vectors.
  * Vector index and data operations plus similarity (top-K) queries.
  * Helpers for bulk vector deletion and bucket policy management.

* **Bug Fixes / Improvements**
  * Stricter NSFS PVC/ULS name selection to avoid invalid names.
  * New OC-backed vector bucket type with health verification behavior.

* **Tests**
  * New fixtures and end-to-end tests covering full S3 Vector workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/red-hat-storage/ocs-ci/pull/15195?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/ocs_ci/ocs/resources/bucketclass.py
+++ b/ocs_ci/ocs/resources/bucketclass.py
@@ -129,8 +129,22 @@ def bucket_class_factory(
             backingstores = None
             namespacestores = None
             placement_policy = None
+            vector_policy = None
 
-            if "namespace_policy_dict" in bucket_class_dict:
+            if "vector_policy" in bucket_class_dict and interface.lower() != "oc":
+                raise RuntimeError(
+                    f"vector_policy is supported only with 'oc' interface, got: {interface}"
+                )
+
+            if "vector_policy" in bucket_class_dict:
+                # Vector policy for vector buckets
+                vector_policy = {
+                    "resource": bucket_class_dict["vector_policy"]["resource"],
+                    "vectorDBType": bucket_class_dict["vector_policy"].get(
+                        "vector_db_type", "lance"
+                    ),
+                }
+            elif "namespace_policy_dict" in bucket_class_dict:
                 if "namespacestore_dict" in bucket_class_dict["namespace_policy_dict"]:
                     nss_dict = bucket_class_dict["namespace_policy_dict"][
                         "namespacestore_dict"
@@ -222,6 +236,7 @@ def bucket_class_factory(
                     placement_policy,
                     namespace_policy,
                     replication_policy,
+                    vector_policy,
                 )
             elif interface.lower() == "cli" and backingstores:
                 mcg_obj.cli_create_bucketclass_over_backingstores(
--- a/ocs_ci/ocs/resources/mcg.py
+++ b/ocs_ci/ocs/resources/mcg.py
@@ -73,7 +73,9 @@ class MCG:
         noobaa_user,
         noobaa_password,
         data_to_mask,
-    ) = (None,) * 12
+        vectors_endpoint,
+        vectors_internal_endpoint,
+    ) = (None,) * 14
 
     def __init__(self, *args, **kwargs):
         """
@@ -157,6 +159,24 @@ class MCG:
                 .get("serviceMgmt")
                 .get("externalDNS")[0]
             ) + "/rpc"
+
+            # serviceVectors is for S3 Vectors operations
+            service_vectors = (
+                get_noobaa.get("items")[0]
+                .get("status")
+                .get("services")
+                .get("serviceVectors")
+            )
+            if service_vectors:
+                self.vectors_endpoint = service_vectors.get("externalDNS")[0]
+                self.vectors_internal_endpoint = service_vectors.get("internalDNS")[0]
+                logger.info(f"S3 Vectors endpoint: {self.vectors_endpoint}")
+            else:
+                logger.warning(
+                    "serviceVectors not found in NooBaa services - S3 Vectors endpoint will not be available"
+                )
+                self.vectors_endpoint = None
+                self.vectors_internal_endpoint = None
 
             # serviceIam is not available on some platforms (e.g., Azure ARO)
             service_iam = (
@@ -651,15 +671,19 @@ class MCG:
         placement_policy,
         namespace_policy,
         replication_policy,
+        vector_policy=None,
     ):
         """
         Creates a new NooBaa bucket class using a template YAML
+
         Args:
             name (str): The name to be given to the bucket class
             backingstores (list): The backing stores to use as part of the policy
             placement_policy (str): The placement policy to be used - Mirror | Spread
             namespace_policy (dict): The namespace policy to be used
             replication_policy (dict): The replication policy dictionary
+            vector_policy (dict): The vector policy for vector buckets with
+                resource and vectorDBType fields
 
         Returns:
             OCS: The bucket class resource
@@ -711,6 +735,12 @@ class MCG:
                     else json.dumps({"rules": replication_policy})
                 ),
             )
+
+        if vector_policy:
+            bc_data["spec"]["vectorPolicy"] = {
+                "resource": vector_policy["resource"],
+                "vectorDBType": vector_policy["vectorDBType"],
+            }
 
         return create_resource(**bc_data)
 
--- a/ocs_ci/ocs/resources/namespacestore.py
+++ b/ocs_ci/ocs/resources/namespacestore.py
@@ -452,9 +452,14 @@ def namespace_store_factory(
                 for nss_tup in nss_lst:
                     for _ in range(nss_tup[0] if isinstance(nss_tup[0], int) else 1):
                         if platform.lower() == "nsfs":
-                            uls_name = nss_tup[0] or create_unique_resource_name(
-                                constants.PVC.lower(), platform
-                            )
+                            # Use nss_tup[0] as PVC name only if it's a string
+                            # If it's an int (amount) or None, generate a unique name
+                            if isinstance(nss_tup[0], str) and nss_tup[0].strip():
+                                uls_name = nss_tup[0].strip()
+                            else:
+                                uls_name = create_unique_resource_name(
+                                    constants.PVC.lower(), platform
+                                )
                             pvc_factory_session(
                                 custom_data=template_pvc(uls_name, size=nss_tup[1])
                             )
--- a/ocs_ci/ocs/resources/objectbucket.py
+++ b/ocs_ci/ocs/resources/objectbucket.py
@@ -169,6 +169,7 @@ class ObjectBucket(ABC):
     """
 
     mcg, name = (None,) * 2
+    _verify_s3_listing = True  # Set to False in subclasses to skip S3 bucket check
 
     def __init__(
         self,
@@ -312,7 +313,11 @@ class ObjectBucket(ABC):
             ):
                 if not health_check:
                     logger.info(f"{self.name} is unhealthy. Rechecking.")
-                elif self.mcg and not self.mcg.s3_verify_bucket_exists(self.name):
+                elif (
+                    self._verify_s3_listing
+                    and self.mcg
+                    and not self.mcg.s3_verify_bucket_exists(self.name)
+                ):
                     logger.info(
                         f"{self.name} is healthy, but not found in S3. Rechecking."
                     )
@@ -626,6 +631,36 @@ class MCGOCBucket(OCBucket):
         create_resource(**obc_data)
 
 
+class VectorOCBucket(OCBucket):
+    """
+    Implementation of an MCG Vector bucket using the OC CLI
+    """
+
+    _verify_s3_listing = False  # Skip S3 bucket existence check for vector buckets
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        obc_data = templating.load_yaml(constants.VECTOR_OBC_YAML)
+        if self.name is None:
+            self.name = create_unique_resource_name("vector-oc", "obc")
+        obc_data["metadata"]["name"] = self.name
+        obc_data["spec"]["bucketName"] = self.name
+        obc_data["spec"]["storageClassName"] = f"{self.namespace}.noobaa.io"
+        obc_data["metadata"]["namespace"] = self.namespace
+
+        # Ensure additionalConfig exists
+        obc_data.setdefault("spec", {}).setdefault("additionalConfig", {})
+
+        # Always set bucketType to vector
+        obc_data["spec"]["additionalConfig"]["bucketType"] = "vector"
+
+        # Add bucketclass if specified
+        if self.bucketclass:
+            obc_data["spec"]["additionalConfig"]["bucketclass"] = self.bucketclass.name
+
+        create_resource(**obc_data)
+
+
 class RGWOCBucket(OCBucket):
     """
     Implementation of an RGW bucket using the S3 API
@@ -743,4 +778,5 @@ BUCKET_MAP = {
     "cli": MCGCLIBucket,
     "rgw-oc": RGWOCBucket,
     "mcg-namespace": MCGNamespaceBucket,
+    "vector-oc": VectorOCBucket,
 }
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -66,6 +66,11 @@ from ocs_ci.ocs.bucket_utils import (
     update_replication_policy,
     put_bucket_versioning_via_awscli,
 )
+from ocs_ci.ocs.vector_utils import (
+    list_indexes,
+    delete_index,
+    create_s3vectors_client,
+)
 from ocs_ci.ocs.cnv.virtual_machine import VirtualMachine, VMCloner
 from ocs_ci.ocs.dr.dr_workload import (
     BusyBox,
@@ -126,7 +131,7 @@ from ocs_ci.ocs.resources.cloud_uls import (
 )
 from ocs_ci.ocs.node import check_nodes_specs
 from ocs_ci.ocs.resources.mcg import MCG
-from ocs_ci.ocs.resources.objectbucket import BUCKET_MAP
+from ocs_ci.ocs.resources.objectbucket import BUCKET_MAP, OBC
 from ocs_ci.ocs.resources.ocs import OCS
 from ocs_ci.ocs.resources.pod import (
     get_rgw_pods,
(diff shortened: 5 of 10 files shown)


ocs_ci/ocs/resources/mcg.py
```diff
@@ -158,6 +161,24 @@ def __init__(self, *args, **kwargs):
...
+            # serviceVectors is for S3 Vectors operations
+            service_vectors = (
+                get_noobaa.get("items")[0]
+                .get("status")
+                .get("services")
+                .get("serviceVectors")
+            )
+            if service_vectors:
+                self.vectors_endpoint = service_vectors.get("externalDNS")[0]
+                self.vectors_internal_endpoint = service_vectors.get("internalDNS")[0]
```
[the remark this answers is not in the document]
udaysk23: If service_vectors is present in cluster then by default it will have `externalDNS` and `internalDNS`

tests/conftest.py
```diff
@@ -3625,6 +3625,220 @@ def bucket_cleanup():
...
+                    *args,
+                    **kwargs,
+                )
+                current_call_created_buckets.append(created_bucket)
+                created_buckets.append(created_bucket)
+                if verify_health:
+                    created_bucket.verify_health(
+                        timeout=kwargs.pop("timeout") if "timeout" in kwargs else 180,
+                        **kwargs,
+                    )
```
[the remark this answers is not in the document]
udaysk23: Fixed

ocs_ci/ocs/resources/mcg.py
```diff
@@ -1266,6 +1296,16 @@ def update_s3_creds(self):
...
 
+        self.s3vectors_client = boto3.client(
+            "s3vectors",
+            region_name=self.region,
+            verify=retrieve_verification_mode(),
+            endpoint_url=self.vectors_endpoint,
+            aws_access_key_id=self.access_key_id,
+            aws_secret_access_key=self.access_key,
+            config=retry_cfg,
+        )
```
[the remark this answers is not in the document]
udaysk23: fixed

ocs_ci/ocs/resources/namespacestore.py
```diff
@@ -452,9 +452,14 @@ def _create_nss(method, nss_dict):
...
                     for _ in range(nss_tup[0] if isinstance(nss_tup[0], int) else 1):
                         if platform.lower() == "nsfs":
-                            uls_name = nss_tup[0] or create_unique_resource_name(
-                                constants.PVC.lower(), platform
-                            )
+                            # Use nss_tup[0] as PVC name only if it's a string
+                            # If it's an int (amount) or None, generate a unique name
+                            if isinstance(nss_tup[0], str):
+                                uls_name = nss_tup[0]
+                            else:
```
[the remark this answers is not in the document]
udaysk23: Fixed

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -0,0 +1,206 @@
...
+        logger.info(f"✓ Deleted {deleted_count} vectors")
+
+        # Verify vectors are deleted
+        vectors_after_delete = vector_utils.list_vectors(
+            mcg_obj, vectorBucketName=vector_bucket_name, indexName=index_name
+        )
+        remaining = vectors_after_delete.get("Vectors", [])
+        assert (
+            len(remaining) == 0
+        ), f"Vectors still exist after deletion: {len(remaining)}"
```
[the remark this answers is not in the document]
udaysk23: fixed

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -0,0 +1,206 @@
...
+        # Verify bucket is deleted
+        buckets_after_delete = vector_utils.list_vector_buckets(mcg_obj)
+        remaining_buckets = [
+            bucket["VectorBucketName"]
+            for bucket in buckets_after_delete.get("VectorBuckets", [])
+        ]
+        assert (
+            vector_bucket_name not in remaining_buckets
+        ), f"Vector bucket {vector_bucket_name} still exists after deletion"
+        logger.info(f"✓ Verified vector bucket deletion")
```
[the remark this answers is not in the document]
udaysk23: fixed

ocs_ci/ocs/resources/namespacestore.py
```diff
@@ -452,9 +452,14 @@ def _create_nss(method, nss_dict):
                 for nss_tup in nss_lst:
                     for _ in range(nss_tup[0] if isinstance(nss_tup[0], int) else 1):
                         if platform.lower() == "nsfs":
-                            uls_name = nss_tup[0] or create_unique_resource_name(
-                                constants.PVC.lower(), platform
-                            )
+                            # Use nss_tup[0] as PVC name only if it's a string
+                            # If it's an int (amount) or None, generate a unique name
+                            if isinstance(nss_tup[0], str) and nss_tup[0].strip():
+                                uls_name = nss_tup[0]
```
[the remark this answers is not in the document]
udaysk23: Fixed

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -0,0 +1,205 @@
...
+        remaining_buckets = [
+            bucket["VectorBucketName"]
+            for bucket in buckets_after_delete.get("VectorBuckets", [])
+        ]
+        assert (
+            vector_bucket_name not in remaining_buckets
+        ), f"Vector bucket {vector_bucket_name} still exists after deletion"
+        logger.info("✓ Verified vector bucket deletion")
+
+        logger.info("✅ All S3 vector operations completed successfully!")
```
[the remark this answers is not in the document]
udaysk23: Cleanup is handled in fixture for early failure

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -0,0 +1,208 @@
...
+            "Creating vector bucket (auto-creates nsfs + vectorPolicy bucketclass)"
+        )
+        vector_bucket_name = vector_bucket_factory(amount=1)[0].name
+        logger.info(f"✓ Vector bucket created: {vector_bucket_name}")
+
+        # Step 2: Validate vector bucket exists and update mcg_obj withOBC creds
+        logger.info("Validating vector bucket exists and logging OBC credentials")
+        obc_obj = OBC(vector_bucket_name)
+        mcg_obj.access_key = obc_obj.access_key
+        mcg_obj.access_key_id = obc_obj.access_key_id
```
ypersky1980: @udaysk23  This credential swap is inconsistent. 

After these two lines:
  - `mcg_obj.access_key` / `mcg_obj.access_key_id` → OBC creds
  - `mcg_obj.s3vectors_client` → still initialized with the **original** admin creds (set during `update_s3_creds()`)

  So all subsequent `vector_utils.*` calls in this test still use admin auth via the unchanged `s3vectors_client`, while these two attributes now hold OBC creds that nothing reads.

  Either:
  1. Remove these two lines if admin creds work for vector operations (they're misleading as-is), or
  2. If OBC creds are required, reinitialize `s3vectors_client` after the swap (e.g. call `mcg_obj.update_s3_creds()` or rebuild the boto3 client)

ocs_ci/ocs/resources/namespacestore.py
ypersky1980: @udaysk23  The condition on the line above correctly checks `.strip()`, but the assignment here uses the original unstripped value. A whitespace-padded input like `" my-pvc "` passes the check but
  creates an invalid K8s resource name.

  Should be:
      uls_name = nss_tup[0].strip()

ocs_ci/ocs/resources/bucketclass.py
```diff
@@ -129,8 +129,22 @@ def _create_bucket_class(bucket_class_dict):
...
-            if "namespace_policy_dict" in bucket_class_dict:
+            if "vector_policy" in bucket_class_dict and interface.lower() != "oc":
+                raise RuntimeError(
+                    f"vector_policy is supported only with 'oc' interface, got: {interface}"
+                )
+
+            if "vector_policy" in bucket_class_dict:
+                # Vector policy for vector buckets
+                vector_policy = {
+                    "resource": bucket_class_dict["vector_policy"]["resource"],
```
ypersky1980, line 142: @udaysk23   If `vector_policy` is present but missing the `resource` key, this raises a raw `KeyError`. Consider validating upfront, e.g.:

      vp = bucket_class_dict["vector_policy"]
      if "resource" not in vp:
          raise ValueError("vector_policy must include a 'resource' field")

tests/functional/object/mcg/test_s3_vector.py
udaysk23: Vector operation needs OBC creds and it wont work with noobaa admin keys yet(Dev confirmed that it is not supported)
Right now, Above code overwrites mcg creds with OBC creds and I dont see any code which is resetting the credentials in this code.
 
If we consider, update_s3_creds() method, it will again assign noobaa admin creds.

ocs_ci/ocs/resources/bucketclass.py
udaysk23: @ypersky1980 , In our CI suite, script always pass `resource` info while creating bucketclass and there is no user intervention possible here.
When `vector_bucket_factory_fixture` is called, it will create NSFS and that info will get passed to `bucketclass fixture` with all info.

At this point, I dont think we need to consider one more check for already present info

ocs_ci/ocs/resources/namespacestore.py
udaysk23: Fixed

ocs_ci/ocs/resources/mcg.py
```diff
@@ -158,6 +161,24 @@ def __init__(self, *args, **kwargs):
...
+                .get("status")
+                .get("services")
+                .get("serviceVectors")
+            )
+            if service_vectors:
+                self.vectors_endpoint = service_vectors.get("externalDNS")[0]
+                self.vectors_internal_endpoint = service_vectors.get("internalDNS")[0]
+                logger.info(f"S3 Vectors endpoint: {self.vectors_endpoint}")
+            else:
+                logger.info(
```
sagihirshfeld:
```suggestion
                logger.warning(
```

ocs_ci/ocs/resources/objectbucket.py
```diff
@@ -627,6 +627,67 @@ def __init__(self, *args, **kwargs):
...
+                f"{self.name} did not reach a healthy state within {timeout} seconds."
+            )
+            obc_obj = OCP(kind="obc", namespace=self.namespace, resource_name=self.name)
+            obc_yaml = obc_obj.get()
+            obc_description = obc_obj.describe(resource_name=self.name)
+            raise UnhealthyBucket(
+                f"{self.name} did not reach a healthy state within {timeout} seconds.\n"
+                f"OBC YAML:\n{json.dumps(obc_yaml, indent=2)}\n\n"
+                f"OBC description:\n{obc_description}"
+            )
```
sagihirshfeld: This seems to override and mostly duplicate the parent's (ObjectBucket) `verify_health`. The only difference I see is that in this implementation omits this part from the original:
```
                elif self.mcg and not self.mcg.s3_verify_bucket_exists(self.name):
```
I suppose `s3_verify_bucket_exists` would fail to list a vector bucket, hence the need.

Let's use a simpler approach instead of duplicating code - add a class attribute to the base class:
```
  # In ObjectBucket (base):
  _verify_s3_listing = True
```
Add it to the original's verify_health check:
```
  # In verify_health, line 316 becomes:
  elif self._verify_s3_listing and self.mcg and not self.mcg.s3_verify_bucket_exists(self.name):
```
Then set in this class:
```
  class VectorOCBucket(OCBucket):
      _verify_s3_listing = False
```

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -0,0 +1,208 @@
...
+from ocs_ci.framework.testlib import MCGTest
+from ocs_ci.ocs import vector_utils
+from ocs_ci.ocs.resources.objectbucket import OBC
+
+logger = logging.getLogger(__name__)
+
+
+@mcg
+@red_squad
+class TestS3VectorWithFactory(MCGTest):
```
sagihirshfeld: The use the factory fixture is not important enough from a testing POV to be in the title:
```suggestion
class TestS3Vectors(MCGTest):
```

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    Test S3 Vector functionality against noobaa
+    """
+
+    @tier2
+    def test_s3_vector_operations_with_factory(self, vector_bucket_factory, mcg_obj):
```
sagihirshfeld: The use the factory fixture is not important enough from a testing POV to be in the title:
```suggestion
    def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
```

tests/functional/object/mcg/test_s3_vector.py
sagihirshfeld: Yuli's right - `mcg_obj.s3vectors_client` was created using the admin creds, and the assignments in lines 63 and 64 don't change that. In the following calls to vector_utils functions we're still using the same client that uses the admin creds.

However, calling `mcg_obj.update_s3_creds()` after wouldn't work because it reassigns the admin creds before recreating `s3vectors_client`. 

Given that the vectors client need creds from an OBC, let's remove the initialization of `s3vectors_client` from the MCG class. The simplest solution would be to create another util function under `vector_utils` that accepts the OBC (or its credentials), an instance of the MCG class and returns an `s3vectors_client`:
```
  def create_s3vectors_client(mcg_obj, obc_obj):
      """
      Create an S3 Vectors client using OBC credentials and the MCG vectors endpoint.

      Args:
          mcg_obj (MCG): MCG object with vectors_endpoint
          obc_obj (OBC): OBC object with access credentials

      Returns:
          boto3.client: S3 Vectors client

      """
      return boto3.client(
          "s3vectors",
          region_name=mcg_obj.region,
          verify=retrieve_verification_mode(),
          endpoint_url=mcg_obj.vectors_endpoint,
          aws_access_key_id=obc_obj.access_key_id,
          aws_secret_access_key=obc_obj.access_key,
      )
```
This will require to change the vector util functions to accept a client and not an obj that has the client. You can avoid this by returning a simple wrapper object instead:
```
  from types import SimpleNamespace

  def create_s3vectors_client(mcg_obj, obc_obj):
      client = boto3.client(
          "s3vectors",
          region_name=mcg_obj.region,
          verify=retrieve_verification_mode(),
          endpoint_url=mcg_obj.vectors_endpoint,
          aws_access_key_id=obc_obj.access_key_id,
          aws_secret_access_key=obc_obj.access_key,
      )
      return SimpleNamespace(s3vectors_client=client)
```

ocs_ci/ocs/resources/mcg.py
```diff
@@ -1266,6 +1297,17 @@ def update_s3_creds(self):
...
+        if self.vectors_endpoint:
+            self.s3vectors_client = boto3.client(
+                "s3vectors",
+                region_name=self.region,
+                verify=retrieve_verification_mode(),
+                endpoint_url=self.vectors_endpoint,
+                aws_access_key_id=self.access_key_id,
+                aws_secret_access_key=self.access_key,
+                config=retry_cfg,
+            )
```
sagihirshfeld: Let's remove this class member from the MCG class - see the comment that suggests creating it elsewhere.

ocs_ci/ocs/resources/mcg.py
```diff
@@ -73,7 +73,10 @@ class MCG:
         noobaa_user,
         noobaa_password,
         data_to_mask,
-    ) = (None,) * 12
+        s3vectors_client,
+        vectors_endpoint,
+        vectors_internal_endpoint,
+    ) = (None,) * 15
```
sagihirshfeld: Let's remove this class member from the MCG class - see the comment that suggests creating it elsewhere.
```suggestion
        vectors_endpoint,
        vectors_internal_endpoint,
    ) = (None,) * 14
```

tests/conftest.py
```diff
@@ -3627,6 +3628,238 @@ def bucket_cleanup():
...
+                    remaining_indices = indices_response.get("indexes", [])
+                    if remaining_indices:
+                        for index in remaining_indices:
+                            try:
+                                delete_index(
+                                    mcg_obj,
+                                    vectorBucketName=bucket.name,
+                                    indexName=index["indexName"],
+                                )
```
sagihirshfeld: `list_indexes` and `delete_index` will fail here because `mcg.s3vectors_client` is still using the admin credentials. See the above comment about creating the `s3vectors_client` using a util func.

Given such a util func, you can fix this using this instead:


```suggestion
                    obc_obj = OBC(bucket.name)
                    s3vectors_client = create_s3vectors_client(mcg_obj, obc_obj)
                    indices_response = s3vectors_client.list_indexes(vectorBucketName=bucket.name)
                    remaining_indices = indices_response.get("indexes", [])
                    if remaining_indices:
                        for index in remaining_indices:
                            try:
                                s3vectors_client.delete_index(
                                    vectorBucketName=bucket.name,
                                    indexName=index["indexName"],
                                )

```

ocs_ci/ocs/resources/bucketclass.py
ypersky1980: Agree.

udaysk23:
@sagihirshfeld @ypersky1980 I have fixed all the review comments in latest code change

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -0,0 +1,211 @@
...
+            s3vectors_client, vectorBucketName=vector_bucket_name, indexName=index_name
+        )
+
+        assert deleted_count >= num_vectors, (
+            f"Not all vectors were deleted. Deleted: {deleted_count}, "
+            f"Expected: {num_vectors}"
+        )
+        logger.info(f"✓ Deleted {deleted_count} vectors")
+
+        # Verify vectors are deleted
```
ypersky1980: @udaysk23  The test manually deletes everything, then the fixture tries to delete the same things again. 
  
  Steps 8-10 explicitly delete all vectors, the index, and the vector bucket. But the vector_bucket_cleanup finalizer in conftest.py:3830 also tries to list indexes, delete them, and delete the bucket. So every successful test run produces a round of "cleanup failed" warnings for resources that are already gone. Either let the fixture handle all cleanup (remove steps 8-10 from the test), or skip cleanup in the fixture for resources the test already deleted. Right now it's noisy and makes real cleanup failures harder to spot in logs.

ypersky1980 (commented):
Left one more comment.

sagihirshfeld (dismissed):
LGTM

tests/functional/object/mcg/test_s3_vector.py
ypersky1980: @udaysk23 what do you think about this? I see that it's not fixed.

udaysk23: @ypersky1980 Step 8 to 10 are needed as part of Happy path validation.
The reason for adding finalizer in fixture is that, if test fails for step 1 to 7, it should do clean up

ypersky1980: OK.

ypersky1980 (approved):
LGTM

sagihirshfeld (approved):
LGTM

PR #15231, opened 2026-05-26
Adds tests for disable RGW http feature

udaysk23:
Adds tests for disable RGW http feature

## Summary by CodeRabbit

* **Tests**
  * Added a functional test validating RGW HTTP enable/disable flows: verifies unsecured and secure route behavior and accessibility changes when HTTP is toggled.
  * Ensures original cluster configuration is restored after the test and waits for RGW pods and routes to reach expected states.
  * Includes automated setup/teardown, helper utilities, and polling for reliable, isolated test runs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/red-hat-storage/ocs-ci/pull/15231?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/functional/object/rgw/test_rgw_http_endpoint.py
+++ b/tests/functional/object/rgw/test_rgw_http_endpoint.py
@@ -1,0 +1,348 @@
+import logging
+import requests
+import pytest
+
+from ocs_ci.framework import config
+from ocs_ci.framework.pytest_customization.marks import (
+    on_prem_platform_required,
+    post_upgrade,
+    red_squad,
+    skipif_disconnected_cluster,
+    skipif_external_mode,
+    skipif_ocs_version,
+    skipif_proxy_cluster,
+    tier2,
+    rgw,
+    runs_on_provider,
+)
+from ocs_ci.ocs import constants, ocp
+from ocs_ci.ocs.exceptions import CommandFailed, TimeoutExpiredError
+from ocs_ci.ocs.resources.pod import wait_for_pods_to_be_running
+from ocs_ci.ocs.utils import get_pod_name_by_pattern
+from ocs_ci.utility.utils import TimeoutSampler
+
+log = logging.getLogger(__name__)
+
+
+@rgw
+@red_squad
+@skipif_disconnected_cluster
+@skipif_proxy_cluster
+@runs_on_provider
+@on_prem_platform_required
+class TestRGWHTTPEndpoint:
+    """
+    Test the RGW HTTP endpoint disable functionality
+
+    """
+
+    @pytest.fixture(autouse=True)
+    def setup_and_teardown(self, request):
+        """
+        Setup and teardown fixture to restore original state after test
+
+        """
+        self.storage_cluster_obj = ocp.OCP(
+            kind=constants.STORAGECLUSTER,
+            namespace=config.ENV_DATA["cluster_namespace"],
+        )
+        self.route_obj = ocp.OCP(
+            kind=constants.ROUTE, namespace=config.ENV_DATA["cluster_namespace"]
+        )
+
+        # Get storage cluster name with fallback
+        self.storage_cluster_name = (
+            self.storage_cluster_obj.resource_name or "ocs-storagecluster"
+        )
+
+        # Get initial storage cluster state
+        self.storage_cluster_data = self.storage_cluster_obj.get(
+            resource_name=self.storage_cluster_name
+        )
+
+        # Check if disableHttp parameter exists
+        ceph_object_stores = (
+            self.storage_cluster_data.get("spec", {})
+            .get("managedResources", {})
+            .get("cephObjectStores", {})
+        )
+        self.disable_http_exists = "disableHttp" in ceph_object_stores
+        self.original_disable_http = ceph_object_stores.get("disableHttp", False)
+
+        def teardown():
+            """Teardown: Restore original state"""
+            if self.disable_http_exists:
+                # Restore original value if it existed
+                log.info(
+                    f"Restoring original disableHttp value: {self.original_disable_http}"
+                )
+                self.set_disable_http_value(
+                    self.storage_cluster_name, self.original_disable_http
+                )
+            else:
+                # Remove the parameter if it didn't exist originally
+                log.info("Removing disableHttp parameter (it didn't exist originally)")
+                params = '[{"op": "remove", "path": "/spec/managedResources/cephObjectStores/disableHttp"}]'
+                try:
+                    self.storage_cluster_obj.patch(
+                        resource_name=self.storage_cluster_name,
+                        params=params,
+                        format_type="json",
+                    )
+                except CommandFailed:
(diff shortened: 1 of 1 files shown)


tests/functional/object/rgw/test_rgw_http_endpoint.py
```diff
@@ -0,0 +1,307 @@
...
+        log.info(f"Waiting for route {route_name} to exist")
+        for sample in TimeoutSampler(
+            timeout=timeout, sleep=10, func=self.get_route, route_name=route_name
+        ):
+            if sample:
+                log.info(f"Route {route_name} exists")
+                return sample
+        raise AssertionError(
+            f"Route {route_name} did not get created within {timeout} seconds"
+        )
```
sagihirshfeld: When the TimeoutSampler expires it raises a TimeoutExpiredError. This means that we never reach the raised assertion at line 164. If you still wish to "clean" the raised error:
```suggestion
          try:
              for sample in TimeoutSampler(
                  timeout=timeout, sleep=10, func=self.get_route, route_name=route_name
              ):
                  if sample:
                      log.info(f"Route {route_name} exists")
                      return sample
          except TimeoutExpiredError:
              raise AssertionError(
                  f"Route {route_name} did not get created within {timeout} seconds"
              )
```

tests/functional/object/rgw/test_rgw_http_endpoint.py
```diff
@@ -0,0 +1,307 @@
...
+)
+from ocs_ci.ocs import constants, ocp
+from ocs_ci.ocs.exceptions import CommandFailed
+from ocs_ci.ocs.resources.pod import wait_for_pods_to_be_running
+from ocs_ci.ocs.utils import get_pod_name_by_pattern
+from ocs_ci.utility.utils import TimeoutSampler
+
+log = logging.getLogger(__name__)
+
+
```
sagihirshfeld, line 26: Would `check_http_accessibility` work normally on disconnected and proxy clusters? Consider adding these: 
```suggestion
@skipif_disconnected_cluster
@skipif_proxy_cluster
```

sagihirshfeld (dismissed):

tests/functional/object/rgw/test_rgw_http_endpoint.py
```diff
@@ -0,0 +1,307 @@
...
+                return sample
+        raise AssertionError(
+            f"Route {route_name} did not get created within {timeout} seconds"
+        )
+
+    @skipif_external_mode
+    @tier2
+    @post_upgrade
+    @skipif_ocs_version(">4.22")
+    def test_rgw_disable_http_endpoint(self):
```
ypersky1980, line 187: Version skip logic appears inverted. 

  @skipif_ocs_version(">4.22")

  This skips the test on versions greater than 4.22, so it only runs on <=4.22. If disableHttp is a feature introduced in 4.22, the skip should be @skipif_ocs_version("<4.22") to skip versions
  that don't have it. As written, this test will stop running on every future release (4.23, 4.24, …), which is almost certainly not the intent. Please verify the direction — if the feature
  persists beyond 4.22, this should be "<4.22".

udaysk23: @ypersky1980 , The upgrade path should test only from 4.21 to 4.22.

tests/functional/object/rgw/test_rgw_http_endpoint.py
udaysk23: Fixed

sagihirshfeld (dismissed):

tests/functional/object/rgw/test_rgw_http_endpoint.py
```diff
@@ -0,0 +1,326 @@
...
+        self.wait_for_route_to_exist(
+            constants.RGW_ROUTE_INTERNAL_MODE_SECURE, timeout=180
+        )
+        # Step 3: Change disableHttp value to true
+        log.info("Step 3: Changing disableHttp value to true")
+        self.set_disable_http_value(self.storage_cluster_name, True)
+
+        # Wait for RGW pods to be ready
+        self.wait_for_rgw_pods_ready()
+
```
ypersky1980, line 240: @udaysk23 Accepting HTTP 500 as "endpoint is inaccessible" is wrong. 
  
  The test treats status codes 503, 404, 502, and 500 all as proof that HTTP is disabled. But if you get any HTTP response — even a 500 — it means something is still listening on the HTTP port.
   That contradicts "disable HTTP." A 500 could mean RGW is running on HTTP but has an internal error. The only reliable proof that HTTP is truly disabled is a connection error (the except
  branch). The test should fail on any successful HTTP response, regardless of status code.

tests/functional/object/rgw/test_rgw_http_endpoint.py
```diff
@@ -0,0 +1,326 @@
...
+        assert (
+            http_host
+        ), "HTTP route host is missing; cannot validate endpoint accessibility"
+        http_endpoint_url = f"http://{http_host}"
+        log.info(f"Saved HTTP endpoint URL: {http_endpoint_url}")
+
+        # Verify HTTPS route exists
+        self.wait_for_route_to_exist(
+            constants.RGW_ROUTE_INTERNAL_MODE_SECURE, timeout=180
+        )
```
ypersky1980, line 233: After disabling HTTP, the test only checks that the HTTPS route exists — not that it actually works
  
After setting disableHttp=true, the test verifies the HTTP route is deleted and the HTTPS route still exists (wait_for_route_to_exist). But it never makes an actual HTTPS request
  to confirm the secure endpoint is still functional. If a bug in the operator accidentally breaks both HTTP and HTTPS when toggling the flag, this test wouldn't catch it. The test should curl
  the HTTPS endpoint and assert a successful response, similar to what step 6 does for HTTP after re-enabling.

ypersky1980 (commented):
Left comments.

tests/functional/object/rgw/test_rgw_http_endpoint.py
udaysk23: @ypersky1980 This is expected. The Router receives the request on port 80 but can't find a matching Route for that hostname (since it was deleted), so it returns Service Unavailable.

tests/functional/object/rgw/test_rgw_http_endpoint.py
udaysk23: Fixed

sagihirshfeld (approved):

tests/functional/object/rgw/test_rgw_http_endpoint.py
ypersky1980: OK.

tests/functional/object/rgw/test_rgw_http_endpoint.py
ypersky1980: ok.

ypersky1980 (approved):
LGTM

PR #15251, opened 2026-05-29
Adds tests for qualifying RGW STS feature

udaysk23:
Adds tests for qualifying RGW STS feature

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end functional test for RGW STS that verifies StorageCluster STS enablement and presence of the STS secret, validates object store STS configuration, exercises IAM user and role provisioning, performs assume-role to obtain temporary credentials, and confirms S3 operations (list, upload, download) using those temporary credentials, including best-effort cleanup of created identities.

### DIFF
--- a/tests/functional/object/rgw/test_rgw_sts.py
+++ b/tests/functional/object/rgw/test_rgw_sts.py
@@ -1,0 +1,526 @@
+"""
+Test RGW STS (Security Token Service) functionality in OCS-CI
+"""
+
+import json
+import logging
+import uuid
+
+import boto3
+import pytest
+from botocore.exceptions import ClientError
+
+from ocs_ci.framework import config
+from ocs_ci.framework.pytest_customization.marks import (
+    post_upgrade,
+    red_squad,
+    rgw,
+    runs_on_provider,
+    skipif_mcg_only,
+    tier2,
+)
+from ocs_ci.helpers.helpers import create_unique_resource_name
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.ocp import OCP
+from ocs_ci.ocs.resources.pod import get_ceph_tools_pod
+
+logger = logging.getLogger(__name__)
+
+
+@rgw
+@red_squad
+@runs_on_provider
+@skipif_mcg_only
+class TestRGWSTS:
+    """
+    Test RGW STS (Security Token Service) functionality
+    """
+
+    @pytest.fixture(scope="function")
+    def rgw_iam_client_creation(self, rgw_obj, request):
+        """
+        Create RGW account, root user, and IAM client for STS operations
+
+        Steps:
+        1. Create RGW account using radosgw-admin (20-byte ID: RGW + 17 numeric digits)
+        2. Create account root user using radosgw-admin
+        3. Create boto3 IAM client with user credentials
+
+        Cleanup:
+        - Deletes the user and account after test completion
+
+        Returns:
+            tuple: (iam_client, account_info_dict)
+                - iam_client: boto3 IAM client configured with account credentials
+                - account_info_dict: Dictionary containing:
+                    - account_id: RGW account ID
+                    - user_id: User ID
+                    - access_key: AWS access key
+                    - secret_key: AWS secret key
+        """
+        cephobjectstore_name = OCP(
+            kind=constants.CEPHOBJECTSTORE,
+            resource_name="ocs-storagecluster-cephobjectstore",
+            namespace=config.ENV_DATA["cluster_namespace"],
+        ).get()["metadata"]["name"]
+        toolbox = get_ceph_tools_pod()
+
+        # Generate unique IDs to avoid conflicts
+        # RGW account ID must be exactly 20 bytes long, start with "RGW", and end with numeric digits
+        # Format: RGW + 17 numeric digits = 20 bytes total
+        numeric_suffix = str(uuid.uuid4().int)[:17]
+        account_id = f"RGW{numeric_suffix}"
+        user_id = create_unique_resource_name("sts", "user")
+
+        def finalizer():
+            """
+            Cleanup RGW account and user after test completion
+            """
+            logger.info("Starting cleanup of RGW account and user")
+            toolbox = get_ceph_tools_pod()
+
+            try:
+                # Delete user first (must delete user before account)
+                logger.info(f"Deleting RGW user: {user_id}")
+                user_delete_cmd = f"radosgw-admin user rm --uid={user_id} --purge-data"
+                toolbox.exec_cmd_on_pod(user_delete_cmd, out_yaml_format=False)
+                logger.info(f"RGW user {user_id} deleted successfully")
+            except Exception as e:
+                logger.warning(f"Failed to delete RGW user {user_id}: {e}")
+
+            try:
+                # Delete account
+                logger.info(f"Deleting RGW account: {account_id}")
+                account_delete_cmd = (
+                    f"radosgw-admin account rm --account-id={account_id} --purge-data"
+                )
+                toolbox.exec_cmd_on_pod(account_delete_cmd, out_yaml_format=False)
+                logger.info(f"RGW account {account_id} deleted successfully")
+            except Exception as e:
+                logger.warning(f"Failed to delete RGW account {account_id}: {e}")
+
+        request.addfinalizer(finalizer)
+
+        # Step 1: Create RGW account
+        logger.info(f"Creating RGW account with ID: {account_id}")
+        account_create_cmd = (
+            f"radosgw-admin account create "
+            f"--account-id={account_id} "
+            f"--rgw-zone={cephobjectstore_name} "
+            f"--rgw-zonegroup={cephobjectstore_name} "
+            f"--rgw-realm={cephobjectstore_name}"
+        )
+        toolbox.exec_cmd_on_pod(account_create_cmd, out_yaml_format=False)
+        logger.info(f"RGW account {account_id} created successfully")
+
+        # Step 2: Create account root user
+        logger.info(f"Creating account root user: {user_id}")
+        user_create_cmd = (
+            f"radosgw-admin user create "
+            f"--uid={user_id} "
+            f'--display-name="STSUser" '
+            f"--account-id={account_id} "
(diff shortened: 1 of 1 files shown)


tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -0,0 +1,534 @@
...
+                logger.info(f"Deleting RGW account: {account_id}")
+                account_delete_cmd = (
+                    f"radosgw-admin account rm --account-id={account_id}"
+                )
+                toolbox.exec_cmd_on_pod(account_delete_cmd, out_yaml_format=False)
+                logger.info(f"RGW account {account_id} deleted successfully")
+            except Exception as e:
+                logger.warning(f"Failed to delete RGW account {account_id}: {e}")
+
+        request.addfinalizer(finalizer)
```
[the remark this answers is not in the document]
udaysk23: fixed

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -0,0 +1,534 @@
...
+            assert assumed_secret_key, "Assumed role secret key should not be empty"
+            assert assumed_session_token, "Session token should not be empty"
+
+        except ClientError as e:
+            logger.error(f"Failed to assume role: {e}")
+            raise
+
+        # Step 9: Access bucket using assumed role credentials
+        logger.info("Step 9: Accessing S3 bucket using assumed role credentials")
+
```
ypersky1980, line 482: No assertion on temporary credential expiration: 

  assert assumed_access_key, "Assumed role access key should not be empty"
  assert assumed_secret_key, "Assumed role secret key should not be empty"
  assert assumed_session_token, "Session token should not be empty"

  The test asserts the three credential fields are non-empty but never checks credentials["Expiration"]. The core contract of STS AssumeRole is that it returns temporary credentials with a
  bounded lifetime. The test should assert that Expiration exists and is a future timestamp. Without this, a bug where RGW returns permanent credentials instead of temporary ones would pass
  silently.

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -0,0 +1,534 @@
...
+            # Create bucket
+            logger.info(f"Creating S3 bucket: {bucket_name}")
+            iam_user_s3_client.create_bucket(Bucket=bucket_name)
+            logger.info(f"Bucket {bucket_name} created successfully")
+
+        except ClientError as e:
+            logger.error(f"Failed to create IAM user or bucket: {e}")
+            raise
+
```
ypersky1980, line 389: Resources created in the test body have no cleanup

  The test creates an IAM user, access key, S3 bucket, object, IAM role, and inline policy — but none of these are cleaned up. The fixture finalizer only deletes the RGW root user and account.
  If account rm doesn't cascade (which it won't without --purge-data), these resources leak on the cluster and can cause conflicts in subsequent runs. A try/finally block or additional
  finalizer calls are needed for the resources created inside the test method.

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -0,0 +1,534 @@
...
+            "Statement": [
+                {
+                    "Effect": "Allow",
+                    "Principal": {"AWS": "*"},
+                    "Action": "sts:AssumeRole",
+                }
+            ],
+        }
+        return policy
+
```
ypersky1980, line 175: Trust policy allows anyone to assume the role. 

  The trust policy sets "Principal": {"AWS": "*"}, which means any authenticated user can assume this role. Even in a test, the principal should be restricted to the specific user or account
  that will call AssumeRole. As written, the test doesn't actually verify that RGW enforces who is allowed to assume a role — it just grants access to everyone. This also sets a bad example if
  someone copies this pattern into production code.

More detailed explanation: 

  STS lets a user "assume a role" to get temporary permissions. The trust policy controls who is allowed to assume that role — it's the security gate.

  When the policy says "Principal": {"AWS": "*"}, it means "allow literally anyone". So the test proves that STS works mechanically (you call AssumeRole, you get credentials), but it doesn't
  prove that RGW actually blocks unauthorized users from assuming the role.

  Imagine a bug where RGW ignores the trust policy entirely and lets every user assume every role. This test would still pass, because it never tries to have an unauthorized user assume the
  role and checks that it fails.

  A better test would:
  1. Set the principal to a specific user (e.g., arn:aws:iam::{account_id}:user/{iam_user_name})
  2. Verify that this user can assume the role
  3. Optionally verify that a different user gets denied

ypersky1980 (commented):
Left comments.

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -0,0 +1,534 @@
...
+    """
+    Test RGW STS (Security Token Service) functionality
+
+    This test validates the complete STS workflow including:
+    - StorageCluster STS configuration
+    - Ceph RGW STS settings
+    - Secret creation for STS key
+    - CephObjectStore configuration
+    - IAM user and role creation
+    - AssumeRole functionality
```
sagihirshfeld: This only applies to `test_rgw_sts_configuration`, and not the entire `TestRGWSTS` class. Please either remove this or combine it with the test's docstring.

tests/functional/object/rgw/test_rgw_sts.py
sagihirshfeld: It's fine for this test - the accuracy of tighter policies should be tested in a separate test case.

That said, I think we should extend this method so it can accept one or more users to target (if none are given it will target * like it currently does).

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -0,0 +1,534 @@
...
+                {
+                    "Effect": "Allow",
+                    "Principal": {"AWS": "*"},
+                    "Action": "sts:AssumeRole",
+                }
+            ],
+        }
+        return policy
+
+    def role_permission_policy(self):
```
sagihirshfeld, line 176: We should add an arg so you can at least set one or more specific resources (for future tests)

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -0,0 +1,534 @@
...
+    - StorageCluster STS configuration
+    - Ceph RGW STS settings
+    - Secret creation for STS key
+    - CephObjectStore configuration
+    - IAM user and role creation
+    - AssumeRole functionality
+    """
+
+    @pytest.fixture(scope="function")
+    def rgw_iam_client_creation(self, rgw_obj, request):
```
sagihirshfeld, line 40: Please move this fixture to tests/conftest.py for reusability

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -0,0 +1,534 @@
...
+
+            # Create S3 client with IAM user credentials
+            iam_user_s3_client = boto3.client(
+                "s3",
+                endpoint_url=endpoint,
+                aws_access_key_id=iam_user_access_key,
+                aws_secret_access_key=iam_user_secret_key,
+                region_name=rgw_obj.region,
+                verify=False,
+            )
```
sagihirshfeld, line 379: I recommend wrapping this in a fixture (i.e `create rgw_iam_user`) that builds on top of `rgw_iam_client_creation` to simplify future testing

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -0,0 +1,534 @@
...
+            logger.info(f"Session token length: {len(assumed_session_token)}")
+
+            assert assumed_access_key, "Assumed role access key should not be empty"
+            assert assumed_secret_key, "Assumed role secret key should not be empty"
+            assert assumed_session_token, "Session token should not be empty"
+
+        except ClientError as e:
+            logger.error(f"Failed to assume role: {e}")
+            raise
```
sagihirshfeld, line 478: Also consider wrapping these steps in a follow-up fixture that builds on the previous two: create_rgw_sts_client. It should be a factory fixture that accepts an optional list of resources to allow access to.

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: @ypersky1980  We have separate test for it

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: @sagihirshfeld it will be covered in upcoming test script automation

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: @ypersky1980 for now, I would like to keep it here since STS feature is tested with this script only. Will move out of this test if we have any other STS dependent feature

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: IAM user is created using main user and account. While cleanup of user and account, it will delete all kind of data and buckets from cluster

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: @sagihirshfeld This is workflow validation test and each step is build on previous step. For now I would like to skip it and will keep as linear test steps for clarity

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: @sagihirshfeld This is workflow validation test and each step is build on previous step. For now I would like to skip it and will keep as linear test steps for clarity

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: Fixed

sagihirshfeld (approved):

tests/functional/object/rgw/test_rgw_sts.py
ypersky1980: ok.

ypersky1980 (approved):
LGTM

PR #15297, opened 2026-06-04
Adds high priority tests for s3 vector feature

udaysk23:
Adds high priority tests for s3 vector feature

## Summary by CodeRabbit

* **New Features**
  * S3 Vectors client can optionally scope requests to a specific namespace resource; scoped requests include the header required to target that namespace.

* **Tests**
  * New tests validate vector bucket naming rules (patterns, length boundaries, duplicates, and client vs. server validation).
  * New tests validate vector index dimensions (valid/invalid boundaries, deletion behavior, duplicate-index handling).
  * New tests exercise multi-dimensional index behavior, insertion, overwrite, and cleanup.

### DIFF
--- a/ocs_ci/ocs/vector_utils.py
+++ b/ocs_ci/ocs/vector_utils.py
@@ -13,13 +13,14 @@ from botocore.config import Config
 logger = logging.getLogger(__name__)
 
 
-def create_s3vectors_client(mcg_obj, obc_obj):
+def create_s3vectors_client(mcg_obj, obc_obj, nsr_name=None):
     """
     Create an S3 Vectors client using OBC credentials
 
     Args:
         mcg_obj (obj): MCG object containing vectors endpoint and region
         obc_obj (obj): OBC object containing access credentials
+        nsr_name (str): Optional NooBaa namespace resource name for x-noobaa-nsr header
 
     Returns:
         boto3.client: S3 Vectors client configured with OBC credentials
@@ -30,7 +31,7 @@ def create_s3vectors_client(mcg_obj, obc_obj):
 
     retry_cfg = Config(retries={"max_attempts": 10, "mode": "standard"})
 
-    return boto3.client(
+    client = boto3.client(
         "s3vectors",
         region_name=mcg_obj.region,
         verify=retrieve_verification_mode(),
@@ -39,6 +40,26 @@ def create_s3vectors_client(mcg_obj, obc_obj):
         aws_secret_access_key=obc_obj.access_key,
         config=retry_cfg,
     )
+
+    # Add custom NooBaa header if NSR name is provided
+    if nsr_name:
+
+        def add_nsr_header(params, **kwargs):
+            params["headers"]["x-noobaa-nsr"] = nsr_name
+
+        client = boto3.client(
+            "s3vectors",
+            region_name=mcg_obj.region,
+            verify=retrieve_verification_mode(),
+            endpoint_url=mcg_obj.vectors_endpoint,
+            aws_access_key_id=mcg_obj.access_key_id,
+            aws_secret_access_key=mcg_obj.access_key,
+            config=retry_cfg,
+        )
+
+        client.meta.events.register("before-call", add_nsr_header)
+
+    return client
 
 
 def generate_test_vectors_with_metadata(
--- a/tests/functional/object/mcg/test_s3_vector.py
+++ b/tests/functional/object/mcg/test_s3_vector.py
@@ -1,19 +1,24 @@
 import logging
+import math
 import random
 import string
 
-from botocore.exceptions import ClientError
+from botocore.exceptions import ClientError, ParamValidationError
 
 from ocs_ci.framework.pytest_customization.marks import (
     tier2,
     red_squad,
     post_upgrade,
     mcg,
+    jira,
 )
 from ocs_ci.framework.testlib import MCGTest
+from ocs_ci.framework import config
 from ocs_ci.ocs import vector_utils
 from ocs_ci.ocs.resources.objectbucket import OBC
+from ocs_ci.ocs.ocp import OCP
 from ocs_ci.utility.utils import TimeoutSampler
+
 
 logger = logging.getLogger(__name__)
 
@@ -24,6 +29,46 @@ class TestS3Vector(MCGTest):
     """
     Test S3 Vector functionality against noobaa
     """
+
+    def _create_index(
+        self,
+        s3vectors_client,
+        index_name,
+        dimension,
+        vector_bucket_name,
+        data_type="float32",
+        distance_metric="cosine",
+    ):
+        """
+        Create a vector index, tolerating the known NooBaa VECTOR_INDEX_ALREADY_OWNED_BY_YOU
+        error that occurs when the index is created successfully but the response erroneously
+        indicates a duplicate.
+
+        Args:
+            s3vectors_client: boto3 s3vectors client
+            index_name (str): Name of the index to create
+            dimension (int): Dimensionality of the index
+            vector_bucket_name (str): Name of the vector bucket
+            data_type (str): Vector data type (default: float32)
+            distance_metric (str): Distance metric to use (default: cosine)
+        """
+        try:
+            vector_utils.create_index(
+                s3vectors_client,
+                index_name=index_name,
+                data_type=data_type,
+                dimension=dimension,
+                distance_metric=distance_metric,
+                vectorBucketName=vector_bucket_name,
+            )
+            logger.info(f"✓ Created {dimension}-dimensional index: {index_name}")
+        except ClientError as e:
+            if "VECTOR_INDEX_ALREADY_OWNED_BY_YOU" in str(e):
+                logger.warning(
+                    f"Index {index_name} already exists (NooBaa bug), continuing..."
+                )
+            else:
+                raise
 
     @tier2
     @post_upgrade
@@ -76,26 +121,7 @@ class TestS3Vector(MCGTest):
 
         # Step 3: Create 3-dimensional vector index
         logger.info(f"Creating {dimension}-dimensional vector index")
-        try:
-            vector_utils.create_index(
-                s3vectors_client,
-                index_name=index_name,
-                data_type="float32",
-                dimension=dimension,
-                distance_metric="cosine",
-                vectorBucketName=vector_bucket_name,
-            )
-            logger.info(f"✓ Vector index '{index_name}' created")
-        except ClientError as e:
-            error_message = str(e)
-            if "VECTOR_INDEX_ALREADY_OWNED_BY_YOU" in error_message:
-                logger.warning(
-                    f"Index '{index_name}' creation got ALREADY_OWNED error. "
-                    "This is a known NooBaa issue where the index is created successfully "
-                    "but an error is returned, causing retries to fail. Continuing test..."
-                )
-            else:
-                raise
+        self._create_index(s3vectors_client, index_name, dimension, vector_bucket_name)
 
         # Step 4: List indices and validate
         logger.info("Listing vector indices")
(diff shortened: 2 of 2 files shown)


ocs_ci/ocs/vector_utils.py
```diff
@@ -40,6 +41,26 @@ def create_s3vectors_client(mcg_obj, obc_obj):
...
+            params["headers"]["x-noobaa-nsr"] = nsr_name
+
+        client = boto3.client(
+            "s3vectors",
+            region_name=mcg_obj.region,
+            verify=retrieve_verification_mode(),
+            endpoint_url=mcg_obj.vectors_endpoint,
+            aws_access_key_id=mcg_obj.access_key_id,
+            aws_secret_access_key=mcg_obj.access_key,
+            config=retry_cfg,
```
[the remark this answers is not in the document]
udaysk23: This is needed as direct s3 API call needs admin creds

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +236,401 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+        # Step 1: Create vector bucket
+        logger.info("Creating vector bucket for index dimension tests")
+        vector_bucket_name = vector_bucket_factory(amount=1)[0].name
+        logger.info(f"✓ Vector bucket created: {vector_bucket_name}")
+
+        obc_obj = OBC(vector_bucket_name)
+        s3vectors_client = vector_utils.create_s3vectors_client(mcg_obj, obc_obj)
+
+        valid_index_names = []
+
```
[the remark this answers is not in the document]
udaysk23: Code is already handling clean up task in finally block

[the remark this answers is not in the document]
udaysk23: vector_bucket_factory will take care of clean up

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -25,6 +29,46 @@ class TestS3Vector(MCGTest):
...
+        index_name,
+        dimension,
+        vector_bucket_name,
+        data_type="float32",
+        distance_metric="cosine",
+    ):
+        """
+        Create a vector index, tolerating the known NooBaa VECTOR_INDEX_ALREADY_OWNED_BY_YOU
+        error that occurs when the index is created successfully but the response erroneously
+        indicates a duplicate.
```
sagihirshfeld, line 45: Do we have an open bug for this? Please mention it (i.e DFBUGS-XXXX) in the docstring if we do for reference

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +258,981 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+            try:
+                vector_utils.create_vector_bucket(s3vectors_client, duplicate_name)
+                assert (
+                    False
+                ), f"Duplicate bucket creation should have failed for {duplicate_name}"
+            except ClientError as e:
+                error_code = e.response.get("Error", {}).get("Code", "")
+                assert error_code in [
+                    "BUCKET_ALREADY_OWNED_BY_YOU",
+                    "InternalFailure",
```
sagihirshfeld, line 371: Getting "InternalFailure" here seems like a (minor) product bug. Do we have an open bug for this? If we do, please mention it in a comment
```suggestion
                    "InternalFailure", # DFBUGS-XXXXX
```

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +258,981 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+                assert (
+                    len(empty_results) == 0
+                ), f"Expected no results from empty {dim_label} index, got {len(empty_results)}"
+                logger.info(
+                    f"✓ [{dim_label}] Empty index query returned 0 results as expected"
+                )
+            except ClientError as e:
+                # NooBaa raises InternalFailure when querying an empty index because
+                # the vector search engine is not initialized until the first vector is inserted.
+                # Treat this as an expected "no results" outcome.
```
sagihirshfeld: Sounds like a (minor) product bug - ideally we'd be told this in the error message instead of an obscure InternalFailure.  If it wasn't raised yet and dismissed by the feature owener, please raise a bug and mention it here.

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +258,981 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+                query_vector=query_vector,
+                top_k=top_k,
+                vectorBucketName=vector_bucket_name,
+                indexName=index_name,
+                returnDistance=True,
+                returnMetadata=True,
+            ).get("vectors", [])
+            assert (
+                0 < len(results_step5) <= top_k
+            ), f"[{dim_label}] Expected 1–{top_k} results, got {len(results_step5)}"
```
sagihirshfeld: Shouldn't this query return exactly `top_k`?

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +258,981 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+            )
+            vector_utils.put_vectors(
+                s3vectors_client,
+                vectors,
+                vectorBucketName=vector_bucket_name,
+                indexName=index_name,
+            )
+            logger.info(f"✓ [{dim_label}] Inserted {num_vectors} vectors")
+
+            # Step 5: Query for top 5 using cosine metric
```
sagihirshfeld, line 940: Looks like we're only verifying that we're getting a certain number of vectors as output and displaying their distance from the query_vector, but we're not verifying that they're indeed the closest out of the 100 that were uploaded.

Would it be possible to add this verification to this step? Please correct me if you did add such a verification somewhere else in this test suite and I missed it (it's a pretty big PR after all 🙂)

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +258,981 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+                0 < len(results_step5) <= top_k
+            ), f"[{dim_label}] Expected 1–{top_k} results, got {len(results_step5)}"
+            logger.info(f"✓ [{dim_label}] Query returned {len(results_step5)} results")
+            for rank, result in enumerate(results_step5, 1):
+                logger.info(
+                    f"  [{dim_label}] Rank {rank}: key={result.get('key')}, "
+                    f"distance={result.get('distance', 'N/A')}"
+                )
+
+            # Step 6: Update one entry and re-run the same query
```
sagihirshfeld: After the update, are we expecting that vector to still be in the top 5 closest vectors? Would it be possible to check if that update affected the result of the query? It might put it farther, resulting in a new vector in the top 5 instead of it.

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +258,981 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+                s3vectors_client,
+                query_vector=query_vector,
+                top_k=top_k,
+                vectorBucketName=vector_bucket_name,
+                indexName=index_name,
+                returnDistance=True,
+                returnMetadata=True,
+            ).get("vectors", [])
+            assert (
+                0 < len(results_step6) <= top_k
```
sagihirshfeld: Shouldn't it be exactly top_k?

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +258,981 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+            logger.info(
+                f"✓ [{dim_label}] Re-query after update returned {len(results_step6)} results"
+            )
+            for rank, result in enumerate(results_step6, 1):
+                logger.info(
+                    f"  [{dim_label}] Rank {rank}: key={result.get('key')}, "
+                    f"distance={result.get('distance', 'N/A')}"
+                )
+
+            # Step 7: Query with zero vector
```
sagihirshfeld: Why is this a special case, and shouldn't we verify the result vectors in some way?

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +258,981 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+                s3vectors_client,
+                query_vector=query_vector,
+                top_k=top_k,
+                vectorBucketName=vector_bucket_name,
+                indexName=index_name,
+                returnDistance=True,
+                returnMetadata=True,
+            ).get("vectors", [])
+            assert (
+                0 < len(results[metric]) <= top_k
```
sagihirshfeld: Shouldn't it be exactly top_k?

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +258,981 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+        logger.info(f"  Cosine    top keys:      {cosine_keys}")
+        logger.info(f"  Euclidean top keys:      {euclidean_keys}")
+        logger.info(f"  Cosine    distances:     {cosine_distances}")
+        logger.info(f"  Euclidean distances:     {euclidean_distances}")
+
+        # Cosine distance is in [0, 2]; euclidean is unbounded — values must differ
+        assert (
+            cosine_distances == euclidean_distances
+        ), "Cosine and euclidean distances should be identical for the same vectors"
+        logger.info("✓ Distance values differ between cosine and euclidean as expected")
```
sagihirshfeld: Might be a misunderstanding on my part, but isn't the comment in 1143 contradicting the assertion that comes after it?

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +258,981 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+                        s3vectors_client,
+                        [edge_vector],
+                        vectorBucketName=vector_bucket_name,
+                        indexName=index_name,
+                    )
+                    logger.info(f"✓ Inserted {description} into {metric} index")
+                except ClientError as e:
+                    logger.info(
+                        f"  {metric} index rejected {description}: {e.response.get('Error', {}).get('Code', 'Unknown')}"
+                    )
```
sagihirshfeld: Looks like we're expecting to fail the test in case of a ClientError here - currently we're just logging it and continuing
```suggestion
                except ClientError as e:
                    logger.error(
                        f"  {metric} index rejected {description}: {e.response.get('Error', {}).get('Code', 'Unknown')}"
                    )
                    raise e
```

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +258,981 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+                        returnDistance=True,
+                    ).get("vectors", [])
+                    logger.info(
+                        f"  [{metric}] {description} query: {len(edge_results)} results returned"
+                    )
+                except ClientError as e:
+                    logger.info(
+                        f"  [{metric}] {description} query raised "
+                        f"{e.response.get('Error', {}).get('Code', 'Unknown')} (noted)"
+                    )
```
sagihirshfeld: Same problem here - we're swallowing the error instead of raising it and failing the test

tests/functional/object/mcg/test_s3_vector.py
udaysk23: @sagihirshfeld This is not a bug, Noobaa is taking some time before responding and in that timeframe, boto3 is sending same new request and that time noobaa is sending `VECTOR_INDEX_ALREADY_OWNED_BY_YOU` exception since first request created it

tests/functional/object/mcg/test_s3_vector.py
udaysk23: NA

tests/functional/object/mcg/test_s3_vector.py
udaysk23: It throws InternalFailure along with error message

sagihirshfeld: What's the error message? Usually `InternalFailure ` errors are pretty obscure

tests/functional/object/mcg/test_s3_vector.py
udaysk23: Fixed

tests/functional/object/mcg/test_s3_vector.py
udaysk23: Fixed

tests/functional/object/mcg/test_s3_vector.py
udaysk23: Fixed

tests/functional/object/mcg/test_s3_vector.py
udaysk23: Added comment in code

tests/functional/object/mcg/test_s3_vector.py
udaysk23: Fixed

tests/functional/object/mcg/test_s3_vector.py
udaysk23: Fixed

tests/functional/object/mcg/test_s3_vector.py
udaysk23: invalid metric rejection is correct — both are raised by boto3 depending on client-side vs server-side validation

tests/functional/object/mcg/test_s3_vector.py
udaysk23: invalid metric rejection is correct — both are raised by boto3 depending on client-side vs server-side validation

tests/functional/object/mcg/test_s3_vector.py
udaysk23: Fixed

tests/functional/object/mcg/test_s3_vector.py
sagihirshfeld: I'm not sure I'm following. Are we expecting `put_vectors` to fail for all of the edge cases? Some of them? If we do this should be clearly documented and logged as part of the test, and if we're not then the test should fail clearly when `put_vectors` fails with a ClientError instead of ignoring the exception.

tests/functional/object/mcg/test_s3_vector.py
sagihirshfeld: Still unclear. Are we expecting some or all of the `vector_utils.query_vectors` calls to fail here or not? If we are, then we should document it clearly. If we're not, we should raise the exception so the test fails for us to notice the problem.

ypersky1980:
I agree with Sagi's comments.

tests/functional/object/mcg/test_s3_vector.py
udaysk23: ```An error occurred (InternalFailure) when calling the QueryVectors operation (reached max retries: 4): Cannot read properties of undefined (reading 'vectorSearch')```

tests/functional/object/mcg/test_s3_vector.py
udaysk23: Fixed

tests/functional/object/mcg/test_s3_vector.py
udaysk23: fixed

tests/functional/object/mcg/test_s3_vector.py
udaysk23: Added validation for "vectorSearch" string as well

sagihirshfeld (dismissed):
LGTM

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -233,3 +259,1270 @@ def test_s3_vector_operations(self, vector_bucket_factory, mcg_obj):
...
+                assert False, f"Index with invalid metric '{metric}' should have failed"
+            except (ClientError, ParamValidationError, ValueError) as e:
+                logger.info(
+                    f"✓ Distance metric '{metric}' correctly rejected: {type(e).__name__}"
+                )
+
+        logger.info("✅ Vector distance metrics test completed successfully!")
+
+    @tier2
+    def test_vector_bucket_policy_operations(self, vector_bucket_factory, mcg_obj):
```
ypersky1980: test_vector_bucket_policy_operations only tests CRUD, never verifies policy enforcement.
  
  The test sets Deny rules on DeleteVectors and GetVectors, but never actually attempts those operations to verify they're denied. It only checks that the policy JSON is stored and retrieved correctly. 

A policy feature could be "stored but never enforced" and this test would still pass. After Step 5 (verify policy contents), I suggest that you add a step that creates an index + vector, then tries the denied action (e.g., delete_vectors) and asserts it fails with AccessDenied.

ypersky1980 (commented):
Left a comment.

ypersky1980 (dismissed):
LGTM

tests/functional/object/mcg/test_s3_vector.py
udaysk23: @ypersky1980 Thanks for pointing out. This test code should not suppose to be here since it is still WIP and mistakenly pushed.
Removing it in latest commit.

sagihirshfeld (approved):

ypersky1980 (approved):
LGTM

PR #15312, opened 2026-06-08
Adds negative tests for RGW disablehttp feature

udaysk23:
Adds negative tests for RGW disablehttp feature

## Summary by CodeRabbit

* **Tests**
  * Enhanced coverage for RGW HTTP endpoint edge cases when the HTTP-disable flag is unset or removed.
  * Applied external-mode skipping at the test-class level for consistent behavior across environments.
  * Refactored accessibility checks to use a retry-friendly helper for more reliable verification of route removal and recovery.
  * Added parameterized scenarios asserting HTTP is re-enabled while HTTPS remains functional.

### DIFF
--- a/tests/functional/object/rgw/test_rgw_http_endpoint.py
+++ b/tests/functional/object/rgw/test_rgw_http_endpoint.py
@@ -27,6 +27,7 @@ log = logging.getLogger(__name__)
 @rgw
 @red_squad
 @skipif_disconnected_cluster
+@skipif_external_mode
 @skipif_proxy_cluster
 @runs_on_provider
 @on_prem_platform_required
@@ -152,6 +153,52 @@ class TestRGWHTTPEndpoint:
         except CommandFailed:
             return None
 
+    def check_https_accessibility(self, https_endpoint_url):
+        """
+        Check if HTTPS endpoint returns a functional response
+
+        Args:
+            https_endpoint_url (str): Full HTTPS URL to check
+
+        Returns:
+            bool: True if endpoint returns a functional status, False otherwise
+
+        """
+        try:
+            response = requests.get(https_endpoint_url, timeout=10, verify=False)
+            if response.status_code in [200, 403, 401, 404]:
+                log.info(
+                    f"HTTPS endpoint accessible with status {response.status_code}"
+                )
+                return True
+            log.info(f"Got status {response.status_code}, retrying...")
+            return False
+        except requests.exceptions.RequestException as e:
+            log.info(f"Connection failed: {e}, retrying...")
+            return False
+
+    def check_http_accessibility(self, http_endpoint_url):
+        """
+        Check if HTTP endpoint returns a 200 response
+
+        Args:
+            http_endpoint_url (str): Full HTTP URL to check
+
+        Returns:
+            bool: True if endpoint returns 200, False otherwise
+
+        """
+        try:
+            response = requests.get(http_endpoint_url, timeout=10, verify=False)
+            if response.status_code == 200:
+                log.info(f"HTTP endpoint accessible with status {response.status_code}")
+                return True
+            log.info(f"Got status {response.status_code}, retrying...")
+            return False
+        except requests.exceptions.RequestException as e:
+            log.info(f"Connection failed: {e}, retrying...")
+            return False
+
     def wait_for_route_to_exist(self, route_name, timeout=180):
         """
         Wait for a route to exist
@@ -180,7 +227,6 @@ class TestRGWHTTPEndpoint:
                 f"Route {route_name} did not get created within {timeout} seconds"
             )
 
-    @skipif_external_mode
     @tier2
     @post_upgrade
     @skipif_ocs_version(">4.22")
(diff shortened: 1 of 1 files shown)


tests/functional/object/rgw/test_rgw_http_endpoint.py
```diff
@@ -318,31 +340,127 @@ def test_rgw_disable_http_endpoint(self):
...
+            for sample in TimeoutSampler(
+                timeout=300,
+                sleep=10,
+                func=self.check_http_accessibility,
+                http_endpoint_url=http_endpoint_url,
+            ):
+                if sample:
+                    log.info(f"HTTP endpoint accessible after patch_action={patch_action}")
+                    break
+
```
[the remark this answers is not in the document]
udaysk23: patch_action == "null" is expected to be rejected

tests/functional/object/rgw/test_rgw_http_endpoint.py
```diff
@@ -318,31 +340,144 @@ def test_rgw_disable_http_endpoint(self):
...
+        log.info("HTTP route deleted as expected after setting disableHttp=true")
+
+        # Step 2: Apply null or remove patch — these may not trigger reconciliation
+        if patch_action == "null":
+            log.info("Step 2: Setting disableHttp to null via merge patch")
+            params = '{"spec":{"managedResources":{"cephObjectStores":{"disableHttp": null}}}}'
+            try:
+                self.storage_cluster_obj.patch(
+                    resource_name=self.storage_cluster_name,
+                    params=params,
```
ypersky1980: The null case can pass without actually proving anything.

  The test expects that setting disableHttp=null via merge patch will be rejected by the API (because it's a boolean field). But the code handles both outcomes — rejection (CommandFailed) and
  success — without distinguishing between them.

  The problem: 

if the patch is not rejected and actually removes disableHttp from the spec, the route check right after (self.get_route() is None) will still pass — not because the null patch had no effect, but simply because the operator hasn't had time to reconcile and recreate the route yet. The route is still gone from Step 1.

  So the test would report "null patch correctly had no effect" when in reality the patch did have an effect — it just hasn't manifested yet. That's a false positive.

 Suggested fix: explicitly assert that CommandFailed was raised. If it wasn't, the test should fail immediately rather than relying on a race-condition-dependent route check.

tests/functional/object/rgw/test_rgw_http_endpoint.py
```diff
@@ -318,31 +340,144 @@ def test_rgw_disable_http_endpoint(self):
...
+            params = '[{"op": "remove", "path": "/spec/managedResources/cephObjectStores/disableHttp"}]'
+            try:
+                self.storage_cluster_obj.patch(
+                    resource_name=self.storage_cluster_name,
+                    params=params,
+                    format_type="json",
+                )
+                log.info("disableHttp field removed via JSON patch")
+            except CommandFailed as e:
+                log.warning(f"Removing disableHttp field raised an error: {e}")
```
ypersky1980: The remove patch failing should fail the test, not be silently swallowed
                                
  The remove case catches CommandFailed and logs a warning, then continues as if the patch worked. But if the remove patch failed, disableHttp is still true in the StorageCluster — the HTTP route will never come back. The test will then spend 180 seconds waiting in wait_for_route_to_exist before failing with a confusing timeout error like "Route did not get created within 180 seconds." 
That hides the real problem (patch failed) behind a misleading symptom (route timeout). Either don't catch the exception here, or re-raise it after logging.

tests/functional/object/rgw/test_rgw_http_endpoint.py
udaysk23: Fixed

tests/functional/object/rgw/test_rgw_http_endpoint.py
udaysk23: Fixed

ypersky1980 (dismissed):
LGTM

tests/functional/object/rgw/test_rgw_http_endpoint.py
```diff
@@ -318,31 +340,147 @@ def test_rgw_disable_http_endpoint(self):
...
+        log.info("Step 4: Verifying HTTPS route is still functional")
+        https_route_data = self.wait_for_route_to_exist(
+            constants.RGW_ROUTE_INTERNAL_MODE_SECURE, timeout=60
+        )
+        https_host = https_route_data.get("spec", {}).get("host")
+        assert (
+            https_host
+        ), "HTTPS route host is missing; cannot validate endpoint accessibility"
+        try:
+            response = requests.get(f"https://{https_host}", timeout=10, verify=False)
```
ypersky1980: This HTTPS check does a single request with no retry, but the HTTP check above (lines 458-465) uses TimeoutSampler(timeout=300, sleep=10) because after pod restarts the endpoint needs time to start serving. 
The same applies here — in the "remove" case, RGW pods were just restarted, so the HTTPS endpoint might not be ready yet either. I suggest that you use check_http_accessibility (or a similar helper) with TimeoutSampler for consistency and to avoid flaky failures.

ypersky1980 (commented):
Left one comment.

tests/functional/object/rgw/test_rgw_http_endpoint.py
```diff
@@ -318,31 +340,147 @@ def test_rgw_disable_http_endpoint(self):
...
+        """
+        Test that HTTP endpoint is re-enabled when disableHttp is unset via
+        null value (merge patch) or field removal (JSON patch).
+
+        Test Steps:
+            1. Set disableHttp=true to disable HTTP and verify route is deleted
+            2. Apply patch to unset disableHttp (either null merge or remove JSON patch)
+            3. Verify HTTP route is recreated and endpoint is accessible
+            4. Verify HTTPS route is still functional
+        """
```
sagihirshfeld, line 409: Looks like a test name mismatch

```suggestion
    def  test_rgw_disable_http_unset(self, patch_action):
        """
        Test that HTTP endpoint is re-enabled when disableHttp is unset via
        null value (merge patch) or field removal (JSON patch).

        Test Steps:
            1. Set disableHttp=true to disable HTTP and verify route is deleted
            2. Apply patch to unset disableHttp (either null merge or remove JSON patch)
            3. Verify HTTP route is recreated and endpoint is accessible
            4. Verify HTTPS route is still functional
        """
```

tests/functional/object/rgw/test_rgw_http_endpoint.py
sagihirshfeld: Patching `"disableHttp": null` should work to remove a value or a sub-dictionary in a YAML. i.e [this example](https://github.com/red-hat-storage/ocs-ci/blob/master/tests/cross_functional/scale/test_multiple_storage_coexistence.py#L112-L114) from OCS-CI.

I tried this approach manually and I didn't run into any issue:
```
❯ oc get route | grep cephobjectstore
ocs-storagecluster-cephobjectstore          ocs-storagecluster-cephobjectstore-openshift-storage.apps.ukurundw-50.qe.rh-ocs.com                 rook-ceph-rgw-ocs-storagecluster-cephobjectstore   http            edge/Allow           None
ocs-storagecluster-cephobjectstore-secure   ocs-storagecluster-cephobjectstore-secure-openshift-storage.apps.ukurundw-50.qe.rh-ocs.com          rook-ceph-rgw-ocs-storagecluster-cephobjectstore   https           reencrypt/Redirect   None

❯  oc patch storagecluster ocs-storagecluster --type merge -p '{"spec":{"managedResources":{"cephObjectStores":{"disableHttp":true}}}}'

storagecluster.ocs.openshift.io/ocs-storagecluster patched

❯ oc get route | grep cephobjectstore
ocs-storagecluster-cephobjectstore-secure   ocs-storagecluster-cephobjectstore-secure-openshift-storage.apps.ukurundw-50.qe.rh-ocs.com          rook-ceph-rgw-ocs-storagecluster-cephobjectstore   https           reencrypt/Redirect   None

❯  oc patch storagecluster ocs-storagecluster --type merge -p '{"spec":{"managedResources":{"cephObjectStores":{"disableHttp":null}}}}'

storagecluster.ocs.openshift.io/ocs-storagecluster patched

❯ oc get route | grep cephobjectstore
ocs-storagecluster-cephobjectstore          ocs-storagecluster-cephobjectstore-openshift-storage.apps.ukurundw-50.qe.rh-ocs.com                 rook-ceph-rgw-ocs-storagecluster-cephobjectstore   http            edge/Allow           None
ocs-storagecluster-cephobjectstore-secure   ocs-storagecluster-cephobjectstore-secure-openshift-storage.apps.ukurundw-50.qe.rh-ocs.com          rook-ceph-rgw-ocs-storagecluster-cephobjectstore   https           reencrypt/Redirect   None
```

Unless I'm missing something here, we should expect the same behavior out of both params instead of this massive divergence in test behavior

ypersky1980 (approved):
LGTM

nehaberry (approved):
Approved based on logic. Cosmetic changes are already addressed.

PR #15364, opened 2026-06-10
Cache oc plugin list output in exec_cmd to avoid redundant subprocess calls

AviadP:
- Cache the output of `oc plugin list` in a module-level variable so it runs once per process instead of on every `exec_cmd` call
- Only cache on success (`returncode == 0`) — when `oc plugin list` fails (e.g. no plugins installed, exit code 1), the cache stays empty and retries every call, preserving the original behavior
- Fall back to empty list when cache is not populated, matching the original behavior for environments without plugins

### Context

Every `oc` command routed through `exec_cmd` spawns a separate `oc plugin list` subprocess to determine where to insert `--kubeconfig` (index 1 for regular commands, index 2 for plugin commands). This adds ~60ms of overhead per call — scanning every directory in `$PATH` for `oc-*` binaries — even though the plugin list never changes during a test run.

With this change:
- **Plugins installed + `oc plugin list` succeeds**: cached on first call, reused for the rest of the process → ~52% reduction in `oc` subprocess overhead
- **No plugins + `oc plugin list` fails**: retries every call → identical to current behavior, no regression

## Summary by CodeRabbit

* **Chores**
  * Optimized command execution performance by caching plugin list queries, reducing unnecessary system calls during repeated operations.

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -73,6 +73,7 @@ from ocs_ci.ocs.constants import HCI_PROVIDER_CLIENT_PLATFORMS
 log = logging.getLogger(__name__)
 
 # variables
+_oc_plugin_list_cache = None
 mounting_dir = "/mnt/cephfs/"
 clients = []
 md5sum_list1 = []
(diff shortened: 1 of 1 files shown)


AviadP:
> PR validation
> 
> Cluster Name: Cluster Configuration: PR Test Suite: tier1 PR Test Path: tests/ Additional Test Params: OCP VERSION: 4.22 OCS VERSION: 4.22 tested against branch: master
> 
> [Job FAILED](https://url.corp.redhat.com/7a4fdc9) (installation failed, tests not executed).

12 failures, nothing is related to changes in this PR

petr-balogh (approved):

dahorak (approved):

PR #15386, opened 2026-06-11
Add NooBaa health check to health_checker fixture

### LINKED ISSUE #1962: ocs_ci.ocs.cluster.CephCluster.cluster_health_check should check the storagecluster PHASE
Storage cluster phase should be validated as part of cluster health check. It should be 'Ready'

AviadP:
Fixes https://github.com/red-hat-storage/ocs-ci/issues/1962

## Summary

- Add NooBaa CR phase check to `health_checker` setup alongside the existing Ceph health check
- Skip NooBaa-dependent tests (`tests/functional/object/`) when NooBaa is not Ready; let Ceph tests proceed
- Modify `verify_storage_cluster()` to skip SC phase verification when NooBaa was already unhealthy at test setup, preventing misleading teardown failures
- Track NooBaa health state in `config.RUN` for `failure_propagator` reporting

The skip in `verify_storage_cluster()` is guarded by 4 safety conditions: the NooBaa flag must be set, `NoobaaInitializing` must be the Progressing reason, no CephCluster conditions present, and NooBaa phase must match what was recorded at setup.

## Test plan

- [x] Run tier1 on healthy cluster — verify no behavior change (+1 API call per test)
- [ ] Simulate NooBaa failure (scale down noobaa-core-0) — verify MCG tests skip, Ceph tests pass
- [ ] Verify flag clears when NooBaa recovers between tests
- [ ] Verify `verify_storage_cluster()` re-raises when Ceph has issues alongside NooBaa

## Summary by CodeRabbit

* **Improvements**
  * Added NooBaa/Ceph health gating at test setup to detect NooBaa initialization delays and adjust which health checks run.
  * Introduced new runtime configuration flags to track NooBaa readiness, health-failure origin, and Ceph health mismatch context.
  * Enhanced failure reporting to append NooBaa and Ceph mismatch summaries to skip/failure messages and email-visible output.

* **Tests**
  * Updated health-check fixture logic to record NooBaa phase at setup, re-evaluate at teardown, and conditionally skip affected tests.

### DIFF
--- a/ocs_ci/framework/conf/default_config.yaml
+++ b/ocs_ci/framework/conf/default_config.yaml
@@ -39,9 +39,13 @@ RUN:
   load_status: None
   skip_reason_test_found: {}
   skipped_tests_ceph_health: 0
+  skipped_tests_noobaa_health: 0
   number_of_tests: None
   skipped_on_ceph_health_ratio: 0
   skipped_on_ceph_health_threshold: 0
+  noobaa_not_ready_at_setup: {}
+  noobaa_health_failure_source: {}
+  sc_ceph_health_mismatch: {}
 
 # In this section we are storing all deployment related configuration but not
 # the environment related data as those are defined in ENV_DATA section.
--- a/ocs_ci/ocs/resources/storage_cluster.py
+++ b/ocs_ci/ocs/resources/storage_cluster.py
@@ -9,6 +9,7 @@ import re
 import tempfile
 import json
 import time
+from typing import Optional
 
 from jsonschema import validate
 from jsonschema.exceptions import ValidationError
@@ -1303,6 +1304,55 @@ def verify_storage_system():
         )
 
 
+def get_noobaa_phase(namespace: str) -> Optional[str]:
+    """
+    Get the current phase of the NooBaa CR via a single lightweight API call.
+
+    Args:
+        namespace (str): Kubernetes namespace where NooBaa is deployed.
+
+    Returns:
+        Optional[str]: NooBaa phase string (e.g. "Ready", "Configuring"),
+            or None if the CR cannot be retrieved.
+
+    """
+    try:
+        noobaa = OCP(kind="noobaa", namespace=namespace).get(
+            resource_name=constants.NOOBAA_RESOURCE_NAME
+        )
+        return noobaa.get("status", {}).get("phase", "")
+    except CommandFailed as ex:
+        log.warning("Failed to get NooBaa CR (%s) — skipping NooBaa health check", ex)
+        return None
+
+
+def get_storage_cluster_phase(namespace: str) -> Optional[str]:
+    """
+    Get the current phase of the StorageCluster CR via a single lightweight
+    API call.
+
+    Args:
+        namespace (str): Kubernetes namespace where the StorageCluster is
+            deployed.
+
+    Returns:
+        Optional[str]: StorageCluster phase string (e.g. "Ready",
+            "Progressing"), or None if the CR cannot be retrieved.
+
+    """
+    try:
+        sc_name = config.ENV_DATA["storage_cluster_name"]
+        sc = OCP(kind=constants.STORAGECLUSTER, namespace=namespace).get(
+            resource_name=sc_name
+        )
+        return sc.get("status", {}).get("phase", "")
+    except CommandFailed as ex:
+        log.warning(
+            "Failed to get StorageCluster CR (%s) — skipping SC phase check", ex
+        )
+        return None
+
+
 def verify_storage_cluster():
     """
     Verify storage cluster status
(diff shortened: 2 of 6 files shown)


ocs_ci/framework/conf/default_config.yaml
```diff
@@ -42,6 +42,8 @@ RUN:
   number_of_tests: None
   skipped_on_ceph_health_ratio: 0
   skipped_on_ceph_health_threshold: 0
+  noobaa_not_ready_at_setup: {}
+  noobaa_health_failure_source: {}
```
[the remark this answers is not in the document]
AviadP: Thanks for the suggestion, but intentionally skipping this one.

These keys are documented by their usage — `conftest.py` (the producer) and `test_failure_propagator.py` (the consumer) make the purpose and expected contents clear. Adding YAML comments would duplicate what the code already says and add maintenance overhead when the structure changes.

Project convention is to keep comments minimal — inline YAML comments on config keys aren't used elsewhere in this file.

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -1303,6 +1303,28 @@ def verify_storage_system():
...
+
+    """
+    try:
+        noobaa = OCP(kind="noobaa", namespace=namespace).get(
+            resource_name=constants.NOOBAA_RESOURCE_NAME
+        )
+        return noobaa.get("status", {}).get("phase", "")
+    except CommandFailed:
+        log.warning("Failed to get NooBaa CR — skipping NooBaa health check")
+        return None
```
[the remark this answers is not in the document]
AviadP: Traced the full flow — this can't happen. Here's why:

**`None` can never suppress the exception.** `get_noobaa_phase()` returns `None` only on `CommandFailed`. At setup, `None` triggers an early return (conftest.py:2192-2193) and the `noobaa_not_ready_at_setup` flag is **never set**. So in `verify_storage_cluster()`, `config.RUN.get("noobaa_not_ready_at_setup")` returns `None` → the entire NooBaa handling block is skipped → exception re-raised normally.

**`""` (empty string) is also safe.** If the NooBaa CR exists but has no `phase` field, both setup and teardown get `""`. The phases match, Ceph is confirmed healthy, and the NooBaa condition is present — this is a genuine pre-existing issue. Skipping SC verification is the correct behavior here.

**If current gets `None` but setup recorded a real phase:** `None != "SomePhase"` → `True` → raises with phase-change warning. Also correct.

The suggested `None` guards add code for a path that can't be reached. The cross-function state flow (setup early-return prevents the flag from being set) is the actual safety mechanism — not an oversight.

tests/conftest.py
```diff
@@ -2206,6 +2251,34 @@ def finalizer():
...
+                                    "phase": nb_phase,
+                                }
+                        elif nb_was_broken and nb_is_broken:
+                            log.info(
+                                "NooBaa still not Ready at teardown (phase: %s)",
+                                nb_phase,
+                            )
+                        elif nb_was_broken and not nb_is_broken:
+                            log.info("NooBaa recovered to Ready during test execution")
+
```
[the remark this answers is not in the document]
AviadP: This is by design — the teardown NooBaa tracking is **purely informational**. It logs and records data for the failure propagator. It doesn't skip tests, doesn't raise exceptions, and doesn't change any test outcome.

If `nb_phase` is `None` (API failure during teardown), the worst case is one misleading INFO log line. The flag itself (`noobaa_not_ready_at_setup`) isn't cleared here — it's always re-evaluated from scratch at the next test's setup. So there's no state corruption.

Adding a `None` branch would add complexity to handle a rare edge case (API failure during teardown) with zero functional impact. The "do no harm" principle for this PR means: if we're unsure, don't add code that could introduce new failure modes in teardown.

AviadP:
Addressing the findings from the review that couldn't be posted inline:

---

**conftest.py:2357-2362 — "Run NooBaa gate in recovered-Ceph path"**

This conflicts with the `CephHealthRecoveredException` design. When this exception fires, the test runs but **is already scheduled to fail in teardown** (lines 2306-2319 raise the stored exception). Adding `check_noobaa_health_at_setup()` here would:
- For MCG tests: `pytest.skip()` — replacing a planned *failure* with a *skip*, losing the Ceph recovery signal entirely
- For non-MCG tests: set the leniency flag on `verify_storage_cluster()` — wrong, because we *want* it to fail when Ceph just had issues

The deferred-failure pattern exists precisely so the issue gets investigated. Skipping on top of it defeats the purpose.

---

**conftest.py:2192-2210 — "Don't treat None as healthy at setup"**

Two separate concerns here:

1. **Initial `None` return (line 2192):** This is a deliberate **fail-open** design. `None` means `CommandFailed` — a transient API failure, not a NooBaa health problem. Blocking MCG tests because of one failed API call would create more false skips than it prevents.

2. **Retry clearing flag on `None` (line 2208):** Also intentional. If the first check saw NooBaa unhealthy but the retry can't reach the API, clearing the flag falls back to **today's behavior** — no new skips, no new leniency. The alternative (keeping the flag) would skip MCG tests and soften `verify_storage_cluster()` based on one reading + one API failure. That creates new blind spots, which is what this PR is designed to avoid.

---

**storage_cluster.py:3908-3912 — "Use log.exception()"**

Pre-existing code, not introduced by this PR. Out of scope.

---

**test_failure_propagator.py:113-119 — "Use defensive .get() access"**

We control both the producer (conftest.py:2270-2272) and consumer. The dict is always created with both `phase` and `test_name` keys — it's never partially constructed. Using `[]` correctly surfaces a programming error if someone later breaks the producer. Using `.get('phase', 'Unknown')` would silently produce a wrong report, which is worse during an already-bad failure investigation.

tests/conftest.py
```diff
@@ -2206,6 +2254,36 @@ def finalizer():
                         )
                         multi_storagecluster_external_health_passed = True
 
+                    if not ocsci_config.DEPLOYMENT.get(
+                        "external_mode"
+                    ) and not ocsci_config.COMPONENTS.get("disable_noobaa"):
+                        from ocs_ci.ocs.resources.storage_cluster import (
+                            get_noobaa_phase,
+                        )
```
udaysk23: [Very minor]I can see `get_noobaa_phase` twice in `health_checker` fixture.

udaysk23 (dismissed):

tests/conftest.py
```diff
@@ -2264,6 +2342,7 @@ def finalizer():
                 if not ocsci_config.DEPLOYMENT.get("multi_storagecluster"):
                     if status:
                         log.info("Ceph health check passed at setup")
+                        check_noobaa_health_at_setup()
```
ebenahar: how about first, before ceph and noobaa health at setup, to check the storage cluster state? this way we get a higher level health check and if needed proceed to ceph and noobaa health checks?

AviadP: so something like that:
1. check SC->if ok proceed, if not ok check each component?

tests/conftest.py
AviadP: fixed

tests/conftest.py
AviadP: done

AviadP:
> PR validation
> 
> Cluster Name: Cluster Configuration: PR Test Suite: tier1 PR Test Path: tests/ Additional Test Params: OCP VERSION: 4.22-ga OCS VERSION: 4.22 tested against branch: master
> 
> [Job UNSTABLE](https://url.corp.redhat.com/e68a35c) (some or all tests failed).

non of the failure is related to PR. one test skipped instead of fail due to noobaa unhealthy

udaysk23 (approved):

ebenahar (approved):

PR #15401, opened 2026-06-14
Wait for CephFS nodeplugin pods before configuring image registry

yitzhak12:
After ODF deployment, `change_registry_backend_to_ocs()` was called immediately without verifying that the CephFS CSI node plugin DaemonSet was fully running on all worker nodes. If a node rebooted during deployment (or image pulls were slow), the nodeplugin pods could be stuck in ContainerCreating, leaving the CephFS driver unregistered on those nodes. Any attempt to attach a CephFS PVC to such a node then fails with:

 

>  CSINode <node> does not contain driver
>   openshift-storage.cephfs.csi.ceph.com

Add `get_cephfs_nodeplugin_pods()` helper in `pod.py`. Use it in `deployment.py` to restart the nodeplugin pods before the wait to clear any transient ContainerCreating state caused by kubelet cache sync issues at reboot time, then block until all `worker_count` pods are Running (up to 600s) before reconfiguring the image registry backend.

See more details in the job: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/69130/console

## Release Notes

* **Bug Fixes**
  * Improved CephFS CSI node plugin pod initialization during OCS deployment to prevent transient container state issues before registry backend configuration.

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -123,6 +123,7 @@ from ocs_ci.ocs.resources.pod import (
     wait_for_ceph_cmd_execute_successfully,
     get_operator_pods,
     delete_pods,
+    get_cephfs_nodeplugin_pods,
     wait_for_pods_by_label_count,
     wait_for_pods_to_be_running,
     wait_for_pods_to_be_in_statuses,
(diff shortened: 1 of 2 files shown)


ocs_ci/deployment/deployment.py
```diff
@@ -2321,6 +2322,26 @@ def deploy_ocs(self):
             )
 
         if not config.COMPONENTS["disable_cephfs"]:
+            worker_count = len(get_worker_nodes())
```
ebenahar, line 2325: since this is added as part of a very long function, for the purpose of modularity and code simplification, how about putting this in a separate function to will be called from here?

PR #15413, opened 2026-06-15
Adds high priority tests for RGW STS feature

udaysk23:
Adds high priority tests for RGW STS feature

## Summary by CodeRabbit

* **Tests**
  * Enhanced RGW STS end-to-end coverage by standardizing service client creation for consistent endpoint and region handling.
  * Updated RGW STS configuration testing to use the standardized setup and consistent CephObjectStore selection.
  * Added AssumeRole scenarios covering successful role assumption, invalid role ARN handling, and missing RoleArn parameter validation.
  * Added token validation to verify S3 access with assumed-role credentials, including invalid credential error codes and expired-token behavior.

### DIFF
--- a/tests/functional/object/rgw/test_rgw_sts.py
+++ b/tests/functional/object/rgw/test_rgw_sts.py
@@ -4,11 +4,12 @@ Test RGW STS (Security Token Service) functionality in OCS-CI
 
 import json
 import logging
+import time
 import uuid
 
 import boto3
 import pytest
-from botocore.exceptions import ClientError
+from botocore.exceptions import ClientError, ParamValidationError
 
 from ocs_ci.framework import config
 from ocs_ci.framework.pytest_customization.marks import (
@@ -23,6 +24,7 @@ from ocs_ci.helpers.helpers import create_unique_resource_name
 from ocs_ci.ocs import constants
 from ocs_ci.ocs.ocp import OCP
 from ocs_ci.ocs.resources.pod import get_ceph_tools_pod
+from ocs_ci.utility.utils import TimeoutSampler
 
 logger = logging.getLogger(__name__)
 
@@ -58,11 +60,7 @@ class TestRGWSTS:
                     - access_key: AWS access key
                     - secret_key: AWS secret key
         """
-        cephobjectstore_name = OCP(
-            kind=constants.CEPHOBJECTSTORE,
-            resource_name="ocs-storagecluster-cephobjectstore",
-            namespace=config.ENV_DATA["cluster_namespace"],
-        ).get()["metadata"]["name"]
+        cephobjectstore_name = constants.CEPHOBJECTSTORE_NAME
         toolbox = get_ceph_tools_pod()
 
         # Generate unique IDs to avoid conflicts
@@ -138,13 +136,8 @@ class TestRGWSTS:
 
         endpoint, _, _ = rgw_obj.get_credentials()
 
-        iam_client = boto3.client(
-            "iam",
-            endpoint_url=endpoint,
-            aws_access_key_id=access_key,
-            aws_secret_access_key=secret_key,
-            region_name=rgw_obj.region,
-            verify=False,
+        iam_client = self.make_boto3_client(
+            "iam", endpoint, rgw_obj.region, access_key, secret_key
         )
 
         return iam_client, {
@@ -152,6 +145,8 @@ class TestRGWSTS:
             "user_id": user_id,
             "access_key": access_key,
             "secret_key": secret_key,
+            "endpoint": endpoint,
+            "region": rgw_obj.region,
         }
 
     def trust_policy_creation(self):
@@ -185,6 +180,21 @@ class TestRGWSTS:
             "Statement": [{"Effect": "Allow", "Action": ["s3:*"], "Resource": "*"}],
         }
         return policy
+
+    def make_boto3_client(
+        self, service, endpoint, region, access_key, secret_key, session_token=None
+    ):
+        """Create a boto3 client with common RGW connection parameters."""
+        kwargs = dict(
+            endpoint_url=endpoint,
+            aws_access_key_id=access_key,
+            aws_secret_access_key=secret_key,
+            region_name=region,
+            verify=False,
+        )
+        if session_token is not None:
+            kwargs["aws_session_token"] = session_token
+        return boto3.client(service, **kwargs)
 
     def put_object(self, s3_client, bucket_name, object_key, data):
         """
@@ -226,7 +236,6 @@ class TestRGWSTS:
     @post_upgrade
     def test_rgw_sts_configuration(
         self,
-        rgw_obj,
         rgw_iam_client_creation,
     ):
         """
@@ -246,13 +255,15 @@ class TestRGWSTS:
 
         # Unpack the fixture return values
         iam_client, rgw_account_and_user = rgw_iam_client_creation
+        endpoint = rgw_account_and_user["endpoint"]
+        region = rgw_account_and_user["region"]
 
+        cephobjectstore_name = constants.CEPHOBJECTSTORE_NAME
         cephobjectstore_obj = OCP(
             kind=constants.CEPHOBJECTSTORE,
-            resource_name="ocs-storagecluster-cephobjectstore",
+            resource_name=cephobjectstore_name,
             namespace=config.ENV_DATA["cluster_namespace"],
         )
-        cephobjectstore_name = cephobjectstore_obj.get()["metadata"]["name"]
 
         # Step 1: Check EnableSTS parameter in StorageCluster
         logger.info("Step 1: Checking EnableSTS parameter in StorageCluster")
@@ -365,17 +376,9 @@ class TestRGWSTS:
 
             logger.info(f"IAM user access key created: {iam_user_access_key[:10]}...")
 
-            # Get endpoint from rgw_obj
-            endpoint, _, _ = rgw_obj.get_credentials()
-
             # Create S3 client with IAM user credentials
-            iam_user_s3_client = boto3.client(
-                "s3",
-                endpoint_url=endpoint,
-                aws_access_key_id=iam_user_access_key,
-                aws_secret_access_key=iam_user_secret_key,
-                region_name=rgw_obj.region,
-                verify=False,
+            iam_user_s3_client = self.make_boto3_client(
+                "s3", endpoint, region, iam_user_access_key, iam_user_secret_key
             )
 
             # Create bucket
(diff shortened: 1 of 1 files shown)


tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -524,3 +520,306 @@ def test_rgw_sts_configuration(
...
+                f"Step 7: Got expected error with invalid Session Token - Error code: {error_code}"
+            )
+            assert (
+                error_code == "InvalidArgument"
+            ), f"Unexpected error code for invalid session token: {error_code}"
+
+        # Step 8: Wait for STS token to expire
+        logger.info(f"Step 8: Waiting {token_duration + 10}s for STS token to expire")
+        time.sleep(token_duration + 10)
+        logger.info("Step 8: Token expiration wait complete")
```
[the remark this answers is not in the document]
udaysk23: E               botocore.exceptions.ParamValidationError: Parameter validation failed:
E               Invalid value for parameter DurationSeconds, value: 120, valid min value: 900

../../env/lib/python3.11/site-packages/botocore/validate.py:424: ParamValidationError

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -524,3 +520,336 @@ def test_rgw_sts_configuration(
...
+            3. Fetch credentials using incorrect AssumeRole name - expect failure
+            4. Fetch credentials without passing RoleARN - expect failure
+        """
+        iam_client, rgw_account_and_user = rgw_iam_client_creation
+        endpoint = rgw_account_and_user["endpoint"]
+
+        # Step 1: Create IAM user, role, and attach permission policy
+        logger.info("Step 1: Creating IAM user, role, and attaching permission policy")
+        iam_user_name = create_unique_resource_name("sts-iam", "user")
+        iam_client.create_user(UserName=iam_user_name)
```
ypersky1980, line 545: Both new tests (test_rgw_sts_assumerole_scenarios and test_rgw_sts_token_validation) create IAM users, roles, policies, access keys, and S3 buckets but register no finalizer. The rgw_iam_client_creation fixture only removes the radosgw account user via radosgw-admin user rm --purge-data. IAM roles, inline policies, and sub-users created inside the test body may not be cascade-deleted, causing resource leaks. Both tests need request.addfinalizer to clean up  the IAM users/roles/policies/buckets they create (add request to the test parameters).

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -524,3 +520,336 @@ def test_rgw_sts_configuration(
...
+                error_code == "InvalidAccessKeyId"
+            ), f"Unexpected error code for invalid access key: {error_code}"
+
+        # Step 6: AssumeRole with DurationSeconds=0 - expect failure
+        logger.info("Step 6: AssumeRole with DurationSeconds=0")
+        try:
+            sts_client.assume_role(
+                RoleArn=role_arn,
+                RoleSessionName=session_name,
+                DurationSeconds=0,
```
ypersky1980, line 785: Steps 6–7 (DurationSeconds=0 and DurationSeconds=-1) catch ParamValidationError, which is raised by botocore locally before any request reaches RGW. These steps test the boto3 SDK's client-side input validation, not the RGW STS server. To test RGW's server-side validation, use a value that passes botocore but is rejected by the server — e.g., DurationSeconds=899 (below RGW's 900s minimum but above botocore's floor), and expect a ClientError from the server.

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -524,3 +520,336 @@ def test_rgw_sts_configuration(
...
+        logger.info(
+            "Step 1: Creating IAM user, STS client, role, and attaching permission policy"
+        )
+        iam_user_name = create_unique_resource_name("sts-iam", "user")
+        iam_client.create_user(UserName=iam_user_name)
+        logger.info(f"IAM user {iam_user_name} created successfully")
+
+        iam_client.attach_user_policy(
+            UserName=iam_user_name,
+            PolicyArn="arn:aws:iam::aws:policy/AmazonS3FullAccess",
```
ypersky1980, line 666: attach_user_policy(PolicyArn="arn:aws:iam::aws:policy/AmazonS3FullAccess") references an AWS-managed policy ARN. RGW has no concept of AWS-managed policy ARNs — this call either silently
  fails (meaning the IAM user operates on default permissions and the test passes for the wrong reason) or relies on undocumented RGW behavior. The existing test_rgw_sts_configuration correctly
  uses put_role_policy with an inline policy document, which is the reliable RGW approach.

ypersky1980 (commented):
Left 3 comments.

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: @ypersky1980 All users, data, buckets are getting deleted when  radosgw-admin user rm --purge-data is getting triggered from RGW side since that is main user like admin for all IAM user created by the test

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: I tested attach_user_policy with arn:aws:iam::aws:policy/AmazonS3FullAccess on a real RGW cluster and it works as expected — the IAM user receives S3 permissions and bucket operations succeed. RGW does appear to recognize and apply this AWS-managed policy ARN correctly in practice.

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: You're right that DurationSeconds=0 and DurationSeconds=-1 trigger botocore's client-side validation (ParamValidationError) before the request reaches RGW. However, using DurationSeconds=899 would also be rejected by botocore since it enforces a minimum of 900 at the SDK level — there is no sub-900 value that bypasses botocore and reaches the RGW server.

These steps intentionally test that the STS client rejects invalid duration values before making a network call, which is still a valid boundary condition to cover.

ypersky1980 (dismissed):
LGTM

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -524,3 +520,336 @@ def test_rgw_sts_configuration(
...
+            assert (
+                error_code == "InvalidArgument"
+            ), f"Unexpected error code for invalid session token: {error_code}"
+
+        # Step 10: Wait for STS token to expire
+        logger.info(f"Step 10: Waiting {token_duration + 10}s for STS token to expire")
+        time.sleep(token_duration + 10)
+        logger.info("Step 10: Token expiration wait complete")
+
+        # Step 11: Get bucket content after token expiration - expect failure
```
sagihirshfeld, line 856: Are you sure enough of the timing? If infra issues such as network disruptions can delay the expiration by 10 seconds or more, then step 11 will fail with a false negative.

Please consider wrapping step 11 in a TimeoutSampler so it tries a few times before failing the test

udaysk23: Fixed in latest commit

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -524,3 +521,346 @@ def test_rgw_sts_configuration(
...
+        correct_role_arn = f"arn:aws:iam::{account_id}:role/{role_name}"
+        logger.info(f"Assuming role with correct ARN: {correct_role_arn}")
+
+        assume_role_response = sts_client.assume_role(
+            RoleArn=correct_role_arn,
+            RoleSessionName=session_name,
+        )
+        credentials = assume_role_response["Credentials"]
+        assert credentials["AccessKeyId"], "AccessKeyId should not be empty"
+        assert credentials["SecretAccessKey"], "SecretAccessKey should not be empty"
```
ypersky1980, line 591: Step 2 proves credentials are returned but never that they actually work. The test asserts AccessKeyId, SecretAccessKey, and SessionToken are non-empty strings, then moves on. It never makes a single API call with those credentials. Compare with test_rgw_sts_configuration which uses assumed credentials to list_buckets, put_object, and get_object. Without at least one S3 call (e.g. list_buckets), this step only validates the AssumeRole response format, not that the credentials grant actual access.

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -524,3 +521,346 @@ def test_rgw_sts_configuration(
...
+            assert False, "Expected ClientError for incorrect role ARN but got success"
+        except ClientError as e:
+            error_code = e.response["Error"]["Code"]
+            logger.info(
+                f"Step 3: Got expected error for incorrect role ARN - Error code: {error_code}"
+            )
+            assert error_code in (
+                "NoSuchEntity",
+                "AccessDenied",
+                "InvalidClientTokenId",
```
ypersky1980: Step 3 accepts 3 different error codes for a non-existent role — NoSuchEntity, AccessDenied, or InvalidClientTokenId. A well-defined API should return one specific error for "role doesn't exist." This over-permissive assertion could mask a regression if RGW changes its error behavior. Pin the assertion to the actual error code RGW returns (which from the passing CI run should  be known), and add a comment explaining why that specific code is expected.

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -524,3 +521,346 @@ def test_rgw_sts_configuration(
...
+        logger.info(
+            "Step 4: Listing bucket content using valid assumed role credentials"
+        )
+        assumed_s3_client = self.make_boto3_client(
+            "s3",
+            endpoint,
+            region,
+            assumed_access_key,
+            assumed_secret_key,
+            assumed_session_token,
```
ypersky1980, line 739: Step 4 only validates list_objects_v2 on an empty bucket — weak proof that the token works. The bucket created in step 2 is never populated. Listing an empty bucket is a minimal operation.
Before testing invalid credentials in steps 5-9, do a put_object + get_object with the valid assumed-role credentials. This proves the token grants full data-path access (not just metadata listing), making the subsequent "invalid credential" steps more meaningful — you first prove the token works end-to-end, then prove it fails when tampered with.

ypersky1980 (commented):
Left 3 comments.

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: This test's scope is specifically ARN handling (correct/incorrect/missing). Whether the returned credentials actually work on S3 is already covered end-to-end in test_rgw_sts_token_validation Steps 3-4. Adding an S3 call here would duplicate that coverage and blur the test's purpose.

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: Fixed in latest commit

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: Fixed in latest commit

sagihirshfeld (approved):
LGTM 👍

ypersky1980 (approved):
LGTM

PR #15425, opened 2026-06-16
Adds vector policy and medium proirity tests for s3 vector feature

udaysk23:
Adds vector policy and medium proirity tests for s3 vector feature

## Summary by CodeRabbit

* **New Features**
  * Enhanced bucket policy generation with configurable action and resource ARN service prefixes, enabling S3 Vectors–specific `s3vectors:` actions and `arn:aws:s3vectors:::` resources.
  * Added a public helper to generate S3 Vectors bucket policies with the correct preset formatting.
* **Tests**
  * Improved S3 Vector update/query checks by polling until changes are visible and the updated vector behavior matches expectations.
  * Added functional coverage for vector bucket policy create/update/delete and for vector metadata updates verified via filter queries.

### DIFF
--- a/ocs_ci/ocs/vector_utils.py
+++ b/ocs_ci/ocs/vector_utils.py
@@ -8,6 +8,7 @@ import logging
 import random
 
 from ocs_ci.ocs.bucket_utils import retrieve_verification_mode
+from ocs_ci.ocs.resources.bucket_policy import gen_bucket_policy
 from botocore.config import Config
 
 logger = logging.getLogger(__name__)
@@ -452,4 +453,41 @@ def delete_vector_bucket_policy(s3vectors_client, vector_bucket_name):
     """
     return s3vectors_client.delete_vector_bucket_policy(
         vectorBucketName=vector_bucket_name
+    )
+
+
+def gen_vector_bucket_policy(
+    user_list,
+    actions_list,
+    resources_list,
+    effect=None,
+    sid="statement",
+):
+    """
+    Generate a vector bucket policy dict for the S3 Vectors API.
+
+    Thin wrapper around gen_bucket_policy that applies the s3vectors: action
+    prefix and arn:aws:s3vectors::: resource format.
+
+    Args:
+        user_list (list or str): Principal ARNs, e.g. ["*"] or "arn:aws:iam::id:root"
+        actions_list (list): Action names without the "s3vectors:" prefix,
+                             e.g. ["ListVectors"] or ["DeleteVectors", "GetVectors"]
+        resources_list (list): Resource suffixes or full ARNs. Bare strings are
+                               prefixed with "arn:aws:s3vectors:::"; strings already
+                               starting with "arn:" are used as-is.
+        effect (str): "Allow" or "Deny". Default: "Allow"
+        sid (str): Statement ID. Default: "statement"
+
+    Returns:
+        dict: Vector bucket policy dict with a single statement, ready for json.dumps()
+    """
+    return gen_bucket_policy(
+        user_list=user_list,
+        actions_list=actions_list,
+        resources_list=resources_list,
+        effect=effect,
+        sid=sid,
+        action_prefix="s3vectors",
+        resource_arn_service="s3vectors",
     )
--- a/tests/functional/object/mcg/test_s3_vector.py
+++ b/tests/functional/object/mcg/test_s3_vector.py
@@ -1,7 +1,11 @@
+import json
 import logging
 import math
 import random
 import string
+import time
+
+import pytest
 
 from botocore.exceptions import ClientError, ParamValidationError
 
@@ -10,7 +14,6 @@ from ocs_ci.framework.pytest_customization.marks import (
     red_squad,
     post_upgrade,
     mcg,
-    jira,
 )
 from ocs_ci.framework.testlib import MCGTest
 from ocs_ci.framework import config
@@ -69,6 +72,20 @@ class TestS3Vector(MCGTest):
                 )
             else:
                 raise
+
+    @staticmethod
+    def _assert_access_denied(fn, label, *args, **kwargs):
+        """Assert that fn raises AccessDeniedException; fail if it succeeds."""
+        try:
+            fn(*args, **kwargs)
+        except ClientError as e:
+            error_code = e.response.get("Error", {}).get("Code", "")
+            if error_code == "AccessDeniedException":
+                logger.info(f"  ✓ {label} denied as expected by policy")
+            else:
+                raise
+        else:
+            pytest.fail(f"{label} succeeded when policy should have denied it")
 
     @tier2
     @post_upgrade
@@ -997,7 +1014,14 @@ class TestS3Vector(MCGTest):
             )
             logger.info(f"✓ [{dim_label}] Updated vector: {target_key}")
 
-            results_step6 = vector_utils.query_vectors(
+            # Lance DB indexes updates asynchronously — poll until the distance
+            # changes or the key drops out of results, rather than querying once
+            # immediately after the put (which races the index flush on Jenkins).
+            results_step6 = []
+            for _sample in TimeoutSampler(
+                60,
+                5,
+                vector_utils.query_vectors,
                 s3vectors_client,
                 query_vector=query_vector,
                 top_k=top_k,
@@ -1005,7 +1029,21 @@ class TestS3Vector(MCGTest):
                 indexName=index_name,
                 returnDistance=True,
                 returnMetadata=True,
-            ).get("vectors", [])
+            ):
+                results_step6 = _sample.get("vectors", [])
+                if len(results_step6) != top_k:
+                    continue
+                keys_now = [r.get("key") for r in results_step6]
+                if target_key not in keys_now:
+                    break
+                dist_now = next(
+                    r.get("distance")
+                    for r in results_step6
+                    if r.get("key") == target_key
+                )
+                if dist_now != target_distance_before:
+                    break
+
             assert len(results_step6) == top_k, (
                 f"[{dim_label}] Expected exactly {top_k} results after update, "
                 f"got {len(results_step6)}"
@@ -1093,7 +1131,6 @@ class TestS3Vector(MCGTest):
 
         logger.info("✅ Vector query operations test completed successfully!")
 
-    @jira("DFBUGS-7337")
     @tier2
     def test_vector_distance_metrics(self, vector_bucket_factory, mcg_obj):
         """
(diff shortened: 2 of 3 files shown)


tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -1310,3 +1310,558 @@ def test_vector_distance_metrics(self, vector_bucket_factory, mcg_obj):
...
+            # Step 13: Policy-not-found error is the expected NooBaa/S3 response
+            error_code = e.response.get("Error", {}).get("Code", "")
+            assert (
+                error_code == "NoSuchVectorBucketPolicy"
+            ), f"Unexpected error after policy deletion: {error_code} — {e}"
+            logger.info(
+                f"Step 13: ✓ GetVectorBucketPolicy raised '{error_code}' after deletion — "
+                f"default access rules apply, no explicit policy set"
+            )
+
```
ypersky1980, line 1724: After policy deletion, the test only verifies the API says "no policy" but never checks that previously-denied operations (GetVectors, DeleteVectors, PutVectors) are actually allowed again.
  Without that, a bug where enforcement persists after deletion would pass this test. Consider adding a step that re-runs the denied operations and asserts they succeed.

ypersky1980 (commented):
Left one comment

tests/functional/object/mcg/test_s3_vector.py
udaysk23: Addressed in latest commit

ocs_ci/ocs/vector_utils.py
```diff
@@ -453,3 +453,62 @@ def delete_vector_bucket_policy(s3vectors_client, vector_bucket_name):
...
+    user_list,
+    actions_list,
+    resources_list,
+    effect=None,
+    sid="statement",
+):
+    """
+    Generate a vector bucket policy dict for the S3 Vectors API.
+
+    Mirrors gen_bucket_policy from bucket_policy.py but uses the s3vectors:
```
sagihirshfeld: Rather than duplicating the logic, we could parameterize `gen_bucket_policy` with prefix arguments (backward-compatible defaults keep existing callers untouched):

```python
# In gen_bucket_policy, add two params with S3 defaults:
def gen_bucket_policy(
    user_list,
    actions_list,
    resources_list,
    effect=None,
    sid="statement",
    principal_property=None,
    action_property=None,
    resource_property=None,
    action_prefix="s3",
    resource_arn_service="s3",
):
    ...
    actions = list(map(lambda action: f"{action_prefix}:{action}", actions_list))
    resources = list(
        map(
            lambda r: r if r.startswith("arn:") else f"arn:aws:{resource_arn_service}:::{r}",
            resources_list,
        )
    )
    ...
```

Then `gen_vector_bucket_policy` becomes a thin wrapper:

```python
def gen_vector_bucket_policy(
    user_list, actions_list, resources_list, effect=None, sid="statement"
):
    """
    Generate a vector bucket policy dict for the S3 Vectors API.

    Thin wrapper around gen_bucket_policy with s3vectors prefixes.
    """
    return gen_bucket_policy(
        user_list=user_list,
        actions_list=actions_list,
        resources_list=resources_list,
        effect=effect,
        sid=sid,
        action_prefix="s3vectors",
        resource_arn_service="s3vectors",
    )
```

The `startswith("arn:")` passthrough for pre-formed ARNs is a nice bonus that benefits both S3 and S3Vectors callers.

udaysk23: Removed duplicates and code is using existing code

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -1310,3 +1331,592 @@ def test_vector_distance_metrics(self, vector_bucket_factory, mcg_obj):
...
+        logger.info(
+            f"✓ Initial policy defined with {len(initial_policy['Statement'])} statements: "
+            f"{[s['Sid'] for s in initial_policy['Statement']]}"
+        )
+
+        # Step 3: Call PutVectorBucketPolicy API
+        logger.info("Step 3: Calling PutVectorBucketPolicy API")
+        put_response = vector_utils.put_vector_bucket_policy(
+            s3vectors_client, vector_bucket_name, initial_policy_str
+        )
```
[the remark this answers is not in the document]
udaysk23: No need to delete bucket policy

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -1310,3 +1331,592 @@ def test_vector_distance_metrics(self, vector_bucket_factory, mcg_obj):
...
+            query_vector=query_vector,
+            top_k=10,
+            vectorBucketName=vector_bucket_name,
+            indexName=index_name,
+            filter={"genre": {"$eq": "genre_updated"}},
+            returnMetadata=True,
+            returnDistance=True,
+        ).get("vectors", [])
+        new_keys = [r.get("key") for r in new_filter_results]
+        assert target_key in new_keys, (
```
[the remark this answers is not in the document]
udaysk23: fixed

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -1310,3 +1331,592 @@ def test_vector_distance_metrics(self, vector_bucket_factory, mcg_obj):
...
+        except ClientError as e:
+            error_code = e.response.get("Error", {}).get("Code", "")
+            if error_code == "AccessDeniedException":
+                logger.info("  ✓ PutVectors denied as expected by policy")
+            else:
+                raise
+        else:
+            assert False, "PutVectors succeeded when policy should have denied it"
+
+        logger.info("✓ Updated policy enforcement verified")
```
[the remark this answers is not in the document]
udaysk23: fixed

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -1310,3 +1331,592 @@ def test_vector_distance_metrics(self, vector_bucket_factory, mcg_obj):
...
+            indexName=index_name,
+            filter={"genre": {"$eq": "scifi"}},
+            returnMetadata=True,
+            returnDistance=True,
+        ).get("vectors", [])
+        returned_keys = [r.get("key") for r in old_filter_results]
+        assert target_key not in returned_keys, (
+            f"Updated vector '{target_key}' still appears in genre='scifi' filter results: "
+            f"{returned_keys}"
+        )
```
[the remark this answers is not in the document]
udaysk23: fixed

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -1310,3 +1333,635 @@ def test_vector_distance_metrics(self, vector_bucket_factory, mcg_obj):
...
+
+        # Step 4: Verify policy application response
+        logger.info("Step 4: Verifying policy application response")
+        assert (
+            put_response.get("ResponseMetadata", {}).get("HTTPStatusCode") == 200
+        ), f"PutVectorBucketPolicy failed: {put_response}"
+        logger.info("✓ PutVectorBucketPolicy succeeded (HTTP 200)")
+
+        # Hardcoded sleep is needed for policy to reflect
```
ypersky1980, line 1439: test_vector_bucket_policy_operations has 3 hardcoded time.sleep(60) calls (lines ~1428, ~1570, ~1649), adding a minimum of 3 minutes of dead waiting to every CI run. The comment says  "even GetVectorBucketPolicy returning the policy does not guarantee enforcement" — fair for the initial apply. But the 2nd and 3rd sleeps (after policy update and after policy deletion) could use TimeoutSampler polling of the actual enforcement behavior instead: try the operation, if it doesn't match expected behavior yet, retry. This would cut minutes off CI runtime when propagation is faster than 60s.

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -1310,3 +1333,635 @@ def test_vector_distance_metrics(self, vector_bucket_factory, mcg_obj):
...
+        )
+        logger.info(
+            f"✓ '{target_key}' found in genre='genre_updated' results — "
+            f"new metadata filter works correctly"
+        )
+
+        # Step 7: Update vectors with category='A' and category='B' metadata
+        logger.info(
+            "Step 7: Updating vectors with category='A' and category='B' metadata"
+        )
```
ypersky1980, line 1929: Step 3 changes the genre on the server but not in the local Python variable. Step 7 re-sends the old local variable to the server, which undoes the genre change. The test passes by luck of ordering, not correctness. To be more specific:  
In step 3, vectors_a[0]'s genre is updated to "genre_updated" on the server — but only via the API call. The local Python dict vectors_a[0]["metadata"] still holds the original genre  (e.g. "scifi"). Then here in step 7, {**v["metadata"], "category": "A"} spreads that stale local dict, so vectors_a[0] is sent to the server with genre="scifi" + category="A" — silently  reverting the genre change from step 3. Steps 5-6 pass only because they run before step 7 overwrites the value. Fix: after step 3's put_vectors, update the local dict too:  vectors_a[0]["metadata"] = updated_metadata, so step 7 preserves the server state instead of reverting it.

tests/functional/object/mcg/test_s3_vector.py
```diff
@@ -1310,3 +1333,635 @@ def test_vector_distance_metrics(self, vector_bucket_factory, mcg_obj):
...
+            "DenyPutVectors" in updated_stmts
+        ), f"DenyPutVectors not found in updated policy. Sids: {set(updated_stmts)}"
+        assert (
+            "AllowListBucket" not in updated_stmts
+        ), f"Old rule AllowListBucket still present after update. Sids: {set(updated_stmts)}"
+        assert (
+            "DenyDeleteVector" not in updated_stmts
+        ), f"Old rule DenyDeleteVector still present after update. Sids: {set(updated_stmts)}"
+        assert (
+            updated_stmts["AllowQueryVectors"].get("Effect") == "Allow"
```
ypersky1980, line 1597: The policy test verifies ListVectors is allowed under the initial policy (step 6a), then the updated policy removes the AllowListBucket rule. Step 8 verifies AllowListBucket is gone  from the JSON, but step 9 never tests whether ListVectors is now denied or still allowed by default. This is a gap — the test verifies new allow/deny rules are enforced but doesn't verify that a removed allow rule actually changes behavior. Add a step 9d that calls ListVectors and asserts the expected outcome (denied or still allowed by default).

ypersky1980 (commented):
Left 3 comments.

tests/functional/object/mcg/test_s3_vector.py
udaysk23: fixed in latest commit

tests/functional/object/mcg/test_s3_vector.py
udaysk23: fixed in latest commit

tests/functional/object/mcg/test_s3_vector.py
udaysk23: fixed in latest commit

sagihirshfeld (approved):
LGTM

ypersky1980 (approved):
LGTM

PR #15456, opened 2026-06-19
Adds medium proirity UI tests for s3 vector feature

udaysk23:
Adds medium proirity UI tests for s3 vector feature

## Summary by CodeRabbit

* **New Features**
  * Expanded S3 Vector console UI automation to delete a specific index from the S3 Vector tab with name-based confirmation.
  * Added support for deleting multiple indices sequentially from the S3 Vector tab.
  * Added UI automation to delete an S3 Vector bucket from the bucket list with name-based confirmation.
* **Tests**
  * Added a functional UI test that provisions a bucket and indices, deletes one index, deletes remaining indices, deletes the bucket, and validates each step via backend API checks.
* **Bug Fixes**
  * Made index creation more idempotent by treating “already owned” errors as a successful outcome.

### DIFF
--- a/ocs_ci/ocs/ui/views.py
+++ b/ocs_ci/ocs/ui/views.py
@@ -3395,6 +3395,38 @@ s3_vector_tab = {
         "//button[@id='{}-link']",
         By.XPATH,
     ),
+    # Shared kebab toggle for any named table row (index rows and bucket rows
+    # share the same DOM pattern; callers differ only by which page they are on).
+    "row_kebab_by_name": (
+        "//tr[.//a[normalize-space()='{}']]//button[@aria-label='Kebab toggle']",
+        By.XPATH,
+    ),
+    # Delete index via row kebab on the bucket detail page
+    "delete_index_option": (
+        "//button[normalize-space()='Delete index']",
+        By.XPATH,
+    ),
+    "delete_index_name_input": (
+        "//input[@aria-label='Index name input']",
+        By.XPATH,
+    ),
+    "delete_index_confirm_button": (
+        "//div[@role='dialog']//button[.//span[text()='Delete index']]",
+        By.XPATH,
+    ),
+    # Delete vector bucket from the S3 Vector tab bucket list
+    "delete_vector_bucket_option": (
+        "//button[normalize-space()='Delete bucket']",
+        By.XPATH,
+    ),
+    "delete_vector_bucket_name_input": (
+        "//input[@aria-label='Bucket name input']",
+        By.XPATH,
+    ),
+    "delete_vector_bucket_confirm_button": (
+        "//div[@role='dialog']//button[.//span[text()='Delete bucket']]",
+        By.XPATH,
+    ),
 }
 locate_aws_regions = {
     "region_table": ('//*[@id="main-col-body"]/div[4]/div/table', By.XPATH)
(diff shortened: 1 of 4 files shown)


ocs_ci/ocs/ui/views.py
```diff
@@ -3399,6 +3395,40 @@
         "//button[@id='{}-link']",
         By.XPATH,
     ),
+    # Delete index via row kebab on the bucket detail page
+    "index_row_kebab_by_name": (
+        "//tr[.//a[normalize-space()='{}']]//button[@aria-label='Kebab toggle']",
+        By.XPATH,
```
ypersky1980, line 3402: index_row_kebab_by_name and vector_bucket_row_kebab_by_name (line ~3418) have the exact same XPath: "//tr[.//a[normalize-space()='{}']]//button[@aria-label='Kebab toggle']". Nothing distinguishes an index kebab from a bucket kebab. If the UI ever renders both tables on one page, or if a bucket and index share a name, the wrong element gets clicked. 
I suggest you either consolidate  into a single shared locator, or scope each one to its specific section (e.g., by a parent data-test attribute or table identifier).

ocs_ci/ocs/ui/views.py
```diff
@@ -3399,6 +3395,40 @@
...
+        "//button[normalize-space()='Delete index']",
+        By.XPATH,
+    ),
+    "delete_index_name_input": (
+        "//input[@aria-label='Index name input']",
+        By.XPATH,
+    ),
+    "delete_index_confirm_button": (
+        "//button[.//span[text()='Delete index']]",
+        By.XPATH,
```
ypersky1980, line 3415: delete_index_confirm_button uses "//button[.//span[text()='Delete index']]" — a page-wide selector. 
The kebab menu option delete_index_option above uses "//button[normalize-space()='Delete index']" with nearly identical text. If the kebab dropdown hasn't fully closed when the confirmation modal appears, this XPath could match the dropdown item instead of the modal button. Same issue with delete_vector_bucket_confirm_button. Please scope both confirm locators to the modal dialog, e.g.:
    "//div[contains(@class,'modal')]//button[.//span[text()='Delete index']]"

tests/functional/object/mcg/ui/test_s3_vector_ui.py
```diff
@@ -431,3 +431,188 @@ def teardown():
...
+            timeout=60,
+            sleep=3,
+            func=lambda: [
+                idx["indexName"]
+                for idx in vector_utils.list_indexes(
+                    s3vectors_client, vectorBucketName=noobaa_bucket_name
+                ).get("indexes", [])
+            ],
+        ):
+            if set(api_indices) == set(remaining_indices):
```
ypersky1980, line 570: Steps 8, 10, and 13 verify deletion outcomes only via S3 API, but this is a UI test. After deleting an index (steps 6-7), please also assert the deleted row is gone from the UI table (e.g., call get_index_names_from_bucket() and confirm the index is absent). Same for step 10 (UI table should be empty) and step 13 (bucket should be absent from the list via get_vector_bucket_names_from_tab()). The API checks confirm the backend processed the delete, but without UI assertions you'd miss bugs like a stale row remaining after delete or the table  showing an error state.

ypersky1980 (commented):
Left 3 comments.

tests/functional/object/mcg/ui/test_s3_vector_ui.py
udaysk23: Fixed in latest commit

ocs_ci/ocs/ui/views.py
udaysk23: Fixed in latest commit

ocs_ci/ocs/ui/views.py
udaysk23: Fixed in latest commit

sagihirshfeld (approved):

ypersky1980 (approved):
LGTM

anajain16 (approved):

PR #15470, opened 2026-06-22
Add df size check when mounting the export externally

amr1ta:
Testfix for bug, https://redhat.atlassian.net/browse/DFBUGS-3970

## Problem
When mounting a CephNFS PVC export externally (via the LoadBalancer ingress), df reports the total Ceph cluster capacity (~95 TB) instead of the PVC-requested size (e.g. 5 Gi). This is confusing for customers who do not want to expose their storage layer's total capacity to end users.

## What changed
Extended the existing test_outcluster_nfs_export (OCS-4273) with a new verification step (step 12) that:

Runs df -h --output=size <mount_point> on the external NFS client VM after the export is mounted.
Parses the reported total size.
Asserts it is the PVC size. Cluster capacity (~95 TB) is orders of magnitude larger and will cause the assertion to fail with a clear diagnostic message identifying the regression.
No new test is added — the check is woven into the existing outcluster export test where the external mount is already set up.

## Summary by CodeRabbit

* **Tests**
  * Enhanced NFS export validation by parameterizing the NFS PVC size instead of a hardcoded amount.
  * Added an external-client `df -h` verification to ensure the mounted filesystem reports an expected gigabyte-scale size and not a terabyte-scale value.
  * Updated the test step sequence to include the new external-client `df -h` verification.

### DIFF
--- a/tests/functional/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
+++ b/tests/functional/nfs_feature/test_nfs_feature_enable_for_ODF_clusters.py
@@ -519,17 +519,20 @@ class TestNfsEnable(ManageTest):
         9:- Verify able to read exported volume
         10:- Verify able to write to the exported volume from external client
         11:- Able to read updated /var/lib/www/html/index.html file from inside the pod
-        12:- unmount
-        13:- Deletion of Pods and PVCs
+        12:- Verify df on the external client reports PVC size, not cluster capacity
+        13:- unmount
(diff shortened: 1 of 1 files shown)


amr1ta:
verification job: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-deploy-ocs-cluster/69443/

AviadP (approved):

ebenahar (approved):

Issue #15043, opened 2026-05-03
test_mcg_namespace_basic_s3_ops_crd[AWS-OC-Single] and test_mcg_namespace_basic_s3_ops_crd[Azure-OC-Single] and test_mcg_namespace_basic_s3_ops_crd[AWS-OC-Cache]are failing in 4.19


ypersky1980:
test_mcg_namespace_basic_s3_ops_crd[AWS-OC-Single] 

Relevant runs: 

http://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/852/45601/2056927/2057140/log
https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/852/47196/2107413/2107613/log

Backtrace: 

self = <test_mcg_namespace_s3_ops_crd.TestMcgNamespaceS3OperationsCrd object at 0x7f99503071d0>
mcg_obj = <ocs_ci.ocs.resources.mcg.MCG object at 0x7f9957113c50>
cld_mgr = <ocs_ci.ocs.resources.cloud_manager.CloudManager object at 0x7f9954ae5290>
bucket_factory = <function bucket_factory_fixture.._create_buckets at 0x7f990f941300>
bucketclass_dict = {'interface': 'OC', 'namespace_policy_dict': {'namespacestore_dict': {'aws': [(1, None)]}, 'type': 'Single'}}

@pytest.mark.parametrize(
    argnames=["bucketclass_dict"],
    argvalues=[
        pytest.param(
            {
                "interface": "OC",
                "namespace_policy_dict": {
                    "type": "Single",
                    "namespacestore_dict": {"aws": [(1, None)]},
                },
            },
        ),
        pytest.param(
            {
                "interface": "OC",
                "namespace_policy_dict": {
                    "type": "Single",
                    "namespacestore_dict": {"azure": [(1, None)]},
                },
            },
        ),
        pytest.param(
            {
                "interface": "OC",
                "namespace_policy_dict": {
                    "type": "Single",
                    "namespacestore_dict": {"rgw": [(1, None)]},
                },
            },
            marks=on_prem_platform_required,
        ),
        pytest.param(
            {
                "interface": "OC",
                "namespace_policy_dict": {
                    "type": "Cache",
                    "ttl": 600000,
                    "namespacestore_dict": {
                        "aws": [(1, "eu-central-1")],
                    },
                },
                "placement_policy": {
[...]

PR #15482, opened 2026-06-23
[release-4.19] Fix s3_copy_object for when metadata is not provided

ypersky1980:
Cherry-pick of #12902 to release-4.19.

Fixes https://github.com/red-hat-storage/ocs-ci/issues/15043
  
The `metadata` parameter in `s3_copy_object` defaulted to `""` (empty string), but botocore's `copy_object` expects a `dict` for the `Metadata` kwarg. When tests call `s3_copy_object` without metadata, it fails with `ParamValidationError`. Fix changes the default to `None` and converts to `{}` before the API call.

### DIFF
--- a/ocs_ci/ocs/bucket_utils.py
+++ b/ocs_ci/ocs/bucket_utils.py
@@ -1814,7 +1814,7 @@ def compare_directory(
     return all(comparisons)
 
 
-def s3_copy_object(s3_obj, bucketname, source, object_key, metadata=""):
+def s3_copy_object(s3_obj, bucketname, source, object_key, metadata=None):
     """
     Boto3 client based copy object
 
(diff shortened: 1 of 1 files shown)


ypersky1980:
**NOTE FOR MERGING :** 
Merge target: release-4.19 only.
  4.20 and later already have a broader fix that replaced the metadata parameter with **kwargs, which passes arguments directly to boto3's copy_object.
  4.18 and earlier don't have the metadata parameter at all, so the bug doesn't exist there.
  The broken metadata="" default was only introduced in 4.19, making this the only affected branch.

sagihirshfeld (approved):

ocs_ci/ocs/bucket_utils.py
```diff
@@ -1814,7 +1814,7 @@ def compare_directory(
     return all(comparisons)
 
 
-def s3_copy_object(s3_obj, bucketname, source, object_key, metadata=""):
+def s3_copy_object(s3_obj, bucketname, source, object_key, metadata=None):
```
udaysk23, line 1817: How about passing kwargs here so that we dont need to worry about extra args in future?

udaysk23: Resolving it considering it is cherry pick

udaysk23 (approved):

PR #15494, opened 2026-06-24
Fix test_nsfs_list_buckets and add nsfs_only access test

sagihirshfeld:
## Summary
- The old `test_nsfs_list_buckets` expected cross-account bucket visibility, but since [DFBUGS-4449](https://redhat.atlassian.net/browse/DFBUGS-4449) ListBuckets is ownership-scoped - each account only sees its own buckets. Refactored to create two accounts with two buckets each, then verify each account lists exactly its own.
- Added `test_nsfs_only_account_access` - a dedicated test for the `nsfs_only` account flag, verifying it blocks access to non-NSFS buckets while still allowing cross-account NSFS bucket access.

## Changes
- **test_nsfs_list_buckets**: Replaced duplicated account setup with a loop, dropped the `nsfs_only` angle (it doesn't affect listing scope - visibility is ownership-based)
- **test_nsfs_only_account_access** (new): Creates one `nsfs_only=True` and one regular NSFS account, verifies the regular account can access a public non-NSFS bucket but the nsfs_only account gets AccessDenied, and verifies both can access each other's NSFS buckets
- Exposed `s3_resource` on the NSFS dataclass so tests can create buckets using an account's own credentials

### DIFF
--- a/tests/functional/object/mcg/test_nsfs.py
+++ b/tests/functional/object/mcg/test_nsfs.py
@@ -1,5 +1,6 @@
 import logging
 import types
+
 import pytest
 
 from ocs_ci.framework.testlib import MCGTest, tier1, tier3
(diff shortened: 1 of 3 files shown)


tests/functional/object/mcg/test_nsfs.py
```diff
@@ -283,50 +284,44 @@ def test_nsfs_list_objects(
...
-        Test NSFS bucket listing:
+        Test NSFS bucket listing with ownership-scoped ListBuckets:
 
-        1. Create two NSFS accounts - one with NSFS_ONLY=True and one with NSFS_ONLY=False
-        2. Create two NSFS buckets - one from each account
-        3. Verify that the NSFS_ONLY=True account can only see the NSFS buckets
-        4. Verify that the NSFS_ONLY=False account can also see first.bucket
+        1. Create two NSFS accounts with one bucket each
+        2. Create a second bucket from each account
+        3. Verify that each account can only list its own buckets
```
udaysk23, line 297: We can combine entire test with loop instead of writing it twice

sagihirshfeld: Fixed in https://github.com/red-hat-storage/ocs-ci/pull/15494/changes/ba77596f49bfe384bb225bf9c19b7d3bf4ced93d

tests/functional/object/mcg/test_nsfs.py
```diff
@@ -283,50 +284,36 @@ def test_nsfs_list_objects(
...
         """
-        # 1. Create two NSFS accounts - one with NSFS_ONLY=True and one with NSFS_ONLY=False
-        # 2. Create two NSFS buckets - one from each account
-        nsfs_obj_1 = NSFS(
```
ypersky1980, line 297: @sagihirshfeld  This line drops all nsfs_only coverage from the test suite. The old code created one account with nsfs_only=True and one with nsfs_only=False — this was the only test in the repo that ever
exercised nsfs_only=True (confirmed via code search — the flag only appears in mcg_params.py and this file).

  I understand that nsfs_only restricts access to non-NSFS bucket types rather than affecting listing scope, so it's orthogonal to what this test now verifies (ownership-scoped ListBuckets).
  But by removing it here without a replacement, we lose our only coverage for the nsfs_only=True account behavior entirely.

  Could you either:
  1. Add a separate focused test (even a minimal one) that creates an nsfs_only=True account and verifies it can't access non-NSFS bucket types, or
  2. Keep one of the two accounts in this test as nsfs_only=True — since the ownership-scoped listing logic shouldn't depend on that flag, it would work the same while preserving coverage as a
  side effect

  If nsfs_only is intentionally being deprecated or is already covered by a NooBaa-side test, a brief note in the PR description would help.

ypersky1980 (commented):
Left a comment.

tests/functional/object/mcg/test_nsfs.py
sagihirshfeld: Good catch. You're right that `nsfs_only` coverage was lost here.

I looked into what `nsfs_only=True` actually does in noobaa-core - it blocks the account from accessing non-NSFS buckets (any S3 operation on a non-NSFS bucket returns AccessDenied). It does NOT affect ListBuckets filtering though - bucket visibility is purely ownership-scoped, so an nsfs_only account won't see non-NSFS buckets in listing simply because it doesn't own any, not because of the flag.

Rather than keeping it as a side-effect in this test (where all buckets are NSFS anyway, so the flag wouldn't change anything), I added a dedicated test (`test_nsfs_only_account_access`, d638829cfed13784714da4219a663c4de49fbf4c) that exercises the real distinction:

1. Creates one `nsfs_only=True` account and one regular NSFS account
2. Creates an admin-owned non-NSFS bucket with a public policy
3. Verifies the regular account can access it but the nsfs_only account gets AccessDenied
4. Verifies both accounts can still access each other's NSFS buckets

ypersky1980 (approved):
LGTM

udaysk23 (approved):

PR #15511, opened 2026-06-25
Adds medium and low tests for validating RGW STS feature and optimised code to avoid duplication

udaysk23:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RGW STS S3 object handling: large uploads now automatically use multipart upload, and large downloads now use ranged reads to retrieve data in chunks.
  * Standardized AWS/STS error assertions for consistent validation of invalid credentials and authorization failures.

* **Tests**
  * Refactored RGW STS scenarios to use reusable IAM/STS/S3 helper methods, reducing duplicated setup and simplifying failure checks.
  * Added end-to-end tests for inline role policy enforcement and a comprehensive S3 operation matrix, including multipart and mixed-credential download/delete flows.


tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -207,16 +208,47 @@ def put_object(self, s3_client, bucket_name, object_key, data):
...
+            part_number += 1
+        s3_client.complete_multipart_upload(
+            Bucket=bucket_name,
+            Key=object_key,
+            UploadId=upload_id,
+            MultipartUpload={"Parts": parts},
+        )
+        logger.info(
+            f"Object {object_key} uploaded via multipart ({part_number - 1} parts)"
+        )
```
ypersky1980, line 254: 1. No error handling: If upload_part or complete_multipart_upload fails, the upload ID is leaked on the RGW server — no abort_multipart_upload is called. This can consume storage and block
  bucket deletion in teardown. Please wrap the upload loop in try/except and call s3_client.abort_multipart_upload(Bucket=bucket_name, Key=object_key, UploadId=upload_id) in the except block
  before re-raising.
  2. Missing return value: The multipart path returns None while the simple upload path returns the response dict. complete_multipart_upload returns a response dict — please return it so the
  API is consistent (return s3_client.complete_multipart_upload(...)).

tests/functional/object/rgw/test_rgw_sts.py
```diff
@@ -877,4 +876,336 @@ def _is_token_expired():
...
+        access_key = rgw_account_and_user["access_key"]
+        secret_key = rgw_account_and_user["secret_key"]
+
+        # Step 1: Create role with full S3 permissions and assume role
+        logger.info("Step 1: Creating role with full S3 permissions and assuming role")
+        std_s3_client = self.make_boto3_client(
+            "s3", endpoint, region, access_key, secret_key
+        )
+        sts_client = self.make_boto3_client(
+            "sts", endpoint, region, access_key, secret_key
```
ypersky1980, line 1105: Step 3 uses raw s3_client.get_object(Bucket=..., Key=...)["Body"].read() instead of self.get_object(), which was specifically enhanced earlier in this PR to handle ranged GET for large objects. If the test ever uses data larger than 5 MiB, this raw call will work but bypass the chunked download logic, making that enhancement dead code in this test path. Please use  self.get_object() here for consistency.

ypersky1980 (commented):
Left 2 comments.

tests/functional/object/rgw/test_rgw_sts.py
ypersky1980: @udaysk23 do you plan to fix it?

sagihirshfeld (dismissed):

tests/functional/object/rgw/test_rgw_sts.py
udaysk23: Steps 4/11 objects are small test literals (<100 bytes), so self.get_object() would take the same single-request path as the raw call — no functional difference. Those steps also check the full HTTP response metadata, which the helper doesn't expose. The ranged GET path is exercised in the multipart sections (steps with 10 MiB objects) which already call self.get_object().

ypersky1980 (dismissed):
LGTM

sagihirshfeld (approved):

ypersky1980 (approved):
LGTM

PR #15516, opened 2026-06-26
correct the parametrizations

vkathole:
Problem: EC cases used pytest.mark.skipif(not is_ec_pool_supported(), ...), so on clusters without EC support those parametrizations were still collected and reported as SKIPPED, even though the replicated-pool cases (erasure_coded=False) should run.

Fix: Added get_erasure_coded_rbd_sc_params() in ocs_ci/ocs/cluster.py. It returns EC pytest.param entries only when is_ec_pool_supported() is true; otherwise it returns an empty list.

On a non-EC cluster, pytest tests/functional/workloads/test_new_sc_rbd_e2e_workloads.py --collect-only should show 4 collected tests. On an EC-capable cluster it should show 6.

## Summary by CodeRabbit

* **New Features**
  * Expanded test coverage for erasure-coded RBD storage class scenarios in end-to-end and snapshot/clone workflows.
  * Added a shared way to generate parameter sets for storage class test variations.

* **Refactor**
  * Simplified test parameter setup by moving repeated configuration into shared reusable logic.
  * Reduced inline test markers and conditional setup, making the affected tests easier to maintain.

### DIFF
--- a/ocs_ci/ocs/cluster.py
+++ b/ocs_ci/ocs/cluster.py
@@ -2439,6 +2439,45 @@ def is_ec_pool_supported():
     return True
 
 
+def get_erasure_coded_rbd_sc_params(param_configs):
+    """
+    Build pytest.param entries for erasure-coded RBD storage class test cases.
+
+    Returns an empty list when EC pools are not supported on the cluster,
+    so non-EC clusters run only the replicated-pool parametrizations without
+    skipped EC variants.
+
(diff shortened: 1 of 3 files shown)


vkathole:
Closing as the "test never passed" status for EC tests on non-EC clusters is expected until the dedicated EC pipeline
 runs them, which is being prepared before 4.22 GA.

PR #15518, opened 2026-06-28
Add test for bound CephFS snapshot delete rejection (RHSTOR-7731)

yitzhak12:
Add `test_cephfs_bound_snapshot_delete_rejected` to verify that the odf CLI rejects deletion of a Bound CephFS snapshot. The test reuses the existing setup fixture (Retain snapclass, bound and orphaned snapshots) and asserts the CLI fails or leaves the snapshot intact, then confirms the snapshot remains in Bound state via a short TimeoutSampler poll.

Extract two helpers into `odf_cephfs_snap.py`:
- `verify_bound_snapshot_delete_rejected`: tries the CLI delete, handles both CommandFailed (expected path) and silent no-op (fallback check).
- `wait_and_verify_snapshot_bound`: polls with TimeoutSampler then does a final explicit Bound-state assertion.

Fix `find_consumer_for_storage_client` and `get_consumer_svg_on_provider` in `storageconsumer.py` to accept an optional `client_cluster_name` parameter. When multiple consumers share the same storage-client name, the caller passes `config.ENV_DATA["cluster_name"]` (captured before switching to the provider context) so the correct consumer is found by matching the prefix of `status.client.clusterName`.

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end negative test coverage ensuring deletion attempts for **Bound** CephFS snapshots are correctly rejected and the snapshot set remains unchanged.
  * Improved multicluster SVG resolution by using the client cluster context to select the correct provider mapping.

* **Bug Fixes**
  * Strengthened verification that **Bound** CephFS snapshot deletions are rejected, including more robust CLI output matching.
  * Added polling-based confirmation that the target snapshot remains in **Bound** state after the failed delete attempt.

### DIFF
--- a/ocs_ci/helpers/odf_cephfs_snap.py
+++ b/ocs_ci/helpers/odf_cephfs_snap.py
@@ -320,12 +320,14 @@ def wait_and_verify_snapshot_bound(snap_runner, snap_data, timeout=30, sleep=5):
         sleep (int): Seconds between polls (default 5).
 
     Raises:
-        AssertionError: If the snapshot disappears or transitions out of
-            Bound state at any point during the sampling window.
+        AssertionError: If the snapshot disappears or is not in Bound state
+            when first observed after the delete attempt.
+        TimeoutExpiredError: If ``get_cephfs_snap_entries`` keeps failing
+            for the entire ``timeout`` window.
     """
     ceph_snap_name = snap_data["ceph_snap_name"]
     log.info(
-        "Polling to confirm snapshot '%s' remains in Bound state",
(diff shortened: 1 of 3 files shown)


yitzhak12:
The PR validation passed with provider-client: https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-odf-provider-client-multicluster/1652/consoleFull

DanielOsypenko (dismissed):

ebenahar (dismissed):

DanielOsypenko (approved):

ebenahar (approved):

PR #15519, opened 2026-06-28
resolve client profile mapping for non-default storage ns

DanielOsypenko:
Instead of assuming the ClientProfile name equals the namespace, fetch the actual ClientProfile name from the cluster
Assumption is `ConfigMap value == ClientProfile name`, which happens to be true on problematic cluster, based on MG logs

Fix the problem, seen in run 21678, 21715
```
2026-06-20 13:09:48  >       assert all(ceph_data_on_consumer_match.values()), (
2026-06-20 13:09:48              f"StorageConsumer config map data does not match expected values. Consumer name: {consumer_name};"
2026-06-20 13:09:48              f"ODF Operator upgraded: {is_cluster_y_version_upgraded()}; "
2026-06-20 13:09:48          )
2026-06-20 13:09:48  E       AssertionError: StorageConsumer config map data does not match expected values. Consumer name: internal;ODF Operator upgraded:
```
https://url.corp.redhat.com/83f4224

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved storage configuration verification by deriving the CSI ClientProfile stable name dynamically from existing `CLIENT_PROFILE` entries, with a fallback when none are found.
  * Updated internal-client post-convergence checks to use the derived ClientProfile name consistently, including when CephFS is disabled.
  * Fixed IBM Cloud COS secret generation by ensuring the secret manifest includes the correct cluster namespace metadata.

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -1735,6 +1735,9 @@ class Deployment(object):
                 key_secret = config.AUTH["ibmcloud"]["ibm_cos_secret_access_key"]
                 cos_secret_data["data"]["IBM_COS_ACCESS_KEY_ID"] = key_id
                 cos_secret_data["data"]["IBM_COS_SECRET_ACCESS_KEY"] = key_secret
+                cos_secret_data["metadata"]["namespace"] = config.ENV_DATA[
+                    "cluster_namespace"
+                ]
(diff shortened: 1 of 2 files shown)


DanielOsypenko:
verification run on IBM cloud and non-standard ns:
https://url.corp.redhat.com/83a8cf7

vavuthu (approved):

petr-balogh (approved):

PR #15521, opened 2026-06-29
Implement RDR-Provider Upgrade scenario

shylesh:
## Summary by CodeRabbit

* **New Features**
  * Added support for a new multicluster RDR provider mode (“rdr-provider”).
  * Introduced provider-mode role handling, including HostedClient role/rank support and hosted-client topology coverage in upgrade scenarios.
* **Bug Fixes**
  * Improved upgrade sequencing and zone grouping to keep provider and hosted-client scheduling consistent across runs.
  * Enabled multicluster upgrade parametrization for provider-client platforms when running in “rdr-provider” mode.
* **Chores**
  * Updated multicluster configuration and extended upgrade test scope with additional multicluster role coverage.

### DIFF
--- a/ocs_ci/deployment/hub_spoke.py
+++ b/ocs_ci/deployment/hub_spoke.py
@@ -1464,7 +1464,7 @@ class HostedClients(HyperShiftBase):
             bool: True if all hosted clusters passed verification, False otherwise
 
         """
-        cluster_names = list(config.ENV_DATA.get("clusters").keys())
+        cluster_names = list((config.ENV_DATA.get("clusters") or {}).keys())
         cluster_names = [
             name
             for name in cluster_names
@@ -1652,7 +1652,7 @@ class HostedClients(HyperShiftBase):
                 f'--patch \'{{"data":{{"ack-4.19-admissionregistration-v1beta1-api-removals-in-4.20":"true"}}}}\' '
             )
 
-        cluster_names = list(config.ENV_DATA.get("clusters").keys())
+        cluster_names = list((config.ENV_DATA.get("clusters") or {}).keys())
         cluster_names = [
             name
             for name in cluster_names
--- a/ocs_ci/ocs/provider_client_upgrade.py
+++ b/ocs_ci/ocs/provider_client_upgrade.py
@@ -150,7 +150,13 @@ class OperatorUpgrade(ProviderUpgrade):
         )
 
         if not cluster_names:
-            cluster_names = list(config.ENV_DATA.get("clusters").keys())
+            cluster_names = list((config.ENV_DATA.get("clusters") or {}).keys())
+        if not cluster_names:
+            from ocs_ci.deployment.helpers.hypershift_base import (
+                get_hosted_cluster_names,
+            )
+
+            cluster_names = get_hosted_cluster_names()
 
         for cluster_name in cluster_names:
             log.info(
--- a/ocs_ci/utility/multicluster.py
+++ b/ocs_ci/utility/multicluster.py
@@ -16,7 +16,14 @@ from ocs_ci.ocs.utils import (
     get_active_acm_index,
     get_all_acm_indexes,
 )
-from ocs_ci.ocs.constants import MDR_ROLES, RDR_ROLES, ACM_RANK, MANAGED_CLUSTER_RANK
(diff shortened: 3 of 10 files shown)


ocs_ci/ocs/provider_client_upgrade.py
```diff
@@ -198,20 +274,78 @@ def __init__(self):
...
             log.info("Starting the operator upgrade process...")
             operator_upgrade = OperatorUpgrade()
-            # delete old idms, if exist from the previous upgrades
-            prune_old_df_repo_idms(force_delete_pods=True)
-            # Bump OCS version on clients. This function will not fail upgrade if bump of any client fails
-            operator_upgrade.bump_ocs_version_on_clients()
-            ocs_upgrade.run_ocs_upgrade()
+            is_fdf = _is_fdf_upgrade()
+
+            if is_fdf:
```
shylesh: @clacroix12 suggested to move these if/else into seperate functions to keep run_provider_upgrade() clean.

ocs_ci/deployment/hub_spoke.py
```diff
@@ -2440,6 +2459,62 @@ def patch_nodepool_for_ocp_upgrade(
             logger.error(f"Failed to patch NodePool '{self.name}' for OCP upgrade: {e}")
             return False
 
+    def patch_nodepool_root_volume(self, size):
```
clacroix12, line 2462: This function returns a boolean based on whether the patch was successful or not, but we don't do anything with that boolean on line 1703 where it's called. It seems like we will just be logging the error and moving on.

If this is something we should be attempting again perhaps it would be better to have this function raise an error (instead of returning false) when it fails and decorate it with a retry to attempt the patch again?

ocs_ci/ocs/provider_client_upgrade.py
```diff
@@ -139,41 +158,98 @@ def run_operators_upgrade(self):
...
+            log.info(f"Bumping catalog on hosted OCP cluster '{cluster_name}'")
             try:
-                hosted_odf = HostedODF(cluster_name)
-                if not hosted_odf.odf_client_installed():
-                    log.info(
-                        f"ODF client operator not installed on HCP cluster '{cluster_name}', skipping this client"
+                if is_fdf:
+                    hosted_odf = HostedFDF(cluster_name)
+                    if not hosted_odf.odf_client_installed():
+                        log.info(
```
clacroix12, line 191: This probably doesn't need to be logged at INFO. DEBUG would probably suffice. I would rather see confirmation of which catalog source is created instead of the one we skip (at info level).

ocs_ci/ocs/provider_client_upgrade.py
```diff
@@ -198,22 +274,85 @@ def __init__(self):
...
+               image for offline racks)
+            4. Propagate IDMS mirror config to hosted clusters
+            5. Verify FDF auto-upgraded on all client clusters
+        """
+        from ocs_ci.deployment.fusion_data_foundation import (
+            FusionDataFoundationDeployment,
+        )
+        from ocs_ci.ocs.fdf_upgrade import FDFUpgrade
+
+        # Step 1: Upgrade Fusion + FDF operator on the provider.
```
clacroix12, line 314: Consider using `log.test_step` to outline these so we get the same clarity in the logs.

ocs_ci/utility/multicluster.py
```diff
@@ -303,9 +311,254 @@ def reeval_upgrade_order(self, phase_order, zrank, role_rank):
...
+
+    def __init__(self):
+        self.dr_type = "rdr-provider"
+        super().__init__()
+        # Start from the static role list; dynamic entries are appended below
+        self.all_roles = list(RDR_PROVIDER_ROLES)
+
+        # Optional secondary ACM
+        if get_passive_acm_index():
+            self.all_roles.append("PassiveACM")
```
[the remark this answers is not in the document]
clacroix12: @shylesh please take a look at this issue and the proposed fix. If we are unintentionally missing the 0 index here, we should implement the proposed fix. If we are intentionally skipping the 0 index, we should probably still retrieve the index in the same way but change the if statement to reflect that we want the value to be `> 0`. and make a comment why.

PR #15553, opened 2026-06-30
use RHCOS in OCP 5.0

vavuthu:

## Summary by CodeRabbit

* **New Features**
  * Updated the OCP 5.0 nightly environment to use the latest RHCOS 10 virtual machine template.
  * Improved VM template selection to support multiple RHCOS major versions via a version-keyed template mapping, with a corrected legacy fallback.
* **Documentation**
  * Added in-config guidance indicating how to select the RHCOS 10 template using the appropriate RHCOS version setting.

### DIFF
--- a/ocs_ci/framework/conf/ocp_version/ocp-5.0-config.yaml
+++ b/ocs_ci/framework/conf/ocp_version/ocp-5.0-config.yaml
@@ -9,7 +9,13 @@ DEPLOYMENT:
   # https://docs.openshift.com/container-platform/4.7/post_installation_configuration/machine-configuration-tasks.html#machine-config-overview-post-install-machine-configuration-tasks
   ignition_version: "3.2.0" # TODO: We might want to update to 3.5 or later: https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/release_notes/ocp-4-19-release-notes
 ENV_DATA:
-  vm_template: "scos-10.0.20251103-0-vmware.x86_64"
+  # Dictionary of available RHCOS templates by major version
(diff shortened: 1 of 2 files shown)


ocs_ci/framework/conf/ocp_version/ocp-5.0-config.yaml
```diff
@@ -9,7 +9,14 @@ DEPLOYMENT:
...
 ENV_DATA:
-  vm_template: "scos-10.0.20251103-0-vmware.x86_64"
+  # Dictionary of available RHCOS templates by major version
+  # To use RHCOS 10, set rhcos_version: "10" in your custom config
+  vm_templates:
+    "9": "rhcos-9.8.20260428-0-vmware.x86_64"
+    "10": "rhcos-10.2.20260423-0-vmware.x86_64"
+  # Legacy single template (fallback if vm_templates not defined)
+  vm_template: "rhcos-10.2.20260423-0-vmware.x86_64"
+  rhcos_version: "10"
```
petr-balogh: I am bit afraid to set this value here with the way how we reload OCP config file based on ODF version during the process we load configs.  So even though we will pass custom config file with:
```
rhcos_version: "9"
```

I think it will get overwritten by 10 once OCP config file is reloaded.

So we need to verify this works, or we will need to find other way to define maybe here 
```
default_rhcos_version: "10"
```

And this case check first if user provided value of `rhcos_version` is passed - and if not - we will be using `default_rhcos_version`

vavuthu: removed

petr-balogh (approved):

dahorak (approved):

PR #15476, opened 2026-06-22
Add operator restart PriorityClass persistence test (RHSTOR-7764)

sagihirshfeld:
## Summary
- Add `test_priority_class_persists_after_operator_restart` to verify that
  PriorityClassName settings on the NooBaa CR survive an operator pod restart
  without triggering unnecessary pod reconciliation

## Changes
- New test patches the NooBaa CR with per-component PriorityClassNames,
  verifies propagation, then restarts the operator pod and confirms that
  all priority classes persist and no pods were unnecessarily recreated
  (by comparing creation timestamps before and after the restart)
- Adds `request.addfinalizer` cleanup to revert the NooBaa CR on teardown

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for NooBaa operator stability, validating that component priority class configurations persist correctly across operator restarts and that system pods remain unaffected during operator maintenance operations.

PR #15561, opened 2026-07-01
Add multi-target and pod-restart tests for replication alert (RHSTOR-8110)

sagihirshfeld:
## Summary
- Add a test for the `NooBaaReplicationTargetUnreachable` alert with multiple source and target
  buckets (OCS-8034, tier3), verifying per-source-per-target alert granularity
- Add a test for alert persistence across a noobaa-core pod restart (OCS-8035, tier4c), verifying
  the alert resumes firing after the scanner re-detects the unreachable target

## Changes
- `tests/.../test_mcg_replication_target_unreachable.py` - two new test methods:
  `test_noobaa_replication_target_unreachable_multi_source_multi_target` and
  `test_noobaa_replication_target_unreachable_core_pod_restart`; new imports for
  `update_replication_policy`, `get_noobaa_core_pod`, `wait_for_noobaa_pods_running`,
  `tier3`, `tier4c`

## Summary by CodeRabbit

* **Tests**
  * Expanded replication monitoring coverage for multiple source and target buckets.
  * Added validation that unreachable-target alerts identify the correct buckets and remain isolated to affected targets.
  * Added coverage for alert recovery after restarting the NooBaa core service.
  * Added verification that replication resumes and alerts clear after restoring unavailable targets.
  * Improved validation of expected objects during bucket replication checks.

### DIFF
--- a/tests/functional/object/mcg/test_mcg_replication_target_unreachable.py
+++ b/tests/functional/object/mcg/test_mcg_replication_target_unreachable.py
@@ -13,6 +13,7 @@ from ocs_ci.ocs import constants
 from ocs_ci.ocs.bucket_utils import (
     compare_bucket_object_list,
     update_replication_policy,
+    wait_for_expected_objects_in_bucket,
     write_random_test_objects_to_bucket,
 )
 from ocs_ci.ocs.exceptions import CommandFailed
@@ -462,13 +463,8 @@ class TestMCGReplicationTargetUnreachableAlert(MCGTest):
                 for label, t in zip(("a", "b", "c"), target_buckets)
             ]
         }
-        initial_replication_policy = ("rule-a", target_a.name, None)
-        source_a = bucket_factory(
-            1, "OC", replication_policy=initial_replication_policy
-        )[0]
-        source_b = bucket_factory(
-            1, "OC", replication_policy=initial_replication_policy
-        )[0]
+        source_a = bucket_factory(1, "OC")[0]
+        source_b = bucket_factory(1, "OC")[0]
         for src in (source_a, source_b):
             update_replication_policy(src.name, multi_rule_policy)
             logger.info(
@@ -480,7 +476,7 @@ class TestMCGReplicationTargetUnreachableAlert(MCGTest):
         logger.test_step(
             "Write objects and verify replication from both sources to all targets"
         )
-        write_random_test_objects_to_bucket(
+        src_a_objects = write_random_test_objects_to_bucket(
             awscli_pod_session,
             source_a.name,
             f"{test_directory_setup.origin_dir}/src_a_init",
@@ -488,7 +484,7 @@ class TestMCGReplicationTargetUnreachableAlert(MCGTest):
             pattern="src-a-init-",
             mcg_obj=mcg_obj,
         )
-        write_random_test_objects_to_bucket(
+        src_b_objects = write_random_test_objects_to_bucket(
             awscli_pod_session,
             source_b.name,
             f"{test_directory_setup.origin_dir}/src_b_init",
@@ -496,12 +492,15 @@ class TestMCGReplicationTargetUnreachableAlert(MCGTest):
             pattern="src-b-init-",
             mcg_obj=mcg_obj,
         )
-        for src in (source_a, source_b):
+        for src_name, obj_names in (
+            (source_a.name, src_a_objects),
+            (source_b.name, src_b_objects),
+        ):
             for tgt in target_buckets:
-                logger.assertion(f"Replication check: {src.name} -> {tgt.name}")
-                assert compare_bucket_object_list(
-                    mcg_obj, src.name, tgt.name, timeout=600
-                ), f"Replication failed: {src.name} -> {tgt.name}"
+                logger.assertion(f"Replication check: {src_name} -> {tgt.name}")
+                assert wait_for_expected_objects_in_bucket(
+                    mcg_obj, obj_names, tgt.name, timeout=600
+                ), f"Replication failed: {src_name} -> {tgt.name}"
         logger.info("Initial replication verified for all source-target pairs")
 
         # 4. Delete the underlying MCG bucket of targetA's backingstore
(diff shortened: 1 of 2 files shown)


tests/functional/object/mcg/test_mcg_replication_target_unreachable.py
```diff
@@ -394,3 +396,398 @@ def test_noobaa_replication_target_unreachable_underlying_bucket_deleted(
...
+                    mcg_obj, src.name, tgt.name, timeout=600
+                ), f"Replication not recovered: {src.name} -> {tgt.name}"
+        logger.info("Replication recovered for all source-target pairs")
+
+        # 13. Wait for all alerts to clear
+        logger.test_step("Wait for all alerts to clear")
+        for src in (source_a, source_b):
+            _wait_for_replication_alert(
+                threading_lock, src.name, timeout=600, sleep=10, cleared=True
+            )
```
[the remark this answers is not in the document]
sagihirshfeld: Skipping this suggestion. The two sequences aren't actually identical - the healthy-target verification (step 7) only happens after the first deletion, not the second. Extracting a helper would require passing 8+ parameters (target, underlying name, sources, healthy targets list, awscli pod, test directory, mcg_obj, threading_lock, pattern suffix) and would obscure the test's narrative progression: first delete A and verify B/C are unaffected, then delete B. The test is telling a deliberate story, not running the same operation twice.

PLR0915 isn't enforced in this repo (pre-commit hooks passed), and the statement count is inherent to a 13-step integration test.

tests/functional/object/mcg/test_mcg_replication_target_unreachable.py
```diff
@@ -394,3 +397,404 @@ def test_noobaa_replication_target_unreachable_underlying_bucket_deleted(
...
+
+        # 4. Wait for the alert to fire
+        logger.test_step("Wait for the alert to fire")
+        _wait_for_replication_alert(
+            threading_lock, source_bucket.name, timeout=600, sleep=10
+        )
+
+        # 5. Restart the noobaa-core pod and wait for it to become ready
+        logger.test_step("Restart the noobaa-core pod and wait for it to become ready")
+        get_noobaa_core_pod().delete()
```
ypersky1980: delete() without `wait=True` races with `wait_for_noobaa_pods_running`. 

  `delete()` without `wait=True` sends the DELETE request and returns immediately. The old pod enters `Terminating` state but **still has `phase: Running`** in the Kubernetes API —`Terminating` is a kubectl display state, not an API phase.

  `wait_for_noobaa_pods_running` then calls `get_pods_having_label(pod_label, statuses=[STATUS_RUNNING])`, sees the old pod (still phase=Running), and returns success immediately. Step 5
  effectively becomes a no-op:

  1. `delete()` → sends DELETE, returns instantly
  2. Old pod still Running (Terminating but phase hasn't changed yet)
  3. `wait_for_noobaa_pods_running` → sees old pod Running → returns True
  4. Test continues to Step 6 while old pod is still shutting down
  5. Old pod terminates, new pod starts — but test is already polling for alerts

  The test still passes because Step 6 has a generous 600s timeout that absorbs the pod restart time. But the log message "NooBaa core pod is ready" on line 736 is misleading — it's reporting  on the OLD pod.

  **Suggested fix:**

  get_noobaa_core_pod().delete(wait=True)

  This ensures the old pod is fully gone before wait_for_noobaa_pods_running starts checking, so it waits for the new pod. This also matches the pattern used in other NooBaa tests (e.g., PR  #15476 uses operator_pod.delete(wait=True)).

tests/functional/object/mcg/test_mcg_replication_target_unreachable.py
```diff
@@ -394,3 +397,404 @@ def test_noobaa_replication_target_unreachable_underlying_bucket_deleted(
...
+
+        # 10. Wait for two more alerts for targetB
+        logger.test_step(
+            "Wait for two more alerts to fire for targetB (one per source)"
+        )
+        _wait_for_replication_alert(
+            threading_lock, source_a.name, target_b.name, timeout=600, sleep=10
+        )
+        _wait_for_replication_alert(
+            threading_lock, source_b.name, target_b.name, timeout=600, sleep=10
```
ypersky1980, line 594: Missing targetC health verification after targetB deletion; 

  Step 7 (lines 540-557) correctly verifies that targetB and targetC have no alerts after breaking targetA. But after breaking targetB and confirming its alerts fire here in Step 10, there's no equivalent check that targetC is still healthy.

  This is a gap for a test specifically designed to verify "per-source-per-target alert granularity." If deleting targetB's underlying bucket causes a side effect that triggers a spurious alert  for targetC (e.g., shared backing infrastructure, NooBaa scanner bug treating all remaining targets as suspect), the test won't catch it.

  Suggested fix — add after this line, before Step 11:

  # Verify targetC is still healthy after targetB deletion
  logger.test_step(
      "Verify no alerts are firing for healthy target (targetC)"
  )
  alerts = api.get_alerts_by_labels(
      alert_name=alert_name,
      labels_dict={"target_bucket": target_c.name},
  )
  logger.assertion(
      f"No alerts for healthy target {target_c.name}: "
      f"found={len(alerts)}, expected=0"
  )
  assert (
      not alerts
  ), f"Unexpected alert firing for healthy target {target_c.name} after targetB deletion: {alerts}"

 This mirrors the Step 7 check but specifically after the second disruption, closing the verification gap.

ypersky1980 (commented):
Left 2 comments.

tests/functional/object/mcg/test_mcg_replication_target_unreachable.py
sagihirshfeld: Fixed in bd8b1bd0836abd7f5cd1e8a2b14c831bca775285

tests/functional/object/mcg/test_mcg_replication_target_unreachable.py
sagihirshfeld: Fixed in bd8b1bd0836abd7f5cd1e8a2b14c831bca775285

ypersky1980 (approved):
LGTM

udaysk23 (approved):

PR #15562, opened 2026-07-01
Add scale test for multiple STS backingstores and namespacestores

sagihirshfeld:
## Summary
- Add `test_scale_sts_stores` to validate that multiple STS backingstores and namespacestores can coexist on the same cluster ([RHSTOR-5085](https://redhat.atlassian.net/browse/RHSTOR-5085))
- Parametrized for both AWS STS and Azure STS platforms, gated by platform markers
- Creates 3 backingstores + 3 namespacestores via CLI, then verifies IO through buckets backed by them

## Changes
- New test `test_scale_sts_stores` in `tests/functional/object/mcg/test_sts_bucket.py` with two parametrized variants:
  - `AWS-STS` (tier3, `aws_platform_required`, OCS-8039)
  - `AZURE-STS` (tier3, `azure_platform_required`, OCS-8038)
- Moved `polarion_id("OCS-5479")` from function-level decorator to the AWS-STS pytest.param marks on `test_sts_bucket_ops` for consistency

## Summary by CodeRabbit

* **Tests**
  * Expanded end-to-end STS bucket coverage to include both AWS and Azure scenarios with improved parametrization and clearer per-case identifiers.
  * Added a stronger scale test that creates multiple STS backingstores and namespacestores, uploads random objects, performs sync/download validation, and verifies data integrity.
  * After deleting objects, now explicitly confirms buckets are empty via post-deletion listing to ensure cleanup is fully effective.

### DIFF
--- a/tests/functional/object/mcg/test_sts_bucket.py
+++ b/tests/functional/object/mcg/test_sts_bucket.py
@@ -5,6 +5,7 @@ import pytest
 from ocs_ci.framework.pytest_customization.marks import (
     tier1,
     tier2,
+    tier3,
     sts_deployment_required,
     aws_platform_required,
     azure_platform_required,
@@ -15,6 +16,7 @@ from ocs_ci.framework.pytest_customization.marks import (
 from ocs_ci.ocs.bucket_utils import (
     write_random_test_objects_to_bucket,
     s3_delete_objects,
+    s3_list_objects_v2,
     sync_object_directory,
     compare_directory,
     check_if_objects_expired,
@@ -27,7 +29,6 @@ logger = logging.getLogger(__name__)
 @red_squad
 @sts_deployment_required
 class TestSTSBucket:
-    @polarion_id("OCS-5479")
     @pytest.mark.parametrize(
         argnames=["bucketclass"],
         argvalues=[
@@ -38,7 +39,7 @@ class TestSTSBucket:
                         "backingstore_dict": {"aws-sts": [(1, "eu-central-1")]},
                     },
                 ],
-                marks=[tier2, aws_platform_required],
+                marks=[tier2, aws_platform_required, polarion_id("OCS-5479")],
             ),
             pytest.param(
                 *[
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_sts_bucket.py
```diff
@@ -134,3 +135,117 @@ def test_sts_bucket_ops(
...
+                "namespace_policy_dict": {
+                    "type": "Single",
+                    "namespacestores": [namespacestores[0]],
+                },
+            },
+        ]
+        bucketnames = [
+            bucket_factory(bucketclass=bc)[0].name for bc in bucketclass_dicts
+        ]
+        logger.info(f"Created buckets: {bucketnames}")
```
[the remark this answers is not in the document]
sagihirshfeld: Actually, fair point - the intent was to create a bucketclass and OBC for each store (6 total), not just 2. Fixed to iterate over all backingstores and namespacestores.

tests/functional/object/mcg/test_sts_bucket.py
```diff
@@ -134,3 +135,117 @@ def test_sts_bucket_ops(
...
+            remaining = s3_list_objects_v2(
+                s3_obj=mcg_obj_session, bucketname=bucketname
+            )
+            logger.assertion(
+                f"Bucket emptied after delete: bucket='{bucketname}', "
+                f"remaining_objects={remaining.get('KeyCount', 0)}"
+            )
+            assert (
+                remaining.get("KeyCount", 0) == 0
+            ), f"Bucket {bucketname} still has objects after deletion"
```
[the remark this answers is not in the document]
sagihirshfeld: `check_if_objects_expired` does not have any polling or timeout - it is a single-shot `s3_list_objects_v2` + `KeyCount` check, identical to what is already here. Swapping it in would be cosmetic, not a stability improvement.

Also, `DeleteObjects` against NooBaa is synchronous - there is no eventual consistency concern (this is not a cloud provider endpoint with lazy replication). The first test uses `check_if_objects_expired` because it is verifying lifecycle-based expiration, which is semantically different from an explicit delete.

tests/functional/object/mcg/test_sts_bucket.py
ypersky1980, line 66: STS-DEFAULT variant of test_sts_bucket_ops lost its Polarion ID

The PR correctly moves @polarion_id("OCS-5479") from function-level to the AWS-STS param marks. The Azure variants already had their own IDs (OCS-7949, OCS-7950) so they're fine. But the 4th
  variant — STS-DEFAULT — only has marks=[tier1] and no polarion_id:

  pytest.param(
      *[None],
      marks=[
          tier1,
      ],
  ),

Previously it inherited OCS-5479 from the function-level decorator. Now it has no Polarion tracking at all. Either add a dedicated Polarion ID for this variant or confirm it intentionally doesn't need one.

tests/functional/object/mcg/test_sts_bucket.py
```diff
@@ -134,3 +135,119 @@ def test_sts_bucket_ops(
...
+        4. Upload objects to all buckets, download, and verify data integrity
+        """
+        # 1. Create 3 STS backingstores via CLI
+        logger.test_step(f"Create 3 {platform} backingstores via CLI")
+        backingstores = backingstore_factory("cli", {platform: [(3, region)]})
+
+        # 2. Create 3 STS namespacestores via CLI
+        logger.test_step(f"Create 3 {platform} namespacestores via CLI")
+        namespacestores = namespace_store_factory("cli", {platform: [(3, region)]})
+
```
ypersky1980, line 177: No assertion that factories returned the expected 3 stores each
The test's purpose (RHSTOR-5085) is validating that multiple STS stores can coexist. But the store count is never verified:

  backingstores = backingstore_factory("cli", {platform: [(3, region)]})
  # missing: assert len(backingstores) == 3

  namespacestores = namespace_store_factory("cli", {platform: [(3, region)]})
  # missing: assert len(namespacestores) == 3

If the factory silently returns fewer stores (due to a partial failure or factory bug), the test passes with fewer stores than intended — undermining the scale coexistence validation.

ypersky1980 (commented):
Left 2 comments.

tests/functional/object/mcg/test_sts_bucket.py
sagihirshfeld: Fixed in 68c80a7642bf654b2684acb2b2eb332104e2e3af

tests/functional/object/mcg/test_sts_bucket.py
sagihirshfeld: The factory can't silently return fewer stores - if any of the 3 fails, the test errors out before reaching the assertions. Here's the chain:

1. `cli_create_*_backingstore` calls `exec_mcg_cmd` with `ignore_error=False` (the default), so a CLI failure raises `CommandFailed`
2. After each creation, the factory calls `check_backingstore_state(name, BS_OPTIMAL, timeout=600)`, which raises `TimeoutExpiredError` if the store never reaches `OPTIMAL`
3. Both checks happen inside the per-store loop, so the exception propagates immediately - the factory never returns a partial list

An `assert len(...) == 3` would be unreachable in the failure case since the factory raises before returning.

ypersky1980: Good point — I see the factory raises before returning if any store fails, so a partial list isn't possible with the current implementation. Accepted, thanks for the detailed explanation.

tests/functional/object/mcg/test_sts_bucket.py
ypersky1980: thanks.

ypersky1980 (approved):
LGTM

udaysk23 (approved):

PR #15579, opened 2026-07-03
Handle PV deletion race during VM teardown

### LINKED ISSUE #15702: [Bug]: VM teardown fails when backing PV is deleted between existence check and reload()
### Description

A race condition exists during CNV VM teardown in `VirtualMachine._cleanup_backing_pv()`.

The code checks whether the backing PersistentVolume exists and later calls `pv_obj.reload()`. If the PV is deleted by another cleanup operation between these two operations, `reload()` raises a `CommandFailed` exception and the entire VM teardown fails.

### Steps to Reproduce

1. Create a VM with a backing PVC.
2. Delete the VM and trigger teardown.
3. Ensure the backing PersistentVolume is deleted after the existence check but before `pv_obj.reload()` is executed.
4. Observe the teardown failure.

### Actual Behavior

VM teardown fails with an exception similar to:

```text
CommandFailed:
PersistentVolume "<pv-name>" not found
```

even though the PV has already been successfully removed.

### Expected Behavior

If the backing PV has already been deleted, teardown should log the condition, skip PV cleanup, and continue successfully.

Unexpected `CommandFailed` exceptions should still be propagated.

### Impact

- Causes intermittent teardown failures in CNV workload tests.
- Results in false-negative test failures.
- Reduces reliability of automation due to a timing race.

### Screenshots

N/A

### Environment

- Test Suite(s): CNV Workloads
- Platform(s): OpenShift Data Foundation
- Version(s): Observed during CNV teardown
- OS: N/A

### Additional Context

The fix catches the expected `CommandFailed` when the backing PV has already been removed, logs the event, skips PV cleanup, and allows teardown to complete successfully.

tchaitan-web:
Fixes: #15702 
## Description

Handle a race condition during VM teardown where the backing PersistentVolume
may be deleted after the existence check but before `reload()` is called.

Instead of failing teardown with a `CommandFailed` exception when the PV is
already deleted, log the condition and skip PV cleanup. Unexpected
`CommandFailed` exceptions continue to be raised.

## Verification

Verified by rerunning:

tests/functional/workloads/cnv

The teardown completed successfully without the previous
`PersistentVolume ... not found` failure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved virtual machine and cloner teardown to reliably clean up backing persistent volumes, even when resources are removed concurrently.
  * Added more resilient handling for “not found” scenarios during storage cleanup, reducing teardown failures and allowing deletion to proceed smoothly.

### DIFF
--- a/ocs_ci/ocs/cnv/virtual_machine.py
+++ b/ocs_ci/ocs/cnv/virtual_machine.py
@@ -736,6 +736,26 @@ chpasswd:
         """
         return self.get().get("status").get("printableStatus")
 
+    def _cleanup_backing_pv(self, pv_obj):
+        """
+        Clean up the backing PV.
+
+        Handles the case where the PV was already deleted due to the
+        Delete reclaim policy before VM teardown reaches it.
+        """
+        try:
+            pv_obj.reload()
+        except CommandFailed as ex:
+            if "not found" in str(ex).lower():
+                logger.info(
+                    "PV %s already deleted. Skipping PV cleanup.",
+                    pv_obj.name,
+                )
+            else:
+                raise
+        else:
+            delete_pv_with_force_and_finalizers(pv_obj, timeout=600)
+
     def delete(self):
         """
         Delete the VirtualMachine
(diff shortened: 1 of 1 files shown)


ocs_ci/ocs/cnv/virtual_machine.py
```diff
@@ -736,6 +736,26 @@ def printableStatus(self):
         """
         return self.get().get("status").get("printableStatus")
 
+    def _cleanup_backing_pv(self, pv_obj):
+        """
+        Clean up the backing PV.
+
+        Handles the case where the PV was already deleted due to the
+        Delete reclaim policy before VM teardown reaches it.
+        """
```
PrasadDesala, line 748: missing doc string for fun arg

ocs_ci/ocs/cnv/virtual_machine.py
```diff
@@ -736,6 +736,26 @@ def printableStatus(self):
         """
         return self.get().get("status").get("printableStatus")
 
+    def _cleanup_backing_pv(self, pv_obj):
```
PrasadDesala, line 739: The VMCloner.delete() method (lines ~979, ~983) still uses the old pattern. VMCloner.delete() can still hit CommandFailed- not found during teardown. Should we fix that one too?

ocs_ci/ocs/cnv/virtual_machine.py
tchaitan-web: Thanks! Added the missing docstring for the pv_obj argument.

ocs_ci/ocs/cnv/virtual_machine.py
tchaitan-web: Updated VMCloner.delete() to reuse _cleanup_backing_pv() so the PV deletion race is handled consistently in both VirtualMachine and VMCloner cleanup paths.

tchaitan-web:
PR validation job was successful

PrasadDesala (approved):

hnallurv (approved):

PR #15580, opened 2026-07-05
Improve TimeoutSampler failure diagnostics

AviadP:
When a TimeoutSampler wait times out, the failure reason is hard to find - wait_for_func_status() swallows the exception entirely, and the raised TimeoutExpiredError carries no link back to what actually failed.

- Chain TimeoutExpiredError to the last exception raised by the sampled function so the root cause shows up in the failure traceback
- Expose attempt, last_result and last_exception attributes for post-mortem inspection after a timeout
- Include the failure reason (exception type only) in the rate-limited INFO log, keeping full messages at DEBUG to avoid leaking sensitive text
- Report the last sampled value in the wait_for_func_value timeout error, sanitized for safe logging
- Fix a crash when sampling callables without __name__ (e.g. functools.partial)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout and retry handling with better tracking of the latest attempt, result, and exception.
  * Timeout errors now preserve the original failure as the cause when applicable.
  * Logging was tightened to avoid exposing full exception messages or large sampled values, while still showing useful summaries.
  * `wait_for_func_value()` now reports clearer diagnostics when a value never succeeds or times out after transient failures.

* **Tests**
  * Added coverage for timeout chaining, logging behavior, and post-timeout state across several retry scenarios.

### DIFF
--- a/ocs_ci/utility/tests/test_utils_timeout_sampler.py
+++ b/ocs_ci/utility/tests/test_utils_timeout_sampler.py
@@ -1,5 +1,6 @@
 # -*- coding: utf8 -*-
 
+import functools
 import logging
 import time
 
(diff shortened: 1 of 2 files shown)


AviadP:
> PR validation
> 
> Cluster Name: Cluster Configuration: PR Test Suite: tier1 PR Test Path: tests/ Additional Test Params: OCP VERSION: 4.22-ga OCS VERSION: 4.22 tested against branch: master
> 
> [Job UNSTABLE](https://url.corp.redhat.com/5d45aaf) (some or all tests failed).

full tier1, none of the failure related to the fix

dahorak (approved):

clacroix12 (approved):

PR #15586, opened 2026-07-05
Fix bucket policy UI principal to use ARN format on OCS 4.21+

AviadP:
- Use OBC.id_for_policy_principal instead of the stale obc_account field when resolving the real account for account-dependent bucket policy tests
- Teach gen_bucket_policy_ui_compatible to recognize already-formatted ARN strings and pass them through instead of re-wrapping them
- Fixes PolicyApplicationError: "Invalid principal in policy" on OCS >= 4.21 for test_set_bucket_policy_ui[policy_config1] and [policy_config3]

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bucket policy principal handling so existing identity strings are preserved correctly instead of being converted to a different format.
  * Updated account identification used in bucket permissions to better match current platform behavior, including support for ARN-based values.

### DIFF
--- a/ocs_ci/ocs/ui/page_objects/bucket_tab_permissions.py
+++ b/ocs_ci/ocs/ui/page_objects/bucket_tab_permissions.py
@@ -116,7 +116,7 @@ class BucketsTabPermissions(ObjectStorage, ConfirmDialog):
             bucket_name (str): Name of the bucket to get account ID for.
 
         Returns:
-            str: Real account ID associated with the bucket.
+            str: Real account ID (or ARN, on OCS >= 4.21) associated with the bucket.
 
         Raises:
             ValueError: If bucket is not found or account ID cannot be retrieved.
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/ui/page_objects/bucket_tab_permissions.py
```diff
@@ -116,7 +116,7 @@ def _get_real_account_id_from_bucket(self, bucket_name: str) -> str:
             bucket_name (str): Name of the bucket to get account ID for.
 
         Returns:
-            str: Real account ID associated with the bucket.
+            str: Real account ID (or ARN, on OCS >= 4.21) associated with the bucket.
```
DanielOsypenko, line 119: is this chance applicable to 4.21?

AviadP: yes, will backport it after merge

DanielOsypenko (approved):

ebenahar (approved):

PR #15610, opened 2026-07-07
Fix revert_noobaa_endpoint_scc for Provider Client clusters

sagihirshfeld:
## Summary
- Fix `revert_noobaa_endpoint_scc_fixture` failing on Provider Client clusters
  because the `noobaa-endpoint` SCC doesn't exist on the client cluster.

## Changes
- Add `@config.run_with_provider_context_if_available` to
  `revert_noobaa_endpoint_scc_fixture` so the initial SCC lookup runs
  against the provider cluster.
- Add the same decorator to the inner `revert_endpoint_scc_implementation`
  and `finalizer` functions so SCC patching and teardown also target the
  provider cluster (the finalizer decorator is critical since pytest calls
  it outside the outer function's context).

## Summary by CodeRabbit

* **Tests**
  * Updated test fixture behavior so SCC patching and restoration execute within the correct provider execution context when available.
  * Improved teardown finalization to ensure settings are consistently reverted during cleanup.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7205,6 +7205,7 @@ def revert_noobaa_endpoint_scc_class(request):
     return revert_noobaa_endpoint_scc_fixture(request)
 
 
+@config.run_with_provider_context_if_available
 def revert_noobaa_endpoint_scc_fixture(request):
     """
     This fixture reverts the noobaa-endpoint SCC back to the way it was before ODF 4.12.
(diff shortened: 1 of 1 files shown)


sagihirshfeld:
Adding the Verfied label due to the limited nature of the change

udaysk23 (approved):

ypersky1980 (approved):
LGTM

PR #15378, opened 2026-06-11
introduce a base class for fdf mirror and fdf deploy

rh-aaleem:
## Refactor
* Restructured framework initialization logic by extracting common functionality into a centralized base class, improving code organization and reducing duplication across initialization modules.

## Release Notes

* **Refactor**
  * Centralized shared initialization for framework runs, including run ID generation, configuration loading, per-run logging setup, and cluster connectivity validation.
  * Updated fusion FDF initializer to reuse the shared initialization flow and removed now-redundant local helpers.
* **Bug Fixes**
  * Only set reporting output path when reporting is explicitly provided and enabled.

PR #15652, opened 2026-07-09
[release-4.21] introduce-base-class

rh-aaleem:
This is  cherry-pick of https://github.com/red-hat-storage/ocs-ci/pull/15378

### DIFF
--- a/ocs_ci/utility/framework/base_initializer.py
+++ b/ocs_ci/utility/framework/base_initializer.py
@@ -1,0 +1,201 @@
+"""
+Base Initializer Module
+
+Provides common initialization functionality for all initializer types.
+"""
+
+import logging
+import os
+import time
+
+from ocs_ci import framework
+from ocs_ci.framework import config
+from ocs_ci.framework.exceptions import ClusterNameNotProvidedError
+from ocs_ci.ocs import constants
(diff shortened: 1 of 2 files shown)


dahorak (dismissed):

dahorak (approved):

ocs_ci/utility/framework/base_initializer.py
```diff
@@ -0,0 +1,201 @@
...
+        # Create kubeconfig if doesn't exist
+        create_kubeconfig(config.RUN["kubeconfig"])
+
+        self.setup_bin_dir()
+        self.check_cluster_access(config.RUN["kubeconfig"])
+
+        # Load OCP version config
+        self.load_ocp_version_config()
+
+    def load_ocp_version_config(self) -> None:
```
clacroix12, line 148: I noticed this is implemented in both the base class and in the Initializer class in `fusion_fdf_init`. It is functionally not a problem but since the logic doesn't appear to change we probably want to fix that in a follow-up pr. This appears to be the case in the original PR so not really an issue with the backport.

clacroix12 (approved):

petr-balogh (approved):

PR #15671, opened 2026-07-10
Fix azure aro destroy release 4.20

petr-balogh:


dahorak (approved):

petr-balogh:
Verified as part of 4.19 cherry-pick, so probably not much risk to merge it here as well as it's same cherry-pick with no conflict.
Verification job:
https://url.corp.redhat.com/4133d7a

PR #15672, opened 2026-07-10
Fix ARO deployment to clean up orphaned Azure resources

petr-balogh:
ARO deployments were leaving orphaned resources:
- Azure AD applications named 'aro-{cluster-name}'
- VNETs named 'aro-vnet-{cluster-name}'
- DNS records

These resources were not cleaned up during normal teardown, leading to resource accumulation.

Changes:
- Add delete_azure_ad_app() method to clean up AD applications
- Enhance destroy_cluster() to delete all resources (DNS, cluster, VNET, AD app)
- Use ignore_error=True for all delete operations

All cleanup operations are idempotent - they handle missing resources gracefully and will not fail if resources have already been deleted or were never created.

The destroy_cluster() method now:
1. Deletes DNS records
2. Deletes ARO cluster (which also deletes cluster resource group)
3. Deletes VNET
4. Deletes Azure AD application

Each operation logs whether the resource was found and deleted, or was already gone. This ensures cleanup works even when deployment partially fails or resources were manually deleted.

### DIFF
--- a/ocs_ci/utility/azure_utils.py
+++ b/ocs_ci/utility/azure_utils.py
@@ -631,13 +631,21 @@ class AzureAroUtil(AZURE):
         logger.info(f"Deleting service principal: {sp_name}")
 
         try:
-            # Get the app ID first
+            # Get the service principal by display name
+            # NOTE: --display-name does substring matching, so we need to filter for exact match
             list_cmd = f"az ad sp list --display-name {sp_name}"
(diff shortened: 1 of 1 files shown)


dahorak (approved):

petr-balogh:
Verified as part of 4.19 cherry-pick, so probably not much risk to merge it here as well as it's same cherry-pick with no conflict.
Verification job:
https://url.corp.redhat.com/4133d7a

PR #15678, opened 2026-07-11
Skip submariner MG if current phase is import clusters as submariner wouldn't be present

shylesh:
Due to https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/utils.py#L7696 , submariner_source will always be true and hence it forced submariner mustgather everytime.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostic log collection for Submariner: logs are now gathered whenever cluster import to ACM is disabled, even if a Submariner source is configured.
  * Log collection is skipped only when cluster import to ACM is enabled.

### DIFF
--- a/ocs_ci/ocs/utils.py
+++ b/ocs_ci/ocs/utils.py
@@ -1380,9 +1380,7 @@ def _collect_ocs_logs(
                 )
 
             # We want to skip submariner log collection if it's in import clusters phase
-            if not cluster_config.ENV_DATA.get(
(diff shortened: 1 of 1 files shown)


vavuthu (approved):

shylesh:
https://url.corp.redhat.com/subctlskip - now in case of import clusters scenario we will skip submariner MG collection

sidhant-agrawal (approved):

PR #15788, opened 2026-07-23
New system test case for in-cluster and out-cluster volume mount and related operations

dgaikwad:
Added a new NFS export system test validating simultaneous in-cluster (pod) and
out-of-cluster (NFS client VM) access across both RWX and RWO access modes.

Covers:
- Continuous background I/O during NFS server pod node reboot; file is readable and able to write new data after reboot as well.
- PVC snapshot and restore: writes I/O to the restored PVC and cross-validates
  checksums between the in-cluster pod and the out-of-cluster NFS mount.
- PVC clone from restored snapshot: writes I/O, resizes clone from 10Gi to 15Gi,
  verifies data integrity across both mount types.
- Second snapshot/restore cycle on the resized clone: writes I/O, resizes to 20Gi,
  verifies checksums and line counts.
- Non-graceful cluster stop/start (ordered shutdown): force-stops workers then
  control-plane, starts in reverse order, verifies Ceph health, confirms all NFS
  mount points are accessible, and validates post-recovery I/O on all pods.

Base class (NFSClientTestBase) added to ocs_ci/utility/nfs_base.py providing
SSH connection management, automatic NFS client VM reboot on connection failure,
and NFS mount retry logic shared across NFS test classes.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive end-to-end NFS export validation for in-cluster and external clients, including node reboot, snapshot/restore, clone/resize, and forced recovery checks using checksum and line-count validation.

* **Bug Fixes**
  * Improved NFS enable/disable and cleanup with added readiness waits, conditional StorageClass deletion, and more resilient external NFS connectivity and mount retries.

* **Other Changes**
  * Increased long-running I/O timeout and improved remote file probing for reliability; added safety delay for IBM Cloud load balancer rule propagation.

### DIFF
--- a/ocs_ci/helpers/nfs_helpers.py
+++ b/ocs_ci/helpers/nfs_helpers.py
@@ -1,0 +1,161 @@
+"""
+Base class for NFS tests that require out-of-cluster client connection.
+
+This module provides a base test class with common functionality for NFS tests
+that need to connect to an external NFS client VM for testing NFS exports.
+"""
+
+import logging
+import socket
+import time
+
+from ocs_ci.framework import config
+from ocs_ci.framework.testlib import ManageTest
+from ocs_ci.ocs.exceptions import CommandFailed, ConfigurationError
+from ocs_ci.utility import nfs_utils
+from ocs_ci.utility.connection import Connection
+from ocs_ci.utility.retry import retry
+from ocs_ci.utility.utils import exec_cmd
+
+log = logging.getLogger(__name__)
+
+
+class NFSClientTestBase(ManageTest):
+    """
+    Base class for NFS tests that need out-of-cluster client connection.
+
+    This class provides common functionality for:
+    - Establishing SSH connection to NFS client VM
+    - Handling connection failures with automatic VM reboot
+    - Mounting NFS exports with retry logic
+    - Hostname resolution and /etc/hosts management
+
+    Test classes should inherit from this base class and will have access to:
+    - self.con: Property that returns an active connection to the NFS client VM
+    - self.get_nfs_client_connection(): Method to create a new connection
+    - self._mount_nfs_with_retry(): Method to mount NFS with retry logic
+    """
+
+    # If the connection to the NFS Client VM fails, it's possible that
+    # the NFS Client VM might not be healthy, so rebooting it and re-trying
+    @property
+    @retry((TimeoutError, socket.gaierror), tries=3, delay=60, backoff=1)
+    def con(self):
+        """
+        Create connection to NFS Client VM, if not accessible, try to restart it.
+
+        This property provides a cached connection to the NFS client VM. If the
+        connection fails, it will attempt to reboot the VM using OpenStack CLI
+        (if configured) and retry the connection.
+
+        Returns:
+            Connection: Active SSH connection to the NFS client VM
+
+        Raises:
+            ConfigurationError: If VM is not accessible and reboot config is missing
+            TimeoutError: If connection cannot be established after retries
+            socket.gaierror: If hostname resolution fails
+        """
+        if (
+            not hasattr(self, "__nfs_client_connection")
+            or not self.__nfs_client_connection
+        ):
+            try:
+                self.__nfs_client_connection = self.get_nfs_client_connection(
+                    re_try=False
+                )
+            except (TimeoutError, socket.gaierror):
+                nfs_client_vm_cloud = config.ENV_DATA.get("nfs_client_vm_cloud")
+                nfs_client_vm_name = config.ENV_DATA.get("nfs_client_vm_name")
+                if not nfs_client_vm_cloud or not nfs_client_vm_name:
+                    raise ConfigurationError(
+                        "NFS Client VM is not accessible and ENV_DATA nfs_client_vm_cloud and/or nfs_client_vm_name "
+                        "parameters are not configured to be able to automatically reboot the NFS Client VM."
+                    )
+                cmd = f"openstack --os-cloud {nfs_client_vm_cloud} server reboot --hard --wait {nfs_client_vm_name}"
+                exec_cmd(cmd)
+
+                time.sleep(60)
+                self.__nfs_client_connection = self.get_nfs_client_connection()
+        return self.__nfs_client_connection
+
+    def get_nfs_client_connection(self, re_try=True):
+        """
+        Create connection to NFS Client VM.
+
+        After establishing the SSH connection, if the NFS LB endpoint is a
+        hostname (not a raw IP), the hostname is resolved from within the
+        cluster and /etc/hosts on the client VM is updated. This is required
+        when the NFS client VM is in a different VPC from the OpenShift cluster
+        and cannot resolve IBM Cloud VPC LB hostnames via its DNS servers.
+
+        If hostname resolution from the cluster fails (timeout), the code will
+        proceed without updating /etc/hosts, assuming the NFS client VM can
+        resolve the hostname via its own DNS configuration.
+
+        Args:
+            re_try (bool): Whether to retry connection on failure (default: True)
+
+        Returns:
+            Connection: SSH connection object to the NFS client VM
+
+        Raises:
+            TimeoutError: If connection cannot be established
+            socket.gaierror: If hostname resolution fails
+        """
+        log.info("Connecting to nfs client test VM")
+        tries = 3 if re_try else 1
+
+        @retry((TimeoutError, socket.gaierror), tries=tries, delay=60, backoff=1)
+        def __make_connection():
+            return Connection(
+                self.nfs_client_ip,
+                self.nfs_client_user,
+                private_key=self.nfs_client_private_key,
+            )
+
+        con = __make_connection()
+        hostname_add = getattr(self, "hostname_add", None)
+        if hostname_add:
+            try:
+                nfs_utils.update_etc_hosts_on_nfs_client(con, hostname_add)
+            except TimeoutError:
+                log.warning(
+                    f"Timed out resolving hostname {hostname_add} from cluster; "
+                    "continuing without /etc/hosts update on NFS client VM"
+                )
+        return con
+
+    def _mount_nfs_with_retry(self, mount_dir, export_path, options="", retries=3):
+        """
+        Mount NFS export with retry logic and cleanup.
+
+        Args:
+            mount_dir (str): Local mount point directory
+            export_path (str): NFS export path in format server:/path
+            options (str): Additional mount options
+            retries (int): Number of retry attempts
+
+        Raises:
+            CommandFailed: If mount fails after all retries
+        """
+        mount_cmd = f"mount -t nfs {options} {export_path} {mount_dir}".strip()
+
+        for attempt in range(retries):
+            retcode, stdout, stderr = self.con.exec_cmd(mount_cmd)
+            if retcode == 0:
+                log.info(f"Successfully mounted NFS export on attempt {attempt + 1}")
+                return
+
+            msg = (
+                f"Mount attempt {attempt + 1} failed (rc={retcode}): {stderr or stdout}"
+            )
+            if attempt == retries - 1:
+                raise CommandFailed(msg)
+
+            log.warning(f"{msg}. Retrying in 10 seconds...")
+            self.con.exec_cmd(f"umount -f {mount_dir}")
+            time.sleep(10)
+
+
+# Made with Bob
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -9,6 +9,7 @@ from datetime import datetime, timedelta
 import logging
 import os
 import re
+import shlex
 import yaml
 import tempfile
 import time
@@ -47,12 +48,11 @@ from ocs_ci.utility.utils import (
     TimeoutSampler,
     exec_cmd,
 )
-from ocs_ci.utility.utils import check_if_executable_in_path
 from ocs_ci.utility.retry import retry
 from ocs_ci.ocs.constants import CSI_RBD_ADDON_NODEPLUGIN_LABEL_420
 
 logger = logging.getLogger(__name__)
-FIO_TIMEOUT = 600
+FIO_TIMEOUT = 900
 
 TEXT_CONTENT = (
     "Lorem ipsum dolor sit amet, consectetur adipiscing elit, "
@@ -1376,14 +1376,19 @@ def check_file_existence(pod_obj, file_path):
     Returns:
         bool: True if the file exist, False otherwise
     """
+    # try:
+    #     check_if_executable_in_path(pod_obj.exec_cmd_on_pod("which find"))
+    # except CommandFailed:
+    #     pod_obj.install_packages("findutils")
     try:
-        check_if_executable_in_path(pod_obj.exec_cmd_on_pod("which find"))
+        ret = pod_obj.exec_cmd_on_pod(
+            f"bash -c 'ls {shlex.quote(file_path)}'", out_yaml_format=False
+        )
+        if file_path in ret:
+            return True
+        return False
     except CommandFailed:
-        pod_obj.install_packages("findutils")
-    ret = pod_obj.exec_cmd_on_pod(f'bash -c "find {file_path}"')
-    if re.search(file_path, ret):
-        return True
-    return False
+        return False
 
 
 def get_file_path(pod_obj, file_name):
--- a/ocs_ci/utility/nfs_utils.py
+++ b/ocs_ci/utility/nfs_utils.py
@@ -7,6 +7,7 @@ import json
 import logging
 import yaml
 import pytest
+import time
 from ocs_ci.ocs import constants, resources, ocp
 from ocs_ci.helpers import helpers
 from ocs_ci.ocs.resources import pod
@@ -140,8 +141,43 @@ def nfs_disable(
     # Wait untill nfs-ganesha pod deleted
     pod_obj.wait_for_delete(resource_name=nfs_ganesha_pod_name)
 
-    # Delete the nfs StorageClass
-    sc_obj.delete(resource_name=constants.NFS_STORAGECLASS_NAME)
+    # Wait for the StorageCluster to reach Ready phase before touching the SC.
+    # The nfs-controller reconciler hits a retriggered reconcile loop during NFS
+    # disable (confirmed: "failed to remove ganesha from grace db" causes 3x
+    # retriggers). Each loop recreates the SC as part of the NFS cleanup sequence,
+    # so manually deleting the SC before reconciliation finishes races with the
+    # operator and the SC reappears on every poll.
+    # Waiting for Ready ensures all retrigger loops have completed and the operator
+    # has either removed the SC itself or will no longer recreate it.
+    log.info(
+        "Waiting for StorageCluster to reach Ready phase after disabling NFS "
+        "(ensures reconciler retrigger loops are complete)..."
+    )
+    storage_cluster = StorageCluster(
+        resource_name=config.ENV_DATA["storage_cluster_name"],
+        namespace=config.ENV_DATA["cluster_namespace"],
+    )
+    storage_cluster.wait_for_phase(phase="Ready", timeout=600)
+    log.info("StorageCluster is Ready — safe to delete NFS StorageClass")
+
+    # Only delete the SC if the operator has not already removed it during
+    # reconciliation — after a clean reconcile it may already be gone.
+    if sc_obj.is_exist(resource_name=constants.NFS_STORAGECLASS_NAME):
+        log.info(f"Deleting NFS StorageClass {constants.NFS_STORAGECLASS_NAME}")
+        sc_obj.delete(resource_name=constants.NFS_STORAGECLASS_NAME)
+        log.info(
+            f"Waiting for NFS StorageClass {constants.NFS_STORAGECLASS_NAME} to be deleted..."
+        )
+        sc_obj.wait_for_delete(
+            resource_name=constants.NFS_STORAGECLASS_NAME, timeout=120
+        )
+        log.info(
+            f"NFS StorageClass {constants.NFS_STORAGECLASS_NAME} deleted successfully"
+        )
+    else:
+        log.info(
+            f"NFS StorageClass {constants.NFS_STORAGECLASS_NAME} already removed by operator"
+        )
 
 
 def create_nfs_load_balancer_service(
@@ -217,7 +253,12 @@ def create_nfs_load_balancer_service(
             configure_nfs_lb_security_group,
         )
 
+        log.info("Configuring IBM Cloud security group for NFS LoadBalancer...")
         configure_nfs_lb_security_group()
+        log.info(
+            "Security group configured. Waiting 60 seconds for rules to propagate..."
+        )
+        time.sleep(60)
 
     return hostname_add
 
@@ -850,3 +891,94 @@ def fetch_nfs_server_details_on_client_cluster(default_server=False):
                 server,
             )
             return server
+
+
+def get_file_checksum_from_nfs(con, file_path):
+    """
+    Get MD5 checksum for a file on NFS mount (out-of-cluster).
+
+    Args:
+        con: Connection object to NFS client VM
+        file_path (str): Full path to the file on NFS mount
+
+    Returns:
+        str: MD5 checksum of the file
+
+    Raises:
+        CommandFailed: If the md5sum command fails
+    """
+    checksum_cmd = f"md5sum {file_path}"
+    retcode, stdout, stderr = con.exec_cmd(checksum_cmd)
+    if retcode != 0:
+        raise CommandFailed(
+            f"Failed to calculate checksum for {file_path}: {stderr or stdout}"
+        )
+    return stdout.split()[0]
+
+
+def get_file_line_count_from_nfs(con, file_path):
+    """
+    Get line count for a file on NFS mount (out-of-cluster).
+
+    Args:
+        con: Connection object to NFS client VM
+        file_path (str): Full path to the file on NFS mount
+
+    Returns:
+        int: Line count of the file
+
+    Raises:
+        CommandFailed: If the wc -l command fails
+    """
+    line_cmd = f"wc -l {file_path}"
+    retcode, stdout, stderr = con.exec_cmd(line_cmd)
+    if retcode != 0:
+        raise CommandFailed(
+            f"Failed to get line count for {file_path}: {stderr or stdout}"
+        )
+    return int(stdout.split()[0])
+
+
+def frame_deployment_config(deployment_name, pvc_name, node_name=None):
+    """
+    Frame a deployment configuration for an app pod that mounts a given PVC.
+
+    The returned dict can be passed directly to ``helpers.create_resource``.
+    Pass ``node_name`` to pin the pod to a specific worker node — use this to
+    guarantee the pod never lands on a node that will be rebooted during the
+    test (e.g. the node hosting the NFS server pod).
+
+    Args:
+        deployment_name (str): Name for the Deployment and its pod label.
+        pvc_name (str): Name of the PVC to mount at ``/mnt``.
+        node_name (str): Optional worker node hostname. When provided,
+            ``spec.template.spec.nodeName`` is set so Kubernetes schedules the
+            pod directly onto that node.
+
+    Returns:
+        dict: Deployment manifest ready for creation.
+    """
+    from ocs_ci.utility import templating
+
+    # Load base deployment template
+    deployment_data = templating.load_yaml(constants.NFS_APP_POD_YAML)
+
+    # Set deployment name
+    deployment_data["metadata"]["name"] = deployment_name
+
+    # Set label values (used in multiple places for pod selection)
+    deployment_data["metadata"]["labels"]["app"] = deployment_name
+    deployment_data["spec"]["selector"]["matchLabels"]["name"] = deployment_name
+    deployment_data["spec"]["template"]["metadata"]["labels"]["name"] = deployment_name
+
+    # Set PVC claimName
+    deployment_data["spec"]["template"]["spec"]["volumes"][0]["persistentVolumeClaim"][
+        "claimName"
+    ] = pvc_name
+
+    # Pin the pod to a specific node so it never lands on the node being rebooted
+    if node_name:
+        deployment_data["spec"]["template"]["spec"]["nodeName"] = node_name
+        log.info(f"Deployment '{deployment_name}': nodeName set to '{node_name}'")
+
+    return deployment_data
--- a/tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
+++ b/tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
@@ -1,0 +1,2161 @@
+import inspect
+import logging
+import os
+import random
+import threading
+import time
+from threading import Thread
+
+import pytest
+
+from ocs_ci.framework import config
+from ocs_ci.framework.pytest_customization.marks import (
+    skipif_rosa_hcp,
+    skipif_lean_deployment,
+)
+from ocs_ci.framework.testlib import (
+    skipif_ocs_version,
+    tier1,
+    skipif_ocp_version,
+    skipif_managed_service,
+    skip_for_provider_or_client_if_ocs_version,
+    skipif_disconnected_cluster,
+    skipif_proxy_cluster,
+    skipif_external_mode,
+    skipif_hci_client,
+)
+from ocs_ci.helpers.nfs_helpers import NFSClientTestBase
+from ocs_ci.helpers import helpers
+from ocs_ci.ocs import constants, ocp, platform_nodes
+from ocs_ci.ocs.node import (
+    wait_for_nodes_status,
+    get_node_objs,
+    get_all_nodes,
+    get_worker_nodes,
+)
+from ocs_ci.ocs.resources import pod, ocs, pvc
+from ocs_ci.ocs.resources.pod import (
+    get_all_pods,
+    wait_for_pods_to_be_running,
+)
+from ocs_ci.ocs.resources.pvc import create_pvc_snapshot
+from ocs_ci.utility.nfs_utils import (
+    frame_deployment_config,
+)
+from ocs_ci.utility import nfs_utils
+from ocs_ci.utility import version as version_module
+from ocs_ci.utility.utils import ceph_health_check
+
+log = logging.getLogger(__name__)
+
+
+@skipif_rosa_hcp
+@skipif_external_mode
+@skipif_ocs_version("<4.11")
+@skipif_ocp_version("<4.11")
+@skipif_managed_service
+@skipif_hci_client
+@skip_for_provider_or_client_if_ocs_version("<4.19")
+@skipif_disconnected_cluster
+@skipif_proxy_cluster
+@skipif_lean_deployment
+class TestNfsExport(NFSClientTestBase):
+    """ """
+
+    @pytest.fixture(scope="class", autouse=True)
+    def setup_teardown(
+        self,
+        request,
+    ):
+        """
+        Setup-Teardown for the class
+
+        Pre-Req:
+        By default in our jenkins jobs we are creating one of our custom config file,
+        so we can make sure
+        ENV_DATA:
+            nfs_client_ip: "10.xx.xxx.xx"
+            nfs_client_user: "root"
+            nfs_client_private_key: constants.SSH_PRIV_KEY
+        these values are provided in all our automation runs in Jenkins.
+
+        But if someone will run locally, they will need to create custom config file and provide that via
+        --ocsci-conf in order to run the external nfs consume tests. Example:
+        ENV_DATA:
+            nfs_client_ip: "10.xx.xxx.xx"
+            nfs_client_user: "root"
+            nfs_client_private_key: "<path to ssh private key>"
+
+        If this VM IP is not available in config, then the external nfs consume tests will be skipped.
+
+        Steps:
+        ---Setup---
+        1:- Create objects for storage_cluster, configmap, pod, pv, pvc, service and storageclass
+        2:- Fetch number of cephfsplugin and cephfsplugin_provisioner pods running
+        3:- Enable nfs feature
+        4:- Create loadbalancer service for nfs
+        ---Teardown---
+        5:- Disable nfs feature
+        6:- Delete ocs nfs Service
+
+        """
+        cls = request.cls
+        log.info("-----Setup-----")
+        cls.nfs_app_deployment = "nfs-test-pod"
+        cls.namespace = config.ENV_DATA["cluster_namespace"]
+        cls.storage_cluster_obj = ocp.OCP(
+            kind=constants.STORAGECLUSTER, namespace=cls.namespace
+        )
+        cls.sc_obj = ocp.OCP(kind=constants.STORAGECLASS)
+        cls.config_map_obj = ocp.OCP(kind=constants.CONFIGMAP, namespace=cls.namespace)
+        cls.pod_obj = ocp.OCP(kind=constants.POD, namespace=cls.namespace)
+        cls.service_obj = ocp.OCP(kind=constants.SERVICE, namespace=cls.namespace)
+        cls.pvc_obj = ocp.OCP(kind=constants.PVC, namespace=cls.namespace)
+        cls.pv_obj = ocp.OCP(kind=constants.PV, namespace=cls.namespace)
+        cls.nfs_sc = constants.NFS_STORAGECLASS_NAME
+        cls.sc = ocs.OCS(kind=constants.STORAGECLASS, metadata={"name": cls.nfs_sc})
+        cls.retain_nfs_sc_name = "ocs-storagecluster-ceph-nfs-retain"
+        platform = config.ENV_DATA.get("platform", "").lower()
+        cls.is_vsphere = platform == constants.VSPHERE_PLATFORM
+        if cls.is_vsphere:
+            log.info(
+                "vSphere platform detected: out-of-cluster NFS operations will be skipped. "
+                "All I/O and data validation will be performed from in-cluster pods."
+            )
+        cls.run_id = config.RUN.get("run_id")
+        cls.test_folder = f"mnt/test_nfs_{cls.run_id}"
+        log.info(f"nfs mount point out of cluster is----- {cls.test_folder}")
+        cls.nfs_client_ip = config.ENV_DATA.get("nfs_client_ip")
+        log.info(f"nfs_client_ip is: {cls.nfs_client_ip}")
+
+        cls.nfs_client_user = config.ENV_DATA.get("nfs_client_user")
+        log.info(f"nfs_client_user is: {cls.nfs_client_user}")
+
+        # Only load NFS client credentials when not on vSphere
+        if not cls.is_vsphere:
+            cls.nfs_client_private_key = os.path.expanduser(
+                config.ENV_DATA.get("nfs_client_private_key")
+                or config.DEPLOYMENT["ssh_key_private"]
+            )
+
+        # Enable nfs feature
+        log.info("----Enable nfs----")
+        if (
+            config.default_cluster_ctx.ENV_DATA["cluster_type"].lower()
+            == constants.HCI_CLIENT
+        ):
+            nfs_ganesha_pod, cls.hostname_add = nfs_utils.nfs_access_for_clients(
+                cls.nfs_sc
+            )
+
+            # Create a duplicate sc of nfs-sc and update the server details with hostname_add
+            if (
+                version_module.get_semantic_ocs_version_from_config()
+                < version_module.VERSION_4_21
+            ):
+                _ = nfs_utils.create_nfs_sc(
+                    sc_name_to_create=constants.COPY_NFS_STORAGECLASS_NAME,
+                    sc_name_to_copy=cls.nfs_sc,
+                    server=cls.hostname_add,
+                )
+                cls.nfs_sc = constants.COPY_NFS_STORAGECLASS_NAME
+            yield
+            # Remove NFS SC from distributed storage classes on the provider
+            nfs_utils.remove_nfs_storage_class_from_all_consumers(
+                constants.NFS_STORAGECLASS_NAME
+            )
+            # Disable nfs feature
+            nfs_utils.disable_nfs_service_from_provider(cls.sc, nfs_ganesha_pod)
+
+            # delete nfs non default storageclass if available
+            if ocp.OCP(kind=constants.STORAGECLASS).is_exist(
+                resource_name=constants.COPY_NFS_STORAGECLASS_NAME
+            ):
+                cls.sc_obj.delete(resource_name=constants.COPY_NFS_STORAGECLASS_NAME)
+
+        else:
+            nfs_ganesha_pod_name = nfs_utils.nfs_enable(
+                cls.storage_cluster_obj,
+                cls.config_map_obj,
+                cls.pod_obj,
+                cls.namespace,
+            )
+
+            # Load balancer is not supported on vSphere; skip LB creation there
+            if (
+                platform == constants.AWS_PLATFORM
+                or platform == constants.IBMCLOUD_PLATFORM
+                or platform == constants.HCI_BAREMETAL
+            ):
+                # Create loadbalancer service for nfs
+                cls.hostname_add = nfs_utils.create_nfs_load_balancer_service(
+                    cls.storage_cluster_obj,
+                )
+
+            yield
+
+            log.info("-----Teardown-----")
+            # Disable nfs feature
+            nfs_utils.nfs_disable(
+                cls.storage_cluster_obj,
+                cls.config_map_obj,
+                cls.pod_obj,
+                cls.sc,
+                nfs_ganesha_pod_name,
+            )
+            if (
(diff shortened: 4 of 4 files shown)


dgaikwad:
Screenshot 2026-08-25 at 4 42 14 PM
@amr1ta , intermittently we are seeing this error where the NFS StorageClass is reported as a leftover, but from the teardown logs it appears to be deleted successfully. However, on one occasion when I manually checked the cluster, the StorageClass was still present. It might be getting recreated, and I noticed its owner/reference kind was StorageClient. Have you observed this issue earlier?

amr1ta:
> Screenshot 2026-08-25 at 4 42 14 PM @amr1ta , intermittently we are seeing this error where the NFS StorageClass is reported as a leftover, but from the teardown logs it appears to be deleted successfully. However, on one occasion when I manually checked the cluster, the StorageClass was still present. It might be getting recreated, and I noticed its owner/reference kind was StorageClient. Have you observed this issue earlier?

@dgaikwad I have not observed this, if you think it is a repetitive incidence then please open an issue on my name I will investigate.

ocs_ci/helpers/nfs_helpers.py
```diff
@@ -0,0 +1,161 @@
...
+    - self.con: Property that returns an active connection to the NFS client VM
+    - self.get_nfs_client_connection(): Method to create a new connection
+    - self._mount_nfs_with_retry(): Method to mount NFS with retry logic
+    """
+
+    # If the connection to the NFS Client VM fails, it's possible that
+    # the NFS Client VM might not be healthy, so rebooting it and re-trying
+    @property
+    @retry((TimeoutError, socket.gaierror), tries=3, delay=60, backoff=1)
+    def con(self):
```
PrasadDesala: TimeoutError is not imported

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+                raise AssertionError(f"Too many I/O errors detected: {len(io_errors)}")
+        else:
+            log.info("No I/O errors detected - all operations successful!")
+
+        # Verify data integrity: pod checksum is the source of truth on all platforms.
+        # On non-vSphere, also cross-check via NFS mount.
+        # No fixed expected_checksum here — the I/O thread was writing continuously,
+        # so we just confirm the file is non-empty and readable from the pod.
+        log.info("Verifying data integrity after node reboot...")
+        self.verify_data_integrity_from_pod(
```
PrasadDesala: there is no need to pass expected_checksum or expected_line_count?

ocs_ci/helpers/nfs_helpers.py
```diff
@@ -0,0 +1,161 @@
...
+            )
+            if attempt == retries - 1:
+                raise CommandFailed(msg)
+
+            log.warning(f"{msg}. Retrying in 10 seconds...")
+            self.con.exec_cmd(f"umount -f {mount_dir}")
+            time.sleep(10)
+
+
+# Made with Bob
```
PrasadDesala: you can remove it

ocs_ci/helpers/nfs_helpers.py
```diff
@@ -0,0 +1,161 @@
+"""
+Base class for NFS tests that require out-of-cluster client connection.
```
PrasadDesala: If this is the base class for NFS tests, should be into some other place and not in helpers?

ocs_ci/utility/nfs_utils.py
```diff
@@ -850,3 +891,94 @@ def _get_nfs_server():
                 server,
             )
             return server
+
+
+def get_file_checksum_from_nfs(con, file_path):
```
PrasadDesala: This new function is added to nfs_utils.py but are never imported or called from the test

ocs_ci/utility/nfs_utils.py
```diff
@@ -850,3 +891,94 @@ def _get_nfs_server():
...
+    checksum_cmd = f"md5sum {file_path}"
+    retcode, stdout, stderr = con.exec_cmd(checksum_cmd)
+    if retcode != 0:
+        raise CommandFailed(
+            f"Failed to calculate checksum for {file_path}: {stderr or stdout}"
+        )
+    return stdout.split()[0]
+
+
+def get_file_line_count_from_nfs(con, file_path):
```
PrasadDesala: This new func is created but never imported or called from the test or any other places

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+
+        new_size = "15Gi"
+        log.info(f"Expanding cloned PVC from 10Gi to {new_size}")
+
+        # Patch the PVC to request more storage
+        patch_cmd = (
+            f"patch pvc {cloned_pvc_name} -p "
+            f'\'{{"spec":{{"resources":{{"requests":{{"storage":"{new_size}"}}}}}}}}\''
+        )
+        result = cloned_pvc_obj.ocp.exec_oc_cmd(patch_cmd)
```
PrasadDesala: you can use the existing pvc resizr fun

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+        log.info(
+            f"Test case execution started: {inspect.currentframe().f_code.co_name}"
+        )
+        # On vSphere there is no load balancer so no out-of-cluster access is possible.
+        # Skip the nfs_client check only when NOT on vSphere.
+        if not self.is_vsphere:
+            nfs_utils.skip_test_if_nfs_client_unavailable(self.nfs_client_ip)
+
+        # Generate unique names using timestamp to avoid conflicts between test run
+        unique_suffix = f"{int(time.time())}-{random.randint(1000, 9999)}"
```
PrasadDesala: wouldn't random suffix alone is not sufficient?

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+
+        Cleanup:
+        --------
+        - Unmount NFS exports from both clients
+        - Delete all created resources (PVCs, snapshots, clones, pods, deployments)
+        - Verify all resources are properly cleaned up
+        """
+
+        log.info(
+            f"Test case execution started: {inspect.currentframe().f_code.co_name}"
```
PrasadDesala: This is an unusual pattern for logging the test name. The pytest request fixture is already available and provides request.node.name, which is cleaner and includes the parametrize suffix.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+            try:
+                chk = self.calculate_checksum_and_lines_from_pod(pod_obj, pod_io_file)
+                log.info(
+                    f"Final file: {chk['line_count']} lines, "
+                    f"checksum {chk['checksum']}"
+                )
+            except Exception as stat_error:
+                log.warning(f"Failed to get final file statistics: {stat_error}")
+
+    @tier1
```
PrasadDesala: why did we add tier1 here?. It is system test

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+@skipif_external_mode
+@skipif_ocs_version("<4.11")
+@skipif_ocp_version("<4.11")
+@skipif_managed_service
+@skipif_hci_client
+@skip_for_provider_or_client_if_ocs_version("<4.19")
+@skipif_disconnected_cluster
+@skipif_proxy_cluster
+@skipif_lean_deployment
+class TestNfsExport(NFSClientTestBase):
```
PrasadDesala, line 60: class doc string is empty

ocs_ci/helpers/nfs_helpers.py
```diff
@@ -0,0 +1,161 @@
...
+
+            msg = (
+                f"Mount attempt {attempt + 1} failed (rc={retcode}): {stderr or stdout}"
+            )
+            if attempt == retries - 1:
+                raise CommandFailed(msg)
+
+            log.warning(f"{msg}. Retrying in 10 seconds...")
+            self.con.exec_cmd(f"umount -f {mount_dir}")
+            time.sleep(10)
```
PrasadDesala: The _mount_nfs_with_retry method calls self.con.exec_cmd(mount_cmd) and self.con.exec_cmd(f"umount -f {mount_dir}") on failure. self.con is a @property decorated with @retry. Accessing self.con inside the retry loop means each failed attempt re-evaluates the property — potentially triggering VM reboots. The connection object should be retrieved once before the loop

ocs_ci/helpers/nfs_helpers.py
```diff
@@ -0,0 +1,161 @@
...
+                nfs_client_vm_name = config.ENV_DATA.get("nfs_client_vm_name")
+                if not nfs_client_vm_cloud or not nfs_client_vm_name:
+                    raise ConfigurationError(
+                        "NFS Client VM is not accessible and ENV_DATA nfs_client_vm_cloud and/or nfs_client_vm_name "
+                        "parameters are not configured to be able to automatically reboot the NFS Client VM."
+                    )
+                cmd = f"openstack --os-cloud {nfs_client_vm_cloud} server reboot --hard --wait {nfs_client_vm_name}"
+                exec_cmd(cmd)
+
+                time.sleep(60)
```
PrasadDesala: why do we need a hardcoded sleep time here?

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+        if io_errors:
+            log.warning(f"I/O errors detected: {len(io_errors)} errors")
+            for error in io_errors[:10]:  # Log first 10 errors
+                log.warning(f"  - {error}")
+            # Note: Some transient errors during reboot might be acceptable
+            # Fail only if there are persistent errors after recovery
+            if len(io_errors) > 20:
+                raise AssertionError(f"Too many I/O errors detected: {len(io_errors)}")
+        else:
+            log.info("No I/O errors detected - all operations successful!")
```
PrasadDesala: IO errors are expected during reboots? 
The threshold of 20 means the test could pass with up to 20 distinct data-integrity errors (checksum unchanged, data mismatch) in addition to the one pod-unavailability entry, masking real data corruption. The reboot errors (pod unavailable) should be explicitly separated from data-integrity errors.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+        log.info("=" * 80)
+        log.info("Starting Snapshot of Resized PVC, Restore, and Re-Resize scenario")
+        log.info("=" * 80)
+
+        # Step 1: Create snapshot of the resized cloned PVC
+        log.info("Step 1: Creating snapshot of resized cloned PVC")
+
+        cloned_snapshot_name = f"{cloned_pvc_name}-snapshot"
+        snap_yaml = constants.CSI_CEPHFS_SNAPSHOT_YAML
+        nfs_snapshotclass_name = "ocs-storagecluster-nfsplugin-snapclass"
```
PrasadDesala: should we move it to constants.py?

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+        patch_cmd = (
+            f"patch pvc {cloned_pvc_name} -p "
+            f'\'{{"spec":{{"resources":{{"requests":{{"storage":"{new_size}"}}}}}}}}\''
+        )
+        result = cloned_pvc_obj.ocp.exec_oc_cmd(patch_cmd)
+        log.info(f"PVC resize request submitted: {result}")
+
+        # Wait for PVC to be resized
+        log.info("Waiting for PVC resize to complete...")
+        time.sleep(30)  # Give some time for resize operation
```
PrasadDesala: PVC resize verification uses a bare time.sleep(30) followed by a hand rolled for attempt in range(12) polling loop, instead use TimeoutSampler.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+            current_size = cloned_pvc_obj.ocp.exec_oc_cmd(size_cmd)
+            if current_size and current_size.strip() == new_size:
+                log.info(f"PVC successfully resized to {new_size}")
+                break
+            log.info(
+                f"Waiting for resize... Current size: {current_size.strip() if current_size else 'unknown'}"
+            )
+            time.sleep(10)
+        else:
+            log.warning("PVC resize may not have completed within timeout")
```
PrasadDesala: this silently continues the test with a PVC that may not have been resized, making the subsequent checksum assertion meaningless.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+        log.info(f"Deployment {cloned_pod_name} is ready")
+
+        # Get the actual pod created by the deployment (it will have a generated suffix)
+        cloned_pod_objs = pod.get_all_pods(
+            namespace=self.namespace, selector=[cloned_pod_name], selector_label="name"
+        )
+        assert (
+            len(cloned_pod_objs) > 0
+        ), f"No pods found for deployment {cloned_pod_name}"
+        # cloned_pod_obj = cloned_pod_objs[0]
```
PrasadDesala: remove the commented line

ocs_ci/helpers/nfs_helpers.py
```diff
@@ -0,0 +1,161 @@
...
+    Test classes should inherit from this base class and will have access to:
+    - self.con: Property that returns an active connection to the NFS client VM
+    - self.get_nfs_client_connection(): Method to create a new connection
+    - self._mount_nfs_with_retry(): Method to mount NFS with retry logic
+    """
+
+    # If the connection to the NFS Client VM fails, it's possible that
+    # the NFS Client VM might not be healthy, so rebooting it and re-trying
+    @property
+    @retry((TimeoutError, socket.gaierror), tries=3, delay=60, backoff=1)
```
PrasadDesala: @retry wraps the function, but a @property is evaluated by attribute access (self.con). When the retry decorator wraps the underlying function, the @property descriptor is applied to the wrapped function, so on retry attempts the function is called as a regular function, not through the property mechanism.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+            nfs_client_user: "root"
+            nfs_client_private_key: constants.SSH_PRIV_KEY
+        these values are provided in all our automation runs in Jenkins.
+
+        But if someone will run locally, they will need to create custom config file and provide that via
+        --ocsci-conf in order to run the external nfs consume tests. Example:
+        ENV_DATA:
+            nfs_client_ip: "10.xx.xxx.xx"
+            nfs_client_user: "root"
+            nfs_client_private_key: "<path to ssh private key>"
```
PrasadDesala: is this info needed in the doc string?

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+        cls.retain_nfs_sc_name = "ocs-storagecluster-ceph-nfs-retain"
+        platform = config.ENV_DATA.get("platform", "").lower()
+        cls.is_vsphere = platform == constants.VSPHERE_PLATFORM
+        if cls.is_vsphere:
+            log.info(
+                "vSphere platform detected: out-of-cluster NFS operations will be skipped. "
+                "All I/O and data validation will be performed from in-cluster pods."
+            )
+        cls.run_id = config.RUN.get("run_id")
+        cls.test_folder = f"mnt/test_nfs_{cls.run_id}"
```
PrasadDesala: what is the use of ths?

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+
+        # Only load NFS client credentials when not on vSphere
+        if not cls.is_vsphere:
+            cls.nfs_client_private_key = os.path.expanduser(
+                config.ENV_DATA.get("nfs_client_private_key")
+                or config.DEPLOYMENT["ssh_key_private"]
+            )
+
+        # Enable nfs feature
+        log.info("----Enable nfs----")
```
PrasadDesala: we do not have a fixture for nfs setup?

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+        log.info(f"Volume name for PVC {pvc_name}: {vol_name}")
+
+        # Get NFS share details from PV
+        fetch_share_cmd = (
+            f"get pv {vol_name} " f"-o jsonpath='{{.spec.csi.volumeAttributes.share}}'"
+        )
+        share_details = self.pv_obj.exec_oc_cmd(fetch_share_cmd)
+        log.info(f"NFS share for PVC {pvc_name}: {share_details}")
+
+        return {"volume_name": vol_name, "share_details": share_details}
```
PrasadDesala: you can replace all these with 2 lines if nfs_pvc_obj is used as an arg

pv_obj = nfs_pvc_obj.backed_pv_obj
share_details = pv_obj.get()["spec"]["csi"]["volumeAttributes"]["share"]

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,2211 @@
...
+@skipif_hci_client
+@skip_for_provider_or_client_if_ocs_version("<4.19")
+@skipif_disconnected_cluster
+@skipif_proxy_cluster
+@skipif_lean_deployment
+class TestNfsExport(NFSClientTestBase):
+    """ """
+
+    @pytest.fixture(scope="class", autouse=True)
+    def setup_teardown(
```
PrasadDesala: The PR's setup_teardown fixture is line-for-line identical to the one in TestNfsEnable.setup_teardown. Since NFSClientTestBase is a new base class, the setup fixture should either live there (once) or the new test should inherit from the existing TestNfsEnable test class as a base. The identical fixture belongs in NFSClientTestBase so both test classes share it.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: We intentionally skip expected_checksum and expected_line_count here because the I/O thread was writing to the file the whole time — before, during, and after the reboot. There's no point in time where we could have frozen the file and captured a reliable baseline. If we passed an old checksum it would always fail since the file kept growing. The check still makes sure the file is readable from the pod and, on non-vSphere, that both the pod and the external NFS mount agree on the same content.

ocs_ci/helpers/nfs_helpers.py
dgaikwad: is it okay if I move code to inside` tests/ `directory?
@PrasadDesala

ocs_ci/utility/nfs_utils.py
dgaikwad: These were the original individual helpers added early in development. They were later replaced by calculate_checksum_and_lines_from_nfs_mount in the test class, which combines both into a single call and returns a structured result with success/error handling. The old functions in nfs_utils.py were never removed after that refactor. They are dead code and will be deleted.

ocs_ci/utility/nfs_utils.py
dgaikwad: These were the original individual helpers added early in development. They were later replaced by calculate_checksum_and_lines_from_nfs_mount in the test class, which combines both into a single call and returns a structured result with success/error handling. The old functions in nfs_utils.py were never removed after that refactor. They are dead code and will be deleted.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: This is mistake, II will remove it.
 Took all decorators from one of the NFS test, will ask Amrita to review once.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Yes, I/O errors are expected in this scenario — we are doing a node reboot during active I/O, so write/read failures and pod-unavailable errors will naturally occur until the NFS pod recovers. The threshold of 20 is intentional to tolerate those transient errors. That said, your point about visibility is valid, so I have updated the code to categorise the errors (pod-unavailable, integrity, transient) in both the log warning and the AssertionError message, making it clear what type of errors occurred without changing the tolerance logic.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Now Using resize_pvc function so every verification will be take caren from that code only.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: cls.test_folder is the base directory created on the external NFS client VM where the NFS export is mounted. The run_id makes it unique per test run so two runs on the same VM don't use the same directory. All other mount points in the test (-restored, -cloned, -final) are sub-folders derived from it.

ocs_ci/helpers/nfs_helpers.py
dgaikwad: It's a Python built-in, no import needed.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: No.
After the node reboot, this call confirms two things:
The file is still readable from the pod — proving the NFS PVC survived the reboot.
On non-vSphere, both the pod and the external NFS client see the same data — proving NFS consistency from inside and outside the cluster.

ocs_ci/utility/nfs_utils.py
dgaikwad: These were the original individual helpers added early in development. They were later replaced by calculate_checksum_and_lines_from_nfs_mount in the test class, which combines both into a single call and returns a structured result with success/error handling. The old functions in nfs_utils.py were never removed after that refactor. They are dead code and will be deleted.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: It's the base directory created on the external NFS client VM where the NFS exports are mounted during the test. The run_id suffix makes it unique per test run to avoid directory collisions between parallel runs.

ocs_ci/ocs/resources/pod.py
```diff
@@ -1376,14 +1376,19 @@ def check_file_existence(pod_obj, file_path):
     Returns:
         bool: True if the file exist, False otherwise
     """
+    # try:
```
vkathole: You can remove the unnecessary code

dgaikwad: Done

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: moved code in new nfs_base.py with fixture and its relevant functionality in the NFSClientTestBase, Now we are inheriting this class in the new class.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: done

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Now yes.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: No, removed.

ocs_ci/helpers/nfs_helpers.py
dgaikwad: Expected changes are done.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Removed

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Moved.

ocs_ci/helpers/nfs_helpers.py
dgaikwad: Updated code and now "Wait until SSH port is reachable before attempting connection" added.

ocs_ci/helpers/nfs_helpers.py
dgaikwad: now creating new copy of con before the loop and use new instance in the loop.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: added doc str

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Updated code with:
        log.info(f"Test case execution started: {request.node.name}")

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Done

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Done.

ocs_ci/helpers/nfs_helpers.py
dgaikwad: moved to the ocs_ci/utility/nfs_base.py

ocs_ci/helpers/nfs_helpers.py
dgaikwad: Done

dgaikwad:
> > Screenshot 2026-08-25 at 4 42 14 PM @amr1ta , intermittently we are seeing this error where the NFS StorageClass is reported as a leftover, but from the teardown logs it appears to be deleted successfully. However, on one occasion when I manually checked the cluster, the StorageClass was still present. It might be getting recreated, and I noticed its owner/reference kind was StorageClient. Have you observed this issue earlier?
> 
> @dgaikwad I have not observed this, if you think it is a repetitive incidence then please open an issue on my name I will investigate.

Okay

ocs_ci/utility/nfs_base.py
```diff
@@ -0,0 +1,404 @@
...
+            Connection: Active SSH connection to the NFS client VM
+
+        Raises:
+            ConfigurationError: If the VM is unreachable and nfs_client_vm_cloud /
+                nfs_client_vm_name are not set in ENV_DATA
+            TimeoutError: If the SSH connection cannot be established after retries
+            socket.gaierror: If hostname resolution fails
+        """
+        if (
+            not hasattr(self, "__nfs_client_connection")
```
PrasadDesala, line 72: Python mangles __nfs_client_connection to _NFSClientTestBase__nfs_client_connection inside class methods. The literal string "__nfs_client_connection" passed to hasattr is not mangled, so it will never find the attribute — meaning hasattr(...) always returns False, and a new SSH connection is established on every call to self.con

ocs_ci/utility/nfs_base.py
```diff
@@ -0,0 +1,404 @@
...
+        cls.retain_nfs_sc_name = "ocs-storagecluster-ceph-nfs-retain"
+        platform = config.ENV_DATA.get("platform", "").lower()
+        cls.is_vsphere = platform == constants.VSPHERE_PLATFORM
+        if cls.is_vsphere:
+            log.info(
+                "vSphere platform detected: out-of-cluster NFS operations will be skipped. "
+                "All I/O and data validation will be performed from in-cluster pods."
+            )
+        cls.run_id = config.RUN.get("run_id")
+        cls.test_folder = f"mnt/test_nfs_{cls.run_id}"
```
PrasadDesala: test_folder is set without a leading /, making it a relative path used as a mount point and with mkdir, findmnt, umount, and rm -rf.
All derivative paths (test_folder_for_pod, restored_test_folder, cloned_test_folder, final_restored_test_folder) appears to be inherit this problem. shoul;d NFS mount points must be absolute paths?

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,1945 @@
...
+
+        log.info(
+            "Post-recovery I/O and data integrity verified for all pods "
+            "after non-graceful cluster stop/start (ordered shutdown)"
+        )
+
+    @pytest.mark.ignore_leftover_label(constants.MON_APP_LABEL)
+    def test_nfs_export_operations_in_out_cluster(
+        self,
+        pod_factory,
```
PrasadDesala: pod_factory fixture is declared in the test signature but never used in the body

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,1945 @@
...
+        log.info(f"Test case execution started: {request.node.name}")
+        if not self.is_vsphere:
+            nfs_utils.skip_test_if_nfs_client_unavailable(self.nfs_client_ip)
+        ACCESS_MODES = [constants.ACCESS_MODE_RWX, constants.ACCESS_MODE_RWO]
+        data_sets = dict()
+        for access_mode in ACCESS_MODES:
+            data_sets[access_mode] = {}
+            unique_suffix = random.randint(1000, 9999)
+            data_sets[access_mode]["pod_name"] = f"test-pod-outcluster-{unique_suffix}"
+            data_sets[access_mode]["pvc_name"] = f"test-pvc-{unique_suffix}"
```
PrasadDesala, line 309: RWO volumes are intended for single-node access, are we simultaneously mounting from two different clients (in-cluster pod on one node, NFS client VM as another node) here?

ocs_ci/utility/nfs_utils.py
```diff
@@ -7,6 +7,7 @@
 import logging
 import yaml
 import pytest
+import time
```
PrasadDesala, line 10: import time is added without a blank line separating it from the next from import group

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,1945 @@
...
+        e) Snapshot the resized clone, restore it, write I/O, resize to 20Gi, verify integrity.
+        f) Non-graceful cluster stop/start (ordered shutdown, force=True; workers first,
+           then control-plane), start in reverse order; verify Ceph health, NFS mounts
+           accessible, and post-recovery I/O from all pods.
+        """
+
+        log.info(f"Test case execution started: {request.node.name}")
+        if not self.is_vsphere:
+            nfs_utils.skip_test_if_nfs_client_unavailable(self.nfs_client_ip)
+        ACCESS_MODES = [constants.ACCESS_MODE_RWX, constants.ACCESS_MODE_RWO]
```
PrasadDesala: it is a local variable inside the test method. should we rename it to lowercase?

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,1945 @@
...
+        success = len(errors) <= max_errors
+        if errors:
+            log.warning(
+                f"{len(errors)} write error(s) on {file_path} "
+                f"({'tolerated' if success else 'exceeded threshold'},"
+                f" max_errors={max_errors}): {errors}"
+            )
+        return success, errors
+
+    def calculate_checksum_and_lines_from_nfs_mount(self, con, file_path):
```
PrasadDesala: (write_io_to_single_file, calculate_checksum_and_lines_from_nfs_mount, etc.) are all defined inside TestNfsExport but they're generic enough to belong in NFSClientTestBase or other helpers files.
Should we move these out from the test?

ocs_ci/utility/nfs_base.py
```diff
@@ -0,0 +1,404 @@
...
+
+        Raises:
+            TimeoutError: If connection cannot be established
+            socket.gaierror: If hostname resolution fails
+        """
+        log.info("Connecting to nfs client test VM")
+        tries = 3 if re_try else 1
+
+        @retry((TimeoutError, socket.gaierror), tries=tries, delay=60, backoff=1)
+        def __make_connection():
```
PrasadDesala: since this is inside a func, should we name it with a single underscore to be cleaner?

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,1945 @@
...
+                        log.info(
+                            f"✓ Pod {pname} data confirmed on NFS mount {mount_point}"
+                        )
+
+            except Exception as e:
+                log.error(
+                    f"✗ Non-graceful stop/start (ordered shutdown) post-recovery "
+                    f"I/O failed for pod {pname}: {e}"
+                )
+                raise
```
PrasadDesala: If the first pod in the list fails, subsequent pods are never tested. should we rather collect failures for all pods and assert at the end ?

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,1945 @@
...
+@skip_for_provider_or_client_if_ocs_version("<4.19")
+@skipif_disconnected_cluster
+@skipif_proxy_cluster
+@skipif_lean_deployment
+class TestNfsExport(NFSClientTestBase):
+    """
+    Test NFS export operations with in-cluster and out-of-cluster access.
+    """
+
+    def write_io_to_single_file(
```
PrasadDesala: You seems to be doing very aggressively. I can grow file with potentially thousands of iterations and this incurs substantial per-iteration overhead. should we consider sampling the checksum every N iterations rather than on every write?

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Yes.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Moving below generic methods.
write_io_to_single_file
calculate_checksum_and_lines_from_nfs_mount
calculate_checksum_and_lines_from_pod
verify_data_integrity
verify_data_integrity_from_pod
get_nfs_export_details
mount_nfs_export
continuous_io_operations

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,1945 @@
...
+        Args:
+            pvc_obj (PVC): PVC object
+
+        Returns:
+            str: NFS share path
+        """
+        pv_obj = pvc_obj.backed_pv_obj
+        return pv_obj.get()["spec"]["csi"]["volumeAttributes"]["share"]
+
+    def mount_nfs_export(self, con, share_details, mount_point):
```
vkathole: If possible move these functions to common file

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,1945 @@
...
+from ocs_ci.utility.nfs_utils import (
+    frame_deployment_config,
+)
+from ocs_ci.utility import nfs_utils
+from ocs_ci.utility.utils import ceph_health_check
+
+log = logging.getLogger(__name__)
+
+
+@magenta_squad
```
vkathole, line 51: add system_test marker and polarion id

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,1945 @@
...
+            "PVC Clone, Resize, and Data Integrity scenario completed successfully!"
+        )
+        log.info("=" * 80)
+        log.info(
+            "Scenario: Snapshot Resized PVC, Restore, Resize Again, Verify Integrity"
+        )
+        log.info("=" * 80)
+
+        snap_yaml = constants.CSI_CEPHFS_SNAPSHOT_YAML
+        nfs_snapshotclass_name = constants.NFS_SNAPSHOT_CLASS_NAME
```
vkathole: these are already declared at https://github.com/red-hat-storage/ocs-ci/pull/15788/changes#diff-8665e3d35eadb8813aeba041d8671c5ab23128f1636096ddc1df4fc5821b8717R1139 , Could we avoid redeclaring them here and keep this simpler by using the constants directly where needed?
` log.info("Step 2: Creating snapshot of NFS PVC")

        snap_yaml = constants.CSI_CEPHFS_SNAPSHOT_YAML
        nfs_snapshotclass_name = constants.NFS_SNAPSHOT_CLASS_NAME`

dgaikwad: Done.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Added new marker "system". From where I can get polarion id?
@vkathole

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Done

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
vkathole: You have create new polarion id for this test case if not done already, ex. https://polarion.engineering.redhat.com/polarion/#/project/OpenShiftContainerStorage/workitem?id=OCS-4852

dgaikwad: Done. Added.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Now doing after 10 iteration.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Done

ocs_ci/utility/nfs_base.py
dgaikwad: Updated.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Updated.

ocs_ci/utility/nfs_utils.py
dgaikwad: Added Space.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Removed.

ocs_ci/utility/nfs_base.py
dgaikwad: Done.

ocs_ci/utility/nfs_base.py
dgaikwad: Done

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: Updated the code. For RWO volumes, the volume is now mounted only inside the pod. We mount the RWO volume from both the pod and the NFS client and perform the corresponding data integrity checks accordingly.

ocs_ci/utility/nfs_base.py
PrasadDesala: this comment doesn't seem to be incorporated

ocs_ci/ocs/resources/pod.py
```diff
@@ -1377,13 +1377,14 @@ def check_file_existence(pod_obj, file_path):
         bool: True if the file exist, False otherwise
     """
     try:
-        check_if_executable_in_path(pod_obj.exec_cmd_on_pod("which find"))
+        ret = pod_obj.exec_cmd_on_pod(
+            f"bash -c 'ls {shlex.quote(file_path)}'", out_yaml_format=False
+        )
+        if file_path in ret:
+            return True
+        return False
```
PrasadDesala, line 1385: why was this func updated? the old implementation checked whether find returned a match,  a filesystem-level search. The new code runs ls and then checks if file_path in ret, meaning if ls succeeds but ret is empty (e.g. ls output is consumed differently by exec_cmd_on_pod), the function returns False silently. It can cause a regression to the entire function

ocs_ci/utility/nfs_base.py
```diff
@@ -0,0 +1,869 @@
...
+            f"Mounting NFS export: mount -t nfs {mount_options} {export_path} {mount_point}"
+        )
+
+        platform = config.ENV_DATA.get("platform", "").lower()
+        if platform == constants.IBMCLOUD_PLATFORM:
+            log.info(
+                "IBM Cloud: waiting 30s before mount to ensure security group rules "
+                "and DNS resolution are fully active..."
+            )
+            time.sleep(30)
```
PrasadDesala, line 712: replace sleep at all places with a TimeoutSampler that is more robust and faster

ocs_ci/utility/nfs_base.py
```diff
@@ -0,0 +1,869 @@
...
+                return
+
+            msg = (
+                f"Mount attempt {attempt + 1} failed (rc={retcode}): {stderr or stdout}"
+            )
+            if attempt == retries - 1:
+                raise CommandFailed(msg)
+
+            log.warning(f"{msg}. Retrying in 10 seconds...")
+            con.exec_cmd(f"umount -f {mount_dir}")
```
PrasadDesala, line 234: you are looping it but not checking the return code. If umount itself fails (e.g. directory never actually mounted), subsequent mount attempts will hit a stale or dirty mount point.

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
```diff
@@ -0,0 +1,1512 @@
...
+                    log.warning(f"[{am}] Failed to unmount restored NFS: {e}")
+
+        if not self.is_vsphere:
+            request.addfinalizer(cleanup_restored_mount)
+
+        log.info("NFS Snapshot and Restore scenario completed successfully!")
+        log.info("=" * 80)
+        log.info("Scenario: Clone Restored PVC, Resize, and Verify Data Integrity")
+        log.info("=" * 80)
+        log.info("Step 1: Creating clone of restored PVC")
```
vkathole, line 907: Correct the log info step numbers, it has log.info("Step 5: Verifying data integrity for restored PVC") then again started with Step 1, and similarly followed in rest of the code 4 times as well, it will create a confusion. If you want to keep number then give something like 1.1,1.2,... then 2.1,2.2,...

ocs_ci/utility/nfs_base.py
dgaikwad: ---

## `time.sleep` audit — replaced with `TimeoutSampler` where applicable

### ✅ Replaced with `TimeoutSampler`

**1. Mount retry backoff — `nfs_base.py` `_mount_nfs_with_retry`**

```python
# Before
time.sleep(10)  # fixed wait between mount attempts

# After
for sample in TimeoutSampler(timeout=retries * 15, sleep=10, func=con.exec_cmd, command=mount_cmd):
    retcode, stdout, stderr = sample
    if retcode == 0:
        return
    ...
```
Condition polled: `con.exec_cmd(mount_cmd)` returns `retcode == 0`. Exits as soon as mount succeeds rather than always sleeping 10 s.

---

**2. Pod container unavailable backoff — `nfs_base.py` `continuous_io_operations`**

```python
# Before
if "container not found" in exc_str:
    time.sleep(10)
    continue

# After
if "container not found" in exc_str:
    for _ in TimeoutSampler(timeout=600, sleep=10, func=pod_obj.exec_sh_cmd_on_pod, command="echo ready"):
        break
    continue
```
Condition polled: `exec_sh_cmd_on_pod` succeeds without raising `container not found`. Exits as soon as the container is reachable.

---

**3. NFS export release after unmount — `test_nfs_export_in_out_cluster.py`**

```python
# Before
nfs_utils.unmount(con, test_folder)
time.sleep(10)

# After
nfs_utils.unmount(con, test_folder)
for _ in TimeoutSampler(timeout=60, sleep=5, func=lambda: con.exec_cmd(f"findmnt -M {test_folder}")[0] != 0):
    break
```
Condition polled: `findmnt -M {test_folder}` returns non-zero (mount no longer active).

---

**4. I/O thread initialisation — `test_nfs_export_in_out_cluster.py`**

```python
# Before
time.sleep(10)
log.info("Initial I/O operations running successfully")

# After
for _ in TimeoutSampler(timeout=60, sleep=2,
    func=lambda: all(data_sets[am]["io_thread"].is_alive() and len(data_sets[am]["io_errors"]) == 0 for am in access_modes)):
    break
log.info("Initial I/O operations running successfully")
```
Condition polled: all I/O threads alive with zero errors.

---

### ❌ Cannot replace — reason per location

| Location | Sleep | Reason |
|---|---|---|
| `nfs_base.py` — `write_io_to_single_file` loop | `time.sleep(delay)` | Intentional I/O pacing between write iterations — no condition to poll, the sleep IS the feature |
| `nfs_base.py` — `continuous_io_operations` loop | `time.sleep(2)` | Deliberate throttle between background I/O iterations — removing it changes write frequency semantics |
| `nfs_base.py` — `mount_nfs_export` IBM Cloud block | `time.sleep(30)` | IBM Cloud SG rule enforcement has no queryable "active" signal — empirical propagation buffer |
| `nfs_utils.py` — after `configure_nfs_lb_security_group()` | `time.sleep(60)` | Same — IBM Cloud SG propagation is asynchronous with no observable done-state |
| `test_nfs_export_in_out_cluster.py` — post-recovery soak | `time.sleep(60)` | Deliberate soak to prove I/O stability over 60 s, not just one successful write — `TimeoutSampler` exits on first success which defeats the purpose |
| `test_nfs_export_in_out_cluster.py` — after force-stop all nodes | `time.sleep(300)` | After masters are force-stopped the Kubernetes API is gone — `oc get nodes` fails with `connection refused` so no `TimeoutSampler` condition function can work. No platform-agnostic "node powered off" signal exists on AWS/vSphere/IBM Cloud |

ocs_ci/utility/nfs_base.py
dgaikwad: The umount return code is now captured and checked. If it fails, a WARNING is logged with the exact error so you can see in the logs whether the mount point was dirty going into the next attempt. The loop continues regardless — intentionally — because the subsequent mount attempt may still succeed (e.g. if umount failed with not mounted it means the slate is already clean). The important thing is that the failure is now visible rather than silently ignored.

ocs_ci/utility/nfs_base.py
dgaikwad: Done. What changed and why:

- **Removed** `hasattr(self, "__nfs_client_connection")` — string `"__nfs_client_connection"` is never mangled so it always returned `False`, causing a new SSH connection on every `self.con` access
- **Replaced with** `getattr(self, "_NFSClientTestBase__nfs_client_connection", None)` — uses the correctly mangled name with a `None` default to safely handle the first call before the attribute is set
- `self.__nfs_client_connection` on lines below **stays as-is** — those are source-code identifiers so Python mangles them correctly at compile time

tests/cross_functional/system_test/test_nfs_export_in_out_cluster.py
dgaikwad: To remove confusion removed steps from the logger, Keeping scenario 1, 2, 3 like these only

ocs_ci/ocs/resources/pod.py
dgaikwad: Actually, we were using this code earlier, and at that time we considered updating it. However, the update is no longer needed, so we are reverting these changes.

PrasadDesala (approved):

vkathole (approved):

PR #15794, opened 2026-07-23
INTEROP-9285: Filter non-ODF leftovers in environment_check

amp-rh:
Add provisioner detection helper and filtering logic to environment_check.py so that PVs/PVCs from non-ODF CSI drivers and StorageClient-owned StorageClasses are excluded from the pre/post comparison, preventing false-positive ResourceLeftoversException errors.

In tests/conftest.py, add StorageClient ownerReference stripping before StorageClass deletion and NotFound tolerance so the finalizer does not fail when the SC is already removed.

Supersedes #15714, #15576, and #15157.

Tracked in: [INTEROP-9285](https://redhat.atlassian.net/browse/INTEROP-9285)

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined environment snapshot/leftover checks to better identify which PV/PVC resources to consider using CSI provisioner detection.
  * Improved filtering so StorageClass resources that are StorageClient-owned are excluded from leftover comparisons.
* **Tests**
  * Improved StorageClass teardown reliability by removing only StorageClient-related owner references, tolerating already-deleted resources, and surfacing unexpected teardown failures.

### DIFF
--- a/ocs_ci/utility/environment_check.py
+++ b/ocs_ci/utility/environment_check.py
@@ -13,6 +13,35 @@ from ocs_ci.ocs import ocp, defaults, constants, exceptions
 log = logging.getLogger(__name__)
 
 
+def _pv_or_pvc_storage_provisioner(item):
+    """
+    Extract the CSI provisioner driver name from a PV or PVC item.
+
+    Args:
+        item (dict): Kubernetes resource dictionary (PV or PVC)
+
+    Returns:
+        str: The provisioner driver name, or None if not detected
+    """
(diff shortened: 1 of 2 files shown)


amp-rh:
Closing: the OCS acceptance step (interop-tests-ocs-tests) is being replaced with an ODF health check (openshift/release#82999 / [INTEROP-9384](https://redhat.atlassian.net/browse/INTEROP-9384)), so the environment_check code path will no longer execute in OPP jobs. This fix is no longer needed.

PR #15804, opened 2026-07-24
fix active mgr pod issue

### LINKED ISSUE #15802: test_monitoring_after_rebooting_node_where_mgr_is_running fails sometimes with AssertionError
Root Cause:
The test `test_monitoring_after_rebooting_node_where_mgr_is_running` fails because the verification function `wait_to_update_mgrpod_info_prometheus_pod` checks for Prometheus metrics from the wrong MGR pod. After the node reboot triggers MGR failover, the active MGR changes from pod 'a' to pod 'b'. However, the verification function retrieves the first pod from the MGR pod list (which happens to be the standby MGR-a) instead of identifying the active MGR. Since the `ceph_health_status` metric is only exported by the active MGR pod, the check fails when it looks for metrics from the standby pod. The test correctly identifies the active MGR before the reboot but fails to do so in the verification function.

RP Links: 
https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/all/47201/2107908/2107968/log
https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/all/49457/2175109/2175167/log

vkathole:
Fixes https://github.com/red-hat-storage/ocs-ci/issues/15802

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of the active Ceph manager pod.
  * Monitoring workflows now reliably target the correct active manager pod after node reboots and in negative scenarios.
  * Increased reliability of Prometheus/metric validation tied to Ceph health status.

### DIFF
--- a/ocs_ci/ocs/resources/pod.py
+++ b/ocs_ci/ocs/resources/pod.py
@@ -1787,6 +1787,29 @@ def get_mgr_pods(mgr_label=constants.MGR_APP_LABEL, namespace=None):
     return mgr_pods
 
 
+def get_active_mgr_pod():
+    """
+    Get the active MGR pod using ceph mgr stat
+
+    Returns:
+        Pod: The active MGR pod object
+
+    Raises:
+        ValueError: If the active MGR pod is not found
+
+    """
+    toolbox = get_ceph_tools_pod()
+    active_mgr_pod_output = toolbox.exec_cmd_on_pod("ceph mgr stat")
+    active_mgr_pod_suffix = active_mgr_pod_output.get("active_name")
+    logger.info(f"The active MGR pod suffix is {active_mgr_pod_suffix}")
+
+    for mgr_pod in get_mgr_pods():
+        if active_mgr_pod_suffix in mgr_pod.name:
+            logger.info(f"Active MGR pod: {mgr_pod.name}")
+            return mgr_pod
+    raise ValueError(f"Active MGR pod with suffix '{active_mgr_pod_suffix}' not found")
+
+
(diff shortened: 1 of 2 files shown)


ocs_ci/ocs/resources/pod.py
```diff
@@ -1787,6 +1787,29 @@ def get_mgr_pods(mgr_label=constants.MGR_APP_LABEL, namespace=None):
...
+    Returns:
+        Pod: The active MGR pod object
+
+    Raises:
+        ValueError: If the active MGR pod is not found
+
+    """
+    toolbox = get_ceph_tools_pod()
+    active_mgr_pod_output = toolbox.exec_cmd_on_pod("ceph mgr stat")
+    active_mgr_pod_suffix = active_mgr_pod_output.get("active_name")
```
PrasadDesala, line 1803: active_mgr_pod_suffix could be None, leading to a misleading ValueError if the key is missing for any reason. If no active_mgr_pod_suffix found raise an error that it could not determine active MGR from 'ceph mgr stat' output

ocs_ci/ocs/resources/pod.py
```diff
@@ -1787,6 +1787,29 @@ def get_mgr_pods(mgr_label=constants.MGR_APP_LABEL, namespace=None):
     return mgr_pods
 
 
+def get_active_mgr_pod():
```
PrasadDesala, line 1790: Should we consider making the function more resilient with a retry? get_active_mgr_pod() could also failtransiently if called during an MGR failover and adding retry would help in those cases

PR #15810, opened 2026-07-24
Fix OCP release image URL for OCP 5.0+

petr-balogh:
OCP 5.x release images are published under ocp/release-5 instead of ocp/release. Update the upgrade config, add a version-aware mirror path helper, and use it in disconnected mirroring so both standard and disconnected upgrades resolve the correct image repository.

## Summary by CodeRabbit

* **New Features**
  * Enhanced disconnected image mirroring to use version-specific OpenShift release mirror subpaths.
  * OpenShift 5 references now mirror from the `release-5` location when applicable, including digest-based scenarios.
* **Bug Fixes**
  * Added a vSphere UPI workaround to restore missing `platform-type` node labels after deployment for OCP 4.22+.
* **Configuration**
  * Updated the OCP 5.0 upgrade configuration to use the dedicated `release-5` release image path.

### DIFF
--- a/ocs_ci/utility/utils.py
+++ b/ocs_ci/utility/utils.py
@@ -1330,9 +1330,29 @@ def get_registry_svc(version):
 
     """
     major = version.split(".")[0]
-    if int(major) >= 5:
-        return "registry.ci.openshift.org/ocp/release-5"
+    if int(major) == 5:
+        return constants.REGISTRY_SVC_V5
     return constants.REGISTRY_SVC
+
+
+def get_ocp_release_mirror_path(version):
+    """
+    Get the OCP release image mirror path for the given OCP version.
+
+    OCP 5.x release images are published under ocp/release-5
+    instead of ocp/release.
+
+    Args:
+        version (str): OCP version string (e.g. "5.0.0-0.nightly-2026-06-18-000016")
+
+    Returns:
+        str: Mirror path (e.g. "ocp/release-5")
+
+    """
+    major = version.split(".")[0]
+    if int(major) == 5:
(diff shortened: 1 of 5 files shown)


vavuthu (dismissed):

petr-balogh:
Triggered verification:
https://url.corp.redhat.com/a20323c

ocs_ci/utility/utils.py
```diff
@@ -1331,10 +1331,30 @@ def get_registry_svc(version):
...
+
+    Args:
+        version (str): OCP version string (e.g. "5.0.0-0.nightly-2026-06-18-000016")
+
+    Returns:
+        str: Mirror path (e.g. "ocp/release-5")
+
+    """
+    major = version.split(".")[0]
+    if int(major) >= 5:
```
dahorak: The `>=` doesn't make much sense here, when we are setting the `release-5` in the path... More meaningful would be `==`, but I know, that version 6 is not something we should consider now, so probably no need to change it.

ocs_ci/utility/utils.py
```diff
@@ -1331,10 +1331,30 @@ def get_registry_svc(version):
     """
     major = version.split(".")[0]
     if int(major) >= 5:
```
dahorak: Same as bellow, the `>=` doesn't make much sense in this condition - I would use `==`.

dahorak (dismissed):
Overall LGTM.

vavuthu (approved):

dahorak (approved):

petr-balogh:
I see OCP upgrade passed so marking as verified

PR #15828, opened 2026-07-28
fix(ui): update ACM perspective-switcher locators for OCP 4.22

prsurve:
In OCP 4.22, the perspective-switcher dropdown menu no longer uses "local-cluster" as the perspective label. The local OCP cluster perspective is now labelled "Core platform", and menu options are rendered as <li data-test-id="perspective-switcher-menu-option"> elements inside the toggle button DOM.

Update acm_configuration_4_22 in views.py:
- local-cluster_dropdown_item: target the "Core platform" <button role="option"> menu item instead of searching for "local-cluster" text inside the toggle button
- local-cluster_dropdown: detect the active "Core platform" label in the toggle button instead of the now-absent <h2>Administrator

Fixes setup_acm_ui fixture failure in navigate_to_local_cluster() which timed out waiting for a "local-cluster" element that no longer exists in the ACM console on OCP 4.22.3.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated perspective switcher targeting for OpenShift Container Platform 4.22 to reliably select the **Core platform** perspective and correct menu options (with sensible fallbacks).
  * Improved ACM console login dropdown detection by recognizing an additional **Core platform** option and refining the “no dropdown found” messaging.
  * Enhanced navigation from OpenShift to ACM cluster pages with a dedicated **Fleet Management** path.
  * Ensured disaster-recovery navigation supports accurate DOM copying when validating cleanup alerts.

### DIFF
--- a/ocs_ci/ocs/ui/views.py
+++ b/ocs_ci/ocs/ui/views.py
@@ -1867,30 +1867,32 @@ acm_configuration_4_22 = {
         "//button[@data-test-id='perspective-switcher-toggle']",
         By.XPATH,
     ),
-    # In OCP 4.22 the perspective-switcher menu renders options as <li data-test-id="perspective-switcher-menu-option">
-    # inside the toggle button DOM. The local OCP cluster perspective is labelled "Core platform".
     "local-cluster_dropdown_item": (
         "//li[@data-test-id='perspective-switcher-menu-option']"
-        "//button[@role='option' and .//h2[normalize-space()='Core platform']]",
+        "//button[@role='option' and .//h2[normalize-space()='Core platform']] | "
+        "//span[contains(@class, 'c-menu__item-text') and text()='local-cluster']/.. | "
+        "//h2[normalize-space()='Administrator']",
         By.XPATH,
     ),
-    # When already on the local OCP console the toggle button shows "Core platform" as the active perspective.
     "local-cluster_dropdown": (
         "//button[@data-test-id='perspective-switcher-toggle']"
-        "//*[normalize-space()='Core platform']",
+        "//*[normalize-space()='Core platform'] | "
+        "//h2[text()='local-cluster'] | "
+        "//span[contains(@class, 'c-menu-toggle__text') and text()='local-cluster']/.. | "
+        "//h2[normalize-space()='Fleet Management'] | "
+        "//h2[normalize-space()='Administrator']",
         By.XPATH,
     ),
-    # OCP 4.22: detects that the current perspective is "Core platform" (local OCP cluster).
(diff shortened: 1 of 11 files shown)


ocs_ci/ocs/ui/views.py
```diff
@@ -1867,12 +1867,30 @@
         "//button[@data-test-id='perspective-switcher-toggle']",
         By.XPATH,
     ),
+    # In OCP 4.22 the perspective-switcher menu renders options as <li data-test-id="perspective-switcher-menu-option">
```
DanielOsypenko: please remove new comments here and bellow

ocs_ci/ocs/ui/views.py
```diff
@@ -1867,12 +1867,30 @@
...
     "local-cluster_dropdown_item": (
-        "//button[@data-test-id='perspective-switcher-toggle']//*[normalize-space()='local-cluster']",
+        "//li[@data-test-id='perspective-switcher-menu-option']"
+        "//button[@role='option' and .//h2[normalize-space()='Core platform']]",
         By.XPATH,
     ),
+    # When already on the local OCP console the toggle button shows "Core platform" as the active perspective.
     "local-cluster_dropdown": (
-        "//h2[normalize-space()='Administrator']",
+        "//button[@data-test-id='perspective-switcher-toggle']"
```
DanielOsypenko: in this case we should preserve old locator and add a new one with "|"
this is necessary for Upgrade scenario on both pre-change and post-change versions

ocs_ci/ocs/ui/views.py
```diff
@@ -1867,12 +1867,30 @@
         "//button[@data-test-id='perspective-switcher-toggle']",
         By.XPATH,
     ),
+    # In OCP 4.22 the perspective-switcher menu renders options as <li data-test-id="perspective-switcher-menu-option">
+    # inside the toggle button DOM. The local OCP cluster perspective is labelled "Core platform".
     "local-cluster_dropdown_item": (
-        "//button[@data-test-id='perspective-switcher-toggle']//*[normalize-space()='local-cluster']",
+        "//li[@data-test-id='perspective-switcher-menu-option']"
```
DanielOsypenko: in this case we should preserve old locator and add a new one with "|"
this is necessary for Upgrade scenario on both pre-change and post-change versions

ocs_ci/ocs/ui/views.py
DanielOsypenko: now we have too much ) there should be 2 locators but we have 6 instead

prsurve: 🤣 let me fix it

PR #15837, opened 2026-07-28
W/A for bug OCPBUGS-100052

petr-balogh:
On OCP 4.22+ vSphere UPI deployments, nodes are missing the node.openshift.io/platform-type=vsphere label which is required for the upgrade to proceed. Label all nodes with the platform-type in post_ocp_deploy as a workaround until the bug is fixed.

Slack thread with details:
https://redhat-internal.slack.com/archives/CBQHQFU0N/p1785223848177199

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -1071,6 +1071,22 @@ class Deployment(object):
         )
         if ibmcloud_ipi:
             ibmcloud.label_nodes_region()
+        vsphere_upi = (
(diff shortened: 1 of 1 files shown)


petr-balogh:
Verification will be done here:
https://url.corp.redhat.com/6da42e4

As the same commit is also in master PR.

vavuthu (approved):

dahorak (approved):

petr-balogh:
I see OCP upgrade passed so marking as verified

PR #15904, opened 2026-08-05
Fix TLS scan RGW endpoint discovery and skip on disconnected clusters

AviadP:
## Summary
- Fix RGW endpoint discovery when pod spec has no `ports` — fall back to well-known ports (8080, 443)
- Skip TLS profile tests on disconnected clusters (scanner needs internet access)

## Test plan
- [x] Validated locally: wildcard selector (MCG + RGW)
- [x] Validated locally: noobaa.io selector
- [x] Validated locally: ceph.rook.io (RGW) selector

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RGW TLS endpoint discovery with HTTP and HTTPS fallback ports when container ports are not declared.
  * Preserved explicitly declared container ports when available.
  * Improved endpoint selection across RGW components.

* **Test Improvements**
  * Centralized TLS profile tests are now skipped on disconnected clusters where they cannot run successfully.

### DIFF
--- a/ocs_ci/helpers/tlsprofile_helper.py
+++ b/ocs_ci/helpers/tlsprofile_helper.py
@@ -303,6 +303,55 @@ def _tls_scan_run_oc(args, kubeconfig=None, timeout=60):
     return completed.stdout.decode()
 
 
+def _find_fallback_ports(selector, component, pod):
+    fallback_ports = selector.get("fallback_ports", [])
+    if fallback_ports or component != "all":
+        return fallback_ports
+    pod_labels = pod.get("metadata", {}).get("labels", {})
+    for comp_sel in TLS_SCAN_COMPONENT_SELECTORS.values():
+        comp_label = comp_sel.get("label", "")
+        comp_fb = comp_sel.get("fallback_ports", [])
+        if comp_fb and "=" in comp_label:
+            key, val = comp_label.split("=", 1)
+            if pod_labels.get(key) == val:
+                return comp_fb
+    return []
+
+
+def _build_container_endpoints(container, pod_name, pod_ns, pod_ip, fallback_ports):
+    c_name = container["name"]
+    cmd_parts = container.get("command", []) + container.get("args", [])
+    process = ""
+    if cmd_parts:
+        process = cmd_parts[0].rsplit("/", 1)[-1][:15]
+    if not process:
+        process = container.get("image", "").split("/")[-1].split(":")[0][:15]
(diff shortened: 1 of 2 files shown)


AviadP:
Fixed in b18fbe6c6. When `component="all"`, the discovery now checks each pod's labels against component-specific selectors that define `fallback_ports`, so RGW pods without declared ports are correctly discovered in `all` scans too.

ocs_ci/helpers/tlsprofile_helper.py
```diff
@@ -328,6 +334,17 @@ def _tls_scan_discover_endpoints(kubeconfig, namespaces, component="all"):
...
+            if not fallback_ports and component == "all":
+                pod_labels = pod.get("metadata", {}).get("labels", {})
+                for comp_sel in TLS_SCAN_COMPONENT_SELECTORS.values():
+                    comp_label = comp_sel.get("label", "")
+                    comp_fb = comp_sel.get("fallback_ports", [])
+                    if comp_fb and "=" in comp_label:
+                        key, val = comp_label.split("=", 1)
+                        if pod_labels.get(key) == val:
+                            fallback_ports = comp_fb
+                            break
```
yitzhak12: I would move the code here to a private function with a name like `_find_fallback_ports`

ocs_ci/helpers/tlsprofile_helper.py
```diff
@@ -338,19 +355,32 @@ def _tls_scan_discover_endpoints(kubeconfig, namespaces, component="all"):
...
+                ]
+                if not port_numbers and fallback_ports:
+                    log.info(
+                        "TLS scan: pod %s container %s has no declared ports, "
+                        "using fallback ports %s",
+                        pod_name,
+                        c_name,
+                        fallback_ports,
+                    )
+                    port_numbers = fallback_ports
```
yitzhak12: I think we can also move this to a private function here. Because it's inside a for loop, and it's a little long.

ocs_ci/helpers/tlsprofile_helper.py
AviadP: fixed

ocs_ci/helpers/tlsprofile_helper.py
AviadP: fixed

yitzhak12 (approved):
Looks good.

suchita-g (approved):
LGTM

ebenahar (approved):

Issue #13242, opened 2025-09-18
We may need to change the way we check the crushDeviceClass in the Cephcluster


yitzhak12:
See the related failure here: https://reportportal-ocs4.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#ocs/launches/954/37446/1770944/1771152/log. 
In the [cephcluster](http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/j045gism1c33t1/j045gism1c33t1_20250911T082741/logs/failed_testcase_ocs_logs_1757598714952/test_resize_osd_ocs_logs/j045gism1c33t1/ocs_must_gather/quay-io-rhceph-dev-odf4-odf-must-gather-rhel9-sha256-1e405884eadb3a98cdfaa2768712975ffbb87f55fc2b0d925333075a3276426d/ceph/namespaces/openshift-storage/ceph.rook.io/cephclusters/ocs-storagecluster-cephcluster.yaml) and in the [Ceph OSD tree](http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/j045gism1c33t1/j045gism1c33t1_20250911T082741/logs/failed_testcase_ocs_logs_1757598714952/test_resize_osd_ocs_logs/j045gism1c33t1/ocs_must_gather/quay-io-rhceph-dev-odf4-odf-must-gather-rhel9-sha256-1e405884eadb3a98cdfaa2768712975ffbb87f55fc2b0d925333075a3276426d/ceph/must_gather_commands/ceph_osd_tree), I see the device crush `europe-west1-b`, but in the [storagecluster](http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/j045gism1c33t1/j045gism1c33t1_20250911T082741/logs/failed_testcase_ocs_logs_1757598714952/test_resize_osd_ocs_logs/j045gism1c33t1/ocs_must_gather/quay-io-rhceph-dev-odf4-odf-must-gather-rhel9-sha256-1e405884eadb3a98cdfaa2768712975ffbb87f55fc2b0d925333075a3276426d/namespaces/openshift-storage/oc_output/storagecluster), I don't see the this device. Is it a valid behaviour?

ebenahar:
@yitzhak12 please tag the relevant person to answer the query

ebenahar:
@yitzhak12 to provide update this week (Oct 29th)

yitzhak12:
[...]

PR #15917, opened 2026-08-09
Fix device class verification to skip operator-generated CephCluster sets

### LINKED ISSUE #14697: [Improvement]: "verify_storage_device_class" fails on arbiter clusters due to operator-generated zone device sets
**Description:**

On arbiter deployments that use failureDomain: zone, the ODF operator automatically creates additional storageClassDeviceSets in the CephCluster CR. These device sets are named after the zone (for example: us-south-1 or europe-west1-b). For these operator-created device sets, the operator sets crushDeviceClass to the zone name instead of using the expected value "ssd".

The `verify_storage_device_class` function checks all storageClassDeviceSets in the CephCluster
and expects crushDeviceClass to be "ssd". Because of this, the check fails on arbiter clusters,
even though the configuration is correct.

Example failure:

> AssertionError: crushDeviceClass for us-south-1 is set to us-south-1, but it should be set to ssd

Affected platforms:

- IBM Cloud arbiter clusters (zones: us-south-1, us-south-2, us-south-3)
- GCP arbiter clusters (zones: europe-west1-b, europe-west1-c, europe-west1-d)
- Any arbiter deployment that uses zone-based failure domains

Root cause:

- The CephCluster ends up with 6 storageClassDeviceSets:
  - 3 defined by the StorageCluster, with crushDeviceClass set to "ssd"
  - 3 created by the operator for arbiter, with crushDeviceClass set to the zone name
- The verify_storage_device_class function does not distinguish between user-defined
  device sets and operator-generated ones
- The same problem happens later in the function when validating deviceClasses, where
  zone-based values (for example us-south-1) also cause the assertion to fail

Fix:

- Add a helper function (`filter_user_defined_device_sets`) that only selects device sets
  defined in the StorageCluster, and skips operator-generated ones
- While checking device classes, track only the expected values and ignore device classes
  that come from operator-generated arbiter device sets

yitzhak12:
## Description
  On clusters where the ODF operator automatically creates additional CephCluster `storageClassDeviceSets` (such as arbiter/zone topologies and non-resilient pools), these sets use zone or rack names like `europe-west1-b` or `rack0` as their `crushDeviceClass` instead of `ssd`. The functions `verify_storage_device_class` and `verify_device_class_in_osd_tree` previously checked all device sets and classes within the CephCluster, which led to false failures during post-resize or autoscaler verification, even when the cluster remained healthy. This update filters CephCluster device sets to only those matching StorageCluster names and excludes operator-generated device classes from status checks and from the `ceph osd tree` output.
  fix https://github.com/red-hat-storage/ocs-ci/issues/14697
  Related: https://github.com/red-hat-storage/ocs-ci/issues/13242

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved post-resize storage validation when multiple device sets are present.
  * Device-class checks now focus on user-defined storage configurations and ignore operator-generated resources.
  * Enhanced validation for arbiter, zone, and non-resilient pool configurations.
  * OSD-tree checks now correctly exclude internally generated device classes, reducing false validation failures.
  * Improved handling of storage device sets with multi-digit suffixes.


amr1ta (approved):

ocs_ci/ocs/resources/storage_cluster.py
```diff
@@ -1546,79 +1550,180 @@ def verify_storage_device_class(
...
+            msg = (
+                f"tuneFastDeviceClass for {device_set_name}"
+                f" is set to {tune_fast_device_class}"
+            )
             log.debug(msg)
-            assert (
-                tune_fast_device_class
-            ), f"{msg} when {constants.DEVICECLASS} is not selected explicitly"
+            assert tune_fast_device_class, (
+                f"{msg} when {constants.DEVICECLASS}" " is not selected explicitly"
```
suchita-g, line 1593: f"{msg} when {constants.DEVICECLASS} is not selected explicitly"

suchita-g (commented):
Added Minor comment : Unnecessary String Concatenation

ebenahar (approved):

PR #15924, opened 2026-08-10
`test_multiregion_mirror` - use Jira mark instead of unconditionally skipping the test

sagihirshfeld:
## Summary
Replace the unconditional `@pytest.mark.skip` on `test_multiregion_mirror` with a `@jira("[DFBUGS-9429](https://redhat.atlassian.net/browse/DFBUGS-9429)")` marker so the test automatically resumes running once the bug is resolved, instead of requiring a separate PR to remove the skip.

## Summary by CodeRabbit

* **Tests**
  * Updated multi-region mirror test tracking and removed its unconditional skip, allowing the test to run.

### DIFF
--- a/tests/functional/object/mcg/test_multi_region.py
+++ b/tests/functional/object/mcg/test_multi_region.py
@@ -14,6 +14,7 @@ from ocs_ci.framework.pytest_customization.marks import (
     runs_on_provider,
     mcg,
     tier2,
+    jira,
 )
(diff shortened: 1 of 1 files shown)


sagihirshfeld:
In the above verification, we can see that the test was skipped with the expected message: 
image

udaysk23 (approved):

DanielOsypenko (approved):

PR #15957, opened 2026-08-13
Test MDR multi ns discovered app failover-relocate for HCI

jilju:
## Summary by CodeRabbit

* **Tests**
  * Added disaster-recovery coverage for failing over and relocating discovered applications across multiple namespaces.
  * Validates scenarios with the primary cluster available or unavailable.
  * Covers both Ceph block and Ceph filesystem storage interfaces.
  * Includes verification of fencing, resource creation, application recovery, and successful relocation throughout the workflow.
  * Adds cleanup handling for workflows where resource-deletion verification is deferred.


tests/functional/disaster-recovery/metro-dr-pc/test_failover_relocate_discovered_apps_multi_ns.py
```diff
@@ -0,0 +1,207 @@
...
+            failover_cluster=secondary_cluster_name,
+            namespace=workloads[0].workload_namespace,
+            discovered_apps=True,
+            workload_placement_name=workloads[0].discovered_apps_placement_name,
+            old_primary=self.primary_cluster_name,
+            skip_odf_cli_validation=primary_cluster_down,
+        )
+
+        logger.info("Doing Cleanup Operations")
+        do_discovered_apps_cleanup_multi_ns(
```
sidhant-agrawal: Cluster will remain fenced at this stage, so PVC will be stuck in the terminating state.
Test will fail at this step.

PR #15960, opened 2026-08-13
delete security groups early - unblock hcp destroy

DanielOsypenko:
```
 1. VPC peering cleanup moved before hypershift destroy cluster (line 4288)   
                  
  The _cleanup_vpc_peering() call is now executed before hypershift destroy
  cluster, not after. This removes:
  - VPC peering connections between client and management VPCs
  - Ceph port ingress rules from management VPC SGs referencing client VPC CIDR
  - Routes to client VPC from management route tables

  This eliminates cross-VPC dependencies that caused the hypershift destroy
  command to loop for 20 minutes on DependencyViolation: resource sg-xxx has a
  dependent object errors before timing out.

  2. New _discover_oidc_config() method (line 4357)

  Discovers OIDC bucket name/region during destroy using a 3-tier strategy:
  1. Cluster config
  2. K8s secret on management cluster
  3. Naming convention fallback: {name}-oidc-bucket

  This fixes the root cause of OIDC/S3 leftovers: oidc_bucket_name was None
  during destroy, so _cleanup_iam_resources(), _cleanup_s3_oidc_objects(), and
  _cleanup_stale_oidc_providers() all silently skipped their work.

  3. Enhanced _cleanup_s3_oidc_objects() (line 5264)

  Now distinguishes between:
  - Per-cluster OIDC bucket ({name}-oidc-bucket): deletes the entire bucket +
  contents
  - Shared OIDC bucket: deletes only objects under {infra_id}/ prefix

  Also added upfront head_bucket check to gracefully skip if bucket is already
  gone.
  ```

### DIFF
--- a/ocs_ci/deployment/hub_spoke.py
+++ b/ocs_ci/deployment/hub_spoke.py
@@ -4270,6 +4270,8 @@ class HypershiftAWSHostedOCP(SpokeOCP, HyperShiftBase, Deployment, MCEInstaller,
         """
         all_clean = True
 
+        self._discover_oidc_config()
+
         hypershift_cli_available = self._ensure_sts_credentials()
 
         if not self.role_arn:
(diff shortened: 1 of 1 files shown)


DanielOsypenko:
verified: https://url.corp.redhat.com/9ccc0e2

DanielOsypenko:
verified: https://url.corp.redhat.com/6e45d9a

fbalak (approved):

dahorak (approved):

PR #15978, opened 2026-08-17
normilize-hyperconverged-version

DanielOsypenko:
Fix HCO version mapping for OCP 5.x by normalizing 5.0 to 4.23 equivalent

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved version matching for Hyperconverged deployments across supported OpenShift Container Platform release formats.
  - Added support for incomplete version values and equivalent version notation, helping ensure the correct corresponding release is selected.
  - Improved compatibility when working with newer minor release numbers.

### DIFF
--- a/ocs_ci/deployment/hyperconverged.py
+++ b/ocs_ci/deployment/hyperconverged.py
@@ -191,6 +191,26 @@ class HyperConverged:
         HyperConverged.create_hyperconverged_instance(self)
 
 
+def _normalize_ocp_version(ocp_version: str) -> semantic_version.Version:
+    """
+    Parse an OCP version string and normalize OCP 5.x to the 4.x equivalent.
+
+    OCP 5.0 is equivalent to OCP 4.23, so 5.x maps to 4.(23 + x).
+
+    Args:
+        ocp_version: OCP version as a string (e.g., "4.18", "5.0.0-0.nightly-...")
+    Returns:
+        semantic_version.Version normalized to the 4.x series
+    """
+    if not semantic_version.validate(ocp_version):
+        ocp_version += ".0"
+    ocp_semver = semantic_version.Version(ocp_version)
+    if ocp_semver.major >= 5:
+        normalized_minor = 23 + (ocp_semver.major - 5) * 23 + ocp_semver.minor
+        ocp_semver = semantic_version.Version(f"4.{normalized_minor}.0")
+    return ocp_semver
+
+
 def get_hyperconverged_corresponding_version(ocp_version: str) -> str:
     """
     Given an OCP version, return the corresponding Hyperconverged version.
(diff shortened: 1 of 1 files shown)


fbalak (approved):

dahorak (approved):

DanielOsypenko:
the PR is succesfully creates expected tag that matches the OCP supportability.
Although the job failed to deploy healthy hyperconverged - operator is in CLBO. 
AI Analysis is saying about bug of the latest build of  `1.19.0-unstable`
I propose to merge this PR and if necessary to go with older version (Fallback version) or with CNV for some time until problem will be resolved.  

https://url.corp.redhat.com/3d2f857

RCA from AI:
```
  1. Why did it fail? (the simple version)

  The KubeVirt/CNV operator crashes the instant it starts, so it never finishes installing. Your test waits 10 minutes for the virt-operator pods to appear, they never do, and the job fails with a timeout.

  The chain:
  HCO operator starts → reads its config (the HyperConverged CR)
     → can't understand the "featureGates" field → exits immediately
     → CrashLoopBackOff → never creates virt-operator (0/2)
     → ocs-ci waits 600s → TimeoutExpiredError → job FAILURE

  The exact reason it can't understand its config, proven on the cluster:
  - The CRD (schema) says featureGates is an object of on/off flags, and even fills in a default one:
  {"decentralizedLiveMigration":true, "declarativeHotplugVolumes":true, ...}
  - The operator binary in the same image rejects that object:
  cannot unmarshal object into Go struct field spec.featureGates of type HyperConvergedFeatureGates

  So the operator's own schema (CRD) and its own program (binary) disagree about what featureGates should look like. The image is internally inconsistent.

  ---
  2. Is it a hyperconverged (CNV) bug?
  
  Yes. This is an upstream CNV/HCO bug in this specific image:
  quay.io/kubevirt/hyperconverged-cluster-index:1.19.0-unstable
  The -unstable tag is a floating, in-development community build. In today's snapshot, the CRD and the operator binary are out of sync on the featureGates type (mid-migration, one side updated and the other
  not). It has nothing to do with: 
  - ODF/storage (not installed yet at this phase — expected),
  - TechPreviewNoUpgrade (a separate, cluster-level setting),
  - anything ocs-ci sets (your HyperConverged CR is bare — it doesn't set featureGates at all; the bad default comes from the operator's own CRD).

  ---
  3. Is it fixable?

  Yes, but the real fix is on the CNV side, not yours. Options:

  - Wait / retry — because it's the floating -unstable build, a later snapshot with a matching CRD+binary will just start working with no code change.
  - Pin a known-good version — instead of always using {version}.0-unstable, point the catalog image at a stable/released CNV index (or a specific unstable digest known to work). This makes deployments
  reproducible instead of "whatever unstable is today."
  - You cannot fix it by editing the CR — the broken default is injected by the operator's CRD, not by you.

  ---
  4. Is your PR working properly?
  
  Yes — your PR's logic did exactly what it's supposed to. I traced it:

  - PR 15978 = "normilize-hyperconverged-version", changes ocs_ci/deployment/hyperconverged.py.
  - Its version mapping (HCO major = OCP major − 3, HCO minor = OCP minor − 4) correctly turned OCP 4.23 → HCO 1.19, and built the image hyperconverged-cluster-index:1.19.0-unstable. ✅
  - The deployment steps all ran in order: namespace → operator group → catalog source → subscription → CR → wait for virt-operator. ✅

  Your code picked the right version and deployed it correctly. It then died on a broken upstream image, which your PR has no control over.

  One caveat worth considering: your PR always targets the -unstable floating tag. That means your deployments are at the mercy of whatever community build exists that day — exactly like this failure. If the
  goal of "normalize the version" is stable, reproducible deployments, you may want to pin to a released/GA CNV index (or a fixed digest) rather than -unstable. That's an improvement to consider, not a bug in
  what you wrote.

  ---
  TL;DR

  Not your bug, not a storage bug, not TPNU. It's a broken community 1.19.0-unstable CNV image whose CRD and operator binary disagree about featureGates. Your PR correctly computed and deployed HCO 1.19 for OCP
  4.23; it just landed on a bad build. Re-run later or pin a known-good CNV version.
  ```

PR #16000, opened 2026-08-19
Fixes to stabalize logging with lokistack

posharma1:
## Summary by CodeRabbit

- **Improvements**
  - Updated LokiStack deployments to use the extra-small configuration.
  - Improved logging deployment checks when PVC availability varies or resources already exist.
  - Enhanced handling when required logging packages are unavailable.
  - Improved polling for generated configuration resources until required values are available.
  - Enhanced log validation retries and error messages when no results are returned.
  - Added clearer reporting for unavailable logging data and configuration resources.

### DIFF
--- a/ocs_ci/utility/deployment_openshift_logging.py
+++ b/ocs_ci/utility/deployment_openshift_logging.py
@@ -381,9 +381,11 @@ def create_clusterlogforwarder(yaml_file, skip_resource_exists=False):
     try:
         clf_obj.create(yaml_file=yaml_file)
     except CommandFailed as e:
-        if "AlreadyExists" in str(e) and skip_resource_exists:
-            logger.warning("clusterlogforwader already exists")
-            return True
+        if "AlreadyExists" in str(e):
+            logger.warning(
+                "clusterlogforwarder already exists, skipping creation "
+                "and continuing with pod/PVC verification"
+            )
         else:
             raise
 
--- a/ocs_ci/utility/deployment_openshift_logging.py
+++ b/ocs_ci/utility/deployment_openshift_logging.py
@@ -10,7 +10,11 @@ import pytest
 
 from ocs_ci.ocs import constants, ocp
 from ocs_ci.utility import templating
-from ocs_ci.ocs.exceptions import CommandFailed, ResourceNotFoundError
+from ocs_ci.ocs.exceptions import (
+    CommandFailed,
+    ConfigMapDataNotAvailable,
+    ResourceNotFoundError,
+)
 from ocs_ci.helpers import helpers
 from ocs_ci.utility import deployment_openshift_logging as ocp_logging_obj
 from ocs_ci.utility.utils import (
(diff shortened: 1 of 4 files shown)


ocs_ci/utility/deployment_openshift_logging.py
```diff
@@ -360,7 +366,6 @@ def setup_sa_permissions():
     exec_cmd(cmd)
 
 
-@retry(CommandFailed, tries=5, delay=60, backoff=2)
```
PrasadDesala, line 363: why is the retry removed?

ocs_ci/utility/deployment_openshift_logging.py
```diff
@@ -549,6 +556,19 @@ def install_logging(skip_resource_exists=False):
...
+        package_manifest = PackageManifest(
+            resource_name="cluster-logging",
+            selector="catalog=redhat-operators",
+        )
+        package_manifest.get()  # Verify it exists
+    except ResourceNotFoundError:
+        pytest.skip(
+            "cluster-logging operator not available in catalog. "
+            "Skipping openshift-logging tests."
+        )
```
PrasadDesala, line 570: pytest.skip() called inside a utility module (deployment_openshift_logging.py), please skip it in the test or fixture

ocs_ci/utility/deployment_openshift_logging.py
```diff
@@ -605,23 +625,37 @@ def install_logging(skip_resource_exists=False):
...
+        resource_name=constants.OBJECT_BUCKET_CLAIM,
+    ):
+        data = (cm or {}).get("data", {})
+        if data.get("BUCKET_NAME") and data.get("BUCKET_HOST"):
+            cm_dict = cm
+            break
+    if cm_dict is None:
+        raise ConfigMapDataNotAvailable(
+            f"ConfigMap {constants.OBJECT_BUCKET_CLAIM} exists but "
+            "BUCKET_NAME and BUCKET_HOST were not populated within the timeout"
```
PrasadDesala, line 657: will it be ever called? TimeoutSampler raises TimeoutExpiredError on exhaustion, it never exits the for loop silently

ocs_ci/utility/deployment_openshift_logging.py
```diff
@@ -404,7 +411,7 @@ def create_clusterlogforwarder(yaml_file, skip_resource_exists=False):
     )
     pvc_status = pvc_obj.wait_for_resource(
         condition=constants.STATUS_BOUND,
-        resource_count=nodes_in_cluster,
+        selector="app.kubernetes.io/instance=logging-loki",
```
PrasadDesala, line 414: how many pvc do we expect to run bound state here?

you have dropped esource_count entirely, using a selector without resource_count means wait_for_resource will return True as soon as any PVC matching the selector is Bound. If provisioning is partial for eg 1 of 3 PVCs Bound, the check passes prematurely.

ocs_ci/utility/deployment_openshift_logging.py
PrasadDesala: Logging should be available by default and it is not an optional one in the packages. I think we can even remove it

dgaikwad:
Apart from the Prasad's Comments it looks good to me.

PR #16005, opened 2026-08-19
Fix RHCOS version defaults for OCP 4.23 and 5.0

petr-balogh:
OCP 4.23: RHEL 9 is the default OS (RHEL 10 is still Tech Preview). The config incorrectly defaulted to rhcos_version: "10", which would set TechPreviewNoUpgrade and osImageStream in install-config.yaml. Removed that default so 4.23 uses RHEL 9 by default. Users can still opt-in to RHEL 10 via rhcos10_testing.yaml overlay.

OCP 5.0: RHEL 10 is GA and the default for new installs, so TechPreviewNoUpgrade must NOT be set. Added OCP version gating in all 4 deployment files (ocp.py, vmware.py UPI/IPI, baremetal.py) to skip TechPreviewNoUpgrade and osImageStream when OCP >= 5.0. Also set rhcos_version: "10" as the default in the 5.0 config.

## Summary by CodeRabbit

* **Documentation**
  * Updated RHCOS version guidance for OCP 4.x and OCP 5.0+.
  * Clarified default RHCOS versions and available opt-in configurations.

* **Bug Fixes**
  * Adjusted RHEL 10 installation settings for compatibility with OCP versions below 5.0.
  * OCP 5.0 and later now use default RHEL 10 behavior without legacy configuration overrides.
  * Corrected the OCP 4.23 fallback to use the RHCOS 9 template.

### DIFF
--- a/conf/README.md
+++ b/conf/README.md
@@ -266,7 +266,7 @@ higher priority).
 * `vm_template` - VMWare template to use for RHCOS images (legacy single template, used as fallback)
 * `vm_template_overwrite` - VM template to overwirthe for early testing deployment e.g. rhcos-47.84.202103151537-0-vmware.x86_64
 * `vm_templates` - Dictionary of available RHCOS templates by major version for VMWare deployments (e.g., `{"9": "rhcos-9.6...", "10": "rhcos-10.2..."}`)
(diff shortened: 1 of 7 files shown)


vavuthu (approved):

clacroix12 (approved):

dahorak (approved):

petr-balogh:
I see from validation jobs:
5.0:
```
Red Hat Enterprise Linux CoreOS 10.2.20260817-0 (Coughlan)   6.12.0-211.47.1.el10_2.x86_64 (amd64)   
```

From 4.23:
```
Red Hat Enterprise Linux CoreOS 9.8.20260817-0 (Plow)  5.14.0-687.39.1.el9_8.x86_64 (amd64)
```

PR #16023, opened 2026-08-24
Fix: Use dynamic agentInstallNamespace instead of hardcoded constant

prsurve:
In configure_acm_to_import_mce_clusters(), the pod-label wait loop in the automated way (ACM > 2.14) was using the hardcoded constant ACM_ADDONS_NAMESPACE_DISCOVERY as the namespace. This fails when the actual agentInstallNamespace configured in the hypershift-addon-deploy-config AddonDeploymentConfig differs from the default.

Fix by fetching the live AddonDeploymentConfig resource after the patch and reading agentInstallNamespace from its spec dynamically.


shylesh (approved):
lgtm

PR #16031, opened 2026-08-26
Adds Happy path validation for CNPG TLS feature

udaysk23:
Adds Happy path validation for CNPG TLS feature

## Summary by CodeRabbit

- **New Features**
  - Added support for secure TLS connections between NooBaa and its PostgreSQL database.
  - Added configuration for database certificates and enforced encrypted database communication.

- **Tests**
  - Added end-to-end validation of certificate setup and encrypted connections.
  - Verified TLS enforcement, protection of private key material, and object data integrity before and after upgrades.


tests/functional/object/mcg/test_noobaa_db_tls.py
```diff
@@ -0,0 +1,389 @@
...
+        f"user={db_user} sslmode=require' "
+        f"-tAc '{ssl_query}'"
+    )
+    ssl_used = (
+        nb_db_pod.exec_sh_cmd_on_pod(command=psql_cmd, secrets=[db_password])
+        .strip()
+        .lower()
+    )
+    logger.info(f"pg_stat_ssl.ssl for the current backend returned: {ssl_used}")
+    if ssl_used != "t":
```
ypersky1980, line 222: Missing exception handling breaks test pattern:

If the psql command fails (connection refused, auth failure, pod not ready), exec_sh_cmd_on_pod() will raise an exception rather than returning False. This breaks the boolean return pattern expected by the assertions in test_noobaa_db_cnpg_tls() 

Fix: Wrap in try-except and return False on failure:

try:
    ssl_used = (
        nb_db_pod.exec_sh_cmd_on_pod(command=psql_cmd, secrets=[db_password])
        .strip()
        .lower()
    )
except Exception as e:
    logger.error(f"Failed to query pg_stat_ssl: {e}")
    return False

tests/functional/object/mcg/test_noobaa_db_tls.py
```diff
@@ -0,0 +1,389 @@
...
+        assert (
+            verify_postgres_ssl_required()
+        ), "NooBaa core does not require PostgreSQL TLS (POSTGRES_SSL_REQUIRED)"
+        assert (
+            verify_encrypted_db_connection()
+        ), "NooBaa DB connection is not encrypted (TLS not in use)"
+        assert (
+            verify_no_tls_key_material_in_env()
+        ), "TLS key/cert material is exposed in NooBaa container env vars"
+
```
ypersky1980, line 310: Missing TLS validation for read-only host:

The test only validates TLS on CNPG_READ_WRITE_HOST (line 192) but doesn't test the read-only host (CNPG_READ_ONLY_HOST from constants.py line 1003). This is a coverage gap - read-only replicas should also serve TLS connections.

Recommendation: Add a similar validation for the read-only host, either by extending verify_encrypted_db_connection() to test both hosts, or adding a separate assertion in this test method.

ypersky1980 (commented):
Left 2 comments.

tests/functional/object/mcg/test_noobaa_db_tls.py
udaysk23: Fixed

tests/functional/object/mcg/test_noobaa_db_tls.py
udaysk23: Fixed

ypersky1980 (approved):
LGTM

fbalak (approved):

PR #16036, opened 2026-08-26
moving temp catsrc creation inside gitops deploy function

keemano:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GitOps deployment on OpenShift 5.0 and later by ensuring the required temporary catalog source is created consistently across clusters.
  * Updated the deployment template location to ensure the configuration is applied reliably.
* **Deployment**
  * Added a temporary catalog source using the OpenShift Data Foundation catalog image.
  * Configured periodic gRPC registry polling every 15 minutes to keep catalog information current during deployment.


ocs_ci/deployment/deployment.py
```diff
@@ -487,6 +487,11 @@ def do_gitops_deploy(self):
         Returns:
 
         """
+        # Temporary workaround: create the GitOps CatalogSource on all clusters
+        # before deploying the GitOps operator and OADP.
+        if version.get_semantic_ocp_version_from_config() >= version.VERSION_5_0:
+            run_cmd_multicluster(f"oc apply -f {constants.TEMP_GITOPS_WA_YAML}")
+
```
dahorak: This has to be under the `if config.multicluster:` condition, otherwise it will not solve anything, because it will be executed everytime similarly like the original version.

dahorak (changes requested):

ocs_ci/deployment/deployment.py
keemano: done

dahorak (approved):

keemano:
Verified - https://jenkins-csb-odf-qe-ocs4.dno.corp.redhat.com/job/qe-rdr-setup/5540/console

petr-balogh (approved):

petr-balogh:
Running also verification here:
 https://url.corp.redhat.com/813e3ee
 On normal disconnected deployment

PR #16041, opened 2026-08-27
Adds Happy path validation and upgrade scenario for noobaa core HA feature

udaysk23:
Adds Happy path validation and upgrade scenario for noobaa core HA feature

## Summary by CodeRabbit

* **Tests**
  * Added functional coverage for NooBaa core high availability.
  * Verifies core pods run on separate nodes and leader election remains consistent.
  * Confirms switching HA off and back on produces the expected number of pods.
  * Validates S3 object upload and download integrity during HA testing.
  * Added cleanup safeguards to restore HA and wait for core availability after failures.

### DIFF
--- a/tests/functional/object/mcg/test_noobaa_core_ha.py
+++ b/tests/functional/object/mcg/test_noobaa_core_ha.py
@@ -1,0 +1,243 @@
+import logging
+
+import pytest
+
+from ocs_ci.framework import config
+from ocs_ci.framework.pytest_customization.marks import (
+    mcg,
+    post_upgrade,
+    red_squad,
+    runs_on_provider,
+    skipif_ocs_version,
+    tier2,
+)
+from ocs_ci.framework.testlib import MCGTest
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.ocp import OCP
+from ocs_ci.ocs.resources.pod import (
+    Pod,
+    get_pod_node,
+    get_pods_having_label,
+    wait_for_pods_by_label_count,
+    wait_for_pods_to_be_running,
+)
+from ocs_ci.ocs.bucket_utils import (
+    sync_object_directory,
+    verify_s3_object_integrity,
+    write_random_objects_in_pod,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def get_noobaa_core_pods():
+    """
+    Fetch all NooBaa core pods in the cluster namespace.
+
+    Returns:
+        list: List of NooBaa core Pod objects
+
+    """
+    core_pods = get_pods_having_label(
+        label=constants.NOOBAA_CORE_POD_LABEL,
+        namespace=config.ENV_DATA["cluster_namespace"],
+    )
+    return [Pod(**pod) for pod in core_pods]
+
+
+@mcg
+@red_squad
+@runs_on_provider
+class TestNoobaaCoreHA(MCGTest):
+    """
+    Test NooBaa core high availability (HA).
+    """
+
+    @pytest.fixture(autouse=True)
+    def teardown(self, request):
+        """
+        Ensure NooBaa core HA is re-enabled after the test, even if it fails
+        mid-way, so the cluster is not left with a single core pod.
+        """
+
+        def finalizer():
+            core_pods = get_noobaa_core_pods()
+            if len(core_pods) == 2:
+                return
+            logger.warning(
+                f"Found {len(core_pods)} NooBaa core pod(s) during teardown, "
+                "re-enabling core HA"
+            )
+            self._set_core_ha(disabled=False)
+            wait_for_pods_by_label_count(
+                constants.NOOBAA_CORE_POD_LABEL, expected_count=2
+            )
+
+        request.addfinalizer(finalizer)
+
+    def _set_core_ha(self, disabled):
+        """
+        Patch the NooBaa CR to enable/disable core HA.
+
+        Args:
+            disabled (bool): True disables core HA (single core pod),
+                False enables core HA (two core pods)
+
+        """
+        noobaa_obj = OCP(
+            kind=constants.NOOBAA_RESOURCE_NAME,
+            namespace=config.ENV_DATA["cluster_namespace"],
+        )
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_noobaa_core_ha.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # 7. Validate there are 2 core pods
+        logger.info("Verifying that two NooBaa core pods are running again")
+        wait_for_pods_by_label_count(constants.NOOBAA_CORE_POD_LABEL, expected_count=2)
+        core_pods = get_noobaa_core_pods()
+        wait_for_pods_to_be_running(
+            namespace=config.ENV_DATA["cluster_namespace"],
+            pod_names=[pod.name for pod in core_pods],
+            timeout=300,
+        )
```
ypersky1980: The two core HA guarantees this test exists to protect — 2 pods on separate nodes and a working leader-election lease — are only validated in the initial state (steps 2–3). After the disable → re-enable cycle (step 7), which is the actual transition this test exercises, only the pod count is re-checked. A regression where the re-enabled pods co-locate on the same node, or where leader election fails to recover, would pass silently. Please re-run the node-distinctness check and self._assert_lease_holder([...]) after step 7.

tests/functional/object/mcg/test_noobaa_core_ha.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # 8. Create S3 bucket, PUT object, GET object, verify checksum
```
ypersky1980: Object I/O is only exercised after the HA toggle, so the test never verifies that pre-existing data survives the core-pod scale-down/up. Given the PR claims an "upgrade scenario", consider writing + uploading a baseline set of objects before disabling HA (step 4) and verifying their checksums after re-enabling (step 7), so data durability across the transition is actually covered — not just that the S3 path works at the end.

tests/functional/object/mcg/test_noobaa_core_ha.py
```diff
@@ -0,0 +1,222 @@
...
+            core_pod_names (list): Names of the running NooBaa core pods
+
+        """
+        namespace = config.ENV_DATA["cluster_namespace"]
+        lease_ocp = OCP(kind="Lease", namespace=namespace)
+        leases = lease_ocp.get().get("items", [])
+        assert leases, f"No Lease objects found in namespace {namespace}"
+
+        holder = None
+        for lease in leases:
```
ypersky1980: _assert_lease_holder iterates over every Lease in the namespace and matches on pod-name only. It should scope to the specific NooBaa core leader-election lease (by name/label), otherwise a missing or renamed NooBaa lease won't be caught, and the match relies on an unrelated lease happening to not collide. Also, holder_identity.startswith(f"{name}_") assumes a specific holderIdentity format — if NooBaa's format differs this becomes a false failure (flake). Worth anchoring to the known lease name.

tests/functional/object/mcg/test_noobaa_core_ha.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # 3. Confirm lease exists and holderIdentity matches one pod
+        self._assert_lease_holder([pod.name for pod in core_pods])
+
+        # 4. Disable HA flag from NooBaa CR
+        self._set_core_ha(disabled=True)
+
+        # 5. Validate there is only 1 core pod
+        logger.info("Verifying that a single NooBaa core pod remains")
+        wait_for_pods_by_label_count(constants.NOOBAA_CORE_POD_LABEL, expected_count=1)
```
ypersky1980: After disabling HA, the test waits for the count to reach 1 but never confirms that the remaining single pod is actually Running/Ready (steps 1 and 7 both call wait_for_pods_to_be_running, step 5 does not). A crash-looping single core pod would still pass "HA disabled". Add a readiness wait here for consistency.

tests/functional/object/mcg/test_noobaa_core_ha.py
```diff
@@ -0,0 +1,222 @@
...
+        logger.info(f"NooBaa core pods are scheduled on nodes: {node_names}")
+        assert len(set(node_names)) == 2, (
+            "NooBaa core pods are not running on different nodes: " f"{node_names}"
+        )
+
+        # 3. Confirm lease exists and holderIdentity matches one pod
+        self._assert_lease_holder([pod.name for pod in core_pods])
+
+        # 4. Disable HA flag from NooBaa CR
+        self._set_core_ha(disabled=True)
```
ypersky1980: The test mutates a cluster-wide singleton (disableCoreHA on the NooBaa CR) and relies solely on the autouse finalizer to restore it. If the process is interrupted between disable and re-enable — or the finalizer's wait times out — the cluster is left degraded (single core pod) for every subsequent test in the run. Consider wrapping the disabled window in a try/finally inside the test that restores HA, in addition to the finalizer. (Distinct from CodeRabbit's point, which only asks you to assert the finalizer's return value.)

ypersky1980 (commented):
Left 5 comments

tests/functional/object/mcg/test_noobaa_core_ha.py
udaysk23: Fixed

tests/functional/object/mcg/test_noobaa_core_ha.py
udaysk23: fixed

tests/functional/object/mcg/test_noobaa_core_ha.py
udaysk23: fixed

tests/functional/object/mcg/test_noobaa_core_ha.py
udaysk23: fixed

tests/functional/object/mcg/test_noobaa_core_ha.py
udaysk23: fixed

ypersky1980 (approved):
LGTM

fbalak (approved):

PR #16089, opened 2026-09-02
Add negative scenario test case for user namespace

anajain16:
Add test case for [RHSTOR-8028](https://redhat.atlassian.net/browse/RHSTOR-8028) to check uid remapping not happening in case user namespace is not enabled 

## Summary by CodeRabbit

- **Tests**
  - Expanded CephFS functional coverage for restricted pod user namespaces.
  - Added validation that files use the namespace-assigned UID and filesystem group without host UID remapping when host user namespaces are not disabled.
  - Updated test categorization to distinguish tier-one and tier-two coverage.
  - Added coverage for supported environment configurations, including managed services and specialized storage deployments.

### DIFF
--- a/tests/functional/pv/pv_services/test_user_namespace_cephfs.py
+++ b/tests/functional/pv/pv_services/test_user_namespace_cephfs.py
@@ -7,11 +7,12 @@ from ocs_ci.framework.pytest_customization.marks import (
     green_squad,
     skipif_ocs_version,
 )
-from ocs_ci.framework.testlib import ManageTest, tier1, polarion_id
+from ocs_ci.framework.testlib import ManageTest, tier1, tier2, polarion_id
 from ocs_ci.ocs import constants, node
 from ocs_ci.helpers.helpers import (
     create_userns_project,
     create_userns_pod,
+    create_pod,
     verify_file_ownership,
     wait_for_resource_state,
 )
@@ -23,7 +24,6 @@ SUPPLEMENTAL_GROUPS_BASE = 10000
 USERNS_UID_RANGE = "10000/1000"
 
 
-@tier1
 @green_squad
 @skipif_ocs_version("<4.23")
 class TestUserNamespaceCephFS(ManageTest):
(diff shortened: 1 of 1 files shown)


tests/functional/pv/pv_services/test_user_namespace_cephfs.py
```diff
@@ -256,3 +258,107 @@ def test_user_namespace_rwo_ownership_and_persistence(
...
+        pod_node = pod_obj.get()["spec"]["nodeName"]
+        logger.info(f"Pod {pod_obj.name} running on {pod_node}")
+
+        logger.test_step("Step 4: Write/read file, read container UID and fsGroup")
+        out = pod_obj.exec_sh_cmd_on_pod(
+            "id -u && echo control > /mnt/test/file1 && sync"
+        )
+        logger.info(f"Container id / write output: {out}")
+        container_uid = int(out.split()[0])
+        fs_group = pod_obj.get()["spec"]["securityContext"]["fsGroup"]
```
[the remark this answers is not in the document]
anajain16: This test's purpose is to prove that without hostUsers: false there's no UID remapping (host_uid == container_uid), which is asserted independently and isn't tautological. Validating that the SCC assigns UID/fsGroup from the namespace ranges is testing OpenShift admission behavior, not the CephFS user-namespace feature under test — it's a guaranteed restricted-v2 default we can rely on, so adding it here would be scope creep.

anajain16:
verified - https://url.corp.redhat.com/fd69d9d

fbalak (dismissed):

tests/functional/pv/pv_services/test_user_namespace_cephfs.py
sidhant-agrawal, line 1: Is this test valid on external mode clusters?

anajain16: not valid if external cluster doesn't provide CephFS , added checks

sidhant-agrawal (approved):

fbalak (approved):

PR #16105, opened 2026-09-04
[release-4.20] Use config.RUN.get for resource_checker to avoid KeyError

clacroix12:
The resource_checker RUN config key is not populated in all deployment entry points (e.g. deploy-fdf), causing a KeyError when creating resources via a yaml file. Use .get() so a missing key is treated as falsy instead of raising.


vavuthu (approved):

vavuthu (approved):

shylesh (approved):
lgtm

PR #16107, opened 2026-09-04
[release-4.18] Handle missing resource_checker RUN config in FDF deployments

clacroix12:
The resource_checker RUN config key is not populated by the deploy-fdf entry point, causing a KeyError when creating resources via a yaml file. Set resource_checker to False during FDF init (mirroring the approach on release-4.20) and use .get() in ocp.py so a missing key is treated as falsy instead of raising.


vavuthu (approved):

petr-balogh (approved):

clacroix12:
Verified: https://url.corp.redhat.com/e2364a8

PR #16116, opened 2026-09-07
Test BackingStore and NamespaceStore endpoint updates (RHSTOR-8718)

ypersky1980:
This PR adds automated coverage for [RHSTOR-8718](https://redhat.atlassian.net/browse/RHSTOR-8718), which adds the CLI command:

noobaa connection update --old-endpoint <OLD> --new-endpoint <NEW>
an endpoint-scoped, all-or-nothing bulk update that matches every store on OLD, pre-validates against NEW, then patches specs and updates the NooBaa-core connections (auto-rollback on failure). The scenarios were run manually against a live cluster (ODF/MCG 5.0.0-36.stable) before being automated.

What is added

New module tests/functional/object/mcg/test_endpoint_update.py — 14 tests, 24 parametrised cases in three classes:

- TestBackingStoreEndpointUpdate (11 cases) — core behaviour: single switch and revert, bulk switch with connection de-duplication, matching semantics (no-match on a wrong endpoint, trailing slash, DNS-vs-IP spelling, rerun-after-success, idempotent no-op), selective matching with a data-path cross-talk check once two connections share one endpoint, the noobaa.io/pause-reconcile annotation lifecycle, switch under active I/O, and the default backingstore.
- TestConnectionUpdateNegative (11 cases) — failure paths: pre-validation aborts the whole batch (unreachable endpoint, missing target bucket, a bucket that exists but is not a NooBaa location, bad credentials, one bad store in a batch), CLI argument handling (missing flag, whitespace trimming, malformed URL, over-long URL), the direct-CR-edit path (regression guard for [DFBUGS-10346](https://redhat.atlassian.net/browse/DFBUGS-10346)), and the update's behaviour when the matched stores are being written to concurrently.
- TestNamespaceStoreEndpointUpdate (2 cases) — NamespaceStore switch and revert, and the mixed BackingStore + NamespaceStore batch. Both assert the intended behaviour (a successful switch) and are skipped while [DFBUGS-10744](https://redhat.atlassian.net/browse/DFBUGS-10744) is unfixed in the build under test.

Bugs found by this automation

Two defects were filed off these tests and are referenced in the code where they shape an assertion:

- [DFBUGS-10937](https://redhat.atlassian.net/browse/DFBUGS-10937) — noobaa connection update aborts fatally on a retryable Kubernetes resource-version conflict; util.KubeUpdate issues a single Update with no re-read and no RetryOnConflict. test_switch_survives_concurrent_store_writes carries @jira("[DFBUGS-10937](https://redhat.atlassian.net/browse/DFBUGS-10937)") and is skipped while it is open. Tests that are not about this race wait for the stores to quiesce first, so they fail only for their own reasons.
- [DFBUGS-10938](https://redhat.atlassian.net/browse/DFBUGS-10938) — check_external_connection misclassifies network and bucket errors as UNKNOWN_FAILURE, because aws_error_mapping still keys on AWS SDK v2 error names in front of a v3 client. The pre-validation tests therefore assert on message substrings rather than status codes; a TODO([DFBUGS-10938](https://redhat.atlassian.net/browse/DFBUGS-10938)) marks the spot to tighten them once core classifies properly.

Also observed, on the NamespaceStore path: [DFBUGS-10743](https://redhat.atlassian.net/browse/DFBUGS-10743) — a failed endpoint patch leaves the store annotated pause-reconcile=true, because the rollback reuses an in-memory object that already carries the new endpoint and is denied by the same webhook. The store_factory teardown clears the annotation before deleting, so a paused store cannot strand the rest of the run.

No environment prerequisites

The earlier revision of this PR required two S3 endpoints to be supplied via config.ENV_DATA["mcg_endpoint_pair"]. That key exists nowhere in ocs-ci, so every test skipped on a standard job. The endpoint_conf fixture now resolves an endpoint pair by itself:

1. config.ENV_DATA["mcg_endpoint_pair"] if set — an explicit override, for labs with two genuinely separate S3 services.
2. Otherwise the pair is derived from the cluster — RGW first, then MCG's own S3 (self-ref). A second Service is created in front of the very same pods, so the cluster hands out a second ClusterIP for one unchanged backend, and the two http://<ip>:80 strings become a real endpoint pair. The target buckets come from cloud_uls_factory, seeded with a noobaa_blocks/ marker object (what CLI pre-validation looks for when checking the bucket is "a valid location used by noobaa"); credentials come from the cloud manager. Everything is torn down by the fixtures that created it.
3. Only if neither is available do the tests skip.

The override accepts three optional keys beyond old / new / target_bucket / secret:

ENV_DATA:
  mcg_endpoint_pair:
    target_buckets:   # one per store for the tests that stand several
                      # stores on one endpoint; defaults to [target_bucket]
    empty_bucket:     # exists on both endpoints but holds no noobaa_blocks/
                      # prefix; without it the not_noobaa_location case skips
    old_dns:          # a second, equivalent spelling of `old` - its DNS name
                      # where `old` is an IP; without it the dns_form case skips
Neither can be conjured from the other keys, so the cases that need them skip rather than guess. A derived pair always supplies all three.

A same-backend pair is deliberate: it keeps data-path continuity assertable and avoids the two-backend "stranding" artifact described in the test plan. Plain http against a ClusterIP is also deliberate — pre-validation resolves the target bucket virtual-hosted style, so a DNS endpoint fails (there is no wildcard DNS for in-cluster services), and an https ClusterIP fails the certificate check, since service-serving certs carry DNS SANs only.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for updating connection endpoints through the CLI.
  * Validates endpoint switching, reverting, matching, duplicate handling, idempotence, active I/O, default stores, concurrent updates, and rollback behavior.
  * Added validation for invalid input, credential failures, webhook denials, reconciliation pauses, and differences between configuration and runtime state.
  * NamespaceStore endpoint scenarios remain skipped due to a known issue.

### DIFF
--- a/tests/functional/object/mcg/test_endpoint_update.py
+++ b/tests/functional/object/mcg/test_endpoint_update.py
@@ -1,0 +1,954 @@
+"""
+Automated coverage for RHSTOR-8718 - "Support Changes in Target Endpoint for
+BackingStores (and NamespaceStores)".
+
+The feature adds the CLI command::
+
+    noobaa connection update --old-endpoint <OLD> --new-endpoint <NEW>
+
+which performs an endpoint-scoped, all-or-nothing bulk update: it matches every
+BackingStore (and NamespaceStore) whose spec endpoint == OLD, pre-validates each
+one against NEW, and only then patches the specs and updates the underlying
+NooBaa-core connections (rolling back automatically on any failure).
+
+The coverage is organised into three test classes:
+
+  * ``TestBackingStoreEndpointUpdate`` - positive / core behaviour: single-store
+    switch and revert, bulk switch with connection de-duplication, endpoint
+    matching semantics (no-match, idempotent no-op, selective matching),
+    switching under active I/O, and the default backingstore.
+  * ``TestConnectionUpdateNegative`` - failure paths: pre-validation aborts the
+    whole batch (unreachable endpoint, non-NooBaa bucket, bad credentials, one
+    bad store in a batch), CLI argument handling, and the direct-CR-edit path.
+  * ``TestNamespaceStoreEndpointUpdate`` - NamespaceStore endpoint update and
+    the mixed BackingStore + NamespaceStore batch (regression guards for
+    DFBUGS-10744 / DFBUGS-10743).
+
+Deferred scenarios that need infrastructure not available in a plain MCG job
+(core-connection fault injection, ODF upgrade, provider mode) are present as
+skipped placeholders so the coverage matrix stays complete.
+
+ENVIRONMENT PREREQUISITES
+-------------------------
+Most tests need TWO reachable S3 endpoints and a secret with credentials valid
+on both.  Prefer a *same-backend* endpoint pair (two routes / LB VIPs fronting
+the SAME bucket, e.g. two RGW endpoints) so data-path continuity can be asserted
+without the two-backend "stranding" simulation artifact described in the test
+plan.  Provide them via ``config.ENV_DATA``::
+
+    ENV_DATA:
+      mcg_endpoint_pair:
+        old: https://<old-endpoint>
+        new: https://<new-endpoint>
+        target_bucket: <bucket-that-is-a-valid-noobaa-location-on-both>
+        secret: <k8s-secret-with-AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY>
+        signature_version: v4        # v2 for plain-http endpoints
+
+Tests requiring this configuration skip cleanly when it is absent.
+"""
+
+import logging
+import re
+
+import pytest
+
+from ocs_ci.framework import config
+from ocs_ci.framework.pytest_customization.marks import (
+    mcg,
+    red_squad,
+    tier1,
+    tier2,
+    tier3,
+    jira,
+)
+from ocs_ci.helpers.helpers import create_resource, create_unique_resource_name
+from ocs_ci.ocs import constants
+from ocs_ci.ocs.ocp import OCP
+
+logger = logging.getLogger(__name__)
+
+NOOBAA_API_VERSION = "noobaa.io/v1alpha1"
+
+
+# ---------------------------------------------------------------------------
+# CLI helper + output parser (reusable by every test)
+# ---------------------------------------------------------------------------
+def run_connection_update(
+    mcg_obj, old_endpoint, new_endpoint, namespace=None, extra_args=""
+):
+    """
+    Run ``noobaa connection update`` via the MCG CLI (odf-cli / mcg-cli, resolved
+    automatically by :meth:`MCG.exec_mcg_cmd`) and return a parsed result.
+
+    Args:
+        mcg_obj (MCG): The MCG fixture object.
+        old_endpoint (str): Value passed to ``--old-endpoint``.
+        new_endpoint (str): Value passed to ``--new-endpoint``.
+        namespace (str): Namespace override (defaults to the MCG namespace).
+        extra_args (str): Any additional raw CLI arguments.
+
+    Returns:
+        dict: Parsed output, see :func:`parse_connection_update_output`.
+    """
+    cmd = (
+        f"connection update --old-endpoint {old_endpoint} "
+        f"--new-endpoint {new_endpoint} {extra_args}"
+    ).strip()
+    logger.info(f"Running MCG CLI: {cmd}")
+    result = mcg_obj.exec_mcg_cmd(cmd, namespace=namespace, ignore_error=True)
+    parsed = parse_connection_update_output(result)
+    summary = {k: v for k, v in parsed.items() if k != "raw"}
+    logger.info(f"connection update parsed result: {summary}")
+    return parsed
+
+
+def _search_int(pattern, text):
+    match = re.search(pattern, text)
+    return int(match.group(1)) if match else None
+
+
+def parse_connection_update_output(result):
+    """
+    Parse the ``connection update`` output into a structured dict.
+
+    The CLI logs to both stdout and stderr (logrus), so both are combined.
+
+    Returns:
+        dict: {
+            "raw": str,                    # combined stdout+stderr
+            "returncode": int | None,
+            "matched": int | None,         # "Found N store(s) matching endpoint"
+            "stores_updated": int | None,  # "Stores updated: N"
+            "connections_updated": int | None,
+            "unique_connections": int | None,  # "Found N unique connection(s)"
+            "prevalidation_passed": bool,
+            "prevalidation_failed": bool,
+            "aborted": bool,               # "No changes have been made"
+            "no_match": bool,              # "no matching stores"
+            "webhook_denied": bool,
+            "status_codes": list[str],     # e.g. ["INVALID_ENDPOINT"]
+        }
+    """
+    out = "\n".join(
+        [
+            (getattr(result, "stdout", "") or ""),
+            (getattr(result, "stderr", "") or ""),
+        ]
+    )
+    return {
+        "raw": out,
+        "returncode": getattr(result, "returncode", None),
+        "matched": _search_int(r"Found (\d+) store", out),
+        "stores_updated": _search_int(r"Stores updated:\s*(\d+)", out),
+        "connections_updated": _search_int(r"Connections updated:\s*(\d+)", out),
+        "unique_connections": _search_int(r"Found (\d+) unique connection", out),
+        "prevalidation_passed": "passed pre-validation" in out,
+        "prevalidation_failed": "Pre-validation failed" in out,
+        "aborted": "No changes have been made" in out,
+        "no_match": "no matching stores" in out.lower(),
+        "webhook_denied": (
+            "admissionwebhook.noobaa.io" in out and "denied the request" in out
+        ),
+        "status_codes": re.findall(r"status=([A-Z_]+)", out),
+    }
+
+
+# ---------------------------------------------------------------------------
+# Resource helpers
+# ---------------------------------------------------------------------------
+def _store_dict(
+    kind, name, namespace, endpoint, target_bucket, secret_name, signature_version
+):
+    """Build an s3-compatible BackingStore/NamespaceStore CR dict."""
+    return {
+        "apiVersion": NOOBAA_API_VERSION,
+        "kind": kind,
+        "metadata": {"name": name, "namespace": namespace},
+        "spec": {
+            "type": "s3-compatible",
+            "s3Compatible": {
+                "endpoint": endpoint,
+                "targetBucket": target_bucket,
+                "signatureVersion": signature_version,
+                "secret": {"name": secret_name, "namespace": namespace},
+            },
+        },
+    }
+
+
+def get_store_endpoint(kind, name, namespace):
+    """Return the current spec endpoint of a BackingStore/NamespaceStore."""
+    obj = OCP(kind=kind, namespace=namespace, resource_name=name).get()
+    return obj["spec"]["s3Compatible"]["endpoint"]
+
+
+def get_store_pause_annotation(kind, name, namespace):
+    """Return the value of the noobaa.io/pause-reconcile annotation (or None)."""
+    annotations = (
+        OCP(kind=kind, namespace=namespace, resource_name=name)
+        .get()
+        .get("metadata", {})
+        .get("annotations", {})
+    )
+    return annotations.get("noobaa.io/pause-reconcile")
+
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+@pytest.fixture
+def endpoint_conf():
+    """
+    Return the configured endpoint pair / bucket / secret, or skip.
+
+    See the module docstring for the expected ``ENV_DATA['mcg_endpoint_pair']``
+    structure.
+    """
+    conf = config.ENV_DATA.get("mcg_endpoint_pair")
+    required = ("old", "new", "target_bucket", "secret")
+    if not conf or any(not conf.get(k) for k in required):
+        pytest.skip(
+            "Endpoint-update tests require ENV_DATA['mcg_endpoint_pair'] "
+            "with keys: old, new, target_bucket, secret."
+        )
+    conf.setdefault("signature_version", "v4")
+    return conf
+
+
+@pytest.fixture
+def store_factory(request):
+    """
+    Factory that creates s3-compatible BackingStore / NamespaceStore CRs on a
+    given endpoint, waits for Ready, and cleans them up afterwards.
+
+    Usage::
+
+        bs = store_factory(constants.BACKINGSTORE, endpoint, bucket, secret)
+    """
+    namespace = config.ENV_DATA["cluster_namespace"]
+    created = []
+
+    def _factory(
+        kind,
+        endpoint,
+        target_bucket,
+        secret_name,
+        signature_version="v4",
+        name=None,
+        wait=True,
+    ):
+        name = name or create_unique_resource_name("eps", kind.lower())
+        body = _store_dict(
+            kind,
+            name,
+            namespace,
+            endpoint,
+            target_bucket,
+            secret_name,
+            signature_version,
+        )
+        ocs_obj = create_resource(**body)
+        created.append((kind, name))
+        if wait:
+            OCP(kind=kind, namespace=namespace, resource_name=name).wait_for_resource(
+                condition=constants.STATUS_READY,
+                column="PHASE",
+                timeout=300,
+                sleep=10,
+            )
+        return ocs_obj
+
+    yield _factory
+
+    for kind, name in reversed(created):
+        try:
+            store_ocp = OCP(kind=kind, namespace=namespace)
+            # A paused NamespaceStore (DFBUGS-10743) must have the annotation
+            # removed before it can be reconciled/deleted cleanly.
+            if get_store_pause_annotation(kind, name, namespace):
+                store_ocp.annotate(
+                    annotation="noobaa.io/pause-reconcile-", resource_name=name
+                )
+            store_ocp.delete(resource_name=name, wait=True)
+        except Exception as ex:  # noqa - best-effort teardown
+            logger.warning(f"Teardown of {kind}/{name} failed: {ex}")
+
+
+@pytest.fixture
+def bad_creds_secret(request):
+    """Create a secret carrying invalid S3 credentials (for the wrong-credentials case)."""
+    namespace = config.ENV_DATA["cluster_namespace"]
+    name = create_unique_resource_name("eps-badcreds", "secret")
+    body = {
+        "apiVersion": "v1",
+        "kind": "Secret",
+        "metadata": {"name": name, "namespace": namespace},
+        "type": "Opaque",
+        "stringData": {
+            "AWS_ACCESS_KEY_ID": "not-a-real-access-key",
+            "AWS_SECRET_ACCESS_KEY": "not-a-real-secret-key",
+        },
+    }
+    create_resource(**body)
+
+    def _finalizer():
+        try:
+            OCP(kind="Secret", namespace=namespace).delete(
+                resource_name=name, wait=False
+            )
+        except Exception as ex:  # noqa
+            logger.warning(f"Teardown of secret {name} failed: {ex}")
+
+    request.addfinalizer(_finalizer)
+    return name
+
+
+# ===========================================================================
+# Module A - positive / core behaviour
+# ===========================================================================
+@mcg
+@red_squad
+@jira("RHSTOR-8718")
+class TestBackingStoreEndpointUpdate:
+    """Happy-path and matching-semantics coverage for the endpoint update CLI."""
+
+    @tier1
+    # TODO: assign polarion id
+    def test_switch_and_revert_single_bs(self, mcg_obj, endpoint_conf, store_factory):
+        """
+        Switch a single s3-compatible BackingStore's endpoint from OLD to NEW,
+        then revert it back.
+
+        Flow:
+            1. Create a BackingStore on the OLD endpoint and wait for Ready.
+            2. Run the connection update OLD -> NEW and assert exactly one store
+               was updated, its spec endpoint now points at NEW, and the
+               transient pause-reconcile annotation was cleaned up.
+            3. Run the reverse update NEW -> OLD and assert the store is switched
+               back to its original endpoint.
+        """
+        ns = config.ENV_DATA["cluster_namespace"]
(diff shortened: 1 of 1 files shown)


tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,954 @@
...
+            endpoint_conf["target_bucket"],
+            endpoint_conf["secret"],
+            endpoint_conf["signature_version"],
+        )
+
+        if variant == "wrong_endpoint":
+            bad_old = "https://no-such-endpoint.example.invalid:9000"
+            result = run_connection_update(mcg_obj, bad_old, new)
+        elif variant == "trailing_slash":
+            result = run_connection_update(mcg_obj, old.rstrip("/") + "/", new)
```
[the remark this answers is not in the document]
ypersky1980: Done.

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,954 @@
...
+
+        NOTE: assert continuity only on post-switch I/O because of the
+        endpoint-config propagation window; a same-backend endpoint pair avoids
+        the two-backend stranding artifact.
+        """
+        pytest.skip(
+            "Needs an OBC bound to the store under test plus a background I/O "
+            "loop; wire this to the data-path harness (bucketclass -> OBC -> "
+            "write/read) once the same-backend endpoint pair is available."
+        )
```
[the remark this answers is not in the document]
ypersky1980: Done.

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,954 @@
...
+        if wait:
+            OCP(kind=kind, namespace=namespace, resource_name=name).wait_for_resource(
+                condition=constants.STATUS_READY,
+                column="PHASE",
+                timeout=300,
+                sleep=10,
+            )
+        return ocs_obj
+
+    yield _factory
```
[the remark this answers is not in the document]
ypersky1980: Done.

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,954 @@
...
+    """
+    conf = config.ENV_DATA.get("mcg_endpoint_pair")
+    required = ("old", "new", "target_bucket", "secret")
+    if not conf or any(not conf.get(k) for k in required):
+        pytest.skip(
+            "Endpoint-update tests require ENV_DATA['mcg_endpoint_pair'] "
+            "with keys: old, new, target_bucket, secret."
+        )
+    conf.setdefault("signature_version", "v4")
+    return conf
```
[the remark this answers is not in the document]
ypersky1980: Done.

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2138 @@
...
+        pre_new = set(
+            write_random_test_objects_to_bucket(
+                awscli_pod_session,
+                obc_new.name,
+                test_directory_setup.origin_dir,
+                amount=2,
+                pattern="from-new-store-",
+                mcg_obj=mcg_obj,
+            )
+        )
```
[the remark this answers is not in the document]
ypersky1980: Done.

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2409 @@
...
+    where no such secret exists, so non-RGW clusters keep the coverage.
+  * ``TestConnectionUpdateNegative`` - failure paths: pre-validation aborts the
+    whole batch (unreachable endpoint, missing target bucket, a bucket that
+    exists but is not a NooBaa location, bad credentials, one bad store in a
+    batch), CLI argument handling, the direct-CR-edit path, and the update's
+    behaviour when the matched stores are being written to concurrently.
+  * ``TestNamespaceStoreEndpointUpdate`` - NamespaceStore switch and revert, and
+    the mixed BackingStore + NamespaceStore batch. Both skipped on builds where
+    DFBUGS-10744 (webhook denies the change) is not yet fixed; as of 2026-09-14
+    that fix is merged but not yet in a build, so the marker still applies.
```
udaysk23, line 33: Move these docstrings to respective classes

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2409 @@
...
+     created in front of the very same pods, so the cluster hands out a second
+     ClusterIP for one unchanged backend, and the two ``http://<ip>:80`` strings
+     become a genuine endpoint pair.  The target buckets come from
+     ``cloud_uls_factory`` and are seeded with a ``noobaa_blocks/`` marker object,
+     which is what CLI pre-validation looks for when it checks that the bucket is
+     "a valid location used by noobaa"; one further bucket is provisioned and
+     deliberately left unseeded, to serve the negative case of a bucket that
+     exists but is not a NooBaa location.  The credentials secret comes from the
+     cloud manager.  Everything is torn down with the fixtures that made it.
+  3. Only if neither is available do the tests skip.
```
udaysk23, line 58: Same for this

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2409 @@
...
+                                     # skips - it cannot be derived from the
+                                     # other keys.
+        old_dns: http://<host>:<port>
+                                     # optional; a second, equivalent spelling
+                                     # of `old` - its DNS name where `old` is an
+                                     # IP, or the reverse. Used only as an
+                                     # --old-endpoint that must match nothing.
+                                     # Without it the dns_form no-match case
+                                     # skips.
+"""
```
udaysk23, line 90: Do we need this info?

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2409 @@
...
+    ).strip()
+    logger.info(f"Running MCG CLI: {cmd}")
+    result = mcg_obj.exec_mcg_cmd(cmd, namespace=namespace, ignore_error=True)
+    parsed = parse_connection_update_output(result)
+    summary = {k: v for k, v in parsed.items() if k != "raw"}
+    logger.info(f"connection update parsed result: {summary}")
+    return parsed
+
+
+def _search_int(pattern, text):
```
udaysk23, line 157: Add doc string

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2409 @@
...
+        "spec": {
+            "type": "s3-compatible",
+            "s3Compatible": {
+                "endpoint": endpoint,
+                "targetBucket": target_bucket,
+                "signatureVersion": signature_version,
+                "secret": {"name": secret_name, "namespace": namespace},
+            },
+        },
+    }
```
udaysk23, line 240: Consider building the CR from the existing MCG templates instead of assembling the whole dict by hand. oc_create_rgw_backingstore in ocs_ci/ocs/bucket_utils.py already follows this pattern — templating.load_yaml(constants.MCG_BACKINGSTORE_YAML), then override metadata/spec.

ocs_ci/templates/mcg/BackingStore.yaml and NamespaceStore.yaml already carry apiVersion, the exact kind, the app: noobaa label, and the noobaa.io/finalizer. Loading them lets you drop _CR_KIND and the NOOBAA_API_VERSION literal, use constants.BACKINGSTORE_TYPE_S3_COMP for the type, and — since the current _store_dict omits them — pick up the standard label and finalizer that the rest of the MCG stores carry.

```suggestion
from ocs_ci.utility import templating

_STORE_TEMPLATE = {
    constants.BACKINGSTORE: constants.MCG_BACKINGSTORE_YAML,
    constants.NAMESPACESTORE: constants.MCG_NAMESPACESTORE_YAML,
}

def _store_dict(kind, name, namespace, endpoint, target_bucket, secret_name, signature_version):
    """Build an s3-compatible BackingStore/NamespaceStore CR from the MCG template."""
    body = templating.load_yaml(_STORE_TEMPLATE[kind])
    body["metadata"]["name"] = name
    body["metadata"]["namespace"] = namespace
    body["spec"] = {
        "type": constants.BACKINGSTORE_TYPE_S3_COMP,
        "s3Compatible": {
            "endpoint": endpoint,
            "targetBucket": target_bucket,
            "signatureVersion": signature_version,
            "secret": {"name": secret_name, "namespace": namespace},
        },
    }
    return body

```

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2409 @@
...
+        stable = stable + 1 if current == previous else 0
+        if stable >= stable_samples:
+            logger.info(f"Stores settled at resourceVersions {current}")
+            return True
+        previous = current
+        time.sleep(interval)
+    logger.warning(
+        f"Stores were still being written to after {timeout}s; running anyway"
+    )
+    return False
```
udaysk23, line 392: Consider using TimeoutSampler here instead of the manual deadline / while time.time() < deadline / time.sleep(interval) loop — it's already imported, and it's the OCS-CI convention for polling. It can own the timing while the stable-sample counter stays as-is

```suggestion
previous = None
stable = 0
try:
    for current in TimeoutSampler(timeout, interval, _versions):
        stable = stable + 1 if current == previous else 0
        if stable >= stable_samples:
            logger.info(f"Stores settled at resourceVersions {current}")
            return True
        previous = current
except TimeoutExpiredError:
    logger.warning(f"Stores were still being written to after {timeout}s; running anyway")
    return False
```

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2409 @@
...
+
+# How long to watch the operator log for that line before giving up. The
+# operator requeues a paused store every ~5s, so this is many chances over.
+PAUSE_LOG_TIMEOUT = 90
+PAUSE_LOG_INTERVAL = 10
+
+# After the annotation is removed, how long to let the operator pick the store
+# back up before asserting the skip lines have stopped. Several requeue periods,
+# so an in-flight skip logged just before the removal cannot fail the check.
+PAUSE_RESUME_SETTLE = 20
```
udaysk23, line 417: Move all constants at the start of the class or file

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2409 @@
...
+
+def get_store_pause_annotation(kind, name, namespace):
+    """Return the value of the noobaa.io/pause-reconcile annotation (or None)."""
+    annotations = (
+        OCP(kind=kind, namespace=namespace, resource_name=name)
+        .get()
+        .get("metadata", {})
+        .get("annotations", {})
+    )
+    return annotations.get(PAUSE_ANNOTATION)
```
udaysk23, line 428:
```suggestion
def get_store_pause_annotation(kind, name, namespace):
    """Return the value of the noobaa.io/pause-reconcile annotation (or None)."""
    return (
        OCP(kind=kind, namespace=namespace, resource_name=name)
        .get()
        .get("metadata", {})
        .get("annotations", {})
        .get(PAUSE_ANNOTATION)
    )
```

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2409 @@
...
+
+
+def get_noobaa_operator_pod_name(namespace):
```
udaysk23, line 431: There's an existing get_noobaa_operator_pod in ocs_ci/ocs/resources/pod.py that fetches the operator pod. It returns a Pod object rather than a name and doesn't filter by status, so it's not a drop-in — but if grabbing any operator pod is acceptable here, you could reuse it and drop this helper:

```
from ocs_ci.ocs.resources.pod import get_noobaa_operator_pod
operator_pod = get_noobaa_operator_pod(namespace=ns).name
```

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done — moved into each class's docstring.

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done — moved to the endpoint_pair fixture docstring.

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: The override schema is the only documentation of mcg_endpoint_pair anywhere in ocs-ci, so I kept it but moved it to the endpoint_pair fixture. Dropped the http/ClusterIP paragraph — it was already on _S3_HTTP_PORT.

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done — both kinds now build from the templates. Dropped _CR_KIND and NOOBAA_API_VERSION.

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done — all module constants are now in one block at the top.

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done — using get_noobaa_operator_pod(namespace=ns).name.

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2430 @@
...
+
+# NooBaa core's check_external_connection() reports a coarse UNKNOWN_FAILURE for
+# DNS failures, unreachable hosts and missing buckets alike - its AWS error map
+# is keyed on SDK v2 error names while the client is v3, so everything
+# client-side falls through to the catch-all (DFBUGS-10938). The real reason is
+# only in the error text, which is why the tests below assert on the message as
+# well as the code. INVALID_ENDPOINT is accepted too, so the tests keep passing
+# once DFBUGS-10938 is fixed and core starts classifying these properly.
+# TODO(DFBUGS-10938): once core reports precise codes, drop the message
+# assertions and pin each variant to its own status code.
```
udaysk23: Can we make it one liner comment

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2430 @@
...
+    """
+    matched = get_pod_logs(
+        pod_name=operator_pod,
+        namespace=namespace,
+        since=since,
+        grep=PAUSE_SKIP_LOG_PATTERN.format(name=re.escape(store_name)),
+        regex=True,
+        first_match_only=False,
+    )
+    return bool(matched and matched.strip())
```
udaysk23: ocs_ci/ocs/resources/pod.py already has search_pattern_in_pod_logs(pod_name, pattern, namespace=..., since=...), which searches a pod's logs for a regex within a since window — the same job this helper does. It returns the list of matched lines rather than a bool, but that's a one-liner to adapt:
```suggestion
    pattern = PAUSE_SKIP_LOG_PATTERN.format(name=re.escape(store_name))
    return bool(
        search_pattern_in_pod_logs(operator_pod, pattern, namespace=namespace, since=since)
    )
```

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2430 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
```
udaysk23: Please remove this comment

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2430 @@
...
+    "UNKNOWN FAILURE - https://redhat.atlassian.net/browse/DFBUGS-10938"
+)
+
+
+# ===========================================================================
+# Module A - positive / core behaviour
+# ===========================================================================
+@mcg
+@red_squad
+class TestBackingStoreEndpointUpdate:
```
udaysk23: The three test classes don't extend an OCS-CI base class.

TestBackingStoreEndpointUpdate, TestConnectionUpdateNegative, and TestNamespaceStoreEndpointUpdate are all declared as bare class Test...:. MCG tests are expected to extend MCGTest (from ocs_ci.framework.testlib — it chains MCGTest → ManageTest → BaseTest), which wires in the framework-level teardown/health behavior that runs for BaseTest subclasses on the shared CI cluster. Extending it also keeps this module consistent with the rest of the MCG suite.
```suggestion
from ocs_ci.framework.testlib import MCGTest

@mcg
@red_squad
class TestBackingStoreEndpointUpdate(MCGTest):
    ...

```

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2430 @@
...
+        # entirely and leak the store, which is how orphaned Rejected stores
+        # have been left on test clusters before.
+        try:
+            if get_store_pause_annotation(kind, name, namespace):
+                store_ocp.annotate(
+                    annotation="noobaa.io/pause-reconcile-", resource_name=name
+                )
+        except CommandFailed as ex:
+            logger.warning(f"Could not clear pause annotation on {kind}/{name}: {ex}")
+        deadline = time.time() + STORE_DELETE_TIMEOUT
```
udaysk23: _delete_store polls the webhook's "still being deleted" denial with a manual deadline = time.time() + STORE_DELETE_TIMEOUT / while True / time.sleep(STORE_DELETE_INTERVAL) loop (:759–773). TimeoutSampler is already imported (:54) and used elsewhere in this file (e.g. wait_for_stores_quiesced), and it would take over the deadline/sleep bookkeeping while you keep the retry-on-marker vs. re-raise logic inside the sampled call. Not blocking — just brings this loop in line with the rest of the module.

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2430 @@
...
+        """
+        ns = config.ENV_DATA["cluster_namespace"]
+        old, new = endpoint_conf["old"], endpoint_conf["new"]
+        bs = store_factory(
+            constants.BACKINGSTORE,
+            old,
+            endpoint_conf["target_bucket"],
+            endpoint_conf["secret"],
+            endpoint_conf["signature_version"],
+        )
```
udaysk23: The 5-arg store_factory(constants.BACKINGSTORE, <endpoint>, <bucket>, endpoint_conf["secret"], endpoint_conf["signature_version"]) call recurs ~13× across the tests, re-passing the same endpoint_conf["secret"] / signature_version tail each time. A thin local helper that unpacks endpoint_conf — e.g. _make_store(store_factory, endpoint_conf, endpoint, bucket, kind=constants.BACKINGSTORE) — would cut the repetition and let each call site show only what it actually varies (kind, endpoint, bucket).

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2430 @@
...
+    pause-reconcile annotation's lifecycle, switching under active I/O, and the
+    default backingstore.
+
+    The seven that need the update to succeed are skipped on clusters carrying a
+    Rook-keyed Ceph object-user secret, where DFBUGS-10975 aborts every update;
+    see ``DFBUGS_10975_SKIP``. They still run where no such secret exists, so
+    non-RGW clusters keep the coverage.
+    """
+
+    @tier1
```
udaysk23: Please check that we really need tier1 here or not

tests/functional/object/mcg/test_endpoint_update.py
```diff
@@ -0,0 +1,2430 @@
...
+                        "targetPort": http_port["targetPort"],
+                        "protocol": "TCP",
+                    }
+                ],
+            },
+        }
+    )
+    cluster_ip = OCP(kind="Service", namespace=namespace, resource_name=name).get()[
+        "spec"
+    ]["clusterIP"]
```
udaysk23: We can remove this code and read cluster_ip from svc_obj wherever needed like `svc_obj.get()["spec"]["clusterIP"]`

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done — cut to a single line: # Core lumps unrelated failures together (DFBUGS-10938), so tests assert the message. I applied the same rule to every other comment in the file, so there are no multi-line comment blocks left.

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done.

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done — removed this one and the five other `# ---- section ----` banners
for consistency.

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Checked — moved to tier2.

tests/functional/object/mcg/test_endpoint_update.py
ypersky1980: Done — added make_store(), 20 call sites collapsed. It also picks the
signature version from the store kind: the webhook rejects v4 on a
plain-http NamespaceStore but accepts it for a BackingStore, so a new
NamespaceStore call site can't get that wrong now.

udaysk23 (approved):

fbalak (approved):

PR #16137, opened 2026-09-07
Use incompressible fio data in FillPoolJob to raise Ceph used capacity

yitzhak12:
## Summary
  RBD discards `/dev/zero` writes, so FillPoolJob never raised Ceph used-raw and autoscaler tests never hit the
  scale threshold.
  Add `fill_mode=incompressible` (fio libaio, scrambled buffers, iodepth 16) on `quay.io/ocsci/fedora:fio`, and use
  it for autoscaler fill and the FillPoolJob libtest. `zero` and `random` (dd) are unchanged.
  ## Test plan
  - [ ] `tests/libtest/test_fill_pool_job.py` — used-raw increases, job completes
  - [ ] Storage autoscaler fill reaches the intended used-capacity percent

## Summary by CodeRabbit

- **New Features**
  - Added zero, random, and incompressible storage fill modes.
  - Incompressible mode uses Fedora FIO tooling to increase reported raw storage usage.
  - Fill jobs now recognize expected “no space left” conditions as successful completion.
  - Updated fill-job resource defaults and block sizes for larger workloads.

- **Bug Fixes**
  - Improved validation for unsupported fill modes and invalid storage sizes.
  - Improved reliability of package installation and file checks in pods.

- **Tests**
  - Updated storage scaling and fill-job tests for incompressible data generation and longer completion times.

### DIFF
--- a/ocs_ci/ocs/fill_pool_job.py
+++ b/ocs_ci/ocs/fill_pool_job.py
@@ -13,6 +13,20 @@ from ocs_ci.utility.utils import exec_cmd
 
 log = logging.getLogger(__name__)
 
+# RBD discards actual zero writes (discard_on_zeroed_write_same). /dev/urandom is
+# CPU-bound under the default CPU cap. fio libaio with scrambled buffers is stored
+# as real used capacity and can issue multiple outstanding IOs.
+FILL_MODES = ("zero", "random", "incompressible")
+
+
+def _fio_size_from_pvc_storage(storage):
+    """92% of a Gi PVC, so fio --size fits on the filesystem (fill_fs OOMs this image)."""
+    text = str(storage).strip().lower()
+    if not text.endswith("gi"):
+        raise ValueError(f"incompressible fill expects storage in Gi, got {storage!r}")
+    gib = float(text[:-2])
+    return f"{max(1, int(gib * 1024 * 0.92))}M"
+
 
 class FillPoolJob(object):
     """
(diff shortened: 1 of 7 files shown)


ocs_ci/ocs/fill_pool_job.py
```diff
@@ -13,6 +14,90 @@
 
 log = logging.getLogger(__name__)
 
+# RBD discards actual zero writes (discard_on_zeroed_write_same). /dev/urandom is
+# CPU-bound under the default CPU cap. fio libaio with scrambled buffers is stored
+# as real used capacity and can issue multiple outstanding IOs.
+FILL_MODES = ("zero", "random", "incompressible")
+
+# fio 3.21 in fedora:fio is OOMKilled at 1Gi. incompressible mode raises the
+# caller-supplied resources to at least these values; larger requests are kept.
```
AviadP, line 23: This code takes whatever cpu/memory the caller passed and bumps it up to a minimum. But no caller passes cpu/memory — all 3 call sites only pass fill_mode, storage, block_size. So the input is always create()'s defaults, which are always smaller than the minimums.

ocs_ci/templates/workloads/fio/fill_pool_job.yaml
```diff
@@ -13,24 +13,23 @@ spec:
       restartPolicy: Never
       containers:
       - name: filler
-        image: quay.io/ocsci/busybox:latest
+        # fedora:fio provides GNU dd and fio. FillPoolJob.create overrides command.
```
AviadP, line 16: create() overwrites image, resources, command and BLOCK_SIZE anyway, so nothing you changed in this YAML reaches the pod

PR #16201, opened 2026-09-10
Adding support to BM deployment on IBM Cloud VPC BM 

shivamdurgbuns:
Adding support to BM deployment on IBM Cloud VPC BM for this added new lb and dns creation/deletion capablility for AI deployment

## Summary by CodeRabbit

- **New Features**
  - Added support for IBM Cloud VPC bare-metal deployments using Assisted Installer.
  - Added a compact deployment option with three control-plane nodes and no worker nodes.
  - Added automatic API and ingress load balancer and DNS configuration for VPC environments.
  - Added direct iPXE-based server booting with improved node restart and recovery handling.
  - Added lifecycle management for VPC bare-metal servers, including start, stop, restart, and cleanup.
  - Existing classic infrastructure deployment behavior remains unchanged.


ocs_ci/deployment/baremetal.py
```diff
@@ -890,7 +891,70 @@ def deploy_prereq(self):
                     {"clusterName": self.cluster_name, "infraID": self.cluster_name},
                     metadata_file,
                 )
-            # load API and Ingress IPs from config
+            # IBM Cloud VPC BM: Create Network Load Balancers for VIPs
+            if config.ENV_DATA.get("platform") == constants.IBM_VPC_BM_PLATFORM:
```
petr-balogh:
```suggestion
            if config.ENV_DATA.get("infra_type", "classic") == constants.VPC_INFRA_TYPE:
```

ocs_ci/ocs/platform_nodes.py
```diff
@@ -116,6 +116,7 @@ def __init__(self):
             "baremetal_ai": IBMCloudBMNodes,
             "ibm_hci_ipi": IBMHCINode,
             "ibm_hci_agent": IBMHCIAgentNode,
+            "ibm_vpc_bm": IBMCloudVPCBMNodes,
```
petr-balogh, line 119: Here we need to do adjustments in get_nodes_platform
It properly return the platform key matching this one, which should be similarly like we do:
baremetal_ai
But we need to inject somehow vpc from infra_type there if it's VPC and then use this proper class `IBMCloudVPCBMNodes`

conf/deployment/baremetal/ai_1az_rhcos_vpc_compact_mode_3m_0w.yaml
```diff
@@ -0,0 +1,11 @@
+---
+# VPC Bare Metal - Compact Mode (3 masters, 0 workers)
+DEPLOYMENT:
+  local_storage: true
+
+ENV_DATA:
+  platform: 'baremetal'
+  deployment_type: 'ai'
+  infra_type: 'vpc'
```
dahorak: What about placing this configuration into the "environment specific credentials file"? So we could use the existing config like `conf/deployment/baremetal/ai_1az_rhcos_compact_mode_3m_0w.yaml`.

PR #16209, opened 2026-09-11
Add temporary submariner 0.24.1 lighthouse import-namespace-deny-list…

keemano:
Temp fix for service export's restriction issue.
… workaround

For globalnet RDR with submariner 0.24.1, create the submariner-lighthouse-agent configmap with import-namespace-deny-list set to "kube-" and restart the ocs-operator pods on both managed clusters after the multicluster service is enabled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multicluster disaster recovery deployment compatibility with Submariner 0.24.1 in globalnet environments.
  * Automatically applies the required namespace import configuration during applicable deployments.
  * Refreshes OCS operator pods so the updated settings take effect immediately.
  * Deployments using other Submariner versions remain unaffected.
  * Clusters that do not use the applicable Submariner version are skipped without changing their existing configuration.

### DIFF
--- a/ocs_ci/deployment/deployment.py
+++ b/ocs_ci/deployment/deployment.py
@@ -61,6 +61,7 @@ from ocs_ci.utility.iscsi_config import iscsi_setup
 from ocs_ci.framework.logger_helper import log_step
 from ocs_ci.helpers.dr_helpers import (
     configure_drcluster_for_fencing,
+    configure_submariner_lighthouse_import_namespace_deny_list,
     get_cluster_set_name,
     create_service_exporter,
     is_cg_enabled,
@@ -639,6 +640,11 @@ class Deployment(object):
                                 timeout=2000,
                                 ocs_registry_image=ocs_registry_image,
                             )
+                        # Temporary workaround for submariner 0.24.1: after the
+                        # multicluster service is enabled on both managed
+                        # clusters, create the submariner-lighthouse-agent
+                        # configmap and restart ocs-operator pods
+                        configure_submariner_lighthouse_import_namespace_deny_list()
                     config.reset_ctx()
                 if config.REPORTING["collect_logs_on_success_run"]:
                     try:
--- a/ocs_ci/helpers/dr_helpers.py
+++ b/ocs_ci/helpers/dr_helpers.py
@@ -4771,3 +4771,61 @@ def validate_cluster_odf_cli(retries=5, retry_interval=60):
         f"ODF DR validate clusters did not report success after {retries} attempts. "
         f"Output:\n{last_stdout}"
     )
+
+
+def configure_submariner_lighthouse_import_namespace_deny_list():
+    """
+    Temporary workaround required only for submariner version 0.24.1 in a
+    globalnet RDR setup, to be run after the multicluster service is enabled on
+    both managed clusters.
+
+    On each managed cluster, if the deployed submariner version is 0.24.1,
+    create the submariner-lighthouse-agent configmap with the
+    import-namespace-deny-list set to "kube-" and restart the ocs-operator pods
+    so that the change takes effect.
+
+    """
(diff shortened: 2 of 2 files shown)


ocs_ci/deployment/deployment.py
```diff
@@ -639,6 +640,11 @@ def do_deploy_ocs(self):
                                 timeout=2000,
                                 ocs_registry_image=ocs_registry_image,
                             )
+                        # Temporary workaround for submariner 0.24.1: after the
+                        # multicluster service is enabled on both managed
+                        # clusters, create the submariner-lighthouse-agent
+                        # configmap and restart ocs-operator pods
+                        configure_submariner_lighthouse_import_namespace_deny_list()
```
prsurve, line 647: How about we do this After submariner install

keemano: This is inside ocs install function. So this will be done post submariner installation.

prsurve: This will be done after patching of multiClusterService is done .

prsurve: Also this will fail for c2 as in your function it's applying for all non_acm cluster config

keemano: Yes. We are doing it after patching multiclusterservice. You want to move before that?
why will it fail in c2?

prsurve: What is the reason to do it after patching? Why to complicate things if we done this patching as soon as subamriner install in do_deploy_submariner it should work  fine

prsurve: after patching multiclusterservice it will validate_serviceexport and validation will fail and WA will never be applied

prsurve (dismissed):

sidhant-agrawal (dismissed):

ocs_ci/helpers/dr_helpers.py
```diff
@@ -4771,3 +4771,53 @@ def validate_cluster_odf_cli(retries=5, retry_interval=60):
...
+    import-namespace-deny-list set to "kube-".
+
+    """
+    # Local imports to keep this temporary workaround self-contained
+    from ocs_ci.utility import version
+
+    target_version = version.get_semantic_version("0.24.1")
+    restore_index = config.cur_index
+    try:
+        for cluster in get_non_acm_cluster_config():
```
jilju, line 4793: When the RDR pair is client clusters this will iterate over provider clusters as well and fail "oc create configmap submariner-lighthouse-agent" command because it isw already created while deploying provider-provider pair RDR.

shylesh (dismissed):
lgtm

ocs_ci/helpers/dr_helpers.py
petr-balogh: Should we simply check if config map is already in place and if so , continue and not create it again?

<del>Or instead of create - use apply - then this will not fail when resource already exists.</del>

Ah, above comment about apply will not work as we are not applying from file.

Gemini suggests something like:
```
exec_cmd(
    'oc create configmap submariner-lighthouse-agent '
    '--from-literal=import-namespace-deny-list="kube-" '
    f'-n {constants.SUBMARINER_OPERATOR_NAMESPACE} '
    '--dry-run=client -o yaml | oc apply -f -',
    use_shell=True
)
```

But more safe is just to check if config map exists and if so - do not create it I guess.

prsurve (approved):

petr-balogh (approved):

prsurve:
Verified :- https://url.corp.redhat.com/1ef3d30

ocs_ci/helpers/dr_helpers.py
jilju: Create if not present is the right option.

PR #16242, opened 2026-09-14
Fix ODF CLI failures in containerized Jenkins agents

petr-balogh:
Remove KUBERNETES_* environment variables at startup to prevent ODF CLI and other Kubernetes-aware tools from detecting the wrong cluster.

When ocs-ci runs in a containerized Jenkins agent that itself runs in a different OpenShift cluster, environment variables like KUBERNETES_SERVICE_HOST point to the Jenkins cluster, not the test cluster being validated.

This causes the ODF CLI (and potentially other tools) to incorrectly assume they're running inside the test cluster. The CLI then tries to connect directly to internal service addresses (e.g., noobaa-mgmt.openshift-storage.svc.cluster.local), which fail DNS resolution because the container only has access to the Jenkins cluster's DNS.

By removing these variables at initialization, the ODF CLI correctly detects it's running outside the cluster and automatically uses port-forwarding via localhost instead of attempting direct internal service connections.

This fix is applied globally in init_ocsci_conf() so all commands benefit, not just ODF CLI operations.

Fixes test failures with errors like:
  "dial tcp: lookup noobaa-mgmt.openshift-storage.svc.cluster.local on
   172.31.0.10:53: no such host"

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented configuration processing from using Kubernetes environment metadata inherited from Jenkins agents.


petr-balogh:
Verification job:
https://url.corp.redhat.com/90a120f

clacroix12 (approved):

dahorak (approved):

dahorak:
We should probably implement this also for deploy-fusion/deploy-fdf commands here: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/framework/fusion_fdf_init.py or even here: https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/utility/framework/base_initializer.py

PR #16287, opened 2026-09-16
[release-4.21] checking if it is a hci in case of storage setup (#14490)

clacroix12:
* checking if it is a hci in case of storage setup


petr-balogh (approved):

clacroix12:
Verified on the 4.18 cherry-pick: https://url.corp.redhat.com/fa6d372

PR #16292, opened 2026-09-17
Update SAN cluster teardown steps via CLI

posharma1:
## Summary by CodeRabbit

* **Tests**
  * Updated Fusion Access validation to use centralized SAN settings for iSCSI, image registry, and Quay configuration.
  * Updated the default image registry to `quay.io`.
  * Improved virtual machine lifecycle cleanup by verifying removal of filesystems, storage classes, local disks, and clusters.
  * Added polling and warning handling to accommodate delayed resource deletion during test teardown.
* **Documentation**
  * Updated SAN configuration guidance to reflect the current settings location and registry default.


tests/cross_functional/ui/fusion_access/test_virtualmachine_lifecycle.py
avd-sagare, line 61: Instead of adding fixture in testcase can we move it to common place and call here??

avd-sagare, line 45: Instead of adding fixture in testcase can we move it to common place and call here?

PR #16306, opened 2026-09-21
Use oc config new-admin-kubeconfig for permanent kubeconfig creation

dahorak:
Replace the oc login approach with oc config new-admin-kubeconfig to create permanently valid kubeconfig files. The previous implementation created kubeconfigs with tokens that expire after 24 hours. The new approach uses certificate-based authentication which never expires.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved kubeconfig creation by handling authentication failures explicitly and reporting login issues clearly.
  - Ensured failures while generating persistent kubeconfig files are surfaced instead of proceeding silently.
  - Applied the configured proxy URL to newly created kubeconfig files, improving connectivity for clients using those files.


dahorak:
Verified locally.

PR #16317, opened 2026-09-22
fix: make discover_hosted_clusters and automate_import_of_hosted_clus…

prsurve:
…ters idempotent

discover_hosted_clusters: blindly appending discoveryPrefix to customizedVariables caused a duplicate-key error on re-runs. Now checks if the entry already exists with the expected value and skips the patch; replaces a stale value if the entry exists with a different value.

automate_import_of_hosted_clusters: replaced create_resource (oc create) with apply_resource (oc apply) for all three resources (Policy, Placement, PlacementBinding) so the function no longer fails with AlreadyExists on re-runs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved hosted cluster discovery to update configuration only when necessary, avoiding redundant changes.
  - Corrected hosted cluster import automation so policies and placement bindings are consistently applied.
  - Preserved existing configuration when the required discovery value is already present.


shylesh (approved):
lgtm

prsurve:

jilju (approved):

PR #16318, opened 2026-09-22
acm: skip import for already-imported ManagedClusters

prsurve:
Add `is_cluster_already_imported()` helper that checks whether a ManagedCluster resource already exists and has
`ManagedClusterConditionAvailable=True` in ACM.

Use this check in both import paths:
- `import_clusters_via_cli`: skip the full import sequence (manifests, secret, wait, klusterlet) for any cluster that is already available.
- `import_clusters_with_acm` (UI path): skip `acm_nav.import_cluster()` if the cluster is already imported.

This makes repeated runs and partial-import scenarios idempotent instead of failing or duplicating import work.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate imports for clusters that are already registered and available.
  * Skipped unnecessary project creation when a cluster project already exists.
  * Ensured all configured clusters are evaluated during UI imports, rather than only the primary cluster.
  * Preserved the existing import flow for unavailable or unregistered clusters.
  * Continued to surface unexpected errors during cluster status checks.


ocs_ci/ocs/acm/acm.py
```diff
@@ -1113,10 +1144,15 @@ def import_clusters_with_acm():
     if config.DEPLOYMENT.get("ui_acm_import"):
         login_to_acm()
         acm_nav = AcmAddClusters()
-        acm_nav.import_cluster(
-            cluster_name=cluster_name_a,
-            kubeconfig_location=kubeconfig_a,
-        )
+        if is_cluster_already_imported(cluster_name_a):
```
shylesh: where is the check for another cluster ?  what if cluster_name_b has already been imported ?

prsurve: Good catch.  let me fix it

prsurve: fixed

shylesh (dismissed):
lgtm

prsurve:

amr1ta (approved):

amr1ta:
lgtm

PR #16324, opened 2026-09-22
fix(deployment): skip MCH apply and use CSV+deployment check for back…

prsurve:
…up pod validation

- In enable_cluster_backup(), detect if cluster-backup is already enabled in the MCH spec before patching. If already True, skip oc apply and wait_for_phase (which triggered a redundant 2-3hr wait on re-runs).

- Replace backup_pod_status_check() hardcoded pod count logic with:
  1. OADP CSV presence and Succeeded phase check
  2. Deployment availableReplicas == spec.replicas check for all deployments in open-cluster-management-backup namespace

  The old approach called get_oadp_version() which returned None while
  OADP was still installing, causing compare_versions("None >= 1.6") to
  raise WrongVersionExpression on every retry. Combined with the default
  backoff=2, this silently looped for ~2hr 20min before failing.

- Add backoff=1 to the @retry decorator on backup_pod_status_check() to keep the delay flat at 10s (19 * 10s = ~3min total) instead of growing exponentially to the 600s cap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cluster backup readiness checks by validating that the backup operator has completed successfully.
  * Confirmed backup deployments are present and all desired replicas are available before reporting readiness.
  * Avoided unnecessary resource updates when cluster backup is already enabled.
  * Provides clearer failure handling when the backup operator or deployments are missing or not ready.


prsurve:

vavuthu (approved):

sidhant-agrawal (approved):

amr1ta (approved):

PR #16328, opened 2026-09-22
fix(oadp): skip subscription creation in do_deploy_oadp if already pr…

prsurve:
…esent

When re-running deployment on a cluster where a prior attempt created the OADP subscription but the CSV had not yet appeared, the CSV-only check was insufficient to skip re-creation. The code would re-apply the namespace, re-apply the subscription, and wait 120s unnecessarily.

Add a subscription existence check after the CSV check so that any cluster with an existing OADP subscription is skipped immediately.


prsurve:

sidhant-agrawal (approved):

amr1ta (approved):
looking good

